PR #568, opened 2019-03-05
STYLE: Class derived from TransformBase need explicit instantiation

hjmjohnson:
Move definition code from hxx to cxx file.

These classes *must* be explicitly
instantiated for each valid template type to support
dynamic casting robustly (one-definiton-rule).

As a consequence, the only place that the definitions
for these class member functions is needed is in the
cxx compiled files.


hjmjohnson:
```
Total Test time (real) = 102.89 sec
	Add file: /home/circleci/ITK/.circleci/config.yml
	Add file: /home/circleci/ITK-dashboard/circleci.cmake
	Add file: /home/circleci/ITK-dashboard/itk_common.cmake
Submit files (using http)
   Using HTTP submit method
   Drop site:http://open.cdash.org/submit.php?project=Insight
   Error when uploading file: /home/circleci/ITK-build/Testing/20190305-2049/Build.xml
   Error message was: Failed to connect to open.cdash.org port 80: Connection timed out
   Problems when submitting via HTTP
Exited with code 255
```

PR #570, opened 2019-03-05
COMP: Fix warning in IO/DCMTK about unused variable

phcerdan:
```
Modules/IO/DCMTK/src/itkDCMTKImageIO.cxx:153:45: warning: unused parameter ‘name’ [-Wunused-parameter]
 bool DCMTKImageIO::CanWriteFile(const char *name)
```


phcerdan:
Already fixed... forget about it! :)

PR #573, opened 2019-03-06
BUG: Fix OFFIS_DMCTK_VERSION_NUMBER where wrunlock was introduced

phcerdan:
The change in the API of DcmDataDictionary was introduced in the commit:
https://github.com/DCMTK/dcmtk/commit/1e6aefdf8081263d192a9ed6a8241dac31401b47#diff-7f734e6993454e7b234c1f352bad0fb6

This corresponds to version 364 of DCMTK


hjmjohnson:
@phcerdan I thought I had that fixed already.  Sorry that it slipped through the cracks.

phcerdan:
Thanks @hjmjohnson, the fix was there, but the version was wrong by just one digit. The change on the API was introduced in 364, but the current fix was for (> 362). So if DMCTK version is 363 it was failing!

PR #580, opened 2019-03-09
BUG: Do no install any Eigen3 targets if ITK_USE_SYSTEM_EIGEN=ON

phcerdan:
Also added documentation about the logic having two Eigen3 targets,
and a minimal CMake example for usage in external modules.

Raised in: https://github.com/Microsoft/vcpkg/pull/5115#issuecomment-471109666


Modules/ThirdParty/Eigen3/CMakeLists.txt
```diff
@@ -12,18 +12,51 @@ set(_Eigen3_min_version 3.3)
...
-if(ITK_USE_SYSTEM_EIGEN)
-  set(_Eigen3_SYSTEM_OR_INTERNAL "Eigen3")
-else()
-  set(_Eigen3_SYSTEM_OR_INTERNAL "ITKInternalEigen3")
-endif()
+# Docs:
+# find_package(ITK REQUIRED COMPONENTS
+# ITKEigen3
+# )
+# When requiring finding this module, the following cmake is invoked:
```
thewtex: Not sure how to interpret this -- "When requiring finding this module" -> "When this module is a dependency"?

Modules/ThirdParty/Eigen3/CMakeLists.txt
```diff
@@ -12,18 +12,51 @@ set(_Eigen3_min_version 3.3)
...
+# find_package(Eigen3 CONFIG) if ITK_USE_SYSTEM_EIGEN=ON
+# Or
+# find_package(ITKInternalEigen3 CONFIG) otherwise
+#
+# ITKInternalEigen3 does NOT allow the use of #include <Eigen/Core>, it requires #include <itkeigen/Eigen/Core>
+# This logic is handled automatically by the macro used internally `#include ITK_EIGEN(Core)`
+
+# However, this does not scales well for external ITK modules that want
```
thewtex: scales -> scale

Modules/ThirdParty/Eigen3/CMakeLists.txt
```diff
@@ -12,18 +12,51 @@ set(_Eigen3_min_version 3.3)
...
+# to link to third party libraries that are already using Eigen3.
```
thewtex: that -> which

Modules/ThirdParty/Eigen3/CMakeLists.txt
phcerdan: Thanks a lot Matt.
I went for: When this module is required

Let me know if When this module is a dependency reads better.

PR #583, opened 2019-03-11
Update vnl fix testing path

hjmjohnson:


hjmjohnson:
@thewtex @dzenanz Thank you for the quick review.  This was a stumbling block for a project I was working on where an external tool was using ITK's internal testing infrastructure.

PR #585, opened 2019-03-13
ENH: SpatialObject Refactoring for ITKv5

aylward:
High level description of changes made:

- Eliminate IndexToX transforms. SpatialObjects exist purely in physical space coordinates
- Eliminate vnl TreeNode usage. SpatialObjects now track their transforms, parents, and children directly.
- Eliminate AffineGeometryFrame
- Eliminate OffsetTable computation
- Dimension renamed to ObjectDimension - for consistency
- AddSpatialObject() renamed to AddChild() - for consistency
- RemoveSpatialObject() renamed to RemoveChild() - for consistency
- Clear() renamed to ClearChildren() - for clarity and consistency
- AddChildrenToList added to speed parsing of children in tree
- GetSpatialObjectTypeAsString() is now GetClassNameAndDimension()
- PropertyType is templated over ScalarType, not float
- Converted API from using char * to using std::string
- SetParent and AddChild now create consistent trees and do not cause a spatial object to move in physical space (i.e., transforms are updated appropriately).
- RemoveChild and RemoveAllChildren fixed to remove all pointers to / from those children to / from the tree
- Helper functions simplify the specification of IsInside, ValueAt, and other computations that potentially traverse an SO tree.
- Derived classes typically only need to implement IsInside and ComputeObjectBoundingBox member functions. Logic for ValueAt and such is improved.


Examples/SpatialObjects/SpatialObjectTransforms.cxx
```diff
@@ -19,17 +19,55 @@
...
-//
-// \begin{figure} \center
-// \includegraphics[width=0.9\textwidth]{SpatialObjectTransforms}
-// \itkcaption[SpatialObject Transformations]{Set of transformations associated
-// with a Spatial Object}
-// \label{fig:SpatialObjectTransforms}
-// \end{figure}
-//
-// Figure~\ref{fig:SpatialObjectTransforms} shows our set of transformations.
+// transformations and "spaces" ssociated with a spatial object.
```
phcerdan: Great documentation!

Modules/Core/SpatialObjects/include/itkArrowSpatialObject.hxx
```diff
@@ -28,182 +27,162 @@ template< unsigned int TDimension >
...
-    PointType pnt2;
-    for ( unsigned int i = 0; i < TDimension; i++ )
-      {
-      pnt2[i] = pnt[i] + m_Length * m_Direction[i];
-      }
+  PointType pnt = this->GetPositionInObjectSpace();
 
-    pnt = this->GetIndexToWorldTransform()->TransformPoint(pnt);
-    pnt2 = this->GetIndexToWorldTransform()->TransformPoint(pnt2);
+  const_cast< typename Superclass::BoundingBoxType * >(
```
phcerdan: Maybe these const_casts can be avoided with either:
a) Create non const version of GetMyBoundingBoxInObjectSpace()
b) Make the bounding box protected, instead of private. So derived class can access to it with `this->m_MyBoundingBoxInObjectSpace.SetXXX(pnt)`
Not sure if there is a performance hit on creating protected members. Should be double checked.

Modules/Core/SpatialObjects/include/itkImageMaskSpatialObject.hxx
```diff
@@ -25,292 +25,150 @@
...
-              rit.GoToBegin(); //Skip to reverse end
-              break;
-              }
-            --rit;
-            }
-          rit.PreviousLine();
-          }
-        rit.PreviousSlice();
+        first = false;
+        this->GetMyBoundingBoxInObjectSpace()->SetMinimum( tmpPoint );
```
phcerdan: Related to the other comment, if this is already valid, the const_cast can be removed?

Examples/SpatialObjects/SpatialObjectTransforms.cxx
fbudin69500: The 'a' from associated was removed.

Modules/Core/SpatialObjects/include/itkEllipseSpatialObject.h
```diff
@@ -54,65 +53,61 @@ class ITK_TEMPLATE_EXPORT EllipseSpatialObject:
...
-  static constexpr unsigned int NumberOfDimension = TDimension;
+  static constexpr unsigned int ObjectDimension = TDimension;
 
   itkNewMacro(Self);
   itkTypeMacro(EllipseSpatialObject, SpatialObject);
 
   /** Set all radii to the same radius value.  Each radius is
    *  half the length of one axis of the ellipse.  */
-  void SetRadius(double radius);
+  void SetRadiusInObjectSpace(double radius)
```
fbudin69500: 2 comments concerning this function:
* Since the function was renamed, it would be important to also update the documentation and highlight that there are 2  spaces: World and Object.

* In `Set` functions, the input value should be compared to the current value. It should only update the current value and call `Modified` if the new value is different.

Modules/Core/SpatialObjects/include/itkSpatialObjectPoint.h
```diff
@@ -45,25 +51,27 @@ class ITK_TEMPLATE_EXPORT SpatialObjectPoint
...
-  const PointType & GetPosition() const;
+  /** Get the SpatialObjectPoint Id. */
+  int GetId() const
+  { return m_Id; }
 
-  /** Set the point object. Couldn't use macros for these methods. */
-  void SetPosition(const PointType & newX);
+  /** Set the point object. */
+  void SetPositionInObjectSpace(const PointType & newPositionInObjectSpace)
+  { m_PositionInObjectSpace = newPositionInObjectSpace; }
```
fbudin69500: Shouldn't this function call `Modified` if the new value is different from the old one?

Modules/Core/SpatialObjects/include/itkSpatialObjectPoint.h
```diff
@@ -45,25 +51,27 @@ class ITK_TEMPLATE_EXPORT SpatialObjectPoint
...
 
   template <typename... TCoordinate>
-  void SetPosition(
+  void SetPositionInObjectSpace(
```
fbudin69500: Same as above

Modules/Core/SpatialObjects/include/itkSpatialObjectPoint.h
```diff
@@ -73,56 +81,86 @@ class ITK_TEMPLATE_EXPORT SpatialObjectPoint
...
       };
-    m_X = coordinates;
+    m_PositionInObjectSpace = coordinates;
     }
 
+  /** Return a pointer to the point object. */
+  const PointType & GetPositionInObjectSpace() const
+  { return m_PositionInObjectSpace; }
+
+  void SetSpatialObject( SpatialObjectType * so )
```
fbudin69500: Call `Modified()`?

Examples/SpatialObjects/SpatialObjectTransforms.cxx
fbudin69500: I think explicitly writing World space and Object space would be more explicit than just "spaces".

Modules/Core/SpatialObjects/include/itkArrowSpatialObject.hxx
fbudin69500: Since the API is changed a lot throughout this Pull request, I agree with @phcerdan 
that it is a good time to make this API `const` consistent.

Modules/Core/SpatialObjects/include/itkSpatialObject.hxx
```diff
@@ -489,113 +670,137 @@ SpatialObject< TDimension >
...
+        m_FamilyBoundingBoxInObjectSpace->SetMaximum( pointMax );
+        break;
         }
+      }
+    }
 
-      if ( !bbDefined )
+  if( depth > 0 )
+    {
+    PointType pnt;
```
fbudin69500: Shadowing variable declared line 747

Modules/Core/SpatialObjects/include/itkPointBasedSpatialObject.hxx
```diff
@@ -24,21 +24,225 @@
...
+
+/** Set Points from a list */
+template< unsigned int TDimension, class TSpatialObjectPointType >
+void
+PointBasedSpatialObject< TDimension, TSpatialObjectPointType >
+::SetPoints( const SpatialObjectPointListType & newPoints )
+{
+  m_Points.clear();
+
+  SpatialObjectPointListType::const_iterator it = newPoints.begin();
```
fbudin69500: `typename` required

Modules/Core/SpatialObjects/include/itkSpatialObject.hxx
```diff
@@ -634,43 +869,214 @@ void
...
+  while ( it != itEnd )
+    {
+    id = (*it)->GetId();
+    it2 = ++it;
+    while( it2 != itEnd )
+      {
+      id2 = (*it2)->GetId();
+      if( id == id2 || id2 == -1 )
+        {
+        int id = this->GetNextAvailableId();
```
fbudin69500: `id` declaration shadows `id` declared line 1027

Modules/Core/SpatialObjects/include/itkArrowSpatialObject.hxx
aylward: I will create a protected member function that is non-const:  GetModifiableMyBoundingBox()

Modules/Core/SpatialObjects/include/itkImageMaskSpatialObject.hxx
aylward: This shouldn't be valid.  GetMyBoundingBoxInObjectSpace is declared as const, but was returning a non-const pointer to the member variable.   Fixed GetMyBoundingBoxInObjectSpace to return a const pointer (and be a const function) and added the GetModifiableMyBoundingBoxInObjectSpace as a protected member function.

Modules/Core/SpatialObjects/include/itkSpatialObjectPoint.h
aylward: Thanks!  Should use the ITK macro.   I'll make another pass through fixing these Set (and Get) calls that are paired to member variables.

Modules/Core/SpatialObjects/include/itkEllipseSpatialObject.h
aylward: This (and others) should be updated to use ITK Set Macros.   Thanks for catching it.    Went through a couple of iterations before deciding the relevant space (object or world) should be explicit in member functions and variables.   These functions are remnants of a best forgotten version :)

Modules/Core/SpatialObjects/include/itkSpatialObjectPoint.h
aylward: Actually, a "point" is a lightweight data container.  It is not an ITK-derived class, so it is not using Update/Modified or ITK macros.   I had thought about deriving it from an ITK point, but none of the API overlaps.  I think it is ok for this to not have Modified/Update calls, but I'm open to other suggestions.

Modules/Core/SpatialObjects/include/itkSpatialObjectPoint.h
aylward: As above, I think Points should remain non-ITK classes.  They are simply nearly atomic data containers.

Examples/SpatialObjects/SpatialObjectTransforms.cxx
aylward: Fixed

Modules/Core/SpatialObjects/include/itkSpatialObjectPoint.h
fbudin69500: Indeed. I hadn't realized that this SpatialObjectPoint class was different. I am not familiar enough with where it is used to know if it should be changed. It really depends if changing the value of the point should lead to updating ITK pipelines in which it is used. I'll let you make that decision.

aylward: I consider SpatialObjectPoints to be relatively static data items.   They are typically added to or removed from a pointbasedspatialobject - which causes that spatialobject's Modified() function to be called, so I do think we're doing the right thing as-is.

aylward:
Looks like all tests are passing except the format of the pull request.   I will update the documentation and commit it over the next 2-3 days, and then perhaps we can merge?

fbudin69500:
Thanks for this nice PR! The Python tests are not done running, but if they pass, I think we can go ahead and merge everything.

thewtex:
:green_heart: 

`ObjectSpace` or `WorldSpace`, in whatever space you at it, this is monumental! :-)

Please remove the `WIP:` from the PR title when ready to merge.

aylward:
I think it is now ready to squash-and-merge - perhaps keep only the first commit message.

aylward:
That is up to you.  I am fine either way.  Some of the commits are quite
irregular, but some may be helpful.

Thanks
S

---
Sent via phone.

dzenanz:
Maybe someone could go through the commits and squash smaller changes, while leaving bigger changes and different author's commits separate? E.g. there is a commit `COMP: Squash me toooooo` etc. I could do it if no one else wants to.

aylward:
That is fine with me.  Thanks!

---
Sent via phone.

dzenanz:
OK, I am squashing and rebasing.

dzenanz:
Squashed in #613.

PR #590, opened 2019-03-15
BUG: Use Experimental dashboard_model with master and release branch …

thewtex:
…builds

The Continuous dashboard model waits for additional builds. This is
causing the builds to timeout and be marked as "Cancelled".


blowekamp:
Ah, that explains what's going on the landing page with the Azure badges.

blowekamp:
If the build should still be placed in the continuous section using `set(dashboard_loop 0)` should work with the continuous model:
https://github.com/SimpleITK/SimpleITK/blob/master/Testing/CI/Azure/azure.cmake#L51

thewtex:
> If the build should still be placed in the continuous section using `set(dashboard_loop 0)` should work with the continuous model:
> https://github.com/SimpleITK/SimpleITK/blob/master/Testing/CI/Azure/azure.cmake#L51

Good idea. Let's try that...

PR #593, opened 2019-03-15
ENH: refactoring NarrowBandImageFilterBase to not use Barrier

dzenanz:
This is second-last filter to use `Barrier`. The last one is `ParallelSparseFieldLevelSetImageFilter`.


hjmjohnson:
@dzenanz All passing except the "CDash" Build analysis.  I don't understand why.

PR #594, opened 2019-03-16
DOC: Fix PR template not being displayed.

jhlegarreta:
Fix PR template not being displayed: move the PR template file one level
up to the `.github` folder.


jhlegarreta:
After many, many tests in a private repository, and a long exchange with the GitHub support team, and even if IMHO this is not clearly stated in the documentation, when a single PR template exists, it should apparently be placed directly in `.github` so that it gets actually recognized when opening a PR.

PR #596, opened 2019-03-16
DOC: Add emojis to GitHub community issue templates.

jhlegarreta:
Add emojis to GitHub issue templates.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,7 +1,7 @@
 ---
-name: Bug Report
+name: "Bug Report \U0001F41B"
 labels: type:Bug
-about: File a bug report to help us improve ITK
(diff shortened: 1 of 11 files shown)


jhlegarreta:
Inspired by:
https://github.com/PowerShell/PowerShell/issues/new/choose
https://github.com/angular/angular-cli/issues/new/choose

The YAML markup does not accept markdown-syntax emojis, and the unicode code points need to be used:
https://unicode.org/emoji/charts/full-emoji-list.html

However, the full list is not supported, and I found no document explaining which ones are actually supported.

Files are kept as text files.

The double inverted commas are necessary for templates to be correctly interpreted when adding the unicode code points (the ones in the `about:` field are not necessary, but have been added for the sake of consistency).

jhlegarreta:
> However, the full list is not supported

Needless to say that the ones used in this PR (should) work since I have tested them in a private repository.

PR #601, opened 2019-03-19
BUG: Returns a copy of the VNL matrix instead of a reference

fbudin69500:
When getting a VNL matrix from an ITK matrix, a reference is returned. This
creates problem in Python if on uses the variable multiple times in one line
(e.g. m2=m*m).
To work araound this problem, we define a Python function that copies the
VNL matrix. Another solution would have been to use a SWIG `typemap`.


fbudin69500:
Superseded by #605

PR #603, opened 2019-03-20
ENH: Add helper functions to convert NumPy arrays to ITK matrices and…

fbudin69500:
… back

The new helper functions are wrappers around the ITK Matrix to VNL matrix
(and back).

### DIFF
--- a/Wrapping/Generators/Python/itkExtras.py
+++ b/Wrapping/Generators/Python/itkExtras.py
@@ -371,6 +371,21 @@ def GetVnlMatrixFromArray(arr):
 
 vnl_matrix_from_array = GetVnlMatrixFromArray
 
+def GetArrayFromMatrix(itk_matrix):
+    return GetArrayFromVnlMatrix(itk_matrix.GetVnlMatrix().as_matrix())
+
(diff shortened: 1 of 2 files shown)


Wrapping/Generators/Python/itkExtras.py
```diff
@@ -371,6 +371,17 @@ def GetVnlMatrixFromArray(arr):
 
 vnl_matrix_from_array = GetVnlMatrixFromArray
 
+def GetArrayFromITKMatrix(itk_matrix):
+    return GetArrayFromVnlMatrix(itk_matrix.GetVnlMatrix().as_matrix())
+
+def GetITKMatrixFromArray(arr):
```
thewtex: Awesome !! Thanks, @fbudin69500 !

`GetMatrixFromArray` would be shorter, and the implicit `ITK` I think makes sense since this is an ITK API and it is called with `itk.`.

Also

```
matrix_from_array = GetMatrixFromArray
array_from_matrix = GetArrayFromMatrix
```
Python aliases would we nice.

fbudin69500: Done.

PR #604, opened 2019-03-20
ENH: Add `ttype` keyword in New() function for Python templated classes

fbudin69500:
`ttype` keyword allows the user to specify the type of the object to
instantiate directly as a parameter of the `New()` function call.
This is especially useful to be able to use ITK's functional interface
with filters that require to set a type. It is similar to NumPy `dtype`.


fbudin69500:
@thewtex This has been corrected.

PR #606, opened 2019-03-20
ENH: Add helper functions to convert NumPy arrays to ITK matrices and…

fbudin69500:
… back

The new helper functions are wrappers around the ITK Matrix to VNL matrix
(and back).


fbudin69500:
#603 merged.

PR #607, opened 2019-03-20
ENH: Add `ttype` keyword in New() function for Python templated classes

fbudin69500:
`ttype` keyword allows the user to specify the type of the object to
instantiate directly as a parameter of the `New()` function call.
This is especially useful to be able to use ITK's functional interface
with filters that require to set a type. It is similar to NumPy `dtype`.


Wrapping/Generators/Python/itkTemplate.py
```diff
@@ -403,7 +403,14 @@ def New(self, *args, **kwargs):
             # to deal with checking if both keyword arguments are given.
             return self._NewImageReader(itk.ImageSeriesReader, True, 'FileNames', *args, **kwargs)
         primary_input_methods = ('Input', 'InputImage', 'Input1')
-        if len(args) != 0:
+        if 'ttype' in kwargs and keys:
+            type_params = kwargs.pop('ttype')
```
manuelcastr: If a user passes the parameters as anything other than a **list()**, say: a **tuple()** or even a NumPy array, semantically it wouldn't be wrong but it would fail on the comparison on line 412.
Is this something that must be taken into account by the user or is it something that can be addressed here by writing that line as:
```suggestion
            type_params = list(kwargs.pop('ttype'))
```

fbudin69500: Great idea. I have updated the PR taking your suggestion into account, but doing it the other way around. The keys actually are tuples, so I am converting the `ttype` argument into a tuple. If the user gives a list, which they are likely to do since it would look very similar to what they are currently typing when creating an ITK template object (e.g. `itk.Image[itk.F, 3].New()`), it should still work.

fbudin69500:
#604 merged.

PR #610, opened 2019-03-20
ENH: Wrap ExtrapolateImageFunction

thewtex:


thewtex:
Yes, we are only wrapping `double` here for TCoordRep as with the other `itk::ImageFunction`'s:

https://github.com/InsightSoftwareConsortium/ITK/blob/c06eb6859b561ae1fb7aca8149e46769cbfe6e99/Modules/Core/ImageFunction/include/itkExtrapolateImageFunction.h#L42

PR #617, opened 2019-03-22
ENH: `isinstance()` can be called without template parameters of clas…

fbudin69500:
…sinfo

`isinstance()` could already be called on ITK objects if the developer
was correctly comparing the object to a class with template arguments:

obj = itk.MedianImageFilter.IF2IF2.New()
isinstance(obj, itk.MedianImageFilter.IF2IF2)

Now, a developer can simply write:

obj = itk.MedianImageFilter.IF2IF2.New()
isinstance(obj, itk.MedianImageFilter)

This is especially convenient when trying to abstract from the template
parameters like when using `imread()` or calling filters using
template parameter deduction.


fbudin69500:
#616 merged.

PR #623, opened 2019-03-22
COMP: Allow compiler option fixed in gcc 4.1.0

hjmjohnson:
GCC version 4.0.4 had a bug in the -frename-registers
option used by -funroll-loops.

As of GCC version 4.1.0 and greater, this optimiation
was fixed.

ITK should no longer prohibit this optimization.

GCC 4.0.4 documentation
  -frename-registers
  Attempt to avoid false dependencies in scheduled code by making use of registers
  left over after register allocation. This optimization will most benefit processors
  with lots of registers. Depending on the debug information format adopted by
  the target, however, it can make debugging impossible, since variables will no
  longer stay in a “home register”.
 *** Not enabled by default at any level because it has known bugs.

GCC 4.1.0 documentation
  -frename-registers
  Attempt to avoid false dependencies in scheduled code by making use of registers
  left over after register allocation. This optimization will most benefit processors
  with lots of registers. Depending on the debug information format adopted by
  the target, however, it can make debugging impossible, since variables will no
  longer stay in a “home register”.
 *** Enabled by default with ‘-funroll-loops’.

### DIFF
--- a/CMake/ITKSetStandardCompilerFlags.cmake
+++ b/CMake/ITKSetStandardCompilerFlags.cmake
@@ -292,21 +292,6 @@ macro(check_compiler_platform_flags)
     set(ITK_REQUIRED_LINK_FLAGS "${ITK_REQUIRED_LINK_FLAGS} -mthreads")
   endif()
 
-
-  #-----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


dzenanz:
Making way towards having SSE4.2 required by default?

PR #625, opened 2019-03-22
PERF: Allow overwriting ABI/Optimization & Warning flags

hjmjohnson:
Provide a mechanism for changing the default warning
and ABI/OPTIMIZATION flags.

The default ABI/Optimization flag settings are
set to:
```bash
    -mtune=native # Tune the code for the computer used compile
                  # ITK, but allow running on generic cpu
                  # archetectures
    -march=corei7 # Use ABI settings to support corei7
                  # (circa 2008 ABI feature sets)
```
The defaults can easily be overriden from the cmake comand line with:
```bash
  -DITK_C_OPTIMIZATION_FLAGS:STRING="-march=native"
  -DITK_CXX_OPTIMIZATION_FLAGS:STRING="-march=native"
  -DITK_C_WARNING_FLAGS:STRING="-Wall -Werror"
  -DITK_CXX_WARNING_FLAGS:STRING="-Wall -Werror"
```

### DIFF
--- a/CMake/ITKSetStandardCompilerFlags.cmake
+++ b/CMake/ITKSetStandardCompilerFlags.cmake
@@ -26,7 +26,7 @@ if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.14.0)
   check_pie_supported()
 endif()
 
-function(check_c_compiler_warning_flags c_flag_var)
+function(check_c_compiler_flags c_flag_var)
   set(local_c_flags "")
   set(flag_list "${ARGN}")
   foreach(flag IN LISTS flag_list)
@@ -40,7 +40,7 @@ function(check_c_compiler_warning_flags c_flag_var)
 endfunction()
 
 
-function(check_cxx_compiler_warning_flags cxx_flag_var)
+function(check_cxx_compiler_flags cxx_flag_var)
   set(local_cxx_flags "")
   set(flag_list "${ARGN}")
   foreach(flag IN LISTS flag_list)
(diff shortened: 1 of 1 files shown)


hjmjohnson:
For all test, an initial indication on my laptop (not well-controlled experiment) indicates
a 5% speed increase across all tests.

The most important part of this PR is that you can now modify the default values for ABI and Warnings without the defaults clobbering your desired flags.

CMake/ITKSetStandardCompilerFlags.cmake
```diff
@@ -130,14 +130,62 @@ function(check_compiler_warning_flags c_warning_flags_var cxx_warning_flags_var)
...
   set(${c_warning_flags_var} "${CMAKE_C_WARNING_FLAGS}" PARENT_SCOPE)
   set(${cxx_warning_flags_var} "${CMAKE_CXX_WARNING_FLAGS}" PARENT_SCOPE)
 endfunction()
 
 
+function(check_compiler_optimization_flags c_optimization_flags_var cxx_optimization_flags_var)
+  set(${c_optimization_flags_var} "" PARENT_SCOPE)
+  set(${cxx_optimization_flags_var} "" PARENT_SCOPE)
+
+  if( WIN32 )
```
thewtex: This should just be MSVC since these flags will be different with MinGW64.

CMake/ITKSetStandardCompilerFlags.cmake
```diff
@@ -130,14 +130,62 @@ function(check_compiler_warning_flags c_warning_flags_var cxx_warning_flags_var)
...
+  # Check this list on C++ compiler only
+  set(cxx_flags "")
+
+  # Check this list on both C and C++ compilers
+  set(c_and_cxx_flags
+     # https://gcc.gnu.org/onlinedocs/gcc-4.8.0/gcc/i386-and-x86_002d64-Options.html
+     # NOTE the corei7 release date was 2008
+     ${InstructionSetOptimizationFlags}
+     -mtune=native # Tune the code for the computer used compile ITK, but allow running on generic cpu archetectures
+     -march=corei7 # Use ABI settings to support corei7 (circa 2008 ABI feature sets)
```
thewtex: :+1:

CMake/ITKSetStandardCompilerFlags.cmake
```diff
@@ -130,14 +130,62 @@ function(check_compiler_warning_flags c_warning_flags_var cxx_warning_flags_var)
...
+function(check_compiler_optimization_flags c_optimization_flags_var cxx_optimization_flags_var)
+  set(${c_optimization_flags_var} "" PARENT_SCOPE)
+  set(${cxx_optimization_flags_var} "" PARENT_SCOPE)
+
+  if(MSVC)
+      set(InstructionSetOptimizationFlags /arch:SSE /arch:SSE2 /arch:/AVX ) # /arch:/AVX2
+      set(c_and_cxx_flags
+         # https://docs.microsoft.com/en-us/cpp/build/reference/arch-x64?view=vs-2015
+         ${InstructionSetOptimizationFlags}
+         /arch:SSE /arch:SSE2 /arch:/AVX # Circa 2008 CPU's use /arch:/AVX2 for circa 2013 CPU support
```
dzenanz: Line 150 is a repeat of either line 149 or 146.

CMake/ITKSetStandardCompilerFlags.cmake
```diff
@@ -130,14 +130,62 @@ function(check_compiler_warning_flags c_warning_flags_var cxx_warning_flags_var)
...
 endfunction()
 
 
+function(check_compiler_optimization_flags c_optimization_flags_var cxx_optimization_flags_var)
+  set(${c_optimization_flags_var} "" PARENT_SCOPE)
+  set(${cxx_optimization_flags_var} "" PARENT_SCOPE)
+
+  if(MSVC)
+      set(InstructionSetOptimizationFlags /arch:SSE /arch:SSE2 /arch:/AVX ) # /arch:/AVX2
+      set(c_and_cxx_flags
```
dzenanz: `c_and_cxx_flags` set here get overwritten on line 172.

CMake/ITKSetStandardCompilerFlags.cmake
```diff
@@ -130,14 +130,62 @@ function(check_compiler_warning_flags c_warning_flags_var cxx_warning_flags_var)
...
+
+  # Check this list on C++ compiler only
+  set(cxx_flags "")
+
+  # Check this list on both C and C++ compilers
+  set(c_and_cxx_flags
+     # https://gcc.gnu.org/onlinedocs/gcc-4.8.0/gcc/i386-and-x86_002d64-Options.html
+     # NOTE the corei7 release date was 2008
+     ${InstructionSetOptimizationFlags}
+     -mtune=native # Tune the code for the computer used compile ITK, but allow running on generic cpu archetectures
```
dzenanz: Will this be ignored on Visual Studio?

CMake/ITKSetStandardCompilerFlags.cmake
```diff
@@ -304,16 +352,59 @@ macro(check_compiler_platform_flags)
   endif()
 endmacro()#End the platform check function
 
+if(NOT ITK_C_WARNING_FLAGS OR NOT ITK_CXX_WARNING_FLAGS ) # Only check once if not explicitly set on command line
+  #-----------------------------------------------------------------------------
+  #Check the set of warning flags the compiler supports
+  check_compiler_warning_flags(C_WARNING_FLAGS CXX_WARNING_FLAGS)
+endif()
+if(NOT ITK_C_WARNING_FLAGS) #Not set on cmake command line option -DITK_C_WARNING_FLAGS:STRING=""
+  set(ITK_C_WARNING_FLAGS ${C_WARNING_FLAGS} CACHE STRING "ITK C Compiler Warning flags, Modify to suit your needs.")
```
dzenanz: Either put `.` instead of `,`, or modify should be lowercase. Also copy-pasted a few time below.

CMake/ITKSetStandardCompilerFlags.cmake
```diff
@@ -304,16 +352,59 @@ macro(check_compiler_platform_flags)
...
 # Append ITK warnings to the CMake flags.
 # We do not set them in ITK_REQUIRED FLAGS because all project which
 # use ITK don't require these flags .
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${C_WARNING_FLAGS}")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CXX_WARNING_FLAGS}")
 
-#-----------------------------------------------------------------------------
-#Check the set of platform flags the compiler supports
-check_compiler_platform_flags()
+string(REPLACE " " ";" CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${ITK_C_OPTIMIZATION_FLAGS} ${ITK_C_WARNING_FLAGS}")
```
dzenanz: Could `list (APPEND CMAKE_C_FLAGS ${ITK_C_OPTIMIZATION_FLAGS}  ${ITK_C_WARNING_FLAGS})` be used instead of this acrobatics?

CMake/ITKSetStandardCompilerFlags.cmake
hjmjohnson: I cleaned this up to be a little more clear, and fixed indenting.

CMake/ITKSetStandardCompilerFlags.cmake
hjmjohnson: Line 152 is the else part, so Yes it will be ignored.  I fixed indentation of lines 165-177 to reflect the else indentation problem.

CMake/ITKSetStandardCompilerFlags.cmake
hjmjohnson: cleaned, and reorganized to remove duplications.

CMake/ITKSetStandardCompilerFlags.cmake
hjmjohnson: @dzenanz I tried what you recommend, but it did not seem to work.  I hate this acrobatics, but it was what could work.

I tried what you recommended but my testing failed.

PR #626, opened 2019-03-23
ENH: Update a few remote modules to their latest commit versions

jhlegarreta:
Update a few remote modules to their latest commit versions.

### DIFF
--- a/Modules/Remote/AnisotropicDiffusionLBR.remote.cmake
+++ b/Modules/Remote/AnisotropicDiffusionLBR.remote.cmake
@@ -20,5 +20,5 @@ itk_fetch_module(AnisotropicDiffusionLBR
     http://insight-journal.org/browse/publication/953
   "
   GIT_REPOSITORY ${git_protocol}://github.com/InsightSoftwareConsortium/ITKAnisotropicDiffusionLBR.git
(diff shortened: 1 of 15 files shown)


jhlegarreta:
Let's see if this helps bringing the [Expected Nightly Remotes in CDash](http://testing.cdash.org/index.php?project=Insight#!#Expected_Nightly_Remote_Modules) to the green.

gdevenyi:
Shouldn't these be done separately?

jhlegarreta:
@gdevenyi There is no fixed rule. See the following commits as an example: https://github.com/InsightSoftwareConsortium/ITK/commit/fbfabd8cbdf331b5ca203034b86a0266ef572905, https://github.com/InsightSoftwareConsortium/ITK/commit/4792824e3dbb163ab2d9ee1ab07502b73f01c67d

jhlegarreta:
Also, and although I may have missed some in this PR, updating each separately brings an overhead in terms of PRs that may not pay off. IMHO.

jhlegarreta:
@gdevenyi This should facilitate maintaining the remote modules up-to-date in ITK: https://github.com/InsightSoftwareConsortium/ITK/pull/632

PR #629, opened 2019-03-23
ENH: Remove CircleCI config files

jhlegarreta:
Remove CircleCI config files: Azure Pipelines are being used for CI since
commit 325fb75.


jhlegarreta:
I'd dare to say that Circle CI builds for the project are still enabled. We should may be deactivate them? Same thing for the SW Guide?

thewtex:
We recently (last Thursday) reduced the amount of CircleCI resources. If builds start to excessively queue, we can remove them. Until that time, we should keep them around because they are generally helpful.

Please re-open if we start to see a large queue.

PR #632, opened 2019-03-25
ENH: Add a script to update the remote module commit hashes

jhlegarreta:
Add a script to update the remote module commit hashes.

Update the `Release.md` guide to add the remote module update step prior
to a release.

### DIFF
--- a/Documentation/Maintenance/Release.md
+++ b/Documentation/Maintenance/Release.md
@@ -327,6 +327,22 @@ Add any new remote modules to nightly builds. Some builds may be difficult to ad
 due to third-party dependencies.
 
 
+Update Remote Modules
+---------------------
+
+In order to have the latest versions for all remote modules, and have them use
+the latest ITK tag, the following steps should be performed:
+
+1. Update the ITK tag used in the `azure-pipelines.yml` CI configuration and the
+`setup.py` Python setup files using the [](https://github.com/InsightSoftwareConsortium/ITK/tree/master/Utilities/Maintenance/)
+script. This will involve merging a new pull request to each remote module
+repository.
+
+2. Update the remote modules to their latest commits using the
+[UpdateRemoteModules.sh](https://github.com/InsightSoftwareConsortium/ITK/tree/master/Utilities/Maintenance/UpdateRemoteModules.sh)
+script.
+
+
 Create Tarballs
 ---------------
 
--- a/Utilities/Maintenance/UpdateRemoteModules.sh
+++ b/Utilities/Maintenance/UpdateRemoteModules.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+
+#==========================================================================
+#
+#   Copyright Insight Software Consortium
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#          http://www.apache.org/licenses/LICENSE-2.0.txt
+#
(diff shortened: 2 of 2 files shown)


jhlegarreta:
As discussed some time ago by mail.

A few comments:
- Have not tested if it would fail if for some reason the remote repository URL does not have a trailing _.git_
- May be the coding style is not too bash-like, so feel free to suggest changes.

Utilities/Maintenance/UpdateRemoteModules.sh
```diff
@@ -0,0 +1,45 @@
...
+  echo "  param1 Remote modules' path"
+  exit 1
+}
+
+# Display usage
+if [ ${#@} != 1 ]; then
+  usage
+fi
+
+remote_modules_path=$1
```
thewtex: We could use `BASH_SOURCE` and the relative path of *ITK/Modules/Remote* from this script to avoid the need for this argument -- it will always be the same relative directory. See e.g.:

https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/GitSetup/setup-upstream#L32

Utilities/Maintenance/UpdateRemoteModules.sh
```diff
@@ -0,0 +1,45 @@
...
+
+git_repository_tag_label='GIT_REPOSITORY'
+git_tag_label='GIT_TAG'
+
```
thewtex: We could check here whether `jq` is available, and if it is not, provide a suggestion for how to install it!

thewtex:
This also provides a dependency-free way to get the latest commit on GitHub:

https://stackoverflow.com/questions/15677439/command-to-get-latest-git-commit-hash-from-a-branch

Utilities/Maintenance/UpdateRemoteModules.sh
jhlegarreta: Done in b68921c.

Utilities/Maintenance/UpdateRemoteModules.sh
jhlegarreta: Done in b68921c.

jhlegarreta:
> This also provides a dependency-free way to get the latest commit on GitHub:
> 
> https://stackoverflow.com/questions/15677439/command-to-get-latest-git-commit-hash-from-a-branch

Nice ! Done in e9d1318 and a variable name error fixed !

Thanks both.

Documentation/Maintenance/Release.md
```diff
@@ -327,6 +327,22 @@ Add any new remote modules to nightly builds. Some builds may be difficult to ad
...
 
+Update Remote Modules
+---------------------
+
+In order to have the latest versions for all remote modules, and have them use
+the latest ITK tag, the following steps should be performed:
+
+1. Update the ITK tag used in the `azure-pipelines.yml` CI configuration and the
+`setup.py` Python setup files using the [](https://github.com/InsightSoftwareConsortium/ITK/tree/master/Utilities/Maintenance/)
+script. This will involve merging a new pull request to each remote module
```
jhlegarreta: Used the term **involve** because the commit/PR is not yet automatic.

Eventually a `Documentation/Maintenance/UpdatingRemotes.md` could be created, much like it exists for the `ThirdParty` modules, and that document could be linked from here. But that is left for an eventual future PR.

jhlegarreta:
Added a note in the `Release.md` document concerning the update of the `Remote` modules.

phcerdan:
This is great @jhlegarreta!

PR #633, opened 2019-03-25
ENH: Add a script to update the ITK tag in Python `setup.py` files

jhlegarreta:
Add a script to update the ITK tag in Python `setup.py` files.

### DIFF
--- a/Utilities/Maintenance/UpdateITKGitTagPythonSetup.sh
+++ b/Utilities/Maintenance/UpdateITKGitTagPythonSetup.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+program_name=$0
+
+function usage {
+  echo "Usage: ${program_name} "
+  echo "  param1 Setup.py file path"
+  exit 1
+}
+
+# Display usage
+if [ ${#@} != 1 ]; then
+  usage
+fi
+
+filename=$1
+
+# Get the latest ITK git tag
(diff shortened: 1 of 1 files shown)


jhlegarreta:
A few comments:
- `jq` must be available.
- Assuming that the requirement filed is not empty.
- Assuming that no changes are made to the ``=` or `>=` symbol (although we will want to change to the latter I guess).
- Assuming all tags have the _v_ prefix.
- We may need to ask for the username so that the user is authenticated in the `curl` command. I think there is a limit in the anonymous gets.
- I guess generating a branch and automatically committing/pushing as @hjmjohnson has done in some scripts is also doable, but this is a first version.

Utilities/Maintenance/UpdateITKGitTagPythonSetup.sh
```diff
@@ -0,0 +1,37 @@
...
+  echo "  param1 Setup.py file path"
+  exit 1
+}
+
+# Display usage
+if [ ${#@} != 1 ]; then
+  usage
+fi
+
+filename=$1
```
dzenanz: I guess that same remark from remote update applies here regarding relative path to `Setup.py` file.

jhlegarreta: Thanks @dzenanz . Answering in #641.

jhlegarreta:
Superseded by #641 for the reasons stated in #634.

PR #634, opened 2019-03-25
ENH: Add a script to update the ITK tag in Azure Pipelines CI config …

jhlegarreta:
…files

Add a script to update the ITK tag in Azure Pipelines CI config files.


jhlegarreta:
A few comments:
- `jq` must be available.
- We may need to ask for the username so that the user is authenticated in the `curl` command. I think there is a limit in the anonymous gets.
- I guess generating a branch and automatically committing/pushing as @hjmjohnson has done in some scripts is also doable, but this is a first version.

jhlegarreta:
Superseded by #641 to address @thewtex 's comment more cleanly.
Closing this PR.

PR #637, opened 2019-03-27
BUG: Python3 compatibility

maekclena:
In Python3 the keys of an `OrderedDict` are of type `odict_keys` and cannot use indexing.

These changes allow `PythonTemplateTest.py` to pass with Python3.


fbudin69500:
@maekclena Thanks for the contribution! I created a PR to run Python tests (#639). It is not ideal, but currently, when updating something that needs Python testing, you need to create two PRs: a PR against the `python-builds` branch and one against `master`. The one against `master` is still required though to merge the modifications once all the tests pass.

fbudin69500:
Test failing is not related to this PR.

PR #638, opened 2019-03-27
STYLE: ImageFileReader uses std::unique_ptr<[]>, try-catch-throw removed

N-Dekker:
Simplified `ImageFileReader::GenerateData()` implementation by declaring
its local `loadBuffer` as `const std::unique_ptr<char[]>`, instead of as
a raw `char` pointer.

Removed try-catch-throw and `delete[]` statements that are no longer necessary
now, due to the C++ "RAII" idiom implemented by `std::unique_ptr`.

Replaced a `std::copy` by `std::copy_n`, to further simplify the code.

### DIFF
--- a/Modules/IO/ImageBase/include/itkImageFileReader.hxx
+++ b/Modules/IO/ImageBase/include/itkImageFileReader.hxx
@@ -26,6 +26,7 @@
 #include "itkVectorImage.h"
 
 #include "itksys/SystemTools.hxx"
+#include <memory>  // For unique_ptr
 #include <fstream>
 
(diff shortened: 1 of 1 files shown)


N-Dekker:
The next amend (force-push) will just adjust the commit text a little bit.

N-Dekker:
When reviewing this commit, you might want to ignore lines that only have white space differences:
https://github.blog/2018-05-01-ignore-white-space-in-code-review/

When removing the try-catch, I had to adjust the indentation, that's why there are those white space differences!

PR #641, opened 2019-03-28
ENH: Add a script to update the ITK tag as required in files

jhlegarreta:
Add a script to update the ITK tag as required in
- `azure-pipelines.yaml` Azure pipelines configuration file.
- `setup.py` Python setup files.

usually in the `Remote` modules.

## PR Checklist
- [x] Adds the License notice to new files.
- [ ] Adds Python wrapping.
- [ ] Adds Documentation.

### DIFF
--- a/Utilities/Maintenance/UpdateRequiredITKVersionInFiles.sh
+++ b/Utilities/Maintenance/UpdateRequiredITKVersionInFiles.sh
@@ -1,0 +1,69 @@
+#!/bin/bash
+
+#==========================================================================
+#
+#   Copyright Insight Software Consortium
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


jhlegarreta:
> I guess that same remark from remote update applies here regarding relative path to Setup.py file.

@dzenanz I do agree that this would make things easier and cleaner.

That would require cloning each remote, may be using CMake to checkout them out. I usually do the editions to the remotes outside my ITK source tree, and I am not too familiar with dealing repositories or submodules inside another repository. I would need to investigate how they are cloned, whether cloning within the ITK folder creates a per-remote git history and corresponding branches, etc.

Also, the **relative path solution assumes that all** `azure-pipelines.yml` **files are all in the same location**. Although we will eventually get there, it is not the case yet.

**I haven't got the bandwith to address the relative location improvement right now**, so for now I will stick to the input arguments. 

But at some point in time (i.e. not in the short term) I will investigate the above and propose the changes to the script. That should also facilitate using scripts that apply changes over all modules.

PR #643, opened 2019-03-28
ENH: Cleanup python wrapping tests

### LINKED ISSUE #96: Re-enable the `PythonGetNameOfClass` test
This test currently crashes unreliably with

```
itkTestDriver: Process exception: Illegal instruction
```

on macOS

or

```
itkTestDriver: Process exception: Segmentation fault
```

on Linux.

maekclena:
- Remove last remaining modules tests
 - Remove outdated "notYetUsable" tests
 - Remove Numpy condition, tests work fine without it
 - Update `exclude-filters.txt`
 - Rewrite and enable `findEmptyClasses.py`
 - Exclude `ScanlineFilterCommon` from `getNameOfClass.py` to fix #96.

I initially wanted to rewrite `ScanlineFilterCommon` to have it inherit `ImageToImageFilter` but this introduces a diamond inheritance with `InPlaceImageFilter`, which I imagine is why is used composition in the first place. Honestly, I don't know why it needs to be wrapped.


Wrapping/Generators/Python/Tests/CMakeLists.txt
```diff
@@ -49,70 +37,13 @@ if(ITK_WRAP_unsigned_char AND WRAP_2)
...
-#   itk_python_add_test(NAME PythonStrelFromImageGrayscaleDilateImageFilter
-#     StrelFromImageGrayscaleDilateImageFilter.py
-#     DATA{${WrapITK_SOURCE_DIR}/images/cthead1.png}
-#     StrelFromImageGrayscaleDilateImageFilter.png
-#     DATA{${WrapITK_SOURCE_DIR}/images/StrelFromImage.png}
-#     --compare StrelFromImageGrayscaleDilateImageFilter.png DATA{${WrapITK_SOURCE_DIR}/images/StrelFromImageGrayscaleDilateImageFilter.png}
-#   )
-
-#   itk_python_add_test(NAME PythonFlatStructuringElementBall
-#     FlatStructuringElement.py
```
fbudin69500: Instead of simply removing this test, could you remove the part that does not work anymore (`GetImageUC`) and move the test to where it should be (`Modules/Filtering/MathematicalMorphology/test/`)?

fbudin69500:
Awesome! Thanks! :+1:

PR #646, opened 2019-03-28
BUG: Wrap itkImageDuplicator for all image types

fbudin69500:
NumPyBridge uses the itkImageDuplicator filter to copy images, so all image
types need to be supported (wrapped) for this filter to be able to convert
all ITK images into NumPy arrays.


fbudin69500:
#647 merged in master.

PR #648, opened 2019-03-28
ENH: itkPCAShapeSignedDistanceFunction supports float images

fbudin69500:
Exposing 3rd template argument of itkPCAShapeSignedDistanceFunction
which defaults to double images in C++. This patch allows to use
another image type such as float images.

### DIFF
--- a/Modules/Segmentation/SignedDistanceFunction/wrapping/itkPCAShapeSignedDistanceFunction.wrap
+++ b/Modules/Segmentation/SignedDistanceFunction/wrapping/itkPCAShapeSignedDistanceFunction.wrap
@@ -1,7 +1,14 @@
 itk_wrap_class("itk::PCAShapeSignedDistanceFunction" POINTER)
   foreach(d ${ITK_WRAP_IMAGE_DIMS})
-    # Can only be wrapped for `double` because of OptimizerParameters that
-    # is hardcoded as templated over `double`.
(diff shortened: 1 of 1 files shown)


Modules/Segmentation/SignedDistanceFunction/wrapping/itkPCAShapeSignedDistanceFunction.wrap
```diff
@@ -1,7 +1,12 @@
 itk_wrap_class("itk::PCAShapeSignedDistanceFunction" POINTER)
   foreach(d ${ITK_WRAP_IMAGE_DIMS})
+    # Force wrapping for double images as it is the default image template type in
+    # C++.
+    UNIQUE(types "${WRAP_ITK_REAL};D")
     # Can only be wrapped for `double` because of OptimizerParameters that
```
dzenanz: This comment should now be modified.

fbudin69500: I guess it could be interpreted ambiguously as it adds `double` on top of other real types. I'll try to think of a better phrasing.

fbudin69500: @dzenanz : Let me know if the new phrasing is more explicit.

PR #649, opened 2019-03-28
ENH: itkPCAShapeSignedDistanceFunction supports float images

fbudin69500:
Exposing 3rd template argument of itkPCAShapeSignedDistanceFunction
which defaults to double images in C++. This patch allows to use
another image type such as float images.


fbudin69500:
#648 merged in master

PR #653, opened 2019-03-29
ENH: ModifiedTimeType uses 64 bits on Windows - it was 32 before

dzenanz:


dzenanz:
This is missing `circle.png`+resample test failure.

dzenanz:
As #654 needs more work, I will merge this now instead of waiting to rebase.

PR #662, opened 2019-03-30
ENH: Improve test coverage of ImageMoments

hjmjohnson:
Add masking from SpatialObjects for
computing the ImageMoments.

### DIFF
--- a/Modules/Core/SpatialObjects/test/itkContourSpatialObjectTest.cxx
+++ b/Modules/Core/SpatialObjects/test/itkContourSpatialObjectTest.cxx
@@ -63,11 +63,6 @@ int itkContourSpatialObjectTest(int, char* [])
 
 
   //
-  // Test Update() before data added
-  //
-  contour->Update();
-
-  //
   // Test Control Points (SetControlPoints, GetControlPoints,
   // GetNumberOfControlPoints, GetControlPoint)
   //
@@ -78,6 +73,11 @@ int itkContourSpatialObjectTest(int, char* [])
   controlPointList.push_back(pt4);
 
   contour->SetControlPoints(controlPointList);
+
+  //
+  // Test Update() before data added
+  //
+  contour->Update();
 
   // check number of points
   if (contour->GetNumberOfControlPoints() != 4)
--- a/Modules/Core/SpatialObjects/test/itkContourSpatialObjectTest.cxx
+++ b/Modules/Core/SpatialObjects/test/itkContourSpatialObjectTest.cxx
@@ -63,7 +63,7 @@ int itkContourSpatialObjectTest(int, char* [])
 
 
   //
-  // Test ComputeMyBoundingBox before data added
+  // Test Update() before data added
   //
   contour->Update();
 
@@ -261,14 +261,18 @@ int itkContourSpatialObjectTest(int, char* [])
 
 
   //
-  // Test ComputeLocalBoundingBox
+  // Test Update()
   //
-  if (!contour->ComputeMyBoundingBox())
+  try
     {
-    std::cout << "[FAILED] faild bounding box computation" << std::endl;
+    contour->Update();
+    }
+  catch ( itk::ExceptionObject excp )
+    {
+    std::cout << "[FAILED] failed Update()" << std::endl;
     return EXIT_FAILURE;
     }
-  std::cout << "[PASSED] ComputeLocalBoundingBox" << std::endl;
+  std::cout << "[PASSED] Update()" << std::endl;
 
 
   //
--- a/Modules/Core/SpatialObjects/test/itkEllipseSpatialObjectTest.cxx
+++ b/Modules/Core/SpatialObjects/test/itkEllipseSpatialObjectTest.cxx
@@ -127,8 +127,9 @@ int itkEllipseSpatialObjectTest(int, char* [])
   }
   std::cout<<"[PASSED]"<<std::endl;
 
-  myEllipse->ComputeFamilyBoundingBox( EllipseType::MaximumDepth );
+  //NOTE: ORDER OF Update() and ComputeFamilyBoundingBox() is important.
   myEllipse->Update();
+  myEllipse->ComputeFamilyBoundingBox( EllipseType::MaximumDepth );
   const EllipseType::BoundingBoxType * boundingBox
     = myEllipse->GetFamilyBoundingBoxInWorldSpace();
   std::cout << "Bounds = " << boundingBox->GetBounds() << std::endl;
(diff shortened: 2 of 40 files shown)


hjmjohnson:
@dzenanz The error was the whole point of this patch set.  It demonstrates that this code fails where previously it had succeeded (see #663 for the same code passing).

I've been discussing with @aylward and refactored to provide a more clear path to working code. in subsequent patches.

hjmjohnson:
@aylward it would be great if you could look at these changes.
@N-Dekker This is related to the conversation you and @aylward were having on a different PR.  I went ahead and implement part of it because the current implementation was breaking my codebase.

Modules/Core/SpatialObjects/test/itkImageMaskSpatialObjectTest.cxx
```diff
@@ -72,7 +72,7 @@ int itkImageMaskSpatialObjectTest(int, char* [])
   maskSO->SetImage(image);
   maskSO->Update();
 
-  maskSO->ComputeObjectToWorldTransform();
+  maskSO->Update();
```
aylward: Perhaps redundant.

Modules/Core/SpatialObjects/test/itkImageMaskSpatialObjectTest2.cxx
```diff
@@ -112,7 +112,7 @@ int itkImageMaskSpatialObjectTest2(int, char* [])
   maskSO->SetImage(image);
   maskSO->Update();
 
-  maskSO->ComputeObjectToWorldTransform();
+    maskSO->Update();
```
aylward: Redundant

Modules/Core/SpatialObjects/test/itkImageMaskSpatialObjectTest2.cxx
```diff
@@ -255,7 +255,7 @@ int itkImageMaskSpatialObjectTest2(int, char* [])
...
```
aylward: Redundant

Modules/Filtering/ImageStatistics/test/itkImageMomentsTest.cxx
```diff
@@ -107,6 +115,32 @@ itkImageMomentsTest( int itkNotUsed(argc), char * itkNotUsed(argv) [] )
...
+      maskimg->SetRegions(image->GetLargestPossibleRegion());
+      maskimg->Allocate();
+      // Masking the entire image should not change the computation results.
+      maskimg->FillBuffer( itk::NumericTraits<PixelType>::OneValue() );
+
+      // convert mask image to mask
+      using LFFImageMaskSpatialObjectType = typename itk::ImageMaskSpatialObject<MaskImageType::ImageDimension>;
+      typename LFFImageMaskSpatialObjectType::Pointer mask = LFFImageMaskSpatialObjectType::New();
+      mask->SetImage(maskimg.GetPointer());
+      mask->ComputeObjectToWorldTransform();
```
aylward: Perhaps change this to Update()?

Modules/Core/SpatialObjects/test/itkImageMaskSpatialObjectTest.cxx
hjmjohnson: Removed.

Modules/Core/SpatialObjects/test/itkContourSpatialObjectTest.cxx
```diff
@@ -79,6 +74,12 @@ int itkContourSpatialObjectTest(int, char* [])
 
   contour->SetControlPoints(controlPointList);
 
+    //
```
dzenanz: comment misaligned

Modules/Core/SpatialObjects/test/itkEllipseSpatialObjectTest.cxx
```diff
@@ -127,17 +127,19 @@ int itkEllipseSpatialObjectTest(int, char* [])
...
   const EllipseType::BoundingBoxType * boundingBox
     = myEllipse->GetFamilyBoundingBoxInWorldSpace();
   std::cout << "Bounds = " << boundingBox->GetBounds() << std::endl;
 
   std::cout << "Update(): ";
   for(unsigned int i=0;i<3;i++)
   {
-    if(   itk::Math::NotAlmostEquals(boundingBox->GetBounds()[2*i], 7 )
-       || itk::Math::NotAlmostEquals(boundingBox->GetBounds()[2*i+1], 16 )
+    const EllipseType::BoundingBoxType::BoundsArrayType bnds = boundingBox->GetBounds();
```
dzenanz: not much was saved by abbreviating `bounds` to `bnds`

hjmjohnson: It allowed me to see the variable in the debugger when I was trying to figure out what was going wrong.

dzenanz: That is a reason I frequently introduce a variable too. But you could have named it `bounds` instead of `bnds`. You saved 2 characters, but made the name more cryptic.

Modules/Core/SpatialObjects/test/itkContourSpatialObjectTest.cxx
hjmjohnson: Fixed.

Modules/Core/SpatialObjects/test/itkEllipseSpatialObjectTest.cxx
hjmjohnson: Ok.  I fixed the variable name to be bounds.

Modules/Core/SpatialObjects/test/itkEllipseSpatialObjectTest.cxx
```diff
@@ -127,17 +127,19 @@ int itkEllipseSpatialObjectTest(int, char* [])
   }
   std::cout<<"[PASSED]"<<std::endl;
 
-  myEllipse->ComputeFamilyBoundingBox( EllipseType::MaximumDepth );
```
hjmjohnson: @aylward I'm a little concerned that the ordering of these calls changes the behavior.  For example, calling "ComputeFamilyBoundingBox" must occur after calling "Update".

Should "ComputeFamilyBoundingBox" implicitly call update?

aylward: As long as ComputeMyBoundingBox() has been called via Update(), there is no need for ComputeFamilyBoundingBox() to call Update().   We perhaps need to make it clear that function calls to get BoundingBoxes, PerimeterSize, Areas, and other values from the objects must occur after defining the parameters of the object and then calling Update().   I think this is a logical flow, and relatively easy to document.     My concern with calling Update() from ComputeFamilyBoundingBox() or any other function is that Update() could be a computationally expensive call.   We would need to make certain all computations within Update() are conditional on modified times, and they are not at this time.  One thing I did  not clear up in the refactoring is the odd handling of modified times by certain spatial objects - there are still some inconsistencies (and some pretzel logic, imo), but I think that the appropriate use of Update() mitigates those deficiencies and will not change behavior once they are resolved.

PR #666, opened 2019-04-01
ENH: Bump remote modules.

jhlegarreta:
Bump remote modules.

## PR Checklist
- [ ] Adds the License notice to new files.
- [ ] Adds Python wrapping.
- [ ] Adds Documentation.

### DIFF
--- a/Modules/Remote/AnisotropicDiffusionLBR.remote.cmake
+++ b/Modules/Remote/AnisotropicDiffusionLBR.remote.cmake
@@ -20,5 +20,5 @@ itk_fetch_module(AnisotropicDiffusionLBR
     http://insight-journal.org/browse/publication/953
   "
   GIT_REPOSITORY ${git_protocol}://github.com/InsightSoftwareConsortium/ITKAnisotropicDiffusionLBR.git
-  GIT_TAG 087c5fda829a1b08f4ce6d1a7cebdb9af23d1080
+  GIT_TAG ebd95bf5fc6f90f76735c4388b87d926ac38d9bb
   )
--- a/Modules/Remote/BioCell.remote.cmake
+++ b/Modules/Remote/BioCell.remote.cmake
@@ -4,5 +4,5 @@ It has classes to represent cells' shape, color, and growth state.
 It also has classes to represent a cell genome,
 whose expression is modeled by differential equations."
   GIT_REPOSITORY ${git_protocol}://github.com/InsightSoftwareConsortium/ITKBioCell.git
-  GIT_TAG 9fae5e04e558b4d0918384f0f8b91c22da5067f7
+  GIT_TAG 6dd1b454f64a17a60d88d58a84d895014714b181
   )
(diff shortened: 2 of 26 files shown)


jhlegarreta:
Done using the [UpdateRemoteModules.sh](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/Maintenance/UpdateRemoteModules.sh) script :smiley:

This should help cleaning the Expected Nightly Remote Modules dashboard
http://testing.cdash.org/index.php?project=Insight#!#Expected+Nightly+Remote+Modules

dzenanz:
Good job! I guess `WikiExamples` needs some love.

Modules/Remote/RTK.remote.cmake
```diff
@@ -2,5 +2,5 @@
 itk_fetch_module(RTK
     "Reconstruction Toolkit (RTK) http://www.openrtk.org/"
   GIT_REPOSITORY ${git_protocol}://github.com/SimonRit/RTK.git
-  GIT_TAG v2.0.0
```
phcerdan: Maybe do not update this one?
- Not owned by ISC or KitwareMedical.
- It is tagged, so maybe master is no stable?

phcerdan: Maybe your awesome bash script to automatic update remote modules could accept an option to only update repos owned by ISC?
Not all repositories would have master as a stable version.

dzenanz: It is easy to manually revert files changes for a few remotes. And adding options complicates the usage, in addition to requiring development and maintenance of the feature.

jhlegarreta: > Not owned by ISC or KitwareMedical.

It must be one of the few/the only one. I ignore why it has not been transferred. 

> It is tagged, so maybe master is not stable?

Unfortunately, most of remotes are not tagged.

As @dzenanz says, adding options brings flexibility but may yield things complicated (at least without careful design, and a set of rules that all remotes stick to I guess).

For this one I would manually revert it if necessary.

IMHO, what would be ideal would be to check for the latest tag if it exists and update accordingly, otherwise stick to the latest commit, as it now does. And in both cases **automatically checking the CI build state**, and updating only if it is passing. I ignore if that is doable. Even if it is, I will not be able to tackle it in the short term.

jhlegarreta: @phcerdan If you wish, you can open an issue to keep track of these ideas.

phcerdan: Your comments make sense. I just wanted to pointed out that some developers treat master as a `dev` branch. Even though that's not the case for the InsightSoftwareConsortium repositories.
> I ignore why it has not been transferred.

I happily transfer repositories to ISC to ease integration, but I don't think transferring is the norm, or an assumption that can be made. I would black list RTK. or at least make sure that the owner is happy to continuously integrate its master into ITK. Or for now, revert that RTK line, and add it later. Just my two cents, it looks good to me!

phcerdan: @SimonRit

jhlegarreta: Removed the RTK update in  60e8310 for the time being.

SimonRit: Ok too bad... I'm trying to address this error (which originates from an ITK backward incompatible change), see #674.

PR #668, opened 2019-04-01
ENH: Expand ResampleImageFilter requested region by interpolator domain

thewtex:
When streaming during resampling, the requested region is extended by
the domain of the interpolator. This is especially important for
consistent resampling results with interpolators other than the nearest
neighbor or linear interpolator, which use many surrounding pixels to
determine their value.

A new, abstract GetRadius method is added to InterpolateImageFunction.
This method is used by the ResampleImageFilter to extend its required
input region.

Suggested-by: Bradley Lowekamp <blowekamp@mail.nih.gov>

## PR Checklist

### DIFF
--- a/Modules/Core/ImageFunction/include/itkBSplineInterpolateImageFunction.h
+++ b/Modules/Core/ImageFunction/include/itkBSplineInterpolateImageFunction.h
@@ -111,6 +111,9 @@ public:
   /** Index type alias support */
   using IndexType = typename Superclass::IndexType;
 
+  /** Size type alias support */
+  using SizeType = typename Superclass::SizeType;
+
   /** ContinuousIndex type alias support */
   using ContinuousIndexType = typename Superclass::ContinuousIndexType;
 
@@ -311,6 +314,11 @@ public:
   itkSetMacro(UseImageDirection, bool);
   itkGetConstMacro(UseImageDirection, bool);
   itkBooleanMacro(UseImageDirection);
+
+  SizeType GetRadius() const override
+    {
+    return SizeType::Filled(m_SplineOrder + 1);
+    }
 
 protected:
 
--- a/Modules/Core/ImageFunction/include/itkGaussianInterpolateImageFunction.hxx
+++ b/Modules/Core/ImageFunction/include/itkGaussianInterpolateImageFunction.hxx
@@ -222,6 +222,30 @@ GaussianInterpolateImageFunction<TImageType, TCoordRep>
     }
 }
 
+
+template<typename TImageType, typename TCoordRep>
+typename GaussianInterpolateImageFunction<TImageType, TCoordRep>::SizeType
+GaussianInterpolateImageFunction<TImageType, TCoordRep>
+::GetRadius() const
+{
+  SizeType radius;
+
+  if( !this->GetInputImage() )
+    {
+    itkExceptionMacro("Input image required!");
+    }
+
+  const InputImageType * input = this->GetInputImage();
+  const typename InputImageType::SpacingType spacing = input->GetSpacing();
+
+  for( unsigned int dim = 0; dim < ImageDimension; ++dim )
+    {
+    radius[dim] = static_cast< SizeValueType >( Math::ceil( m_CutOffDistance[dim] / spacing[dim] ) );
+    }
+  return radius;
+}
+
+
 template<typename TImageType, typename TCoordRep>
 void
 GaussianInterpolateImageFunction<TImageType, TCoordRep>
--- a/Modules/Core/ImageFunction/include/itkInterpolateImageFunction.h
+++ b/Modules/Core/ImageFunction/include/itkInterpolateImageFunction.h
@@ -76,6 +76,9 @@ public:
   using IndexType = typename Superclass::IndexType;
   using IndexValueType = typename Superclass::IndexValueType;
 
+  /** Size type alias support */
+  using SizeType = typename InputImageType::SizeType;
+
   /** ContinuousIndex type alias support */
   using ContinuousIndexType = typename Superclass::ContinuousIndexType;
 
@@ -123,6 +126,13 @@ public:
   {
     return ( static_cast< RealType >( this->GetInputImage()->GetPixel(index) ) );
   }
+
+  /** Get the radius required for interpolation.
+   *
+   * This defines the number of surrounding pixels required to interpolate at
+   * a given point.
+   */
+  virtual SizeType GetRadius() const = 0;
 
 protected:
   InterpolateImageFunction()= default;
--- a/Modules/Core/ImageFunction/include/itkLinearInterpolateImageFunction.h
+++ b/Modules/Core/ImageFunction/include/itkLinearInterpolateImageFunction.h
@@ -82,6 +82,9 @@ public:
   /** Index type alias support */
   using IndexType = typename Superclass::IndexType;
 
+  /** Size type alias support */
+  using SizeType = typename Superclass::SizeType;
+
   /** ContinuousIndex type alias support */
   using ContinuousIndexType = typename Superclass::ContinuousIndexType;
   using InternalComputationType = typename ContinuousIndexType::ValueType;
@@ -100,6 +103,11 @@ public:
   {
     return this->EvaluateOptimized(Dispatch< ImageDimension >(), index);
   }
+
+  SizeType GetRadius() const override
+    {
+    return SizeType::Filled(1);
+    }
 
 protected:
   LinearInterpolateImageFunction() = default;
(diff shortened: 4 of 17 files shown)


romangrothausmann:
Many thanks @thewtex for tackling this problem. I had a look at the failing tests and noticed that the comparison with the baseline is disabled in current master:
https://github.com/InsightSoftwareConsortium/ITK/blob/651ee50286385b5913dc7c34a8cb67fa3fcab85c/Modules/Filtering/ImageGrid/test/CMakeLists.txt#L271-L280
It is not clear to me why and what was the last reason for that decision since they use linear interpolation
https://github.com/InsightSoftwareConsortium/ITK/blob/651ee50286385b5913dc7c34a8cb67fa3fcab85c/Modules/Filtering/ImageGrid/test/itkResampleImageTest2Streaming.cxx#L83
which already works with `itkResampleImageTest7`.
So I'd say that the failure of  `itkResampleImageTest7` is actually the critical one.

Modules/Filtering/ImageGrid/include/itkResampleImageFilter.hxx
```diff
@@ -540,6 +539,8 @@ ResampleImageFilter< TInputImage, TOutputImage, TInterpolatorPrecisionType, TTra
       // Input requested region is partially outside the largest possible region.
       //   or
       // Input requested region is completely inside the largest possible region.
+      inputRequestedRegion.PadByRadius( m_Interpolator->GetRadius() );
```
romangrothausmann: Concerning the linear interpolator I'd say that the former code would assume a radius of `0` for the padding but with 
https://github.com/InsightSoftwareConsortium/ITK/pull/668/commits/17d23c4f0a3d31c89f11d2bf402f477e7c0c07fd#diff-b143a8bc7c8f5f43d436547ffdb53fddR110
this now extents the requested region by `1`. 
So I assume the failure of `itkResampleImageTest7` originates from this padding but it is not clear to me why.

Modules/Core/ImageFunction/include/itkBSplineInterpolateImageFunction.h
```diff
@@ -312,6 +315,13 @@ class ITK_TEMPLATE_EXPORT BSplineInterpolateImageFunction:
   itkGetConstMacro(UseImageDirection, bool);
   itkBooleanMacro(UseImageDirection);
 
+  SizeType GetRadius() const override
+    {
+    SizeType radius;
+    radius.Fill( m_SplineOrder + 1 );
+    return radius;
```
N-Dekker: For what it's worth, this function could be written as `return SizeType::Filled(m_SplineOrder + 1);`

Modules/Core/ImageFunction/include/itkLinearInterpolateImageFunction.h
```diff
@@ -101,6 +104,13 @@ class ITK_TEMPLATE_EXPORT LinearInterpolateImageFunction:
     return this->EvaluateOptimized(Dispatch< ImageDimension >(), index);
   }
 
+  SizeType GetRadius() const override
+    {
+    SizeType radius;
+    radius.Fill( 1 );
+    return radius;
+    }
```
N-Dekker: FWIW, this function could be written as `return SizeType::Filled(1)`.

Modules/Core/ImageFunction/include/itkNearestNeighborInterpolateImageFunction.h
```diff
@@ -86,6 +89,13 @@ class ITK_TEMPLATE_EXPORT NearestNeighborInterpolateImageFunction:
     return static_cast< OutputType >( this->GetInputImage()->GetPixel(nindex) );
   }
 
+  SizeType GetRadius() const override
+    {
+    SizeType radius;
+    radius.Fill( 0 );
+    return radius;
+    }
```
N-Dekker: FWIW, could be written as `return SizeType()`.

dzenanz:
I applied Niels' suggestions and rebased. Doing this to refresh CI results. PR is not yet finished.

dzenanz:
Separating test enablement into https://github.com/InsightSoftwareConsortium/ITK/pull/872.

blowekamp:
I have worked on the interpolation in the `LinearThreadedGenerateData` and encountered many issues the portability of  round and numerical precissions. I'd recommend extending the padding by at least 0.5 in continuous index space before rounding or 1 afterwards.

Modules/Filtering/ImageGrid/include/itkResampleImageFilter.hxx
dzenanz: The cover region is now bigger by one, so the indices are wider by 1 and size is greater by 2.

Modules/Core/ImageFunction/include/itkRayCastInterpolateImageFunction.h
```diff
@@ -164,6 +165,13 @@ class ITK_TEMPLATE_EXPORT RayCastInterpolateImageFunction:
     return true;
   }
 
+  SizeType GetRadius() const override
+    {
+    SizeType radius;
+    radius.Fill( NumericTraits< SizeValueType >::max() );
```
blowekamp: This looks like it will likely lead to numeric overflow. Could we use the Image size instead?

Modules/Core/ImageFunction/include/itkInterpolateImageFunction.h
```diff
@@ -124,6 +127,13 @@ class ITK_TEMPLATE_EXPORT InterpolateImageFunction:
     return ( static_cast< RealType >( this->GetInputImage()->GetPixel(index) ) );
   }
 
+  /** Get the radius required for interpolation.
+   *
+   * This defines the number of surrounding pixels required to interpolate at
+   * a given point.
+   */
+  virtual SizeType GetRadius() const = 0;
```
blowekamp: This requires all derived interpolators to implement this method ( not too bad of a thing ). An alternative is to return the size of the image, so that the default usage would indicate the whole image is needed.

blowekamp:
Looks pretty good to me. Just  a couple questions about how best to indicate that an interpolator needs the whole image.

Modules/Core/ImageFunction/include/itkInterpolateImageFunction.h
dzenanz: Implementing this is not hard. Let's require all interpolators to do it. But we should add a note to migration guide.

dzenanz:
My next force-push applies Brad's suggestions. I am waiting for current CI results to come in before pushing this next amend.

Modules/Core/ImageFunction/include/itkGaussianInterpolateImageFunction.hxx
```diff
@@ -222,6 +222,30 @@ GaussianInterpolateImageFunction<TImageType, TCoordRep>
...
+template<typename TImageType, typename TCoordRep>
+typename GaussianInterpolateImageFunction<TImageType, TCoordRep>::SizeType
+GaussianInterpolateImageFunction<TImageType, TCoordRep>
+::GetRadius() const
+{
+  SizeType radius;
+
+  if( !this->GetInputImage() )
+    {
+    itkExceptionMacro("Input image required!");
```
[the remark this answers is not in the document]
dzenanz: > One of my new tests

Can you point us to the source code?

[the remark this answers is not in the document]
romangrothausmann: Sure, sorry. The test 
https://github.com/romangrothausmann/ITK-CLIs/blob/593b75ec483e07683a4c4bfc056b4701cc58713d/.gitlab-ci.yml#L154-L158
runs `resample.cxx` with parameters selecting `GaussianInterpolateImageFunction` (Interpolator_Type 10):
https://github.com/romangrothausmann/ITK-CLIs/blob/593b75ec483e07683a4c4bfc056b4701cc58713d/resample.cxx#L162-L173
once without streaming (https://github.com/romangrothausmann/ITK-CLIs/blob/593b75ec483e07683a4c4bfc056b4701cc58713d/.gitlab-ci.yml#L157 where it now fails)
and once with streaming 10 chunks (https://github.com/romangrothausmann/ITK-CLIs/blob/593b75ec483e07683a4c4bfc056b4701cc58713d/.gitlab-ci.yml#L158).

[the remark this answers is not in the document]
dzenanz: I figured out what the bug is. `m_Interpolator->SetInputImage( this->GetInput() );` is invoked to late. Patch coming up.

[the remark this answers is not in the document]
romangrothausmann: That's great! Many thanks @dzenanz for looking into this.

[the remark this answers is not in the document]
dzenanz: @romangrothausmann can you please test #958?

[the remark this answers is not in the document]
romangrothausmann: #958 solves this problem

PR #672, opened 2019-04-01
COMP: Support ITK_FUTURE_LEGACY_REMOVE for SpatialObjects

hjmjohnson:
When building with ITK_FUTURE_LEGACY_REMOVE option, the
itkLegacyMacro leads to incorrect syntax when the code
is not explicitly skipped.

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkSpatialObject.h
+++ b/Modules/Core/SpatialObjects/include/itkSpatialObject.h
@@ -494,11 +494,13 @@ public:
   bool Evaluate(const PointType & point) const
   { return this->IsInsideInWorldSpace(point); }
 
+#if ! defined ( ITK_LEGACY_REMOVE )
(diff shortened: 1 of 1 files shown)


Modules/Core/SpatialObjects/include/itkSpatialObject.h
```diff
@@ -494,11 +494,13 @@ class ITK_TEMPLATE_EXPORT SpatialObject:
   bool Evaluate(const PointType & point) const
   { return this->IsInsideInWorldSpace(point); }
 
+#if ! defined ( ITK_FUTURE_LEGACY_REMOVE )
```
dzenanz: Should this be `ITK_LEGACY_REMOVE` instead of `ITK_FUTURE_LEGACY_REMOVE`?

hjmjohnson: Thanks.  Fixed now.

PR #675, opened 2019-04-02
COMP: itkLegacyMacro should encompass inline definition (fixes ea784ea)

SimonRit:
This patch fixes RTK's compilation with -DITK_LEGACY_REMOVE=ON, see [RTK dashboard](https://my.cdash.org/viewBuildError.php?buildid=1629970) for the errors that showed up.
@hjmjohnson


hjmjohnson:
Fixed in #672

PR #679, opened 2019-04-03
WIP: BUG: index overflow in BinaryMask3DMeshSource

issakomi:
Reference [Email](https://public.kitware.com/pipermail/community/2017-August/013509.html)

### DIFF
--- a/Modules/Core/Mesh/include/itkBinaryMask3DMeshSource.h
+++ b/Modules/Core/Mesh/include/itkBinaryMask3DMeshSource.h
@@ -205,13 +205,13 @@ private:
   IdentifierType **m_CurrentRow{nullptr};
   IdentifierType **m_CurrentFrame{nullptr};
 
-  unsigned int m_CurrentRowIndex{0};
-  unsigned int m_CurrentFrameIndex{0};
-  unsigned int m_LastRowNum{0};
-  unsigned int m_LastFrameNum{0};
-  unsigned int m_CurrentRowNum{200};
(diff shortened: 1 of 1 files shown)


Modules/Core/Mesh/include/itkBinaryMask3DMeshSource.h
```diff
@@ -205,12 +205,12 @@ class ITK_TEMPLATE_EXPORT BinaryMask3DMeshSource:public ImageToMeshFilter< TInpu
   IdentifierType **m_CurrentRow{nullptr};
   IdentifierType **m_CurrentFrame{nullptr};
 
-  unsigned short m_CurrentRowIndex{0};
-  unsigned short m_CurrentFrameIndex{0};
-  unsigned short m_LastRowNum{0};
-  unsigned short m_LastFrameNum{0};
-  unsigned short m_CurrentRowNum{200};
-  unsigned short m_CurrentFrameNum{2000};
+  int m_CurrentRowIndex{0};
```
dzenanz: Why not use an unsigned type such as uint, ulong, size_t or `IdentifierType`?

issakomi: i guess for those (originally "unsigned short") variables the only correct type is "int", with "unsigned int" any test will crash (probably with "long" or "unsigned long" too). I don't remember exactly what i saw in debugger, can reproduce and post, some numeric-trait-whatever issue. But "unsigned short" is too small, test case in email can be reproduced (segfault).

PR #674, opened 2019-04-02
COMP: Add backwards compatible EnlargeRegionOverBox

thewtex:
This adds a backwards compatible signature following the addition of
another argument in

  609319760ae5c8c780c4b9be156a4a865ab79439

@romangrothausmann please take a look.

PR #680, opened 2019-04-03
COMP: Add backwards compatible EnlargeRegionOverBox

SimonRit:
This PR modifies PR #674 and now compiles in RTK, see [dashboard](https://my.cdash.org/buildSummary.php?buildid=1630838) . I don't think the solution is very elegant but I could not find a better solution. Other options:
- use a default template argument (e.g., itk::IdentityTransform) but no Transform is available in this module (Core/Common).
- add another templated function containing the TransformPoint line with a specialization doing nothing for a dummy type (e.g. nullptr_t).
Please have a look @thewtex @romangrothausmann. If you have another suggestion, don't hesitate! BTW, the problem does not seem to arise on linux, see [dashboard](https://my.cdash.org/buildSummary.php?buildid=1630585), I really don't know why.

### DIFF
--- a/Modules/Core/Common/include/itkImageAlgorithm.h
+++ b/Modules/Core/Common/include/itkImageAlgorithm.h
@@ -110,12 +110,18 @@ struct ImageAlgorithm
    * the physical space covered by the input
    * region of the input image
    */
+  template<typename InputImageType, typename OutputImageType>
+  static typename OutputImageType::RegionType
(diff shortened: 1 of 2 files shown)


romangrothausmann:
Look good to me, too. I have little experience with compiling outside linux. Does ITK have OS specific cpp differentiations?

thewtex:
@romangrothausmann generally, the different compilers catch different issues (or have different bugs :-P), and we make an effort to be as compatible with as many compilers as possible.

PR #683, opened 2019-04-04
ENH: Disable test drivers compilation if Python wrapping ON for Azure…

fbudin69500:
… pipelines

Disable test drivers compilation should have a significant positive impact
on build time. It also does not impact testing negatively as these tests
are run on non-Python builds.


fbudin69500:
See #684

PR #684, opened 2019-04-04
ENH: Add advanced CMake option `ITK_DO_NOT_BUILD_TESTDRIVERS`

fbudin69500:
This option allows to skip building the itkTestDrivers and therefore
can speed up compilation time. This is especially useful on CI
where the tests using these itkTestDrivers can be run for certain
builds that do not risk to time out (i.e. no wrapping).

### DIFF
--- a/CMake/ITKModuleTest.cmake
+++ b/CMake/ITKModuleTest.cmake
@@ -12,9 +12,10 @@
 #   ADDITIONAL_SRC (optional) - additional source files, which don't contain tests
 
 macro(CreateTestDriver KIT KIT_LIBS KitTests)
-  set(ADDITIONAL_SRC ${ARGN})
-  if(EMSCRIPTEN)
-    set(emscripten_before "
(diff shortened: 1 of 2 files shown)


fbudin69500:
See #683

fbudin69500:
The C++ tests are not deactivated because that would require a little bit more CMake code to be changed (i.e. recognize if the test is a Python test or a C++ test, which require an extra argument to be passed to `add_itk_test` from `add_itk_python_test` which is doable but more complicated). This works around the potential failing tests by running only the Python tests with `-R Python` (maybe using labels would be cleaner).

blowekamp:
I would recommend using CTest labels to label the language of the tests:
https://github.com/SimpleITK/SimpleITK/blob/master/CMake/sitkAddTest.cmake#L70

It would the following to have variables something like: `ITK_CXX_TESTING` and `ITK_PYTHON_TESTING` initialized to the global enable testing variable.

fbudin69500:
#683 was merged.

PR #687, opened 2019-04-04
ENH: Add advanced CMake option `ITK_DO_NOT_BUILD_TESTDRIVERS`

fbudin69500:
This option allows to skip building the itkTestDrivers and therefore
can speed up compilation time. This is especially useful on CI
where the tests using these itkTestDrivers can be run for certain
builds that do not risk to time out (i.e. no wrapping).


fbudin69500:
Note that locally this leads to a speed up of about 25% (see [here](https://github.com/fbudin69500/ITKBuildTimeAnalysis/blob/master/AnalyzeTimes.ipynb))

PR #690, opened 2019-04-05
STYLE: Remove globals, ProcessObject MakeNameFromIndex use std to_string

N-Dekker:
Removed `ITK_GLOBAL_INDEX_NAMES_NUMBER`, `ITK_GLOBAL_INDEX_NAMES_LENGTH`, and
`globalIndexNames` from the unnamed namespace of itkProcessObject.cxx.
Simplified `ProcessObject::MakeNameFromIndex` by calling `std::to_string`,
instead of using `globalIndexNames`.

Triggered by forum topic "Build ITK 5.0 in Windows", started by Sandro Queiros,
who reported compile errors from VS2015 (before installing VS updates), at
https://discourse.itk.org/t/build-itk-5-0-in-windows/1748 saying:

    itkProcessObject.cxx(54): error C2440: ‘initializing’: cannot convert from ‘const char [3]’ to ‘const char’.

### DIFF
--- a/Modules/Core/Common/src/itkProcessObject.cxx
+++ b/Modules/Core/Common/src/itkProcessObject.cxx
@@ -31,30 +31,10 @@
 #include <cstdio>
 #include <sstream>
 #include <algorithm>
+#include <string>
 
 namespace itk
 {
-
-
-namespace
-{ // local namespace for managing globals
-constexpr size_t ITK_GLOBAL_INDEX_NAMES_NUMBER  = 100;
-constexpr size_t ITK_GLOBAL_INDEX_NAMES_LENGTH = 4;
-constexpr char globalIndexNames[ITK_GLOBAL_INDEX_NAMES_NUMBER][ITK_GLOBAL_INDEX_NAMES_LENGTH] =
-{ "_0", "_1", "_2", "_3", "_4", "_5", "_6", "_7", "_8", "_9",
-  "_10", "_11", "_12", "_13", "_14", "_15", "_16", "_17", "_18", "_19",
-  "_20", "_21", "_22", "_23", "_24", "_25", "_26", "_27", "_28", "_29",
-  "_30", "_31", "_32", "_33", "_34", "_35", "_36", "_37", "_38", "_39",
-  "_40", "_41", "_42", "_43", "_44", "_45", "_46", "_47", "_48", "_49",
-  "_50", "_51", "_52", "_53", "_54", "_55", "_56", "_57", "_58", "_59",
-  "_60", "_61", "_62", "_63", "_64", "_65", "_66", "_67", "_68", "_69",
-  "_70", "_71", "_72", "_73", "_74", "_75", "_76", "_77", "_78", "_79",
-  "_80", "_81", "_82", "_83", "_84", "_85", "_86", "_87", "_88", "_89",
-  "_90", "_91", "_92", "_93", "_94", "_95", "_96", "_97", "_98", "_99"
-};
-
-}
-
 
 ProcessObject
 ::ProcessObject() :
@@ -1115,16 +1095,7 @@ ProcessObject::DataObjectIdentifierType
 ProcessObject
 ::MakeNameFromIndex(DataObjectPointerArraySizeType idx) const
 {
-  if ( idx < ITK_GLOBAL_INDEX_NAMES_NUMBER )
-    {
-    return ProcessObject::DataObjectIdentifierType( globalIndexNames[idx] );
-    }
-  else
-    {
-    char buf[2+21]; // a 64-bit integer is ~20 decimal places max
-    sprintf(buf, "_%u", static_cast<unsigned int>(idx));
-    return buf;
-    }
+  return '_' + std::to_string(idx);
 }
 
 


N-Dekker:
Almost six years ago, Bradley (@blowekamp) added `globalIndexNames` to ProcessObject: 8bcd6e985b46593ab55b8c33b239ed08eaad7318  At that time, I guess that was beneficial, as it would avoid expensive `sprintf` calls or `ostringstream` operations, within `ProcessObject::MakeNameFromIndex`.

Now that we are using C++11, I think it is no longer necessary to use `globalIndexNames`, as the C++11 `std::to_string` function should be fast enough. Right?

blowekamp:
Occasionally a GetInput is done on a per pixel basis. So I believe the is still a good trade off of a little memory for time.

I’m currious of the timings for the 3 different operations.

N-Dekker:
@blowekamp 
> Occasionally a GetInput is done on a per pixel basis. So I believe this is still a good trade off of a little memory for significant time.
> 
> I’m curious of the timings for the 3 different operations.

Would `ProcessObject::MakeNameFromIndex` be called for every pixel of an image? Assuming that an image often has more than 100 pixels, I guess it would mostly be called for idx > 100, right?

So then, we should focus on comparing the timing of `std::to_string(idx)` to:

    char buf[2+21];
    sprintf(buf, "_%u", static_cast<unsigned int>(idx));
    return buf;
   
Right?

N-Dekker:
With C++11, I think `std::to_string` is in most cases preferable to hard-written int-to-string conversions. I think we should trust the `std` library to do it right.

Of course, it could still be interesting to know the performance effect of this commit. But then, do you have a realistic use case that calls `MakeNameFromIndex(idx)` very frequently with `idx` < `100`?

blowekamp:
> Would `ProcessObject::MakeNameFromIndex` be called for every pixel of an image? Assuming that an image often has many more than 100 pixels, I guess it would mostly be called for idx > 100, right?
> 
> So then, we should focus on comparing the timing of `std::to_string(idx)` to:
> 
> ```
> char buf[2+21];
> sprintf(buf, "_%u", static_cast<unsigned int>(idx));
> return buf;
> ```
> Right?

Umm.... Nope. The methods are for itk::DataObject inputs to a filter, not pixels. So an ITK file commonly has 1 or two inputs. Occasionally, the there might be a hundred indexed input for filters like `JoinSeriesImageFilter`.

The data inputs to a filter/process object are index be integer and/or string with the preference being having both. This is actually an over complex system, but it is highly optimized and efficient! **Please** don't add this work for each look up!

Also please use "code coverage" as guidance for the significant of these changes. Unfortunately, our new coverage doesn't seem to upload the line by line results the way it use to (Issue to report!)

My suggestion to fix this compilation it to make all the string the same length by adding a null at the end.  ie. ` "_0\0", "_0",...`.

N-Dekker:
@blowekamp Do I understand correctly that you really like to keep the local `globalIndexNames` look-up table of 100 literal strings in there? I was hoping that C++11 `std::to_string` would allow us to clean up this code.

blowekamp:
I don't understand how it's not clean. Even an efficient implementation of "std::to_string" is going to involve expensive memory allocations and deallocations. This is avoided with this simple and efficient indexed look up.

What is the downside to the current approach? 🤷‍♂️

dzenanz:
> What is the downside to the current approach?

It is not elegant 😈

But we could reduce the size of that array from 100 to 10, as the common number of inputs into a filter is 1, 2 and 3. Adding nulls would then be easy:

> make all the string the same length by adding a null at the end. ie. "_0\0", "_0",....

Modules/Core/Common/src/itkProcessObject.cxx
```diff
@@ -1115,16 +1095,7 @@ ProcessObject::DataObjectIdentifierType
...
-    {
-    return ProcessObject::DataObjectIdentifierType( globalIndexNames[idx] );
-    }
-  else
-    {
-    char buf[2+21]; // a 64-bit integer is ~20 decimal places max
-    sprintf(buf, "_%u", static_cast<unsigned int>(idx));
-    return buf;
-    }
+  return '_' + std::to_string(idx);
```
blowekamp: How many dynamic memory allocations occur in this line? 2? 3?

blowekamp:
> But we could reduce the size of that array from 100 to 10, as the common number of inputs into a filter is 1, 2 and 3. Adding nulls would then be easy.

If all the elements of the array are 2 characters, I don't think an additional null would be needed.

Modules/Core/Common/src/itkProcessObject.cxx
N-Dekker: @blowekamp Good question! When `idx` is less than 100, I expect no dynamic allocation at all. Most common `std::string` implementations nowadays use small string optimization (SSO). 

I just had a closer look: With VS2017 x64, it appears that an `std::string` can hold up to 15 characters without any dynamic memory allocation! So with the proposed commit, `MakeNameFromIndex(idx)` would not do any dynamic memory allocation, as long as `idx` is less than `1.0e+14` (= 10^14). So the function would produce "_99999999999999", without dynamic allocation  `:-)`

N-Dekker: Update: I just checked that GCC and Clang also do small-string optimization for `std::string`, to avoid expensive dynamic memory allocation. So when `idx` has less than 15 decimal digits, I really don't think there will be any dynamic allocation, when doing `'_' + std::to_string(idx)`.

Note also that the current ITK git master version of `MakeNameFromIndex(idx)` has an integer overflow when `idx` has more than 10 decimal digits (specifically, when `idx > UINT_MAX`). The proposed commit will fix this bug, as it will support the full range of the type of `idx`.

dzenanz:
> Occasionally a GetInput is done on a per pixel basis.

Why would that be done? Can you provide an example of that @blowekamp?

And this example could then be bench-marked, with and without this patch.

blowekamp:
Please change only the else clause to use the proposed method.

This method can be timed alone.

blowekamp:
I'll just do it myself.

PR #695, opened 2019-04-05
Enh make test of non orthogonal image spatial objects

hjmjohnson:
While investigating other issues that appear to be related to SpatialObject update, I have fortified this test.  This is an attempt to isolate the issue with the new spatial objects (which this patch set does not do), but is still a valuable enhancement to the test.

### DIFF
--- a/Modules/Core/SpatialObjects/test/itkImageMaskSpatialObjectTest2.cxx
+++ b/Modules/Core/SpatialObjects/test/itkImageMaskSpatialObjectTest2.cxx
@@ -33,7 +33,7 @@
 #include "itkImageRegionIterator.h"
 #include "itkImageMaskSpatialObject.h"
 #include "itkMath.h"
-
+#include "itkEuler3DTransform.h"
 
 int itkImageMaskSpatialObjectTest2(int, char* [])
 {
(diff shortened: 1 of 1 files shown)


Modules/Core/SpatialObjects/test/itkImageMaskSpatialObjectTest2.cxx
```diff
@@ -94,45 +94,73 @@ int itkImageMaskSpatialObjectTest2(int, char* [])
...
+      ++it;
+      }
+  }
 
   ImageMaskSpatialObject::Pointer maskSO = ImageMaskSpatialObject::New();
-
   maskSO->SetImage(image);
   maskSO->Update();
 
+  { // Replicate use of MaskSpatialObject behavior from in itk::ImageToImageMetric.hxx
```
dzenanz: spelling: `from in` -> either `from` or `in`, not both

Modules/Core/SpatialObjects/test/itkImageMaskSpatialObjectTest2.cxx
```diff
@@ -94,45 +94,73 @@ int itkImageMaskSpatialObjectTest2(int, char* [])
...
+    const ImageType::IndexType constIndex = itr.GetIndex();
 
     ImageType::PointType point;
-    image->TransformIndexToPhysicalPoint( constIndex, point );
-    const bool test      = maskSO->IsInsideInWorldSpace( point );
-    if( test != reference )
+    image->TransformIndexToPhysicalPoint(constIndex, point);
+    const bool isInsideTest = maskSO->IsInsideInWorldSpace(point);
+
+      double outsideIfZeroValue;
```
dzenanz: this line and next two should have indentation level lower (indented by 2 spaces less)

thewtex:
Thanks @hjmjohnson !

@aylward @N-Dekker please take a look.

PR #696, opened 2019-04-05
BUG: Add Clear() to spatial objects

### LINKED ISSUE #691: Implement a Clear() function in SpatialObjects v5
In the refactoring of SpatialObjects to ITKv5, the Clear() function was removed since it wasn't consistently implemented.   However, there is a need for such a function to revert a SpatialObject to its initial state.   This function should remove all points in a PointBasedSpatialObject and sets its ObjectToParentTransform to identity, for example.

This Clear() function should not delete/remove children.  Removing children should be accomplished using the existing RemoveAllChildren() function.

aylward:
This PR resolves #691 by adding the Clear() function to ITK's SpatialObjects.   The Clear() function resets the spatial object to its initial state, except for the object's Id, Parent, and Children.


aylward:
Rebasing

PR #697, opened 2019-04-06
Move CI configuration into Azure YML files

fbudin69500:
CI configuration was done through environment variables that were then
interpreted in a CTest configuration script shipped in the dashboard branch
of ITK. This implementation required to modified both the master branch
and the dashboard branch if new options had to be controled from the YML
configuration files. For ITK remote modules, the configuration is already
done directly inside the YML configuration file. This patch moves the CI
configuration for ITK inside the master branch the same way it is done for
remote modules.


thewtex:
Cool -- the move away from environmental variables only is nice :+1: 

Can we put common code in the `dashboard` branch, then include it, `include(azure_common.cmake)`, to avoid duplication across the YAML files?

PR #698, opened 2019-04-06
Move CI configuration into Azure YML files

fbudin69500:
CI configuration was done through environment variables that were then
interpreted in a CTest configuration script shipped in the dashboard branch
of ITK. This implementation required to modified both the master branch
and the dashboard branch if new options had to be controled from the YML
configuration files. For ITK remote modules, the configuration is already
done directly inside the YML configuration file. This patch moves the CI
configuration for ITK inside the master branch the same way it is done for
remote modules.

### DIFF
--- a/Testing/ContinuousIntegration/AzurePipelinesBatch.yml
+++ b/Testing/ContinuousIntegration/AzurePipelinesBatch.yml
@@ -36,23 +36,29 @@ jobs:
         workingDirectory: $(Agent.BuildDirectory)
         displayName: 'Download dashboard script and testing data'
       - bash: |
-          ctest -V -S ITK-dashboard/azure_dashboard.cmake
-        displayName: Build and test with coverage
+          cat > dashboard.cmake << EOF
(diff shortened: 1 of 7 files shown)


fbudin69500:
This integrates with #703.

blowekamp:
I think this is a step in the right direction by making things more direct and straight forward. I have been pondering how to reduce the churn of commits for these CI build configuration files.

Perhaps the next step is to keep the CMake/CTest build scripts in a separate repository, and download them for the CI build? This way the configuration can be updated with out noise commit in the ITK repo?

fbudin69500:
Separating files can be good to avoid noise, but I would anticipate that most of the time we want to actually change the CMake configuration, it has to do with a specific commit in ITK. At least that is why I am working on this now, and having both the update of the CI configuration and my patch in the same commit seems easier to understand to track down possible issues due to these modifications in the future. Maybe we can keep that in mind and see if in the future we end up having to change the CMake configuration for the CI for a reason that is not directly linked to a patch?

PR #699, opened 2019-04-07
DOC: Update recommendation for computing a Bounding Region in IndexSpace

aylward:
Since MyBoundingBoxInObjectSpace is computed via Update(), it can be converted to a bounding itk::Region defined in IndexSpace.

Updated example code is provided.  It offers significant speed improvement by using the already-computed MyBoundingBoxInObjectSpace as a basis for the index space box.


thewtex:
Nice!

Documentation/ITK5MigrationGuide.md
```diff
@@ -352,21 +352,50 @@ As implied above, the changes to SpatialObject are extensive.   They include the
...
+  indexBoundingBox->SetPoints(transformedCorners);
+  indexBoundingBox->ComputeBoundingBox();
+
+  RegionType boundingRegion;
+  RegionType::IndexType indx;
+  RegionType::SizeType size;
+  for( unsigned int i=0; i<Dimensions; ++i )
+      {
+      indx[i] = indexBoundingBox->GetMinimum()[i];
+      size[i] = indexBoundingBox->GetMaximum()[i] - indexBoundingBox->GetMinimum()[i] + 1;
```
N-Dekker: @aylward Please note: It appears that this will do `size[i] = 1` on a black image (all zero-valued pixels). For such an image, the legacy `GetAxisAlignedBoundingBoxRegion()` yields a size of zero values instead.

PR #703, opened 2019-04-08
Move configuration of _dashboard_cache out of azure_dashboard.cmake

fbudin69500:
`_dashboard_cache` which contains the initialization of ITK configuration
should now be done directly in the file calling azure_dashboard.cmake.
In the current setting, this means that the YML configuration files of
the CI should create a CMake file that initializes this variable and
then include `azure_dashboard.cmake`.


fbudin69500:
This integrates with #698.

PR #688, opened 2019-04-05
WIP: Speed up ci compilation with wrapping on follow up

fbudin69500:

PR #706, opened 2019-04-09
Speed up ci compilation with wrapping on follow up

fbudin69500:


fbudin69500:
Close #688 after merging this PR.

Issue #702, opened 2019-04-08
Code Coverage is not being reported by files in CDash


blowekamp:
For the new Azure code coverage builds, only the total number of lines and percentage are begin reported on CDash. On CDash when clicking coverage percentage to drill down to the per file level coverage, there are no files lists.

This may be a CDash issue as this info is not available for CMake either: https://open.cdash.org/viewCoverage.php?buildid=5845858

thewtex:
Per-line coverage is working in VTK:

https://open.cdash.org/viewCoverage.php?buildid=5849102

thewtex:
CC: @zackgalbreath

zackgalbreath:
A few weeks ago we upgraded the database server that hosts CDash. One of the default settings changed and this broke per-file coverage. This situation was fixed back on April 8th.

Let me know if there's anything I can do to help help get ITK coverage data displaying correctly on CDash.

blowekamp:
That follows with what I saw. Thanks!

PR #716, opened 2019-04-10
BUG: Set dashboard_do_coverage in Azure configuration

thewtex:
Also use CTEST_COVERAGE_COMMAND.

Hopefully this will address #702 .


dzenanz:
Where does the coverage get reported then?

PR #717, opened 2019-04-10
STYLE: Removed five unused nested types from ImageSpatialObject

N-Dekker:
Removed unused nested type aliases from `itk::ImageSpatialObject`:
 * RegionType
 * TransformType
 * BoundingBoxType
 * PointContainerType
 * PointContainerPointer


N-Dekker:
FYI, the first three nested types remain available via its base class. Only the last two are really removed, with this pull request:

    itk::ImageSpatialObject<>::RegionType* p1; // OK
    itk::ImageSpatialObject<>::TransformType* p2; // OK
    itk::ImageSpatialObject<>::BoundingBoxType* p3; // OK
    itk::ImageSpatialObject<>::PointContainerType* p4; // Error, after this pull request
    itk::ImageSpatialObject<>::PointContainerPointer* p5; // Error, after this pull request

PR #719, opened 2019-04-10
Add force snake case igenerator

fbudin69500:


fbudin69500:
@thewtex : The new test is passing [see here](https://open.cdash.org/viewTest.php?onlypassed&buildid=5851478&filtercount=1&showfilters=1&field1=testname&compare1=63&value1=itkImageDuplicatorPythonTest). I think it is good to be merged if you agree with this strategy.

PR #731, opened 2019-04-14
COMP: VtkGlue module-Provide support for VTK new cmake targets

schmidje:
Since VTK 8.2, the CMake building system is now using more intensively the CMake
targets system as proposed in "modern cmake". As a result, some CMake macros
used in the old VTK have been removed and the vtkGlue module of ITK does not
compile anymore.

The provided fix should support both new and previous VTK versions.

## PR Checklist
- [ ] Adds the License notice to new files.
- [ ] Adds Python wrapping.
- [ ] Adds Documentation.


schmidje:
the macro is repeated in the sections corresponding to `ITKVtkGlue_EXPORT_CODE_BUILD` and `ITKVtkGlue_EXPORT_CODE_INSTALL` that are respectively used at the generation stage to produce files necessary in the future to link against ITK in the build or install directories. For instance `ITKVtkGlue_EXPORT_CODE_BUILD` will produce the file `<build_dir>\ITK\Modules\Bridge\VtkGlue\CMakeFiles\ITKVtkGlue.cmake`. 

If you look closely the macro code is written within a section enclosed by "".

schmidje:
Closing the PR, thanks for the review. Hope the fix will work for others.

BR

mathstuf:
I'll note that this approach is basically trying to keep the old API up-to-date rather than properly migrating to the new one. Since ITK aims to support both, maybe this is OK. I'll note that one thing missing here is `vtk_module_autoinit` calls which means that object factories are not guaranteed to be properly initialized.

schmidje:
I agree with you, ITK should eventually properly migrate to the new CMake modules architecture, modern cmake offers so much advantages. However users should be better educated at linking with only the modules they really need, many of us, included myself, were too used to include the ITK_USE_FILE and link with all the ITK libraries.

Regarding the `vtk_module_autoinit` I am not sure the `ITKVtkGlue CMakeLists.txt` is the right place to put it as `autoinit `is used from my understanding at **runtime**. So I use it e.g. in the CMakeLists.txt of an exec using VTK and ITK.

dzenanz:
Rendering part of `ITKVtkGlue` still does not work for me. [This](https://open.cdash.org/testDetails.php?test=670818933&build=5860842) is a typical failure.

mathstuf:
That failure means that autoinit macros were not generated, so `vtkRenderer` never had an override register itself. It's needed at every place that VTK libraries are linked. Brad and I discussed how to do it with usage requirements and generator expressions, but there's no way with CMake today and the logic CMake would need to provide is non-trivial.

dzenanz:
I created an [issue](https://github.com/InsightSoftwareConsortium/ITK/issues/739) to track this.

schmidje:
Indeed we the new VTK I add to something like:

```
find_package(VTK COMPONENTS 
  ...)

# with vtk > 8.2 we must use the following
# otherwise it will fail at *runtime* for the rendering
vtk_module_autoinit(
  TARGETS my-target
MODULES ${VTK_LIBRARIES})
```

PR #732, opened 2019-04-14
BUG: Fix #734 and extend GTest for GetAxisAlignedBoundingBoxRegion

### LINKED ISSUE #734: Incorrect results GetAxisAlignedBoundingBoxRegion() for 2D
### Description

`ImageMaskSpatialObject::GetAxisAlignedBoundingBoxRegion()` appears to produce incorrect results for image dimensions other than 3D, in two cases:

1. When the image is entirely black (all zero's). In this case, a bounding box region size of zero is expected, but instead, the returned region appears to have a very large size! For example, { 18446744073709551613, 18446744073709551611 }
2. When the (N-dimensional) index of the requested region of the image has non-zero coordinates. In this case, the size of the returned bounding box region is also incorrect, apparently because the function just does not take the image region index into account. 

### Steps to Reproduce

In order to reproduce this issue, run the test from https://github.com/InsightSoftwareConsortium/ITK/blob/d19a05a841d4b89c7646d2ffef36f336017af54a/Modules/Core/SpatialObjects/test/itkImageMaskSpatialObjectGTest.cxx

Test failures are reported at https://open.cdash.org/viewTest.php?onlyfailed&buildid=5858683

### Versions

This issue was introduced in 2008, so it applies to both ITK 4.13.2 and ITK 5.0 RC1.

It appears that both of these bugs were there already in the first version that included support for image dimensions other than 3D:

  "ENH: implemented GetAxisAlignedBoundingBoxRegion() for image dimensions other than 3D"
  cd0a80cecf0dd9ea4306d6323d2fbf0413c704c5 (Marius Staring, May 22, 2008)

### Additional Information

I already discussed with Marius, and he agrees that it should be fixed. I'm now preparing a fix.

N-Dekker:
* Extended GTest for `ImageMaskSpatialObject::GetAxisAlignedBoundingBoxRegion()`
* Fixed #734 ("Incorrect results GetAxisAlignedBoundingBoxRegion() for 2D")

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkImageMaskSpatialObject.hxx
+++ b/Modules/Core/SpatialObjects/include/itkImageMaskSpatialObject.hxx
@@ -191,6 +191,7 @@ ImageMaskSpatialObject< TDimension, TPixel >
   RegionType region;
 
   ImagePointer image = this->GetImage();
+  const RegionType imageRegion = image->GetRequestedRegion();
 
   IndexType index;
   SizeType  size;
@@ -219,7 +220,7 @@ ImageMaskSpatialObject< TDimension, TPixel >
         }
 
       // Create the forward iterator to find lower bound
-      SliceIteratorType fit( image,  image->GetRequestedRegion() );
+      SliceIteratorType fit( image,  imageRegion );
       fit.SetFirstDirection(direction[1]);
       fit.SetSecondDirection(direction[0]);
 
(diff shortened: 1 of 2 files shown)


N-Dekker:
Thanks for your approval, @aylward and @dzenanz !

For your information, my last force-push only just reordered to two commits, to avoid a possible test failure on the master branch. (The GTest enhancement is now the last commit, instead of the first.) I think this pull request is ready now.

Issue #713, opened 2019-04-10
7 tests failing on MacOS (Azure Pipeline) when build type is `MinSizeRel`


fbudin69500:
### Description

7 tests are failing (Segfault or just failing) when compiling ITK on MacOS, at least on Azure Pipeline CI, when build type is `MinSizeRel`. These failure do not appear when compiling ITK in `Release`.

itkAnisotropicFourthOrderLevelSetImageFilterTest : SEGFAULT
itkGaussianInterpolateImageFunctionTest : Failed
itkImplicitManifoldNormalVectorFilterTest : SEGFAULT
itkIsotropicFourthOrderLevelSetImageFilterTest :SEGFAULT
itkLinearInterpolateImageFunctionTest : Failed
itkSparseFieldFourthOrderLevelSetImageFilterTest : SEGFAULT
itkUnsharpMaskLevelSetImageFilterTest : SEGFAULT

https://open.cdash.org/viewTest.php?onlyfailed&buildid=5848875

### Steps to Reproduce

Compile ITK on MacOS with build type set to MinSizeRel. This can be done on CI by creating a PR in which [this](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Testing/ContinuousIntegration/AzurePipelinesMacOS.yml#L44) is modified.

### Reproducibility

100% of the time, these tests fail.

### Versions

Current master:  7dbfeb6

### Environment

MacOS, Azure CI environment.

thewtex:
@maekclena did your PR's address all of these tests?

fbudin69500:
This could be tested on CI by changing build type in the yaml configuration file :)

fbudin69500:
https://github.com/InsightSoftwareConsortium/ITK/pull/758

thewtex:
> #758

Nice, thanks!

fbudin69500:
All these tests now pass on MacOS when build type is `MinSizeRel` (See [here](https://open.cdash.org/viewTest.php?onlypassed&buildid=5875215&filtercount=1&showfilters=1&field1=testname&compare1=63&value1=))

PR #733, opened 2019-04-15
BUG: Remove outdated class declaration

maekclena:
This is a fix for the segfaults in #713.

This test was redeclaring `NormalBandNode` but the class had since been updated. It seems that when using MinSizeRel, AppleClang was favoring this outdated declaration, causing segmentation faults when confusing attributes.

Concerning the two tests failing without segfaults in #713, I was never able to reproduce the problem.


thewtex:
Thank you @maekclena !

PR #736, opened 2019-04-15
WIP: itkObjectFactoryBasePrivateDestructor was not compiling with MSVC

fbudin69500:
Export and Import symbols was not generated and added to the libraries
that are created for the test `itkObjectFactoryBasePrivateDestructor`.
Executable for test could not be compiled correctly on MSVC.


fbudin69500:
#735 merged in master

Issue #685, opened 2019-04-04
Avoid updating remotes that are failing to build


jhlegarreta:
### Description
[UpdateRemoteModules.sh](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/Maintenance/UpdateRemoteModules.sh) updates all remotes without checking whether their actual builds are succeeding.

It should check for a green CI, and only update those remotes whose CI build succeeds. It could show a report in the std output, including those it updated, and those whose CI build is failing.

### Steps to Reproduce

Execute [UpdateRemoteModules.sh](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/Maintenance/UpdateRemoteModules.sh).

### Expected behavior

Update remote modules whose CI build is succeeding.

### Actual behavior

Updates all remote modules regardless of their CI build state.

### Reproducibility

100%

### Versions

`master`. Added in 8fdb4542f2203fddee463a02cbc14587ade2375c.

### Environment

Any.

### Additional Information

None.

jhlegarreta:
It was actually solved in PR #740.

PR #740, opened 2019-04-16
ENH: Update remote module maintenance script

### LINKED ISSUE #682: `UpdateRemoteModules.sh` does not handle correctly commit hashes with trailing characters
### Description
[UpdateRemoteModules.sh](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/Maintenance/UpdateRemoteModules.sh) does not handle correctly commit hashes with trailing characters.

### Steps to Reproduce

Execute [UpdateRemoteModules.sh](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/Maintenance/UpdateRemoteModules.sh).

### Expected behavior

Pretending the `+`/`-` signs are shown by `git diff`:
```
- set(GIT_TAG 70a5a174a0ca0bebe00d6b1f817caacceef8cf35)
+ set(GIT_TAG 07f0e1622ca3eeed8b9c5b9665b066b785af86e8)
```

### Actual behavior

Current result on applying the script to [WikiExamples.remote.cmake](Modules/Remote/WikiExamples.remote.cmake) pretending the `+`/`-` signs are shown by `git diff`
```
- set(GIT_TAG 70a5a174a0ca0bebe00d6b1f817caacceef8cf35)
+ set(GIT_TAG 07f0e1622ca3eeed8b9c5b9665b066b785af86e8
```

### Reproducibility

100%

### Versions

`master`. Added in 8fdb4542f2203fddee463a02cbc14587ade2375c.

### Environment

Any.

### Additional Information

None.

maekclena:
Fixes #682 and #685.

This updated version of `UpdateRemoteModules.sh` uses more constraining parsing rules to avoid issues and removes version specific syntax to be more portable.

It also adds the option `-ongreen` to update only if the latest build was successful.

### DIFF
--- a/Utilities/Maintenance/UpdateRemoteModules.sh
+++ b/Utilities/Maintenance/UpdateRemoteModules.sh
@@ -24,17 +24,25 @@ cd "${BASH_SOURCE%/*}" &&
 
 remote_modules_path=$(cd ../../Modules/Remote && pwd)
 
+updated=()
+if [ $# -ge 1 ] && [ $1 = "-ongreen" ]; then
+  ongreen=1
+  failing=()
+else
+  ongreen=0
+fi
+
 # Loop over the remote modules' CMake files
 for filename in ${remote_modules_path}/*.cmake; do
+  echo -n "."
 
   # Get the current commit hash
-  curr_commit=$(grep -m 1 -o "GIT_TAG \(\w\|\.\|\-\|_\)*" $filename)
-  curr_commit=${curr_commit/GIT_TAG /}
+  curr_commit=$(grep -m 1 -o "[^#]GIT_TAG \(\w\|\.\|\-\|_\)*" $filename)
+  curr_commit=${curr_commit/*GIT_TAG /}
 
   # Read the git repository information
-  repository=$(grep -m 1 -o "^\s*GIT_REPOSITORY \${git_protocol}://github.com/\(\w\|\.\|\-\|_\|/\)*" $filename)
+  repository=$(grep -m 1 -o "[^#]GIT_REPOSITORY \${git_protocol}://github.com/\(\w\|\-\|_\|/\)*" $filename)
   repository=${repository/*GIT_REPOSITORY \${git_protocol\}:\/\/github.com\//}
-  repository=${repository/.git/}
 
   # Get the latest git commit hash of the remote module.
   # Remotes will usually not be tagged.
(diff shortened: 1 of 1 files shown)


jhlegarreta:
May be the PR subject could be reformatted to *ENH: Update remote module maintenance script* so that it becomes clearer.

dzenanz:
@maekclena Can you run the script and make a new PR with updated remotes?

jhlegarreta:
I've just tested the PR and it works overall:
- **Nice output of the taken actions**. Could be improved by **grouping the updated vs. non-updated ones vs. those not update because they are failing** (thus avoiding the *(...) is already up to date* and *(...) was updated to latest commit*).  But this is not a show stopper. 
- We still have the tag issue with the remotes (i.e. `RTK` would be updated to the master commit), but I did not address that either and is closely related to the way remotes can be handled concerning versions (i.e. not an easy thing to be systematical), so for the time being, this is correct to me.
- Checked that the PR **does fix** #682. **Great**.
- **The ancient version was introducing some undesired CR/LFs** (my bad) to all files that were **later filtered by git** (ITK hooks may be) so as to avoid unnecessary commits. **The new version does not introduce them**. Which is **simply great**. Thanks @maekclena.

**Unfortunately I checked that**:
- The script updated the [ITKWikiExamples](https://github.com/InsightSoftwareConsortium/ITKWikiExamples) remote, which shows a **failing Azure Pipeline badge**, and I trust it is failing. The script updated the remote anyway, though, so the CI check in the script seems not to be working correctly.

@maekclena whenever you find the time, could you have a look?

Note that all remote CIs will eventually be migrated to Azure Pipelines, if this is of any interest for having the `status` answer.

jhlegarreta:
I've re-tested the script:
- The **output message now shows only the updated modules**. **Nice**.
- Looks like the **status is now taken into account**. **Great**.

Thanks @maekclena  !

However, there is still at least **one more thing that I've detected**: for remotes with upstream URLs in comment (e.g. [SmoothingRecursiveYvvGaussianFilter.remote.cmake](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Remote/SmoothingRecursiveYvvGaussianFilter.remote.cmake)), it looks like **the script takes the URL that first matches the regex, even if it lives on a commented line**. Thus, in the above case, instead of updating the commit hash from the [expected repository](https://github.com/InsightSoftwareConsortium/ITKSmoothingRecursiveYvvGaussianFilter), it will do so from the [upstream repository](https://github.com/Inria-Asclepios/SmoothingRecursiveYvvGaussianFilter), and it will be printed as such in the output:
```
Modules updated:
Inria-Asclepios/SmoothingRecursiveYvvGaussianFilter
```

The commit to which the remote module `cmake` file has been updated correspond's to the upstream latest, which, in this case, is well behind ISC's latest. Even if ISC's was behind (which may happen if the upstream has a merge commit from ISC), we'd like to stick to the appropriate repository (which may not necessarily be the upstream).

So I'd dare to say that the fix boils down to discarding the regex results that lie on a comment line.

dzenanz:
> So I'd dare to say that the fix boils down to discarding the regex results that lie on a comment line.

Equally valid fix is to remove the commented lines, or put them later so they don't throw of this script.

jhlegarreta:
@maekclena  Thanks ! Checked and now it is working as expected !!

:rocket: :tada: :fireworks: 

This is an awesome improvement !

jhlegarreta:
I will merge when Windows and CDash builds are completed successfully.

PR #741, opened 2019-04-16
PERF: ImageMaskSpatialObject::GetAxisAlignedBoundingBoxRegion() faster

N-Dekker:
Made `ImageMaskSpatialObject::GetAxisAlignedBoundingBoxRegion()` many times (!)
faster than before.

The amount of the speed improvement depends on the input image: For a black 3D
image of 256x256x256, it was observed than the function runs more than 10 x
faster than before. For a 2D image of 1024x1024 with non-zero border pixels,
it was observed than the function runs even more than 100 x faster!

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkImageMaskSpatialObject.hxx
+++ b/Modules/Core/SpatialObjects/include/itkImageMaskSpatialObject.hxx
@@ -20,6 +20,7 @@
 
 #include "itkMath.h"
 #include "itkImageMaskSpatialObject.h"
+#include "itkImageRegionConstIterator.h"
 #include "itkImageRegionConstIteratorWithIndex.h"
 
 namespace itk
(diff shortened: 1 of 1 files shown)


N-Dekker:
The implementation of this proposed revision is based on `LKEB::EstimateAxisAlignedBoundingBoxRegion(const TImage&)` from https://gist.github.com/N-Dekker/9e1f5bcc6f65c7652c11d1cbd3d519ee which was already read through thoroughly by @aylward

N-Dekker:
Note that the proposed implementation has only 64 lines of code, while the original had 116 lines `:-)` That's because the original implementation had separate code for 3D and for other dimensions. The proposed implementation works for each number of dimensions the same way.

The original implementation is especially slow for dimensions _other_ than 3D, because it does exhaustively check _all_ pixels of the image, to find the bounding box. The proposed implementation iterates from the outside to the inside of the requested image region, and it stops directly, as soon as it has found the bounding box.

PR #745, opened 2019-04-17
BUG: fix gaussian interpolate test

maekclena:
This is a fix for the 2 failing tests without segfault in #713.

It turns out that `itkGaussianInterpolateImageFunctionTest` was always broken, but when it called its `RunTest` function, the one in `itkLinearInterpolateImageFunctionTest` was ran instead. AppleClang with MinSizeRel just happened to confuse them the other way around.

I replaced the test with a basic regression check.

### DIFF
--- a/Modules/Core/ImageFunction/test/itkGaussianInterpolateImageFunctionTest.cxx
+++ b/Modules/Core/ImageFunction/test/itkGaussianInterpolateImageFunctionTest.cxx
@@ -19,258 +19,93 @@
 #include <iostream>
 
 #include "itkImage.h"
-#include "itkVectorImage.h"
 #include "itkGaussianInterpolateImageFunction.h"
 #include "itkMath.h"
 #include "itkTestingMacros.h"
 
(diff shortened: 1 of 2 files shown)


Modules/Core/ImageFunction/test/itkGaussianInterpolateImageFunctionTest.cxx
```diff
@@ -19,258 +19,93 @@
...
-      }
+      image->SetPixel({{i, j}}, i + j);
     }
+  }
 
-  typename InterpolatorType::Pointer interpolator = InterpolatorType::New();
+  interpolator->SetInputImage(image);
 
-  interpolator->SetInputImage( image );
+  InterpolatorType::OutputType expectedValues[5][5] = {{0.773964, 0.886982, 1.38698, 1.88698, 2.0},
```
dzenanz: Perhaps put everything after `=` on a new line

thewtex:
Outstanding @maekclena !

PR #749, opened 2019-04-19
DOC: Improve diagnostic message from ExtractImageFilter

hjmjohnson:
The original "OutputImageDimension-InputImageDimension" computation
resulted in underflow error and a very confusing message when the output
dimension was less than the input dimension.  Provide an expanded
diagnostic message to improve the understanding of what the failure
was."

### DIFF
--- a/Modules/Core/Common/include/itkExtractImageFilter.hxx
+++ b/Modules/Core/Common/include/itkExtractImageFilter.hxx
@@ -68,6 +68,7 @@ void
 ExtractImageFilter< TInputImage, TOutputImage >
 ::SetExtractionRegion(InputImageRegionType extractRegion)
 {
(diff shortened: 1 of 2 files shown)


Modules/Core/Common/include/itkExtractImageFilter.hxx
```diff
@@ -68,6 +68,7 @@ void
 ExtractImageFilter< TInputImage, TOutputImage >
 ::SetExtractionRegion(InputImageRegionType extractRegion)
 {
+  static_assert(InputImageDimension >= OutputImageDimension, "InputImageDimension must be greater than OutputImageDimension");
```
thewtex: Cool! :sunglasses:

PR #751, opened 2019-04-21
BUG: Fix issue #785 removing BoundingBox check from ImageMask IsInside

N-Dekker:
Fixed issue #785 (ImageMaskSpatialObject IsInside should not depend on
distant pixels) by removing the BoundingBox check from
`ImageMask::IsInsideInObjectSpace`

Added various GTest unit tests for `ImageMaskSpatialObject::IsInside`


N-Dekker:
Discussion topic: [ImageMaskSpatialObject IsInside less than half a pixel away?](https://discourse.itk.org/t/imagemaskspatialobject-isinside-less-than-half-a-pixel-away/1807)

PR #757, opened 2019-04-24
ENH: Add begin() and end() to FixedArray, support range-based for-loops

N-Dekker:
Added `begin()`, `end()`, `cbegin()`, `cend()`, `iterator`, and
`const_iterator` to `itk::FixedArray`. Included GTest unit tests.

This commit allows iteration over a `FixedArray` by means of a
C++11 range-based for-loop.

### DIFF
--- a/Modules/Core/Common/include/itkFixedArray.h
+++ b/Modules/Core/Common/include/itkFixedArray.h
@@ -128,6 +128,12 @@ public:
   /** A const reference to the ValueType. */
   using const_reference = const ValueType &;
 
+  /** The return type of the non-const overloads of begin() and end(). */
+  using iterator = ValueType *;
+
+  /** The return type of cbegin() and cend(), and the const overloads of begin() and end(). */
+  using const_iterator = const ValueType *;
+
   using SizeType = unsigned int;
 
 public:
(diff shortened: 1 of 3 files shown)


N-Dekker:
Note that `itk::FixedArray` already has its `Begin()` and `End()` member functions, which should also be fine, but they just don't get picked by range-based for-loops, because of the way they are spelled.

phcerdan:
Great idea!

Could you add rbegin and rend as well?

https://github.com/InsightSoftwareConsortium/ITK/blob/d6b48deea2ee9e7e6f006defa6823a136be056e6/Modules/Core/Common/include/itkFixedArray.hxx#L149-L191

N-Dekker:
FYI, the amend that I just did _only_ extended the unit tests for this feature.

N-Dekker:
@phcerdan
> Great idea!

Thanks, Pablo  `:-)`

> Could you add rbegin and rend as well?

`rbegin` and `rend` might also be interesting, indeed, but I would prefer to have it in a separate pull request. Is that OK to you? My aim for this pull request (#757) was just to add range-based for-loop support. By keeping the PR small, I hope that it can be reviewed and merged easily.

phcerdan:
Sure! Approved. Sorry for the brevity (answering from the phone)

PR #759, opened 2019-04-24
COMP: Add export specification to ITKSpatialObjects

### LINKED ISSUE #743: Potential link errors classes of SpatialObjects library
With commit 198b5c8f59fdb2b67b443f0455ce61b203c3fddd (Feb 10, 2019), `SpatialObjectProperty` member function implementations were moved to a cxx file. In order to export this functionality, I _think_ the class should have some kind of EXPORT macro. Something like `ITKSpatialObjects_EXPORT`? Possibly here:

https://github.com/InsightSoftwareConsortium/ITK/blob/2e23f710b387e8465dd37d102e245b5365c5831c/Modules/Core/SpatialObjects/include/itkSpatialObjectProperty.h#L36

Can you please check this (and please fix it, if necessary)? Stephen @aylward ? Or Hans @hjmjohnson ?

The issue might also apply to `MetaEvent` and `SpatialObjectFactoryBase`.

Note: This is not a high priority issue for me personally, because I usually link _statically_ to ITK. But it could be of importance to users who link dynamically (`BUILD_SHARED_LIBS`) to ITK.

thewtex:
Closes #743

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkMetaEvent.h
+++ b/Modules/Core/SpatialObjects/include/itkMetaEvent.h
@@ -21,6 +21,8 @@
 #include "itkMacro.h"
 #include "metaEvent.h"
 
+#include "ITKSpatialObjectsExport.h"
+
 namespace itk
 {
 /** \class MetaEvent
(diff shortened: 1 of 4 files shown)


thewtex:
Thanks for the review.

MetaImageMaskConverter, SpatialObjectFactory, and ObjectFactory are header-only, so they should not have the export specification. NiftiImageIOFactory looks like it has the export specification.

fbudin69500:
You are correct, those two other files that I mentioned are header only. Should we come up with a script (used in a test) that verifies the export symbols?

thewtex:
We should have errors on the dashboard if export specifications are not correct. It was not an issue with the *ITKSpatialObjects* module because `ENABLED_SHARED` was not set on the module -- it was always being built statically.

Issue #622, opened 2019-03-22
Clean up CMake options


dzenanz:
Some CMake settings are not marked as advanced, when they should be. And many of them are coming from third party libraries.

issakomi:
Good idea. BTW, several a little bit confusing options are coming from GDCM's openjpeg cmake file, there are usually on top of cmake-gui (e.g. BUILD_VIEWER, BUILD_JAVA). User can assume it is a kind of ITK's viewer or ITK's whatever java thingy.

PR #761, opened 2019-04-24
CMake advanced options

thewtex:
Mark options as advanced in the CMake GUI that are related to third party libraries.


thewtex:
Closes #622

thewtex:
It may still be possible to set them to use them -- this is just removing them from the CMake GUI options.

fbudin69500:
Possible but difficult as they are not exposed at all.

PR #769, opened 2019-04-25
BUG: Allow Python tests in module test/CMakeLists.txt

thewtex:
With ITK 5, the preferred location for Python tests is in the module
wrapping/test/CMakeLists.txt file. However, allow backwards-compatible
test specification in test/CMakeLists.txt for remote modules that have
not yet been updated.

To address:

  CMake Error at CMake/ITKModuleTest.cmake:224 (message):
    `itk_python_add_test` should never be called if `ITK_WRAP_PYTHON` if OFF
    Modules/Remote/FixedPointInverseDisplacementField/test/CMakeLists.txt:17 (itk_python_expression_add_test)


thewtex:
> Could the return be called only if we are building a remote module? I feel that this would be safer. This check allowed me to catch a few tests that I had missed when I moved the Python tests, and it would make sense to ensure that developers do not add new Python tests in the wrong place in the future.

Good idea -- added.

thewtex:
Merging to hit the Nightly

PR #774, opened 2019-04-25
DOC: Correct documented Canny Segmentation filter example

blowekamp:
The advection term is negative in the test, which is require for
outward propagation of the level set.


blowekamp:
Please see: https://github.com/InsightSoftwareConsortium/ITK/blob/c351787a261641a8b1c73e03f3dae37a40be967a/Examples/Segmentation/test/CMakeLists.txt#L229

PR #779, opened 2019-04-28
MI sample point histogram setting

hjmjohnson:
ENH: Identify histogram min/max for UseSampledPointSet
    
    When UseSampledPointSet is true (i.e. A sparse sampling point set is
    explicitly requested), compute the joint histogram range for the fixed
    image axis from those explicilty specified points.
    
    If m_UseSampledPointSet is true, then *ONLY* the sparse sampled points
    are used for analysis of the metric, and the fixed space intensity range
    values will be fixed to those values identified in the sparse sampled
    points.  The fixed masked value items are not relevant when the sparse
    sampling is set.


blowekamp:
The changes look good to me.

To nit pick about style. There has been criticism of using the C++ ternary operator. For this case it may be better, clearer, and shorter to use the `std::min` and `std::max` methods.

PR #781, opened 2019-04-28
ENH: Add a script to apply another script to all remotes

jhlegarreta:
Add a script to apply another script to all remotes.

This script is intended to ease the extension of toolkit-wide changes that
are applied through a bash script to the remote modules that dwell in/are
referenced from the toolkit's `Modules/Remote` module.

## PR Checklist
- [x] Adds the License notice to new files.
- [ ] Adds Python wrapping.
- [ ] Adds Documentation.

### DIFF
--- a/Utilities/Maintenance/ApplyScriptToRemotes.sh
+++ b/Utilities/Maintenance/ApplyScriptToRemotes.sh
@@ -1,0 +1,203 @@
+#!/bin/bash
+
+#==========================================================================
+#
+#   Copyright Insight Software Consortium
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#          http://www.apache.org/licenses/LICENSE-2.0.txt
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+#==========================================================================*/
+
+
+# This script, given a path to another script, applies this to the source
+# code of the remote modules in the ITK source tree whose CI status shows
+# a successful build.
+#
+# The script clones the successful remote modules to the ITK source tree,
+# creates a feature branch on each module, applies the provided script to the
+# source code of each module, adds the modified files to the git stage area,
+# commits them using the provided commit message, and pushes the commit to the
+# remote module repository.
+#
+# Please, review the commit message directives at:
+# https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md
+
+
+# Utility functions
+usage() {
+cat << EOF
+Usage: $0 <script_filename> <feature_branch> <commit_message>
+
+Use this script to apply another script to the source code of the remote
+remote modules in the ITK source tree whose CI status shows a successful
+build.
+
+The script clones the successful remote modules to the ITK source tree,
+creates a feature branch on each module, applies the provided script to the
+source code of each module, adds the modified files to the git stage area,
+commits them using the provided commit message, and pushes the commit to the
+remote module repository.
+
+Please, review the commit message directives at:
+https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md
+EOF
+}
+
+die() {
+  echo "$@" 1>&2; exit 1
+}
+
+# Parse arguments
+help=false
+script=""
+feature_branch=""
+commit_message=""
+while test $# -gt 0;
+do
+  opt="$1";
+  case "$opt" in
+    "-h"|"--help")
+      shift;
+      help=true
+      break;;
+    *)
+      break;;
+  esac
+done
+
+script="$1"
+feature_branch="$2"
+commit_message="$3"
+
+if test "${script}" = "" || "${feature_branch}" = "" || "${commit_message}" = "" || $help; then
+  usage
+  exit 1
+fi
+
+# Make sure we are inside the repository
+cd "${BASH_SOURCE%/*}" &&
+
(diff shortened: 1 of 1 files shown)


Utilities/Maintenance/ApplyScriptToRemotes.sh
```diff
@@ -0,0 +1,157 @@
...
+  cd ../../../ITK-build
+
+  cmake_remote_flags=()
+
+  # Switch on all remotes that have a success build status
+  for remote in ${remotes[@]}; do
+    cmake_remote_flags+=(-D'Module_'$remote':BOOL=ON' )
+  done
+
+  cmake ${cmake_remote_flags[@]} ../ITK
```
jhlegarreta: Completing all CMake configuration steps is not necessary, so should we just extract (if that is possible) the relevant parts from the CMakeLists.txt to clone the remotes in the ITK source tree?

Utilities/Maintenance/ApplyScriptToRemotes.sh
```diff
@@ -0,0 +1,157 @@
...
+    cd $remote
+
+    repository_basename=$(basename -s .git `git config --get remote.origin.url`)
+
+    git checkout master
+    git checkout -b $feature_branch origin/master
+    $script
+
+    # Add the files, adding filters if necessary
+    git add -u *.txt ./*.cmake ./*.rst ./*.py ./*.yml ./*.c ./*.cxx ./*.h ./*.hxx ./*.wrap
```
jhlegarreta: **Removed files are not added to the commit**. We could assume that a `git add . ` could be done since it will be a new branch, with no files and changes other than those applied by the script.

Utilities/Maintenance/ApplyScriptToRemotes.sh
```diff
@@ -0,0 +1,157 @@
...
+
+    # Commit and push to the feature branch
+    git commit -m $commit_message
+    git push --quiet https://$username:$password@github.com/$username/$repository_basename $feature_branch
```
jhlegarreta: - Rather that keeping the remote URLs from the `list_candidate_remotes` function, I do get the URL again in `apply_script_and_push_remotes`. I though it was easier than keeping a list from the former method, then finding the appropriate element for the remote at issue (partly because the URL includes the username and password for the push to be done). If you feel there is an alternative to this, please go ahead.

Utilities/Maintenance/ApplyScriptToRemotes.sh
```diff
@@ -0,0 +1,157 @@
...
+cd "${BASH_SOURCE%/*}" &&
+
+remote_modules_path='../../Modules/Remote'
+
+# Ask for GitHub username and password once
+echo "Please provide your GitHub (https://github.com)"
+read -p "username: " username
+read -p "password: " -s password
+
+script=$1
```
jhlegarreta: The **full path of the script** to be applied must be provided. Otherwise, we **could assume that only scripts that dwell in `Utilities` will be applied**, since they anyways will be worthwhile maintaining there for the records and for the sake of our philosophy.

Utilities/Maintenance/ApplyScriptToRemotes.sh
```diff
@@ -0,0 +1,157 @@
...
+feature_branch=$2
+commit_message=$3
```
jhlegarreta: The **commit message must be provided explicitly**. Many of the [ITKv5 preparation](https://github.com/InsightSoftwareConsortium/ITK/tree/master/Utilities/ITKv5Preparation) scripts do it all at once: do the changes, add them and commit with a message contained in the script itself. We should consider whether we prefer that the add and commit steps will generally be contained in the script that is applied.

jhlegarreta:
As discussed some time ago by mail. Took me a while to gather some time to do it, but here it goes.

Just as a reminder, this script can be used to apply changes such as the ones done for the [ITKv5 preparation](https://github.com/InsightSoftwareConsortium/ITK/tree/master/Utilities/ITKv5Preparation), or changes like removing the AppVeyor, Circle CI, and Travis CI files and adding the `azure-pipelines.yml` configuration, updating the `README.rst` files accordingly, etc., changes that may be related to C++14, 17 or 20, etc.

A few **thoughts** in-line, and other additional ones (**some may require careful consideration, and further work, so we should assume it will not be immediate to answer/address them**):
- We can rename the script if we find it is not too descriptive.
- We can change the help message formatting if it is not too bash script-like.
- The script **assumes that the user has all remotes forked**.
- However no **upstream** remote is configured. This is may be something that should be done (I have not tested push step).
- Not sure about the behavior of the script **if a few remotes have already been cloned, have local untracked or uncommitted changes, etc.** This should be considered at some point. 
- The script assumes that a **local branch with the same name as the one provided as the feature branch does not exist**. Should we remove it in case it exists?
- The script assumes that a **remote branch with the same name as the one provided as the feature branch does not exist**. Should we push force in case it exists?
- **Remote modules that are checked out are not deleted from the ITK source tree**. Should we delete them? I'd dare to say that if yes, we should handle the case where some remotes had already been cloned and had untracked, uncommitted, local branches, etc.


Many of the branch collision cases are covered by @fbudin69500 's [Slicer 3D extensions script](https://github.com/fbudin69500/NIRALSystemScripts/blob/master/UpdateSlicerExtensions.script). So it could be of help.

Once these aspects fixed, we could think of updating the [UpdateRequiredITKVersionInRemoteModules.sh](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/Maintenance/UpdateRequiredITKVersionInRemoteModules.sh) so that given the remote module name, it updates the `azure-pipelines.yml` and `setup.py` files automatically (provided that the former is always at the same location).

@maekclena This is related to the [remote module maintenance script](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/Maintenance/UpdateRemoteModules.sh), so you may want to have a look/improve it. Thanks.

jhlegarreta:
Thanks Hans. The PR had a few unresolved comments and requests for comment, but I think it is fine. We will improve it as we use it.

Issue #772, opened 2019-04-25
Pass image orientation information in ITK / VTK pipelines


thewtex:
### Description

Transfer the image orientation *Direction* matrix from ITK to VTK and VTK to ITK in ITKVtkGlue.

VTK Git `master` recently added orientation to *vtkImageData*. :tada: :taco: h/t @agirault 

### Steps to Reproduce

In [itkImageToVTKImageFilterTest](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Bridge/VtkGlue/test/itkImageToVTKImageFilterTest.cxx), set the *Direction* on the input *itk::Image*, and verify that it is correct on the output *vtkImageData*. Verify the fix in [itkVTKImageToImageFilterTest](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Bridge/VtkGlue/test/itkVTKImageToImageFilterTest.cxx) similarly.

To implement *Direction* passing, transfer the information in a similar way to the *Origin* and *Spacing* [in the ITKVTK module](https://github.com/InsightSoftwareConsortium/ITK/tree/master/Modules/Bridge/VTK/include).

To build the code and tests, enable the *Module_ITKVtkGlue* CMake option in ITK's build configuration, and point the build to the desired *VTK_DIR*.

### Expected behavior

Image orientation information is passed in ITK/VTK pipelines.

### Actual behavior

Image orientation information is lost.

### Reproducibility

100%.

### Versions

This will require VTK Git `master` (9.X) to be fixed.

CC: @aylward @martinken @lassoan @jcfr

agirault:
Followed your steps, and I get this with [itk/c351787](https://github.com/InsightSoftwareConsortium/ITK/commits/c351787) and [vtk/e2f92f6e](https://gitlab.kitware.com/vtk/vtk/commits/e2f92f6e) :
```ninja
[5481/5508] Building CXX object Modules/Bridge/VtkGlue/src/CMakeFiles/ITKVtkGlue.dir/QuickView.cxx.o
/itk/src/Modules/Bridge/VtkGlue/src/QuickView.cxx:25:10: fatal error: 'vtkVersion.h' file not found
#include "vtkVersion.h"
         ^~~~~~~~~~~~~~
1 error generated.

[...]

PR #790, opened 2019-04-30
Pass direction between itk and vtk images

agirault:
Address #772:

- [x] Improve VTKImage import/export tests to check for correct image properties
- [x] Pass directions in VTKImage import/export classes
- [x] Improve `itkVTKImageImport` to ~~consider which direction vectors to conserve when importing a VTK image (always in a 3D coordinate system) to an ITK image of dimension lower than 3. Currently picking the top left submatrix. (but that's already an issue for origin, spacing, and size?)~~ throw an exception when the direction matrix cannot be converted down to a lower dimension in ITK.
- [x] Check VTK version and add preprocessor definitions to only call VTK's direction API if it exists (past [868ba30c](https://gitlab.kitware.com/vtk/vtk/commits/868ba30c))

Thanks to @fbudin69500 for helping me getting acclimated to the ITK way 🙏

### DIFF
--- a/Modules/Bridge/VTK/include/itkVTKImageImport.hxx
+++ b/Modules/Bridge/VTK/include/itkVTKImageImport.hxx
@@ -98,6 +98,7 @@ VTKImageImport< TOutputImage >
   m_FloatSpacingCallback = nullptr;
   m_OriginCallback = nullptr;
   m_FloatOriginCallback = nullptr;
+  m_DirectionCallback = nullptr;
   m_UpdateInformationCallback = nullptr;
   m_ScalarTypeCallback = nullptr;
   m_DataExtentCallback = nullptr;
@@ -235,6 +236,19 @@ VTKImageImport< TOutputImage >
       }
     output->SetOrigin(outOrigin);
     }
+  if ( m_DirectionCallback )
+    {
+    double *inDirection = (m_DirectionCallback)( m_CallbackUserData );
+    typename TOutputImage::DirectionType outDirection;
+    for ( unsigned int i = 0; i < OutputImageDimension; ++i )
+      {
+      for ( unsigned int j = 0; j < OutputImageDimension; ++j )
+        {
+        outDirection[i][j] = inDirection[i*3+j];
+        }
+      }
+    output->SetDirection(outDirection);
+    }
   if ( m_NumberOfComponentsCallback )
     {
     const unsigned int components =
@@ -356,6 +370,10 @@ VTKImageImport< TOutputImage >
   if ( m_FloatOriginCallback )
     {
     os << "FloatOriginCallback: " << m_FloatOriginCallback << std::endl;
+    }
+  if ( m_DirectionCallback )
+    {
+    os << "DirectionCallback: " << m_DirectionCallback << std::endl;
     }
   if ( m_UpdateInformationCallback )
     {
--- a/Modules/Bridge/VTK/include/itkVTKImageImport.hxx
+++ b/Modules/Bridge/VTK/include/itkVTKImageImport.hxx
@@ -240,11 +240,30 @@ VTKImageImport< TOutputImage >
     {
     double *inDirection = (m_DirectionCallback)( m_CallbackUserData );
     typename TOutputImage::DirectionType outDirection;
-    for ( unsigned int i = 0; i < OutputImageDimension; ++i )
+    for ( unsigned int i = 0; i < 3; ++i )
       {
-      for ( unsigned int j = 0; j < OutputImageDimension; ++j )
+      for ( unsigned int j = 0; j < 3; ++j )
         {
-        outDirection[i][j] = inDirection[i*3+j];
+        if (i < OutputImageDimension && j < OutputImageDimension)
+          {
+          outDirection[i][j] = inDirection[i*3+j];
+          }
+        else if ((i != j) && (inDirection[i*3+j] != 0.0))
+          {
+            std::string ijk = "IJK";
+            std::string xyz = "XYZ";
+            itkExceptionMacro(<< "Cannot convert a VTK image to an ITK image of dimension "
+            << OutputImageDimension << " since the VTK image direction matrix element at ("
+            << i << "," << j << ") is not equal to 0.0:\n"
+            << "   I  J  K\n"
+            << "X  " << inDirection[0] << ", " << inDirection[1] << ", " << inDirection[2] << "\n"
+            << "Y  " << inDirection[3] << ", " << inDirection[4] << ", " << inDirection[5] << "\n"
+            << "Z  " << inDirection[6] << ", " << inDirection[7] << ", " << inDirection[8] << "\n"
+            << "This means that the " << ijk[j] << " data axis has a " << xyz[i]
+            << " component in physical space, but the ITK image can only represent values"
+            << " along " << ijk.substr(0,OutputImageDimension) << " projected on "
+            << xyz.substr(0,OutputImageDimension) << "." << std::endl);
+          }
         }
       }
     output->SetDirection(outDirection);
--- a/Modules/Bridge/VtkGlue/test/itkImageToVTKImageFilterTest.cxx
+++ b/Modules/Bridge/VtkGlue/test/itkImageToVTKImageFilterTest.cxx
@@ -85,6 +85,7 @@ int itkImageToVTKImageFilterTest(int, char *[])
       std::cerr << "Error: origins do not match for component (" << i << ")." << std::endl;
       return EXIT_FAILURE;
     }
+#if VTK_MAJOR_VERSION >= 9 || (VTK_MAJOR_VERSION == 8 && VTK_MINOR_VERSION >= 90)
     for (int j = 0; j < dim; ++j)
     {
       if (input->GetDirection()[i][j] != output->GetDirectionMatrix()->GetData()[i*3+j])
@@ -93,6 +94,7 @@ int itkImageToVTKImageFilterTest(int, char *[])
         return EXIT_FAILURE;
       }
     }
+#endif
   }
 
   return EXIT_SUCCESS;
--- a/Modules/Bridge/VtkGlue/test/itkImageToVTKImageFilterTest.cxx
+++ b/Modules/Bridge/VtkGlue/test/itkImageToVTKImageFilterTest.cxx
@@ -20,29 +20,80 @@
 
 #include "itkRandomImageSource.h"
 
+#include "vtkMatrix3x3.h"
+
 int itkImageToVTKImageFilterTest(int, char *[])
 {
-  using ImageType = itk::Image<float, 2 >;
+  const int dim = 3;
+  using ImageType = itk::Image<float, dim >;
   using SourceType = itk::RandomImageSource<ImageType>;
+  using SpacingType = SourceType::SpacingType;
+  using OriginType = SourceType::PointType;
+  using DirectionType = SourceType::DirectionType;
   using ConnectorType = itk::ImageToVTKImageFilter<ImageType>;
 
   ImageType::SizeType size;
-  size.Fill(20);
+  size[0] = 40;
+  size[1] = 10;
+  size[2] = 20;
   SourceType::Pointer source = SourceType::New();
   source->SetSize(size);
 
+  SpacingType spacing;
+  spacing[0] = 0.5;
+  spacing[1] = -2;
+  spacing[2] = 1;
+  OriginType origin;
+  origin[0] = -1.5;
+  origin[1] = 0.222;
+  origin[2] = 0;
+  DirectionType direction;
+  direction.Fill(0.0);
+  direction[0][1] = 1;
+  direction[1][0] = -1;
+  direction[2][2] = 0.7;
+  source->SetSpacing(spacing);
+  source->SetOrigin(origin);
+  source->SetDirection(direction);
+
+  source->Update();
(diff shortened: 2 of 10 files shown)


agirault:
FYI @thewtex: I mentioned the third point to Ken and it seems that we should just always import images from VTK as 3D image because there is no notion of a 2D image in VTK. Also, the issue we discussed for direction already applies to origin, spacing, and size, and ITK already just ignores the latest elements.

agirault:
FYI: @martinkem, @jcfr

Modules/Bridge/VTK/include/itkVTKImageImport.hxx
```diff
@@ -235,6 +236,31 @@ VTKImageImport< TOutputImage >
...
+        {
+        if (i < OutputImageDimension && j < OutputImageDimension)
+          {
+          outDirection[i][j] = inDirection[i*3+j];
+          }
+        else if ((i != j) && (inDirection[i*3+j] != 0.0))
+          {
+            char ijk[4] = "IJK";
+            char xyz[4] = "XYZ";
+            itkExceptionMacro(<< "Can not convert a VTK image to an ITK image of dimension "
```
fbudin69500: This is probably the safe way to do it but it does raise the problem of backward compatibility. Maybe this exception should be surrounded by `#ifndef ITK_LEGACY_SILENT`. @thewtex : what do you think? Also, `can not` ->`cannot`

Modules/Bridge/VtkGlue/test/itkVTKImageToImageFilterTest.cxx
```diff
@@ -33,15 +35,54 @@ int itkVTKImageToImageFilterTest(int, char*[])
...
+      {
+        std::cerr << "Error: directions do not match for component (" << i << "," << j << ")." << std::endl;
+        return EXIT_FAILURE;
+      }
+    }
+  }
 
-  connector->Print(std::cout);
+  // Try again with an invalid direction matrix and ensure it throws an exception
+  input->SetDirectionMatrix(0, 1, 0, 1, 0, 0, 1, 0, 0);
```
fbudin69500: Could you also test that it the matrix cannot be cropped, the appropriate exception is thrown?

Modules/Bridge/VtkGlue/test/itkVTKImageToImageFilterTest.cxx
```diff
@@ -33,15 +35,54 @@ int itkVTKImageToImageFilterTest(int, char*[])
...
+  try {
+    connector->Update();
+  } catch (itk::ExceptionObject & e) {
+    std::cout << "Expected exception was correctly thrown: " << e << std::endl;
```
fbudin69500: Make sure that the exception that is caught is the exception that we are expecting. There could be another problem and another exception could be thrown and caught.

Modules/Bridge/VtkGlue/test/itkImageToVTKImageFilterTest.cxx
```diff
@@ -20,29 +20,84 @@
...
 
+  ConnectorType::Pointer connector = ConnectorType::New();
+  connector->SetInput(input);
   connector->UpdateLargestPossibleRegion();
+  connector->Update();
 
-  connector->GetOutput()->Print(std::cout);
-  connector->GetImporter()->Print(std::cout);
-  connector->GetExporter()->Print(std::cout);
+  // connector->GetImporter()->Print(std::cout);
```
fbudin69500: I am guessing this is WIP?

Modules/Bridge/VTK/include/itkVTKImageImport.hxx
thewtex: I don't think we have to worry about backwards compatibility here because the *Direction* was never transferred.

fbudin69500: What I meant by backward compatibility is that someone who had some code working until now could see this exception appear

Modules/Bridge/VTK/include/itkVTKImageImport.hxx
```diff
@@ -235,6 +236,31 @@ VTKImageImport< TOutputImage >
...
+      {
+      for ( unsigned int j = 0; j < 3; ++j )
+        {
+        if (i < OutputImageDimension && j < OutputImageDimension)
+          {
+          outDirection[i][j] = inDirection[i*3+j];
+          }
+        else if ((i != j) && (inDirection[i*3+j] != 0.0))
+          {
+            char ijk[4] = "IJK";
```
phcerdan: Tiny comment, you can make both char arrays `const`.

Modules/Bridge/VtkGlue/test/itkVTKImageToImageFilterTest.cxx
agirault: That's what this should be testing here.

Modules/Bridge/VtkGlue/test/itkVTKImageToImageFilterTest.cxx
agirault: Makes sense. Just a string compare?

Modules/Bridge/VtkGlue/test/itkImageToVTKImageFilterTest.cxx
agirault: There were a lot of print calls which were not testing anything and where pretty verbose so I commented them. Didn't remove them because I figured someone could just uncomment them and inspect the log if the tests failed, but happy to either remove them entirely or uncomment them if that's preferred. Same for other test I believe.

Modules/Bridge/VtkGlue/test/itkVTKImageToImageFilterTest.cxx
fbudin69500: yes, that's probably the only way you can test that.

Modules/Bridge/VtkGlue/test/itkVTKImageToImageFilterTest.cxx
fbudin69500: Looking at your matrix, it looked like your matrix is just not a great direction matrix as it is not a rotation matrix since the second and third rows are equal (determinant of your matrix is 0).

Modules/Bridge/VtkGlue/test/itkImageToVTKImageFilterTest.cxx
fbudin69500: I don't have a strong opinion on this, but I would prefer for these to either be kept (that might allow easier debugging in the future, and it the test pass we don't have to look at it) or completely removed (in a separate commit that has a short explanation). The second solution would allow one to understand why this was removed when looking at the git log.

Modules/Bridge/VTK/include/itkVTKImageImport.hxx
agirault: So:
- `ITK_LEGACY_SILENT` to ensure we don't throw an exception if we didn't before
- another preprocessor definition based on the VTK version to avoid importing/exporting the direction altogether

Is that right?

Modules/Bridge/VtkGlue/test/itkVTKImageToImageFilterTest.cxx
agirault: Yeah I messed up while testing around, nice catch thanks.
I plan to have:
```
0 1 0
0 0 1
1 0 0
```

Modules/Bridge/VtkGlue/test/itkVTKImageToImageFilterTest.cxx
agirault: I can check `e.GetLocation()` and `e.GetDescription()`: maybe checking the location would make more sense?

Modules/Bridge/VtkGlue/test/itkImageToVTKImageFilterTest.cxx
agirault: I'll remove them since they're not useful for the test.

agirault:
@fbudin69500 @phcerdan thanks for the review. I addressed everything apart from the preprocessor definitions I still need to figure out.

agirault:
Fixed warnings [here](https://github.com/InsightSoftwareConsortium/ITK/compare/1d4c467df351fd9751b022326d0a6de00a383561..43ae9cb205bad69bbc509112cd9ddec184252398).
```cpp
warning: comparison of integers of different signs: 'const itk::Size::value_type' (aka 'const unsigned long') and 'int' [-Wsign-compare]
    if (output->GetLargestPossibleRegion().GetSize()[i] != input->GetDimensions()[i])
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~
```

cc: @thewtex

Modules/Bridge/VtkGlue/test/itkImageToVTKImageFilterTest.cxx
```diff
@@ -20,29 +20,80 @@
...
+  connector->Update();
 
-  connector->Print(std::cout);
+  auto output = connector->GetOutput();
+  output->Print(std::cout);
 
-  connector->Update();
+  for (int i = 0; i < dim; ++i)
+  {
+    if (static_cast<int>(input->GetLargestPossibleRegion().GetSize()[i]) != output->GetDimensions()[i])
```
fbudin69500: It would be safer to cast the VTK dimensions to  `itk::SizeValueType` since it is (at least currently) `unsigned long` which can be larger than `int`, although that doesn't solve the sign issue. Best might be to make sure that `output->GetDimension()[i]` is positive before it is casted, to avoid any possible issue that would be hard to track if a bug was arising from this.

Modules/Bridge/VtkGlue/CMakeLists.txt
```diff
@@ -46,6 +46,8 @@ endif()
 set(_target_prefix "vtk")
 if(VTK_VERSION VERSION_GREATER_EQUAL 8.90.0)
   set(_target_prefix "VTK::")
+  message(STATUS "VTK supports image direction")
+  set(VTK_SUPPORTS_DIRECTION 1) #definition used in itkVtkGlueConfigure.h.in
```
thewtex: It would be better to use VTK version C preprocessor macros defined in *vtkVersion.h* instead of CMake variables. Although we could technically say that we do not support it, someone may reasonably try to use the ITKVtkGlue headers in a project where they use a different VTK than the version that ITKVtkGlue was built with. Also, it keeps things simpler and more direct in the implementation. For example, there is a bugs related to the complexity right now, i.e. *itkVtkGlueConfigure.h* is not installed.

jcfr: Now that i see the use case, what about adding `VTK_SUPPORT_DIRECTION` directly in `vtkConfigure.h` (or `vtkCommonDataModelConfigure.h`)

fbudin69500: You are correct because the value of this macro is not used inside any `cxx` file build as part of this library. If it were the case, then for consistency we would want to define that extra macro.

Modules/Bridge/VTK/include/itkVTKImageImport.hxx
```diff
@@ -240,11 +240,23 @@ VTKImageImport< TOutputImage >
...
-        outDirection[i][j] = inDirection[i*3+j];
+        if (i < OutputImageDimension && j < OutputImageDimension)
+          {
+          outDirection[i][j] = inDirection[i*3+j];
+          }
+        else if ((i != j) && (inDirection[i*3+j] != 0.0))
+          {
+            char ijk[4] = "IJK";
+            char xyz[4] = "XYZ";
+            itkExceptionMacro(<< "Can not convert a VTK image to an ITK image of dimension "
```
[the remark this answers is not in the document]
agirault: It's always in 3d

Modules/Bridge/VTK/include/itkVTKImageImport.hxx
```diff
@@ -240,11 +240,23 @@ VTKImageImport< TOutputImage >
...
+            << OutputImageDimension << " since the VTK image direction matrix has a non-zero element at ("
```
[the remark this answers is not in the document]
agirault: I can make some of those changes, apart from "The most probable issue": it's not the most probable, it is the issue.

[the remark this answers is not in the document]
agirault: Description updated:
Screen Shot 2019-05-02 at 11 09 18 AM

Modules/Bridge/VtkGlue/CMakeLists.txt
agirault: @thewtex updates as recommended. Current use is:
```c++
#if VTK_MAJOR_VERSION >= 9 || (VTK_MAJOR_VERSION == 8 && VTK_MINOR_VERSION >= 90)
```
but we could update to just this in the future:
```c++
#if VTK_MAJOR_VERSION >= 9
```

Let me know if we want to consider adding another preprocessor definition in VTK (but maybe it becomes confusing to have that many, and checking the version is easier for users?)

jcfr: Direction support being a major addition, i think we should add a proper macro in VTK.

PR #793, opened 2019-04-30
WIP: Imread fallback only

fbudin69500:
`imread()` Python function allows to read an image, either by letting the
function deduce the input image pixel type or by manually selecting the
pixel type used to read it. The image dimension is always automatically
deduced.

In certain case, the user want to use the actual image pixel type if
that is possible, and use a fallback type if the input image type is
not supported. The new `fallback_only` parameter allows this.


fbudin69500:
#794 has been merged.

PR #801, opened 2019-05-01
BUG: PythonFindEmptyClasses test for windowed-sinc functions

### LINKED ISSUE #797: Broken test `PythonFindEmptyClasses`
### Description

The test `PythonFindEmptyClasses` is broken. This is most likely due to [this commit](https://github.com/InsightSoftwareConsortium/ITK/commit/37e0474669a6f68026e718114f584a8f631eec84).

The result of the test can be seen [here](https://open.cdash.org/viewTest.php?onlyfailed&buildid=5885143).

thewtex:
Exclude these classes because they only expose `operator()`.

Also do not wrap with POINTER because they do not inherit from
itk.LightObject.

Closes #797

### DIFF
--- a/Wrapping/Generators/Python/Tests/findEmptyClasses.py
+++ b/Wrapping/Generators/Python/Tests/findEmptyClasses.py
@@ -67,6 +67,10 @@ exclude = ["AuthalicMatrixCoefficients",
            "vnl_file_matrix",
            "vnl_file_vector",
            "vnl_fortran_copy",
+           "CosineWindowFunction",
+           "HammingWindowFunction",
+           "LanczosWindowFunction",
+           "WelchWindowFunction",
            ]
 
 total = 0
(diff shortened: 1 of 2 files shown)


Wrapping/Generators/Python/Tests/findEmptyClasses.py
```diff
@@ -67,6 +67,10 @@ def isEmpty(o):
            "vnl_file_matrix",
            "vnl_file_vector",
            "vnl_fortran_copy",
+           "CosineWindowFunction",
```
fbudin69500: Those functions are not currently wrapped, so I don't think it is necessary to exclude them (at least for now).

fbudin69500: https://github.com/InsightSoftwareConsortium/ITK/blob/651ee50286385b5913dc7c34a8cb67fa3fcab85c/Modules/Core/ImageFunction/wrapping/itkWindowedSincInterpolateImageFunction.wrap#L5-L8

fbudin69500: Those functions are indeed wrapped here: https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/ImageFunction/wrapping/itkWindowedSincInterpolateImageFunction.wrap#L7

PR #805, opened 2019-05-01
STYLE: SpatialObject() = default, const BoundingBox, Transform members

N-Dekker:
Explicitly-defaulted `SpatialObject()` default-constructor.

Added `const` to `BoundingBoxPointer` and `TransformPointer` data
members of `SpatialObject`. The aim is to ensure that these pointer
are always valid (never null) during the lifetime of the spatial object.

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkSpatialObject.h
+++ b/Modules/Core/SpatialObjects/include/itkSpatialObject.h
@@ -522,8 +522,11 @@ protected:
   /** Compute bounding box for the object in world space */
   virtual void ComputeMyBoundingBox();
 
-  /** Constructor. */
-  SpatialObject();
+  /** Default constructor. Ensures that its bounding boxes are empty (all
+  * bounds zero-valued), its list of children is empty, and its transform
+  * objects identical to the identity transform, initially.
+  */
+  SpatialObject() = default;
 
   /** Destructor. */
   ~SpatialObject() override;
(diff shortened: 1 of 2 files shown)


Modules/Core/SpatialObjects/include/itkSpatialObject.h
```diff
@@ -522,8 +522,11 @@ class ITK_TEMPLATE_EXPORT SpatialObject:
   /** Compute bounding box for the object in world space */
   virtual void ComputeMyBoundingBox();
 
-  /** Constructor. */
-  SpatialObject();
+  /** Default constructor. Ensures that its bounding boxes are empty (all
+  * bounds zero-valued), its list of children is empty, and its transform
+  * objects identical to the identity transform, initially.
+  */
+  SpatialObject() = default;
```
blowekamp: For libraries with compiles objects it is best practice to keep the implementation in the compiled cxx files. Here is a little example of the problems that can arise with shared libraries when the descrutor is implicitly instantiated:
https://github.com/blowekamp/PrivateDynamicCastTest/pull/1

The destructor should be explicitly instantiated and compiled in the library.

blowekamp:
Nice update to use a simpler constructor with member initialization!

It's OK to have inline destructors for template classes.

N-Dekker:
Thanks, Bradley @blowekamp 

The force-push that I just did does not have any code changes. It is a rebase and an amend of the author date, just to trigger a rebuild, hoping to get rid of the Linux-only test failure:

> Test: itkStatisticsUniqueLabelMapFilterTest1 (Failed)
> Build: Linux-Build1706-PR805-Defaulted-Constructor-SpatialObject (Azure.fv-az619) on 2019-05-01 19:22:20
> Repository revision: 3b6785fd91d5d93282542ee36c392546edb03c8d
> Test Details: Completed (Failed)

https://open.cdash.org/testDetails.php?test=759766053&build=5886758

N-Dekker:
All tests passed successfully  `:-)`. I think this nice little pull request is ready for merge!

N-Dekker:
Please merge this one! I really think it's OK, it has a nice comment from @blowekamp and an approval from @dzenanz

PR #807, opened 2019-05-02
Improve python wrapping error messages

fbudin69500:
This PR improves Python error messages when trying to instantiates an ITK object with template parameters that have not been wrapped. The error message relies on a new ITK Python function that prints the Python type of an ITK object in a way that can simply be copy/pasted by the user in Python instead of a C++ style.


fbudin69500:
The failing test seems unrelated and should be resolved by #801.

Issue #809, opened 2019-05-02
Deprecate WarpImageFilter and WarpVectorImageFilter


blowekamp:
These two filers apply deformation fields to images ( and vector images ). The same functionality now exist in the `ResampleImageFilter`. 

Should these filters be deprecated to avoid confusion?

blowekamp:
@thewtex @ntustison @hjmjohnson 
Any thought on these classes?

hjmjohnson:
@blowekamp I am in favor of deprecating as long a documentation clearly expresses how to get the same results with ResampleImageFilter (Including an example refactoring).

Thanks Brad.

blowekamp:
Here is list of examples which use the `WarpImageFilter`:
```
Filtering/WarpImageFilter1.cxx
RegistrationITKv3/DeformableRegistration10.cxx
RegistrationITKv3/DeformableRegistration16.cxx
RegistrationITKv3/DeformableRegistration17.cxx
RegistrationITKv3/DeformableRegistration2.cxx
RegistrationITKv3/DeformableRegistration3.cxx
RegistrationITKv3/DeformableRegistration5.cxx
RegistrationITKv3/DeformableRegistration9.cxx
RegistrationITKv3/LandmarkWarping2.cxx
RegistrationITKv4/DeformableRegistration10.cxx
RegistrationITKv4/DeformableRegistration16.cxx
RegistrationITKv4/DeformableRegistration17.cxx
RegistrationITKv4/DeformableRegistration2.cxx
RegistrationITKv4/DeformableRegistration3.cxx
RegistrationITKv4/DeformableRegistration5.cxx
RegistrationITKv4/DeformableRegistration9.cxx
RegistrationITKv4/LandmarkWarping2.cxx
```

Looks like step one is to update those examples to use the `ResampleImageFilter`.

blowekamp:
We need to figure out what to do with the ITKv3 Registration examples...

hjmjohnson:
The ITKv3 examples should definitely be deprecated, and I would recommend removing them (as they are now confusing to users who accidentally end up in that directory.

hjmjohnson:
On second thought, perhaps we should just move them out of the "examples" directory and into a "testing" subdirectory as they may still have code coverage utility.

ITKv6 should evaluate if the ITKv3 framework is completely removed.

blowekamp:
I agree that it is important to keep them as testing while we have the ITKv3 registration framework still there.

[...]

PR #813, opened 2019-05-02
STYLE: Remove ITKv3 Registration examples

hjmjohnson:
Remove redundancy for ITKv3 examples with respect to ITKv4.

As ITKv5 does not support ITKv3 registration any longer, completely remove the examples.

## PR Checklist
- No [Makes breaking changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes)
- No [Makes design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes)
- No Adds the License notice to new files.
- No Adds Python wrapping.
- No Adds tests and baseline comparison (quantitative).
- No [Adds test data](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/UploadBinaryData.md).
- [x] Removes outdated ITKv3 Registration Examples to [ITKExamples](https://github.com/InsightSoftwareConsortium/ITKExamples)
- No Adds Documentation.


hjmjohnson:
Based on comments in #809 we wanted to keep ITKv3 registration frameworks there for testing.  After a considerable effort to minimize the duplication between ITKv3 and ITKv4 examples, I identified that the ITKv3 registration code was NOT USED FOR TESTING, and had been turned off and inaccessible since early in the ITKv5 conversion (as evidenced by lack of support for new spatial objects, and many C++11 warning oversights, or refactoring skips).

In the end I think that removing the directory is the best option, but the intermediate path to getting there will provide granular restore points if it becomes necessary to restore these outdated examples.  I'd rather not squash them (which would be cleaner), but less granular for restore points.

blowekamp:
I got most of them working here #814

- likely c++11 warnings
- many of the RegistrationITKv3 examples are duplicated in the RegistrationITKv4 directory.
- many of the RegistrationITKv4 examples still use the old registration frame work

not sure what the path forward is here.

blowekamp:
Nice work cleaning up the duplicates! I'm totally for merging that ASAP.

As for removing the reminder, I'm not on that page yet.

hjmjohnson:
@blowekamp Give me an hour or two.  I think I have a good solution. When "BUILD_EXAMPLES" is turned on, the non-duplicated code will be built and tested (i.e. improve code coverage, and exercise old code in a way that was at one point our best practices suggestions).

I'm working through a linking issues now.

hjmjohnson:
@blowekamp @dzenanz @thewtex  This took a bit longer than expected to move the code around.  I'd love to have this merged in soon so that the content in the Examples directory is only demonstrating modern best practices, and so that the long-dormant untested code starts to be tested again (at least when Examples are built).

blowekamp:
Looks great!

Could we enable the examples and the legacy testing, just to test how the CI does with these examples?

dzenanz:
I will add examples and ITKFEMRegistration to this [build](https://open.cdash.org/buildSummary.php?buildid=5894171).

PR #815, opened 2019-05-03
WIP: Improve python error message if image file doesnt exist

fbudin69500:
… exist

When opening an image file in Python through the `itkTemplate` class (i.e.
`itk.ImageFileReader.New(FileName="my_file.png")`, the error message
warns the user that the file does not exist. The error message matches
the error message printed in C++ or when using directly the `ImageFileReader`
class (i.e. `itk.ImageFileReader[ImageType].New(FileName="my_file.png")`).


thewtex:
Merged to `master` in https://github.com/InsightSoftwareConsortium/ITK/pull/910

PR #816, opened 2019-05-03
ENH: Warning when trying to open in Python an image file that doesn't…

fbudin69500:
… exist

When opening an image file in Python through the `itkTemplate` class (i.e.
`itk.ImageFileReader.New(FileName="my_file.png")`, the error message
warns the user that the file does not exist. The error message matches
the error message printed in C++ or when using directly the `ImageFileReader`
class (i.e. `itk.ImageFileReader[ImageType].New(FileName="my_file.png")`).

### DIFF
--- a/Wrapping/Generators/Python/Tests/PythonTemplateTest.py
+++ b/Wrapping/Generators/Python/Tests/PythonTemplateTest.py
@@ -76,18 +76,27 @@ reader2 = readerType.New()
 reader = readerType.New(FileName='test.png')
 assert reader.GetFileName() == 'test.png'
 
-# wwith a wrong attribute name
+# with a wrong attribute name
 try:
     reader = readerType.New(WrongName='test.png')
     raise Exception('no exception sent for wrong attribute name')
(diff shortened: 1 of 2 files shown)


fbudin69500:
Matches [C++ error message](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/IO/ImageBase/include/itkImageFileReader.hxx#L259-L269).

fbudin69500:
@cdeepakroy

Wrapping/Generators/Python/Tests/PythonTemplateTest.py
```diff
@@ -90,6 +90,15 @@
 except:
     pass
 
+# wwith a wrong file name
```
hjmjohnson: wwith -> with

fbudin69500: Good catch. I had copy pasted the string that already contained the error. I updated my PR and corrected the original mistake too.

PR #817, opened 2019-05-03
COMP: Correct continuous index type

blowekamp:
Support when the interpolator precision is a different type than then
transform precision type.

### DIFF
--- a/Modules/Filtering/ImageGrid/include/itkResampleImageFilter.h
+++ b/Modules/Filtering/ImageGrid/include/itkResampleImageFilter.h
@@ -169,7 +169,7 @@ public:
 
   /** Input pixel continuous index typdef */
   using ContinuousInputIndexType =
-      ContinuousIndex< TTransformPrecisionType, ImageDimension >;
+      ContinuousIndex< TInterpolatorPrecisionType, ImageDimension >;
 
   /** Typedef to describe the output image region type. */
   using OutputImageRegionType = typename TOutputImage::RegionType;
--- a/Modules/Filtering/ImageGrid/test/itkResampleImageFilterGTest.cxx
+++ b/Modules/Filtering/ImageGrid/test/itkResampleImageFilterGTest.cxx
@@ -66,6 +66,11 @@ void Expect_ResampleImageFilter_preserves_pixel_value(const TPixel inputPixel)
 
 } // namespace
 
+// Compile time check of mixing transform and precision types
+template class itk::ResampleImageFilter<itk::Image<int>, itk::Image<int>, float, float >;
+template class itk::ResampleImageFilter<itk::Image<int>, itk::Image<int>, double, float >;
+template class itk::ResampleImageFilter<itk::Image<int>, itk::Image<int>, float , double >;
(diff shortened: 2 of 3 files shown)


Modules/Filtering/ImageGrid/include/itkResampleImageFilter.h
```diff
@@ -169,7 +169,7 @@ class ITK_TEMPLATE_EXPORT ResampleImageFilter :
 
   /** Input pixel continuous index typdef */
   using ContinuousInputIndexType =
-      ContinuousIndex< TTransformPrecisionType, ImageDimension >;
+      ContinuousIndex< TInterpolatorPrecisionType, ImageDimension >;
```
N-Dekker: Just checked: this change does indeed fix compilation errors when TInterpolatorPrecisionType != TTransformPrecisionType.

Modules/Filtering/ImageGrid/test/itkResampleImageFilterGTest.cxx
```diff
@@ -66,6 +66,17 @@ void Expect_ResampleImageFilter_preserves_pixel_value(const TPixel inputPixel)
...
+{
+  using PixelType = int;
+  using ImageType = itk::Image<PixelType>;
+
+  const auto filter1 = itk::ResampleImageFilter<ImageType, ImageType >::New();
+  const auto filter2 = itk::ResampleImageFilter<ImageType, ImageType, double, double >::New();
+  const auto filter3 = itk::ResampleImageFilter<ImageType, ImageType, double, float >::New();
+  const auto filter4 = itk::ResampleImageFilter<ImageType, ImageType, float, double >::New();
+  const auto filter5 = itk::ResampleImageFilter<ImageType, ImageType, float, float >::New();
+}
```
N-Dekker: Instead of creating all those filters at run-time, you could also just instantiate the templates at compile-time, without  using them. For example, at file scope:

    template class itk::ResampleImageFilter<itk::Image<int>, itk::Image<int>, double, float >;
    template class itk::ResampleImageFilter<itk::Image<int>, itk::Image<int>, float , double >;

Then the `TEST(...)` function is not necessary anymore.

Modules/Filtering/ImageGrid/include/itkResampleImageFilter.h
blowekamp: The driving problem is replacing the WarpImageFilter with this one, and ran into this issue.

N-Dekker:
@blowekamp Looks like you placed your template instantiations within the existing unnamed namespace of the test. GCC and Clang don't seem to like that (https://open.cdash.org/viewBuildError.php?buildid=5890808):

>error: explicit instantiation of 'class itk::ResampleImageFilter<itk::Image<int>, itk::Image<int>, float, float>' in namespace '{anonymous}'... 

For `IndexRange`, I placed the tested template instantiations outside any namespace. That appeared to work fine, for any compiler: 

https://github.com/InsightSoftwareConsortium/ITK/blob/v5.0rc02/Modules/Core/Common/test/itkIndexRangeGTest.cxx#L26

Maybe you should do the same for `ResampleImageFilter`.

PR #818, opened 2019-05-03
ENH: Prefer the usage of the ResampleImageFilter over the Warp

blowekamp:


blowekamp:
@N-Dekker You asked.

blowekamp:
@hjmjohnson Before take this patch any further, this ended up being more verbose than expected. Do you still think it is a good idea to make the change the resample filter? or have any suggestions of improving the style of the code?

PR #821, opened 2019-05-05
BUG: itkConfigure.h defines ITK_DEFAULT_MAX_THREADS

hjmjohnson:
Missing header for definition of ITK_MAX_THREADS was affected by order
of include files in external programs using ITK.

ITK/Modules/Core/Common/include/itkThreadSupport.h:46:43: error: use of undeclared identifier 'ITK_DEFAULT_MAX_THREADS'
  constexpr std::size_t ITK_MAX_THREADS = ITK_DEFAULT_MAX_THREADS;

### DIFF
--- a/Modules/Core/Common/include/itkThreadSupport.h
+++ b/Modules/Core/Common/include/itkThreadSupport.h
@@ -36,6 +36,7 @@
 #include "itkWindows.h"
 #include <winbase.h>
 #endif
+#include "itkConfigure.h"
 
 
 namespace itk


Modules/Core/Common/src/itkTBBMultiThreader.cxx
```diff
@@ -137,7 +137,6 @@ namespace
 {
 struct TBBImageRegionSplitter : public itk::ImageIORegion
 {
-  static const bool is_splittable_in_proportion = true;
```
dzenanz: I think we need this, because TBB checks its existence.
https://software.intel.com/en-us/node/506143

Modules/Core/Common/include/itkThreadSupport.h
```diff
@@ -36,6 +36,7 @@
 #include "itkWindows.h"
 #include <winbase.h>
 #endif
+#include "itkConfigure.h"
```
dzenanz: Thanks for fixing this.

Modules/Core/Common/src/itkTBBMultiThreader.cxx
dzenanz: It is better to have TBB call our proportional split, then to emulate proportional splitting by repeated calls to our 1:1 split.

hjmjohnson: @dzenanz I'm splitting this into two PR's. This PR is now only the single header fix.

Regarding the TBB, I don't know this behavior of TBB well enough to answer your question.
I'm added a proposal in #826, but it is not very satisfying.  Your advice would be greatly appreciated.

dzenanz:
Superseded by #826.

Modules/Core/Common/src/itkTBBMultiThreader.cxx
dzenanz: I saw this comment after closing this PR. #826 also includes the header fix. Feel free to split and reopen if you wish.

hjmjohnson:
Properly split now.  Sorry @dzenanz for being sloppy.

PR #822, opened 2019-05-05
ENH: colorspace for DICOM YBR greenish-blue images

issakomi:
@thewtex  @dzenanz 

Correct color for YBR_FULL_422 (usually JPEG lossy compressed Ultrasound) and YBR_FULL (usually RLE compressed Ultrasound). GDCM IO is highly likely the right place to do.

Sample [data sets](https://drive.google.com/file/d/1WHlKAaryx5ujD2HhWPTppfJtSY8aNNAQ/view?usp=sharing).

### DIFF
--- a/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
+++ b/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
@@ -339,6 +339,36 @@ void GDCMImageIO::Read(void *pointer)
     len = len * outputpt.GetPixelSize() / pixeltype.GetPixelSize();
     }
 
+  if ( ( pi == gdcm::PhotometricInterpretation::YBR_FULL_422 ||
+         pi == gdcm::PhotometricInterpretation::YBR_FULL ) &&
+       ( pixeltype == gdcm::PixelFormat::UINT8 ||
+         pixeltype == gdcm::PixelFormat::INT8 ) &&
+       m_NumberOfComponents == 3 )
+  {
+    unsigned char * copy = reinterpret_cast<unsigned char *>(pointer);
+    for (unsigned long x = 0; x < len; x+=3)
+    {
+      const unsigned char a = copy[x  ];
+      const unsigned char b = copy[x+1];
+      const unsigned char c = copy[x+2];
+      int R = 38142*(a-16)+52298*(c-128);
+      int G = 38142*(a-16)-26640*(c-128)-12845*(b-128);
+      int B = 38142*(a-16)+66093*(b-128);
+      R = (R+16384)>>15;
+      G = (G+16384)>>15;
+      B = (B+16384)>>15;
+      if (R <   0) R =   0;
+      if (R > 255) R = 255;
+      if (G <   0) G =   0;
+      if (G > 255) G = 255;
+      if (B <   0) B =   0;
+      if (B > 255) B = 255;
+      copy[x  ]=(unsigned char)(R);
+      copy[x+1]=(unsigned char)(G);
+      copy[x+2]=(unsigned char)(B);
+    }
+  }
+
 #ifndef NDEBUG
   // \postcondition
   // Now that len was updated (after unpacker 12bits -> 16bits, rescale...) ,
--- a/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
+++ b/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
@@ -342,8 +342,7 @@ void GDCMImageIO::Read(void *pointer)
               pi == gdcm::PhotometricInterpretation::YBR_FULL ) &&
             ( pixeltype == gdcm::PixelFormat::UINT8 ||
               pixeltype == gdcm::PixelFormat::INT8 ) &&
-            ( len%3 == 0 ) &&
-            m_NumberOfComponents == 3 )
+            len%3 == 0 && m_NumberOfComponents == 3 )
     {
     unsigned char * copy = reinterpret_cast<unsigned char *>(pointer);
     for (unsigned long long x = 0; x < len; x+=3)
--- a/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
+++ b/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
@@ -342,7 +342,7 @@ void GDCMImageIO::Read(void *pointer)
               pi == gdcm::PhotometricInterpretation::YBR_FULL ) &&
             ( pixeltype == gdcm::PixelFormat::UINT8 ||
               pixeltype == gdcm::PixelFormat::INT8 ) &&
-            len%3 == 0 && m_NumberOfComponents == 3 )
+            ( len%3 == 0 ) && m_NumberOfComponents == 3 )
     {
     unsigned char * copy = reinterpret_cast<unsigned char *>(pointer);
     for (unsigned long long x = 0; x < len; x+=3)
--- a/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
+++ b/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
@@ -338,16 +338,15 @@ void GDCMImageIO::Read(void *pointer)
     // WARNING: sizeof(Real World Value) != sizeof(Stored Pixel)
     len = len * outputpt.GetPixelSize() / pixeltype.GetPixelSize();
     }
-
-  if ( ( pi == gdcm::PhotometricInterpretation::YBR_FULL_422 ||
-         pi == gdcm::PhotometricInterpretation::YBR_FULL ) &&
-       ( pixeltype == gdcm::PixelFormat::UINT8 ||
-         pixeltype == gdcm::PixelFormat::INT8 ) &&
-       m_NumberOfComponents == 3 )
-  {
-    unsigned char * copy = reinterpret_cast<unsigned char *>(pointer);
-    for (unsigned long x = 0; x < len; x+=3)
+  else if ( ( pi == gdcm::PhotometricInterpretation::YBR_FULL_422 ||
+              pi == gdcm::PhotometricInterpretation::YBR_FULL ) &&
+            ( pixeltype == gdcm::PixelFormat::UINT8 ||
+              pixeltype == gdcm::PixelFormat::INT8 ) &&
+            m_NumberOfComponents == 3 )
     {
+    unsigned char * copy = reinterpret_cast<unsigned char *>(pointer);
+    for (unsigned long long x = 0; x < len; x+=3)
+      {
       const unsigned char a = copy[x  ];
       const unsigned char b = copy[x+1];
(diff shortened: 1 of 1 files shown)


dzenanz:
@malaterre is the go-to person for GDCM.

issakomi:
OK, let them be greenish-blue, like in ITK-Snap and Slicer since at least 7 years, i don't have the problem in my app.

malaterre:
@issakomi Thanks for the patch, I guess next time you could re-use gdcmImageChangePhotometricInterpretation (http://gdcm.sourceforge.net/3.0/html/classgdcm_1_1ImageChangePhotometricInterpretation.xhtml). 

Anyway I agree with @zivy the color space in ITK is implicit. You know you have one component, or three components but that is all. I suspect you could reproduce this issue with other format with an ICC profile. In any case both non-RGB (YBR style) and MONOCHROME1 will give those visual artifacts. When I wrote the first implementation I left it this way because the ITK design is heavily templated over the pixel type, so if all your inputs is MONOCHROME1, then you really mean to process your inputs using MONOCHROME1 (same goes for YBR). ITK is about processing the pixel data buffer not displaying it on RGB monitor (AFAIK).

The visual artifacts could be described in the doxygen doc though.

issakomi:
@malaterre 
Sure, it is reason i didn't try this in GDCM itself, but here, it ITK's IO, ITK's GDCMIO returns
RGB image, it is a final image for viewing, not for a kind of processing,
there are also no clear instruction for GDCMIO user what to do. It is reason why all ITK
based viewers (Slicer, ITK-Snap, all i've seen except my own one) return to users
greenish-blue something. Other DICOM viewers, not ITK-based,  open those images
correctly, simply try Aeskulap for e.g.
Also if i save binary stream "as is" is from DICOM JPEG lossy 2D Ultrasound image and
re-name to .jpg - all viewers show correct picture, not greensh-blue.
So i am still sure it is bad in ITK  GDCMIO, but having a flag as @zivy suggests is sure a
good idea. If @malaterre  @dzenanz want - i'll open another PR. Let me know.
Thanks.

malaterre:
@issakomi I understand what you were trying to do. I am ok with it as long as someone from ITK dev, finally decide that all IO drivers are required to do colorspace conversion to grayscale (min-is-black) and RGB. To address @zivy concern one simply need to introduce the opposite step in the writing step to achieve full round-trip operation (Am I doing it for PALETTE anyway?).

issakomi:
BTW FYI, my patch doesn't touch PALETTE_COLOR at all, it work great,
returns correct RGB image or gray image, depends on reader's image type
`typedef itk::ImageFileReader<T> ReaderType;`

issakomi:
@malaterre 
I am not sure it is better to use gdcm::ImageChangePhotometricInterpretation,
my version doesn't need to allocate double memory, doesn't allocate memory
at all, like in change filter, s below

```
  unsigned long len = image.GetBufferLength();
  unsigned char *p = new unsigned char[len];
  image.GetBuffer( (char*)p );
```

Anyway i have tried to test gdcm::ImageChangePhotometricInterpretation, could
not get it working, may be i have done something wrong
`itk::ERROR: GDCMImageIO(0x557e44a05b80): Failed to get the buffer!`

```
   gdcm::PhotometricInterpretation pi = image.GetPhotometricInterpretation();
   if ( pi == gdcm::PhotometricInterpretation::PALETTE_COLOR )
     {
     gdcm::ImageApplyLookupTable ialut;
     ialut.SetInput(image);
     ialut.Apply();
     image = ialut.GetOutput();
     len *= 3;
     }
+  else if ( true && // option on/off
+            ( pi == gdcm::PhotometricInterpretation::YBR_FULL_422 ||
+              pi == gdcm::PhotometricInterpretation::YBR_FULL ) )
+    {
+    gdcm::ImageChangePhotometricInterpretation iybr;
+    iybr.SetInput(image);
+    iybr.SetPhotometricInterpretation(gdcm::PhotometricInterpretation::RGB);
+    iybr.Change();
+    image = iybr.GetOutput();
+    }

   if ( !image.GetBuffer( (char*)pointer ) )
     {
     itkExceptionMacro(<< "Failed to get the buffer!");
     }
```

blowekamp:
The automatic conversion in ITK from multi-component to gray scale does not take into consideration any photometric interpretation of the multicomponent image. Please see the code here:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/IO/ImageBase/include/itkConvertPixelBuffer.hxx#L62-L174

In general there is no unified photometric meta-data associated with an Image after it has been read. Ideally, each ImageIO, would attach some meta-data describing the pixel's interpretation, then there would be filter(s) aware of how to convert the buffer.

issakomi:
@blowekamp 

> The automatic conversion in ITK from multi-component to gray scale
thanks, i mean color view
```
typedef itk::RGBPixel<unsigned char>  RGBPixelUCType;
typedef itk::Image<RGBPixelUCType, 3> RGBImageTypeUC;
typedef itk::ImageFileReader<RGBImageTypeUC> ReaderType;

```

> In general there is no unified photometric meta-data associated with an Image 
i think **itk::RGBPixel** defines with the name that it is RGB pixel, not anything else, s. code above

blowekamp:
What `[PixelType](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/IO/ImageBase/include/itkImageIOBase.h#L105-L107)` does GDCMImageIO report this file as? I don't think the correct type is even available in the enum. Perhaps there needs to be something like "COLOR_PHOTOMETRIC" added.

The ImageIO does not know what the itk::Image type its filling. Perhaps logic would be better applied in a similar fashion as the convert buffer?
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/IO/ImageBase/include/itkImageFileReader.hxx#L398-L423

issakomi:
@blowekamp 
> Perhaps logic would be better applied in a similar fashion as the convert buffer?

May be, i don't know, my [patch](https://github.com/InsightSoftwareConsortium/ITK/compare/master...issakomi:ybr_to_rgb) is single function and could be disabled, very simple, doesn't break things. I don't insits, have solved the problem for a long time for. Also if i save raw binary stream from DICOM Jpeg Lossy compressed file and rename it to
*.jpg - ITK opens the file (1st frame) in normal colors, not greenish-blue, BTW. Don't understand who can need that greenish-blue thingy.

[DICOM file](https://drive.google.com/file/d/1YXaIbyRAY8ZsZvvshS2vaQWCofThozs2/view?usp=sharing)
[raw binary stream](https://drive.google.com/file/d/16iO-LJpszApkGs6SdvzfWwHbsxjsb2wF/view?usp=sharing)

zivy:
Hi @issakomi 

You are confusing two things here: 
1. Data, a color image in some color space. 
2. Data visualization, display of the image.

Most viewers assume that if you give them an image with three channels these are RGB, this isn't right or wrong it is just an assumption. 

Why you may want the greenish-blue thingy (YCbCr colorspace):
For image analysis we often only work on a luminance channel (e.g. Canny edge detector), this would be the Y channel in your image. For a color image that means you either store it in YCbCr and readily use the Y channel or you store it in RGB and then compute the Y channel before analysis.

Hopefully this clarifies things a bit. 

As I said earlier, the patch looks good to me with the caveat that you add a flag letting the user specify whether the channels are the original or converted ones.

issakomi:
@zivy 
If i am confusing something be so kind and make the patch yourself. Thank you.

thewtex:
> thanks, i mean color view
> 
> ```
> typedef itk::RGBPixel<unsigned char>  RGBPixelUCType;
> typedef itk::Image<RGBPixelUCType, 3> RGBImageUCType;
> typedef itk::ImageFileReader<RGBImageUCType> ReaderType;
> ```
> 
> > In general there is no unified photometric meta-data associated with an Image
> 
> i think **itk::RGBPixel** defines with the name that it is RGB pixel, or at least can be

@issakomi good point. Thanks for contributing the patch.

We want to add the provided test data and the nice fix demonstrated when the issue was opened in a test to verify the fix.

issakomi:
@thewtex  Thank you.
BTW, i have deleted branch after closed PR, [this one,](https://github.com/InsightSoftwareConsortium/ITK/compare/master...issakomi:ybr_to_rgb) the same as in the link in the above post, has requested _on/off_ option to preserve old behavior.

blowekamp:
@thewtex @sjh26 
Please include in the test that the GDCMImageIO reports the pixel type as `IOPixelType::RBG` if the conversion does occur.

If the conversion is disabled, then it's not clear what the `IOPixelType` should be perhaps `FIXEDARRAY` is appropriate, but reporting as `RGB` still would be misleading.

issakomi:
> If the conversion is disabled, then it's not clear what the IOPixelType should be perhaps FIXEDARRAY is appropriate, but reporting as RGB still would be misleading.

BTW, there are also YBR_ICT and YBR_RCT. It is for JPEG2000 (irreversible, reversible). They are
displayed **correctly**, not greenish-blue. Sample color data sets from D. Clunie:
[YBR_ICT data sets](https://drive.google.com/file/d/1A9bXMc6qR6ndKeQ_ywctKCvmrQuys6QK/view?usp=sharing)
[YBR_RCT data sets](https://drive.google.com/file/d/1ahM-pDTX8v44ik9VCGVfk-c9Sb6oLQ8Q/view?usp=sharing)

thewtex:
> Please include in the test a check that the GDCMImageIO reports the pixel type as `IOPixelType::RBG` if the conversion does occur.

`ImageIOBase::GetPixelType` should return the pixel type of image in the file -- it should not be dependent on how it is used later, i.e. if it converted downstream in its use.


> If the conversion is disabled, then it's not clear what the `IOPixelType` should be perhaps `FIXEDARRAY` is appropriate, but reporting as `RGB` still would be misleading.

`IOPixelType` is something that should have attention, but this fix can proceed without it. It is unclear what it should be, maybe `FIXEDARRAY`, maybe not. Ideally it would be `YBR`, but that is not supported.

blowekamp:
> ImageIOBase::GetPixelType should return the pixel type of image in the file -- it should not be dependent on how it is used later, i.e. if it converted downstream in its use.

I have always interpreted this field as: what is is the produced buffer by the ImageIO, NOT what is in the file.  This would include any conversion done but the ImageIO, but not many conversions done by the `ImageFileReader`'s `DefaultConvertPixelTraits`.

The proposed change is consistent with this model, as long as the `IOPixelType` is set to `RGB`. Which is why I request a check for consistency in the test.

issakomi:
BTW, currently GDCMImageIO reports RGB for all of them

```
r@deb:~/tmp/test012/build$ ./test012 dicom_ybr/CLIP0001
GDCMImageIO 
Image dimensions: 3
Pixel type: RGB
Component type: unsigned char (3)
```

[Show Image IO test prog.](https://drive.google.com/file/d/1BhZj8uujkoMyneqwcemRQXe9FdBGEXe1/view?usp=sharing)

Issue #820, opened 2019-05-05
SpatialObject IsInsideInObjectSpace PERF penalty GetTypeName().find


N-Dekker:
### Description

Looking at the performance of `IsInsideInObjectSpace(point, depth=0,  name="")`, I observed a large performance gain when I locally removed the following line, when using the default argument values (so `name=""`):

    if( this->GetTypeName().find( name ) != std::string::npos )

https://github.com/InsightSoftwareConsortium/ITK/blob/v5.0rc02/Modules/Core/SpatialObjects/include/itkImageMaskSpatialObject.hxx#L45

### Impact analysis

For `ImageMaskSpatialObject` it appears that `IsInsideInObjectSpace(point)` will be more than twice as fast, and sometimes even more than 3 X as fast, when the line of code that does the `GetTypeName()` check is removed.

I measured the performance of something like this: 

    for (unsigned i = 0; i < n; ++i)
    {
      typename TImage::PointType point{};
      point[0] = i;
      spatialObject->IsInsideInObjectSpace(point);
    }

### Versions

ITK5 RC2

### Environment

Visual C++ 2017 x64 Release 

### Question

Is there a common use case, calling `IsInsideInObjectSpace(point, depth, name)` with a non-empty `name` argument?

hjmjohnson:
@N-Dekker Could you simply make 2 explicit separate signatures and avoid the default parameter for name removed?

```
IsInsideInObjectSpace(point)
```
```
IsInsideInObjectSpace(point, depth) // Perhaps have depth=0 defaulted and skip the first signature
```

```
IsInsideInObjectSpace(point, depth, name)
{
if( this->GetTypeName().find( name ) != std::string::npos )
{}

return IsInsideInObjectSpace(point, depth)
}
```

This would be backward compatible and have the performance gains.

N-Dekker:
@hjmjohnson Thanks for your suggestion. Surely the PERF issue could be solved by adding an overload without the `name` parameter. However, it isn't entirely as simple as you suggested! When `depth` > 0, the function returns `IsInsideChildrenInObjectSpace(point, depth, name)`, so at that point, the name argument must still be available.

N-Dekker:
In the mean time I did figure out that the performance issue can already be fixed easily, by ensuring that `GetTypeName().find(name)` is _only_ called when the name is non-empty. So by doing:

    if(name.empty() || (this->GetTypeName().find( name ) != std::string::npos) )

[...]

PR #828, opened 2019-05-06
PERF: Overload SpatialObject IsInsideInObjectSpace without name argument

N-Dekker:
Added virtual `IsInsideInObjectSpace(const PointType &)` member function
to `itk::SpatialObject`, as an overload of the original one,
`IsInsideInObjectSpace(const PointType&, unsigned depth, const std::string& name)`.
Removed the `virtual` keyword from the original member function.

Ensured that the 11 derived classes now override the new overload
(without the `name` parameter), instead of the original one. The
original member function now calls the new overload.

A major performance improvement was observed by this commit, especially
because it avoids many expensive `GetTypeName().find(name)` calls.
`IsInsideInObjectSpace` often appears more than twice (and sometimes
even more than 3 times) as fast as before.

Fixes issue #820 ("SpatialObject IsInsideInObjectSpace PERF penalty
GetTypeName().find"), based on a suggestion by Hans @hjmjohnson.

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkArrowSpatialObject.h
+++ b/Modules/Core/SpatialObjects/include/itkArrowSpatialObject.h
@@ -79,8 +79,10 @@ public:
   itkGetConstReferenceMacro(LengthInObjectSpace, double);
 
   /** Returns true if the point is inside the line, false otherwise. */
-  bool IsInsideInObjectSpace(const PointType & point, unsigned int depth=0,
-    const std::string & name="") const override;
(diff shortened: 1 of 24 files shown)


N-Dekker:
Tip of the day: when reviewing this commit on github, consider Diff settings ->  Hide whitespace changes (Just for now)!

The most important changes are in itkSpatialObject.h  and itkSpatialObject.hxx

And yes, it's really a major performance improvement  `:-)`

PR #831, opened 2019-05-06
BUG: Address valgrid defects related to the PointSpatialObject

blowekamp:
The PointSpatialObject keeps a reference to its parent spatial object
for the spatial context. This created additional reference counts for
the parent class not released. A WeakPointer is now used.

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkSpatialObjectPoint.h
+++ b/Modules/Core/SpatialObjects/include/itkSpatialObjectPoint.h
@@ -158,7 +158,10 @@ protected:
   /** Color of the point */
   ColorType m_Color;
 
-  typename SpatialObjectType::Pointer m_SpatialObject;
+  // The SpatialObjectPoint keeps a reference to its owning parent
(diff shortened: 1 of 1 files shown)


aylward:
Lgtm!

Modules/Core/SpatialObjects/include/itkSpatialObjectPoint.h
```diff
@@ -158,7 +158,7 @@ class ITK_TEMPLATE_EXPORT SpatialObjectPoint
   /** Color of the point */
   ColorType m_Color;
 
-  typename SpatialObjectType::Pointer m_SpatialObject;
+  WeakPointer<SpatialObjectType> m_SpatialObject;
```
hjmjohnson: Perhaps add a comment "//The PointSpatialObject keeps a reference to its parent spatial object for its spatial context, so use a WeakPointer here to avoid a memory leak."

fbudin69500:
Looks good to me minus the typos in the commit message (it's->its)

PR #846, opened 2019-05-08
BUG: Set all component of 3D points

blowekamp:
This addresses valgrind defects related to using uninitialized values
related to the 3d component of the point being unset.

### DIFF
--- a/Examples/Filtering/SpatialObjectToImage1.cxx
+++ b/Examples/Filtering/SpatialObjectToImage1.cxx
@@ -188,14 +188,16 @@ int main( int argc, char *argv[] )
   typename TubeType::PointType point;
   typename TubeType::TubePointType tubePoint;
   typename TubeType::TubePointListType tubePointList;
-  point[0] =  size[0] * 0.2 * spacing[0];
-  point[1] =  size[1] * 0.2 * spacing[1];
+  point[0] = size[0] * 0.2 * spacing[0];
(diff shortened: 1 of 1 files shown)


blowekamp:
After this is merged I expect Valgrind to be clean.

dzenanz:
> I expect Valgrind to be clean

Awesome!

Examples/Filtering/SpatialObjectToImage1.cxx
```diff
@@ -204,12 +206,14 @@ int main( int argc, char *argv[] )
...
   point[0] =  size[0] * 0.2 * spacing[0];
   point[1] =  size[1] * 0.8 * spacing[1];
+  point[2] =  size[2] * 0.2 * spacing[2];
   tubePoint.SetPositionInObjectSpace( point );
   tubePoint.SetRadiusInObjectSpace( size[0] * 0.05 * spacing[0] );
   tubePointList.push_back( tubePoint );
 
   point[0] =  size[0] * 0.8 * spacing[0];
   point[1] =  size[1] * 0.8 * spacing[1];
+  point[2] =  size[2] * 0.8 * spacing[1];
```
N-Dekker: @blowekamp Just checking... shouldn't that be `spacing[2]`? (Not sure if it matters, it's just an example, right?)

dzenanz: All of us missed it 😄

PR #847, opened 2019-05-08
Consolidating setting compression level interface to ImageIOBase

blowekamp:
There are two primary characterizations of compression level. 1) the
0-100 version for JPEG-like lossy compression which ranges 0-100 and
2) the loss-less with quality 0-9.

This implementation has the ImageIO set the expected range and enforce
the range by clamping when setting the value.


Modules/IO/ImageBase/include/itkImageIOBase.h
```diff
@@ -205,6 +205,28 @@ class ITKIOImageBase_EXPORT ImageIOBase:public LightProcessObject
...
+  /** \brief Set/Get a compression level hint in range 0-100
+   *
+   * If compression is enabled by UseCompression, then this is the
+   * hint to the actually ImageIO of the requested compression level
+   * range is dependent on the ImageIO.
+   **/
+  itkSetClampMacro(CompressionLevel, int, 1, this->GetMaximumCompressionLevel() );
+  itkGetConstMacro(CompressionLevel, int);
+
+  /** \brief Set/Get the compression algorithm to used
```
dzenanz: to used -> to **be** used

Modules/IO/ImageBase/include/itkImageIOBase.h
```diff
@@ -205,6 +205,28 @@ class ITKIOImageBase_EXPORT ImageIOBase:public LightProcessObject
...
+   *
+   * If compression is enabled by UseCompression, then the string is
+   * used as a suggestion for the compression algorithm to use. If the
+   * string identifier is not recognized, then a warning is produced,
+   * and the default is used.
+   *
+   * \nb Not all ImageIO's support compression, to this all
```
dzenanz: ImageIO's -> ImageIOs

Sentence -> meaning

Modules/IO/LSM/src/itkLSMImageIO.cxx
```diff
@@ -357,7 +357,7 @@ void LSMImageIO::Write(const void *buffer)
 
     if ( compression == COMPRESSION_JPEG )
       {
-      TIFFSetField(tif, TIFFTAG_JPEGQUALITY, 75); // Parameter
+      TIFFSetField(tif, TIFFTAG_JPEGQUALITY, this->GetCompressionLevel()); // Parameter
```
dzenanz: This will change default JPEG quality from 75 to 30, which is probably not desired.

blowekamp:
> Have you looked into exposing different compressors for TIFF? Packbits, LZW, RLE? If easy, it would be a good example of use for compressor API.

Thanks for the look.

The proposed "string" name interface for a compression algos will be a little tedious ( error prone ) to implement, but not "hard".

TIFF, GDCM, and NRRD have some different compression method to explore this interface with.

blowekamp:
@malaterre @gdevenyi Asking the supporters of ITK's IOs to review the extension of the ImageIOBase to generically support compression options.

Modules/IO/JPEG/include/itkJPEGImageIO.h
```diff
@@ -28,6 +28,9 @@ namespace itk
  *
  * \brief ImageIO object for reading and writing JPEG images
  *
+ * Compression is support with only the default compressor. The
```
dzenanz: support -> support**ed**

Modules/IO/MINC/include/itkMINCImageIO.h
```diff
@@ -55,6 +55,9 @@ struct ITKIOMINC_HIDDEN MINCImageIOPImpl;
  * reads the regularly sampled dimensions as I am not sure how to deal
  * with "iregularly sampled" dimensions yet!
  *
+ * Compression is support with only the default compressor. The
```
dzenanz: support -> support**ed**

Modules/IO/NRRD/src/itkNrrdImageIO.cxx
```diff
@@ -1102,6 +1104,8 @@ void NrrdImageIO::Write(const void *buffer)
        && this->m_NrrdCompressionEncoding->available() )
     {
     nio->encoding = this->m_NrrdCompressionEncoding;
+    nio->zlibLevel = this->GetCompressionLevel();
```
dzenanz: It might be this simple. Have you tried writing some images with different compression levels to see whether this gets propagated all the way?

Modules/IO/TIFF/test/itkTIFFImageIOCompressionTest.cxx
```diff
@@ -53,24 +79,43 @@ int itkTIFFImageIOCompressionTestHelper( int, char * argv[], int JPEGQuality )
 
   if( compression == "Deflate" )
     {
+
     io->SetCompressionToDeflate();
+
+    TEST_EXPECT_EQUAL(compression, io->GetCompressor());
+    TEST_EXPECT_TRUE(!io->GetUseCompression());
```
dzenanz: Is negation here intentional?

Modules/IO/GDCM/include/itkGDCMImageIO.h
```diff
@@ -53,6 +53,9 @@ namespace itk
  *
  * GDCM build, instead of the one included within ITK itself.
  *
+ * The compressors supported include "JPEG2000" (default), and
```
malaterre: At least in my fork, I can do JPEG 2000/JPEG/JPEG-LS/RLE. So I need to do my homework here, first.

Modules/IO/GDCM/src/itkGDCMImageIO.cxx
```diff
@@ -1444,6 +1444,24 @@ bool GDCMImageIO::GetLabelFromTag(const std::string & tag,
   return false;
 }
 
+void GDCMImageIO::InternalSetCompressor(const std::string &_compressor )
+{
+
+  if (_compressor == "" ||
```
malaterre: As said above this is missing other important compressor (RLE & JPEG-LS)

Modules/IO/GDCM/include/itkGDCMImageIO.h
```diff
@@ -53,6 +53,9 @@ namespace itk
...
+ * "JPEG". The compression level parameter is not supported.
```
malaterre: Technicaly GDCM does not do much here, simply forward parameter. So saying out loud 'compression level parameter is not supported' feels a bit wrong, maybe 'compression level parameter is not implemented' feels more accurate.

Modules/IO/JPEG/include/itkJPEGImageIO.h
```diff
@@ -49,8 +52,8 @@ class ITKIOJPEG_EXPORT JPEGImageIO:public ImageIOBase
   itkTypeMacro(JPEGImageIO, ImageIOBase);
 
   /** Set/Get the level of quality for the output images. */
-  itkSetMacro(Quality, int);
-  itkGetConstMacro(Quality, int);
+  virtual void SetQuality(int _JPEGQuality) { this->SetCompressionLevel(_JPEGQuality); }
```
malaterre: That would need to extra documentation. As a user I would find it very hard to understand that a compression ratio is now an image quality (from the top of my head jpeg quality here is a PSNR right?).

Modules/IO/MINC/include/itkMINCImageIO.h
```diff
@@ -55,6 +55,9 @@ struct ITKIOMINC_HIDDEN MINCImageIOPImpl;
  * reads the regularly sampled dimensions as I am not sure how to deal
  * with "iregularly sampled" dimensions yet!
  *
+ * Compression is supported with only the default compressor. The
+ * compression level option is supported in the range 0-9.
```
malaterre: Also right here, you are not exposing (implementation level) that max level is actually '9'. A program would have a hard time exposing this max value (slide bar) to the user if need be (and instead would display '100').

malaterre:
Overall I am not very convinced this will add any value to the ITK toolkit. In particular this may even introduce more doubt to the user. Confusing lossless compression ratio and lossy compression quality feels awkward. Could someone kind add `gimp` screenshot comparing the UI for PNG (lossless) vs JPEG (lossy) to see what other people are exposing ?

Modules/IO/TIFF/test/itkTIFFImageIOCompressionTest.cxx
blowekamp: nope, good find!

Modules/IO/NRRD/src/itkNrrdImageIO.cxx
blowekamp: I'll check...

Modules/IO/MINC/include/itkMINCImageIO.h
blowekamp: The `ImageIOBase` class has a `GetMaximumCompressionLevel` method which would give the maximum value for a slider.

Modules/IO/GDCM/include/itkGDCMImageIO.h
blowekamp: There are quite a few definitions related to JPEG [here](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/ThirdParty/GDCM/src/gdcm/Source/DataDictionary/gdcmUIDs.h#L349-L376), it's not clear which are supported compressors. We can save exposing additional one for another patch. This documentation is for the features supported by the `ImageIO` not the library. Perhaps it should indicate that these are the "compressors" supported for writing.

Modules/IO/GDCM/src/itkGDCMImageIO.cxx
blowekamp: future work.

Modules/IO/GDCM/include/itkGDCMImageIO.h
blowekamp: Currently, this parameter is not being forwarded  and there is not compression quality forward hence the parameter is not supported.

maekclena:
@malaterre see https://docs.gimp.org/2.10/en/gimp-images-out.html

blowekamp:
Thanks @maekclena for providing that like. The interface provides for both  lossless "compression level" (0-9) and JPEG "quality" (0-100) is a slider bar with the high numbers to the right. This is similar to API provided, and one could use this proposed API to implement such a slider with the proper range.

There are no better alternative proposed for the interface. As this interface provide significant value as access to more current and futures compression features it is well worth merging.

dzenanz:
If no one proposes and alternative API design for this by, let's say tomorrow morning, I vote this be merged. Any opposition?

PR #850, opened 2019-05-09
BUG: Fix reverse iterator increment return type

### LINKED ISSUE #775: Incorrect return value FixedArray::ReverseIterator::operator++()
### Description

The return value of `operator++()` and `operator++(int)` of both `FixedArray::ReverseIterator` and `FixedArray::ConstReverseIterator` appears incorrect. Both operators appear to return an iterator that points to the wrong element.

### Steps to Reproduce

The following GTest expects the same behavior for the reverse iterators of `FixedArray` and `std::array`, but it fails:

    TEST(FixedArray, IncrementReverseIterator)
    {
      const std::array<int, 3> stdArray{ {1, 2, 3} };
      const itk::FixedArray<int, 3> fixedArray{ stdArray };
      auto fixedArrayIterator = fixedArray.rBegin();
      auto stdArrayIterator = stdArray.rbegin();

      EXPECT_EQ(*++fixedArrayIterator, *++stdArrayIterator);
      EXPECT_EQ(*fixedArrayIterator++, *stdArrayIterator++);
    }

Output (twice):
>  error: Expected equality of these values:
>   *++fixedArrayIterator
>     Which is: 3
>   *++stdArrayIterator
>     Which is: 2

### Reproducibility

100% reproducible

### Versions

Apparently, this bug was there already in 2002, but it is still there in ITK5 RC1, and in the current revision from git master branch: c351787a261641a8b1c73e03f3dae37a40be967a

### Additional Information

Instead of fixing this bug, I would prefer if the old FixedArray reverse iterators would be deprecated (and eventually removed), in favor of the new reverse iterators proposed with pull request #771 ("ENH: Add std style reverse iterators to FixedArray") and merged with 07f766079ffabbdc254e30c838b62d255742cc98

maekclena:
Fixes #775.

### DIFF
--- a/Modules/Core/Common/include/itkFixedArray.h
+++ b/Modules/Core/Common/include/itkFixedArray.h
@@ -80,10 +80,10 @@ public:
   {
   public:
     explicit ReverseIterator(Iterator i):m_Iterator(i) {}
-    Iterator operator++()        { return --m_Iterator; }
-    Iterator operator++(int)     { return m_Iterator--; }
-    Iterator operator--()        { return ++m_Iterator; }
-    Iterator operator--(int)     { return m_Iterator++; }
+    ReverseIterator operator++()        { return ReverseIterator( --m_Iterator ); }
+    ReverseIterator operator++(int)     { return ReverseIterator( m_Iterator-- ); }
+    ReverseIterator operator--()        { return ReverseIterator( ++m_Iterator ); }
+    ReverseIterator operator--(int)     { return ReverseIterator( m_Iterator++ ); }
     Iterator operator->() const { return ( m_Iterator - 1 ); }
     ValueType & operator*() const { return *( m_Iterator - 1 ); }
     bool operator!=(const ReverseIterator & rit) const { return m_Iterator != rit.m_Iterator; }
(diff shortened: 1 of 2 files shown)


blowekamp:
Please add a test to verify what you have written has correct behavior. The issue reported has a test which can be expanded.

phcerdan:
> Still I hope that FixedArray::ReverseIterator can eventually be removed from ITK.  @phcerdan Wasn't your idea to eventually remove the old ReverseIterator, when you suggested me to write a new one, for FixedArray, at #757 (comment) ?

No, it wasn't my idea and the first time I heard that removing ReverseIterator is desired. In #757 you were adding compatibility of FixedArray to the `std` algorithms adding `begin()` and `end()` (which was great!), and I asked to be more feature complete adding `rbegin` and `rend`.

What is really desired is to fix bugs, so +100 to this PR and @maekclena!

N-Dekker:
@dzenanz I still had a technical question about this PR: https://github.com/InsightSoftwareConsortium/ITK/pull/850#pullrequestreview-236183925 So why did you already merge?

N-Dekker:
@maekclena For your information, this example code still fails now, even with your fix being merged:

    auto newReverseIterator = fixedArray.rbegin();
    auto oldReverseIterator = fixedArray.rBegin();

    // Move two steps in reverse direction!
    ++++oldReverseIterator;
    ++++newReverseIterator;

    EXPECT_EQ(*oldReverseIterator, *newReverseIterator);

dzenanz:
@N-Dekker this PR did what its title says "Fix ... return type". All the increment/decrement operators return `ReverseIterator`s so they cannot return `Iterator`s. And your comment `...` was a bit confusing, I thought you figured out the question from previous comment. Hence I merged it. Sorry for misunderstanding.

You could expand and reopen the issue, which might get addressed in another PR. Which you might attempt yourself?

N-Dekker:
@dzenanz  First of all, I do agree that the PR as it is merged now does already improve the behavior of the old reverse iterator. But then again, it still does not seem interesting to me to keep maintaining the old `FixedArray::ReverseIterator`. I think users should just switch to the new one!

Advantages of the new `FixedArray::reverse_iterator`:
1. Can be passed directly to `std` algorithms, for example `std::sort(fixedArray.rbegin(), fixedArray.rend())`, which won't compile for the old `rBegin()` and `rEnd()`
2. Supports random access iteration (while the old `ReverseIterator` _only_ supports bidirectional iteration)
3. Has a `base()` function to return the original (forward) iterator
4. Takes less maintenance from ITKdevelopers, as it just uses the `std::reverse_iterator` implementation.

The old `FixedArray::ReverseIterator` does not seem to offer anything extra, does it?

dzenanz:
Can you make a PR which deprecates it?

N-Dekker:
>Can you make a PR which deprecates it?

I would like to, thanks @dzenanz, but isn't it better now to postpone such deprecation until after the release of ITK5?

dzenanz:
If we are going to deprecate it soon, before the release is probably better than after.

PR #855, opened 2019-05-10
Add streaming base classes and streaming statistics

blowekamp:


blowekamp:
Changing the base class for the `StatisticsFilters` to the `ImageSink` with streaming support when very smoothly. More testing and documentation passes are still needed.

Please review the style interface for the `StreamingProcessObject` and `ImageSink` classes.

A backwards incompatibility change introduced is that the `LabelStatisticsImageFilter` and `StatisticsImageFilter` no longer produce an image as output since they are no longer derived from `ImageToImageFilter`.

Modules/Core/Common/include/itkImageSink.hxx
```diff
@@ -0,0 +1,176 @@
...
+  this->GetMultiThreader()->ParallelizeArray(
+    0,
+    total,
+    [this, total](SizeValueType threadId)
+    {
+        InputImageRegionType splitRegion = this->m_CurrentInputRegion;
+        this->GetRegionSplitter()->GetSplit( threadId, total, splitRegion );
+        this->ThreadedStreamedGenerateData(splitRegion );
+    },
+    nullptr); // \Modules\Core\Common\include\itkProgressTransformer.h would need to be used to properly report progress here
```
dzenanz: This comment was more for you @blowekamp. Adding `ProgressTransformer` should be only few extra lines of code, and would increase granularity of progress reporting.

blowekamp: See line ~85 of `StreamingProceessObject` where the progress is reported just at the per streaming chunk level. Do you have some example code which would be similar to this situation with 2 levels of progress?

dzenanz: It should be similar to this:
https://github.com/InsightSoftwareConsortium/ITK/blob/91ab86f63eb5fc9799bb25fa49e76adf0d1b5251/Modules/Filtering/DisplacementField/include/itkInvertDisplacementFieldImageFilter.hxx#L136-L137

dzenanz: But then the other progress reporting is redundant.

blowekamp: Thanks for the link.

Documentation/ITK5MigrationGuide.md
```diff
@@ -402,6 +402,14 @@ b40f74e07d74614c75be4aceac63b87e80e589d1 on 2018-11-14.
 
 `itk::Barrier` has been moved to `ITKDeprecated` module.
 
+`itk::ImageTransformer` has been moved to `ITKDeprecated` module. The new `itk::ImageSink` filter can be used in it's place.
```
blowekamp: its

Modules/Core/Common/include/itkImageSink.hxx
```diff
@@ -0,0 +1,287 @@
...
+  const float oldProgress = float(inputRequestedRegionNumber)/(total);
+  const float newProgress = float(inputRequestedRegionNumber+1)/(total);
+  ProgressTransformer pt( oldProgress, newProgress, this );
+
+
+  this->GetMultiThreader()->template ParallelizeImageRegion<InputImageDimension>(
+    this->m_CurrentInputRegion,
+    [this](const InputImageRegionType & inputRegionForThread)
+    { this->ThreadedStreamedGenerateData(inputRegionForThread); },
+    pt.GetProcessObject());
```
blowekamp: @dzenanz This was updated further to use the `ParallelizeImageRegion` method. The internal splitter is just used for the streamed region splitting now.

Modules/Filtering/ImageStatistics/test/itkStatisticsImageFilterTest.cxx
```diff
@@ -24,14 +24,27 @@
...
+  std::cout << "itkStatisticsImageFilterTest  [numberOfStreamDivisions]" << std::endl;
 
   int status = 0;
+
+
+  unsigned int numberOfStreamDivisions = 1;
+
+  if (argc > 2)
+    {
+    numberOfStreamDivisions = std::max( atoi( argv[1] ), 1 );
```
dzenanz: use `std::stoi`

Modules/Filtering/ImageStatistics/include/itkStatisticsImageFilter.h
```diff
@@ -163,27 +133,30 @@ class ITK_TEMPLATE_EXPORT StatisticsImageFilter:
...
+  itkSetDecoratedOutputMacro(Mean, RealType);
+  itkSetDecoratedOutputMacro(Sigma, RealType);
+  itkSetDecoratedOutputMacro(Variance, RealType);
+  itkSetDecoratedOutputMacro(Sum, RealType);
+  itkSetDecoratedOutputMacro(SumOfSquares, RealType);
+
 private:
-  CompensatedSummation<RealType> m_ThreadSum;
-  CompensatedSummation<RealType> m_SumOfSquares;
+  CompensatedSummation<RealType> m_ThreadSum{1};
```
dzenanz: shouldn't these two be initialized to 0?

Modules/Numerics/Statistics/include/itkImageToHistogramFilter.h
```diff
@@ -117,21 +118,29 @@ class ITK_TEMPLATE_EXPORT ImageToHistogramFilter:public ImageTransformer<TImage>
...
+  void BeforeStreamedGenerateData() override;
+  void AfterStreamedGenerateData() override;
 
   /** Method that construct the outputs */
   using DataObjectPointerArraySizeType = ProcessObject::DataObjectPointerArraySizeType;
   using Superclass::MakeOutput;
   DataObject::Pointer  MakeOutput(DataObjectPointerArraySizeType) override;
 
-  virtual void ThreadedComputeHistogram(const RegionType &);
+  /* Override method to disable streaming when the minimal and
```
dzenanz: minimal -> minimum

dzenanz:
Looking good so far.

Modules/Core/Common/include/itkImageSink.h
```diff
@@ -0,0 +1,189 @@
...
+
+/** \class ImageSink
+ *
+ * ImageSink is the base class for process objects which consume image
+ * data. This class defaults to having at least one input of the
+ * templated image type. The framework enables derived algorithms to
+ * stream the input image as it's being consumed by the algorithm.
+ *
+ * The framework provides multi-threading of the streamed image regions.
+ * The input image's pipeline is updated multiple time with the
```
zivy: time -> times

Modules/Core/Common/include/itkImageSink.h
```diff
@@ -0,0 +1,189 @@
...
+ * streaming requested regions, then the  fulfilled requested region
+ * are split again for multi-threading.
+ *
```
zivy: * By default, the NumberOfStreamDivisions is just 1 (no streaming),
 * derived implementations must change the access specification for
 * this method to be public to expose the streaming feature.

Modules/Filtering/ImageStatistics/include/itkStatisticsImageFilter.h
blowekamp: The value doesn't matter much as it get's properly initialized int BeforeStreamedGenerateData. These were the values in the constructor.

Modules/Core/Common/include/itkImageSink.hxx
```diff
@@ -0,0 +1,287 @@
...
+  itkDebugMacro( "Generating " << inputRequestedRegionNumber << " chunk as " <<  m_CurrentInputRegion );
+
+
+  for (auto &inputName: this->GetInputNames())
+    {
+    if ( this->ProcessObject::GetInput(inputName) )
+      {
+      // Check whether the input is an image of the appropriate
+      // dimension (use ProcessObject's version of the GetInput()
+      // method sinke it returns the input as a pointer to a
```
zivy: sinke -> since

Modules/Core/Common/include/itkImageSink.hxx
```diff
@@ -0,0 +1,287 @@
...
+    // images, and not an image and a constant.
+    if( inputPtrN )
+      {
+      // check that the image occupy the same physical space, and that
+      // each index is at the same physical location
+
+      // tolerance for origin and spacing depends on the size of pixel
+      // tolerance for directions a fraction of the unit cube.
+      const SpacePrecisionType coordinateTol
+        = std::abs(this->m_CoordinateTolerance * inputPtr1->GetSpacing()[0]); // use first dimension spacing
```
zivy: why is there a need for abs here? I expected tolerance and spacing to be positive. The tolerance is double but I expect that when it is set, ITK either forces abs or throws an exception if less than zero.

blowekamp:
I have completed refactoring the classes. I'll write some detailed/corner cases tomorrow to wrap things up.

@romangrothausmann Please check out this PR and test the classes. Thank you!

Modules/Core/Common/include/itkImageSink.hxx
```diff
@@ -0,0 +1,287 @@
...
+
+      if ( !inputPtr1->GetOrigin().GetVnlVector().is_equal(inputPtrN->GetOrigin().GetVnlVector(), coordinateTol) ||
```
zivy: Not sure about this, but I thought there was an attempt to avoid usage of vnl structures so that at some point in the future we can move away from vnl. Is there some other way of doing the comparisons that isn't vnl dependent? I may be wrong about this.

Modules/Core/Common/include/itkStreamingProcessObject.h
```diff
@@ -0,0 +1,100 @@
...
+{
+
+/** \class Base class interface to process data on multiple requested input chunks
+ *
+ * Streaming allows the data to be split into chunks and processed
+ * separately. The StreamingProcessObject class extends functionally
+ * to execute the primary input's pipeline multiple times over
+ * different requested regions. After each requested region is
+ * generated by the upstream pipeline the StreamedGenerateData method
+ * is call.
```
zivy: call -> called

romangrothausmann:
Many thanks @blowekamp, this is a very great contribution! Your former module for ITK4 has helped us a lot when we were in the need to get statistics of very large datasets and we worried about its missing compatibility with the transition to ITK-5. With this PR this would be solved and its future maintenance ensured. In addition it is great that this PR also includes the ideas of https://github.com/blowekamp/itkStreamingSinc/pull/3.

Modules/Filtering/ImageStatistics/include/itkStatisticsImageFilter.hxx
```diff
@@ -272,19 +97,20 @@ StatisticsImageFilter< TInputImage >
...
 
   // Set the outputs
-  this->GetMinimumOutput()->Set(minimum);
-  this->GetMaximumOutput()->Set(maximum);
-  this->GetMeanOutput()->Set(mean);
-  this->GetSigmaOutput()->Set(sigma);
-  this->GetVarianceOutput()->Set(variance);
-  this->GetSumOutput()->Set(sum);
-  this->GetSumOfSquaresOutput()->Set(sumOfSquares);
+  std::cout << "Setting Sigma: " << sigma << std::endl;
```
zivy: remove std::cout line

Modules/Filtering/ImageStatistics/include/itkStatisticsImageFilter.hxx
```diff
@@ -28,224 +28,47 @@ namespace itk
...
-}
-
-template< typename TInputImage >
-void
-StatisticsImageFilter< TInputImage >
-::GenerateInputRequestedRegion()
-{
-  Superclass::GenerateInputRequestedRegion();
-  if ( this->GetInput() )
+  if ( name == "Mean" ||
```
zivy: This is ugly, but I guess this is C++, Python is nicer for this. I thought about using a const string results[] = {"Mean", "Sigma", "Variance", "Sum", "SumOfSquares"} and std::find, but probably doesn't look more elegant.

Modules/Core/Common/include/itkImageSink.hxx
blowekamp: This code is just a dup of what's in ImageToImage. It improves robustness with no downside.

Modules/Core/Common/include/itkImageSink.hxx
blowekamp: This is the best method available. Same as imagetoimage.

Modules/Core/Common/include/itkImageToImageFilter.h
```diff
@@ -173,8 +173,8 @@ class ITK_TEMPLATE_EXPORT ImageToImageFilter:public ImageSource< TOutputImage >,
 
   /** get/set the Coordinate tolerance
    *  This tolerance is used when comparing the space defined
-   *  by the input images.  With ITK 4.x a requirement has
-   *  been added that both input images be congruent in space.
+   *  by the input images.  ITK has a requirement that multiple input
+   *  images be congruent in space by default.
    */
```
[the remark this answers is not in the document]
blowekamp: out of scope

Modules/Core/Common/src/itkStreamingProcessObject.cxx
```diff
@@ -0,0 +1,264 @@
...
+    {
+    if (this->GetOutput(outputName))
+      {
+      this->GetOutput(outputName)->DataHasBeenGenerated();
+      }
+    }
+
+  /* DO NOT Restore the state of any input ReleaseDataFlags
+   */
+  //this->RestoreInputReleaseDataFlags();
```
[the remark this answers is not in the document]
blowekamp: This `StreamingProcessObject::UpdateOutputData` is a dup of `StreamingImageFilter::UpdateData` which is very similar to `ProcessObject::UpdateOutputData`. Frankly I don't know why `RestoreInputReleaseDataFlags` should not be called here. IMHO keeping the line commented out aids in comparing the difference versions.

Testing/Data/Baseline/Filtering/ResampleImageFilter9TestPixelCentered.4.png.sha512
```diff
@@ -0,0 +1 @@
+a44632067144ed3d62e538615b1b3503162b9534879abe62100495cc96c69829a5b654a7bb9dce172d04478c5166d1d7d3ce4203882a539072d1047359c8db1f
```
blowekamp: extra file

blowekamp:
This patch should be ready to go. I'm just refreshing the CI after one VNL test failure due to time profiling.

romangrothausmann:
Trying to test this with my ITK-CLIs and am hitting the backwards incompatibility that the StatisticsImageFilter no longer produce an image as output. What would be the recommended way to adjust the code accordingly? Basically I would replace e.g.
```
stat->SetInput(extractor->GetOutput());
adder->SetInput1(stat->GetOutput());
adder->Update();
std::cerr << stat->GetMaximum();
```
by 
```
stat->SetInput(extractor->GetOutput());
adder->SetInput1(extractor->GetOutput());
adder->Update();
stat->Update();
std::cerr << stat->GetMaximum();
```
Is there a more elegant way to trigger stat to update if adder is updated, similar to how it was possible before?

blowekamp:
Great, thank you for testing!

I take it there is no streaming involved here. All you should have to do is simple call `stat->Update()`; seems elegant to me. :sunglasses:

romangrothausmann:
Thanks for the feedback. No, no streaming there. I'll just  use the additional `stat->Update();`
I am hitting this though with `Module_ParabolicMorphology=ON`:
```
/opt/itk/include/ITK-5.0/itkParabolicOpenCloseSafeBorderImageFilter.hxx:48:48: error: no matching function for call to 'itk::StatisticsImageFilter<itk::Image<unsigned char, 3u> >::GetOutput()'
     typename TInputImage::SpacingType spcing = m_StatsFilt->GetOutput()->GetSpacing();
```
Currently trying to find where the sources for that module reside.

blowekamp:
Here is a PR to improve that filter: InsightSoftwareConsortium/ITKParabolicMorphology#26

romangrothausmann:
@blowekamp I'm getting warnings like:
```
WARNING: In /opt/itk/include/ITK-5.0/itkImageSink.hxx, line 88
ImageToHistogramFilter (0x2955840): Unable to convert input "HistogramBinMaximum" to type N3itk5ImageIfLj3EEE
```
when using ImageToHistogramFilter of this PR with my [`hist`](https://github.com/romangrothausmann/ITK-CLIs/commit/8ebfd7d3d8e1196e4b8464c24cda4de39b566401#diff-cfcd3f0f93f9cfa7ee7679c3f76c7753) see e.g. the log here for SDI:
https://gitlab.com/romangrothausmann/ITK-CLIs/-/jobs/214553760
and also if not using SDI:
https://gitlab.com/romangrothausmann/ITK-CLIs/-/jobs/214553759
It is unclear to my why and what to do about them. The result is still the same though.

hjmjohnson:
@blowekamp  I am also getting new warnings for every run of my application:

```
WARNING: In /Shared/johnsonhj/Binaries/20190513/Linux/NEP-14/ITKv5/Modules/Core/Common/include/itkImageSink.hxx, line 88
LabelStatisticsImageFilter (0x278c2d0): Unable to convert input "LabelInput" to type N3itk5ImageIdLj3EEE

WARNING: In /Shared/johnsonhj/Binaries/20190513/Linux/NEP-14/ITKv5/Modules/Core/Common/include/itkImageSink.hxx, line 88
ImageToHistogramFilter (0x279bd70): Unable to convert input "AutoMinimumMaximum" to type N3itk5ImageIsLj3EEE

WARNING: In /Shared/johnsonhj/Binaries/20190513/Linux/NEP-14/ITKv5/Modules/Core/Common/include/itkImageSink.hxx, line 88
ImageToHistogramFilter (0x279bd70): Unable to convert input "HistogramSize" to type N3itk5ImageIsLj3EEE

WARNING: In /Shared/johnsonhj/Binaries/20190513/Linux/NEP-14/ITKv5/Modules/Core/Common/include/itkImageSink.hxx, line 88
ImageToHistogramFilter (0x279bd70): Unable to convert input "MarginalScale" to type N3itk5ImageIsLj3EEE

```

PR #786, opened 2019-04-29
ENH: Deprecate some vector specializations

maekclena:
Addresses #725.

Deprecates the following classes:
- `VectorAnisotropicDiffusionFunction`
- `VectorCastImageFilter`
- `VectorCentralDifferenceImageFunction`
- ~~`VectorConfidenceConnectedImageFilter`~~
- ~~`VectorConnectedComponentImageFilter`~~
- `VectorCurvatureAnisotropicDiffusionImageFilter`
- `VectorCurvatureNDAnisotropicDiffusionFunction`
- `VectorExpandImageFilter`
- `VectorGradientAnisotropicDiffusionImageFilter`
- `VectorGradientMagnitudeImageFilter`
- `VectorGradientNDAnisotropicDiffusionFunction`
- ~~`VectorMeanImageFunction`~~
- `VectorNearestNeighborInterpolateImageFunction`
- `VectorResampleImageFilter`
- `VectorThresholdSegmentationLevelSetFunction`
- `VectorThresholdSegmentationLevelSetImageFilter`

Some classes remain because replacing them by their generic counterparts causes compilation errors (I still replaced as many references as possible):
- `VectorImageToImageMetricTraitsv4`
- `VectorLinearInterpolateImageFunction`
- `VectorInterpolateImageFunction`
- `VectorNeighborhoodInnerProduct`
- `VectorImageNeighborhoodAccessorFunctor`
- `VectorRescaleIntensityImageFilter`
- `VectorNeighborhoodOperatorImageFilter`

PR #856, opened 2019-05-10
Depricate vector resample image filter

blowekamp:
Pulling out some parts of #786


blowekamp:
Note: I am getting a local test values with `ResampleImageFilter9TestPixelCentered`. This test uses the nearest neighbor interpolator. This is like related to these diverging improvements done to the [`ResampleImageFilter`](https://github.com/InsightSoftwareConsortium/ITK/commit/9c448f1b9446e70ee661cd585d5ac7098eeeb5d4#diff-a446710011999e954fe6b67b68d64beb). This divergence is more reason to remove this duplication.

I'll update the baseline after I see consistent results of the change in the CI.

blowekamp:
For the nearest neighbor interpolation, 2/4 tests agree with the baseline. I will be introducing an alternate baseline image.

PR #860, opened 2019-05-12
STYLE: Replace std::swap_ranges by C++11 std::swap of m_InternalArray

N-Dekker:
Instead of calling `std::swap_ranges`, it appears preferable to call
the dedicated C++11 `std::swap` overload for arrays, when swapping the
elements of one `m_InternalArray` and another.

### DIFF
--- a/Modules/Core/Common/include/itkFixedArray.h
+++ b/Modules/Core/Common/include/itkFixedArray.h
@@ -334,7 +334,7 @@ public:
 
   void swap(FixedArray &other)
     {
-      std::swap_ranges(this->Begin(), this->End(), other.Begin());
+      std::swap(m_InternalArray, other.m_InternalArray);
     }
 
 private:
--- a/Modules/Core/Common/include/itkIndex.h
+++ b/Modules/Core/Common/include/itkIndex.h
@@ -324,7 +324,7 @@ public:
 
   void swap(Index & other)
   {
-    std::swap_ranges(begin(), end(), other.begin() );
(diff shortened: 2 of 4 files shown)


N-Dekker:
The dedicated C++11 `std::swap` overload for arrays is actually based on an accepted proposal of mine, for the C++ Standard: [LWG issue 809, std::swap should be overloaded for array types
](https://cplusplus.github.io/LWG/issue809) :blush:

thewtex:
> The dedicated C++11 `std::swap` overload for arrays is actually based on an accepted proposal of mine, for the C++ Standard: [LWG issue 809, std::swap should be overloaded for array types ](https://cplusplus.github.io/LWG/issue809)

Awesome!

N-Dekker:
Thanks very much for accepting this pull request! I'm very glad that ITK now uses the `std::swap` overload that I originally proposed to the Standard Library! 

Two more thought:
1. Ideally, `swap` functions should have a `noexcept` specifier. I did not add them here, because it might slow down the process, and I considered it a separate change (One Commit One Change).
2. With C++03, it was very important to add fast non-throwing `swap` functions to user-defined classes. However, with C++11, it is not so important anymore! C++11 `std::swap` already makes use of fast non-throwing move-semantics, whenever provided, by the user-defined type. 

So eventually I think it would be OK to just remove those ITK-specific `swap` functions, and rely on `std::swap` to do the right thing!

thewtex:
Awesome, thanks for the information @N-Dekker !

PR #861, opened 2019-05-12
COMP: ImageMaskSpatialObjectGTest add braces for initialization

thewtex:


thewtex:
Thanks for the review @N-Dekker !

Just trying to get a clean build on my old work horse :horse:

PR #865, opened 2019-05-14
STYLE: Replace unsigned short by unsigned int as FixedArray index type

N-Dekker:
Replaced `unsigned short` by `unsigned int` as type of `index` parameter,
for both `GetElement(index)` and `SetElement(index, value)` member
functions of `FixedArray`. This improved consistency, as `unsigned int`
is being used throughout the entire definition of `FixedArray`.

Style improvements in the implementation of `FixedArray`:
* Fixed a typo in the documentation.
* Replaced `while` loops by range-based for-loops.
* Replaced `std::copy` by C++11 `std::copy_n`.
* Removed two dangling backslashes.

### DIFF
--- a/Modules/Core/Common/include/itkFixedArray.h
+++ b/Modules/Core/Common/include/itkFixedArray.h
@@ -26,7 +26,7 @@ namespace itk
 {
 
 /** \class FixedArray
- *  \brief Simulate a standard C array with copy semnatics.
+ *  \brief Simulate a standard C array with copy semantics.
  *
  * Simulates a standard C array, except that copy semantics are used instead
  * of reference semantics.  Also, arrays of different sizes cannot be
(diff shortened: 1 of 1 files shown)


Modules/Core/Common/include/itkFixedArray.h
```diff
@@ -163,33 +163,30 @@ class ITK_TEMPLATE_EXPORT FixedArray
...
-    }
+  {
+    std::copy_n(r, VLength, m_InternalArray);
+  }
 
   /** Operator= defined for a variety of types. */
   template< typename TFixedArrayValueType >
   FixedArray & operator=(const FixedArray< TFixedArrayValueType, VLength > & r)
   {
-    if ( (const void *)r.Begin() != (const void *)m_InternalArray )
```
dzenanz: The equality check has been removed. If it was intentional, what is the reasoning behind it?

N-Dekker: @dzenanz Thanks for your remark, I appreciate that you review so carefully!

Yes indeed, the check was intentionally removed. How could `this` and `r` possibly have their internal arrays starting at the very same memory address? Note that their value types are different, already, for this `operator=` overload. An assignment that just copies fixed arrays of the same type calls the default `operator=(const FixedArray &)`, instead.

dzenanz:
Did this PR cause this warning:
https://open.cdash.org/viewBuildError.php?type=1&buildid=5913488

N-Dekker:
@dzenanz 
> Did this PR cause this warning: https://open.cdash.org/viewBuildError.php?type=1&buildid=5913488

Hmm, I see now:

    itkFixedArrayGTest.cxx:35:42: warning: missing initializer for member 'std::array<double, 2ul>::_M_elems' [-Wmissing-field-initializers]
     std::array<TValue, VLength> stdArray{};

                                          ^
I think the warning was not caused by this PR. I think it was pull request #757 (merged 22 days ago). Still my fault anyway :frowning:

The warning seems to come from old compiler versions only. I think it can be fixed easily, by simply removing that particular pair of curly braces `{}`. Can you fix it? Or do you want me to do it?

dzenanz:
I was to lazy to find the exact PR which introduced the issue, so I commented on this recent one. I kind of remember that you created/touched that file last. I prefer if you do the fix.

N-Dekker:
@dzenanz 
>. I prefer if you do the fix.

OK, but today I won't have time anymore. I guess it can still be done next week, right? Just removing that pair of curly braces, `{}`.

dzenanz:
Of course, that is not a feature 😄

dzenanz:
Tagging is planned for Tuesday or Wednesday.

N-Dekker:
@dzenanz OK, here it is: pull request #914, "COMP: Fixed GCC4 missing-field-initializers warning on `{}` in GTest"

PR #868, opened 2019-05-14
ENH: Convenience overloads for ImageBase Transform member functions

N-Dekker:
Added convenience overloads to the six `ImageBase` Transform member
functions, returning the result by value, instead of having an output
parameter:

    TransformPhysicalPointToIndex
    TransformPhysicalPointToContinuousIndex
    TransformContinuousIndexToPhysicalPoint
    TransformIndexToPhysicalPoint
    TransformLocalVectorToPhysicalVector
    TransformPhysicalVectorToLocalVector

These overloads allow users to have a more "functional" programming
style, or at least, place the transformation result in a `const` variable.

So instead of:

    Bar bar;
    image->TransformFooToBar(foo, bar);

It typically allows users to write:

    const auto bar = image->TransformFooToBar(foo);

Added a unit test to itkImageBaseGTest

### DIFF
--- a/Modules/Core/Common/include/itkImageBase.h
+++ b/Modules/Core/Common/include/itkImageBase.h
@@ -398,15 +398,27 @@ public:
   virtual void SetSpacing(const double spacing[VImageDimension]);
   virtual void SetSpacing(const float spacing[VImageDimension]);
 
-  /** Get the index (discrete) of a voxel from a physical point.
+  /** Returns the index (discrete) of a voxel from a physical point.
    * Floating point index results are rounded to integers
-   * Returns true if the resulting index is within the image, false otherwise
+   * \note This specific overload does not figure out whether or not
+   * the returned index is inside the image. Of course, the user can
+   * still test this afterwards by calling ImageRegion::IsInside(index):
+     \code
+      auto index = image->TransformPhysicalPointToIndex(point);
+      if (image->GetLargestPossibleRegion().IsInside(index)) // Et cetera...
+     \endcode
+   * Which is equivalent to the following code, which calls the other overload:
+     \code
+      IndexType index;
+      if (image->TransformPhysicalPointToIndex(point, index)) // Et cetera...
+     \endcode
    * \sa Transform */
   template< typename TCoordRep >
-  bool TransformPhysicalPointToIndex(
-    const Point< TCoordRep, VImageDimension > & point,
-    IndexType & index) const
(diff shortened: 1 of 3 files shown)


dzenanz:
I have been annoyed by absence of these more than once!

Modules/Core/Common/include/itkImageBase.h
```diff
@@ -398,15 +398,15 @@ class ITK_TEMPLATE_EXPORT ImageBase:public DataObject
   virtual void SetSpacing(const double spacing[VImageDimension]);
   virtual void SetSpacing(const float spacing[VImageDimension]);
 
-  /** Get the index (discrete) of a voxel from a physical point.
+  /** Returns the index (discrete) of a voxel from a physical point.
    * Floating point index results are rounded to integers
-   * Returns true if the resulting index is within the image, false otherwise
```
hjmjohnson: Minor changes request to documentation.  I think it is worth explicitly noting that these functions do not verify that the resulting indices fall within the image.

N-Dekker:
Update: Still preparing an amend that should address the comment by @hjmjohnson _and_ add at least a _very little_ unit test. Hopefully tomorrow...  :-)

Modules/Core/Common/include/itkImageBase.h
N-Dekker: @hjmjohnson Please check the added note:

https://github.com/InsightSoftwareConsortium/ITK/blob/ec7e47678843eddf7ea45d9b83bedaadd24d1c99/Modules/Core/Common/include/itkImageBase.h#L403

https://github.com/InsightSoftwareConsortium/ITK/blob/ec7e47678843eddf7ea45d9b83bedaadd24d1c99/Modules/Core/Common/include/itkImageBase.h#L452 (copied from the previous note, but adapted code snippets)

hjmjohnson:
I approve the wording of the commit message.

N-Dekker:
The GTest that I added compiles well on ITK.Windows, but it fails on macOS and Linux, because I should have added a template keyword on various places. For example:

    imageBase.TransformPhysicalPointToContinuousIndex<float>(PointType())

Should be:

    imageBase.template TransformPhysicalPointToContinuousIndex<float>(PointType())

I'll try to fix it.

N-Dekker:
I think this pull request is OK now. CDash appears stuck "in progress" for no reason. It has succeeded at Linux, Windows and macOS. `:-)`

PR #870, opened 2019-05-15
ENH: Deprecate VectorCastImageFilter

maekclena:
Replace VectorCastImageFilter with the generic CastImageFilter.

I only had to implement a converter from RGBPixel to Vector.

I also fixed a small mistake from the VectorResampleImageFilter deprecation.

### DIFF
--- a/Modules/Core/Common/include/itkRGBAPixel.h
+++ b/Modules/Core/Common/include/itkRGBAPixel.h
@@ -24,7 +24,7 @@
 #endif
 
 #include "itkIndent.h"
-#include "itkFixedArray.h"
+#include "itkVector.h"
 #include "itkMath.h"
 
 namespace itk
@@ -109,6 +109,18 @@ public:
   bool operator<(const Self & vec) const;
 
   bool operator==(const Self & vec) const;
+
+  /** Implicit casting operator. Return an equivalent itk::Vector. */
+  template< typename TPixelOutput >
+  operator Vector< TPixelOutput, 4 >() const
+  {
+    Vector< TPixelOutput, 4 > v;
+    v[0] = static_cast< TPixelOutput >( this->operator[](0) );
+    v[1] = static_cast< TPixelOutput >( this->operator[](1) );
+    v[2] = static_cast< TPixelOutput >( this->operator[](2) );
+    v[3] = static_cast< TPixelOutput >( this->operator[](3) );
+    return v;
+  }
 
   /** Return the number of components. */
   static unsigned int GetNumberOfComponents() { return 4; }
--- a/Modules/Core/Common/include/itkRGBAPixel.h
+++ b/Modules/Core/Common/include/itkRGBAPixel.h
@@ -24,7 +24,7 @@
 #endif
 
 #include "itkIndent.h"
-#include "itkVector.h"
+#include "itkFixedArray.h"
 #include "itkMath.h"
 
 namespace itk
--- a/Modules/Core/Common/include/itkRGBAPixel.h
+++ b/Modules/Core/Common/include/itkRGBAPixel.h
@@ -110,18 +110,6 @@ public:
 
   bool operator==(const Self & vec) const;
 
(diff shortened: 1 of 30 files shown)


Modules/Core/Common/include/itkRGBAPixel.h
```diff
@@ -110,6 +110,18 @@ class ITK_TEMPLATE_EXPORT RGBAPixel:public FixedArray< TComponent, 4 >
...
+  operator Vector< TPixelOutput, 4 >() const
+  {
+    Vector< TPixelOutput, 4 > v;
+    v[0] = static_cast< TPixelOutput >( this->operator[](0) );
+    v[1] = static_cast< TPixelOutput >( this->operator[](1) );
+    v[2] = static_cast< TPixelOutput >( this->operator[](2) );
+    v[3] = static_cast< TPixelOutput >( this->operator[](3) );
+    return v;
+  }
+
```
blowekamp: The addition of the *implicit* conversion is significant. Which test/example required this addition?

Due to its significance please make it a separate commit in this topic.

I would like to discuss if there are/were alternatives to this addition.

Modules/Core/Common/include/itkRGBPixel.h
```diff
@@ -119,6 +118,17 @@ class ITK_TEMPLATE_EXPORT RGBPixel:public FixedArray< TComponent, 3 >
...
+  template< typename TPixelOutput >
+  operator Vector< TPixelOutput, 3 >() const
+  {
+    Vector< TPixelOutput, 3 > v;
+    v[0] = static_cast< TPixelOutput >( this->operator[](0) );
+    v[1] = static_cast< TPixelOutput >( this->operator[](1) );
+    v[2] = static_cast< TPixelOutput >( this->operator[](2) );
+    return v;
+  }
+
```
blowekamp: *ditto*

blowekamp:
Great to see you reviving this deprivation piece by piece! Looks very good.


Please at a note of this depreciation in `Documentation/ITK5MigrationGuide.md`

I'd like to discuss of there are any alternative to the implicit conversions from RGB pixel types to vector types.

Modules/Core/Common/include/itkRGBAPixel.h
maekclena: It is required to compile the example WatershedSegmentation1. An alternative is to use `RGBToVectorImageAdaptor` like [here](https://itk.org/Doxygen/html/WikiExamples_2Smoothing_2VectorGradientAnisotropicDiffusionImageFilter_8cxx-example.html).

blowekamp: The example of using the adaptor with the CastImageFilter looks good to me, and preferable to adding implicit conversions.

As other future work, we might be able to "teach" the `CastImageFilter` and/or `ImageAlgorithm::Copy`  that these pixel types have the same data layout and block copies with out conversion.

romangrothausmann:
@maekclena @dzenanz 
The changes of this PR cause trouble for me to adapt to. I only managed to adjust my  `gradient_mag_vec` CLI for RGB:
https://github.com/romangrothausmann/ITK-CLIs/commit/e558ef2906b75de06d5c3845936d777766766e67
but not for RGBA or VECTOR types, 
nor do I see a way to adjust e.g. [`anisoDiff-grad`](https://github.com/romangrothausmann/ITK-CLIs/commit/e558ef2906b75de06d5c3845936d777766766e67#diff-e81d13253020715f8919cc0c1ceebf67) to work with the deprecation, see e.g. the compilation log:
https://gitlab.com/romangrothausmann/ITK-CLIs/-/jobs/214458383

Is there any way to adapt these CLIs to the new API without loosing the generalization that was working before?

blowekamp:
I'm looking into adding some meta-programming intelligence to `CastImageFilter`.

romangrothausmann:
Many thanks @blowekamp for taking a look. Some "meta-programming intelligence" to CastImageFilter would be really great.

PR #874, opened 2019-05-15
ENH: Add environmental override for git tag of remote module

hjmjohnson:
An environmental variable ${_name}_GIT_TAG_OVERRIDE can be set
in the environment to override the value in the remote module.
The intent of the environmental variable override is to
facilitate testing of remote module branch behaviors without
requiring changes to the ITK code base.


blowekamp:
Why are we using environment variables over CMake variables? We are not frequently using environment variable to change CMake configurations. Additionally the variable name may not start with an ITK prefix like I would expect.

hjmjohnson:
@blowekamp I've moved from Environmental variables to cmake variables as suggested.

blowekamp:
Nice!

PR #878, opened 2019-05-16
Fftw registration warnings

SimonRit:
These two commits fix the warnings of the dashboard results [here](https://my.cdash.org/viewBuildError.php?type=1&buildid=1649833) and [here](https://my.cdash.org/viewBuildError.php?type=1&buildid=1649889).


SimonRit:
Thanks for the review and sorry for not checking the style before hand. It seems to be fixed now.

Issue #871, opened 2019-05-15
Problems with the python wrapping of ShapePriorMAPCostFunction


bpaniagua:
### Description

There seems to be a problem using the ShapePriorMAPCostFunction class in python.

 When assigning the mean and stddev vectors and then printing the cost function object, those arrays seem to be empty. This assignment [ does not create a problem in C++](https://github.com/InsightSoftwareConsortium/ITK/blob/c95b99852a9109a8fc176a06f88b2aa6d442154b/Modules/Segmentation/LevelSets/test/itkShapePriorMAPCostFunctionTest.cxx#L149-L150). The test [here](https://github.com/InsightSoftwareConsortium/ITK/blob/c95b99852a9109a8fc176a06f88b2aa6d442154b/Modules/Segmentation/LevelSets/test/itkShapePriorMAPCostFunctionTest.cxx) uses the ShapePriorMAPCostFunction and it works perfectly.

### Steps to Reproduce
In python, I create a ShapePriorMAPCostFunction in the following way:

costfunction=itk.ShapePriorMAPCostFunction[ImageType, InputPixelType].New()
weights=itk.FixedArray[itk.D,4]()
weights[0] =  1.0  # weight for contour fit term
weights[1] =  20.0 # weight for image fit term
weights[2] =  1.0  # weight for shape prior term
weights[3] =  1.0  # weight for pose prior term
costfunction.SetWeights(weights)
mean=itk.Array[itk.D](pca.GetNumberOfShapeParameters())
mean.Fill(0.0)
stddev=itk.Array[itk.D](pca.GetNumberOfShapeParameters())
stddev.Fill(1.0)
costfunction.SetShapeParameterMeans(mean)
costfunction.SetShapeParameterStandardDeviations(stddev)

I try to use this costfunction object in a GeodesicActiveContourShapePriorLevelSetImage filter, and when I update the filter I get the following error:

RuntimeError: /work/standalone-x64-build/ITKs/Modules/Segmentation/LevelSets/include/itkShapePriorMAPCostFunction.hxx:187:
itk::ERROR: ShapePriorMAPCostFunction(0x9675fa0): ShapeParameterMeans does not have at least 3 number of elements.

When I print the costfunction object I created in the way I described I see this information:

ShapePriorMAPCostFunction (0x9675fa0)
  RTTI typeinfo:   itk::ShapePriorMAPCostFunction<itk::Image<float, 2u>, float>
  Reference Count: 3
[...]

PR #881, opened 2019-05-16
WIP: DECL_PYTHON_VARLEN_SEQ_TYPEMAP typenames only works with Python …

fbudin69500:
…sequence

Addresses issue #871.

an error message.


thewtex:
Repushed in https://github.com/InsightSoftwareConsortium/ITK/pull/1857

PR #882, opened 2019-05-16
BUG: DECL_PYTHON_VARLEN_SEQ_TYPEMAP typenames only works with Python …

fbudin69500:
…sequence

Addresses issue #871.

an error message.


Wrapping/Generators/Python/PyBase/pyBase.i
```diff
@@ -580,19 +580,27 @@ str = str
...
-                    itks[i] = (value_type)PyInt_AsLong(o);
-                } else if (PyFloat_Check(o)) {
-                    itks[i] = (value_type)PyFloat_AsDouble(o);
-                } else {
-                    PyErr_SetString(PyExc_ValueError,"Expecting a sequence of int or float");
-                    return NULL;
+            if (PySequence_Check($input)) {
+                itks = type( PyObject_Length($input) );
+                for (unsigned int i =0; i < itks.Size(); i++) {
+                    PyObject *o = PySequence_GetItem($input,i);
```
thewtex: According to:

https://docs.python.org/3/c-api/sequence.html#c.PySequence_GetItem

`PySequence_GetItem` returns a new reference -- if I recall correctly, the CPython decrement reference macros need to be called after you are finished using the object and in the error conditions to avoid a memory leak.

https://docs.python.org/2/c-api/refcounting.html

Wrapping/Generators/Python/PyBase/pyBase.i
```diff
@@ -611,19 +619,30 @@ str = str
...
-                itks[i] = (value_type)PyInt_AsLong(o);
-            } else if (PyFloat_Check(o)) {
-                itks[i] = (value_type)PyFloat_AsDouble(o);
-            } else {
-                PyErr_SetString(PyExc_ValueError,"Expecting a sequence of int or float");
-                return NULL;
+        if (PySequence_Check($input)) {
+            itks = type( PyObject_Length($input) );
+            for (unsigned int i =0; i < itks.Size(); i++) {
+                PyObject *o = PySequence_GetItem($input,i);
```
thewtex: decref

thewtex:
Rebased on `release`.

thewtex:
Re-opened for cleanliness :broom: in #1857

PR #900, opened 2019-05-17
Add cast image filter vector support

blowekamp:


Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.h
```diff
@@ -119,22 +123,33 @@ class ITK_TEMPLATE_EXPORT CastImageFilter:
 
 #ifdef ITK_USE_CONCEPT_CHECKING
   // Begin concept checking
-  itkConceptMacro( InputConvertibleToOutputCheck,
-                   ( Concept::Convertible< typename TInputImage::PixelType,
-                                           typename TOutputImage::PixelType > ) );
+  //static_assert(std::is_convertible<InputPixelType, OutputPixelType>::value );
```
hjmjohnson: Did you intend for this to be commented out?

blowekamp: This patch gets RGB->Vector pixel types to be "casted" or converted by the filter. Those types are not convertible. But `InputPixelType::ValueType`->`OutputPixelType::ValueType` is... but for scalar don't have the `ValueType` typedef. 

@N-Dekker 
I guess what's needed is a meta-programming "trait" that defaults to fail when SFINAE. for 

```
std::is_convertible<typename InputPixelType::ValueType, typename OutputPixelType::ValueType>::value 
```

Then we can just or the two trait's values.

Examples/Segmentation/WatershedSegmentation1.cxx
```diff
@@ -103,9 +102,8 @@ int main( int argc, char *argv[] )
 
   // Software Guide : BeginCodeSnippet
   using FileReaderType = itk::ImageFileReader< RGBImageType >;
-  using AdaptorType = itk::RGBToVectorImageAdaptor< RGBImageType >;
   using CastFilterType =
-      itk::CastImageFilter< AdaptorType, VectorImageType >;
+      itk::CastImageFilter< RGBImageType, VectorImageType >;
```
dzenanz: This could be on the previous line too.

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.h
dzenanz: Then perhaps remove this commented code?

blowekamp:
> Looks good. Does it work? 😄

It worked on my system!

I fixed up a couple issues with `static_assert`, but the Windows compilation error is still a little bit of a mystery.

dzenanz:
Yes, it is sometimes hard to please all the compilers simultaneously 😠

Here is a recent code written by me, I spent quite a lot time getting it right:
https://github.com/InsightSoftwareConsortium/ITKMontage/blob/c9285cf8c51c08ac4de9e1be5bc4e7fbc7edbe5a/test/itkMontageTestHelper.hxx#L76-L93

blowekamp:
I just tried something different.

dzenanz:
#929 is a Python-only change, it would not affect your C++ code.

blowekamp:
I am also getting a problem with SimpleITK and conversion with std::complex.

romangrothausmann:
Thanks @dzenanz  for that clarification, I wasn't sure if it concerned "wrapping" RGB(A) as Vector in C++ as well. Should I open an issue for the remaining problem or is RGB(A) compatibility for C++ covered by some other open PR?

dzenanz:
You should turn your comment from an hour ago into a new issue.

PR #903, opened 2019-05-17
HDF5 is clobbering ctest TIMEOUT variables and others

blowekamp:


blowekamp:
Not sure what should be done about this.

blowekamp:
@dzenanz @thewtex Having a timeout greater than 600 (IIRC) is required to run many tests. Currently this is overriding any user set variable to enable these test. Do you have a suggestions on what to commit so we can run this tests/examples in the release?

This here is the "quick hack" to disable the HDF5 override.

dzenanz:
It is hard to do something better than this. Hopefully that commenting-out will propagate through future HDF5 updates.

PR #914, opened 2019-05-19
COMP: Fixed GCC4 missing-field-initializers warning on `{}` in GTest

N-Dekker:
Removed a pair of curly braces, `{}`, from the declaration of a local
`std::array` variable in itkFixedArrayGTest, to avoid a warning from
GCC4, reported by @dzenanz:

> itkFixedArrayGTest.cxx:35:42: warning: missing initializer for member
> 'std::array<double, 2ul>::_M_elems' [-Wmissing-field-initializers]

After removing those curly braces, the array may no longer be initialized.
Fortunately the array does not really need to be initialized at the point
of declaration, as each array element is assigned a new value directly
after the declaration.

### DIFF
--- a/Modules/Core/Common/test/itkFixedArrayGTest.cxx
+++ b/Modules/Core/Common/test/itkFixedArrayGTest.cxx
@@ -32,7 +32,7 @@ namespace
   template< typename TValue, unsigned int VLength >
   void Check_FixedArray_supports_retrieving_values_by_range_based_for_loop()
   {
-    std::array<TValue, VLength> stdArray{};
+    std::array<TValue, VLength> stdArray;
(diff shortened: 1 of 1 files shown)


N-Dekker:
Note that in principle, it is entirely correct to use the form `T var{};` to declare and initialize a variable `var` of type `T` in C++11. in this case, GCC should not produce such a warning message: 

[GCC Bug 61489 - Wrong warning with -Wmissing-field-initializers
](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61489)  

Indeed, GCC5 and newer compiler versions don't produce a warning for such a use case anymore.

PR #916, opened 2019-05-20
DOC: Fixed Doxygen description BoundingBox Get member functions

N-Dekker:
The documentation of `BoundingBox` member functions `GetCenter()`,
`GetMinimum()`, and `GetMaximum()` incorrectly stated that these
member functions might return a `nullptr`. This commit uses the term
"a point at the origin" instead.

Also mentioned that `GetCorners()` is a legacy member function.

### DIFF
--- a/Modules/Core/Common/include/itkBoundingBox.h
+++ b/Modules/Core/Common/include/itkBoundingBox.h
@@ -121,8 +121,9 @@ public:
   const PointsContainer * GetPoints() const;
 
   /** Compute and return the corners of the bounding box.
-   *\note This function returns the same points as `GetCorners()`, but it is
-   * `const`, and it avoids dynamic memory allocation by using `std::array`.
+   *\note This function returns the same points as the legacy member function
+   * `GetCorners()`, but it is `const`, and it avoids dynamic memory allocation
+   * by using `std::array`.
   */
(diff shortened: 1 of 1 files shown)


N-Dekker:
It's not so easy to _completely_ describe what these `Get` functions return, in general, as it depends on previous calls to `SetMinimum`, `SetMaximum`, `SetPoints`, `ConsiderPoint`, and `Modified()`. But at least, I believe that the text I'm proposing here is error-free.

N-Dekker:
Thanks for the merge, @thewtex  For the record, here an example to show why I found it hard to _complete_ describe the behavior of those `Get` functions:

    const auto boundingBox = itk::BoundingBox<>::New();
    const itk::Point<float, 3> maximumPoint{ {99, 99, 99} };

    boundingBox->SetMaximum(maximumPoint);

    // This expectation is OK, of course:
    EXPECT_EQ(boundingBox->GetMaximum(), maximumPoint);

    // Now set the object time to the current time.
    boundingBox->Modified();

    // And then, the following expectation fails:
    EXPECT_EQ(boundingBox->GetMaximum(), maximumPoint);

Even though it challenged my expectations, I don't think this behavior is a bug, I think it is _by design_.

PR #917, opened 2019-05-20
Follow up commit to fix itkBoundingBox with LEGACY support

fbudin69500:
When compiling ITK with ITK_LEGACY_SILENT set to ON, e.g. when compiling
with Python wrapping ON, ITK was not compiling because itkVectorContainer
class does not have a member function called `clear()` or a function called
`reserve()`. This regression was introduced in [1] and the PR is was a
follow up of.

[1] 5b508a5e88bb70ec6667569f6adcf9c915225f08

### DIFF
--- a/Modules/Core/Common/include/itkBoundingBox.hxx
+++ b/Modules/Core/Common/include/itkBoundingBox.hxx
@@ -121,8 +121,8 @@ const typename BoundingBox< TPointIdentifier, VPointDimension, TCoordRep,
 BoundingBox< TPointIdentifier, VPointDimension, TCoordRep, TPointsContainer >
 ::GetCorners()
 {
-  m_CornersContainer->clear();
-  m_CornersContainer->reserve(NumberOfCorners);
+  m_CornersContainer->Initialize();
+  m_CornersContainer->Reserve(NumberOfCorners);
 
   for (const PointType& pnt: this->ComputeCorners())
     {


fbudin69500:
@N-Dekker : Those functions are indeed in the file, but do not appear to be available if you look at the doxygen documentation [here](https://itk.org/Doxygen/html/classitk_1_1VectorContainer-members.html) or at the compilation that is failing line 6822 [here](https://dev.azure.com/itkrobotlinuxpython/ITK.Linux.Python/_build/results?buildId=858).

thewtex:
The call is on `MapContainer` instead of `VectorContainer`:

```
::GetCorners() [with TPointIdentifier = long unsigned int; int
VPointDimension = 3; TCoordRep = float; TPointsContainer =
itk::MapContainer<long unsigned int, itk::Point<float, 3> >;
itk::BoundingBox<TPointIdentifier, VPointDimension, TCoordRep,
TPointsContainer>::PointsContainer = itk::MapContainer<long unsigned
int, itk::Point<float, 3> >]’:
Wrapping/Modules/ITKCommon/itkBoundingBoxPython.cpp:9659:59:
required from here
/home/kitware/mattm/ITK/Modules/Core/Common/include/itkBoundingBox.hxx:125:23:
error: ‘using ObjectType = class itk::MapContainer<long unsigned int,
itk::Point<float, 3> > {aka class itk::MapContainer<long unsigned int,
itk::Point<float, 3> >}’ has no member named ‘reserve’; did you mean
‘Reserve’?
m_CornersContainer->reserve(NumberOfCorners);
~~~~~~~~~~~~~~~~~~~~^~~~~~~
Reserve
```

thewtex:
Unrelated Python tests are failing.

N-Dekker:
Thanks for your explanations @fbudin69500 and @thewtex, I do understand the issue now. 

Now that I understand, I would have liked to get the opportunity to review this more carefully. But it is merged so quickly, while I did not yet approve!

`m_CornersContainer->clear()` has been in there already on Feb 23, 2005, by Julien @jjomier, so it should not have been a problem:

https://github.com/InsightSoftwareConsortium/ITK/blob/423562511fff3e18284572c8fb267c16f249485e/Code/Common/itkBoundingBox.txx#L84

`m_CornersContainer->reserve()` was wrong indeed [(my fault)](https://github.com/InsightSoftwareConsortium/ITK/commit/8383f1bf94df8a9b16266dcc34e422c12befd213) as I did not realize before that `TPointsContainer` could be a map type, instead of a vector type. However, the proposed alternative is also wrong:

    m_CornersContainer->Reserve(NumberOfCorners)

This function call does not just `reserve` the way it is specified for `std::vector`. Instead, it resizes the container! This yields an incorrect result! Seriously!

fbudin69500:
@N-Dekker : Could you explain why you think that `Reserve()` does not work in this case? It does seems to compile (at least in the cases used in ITK Python wrapping). I created another PR to address your concerns about replacing the call to the method `clear()`.

N-Dekker:
@fbudin69500 Thanks for asking `:-)` I don't know why `itk::VectorContainer::Reserve` was implemented that way. But the following expectations (which are not yet in ITK) fail, now that GetCorners() calls `Reserve`:

    // Expect four corners for a two-dimensional box:
    const auto boundingBox_2D = itk::BoundingBox<itk::IdentifierType, 2>::New();
    EXPECT_EQ(boundingBox_2D->GetCorners()->size(), 4);

    // Expect eight corners for a three-dimensional box:
    const auto boundingBox_3D = itk::BoundingBox<itk::IdentifierType, 3>::New();
    EXPECT_EQ(boundingBox_3D->GetCorners()->size(), 8);

GTest output:

> error: Expected equality of these values:
>  boundingBox_2D->GetCorners()->size()
>    Which is: 8
>  4
> error: Expected equality of these values:
>  boundingBox_3D->GetCorners()->size()
>    Which is: 16
>  8

N-Dekker:
@fbudin69500 The following expectations do pass _succesfully_! Which meanst that `vectorContainer->Reserve(42)` does resize the container!

    const auto vectorContainer = itk::VectorContainer<std::size_t, int>::New();
    vectorContainer->Reserve(42);
    EXPECT_EQ(vectorContainer->Size(), 42);
    EXPECT_EQ(vectorContainer->size(), 42);

(Would be nice to have these tests as GoogleTest in ITK, in my opinion.)

fbudin69500:
@N-Dekker : From the tests you posted, it seems that what fails is not the `Reserve()` part of the code, but the size itself that is wrong. This means that maybe [this line](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkBoundingBox.h#L97) is incorrect. The fact that the container is resized does not seem to me to be a problem (except that potentially speed could be improved if this was reserved only the first time or at compile time since the size is based on the dimension which is a template parameter. The only reason why maybe changing the size at initialization time could be dangerous in my opinion, is that one could get the uninitialized values if they call this functions before it is initialized correctly, but I don't even know if this is possible. I have to look more at the code.

fbudin69500:
Looking at the code, it looks like it would make more sense to initialize the size of the bounding box at compilation time, since the size is always the same, and it doesn't seem to be possible to get this array without initializing it. This class is not derived inside ITK. The only way I could think that could create a problem would be if someone was creating a derived class outside of ITK and changing the behavior of this function. That seems like a bad idea to do so anyway.

N-Dekker:
@fbudin69500 
> Looking at the code, it looks like it would make more sense to initialize the size of the bounding box at compilation time, since the size is always the same

This is indeed one of the reasons why I wanted GetCorners() to be deprecated, as it is right now: https://github.com/InsightSoftwareConsortium/ITK/pull/892/commits/814419ae9f989c10c597a5e913166d3e8c949267 It is superseded by ComputeCorners(), which returns a compile-time fixed array of corners.

I do strongly believe that this code correctly calculates the number of corners of a bounding box (4 in 2D, 8 in 3D):

https://github.com/InsightSoftwareConsortium/ITK/blob/8383f1bf94df8a9b16266dcc34e422c12befd213/Modules/Core/Common/include/itkBoundingBox.h#L97

Do you still doubt it?

fbudin69500:
I agree with you, this code that I linked earlier works correctly. I found the origin of the problem and addressed it in my follow up PR #927

PR #920, opened 2019-05-20
std::list of `LineSpatialObject` needs to be wrapped

fbudin69500:
The ITK filter `HoughTransform2DLinesImageFilter` return an `std::list` of
`LineSpatialObject` and therefore this type needs to be wrapped in Python
for the filter to be usable in Python.


fbudin69500:
The failing tests seem unrelated and should be fixed with #934

PR #922, opened 2019-05-20
WIP: Out of bounds read in testMeta6Surface

blowekamp:


blowekamp:
This patch just guards against an out of bounds access. The problem is that `pntDim` is 9 and not 10 for the number of components in the SurfacePnt structure when read.

dzenanz:
Replicating this in MetaIO:
https://github.com/Kitware/MetaIO/pull/78

dzenanz:
Superseded by #935.

PR #926, opened 2019-05-21
STYLE: Rename ITK macro EXPECT_VECTOR_NEAR to ITK_EXPECT_VECTOR_NEAR

N-Dekker:
Renamed the unit test macro `EXPECT_VECTOR_NEAR` to `ITK_EXPECT_VECTOR_NEAR`,
to explicitly indicate that it is an ITK specific extension to GTest.

### DIFF
--- a/Modules/Core/Common/test/itkAggregateTypesGTest.cxx
+++ b/Modules/Core/Common/test/itkAggregateTypesGTest.cxx
@@ -152,27 +152,27 @@ public:
     index1.at(1)  = 6; //non-const at()
     index1.m_InternalArray[2] = 6; //Direct access
     index1.SetElement(3,6); //SetElement
-    EXPECT_VECTOR_NEAR(index1, knownAll6s, 0);
+    ITK_EXPECT_VECTOR_NEAR(index1, knownAll6s, 0);
 
     AggregateType index2 = index1; // Copy constructor
-    EXPECT_VECTOR_NEAR(index2, index1, 0);
+    ITK_EXPECT_VECTOR_NEAR(index2, index1, 0);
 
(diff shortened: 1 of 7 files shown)


fbudin69500:
Does the new name help differentiate from other functions that have a similar name (e.g. Google test macros or functions?)

N-Dekker:
> Does the new name help differentiate from other functions that have a similar name (e.g. Google test macros or functions?)

That's an extension to the philosophical question of @dzenanz   `:-)`

I thought it was a coding convention of ITK to always use "ITK" or "itk" as prefix for macro names. And I like this convention, as it helps me to find the origin of a macro, when I encounter the macro being used in source code. In general, this convention also prevents naming conflicts with external libraries. What do you think?

dzenanz:
Avoiding collisions with other libraries might be the most significant. People really like calling their class `vector`, so we need prefixes like `std`, `itk` and `Eigen`. As macros don't have namespaces, using prefixes is the norm.

blowekamp:
> Avoiding collisions with other libraries might be the most significant. People really like calling their class `vector`, so we need prefixes like `std`, `itk` and `Eigen`. As macros don't have namespaces, using prefixes is the norm.

The macro implementation is generic, it only requires `itk::NumericTraits<T1>::GetLength` and `operator[]` so it already works with `std::vector` and combination with itk Array/vector like objects.

thewtex:
In general we should have an ITK prefix for macros so we know where they come from and avoid conflicts. However, this should be the case with all the testing macros:

  https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/TestKernel/include/itkTestingMacros.h

PR #927, opened 2019-05-21
BUG: Revert changing call to `Initialize()` with `clear()`

fbudin69500:
This is a follow up of commit [1].
The method `clear()` has been used successfully in itkBoundingBox
for more than 10 years. There is no reason to replace it with a
call of the method `Initialize()`.

[1] 3a7ce4b6d8c76e30e6b89ddc850d5745b18a45d3

### DIFF
--- a/Modules/Core/Common/include/itkBoundingBox.hxx
+++ b/Modules/Core/Common/include/itkBoundingBox.hxx
@@ -121,7 +121,7 @@ const typename BoundingBox< TPointIdentifier, VPointDimension, TCoordRep,
 BoundingBox< TPointIdentifier, VPointDimension, TCoordRep, TPointsContainer >
 ::GetCorners()
 {
-  m_CornersContainer->Initialize();
+  m_CornersContainer->clear();
   m_CornersContainer->Reserve(NumberOfCorners);
 
   for (const PointType& pnt: this->ComputeCorners())
--- a/Modules/Core/Common/include/itkBoundingBox.hxx
+++ b/Modules/Core/Common/include/itkBoundingBox.hxx
@@ -34,7 +34,7 @@ namespace itk
 /**
  * Print out the bounding box.
  */
-template< typename TPointIdentifier, int VPointDimension,
+template< typename TPointIdentifier, unsigned int VPointDimension,
           typename TCoordRep, typename TPointsContainer >
 void
 BoundingBox< TPointIdentifier, VPointDimension, TCoordRep, TPointsContainer >
@@ -53,7 +53,7 @@ BoundingBox< TPointIdentifier, VPointDimension, TCoordRep, TPointsContainer >
 /**
  * Access routine to set the points container.
  */
-template< typename TPointIdentifier, int VPointDimension,
+template< typename TPointIdentifier, unsigned int VPointDimension,
           typename TCoordRep, typename TPointsContainer >
 void
 BoundingBox< TPointIdentifier, VPointDimension, TCoordRep, TPointsContainer >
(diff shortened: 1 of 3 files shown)


fbudin69500:
This is to address concerns expressed in comments of PR #917.

fbudin69500:
@N-Dekker Would you be able to update this PR with the tests you suggested in #917 ?

fbudin69500:
@N-Dekker : This updated version of the PR should correct the wrong size of the list. If you integrate the tests you mentioned in #917, they should pass.

N-Dekker:
> Would you be able to update this PR with the tests you suggested in #917 ?

Sorry, not today anymore. So if there is going to be a 5.0.0 tag today, maybe your bug fix could be merged without unit test...? (I hate to say it! Fix without test! :smiling_imp: )

fbudin69500:
I'll add a simple test that doesn't need google test.

dzenanz:
See this regarding legacy test warnings:
https://github.com/InsightSoftwareConsortium/ITK/blob/fb6cf97441467e022834c651ab02b61bf30543e3/Modules/Core/Common/include/itkMacro.h#L538-L542

fbudin69500:
@dzenanz Thanks, I have updated my PR.

Modules/Core/Common/test/itkBoundingBoxTest.cxx
```diff
@@ -205,6 +206,21 @@ int itkBoundingBoxTest (int, char*[])
     }
   }
 
+  // Expect four corners for a two-dimensional box:
```
N-Dekker: Before this line, I think you should do `#ifndef ITK_LEGACY_REMOVE`

Modules/Core/Common/test/itkBoundingBoxTest.cxx
```diff
@@ -205,6 +206,21 @@ int itkBoundingBoxTest (int, char*[])
...
+    return EXIT_FAILURE;
+    }
+
+  // Expect eight corners for a three-dimensional box:
+  const auto boundingBox_3D = itk::BoundingBox<itk::IdentifierType, 3>::New();
+  if(boundingBox_3D->GetCorners()->size() != 8)
+    {
+    std::cerr << "Wrong number of corners for 3D box (expects 8):" << boundingBox_3D->GetCorners()->size() << std::endl;
+    return EXIT_FAILURE;
+    }
```
N-Dekker: `#endif // Not ITK_LEGACY_REMOVE`

Modules/Core/Common/include/itkBoundingBox.hxx
```diff
@@ -124,10 +124,10 @@ BoundingBox< TPointIdentifier, VPointDimension, TCoordRep, TPointsContainer >
   m_CornersContainer->clear();
   m_CornersContainer->Reserve(NumberOfCorners);
 
+  SizeValueType index = 0;
   for (const PointType& pnt: this->ComputeCorners())
     {
-    // Push back is not defined so we insert at the end of the list
-    m_CornersContainer->InsertElement(m_CornersContainer->Size(), pnt);
+    m_CornersContainer->SetElement(index++, pnt);
```
N-Dekker: For what it's worth, the original code from 2005 simply did `push_back(pnt)`!

https://github.com/InsightSoftwareConsortium/ITK/blob/423562511fff3e18284572c8fb267c16f249485e/Code/Common/itkBoundingBox.txx#L101 

But `SetElement` should also be OK in this case.

Modules/Core/Common/test/itkBoundingBoxTest.cxx
fbudin69500: done

Modules/Core/Common/test/itkBoundingBoxTest.cxx
fbudin69500: done.

fbudin69500:
Failed because of unrelated failing tests.

PR #929, opened 2019-05-21
BUG: Add casting for vector pixel types

blowekamp:
A transfer of types wrapped for the VectorCastImageFilter to
the CastImageFilter.


blowekamp:
This patch was just to transfer what was wrapped with `VectorCastImageFilter` to the `CastImageFilter`.

fbudin69500:
Are these two filters doing exactly the same thing then? Does it make sense to wrap that twice? Should we remove the wrapping of `VectorCastImageFilter` to limit the size of the package?

blowekamp:
https://github.com/InsightSoftwareConsortium/ITK/pull/870

PR #933, opened 2019-05-21
BUG: ITK could not find Eigen3::Eigen target when using system Eigen

fbudin69500:
The CMake command `find_package()` used for Eigen3 imports the Eigen3::Eigen
target, but this target is only visible in that folder and its subfolders.
For this target to be used in other ITK module, the command `find_package()`
needs to be called for each ITK modules that needs Eigen. We removed
the condition around `find_package()` that is called in
`ITKEigen3_EXPORT_CODE_BUILD` to always call `find_package()`, even when
it is done inside the ITK build tree.

### DIFF
--- a/Modules/ThirdParty/Eigen3/CMakeLists.txt
+++ b/Modules/ThirdParty/Eigen3/CMakeLists.txt
@@ -1,5 +1,9 @@
 project(ITKEigen3)
 set(ITKEigen3_THIRD_PARTY 1)
+
+option(ITK_USE_SYSTEM_EIGEN "Use an outside build of Eigen3." ${ITK_USE_SYSTEM_LIBRARIES})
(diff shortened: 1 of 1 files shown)


thewtex:
```
CMake Error at /Users/vsts/agent/2.150.3/work/1/s-build/Modules/ThirdParty/Eigen3/src/itkeigen/ITKInternalEigen3Config.cmake:20 (include):
  The file

    /Users/vsts/agent/2.150.3/work/1/s-build/Modules/ThirdParty/Eigen3/src/itkeigen/ITKInternalEigen3Targets.cmake

  was generated by the export() command.  It may not be used as the argument
  to the include() command.  Use ALIAS targets instead to refer to targets by
  alternative names.
```

PR #934, opened 2019-05-21
ImageSink outputs

thewtex:


Modules/Core/Common/include/itkImageSink.hxx
```diff
@@ -94,12 +94,20 @@ ImageSink<TInputImage>
 template <typename TInputImage>
 void
 ImageSink<TInputImage>
-::Update( )
+::UpdateLargestPossibleRegion()
```
dzenanz: Is this left empty on purpose?

blowekamp:
Looks like good stuff +1

I am a little uncertain about the why the empty implementation os `UpdateLargestPossbileRegion` is required.

thewtex:
After updates, `StatisticsImageFilterTest_1`, etc. would fail because they are calling `UpdateLargestPossibleRegion()`, and this was not triggering the update.

Modules/Core/Common/include/itkImageSink.hxx
thewtex: Good catch, missed in a rebase -- it should be calling `this->Update()`. I will push a follow-up.

blowekamp: @thewtex This does not correspond to you comment.

There is still the default ProcessObject's Primary output, since we are not setting one in the filters now. The construction of it looks very odd:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/src/itkProcessObject.cxx#L67-L69

blowekamp:
OK, so the effect is that the "PrimayOutput" is now a "null" upon construction, but may be reconstructed as a `DataObject` by `MakeOutput` some time. The need to override `UpdateLargestPossibleRegion` is because the `PrimayOutput` is null:
https://github.com/InsightSoftwareConsortium/ITK/blob/073f17e0209bd6c7605299193c50c87865660e23/Modules/Core/Common/src/itkProcessObject.cxx#L1899

I don't think we have other example in ITK of filters without a primary output. There are aspects of ProcessObject that should be refunded and defined for these use cases. Perhaps the `Primary` output should alway be null for this usage?

thewtex:
> There are aspects of ProcessObject that should be refunded and defined for these use cases. Perhaps the `Primary` output should alway be null for this usage?

There are possibly other bugs with ImageSink / ProcessObject. Setting the Primary output to null may work, but it could also have other side effects given the complexity of ProcessObject and related code. If pursued, it should come after 5.0.0.

Ideally, too, PrimaryOutput would not be restricted to an IndexedOutput. But that is a topic for ITK 6 :-).

blowekamp:
> > There are aspects of ProcessObject that should be refunded and defined for these use cases. Perhaps the `Primary` output should alway be null for this usage?
> 
> There are possibly other bugs with ImageSink / ProcessObject. Setting the Primary output to null may work, but it could also have other side effects given the complexity of ProcessObject and related code. If pursued, it should come after 5.0.0.
> 
> Ideally, too, PrimaryOutput would not be restricted to an IndexedOutput. But that is a topic for ITK 6 :-).

The Primary output/input is always index 0. It also has a "name" too. I think the most less risky thing to do here to to add "GetOutput" methods to the problematic cases and keep the name/0/primary output as it was.
 e.g.:
```
--- a/Modules/Filtering/ImageStatistics/include/itkMinimumMaximumImageFilter.h
+++ b/Modules/Filtering/ImageStatistics/include/itkMinimumMaximumImageFilter.h
@@ -82,6 +82,7 @@ public:
 
   /** Return the computed Minimum. */
   itkGetDecoratedOutputMacro(Minimum, PixelType);
+  const PixelObjectType *GetOutput() const {this->GetMinimum();};
```

The downside/compatibility issue is that the "GetOutput" use to be an image.

thewtex:
> +  const PixelObjectType *GetOutput() const {this->GetMinimum();};

This will not fix the failing test.

Changing the content of `GetOutput` will be unexpected for existing clients of the code. There is no reason that "Minimum" should be considered Primary over "Maximum" or that "Sigma" is Primary relative to the other statistics or that these outputs are indexed when the others are not.

blowekamp:
> This will not fix the failing test.
As I really from Friday, it did. Did you try it?

> Changing the content of `GetOutput` will be unexpected for existing clients of the code. There is no reason that "Minimum" should be considered Primary over "Maximum" or that "Sigma" is Primary relative to the other statistics or that these outputs are indexed when the others are not.

Yes, I agree with those points. But this patch which has the "Primary" be `nullptr` may have other latent bug, similar to #936. It's a matter of choosing the better imperfect choice, unless there is another proposal.

thewtex:
We need a clean build for the release, which needs to get under way, so this will have to go in as-is.

#936 is a separate bug that can be merged when passing.

There is a lot of complexity in ProcessObject related to an attempt to have an input / output be both indexed and named. I would be surprised if there are not more latent bugs.

blowekamp:
I’ll do a code review looking for additional uses of GetPrimaryOutput that are not protected with a is null check tomorrow.

PR #937, opened 2019-05-21
COMP: Address missing fftw include in CurvatureRegistrationFilter

blowekamp:
When FFTW is enable the CurvatureRegistrationFilter is dependent upon
the ITKFFTW module.


Modules/Registration/PDEDeformable/itk-module.cmake
```diff
@@ -2,8 +2,14 @@ set(DOCUMENTATION "This module contains classes for deformable image
 registration based on intensity differences by solving the PDE, optical flow
 problem.  This includes Thirion's popular \"demons\" algorithm.")
 
+
```
fbudin69500: You may want to add a `set(_FFTW_DEPENDS)` above to make sure it is empty when starting this function (except if you can think of a case when it should be possible to manually pass this variable.

blowekamp: done

Issue #925, opened 2019-05-21
GDCM warnings


SimonRit:
When compiling ITK with Module_RTK, I have a bunch of warnings which are related to GDCM, see the list in the following dashboard results
[https://my.cdash.org/viewBuildError.php?type=1&buildid=1652539](https://my.cdash.org/viewBuildError.php?type=1&buildid=1652539)
[https://my.cdash.org/viewBuildError.php?type=1&buildid=1652545](https://my.cdash.org/viewBuildError.php?type=1&buildid=1652545)
[https://my.cdash.org/viewBuildError.php?type=1&buildid=1652594](https://my.cdash.org/viewBuildError.php?type=1&buildid=1652594)
Should I (1) submit a patch to fix them in ITK, (2) disable warnings in ITK for GDCM with itk_module_warnings_disable (I'm not sure what to do with the "has no symbols" warnings then) or maybe @malaterre wants to fix them both in GDCM and ITK? Thanks!

malaterre:
@SimonRit If you write a patch I can either git cherry-pick / am in GDCM if needed. Thanks

SimonRit:
Alright, I'll propose something. Can you explain empty files like [gdcmItem.cxx](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/ThirdParty/GDCM/src/gdcm/Source/DataStructureAndEncodingDefinition/gdcmItem.cxx) which generate those "has no symbols" warnings? Are they necessary for some reason or can my patch remove them?

malaterre:
Well really the warning is annoying to me. Anyway I'd prefer you move the gdcm::Item::Clear() definition to the cxx file. This should silence the warning.

PR #939, opened 2019-05-21
COMP: fix GDCM warnings

SimonRit:
Warnings were of three types:
.* GCC 4.8.5: "warning: unused parameter"
* AppleClang 10.0.0.10001145: warning: cast from 'const char *' to 'const gdcm::Element<256, 218367>::Type *' (aka 'const float *') increases required alignment from 1 to 4 [-Wcast-align]
* AppleClang 10.0.0.10001145: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: (...) has no symbols

See issue GH-925, please review @malaterre.

### DIFF
--- a/Modules/ThirdParty/GDCM/src/gdcm/Source/Common/CMakeLists.txt
+++ b/Modules/ThirdParty/GDCM/src/gdcm/Source/Common/CMakeLists.txt
@@ -131,33 +131,21 @@ set(Common_SRCS
   gdcmRegion.cxx
   gdcmBoxRegion.cxx
   gdcmEvent.cxx
-  gdcmDataEvent.cxx
-  gdcmProgressEvent.cxx
-  gdcmFileNameEvent.cxx
   gdcmCommand.cxx
   gdcmMD5.cxx
   gdcmBase64.cxx
   gdcmSHA1.cxx
   gdcmDummyValueGenerator.cxx
-  #gdcmCryptographicMessageSyntax.cxx
-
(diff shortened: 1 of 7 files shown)


malaterre:
That's just removing the type, the compiler report looks accurate to me, so please do not merge this portion of the patch.

`const Type* array = (const Type*)(void*)bv->GetPointer();`

SimonRit:
I know! I just want to silence this warning but I don't know if there is a better solution. This solution is based on a mozilla patch, see [here](https://bugzilla.mozilla.org/show_bug.cgi?id=659546). Any better suggestion for removing the "increases required alignment" blabla?

malaterre:
The only remaining warning that is not addressed in 942 is the one about no symbols. However this should not be addressed within GDCM itself since ITK is hiding this warning globally:

```
$ git grep ranlib
CMake/CTestCustom.cmake.in:  "ranlib:.*file:.*has no symbols"
CMake/CTestCustom.cmake.in:  "ranlib: file: .+ has no symbols"
```

I suspect the warnings are only seen from project outside of ITK.

dzenanz:
Thanks @SimonRit and @malaterre.

malaterre:
@SimonRit I do not understand either. Any ctest expert here ?

dzenanz:
@thewtex or @blowekamp might be familiar with suppression. @bradking might give us more insight into CTest's behavior.

PR #941, opened 2019-05-22
BUG: InterpolateImageFunction::GetRadius hidden in ITKV4_COMPATIBILITY

thewtex:
This function is required for streaming with ResampleImageFilter in ITK
5.0. The abstract method is hidden with ITKV4_COMPATIBILITY so classes
that derive from InterpolateImageFunction do not encounter build errors.
When compatibility is turned off interpolator authors with get a compile
time error until the method is defined to ensure correct behavior when
resampling.

### DIFF
--- a/Modules/Core/ImageFunction/include/itkInterpolateImageFunction.h
+++ b/Modules/Core/ImageFunction/include/itkInterpolateImageFunction.h
@@ -19,6 +19,8 @@
 #define itkInterpolateImageFunction_h
 
 #include "itkImageFunction.h"
+#include "itkConfigure.h"
+
 
 namespace itk
 {
@@ -127,12 +129,14 @@ public:
     return ( static_cast< RealType >( this->GetInputImage()->GetPixel(index) ) );
   }
 
+#if !defined(ITKV4_COMPATIBILITY)
   /** Get the radius required for interpolation.
    *
    * This defines the number of surrounding pixels required to interpolate at
    * a given point.
    */
   virtual SizeType GetRadius() const = 0;
+#endif
 
 protected:
   InterpolateImageFunction()= default;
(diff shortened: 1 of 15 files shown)


Modules/Core/ImageFunction/include/itkInterpolateImageFunction.h
```diff
@@ -127,12 +129,14 @@ class ITK_TEMPLATE_EXPORT InterpolateImageFunction:
     return ( static_cast< RealType >( this->GetInputImage()->GetPixel(index) ) );
   }
 
+#if !defined(ITKV4_COMPATIBILITY)
   /** Get the radius required for interpolation.
    *
    * This defines the number of surrounding pixels required to interpolate at
    * a given point.
    */
   virtual SizeType GetRadius() const = 0;
```
blowekamp: I suggest if compatibility in enabled the the implementation then change this from an abstract implementation to an implementation which returns the largest possible region. This would retain the streaming feature when compatibility is enabled, and still not require derived interpolators to have it implemented, as it would default to a value which disables streaming.

blowekamp:
-1

N-Dekker:
@blowekamp wrote:
> -1

Do you mean this PR should be changed?

blowekamp:
This patch understandably came through quickly.

I think there are solutions to the problem which can provide consistent behavior and not have this feature disabled when compatibility is enabled

thewtex:
The abstract behavior is intended. An Interpolator author should implement the correct radius required, and they need to know that it needs to be implemented.

Modules/Core/ImageFunction/include/itkInterpolateImageFunction.h
thewtex: This will silently defeat the purpose of the feature.

dzenanz: This patch completely removed the feature. And Brad's proposal leaves the feature implemented for built-in classes, but does not force its implementation for external classes when `ITKV4_COMPATIBILITY` is defined. Which is a good path from V4 to V5.

PR #942, opened 2019-05-22
replace 939

malaterre:
## PR Checklist


malaterre:
If not clear this PR replaces PR 939

@SimonRit

issakomi:
Doesn't have any effect, but that [here](https://github.com/InsightSoftwareConsortium/ITK/blob/ef3e412eb2396786e5bf5a2cce8f4cb891406482/Modules/ThirdParty/GDCM/src/gdcm/Source/DataStructureAndEncodingDefinition/gdcmElement.h#L660) (near that cast) is useless assert

```
      if(array)
      {
        assert(array); // That would be bad
```

SimonRit:
> If not clear this PR replaces PR #939.
> 
> @SimonRit

Thanks. I approve this PR. Can you comment on why this is better than what was in #939 for the alignment warning? It seems to be doing the same thing to me... Any pointer to an explanation?

malaterre:
@SimonRit Technically you are correct, I am indeed using your trick to avoid a compiler warning. However the consumer of the `GetPointer` API should not be responsible for specifiying the alignement of the char * buffer. Since there is no way in c++ (even c++20) [to do so](https://stackoverflow.com/questions/56256431/warning-cast-from-char-to-float-increases-required-alignment-from-1-to). I simply exposed a void* API to the user.

thewtex:
Failing test is unrelated.

PR #947, opened 2019-05-22
Keep interpolator radius behavior in itkv4 compatibility

blowekamp:


thewtex:
`GetRadius` was not present in ITKv4.

blowekamp:
> `GetRadius` was not present in ITKv4.

Please elaborate.

This patch preserves ITKv5 features and enables old interpolators to work as they did before with out streaming.

romangrothausmann:
Hm, I find this difficult to judge about. Do we have a definition of what `ITKV4_COMPATIBILITY` is meant to mean? This could serve as a basis for deciding which way to go.

As I understand it (please correct me if I got it wrong), #941 is meant to force a contributor of a new class that derives from `InterpolateImageFunction` to implement `GetRadius` for ITK-5 contribs but to not have to bother about this in case `ITKV4_COMPATIBILITY` is enabled; while this PR would pre-define a `GetRadius` that would disable streaming as long as `GetRadius` is not overridden in a class derived from `InterpolateImageFunction`, therefore not enforcing an author to think about implementing streaming capabilities.
Allowing to have `GetRadius` in `ITKV4_COMPATIBILITY` (but hidden) -- while `GetRadius` was not in ITK4 -- is, I would say, depending on what `ITKV4_COMPATIBILITY` is meant to define.

From a general point of view, I would say that forcing an author to think about an implementation for  `GetRadius` is a good idea, since the author is likely best immersed into the problem. If not, it would motivate the author to consider how the contribution could be generalized such that it allows streaming.

blowekamp:
That is not quite correct on this patch. 

With this patch the base implementation will   only defined when ITKV4_COMPATIBILITY is enabled. Without that option, derived interpolator a still are required to implement GetRadius. 

The difference with this patch is that when ITKV4_COMPATIBILITY is enabled the ResampleImageFilter still retains its streaming feature.

The question that needs to be asked and documented is: Does streaming in ResampleImageFilter break “compatibility”?

romangrothausmann:
> With this patch the base implementation will only defined when ITKV4_COMPATIBILITY is enabled. Without that option, derived interpolator a still are required to implement GetRadius.

OK, I was not aware of that. Thanks for that clarification.

> The difference with this patch is that when ITKV4_COMPATIBILITY is enabled the ResampleImageFilter still retains its streaming feature.

That I got, and that I would find interesting also for `ITKV4_COMPATIBILITY`.

> The question that needs to be asked and documented is: Does streaming in ResampleImageFilter break “compatibility”?

Hm, to my understanding it would not break “compatibility” but might not be expected (which not necessary falls under “compatibility”).

Modules/Core/ImageFunction/include/itkInterpolateImageFunction.h
```diff
@@ -129,14 +127,26 @@ class ITK_TEMPLATE_EXPORT InterpolateImageFunction:
...
+    // largest by default.
+    const InputImageType* input = this->GetInputImage();
+    if ( !input )
+      {
+      itkExceptionMacro( "Input image required!" );
+      }
+    return input->GetLargestPossibleRegion().GetSize();
+  }
+  #else
   virtual SizeType GetRadius() const = 0;
```
romangrothausmann: I see, so this enforces a derived interpolator to implement `GetRadius` while providing a default  `GetRadius` definition that disables streaming in case of `ITKV4_COMPATIBILITY` except for the already implemented overrides.

dzenanz:
Streaming is not requested by default. If someone requests it, it means they want it. Previously, resample filter didn't offer that. With this patch they get that with ITK 5, even if they operate under `ITKV4_COMPATIBILITY`. It is a good surprise. But I expect this case to be rare. Also, this patch reduces amount of code, which is always a good thing.

dzenanz:
The test failure seems like an unrelated fluke.

blowekamp:
> Streaming is not requested by default. If someone requests it, it means they want it. Previously, resample filter didn't offer that. With this patch they get that with ITK 5, even if they operate under `ITKV4_COMPATIBILITY`. It is a good surprise. But I expect this case to be rare. Also, this patch reduces amount of code, which is always a good thing.

Just a detail, the ITKv4 Resampling did produce an arbitrary output region ( part of streaming ), which is a feature I have used many times. So the output of the v5 Resampler should be the same size as before, the difference occurs with the size of the upstream RR and the reduced computations. *surprise* your code runs faster.

dzenanz:
Please rebase on top of #973, once that is merged in. That should also re-trigger the CI tests.

blowekamp:
rebased after #973, and reverted #973

dzenanz:
Thanks, and sorry for the inconvenience.

PR #949, opened 2019-05-22
BUG: Add test/CMakeLists.txt stub to NumPy bridge

thewtex:
To address:

  CMake Error at CMake/ITKModuleEnablement.cmake:319 (add_subdirectory):
     The source directory

       /home/matt/src/ITK/Modules/Bridge/NumPy/test

     does not contain a CMakeLists.txt file.
     CMakeLists.txt:433 (include)

When BUILD_EXAMPLES, ITK_WRAP_PYTHON, and BUILD_TESTING are enabled.

### DIFF
--- a/Modules/Bridge/NumPy/test/CMakeLists.txt
+++ b/Modules/Bridge/NumPy/test/CMakeLists.txt
@@ -1,0 +1,1 @@
+itk_module_test()


fbudin69500:
I cannot reproduce this error. Could you give more information about your configuration?

thewtex:
CMake version 3.14.3

fbudin69500:
I am still unable to reproduce this. Were you able to reproduce it in a clean, freshly cloned repository?

thewtex:
Thanks for the attempt to reproduce. I am surprised it only occurs on my system. I could reproduce with a clean source tree, fresh configure of ITK v5.0.0.

hjmjohnson:
@fbudin69500 Can this be merged?

fbudin69500:
yes. I could not reproduce the issue, but this patch should not create any issue.

dzenanz:
Merged to master too.

PR #951, opened 2019-05-22
BUG: ImageBase regions are returned as reference

fbudin69500:
ImageBase `LargestPossibleRegion`, `RequestedRegion`, and `BufferedRegion`
are returned as const references. However, it is possible in Python to
change the value they contain, and this can lead to crashes due to
the pipeline not being properly updated once the region is altered inside
the image.

To solve this problem, we hide the original functions that return references
and create new functions that return a copy of the region instead.


fbudin69500:
This PR should address the error reported in the ITKExamples when updated ITK ([here](https://github.com/InsightSoftwareConsortium/ITKExamples/issues/99)).

thewtex:
A rebase should help determine if there are issues -- there are unrelated test failures, but there is one in `itkExtras` that may be related.

fbudin69500:
This is tested [here](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Wrapping/Generators/Python/Tests/extras.py#L111-L114).

fbudin69500:
#962 merged in `release` and `release` merged in `master`

PR #953, opened 2019-05-22
Fix CastImageFilter to support explicit pixel type cast

blowekamp:


dzenanz:
```text
/home/vsts/work/1/s/Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.h:144: error: Template definition (typename) doesn't match regular expression
/home/vsts/work/1/s/Modules/Filtering/ImageFilterBase/test/itkCastImageFilterTest.cxx:17: error: namespace not defined
/home/vsts/work/1/s/Modules/Filtering/ImageFilterBase/test/itkCastImageFilterTest.cxx:44: error: Empty lines exceed 3 (max=2)
```

thewtex:
After confirmed by @romangrothausmann , please create a PR for the `release` branch.

blowekamp:
Thank you for the link to the builds as it's quite helpful. There we can see error messages such as:
```
/opt/itk/include/ITK-5.0/itkCastImageFilter.hxx:127:3: error: static assertion failed: Vector dimensions are required to match!
/opt/itk/include/ITK-5.0/itkCastImageFilter.hxx: In instantiation of 'void itk::CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatched(const OutputImageRegionType&, std::false_type) [with TInputPixelType = itk::RGBAPixel<short int>; TInputImage = itk::Image<itk::RGBAPixel<short int>, 3u>; TOutputImage = itk::Image<itk::Vector<float, 3u>, 3u>; itk::CastImageFilter<TInputImage, TOutputImage>::OutputImageRegionType = itk::ImageRegion<3u>; std::false_type = std::integral_constant<bool, false>]':
```
I do lnke the `static_assert` as it is gives a clear error message.

We can also see the instantiated type is `CastImageFilter<itk::Image<itk::RGBAPixel<short int>, 3u>, Image<itk::Vector<float, 3u>, 3u>` where `RGBAPixel` has 4 components and the `Vector` only has 3, as you described. I'm assuming this use to "work" with the `VectorCastImageFilter`. If we look at it's [functor](https://github.com/InsightSoftwareConsortium/ITK/blob/f69f2f465611afc9a93f34e4db67f127a1a421fc/Modules/Compatibility/Deprecated/include/itkVectorCastImageFilter.h#L60-L70)'s implementation we can see there it no size checking for the input vector. 

Your old usage of `VectorCastImageFilter` had *undefined* behavior related to uninitialized pixel elements and out of bounds access. I would classify this behavior as a bug, at that it has been fixed in ITKv5.

blowekamp:
For you particular problem, Does `VectorCurvatureAnisotropicDiffusionImageFilter` work with `VectorImages`?

romangrothausmann:
> Your old usage of VectorCastImageFilter had undefined behavior related to uninitialized pixel elements and out of bounds access. I would classify this behavior as a bug, at that it has been fixed in ITKv5.

Yes, I gues that is what I vaguely tried to express with "freedom". 

> For you particular problem, Does VectorCurvatureAnisotropicDiffusionImageFilter work with VectorImages?

It does. That is (I think) what is used for processing when the input is an RG image (lacking the blue channel). The idea of the CLIs in question was to have a program to do CurvatureAnisotropicDiffusion no matter whether the input is scalar, vector, or RGB/RGBA image of dimension 2 or 3.

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.h
```diff
@@ -138,6 +138,21 @@ class ITK_TEMPLATE_EXPORT CastImageFilter:
...
+    : std::false_type
+  {};
+
+  template <typename TFromType, typename TToType>
+    struct is_static_castable<TFromType,
+                              TToType,
+                              decltype(static_cast<TToType>(std::declval<TFromType>()))>
+    : std::true_type
+  {};
+
```
blowekamp: @N-Dekker I added you to review due to the C++ meta-programming and type trait usage.

PR #958, opened 2019-05-23
BUG: resample filter no longer triggers unnecessary exception

dzenanz:
If interpolator's `GetRadius` required an image, and it was not set
outside of resample filter, an exception `Input image required!` was raised.

## PR Checklist

- [X] :no_entry_sign: [Makes breaking changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes)
- [X] :no_entry_sign: [Makes design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes)
- [X] :no_entry_sign: Adds the License notice to new files.
- [X] :no_entry_sign: Adds Python wrapping.
- [X] :no_entry_sign: Adds tests and baseline comparison (quantitative).
- [X] :no_entry_sign: [Adds test data](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/UploadBinaryData.md).
- [X] :no_entry_sign: Adds Examples to [ITKExamples](https://github.com/InsightSoftwareConsortium/ITKExamples)
- [X] :no_entry_sign: Adds Documentation.

PR #959, opened 2019-05-23
BUG: resample filter no longer triggers unnecessary exception

dzenanz:
If interpolator's `GetRadius` required an image, and it was not set
outside of resample filter, an exception `Input image required!` was raised.

This is a clone of #958 which goes into `release` branch, slated for 5.0.1.


hjmjohnson:
@dzenanz How is this different from #958 ?

dzenanz:
#958 -> master, #959 (this) -> release

dzenanz:
`master` will eventually become 5.1.0, whereas `release` will become 5.0.1 patch release.

PR #961, opened 2019-05-23
Demonstrate type in DataObject::MakeOutput

blowekamp:
This but was encountered when setting an output to nullptr.

### DIFF
--- a/Modules/Core/Common/src/itkProcessObject.cxx
+++ b/Modules/Core/Common/src/itkProcessObject.cxx
@@ -90,7 +90,8 @@ ProcessObject
     {
     return this->MakeOutput( this->MakeIndexFromOutputName(name) );
     }
-  return static_cast<DataObject *>(DataObject::New().GetPointer());
+
+  return itkDynamicCastInDebugMode<DataObject *>(DataObject::New().GetPointer());
 }
 
 
--- a/Modules/Core/Common/include/itkDataObject.h
+++ b/Modules/Core/Common/include/itkDataObject.h
@@ -33,6 +33,7 @@
 #include "itkSingletonMacro.h"
 #include "itkWeakPointer.h"
 #include "itkRealTimeStamp.h"
+#include "itkObjectFactory.h"
 
 namespace itk
 {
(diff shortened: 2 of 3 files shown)


Modules/Core/Common/src/itkProcessObject.cxx
```diff
@@ -90,6 +90,7 @@ ::MakeOutput( const DataObjectIdentifierType & name )
     {
     return this->MakeOutput( this->MakeIndexFromOutputName(name) );
     }
+  // DataObject::New() returns is from Object, and returns only Object!
   return static_cast<DataObject *>(DataObject::New().GetPointer());
```
blowekamp: Here an invalid upcast occours on a concrete `itk::Object`.

blowekamp:
I'm not sure what to do about this bug I encounter.

Modules/Core/Common/src/itkProcessObject.cxx
```diff
@@ -90,6 +90,7 @@ ::MakeOutput( const DataObjectIdentifierType & name )
     {
     return this->MakeOutput( this->MakeIndexFromOutputName(name) );
     }
+  // DataObject::New() returns is from Object, and returns only Object!
```
dzenanz: itkDataObject.h is missing `itkTypeMacro`.

dzenanz: Oops, it has that but it is missing definition of new.

dzenanz: `Object` has manually defined new, whereas `DataObject` has neither `itkNewMacro` nor manually defined/overridden `New`.

dzenanz: Perhaps both `Object` and `DataObject` should use `itkNewMacro`?

blowekamp: Yup. The questions is: Should` DataObject` be instantiable or an abstract base class?

dzenanz: Is `DataObject` currently instantiable? It looks like it is. So just fix the way it is instantiated? If I am wrong, what are the abstract members?

blowekamp: New method for `DataObject` seems to work just fine. Re: `Object::New` likely has to be in the cxx file due to a circular dependency with using the `ObjectFactory`

PR #952, opened 2019-05-22
WIP: ImageBase regions are returned as reference

fbudin69500:
Python PR for #951

PR #964, opened 2019-05-24
WIP: HoughTransform2D GetLines(), GetCircles() return ListType by-value

N-Dekker:
Changed return types of `HoughTransform2DLinesImageFilter::GetLines()`
and `HoughTransform2DCirclesImageFilter::GetCircles()` from
"by reference" (`ListType &`) to "by value" (`ListType`).

This commit improves encapsulation of the internal data of these
filters, and avoids segmentation faults in Python user code (ITK wrapped
by SWIG), when the result is returned by a Python function, as described
by Nick at the discussion form, 24 May 2019:
https://discourse.itk.org/t/how-to-deal-with-houghtransform2dlinesimagefilter-getlines-method-in-python/1861/16

### DIFF
--- a/Modules/Filtering/ImageFeature/include/itkHoughTransform2DCirclesImageFilter.h
+++ b/Modules/Filtering/ImageFeature/include/itkHoughTransform2DCirclesImageFilter.h
@@ -148,7 +148,7 @@ public:
   * The pixel grid coordinates of the center of a circle from the list can
   * be retrieved by calling circle->GetCenterPoint().
   */
-  CirclesListType & GetCircles();
+  CirclesListType GetCircles();
 
   /** Set/Get the number of circles to extract. */
   itkSetMacro(NumberOfCircles, CirclesListSizeType);
--- a/Modules/Filtering/ImageFeature/include/itkHoughTransform2DCirclesImageFilter.hxx
+++ b/Modules/Filtering/ImageFeature/include/itkHoughTransform2DCirclesImageFilter.hxx
@@ -188,7 +188,7 @@ HoughTransform2DCirclesImageFilter< TInputPixelType, TOutputPixelType, TRadiusPi
 }
 
 template< typename TInputPixelType, typename TOutputPixelType, typename TRadiusPixelType >
-typename HoughTransform2DCirclesImageFilter< TInputPixelType, TOutputPixelType, TRadiusPixelType >::CirclesListType &
+typename HoughTransform2DCirclesImageFilter< TInputPixelType, TOutputPixelType, TRadiusPixelType >::CirclesListType
 HoughTransform2DCirclesImageFilter< TInputPixelType, TOutputPixelType, TRadiusPixelType >
 ::GetCircles()
 {
--- a/Modules/Filtering/ImageFeature/include/itkHoughTransform2DLinesImageFilter.h
+++ b/Modules/Filtering/ImageFeature/include/itkHoughTransform2DLinesImageFilter.h
@@ -132,7 +132,7 @@ public:
   itkGetModifiableObjectMacro( SimplifyAccumulator, OutputImageType );
 
   /** Get the list of lines. This recomputes the lines. */
-  LinesListType & GetLines();
+  LinesListType GetLines();
 
   /** Set/Get the number of lines to extract */
   itkSetMacro( NumberOfLines, LinesListSizeType );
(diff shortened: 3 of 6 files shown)


fbudin69500:
It's interesting that the current returned value is not `const`! I am not against this PR, as it should address the problem that @N-Dekker is talking about, but I would prefer a more general solution to address this problem throughout ITK, if this was possible (See my comment on the discourse forum [here](https://discourse.itk.org/t/how-to-deal-with-houghtransform2dlinesimagefilter-getlines-method-in-python/1861/23?u=fbudin)).

fbudin69500:
One detail about this PR that I think should be addressed is that since this solves a bug, there should be a test to verify that this is the case. @N-Dekker , could you add some code similar to the Python code given by Nick in the discourse forum to ensure that his problem is solved?

N-Dekker:
>  @N-Dekker , could you add some code similar to the Python code given by Nick in the discourse forum to ensure that his problem is solved?

Sorry @fbudin69500 I'm not a regular Python programmer, so I would prefer to stick with C++ for now, and leave writing ITK Python tests to more experienced developers.

I could adapt the commit message from "BUG" to "STYLE", though, as it improves encapsulation of the filter data, which is really a matter of programming style.

fbudin69500:
Do you mind if I add the Python test to this PR? That will update the branch inside your repo if you have enabled that.

N-Dekker:
> Do you mind if I add the Python test to this PR? That will update the branch inside your repo if you have enabled that.

OK: Allowed edits from maintainers.  `:-)`

hjmjohnson:
@fbudin69500 That is a great idea. Thanks!

fbudin69500:
I have tested this PR on my computer and it breaks the Python wrapping. Please do not merge until this problem is solved!

fbudin69500:
I dug more into it and replacing the reference by the actually objects creates more issues. I don't know enough about SWIG to know exactly what happens and how to correct it, but it may not be worth it spending too much time trying to modify the C++ code for now. What we could do it [this PR](https://github.com/InsightSoftwareConsortium/ITK/pull/952/files) works, is use the SWIG macro that I added there to create a copy at the Python wrapping level.

N-Dekker:
> I dug more into it and replacing the reference by the actually objects creates more issues. I don't know enough about SWIG to know exactly what happens and how to correct it, but it may not be worth it spending too much time trying to modify the C++ code for now.

Thanks for dugging into this, Francois @fbudin69500 ! It's a pity that it seems so hard to wrap return-by-value for an `std::list`. It should be possible, at least for `std::vector`: https://stackoverflow.com/questions/46180537/wrapping-return-vectort-on-swig

In my opinion, removing the `&` would improve the C++ style, as it would increase member data encapsulation. (I would find it even better to return `std::vector`, instead of `std::list`, as `std::vector` is more commonly used, more efficiently stored, and offers random access.)

So I do think it's worth figuring out how to return the container by-value. But of course, if you have other priorities, this issue might need to be postponed.

fbudin69500:
@N-Dekker Returning by value is definitely possible, but I think the problem here is as much the returning by value as the smart pointer inside the list. When we keep the code as is it (return by reference), the list of smart pointers gets wrapped and during the wrapping, its output type is modified to return raw pointers instead of smart pointers (with a `typemap(out)`). However, when returning by value, this does not seem to be happening. If we can figure that out, then we can definitely change the API to return values instead of references.

N-Dekker:
> during the wrapping, its output type is modified to return raw pointers instead of smart pointers

Thanks Francois. But that sounds risky to me (as C++ programmer, I don't know SWIG/Python well enough). When the returned pointers are "raw" in Python, will the spatial objects still be kept alive long enough? Will these spatial objects not be destroyed when the filter gets out of scope?

fbudin69500:
@N-Dekker The way it works, at least as far as I understand, is that when an ITK object, typically within a C++ Smart Pointer, is passed to Python, only the raw pointer is passed to Python, and we also give the ownership of the raw pointer to Python. Python is then responsible to manage the memory.
What happens is that when you work in Python, you don't work directly with the raw pointers. You actually work with a Python object that owns the raw pointer. In a sense, one could consider that a "Python Smart Pointer". It is just a wrapper around your C++ data that has extra information for Python, and it allows Python to know how to manage the memory.
If we were working with the Smart Pointers in Python, that would create a lot of complications.

PR #970, opened 2019-05-29
PERF: Replace mutex by atomic in MersenneTwisterRandomVariateGenerator

N-Dekker:
Declared internal integer data members `m_Seed` and `m_StaticDiffer`
`atomic`, and removed three locks (`std::lock_guard<std::mutex>`) from
`MersenneTwisterRandomVariateGenerator`, to improve the performance.

For performance reasons, it is in general recommended to use atomics,
instead of mutex locking, whenever possible.

### DIFF
--- a/Modules/Core/Common/include/itkMersenneTwisterRandomVariateGenerator.h
+++ b/Modules/Core/Common/include/itkMersenneTwisterRandomVariateGenerator.h
@@ -23,8 +23,10 @@
 #include "itkRandomVariateGeneratorBase.h"
 #include "itkIntTypes.h"
 #include "itkMath.h"
-#include <mutex>
 #include "itkSingletonMacro.h"
+
+#include <atomic>
+#include <mutex>
 #include <climits>
 #include <ctime>
 
@@ -277,7 +279,7 @@ protected:
   int          m_Left;
 
   // Seed value
-  IntegerType  m_Seed;
+  std::atomic<IntegerType>  m_Seed;
 
 private:
 
@@ -372,9 +374,7 @@ MersenneTwisterRandomVariateGenerator::SetSeed()
 inline MersenneTwisterRandomVariateGenerator::IntegerType
 MersenneTwisterRandomVariateGenerator::GetSeed()
 {
-  std::lock_guard<std::mutex> mutexHolder(m_InstanceLock);
-  volatile IntegerType seed = this->m_Seed;
-  return seed;
+  return this->m_Seed;
 }
 
 /** Get an integer variate in [0, 2^32-1] */
--- a/Modules/Core/Common/src/itkImageSourceCommon.cxx
+++ b/Modules/Core/Common/src/itkImageSourceCommon.cxx
@@ -19,7 +19,6 @@
 #include "itkImageRegionSplitterSlowDimension.h"
 #include "itkImageSourceCommon.h"
 #include <mutex>
-#include <mutex>
 
 namespace itk
 {
--- a/Modules/Core/Common/src/itkMersenneTwisterRandomVariateGenerator.cxx
+++ b/Modules/Core/Common/src/itkMersenneTwisterRandomVariateGenerator.cxx
@@ -19,6 +19,7 @@
 
 #include "itkSingleton.h"
 
+#include <atomic>
 
 namespace itk
 {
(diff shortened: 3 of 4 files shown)


dzenanz:
There is some error downloading testing data. I hope it will go away later today. @N-Dekker could you also make a WIP PR to python-builds branch after the tests pass here?

N-Dekker:
> could you also make a WIP PR to python-builds branch after the tests pass here?

Thanks for your approval, @dzenanz but I'm not so much into Python. I don't think the PR affects Python bindings anyway, as it just proposes an improvement to the internal implementation of the C++ class. It does not propose any change of the public interface.

If you still think this PR would need extra Python testing, please try!

dzenanz:
I vaguely remember that some global variable requires careful synchronization, and `m_PimplGlobals` is reminiscent of that. @fbudin69500 could tell us more.

N-Dekker:
Just added a _slightly_ related style commit (a no-brainer), hoping to get Azure Pipelines to do its job again. Unfortunately I see, ITK.macOS failing already.

fbudin69500:
@dzenanz : honestly I don't know if the atomic would create a problem with the Python wrapping. We definitely need to carefully think about it and have some Python test that should ensure that the seed is still synchronized. I cannot think of a place in ITK where this specific variable would be modified simultaneously in different ITK Python modules, but if you can find one location where this could happen, this would be a great test!

blowekamp:
> I vaguely remember that some global variable requires careful synchronization, and `m_PimplGlobals` is reminiscent of that. @fbudin69500 could tell us more.

My understanding is that simply the `m_PimplGlobals` being a pointer and the pointer being synchronized solved the non-unique global symbol problem here. The content or usage of the member of m_PimplGlobals does not matter, since if the pointer is managed/synconized properly they are still the same object.

I don't think this should change anything with respect to the Python global issues.

N-Dekker:
Thanks to you all for reviewing this PR so carefully. I do realize that multi-threading code needs that extra bit of care! 

Note that this PR does _not_ remove _all_ the locks from `MersenneTwisterRandomVariateGenerator`. For example, I left the `std::lock_guard` in here:

https://github.com/InsightSoftwareConsortium/ITK/blob/2dbf82c2cf53a4c414fb790f6bbe9e68c02425a4/Modules/Core/Common/include/itkMersenneTwisterRandomVariateGenerator.h#L306 

Only for those cases where the mutex was only protecting the access to a _single_ integer (`m_Seed` or `m_StaticDiffer`), using atomic instead of locking a mutex appeared the right thing to do.

N-Dekker:
https://github.com/InsightSoftwareConsortium/ITK/blob/v5.0.0/Modules/Core/Common/include/itkMersenneTwisterRandomVariateGenerator.h#L38 says:

    It is recommended to create a separate object in each thread. By
    default, each instantiated class will have a different seed created
    by the GetNextSeed method. The creation of the initial seeds are
    initialized once from the time. For deterministic behavior, the
    individual instances' seeds should be manual set to separate
    values in each thread.
  
    It is no longer recommended to use this class using a
    "Singleton-like" GetInstance method for the global instance of this
    class. This usage may result in unsafe concurrent access to
    the global instance.
 
Interesting... in that case I guess both mutex locking and atomics are unnecessary! Right?

dzenanz:
That might be a recommendation, but do not trust the users to follow it. We should still have atomics.

blowekamp:
> It is recommended to create a separate object in each thread. By
> default, each instantiated class will have a different seed created
> by the GetNextSeed method. The creation of the initial seeds are
> initialized once from the time. For deterministic behavior, the
> individual instances' seeds should be manual set to separate
> values in each thread.
> 
> It is no longer recommended to use this class using a
> "Singleton-like" GetInstance method for the global instance of this
> class. This usage may result in unsafe concurrent access to
> the global instance.

I think that comment is totally out of date an incorrect.

There have been a few efforts to make this generator reproducible and random over the years. So the details of the implementation and recommend usage of changed over time and the documentation has not kept up.

N-Dekker:
Thanks for your explanation, Dzenan and Bradley.

@blowekamp wrote:
> I think that comment is totally out of date an incorrect.

Are you sure? Because it looks like you that comment less than two years ago: "DOC: recommend using separate instances", https://github.com/InsightSoftwareConsortium/ITK/commit/3563b8601d442e85a1e9570589e7dc88062079a2  

Anyway, no problem!

blowekamp:
😆 Well... looks like I fixed it.

I wonder if the `GetInstance` method could be deprecated in the future...

PR #978, opened 2019-05-31
ENH: MersenneTwisterRandomVariateGenerator GTest (+ const keyword)

N-Dekker:
`MersenneTwisterRandomVariateGenerator` improvements:
- STYLE: Added `const` to `GetSeed()`
- ENH: Added GoogleTest, testing `GetIntegerVariate()`


N-Dekker:
Very happy to see that the old MersenneTwister `GetIntegerVariate()`, originally added to ITK by Daniel @blezek at 352e56087fadc01f52ac9f076554555a0799cbc0 (Oct 6, 2005), still behaves exactly like a C++11 `std::mt19937` function object. Only the default initial seed is different!

PR #979, opened 2019-05-31
COMP: fix warning about missing override in CastImageFilter

dzenanz:
```text
/Users/builder/externalModules/Filtering/ImageFilterBase/include/itkCastImageFilter.h:128:11: warning: '~CastImageFilter' overrides a destructor but is not marked 'override' [-Winconsistent-missing-destructor-override]
  virtual ~CastImageFilter() = default;
          ^
```
This was introduced in 75d98250d5c87c274c87c9a33c1f442c4addab86.

### DIFF
--- a/Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.h
+++ b/Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.h
@@ -125,7 +125,7 @@ public:
 
 protected:
   CastImageFilter();
-  virtual ~CastImageFilter() = default;
+  ~CastImageFilter() override = default;
 
   void GenerateOutputInformation() override;
 


blowekamp:
`release` likely has this issue too

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.h
```diff
@@ -125,7 +125,7 @@ class ITK_TEMPLATE_EXPORT CastImageFilter:
 
 protected:
   CastImageFilter();
-  virtual ~CastImageFilter() = default;
+  virtual ~CastImageFilter() override = default;
```
hjmjohnson: virtual and override should not both be used.

if override is used, then virtual should be removed.

dzenanz:
Updated. @blowekamp I am already targeting `release` branch with this PR.

dzenanz:
Merged to master too.

PR #984, opened 2019-06-01
ENH: itkObjectToObjectOptimizerBase Must be explicitly instantiated

hjmjohnson:
Move code to cxx file because this class must be explicitly
    instantiated for each valid template type to support
    dynamic casting necessary.


hjmjohnson:
@dzenanz removing the number of file dependancies is important.  The .hxx file should be moved to the src directory.

The main point is to be clear that the definitions are only used in the compilation unit of the cxx file where explicit instantiation occurs.

hjmjohnson:
@dzenanz Any further concerns?

PR #985, opened 2019-06-01
DOC: Improve the documentation clarity for NormalizeImage.

hjmjohnson:
Expand upon the existing caveats to clarify the intended
output values of the image.


blowekamp:
Should it be a warning or error if the output type is an integer?
Should the default output type default to the real pixel type?
This seems more than just a documentation issue to me.

hjmjohnson:
@blowekamp I wanted to keep the scope of this PR small.  If you want to make mandated errors/warnings, I would be supportive.

This is backwards compatible.  What you propose is not, although I agree it is probably good.

PR #991, opened 2019-06-04
COMP: add FindTBB.cmake from VTK

dyollb:
- this allows ITK to find TBB also on Windows (TBBConfig.cmake is not generated on Windows)
- I tried to make it backwards compatible, i.e. use the current mechanism (find TBBConfig.cmake), and only if it fails use the FindTBB.cmake

## PR Checklist

### DIFF
--- a/Modules/ThirdParty/TBB/CMake/FindTBB.cmake
+++ b/Modules/ThirdParty/TBB/CMake/FindTBB.cmake
@@ -1,0 +1,420 @@
+# - Find ThreadingBuildingBlocks include dirs and libraries
+# Use this module by invoking find_package with the form:
+#  find_package(TBB
+#    [REQUIRED]             # Fail with error if TBB is not found
+#    )                      #
+# Once done, this will define
+#
+#  TBB_FOUND - system has TBB
+#  TBB_INCLUDE_DIRS - the TBB include directories
+#  TBB_LIBRARIES - TBB libraries to be lined, doesn't include malloc or
+#                  malloc proxy
+#  TBB::tbb - imported target for the TBB library
+#
+#  TBB_VERSION_MAJOR - Major Product Version Number
+#  TBB_VERSION_MINOR - Minor Product Version Number
+#  TBB_INTERFACE_VERSION - Engineering Focused Version Number
+#  TBB_COMPATIBLE_INTERFACE_VERSION - The oldest major interface version
+#                                     still supported. This uses the engineering
+#                                     focused interface version numbers.
+#
+#  TBB_MALLOC_FOUND - system has TBB malloc library
+#  TBB_MALLOC_INCLUDE_DIRS - the TBB malloc include directories
+#  TBB_MALLOC_LIBRARIES - The TBB malloc libraries to be lined
+#  TBB::malloc - imported target for the TBB malloc library
+#
+#  TBB_MALLOC_PROXY_FOUND - system has TBB malloc proxy library
+#  TBB_MALLOC_PROXY_INCLUDE_DIRS = the TBB malloc proxy include directories
+#  TBB_MALLOC_PROXY_LIBRARIES - The TBB malloc proxy libraries to be lined
+#  TBB::malloc_proxy - imported target for the TBB malloc proxy library
+#
+#
+# This module reads hints about search locations from variables:
+#  ENV TBB_ARCH_PLATFORM - for eg. set it to "mic" for Xeon Phi builds
+#  ENV TBB_ROOT or just TBB_ROOT - root directory of tbb installation
+#  ENV TBB_BUILD_PREFIX - specifies the build prefix for user built tbb
+#                         libraries. Should be specified with ENV TBB_ROOT
+#                         and optionally...
+#  ENV TBB_BUILD_DIR - if build directory is different than ${TBB_ROOT}/build
+#
+#
+# Modified by Robert Maynard from the original OGRE source
+#
+#-------------------------------------------------------------------
+# This file is part of the CMake build system for OGRE
+#     (Object-oriented Graphics Rendering Engine)
+# For the latest info, see http://www.ogre3d.org/
+#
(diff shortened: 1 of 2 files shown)


Modules/ThirdParty/TBB/CMake/FindTBB.cmake
```diff
@@ -0,0 +1,420 @@
...
+    set(COMPILER_PREFIX "vc8")
+  elseif(MSVC_VERSION EQUAL 1500)
+    set(COMPILER_PREFIX "vc9")
+  elseif(MSVC_VERSION EQUAL 1600)
+    set(COMPILER_PREFIX "vc10")
+  elseif(MSVC_VERSION EQUAL 1700)
+    set(COMPILER_PREFIX "vc11")
+  elseif(MSVC_VERSION EQUAL 1800)
+    set(COMPILER_PREFIX "vc12")
+  elseif(MSVC_VERSION EQUAL 1900)
```
dzenanz: VS2019 is missing:
`1910-1919 = VS 15.0 (v141 toolset)`

dzenanz: Missing 1920 (Visual Studio 2019 RTW (16.0)). Both 15.0 and 16.0 use vc14 folder.

dzenanz:
I tried it on my computer, and it either worked or was not triggered due to the existence of `TBBConfig.cmake`.

Hrach97:
Hi,

I am trying to build ITK with TBB.
In contrast of 4.13 version 5.0.0 gives TBB_DIR variable.
Do you have any idea which path I have to give to that variable? 

Thank you,
Hach

dzenanz:
On my computer, `TBB_DIR` is equal to `C:/Libs/tbb2019_20190410oss/cmake`. I am using Intel's binary distribution.

dzenanz:
As [reported](https://discourse.itk.org/t/itk-build-with-tbb/1954/16?u=dzenanz) on the forum, this patch does not work as expected.

Modules/ThirdParty/TBB/CMake/FindTBB.cmake
thewtex: @dyollb or @dzenanz can we please add these to get this in?

dyollb: > @dyollb or @dzenanz can we please add these to get this in?

I could add lines like:

elseif(MSVC_VERSION GREATER 1900 AND MSVC_VERSION LESS 1919)
    set(COMPILER_PREFIX "vc15")
  elseif(MSVC_VERSION EQUAL 1920 OR MSVC_VERSION EQUAL 1921)
    set(COMPILER_PREFIX "vc16")

But the most recent pre-compiled version of TBB does not come with those folders (yet?).

dyollb:
> As [reported](https://discourse.itk.org/t/itk-build-with-tbb/1954/16?u=dzenanz) on the forum, this patch does not work as expected.

Commented in the forum. I don't understand what nm97 means by "run examples". I am happy to spend a bit more time to improve/fix the patch.

dyollb:
> I tried it on my computer, and it either worked or was not triggered due to the existence of `TBBConfig.cmake`.

Initially, I added a message to the FindTBB.cmake, to debug if cmake would enter into that file. It did for my build of TBB.

Modules/ThirdParty/TBB/CMake/FindTBB.cmake
dzenanz: The newer versions all use vc14's compiled libraries, because they are binary compatible with it. So put something like:
```CMake
elseif(MSVC_VERSION GREATER 1900 AND MSVC_VERSION LESS 1920)
  set(COMPILER_PREFIX "vc14")
elseif(MSVC_VERSION EQUAL 1920 OR MSVC_VERSION EQUAL 1921)
  set(COMPILER_PREFIX "vc14")
```
or:
```CMake
elseif(MSVC_VERSION GREATER 1800 AND MSVC_VERSION LESS 1929) 
  set(COMPILER_PREFIX "vc14") # expect all compilers in 192X series to have same binary compatibility as 1920 and 1921.
```

sjh26:
Testing ....

seanm:
CC: @seanm @vfonov 

I stumbled upon this PR trying to figure out how to build ITK with TBB.  I succeeded to build VTK with TBB, I just had to provide a path for each lib and the headers folder.  But for ITK, it seems to want me to point to `TBBConfig.cmake` which doesn't seem to exist when one builds TBB oneself.  I need to build TBB myself because I need to have ASan and TSan versions of it.

Is the purpose of this patch to make ITK work like VTK, where one would just need to point to the libraries?

dzenanz:
If I need to specify path to 10 lib files once, that is fine but not ideal. If I need to do that frequently, it is super annoying.

The purpose of this patch seems to be ability to specify TBB via means other than `TBBConfig.cmake`.

The best solution is to bite the bullet and add code which generates `TBBConfig.cmake` by following the [instructions](https://github.com/intel/tbb/tree/tbb_2019/cmake) provided by Intel.

If you think you can make TBB work without `TBBConfig.cmake`, start by removing `config` from the `find` command [here](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/ThirdParty/TBB/itk-module-init.cmake).

Additional relevant links:
https://github.com/intel/tbb/pull/2
https://github.com/intel/tbb/issues/6
https://github.com/wjakob/tbb/issues/55

blowekamp:
Here is a PR I use to get TBB working for me, which automatically downloaded it: 
https://review.source.kitware.com/%23/c/23436/

As I recall the proposed idea was not general enough to be accepted.

dyollb:
I ended up patching the TBBConfig.cmake to include a tbb prefix. Our build of tbb has a prefix to avoid conflicts with thirdparty libs pulling in a potentially different version of tbb.

PR #997, opened 2019-06-06
Replace calls to FixedArray::Begin() and FixedArray::End()

N-Dekker:
Four STYLE commits that replace calls to `FixedArray::Begin()` and `FixedArray::End()`:

- `Vector(value)`, `CovariantVector(value)` constructors call base class `FixedArray(value)` instead
- `FastMarching` does a range-based for-loop instead
- Const iteration calls `cbegin()` and `cend()` instead
- itkFixedArray.hxx using std::equal, and C++11 fill_n and copy_n

### DIFF
--- a/Modules/Core/Common/include/itkCovariantVector.hxx
+++ b/Modules/Core/Common/include/itkCovariantVector.hxx
@@ -27,11 +27,9 @@ namespace itk
 template< typename T, unsigned int TVectorDimension >
 CovariantVector< T, TVectorDimension >
 ::CovariantVector(const ValueType & r)
+  :
+  Superclass{ r }
 {
-  for ( typename BaseArray::Iterator i = BaseArray::Begin(); i != BaseArray::End(); ++i )
-    {
-    *i = r;
-    }
 }
 
 template< typename T, unsigned int NVectorDimension >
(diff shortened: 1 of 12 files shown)


N-Dekker:
Thanks for your approval, @blowekamp, @hjmjohnson, @dzenanz! Please allow me one more commit for this pull request  `:-)`. In a minute from now!

N-Dekker:
Very small amend of the last commit ("STYLE: itkFixedArray.hxx using std::equal..."), only just to remove `CLANG_SUPPRESS_Wfloat_equal`.

It appears that even when `-Wfloat-equal` warnings are enabled, a call to `std::equal` will not trigger this warning  `:-)` 

    #include <algorithm>

    bool StdEqualFloats(const float f1, const float f2)
    {
        return std::equal(&f1, &f1 + 1, &f2); // No warning!  :-)
    }

    bool EqualFloats(const float f1, const float f2)
    {
        return f1 == f2; // warning: comparing floating point with == or != is unsafe [-Wfloat-equal]
    }

See https://godbolt.org/z/seTRxf

N-Dekker:
I think this PR is ready now  `:-)`

PR #994, opened 2019-06-05
hessianToObjectness strict weak ordering

yjcchen0913:
## PR Checklist

- [X] :no_entry_sign: [Makes breaking changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes)
- [X] :no_entry_sign: [Makes design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes)
- [X] :no_entry_sign: Adds the License notice to new files.
- [X] :no_entry_sign: Adds Python wrapping.
- [X] :no_entry_sign: Adds tests and baseline comparison (quantitative).
- [X] :no_entry_sign: [Adds test data](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/UploadBinaryData.md).
- [X] :no_entry_sign: Adds Examples to [ITKExamples](https://github.com/InsightSoftwareConsortium/ITKExamples)
- [X] :no_entry_sign: Adds Documentation.

Refer to the [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) for
further development details if necessary.

if two eigen value is equal it will return true and cause error "invalid operator<".
Because strict weak ordering of sort.
So i think it should be modified to "return itk::Math::abs(a) '<' itk::Math::abs(b);"

PR #1004, opened 2019-06-10
BUG: Ensure strict weak ordering in HessianToObjectnessMeasure's sort

dzenanz:
This is a squash and rebase of #994.


Modules/Filtering/ImageFeature/include/itkHessianToObjectnessMeasureImageFilter.h
```diff
@@ -140,13 +140,13 @@ class ITK_TEMPLATE_EXPORT HessianToObjectnessMeasureImageFilter:public
 
 private:
   // functor used to sort the eigenvalues are to be sorted
-  // |e1|<=|e2|<=...<=|eN|
```
blowekamp: The original statement is correct. As there can be Eigen values were are the same.

dzenanz: You are right. Amending...

PR #1008, opened 2019-06-10
Add swig include directory to command line

fbudin69500:
This PR may need more testing by I ran into issues when trying to use swig 4.0.0 (current master of the repo) and this patch addressed the problem).


fbudin69500:
@thewtex Pushing this branch here so it does not get lost.

thewtex:
@fbudin69500 thanks!

PR #1010, opened 2019-06-11
STYLE: Add const to SetStartingShrinkFactors parameter

N-Dekker:
Declared the `factors` parameter of `SetStartingShrinkFactors` as
pointer-to-const for both `MultiResolutionPyramidImageFilter` and
`MRIBiasFieldCorrectionFilter`, as these member functions do not
modify their argument.

Adjusted the "single factor overloads" of these member functions, to
make use of the improved const-correctness.


N-Dekker:
Thanks for your approval, @dzenanz  I think this pull request is ready!

PR #1023, opened 2019-06-14
WIP: Add PointSet percentile selection to PointSetToPointSetMetric

fbudin69500:
A percentile option allows to drive the registration using only the
`percentile` points returning the smallest error. This is a reasonable
hypothesis if the moving and fixed point sets are initialized in
a position where corresponding points are close. Setting the
percentile option to `100` returns the same result as the algorithm prior
to this patch would have returned.

### DIFF
--- a/Modules/Registration/Metricsv4/include/itkPointSetToPointSetMetricv4.h
+++ b/Modules/Registration/Metricsv4/include/itkPointSetToPointSetMetricv4.h
@@ -301,6 +301,31 @@ public:
   }
 
   /**
+  * Get/Set the cut off percentile cut off value.
+  */
+  void SetPercentile( unsigned int percentile )
+    {
+    if( percentile > 0 && percentile <= 1000 )
+      {
+      m_Percentile = percentile;
+      }
+    else
+      {
+      itkExceptionMacro( "Percentile value must belong to ]0;100]." )
+      }
+    }
+  itkGetMacro( Percentile, unsigned int );
+
+  /**
+  * Cut off value to filter the number of points used to drive the registration
+  * at each iteration. Points from the fixed point set whose error measurement
+  * are within the given percentile are kept (i.e. if percentile is set to 80,
+  * all the points below the 80th percentile are kept) to compute the error
+  * and the derivative. Other points are ignored.
+  */
+  unsigned int m_Percentile;
+
+  /**
    * By default, the point set metric derivative for a displacement field transform
    * is stored by saving the gradient for every voxel in the displacement field (see
    * the function StorePointDerivative()).  Since the "fixed points" will typically
--- a/Modules/Registration/Metricsv4/include/itkPointSetToPointSetMetricv4.hxx
+++ b/Modules/Registration/Metricsv4/include/itkPointSetToPointSetMetricv4.hxx
@@ -56,6 +56,7 @@ PointSetToPointSetMetricv4<TFixedPointSet, TMovingPointSet, TInternalComputation
   this->m_StoreDerivativeAsSparseFieldForLocalSupportTransforms = true;
 
   this->m_CalculateValueAndDerivativeInTangentSpace = false;
+  this->m_Percentile = 100;
 }
 
 /** Initialize the metric */
@@ -161,8 +162,6 @@ PointSetToPointSetMetricv4<TFixedPointSet, TMovingPointSet, TInternalComputation
 {
   this->InitializeForIteration();
 
-  CompensatedSummation< MeasureType > value;
-
   PointsConstIterator It = this->m_FixedTransformedPointSet->GetPoints()->Begin();
   // Virtual point set will be the same size as fixed point set as long as it's
   // generated from the fixed point set.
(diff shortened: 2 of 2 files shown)


samuelgerber:
One additional option to consider would be using a threshold in lieu of the percentile. For cases in which you know the point sets only have partial overlap and a good initialization this might be more intuitive than selecting a percentile (and it would be marginally faster since you don't need to sort the distances).

samuelgerber:
Maybe consider checking if m_Percentile is set to 100 and in that case not sort the distances.

fbudin69500:
Good ideas. Any takers to implement @samuelgerber 's ideas? :)

Modules/Registration/Metricsv4/include/itkPointSetToPointSetMetricv4.h
```diff
@@ -300,6 +300,31 @@ class ITK_TEMPLATE_EXPORT PointSetToPointSetMetricv4
     return false;
   }
 
+  /**
+  * Get/Set the cut off percentile cut off value.
+  */
+  void SetPercentile( unsigned int percentile )
+    {
+    if( percentile > 0 && percentile <= 1000 )
```
dzenanz: 1000 -> 100

Modules/Registration/Metricsv4/include/itkPointSetToPointSetMetricv4.h
```diff
@@ -300,6 +300,31 @@ class ITK_TEMPLATE_EXPORT PointSetToPointSetMetricv4
...
+      {
+      m_Percentile = percentile;
+      }
+    else
+      {
+      itkExceptionMacro( "Percentile value must belong to ]0;100]." )
```
dzenanz: `]0;100]` -> `(0,100]`

ntustison:
We briefly toyed around with adding something like this in the original implementation but ultimately decided against it.  We felt that, similar to the image analog, any preprocessing or pruning functionality of the input data should be performed outside any metric function.  For example, in considering the image analog, one wouldn't implement cropping within the ImageToImageMetric (which would currently be done prior and with a separate ITK class).  Also, it seems to me that the way that the point set is pruned in this case is based on distance determined on a point-by-point basis which isn't necessarily what you want to do in general for all point-set cases.  I would vote against merging.

samuelgerber:
@ntustison Having it as a preprocessing step would keep the registration framework simpler and apply to all metrics. On the other hand one can argue that it acts similar to setting a mask in the image registration framework. And if it is integrated in the metric the points used are adapted during the optimization which does not seem to be feasible when doing it as a preprocessing step.

samuelgerber:
@ntustison since you're here. I was considering adding a caching option to the metric, such that the nearest point index does not have to be recomputed during every iteration of the optimization. Any thoughts on this?

ntustison:
@samuelgerber  Yeah, I can see the analogy to masking but only in the loose sense of limiting the data considered within the metric.  However, in the context of ITK-usage, we have spatial objects specific for masking and that’s not what is being used here.  This is more akin to a rather blunt heuristic for pruning data.  I think the analogy to masking is also stretched when you consider how the “masking” here is actually being performed.  Instead of limiting the points according to the boundaries of some spatial object, the point-wise distances are being calculated, sorted, and then pruned according to a desired percentage.  I don’t see where this is analogous to the image case.  

The adaptive argument in favor of this change also seems a bit suspect to me.  This pruning process appears to be somewhat computationally redundant as the individual metrics themselves consider point-wise distances at each iteration and essentially limit the influence of points which are far away.  And this is done in a much more systematic and formal manner, as defined by the individual metrics, than is being performed here.  I honestly don’t see how this would be useful and warrant the additional costs of code maintenance and potential feature creep.

Regarding the caching idea—sure, any possibilities for speed-up would be great but I can see where the transform updates at each iteration would require that the nearest point index be updated and I don’t see how this bookkeeping could be performed in a way such that you’d save processing time in the end.

samuelgerber:
@ntustison I think this implementation was mostly targeted at the EuclideanDistancePointMetric in which far away point can drive the optimization. My comment on caching was also with respect to that metric.

While the other metrics might do a good job at weighting far away distances, ultimately they implement a distance between probability measures as far as I understand, and thus consider always the overall match. If one would like to achieve a partial matching they might not be suited for that either, since even if there is a good partial match it might be beneficial to spread the points over a larger region to achieve an overall closer match between the probability distributions.

ntustison:
@samuelgerber  Correct me if I’m misunderstanding what you wrote but I would agree that it’s quite clear that the ICP, or Euclidean distance, metric is particularly susceptible to the presence of outliers which has partially motivated the development of other point set metrics.  If this merge candidate is meant to address that susceptibility, then I would think it should be implemented within that metric.  However, even then I’m highly skeptical that this blunt pruning procedure actually addresses this deficiency or addresses the more complicated issue of partial matching—at least I haven’t seen evidence of such presented in the literature (but am certainly open to reading something on the topic).   Please understand that my only motivation for down-voting something like this is my own subjective simple cost/benefit weighting of the proposed changes (code maintenance and feature creep vs. real world-utility).  I’d certainly go along with the other reviewers in the case of otherwise mutual disagreement.

samuelgerber:
@ntustison I don't have a strong preference. But I think it's an approach worth considering and has been around in the literature for while: https://ieeexplore.ieee.org/abstract/document/1047997

Whether this is better implemented as a pruning step or in the metric is up for debate. I think the possibility to not completely throw out points and allow refinement during the optimization has some merit.

samuelgerber:
@ntustison I guess an alternative middle ground would be to subclass the Euclidean Metric with a Trimmed Euclidean Metric.

ntustison:
@samuelgerber Yeah, that would be a great idea.  Thanks for the link.  I didn't get time to look at it today but will probably tomorrow.

samuelgerber:
Closing this. I will implement a version of this in a remote module here:
https://github.com/samuelgerber/ITKTrimmedPointSetRegistration
If it ends up working well and it can later be integrated.

samuelgerber:
A working implementation is now available at:
https://github.com/samuelgerber/ITKTrimmedPointSetRegistration
an initial simple example suggest it could be quite useful

PR #1030, opened 2019-06-18
Remote module update Python ITK version

thewtex:


Utilities/Maintenance/UpdateRequiredITKVersionInRemoteModules.sh
```diff
@@ -78,11 +78,11 @@ sort -n - | tail -n1)
 git_tag_label='ITKGitTag: '
 
 # Read the ITK git tag information
-curr_git_str=($(grep $git_tag_label $azure_pipelines_ci_filename))
-curr_git_tag=${curr_git_str[1]}
+current_git_str=($(grep $git_tag_label $azure_pipelines_ci_filename))
```
dzenanz: It took me almost a minute to figure out what the change was (`curr` -> `current`).

PR #1032, opened 2019-06-19
ENH: Add ImageRegionRange<TImage> to itk::Experimental namespace

N-Dekker:
Added `ImageRegionRange<TImage>` to the `itk::Experimental` namespace,
supporting modern C++ iteration (bidirectional) over a range of pixels
of an image region.

Replaced the classic `ImageRegionConstIterator<ImageType>` by
`ImageRegionRange<const ImageType>`, when doing bounding box computation
in `ImageMaskSpatialObject`. A performance improvement of 5 % or more
was observed.

### DIFF
--- a/Modules/Core/Common/test/itkImageRegionRangeGTest.cxx
+++ b/Modules/Core/Common/test/itkImageRegionRangeGTest.cxx
@@ -1,0 +1,601 @@
+/*=========================================================================
+ *
+ *  Copyright Insight Software Consortium
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+ // First include the header file to be tested:
+#include "itkImageRegionRange.h"
+
+#include "itkImage.h"
+#include "itkImageBufferRange.h"
+#include "itkImageRegionIterator.h"
+#include "itkMacro.h" // For itkNotUsed.
+#include "itkRGBPixel.h"
+#include "itkRangeGTestUtilities.h"
+#include "itkVectorImage.h"
+
+#include <gtest/gtest.h>
+#include <algorithm>  // For std::reverse_copy, std::equal, etc.
+#include <numeric>  // For std::inner_product
+#include <type_traits>  // For std::is_reference.
+
+// Test template instantiations for various ImageDimension values, and const Image:
+template class itk::Experimental::ImageRegionRange<itk::Image<short, 1>>;
+template class itk::Experimental::ImageRegionRange<itk::Image<short, 2>>;
+template class itk::Experimental::ImageRegionRange<itk::Image<short, 3>>;
+template class itk::Experimental::ImageRegionRange<itk::Image<short, 4>>;
+template class itk::Experimental::ImageRegionRange<const itk::Image<short>>;
+template class itk::Experimental::ImageRegionRange<itk::VectorImage<short>>;
+template class itk::Experimental::ImageRegionRange<const itk::VectorImage<short>>;
+template class itk::Experimental::ImageRegionRange<itk::Image<itk::RGBPixel<std::uint8_t>>>;
+template class itk::Experimental::ImageRegionRange<itk::Image<itk::Vector<long, 11>>>;
+
+using itk::Experimental::ImageRegionRange;
+
+
+namespace
+{
+  // Tells whether or not ImageRegionRange<TImage>::iterator::operator*() returns a reference.
+  // (If it does not return a reference, it actually returns a proxy to the pixel.)
+  template <typename TImage>
+  constexpr bool DoesImageRegionRangeIteratorDereferenceOperatorReturnReference()
+  {
+    using IteratorType = typename ImageRegionRange<TImage>::iterator;
+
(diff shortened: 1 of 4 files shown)


N-Dekker:
While the existing [ImageBufferRange<TImage>](https://itk.org/Doxygen/html/classitk_1_1Experimental_1_1ImageBufferRange.html) iterates over the entire image buffer, this `ImageRegionRange<TImage>` would iterate over the image region, specified by the user. (By default, it would do the _requested region_)

Note: Still under construction!

N-Dekker:
When the region of interest has only a few pixels per row, the proposed `ImageRegionRange` appears much faster than the classic `itk::ImageRegionConstIterator`. For example, with region size [2, 65535], I observed that the proposed range appears more than 10 times (!) as fast as the classic ITK iterator.

When  the region of interest has a large number pixels per row,  the proposed range and the classic ITK iterator appear to have a similar performance.

blowekamp:
@N-Dekker I am really liking these new C++11 interfaces.

Just an FYI, I wrote the[ ImageScanlineIterator](https://itk.org/Doxygen/html/classitk_1_1ImageScanlineConstIterator.html). This iteration patter had two loops, the inner loop which had a trivial check in "IsAtEndOfLine" which was just a comparison. While the "NextLine" operations did the work of moving the iterator through the N dimensions. There was quite a bit of speed up because the interloop could be easily vectorized by the compiler. This is generally my go to iterator for performance.

A filter that may significantly benefit from better iterator is the BilateralImageFilter, I believe 10x practical speed up in the execution is possible.

Modules/Core/Common/test/itkImageRegionRangeGTest.cxx
```diff
@@ -0,0 +1,598 @@
...
+#include <algorithm>  // For std::reverse_copy, std::equal, etc.
+#include <numeric>  // For std::inner_product
+#include <type_traits>  // For std::is_reference.
+
+// Test template instantiations for various ImageDimension values, and const Image:
+template class itk::Experimental::ImageRegionRange<itk::Image<short, 1>>;
+template class itk::Experimental::ImageRegionRange<itk::Image<short, 2>>;
+template class itk::Experimental::ImageRegionRange<itk::Image<short, 3>>;
+template class itk::Experimental::ImageRegionRange<itk::Image<short, 4>>;
+template class itk::Experimental::ImageRegionRange<const itk::Image<short>>;
```
dzenanz: Since you are being this comprehensive, maybe you should also add `RGBPixel<uint8_t>` and `Vector<long, 11>` as pixel types.

N-Dekker: Thanks for the suggestion @dzenanz  I've put it onto my TODO list  `:-)`

N-Dekker:
@blowekamp 
> Just an FYI, I wrote the[ ImageScanlineIterator](https://itk.org/Doxygen/html/classitk_1_1ImageScanlineConstIterator.html). This iteration patter had two loops, the inner loop which had a trivial check in "IsAtEndOfLine" which was just a comparison. While the "NextLine" operations did the work of moving the iterator through the N dimensions. There was quite a bit of speed up because the interloop could be easily vectorized by the compiler. This is generally my go to iterator for performance.

Thanks Bradley! Very interesting! Actually I already noticed that the old `itk::ImageRegionConstIterator` is optimized for iteration within one and the same row (or 'span'): https://github.com/InsightSoftwareConsortium/ITK/blob/v5.0.0/Modules/Core/Common/include/itkImageRegionConstIterator.h#L233

Such an optimization appears very beneficial _in general_, for many use cases. However, when the number of pixels per row is small, it can yield a performance penalty instead! I think that's why the `Experimental::ImageRegionRange` that I'm proposing here is much faster than old `itk::ImageRegionConstIterator`, for a region that has only a few pixels per row.

Modules/Core/Common/test/itkImageRegionRangeGTest.cxx
N-Dekker: @dzenanz 
> maybe you should also add `RGBPixel<uint8_t>` and `Vector<long, 11>` as pixel types.

I just did, please check the latest force-push!

blowekamp:
@N-Dekker Your link is to an "if" statement. Traditionally compiler auto-vectorization has difficulties with `if` statements when trying to vectorize the code because it's trying to do multiple loop passes is a vectorized fashion. That is why the ImageScanlineIterator removes the if statement in the [operator++](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkImageScanlineConstIterator.h#L238), and  [IsAtEndOfLine](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkImageScanlineConstIterator.h#L199), is just a comparison.

One suggestion I believe Hans made, was to extend the "optimized" inner loop to be the maximum continuous region, which could be multiple slices. Food for your thoughts.

N-Dekker:
@blowekamp Just did a preliminary performance comparison between `ImageScanlineConstIterator`, `ImageRegionConstIterator`, and `ImageRegionRange`, on VS2017 Release. For an image region with a small number of pixels per row,  `ImageRegionRange` still appears fastest. But for an image region with a large number of rows, your  `ImageScanlineConstIterator` is the winner  :-) 

Some (preliminary) results:

    Region size [2, 1048575]:
    ImageRegionConstIterator.....  0.0603093 seconds
    ImageRegionRange.............  0.00438488 seconds
    ImageScanlineConstIterator...  0.0342728 seconds
     -- Slowest: ImageRegionConstIterator.....
     -- Fastest: ImageRegionRange.............
    
    Region size [65535, 65535]:
    ImageRegionConstIterator.....  3.51961 seconds
    ImageRegionRange.............  3.39576 seconds
    ImageScanlineConstIterator...  2.19651 seconds
     -- Slowest: ImageRegionConstIterator.....
     -- Fastest: ImageScanlineConstIterator...

thewtex:
@N-Dekker awesome!!

N-Dekker:
For new use cases (user code), I think the choice should be either `Experimental::ImageRegionRange` or `itk::ImageScanlineIterator`. Either of these two always seems a better choice than the old `itk::ImageRegionIterator`. At least, that's what I think...!

PR #1034, opened 2019-06-20
STYLE: Replace std::copy(input, input + n, output) by C++11 std::copy_n

N-Dekker:
Replaced each call of the form `std::copy(input, input + n, output)` by
the equivalent C++11 `std::copy_n(input, n, output)` call. Added missing
`#include <algorithm>` directives.

Does simplify the code, and might in some cases improve the performance,
especially when the original code did two function calls to retrieve the
input iterator, as in `std::copy(f(), f() + n, output)`.

### DIFF
--- a/Modules/Core/Common/include/itkGaussianDerivativeOperator.hxx
+++ b/Modules/Core/Common/include/itkGaussianDerivativeOperator.hxx
@@ -192,7 +192,7 @@ GaussianDerivativeOperator< TPixel, VDimension, TAllocator >
   // Make symmetric
   size_t s = coeff.size() - 1;
   coeff.insert(coeff.begin(), s, 0);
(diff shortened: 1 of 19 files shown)


N-Dekker:
Thanks for your approval, Dzenan and Matt!

@dzenanz wrote:
> Looks good. I think you need to add `#include <algorithm>` somewhere.

Honestly I'm surprised that an extra `#include <algorithm>` appears necessary, as this pull request only replaces existing `std::copy` calls. And `std::copy` is also from `<algorithm>`! Anyway, I just added the missing `#include` to "itkSize.h".

dzenanz:
Now you need to add it someplace else too 😄

N-Dekker:
Ready! `:-)`

PR #1035, opened 2019-06-23
DOC: Update supported Python versions warning

thewtex:


Wrapping/Generators/Python/CMakeLists.txt
```diff
@@ -3,9 +3,13 @@
 find_package(PythonInterp)
 find_package(PythonLibs)
 
-# require at least python 2.6
-if(PYTHON_VERSION_STRING VERSION_LESS 2.6)
-  message(WARNING "Python version less than 2.6: \"${PYTHON_VERSION_STRING}\".")
+# require at least python 2.7
+if(PYTHON_VERSION_STRING VERSION_LESS 2.7)
```
dzenanz: Check order is wrong. If version is 3.3, the warning message will not be displayed because the outer condition checks <=2.7.

dzenanz: (On Windows)

thewtex: Oops -- good catch!

Wrapping/Generators/Python/CMakeLists.txt
```diff
@@ -3,9 +3,11 @@
 find_package(PythonInterp)
 find_package(PythonLibs)
 
-# require at least python 2.6
-if(PYTHON_VERSION_STRING VERSION_LESS 2.6)
-  message(WARNING "Python version less than 2.6: \"${PYTHON_VERSION_STRING}\".")
+# Check for required at least python versions
+if(WIN32 AND PYTHON_VERSION_STRING VERSION_LESS 3.5)
+  message(WARNING "Python versions less than 3.5 are not supported on Windows. Python version: \"${PYTHON_VERSION_STRING}\".")
+elseif(PYTHON_VERSION_STRING VERSION_LESS 2.7)
```
hjmjohnson: Can we start moving towards supporting only python 3.5 and greater with new ITK wrappings in ITKv5?

thewtex: Since it is fairly standard in the Scientific Python community, it seems reasonable for ITK 5.1.0.

hjmjohnson: FYI:  https://pythonclock.org/

thewtex: PR here: https://github.com/InsightSoftwareConsortium/ITK/pull/1040

Discourse thread: https://discourse.itk.org/t/dropping-legacy-python-support-support-for-python-3-5-and-newer-with-itk-5-1/1998

PR #1040, opened 2019-06-26
ENH: Require Python 3.5 or newer on all platforms

thewtex:


thewtex:
Related Discourse thread: https://discourse.itk.org/t/dropping-legacy-python-support-support-for-python-3-5-and-newer-with-itk-5-1/1998

Issue #725, opened 2019-04-12
Deprecate Vector* specializations


dzenanz:
Deprecate Vector* specializations such as `VectorResampleImageFilter` and `VectorLinearInterpolateImageFunction`. The general versions now work with vector types too.

This is not trivial. See this [comment](https://github.com/InsightSoftwareConsortium/ITK/pull/786#issuecomment-488660335).

blowekamp:
There are some filters or algorithms that are specialized just to deal with vector pixel type, while others are specialized to compute different distance, threshold or gradient metrics. Those that treat the vector components independely are likely not to need a specialized implementation. The following is a list of candidates:
- VectorResampleImageFilter
- VectorExpandImageFilter
- VectorLinearInterpolateImageFunction
- VectorCastImageFilter
- VectorCentralDifferenceImageFunction

dzenanz:
Quite some work was done on this already. Is more work expected, or should we close the issue?

blowekamp:
Much of this work was done in : #1054, #900, #856  and others.

This issue has mostly been accomplished and should be closed.

PR #1045, opened 2019-06-28
Deprecate some Vector specializations

maekclena:
Deprecate `VectorExpandImageFilter` and `VectorCentralDifferenceImageFunction` in favor of their generic versions.

According to #725.

### DIFF
--- a/Wrapping/Generators/Python/CMakeLists.txt
+++ b/Wrapping/Generators/Python/CMakeLists.txt
@@ -3,9 +3,11 @@
 find_package(PythonInterp)
 find_package(PythonLibs)
 
-# require at least python 2.6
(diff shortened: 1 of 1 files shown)


maekclena:
It is `VectorCentralDifferenceImageFilter` which is from `ITKReview`. Did you mean to swap the names ?

dzenanz:
@maekclena I think Matt did not confuse the classes. `VectorExpandImageFilter` is in `ITKImageIntensity` and should stay there while ITK major version is 5. But it can/should be deprecated.

`VectorCentralDifferenceImageFilter` is currently in `ITKReview`. `ITKReview` is meant as a temporary place for filters, and filters from it can be moved to other modules at any time. So moving `VectorCentralDifferenceImageFilter` into `ITKDeprecated` is fine.

PR #1046, opened 2019-06-29
STYLE: clang format style enforcement

hjmjohnson:
[X] STYLE: no logic impact (indentation, comments)

# Background
This is primarily the start of a long process to evaluate how to simplify the
burden of code formatting styles.  The goal is to have at least one tool that can automatically
enforce the desired style.

The benefit of choosing clang-format is that the tool:
   -  is robust (does not introduce compiler errors), 
   - is integrated with many IDE's/Editors
   - is used by many other tools


Modules/Bridge/VTK/include/itkVTKImageExport.h
```diff
@@ -64,55 +64,67 @@ class ITK_TEMPLATE_EXPORT VTKImageExport:public VTKImageExportBase
...
+  void
+  PrintSelf( std::ostream & os, Indent indent ) const override;
 
   using InputImagePointer = typename InputImageType::Pointer;
   using InputRegionType = typename InputImageType::RegionType;
   using InputSizeType = typename InputRegionType::SizeType;
   using InputIndexType = typename InputRegionType::IndexType;
 
-  int * WholeExtentCallback() override;
+  int *
```
dzenanz: This kind of blindness is the price to pay for automated consistency. But I am willing to do it in order to have good, automated consistency.

Modules/Bridge/NumPy/include/itkPyBuffer.hxx
```diff
@@ -23,136 +23,137 @@
...
-  Py_buffer                   pyBuffer;
-  memset(&pyBuffer, 0, sizeof(Py_buffer));
+  Py_ssize_t bufferLength;
+  Py_buffer  pyBuffer;
+  memset( &pyBuffer, 0, sizeof( Py_buffer ) );
 
-  SizeType size;
-  SizeType sizeFortran;
+  SizeType      size;
+  SizeType      sizeFortran;
```
dzenanz: Can we have local variable declarations not aligned on the name? And only keep them for class member variable declarations?

Modules/Core/Common/include/itkArray.h
```diff
@@ -154,67 +171,73 @@ class ITK_TEMPLATE_EXPORT Array : public vnl_vector< TValue >
...
+  }
   os << "]";
   return os;
 }
 
 // declaration of specialization
-template<> ITKCommon_EXPORT std::ostream & operator<< <double> (std::ostream & os, const Array< double > & arr);
-template<> ITKCommon_EXPORT std::ostream & operator<< <float> (std::ostream & os, const Array< float > & arr);
+template <>
+ITKCommon_EXPORT std::ostream & operator<<< double >( std::ostream & os, const Array< double > & arr );
```
dzenanz: I guess we plan to manually override `clang-format` in instances like this?

Modules/Bridge/VTK/include/itkVTKImageExport.h
hjmjohnson: https://clang.llvm.org/docs/ClangFormatStyleOptions.html

*** . AlwaysBreakAfterReturnType = All,  but we could go with "top level" or "None"

Modules/Bridge/NumPy/include/itkPyBuffer.hxx
hjmjohnson: https://clang.llvm.org/docs/ClangFormatStyleOptions.html
AlignConsecutiveDeclarations = False

assume 
AlignConsecutiveAssignments = False  also

Modules/Core/Common/include/itkArray.h
hjmjohnson: Why?  

AlwaysBreakTemplateDeclarations = Yes .  -- Perhaps we want "MultiLine"

Modules/Bridge/VTK/include/itkVTKImageExport.h
dzenanz: The issue about whether to put return type definition on a separate line is primarily its length. An `int` or a `void` don't benefit from being put on a separate line, but [`typename MorphologicalContourInterpolator< TImage >::BoolSliceType::Pointer`](https://github.com/KitwareMedical/ITKMorphologicalContourInterpolation/blob/4f4fba267b93af6af379d676160d871ee1fdd3d2/include/itkMorphologicalContourInterpolator.hxx#L259) does.

` clang-format` does not seem to offer the option we like, so we should stick with `All`, which is probably the closest to what we want and already part of ITK's recommended style.

Modules/Bridge/NumPy/include/itkPyBuffer.hxx
dzenanz: There is no separate setting for local variables vs members variables, so we should leave `AlignConsecutiveDeclarations = True`.

Modules/Core/Common/include/itkArray.h
dzenanz: `operator<< <double>` is clearer than `operator<<< double >`, at least to me. At first, the formatted version was confusing - is there an `<<<` operator?

blowekamp:
Just my 2 cents: I am in favor of a relaxed style (comparatively) which only enforces very important things, as opposed to a programmatically mandated style. I just like the programmer to be able to do things like separate arguments on lines when it improves clarity and readability, as opposed to following strict rules. Perhaps just start with some simple things things like indents and curly braces would go a long way to improving the current state of affairs.

I could see header files being more ridging in style requirements than implementation files.

blowekamp:
And of course thanks for improving our tools and easing development practices.

hjmjohnson:
> I just like the program to be able to do things like separate arguments on lines when it improves clarity and readability, as opposed to following strict rules.

Brad,  I used to agree with your statement above, but have recently started to convert ONLY IF THE STRICT RULES can be fully automated.  A flexible choice that requires manual decisions, in my opinion, is less effective than strict rules that can be widely automated.

I have been partially converted based on the doctrine espoused by the python "Black" formatting tool.  Since submitting to "Black" as the authority for formatting in python, and allowing my editors to auto enforce that style, I find that I am saving a lot of time doing house cleaning. I can write python code as a stream-of-algorithmic-conciousness, press [File][Save], and Black reformats the code as needed for style and future readability.  I LOVE IT!

Now... The trick is finding tools, and providing documentation, that can be widely automated.  CLion, Visual studio, vim, emacs all have plugins that automate compliance by relying on clang-format.  I'm working on that this summer. I think I have a "clang-format" based solution that will be no more difficult than the "kwstyle" tests.

thewtex:
@hjmjohnson This is fantastic!

Perhaps more ideas and feedback from a discussion on Discourse would help?

jhlegarreta:
It is great that we get to have further style enforcement/automation. Thanks @hjmjohnson.

As a side-effect, once this gets merged, the related ITK Software Guide sections will need to be updated/adapted :book:.

hjmjohnson:
@blowekamp Thanks for fixing the circle/ci builds

@thewtex What happened to the windows builds?  I see Azure is a mac build,  circleci is a linux build, but there is no windows build.

hjmjohnson:
https://discourse.itk.org/t/update-coding-style-for-itk/2055

thewtex:
@hjmjohnson the Windows build is now there - the GitHub status checks are clobbering each other as *InsightSoftwareConsortium.ITK*. I am working on this and tracking progress in #1102 


Amazing work on this!

phcerdan:
This is great! As a note to @blowekamp comments, clang-format can be switched off by regions with:
```
// clang-format off
... code here ...
// clang-format on
```
for regions that are manually crafted to look good but are disturbed by the clang-format style.

hjmjohnson:
remove commit that makes massive changes

```
    STYLE: Enforce a commmon style based on clang-format
    
    ./Utilities/Maintenance/clang-format.bash \
         --clang-format ../ITK-bld/clang-format-Darwin \
         --tracked
    
    With this change, the ITK prefered format can be defined
    simply the the set of rules defined in the .clang-format
    file with examples of the desired format provided
    by the clang-format version 8.0 documentation.
```

hjmjohnson:
Failing test is unrelated:


The following tests FAILED:
	222 - itkPyBufferMemoryLeakTest (Failed)

PR #1054, opened 2019-07-02
Deprecate some Vector specializations

maekclena:
This is a followup to #1045 which I accidentally closed.


Modules/Filtering/ImageIntensity/include/itkVectorExpandImageFilter.h
```diff
@@ -62,9 +62,10 @@ namespace itk
  * \sa Vector
  * \sa VectorInterpolateImageFunction
  * \sa VectorLinearInterpolationImageFunction
- *
  * \sa ExpandImageFilter
  *
+ * \deprecated Please use ExpandImageFilter instead.
```
dzenanz: Please wrap the whole class in `if ! defined ( ITK_LEGACY_REMOVE )`.

Modules/Filtering/ImageGrid/include/itkExpandImageFilter.h
```diff
@@ -50,15 +50,11 @@ namespace itk
  * This filter is implemented as a multithreaded filter and supports
  * streaming.
  *
- * \warning This filter only works for image with scalar pixel types.
- * For vector images use VectorExpandImageFilter.
- *
```
blowekamp: Do we have a test with Image of Vector's or itk::VectorImage some place? 

I just double checked and SimpleITK is running this filter separately for each component instead of instantiating it with a `VectorImage`.

thewtex: From the diff, it looks like it is used by `itkFEMRegistrationFilter.h`, but I created #1066 to track this.

PR #1057, opened 2019-07-02
Address uninitialized value warnings in Ubuntu 18.04

blowekamp:


Modules/Core/Common/include/itkImageBase.hxx
```diff
@@ -146,9 +144,7 @@ void
 ImageBase< VImageDimension >
 ::SetOrigin(const float origin[VImageDimension])
 {
-  Point< float, VImageDimension > of(origin);
-  PointType                       p;
-  p.CastFrom(of);
+  PointType p(origin);
```
dzenanz: Isn't this the same pattern as old line 138? Does float/double distinction matter? What happens if `ITK_USE_FLOAT_SPACE_PRECISION` is enabled?

blowekamp: Good catch, that can be update too.

With `ITK_USE_FLOAT_SPACE_PRECISION` enabled I get about 50% through the compilation, the some compilation errors in GaussianDerivativeImageFunction which appear to be un-releated. I don't know if this option was compilable before.

blowekamp:
Super annoying!

Modules/Core/Common/include/itkImageBase.hxx
blowekamp: There are ALOT of errors related to the usage of the continuous index and interpolators. This configuration option is broken.

PR #1060, opened 2019-07-03
ENH: Prefer c++11 'using' to 'typedef' for structs

hjmjohnson:
Improve consistency in the code base, and provide modern
type aliasing wiht the 'using' syntax.

### DIFF
--- a/Modules/Core/TestKernel/include/itkTestDriverInclude.h
+++ b/Modules/Core/TestKernel/include/itkTestDriverInclude.h
@@ -73,7 +73,7 @@ std::map< std::string, int > RegressionTestBaselines(char *);
 using ComparePairType = std::pair< char *, char * >;
 
 // A structure to hold regression test parameters
-typedef struct
+using RegressionTestParameters = struct _RegressionTestParameters
 {
   std::vector< ComparePairType > compareList;
   double intensityTolerance;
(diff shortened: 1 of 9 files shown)


Modules/Core/TestKernel/include/itkTestDriverInclude.h
```diff
@@ -108,11 +108,11 @@ struct ProcessedOutputType
 };
 
 // A structure to hold redirect output parameters
-typedef struct
+using RedirectOutputParameters = struct _RedirectOutputParameters
```
dzenanz: Why is additional name `_RedirectOutputParameters` being introduced?

Modules/Core/TestKernel/include/itkTestDriverInclude.h
```diff
@@ -73,7 +73,7 @@ std::map< std::string, int > RegressionTestBaselines(char *);
 using ComparePairType = std::pair< char *, char * >;
 
 // A structure to hold regression test parameters
-typedef struct
+using RegressionTestParameters = struct _RegressionTestParameters
```
dzenanz: Why is additional name `_RegressionTestParameters` being introduced?

hjmjohnson: These were needed to avoid an "anonymous struct" linking issue.  The name was needed in this case to be used ensure that the symbol has a name.

PR #1065, opened 2019-07-06
BUG: Release the Python Global Interpreter Lock (GIL) during execution

thewtex:
Release the Python Global Interpreter Lock (GIL) so multi-threaded
Python applications, like Dask, can run in parallel while ITK does heavy
operations.

For more information, see section 32.13 Support for Multithreaded
Applications at

 http://www.swig.org/Doc4.0/Python.html#Python

Also, set `nothreadallow` as suggested in the comments
on https://github.com/swig/swig/issues/927. This avoids calls to
`PyEval_SaveThread()` and `PyEval_RestoreThread()`, which can negatively
effect performance. This change only affects module C++ code -- code
that can call Python code like ITKPyUtils does not have `threads`
enabled.

### DIFF
--- a/Wrapping/Generators/Python/PyUtils/itkPyCommand.cxx
+++ b/Wrapping/Generators/Python/PyUtils/itkPyCommand.cxx
@@ -18,6 +18,26 @@
 
 #include "itkPyCommand.h"
 
+namespace
+{
+// Wrapper to automatics obtain and release GIL
+// RAII idiom
+class PyGILStateEnsure
+{
+public:
+  PyGILStateEnsure()
+    {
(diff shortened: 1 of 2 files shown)


blowekamp:
The GIL should be reacquired when calling Python call backs. See SimpleITK for an example:
https://github.com/SimpleITK/SimpleITK/blob/master/Wrapping/Python/sitkPyCommand.cxx

thewtex:
@blowekamp thanks for the pointer! Added in https://github.com/InsightSoftwareConsortium/ITK/pull/1065/commits/3eb1f6a6e59344a756dece628833b0214145057b

blowekamp:
LGTM

PR #1070, opened 2019-07-07
BUG: Remove VectorExpandImageFilter wrapping

thewtex:
This class has been deprecated in favor of ExpandImageFilter.


blowekamp:
nice. should ExpandImageFilter be instantiated with any vector like types to replace the functionality?

thewtex:
> nice. should ExpandImageFilter be instantiated with any vector like types to replace the functionality?

Good idea! Added in #1076

PR #1079, opened 2019-07-09
BUG: Fix COMPILE_DEFINITIONS of castxml

phcerdan:
itk_wrap_module_castxml now gets the COMPILE_DEFINITIONS of the
module top directory, instead of the wrapping directory.

The old behavior wasn't respecting the add_compile_definitions
set in the CMakeLists of the module.

Triggered by:
https://github.com/InsightSoftwareConsortium/ITKTotalVariation/pull/5

This is a follow on from: e3a908b49b569be0690227614f8e230262ac7bc2

### DIFF
--- a/Wrapping/Generators/CastXML/CMakeLists.txt
+++ b/Wrapping/Generators/CastXML/CMakeLists.txt
@@ -290,7 +290,16 @@ macro(itk_wrap_module_castxml library_name)
   get_directory_property(include_dir_list INCLUDE_DIRECTORIES)
   list(REMOVE_DUPLICATES include_dir_list)
 
+  # Get the compile_definitions of the module added with add_compile_definitions
+  # From the wrapping folder (current)
   get_directory_property(compile_definition_list COMPILE_DEFINITIONS)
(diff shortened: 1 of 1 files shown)


Wrapping/Generators/CastXML/CMakeLists.txt
```diff
@@ -290,7 +290,11 @@ macro(itk_wrap_module_castxml library_name)
   get_directory_property(include_dir_list INCLUDE_DIRECTORIES)
   list(REMOVE_DUPLICATES include_dir_list)
 
-  get_directory_property(compile_definition_list COMPILE_DEFINITIONS)
```
thewtex: Can we still keep this compile definition list around, then append the definitions for `${WRAPPER_LIBRARY_SOURCE_DIR}/..`? These compile definitions may be needed for another use case.

phcerdan: sure!

bpaniagua:
@phcerdan @thewtex 
What is left to do here? Anything I can do to help?

PR #1087, opened 2019-07-13
BUG: Fix possible loss of data on type conversion

N-Dekker:
Fix possible loss of data on type conversion, typically from a (possibly
64-bit) `size_t` to a 32-bit integer type, as suggested by various
occurrences of VS2017 warning C4244 and C4267 (when these warnings are
not disabled):

> video\core\include\itkRingBuffer.hxx(193): warning C4244: 'initializing': conversion from 'itk::RingBuffer<itk::DataObject>::OffsetValueType' to 'unsigned int'

> Filtering\BiasCorrection\src\itkCompositeValleyFunction.cxx(26): warning C4267: 'initializing': conversion from 'size_t' to 'unsigned int'

> Numerics\Polynomials\src\itkMultivariateLegendrePolynomial.cxx(151): warning C4267: 'argument': conversion from 'size_t' to 'int'

> Core\Common\src\itkArrayOutputSpecialization.cxx(29): warning C4267: 'initializing': conversion from 'size_t' to 'unsigned int'

> core\common\include\itkNumericTraitsArrayPixel.h(126): warning C4267: 'return': conversion from 'size_t' to 'unsigned int'

> numerics\statistics\include\itkMeasurementVectorTraits.h(308): warning C4267: 'return': conversion from 'size_t' to 'itk::Statistics::MeasurementVectorTraits::MeasurementVectorLength'

> IO\MeshBase\include\itkMeshIOBase.h(386): warning C4244: 'return': conversion from 'const itk::MeshIOBase::SizeValueType' to 'unsigned int'

### DIFF
--- a/Modules/IO/MeshBase/include/itkMeshIOBase.h
+++ b/Modules/IO/MeshBase/include/itkMeshIOBase.h
@@ -734,7 +734,7 @@ protected:
   unsigned int m_NumberOfCellPixelComponents{0};
 
   /** The number of independent dimensions in the point. */
-  SizeValueType m_PointDimension{3};
+  unsigned int m_PointDimension{3};
 
   /** The number of points and cells */
   SizeValueType m_NumberOfPoints;
--- a/Modules/Core/Common/include/itkNumericTraitsArrayPixel.h
+++ b/Modules/Core/Common/include/itkNumericTraitsArrayPixel.h
@@ -121,7 +121,7 @@ public:
   }
 
   /** Get the length of the input array. */
-  static unsigned int GetLength(const Array< T > & m)
+  static std::size_t GetLength(const Array< T > & m)
   {
     return m.GetSize();
   }
--- a/Modules/Core/Common/src/itkArrayOutputSpecialization.cxx
+++ b/Modules/Core/Common/src/itkArrayOutputSpecialization.cxx
@@ -26,11 +26,11 @@ std::ostream & operator<< <double> (std::ostream & os, const Array< double > & a
 {
   NumberToString<double> convert;
   os << "[";
-  const unsigned int length = arr.size();
+  const std::size_t length = arr.size();
   if ( length >= 1 )
     {
-    const unsigned int   last   = length - 1;
-    for ( unsigned int i = 0; i < last; ++i )
+    const std::size_t last = length - 1;
+    for ( std::size_t i = 0; i < last; ++i )
       {
       os << convert(arr[i]) << ", ";
       }
@@ -45,11 +45,11 @@ std::ostream & operator<< <float> (std::ostream & os, const Array< float > & arr
 {
   NumberToString<float> convert;
   os << "[";
-  const unsigned int length = arr.size();
+  const std::size_t length = arr.size();
   if ( length >= 1 )
     {
-    const unsigned int   last   = length - 1;
-    for ( unsigned int i = 0; i < last; ++i )
+    const std::size_t last = length - 1;
+    for ( std::size_t i = 0; i < last; ++i )
       {
       os << convert(static_cast<float>(arr[i])) << ", ";
       }
--- a/Modules/Filtering/BiasCorrection/src/itkCompositeValleyFunction.cxx
+++ b/Modules/Filtering/BiasCorrection/src/itkCompositeValleyFunction.cxx
@@ -23,7 +23,7 @@ CompositeValleyFunction
 ::CompositeValleyFunction(const MeasureArrayType & classMeans,
                           const MeasureArrayType & classSigmas)
 {
-  unsigned int length = classMeans.size();
+  const std::size_t length = classMeans.size();
 
   if ( length != classSigmas.size() )
     {
@@ -41,7 +41,7 @@ CompositeValleyFunction
     throw ex;
     }
 
-  for ( unsigned int i = 0; i < length; i++ )
+  for ( std::size_t i = 0; i < length; i++ )
     {
     this->AddNewClass(classMeans[i], classSigmas[i]);
     }
--- a/Modules/Numerics/Polynomials/include/itkMultivariateLegendrePolynomial.h
+++ b/Modules/Numerics/Polynomials/include/itkMultivariateLegendrePolynomial.h
@@ -127,14 +127,15 @@ public:
   class CoefficientVectorSizeMismatch
   {
 public:
-    CoefficientVectorSizeMismatch(int given, int required)
-    {
-      m_Required = required;
-      m_Given = given;
-    }
+  CoefficientVectorSizeMismatch(const std::size_t given, const std::size_t required)
+    :
+    m_Required{ required },
+    m_Given{ given }
+  {
+  }
 
-    int m_Required;
-    int m_Given;
+    std::size_t m_Required;
+    std::size_t m_Given;
   };
 
   /** \brief Sets the Legendre polynomials' parameters.
--- a/Modules/Numerics/Polynomials/src/itkMultivariateLegendrePolynomial.cxx
+++ b/Modules/Numerics/Polynomials/src/itkMultivariateLegendrePolynomial.cxx
@@ -125,8 +125,8 @@ void MultivariateLegendrePolynomial
 {
   if ( coefficients.size() != m_NumberOfCoefficients )
     {
-    throw CoefficientVectorSizeMismatch(static_cast<int>( coefficients.size() ),
-                                        static_cast<int>( m_NumberOfCoefficients) );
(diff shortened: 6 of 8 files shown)


Modules/IO/MeshBase/include/itkMeshIOBase.h
```diff
@@ -382,8 +382,8 @@ class ITKIOMeshBase_EXPORT MeshIOBase:public LightProcessObject
   itkGetConstMacro(NumberOfPointPixelComponents, unsigned int);
   itkSetMacro(NumberOfCellPixelComponents, unsigned int);
   itkGetConstMacro(NumberOfCellPixelComponents, unsigned int);
-  itkSetMacro(PointDimension, unsigned int);
-  itkGetConstMacro(PointDimension, unsigned int);
+  itkSetMacro(PointDimension, SizeValueType);
```
dzenanz: I was going to suggest to use `SizeValueType` instead of `size_t`, because that what these sort of types are usually called in ITK. But now I see you also `SizeValueType` here.

What happens with warning if you turn `ITK_USE_64BITS_IDS` off? Is there a difference before/after?

N-Dekker: @dzenanz Thanks for your thoughtful review! Yes I did use ITK's `SizeValueType` for `PointDimension`, because the corresponding data member `m_PointDimension` is actually of type `SizeValueType`! Still I wonder: is there any fundamental difference between `itk::SizeValueType` and `std::size_t`?

> What happens with warning if you turn ITK_USE_64BITS_IDS off?

Haven't checked yet, but honestly my main goal is to fully support the 64-bit address space.  So I'm focusing now on building ITK with `ITK_USE_64BITS_IDS=ON`, and `sizeof(void*)` is 8 bytes. However, I certainly do not want to break any 32-bit user code!

Note that those MSVC "possible loss of data" warnings are disabled now on various places in ITK. So I had to manually "enable" them, locally.

This pull request only fixes a few of those warnings. There are many more! But I thought I'd better first see your reactions on this PR...!

dzenanz: `SizeValueType` usually boils down to `size_t` on Windows, and `unsigned long` on Linux. See [itkIntTypes.h](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkIntTypes.h). But on 32-bit Windows `SizeValueType` boils down to `unsigned long`, which is essentially the same as `unsigned int`. Hence the recommendation to check 32-bit compilation.

N-Dekker: Thanks @dzenanz Is it safe to assume `sizeof(itk::SizeValueType) == sizeof(std::size_t)`?

dzenanz: Not always. But for 64-bit compilation and `ITK_USE_64BITS_IDS=ON`, which is the most common, yes.

N-Dekker: Thanks again, @dzenanz  Personally I think it's complicated enough already to support multiple platforms, with different sizes for `long` and `size_t`. I would be in favor of requiring that `itk::SizeValueType` has the same size as `std::size_t`. Ideally I would think: `using SizeValueType = std::size_t`. Unless there is a very strong use case to have it differently!

I feel that the current complexity makes it very hard to fully support the 64-bit address space. I guess that's why warnings about "loss of data" have just been disabled.

By the way, my interest in this issue was triggered by a post at the [elastix](https://github.com/SuperElastix/elastix) forum, [Elastix crashes without any notice](https://groups.google.com/forum/#!category-topic/elastix-imageregistration/elastix/Q1e1b2UNIQI), by Philipp Heinrich. So users actually want to go beyond the 32 bits...!

Do you also think it would be a good idea to require that `itk::SizeValueType` has the same size as `std::size_t`?

dzenanz: Yes, I think 32-bit platforms have lost relevance about 10 years ago, at least in medical imaging field. And we could go the route of having SizeValueType = size_t. But I think you will find out that it will require quite a bit of refactoring to get right and flush out all the warnings. If you decide to go that way, I am willing to help you. But we should probably have approval, or at least absence of objections from @thewtex, @hjmjohnson and @blowekamp.

blowekamp: It is relatively recently that for windows 64 that ITK_USE64_BIT_IDS is ON by default. This was done for compatibility because a lot of code was written with the unsigned long type. 

This issue should be discussed further on discourse.

hjmjohnson: I am very much in support of having "having SizeValueType = size_t".

dzenanz: Niels, can you start a discourse discussion about the proposed reorganization of various `typedef`s? Matt should be back from SciPy tomorrow, so he will have a chance to pitch in into the discussion.

N-Dekker: @blowekamp 
> It is relatively recently that for windows 64 that ITK_USE64_BIT_IDS is ON by default.

Thanks, Bradley! Just for the record, this appears to be the commit that did it:

Revision: 8bb8c743267ab0a0ac4e7b14f1ec2baf4a0fdc80
Author: Bradley Lowekamp
Date: March 3, 2016
ENH: Use ITK_USE_64BITS_IDS for windows 64 by default

> Support 4GB+ images and data structures by default on Windows 64.

> Note: ITK_USE_64BITS_IDS still defaults to being off on non-windows
systems which continues the use of type "unsigned long" for sizes as
uint64_t and "unsigned long" may not be the same type one these
systems.

Modules/IO/MeshBase/include/itkMeshIOBase.h
```diff
@@ -382,8 +382,8 @@ class ITKIOMeshBase_EXPORT MeshIOBase:public LightProcessObject
...
+  itkGetConstMacro(PointDimension, SizeValueType);
```
blowekamp: For this particular one, I'm not sure this should be changed. For things link number of image dimensions and number of components we are still using a hard coded "unsigned int" type. This appear to be the same kind of type from reading the documentation: https://itk.org/Doxygen/html/classitk_1_1MeshIOBase.html#aa8cbbad10bc2781fd77cdc118ae79da8

N-Dekker: @blowekamp Thanks for your reply. But `MeshIOBase::m_PointDimension` itself is defined as `SizeValueType`! At https://github.com/InsightSoftwareConsortium/ITK/blob/v5.0.0/Modules/IO/MeshBase/include/itkMeshIOBase.h#L737

      SizeValueType m_PointDimension{3};

Modules/IO/MeshBase/include/itkMeshIOBase.h
N-Dekker: @dzenanz wrote:
> Niels, can you start a discourse discussion about the proposed reorganization of various `typedef`s?

 Please have a look: [Could ITK drop ITK_USE_64BITS_IDS and just do SizeValueType = std::size_t?](https://discourse.itk.org/t/could-itk-drop-itk-use-64bits-ids-and-just-do-sizevaluetype-std-size-t/2053)

Modules/IO/MeshBase/include/itkMeshIOBase.h
blowekamp: Then perhaps the correct fix is to change the data type of the IVAR? It may be less disruptive to the API and class users.

N-Dekker: @blowekamp The last force-push I did yesterday includes your suggestion: it fixes a "loss of data warning" by declaring `m_PointDimension` as `unsigned int`, instead of `SizeValueType`. OK?

blowekamp: Great! I think we are just waiting on getting the windows CI back now...

N-Dekker:
After this commit, there are still 225 other C4267 warnings, and 2390 other C4244 warnings about "possible loss of data" in ITK. Some of them are about conversion from 64-bit integer to 32-bit integer, others are about floating point conversions. But they are too many for me to solve them all in one commit. Could this pull request already be merged, as an initial step?

blowekamp:
> others are about floating point conversions. 

I recall the implicit conversion from double to float being warned was a bit annoyance for a while long ago, that may have been part of the motivation for this suppression.

Issue #1050, opened 2019-07-01
Pickling the itk module fails


jakirkham:
### Description

Pickling of the `itk` module fails.

### Steps to Reproduce

```python
import pickle
import itk                                                           

pickle.dumps(itk)
```

### Expected behavior

Pickling succeeds.

### Actual behavior

```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-8fd95ab0c545> in <module>
----> 1 pickle.dumps(itk)

TypeError: can't pickle LazyITKModule objects
```

### Reproducibility

Every time.

### Versions

ITK 5.0.0 with some modules as 5.0.0.post1.

```
$ conda list itk
...
# Name                    Version                   Build  Channel
itk                       5.0.0                    py37_1    conda-forge
itk-core                  5.0.0.post1              pypi_0    pypi
itk-filtering             5.0.0.post1              pypi_0    pypi
itk-numerics              5.0.0.post1              pypi_0    pypi
itk-registration          5.0.0.post1              pypi_0    pypi
itk-segmentation          5.0.0.post1              pypi_0    pypi
```

### Environment

Ubuntu 16.04.5 LTS (Xenial)
Python  3.7.3

thewtex:
xref #1051

thewtex:
Pickling via cloudpickle addressed in #1090 . We will not be able to pickle with the `pickle` module because it does not support pickling modules.

PR #1090, opened 2019-07-14
BUG: Support pickling LazyITKModule with cloudpickle

thewtex:
cloudpickle is used by Dask for transmitting the state of a Python
program.

Since LazyITKModule extends types.ModuleType, a few extras are required
to support its pickling.

__reduce_ex__ is re-implemented because modules using implement this
protocol, and initialization arguments are not the same as
LazyITKModule.

__getstate__ and __setstate__ are implemented to re-constitute the
LazyITKModule after it is de-serialized. Like cloudpickle's
implementation for loading dynamically created modules and imported
modules:

  https://github.com/cloudpipe/cloudpickle/blob/2db58f11991fb655e5b188d385633d63cbfd2841/cloudpickle/cloudpickle.py#L1134-L1143

We restore the LazyITKModule attributes (which can include other
LazyITKModules's, e.g. `itk.ITKCommon` (`ITKCommon` is also an
ITKLazyModule), and we reload the modules on disk with
`itkBase.Module.LoadModule`.

Addresses #1050 @jakirkham @mrocklin

### DIFF
--- a/Wrapping/Generators/Python/Tests/lazy.py
+++ b/Wrapping/Generators/Python/Tests/lazy.py
@@ -21,3 +21,11 @@ import itk
 assert(itk.__package__ == 'itk')
 from itk import ITKCommon
 assert(ITKCommon.__package__ == 'itk')
+
+# Test pickling used bash Dask
+try:
+    import cloudpickle
+    itkpickled = cloudpickle.dumps(itk)
+    cloudpickle.loads(itkpickled)
+except ImportError:
+    pass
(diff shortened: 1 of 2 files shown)


thewtex:
CircleCI failures unrelated per #1095

jakirkham:
Thanks for doing this, @thewtex. Are there nightlies with this change that we could try?

thewtex:
There are some [Linux nightlies available](https://github.com/InsightSoftwareConsortium/ITKPythonPackage/releases/tag/latest), which can be installed with

```
python -m pip install --upgrade pip numpy
python -m pip install itk --upgrade --no-index \
  -f https://github.com/InsightSoftwareConsortium/ITKPythonPackage/releases/tag/latest
```

but ITK 5.0.1 will be released over the next few days.

PR #1093, opened 2019-07-15
ENH: updating remote modules

dzenanz:
## PR Checklist

### DIFF
--- a/Modules/Remote/AnalyzeObjectMapIO.remote.cmake
+++ b/Modules/Remote/AnalyzeObjectMapIO.remote.cmake
@@ -1,5 +1,5 @@
 itk_fetch_module(AnalyzeObjectMapIO
   "AnalyzeObjectMapIO plugin for ITK. From Insight Journal article with handle: https://hdl.handle.net/1926/593"
   GIT_REPOSITORY ${git_protocol}://github.com/InsightSoftwareConsortium/itkAnalyzeObjectMap.git
-  GIT_TAG 54c3e2dbf0b4259aabe6c430df75171469d7bfee
+  GIT_TAG 7388d866433896368bd3049c396b974433b22cc5
   )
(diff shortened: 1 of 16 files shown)


Modules/Remote/RTK.remote.cmake
```diff
@@ -2,5 +2,5 @@
 itk_fetch_module(RTK
     "Reconstruction Toolkit (RTK) http://www.openrtk.org/"
   GIT_REPOSITORY ${git_protocol}://github.com/SimonRit/RTK.git
-  GIT_TAG v2.0.1
+  GIT_TAG 99a6626f42e6654d11a780127610c95d3c1a063e
```
jhlegarreta: `RTK` is one of the few modules, if not the only one, that is tagged consistentl and uses these instead of commit hashes to indicate the appropriate version to be used within ITK.

This is a case that is not currently handled properly by the update script.

dzenanz: I will undo this change, then.

dzenanz:
We added backwards-compatible testing macros without `ITK_` prefix under legacy. So all the remotes should be working with common configuration options.

dzenanz:
It is now easy to run the script [UpdateRemoteModules.sh](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/Maintenance/UpdateRemoteModules.sh). My motivation was [these outdated warnings](https://open.cdash.org/viewBuildError.php?type=1&buildid=6004976) in Montage module. Whenever some improvement is made in remote modules, the script could be run and PR created.

Modules/Remote/RTK.remote.cmake
SimonRit: I admit I'm not following this conversation. Is that fine if I tag RTK before an ITK release after checking that everything compiles ok? How do you handle patch release for remote modules, can I have a non-patch release for RTK included in a patch release of ITK?

dzenanz: Remote modules are far less strictly controlled than main ITK repository. Most remote modules don't have tags and versions, and we just take the latest commit in master branch in most cases.

PR #1099, opened 2019-07-16
BUG: Specify specific CircleCI docker image with platform

blowekamp:


blowekamp:
I am presuming the reason for the compiler change is that the docker image used got bumped from Debian's jessie to buster. The compiler or code issue will still exit.

blowekamp:
No compiler errors but 44 warnings:
https://open.cdash.org/viewBuildError.php?type=1&buildid=6007474

blowekamp:
@N-Dekker  Any insight into this warning:
Modules/Core/Common/include/itkImageRegionRange.h:401:32: warning: missing initializer for member 'itk::Offset<3u>::m_InternalArray' [-Wmissing-field-initializers]

The class `itk::Offset` is an aggregate type with a C array. I presume there is not actual initialization going on with the class when a universal initializer `{}` is used?

N-Dekker:
@blowekamp 
> Modules/Core/Common/include/itkImageRegionRange.h:401:32: warning: missing initializer for member 'itk::Offset<3u>::m_InternalArray' [-Wmissing-field-initializers]
> 
> The class `itk::Offset` is an aggregate type with a C array. I presume there is not actual initialization going on with the class when a universal initializer `{}` is used

The code does correctly initialize the entire aggregate, the way it is meant in C++11:

    OffsetType iterationOffset{};  // All offset values zero.

However, I did (again) forget that some old compiler versions like it better with an extra pair of curly braces:

    OffsetType iterationOffset{ {} };

The warning appears with GCC 4.9.4, but no longer with GCC 5.1

Issue #1102, opened 2019-07-16
Azure Pipelines builds lost GitHub repository access


thewtex:
The Azure Pipelines builds have been loosing their ITK repository GitHub access permissions for unknown reasons over over the last week.

I am working on replacing the access tokens and builds.

PR #1103, opened 2019-07-16
DOC: Update Azure Pipelines badge URL's

thewtex:
For unknown reasons, Azure Pipelines lost GitHub access credentials.
The builds have been re-configured, but the badge URL's need to be
updated.

Windows Python builds have not lost access yet -- if they do, it will be
addressed in another patch.

xref #1102


hjmjohnson:
@thewtex Why is the mac python build named differently?  I expected ITK.macosx.Python for the name.

thewtex:
All builds have now been updated.

PR #1104, opened 2019-07-17
COMP: Prefer anonymous namespace over static template function

hjmjohnson:
Avoid:
warning: unused function template 'itkReadRawBytesAfterSwappingMacro' [-Wunused-template]

### DIFF
--- a/Modules/IO/ImageBase/src/itkRawImageIOUtilities.cxx
+++ b/Modules/IO/ImageBase/src/itkRawImageIOUtilities.cxx
@@ -29,7 +29,7 @@ _WriteRawBytesAfterSwappingUtility( const void * buffer, std::ofstream & file, i
 
   using InternalByteSwapperType = itk::ByteSwapper< TStrongType >;
   const itk::SizeValueType numberOfPixels = numberOfBytes / ( sizeof( TStrongType ) );
-  if ( byteOrder == itk::ImageIOBase::LittleEndian )
+  if ( byteOrder == itk::ImageIOBase::LittleEndian && InternalByteSwapperType::SystemIsBigEndian() )
   {
     TStrongType * tempBuffer = new TStrongType[numberOfPixels];
     memcpy( (char *)tempBuffer, buffer, numberOfBytes );
@@ -37,7 +37,7 @@ _WriteRawBytesAfterSwappingUtility( const void * buffer, std::ofstream & file, i
     file.write( (char *)tempBuffer, numberOfBytes );
     delete[] tempBuffer;
   }
-  else if ( byteOrder == itk::ImageIOBase::BigEndian )
+  else if ( byteOrder == itk::ImageIOBase::BigEndian && InternalByteSwapperType::SystemIsLittleEndian() )
   {
     TStrongType * tempBuffer = new TStrongType[numberOfPixels];
     memcpy( (char *)tempBuffer, buffer, numberOfBytes );
--- a/Modules/IO/ImageBase/src/itkRawImageIOUtilities.cxx
+++ b/Modules/IO/ImageBase/src/itkRawImageIOUtilities.cxx
@@ -1,0 +1,194 @@
+/*=========================================================================
+ *
+ *  Copyright Insight Software Consortium
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+#include "itkImageIOBase.h"
+#include "itkByteSwapper.h"
+
+namespace
+{
+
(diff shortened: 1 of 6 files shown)


hjmjohnson:
@dzenanz This is my attempt to fix the new compiler warning.

Modules/IO/RAW/include/itkRawImageIO.hxx
```diff
@@ -102,7 +150,7 @@ void RawImageIO< TPixel, VImageDimension >
 }
 
 template < typename TStrongType >
-static void
+void
 itkWriteRawBytesAfterSwappingMacro( const void * buffer, std::ofstream & file, ImageIOBase::ByteOrder byteOrder,
```
dzenanz: The existing functions are not removed?

hjmjohnson:
@N-Dekker Thank you. That is very helpful.  I'm working on a more complete solution.  I was trying to work around the need for a cxx as the RAW image filter previously did not have any non-templated components.  I am now making a RAWIO library.

blowekamp:
As I recall you can make this type of template function `inline` and the warning will go away too.

dzenanz:
If you are making non-trivial changes, you might eliminate the need for a temporary buffer when the current and target endianness are the same. Currently, a temporary buffer is created no matter whether we are writing in little endian or big endian order.

N-Dekker:
By the way, `itkReadRawBytesAfterSwappingMacro` and `itkWriteRawBytesAfterSwappingMacro` don't look like macro's to me, but I guess you noticed that as well...  `:-)`

hjmjohnson:
> By the way, `itkReadRawBytesAfterSwappingMacro` and `itkWriteRawBytesAfterSwappingMacro` don't look like macro's to me, but I guess you noticed that as well... `:-)`

Changed.  I was minimizing the differences from the previous code, but at this point that changes are vastly different anyway.

hjmjohnson:
> If you are making non-trivial changes, you might eliminate the need for a temporary buffer when the current and target endianness are the same. Currently, a temporary buffer is created no matter whether we are writing in little endian or big endian order.

Done.

Modules/IO/RAW/include/itkRawImageIO.hxx
```diff
@@ -21,8 +21,27 @@
 #include "itkRawImageIO.h"
 #include "itkIntTypes.h"
 
+
 namespace itk
 {
+
+/** Utility function for writing RAW bytes */
+extern void
+WriteRawBytesAfterSwapping( itk::ImageIOBase::IOComponentType componentType,
```
dzenanz: Isn't this identical re-declaration from .h file?

Modules/IO/RAW/include/itkRawImageIO.hxx
```diff
@@ -21,8 +21,27 @@
...
+    const void * buffer,
+    std::ofstream & file,
+    itk::ImageIOBase::ByteOrder byteOrder,
+    itk::SizeValueType numberOfBytes,
+    itk::SizeValueType numberOfComponents );
+
+/** Utility function for reading RAW bytes */
+extern void
+ReadRawBytesAfterSwapping( itk::ImageIOBase::IOComponentType componentType,
```
dzenanz: re-declaration from .h file?

Modules/IO/ImageBase/src/itkRawImageIOUtilities.cxx
```diff
@@ -29,15 +29,15 @@ _WriteRawBytesAfterSwappingUtility( const void * buffer, std::ofstream & file, i
 
   using InternalByteSwapperType = itk::ByteSwapper< TStrongType >;
   const itk::SizeValueType numberOfPixels = numberOfBytes / ( sizeof( TStrongType ) );
-  if ( byteOrder == itk::ImageIOBase::LittleEndian )
+  if ( byteOrder == itk::ImageIOBase::LittleEndian && InternalByteSwapperType::SystemIsBigEndian() )
   {
     TStrongType * tempBuffer = new TStrongType[numberOfPixels];
     memcpy( (char *)tempBuffer, buffer, numberOfBytes );
     InternalByteSwapperType::SwapRangeFromSystemToLittleEndian( (TStrongType *)tempBuffer, numberOfComponents );
     file.write( (char *)tempBuffer, numberOfBytes );
```
dzenanz: f

Modules/IO/RAW/include/itkRawImageIO.hxx
hjmjohnson: Removed from .h file.  I put these extern declarations in the .hxx file as they are not intended to be used outside of this class.

hjmjohnson:
Added windows export options for shared library builds.

dzenanz:
The following tests FAILED:
	2257 - itkTimeVaryingBSplineVelocityFieldPointSetRegistrationTest (SEGFAULT)

hjmjohnson:
> The following tests FAILED:
> 2257 - itkTimeVaryingBSplineVelocityFieldPointSetRegistrationTest (SEGFAULT)

I'm looking into it. That is very strange. I am very surprised as I would not think that test has anything to do with RAWIO.

dzenanz:
The same test is segfaulting in my PR #1105, but my PR could plausibly interact with it 😄

hjmjohnson:
> itkTimeVaryingBSplineVelocityFieldPointSetRegistrationTest

FYI the exact same hash value does not fail on my computer.:

1/1 Test #2331: itkTimeVaryingBSplineVelocityFieldPointSetRegistrationTest ...   Passed    1.55 sec

The following tests passed:
	itkTimeVaryingBSplineVelocityFieldPointSetRegistrationTest

PR #1108, opened 2019-07-19
Fix a few MSVC Warning Level 4 warnings

N-Dekker:
Two commits that fix a few MSVC Warning Level 4 warnings.

### DIFF
--- a/Modules/IO/ImageBase/test/itkUnicodeIOTest.cxx
+++ b/Modules/IO/ImageBase/test/itkUnicodeIOTest.cxx
@@ -39,8 +39,8 @@ bool checkAlphaExists()
   FILE * tmp = _wfopen(wstr.c_str(), L"r");
 #else
   std::string utf8_str;
-  utf8_str.append(1, (char)(0xCE));
-  utf8_str.append(1, (char)(0xB1));
+  utf8_str.append(1, '\xCE');
+  utf8_str.append(1, '\xB1');
   utf8_str += ".txt";
   FILE * tmp = fopen(utf8_str.c_str(), "r");
 #endif
@@ -65,8 +65,8 @@ bool removeAlpha()
   return (_wunlink(wstr.c_str()) != -1);
 #else
   std::string utf8_str;
-  utf8_str.append(1,(char)(0xCE));
-  utf8_str.append(1,(char)(0xB1));
+  utf8_str.append(1, '\xCE');
+  utf8_str.append(1, '\xB1');
   utf8_str += ".txt";
   return (unlink(utf8_str.c_str()) != -1);
 #endif
@@ -84,8 +84,8 @@ int main( int , char * [] )
 
   // Put alpha.txt encoded in utf8 within a std::string
   std::string utf8_str;
-  utf8_str.append(1, (char)(0xCE));
-  utf8_str.append(1, (char)(0xB1));
+  utf8_str.append(1, '\xCE');
+  utf8_str.append(1, '\xB1');
   utf8_str += ".txt";
 
   // Check if we actually find it is a valid string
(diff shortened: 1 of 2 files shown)


Modules/IO/ImageBase/test/itkUnicodeIOTest.cxx
```diff
@@ -84,8 +84,8 @@ int main( int , char * [] )
 
   // Put alpha.txt encoded in utf8 within a std::string
   std::string utf8_str;
-  utf8_str.append(1, (char)(0xCE));
-  utf8_str.append(1, (char)(0xB1));
+  utf8_str.append(1, static_cast<char>(std::uint8_t{ 0xCE }));
+  utf8_str.append(1, static_cast<char>(std::uint8_t{ 0xB1 }));
```
blowekamp: This looks like is should work, but there has to be a more concise where to do this!

N-Dekker: @blowekamp 
> This looks like is should work, but there has to be a more concise where to do this!

Hmmm... maybe somethis like this:

    utf8_str.append(1, '\xCE');
    utf8_str.append(1, '\xB1');

What do you think?

blowekamp: That is what I would like to think would work. Hard to say if a compiler would warn. 🤷‍♂

hjmjohnson: https://en.cppreference.com/w/cpp/language/string_literal

Do we need a prefix on the string literal?

N-Dekker: @hjmjohnson Thanks for the link, Hans, but I don't think such a prefix is necessary. The '\xnn' escape sequence appears to do the job: https://en.cppreference.com/w/cpp/language/escape

N-Dekker: @blowekamp 
> That is what I would like to think would work. Hard to say if a compiler would warn. 

I just checked at https://rextester.com/VMDIR6746 with Warning Level 4 (`/W4`):

    str.append(1, (char)0xCE);  // <== warning: cast truncates constant value

    str.append(1, '\xCE');  // OK, no warning  :-)

blowekamp: Great work figuring out these details! That link was very helpful Hans. I believe the original code was declaring a signed integer the casting it to char. The single quote syntax directly defines a char type!

N-Dekker: I just checked: `0xCE` is equal to `206`, while the integer value of `(char)0xCE` is `-50`, by default, for MSVC (having signed `char` by default). But yes, `(char)0xCE == \xCE` is true.

Modules/IO/ImageBase/test/itkUnicodeIOTest.cxx
```diff
@@ -39,8 +39,8 @@ bool checkAlphaExists()
   FILE * tmp = _wfopen(wstr.c_str(), L"r");
 #else
   std::string utf8_str;
-  utf8_str.append(1, (char)(0xCE));
-  utf8_str.append(1, (char)(0xB1));
+  utf8_str.append(1, '\xCE');
+  utf8_str.append(1, '\xB1');
   utf8_str += ".txt";
   FILE * tmp = fopen(utf8_str.c_str(), "r");
```
N-Dekker: For the record: Instead of appending each char, it could have been: `std::string utf8_str = "\xCE\xB1.txt"`.

dzenanz: Excellent excuse for another PR!

N-Dekker: @dzenanz 
> Excellent excuse for another PR!

Well, you may be right, but I don't want to spend too much time on itkUnicodeIOTest now. It does not seem to be very relevant to the projects I'm currently working on, here at the Leiden University Medical Center. I'm just happy that it compiles silently now at VS2017 Warning Level 4.

So I just wanted to note that the characters of a literal string may also be specified by hexadecimal numbers!

PR #1110, opened 2019-07-19
ENH: documenting supported compilers

dzenanz:
## PR Checklist

- [X] Adds Documentation.

### DIFF
--- a/Documentation/SupportedCompilers.md
+++ b/Documentation/SupportedCompilers.md
@@ -1,0 +1,21 @@
+ITK requires a compiler with C++11 support.
+
+# Visual Studio
+* VS2013 and earlier: **NOT supported**
+* MSVC toolset v140 (first shipped with VS2015): fully supported
+* MSVC toolset v141 (first shipped with VS2017): fully supported
+* MSVC toolset v142 (first shipped with VS2019): supported with some [caveats](https://discourse.itk.org/t/has-anyone-tried-to-build-visual-studio-c-2019/1561/6?u=dzenanz)
+
+# GNU Compile Collection
+* GCC 4.8.1 and later [should be supported](https://www.gnu.org/software/gcc/projects/cxx-status.html).
+
+# LLVM Clang
+* Clang 3.3 and later [should be supported](https://clang.llvm.org/cxx_status.html)
+
+# Intel C++ Compiler
+* Intel C++ 15.0.4 and later
+
+# Future versions
+Future versions of these compilers should be backwards compatible, and thus ITK is expected to work with them.
+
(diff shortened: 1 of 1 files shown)


Documentation/SupportedCompilers.md
```diff
@@ -0,0 +1,21 @@
+ITK requires a compiler with C++11 support.
+
+# Visual Studio
+* MSVC 12 (VS2013) and earlier: **NOT supported**
+* MSVC 14 (VS2015): fully supported
+* MSVC 15 (VS2017): fully supported
+* MSVC 16 (VS2019): supported with some [caveats](https://discourse.itk.org/t/has-anyone-tried-to-build-visual-studio-c-2019/1561/6?u=dzenanz)
```
blowekamp: This does not appear to be consistent with this:
https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B#Internal_version_numbering

Also please keep in mind a given Visual Studio can have multiple tool sets installed now.

lassoan:
If we keep build instructions in the source code repository then the developer automatically gets the appropriate build instructions automatically. Browsing various versions of the same file is very easy on github web GUI.

Summary/comparison tables give a nice overview but if you need to maintain it manually (in addition to the version-specific build instructions) then it is probably not worth it. Things change too frequently and there are always some small but important details.

I agree that we should move away from the habit of referring to Visual Studio compiler version as VS20xx, as it has become ambiguous since Microsoft decoupled Visual Studio IDE and build toolset. We should mainly use v140, v141, v142 version IDs for identifying the compiler.

lassoan:
I would consider this document to contain the official specification, so we should not require the user to look up limitations or special instructions on discourse but we should have all details here (we can still link to the discourse post for further reading).

Otherwise, this all looks good. Thank you very much for working on this!

jhlegarreta:
It also dawned on me that users could assume that what was supported for a release version would be the version of this file pointed by the tag, so that would be easy to find out.

Let's see how things work out.

dzenanz:
Thanks for feedback everybody! I updated Visual Studio version specifications.

@jhlegarreta A table of supported compilers for previous versions would be handy, but super hard to make. I guess we could switch to the new system of having a list of supported compilers in source code, and relying on git log/blame for history. But it will only be valid from v5 onwards.

Changing target branch from master to release, so we get this information into v5.0.1 too.

hjmjohnson:
CI Failure on mac environment.  I think the CI Environment needs to be updated to python 3.5.

``` 
-- Found PythonLibs: /usr/lib/libpython2.7.dylib (found version "2.7.10") 
CMake Warning at Wrapping/Generators/Python/CMakeLists.txt:16 (message):
  Python executable ("2.7.16") and library ("2.7.10") version mismatch.

```

PR #1126, opened 2019-07-31
BUG: Remove Windows Python 2.7 hypot workaround

thewtex:
This was a workaround for a bug with the compiler required to build
Python 2.7 on Windows and Python 2.7. Since we do not support Python 2.7
anymore, it is no longer required.


dzenanz:
This seems to be a duplicate of #1067.

thewtex:
Superceded by #1161

PR #1127, opened 2019-07-31
BUG: Use ITK_BUILD_DOCUMENTATION to enable Doxygen builds

thewtex:
Using this specific CMake variable prevents conflicts when built with
other packages. For example, this prevents conflicts with the
SphinxExamples remote module.


thewtex:
`itkTimeVaryingBSplineVelocityFieldPointSetRegistrationTest` false positives :-(

Issue #1131, opened 2019-08-01
Flaky itkTimeVaryingBSplineVelocityFieldPointSetRegistrationTest


dzenanz:
I caught `itkTimeVaryingBSplineVelocityFieldPointSetRegistrationTest` crashing in [itkPointSetToPointSetMetricv4.hxx, line 347](https://github.com/InsightSoftwareConsortium/ITK/blob/c376cf6bd17b07e730ee1f146c9e06846dc6aaa7/Modules/Registration/Metricsv4/include/itkPointSetToPointSetMetricv4.hxx#L347). This was most recently modified by @samuelgerber in 687c49176ba97fbbe805fc468bd96ec36b46ffc0.

dzenanz:
I will look into this a bit more, but ideas are welcome.

samuelgerber:
I think I see the issue. I can fix it.

Needs another variable of different dimension to store intermediate results before jacobian transform.

dzenanz:
The failure is non-deterministic, but relatively frequent at about 10-20%.

samuelgerber:
I put a fix here:
https://github.com/InsightSoftwareConsortium/ITK/compare/master...samuelgerber:FixPointSetMetricv4?expand=1

I did not open a pull request yet since i think there is a logic / math error somewhere as well when computing the derivative in tangent space.

hjmjohnson:
Failing test persists here.

dzenanz:
Offending changes reverted by #1139.

PR #1138, opened 2019-08-05
BUG: PointSetToPointSetMetricv4 Wrong Iteration Bounds

samuelgerber:
Fixes https://github.com/InsightSoftwareConsortium/ITK/issues/1131

The iteration bounds in https://github.com/InsightSoftwareConsortium/ITK/blob/c376cf6bd17b07e730ee1f146c9e06846dc6aaa7/Modules/Registration/Metricsv4/include/itkPointSetToPointSetMetricv4.hxx#L347 were over the wrong dimension.

It appears though that there might be a logic error as well and the calculation of the derivative in tangent space should not be allowed depending on parameters of the transformation?

## PR Checklist

### DIFF
--- a/Modules/Registration/Metricsv4/include/itkPointSetToPointSetMetricv4.hxx
+++ b/Modules/Registration/Metricsv4/include/itkPointSetToPointSetMetricv4.hxx
@@ -255,6 +255,7 @@ void
 PointSetToPointSetMetricv4<TFixedPointSet, TMovingPointSet, TInternalComputationValueType>
 ::CalculateValueAndDerivative( MeasureType & calculatedValue, DerivativeType & derivative, bool calculateValue ) const
 {
+
   this->InitializeForIteration();
 
   // Virtual point set will be the same size as fixed point set as long as it's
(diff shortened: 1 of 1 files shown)


hjmjohnson:
Failing test persists.  	

2345 - itkTimeVaryingBSplineVelocityFieldPointSetRegistrationTest (SEGFAULT)

Modules/Registration/Metricsv4/include/itkPointSetToPointSetMetricv4.hxx
```diff
@@ -342,7 +342,7 @@ PointSetToPointSetMetricv4<TFixedPointSet, TMovingPointSet, TInternalComputation
 
       if( this->m_CalculateValueAndDerivativeInTangentSpace )
         {
-        for( DimensionType d = 0; d < PointDimension; ++d )
+        for( DimensionType d = 0; d < this->GetNumberOfLocalParameters(); ++d )
```
dzenanz: If these were different, that would always fail, not just sporadically.

samuelgerber: I thought the behavior is undefined if you write out of bounds? Depending on what you end up writing to it might or might not crash?

hjmjohnson: I agree with @samuelgerber .  What is the size of "pointDerivative" that is indexed by 'd'?

dzenanz: Writing out of bounds usually gets reported in debug mode. I will run this in debugger to double check.

dzenanz: threadLocalTransformDerivative: size 2
pointDerivative: size 2
PointDimension: size 2

samuelgerber: Thanks @dzenanz, now I can stop barking up the wrong tree :)

Modules/Registration/Metricsv4/include/itkPointSetToPointSetMetricv4.hxx
```diff
@@ -276,24 +277,29 @@ PointSetToPointSetMetricv4<TFixedPointSet, TMovingPointSet, TInternalComputation
...
+
+  //Thread lambda for accumlating derivatives
   std::function< void(unsigned int) > sumNeighborhoodValues =
-      [ this, &derivative, &threadDerivatives, &threadValues, &ranges, &calculateValue,
-        &virtualTransformedPointSet, &fixedTransformedPointSet]
+      [ this, &derivative, &threadDerivatives, &threadValues, &ranges, &calculateValue]
       (unsigned int rangeIndex)
     {
+
+    //Use STL container to make sure no unesecarry checks are performed
```
hjmjohnson: unesecarry -> unnecessary

dzenanz:
itkTimeVaryingBSplineVelocityFieldPointSetRegistrationTest is still failing on CircleCI.

hjmjohnson:
any progress on this?

samuelgerber:
I was on vacation but looking at it a bit more now. 
It appears that the segfault only happens when there is more than one thread. This should narrow the potential sources of errors down a bit ( and has me a bit confused).

samuelgerber:
Maybe it is this:
The method GetNumberOfLocalParameters in CompositeTransform is not thread safe:
If a thread updates m_NumberOfLocalParameters the time is updated before the variable is updatde potentially returjning a wrong value to a thread concurrently accessing it:
https://github.com/InsightSoftwareConsortium/ITK/blob/ed41bcc9d091d13841bfc211b35d84119bf2f848/Modules/Core/Transform/include/itkCompositeTransform.hxx#L935

dzenanz:
That sound like just the right kind of explanation! So move `GetNumberOfLocalParameters` before the multi-threaded section?

samuelgerber:
I would fix CompositeTransfrom by moving:
  this->m_LocalParametersUpdateTime = this->GetMTime();
to after
   this->m_NumberOfLocalParameters = result;

dzenanz:
Fixing `CompositeTransfrom` to be thread-safe would be better, but it might be a bit trickier than just moving that assignment.

samuelgerber:
I think that should fix it, but might cause a minor unnecessary re-computations of m_NumberOfLocalParameters

samuelgerber:
Well I guess one could change the number of transforms or the number of parameters of a transform (which won't be captured in this method either way) during the summation but that seems a very unlikely scenario.

samuelgerber:
Any preferences on the way i fix this?
Also should I start from the pointset registration modification branch (https://github.com/InsightSoftwareConsortium/ITK/commit/687c49176ba97fbbe805fc468bd96ec36b46ffc0) and add the fix there?

hjmjohnson:
> Any preferences on the way i fix this?
> Also should I start from the pointset registration modification branch ([687c491](https://github.com/InsightSoftwareConsortium/ITK/commit/687c49176ba97fbbe805fc468bd96ec36b46ffc0)) and add the fix there?

I think you should revert the revert (i.e. restore the code that was failing a week ago).  Then continue until the true problem is resolved completely.

dzenanz:
Superseded by #1166.

PR #1140, opened 2019-08-05
Fix library dependancies

hjmjohnson:
Small changes to fix library dependancies and consistency.

### DIFF
--- a/Modules/Filtering/Convolution/itk-module.cmake
+++ b/Modules/Filtering/Convolution/itk-module.cmake
@@ -3,7 +3,8 @@ with a kernel. Convolution is a fundamental operation in many image
 analysis algorithms.")
 
 itk_module(ITKConvolution
-  COMPILE_DEPENDS
+  ENABLE_SHARED
+  DEPENDS
     ITKFFT
     ITKImageIntensity
     ITKThresholding


Modules/Filtering/Convolution/CMakeLists.txt
```diff
@@ -1,3 +1,3 @@
-project(ITK-Convolution)
-
+project(ITKConvolution)
+set(ITKConvolution_LIBRARIES ITKConvolution)
```
dzenanz: This should probably be removed, as Convolution is a header-only module.

hjmjohnson: This will be needed when the next PR with the enum classes is added.  The enum classes print function needs to be in a library.   I'm just trying to split compound patch sets apart into simpler pieces.

hjmjohnson: I will submit just the 1 patch with the isolated FFT library fix, and then move the ITKConvolution patch back to the enum-c PR.

hjmjohnson:
Failed: itkTimeVaryingBSplineVelocityFieldPointSetRegistrationTest

PR #1150, opened 2019-08-09
BUG: offset was not updated after changing Euler angle order (ZYX)

### LINKED ISSUE #1148: itkCenteredEuler3DTransform fails when applying SetComputeZYX(true) after parameters are set
### Description

 itkCenteredEuler3DTransform is supposed to rotate (around a user specified center) and translate an element in 3D space. When the transformation parameters are set via SetParameters function, the rotation matrix and the offset are calculated following ZXY order (unless previously ZYX order was specified). If SetComputeZYX(true) is applied, the rotation matrix is computed again, but not the offset. The offset in this class depends on the rotation matrix, so when matrix gets updated, the offset should do so as well. If SetComputeZYX(true) is applied before SetParameters function, the rotation matrix and the offset are calculated correctly because when SetParameters is called,  the offset is computed with the coherent rotation matrix (ZYX). If computations are required in ZXY order, there is no problem, as the matrix is calculated this way by default. The problem arises when SetParameters function is called before SetComputeZYX(true).

### Steps to Reproduce

In order to test this bug, a point in space was transformed using itkCenteredEuler3DTransform. The transformation parameters consisted only in 3 rotation parameters (no translation was applied) and the rotation center was set to the point that was transformed.

Code: 

```C++
#include <cstdlib>

#include <itkCenteredEuler3DTransform.h>
#include <itkPoint.h>

int main()
{
	const double dtr = (std::atan(1.0) * 4.0) / 180.0; // cast angles from degrees to radians

	using CenteredEulerTransformType = itk::CenteredEuler3DTransform<double>;
	CenteredEulerTransformType::Pointer centeredtransform = CenteredEulerTransformType::New();
	CenteredEulerTransformType::ParametersType transformParameters = centeredtransform->GetParameters();
	transformParameters[0] = 32;
	transformParameters[1] = -51;
	transformParameters[2] = 17;
	transformParameters[3] = 0;
	transformParameters[4] = 0;
	transformParameters[5] = 0;

	itk::Point<double> rotationCenter; // center of rotation
	rotationCenter[0] = 14.26;
	rotationCenter[1] = -32.819;
	rotationCenter[2] = 48.019;

	CenteredEulerTransformType::ParametersType parameters(9);
	parameters[0] = transformParameters[0] * dtr;
	parameters[1] = transformParameters[1] * dtr;
	parameters[2] = transformParameters[2] * dtr;
	parameters[3] = rotationCenter[0];
	parameters[4] = rotationCenter[1];
	parameters[5] = rotationCenter[2];
	parameters[6] = transformParameters[3];
	parameters[7] = transformParameters[4];
	parameters[8] = transformParameters[5];
	centeredtransform->SetParameters(parameters);
	centeredtransform->SetComputeZYX(true);

	itk::Point<double> testPoint;
	testPoint[0] = rotationCenter[0];
	testPoint[1] = rotationCenter[1];
	testPoint[2] = rotationCenter[2];

	itk::Point<double> outputTestPoint = centeredtransform->TransformPoint(testPoint);

	double computeError = (outputTestPoint[0] - testPoint[0]) * (outputTestPoint[0] - testPoint[0]) +
		(outputTestPoint[1] - testPoint[1]) * (outputTestPoint[1] - testPoint[1]) +
		(outputTestPoint[2] - testPoint[2]) * (outputTestPoint[2] - testPoint[2]);
	computeError = std::sqrt(computeError);

	double errorTolerance = 0.001;

	if (computeError < errorTolerance)
	{
		std::cout << "The point is transformed correctly!!\n";
	}
	else
	{
		std::cout << "There was an error transforming the point!!\n";
	}
}
```

### Expected behavior

The output point should the identical (or really similar) to the input one as the transformation being applied is a rotation through the transformed point itself, with no translation.

### Actual behavior

The result of the transformation is a point far from the theoretical one. The reason is that rotation matrix is computed in SetParameters in ZXY order and recomputed in SetComputeZYX in ZYX order. The offset however, is calculated in ZXY order in SetParameters but is not recomputed coherently in SetComputeZYX.

### Versions

This error was tested in ITK 4.13 and in ITK 5.0 and both showed the same error. 

### Environment

I am working in Windows 10 and Visual Studio 2017. 

### Additional Information

To execute this test correctly, SetComputeZYX should be changed to false, or should be executed before SetParameters.
This could be corrected in two different ways:
1- The offset gets updated when calling SetComputeZYX(true) coherently with the rotation matrix.
2- Make SetComputeZYX update nothing, but user should be "forced" to call this function before calling SetParameters function.

dzenanz:
Closes #1148.

## PR Checklist

- [X] Adds tests and baseline comparison (quantitative).

### DIFF
--- a/Modules/Core/Transform/test/itkCenteredEuler3DTransformTest.cxx
+++ b/Modules/Core/Transform/test/itkCenteredEuler3DTransformTest.cxx
@@ -335,6 +335,55 @@ int itkCenteredEuler3DTransformTest(int, char *[] )
     return EXIT_FAILURE;
     }
 
+  std::cout << "Testing offset updating after changing angle order (ZYX) : ";
+  const double dtr = (std::atan(1.0) * 4.0) / 180.0; // cast angles from degrees to radians
+
+  EulerTransformType::Pointer centeredtransform = EulerTransformType::New();
+  EulerTransformType::ParametersType transformParameters = centeredtransform->GetParameters();
+  transformParameters[0] = 32;
+  transformParameters[1] = -51;
+  transformParameters[2] = 17;
+  transformParameters[3] = 0;
+  transformParameters[4] = 0;
+  transformParameters[5] = 0;
+
+  itk::Point<double> rotationCenter; // center of rotation
+  rotationCenter[0] = 14.26;
+  rotationCenter[1] = -32.819;
+  rotationCenter[2] = 48.019;
+
+  parameters[0] = transformParameters[0] * dtr;
+  parameters[1] = transformParameters[1] * dtr;
+  parameters[2] = transformParameters[2] * dtr;
+  parameters[3] = rotationCenter[0];
+  parameters[4] = rotationCenter[1];
+  parameters[5] = rotationCenter[2];
+  parameters[6] = transformParameters[3];
+  parameters[7] = transformParameters[4];
+  parameters[8] = transformParameters[5];
+  centeredtransform->SetParameters(parameters);
+  centeredtransform->SetComputeZYX(true);
+
+  itk::Point<double> testPoint;
+  testPoint[0] = rotationCenter[0];
+  testPoint[1] = rotationCenter[1];
+  testPoint[2] = rotationCenter[2];
+
+  itk::Point<double> outputTestPoint = centeredtransform->TransformPoint(testPoint);
+
+  double computeError = (outputTestPoint[0] - testPoint[0]) * (outputTestPoint[0] - testPoint[0]) +
+      (outputTestPoint[1] - testPoint[1]) * (outputTestPoint[1] - testPoint[1]) +
+      (outputTestPoint[2] - testPoint[2]) * (outputTestPoint[2] - testPoint[2]);
+  computeError = std::sqrt(computeError);
+
+  double errorTolerance = 0.001;
+  if (computeError > errorTolerance)
+    {
+    std::cout << " [ FAILED ] " << std::endl;
+    return EXIT_FAILURE;
+    }
+  std::cout << " [ PASSED ] " << std::endl;
+
   return EXIT_SUCCESS;
 
 }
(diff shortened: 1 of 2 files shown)


dzenanz:
Please review @IbaiGenuaGamio.

Modules/Core/Transform/test/itkCenteredEuler3DTransformTest.cxx
```diff
@@ -335,6 +335,55 @@ int itkCenteredEuler3DTransformTest(int, char *[] )
...
+  testPoint[0] = rotationCenter[0];
+  testPoint[1] = rotationCenter[1];
+  testPoint[2] = rotationCenter[2];
+
+  itk::Point<double> outputTestPoint = centeredtransform->TransformPoint(testPoint);
+
+  double computeError = (outputTestPoint[0] - testPoint[0]) * (outputTestPoint[0] - testPoint[0]) +
+      (outputTestPoint[1] - testPoint[1]) * (outputTestPoint[1] - testPoint[1]) +
+      (outputTestPoint[2] - testPoint[2]) * (outputTestPoint[2] - testPoint[2]);
+  computeError = std::sqrt(computeError);
```
blowekamp: For future reference there is the [VectorDoubleRMSPredFormat](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/TestKernel/include/itkGTestPredicate.h#L58) to make these comparisons much easier.

dzenanz: This is essentially a copy-paste of @IbaiGenuaGamio's MWE case. I did not want to spend the time to understand the existing test, as reusing that was required to minimize code additions.

blowekamp:
+1 changes look good, and it's very nice a test was added.

dzenanz:
The test failure is unrelated and was not present on the repeated run. Merging.

PR #1061, opened 2019-07-03
Prefer enum class over 'c enumerations'

hjmjohnson:
Mathew Seng is working on a set of refactorings to move from "c enumerations" to "strongly typed enumerations".

#  PR Checklist

- [X] [Makes breaking changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes)
- [X] [Makes design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes)

See: http://www.stroustrup.com/C++11FAQ.html#enum

PR #1160, opened 2019-08-14
COMP: Improve default building of ITK with MSVC 1920-2929 version

hjmjohnson:
See https://docs.microsoft.com/en-us/cpp/build/reference/fs-force-synchronous-pdb-writes?view=vs-2019

and discussions:

https://discourse.itk.org/t/has-anyone-tried-to-build-visual-studio-c-2019/1561

## PR Checklist

### DIFF
--- a/CMake/ITKSetStandardCompilerFlags.cmake
+++ b/CMake/ITKSetStandardCompilerFlags.cmake
@@ -217,6 +217,12 @@ macro(check_compiler_platform_flags)
          if(MSVC_VERSION GREATER 1310)
            set(ITK_REQUIRED_CXX_FLAGS "${ITK_REQUIRED_CXX_FLAGS} /bigobj")
          endif()
+         # Force synchronous writes for building ITK on MSVC 1920-1929
+         # https://docs.microsoft.com/en-us/cpp/build/reference/fs-force-synchronous-pdb-writes?view=vs-201
+         if(MSVC_VERSION GREATER_EQUAL 1920)
+           set(ITK_REQUIRED_C_FLAGS "${ITK_REQUIRED_CXX_FLAGS} /FS")
+           set(ITK_REQUIRED_CXX_FLAGS "${ITK_REQUIRED_CXX_FLAGS} /FS")
+         endif()
        endif()
   endif()
 


dzenanz:
Add `itktestlib` to that list.

hjmjohnson:
@dzenanz I don't know how to do that.  Is this something that could be fixed in upstream VXL?  

I guess I don't fully understand the problem.  I thought this is a pervasive issue with MSVC 1920-1929 regarding parallel builds.  Any help would be greatly appreciated. I spend many hours this weekend trying to get ITK to build and searching for solutions in the wrong place until I came across the itk discourse that recommends adding /FS to the compiler flags.

hjmjohnson:
@dzenanz FYI:  I'm trying to track down this similar error: https://open.cdash.org/viewBuildError.php?buildid=6053536

dzenanz:
`ITKSmoothing` is a header-only module, so the `.lib` file does not exist. Try rebuilding the ***HeaderTest1**.

hjmjohnson:
> `ITKSmoothing` is a header-only module, so the `.lib` file does not exist. Try rebuilding the ***HeaderTest1**.

In the https://github.com/InsightSoftwareConsortium/ITK/pull/1061 PR, Smoothing is not header only any longer.  It now has a library associated with it for new source that needs to be compiled outside the templating.

blowekamp:
Since this is for the debugging symbols is it needed in release mode?

And do dependent project need it, or just internal when building ITK?

CMake/ITKSetStandardCompilerFlags.cmake
```diff
@@ -217,6 +217,12 @@ macro(check_compiler_platform_flags)
          if(MSVC_VERSION GREATER 1310)
            set(ITK_REQUIRED_CXX_FLAGS "${ITK_REQUIRED_CXX_FLAGS} /bigobj")
          endif()
+         # Force synchronous writes for building ITK on MSVC 1920-1929
+         # https://docs.microsoft.com/en-us/cpp/build/reference/fs-force-synchronous-pdb-writes?view=vs-201
+         if(MSVC_VERSION GREATER_EQUAL 1920)
+           set(ITK_REQUIRED_C_FLAGS "${ITK_REQUIRED_CXX_FLAGS} /FS")
```
dzenanz: This might be better added to VXL here:
https://github.com/vxl/vxl/blob/2e003cb38cd08444b7ac82088bc1ba57326d029f/CMakeLists.txt#L151-L152

dzenanz:
I haven't tried building in `Release` mode, only in `RelWithDebInfo` and `Debug` - both of which create debug information.

It is only needed for VXL, as far as I can tell. And since documentation says that `/FS` switch slows down compilation, I would rather not enable it for whole of ITK.

blowekamp:
> I haven't tried building in `Release` mode, only in `RelWithDebInfo` and `Debug` - both of which create debug information.
> 
> It is only needed for VXL, as far as I can tell. And since documentation says that `/FS` switch slows down compilation, I would rather not enable it for whole of ITK.

I have been building SimpleITK against ITKv4 with vs2019 on Azure Pipelines in Release mode without problems.

We likely should add an Azure vs2019 build for ITK. Would on the "batch" on merge script or for every PR?

hjmjohnson:
Superceeded by #1165

PR #1161, opened 2019-08-14
WIP: BUG: Remove Windows Python 2.7 hypot workaround

thewtex:
Applying #1067 file by file to determine the problem location...


hjmjohnson:
@thewtex Is this ready to come out of "WIP" status?

dzenanz:
Probably not. I think he wanted to have a multi-commit PR, to see which change breaks the compilation.

thewtex:
Yes, working on a search here to find the issue with #1067

PR #1168, opened 2019-08-16
DOC: AdaptiveHistogramEqualizationImageFilter default radius

thewtex:
Closes #727

PR #1170, opened 2019-08-16
ENH: Allow histogram matching with reference histogram

hjmjohnson:
Allow histogram matching from with either reference image or reference histogram.

## PR Checklist

-  :no_entry_sign: [Makes breaking changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes)
-  :no_entry_sign: Adds the License notice to new files.
-  :no_entry_sign: Adds Python wrapping.
-  :no_entry_sign: Adds tests and baseline comparison (quantitative).
-  :no_entry_sign: [Adds test data](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/UploadBinaryData.md).

### DIFF
--- a/Modules/Filtering/ImageIntensity/include/itkHistogramMatchingImageFilter.hxx
+++ b/Modules/Filtering/ImageIntensity/include/itkHistogramMatchingImageFilter.hxx
@@ -29,24 +29,18 @@ namespace itk
 template< typename TInputImage, typename TOutputImage, typename THistogramMeasurement >
 HistogramMatchingImageFilter< TInputImage, TOutputImage, THistogramMeasurement >
 ::HistogramMatchingImageFilter() :
-  m_SourceIntensityThreshold(NumericTraits<InputPixelType>::ZeroValue()),
-  m_ReferenceIntensityThreshold(NumericTraits<InputPixelType>::ZeroValue()),
-  m_OutputIntensityThreshold(NumericTraits<OutputPixelType>::ZeroValue()),
   m_SourceMinValue(NumericTraits<THistogramMeasurement>::ZeroValue()),
   m_SourceMaxValue(NumericTraits<THistogramMeasurement>::ZeroValue()),
-  m_SourceMeanValue(NumericTraits<THistogramMeasurement>::ZeroValue()),
   m_ReferenceMinValue(NumericTraits<THistogramMeasurement>::ZeroValue()),
   m_ReferenceMaxValue(NumericTraits<THistogramMeasurement>::ZeroValue()),
-  m_ReferenceMeanValue(NumericTraits<THistogramMeasurement>::ZeroValue()),
-  m_OutputMinValue(NumericTraits<THistogramMeasurement>::ZeroValue()),
-  m_OutputMaxValue(NumericTraits<THistogramMeasurement>::ZeroValue()),
-  m_OutputMeanValue(NumericTraits<THistogramMeasurement>::ZeroValue()),
   m_SourceHistogram(HistogramType::New()),
-  m_ReferenceHistogram(HistogramType::New()),
   m_OutputHistogram(HistogramType::New())
 
 {
-  this->SetNumberOfRequiredInputs(2);
+  this->SetNumberOfRequiredInputs(1);
+  Self::SetPrimaryInputName("SourceImage");
+  Self::AddOptionalInputName("ReferenceImage",1);
+  Self::AddOptionalInputName("ReferenceHistogram",2);
 
   m_QuantileTable.set_size(3, m_NumberOfMatchPoints + 2);
   m_QuantileTable.fill(0);
@@ -70,16 +64,10 @@ HistogramMatchingImageFilter< TInputImage, TOutputImage, THistogramMeasurement >
   os << indent << "ThresholdAtMeanIntensity: ";
   os << m_ThresholdAtMeanIntensity << std::endl;
 
-  os << indent << "SourceIntensityThreshold: ";
-  os << m_SourceIntensityThreshold << std::endl;
-  os << indent << "ReferenceIntensityThreshold: ";
-  os << m_ReferenceIntensityThreshold << std::endl;
-  os << indent << "OutputIntensityThreshold: ";
-  os << m_ReferenceIntensityThreshold << std::endl;
   os << indent << "Source histogram: ";
   os << m_SourceHistogram.GetPointer() << std::endl;
   os << indent << "Reference histogram: ";
-  os << m_ReferenceHistogram.GetPointer() << std::endl;
+  os << this->GetReferenceHistogram() << std::endl;
   os << indent << "Output histogram: ";
   os << m_OutputHistogram.GetPointer() << std::endl;
   os << indent << "QuantileTable: " << std::endl;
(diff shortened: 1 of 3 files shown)


blowekamp:
Looks like a nice refactoring!

A couple comments:
 - The class level documentation should be updated too
 - I do not like *hidden* IVAR like `m_GenerateReferenceHistogramFromImage` that are dependent on the order of setting variables. Please expose it directly the option."GenerateReferenceHistogramFromImage" option.
 - Please don't assign multiple input types to pipeline input 1. I'd configure the filter to use [named optional inputs](https://itk.org/Doxygen/html/classitk_1_1ProcessObject.html#a25e59a4b8b966bdb32f8359bc52f789c) for the "ReferenceImage" and the "ReferenceHistogram" I the constructor.
  - When the above is done you can use the [standard macros](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkMacro.h#L767) so the code and potential errors will be reduced.

Modules/Filtering/ImageIntensity/include/itkHistogramMatchingImageFilter.hxx
```diff
@@ -109,9 +130,17 @@ template< typename TInputImage, typename TOutputImage, typename THistogramMeasur
 const typename HistogramMatchingImageFilter< TInputImage, TOutputImage, THistogramMeasurement >
 ::InputImageType *
 HistogramMatchingImageFilter< TInputImage, TOutputImage, THistogramMeasurement >
-::GetReferenceImage()
+::GetReferenceImage() const
 {
-  return dynamic_cast< TInputImage * >( this->ProcessObject::GetInput(1) );
+  if (m_GenerateReferenceHistogramFromImage)
+    {
+    constexpr unsigned int reference_object_input_idx=1;
```
dzenanz: This variable is unnecessary. If you wanted to force 1 to be unsigned, you could have used `1u`, e.g. `GetInput(1u)`.

Modules/Filtering/ImageIntensity/include/itkHistogramMatchingImageFilter.hxx
```diff
@@ -277,8 +343,8 @@ void
 HistogramMatchingImageFilter< TInputImage, TOutputImage, THistogramMeasurement >
 ::DynamicThreadedGenerateData(const OutputImageRegionType & outputRegionForThread)
 {
-
-  InputImageConstPointer input  = this->GetInput();
+  constexpr unsigned int source_image_idx=0;
+  InputImageConstPointer input  = this->GetInput(source_image_idx);
```
dzenanz: `GetInput(0)` would have worked too, and had one less entity (`source_image_idx`) to contend with.

Modules/Filtering/ImageIntensity/test/itkHistogramMatchingImageFilterTest.cxx
```diff
@@ -57,17 +57,70 @@ class ShowProgressObject
...
+      ++outIter;
+      ++refIter;
+    }
+  return passed;
 }
 
+template <typename HistogramType>
+void
+PrintHistogramInfo( const HistogramType & refHistogram )
+{ // Write one histogram to disk
```
dzenanz: Write to console?

Modules/Filtering/ImageIntensity/test/itkHistogramMatchingImageFilterTest.cxx
```diff
@@ -102,74 +155,161 @@ int itkHistogramMatchingImageFilterTest()
...
+  filterWithReferenceImage->SetNumberOfHistogramLevels( 50 );
+  filterWithReferenceImage->SetNumberOfMatchPoints( 8 );
 
-  ShowProgressObject progressWatch(filter);
-  itk::SimpleMemberCommand<ShowProgressObject>::Pointer command;
-  command = itk::SimpleMemberCommand<ShowProgressObject>::New();
-  command->SetCallbackFunction(&progressWatch,
-                               &ShowProgressObject::ShowProgress);
-  filter->AddObserver(itk::ProgressEvent(), command);
+  ShowProgressObject                                      progressWatch( filterWithReferenceImage );
```
dzenanz: This auto-formatting is better undone.

Modules/Filtering/ImageIntensity/include/itkHistogramMatchingImageFilter.hxx
hjmjohnson: @dzenanz It used to be GetInput(0), but I was trying to remove "magic numbers".  The constexpr could also have been `#define soruce_image_idx=0`.  I was trying to increase readability, and provide a token to search for in the code so that I could isolate all the places where source_image_idx was used.  Searching for `0` did not provide a useful query :).

After addressing comments by Brad, I'm only using named inputs, so this is now not relevant.

Modules/Filtering/ImageIntensity/include/itkHistogramMatchingImageFilter.hxx
hjmjohnson: Removed.

hjmjohnson:
> * The class level documentation should be updated too

@blowekamp Done.

hjmjohnson:
> * I do not like _hidden_ IVAR like `m_GenerateReferenceHistogramFromImage` that are dependent on the order of setting variables. Please expose it directly the option."GenerateReferenceHistogramFromImage" option.

@blowekamp Done.

hjmjohnson:
> * Please don't assign multiple input types to pipeline input 1. I'd configure the filter to use [named optional inputs](https://itk.org/Doxygen/html/classitk_1_1ProcessObject.html#a25e59a4b8b966bdb32f8359bc52f789c) for the "ReferenceImage" and the "ReferenceHistogram" I the constructor.
> * When the above is done you can use the [standard macros](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkMacro.h#L767) so the code and potential errors will be reduced.

Done and Done.

Modules/Filtering/ImageIntensity/test/itkHistogramMatchingImageFilterTest.cxx
hjmjohnson: Fixed.

Modules/Filtering/ImageIntensity/include/itkHistogramMatchingImageFilter.hxx
```diff
@@ -142,58 +116,108 @@ void
...
 ::BeforeThreadedGenerateData()
 {
-  unsigned int j;
+  THistogramMeasurement sourceMeanValue;
+  THistogramMeasurement referenceMeanValue;
+
+  InputPixelType  sourceIntensityThreshold;
+  InputPixelType  referenceIntensityThreshold;
+
+  if ( m_GenerateReferenceHistogramFromImage )
```
blowekamp: The best place to check if all the inputs are set at require is by overloading the [VerifyPreconditions ](https://itk.org/Doxygen/html/classitk_1_1ProcessObject.html#a73000f23064b7d2898fa21135080b4cf)method. This method gets called before any pipeline execution occurs even, `GenerateOutputInformation` and follows the fail quickly principle.

blowekamp:
One tangental thought, I'm unaware of a [histogram equalization](https://en.wikipedia.org/wiki/Histogram_equalization) filter in ITK. This filter could be used that way thanks to this improvement with this addition and the proper configuration with a flat histogram. It might be nice to have a separate ( derived or sibling ) class which just does HE, or perhaps an additional method to explicit setup HE, or perhaps is just up user. Maybe at the very least a mention in the documentation that this filter can do "Histogram Equalization" when given a flat histogram.

gdevenyi:
@blowekamp 	what about https://github.com/InsightSoftwareConsortium/ITK/pull/1168

blowekamp:
> @blowekamp what about #1168

That AdaptiveHE works on local neighborhoods, while that classic HE make the histogram of the entire image flat. There was a user who confused the two a short while ago. Running global  HE before AI may be something people want to easily do.

PR #1165, opened 2019-08-15
COMP: Update vxl to support msvc 2019 v142

hjmjohnson:
Address issues related to #1160 targeted to vxl codebase

@dzenanz suggested vxl only changes.

-  :no_entry_sign: [Makes breaking changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes)
-  :no_entry_sign: [Makes design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes)
-  :no_entry_sign: Adds the License notice to new files.
-  :no_entry_sign: Adds Python wrapping.
- :no_entry_sign: Adds tests and baseline comparison (quantitative).
-  :no_entry_sign: [Adds test data](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/UploadBinaryData.md).
- :no_entry_sign: Adds Examples to [ITKExamples](https://github.com/InsightSoftwareConsortium/ITKExamples)
- :no_entry_sign: Adds Documentation.

PR #1173, opened 2019-08-16
ENH: back-porting #1165 to support Visual Studio 2019

dzenanz:
On my computer, this patch is not required to build with VS2019. Should we still consider merging it?

### DIFF
--- a/Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmIPPSorter.cxx
+++ b/Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmIPPSorter.cxx
@@ -48,7 +48,7 @@ struct dircos_key {
 };
 
 struct dircos_comp {
-  bool operator()( dircos_key const & lhs, dircos_key const & rhs ) {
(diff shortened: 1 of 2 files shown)


dzenanz:
I still get `error C3848:  expression having type 'const gdcm::dircos_comp' would lose some const-volatile qualifiers in order to call 'bool gdcm::dircos_comp::operator ()(const gdcm::dircos_key &,const gdcm::dircos_key &)'`.

thewtex:
Seems to be other issues -- ITK 5 can be required for VS 2019.

blowekamp:
I plan on looking into this GDCM issue this week as I don't think it's to far from working.

blowekamp:
@dzenanz I update you PR with the little fix needed from GDCM.

dzenanz:
It could have been an independent PR, and I could have rebased this on it. This if fine too.

dzenanz:
Azure pipelines are not expected to succeed for 4.x. CircleCI returned 255 despite all the tests succeeding. Good enough?

blowekamp:
Merging is fine with me.

dzenanz:
I swapped the order of commit, so the necessary commit comes before the maybe-needed commit.

Issue #1136, opened 2019-08-04
Consume and produce Numpy arrays


mrocklin:
I would find it convenient if `itk` functions optionally consumed and produced numpy arrays.  Here is a motivating example function that I'm working with:

### Current code

```python
import itk

image = np.array(...)  # 3d array
kernel = np.array(...)  # 3d array    

image = itk.image_view_from_array(image)   # Convert to ITK object
kernel = itk.image_view_from_array(kernel)  # Convert to ITK object
    
deconvolved = itk.richardson_lucy_deconvolution_image_filter(image,
                                                             kernel_image=kernel,
                                                             number_of_iterations=iterations)
    
result = itk.array_from_image(deconvolved)  # Convert back to Numpy array
```

### Desired code

```python
import itk

image = np.array(...)  # 3d array
kernel = np.array(...)  # 3d array    
    
deconvolved = itk.richardson_lucy_deconvolution_image_filter(image,
                                                             kernel_image=kernel,
                                                             number_of_iterations=iterations)
```

### Discussion

It would be nice for `itk` functions to consume and produce Numpy arrays.  This would help them compose with the rest of the SciPy ecosystem.  However I suspect that there is a good reason why this isn't done today. For example maybe they have important metadata.  However, if a user presents with a Numpy array then maybe there are some things that we can do?  For example maybe the `itk.image_view_from_array` function could be called automatically on all inputs that are expected to be ITK Image objects but are actually NumPy arrays.  Additionally, perhaps if the user presents with a Numpy array then we might automatically call  `itk.array_from_image` on our result?

mrocklin:
cc @jakirkham

jakirkham:
Probably as a first pass it would be useful to have a function (maybe a decorator?) that takes an existing ITK function and does the following.

1. Converts the input array into an ITK image
2. Calls the give function with that image
3. Computes the result as a NumPy array, to return

This would at least make it easier to wrap ITK function to match this idiom with hopefully minimal work.

thewtex:
@mrocklin brilliant idea! We will definitely do this!

> For example maybe they have important metadata. 

Yes. Spatial metadata is critical in practice. There is more discussion on this metadata and NumPy-array-like-interface's [in this Discourse thread](https://discourse.itk.org/t/images-in-physical-space-in-python/2124/15).

> itk.image_view_from_array function could be called automatically on all inputs that are expected to be ITK Image objects but are actually NumPy arrays.

Yes!

[...]

PR #1176, opened 2019-08-19
ENH: Support NumPy array-like inputs to image filters

thewtex:
When a NumPy array-like is passed into Python snake_case interface to
filters, the array-like arguments or standard image keyword arguments
are converted into itk.Image views, and passed to the filter. All image
outputs are converted to NumPy ndarray's if array's were used as
inputs.

Addresses #1136

### DIFF
--- a/Wrapping/Generators/Python/itkHelpers.py
+++ b/Wrapping/Generators/Python/itkHelpers.py
@@ -17,8 +17,60 @@
 #==========================================================================*/
 
 import re
+import functools
 
 def camel_to_snake_case(name):
     snake = re.sub('(.)([A-Z][a-z]+)', r'\1_\2', name)
     snake = re.sub('([a-z0-9])([A-Z])', r'\1_\2', snake)
     return snake.replace('__', '_').lower()
+
+def is_arraylike(arr):
+    return hasattr(arr, 'shape') and \
+    hasattr(arr, 'dtype') and \
+    hasattr(arr, '__array__') and \
+    hasattr(arr, 'ndim')
+
+def accept_numpy_array_like(image_filter):
+    """Decorator that allows itk.ProcessObject snake_case functions to accept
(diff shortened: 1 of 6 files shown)


mrocklin:
Well this is exciting!  I'm glad to see that this is such a small code change for you all.

Wrapping/Generators/Python/itkHelpers.py
```diff
@@ -17,8 +17,60 @@
...
+            if key.lower() in potential_image_input_kwargs and is_arraylike(value):
+                have_array_like_input = True
+                array = np.asarray(value)
+                image = itk.image_view_from_array(array)
+                kwargs[key] = image
+
+        if have_array_like_input:
+            # Convert output itk.Image's to numpy.ndarray's
+            output = image_filter(*tuple(args_list), **kwargs)
+            if output is tuple:
```
mrocklin:
```suggestion
            if isinstance(output, tuple):
```

jakirkham:
Would it be possible also to handle anything that supports the buffer protocol?

jakirkham:
Otherwise this looks great! Thanks for putting it together Matt 🙂

thewtex:
Thanks for the reviews!

Addressed the issue identified by @mrocklin

thewtex:
> Would it be possible also to handle anything that supports the buffer protocol?

The intention is to support anything NumPy array-like, e.g. `dask.array`, with `np.asarray`:

https://github.com/InsightSoftwareConsortium/ITK/pull/1176/files#diff-7404aefb5771301dd2c2edfe74f17008R48

but, it sounds like this could / should change to `np.coerce` in the future following [the discussion of NEP 30](https://mail.python.org/pipermail/numpy-discussion/2019-August/079897.html).

mrocklin:
Woo!

PR #1178, opened 2019-08-19
BUG: Do not run itkPyBufferMemoryLeakTest on Windows

thewtex:
The required Python `resource` module is Unix-only.


thewtex:
The tests fails because the `resource` module is not available in Windows:

```
  File "D:/a/1/s/Modules/Bridge/NumPy/wrapping/test/itkPyBufferMemoryLeakTest.py", line 25, in <module>
    import resource
ModuleNotFoundError: No module named 'resource'
itkTestDriver: Process exited with return value:
```
https://open.cdash.org/testDetails.php?test=795040246&build=6065508

https://docs.python.org/3.7/library/resource.html

PR #315, opened 2018-12-16
ENH: Add regression test for reading legacy multi-frame DICOM

thewtex:
Ensure that origin and spacing and identified correctly (the image
comparison checks this metadata).

Test image from David Clunie.

See also the discussion:

  https://discourse.slicer.org/t/dicom-multiframe-support/4806/9

PR #243, opened 2018-11-29
BUG: The Z spacing was set using thickness

dzenanz:
This addresses https://issues.itk.org/jira/browse/ITK-3527

According Dicom standard (DICOM PS3.6 2016b - Data Dictionary)
(0028, 0030) indicates physical X,Y spacing inside a slice;
(0018, 0088) indicates physical Z spacing between slices;
which above are also consistent with Dcom2iix software.

In current code, the zSpace is get from (0x0018,0x0050) which
actually is the thickness from which the signal is created,
not the separation of signal samplings.

This patch set resolves the problem by updating the Z get
spacing method by using (0018, 0088) when it is available.

Supersedes #112.

PR #112, opened 2018-11-05
WIP: BUG: The Z spacing was set using thickness

hjmjohnson:
This addresses https://issues.itk.org/jira/browse/ITK-3527

According Dicom standard (DICOM PS3.6 2016b - Data Dictionary)
(0028, 0030) indicates physical X,Y spacing inside a slice;
(0018, 0088) indicates physical Z spacing between slices;
which above are also consistent with Dcom2iix software.

In current code, the zSpace is get from (0x0018,0x0050) which
actually is the thickness from which the signal is created,
not the separation of signal samplings.

This patch set resolves the problem by updating the Z get
spacing method by using (0018, 0088) when it is available.

If (0018, 0088) is not available, then the code reverts to
the previous behavior of using (0x0018,0x0050).

PR #1179, opened 2019-08-20
Bugfixes in DCMTK

### LINKED ISSUE #1125: itkDCMTKImageIO.h RGB Converter
### Description
In Line 320 ~ 337 in itkDCMTKImageIO.cxx the buffer of the DICOM Image is copied to the itkImage buffer. For the RGB(A) Pixels a reorder of the RGB(A) Pixel (itkDCMTKImageIO.h Line 123 ~134) is made, which is useless, because in Line 337 (itkDCMTKImageIO.cxx) the old buffer is again copied over the new sorted buffer.

### Steps to Reproduce
See above

### Versions

ITK 5.0.0

### Environment

OS: Windows
Platform Toolset: MSVS 2017 Visual Studio 2017 (v141)
Windows SDK Version: 10.0.15063.0

maekclena:
The first commit fixes #1125 and removes an outdated comment.
The second commit adds a metadata read for spacing (according to already existing comments in the code) to fix failing test `itkDCMTKImageIOMultiFrameImageTest`.

## PR Checklist

### DIFF
--- a/Modules/IO/DCMTK/include/itkDCMTKImageIO.h
+++ b/Modules/IO/DCMTK/include/itkDCMTKImageIO.h
@@ -112,7 +112,7 @@ private:
   void OpenDicomImage();
 
   /** Finds the correct type to call the templated function ReorderRGBValues(...) */
-  void ReorderRGBValues(void *buffer, const void* data, unsigned long count, unsigned int voxel_size);
+  void ReorderRGBValues(void *buffer, const void* data, size_t count, unsigned int voxel_size);
   /** Reorders RGB values in an image from color-by-plane (R1R2R3...G1G2G3...B1B2B3...)
    * to color-by-pixel (R1G1B1...R2G2B2...R3G3B3...). `voxel_size` specifies the pixel size: It should
    * be 3 for RGB images, and 4 for RGBA images.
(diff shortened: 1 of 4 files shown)


dzenanz:
I remember some discussions about this, e.g. #112 and #315. Were you aware of it @maekclena?

maekclena:
No, I did not know. It seems like the fallback behavior was mistakenly removed in #243.

hjmjohnson:
> No, I did not know. It seems like the fallback behavior was mistakenly removed in #243.

I don't think it was mistakenly removed.  My recollection is that it was purposefully removed because using SliceThinkness was determined to be the wrong approach.  If 0018, 0088 is not present, then the correct behavior is to explicitly compute the distance between slices as described in 

https://github.com/InsightSoftwareConsortium/ITK/pull/112#pullrequestreview-179175160
and
https://github.com/InsightSoftwareConsortium/ITK/pull/112#issuecomment-442451420

@malaterre is an expert in DICOM conformance.

dzenanz:
Please split this into two PRs. First commit is non-contentious, and can be merged immediately. That will allow reviews for the second commit independently.

malaterre:
> > No, I did not know. It seems like the fallback behavior was mistakenly removed in #243.
> 
> I don't think it was mistakenly removed. My recollection is that it was purposefully removed because using SliceThinkness was determined to be the wrong approach. If 0018, 0088 is not present, then the correct behavior is to explicitly compute the distance between slices as described in
> 
> [#112 (review)](https://github.com/InsightSoftwareConsortium/ITK/pull/112#pullrequestreview-179175160)
> and
> [#112 (comment)](https://github.com/InsightSoftwareConsortium/ITK/pull/112#issuecomment-442451420)
> 
> @malaterre is an expert in DICOM conformance.

1. 0018,0088 is Spacing Between Slices and should represent the actual Z-spacing when present,
2. Never (ever!) confuse Spacing Between Slices and Slice Thickness those are totally independent attributes
3. During the course of ITK dev, it was found that Spacing Between Slices may be missing (eg. CT Image Storage), or incorrect. Computing the actual Z-spacing from the IOP & IPP make sure to report any missing slices or inconsistency. Which is the reason why it is acceptable to loop over all frames to (re)compute the Slice Interval.

Ref:
* https://stackoverflow.com/a/15116116/136285

PR #1184, opened 2019-08-21
COMP: Address internal compiler error on MSVC 19.0.24234.1 (v140)

blowekamp:
Work around internal compiler error with the MSVS v140 compiler
apparently related to default copy constructor used in the functor
generator filter.

## PR Checklist


blowekamp:
> When does this trigger the ICE? Does it happen with VS2015 Update3, or an earlier version?

It is happening on the newly add AZP continuous batch build with 'vs2017-win2016' running VS2017 and v140 toolset with cl version 19.0.24234.1:
https://open.cdash.org/viewBuildError.php?buildid=6071335

I also reproduced in with VS2019 and the v140 toolset where the cl.exe version is reporting as 19.00.24245.

I don't know what version of cl.exe comes with VS2015 update 3.

I was also able to reproduce the ICE on the release branch so it needs to be ported there too.

This is a very important compiler as Python 3.5+ is using this toolset. It may be better to have the v140 toolset on the CI, than the current v141.

PR #1188, opened 2019-08-22
BUG: Address bug with small size in output of SliceImageFilter

### LINKED ISSUE #1177: Image slicing
### Description

I am trying to slice a 3D image using indexing but I get one slice short of what one would expect. After looking into it, it seems the last slice is always missing. 

### Steps to Reproduce

Here’s a simple example using a simple array:
```{python}
import numpy as np
import SimpleITK as itk
 
test_array = np.array([0,1,2,3,4] * 6).reshape(3,2,5)
test_array = test_array.swapaxes(2,0)
test_array
```
```{python}
Out[2]:
array([[[0, 0, 0],
        [0, 0, 0]],
       [[1, 1, 1],
        [1, 1, 1]],
       [[2, 2, 2],
        [2, 2, 2]],
       [[3, 3, 3],
        [3, 3, 3]],
       [[4, 4, 4],
        [4, 4, 4]]])
```
If we convert this array to an Image and take every second slice in the third dimension and convert it back to a numpy array (for easier visualization), we expect to get an array of size (3,2,3), the same as one would by using the same operation on the array directly. Instead we get this

```{python}
test_img = itk.GetImageFromArray(test_array)
sliced = test_img[:,:,::2]
sliced_array = itk.GetArrayFromImage(sliced)
sliced_array
```
```{python}
Out[3]:
array([[[0, 0, 0],
        [0, 0, 0]],
       [[2, 2, 2],
        [2, 2, 2]]])
```
So the last slice 
```{python}
[[4,4,4],
[4,4,4]]
```
is missing. The same issue occurs when slicing with SliceImageFilter
```{python}
slice = itk.SliceImageFilter()
slice.SetStep((1,1,2))
sliced_filter = slice.Execute(test_img)
sliced_filter.GetSize()
```
```{python}
Out[14]: (3, 2, 2)
```
```{python}
sliced_filter = itk.GetArrayFromImage(sliced_filter)
sliced_filter
```
```{python}
Out[16]: 
array([[[0, 0, 0],
        [0, 0, 0]],
       [[2, 2, 2],
        [2, 2, 2]]])
```

### Environment
I have run the above experiment on Ubuntu 18.04 using both Python 3.6.8 and Python 3.5.2 and SimpleITK version 1.2.0.

blowekamp:
The size of the output with a non-one step was being floored, instead
include an extra index if available. For example for a size of 5 and a
step of two the filter will now produce a size of 3.

Closes #1177.

## PR Checklist

### DIFF
--- a/Modules/Filtering/ImageGrid/include/itkSliceImageFilter.hxx
+++ b/Modules/Filtering/ImageGrid/include/itkSliceImageFilter.hxx
@@ -284,7 +284,8 @@ SliceImageFilter< TInputImage, TOutputImage >
     if ( (m_Step[i] > 0 && stop > start) ||
          ( m_Step[i] < 0 && stop < start ) )
       {
-      outputSize[i] = (stop-start)/m_Step[i];
+      outputSize[i] = (stop-start-Math::sgn(m_Step[i]))/m_Step[i];
+      outputSize[i] += 1u;
       }
     else
       {


blowekamp:
I got them working again a while back, but then additional changes were made to the dashboard branch so they broke again. There are positive and negative with keeping the CI configuration at part of the code as opposed to separate.

blowekamp:
The AZP ITK.macOS has one failing test "itkCastImageFilterTest", and it's not surprising there are a bunch of warning on the platforms. I think this is good enough for a merge?

PR #1094, opened 2019-07-15
STYLE: Prepare for clang format style enforcement

hjmjohnson:
Provide initial code changes to facilitate enforcement of an ITK style that can be uniformly applied in an automated way.

These problems appeared when enforcing a consistent style with the clang-formatting tool.  It is a first step towards providing an easier code formatting style enforcement throughout ITK.

PR #1191, opened 2019-08-23
STYLE: Enforce ITK style defined by .clang-format

hjmjohnson:
ITK/Utilities/Maintenance/clang-format.bash \
    --clang-format ~/local/llvm/clang+llvm-8.0.0-x86_64-linux-gnu-ubuntu-18.04/bin/clang-format \
    --tracked

This patch updates the ITK preferred style to one that can be
automatically enforced, integrated with many IDE's, and based
on externally defined documentation and examples.

See: https://discourse.itk.org/t/update-coding-style-for-itk/2055

Most of the changes make inconsistent coding practices consistent
throughout the toolkit.

One large change is changing the {} indentation style from
`Whitesmiths` style https://en.wikipedia.org/wiki/Indentation_style#Whitesmiths_style
to an indentation style consistent with many of the other common
indentation styles.  The primary reason for this change is to
overcome the common deficiency of IDE's and formatters to support
this uncommon style choice.
Completes the work started in #1094 

## PR Checklist


gdevenyi:
Is it possible to add a test to new PRs that check for style?

hjmjohnson:
> Is it possible to add a test to new PRs that check for style?

That should be possible, and probably not too hard for someone who better understands how to make those checks.

bradking:
@hjmjohnson for the commit that actually applies the mass style change I suggest commiting with `--author=...` to set the author to a robot user (e.g. kwrobot).  That way you won't be `git blame`d for all the code.  For example, see [CMake MR 2123](https://gitlab.kitware.com/cmake/cmake/merge_requests/2123/commits) where we updated style there.  We also added an empty commit just before the transition and instructions in the commit messages for people to rebase their topics past the formatting changes.

bradking:
> Is it possible to add a test to new PRs that check for style?

Yes, the `ghostflow-check-master` check can be updated to also verify style with `clang-format`.  However, `clang-format`'s format is not quite stable across versions.  Therefore we must pick a single major version of `clang-format` for this purpose.  CMake currently uses `clang-format-6.0`.

bradking:
For reference, if we choose a `clang-format` version other than 6.0 it will take some time to make `ghostflow-check-master` check it because some infrastructure needs to be upgraded.

hjmjohnson:
> @hjmjohnson for the commit that actually applies the mass style change I suggest commiting with `--author=...` to set the author to a robot user (e.g. kwrobot). That way you won't be `git blame`d for all the code. For example, see [CMake MR 2123](https://gitlab.kitware.com/cmake/cmake/merge_requests/2123/commits) where we updated style there. We also added an empty commit just before the transition and instructions in the commit messages for people to rebase their topics past the formatting changes.

@bradking EXCELLENT IDEAS!!! Thank you.  I'll do that.

hjmjohnson:
> For reference, if we choose a `clang-format` version other than 6.0 it will take some time to make `ghostflow-check-master` check it because some infrastructure needs to be upgraded.

I'm using a modified version of the cmake mechanism with standardization on clang 8.0 (for a new refined C++11 feature support, and because that is the clang-format version that is standard with one of the Visual studio plugins, and CLion plugins.

hjmjohnson:
> VS and Clion come with an embedded version for clang-format,

I choose to standardize on clang 8 to be most aligned with those two IDE's :).

Examples/DataRepresentation/Containers/TreeContainer.cxx
```diff
@@ -52,7 +52,8 @@
 #include "itkTreeIteratorClone.h"
 // Software Guide : EndCodeSnippet
 
-int main(int, char* [])
+int
+main( int, char *[] )
```
N-Dekker: Do you really want to enforce having the return type of a function on a separate line? It's OK by me, I just have to get used to it. But if you do, you might as well consider using C++11 trailing return types everywhere:

    auto main( int, char *[] ) ->
    int

Examples/DataRepresentation/Containers/TreeContainer.cxx
```diff
@@ -64,7 +65,7 @@ int main(int, char* [])
 
   // Software Guide : BeginCodeSnippet
   using NodeType = int;
-  using TreeType = itk::TreeContainer<NodeType>;
+  using TreeType = itk::TreeContainer< NodeType >;
```
N-Dekker: For me personally these spaces around an argument are not really necessary. (But I can get used to it.)

Examples/DataRepresentation/Containers/TreeContainer.cxx
```diff
@@ -77,7 +78,7 @@ int main(int, char* [])
 
 
   // Software Guide : BeginCodeSnippet
-  tree->SetRoot(0);
+  tree->SetRoot( 0 );
```
N-Dekker: Again, for me personally these spaces around an argument are not really necessary. (But I can get used to it.)

Examples/DataRepresentation/Image/Image4.cxx
```diff
@@ -347,8 +346,8 @@ int main(int, char *[])
 
   // Software Guide : BeginCodeSnippet
   const ImageType::IndexType LeftEyeIndex = GetIndexFromMouseClick();
-  ImageType::PointType LeftEyePoint;
-  image->TransformIndexToPhysicalPoint(LeftEyeIndex,LeftEyePoint);
+  ImageType::PointType       LeftEyePoint;
```
N-Dekker: I very much dislike alignment somewhere in the middle of a line of code. It would mean that whenever line 348 is modified, line 349 would also be modified to keep the alignment in sync. Please no!

Examples/DataRepresentation/Mesh/AutomaticMesh.cxx
```diff
@@ -100,12 +100,10 @@ main(int, char *[])
...
-  meshSource->AddTetrahedron(
-    meshSource->AddPoint( -1, -1, -1 ),
-    meshSource->AddPoint(  1,  1, -1 ),
-    meshSource->AddPoint(  1, -1,  1 ),
-    meshSource->AddPoint( -1,  1,  1 )
-  );
+  meshSource->AddTetrahedron( meshSource->AddPoint( -1, -1, -1 ),
+                              meshSource->AddPoint( 1, 1, -1 ),
+                              meshSource->AddPoint( 1, -1, 1 ),
+                              meshSource->AddPoint( -1, 1, 1 ) );
```
N-Dekker: I like the previous much better. A change of the name of the variable `meshSource` or the function `AddTetrahedron` should only affect line 103, not the following lines.

Examples/DataRepresentation/Containers/TreeContainer.cxx
hjmjohnson: I have mixed feelings on this.  1. condensed is nice for more content on one line, but 2. My eyesight is really bad, and the spaces do facilitate the mechanics of reading code with spaces.

In reviewing the historical code, the use/disuse of spaces in cases like this was mixed without a definitive common practice.  Either choice was going to make many code changes.

Again the goal is to provide a more easily enforced common standard.

Examples/DataRepresentation/Image/Image4.cxx
hjmjohnson: @N-Dekker, Personally I agree, but this has been discussed several times before, and the historical community agreement is that the benefits in readability outweigh the pain.

Examples/DataRepresentation/Mesh/AutomaticMesh.cxx
hjmjohnson: This is an example of a `con` that is outweighed by the `pro`s of having a uniform, easily automated fixed format style.

Examples/DataRepresentation/Image/Image4.cxx
seanm: FWIW, I agree with @N-Dekker. It will mess up blame & history to re-indent like that.

Examples/DataRepresentation/Containers/TreeContainer.cxx
lassoan: Space is _required_ by some compilers when there are multiple embedded templates. For example: ` itk::TreeContainer<std::vector<NodeType>>` throws error on some compilers. Always adding the space (even if it's not required by the compiler) makes things a bit more consistent.

Examples/DataRepresentation/Image/Image4.cxx
lassoan: I agree, it's very hard to maintain this alignment. Even if this was discussed before, since formatting will change now anyway, it would be a good time to re-check what most people prefer.

There are many similar questions where people can have different preferences, so maybe all the controversial decisions could be put up on a public vote. We could give different weight to votes depending on involvement in ITK project (e.g., a core ITK developer could have 10x more vote than an ITK user).

blowekamp:
I am in favor of changes to style which reduce the style noise in our history. Looking through history to resolve issues and find the cause of problems is time consuming. Find the real algorithm changes could be made easier if there was less style noise.

Examples/DataRepresentation/Mesh/AutomaticMesh.cxx
lassoan: It seems that you can configure `AlignAfterOpenBracket: DontAlign` and then fixed `ContinuationIndentWidth` is used for alignment instead.

Examples/DataRepresentation/Containers/TreeContainer.cxx
dzenanz: This is mainly done to accommodate templated return types, such as `itk::SmartPointer<itk::Image<unsigned short int,3>>`. Having `int` and `void` on a line all by itself is quite annoying, but a price we will have to pay for consistency. I guess we will get used to it over time.

Examples/DataRepresentation/Containers/TreeContainer.cxx
dzenanz: Same, I got used to it after a while.

Examples/DataRepresentation/Containers/TreeContainer.cxx
dzenanz: Same, I got used to it after a while.

Examples/DataRepresentation/Image/Image4.cxx
dzenanz: Yes, this is annoying inside functions/methods. Unfortunately, clang-format does not distinguish that from declarations inside classes.

@hjmjohnson Maybe we could completely abolish alignment of declarations because of this?

seanm:
Perhaps a mega commit reformatting the whole codebase should come later? Perhaps we should start by enforcing the new style only on new commits? This would give a chance to tweak the style based on real-world usage.

Examples/DataRepresentation/Containers/TreeContainer.cxx
seanm: @lassoan those spaces were only required pre-C++11, no?

Examples/DataRepresentation/Containers/TreeContainer.cxx
lassoan: It should be possible to manage long return types by tuning PenaltyReturnTypeOnItsOwnLine.

dzenanz:
@seanm This is the mega-commit. The style was last discussed in #1094.

Examples/DataRepresentation/Containers/TreeContainer.cxx
lassoan: I just remember that I had to add those spaces for some compilers. Maybe they are no longer needed.

dzenanz:
I second Andras' suggestion to have a vote about contentions/annoying style, and in case of ties go with the old style.

Examples/DataRepresentation/Containers/TreeContainer.cxx
dzenanz: With C++11 (which we require since 5.0) space between closing template angle brackets is no longer required, so `itk::TreeContainer<std::vector<NodeType>>` is fine.

I prefer no spaces, but I have gotten used to spaces since using ITK extensively.

bradking:
I see some posts above concerned about `git blame` stopping on the style change.  This is inevitable, but we can help users that reach it.  In CMake, our latest style change commit is `d7204e649ed4ebb19bb341b4e49eb51514364922`, and its commit message helps readers move past it during `git blame`:

```
Revise C++ coding style using clang-format-6.0

Run the `clang-format.bash` script to update all our C and C++ code to a
new style defined by `.clang-format`.  Use `clang-format` version 6.0.

* If you reached this commit for a line in `git blame`, re-run the blame
  operation starting at the parent of this commit to see older history
  for the content.

* See the parent commit for instructions to rebase a change across this
  style transition commit.
```

Typically if I see a blame that reaches that commit, I then run `git blame d7204e649~1 -- ...` to look at content before the style change.

hjmjohnson:
> Also related, is it possible to add this as an (optional) precommit hook for ITK developers?
> 
> https://gist.github.com/alexeagle/c8ed91b14a407342d9a8e112b5ac7dab

I am very supportive of this pre-commit, but I want to separate that effort from this one.  See #1192

hjmjohnson:
> Hopefully a pre-commit hook could be overridden? It would be a shame to discourage small first time patches just because someone doesn't have a specific version of clang-format on their computer.

I think any pre-commit hook would be disabled by default.

N-Dekker:
@dzenanz wrote:
> @seanm This is the mega-commit. The style was last discussed in #1094.

Indeed, a MEGA-commit! https://github.com/InsightSoftwareConsortium/ITK/pull/1191/commits/99e75a8a4ab970365c96ecc148b5a23305a03578 says:

> The diff you're trying to view is too large. We only load the first 3000 changed files. 

😮

Examples/DataRepresentation/Image/Image4.cxx
hjmjohnson: @dzenanz @lassoan @N-Dekker @thewtex I want to be VERY CLEAR:  I am doing my best to provide better automation of style, and not dictate the style.  I spent a lot of time finding the set of options that minimizes changes to the current style while providing AUTOMATION.

I'm willing to entertain other global style changes, but to the extent possible I would rather not let the scope of this PR require that every style be re-considered.

Please vote now, and recruit others to support/reject this coding change.

NOTE:  Making this change will also require changing KWSTYLE rules, as the current suggestions will cause the KWSTYLE checks to fail.

--- NOW my personal opinion:  Let's change KWSTYLE and remove it's enforcement, and change .clang-format `AlignConsecutiveDeclarations: true` to `AlignConsecutiveDeclarations: false`

Examples/DataRepresentation/Mesh/AutomaticMesh.cxx
hjmjohnson: @lassoan Thank you, I'll try that.

hjmjohnson:
> Perhaps a mega commit reformatting the whole codebase should come later? Perhaps we should start by enforcing the new style only on new commits? This would give a chance to tweak the style based on real-world usage.

I am opposed to a graduate change.  In the long run a gradual change to the code base will be more disruptive than a reset.

hjmjohnson:
> I second Andras' suggestion to have a vote about contentions/annoying style, and in case of ties go with the old style.

@dzenanz Do you have a suggestion about how to start/coordiante this voting process?

Examples/DataRepresentation/Image/Image4.cxx
blowekamp: Thank you for taking the time for this, I look forward to not thinking about style anymore. 👔 🕴 

Are we going to need KWStyle any more?

My issue with looking through history, is not a single change like this one ( seems to be a very good plan for it ). It is more the frequent white space changes cause by this case and the one below. Thank you for being open to a couple be for the big commit.

hjmjohnson:
> > I second Andras' suggestion to have a vote about contentions/annoying style, and in case of ties go with the old style.
> 
> @dzenanz Do you have a suggestion about how to start/coordiante this voting process?

The style must be enforceable by a configuration in : http://releases.llvm.org/8.0.0/tools/clang/docs/ClangFormatStyleOptions.html

samuelgerber:
I am partial to the K&R indentation style, particularly the Java variant (https://en.wikipedia.org/wiki/Indentation_style#Variant:_Java) with mandatory braces.
I believe it makes code more compact and easy to read.

Examples/DataRepresentation/Image/Image4.cxx
```diff
@@ -136,7 +138,7 @@ int main(int, char *[])
   //  Software Guide : EndLatex
 
   // Software Guide : BeginCodeSnippet
-  const ImageType::SpacingType& sp = image->GetSpacing();
+  const ImageType::SpacingType & sp = image->GetSpacing();
```
dzenanz: Isn't `&` and `*` alignment with type's name? My memory might be wrong.

Examples/DataRepresentation/Image/Image4.cxx
dzenanz: Since clang-format doesn't distinguish between in-class and in-function declarations, maybe we should abandon alignment of declarations? Alignment will increase diff sizes for simple and style changes - not problematic, purely annoying.

Examples/DataRepresentation/Image/ImageAdaptor1.cxx
```diff
@@ -81,87 +83,88 @@ class CastPixelAccessor
...
 
-//  Software Guide : BeginLatex
-//
-//  The CastPixelAccessor class simply applies a
-//  \code{static\_cast} to the pixel values. We now use this pixel accessor
-//  to define the image adaptor type and create an instance using
-//  the standard \code{New()} method.
-//
-//  Software Guide : EndLatex
+  //  Software Guide : BeginLatex
```
dzenanz: Does this indentation interfere with the software guide?

Examples/DataRepresentation/Image/ImageAdaptor2.cxx
```diff
@@ -145,9 +146,7 @@ int main( int argc, char *argv[] )
 
   // Software Guide : BeginCodeSnippet
   using OutputImageType = itk::Image< unsigned char, Dimension >;
-  using RescalerType = itk::RescaleIntensityImageFilter<
-                                            ImageAdaptorType,
-                                            OutputImageType >;
+  using RescalerType = itk::RescaleIntensityImageFilter< ImageAdaptorType, OutputImageType >;
```
dzenanz: Some of these might be manual edits to keep the lines from going off the page in LaTeX. Maybe we can add additional `.clang-format` file in the examples directory to treat `// Software Guide : BeginCodeSnippet` and `// Software Guide : EndCodeSnippet` as special format on/off "macros".

Modules/Core/Common/include/itkMultiThreaderBase.h
```diff
@@ -61,60 +61,75 @@ class ProcessObject;
...
    * Deprecated: use Get/Set GlobalDefaultThreader. */
-  itkLegacyMacro(static void SetGlobalDefaultUseThreadPool( const bool GlobalDefaultUseThreadPool ));
-  itkLegacyMacro(static bool GetGlobalDefaultUseThreadPool( ));
+  itkLegacyMacro( static void SetGlobalDefaultUseThreadPool( const bool GlobalDefaultUseThreadPool ) );
+  itkLegacyMacro( static bool GetGlobalDefaultUseThreadPool() );
 
   /** Currently supported types of multi-threader implementations.
    * Last will change with additional implementations. */
-  enum ThreaderType { Platform = 0, First = Platform, Pool, TBB, Last = TBB, Unknown = -1 };
+  enum ThreaderType
```
dzenanz: Weren't `enum`s now supposed to be `enum class`?

Modules/Core/Common/include/itkMultiThreaderBase.h
```diff
@@ -173,28 +189,32 @@ class ITKCommon_EXPORT MultiThreaderBase : public Object
...
-  struct ThreadInfoStruct
-#else
-  struct [[deprecated( "Use WorkUnitInfo, ThreadInfoStruct is deprecated since ITK 5.0" )]] ThreadInfoStruct
-#endif
-CLANG_PRAGMA_POP
-INTEL_PRAGMA_WARN_POP
+  ITK_GCC_PRAGMA_DIAG_PUSH()
+  ITK_GCC_PRAGMA_DIAG( ignored "-Wattributes" )
+  INTEL_PRAGMA_WARN_PUSH
+  INTEL_SUPPRESS_warning_1292 CLANG_PRAGMA_PUSH CLANG_SUPPRESS_Wc__14_extensions
```
dzenanz: This change definitely looks wrong.

Modules/Core/Common/include/itkMultiThreaderBase.h
```diff
@@ -173,28 +189,32 @@ class ITKCommon_EXPORT MultiThreaderBase : public Object
...
+#  ifdef ITK_LEGACY_SILENT
+    struct ThreadInfoStruct
+#  else
+    struct [[deprecated( "Use WorkUnitInfo, ThreadInfoStruct is deprecated since ITK 5.0" )]] ThreadInfoStruct
+#  endif
+      CLANG_PRAGMA_POP INTEL_PRAGMA_WARN_POP
```
dzenanz: This change too.

Modules/Core/Common/include/itkMultiThreaderBase.h
```diff
@@ -243,137 +263,133 @@ ITK_GCC_PRAGMA_DIAG_POP()
...
   /** Break up region into smaller chunks, and call the function with chunks as parameters.
    * If filter argument is not nullptr, this function will update its progress
    * as each work unit is completed. Delegates work to non-templated version. */
-  template<unsigned int VDimension>
-  ITK_TEMPLATE_EXPORT void ParallelizeImageRegion(const ImageRegion<VDimension> & requestedRegion,
-                                     TemplatedThreadingFunctorType<VDimension> funcP,
-                                     ProcessObject* filter)
+  template < unsigned int VDimension >
+  ITK_TEMPLATE_EXPORT void
+  ParallelizeImageRegion( const ImageRegion< VDimension > &           requestedRegion,
```
dzenanz: This alignment of parameter names is weird, and perhaps counterproductive. I guess there is no setting which says: if parameters cannot all fit on a single line, make each parameter have its own line? If there is, can we have that please?

Examples/DataRepresentation/Image/Image4.cxx
hjmjohnson: The code is consistent.  I don't know if there was a style.

`PointerAlignment: Middle` --> `PointerAlignment: Left` ?

Examples/DataRepresentation/Image/Image4.cxx
hjmjohnson: Abandoned in new code.

Examples/DataRepresentation/Image/ImageAdaptor1.cxx
hjmjohnson: I don't think so, but we could test.

Modules/Core/Common/include/itkMultiThreaderBase.h
hjmjohnson: Not uniformly.  

This one looks like it could be upon first glance.  Out of scope for this PR. Please make an issue and we'll update later.

Examples/DataRepresentation/Image/ImageAdaptor2.cxx
hjmjohnson: This is an interesting idea.  Can we make the fixing of the software guide a separate issue to be followed up on after the initial commit?

Modules/Core/Common/include/itkMultiThreaderBase.h
hjmjohnson: Out of scope for this PR.  See #1194

Examples/DataRepresentation/Image/Image4.cxx
hjmjohnson: I prefer "Middle"

Modules/Core/Common/include/itkMultiThreaderBase.h
hjmjohnson: What looks wrong about this?  It looks fine to me.

Modules/Core/Common/include/itkMultiThreaderBase.h
hjmjohnson: Why?

Modules/Core/Common/include/itkMultiThreaderBase.h
hjmjohnson: There is!

`
AllowAllParametersOfDeclarationOnNextLine (bool)
If the function declaration doesn’t fit on a line, allow putting all parameters of a function declaration onto the next line even if BinPackParameters is false.
`

Modules/Core/Common/include/itkProcessObject.h
```diff
@@ -411,17 +436,28 @@ class ITKCommon_EXPORT ProcessObject:public Object
...
    * ReleaseDataBeforeUpdateFlag. */
-  virtual void SetReleaseDataFlag(bool flag);
-  virtual bool GetReleaseDataFlag() const;
-  void ReleaseDataFlagOn() { this->SetReleaseDataFlag(true); }
-  void ReleaseDataFlagOff() { this->SetReleaseDataFlag(false); }
+  virtual void
+  SetReleaseDataFlag(bool flag);
+  virtual bool
+  GetReleaseDataFlag() const;
+  void
```
dzenanz: Here, clang-format turns one-liners into five-liners.

Modules/Core/Common/include/itkProcessObject.h
```diff
@@ -443,38 +479,38 @@ class ITKCommon_EXPORT ProcessObject:public Object
   itkSetClampMacro(NumberOfWorkUnits, ThreadIdType, 1, ITK_MAX_THREADS);
   itkGetConstReferenceMacro(NumberOfWorkUnits, ThreadIdType);
 
-#if !defined( ITK_LEGACY_REMOVE ) || defined( ITKV4_COMPATIBILITY )
-  itkLegacyMacro(void SetNumberOfThreads(ThreadIdType count))
-  {
-    this->SetNumberOfWorkUnits(count);
-  }
+#if !defined(ITK_LEGACY_REMOVE) || defined(ITKV4_COMPATIBILITY)
+  itkLegacyMacro(void SetNumberOfThreads(ThreadIdType count)) { this->SetNumberOfWorkUnits(count); }
```
dzenanz: Here it does the opposite. I guess the macro confuses it.

Modules/Core/TestKernel/include/itkRandomImageSource.h
```diff
@@ -92,70 +92,77 @@ class ITK_TEMPLATE_EXPORT RandomImageSource:public ImageSource< TOutputImage >
...
-  void GenerateOutputInformation() override;
+  void
+  GenerateOutputInformation() override;
 
 private:
-  SizeType      m_Size;      //size of the output image
-  SpacingType   m_Spacing;   //spacing
-  PointType     m_Origin;    //origin
-  DirectionType m_Direction; //direction
+  SizeType m_Size;           // size of the output image
```
dzenanz: Comments are aligned. Interesting!

Modules/Core/Common/include/itkImageSource.hxx
```diff
@@ -233,17 +219,19 @@ ImageSource< TOutputImage >
...
-    {
+  {
     this->GetMultiThreader()->SetNumberOfWorkUnits(this->GetNumberOfWorkUnits());
     this->GetMultiThreader()->template ParallelizeImageRegion<OutputImageDimension>(
-        this->GetOutput()->GetRequestedRegion(),
-        [this](const OutputImageRegionType & outputRegionForThread)
-          { this->DynamicThreadedGenerateData(outputRegionForThread); }, this);
-    }
+      this->GetOutput()->GetRequestedRegion(),
+      [this](const OutputImageRegionType & outputRegionForThread) {
```
dzenanz: Lambda braces inconsistent with other brace indentation.

Modules/Core/Common/include/itkImageSource.hxx
```diff
@@ -252,28 +240,27 @@ ImageSource< TOutputImage >
...
-template< typename TOutputImage >
+template <typename TOutputImage>
 void
-ImageSource< TOutputImage >
-::ThreadedGenerateData(const OutputImageRegionType &
-#if ! defined ( ITK_LEGACY_REMOVE )
-    region
+ImageSource<TOutputImage>::ThreadedGenerateData(const OutputImageRegionType &
+#if !defined(ITK_LEGACY_REMOVE)
+                                                  region
```
dzenanz: I wonder what is this "region" aligned with?

Modules/Core/Common/include/itkMultiThreaderBase.h
dzenanz: Don't macros each need a line of its own? If not, why do preceding macros each get a line all to themselves?

dzenanz:
@imikejackson Review welcome.

Modules/Core/Common/include/itkMultiThreaderBase.h
```diff
@@ -244,136 +264,132 @@ ITK_GCC_PRAGMA_DIAG_POP()
...
         SplitDimension,
         splitRegion.GetIndex().m_InternalArray,
         splitRegion.GetSize().m_InternalArray,
-        [&](const IndexValueType index[], const SizeValueType size[])
-        {
-        ImageRegion<VDimension> restrictedRequestedRegion;
-        restrictedRequestedRegion.SetIndex( restrictedDirection, requestedRegion.GetIndex( restrictedDirection ) );
-        restrictedRequestedRegion.SetSize( restrictedDirection, requestedRegion.GetSize( restrictedDirection ) );
-        for( unsigned int splitDimension = 0, dimension = 0; dimension < VDimension; ++dimension )
+        [&](const IndexValueType index[], const SizeValueType size[]) {
```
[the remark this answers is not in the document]
hjmjohnson: @dzenanz Do you have a recommendation for a clang-format change that you propose?

This is not intentional, but perhaps an oversight of a better option?

http://releases.llvm.org/8.0.0/tools/clang/docs/ClangFormatStyleOptions.html

Modules/Core/Common/include/itkProcessObject.h
hjmjohnson: I only see 2 lines.

Modules/Core/TestKernel/include/itkRandomImageSource.h
hjmjohnson: `
AlignTrailingComments (bool)
If true, aligns trailing comments.
`

Modules/Core/Common/include/itkMultiThreaderBase.h
hjmjohnson: Ahh, I see.  This may be a rare case where we need
// clang-format off

// clang-format on

-- The macros that are now nested are defined as empty
```
#  define INTEL_SUPPRESS_warning_1292
#  define CLANG_PRAGMA_PUSH
#  define CLANG_PRAGMA_POP
#  define CLANG_SUPPRESS_Wfloat_equal
#  define CLANG_SUPPRESS_Wc__14_extensions
```
On this platform, so they are can be nested for this preprocessor options.

Examples/DataRepresentation/Image/Image4.cxx
dzenanz: As I was initially reviewing, you were making changes. Parallel synchronization of humans is non-trivial 🤖

Examples/DataRepresentation/Image/ImageAdaptor2.cxx
dzenanz: Created issue https://github.com/InsightSoftwareConsortium/ITK/issues/1195

hjmjohnson:
@lassoan, @dzenanz, @N-Dekker , @gdevenyi , @blowekamp , @seanm , @bradking Thank you everyone for the reviews.  I really appreciate your help.  I have updated the PR trying to incorporate as much of the discussion as possible into this set of patches.

We are are a nice low point in outstanding PR's, and it would be great to make this change during this relative down-time.

Examples/.clang-format
```diff
@@ -0,0 +1,150 @@
...
+  SplitEmptyNamespace: false
+BreakBeforeBinaryOperators: None
+#clang 6.0 BreakBeforeInheritanceComma: true
+BreakInheritanceList: BeforeComma
+BreakBeforeTernaryOperators: true
+#clang 6.0 BreakConstructorInitializersBeforeComma: true
+BreakConstructorInitializers: BeforeComma
+BreakAfterJavaFieldAnnotations: false
+BreakStringLiterals: true
+## The following line is the only line that should be different from the ../.clang-format file
```
dzenanz: I think that clang-format uses hierarchical configuration. Meaning, this file could only have the next line and line 19 (`Language: Cpp`). Everything else should be inherited from parent folder(s). That way we will only have one important configuration file.

Examples/.clang-format
```diff
@@ -0,0 +1,150 @@
...
+ColumnLimit:     88
```
dzenanz: Did you get 88 somewhere from the software guide? Otherwise 80 or 78 might be a more logical choice.

Modules/Core/Common/include/itkMultiThreaderBase.h
```diff
@@ -213,7 +220,9 @@ class ITKCommon_EXPORT MultiThreaderBase : public Object
       UNKNOWN
     } ThreadExitCode;
   };
-  ITK_GCC_PRAGMA_DIAG_POP()
+// clang-format off
+ITK_GCC_PRAGMA_DIAG_POP()
```
dzenanz: What can this macro nest with? Perhaps `// clang-format off` is not needed for this instance.

Modules/Segmentation/Watersheds/include/itkWatershedSegmenter.hxx
```diff
@@ -176,9 +176,17 @@ Segmenter<TInputImage>::GenerateData()
...
+// clang-format on
+CLANG_PRAGMA_PUSH
+CLANG_SUPPRESS_Wfloat_equal
+// clang-format off
+      &&
       maximum == NumericTraits<InputPixelType>::max())
-    CLANG_PRAGMA_POP { maximum -= NumericTraits<InputPixelType>::OneValue(); }
+// clang-format on
+CLANG_PRAGMA_POP
+// clang-format off
```
dzenanz: Not at risk of nesting?

Modules/Segmentation/Watersheds/include/itkWatershedSegmenter.hxx
```diff
@@ -176,9 +176,17 @@ Segmenter<TInputImage>::GenerateData()
   Self::MinMax(input, regionToProcess, minimum, maximum);
   // cap the maximum in the image so that we can always define a pixel
   // value that is one greater than the maximum value in the image.
-  if (NumericTraits<InputPixelType>::IsInteger CLANG_PRAGMA_PUSH CLANG_SUPPRESS_Wfloat_equal &&
+  if (NumericTraits<InputPixelType>::IsInteger
+// clang-format on
+CLANG_PRAGMA_PUSH
```
dzenanz: `ITK_PRAGMA_PUSH` is normally followed by another macro, whereas `ITK_PRAGMA_POP` is usually alone.

.clang-format
```diff
@@ -8,9 +8,9 @@
 #  BasedOnStyle: Mozilla
```
dzenanz: We could have this uncommented.

dzenanz:
Here is a poll for pointer alignment: https://doodle.com/poll/yn8bxxtdzz97xg38

dzenanz:
Here is a poll for return type on its own line: https://doodle.com/poll/ktqw5tgvt83hniwe

Modules/Core/Common/include/itkMultiThreaderBase.h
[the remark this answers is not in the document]
dzenanz: Yes, lambdas have no options in clang format 8.0 😞

jhlegarreta:
Sorry, I've been away from keyboard, and it's been a while since I have not contributed.

Thanks to all of you who have participated and commented.

I generally agree on the idea of reaching an agreement, and enforcing/sticking to a style.

I have gone through the comments, and two things are clear: we agree that we will disagree in many aspects of the style, and we agree that, as @hjmjohnson and @blowekamp have explicited, we would like to focus on methods, and have the style be enforced automatically.

Now, I think that it takes a CI service for the style check. I currently do not know how this could be done.

jhlegarreta:
@seanm if by [your comment](https://github.com/InsightSoftwareConsortium/ITK/pull/1191#issuecomment-524331564) you mean how to skip the pre-commit hook, this can be done using the `-n` or `--no-verify` options of the `git commit` command.

Modules/Segmentation/Watersheds/include/itkWatershedSegmenter.hxx
hjmjohnson: True.  What are you proposing?

Modules/Segmentation/Watersheds/include/itkWatershedSegmenter.hxx
hjmjohnson: Perhaps not, but macros that are required to be on their own line when defined, or irrelevant when empty,  should be protected with clang-format guards.

Examples/.clang-format
hjmjohnson: :) . 88 is the line length identified by the python community "black" project as having an optimal (linelength/numberoflines) ratio for a very large number of python projects.   Python projects tend to use long variable names (like ITK does), and tends to have long chained functions like ITK does.

The historical 80 characters is a standard because the IBM punch card had 80 columns.  This seems like an arbitrary choice with modern computers.  

Ideally, we should choose a line length that best fits the latex rendering in the software guide.  That may be 60 characters, or 94 characters.

hjmjohnson: Followup:  It seems that about 91 characters will fit in the latex rendering of code for the software guide.

hjmjohnson:
> Here is a poll for return type on its own line: https://doodle.com/poll/ktqw5tgvt83hniwe

I'm ok putting return types on one line.  
PRO: Visually I think it looks better for short functions.
CON: Writing code requires counting characters to determine if the return goes on a new line or not.  It really requires an automated solution in order to enforce.  The "return is always on it's own line" setting is easier to manually enforce.

.clang-format
```diff
@@ -5,12 +5,12 @@
...
 # This configuration requires clang-format version 8.0.0 exactly.
-#  BasedOnStyle: Mozilla
+BasedOnStyle: Mozilla
 Language:        Cpp
 AccessModifierOffset: -2
-AlignAfterOpenBracket: Align
+AlignAfterOpenBracket: DontAlign
 AlignConsecutiveAssignments: false
-AlignConsecutiveDeclarations: true
+AlignConsecutiveDeclarations: false
```
thewtex: I strongly disagree with this change. Readability is paramount. We read the code hundreds of times more than when we write it. Readability was worth aligning the code before -- now we will have an automated way to keep the code readable. The code is much more important than the diffs.

N-Dekker:
Before merging this pull request, please consider Kevlin Henney "Seven Ineffective Coding Habits of Many Programmers": https://www.youtube.com/watch?v=ZsHMHukIlJY

phcerdan:
This is outstanding @hjmjohnson! I have been enjoying clang-format recently in personal projects, and forgetting about style (or only think about it once) is a nice productivity boost.
It's impossible to have the perfect style for everyone and for every case, but the gain in automation is huge.

I agree with @thewtex, issues with `git blame` due to alignment are going to be minimal, and when they happen, should be straight forward to spot the important change. The annoying part of the alignment is that the user had to do it after any change, if we automatize it, the main issue is gone.

The problem with clang-format changing the output depending on the version is important. @bradking mentioned that they enforce a version (6.0) in CMake. I would propose to live at head, and use the latest clang-format instead (or at least the last release), but I understand if we enforce 6.0 for now.

This is great, looking forward to use it!

By the way, I am using this vim plugin for clang-format: https://github.com/rhysd/vim-clang-format

With this mapping:
```
" <Leader> is Space in my case
" let mapleader=" "
autocmd FileType c,cpp,objc nnoremap <buffer><Leader>x :<C-u>ClangFormat<CR>
autocmd FileType c,cpp,objc vnoremap <buffer><Leader>x :ClangFormat<CR>
```
And it is awesome! If the files are relatively small, I even `:ClangFormatAutoEnable`, which applies the format at every save.

N-Dekker:
> Here is a poll for return type on its own line: https://doodle.com/poll/ktqw5tgvt83hniwe

Nice polls, @dzenanz ! Can you please add a "neutral" option to the next poll(s)? In some cases I would like to indicated that I'm fine with each possible outcome.

Sometimes it might even be useful to distinguish between "strongly in favor" and "weakly in favor" of a certain outcome.

Modules/Segmentation/Watersheds/include/itkWatershedSegmenter.hxx
dzenanz: I thought that `// clang-format on` might not be needed for this case, but now I see that the macro gets put on the same line with some ordinary code.

dzenanz:
@N-Dekker neutral: don't vote. That's what I did in one of the polls.

N-Dekker:
> Poll about declaration alignment: https://doodle.com/poll/sb4aqtpsy6qs7z3h

Example of declaration alignment (please check):

    typename ObjectSizeInPixelsContainerType::const_iterator it;
    ObjectSizeInPhysicalUnitsContainerType::const_iterator   fit;
    LabelType                                                i;

From https://github.com/InsightSoftwareConsortium/ITK/blob/v5.0.1/Modules/Segmentation/ConnectedComponents/include/itkRelabelComponentImageFilter.hxx#L240-L242

hjmjohnson:
@thewtex All Green!

hjmjohnson:
@thewtex @dzenanz I have rebased on the merges from this morning.  Last nights builds were clean green from the CI perspective.

This PR required a manual merging of 2 small blocks of conflicts in from the DCMTK merge that occurred this morning.

I tested, and the build succeeded, and all tests pass.

As this has 2 approvals, I'd really like to "rebase and merge" immediately (i.e. not wait 5 hours for another green CI for the small formatting fixes induced by the DCMTK code merges ) so that new PR's can be generated against this rigorously enforced code style.

Please provide a +1 or just rebase and merge at your earliest convenience,
Hans

dzenanz:
In some of the previous builds there was a warning about multi-line comment. Was that also addressed? Not a showstopper if it wasn't. But as there are no other PRs which are ready for merging we could wait a few hours for the CI builds too.

hjmjohnson:
Yes it was addressed last night.

 

A “// clang-format off” “// clang-format on” protection around a comment that had a large latex table imbedded in a comment block.

 

Hans

PR #1198, opened 2019-08-24
BUG: Resolve corner case bug in ComputeTangentsAndNormals in TubeSO

aylward:
TubeSpatialObjects can computer their tangets and normals in a frenet
frame, but normal directions can be ambiguous along a straight line, so
heuristics are used.   This fix adds to those heuristics so that they
work on a tube that is perfectly straight in the z-direction.
Previously such cases would produce NAN normals.

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkTubeSpatialObject.hxx
+++ b/Modules/Core/SpatialObjects/include/itkTubeSpatialObject.hxx
@@ -48,7 +48,7 @@ TubeSpatialObject<TDimension, TTubePointType>::Clear(void)
 
   m_Root = false;
   m_ParentPoint = -1;
-  m_EndRounded = false; // default end-type is flat
+  m_EndRounded = true; // default end-type is flat
 
   this->Modified();
 }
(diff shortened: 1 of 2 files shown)


Modules/Core/SpatialObjects/include/itkTubeSpatialObject.hxx
```diff
@@ -345,7 +345,7 @@ TubeSpatialObject< TDimension, TTubePointType >
     l = std::sqrt(l);
     // if the adjacent points correspond, use the current point and one
     //   forward point
-    if ( Math::AlmostEquals( l, 0.0 ) )
+    if ( Math::AlmostEquals( l, 0.0 ) || l != l )
```
hjmjohnson: `||    l != l`

This seems like an always false statement that will cause compiler warnings about autological comparisons.

Was this used for debugging?

dzenanz: Using `std::isnan` is far clearer, if that is what you wanted to check.

aylward:
I will change this to isnan as Dzenan suggested.

Thanks!

S

---
Sent via phone.

Modules/Core/SpatialObjects/include/itkTubeSpatialObject.hxx
aylward: Changed to std::isnan() call.

PR #1203, opened 2019-08-26
ENH: Download clang-format 8.0.1 executable

thewtex:
A clang-format executable for all major platforms is downloaded with
development builds with a version compatible with ITK's clang-format
configuration, currently 8.0.1. Git is configured with the executable's
path.

In the future, this will help automatically apply ITK's Coding Style
Guidelines in Git hooks and aliases.

### DIFF
--- a/CMake/ITKModuleClangFormat.cmake
+++ b/CMake/ITKModuleClangFormat.cmake
@@ -1,0 +1,15 @@
+# Download and configure clang-format for style uniformity during development.
+#
+# This code helps download a version of clang-format binaries consistent with
+# ITK's clang-format configuration. Git configuration is added during
+# development to help automatically apply the formatting.
+#
+# Globally, the ITK/.clang-format file provides the base style configuration.
+# For more information on ITK coding style, see the ITK Coding Style Guide in
+# the ITK Software Guide.
+option(ITK_USE_CLANG_FORMAT "Enable the use of clang-format for coding style formatting." ${BUILD_TESTING})
+mark_as_advanced(ITK_USE_CLANG_FORMAT)
+
+if(BUILD_TESTING AND ITK_USE_CLANG_FORMAT AND NOT CMAKE_CROSSCOMPILING)
+  include(${ITK_CMAKE_DIR}/../Utilities/ClangFormat/DownloadClangFormat.cmake)
+endif()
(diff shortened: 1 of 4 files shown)


CMake/ITKModuleClangFormat.cmake
```diff
@@ -0,0 +1,15 @@
+# Download and configure clang-format for style uniformity during development.
+#
+# This code helps download a version of clang-format binaries consistent with
+# ITK's clang-format configuration. Git configuration is added during
+# development to help automatically apply the formatting.
+#
+# Globally, the ITK/.clang-format file provides the base style configuration.
+# For more information on ITK coding style, see the ITK Coding Style Guide in
+# the ITK Software Guide.
+option(ITK_USE_CLANG_FORMAT "Enable the use of clang-format for coding style formatting." ${BUILD_TESTING})
```
jhlegarreta, line 10: @thewtex Should this be `ITK_USE_CLANGFORMAT`, cross-referencing
https://github.com/InsightSoftwareConsortium/ITK/blob/master/CMake/ITKClangFormatSetup.cmake#L11
or else, should the later be renamed to `ITK_USE_CLANG_FORMAT`?

I've just realized that both options are available in CMake, and I am not sure about their difference. Thanks

dzenanz: I think one option is for downloading it, and the other for using it.

jhlegarreta: > I think one option is for downloading it, and the other for using it.

Then maybe the variable names and their descriptions should honor better their purpose?

dzenanz: I am not entirely sure of the distinction. If you could double-check it, feel free to update the names and descriptions.

thewtex: @jhlegarreta good catch! They were intended to be the same. Patch here:https://github.com/InsightSoftwareConsortium/ITK/pull/2594

jhlegarreta: Thanks both for caring about this. Resolving.

N-Dekker:
@thewtex @hjmjohnson @dzenanz  Would it be an idea to upgrade clang-format before the final release of ITK 5.4.0? I got the impression that clang-format 8.0.1 does not recognize all the latest C++17 syntax anymore. For example, those CTAD guides look a bit funny, as clang-format 8.0.1 has removed the surrounding spaces from `->`, in a line of code like:

    ImageBufferRange(TImage &) -> ImageBufferRange<TImage>

https://github.com/InsightSoftwareConsortium/ITK/blob/65306d5f6230eb3c5dc9a5df89a4740ba383041e/Modules/Core/Common/include/itkImageBufferRange.h#L662-L663



I suspect that an upgrade of clang-format will cause _many_ cosmetic code changes. So maybe it's better to do it _before_ the upcoming major release. Otherwise it may be harder to merge patches between the master branch and the next release maintenance branch.

Anyway, I'm just asking. I'm not volunteering myself, I just hope one of you can do it 🤷

blowekamp:
> Otherwise it may be harder to merge patches between the master branch and the next release maintenance branch.

I agree.

PR #1218, opened 2019-08-30
BUG: Wrong component type for non supported tiff with USHORT palette

YannLePoul:
## PR Checklist

- [] :no_entry_sign: [Makes breaking changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes)
- [] :no_entry_sign: [Makes design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes)
- [] :no_entry_sign: Adds the License notice to new files.
- [] :no_entry_sign: Adds Python wrapping.
- [] :no_entry_sign: Adds tests and baseline comparison (quantitative).
- [] :no_entry_sign: [Adds test data](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/UploadBinaryData.md).
- [] :no_entry_sign: Adds Examples to [ITKExamples](https://github.com/InsightSoftwareConsortium/ITKExamples)
- [] :no_entry_sign: Adds Documentation.

### DIFF
--- a/Modules/IO/TIFF/src/itkTIFFImageIO.cxx
+++ b/Modules/IO/TIFF/src/itkTIFFImageIO.cxx
@@ -456,10 +456,12 @@ TIFFImageIO::ReadImageInformation()
       this->SetPixelType(RGBA);
   }
 
-  if ((this->GetFormat() == TIFFImageIO::PALETTE_GRAYSCALE || this->GetFormat() == TIFFImageIO::PALETTE_RGB) &&
-      m_TotalColors > 0 && this->GetExpandRGBPalette())
+  bool isPalette =
+    (this->GetFormat() == TIFFImageIO::PALETTE_GRAYSCALE || this->GetFormat() == TIFFImageIO::PALETTE_RGB) &&
+    m_TotalColors > 0;
+  bool isPaletteShortType = false;
+  if (isPalette)
   {
-    m_ComponentType = UCHAR;
     // detect if palette appears to be 8-bit or 16-bit
     for (unsigned int cc = 0; cc < 256; ++cc)
     {
@@ -467,15 +469,16 @@ TIFFImageIO::ReadImageInformation()
       this->GetColor(cc, &red, &green, &blue);
       if (red > 255 || green > 255 || blue > 255)
       {
-        m_ComponentType = USHORT;
+        isPaletteShortType = true;
         break;
       }
     }
   }
-  if (!m_IsReadAsScalarPlusPalette)
+
+  if (isPalette && this->GetExpandRGBPalette())
   {
-    // make sure the palette is empty
-    m_ColorPalette.resize(0);
+    // put the component type the same as the palette type
+    m_ComponentType = (isPaletteShortType) ? USHORT : UCHAR;
   }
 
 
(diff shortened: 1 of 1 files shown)


thewtex:
Thanks @YannLePoul !

CI identifies build errors.

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
```diff
@@ -456,21 +456,33 @@ TIFFImageIO::ReadImageInformation()
       this->SetPixelType(RGBA);
   }
 
+  bool isPaletteShortType = false;
```
hjmjohnson: I think this logic can be simpler and avoid the extra isPaletteShortType variable.

The logic before line 474 could be reverted.
m_ComponentType should not be changed after line 474.   (i.e. line 503 was in error before).

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
```diff
@@ -496,16 +508,25 @@ TIFFImageIO::ReadImageInformation()
...
+      this->SetNumberOfComponents(3);
+      this->SetPixelType(RGB); // RGBA is not valid for palette, so cannot be
+      if (isPaletteShortType)
+      {
+        m_ComponentType = USHORT;
+      }
+    }
+    else
+    {
+      tkDebugMacro(<< "Using TIFFReadRGBAImage");
```
jhlegarreta: Typo: _tkDebugMacro_ -> _itkDebugMacro_

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
```diff
@@ -456,21 +456,33 @@ TIFFImageIO::ReadImageInformation()
...
   }
 
+  bool isPaletteShortType = false;
   if ((this->GetFormat() == TIFFImageIO::PALETTE_GRAYSCALE || this->GetFormat() == TIFFImageIO::PALETTE_RGB) &&
       m_TotalColors > 0 && this->GetExpandRGBPalette())
+
   {
-    m_ComponentType = UCHAR;
     // detect if palette appears to be 8-bit or 16-bit
     for (unsigned int cc = 0; cc < 256; ++cc)
```
jhlegarreta: Although the 256 the 8-bit grayscale values can be identified automatically, I am wondering whether we could use `UCHAR_MAX` instead, or 2^sizeof(UCHAR). @hjmjohnson, @thewtex Is the proposal adding more generalization than necessary? Should may be done in another PR?

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
```diff
@@ -456,21 +456,33 @@ TIFFImageIO::ReadImageInformation()
...
     {
       unsigned short red, green, blue;
       this->GetColor(cc, &red, &green, &blue);
       if (red > 255 || green > 255 || blue > 255)
```
jhlegarreta: Same comment as above.

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
YannLePoul: I'm not sure to understand the problem, I hope I'm not missunderstanding you, but here, the first 256 correspond to the maximum number of colors in the palette of a tiff file. Tiff palette is saved in the file as 256 x 3 integers of  2^16 bits whether they are actually unsigned char or not (https://www.awaresystems.be/imaging/tiff/tifftags/colormap.html) so I guess it doesn't depends on the platform.

jhlegarreta: @YannLePoul thanks for the explanation. Looks like I was then confused by the `m_ComponentType = UCHAR;` statement that your PR removes. Reading the comment line just above clarifies things :sweat_smile: 

Resolving the conversation.

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
YannLePoul: You must be right. To be honest, I wrote this code 2-3 years ago. I'm going to run some test on a simplified version, and amend the pull request!

YannLePoul:
Ok, I rewrote the thing. To state clearly the problem, if canRead is false, the image was read as RGBA. But If a expand_palette is false, the component type was not set to the palette type, and was kept as the component type. 
So here we have to detect the palette type whatever expand_palette is true or not (to be able to reset it to the right type), and change the type if is palette and expand_palette is true, or if canRead is false, and expand_palette is false, to be reset it to the correct RGB component_type.

I didn't see a simpler way to do that.

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
YannLePoul: done!

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
```diff
@@ -456,26 +456,29 @@ TIFFImageIO::ReadImageInformation()
       this->SetPixelType(RGBA);
   }
 
-  if ((this->GetFormat() == TIFFImageIO::PALETTE_GRAYSCALE || this->GetFormat() == TIFFImageIO::PALETTE_RGB) &&
-      m_TotalColors > 0 && this->GetExpandRGBPalette())
```
blowekamp: Has the check for `ExpandRGBPalette` been intentionally removed?

YannLePoul: Yes, because with a non supported palette file and GetExpandRGBPalette to false, we have to reset to the proper RGB component type (line 518)

PR #1222, opened 2019-09-02
ENH: update remote modules to comply with new enum class style

dzenanz:
There are probably some other changes in some of the remotes. I didn't check whether all the modules have the new changes.


dzenanz:
The failing test seems unrelated. Merging.

PR #1236, opened 2019-09-05
COMP: avoid integer type conversion and use static_cast for necessary…

YannLePoul:
- use static_cast insted of old style cast
- use variable type that avoid integer conversion
- use directly native libTiff types when possible
- makes sure no overflow happens when counting the number of colors in the palette

## PR Checklist

- [] :no_entry_sign: [Makes breaking changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes)
- [] :no_entry_sign: [Makes design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes)
- [] :no_entry_sign: Adds the License notice to new files.
- [] :no_entry_sign: Adds Python wrapping.
- [] :no_entry_sign: Adds tests and baseline comparison (quantitative).
- [] :no_entry_sign: [Adds test data](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/UploadBinaryData.md).
- [] :no_entry_sign: Adds Examples to [ITKExamples](https://github.com/InsightSoftwareConsortium/ITKExamples)
- [] :no_entry_sign: Adds Documentation.

### DIFF
--- a/Modules/IO/TIFF/src/itkTIFFImageIO.cxx
+++ b/Modules/IO/TIFF/src/itkTIFFImageIO.cxx
@@ -78,7 +78,7 @@ TIFFImageIO::ReadGenericImage(void * out, unsigned int width, unsigned int heigh
 }
 
 void
-TIFFImageIO::GetColor(unsigned int index, unsigned short * red, unsigned short * green, unsigned short * blue)
+TIFFImageIO::GetColor(uint64_t index, uint16_t * red, uint16_t * green, uint16_t * blue)
 {
   *red = 0;
   *green = 0;
@@ -119,9 +119,9 @@ TIFFImageIO::GetFormat()
       {
         if (this->GetExpandRGBPalette())
         {
-          for (unsigned int cc = 0; cc < static_cast<unsigned int>(m_TotalColors); ++cc)
+          for (uint64_t cc = 0; cc < m_TotalColors; ++cc)
           {
-            unsigned short red, green, blue;
+            uint16_t red, green, blue;
             this->GetColor(cc, &red, &green, &blue);
             if (red != green || red != blue)
             {
@@ -148,10 +148,10 @@ TIFFImageIO::GetFormat()
 void
 TIFFImageIO::ReadVolume(void * buffer)
 {
-  const int width = m_InternalImage->m_Width;
-  const int height = m_InternalImage->m_Height;
+  const size_t width{ m_InternalImage->m_Width };
+  const size_t height{ m_InternalImage->m_Height };
 
-  for (unsigned int page = 0; page < m_InternalImage->m_NumberOfPages; page++)
+  for (uint16 page = 0; page < m_InternalImage->m_NumberOfPages; page++)
   {
     if (m_InternalImage->m_IgnoredSubFiles > 0)
     {
@@ -168,8 +168,7 @@ TIFFImageIO::ReadVolume(void * buffer)
     }
 
 
-    const size_t pixelOffset = static_cast<size_t>(width) * static_cast<size_t>(height) *
-                               static_cast<size_t>(this->GetNumberOfComponents()) * static_cast<size_t>(page);
+    const size_t pixelOffset = width * height * this->GetNumberOfComponents() * page;
 
     ReadCurrentPage(buffer, pixelOffset);
 
@@ -297,7 +296,7 @@ TIFFImageIO::InitializeColors()
   m_ColorRed = nullptr;
   m_ColorGreen = nullptr;
   m_ColorBlue = nullptr;
-  m_TotalColors = -1;
+  m_TotalColors = 0;
   m_ImageFormat = TIFFImageIO::NOFORMAT;
 
   if (m_InternalImage == nullptr)
@@ -323,7 +322,7 @@ TIFFImageIO::InitializeColors()
       itkExceptionMacro(<< "Sorry, can not handle image with " << m_InternalImage->m_BitsPerSample << "-bit samples");
   }
 
-  m_TotalColors = (1L << m_InternalImage->m_BitsPerSample);
+  m_TotalColors = uint64_t{ 1 } << m_InternalImage->m_BitsPerSample;
 
   m_ColorRed = red_orig;
   m_ColorGreen = green_orig;
@@ -374,24 +373,21 @@ TIFFImageIO::ReadImageInformation()
   {
     if (m_InternalImage->m_ResolutionUnit == 2) // inches
     {
-      m_Spacing[0] = 25.4 / m_InternalImage->m_XResolution;
-      m_Spacing[1] = 25.4 / m_InternalImage->m_YResolution;
+      m_Spacing[0] = 25.4 / static_cast<double>(m_InternalImage->m_XResolution);
+      m_Spacing[1] = 25.4 / static_cast<double>(m_InternalImage->m_YResolution);
     }
     else if (m_InternalImage->m_ResolutionUnit == 3) // cm
     {
-      m_Spacing[0] = 10.0 / m_InternalImage->m_XResolution;
-      m_Spacing[1] = 10.0 / m_InternalImage->m_YResolution;
+      m_Spacing[0] = 10.0 / static_cast<double>(m_InternalImage->m_XResolution);
+      m_Spacing[1] = 10.0 / static_cast<double>(m_InternalImage->m_YResolution);
     }
   }
 
   m_Origin[0] = 0.0;
   m_Origin[1] = 0.0;
 
-  int width = m_InternalImage->m_Width;
-  int height = m_InternalImage->m_Height;
-
-  m_Dimensions[0] = width;
-  m_Dimensions[1] = height;
+  m_Dimensions[0] = m_InternalImage->m_Width;
+  m_Dimensions[1] = m_InternalImage->m_Height;
 
   if (m_InternalImage->m_BitsPerSample <= 8)
   {
@@ -463,9 +459,9 @@ TIFFImageIO::ReadImageInformation()
   if (isPalette)
   {
     // detect if palette appears to be 8-bit or 16-bit
-    for (unsigned int cc = 0; cc < static_cast<unsigned int>(m_TotalColors); ++cc)
+    for (uint64_t cc = 0; cc < m_TotalColors; ++cc)
     {
-      unsigned short red, green, blue;
+      uint16_t red, green, blue;
       this->GetColor(cc, &red, &green, &blue);
       if (red > 255 || green > 255 || blue > 255)
       {
@@ -559,24 +555,24 @@ TIFFImageIO::Write(const void * buffer)
 void
 TIFFImageIO::InternalWrite(const void * buffer)
 {
-  const auto * outPtr = (const char *)buffer;
+  const auto * outPtr = static_cast<const char *>(buffer);
 
-  unsigned int page, pages = 1;
+  uint16 page, pages = 1;
 
   const SizeValueType width = m_Dimensions[0];
   const SizeValueType height = m_Dimensions[1];
   if (m_NumberOfDimensions == 3)
   {
-    pages = m_Dimensions[2];
+    pages = static_cast<uint16>(m_Dimensions[2]);
   }
 
-  int  scomponents = this->GetNumberOfComponents();
-  auto resolution_x = static_cast<float>(m_Spacing[0] != 0.0 ? 25.4 / m_Spacing[0] : 0.0);
-  auto resolution_y = static_cast<float>(m_Spacing[1] != 0.0 ? 25.4 / m_Spacing[1] : 0.0);
+  auto   scomponents = static_cast<uint16>(this->GetNumberOfComponents());
+  double resolution_x{ m_Spacing[0] != 0.0 ? 25.4 / m_Spacing[0] : 0.0 };
+  double resolution_y{ m_Spacing[1] != 0.0 ? 25.4 / m_Spacing[1] : 0.0 };
   // rowsperstrip is set to a default value but modified based on the tif scanlinesize before
   // passing it into the TIFFSetField (see below).
-  auto rowsperstrip = (uint32)-1;
-  int  bps;
+  auto     rowsperstrip = uint32{ 0 };
+  uint16_t bps;
 
   switch (this->GetComponentType())
   {
@@ -636,8 +632,8 @@ TIFFImageIO::InternalWrite(const void * buffer)
     TIFFSetField(tif, TIFFTAG_SAMPLEFORMAT, SAMPLEFORMAT_IEEEFP);
   }
 
-  uint32 w = width;
-  uint32 h = height;
+  auto w = static_cast<uint32>(width);
+  auto h = static_cast<uint32>(height);
 
   if (m_NumberOfDimensions == 3)
   {
@@ -669,8 +665,7 @@ TIFFImageIO::InternalWrite(const void * buffer)
       uint16 extra_samples = scomponents - 3;
       auto * sample_info = new uint16[scomponents - 3];
       sample_info[0] = EXTRASAMPLE_ASSOCALPHA;
-      int cc;
-      for (cc = 1; cc < scomponents - 3; cc++)
+      for (uint16 cc = 1; cc < scomponents - 3; cc++)
       {
         sample_info[cc] = EXTRASAMPLE_UNSPECIFIED;
       }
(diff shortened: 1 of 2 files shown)


Modules/IO/TIFF/src/itkTIFFImageIO.cxx
```diff
@@ -86,7 +86,7 @@ TIFFImageIO::GetColor(unsigned int index, unsigned short * red, unsigned short *
 
   if (m_TotalColors > 0 && m_ColorRed && m_ColorGreen && m_ColorBlue)
   {
-    index %= m_TotalColors;
+    index %= static_cast<unsigned int>(m_TotalColors);
```
N-Dekker: This line (your added `static_cast`) looks fine to me, but I'm just wondering why `m_TotalColors` wasn't declared `unsigned`. Do you know?

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
```diff
@@ -148,8 +148,8 @@ TIFFImageIO::GetFormat()
 void
 TIFFImageIO::ReadVolume(void * buffer)
 {
-  const int width = m_InternalImage->m_Width;
-  const int height = m_InternalImage->m_Height;
+  const size_t width = m_InternalImage->m_Width;
+  const size_t height = m_InternalImage->m_Height;
```
N-Dekker: I agree with you that `width` and `height` should not be signed int, but I think they should be `auto`, or `uint32_t` (not `size_t`), in order to pass them safely to `TIFFReadRGBAImageOriented` (a few lines later). Please check!

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
```diff
@@ -323,7 +323,7 @@ TIFFImageIO::InitializeColors()
       itkExceptionMacro(<< "Sorry, can not handle image with " << m_InternalImage->m_BitsPerSample << "-bit samples");
   }
 
-  m_TotalColors = (1L << m_InternalImage->m_BitsPerSample);
+  m_TotalColors = static_cast<int>(1L << m_InternalImage->m_BitsPerSample);
```
N-Dekker: Do you know why the bit-shift was done on a `long` integer, "1L"?  This `static_cast` appears to silence a potential issue without solving it. If `int` is good enough (instead of `long`), I would think the line of code should simply look like this (without cast, and without `long` "1L"):

     m_TotalColors = (1 << m_InternalImage->m_BitsPerSample);

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
```diff
@@ -387,8 +387,8 @@ TIFFImageIO::ReadImageInformation()
   m_Origin[0] = 0.0;
   m_Origin[1] = 0.0;
 
-  int width = m_InternalImage->m_Width;
-  int height = m_InternalImage->m_Height;
+  size_t width = m_InternalImage->m_Width;
+  size_t height = m_InternalImage->m_Height;
```
N-Dekker: Again, I agree with you that `width` and `height` should not be signed int, but I'm not convinced about `size_t either, here. in this case I would not even declare them. `m_InternalImage->m_Width` and `m_Height` could be assigned directly to m_Dimensions[i], right?

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
```diff
@@ -570,12 +570,12 @@ TIFFImageIO::InternalWrite(const void * buffer)
     pages = m_Dimensions[2];
   }
 
-  int  scomponents = this->GetNumberOfComponents();
-  auto resolution_x = static_cast<float>(m_Spacing[0] != 0.0 ? 25.4 / m_Spacing[0] : 0.0);
-  auto resolution_y = static_cast<float>(m_Spacing[1] != 0.0 ? 25.4 / m_Spacing[1] : 0.0);
+  unsigned int scomponents = this->GetNumberOfComponents();
+  auto         resolution_x = static_cast<double>(m_Spacing[0] != 0.0 ? 25.4 / m_Spacing[0] : 0.0);
+  auto         resolution_y = static_cast<double>(m_Spacing[1] != 0.0 ? 25.4 / m_Spacing[1] : 0.0);
```
N-Dekker: Please reconsider the original `static_cast<float>` for both resolution_x and resolution_y. They must be passed to `TIFFSetField` as 32-bit float TIFF tag values. Right?

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
```diff
@@ -559,9 +559,9 @@ TIFFImageIO::Write(const void * buffer)
 void
 TIFFImageIO::InternalWrite(const void * buffer)
 {
-  const auto * outPtr = (const char *)buffer;
+  const auto * outPtr = static_cast<const char *>(buffer);
 
-  unsigned int page, pages = 1;
+  size_t page, pages = 1;
```
N-Dekker: Why do you prefer `size_t` to `unsigned int`, for page and pages? I'm not sure but it looks like they should actually be 16-bit unsigned, for `TIFFTAG_PAGENUMBER`! http://www.libtiff.org/man/TIFFSetField.3t.html

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
YannLePoul: m_TotalColors is not unsigned because it can be -1 if image on disk is not palette image. That's why you will see these `m_TotalColors > 0` or `m_TotalColors >-1` everywhere. It is not so clear of a name, but I was thinking doing a different PR for that.

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
YannLePoul: The goal of static cast here is to be explicit on the unsigned -> signed casting, because  m_TotalColors is signed. I don't know why bit shift is done on a long. Code was introduced in 40afd95161727be3274f8094ce19fe45e2c3aa50 . If I had to guess, I would say for safety reason with bit shifts overhead ?

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
YannLePoul: yep

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
YannLePoul: yep

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
YannLePoul: Did that because of the type of `m_Dimensions`, but I'll do the other way around.

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
YannLePoul: In ReadVolume function, they are just used by  ReadCurrentPage function, which expect a size_t.

YannLePoul:
> I'm always glad if unnecessary type conversions are removed, thanks @YannLePoul However, it appears that the TIFF specification (as well as its implementation) expects certain very specific scalar types. I'm not sure if everything goes well when you change them! So please double-check!

Thanks for helping pointing out the problems. I see this commit as a draft anyway (I guess there are many implicit cast lying around)

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
N-Dekker: Thanks for the clarification! So `TIFFSetField` supports 64-bit `double` as argument (32-bit `float` arguments get promoted to `double`: https://stackoverflow.com/questions/11270588/variadic-function-va-arg-doesnt-work-with-float), but then its value is internally casted to 32-bit `float` anyway 😞 as you pointed out.

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
N-Dekker: As far as I can see, the expression `(1 << m_InternalImage->m_BitsPerSample)` is already signed. So the `static_cast<int>` does not seem necessary.  Right?

By the way, C++11 braced conversion `int{x}` is more type-safe than `static_cast<int>(x)`. Would you like that better?

    m_TotalColors = int{ 1 << m_InternalImage->m_BitsPerSample };

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
N-Dekker: But still I would like to remark that the expression `m_Spacing[0] != 0.0 ? 25.4 / m_Spacing[0] : 0.0` already is of type `double`. So  `static_cast<double>` seems unnecessary. If you really want to make the type explicit, why not just replace `auto` by `double`? You might consider braced initialization, to improve type safety, something like this:

       double resolution_x { m_Spacing[0] != 0.0 ? 25.4 / m_Spacing[0] : 0.0 };

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
N-Dekker: Oops, I see now, the `TIFFReadRGBAImageOriented(m_InternalImage->m_Image, width, height...)` call is in another member function.  Sorry for the noise!

N-Dekker: To ensure that no narrowing (signed-to-unsigned or otherwise) conversion takes place, you could use C++11 curly-braced list-initialization:

    const size_t width{ m_InternalImage->m_Width };
    const size_t height{ m_InternalImage->m_Height };

I think that's preferable, especially because the size of `size_t` is platform dependent. What do you think?

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
YannLePoul: In theory, int must be 16bit at least, so you could have a 16bit image, and it would be overflow. It shouldn't happen on 3ébit platform, and anyway m_TotalColors is int. Or We could set m_TotalColors to a long. What do you think ?

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
YannLePoul: I'll do that

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
N-Dekker: On Linux, `long` is typically 64-bit, but on Windows, `int` and `long` are both 32-bits (nowadays).s. When `m_TotalColors` is a signed 32-bits integer, BitsPerSample > 30 will cause overflow. Is that OK? I mean, is BitsPerSample in practice always below 31? Otherwise maybe `m_TotalColors` should be defined as `int64_t`. Or `uint64_t`.

YannLePoul:
I've double checked all `TIFFSetField` use in `itkTIFFImageIO.cxx` to make sure the right variable type is given

> However, it appears that the TIFF specification (as well as its implementation) expects certain very specific scalar types. I'm not sure if everything goes well when you change them! So please double-check!

I've just double checked all `TIFFSetField` in `itkTIFFImageIO.cxx` to make sure the right variable type is given. Few corrections again, but that should be it now

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
YannLePoul: I've defined it as `int64_t`, to be on the really safe side (`uint64_t` not possible because of the `-1` in case no palette)

YannLePoul:
@N-Dekker 
> It still seems to me that `m_TotalColors` = 0 would be just fine, for the "no palette case". Don't you think?

Ok after double checking, There is no way to have m_TotalColor to be 0 in case of a palette image, so 0 is a safe default value.

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
```diff
@@ -1405,11 +1401,11 @@ TIFFImageIO::ReadGenericImage(void * _out, unsigned int width, unsigned int heig
 
     if (m_InternalImage->m_Orientation == ORIENTATION_TOPLEFT)
     {
-      image = out + (size_t)(row)*width * inc;
+      image = out + row * width * inc;
```
N-Dekker: VS2017 Code Analysis says:
> itktiffimageio.cxx(1404): warning C26451: Arithmetic overflow: Using operator '*' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '*' to avoid overflow (io.2).

The problem introduced here is that `width * inc` is calculated in 32-bit, while on a 64-bit build, the multiplication was  previously calculated in 64-bit: `(size_t)(row)*width`.

A minimal fix could be to add a pair of parentheses:

    image = out + row * (width * inc);

`inc` is 64-bit size_t (on a 64-bit build), so then `(width * inc)` is calculated in 64-bit, and then `row * (width * inc)` is also calculated in 64-bit! What do you think? Anyway, I think this should be fixed one way or another before merging!

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
```diff
@@ -119,9 +119,9 @@ TIFFImageIO::GetFormat()
       {
         if (this->GetExpandRGBPalette())
         {
-          for (unsigned int cc = 0; cc < static_cast<unsigned int>(m_TotalColors); ++cc)
+          for (uint64_t cc = 0; cc < static_cast<uint64_t>(m_TotalColors); ++cc)
```
N-Dekker: This `static_cast` is no longer necessary now 😄  Please remove!

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
```diff
@@ -559,24 +555,24 @@ TIFFImageIO::Write(const void * buffer)
...
   if (m_NumberOfDimensions == 3)
   {
-    pages = m_Dimensions[2];
+    pages = static_cast<uint16>(m_Dimensions[2]);
   }
 
-  int  scomponents = this->GetNumberOfComponents();
-  auto resolution_x = static_cast<float>(m_Spacing[0] != 0.0 ? 25.4 / m_Spacing[0] : 0.0);
-  auto resolution_y = static_cast<float>(m_Spacing[1] != 0.0 ? 25.4 / m_Spacing[1] : 0.0);
+  uint16 scomponents = static_cast<uint16>(this->GetNumberOfComponents());
```
N-Dekker: (Minor) You could use `auto`, as the type is clearly redundant here. (Suggested by Visual Assist Code Inspection).

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
YannLePoul: > would it be worth mentioning in the commit message that your commit improves BitsPerSample support, as it now supports up to 63 bits per sample, instead of < 31 bits...?

I would say `uint64` type allows full (safe) support **palette** for 16-bit images (overflow if `int` is 16-bit only), and theoretical support for up to 64-bit palette (for what `m_TotalColors` is concerned), but for tiff above 16bit, an exception is raised elsewhere anyway. There may be other things to change/verify in the code so that it actually works, so I would not particularly insist on that...

> and then `row * (width * inc)`

or just move `inc` on the left !

N-Dekker: > overflow if int is 16-bit only

Does ITK support any platform that has 16-bit int? @dzenanz ?

N-Dekker:
No show stopper to me, but you might still fix the following line in `TIFFImageIO::InternalWrite`, if you want to:

    int rowLength; // in bytes

I think it might be better to declare it `size_t`, or `SizeValueType`. Especially because of:

    rowLength *= width;

https://github.com/InsightSoftwareConsortium/ITK/blob/81e1ada557bbf635bc673cf6b94a5bbbca1b1223/Modules/IO/TIFF/src/itkTIFFImageIO.cxx#L817

Which triggers a (usually suppressed) warning, from MSVC:
> warning C4244: '*=': conversion from 'const itk::ImageIOBase::SizeValueType' to 'int', possible loss of data

But you could also leave it to another commit, or someone else  😄

PR #1240, opened 2019-09-09
WIP: Convert GDCM image to RGB color

blowekamp:
When an ImageIO reports it's buffer as RGB it is expected to produce
and RGB image, not a color image in anther encoding format.

Co-Author: Mihail Isakov <mihail.isakov@gmail.com>

## PR Checklist

### DIFF
--- a/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
+++ b/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
@@ -55,6 +55,7 @@
 
 #include <fstream>
 #include <sstream>
+#include <gdcmImageChangePhotometricInterpretation.h>
 
 namespace itk
 {
@@ -301,6 +302,21 @@ GDCMImageIO::Read(void * pointer)
     image = ialut.GetOutput();
     len *= 3;
   }
+  else if (pi == gdcm::PhotometricInterpretation::YBR_FULL || pi == gdcm::PhotometricInterpretation::YBR_FULL_422 ||
+           pi == gdcm::PhotometricInterpretation::YBR_PARTIAL_420 ||
+           pi == gdcm::PhotometricInterpretation::YBR_PARTIAL_422)
+  {
+    gdcm::ImageChangePhotometricInterpretation icpi;
+    icpi.SetInput(image);
+    icpi.SetPhotometricInterpretation(gdcm::PhotometricInterpretation::RGB);
+    if (!icpi.Change())
+    {
+      itkExceptionMacro("Failure to change photometric interpretation to RGB from "
+                        << gdcm::PhotometricInterpretation::GetPIString(pi) << "!");
+    }
+    image = icpi.GetOutput();
+  }
+
 
   if (!image.GetBuffer((char *)pointer))
   {


Modules/IO/GDCM/src/itkGDCMImageIO.cxx
```diff
@@ -301,6 +302,21 @@ GDCMImageIO::Read(void * pointer)
     image = ialut.GetOutput();
     len *= 3;
   }
+  else if (pi == gdcm::PhotometricInterpretation::YBR_FULL || pi == gdcm::PhotometricInterpretation::YBR_FULL_422 ||
+           pi == gdcm::PhotometricInterpretation::YBR_PARTIAL_420 ||
+           pi == gdcm::PhotometricInterpretation::YBR_PARTIAL_422)
```
blowekamp: I don't believe PARTIAL is handled correctly but gdcm::ImageChangePhotometricInterpretation. There are also many other PhotometricInterpretation that need to be looked into, at minimal if the image provided by GDCM is not RGB a warning should be printed, that the color space/Photometric Interpretation is not supported.

blowekamp:
@malaterre Any insight if any other [photometric interpretations](http://gdcm.sourceforge.net/html/classgdcm_1_1PhotometricInterpretation.html) are being converted to RGB already?

issakomi:
In the latest version of the branch from that old PR (deleted) was also an option to preserve the old behavior. And i've added a block to avoid double memory allocation by _gdcm::ImageChangePhotometricInterpretation icpi_, which also crashed (i didn't investigate), the math is the same. It is just FYI, of course, your decision. Thank you.

```
+  else if (/*!m_PreserveYBR &&*/
+           (pi == gdcm::PhotometricInterpretation::YBR_FULL_422 || pi == gdcm::PhotometricInterpretation::YBR_FULL) &&
+           (pixeltype == gdcm::PixelFormat::UINT8 || pixeltype == gdcm::PixelFormat::INT8) && len % 3 == 0 &&
+           m_NumberOfComponents == 3)
+  {
+    unsigned char * copy = reinterpret_cast<unsigned char *>(pointer);
+    for (unsigned long long x = 0; x < len; x += 3)
+    {
+      const unsigned char a = copy[x];
+      const unsigned char b = copy[x + 1];
+      const unsigned char c = copy[x + 2];
+      int                 R = 38142 * (a - 16) + 52298 * (c - 128);
+      int                 G = 38142 * (a - 16) - 26640 * (c - 128) - 12845 * (b - 128);
+      int                 B = 38142 * (a - 16) + 66093 * (b - 128);
+      R = (R + 16384) >> 15;
+      G = (G + 16384) >> 15;
+      B = (B + 16384) >> 15;
+      if (R < 0)
+        R = 0;
+      if (R > 255)
+        R = 255;
+      if (G < 0)
+        G = 0;
+      if (G > 255)
+        G = 255;
+      if (B < 0)
+        B = 0;
+      if (B > 255)
+        B = 255;
+      copy[x] = (unsigned char)R;
+      copy[x + 1] = (unsigned char)G;
+      copy[x + 2] = (unsigned char)B;
+    }
+  }
```

malaterre:
> @malaterre Any insight if any other [photometric interpretations](http://gdcm.sourceforge.net/html/classgdcm_1_1PhotometricInterpretation.html) are being converted to RGB already?

<deep breath>Originally when GDCM was designed it would closely follow ITK Image representation. At that time ITK did not support colorspace (did that change?). In which case when an ITK pipeline is fed with an input JPEG/YBR_FULL_422, the expected behavior is to load an RGB image (I cannot think of another sane behavior). At this point, ITK sees an RGB buffer, execute the pipeline and if needed writes the output. By design I was expecting the output buffer to be used in another quantitative pipeline, which means I need to use a lossless representation in DICOM (yes: YBR_FULL_422 is a lossy colorspace) . So IMHO an output with PhotometricInterpretation of RGB is what is expected or at least that's what I designed and implemented.

Think of it this way, what is the expected behavior of ITK in these cases:

1. JPEG -> filter -> TIFF
2. TIFF/JPEG -> filter -> TIFF

Do you expect TIFF output to have lossy compressed buffer ?

Having said that, your point may be that : "Yes, but I did set YBR_FULL_422 in the ITK MetaData dictionary, so I am an advanced user and I did read the DICOM standard".

My answer is: "Well then please re-read it carefully", in particular part 5. For example DICOM/JPEG-2000 will not accept YBR_FULL_422 as Photometric Interpretation:

* http://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_8.2.4.html#table_8.2.4-1

At this point if you are thinking: "Yes, thank you, but I knew all of this by heart, stop being pedantic, I simply want a particular combination".

Then my answer is: the only case where you would have YBR_FULL_422 as input is with (1) JPEG Lossy 8bits or (2) plain raw buffer. In the first case PhotometricInterpretation=YBR_FULL_422 is simply a low level detail of the internal JPEG stream. I guess some advanced user may expect to preserve their compression ratio of the input image. In the second case people are loading raw/YBR_FULL_422 image, converting the buffer to RGB, filtering the buffer and then writing it out as raw/YBR_FULL_422. There would need some clarification about this pipeline since I would have assumed people would want to process directly the raw/YBR_FULL_422 buffer.

@zivy which cases described above do you want (1) or (2) ?

malaterre:
@blowekamp Nevermind, looks like I answered another thread here. The patch looks almost correct. By default GDCM will convert to RGB an input JPEG/YBR_FULL_422. So the only remaining bug is what the input is:

1. RLE/YBR_FULL
2. raw/YBR_FULL_422
3. raw/YBR_FULL (who would want to do that?)

@issakomi  You should update your `if()`

zivy:
@malaterre Thanks for the comprehensive explanation. To tell the truth I had no preference, this is more of a convention issue. I was only trying to make the minimal changes to original DICOM images that were originally in YBR_FULL_422. In the end I just saved in RGB and accepted that that is an ITK convention.

blowekamp:
The writing is a separate issue, not here.

For the reading, if the loaded image buffer is report as `ImageIOBase::IOPixelType::RGB`, then it needs to be RGB or else it's a bug. Like for this case.

> @blowekamp Nevermind, looks like I answered another thread here. The patch looks almost correct. By default GDCM will convert to RGB an input JPEG/YBR_FULL_422. So the only remaining bug is when the input is:
> 
> 1. RLE/YBR_FULL
> 2. raw/YBR_FULL_422
> 3. raw/YBR_FULL (who would want to do that?)
> 
> GDCM does not handle the YBR_PARTIAL_422 stuff. I suspect JPEG/YBR_PARTIAL_420 is handled by the internal JPEG library, but I never received one such sample.
> 

Do you have test images for 1,2,3?

issakomi:
[GE RLE YBR_FULL](https://drive.google.com/file/d/1vouULQY6_sNmkuxpqklRLT3i45RF0Bsl/view?usp=sharing)

[Philips Jpeg YBR_FULL_422](https://drive.google.com/file/d/0ByBLJ2-U1t9xWHczTkhDMVZST3M/view?usp=sharing)

issakomi:
To validate one can save raw data from lossy Jpeg DICOM as _jpg_, highly likely you will be able to open that file with any image viewer (for multi-frames only 1st frame will be visible in usual viewer). Surely also possible with GDCM tools, don't remember syntax.

```
// input must be Dicom file in lossy Jpeg transfer syntax 

#include <gdcmReader.h>
#include <gdcmTag.h>
#include <gdcmDataSet.h>
#include <gdcmDataElement.h>
#include <gdcmSequenceOfFragments.h>
#include <fstream>
#include <iostream>

int main(int argc, char ** argv)
{
  gdcm::Reader reader;
  reader.SetFileName(argv[1]);
  if (!reader.Read()) { std::cout << "failed (1)" << std::endl; return 1; }
  const gdcm::DataSet & ds = reader.GetFile().GetDataSet();
  const gdcm::DataElement & e = ds.GetDataElement(gdcm::Tag(0x7fe0,0x0010));
  const gdcm::SequenceOfFragments * sf = e.GetSequenceOfFragments();
  if(!sf) { std::cout << "failed (2)" << std::endl; return 1; }
  std::ofstream out("binary.jpg");
  sf->WriteBuffer(out);
  out.close();
  return 0;
}
```



```
CMAKE_MINIMUM_REQUIRED(VERSION 3.1.0)
PROJECT(test35)
FIND_PACKAGE(ITK REQUIRED)
INCLUDE(${ITK_USE_FILE})
ADD_EXECUTABLE(test35 main.cxx)
TARGET_LINK_LIBRARIES(test35 ${ITK_LIBRARIES})
```

blowekamp:
When I ran the code on a YBR_FULL_422 image. I got an exception on line 323.

@malaterre Any insight into why the resulting buffer is null?

issakomi:
`By default GDCM will convert to RGB an input JPEG/YBR_FULL_422. So the only remaining bug is when the input is`
May be _will_ is future tense, not sure, but actually lossy Jpeg/YBR_FULL_422 and RLE/YBR_FULL, both, are greenish blue. Plus this crash, BTW, i mentioned twice, in the 1st PR and here. There is probably a bug with lossy Jpeg/YBR_FULL_422, not a feature.

Modules/IO/GDCM/src/itkGDCMImageIO.cxx
issakomi: I didn't see PARTIALs for any related transfer syntax and don't have examples, so didn't add them in 1st PR. May be such files exist, who knows, and they will be handled by decoder correct, difficult to say without samples.

S. valid photometric/transfer systax combinations from current Dicom standard.

[RLE](http://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_8.2.2.html#table_8.2.2-1)
[Jpeg2000](http://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_8.2.4.html#table_8.2.4-1)
[JpegLS](http://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_8.2.3.html#table_8.2.3-1)
[Lossy Jpeg](http://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_8.2.html#table_8.2.1-1)
[Lossless Jpeg](http://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_8.2.html#table_8.2.1-2)

YBR_PARTIAL_420 is valid for MPEG files.

Also the above tables mention many extremely rare are or hypothetical combinations, eg Jpeg2000 YBR_FULL. In real life almost always there are lossy Jpeg/YBR_FULL_422 and RLE/YBR_FULL.

issakomi:
In GDCM [these 2 lines: 886, 887](https://github.com/InsightSoftwareConsortium/ITK/blob/2b82a99b888657fb5edcc7f02ae576860ac33b10/Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmJPEGBITSCodec.hxx#L861) make a difference. If commented out - image colors for lossy Jpeg are correct. It not a request to change it now, just FYI, if the PR is still alive.

blowekamp:
> In GDCM [these 2 lines: 861, 862](https://github.com/InsightSoftwareConsortium/ITK/blob/2b82a99b888657fb5edcc7f02ae576860ac33b10/Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmJPEGBITSCodec.hxx#L861) make a difference. If commented out - image colors for lossy Jpeg are correct. It not a request to change it now, just FYI, if the PR is still alive.

Are you saying that your proposed change solves the issue alone? or is it needed with the PR?

issakomi:
@blowekamp it would solve issue with YBR_FULL_422 without additional changes.
May be [here](https://github.com/InsightSoftwareConsortium/ITK/blob/2b82a99b888657fb5edcc7f02ae576860ac33b10/Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmJPEGBITSCodec.hxx#L858) were possible to change
```
if( GetPhotometricInterpretation() == PhotometricInterpretation::YBR_FULL
      || GetPhotometricInterpretation() == PhotometricInterpretation::YBR_FULL_422 )
```
to
`if( GetPhotometricInterpretation() == PhotometricInterpretation::YBR_FULL)`

for hypothetical case of YBR_FULL Jpeg files (never seen).
And in PR also let color space change only for YBR_FULL (is required at least for RLE).
It were in sync with @malaterre statement

> By default GDCM will convert to RGB an input JPEG/YBR_FULL_422. So the only remaining bug is when the input is:

The remaining case were _raw/YBR_FULL_422_, but i don't think such files exists.
Just FYI. Not sure what is better.

blowekamp:
Outstanding! Can you please make the PR to [GDCM](https://github.com/malaterre/GDCM), then we can update it in ITK in time for the 5.1 release!

issakomi:
I'll try, may be @malaterre will accept. But it doesn't change RLE, they are rather common (RLE/YBR_FULL), it will only fix YBR_FULL_422.

issakomi:
I've closed PR, i think @malaterre read about and don't want, i can understand, may be some logic in tools/exes with many option exists, i can not test all of them and don't know what they, just guessing. 

**PS** Probably i shall make other PR, style only (warning) in some codecs to be in sync with actual [DICOM tables for valid values](https://github.com/InsightSoftwareConsortium/ITK/pull/1240#discussion_r324430552).

PR #1250, opened 2019-09-13
STYLE: Remove reinterpret_cast's from void to non-void pointer

N-Dekker:
Replaced `reinterpret_cast` by `static_cast`, when converting a void to
a non-void pointer type. `static_cast` is in general less unsafe than
`reinterpret_cast`.

Fixes Visual C++ 2017 Code Analysis warning C26471: Don't use
reinterpret_cast. A cast from void* can use static_cast (type.1).

### DIFF
--- a/Modules/Core/Common/src/itkObjectFactoryBase.cxx
+++ b/Modules/Core/Common/src/itkObjectFactoryBase.cxx
@@ -876,7 +876,7 @@ ObjectFactoryBase ::SynchronizeObjectFactoryBase(void * objectFactoryBasePrivate
   ObjectFactoryBasePrivate * previousObjectFactoryBasePrivate;
   previousObjectFactoryBasePrivate = GetPimplGlobalsPointer();
 
-  m_PimplGlobals = reinterpret_cast<ObjectFactoryBasePrivate *>(objectFactoryBasePrivate);
+  m_PimplGlobals = static_cast<ObjectFactoryBasePrivate *>(objectFactoryBasePrivate);
   if (m_PimplGlobals && previousObjectFactoryBasePrivate)
   {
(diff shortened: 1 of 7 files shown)


Modules/Core/Common/src/itkObjectFactoryBase.cxx
```diff
@@ -876,7 +876,7 @@ ObjectFactoryBase ::SynchronizeObjectFactoryBase(void * objectFactoryBasePrivate
   ObjectFactoryBasePrivate * previousObjectFactoryBasePrivate;
   previousObjectFactoryBasePrivate = GetPimplGlobalsPointer();
 
-  m_PimplGlobals = reinterpret_cast<ObjectFactoryBasePrivate *>(objectFactoryBasePrivate);
+  m_PimplGlobals = static_cast<ObjectFactoryBasePrivate *>(objectFactoryBasePrivate);
```
hjmjohnson: Should this one be a dynamic_cast?

N-Dekker: Thanks for your question, HansI just double-checked: `objectFactoryBasePrivate` is really a void pointer, so it cannot by casted by `dynamic_cast`.

N-Dekker:
@YannLePoul FYI I noticed those warnings about `reinterpret_cast` for the first time when looking at your pull request #1236  !

This pull request also fixed these warnings in itkTIFFImageIO.cxx

YannLePoul:
Yes I saw those, but I wasn't feeling confident enough with casting to change it...! Glad you did!

N-Dekker:
Thanks for your comments and approvals! Can one of you please merge my PR? Technically I think I can merge my own pull requests now (which is great!), but it still feels better if you do it 😃

PR #1251, opened 2019-09-13
BUG: Support Numpy inputs in keywords that include "image"

mrocklin:
Previously there was a whitelist on which keywords would be
automatically converted from Numpy arrays to ITK Image objects.

Now, we accept any keyword that includes the text "image" in the name.


mrocklin:
Thanks for taking this over the finish line @thewtex

PR #1253, opened 2019-09-16
 BUG: Fixes bugs in SpatialObject to MetaIO conversion

aylward:
In conjunction with updates to MetaIO (https://github.com/Kitware/MetaIO/pull/81),
fixes the object factory mechanism of itkMetaSceneConverter and Read/Write Spatial Objects.

In particular for Reading Spatial Objects, you may now call
GetOutput() to get the exact objects read from a file or GetGroup()
which appends a group as the root of the tree of objects read from
a file.   GetGroup() was a pre-existing command, but as the only
command to read from a file, the WriteSpatialObjects class would (sometimes)
strip the top-level group from a tree of spatialObjects before
writting them to a file - this is bad, e.g., if the top level group
contained a transform that needs to be applied to the children.

Additionally, type conversion for MetaToSpatialObjects would fail
if the Meta objects were directly created instead of read from
a file.  Since directly creating a MetaIO object would not correctly
set its ObjectTypeName used by the factory.

Other minor changes were in regard to line length.  Long lines were
shortened to make the code more readable on 80 char windows.

### DIFF
--- a/Modules/Core/Common/include/itkByteSwapper.hxx
+++ b/Modules/Core/Common/include/itkByteSwapper.hxx
@@ -275,7 +275,7 @@ template <typename T>
 void
 ByteSwapper<T>::Swap2(void * pin)
 {
-  auto *               p = reinterpret_cast<unsigned short *>(pin);
+  auto *               p = static_cast<unsigned short *>(pin);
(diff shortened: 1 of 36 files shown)


aylward:
The one failing test "itkNewMetaObjectTypeTest" will pass once MetaIO is updated by #1257 .

PR #1256, opened 2019-09-17
STYLE: Use C++11 nullptr in itkDynamicCastInDebugMode

N-Dekker:
Fixes Visual Studio 2017 Code Analysis warnings from itkMacro.h:

Note that the formatting changes in the macro definition of
`itkGPUKernelClassMacro(kernel)` are from a post-commit hook.
These are unrelated to the fixes of those VS2017 warnings.

### DIFF
--- a/Modules/Core/Common/include/itkMacro.h
+++ b/Modules/Core/Common/include/itkMacro.h
@@ -1100,9 +1100,9 @@ compilers.
  * provides the GPU kernel source code as a const char*
  */
 #define itkGPUKernelClassMacro(kernel)                                                                                 \
-  /**\class kernel                                                                                                     \
-   * Workaround KWstyle bug                                                                                            \
-   * \ingroup ITKCommon                                                                                                \
+  /**\class kernel \                                                                                                                     \
+   * Workaround KWstyle bug \                                                                                                                     \
+   * \ingroup ITKCommon \                                                                                                                     \
    */                                                                                                                  \
   class kernel                                                                                                         \
   {                                                                                                                    \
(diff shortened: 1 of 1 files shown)


N-Dekker:
Hans, @hjmjohnson Can you please check if the reformatting of `itkGPUKernelClassMacro(kernel)` is OK? I did not intend to touch that macro, but apparently the post-commit hook did so, automatically.

I just wanted to fix the `itkDynamicCastInDebugMode` macro!

Modules/Core/Common/include/itkMacro.h
```diff
@@ -1100,9 +1100,9 @@ compilers.
  * provides the GPU kernel source code as a const char*
  */
 #define itkGPUKernelClassMacro(kernel)                                                                                 \
-  /**\class kernel                                                                                                     \
-   * Workaround KWstyle bug                                                                                            \
-   * \ingroup ITKCommon                                                                                                \
+  /**\class kernel \                                                                                                                     \
```
blowekamp: I don't think Doxygen will use that at all because it is configured to only used predefined macros: https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/Doxygen/doxygen.config.in#L2132

N-Dekker: Thanks @blowekamp ! It says "Workaround KWstyle bug", so is the KWstyle bug still relevant?

blowekamp: I don't know. The usual way to work around a KWStyle issue is to add an override: https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/ITKKWStyleOverwrite.txt

N-Dekker: Thanks for the information, Bradley. For now I would like to leave it this way. This pull request is just meant to fix the `itkDynamicCastInDebugMode` macro, and I think it's ready now! 😃

N-Dekker:
Thanks for your approval and merge, @dzenanz 😄

PR #1258, opened 2019-09-17
STYLE: Use nullptr for isSpecialCoordinatesImage in ResampleImageFilter

N-Dekker:
Explicitly compares results of `dynamic_cast` to `nullptr`, when
estimating `isSpecialCoordinatesImage`, in `ResampleImageFilter`.

Fixes a Visual Studio 2017 Code Analysis warning:

### DIFF
--- a/Modules/Filtering/ImageGrid/include/itkResampleImageFilter.hxx
+++ b/Modules/Filtering/ImageGrid/include/itkResampleImageFilter.hxx
@@ -177,8 +177,9 @@ ResampleImageFilter<TInputImage, TOutputImage, TInterpolatorPrecisionType, TTran
     return;
   }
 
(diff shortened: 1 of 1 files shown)


N-Dekker:
Thanks for your approval and merge, @dzenanz 

For the record, although it seems reasonable to me that the original code triggered a compiler warning, the one from VS2017 Code Analysis doesn't seem entirely appropriate:

> warning C26477: Use 'nullptr' rather than 0 or NULL

It seemed to suggest that the original code was using `0` or `NULL`, which wasn't the case! The same warning (misplaced, in my opinion) occurs for this code:

    struct A
    {
        virtual ~A() = 0;
    };

    struct B: A {};

    int main()
    {
      A* p = nullptr;
      bool b = dynamic_cast<B*>(p); // warning C26477: Use 'nullptr' rather than 0 or NULL
    }

As I observed at https://godbolt.org/z/cjR8_M

PR #1260, opened 2019-09-18
STYLE: Remove redundant code from operator<< for enum's, using lambda's

N-Dekker:
This commit removed a significant amount of redundant lines of code from
the implementation of `operator<<(out, value)` for `enum` value types,
using C++11 lambda's. It removed local `s` variables, that appeared
unnecessary.

Thereby, it fixes many Visual Studio 2017 Code Analysis warnings,
regarding `const char * s = 0`, which said:

A minimal fix of this warning, `const char * s = nullptr`, would still
be risky, as `out << s` might crash when `s` is a `nullptr`.


hjmjohnson:
@mseng10 Please review.

PR #1262, opened 2019-09-20
COMP: remove Eigen3 shadow warnings with GCC 4.8.5

### LINKED ISSUE #1254: Eigen warnings with GCC 4.8.5
I have a bunch of shadow warnings when compiling with GCC 4.8.5, see
[RTK dashoard](https://my.cdash.org/buildSummary.php?buildid=1702853). I am willing to fix them but I'm not sure what's the best procedure. Since they are coming from a third party software, Eigen3, should I add a rule in CMake/CTestCustom.cmake.in to ignore them or try to correct the Eigen3 code?

SimonRit:
Fixes #1254

## PR Checklist

### DIFF
--- a/Modules/ThirdParty/Eigen3/src/itkeigen/Eigen/src/Geometry/AlignedBox.h
+++ b/Modules/ThirdParty/Eigen3/src/itkeigen/Eigen/src/Geometry/AlignedBox.h
@@ -143,7 +143,7 @@ EIGEN_MAKE_ALIGNED_OPERATOR_NEW_IF_VECTORIZABLE_FIXED_SIZE(_Scalar,_AmbientDim)
     * For 3D bounding boxes, the following names are added:
     * BottomLeftCeil, BottomRightCeil, TopLeftCeil, TopRightCeil.
     */
-  EIGEN_DEVICE_FUNC inline VectorType corner(CornerType corner) const
+  EIGEN_DEVICE_FUNC inline VectorType corner(CornerType c) const
   {
     EIGEN_STATIC_ASSERT(_AmbientDim <= 3, THIS_METHOD_IS_ONLY_FOR_VECTORS_OF_A_SPECIFIC_SIZE);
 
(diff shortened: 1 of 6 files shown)


phcerdan:
This is great @SimonRit, thanks. 
We have an Eigen fork/branch for itk in [here](https://gitlab.kitware.com/phcerdan/eigen/tree/for/itk). This patch should be integrated there first, and then call the UpdateFromUpstream script to merge into ITK, instead of directly merging here.

Eigen is a third party library, I don't think it scales well to diverge from upstream to omit warnings on a 5 years old compiler, however:
a) we can propose these changes into upstream, [here](https://bitbucket.org/eigen/eigen/src/default/). 
b) we can ignore this, and the user add the compiler flag `-Wno-shadow` to discard the spurious warnings for this particular compiler.

I will let ITK main maintainers to decide about it. I can integrate this patch in the for/itk branch of our Eigen fork, however I don't feel comfortable proposing this into upstream Eigen, I have the feeling that we would only solve it for a subset of Eigen, and only for old compilers. But of course, if @SimonRit is keen, you just have to flick a mail to Eigen mantainer, and shortly after you would be able to open merge requests there.

SimonRit:
Sure, I understand your point. There is another option
c) Add a line in CMake/CTestCustom.cmake.in to ignore these warnings during ctest.

phcerdan:
> c) Add a line in CMake/CTestCustom.cmake.in to ignore these warnings during ctest.

I would prefer that one too if it is only added when using gcc4.8 or older.

thewtex:
Thanks for pushing this forward @SimonRit @phcerdan :clap: . I see a number of Eigen warnings locally beyond GCC 8 (not sure if it is these, though), and it would be excellent to clean up the noise :droplet: .

This patch seems like a very reasonable fix -- I think it would be worthwhile submitting a PR to upstream to see if they would accept it.

dzenanz:
Does a recent update of Eigen interact with this PR?

hjmjohnson:
No.  I was hopeful that the recent update would fix the problem, but alas it did not.

I had this problem before and after the recent update.


I can replicate on both mac and linux with pretty new CPU's, so I'm thinking it's a problem with Eigen conditional compilation code.

hjmjohnson:
@phcerdan Is this PR still relevant after the recent Eigen work?

phcerdan:
I don't have that GCC to reproduce it. The warnings are not there with a modern compiler.

N-Dekker:
Slightly related: A fix that I proposed for another warning from an Eigen header file was merged into the Eigen master branch two weeks ago, by Rasmus Munk Larsen (@rmlarsen):

_Fix half_impl::float_to_half_rtne(float) warning: '<<' causes overflow_
https://gitlab.com/libeigen/eigen/-/merge_requests/202
https://gitlab.com/libeigen/eigen/-/commit/5328c9be4356cc3d7ab2e359acb1e9ebeeeea631

The warning did appear when building an ITK based application.

PR #1263, opened 2019-09-21
STYLE: Replace 0 by nullptr when appropriate

N-Dekker:
Fixes Visual Studio 2017 Code Analysis warnings:


blowekamp:
😎 Nice work.

N-Dekker:
Thanks for your encouragements, Hans @hjmjohnson and Bradley @blowekamp  I just  did a force-push, as I found two more `nullptr` opportunities  😄

N-Dekker:
Thanks for merging, @dzenanz As far as I can see, all those "Use 'nullptr` warnings" from from ITK source files are now fixed. Third party files do still trigger many of those warning, though, including the versions of eigen3 and googletest used by ITK. I hope other people will take care of those, some day...  😄

Issue #1162, opened 2019-08-15
meshread / meshwrite function


thewtex:
Add functions similar to `itk.imread`, `itk.imwrite` for meshes. Perhaps `itk.meshread`, `itk.meshwrite`.

### Description

This functions will enable reading meshes from disk without specification of their pixel type and dimension.

### Expected behavior

```
mesh = itk.meshread(filename)
itk.meshwrite(mesh, filename)
```

### Actual behavior

Currently, `itk.MeshFileReader` with an explicit pixel type and dimension, which may not be the same as the file on disk, is required.

PR #1274, opened 2019-09-25
itk.meshread, itk.meshwrite

thewtex:
Addresses #1162

### DIFF
--- a/Wrapping/Generators/Python/itkTemplate.py
+++ b/Wrapping/Generators/Python/itkTemplate.py
@@ -61,9 +61,10 @@ class TemplateTypeError(TypeError):
                 itk.python_type(t)
 
         import itk
-        # Special case for ITK image readers: Add extra information.
+        # Special case for ITK readers: Add extra information.
         extra_eg = ""
-        if template_type in [itk.ImageFileReader, itk.ImageSeriesReader]:
+        if template_type in [itk.ImageFileReader, itk.ImageSeriesReader,
+                itk.MeshFileReader]:
             extra_eg="""
 
 or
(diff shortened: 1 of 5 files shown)


Wrapping/Generators/Python/itkTemplate.py
```diff
@@ -543,14 +543,82 @@ def firstIfList(arg):
...
+            if not os.path.isfile(inputFileName):
+                msg += ("\nThe file doesn't exist. \n" +
+                       "Filename = %s" % inputFileName)
+            raise RuntimeError("Could not create IO object for reading file %s" % inputFileName + msg)
+        # Read the metadata from the mesh file.
+        meshIO.SetFileName(inputFileName)
+        meshIO.ReadMeshInformation()
+        dimension = meshIO.GetPointDimension()
+        componentAsString = meshIO.GetComponentTypeAsString(meshIO.GetPointPixelComponentType())
+        _io_component_type_dict = {
```
dzenanz: This is a duplicated definition, and therefore prone to discrepancies. Can this be shared with `_NewImageReader`'s definition?

maekclena:
The failing test should be removed since `itk.MeshFileReader.New` now requires a `FileName` argument. See https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Mesh/wrapping/test/CMakeLists.txt#L13.

Wrapping/Generators/Python/itkTemplate.py
thewtex: There is a subtle difference between the definitions -- to keep it cleaner they were defined this way. I added a comment to explain that it is intentional and what was done.

thewtex:
> The failing test should be removed since `itk.MeshFileReader.New` now requires a `FileName` argument. See https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Mesh/wrapping/test/CMakeLists.txt#L13.

Good catch!

thewtex:
@dzenanz @maekclena thank you for the reviews!

PR #1282, opened 2019-10-01
N4minor optimization

### LINKED ISSUE #1281: Remove repeated intermediate sharpened image allocation in N4BiasCorrectionImageFilter
### Description

In every iteration of the N4BiasCorrectionImageFilter, a new pixel buffer is allocated for the intermediate sharpened image, which can be a costly operation

https://github.com/InsightSoftwareConsortium/ITK/blob/970e7fa9ce42f9d3da78c47213a320abc382d613/Modules/Filtering/BiasCorrection/include/itkN4BiasFieldCorrectionImageFilter.hxx#L456

### Impact analysis

Decreased performance.

### Expected behavior

The image is allocated once in `GenerateData()` and re-used. This could possibly be achieved with a new private member variable.

dzenanz:
Closes #1281.

### DIFF
--- a/Modules/Filtering/BiasCorrection/include/itkN4BiasFieldCorrectionImageFilter.h
+++ b/Modules/Filtering/BiasCorrection/include/itkN4BiasFieldCorrectionImageFilter.h
@@ -374,6 +374,12 @@ public:
    */
   itkGetConstMacro(CurrentLevel, unsigned int);
 
+  /**
+   * Reconstruct bias field given the control point lattice.
+   */
+  RealImagePointer
+  ReconstructBiasField(const BiasFieldControlPointLatticeType *);
+
 protected:
   N4BiasFieldCorrectionImageFilter();
   ~N4BiasFieldCorrectionImageFilter() override = default;
@@ -407,12 +413,6 @@ private:
    */
   RealImagePointer
   UpdateBiasFieldEstimate(RealImageType *, std::size_t);
-
-  /**
-   * Reconstruct bias field given the control point lattice.
-   */
-  RealImagePointer
-  ReconstructBiasField(BiasFieldControlPointLatticeType *);
 
   /**
    * Convergence is determined by the coefficient of variation of the difference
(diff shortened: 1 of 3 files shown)


Modules/Filtering/BiasCorrection/include/itkN4BiasFieldCorrectionImageFilter.h
```diff
@@ -374,6 +374,12 @@ class N4BiasFieldCorrectionImageFilter : public ImageToImageFilter<TInputImage,
    */
   itkGetConstMacro(CurrentLevel, unsigned int);
 
+  /**
+   * Reconstruct bias field given the control point lattice.
+   */
+  RealImagePointer
+  ReconstructBiasField(BiasFieldControlPointLatticeType *);
+
```
thewtex: Was this intentionally made `public`?

dzenanz: Yes. If this remained private, if a user wants to apply a bias field to a different image (e.g. a full resolution image), they first have to reconstruct the bias field. Which requires duplicating this method. It is better to just set a new input image to N4 filter, and call this method. Perhaps I should update the test to show how it reduces code duplication.

gdevenyi:
While someone is working on this, how hard would it be to give an initialization bias field as a guess?

dzenanz:
It would be easy. It is [initialized to zero](https://github.com/InsightSoftwareConsortium/ITK/blob/2074c2f9e5ed3f087d0e2059f8e0e8992fcad7ef/Modules/Filtering/BiasCorrection/include/itkN4BiasFieldCorrectionImageFilter.hxx#L160-L163). You just need an ivar for it, and expose it via setter (and maybe getter).

blowekamp:
> It would be easy. It is [initialized to zero](https://github.com/InsightSoftwareConsortium/ITK/blob/2074c2f9e5ed3f087d0e2059f8e0e8992fcad7ef/Modules/Filtering/BiasCorrection/include/itkN4BiasFieldCorrectionImageFilter.hxx#L160-L163). You just need an ivar for it, and expose it via setter (and maybe getter).

It may be better to consider it an optional pipeline input. So what Dzenan said plus a couple pipeline configuration lines in the constructor for an optional named input.

PR #1286, opened 2019-10-02
BUG: In GDCMIO convert to RGB, if required, option

issakomi:
@blowekamp  @malaterre  S. #1240 

- to skip `gdcmIO->SetYBRtoRGB(false);`
- colors improved
- s. comment

### DIFF
--- a/Modules/IO/GDCM/include/itkGDCMImageIO.h
+++ b/Modules/IO/GDCM/include/itkGDCMImageIO.h
@@ -164,6 +164,12 @@ public:
   itkGetConstMacro(LoadPrivateTags, bool);
   itkBooleanMacro(LoadPrivateTags);
 
+  /** Convert YBR to RGB
+   */
+  itkSetMacro(YBRtoRGB, bool);
+  itkGetConstMacro(YBRtoRGB, bool);
+  itkBooleanMacro(YBRtoRGB);
+
 #if defined(ITKIO_DEPRECATED_GDCM1_API)
   /** Convenience methods to query patient information and scanner
    * information. These methods are here for compatibility with the
@@ -360,6 +366,8 @@ protected:
   bool m_KeepOriginalUID;
 
   bool m_LoadPrivateTags;
+
+  bool m_YBRtoRGB;
 
 private:
 #if defined(ITKIO_DEPRECATED_GDCM1_API)
--- a/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
+++ b/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
@@ -91,6 +91,8 @@ GDCMImageIO::GDCMImageIO()
 
   m_LoadPrivateTags = false;
 
+  m_YBRtoRGB = true;
+
   m_InternalComponentType = UNKNOWNCOMPONENTTYPE;
 
   // by default assume that images will be 2D.
@@ -330,6 +332,33 @@ GDCMImageIO::Read(void * pointer)
     delete[] copy;
     // WARNING: sizeof(Real World Value) != sizeof(Stored Pixel)
     len = len * outputpt.GetPixelSize() / pixeltype.GetPixelSize();
+  }
+  /*
+   * YBR to RGB
+   * Currently (GDCM 3.0.2) processing for lossy Jpeg YBR_FULL_422
+   * is not fully clear. May be later the files will be converted
+   * internally in GDCM, in that case remove YBR_FULL_422 below.
+   *
+   * TODO: UINT16
+   *
+   */
+  if (m_NumberOfComponents == 3 &&
+      (pi == gdcm::PhotometricInterpretation::YBR_FULL || pi == gdcm::PhotometricInterpretation::YBR_FULL_422) &&
+      m_YBRtoRGB && (pixeltype == gdcm::PixelFormat::UINT8 || pixeltype == gdcm::PixelFormat::INT8) && (len % 3 == 0))
+  {
+    unsigned char * copy = reinterpret_cast<unsigned char *>(pointer);
+    for (size_t x = 0; x < len; x += 3)
+    {
+      const double Y = static_cast<double>(copy[x]);
+      const double Cb = static_cast<double>(copy[x + 1]);
+      const double Cr = static_cast<double>(copy[x + 2]);
+      const int    R = static_cast<int>((Y + 1.402 * (Cr - 128)) + 0.5);
+      const int    G = static_cast<int>((Y - (0.114 * 1.772 * (Cb - 128) + 0.299 * 1.402 * (Cr - 128)) / 0.587) + 0.5);
+      const int    B = static_cast<int>((Y + 1.772 * (Cb - 128)) + 0.5);
+      copy[x] = static_cast<unsigned char>((R < 0) ? 0 : R);
+      copy[x + 1] = static_cast<unsigned char>((G < 0) ? 0 : G);
+      copy[x + 2] = static_cast<unsigned char>((B < 0) ? 0 : B);
+    }
   }
 
 #ifndef NDEBUG


Modules/IO/GDCM/src/itkGDCMImageIO.cxx
```diff
@@ -331,6 +333,33 @@ GDCMImageIO::Read(void * pointer)
...
+   * YBR to RGB
+   * Currently (GDCM 3.0.2) processing for lossy Jpeg YBR_FULL_422
+   * is not fully clear. May be later the files will be converted
+   * internally in GDCM, in that case remove YBR_FULL_422 below.
+   *
+   * TODO: UINT16
+   *
+   */
+  if (m_NumberOfComponents == 3 &&
+      (pi == gdcm::PhotometricInterpretation::YBR_FULL || pi == gdcm::PhotometricInterpretation::YBR_FULL_422) &&
```
malaterre: You're doing twice YBR_FULL_422 -> RGB. I suspect this is wrong, please test with JPEG Lossy 8 bits this chunk.

Modules/IO/GDCM/src/itkGDCMImageIO.cxx
```diff
@@ -331,6 +333,33 @@ GDCMImageIO::Read(void * pointer)
...
+      m_YBRtoRGB && (pixeltype == gdcm::PixelFormat::UINT8 || pixeltype == gdcm::PixelFormat::INT8) && (len % 3 == 0))
+  {
+    unsigned char * copy = reinterpret_cast<unsigned char *>(pointer);
+    for (size_t x = 0; x < len; x += 3)
+    {
+      const double Y = static_cast<double>(copy[x]);
```
malaterre: I would have used the gdcm filter YBR to RGB instead of copy/pasting code.

malaterre:
@issakomi You have room for improvement. Thanks for preparing it.

Modules/IO/GDCM/src/itkGDCMImageIO.cxx
issakomi: Thank you. Sorry, change photometric interpretation from GDCM crashes with lossy Jpeg, s, previous closed PR and allocates memory. Sorry.

Modules/IO/GDCM/src/itkGDCMImageIO.cxx
malaterre: Could you add a comment here, explaining how JPEG Lossy/YBR_FULL_422 is not re-entering this loop ? Thanks much

issakomi:
@malaterre  Sorry, i will not change now, i believe there is a bug GDCM. Sorry. If it will be resolved i shall change of course, in next PR. Thank you.

Modules/IO/GDCM/src/itkGDCMImageIO.cxx
```diff
@@ -331,6 +333,33 @@ GDCMImageIO::Read(void * pointer)
...
+      const double Cb = static_cast<double>(copy[x + 1]);
+      const double Cr = static_cast<double>(copy[x + 2]);
+      const int    R = static_cast<int>((Y + 1.402 * (Cr - 128)) + 0.5);
+      const int    G = static_cast<int>((Y - (0.114 * 1.772 * (Cb - 128) + 0.299 * 1.402 * (Cr - 128)) / 0.587) + 0.5);
+      const int    B = static_cast<int>((Y + 1.772 * (Cb - 128)) + 0.5);
+      copy[x] = static_cast<unsigned char>((R < 0) ? 0 : R);
```
malaterre: In GDCM I am doing a full clamp here (<= 255):

* https://github.com/malaterre/GDCM/blob/master/Source/MediaStorageAndFileFormat/gdcmImageChangePhotometricInterpretation.h#L66

Modules/IO/GDCM/src/itkGDCMImageIO.cxx
issakomi: @malaterre  there is no loop, please kindly test

Modules/IO/GDCM/src/itkGDCMImageIO.cxx
issakomi: I think it is not required with proper cast, it is many years tested code.

issakomi: @malaterre  If you  [comment two lines here in GDCM,](https://github.com/InsightSoftwareConsortium/ITK/blob/3526b67168ded206a8b9a3e7100fb1849c476d8e/Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmJPEGBITSCodec.hxx#L858) convert wiil be not required for YBR_FULL_422, i have tried PR in GDCM and closed, please see yourself, may be it will have impact on some GDCM tools or whatever

issakomi: @malaterre 
```
      if( GetPhotometricInterpretation() == PhotometricInterpretation::YBR_FULL
      || GetPhotometricInterpretation() == PhotometricInterpretation::YBR_FULL_422 )
        {
// <------       cinfo.jpeg_color_space = JCS_UNKNOWN;
// <------        cinfo.out_color_space = JCS_UNKNOWN;
        //this->PlanarConfiguration = 1;
        }
```

issakomi:
BTW, i have tons of sample files, also s. here
[https://www.creatis.insa-lyon.fr/~jpr/PUBLIC/gdcm/gdcmSampleData/US_DataSet/](https://www.creatis.insa-lyon.fr/~jpr/PUBLIC/gdcm/gdcmSampleData/US_DataSet/)

In the first folder are Acusson_US are lossy Jpeg YBR_FULL_422
RLEs YBR_FULL are also there

Modules/IO/GDCM/include/itkGDCMImageIO.h
```diff
@@ -164,6 +164,12 @@ class ITKIOGDCM_EXPORT GDCMImageIO : public ImageIOBase
   itkGetConstMacro(LoadPrivateTags, bool);
   itkBooleanMacro(LoadPrivateTags);
 
+  /** Convert YBR to RGB
+   */
+  itkSetMacro(YBRtoRGB, bool);
+  itkGetConstMacro(YBRtoRGB, bool);
+  itkBooleanMacro(YBRtoRGB);
```
blowekamp: I don't think this option/IVAR is needed. The expectation is that if the ImageIO's PixelType is RGB, then the image buff is RGB not YBR.

If this IVAR is kept than it will need to be added to the "PrintSelf" method.

blowekamp:
With out fixed in GDCM to fix automatic conversion or to fix ChangePhotometricInterpretation class this is the best way to proceed.

We should add a test or two for these cases so that we know if the behavior in GDCM changes with versions.

Modules/IO/GDCM/include/itkGDCMImageIO.h
issakomi: Please kindly preserve the option, it was already discussed in the very first PR. Thank you. There is some advantage to take only Y or something like and it will be backward compatible if someone have already some workaround, IMHO.

issakomi:
Again about YBR_FULL_422, on GDCM site it can be switched [here](https://github.com/malaterre/GDCM/pull/98/files)

Modules/IO/GDCM/include/itkGDCMImageIO.h
blowekamp: Yes it was discussing in the other PR: https://github.com/InsightSoftwareConsortium/ITK/pull/822

My conclusion is that the ImageIO::PixelType must be respected and consistent with RGB. Hence the current ITK behavior is a bug, and requires correction. If one wants a YBR buffer that must not be reported as an RGB PixelType by ImageIO. 

The recommendation is to remove the option, otherwise there will need to be more effort in ensure a YBR buffer is consistent with the ImageIO state.

issakomi:
About using _change photometric interpretation filter_ from GDCM, s. [here it allocates memory](https://github.com/InsightSoftwareConsortium/ITK/blob/3526b67168ded206a8b9a3e7100fb1849c476d8e/Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmImageChangePhotometricInterpretation.cxx#L111), so IMHO in particular case, 6 lines with simple math, in-place is real advantage. And it crashes, BTW.

Modules/IO/GDCM/include/itkGDCMImageIO.h
issakomi: > We should add a test or two for these cases so that we know if the behavior in GDCM changes with versions.

BTW, that option will allow very easy tests what is true and what is not, and workaround even behavior for lossy Jpeg will change in GDCM

issakomi: S. [post](https://github.com/InsightSoftwareConsortium/ITK/pull/822#issuecomment-490552962) 
 in old PR

malaterre:
@blowekamp Give me a few more days to polish support for YBR_FULL_422/raw in GDCM. This will simplify the code on the ITK side.

malaterre:
> About using _change photometric interpretation filter_ from GDCM, s. [here it allocates memory](https://github.com/InsightSoftwareConsortium/ITK/blob/3526b67168ded206a8b9a3e7100fb1849c476d8e/Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmImageChangePhotometricInterpretation.cxx#L111), so IMHO in particular case, 6 lines with simple math, in-place is real advantage. And it crashes, BTW.

Are you able to reproduce the issue with GDCM command line ? I am not. As a side note I am exposing the low level ybr <-> rgb functions. So you may want to re-use them if you really care for non-reallocation.

Modules/IO/GDCM/src/itkGDCMImageIO.cxx
issakomi: @malaterre  If you  [comment two lines here in GDCM,](https://github.com/InsightSoftwareConsortium/ITK/blob/3526b67168ded206a8b9a3e7100fb1849c476d8e/Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmJPEGBITSCodec.hxx#L861) convert wiil be not required for YBR_FULL_422, i have tried PR in GDCM and closed, please see yourself, may be it will have impact on some GDCM tools or whatever

```
      if( GetPhotometricInterpretation() == PhotometricInterpretation::YBR_FULL
      || GetPhotometricInterpretation() == PhotometricInterpretation::YBR_FULL_422 )
        {
// <------  cinfo.jpeg_color_space = JCS_UNKNOWN;
// <------  cinfo.out_color_space = JCS_UNKNOWN;
        //this->PlanarConfiguration = 1;
        }
```

issakomi:
> @blowekamp Give me a few more days to polish support for YBR_FULL_422/raw in GDCM. This will simplify the code on the ITK side.

closed

zivy:
Hello @malaterre, do I understand correctly that the intent is that GDCM will always convert the YBR_FULL_422 images to RGB colorspace? That will work for me as it makes the IO consistent and then I'll convert to grayscale when needed instead of grabbing one of the raw channels. Also as you said that will simplify the code on the ITK side.

@malaterre , @issakomi @blowekamp  I really appreciate all the work you're putting into this. The only thing I can say is that DICOM is a complex file format and it is very hard to support all the options. Again, thanks for all the hard work and I apologize if this resulted in a dive into a deep rabbit hole .

issakomi:
The [branch](https://github.com/issakomi/ITK/tree/ybr2rgb02) is alive for my needs, i've added support for compressions, also write RLE/YBR_FULL and lossy Jpeg/YBR_FULL_422. May be i'll re-open later, after it will be clear with GDCM, or let me know. The branch is now WIP.

Modules/IO/GDCM/include/itkGDCMImageIO.h
issakomi: E.g. example for that option, access as vector[3] image and extract luma.

```
#include <itkImage.h>
#include <itkVectorImage.h>
#include <itkImageFileReader.h>
#include <itkImageFileWriter.h>
#include <itkGDCMImageIO.h>
#include <itkVectorIndexSelectionCastImageFilter.h>
// Input Y'CbCr color-space image, output is Y' scalar image (luma)
int main(int argc, char ** argv)
{
  typedef itk::VectorImage<unsigned char, 3> VectorImageType;
  typedef itk::Image<unsigned char, 3> ImageType;
  typedef itk::ImageFileReader<VectorImageType> ReaderType;
  typedef itk::ImageFileWriter<ImageType> WriterType;
  itk::GDCMImageIO::Pointer imageIO = itk::GDCMImageIO::New();
  ReaderType::Pointer reader = ReaderType::New();
  //imageIO->SetInputYBRtoRGB(false); <----------------------------------------------------
  reader->SetImageIO(imageIO);
  reader->SetFileName(argv[1]);
  reader->Update();
  typedef itk::VectorIndexSelectionCastImageFilter<VectorImageType, ImageType> FilterType;
  FilterType::Pointer filter = FilterType::New();
  filter->SetIndex(0);
  filter->SetInput(reader->GetOutput());
  filter->Update();
  WriterType::Pointer writer = WriterType::New();
  writer->SetUseCompression(true);
  writer->SetFileName("luma.tiff");
  writer->SetInput(filter->GetOutput());
  writer->Update();
  return 0;
}
```

malaterre:
> @blowekamp Give me a few more days to polish support for YBR_FULL_422/raw in GDCM. This will simplify the code on the ITK side.

Support for raw + YBR_FULL_422 has been added in GDCM. You'll need GDCM >= 3.0.3 (contains commit 24a7225ff9a4be995138e01d7b4f5ded5dbc9628)

issakomi:
> Support for raw + YBR_FULL_422 has been added in GDCM. You'll need GDCM >= 3.0.3 (contains commit [24a7225ff9a4be995138e01d7b4f5ded5dbc9628](https://github.com/malaterre/GDCM/commit/24a7225ff9a4be995138e01d7b4f5ded5dbc9628))

It really really good. Thank you!

PR #1292, opened 2019-10-04
BUG: wrong number of parameters was checked in itkTestDriverInclude

### LINKED ISSUE #1289: TestDriverInclude parameter checking inconsistency
Some parameters are checking for one extra `argc`,
https://github.com/InsightSoftwareConsortium/ITK/blob/2074c2f9e5ed3f087d0e2059f8e0e8992fcad7ef/Modules/Core/TestKernel/src/itkTestDriverInclude.cxx#L285
while seemingly equivalent others are checking for two extra `argc`:
https://github.com/InsightSoftwareConsortium/ITK/blob/2074c2f9e5ed3f087d0e2059f8e0e8992fcad7ef/Modules/Core/TestKernel/src/itkTestDriverInclude.cxx#L296

Is this an oversight, or is there some intention behind it?

dzenanz:
This rarely mattered, as there was always more parameters afterwards.
Closes #1289.

### DIFF
--- a/Modules/Core/TestKernel/src/itkTestDriverInclude.cxx
+++ b/Modules/Core/TestKernel/src/itkTestDriverInclude.cxx
@@ -248,7 +248,7 @@ ProcessArguments(int * ac, ArgumentStringType * av, ProcessedOutputType * proces
     }
     else if (!skip && strcmp((*av)[i], "--with-threads") == 0)
     {
-      if (i + 2 >= *ac)
+      if (i + 1 >= *ac)
       {
         usage();
         return 1;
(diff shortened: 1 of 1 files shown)


blowekamp:
I'm not thinking the +1 check then the +2 increment is correct, not what was done here. In the test driver in SimpleITK have have it this way:
https://github.com/SimpleITK/SimpleITK/blob/4a0f36836ca0376b5a8e354da836aee2c367df38/Testing/Unit/TestBase/sitkCompareDriver.cxx#L241-L250

And I wrote a test:
https://github.com/SimpleITK/SimpleITK/blob/4a0f36836ca0376b5a8e354da836aee2c367df38/Testing/Unit/TestBase/CMakeLists.txt#L42-L48

dzenanz:
OK I will change it.

PR #1307, opened 2019-10-09
STYLE: Modernize NeighborhoodAllocator data management

N-Dekker:
Declares `NeighborhoodAllocator::m_Data` as a `unique_ptr`, instead of
a raw pointer. Removes low level `delete[]` from `NeighborhoodAllocator`
implementation.

Fixes a potential memory leak when an ITK user would directly call
`Allocate(n)`, and a potential crash when a user would directly call
`Deallocate()`.

Added unit tests for `NeighborhoodAllocator` member functions,
`Allocate(n)` and `Deallocate()`.

Note: it appears that these functions should not be called directly, by
the end user.


blowekamp:
nice!

thewtex:
CI failing tests unrelated.

PR #1315, opened 2019-10-15
STYLE: For a const image, iterator and const_iterator should be the same

N-Dekker:
Ensured that `RangeType::iterator` and `RangeType::const_iterator` are
the same type, when the range is declared to iterate over a `const`
image:

 - Added compile-time tests to `ImageRegionRange` and
   `ShapedImageNeighborhoodRange`
 - Adjusting `ShapedImageNeighborhoodRange::iterator` accordingly.

Note that `ImageBufferRange` was already OK.

This commit aims to improve consistency between ITK image range types.

### DIFF
--- a/Modules/Core/Common/include/itkShapedImageNeighborhoodRange.h
+++ b/Modules/Core/Common/include/itkShapedImageNeighborhoodRange.h
@@ -637,7 +637,7 @@ private:
 
 public:
   using const_iterator = QualifiedIterator<true>;
-  using iterator = QualifiedIterator<false>;
+  using iterator = QualifiedIterator<IsImageTypeConst>;
   using reverse_iterator = std::reverse_iterator<iterator>;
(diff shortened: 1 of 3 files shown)


blowekamp:
Very nice. Thank you for adding a test.

It this really just style? Doesn't it change a type under some circumstances?

N-Dekker:
> It this really just style? Doesn't it change a type under some circumstances?

Good question @blowekamp With this commit, `iterator` and  `const_iterator` become exactly the same type, for a `const` image type. With `ShapedImageNeighborhoodRange<const Image<T>>`, they used to be different types (prior to this commit). But even in the old situation, `ShapedImageNeighborhoodRange<const Image<T>::iterator` did not allow modifying the image data.

When the image type of a range is `const`, it appears to me that it does not make sense to distinguish between `const_iterator` and non-const `iterator`.

PR #1316, opened 2019-10-15
WIP: STYLE: Move ITK5 ranges, shapes, policies out of Experimental namespace

N-Dekker:
Moved the iterator ranges, image neighborhood shapes, and pixel access
policies, introduced with ITK 5, out of the `Experimental` namespace.

### DIFF
--- a/Modules/Core/Common/include/itkConnectedImageNeighborhoodShape.h
+++ b/Modules/Core/Common/include/itkConnectedImageNeighborhoodShape.h
@@ -38,8 +38,6 @@
 
 namespace itk
 {
-namespace Experimental
-{
 
 /**
  * \class ConnectedImageNeighborhoodShape
@@ -299,7 +297,15 @@ GenerateConnectedImageNeighborhoodShapeOffsets() ITK_NOEXCEPT
   return offsets;
 }
 
+
+#if !defined(ITK_LEGACY_REMOVE)
+namespace Experimental // For ITK 5.0 legacy support
+{
+using ::itk::ConnectedImageNeighborhoodShape;
+using ::itk::GenerateConnectedImageNeighborhoodShapeOffsets;
 } // namespace Experimental
+#endif // ITK_LEGACY_REMOVE
+
 } // namespace itk
 
 #undef ITK_X_ASSERT
--- a/Modules/Core/Common/include/itkConstShapedNeighborhoodIterator.h
+++ b/Modules/Core/Common/include/itkConstShapedNeighborhoodIterator.h
@@ -334,7 +334,7 @@ public:
   }
 
   /** Activates a whole range of offsets, for example, an std::vector<OffsetType>,
-   * which could be from Experimental::GenerateImageNeighborhoodOffsets(shape). */
+   * which could be from GenerateImageNeighborhoodOffsets(shape). */
   template <typename TOffsets>
   void
   ActivateOffsets(const TOffsets & offsets)
--- a/Modules/Core/Common/include/itkConstantBoundaryImageNeighborhoodPixelAccessPolicy.h
+++ b/Modules/Core/Common/include/itkConstantBoundaryImageNeighborhoodPixelAccessPolicy.h
@@ -25,8 +25,6 @@
 
 namespace itk
 {
-namespace Experimental
-{
 
 /**
  * \class ConstantBoundaryImageNeighborhoodPixelAccessPolicy
@@ -148,7 +146,14 @@ public:
   }
 };
 
-} // namespace Experimental
+
+#if !defined(ITK_LEGACY_REMOVE)
+namespace Experimental // For ITK 5.0 legacy support
+{
+using ::itk::ConstantBoundaryImageNeighborhoodPixelAccessPolicy;
+}
+#endif // ITK_LEGACY_REMOVE
+
 } // namespace itk
 
 #endif
--- a/Modules/Core/Common/include/itkImageBufferRange.h
+++ b/Modules/Core/Common/include/itkImageBufferRange.h
@@ -34,8 +34,6 @@
 
 namespace itk
 {
-namespace Experimental
-{
 
 /**
  * \class ImageBufferRange
@@ -686,7 +684,14 @@ MakeImageBufferRange(TImage * const image)
   }
 }
 
-} // namespace Experimental
-} // namespace itk
 
+#if !defined(ITK_LEGACY_REMOVE)
+namespace Experimental // For ITK 5.0 legacy support
+{
+using ::itk::ImageBufferRange;
+using ::itk::MakeImageBufferRange;
+} // namespace Experimental
+#endif // ITK_LEGACY_REMOVE
+
+} // namespace itk
 #endif
--- a/Modules/Core/Common/include/itkImageNeighborhoodOffsets.h
+++ b/Modules/Core/Common/include/itkImageNeighborhoodOffsets.h
@@ -24,8 +24,6 @@
 
 namespace itk
 {
-namespace Experimental
-{
 
 /** Generates the offsets for a neighborhood of the specified shape. */
 template <typename TImageNeighborhoodShape>
@@ -47,7 +45,14 @@ GenerateRectangularImageNeighborhoodOffsets(const Size<VImageDimension> & radius
   return GenerateImageNeighborhoodOffsets(shape);
 }
 
-} // namespace Experimental
-} // namespace itk
 
+#if !defined(ITK_LEGACY_REMOVE)
+namespace Experimental // For ITK 5.0 legacy support
+{
+using ::itk::GenerateImageNeighborhoodOffsets;
+using ::itk::GenerateRectangularImageNeighborhoodOffsets;
+} // namespace Experimental
+#endif // ITK_LEGACY_REMOVE
+
+} // namespace itk
 #endif
--- a/Modules/Core/Common/include/itkImageRegionRange.h
+++ b/Modules/Core/Common/include/itkImageRegionRange.h
@@ -33,8 +33,6 @@
 
 namespace itk
 {
-namespace Experimental
-{
 
 /**
  * \class ImageRegionRange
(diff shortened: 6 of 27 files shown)


N-Dekker:
Doxygen generated overview of everything that is now in the `Experimental` namespace: https://itk.org/Doxygen/html/namespaceitk_1_1Experimental.html

N-Dekker:
Bradley @blowekamp, Matt @thewtex It was originally a suggestion of both of you to place those ranges in an "experimental" namespace. Do you agree that it's now the right time to move them out of `itk::Experimental`? I think they are quite stable now, I don't expect fundamental design changes to them anymore.

Modules/Core/Common/include/itkConnectedImageNeighborhoodShape.h
```diff
@@ -299,7 +297,15 @@ GenerateConnectedImageNeighborhoodShapeOffsets() ITK_NOEXCEPT
   return offsets;
 }
 
+
+#if !defined(ITK_LEGACY_REMOVE)
+namespace Experimental // For ITK 5.0 legacy support
```
dzenanz: Are any remote modules using `Experimental` namespace? If not, we don't need this. By definition, experimental things are subject to change, and my expectation is without backwards compatibility.

dzenanz:
@N-Dekker were the some other STL-style classes you wanted to add, but meanwhile gave up? Or was this it? Or you still plan to add more, but first wanted to get this out of experimental status?

Modules/Core/Common/include/itkConnectedImageNeighborhoodShape.h
N-Dekker: @dzenanz 
> Are any remote modules using Experimental namespace? 

Honestly, I don't know. 

I really hope that some ITK users are already using things from `Experimental`, but again, I don't know. The projects I'm involved with only _indirectly_ make use of them, via `itk::GaussianDerivativeImageFunction`, `itk::N4BiasFieldCorrectionImageFilter`, and `itk::ImageMaskSpatialObject` (which do now internally use `Experimental` things).

N-Dekker:
@thewtex 
> even if there are not fundamental design changes, do you anticipate any API refinements?
> 
> E.g.:
> 
> [8883d3e](https://github.com/InsightSoftwareConsortium/ITK/commit/8883d3eea91b9708784460dcd9f26d0b30d95a39)
> 
> is a refinement from yesterday, which is excellent, but we should leave it marked as `Experimental` unless we think the API is ready to be "frozen".

I feel that that particular refinement (ensuring that `iterator` and `const_iterator` are the same type when the image type is `const`) is very minor, and probably won't be noticed by users of the range class. (The API documentation does not even specify whether or not `iterator` and `const_iterator` are the same type, for `const` images.) Such a small refinement seems acceptable to me, even for non-experimental classes.

Requests for other API changes could come from ITK users. But I'm afraid ITK users might not use the ranges, as long as they are still in the `Experimental` namespace. 

No problem, if you think it's too early to move the ranges (and related classes) out of `Experimental`. But then, what do you think should still be done, to get them out of `Experimental`? Or should it just wait a little longer? Maybe until ITK 5.2? 😃

blowekamp:
My quick 2 cents:
  - I'd like to see more classes updated to use these new tools. This will server as more examples and test different use cases for the new iterators.
 - I wish I used these classes by now. I had though about update the SLIC filter to use these great new iterators.
 - As I recall, I had the preference to be more explicit on the region being iterated on to enable an assumption/prerequisite check. Perhaps this come from my experience with streaming buffers where there are many different regions streaming about, and it's not uncommon for a bug or wrong assumption to be made.

> Requests for other API changes could come from ITK users. But I'm afraid ITK users might not use the ranges, as long as they are still in the Experimental namespace.

If they are used in more examples and existing ITK classes they will be adopted quicker ( no matter which namespace they are in ).

blowekamp:
I'll also add that you have done an amazing job consistently writing little tests for your code. Big complements there.

N-Dekker:
Thanks for your comments, Bradley @blowekamp 

> I'd like to see more classes updated to use these new tools.

I would like to see that as well. However, personaly I feel reluctant to make fundamental chances to old ITK filter, unless I'm actively using the class in a project that I'm involved with. Technically, changing the implementation of an old ITK filter to use the new iterator ranges could yield a coding style improvement, as well as a performance boost. Please note that such a change may _not always_ increase the performance! It still depends on the use case!  Certainly, `itk::GaussianDerivativeImageFunction`, `itk::N4BiasFieldCorrectionImageFilter`, and `itk::ImageMaskSpatialObject` did get a significant performance boost from using the new iterator ranges.


> As I recall, I had the preference to be more explicit on the region being iterated on to enable an assumption/prerequisite check.

Both `ImageRegionIndexRange` (the specialization of `IndexRange` for image regions) and `ImageRegionRange` do explicitly require the region to be specified.

`ShapedImageNeighborhoodRange` iterates over a (typically very small) neighborhood of pixels, not over an entire region, so it does not seem necessary for users to explicitly specify the region of the neighborhood.

`ImageBufferRange` iterates over the entire buffered region, I think that's clear already from the name "ImageBufferRange". I hope you agree. 😃  We have extensively discussed this before, starting at https://discourse.itk.org/t/imagerange-a-range-of-iterators-to-the-pixels-of-an-image/1455/6

Kind regards, Niels

N-Dekker:
Back to WIP - work in progress. A few recent commits assume that the iterator ranges are defined within the `Experimental` namespace (commit 321843a4b6eaa4f2415315149375bfcf099f744c, commit e3fd64de1408c79c50fb17e3f046e1d5d98a89c6, commit 6d7f192f2a1e4d111af2dc9257e672851a24b2bc), so  a merge would now cause compilation errors.

More importantly, there is no consensus on whether this is the right time to move the iterator ranges out of the `Experimental` namespace right now. Maybe then ITK 5.2? 😄

dzenanz:
5.2 sounds good to me. Tagging 5.1 isn't so far away any more. And an updated version of this could be merged right after.

thewtex:
+1 for 5.2 :-D

N-Dekker:
How can this pull request be marked as "targeted for ITK 5.2"?

dzenanz:
Click on the `Milestone` link + gear on the right, and choose 5.2. But that milestone needs to be created first, as it currently doesn't exist. See if your newly acquired powers allow you to do that 😉

dzenanz:
We are still on this 😄

dzenanz:
This PR could and should now be revived.

N-Dekker:
> This PR could and should now be revived.

Thanks for the reminder, @dzenanz Before merging, should we double-check that the chosen identifiers are OK? It might be somewhat harder to rename them afterwards, once they are out of the `Exerimental` namespace.

    Classes:
      IndexRange
      ImageBufferRange
      ImageRegionRange
      ShapedImageNeighborhoodRange
      ConnectedImageNeighborhoodShape
      RectangularImageNeighborhoodShape
      ConstantBoundaryImageNeighborhoodPixelAccessPolicy
      ZeroFluxNeumannImageNeighborhoodPixelAccessPolicy

    Typedefs:
      ImageRegionIndexRange
      ZeroBasedIndexRange

    Functions:
      MakeImageBufferRange(image)
      GenerateImageNeighborhoodOffsets(shape)
      GenerateConnectedImageNeighborhoodShapeOffsets()
      GenerateRectangularImageNeighborhoodOffsets(radius)

I find some of the identifiers a bit lengthy. But then, it would already help if they do not need to be prefixed by `Exerimental::`  😄 

As you already know, I never really liked "ZeroFluxNeumann", but we discussed that extensively before, no need to redo the discussion! 😃 

And then I feel somewhat uncertain about [ConnectedImageNeighborhoodShape](https://itk.org/Doxygen/html/classitk_1_1Experimental_1_1ConnectedImageNeighborhoodShape.html) now. Although IMO it does describe the class correctly, it appears to look similar to [itk::NeighborhoodConnectedImageFilter](https://itk.org/Doxygen/html/classitk_1_1NeighborhoodConnectedImageFilter.html), while these classes are actually unrelated.

If it's alright to you right now @thewtex @blowekamp @dzenanz , I will just prepare to make this PR ready.

dzenanz:
I remember we discussed most of the names already, and as such I think they are OK. Most of them are a bit lengthy, so if you or someone else can come up with shorter names, that would be good.

N-Dekker:
> I remember we discussed most of the names already, and as such I think they are OK. Most of them are a bit lengthy, so if you or someone else can come up with shorter names, that would be good.

Thanks @dzenanz Honestly I have no shorter names in mind. At a certain moment I though we should distinguish between a range that yields _index_ _values_ (`ImageRegionIndexRange`, `ZeroBasedIndexRange`) and a range that yields _pixel_ _values_. The first category ends with "IndexRange", so then the second category should then have a name ending with "PixelRange". But then we would get something like "ShapedImageNeighborhood _PixelRange_". I thought that would be far too lengthy, so I dropped the idea long ago  😸

Modules/Core/Common/include/itkConnectedImageNeighborhoodShape.h
N-Dekker: @dzenanz You wrote on Oct 16, 2019 https://github.com/InsightSoftwareConsortium/ITK/pull/1316/commits/27b703931bbf730ae3945cca22ec7adaffd2a9ba#r335469363 
>Are any remote modules using Experimental namespace? If not, we don't need this

Do you still think so? (Just double-checking.) Then I will make sure that the PR will no longer add `#if !defined(ITK_LEGACY_REMOVE)`  ITK 5.0 backwards compatibility.

dzenanz: Number of remotes to use this is probably 0-2 (skewed towards zero). Removing `Experimental::` prefix should be easy to do, and easy to deduce as the solution in the case the problem arises.

N-Dekker:
Hmmm... the original PR branch isn't there anymore. You see, it says:

> N-Dekker wants to merge 1 commit into InsightSoftwareConsortium:master from unknown repository

I think it should be from N-Dekker:Move-out-of-Experimental !!!

dzenanz:
You can make a new PR, and reference this PR for discussion. I don't know why GitHub lost track of your branch.

thewtex:
:+1: for moving ahead with the current names and out of `::Experimental`!

Exciting! :tada: :taco: :1st_place_medal:

N-Dekker:
Thank you @thewtex  I will now close this PR, as it is superseded by  #1880

(Somehow I could not figure out how to continue on PR #1316 anymore, as it appears "from unknown repository")

PR #1321, opened 2019-10-16
Clang format

dzenanz:
I guess the tube SO change was in a PR parallel to the style change, thus was missed by the new style. The other changes might be due to difference in versions 8.0.0 and 8.0.1 (which I have).

### DIFF
--- a/Modules/Core/Common/test/itkMemoryProbesCollecterBaseTest.cxx
+++ b/Modules/Core/Common/test/itkMemoryProbesCollecterBaseTest.cxx
@@ -69,10 +69,9 @@ itkMemoryProbesCollecterBaseTest(int, char *[])
   delete[] buf;
   Sleep(5000);
   probe.Stop();
-  if ( total != 0 && total < probe.GetTotal() )
+  if (total != 0 && total < probe.GetTotal())
   {
-    std::cerr << "Freeing memory should result in less memory but it is "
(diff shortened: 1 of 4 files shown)


dzenanz:
@hjmjohnson Can this script be used to format remote modules? I ran into this when I tried:
```text
Dzenan@Ryzenator MINGW64 /c/Dev/ITK-git/Modules/Remote/PhaseSymmetry (master)
$ ../../../Utilities/Maintenance/clang-format.bash --clang-format /c/Dev/ITK-2019/ClangFormat-prefix/src/ClangFormat/clang-format.exe --tracked
error: cannot use -i when reading from stdin.
```

hjmjohnson:
> @hjmjohnson Can this script be used to format remote modules? I ran into this when I tried:
> 
> ```
> Dzenan@Ryzenator MINGW64 /c/Dev/ITK-git/Modules/Remote/PhaseSymmetry (master)
> $ ../../../Utilities/Maintenance/clang-format.bash --clang-format /c/Dev/ITK-2019/ClangFormat-prefix/src/ClangFormat/clang-format.exe --tracked
> error: cannot use -i when reading from stdin.
> ```

@dzenanz I'm not 100% sure.  Seems like we may need to add a directory change so that git uses the correct information for determining which files have changed.

PR #1323, opened 2019-10-19
STYLE: Only python > 3.5 is supported

hjmjohnson:
Remove conditional code that is no longer valid for
old support of python 2.6.


Modules/Bridge/VtkGlue/wrapping/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 itk_wrap_module(ITKVtkGlue)
-  if(PYTHON_VERSION_STRING VERSION_GREATER 3.0 AND "${VTK_VERSION}" VERSION_LESS 7.0.0)
+  if("${VTK_VERSION}" VERSION_LESS 7.0.0)
     message(WARNING
       "The ITKVtkGlue module can only built with Python 3 and VTK >= 7.\n"
```
blowekamp: The error message should be updated as well.

hjmjohnson: The error message is still correct, although redundant in stating that Python 3 is required.  In this case, I chose to leave the redundant documemtation.

PR #1324, opened 2019-10-20
PERF: Use ShapedImageNeighborhoodRange in SumOfSquaresImageFunction

N-Dekker:
Improved the performance of `SumOfSquaresImageFunction::EvaluateAtIndex`

Using `ShapedImageNeighborhoodRange` this way will improve the
performance of `EvaluateAtIndex`, because the range does not do any
dynamic memory allocation, and because its "shape offsets",
`m_NeighborhoodOffsets` are generated beforehand (outside of
`EvaluateAtIndex`).

`ImageFunction::EvaluateAtIndex` is typically called repetitively
("for each pixel"), so it appears worth optimizing this member function.

### DIFF
--- a/Modules/Core/ImageFunction/include/itkSumOfSquaresImageFunction.h
+++ b/Modules/Core/ImageFunction/include/itkSumOfSquaresImageFunction.h
@@ -19,8 +19,11 @@
 #define itkSumOfSquaresImageFunction_h
 
 #include "itkImageFunction.h"
+#include "itkImageNeighborhoodOffsets.h"
 #include "itkNumericTraits.h"
-#include "itkNeighborhood.h"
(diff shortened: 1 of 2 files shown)


N-Dekker:
> But the class was only used in ConfidenceConnectedImageFilter. I'm not sure it's tested any more either.

@blowekamp Very well observed! Actually I already wrote a (very drafty) test for `SumOfSquaresImageFunction` but I wasn't sure if it was worth it! Honestly my main aim was to present some examples of how to use the new ranges effectively.

Do you think it's worth adding a `SumOfSquaresImageFunction` test?

blowekamp:
Since you already wrote something... You might as well check it in, with the least effort. There should be something that compiles runs this code.

PR #1326, opened 2019-10-20
PERF: Use ShapedImageNeighborhoodRange in ConfidenceConnectedImageFilter

N-Dekker:
Improved the performance of `ConfidenceConnectedImageFilter` by
removing its local `MeanImageFunction` and `SumOfSquaresImageFunction`
variables, and doing the calculation of mean and sum of squares by using `Experimental::ShapedImageNeighborhoodRange`.

Using `ShapedImageNeighborhoodRange` this way will improve the
performance of `GenerateData()`, because the range does not do any
dynamic memory allocation, and the calculation of the mean and the sum
of squares of neighborhood pixels happens in one and the same iteration.


blowekamp:
Very nice!

PR #1328, opened 2019-10-21
ENH: Generate external module export header in ITKCommon_BINARY_DIR

thewtex:
This reduces complexity and the required number of include directories
in a build.


thewtex:
It turns out that `ITKCommon_BINARY_DIR` is not defined with an external module build. Also, we may not be able to write to its directory; abandoning.

PR #1329, opened 2019-10-21
PERF: Use ShapedImageNeighborhoodRange in MeanImageFunction

N-Dekker:
Improved the performance of `MeanImageFunction::EvaluateAtIndex`

Similar to a previous commit:
"PERF: Use ShapedImageNeighborhoodRange in SumOfSquaresImageFunction"
SHA-1 e3fd64de1408c79c50fb17e3f046e1d5d98a89c6

### DIFF
--- a/Modules/Core/ImageFunction/include/itkMeanImageFunction.hxx
+++ b/Modules/Core/ImageFunction/include/itkMeanImageFunction.hxx
@@ -19,7 +19,9 @@
 #define itkMeanImageFunction_hxx
 
 #include "itkMeanImageFunction.h"
-#include "itkConstNeighborhoodIterator.h"
+
+#include "itkImage.h"
+#include "itkShapedImageNeighborhoodRange.h"
 
 namespace itk
 {
@@ -37,7 +39,9 @@ MeanImageFunction<TInputImage, TCoordRep>::EvaluateAtIndex(const IndexType & ind
 
   sum = NumericTraits<RealType>::ZeroValue();
 
-  if (!this->GetInputImage())
+  const InputImageType * const image = this->GetInputImage();
+
+  if (image == nullptr)
   {
     return (NumericTraits<RealType>::max());
   }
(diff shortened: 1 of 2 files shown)


blowekamp:
There appears to be an existing test for this class: Modules/Core/ImageFunction/test/itkMeanImageFunctionTest.cxx

But it seemed to be only used but the ConfidenceConnectedImageFilter.

Modules/Core/ImageFunction/include/itkMeanImageFunction.hxx
```diff
@@ -47,27 +51,28 @@ MeanImageFunction<TInputImage, TCoordRep>::EvaluateAtIndex(const IndexType & ind
...
-  return sum;
+  return sum / static_cast<double>(neighborhoodRange.size());
 }
 
+
+template <typename TInputImage, typename TCoordRep>
+void
+MeanImageFunction<TInputImage, TCoordRep>::SetNeighborhoodRadius(const unsigned int radius)
+{
+  m_NeighborhoodOffsets = Experimental::GenerateRectangularImageNeighborhoodOffsets(ImageSizeType::Filled(radius));
```
thewtex: Here we want to compare to the previous value of `radius` and call `this->Modified()` if son, per:

https://github.com/InsightSoftwareConsortium/ITK/blob/8e4ab68c240f439a0b188796794f2c7066ea8a46/Modules/Core/Common/include/itkMacro.h#L875-L879

N-Dekker: @thewtex Thanks, good catch, Matt! I just did a force-push , having added a compare to the previous radius value and `this->Modified()` call. Now I see, `itkSetMacro` declares the function `virtual`: https://github.com/InsightSoftwareConsortium/ITK/blob/8e4ab68c240f439a0b188796794f2c7066ea8a46/Modules/Core/Common/include/itkMacro.h#L871 

Do you still want `SetNeighborhoodRadius` to be `virtual`? Personally I don't think it's necessary. You know, we have put a lot of effort into removing `virtual` keywords from the past!

N-Dekker:
The remaining question is: should `SetNeighborhoodRadius` be `virtual`? (Personally I'd rather have it non-virtual, as the pull request is right now.)

PR #1333, opened 2019-10-22
DOC: Add HDF5 copyright to NOTICE file

thewtex:


blowekamp:
This PR looks good.

We should consider spending the effort to "modularize" the copyright license files. I would propose added a new option to the `itk-module` macro to specify the additional copyrights and licenses.

thewtex:
> We should consider spending the effort to "modularize" the copyright license files. I would propose added a new option to the `itk-module` macro to specify the additional copyrights and licenses.

An implementation like this could work. Considering the added complexity and sparse addition of third party licenses, I do not think it should be a priority.

thewtex:
Also merged to the `master` branch.

PR #1335, opened 2019-10-22
PERF: Call FillBuffer in watershed Segmenter::InitializeBoundary()

N-Dekker:
Improved the performance of `watershed::Segmenter::InitializeBoundary()`
by replacing its iteration over the buffered region of a face image by
a simple `Image::FillBuffer` call.

### DIFF
--- a/Modules/Segmentation/Watersheds/include/itkWatershedSegmenter.hxx
+++ b/Modules/Segmentation/Watersheds/include/itkWatershedSegmenter.hxx
@@ -405,10 +405,9 @@ template <typename TInputImage>
 void
 Segmenter<TInputImage>::InitializeBoundary()
 {
-  ImageRegionIterator<typename BoundaryType::face_t> faceIt;
-  typename BoundaryType::face_t::Pointer             face;
-  typename BoundaryType::face_pixel_t                fps;
-  BoundaryIndexType                                  idx;
+  typename BoundaryType::face_t::Pointer face;
+  typename BoundaryType::face_pixel_t    fps;
+  BoundaryIndexType                      idx;
 
   fps.flow = NULL_FLOW;
   fps.label = NULL_LABEL;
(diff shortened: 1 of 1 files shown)


Modules/Segmentation/Watersheds/include/itkWatershedSegmenter.hxx
```diff
@@ -423,11 +424,8 @@ Segmenter<TInputImage>::InitializeBoundary()
       }
       this->GetBoundary()->GetFlatHash(idx)->clear();
       face = this->GetBoundary()->GetFace(idx);
-      faceIt = ImageRegionIterator<typename BoundaryType::face_t>(face, face->GetBufferedRegion());
-      for (faceIt.GoToBegin(); !faceIt.IsAtEnd(); ++faceIt)
-      {
-        faceIt.Set(fps);
-      }
+      const auto imageBufferRange = Experimental::MakeImageBufferRange(face.GetPointer());
+      std::fill(imageBufferRange.begin(), imageBufferRange.end(), fps);
```
N-Dekker: Or maybe just `face->FillBuffer(fps)`...?

dzenanz: I expect that `FillBuffer` should be as fast or faster than `std::fill`.

dzenanz: Assuming `face` is an image.

N-Dekker: > Assuming `face` is an image.

Yes it is  😄 `face`is a (smart) pointer to `face_t`, which is defined as: 

    using face_t = Image<face_pixel_t, TDimension>;

https://github.com/InsightSoftwareConsortium/ITK/blob/v5.1b01/Modules/Segmentation/Watersheds/include/itkWatershedBoundary.h#L117

`face->FillBuffer(fps)` is clearer and faster than the original code, but it's also simpler than `std::fill(imageBufferRange.begin(), imageBufferRange.end(), fps)`, I must admit. 😸

N-Dekker:
"All checks have passed", "4 successful checks", but it did not rebuild (after my force-push), so it seems!

PR #1337, opened 2019-10-22
PERF: Use ImageBufferRange in Testing::StretchIntensityImageFilter

N-Dekker:
Improved the performance of `Testing::StretchIntensityImageFilter`
member function `BeforeThreadedGenerateData()` by replacing its local
`ImageRegionConstIteratorWithIndex` variable by an
`Experimental::ImageBufferRange`.

In general, when a function iterates over an entire image buffer,
`ImageBufferRange` is significantly faster than
`ImageRegionConstIteratorWithIndex`.

### DIFF
--- a/Modules/Core/TestKernel/include/itkTestingStretchIntensityImageFilter.hxx
+++ b/Modules/Core/TestKernel/include/itkTestingStretchIntensityImageFilter.hxx
@@ -29,6 +29,7 @@
 #define itkTestingStretchIntensityImageFilter_hxx
 
 #include "itkTestingStretchIntensityImageFilter.h"
+#include "itkImageBufferRange.h"
 #include "itkImageRegionIterator.h"
 #include "itkImageRegionConstIterator.h"
 #include "itkMath.h"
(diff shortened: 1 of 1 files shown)


N-Dekker:
> It would be good if you did some actual performance tests for this recent batch of changes regarding the use of ImageBufferRange.

@dzenanz I just observed the effect of this particular commit: for a filter of type `StretchIntensityImageFilter` and a large input image (3-D: 1024x1024x1024), using VS2017 64-bit Release, I observed that the duration of `filter->Update()` dropped from 2.7 seconds to 1.6 seconds, approximately. You may give it a try, if you want to:

    #include <itkTestingStretchIntensityImageFilter.h>
    #include <itkImage.h>
    #include <chrono>
    #include <iostream>

    int main()
    {
      image->SetRegions(ImageType::SizeType::Filled(1024));
      image->Allocate(true);

      const auto filter = itk::Testing::StretchIntensityImageFilter<ImageType>::New();

      for (int i{}; i < 8; ++i)
      {
        filter->SetInput(nullptr);
        filter->SetInput(image);

        using namespace std::chrono;

        filter->Update();

        const auto durationSeconds =
        std::cout
          << std::endl;
      }
    }

dzenanz:
> filter->Update() dropped from 2.7 seconds to 1.6 seconds

For me it goes from 4.2 to 1.5 seconds (VS2019 x64 RelWithDebInfo on 8-core Ryzen). Awesome!

N-Dekker:
> For me it goes from 4.2 to 1.5 seconds (VS2019 x64 RelWithDebInfo on 8-core Ryzen). Awesome!

Thanks for trying, @dzenanz ! Honestly, the example already would have performed much better if `itk::ImageRegionConstIteratorWithIndex` would have been replaced, simply by the old `itk::ImageRegionConstIterator` (without `WithIndex`). But as far as I can see, `Experimental::ImageBufferRange` would still be the fastest anyway  😄

PR #1339, opened 2019-10-22
COMP: compile the test with legacy disabled

dzenanz:
This should have been part of 71c90c83de8631e655144ad9eb6a462bcf059f56.


dzenanz:
Conflicts with #1338.

PR #1340, opened 2019-10-22
ENH: Update CI ExternalDataVersion to 5.0.1

thewtex:
It looks like GitHub releases is failing to serve 5.0.0 ExternalData tarballs. This may not resolve the issue, but it will not hurt.


thewtex:
GitHub is currently experiencing some issues according to:

  https://www.githubstatus.com

PR #1344, opened 2019-10-23
PERF: Use ShapedImageNeighborhoodRange in MedianImageFunction

N-Dekker:
Improved the performance of `MedianImageFunction::EvaluateAtIndex`

Analog to recent commits on other `ImageFunction` templates:
- "PERF: Use ShapedImageNeighborhoodRange in MeanImageFunction"
  SHA-1 7e6a391160df7551cb9f3da416d3ac3d3f02e441
- "PERF: Use ShapedImageNeighborhoodRange in SumOfSquaresImageFunction"
  SHA-1 e3fd64de1408c79c50fb17e3f046e1d5d98a89c6

### DIFF
--- a/Modules/Core/ImageFunction/include/itkMedianImageFunction.hxx
+++ b/Modules/Core/ImageFunction/include/itkMedianImageFunction.hxx
@@ -19,7 +19,8 @@
 #define itkMedianImageFunction_hxx
 
 #include "itkMedianImageFunction.h"
-#include "itkConstNeighborhoodIterator.h"
+#include "itkImage.h"
+#include "itkShapedImageNeighborhoodRange.h"
 
 #include <vector>
 #include <algorithm>
@@ -32,6 +33,19 @@ namespace itk
 template <typename TInputImage, typename TCoordRep>
 MedianImageFunction<TInputImage, TCoordRep>::MedianImageFunction()
 {}
+
+
+template <typename TInputImage, typename TCoordRep>
+void
+MedianImageFunction<TInputImage, TCoordRep>::SetNeighborhoodRadius(const unsigned int radius)
+{
+  if (m_NeighborhoodRadius != radius)
+  {
+    m_NeighborhoodOffsets = Experimental::GenerateRectangularImageNeighborhoodOffsets(ImageSizeType::Filled(radius));
+    m_NeighborhoodRadius = radius;
+    this->Modified();
+  }
+}
 
 
 /**
@@ -52,9 +66,9 @@ template <typename TInputImage, typename TCoordRep>
 typename MedianImageFunction<TInputImage, TCoordRep>::OutputType
 MedianImageFunction<TInputImage, TCoordRep>::EvaluateAtIndex(const IndexType & index) const
 {
-  unsigned int i;
+  const InputImageType * const image = this->GetInputImage();
 
-  if (!this->GetInputImage())
+  if (image == nullptr)
   {
     return (NumericTraits<OutputType>::max());
   }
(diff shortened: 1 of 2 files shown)


N-Dekker:
It appears that with this commit, `MedianImageFunction::EvaluateAtIndex` may become more than 4 times as fast as before! For a 1024x1024 image, I observed that the duration of calling `EvaluateAtIndex` on each pixel dropped from ~0.71 second (before the commit) to ~0.15 second (after the commit)!

I tested the following, using Visual C++ 2017 64-bit (Release):

    #include <itkMedianImageFunction.h>
    #include <itkIndexRange.h>
    #include <itkImage.h>

    #include <chrono>
    #include <iostream>

    int main()
    {
      using ImageType = itk::Image<char>;
      const auto imageSize = ImageType::SizeType{ {1024, 1024} };
      image->SetRegions(imageSize);
      image->Allocate(true);

      const auto imageFunction = itk::MedianImageFunction<ImageType>::New();

      for (int i{}; i < 8; ++i)
      {
        imageFunction->SetInputImage(nullptr);
        imageFunction->SetInputImage(image);

        using namespace std::chrono;

        for (const auto index : itk::Experimental::ZeroBasedIndexRange<2>(imageSize))
        {
          imageFunction->EvaluateAtIndex(index);
        }

        const auto durationSeconds =
        std::cout
          << std::endl;
      }
    }

Modules/Core/ImageFunction/include/itkMedianImageFunction.hxx
```diff
@@ -34,6 +35,19 @@ MedianImageFunction<TInputImage, TCoordRep>::MedianImageFunction()
 {}
 
 
+template <typename TInputImage, typename TCoordRep>
+void
+MedianImageFunction<TInputImage, TCoordRep>::SetNeighborhoodRadius(const unsigned int radius)
+{
+  if (m_NeighborhoodRadius != radius)
+  {
+    m_NeighborhoodOffsets = Experimental::GenerateRectangularImageNeighborhoodOffsets(ImageSizeType::Filled(radius));
```
N-Dekker: I see now, there should have been a `GenerateSquareImageNeighborhoodOffsets(unsigned radius)` helper function, creating a _square_ neighborhood shape, instead of any rectangular shape. Then the `ImageSizeType::Filled(radius)`call would not have been necessary anymore.

dzenanz: `GenerateSquareImageNeighborhoodOffsets` would have been to specialized. There is a maintenance burden associated with more code. Rectangular + `ImageSizeType::Filled(radius)` is not hard to write.

dzenanz:
> Is this used by MedianImage**Filter**? Median filter sees a lot of use.

It is not.

Modules/Core/ImageFunction/include/itkMedianImageFunction.hxx
N-Dekker: @dzenanz  I see that there's a trade-off here. It appears to me that the majority of neighborhood shapes are just square (or cubic, in 3-D). Please have a look at pull request #1347 and reject it, if you still believe the maintenance burden would be too much!

N-Dekker:
> > Is this used by MedianImage**Filter**? Median filter sees a lot of use.
> 
> It is not.

I see, MedianImage**Filter** still uses the old `itk::ConstNeighborhoodIterator`. Its performance _might_ also be improved by using `Experimental::ShapedImageNeighborhoodRange`, but I don't expect such a great performance improvement in this case!

In general , the largest performance costs of using `itk::ConstNeighborhoodIterator` came from its constructor and its `SetLocation` member function. `MedianImageFilter` only calls the `itk::ConstNeighborhoodIterator` constructor _outside_ of the region iteratation loop, and it does not call  `SetLocation` at all.

https://github.com/InsightSoftwareConsortium/ITK/blob/v5.1b01/Modules/Filtering/Smoothing/include/itkMedianImageFilter.hxx#L66

N-Dekker:
FYI, the force-push that I just did, did not modify any code. It only extended the commit message a little bit.

N-Dekker:
@dzenanz 
> Median filter sees a lot of use.

In the mean time I did find out that the performance of `MedianImageFilter` could really be improved very much (!) by using `ShapeImageNeighborhoodRange`, instead of  `ConstNeighborhoodIterator`. However, it's not as easy as a simple find-and-replace. `ShapeImageNeighborhoodRange` would need to be extended to support faster pixel access for the inner image region. For the inner region, border extrapolation (boundary conditions) is irrelevant, and boundary checks can be skipped. `ConstNeighborhoodIterator` already does so, but `ShapeImageNeighborhoodRange` can do it even faster... At least once it is implemented 😄

Anyway, that's far beyond the scope of this particular PR!

PR #1347, opened 2019-10-24
ENH: Add Experimental::SquareImageNeighborhoodShape

N-Dekker:
Added the function `GenerateSquareImageNeighborhoodOffsets(size_t radius)`
and the class `SquareImageNeighborhoodShape` to the `Experimental`
namespace, to allow specifying more explicitly and more efficiently that
the neighborhood shape should be square (or n-dimensional hypercubic).

Replaced the more general `GenerateRectangularImageNeighborhoodOffsets`
by the newly added `GenerateSquareImageNeighborhoodOffsets` in
`MeanImageFunction`, `SumOfSquaresImageFunction`, and
`ConfidenceConnectedImageFilter`

### DIFF
--- a/Modules/Core/Common/include/itkImageNeighborhoodOffsets.h
+++ b/Modules/Core/Common/include/itkImageNeighborhoodOffsets.h
@@ -20,6 +20,7 @@
 #define itkImageNeighborhoodOffsets_h
 
 #include "itkRectangularImageNeighborhoodShape.h"
+#include "itkSquareImageNeighborhoodShape.h"
 #include <vector>
 
 namespace itk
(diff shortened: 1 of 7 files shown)


N-Dekker:
Thanks for your feedback, @dzenanz 

First of all, I don't feel too attached to this pull request, I wouldn't feel too bad about a rejection. However...
 - It appears that the fast majority of neighborhood shapes are in practice square-shaped (or cubic-shaped in 3-D). I would like to have it as easy as possible for the majority.
 - It seems clearer to me, when you want a square shape, that you can say so explicitly (instead of by saying you want a rectangular shape of size N x N).
 - The proposed `SquareImageNeighborhoodShape` might serve as a simple example for others, of how to write an ITK neighborhood shape class.

What do you think? 👍 or 👎   😃

N-Dekker:
OK, I can give it another 24 hours, but tomorrow I think I'll just close and abandon this pull request. As it seems that nobody really likes the little square (or cubic) neighborhood shape utility that I'm proposing here... 😢

N-Dekker:
OK: @jhlegarreta like the idea (thanks Jon  😄 ) but @dzenanz does not, and the pull request does not seem to have a priority to anyone else, so I give it up for now...  Close without merge!

PR #1349, opened 2019-10-25
ENH: ITK python cleanup

hjmjohnson:
Default to building python bindings without backward compatibility support.

### DIFF
--- a/Modules/Core/Common/include/itkMacro.h
+++ b/Modules/Core/Common/include/itkMacro.h
@@ -522,9 +522,14 @@ OutputWindowDisplayDebugText(const char *);
 //----------------------------------------------------------------------------
 // Setup legacy code policy.
 //
-// CMake options ITK_LEGACY_REMOVE and ITK_LEGACY_SILENT are converted
-// They may be used to completely remove legacy code or silence the
-// warnings.  The default is to warn about their use.
+// CMake options:
+//  - When ITK_LEGACY_REMOVE:BOOL=ON, legacy code is hidden, thus causing compiler errors for code that depends on it
+//  - When ITK_LEGACY_REMOVE:BOOL=OFF, and ITK_LEGACY_SILENT:BOOL=ON, use
+//    of legacy code will not produce compiler warnings.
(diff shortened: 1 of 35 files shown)


hjmjohnson:
> Looks good, but should similar thing be done for files touched by [0de84a3](https://github.com/InsightSoftwareConsortium/ITK/commit/0de84a3b9b156d6c9e80f34d50aa9c679709d4db) and [8f7f709](https://github.com/InsightSoftwareConsortium/ITK/commit/8f7f709b2331ee7fcfa1b818cc4490583afe84e3)? Do we need to do the same for remote modules which have wrapping? A recent round of fixes was finalized by [9e04dc9](https://github.com/InsightSoftwareConsortium/ITK/commit/9e04dc940e7e697dfaad52c3c847fa807bac9f2d).

I don't' think so.  Only the wrapping of the enums themselves needs to be wrapped to expose the individual elements at the parent level for wrapping purposes.

Hans

Modules/Core/Common/include/itkMacro.h
```diff
@@ -545,18 +550,20 @@ OutputWindowDisplayDebugText(const char *);
...
-#  define itkLegacyMacro(method) method
 #else
-// Setup compile-time warnings for uses of deprecated methods if
-// possible on this compiler.
-#  if defined(__GNUC__) && !defined(__INTEL_COMPILER)
-#    define itkLegacyMacro(method) method __attribute__((deprecated))
-#  elif defined(_MSC_VER)
-#    define itkLegacyMacro(method) __declspec(deprecated) method
-#  else
+#  if defined(ITK_LEGACY_SILENT) || defined(ITK_LEGACY_TEST)
```
dzenanz: This way of defining affects `Modules/Core/Common/test/itkSTLThreadTest.cxx`

dzenanz: Disregard - I tried to delete this comment.

hjmjohnson:
@dzenanz All green.  I'm going to do a force push with the ',' -> '.' comment change and then merge.

PR #1351, opened 2019-10-25
ENH: Run larger memory tests if supported

hjmjohnson:
Allow for tests that may use upto 80% of the computer memory to be used for testing.
Use introspection to automatically identify how much memory is available on the
computer where testing is being performed.


CMakeLists.txt
```diff
@@ -407,9 +407,13 @@ set(ITK_DATA_ROOT ${ITK_SOURCE_DIR}/Testing/Data)
 # Location of ITK Example Data.
 set(ITK_EXAMPLE_DATA_ROOT "${ITK_SOURCE_DIR}/Examples/Data")
 
-#This flag is used in particular, to enable some tests that require large memory to run.
-# This probably can be discovered at configuration time by CMake. (Check with CMake developers).
-set(ITK_COMPUTER_MEMORY_SIZE 1 CACHE STRING "Provide here the size of your RAM in gibibytes")
+# This flag is used in particular, to enable some tests that require large memory to run.
+# Allow for tests that may use upto 80% of the computer memory to be used for testing.
```
blowekamp: We should also document that tests which use a significant amount of memory, also need to add the test property "RESOURCE_LOCK MEMORY_SIZE". This restricts test execution so that only one memory intensive test is run at once.

dzenanz: I think that tests with _large_ in their name are good candidates for checking, as well as those with label `RUNS_LONG`.

seanm:
What happens when you use say 'ctest -j16' to run tests in parallel?  If 16 tests try to use all the host's memory, it will be crushing.

CMakeLists.txt
hjmjohnson: This was already instrumented.  I can not find any cases where this is not true.

blowekamp:
> What happens when you use say 'ctest -j16' to run tests in parallel? If 16 tests try to use all the host's memory, it will be crushing.

@seanm As per my comment we are setting the the ctest proper `[RESOURCE_LOCK MEMORY_SIZE`](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Registration/Common/test/RegistrationITKv3/test/CMakeLists.txt#L483). This is a lock on the memory resource.

So for your `-j 16` example only one test can hold the "MEMORY_SIZE" resource lock at a time. So only one of the big memory tests will be running at once.

Hence, my request to **document** how this value is used.

seanm:
So "ITK_COMPUTER_MEMORY_SIZE" is now gone?  Bot maintainers like me should therefore remove it from their scripts right?

hjmjohnson:
> So "ITK_COMPUTER_MEMORY_SIZE" is now gone? Bot maintainers like me should therefore remove it from their scripts right?

Not gone.  Just automatically set in the cache if not already set.  See line 418.

Your scripts will work the same way as always, and you can always override the default CACHE setting.

PR #1355, opened 2019-10-25
COMP: Removed check for old __sync builtins, since they're no longer …

seanm:


seanm:
@thewtex @dzenanz @vfonov

PR #1356, opened 2019-10-25
COMP: Fixed the only -Wmissing-braces warning

seanm:


seanm:
@thewtex @dzenanz @vfonov

Issue #1115, opened 2019-07-22
Inconsistent _snprintf define 


KBaur:
### Description

Above VS2015 the 

> #define snprintf _snprintf

is no longer necessary and will cause problemens on compiling with windows.

### Steps to Reproduce
In
itkImageSeriesWriter.hxx line 32
libsrc2/minc2_pricate.h line 213
include/minc2_private.h line 213
libmin/libsrc2/minc2_private.h line 213

> #if _MSC_VER
> #define snprintf _snprintf
> #endif // _MSC_VER

change to

> #if _MSC_VER < 1900 /* MSVC-14 defines real snprintf()... just about time! */
> #define snprintf _snprintf
> #endif // _MSC_VER

### Actual behavior

```none
```

### Versions

ITK 5.0.0

### Environment
OS: Windows
Platform Toolset: MSVS 2017 Visual Studio 2017 (v141)
Windows SDK Version: 10.0.15063.0

dzenanz:
@gdevenyi I guess this should be changed in MINC, and then we should update ITK's version of MINC.

TMeindl:
I think the patch for libsrc2/minc2_private.h still needs 

#define strncasecmp strnicmp
#define strcasecmp stricmp

in the common case if _MSC_VER is defined.

See attached patch:
[ITK-5.0.0-snprintf.patch.txt](https://github.com/InsightSoftwareConsortium/ITK/files/3421780/ITK-5.0.0-snprintf.patch.txt)

seanm:
I made the same conditionalization in two places in ITK itself: https://github.com/InsightSoftwareConsortium/ITK/pull/1359

and in GDCM: https://github.com/malaterre/GDCM/pull/99

I believe this can be closed after those are in.

N-Dekker:
Please note that `snprintf` and `_snprintf` behave slightly different.

    char str1[4] = "";
    char str2[4] = "";

    snprintf(str1, 3, "1234567");  // str1 = "12"
    _snprintf(str2, 3, "1234567");  // str2 = "123"

As I observed with VS2019 (version 16.3.6)

seanm:
I think the only ones left are in:
- MINC's config.h.msvc-win32
- teem nrrd
- googletest

dzenanz:
Has this been resolved? Can we close it?

hjmjohnson:
```
git grep "define.*_snprintf"

[...]

PR #1359, opened 2019-10-25
COMP: Properly conditionalize _snprintf hack to only old MS compilers

seanm:
For github issue #1115.


dzenanz:
I agree with Niels: the old stuff can be completely removed, now that we require at least VS2015.

PR #1365, opened 2019-10-28
PERF: MeanImageFilter faster, using BufferedImageNeighborhoodPixelAccessPolicy

N-Dekker:
Improved the performance `MeanImageFilter`, using `ShapedImageNeighborhoodRange` with a newly added pixel access policy, `BufferedImageNeighborhoodPixelAccessPolicy`. Tested performance by:

    #include <itkMeanImageFilter.h>
    #include <itkImage.h>
    #include <chrono>
    #include <iostream>

    int main()
    {
      const auto size = ImageType::SizeType{ 1024, 1024, 16 };
      image->SetRegions(size);
      image->Allocate(true);

      const auto filter = itk::MeanImageFilter<ImageType, ImageType>::New();

      for (int i{}; i < 8; ++i)
      {
        filter->SetInput(nullptr);
        filter->SetInput(image);

        using namespace std::chrono;

        filter->Update();

        const auto durationSeconds =
        std::cout
          << std::endl;
      }
    }

### DIFF
--- a/Modules/Core/Common/include/itkBufferedImageNeighborhoodPixelAccessPolicy.h
+++ b/Modules/Core/Common/include/itkBufferedImageNeighborhoodPixelAccessPolicy.h
@@ -1,0 +1,137 @@
+/*=========================================================================
+ *
+ *  Copyright Insight Software Consortium
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+#ifndef itkBufferedImageNeighborhoodPixelAccessPolicy_h
+#define itkBufferedImageNeighborhoodPixelAccessPolicy_h
+
+#include "itkIndex.h"
+#include "itkOffset.h"
+#include "itkSize.h"
+
+#include <cassert>
+
+namespace itk
+{
+namespace Experimental
+{
+
+/**
+ * \class BufferedImageNeighborhoodPixelAccessPolicy
+ * ImageNeighborhoodPixelAccessPolicy class for ShapedImageNeighborhoodRange.
+ * Allows getting and setting the value of a pixel, located in a specified
+ * neighborhood location, at a specified offset. Assumes that the pixel index
+ * is within the buffered region.
+ *
+ * \note This policy offers faster pixel access than other pixel access policies,
+ * because it does not do any bounds checking or border extrapolation. An
+ * attempt to access pixels outside the buffered region will have undefined behavior.
+ *
+ * \author Niels Dekker, LKEB, Leiden University Medical Center
+ *
+ * \see ShapedNeighborhoodIterator
+ * \see ConstantBoundaryImageNeighborhoodPixelAccessPolicy
+ * \see ZeroFluxNeumannImageNeighborhoodPixelAccessPolicy
+ * \ingroup ImageIterators
+ * \ingroup ITKCommon
+ */
+template <typename TImage>
+class BufferedImageNeighborhoodPixelAccessPolicy final
+{
+private:
+  using NeighborhoodAccessorFunctorType = typename TImage::NeighborhoodAccessorFunctorType;
+  using PixelType = typename TImage::PixelType;
+  using InternalPixelType = typename TImage::InternalPixelType;
+
+  using ImageDimensionType = typename TImage::ImageDimensionType;
+  static constexpr ImageDimensionType ImageDimension = TImage::ImageDimension;
+
+  using IndexType = Index<ImageDimension>;
+  using OffsetType = Offset<ImageDimension>;
+  using ImageSizeType = Size<ImageDimension>;
+
+  // Index value to the image buffer, indexing the current pixel.
+  const IndexValueType m_PixelIndexValue;
+
+  // A reference to the accessor of the image.
+  const NeighborhoodAccessorFunctorType & m_NeighborhoodAccessor;
+
+
+  // Private helper function. Calculates and returns the index value of the
+  // current pixel within the image buffer.
+  static IndexValueType
+  CalculatePixelIndexValue(const ImageSizeType &
+#if !defined(NDEBUG)
+                             // Parameter that is only needed for debug assert.
+                             imageSize
+#endif
+                           ,
+                           const OffsetType & offsetTable,
+                           const IndexType &  pixelIndex) ITK_NOEXCEPT
+  {
+    IndexValueType result = 0;
+
+    for (ImageDimensionType i = 0; i < ImageDimension; ++i)
+    {
+      const auto pixelIndexValue = pixelIndex[i];
+      assert((pixelIndexValue >= 0) && (static_cast<SizeValueType>(pixelIndexValue) < imageSize[i]));
+      result += pixelIndexValue * offsetTable[i];
+    }
+    return result;
+  }
+
+public:
+  // Deleted member functions:
+  BufferedImageNeighborhoodPixelAccessPolicy() = delete;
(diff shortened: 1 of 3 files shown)


dzenanz:
Does median filter have similar implementation pattern? Can this approach be applied to it?

blowekamp:
> Does median filter have similar implementation pattern? Can this approach be applied to it?

The median filter should be exploiting coherence of a sorted neighborhood, so the method should be quite difference. Come to think of it the mean should be exploit a moving average as well to reduce repeated adjacent computation.

blowekamp:
As for the PR, are there any asserts of validations of the buffered region assumptions? or will be just get "undefined" behavior when the assumptions are not met?

N-Dekker:
Thank you both for your comments!

@dzenanz 
> Does median filter have similar implementation pattern? Can this approach be applied to it?

Yes it does 😄 I just did `MeanImageFilter` with this PR, because it's slightly easier than `MedianImageFilter`. I do think that if this approach also appears applicable on other places, some extra utility could be helpful to limit the amount of copy-and-paste boilerplate code.

@blowekamp 
> As for the PR, are there any asserts of validations of the buffered region assumptions? or will be just get "undefined" behavior when the assumptions are not met?

An attempt to access a location outside the buffered region, using the proposed new pixel access policy, will produce undefined behavior, indeed. The calculation of the index value into the image buffer does not do bounds checking:
https://github.com/InsightSoftwareConsortium/ITK/blob/582b4c9792839098dbac3740d8bb8f5f792e0d95/Modules/Core/Common/include/itkBufferedImageNeighborhoodPixelAccessPolicy.h#L78

That's really a fundamental property of this new pixel access policy. It should be as fast as possible. Within `MeanImageFilter`, it is essential that it is _only_ being used with the first region produces by `ImageBoundaryFacesCalculator`, which is the center (non-boundary) region.

Some more profiling results from VS2017 Release:

| image size | duration before PR |  duration after PR |
| :--- | :--- | :--- |
| 256x256x256 | 0.26 sec | 0.15 sec |
| 16x1024x1024 | 0.49 sec | 0.16 sec |
| 1024x1024x16 | 1.4 (!!!) sec | 0.2 sec |

Could you please check if you can observe similar performance improvements from this PR?

blowekamp:
I hate always being the nay sayer. But the checks for these assumptions (at least in debug mode) is very important. I have spent a lot of time debugging filters and pipeline issues. With the removal of these safety checks it will make development and maintenance even more difficult and time consuming than it is.

blowekamp:
I would suggest looking at the ConvolutionImageFilter next. There is easily room for 10x-100x performance improvement there.

dzenanz:
`MeanImageFilter` being slightly easier than `MedianImageFilter` was my guess - I am glad to hear I was right 😄 

> But the checks for these assumptions (at least in debug mode)

`itkAssertInDebugAndIgnoreInReleaseMacro` or a regular C++ `assert` should work well. Of course, it will make debug mode somewhat slower, but it will be much easier to spot an error when it is introduced.

With included test, before: 0.3 sec, with this PR: 0.15 sec. 2x speedup, nice!

N-Dekker:
Thanks for your comments @blowekamp 

> There are faster implementations of the Mean and Median filters in other implementations in ITK.

Which  faster implementation of the Mean filter do you mean?

Is the current (v5.1b01) `itk::MeanImageFilter` still being used, or is it outdated? It looks like my PR would make it twice as fast. Do you think that's worth the effort?

blowekamp:
> Which faster implementation of the Mean filter do you mean?

The BoxMeanImageFilter performs better for higher dimension images and larger radius as it uses a moving accumulation. This implementation likely performs better for 2-D and smaller radius.

> Is the current (v5.1b01) `itk::MeanImageFilter` still being used, or is it outdated? It looks like my PR would make it twice as fast. Do you think that's worth the effort?

ITK is not about THE implementation for an algorithm. Many algorithms have multiple implementations which perform better in different situations. Others are frequent exemplar of how to do common operations in a generalized templated way. This filter I consider an exemplar filter for doing neighborhood operations, as well as giving good performance for small radius. It is most certainly worth the effort and is a great contribution.

Also do you know if these iterators work well with VectorImages? This filter did not work with them before.

N-Dekker:
@dzenanz 
> `itkAssertInDebugAndIgnoreInReleaseMacro` or a regular C++ `assert` should work well.

Just added  an `assert`, please check:
https://github.com/InsightSoftwareConsortium/ITK/blob/16c31cad9fdf628b2cdcbde0761c39e8e55d7a2c/Modules/Core/Common/include/itkBufferedImageNeighborhoodPixelAccessPolicy.h#L85

N-Dekker:
> Looks good, but CI has qualms with it.

Thanks @dzenanz I see at https://open.cdash.org/viewBuildError.php?type=1&buildid=6181856 
> itkBufferedImageNeighborhoodPixelAccessPolicy.h:76:50: warning: unused parameter 'imageSize' [-Wunused-parameter]

Even though I did `itkNotUsed(imageSize)`! I guess I have _not used_ `itkNotUsed` correctly  😺  Instead, I'll force-push an amend to just do a silly `void`-cast, `(void)imageSize;`, to silence the warning... OK?

dzenanz:
What happens if you use `itkAssertInDebugAndIgnoreInReleaseMacro` instead of `assert`? If that doesn't help, you could apply this trick:
https://github.com/InsightSoftwareConsortium/ITK/blob/49671655c88612fabeb5f0784d3f40123100f166/Modules/Core/Common/include/itkImageSource.hxx#L247-L250

N-Dekker:
@blowekamp 
> The BoxMeanImageFilter performs better for higher dimension images and larger radius as it uses a moving accumulation. This implementation likely performs better for 2-D and smaller radius.

Indeed: I just observed that with the "default" radius value (1), `MeanImageFilter` (with my proposed PR) appears significantly faster than `BoxMeanImageFilter`, even for 3-D... For a larger radius, I get mixed results: sometimes one is faster, sometimes the other.

The output of `BoxMeanImageFilter` is also slightly different than `MeanImageFilter`. It appears that `BoxMeanImageFilter` produces different values for border pixels. Is that a known issue?

> Also do you know if these iterators work well with VectorImages? This filter did not work with them before.

The iterator ranges from `Experimental` should work well with `VectorImage`. Looks like the incompatibility is caused by `sum = NumericTraits<InputRealType>::ZeroValue()`, in the filter itself: https://github.com/InsightSoftwareConsortium/ITK/blob/v5.1b01/Modules/Filtering/Smoothing/include/itkMeanImageFilter.hxx#L73

With this PR, I did not want to change that logic, because this PR is intended to just improve the performance of existing use cases. If you want to add `VectorImage` support to this filter, can you please make it a separate PR? Of course, I would like it if my PR (#1365) could be handled first  😄

N-Dekker:
Succeeded on ITK.Linux, ITK.Windows, and ITK.macOS 🎉 

The ITK.Linux.Python failures at the dashboard seem unrelated to this PR: 

> AttributeError: module 'ITKMathematicalMorphology' has no attribute 'swig'

https://open.cdash.org/testDetails.php?test=811772611&build=6183942

Modules/Core/Common/include/itkBufferedImageNeighborhoodPixelAccessPolicy.h
```diff
@@ -0,0 +1,137 @@
...
+                           ,
+                           const OffsetType & offsetTable,
+                           const IndexType &  pixelIndex) ITK_NOEXCEPT
+  {
+    IndexValueType result = 0;
+
+    for (ImageDimensionType i = 0; i < ImageDimension; ++i)
+    {
+      const auto pixelIndexValue = pixelIndex[i];
+      assert((pixelIndexValue >= 0) && (static_cast<SizeValueType>(pixelIndexValue) < imageSize[i]));
```
dzenanz: This assertion gets triggered 😄
https://open.cdash.org/testDetails.php?test=819819335&build=6195188

dzenanz:
The assertion also gets triggered [here](https://open.cdash.org/testDetails.php?test=819819909&build=6195188), by a test you wrote @N-Dekker. So much about relying on people being careful, or however you phrased it Niels 😄

N-Dekker:
> The assertion also gets triggered [here](https://open.cdash.org/testDetails.php?test=819819909&build=6195188), by a test you wrote @N-Dekker.

Thanks @dzenanz. I think it's related to issue #1397 (Valgrind Defects related to the MedianImageFilter), which should be fixed by pull request #1398 (BUG: Fix IndexRange for itk::Size that has one or more zero values)

PR #1378, opened 2019-10-31
ENH: Preserving information about auto-flipped negative steps in metadata

vfonov:
We need information weather the image was flipped to compensate for the negative steps.
This fix can also help resolve the issues menioned in https://github.com/InsightSoftwareConsortium/ITK/commit/7501479a970694b0dd4a8c4bbf7cbcc033fe059c

## PR Checklist

### DIFF
--- a/Modules/IO/ImageBase/include/itkImageFileReader.hxx
+++ b/Modules/IO/ImageBase/include/itkImageFileReader.hxx
@@ -24,6 +24,7 @@
 #include "itkConvertPixelBuffer.h"
 #include "itkPixelTraits.h"
 #include "itkVectorImage.h"
+#include "itkMetaDataObject.h"
 
 #include "itksys/SystemTools.hxx"
 #include <memory> // For unique_ptr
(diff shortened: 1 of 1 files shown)


vfonov:
CC @seanm

blowekamp:
Would just storing the original spacing in the MetaDataDictionary as an Array, be more useful and easyier? These string tags seems a little cumbersome to manage.

seanm:
@vfonov I think I agree with @blowekamp. If we change your patch to just store the original spacing, then apps like ours can consult it to know if ITK flipped things.  Since ITK also flips `direction`, maybe we should store the original of that too.

seanm:
@blowekamp what do you think of @vfonov 's newest patch? It LGTM.

vfonov:
Oops, I think my last commit didn't include needed changes

blowekamp:
That looks a lot simpler and easier to use! Thanks!

couple more questions:
- should something similar  also be added to the ImageSeries reader?
- Do you see an easy way to add a little test?

PR #1382, opened 2019-11-01
ENH: Separate boundary/non-boundary regions ImageBoundaryFacesCalculator

N-Dekker:
`NeighborhoodAlgorithm::ImageBoundaryFacesCalculator::operator()` places
the non-boundary region (at the center) in a list together with all the
boundary regions (the faces) that it has found. It appears clearer to
keep the non-boundary region and the boundary regions separate.

This commit adds a `Compute(image, region, radius)` member function
and a result type `Result` to `ImageBoundaryFacesCalculator`, which
keep them separate.

It replaces the old `operator()` call by a call to the new `Compute`, in
`MeanImageFilter`, to shorten the code and improve readability.

`operator()` is now implemented by calling `Execute`. `operator()` still
returns the list that has all regions together, as it did before.

### DIFF
--- a/Modules/Core/Common/include/itkNeighborhoodAlgorithm.h
+++ b/Modules/Core/Common/include/itkNeighborhoodAlgorithm.h
@@ -69,6 +69,55 @@ struct ImageBoundaryFacesCalculator
   using FaceListType = std::list<RegionType>;
   static constexpr unsigned int ImageDimension = TImage::ImageDimension;
 
+  /** \class Result
+   *
+   * Represents the result of ImageBoundaryFacesCalculator::Compute
+   *
+   * \ingroup ITKCommon
+   */
+  class Result
+  {
+  public:
+    /** Returns the center (non-boundary) region. */
+    RegionType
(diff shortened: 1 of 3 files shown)


N-Dekker:
My latest force-push only did a rebase, to ensure that the new `MeanImageFilter` GTests that Matt @thewtex just merged onto the master still pass successfully.

blowekamp:
This looks like a good thing to add to me!

However, the name `Execute` does not seem right to me. I think ITK "calculator" classes are suppose to have a `Compute` method.

N-Dekker:
> This looks like a good thing to add to me!

Thanks, Bradley @blowekamp 

> However, the name `Execute` does not seem right to me. I think ITK "calculator" classes are suppose to have a `Compute` method.

Originally I thought of naming the function `Calculate`, because isn't that what calculators do? 😸  Then I thought, a calculation may be "executed". And `Execute` is quite a common term in ITK.

However, `Compute` is also fine to me. 😄  I'll prepare a force-push in 20 minutes from now!

PR #1384, opened 2019-11-02
ENH: Improve diagnostic warnings for ResampleImage misuse

hjmjohnson:
Using a "ReferenceImage" for the ResampleImageFilter is a two step
process requiring both 'SetReferenceImage(im)' and 'UseReferenceImageOn()'.

Failing to use both calls silently generates a resampled image of size
zero in all dimensions.


blowekamp:
The check looks reasonable.

Please consider implementing this check in an overloaded [`VerifyPreconditions`](https://itk.org/Doxygen/html/classitk_1_1ProcessObject.html#a73000f23064b7d2898fa21135080b4cf) method. This method is executed before input information is propagated through the pipeline. You should be able to check if the output size ivar is zero and reference image is set but not enable then throw.

blowekamp:
Great! Now the pipeline will fail earlier. Thanks! :thumbs up:

PR #1385, opened 2019-11-03
PERF: Use ShapedImageNeighborhoodRange in MedianImageFilter

N-Dekker:
Improved the performance of `MedianImageFilter::Update`
`Experimental::ShapedImageNeighborhoodRange` variables, using
`BufferedImageNeighborhoodPixelAccessPolicy` for the non-boundary
(center) subregion.

~25% reduction of the run-time duration was observed on a
`filter->Update()` call, using VS2017, processing a 256x256x256 image.
For a 1024x1024x16 image, the performance improvement appeared even higher.

### DIFF
--- a/Modules/Filtering/Smoothing/include/itkMedianImageFilter.hxx
+++ b/Modules/Filtering/Smoothing/include/itkMedianImageFilter.hxx
@@ -19,12 +19,13 @@
 #define itkMedianImageFilter_hxx
 #include "itkMedianImageFilter.h"
 
-#include "itkConstNeighborhoodIterator.h"
-#include "itkNeighborhoodInnerProduct.h"
-#include "itkImageRegionIterator.h"
+#include "itkBufferedImageNeighborhoodPixelAccessPolicy.h"
+#include "itkImageNeighborhoodOffsets.h"
+#include "itkImageRegionRange.h"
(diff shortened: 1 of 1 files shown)


N-Dekker:
It would be great if one of you could confirm that this commit does indeed yield a significant performance improvement. If you want to, you could run the following little program, both without and with the commit, and compare the duration:

    #include <itkMedianImageFilter.h>
    #include <itkImage.h>
    #include <chrono>
    #include <iostream>

    int main()
    {

      image->SetRegions(ImageType::SizeType{ {1024, 1024, 16} });
      image->Allocate(true);

      const auto filter = itk::MedianImageFilter<ImageType, ImageType>::New();

      for (int i{}; i < 8; ++i)
      {
        filter->SetInput(nullptr);
        filter->SetInput(image);

        using namespace std::chrono;

        filter->Update();

        const auto durationSeconds =
        std::cout
          << std::endl;
      }
    }

N-Dekker:
Oops, typo in the subject of the commit message!

N-Dekker:
Typo fixed ("Shared" --> "Shaped") 😸

dzenanz:
I was trying to find the typo, unsuccessfully :smile:

Issue #1394, opened 2019-11-07
Python Dilate & Erode filters are not working


hjmjohnson:
### Description
Using python to perform binary dilation and erosion operations are failing.

Following the basic approach as defined at 

https://itk.org/ITKExamples/src/Filtering/BinaryMathematicalMorphology/DilateABinaryImage/Documentation.html 

does not result in the dilation of the binary image.

### Steps to Reproduce

A unit test that demonstrates the problem.  The expectation is that a dilated image is different than the input image.

I have confirmed that the same behavior occurs in both 5.0.0 and 5.1.0 versions of the python bindings for ITK.

```python
# \author Hans J. Johnson
# Demonstrate that DilateFilter does not work
import itk

PixelType = itk.UC
Dimension = 2
ImageType = itk.Image[PixelType, Dimension]

im_orig = ImageType.New()
RegionType = itk.ImageRegion[Dimension]
r = RegionType()
r.SetSize([9, 9])
im_orig.SetRegions(r)
im_orig.Allocate(True)

IndexType = itk.Index[Dimension]
i = IndexType()
i.Fill(4)
im_orig.SetPixel(i, 1)

StructuringElementType = itk.FlatStructuringElement[Dimension]

if 0 == 1:
    FilterType = itk.BinaryDilateImageFilter[
        ImageType, ImageType, StructuringElementType
    ]
else:
    FilterType = itk.BinaryErodeImageFilter[
        ImageType, ImageType, StructuringElementType
    ]

structuringElement = StructuringElementType.Ball(2)
structuringElement.SetRadius(2)

# filter_to_test = FilterType.New()
filter_to_test = FilterType.New()
filter_to_test.SetInput(im_orig)
filter_to_test.SetKernel(structuringElement)
filter_to_test.Update()
im_output = filter_to_test.GetOutput()

array1 = itk.GetArrayFromImage(im_orig)
array2 = itk.GetArrayFromImage(im_output)

import numpy as np

if np.array_equal(array1, array2):
    print(
        f"""Error Values should be different after dilating by 2:
\nBefore Filtering\n{array1}
\nAfter Filtering\n{array2}"""
    )

print(itk.Version.GetITKVersion())
```
### Expected behavior

the dilated image should have more 1's than the original image.

### Actual behavior

the output image is identical to the original image.

```
Error Values should be different after dilating by 2:

Before Filtering
[[0 0 0 0 0 0 0 0 0]
[...]

PR #1395, opened 2019-11-07
DOC: More clearly describe foreground/background

hjmjohnson:
The generalalization of "binary" as '0' and 'not zero'
is not used in the "Binary[Dilate|Erode]ImageFilter.

Images with values of 0 or 1 require an explicit setting
of the DilateValue or ErodeValue to 1 in order to
designate 1 as the foreground value to be dilated/eroded.

Expand the documentation to more forcefully convey this
important consideration.

## PR Checklist

### DIFF
--- a/Modules/Filtering/BinaryMathematicalMorphology/include/itkBinaryErodeImageFilter.h
+++ b/Modules/Filtering/BinaryMathematicalMorphology/include/itkBinaryErodeImageFilter.h
@@ -27,10 +27,14 @@ namespace itk
 {
 /**
  * \class BinaryErodeImageFilter
- * \brief Fast binary erosion
+ * \brief Fast binary erosion of a single intensity value in the image.
  *
  * BinaryErodeImageFilter is a binary erosion
- * morphologic operation. This implementation is based on the papers:
+ * morphologic operation on the foreground of an image. Only the value designated
+ * by the single "DilateValue" is considered as foreground, and all other values
+ * are considered background.
+ *
+ * This implementation is based on the papers:
  *
  * L.Vincent "Morphological transformations of binary images with
  * arbitrary structuring elements", and
--- a/Modules/Filtering/BinaryMathematicalMorphology/include/itkBinaryErodeImageFilter.h
+++ b/Modules/Filtering/BinaryMathematicalMorphology/include/itkBinaryErodeImageFilter.h
@@ -31,8 +31,24 @@ namespace itk
  *
  * BinaryErodeImageFilter is a binary erosion
  * morphologic operation on the foreground of an image. Only the value designated
- * by the single "DilateValue" is considered as foreground, and all other values
- * are considered background.
+ * by the intensity value "SetForegroundValue()" (alias as SetErodeValue()) is considered
+ * as foreground, and other intensity values are considered background.
+ *
+ * Gray scale images can be processed as binary images by selecting a
+ * "ForegroundValue" (alias "ErodeValue").  Pixel values matching the erode value are
+ * considered the "foreground" and all other pixels are
+ * "background". This is useful in processing segmented images where
+ * all pixels in segment #1 have value 1 and pixels in segment #2 have
+ * value 2, etc. A particular "segment number" can be processed.
+ * ForegroundValue defaults to the maximum possible value of the
+ * PixelType. The eroded pixels will receive the BackgroundValue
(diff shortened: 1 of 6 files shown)


Modules/Filtering/BinaryMathematicalMorphology/include/itkBinaryErodeImageFilter.h
```diff
@@ -27,10 +27,14 @@ namespace itk
...
 /**
  * \class BinaryErodeImageFilter
- * \brief Fast binary erosion
+ * \brief Fast binary erosion of a single intensity value in the image.
  *
  * BinaryErodeImageFilter is a binary erosion
- * morphologic operation. This implementation is based on the papers:
+ * morphologic operation on the foreground of an image. Only the value designated
+ * by the single "DilateValue" is considered as foreground, and all other values
+ * are considered background.
```
blowekamp: What value are the eroded pixels replaced with?

dzenanz: I would assume zero.

hjmjohnson: m_BackgroundValue = NumericTraits<OutputPixelType>::NonpositiveMin(); 

and it can be overridden.

blowekamp:
For the erosion operation only the BackgroundValue matters, and for dilation only the ForegroundValue matters. It would be good if the updated documentation reflected this complementary nature of the operations instead of repeating the same text.

hjmjohnson:
> For the erosion operation only the BackgroundValue matters, and for dilation only the ForegroundValue matters. It would be good if the updated documentation reflected this complementary nature of the operations instead of repeating the same text.

@blowekamp The above statement is not strictly true.  I'm working on documentation to clarify the situation.

For dilate, "ForegroundValue" value is used to identify voxels that need changing & "ForegroundValue" is the value used for changed voxels.

For erode, "ForegroundValue" value is used to identify voxels that need changing & "BackgroundValue" is the value used for changed voxels.

blowekamp:
> > For the erosion operation only the BackgroundValue matters, and for dilation only the ForegroundValue matters. It would be good if the updated documentation reflected this complementary nature of the operations instead of repeating the same text.
> 
> @blowekamp The above statement is not strictly true. I'm working on documentation to clarify the situation.
> 
> For dilate, "ForegroundValue" value is used to identify voxels that need changing & "ForegroundValue" is the value used for changed voxels.
> 
> For erode, "ForegroundValue" value is used to identify voxels that need changing & "BackgroundValue" is the value used for changed voxels.

For dilate: If `any` pixels in neighborhood are FG then the value is FG
For erode: If `any` pixel is neighborhood is *not* FG then the value is BG

So these operations are *only* duals of each other when the image are binary.

Binary morphology with 3 classes is confusing :)

Modules/Filtering/BinaryMathematicalMorphology/include/itkBinaryErodeImageFilter.h
```diff
@@ -27,10 +27,14 @@ namespace itk
 {
 /**
  * \class BinaryErodeImageFilter
- * \brief Fast binary erosion
+ * \brief Fast binary erosion of a single intensity value in the image.
  *
  * BinaryErodeImageFilter is a binary erosion
- * morphologic operation. This implementation is based on the papers:
+ * morphologic operation on the foreground of an image. Only the value designated
+ * by the single "DilateValue" is considered as foreground, and all other values
```
thewtex: DilateValue -> ErodeValue

hjmjohnson:
> Binary morphology with 3 classes is confusing :)

YES!  I wasted 4 hours yesterday on the stupid assumption that binary meant "zero or not zero".  That is to say that I assumed that an unsigned char image with values of 1 representing the mask could be dilated without explicitly setting the DilateValue (which I also incorrectly assumed was the size of the dilation).  I was trying to infer the behavior of the filter based on python introspection results rather than reading deep into the doxygen pages.  

I went so far as to make a unit test and submit a bug report:   #1394
because I was convinced that there was a python bug in the wrapping.

I felt really stupid in the end, because this exact same problem was encountered by a student in my class last spring, and that student was really upset about my homework assignment.   As soon as I found DilateValue, that triggered my memory.

Issue #1397, opened 2019-11-07
Valgrind Defects related to the MedianImageFilter


blowekamp:
https://open.cdash.org/viewDynamicAnalysis.php?buildid=6196748

```
IPW ==8471== Invalid write of size 4
==8471==    at 0x173C9A: itk::MedianImageFilter, itk::Image >::DynamicThreadedGenerateData(itk::ImageRegion<2u> const&) (itkMedianImageFilter.hxx:98)
==8471==    by 0x155DF3: itk::ImageSource >::GenerateData()::{lambda(itk::ImageRegion<2u> const&)#1}::operator()(itk::ImageRegion<2u> const&) const (itkImageSource.hxx:232)
==8471==    by 0x1680AA: std::_Function_handler const&), itk::ImageSource >::GenerateData()::{lambda(itk::ImageRegion<2u> const&)#1}>::_M_invoke(std::_Any_data const&, itk::ImageRegion<2u> const&) (std_function.h:316)
==8471==    by 0x168294: std::function const&)>::operator()(itk::ImageRegion<2u> const&) const (std_function.h:706)
==8471==    by 0x1624E0: void itk::MultiThreaderBase::ParallelizeImageRegion<2u>(itk::ImageRegion<2u> const&, std::function const&)>, itk::ProcessObject*)::{lambda(long const*, unsigned long const*)#1}::operator()(long const*, unsigned long const*) const (itkMultiThreaderBase.h:300)
==8471==    by 0x16BF03: std::_Function_handler(itk::ImageRegion<2u> const&, std::function const&)>, itk::ProcessObject*)::{lambda(long const*, unsigned long const*)#1}>::_M_invoke(std::_Any_data const&, long const*&&, unsigned long const*&&) (std_function.h:316)
==8471==    by 0x50028E2: std::function::operator()(long const*, unsigned long const*) const (std_function.h:706)
==8471==    by 0x50631D4: itk::PoolMultiThreader::ParallelizeImageRegion(unsigned int, long const*, unsigned long const*, std::function, itk::ProcessObject*) (itkPoolMultiThreader.cxx:243)
==8471==    by 0x162607: void itk::MultiThreaderBase::ParallelizeImageRegion<2u>(itk::ImageRegion<2u> const&, std::function const&)>, itk::ProcessObject*) (itkMultiThreaderBase.h:289)
==8471==    by 0x155F24: itk::ImageSource >::GenerateData() (itkImageSource.hxx:229)
[...]

PR #1398, opened 2019-11-07
BUG: Fix IndexRange for itk::Size that has both zero's and non-zero's

N-Dekker:
`IndexRange` did behave incorrectly when it would be constructed with an
`itk::Size` that has one or more zero values _and_ a non-zero value at
its last position, for example `Size<3>{{0, 0, 8}}`. Specifically, in
that case the iterators returned by `IndexRange::begin()` and
`IndexRange::end()` were different, while they should compare equal.
This bug caused Valgrind defects, issue #1397, reported by Bradley
@blowekamp

The bug affects both `IndexRange` specializations,
`ImageRegionIndexRange` and `ZeroBasedIndexRange`.

This bug fix ensures that a `itk::Size` that has one or more zero values
is treated the same way as a `itk::Size` that has _only_ zero values.

Fixes issue #1397 Valgrind Defects related to the MedianImageFilter

### DIFF
--- a/Modules/Core/Common/include/itkIndexRange.h
+++ b/Modules/Core/Common/include/itkIndexRange.h
@@ -430,11 +430,25 @@ private:
   static IndexType
   CalculateMaxIndex(const MinIndexType & minIndex, const SizeType & size)
   {
+    const bool sizeHasZeroValue = [&size] {
+      for (const auto sizeValue : size)
+      {
+        if (sizeValue == 0)
+        {
+          return true;
+        }
+      }
+      return false;
+    }();
+
+    // Treat any size that has a zero value equally.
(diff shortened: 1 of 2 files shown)


N-Dekker:
The [ITK.Windows failure](https://open.cdash.org/viewBuildError.php?buildid=6197271) seems unrelated to this pull request

dzenanz:
Some compiler warnings to address, e.g. `warning: declaration of 'const SizeType& size' shadows a parameter [-Wshadow]`

N-Dekker:
FYI, My last force-push only adjusted the commit text. It did not include any more code changes. I hope the commit text is clear enough. 😄

PR #1401, opened 2019-11-07
ENH: update KWStyle to fix potential buffer overflow

dzenanz:
warning: '%ld' directive writing between 1 and 20 bytes into a region of size 10 [-Wformat-overflow=]
note: '__builtin___sprintf_chk' output between 2 and 21 bytes into a destination of size 10


dzenanz:
Python test failures are unrelated. Merging.

PR #1405, opened 2019-11-08
STYLE: MeanImageFilter use vector<Offset>, range-based-for on IndexRange

N-Dekker:
Replaced `std::unique_ptr<Offset>[]>` by `std::vector<Offset>`, storing
the rectangular neighborhood shape offsets of the filter.

Replaced old-style `for` by C++11 range-based for-loop on
`ImageRegionIndexRange`.

Removed unnecessary local `ImageRegionRange` variable.

This commit aims to improve code readability, while it does not seem to
make the filter any slower. Actually, the filter seems slightly faster
than before (as observed with VS2017 Release).

### DIFF
--- a/Modules/Filtering/Smoothing/include/itkMeanImageFilter.h
+++ b/Modules/Filtering/Smoothing/include/itkMeanImageFilter.h
@@ -22,6 +22,8 @@
 #include "itkImage.h"
 #include "itkNumericTraits.h"
 
+#include <vector>
+
 namespace itk
 {
 /** \class MeanImageFilter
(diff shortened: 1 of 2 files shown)


N-Dekker:
Thanks to you all for approving and merging my pull request!

@dzenanz wrote:
> Does `sum` variable retain its type?

The variable is now declared by `auto sum = NumericTraits<InputRealType>::ZeroValue()`:

https://github.com/InsightSoftwareConsortium/ITK/blob/70c8737c81a08b7d104a80236ded3770506a4039/Modules/Filtering/Smoothing/include/itkMeanImageFilter.hxx#L89

Before I got involved with `MeanImageFilter`, the variable was declared as `InputRealType sum` (without `auto`): https://github.com/InsightSoftwareConsortium/ITK/blob/v5.1b01/Modules/Filtering/Smoothing/include/itkMeanImageFilter.hxx#L59

Does that answer your question? 😃

PR #1409, opened 2019-11-11
STYLE: Fix small inconsistency between IndexRange constructors

N-Dekker:
There was an inconsistency between the constructors of `IndexRange`, in
the way they initialized the private `m_MaxIndex` member variable for an
empty range. For example, for `IndexRange<3, true>{ itk::Size<3>() }`,
m_MaxIndex is `{-1, -1, -1}`, whereas for a default-constructed
`IndexRange<3, true>()`, m_MaxIndex used to be `{0, 0, -1}`.

This commit fixes the inconsistency by ensuring that m_MaxIndex is
filled entirely by `-1` (for 3-D, that is `{-1, -1, -1}`), when the
range is default-constructed.

It also reduces the complexity of `IndexRange::empty()` from O(n) to
O(1), as it is sufficient now to just compare the first elements of
`m_MinIndex` and `m_MaxIndex`.

### DIFF
--- a/Modules/Core/Common/include/itkIndexRange.h
+++ b/Modules/Core/Common/include/itkIndexRange.h
@@ -289,19 +289,12 @@ public:
   using reverse_iterator = std::reverse_iterator<iterator>;
   using const_reverse_iterator = std::reverse_iterator<const_iterator>;
 
-  /** Default-constructor. Constructs an empty range.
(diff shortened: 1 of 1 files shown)


N-Dekker:
Honestly this is why I wanted to have `itk::Index::Filled(value)` in the first place 😸 I'm glad pull request #1404 was accepted so quickly!

PR #1412, opened 2019-11-12
STYLE: Add additional braces around initialization of subobject

Nekto89:
Fixes warning on AppleClang
https://open.cdash.org/viewBuildError.php?type=1&buildid=6201669


N-Dekker:
I see some failures from ITK.Linux.Python at https://open.cdash.org/viewTest.php?onlyfailed&buildid=6205647 but I guess these are unrelated to this pull request, right?

PR #1413, opened 2019-11-12
STYLE: Remove run-time estimation of operator radius in gradient filters

N-Dekker:
It appears unnecessary to evaluate `operator.GetRadius()[0]` for a
`DerivativeOperator` that has a default direction and a default order,
after calling `DerivativeOperator::CreateDirectional()`. The radius
value estimated that way will always be exactly one.

This information is used to simplify the implementation of
`GradientImageFilter` and `GradientMagnitudeImageFilter`.


N-Dekker:
> Nice code simplification. Why did you even look here? 😈

Thanks @dzenanz I was looking for classes in ITK that could possibly benefit from using the `Experimental` ranges, and then I encountered these gradient filters. However, before even trying to get them to use these ranges, I would need to get a better understanding of what they exactly do. Which is why I'm now trying to simplify them!

PR #1415, opened 2019-11-13
STYLE: Simplify Neighborhood::ComputeNeighborhoodStrideTable()

N-Dekker:
Simplified the implementation of the protected `Neighborhood` member
function `ComputeNeighborhoodStrideTable()`, removing an unnecessary
inner `for` loop and a local `stride` variable.


dzenanz:
And that happened long ago, in d4b070867f83f0c215e9feea871556b86252f7ca.

N-Dekker:
> And that happened long ago, in [d4b0708](https://github.com/InsightSoftwareConsortium/ITK/commit/d4b070867f83f0c215e9feea871556b86252f7ca).

Indeed, @dzenanz While my PR doesn't do any modern C++. It should still be compatible with the programming style of that long ago 😄

PR #1417, opened 2019-11-13
STYLE: C++ Rule of Zero for Neighborhood Operators, support nothrow move

N-Dekker:
Adjusted `NeighborhoodOperator` and eight derived classes according to
the C++ Rule of Zero for special member functions, by removing
used-declared default-constructors, copy-constructors, and assignment
operators.

Implicitly fixed an old bug in the assignment of `LaplacianOperator`, as
it previously did not copy its `m_DerivativeScalings` data member.

This commit effectively makes move-construction and move-assignment of
those classes "noexcept", as is checked in newly added static
assertions, in itkNeighborhoodOperatorTest. It could also improves the
performance of a "move" of such a type of object.

### DIFF
--- a/Modules/Core/Common/include/itkLaplacianOperator.h
+++ b/Modules/Core/Common/include/itkLaplacianOperator.h
@@ -50,6 +50,10 @@ namespace itk
  *  of doubles that is of length VDimension (the dimensionality of the image).
  *  Make sure to use 1/pixel_spacing to properly scale derivatives.
  *
+ * \note LaplacianOperator does not have any user-declared "special member function"
+ * for copy, move, or destruction, following the C++ Rule of Zero: the compiler will
+ * generate them if necessary.
+ *
  * \sa NeighborhoodOperator
  * \sa Neighborhood
  * \ingroup Operators
@@ -81,27 +85,9 @@ public:
     }
   }
 
-  /** Copy constructor   */
-  LaplacianOperator(const Self & other)
-    : NeighborhoodOperator<TPixel, VDimension, TAllocator>(other)
-  {
-    for (unsigned i = 0; i < VDimension; ++i)
-    {
-      m_DerivativeScalings[i] = other.m_DerivativeScalings[i];
-    }
-  }
-
   /** This function is called to create the operator  */
   void
   CreateOperator();
-
-  /** Assignment operator   */
-  Self &
-  operator=(const Self & other)
-  {
-    Superclass::operator=(other);
-    return *this;
-  }
 
   /** Prints some debugging information   */
   void
--- a/Modules/Core/Common/test/itkNeighborhoodOperatorTest.cxx
+++ b/Modules/Core/Common/test/itkNeighborhoodOperatorTest.cxx
@@ -16,14 +16,54 @@
  *
  *=========================================================================*/
 
+#include "itkAnnulusOperator.h"
+#include "itkBackwardDifferenceOperator.h"
 #include "itkDerivativeOperator.h"
 #include "itkForwardDifferenceOperator.h"
-#include "itkBackwardDifferenceOperator.h"
+#include "itkGaussianDerivativeOperator.h"
 #include "itkGaussianOperator.h"
+#include "itkImageKernelOperator.h"
 #include "itkLaplacianOperator.h"
+#include "itkNeighborhoodOperator.h"
 #include "itkSobelOperator.h"
+
+#include <type_traits> // For is_default_constructible, is_copy_constructible, etc.
+
 namespace
 {
+template <typename T>
+constexpr bool
+IsDefaultConstructibleCopyableNoThrowMovableAndDestructible()
+{
+  return std::is_default_constructible<T>::value && std::is_copy_constructible<T>::value &&
+         std::is_copy_assignable<T>::value && std::is_nothrow_move_constructible<T>::value &&
+         std::is_nothrow_move_assignable<T>::value && std::is_nothrow_destructible<T>::value;
+}
+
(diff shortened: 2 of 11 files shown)


Modules/Core/Common/include/itkLaplacianOperator.h
```diff
@@ -81,28 +85,10 @@ class ITK_TEMPLATE_EXPORT LaplacianOperator : public NeighborhoodOperator<TPixel
...
   /** This function is called to create the operator  */
   void
   CreateOperator();
 
-  /** Assignment operator   */
-  Self &
-  operator=(const Self & other)
-  {
-    Superclass::operator=(other);
-    return *this;
```
N-Dekker: Looks like the old assignment of `LaplacianOperator` had a bug, as it forgot to copy `m_DerivativeScalings`!  😨

Modules/Core/Common/test/itkNeighborhoodOperatorTest.cxx
```diff
@@ -16,14 +16,52 @@
...
+constexpr bool
+IsDefaultConstructibleCopyableNoThrowMovableAndDestructible()
+{
+  return std::is_default_constructible<T>::value && std::is_copy_constructible<T>::value &&
+         std::is_copy_assignable<T>::value && std::is_nothrow_move_constructible<T>::value &&
+         std::is_nothrow_move_assignable<T>::value && std::is_nothrow_destructible<T>::value;
+}
+
+static_assert(std::is_copy_assignable<itk::NeighborhoodOperator<int, 3>>::value,
+              "NeighborhoodOperator be copy-assignable.");
```
dzenanz: be -> should be

Modules/Core/Common/test/itkNeighborhoodOperatorTest.cxx
```diff
@@ -16,14 +16,52 @@
...
+static_assert(std::is_nothrow_move_assignable<itk::NeighborhoodOperator<int, 3>>::value,
+              "NeighborhoodOperator be noexcept move-assignable.");
```
dzenanz: be -> should be

Modules/Core/Common/test/itkNeighborhoodOperatorTest.cxx
```diff
@@ -16,14 +16,52 @@
...
+
+static_assert(IsDefaultConstructibleCopyableNoThrowMovableAndDestructible<itk::AnnulusOperator<int>>(),
+              "AnnulusOperator should be default-constructible, copyable, noexcept movable, and destructible.");
+static_assert(IsDefaultConstructibleCopyableNoThrowMovableAndDestructible<itk::DerivativeOperator<int>>(),
+              "DerivativeOperator should be default-constructible, copyable, noexcept movable, and destructible.");
+static_assert(
```
dzenanz: `BackwardDifferenceOperator` is missing from this list of asserts

N-Dekker: > `BackwardDifferenceOperator` is missing from this list of asserts

Very well observed, @dzenanz ! I skipped that one, because it does `ITK_DISALLOW_COPY_AND_ASSIGN(BackwardDifferenceOperator)`! https://github.com/InsightSoftwareConsortium/ITK/blob/v5.1b01/Modules/Core/Common/include/itkBackwardDifferenceOperator.h#L48

The "Rule of Zero" would allow copy-construction and assignment, for `BackwardDifferenceOperator`. It seems like that was not intended. I don't know why specifically `BackwardDifferenceOperator` does not allow copying.

The other eight `NeighborhoodOperator` derived classes already allowed copying before, and they still do with this PR. (Only with less code, and added nothrow-movability).

Modules/Core/Common/test/itkNeighborhoodOperatorTest.cxx
N-Dekker: It _should be_ "should be", certainly 😸

Modules/Core/Common/test/itkNeighborhoodOperatorTest.cxx
dzenanz: You could add a new commit, to this PR or a follow-up PR, which removes `ITK_DISALLOW_COPY_AND_ASSIGN` from _BackwardDifferenceOperator_, in order to make things uniform. @hjmjohnson and @blowekamp what do you think?

N-Dekker: I leave it up to you @dzenanz @hjmjohnson @blowekamp whether or not you want this PR to remove `ITK_DISALLOW_COPY_AND_ASSIGN(BackwardDifferenceOperator)`. Please let me know 😃

Modules/Core/Common/test/itkNeighborhoodOperatorTest.cxx
N-Dekker: should be fixed now  😸

Modules/Core/Common/include/itkLaplacianOperator.h
N-Dekker: For the record, the bug (missing assignment of `LaplacianOperator::m_DerivativeScalings`) appears to be introduced with revision 670af6e15e7f416cdd75cf8f61748fd3cda8a485, May 13, 2003, "ENH: Operator now takes image spacing into account", which added the data member without extending the `operator=`.

Modules/Core/Common/test/itkNeighborhoodOperatorTest.cxx
dzenanz: Let's merge this now, and fix `BackwardDifferenceOperator` in another PR.

PR #1418, opened 2019-11-15
ENH: ImageIO check format support in two phases

### LINKED ISSUE #1410: ImageIO check format support in two phases
### Description

The way the IO factory works is that for every ImageIO it is constructed and then "CanRead" is called to determine if it's the correct one to use. If every ImageIO open the file, checks, and closes the file it will slow down this process ( consider network or remote files ).

### Expected behavior

There have been recent improvements for the ImageIOs to report the extensions it can read. To pursue this feature further I'd recommend making the IO Factory construction two passes:

1. Check the extension of the file is supported for each IO, if found then use that ImageIO.
1. Then check the CanRead method which may open the file.

### Additional Information

Suggested by @blowekamp in #1407.

Nekto89:
Closes #1410. HasSupportedReadExtension/HasSupportedWriteExtension in ImageIOBase are now public.

### DIFF
--- a/Modules/IO/ImageBase/src/itkImageIOFactory.cxx
+++ b/Modules/IO/ImageBase/src/itkImageIOFactory.cxx
@@ -19,49 +19,99 @@
 #include "itkImageIOFactory.h"
 
 #include <mutex>
-
+#include <vector>
 
 namespace itk
 {
 
 namespace
 {
-std::mutex createImageIOLock;
+std::vector<ImageIOBase::Pointer>
+GetAllImageIOInstances()
+{
+  static std::mutex           createImageIOLock;
+  std::lock_guard<std::mutex> mutexHolder(createImageIOLock);
+
+  const auto                        allInstances = ObjectFactoryBase::CreateAllInstance("itkImageIOBase");
+  std::vector<ImageIOBase::Pointer> result;
+  result.reserve(allInstances.size());
+  for (auto & possibleImageIO : allInstances)
+  {
+    auto * io = dynamic_cast<ImageIOBase *>(possibleImageIO.GetPointer());
+    if (io)
+    {
+      result.emplace_back(io);
+    }
+    else
+    {
+      std::cerr << "Error ImageIO factory did not return an ImageIOBase: " << possibleImageIO->GetNameOfClass() << '\n';
+    }
+  }
+  return result;
 }
+} // namespace
 
 ImageIOBase::Pointer
 ImageIOFactory::CreateImageIO(const char * path, FileModeType mode)
 {
-  std::list<ImageIOBase::Pointer> possibleImageIO;
-
-  std::lock_guard<std::mutex> mutexHolder(createImageIOLock);
-
-  for (auto & allobject : ObjectFactoryBase::CreateAllInstance("itkImageIOBase"))
+  if (mode != FileModeType::ReadMode && mode != FileModeType::WriteMode)
   {
-    auto * io = dynamic_cast<ImageIOBase *>(allobject.GetPointer());
-    if (io)
-    {
-      possibleImageIO.emplace_back(io);
-    }
-    else
-    {
-      std::cerr << "Error ImageIO factory did not return an ImageIOBase: " << allobject->GetNameOfClass() << std::endl;
-    }
+    return nullptr;
   }
-  for (auto & k : possibleImageIO)
+
+  auto       imageIOInstances = GetAllImageIOInstances();
+  const bool ignoreCase = true;
+  // check instances that support file extension
+  for (auto & imageIO : imageIOInstances)
   {
     if (mode == FileModeType::ReadMode)
     {
-      if (k->CanReadFile(path))
+      if (imageIO->HasSupportedReadExtension(path, ignoreCase))
       {
-        return k;
+        if (imageIO->CanReadFile(path))
+        {
+          return imageIO;
+        }
+        else
+        {
+          imageIO = nullptr; // don't check it again
+        }
       }
     }
-    else if (mode == FileModeType::WriteMode)
+    else
     {
-      if (k->CanWriteFile(path))
+      if (imageIO->HasSupportedWriteExtension(path, ignoreCase))
       {
-        return k;
+        if (imageIO->CanWriteFile(path))
+        {
+          return imageIO;
+        }
+        else
+        {
+          imageIO = nullptr;
+        }
+      }
+    }
+  }
+
+  for (auto & imageIO : imageIOInstances)
+  {
+    if (!imageIO)
+    {
+      continue;
+    }
+    if (mode == FileModeType::ReadMode)
+    {
+      if (imageIO->CanReadFile(path))
+      {
+        return imageIO;
+      }
+    }
+    else
+    {
+      if (imageIO->CanWriteFile(path))
+      {
+        return imageIO;
       }
     }
   }
--- a/Modules/IO/ImageBase/include/itkImageIOBase.h
+++ b/Modules/IO/ImageBase/include/itkImageIOBase.h
@@ -663,6 +663,14 @@ public:
     this->SetComponentType(ImageIOBase::LONG);
   }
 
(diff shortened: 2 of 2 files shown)


blowekamp:
This looks like a very good start.

While I did suggest this change, I do have some reservations about the patch changing the behavior of which ImageIO is automatically selected. Another thing to know is that the order of the ImageIO used to be well defined and ordered by priority. One case where order of the ImageIO  is import is when ITK is compiled with SCIFIO, SCIFIO will take reading TIFF files, some times for the better many times not. However, TIFFImageIO specifies is can read '*.tiff', but SCIFIO does not specify any extensions. With this new code TIFFImageIO will always automatically be used. There are other format that SCIFIO can read NRRD, where I don't know which ImageIO is used as it works and I didn't bother.

So this will change behavior in additional to improving performance. So this will need to be clear documented, and it should be discussed if this merge should be delayed until after 5.1 release to ensure the community has time to evaluate its impact.

Modules/IO/ImageBase/src/itkImageIOFactory.cxx
```diff
@@ -19,53 +19,103 @@
...
-      if (k->CanWriteFile(path))
+      if (imageIO->CanWriteFile(path))
       {
-        return k;
+        return imageIO;
       }
     }
   }
-  return nullptr;
+  return {};
```
dzenanz: I haven't seen `return {};` syntax before, so it is a bit confusing. I think that `return nullptr;` is clearer.

dzenanz:
@blowekamp another possibility is for SCIFIO to declare supported extensions. @thewtex @ctrueden is this feasible?

blowekamp:
> @blowekamp another possibility is for SCIFIO to declare supported extensions. @thewtex @ctrueden is this feasible?

SCFIO is a general reader that can support many formats. It should be lower priority than the native ITK ImageIO's such as NrrdImageIO and likely similar with PNG and many other formats. So it not supporting the extension check is a good thing!

TIFF is quite a specialized case with one file format with many different standards embedded in it like OME TIFF and other microscopy variants. It's always a bit of a surprise which ImageIO is used, so I have been explicitly specifying it for writing.

hjmjohnson:
> This looks like a very good start.
> 
> While I did suggest this change, I do have some reservations about the patch changing the behavior of which ImageIO is automatically selected. Another thing to know is that the order of the ImageIO used to be well defined and ordered by priority. One case where order of the ImageIO is import is when ITK is compiled with SCIFIO, SCIFIO will take reading TIFF files, some times for the better many times not. However, TIFFImageIO specifies is can read '*.tiff', but SCIFIO does not specify any extensions. With this new code TIFFImageIO will always automatically be used. There are other format that SCIFIO can read NRRD, where I don't know which ImageIO is used as it works and I didn't bother.
> 
> So this will change behavior in additional to improving performance. So this will need to be clear documented, and it should be discussed if this merge should be delayed until after 5.1 release to ensure the community has time to evaluate its impact.

I agree with @blowekamp that there is a huge risk of changing behavior in subtle ways that will be very disruptive to the community.  The default ordering of image IO is depended upon in many cases, and has never been rigorously defined or enforced.

I know that my next suggestion is a fair amount of work, but I think that a prioritization value is needed for each IO mechanism.  Each IOObject should have a 2 prioritization values added:  
```c++
   unsigned char m_InspectFileContentsCanReadPriority={255};
   unsigned char m_FileExtentionCanReadPriority={0};
```
Then the order of checking filters is done based on priority (with any filters that have the same priority be indeterminant which is read first.  The ordering of filters ONLY considers the values of these priorities when they are registered with the factory mechanism (i.e. changing the IO value requires re-registering with the factory to take effect).

blowekamp:
> I know that my next suggestion is a fair amount of work, but I think that a prioritization value is needed for each IO mechanism. Each IOObject should have a 2 prioritization values added:
> 
> ```c++
>    unsigned char m_InspectFileContentsCanReadPriority={255};
>    unsigned char m_FileExtentionCanReadPriority={0};
> ```
> 
> Then the order of checking filters is done based on priority (with any filters that have the same priority be indeterminant which is read first. The ordering of filters ONLY considers the values of these priorities when they are registered with the factory mechanism (i.e. changing the IO value requires re-registering with the factory to take effect).

I have suggested similar things before, including adding a priority value be include in the itk-module "FACTOR_NAMES" parameter e.g.: https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/IO/TIFF/itk-module.cmake#L12

@Nekto89 This is a can of worms that has been opened, with no easy answer to address properly.

blowekamp:
For reference here are ImageIO's registered in SimpleITK sorted alphabetically and their extensions:
```
  Registered ImageIO:
	BMPImageIO ( *.bmp, *.BMP )
	BioRadImageIO ( *.PIC, *.pic )
	Bruker2dseqImageIO
	GDCMImageIO ( *.dcm, *.DCM, *.dicom, *.DICOM )
	GE4ImageIO
	GE5ImageIO
	GiplImageIO
	HDF5ImageIO ( *.hdf, *.h4, *.hdf4, *.h5, *.hdf5, *.he4, *.he5, *.hd5 )
	JPEGImageIO ( *.jpg, *.JPG, *.jpeg, *.JPEG )
	JPEG2000ImageIO ( *.j2k, *.jp2, *.jpt )
	LSMImageIO ( *.tif, *.TIF, *.tiff, *.TIFF, *.lsm, *.LSM )
	MINCImageIO ( *.mnc, *.MNC, *.mnc2, *.MNC2 )
	MRCImageIO ( *.mrc, *.rec )
	MetaImageIO ( *.mha, *.mhd )
	NiftiImageIO ( *.nia, *.nii, *.nii.gz, *.hdr, *.img, *.img.gz )
	NrrdImageIO ( *.nrrd, *.nhdr )
	PNGImageIO ( *.png, *.PNG )
	StimulateImageIO ( *.spr )
	TIFFImageIO ( *.tif, *.TIF, *.tiff, *.TIFF )
	VTKImageIO ( *.vtk )

```

Modules/IO/ImageBase/src/itkImageIOFactory.cxx
```diff
@@ -19,53 +19,103 @@
...
 
 ImageIOBase::Pointer
 ImageIOFactory::CreateImageIO(const char * path, FileModeType mode)
 {
-  std::list<ImageIOBase::Pointer> possibleImageIO;
-
-  std::lock_guard<std::mutex> mutexHolder(createImageIOLock);
+  if (mode != FileModeType::ReadMode && mode != FileModeType::WriteMode)
+  {
+    return {};
```
thewtex: `return nullptr`

blowekamp:
The LSMImageIO extensions look odd, so upon inspection I created a patch (#1428) to remove the listed tiff extension. Likely, the other ImageIO's should have a review too.

Nekto89:
> This does slightly change undocumented behavior, but it is an improvement.
We can probably override "HasSupportedReadExtension" for ImageIO objects that might have conflicts and don't have their own extensions. Then the behavior would be the same.

I agree that system should be rewritten with some priorities, but can we do this in small steps?

blowekamp:
> I agree that system should be rewritten with some priorities, but can we do this in small steps?

Yes, with the proper schedule and plan. IMHO we should avoid frequent changes in behavior at releases. This change will have some subtle changes with which ImageIOs are loaded. And we would like to make additional changes. However, we are schedule to release v5.1b02 in a couple days and [v5.1 on December 9th](https://github.com/InsightSoftwareConsortium/ITK/milestone/6). IMHO we should avoid ImageIO selection changing behavior in 5.1 AND 5.2.

Hence, my conclusion in my first post that it may be best to wait until after the 5.1 release.

blowekamp:
I would like to take on updating this PR to get it in for 5.3.

I will also be adding a (global?) flag for this new behavior to the ImageIOBase class. The determination later can be made for what the default is for what legacy/compatibility CMake options.

blowekamp:
I am unable to push to the Nekto89's repository. I have created a new PR: #2511

PR #1419, opened 2019-11-15
STYLE: Move #include "itk_H5Cpp.h" from itkHDF5TransformIO.h to cxx file

N-Dekker:
Forward-declared `H5::PredType`, and moved the `#include "itk_H5Cpp.h"`
from itkHDF5TransformIO.h to the corresponding cxx file.

Removed `ITKHDF5` from its itk_module `TEST_DEPENDS`, as suggested by
Bradley @blowekamp

Could reduce compilation time and header file dependencies.

### DIFF
--- a/Modules/IO/TransformHDF5/include/itkHDF5TransformIO.h
+++ b/Modules/IO/TransformHDF5/include/itkHDF5TransformIO.h
@@ -23,7 +23,6 @@
 #include "itkTransformIOBase.h"
 #include <memory>
 #include <string>
-#include "itk_H5Cpp.h"
 
 // Avoids KWStyle error from forward declaration below.
 namespace itk
@@ -33,7 +32,8 @@ namespace itk
 namespace H5
 {
 class H5File;
-}
+class PredType;
+} // namespace H5
 
 namespace itk
 {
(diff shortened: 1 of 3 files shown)


blowekamp:
Also this should make the HDF5 library a only a private depends. Please update the itk_module.cmake file: https://github.com/InsightSoftwareConsortium/ITK/blob/725689ba2aec77a7d379517dd559fe8399e2c007/Modules/IO/TransformHDF5/itk-module.cmake#L12

This will have a very nice impact of reducing include paths when using this module.

N-Dekker:
> Also this should make the HDF5 library a only a private depends. Please update the itk_module.cmake file:
> 
> https://github.com/InsightSoftwareConsortium/ITK/blob/725689ba2aec77a7d379517dd559fe8399e2c007/Modules/IO/TransformHDF5/itk-module.cmake#L12

Thanks for the suggestion, @blowekamp  I'll remove `ITKHDF5` from `TEST_DEPENDS`. But then I wonder, why was it not a "public" `DEPENDS`, originally?

blowekamp:
> Thanks for the suggestion, @blowekamp I'll remove `ITKHDF5` from `TEST_DEPENDS`. But then I wonder, why was it not a "public" `DEPENDS`, originally?

It should have been in public before. Likely the only place this module was directly used was in the test, hence the dependency was just added there.

PR #1424, opened 2019-11-18
PERF: Do not allocate buffer in HDF5TransformIOTemplate::WriteParameters

N-Dekker:
This commit avoids copying `parameters` to a local dynamically allocated
buffer, in `WriteParameters` and `WriteFixedParameters`. This could
improve the performance, as well as exception safety.

It uses `std::unique_ptr`, instead of a raw pointer, in `ReadParameters`
and `ReadFixedParameters`, just to improve exception safety.

This commit fixes six warning C26407 messages from VS2017 Code Analysis:
> Prefer scoped objects, don't heap-allocate unnecessarily

### DIFF
--- a/Modules/IO/TransformHDF5/src/itkHDF5TransformIO.cxx
+++ b/Modules/IO/TransformHDF5/src/itkHDF5TransformIO.cxx
@@ -117,11 +117,6 @@ HDF5TransformIOTemplate<TParametersValueType>::WriteParameters(const std::string
                                                                const ParametersType & parameters)
 {
   const hsize_t dim(parameters.Size());
-  auto *        buf = new ParametersValueType[dim];
-  for (hsize_t i{ 0 }; i < dim; i++)
-  {
-    buf[i] = parameters[i];
-  }
   H5::DataSpace paramSpace(1, &dim);
 
   H5::DataSet paramSet;
@@ -146,9 +141,8 @@ HDF5TransformIOTemplate<TParametersValueType>::WriteParameters(const std::string
   {
     paramSet = this->m_H5File->createDataSet(name, h5StorageIdentifier, paramSpace);
   }
-  paramSet.write(buf, h5StorageIdentifier);
+  paramSet.write(parameters.data_block(), h5StorageIdentifier);
   paramSet.close();
-  delete[] buf;
 }
 
 template <typename TParametersValueType>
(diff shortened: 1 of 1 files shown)


N-Dekker:
Similar improvements appear possible in itkHDF5ImageIO.cxx  I could add them to this commit, as a force-push, if you like that...

dzenanz:
Sure, go ahead!

N-Dekker:
> Sure, go ahead!

Thansk @dzenanz  But then we're back to WIP, while we were so close to a merge 😸

dzenanz:
You can also make that a new PR.

dzenanz:
`itkPyBufferMemoryLeakTest (Failed)` - was this also the problem before?

thewtex:
> itkPyBufferMemoryLeakTest (Failed) - was this also the problem before?

This was the test that failed -- let's see if it fails again. :eyeglasses:

N-Dekker:
>> itkPyBufferMemoryLeakTest (Failed) - was this also the problem before?

Very interesting! The name of the test suggests that the pull request would introduce a memory leak, but I don't see how that could be possible! Actually I think the pull request might actually _fix_ some memory leaks! I mean, the old code would leak when an exception would occur between a "`buf = new` ..." and a "`delete[] buf`".

Can someone please explain what the `itkPyBufferMemoryLeakTest` is doing? I see it has three times the same piece of code:

    if M[5] - M[4] > 1000:
        print('Memory leak!')
        sys.exit(1)

What does that mean, exactly? Why should `M[5] - M[4]` be _at most_ 1000?

https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Bridge/NumPy/wrapping/test/itkPyBufferMemoryLeakTest.py#L40

N-Dekker:
FYI, My last force-push has only amended to commit subject (no code changes). I just hope a rebuild will chase away that `itkPyBufferMemoryLeakTest` failure, because I don't understand it!

N-Dekker:
> Can someone please explain what the itkPyBufferMemoryLeakTest is doing?

It's still not clear to me. Why should `M[5] - M[4] <= 1000`? https://github.com/InsightSoftwareConsortium/ITK/blob/v5.1b01/Modules/Bridge/NumPy/wrapping/test/itkPyBufferMemoryLeakTest.py#L40

dzenanz:
Tracing blames leads to this [commit](https://github.com/InsightSoftwareConsortium/ITKBridgeNumPy/commit/4d006eb39d2eb1db13a3f7ce6e639ee715d4360e).

PR #1426, opened 2019-11-18
Update Meta IO

dzenanz:


blowekamp:
This PR appears to contain no changes...?

dzenanz:
Maybe all the changes are getting stripped away by the ITK import script? The changes were concerning the standalone builds of MetaIO.

dzenanz:
> but it doesn't look like it is needed.

Indeed, now we can say that. But let's update so we are fully in sync with MetaIO master.

PR #1427, opened 2019-11-19
ENH: wrap StreamingImageFilter for all the dimensions

dzenanz:
Needed for fixing the wrapping in https://github.com/InsightSoftwareConsortium/ITKBoneEnhancement/pull/20.

PR #1429, opened 2019-11-19
ENH: wrap StreamingImageFilter for vector images

dzenanz:
Re-creation of #1427 because I could not reopen it.


dzenanz:
The test failure is unrelated. I think the Python build simply timed out.

PR #1434, opened 2019-11-21
BUG: Wait for all threads when abort or exception occurs

### LINKED ISSUE #1437: Crash in SignedMaurerDistanceMapImageFilter when aborting
### Description

I am trying to track down a crash I am seeing when aborting SignedMaurerDistanceMapImageFilter while it is generating a distance map. The crash occurs at line 299 in itkImageSource.hxx, ThreaderCallback() :

ThreadIdType total = str->Filter->SplitRequestedRegion(threadId, threadCount, splitRegion);

The issue is that the Filter smart pointer is pointing to nullptr at this point, therefore causing a crash.

I have tracked this down to the clear up code in the PoolMultiThreader class, SingleMethodExecute() function. The ProcessAborted exception is caught (line 110), and the threads are then waited on to ensure they have all terminated. The issue, I believe is that if any of the remaining threads also throw an exception (likely to be ProcessAborted) during this clear up process, it is re-thrown (line 123) meaning that any remaining threads are not waited for.

In SignedMaurerDistanceMapImageFilter::GenerateData(), the ThreadStruct is allocated on the stack, meaning that when the exception gets to this level, the variable str goes out of scope and therefore its Filter member is cleared up. I believe this leads to the crash in ImageSource as any threads still running will attempt to access the Filter member (and the now out of scope and destructed str variable).

I have commented out the throw on line 123 of itkPoolMultiThreader.cxx and rebuilt ITK to check my theory and this change does seems to solve the problem. Does anyone have any thoughts on this? Is this the best solution to this or is there a better one?

I am using ITK 5.0.1 and the machine I am running on is a 12 core hyperthreaded AMD Ryzen so there will potentially be a lot of thread pool threads running concurrently which may cause the issue to show more easily.

### Steps to Reproduce

Attached is code to reproduce the issue.

[Abort Crash.zip](https://github.com/InsightSoftwareConsortium/ITK/files/3879039/Abort.Crash.zip)

### Expected behavior

Not to crash, but to abort cleanly.

### Actual behavior

The crash (access violation) occurs either in the SignedMaurerDistanceMapImageFilter::voronoi or ImageSource::ThreaderCallback functions.

### Reproducibility

The crash occurs about once in every three runs on my machine.

### Versions

5.0.1

### Environment

Visual Studio 2017 (15.9.17), Debug x64 build.
CMake is version 3.15.4.
OS is Windows 10, 1909, build 18363.476.
Machine is an AMD Ryzen 9 3900X (12 cores, hyperthreaded) with 32 GB of RAM.

### Additional Information

If I comment out the throw on line 123 of itkPoolMultiThreader.cxx, the issue seems to be resolved, however this may lead to other unexpected behaviour as the secondary exception will not be re-thrown.

blowekamp:
closes #1437

## PR Checklist

### DIFF
--- a/Modules/Core/Common/src/itkPoolMultiThreader.cxx
+++ b/Modules/Core/Common/src/itkPoolMultiThreader.cxx
@@ -174,30 +174,27 @@ PoolMultiThreader ::ParallelizeArray(SizeValueType             firstIndex,
       chunkSize++; // we want slightly bigger chunks to be processed first
     }
 
+    auto lambda = [aFunc](SizeValueType start, SizeValueType end) {
+      for (SizeValueType ii = start; ii < end; ii++)
+      {
+        aFunc(ii);
+      }
+      // make this lambda have the same signature as m_SingleMethod
+      return ITK_THREAD_RETURN_DEFAULT_VALUE;
+    };
+
     SizeValueType workUnit = 1;
     for (SizeValueType i = firstIndex + chunkSize; i < lastIndexPlus1; i += chunkSize)
     {
-      m_ThreadInfoArray[workUnit++].Future = m_ThreadPool->AddWork(
-        [aFunc](SizeValueType start, SizeValueType end) {
-          for (SizeValueType ii = start; ii < end; ii++)
-          {
-            aFunc(ii);
-          }
-          // make this lambda have the same signature as m_SingleMethod
-          return ITK_THREAD_RETURN_DEFAULT_VALUE;
-        },
-        i,
-        std::min(i + chunkSize, lastIndexPlus1));
+      m_ThreadInfoArray[workUnit++].Future = m_ThreadPool->AddWork(lambda, i, std::min(i + chunkSize, lastIndexPlus1));
     }
     itkAssertOrThrowMacro(workUnit <= m_NumberOfWorkUnits, "Number of work units was somehow miscounted!");
 
     ExceptionHandlingPrologue;
 
     // execute this thread's share
-    for (SizeValueType ii = firstIndex; ii < firstIndex + chunkSize; ii++)
-    {
-      WrapExceptionHandling(aFunc(ii));
-    }
+    WrapExceptionHandling(lambda(firstIndex, firstIndex + chunkSize));
+
     // now wait for the other computations to finish
     for (SizeValueType i = 1; i < workUnit; i++)
     {
--- a/Modules/Core/Common/src/itkPoolMultiThreader.cxx
+++ b/Modules/Core/Common/src/itkPoolMultiThreader.cxx
@@ -77,6 +77,50 @@ PoolMultiThreader::SetMaximumNumberOfThreads(ThreadIdType numberOfThreads)
   m_MaximumNumberOfThreads = m_ThreadPool->GetMaximumNumberOfThreads();
 }
 
+#define ExceptionHandlingPrologue                                                                                      \
+  bool abortOccurred = false;                                                                                          \
+  bool exceptionOccurred = false;                                                                                      \
+                                                                                                                       \
+  ProcessAborted abortException;                                                                                       \
+  std::string    exceptionDetails;
+
+#define WrapExceptionHandling(statement)                                                                               \
+  try                                                                                                                  \
+  {                                                                                                                    \
+    statement;                                                                                                         \
+  }                                                                                                                    \
+  catch (ProcessAborted & e)                                                                                           \
+  {                                                                                                                    \
+    abortOccurred = true;                                                                                              \
+    abortException = e;                                                                                                \
+  }                                                                                                                    \
+  catch (std::exception & e)                                                                                           \
+  {                                                                                                                    \
+    exceptionDetails = e.what();                                                                                       \
+    exceptionOccurred = true;                                                                                          \
+  }                                                                                                                    \
+  catch (...)                                                                                                          \
+  {                                                                                                                    \
+    exceptionOccurred = true;                                                                                          \
+  }
+
+#define ExceptionHandlingEpilogue                                                                                      \
+  if (abortOccurred)                                                                                                   \
+  {                                                                                                                    \
+    throw abortException;                                                                                              \
+  }                                                                                                                    \
+  if (exceptionOccurred)                                                                                               \
+  {                                                                                                                    \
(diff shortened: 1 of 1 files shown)


blowekamp:
This may address the issue raised here:
https://discourse.itk.org/t/crash-in-signedmaurerdistancemapimagefilter-when-aborting/2432/3

Modules/Core/Common/src/itkPoolMultiThreader.cxx
```diff
@@ -99,55 +102,47 @@ PoolMultiThreader::SingleMethodExecute()
...
-  try
-  {
-    // Now, the parent thread calls this->SingleMethod() itself
-    m_ThreadInfoArray[0].UserData = m_SingleData;
-    m_ThreadInfoArray[0].NumberOfWorkUnits = m_NumberOfWorkUnits;
-    m_SingleMethod((void *)(&m_ThreadInfoArray[0]));
+  // Now, the parent thread calls this->SingleMethod() itself
+  m_ThreadInfoArray[0].UserData = m_SingleData;
+  m_ThreadInfoArray[0].NumberOfWorkUnits = m_NumberOfWorkUnits;
+  m_SingleMethod((void *)(&m_ThreadInfoArray[0]));
```
blowekamp: The call should be wrapped with exception handling too.

blowekamp:
> Should analogous changes be made in the other 2 threaders?

The platform multithreaded should be ok. One issue here is that the [std::future::get](https://en.cppreference.com/w/cpp/thread/future/get) will re-throw the an exception. So the handling needs to be quite different.

Unfortunately, I wasn't able to reproduce the issue, much less create a test. Clearly, all the threads were not waited for if an exception occurred, so I fixed that.

I'll leave the TBB threader to you.

Modules/Core/Common/src/itkPoolMultiThreader.cxx
```diff
@@ -205,6 +204,7 @@ PoolMultiThreader ::ParallelizeArray(SizeValueType             firstIndex,
       {
         filter->UpdateProgress(i / float(workUnit));
       }
+      // TODO: Wrap in exception handling
```
dzenanz: I don't think that parallelize array can be aborted. Also, I think we want the exception to be propagated to the caller in its unaltered form.

blowekamp: But the `.get` call may throw an exception, when other threads are still running.

dzenanz: It should not be hard to record the exception and re-throw it after the for loop, in a way similar the the rest of this patch?

dzenanz: OK I am writing a follow-up commit for this.

dzenanz:
TBB multi threader already correctly handles exceptions.

@blowekamp should we turn @ianlindsay's code from #1437 into a unit test?

blowekamp:
> @blowekamp should we turn @ianlindsay's code from #1437 into a unit test?

That would be nice. I just dated a SimpleITK abort test to use the SignedMaurerDistanceMap over the StatisticsImageFilter:
https://github.com/SimpleITK/SimpleITK/blob/d705c3b2850a993e33040f1ccd54439860e7bb25/Testing/Unit/sitkBasicFiltersTests.cxx#L827-L874

I was not able to readily reproduce the issue with is. Adding more abort testing in ITK would be very good todo. I think the progress reporting and abort handling in ITKv5 are still a little rough.

N-Dekker:
It's already merged! I still wanted to ask if it's necessary to use macro's here. I think `#define`'s should only be used when really necessary.

dzenanz:
There is a lot of boilerplate code. And these macros are only used within this file.

dzenanz:
> It's already merged!

I wanted someone else besides me and Brad take a look. Hans was the fastest 😄

dzenanz:
Not all people provide a review when requested, and not all provide it fast. So that's why I usually request a few people take a look, and I usually merge after first approval - especially for non-contentious and minor changes.

N-Dekker:
This looks to me like a non-trivial change in a critical part of ITK, that would deserve some extra time to review.

dzenanz:
I am usually waiting for [people watching the repository](https://github.com/InsightSoftwareConsortium/ITK/watchers) to review for some time before prodding them to review explicitly. If you wanted more involvement you could also watch the repo and review proactively - review (some of) the PRs before the author asks you to.

The automatically enforced rules are that a green build and one approving review by a maintainer is enough. I don't know who has time to review, so I select a few people. Usually, some don't review. In these recent PRs #1426 #1421 #1414 someone requested your review along with other people, and you didn't review before those PRs were merged. I can't know who is willing to review any particular PR.

Waiting too long for reviews slows down development cycle. Adding more exception handling to cover all the cases should be non-contentious - at least I though so.  But it is in an important place, so I should have waited more for reviews.

I get merge-happy in the late afternoons, because I want the patch to make it into the master before the nightly testing begins. Perhaps I could be more mindful of people's time zones before I invite them to review?

N-Dekker:
@dzenanz Thanks for your explanation. I guess it was just bad luck: I was very much busy studying the pull request, and while I was half way, you already merged.

I did not know that this PR would be merged so quickly. While you did not know that I was actually busy studying the code.

It would have helped me if I would have known beforehand when the PR would be merged. For example, if you could indicate that the PR would be merged in 30 minutes, unless anyone would object. On the other hand, it might have helped you if I would have indicated that I was still busy reviewing. 

It would be nice if GitHub would support both these: indicate how much time is left before a planned merge, and allow a reviewer to indicate "busy reviewing, please wait", by means of a fancy button click.

blowekamp:
@N-Dekker I like your macro free suggestion. I think it is worth pursuing.

N-Dekker:
@blowekamp Thanks! I guess it might also be interesting to consider storing the exception as a C++11  `std::exception_ptr`, in your `catch` clause, by calling `std::current_exception()`. I think `std::exception_ptr' was specifically introduced for the kind of issue that your PR is about.

Note that  `std::exception_ptr` could store any type of exception, not just `std::exception`.

dzenanz:
> if you could indicate that the PR would be merged in 30 minutes

This could be a nice solution, assuming GitHub doesn't introduce their own (better) solution.

Issue #1431, opened 2019-11-20
Visual Studio errors and warnings


dzenanz:
With the latest master (75ec4e508e31ef93cac17235db5a5600590bc1a8), VS2019 gives the below errors and warnings, most important of which are VNL link errors of the style `unresolved external symbol "public: __cdecl vnl_matrix<double>::vnl_matrix<double>`.
```text
In project ITKDCMTK_ExtProject (ExternalProjectTargets\ITKDCMTK_ExtProject\ITKDCMTK_ExtProject), file C:\Dev\ITK-2019\Modules\ThirdParty\DCMTK\CUSTOMBUILD, line 1: Warning : PNG support will be disabled because libpng directory was not specified. Correct path and re-enable DCMTK_WITH_PNG.
In project ITKDCMTK_ExtProject (ExternalProjectTargets\ITKDCMTK_ExtProject\ITKDCMTK_ExtProject), file C:\Dev\ITK-2019\Modules\ThirdParty\DCMTK\CUSTOMBUILD, line 1: Warning : TIFF support will be disabled because libtiff directory was not specified. Correct path and re-enable DCMTK_WITH_TIFF.
In project ITKDCMTK_ExtProject (ExternalProjectTargets\ITKDCMTK_ExtProject\ITKDCMTK_ExtProject), file C:\Dev\ITK-2019\Modules\ThirdParty\DCMTK\CUSTOMBUILD, line 1: Warning : ZLIB support will be disabled because zlib directory was not specified. Correct path and re-enable DCMTK_WITH_ZLIB.
In project ITKDCMTK_ExtProject (ExternalProjectTargets\ITKDCMTK_ExtProject\ITKDCMTK_ExtProject), file C:\Dev\ITK-2019\Modules\ThirdParty\DCMTK\CUSTOMBUILD, line 1: Warning : 
In project ITKGPUCommon, file C:\Dev\ITK-git\Modules\Core\GPUCommon\src\itkGPUContextManager.cxx, line 71: Warning C4996: 'clCreateCommandQueue': was declared deprecated
In project ITKGPUCommon, file C:\Program Files (x86)\OpenCL_SDK_Light\include\CL\cl.h, line 1359: Message : see declaration of 'clCreateCommandQueue'
[...]

PR #1436, opened 2019-11-21
COMP: fixing a couple of VNL related linking errors

dzenanz:
These two errors are getting fixed. Ref #1431.

In project ITKCommon1TestDriver, file C:\Dev\ITK-2019\Modules\Core\Common\test\itkPointGeometryTest.obj, line 1: Error LNK2019: unresolved external symbol "public: __cdecl vnl_vector_ref<double>::vnl_vector_ref<double>(unsigned int,double *)" (??0?$vnl_vector_ref@N@@QEAA@IPEAN@Z) referenced in function "public: class vnl_vector_ref<double> __cdecl itk::Point<double,3>::GetVnlVector(void)" (?GetVnlVector@?$Point@N$02@itk@@QEAA?AV?$vnl_vector_ref@N@@XZ)

In project ITKCommon1TestDriver, file C:\Dev\ITK-2019\Modules\Core\Common\test\itkPointGeometryTest.obj, line 1: Error LNK2019: unresolved external symbol "public: __cdecl vnl_vector_ref<double>::~vnl_vector_ref<double>(void)" (??1?$vnl_vector_ref@N@@QEAA@XZ) referenced in function "int __cdecl itkPointGeometryTest(int,char * * const)" (?itkPointGeometryTest@@YAHHQEAPEAD@Z)

### DIFF
--- a/Modules/Core/Common/test/itkPointGeometryTest.cxx
+++ b/Modules/Core/Common/test/itkPointGeometryTest.cxx
@@ -98,7 +98,7 @@ itkPointGeometryTest(int, char *[])
   std::cout << "Squared Euclidean distance between pc and pb = ";
   std::cout << distance2 << std::endl;
 
-  vnl_vector_ref<ValueType> vnlVector = pa.GetVnlVector();
+  auto vnlVector = pa.GetVnlVector();
   std::cout << "vnl_vector = ";
   {
     for (unsigned int i = 0; i < N; i++)


dzenanz:
@hjmjohnson I have further linking errors, mostly constructors and destructors #1431. I feel what I am proposing here is a workaround and not a fix. Why is this needed? I suspect the root cause is recent refactoring of `vnl_vector_ref` and `vnl_matrix_ref` classes.

Modules/Core/Common/test/itkPointGeometryTest.cxx
```diff
@@ -98,7 +98,7 @@ itkPointGeometryTest(int, char *[])
   std::cout << "Squared Euclidean distance between pc and pb = ";
   std::cout << distance2 << std::endl;
 
-  vnl_vector_ref<ValueType> vnlVector = pa.GetVnlVector();
+  auto vnlVector = pa.GetVnlVector();
```
hjmjohnson: // This is fine as is,  const version would be slightly better.
const auto vnlVector = pa.GetVnlVector();

PR #1443, opened 2019-11-25
ENH: update BoneEnhancement remote module

dzenanz:
Some of the filters are now streamable,
greatly reducing main filter's memory consumption.


dzenanz:
The big merged happened in https://github.com/InsightSoftwareConsortium/ITKBoneEnhancement/pull/20. The python packaging part of CI should start working once #1429 makes it into the next release of ITK.

PR #1445, opened 2019-11-27
ENH: Add UnsignedPower(base, exp) and UnsignedProduct(a, b) to itk::Math

N-Dekker:
Added `constexpr` functions `UnsignedPower(base, exponent)` and
`UnsignedProduct(a, b)` to the `itk::Math` namespace.

Especially `UnsignedPower` appears interesting, as it allows estimating
the number of voxels of an N-dimensional hypercube at compile-time
without conversions between `unsigned` and floating point.
`UnsignedProduct` is a helper function of `UnsignedPower`, but it is also
being used in `itk::Experimental::ConnectedImageNeighborhoodShape`, with
this commit.

Allowed specifying the return type by a template argument, as requested
by Bradley @blowekamp

Replaced `ShapeLabelObject::IntegerPow` by a call to `UnsignedPower`, as
suggested by Bradley @blowekamp

### DIFF
--- a/Modules/Core/Common/include/itkConnectedImageNeighborhoodShape.h
+++ b/Modules/Core/Common/include/itkConnectedImageNeighborhoodShape.h
@@ -19,6 +19,7 @@
 #ifndef itkConnectedImageNeighborhoodShape_h
 #define itkConnectedImageNeighborhoodShape_h
 
+#include "itkMath.h"
 #include "itkOffset.h"
 
 #include <array>
@@ -190,15 +191,6 @@ private:
   }
 
 
-  // Calculates a * b. Numeric overflow triggers a compilation error in
-  // "constexpr context" and a debug assert failure at run-time.
-  static constexpr std::uintmax_t
-  CalculateProduct(const std::uintmax_t a, const std::uintmax_t b) ITK_NOEXCEPT
-  {
-    return (((a * b) / a) == b) && (((a * b) / b) == a) ? (a * b) : (ITK_X_ASSERT(!"CalculateProduct overflow!"), 0);
-  }
-
-
   // Calculates 2 ^ n. Numeric overflow triggers a compilation error in
   // "constexpr context" and a debug assert failure at run-time.
   static constexpr std::uintmax_t
@@ -206,14 +198,6 @@ private:
   {
     return (n < std::numeric_limits<std::uintmax_t>::digits) ? (std::uintmax_t{ 1 } << n)
                                                              : (ITK_X_ASSERT(!"CalculatePowerOfTwo overflow!"), 0);
-  }
-
-
-  // Calculates 3 ^ n
-  static constexpr std::uintmax_t
-  CalculatePowerOfThree(const std::size_t n) ITK_NOEXCEPT
-  {
-    return (n == 0) ? 1 : CalculateProduct(3, CalculatePowerOfThree(n - 1));
   }
 
 
@@ -225,7 +209,7 @@ private:
   CalculateBinomialCoefficient(const std::uintmax_t n, const std::uintmax_t k) ITK_NOEXCEPT
   {
     return (k > n) ? (ITK_X_ASSERT(!"Out of range!"), 0)
-                   : (k == 0) ? 1 : CalculateProduct(n, CalculateBinomialCoefficient(n - 1, k - 1)) / k;
+                   : (k == 0) ? 1 : Math::UnsignedProduct(n, CalculateBinomialCoefficient(n - 1, k - 1)) / k;
   }
 
 
@@ -236,13 +220,13 @@ private:
   CalculateNumberOfHypercubesOnBoundaryOfCube(const std::size_t m, const std::size_t n) ITK_NOEXCEPT
   {
     // Calculate 2^(n-m) * BinomialCoefficient(n, m)
-    return CalculateProduct(CalculatePowerOfTwo(n - m),
-                            (((2 * m) < n) ?
-                                           // Calculate either the binomial coefficient of (n, m) or (n, n - m).
-                                           // Mathematically, both should yield the same number, but the
-                                           // calculation is optimized for a smaller second argument.
-                               CalculateBinomialCoefficient(n, m)
-                                           : CalculateBinomialCoefficient(n, n - m)));
+    return Math::UnsignedProduct(CalculatePowerOfTwo(n - m),
+                                 (((2 * m) < n) ?
+                                                // Calculate either the binomial coefficient of (n, m) or (n, n - m).
+                                                // Mathematically, both should yield the same number, but the
+                                                // calculation is optimized for a smaller second argument.
+                                    CalculateBinomialCoefficient(n, m)
+                                                : CalculateBinomialCoefficient(n, n - m)));
   }
 
 
@@ -263,7 +247,7 @@ private:
     return (((maximumCityblockDistance == 0) || (ImageDimension == 0))
               ? 0
               : ((maximumCityblockDistance >= ImageDimension)
-                   ? (CalculatePowerOfThree(ImageDimension) - 1)
+                   ? (Math::UnsignedPower(3, ImageDimension) - 1)
                    : CalculateSumOfNumberOfHypercubesOnBoundaryOfCube(ImageDimension - 1,
                                                                       (ImageDimension - maximumCityblockDistance))));
   }
--- a/Modules/Core/Common/include/itkMath.h
+++ b/Modules/Core/Common/include/itkMath.h
@@ -762,6 +762,57 @@ GreatestPrimeFactor(unsigned long n);
 ITKCommon_EXPORT unsigned long long
 GreatestPrimeFactor(unsigned long long n);
 
+// C++11 does not guarantee that assert can be used in constexpr
+// functions. This is a work-around for GCC 4.8, 4.9. Originating
+// from Andrzej's C++ blog:
+//  https://akrzemi1.wordpress.com/2017/05/18/asserts-in-constexpr-functions/
+#if defined NDEBUG
+#  define ITK_X_ASSERT(CHECK) void(0)
+#else
+#  define ITK_X_ASSERT(CHECK) ((CHECK) ? void(0) : [] { assert(!#CHECK); }())
+#endif
+
+/**  Returns `a * b`. Numeric overflow triggers a compilation error in
+ * "constexpr context" and a debug assert failure at run-time.
+ */
+template <typename TReturnType = std::uintmax_t>
+constexpr TReturnType
+UnsignedProduct(const std::uintmax_t a, const std::uintmax_t b) ITK_NOEXCEPT
+{
+  static_assert(std::is_unsigned<TReturnType>::value, "UnsignedProduct only supports unsigned return types");
+
+  // Note that numeric overflow is not "undefined behavior", for unsigned numbers.
+  // This function checks if the result of a*b is mathematically correct.
+  return (a == 0) || (b == 0) ||
+             (((static_cast<TReturnType>(a * b) / a) == b) && ((static_cast<TReturnType>(a * b) / b) == a))
+           ? static_cast<TReturnType>(a * b)
+           : (ITK_X_ASSERT(!"UnsignedProduct overflow!"), 0);
+}
+
+
+/** Calculates base ^ exponent. Numeric overflow triggers a compilation error in
+ * "constexpr context" and a debug assert failure at run-time. Otherwise equivalent to
+ * C++11 `static_cast<std::uintmax_t>(std::pow(base, exponent))`
+ *
+ * \note `UnsignedPower(0, 0)` is not supported, as zero to the power of zero has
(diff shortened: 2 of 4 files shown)


Modules/Core/Common/include/itkMath.h
```diff
@@ -762,6 +762,54 @@ GreatestPrimeFactor(unsigned long n);
...
+           : (ITK_X_ASSERT(!"UnsignedProduct overflow!"), 0);
+}
+
+
+/** Calculates base ^ exponent. Numeric overflow triggers a compilation error in
+ * "constexpr context" and a debug assert failure at run-time. Otherwise equivalent to
+ * C++11 `static_cast<std::uintmax_t>(std::pow(base, exponent))`
+ *
+ * \note `UnsignedPower(0, 0)` is not supported, as zero to the power of zero does
+ * have no agreed-upon value: https://en.wikipedia.org/wiki/Zero_to_the_power_of_zero
```
dzenanz: does have no agreed-upon value -> has no agreed-upon value **OR** does not have an agreed-upon value

Modules/Core/Common/include/itkMath.h
```diff
@@ -762,6 +762,54 @@ GreatestPrimeFactor(unsigned long n);
...
+constexpr std::uintmax_t
+UnsignedPower(const std::uintmax_t base, const std::uintmax_t exponent) ITK_NOEXCEPT
+{
+  // Uses recursive function calls because C++11 does not support other ways of
+  // iterations for a constexpr function.
+  return (exponent == 0)
+           ? (ITK_X_ASSERT(base > 0), 1)
+           : (exponent == 1)
+               ? base
+               : UnsignedProduct(UnsignedPower(base, exponent / 2), UnsignedPower(base, (exponent + 1) / 2));
```
dzenanz: Neat trick!

blowekamp:
Cool! I don't think there is a need to place these little functions into an Experimental namespace. Also It looks like you could replace [this code](https://github.com/InsightSoftwareConsortium/ITK/blob/c71fe7d6311a08cc867cf12d835d22e385475d2b/Modules/Filtering/LabelMap/include/itkShapeLabelObject.h#L326-L336) with the function as well.

Modules/Core/Common/include/itkConnectedImageNeighborhoodShape.h
```diff
@@ -235,14 +221,16 @@ class ConnectedImageNeighborhoodShape
...
+
     // Calculate 2^(n-m) * BinomialCoefficient(n, m)
-    return CalculateProduct(CalculatePowerOfTwo(n - m),
-                            (((2 * m) < n) ?
-                                           // Calculate either the binomial coefficient of (n, m) or (n, n - m).
-                                           // Mathematically, both should yield the same number, but the
-                                           // calculation is optimized for a smaller second argument.
-                               CalculateBinomialCoefficient(n, m)
-                                           : CalculateBinomialCoefficient(n, n - m)));
+    return UnsignedProduct(CalculatePowerOfTwo(n - m),
```
dzenanz: Why not also replace `CalculatePowerOfTwo`?

Modules/Core/Common/include/itkMath.h
```diff
@@ -762,6 +762,54 @@ GreatestPrimeFactor(unsigned long n);
...
+
+/** Calculates base ^ exponent. Numeric overflow triggers a compilation error in
+ * "constexpr context" and a debug assert failure at run-time. Otherwise equivalent to
+ * C++11 `static_cast<std::uintmax_t>(std::pow(base, exponent))`
+ *
+ * \note `UnsignedPower(0, 0)` is not supported, as zero to the power of zero does
+ * have no agreed-upon value: https://en.wikipedia.org/wiki/Zero_to_the_power_of_zero
+ */
+constexpr std::uintmax_t
+UnsignedPower(const std::uintmax_t base, const std::uintmax_t exponent) ITK_NOEXCEPT
```
blowekamp: You should template this method over the output type. I don't think having exponent be a larger type than unsigned byte is useful.

Modules/Core/Common/include/itkMath.h
N-Dekker: Thanks, @dzenanz  Honestly, I looked at `vnl_finite_int_poly::Ntothe` (which is private, and non-`constexpr`, unfortunately): https://github.com/InsightSoftwareConsortium/ITK/blob/v5.1b01/Modules/ThirdParty/VNL/src/vxl/core/vnl/vnl_finite.h#L439

Modules/Core/Common/include/itkConnectedImageNeighborhoodShape.h
N-Dekker: @dzenanz `CalculatePowerOfTwo` does a simple bit-shift, no need to do recursive function calls. I like that better for powers of 2. OK?

Modules/Core/Common/include/itkMath.h
N-Dekker: @blowekamp 
> I don't think there is a need to place these little functions into an Experimental namespace

My main reason for putting them in an `Experimental` namespace is because I was not 100% sure about the names of the new functions. I thought about `CalculateUnsignedPower`,  `CalculatePowerOfUnsigned`, etc., and `CalculateUnsignedProduct` or `MultiplyUnsigned`...

But I think the current function names are also fine. 😃 So if no one objects, I'll adjust the PR, so that they are directly placed in the `itk::Math` namespace.

N-Dekker: @blowekamp 
> You should template this method over the output type

The result can easily become very large, that 's why I thought `std::uintmax_t` would be appropriate as a return type. (Note that on most platforms, `uintmax_t` and `size_t` have the same size anyway.)

> I don't think having exponent be a larger type than unsigned byte is useful.

A common use case would be to have an `ImageDimension` as exponent, which is often defined as 32-bit or 64-bit unsigned. I would not want to require the user to explicitly cast their argument from 32/64-bit to byte, when using the value as exponent. Would you?

Can you show me how you would want the function to look like?

blowekamp: > The result can easily become very large, that 's why I thought std::uintmax_t would be appropriate as a return type. (Note that on most platforms, uintmax_t and size_t have the same size anyway.)

This would not be true on 32-bit platforms.

>A common use case would be to have an ImageDimension as exponent, which is often defined as 32-bit or 64-bit unsigned. I would not want to require the user to explicitly cast their argument from 32/64-bit to byte, when using the value as exponent. Would you?

I agree avoiding casting is good, and that is logical for the exponent type. The same reasoning can be made for having a templated output type. 😄 

```
template <typename TValueType=std::uintmax_t>
constexpr TValueType
UnsignedPower(const TValueType base, const std::uintmax_t exponent)
```

N-Dekker: Thanks for your suggestion @blowekamp Unfortunately I haven't got it working yet. When I modified the signature as you suggested, the test `UnsignedPower(2, 3)` deduced `TValueType = int`.  It completely ignored the default, `TValueType=std::uintmax_t`, which I would like to have! I'll have another look...

blowekamp: Ahh, templates are a complicated thing. It should have been obvious that the temple parameters will automatically be deduced, and the default parameter for the function should not make much sense.

N-Dekker: @blowekamp I'm preparing a force-push, which allows specifying the return type of `UnsignedPower`. It will still be `uintmax_t` by default, though. Hope that's OK to you. Please wait  😄

N-Dekker:
FYI: The force-push that I just did moved the two functions directly into `itk::Math` (no longer introducing an `itk::MathExperimental` namespace. It fixed a small English style issue (" no agreed-upon value") found by @dzenanz, and it allowed specifying the return type, by means of a template argument, as requested by @blowekamp

Modules/Core/Common/include/itkMath.h
N-Dekker: @blowekamp My last force-push does include compile-time tests, that the return type of a function call `UnsignedPower<T>(1, 1)` is `T`. Please check: 

https://github.com/InsightSoftwareConsortium/ITK/blob/996f95938ed19da6e9166ee7a2a784b64db2b38f/Modules/Core/Common/test/itkMathTest.cxx#L50

Modules/Core/Common/include/itkMath.h
```diff
@@ -762,6 +762,55 @@ GreatestPrimeFactor(unsigned long n);
...
+#  define ITK_X_ASSERT(CHECK) ((CHECK) ? void(0) : [] { assert(!#CHECK); }())
+#endif
+
+/**  Returns `a * b`. Numeric overflow triggers a compilation error in
+ * "constexpr context" and a debug assert failure at run-time.
+ */
+template <typename TValueType = std::uintmax_t>
+constexpr TValueType
+UnsignedProduct(const std::uintmax_t a, const std::uintmax_t b) ITK_NOEXCEPT
+{
```
N-Dekker: It would be nice to (also) check here that `TValueType` is unsigned:

      static_assert(std::is_unsigned<TValueType>::value, "Only supports unsigned return types");

N-Dekker:
The last force-push just has a few very small stylish improvements: 
- Renamed `TValueType` to `TReturnType`.
- Added a check that  that the user-specified `TReturnType` is unsigned, also for `UnsignedProduct(a, b)`
- Improved an error message of a `static_assert` in MathTest.
- Ordered standard C++ #include's alphabetically

N-Dekker:
@blowekamp wrote:
> Also It looks like you could replace [this code](https://github.com/InsightSoftwareConsortium/ITK/blob/c71fe7d6311a08cc867cf12d835d22e385475d2b/Modules/Filtering/LabelMap/include/itkShapeLabelObject.h#L326-L336) with the function as well.

My last force-push does replace `ShapeLabelObject::IntegerPow` by a call to `UnsignedPower`, as you suggested  😃 Please check!

Modules/Filtering/LabelMap/include/itkShapeLabelObject.h
```diff
@@ -343,7 +330,7 @@ class ShapeLabelObject : public LabelObject<TLabel, VImageDimension>
   using OrientedBoundingBoxSizeType = Vector<double, VImageDimension>;
 
   using OrientedBoundingBoxVerticesType =
-    FixedArray<OrientedBoundingBoxPointType, IntegerPow<2, ImageDimension>::Result>;
+    FixedArray<OrientedBoundingBoxPointType, Math::UnsignedPower<unsigned int>(2, ImageDimension)>;
```
N-Dekker: Here it is, @blowekamp ensuring _at compile time_ that the power can be represented by an `unsigned int`, using the template argument that you requested 😃

N-Dekker:
I think this pull request is ready now. The `itkPyBufferMemoryLeakTest` failure seems unrelated: https://open.cdash.org/testDetails.php?test=823766290&build=6229318

The same test failure occurred before: https://github.com/InsightSoftwareConsortium/ITK/pull/1424#issuecomment-557077795

N-Dekker:
Just did a `git rebase master` to see if the `itkPyBufferMemoryLeakTest` failure might just go away "automagically".

Modules/Filtering/LabelMap/include/itkShapeLabelObject.h
blowekamp: Great! Thank you for taking the time to improve the utility and generalization for your contribution.

dzenanz:
Did this PR introduce these warnings?
https://open.cdash.org/viewBuildError.php?type=1&onlydeltap&buildid=6234485

N-Dekker:
> Did this PR introduce these warnings?
> https://open.cdash.org/viewBuildError.php?type=1&onlydeltap&buildid=6234485

Is that a rhetorical question?  😺 Thanks @dzenanz  I'll have a look:

> [CTest: warning matched] Modules/Core/Common/include/itkMath.h:802:1: warning: inline function 'itk::Math::UnsignedPower<unsigned long>' is not defined [-Wundefined-inline]
UnsignedPower(const std::uintmax_t base, const std::uintmax_t exponent) ITK_NOEXCEPT
^
[CTest: warning matched] Modules/Core/Common/include/itkMath.h:810:75: note: used here
                                           : UnsignedProduct<TReturnType>(UnsignedPower(base, exponent / 2),

dzenanz:
The question was not entirely rhetorical, because these warning appeared on the dashboard only last night, even though this PR was merged 5 days ago.

N-Dekker:
@dzenanz It appears an old Clang compiler bug. I'm able to reproduce it with the old clang 3.8.1, at https://godbolt.org/z/Rwskcq with the following simple program:

    template <typename T = unsigned>
    constexpr T f(unsigned i)
    {
      return (i == 0) ? 0 : f(i - 1);
    }

    int main() {}

A workaround could be to pass `TReturnType` as template argument to both of the recursive `UnsignedPower` function calls, inside the body of `UnsignedPower`. But I can't try that out today anymore!

PR #1448, opened 2019-11-28
BUG: avoid undefined behavior (dereferencing nullptr). Closes #1358

### LINKED ISSUE #1358: null deref in 'itkTreeContainerTest' found by UBSan
itkTreeContainerTest fails under UBSan on my dashboards, ex:

https://open.cdash.org/testDetails.php?test=799482596&build=6175118

```
/Users/builder/external/ITK/Modules/Core/Common/include/itkChildTreeIterator.hxx:112:28: runtime error: member call on null pointer of type 'itk::TreeNode<int>'
```

in the debugger:

```
frame #4: 0x0000000100140292 ITKCommon2TestDriver`itk::ChildTreeIterator<itk::TreeContainer<int> >::Next(this=<unavailable>) at itkChildTreeIterator.hxx:112:28 [opt]
   109 	{
   110 	  m_ListPosition++;
   111 	  this->m_Position = m_ParentNode->GetChild(m_ListPosition);
-> 112 	  return this->m_Position->Get();
   113 	}
   114 	
   115 	/** Clone function */
```

@dzenanz I thought you fixed this last year, but maybe this is another variation:

https://discourse.itk.org/t/help-with-null-deref-in-itktreecontainertest/826/1

dzenanz:
## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkLeafTreeIterator.h
+++ b/Modules/Core/Common/include/itkLeafTreeIterator.h
@@ -134,6 +134,10 @@ const typename LeafTreeIterator<TTreeType>::ValueType &
 LeafTreeIterator<TTreeType>::Next()
 {
   this->m_Position = const_cast<TreeNodeType *>(FindNextNode());
+  if (this->m_Position == nullptr)
+  {
+    return this->m_Root->Get(); // value irrelevant, but we have to return something
+  }
   return this->m_Position->Get();
 }
 


seanm:
Look reasonable to me, but I don't know this code really.

Modules/Core/Common/include/itkLeafTreeIterator.h
```diff
@@ -134,6 +134,10 @@ const typename LeafTreeIterator<TTreeType>::ValueType &
 LeafTreeIterator<TTreeType>::Next()
 {
   this->m_Position = const_cast<TreeNodeType *>(FindNextNode());
+  if (this->m_Position == nullptr)
+  {
+    return this->m_Root->Get(); // value irrelevant, but we have to return something
```
N-Dekker: Sorry, I'm not familiar with this class. Is it still being used? If so, could a user application actually get at this line of code, during program execution? Should it get an assert failure here?

seanm: A test gets there: https://open.cdash.org/testDetails.php?test=817215405&build=6229026

dzenanz: I am not familiar with this class either, and I don't know if it is being used.

The unit test gets to here, but the result is commonly discarded, so the whole operation is commonly optimized away. But as Sean's tool notes, it is undefined behavior.

seanm: The tool in question is -fsanitize=null https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html

N-Dekker: If the return value is commonly ignored, maybe just make it `void`. Otherwise I think an assert failure might be helpful. But again, I don't really know.

dzenanz: `operator++` (iterator increment) returns this reference, which is common in C++, but the [usual use pattern](https://github.com/InsightSoftwareConsortium/ITK/blob/2074c2f9e5ed3f087d0e2059f8e0e8992fcad7ef/Modules/Core/Common/test/itkTreeContainerTest.cxx#L160-L167) does not use that reference. And since this group of classes seems to be used rarely (if at all), refactoring them is not a priority.

dzenanz:
Windows Python packages timed out.

seanm:
@dzenanz there are still 2 more null deref related treecontainer failures:

Test: itkTreeContainerTest and itkTreeContainerTest2
https://open.cdash.org/viewTest.php?onlyfailed&buildid=6239707

PR #1451, opened 2019-11-29
STYLE: Replace macro's in PoolMultiThreader by ExceptionHandling class

N-Dekker:
Replaces the macro's `ExceptionHandlingPrologue`, `WrapExceptionHandling`,
and `ExceptionHandlingEpilogue` by a class, `ExceptionHandling`, aiming
to improve code readability and maintainability.

Bradley @blowekamp encouraged pursuing such a "macro free" approach,
during the discussion following the merge of pull request #1434, "BUG:
Wait for all threads when abort or exception occurs".

### DIFF
--- a/Modules/Core/Common/src/itkPoolMultiThreader.cxx
+++ b/Modules/Core/Common/src/itkPoolMultiThreader.cxx
@@ -30,12 +30,60 @@
 #include "itkProcessObject.h"
 #include "itkImageSourceCommon.h"
 #include <algorithm>
+#include <exception>
 #include <iostream>
 #include <string>
-#include <algorithm>
 
 namespace itk
 {
+namespace
+{
+class ExceptionHandler
+{
+public:
+  template <typename TFunction>
+  explicit ExceptionHandler(const TFunction & function)
+  {
+    try
+    {
+      function();
+    }
+    catch (...)
+    {
+      m_FirstCaughtException = std::current_exception();
+    }
+  }
+
+  void
+  TryToGetFuture(PoolMultiThreader::ThreadPoolInfoStruct & threadPoolInfo)
+  {
+    try
+    {
+      threadPoolInfo.Future.get();
+    }
+    catch (...)
+    {
+      if (m_FirstCaughtException == nullptr)
+      {
+        m_FirstCaughtException = std::current_exception();
+      }
+    }
+  }
+
+  void
+  RethrowFirstCaughtException() const
+  {
+    if (m_FirstCaughtException != nullptr)
+    {
+      std::rethrow_exception(m_FirstCaughtException);
+    }
+  }
+
+private:
+  std::exception_ptr m_FirstCaughtException;
+};
+} // namespace
+
 
 PoolMultiThreader::PoolMultiThreader()
   : m_ThreadPool(ThreadPool::GetInstance())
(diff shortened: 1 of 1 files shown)


blowekamp:
The class looks a lot cleaner to me than macros.

I believe the original code kept and re-threw the original exception not the subsequent... It looks like it has gotten changed along the way. The first exception may be the reason the other exceptions occur ( like compilation errors ), so it may be important to keep the original behavior.

N-Dekker:
> I believe the original code kept and re-threw the original exception not the subsequent... It looks like it has gotten changed along the way. The first exception may be the reason the other exceptions occur ( like compilation errors ), so it may be important to keep the original behavior.

@blowekamp Thanks for your suggestion. 😃  So what do you think of my [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/b2963545538b763ba59cd79ff20d605eab0a36f1..01e30cdd42d759114da15973c301699ef7c49021)?

Modules/Core/Common/src/itkPoolMultiThreader.cxx
```diff
@@ -30,12 +30,60 @@
...
+{
+public:
+  template <typename TFunction>
+  explicit ExceptionHandler(const TFunction & function)
+  {
+    try
+    {
+      function();
+    }
+    catch (...)
```
N-Dekker: Honestly I believe an exception should always be derived from `std::exception`, as a general C++ coding conversion, so it should be sufficient to do `catch (const std::exception&)`, instead of `catch (...)`. What do you think?

Modules/Core/Common/src/itkPoolMultiThreader.cxx
```diff
@@ -190,10 +192,8 @@ PoolMultiThreader ::ParallelizeArray(SizeValueType             firstIndex,
     }
     itkAssertOrThrowMacro(workUnit <= m_NumberOfWorkUnits, "Number of work units was somehow miscounted!");
 
-    ExceptionHandlingPrologue;
-
     // execute this thread's share
-    WrapExceptionHandling(lambda(firstIndex, firstIndex + chunkSize));
+    ExceptionHandler exceptionHandler([lambda, firstIndex, chunkSize] { lambda(firstIndex, firstIndex + chunkSize); });
```
dzenanz: Hans thought that variable name `lambda` is not descriptive enough, now is a good time to change it. Not necessary, though.

Modules/Core/Common/src/itkPoolMultiThreader.cxx
blowekamp: This code needs to be bullet proof. If an exception is not caught it can lead to segmentation faults, abandoned threads, and other bad things.  It wraps the execution of user's threaded code. We don't have any control of what some one may running threaded. We need to be able to handle anything they throw at us ( pun intended ).

dzenanz: That will hold water until someone comes along who writes `throw "Error HEREEEEEEE";` and then comes crying to the forum because his program is crashing without explanation.

People don't always follow conventions, so we should handle all exceptions.

N-Dekker: OK, no problem, I will leave the `catch(...)` in there, if this is the way you want it.  It's just that https://www.boost.org/community/error_handling.html said so, many years ago:

> Avoid catch(...) when possible

dzenanz: This handles exceptions both from our and other people's code. We can't expect everyone to be nice. Avoiding `catch(...)` here is quite detrimental.

Modules/Core/Common/src/itkPoolMultiThreader.cxx
N-Dekker: Honestly it's actually slightly worse now because I placed the variable named `lambda` inside a newly introduced _unnamed_ lambda! I could entirely get rid of the named `lambda` and just use two small unnamed lambda's. Would you like that? @hjmjohnson

dzenanz: The reason for introducing a named lambda was to reduce code duplication. It being invoked from another, unnamed lambda is not a problem.

N-Dekker:
Looking at the code before (or even after) this PR, it appears that an exception is sometimes thrown between the `m_ThreadPool->AddWork` and the corresponding "future-get". Is that OK? Here (at the master branch):
https://github.com/InsightSoftwareConsortium/ITK/blob/bde411e2b40fc325d83c54f41a4e57faf3d7a88c/Modules/Core/Common/src/itkPoolMultiThreader.cxx#L266

dzenanz:
Yes, that is an exception we want to throw here, because our assumptions are not satisfied. If that is thrown, something is horribly wrong.

N-Dekker:
> Yes, that is an exception we want to throw here, because our assumptions are not satisfied. If that is thrown, something is horribly wrong.

So what would happen if some threads from `m_ThreadInfoArray` would still be running at that line of code?  Would they get aborted?

dzenanz:
Assuming their data does not get deallocated, they would continue running until completion. Otherwise they would probably crash the program due to trying to access wrong memory.

N-Dekker:
Thanks @dzenanz. Another question: my PR does not explicitly distinguish between `ProcessAborted` exceptions and other exceptions. It just catches all of them and rethrows the first one. Is that OK? I'm just asking, to be sure 😃 Of course, the user may still catch `ProcessAborted`, if it has been rethrown.

PR #1457, opened 2019-12-02
STYLE: Improve ImageIORegion special member functions, move-semantics

N-Dekker:
Added fast, noexcept move-semantics, by a defaulted move-constructor and
a defaulted move-assignment operator. Also defaulted the existing
default-constructor and copy-constructor.

Adjusted the converting constructor, `ImageIORegion(unsigned int)` to
initialize the member data by a member-initializer-list.

Let assignments return `*this` (instead of `void`), which is the
"standard" return value of an assignment. Made the assignment
exception safe.

Added GoogleTest unit tests, with help from includecpp.org members.

### DIFF
--- a/Modules/Core/Common/src/itkImageIORegion.cxx
+++ b/Modules/Core/Common/src/itkImageIORegion.cxx
@@ -17,44 +17,41 @@
  *=========================================================================*/
 
 #include "itkImageIORegion.h"
+#include <algorithm> // For copy_n.
 
 namespace itk
 {
-ImageIORegion ::ImageIORegion()
-{
-  m_ImageDimension = 2;
-  m_Index.resize(2);
-  m_Size.resize(2);
-  std::fill(m_Index.begin(), m_Index.end(), 0);
-  std::fill(m_Size.begin(), m_Size.end(), 0);
-}
 
 ImageIORegion ::~ImageIORegion() = default;
 
 ImageIORegion ::ImageIORegion(unsigned int dimension)
-{
-  m_ImageDimension = dimension;
-  m_Index.resize(m_ImageDimension);
-  m_Size.resize(m_ImageDimension);
-  std::fill(m_Index.begin(), m_Index.end(), 0);
-  std::fill(m_Size.begin(), m_Size.end(), 0);
-}
+  : m_ImageDimension{ dimension }
+  , m_Index(dimension)
+  , m_Size(dimension)
+{}
 
-ImageIORegion ::ImageIORegion(const Self & region)
-  : Region()
-{
-  m_Index = region.m_Index;
-  m_Size = region.m_Size;
-  m_ImageDimension = region.m_ImageDimension;
-}
 
-void
+ImageIORegion &
 ImageIORegion ::operator=(const Self & region)
 {
-  m_Index = region.m_Index;
-  m_Size = region.m_Size;
-  m_ImageDimension = region.m_ImageDimension;
+  if ((region.m_Index.size() == m_Index.size()) && (region.m_Size.size() == m_Size.size()))
+  {
+    // Copy the values from 'region', but do not change the size of m_Index and m_Size.
+    std::copy_n(region.m_Index.cbegin(), m_Index.size(), m_Index.begin());
+    std::copy_n(region.m_Size.cbegin(), m_Size.size(), m_Size.begin());
+    m_ImageDimension = region.m_ImageDimension;
+  }
+  else
+  {
+    // Copy the region to 'temp' and then do a non-throwing move.
+    Self temp(region);
+    static_assert(noexcept(*this = std::move(temp)),
+                  "Move-assignment should be noexcept, to provide the strong exception guarantee for copy-assignment.");
+    *this = std::move(temp);
+  }
+  return *this;
 }
+
 
 std::ostream &
 operator<<(std::ostream & os, const ImageIORegion & region)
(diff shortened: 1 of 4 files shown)


Modules/Core/Common/test/itkImageIORegionGTest.cxx
```diff
@@ -0,0 +1,186 @@
...
+// First include the header file to be tested:
+#include "itkImageIORegion.h"
+#include <gtest/gtest.h>
+#include <random>
+#include <type_traits> // For is_default_constructible, is_copy_constructible, etc.
+
+namespace
+{
+template <typename T>
+constexpr bool                               IsDefaultConstructibleCopyableNoThrowMovableAndDestructible =
```
dzenanz: What is this aligned with? Does `clang-format` produce this?

N-Dekker: @dzenanz Well observed! But honestly, I swear I only typed a single space... `clang-format` alignment automation messed it up quite badly :astonished: in this particular case, I guess it's just a clang-format bug. But of course, I'm still in favor of entirely reconsidering the auto-align option https://discourse.itk.org/t/update-coding-style-for-itk/2055/10

Modules/Core/Common/src/itkImageIORegion.cxx
```diff
@@ -17,45 +17,42 @@
...
-  m_Size = region.m_Size;
-  m_ImageDimension = region.m_ImageDimension;
-}
 
-void
+ImageIORegion &
 ImageIORegion ::operator=(const Self & region)
 {
-  m_Index = region.m_Index;
-  m_Size = region.m_Size;
```
N-Dekker: FYI Here is where the original `ImageIORegion` assignment operator was exception unsafe: if `m_Size = region.m_Size` would throw an exception, it could leave `this` object in a corrupted state: `m_Index.size()` and `m_Size.size()` could then be different.

N-Dekker: Note that a compiler-generated _(defaulted)_ copy-assignment would not be exception-safe either, as it would just do a member-wise assignment. Which is why I did not follow the Rule Of Zero, this time.

dzenanz: I was wondering why not follow the rule of zero this time 😄

N-Dekker: > I was wondering why not follow the rule of zero this time

Well... I guess there's a trade-off between code simplicity and exception-safety this time. Do we care about the scenario when the `ImageIORegion` assignment _might_ get an exception when it's only half way...? Or would the Rule Of Zero be good enough for `ImageIORegion`?

blowekamp: I think Dzenan was thanking you for the explanation as these details for the motivation of the code were not self evident. Perhaps a comment is needed as to why the rule of zero was not followed.

Thank you for paying attention to these details.

dzenanz: Yes, I forgot to thank for the explanation 😕

N-Dekker:
FYI: While I was writing tests for the "special member functions" of this class, I felt that it would have been nice if GoogleTest would already have some helper functions to test copy and move semantics. So I posted the following at the Google Group on GTest:  [How to test CopyConstructible, MoveConstructible, Assignable requirements?
](https://groups.google.com/forum/#!topic/googletestframework/uZ4kJr7VvF8)

dzenanz:
Some problems on the dashboard:
https://open.cdash.org/viewBuildError.php?type=0&buildid=6237834

N-Dekker:
Thanks @dzenanz  Regarding the first compile error from  RogueResearch13/Mac10.10-AppleClang-dbg-x86_64-static :
> /Users/builder/externalModules/Core/Common/test/itkImageIORegionGTest.cxx:173:28: error: default initialization of an object of const type 'const itk::ImageIORegion' without a user-provided default constructor

That looks like an old Clang <= 3.8.1 issue (again). The following produces the same error for old Clang versions, but compiles well for Clang >= 3.9:

    struct MyStruct { int i = 2; };

    int main()
    {
        const MyStruct myConstStruct;
    }

See https://godbolt.org/z/WqxVVN

seanm:
@N-Dekker removing that const at line 173 results in everything compiling. Thanks!

Clang 3.8 is not soooo old, only 3.5 years. :)  SupportedCompilers.md claims support back to Clang 3.3 (from 2013).

CC also @vfonov

N-Dekker:
@ seanm Could you possibly make the PR?  😄 

> Clang 3.8 is not soooo old, only 3.5 years.  :)

Maybe I should say: C++ is modernizing soooo rapidly! 😃 I wish I could be using modern C++17 features, instead of sticking to the old (2011) C++11 standard.

seanm:
:)

Patch coming...

N-Dekker:
Note that the Clang suggested adding `{}`, rather than removing the `const` keyword:

    /Users/builder/externalModules/Core/Common/test/itkImageIORegionGTest.cxx:173:52: note: add an explicit initializer to initialize 'defaultInitializedRegion'
      const itk::ImageIORegion defaultInitializedRegion;
                                                       ^
                                                       {}

This is in general a nice suggestion. However, it's not what this test was meant for!!! I wanted to test that a default-initialization _without_ `{}` still properly initializes the object! So for the sake of the test, it would be preferable to just remove the `const`.

PR #1458, opened 2019-12-02
ENH: DiscreteGaussian filter is compatible with SmoothingRecursive one

dzenanz:
* add sigma setters and getters to DiscreteGaussianImageFilter
* make second template parameter equal to first one by default


dzenanz:
The force-push adds basic setter/getter tests.

PR #1461, opened 2019-12-04
ENH: use RTK v2.1.0 (remote module)

SimonRit:
RTK has been regularly updated to account for ITK changes (master branch). I just released a new RTK version for inclusion in ITK v5.1rc01 (if it's still time).


dzenanz:
This is probably just in time. @thewtex?

PR #1462, opened 2019-12-04
COMP: Workaround Clang 3.8 -Wundefined-inline warning Math UnsignedPower

N-Dekker:
RogueResearch13/Mac10.10-AppleClang-dbg-x86_64-static produced the
following warning, as reported by Dženan @dzenanz at pull request #1445:

> [CTest: warning matched] Modules/Core/Common/include/itkMath.h:802:1:
> warning: inline function `itk::Math::UnsignedPower<unsigned long>`
> is not defined [-Wundefined-inline]

It appears to be an old Clang 3.8.1 compiler issue. This commit aims to
work around the issue by passing `TReturnType` as template argument to
both of the recursive `UnsignedPower` function calls, inside the body
of `UnsignedPower`.

### DIFF
--- a/Modules/Core/Common/include/itkMath.h
+++ b/Modules/Core/Common/include/itkMath.h
@@ -805,10 +805,11 @@ UnsignedPower(const std::uintmax_t base, const std::uintmax_t exponent) ITK_NOEX
 
   // Uses recursive function calls because C++11 does not support other ways of
   // iterations for a constexpr function.
-  return (exponent == 0) ? (ITK_X_ASSERT(base > 0), 1)
-                         : (exponent == 1) ? base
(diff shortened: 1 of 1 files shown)


seanm:
Just spent a while looking into this warning then ultimately discovered this PR. :)  But thanks for fixing it!  If you could remember to CC me on any fixes related to my bots that'd help reduce the chance of duplicate work. :)

dzenanz:
It didn't even occur to me that you might be looking for a fix for this too 😞

seanm:
No worries. We've been working lately to get our ITK dashboards green. Sometimes I'll take a stab at warnings as they are usually easier than failing tests :)

PR #1464, opened 2019-12-05
COMP: fixed compilation with AppleClang 7 and 8.

seanm:
Just removed a 'const' as suggested by Niels Dekker.

Also added a note to the supported compilers list about AppleClang.


seanm:
@N-Dekker @vfonov @dzenanz

N-Dekker:
Thank you @seanm Can you please add the fixed error message from https://open.cdash.org/viewBuildError.php?type=0&buildid=6237834 to the commit message?

thewtex:
@seanm @N-Dekker @dzenanz thank you for the quick follow-up on this!

PR #1465, opened 2019-12-05
BUG: Fix wrapping for TransformBaseTemplate class

thewtex:
The class was re-named from TransformBase to TransformBaseTemplate, but
the wrapping name was not updated.

Also wrap double transform's, the most common.


thewtex:
> How did you notice this error?

I made an attempt to write transforms in Python and ran into this issue. I believe this does not fully address that use case, yet, but it is a step in the right direction.

dzenanz:
Do you think this PR was merged to late to fix [these errors](https://open.cdash.org/viewBuildError.php?type=0&buildid=6239470) for the last nightly?

thewtex:
```
	/home/lorensen/ProjectsGIT/Dashboards/ITK-SkullStrip/Modules/Filtering/FastMarching
Exit Condition	1
Standard Error	
../../../lib/libITKSpatialObjects-5.1.so.1: error: undefined reference to 'vtable for itk::TransformBaseTemplate<double>'
../../../lib/libITKSpatialObjects-5.1.so.1: error: undefined reference to 'typeinfo for itk::TransformBaseTemplate<double>'
collect2: error: ld returned 1 exit status
```

This does not look related.

PR #1467, opened 2019-12-05
BUG: Use itk.array_from_image for __array__

thewtex:
Experimentation through use has found that the use of a view
for np.asarray can result in unexpected behavior or crashes.


thewtex:
> Does this approach use copy-on-write, or immediate copy?

An immediate copy.

PR #1468, opened 2019-12-05
ENH: Gracefully convert non-contiguous NumPy arrays

### LINKED ISSUE #1243: ITK package requires numpy array be of order='C' before using ITK.GetImageFromArray
### Description

Regarding the conversion of Numpy arrays to ITK Images, I have found that an ITK package will report an error if the numpy array is not of order type 'C'. Specifically, when using an ITK Image that has been created from the conversion of a numpy array with itk.CoocurrenceTextureFeaturesImageFilter(), I will get the following error:

> Requested region is (at least partially) outside the largest possible region.

Changing the numpy array to order 'C' resolved this issue. This doesn't seem to happen with arbitrarily created arrays using, for example, np.random.rand(), but does happen in my case with a pre-processed image. 

### Expected information

The proper format for a numpy array before conversion to ITKImage type.

### Actual information

https://itkpythonpackage.readthedocs.io/en/latest/Quick_start_guide.html#mixing-itk-and-numpy

### Versions

itk                       5.0.1                    pypi_0    pypi
itk-core                  5.0.1                    pypi_0    pypi
itk-filtering             5.0.1                    pypi_0    pypi
itk-io                    5.0.1                    pypi_0    pypi
itk-numerics              5.0.1                    pypi_0    pypi
itk-registration          5.0.1                    pypi_0    pypi
itk-segmentation          5.0.1                    pypi_0    pypi
itk-texturefeatures       3.2.2                    pypi_0    pypi

thewtex:
Instead of raising an exception when converting a NumPy array to an
itk.Image, just make the array contiguous with numpy.ascontiguousarray.
This improves ease of use, and internal conversion is a better default
behavior.

Closes #1243

### DIFF
--- a/Modules/Bridge/NumPy/wrapping/PyBuffer.i.in
+++ b/Modules/Bridge/NumPy/wrapping/PyBuffer.i.in
@@ -92,9 +92,8 @@
 
         assert ndarr.ndim in ( 2, 3, 4 ), \
             "Only arrays of 2, 3 or 4 dimensions are supported."
-        if( not ndarr.flags['C_CONTIGUOUS'] and not ndarr.flags['F_CONTIGUOUS'] ):
-            raise ValueError("Array memory is not contiguous. Try converting your array with "
-                  + "`ascontiguousarray()` or `copy()` or use `GetImageFromArray()`")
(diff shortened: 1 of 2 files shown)


thewtex:
@kyochujoho please take a look

Issue #1469, opened 2019-12-05
Disable CMake user registry lookup by find_package


jcfr:
### Description

Configuring ITK on a system where unrelated build exported their build tree using [export(PACKAGE)](https://cmake.org/cmake/help/latest/command/export.html#command:export) can have unintended side effect.

For example, on my system the package registry contains the following:

### Steps to Reproduce

Configure ITK with `ITK_USE_SYSTEM_EIGEN` set to `ON` on a system where `Eigen3` is found in the cmake registry:

```
ls -1 ~/.cmake/packages/
Eigen3
```

### Expected behavior

Find Eigen3 installed with the package manager or fail if not available.

### Actual behavior

`CMakeCache.txt` has the following value:

```
Eigen3_DIR:PATH=/home/jcfr/Projects/SlicerSALT-Release-Qt5/Slicer-build/E/Eigen3
```

### Proposed solution

Set `CMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY` to `1`

See https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#disabling-the-package-registry

### Versions

Observed with d2e511437d

### Environment

Any

phcerdan:
Closed in #1470

PR #1470, opened 2019-12-05
COMP: Disable CMake user registry lookup by find_package.

jcfr:
See #1469

## PR Checklist


thewtex:
@jcfr thanks for taking the time to file the issue and submit patch on this tricky time sink :potable_water: that others will hopefully not need to encounter :pray: :1st_place_medal: 

Per the link in the report, since `CMAKE_FIND_NO_PACKAGE_REGISTRY ` [is deprecated](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY.html#variable:CMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY), should [`CMAKE_FIND_USE_PACKAGE_REGISTRY`](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_USE_PACKAGE_REGISTRY.html#variable:CMAKE_FIND_USE_PACKAGE_REGISTRY`]) be used instead?

jcfr:
> should CMAKE_FIND_USE_PACKAGE_REGISTRY be used instead?

Until ITK depends on CMake >= 3.16, I suggest we keep using the deprecated variable. Indeed, the new one `CMAKE_FIND_USE_PACKAGE_REGISTRY` was introduced in CMake 3.16
See https://cmake.org/cmake/help/latest/release/3.16.html?highlight=cmake_find_use_package_registry#commands

We could check for the version of CMake and set one or the other ?

hjmjohnson:
> We could check for the version of CMake and set one or the other ?

I think we should check version numbers.  That way when we require CMake 3.16 (in a few years :) ), it will be easy to remove the outdated variable name.

jcfr:
> I think we should check version numbers. That way when we require CMake 3.16 (in a few years :) ), it will be easy to remove the outdated variable name.

@hjmjohnson  Great idea :bulb: , let me know what you think.

CMakeLists.txt
```diff
@@ -242,8 +242,15 @@ set(CMAKE_DEBUG_POSTFIX "" CACHE STRING "Generate debug library name with a post
...
 
 # Configure find_package behavior
-if(NOT DEFINED CMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY)
-  set(CMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY 1)
+if("${CMAKE_VERSION}" VERSION_GREATER_EQUAL "3.16")
+  if(NOT DEFINED CMAKE_FIND_USE_PACKAGE_REGISTRY)
+    set(CMAKE_FIND_USE_PACKAGE_REGISTRY 1)
+  endif()
+else()
+  message("HERE")
```
hjmjohnson: OOPS:  Remove "HERE" message.

jcfr: Good catch. Thanks

CMakeLists.txt
```diff
@@ -242,8 +242,14 @@ set(CMAKE_DEBUG_POSTFIX "" CACHE STRING "Generate debug library name with a post
 mark_as_advanced(CMAKE_DEBUG_POSTFIX)
 
 # Configure find_package behavior
-if(NOT DEFINED CMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY)
-  set(CMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY 1)
+if("${CMAKE_VERSION}" VERSION_GREATER_EQUAL "3.16")
+  if(NOT DEFINED CMAKE_FIND_USE_PACKAGE_REGISTRY)
+    set(CMAKE_FIND_USE_PACKAGE_REGISTRY 1)
```
dzenanz: I thought the logic here is positive, so this should be `FALSE`, not `1`.

jcfr: An other good catch ... :pray:  Thanks for the review

jcfr:
Thanks for the review. I will integrate once the CI finishes.

PR #1472, opened 2019-12-06
COMP: reformulate a self-assignment to fix -Wself-assign-overloaded w…

seanm:
…arning

The warning was:

Modules/Core/Common/test/itkImageIORegionGTest.cxx:82:8: warning: explicitly assigning value of variable of type 'T' to itself [-Wself-assign-overloaded]
  self = self;
  ~~~~ ^ ~~~~


seanm:
@N-Dekker @vfonov @dzenanz

PR #1473, opened 2019-12-06
BUG: Fixed signed overflow in float ULP comparisons

seanm:
Found by UBSan.

Removed a previous attempt to fix this (207a801f2).

Use the technique described in the already-referenced blog post to avoid overflow by first checking if the two floats being compared have different signs.

This will fix several unit test failures, example: itkDiscreteGaussianDerivativeImageFilterTest

### DIFF
--- a/Modules/Core/Common/include/itkMath.h
+++ b/Modules/Core/Common/include/itkMath.h
@@ -28,6 +28,7 @@
 #ifndef itkMath_h
 #define itkMath_h
 
+#include <cmath>
 #include "itkMathDetail.h"
 #include "itkConceptChecking.h"
 #include <vnl/vnl_math.h>
@@ -280,8 +281,8 @@ FloatAddULP(T x, typename Detail::FloatIEEE<T>::IntType ulps)
  *
  * The implementation is based off the explanation in the white papers:
  *
- * - http://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/
- * - http://randomascii.wordpress.com/category/floating-point/
+ * - https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/
+ * - https://randomascii.wordpress.com/category/floating-point/
  *
  * This function is not a cure-all, and reading those articles is important
  * to understand its appropriate use in the context of ULPs, zeros, subnormals,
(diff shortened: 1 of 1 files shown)


seanm:
@N-Dekker @hjmjohnson @dzenanz @vfonov another thing found by UBSan causing several of our dashboard test failures....

Modules/Core/Common/include/itkMath.h
```diff
@@ -318,25 +319,26 @@ FloatAlmostEqual(T                                        x1,
...
+  // result and overflow an int.
+  if (std::signbit(x1) != std::signbit(x2))
   {
-    ulps = -ulps;
+    // Check for equality to make sure +0 == -0
+    if (x1 == x2)
+    {
+      return true;
+    }
+    return false;
```
N-Dekker: Interesting issue @seanm By the way, it looks like these five lines of code are equivalent to just `return  x1 == x2`. In this case I would prefer one line of code, instead on five!

seanm: Indeed. I blindly copied https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/

Will change.

Modules/Core/Common/include/itkMath.h
```diff
@@ -318,25 +319,22 @@ FloatAlmostEqual(T                                        x1,
...
-  if (ulps < 0)
+  // This check for different signs is necessary for several reasons, see the blog link above.
+  // Subtracting the signed-magnitude representation of floats using twos-complement
+  // math isn't particularly meaningful, and the subtraction would produce a 33-bit
+  // result and overflow an int.
+  if (std::signbit(x1) != std::signbit(x2))
   {
-    ulps = -ulps;
+    // Check for equality to make sure +0 == -0
+    return (x1 == x2);
```
N-Dekker: Thanks for following my suggestion @seanm Now I'm sorry I just realize that this might still produce a warning! Like this:
> warning: comparing floating point with == or != is unsafe [-Wfloat-equal]

seanm: Which is an oddly worded warning. There's nothing "unsafe" about such comparisons, they just aren't always what one wants/expects. But here exact comparison is what's wanted, so such a warning is a false positive. The only workaround I know is to suppress it with a pragma. I could make an `AreFloatsExactlyEqual` function to wrap it...

N-Dekker: @seanm  We're lucky, it's there already  😃 
 
https://github.com/InsightSoftwareConsortium/ITK/blob/9531f046a35134861386854ae491b0956f49fb60/Modules/Core/Common/include/itkMath.h#L723

seanm: Reading the blog more carefully, and in fact we don't even need it.

Modules/Core/Common/include/itkMath.h
```diff
@@ -318,25 +319,21 @@ FloatAlmostEqual(T                                        x1,
...
-  Detail::FloatIEEE<T> x2f(x2);
-  double               x1fAsULP = static_cast<double>(x1f.AsULP());
-  double               x2fAsULP = static_cast<double>(x2f.AsULP());
-  double               ulps = x1fAsULP - x2fAsULP;
-  if (ulps < 0)
+  // This check for different signs is necessary for several reasons, see the blog link above.
+  // Subtracting the signed-magnitude representation of floats using twos-complement
+  // math isn't particularly meaningful, and the subtraction would produce a 33-bit
+  // result and overflow an int.
+  if (std::signbit(x1) != std::signbit(x2))
```
N-Dekker: This looks like a major bug fix to me, thanks again Sean. Apparently this essential sign comparison was missing already from the very first version by @thewtex and @bradking, Feb 25, 2012: 0d8ca651d64e8c03efbdb807ed8d2f5893dc8ee5

seanm: It's funny, because when I was debugging this, I thought: "I remember reading a great blog about this", then was surprised to see the blog mentioned in the ITK comments!  But for some reason the blogger's implementation was not copied fully. ITK is now doing basically exactly the same as his `AlmostEqualUlpsAndAbs` function. (Which BTW excludes the 0 == -0 part, for the reason you surmised).

It was sorta caught in late 2015 with 207a801, but the change only hid the problem really.

Floating point math is hard. :)

thewtex: The implementation deviated, e.g. adding support for both `float` and `double`. But, this comparison was lost in the process. Lucky us, we now have `std::signbit` we can use with C++11 :-).

PR #1477, opened 2019-12-07
BUG: fixed off-by-one error found by ASan

seanm:


seanm:
@dzenanz @thewtex @N-Dekker @vfonov this fixes https://open.cdash.org/testDetails.php?test=829568168&build=6241376

I think it's the correct change...

PR #1479, opened 2019-12-09
Update VXL library to the latest upstream master

vfonov:
Updating VXL to incorporate changes from vxl/vxl#729
@seanm 

## PR Checklist

- [ ] :no_entry_sign: [Makes breaking changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes)
- [ ] :no_entry_sign: [Makes design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes)
- [ ] :no_entry_sign: Adds the License notice to new files.
- [ ] :no_entry_sign: Adds Python wrapping.
- [ ] :no_entry_sign: Adds tests and baseline comparison (quantitative).
- [ ] :no_entry_sign: [Adds test data](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/UploadBinaryData.md).
- [ ] :no_entry_sign: Adds Examples to [ITKExamples](https://github.com/InsightSoftwareConsortium/ITKExamples)
- [ ] :no_entry_sign: Adds Documentation.

Refer to the [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) for
further development details if necessary.

PR #1480, opened 2019-12-09
Vxl update 20191208

hjmjohnson:
Supersedes and closes #1479.


dzenanz:
Test failures are unrelated.

dzenanz:
Wait, a merge commit is missing for this.

hjmjohnson:
@dzenanz If this passes, could we merge it right away, or should we wait for other PR's to complete?

dzenanz:
I am OK with merging right away. But that will be about 5 hours from now, so it will not make it into this night's nightly.

hjmjohnson:
The windows failure is new.  I won't be able to look into why windows is failing for a long while. ;(.

seanm:
CC @seanm @vfonov 

So @hjmjohnson so if this can't go in yet, can we at least merge the fix to https://github.com/vxl/vxl/pull/729 ?

hjmjohnson:
Python failures are unrelated:

```
AttributeError: module 'ITKConnectedComponents' has no attribute 'swig'
```

hjmjohnson:
@seanm I finally found time to track down the failure of the windows build.  I had made the static array in the test too big for windows to handle.  Reducing the size of the test buffer fixed the problem.

seanm:
Sweet, thanks!

PR #1486, opened 2019-12-11
ENH: Add boundary condition getters and setters

arobert01:
Add boundary condition setters and getters to DiscreteGaussianImageFilter.

We are currently trying to implement a new projector in RTK, and we need to use the DiscreteGaussianImageFilter with another boundary condition than the default one. 

The difficulty here is that the DiscreteGaussianImageFilter uses intermediate filter with real pixel type. In this PR, I implemented one setter and getter for boundary condition with input pixel type and another one for boundary condition with real pixel type. However, I'm not sure it's the best way to do since the user would have to set boundary condition with two different image type. 

Can you think of a better solution?

## PR Checklist

- [ ]  [Makes design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes)
- [ ] : Adds the License notice to new files.
- [ ]  Adds Python wrapping.
- [ ]  Adds tests and baseline comparison (quantitative).
- [ ]  [Adds test data](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/UploadBinaryData.md).
- [ ]  Adds Examples to [ITKExamples](https://github.com/InsightSoftwareConsortium/ITKExamples)
- [ ] Adds Documentation.

### DIFF
--- a/Modules/Filtering/Smoothing/include/itkDiscreteGaussianImageFilter.hxx
+++ b/Modules/Filtering/Smoothing/include/itkDiscreteGaussianImageFilter.hxx
@@ -140,19 +140,12 @@ DiscreteGaussianImageFilter<TInputImage, TOutputImage>::GenerateData()
     return;
   }
 
-  // Type of the pixel to use for intermediate results
-  using RealOutputPixelType = typename NumericTraits<OutputPixelType>::RealType;
-  using RealOutputImageType = Image<OutputPixelType, ImageDimension>;
-
-  using RealOutputPixelValueType = typename NumericTraits<RealOutputPixelType>::ValueType;
-
   // Type definition for the internal neighborhood filter
   //
   // First filter convolves and changes type from input type to real type
   // Middle filters convolves from real to real
   // Last filter convolves and changes type from real type to output type
   // Streaming filter forces the mini-pipeline to run in chunks
-
 
   using FirstFilterType =
     NeighborhoodOperatorImageFilter<InputImageType, RealOutputImageType, RealOutputPixelValueType>;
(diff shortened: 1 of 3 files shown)


dzenanz:
Looks good to me. Can you rebase on top of current master, for CI to run and to avoid the merge conflicts?

arobert01:
@dzenanz Rebased on master.

Modules/Filtering/Smoothing/include/itkDiscreteGaussianImageFilter.hxx
```diff
@@ -24,6 +24,7 @@
 #include "itkImageRegionIterator.h"
 #include "itkProgressAccumulator.h"
 #include "itkImageAlgorithm.h"
+#include <type_traits>
```
blowekamp: why is this include needed?

dzenanz:
Have you measured performance impact of this? Totally negligible or approximately negligible?

Modules/Filtering/Smoothing/include/itkDiscreteGaussianImageFilter.hxx
arobert01: My bad, it's not needed

arobert01:
> Please add m_RealBoundaryCondition to the print self method.
> 
> It would be great if the new code could be tested so we know it works correctly.

The force-push adds basic setter/getter tests and adds m_RealBoundaryCondition to the print self method.


> Have you measured performance impact of this? Totally negligible or approximately negligible?

I didn't measured performance impact, but since the changes only consist in adding setters and getters for boundary condition, I think it's totally negligible. Futhermore, I'm already using the changes in local for my RTK development without noticing any impact.

PR #1488, opened 2019-12-11
BUG: fix null pointer dereference in itkInOrderTreeIterator.h:102

dzenanz:
/Users/builder/external/ITK/Modules/Core/Common/include/itkInOrderTreeIterator.h:102:28: runtime error: member call on null pointer of type 'itk::TreeNode<int>'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /Users/builder/external/ITK/Modules/Core/Common/include/itkInOrderTreeIterator.h:102:28

### DIFF
--- a/Modules/Core/Common/include/itkInOrderTreeIterator.h
+++ b/Modules/Core/Common/include/itkInOrderTreeIterator.h
@@ -99,6 +99,10 @@ const typename InOrderTreeIterator<TTreeType>::ValueType &
 InOrderTreeIterator<TTreeType>::Next()
 {
   this->m_Position = const_cast<TreeNodeType *>(FindNextNode());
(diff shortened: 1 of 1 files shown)


dzenanz:
The [other problem](https://open.cdash.org/testDetails.php?test=799482597&build=6239707) seems to be a design error in either the classes or the test, an is not trivial to fix.

seanm:
Seems just like the previous fixes, so +1 from me, even though I don't know this class. :)

seanm:
So for the other one, that's not trivial to fix... could perhaps the test itself be changed to not invoke the buggy behaviour of the class?

dzenanz:
See #1490.

PR #1492, opened 2019-12-14
COMP: Exceptions should be caught by constant reference

hjmjohnson:
Modules/Core/SpatialObjects/test/itkContourSpatialObjectTest.cxx:244:31:
warning: catching polymorphic type ‘class itk::ExceptionObject’ by value [-Wcatch-value=]

Exceptions should be caught by constant reference, not by value.


hjmjohnson:
> Great! Do we also need to update the software guide with recommendations and examples regarding catching exceptions by reference?

That is "C++" best practices, so I think it would be OK, but not necessary.  I'd be cautious to not be too TLDR; regarding details like this.

dzenanz:
What I meant, update the examples in the software guide to follow this, and if there are statements opposite to this update or remove them. I did not think of expansion.

hjmjohnson:
@dzenanz Oh.  the Examples are already updated in the PR.  I am working on a bash script to detect this in the SphinxExamples too, but that might be a few weeks before I get to it.

dzenanz:
I meant update code snippets such as [these](https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/blob/8fbe524e12db703fbf9254fc348e13cba883e5db/SoftwareGuide/Latex/Appendices/CodingStyleGuide.tex#L3192-L3219).

hjmjohnson:
https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/pull/136

hjmjohnson:
Python failures are not related to these changes.  @thewtex something is happening related to swig.

https://open.cdash.org/viewTest.php?onlyfailed&buildid=6252493

Would @thewtex  or @dzenanz  merge when satisfied that python failures are some other problem?

thewtex:
> https://open.cdash.org/viewTest.php?onlyfailed&buildid=6252493

Something that is not reproducible sometimes causes these test failures on Linux CI builds.

PR #1501, opened 2019-12-18
ENH: Add VectorImageSupport to MeanImageFilter

### LINKED ISSUE #1377: Improve MeanImageFilter with VectorImage support
Follow up to  #1365

blowekamp:
closes #1377 

Add support to MeanImageFilter to support the VectorImage with it's run time variable number of components.

### DIFF
--- a/Modules/Filtering/Smoothing/include/itkMeanImageFilter.hxx
+++ b/Modules/Filtering/Smoothing/include/itkMeanImageFilter.hxx
@@ -26,6 +26,7 @@
 #include "itkNeighborhoodAlgorithm.h"
 #include "itkOffset.h"
 #include "itkShapedImageNeighborhoodRange.h"
+#include "itkDefaultConvertPixelTraits.h"
 
 namespace itk
 {
@@ -54,26 +55,31 @@ MeanImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateData(
 
   // Process the non-boundary subregion, using a faster pixel access policy without boundary extrapolation.
   GenerateDataInSubregion<BufferedImageNeighborhoodPixelAccessPolicy<InputImageType>>(
-    *input, *output, calculatorResult.GetNonBoundaryRegion(), neighborhoodOffsets);
+    *input,
+    *output,
+    calculatorResult.GetNonBoundaryRegion(),
+    neighborhoodOffsets,
+    static_cast<InputPixelType *>(nullptr));
 
   // Process each of the boundary faces. These are N-d regions which border
   // the edge of the buffer.
   for (const auto & boundaryFace : calculatorResult.GetBoundaryFaces())
   {
     GenerateDataInSubregion<ZeroFluxNeumannImageNeighborhoodPixelAccessPolicy<InputImageType>>(
-      *input, *output, boundaryFace, neighborhoodOffsets);
+      *input, *output, boundaryFace, neighborhoodOffsets, static_cast<InputPixelType *>(nullptr));
   }
 }
 
 
 template <typename TInputImage, typename TOutputImage>
-template <typename TPixelAccessPolicy>
+template <typename TPixelAccessPolicy, typename TPixelType>
 void
 MeanImageFilter<TInputImage, TOutputImage>::GenerateDataInSubregion(
   const TInputImage &                              inputImage,
   TOutputImage &                                   outputImage,
   const ImageRegion<InputImageDimension> &         imageRegion,
-  const std::vector<Offset<InputImageDimension>> & neighborhoodOffsets)
+  const std::vector<Offset<InputImageDimension>> & neighborhoodOffsets,
+  const TPixelType *)
 {
   const auto neighborhoodSize = static_cast<double>(neighborhoodOffsets.size());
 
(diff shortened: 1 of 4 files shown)


Modules/Filtering/Smoothing/include/itkMeanImageFilter.hxx
```diff
@@ -82,11 +83,29 @@ MeanImageFilter<TInputImage, TOutputImage>::GenerateDataInSubregion(
     inputImage, Index<InputImageDimension>(), neighborhoodOffsets);
   auto outputIterator = ImageRegionRange<OutputImageType>(outputImage, imageRegion).begin();
 
+  InputRealType sumRef;
+  using PixelConvertType = DefaultConvertPixelTraits<InputRealType>;
+  using PixelComponentType = typename PixelConvertType::ComponentType;
+
+  PixelComponentType zeroComponent = NumericTraits<PixelComponentType>::ZeroValue(zeroComponent);
```
dzenanz: This definition seems self-referential.

blowekamp: I am not sure why/how it works, but I got this block of code from the ResampleImageFilter.

N-Dekker: Interesting PR, @blowekamp I'll have a look!

N-Dekker: Sorry, I won't be able to finish the review today.

Modules/Filtering/Smoothing/test/itkMeanImageFilterTest.cxx
```diff
@@ -20,10 +20,14 @@
 #include "itkMeanImageFilter.h"
 #include "itkTextOutput.h"
 #include "itkTestingMacros.h"
+#include "itkVectorImage.h"
 
 int
 itkMeanImageFilterTest(int, char *[])
 {
+  using VectorImageType = itk::VectorImage<float, 3>;
+  itk::MeanImageFilter<VectorImageType, VectorImageType>::New();
```
N-Dekker: @blowekamp It would be really great if you could add some more elaborate tests for vector images than just doing `MeanImageFilter::New()` 😃 Just testing the filter for a few small vector images. If you like, there's a GTest for this filter as well.

blowekamp: Working on it now...

N-Dekker:
FYI, I just observed a small but reproducible performance penalty (~5%) from the current version of this PR (compared to the current master), on an `itk::Image<char, 3>` of 512x512x512 voxels.

blowekamp:
> FYI, I just observed a small but reproducible performance penalty (~5%) from the current version of this PR (compared to the current master), on an `itk::Image<char, 3>` of 512x512x512 voxels. My test was similar to [#1365 (comment)](https://github.com/InsightSoftwareConsortium/ITK/pull/1365#issue-333232475) (but with a different image size)

I could only attribute that change would be due to reduced vectorization due to the reuse of the "sum" variable.

I updated the patch to include test, and pull out more variables from the loop to avoid the memory allocation associated with the variable length vectors on a per pixel basis. So this very well will have even worse performance. 🥂 😁

N-Dekker:
> I updated the patch to include test, and pull out more variables from the loop to avoid the memory allocation associated with the variable length vectors on a per pixel basis. So this very well will have even worse performance. 🥂 😁

Thanks for the test, Bradley, but indeed, the performance for a regular `itk::Image<char, 3>` is even worse now (getting close to 10%, compared to master) 😢

blowekamp:
I ran your test profiling code on my Ubuntu system with gcc (`gcc (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0`) with a 512^3 image. I get the same performance with or without vector with 1 thread or many. 

Without vector support 1 thread:
```
2428:  Duration: 7.71682 second(s)
2428:  Duration: 7.63101 second(s)
2428:  Duration: 7.69318 second(s)
2428:  Duration: 7.66172 second(s)
2428:  Duration: 7.57051 second(s)
2428:  Duration: 7.63201 second(s)
2428:  Duration: 7.62434 second(s)
2428:  Duration: 7.60287 second(s)
```

With vector support 1 thread:
```
2428:  Duration: 7.77182 second(s)
2428:  Duration: 7.69744 second(s)
2428:  Duration: 7.69085 second(s)
2428:  Duration: 7.62326 second(s)
2428:  Duration: 7.62163 second(s)
2428:  Duration: 7.68784 second(s)
2428:  Duration: 7.62271 second(s)
2428:  Duration: 7.62398 second(s)
```

Then converted the timing code to use a VectorImage of 512x512x128 with 4 components. At first it timed at 25s. Nearly 4x the time of the same number of scalars. I updated the code ( here too) and  got it down to ~8.8s. So that is a 3x improvement by removing some variable length vector temporaries created by static casts.

Modules/Filtering/Smoothing/test/itkMeanImageFilterTest.cxx
blowekamp: test added to the GTest file above.

blowekamp:
@N-Dekker The ~10% performance reduction you reported does not seem to effect the gcc I'm using. And it may be specific to your compiler.

How can we move this forward:
- Should I commit my performance profiling code so we can see the results on the CI systems? 
- Do you want to look to see if you can make some tweaks so there is less of a penalty for the generalized code here?
- Is any amount of performance reduction not acceptable?

I suspect that the reason for the reduction is due to the addition of the `sumAcc` outside the for loop. This is needed for the VariableLength to prevent a memory allocation at every look iterations. However, I suspect the compile with the 10% performance reduction is not longer able to vectorize the loop for scalars.

N-Dekker:
@blowekamp Nice to see that you're still further improving this pull request. I hope to have a closer look later today, or tomorrow (if you can wait).

> * Is any amount of performance reduction not acceptable?

I would not say "not acceptable". There may be a trade-off between functionality and performance. 

Personally I do not need `VectorImage` support for `MeanImageFilter`. So I would rather not "pay" for it, by reduced performance for regular ITK images. But if you think adding `VectorImage` support to this filter would make other users happy, well, it's up to you to decide  😃  

A previous commit of mine b657c12998509b2fad1244650d40d27ac87128e1 appeared to produce 40% performance gain for this filter, so if we would now lose 10%, there would still be about 30% left  😄

N-Dekker:
I think performance loss could be avoided by having separate template specializations of internal helper functions for regular images and vector-images. But honestly I haven't yet figured out how 😸

blowekamp:
@N-Dekker Thanks for the suggestion. I agree that we should have two implementations for the critical section.

I generally prefer to use function overloading before meta-programming as generally easier to maintain and I fine it simpler and clearer. The new method overloaded with a pointer to the pixel type should suffice. Here is the first similar example I found:

https://github.com/InsightSoftwareConsortium/ITK/blob/b176feac1ba33daa178018944174b375f7c66686/Modules/Filtering/ImageIntensity/include/itkMaskImageFilter.h#L100-L113

Modules/Filtering/Smoothing/include/itkMeanImageFilter.hxx
blowekamp: The line is no longer in the latest version.

N-Dekker:
Busy reviewing... please wait 😄

Modules/Filtering/Smoothing/include/itkMeanImageFilter.hxx
```diff
@@ -98,6 +104,56 @@ MeanImageFilter<TInputImage, TOutputImage>::GenerateDataInSubregion(
...
+  const VariableLengthVector<TValueType> *)
+{
+  const auto neighborhoodSize = static_cast<double>(neighborhoodOffsets.size());
+
+  using namespace Experimental;
+  auto neighborhoodRange = ShapedImageNeighborhoodRange<const InputImageType, TPixelAccessPolicy>(
+    inputImage, Index<InputImageDimension>(), neighborhoodOffsets);
+  auto outputIterator = ImageRegionRange<OutputImageType>(outputImage, imageRegion).begin();
+
+  InputRealType sumRef(inputImage.GetNumberOfComponentsPerPixel());
```
N-Dekker: The name `sumRef` is not really clear to me. Can you explain it to me?

Modules/Filtering/Smoothing/include/itkMeanImageFilter.hxx
```diff
@@ -98,6 +104,56 @@ MeanImageFilter<TInputImage, TOutputImage>::GenerateDataInSubregion(
...
+  // basis.
+  const InputRealType &               zeroSum = sumRef;
+  InputRealType                       sum;
+  typename OutputImageType::PixelType out;
+
+  for (const auto & index : ImageRegionIndexRange<InputImageDimension>(imageRegion))
+  {
+    neighborhoodRange.SetLocation(index);
+
+    sum = zeroSum;
```
N-Dekker: Is this (`sum = zeroSum`) equivalent to `sum.Fill(NumericTraits<PixelComponentType>::Zero)`?

blowekamp: Yup, I think that would remove the need for zeroSum, and sumRef. It's good to remove the hoops that had to be jumped through to get the code to work all the pixel types.

blowekamp:
@N-Dekker Thanks for the review, and the comments to improve the code!

PR #1522, opened 2019-12-28
BUG: Handle CellData in Decimation Filter

DVigneault:
Currently, cell data is not properly handled in the
itk::SquaredEdgeLengthDecimation filter.  This patch adds entries to
the cell data array as faces are added in the Zip filter, and then
deletes unreferenced entries in a new member function added to
itk::Mesh.

## PR Checklist

- [X] Adds the License notice to new files.
- [X] Adds tests and baseline comparison (quantitative).
- [X] Adds Documentation.

### DIFF
--- a/Modules/Core/Mesh/include/itkMesh.h
+++ b/Modules/Core/Mesh/include/itkMesh.h
@@ -359,6 +359,12 @@ public:
   const CellDataContainer *
   GetCellData() const;
 
+  /** Delete entries in m_CellDataContainer which do not have a corresponding
+   * entry in m_CellsContainer.
+   */
+  void
+  DeleteUnusedCellData();
+
 #if !defined(ITK_WRAPPING_PARSER)
   /**
    * Set/get the BoundaryAssignmentsContainer for a given dimension.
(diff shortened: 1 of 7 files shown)


DVigneault:
Here's an example of a high-poly input mesh (left) and a low-poly decimated mesh (right), where the cell faces are colored according to their corresponding cell data entries using VTK.

Decimation

DVigneault:
Not sure what accounts for the `ITK.Windows` failure--looking at the output, it appears that none of the test executables could be found and so no tests were run.

DVigneault:
@dzenanz I changed the signature of the test from `(int, char**)` to `(int, char*[])`, which seems to have fixed the failure on `ITK.Windows`.  Unfortunately, there is now a failure on `ITK.macOS`, though this seems to be unrelated (something to do with the `vnl_symmetric_eigensystem`).  Here are the failing lines from Azure Pipelines:

````
109: Start Testing test_symmetric_eigensystem:
...
109:  Test 009: Specialised version is faster                         --> **FAILED**
````

````
Start Testing test_symmetric_eigensystem:
...
Test 009: Specialised version is faster                         --> **FAILED**
````

dzenanz:
That's a known flaky test.

dzenanz:
It looks like this introduces some problem with TBB multi threader:
https://open.cdash.org/testDetails.php?test=837479003&build=6285497
We could wait another nightly test cycle to be sure.

PR #1530, opened 2020-01-02
STYLE: Updates for ImageIORegionGTest, DiscreteGaussianImageFilter

thewtex:
With clang-format.

### DIFF
--- a/Modules/Core/Common/test/itkImageIORegionGTest.cxx
+++ b/Modules/Core/Common/test/itkImageIORegionGTest.cxx
@@ -81,7 +81,7 @@ Expect_CopyAssignableToSelf(const T & value)
   T self(value);
 
   // Slight contortion to avoid self-assignment warning with 'self = self'.
-  const T& referenceToSelf = self;
+  const T & referenceToSelf = self;
   self = referenceToSelf;
 
   EXPECT_EQ(self, value);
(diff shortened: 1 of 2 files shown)


Modules/Core/Common/test/itkImageIORegionGTest.cxx
```diff
@@ -81,7 +81,7 @@ Expect_CopyAssignableToSelf(const T & value)
   T self(value);
 
   // Slight contortion to avoid self-assignment warning with 'self = self'.
-  const T& referenceToSelf = self;
+  const T & referenceToSelf = self;
```
N-Dekker: Personally I would prefer the original, to indicate more clearly that the ampersand is part of the type:

    T& reference = variable;

Whereas I like to see the ampersand at the right, when it means "address-of":

    ptr = &variable;

And in the middle on binary AND operation:

    result = byte1 & byte2; 

My two cents.... and best wishes for 2020! 🎉

Modules/Filtering/Smoothing/include/itkDiscreteGaussianImageFilter.h
```diff
@@ -356,9 +356,8 @@ class ITK_TEMPLATE_EXPORT DiscreteGaussianImageFilter : public ImageToImageFilte
   /** Boundary condition use for the intermediate filters */
   RealBoundaryConditionPointerType m_RealBoundaryCondition;
 
- /** Default boundary condition use for the intermediate filters */
+  /** Default boundary condition use for the intermediate filters */
```
N-Dekker: This indentation (2 spaces) obviously looks better than the original (1 space) 👍  Do you know why the original wasn't fixed by clang-format initially?

N-Dekker:
Can anyone please explain why clang-format did not do this initially, for ImageIORegionGTest, DiscreteGaussianImageFilter?

PR #1531, opened 2020-01-04
ENH: Adding function to SOToImageFilter simplify output definition

aylward:
Added SetMatchImage( img ) function.   The size, spacing, directions,
etc of the image passed to this function are used to define the
output image format of the filter.

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkSpatialObjectToImageFilter.h
+++ b/Modules/Core/SpatialObjects/include/itkSpatialObjectToImageFilter.h
@@ -90,6 +90,18 @@ public:
   const InputSpatialObjectType *
   GetInput(unsigned int idx);
 
+  /** Generate an output image that matches the origin,
+   * size, direction, and spacing of this image. */
+  template <class TMatchImage>
+  void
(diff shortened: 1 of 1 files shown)


Modules/Core/SpatialObjects/include/itkSpatialObjectToImageFilter.h
```diff
@@ -90,6 +90,18 @@ class ITK_TEMPLATE_EXPORT SpatialObjectToImageFilter : public ImageSource<TOutpu
   const InputSpatialObjectType *
   GetInput(unsigned int idx);
 
+  /** Generate an output image that matches the origin,
+   * size, direction, and spacing of this image. */
+  template <class TMatchImage>
+  void
+  SetMatchImage(TMatchImage * matchImage)
```
dzenanz: The equivalent thing is called `ReferenceImage` in resample filter, it would be good if we kept that name here too.

aylward: Fixed!

PR #1532, opened 2020-01-05
BUG: itkMetaImageIO.cxx supports more MetaDataDictionary field types

aylward:
MetaDataDictionary now routinely includes itkMatrix and std::vector
data items.  These changes add support to MetaIO for those types when
parsing non-standard MetaDataDictionary fields.


thewtex:
LGTM :-)

PR #1536, opened 2020-01-06
ENH: When using cmake > 3.12 use FindPython3 module

hjmjohnson:
The FindPython3 module is more robust for ensuring that python 3 or greater is found.

The FindPythonInterp and FindPythonLib modules are deprecated
in cmake greater than 3.12.0

The GoogleTest framework uses the deprecated FindPythonInterp, so this
the patch uses the FindPython3 module, and then uses that more robust
information to set the FindPythonInterp cmake variables so that
wrapping, google tests, documentation generation, etc.. all use the
same found a version of Python 3.

## PR Checklist
- [X] [Makes design changes]

### DIFF
--- a/Testing/ContinuousIntegration/AzurePipelinesMacOSPython.yml
+++ b/Testing/ContinuousIntegration/AzurePipelinesMacOSPython.yml
@@ -44,12 +44,14 @@ jobs:
 
     - bash: |
         python_executable=`which python3`
+        python_root_dir=$(dirname $(dirname ${python_executable}))
         cat > dashboard.cmake << EOF
         set(CTEST_BUILD_CONFIGURATION "Release")
         set(CTEST_CMAKE_GENERATOR "Ninja")
         set(BUILD_NAME_SUFFIX "-Python")
         set(dashboard_cache "
-          PYTHON_EXECUTABLE:FILEPATH=${python_executable}
+          Python3_ROOT_DIR:PATH=${python_root_dir}
+          Python3_EXECUTABLE:FILEPATH=${python_executable}
           DISABLE_MODULE_TESTS:BOOL=ON
           BUILD_SHARED_LIBS:BOOL=OFF
           BUILD_EXAMPLES:BOOL=OFF
--- a/CMake/ITKModuleDoxygen.cmake
+++ b/CMake/ITKModuleDoxygen.cmake
@@ -6,7 +6,23 @@
 # python is needed to verify the presence of the module name in the doxygen header
 # Don't require it to not force the developers to install python to be able to build
 # ITK. The tests will simply not be run if python is not available.
-find_package(PythonInterp)
+# Prefer to use more robust FindPython3 module if greater than cmake 3.12.0
+if("${CMAKE_VERSION}" VERSION_LESS_EQUAL "3.12.0")
+  # Use of PythonInterp and PythonLibs is dprecated since cmake version 3.12.0
+
+  # configure python (find PythonInterp first, as of cmake 3.1)
+  find_package(PythonInterp)
+  # Check for supported python versions
+  if(PYTHON_VERSION_STRING VERSION_LESS 3.5)
+    message(FATAL_ERROR "Python versions less than 3.5 are not supported. Python version: \"${PYTHON_VERSION_STRING}\".")
+  endif()
+
+  ## For forward compatibility with cmake 3.12.0 or greater
+  set(Python3_EXECUTABLE ${PYTHON_EXECUTABLE})
+else()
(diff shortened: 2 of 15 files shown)


Testing/ContinuousIntegration/AzurePipelinesMacOSPython.yml
```diff
@@ -44,12 +44,14 @@ jobs:
...
     - bash: |
         python_executable=`which python3`
+        python_root_dir=$(dirname $(dirname ${python_executable}))
         cat > dashboard.cmake << EOF
         set(CTEST_BUILD_CONFIGURATION "Release")
         set(CTEST_CMAKE_GENERATOR "Ninja")
         set(BUILD_NAME_SUFFIX "-Python")
         set(dashboard_cache "
-          PYTHON_EXECUTABLE:FILEPATH=${python_executable}
+          Python3_ROOT_DIR:PATH=${python_root_dir}
```
dzenanz: This change is not done for Windows. Intentional or oversight?

hjmjohnson: @dzenanz The variable "PYTHON_EXECUTABLE" did not exist in the current windows path.  Perhaps it should, but I don't know the windows environment well enough to know what to set the variable to.  There is also no `which python3` shell script for windows either.

Can this intermediate refactoring be done separately from a windows/python refactoring?

dzenanz: Sure! Just making sure it is not an oversight.

blowekamp:
I am not getting a configuration error on my Mac system because it only finds py2 and not py3. Python 3 is not required for the Doxygen, or HeaderTests. OSX, and and many linux based systems still come with Python 2 by default. And this patch breaks the builds.

hjmjohnson:
> I am _now_ getting a configuration error on my Mac system because it only finds py2 and not py3. Python 3 is not required for the Doxygen, or HeaderTests. OSX, and and many linux based systems still come with Python 2 by default. And this patch breaks the builds.

@blowekamp I thought that it was already established that we would only support Python3 (for wrapping at least).  From a maintenance standpoint, it would be easier if we simply require python3 for any use of python in ITK.  I think that Doxygen and HeaderTests are both "developer-level" features that are OK to require Python3 (even if they don't strictly need them).    I know we could provide support for python2, but I want to ensure that the extra complexity and maintenance burden will be worth that level of complexity.  (If Doxygen and not wrapping, then python2, else python3 is required).

My motivation for this change is that during the configuration of cmake with ITK I was finding that several versions of Python were being found and in some cases overwriting the intended version.  I found it really confusing when multiple versions of the python executable were identified for different parts of the code base.

I'm willing to re-add python2 support for developer builds that only use HeaderTests or Doxygen tests if that is the consensus, but I think it is important to have a conversation regarding the pros/cons of the added complexity.  Let me know what you think.

Hans

blowekamp:
On a new configuration this is the error I'm getting:
```
CMake Error at /Users/blowekamp/Applications/CMake.app/Contents/share/cmake-3.14/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find Python3 (missing: Python3_EXECUTABLE Interpreter)
  /Users/blowekamp/Applications/CMake.app/Contents/share/cmake-3.14/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  /Users/blowekamp/Applications/CMake.app/Contents/share/cmake-3.14/Modules/FindPython/Support.cmake:1170 (find_package_handle_standard_args)
  /Users/blowekamp/Applications/CMake.app/Contents/share/cmake-3.14/Modules/FindPython3.cmake:174 (include)
  CMake/ITKModuleDoxygen.cmake:23 (find_package)
  CMake/ITKModuleMacros.cmake:6 (include)
  CMakeLists.txt:142 (include)
```

I don't even have Doxygen enabled.

The options are:
1. When python 2 or no python is detect, gracefully disable the "developer-level" features.
2. Allow python 2 and python3 for the developer-level feature. This would require changing from CMake "FindPython3" to "FindPython".

hjmjohnson:
@blowekamp could you please try the following:
 At line 547 of the primary CMakeLists.txt file: Remove "REQUIRED" from the `      find_package(Python3 COMPONENTS Interpreter REQUIRED)`.  I think that was a mistake.

We would also need to add 'if(Python3_Interpreter_FOUND ` to line 550.

hjmjohnson:
@blowekamp I'm working on a solution.  it is almost ready for option #1 given above.  I've removed python3 from one of my (slower) computers and I am testing the build now.

PR #1537, opened 2020-01-07
Revert "BUG: SetScale() and SetSkew() did not recompute offset: ScaleSkewVersorT"

blowekamp:
Reverts InsightSoftwareConsortium/ITK#1523, to preserve existing transform files the transform numerics and number of parameters must not change.


dzenanz:
Python configuration failure is unrelated.

PR #1539, opened 2020-01-07
ENH: Adding ComposeScaleSkewVersor 3D Transform

aylward:
This transform uses matrix compose to combine matrix defined by
versor, scale, and skew parameters.  Implementing the
jacobian of this transform is left as an exercise for future
developers.

### DIFF
--- a/Modules/Core/Transform/include/itkComposeScaleSkewVersor3DTransform.h
+++ b/Modules/Core/Transform/include/itkComposeScaleSkewVersor3DTransform.h
@@ -145,10 +145,8 @@ public:
   void
   SetIdentity() override;
 
-  /** This method computes the Jacobian matrix of the transformation.
-   * given point or vector, returning the transformed point or
-   * vector. The rank of the Jacobian will also indicate if the
-   * transform is invertible at this point. */
+  /* This function is not implemented for this transform.  An exception
+   *   is thrown if this function is called. */
   void
   ComputeJacobianWithRespectToParameters(const InputPointType & p, JacobianType & jacobian) const override;
 
--- a/Modules/Core/Transform/include/itkComposeScaleSkewVersor3DTransform.h
+++ b/Modules/Core/Transform/include/itkComposeScaleSkewVersor3DTransform.h
@@ -1,0 +1,196 @@
+/*=========================================================================
+ *
+ *  Copyright Insight Software Consortium
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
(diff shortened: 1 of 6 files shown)


dzenanz:
The new class is not Python-wrapped.

aylward:
Added python wrapping.

Modules/Core/Transform/include/itkComposeScaleSkewVersor3DTransform.hxx
```diff
@@ -0,0 +1,372 @@
...
+
+  jacobian[0][2] = 2.0 * (-2 * vzw * px + (vzz - vww) * py + (vxw - vyz) * pz) / vw;
+  jacobian[1][2] = 2.0 * ((vww - vzz) * px - 2 * vzw * py + (vyw + vxz) * pz) / vw;
+  jacobian[2][2] = 2.0 * ((vxw + vyz) * px + (vyw - vxz) * py) / vw;
+
+  jacobian[0][3] = 1.0;
+  jacobian[1][4] = 1.0;
+  jacobian[2][5] = 1.0;
+
+  jacobian[0][6] = px;
```
blowekamp: The scale dependent Jacobian values an incorrect as well.

Maybe this whole implementation should be removed and only leave the exception?

Modules/Core/Transform/test/itkComposeScaleSkewVersor3DTransformTest.cxx
```diff
@@ -0,0 +1,554 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+/**
+ *
+ *  This program illustrates the use of Similarity3DTransform
```
blowekamp: Similarity3DTransform->ComposeScaleSkewVersor3DTransform

Modules/Core/Transform/include/itkComposeScaleSkewVersor3DTransform.h
```diff
@@ -0,0 +1,196 @@
...
+
+  itkGetConstReferenceMacro(Skew, SkewVectorType);
+
+  void
+  SetIdentity() override;
+
+  /** This method computes the Jacobian matrix of the transformation.
+   * given point or vector, returning the transformed point or
+   * vector. The rank of the Jacobian will also indicate if the
+   * transform is invertible at this point. */
```
blowekamp: Change the comment to indicate that this is not implemented and an exception will be thrown.

Modules/Core/Transform/test/itkComposeScaleSkewVersor3DTransformTest.cxx
dzenanz: The same change should probably be applied to the original class too?

Modules/Core/Transform/include/itkComposeScaleSkewVersor3DTransform.hxx
aylward: Removed and appropriate comments added.

Modules/Core/Transform/test/itkComposeScaleSkewVersor3DTransformTest.cxx
aylward: Changed in new and in old/original class too.

Modules/Core/Transform/include/itkComposeScaleSkewVersor3DTransform.h
aylward: Done.

PR #1547, opened 2020-01-08
DOC: Remove CONTRIBUTING note about ApplyClangFormat action

thewtex:
As of 2020-01-08, this action [currently does not work from forked
repositories](https://github.community/t5/GitHub-Actions/Token-permissions-for-forks-once-again/td-p/33839)
because the [*GITHUB_TOKEN* has limited
permissions](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/authenticating-with-the-github_token#permissions-for-the-github_token). Branches need to be created on the upstream repository.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -281,10 +281,6 @@ one, but further back in your history-, and the* Pro Git: Rebasing *resource on
 taking all the changes that were committed on one branch and replaying them on
 another one.*)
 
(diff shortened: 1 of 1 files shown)


thewtex:
The action still could be quite useful in remote module repositories, but it is going to have limited applicability in the ITK repository until GitHub fixes the permissions.

PR #1551, opened 2020-01-09
BUG: Prevent integer underflow in CropImageFilter

blowekamp:
Check that the crop size is not greater that the image size to prevent
integer underflow.

## PR Checklist

### DIFF
--- a/Modules/Filtering/ImageGrid/include/itkCropImageFilter.h
+++ b/Modules/Filtering/ImageGrid/include/itkCropImageFilter.h
@@ -111,6 +111,9 @@ protected:
   void
   GenerateOutputInformation() override;
 
+  void
+  VerifyInputInformation() ITKv5_CONST override;
+
 private:
   SizeType m_UpperBoundaryCropSize;
   SizeType m_LowerBoundaryCropSize;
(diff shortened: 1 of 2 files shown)


Modules/Filtering/ImageGrid/include/itkCropImageFilter.hxx
```diff
@@ -56,6 +58,25 @@ CropImageFilter<TInputImage, TOutputImage>::GenerateOutputInformation()
   Superclass::GenerateOutputInformation();
 }
 
+
+template <typename TInputImage, typename TOutputImage>
+void
+CropImageFilter<TInputImage, TOutputImage>::VerifyInputInformation()
+{
+
```
dzenanz: This empty line is not needed

blowekamp:
This issue was discovered/reported on the 3D Slicer discussion forum:
https://discourse.slicer.org/t/cropping-an-image-in-jupyter-notebook-fails/9741/10

Modules/Filtering/ImageGrid/include/itkCropImageFilter.h
```diff
@@ -111,6 +111,9 @@ class ITK_TEMPLATE_EXPORT CropImageFilter : public ExtractImageFilter<TInputImag
   void
   GenerateOutputInformation() override;
 
+  void
+  VerifyInputInformation() ITKv5_CONST override;
```
dzenanz: `ITKv5_CONST` is meant for user code which need to build against both v4 and v5. Here we should use plain `const`, I believe.

blowekamp: The const-ness was added in ITKv5 in the base class and it's toggled with the compatibility flags. So it's needed.
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkProcessObject.h#L789

dzenanz: Human memory is faulty 😄

PR #1471, opened 2019-12-06
ENH: Consistently use strongly typed enums

thewtex:
To finish addressing https://github.com/InsightSoftwareConsortium/ITK/issues/1049

PR #1557, opened 2020-01-10
ENH: Changed NIFTI IO to support different flavors of Analyze readers

vfonov:
Changed ITK NIFTI Reader, according to the discussion from https://discourse.itk.org/t/nifti-file-orientation-change-between-2019-08-26-and-itk-v5-1rc01/2594 .
New behaviour is to reject Analyze files, unless one of the flavours is specified:

Reject Analyze files (default) to avoid confusion:
`imageIO->SetLegacyAnalyze75Mode(itk::NiftiImageIO::AnalyzeReject);`
Use NIFTI reader behaviour introduced in ITK v4 (ignore orientation code, flip dimensions if step size is negative):
`imageIO->SetLegacyAnalyze75Mode(itk::NiftiImageIO::AnalyzeITK4);`
Simulate SPM reader, mode similar to depricated Analyze reader from ITK3 (interpret orientation code, flip dimensions if step size is negative):
`imageIO->SetLegacyAnalyze75Mode(itk::NiftiImageIO::AnalyzeSPM);`
Simulate FSL reader (respect orientation code, ignore negative step size):
`imageIO->SetLegacyAnalyze75Mode(itk::NiftiImageIO::AnalyzeFSL);`

## PR Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -295,6 +295,23 @@ if (ITKV3_COMPATIBILITY)
   message(FATAL_ERROR "ITKV3_COMPATIBILITY is removed starting in ITK version 5.0, this option is no longer valid.")
 endif()
 
+# Chose flavour of Analyze reader
+set(ITK_NIFTI_IO_ANALYZE_FLAVOR_DOC_STRING "Which flavour of Analyze reader to use by default")
+
+set(ITK_NIFTI_IO_ANALYZE_FLAVOR "ITK4Warning" CACHE STRING ${ITK_NIFTI_IO_ANALYZE_FLAVOR_DOC_STRING})
+
+set_property(CACHE ITK_NIFTI_IO_ANALYZE_FLAVOR PROPERTY STRINGS ITK4Warning ITK4 SPM FSL Reject)
+
+if( NOT ("${ITK_NIFTI_IO_ANALYZE_FLAVOR}" STREQUAL "ITK4Warning" OR
+         "${ITK_NIFTI_IO_ANALYZE_FLAVOR}" STREQUAL "ITK4" OR
+         "${ITK_NIFTI_IO_ANALYZE_FLAVOR}" STREQUAL "SPM" OR
+         "${ITK_NIFTI_IO_ANALYZE_FLAVOR}" STREQUAL "FSL" OR
+         "${ITK_NIFTI_IO_ANALYZE_FLAVOR}" STREQUAL "Reject"
+         ) )
+  set(ITK_NIFTI_IO_ANALYZE_FLAVOR "ITK4Warning" CACHE STRING ${ITK_NIFTI_IO_ANALYZE_FLAVOR_DOC_STRING} FORCE)
+endif()
+
+
 #-----------------------------------------------------------------------------
 # ITK build classes that are in the review process
 # ITK_USE_REVIEW is deprecated, only kept for backward compatibility
--- a/Modules/IO/ImageBase/include/itkIOTestHelper.h
+++ b/Modules/IO/ImageBase/include/itkIOTestHelper.h
@@ -31,12 +31,18 @@ class IOTestHelper
 public:
   template <typename TImage>
   static typename TImage::Pointer
-  ReadImage(const std::string & fileName, const bool zeroOrigin = false)
+  ReadImage(const std::string &           fileName,
+            const bool                    zeroOrigin = false,
+            typename ImageIOBase::Pointer imageio = nullptr)
   {
     using ReaderType = itk::ImageFileReader<TImage>;
 
     typename ReaderType::Pointer reader = ReaderType::New();
     {
+      if (imageio)
+      {
+        reader->SetImageIO(imageio);
+      }
       reader->SetFileName(fileName.c_str());
       try
       {
@@ -69,13 +75,18 @@ public:
 
   template <typename ImageType, typename ImageIOType>
   static void
-  WriteImage(typename ImageType::Pointer & image, const std::string & filename)
+  WriteImage(typename ImageType::Pointer & image,
+             const std::string &           filename,
+             typename ImageIOType::Pointer imageio = nullptr)
   {
 
     using WriterType = itk::ImageFileWriter<ImageType>;
     typename WriterType::Pointer writer = WriterType::New();
 
-    typename ImageIOType::Pointer imageio(ImageIOType::New());
+    if (imageio.IsNull())
+    {
+      imageio = ImageIOType::New();
+    }
 
     writer->SetImageIO(imageio);
 
--- a/Modules/IO/NIFTI/include/itkNiftiImageIO.h
+++ b/Modules/IO/NIFTI/include/itkNiftiImageIO.h
@@ -27,6 +27,30 @@
 
 namespace itk
 {
+
+
+/** \class Analyze75Flavor
+ * \ingroup ITKIONIFTI
+ * Enum used to define way to treat legacy Analyze75 files
+ */
+enum class Analyze75Flavor : uint8_t
+{
+  /** Behavior introduced in ITK4.0 by NIFTI reader interpreting Analyze files */
+  AnalyzeITK4 = 4,
+  /** Will ignore orientation code and negative pixel dimensions */
+  AnalyzeFSL = 3,
+  /** Will ignore orientation code and respect negative pixel dimensions */
+  AnalyzeSPM = 2,
+  /** Same as AnalyzeITK4 but will show warning about deprecated file format (Default)*/
+  AnalyzeITK4Warning = 1,
+  /** Reject Analyze files as potentially wrong  */
+  AnalyzeReject = 0
+};
+
+/** Define how to print enumerations */
+extern ITKIONIFTI_EXPORT std::ostream &
+                         operator<<(std::ostream & out, const Analyze75Flavor value);
+
 /** \class NiftiImageIO
  *
  * \author Hans J. Johnson, The University of Iowa 2002
@@ -70,6 +94,7 @@ public:
     /** Some other file format, or file system error. */
     OtherOrError = -1,
   };
+
 
   /** Reads the file to determine if it can be read with this ImageIO implementation,
    * and to determine what kind of file it is (Analyze vs NIfTI). Note that the value
@@ -131,10 +156,10 @@ public:
 
   /** A mode to allow the Nifti filter to read and write to the LegacyAnalyze75 format as interpreted by
    * the nifti library maintainers.  This format does not properly respect the file orientation fields.
-   * By default this is set to true.
+   * By default this is set by configuration option ITK_NIFTI_IO_ANALYZE_FLAVOR
    */
-  itkSetMacro(LegacyAnalyze75Mode, bool);
-  itkGetConstMacro(LegacyAnalyze75Mode, bool);
+  itkSetMacro(LegacyAnalyze75Mode, Analyze75Flavor);
+  itkGetConstMacro(LegacyAnalyze75Mode, Analyze75Flavor);
 
 protected:
   NiftiImageIO();
@@ -188,8 +213,10 @@ private:
 
   IOComponentType m_OnDiskComponentType{ UNKNOWNCOMPONENTTYPE };
 
-  bool m_LegacyAnalyze75Mode{ true };
+  Analyze75Flavor m_LegacyAnalyze75Mode;
 };
+
+
 } // end namespace itk
 
 #endif // itkNiftiImageIO_h
--- a/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
+++ b/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
@@ -21,6 +21,8 @@
 #include "itkSpatialOrientationAdapter.h"
 #include <nifti1_io.h>
 
+#include "itkNiftiImageIOConfigurePrivate.h"
+
 namespace itk
 {
 //#define ITK_USE_VERY_VERBOSE_NIFTI_DEBUGGING
@@ -446,6 +448,7 @@ public:
 NiftiImageIO::NiftiImageIO()
   : m_NiftiImageHolder(new NiftiImageProxy(nullptr))
   , m_NiftiImage(*m_NiftiImageHolder.get())
+  , m_LegacyAnalyze75Mode(ITK_NIFTI_IO_ANALYZE_FLAVOR_DEFAULT)
 
 {
   this->SetNumberOfDimensions(3);
@@ -791,9 +794,9 @@ NiftiImageIO::DetermineFileType(const char * FileNameToRead)
   //      == 1 for a nifti file (header+data in 1 file)
   //      == 0 for an analyze 7.5 file,
   //      == -1 for an error,
-  const int image_FTYPE = is_nifti_file(FileNameToRead);
+  const int imageFTYPE = is_nifti_file(FileNameToRead);
 
-  return static_cast<NiftiImageIO::FileType>(image_FTYPE);
+  return static_cast<NiftiImageIO::FileType>(imageFTYPE);
 }
 
 // This method will only test if the header looks like an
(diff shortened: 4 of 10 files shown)


vfonov:
Oops, looks like I left some debug code...

Modules/IO/NIFTI/include/itkNiftiImageIO.h
```diff
@@ -71,6 +71,19 @@ class ITKIONIFTI_EXPORT NiftiImageIO : public ImageIOBase
     OtherOrError = -1,
   };
 
+  /** Way to treat legacy Analyze75 files */
+  enum Analyze75Flavour
```
dzenanz: We are in the process of switching to `enum class` style, so new code should be using that.

Modules/IO/NIFTI/include/itkNiftiImageIO.h
```diff
@@ -188,7 +201,7 @@ class ITKIONIFTI_EXPORT NiftiImageIO : public ImageIOBase
 
   IOComponentType m_OnDiskComponentType{ UNKNOWNCOMPONENTTYPE };
 
-  bool m_LegacyAnalyze75Mode{ true };
+  Analyze75Flavour m_LegacyAnalyze75Mode{ AnalyzeReject }; // AnalyzeITK4
```
dzenanz: remove debugging comment

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -808,12 +808,14 @@ NiftiImageIO ::CanReadFile(const char * FileNameToRead)
   //      == 0 for an analyze 7.5 file,
   //      == -1 for an error,
   const int image_FTYPE = is_nifti_file(FileNameToRead);
-
+  // DEBUG
```
dzenanz: Either remove or wrap into `if (this->GetDebug())`

Modules/IO/NIFTI/test/itkNiftiImageIOTest.cxx
```diff
@@ -125,13 +125,17 @@ itkNiftiImageIOTest(int ac, char * av[])
...
-    ImageType::Pointer input;
+    ImageType::Pointer         input;
+    itk::NiftiImageIO::Pointer imageIO = itk::NiftiImageIO::New();
+    // enable old behavior of NIFTI reader
+    imageIO->SetLegacyAnalyze75Mode(itk::NiftiImageIO::AnalyzeITK4);
     for (int imagenameindex = 1; imagenameindex < ac; imagenameindex++)
     {
-      // std::cout << "Attempting to read " << av[imagenameindex] << std::endl;
+      // DEBUG
+      std::cout << "Attempting to read " << av[imagenameindex] << std::endl;
```
dzenanz: `if (this->GetDebug())` or remove

Modules/IO/NIFTI/test/itkNiftiImageIOTest.cxx
```diff
@@ -125,13 +125,17 @@ itkNiftiImageIOTest(int ac, char * av[])
...
     using ImageType = itk::Image<unsigned char, 3>;
-    ImageType::Pointer input;
+    ImageType::Pointer         input;
+    itk::NiftiImageIO::Pointer imageIO = itk::NiftiImageIO::New();
+    // enable old behavior of NIFTI reader
+    imageIO->SetLegacyAnalyze75Mode(itk::NiftiImageIO::AnalyzeITK4);
     for (int imagenameindex = 1; imagenameindex < ac; imagenameindex++)
     {
-      // std::cout << "Attempting to read " << av[imagenameindex] << std::endl;
+      // DEBUG
```
dzenanz: remove comment

Modules/IO/NIFTI/test/itkNiftiImageIOTestHelper.cxx
```diff
@@ -88,6 +88,7 @@ TestNiftiByteSwap(const std::string & prefix)
   catch (const itk::ExceptionObject & e)
   {
     e.Print(std::cerr);
+    // DEBUG
```
dzenanz: remove comment

Modules/IO/NIFTI/include/itkNiftiImageIO.h
vfonov: enum class doesn't seem to work with itkSetMacro

dzenanz: Matt just made a big commit in enum class PR, see this comment:
https://github.com/InsightSoftwareConsortium/ITK/pull/1471#issuecomment-573191946

jhlegarreta:
Sorry to chime in. I have loosely followed the related conversation on discourse. Forgive me if my comments are not totally accurate. I have a few comments:
- Although necessary, this patch set breaks the current API, if I have read the code appropriately (e.g. the changes to the` itk::NiftiImageIO::GetLegacyAnalyze75Mode,` `itk::NiftiImageIO:SetLegacyAnalyze75Mode` methods). Thus, this should be guarded with some macro to make it compatible with previous version until a new major version is released, if we are to follow the usual practice.
- I have not read the new tests, but given the breadth of the discussion on discourse, and the  implications it has for third-party applications, and the time scale of the first considerations made by Hans, I think having this thoroughly tested would facilitate the understanding and rationale for the reasons of the change. This may imply using new or more image tests to demonstrate the behavior Andras showed on discourse.
- May be the rationale or the discussion should be cast into the `ITK5MigrationGuide.md` document or the class headers? Not sure if editing the `ITK5MigrationGuide.md` is appropriate since on my mind the `ITKxMigrationGuide.md`-like files are to be edited only at a major version change.

I understand that this is a much needed change, that it implies a great deal of work and testing, and that some third-party applications require this to be merged. Thanks to all that strived to improve and make more robust the current behavior.

vfonov:
I can make default behavior to still accept Analyze files using ITK4 convention , but display a warning about depricated file format.

vfonov:
New behaviour is to reject warn about Analyze files, unless one of the flavours is specified:

- Reject Analyze files to avoid confusion: 
  `imageIO->SetLegacyAnalyze75Mode(itk::NiftiImageIO::AnalyzeReject);`
- Use NIFTI reader behaviour introduced in ITK v4 (ignore orientation code, flip dimensions if step size is negative): 
  `imageIO->SetLegacyAnalyze75Mode(itk::NiftiImageIO::AnalyzeITK4);`
- Simulate SPM reader, mode similar to depricated Analyze reader from ITK3 (interpret orientation code, flip dimensions if step size is negative): 
  `imageIO->SetLegacyAnalyze75Mode(itk::NiftiImageIO::AnalyzeSPM);`
- Simulate FSL reader (respect orientation code, ignore negative step size): 
  `imageIO->SetLegacyAnalyze75Mode(itk::NiftiImageIO::AnalyzeFSL);`

Modules/IO/NIFTI/include/itkNiftiImageIO.h
```diff
@@ -27,6 +27,30 @@
...
+ */
+enum class Analyze75Flavour : uint8_t
+{
+  /** Behavioir introduced in ITK4.0 by NIFTI reader interpreting Analyze files */
+  AnalyzeITK4 = 4,
+  /** Will ignore orientation code and negative pixel dimensios */
+  AnalyzeFSL = 3,
+  /** Will ignore orientation code and respect negative pixel dimensios */
+  AnalyzeSPM = 2,
+  /** Same as AnalyzeITK4 but will show warning about depricated file format (Default)*/
```
lassoan:
```suggestion
  /** Same as AnalyzeITK4 but will show warning about deprecated file format (Default)*/
```

Modules/IO/NIFTI/include/itkNiftiImageIO.h
```diff
@@ -133,8 +158,8 @@ class ITKIONIFTI_EXPORT NiftiImageIO : public ImageIOBase
    * the nifti library maintainers.  This format does not properly respect the file orientation fields.
    * By default this is set to true.
```
lassoan: Need to update the method documentation: default is not "true", also add some more details (summary of the discussions, reason for default value choice)

vfonov: oops , looks like i forgot to set the default to AnalyzeITK4Warning

vfonov:
Now there is a new cmake configuration option `ITK_ANALYZE_FLAVOUR` with possible values `ITK4Warning` `ITK4` `SPM` `FSL` `Reject`

CMakeLists.txt
```diff
@@ -295,6 +295,23 @@ if (ITKV3_COMPATIBILITY)
   message(FATAL_ERROR "ITKV3_COMPATIBILITY is removed starting in ITK version 5.0, this option is no longer valid.")
 endif()
 
+# Chose flavour of Analyze reader
+set(ITK_ANALYZE_FLAVOUR_DOC_STRING "Which flavour of Analyze reader to use by default")
+
+set(ITK_ANALYZE_FLAVOUR "ITK4Warning" CACHE STRING ${ITK_ANALYZE_FLAVOUR_DOC_STRING})
```
lassoan: ITK_ANALYZE_FLAVOUR name is a bit too generic (people who don't know that Analyze is an image file format may misinterpret it). Also, "flavour" is British spelling, while in ITK uses US speclling (it is always spelled as "flavor" in ITK). So, an option name like these could work better:
- ITK_ANALYZE_IMAGE_IO_FLAVOR
- ITK NIFTI_IO_ANALYZE_FLAVOR
- ITK NIFTI_IMAGE_IO_ANALYZE_FLAVOR

Modules/IO/NIFTI/include/itkNiftiImageIO.h
```diff
@@ -27,6 +27,30 @@
...
+  /** Will ignore orientation code and negative pixel dimensios */
+  AnalyzeFSL = 3,
+  /** Will ignore orientation code and respect negative pixel dimensios */
+  AnalyzeSPM = 2,
+  /** Same as AnalyzeITK4 but will show warning about deprecated file format (Default)*/
+  AnalyzeITK4Warning = 1,
+  /** Reject Analyze files as potentially wrong  */
+  AnalyzeReject = 0
+};
+
```
seanm: Couple of typos: Behavioir, dimensios

Can you add a comment indicating which behaves like ITK 3/4 AnalyzeImageIO.

vfonov: It's a tricky question: ITK3 allowed negative spacing, the most similar behavior of Analyze reader would be SPM

Modules/IO/ImageBase/include/itkIOTestHelper.h
```diff
@@ -69,13 +73,16 @@ class IOTestHelper
...
+             const std::string &           filename,
+             typename ImageIOType::Pointer imageio = nullptr)
   {
 
     using WriterType = itk::ImageFileWriter<ImageType>;
     typename WriterType::Pointer writer = WriterType::New();
 
-    typename ImageIOType::Pointer imageio(ImageIOType::New());
+    if (imageio.IsNull())
+      imageio = ImageIOType::New();
```
seanm: IIRC ITK style requires mandatory braces, even for single line `if`s.  There are a few other instances too.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -991,6 +993,25 @@ NiftiImageIO::SetImageIOMetadataFromNIfTI()
 void
 NiftiImageIO ::ReadImageInformation()
 {
+  const int image_FTYPE = is_nifti_file(this->GetFileName());
```
seanm: IIRC ITK style doesn't allow underscores in variable names.

lassoan:
Thanks a lot @vfonov it looks good to me, I have no further comments!

Modules/IO/NIFTI/include/itkNiftiImageIO.h
```diff
@@ -27,6 +27,30 @@
 
 namespace itk
 {
+
+
+/** \class Analyze75Flavor
+ * \ingroup ITKIONIFTI
+ * Enum used to define way to treat legacy Analyze75 files
+ */
+enum class Analyze75Flavor : uint8_t
```
dzenanz: I guess wrapping is needed for this class
`itkNiftiImageIO: warning(4): ITK type not wrapped, or currently not known: itk::Analyze75Flavor`
A recent example of how to do it:
https://github.com/InsightSoftwareConsortium/ITKMontage/commit/c65afeef7a5f35e0975d2fc07908552a8b6ca908#diff-517bbaeb1b2d54fcf3da514456be02e7R5

Modules/IO/ImageBase/include/itkIOTestHelper.h
```diff
@@ -69,13 +75,18 @@ class IOTestHelper
...
+  WriteImage(typename ImageType::Pointer & image,
+             const std::string &           filename,
+             typename ImageIOType::Pointer imageio = nullptr)
   {
 
     using WriterType = itk::ImageFileWriter<ImageType>;
     typename WriterType::Pointer writer = WriterType::New();
 
-    typename ImageIOType::Pointer imageio(ImageIOType::New());
+    if (imageio.IsNull())
```
seanm: Is there a difference between doing `if (imageio.IsNull())` vs line 42?

Modules/IO/NIFTI/include/itkNiftiImageIO.h
```diff
@@ -131,10 +156,10 @@ class ITKIONIFTI_EXPORT NiftiImageIO : public ImageIOBase
 
   /** A mode to allow the Nifti filter to read and write to the LegacyAnalyze75 format as interpreted by
    * the nifti library maintainers.  This format does not properly respect the file orientation fields.
-   * By default this is set to true.
+   * By default this is set by configuration option ITK_NIFTI_IO_ANALYZE_FLAVOR
    */
-  itkSetMacro(LegacyAnalyze75Mode, bool);
-  itkGetConstMacro(LegacyAnalyze75Mode, bool);
+  itkSetMacro(LegacyAnalyze75Mode, Analyze75Flavor);
```
seanm: We are all ok with changing this public API's signature?

Modules/IO/ImageBase/include/itkIOTestHelper.h
dzenanz: I believe `if (imageio)` is equivalent to `if (imageio.IsNotNull())`, so opposite to `if (imageio.IsNull())`.

Modules/IO/NIFTI/include/itkNiftiImageIO.h
dzenanz: I think this API was added since last release, so only people living on `master` will feel the effect of this change.

seanm:
OK +1 for me then

blowekamp:
I believe all the interested parties have agreed with this flexible implementation. It looks good to go to me.

Modules/IO/NIFTI/include/itkNiftiImageIO.h
seanm: In fact, no, that API is present in ITK 4.13.2, and probably earlier too.

dzenanz: OK, thanks for clarification Sean. I am fine with changing the signature, as people should have moved away from Analyze files. But we should document the change in [migration guide](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/ITK5MigrationGuide.md).

dzenanz: @vfonov and @seanm, can one of you make a PR with migration guide update?

seanm: In fact, if you've "moved away from Analyze files" then you might already be deliberately calling SetLegacyAnalyze75Mode(false) to make sure you *don't* open Analyze files. Now that won't compile.

dzenanz: Well, a PR which restores the old API, and gives the new API a new name is a possible alternative. As I never used Analyze files, I don't have much preference for one over the other. I believe most people just don't use IOs directly, and those who do, mostly use GDCMImageIO.

lassoan: It may be useful to break the API to make sure that developers notice the potential change in behavior. A comment can be added near where the build error occurs that points to corresponding section in the migration guide.

PR #1561, opened 2020-01-13
BUG: SetLabelForUndecidedPixels value ignored.

zivy:
The user requested label for undecided values was ignored and the
maxLabel+1 was always used. The test also had a bug in its ground
truth values.


blowekamp:
The patch addresses the bug reported here: 
SimpleITK/SimpleITK#943

LGTM

PR #1571, opened 2020-01-20
Granular progress updates

dzenanz:
Implementing a feature requested in https://discourse.itk.org/t/no-progress-reporting-in-itk5/2630/.

## PR Checklist

- [X] Adds Documentation.

### DIFF
--- a/Modules/Core/Common/include/itkProcessObject.h
+++ b/Modules/Core/Common/include/itkProcessObject.h
@@ -37,6 +37,9 @@
 #include <map>
 #include <set>
 #include <algorithm>
+#include <atomic>
+#include <chrono>
+#include <thread>
 
 namespace itk
 {
@@ -334,6 +337,41 @@ public:
    */
   void
   UpdateProgress(float progress);
+
+  /** Indicate to this class that some of the pixels/items have been processed.
+   * It will decide whether this warrants a progress report or not. */
+  void
+  PixelsProcessed(SizeValueType numberOfPixels);
+
+  /** Should MultiThreader report coarse-grained progress? */
+  itkGetConstReferenceMacro(MultiThreaderUpdatesProgress, bool);
+  virtual void
+  SetMultiThreaderUpdatesProgress(bool arg)
+  {
+    this->m_MultiThreaderUpdatesProgress = arg;
+    // calling this->Modified to update timestamp messes with the pipeline execution
+  }
+
+  /** How many pixels or items this class will work on. */
+  itkGetConstReferenceMacro(TotalPixelCount, SizeValueType);
+  virtual void
+  SetTotalPixelCount(SizeValueType totalPixelCount)
+  {
+    this->m_TotalPixelCount = totalPixelCount;
+    // calling this->Modified to update timestamp messes with the pipeline execution
+  }
+
+  /** How frequently would you like progress to be reported. */
+  itkGetConstReferenceMacro(DesiredUpdateInterval, std::chrono::microseconds);
+  virtual void
+  SetDesiredUpdateInterval(const std::chrono::microseconds microseconds)
+  {
+    if (this->m_DesiredUpdateInterval != microseconds)
+    {
+      this->m_DesiredUpdateInterval = microseconds;
+      this->Modified();
+    }
+  }
 
   /** \brief Bring this filter up-to-date.
    *
@@ -914,6 +952,16 @@ private:
   /** These support the progress method and aborting filter execution. */
   bool  m_AbortGenerateData;
   float m_Progress;
+
+  /** Internal variables used in new-style progress reporting. */
+  bool                                  m_MultiThreaderUpdatesProgress = true;
+  std::atomic<SizeValueType>            m_CurrentPixelCount{ 0 };
+  IdentifierType                        m_TotalPixelCount{ 0 };
+  IdentifierType                        m_PreviousProgressReport{ 0 }; // pixel count during last report
+  std::chrono::steady_clock::time_point m_LastProgressReport = std::chrono::steady_clock::now(); // last report time
+  std::chrono::microseconds             m_DesiredUpdateInterval =
+    std::chrono::milliseconds(100);                               // TODO: have a statically settable global default;
+  std::thread::id m_CallingThreadID = std::this_thread::get_id(); // we can only report from this thread
 
   /** Support processing data in multiple threads. Used by subclasses
    * (e.g., ImageSource). */
--- a/Modules/Core/Common/include/itkProgressReporter.h
+++ b/Modules/Core/Common/include/itkProgressReporter.h
@@ -110,6 +110,48 @@ protected:
   float           m_InitialProgress;
   float           m_ProgressWeight;
 };
+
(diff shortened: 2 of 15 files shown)


dzenanz:
This is not quite finished yet, but it should be working, and can be tested.

lassoan:
Thanks a lot for the quick action. I'm at the Slicer project week and the program is very busy, but I'll test this as soon as I can.

blowekamp:
Dzenan, thanks for the quick turn around.

IMHO, it is a poor design choice to integrate so much for the progress reporting into the `ProcessObject`. Functionality should be in separate objects. It also appears that the "ProcessObject" now has two sets of variable to keep track of progress. This is confusing and will lead to difficulties in maintanance. Please take some time to think about how to separate the keeping track of progress (The process object's responsibility), mapping of ranges ( The progress reports) and when a progress even should occour.

What is your reasoning to consolidate the old `ProgressReporter` and the new style progress reporting? One object one behavior. 

Yes, these are difficult design problems to address, and likely to tedious to address online. Hopefully, these thought help moving the implementation to a more module design.

dzenanz:
> What is your reasoning to consolidate the old ProgressReporter and the new style progress reporting?

The reasoning is to bring improved (time-limited progress reporting) to old code which uses `ProgressReporter`.

Modules/Core/Common/include/itkProcessObject.h
```diff
@@ -880,6 +908,18 @@ class ITKCommon_EXPORT ProcessObject : public Object
...
+  bool m_MultiThreaderUpdatesProgress =
+    true; // coarse-grained progress reporting does not require cooperation of filters
+  std::atomic<SizeValueType>            m_CurrentPixelCount{ 0 };
+  IdentifierType                        m_TotalPixelCount{ 0 };
+  IdentifierType                        m_PreviousProgressReport{ 0 }; // pixel count during last report
+  std::chrono::steady_clock::time_point m_LastProgressReport =
+    std::chrono::steady_clock::now(); // time during last report
+  std::chrono::milliseconds m_DesiredUpdateInterval =
+    std::chrono::milliseconds(100);                               // TODO: have a statically settable global default;
+  std::thread::id m_CallingThreadID = std::this_thread::get_id(); // we can only report from this thread
```
blowekamp: This would be the thread of construction. I believe the progress is only for the `GenerateData` method. The best place to get the thread id of the pipeline execution might be right before the GenerateData method is called in `itk::ProcessObject::UpdateOutputData`. Then the variable could be made private and some accessor like "IsPipelineThread" could be used.

Modules/Core/Common/include/itkProgressReporter.h
```diff
@@ -62,14 +62,36 @@ class ITKCommon_EXPORT ProgressReporter
...
-                   float           initialProgress = 0.0f,
+                   SizeValueType   numberOfUpdates,
+                   float           initialProgress,
                    float           progressWeight = 1.0f);
 
+
+  /** New-style progress reporting. */
+  ProgressReporter(ProcessObject * filter,
+                   ThreadIdType,
+                   SizeValueType numberOfPixels,
```
blowekamp: This in on location it will be very confusing combining the different behaviors the new-style expects the total number of "pixels" while the old expect the number just for the current threads.

Modules/Core/Common/include/itkProcessObject.h
dzenanz: Good catch.

Modules/Core/Common/include/itkProgressReporter.h
dzenanz: I now think it is best if we keep the new progress reporting separate. Trying to make it backwards compatible and elegantly plugging it into existing infrastructure seems like too much work.

dzenanz: I am thinking about calling the new class `ProgressReporter2`, and defining it in the same header/source file as the original `ProgressReporter`. What do you think @blowekamp?

blowekamp:
- There is still a confusion duplication of progress in the ProcessObject. The current progress is float, which doesn't have atomic operations. ☹️ However, given that that range is fixed at [0,1], and we have encapsulated the private IVAR enough so that we could change the internal representation to a uint64, and have the setters getters treat it as a fixed point floating number.

- Introducing the concept of pixels at the ProcessObject is not appropriate. And event for the the images, it's not always the correct term/representation. This mapping should be occurring the the ProgressReporter.

I can volunteer to quickly do the first Item above, or I could take a couple days and come up with my own version.

dzenanz:
I would much prefer if you took over this PR. I don't think there is hurry to get it done, aside that it would be good if it makes it into 5.1 final.

The "TODO" about configurable default duration: I planned to put a statically settable default value into MultiThreaderBase, because it already has logic for ensuring there is only one instance of [globals](https://github.com/InsightSoftwareConsortium/ITK/blob/bcd762e7bd743341c2dce84852ced49d8292f293/Modules/Core/Common/src/itkMultiThreaderBase.cxx#L54-L104) around, even when used from DLLs in Python.

dzenanz:
Work continues in #1575.

lassoan:
Thanks for working on this. Let me know if you would like me to test it.

dzenanz:
While Brad works on the other PR, you can test this one.

lassoan:
Thank you @dzenanz I've tested this in Slicer on a number of filters and it worked perfectly!

blowekamp:
@lassoan What first's did you test?
 @dzenanz Is this update suppose to increase reporting for filters ( or child classes ) not modified in this PR?

lassoan:
I tested dzenanz:granularProgressUpdates branch with latest Slicer master and I got nice continuous progress update.

dzenanz:
Smoother progress updates are not expected for filters untouched by this PR. But a lot of filters inherit from unary functor.

thewtex (commented):
Awesome work.

Modules/Core/Common/include/itkProcessObject.h
```diff
@@ -915,6 +953,16 @@ class ITKCommon_EXPORT ProcessObject : public Object
   bool  m_AbortGenerateData;
   float m_Progress;
 
+  /** Internal variables used in new-style progress reporting. */
+  bool                                  m_MultiThreaderUpdatesProgress = true;
```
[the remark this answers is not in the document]
dzenanz: we would have no progress updates for a lot of filters

Modules/Core/Common/include/itkProgressReporter.h
```diff
@@ -110,6 +110,48 @@ class ITKCommon_EXPORT ProgressReporter
   float           m_InitialProgress;
   float           m_ProgressWeight;
 };
+
+/** \class ProgressReporter2
+ * \brief Implements new-style progress tracking for a filter.
+ *
```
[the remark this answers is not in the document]
dzenanz: I did not want to write a detailed description of it before the work is finalized. And as Brad is working on a different way to reintroduce fine-grained progress, this PR might be abandoned.

dzenanz:
Closing in favor of #1575.

PR #1573, opened 2020-01-22
PERF: Refactor RelabelComponentImageFilter

blowekamp:
Parallelized applying the output production step. Modernizes the use
of STL data structures and loops. Stream-lined cases. Save iterators
used for searching to exploit coherence in images.

## PR Checklist

### DIFF
--- a/Modules/Segmentation/ConnectedComponents/include/itkRelabelComponentImageFilter.h
+++ b/Modules/Segmentation/ConnectedComponents/include/itkRelabelComponentImageFilter.h
@@ -21,6 +21,7 @@
 #include "itkInPlaceImageFilter.h"
 #include "itkImage.h"
 #include <vector>
+#include <mutex>
 
 namespace itk
 {
@@ -131,14 +132,14 @@ public:
   itkNewMacro(Self);
 
   /** Type used as identifier for the different component labels. */
-  using LabelType = IdentifierType;
+  using LabelType = InputPixelType;
 
   /** Type used to count number of pixels in objects. */
   using ObjectSizeType = SizeValueType;
 
   /** Get the number of objects in the image. This information is only
    * valid after the filter has executed. */
-  itkGetConstMacro(NumberOfObjects, LabelType);
+  itkGetConstMacro(NumberOfObjects, SizeValueType);
 
   using ObjectSizeInPixelsContainerType = std::vector<ObjectSizeType>;
   using ObjectSizeInPhysicalUnitsContainerType = std::vector<float>;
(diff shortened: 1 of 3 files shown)


Modules/Segmentation/ConnectedComponents/include/itkRelabelComponentImageFilter.h
```diff
@@ -266,37 +266,7 @@ class ITK_TEMPLATE_EXPORT RelabelComponentImageFilter : public InPlaceImageFilte
 
   struct RelabelComponentObjectType
```
dzenanz: This struct now has only one member. Can it be eliminated entirely and replaced by just `ObjectSizeType`?

Modules/Segmentation/ConnectedComponents/include/itkRelabelComponentImageFilter.hxx
```diff
@@ -45,176 +48,179 @@ template <typename TInputImage, typename TOutputImage>
...
-  m_SizeOfObjectsInPhysicalUnits.clear();
-  m_SizeOfObjectsInPhysicalUnits.resize(m_NumberOfObjects);
-  int NumberOfObjectsRemoved = 0;
-  for (i = 0, vit = sizeVector.begin(); vit != sizeVector.end(); ++vit)
+  int             NumberOfObjectsRemoved = 0;
+  OutputPixelType outputLabel = 0;
+  for (auto & sizeVectorItem : sizeVector)
   {
     // if we find an object smaller than the minimum size, we
     // terminate the loop.
```
dzenanz: This comment is no longer true.

Modules/Segmentation/ConnectedComponents/include/itkRelabelComponentImageFilter.h
blowekamp: It could. But as it's used in a `std::pair`, the extra verbosity of `second.m_SizeInPixels` vs just `second` is useful.

blowekamp:
I did a performance testing on 512^image of unsigned short pixels. The tests are included in the PR. Neither of the a "real world" but represent two extreme cases. Here are the results:

|   | old | old - no progress| pool 16/64 | pool 1/1 | platform 16
| ------------- | ------------- |-|-| -|-|
| zeros  |  1.41s | 0.5s | 0.15s | 0.3s | 0.098s
| random  | 37.1s  | 35s| 3.5s | 27s | 3.28s

*old* - the prior version
*old* - no progress* - the prior version with `CompetePixel` commented out
*pool 16/64* - the pool multithreaded with 16 threads and 64 work units
*pool 1/1* - the pool multithread with just one thread
*platform 16* - the platform multithreaded with 16 threads 

The results are from a Mac Pro workstation with 8 core + HyperThreading.

The new revision has performance improvement across the board even with single threads. The platform multithread has a 11x and 14x speed up vs the old single threaded implementation.

The thread pool using 64 work units has more progress events than the old implementation which may account for some of the performance difference when compared to the platform threader.

PR #1575, opened 2020-01-24
Progress reporting revisionv5

blowekamp:
This is early in the hacking of revising the progress reporting system. The initial goal of this approach was to have one ProcessObject variable with the progress. This should allow all the different types of progress reporting to work together.

The progress reporting in the Multithreaders was commented out. After looking at the code a little I am wonder is much of progress reporting in the multithreads should be replaced by a separate progress reporting object.

### DIFF
--- a/Modules/Core/Common/include/itkProcessObject.h
+++ b/Modules/Core/Common/include/itkProcessObject.h
@@ -29,17 +29,20 @@
 #define itkProcessObject_h
 
 #include "itkDataObject.h"
-#include "itkDomainThreader.h"
-#include "itkMultiThreaderBase.h"
 #include "itkObjectFactory.h"
 #include "itkNumericTraits.h"
+#include "itkThreadSupport.h"
 #include <vector>
 #include <map>
 #include <set>
 #include <algorithm>
+#include <thread>
 
 namespace itk
 {
+
+class MultiThreaderBase;
+
 /** \class ProcessObject
  * \brief The base class for all process objects (source,
  *        filters, mappers) in the Insight data processing pipeline.
@@ -303,19 +306,13 @@ public:
    * the filter has completed execution.  The ProgressEvent is NOT
    * invoked.
    * This method is deprecated because filters should not be calling
-   * SetProgress directly but should be using UpdateProgress instead.
-   * We avoid the use of the itkSetClampMacro because that macro calls
-   * Modified on the filter, which will cause the filter to rerun even
-   * if it doesn't need to.
-   * Thus, we implement the SetClampMacro directly without the call to
-   * Modified. */
+   * SetProgress directly but should be using UpdateProgress or IncrementProgress instead.
+   */
 #if !defined(ITK_LEGACY_REMOVE)
   void
   SetProgress(float progress)
   {
-    // Clamp the value to be between 0 and 1.
-    m_Progress = std::max(progress, 0.0f);
-    m_Progress = std::min(m_Progress, 1.0f);
+    m_Progress = progressFloatToFixed(progress);
   }
 #endif
 
@@ -324,16 +321,33 @@ public:
    * The progress is a floating number in [0,1] with 0 meaning no
    * progress and 1 meaning the filter has completed execution.
    */
-  itkGetConstReferenceMacro(Progress, float);
+  virtual float
+  GetProgress() const
+  {
+    return progressFixedToFloat(m_Progress);
+  }
 
   /** \brief Update the progress of the process object.
    *
    * Sets the Progress ivar to amount and invokes any observers for
    * the ProgressEvent. The parameter amount should be in [0,1] and is
    * the cumulative (not incremental) progress.
+   *
+   * Multiple thread should not call this method, since they may have computed different values and increments may get
+   * lost.
    */
   void
   UpdateProgress(float progress);
+
+  /** \brief Increment the progress of the process object.
+   *
+   * Atomically add the the current progress and may invoke observers of the ProgressEvent. Progress is represented in
+   * [0.0,1.0] or percentage. This method will invoke the ProgressEvent when called by the same as the pipeline.
+   *
+   * Multiple threads may call this method and the total progress will be atomically incremented.
+   */
+  void
+  IncrementProgress(float increment);
 
   /** \brief Bring this filter up-to-date.
    *
@@ -873,6 +887,43 @@ protected:
   virtual void
   RestoreInputReleaseDataFlags();
 
+  /**
+   * When true, the MultiThreader will report course grain progress. If set to false, a progress must be explicitly
+   * updated in derived filters.
+   */
+  itkGetConstMacro(ThreaderUpdateProgress, bool);
+  itkBooleanMacro(ThreaderUpdateProgress);
+  virtual void
+  SetThreaderUpdateProgress(bool arg);
+
+
+  /**
+   * Internal method to convert internal integer progress to float [0.0, 1.0]
+   */
+  static inline constexpr float
+  progressFixedToFloat(uint32_t fixed)
+  {
+    return double(fixed) / double(std::numeric_limits<uint32_t>::max());
+  };
+
+  /**
+   * Internal method convert floating point progress [0.0, 1.0] to internal integer representation. Values outside the
+   * [0.0, 1.0] range are clamped.
+   */
+  static inline uint32_t
+  progressFloatToFixed(float f)
+  {
+    if (f <= 0.0f)
+    {
+      return 0;
+    }
+    if (f >= 1.0f)
+    {
+      return std::numeric_limits<uint32_t>::max();
+    }
+    return static_cast<double>(f) * std::numeric_limits<uint32_t>::max();
+  };
+
   /** These ivars are made protected so filters like itkStreamingImageFilter
    * can access them directly. */
 
@@ -912,13 +963,18 @@ private:
   NameSet m_RequiredInputNames;
 
   /** These support the progress method and aborting filter execution. */
-  bool  m_AbortGenerateData;
-  float m_Progress;
+  bool                  m_AbortGenerateData;
+  std::atomic<uint32_t> m_Progress;
+
+
+  std::thread::id m_UpdateThreadID;
 
   /** Support processing data in multiple threads. Used by subclasses
    * (e.g., ImageSource). */
-  MultiThreaderType::Pointer m_MultiThreader;
-  ThreadIdType               m_NumberOfWorkUnits;
+  itk::SmartPointer<MultiThreaderType> m_MultiThreader;
+  ThreadIdType                         m_NumberOfWorkUnits;
+
+  bool m_ThreaderUpdateProgress{ true };
 
   /** Memory management ivars */
   bool m_ReleaseDataBeforeUpdateFlag;
--- a/Modules/Core/Common/include/itkTotalProgressReporter.h
+++ b/Modules/Core/Common/include/itkTotalProgressReporter.h
@@ -1,0 +1,132 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+#ifndef itkTotalProgressReporter_h
+#define itkTotalProgressReporter_h
+
+#include "itkIntTypes.h"
+#include "itkProcessObject.h"
+
+namespace itk
+{
+/** \class TotalProgressReporter
+ *  \brief A progress reporter for concurrent threads.
+ *
+ * Each thread should construct their own instance of the class. The ProcessObject::IncrementProgress method will be
+ * called to update the progress from all threads. The ProcessObject's method will automatically create ProgressEvents
+ * when the pipeline invoking thread updates the progress.
+ *
+ * All threads concurrently contribute parts of the progress to the *total* number of pixels for all threads. This
+ * report will update the progress after a sufficient number of pixel to meet the numberOfUpdates requirement between
+ * all threads. Also when the object is deconstructed, all remaining pixels will increment the progress.
+ *
+ *
+ * \ingroup ITKCommon
+ */
+class ITKCommon_EXPORT TotalProgressReporter
+{
+public:
+  ITK_DISALLOW_COPY_AND_ASSIGN(TotalProgressReporter);
+
+  /** \brief Construct a TotalProgressReporter
+   *
+   * @param filter - the ProcessObject which whose progress will be updated. If nullptr then no updates will occur.
+   * @param totalNumberOfPixels - the number of pixels between all threads and chunks that will be updated
+   * @param numberOfUpdates - controls how often the ProcessObject's progress will be incremented
+   * @param progressWeight - A percentage of the filters progress, this total number of pixels will contribute
+   */
+  TotalProgressReporter(ProcessObject * filter,
+                        SizeValueType   totalNumberOfPixels,
+                        SizeValueType   numberOfUpdates = 100,
+                        float           progressWeight = 1.0f);
+
+  /** Destructor sets progress to 1 because the filter has finished.  */
+  ~TotalProgressReporter();
+
+  /** Called by a filter once per pixel.  */
+  void
+  CompletedPixel()
+  {
+    // Inline implementation for efficiency.
+    if (--m_PixelsBeforeUpdate == 0)
+    {
+      m_PixelsBeforeUpdate = m_PixelsPerUpdate;
+      m_CurrentPixel += m_PixelsPerUpdate;
+
+      if (m_Filter)
+      {
+        m_Filter->IncrementProgress(m_PixelsPerUpdate * m_InverseNumberOfPixels * m_ProgressWeight);
+
+        // all threads needs to check the abort flag
+        if (m_Filter->GetAbortGenerateData())
+        {
+          std::string    msg;
+          ProcessAborted e(__FILE__, __LINE__);
+          msg += "Object " + std::string(m_Filter->GetNameOfClass()) + ": AbortGenerateDataOn";
+          e.SetDescription(msg);
+          throw e;
+        }
+      }
+    }
+  }
+
+
+  /** Called by a filter when a chunk, region, scan-line, etc. is completed. */
+  void
+  Completed(SizeValueType count)
+  {
+
+    if (count >= m_PixelsBeforeUpdate)
+    {
+      SizeValueType total = static_cast<SizeValueType>(m_PixelsPerUpdate - m_PixelsBeforeUpdate) + count;
+      SizeValueType numberOfUpdates = total / m_PixelsPerUpdate;
+
+      m_PixelsBeforeUpdate = m_PixelsPerUpdate - total % m_PixelsPerUpdate;
+      m_CurrentPixel += numberOfUpdates * m_PixelsPerUpdate;
+
+      if (m_Filter)
+      {
+        m_Filter->IncrementProgress(numberOfUpdates * m_PixelsPerUpdate * m_InverseNumberOfPixels * m_ProgressWeight);
+
+        // all threads needs to check the abort flag
+        if (m_Filter->GetAbortGenerateData())
+        {
+          std::string    msg;
+          ProcessAborted e(__FILE__, __LINE__);
+          msg += "Object " + std::string(m_Filter->GetNameOfClass()) + ": AbortGenerateDataOn";
+          e.SetDescription(msg);
+          throw e;
+        }
+      }
+    }
+    else
+    {
+      m_PixelsBeforeUpdate -= count;
+    }
+  }
+
+protected:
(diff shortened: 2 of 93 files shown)


Modules/Core/Common/include/itkProcessObject.h
```diff
@@ -297,25 +297,35 @@ class ITKCommon_EXPORT ProcessObject : public Object
...
-   * invoked.
-   * This method is deprecated because filters should not be calling
-   * SetProgress directly but should be using UpdateProgress instead.
-   * We avoid the use of the itkSetClampMacro because that macro calls
-   * Modified on the filter, which will cause the filter to rerun even
-   * if it doesn't need to.
-   * Thus, we implement the SetClampMacro directly without the call to
-   * Modified. */
+  static constexpr float
+  progressFixedToFloat(uint32_t f)
```
dzenanz: Why 32-bit integer? We already commonly process images bigger than that. If this is meant to be performant on 32-bit processors, we could use `SizeValueType`. Answering self: 32-bit float has less significant bits and 32-bit integer, so it is a fine replacement.

Modules/Core/Common/src/itkProcessObject.cxx
```diff
@@ -1622,13 +1639,14 @@ void
 ProcessObject ::UpdateOutputData(DataObject * itkNotUsed(output))
 {
   /**
-   * prevent chasing our tail
+   * prevent chasing our tailD
```
dzenanz: accidentally modified

Modules/Core/Common/src/itkProcessObject.cxx
```diff
@@ -1163,8 +1163,17 @@ ProcessObject::IncrementProgress(float increment)
   // Clamp the value to be between 0 and 1.
   uint32_t integerIncrement = progressFloatToFixed(increment);
 
+  uint32_t oldProgress = m_Progress;
+
   m_Progress.fetch_add(integerIncrement);
 
+  if (oldProgress < m_Progress)
```
dzenanz: This check seems tautological: condition should not be met only only if no progress was made, or there are some multiple thread race conditions.

blowekamp: The line has been corrected, and a comment added for its reason. Thank you for point it out.

blowekamp:
@dzenanz  What do you think of reusing progress reported in the `MultiThreaders`?

Modules/Core/Common/include/itkTotalProgressReporter.h
```diff
@@ -0,0 +1,120 @@
...
+ */
+class ITKCommon_EXPORT TotalProgressReporter
+{
+public:
+  ITK_DISALLOW_COPY_AND_ASSIGN(TotalProgressReporter);
+
+
+  TotalProgressReporter(ProcessObject * filter,
+                        SizeValueType   totalNumberOfPixels,
+                        SizeValueType   numberOfUpdates = 100,
```
blowekamp: Comparing to PR #1571, this uses the old "numberOfUpdates" while #1571 uses a pixelsPerUpdate. The later in #1571, would cause fewer updates on smaller images or smaller threaded regions. Additionally, the pixelPerUpdate would have a simpler implementation in the `Complete` methods.

Some logic related to timed progress here could be added too. I'm not sure if the timing interval parameter is to get the progress events to occur more frequently, or less.

Thoughts?

dzenanz:
It looks like a clever re-use of `ProgressReporter`s.

Modules/Core/Common/include/itkTotalProgressReporter.h
dzenanz: The time parameter is to make reports more predictable in time. For some images it would be more frequent, for some less frequent.

blowekamp: In #1571, the minimum update interval is controlled by the time parameter, and the frequency that the main calling thread calls the `UpdateProgress` methods. So #1571 would suppress duplicate calls UpdateProcess call, but during slow filters the time interval could be would not be any faster. For example, median with a really big neighborhood more frequent events will not occur.

Just some thought on the time interval, there are many ways to handle it.

dzenanz: Surely. This was a low implementation effort approach. Another way might be to create a timer which checks the current progress?

The approach in #1571 would work well if filters could predict how slow they would be, and invoke pixel completed appropriately. E.g., median with small radius reports pixel completion every 1000 pixels, while median with very large radius reports every 10 pixels.

dzenanz:
Can this PR be finished before [5.1RC2](https://github.com/InsightSoftwareConsortium/ITK/milestone/13) (scheduled sometime next week)?

blowekamp:
It wasn't clear to me if we were converging on the approaches used here.

I can plan to work on in Monday and Tuesday, but I will likely take a little more time to finish the details.

I was unaware of any rescheduling for the RCs, I had been wondering about the plan.

dzenanz:
The enum refactoring was causing compile problems in a lot of places, and its PR was the big blocker for RC2. With that now merged, we can make the next release soon. This doesn't have to make it into RC2, but we should have one improvement to progress reporting as part of 5.1 final.

blowekamp:
@dzenanz With the pool multi-threader, when the number of work units>1, the invoking thread does only one job then it waits for ALL other jobs. For the current default of 4x work load factor, the first quarter of progress is frequent, but then the main thread will not produce anymore progress events as it is just waiting.

This is an odd implementation in the Pool threader, which remove the invoking thread from being a worker thread after only 1 job. This will not use all the cores with more that one for the load factor.

dzenanz:
This uses all the cores, but indeed the invoked thread only executes the first work unit. The other work units are executed by the thread pool, which by default has as many threads as there are CPU cores.

What we could do instead of simply waiting for other threads, is wait with a timeout, then check/report progress e.g. by invoking `filter->pixelsCompleted(0);`, then wait again.

blowekamp:
> This uses all the cores, but indeed the invoked thread only executes the first work unit. The other work units are executed by the thread pool, which by default has as many threads as there are CPU cores.

Indeed, I was able to confirm this by setting ITK_GLOBAL_DEFAULT_THREADER=POOL, and ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS=2 then observing 3 threads and 300% CPU usage. That is good that it is using all the thread, but it is using more resources that I expect. In a cluster environment, this could get your processes kill by using too many resource.

@lassoan 
This behavior of the invoke thread in the thread pool sheds new light on the progress reporting behavior [reported in discourse](https://discourse.itk.org/t/no-progress-reporting-in-itk5/2630).

blowekamp:
> This uses all the cores, but indeed the invoked thread only executes the first work unit. The other work units are executed by the thread pool, which by default has as many threads as there are CPU cores.

Indeed, I was able to confirm this by setting ITK_GLOBAL_DEFAULT_THREADER=POOL, and ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS=2 then observing 3 threads and 300% CPU usage. That is good that it is using all the thread, but it is using more resources that I expect. In a cluster environment, this could get your processes kill by using too many resource.

@lassoan 
This behavior of the invoke thread in the thread pool sheds new light on the progress reporting behavior [reported in discourse](https://discourse.itk.org/t/no-progress-reporting-in-itk5/2630).

blowekamp:
> What we could do instead of simply waiting for other threads, is wait with a timeout, then check/report progress e.g. by invoking `filter->pixelsCompleted(0);`, then wait again.

@dzenanz The other option is to have the invoking thread participate in the work. This would have consistency behavior with TBB and the Platform multi-threaders. What ever is decided here I think we should do it in a subsequent patch.

This patch is about ready for review and testing.

Modules/Core/Common/include/itkTotalProgressReporter.h
blowekamp: As opposed to timed events, it would be better to recommend users to "poll" the progress every at their chosen time rather than rely on evens occurring at regular intervals. With 3 different multi-threaders getting consistent behavior from the invoking thread would require a bit of refactoring and rewriting.

Modules/Filtering/Smoothing/include/itkMeanImageFilter.hxx
```diff
@@ -34,6 +34,7 @@ template <typename TInputImage, typename TOutputImage>
 MeanImageFilter<TInputImage, TOutputImage>::MeanImageFilter()
 {
   this->DynamicMultiThreadingOn();
+  this->ThreaderUpdateProgressOff();
```
blowekamp: This is why the Mean Python test failed due to no progress reporting.

Modules/Core/Common/include/itkTotalProgressReporter.h
dzenanz: Polling is impossible in case of Slicer CLI modules - the main executable waits for output to cout with progress information, and then updates the GUI accordingly.

blowekamp:
@lassoan It's not needed to test this patching Slicer yet, as similar to #1571 only a the UnaryGeneratorImageFilter base class, and the MedianImageFilter have resorted the fine grain progress reporting.

I'd like to get some consensus of approval for these approaches before restoring granular progress reporting to more filters. And some agreement that the progress reporter should be added back to most of the filters where is was removed from in 4611871c.

Modules/Core/Common/include/itkTotalProgressReporter.h
blowekamp: There are several different ways Slicer modules are executed. For the CLI modules they can be execute as a sub-process or in a loaded library. I don't know what way the currently work by default. And *impossible* is a narrow(?) word. Certainly, solutions exists with varying amount of work. 

For the SimpleFilters module it is quite do able.

The timed events is a *new* feature request, and is much more burden-some than restoring progress to similar to how it was. Both in terms of performance impact on other users and about of refactoring required in the multi-threaders. IMHO the lack of timed updates has not been a problem until ITKv5, so restoring the behavior should be sufficient. Many good implementation of timed events involve an event queue similar to how GUI behave. That is one reason why timed progress updates are likely better handled by the application rather that ITK, because ITK has no event queue.

While #1571 added the timer, it still did not work smoothly with the pool multi-threader because the invoking thread just waited for the other threads after completion the first work unit. So  regular timed updates did not occurred after the first chunk.

This is a complex issue!

dzenanz: Sure, this PR can restore granular progress updates. A possible subsequent PR could tackle update timing.

blowekamp:
Any strong objects to the name `TotalProgressReporter`? The "Total" comes from the concept that the `totalNumberOfPixels` is the total size from all threads, not just the local size.

Also its member function name remains `CompletedPixels` for similarity to the `ProgressReporter`. However, there is nothing in the progress classes which is specific to pixels, it could be mesh vertices, or number of indexes in an array. If we which to generalize it now is the time.

dzenanz:
The class name is fine. `CompletedPixels` is better than most general name, as this will be used the most in image filters. General name would make this un-intuitive for the common case.

Modules/Filtering/Smoothing/test/itkMedianImageFilterTest.cxx
```diff
@@ -36,7 +36,7 @@ itkMedianImageFilterTest(int, char *[])
   random->SetMax(1000.0);
 
   FloatImage2DType::SizeValueType randomSize[2];
-  randomSize[0] = randomSize[1] = 8;
+  randomSize[0] = randomSize[1] = 5120;
```
blowekamp: This change needs to be reverted.

blowekamp: done

blowekamp:
This PR is ready for review and envaluation.

lassoan:
Thank you very much @blowekamp and @dzenanz. I've rebuilt Slicer using this branch and progress reporting worked nicely with all the filters I tried!

Modules/Core/Common/include/itkProcessObject.h
```diff
@@ -314,8 +317,7 @@ class ITKCommon_EXPORT ProcessObject : public Object
   SetProgress(float progress)
   {
     // Clamp the value to be between 0 and 1.
```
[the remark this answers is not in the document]
blowekamp: done

Modules/Core/Common/include/itkProcessObject.h
```diff
@@ -873,6 +892,43 @@ class ITKCommon_EXPORT ProcessObject : public Object
   virtual void
   RestoreInputReleaseDataFlags();
 
+  /**
+   * When true, the MultiThreader will report course grain progress. If set to false, the derived
```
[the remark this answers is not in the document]
blowekamp: done.

Modules/Core/Common/src/CMakeLists.txt
```diff
@@ -58,6 +58,7 @@ set(ITKCommon_SRCS
   itkLogOutput.cxx
   itkLoggerOutput.cxx
   itkProgressAccumulator.cxx
+        itkTotalProgressReporter.cxx
```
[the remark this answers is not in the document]
blowekamp: My editor, for an illogical reason had a strong opinion on that level of indentation. Fixed.

blowekamp:
@dzenanz @thewtex It looks like with the rc3 schedule in a couple days we should try to get this PR in today?

dzenanz:
Yes, merge immediately if CI comes back green.

blowekamp:
It looks like the ITK Python build issue is not related.

imikejackson:
I am trying to compile my project with 5.1 RC3 on macOS Catalina and I am getting the following error:

```[88/1573] Linking CXX shared library Bin/libSimpleITKExplicit-0.9.1.dylib
FAILED: Bin/libSimpleITKExplicit-0.9.1.dylib
: && /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ -Wmost -Wno-four-char-constants -Wno-unknown-pragmas -mfpmath=sse -Wno-inconsistent-missing-override    -O3 -DNDEBUG -arch x86_64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk -mmacosx-version-min=10.10 -dynamiclib -Wl,-headerpad_max_install_names -compatibility_version 1.0.0 -current_version 1.0.0 -o Bin/libSimpleITKExplicit-0.9.1.dylib -install_name @rpath/libSimpleITKExplicit-0.9.1.dylib SIMPLView/SIMPL/SimpleITKExplicit/CMakeFiles/SimpleITKExplicit.dir/SimpleITKExplicit_autogen/mocs_compilation.cpp.o SIMPLView/SIMPL/SimpleITKExplicit/CMakeFiles/SimpleITKExplicit.dir/Explicit/src/sitkExplicitITKComposeImageFilter.cxx.o SIMPLView/SIMPL/SimpleITKExplicit/CMakeFiles/SimpleITKExplicit.dir/Explicit/src/sitkExplicitITKDefaultPixelAccessor.cxx.o SIMPLView/SIMPL/SimpleITKExplicit/CMakeFiles/SimpleITKExplicit.dir/Explicit/src/sitkExplicitITKDefaultVectorPixelAccessor.cxx.o SIMPLView/SIMPL/SimpleITKExplicit/CMakeFiles/SimpleITKExplicit.dir/Explicit/src/sitkExplicitITKImageBase.cxx.o SIMPLView/SIMPL/SimpleITKExplicit/CMakeFiles/SimpleITKExplicit.dir/Explicit/src/sitkExplicitITKImage.cxx.o SIMPLView/SIMPL/SimpleITKExplicit/CMakeFiles/SimpleITKExplicit.dir/Explicit/src/sitkExplicitITKImageRegion.cxx.o SIMPLView/SIMPL/SimpleITKExplicit/CMakeFiles/SimpleITKExplicit.dir/Explicit/src/sitkExplicitITKImageSource.cxx.o SIMPLView/SIMPL/SimpleITKExplicit/CMakeFiles/SimpleITKExplicit.dir/Explicit/src/sitkExplicitITKImageToImageFilter.cxx.o SIMPLView/SIMPL/SimpleITKExplicit/CMakeFiles/SimpleITKExplicit.dir/Explicit/src/sitkExplicitITKImportImageContainer.cxx.o SIMPLView/SIMPL/SimpleITKExplicit/CMakeFiles/SimpleITKExplicit.dir/Explicit/src/sitkExplicitITKInPlaceImageFilter.cxx.o SIMPLView/SIMPL/SimpleITKExplicit/CMakeFiles/SimpleITKExplicit.dir/Explicit/src/sitkExplicitITKLabelMap.cxx.o SIMPLView/SIMPL/SimpleITKExplicit/CMakeFiles/SimpleITKExplicit.dir/Explicit/src/sitkExplicitITKVectorImage.cxx.o SIMPLView/SIMPL/SimpleITKExplicit/CMakeFiles/SimpleITKExplicit.dir/Explicit/src/sitkExplicitITKVectorIndexSelectionCastImageFilter.cxx.o SIMPLView/SIMPL/SimpleITKExplicit/CMakeFiles/SimpleITKExplicit.dir/Explicit/src/sitkExplicitITKImageConstIterator.cxx.o SIMPLView/SIMPL/SimpleITKExplicit/CMakeFiles/SimpleITKExplicit.dir/Explicit/src/sitkExplicitITKImageRegionConstIterator.cxx.o SIMPLView/SIMPL/SimpleITKExplicit/CMakeFiles/SimpleITKExplicit.dir/Explicit/src/sitkExplicitITKImageScanlineConstIterator.cxx.o SIMPLView/SIMPL/SimpleITKExplicit/CMakeFiles/SimpleITKExplicit.dir/Explicit/src/sitkExplicitITKImageScanlineIterator.cxx.o  -Wl,-rpath,/Users/Shared/DREAM3D_SDK/ITK-5.1-Release/lib  /Users/Shared/DREAM3D_SDK/ITK-5.1-Release/lib/libITKCommon-5.1.1.dylib  /Users/Shared/DREAM3D_SDK/ITK-5.1-Release/lib/libitksys-5.1.1.dylib  /Users/Shared/DREAM3D_SDK/ITK-5.1-Release/lib/libITKVNLInstantiation-5.1.1.dylib  /Users/Shared/DREAM3D_SDK/ITK-5.1-Release/lib/libitkvnl_algo-5.1.1.dylib  /Users/Shared/DREAM3D_SDK/ITK-5.1-Release/lib/libitkvnl-5.1.1.dylib  /Users/Shared/DREAM3D_SDK/ITK-5.1-Release/lib/libitkv3p_netlib-5.1.1.dylib  /Users/Shared/DREAM3D_SDK/ITK-5.1-Release/lib/libitknetlib-5.1.1.dylib  /Users/Shared/DREAM3D_SDK/ITK-5.1-Release/lib/libitkvcl-5.1.1.dylib  -lm  -lm && :
  "itk::ProcessObject::IncrementProgress(float)", referenced from:
      itk::ComposeImageFilter<itk::Image<double, 2u>, itk::VectorImage<double, 2u> >::DynamicThreadedGenerateData(itk::ImageRegion<2u> const&) in sitkExplicitITKComposeImageFilter.cxx.o
      itk::ComposeImageFilter<itk::Image<double, 3u>, itk::VectorImage<double, 3u> >::DynamicThreadedGenerateData(itk::ImageRegion<3u> const&) in sitkExplicitITKComposeImageFilter.cxx.o
      itk::ComposeImageFilter<itk::Image<float, 2u>, itk::VectorImage<float, 2u> >::DynamicThreadedGenerateData(itk::ImageRegion<2u> const&) in sitkExplicitITKComposeImageFilter.cxx.o
      itk::ComposeImageFilter<itk::Image<float, 3u>, itk::VectorImage<float, 3u> >::DynamicThreadedGenerateData(itk::ImageRegion<3u> const&) in sitkExplicitITKComposeImageFilter.cxx.o
      itk::ComposeImageFilter<itk::Image<int, 2u>, itk::VectorImage<int, 2u> >::DynamicThreadedGenerateData(itk::ImageRegion<2u> const&) in sitkExplicitITKComposeImageFilter.cxx.o
      itk::ComposeImageFilter<itk::Image<int, 3u>, itk::VectorImage<int, 3u> >::DynamicThreadedGenerateData(itk::ImageRegion<3u> const&) in sitkExplicitITKComposeImageFilter.cxx.o
      itk::ComposeImageFilter<itk::Image<long, 2u>, itk::VectorImage<long, 2u> >::DynamicThreadedGenerateData(itk::ImageRegion<2u> const&) in sitkExplicitITKComposeImageFilter.cxx.o
      ...
  "itk::ProcessObject::SetThreaderUpdateProgress(bool)", referenced from:
      vtable for itk::ComposeImageFilter<itk::Image<double, 2u>, itk::VectorImage<double, 2u> > in sitkExplicitITKComposeImageFilter.cxx.o
      vtable for itk::ComposeImageFilter<itk::Image<double, 3u>, itk::VectorImage<double, 3u> > in sitkExplicitITKComposeImageFilter.cxx.o
      vtable for itk::ComposeImageFilter<itk::Image<float, 2u>, itk::VectorImage<float, 2u> > in sitkExplicitITKComposeImageFilter.cxx.o
      vtable for itk::ComposeImageFilter<itk::Image<float, 3u>, itk::VectorImage<float, 3u> > in sitkExplicitITKComposeImageFilter.cxx.o
      vtable for itk::ComposeImageFilter<itk::Image<int, 2u>, itk::VectorImage<int, 2u> > in sitkExplicitITKComposeImageFilter.cxx.o
      vtable for itk::ComposeImageFilter<itk::Image<int, 3u>, itk::VectorImage<int, 3u> > in sitkExplicitITKComposeImageFilter.cxx.o
      vtable for itk::ComposeImageFilter<itk::Image<long, 2u>, itk::VectorImage<long, 2u> > in sitkExplicitITKComposeImageFilter.cxx.o
      ...
﻿```

Any ideas what I might be needing to link against?

blowekamp:
Please create a new issue in ITK/SimpleITK as appropriate. I see you built `libSimpleITKExplicit-0.9.1.dylib` which is an old version of a SimpleITK library. Some explanation is needed to what you are doing in the *new* issue.

Issue #1133, opened 2019-08-01
Wrap CastImageFilter for vector pixel types


thewtex:
Wrap `CastImageFilter` to convert to and from vector pixel types like `Vector`, `CovariantVector`, etc. and `VectorImage`.

See:

  https://discourse.itk.org/t/installing-and-using-python-wrapped-remote-module/2104/9

PR #1579, opened 2020-01-29
WIP: ENH: Wrap CastImageFilter for VectorImage <-> Vector

thewtex:
Addresses #1133

### DIFF
--- a/Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
+++ b/Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
@@ -120,7 +120,6 @@ CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatche
   // Implementation for non-implicit convertible pixels which are
   // itk-array-like.
 
-  static_assert(OutputPixelType::Dimension == InputPixelType::Dimension, "Vector dimensions are required to match!");
   static_assert(std::is_convertible<typename InputPixelType::ValueType, typename OutputPixelType::ValueType>::value,
                 "Component types are required to be convertible.");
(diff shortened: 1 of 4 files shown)


Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
```diff
@@ -152,7 +153,7 @@ CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatche
     {
       const InputPixelType & inputPixel = inputIt.Get();
       OutputPixelType        value;
```
blowekamp: This will not get set to the correct length for vector pixels.

blowekamp: Outside the loops add something like:

```
OutputPixelType value;
NumericTraits<OutputPixelType>::SetLength(value, componentsPerPixels)
```

I'm not sure if its a `NumericTraits` or some other pixel traits class needed to set the  number components for all types OK.

Pulling it out size the loop also remove per-pixel allocation which is a performance killer.

thewtex:
Superceded by https://github.com/InsightSoftwareConsortium/ITK/pull/2073/files

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
thewtex: @blowekamp thanks for the note.

@Leengit it looks this addition was missed when with https://github.com/InsightSoftwareConsortium/ITK/pull/2073 -- could you please add it in a follow-up patch?

PR #1582, opened 2020-01-31
BUG: Call itk.down_cast on ProcessObject outputs in functional call

thewtex:
For functional Python ProcessObject or Filter calls, call itk.down_cast
on outputs to return the most derived type.

### DIFF
--- a/Wrapping/Generators/Python/PyBase/pyBase.i
+++ b/Wrapping/Generators/Python/PyBase/pyBase.i
@@ -361,12 +361,13 @@ str = str
                 """
                 self.UpdateLargestPossibleRegion()
                 try:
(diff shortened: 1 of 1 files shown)


jhlegarreta:
macOS CI failures seem to be related to the patch set though ! :zap:

thewtex:
@jhlegarreta thanks for the reviews! After a little effort we now have the tests passing. All of this hinges on `GetClassName`, which we have working well now thanks to your testing macros and the Python tests. :clap: :green_heart:

jhlegarreta:
Glad to read that Matt. Thanks for your words. I miss contributing actively ! Keep on the good work !

PR #1585, opened 2020-02-01
Input spatial object const

thewtex:
We can rebase #1582 #1583 after merging to re-check the CI.


dzenanz:
@jhlegarreta commit messages were fine, only the PR description was uncommon.

PR #1592, opened 2020-02-06
ENH: Explicitly instantiate MultiThreaderBase methods

blowekamp:
The MultiThreaderBase ParallelizeImageRegion methods are explicitly
instantiated in the ITK Common library for common dimensions of
1-5. This will improve compilation time, reduce the size of binaries
and may improve execution time from reduced code size.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkMultiThreaderBase.h
+++ b/Modules/Core/Common/include/itkMultiThreaderBase.h
@@ -466,3 +466,50 @@ ITKCommon_EXPORT std::ostream &
 
 } // end namespace itk
 #endif
+
+#ifndef ITK_TEMPLATE_EXPLICIT_MultiThreaderBase
+
+namespace itk
+{
+extern template void ITKCommon_EXPORT_EXPLICIT
+                     MultiThreaderBase::ParallelizeImageRegion<1>(const ImageRegion<1> &, TemplatedThreadingFunctorType<1>, ProcessObject *);
+extern template void ITKCommon_EXPORT_EXPLICIT
+                     MultiThreaderBase::ParallelizeImageRegion<2>(const ImageRegion<2> &, TemplatedThreadingFunctorType<2>, ProcessObject *);
+extern template void ITKCommon_EXPORT_EXPLICIT
+                     MultiThreaderBase::ParallelizeImageRegion<3>(const ImageRegion<3> &, TemplatedThreadingFunctorType<3>, ProcessObject *);
+extern template void ITKCommon_EXPORT_EXPLICIT
+                     MultiThreaderBase::ParallelizeImageRegion<4>(const ImageRegion<4> &, TemplatedThreadingFunctorType<4>, ProcessObject *);
+extern template void ITKCommon_EXPORT_EXPLICIT
+                     MultiThreaderBase::ParallelizeImageRegion<5>(const ImageRegion<5> &,
+                                             TemplatedThreadingFunctorType<5> funcP,
+                                             ProcessObject *);
+
+
+extern template void ITKCommon_EXPORT_EXPLICIT
+                     MultiThreaderBase::ParallelizeImageRegionRestrictDirection<1>(unsigned int,
+                                                              const ImageRegion<1> &,
+                                                              TemplatedThreadingFunctorType<1>,
+                                                              ProcessObject *);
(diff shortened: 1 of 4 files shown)


Modules/Core/Common/src/itkMultiThreaderBase.cxx
```diff
@@ -650,4 +651,52 @@ MultiThreaderBase::PrintSelf(std::ostream & os, Indent indent) const
 
 MultiThreaderBaseGlobals * MultiThreaderBase::m_PimplGlobals;
 
+
+template void ITKCommon_EXPORT
+              MultiThreaderBase::ParallelizeImageRegion<1>(const ImageRegion<1> &           requestedRegion,
```
dzenanz: I guess that naming these parameters is not important?

Modules/Core/Common/src/itkMultiThreaderBase.cxx
```diff
@@ -650,4 +651,52 @@ MultiThreaderBase::PrintSelf(std::ostream & os, Indent indent) const
...
+              MultiThreaderBase::ParallelizeImageRegion<4>(const ImageRegion<4> &           requestedRegion,
+                                             TemplatedThreadingFunctorType<4> funcP,
+                                             ProcessObject *                  filter);
+template void ITKCommon_EXPORT
+              MultiThreaderBase::ParallelizeImageRegion<5>(const ImageRegion<5> &           requestedRegion,
+                                             TemplatedThreadingFunctorType<5> funcP,
+                                             ProcessObject *                  filter);
+
+template void ITKCommon_EXPORT
+              MultiThreaderBase::ParallelizeImageRegionRestrictDirection<1>(unsigned int,
```
dzenanz: These parameters are not named.

Modules/Core/Common/include/itkMultiThreaderBase.h
```diff
@@ -466,3 +466,50 @@ ITKCommon_EXPORT std::ostream &
 
 } // end namespace itk
 #endif
+
+#ifndef ITK_TEMPLATE_EXPLICIT_MultiThreaderBase
+
+namespace itk
+{
+extern template void ITKCommon_EXPORT_EXPLICIT
+                     MultiThreaderBase::ParallelizeImageRegion<1>(const ImageRegion<1> &, TemplatedThreadingFunctorType<1>, ProcessObject *);
```
dzenanz: Does code completion pick up the previous definition with named parameters? If not, we should add parameter names here.

thewtex:
We should not be adding explicit instantiations like this, because it limits what can be instantiated (we cannot instantiate other dimensions). It also adds bloat to binaries generated -- for many binaries, they are not using the instantiated dimensions.

blowekamp:
> We should not be adding explicit instantiations like this, because it limits what can be instantiated (we cannot instantiate other dimensions).

This is not true for this code. Other dimensions can me implicitly instantiated.

> It also adds bloat to binaries generated -- for many binaries, they are not using the instantiated dimensions.

I have a different concept of code bloat than you.

This patch is not worth pursuing further.

dzenanz:
Dimensions 2 and 3 are surely used. We could keep at least those?

thewtex:
Dimension 2 and 3 are not always used. Including them increases the size for size-critical applications like WebAssembly builds.

The other problem with explicit instantiation is that it prevents instantiation of other types in shared build configurations where explicitly exported symbols do not mix with implicitly instantiated symbols.

blowekamp:
> The other problem with explicit instantiation is that it prevents instantiation of other types in shared build configurations where explicitly exported symbols do not mix with implicitly instantiated symbols.

This is certainly a valid concern in general with explicit instantiated object. I believe you are referring to the `dynamic_cast`ing issue. For this particular case, no dynamic_casts are used in the code. And the instantiation is a member function not a class with a vtable.



> Did you check how much code size is reduced?

The object files were reduced in size a bit, but the final library was about the same size. I concluded it was not worth the risk of other potential issues.

I am curious about the library size/expense of using lambda functions and their impact on the size of the symbol tables of libraries.

thewtex:
>  I believe you are referring to the dynamic_casting issue.

The symbols problems are not just related to `dynamic_cast`ing -- this also causes issues for build-time linking and run-time library loading.

PR #1599, opened 2020-02-08
BUG: operator() can not be const

hjmjohnson:
The member variable is changed, so the operator() can not be const.

Bug identified with clang10 compilation


hjmjohnson:
Failure is unrelated.

PR #1589, opened 2020-02-03
BUG: incorrect neighbor determination in scan-line derived filters

dzenanz:
Thanks to @DVigneault for finding the bug and creating a MWE for demonstrating it. Forum [thread](https://discourse.itk.org/t/incorrect-result-from-itk-connectedcomponentimagefilter/2676).

PR #1600, opened 2020-02-10
BUG: remove assertion which frequently fails

dzenanz:
We are not checking difference along dimension zero,
so it is possible for diffSum to be 0 despite neighbors being different.

This is a follow-up to #1589.

### DIFF
--- a/Modules/Filtering/ImageLabel/include/itkScanlineFilterCommon.h
+++ b/Modules/Filtering/ImageLabel/include/itkScanlineFilterCommon.h
@@ -247,7 +247,6 @@ protected:
       diffSum += diff;
     }
 
-    assert(diffSum > 0); // we are checking potential neighbors, index difference must exist
     if (!this->m_FullyConnected)
     {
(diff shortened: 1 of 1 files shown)


dzenanz:
This is motivated by an increase in number of test failures on the [dashboard](https://open.cdash.org/index.php?project=Insight).

dzenanz:
Running tests in a debug build takes a long time. Would that work with current CI limitations?

DVigneault:
If I understand this correctly, `diffSum` will only equal zero (and trigger the `assert`) if both indices come from the same scanline--in which case I would say that `A` and `B` are not neighbors, and expect the function to return `false`.  Currently, I think it returns `true` in this case.  Would it cause any problems to instead replace the `assert` with:

`if (diffSum == 0) { return false; }`

?

Apologies if I've misunderstood.  Otherwise LGTM!

dzenanz:
We might replace it by `if (diffSum == 0) { return true; }`, but that is subsumed in the code after the assert.

PR #1602, opened 2020-02-10
WIP: Hide GDCM symbols

blowekamp:
When compiling ITK with static library and with explicit hidden
visibility:

 -DCMAKE_VISIBILITY_INLINES_HIDDEN:BOOL=ON
 -DCMAKE_C_VISIBILITY_PRESET:STRING=hidden
 -DCMAKE_CXX_VISIBILITY_PRESET:STRING=hidden

numerous GDCM symbols are still export on linux and OSX. GDCM was
always defining it's libraries and several utility libraries as
"default" even when static libraries with explicit hidden visibility
was requested.

This can cause conflicting runtime environment if multiple libraries
or systems are using ITK or GDCM. For example SimpleITK's distribution
compiles with ITK symbols hidden, but GDCM symbols are still in it's
library interface. This cause a conflict when imported into 3D
Slicer.

## PR Checklist


blowekamp:
This patch will need to be pushed to upstream GDCM. But I thought this would be a good place to get the discussion started.

@malaterre The GDCM library and utility libraries are always exporting their symbols with "default" visibility. For this case these libraries are compiled static, and the symbols should be hidden. Any information about why GDCM currently has this behavior?

malaterre:
> This patch will need to be pushed to upstream GDCM. But I thought this would be a good place to get the discussion started.
> 
> @malaterre The GDCM library and utility libraries are always exporting their symbols with "default" visibility. For this case these libraries are compiled static, and the symbols should be hidden. Any information about why GDCM currently has this behavior?

Off the top of my head this had to do with the python wrapping. I am pretty sure this broke the python 2.7 wrapping I tried at the time. I'll test your patch, verify this pass this time and merge it upstream. Thanks for preparing it !

blowekamp:
A proper update of GDCM and subtree update is in PR #1606.

Issue #1542, opened 2020-01-08
Multi-Frame DICOM load performace decreased substantially in 4.13.2+


timbra339:
I've updated my ITK libraries from 4.13.0 to 5.0.1 recently and noticed a quite substantial drop in DICOM data load performance.

I've looked a bit into this using the attached python test-file which loads a ~800MB multi-frame DICOM file. Here are the resulting load times using various versions of ITK (used python-wheels for that):

ITK Version: 4.13.0  (python-package itk==4.13.1.post1)
Loading took **2.3250107765197754** secs.

ITK Version: 4.13.2
Loading took **16.011502027511597** secs.

ITK Version: 5.0.0
Loading took 15.990726470947266 secs.

ITK Version: 5.0.1
Loading took 15.931232690811157 secs.

ITK Version: 5.1.0
Loading took 15.805387496948242 secs.

It looks like the performance drop cause was introduced with ITK 4.13.2, whose release notes state several changes to GDCM, so that appears to be a likely cause? The issues seems to be still in the current codebase, maybe it can be fixed for the upcoming 5.1 version? I can supply the test data if needed.

```python
# -*- coding: utf-8 -*-

import itk
import time

print(f"ITK Version: {itk.Version.GetITKVersion()}")

reader = itk.ImageFileReader[itk.Image[itk.US, 3]].New(FileName = r"Volume_0000.dcm")
start = time.time()
reader.Update()
stop = time.time()

print(f"Loading took {stop-start} secs.")
```

dzenanz:
Please give a link to the test data.

timbra339:
Uploading the test data. 
I've zeroed out the actual image content to make the file compressible, the runtimes for 4.13.0 and 5.1 are still  as reported.
[Volume_0000.zip](https://github.com/InsightSoftwareConsortium/ITK/files/4035688/Volume_0000.zip)

thewtex:
ITK 5.1 RC 1 contains many GDCM updates.

You can install the wheels for testing with:

```
pip install --upgrade --pre itk
```

@malaterre

timbra339:
@thewtex: This command installs the version 5.1rc1.post1 on my machine, which was the one that produced the 

ITK Version: 5.1.0
Loading took 15.805387496948242 secs.

output. Should there be a more recent wheel instead?

dzenanz:
@issakomi might also pitch in.

issakomi:
[...]

PR #1606, opened 2020-02-12
Update gdcm 2020 02 12

blowekamp:
## PR Checklist

- [x] :no_entry_sign: [Makes breaking changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes)
- [ ] :no_entry_sign: [Makes design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes)
- [x] :no_entry_sign: Adds the License notice to new files.
- [X] :no_entry_sign: Adds Python wrapping.
- [X] :no_entry_sign: Adds tests and baseline comparison (quantitative).
- [X] :no_entry_sign: [Adds test data](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/UploadBinaryData.md).
- [X] :no_entry_sign: Adds Examples to [ITKExamples](https://github.com/InsightSoftwareConsortium/ITKExamples)
- [X] :no_entry_sign: Adds Documentation.

Refer to the [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) for
further development details if necessary.

PR #1608, opened 2020-02-12
BUG: Fix DICOM loading speed regression

lassoan:
GDCM has changed so that it masks unused bits more frequently.
However, this masking operation was implemented so inefficiently that loading time increased by a factor of 5-10x (on Windows, VS2017).

Fixed by reading/writing data in the data stream in larger chunks.

I know that this should be fixed in GDCM but I've already spent more time with this than I wanted. I hope that @malaterre and others will take this and integrate in GDCM/ITK as appropriate.

It fixes https://github.com/InsightSoftwareConsortium/ITK/issues/1542.

## PR Checklist

### DIFF
--- a/Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmImageCodec.cxx
+++ b/Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmImageCodec.cxx
@@ -538,14 +538,22 @@ bool ImageCodec::DoOverlayCleanup(std::istream &is, std::ostream &os)
     else // Pixel are unsigned
       {
 #if 1
-      uint16_t c;
-      while( is.read((char*)&c,2) )
+      // On Windows, is.read and os.write are expensive operations.
+      // If we called it for each value then conversion would be extremely slow.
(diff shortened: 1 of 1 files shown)


malaterre:
@lassoan thanks for the patch ! Can you comment on the speed improvement given by this patch ?

As a side note, I would have picked 1024 instead of 1000.

malaterre:
If this is not clear, I'll import in GDCM. There might be some conflict with a recent work from @blowekamp and PR #1606.

lassoan:
> Can you comment on the speed improvement given by this patch ?

Loading of a JPEG-compressed image used to take about 10 seconds.
After recent GDCM changes, this slowed down to about 1 minute.
With the changes in this pull request, loading takes about 10 seconds again.

> As a side note, I would have picked 1024 instead of 1000.

Most likely it does not matter, but I just wanted to remain somewhat below a 2^n round number (to make sure that a few extra bytes does not make the memory area larger than 2^n).

malaterre:
@lassoan I've chery-picked your commit in GDCM upstream:

* https://github.com/malaterre/GDCM/commit/e7daa53652fbbc146a17db7ad2512957e1b31ac6

thanks !

PR #1610, opened 2020-02-12
ENH: Update RTK for future proofing.

hjmjohnson:
Using explicit typecasts for future-proofing
enums.  RTK treats enum values as integers and
enumerations interchangeably for pushing onto
a stack of integers.  Using explicit static_casts
makes code future proof when enum class is used.


SimonRit:
Thanks @hjmjohnson. It was my intention to update it before the new ITK release candidate. If this is still planned for tomorrow, then I'll finish a few things today and suggest a newer RTK tag tomorrow morning if all nightly builds went fine.

hjmjohnson:
@SimonRit This small change is needed for the updates to enumerations.  I'm going to merge this version to remove one complication in expanded testing.  A new PR with even newer RTK should be considered as a separate issue.

PR #1612, opened 2020-02-12
COMP: Need to explicitly define std::min<> template values

hjmjohnson:
Compiling with clang++-11 compiler identified ambiguous
types for min function.


blowekamp:
Thanks for fixing this Hans. I just saw this error on the SimpleITK dashboard with VS v140.

Issue #1607, opened 2020-02-12
Hypersensitive "Non unform sampling or missing slices detected" warning


lassoan:
### Description

ITK's new non-uniform "sampling" detection (actually, slice spacing) is so sensitive that it is useless. It also generates so many warnings that it floods the log and slows down DICOM loading.

Example:

WARNING: In c:\d\s42d\itk\modules\io\imagebase\include\itkImageSeriesReader.hxx, line 440
ImageSeriesReader (000001F94F144440): Non unform sampling or missing slices detected , expected 0.3 got: 0.3

The compared values are 0.29999999999999716 and 0.29999999999999999 (absolute difference is 2.8310687127941492e-15).

### Steps to Reproduce

Load any DICOM series - but to give a specific example:
https://wiki.cancerimagingarchive.net/display/Public/QIN-HEADNECK
TCGA-BA-4074

### Expected behavior

Only give a warning when spacing variation is non-negligible.

If the difference is magnitudes lower then the expected slice spacing then I definitely don't want to see any warning and don't want the image to be considered as having varying slice spacing.

### Actual behavior

A warning is logged for each slice.

### Reproducibility

100%

### Versions

Latest ITK 5.0.1rc01

### Environment

Windows.

vfonov:
didn't we address this in #1476 ?

dzenanz:
Perhaps instead of AlmostEquals use [FloatAlmostEqual](https://itk.org/Doxygen/html/namespaceitk_1_1Math.html#ae9f0d6137957033eecb66c0e1356d022) with some medium tolerance (1e-6)?

vfonov:
I suppose it will depend on the context?

dzenanz:
Perhaps use a more strict tolerance (e.g. 1e-14)?

vfonov:
we can also disable the warning completely and keep metadata information only

dzenanz:
Yes, that is one way too.

lassoan:
I use the latest release candidate (ITK v5.1-rc01), which does not have this yet.

I've cherry-picked the content of https://github.com/InsightSoftwareConsortium/ITK/pull/1476 but it did not help much. It still pollutes the log with hundreds of these messages (but at least it now shows the "deviation", too):

WARNING: In c:\d\s42rdi\itk\modules\io\imagebase\include\itkImageSeriesReader.hxx, line 444
[...]

PR #1616, opened 2020-02-13
BUG: reduced number of warnings for non-uniform sampling

vfonov:
BUG: Related to #1607, now only one warning per series is emitted, when maximum 
slice sampling error is greater then 1e-6

## PR Checklist

### DIFF
--- a/Modules/IO/ImageBase/include/itkImageSeriesReader.h
+++ b/Modules/IO/ImageBase/include/itkImageSeriesReader.h
@@ -169,6 +169,10 @@ public:
   itkGetConstReferenceMacro(UseStreaming, bool);
   itkBooleanMacro(UseStreaming);
 
+  /** Set the relative threshold for issuing warnings about non-uniform sampling */
+  itkSetMacro(SpacingWarningRelThreshold, double);
+  itkGetConstMacro(SpacingWarningRelThreshold, double);
+
 protected:
   ImageSeriesReader()
     : m_ImageIO(nullptr)
@@ -208,6 +212,8 @@ protected:
   bool m_UseStreaming{ true };
 
   bool m_SpacingDefined{ false };
+
+  double m_SpacingWarningRelThreshold{ 1e-4 };
 
 private:
   using ReaderType = ImageFileReader<TOutputImage>;
--- a/Modules/IO/ImageBase/include/itkImageSeriesReader.hxx
+++ b/Modules/IO/ImageBase/include/itkImageSeriesReader.hxx
@@ -439,18 +439,12 @@ ImageSeriesReader<TOutputImage>::GenerateData()
         {
           nonUniformSampling = true;
           spacingDeviation = Math::abs(outputSpacing[this->m_NumberOfDimensionsInImage] - dirNnorm);
-          itkWarningMacro(<< "Non uniform sampling or missing slices detected , expected " << std::setprecision(14)
-                          << outputSpacing[this->m_NumberOfDimensionsInImage] << " got: " << dirNnorm
-                          << " Deviation of:" << spacingDeviation);
-
-          needToUpdateMetaDataDictionaryArray = true;
           if (spacingDeviation > maxSpacingDeviation)
           {
             maxSpacingDeviation = spacingDeviation;
-            EncapsulateMetaData<double>(output->GetMetaDataDictionary(),
-                                        "ITK_non_uniform_sampling_deviation",
-                                        maxSpacingDeviation); // maximum deviation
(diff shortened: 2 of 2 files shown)


Modules/IO/ImageBase/include/itkImageSeriesReader.hxx
```diff
@@ -479,6 +476,12 @@ ImageSeriesReader<TOutputImage>::GenerateData()
 
   } // end per slice loop
 
+  if (maxSpacingDeviation > 1e-6) // arbitrary number
```
dzenanz: This should probably be proportional to the spacing, e.g. if `if (maxSpacingDeviation > 1e-6 * outputSpacing[this->m_NumberOfDimensionsInImage])`

hjmjohnson: @dzenanz I don't really understand what you are suggesting here.

The psuedo code you suggest does not seem valid to me.  Do you mean the product of all spacing sizes times 1e-6?

dzenanz: `outputSpacing[this->m_NumberOfDimensionsInImage]` would mostly mean `zSpacing` (a common case with a series of 2D DICOM slices).

Modules/IO/ImageBase/include/itkImageSeriesReader.hxx
```diff
@@ -439,9 +439,6 @@ ImageSeriesReader<TOutputImage>::GenerateData()
         {
           nonUniformSampling = true;
           spacingDeviation = Math::abs(outputSpacing[this->m_NumberOfDimensionsInImage] - dirNnorm);
-          itkWarningMacro(<< "Non uniform sampling or missing slices detected , expected " << std::setprecision(14)
-                          << outputSpacing[this->m_NumberOfDimensionsInImage] << " got: " << dirNnorm
-                          << " Deviation of:" << spacingDeviation);
 
           needToUpdateMetaDataDictionaryArray = true;
```
lassoan: metadata field "ITK_non_uniform_sampling_deviation" should only be updated once, at the end

vfonov: ImageSeriesReader doesn't work for 4D or 5D images.

blowekamp: I'm pretty sure it works to create a N+1D image from ND images.... Maybe not for DICOM but other formats.

vfonov: I tried with fMRI series - it doesn't.

Modules/IO/ImageBase/include/itkImageSeriesReader.hxx
vfonov: you are reviewing previous version , by the way

Modules/IO/ImageBase/include/itkImageSeriesReader.hxx
```diff
@@ -476,9 +470,23 @@ ImageSeriesReader<TOutputImage>::GenerateData()
       }
       m_MetaDataDictionaryArray.push_back(newDictionary);
     }
-
   } // end per slice loop
 
+
+  if (maxSpacingDeviation >
+      m_SpacingWarningRelThreshold * outputSpacing[this->m_NumberOfDimensionsInImage]) // arbitrary number
```
dzenanz: The "arbitrary number" comment could be removed.

Modules/IO/ImageBase/include/itkImageSeriesReader.h
```diff
@@ -209,6 +213,8 @@ class ITK_TEMPLATE_EXPORT ImageSeriesReader : public ImageSource<TOutputImage>
 
   bool m_SpacingDefined{ false };
 
+  double m_SpacingWarningRelThreshold{ 1e-6 };
```
pieper: Nitpick: a hard coded number is probably fine for most use cases, but @lassoan 's original suggestion of 1% of slice spacing (or 0.1% even if you must) would be much more practical.  For both CT and MR scans there are physical reasons why the table position or slice reconstruction intervals may have discrete but negligible errors that aren't due simply to floating point rounding.

hjmjohnson: I agree with the observation indicated by @pieper.  Personal preference is to err on the side of 0.1% until we identify that this is too stringent, and then increase based on data.

PR #1623, opened 2020-02-14
COMP: Implicit conversion warnings

hjmjohnson:
Core/Common/include/itkColorTable.hxx:182:17:
Core/Common/include/itkColorTable.hxx:198:19:
Core/Common/include/itkColorTable.hxx:91:17:
Filtering/ImageIntensity/test/itkClampImageFilterTest.cxx:111:26:
Core/Common/include/itkColorTable.hxx:139:20: warning: implicit conversion from
'itk::NumericTraits<long long>::ValueType' (aka 'long long') to 'double'
changes value from 9223372036854775807 to 9223372036854775808 [-Wimplicit-int-float-conversion]

### DIFF
--- a/Modules/Core/Common/include/itkColorTable.hxx
+++ b/Modules/Core/Common/include/itkColorTable.hxx
@@ -88,7 +88,10 @@ ColorTable<TPixel>::UseDiscreteColors()
   // set to NumericTraits<TPixel>::max().
   typename NumericTraits<TPixel>::RealType realMax(1.0 * scale + shift);
   TPixel                                   pixelMax(NumericTraits<TPixel>::max());
-  if (realMax < NumericTraits<TPixel>::max())
+  // Converting from TPixel to RealType may introduce a rounding error, so do static_cast
+  constexpr auto max_value_converted =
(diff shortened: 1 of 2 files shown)


Modules/Core/Common/include/itkColorTable.hxx
```diff
@@ -88,7 +88,10 @@ ColorTable<TPixel>::UseDiscreteColors()
   // set to NumericTraits<TPixel>::max().
   typename NumericTraits<TPixel>::RealType realMax(1.0 * scale + shift);
   TPixel                                   pixelMax(NumericTraits<TPixel>::max());
-  if (realMax < NumericTraits<TPixel>::max())
+  // Converting from TPixel to RealType may introduce a rounding error, so do static_cast
+  constexpr typename NumericTraits<TPixel>::RealType max_value_converted =
```
dzenanz: Shouldn't this be `constexpr auto max_value_converted` due to explicit cast? Also repeated twice below.

hjmjohnson:
It could be.

Modules/Core/Common/include/itkColorTable.hxx
hjmjohnson: Fixed in upcomming rebase.

PR #1625, opened 2020-02-15
BUG: Missing backward compatible enumeration name

hjmjohnson:
Building deprecated code and remote modules identified a few missing backward compatibility issues.

Backward compatible enumerations for Remote Modules

Build the remote modules identified missing backward compatible
names that were missing.

Incomplete refactoring for deprecated enumeration.

Incorrect type change from enumeration update

### DIFF
--- a/Modules/Filtering/Colormap/include/itkScalarToRGBColormapImageFilter.h
+++ b/Modules/Filtering/Colormap/include/itkScalarToRGBColormapImageFilter.h
@@ -54,6 +54,7 @@ public:
 };
 #if !defined(ITK_LEGACY_REMOVE)
 using RGBColormapFilterEnum = ScalarToRGBColormapImageFilterEnums::RGBColormapFilter;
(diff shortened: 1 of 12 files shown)


Modules/Filtering/Colormap/include/itkScalarToRGBColormapImageFilter.h
```diff
@@ -54,6 +54,7 @@ class ScalarToRGBColormapImageFilterEnums
 };
 #if !defined(ITK_LEGACY_REMOVE)
 using RGBColormapFilterEnum = ScalarToRGBColormapImageFilterEnums::RGBColormapFilter;
```
dzenanz: Do we need line 56 if we have line 57?

hjmjohnson: I am not sure which is needed for backward compatibility.  Error on the side of defining an unused name in the ITK_LEGACY_REMOVE section.

PR #1626, opened 2020-02-16
ENH: Update DCMTK to master branch of upstream DCMTK

hjmjohnson:
DCMTK no longer requires as many custom patches to build ITK as many patches have all been incorporated into the upstream versions.


hjmjohnson:
> Looks good, but I don't see a reduction in custom patches implied by the commit message.

The reduction is in customizations done in the DCMTK project at 14393c0baa743469b5826bea724c048e221dd08b

PR #1631, opened 2020-02-17
BUG: Incorrect type specified for enumeration

### LINKED ISSUE #1629: PhilipsRECImageIO does not compile
`PhilipsRECImageIO`, a module which is not enabled by default, no longer compiles when legacy code is disabled. The likely culprit is 0dd31a62f851a0ae95ef4c6d6c85e531a4ebca59.

hjmjohnson:
Fixes #1629 when legacy code is disabled.


thewtex:
Unlikely that failing Python tests are related.

PR #1632, opened 2020-02-18
ENH: Add itk.xarray_from_image and itk.image_from_xarray

thewtex:
These functions convert an itk.Image to and from an xarray.DataArray. The
xarray.DataArray preserves spatial metadata while processing in Python.
xarray enables parallel computing with Dask, machine learning, and
modern seralization with Zarr.

These interfaces are complete but experimental. They are available for
testing but are subject to possible future changes and refinements.

### DIFF
--- a/Wrapping/Generators/Python/Tests/extras.py
+++ b/Wrapping/Generators/Python/Tests/extras.py
@@ -147,189 +147,246 @@ except Exception as e:
   else:
     raise e
 
-# test mesh read / write
-mesh = itk.meshread(mesh_filename)
-assert type(mesh) == itk.Mesh[itk.F, 3]
-mesh = itk.meshread(mesh_filename, itk.UC)
-assert type(mesh) == itk.Mesh[itk.UC, 3]
-mesh = itk.meshread(mesh_filename, itk.UC, fallback_only=True)
-assert type(mesh) == itk.Mesh[itk.F, 3]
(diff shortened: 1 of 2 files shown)


Wrapping/Generators/Python/Tests/extras.py
```diff
@@ -147,189 +147,246 @@ def custom_callback(name, progress):
...
-
-# test reading image series with `itk.imread()` and check that dimension is
-# not increased if last dimension is 1.
-image_series = itk.imread([image_series3d_filename, image_series3d_filename])
-assert image_series.GetImageDimension() == 3
-
-# pipeline, auto_pipeline and templated class are tested in other files
-
-# BridgeNumPy
+# # test mesh read / write
```
dzenanz: Why was this code commented out? It does not seem related to xarray.

thewtex: Good catch -- I will re-enable.

thewtex: Looks like @hjmjohnson caught it and fixed it -- thanks, Hans!

dzenanz: I don't see the fix.

thewtex: I was looking at the wrong file :-P -- I will patch it.

PR #1637, opened 2020-02-19
BUG: Remove duplicate wrapping of itk.MeshEnums

thewtex:
Following strongly typed enums refactoring, this is now wrapped in
ITKCommon.

To address #1636


thewtex:
Windows failure related to network download of testing data.

PR #1638, opened 2020-02-19
BUG: Wrap TriangleMeshToBinaryImageFilter for float types

thewtex:
xref: https://discourse.itk.org/t/convert-from-itk-mesh-itk-f-3-to-itk-mesh-itk-ss-3/2719/2


Modules/Core/Mesh/wrapping/itkTriangleMeshToBinaryImageFilter.wrap
```diff
@@ -6,7 +6,8 @@ if(d3)
   itk_wrap_include("itkTriangleMeshToBinaryImageFilter.h")
   itk_wrap_simple_class("itk::Point1D")
   itk_wrap_class("itk::TriangleMeshToBinaryImageFilter" POINTER)
-    foreach(t ${WRAP_ITK_INT})
+    UNIQUE(types "${WRAP_ITK_SCALAR};D")
+    foreach(t ${types})
       foreach(d 3)
         itk_wrap_template("M${ITKM_${t}}${d}${ITKM_I${t}${d}}" "itk::Mesh< ${ITKT_${t}},${d} >, ${ITKT_I${t}${d}} ")
```
dzenanz: Maybe `itk::Mesh< ${ITKT_${t}},${d} >` needs to be replaced by `itk::Mesh< D,${d} >` (or something similar), to keep the mesh positions always `double`, even if image pixel type is an `int`.

thewtex: The mesh positions type, a floating point type, is determined by a different template parameter. We are using the default here and in `itk::MeshFileReader`.

dzenanz:
itkTriangleMeshToBinaryImageFilter: warning(4): ITK type not wrapped, or currently not known: itk::ImageSource< itk::Image< double, 3 > >

PR #1644, opened 2020-02-19
DOC: Adding ORCID for Ziv Yaniv.

zivy:


.zenodo.json
```diff
@@ -327,6 +327,7 @@
     },
     {
       "name": "Yaniv, Ziv"
```
thewtex: Please add a `,` for valid JSON.

zivy: Fixed and force pushed. Can't seem to add one line without making a mistake ;).

PR #1646, opened 2020-02-20
ENH: updating remote module PolarTransform

romangrothausmann:
Make use of PR to ITKPolarTransform: 
https://github.com/InsightSoftwareConsortium/ITKPolarTransform/pull/13

### DIFF
--- a/Modules/Remote/PolarTransform.remote.cmake
+++ b/Modules/Remote/PolarTransform.remote.cmake
@@ -11,5 +11,5 @@ For more information, see the Insight Journal article:
   http://insight-journal.org/browse/publication/130
   "
   GIT_REPOSITORY ${git_protocol}://github.com/InsightSoftwareConsortium/ITKPolarTransform.git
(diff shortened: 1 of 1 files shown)


romangrothausmann:
Thanks for merging @thewtex. I noticed that some change between ITK-5.1rc01 and master @ 1d1ad8 causes 2 of my ITK-CLI test to fail:
https://gitlab.com/romangrothausmann/ITK-CLIs/pipelines/119648212
the one for `project` writes a TIF, whose content seems to be the same as before (diff in Fiji) and DM3 reading seems to lead to the same content as well but both yield a different MD5 than before.
Has there been a change to ITK that would explain this? 
I.e. should I just update the MD5s to be expected for newer ITK?

romangrothausmann:
Seems like the TIFs now output by ITK contain a time-stamp or uuid or similar. Each re-run of the same command leads to a different MD5.
Conversion of the DM3 leads to a constant but different MD5 (of an MHA).

PR #1647, opened 2020-02-20
BUG: Specify CMAKE_GENERATOR for Eigen external project build

thewtex:
On Windows, with the Ninja Generator and a non-default Visual Studio
toolset, the Eigen External Project will fail to build, attempting to
use MSBuild, without the generator specified.


phcerdan:
Good fix, that's the only caveat of `execute_process`.

It would be great if `CMake` refactors all the logic used in `ExternalProject_add` to handle toolchains and generators into some kind of variable i.e `${CMAKE_TOOLCHAIN_VARIABLES}` or similar.
The good thing is that we might only need to worry about the generator in this case, because we are not really compiling anything.

PR #1648, opened 2020-02-20
Python3 executable windows

thewtex:


thewtex:
Verified locally. Merging for the release.

CMakeLists.txt
```diff
@@ -576,6 +576,10 @@ if(BUILD_TESTING OR ITK_BUILD_DOCUMENTATION OR ITK_WRAP_PYTHON)
     else()
       find_package(Python3 COMPONENTS Interpreter)
     endif()
+    if(NOT Python3_EXECUTABLE AND _Python3_EXECUTABLE)
+      set(Python3_EXECUTABLE ${_Python3_EXECUTABLE} CACHE INTERNAL
+        "Path to the Python interpreter" FORCE)
```
jcfr: Nitpick: `FORCE` is not needed, It is implied. See https://cmake.org/cmake/help/latest/command/set.html#set-cache-entry

thewtex: @jcfr good to know -- thanks for the note.

Issue #66, opened 2018-09-24
Building ITK on Windows


muschellij2:
### Building ITK on Windows (through R)

I'm trying to build ITK on Windows, through R in the ITKR package (https://github.com/muschellij2/ITKR).  I'm testing this through appveyor (https://ci.appveyor.com/project/muschellij2/itkr).  I'm getting the following error: 

```
[ 73%] Building C object Modules/ThirdParty/HDF5/src/itkhdf5/src/CMakeFiles/hdf5-static.dir/H5Defl.c.obj
In file included from C:\run\ITKR.Rcheck\00_pkg_src\ITKR\src\itks\Modules\ThirdParty\HDF5\src\itkhdf5\src\H5private.h:615:0,
                 from C:\run\ITKR.Rcheck\00_pkg_src\ITKR\src\itks\Modules\ThirdParty\HDF5\src\itkhdf5\src\H5Defl.c:29:
C:\run\ITKR.Rcheck\00_pkg_src\ITKR\src\itks\Modules\ThirdParty\HDF5\src\itkhdf5\src\H5Defl.c: In function 'H5D__efl_read':
C:\run\ITKR.Rcheck\00_pkg_src\ITKR\src\itks\Modules\ThirdParty\HDF5\src\itkhdf5\src\H5win32defs.h:57:66: error: expected expression before ')' token
 #define HDopen(S,F,...)       _open(S, F | _O_BINARY, __VA_ARGS__)
                                                                  ^
C:\run\ITKR.Rcheck\00_pkg_src\ITKR\src\itks\Modules\ThirdParty\HDF5\src\itkhdf5\src\H5Defl.c:291:18: note: in expansion of macro 'HDopen'
         if((fd = HDopen(full_name, O_RDONLY)) < 0)
                  ^
Modules\ThirdParty\HDF5\src\itkhdf5\src\CMakeFiles\hdf5-static.dir\build.make:770: recipe for target 'Modules/ThirdParty/HDF5/src/itkhdf5/src/CMakeFiles/hdf5-static.dir/H5Defl.c.obj' failed
```

This worked using commit 4f89edd8334da198099c686b993ec9f9f12dccdf, but not the latest or anything with respect to v5.0.

### Steps to Reproduce

The steps can be seen here: https://ci.appveyor.com/project/muschellij2/itkr

1.   The main configure script is https://github.com/muschellij2/ITKR/blob/master/configure.win
[...]

PR #1652, opened 2020-02-24
BUG: Pass CMAKE_<lang>_COMPILER to Eigen configuration

phcerdan:
Explicitly pass `CMAKE_C_COMPILER` `CMAKE_CXX_COMPILER` and
extra `CMAKE_GENERATOR_` variables to the internal Eigen configuration.

Raised here: https://github.com/InsightSoftwareConsortium/ITK/issues/66

### DIFF
--- a/CMake/itkExternal_Eigen3.cmake
+++ b/CMake/itkExternal_Eigen3.cmake
@@ -28,7 +28,9 @@ else()
   # ExternalProject_add creates a target for the build tree,
   # too late for our purposes of find_package at configure time.
   # We execute the configuration manually to create the Config and Targets.cmake files.
-  # Nothing to link or compile, so no need to pass generators or toolchains.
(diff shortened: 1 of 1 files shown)


phcerdan:
Hi @bradking, we are using `execute_process` instead of `ExternalProject_add` or `FetchContent` for simplicity. Eigen project only needs to generate its Targets and Config.cmake files at configure time, nothing to compile or link at build time.

I am wondering what is the minimal set of CMake variables that we need to explicitly pass to this "sub project".
So far, we are passing the Generator and the C/CXX compiler, because the lack of them have raised errors, would you suggest any other `CMAKE` variable to make it more robust?

bradking:
Also pass `CMAKE_{C,CXX}_FLAGS`.  Technically a few other variables might be relevant but that should handle most cases.

Issue #1655, opened 2020-02-25
Add rating scale for ITK Remote modules


hjmjohnson:
### Description

See https://discourse.itk.org/t/remote-module-grading-ranking-system-needed/2750

### Impact analysis

Provide a mechanism for remote module developers to convey information about the status of maturity for each remote module.  

Core ITK developers can then more easily leverage the remote modules for testing backward compatibility, and future features against the higher quality remote modules, without being bogged down by fixing bugs/known failures in remote modules that are in alpha or beta coding stages.

### Expected behavior

Be backward compatible with the current build system.  Current remote module files that are not distributed with ITK should continue to work (albeit with a warning message).

### Versions

Should be included with ITK 5.2 final release. 

### Additional Information
```
ITK_MINIMUM_COMPLIANCE_LEVEL defaults to 3, and can be changed provide a different limit on which modules are made available.
```

jhlegarreta:
The issue was not automatically closed when merging #1656 due to the cross-ref statement

> Resolves issue #1655

not fully fitting GitHub's issue [closing keyword statements](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

@hjmjohnson closing this manually.

PR #1656, opened 2020-02-25
ENH: Provide rating scale for ITK Remote modules

hjmjohnson:
SEE https://discourse.itk.org/t/remote-module-grading-ranking-system-needed/2750

Resolves issue #1655

## PR Checklist
- [X] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKExamples) for all new major features (if any)

### DIFF
--- a/CMake/ITKModuleRemote.cmake
+++ b/CMake/ITKModuleRemote.cmake
@@ -147,8 +147,15 @@ endfunction()
 # the CMake user interface.
 #
 # The following options are currently supported:
-#    [GIT_REPOSITORY url]        # URL of git repo
-#    [GIT_TAG tag]               # Git branch name, commit id or tag
+#    [MODULE_COMPLIANCE_LEVEL [5|4|3|2|1|0] ] # The compliance level of the module, used for filtering
+#            Compliance level 5 star (AKA ITK main modules, or remote modules that could become core modules)
+#            Compliance Level 4 star (Very high-quality code, perhaps small community dependance)
+#            Compliance Level 3 star (Quality beta code)
+#            Compliance Level 2 star (Alpha code feature API development or niche community/exectution environment dependance )
+#            Compliance Level 1 star (Pre-alpha features under development and code of unkown quality)
+#            Compliance Level 0 star ( Code/Feature of known poor-quality or deprecated status )
+#    [GIT_REPOSITORY url]            # URL of git repo
+#    [GIT_TAG tag]                   # Git branch name, commit id or tag
 #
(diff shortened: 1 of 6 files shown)


hjmjohnson:
> Thanks again for the effort Hans.
> 
> A general comment: I would assume that when a module complies with level `N`, it automatically complies with all lower levels `[N-1,0]`.
> 
> As for a full review/approval, I will defer the decision to others as I am not fully convinced that this is the solution to the current situation. Unfortunately, I have no better proposals or time to develop the aspects I brought in the discourse thread. I am sorry 😔.

@jhlegarreta Thanks!  You are correct that the selection defines the minimum.

jhlegarreta:
@hjmjohnson I may have not made clear my point :thinking: : if a module complies with e.g. level 4, then I would assume that checking `[X] All requirements below` at level 4 would imply that either **all** checkboxes of levels 0, 1, 2, and 3 should be checked or left empty (whatever it is decided to be best). Hence, I am not sure about the rationale behind checking `[X] All requirements below` at e.g. level 4 and checking only **some** of the checkboxes in lower levels.

lassoan:
Instead of "All requirements below", it would be more clear to say "All requirements for Level 1-3 fulfilled" (otherwise Level 0 would be contradictory).

I would not repeat the grading checklist in each cmake file, I would just keep the MODULE_COMPLIANCE_LEVEL value and a comment that explains why the software is at the current level (what are the main limitations that would need to be fixed to get to the next level). This way the checklist could be maintained more easily (it would not be replicated in each of the 44 modules) and would better match reality: that the checklist is a guideline only (there is no way it can reflect the current state of all modules all the time).

hjmjohnson:
> Instead of "All requirements below", it would be more clear to say "All requirements for Level 1-3 fulfilled" (otherwise Level 0 would be contradictory).
> 
> I would not repeat the grading checklist in each cmake file, I would just keep the MODULE_COMPLIANCE_LEVEL value and a comment that explains why the software is at the current level (what are the main limitations that would need to be fixed to get to the next level). This way the checklist could be maintained more easily (it would not be replicated in each of the 44 modules) and would better match reality: that the checklist is a guideline only (there is no way it can reflect the current state of all modules all the time).

@lassoan I agree 100% with your first comment and will implement that in the next few days. 

Regarding having the checklist in each file, as each module will have a different status, and meet different criteria on the checklist, I think it is important to act as a per-module report-card/roadmap of what needs to be done to elevate the module to the next grading level.

hjmjohnson:
Thanks, everyone for your feedback. I think we have a good, low burden solution that is better than the previous one.

PR #1657, opened 2020-02-25
COMP: Unused exception object in SmapsFileParser::ReadFile

thewtex:
To address:

  [CTest: warning matched] /Users/builder/externalModules/Core/Common/include/itkSmapsFileParser.hxx:128:34: warning: unused exception parameter 'excp' [-Wunused-exception-parameter]
  catch (const ExceptionObject & excp)
                                 ^


thewtex:
Yes, updating.

PR #1662, opened 2020-02-26
COMP: Use nullptr rather than 0 in IOFactoryRegisterRegisterList (3x)

N-Dekker:
Three "IOFactoryRegisterManager.h.in" files each declared an array of function
pointers, "IOFactoryRegisterRegisterList", which was terminated by `0`. This
triggered a Visual C++ (VS2017) Code Analysis warning:

Fixed by replacing `0` by `nullptr`.

### DIFF
--- a/CMake/itkImageIOFactoryRegisterManager.h.in
+++ b/CMake/itkImageIOFactoryRegisterManager.h.in
@@ -49,7 +49,7 @@ namespace {
 
   void (*ImageIOFactoryRegisterRegisterList[])(void) = {
     @LIST_OF_FACTORY_NAMES@
-    0};
+    nullptr};
   ImageIOFactoryRegisterManager ImageIOFactoryRegisterManagerInstance(ImageIOFactoryRegisterRegisterList);
 
 }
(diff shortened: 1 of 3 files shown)


N-Dekker:
Thanks, @dzenanz  As far as I can see, this is the minimal fix for those warnings. I have been looking into a slightly more ambitious adjustment, defining the "ManagerInstance" variable (in each of the three *.in files) as follows:

    const int ImageIOFactoryRegisterManagerInstance = []
    {
      constexpr void (*functionPtrArray[])() = { @LIST_OF_FACTORY_NAMES@ nullptr };

      for(const auto functionPtr: functionPtrArray)
      {
        if (functionPtr != nullptr)
        {
          functionPtr();
        }
      }
      return 0;
    }();

This would allow removing the undocumented "IOFactoryRegisterManager" helper classes from the `itk` namespace and the `IOFactoryRegisterRegisterList` variables from the unnamed namespace.

But for now it might be sufficient to just do the minimal fix of using `nullptr`!

PR #1670, opened 2020-02-29
Update MetaIO and associated SpatialObjects

aylward:
Update MetaIO [Kitware/MetaIO/#88](https://github.com/Kitware/MetaIO/pull/88):

"The MetaVesselTube and MetaTube classes differed only by SetArtery(true/false) function once
ridgeness, medialness, and other non-vessel-specific measures were moved into the MetaTube class (to make them commensurate with ITK's TubeSpatialObject). With this one (SetArtery) member function being the only difference, it was decided that Arterial is a generic enough term to also be moved (i.e.., arterial = "a through road" according to Websters); thereby eliminating the uniqueness of the MetaVesselTube class. Here it is kept as a facade class of MetaTube for backward compatibility. No tests needed to be changed in MetaIO or ITK.

Potential bug fix: reading tubes with binary point data encoding assumed the binary data was in a specific order and contained specific data, irrespective of the PointDim variable which, for ascii files, specifies the order of data specified for each point. As long as MetaIO was used to read/write binary files, this didn't cause an error but it was inconsistent. With this commit the PointDim order of information per point is respected by both ascii and binary encodings of point data."


aylward:
This pull request has two interdependent commits.   The first merges a branch that updates MetaIO, the second makes associated changes to itkSpatialObjects.

All of the changes in the attached files are backward compatible regarding file formats and APIs.  That is, old files can be read without changes and existing code does not need to be updates.  To provide this, none of the tests in MetaIO or ITK were updated for these changes and continue to pass.

These changes are needed so that MetaIO and SpatialObjects can optionally associate additional measures at each extracted tube point (e.g., branchness - an estimate of how likely a branch occurs at the current tube point).

PR #1678, opened 2020-03-04
BUG: PoolMultiThreader catch and re-throw exceptions from progress

blowekamp:
The progress reporter's CompletedPixel method may throw an
ProgressAborted exception. These call are now wrapped in the
ExceptionHandler class to catch and rethrow exceptions.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/src/itkPoolMultiThreader.cxx
+++ b/Modules/Core/Common/src/itkPoolMultiThreader.cxx
@@ -41,25 +41,15 @@ namespace
 class ExceptionHandler
 {
 public:
+  // This class follows the rule of zero
+
   template <typename TFunction>
-  explicit ExceptionHandler(const TFunction & function)
+  void
+  TryAndCatch(const TFunction & function)
   {
     try
     {
       function();
-    }
-    catch (...)
-    {
-      m_FirstCaughtException = std::current_exception();
-    }
-  }
-
-  void
-  TryToGetFuture(PoolMultiThreader::ThreadPoolInfoStruct & threadPoolInfo)
-  {
-    try
-    {
-      threadPoolInfo.Future.get();
     }
     catch (...)
     {
@@ -148,13 +138,14 @@ PoolMultiThreader::SingleMethodExecute()
   // Now, the parent thread calls this->SingleMethod() itself
   m_ThreadInfoArray[0].UserData = m_SingleData;
   m_ThreadInfoArray[0].NumberOfWorkUnits = m_NumberOfWorkUnits;
-  ExceptionHandler exceptionHandler([this] { m_SingleMethod(&m_ThreadInfoArray[0]); });
+  ExceptionHandler exceptionHandler;
+  exceptionHandler.TryAndCatch([this] { m_SingleMethod(&m_ThreadInfoArray[0]); });
 
   // The parent thread has finished SingleMethod()
   // so now it waits for each of the other work units to finish
   for (threadLoop = 1; threadLoop < m_NumberOfWorkUnits; ++threadLoop)
   {
-    exceptionHandler.TryToGetFuture(m_ThreadInfoArray[threadLoop]);
+    exceptionHandler.TryAndCatch([this, threadLoop] { m_ThreadInfoArray[threadLoop].Future.get(); });
   }
 
   exceptionHandler.RethrowFirstCaughtException();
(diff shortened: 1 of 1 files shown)


Modules/Core/Common/src/itkPoolMultiThreader.cxx
```diff
@@ -41,16 +41,22 @@ namespace
 class ExceptionHandler
 {
 public:
+  ExceptionHandler() = default;
```
N-Dekker: If you really want to remove the explicit `ExceptionHandler(const TFunction &)` constructor, you might as well follow the Rule of Zero. (It is not necessary to declare the default-constructor anymore.)

Modules/Core/Common/src/itkPoolMultiThreader.cxx
```diff
@@ -80,7 +86,7 @@ class ExceptionHandler
   }
 
 private:
-  std::exception_ptr m_FirstCaughtException;
+  std::exception_ptr m_FirstCaughtException{ nullptr };
```
N-Dekker: I guess this is just a matter of style, right? An `std::exception_ptr` is already _null_ by default, not really necessary to initialize to `nullptr`: 
> A default-constructed std::exception_ptr is a null pointer; it does not point to an exception object. 

https://en.cppreference.com/w/cpp/error/exception_ptr

N-Dekker:
Just added some minor "nitpicks". If you want me to review more carefully, I'll try to have another look this evening.

Modules/Core/Common/src/itkPoolMultiThreader.cxx
```diff
@@ -193,15 +200,19 @@ PoolMultiThreader ::ParallelizeArray(SizeValueType             firstIndex,
...
+      lambda(firstIndex, firstIndex + chunkSize);
+      reporter.CompletedPixel();
+    });
 
     // now wait for the other computations to finish
     for (SizeValueType i = 1; i < workUnit; i++)
     {
-      reporter.CompletedPixel();
-
-      exceptionHandler.TryToGetFuture(m_ThreadInfoArray[i]);
```
dzenanz: We can probably remove the last use of `TryToGetFuture` and replace it by a call to  more general `TryAndCatch`.

Modules/Core/Common/src/itkPoolMultiThreader.cxx
blowekamp: done

Modules/Core/Common/src/itkPoolMultiThreader.cxx
blowekamp: done.

Modules/Core/Common/src/itkPoolMultiThreader.cxx
blowekamp: done

Modules/Core/Common/src/itkPoolMultiThreader.cxx
```diff
@@ -148,13 +138,14 @@ PoolMultiThreader::SingleMethodExecute()
...
-  ExceptionHandler exceptionHandler([this] { m_SingleMethod(&m_ThreadInfoArray[0]); });
+  ExceptionHandler exceptionHandler;
+  exceptionHandler.TryAndCatch([this] { m_SingleMethod(&m_ThreadInfoArray[0]); });
 
   // The parent thread has finished SingleMethod()
   // so now it waits for each of the other work units to finish
   for (threadLoop = 1; threadLoop < m_NumberOfWorkUnits; ++threadLoop)
   {
-    exceptionHandler.TryToGetFuture(m_ThreadInfoArray[threadLoop]);
+    exceptionHandler.TryAndCatch([&] { m_ThreadInfoArray[threadLoop].Future.get(); });
```
N-Dekker: I would prefer not to use a default capture, `[&]`. Instead I think it could be `[this, threadLoop]`. This looks preferable to me, because it treats `threadLoop` inside the lambda as a `const unsigned` value, instead of a non-const reference. See also Scott Meyers - Effective Modern C++ item "Avoid default capture modes"

Modules/Core/Common/src/itkPoolMultiThreader.cxx
```diff
@@ -193,15 +184,19 @@ PoolMultiThreader ::ParallelizeArray(SizeValueType             firstIndex,
...
+      reporter.CompletedPixel();
+    });
 
     // now wait for the other computations to finish
     for (SizeValueType i = 1; i < workUnit; i++)
     {
-      reporter.CompletedPixel();
-
-      exceptionHandler.TryToGetFuture(m_ThreadInfoArray[i]);
+      exceptionHandler.TryAndCatch([&] {
```
N-Dekker: Instead of default capture `[&]`. I would prefer something like `[this, i, &reporter]`, as I wrote before.

Modules/Core/Common/src/itkPoolMultiThreader.cxx
blowekamp: Yes, I began pondering the best ways to specify the captures with reference vs value. And thinking about the cost of passing lambdas (and other objects) by value vs reference which are done in several places. Do you have and thoughts on how these captures should be pass?

I'll update the PR to explicitly specify the captures.

N-Dekker: @blowekamp 
> Do you have and thoughts on how these captures should be pass?

I _think_ it's like this:
* If you need to modify the object inside the lambda, do capture by reference `[&obj]`.
* If the object has an expensive copy-constructor, consider doing capture by reference `[&obj]`.
* Otherwise, do capture by value `[obj]`.

Capture "by value" is often more safe than capture "by reference", because the value is treated as `const` value within the lambda (unless it's a `mutable` lambda, of course). When doing  capture "by reference" you must ensure that the reference isn't _dangling_ once the lambda is executed. 

Note that if you capture a (smart) pointer by value, you may still modify the object it points to!

Regarding the cost, I'm afraid you'd just have to measure!

> I'll update the PR to explicitly specify the captures.

Thanks, cool  😃

blowekamp: For these particular cases where the lambda is executed immediately, there is not a risk of dangling references.

PR #1682, opened 2020-03-05
ENH: Wrap ExhaustiveOptimizerv4

thewtex:


hjmjohnson:
I don't think the failures are related to this patch so approved.

image

PR #1688, opened 2020-03-07
DOC: Document python snakecase filters' kwargs

HastingsGreer:
Currently the docstring of python snakecase filters looks like:
```
>itk.extract_image_filter?
Proxy of C++ itkExtractImageFilterIRGBUC2IRGBUC2 class.
```
With this change, the docstring is much more verbose, but also much more helpful
```
>itk.extract_image_filter?
Proxy of C++ itkExtractImageFilterIRGBUC2IRGBUC2 class.
Args are Input(s) to filter.
Available Keyword Arguments:
  AbortGenerateData
  CoordinateTolerance
  Debug
  DirectionCollapseToGuess
  DirectionCollapseToIdentity
  DirectionCollapseToStrategy
  DirectionCollapseToSubmatrix
  DirectionTolerance
  ExtractionRegion
  GlobalWarningDisplay
  InPlace
  Input
  MetaDataDictionary
  MultiThreader
  NumberOfThreads
  NumberOfWorkUnits
  ObjectName
  Progress
  ReleaseDataBeforeUpdateFlag
  ReleaseDataFlag
```

Arguably some of the more obscure keyword arguments common to all filters, such as "ReleaseDataBeforeUpdateFlag" or "AbortGenerateData" could be omitted from the docstring for brevity.


HastingsGreer:
@aylward @thewtex

Wrapping/Generators/SwigInterface/igenerator.py
```diff
@@ -533,9 +533,17 @@ def generate_process_object_snake_case_functions(self, typedefs):
                 self.outputFile.write('    import itk\n')
                 self.outputFile.write('    import itkTemplate\n')
                 self.outputFile.write('    if isinstance(itk.%s, itkTemplate.itkTemplate):\n' % processObject)
-                self.outputFile.write('        %s.__doc__ = itk.%s.values()[0].__doc__\n' % (snakeCase, processObject))
+                self.outputFile.write('        filter_object = itk.%s.values()[0]\n' % (processObject))
                 self.outputFile.write('    else:\n')
-                self.outputFile.write('        %s.__doc__ = itk.%s.__doc__\n\n' % (snakeCase, processObject))
+                self.outputFile.write('        filter_object = itk.%s\n\n' % (processObject))
+                self.outputFile.write('    %s.__doc__ = filter_object.__doc__\n' % (snakeCase))
+                self.outputFile.write('    %s.__doc__ += "\\n Args are Input(s) to filter.\\n"\n' % (snakeCase))
```
thewtex: `to filter` -> `to the filter`

Wrapping/Generators/SwigInterface/igenerator.py
```diff
@@ -533,9 +533,17 @@ def generate_process_object_snake_case_functions(self, typedefs):
...
+                self.outputFile.write('    %s.__doc__ += "Available Keyword Arguments:\\n"\n' % (snakeCase))
+                self.outputFile.write('    %s.__doc__ += "".join([\n' % (snakeCase))
+                self.outputFile.write('        "  " + item[3:] + "\\n"\n')
```
thewtex: Add `self.camelCaseToSnakeCase(item[3:])` to make this more Pythonic (the interface supports both CamelCase and snake_case keyword arguments).

HastingsGreer: It's tricky: `item` only exists when the generated code runs, and the generated code doesn't have access to an IdxGenerator object because igenerator.py isn't part of the finished itk package. Should I move camelCaseToSnakeCase to itkExtras? It is already a static method.

HastingsGreer: Alternately, I probably should generate the docstring at compile time

Wrapping/Generators/SwigInterface/igenerator.py
HastingsGreer: fixed!

PR #1691, opened 2020-03-10
BUG: Module COMPLIANCE level processing improved

hjmjohnson:
There were bugs in the previous implementation
of the module COMPLIANCE level filtering.  The
COMPLIANCE level filtering was confounded with
the module ON/OFF status.

### DIFF
--- a/CMake/ITKModuleRemote.cmake
+++ b/CMake/ITKModuleRemote.cmake
@@ -172,15 +172,23 @@ function(itk_fetch_module _name _description)
     set(MODULE_COMPLIANCE_LEVEL ${DEFAULT_MODULE_COMPLIANCE_LEVEL})
   endif()
 
+  set(Module_${_name}_REMOTE_COMPLIANCE_LEVEL ${MODULE_COMPLIANCE_LEVEL} CACHE INTERNAL "Variable to indicate the Module_${_name} compliance level")
+
+  if(NOT DEFINED Module_${_name} )
+    option(Module_${_name} "(Remote-${MODULE_COMPLIANCE_LEVEL}) ${_description}" OFF)
+  else()
+    # If Module_${_name} is set manually, put it's value in the CACHE
+    option(Module_${_name} "(Remote-${MODULE_COMPLIANCE_LEVEL}) ${_description}" ${Module_${_name}})
+  endif()
+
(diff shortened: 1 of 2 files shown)


hjmjohnson:
Related to #1675

This has not been rigorously tested yet. 

@aylward Would you please test this to determine if it now works as expected?

CMake/ITKModuleRemote.cmake
```diff
@@ -230,9 +238,6 @@ function(itk_fetch_module _name _description)
     if(DEFINED REMOTE_GIT_TAG_${_name})
       set_property(CACHE REMOTE_GIT_TAG_${_name} PROPERTY TYPE INTERNAL)
     endif()
-    if(DEFINED Module_${_name})
```
hjmjohnson: This was the primary bug.  If the module was off, it was also marked as an internal variable.  THAT IS A MISTAKE!

blowekamp:
I have not groked through the code. What was the incorrect old behavior and what is the new behavior it was corrected to?

hjmjohnson:
> I have not groked through the code. What was the incorrect old behavior and what is the new behavior it was corrected to?
> 
> * Please consider improving the commit message.

See comment related to line 233 above.  Internal variables are not visible, and this line was setting all Module_${_name} variables that were "OFF" to also be "INTERNAL", thus preventing them from being visible to turn "ON".

blowekamp:
So, modules which exceed the compliance level should be non-advanced options, while those below the selected compliance level are advanced options?

hjmjohnson:
> So, modules which exceed the compliance level should be non-advanced options, while those below the selected compliance level are advanced options?

That is the way it is currently implemented.  I open to other suggestions.  I feel that having all the modules listed all the time is not the correct answer.  You can still select modules with lower compliance, but you must be in advanced mode to view them.

SimonRit:
I quickly checked locally and it behaves as you describe. It looks good to me. Currently, with ITK_MINIMUM_COMPLIANCE_LEVEL set to 4, only one module shows up, SphinxExamples. Have you considered lowering it to 3 to show four modules?

dzenanz:
Yes, 4 is a very high bar, hiding many useful modules. 3 would be a much better default.

hjmjohnson:
> Yes, 4 is a very high bar, hiding many useful modules. 3 would be a much better default.

Done.

blowekamp:
But the internal ITK modules options are advanced options...

PR #1694, opened 2020-03-10
BUG: restoring libPNG branch name for UpdateFromUpstream.sh

dzenanz:
The change was accidental commit in
b644f1a3818c9dbbc9aa5751a311f01a109f4174.


blowekamp:
Thanks for the follow up!

I had the fix in my branch locally, it looks like when I tries to do a force push to your PR it failed, and I didn't notice.

PR #1696, opened 2020-03-11
BUG: Fix BUILD_SHARED_LIBS state consistency for HDF5

### LINKED ISSUE #1658: BUILD_SHARED_LIBS forced to off on Win32 causing wrong system hdf5 libs to be selected
### Description

When building shared libraries for ITK and using system hdf5, static hdf5 libraries are selected under Windows.

### Steps to Reproduce

1. Install HDF5 manually with cmake (in my case 1.10.3)
2. With CMake, select BUILD_SHARED_LIBS=ON and ITK_USE_SYSTEM_HDF5
3. Configure

You will see that BUILD_SHARED_LIBS is forced to OFF in ITKSetStandardCompilerFlags on Windows. Then, in Modules/ThirdParty/HDF5/itk-module-init.cmake and Modules/ThirdParty/HDF5/CMakeLists.txt:29, the wrong libraries are searched and selected.

### Expected behavior

Selection of the HDF5 shared libraries when BUILD_SHARED_LIBS=ON. 

### Actual behavior

Selection of the HDF static libraries.

### Reproducibility

100%

### Versions

ITK version 5.0.1

### Environment

Windows 10 64 bits
Visual Studio 2019
CMake 3.15

kenavolic:
Reset BUILD_SHARED_LIBS to the root project initial state in case
this option is modified before processing the configuration of
a third-party module.

Closes #1658

### DIFF
--- a/Modules/ThirdParty/HDF5/CMakeLists.txt
+++ b/Modules/ThirdParty/HDF5/CMakeLists.txt
@@ -1,6 +1,12 @@
 project(ITKHDF5)
 set(ITKHDF5_THIRD_PARTY 1)
 
+# Reset BUILD_SHARED_LIBS to ITK_BUILD_SHARED_LIBS
+# that keeps track of the initial state of the root
+# project BUILD_SHARED_LIBS option (which could be
+# modified before reaching this point)
+set(BUILD_SHARED_LIBS ${ITK_BUILD_SHARED_LIBS})
+
 if (BUILD_SHARED_LIBS)
    add_definitions(-DH5_BUILT_AS_DYNAMIC_LIB=1)
 endif()
--- a/Modules/ThirdParty/HDF5/itk-module-init.cmake
+++ b/Modules/ThirdParty/HDF5/itk-module-init.cmake
@@ -1,7 +1,8 @@
 option(ITK_USE_SYSTEM_HDF5 "Use an outside build of HDF5." ${ITK_USE_SYSTEM_LIBRARIES})
 mark_as_advanced(ITK_USE_SYSTEM_HDF5)
+
 if(ITK_USE_SYSTEM_HDF5)
-  if(BUILD_SHARED_LIBS)
+  if(ITK_BUILD_SHARED_LIBS)
     find_package(HDF5 QUIET NO_MODULE COMPONENTS CXX C shared)
   else()
     find_package(HDF5 QUIET NO_MODULE COMPONENTS CXX C static)


thewtex:
These modules likely do not have the symbol export specifications setup for building as shared libraries.

kenavolic:
My changes broke windows CI due to issue with jpeg lin in shared mode. I'll investigate this as soon as possible!

kenavolic:
By inspecting the ci failures, it seems that some of the 3rd parties did not export any symbols in shared mode. Thus I disable the shared mode for these libs on Windows. This might not be the behavior you expect since there are some features like `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` now.

kenavolic:
Moreover, I have some unexpected linker failure with vnl/vxl libraries with Visual Studio 2019:
```
6>itkvnl-5.1.lib(itkvnl-5.1.dll) : error LNK2005: "public: double & __cdecl vnl_matrix_fixed<double,6,6>::operator()(unsigned int,unsigned int)" (??R?$vnl_matrix_fixed@N$05$05@@QEAAAEANII@Z) already defined in ITKTestKernel-5.1.lib(itkTestDriverInclude.obj)
6>itkvnl-5.1.lib(itkvnl-5.1.dll) : error LNK2005: "public: double const & __cdecl vnl_matrix_fixed<double,6,6>::operator()(unsigned int,unsigned int)const " (??R?$vnl_matrix_fixed@N$05$05@@QEBAAEBNII@Z) already defined in ITKTestKernel-5.1.lib(itkTestDriverInclude.obj)
6>itkvnl-5.1.lib(itkvnl-5.1.dll) : error LNK2005: "public: class vnl_matrix_fixed<double,6,6> & __cdecl vnl_matrix_fixed<double,6,6>::set_identity(void)" (?set_identity@?$vnl_matrix_fixed@N$05$05@@QEAAAEAV1@XZ) already defined in ITKTestKernel-5.1.lib(itkTestDriverInclude.obj)
6>itkvnl-5.1.lib(itkvnl-5.1.dll) : error LNK2005: "public: void __cdecl vnl_matrix_fixed<double,6,6>::print(class std::basic_ostream<char,struct std::char_traits<char> > &)const " (?print@?$vnl_matrix_fixed@N$05$05@@QEBAXAEAV?$basic_ostream@DU?$char_traits@D@std@@@std@@@Z) already defined in ITKTestKernel-5.1.lib(itkTestDriverInclude.obj)
1
```

hjmjohnson:
@dzenanz I don't have easy access to windows, and I am not good at windows shared library problems.  I don't have useful information to add to this PR.

dzenanz:
OK I will build it locally with VS2019 and report back.

dzenanz:
The error is slightly different for me:
```text
2>itkvnl-5.1.lib(itkvnl-5.1.dll) : error LNK2005: "public: double & __cdecl vnl_matrix_fixed<double,6,6>::operator()(unsigned int,unsigned int)" (??R?$vnl_matrix_fixed@N$05$05@@QEAAAEANII@Z) already defined in ITKTestKernel-5.1.lib(itkTestDriverInclude.obj)
2>itkvnl-5.1.lib(itkvnl-5.1.dll) : error LNK2005: "public: double const & __cdecl vnl_matrix_fixed<double,6,6>::operator()(unsigned int,unsigned int)const " (??R?$vnl_matrix_fixed@N$05$05@@QEBAAEBNII@Z) already defined in ITKTestKernel-5.1.lib(itkTestDriverInclude.obj)
2>itkvnl-5.1.lib(itkvnl-5.1.dll) : error LNK2005: "public: class vnl_matrix_fixed<double,6,6> & __cdecl vnl_matrix_fixed<double,6,6>::set_identity(void)" (?set_identity@?$vnl_matrix_fixed@N$05$05@@QEAAAEAV1@XZ) already defined in ITKTestKernel-5.1.lib(itkTestDriverInclude.obj)
2>itkvnl-5.1.lib(itkvnl-5.1.dll) : error LNK2005: "public: void __cdecl vnl_matrix_fixed<double,6,6>::print(class std::basic_ostream<char,struct std::char_traits<char> > &)const " (?print@?$vnl_matrix_fixed@N$05$05@@QEBAXAEAV?$basic_ostream@DU?$char_traits@D@std@@@std@@@Z) already defined in ITKTestKernel-5.1.lib(itkTestDriverInclude.obj)
2>   Creating library C:/Dev/ITK-19DLL/lib/RelWithDebInfo/ITKStatisticsTestDriver.lib and object C:/Dev/ITK-19DLL/lib/RelWithDebInfo/ITKStatisticsTestDriver.exp
2>C:\Dev\ITK-19DLL\bin\RelWithDebInfo\ITKStatisticsTestDriver.exe : fatal error LNK1169: one or more multiply defined symbols found
```

thewtex:
> By inspecting the ci failures, it seems that some of the 3rd parties did not export any symbols in shared mode. Thus I disable the shared mode for these libs on Windows. This might not be the behavior you expect since there are some features like CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS now.

Yes, this is why these libraries are intentionally not built as shared.

kenavolic:
As modifying the BUILD_SHARED_LIBS state for all third parties was too ambitious for a first ITK PR and with my current workload (no time for an in-depth impact analysis), I decided to focus only on the issue with HDF5. Here are the links towards the experimental builds:

- [Build with shared libs, internal hdf5, CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=OFF](https://open.cdash.org/buildSummary.php?buildid=6416136)
- [Build with shared libs, internal hdf5, CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=ON](https://open.cdash.org/buildSummary.php?buildid=6415951)
- [Build with shared libs, system hdf5, CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=OFF](https://open.cdash.org/buildSummary.php?buildid=6415920)

Neumann-A:
Can somebody tell me why `ITK_BUILD_SHARED_LIBS` is linked to hdf5 being static or shared? Is there any good reason to do that in ITK? Normally I would suspect that it does not matter if hdf5 is linked in static or dynamically however ITK will be build.

dzenanz:
In order to fix #1658?

Neumann-A:
If I have an external HDF5 it might be built statically or dynamically.  It should not matter how HDF5 is built for ITK.(which is obviously indicated by the third find_package call in the file) 
The correct way would be to always only call `find_package(HDF5 QUIET NO_MODULE COMPONENTS CXX C)`. Depending wehter ITK **prefers** to link HDF5 dynamically it could check if the dynamic HDF5 target is available and if not link against the static target (If non is found the whole module is not found). The order here which is first (dynamic or static) is just defined but what ITK wants to prefer to link. Having code for three `find_package` calls is just wrong considering you have basically the same conditional code in the `CMakeLists.txt`
You probably want to change part of this monster:
https://github.com/InsightSoftwareConsortium/ITK/blob/7900a8b6d63ec8201405eff3b46c8d2a4e563c45/Modules/ThirdParty/HDF5/CMakeLists.txt#L34-L40 
e.g.
```
  if(ITK_BUILD_SHARED_LIBS AND HDF5_BUILD_SHARED_LIBS)
    list(APPEND ITKHDF5_LIBRARIES hdf5::hdf5_cpp-shared hdf5::hdf5-shared)
  else()
    list(APPEND ITKHDF5_LIBRARIES hdf5::hdf5_cpp-static hdf5::hdf5-static)
  endif()
```
(Allows ITK:HDF5 || shared:shared | static:static | shared:static || Disallows || static:shared)
or provide an `hdf5::hdf5 ALIAS` target which does not care whether hdf5 is static or dynamic and reflects ITK linkage preference

dzenanz:
@kenavolic do you have time and will to look into @Neumann-A's suggestion? I assume this is reported as #1799?

Neumann-A:
@dzenanz: #1799 is another build error but I also mentioned the issue above. The solution for #1799 is also in there

kenavolic:
@Neumann-A Thanks for your feedback. This PR just fixes a state inconsistency in the `BUILD_SHARED_LIBS` variable. I've got to say I was focused on the bug and missed the big picture. 

First of all, I have no clue why there is such a backward transitivity in libraries types.

What could be achieved is somthing like that:
- If only static libraries are here, link to HDF5 static libraries (HDF5 configured with `BUILD_SHARED_LIBS=OFF`)
- If both are here, set a preference in ITK (HDF5 configured with `BUILD_SHARED_LIBS=ON`)
- If only shared libraries are here, link to HDF5 shared libraries (HDF5 configured with `ONLY_SHARED_LIBS=ON`)

For a project to include HDF5, it has to take into account an inclusion through the path to a configuration file (`-DHDF5_DIR`) or through a package search (`FindHDF5` cmake module). The main issue is that both will not populate the same variables.

For some version of HDF5 configuration file, `find_package(HDF5 QUIET NO_MODULE COMPONENTS CXX C)` without the `static/shared` component won't populate any target related variables. The only variable of use would thus be `${HDF5_PACKAGE_NAME}_BUILD_STATIC_LIBS` and `${HDF5_PACKAGE_NAME}_BUILD_SHARED_LIBS` than can let you decide according to their states what to include according to ITK preference with another call to find_package(HDF5` QUIET NO_MODULE COMPONENTS CXX C Comp)` with the component this time.

On the other side,  the `FindHDF5` (at least for cmake 3.15) will populate `HDF5_CXX_LIBRARIES` according to a variable `HDF5_USE_STATIC_LIBRARIES` to be set before.

I think you are right in the sense that shared ITK libs should not imply shared third-party libraries but the modification could be a little bit more complicated to handle all cases.

I gave it a try though and I ended up with an issue at runtime. Here is the scenario:
* `ITKIOHDF5` is built as shared and static system HDF5 libraries are used,
* I have a libfoo.so that relies as well on HDF5,
* And then an executable that relies on libfoo.so and libITKIOHDF5.so.

When I run the executable, I get an exception due to a double initialization in an HDF5 singleton (H5DataSpace.cpp):
```
DataSpace* DataSpace::getConstant() {
...
    if (ALL_ == 0)
        ALL_ = new DataSpace(H5S_ALL);
    else
        throw DataSpaceIException("DataSpace::getConstant", "DataSpace::getConstant is being invoked on an allocated ALL_");
    return(ALL_);
}
```
I tried to track down the problem, and I found that the issue comes from ` libITKIOHDF5.so` that allows deferred binding of the `getConstant` symbol at runtime (`00000000006d370 <_ZN2H59DataSpace11getConstantEv@plt>`). Thus at runtime, as `libfoo.so` is loaded first, its `getConstant` "version" is used first for its own initialization and then for `libITKIOHDF5.so` causing the exception.

I tried to solve the issue with `target_link_options(ITKIOHDF5 PUBLIC -Bsymbolic)` as suggested [here](https://stackoverflow.com/questions/7201667/ld-magically-overrides-statically-linked-symbols), but I must be missing something as it does not fix the problem.

In conclusion, doing what you want is what should be aimed to but I don't know if it is that easy due to possible side effects! 

Regards,

Neumann-A:
>For some version of HDF5 configuration file

would be solved by how e.g. VTK bundles `find<PackageName>.cmake` modules

> Thus at runtime, as libfoo.so is loaded first, its getConstant "version" is used first for its own initialization and then for libITKIOHDF5.so causing the exception.

Funny problem, but it feels like it is not ITK's issue to solve that. The person building the executable should know what can be mixed and what cannot be mixed. In vcpkg for example that problem does never exist since there is only always one type of library allowed for a given triplet

Issue #1700, opened 2020-03-12
FindPython3.cmake unavailable before CMake 3.12


SimonRit:
### Description

I have a warning with cmake v3.11 since 07705bd (@thewtex), see [RTK dashboard](https://my.cdash.org/viewConfigure.php?buildid=1788739). That's because FindPython3.cmake is available since CMake 3.12 only. I'm not sure what's the strategy to fix it, either raise cmake minimal version or find a workaround backward compatible with cmake 3.10. I'm happy to help if I can...

### Steps to Reproduce

Configure with cmake 3.11.

### Expected behavior

Smooth cmake configuration

### Actual behavior

Cmake warning 

```
CMake Warning at D:/src/itk/ITK-master/CMake/ITKModuleExternal.cmake:22 (find_package):
  By not providing "FindPython3.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Python3", but
  CMake did not find one.

  Could not find a package configuration file provided by "Python3" with any
  of the following names:

    Python3Config.cmake
    python3-config.cmake

  Add the installation prefix of "Python3" to CMAKE_PREFIX_PATH or set
  "Python3_DIR" to a directory containing one of the above files.  If
  "Python3" provides a separate development package or SDK, be sure it has
  been installed.
Call Stack (most recent call first):
  CMakeLists.txt:257 (include)
```

```none
```

### Reproducibility

100%

### Versions

07705bd

### Environment

Tested on Windows but should occur on any platform.

dzenanz:
CC: @hjmjohnson.

thewtex:
@SimonRit thanks for the report.

It is too much of a burden for the community to install a newer CMake when they want to build ITK, so we cannot bump the CMake minimum version.

We could vendor the FindPython3.cmake module. The version from 3.16 or later is the one that has the least worst behavior on Windows. We have to ensure that the module code is not using any CMake features that were introduced past 3.10.

thewtex:
Closing pre #1072

PR #1702, opened 2020-03-12
COMP: Conditional python3 package finding

hjmjohnson:
The code needed to find python3 packages has been consolidated
so that it may be reused for external modules.

This attempts to address the issues identified in #1700 by re-using a previous paradigm for the case of external module dependency on python.

@SimonRit This solution has not been tested with RTK

## PR Checklist

### DIFF
--- a/CMake/ITKModuleExternal.cmake
+++ b/CMake/ITKModuleExternal.cmake
@@ -11,20 +11,8 @@ if(NOT EXISTS ${ITK_CMAKE_DIR}/ITKModuleMacros.cmake)
   message(FATAL_ERROR "Modules can only be built against an ITK build tree; they cannot be built against an ITK install tree.")
 endif()
 
-if(ITK_WRAP_PYTHON)
-  if(DEFINED Python3_EXECUTABLE)
-    set(_specified_Python3_EXECUTABLE ${Python3_EXECUTABLE})
-  endif()
-  find_package(Python3 COMPONENTS Interpreter Development)
-  set(Python3_EXECUTABLE ${_specified_Python3_EXECUTABLE} CACHE INTERNAL
-    "Path to the Python interpreter" FORCE)
-else()
-  find_package(Python3 COMPONENTS Interpreter)
-endif()
-if(NOT Python3_EXECUTABLE AND _Python3_EXECUTABLE)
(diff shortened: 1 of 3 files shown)


CMake/ITKModuleExternal.cmake
```diff
@@ -11,20 +11,8 @@ if(NOT EXISTS ${ITK_CMAKE_DIR}/ITKModuleMacros.cmake)
...
-    "Path to the Python interpreter" FORCE)
-else()
-  find_package(Python3 COMPONENTS Interpreter)
-endif()
-if(NOT Python3_EXECUTABLE AND _Python3_EXECUTABLE)
-  set(Python3_EXECUTABLE ${_Python3_EXECUTABLE} CACHE INTERNAL
-    "Path to the Python interpreter" FORCE)
-endif()
+set(PYTHON_DEVELOPMENT_REQUIRED ${ITK_WRAP_PYTHON})
+include(${CMAKE_CURRENT_SOURCE_DIR}/ITKSetPython3Vars.cmake)
```
SimonRit: Thanks. This does not seem to work when building an external module against an already built ITK, see [here](https://my.cdash.org/buildSummary.php?buildid=1789345). Replacing this line by ` include(ITKSetPython3Vars)` seems to help but a compilation is on-going on my side, I'll send you the dashboard link when it's done.

hjmjohnson: I think that is a chicken and egg problem.  We need to update the ITK environment on the docker image that is used to build external modules.

hjmjohnson: OH!! The path is also wrong!  I see that now.

SimonRit: My compilation [works](https://my.cdash.org/buildSummary.php?buildid=1789429) with this change but I haven't checked the compilation from within ITK yet.

hjmjohnson: @SimonRit I'm dealing with COVID-19 implications of all lectures going virtual, and ISBI 2020 becoming a completely virtual meeting.  I'd really appreciate your response if you feel that your tests are working, then I'll merge after that.  



Thanks,
Hans

PR #1708, opened 2020-03-16
DOC: Remove FindPython3 robust remark

thewtex:
The FindPython3 module has required half a dozen work arounds for broken
behavior and is still causing a failing Windows CI build.


thewtex:
@hjmjohnson it is no fault of yours. We started down module adoption, and it is hard to go back, but we can avoid misleading other projects into the difficulties we have experienced.

PR #1710, opened 2020-03-19
Macro ignore extra semi colon and use anonymous exception passing

hjmjohnson:
Update exception macros to use anonymous exceptions for clang.

### DIFF
--- a/Modules/Core/Common/include/itkMacro.h
+++ b/Modules/Core/Common/include/itkMacro.h
@@ -65,6 +65,13 @@ namespace itk
  * avoiding compile-time warnings. */
 #define itkNotUsed(x)
 
+/** The `do {} while(0)`  idiom is commonly employed for
+ *  allowing a macro to be used anywhere that a statement
+ *  is expected, and to enforce consistent use of ; after
+ *  macro */
+// clang-format off
+#  define ITK_NOOP_STATEMENT    do {} while(0)
+// clang-format on
 
 // Define ITK_PRAGMA macro.
 //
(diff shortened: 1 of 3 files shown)


Modules/Core/Common/include/itkMacro.h
```diff
@@ -120,9 +120,18 @@ namespace itk
 #  define ITK_GCC_PRAGMA_DIAG_PUSH() ITK_GCC_PRAGMA_DIAG(push)
 #  define ITK_GCC_PRAGMA_DIAG_POP() ITK_GCC_PRAGMA_DIAG(pop)
 #else
-#  define ITK_GCC_PRAGMA_DIAG(x)
-#  define ITK_GCC_PRAGMA_DIAG_PUSH()
-#  define ITK_GCC_PRAGMA_DIAG_POP()
+#  define ITK_GCC_PRAGMA_DIAG(x)                                                                                       \
```
blowekamp: It looks like "\" are missing here to me. If it is style compliant, putting it on one line may be clearer.

hjmjohnson: I don't understand about the "".  Do you mean the end of line continuations "\\"?  They are there, but off the standard github display.  You need to scrol to the right.

I did write these as one liners, but clang-format expanded it to 3 lines.  Perhaps this is a good use of a comment to ignore clang-format.

blowekamp: Sorry for the missing char! 😞 

I see the "\" now! That is very annoying that it is off the end of the GitHub display.

dzenanz:
... and failed Windows build supports that :(

hjmjohnson:
> I feel like there are more instances of `itkDebugMacro` and friends than the few which are part of this PR.

I'm not 100% sure what you are refering to, but I did make a mistake and added this paradigm in a pragma push/pop situation that it should not have been added to.  That is will be resolved in a force push in a few minutes.

PR #511, opened 2019-02-16
ENH: Improve Thresholding module filters' coverage.

jhlegarreta:
Improve the ITKThresholding module filters' coverage:
- Exercise basic object methods. Remove redundant calls to print the
  filter.
- Exercise the Set/Get methods using the TEST_SET_GET_VALUE and On/Off
  methods using the TEST_SET_GET_BOOLEAN macro.
- Use the TRY_EXPECT_NO_EXCEPTION macro for all filter update calls to
  save typing try/catch blocks.
- Use TRY_EXPECT_EXCEPTION to try/catch expected exceptions when
  possible/appropriate.
- Refactor the tests to accept (more) input parameters to check the
  classes under a broader range of conditions. In many cases, the
  parameters accepted belong to inherited propertis (e.g.
  itk::HistogramThresholdImageFilter: m_NumberOfHistogramBins,
  m_AutoMinimumMaximum, m_MaskOutput, m_MaskValue). Keep the default
  values of the class for existing tests.
- Make the tests quantitative: compare the compute threshold to the
  expected one.
- Add the hash code to compare to the baseline for new tests.
- Remove unnecessary print messages.
- Style changes: define the image dimension as a constant; use a typedef
  for the pixel type; use a variable to set the output's inside/outside
  value so that its Set/Get methods can be clearly tested; try to use the
  same sequence of sentences across all tests; change '\n' to std::endl.
- Improve the messages in the threshold regression tests to make the
  consistent across the tests/the entire toolkit, and to provide rich
  information.
- Finished the tests with a "Test finished" message; helps telling in the
  the dashboard whether the failure has happened in the test or the
  comparison to the baseline.
- In itkBinaryThresholdImageFilterTest.cxx, set the functor before
  updating the filter.
- Correct the arguments passed to SetSigmaFactor and SetNumberOfIterations
  in itkKappaSigmaThresholdImageFilterTest.cxx (the expected sigmaFactor
  as defined by the order in the CMakeLists command/usage explanation was
  being set to the number of iterations of the filter and vice-versa).
  Update the corresponding baselines accordingly.

Take advantage of the commit to use C++11 features.

Recovered from gerrit:
Change-Id: I9d6050d9e474b17fe3639faa7a5b16c052440e21
https://review.source.kitware.com/%23/c/22134/

PR #1718, opened 2020-03-21
BUG: Have a single histogram threshold calculator for intermodes threshold

jhlegarreta:
Fix the threshold calculator inheritance issue in
`itk::IntermodesThresholdImageFilter` by having a single histogram
threshold calculator variable.

The class was declaring a histogram threshold calculator under the
`m_IntermodesCalculator` ivar, which was of type
`itk::IntermodesThresholdCalculator` type. At the same time, the class was
inheriting from `itk::HistogramThresholdImageFilter`, which was declaring
its own histogram threshod calculator `m_Calculator`, which was of type
`itk::HistogramThredholdCalculator`.

This was making the `itk::IntermodesThresholdImageFilter` objects to have
two histogram calculator instances, and given that the `GenerateData`
method was implemented in the parent class
(`itk::HistogramThresholdImageFilter`) and making use of the
`m_Calculator` ivar, this was yielding a runtime error.

Fixes the runtime exception error mentoned in [this
comment](https://github.com/InsightSoftwareConsortium/ITK/pull/511#issuecomment-497877457).

## PR Checklist

### DIFF
--- a/Modules/Filtering/Thresholding/include/itkHistogramThresholdImageFilter.h
+++ b/Modules/Filtering/Thresholding/include/itkHistogramThresholdImageFilter.h
@@ -20,8 +20,8 @@
 #define itkHistogramThresholdImageFilter_h
 
 #include "itkImageToImageFilter.h"
+#include "itkHistogramThresholdImageFilterBase.h"
 #include "itkHistogram.h"
-#include "itkHistogramThresholdCalculator.h"
 
 namespace itk
 {
@@ -36,16 +36,6 @@ namespace itk
  * applies that threshold to the input image using the
  * BinaryThresholdImageFilter.
  *
- * The filter also has the option of providing a mask, in which case
- * the histogram and therefore the threshold is computed from the
- * parts of the mask with values indicated by MaskValue. The output
- * image is, by default, masked by the same image. This output
- * masking can be disabled using SetMaskOutput(false). Note that there
- * is an inconsistency here. The MaskImageFilter (used internally)
- * masks by non zero values, where as the MaskedImageToHistogramFilter
- * uses explicit values. If this does not match your usage then the
- * output masking will need to be managed by the user.
- *
  * \author Richard Beare. Department of Medicine, Monash University,
  * Melbourne, Australia.
  * \author Gaetan Lehmann. Biologie du Developpement et de la Reproduction, INRA de Jouy-en-Josas, France.
@@ -59,14 +49,15 @@ namespace itk
  */
 
 template <typename TInputImage, typename TOutputImage, typename TMaskImage = TOutputImage>
-class ITK_TEMPLATE_EXPORT HistogramThresholdImageFilter : public ImageToImageFilter<TInputImage, TOutputImage>
+class ITK_TEMPLATE_EXPORT HistogramThresholdImageFilter
+  : public HistogramThresholdImageFilterBase<TInputImage, TOutputImage, TMaskImage>
 {
 public:
   ITK_DISALLOW_COPY_AND_ASSIGN(HistogramThresholdImageFilter);
 
   /** Standard Self type alias */
   using Self = HistogramThresholdImageFilter;
-  using Superclass = ImageToImageFilter<TInputImage, TOutputImage>;
+  using Superclass = HistogramThresholdImageFilterBase<TInputImage, TOutputImage>;
   using Pointer = SmartPointer<Self>;
   using ConstPointer = SmartPointer<const Self>;
 
@@ -74,7 +65,7 @@ public:
   itkNewMacro(Self);
 
   /** Runtime information support. */
-  itkTypeMacro(HistogramThresholdImageFilter, ImageToImageFilter);
+  itkTypeMacro(HistogramThresholdImageFilter, HistogramThresholdImageFilterBase);
 
   using InputImageType = TInputImage;
   using OutputImageType = TOutputImage;
@@ -88,7 +79,6 @@ public:
   /** Image related type alias. */
   using InputImagePointer = typename InputImageType::Pointer;
   using OutputImagePointer = typename OutputImageType::Pointer;
-  using MaskImagePointer = typename MaskImageType::Pointer;
 
   using InputSizeType = typename InputImageType::SizeType;
   using InputIndexType = typename InputImageType::IndexType;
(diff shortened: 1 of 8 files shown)


jhlegarreta:
This is the approach I came up with to solve the issue raised in the comment referenced in the commit message.

I fear the current `ITKThresholding` module tests will not allow to evidence what this patch set tries to address, but if one cherry-picks the commit into PR #511, and makes the necessary adjustments (the `expectedThredshold` parameter tests are reached now and the baseline images can be generated, so they must be adjusted, and the ancestor class needs to be changed for the `ITK_EXERCISE_BASIC_OBJECT_METHODS` for the `itkIntermodesThresholdImageFilter` tests), it can be checked that the tests affected by this issue (e.g. `itkIntermodesThresholdImageFilterTest`, `itkIntermodesThresholdImageFilterTestNoInterMode`, `itkIntermodesThresholdImageFilterTestShort`) pass.

I tried to avoid the duplicated code in the `GenerateData` method, but since the `itk::Statistics::ImageToHistogramFilter` and the `itk::Statistics::MaskedImageToHistogramFilter` classes do not hae an immediate common ancestor, I did not quite find the way to do it, so unless there is some proposal to do that, and if CI builds turn out to be successful, this is ready to be merged.

I will rebase PR #511 on top of `master` once this gets merged, and address the `m_Min`/`m_Max` ivar initialization issue in a separate PR to again rebase PR #511 on top of `master` once that gets solved.

Reviews are welcome.

Modules/Filtering/Thresholding/include/itkHistogramThresholdImageFilter.h
```diff
@@ -59,22 +50,23 @@ namespace itk
  */
 
 template <typename TInputImage, typename TOutputImage, typename TMaskImage = TOutputImage>
-class ITK_TEMPLATE_EXPORT HistogramThresholdImageFilter : public ImageToImageFilter<TInputImage, TOutputImage>
+class ITK_TEMPLATE_EXPORT HistogramThresholdImageFilter
+  : public HistogramThresholdImageFilterBase<TInputImage, TOutputImage>
```
blowekamp: Shouldn't the `TMaskImage` template parameter be passes to the parent class.

Modules/Filtering/Thresholding/include/itkHistogramThresholdImageFilterBase.h
```diff
@@ -0,0 +1,202 @@
...
+ * is an inconsistency here. The MaskImageFilter (used internally)
+ * masks by non zero values, where as the MaskedImageToHistogramFilter
+ * uses explicit values. If this does not match your usage then the
+ * output masking will need to be managed by the user.
+ *
+ * \author Richard Beare. Department of Medicine, Monash University,
+ * Melbourne, Australia.
+ * \author Gaetan Lehmann. Biologie du Developpement et de la Reproduction, INRA de Jouy-en-Josas, France.
+ *
+ * This implementation was taken from the Insight Journal paper:
```
blowekamp: "This implementation was based on the Insight Journal paper:"

It also looks like you are authoring this class :)

Modules/Filtering/Thresholding/include/itkIntermodesThresholdImageFilter.h
```diff
@@ -90,61 +90,57 @@ class IntermodesThresholdImageFilter : public HistogramThresholdImageFilter<TInp
...
+  using HistogramConstPointer = typename HistogramType::ConstPointer;
+  using HistogramSizeType = typename HistogramType::SizeType;
+  using HistogramMeasurementType = typename HistogramType::MeasurementType;
+  using HistogramMeasurementVectorType = typename HistogramType::MeasurementVectorType;
+
   using CalculatorType = IntermodesThresholdCalculator<HistogramType, InputPixelType>;
 
+  /** Set/Get the calculator to use to compute the threshold */
+  itkSetObjectMacro(Calculator, CalculatorType);
+  itkGetModifiableObjectMacro(Calculator, CalculatorType);
```
blowekamp: The parent class has a public Set/Get Calculator method. This method should be overloaded. The set method likely needs a dynamic_cast of the calculator.

Modules/Filtering/Thresholding/include/itkIntermodesThresholdImageFilter.hxx
```diff
@@ -0,0 +1,159 @@
...
+  m_Calculator->SetMaximumSmoothingIterations(10000);
+  m_Calculator->SetUseInterMode(true);
+}
+
+template <typename TInputImage, typename TOutputImage, typename TMaskImage>
+void
+IntermodesThresholdImageFilter<TInputImage, TOutputImage, TMaskImage>::SetMaximumSmoothingIterations(
+  SizeValueType maxSmoothingIterations)
+{
+  m_Calculator->SetMaximumSmoothingIterations(maxSmoothingIterations);
```
blowekamp: Does this cause the modify time of the image filter class to be updated?

The simple way to do this would be to copy the code in the set/get macro which check if the value is the same, if they are different then set the modify time. The more complicated way would involve modify the parent's GetMTime method to return the max of the MTime of itself, and the m_Calculator.

Modules/Filtering/Thresholding/include/itkIntermodesThresholdImageFilter.hxx
```diff
@@ -0,0 +1,159 @@
...
+  return m_Calculator->GetUseInterMode();
+}
+
+template <typename TInputImage, typename TOutputImage, typename TMaskImage>
+void
+IntermodesThresholdImageFilter<TInputImage, TOutputImage, TMaskImage>::GenerateData()
+{
+  if (m_Calculator.IsNull())
+  {
+    itkExceptionMacro(<< "No threshold calculator set.");
```
blowekamp: This type of check should happen the the "VerifyPreconditions" method.

blowekamp:
Looks like a very nice refactor to a base class. However it duplicates code into IntermodesThresholdImageFilter::Gernate data. Would it be simpler to overload the Set/Get Calculator methods to set the calculator ivars in the parent and Intermodes filter to be the same calculator  object ( although in HistogramThresholdImageFilter would point to the parent class, while Intermodes would point to he derived. )?

Modules/Filtering/Thresholding/include/itkIntermodesThresholdImageFilter.h
blowekamp: Consider the alternative solution of overloading these methods so that HistogramThresholdImageFilter::m_Calclulcator and this::m_ItermodeCalculator point to the same instance.

The overloaded HistogramThresholdImageFilter::SetCalculator method would need to dynamic_cast to the Intermodes calculator, the call HistogramThresholdImageFilter::SetCalculator and this::SetCalculator with the dynamic casted object.

jhlegarreta: Not sure how you would create the `m_Calculator` object in [here](https://github.com/InsightSoftwareConsortium/ITK/pull/1718/files#diff-a4eb676b8dccbc553855aabae3ec343fR34), though

Modules/Filtering/Thresholding/include/itkHistogramThresholdImageFilterBase.h
jhlegarreta: Thanks for considering it Brad, but I think the PR/commit authorship is enough credit.

Modules/Filtering/Thresholding/include/itkIntermodesThresholdImageFilter.h
blowekamp: HistogramThresholdImageFilter still have its m_Calculator ivar.
IntermodesThresholdImageFilter maintains its original m_IntermodesCalculator with derived type.

Both SetCalculator methods would set both variables.

Modules/Filtering/Thresholding/include/itkIntermodesThresholdImageFilter.hxx
jhlegarreta: Done in 624c503.

Modules/Filtering/Thresholding/include/itkHistogramThresholdImageFilter.h
jhlegarreta: My bad. Thanks Brad. Done in 624c503.

Modules/Filtering/Thresholding/include/itkIntermodesThresholdImageFilter.hxx
jhlegarreta: Opened issue #1720.

Modules/Filtering/Thresholding/include/itkIntermodesThresholdImageFilter.h
jhlegarreta: Thanks for the comments Brad. Please check commit 624c503. Especially the change in [here](https://github.com/InsightSoftwareConsortium/ITK/pull/1718/files#diff-e8fefc8c747bbff3a35dbb5c9fbfe84eL116) as I did not have a clear criterion to continue using an `itkGetModifiableObjectMacro` macro, and the base class `GetCalculator` implementation in [here](https://github.com/InsightSoftwareConsortium/ITK/pull/1718/files#diff-acc13746dfc9bfba2ffc93d33947b390R186).

blowekamp:
May I take a stab at addressing the issue, as I don't believe the base class refactor is nessacary to address?

dzenanz:
If I remember correctly, he was looking for help 😄

jhlegarreta:
> If I remember correctly, he was looking for help 😄

@dzenanz Brad made a few comments early on and following them I was able to propose a solution that was successful.

> May I take a stab at addressing the issue, as I don't believe the base class refactor is nessacary to address?

@blowekamp I thought that after having addressed the comments the solution I came up with was a pretty neat one, and had tested with the failing tests in #511, but definitely, if you think you have a more elegant or efficient one, please go ahead.

blowekamp:
Ok, hopefully it works out elegantly... I also have interest in making these filters run "inplace" and the extra base class would make it more difficult.

jhlegarreta:
Superseded by #1730. Closing.

PR #1730, opened 2020-03-26
BUG: Use one calculator in ItermodesThresholder

### LINKED ISSUE #1720: Update modified time in `itk::IntermodesThresholdImageFilter::SetMaximumSmoothingIterations` and other set methods.
### Description

When calling the `Set...` methods in the  `itk::IntermodesThresholdImageFilter` class it should be update the classes MTime if the value changes. This will ensure the pipeline reexecutes after a value has changed.

### Steps to Reproduce
None.

### Expected behavior
Reproducing @blowekamp 's proposal in [this comment](https://github.com/InsightSoftwareConsortium/ITK/pull/1718/files#r396414648):

> The simple way to do this would be to copy the code in the [set/get macro](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkMacro.h#L908-L912) which checks if the value is the same, if they are different then set the modify time. The more complicated way would involve modify the parent's GetMTime method to return the max of the MTime of itself, and the m_Calculator.

### Actual behavior
No check about the modified time is made.

### Reproducibility
%100

### Versions
`master`

### Environment
Any environment.

### Additional Information
None.

blowekamp:
Use the parent class's calculator and override GenerateDate method to
update the derived calculators ivars. Store the calculator parameters
as filter ivar, with correct filter modified time management.

Closes #1720.

## PR Checklist

### DIFF
--- a/Modules/Filtering/Thresholding/include/itkIntermodesThresholdImageFilter.h
+++ b/Modules/Filtering/Thresholding/include/itkIntermodesThresholdImageFilter.h
@@ -93,31 +93,15 @@ public:
   using HistogramType = typename Superclass::HistogramType;
   using CalculatorType = IntermodesThresholdCalculator<HistogramType, InputPixelType>;
 
-  void
-  SetMaximumSmoothingIterations(SizeValueType maxSmoothingIterations)
-  {
-    m_IntermodesCalculator->SetMaximumSmoothingIterations(maxSmoothingIterations);
-  }
 
-  SizeValueType
-  GetMaximumSmoothingIterations()
-  {
-    return (m_IntermodesCalculator->GetMaximumSmoothingIterations());
-  }
+  itkSetMacro(MaxSmoothingIterations, SizeValueType);
+  itkGetMacro(MaxSmoothingIterations, SizeValueType);
 
   /** Select whether midpoint (intermode=true) or minimum between
      peaks is used. */
-  void
-  SetUseInterMode(bool useIntermode)
-  {
-    m_IntermodesCalculator->SetUseInterMode(useIntermode);
-  }
-
-  bool
-  GetUseInterMode()
-  {
-    return (m_IntermodesCalculator->GetUseInterMode());
-  }
+  itkSetMacro(UseInterMode, bool);
+  itkGetConstReferenceMacro(UseInterMode, bool);
+  itkBooleanMacro(UseInterMode);
 
   /** Image related type alias. */
   static constexpr unsigned int InputImageDimension = InputImageType::ImageDimension;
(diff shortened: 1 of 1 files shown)


blowekamp:
After reviewing the code, based on my understanding of the issue, and a couple iterations this is what I came up with to address a couple issue and keep the implementations perhaps a little easier to maintain.

blowekamp:
The Printing of the new ivars is not done. The PrintSelf method should be overloaded.

Modules/Filtering/Thresholding/include/itkIntermodesThresholdImageFilter.h
```diff
@@ -93,31 +93,15 @@ class IntermodesThresholdImageFilter : public HistogramThresholdImageFilter<TInp
...
-  {
-    m_IntermodesCalculator->SetMaximumSmoothingIterations(maxSmoothingIterations);
-  }
 
-  SizeValueType
-  GetMaximumSmoothingIterations()
-  {
-    return (m_IntermodesCalculator->GetMaximumSmoothingIterations());
-  }
+  itkSetMacro(UseInterMode, SizeValueType);
```
dzenanz: Should this be `MaximumSmoothingIterations`? `UseInterMode` is declared few lines later.

blowekamp: Good catch!

jhlegarreta:
Could this PR be also used to take [this](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/Thresholding/include/itkHistogramThresholdImageFilter.hxx#L84) and [this](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/Thresholding/include/itkHistogramThresholdImageFilter.hxx#L97) statements outside the `if/else` block and thus use a single statement?

blowekamp:
> Could this PR be also used to take [this](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/Thresholding/include/itkHistogramThresholdImageFilter.hxx#L84) and [this](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/Thresholding/include/itkHistogramThresholdImageFilter.hxx#L97) statements outside the `if/else` block and thus use a single statement?
> 
> Sorry Brad. Just realized that I had noticed this in my PR. Should have realized earlier.

I missed that difference. This PR is ready to be merged now, so can you create separate PR with that change? It's a little unrelated to this one.

jhlegarreta:
> I missed that difference. This PR is ready to be merged now, so can you create separate PR with that change? It's a little unrelated to this one.

Sounds fair. Thanks Brad !!!

PR #1735, opened 2020-03-30
Add lambda function command

blowekamp:
## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkObject.h
+++ b/Modules/Core/Common/include/itkObject.h
@@ -145,14 +145,14 @@ public:
     SetGlobalWarningDisplay(false);
   }
 
-  /** Allow people to add/remove/invoke observers (callbacks) to any ITK
-   * object. This is an implementation of the subject/observer design
+  /** \brief Add an observer/command to this object invoked for event
+   *
+   * This is an implementation of the subject/observer design
    * pattern. An observer is added by specifying an event to respond to
    * and an itk::Command to execute. It returns an unsigned long tag
    * which can be used later to remove the event or retrieve the
-   * command.  The memory for the Command becomes the responsibility of
-   * this object, so don't pass the same instance of a command to two
-   * different objects  */
+   * command.
+   */
   unsigned long
   AddObserver(const EventObject & event, Command *);
   unsigned long
--- a/Modules/Core/Common/include/itkObject.h
+++ b/Modules/Core/Common/include/itkObject.h
@@ -33,6 +33,8 @@
 #include "itkMetaDataDictionary.h"
 #include "itkSingletonMacro.h"
 
+#include <functional>
+
 namespace itk
 {
 // Forward reference because of private implementation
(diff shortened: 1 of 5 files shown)


Modules/Core/Common/src/itkObject.cxx
```diff
@@ -503,6 +503,15 @@ Object ::AddObserver(const EventObject & event, Command * cmd) const
   return this->m_SubjectImplementation->AddObserver(event, cmd);
 }
 
+unsigned long
+Object::AddObserver(const EventObject & event, std::function<void(const EventObject &)> function) const
+{
+  auto cmd = FunctionCommand::New();
```
dzenanz: Will `cmd` be deleted once object is deallocated?

dzenanz:
We should also update documentation and examples, as this greatly reduces amount of code needed to track events.

Modules/Core/Common/include/itkObject.h
```diff
@@ -156,6 +156,8 @@ class ITKCommon_EXPORT Object : public LightObject
 
   unsigned long
   AddObserver(const EventObject & event, Command *) const;
+  unsigned long
+  AddObserver(const EventObject & event, std::function<void(const EventObject &)> function) const;
```
phcerdan: Do we need the non-const version?

Modules/Core/Common/src/itkObject.cxx
blowekamp: Yes the itk::Object holds onto the Commands with SmartPointers as is normal.

blowekamp:
The original callbacks had this signature:
```
void f(const Object * caller, const EventObject & event)
```

Here I removed the object caller parameter to avoid the odd two signatures of the callback and invocation. 

I am considering also removing the event argument too, as that can easily be replaced with lambda captures. Here is usage of the shortened usage:
```
o->AddObserver(itk::AnyEvent(), [&cnt]() { ++cnt; });
```

Any thought if either of those arguments are needed?

Modules/Core/Common/include/itkObject.h
blowekamp: That was part of the reason, I removed the object argument to the callback, I was also considering removing the EventObject argument too.

dzenanz:
Listening to iteration events is probably the most common. I would keep the arguments. The compiler will remove them if they are not needed, and they provide generality. If you really want the (short-form) non-argument lambda to be available, you could provide both signatures.

blowekamp:
> Listening to iteration events is probably the most common. I would keep the arguments. The compiler will remove them if they are not needed, and they provide generality. If you really want the (short-form) non-argument lambda to be available, you could provide both signatures.

So for iteration events why would you need the arguments? The EventObjects contain no additional information beyond the type of event, which is known when you add the command. And the object can easily be captured.

With both arguments the inline addition of observer would look like:
```
 o->AddObserver(itk::AnyEvent(), [](Object * caller, const EventObject & event) {std::cout << event.GetEventName() << std::end; });
```

or 

```
auto cmd = [](Object * caller, const EventObject & event)
{
   std::cout << event.GetEventName() << std::end;
 };
o->AddObserver(itk::AnyEvent(), std::move(cmd));
```

I am no doubt biased by what is implemented in SimpleITK:
https://github.com/SimpleITK/SimpleITK/blob/f8d1ff4328dd2f31d132b01e72d86667e98d3a6d/Examples/FilterProgressReporting/FilterProgressReporting.cxx

dzenanz:
I thought you were referring to arguments of `AddObserver` method. The current proposal is a good balance, in case someone wants to handle different event types differently.

thewtex:
@blowekamp thanks for working on this -- looks very cool!

Please refrain from merging until after the 5.1.0 tag too allow for time to iron out build issues and the interface for this important, new feature.

dzenanz:
Can you fix compile issues and rebase? Now is a good time to merge this.

PR #1745, opened 2020-04-01
ENH: Improve is_vector identification in image_from_xarray

thewtex:
A simpler check that requires a 'c' dim explicitly.


thewtex:
`c` is channel / component / color. More documentation on this is in the works. :-)

PR #1748, opened 2020-04-03
ENH: Update CI testing virtual machine images

thewtex:


Testing/ContinuousIntegration/AzurePipelinesWindowsPython.yml
```diff
@@ -66,7 +66,7 @@ jobs:
 
     - script: |
         cmake --version
-        call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+        call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
```
blowekamp, line 69: Don't we need specific VS tool set for Python binaries? Should we keep the CI the same as the packaging? I am setting the following in my SimpleITK builds (but I'm not sure how it would work here):
```
CTEST_CMAKE_GENERATOR_TOOLSET: v141,host=x64
```

dzenanz: Ninja picks up CL version from the path, and this sets the x64 architecture I belive.

thewtex: It is a good idea to use an explicit version version of the binaries. We can specify this with arguments to vcvars64.bat. I will add that...

Modules/Core/Common/test/itkHashTableTest.cxx
```diff
@@ -79,19 +79,12 @@ itkHashTableTest(int, char *[])
   lookup(Set, "apple");
   lookup(Set, "durian");
 
-  Set.begin();
```
dzenanz, line 82: We could remove this entire file, since we have removed `itk::HashTable` in favor of `std::unordered_map`.

Modules/Video/IO/include/itkVideoFileWriter.hxx
```diff
@@ -253,7 +253,7 @@ VideoFileWriter<TInputVideoStream>::InitializeOutputParameters()
   const VideoStreamType * input = this->GetInput();
 
   // Set dimensions
-  m_Dimensions.empty();
+  m_Dimensions.clear();
```
dzenanz, line 256: > .clear() only returns whether the container is empty or not.

You mean "`.empty()` only returns whether the container is empty or not."

Modules/Core/Common/test/itkHashTableTest.cxx
thewtex: Yes, but it does not hurt to keep it around.

Modules/Video/IO/include/itkVideoFileWriter.hxx
thewtex: Thanks, fixed.

PR #1755, opened 2020-04-07
ENH: Added jacobian computation to ComposeScaleSkewVersorTransform

aylward:
Computed Jacobian using SymPy.  See comments in code for computation
script.

Added test to verify Jacobian with finite difference results.

## PR Checklist

### DIFF
--- a/Modules/Core/Transform/include/itkComposeScaleSkewVersor3DTransform.hxx
+++ b/Modules/Core/Transform/include/itkComposeScaleSkewVersor3DTransform.hxx
@@ -310,19 +310,24 @@ ComposeScaleSkewVersor3DTransform<TParametersValueType>::ComputeJacobianWithResp
   JacobianType &         jacobian) const
 {
   // Jacobian computed via
-  // pip install pysym
-  // from sympy import Function, Derivative, simplify, Transpose
-  // from sympy import symbols from sympy.matrices import Matrix, eye, zeros, ones, diag
-  // x0, x1, x2, v0, v1, v2, w, o0, o1, o2, s0, s1, s2, k0, k1, k2 = symbols('x0 x1 x2 v0 v1 v2 w o0 o1 o2 s0 s1 s2 k0
-  // k1 k2') M = Matrix([[1-2*(v1*v1+v2*v2), 2*(v0*v1-v2*w), 2*(v0*v2+v1*w), o0],
+  // pip install sympy
+  //
+  // from sympy import Derivative, simplify
+  // from sympy import symbols
+  // from sympy.matrices import Matrix
+  // x0, x1, x2, v0, v1, v2, w, o0, o1, o2, s0, s1, s2, k0, k1, k2 =
+  //        symbols('x0 x1 x2 v0 v1 v2 w o0 o1 o2 s0 s1 s2 k0 k1 k2')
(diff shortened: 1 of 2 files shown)


Modules/Core/Transform/include/itkComposeScaleSkewVersor3DTransform.hxx
```diff
@@ -306,71 +306,113 @@ ComposeScaleSkewVersor3DTransform<TParametersValueType>::PrintSelf(std::ostream
...
 {
-  // This document how the jacobian can be computed for certain parameters;
-  //   however, the proper computation of the scale and skew parameters is
-  //   now given or computed, so ultimately an exception is thrown.
+  // Jacobian computed via
+  // pip install pysym
+  // from sympy import Function, Derivative, simplify, Transpose
+  // from sympy import symbols from sympy.matrices import Matrix, eye, zeros, ones, diag
+  // x0, x1, x2, v0, v1, v2, w, o0, o1, o2, s0, s1, s2, k0, k1, k2 = symbols('x0 x1 x2 v0 v1 v2 w o0 o1 o2 s0 s1 s2 k0
+  // k1 k2') M = Matrix([[1-2*(v1*v1+v2*v2), 2*(v0*v1-v2*w), 2*(v0*v2+v1*w), o0],
```
dzenanz, line 317: line break here?

dzenanz: Why is matrix `M` defined this way?

blowekamp: It should be the quaternion to rotation matrix conversion: https://www.euclideanspace.com/maths/geometry/rotations/conversions/quaternionToMatrix/index.htm

dzenanz: Adding that link here as a comment would be quite helpful.

blowekamp:
Very nice!

I recall doing something similar and thinking is was "too" complicated, but this looks simpler. 

It will be interesting to see how it compare for registration with the additive composite transform.

Modules/Core/Transform/include/itkComposeScaleSkewVersor3DTransform.hxx
blowekamp: I think stating "quaternion as matrix" someplace would  be sufficcient documentation. I'm not sure if this is the correct place for the python code though.

dzenanz: If not here, this code could go into the commit message.

aylward: The nice part is that this code generates the jacobian in symbolic form, so I'd like to preserve it.  We can also use an adaption of this code to compute jacobians for any of our other transforms that need it.    For that reason, perhaps it is better if the code stays with the jacobian it generated, rather than moving it to a commit message?

aylward: Will add the link as requested.

dzenanz:
Preferably squash upon merge.

thewtex:
:sunglasses:

PR #1758, opened 2020-04-09
DOC: Release process updates for v5.1rc03

thewtex:
Visual Studio 2019 now comes with CMake.

### DIFF
--- a/Documentation/Maintenance/Release.md
+++ b/Documentation/Maintenance/Release.md
@@ -210,7 +210,7 @@ Run the update script:
 Commit the result:
 
 ```sh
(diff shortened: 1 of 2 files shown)


thewtex:
@hjmjohnson **incredible** work on all the remote modules since v5.1rc02!

hjmjohnson:
> @hjmjohnson **incredible** work on all the remote modules since v5.1rc02!

@mseng10 is working to make the remote modules build and test cleanly.  I am hopeful that we can move many of these to a higher compliance level, and use them as better testbeds for future changes.

PR #1761, opened 2020-04-10
ENH: gitattributes: verify that code adheres to flake8

mathstuf:
@thewtex @bradking 

However, local testing shows that the codebase is not flake8-clean. This should be deferred until that is the case (the robot only looks at changed files for `flake8` enforcement, so this PR is clean).

## PR Checklist

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -27,14 +27,16 @@ vxl_doxy.pl      crlf=input
 zap.pl           crlf=input
 
 
-# Custom attribute to mark sources as using our C++/C code style.
+# Custom attribute to mark sources as using our C++/C/Python code style.
 [attr]our-c-style  whitespace=tab-in-indent,no-lf-at-eof  hooks.style=KWStyle,clangformat
+[attr]our-py-style whitespace=tab-in-indent,no-lf-at-eof  format.flake8
 
 *.c              our-c-style
 *.h              our-c-style
 *.cxx            our-c-style
 *.hxx            our-c-style
 *.txx            our-c-style
+*.py             our-py-style
 *.txt            whitespace=tab-in-indent,no-lf-at-eof
 *.cmake          whitespace=tab-in-indent,no-lf-at-eof
 


.gitattributes
```diff
@@ -29,12 +29,14 @@ zap.pl           crlf=input
 
 # Custom attribute to mark sources as using our C++/C code style.
 [attr]our-c-style  whitespace=tab-in-indent,no-lf-at-eof  hooks.style=KWStyle,clangformat
+[attr]our-py-style whitespace=tab-in-indent,no-lf-at-eof  format.flake8
```
jcfr, line 32: Comment above should be changed from 

```
Custom attribute to mark sources as using our C++/C code style.
```

to 

```
Custom attributes to mark sources as using our C++/C/Python/... code styles.
```

jcfr: Also, I suggest:
* `our-py-style` -> `itk-py-style`
* `our-c-style` -> `itk-c-cpp-style`

mathstuf: The attribute rename can be done separately; it can't conflict with anything anyways.

hjmjohnson:
I would prefer to ensure that all python code is compatible with 'black' autoformatting tool.

some default flake8 features would need to be modified.

Flake8
[Flake8](https://pypi.org/p/flake8/) is a code linter. It warns you of syntax errors, possible bugs, stylistic errors, etc. For the most part, Flake8 follows [PEP 8](https://www.python.org/dev/peps/pep-0008/) when warning about stylistic errors. There are a few deviations that cause incompatibilities with Black.

Configuration
max-line-length = 88
extend-ignore = E203
Why those options above?
In some cases, as determined by PEP 8, Black will enforce an equal amount of whitespace around slice operators. Due to this, Flake8 will raise E203 whitespace before ':' warnings. Since this warning is not PEP 8 compliant, Flake8 should be configured to ignore it via extend-ignore = E203.

When breaking a line, Black will break it before a binary operator. This is compliant with PEP 8 as of [April 2016](https://github.com/python/peps/commit/c59c4376ad233a62ca4b3a6060c81368bd21e85b#diff-64ec08cc46db7540f18f2af46037f599). There’s a disabled-by-default warning in Flake8 which goes against this PEP 8 recommendation called W503 line break before binary operator. It should not be enabled in your configuration.

Also, as like with isort, flake8 should be configured to allow lines up to the length limit of 88, Black’s default. This explains max-line-length = 88.

Formats
.flake8
setup.cfg
tox.iniFlake8
[Flake8](https://pypi.org/p/flake8/) is a code linter. It warns you of syntax errors, possible bugs, stylistic errors, etc. For the most part, Flake8 follows [PEP 8](https://www.python.org/dev/peps/pep-0008/) when warning about stylistic errors. There are a few deviations that cause incompatibilities with Black.

Configuration
max-line-length = 88
extend-ignore = E203
Why those options above?
In some cases, as determined by PEP 8, Black will enforce an equal amount of whitespace around slice operators. Due to this, Flake8 will raise E203 whitespace before ':' warnings. Since this warning is not PEP 8 compliant, Flake8 should be configured to ignore it via extend-ignore = E203.

When breaking a line, Black will break it before a binary operator. This is compliant with PEP 8 as of [April 2016](https://github.com/python/peps/commit/c59c4376ad233a62ca4b3a6060c81368bd21e85b#diff-64ec08cc46db7540f18f2af46037f599). There’s a disabled-by-default warning in Flake8 which goes against this PEP 8 recommendation called W503 line break before binary operator. It should not be enabled in your configuration.

Also, as like with isort, flake8 should be configured to allow lines up to the length limit of 88, Black’s default. This explains max-line-length = 88.

Formats
.flake8
setup.cfg
tox.ini

hjmjohnson:
Pre-commit hooks defined in .pre-commit-config.yaml are now used to ensure consistent formatting.

This approach is succeeded by the use of `- repo: https://github.com/psf/black` to achieve similar results.

mathstuf:
FWIW, `black` is now supported in `source-formatters`, so the robot can enforce it too.

PR #1773, opened 2020-04-20
COMP: Condition HDF5 HDopen on _MSC_VER definition

thewtex:
H5_HAVE_VISUAL_STUDIO may be defined if Visual Studio is present on the
build system even if MinGW-w64 is the currently used compiler.


thewtex:
Builds / tests are now passing: https://open.cdash.org/buildSummary.php?buildid=6482015

PR #1786, opened 2020-04-26
BUG: CurvatureRegistrationFilter fix logic re: depends on FFTW

aylward:
Conditional in .h file did not implement intended logic (resulting in compilation error when used, e.g., with Python wrapping enabled) and logic was not duplicated in .hxx file (resulting in .hxx definitions without a corresponding .h).


dzenanz:
Tricky bug!

PR #1791, opened 2020-04-28
BUG: Double scaling introduced in refactoring

hjmjohnson:
The moving marginal pdf was double scaled by the normalization factor.

During the previous refactorings, the scaling was decoupled from
the computations for the fixed and marginal computations, but
was not removed from the subsequent marginal computation.

Print trace dumps of every variable value before and after
April 10, 2020 changes.

## PR Checklist


hjmjohnson:
NOTE: https://github.com/InsightSoftwareConsortium/ITK/pull/1791/commits/b699519c062afb9b248309245528c49ca66d94cb#diff-e9f98ebc255c5010362c27c3b3945d05L347 

The line removed adds a second scaling to the transform done at the above line.

blowekamp:
Is this related to issue discovered in SimpleITK/SimpleITK#1084.

Looks cool. Is there some testing that should be added to ITK?

hjmjohnson:
> Is this related to issue discovered in [SimpleITK/SimpleITK#1084](https://github.com/SimpleITK/SimpleITK/pull/1084).
> 
> Looks cool. Is there some testing that should be added to ITK?

Yes.  I would love to have a test, but I won't be able to work on that until later this week.

blowekamp:
Outstanding! I have confirmed that this fixes the test failures in SimpleITK with ITK 5.1.

This appears  to be a significant issue with the registration metric. The metric value significantly changed  from this bug e.g -1 vs 10. It is interesting that other registration test suites did not detect this change. SimpleITK, perhaps uniquely, compares RMS of resulting transforms as a displacement fields and not the resulting transformed image.

Hopefully, a patched 5.1.1 can be released quickly to address this major issue with registration.

hjmjohnson:
> This appears to be a significant issue with the registration metric.

@blowekamp I build BRAINSTools and ANTs with all their testing and regression images as well, and neither of those packages reported new regressions either (that is to say the new results were within tolerances of numerical stability margins needed for multithreaded orders of operations accumulation changes/platform changes).    I'm glad we got this tracked down and fixed.

thewtex:
Merged to `release`

PR #1793, opened 2020-04-29
ENH: Add test to monitor metric result value

hjmjohnson:
Add a simple regression test to ensure that the metric returned
values are consistent over time.

When run before recently merged "double normalization bug" fix the following results were obtained.

```bash
2018: 	[FAILED] known values for metric values not computed correctly.-10.50833512288329 != -1.297994750907097
2018: 	[FAILED] known values for metric values not computed correctly.-10.50833574913326 != -1.297995377157083
2018: 	[FAILED] known values for metric values not computed correctly.-10.50833482019903 != -1.297994448222851
2018: 	[FAILED] known values for metric values not computed correctly.-10.508336072699 != -1.297995700722833
```


hjmjohnson:
NOTE:  I guessed a tolerance level for the metric of 1e-7.  We may need to adjust if that is too strict on other platforms.

blowekamp:
I have been using 1e-9 in SimpleITK to compare a static evaluation, relaxing to 1e-7 seems reasonable for a few iterations of the metric during optimization.

hjmjohnson:
@blowekamp I don't understand your comment. Do you approve of this PR?  Note the test fails for different computer environments with more stringent tolerances.

PR #1795, opened 2020-04-29
ENH: Bump TubeTK to version that removes ArrayFire (replaced by cufft)

aylward:
Also enables building TubeTK's applications.

### DIFF
--- a/Modules/Remote/TubeTK.remote.cmake
+++ b/Modules/Remote/TubeTK.remote.cmake
@@ -11,11 +11,11 @@
 #--
 #-- ## Compliance Level 4 star (Very high-quality code, perhaps small community dependance)
 #--   - [ ] Meets all ITK code style standards
-#--   - [ ] No external requirements beyond those needed by ITK proper
-#--   - [ ] Builds and passes tests on all supported platforms within 1 month of each core tagged release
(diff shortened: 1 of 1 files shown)


Modules/Remote/TubeTK.remote.cmake
```diff
@@ -11,11 +11,11 @@
 #--
 #-- ## Compliance Level 4 star (Very high-quality code, perhaps small community dependance)
 #--   - [ ] Meets all ITK code style standards
-#--   - [ ] No external requirements beyond those needed by ITK proper
-#--   - [ ] Builds and passes tests on all supported platforms within 1 month of each core tagged release
-#--            - [ ] Windows Shared Library Build with Visual Studio
-#--            - [ ] Mac with clang compiller
-#--            - [ ] Linux with gcc compiler
+#--   - [X] No external requirements beyond those needed by ITK proper
```
dzenanz, line 14: I assume this means cuFFT is optional?

aylward: Yes - but it is already an option for ITK (I removed ArrayFire since it wasn't already an option for ITK).

dzenanz: Sounds good.

PR #1796, opened 2020-04-30
ENH: updating remote modules to latest versions

dzenanz:


dzenanz:
Can anyone review/merge? It is ready to go.

phcerdan:
I am missing extra information about the changes in the remove modules.
Can we provide a shortlog of the changes between SHA's?
For example for BioCells:
```
git shortlog --no-merges 47a233..1d394ab | cat                                                                15:08:52 
Mathew Seng (1):
      BUG: Fix failing tests for examples Replaced ITKIONRRD with ITKIOPNG and ITKIOMeta, because the examples were trying to build the png and mha images as nrrd, thus throwing an error
```

dzenanz:
We would need to update the [update script](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/Maintenance/UpdateRemoteModules.sh) to include short logs, otherwise updating will become a chore.

PR #1797, opened 2020-04-30
BUG: Disable dynamic threading in noise filter.

### LINKED ISSUE #1794: ImageNoise filters not deterministic with TBB Threading
### Description
 
The ImageNoise filter were designed to be deterministic, with a give seed and a fixed number of threads. The TBB threader uses "dynamic" region splitting making the output no-deterministic.

### Steps to Reproduce

This regression was detected by SimpleITK testing suite. It can be easily reproduced when SimpleITK is compiled with TBB by running:

img = sitk.Image([512,512], sitk.sitkUInt8)   
print(sitk::Hash(sitk.AdditiveGaussianNoise(img, seed=10)))

Where the last line produces different results when re-run.

### Expected behavior

The results should be reproducable.

### Actual behavior

The results are usually different, but dependent on the TBB region division.

### Reproducibility

The more threads available increases the changes, the results will differ.

### Versions

5.0+

blowekamp:
With the TBB threader the results were dependent on how the regions
were dynamically split.

The noise filters are designed to be reproducible for a fixed seed and
a number of threads. The implementation requires that the image is
split into fix deterministic regions.

Closes #1794.

### DIFF
--- a/Modules/Filtering/ImageNoise/include/itkAdditiveGaussianNoiseImageFilter.h
+++ b/Modules/Filtering/ImageNoise/include/itkAdditiveGaussianNoiseImageFilter.h
@@ -106,7 +106,7 @@ protected:
   PrintSelf(std::ostream & os, Indent indent) const override;
 
   void
-  DynamicThreadedGenerateData(const OutputImageRegionType & outputRegionForThread) override;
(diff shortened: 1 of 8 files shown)


dzenanz:
Can you rebase to get a CI re-test?

PR #1801, opened 2020-05-06
BUG: Use ITK_WRAP_IMAGE_DIMS in ViewImage.wrap

phcerdan:
Instead of hard-coded `2;3` dimensions


phcerdan:
It only affects when user change ITK_WRAP_IMAGE_DIMS, which is `2;3` by default.

Issue #91, opened 2018-11-05
Document how to start Python CI builds


thewtex:
Add a note to `CONTRIBUTING.md` that a pull request against the `python-builds` branch can be used to start Python builds.

Add content to the Git client hooks to present the URL required to open the pull request when pushed to GitHub.

thewtex:
Depends on #166

thewtex:
We are currently running Python builds on each PR, which is working well.

PR #1823, opened 2020-05-13
ENH: Bump MetaIO to address style and const params

aylward:
Update to MetaIO tag f473b7b86a15dc2f3578f1e82b5544ef7776e2e7
from May 11, 2020.

ENH: Parse() should handle char *argv[] as const (#91)


aylward:
That script did not work for me.  It initiated many changes throughout itk
and produced conflicts in metaio dir.

It did work for me in the past.

If it works for you, I am happy to abandon this.

---
Sent via phone.

dzenanz:
I also run into numerous merge conflicts when trying to run the script now.

PR #1825, opened 2020-05-14
Examples column limit at 78

dzenanz:
Attempt to address https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/issues/131

PR #1827, opened 2020-05-14
Reduce examples column limit to 78

dzenanz:
Attempt to address InsightSoftwareConsortium/ITKSoftwareGuide#131.


dzenanz:
This is a revival of #1825 targeting the release branch.

PR #1857, opened 2020-06-10
BUG: DECL_PYTHON_VARLEN_SEQ_TYPEMAP typenames only works with Python …

thewtex:
…sequence

Addresses issue #871.

an error message.

### DIFF
--- a/Modules/Segmentation/LevelSets/wrapping/test/CMakeLists.txt
+++ b/Modules/Segmentation/LevelSets/wrapping/test/CMakeLists.txt
@@ -86,4 +86,9 @@ if(ITK_WRAP_PYTHON)
       EXPRESSION "filt = itk.BinaryMaskToNarrowBandPointSetFilter.New()"
     )
   endif()
+  itk_python_add_test(
+    NAME PythonShapePriorMAPCostFunctionTest
+    COMMAND
+      ShapePriorMAPCostFunctionTest.py
+  )
 endif()
--- a/Modules/Segmentation/LevelSets/wrapping/test/CMakeLists.txt
+++ b/Modules/Segmentation/LevelSets/wrapping/test/CMakeLists.txt
@@ -91,4 +91,9 @@ if(ITK_WRAP_PYTHON)
     COMMAND
       ShapePriorMAPCostFunctionTest.py
   )
+  itk_python_add_test(
+    NAME PythonShapePriorMAPCostFunctionInitializeTest
+    COMMAND
+      ShapePriorMAPCostFunctionInitializeTest.py
+  )
 endif()
--- a/Modules/Segmentation/LevelSets/wrapping/test/ShapePriorMAPCostFunctionInitializeTest.py
+++ b/Modules/Segmentation/LevelSets/wrapping/test/ShapePriorMAPCostFunctionInitializeTest.py
@@ -1,0 +1,213 @@
+# ==========================================================================
+#
+#   Copyright NumFOCUS
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#          https://www.apache.org/licenses/LICENSE-2.0.txt
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ==========================================================================
+
+"""End-to-end regression test for the issue #871 failure path.
+
+Issue #871 reporter's observed error:
+
+    RuntimeError: itkShapePriorMAPCostFunction.hxx:187:
+    itk::ERROR: ShapePriorMAPCostFunction: ShapeParameterMeans does not
+    have at least 3 number of elements.
+
+That raise site is inside `ShapePriorMAPCostFunction::Initialize()`,
+which the reporter reached via
+`GeodesicActiveContourShapePriorLevelSetImageFilter.Update()`.  The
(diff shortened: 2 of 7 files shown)


dzenanz (approved):
Looks good on a glance.

hjmjohnson:
Python build failures expired.  Update required to understand what is going wrong.

thewtex:
@hjmjohnson thanks!

I am not sure if the implementation was having the intended effect.

dzenanz:
I don't have time right now to dive into this.

hjmjohnson:
Rebased onto current `upstream/main` and split into three terse commits (BUG + STYLE + ENH); dedicated regression test lands in the ENH commit.  Full walkthrough below for reviewers / future context.

<details>
<summary>Root cause of #871</summary>

`itkSetMacro(name, type)` expands to `virtual void Set##name(type _arg)` — **by value**.  The SWIG by-value typemap emitted by `DECL_PYTHON_VARLEN_SEQ_TYPEMAP` had no branch for the case where `SWIG_ConvertPtr` succeeded.  `$1` (a stack-allocated `type`) was left default-constructed, so the C++ method received an empty container whenever a native wrapped instance (`itk.Array[itk.D](n)`) was passed.

Both typemaps additionally fell through to `PyObject_Length($input)` whenever `SWIG_ConvertPtr` returned -1.  For a non-sequence argument `PyObject_Length` returns -1; the subsequent `type(size_t(-1))` allocation then exploded instead of raising a well-formed `TypeError`.

</details>

<details>
<summary>Fix (BUG commit f5bf182aae)</summary>

- **By-value typemap**: add `else if (s != NULL) { $1 = *s; }` so a wrapped native instance is copied into the by-value slot.
- **Both typemaps**: gate the sequence path with `PySequence_Check($input)` and raise a descriptive `TypeError` otherwise.
- Every error exit in the fixed typemaps now uses `SWIG_fail` so the generated cleanup label runs and `PyObject_GetItem` references acquired earlier in the loop are released on error.
- Companion test `ShapePriorMAPCostFunctionTest.py` uses non-zero sentinel fills (`3.14`, `2.71`) so a regression that preserves size but corrupts content also fails.

</details>

<details>
<summary>STYLE signature extension (commit 0adb0611a3)</summary>

Add a `swig_name` parameter matching the sibling `DECL_PYTHON_VEC_TYPEMAP(swig_name, type, dim)` convention:

```
-%define DECL_PYTHON_VARLEN_SEQ_TYPEMAP(type, value_type)
+%define DECL_PYTHON_VARLEN_SEQ_TYPEMAP(swig_name, type, value_type)
```

SWIG `%define` substitutes parameter tokens inside string literals, so the generated `__repr__` renders as `"itkArrayD (...)"` instead of `"swig_name (...)"`, and the TypeError message carries the real Python class name.  No behavior change for the three current callers (`itk::Array`, `itk::RGBPixel`, `itk::RGBAPixel`) — `swig_name` and the C++ type tag coincide for all of them.

</details>

<details>
<summary>Empirical verification (local, ITK built with pixi + SWIG 4.3.0)</summary>

\`\`\`
\$ pixi run -e python ctest --test-dir build-python -R PythonVarLenSeqTypemapTest -V
...
PASS  Python list round-trips through by-value typemap
PASS  Native itk.Array[itk.D] round-trips bit-exactly (issue #871)
PASS  Size of round-tripped native itk.Array matches input
PASS  TypeError message names the expected wrapped type (swig_name substitution works): Expecting an itkArrayD, or a sequence of values.
PASS  ValueError message describes expected element type: Expecting a sequence of int or float
PASS  itk.Array[itk.D] __repr__ contains the Python class name: itkArrayD ([0, 0, 0])
PASS  itk.Array[itk.D] __repr__ does not leak the unsubstituted macro-parameter token 'swig_name': itkArrayD ([0, 0, 0])
PASS  itk.Array[itk.F] __repr__ substitutes correctly: itkArrayF ([1.125, 1.125])

All DECL_PYTHON_VARLEN_SEQ_TYPEMAP regression checks passed.
\`\`\`

`ShapePriorMAPCostFunctionTest.py` (in the BUG commit) also passes with the native-array sentinel round-trip:

\`\`\`
Baseline mean: itkArrayD ([3.14, 3.14])
mean: itkArrayD ([3.14, 3.14])
Baseline stddev: itkArrayD ([2.71, 2.71])
stddev: itkArrayD ([2.71, 2.71])
\`\`\`

</details>

<details>
<summary>Commit series</summary>

| commit | subject |
|---|---|
| f5bf182aae | BUG: DECL_PYTHON_VARLEN_SEQ_TYPEMAP drops native ITK array input |
| 0adb0611a3 | STYLE: Extend DECL_PYTHON_VARLEN_SEQ_TYPEMAP signature |
| f264818a3f | ENH: Add DECL_PYTHON_VARLEN_SEQ_TYPEMAP regression test |

Each commit is bisect-safe — the test added in each commit passes against that commit's own state (the STYLE-dependent `__repr__` assertion is only in the ENH test, which lands after STYLE).

</details>

dzenanz (commented):
Description of what happens sounds good. A quick look at the actual fix raised on red flags for me. I didn't really look at the tests. LTGM, but someone else should review.

hjmjohnson:
@dzenanz I would not want to move forward if you have red flag concerns.  I canceled your previous approval of this PR because it didn't fix issue #871, and the CI build failed.  

A new set of tests, and passing the scenario of #871 is now in place.  If we can address your red-flag concerns, we can resolve this old PR and issue.

@thewtex This was originally a PR started by you. Could you see if you have concerns or suggestions for financing this effort?

Thanks.

dzenanz:
> raised on red flags

Mistyped. It should have been: raised **no** red flags

thewtex (commented):
@hjmjohnson thank you for finally addressing this!! It was a complex set of type mapping and lifetime/ownership mapping that we did not get a change to nail down. LGTM!

PR #1868, opened 2020-06-10
ENH: Prefer to use SCANNER_ANAT sform over qform when both are set

hjmjohnson:
When the sform format and qform format are both available,
or if sform representation is a orthonormal matrix and qform
is not set, prefer sform as the representation for the ITK
direction cosine when set as NIFTI_XFORM_SCANNER_ANAT.

This gives a slightly higher precision representation
that is closer to the DICOM representation of the
direction cosines than the qform can represent.  The
more accurate representation is imporant for applications
that must match the original DICOM data physical space
representation.

This also is the approach of other popular tools
for dicom format conversions like the widely used
dcm2niix program.

## PR Checklist
- [X] Added test (or behavior not changed) Modified existing test.


dzenanz:
@hjmjohnson should we wait more for possible reviews, or merge?

gdevenyi:
This PR title scared me until I read it, can it mention NIFTI_XFORM_SCANNER_ANAT

dzenanz:
I think titles are limited to 72 characters or so 😄

cookpa:
Hi  @hjmjohnson, thanks for working on this. I'm probably missing something here, but I could not figure out how this function

https://github.com/InsightSoftwareConsortium/ITK/commit/d52dfdc44edfd2d6048a712a88752ec3f5a73a4d#diff-614c78fa46f6db4dd982f8504c665901R1799-R1804

would work if the sform matrix sto_xyz is not a pure rotation. For example, if the voxels were anisotropic, wouldn't sto_xyz include a scaling?

hjmjohnson:
> Hi @hjmjohnson, thanks for working on this. I'm probably missing something here, but I could not figure out how this function
> 
> [d52dfdc#diff-614c78fa46f6db4dd982f8504c665901R1799-R1804](https://github.com/InsightSoftwareConsortium/ITK/commit/d52dfdc44edfd2d6048a712a88752ec3f5a73a4d#diff-614c78fa46f6db4dd982f8504c665901R1799-R1804)
> 
> would work if the sform matrix sto_xyz is not a pure rotation. For example, if the voxels were anisotropic, wouldn't sto_xyz include a scaling?

@cookpa  In the case where sform is not a pure rotation, the highlighted code would return "false" and the sform would not be used.  The new code ONLY works if sform is orthonormal (as is the case with the qform).

hjmjohnson:
@cookpa identified a logical problem with the implementation of testing for orthonormal in the sform code.  This needs more reviewing and a better test case.

hjmjohnson:
@cookpa This should now be ready for the final review.  I build ANTs against it, and all 130 tests pass.

A test was also added, along with a set of testing data.

hjmjohnson:
BRAINSTools (which also heavily uses nifti) also succeeded.   I'm taking the previous approvals, and the two acknowlegements as approval to commit.

PR #1880, opened 2020-06-18
STYLE: Move ITK5 ranges, shapes, policies out of Experimental namespace

N-Dekker:
Moved the iterator ranges, image neighborhood shapes, and pixel access
policies, introduced with ITK 5, out of the `Experimental` namespace.

### DIFF
--- a/Modules/Core/Common/include/itkIndexRange.h
+++ b/Modules/Core/Common/include/itkIndexRange.h
@@ -30,8 +30,6 @@
 
 namespace itk
 {
-namespace Experimental
-{
 
 /**
   * \class IndexRange
@@ -459,7 +457,6 @@ using ImageRegionIndexRange = IndexRange<VDimension, false>;
 template <unsigned VDimension>
 using ZeroBasedIndexRange = IndexRange<VDimension, true>;
 
-} // namespace Experimental
 } // namespace itk
 
 #endif
--- a/Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.hxx
+++ b/Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.hxx
@@ -42,7 +42,7 @@ BSplineInterpolationWeightFunction<TCoordRep, VSpaceDimension, VSplineOrder>::BS
 
   unsigned int counter = 0;
 
-  for (const IndexType index : Experimental::ZeroBasedIndexRange<VSpaceDimension>(m_SupportSize))
+  for (const IndexType index : ZeroBasedIndexRange<VSpaceDimension>(m_SupportSize))
   {
     for (unsigned int j = 0; j < SpaceDimension; j++)
     {
--- a/Modules/Core/Common/include/itkBufferedImageNeighborhoodPixelAccessPolicy.h
+++ b/Modules/Core/Common/include/itkBufferedImageNeighborhoodPixelAccessPolicy.h
@@ -27,9 +27,6 @@
 
 namespace itk
 {
-namespace Experimental
-{
-
 /**
  * \class BufferedImageNeighborhoodPixelAccessPolicy
  * ImageNeighborhoodPixelAccessPolicy class for ShapedImageNeighborhoodRange.
@@ -131,7 +128,6 @@ public:
   }
 };
 
-} // namespace Experimental
 } // namespace itk
 
 #endif
--- a/Modules/Core/Common/include/itkConnectedImageNeighborhoodShape.h
+++ b/Modules/Core/Common/include/itkConnectedImageNeighborhoodShape.h
@@ -39,8 +39,6 @@
 
 namespace itk
 {
-namespace Experimental
-{
 
 /**
  * \class ConnectedImageNeighborhoodShape
@@ -283,7 +281,6 @@ GenerateConnectedImageNeighborhoodShapeOffsets() ITK_NOEXCEPT
   return offsets;
 }
 
-} // namespace Experimental
 } // namespace itk
 
 #undef ITK_X_ASSERT
--- a/Modules/Core/Common/include/itkConstShapedNeighborhoodIterator.h
+++ b/Modules/Core/Common/include/itkConstShapedNeighborhoodIterator.h
@@ -337,7 +337,7 @@ public:
   }
 
   /** Activates a whole range of offsets, for example, an std::vector<OffsetType>,
-   * which could be from Experimental::GenerateImageNeighborhoodOffsets(shape). */
+   * which could be from GenerateImageNeighborhoodOffsets(shape). */
   template <typename TOffsets>
   void
   ActivateOffsets(const TOffsets & offsets)
--- a/Modules/Core/Common/include/itkConstantBoundaryImageNeighborhoodPixelAccessPolicy.h
+++ b/Modules/Core/Common/include/itkConstantBoundaryImageNeighborhoodPixelAccessPolicy.h
@@ -25,8 +25,6 @@
 
 namespace itk
 {
-namespace Experimental
-{
 
 /**
  * \class ConstantBoundaryImageNeighborhoodPixelAccessPolicy
@@ -148,7 +146,7 @@ public:
   }
 };
 
-} // namespace Experimental
+
 } // namespace itk
 
 #endif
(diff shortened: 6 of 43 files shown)


N-Dekker:
Note: This PR is under construction (WIP), comment by @dzenanz (Oct 16, 2019) regarding ITK 5.0 backward compatibility still to be addressed:
> Are any remote modules using Experimental namespace? If not, we don't need this

https://github.com/InsightSoftwareConsortium/ITK/pull/1316#discussion_r335469363

Modules/Core/Common/include/itkIndexRange.h
```diff
@@ -459,7 +457,16 @@ using ImageRegionIndexRange = IndexRange<VDimension, false>;
 template <unsigned VDimension>
 using ZeroBasedIndexRange = IndexRange<VDimension, true>;
 
+
+#if !defined(ITK_LEGACY_REMOVE)
+namespace Experimental // For ITK 5.0 legacy support
```
dzenanz, line 462: Some of these were introduced with 5.1, right? So the comment could be changed to indicate that.

N-Dekker: `ImageRegionRange` was introduced with ITK 5.1, indeed. Very well observed, @dzenanz The other ranges were already in ITK 5.0

My intention for now was to just remove all `#if !defined(ITK_LEGACY_REMOVE)` stuff that I originally added with this PR. Including the comments. OK?

dzenanz: Sounds good. I was not sure whether you would keep `#ifdef`s.

N-Dekker:
Still failed to amend the commit:

> pre-commit hook failure
> -----------------------
>
> clang-format executable was not found.
>
> A clang-format binary will be downloaded and configured when ITK
> is built with the BUILD_TESTING CMake configuration option enabled.
>
> Alternatively, install clang-format version 8.0 or set the executable location with
>
>   git config clangFormat.binary /path/to/clang-format
>
> git did not exit cleanly (exit code 1) (8094 ms @ 18/06/2020 22:34:36)

dzenanz:
I think that happens when you build a remote module externally, it messes up the clang-format in the main ITK repo. @hjmjohnson or @thewtex might know more.

N-Dekker:
Moreover, my VS2019 output said:
>Performing configure step for 'KWStyle'
>CMake Error : error : generator platform: x64
>Does not match the platform used previously:
>Either remove the CMakeCache.txt file and CMakeFiles directory or choose a different binary directory.
>Done building project "KWStyle.vcxproj" -- FAILED.

I'm building for 64-bit (the default, nowadays)

N-Dekker:
Looks like I manually have to set the path of the clang-format executable, right? Like this:

> git config clangFormat.binary /d/X/bin/ITK/clang-format-Windows.exe

thewtex:
> CMake Error : error : generator platform: x64
> Does not match the platform used previously:

This can happen when the initial CMake configuration targeted a 32-bit build, but an attempt was made to switch it to 64-bit. The solution is to remove the build tree and start with 64-bit CMake generator from the start.

> Looks like I manually have to set the path of the clang-format executable, right?

Yes, you can manually specify the executable. However, if ITK is successfully built with `BUILD_TESTING` enabled, the build will have generated the executables and configured Git to where their location :-).

N-Dekker:
@ thewtex 
> > CMake Error : error : generator platform: x64
> > Does not match the platform used previously:
> 
> This can happen when the initial CMake configuration targeted a 32-bit build, but an attempt was made to switch it to 64-bit. 

Thanks Matt, but I did not initially target 32-bit! I just targeted the default platform for VS2019, using CMake (version 3.17), which is actually x64!. When Visual Studio 16 (2019) is selected as generator, cmake-gui says: 

    Optional platform for generator(if empty, generator uses: x64)

But when I leave it empty, I get that CMake Error!

Workaround for ITK users (apparently): explicitly specify the platform, when using Visual Studio as CMake generator, and do not leave it empty!

N-Dekker:
I think this PR is ready now 😃

phcerdan:
Breaking API without previous deprecation is no good IMO. There are third-parties other than ITKModules.

```cpp
namespace Experimental {
  using IndexRange ITK_DEPRECATED_MSG("Use `itk::` namespace instead of `itk::Experimental::`") = itk::IndexRange;
  // Etc.
}
```

N-Dekker:
> Breaking API without previous deprecation is no good IMO. There are third-parties other than ITKModules.
> 
> ```c++
> namespace Experimental {
>   using IndexRange ITK_DEPRECATED_MSG("Use `itk::` namespace instead of `itk::Experimental::`") = itk::IndexRange;
>   // Etc.
> }
> ```

@phcerdan Thanks for your suggestion, Pablo, it would be great if you could make that a PR. I did originally propose to preserve ITK 5.0 backward compatibility, by `#if !defined(ITK_LEGACY_REMOVE)`..., but Dzenan (@dzenanz) though that it would not be necessary:
https://github.com/InsightSoftwareConsortium/ITK/pull/1316#discussion_r335469363  Dzenan, what do you think now?

phcerdan:
> @phcerdan Thanks for your suggestion, Pablo, it would be great if you could make that a PR. I did originally propose to preserve ITK 5.0 backward compatibility, by #if !defined(ITK_LEGACY_REMOVE)..., but Dzenan (@dzenanz) though that it would not be necessary:

Thanks @N-Dekker , I failed to find a way to add [[deprecated]] to a [using directive](https://en.cppreference.com/w/cpp/language/namespace) like this one.
Your solution was fine for me, but if it was rejected, no problems, I will update ITK in my third-party.

I feel ITK is too eager to break API's, people have expectations on semantic versioning.

N-Dekker:
@phcerdan Actually I had to work around this pull request myself, in a test for elastix, which uses _either_ ITK 5.1 _or_ a release candidate of ITK 5.2, by adding both `using namespace itk` and `using namespace itk::Experimental`:

https://github.com/SuperElastix/elastix/blob/1c294607e55e18018874005e57aac458b6f4e013/Core/Main/GTesting/elxCoreMainGTestUtilities.h#L103-L106 

    void
    FillImageRegion(itk::Image<TPixel, VImageDimension> & image,
                    const itk::Index<VImageDimension> &   regionIndex,
                    const itk::Size<VImageDimension> &    regionSize)
    {
      // ImageRegionRange is to be moved from namespace itk::Experimental
      // to namespace itk with ITK version 5.2.
      using namespace itk;
      using namespace itk::Experimental;

      const ImageRegionRange<Image<TPixel, VImageDimension>> imageRegionRange{
        image, ImageRegion<VImageDimension>{ regionIndex, regionSize }
      };
      std::fill(std::begin(imageRegionRange), std::end(imageRegionRange), 1);
    }

So I agree with you that it is a bit cumbersome...!

phcerdan:
That's a nice workaround, I'll use it :)

Thanks @N-Dekker!

N-Dekker:
> That's a nice workaround, I'll use it :)

For the sake of completeness, this is the initial part of the workaround:
https://github.com/SuperElastix/elastix/blob/1c294607e55e18018874005e57aac458b6f4e013/Core/Main/GTesting/elxCoreMainGTestUtilities.h#L37-L43

```
namespace itk
{
namespace Experimental
{
// Workaround to allow using things that may be either in itk or in itk::Experimental.
}
} // namespace itk

```

To ensure that there is actually an `itk::Experimental` namespace defined, even when you use the latest ITK 5.2!!!

phcerdan:
@N-Dekker , how do you ensure that the namespace `itk::Experimental` exists?
I am getting:
```
 error: ‘Experimental’ is not a namespace-name
```

I can fix adding to my cpp file:
```cpp
namespace itk{
    namespace Experimental{
    }
}
```

phcerdan:
Nice, we commented at the same time, that will do. Thanks again @N-Dekker!

dzenanz:
Experimental namespace, by definition, does not provide backwards compatibility or API stability. I didn't think that anyone other than Niels used it, but I guess I was wrong 😄

phcerdan:
@dzenanz Yeah, I think your guess was safe enough!

PR #1882, opened 2020-06-19
ENH: Prevent segfault by failing fast in itkGDCMImageIO

Raphexion:
In InternalReadImageInformation:

When pixel type is larger than the output type, it is likely
something wrong in the dataset currently being read.

An example is if the following two headers are "out-of-sync"
(0028,0100) BitsAllocated
(0028,0101) BitsStored

Instead of ignoring this sanity check in release-mode,
it is better to throw an exception. That way, the caller
can handle the error instead of seeing the application crash.

Please note the following warning in
Documentation/ReleaseNotes/3.14.md

"when using itkAssertInDebugOrThrowInReleaseMacro do not
terminate with a ;"

Signed-off-by: Niklas Johansson <raphexion@gmail.com>

### DIFF
--- a/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
+++ b/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
@@ -477,7 +477,10 @@ GDCMImageIO::InternalReadImageInformation()
     outputpt = r.ComputeInterceptSlopePixelType();
   }
 
-  itkAssertInDebugAndIgnoreInReleaseMacro(pixeltype <= outputpt);
+  if (pixeltype > outputpt)
+  {
+    itkAssertInDebugOrThrowInReleaseMacro("Pixel type larger than output type")
+  }
 
   m_ComponentType = IOComponentEnum::UNKNOWNCOMPONENTTYPE;
   switch (outputpt)


thewtex:
CI is failing

Raphexion:
@thewtex I am unable to really see what I broke. I see this:

```
CMake Error: Unable to read from file 'Testing/Temporary/itkColorTableTest1.txt': Can't lstat Testing/Temporary/itkColorTableTest1.txt
Error creating tar while encoding file: /home/vsts/work/1/s-build/Testing/Temporary/itkColorTableTest1.txt
```

If you have time and want to, could you point me in the right direction.

dzenanz:
https://open.cdash.org/viewBuildError.php?buildid=6604065

I think there should be no `<<`.

@thewtex Can we get CDash link back? Sifting through log files is quite impractical.

Raphexion:
@dzenanz Thank you for the quick help.

thewtex:
> @thewtex Can we get CDash link back? Sifting through log files is quite impractical.

Yes, I will look into the CDash status link.

Raphexion:
@maekclena If you have time and want to, I think there is missing one review for approval.

PR #1898, opened 2020-06-28
External wrapping

thewtex:


LucasGandel:
@thewtex I could not find time to try this with RTK but it looks safe to me. +1

PR #1905, opened 2020-07-02
Restore license

dzenanz:
I think the reason for conflicts is git's imperfect merge heuristic. I might be wrong, but manually resolving the conflict during update (once per year or so) is not that hard.


blowekamp:
The restoration looks fine. What happens when the `ITK/Modules/ThirdParty/PNG/UpdateFromUpstream.sh` is run again?

It seems odd to me that the subtree merge "broke" out of the ITK subdirectory. Perhaps @bradking  might have some insights to this issue.

dzenanz:
If update is run, merge conflict occurs again.

dzenanz:
Also cherry-picked to release.

Issue #1673, opened 2020-03-03
Progress reporting should happen every X (milli)seconds


dzenanz:
Add timed updates, perhaps inspired by #1571.

blowekamp:
Per the discussion in [discourse](https://discourse.itk.org/t/no-progress-reporting-in-itk5/2630) it was concluded that this was an _optional_ feature to implement and there are alternatives to implement for applications which desire more frequent updates while having less impact of the many application which don't use progress.

For the default pool multi-threader, the current issue for smooth progress reporting, is the idle behavior for the invoking threaded after the first chunk. IMHO this should have a higher priority to address.

PR #1914, opened 2020-07-08
ENH: progress reporting does not stall with PoolMultiThreader

dzenanz:
Implementing the suggestion made in [a forum post](https://discourse.itk.org/t/no-progress-reporting-in-itk5/2630/33?u=dzenanz). Related to #1673.
## PR Checklist

### DIFF
--- a/Modules/Core/Common/src/itkPoolMultiThreader.cxx
+++ b/Modules/Core/Common/src/itkPoolMultiThreader.cxx
@@ -38,6 +38,8 @@ namespace itk
 {
 namespace
 {
+std::chrono::milliseconds threadCompletionPollingInterval = std::chrono::milliseconds(10);
+
 class ExceptionHandler
 {
 public:
@@ -157,6 +159,11 @@ PoolMultiThreader ::ParallelizeArray(SizeValueType             firstIndex,
                                      ArrayThreadingFunctorType aFunc,
                                      ProcessObject *           filter)
 {
+  if (!this->GetUpdateProgress())
+  {
+    filter = nullptr;
+  }
+
   if (firstIndex + 1 < lastIndexPlus1)
   {
     SizeValueType chunkSize = (lastIndexPlus1 - firstIndex) / m_NumberOfWorkUnits;
@@ -193,8 +200,16 @@ PoolMultiThreader ::ParallelizeArray(SizeValueType             firstIndex,
     // now wait for the other computations to finish
     for (SizeValueType i = 1; i < workUnit; i++)
     {
-      exceptionHandler.TryAndCatch([this, i, &reporter] {
-        m_ThreadInfoArray[i].Future.get();
+      exceptionHandler.TryAndCatch([this, i, &reporter, &filter] {
+        std::future_status status;
+        do
+        {
+          status = m_ThreadInfoArray[i].Future.wait_for(threadCompletionPollingInterval);
+          if (filter)
+          {
+            filter->IncrementProgress(0);
+          }
+        } while (status != std::future_status::ready);
         reporter.CompletedPixel();
       });
     }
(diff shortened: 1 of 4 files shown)


Modules/Core/Common/src/itkPoolMultiThreader.cxx
```diff
@@ -273,7 +281,12 @@ PoolMultiThreader ::ParallelizeImageRegion(unsigned int         dimension,
...
       {
         exceptionHandler.TryAndCatch([this, i, &reporter] {
-          m_ThreadInfoArray[i].Future.get();
+          std::future_status status;
+          do
+          {
+            status = m_ThreadInfoArray[i].Future.wait_for(threadCompletionPollingInterval);
+            reporter.Completed(0);
+          } while (status != std::future_status::ready);
           reporter.CompletedPixel();
```
blowekamp, line 290: Using the TotalProgressReporter, the progress is going to be incremented by the chuck amount. The filter may  also be doing the progress, the two will be added together causing some inaccuracy/difficulties.

dzenanz: What is "chuck amount"? I followed your other suggestions.

blowekamp: Looks like you reverted to just the progress reported 👍

Modules/Core/Common/src/itkPoolMultiThreader.cxx
```diff
@@ -193,8 +200,16 @@ PoolMultiThreader ::ParallelizeArray(SizeValueType             firstIndex,
...
-      exceptionHandler.TryAndCatch([this, i, &reporter] {
-        m_ThreadInfoArray[i].Future.get();
+      exceptionHandler.TryAndCatch([this, i, &reporter, &filter] {
+        std::future_status status;
+        do
+        {
+          status = m_ThreadInfoArray[i].Future.wait_for(threadCompletionPollingInterval);
+          if (filter)
+          {
+            filter->IncrementProgress(0);
```
blowekamp, line 210: This fine tweak to consider, but this method always trigger a ProgressReport event. There are some cases where regular updates may be nice ( as this should provide ) or situations where the performance of not emitting extra events would be desired.

dzenanz:
I went with your code suggestions Brad. Platform multi-threader does not deal with progress at all, as it is only meant for backwards compatibility.

Writing a good test for this is not trivial. Perhaps you or @spinicist could do that? If not, let's not let perfect be the enemy of the good 😄

spinicist:
Hello,

Thanks for getting round to this and including me. I'm not working on the project that needs this right at the moment - hopefully will get back to it in a couple of weeks.

I'm wondering what a test for this would look like? Print the progress messages to a string buffer and then check they follow a known pattern?

dzenanz:
Come up with some dummy work (could even just be sleeping), add progress observer, and check that progress is reported a few times for 50%-95% range. A number of threads could be programatically fixed to say 4, and number of work units to say 20. Also explicitly request Pool multi-threader.

blowekamp:
I manually tested this patch with the itkMedianImageFilterTest on a large image with a the Pool and Platform threaders with a variety of different number of threads. The progressed reporting seemed to work very well!

Also SimpleITK's tests all passed when build against this PR.

Thanks for coming up with this approach to improve the regularity or reporting.

blowekamp:
I manually tested this patch with the itkMedianImageFilterTest on a large image with a the Pool and Platform threaders with a variety of different number of threads. The progressed reporting seemed to work very well!

Also SimpleITK's tests all passed when build against this PR.

dzenanz:
Tobias, do you want to add a unit test in this PR? Or perhaps plan to do it in a follow-up PR?

spinicist:
Hi, sadly I have not had time to get back to the project that needs this, and I don't know when I may do so. I very much appreciate your efforts on this and don't want to hold you up so please go ahead and merge, and when I finally get back to ITK I will look at adding a test.

PR #1915, opened 2020-07-08
ENH: Update building internal FFTW version

hjmjohnson:
Recent FFTW builds support cmake building which more
naturally fits with ITK building.

Several variables were conditional based on
if SYSTEM version or local version were harmonized
so that the behavior of the builds is more consistent
across configuration options.  This also allows
for some simplification of the dependant code.

## PR Checklist


hjmjohnson:
@thewtex I don't have access to a windows box at home.  I think this may allow FFTW to be built on windows, which would be a nice improvement (previously cmake failed immediately if it detected non-system windows FFTW use).

Modules/ThirdParty/NIFTI/stop
```diff
@@ -0,0 +1,735 @@
+/** \file nifti2_io.h
```
dzenanz, line 1: Was this file committed on purpose?

CMake/itkExternal_FFTW.cmake
```diff
@@ -12,153 +30,152 @@ if(NOT ITK_USE_MKL AND NOT ITK_USE_CUFFTW)
...
+        UPDATE_COMMAND ""
+        LOG_CONFIGURE 1
+        LOG_BUILD 1
+        LOG_INSTALL 1
+        CMAKE_CACHE_ARGS
+           -DBUILD_SHARED_LIBS:BOOL=${BUILD_SHARED_LIBS}
+           -DBUILD_TESTS:BOOL=OFF
+           -DCMAKE_BUILD_TYPE:STRING=${FFTW_BUILD_TYPE}
+           -DCMAKE_INSTALL_PREFIX:PATH=${FFTW_STAGED_INSTALL_PREFIX}
+           -DDISABLE_FORTRAN:BOOL=OFF
```
dzenanz, line 76: Only FORTRAN has a disable prefix?

gdevenyi:
We recently had issues discovering that the FFTW cmake is not feature complete vs the ./configure version https://github.com/BIC-MNI/minc-toolkit-v2/issues/116 in particular building the code properly for a generic release which supports multiple processors.

Our next step was to swap in one of the FFTW-equivalent codebases that are now available: https://github.com/project-gemmi/benchmarking-fft

hjmjohnson:
> We recently had issues discovering that the FFTW cmake is not feature complete vs the ./configure version [BIC-MNI/minc-toolkit-v2#116](https://github.com/BIC-MNI/minc-toolkit-v2/issues/116) in particular building the code properly for a generic release which supports multiple processors.
> 
> Our next step was to swap in one of the FFTW-equivalent codebases that are now available: https://github.com/project-gemmi/benchmarking-fft

Thanks for the info.  This PR states that this should be a reference implementation only.  Not a high performant solution.

CMake/itkExternal_FFTW.cmake
hjmjohnson: yes. for the cmake interface.

Modules/ThirdParty/NIFTI/stop
hjmjohnson: No.  Will remove soon.

CMake/itkExternal_FFTW.cmake
hjmjohnson: I did not notice that earlier, and the new version has DISABLE_FORTRAN:BOOL=ON.

gdevenyi:
> Thanks for the info. This PR states that this should be a reference implementation only. Not a high performant solution.

Just to be clear, the issue we ran into is that the cmake version can build code that can't actually run on the host system.

Issue #1906, opened 2020-07-03
OpenCVImageBridge does not support signed 32-bit integer matrix conversions.


PierreWargnier:
### Description

Attempts to convert an OpenCV cv::Mat single-channel signed 32-bit integers (CV_32SC1) matrix to itk::Image<int, 2> throw an exception saying "Unknown OpenCV type" whereas this type does exist and there is no reason for it not to be compatible with the itk Image class.

The inverse conversion itk::Image<int, 2> to cv::Mat of type CV_32SC1 also throws an exception: "OpenCV does not support the input pixel type" whereas it is actually supported.

By reading the code in itkOpenCVImageBridge.hxx, it appears that this type was simply forgotten because all other OpenCV types are supported.

### Steps to Reproduce

    #include <opencv2/core.hpp>
    #include <itkImage.h>
    #include <itkOpenCVImageBridge.h>

    int main()
    {
        using IntImageType = itk::Image<int, 2>;

        int width = 1920;
        int height = 1080;

        cv::Mat singleChannelIntMat(height, width, cv::CV_32FC1);

        IntImageType::Pointer itkIntImage = itk::OpenCVImageBridge::CVMatToITKImage<IntImageType>(singleChannelIntMat);

       cv::Mat inverseConversion = itk::OpenCVImageBridge::ITKImageToCVMat<IntImageType>(itkIntImage);
   
       return 0;
    }

### Expected behavior

Conversion should work, since both libraries support int images/matrices.

### Actual behavior

- Conversion from CV_32SC1 Mat to Image<int, 2> throws "Unknown OpenCV type".

- Conversion from Image<int, 2> to CV_32SC1 Mat throws "OpenCV does not support the input pixel type".

### Reproducibility

Always

### Versions

ITK 5.2.0
OpenCV 4.3.0

Unless the signed 32-bit int type was ever supported, this is likely to appear with any versions of both libraries.

### Environment

Ubuntu 20.04 Virtual Box virtual machine. But this is clearly an issue independant of the platform since the int matrix type is clearly missing in the itkOpenCVImageBridge.hxx file.

### Additional Information

There is an easy fix that I implemented and tested:

- In itkOpenCVImageBridge.hxx after line 74, adding
`CONVERSION_CASE(IPL_DEPTH_32S, int)`
fixes the OpenCV -> ITK converstion issue.

- In itkOpenCVImageBridge.hxx after line 196, adding
`
[...]

Issue #1651, opened 2020-02-24
OpenCVBridge Compile error ITK 5.1RC1


Hicham-Damae:
### Description
I managed to compile ITK 5.1RC1 with OpenCV Bridge, after some corrections (already fixed on master or ongoing). 
Everything works pretty well, except when I use OpenCVImageBridge function.

I have the folowing error : 
C:\Program Files\ITK\Release\include\ITK-5.1\itkOpenCVImageBridge.hxx(98): error C2440: 'initializing': cannot convert from 'const cv::Mat' to 'const IplImage'

I fixed this error using
```C++
const IplImage converted = cvIplImage(in);
instead of 
const IplImage converted = in;
```
on line 98 (itkOpenCVImageBridge.hxx) 

### Versions
OpenCV 4.2
ITK 5.1RC1

### Environment
Windows 10 x64 
MSVC 2019 16.4

dzenanz:
Is this issue still open? #1650 might have touched it.

Giaco90:
@Hicham-Damae solution worked for ITK 5.0.1 and OpenCV 4.2.0 too.

dzenanz:
Can either of you turn it into a [PR](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md)?

PierreWargnier:
Hi, I was having the same error with ITK 5.2.0 and OpenCV 4.3.0 (latest stable version of both). The solution proposed by @Hicham-Damae works for me too. This is due to the fact that IplImage has been deprecated in OpenCV for quite a long time now. It was eventually removed in version 4.0.0 and only remains in the C API. The fix works because it is explicitely invoking the C API. This however, is only a temporary fix because the OpenCV C API is likely to be completely removed in future versions (it is already absent from the documentation as of version 4.3.0 - see [https://docs.opencv.org/4.3.0/](https://docs.opencv.org/4.3.0/) - although it appears to be still accessible).

PR #1917, opened 2020-07-09
Opencv bridge update

PierreWargnier:
BUG: Fix Mat to Image conversions for OpenCV 4.x; add support for int32

This fixes issue https://github.com/InsightSoftwareConsortium/ITK/issues/1651 in a lasting manner by introducing specific implementations for functions CVMatToITKImage and ITKImageToCVMat that no longer depend on the legacy C API while avoiding unnescessary memory copies.

This also fixes issue https://github.com/InsightSoftwareConsortium/ITK/issues/1906 by adding support for int pixel type (IPL_DEPTH_32S in the legacy C API, CV_32S in the C++ API) for both OpenCV legacy C and C++ APIs.

In addition, exceptions for pixel/component type mismatches and impossible conversions were added to both IplImageToITKImage and CVMatToITKImage functions. Some extra functions had to be introduced to ensure minimal code duplication/redundancy.

Desired behavior was discussed with @dzenanz . It will remain the same as before, only with extra checks for type mismatches and unsupported color conversions.

Note: for future improvements it could be useful to allow 2, 3 and 4-channels conversions for any OpenCV-supported types (uchar, char, ushort, short, int, float, double). The only constraints are that color conversions will only be done when supported and in other cases, a simple buffer copy should be done. It will then be up to the user to manage data representation. Support for RGBA/BGRA would also be nice, but I am not sure if it is supported in ITK.

## PR Checklist

### DIFF
--- a/Modules/Video/BridgeOpenCV/include/itkOpenCVImageBridge.hxx
+++ b/Modules/Video/BridgeOpenCV/include/itkOpenCVImageBridge.hxx
@@ -28,9 +28,7 @@
 
 namespace itk
 {
-//
 // IplImageToITKImage
-//
 template <typename TOutputImageType>
 typename TOutputImageType::Pointer
 OpenCVImageBridge::IplImageToITKImage(const IplImage * in)
@@ -43,17 +41,13 @@ OpenCVImageBridge::IplImageToITKImage(const IplImage * in)
   // trying to implicit cast some IPL_DEPTH_* values
   using DepthIDType = unsigned int;
 
-  //
   // Make sure input isn't null and output type is 2D or 1D
-  //
   if (!in)
   {
     itkGenericExceptionMacro("Input is nullptr");
   }
 
-  //
   // Do the conversion
-  //
   typename ImageType::Pointer out = ImageType::New();
 
 #define CONVERSION_CASE(iplInputDepthID, itkOutputPixelType)                                                           \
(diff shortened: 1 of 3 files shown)


Modules/Video/BridgeOpenCV/include/itkOpenCVImageBridge.hxx
```diff
@@ -251,14 +297,135 @@ template <typename TInputImageType>
...
+  //
+  // Set up the output image
+  //
+  Mat          tmp;
+  unsigned int w = static_cast<unsigned int>(size[0]);
+  unsigned int h = static_cast<unsigned int>(size[1]);
+
+  //
+  // set the depth correctly based on input pixel type
+  //
```
dzenanz, line 350: These empty comment lines before  and after the actual comment are not exactly needed. They reduce code density necessitating more scrolling. Please remove them.

PierreWargnier: I actually was not the one to format the comments like that. It was in the original code so I thought I should leave it like that. I do agree however that this is not a great coding style. I also realised that I left one include (<iostream>) that was only for debug when I was testing my changes.

I am willing to  make these corrections. However, this is the first time I submit a pull request and I am not sure how to submit the changes.

If I understand correctly, you also want me to ammend my commit message to include the comment about ideas for further improvements. Is that correct?

dzenanz: Corrections: amend the commit and then force-push to the same branch (`opencv-bridge-update`).

Ideas for future improvement were my message to you, and should not be included in commit message. Thanks for working on this!

PierreWargnier: Hi @dzenanz, the cleanup and correction of comment format is done. I also removed the useless comments saying "return result".

PierreWargnier:
Regarding the additional support of arbitrary number of channels, it would be nice to have but I do not think it is essential. I personnally do not need it at the moment and do not really have time to implement it right now.

PR #1919, opened 2020-07-13
BUG: Empty image support in image_from_xarray

thewtex:
Required for map_blocks.


thewtex:
@ZhiyLiu please take a look

PR #1923, opened 2020-07-16
ENH: Add image.astype(pixel_type) for casting

thewtex:
Similar to the numpy.ndarray.astype, supporting itk PixelType's and
numpy dtype's.

More convenient than itk.cast_image_filter, similar to
numpy.ndarray.astype. itk.Image PixelType's are supported along with
scalar numpy dtype's. itkType.itkCType.GetCTypeForDType is added to
support conversion.


thewtex:
Failing test unrelated.

PR #1927, opened 2020-07-20
BUG: Add missing enumerate with multi-ndarray-output itk filters

thewtex:
Required for returning the (registered_image, transform_parameters)
output when ndarray's is passed for the fixed and moving image in
itk-elastix.


romangrothausmann:
Does this relate to https://github.com/InsightSoftwareConsortium/ITKElastix/issues/37?

thewtex:
> Does this relate to InsightSoftwareConsortium/ITKElastix#37?

This is a separate issue, but we are on the path to address it :-)

PR #1942, opened 2020-07-30
ENH: accommodate dask_loads for default serializers of dask

ZhiyLiu:
## PR Checklist

Dask loads method may complain TypeError: No dispatch for <class 'itkPyBufferPython.NDArrayITKBase'> due to the missing of a deserialization method for NDArrayITKBase. This commit tries to solve this issue by registering a deserialization method in dask for NDArrayITKBase. 

Correspondingly, the test case verifies the pickling and unpickling of a NDArrayITKBase instance.

### DIFF
--- a/Modules/Bridge/NumPy/wrapping/PyBuffer.i.init
+++ b/Modules/Bridge/NumPy/wrapping/PyBuffer.i.init
@@ -17,6 +17,22 @@ try:
 
 except ImportError:
   HAVE_NUMPY = False
+try:
+    import numpy as np
+    from distributed.protocol import dask_serialize, dask_deserialize
+    from typing import Dict, List, Tuple
+except ImportError:
+    pass
+else:
+    @dask_serialize.register(NDArrayITKBase)
+    def serialize(ndarray_itk_base: NDArrayITKBase) -> Tuple[Dict, List[bytes]]:
+        dumps = dask_serialize.dispatch(np.ndarray)
+        return dumps(ndarray_itk_base)
+
+    @dask_deserialize.register(NDArrayITKBase)
+    def deserialize(header: Dict, frames: List[bytes]) -> NDArrayITKBase:
+        loads = dask_deserialize.dispatch(np.ndarray)
+        return NDArrayITKBase(loads(header, frames))
 
 def _get_numpy_pixelid(itk_Image_type):
     """Returns a ITK PixelID given a numpy array."""
--- a/Modules/Bridge/NumPy/wrapping/test/itkPyBufferTest.py
+++ b/Modules/Bridge/NumPy/wrapping/test/itkPyBufferTest.py
@@ -28,6 +28,45 @@ class TestNumpyITKMemoryviewInterface(unittest.TestCase):
 
     def setUp(self):
         pass
+    def test_NDArrayITKBase_pickle(self):
+        """
+        Test the serialization of itk.NDArrayITKBase
+        """
+        Dimension             = 3
+        ScalarImageType       = itk.Image[itk.UC, Dimension]
+        RegionType            = itk.ImageRegion[Dimension]
+
+        region                = RegionType()
+        region.SetSize(0, 6);
+        region.SetSize(1, 6);
+        region.SetSize(2, 6);
+
+        scalarImage           = ScalarImageType.New()
+        scalarImage.SetRegions(region);
+        scalarImage.Allocate(True);
+        scalarImage.SetPixel([0, 0, 0], 5)
+        scalarImage.SetPixel([0, 0, 1], 3)
+        scalarImage.SetPixel([5, 5, 5], 8)
+        ndarray_itk_base      = itk.array_view_from_image(scalarImage)
+
+        import pickle
+
+        ## test serialization of itk ndarrary itk base
+        pickled = pickle.dumps(ndarray_itk_base)
+        reloaded = pickle.loads(pickled)
+        equal = (reloaded == ndarray_itk_base).all()
+        assert equal, 'Different results before and after pickle'
+
+        try:
+            import dask
+            from distributed.protocol.serialize import dask_dumps, dask_loads
+        except ImportError:
+            pass
+        else:
+            header, frames = dask_dumps(ndarray_itk_base)
+            recon_obj = dask_loads(header, frames)
+            equal = (recon_obj == ndarray_itk_base).all()
+            assert equal, 'Different results before and after pickle'
 
     def test_NumPyBridge_itkScalarImage(self):
         "Try to convert all pixel types to NumPy array view"


thewtex:
@ZhiyLiu thanks!

The `ITK.Linux.Python` CI test is failing -- do the tests pass for you locally from a fresh build?

thewtex:
@mrocklin @jakirkham how does this look to you?

ZhiyLiu:
> @ZhiyLiu thanks!
> 
> The `ITK.Linux.Python` CI test is failing -- do the tests pass for you locally from a fresh build?

Yes. It passes on my local machine. 

Looking at the log. There is an error regarding to the compilation of C++ code, as follows:
> error: ISO C++ forbids declaration of 'ITK_DISALLOW_COPY_AND_ASSIGN' with no type [-fpermissive]
>    ITK_DISALLOW_COPY_AND_ASSIGN(STLContainerAdaptor);

Not sure this is relevant to the python code.

Modules/Bridge/NumPy/wrapping/PyBuffer.i.init
```diff
@@ -17,6 +17,16 @@ try:
...
+try:
+    import numpy as np
+    from distributed.protocol import dask_deserialize
+    from typing import Dict, List
+    @dask_deserialize.register(NDArrayITKBase)
+    def deserialize(header: Dict, frames: List[bytes]) -> NDArrayITKBase:
+        loads = dask_deserialize.dispatch(np.ndarray)
+        return NDArrayITKBase(loads(header, frames))
+except ImportError:
+    pass
```
mrocklin, line 29: It is typical to handle this as follows:

```suggestion
try:
    import numpy as np
    from distributed.protocol import dask_deserialize
    from typing import Dict, List
except ImportError:
    pass
else:
    @dask_deserialize.register(NDArrayITKBase)
    def deserialize(header: Dict, frames: List[bytes]) -> NDArrayITKBase:
        loads = dask_deserialize.dispatch(np.ndarray)
        return NDArrayITKBase(loads(header, frames))
```

Modules/Bridge/NumPy/wrapping/test/itkPyBufferTest.py
```diff
@@ -28,6 +28,34 @@ class TestNumpyITKMemoryviewInterface(unittest.TestCase):
...
+        scalarImage.SetPixel([5, 5, 5], 8)
+        ndarray_itk_base      = itk.array_view_from_image(scalarImage)
+
+        import pickle
+
+        ## test serialization of itk ndarrary itk base
+        pickled = pickle.dumps(ndarray_itk_base)
+        reloaded = pickle.loads(pickled)
+        equal = (reloaded == ndarray_itk_base).all()
+        assert equal, 'Different results before and after pickle'
```
mrocklin, line 58: This test doesn't engage the code that you've written above.  Pickle is the standard python serialization technique.  Dask's serialization is different.  If you wanted to test it you would try the following:

```python
from distributed.protocol import serialize, deserialize

header, frames = serialize(ndarray_itk_base)
obj = deserialize(header, frames)
assert ...
```

Modules/Bridge/NumPy/wrapping/PyBuffer.i.init
mrocklin: I'm not sure that registering a deserialization function would have much of an effect without also registering a serialization function.

jhlegarreta:
> Looking at the log. There is an error regarding to the compilation of C++ code, as follows:
> 
>     error: ISO C++ forbids declaration of 'ITK_DISALLOW_COPY_AND_ASSIGN' with no type [-fpermissive]
>     ITK_DISALLOW_COPY_AND_ASSIGN(STLContainerAdaptor);
> 
> Not sure this is relevant to the python code.

That should have been addressed in db4a897729721ab25f1115ef73c1b07541d37ebb. The Python wrapping and tests succeeded in the related CI builds, so @ZhiyLiu you may want to rebase on `master` when addressing the comments of the rest of the folks.

ZhiyLiu:
@thewtex @mrocklin @jhlegarreta @jakirkham

Thanks for your review and nice comments! This commit aims to solves an issue showed in [this notebook](https://github.com/ZhiyLiu/test_itk_dask/blob/master/test_ndarray_itk_base_serialize.ipynb).

Briefly, the final cell in the notebook complains:
> TypeError: No dispatch for <class 'itk.itkPyBufferPython.NDArrayITKBase'>

if no deserialization method is registered for NDArrayITKBase.

Honestly, this kind usage of dask is not quite common. In the notebook I scatter the chunked data; on the contrary, a common scatter will scatter a numpy array.

However, there will be no problem if the submitted function
`    def denoise(block):`

`...`

`    return x          # x = np.asarray(x)`
        
returns a numpy array instead of NDArrayITKBase. This suggests that NDArrayITKBase objects cannot be used like numpy arrays in dask.

There are a couple of solutions to this issue, some of which are commented out in the notebook, e.g.,
`client = Client(n_workers=8, serializers=['pickle'])`

#### This commit is one of these solutions, that needs no special care from a user.

Hope this makes sense.

jakirkham:
Sorry for the slow reply. Some thoughts on options for approaching efficient serialization with Dask.

If one communicates objects that Dask already knows how to work with (like [NumPy arrays]( https://github.com/dask/distributed/blob/2.22.0/distributed/protocol/numpy.py )), then one will get optimized serialization for free. This may involve a little (or maybe no) code in ITK to make sure NumPy arrays are handled reasonable well and handed to Dask when functions return. The benefit here seems useful even outside the Dask context.

If one adds [pickle protocol 5]( https://docs.python.org/3/library/pickle.html#pickle-oob ) support to objects (possibly extending to older Python's with [`pickle5`]( https://github.com/pitrou/pickle5-backport ) or leveraging NumPy to do this indirectly ( https://github.com/numpy/numpy/pull/12091 )), then one can leverage Dask's own support for pickle protocol 5 ( https://github.com/dask/distributed/pull/3784 ) ( https://github.com/dask/distributed/pull/3849 ), which should yield a similar performance boost to Dask's own custom serialization (as the underlying principles are the same). This would require merely updating how pickling works in ITK. It also will work with anything else that supports pickle protocol 5.

Alternatively one can use [Dask's custom serialization]( https://distributed.dask.org/en/latest/serialization.html ) by following [this example on how to extend to custom types]( https://distributed.dask.org/en/latest/serialization.html#id3 ). The usual strategy here is to put these registration bits in its own module. This will soften the Distributed dependency and provide a single point for registering these functions. The latter is useful as one needs to [`import` this module in `distributed`]( https://github.com/dask/distributed/blob/2.22.0/distributed/protocol/__init__.py#L104-L113 ) for it to work. This is doable with a bit of work in ITK and Distributed, but not too difficult.

Any of these options seems fine. Possibly it's worth doing multiple or even all of them (this is what we did in RAPIDS https://github.com/rapidsai/cudf/pull/5139 ). Just trying to show what's possible and what it would entail to do 🙂

thewtex:
@ZhiyLiu looking good! @mrocklin @jakirkham thank you for the detailed reviews and pointers!

From what I can see, with a squash this is ready to merge.

- The `try`, `except`, `else` syntax suggested by @mrocklin is used.
- `dask_loads`, `dask_dumps` are now exercised in the tests.
- The missing `serialize` function has been added.
- @jakirkham did an awesome job (thanks!) of pointing to additional ways to register serialization functions and optimizations for the pickling. For the `NDArrayITKBase` type, we are leveraging NumPy's serialization, but we should write serialization function for other `itk.DataObject`'s. And use the more advanced methods to make the serialization function available (assuming that this will suffice for most use cases).

If I am missing something, please let me know; otherwise, I will merge this on Friday.

thewtex:
Merged in https://github.com/InsightSoftwareConsortium/ITK/commit/27fe268d9dccf2755414022260ba560d301e6336

Issue #1950, opened 2020-08-06
ImageMaskSpatialObject tries to access pixels outside the image buffer (segfault)


N-Dekker:
### Description

[ImageMaskSpatialObject::IsInsideInObjectSpace](https://github.com/InsightSoftwareConsortium/ITK/blob/v5.0.1/Modules/Core/SpatialObjects/include/itkImageMaskSpatialObject.hxx#L46-L62) does in some border cases try to access pixels outside the image buffer, which sometimes triggers a segmentation fault, or an incorrect result. This issue appears related to some specific [elastix 5.0.0](https://github.com/SuperElastix/elastix/tree/5.0.0) segfaults reported by Theo van Walsum (@tvanwalsum, Erasmus MC, Rotterdam) at [Google Groups - elastix-imageregistration - Issues with mask in el 5.0.0 when mask extent is smaller then image extent](https://groups.google.com/forum/#!category-topic/elastix-imageregistration/elastix/lSLWJq9Zcgk).

### Steps to Reproduce Assert Failure

```
#include <itkImage.h>
#include <itkSize.h>
#include <itkImageMaskSpatialObject.h>
#include <cassert>
#include <cstdlib>

int main()
{
	const auto image = itk::Image<unsigned char>::New();
	image->SetRegions(itk::Size<>{ {2, 2} });
	image->Allocate(true);
	const auto imageMaskSpatialObject = itk::ImageMaskSpatialObject<2>::New();
	imageMaskSpatialObject->SetImage(image);
	const double point[] = { 1.5, 1.5 };
	if (imageMaskSpatialObject->IsInsideInObjectSpace(point))
	{
		assert(!"Incorrectly estimated that the specified point is inside the mask!");
		return EXIT_FAILURE;
	}
	return EXIT_SUCCESS;
}
```
### Expected behavior

return EXIT_SUCCESS.

### Actual behavior

assert failure, return EXIT_FAILURE.

### Steps to Reproduce Segfault

```
#include <itkImage.h>
#include <itkSize.h>
#include <itkImageMaskSpatialObject.h>

int main()
{
	const auto image = itk::Image<unsigned char, 3>::New();
	const itk::Size<3> imageSize{ {256, 128, 32} };
	image->SetRegions(imageSize);
	const double origin[] = { 0.0, -16.0, 0.0 };
	image->SetOrigin(origin);
	image->Allocate(true);

[...]

PR #1951, opened 2020-08-06
BUG: Fix issue #1950, "ImageMaskSpatialObject tries to access pixels outside the image buffer (segfault)"

N-Dekker:
Fixed issue #1950, "ImageMaskSpatialObject tries to access pixels
outside the image buffer (segfault)".

In the original code, `TransformPhysicalPointToContinuousIndex` did
estimate whether the specified point was inside the image buffer, but
then `GetInterpolator()->EvaluateAtContinuousIndex(index)` did in some
cases still try to access a pixel outside the image buffer.

This fix avoids using an interpolator. It only accesses a pixel when its
index is inside the buffered region.

The use of an interpolator appears less relevant for an image mask than
for other spatial objects, as for each mask pixel value, it is usually
only interesting to know whether it is zero or non-zero.

Added ImageMaskSpatialObject.CornerPointIsNotInsideMaskOfZeroValues unit
test.

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkImageMaskSpatialObject.hxx
+++ b/Modules/Core/SpatialObjects/include/itkImageMaskSpatialObject.hxx
@@ -40,20 +40,12 @@ template <unsigned int TDimension, typename TPixel>
 bool
 ImageMaskSpatialObject<TDimension, TPixel>::IsInsideInObjectSpace(const PointType & point) const
 {
-  typename Superclass::InterpolatorType::ContinuousIndexType index;
-  if (this->GetImage()->TransformPhysicalPointToContinuousIndex(point, index))
-  {
-    using InterpolatorOutputType = typename InterpolatorType::OutputType;
-    bool insideMask = (Math::NotExactlyEquals(DefaultConvertPixelTraits<InterpolatorOutputType>::GetScalarValue(
-                                                this->GetInterpolator()->EvaluateAtContinuousIndex(index)),
-                                              NumericTraits<PixelType>::ZeroValue()));
-    if (insideMask)
-    {
-      return true;
-    }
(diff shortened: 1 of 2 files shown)


N-Dekker:
The test failure that this PR should fix, once it's ready:

> D:\a\1\s\Modules\Core\SpatialObjects\test\itkImageMaskSpatialObjectGTest.cxx(324): error: Value of: imageMaskSpatialObject->IsInsideInObjectSpace(cornerPoint)
>  Actual: true
> Expected: false
> [  FAILED  ] ImageMaskSpatialObject.CornerPointIsNotInsideMaskOfZeroValues (1 ms)


https://dev.azure.com/itkrobotwindow/ITK.Windows/_build/results?buildId=4422&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=77aad734-2057-5694-9ae2-ee1f5f26eae8&l=252997

thewtex:
Merged into `release` and `master`.

N-Dekker:
@aylward Hi Stephen, because of the holidays (heatwave) I did not have time to ask you for feedback on this issue. I'm very happy to see that Matt merged my PR so quickly already, to avoid those segfaults. But as you are very much involved with spatial objects, is it also OK to you that `ImageMaskSpatialObject::IsInsideInObjectSpace` no longer makes use of its interpolator?

This PR is certainly OK to me, I just want to be sure it's also fine to you!

aylward:
@N-Dekker - Hi Neal,  This looks good to me!  Thanks for checking, and thanks for the fix!   I do see arguments either way regarding the use of an interpolator - so not using an interpolator seems reasonable and it would clearly be faster.    I appreciate the change!

PR #1953, opened 2020-08-10
BUG: Fix issue #1939: Use -mtune=generic

### LINKED ISSUE #1939: build results vary from build machine CPU
# Description

While working on reproducible builds for openSUSE, I found that
our insighttoolkit package varied from the type of build machine CPU
that randomly gets chosen from the pool of build workers.

This is because `InsightToolkit-5.1.0/CMake/ITKSetStandardCompilerFlags.cmake` has `-mtune=native`
and [dropping this helped](https://build.opensuse.org/request/show/823615) to make builds reproducible.

See https://reproducible-builds.org/ for why this matters.

### Steps to Reproduce

1. build ITK on 2 hosts with different CPU types
2. compare binaries

### Expected behavior

It should be possible to get bit-identical build results on different build machines.
May be triggered via a cmake flag or the [`SOURCE_DATE_EPOCH` environment variable](https://reproducible-builds.org/specs/source-date-epoch/)

### Actual behavior

build results vary from build machine CPU

### Reproducibility

100%

### Versions

5.1

### Environment

openSUSE-Tumbleweed-20200728

### Additional Information

This PR was done while working on reproducible builds for openSUSE.

bmwiedemann:
Use `-mtune=generic` to avoid build-time CPU detection of `-mtune=native`
to get reproducible build results.

See https://reproducible-builds.org/ for why this is good.

Fixes #1939

This PR was done while working on reproducible builds for openSUSE.

## PR Checklist

### DIFF
--- a/CMake/ITKSetStandardCompilerFlags.cmake
+++ b/CMake/ITKSetStandardCompilerFlags.cmake
@@ -241,7 +241,8 @@ function(check_compiler_optimization_flags c_optimization_flags_var cxx_optimiza
     set(InstructionSetOptimizationFlags
        # https://gcc.gnu.org/onlinedocs/gcc-4.8.0/gcc/i386-and-x86_002d64-Options.html
        # NOTE the corei7 release date was 2008
-       -mtune=native # Tune the code for the computer used compile ITK, but allow running on generic cpu archetectures
+       #-mtune=native # Tune the code for the computer used compile ITK, but allow running on generic cpu archetectures
+       -mtune=generic # for reproducible results https://github.com/InsightSoftwareConsortium/ITK/issues/1939
        -march=corei7 # Use ABI settings to support corei7 (circa 2008 ABI feature sets, core-avx circa 2013)
        )
   endif()


bmwiedemann:
@dzenanz Performance would depend on which build machine did the mtune=native build.
I probably wont be able to run the benchmarking (no time).

hjmjohnson:
My concern is that the default builds with the -mtune=generic only use instructions from CPU's that are 20 years old.    There have been many advances since then, and ITK is well suited to take advantage of modern hardware instructions.

bmwiedemann:
@hjmjohnson the allowed CPU instructions are not governed by `mtune=`, but by the `march=corei7` below that expects CPUs younger than 12y. IMHO a good compromise if you do not want to ship multiple copies for different target CPUs.

hjmjohnson:
> @hjmjohnson the allowed CPU instructions are not governed by `mtune=`, but by the `march=corei7` below that expects CPUs younger than 12y. IMHO a good compromise if you do not want to ship multiple copies for different target CPUs.

I agree regarding the instructions available being specified by the '-march' flag, and the '-mtune' flag prefers using instructions for the specified architecture.  

I don't understand why one would "allow use of corei7 instructions" but tune for the "generic lowest common denominator instructions"?  This would potentially generate binaries that require corei7 features, but then seldom use them.

thewtex:
Checking the [GCC docs](https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html):

```
-mtune=cpu-type

    Tune to cpu-type everything applicable about the generated code, except for the ABI and the set of available instructions. While picking a specific cpu-type schedules things appropriately for that particular chip, the compiler does not generate any code that cannot run on the default machine type unless you use a -march=cpu-type option. For example, if GCC is configured for i686-pc-linux-gnu then -mtune=pentium4 generates code that is tuned for Pentium 4 but still runs on i686 machines.

    The choices for cpu-type are the same as for -march. In addition, -mtune supports 2 extra choices for cpu-type:

    ‘generic’

        Produce code optimized for the most common IA32/AMD64/EM64T processors. If you know the CPU on which your code will run, then you should use the corresponding -mtune or -march option instead of -mtune=generic. But, if you do not know exactly what CPU users of your application will have, then you should use this option.

        As new processors are deployed in the marketplace, the behavior of this option will change. Therefore, if you upgrade to a newer version of GCC, code generation controlled by this option will change to reflect the processors that are most common at the time that version of GCC is released.

        There is no -march=generic option because -march indicates the instruction set the compiler can use, and there is no generic instruction set applicable to all processors. In contrast, -mtune indicates the processor (or, in this case, collection of processors) for which the code is optimized.

```

Given:

> -mtune=generic. But, if you do not know exactly what CPU users of your application will have, then you should use this option.

> Therefore, if you upgrade to a newer version of GCC, code generation controlled by this option will change to reflect the processors that are most common at the time that version of GCC is released.

`-mtune=generic` is the best default option for working across generic modern processors (which may be even newer than the instructions allowed by `-march`). @hjmjohnson do you agree?

hjmjohnson:
@thewtex 

mtune and march are somewhat orthogonal. march=cpu-type generate instructions for the machine type cpu-type. In contrast to -mtune=cpu-type, which **merely tunes the generated code for the specified cpu-type**, -march=cpu-type allows GCC to generate code that may not run at all on processors other than the one indicated. Specifying -march=cpu-type implies -mtune=cpu-type, except where noted otherwise.

-mtune=generic is a fine heuristic that makes reasonable tuning of the code (order of instructions, prefetch assumptions, etc) within the instructions chosen by "march" flag (which like defaults to -march=x86-64, A generic CPU with 64-bit extensions).  

march indicates the instruction set the compiler can use.  If march is undefined (the default) then MAXIMUM GENERALIZABILITY is achieved by MINIMIZING the CPU instructions used (i.e. prohibit use of MMX, SSE, SSE2, SSE3, SSSE3, SSE4.1, SSE4.2, POPCNT, AES and PCLMUL instruction set support.).

========
I think that we could easily drop the -mtune flag, and use the default behavior.  
The rationale for having -march=corei7 (circa 2008 instructions) is that 99% of the anticipated deployment market will likely be satisfied with these instructions, and basic simd instructions are allowed to be used for processing.  

‘corei7’
Intel Core i7 CPU with 64-bit extensions, MMX, SSE, SSE2, SSE3, SSSE3, SSE4.1 and SSE4.2 instruction set support.

Hans

CMake/ITKSetStandardCompilerFlags.cmake
```diff
@@ -241,7 +241,8 @@ function(check_compiler_optimization_flags c_optimization_flags_var cxx_optimiza
     set(InstructionSetOptimizationFlags
        # https://gcc.gnu.org/onlinedocs/gcc-4.8.0/gcc/i386-and-x86_002d64-Options.html
        # NOTE the corei7 release date was 2008
-       -mtune=native # Tune the code for the computer used compile ITK, but allow running on generic cpu archetectures
+       #-mtune=native # Tune the code for the computer used compile ITK, but allow running on generic cpu archetectures
+       -mtune=generic # for reproducible results https://github.com/InsightSoftwareConsortium/ITK/issues/1939
```
hjmjohnson, line 245: NOTE: -mtune=generic is fine.  we could also use the equivalent  -mtune=corei7, or equivalently remove this falg all together.

When march=cpu-type  is specified mtune is implied as -mtune=cpu-type as the generic behavior.

The original thought was to "tune" the instruction scheduling for the build host, but restrict the instructions used to 2008 instruction set.

hjmjohnson (approved):

hjmjohnson:
FYI: 
```
FLAGS1="-march=corei7 -mtune=generic"
gcc $(echo ${FLAGS1})  -Q --help=target > /tmp/options1

FLAGS2=""
gcc $(echo ${FLAGS2})  -Q --help=target > /tmp/options2

diff /tmp/options1 /tmp/options2

24c24
<   -march=                     		corei7
---
>   -march=                     		x86-64
53c53
<   -mcx16                      		[enabled]
---
>   -mcx16                      		[disabled]
100c100
<   -mno-sse4                   		[disabled]
---
>   -mno-sse4                   		[enabled]
109c109
<   -mpopcnt                    		[enabled]
---
>   -mpopcnt                    		[disabled]
125c125
<   -msahf                      		[enabled]
---
>   -msahf                      		[disabled]
133,136c133,136
<   -msse3                      		[enabled]
<   -msse4                      		[enabled]
<   -msse4.1                    		[enabled]
<   -msse4.2                    		[enabled]
---
>   -msse3                      		[disabled]
>   -msse4                      		[disabled]
>   -msse4.1                    		[disabled]
>   -msse4.2                    		[disabled]
140c140
<   -mssse3                     		[enabled]
---
>   -mssse3                     		[disabled]

```

CMake/ITKSetStandardCompilerFlags.cmake
thewtex: > When march=cpu-type is specified mtune is implied as -mtune=cpu-type as the generic behavior.

Yes, agreed, this also my understanding.

However, my understanding from the docs and verifying in the source code:

https://github.com/gcc-mirror/gcc/blob/edd716b6b1caa1a5cb320a8cd7f626f30198e098/gcc/config/i386/i386-c.c#L301

is that `-mtune=corei7` (which could happen by setting `-march=corei7`, or by setting `-mtune=native` when the build is on a `corei7` processor, or by setting `-mtune=corei7`), will tune for `corei7`, and that tuning will be different than `-mtune=generic`.

thewtex (approved):
@hjmjohnson yes, it was a major progression to add support for the newer instructions via 'march=corei7' :+1: :+1: 

It looks like we are all in agreement that `-march=corei7` gives us the instructions we want and `-mtune=generic` gives the tuning we want.

Thanks all for the time and discussion on this, merging towards faster, reproducible insights! :rocket: :boom:

Issue #1971, opened 2020-08-29
BUG: hdf5 missing version


thkvc:
### Description
After building latest code on github, all binaries with the suffix "-5.2" before ".so" and ".dll" except hdf5 don't have them.
Tested on Ubuntu and Windows
This doesn't happen in vtk

Note:
I build shared library and the result of hdf5 is:
ITK:
      - Windows: itkhdf5.dll, itkhdf5_cpp.dll
      - Ubuntu: libitkhdf5.so, libitkhdf5_cpp.so
VTK:
      - Windows: vtkhdf5-9.0.dll, vtkhdf5_hl-9.0.dll
      - Ubuntu: libvtkhdf5-9.0.so, libvtkhdf5_hl-9.0.so

dzenanz:
In addition to missing `-5.2` suffix, it also has `lib` prefix, so the file is named `libitkhdf5.lib` instead of `itkhdf5-5.2.lib`.

dzenanz:
@thewtex where does the version suffix come from? I looked at double conversion for comparison and could not spot significant difference.

thewtex:
A call to `itk_module_target_name` in the HDF5 module CMake configuration may help:

https://github.com/InsightSoftwareConsortium/ITK/blob/7f36e8e2ff52f402a4c9377fbdc68c3ba018e6c9/CMake/ITKModuleMacros.cmake#L393-L411

dzenanz:
@thkvc can you give #2229 a try?

PR #1974, opened 2020-08-31
ENH: append properties instead of overwrite in HDF5

dzenanz:
This does not make a difference wrt #1971 on my computer.


hjmjohnson (approved):

hjmjohnson:
This seems reasonable, but a bit more explanation would be helpful.

hjmjohnson:
@dzenanz Do you still think this should be merged?

dzenanz:
I would like @thewtex to review it, but if he doesn't in the next few days then let's merge.

PR #1998, opened 2020-09-08
ENH: Implement functionality to extract index in a bounded region

spyridon97:
**TransformPhysicalPointToIndex** & **TransformPhysicalPointToContinuousIndex** are 2 sets of functions that convert a physical point to an index with the usage of rounding. Geometrically speaking, if a point is outside the region of an image by at max spacing/2 across every dimension, the point is going the get an index that can exist inside the image. 

I developed **TransformPhysicalPointToIndexInBoundedRegion** & **TransformPhysicalPointToContinuousIndexInBoundedRegion** that given a point and a region of the image, performs the same operation, but if a point is geometrically speaking outside of the given region, then an *out of bounds* index is going to be returned.

If the region is the largest possible region, then it can understand if a point is geometrically speaking outside of the image.
So in simple words, it is strict at the boundaries.

For more information see here: [ITK discussion](https://discourse.itk.org/t/transformphysicalpointtoindex-that-remains-inside-the-boundaries-of-the-image/3435)

### DIFF
--- a/Modules/Core/Common/include/itkImageBase.h
+++ b/Modules/Core/Common/include/itkImageBase.h
@@ -87,8 +87,10 @@ namespace itk
  * frames. TransformIndexToPhysicalPoint() converts an Index in the
  * pixel array into its coordinates in physical space.
  * TransformPhysicalPointToIndex() converts a position in physical
- * space into an Index into the pixel array (using
- * rounding). Subpixel locations are supported by methods that
+ * space into an Index into the pixel array (using rounding).
+ * TransformPhysicalPointToIndexInBoundedRegion() converts a position in physical
+ * space into an Index of a given bounded region into the pixel array (using rounding only
+ * for interior points). Subpixel locations are supported by methods that
  * convert to and from ContinuousIndex types.
  *
  * ImageBase also provides helper routines for the ImageIterators
@@ -473,7 +475,6 @@ public:
     return isInside;
   }
 
-
   /** \brief Returns the continuous index from a physical point
    * \note This specific overload does not figure out whether or not
    *  the returned index is inside the image. Of course, the user can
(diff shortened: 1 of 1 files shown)


blowekamp:
Nice work! 

Please look at SimpleITK's fundamental concepts of images to understand the details of the image geometry I'm discussing below: https://simpleitk.readthedocs.io/en/master/fundamentalConcepts.html

Under the ITK's Image indexing concepts the continuous index of -0.49 is considered with the image's region boundary.

The index represents the *center* of each voxel. So each voxel has an extent of 1, and +/- 0.5 from either side. Therefore for a 1-D image with a region of [0,10], the valid continuous index range is [-0.5, 9.5]. I believe
 your implementation needs a correction to account for the correct behavior on the 0.5 boulder.

Additionally, testing is required so we can verify the correctness of the methods and not regression of behavior occur in the future. For this level of functionality, I would suggest looking at the GTest style tests as an example.

spyridon97:
If understand correctly your example, let's assume we have a 1D image with origin = {0,0}, spacing = {1} and size = {11} (since it is [0, 10]). In this case, the valid continuous index is [-0.5, 10.5], but with my additionally bounded methods, i restrict to [0, 10] cause anything out of that is geometrically speaking outside of the image.

blowekamp:
That is a very good illustration of this behavior or issue.

> For example, i have attached an image with origin {0, 0, 0}, spacing = {0.1, 0.6931469, 0.2}, and size = {41, 6 , 21}. If you preview it on ParaView, you will see that the image is not going below {0, 0, 0} geometrically speaking.
> [data.zip](https://github.com/InsightSoftwareConsortium/ITK/files/5195924/data.zip)

3D Slicer and ITK Snap correctly use the ITK standard of origin at the center of the voxel and then volume extent going down to the {-0.5} continuous index. These applications are designed for medical image analysis and have focus on very precise geometric consistency of volumes and data.


> So in my methods, I allow the interior pixels to extend by spacing[i]/2 to every dimension as in theory, but when it comes to the boundary ones, I don't allow that to happen, because of geometric reasons.
> 
> Yeah, I know that this creates conflict with the fundamental definition of the image, but I believe it's useful when you want to represent such bounded/strict domains (grids), as images.

I can understand how this particular implementation could be useful. But the inconsistency with the rest of ITK is going to be confusing and problematic if this method is added to ITK itself.

For example it is not consistent with the `bool ImageRegion::IsInside(const ContinuousIndex<...> & ) const` method. Introducing another "inside" or "out of bounds" concept would be a bad thing.

blowekamp (changes requested):
This PR introduces a new "out of bounds" concept that is not consistent with the current concept of an ITK Image and ITK ImageRegion.

dzenanz:
ParaView is based on VTK, and VTK's current convention on images is that pixel's origin is in it corner. This is different from medical imaging standard (essentially set by DICOM) where pixel's origin is in its center. ITK, ITK-SNAP and Slicer follow this medical imaging convention. So if you are dealing with medical images, ParaView is not the best choice of viewer.

PR #2000, opened 2020-09-11
COMP: variable is uninitialized when passed as a const reference

dzenanz:
Trying to fix the [warning](https://open.cdash.org/viewBuildError.php?type=1&buildid=6756465):
```text
In file included from /Users/builder/externalModules/Core/ImageFunction/test/itkSumOfSquaresImageFunctionGTest.cxx:22:
In file included from /Users/builder/externalModules/Core/Common/include/itkImage.h:346:
[CTest: warning matched] /Users/builder/externalModules/Core/Common/include/itkImage.hxx:152:46: warning: variable 'p' is uninitialized when passed as a const reference argument here [-Wuninitialized-const-reference]
  return NumericTraits<PixelType>::GetLength(p);
                                             ^
[CTest: warning matched] /Users/builder/externalModules/Core/Common/include/itkImage.h:322:3: note: in instantiation of member function 'itk::Image<double, 2>::GetNumberOfComponentsPerPixel' requested here
  Image();
  ^
[CTest: warning matched] /Users/builder/externalModules/Core/Common/include/itkImage.h:99:15: note: in instantiation of member function 'itk::Image<double, 2>::Image' requested here
  itkNewMacro(Self);
              ^
[CTest: warning matched] /Users/builder/externalModules/Core/ImageFunction/test/itkSumOfSquaresImageFunctionGTest.cxx:57:30: note: in instantiation of member function 'itk::Image<double, 2>::New' requested here
  const auto image = TImage::New();
                             ^
[CTest: warning matched] /Users/builder/externalModules/Core/ImageFunction/test/itkSumOfSquaresImageFunctionGTest.cxx:101:3: note: in instantiation of function template specialization '(anonymous namespace)::Expect_EvaluateAtIndex_returns_zero_when_all_pixels_are_zero<itk::Image<double, 2>>' requested here
  Expect_EvaluateAtIndex_returns_zero_when_all_pixels_are_zero<itk::Image<double, 2>>(itk::Size<2>{ { 2, 3 } });
  ^
```

### DIFF
--- a/Modules/Core/Common/include/itkImage.hxx
+++ b/Modules/Core/Common/include/itkImage.hxx
@@ -148,7 +148,7 @@ Image<TPixel, VImageDimension>::GetNumberOfComponentsPerPixel() const
 {
   // use the GetLength() method which works with variable length arrays,
   // to make it work with as much pixel types as possible
-  PixelType p;
+  const PixelType p{};
   return NumericTraits<PixelType>::GetLength(p);
 }
 


seanm:
Looks reasonable.  Do you need me to verify that it actually fixes it?

dzenanz:
@seanm yes please. Note: I just foce-pushed an update.

blowekamp:
I was just looking at this warning and concluded it was bogus IMHO!

The value p is not used by the method. So it doesn't matter that it is not initialized. To address the warning unnecessary work is being added.

This type of pattern is used in many places to dispatch to the correct overload method/procedure ( scalar vs vector pixel type), so my concern is not just this specific instance but this is not a good pattern for correction.

seanm:
@dzenanz ok I'm building now...

@blowekamp is this hot code? If not, we may as well pacify the compiler, even if we do consider it a false positive.  I'll know soon, but I recall there were not many of these warnings, this is perhaps the only one IIRC.

N-Dekker:
In this case, will it just return zero when `PixelType` is a variable length array?

    PixelType p;
    return NumericTraits<PixelType>::GetLength(p);

seanm:
So in fact with that change I get a link error:

```
  "itk::NumericTraits<itk::Index<2u> >::Zero", referenced from:
      itk::NumericTraits<itk::Index<2u> >::ZeroValue() in itkImageRandomNonRepeatingIteratorWithIndexTest2.cxx.o
  "itk::NumericTraits<itk::Index<3u> >::Zero", referenced from:
      itk::NumericTraits<itk::Index<3u> >::ZeroValue() in itkImageLinearIteratorTest.cxx.o
      itk::NumericTraits<itk::Index<3u> >::ZeroValue() in itkImageRandomIteratorTest.cxx.o
      itk::NumericTraits<itk::Index<3u> >::ZeroValue() in itkImageRandomNonRepeatingIteratorWithIndexTest.cxx.o
      itk::NumericTraits<itk::Index<3u> >::ZeroValue() in itkImageSliceIteratorTest.cxx.o
  "itk::NumericTraits<itk::Matrix<float, 3u, 3u> >::Zero", referenced from:
      itk::NumericTraits<itk::Matrix<float, 3u, 3u> >::ZeroValue() in itkSymmetricSecondRankTensorImageReadTest.cxx.o
  "itk::NumericTraits<itk::NodeClass<itk::Image<int, 2u> >*>::Zero", referenced from:
      itk::NumericTraits<itk::NodeClass<itk::Image<int, 2u> >*>::ZeroValue() in itkSparseImageTest.cxx.o
ld: symbol(s) not found for architecture x86_64
clang-12: error: linker command failed with exit code 1 (use -v to see invocation)
```

N-Dekker:
> So in fact with that change I get a link error:

@seanm Thanks for figuring out 👍 Would it compile when `p` is initialized by an empty pair of curly braces instead?

    PixelType p { } ;  // C++11 empty-list-initialization
    return NumericTraits<PixelType>::GetLength(p);

seanm:
@N-Dekker yes, that does it!

dzenanz:
The reason this did not compile on any platform initially was that I was modifying one repository and compiling another. Visual Studio does not have this warning, so I could not tell whether it was working or not 😄

jhlegarreta (approved):
Nice !

dzenanz:
Warning count went down from 199 (is that when compiler stops reporting any more?) [down to 31](https://open.cdash.org/viewBuildError.php?type=1&buildid=6759909). Some work still could be done. @seanm could you try addressing them?

N-Dekker:
I see, Ubuntu-gcc4.6 says:
> warning: missing initializer for member 'itk::Index<3u>::m_InternalArray' [-Wmissing-field-initializers]
   const PixelType p{};

It's really a compiler bug, to produce that 'missing-field-initializers' warning on such a proper initialization, by an empty initializer list `{}`. And it has been fixed with GCC 5. (GCC is now at version 10.2) Maybe it is worth considering to drop GCC4 support?

Anyway, the following syntax might make the old GCC 4 happier:

    const auto p = PixelType();

blowekamp:
> Maybe it is worth considering to drop GCC4 support?

I don't think it is possible to drop the support for the compiler. However, we could limit which compilers we which to aggressively use to detect and address warning on.

N-Dekker:
Note that the warning about a missing initializer for a member isn't always bogus:

    struct Point2D { int x; int y; };
    Point2D point{42}; // <== Warn me please!
    Point2D defaultPoint{}; // <== Please do not warn!

GCC 5.X and later only produce a warning on the first declaration, `point{42}`, which makes sense to me: https://godbolt.org/z/vzzr97

PR #2012, opened 2020-09-23
STYLE: Allow conversion from nullptr to WeakPointer, not from zero

N-Dekker:
Small WeakPointer improvements:

- Defaulted default-constructor (allowing the compiler to consider it
"constexpr" and "noexcept")
- Removed the other user-defined special functions (Rule Of Zero)
- Removed redundant user-defined `operator=(ObjectType*)`
- Added converting constructor, WeakPointer(std::nullptr_t).

No longer allowed implicit conversion from the integer zero.

Included GoogleTest unit tests.

### DIFF
--- a/Modules/Core/Common/include/itkWeakPointer.h
+++ b/Modules/Core/Common/include/itkWeakPointer.h
@@ -47,22 +47,20 @@ public:
   /** Extract information from template parameter. */
   using ObjectType = TObjectType;
 
-  /** Constructor.  */
-  WeakPointer() { m_Pointer = nullptr; }
+  /** Explicitly-defaulted default-constructor.
+   * \note The other five "special member functions" (copy-constructor,
+   * copy-assignment operator, move-constructor, move-assignment operator,
+   * and destructor) are defaulted implicitly, following the C++ "Rule of Zero".
+   */
+  WeakPointer() = default;
 
-  /** Copy constructor.  */
-  WeakPointer(const WeakPointer<ObjectType> & p) = default;
-
-  /** Move constructor */
-  WeakPointer(WeakPointer<ObjectType> && p) = default;
+  /** Constructor, converting from `nullptr`.  */
+  WeakPointer(std::nullptr_t) {}
 
   /** Constructor to pointer p.  */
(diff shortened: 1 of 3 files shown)


hjmjohnson (approved):
Thank you for the exceptional expanded tests.

If no other issues arise during CI testing, I am in favor of this improvement.

dzenanz (approved):

dzenanz:
> No longer allowed implicit conversion from the integer zero.

Where does this feature come from?

Modules/Core/Common/include/itkWeakPointer.h
```diff
@@ -47,23 +47,21 @@ class WeakPointer
   /** Extract information from template parameter. */
   using ObjectType = TObjectType;
 
-  /** Constructor.  */
-  WeakPointer() { m_Pointer = nullptr; }
+  /** Explicitly-defaulted default-constructor.
+   * \note The other five "special member functions" (copy-constructor,
+   * copy-assignment operator, move-constructor, move-assignment operator,
+   * and destructor) are defaulted implicitly, following the C++ "Rule of Zero".
```
N-Dekker, line 53: Thanks for your approval, @hjmjohnson Do you think it's still useful to explain what is meant by "special member functions", like I'm doing here? I meant, is it still useful to mention those five member functions explicitly here?

    copy-constructor, copy-assignment operator, move-constructor, move-assignment operator, and destructor

I feel that the term "special member functions" is a bit "special" (possibly not familiar to everyone), that's why I did. But what do you think?

N-Dekker:
Thanks for your approval, @dzenanz 
> > No longer allowed implicit conversion from the integer zero.
> 
> Where does this feature come from?

With this PR, `WeakPointer` would have a converting constructor from `nullptr_t`, as well as the existing one from `ObjectType *`. This would make implicit conversion from the integer zero ambiguous. 

Without this PR, conversion from the integer zero would be performed by the existing converting constructor, from `ObjectType *`.

N-Dekker:
OK, I think the PR looks fine, but I leave it up to any of you to press Rebase-and-merge   :-)

PR #2013, opened 2020-09-23
STYLE: Add in-class default member initializer to SmartPointer m_Pointer

N-Dekker:
Defaulted the default-constructor of `SmartPointer`. Slightly simplified
the implementation of `SmartPointer(std::nullptr_t)` and its destructor.


hjmjohnson (approved):

N-Dekker:
This pull request is somewhat similar to the previous one (for `WeakPointer`): https://github.com/InsightSoftwareConsortium/ITK/pull/2012/  However, this pull request does not address assignment to zero:

    itk::SmartPointer<itk::Object> ptr;
    ptr = 0; // Assigning an int to the smart pointer!

PR #2020, opened 2020-09-27
STYLE: Turn commented unnecessary commands into useful comments

jhlegarreta:
Turn commented unnecessary commands into useful comments.

## PR Checklist


dzenanz (approved):
This is good as is. These classes are stable now and should not need much debugging, if any.

But in a similar situation we could use the pattern:
```cpp
if (this->GetDebug())
{
  std::cout << "BasicErodeImageFilter" << std::endl;
}
```

jhlegarreta:
Good to know. Maybe worth adding to the SW Guide ? Thanks @dzenanz !

PR #2024, opened 2020-09-29
BUG: Filter inputs are erroneously released

### LINKED ISSUE #2023: Filter inputs are released despite : DataObject::SetGlobalReleaseDataFlag() set to false
Since I am not a registered itk developer I give the fix that worked for me here

### Description

The filters that rely on DataObject class, eg. itk::MaskImageFilter corrupt the input image (clears it) despite that both global and filter relative setting to release data is set to false. The error was traced and the bug found here:

Code error in the \Modules\Core\Common\src\itkDataObject.cxx . The function : 
```cpp
bool
DataObject ::ShouldIReleaseData() const
{
  return (m_GlobalReleaseDataFlag || m_ReleaseDataFlag);
}
```
contais a bug. The m_GlobalReleaseDataFlag is a pointer and not the actual globalreleasedataflag value.

This corection solves it for me:
```cpp
bool
DataObject ::ShouldIReleaseData() const
{
  return (GetGlobalReleaseDataFlag() || m_ReleaseDataFlag);
}
```
### Steps to Reproduce

call: itk::DataObject::SetGlobalReleaseDataFlag(false);
Create a 3D image of 16bit values, sized 512x98x512. Fill it with zeros.
Create a 3D binary image to represent a mask. Fill it with zeros.

Apply the itk::MaskImageFilter with all default params, passing in the original image and the mask image.

### Expected behavior

Mask filter output is ok, the input image is preserved unchanged.

### Actual behavior

The output is produced ok but the original input image is empty!
It should not be cleared, but kept untouched.

### Reproducibility

Always and also for other filters that I use

### Versions

Official ITK 5.1.0 and ITK 5.1.1 releases

### Environment

OS win10 64bit, cmake 3.18.1, Visual studio 2019, no python usage

dzenanz:
The filters that rely on DataObject class, eg. itk::MaskImageFilter corrupt the input image (clears it) despite that both global and filter relative setting to release data is set to false.

Submitted by @urosvovk1. Closes #2023.

### DIFF
--- a/Modules/Core/Common/src/itkDataObject.cxx
+++ b/Modules/Core/Common/src/itkDataObject.cxx
@@ -175,7 +175,7 @@ DataObject ::ReleaseData()
 bool
 DataObject ::ShouldIReleaseData() const
 {
-  return (m_GlobalReleaseDataFlag || m_ReleaseDataFlag);
+  return (GetGlobalReleaseDataFlag() || m_ReleaseDataFlag);
 }
 
 //----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


dzenanz:
Warnings from CircleCI are unrelated:
`Examples/Iterators/ImageRandomConstIteratorWithIndex.cxx:137:3: warning: multi-line comment [-Wcomment]`

blowekamp (approved):
Nicely done!

PR #2031, opened 2020-10-04
COMP: Fix missing initialization braces warnings

jhlegarreta:
Fix missing initialization braces warnings.

Fixes:
```
[CTest: warning matched]
/Users/builder/externalModules/Filtering/ImageGrid/test/itkPasteImageFilterGTest.cxx:121:33:
warning: suggest braces around initialization of subobject
[-Wmissing-braces]
  filter->SetDestinationIndex({ 11, 13 });
                                ^~~~~~
				{     }
```

and warnings of the same kind reported at:
https://open.cdash.org/viewBuildError.php?type=1&buildid=6799229

## PR Checklist

### DIFF
--- a/Modules/Filtering/ImageGrid/test/itkPasteImageFilterGTest.cxx
+++ b/Modules/Filtering/ImageGrid/test/itkPasteImageFilterGTest.cxx
@@ -118,21 +118,21 @@ TEST_F(PasteFixture, ConstantPaste)
   auto inputImage = Utils::CreateImage(100);
 
   filter->SetDestinationImage(inputImage);
-  filter->SetDestinationIndex({ 11, 13 });
+  filter->SetDestinationIndex({ { 11, 13 } });
 
   constexpr int constantValue = -97;
   filter->SetConstant(constantValue);
-  filter->SetSourceRegion(Utils::SizeType{ 3, 3 });
+  filter->SetSourceRegion(Utils::SizeType{ { 3, 3 } });
 
   filter->SetNumberOfWorkUnits(1);
   filter->UpdateLargestPossibleRegion();
 
   auto outputImage = filter->GetOutput();
-  EXPECT_EQ(0, outputImage->GetPixel({ 10, 13 }));
-  EXPECT_EQ(0, outputImage->GetPixel({ 11, 12 }));
+  EXPECT_EQ(0, outputImage->GetPixel({ { 10, 13 } }));
+  EXPECT_EQ(0, outputImage->GetPixel({ { 11, 12 } }));
 
-  EXPECT_EQ(constantValue, outputImage->GetPixel({ 11, 13 }));
-  EXPECT_EQ(constantValue, outputImage->GetPixel({ 13, 15 }));
+  EXPECT_EQ(constantValue, outputImage->GetPixel({ { 11, 13 } }));
+  EXPECT_EQ(constantValue, outputImage->GetPixel({ { 13, 15 } }));
 
   EXPECT_EQ("e73092e3e58f66f13d32a5aca97ca1d9", MD5Hash(filter->GetOutput()));
 
@@ -141,22 +141,22 @@ TEST_F(PasteFixture, ConstantPaste)
   EXPECT_EQ("e73092e3e58f66f13d32a5aca97ca1d9", MD5Hash(filter->GetOutput()));
 
 
-  filter->SetSourceRegion(Utils::SizeType{ 2, 1 });
+  filter->SetSourceRegion(Utils::SizeType{ { 2, 1 } });
   filter->UpdateLargestPossibleRegion();
   EXPECT_EQ("47e4689dc4a86bc6da9eeb9ce0326fa7", MD5Hash(filter->GetOutput()));
 
 
-  filter->SetSourceRegion(Utils::SizeType{ 1, 5 });
+  filter->SetSourceRegion(Utils::SizeType{ { 1, 5 } });
   filter->UpdateLargestPossibleRegion();
   EXPECT_EQ("2bd084a4aa4696409b6019438ef1313f", MD5Hash(filter->GetOutput()));
 
 
-  filter->SetSourceRegion(Utils::SizeType{ 1, 3 });
+  filter->SetSourceRegion(Utils::SizeType{ { 1, 3 } });
   filter->UpdateLargestPossibleRegion();
   EXPECT_EQ("07f0acc6e11ac92b5ea5dfd8ae3f9209", MD5Hash(filter->GetOutput()));
 
 
-  filter->SetSourceRegion(Utils::SizeType{ 1, 1 });
+  filter->SetSourceRegion(Utils::SizeType{ { 1, 1 } });
   filter->UpdateLargestPossibleRegion();
   EXPECT_EQ("2a9bfc9d4548a986f52fb3ba2ae07de2", MD5Hash(filter->GetOutput()));
 }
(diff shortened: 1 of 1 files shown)


jhlegarreta:
So not sure how to solve the conundrum: the brace suggestion:
```
[CTest: warning matched] /Users/builder/externalModules/Filtering/ImageGrid/test/itkPasteImageFilterGTest.cxx:300:45: warning: suggest braces around initialization of subobject [-Wmissing-braces]
  filter->SetDestinationSkipAxes(SkipType({ true, false, false }));
                                            ^~~~~~~~~~~~~~~~~~
                                            {                 }
```
in here:
https://open.cdash.org/viewBuildError.php?type=1&buildid=6799229
seems to cause an ambiguous overload call to `itk::FixedArray`:
https://open.cdash.org/viewBuildError.php?buildid=6799410

@blowekamp @N-Dekker @seanm any idea?

blowekamp:
@jhlegarreta Thank you for trying to address it. I believe previously this type if initialization was resolved with per element assignment, which I find ridiculous that we can't do this kind of universal initialization/list/aggregate initialization for ITK array type classes.

N-Dekker:
@jhlegarreta What about adding the suggested extra pair of curly braces, but then removing the parentheses?

    filter->SetDestinationSkipAxes(SkipType{ { true, false, false } });

Compiles locally on my VS2019. Does that fix the issue?

seanm:
As I said in the email thread, I'd be happy to remove the warning, and indeed I thought I had, as `-Wno-missing-braces` is listed here:

https://open.cdash.org/build/6800558/notes

Any idea why that's not stopping the warnings?

jhlegarreta:
@N-Dekker I push forced your proposal, but it fails with VS 2015 with the following errors:
```
Error C2127 'CreateSourceImage':
illegal initialization of 'constexpr' entity with a non-constant expression 
ITKImageGridGTestDriver C:\SDKs\ITK\itk-head\ITK\Modules\Filtering\ImageGrid\test\itkPasteImageFilterGTest.cxx 93
Error C2127 'CreateImage':
illegal initialization of 'constexpr' entity with a non-constant expression ITKImageGridGTestDriver
C:\SDKs\ITK\itk-head\ITK\Modules\Filtering\ImageGrid\test\itkPasteImageFilterGTest.cxx 92
```

which were also present for me with the first solution. Removing the `constexpr` specifiers generates further errors for me, so no idea here.

@seanm I ignore why the flag is not being properly applied.

@blowekamp I agree that a better solution would be desirable, but I cannot investigate further than that. Following what @N-Dekker pointed, his proposal already works for `Utils::SizeType` types, so it looks like this goes deep into the contents of `itk::FixedArray`.

N-Dekker:
@jhlegarreta @blowekamp It seems to me that the converting constructor that converts `FixedArray` from one `ValueType` to another caused the ambiguity. Would it be helpful to add an extra "dummy" parameter to this constructor, to avoid such ambiguity? For example:

    enum class ValueTypeConversionTag // The "dummy" parameter type.
    {
      value
    };
 
    /** Dummy parameter indicates that ValueType conversion is performed */
    template <typename TFixedArrayValueType>
    FixedArray(const FixedArray<TFixedArrayValueType, VLength> & r, ValueTypeConversionTag)
    {
        Code here as before...
    }

Every call to this constructor should then have `ValueTypeConversionTag::value` as extra argument. This would prevent accidental lossy conversion, as well as ambiguity. However, it would break old user code.... what do you think?

Leengit:
`IndexValueType` is a `signed long` / `int64_t` according to `Modules/Core/Common/include/itkIntTypes.h`.  Does it help to change `filter->SetDestinationIndex({ 11, 13 });` to `filter->SetDestinationIndex({ 11L, 13L });`?

Leengit:
... or to `filter->SetDestinationIndex( { { 11L, 13L } } );` under the thinking that this last form first creates an array of signed long and then uses it to initialize the first (and only) member of `itk::Index`.

jhlegarreta:
@N-Dekker not sure if the solution you propose is the most appropriate one.

@Leengit I hope the double braces makes the compiler happy about the warnings on the `IndexValueType`s; the CI builds were failing on the boolean type arrays.

However, the [previous solution](https://github.com/InsightSoftwareConsortium/ITK/pull/2031#issuecomment-703642633) proposed by @N-Dekker seems to have workd for the CI builds. Now @seanm can you please checkout the branch and see if that solution satisfies your compilers as well? Thanks !

dzenanz (approved):
Is this ready for merging?

jhlegarreta:
Pinged @seanm : will try it locally and will let us know if the build warnings go away with the proposed solution :+1:.

seanm:
Hi folks. I've been on vacation and just back in the office today. This patch seems to fix most of the warnings, but these remain:

```
/Users/builder/external/ITK/Modules/Filtering/ImageGrid/test/itkPasteImageFilterGTest.cxx:183:46: warning: suggest braces around initialization of subobject [-Wmissing-braces]
  filter->SetDestinationSkipAxes(SkipType{ { true, false, false } });
                                             ^~~~~~~~~~~~~~~~~~
                                             {                 }
/Users/builder/external/ITK/Modules/Filtering/ImageGrid/test/itkPasteImageFilterGTest.cxx:192:46: warning: suggest braces around initialization of subobject [-Wmissing-braces]
  filter->SetDestinationSkipAxes(SkipType{ { false, false, true } });
                                             ^~~~~~~~~~~~~~~~~~
                                             {                 }
/Users/builder/external/ITK/Modules/Filtering/ImageGrid/test/itkPasteImageFilterGTest.cxx:198:46: warning: suggest braces around initialization of subobject [-Wmissing-braces]
  filter->SetDestinationSkipAxes(SkipType{ { false, false, true } });
                                             ^~~~~~~~~~~~~~~~~~
                                             {                 }
/Users/builder/external/ITK/Modules/Filtering/ImageGrid/test/itkPasteImageFilterGTest.cxx:205:46: warning: suggest braces around initialization of subobject [-Wmissing-braces]
  filter->SetDestinationSkipAxes(SkipType{ { true, false, false } });
                                             ^~~~~~~~~~~~~~~~~~
                                             {                 }
/Users/builder/external/ITK/Modules/Filtering/ImageGrid/test/itkPasteImageFilterGTest.cxx:213:46: warning: suggest braces around initialization of subobject [-Wmissing-braces]
  filter->SetDestinationSkipAxes(SkipType{ { true, true, true } });
                                             ^~~~~~~~~~~~~~~~
                                             {               }
/Users/builder/external/ITK/Modules/Filtering/ImageGrid/test/itkPasteImageFilterGTest.cxx:217:46: warning: suggest braces around initialization of subobject [-Wmissing-braces]
  filter->SetDestinationSkipAxes(SkipType{ { false, true, true } });
                                             ^~~~~~~~~~~~~~~~~
                                             {                }
/Users/builder/external/ITK/Modules/Filtering/ImageGrid/test/itkPasteImageFilterGTest.cxx:221:46: warning: suggest braces around initialization of subobject [-Wmissing-braces]
  filter->SetDestinationSkipAxes(SkipType{ { true, true, false } });
                                             ^~~~~~~~~~~~~~~~~
                                             {                }
/Users/builder/external/ITK/Modules/Filtering/ImageGrid/test/itkPasteImageFilterGTest.cxx:299:46: warning: suggest braces around initialization of subobject [-Wmissing-braces]
  filter->SetDestinationSkipAxes(SkipType{ { true, false, false } });
                                             ^~~~~~~~~~~~~~~~~~
                                             {                 }
/Users/builder/external/ITK/Modules/Filtering/ImageGrid/test/itkPasteImageFilterGTest.cxx:304:46: warning: suggest braces around initialization of subobject [-Wmissing-braces]
  filter->SetDestinationSkipAxes(SkipType{ { false, true, false } });
                                             ^~~~~~~~~~~~~~~~~~
                                             {                 }
/Users/builder/external/ITK/Modules/Filtering/ImageGrid/test/itkPasteImageFilterGTest.cxx:310:46: warning: suggest braces around initialization of subobject [-Wmissing-braces]
  filter->SetDestinationSkipAxes(SkipType{ { false, false, true } });
                                             ^~~~~~~~~~~~~~~~~~
                                             {                 }
```

Adding the extra braces as it suggests fixes these others (well, I only tried the first).

jhlegarreta:
Thanks @seanm :+1: ! I've push forced the fix. Let's see if the rest of the compilers are happy as well ! If the CIs come green I will merge right away.

jhlegarreta:
Builds are green ! Merging.

Thanks to all for the reviews, suggestions and the effort. Let's see if the affected builds in CDash come also green !

jhlegarreta:
The related warnings are gone from `RogueResearch7` and `RogueResearch18 ` according to today's CDash. Thanks everybody !

PR #2053, opened 2020-10-10
STYLE: Rename ITK_DISALLOW_COPY_AND_ASSIGN to ITK_DISALLOW_COPY_AND_MOVE

N-Dekker:
Clarifies that the macro does not just disallow copy and assign, but
also move operations. Note that in this context, the term "move" refers
to both move-construct and move-assign.

With this commit, the old macro name will remain available, as long as
`ITK_FUTURE_LEGACY_REMOVE = OFF` (which is the default).


dzenanz:
This will cause another round of changes in remote modules, dependent libraries and application to update this macro. I am not sure whether this clarification is worth it.

hjmjohnson (commented):
I think all 3 functions should be listed in the newly named macro `ITK_DISALLOW_COPY_MOVE_ASSIGN` 

I like the clarity.  For the moment we could make the old name an alias (not legacy), but have a note to make it legacy at the next major update.  This will give a long transition period for the new name.

dzenanz:
@hjmjohnson That makes sense. For now, put under future legacy macro.

N-Dekker:
Thanks for your feedback,  Hans and Dženan.

@hjmjohnson 
> I think all 3 functions should be listed in the newly named macro `ITK_DISALLOW_COPY_MOVE_ASSIGN`

3 functions? But the macro actually deletes four of them:

1. The copy-constructor
2. The copy-assignment operator
3. The move-constructor
4. The move-assignment operator

https://github.com/InsightSoftwareConsortium/ITK/blob/v5.1.1/Modules/Core/Common/include/itkMacro.h#L358-L362

You could mention all four in a macro name, but that would then be a very long macro name. Would you like that?

Conceptually the macro disallows both copy and move, so I would still prefer to name it `ITK_DISALLOW_COPY_AND_MOVE`. 

@dzenanz 
> For now, put under future legacy macro.

You mean `ITK_FUTURE_LEGACY_REMOVE`, right, instead of `ITK_LEGACY_REMOVE`?  

    #if !defined(ITK_FUTURE_LEGACY_REMOVE)
    #  define ITK_DISALLOW_COPY_AND_ASSIGN(TypeName) call-the-new-macro
    #endif

Would be fine to me.

hjmjohnson:
@N-Dekker I agree with you regarding COPY_MOVE only in the name.  Thanks for pointing this out.

Also good with future deprecation rather than current deprecation.

Modules/Core/Common/include/itkMacro.h
```diff
@@ -355,12 +355,16 @@ namespace itk
...
+#define ITK_DISALLOW_COPY_AND_MOVE(TypeName)                                                                           \
   TypeName(const TypeName &) = delete;                                                                                 \
   TypeName & operator=(const TypeName &) = delete;                                                                     \
   TypeName(TypeName &&) = delete;                                                                                      \
   TypeName & operator=(TypeName &&) = delete
 
+#if !defined(ITK_FUTURE_LEGACY_REMOVE)
+#  define ITK_DISALLOW_COPY_AND_ASSIGN(TypeName) ITK_DISALLOW_COPY_AND_MOVE(TypeName)
+#endif
+
```
N-Dekker, line 367: This is the most relevant part of the pull request. Please check!

N-Dekker:
The problem I feel about the old macro name "ITK_DISALLOW_COPY_AND_ASSIGN" isn't _only_ about the fact that it does not mention "move". To me, the expression "disallow copy and assign" sounds like "disallow dogs and pets". Even though not all pets are dogs, and not all dogs are pets, it sounds a bit funny to me.  Hope that makes some sense to you 😺

Modules/Core/Common/include/itkMacro.h
dzenanz: Thanks for pointing this out, it is hard to find it in GitHub interface of all the changes. Possibly make this change to `itkMacro.h` a separate first commit, with the second commit being a style commit updating from `ITK_DISALLOW_COPY_AND_ASSIGN` to `ITK_DISALLOW_COPY_AND_MOVE`.

dzenanz (approved):

hjmjohnson (approved):
@N-Dekker THANKS!  This is really great.  Thank you for being persistent.

blowekamp (approved):
Using future legacy seems remove appropriate here.

dzenanz (approved):

jhlegarreta:
Looks like some were left behind?
https://open.cdash.org/index.php?project=Insight
e.g.
https://open.cdash.org/viewBuildError.php?buildid=6817858

Or maybe these machines have some `CMake`  legacy-related flags that make the `itkMacro.h` legacy `#if` block not to be taken into account?

hjmjohnson:
@jhlegarreta  See https://github.com/InsightSoftwareConsortium/ITK/pull/2059

PR #2066, opened 2020-10-21
ENH: allow dependencies on remote modules

dzenanz:
Remote modules can now depend on one another when being built as part of ITK.

External builds do not cause ITK to be reconfigured with their dependencies, and throw a configure-time error if not all the required modules are turned on in the base ITK build. This is same as before, and I don't think it would be desirable to have ITK reconfigured (if needed by an externally built remote module).

Technically, built-in modules can now depend on remote modules, but we do not allow this as a matter of policy.

### DIFF
--- a/CMake/ITKModuleEnablement.cmake
+++ b/CMake/ITKModuleEnablement.cmake
@@ -1,46 +1,57 @@
 # This script sorts out the module dependencies, provides user options for customizing
 # the list of modules to be built, and enables modules accordingly.
 
-# Load the module DAG.
-set(ITK_MODULES_ALL)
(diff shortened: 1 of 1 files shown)


thewtex (approved):
Nice!!

CC @SimonRit

jhlegarreta (approved):
I have not been faced to this use-case, but I think this is a great addition. Not sure about not re-configuring ITK if the remote needs it, but probably there are good reasons.

Thanks for doing this.

SimonRit:
Great, thanks for doing it! cc @LucasGandel

PR #2077, opened 2020-10-28
Python Package Organization

hjmjohnson:
An interface file can be used to provide introspection into the itk python package libraries.

the ITK package was re-organized to place all package files INSIDE the itk package directory:
  itk/__init__.py     # the itk package and mechanisms to implement lazy loading
     /Configuraiton  # A `database` of information (load from formatted python files) describing the swig generated content
    /conf                  # a sub-package needed for backward compatibility so that historical mechanisms of en/dis-abling lazy loading are maintained
    /support             # support files needed for lazy loading, and lifting symbols into the itk python package namespace

Many of the static analysis tools (pytype, mypy) analyze an entire package organized under one directory as a single unit.

The itk/__init__.py file 'lifts symbols' from the individual modules into the itk package namespace in a dynamic way that is not visible by many IDE's, the explicit generation of an interface file can provide the necessary information needed to assist
development.

```
python3 ${ITK_BUILD_DIR}/Wrapping/Generators/Python/itk_generate_pyi.py
```
The .pyi file provides hints to IDE's so that they can help with dynamic introspection
during code development.

## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) was made (or the changes have been approved)
- [ ] major design change to allow IDE's introspection based on the python interface file.


Wrapping/Generators/Python/CMakeLists.txt
```diff
@@ -1002,3 +1002,10 @@ else()
   endmacro()
 
 endif()
+
+add_custom_command(OUTPUT ${ITK_WRAP_PYTHON_BINARY_DIR}/itk/__init__.pyi
+    COMMAND ${Python3_EXECUTABLE} ${ITK_WRAP_PYTHON_BINARY_DIR}/itk_generate_pyi.py
+    DEPENDS ${ITK_WRAP_PYTHON_FILES} ${WRAPPER_LIBRARY_NAME}Swig
```
hjmjohnson, line 1008: I am concerned that these dependencies are insufficient for parallel builds.  This script needs to be the last thing that is run (AFTER all the libraries are created and all the python files are generated).  A fully working python environment is needed for the introspection to work.

dzenanz (commented):
I am not sure I am of any help. Matt should review this.

Wrapping/Generators/Python/CMakeLists.txt
```diff
@@ -151,9 +151,9 @@ if(NOT EXTERNAL_WRAP_ITK_PROJECT)
     itkHelpers
     itk/__init__
     itk/itkInitHelpers
+    itk_generate_pyi # A utility function used to generate the intreface file
```
thewtex, line 154: These are a list of files installed and needed for the runtime -- this build-time script does not need to be installed.

Wrapping/Generators/Python/CMakeLists.txt
thewtex: Yes, we should keep the parallel build capabilities. We could generate a `*.pyi` file for each module's wrapping. This will also make it possible to generate `*.pyi` files for remote modules. The script may need to be tweaked, but this `add_custom_command` could be added in the `itk_end_wrap_module_python` macro where the `${WRAPPER_LIBRARY_NAME}Swig` is already defined (note that there is one per module).  The module to generate types for would be `itk.ITKSmoothingPython`, `itk.ITKCommonPython`, `itk.ITKSpatialObjectsPython`, etc.. Then, I assume that there is a way to reference these from `itk/__init__.pyi`.

Wrapping/Generators/Python/CMakeLists.txt
hjmjohnson: I think we would want this installed as well.   I think that the implied content of the itk package (from __init__.py) dynamically changes based on which modules (or remote modules) are pip installed.  I was thinking that this would need to also be installed so that a savvy user could update the __iniit__.pyi file if additional remote modules are installed.

hjmjohnson:
Matt,  

I am not following the logic you describe above.  I don't think that an interface file __init__.py can reference other interface files.  The problem is that we are not "importing " symbols from the other modules (like itk.ITKSmoothingPython) into the itk package.  We are instead injecting attributes (to be lazy loaded) into the itk package directly bypassing the normal import mechanisms.  By using the interface file, we provide the interface to functions that do not yet exist (i.e. they are not loaded yet) so that the IDE can lean on the interface file for knowing what "could exists" once requested and loaded.

thewtex:
Checking into [PEP 484, Stub Files](https://www.python.org/dev/peps/pep-0484/#stub-files) Section, it suggests that we may be able to keep parallel builds, and break out the definitions into smaller files:

> However, as an exception to the previous bullet, all objects imported into a stub using from ... import * are considered exported.

So, if I understand correctly, `itk/__init__.pyi` would have

```
from .ITKCommonPython import *
from .ITKSmoothingPython import *
[...]
```

and *itk/ITKCommonPython.pyi*, *itk/ITKSmoothingPython.pyi* would need to be generated. This would help the build parallelism. Remote modules could generate their own *itk/ITKMyRemoteModule.pyi* files with your script. `itk/__init__.pyi` would not reference them, but some type information may be better than no type information without an additional manual step.

hjmjohnson:
FYI:  I'm still working on understanding, and hopefully reducing the complexity so that we can get better introspection.  I've had many failures and false starts over the past 2 weeks.

hjmjohnson:
@thewtex This work is not yet complete, but I think that this huge PR represents improvements in maintainability, and improved readability (for developers).  

This set of changes addresses some of the fragility in the current python generation system.  There are many implicit assumptions and many implicit naming conventions that are not well documented. 

This work represents my attempts to untangle some of the circular dependencies, and some attempts to use a more conventional symbol management (i.e. prefer 'from mymodule import *` over manipulation of `mymodule.__dict__[]` elements, and attempts to minimize pollution of the global or itk namespaces. 

I propose that we attempt to get this set of changes polished and included, and then revisit the primary goal of improving introspection capabilities with .pyi files.

Wrapping/Generators/Python/Tests/findEmptyClasses.py
```diff
@@ -70,22 +63,32 @@ def isEmpty(o):
            "HammingWindowFunction",
            "LanczosWindowFunction",
            "WelchWindowFunction",
+           "Tile"  # include from itkTileMontage remote module but only templated over dimension.
```
hjmjohnson: Line 68 is the new item that needs to be ignored, but only when the remote module is present.

The other changes in this file are cosmetic and should have been put in a different patch set in an ideal development situation.

Wrapping/Generators/Python/itk/__init__.py
```diff
@@ -18,9 +18,13 @@
...
 
-from .conf import itkConfig as _itkConfig
+from .conf.itkConfig import ITK_GLOBAL_VERSION_STRING as __version__
 
-__version__ = _itkConfig.ITK_GLOBAL_VERSION_STRING
+from .conf.itkTemplate import image
+from .conf.itkTemplate import output
+from .conf.itkExtras import *
+from .conf.itkInitHelpers import *
+from itkTypes import *
```
hjmjohnson: NOTE:  Much of the work up until now was a result of changes made in search of the mechanisms needed to provide the above "standard" symbol lifting behavior that mypy and pytype can interpret.  

This is a key patch-set change on the way to automating the introspection capabilities that are desired.

hjmjohnson:
NOTE:  I would like to move this work unit through soon with minimal efforts needed to get to the next stage of making python more accessible from IDE's with better introspection.

Wrapping/Generators/Python/itk/__init__.py
```diff
@@ -45,82 +46,55 @@ def _get_lazy_attributes(local_lazy_attributes, l_module, l_data):
...
             # because the identity of sys.modules['itk'] would always be changing.
             sys.modules[__name__].__init__(__name__, lazy_attributes)
             del lazy_attributes
         else:
             # Create a new LazyITKModule
-            lzy_module = itkLazy.LazyITKModule(__name__, lazy_attributes)
-            # Set the __path__ attribute, which is required for this_module
-            # to be used as a package
-            setattr(lzy_module, "__path__", __path__)
-            setattr(lzy_module, "__spec__", __spec__)  # pytype: disable=name-error
```
thewtex, line 76: We still need these

Wrapping/Generators/Python/itk/support/itkTemplate.py
```diff
@@ -22,110 +22,16 @@
...
-            extra_eg=extra_eg,
-        )
-        TypeError.__init__(self, msg)
+from ..support import itkBase
+from ..support.itkExtras import output
+from ..support.itkTypes import itkCType
+import math
 
 
-class itkTemplate(object):
```
thewtex, line 128: Why is the class renamed? I don't think the name is better, and it may cause backwards compatibility issues.

thewtex (changes requested):
There are empty *__init__.py* file(s) added -- perhaps unintentional?

Wrapping/Generators/Python/itk/__init__.py
thewtex: Please also install `dask` in the test environment to enabled the related tests.

Wrapping/Generators/Python/itk/support/itkTemplate.py
hjmjohnson: It was renamed because the name of the module and the name of the class were the same, and the module was included both with full path and by lifting the symbols.  It became impossible to distinguish between functions in the module and static member functions in the class.  There was a bug that I was chasing down that greatly confounded by the ambiguity in the same naming for a different context.

hjmjohnson: PS:  I agree the name is not better by itself, but the fact that it is different from the module is important.  I could also rename the name of the module ... the same potential backward compatibility problem could occur.

CMake/ITKModuleExternal.cmake
```diff
@@ -37,17 +37,22 @@ endif()
...
 # Setup build locations.
 if(NOT CMAKE_RUNTIME_OUTPUT_DIRECTORY)
-  set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${ITK_DIR}/bin)
-endif()
-if(NOT CMAKE_LIBRARY_OUTPUT_DIRECTORY)
-  if(ITK_WRAP_PYTHON)
-    set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${ITK_DIR}/Wrapping/Generators/Python/itk)
-  else()
-    set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${ITK_DIR}/lib)
+  set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${ITK_DIR}/bin CACHE PATH "Runtime output directory for ITK")
```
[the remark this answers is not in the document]
hjmjohnson: How do I do that?

Wrapping/Generators/Python/itk/__init__.py
hjmjohnson: Done for my local environments.  

I don't know what to change for the CI environments.

Wrapping/Generators/Python/itk/__init__.py
```diff
@@ -70,8 +70,10 @@ def _get_lazy_attributes(local_lazy_attributes, l_module, l_data):
             lzy_module = _itkLazy.LazyITKModule(__name__, lazy_attributes)
             # Set the __path__ attribute, which is required for this_module
             # to be used as a package
-            setattr(lzy_module, "__path__", __path__)
-            setattr(lzy_module, "__spec__", __spec__)  # pytype: disable=name-error
+            setattr(lzy_module, "__path__", sys.modules[__name__].__path__)
```
hjmjohnson: @thewtex  NOTE:  These are still set, but now preferring the fully qualified name to avoid a pytype failure when using the implicit __spec__

Wrapping/Generators/Python/itk/support/itkTemplate.py
hjmjohnson: @thewtex I can change this back, but it does make debugging and maintenance a lot harder.  Please let me know if you think the potential for backward compatibility issues outweighs the ability to debug the code more easily.

CMake/ITKModuleExternal.cmake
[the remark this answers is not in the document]
hjmjohnson: I turned ITKAnisotropicDiffusionLBR on with wrapping inside of an ITK build.  Is that sufficient for testing?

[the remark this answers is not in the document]
thewtex: That is good. We should also test building a module externally, i.e. clone the ITKAnisotropicDiffusionLBR repository, create its own build tree, and build it passing `ITK_DIR`.

Wrapping/Generators/Python/itk/support/itkTemplate.py
thewtex: Yes, having a different module and class name would be nice, but I do think we should avoid the change in this case. Thanks, Hans.

CMake/ITKModuleExternal.cmake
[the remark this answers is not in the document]
hjmjohnson: @thewtex  A minor change in the order of processing caused a parameter to be used before being set.  This is now fixed locally.  

```bash
git clone 	https://github.com/InsightSoftwareConsortium/ITKAnisotropicDiffusionLBR.git
mkdir ITKAnisotropicDiffusionLBR-bld
cd ITKAnisotropicDiffusionLBR-bld
cmake -GNinja -DITK_DIR:PATH=../ITK-bld ../ITKAnisotropicDiffusionLBR
ninja
ctest 
```
```bash
100% tests passed, 0 tests failed out of 18

Label Time Summary:
AnisotropicDiffusionLBR    =  23.83 sec*proc (18 tests)

Total Test time (real) =  12.88 sec
````

File: Wrapping/Generators/Python/itk/support/itkTemplate.py (hunk 6 above)
Comment (hjmjohnson) [reply Wrapping/Generators/Python/itk/support/itkTemplate.py:128]: Reverted to support the previous naming.    Broke class into two components, one for the static base class, one for the non-static components.

File: CMake/ITKModuleExternal.cmake (hunk 7 above)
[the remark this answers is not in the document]
Comment (hjmjohnson) [reply CMake/ITKModuleExternal.cmake:40]: Yes!  And they work!

File: Wrapping/Generators/Python/itk/__init__.py (hunk 5 above)
Comment (thewtex) [reply Wrapping/Generators/Python/itk/__init__.py:76]: @hjmjohnson `__path__` and `__spec__` are still being marked as removed in the global diff's.

**Review by thewtex — CHANGES_REQUESTED** [review]

File: Wrapping/Generators/Python/itk/__init__.py (hunk 5 above)
Comment (hjmjohnson) [reply Wrapping/Generators/Python/itk/__init__.py:76]: @thewtex   

Sorry, it took me a moment to remember.   The new code is needed to lift *all* the existing symbols into the lzy_module before overriding the current module with the replacement module.

```diff
-            # Set the __path__ attribute, which is required for this_module
-            # to be used as a package
-            setattr(lzy_module, "__path__", sys.modules[__name__].__path__)
-            setattr(
-                lzy_module, "__spec__", sys.modules[__name__].__spec__
-            )  # pytype: disable=name-error
+
+            # Pre-existing attributes need to be propagated too!
+            # except for the lazy overridden elements
+            exclusion_copy_list = ["__name__", "__loader__", "__builtins__"]
+            for k, v in sys.modules[__name__].__dict__.items():    #<--- SEE HERE   This line includes both __spec__ and __path__
+                if k not in exclusion_copy_list:
+                    setattr(lzy_module, k, v)  #<-- This line sets the the values
+
+             # Now override the default  sys.modules[__name__] (__name__  == 'itk' )
+             sys.modules[__name__] = lzy_module
```

Previously this behavior was done hap-hazardly, and sometimes redundantly when lazy loading was not used.  The following is no longer needed!
```diff
-    # Regardless of how it was loaded, fill up the itk module with the ITK types
-    # and extras.
-    import itkTypes
-
-    for k, v in itkTypes.__dict__.items():
-        if k != "itkCType" and not k.startswith("_"):
-            setattr(sys.modules[__name__], k, v)
-    del itkTypes
-
-    from .conf import itkInitHelpers as _itkInitHelpers
-
-    for k, v in _itkInitHelpers.__dict__.items():
-        if not k.startswith("_"):
-            setattr(sys.modules[__name__], k, v)
-
-    from .conf import itkExtras as _itkExtras
-
-    for k, v in _itkExtras.__dict__.items():
-        if not k.startswith("_"):
-            setattr(sys.modules[__name__], k, v)
-
-    # --
-    # Needed to propagate symbol to itk.image from itkTemplate.image
-    from .conf import itkTemplate as _itkTemplate
-
-    setattr(sys.modules[__name__], "image", _itkTemplate.image)
-    setattr(sys.modules[__name__], "output", _itkTemplate.output)
-    # --
```

FYI:  I used a debugger to verify that both 

```python

for k, v in sys.modules[__name__].__dict__.items():
  if k not in exclusion_copy_list:
    print(k)                            # <--------------------------------MATT Added for debugging locally 
    setattr(lzy_module, k, v)

############## console output
__doc__
__package__
__spec__                            # <------------ Verified included
__path__                             # <------------ Verified included
__file__
__cached__
conf
__version__
support
re
Optional
Union
Dict
```

Thanks for your careful review, I really appreciate it.

Hans

hjmjohnson:
@thewtex  See longer explanation here https://github.com/InsightSoftwareConsortium/ITK/pull/2077#discussion_r528758508

Wrapping/Generators/Python/itk/__init__.py
thewtex: @hjmjohnson great, thanks for the clarification!

Those two attributes were required for some very tricky Python import behavior. Crossing fingers that this refactoring will improve that behavior overall :crossed_fingers:

thewtex (approved):

PR #2080, opened 2020-10-30
STYLE: Remove C-style (const void *) casts

N-Dekker:
Following the C++ Core Guidelines, which says (August 3, 2020):
> Don't cast between pointer types when the conversion could be implicit
...
> Don't use C-style (T)expression or functional T(expression) casts

http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#prosafety-type-safety-profile


dzenanz (approved):

dzenanz:
Unfortunately, this style change doesn't compile on any platform 😭

dzenanz (changes requested):
Make it compile :D

thewtex (approved):
:+1:

dzenanz (approved):

hjmjohnson (approved):

PR #2091, opened 2020-11-03
STYLE: Remove "include guards" from cxx files

N-Dekker:
Include guards (like `itkSomeFileName_h`) are specifically intended to
avoid problems with double inclusion of header files, so they should
only be used in *.h and *.hxx files.

Corresponding elastix pull request:
https://github.com/SuperElastix/elastix/pull/345

### DIFF
--- a/Modules/Core/SpatialObjects/src/itkSpatialObjectProperty.cxx
+++ b/Modules/Core/SpatialObjects/src/itkSpatialObjectProperty.cxx
@@ -15,8 +15,6 @@
  *  limitations under the License.
  *
  *=========================================================================*/
-#ifndef itkSpatialObjectProperty_cxx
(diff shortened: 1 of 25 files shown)


hjmjohnson (approved):

dzenanz (approved):
I guess these sneaked in by converting `.hxx` files into `.cxx`.

thewtex (approved):

N-Dekker:
The itkJPEG2000Test06 failure seems unrelated to this pull request https://open.cdash.org/test/279376382

N-Dekker:
> I guess these sneaked in by converting `.hxx` files into `.cxx`.

@dzenanz  I don't really know, but it sounds like a reasonable explanation 😃

PR #2073, opened 2020-10-27
ENH: Wrap cast image filter for VectorImage

Leengit:
Building on code of @thewtex, this code is wrapping image casting to and from `itk::VectorImage< ... >` from several `itk::Image< ... >` types such as those based upon `itk::Vector< ... >` or `itk::RGBPixel< ... >` pixel types.

## PR Checklist
- [ ] Added test (or behavior not changed)
- [x] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [ ] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKExamples) for all new major features (if any)

Submitting this as a draft pull request to trigger Github actions.  It is not yet ready for review.

PR #2094, opened 2020-11-04
ENH: Support for VariableLengthVector in CastImageFilter

Leengit:
This pull request resolves the issue that the use of the `OutputPixelType` constructor in `itkCastImageFilter::DynamicThreadedGenerateDataDispatched` does not work when the pixel type is a `VariableLengthVector`.

## PR Checklist

### DIFF
--- a/Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
+++ b/Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
@@ -147,12 +147,12 @@ CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatche
 
   inputIt.GoToBegin();
   outputIt.GoToBegin();
+  OutputPixelType value{ outputIt.Get() };
   while (!inputIt.IsAtEnd())
   {
     while (!inputIt.IsAtEndOfLine())
     {
       const InputPixelType & inputPixel = inputIt.Get();
-      OutputPixelType        value;
       for (unsigned int k = 0; k < componentsPerPixel; k++)
       {
         value[k] = static_cast<typename OutputPixelType::ValueType>(inputPixel[k]);
--- a/Modules/Filtering/ImageFilterBase/test/itkCastImageFilterTest.cxx
+++ b/Modules/Filtering/ImageFilterBase/test/itkCastImageFilterTest.cxx
@@ -44,22 +44,32 @@ namespace itk
 
 
 // Compile time check
-template class itk::CastImageFilter<itk::Image<std::complex<float>>, itk::Image<std::complex<float>>>;
-template class itk::CastImageFilter<itk::Image<std::complex<double>>, itk::Image<std::complex<double>>>;
-template class itk::CastImageFilter<itk::Image<std::complex<float>>, itk::Image<std::complex<double>>>;
-template class itk::CastImageFilter<itk::Image<std::complex<double>>, itk::Image<std::complex<float>>>;
+template class itk::CastImageFilter<itk::Image<std::complex<float>, 2>, itk::Image<std::complex<float>, 2>>;
+template class itk::CastImageFilter<itk::Image<std::complex<double>, 2>, itk::Image<std::complex<double>, 2>>;
+template class itk::CastImageFilter<itk::Image<std::complex<float>, 2>, itk::Image<std::complex<double>, 2>>;
+template class itk::CastImageFilter<itk::Image<std::complex<double>, 2>, itk::Image<std::complex<float>, 2>>;
 
+template class itk::CastImageFilter<itk::Image<itk::RGBPixel<unsigned char>, 2>,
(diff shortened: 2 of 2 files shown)


dzenanz (approved):
Code looks good, but style requires slightly different whitespace. Please run clang-format on the changed source files.

Leengit:
@dzenanz, clang-format is trying to change, for example,
```c++
struct PixelFactory< itk::VariableLengthVector< TSubPixel > >
```
to
```c++
struct PixelFactory<itk::VariableLengthVector<TSubPixel>>
```
Is the latter the desired style?  (Has the style guide been updated recently?)

dzenanz:
The style is [what clang-format 8.0.2 produces](https://discourse.itk.org/t/update-coding-style-for-itk/2055/20?u=dzenanz). We agreed not to fight it. Major style update was done maybe a year ago.

blowekamp:
Can you please add a test case for the conversion you are adding to the filter?

The filter works for variable length -> variable length. Based you the commit/PR message its not clear what conversion you are trying to add but I think it's something like fix array -> variable length.

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
```diff
@@ -147,12 +147,12 @@ CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatche
 
   inputIt.GoToBegin();
   outputIt.GoToBegin();
+  OutputPixelType value{ Self::PixelFactory<OutputPixelType>::New(componentsPerPixel) };
```
blowekamp, line 150: The pixel factory is not needed. Just do the following:
```
OutputPixelType value = outputIt.Get();
```
I believe the above should be a "deep copy" but please test the output in the test.

Leengit: Will do.

Leengit:
> Can you please add a test case for the conversion you are adding to the filter?

Will do.

> The filter works for variable length -> variable length. Based you the commit/PR message its not clear what conversion you are trying to add but I think it's something like fix array -> variable length.

A comment by @blowekamp in Issue https://github.com/InsightSoftwareConsortium/ITK/pull/1579 suggests that casting to a `VectorImage` did not work in `CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatche` due to an invalid C++ construction  of an `OutputPixelType`.  This PR fix addresses this.

Yes, the `Image<Vector<type,c>,d>` ↔ `VectorImage<type,d>` was just implemented, in PR https://github.com/InsightSoftwareConsortium/ITK/pull/2073.

blowekamp:
The simplicity of this solution was missed when I looked at this. Thank you for changing to the simpler solution and adding tests.

With this conversion: `VectorImage<type,d> ↔ Image<Vector<type,c>,d>` it may be worth wild to test/verify, that an exception will be thrown ( early in the pipeline e.g. `GenerateOutputInformation`? ) when the dimensions of the vector-like objects don't match.

Leengit:
> With this conversion: `VectorImage<type,d> ↔ Image<Vector<type,c>,d>` it may be worthwhile to test/verify, that an exception will be thrown ( early in the pipeline e.g. `GenerateOutputInformation`? ) when the dimensions of the vector-like objects don't match.

I believe that the desired behavior is to do something reasonable even when the dimensions of the input and and output images are not equal.  For example, the comment at the top of of the definition of `CastImageFilter::GenerateOutputInformation()` is:
```c++
// do not call the superclass' implementation of this method since
// this filter allows the input the output to be of different dimensions
```

blowekamp:
> > With this conversion: `VectorImage<type,d> ↔ Image<Vector<type,c>,d>` it may be worthwhile to test/verify, that an exception will be thrown ( early in the pipeline e.g. `GenerateOutputInformation`? ) when the dimensions of the vector-like objects don't match.
> 
> I believe that the desired behavior is to do something reasonable even when the dimensions of the input and and output images are not equal. For example, the comment at the top of of the definition of `CastImageFilter::GenerateOutputInformation()` is:
> 
> ```c++
> // do not call the superclass' implementation of this method since
> // this filter allows the input the output to be of different dimensions
> ```

There are two "dimensions":
 - The dimension of the Image ( which your comment refers too )
 - The dimension/size/length of the pixel component vector ( which my comment refers too ).

If the length of the input pixel vector and the output vector differ we need to ensure we have "well defined" behavior and don't over run any buffers. My expectation would be an exception would be thrown some where under the `GenerateOutputInformation` when it is detected that length of the pixel vectors are not equal.

Leengit:
> If the length of the input pixel vector and the output vector differ we need to ensure we have "well defined" behavior and don't over run any buffers. My expectation would be an exception would be thrown some where under the `GenerateOutputInformation` when it is detected that length of the pixel vectors are not equal.

Ah, yes, of course, thank you.  Will do.

thewtex (approved):

PR #2096, opened 2020-11-04
STYLE: Remove space between class and member names in C++ source files

N-Dekker:
It appears that clang-format has in some cases introduced a space
character between `ClassName` and `::MemberName`. This typically seems
to have happened when there originally was a line break between the
class name and the member name.

This commit fixes the formatting by removing those spaces, as follows:

    find . \( -iname *.cxx -and ! -path *ThirdParty* \) -exec perl -pi -w -e 's/([A-Z][a-z]+) \:\:([A-Z][a-z])/$1::$2/g;' {} \;
    find . \( -iname *.hxx -and ! -path *ThirdParty* \) -exec perl -pi -w -e 's/([A-Z][a-z]+) \:\:([A-Z][a-z])/$1::$2/g;' {} \;
    find . \( -iname *.h -and ! -path *ThirdParty* \) -exec perl -pi -w -e 's/([A-Z][a-z]+) \:\:([A-Z][a-z])/$1::$2/g;' {} \;

### DIFF
--- a/Examples/RegistrationITKv4/DeformableRegistration13.cxx
+++ b/Examples/RegistrationITKv4/DeformableRegistration13.cxx
@@ -118,7 +118,7 @@ main(int argc, char * argv[])
   }
 
   // For consistent results when regression testing.
-  itk::Statistics::MersenneTwisterRandomVariateGenerator ::GetInstance()
+  itk::Statistics::MersenneTwisterRandomVariateGenerator::GetInstance()
     ->SetSeed(121212);
 
   constexpr unsigned int ImageDimension = 2;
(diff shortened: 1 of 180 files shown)


dzenanz (approved):

N-Dekker:
@hjmjohnson Do you also like this commit? I realize, you wrote that regarding code formatting, ["automation is key"](https://discourse.itk.org/t/update-coding-style-for-itk/2055/20), while this commit is only semi-automatic!

If you don't like this commit, it would be OK to me to close the PR without merge. No problem. Because it's really a very minor cosmetic thing to me. But if you _do_ like it, I think I should adjust the *.hxx files as well.

hjmjohnson (approved):
@N-Dekker  I love this commit !!

Please remove "WIP: and it can be merged".

N-Dekker:
>  I love this commit !!

@hjmjohnson Thanks, I'm glad to hear so  :-)  You know it's partially a matter of taste, so I can image that _some_ people actually like to have a space before the `::MemberName`. But still it looks like a Clang-Format bug anyway.

For the sake of completeness, I think I should also do the *.hxx file, before the PR is ready!

dzenanz (approved):

PR #2107, opened 2020-11-15
STYLE: Remove local possibleTransformIO from TransformIOFactoryTemplate

N-Dekker:
Removed the local `possibleTransformIO` variable from the member
function `TransformIOFactoryTemplate::CreateTransformIO`, as it appears
unnecessary to put all IO objects into that list, before finding the one
that can process the specified file path.

Reduced amount of duplicate code for read and write, and improved
const-correctness.


dzenanz (approved):
Why do you think it was done in two passes before?

N-Dekker:
> Why do you think it was done in two passes before?

@dzenanz Good question! For historical reasons? Maybe the code grew organically like that...?

I guess the idea was originally: let's get all possible IO objects, and then pick the first one that does the job. Now it appears unnecessary to get all of them, when the one that does the job is already found!  😃

thewtex (approved):

PR #2108, opened 2020-11-18
STYLE: TxtTransformIO Read() no longer copy input file to data string

N-Dekker:
Before processing the content of its input file,
`TxtTransformIOTemplate::Read()` copied the content to a raw `char`
buffer, which was then inserted into to an output string stream, whose
string was then copied into `data`.

None of this appears necessary, as the content of the file can also be
processed line by line, by a `while (std::getline(in, line))` loop,
according to this commit.


dzenanz (approved):

N-Dekker:
For the record, it appears that the logic of copying the content of the input file into a `data` string, via an output string stream, has been in there for quite a while: commit 24300b73bd7e931e4c6f471f9dd6c24c8a5df749 (21 September 2005)

Issue #1766, opened 2020-04-13
Error loading TIFF image


brad-t-moore:
### Description

Wrote a float32 single-channel TIFF image using PIL in Python.  PIL and ImageJ can read the file fine, ITK returns garbled image.  I tried just using itk.imread as well as a float-instantiated itk.ImageReader.  itk.imread also loaded the file as unsigned char when it should have been float.

![image](https://user-images.githubusercontent.com/59569148/79123842-0492f780-7d69-11ea-88d9-6897a6518fbd.png)

![image](https://user-images.githubusercontent.com/59569148/79123892-25f3e380-7d69-11ea-905c-a3693d1c6e71.png)

### Steps to Reproduce

Attached the image and notebook.

[BadIOTest.zip](https://github.com/InsightSoftwareConsortium/ITK/files/4469697/BadIOTest.zip)

Below is the output from the .TIF file when PIL saves the data matrix and when ITK save the data matrix.  ITK can read its file.

```
from PIL import Image
from PIL.TiffTags import TAGS

with Image.open('badimage.tif') as img:
    meta_dict = {TAGS[key] : img.tag[key] for key in img.tag.keys()}
print('badimage.tif', meta_dict)
    
with Image.open('badimage-itk.tif') as img:
    meta_dict = {TAGS[key] : img.tag[key] for key in img.tag.keys()}

print('badimage-itk.tif', meta_dict)
```
```
badimage.tif {'ImageWidth': (428,), 'ImageLength': (210,), 'BitsPerSample': (32,), 'Compression': (1,), 'PhotometricInterpretation': (1,), 'StripOffsets': (134,), 'SampleFormat': (2,), 'RowsPerStrip': (210,), 'StripByteCounts': (359520,), 'PlanarConfiguration': (1,)}
badimage-itk.tif {'ImageWidth': (428,), 'ImageLength': (210,), 'BitsPerSample': (32,), 'Compression': (1,), 'PhotometricInterpretation': (1,), 'StripOffsets': (8,), 'Orientation': (1,), 'SampleFormat': (3,), 'SamplesPerPixel': (1,), 'RowsPerStrip': (612,), 'StripByteCounts': (359520,), 'XResolution': ((4294967295, 169093203),), 'YResolution': ((4294967295, 169093203),), 'PlanarConfiguration': (1,), 'ResolutionUnit': (2,), 'Software': ('InsightToolkit',)}
```

### Expected behavior

[...]

PR #2112, opened 2020-11-19
BUG: Add TIFFImageIO support for unsigned int and int pixel types

thewtex:
According to:

awaresystems.be/imaging/tiff/tifftags/sampleformat.html

If SampleFormat is 1, it is unsigned integer data, and if SampleFormat is 2.

Addresses #1766


dzenanz (commented):
This just adds the test, it does not change the code (to fix the bug as implied by commit message).

thewtex:
Ah, yes, adding the missing update.

dzenanz (approved):

PR #2114, opened 2020-11-20
STYLE: Declare OptimizerParameters::m_Helper as C++11 std::unique_ptr

N-Dekker:
Initialized `m_Helper` by an in-class initializer (default member initializer), replacing four `this->Initializer()` calls.

Defaulted the default-constructor and the destructor of `OptimizerParameters`.


Modules/Core/Common/include/itkOptimizerParameters.hxx
```diff
@@ -23,70 +23,51 @@
...
-/** Default contstructor */
+/** Default constructor */
 template <typename TParametersValueType>
-OptimizerParameters<TParametersValueType>::OptimizerParameters()
-  : Array<TParametersValueType>()
-  , m_Helper(nullptr)
-{
-  this->Initialize();
-}
+OptimizerParameters<TParametersValueType>::OptimizerParameters() = default;
```
dzenanz, line 28: Is there a reason for not moving this declaration to .h file now that it is trivial?

Modules/Core/Common/include/itkOptimizerParameters.hxx
```diff
@@ -23,70 +23,51 @@
...
-/** Constructor with Array assignment */
+/** Constructor with Array */
 template <typename TParametersValueType>
 OptimizerParameters<TParametersValueType>::OptimizerParameters(const ArrayType & array)
   : Array<TParametersValueType>(array)
-  , m_Helper(nullptr)
-{
-  this->Initialize();
-}
+{}
```
dzenanz, line 52: All of the constructors look like they could be moved to .h file.

Modules/Core/Common/include/itkOptimizerParameters.hxx
```diff
@@ -23,70 +23,51 @@
...
-}
+OptimizerParameters<TParametersValueType>::~OptimizerParameters() = default;
 
 template <typename TParametersValueType>
 void
 OptimizerParameters<TParametersValueType>::SetHelper(OptimizerParametersHelperType * helper)
 {
-  delete this->m_Helper;
-  this->m_Helper = helper;
+  this->m_Helper.reset(helper);
```
dzenanz, line 70: As all of the methods are now one-liners, maybe we no longer need .hxx file and could move all the implementations into .h file?

dzenanz: If would be good if others pitched in with opinions.

N-Dekker: @dzenanz  If some people still use `ITK_MANUAL_INSTANTIATION`, it may matter. Compilation, link, or run time might then be affected by moving code between "h" and "hxx". (Not sure in what way the duration would be affected!) Do people still use  `ITK_MANUAL_INSTANTIATION`?

Another effect might be that the compiler might derive automatically that a constructor or a destructor is `noexcept` or `constexpr`, when it is defaulted ( `= default`) in the "h file", instead of in the hxx file.

N-Dekker: Correction: a destructor is always `noexcept` by default. But it might be easier for the compiler to figure out that the destructor won't terminate the program (in case of an exception within the destructor), when the destructor is defaulted in the "h file", rather than in the hxx file.

N-Dekker: Personally I would like to have the entire implementation of `OptimizerParameters` in the .h file, and inside the class definition. But of course, it depends on what others think... 😃

hjmjohnson: @N-Dekker I agree that the trivial nature of the content in the .hxx file does not justify the overhead.  In general, I do like large, non-trivial functions to be placed in the .hxx file for readability purposes.

hjmjohnson (approved):

blowekamp:
@N-Dekker I am seeing some compilation error on the Dashboard related to this patch.
```
/home/kitware/Dashboards/TestsModules/Core/Common/include/itkOptimizerParameters.h:53:3: error: converting to 'std::unique_ptr<itk::OptimizerParametersHelper<double>, std::default_delete<itk::OptimizerParametersHelper<double> > >' from initializer list would use explicit constructor 'std::unique_ptr<_Tp, _Dp>::unique_ptr(std::unique_ptr<_Tp, _Dp>::pointer) [with _Tp = itk::OptimizerParametersHelper<double>; _Dp = std::default_delete<itk::OptimizerParametersHelper<double> >; std::unique_ptr<_Tp, _Dp>::pointer = itk::OptimizerParametersHelper<double>*]'
   OptimizerParameters() = default;
```
https://open.cdash.org/viewBuildError.php?buildid=6897873

The system report as gcc 4.8.4 in the configuration (despite the name of the build ).

N-Dekker:
@blowekamp Thanks for reporting this issue, Bradley. Looks like the compile error can be reproduced here at https://godbolt.org/z/1x6zbh right?

    #include <memory>

    template <typename T>
    class OptimizerParameters
    {
      std::unique_ptr<int> m_Helper{new int};
    };

    OptimizerParameters<double> g_optimizerParameters;

It seems like a GCC 4.8.4 - 4.9.0 compiler bug. The error does not appear with GCC 4.9.1 anymore.

GCC 4.8.4 does accept the following syntax, fortunately:

      std::unique_ptr<int> m_Helper = std::unique_ptr<int>{new int};

PR #2118, opened 2020-11-23
BUG: Fix MatrixOffsetTransformBase SetFixedParameters if too few params

N-Dekker:
When a user provided too few parameters as argument to
`MatrixOffsetTransformBase::SetFixedParameters`, the member function
would have undefined behavior, which would typically just cause a crash.

This issue affects transforms derived from `MatrixOffsetTransformBase`, including:

    AffineTransform
    AzimuthElevationToCartesianTransform
    CenteredAffineTransform
    Euler2DTransform
    FixedCenterOfRotationAffineTransform
    QuaternionRigidTransform
    Rigid2DTransform
    Rigid3DTransform
    ScalableAffineTransform
    ScaleLogarithmicTransform
    ScaleSkewVersor3DTransform
    ScaleTransform
    ScaleVersor3DTransform
    Similarity2DTransform
    Similarity3DTransform
    VersorRigid3DTransform
    VersorTransform

A crash could occur when reading a transform from a transform file, when
this file has an insufficient number of fixed parameters. It was
encountered when loading a transform within 3DSlicer (version
4.11.20200930).

The fix consists of throwing an exception when the number of fixed
parameters is too small.

### DIFF
--- a/Modules/Core/Transform/test/itkMatrixOffsetTransformBaseGTest.cxx
+++ b/Modules/Core/Transform/test/itkMatrixOffsetTransformBaseGTest.cxx
@@ -1,0 +1,50 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+// First include the header file to be tested:
+#include "itkMatrixOffsetTransformBase.h"
+
+#include <gtest/gtest.h>
+
+
+namespace
+{
+template <unsigned int NDimensions>
+void
+Assert_SetFixedParameters_throws_when_size_is_less_than_NDimensions()
+{
+  for (unsigned int size{}; size < NDimensions; ++size)
+  {
+    using TransformBaseType = itk::MatrixOffsetTransformBase<double, NDimensions, NDimensions>;
+    using FixedParametersType = typename TransformBaseType::FixedParametersType;
+
+    const auto                transformBase = TransformBaseType::New();
+    const FixedParametersType fixedParameters(size);
+    ASSERT_THROW(transformBase->SetFixedParameters(fixedParameters), itk::ExceptionObject);
+  }
+}
+
+} // namespace
+
+
+TEST(MatrixOffsetTransformBase, SetFixedParametersThrowsWhenSizeIsLessThanNDimensions)
+{
+  Assert_SetFixedParameters_throws_when_size_is_less_than_NDimensions<2>();
+  Assert_SetFixedParameters_throws_when_size_is_less_than_NDimensions<3>();
+  Assert_SetFixedParameters_throws_when_size_is_less_than_NDimensions<4>();
+}
--- a/Modules/Core/Transform/include/itkMatrixOffsetTransformBase.hxx
+++ b/Modules/Core/Transform/include/itkMatrixOffsetTransformBase.hxx
@@ -484,6 +484,11 @@ void
 MatrixOffsetTransformBase<TParametersValueType, NInputDimensions, NOutputDimensions>::SetFixedParameters(
   const FixedParametersType & fp)
 {
+  if (fp.size() < NInputDimensions)
+  {
+    itkExceptionMacro(<< "Error setting fixed parameters: parameters array size (" << fp.size()
+                      << ") is less than expected  (NInputDimensions = " << NInputDimensions << ")");
+  }
   this->m_FixedParameters = fp;
   InputPointType c;
   for (unsigned int i = 0; i < NInputDimensions; i++)
--- a/Modules/Core/Transform/test/CMakeLists.txt
+++ b/Modules/Core/Transform/test/CMakeLists.txt
@@ -181,5 +181,6 @@ set_tests_properties( itkTestTransformGetInverse PROPERTIES COST 50 )
 
 set(ITKTransformGTests
   itkBSplineTransformGTest.cxx
+  itkMatrixOffsetTransformBaseGTest.cxx
 )
 CreateGoogleTestDriver(ITKTransform "${ITKTransform-Test_LIBRARIES}" "${ITKTransformGTests}")


dzenanz (approved):

jhlegarreta:
Sorry to chime in. Thanks for doing this @N-Dekker :100:. Can the condition be tested, e.g. by adding the file that triggered the crash and the corresponding code in the appropriate class test, or by adding a case where a faulty parameter file is simulated by appropriately removing parameters or dimensions for an existing parameter file in the class' test file?

N-Dekker:
> Can the condition be tested, e.g. by adding the file that triggered the crash and the corresponding code in the appropriate class test, or by adding a case where a faulty parameter file is simulated by appropriately removing parameters or dimensions for an existing parameter file in the class' test file?

@jhlegarreta  Thanks for your suggestion, Jon. I agree that a test would be helpful. Actually I'm thinking now of adding a low-level unit test, that directly calls `SetFixedParameters` with an insufficient number of fixed-parameters. Hope that's OK to you as well.

In the mean-time, it would be helpful if you could confirm the issue by loading such a file yourself, for example:

    #Insight Transform File V1.0
    #Transform 0
    Transform: AffineTransform_double_3_3
    Parameters: 1 0 0 0 1 0 0 0 1 0 0 0
    FixedParameters:

(Note: FixedParameters intentionally left blank!)

For example, using 3D Slicer. Are you able to reproduce the application crash?

jhlegarreta:
> Actually I'm thinking now of adding a low-level unit test, that directly calls `SetFixedParameters` with an insufficient number of fixed-parameters. Hope that's OK to you as well.

Sounds good. Thanks Niels.

> In the mean-time, it would be helpful if you could confirm the issue by loading such a file yourself, for example: For example, using 3D Slicer. Are you able to reproduce the application crash?

Sorry Niels, but I will not be able to do this; I do not use 3D Slicer as of today, and I prefer not to start that front on my machine for a few reasons. I do believe the issue exists; I just made the comment because I believe that testing and ensuring that the condition is working is also important. I am well aware that it requires an additional effort, but one that is worthwhile.

N-Dekker:
Another possible approach could be to allow an empty `FixedParameters` argument to specify that the fixed parameters are all zero. `MatrixOffsetTransformBase::SetFixedParameters` would then set `m_Center` to zero (the origin). Would anyone here like that better than just throwing an exception?

dzenanz:
> empty FixedParameters

How many Slicer-written files with no fixed parameters are there? If many, we could make a special case that empty means all zeroes. If not many, I prefer we throw. Specifying all zeroes is easy enough.

N-Dekker:
@lassoan Hi Andras, could you also please have a look at this pull request? Do you agree that it might indeed fix a potential 3D Slicer crash (when the user provides a transform file that does not have enough FixedParameters).

hjmjohnson (approved):
I think this is a good change.

dzenanz (approved):

Modules/Core/Transform/test/itkMatrixOffsetTransformBaseGTest.cxx
```diff
@@ -0,0 +1,67 @@
...
+  }
+  return result;
+}
+
+
+template <unsigned int NDimensions>
+void
+Assert_SetFixedParameters_throws_when_size_is_less_than_NDimensions()
+{
+  for (unsigned int size{}; size < NDimensions; ++size)
```
lassoan, line 51: `unsigned int size{}` is less readable than `unsigned int size=0`. Is there any advantage of initializing with `{}`?

lassoan:
> could you also please have a look at this pull request? Do you agree that it might indeed fix a potential 3D Slicer crash (when the user provides a transform file that does not have enough FixedParameters).

The change looks good to me. It indeed caused problem in Slicer (it caused some exception that was captured at some level, but left the application in an inconsistent state). I've tested this proposed fix and it makes loading fail with a nice descriptive error message.

It would be nice to replace the plain text transform file format with a simple json-based format (and json schema to clearly specify what parts of the file are optional and describe the meaning of the parameters).

Modules/Core/Transform/test/itkMatrixOffsetTransformBaseGTest.cxx
N-Dekker: @lassoan In this specific case, `T var{}` and `T var=0` are basically equivalent. The expression `T var=0` converts an `int` (zero) to the specified type `T`, and when type `T` is `unsigned int`, that's just fine, of course. Zero performance overhead, no rounding errors, no overflow. But in general, when `T` is a user-defined type, `T var=0` may be significantly slower than `T var{}`, because `T var=0` may call a converting constructor of `T`, whereas `T var{}` may just call the default-constructor. 

So in general, I prefer to avoid conversion, when C++11 `{}` initialization can do the job. Conversion may (in cases other than this specific one) involve rounding errors, overflow, and a performance penalty.

Would it be OK to you if I leave the line of code like this? Or do you want me to change it?

dzenanz: `unsigned size = 0u` doesn't have performance penalty. I agree with Andras that `=0` is clearer than `{}` initialization.

dzenanz (approved):

N-Dekker:
> How many Slicer-written files with no fixed parameters are there?

@dzenanz I guess Slicer-written transform files always have the correct number of fixed parameters, but Andras @lassoan please correct me if I'm wrong!

But allowing an empty `FixedParameters` does ease manually editing a transform file. Which is what I did, when trying to load a transform file from elastix into 3DSlicer (as part of pull request https://github.com/SuperElastix/elastix/pull/358).

> Specifying all zeroes is easy enough.

It was not directly obvious to me how many zeroes to specify, for a specific transform. It appears dependent on the dimensionality as well as the transformation type. Please look at the table below here! Anyway, the main aim of this pull request is to avoid application crashes. So if you do not like my suggestion to allow an empty list of FixedParameters to specify _all zeroes_, it would be fine to me to just have the current version of this pull request merged. 😃 

Transform type | # FixedParameters
-- | --
AffineTransform<double,2> | 2
AffineTransform<double,3> | 3
AffineTransform<double,4> | 4
AzimuthElevationToCartesianTransform<double,3> | 3
BSplineDeformableTransform<double,2,2> | 10
BSplineDeformableTransform<double,3,3> | 18
BSplineSmoothingOnUpdateDisplacementFieldTransform<double,2> | 10
BSplineSmoothingOnUpdateDisplacementFieldTransform<double,3> | 18
BSplineTransform<double,2,3> | 10
BSplineTransform<double,3,3> | 18
CenteredAffineTransform<double,2> | 2
CenteredAffineTransform<double,3> | 3
CenteredEuler3DTransform | 4
CenteredRigid2DTransform | 2
CenteredSimilarity2DTransform | 2
CompositeTransform<double,2> | 0
CompositeTransform<double,3> | 0
CompositeTransform<double,4> | 0
ConstantVelocityFieldTransform<double,2> | 10
ConstantVelocityFieldTransform<double,3> | 18
DisplacementFieldTransform<double,2> | 10
DisplacementFieldTransform<double,3> | 18
Euler2DTransform | 2
Euler3DTransform | 4
FixedCenterOfRotationAffineTransform<double,3> | 3
GaussianExponentialDiffeomorphicTransform<double,2> | 10
GaussianExponentialDiffeomorphicTransform<double,3> | 18
GaussianSmoothingOnUpdateDisplacementFieldTransform<double,2> | 10
GaussianSmoothingOnUpdateDisplacementFieldTransform<double,3> | 18
GaussianSmoothingOnUpdateTimeVaryingVelocityFieldTransform<double,2> | 18
GaussianSmoothingOnUpdateTimeVaryingVelocityFieldTransform<double,3> | 28
IdentityTransform<double,2> | 0
IdentityTransform<double,3> | 0
IdentityTransform<double,4> | 0
QuaternionRigidTransform | 3
Rigid2DTransform | 2
Rigid3DPerspectiveTransform | 0
v3::Rigid3DTransform | 3
ScalableAffineTransform<double,3> | 3
ScaleLogarithmicTransform<double,3> | 3
ScaleSkewVersor3DTransform | 3
ScaleTransform<double,2> | 2
ScaleTransform<double,3> | 3
ScaleTransform<double,4> | 4
ScaleVersor3DTransform | 3
Similarity2DTransform | 2
Similarity3DTransform | 3
TimeVaryingBSplineVelocityFieldTransform<double,2> | 18
TimeVaryingBSplineVelocityFieldTransform<double,3> | 28
TimeVaryingVelocityFieldTransform<double,2> | 18
TimeVaryingVelocityFieldTransform<double,3> | 28
TranslationTransform<double,2> | 0
TranslationTransform<double,3> | 0
TranslationTransform<double,4> | 0
VelocityFieldTransform<double,2> | 18
VelocityFieldTransform<double,3> | 28
VersorRigid3DTransform | 3
VersorTransform | 3

dzenanz:
> It was not directly obvious to me how many zeroes to specify

For manual editing, I think adding extra numbers (zeroes) at the end is not a problem.

> it would be fine to me to just have the current version of this pull request merged

OK.

jhlegarreta:
> Sorry Niels, but I will not be able to do this; I do not use 3D Slicer as of today, and I prefer not to start that front on my machine for a few reasons. I do believe the issue exists; I just made the comment because I believe that testing and ensuring that the condition is working is also important. I am well aware that it requires an additional effort, but one that is worthwhile.

@N-Dekker many thanks for having added a test :medal_sports: and the work in this PR.

lassoan:
Slicer uses ITK to write transforms, so all parameters are expected to written correctly.

SlicerElastix extension provide an interface to Elastix (you can specify a binary Elastix installation or use the bundled Elastix version) and takes care of converting Elastix transforms to ITK transforms (linear, bspline, grid) - see https://github.com/lassoan/SlicerElastix/blob/master/Elastix/Elastix.py#L525-L694.

It would be nice if Elastix would use ITK transform format directly, but I can understand that the plain text file format is not appealing and flexible enough.

N-Dekker:
> SlicerElastix extension provide an interface to Elastix (you can specify a binary Elastix installation or use the bundled Elastix version) and takes care of converting Elastix transforms to ITK transforms (linear, bspline, grid) - see https://github.com/lassoan/SlicerElastix/blob/master/Elastix/Elastix.py#L525-L694.

Very interesting, @lassoan I'm currently trying to figure out if (or when) an elastix transform could be converted to a corresponding ITK transform, within elastix core function `TransformBase<TElastix>::WriteToFile`: https://github.com/SuperElastix/elastix/pull/358 But as far as I can see, elastix has [17 transform types](https://github.com/SuperElastix/elastix/pull/358#issuecomment-731571964), and ITK has [58 of them](https://github.com/SuperElastix/elastix/pull/358#issuecomment-731736374). Not sure if a one-to-one correspondence is always possible.

PR #2127, opened 2020-11-26
COMP: Work around GCC-4.8 error converting to OptimizerParametersHelper

N-Dekker:
Worked around an apparent compiler bug in GCC 4.8 (which seems to be
fixed with GCC 4.9.2), causing a compile error from GCC 4.8.4 at
https://open.cdash.org/viewBuildError.php?buildid=6897873 saying:

> Core/Common/include/itkOptimizerParameters.h:53:3:
> error: converting to 'std::unique_ptr< itk::OptimizerParametersHelper>'
> from initializer list would use explicit constructor...

Reported by Bradley Lowekamp (@blowekamp) at
https://github.com/InsightSoftwareConsortium/ITK/pull/2114#issuecomment-733741732

### DIFF
--- a/Modules/Core/Common/include/itkOptimizerParameters.h
+++ b/Modules/Core/Common/include/itkOptimizerParameters.h
@@ -168,7 +168,8 @@ public:
   ~OptimizerParameters() override = default;
 
 private:
-  std::unique_ptr<OptimizerParametersHelperType> m_Helper{ new OptimizerParametersHelperType };
+  std::unique_ptr<OptimizerParametersHelperType> m_Helper =
+    std::unique_ptr<OptimizerParametersHelperType>{ new OptimizerParametersHelperType };
 };
 
 } // namespace itk


dzenanz (approved):

Modules/Core/Common/include/itkOptimizerParameters.h
```diff
@@ -168,7 +168,8 @@ class ITK_TEMPLATE_EXPORT OptimizerParameters : public Array<TParametersValueTyp
   ~OptimizerParameters() override = default;
 
 private:
-  std::unique_ptr<OptimizerParametersHelperType> m_Helper{ new OptimizerParametersHelperType };
+  std::unique_ptr<OptimizerParametersHelperType> m_Helper =
+    std::unique_ptr<OptimizerParametersHelperType>{ new OptimizerParametersHelperType };
```
N-Dekker, line 172: Note that C++14 introduced `std::make_unique`, which should preferably be used here instead of `new`, once ITK has upgraded to >= C++14.

dzenanz: This could be a comment in the source code.

N-Dekker: @dzenanz Well, I considered adding such a comment, but I guess we might as well just run clang-tidy `modernize-make-unique` over the entire ITK, once ITK is upgraded to >= C++14: https://clang.llvm.org/extra/clang-tidy/checks/modernize-make-unique.html

PR #2141, opened 2020-12-02
ENH: ttype can be `ImageType` whenever `(ImageType,)` is accepted

Leengit:
When exactly one type must be supplied via `ttype`, the following three are now equivalent
```python
ttype=ImageType
ttype=(ImageType,)
ttype=[ImageType]
```
so long as `ImageType` is itself a type, not a tuple or list.  When other than exactly one type needs to be supplied, it continues to be the case that only the tuple and list forms work.

### DIFF
--- a/Wrapping/Generators/Python/itk/support/itkTemplate.py
+++ b/Wrapping/Generators/Python/itk/support/itkTemplate.py
@@ -661,9 +661,14 @@ class itkTemplate(Mapping):
             # Note that the function `itk.template()` which returns the template
             # arguments of an object returns tuples and its returned value
             # should be usable in this context.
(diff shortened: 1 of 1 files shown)


Leengit:
Although it is nonsensical as far as I know, the user could supply a Python `str`, as in `ttype="abc"`.  Because strings are convertible to tuple, the old code would interpret this as if the three-tuple `('a', 'b', 'c')` had been supplied (and ultimately give an error because those three characters are not wrapped types), but the new code will treat this as if the one-tuple `('abc',)` is supplied (and ultimately give an error because that three-character string is not a wrapped type).  Does this matter?

thewtex (approved):

hjmjohnson (approved):

PR #2142, opened 2020-12-03
COMP: Adds ComputeTangentAndNormals() function

aylward:
A function name was changed to correct grammar and improve clarity, but
it broke backwards compatibility.   This request adds a function
with the old name to restore backward compatibility.

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkTubeSpatialObject.h
+++ b/Modules/Core/SpatialObjects/include/itkTubeSpatialObject.h
@@ -83,12 +83,10 @@ public:
   bool
   ComputeTangentsAndNormals();
 
+#if !defined(ITK_LEGACY_REMOVE)
   /** Calculate the normalized tangent - Old spelling of function name */
-  bool
-  ComputeTangentAndNormals()
-  {
-    return ComputeTangentsAndNormals();
-  }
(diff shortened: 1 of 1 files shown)


Modules/Core/SpatialObjects/include/itkTubeSpatialObject.h
```diff
@@ -83,6 +83,13 @@ class ITK_TEMPLATE_EXPORT TubeSpatialObject : public PointBasedSpatialObject<TDi
   bool
   ComputeTangentsAndNormals();
 
+  /** Calculate the normalized tangent - Old spelling of function name */
+  bool
```
dzenanz, line 87: Maybe it should be wrapped in legacy macro?

aylward: Done!  Great idea!

dzenanz (approved):

dzenanz:
MacOS error:
```text
-- Configuring done
CMake Error: install(EXPORT "ITKTargets" ...) includes target "gdcmjpeg8" more than once in the export set.
-- Generating done
CMake Generate step failed.  Build files cannot be regenerated correctly.
```

PR #2154, opened 2020-12-05
BUG: Fix multiprocessing imread LazyLoading failures

hjmjohnson:
The loading of modules in python *must* occur as a
single atomic transaction in multiprocessing environments (i.e.
the module should only be loaded by one thread).
    
The LazyLoading of ITK did not treat the loading of
modules as an atomic transaction, and multiple threads
would attempt to load the cascading dependencies out of
order.
    
The `getattr` override that allows LazyLoading to work
in the case where the module is *not* loaded now blocks
while the first thread completes the delayed module loading.
After the first thread completes module load as an atomic
transaction, the other threads fall through (skip loading)
and return the value requested.
    
Need to use a recursive lock for thread ownership so that the
first thread can acquire a RLock as often as needed while
recursively processing dependant modules lazy loads.  Other threads need
to wait until this first thread releases the RLock.

## PR Checklist

### DIFF
--- a/Wrapping/Generators/Python/itk/support/itkLazy.py
+++ b/Wrapping/Generators/Python/itk/support/itkLazy.py
@@ -18,6 +18,12 @@
 import types
 from itk.support import itkBase
 
+# Needed to avoid problem with aliasing of itk.set (itkTemplate)
+# inside the itk namespace.  We need to explictly specify the
+# use of the builtin set
+from builtins import set as _builtin_set
+
+
 not_loaded: str = "not loaded"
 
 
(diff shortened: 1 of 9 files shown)


Wrapping/Generators/Python/itk/support/itkLazy.py
```diff
@@ -51,13 +57,23 @@ def __init__(self, name, lazy_attributes):
...
-            namespace = {}
-            itkBase.itk_load_swig_module(module, namespace)
-            self.loaded_lazy_modules.add(module)
-            for k, v in namespace.items():
-                setattr(self, k, v)
-            value = namespace[attr]
+            with self._module_lock:  # All but one thread will block here.
+                if (
+                    value is not_loaded
+                ):  # Only the first thread needs to run this code, all other blocked threads skip
```
dzenanz, line 63: This comment can be on a separate line, it does not have to be on the same line as the `if` statement. I think that forced the `if` to break into 3 lines.

dzenanz (commented):
Looks great!

hjmjohnson:
@thewtex  This is ready for review.  It fixes the problem with MONAI and multiprocessing parallel loads of data.

The problem was related to the custom lazy loading. 

FYI:  When Python 3.7 becomes the minimum version, there are updated lazy loading mechanisms that are more naturally supported by the language.  

Hans

thewtex (approved):
AWESOME, Hans!

PR #2156, opened 2020-12-08
ENH: Update codebase for Eigen 3.3.8 release

hjmjohnson:
Fixes several bugs related incompatibilities building on
Intel(R) Xeon(R) Gold 6254 CPU @ 3.10GHz with march=native
flags.

This includes changes manually copied in from upstream/master
with commit 655c3a40428e0e6a5490a3a4d4bddceb45af4a3a on
2020-12-07.

with upstream=	https://gitlab.com/libeigen/eigen.git

## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [X] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)

PR #2157, opened 2020-12-08
Update eigen3.3.8

phcerdan:
Update to the top of the 3.3 branch (release 3.3.8)

Changes from last update:

```
git shortlog --no-merges 89449a0821676c5af96cc51973de^..0fd6b4f71dd85b20 | cat
```

```
Alexander Grund (1):
      Make relative path variables of type STRING

Christoph Hertzberg (7):
      Remove .hgignore and copy .gitignore from master branch
      Bug #1788: Fix rule-of-three violations inside the stable modules. This fixes deprecated-copy warnings when compiling with GCC>=9 Also protect some additional Base-constructors from getting called by user code code (#1587)
      Fix some maybe-uninitialized warnings
      Bug #2036 make sure find_standard_math_library_test_program actually compiles (and is guaranteed to call math functions)
      bug #1746: Removed implementation of standard copy-constructor and standard copy-assign-operator from PermutationMatrix and Transpositions to allow malloc-less std::move. Added unit-test to rvalue_types
      Remove unused variable
      Fix doxygen class block that was wrongly named.

David Tellenbach (9):
      Fix StlDeque for GCC 10
      Bump to 3.3.8-rc1
      Fix failure in GEBP kernel when compiling with OpenMP and FMA
      Fix undefined behaviour caused by uncaught exceptions in OMP section
      Bump to 3.3.8
      Define coeff-wise binary array operators for base class
      Mention problems when using potentially throwing scalars and OpenMP
      Rename test/array.cpp to test/array_cwise.cpp
      Bump to 3.3.9

Eugene Zhulenev (1):
      Change typedefs from private to protected to fix MSVC compilation

Florian Maurin (1):
      Fix typo in doc

Gael Guennebaud (11):
      Add missing footer declaration
      remove piwik tracker
      add a banner to advertise the survey
      Fix compilation with AVX512
      Fix compilation with AVX512 and AVX/SSE packet-math functions
      Backport AVX512 implementation from devel branch to 3.3 (the 3.3 version had many issues)
      Fix #1974: assertion when reserving an empty sparse matrix
      Relaxed fastmath unit test: if std::foo fails, then let's only trigger a warning is numext::foo fails too. A true error will triggered only if std::foo works but our numext::foo fails.
      fix #1901: warning in Mode==(Upper|Lower)
      relax number of iterations checks to avoid false negatives
      check two ctors

Janek Kozicki (1):
      Fix Yade high precision Real compilation

Jim Lersch (1):
      Workaround for doxygen class template titles in which the template part of the class signature is lost due to a problem with forward declarations.  The problem is probably caused by doxygen bug #7689. It is confirmed to be fixed in doxygen >= 1.8.19.

Karl Ljungkvist (1):
      Fix typo in Tutorial_BlockOperations_block_assignment.cpp

Luke Peterson (1):
      Remove error counting in OpenMP parallelize_gemm

Martin Vonheim Larsen (1):
      Enable MathJax in Doxygen.in

Rasmus Munk Larsen (1):
      Fix incorrect use of std::abs reported in #1823.

Simon Pfreundschuh (1):
      Replaced call to deprecated 'load' function with appropriate call to 'on'.

Tobias Bosch (1):
      Include <sstream> explicitly, and don't rely on the implicit include via <complex>.

nluehr (1):
      Fix incorrect integer cast in predux<half2>().

szczepaniak bartek (1):
      Fix Paradiso.
```

## PR Checklist

Update triggered by the needs of: https://github.com/InsightSoftwareConsortium/ITK/pull/2156

### DIFF
--- a/Modules/ThirdParty/Eigen3/src/itkeigen/CMakeLists.txt
+++ b/Modules/ThirdParty/Eigen3/src/itkeigen/CMakeLists.txt
@@ -394,22 +394,27 @@ endif()
 
 if(EIGEN_INCLUDE_INSTALL_DIR AND NOT INCLUDE_INSTALL_DIR)
   set(INCLUDE_INSTALL_DIR ${EIGEN_INCLUDE_INSTALL_DIR}
-      CACHE PATH "The directory relative to CMAKE_PREFIX_PATH where Eigen header files are installed")
+      CACHE STRING "The directory relative to CMAKE_PREFIX_PATH where Eigen header files are installed")
 else()
   set(INCLUDE_INSTALL_DIR
       "${CMAKE_INSTALL_INCLUDEDIR}/eigen3"
-      CACHE PATH "The directory relative to CMAKE_PREFIX_PATH where Eigen header files are installed"
+      CACHE STRING "The directory relative to CMAKE_PREFIX_PATH where Eigen header files are installed"
       )
 endif()
 set(CMAKEPACKAGE_INSTALL_DIR
     "${CMAKE_INSTALL_DATADIR}/eigen3/cmake"
-    CACHE PATH "The directory relative to CMAKE_PREFIX_PATH where Eigen3Config.cmake is installed"
+    CACHE STRING "The directory relative to CMAKE_PREFIX_PATH where Eigen3Config.cmake is installed"
     )
 set(PKGCONFIG_INSTALL_DIR
     "${CMAKE_INSTALL_DATADIR}/pkgconfig"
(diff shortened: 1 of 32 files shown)


hjmjohnson:
Downloading locally now.

dzenanz (approved):

phcerdan:
Last force push just to update commit message for the short log.

dzenanz (approved):

hjmjohnson:
Mac failing jpeg test is not related itkJPEG2000Test06

PR #2143, opened 2020-12-03
Update MetaIO

blowekamp:
## PR Checklist
- [ ] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [ ] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [ ] Added test (or behavior not changed)
- [ ] Updated API documentation (or API not changed)
- [ ] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [ ] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [ ] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKExamples) for all new major features (if any)

Refer to the [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) for
further development details if necessary.

PR #2159, opened 2020-12-08
ENH: Test that MetaIO supports a string value of up to 32767 chars

N-Dekker:
Added a field whose value is a string of 32767 characters to the test,
to ensure that the following MetaIO improvement is included:

  "ENH: MetaIO support reading strings of >= 500 chars (up to 32767 chars)"
  Pull request https://github.com/Kitware/MetaIO/pull/93
  Commit https://github.com/Kitware/MetaIO/commit/c9db10ecf0ee2193cfc57f8b01674ca0ef305316

No longer printed the content of a field in case of a correct test
result, as it would produce too much noise with such a large string.


dzenanz (approved):

N-Dekker:
For the record: This test should pass now because pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2143 (Update MetaIO) is merged.

hjmjohnson (approved):

PR #2160, opened 2020-12-08
ENH: add a convenience function WriteImage

dzenanz:
Inspired by #2102 and https://github.com/dzenanz/ITKMorphologicalContourInterpolation/commit/a9ac128f57078cb7c71d1abacd1ce8fce02f1c96.

## PR Checklist
- [x] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [x] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [x] Added test (or behavior not changed)
- [x] Updated API documentation (or API not changed)
- [ ] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [ ] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [ ] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKExamples) for all new major features (if any)

PR #2165, opened 2020-12-09
ENH: remove redundant WriteImage overloads

dzenanz:
This is a follow-up to #2160 which was merged a bit too fast.

### DIFF
--- a/Modules/Core/Common/include/itkMetaProgrammingLibrary.h
+++ b/Modules/Core/Common/include/itkMetaProgrammingLibrary.h
@@ -210,6 +210,23 @@ struct Not : NotC<TF::Value>
   using Type = typename NotC<TF::Value>::Type;
 };
 
+/** MPL type trait to check if type is a SmartPointer.
+ */
+template <typename T>
+struct IsSmartPointer : FalseType
+{};
+
+
+/// \cond SPECIALIZATION_IMPLEMENTATION
+template <typename T>
+struct IsSmartPointer<SmartPointer<T>> : TrueType
+{};
+
+template <typename T>
+struct IsSmartPointer<const SmartPointer<T>> : TrueType
+{};
+/// \endcond
+
 /** MPL relational type trait to check if a static_cast conversion
  * exists.
  *
--- a/Modules/IO/ImageBase/include/itkImageFileWriter.h
+++ b/Modules/IO/ImageBase/include/itkImageFileWriter.h
@@ -22,6 +22,7 @@
 #include "itkProcessObject.h"
 #include "itkImageIOBase.h"
 #include "itkMacro.h"
+#include "itkMetaProgrammingLibrary.h"
 
 namespace itk
 {
@@ -237,61 +238,27 @@ private:
 };
 
 
-/** Convenience function for writing an image. */
-template <typename TImage>
+/** Convenience function for writing an image.
+ *
+ * The image parameter may be a either SmartPointer or a raw pointer and const or non-const.
+ * */
+template <typename TImagePointer>
 ITK_TEMPLATE_EXPORT void
-WriteImage(const TImage * image, const char * filename, bool compress = false)
+WriteImage(TImagePointer && image, const std::string & filename, bool compress = false)
 {
-  using WriterType = ImageFileWriter<TImage>;
-  typename WriterType::Pointer writer = WriterType::New();
+  using NonReferenceImagePointer = typename std::remove_reference<TImagePointer>::type;
+  static_assert(std::is_pointer<NonReferenceImagePointer>::value ||
+                  mpl::IsSmartPointer<NonReferenceImagePointer>::Value,
+                "WriteImage requires a raw pointer or SmartPointer.");
+
+  using ImageType = typename std::remove_const<typename std::remove_reference<decltype(*image)>::type>::type;
+  auto writer = ImageFileWriter<ImageType>::New();
   writer->SetInput(image);
   writer->SetFileName(filename);
   writer->SetUseCompression(compress);
   writer->Update();
 }
-template <typename TImage>
-ITK_TEMPLATE_EXPORT void
-WriteImage(const SmartPointer<const TImage> image, const char * filename, bool compress = false)
-{
-  WriteImage(image.GetPointer(), filename, compress);
-}
-template <typename TImage>
-ITK_TEMPLATE_EXPORT void
-WriteImage(TImage * image, const char * filename, bool compress = false)
-{
-  const TImage * constImage = image;
-  WriteImage(constImage, filename, compress);
-}
-template <typename TImage>
-ITK_TEMPLATE_EXPORT void
-WriteImage(SmartPointer<TImage> image, const char * filename, bool compress = false)
-{
-  WriteImage(image.GetPointer(), filename, compress);
-}
-template <typename TImage>
-ITK_TEMPLATE_EXPORT void
-WriteImage(const TImage * image, const std::string & filename, bool compress = false)
-{
-  WriteImage(image, filename.c_str(), compress);
-}
-template <typename TImage>
-ITK_TEMPLATE_EXPORT void
-WriteImage(const SmartPointer<const TImage> image, const std::string & filename, bool compress = false)
-{
-  WriteImage(image, filename.c_str(), compress);
-}
-template <typename TImage>
-ITK_TEMPLATE_EXPORT void
-WriteImage(TImage * image, const std::string & filename, bool compress = false)
-{
-  WriteImage(image, filename.c_str(), compress);
-}
-template <typename TImage>
-ITK_TEMPLATE_EXPORT void
-WriteImage(SmartPointer<TImage> image, const std::string & filename, bool compress = false)
-{
-  WriteImage(image, filename.c_str(), compress);
-}
+
 } // end namespace itk
 
 #ifndef ITK_MANUAL_INSTANTIATION
--- a/Modules/IO/ImageBase/test/itkWriteImageFunctionGTest.cxx
+++ b/Modules/IO/ImageBase/test/itkWriteImageFunctionGTest.cxx
@@ -1,0 +1,77 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+#include "itkImageFileWriter.h"
+#include "itkImage.h"
+
+#include "itkGTest.h"
+#include "itksys/SystemTools.hxx"
+#include "itkTestDriverIncludeRequiredIOFactories.h"
+
+#define STRING(s) #s
+
+namespace
+{
+
+struct ITKWriteImageFunctionTest : public ::testing::Test
+{
+  void
+  SetUp() override
+  {
+    RegisterRequiredFactories();
+    itksys::SystemTools::ChangeDirectory(STRING(ITK_TEST_OUTPUT_DIR_STR));
+  }
+  using ImageType = itk::Image<float, 2>;
+
+  ImageType::Pointer
+  MakeImage()
+  {
+    auto image = ImageType::New();
+
+    ImageType::RegionType region({ 3, 2 });
+
+    image->SetRegions(region);
+
+    image->Allocate(true);
+    return image;
+  }
+};
+
+
+} // namespace
+
+TEST_F(ITKWriteImageFunctionTest, ImageTypes)
+{
+  ImageType::Pointer image_ptr = MakeImage();
+  itk::WriteImage(image_ptr, "test1.mha");
+  itk::WriteImage(std::move(image_ptr), "test1.mha");
+
+  ImageType::ConstPointer image_cptr = MakeImage();
+  itk::WriteImage(image_cptr, "test1.mha");
+  itk::WriteImage(std::move(image_cptr), "test1.mha");
+
+  const ImageType::ConstPointer image_ccptr = MakeImage();
+  itk::WriteImage(image_ccptr, "test1.mha");
+  itk::WriteImage(std::move(image_ccptr), "test1.mha");
+
+  image_ptr = MakeImage();
+  ImageType * image_rptr = image_ptr.GetPointer();
+  itk::WriteImage(image_rptr, "test1.mha");
+
+  const ImageType * image_crptr = image_ptr.GetPointer();
+  itk::WriteImage(image_crptr, "test1.mha");
+}
--- a/Modules/IO/ImageBase/ITKKWStyleOverwrite.txt
+++ b/Modules/IO/ImageBase/ITKKWStyleOverwrite.txt
@@ -1,1 +1,2 @@
 itkConvertPixelBuffer\.hxx Namespace Disable
+test/itkWriteImageFunctionGTest.cxx Namespace Disable
--- a/Modules/IO/ImageBase/test/CMakeLists.txt
+++ b/Modules/IO/ImageBase/test/CMakeLists.txt
@@ -355,3 +355,11 @@ itk_add_test(NAME itkVectorImageReadWriteTest2
 add_executable(itkUnicodeIOTest itkUnicodeIOTest.cxx)
 itk_module_target_label(itkUnicodeIOTest)
 itk_add_test(NAME itkUnicodeIOTest COMMAND itkUnicodeIOTest)
+
+
+set(ITKIOImageBaseGTests
+        itkWriteImageFunctionGTest.cxx
+        )
+CreateGoogleTestDriver(ITKIOImageBase  "${ITKIOImageBase-Test_LIBRARIES}" "${ITKIOImageBaseGTests}")
+
+target_compile_definitions(ITKIOImageBaseGTestDriver PRIVATE "-DITK_TEST_OUTPUT_DIR=${ITK_TEST_OUTPUT_DIR}")


Modules/IO/ImageBase/include/itkImageFileWriter.h
```diff
@@ -237,10 +237,16 @@ class ITK_TEMPLATE_EXPORT ImageFileWriter : public ProcessObject
 };
 
 
+#if (__cplusplus >= 201703L)
```
hjmjohnson, line 240: This should be in itkMacro.h

N-Dekker:
I would rather postpone the introduction of C++17 `std::string_view`. No need to rush.

N-Dekker:
> I would rather postpone the introduction of C++17 `std::string_view`. No need to rush.

The thing is, I don't have any `std::string_view` experience, don't know if it is always OK to just replace any `const char*` and `const std::string &` parameter by `std::string_view`. Then I wonder whether the convention is to pass `string_view` by value, or by const reference. And if it would be of interest, maybe we could implement a `itk::string_view` class, to be used until ITK is upgraded to C++17... But again, no need to hurry, I think.

dzenanz:
This **convenience** function is not a big reason to go down that hole. And generally, ITK does not deal with strings much - mostly for file names and metadata dictionary.

dzenanz:
Maybe we need the `char *` overloads too: `Modules/IO/Meta/test/itkMetaImageIOMetaDataTest.cxx:265:43: error: call of overloaded 'WriteImage(itk::Image<unsigned char, 2u>::Pointer&, char*&)' is ambiguous`.

dzenanz:
If we do need them, this PR is pointless. Does anyone have a suggestion of doing this with less than 8 overloads?

blowekamp:
> If we do need them, this PR is pointless. Does anyone have a suggestion of doing this with less than 8 overloads?

My initial expectation is that only `WriteImage(const TImage * image, const std::string & filename, bool compress = false)` should be needed. By not having to resolve which function to use implicit conversions should more readily occur with the compiler.

dzenanz:
@blowekamp 1 signature produces:
```text
10>itkImageDuplicatorTest2.cxx
10>C:\Dev\ITK-git\Modules\Core\Common\test\itkImageDuplicatorTest2.cxx(70): error C2672: 'itk::WriteImage': no matching overloaded function found
10>C:\Dev\ITK-git\Modules\Core\Common\test\itkImageDuplicatorTest2.cxx(70): error C2784: 'void itk::WriteImage(const TImage *,const std::string &,bool)': could not deduce template argument for 'const TImage *' from 'itk::SmartPointer<const itk::Image<PixelType,3>>'
10>c:\dev\itk-git\modules\io\imagebase\include\itkImageFileWriter.h(242): note: see declaration of 'itk::WriteImage'
10>itkImageRandomIteratorTest2.cxx
11>ITKCommon2TestDriver.cxx
11>itkSymmetricSecondRankTensorTest.cxx
10>C:\Dev\ITK-git\Modules\Core\Common\test\itkImageRandomIteratorTest2.cxx(85): error C2672: 'itk::WriteImage': no matching overloaded function found
10>C:\Dev\ITK-git\Modules\Core\Common\test\itkImageRandomIteratorTest2.cxx(85): error C2784: 'void itk::WriteImage(const TImage *,const std::string &,bool)': could not deduce template argument for 'const TImage *' from 'itk::SmartPointer<itk::Image<PixelType,2>>'
10>c:\dev\itk-git\modules\io\imagebase\include\itkImageFileWriter.h(242): note: see declaration of 'itk::WriteImage'
10>itkStreamingImageFilterTest3.cxx
12>ITKCommonGTestDriver.vcxproj -> C:\Dev\ITK-git-2015\bin\RelWithDebInfo\ITKCommonGTestDriver.exe
12>ITKCommonGTestDriver.vcxproj -> C:/Dev/ITK-git-2015/bin/RelWithDebInfo/ITKCommonGTestDriver.pdb (Full PDB)
11>Generating Code...
10>itkSymmetricSecondRankTensorImageReadTest.cxx
10>C:\Dev\ITK-git\Modules\Core\Common\test\itkSymmetricSecondRankTensorImageReadTest.cxx(87): error C2672: 'itk::WriteImage': no matching overloaded function found
10>C:\Dev\ITK-git\Modules\Core\Common\test\itkSymmetricSecondRankTensorImageReadTest.cxx(87): error C2784: 'void itk::WriteImage(const TImage *,const std::string &,bool)': could not deduce template argument for 'const TImage *' from 'itk::SmartPointer<itk::Image<MatrixPixelType,3>>'
10>c:\dev\itk-git\modules\io\imagebase\include\itkImageFileWriter.h(242): note: see declaration of 'itk::WriteImage'
10>itkSymmetricSecondRankTensorImageWriteReadTest.cxx
11>ITKCommon2TestDriver.vcxproj -> C:\Dev\ITK-git-2015\bin\RelWithDebInfo\ITKCommon2TestDriver.exe
11>ITKCommon2TestDriver.vcxproj -> C:/Dev/ITK-git-2015/bin/RelWithDebInfo/ITKCommon2TestDriver.pdb (Full PDB)
10>C:\Dev\ITK-git\Modules\Core\Common\test\itkSymmetricSecondRankTensorImageWriteReadTest.cxx(72): error C2672: 'itk::WriteImage': no matching overloaded function found
10>C:\Dev\ITK-git\Modules\Core\Common\test\itkSymmetricSecondRankTensorImageWriteReadTest.cxx(72): error C2784: 'void itk::WriteImage(const TImage *,const std::string &,bool)': could not deduce template argument for 'const TImage *' from 'itk::SmartPointer<itk::Image<TensorPixelType,2>>'
10>c:\dev\itk-git\modules\io\imagebase\include\itkImageFileWriter.h(242): note: see declaration of 'itk::WriteImage'
```
If I also add `WriteImage(const SmartPointer<const TImage> image, const std::string & filename, bool compress = false)`:
```text
4>itkImageRandomIteratorTest2.cxx
4>C:\Dev\ITK-git\Modules\Core\Common\test\itkImageRandomIteratorTest2.cxx(85): error C2672: 'itk::WriteImage': no matching overloaded function found
4>C:\Dev\ITK-git\Modules\Core\Common\test\itkImageRandomIteratorTest2.cxx(85): error C2784: 'void itk::WriteImage(const itk::SmartPointer<const TImage>,const std::string &,bool)': could not deduce template argument for 'const itk::SmartPointer<const TImage>' from 'itk::SmartPointer<itk::Image<PixelType,2>>'
4>c:\dev\itk-git\modules\io\imagebase\include\itkImageFileWriter.h(253): note: see declaration of 'itk::WriteImage'
4>C:\Dev\ITK-git\Modules\Core\Common\test\itkImageRandomIteratorTest2.cxx(85): error C2784: 'void itk::WriteImage(const TImage *,const std::string &,bool)': could not deduce template argument for 'const TImage *' from 'itk::SmartPointer<itk::Image<PixelType,2>>'
4>c:\dev\itk-git\modules\io\imagebase\include\itkImageFileWriter.h(242): note: see declaration of 'itk::WriteImage'
4>itkSymmetricSecondRankTensorImageReadTest.cxx
4>C:\Dev\ITK-git\Modules\Core\Common\test\itkSymmetricSecondRankTensorImageReadTest.cxx(87): error C2672: 'itk::WriteImage': no matching overloaded function found
4>C:\Dev\ITK-git\Modules\Core\Common\test\itkSymmetricSecondRankTensorImageReadTest.cxx(87): error C2784: 'void itk::WriteImage(const itk::SmartPointer<const TImage>,const std::string &,bool)': could not deduce template argument for 'const itk::SmartPointer<const TImage>' from 'itk::SmartPointer<itk::Image<MatrixPixelType,3>>'
4>c:\dev\itk-git\modules\io\imagebase\include\itkImageFileWriter.h(253): note: see declaration of 'itk::WriteImage'
4>C:\Dev\ITK-git\Modules\Core\Common\test\itkSymmetricSecondRankTensorImageReadTest.cxx(87): error C2784: 'void itk::WriteImage(const TImage *,const std::string &,bool)': could not deduce template argument for 'const TImage *' from 'itk::SmartPointer<itk::Image<MatrixPixelType,3>>'
4>c:\dev\itk-git\modules\io\imagebase\include\itkImageFileWriter.h(242): note: see declaration of 'itk::WriteImage'
4>itkSymmetricSecondRankTensorImageWriteReadTest.cxx
4>C:\Dev\ITK-git\Modules\Core\Common\test\itkSymmetricSecondRankTensorImageWriteReadTest.cxx(72): error C2672: 'itk::WriteImage': no matching overloaded function found
4>C:\Dev\ITK-git\Modules\Core\Common\test\itkSymmetricSecondRankTensorImageWriteReadTest.cxx(72): error C2784: 'void itk::WriteImage(const itk::SmartPointer<const TImage>,const std::string &,bool)': could not deduce template argument for 'const itk::SmartPointer<const TImage>' from 'itk::SmartPointer<itk::Image<TensorPixelType,2>>'
4>c:\dev\itk-git\modules\io\imagebase\include\itkImageFileWriter.h(253): note: see declaration of 'itk::WriteImage'
4>C:\Dev\ITK-git\Modules\Core\Common\test\itkSymmetricSecondRankTensorImageWriteReadTest.cxx(72): error C2784: 'void itk::WriteImage(const TImage *,const std::string &,bool)': could not deduce template argument for 'const TImage *' from 'itk::SmartPointer<itk::Image<TensorPixelType,2>>'
4>c:\dev\itk-git\modules\io\imagebase\include\itkImageFileWriter.h(242): note: see declaration of 'itk::WriteImage'
```

dzenanz:
I will leave this open for a while, in case someone gets some brilliant ideas. But I am essentially abandoning this PR, and when I see this after a week has passed I will formally close it (assuming no new ideas).

N-Dekker:
Honestly I'm still in favor of passing the image by reference, instead of by smart/raw pointer. None of the overloads would have been necessary then.

dzenanz:
> passing the image by reference

That just places the burden on the user. It becomes annoying really quickly.

N-Dekker:
> That just places the burden on the user. It becomes annoying really quickly.

Then why, in your opinion, did C++ introduce pass by (const) reference for parameters?

N-Dekker:
Sorry for accidentally editing your comment @dzenanz !

dzenanz:
ITK was designed in the C++98 world, before `shared_ptr` was part of standard library. So our syntax does not mesh perfectly with C++11. ITK uses `SmartPointer`s for images exclusively, and trying to work around just makes your life harder.

Modules/IO/ImageBase/include/itkImageFileWriter.h
```diff
@@ -236,11 +236,10 @@ class ITK_TEMPLATE_EXPORT ImageFileWriter : public ProcessObject
...
 
-
 /** Convenience function for writing an image. */
 template <typename TImage>
 ITK_TEMPLATE_EXPORT void
-WriteImage(const TImage * image, const char * filename, bool compress = false)
+WriteImage(const TImage * image, const std::string & filename, bool compress = false)
 {
   using WriterType = ImageFileWriter<TImage>;
   typename WriterType::Pointer writer = WriterType::New();
```
N-Dekker, line 245: It would be nice to use C++11 auto here: `const auto writer = WriterType::New();`

dzenanz: I tried that (based on your original PR), and I remember there were compile errors.

N-Dekker: Strange! Do you remember what compile errors?

dzenanz:
`const auto writer = ImageFileWriter<TImage>::New();` results in:
```text
4>itkImageRandomIteratorTest2.cxx
4>C:\Dev\ITK-git\Modules\Core\Common\test\itkImageRandomIteratorTest2.cxx(85): error C2672: 'itk::WriteImage': no matching overloaded function found
4>C:\Dev\ITK-git\Modules\Core\Common\test\itkImageRandomIteratorTest2.cxx(85): error C2784: 'void itk::WriteImage(const itk::SmartPointer<const TImage>,const std::string &,bool)': could not deduce template argument for 'const itk::SmartPointer<const TImage>' from 'itk::SmartPointer<itk::Image<PixelType,2>>'
4>c:\dev\itk-git\modules\io\imagebase\include\itkImageFileWriter.h(252): note: see declaration of 'itk::WriteImage'
4>C:\Dev\ITK-git\Modules\Core\Common\test\itkImageRandomIteratorTest2.cxx(85): error C2784: 'void itk::WriteImage(const TImage *,const std::string &,bool)': could not deduce template argument for 'const TImage *' from 'itk::SmartPointer<itk::Image<PixelType,2>>'
4>c:\dev\itk-git\modules\io\imagebase\include\itkImageFileWriter.h(242): note: see declaration of 'itk::WriteImage'
4>itkSymmetricSecondRankTensorImageReadTest.cxx
4>C:\Dev\ITK-git\Modules\Core\Common\test\itkSymmetricSecondRankTensorImageReadTest.cxx(87): error C2672: 'itk::WriteImage': no matching overloaded function found
4>C:\Dev\ITK-git\Modules\Core\Common\test\itkSymmetricSecondRankTensorImageReadTest.cxx(87): error C2784: 'void itk::WriteImage(const itk::SmartPointer<const TImage>,const std::string &,bool)': could not deduce template argument for 'const itk::SmartPointer<const TImage>' from 'itk::SmartPointer<itk::Image<MatrixPixelType,3>>'
4>c:\dev\itk-git\modules\io\imagebase\include\itkImageFileWriter.h(252): note: see declaration of 'itk::WriteImage'
4>C:\Dev\ITK-git\Modules\Core\Common\test\itkSymmetricSecondRankTensorImageReadTest.cxx(87): error C2784: 'void itk::WriteImage(const TImage *,const std::string &,bool)': could not deduce template argument for 'const TImage *' from 'itk::SmartPointer<itk::Image<MatrixPixelType,3>>'
4>c:\dev\itk-git\modules\io\imagebase\include\itkImageFileWriter.h(242): note: see declaration of 'itk::WriteImage'
4>itkSymmetricSecondRankTensorImageWriteReadTest.cxx
4>C:\Dev\ITK-git\Modules\Core\Common\test\itkSymmetricSecondRankTensorImageWriteReadTest.cxx(72): error C2672: 'itk::WriteImage': no matching overloaded function found
4>C:\Dev\ITK-git\Modules\Core\Common\test\itkSymmetricSecondRankTensorImageWriteReadTest.cxx(72): error C2784: 'void itk::WriteImage(const itk::SmartPointer<const TImage>,const std::string &,bool)': could not deduce template argument for 'const itk::SmartPointer<const TImage>' from 'itk::SmartPointer<itk::Image<TensorPixelType,2>>'
4>c:\dev\itk-git\modules\io\imagebase\include\itkImageFileWriter.h(252): note: see declaration of 'itk::WriteImage'
4>C:\Dev\ITK-git\Modules\Core\Common\test\itkSymmetricSecondRankTensorImageWriteReadTest.cxx(72): error C2784: 'void itk::WriteImage(const TImage *,const std::string &,bool)': could not deduce template argument for 'const TImage *' from 'itk::SmartPointer<itk::Image<TensorPixelType,2>>'
4>c:\dev\itk-git\modules\io\imagebase\include\itkImageFileWriter.h(242): note: see declaration of 'itk::WriteImage'
```

blowekamp:
I am working on a patch with only the following signature:

```
WriteImage(const SmartPointer<const TImage> image, const std::string &filename, bool compress = false)
```

It is also nessecatating adding the a couple implicit conversion methods for the SmartPointer class:

```
  template <typename T, typename = typename EnableIfConst<T>::type>
  SmartPointer(const SmartPointer<typename std::remove_const<T>::type> & p) noexcept
```

blowekamp:
I believe some additional conversion with the std::SmartPoitner with const/non-const should help give the convenience of implicit conversion here. Alternatively, we could just a `.GetPointer()` to the call the `WriteImage`.

In the convenience vs verbosity, the traditionally chooses to be clear and verbose.

dzenanz:
Most images we want saved are held in a smart pointer, I believe that signature is more important. Also, since smart pointer has a constructor from raw pointer, there should be an automatic conversion. Sadly, compilers fail to do it.

blowekamp:
> Most images we want saved are held in a smart pointer, I believe that signature is more important.

I believe you may mean the usage is important, not the signature.

This one function seems to work:

```template <typename TImage>
ITK_TEMPLATE_EXPORT void
WriteImage(TImage image, const std::string &filename, bool compress = false)
{
  static_assert(std::is_pointer<TImage>::value || mpl::IsSmartPointer<TImage>::Value,
                "WriteImage requires a raw pointer or SmartPointer.");
  using  ImagePointerType = typename std::conditional< std::is_pointer<TImage>::value,
    SmartPointer<const typename std::remove_pointer<TImage>::type>,
    const TImage>::type;

  using ImageType = typename std::remove_const<typename ImagePointerType::ObjectType>::type;

  ImagePointerType ptr = image;

  auto writer = ImageFileWriter<ImageType>::New();
  writer->SetInput(ptr);
  writer->SetFileName(filename);
  writer->SetUseCompression(compress);
  writer->Update();
}
```

This might be useful too:
```

template <typename T>
struct IsSmartPointer: FalseType
{};


template <typename T>
struct IsSmartPointer<SmartPointer<T>>: TrueType
{};

```

The in function meta-programming could be refactored into a meta-struct, something called like "MakeSmartPointer".

@N-Dekker I bet you could make it work with references too... but that may get little to messy :)

dzenanz:
@blowekamp you are free to take over this PR or create a new one. I believe you should have the right to push/amend this branch in my fork.

Modules/IO/ImageBase/include/itkImageFileWriter.h
```diff
@@ -240,58 +241,26 @@ class ITK_TEMPLATE_EXPORT ImageFileWriter : public ProcessObject
 /** Convenience function for writing an image. */
 template <typename TImage>
 ITK_TEMPLATE_EXPORT void
-WriteImage(const TImage * image, const char * filename, bool compress = false)
+WriteImage(TImage image, const std::string & filename, bool compress = false)
```
N-Dekker, line 244: Maybe rename `TImage` to `TImagePointer`? (Usually `TImage` itself is not a pointer type).

blowekamp: Yes, I have that locally now :)

I am also experimenting with the following signature:
```
WriteImage(TImagePointer &image, const std::string & filename, bool compress = false)
```

So that if TImagePointer is a `SmartPointer` then a copy will not be made.

N-Dekker: > that if TImagePointer is a SmartPointer then a copy will not be made.

You might consider `TImagePointer &&`, if you really want to. I think the single ampersand version won't support passing a temporary pointer (rvalue) returned by a function.

blowekamp: > > that if TImagePointer is a SmartPointer then a copy will not be made.
> 
> You might consider `TImagePointer &&`, if you really want to. I think the single ampersand version won't support passing a temporary pointer (rvalue) returned by a function.

and then I need to remove the the reference...

N-Dekker: Nitpick: This PR looks like a style improvement to me, so "STYLE", rather than "ENH". So I would suggest adjusting the subject line of the commit message

blowekamp: The intent of the original code is still here so not much of an usage change, however this PR now adds substantial new meta-programming which should be considered and enhancement.

My guideline for STYLE vs ENH: When looking for a bug or a change in behavior can the commits with STYLE/DOC should be able to be skipped as no change in logic should occur.

Modules/IO/ImageBase/include/itkImageFileWriter.h
```diff
@@ -240,58 +241,26 @@ class ITK_TEMPLATE_EXPORT ImageFileWriter : public ProcessObject
...
-  using WriterType = ImageFileWriter<TImage>;
-  typename WriterType::Pointer writer = WriterType::New();
+  static_assert(std::is_pointer<TImage>::value || mpl::IsSmartPointer<TImage>::Value,
+                "WriteImage requires a raw pointer or SmartPointer.");
+
+  using ImagePointerType = typename std::conditional<std::is_pointer<TImage>::value,
+                                                     SmartPointer<const typename std::remove_pointer<TImage>::type>,
+                                                     const TImage>::type;
+
+  using ImageType = typename std::remove_const<typename ImagePointerType::ObjectType>::type;
```
N-Dekker, line 253: You might have retrieved the ImageType directly from the image parameter by `decltype(*image)`. Just a suggestion! You'd probably still need remove_const, and also `std::remove_reference`!

Modules/IO/ImageBase/test/itkWriteImageFunctionGTest.cxx
```diff
@@ -0,0 +1,71 @@
...
+{
+  ImageType::Pointer image_ptr = MakeImage();
+  itk::WriteImage(image_ptr, "test1.mha");
+  itk::WriteImage(std::move(image_ptr), "test1.mha");
+
+  ImageType::ConstPointer image_cptr = MakeImage();
+  itk::WriteImage(image_cptr, "test1.mha");
+  itk::WriteImage(std::move(image_cptr), "test1.mha");
+
+  image_ptr = MakeImage();
```
dzenanz, line 65: I don't think this reinitialization is necessary.

Modules/Core/Common/include/itkMetaProgrammingLibrary.h
```diff
@@ -210,6 +210,19 @@ struct Not : NotC<TF::Value>
...
 };
 
+/** MPL type trait to check if type is a SmartPointer.
+ */
+template <typename T>
+struct IsSmartPointer : FalseType
+{};
+
+
+/// \cond SPECIALIZATION_IMPLEMENTATION
```
dzenanz, line 220: What is this for?

dzenanz (commented):
Looks great! Awesome that you added a specific test.

Modules/IO/ImageBase/test/itkWriteImageFunctionGTest.cxx
blowekamp: Following best practices, after `std::move` on an object is should not be used again. It's state is unspecified.

Modules/Core/Common/include/itkMetaProgrammingLibrary.h
blowekamp: Only the non-specialized class should show up in doxygen, so this removes the specialization for Doxygen.

dzenanz:
I can't formally approve my own PR, but I approve!

Modules/IO/ImageBase/test/itkWriteImageFunctionGTest.cxx
```diff
@@ -0,0 +1,71 @@
...
+  ImageType * image_rptr = image_ptr.GetPointer();
+  itk::WriteImage(image_rptr, "test1.mha");
+
+  ImageType * image_crptr = image_ptr.GetPointer();
```
dzenanz, line 69: Shouldn't here be a `const ImageType * image_crptr = image_ptr.GetConstPointer();` or something else involving `const` qualifier?

Modules/IO/ImageBase/test/itkWriteImageFunctionGTest.cxx
```diff
@@ -0,0 +1,71 @@
...
+  void
+  SetUp() override
+  {
+    RegisterRequiredFactories();
+    itksys::SystemTools::ChangeDirectory(STRING(ITK_TEST_OUTPUT_DIR_STR));
+  }
+};
+
+using ImageType = itk::Image<float, 2>;
+typename ImageType::Pointer
```
N-Dekker, line 40: `typename` appears unnecessary here, because it isn't part of a template, right?

Modules/IO/ImageBase/test/itkWriteImageFunctionGTest.cxx
```diff
@@ -0,0 +1,71 @@
...
+MakeImage()
+{
+  auto image = ImageType::New();
+
+  typename ImageType::RegionType region({ 100, 99 });
```
N-Dekker, line 45: Same here: `typename` seems unnecessary.

Modules/IO/ImageBase/test/itkWriteImageFunctionGTest.cxx
N-Dekker, line 69: Did you mean `image_cptr`  (for _const_ smart pointer) here, instead of `image_ptr`?

dzenanz: const raw pointer, I believe

Modules/IO/ImageBase/test/itkWriteImageFunctionGTest.cxx
N-Dekker: Not a big issue, but a size of 100 x 99 x 4 bytes is larger than necessary for this test. In order to minimize use of resources, you might as well do a 2x2 image: `{ 2, 2 }`.

Modules/IO/ImageBase/test/itkWriteImageFunctionGTest.cxx
N-Dekker: The `std::move` test also appears useful to me, to ensure that an r-value reference as first argument is properly supported. Especially because it might not have been supported if the parameter would have been declared as an l-value reference, `TImagePointer &` (with just a single ampersand).

Modules/IO/ImageBase/test/itkWriteImageFunctionGTest.cxx
```diff
@@ -0,0 +1,71 @@
...
+
+  image->SetRegions(region);
+
+  image->Allocate();
```
N-Dekker, line 49: BUG: By default, `image->Allocate()` does not initialize its pixels. Writing such an image might cause undefined behavior 😱  

Suggestion: `image->Allocate( true )`

Modules/IO/ImageBase/test/itkWriteImageFunctionGTest.cxx
blowekamp: yes, done

N-Dekker:
Thanks for addressing my comments about the new test that you added!

I think the test is still somewhat limited, as it only tests that it is possible to call `WriteImage`. It does not test the effect of calling the function! Ideally you would also test that `WriteImage` actually writes an image to a file, by checking if the image can be read back the image from the written file. But I think such an extra test could be added later, once we also have a simple `itk::ReadImage` function  😄   Pseudo code:

    itk::WriteImage(image, fileName);
    ASSERT_EQ(itk::ReadImage(file), image);

Anyway, that would be a nice future extension, in my opinion.

N-Dekker (approved):
I'm very glad that all (seven!) redundant WriteImage overloads can be removed this way  😄

blowekamp:
@N-Dekker That you for your comments and suggestions. This problem seemed rather simple to begin with but with the requirements of implicit template parameters, and support for numerous input proved challenging due to implicit conversion not working with the implicit template parameters. What we have finally iterated towards is a new ( to me at least ), flexible and maintainable way to implement functions taking itk::Image pointers.

> itk::WriteImage(image, fileName);
> ASSERT_EQ(itk::ReadImage(file), image);

I don't believe that we have this functionality with ITK and the GTest framework and it would likely need some measures for tolerance for pixels and metadata.

Modules/IO/ImageBase/test/itkWriteImageFunctionGTest.cxx
```diff
@@ -0,0 +1,76 @@
...
+
+  const ImageType::ConstPointer image_ccptr = MakeImage();
+  itk::WriteImage(image_ccptr, "test1.mha");
+  itk::WriteImage(std::move(image_ccptr), "test1.mha");
+
+  image_ptr = MakeImage();
+  ImageType * image_rptr = image_ptr.GetPointer();
+  itk::WriteImage(image_rptr, "test1.mha");
+
+  const ImageType * image_crptr = image_cptr.GetPointer();
```
dzenanz, line 74: image_cptr is moved on line 64, so a new MakeImage is needed or a different variable on the right side

blowekamp:
I am seeing the following in testing errors:
```
/Users/runner/work/1/s/Modules/IO/ImageBase/test/itkWriteImageFunctionGTest.cxx:26: error: namespace is wrong
```

I am using an anonymous name space for file local class declarations. I guess this is no longer allowed. What is the preferred way?

dzenanz:
> What is the preferred way?

I don't know right off the bat. I would have to search, maybe see other GTests.

dzenanz (commented):
Approved

N-Dekker:
Why did you change `class` to `struct`?

hjmjohnson (approved):
Very Nice.

PR #2170, opened 2020-12-11
BUG: DICOMParser should use `&&` instead of comma in for-loop condition

N-Dekker:
Fixed the following warning from Visual C++ 2019:

> DICOMParser.cxx(830,14): warning C4834: discarding return value of
> function with 'nodiscard' attribute

Potentially a serious bug.


hjmjohnson (approved):
NICE CATCH!!

N-Dekker:
> NICE CATCH!!

Thanks @hjmjohnson  Would you possibly have time to get the fix "upstream"?

Honestly, I didn't seriously get bitten by the bug. While it might cause runtime trouble, it's just a compilation warning to me  😃

thewtex (approved):

dzenanz (approved):

N-Dekker:
Thanks for merging @hjmjohnson For the record, it looks like the bug has been introduced with 74720a64a399f1b49ffac1ce4a51e5d4d3439cc6 31 Mar 2003, "Code cleanup.  Fixed some memory leaks.  Added support for group/element introspection."

Is https://sourceforge.net/projects/dicomparser still the upstream?

PR #2158, opened 2020-12-08
ENH: Hdf5 update

hjmjohnson:
Update to most recent tagged version of HDF5 (hdf5-1_10_6) that is compatible with cmake 3.10.2.

Replace customizations for ITK that are now included in the upstream version of HDF5.

## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [X] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)

PR #2171, opened 2020-12-11
ENH: Ignore huge file warning for HDF5 paths

hjmjohnson:
The ghostflow-check-master check failure needs to be
ignored for ThirdParty HDF5 large files.


hjmjohnson:
@dzenanz I think that the ghost-flow check gets the ignore list from the master branch.   Ths PR needs to be merged before the #2158 so that the ghost flow check passes.

thewtex (approved):

dzenanz (approved):

PR #2183, opened 2020-12-16
BUG: Fix Euler3DTransform::SetFixedParameters crash when too few params

N-Dekker:
When the argument passed to `Euler3DTransform` member function
`SetFixedParameters` would contain less than `InputSpaceDimension`
values, the member function would have undefined behavior, which would
typically just cause a crash.

A crash could occur when reading a transform from a transform file, when
this file has an insufficient number of fixed parameters. It was
encountered when loading a transform within 3DSlicer (Preview Release
version 4.13.0-2020-12-15).

The fix consists of throwing an exception when the number of fixed
parameters is too small. A GoogleTest unit test is included.

Follow-up to:
"BUG: Fix MatrixOffsetTransformBase SetFixedParameters if too few params"
Pull request: https://github.com/InsightSoftwareConsortium/ITK/pull/2118
Commit: fcb85b3a1e4f78d3c9c3742f213e7acefb87314f

### DIFF
--- a/Modules/Core/Transform/include/itkEuler3DTransform.hxx
+++ b/Modules/Core/Transform/include/itkEuler3DTransform.hxx
@@ -135,6 +135,12 @@ template <typename TParametersValueType>
 void
 Euler3DTransform<TParametersValueType>::SetFixedParameters(const FixedParametersType & parameters)
 {
+  if (parameters.size() < InputSpaceDimension)
+  {
+    itkExceptionMacro(<< "Error setting fixed parameters: parameters array size (" << parameters.size()
+                      << ") is less than expected  (InputSpaceDimension = " << InputSpaceDimension << ")");
+  }
+
   InputPointType c;
   for (unsigned int i = 0; i < InputSpaceDimension; i++)
   {
--- a/Modules/Core/Transform/test/CMakeLists.txt
+++ b/Modules/Core/Transform/test/CMakeLists.txt
@@ -181,6 +181,7 @@ set_tests_properties( itkTestTransformGetInverse PROPERTIES COST 50 )
 
 set(ITKTransformGTests
   itkBSplineTransformGTest.cxx
+  itkEuler3DTransformGTest.cxx
   itkMatrixOffsetTransformBaseGTest.cxx
 )
 CreateGoogleTestDriver(ITKTransform "${ITKTransform-Test_LIBRARIES}" "${ITKTransformGTests}")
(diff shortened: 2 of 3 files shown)


N-Dekker:
@lassoan Hi Andras, hereby a followup to pull request: #2118, that I submitted with 3D Slicer in mind  😃 I'm very glad that my previous fix fcb85b3a1e4f78d3c9c3742f213e7acefb87314f was included so quickly with 3D Slicer. My current pull request addresses a similar problem, but now for `Euler3DTransform`.  Hope you can have a look!

Modules/Core/Transform/include/itkEuler3DTransform.hxx
```diff
@@ -135,6 +135,12 @@ template <typename TParametersValueType>
 void
 Euler3DTransform<TParametersValueType>::SetFixedParameters(const FixedParametersType & parameters)
 {
+  if (parameters.size() < InputSpaceDimension)
```
hjmjohnson, line 138: Should this be "  != " rather than " < " ? 

along with a change to the comment below.

blowekamp: Please see line 147/153. This transform supports legacy transform with a missing parameter. 

Additionally, the "<" only checks what is required here. A derived transform could support more fixed parameters.

N-Dekker: > Should this be " != " rather than " < " ?

@hjmjohnson Good question, thanks Hans. I do have an answer, though  😃  First of all, the aim of the commit is to fix a crash. At https://github.com/InsightSoftwareConsortium/ITK/blob/v5.1.2/Modules/Core/Transform/include/itkEuler3DTransform.hxx#L140-L142

    for (unsigned int i = 0; i < InputSpaceDimension; i++)
    {
      c[i] = this->m_FixedParameters[i] = parameters[i];  // Crashes when m_FixedParameters.size() is less than three.
    }

Such a crash _only_ occurs for m_FixedParameters.size() < 3.

Secondly, `m_FixedParameters.size()` may be three or four. Both sizes are supported. I can imagine that in the future, more, larger sizes will be supported. (Extra, optional fixed parameters.) But less than three is always an error!  OK?

N-Dekker: Sorry @blowekamp I just see you already answered the question from Hans.

N-Dekker: @blowekamp There appears to be a strong agreement between the two answers of us  😄

hjmjohnson: Thank you for the explanation.

hjmjohnson (approved):

thewtex (approved):

PR #2184, opened 2020-12-16
ENH: Add convenience function ReadImage

N-Dekker:
Added `itk::ReadImage<TOutputImage>(const std::string & filename)`, to
allows simplifying the C++ code for reading an image.

Suggested by Dženan Zukić (@dzenanz) at
https://github.com/InsightSoftwareConsortium/ITK/pull/2102#issuecomment-723232937

Follow-up to:
"ENH: add a convenience function WriteImage"
Pull request: #2160
Commit: 60b0984d4196841844424e33963e161f3f0c709d

Replaced manual creation of `ImageFileReader` objects in Common tests by
calls to the new `itk::ReadImage` function. Also removed related
redundant try-catch blocks from SymmetricSecondRankTensorImage tests.

### DIFF
--- a/Modules/IO/ImageBase/include/itkImageFileReader.h
+++ b/Modules/IO/ImageBase/include/itkImageFileReader.h
@@ -45,7 +45,7 @@ namespace itk
  * filter. If data stored in the file is stored in a different format
  * then specified by TOutputImage, than this filter converts data
  * between the file type and the external expected type.  The
- * ConvertTraits template argument is used to do the conversion.
+ * `ConvertPixelTraits` template parameter is used to do the conversion.
  *
  * A Pluggable factory pattern is used this allows different kinds of readers
  * to be registered (even at run time) without having to modify the
@@ -167,6 +167,29 @@ private:
   // produce the requested region.
   ImageIORegion m_ActualIORegion;
 };
+
+
+/** Convenience function for reading an image.
+ *
+ * `TOutputImage` is the expected output image type, and the optional
+ * `ConvertPixelTraits` template parameter is used to do the conversion,
+ * as specified by ImageFileReader.
+ *
+ * The function reads the image from the specified file, and returns the
+ * image that it has read.
+ * */
+template <typename TOutputImage,
+          typename ConvertPixelTraits = DefaultConvertPixelTraits<typename TOutputImage::IOPixelType>>
+typename TOutputImage::Pointer
+ReadImage(const std::string & filename)
+{
+  const auto reader = ImageFileReader<TOutputImage, ConvertPixelTraits>::New();
+  reader->SetFileName(filename);
+  reader->Update();
+  return reader->GetOutput();
+}
+
+
 } // namespace itk
 
 #ifndef ITK_MANUAL_INSTANTIATION
--- a/Modules/Core/Common/test/itkImageDuplicatorTest2.cxx
+++ b/Modules/Core/Common/test/itkImageDuplicatorTest2.cxx
@@ -35,19 +35,14 @@ itkImageDuplicatorTest2(int argc, char * argv[])
   constexpr unsigned int Dimension = 3;
   using ImageType = itk::Image<PixelType, Dimension>;
 
-  using ReaderType = itk::ImageFileReader<ImageType>;
-  ReaderType::Pointer reader = ReaderType::New();
   using DuplicatorType = itk::ImageDuplicator<ImageType>;
   DuplicatorType::Pointer dup = DuplicatorType::New();
   using AbsType = itk::AbsImageFilter<ImageType, ImageType>;
   AbsType::Pointer absF = AbsType::New();
 
-  reader->SetFileName(argv[1]);
-
   try
   {
-    reader->Update();
-    ImageType::Pointer inImage = reader->GetOutput();
+    const auto inImage = itk::ReadImage<ImageType>(argv[1]);
 
     ImageType::RegionType lpr = inImage->GetLargestPossibleRegion();
     ImageType::RegionType region = lpr;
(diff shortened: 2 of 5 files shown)


N-Dekker:
Review suggestion: ignore white space when comparing the files: https://github.com/InsightSoftwareConsortium/ITK/pull/2184/files?w=1

hjmjohnson:
Nice!  LGTM, but I think @dzenanz should have a look before merging.

N-Dekker:
@hjmjohnson  Thank you for your approval, Hans. An alternative could be to add the function as a `static`member function to `ImageFileReader`. In order to support a call like:

     auto image = itk::ImageFileReader<TOutputImage>::ReadImage(filename);

That would also be OK to me. But I think people prefer to just call `itk::ReadImage<TOutputImage>(filename)`  😃

dzenanz (approved):

dzenanz:
@thewtex we should definitely mention these convenience functions in 5.2.0 release notes.

Modules/Core/Common/test/itkStreamingImageFilterTest3.cxx
```diff
@@ -42,13 +42,9 @@ itkStreamingImageFilterTest3(int argc, char * argv[])
...
   using ImageType = itk::Image<PixelType, 2>;
 
-  using ReaderType = itk::ImageFileReader<ImageType>;
-  ReaderType::Pointer reader = ReaderType::New();
-  reader->SetFileName(inputFilename);
-
   using SomeFilter = itk::ShiftScaleImageFilter<ImageType, ImageType>;
   SomeFilter::Pointer filter = SomeFilter::New();
-  filter->SetInput(reader->GetOutput());
+  filter->SetInput(itk::ReadImage<ImageType>(inputFilename));
```
blowekamp, line 47: This is a change in the behavior of this test. This test is called "streaming" because it test streaming. This change update the whole image from a file then runs the filter.

Modules/IO/ImageBase/include/itkImageFileReader.h
```diff
@@ -167,6 +167,21 @@ class ITK_TEMPLATE_EXPORT ImageFileReader : public ImageSource<TOutputImage>
   // produce the requested region.
   ImageIORegion m_ActualIORegion;
 };
+
+
+/** Convenience function for reading an image. */
```
blowekamp, line 172: This description seems short.

A description should include that the function updated the whole image. Also perhaps a description of the template parameters or the behaviors of the pixel trains template parameter would be good.

blowekamp (changes requested):
The behavior to the streaming test changes with this PR.

Modules/Core/Common/test/itkStreamingImageFilterTest3.cxx
N-Dekker: OK, thanks Bradley, but would that be a problem? The test still passes successfully, right?

Modules/IO/ImageBase/include/itkImageFileReader.h
N-Dekker: > A description should include that the function updated the whole image.

Thanks again @blowekamp but when a user does `auto image = itk::ReadImage(fileName)`, it seems obvious to me that the whole image is updated. A user cannot do something like `image->Update()` afterwards anyway. OK?

I do agree with you that a description of the template parameter `ConvertPixelTraits` would be good. I'll add one with the next force-push. 😃

Modules/Core/Common/test/itkStreamingImageFilterTest3.cxx
dzenanz: The test passes successfully, but its whole point is to test streaming. I did not catch this in my review.

Modules/IO/ImageBase/include/itkImageFileReader.h
blowekamp: I may be wrong, but I thought one of the ITK python procedural methods return objects ( Image or Filter?) that still needed to be updated?

I think executing `image->Update()` afterwords would still work. While its not this case, in others, this would have the side effect of setting the LargestPossibleRegion to the BufferedRegion.

Modules/Core/Common/test/itkStreamingImageFilterTest3.cxx
N-Dekker: OK, I will remove the proposed itkStreamingImageFilterTest3 changes from this PR, with the next force-push. Thanks @blowekamp and @dzenanz

Modules/IO/ImageBase/include/itkImageFileReader.h
N-Dekker: OK, I did not even know that `image->Update()` would work. But then again, isn't it obvious to you that the image returned by `ReadImage` is the image that was just read?

dzenanz: It is obvious. The docstring is for people who are having some doubts, maybe because they know streaming exists in ITK 😄 and want to check whether this convenience function participates.

Modules/Core/Common/test/itkStreamingImageFilterTest3.cxx
hjmjohnson: Thanks @N-Dekker

Modules/IO/ImageBase/include/itkImageFileReader.h
N-Dekker: @blowekamp @dzenanz Just extended the function description, hope you appreciate 😃 I added the following sentence:

> The function reads the image from the specified file, and returns the image that it has read.

https://github.com/InsightSoftwareConsortium/ITK/blob/3778f98e1f4d2edc53c135d8526ed2ab3da665d5/Modules/IO/ImageBase/include/itkImageFileReader.h#L178-L179

dzenanz (approved):

dzenanz:
Do we need `ITK_TEMPLATE_EXPORT` in the declaration? Sorry I noticed this just now.

PR #2188, opened 2020-12-17
ENH: Add operator== and operator!= to itk::Image

N-Dekker:
Added `operator==` and `operator!=` overloads which tell whether or not
two images are equal, based on their value semantics (not just their
addresses in memory).

Allowed testing if two images (not just image pointers) compare equal or
unequal by using GoogleTest `EXPECT_EQ(image1, image2)` and
`EXPECT_NE(image1, image2)`.

### DIFF
--- a/Modules/Core/Common/include/itkImage.h
+++ b/Modules/Core/Common/include/itkImage.h
@@ -318,6 +318,51 @@ public:
   unsigned int
   GetNumberOfComponentsPerPixel() const override;
 
+  /** Returns (image1 == image2). */
+  friend bool
+  operator==(const Image & lhs, const Image & rhs)
+  {
+    if ((lhs.GetBufferedRegion() != rhs.GetBufferedRegion()) || (lhs.m_Spacing != rhs.m_Spacing) ||
+        (lhs.m_Origin != rhs.m_Origin) || (lhs.m_Direction != rhs.m_Direction) ||
+        (lhs.m_InverseDirection != rhs.m_InverseDirection))
+    {
+      return false;
+    }
+
+    if (lhs.m_Buffer == rhs.m_Buffer)
+    {
+      return true;
+    }
+
+    if ((lhs.m_Buffer == nullptr) || (rhs.m_Buffer == nullptr))
+    {
+      return false;
+    }
+
+    auto & lhsBuffer = *(lhs.m_Buffer);
+    auto & rhsBuffer = *(rhs.m_Buffer);
+
+    const auto bufferSize = lhsBuffer.Size();
+
+    if (bufferSize != rhsBuffer.Size())
+    {
+      return false;
+    }
+
+    const TPixel * const lhsBufferPointer = lhsBuffer.GetBufferPointer();
+    const TPixel * const rhsBufferPointer = rhsBuffer.GetBufferPointer();
+
+    return ((lhsBufferPointer == rhsBufferPointer) ||
+            std::equal(lhsBufferPointer, lhsBufferPointer + bufferSize, rhsBufferPointer));
+  }
+
+  /** Returns (image1 != image2). */
+  friend bool
+  operator!=(const Image & lhs, const Image & rhs)
+  {
+    return !(lhs == rhs);
+  }
+
 protected:
   Image();
   void
--- a/Modules/Core/Common/test/CMakeLists.txt
+++ b/Modules/Core/Common/test/CMakeLists.txt
@@ -608,6 +608,7 @@ set(ITKCommonGTests
       itkConstantBoundaryImageNeighborhoodPixelAccessPolicyGTest.cxx
       itkFixedArrayGTest.cxx
       itkImageNeighborhoodOffsetsGTest.cxx
+      itkImageGTest.cxx
       itkImageBaseGTest.cxx
       itkImageBufferRangeGTest.cxx
       itkImageRegionRangeGTest.cxx
--- a/Modules/Core/Common/test/itkImageGTest.cxx
+++ b/Modules/Core/Common/test/itkImageGTest.cxx
@@ -1,0 +1,177 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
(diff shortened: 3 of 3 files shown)


N-Dekker:
@hjmjohnson Thank you for your approval and assignment! By the way, what does it mean to be assigned to a PR? `pull_request::operator=(const Developer&)`? Does it mean I'm now responsible for merging?

hjmjohnson:
The assignment does not really mean much for ITK.  I assigned it to indicate the most important person for this PR.

dzenanz (approved):

Modules/Core/Common/include/itkImage.h
```diff
@@ -318,6 +318,51 @@ class ITK_TEMPLATE_EXPORT Image : public ImageBase<VImageDimension>
...
 
+  /** Returns (image1 == image2). */
+  friend bool
+  operator==(const Image & lhs, const Image & rhs)
+  {
+    if ((lhs.GetBufferedRegion() != rhs.GetBufferedRegion()) || (lhs.m_Spacing != rhs.m_Spacing) ||
+        (lhs.m_Origin != rhs.m_Origin) || (lhs.m_Direction != rhs.m_Direction) ||
+        (lhs.m_InverseDirection != rhs.m_InverseDirection))
+    {
+      return false;
```
N-Dekker, line 329: Should `operator==` also check if the _requested_ region is the same for both images? I mean, is the "requested region" a fundamental part of the _value_ of an image? Or can two images be considered equal when only their _requested_ region is different? What do you think?

dzenanz: Since we are comparing all the other attributes, we should compare requested region too.

N-Dekker: OK, thanks @dzenanz. I was just thinking of `std::string`. Its `operator==` does not compare the capacity of the strings. Two strings with different capacity may still compare equal, because the capacity is not a fundamental part of the _value_ of a string.

But I'll add requested region to the comparison, if you think that's what we should do.

dzenanz: My opinion is not strong, so let's hear @blowekamp or maybe @hjmjohnson.

N-Dekker: My initial aim is being able to test the new image functions as follows:

    itk::WriteImage(image1, fileName);
    const auto image2 = itk::ReadImage(file)
    ASSERT_EQ(*image2, *image1); // Asserts that *image2 == *image1

In that case the requested region is equal to the buffered region, as far as I know. (Right?)

However I just realized, the requested region would matter when testing `Image::Graft` as follows:

    image2->Graft(image1);
    ASSERT_EQ(*image2, *image1);

`ImageBase::Graft` copies both the buffered and the requested region. So OK, I'll amend the PR! 😃

N-Dekker: Update: The latest [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/20f6b9672f272ae55270d79ece92748e2257a8b2..9bd662bf7e3a6346b0099660ba27aa2cab9b40de) has included the requested region with the comparison  😃

dzenanz (approved):

N-Dekker:
FYI The [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/9bd662bf7e3a6346b0099660ba27aa2cab9b40de..18bc9596f4eba93efb5e85e8b268fcb962840f32) that I just did only cleans up the added tests a little bit. I just realized it should be sufficient to test two `itk::Image` types, instead of four.

Modules/Core/Common/include/itkImage.h
N-Dekker: @dzenanz Honestly I'm still not 100% sure if the requested region should be included with the comparison. Especially now that I see that `itk::WriteImage(inputImage, fileName)` does adjust the requested region of the input image! Internally, via `ImageFileWriter::Write()`. And you would expect that an image _after_ calling `WriteImage` is "equal" to the image _before_ calling `WriteImage`. Right?

So then, would it be a bug of `ImageFileWriter::Write()` that is modifies the requested region?

blowekamp:
This is nicely coded with nice tests. I have mixed opinions about the `Image` objects as a parameter being introduced, and introducing operators for the object not being the normal ITK style.

Aside from that, please look at the following place where ITK has already established a "tolerance" for comparison of the meta-data geometry:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkImageToImageFilter.hxx#L185-L221

You mention that the goal of this PR is to enable GTEST comparison operators. For that an approximation or tolerant testing method is really needed.

Modules/Core/Common/include/itkImage.h
blowekamp: I would consider the requested region a "pipeline state" variable similar to the m_Modified time. We certainly don't want to compare the latter :)

I would be more inclined to ensure that the largest possible region was the same.

N-Dekker: OK, thanks Brad, I just removed requested region again, with the latest [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/18bc9596f4eba93efb5e85e8b268fcb962840f32..895645b296fe30579acce97c38665aca4584e3c3). Feeling like a flip flop  ;-)  Users who need to include the requested region with their comparison must now do so explicitly:

    if ((image1 == image2) && (image1.GetRequestedRegion() == image2.GetRequestedRegion())) // ...
	
Or an extra ASSERT_EQ, in unit tests:

    ASSERT_EQ(image1, image2);
    ASSERT_EQ(image1.GetRequestedRegion(), image2.GetRequestedRegion());

dzenanz (approved):
Sometimes, we might want true equality. And the equality operator a natural place for it.

We could add approximate equality method to `Image`? But that can be a later PR.

hjmjohnson (approved):

PR #2190, opened 2020-12-18
BUG: Update itkEventObjectTest to use new event macros

aylward:
VS2017 produces errors when compiling line 24 of itkEventObjectTest:

>> ITK\Modules\Core\Common\test\itkEventObjectTest.cxx(24): error C2065: 'TestEvent': undeclared identifier
>> ITK\Modules\Core\Common\test\itkEventObjectTest.cxx(24): error C2275: 'itk::UserEvent': illegal use of this type as an expression

Updating that test to use the preferred itkEventDefinitionMacro() and itkEventDeclarationMacro() fixes this issue.

### DIFF
--- a/Modules/Core/Common/test/itkEventObjectTest.cxx
+++ b/Modules/Core/Common/test/itkEventObjectTest.cxx
@@ -21,9 +21,12 @@
 
 namespace itk
 {
-itkEventMacro(TestEvent, UserEvent);
-itkEventMacro(TestDerivedEvent, TestEvent);
-itkEventMacro(TestOtherEvent, AnyEvent);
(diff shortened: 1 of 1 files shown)


dzenanz (commented):
The need for this changes probably means that 92b891d33ae9ee4fcd95055e74dff24a6b114d36 introduced the issue. Perhaps the `itkEventMacro` should be changed instead of the test?

aylward:
It is bracketed by
with the comment "Support Pre 2015 code bases".   I figured that meant that it wasn't expected to work with VS 2017.   Since we would want the test to work on all cases, I figured it should be updated.

Maybe @hjmjohnson has further guidance.

hjmjohnson (approved):

PR #2196, opened 2020-12-21
ENH: Improve coverage for `MathematicalMorphology` module classes.

jhlegarreta:
Improve coverage for `MathematicalMorphology` module classes:
- Use the `ITK_TRY_EXPECT_NO_EXCEPTION` macro to avoid boilerplate code.
- Remove explicit calls to the `Print` method and rely on the basic
  method exercising macro call.
- Improve slightly the test style to make them more consistent (e.g.
  input argument checking, test finishing message, make the types and
  variables dwell closer to the place where they are instantiated, etc.)

## PR Checklist

### DIFF
--- a/Modules/Filtering/MathematicalMorphology/test/itkClosingByReconstructionImageFilterTest.cxx
+++ b/Modules/Filtering/MathematicalMorphology/test/itkClosingByReconstructionImageFilterTest.cxx
@@ -29,7 +29,7 @@ itkClosingByReconstructionImageFilterTest(int argc, char * argv[])
 {
   if (argc < 5)
   {
-    std::cerr << "Missing arguments" << std::endl;
+    std::cerr << "Missing Parameters " << std::endl;
     std::cerr << "Usage: " << itkNameOfTestExecutableMacro(argv)
               << " Inputimage OutputImage Radius PreserveIntensities(0,1) [Diffmage]" << std::endl;
     return EXIT_FAILURE;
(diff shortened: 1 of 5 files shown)


Modules/Filtering/MathematicalMorphology/test/itkClosingByReconstructionImageFilterTest.cxx
```diff
@@ -40,75 +40,67 @@ itkClosingByReconstructionImageFilterTest(int argc, char * argv[])
...
-    }
+
+    writer->SetFileName(argv[5]);
+    writer->SetInput(subtract->GetOutput());
+
+    ITK_TRY_EXPECT_NO_EXCEPTION(writer->Update());
   }
+
+
+  std::cout << "Test finished" << std::endl;
```
blowekamp, line 104: This line in not really necessary.

If it is really needed and useful, perhaps it could be uniformly added to all tests in "itkTestDriverAfterTest.inc".

blowekamp (approved):
LGTM

Modules/Filtering/MathematicalMorphology/test/itkClosingByReconstructionImageFilterTest.cxx
dzenanz: > it could be uniformly added to all tests in "itkTestDriverAfterTest.inc"

This makes sense. It is somewhat of a big effort to remove all manual instances, though. Perhaps leave it to a new PR?

dzenanz (approved):

dzenanz:
There are some compile errors to be fixed.

Modules/Filtering/MathematicalMorphology/test/itkClosingByReconstructionImageFilterTest.cxx
jhlegarreta: I agree that the line is not necessary, but at the time I was more involved in improving the coverage it was not evident at times to see where or how a test had failed, or if was run at all, reading the standard output in the dashboard. This message allows a human reader to unequivocally know whether the test was run and finished.

If it can be applied to all tests with a single line or few lines in `itkTestDriverAfterTest.inc`, and we are satisfied with the result, then I'd be for adding the necessary code to that file. As Dženan says, it would be best to do it in a separate PR and remove the message present in tests in a single sweep. Maybe an issue can be opened to keep track of it/to avoid forgetting it.

thewtex (approved):
:green_heart:

PR #2203, opened 2020-12-22
ENH: Extend ITKWriteImageFunctionTest by reading the written image back

N-Dekker:
Tests that an image written by `itk::WriteImage` can be read back,
without loss of data.

Uses two recently added functions:
 - `itk::ReadImage`: pull request #2184 commit aeea88b1925e400fa6df099871b149a37f1dee17
 - `operator==` for images: pull request #2188 commit 7f3337d27ee684aa4ab6ccbd294603926895ee82

(`operator==` is called internally by GoogleTest `EXPECT_EQ`.)

### DIFF
--- a/Modules/IO/ImageBase/test/itkWriteImageFunctionGTest.cxx
+++ b/Modules/IO/ImageBase/test/itkWriteImageFunctionGTest.cxx
@@ -16,6 +16,7 @@
  *
  *=========================================================================*/
 #include "itkImageFileWriter.h"
+#include "itkImageFileReader.h"
 #include "itkImage.h"
 
 #include "itkGTest.h"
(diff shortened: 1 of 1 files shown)


N-Dekker:
Extended Bradley's unit test to check _WYWIWYR_: What You Write Is What You Read  😸

Modules/IO/ImageBase/test/itkWriteImageFunctionGTest.cxx
```diff
@@ -58,22 +59,30 @@ struct ITKWriteImageFunctionTest : public ::testing::Test
...
   image_ptr = MakeImage();
   ImageType * image_rptr = image_ptr.GetPointer();
-  itk::WriteImage(image_rptr, "test1.mha");
+  itk::WriteImage(image_rptr, fileName);
 
   const ImageType * image_crptr = image_ptr.GetPointer();
-  itk::WriteImage(image_crptr, "test1.mha");
+  itk::WriteImage(image_crptr, fileName);
+
+  // Test that what is written is also read back.
```
dzenanz, line 83: Maybe this should be done after each WriteImage? Also maybe not, as we don't want to waste time needlessly.

dzenanz (approved):

blowekamp (approved):
Looks nice.

Just an FYI: The MHA headers encode some of the meta data number as text. The reading and writing cycle can be lossy for non-trivial values.

Modules/IO/ImageBase/test/itkWriteImageFunctionGTest.cxx
N-Dekker: Good question @dzenanz but I'm not sure either, as I feel that the added value of such extra tests seems rather small, while there could indeed be a waste of time. Maybe someone else who _really_ wants to have a test after each `WriteImage` could submit a follow-up PR  😸

N-Dekker:
> Just an FYI: The MHA headers encode some of the meta data number as text. The reading and writing cycle can be lossy for non-trivial values.

@blowekamp  Thanks for reminding me that this _image_ ==> `WriteImage` ==> `ReadImage` ==> _image_ "round-trip" isn't always 100% lossless! Fortunately, at least, in `ITKWriteImageFunctionTest` it is  😃

hjmjohnson (approved):

PR #2241, opened 2021-01-06
ENH: Set ITK_TEST_OUTPUT_DIR with module path

blowekamp:
Change the default output path for modules testing data to include the
module name as a directory.

This prevents the same file name in defferent modules clobbering one
another causing intermediate testing failures.

## PR Checklist

### DIFF
--- a/CMake/ITKModuleMacros.cmake
+++ b/CMake/ITKModuleMacros.cmake
@@ -325,6 +325,8 @@ macro(itk_module_test)
   foreach(dep IN LISTS ITK_MODULE_${itk-module-test}_DEPENDS)
     list(APPEND ${itk-module-test}_LIBRARIES "${${dep}_LIBRARIES}")
   endforeach()
+  set(ITK_TEST_OUTPUT_DIR "${ITK_TEST_OUTPUT_DIR}/${itk-module}")
+  file(MAKE_DIRECTORY "${ITK_TEST_OUTPUT_DIR}")
 endmacro()
 
 macro(itk_module_examples)


dzenanz (commented):
Does this work as intended? Or to rephrase the question: how thoroughly has this been tested? It has potential to cause minor disruption in some remote modules.

blowekamp:
> how thoroughly has this been tested?

It has not been thoroughly tested. It seemed to work locally. And using the CI to test what happens across systems on clean build seemed like the next step.

>  It has potential to cause minor disruption in some remote modules.

Yes indeed. Alternatively, a new CMake variable could be introduced, and every thing changed to used it.. but that is more work that I care to tackle especially with the remote modules.

jhlegarreta:
If remote modules do provide an appropriate value for the `itk-module` variable, this seems a fair solution to me.

dzenanz:
> remote modules provide ... itk-module

I don't think this is an issue. I think there are some assumptions where the test temp dir is. Updating that should not be hard.

blowekamp:
Looks like all the tests in the CI passed :)

dzenanz:
Probably best to check how it works for externally built remote modules (pick 2-3 to test). I expect that remotes built as part of ITK will behave the same as regular modules.

thewtex:
Seems fine to generally help with filename clobbering issues. I tend to use the consolidated testing output directory with `rm -rf ITK-build/Testing/Temporary` to clean up build trees that are used repeatedly over time and inflate disk usage. This would make it more difficult, but maybe we could also isolate more easily areas that require cleanup.

blowekamp:
> I tend to use the consolidated testing output directory with `rm -rf ITK-build/Testing/Temporary` to clean up build trees that are used repeatedly over time and inflate disk usage.

That will still work as the new test output directory for modules will be like `ITK-build/Testing/Temporary/ITKCommon` and not under the modules build path.

jhlegarreta:
If as @dzenanz said rather than the module name issue this is likely to impact the output dir expected by the remotes, once that has been checked, I'd be for the change.

dzenanz:
Brad, are we still targeting this for 5.3?

hjmjohnson (approved):
I think this is ready for merging now.

PR #2245, opened 2021-01-08
BUG: Set runtime output path for external module binaries

thewtex:
Addresses crashing test driver on Windows with shared libraries.


dzenanz (approved):

dzenanz:
Is this why some remote modules have test failures?

thewtex:
> Is this why some remote modules have test failures?

I tested it with `ITKModuleTemplate`, but it should impact many.

jhlegarreta (approved):
Have not tested, but :100:.

PR #2246, opened 2021-01-10
ENH: Add operator==, operator!= to MetaDataObject and MetaDataDictionary

N-Dekker:
Follow-up to "ENH: Add operator== and operator!= to itk::Image"
pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2188
commit 7f3337d27ee684aa4ab6ccbd294603926895ee82

### DIFF
--- a/Modules/Core/Common/include/itkMetaDataDictionary.h
+++ b/Modules/Core/Common/include/itkMetaDataDictionary.h
@@ -19,6 +19,7 @@
 #define itkMetaDataDictionary_h
 
 #include "itkMetaDataObjectBase.h"
+#include <algorithm> // For std::equal
 #include <vector>
 #include <map>
 #include <string>
@@ -80,6 +81,35 @@ public:
 
   // Destructor
   virtual ~MetaDataDictionary();
+
+  /** Returns (metaDataDictionary1 == metaDataDictionary2). */
+  friend bool
+  operator==(const Self & lhs, const Self & rhs)
+  {
+    using KeyValuePair = MetaDataDictionaryMapType::value_type;
+
+    return (lhs.m_Dictionary == rhs.m_Dictionary) ||
+           ((lhs.m_Dictionary != nullptr) && (rhs.m_Dictionary != nullptr) &&
+            (lhs.m_Dictionary->size() == rhs.m_Dictionary->size()) &&
+            std::equal(lhs.m_Dictionary->cbegin(),
+                       lhs.m_Dictionary->cend(),
+                       rhs.m_Dictionary->cbegin(),
+                       [](const KeyValuePair & keyValuePair1, const KeyValuePair & keyValuePair2) {
+                         const auto & value1 = keyValuePair1.second;
+                         const auto & value2 = keyValuePair2.second;
+                         return (keyValuePair1.first == keyValuePair2.first) &&
+                                ((value1 == value2) ||
+                                 ((value1 != nullptr) && (value2 != nullptr) && (*value1 == *value2)));
+                       }));
+  }
+
+  /** Returns (metaDataDictionary1 != metaDataDictionary2). */
+  friend bool
+  operator!=(const Self & lhs, const Self & rhs)
+  {
+    return !(lhs == rhs);
+  }
+
 
   /** Returns a vector of keys to the key/value entries in the
    * dictionary.  Iterate through the dictionary using these keys.
--- a/Modules/Core/Common/include/itkMetaDataObject.h
+++ b/Modules/Core/Common/include/itkMetaDataObject.h
@@ -130,11 +130,33 @@ public:
   void
   Print(std::ostream & os) const override;
 
+  /** Returns (metaDataObject1 == metaDataObject2). */
+  friend bool
+  operator==(const Self & lhs, const Self & rhs)
+  {
+    return lhs.m_MetaDataObjectValue == rhs.m_MetaDataObjectValue;
+  }
+
+  /** Returns (metaDataObject1 != metaDataObject2). */
+  friend bool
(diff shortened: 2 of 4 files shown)


Modules/Core/Common/include/itkMetaDataDictionary.h
```diff
@@ -81,6 +82,35 @@ class ITKCommon_EXPORT MetaDataDictionary
...
 
+  /** Returns (metaDataDictionary1 == metaDataDictionary2). */
+  friend bool
+  operator==(const Self & lhs, const Self & rhs)
+  {
+    using KeyValuePair = MetaDataDictionaryMapType::value_type;
+
+    return (lhs.m_Dictionary == rhs.m_Dictionary) ||
+           ((lhs.m_Dictionary != nullptr) && (rhs.m_Dictionary != nullptr)) &&
+             (lhs.m_Dictionary->size() == rhs.m_Dictionary->size()) &&
```
dzenanz, line 93: `Modules/Core/Common/include/itkMetaDataDictionary.h:93:69: warning: '&&' within '||' [-Wlogical-op-parentheses]`

Since there is only one `||`, maybe you don't need extra parentheses around the second set of conditions?

dzenanz (approved):
Looks good to me, aside from compile warnings.

Modules/Core/Common/include/itkMetaDataObject.h
```diff
@@ -130,11 +131,32 @@ class ITK_TEMPLATE_EXPORT MetaDataObject : public MetaDataObjectBase
...
 protected:
   MetaDataObject() = default;
   ~MetaDataObject() override = default;
 
 private:
+  /** Internal helper function used to implement operator== for MetaDataObjectBase. */
+  bool
+  Equal(const MetaDataObjectBase & metaDataObject) const override
+  {
+    return (typeid(metaDataObject) == typeid(Self)) && (*this == static_cast<const Self &>(metaDataObject));
```
blowekamp, line 157: Why is this typeid comparison done over using a dynamic_cast?

Beware there are monsters here. There have been issues related to shared libraries, linkage and duplicate symbols ( typeids ) not being equal on some operating symbols (OS X), which cause these types of comparisons or dynamic_cast to fail which the object refer to the same class/lines of code but the type information is from difference complications units. I hope we have addressed they will not reoccur. However, giving this history I need to and some checks to help if this does reoccur.

Please break up this compound statement. When the typeid or dynamic_cast fails we should add an additional check to see if the objects are from the same class. The `type_info`'s `name` method may provide this check. So add if the first comparison fails:  `if ( typeid(metaDataObject).name() == typeid(Self).name() ) itkWarningMacro("...")`. This will be of great use if the linkage issue reoccurs.

N-Dekker: @blowekamp Thanks for your feedback. I used `typeid`, to state _explicitly_ that two meta-data-objects _only_ compare equal when their types are equal. I think the intension of `typeid(metaDataObject) == typeid(Self)` is most clear.

However, if you prefer `dynamic_cast`, that would be fine to me as well. I could certainly make it:

    const auto metaDataObjectAsSelf = dynamic_cast<const Self *>(&metaDataObject);
    return (metaDataObjectAsSelf != nullptr) && (*this == *metaDataObjectAsSelf);


Would you like that better?

Regarding those possible issues related to shared libraries, linkage and duplicate symbols, does it help that `MetaDataObject<T>` is explicitly instantiated for a specific set of types? In https://github.com/InsightSoftwareConsortium/ITK/blob/v5.2rc01/Modules/Core/Common/src/itkMetaDataObject.cxx

I would rather not put a `itkWarningMacro` in there, as I think by definitions, `operator==` should not have any side effects.

Modules/Core/Common/include/itkMetaDataDictionary.h
N-Dekker: Thanks @dzenanz, hopefully fixed by force-push https://github.com/InsightSoftwareConsortium/ITK/compare/b76ceccb3be9be9b0a016088939ecf7dcf08e50d..ce8cf4336409f4521feae6a6bd76a3420d6545d4

Modules/Core/Common/include/itkMetaDataObject.h
blowekamp: > ```
> const auto metaDataObjectAsSelf = dynamic_cast<const Self *>(&metaDataObject);
> return (metaDataObjectAsSelf != nullptr) && (*this == *metaDataObjectAsSelf);
> ```
> 
> 
> Would you like that better?
@N-Dekker  Yes. It is more consistent with ITK, and less likely to be detected as improper use of `static_cast`.

>> I would rather not put a itkWarningMacro in there, as I think by definitions, operator== should not have any side effects.

I will improve robustness for a problematic situation that likely can still occur in some configuration of ITK. It is important that there are checks in ITK to ensure the code is robust and errouneous states and able to be tracked down.  It will not have side effect ( display warning ) unless the problem is encountered, and it will make the issue be able to tracked down.

N-Dekker: My idea with this PR is that `(metaDataObjectBase1 == metaDataObjectBase2)` would return `false`, and `(metaDataObjectBase1 != metaDataObjectBase2)`  would return `true`, when they refer to objects whose value type is different. The `==` and `!=` comparisons themselves would then still be valid. That's why I think a warning or error message is not necessary. OK?

N-Dekker:
This pull request is just a nice-to-have to me. It would be nice to be able to compare meta-data this way. But it isn't something I need right away.

dzenanz:
Well, now is a good time to merge this. Brad?

blowekamp (approved):

PR #2248, opened 2021-01-10
ENH: Array, OptimizerParameters constructors with size and initial value

N-Dekker:
Added explicit constructors to `Array` and `OptimizerParameters`, allowing to specify the size (dimension) and the initial value of each element.

Both `std::vector` and `vnl_vector` have a similar constructor.

Included a GoogleTest unit test for `OptimizerParameters`.

### DIFF
--- a/Modules/Core/Common/include/itkArray.h
+++ b/Modules/Core/Common/include/itkArray.h
@@ -65,8 +65,13 @@ public:
   /** Construct from a VnlVectorType */
   explicit Array(const VnlVectorType &);
 
-  /** Constructor with size. Size can only be changed by assignment */
+  /** Constructor with size. Size can only be changed by assignment.
+   * \note This constructor may not initialize its elements.
+   */
   explicit Array(SizeValueType dimension);
+
+  /** Constructor with size and initial value for each element. */
+  explicit Array(SizeValueType dimension, const ValueType & value);
 
   /** Constructor that initializes array with contents from a user supplied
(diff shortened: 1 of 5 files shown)


dzenanz (approved):

N-Dekker:
Thanks for your approval, Dženan!

@thewtex Matt, do you think that this pull request could still be included with ITK 5.2? I think it's pretty simple enhancement, but it could certainly make the use of `OptimizerParameters` easier. This is how I created a `const` OptimizerParameters variable of two elements for an elastix unit test with ITK 5.1:

    const itk::OptimizerParameters<double> optimizerParameters(itk::Array<double>(vnl_vector<double>(2U, testValue)));  
https://github.com/SuperElastix/elastix/blob/b60640fe6e0ef002ca567768cce404573f0c0109/Common/GTesting/elxTransformIOGTest.cxx#L324 

The pull request would allow, hopefully for ITK 5.2: 

    const itk::OptimizerParameters<double> optimizerParameters(2U, testValue);

dzenanz:
This does not seem like a potentially problematic feature. #2246 might qualify for that. But we did not announce a feature freeze.

If you are not going to request more reviews, you could merge this PR Niels.

N-Dekker:
> If you are not going to request more reviews, you could merge this PR Niels.

@dzenanz Thanks Dženan, could you please do the merge? I always prefer not to merge my own PRs, if possibly 😃 It looks nicer to me when there are two names with a commit (author + committer).

PR #2249, opened 2021-01-13
Add HardenTransform (resample in place)

dzenanz:
## PR Checklist


dzenanz:
I just realized this only handles rigid transforms properly.

dzenanz:
The force-push should fix the scaling.

jhlegarreta:
@dzenanz thanks for doing this :100: ! May I add a couple of comments?
- A first read at the method name without reading the docstring one may think it is the name of a new type of transform named after some researcher that came up with it (the "Harden transform"; although it'd make more sense for a class name) 😅 , so I'd vote for using an alternative name, e.g. `ApplyTransform`, `ChangeTransform` or `EnforceTransform`, or `ForceTransform`. Maybe native English speakers can come up with a better name than these (not sure if the first two are very appropriate or honor what the method really does).
- Maybe the method can be called from one of the existing tests to show how one could use it, and btw improve the coverage (to hopefully become closer to a 90% :rocket: in one of the builds in the mid term).

Thanks for the effort.

thewtex (changes requested):
:+1: to @jhlegarreta 's suggestion to add tests to keep our coverage improving

blowekamp:
This is a much needed method to be added to ITK 👍 . I agree that testing is needed :)

- The `ImageBase` is in the Common/Core module, which currently does not depend on the `Transform` module. Adding this as a member function creates an interdependency between that modules, which should be avoided. Are there alternative places for method? Perhaps as part of the TransformBase class?
- If the transform is not Global/Affine, the method should throw an exception and no just issue a warning.
-  Consider taking as the argument type `const TransformBase<TValueType, VImageDimension, VImageDimension> *` as oppose to the "universal reference" currently implemented. Using the specific raw pointer type will make the expected usage clearer, follow the conventions in ITK, and make it easier to wrap/instantiate. Keeping interface consistent across a toolkit is important to keep the interface as user expect it and avoids surprises and miss usage.

hjmjohnson (commented):
See also https://github.com/BRAINSia/BRAINSTools/blob/master/BRAINSCommonLib/itkResampleInPlaceImageFilter.hxx

https://github.com/BRAINSia/BRAINSTools/blob/master/BRAINSCommonLib/itkResampleInPlaceImageFilter.h

dzenanz:
Thanks for pointing out the code, Hans. I will steal it 😄

The extra code overhead to instantiate a filter seems like unnecessary burden. I like an `ApplyTransform` method in rigid transforms (e.g. [MatrixOffsetTransformBase](https://itk.org/Doxygen/html/classitk_1_1MatrixOffsetTransformBase.html)) much better. Once you have an image (smart)pointer and a transform you don't need any further entities such as a filter.

I don't think I will get to updating this, this week.

blowekamp:
I think implementing it as a filter is consistent with ITK and it's pipeline architecture. Additionally given Han's provided code it may even be easier to integrate that code close to as it is than come up with some new ( inconsistent and debatable way ) to add the functionality to ITK.

dzenanz:
If someone knows how to disable the method in case `InDim!=OutDim`, please share. What I had works only on Visual Studio.

Leengit:
You might try using `enable_if` on the return type instead of having an extra template parameter.  That is, go back a declaration and definition of the method that don't care about dimensions, but instead of `void` as the return type, try
```c++
typename std::enable_if<NInputDimensions == NOutputDimensions, void>::type
```

dzenanz:
@Leengit I tried that first, it doesn't even compile on VS. For SFINAE to work, stuff inside `enable_if` must be dependent on template parameters.

Modules/Core/Transform/include/itkResampleInPlaceImageFilter.h
```diff
@@ -0,0 +1,160 @@
...
+
+namespace itk
+{
+/** \class ResampleInPlaceImageFilter
+ * \brief Resample an image in place.
+ *
+ * The current ITK resample image filter will generate a physical memory-
+ * modified version of the input image if the input transform is not identity. The
+ * abuse use of the filter can be cumbersome in a situation that the image is very
+ * large, and there are lots of transform to be superimposed for the input image, and
```
jhlegarreta, line 33: transform: transforms

Modules/Core/Transform/include/itkResampleInPlaceImageFilter.h
```diff
@@ -0,0 +1,160 @@
...
+#include "itkVersorRigid3DTransform.h"
+
+namespace itk
+{
+/** \class ResampleInPlaceImageFilter
+ * \brief Resample an image in place.
+ *
+ * The current ITK resample image filter will generate a physical memory-
+ * modified version of the input image if the input transform is not identity. The
+ * abuse use of the filter can be cumbersome in a situation that the image is very
```
jhlegarreta, line 32: I would maybe restate this last sentence: I find that "(...) can be cumbersome (...)" is not that informative about the problems that can arise from the abuse. Maybe saying that "(...) can be a source of problems. such as {mention a few of the issues that can arise}. For example, when the image is very large, and (...)".

Modules/Core/Transform/include/itkResampleInPlaceImageFilter.h
```diff
@@ -0,0 +1,160 @@
...
+ * large, and there are lots of transform to be superimposed for the input image, and
+ * we even don't care about those intermediate transformed images.
+ *
+ * If all the transforms are rigid, a far superior way to achieve a similar result
+ * while minimizing the accumulative resampling errors as well as eliminating the expense
+ * on accessing the physical memory of the image is to compose all the
+ * transforms before hand if it is possible, and then we only need to resample the
```
jhlegarreta, line 39: before hand: should it be a single word?

Modules/Core/Transform/include/itkResampleInPlaceImageFilter.h
```diff
@@ -0,0 +1,160 @@
...
+ * input image with the final composed transform once.
```
jhlegarreta, line 40: I would split the sentence so that the reading becomes easier.

Modules/Core/Transform/include/itkResampleInPlaceImageFilter.h
```diff
@@ -0,0 +1,160 @@
...
+ *
+ * Here we present a more compact alternative, all information is stored in the header
```
jhlegarreta, line 42: I'd use an impersonal mode: "Here, a more compact alternative is presented: (...)"

That comment can also be applied to the part a few sentences above: "(...), and we even don't care (...)" , e.g. "and those intermediate transformed images are not important/relevant/used."

Modules/Core/Transform/include/itkResampleInPlaceImageFilter.h
```diff
@@ -0,0 +1,160 @@
...
+ * of the image and there is no need to maintain the final transform any longer. ITK
+ * image class has innate support for doing this.
+ *
+ * \param RigidTransform -- Currently must be a VersorRigid3D
```
jhlegarreta, line 46: Maybe `VersorRigid3D` can be prefixed with the appropriate Doxygen keyword to generate the link to the class documentation?

Modules/Core/Transform/include/itkResampleInPlaceImageFilter.h
```diff
@@ -0,0 +1,160 @@
...
+/** \class ResampleInPlaceImageFilter
+ * \brief Resample an image in place.
+ *
+ * The current ITK resample image filter will generate a physical memory-
+ * modified version of the input image if the input transform is not identity. The
+ * abuse use of the filter can be cumbersome in a situation that the image is very
+ * large, and there are lots of transform to be superimposed for the input image, and
+ * we even don't care about those intermediate transformed images.
+ *
+ * If all the transforms are rigid, a far superior way to achieve a similar result
```
jhlegarreta, line 36: Not sure if I understand why transformations are mentioned in the preceding paragraph and here if the filter name and brief state that this "only" resamples an image in place.

Modules/Core/Transform/include/itkResampleInPlaceImageFilter.h
```diff
@@ -0,0 +1,160 @@
...
+ * on accessing the physical memory of the image is to compose all the
+ * transforms before hand if it is possible, and then we only need to resample the
+ * input image with the final composed transform once.
+ *
+ * Here we present a more compact alternative, all information is stored in the header
+ * of the image and there is no need to maintain the final transform any longer. ITK
+ * image class has innate support for doing this.
+ *
+ * \param RigidTransform -- Currently must be a VersorRigid3D
+ * \param InputImage -- The image to be duplicated and modified to incorporate the
```
jhlegarreta, line 47: I'd simply say that it's the image on which the resampling (if the filter does only a resampling) will take place.

Modules/Core/Transform/include/itkResampleInPlaceImageFilter.h
```diff
@@ -0,0 +1,160 @@
...
+ * those points are converted to PhysMovingImagePoints = TfmF2M( PhysFixedImagePoints )
+ * and then MovingContinuousIndexPoints = movingImage->TransformPhysicalPointToContinuousIndex( PhysMovingImagePoints )
+ * to get image values.
+ *
+ * We desire to change the moving image DirectionCosign [DC] and Origin O such that
+ * we can compute the:
+ * MovingContinuousIndexPoints = newMovingImage->TransformPhysicalPointToContinuousIndex( PhysFixedImagePoints )
+ *
+ *Image Notations:
+ *  DC-DirectionCosign
```
jhlegarreta, line 65: DirectionCosign: DirectionCosine

Modules/Core/Transform/include/itkResampleInPlaceImageFilter.hxx
```diff
@@ -0,0 +1,102 @@
...
+
+#include "itkResampleInPlaceImageFilter.h"
+#include "itkCastImageFilter.h"
+
+namespace itk
+{
+template <typename TInputImage, typename TOutputImage>
+ResampleInPlaceImageFilter<TInputImage, TOutputImage>::ResampleInPlaceImageFilter()
+  : m_OutputImage(nullptr)
+  , m_RigidTransform(nullptr)
```
jhlegarreta, line 30: Same comment as above.

Modules/Core/Transform/include/itkTransform.hxx
```diff
@@ -470,6 +471,58 @@ Transform<TParametersValueType, NInputDimensions, NOutputDimensions>::TransformS
...
+
+  using ImageType = typename std::remove_const<typename std::remove_reference<decltype(*image)>::type>::type;
+
+  static_assert(Self::InputSpaceDimension == ImageType::ImageDimension,
+                "Image and transform must have the same spatial dimension");
+
+  if (!this->IsLinear())
+  {
+    itkWarningMacro(<< "ApplyToImageMetadata was invoked with non-linear transform of type: " << this->GetNameOfClass()
+                    << ". This might produce unexpected result.");
```
jhlegarreta, line 492: "This might produce unexpected result." : "This might produce unexpected results." or "This might produce an unexpected result."

Modules/Core/Transform/include/itkTransform.hxx
```diff
@@ -470,6 +471,58 @@ Transform<TParametersValueType, NInputDimensions, NOutputDimensions>::TransformS
...
+{
+  using NonReferenceImagePointer = typename std::remove_reference<TImagePointer>::type;
+  static_assert(std::is_pointer<NonReferenceImagePointer>::value ||
+                  mpl::IsSmartPointer<NonReferenceImagePointer>::Value,
+                "ApplyToImageMetadata requires a raw pointer or SmartPointer.");
+
+  using ImageType = typename std::remove_const<typename std::remove_reference<decltype(*image)>::type>::type;
+
+  static_assert(Self::InputSpaceDimension == ImageType::ImageDimension,
+                "Image and transform must have the same spatial dimension");
```
jhlegarreta, line 487: dimension : dimension.

Modules/Core/Transform/test/itkResampleInPlaceImageFilterTest.cxx
```diff
@@ -0,0 +1,145 @@
...
+  return result;
+}
+
+int
+itkResampleInPlaceImageFilterTest(int argc, char * argv[])
+{
+  if (argc < 3)
+  {
+    std::cerr << "Wrong arguments!" << std::endl;
+    std::cerr << "Usage: ./" << argv[0] << " inputImage baselineImage outputImage" << std::endl;
```
jhlegarreta, line 86: `argv[0]`: prefer to use `itkNameOfTestExecutableMacro(argv)`

jhlegarreta: Not sure if "./" was intended.

Modules/Core/Transform/test/itkResampleInPlaceImageFilterTest.cxx
```diff
@@ -0,0 +1,145 @@
...
+inline bool
+Validate(double input, double desired, double tolerance)
+{
+  return std::abs<double>(input - desired) > tolerance * std::abs<double>(desired);
+}
+
+// Returns true if images are different
+template <typename ImageType>
+bool
+imagesDifferent(ImageType * baselineImage, ImageType * outputImage)
```
jhlegarreta, line 40: Isn't the testing image comparison method enough in this case? If not, should this be added to the testing framework?

Modules/Core/Transform/test/itkResampleInPlaceImageFilterTest.cxx
```diff
@@ -0,0 +1,145 @@
...
+
+  double rotationAngle = .5; // Radians
+
+  TransformType::Pointer transform = TransformType::New(); // Identity by default
+  transform->SetCenter(center);
+  transform->Translate(translation);
+  transform->SetRotation(rotationAxis, rotationAngle);
+
+  // Set up the resample filter
+  FilterType::Pointer filter = FilterType::New();
```
jhlegarreta, line 128: Please consider exercising the basic image methods:
```
ITK_EXERCISE_BASIC_OBJECT_METHODS(filter, ResampleInPlaceImageFilter, ImageToImageFilter)
```

Modules/Core/Transform/test/itkResampleInPlaceImageFilterTest.cxx
```diff
@@ -0,0 +1,145 @@
...
+  filter->SetInputImage(inputImage);
```
jhlegarreta, line 129: Can the corresponding Get method be exercised and be compared with the expected output?

Modules/Core/Transform/test/itkResampleInPlaceImageFilterTest.cxx
```diff
@@ -0,0 +1,145 @@
...
+  filter->SetRigidTransform(transform);
```
jhlegarreta, line 130: Can we please exercise the corresponding Get method?
```
ITK_TEST_SET_GET_VALUE(transform, filter->GetRigidTransform());
```

jhlegarreta (commented):
Thanks for giving this another try @dzenanz :100:. I made a few comments.

Modules/Core/Transform/include/itkResampleInPlaceImageFilter.h
dzenanz: I don't know off the bat what keyword is required. Maybe `\class`?

Modules/Core/Transform/include/itkResampleInPlaceImageFilter.h
dzenanz: I reworded most of the documentation now.

Modules/Core/Transform/include/itkResampleInPlaceImageFilter.h
dzenanz: This filter duplicates the image, it does not modify the original. This increases (peak) memory use, which is important.

Modules/Core/Transform/include/itkResampleInPlaceImageFilter.h
```diff
@@ -0,0 +1,160 @@
...
+ * TransformPhysicalPointToContinuousIndex:
+ * CI = [SP^-1][DC^-1]( PhysMovingImagePoints - O )
+ * PhysMovingImagePoints = [R](PhysFixedImagePoints - C) + C + T
+ *
+ * After substitutions:
+ * MovingContinuousIndexPoints = [R^-1][DC][SP] * CI + [R^-1] * O - [R^-1] * C - [R^-1]*T + C
+ *                               ----------            --------------------------------------
+ *                                 NewDC                    NewOrigin
+ * NewDC = [R^-1][DC]
+ * NewOrigin = [R^-1] * ( O - C - T ) + C
```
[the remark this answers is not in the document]
dzenanz: If someone has enthusiasm to do it, they are more than welcome!

Modules/Core/Transform/include/itkResampleInPlaceImageFilter.h
```diff
@@ -0,0 +1,160 @@
...
+ * \return -- An image with the same voxels values as the input, but with differnt
+ * physical space representation affected by the rigid transform.
+ *
+ * The purpose of this code is to generate the new origin and direction
+ * that will remove the need for using the transform.
+ *
+ * Given a set of PhysFixedImagePoints (i.e. from the fixedImage space)
+ * those points are converted to PhysMovingImagePoints = TfmF2M( PhysFixedImagePoints )
+ * and then MovingContinuousIndexPoints = movingImage->TransformPhysicalPointToContinuousIndex( PhysMovingImagePoints )
+ * to get image values.
```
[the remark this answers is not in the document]
dzenanz: TfmF2M = Transform Fixed to Moving

Modules/Core/Transform/include/itkResampleInPlaceImageFilter.hxx
```diff
@@ -0,0 +1,102 @@
...
+#define itkResampleInPlaceImageFilter_hxx
+
+#include "itkResampleInPlaceImageFilter.h"
+#include "itkCastImageFilter.h"
+
+namespace itk
+{
+template <typename TInputImage, typename TOutputImage>
+ResampleInPlaceImageFilter<TInputImage, TOutputImage>::ResampleInPlaceImageFilter()
+  : m_OutputImage(nullptr)
```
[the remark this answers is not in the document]
dzenanz: Yes!

Modules/Core/Transform/test/itkResampleInPlaceImageFilterTest.cxx
dzenanz: This test can probably be further simplified. I was trying to reuse it, not spend time rewriting it completely. You are welcome to do so in a follow-up PR.

dzenanz:
I made extra commits for ease of review. Once reviews are finished, I will squash them appropriately.

Modules/Core/Transform/include/itkResampleInPlaceImageFilter.h
jhlegarreta: I think `\class` would make it: https://www.doxygen.nl/manual/commands.html

Modules/Core/Transform/include/itkResampleInPlaceImageFilter.h
```diff
@@ -52,17 +49,17 @@ namespace itk
...
  * that will remove the need for using the transform.
  *
- * Given a set of PhysFixedImagePoints (i.e. from the fixedImage space)
- * those points are converted to PhysMovingImagePoints = TfmF2M( PhysFixedImagePoints )
- * and then MovingContinuousIndexPoints = movingImage->TransformPhysicalPointToContinuousIndex( PhysMovingImagePoints )
- * to get image values.
+ * Given a set of PhysicalFixedImagePoints (i.e. from the fixedImage space)
+ * those points are converted to PhysicalMovingImagePoints = TfmF2M( PhysicalFixedImagePoints )
+ * and then MovingContinuousIndexPoints = movingImage->TransformPhysicalPointToContinuousIndex(
+ *PhysicalMovingImagePoints ) to get image values.
```
jhlegarreta, line 55: Missing whitespace before the comment block.

Modules/Core/Transform/test/itkResampleInPlaceImageFilterTest.cxx
```diff
@@ -145,5 +148,7 @@ itkResampleInPlaceImageFilterTest(int argc, char * argv[])
   TransformType * rawPointerTransform = transform.GetPointer();
   rawPointerTransform->ApplyToImageMetadata(inputImage);
 
+  ITK_EXERCISE_BASIC_OBJECT_METHODS(filter, ResampleInPlaceImageFilter, ImageToImageFilter);
```
jhlegarreta, line 157: Although given the variables in this filter this is not likely, I think that it is desirable to put this right after the instantiation of the filter to be able to spot problems that can otherwise go unnoticed (e.g. uninitialized member variables).

Modules/Core/Transform/include/itkResampleInPlaceImageFilter.h
```diff
@@ -52,17 +49,17 @@ namespace itk
  * The purpose of this code is to generate the new origin and direction
  * that will remove the need for using the transform.
  *
- * Given a set of PhysFixedImagePoints (i.e. from the fixedImage space)
- * those points are converted to PhysMovingImagePoints = TfmF2M( PhysFixedImagePoints )
- * and then MovingContinuousIndexPoints = movingImage->TransformPhysicalPointToContinuousIndex( PhysMovingImagePoints )
- * to get image values.
+ * Given a set of PhysicalFixedImagePoints (i.e. from the fixedImage space)
+ * those points are converted to PhysicalMovingImagePoints = TfmF2M( PhysicalFixedImagePoints )
```
jhlegarreta, line 53: IMO using capital letters for a variable name is misleading.

I think the term `TfmF2M` has not been introduced previously/it would help to explain what it means.

Modules/Core/Transform/include/itkResampleInPlaceImageFilter.h
```diff
@@ -0,0 +1,160 @@
...
+ *Image Notations:
+ *  DC-DirectionCosign
+ *  O-Origin
+ *
+ *Rigid Transform Notations:
+ *  R-Rotation
+ *  C-Center Of Rotation
+ *  T-Translation
+ *
+ * TransformPhysicalPointToContinuousIndex:
```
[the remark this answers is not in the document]
dzenanz: Would `\memberof` work?

dzenanz:
The recent force-push squashes the commits.

Modules/Core/Transform/include/itkResampleInPlaceImageFilter.h
```diff
@@ -0,0 +1,157 @@
...
+ *  DC-DirectionCosine
+ *  O-Origin
+ *
+ * Rigid Transform Notations:
+ *  R-Rotation
+ *  C-Center Of Rotation
+ *  T-Translation
+ *
+ * TransformPhysicalPointToContinuousIndex:
+ * CI = [SP^-1][DC^-1]( PhysicalMovingImagePoints - O )
```
jhlegarreta, line 71: `CI` and `SP` should maybe be defined?

Modules/Core/Transform/include/itkResampleInPlaceImageFilter.h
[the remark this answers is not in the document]
jhlegarreta: As for the latex equation notation, here the manual
https://www.doxygen.nl/manual/formulas.html

And one example in ITK:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/DistanceMap/include/itkHausdorffDistanceImageFilter.h

with its Doxygen result:
https://itk.org/Doxygen/html/classitk_1_1HausdorffDistanceImageFilter.html#a1ad7f5fb53378d18305547652aa1873a

Below is an attempt. Note that I renamed the variables to denote them with a single letter, matrices in (bold) capitals, vectors in (bold) lower cases. Please check for correctness.
`DC` renamed to `D`
`SP` renamed to `S`
`CI` renamed to `i`

```
Image Notations:
  \f$\mathbf{D}\f$: Direction cosine matrix
  \f$\mathbf{o}\f$: Origin vector
  \f$\mathbf{D}^{'}\f$: New direction cosine matrix
  \f$\mathbf{o}^{'}\f$: New origin vector

Rigid Transform Notations:
  \f$\mathbf{R}\f$: Rotation matrix
  \f$\mathbf{c}\f$: Center of rotation vector
  \f$\mathbf{t}\f$: Translation vector

  \f$\mathbf{f}_{p}\f$: fixed image points in physical space
  \f$\mathbf{m}_{p}\f$: moving image points in physical space

\f[
  \mathbf{ci} = \mathbf{S}^{-1}\mathbf{D}^{-1}( \mathbf{m}_{p} - \mathbf{o} ) \\
  \mathbf{m}_{p} = \mathbf{R}(\mathbf{f}_{p} - \mathbf{c}) + \mathbf{c} + \mathbf{T}
\f]

After substitutions:

\f[
\mathbf{m}_{c} = \underbrace{\mathbf{R}^{-1}\mathbf{D}}_\text{new cosine}\mathbf{S} * \mathbf{i} + \underbrace{\mathbf{R}^{-1} * \mathbf{o} - \mathbf{R}^{-1} * \mathbf{c} - \mathbf{R}^{-1}*T}_\text{new origin} + \mathbf{c} \\
\\
\mathbf{D}^{'} = \mathbf{R}^{-1}\mathbf{D} \\
\mathbf{o}^{'} = \mathbf{R}^{-1} * ( \mathbf{o} - \mathbf{c} - \mathbf{t} ) + \mathbf{c}
\f]
```

jhlegarreta (commented):
:+1: thanks @dzenanz. A suggestion to write the equations in LaTeX and another minor comment.

dzenanz:
The force push applies Jon's suggested LaTeX formatting.

blowekamp (approved):
LGTM.

I expected the class to be derived from the InPlaceImageFilter. That can be fixed up in a future patch.

dzenanz:
@jhlegarreta The `\class` keyword does not link to the appropriate page, it just breaks the documentation starting from where it appears, see [doxygen docs](https://itk.org/Doxygen/html/classitk_1_1ResampleInPlaceImageFilter.html). Hotfix in #2484.

jhlegarreta:
> @jhlegarreta The \class keyword does not link to the appropriate page, it just breaks the documentation starting from where it appears, see doxygen docs. Hotfix in #2484.

:pensive: Sorry @dzenanz. Maybe the problem is that Doxygen is confused about which class is being defined; maybe the class name is enough.

PR #2256, opened 2021-01-17
ENH: Support calling a const NumberToString, add unit tests + style improvements

N-Dekker:
Supported calling a `const` NumberToString function object, by declaring all `NumberToString::operator()` overloads `const`.

Added various GoogleTest unit tests, which do call `const` NumberToString objects.

Did various style improvements, including a reduction of the `buf` size, and removal of `builder.Reset()` calls from the implementation of `NumberToString`.

### DIFF
--- a/Modules/Core/Common/src/itkNumberToString.cxx
+++ b/Modules/Core/Common/src/itkNumberToString.cxx
@@ -26,7 +26,7 @@ namespace itk
 
 template <>
 std::string
-NumberToString<double>::operator()(double val)
+NumberToString<double>::operator()(double val) const
 {
   char                                               buf[256];
   const double_conversion::DoubleToStringConverter & converter =
@@ -42,7 +42,7 @@ NumberToString<double>::operator()(double val)
 
 template <>
 std::string
-NumberToString<float>::operator()(float val)
+NumberToString<float>::operator()(float val) const
 {
   char                                               buf[256];
   const double_conversion::DoubleToStringConverter & converter =
--- a/Modules/Core/Common/src/itkNumberToString.cxx
+++ b/Modules/Core/Common/src/itkNumberToString.cxx
@@ -16,10 +16,50 @@
  *
  *=========================================================================*/
 #include "itkNumberToString.h"
-#include "itkNumericTraits.h"
-#include "double-conversion/double-conversion.h"
+#include "double-conversion/double-to-string.h"
 
 #include <sstream>
+
+namespace
+{
+
+bool
+ConvertToShortest(const double_conversion::DoubleToStringConverter & converter,
+                  const double                                       val,
+                  double_conversion::StringBuilder &                 builder)
+{
+  return converter.ToShortest(val, &builder);
+}
+
+bool
+ConvertToShortest(const double_conversion::DoubleToStringConverter & converter,
+                  const float                                        val,
+                  double_conversion::StringBuilder &                 builder)
+{
+  // Call the converter member function that is specific for single-precision `float`.
+  return converter.ToShortestSingle(val, &builder);
+}
+
+template <typename TValue>
+std::string
+FloatingPointNumberToString(const TValue val)
+{
+  // Declare a buffer, large enough for strings like:
+  // "-100000000000000000000" (-1e20, either float or double, 23 chars)
+  // "-1.7976931348623157e+308" (-DBL_MAX, 25 chars)
+  // "-0.0000033333333333333333" (-3e-005/9.0, 26 chars)
+  char buf[32];
+
+  double_conversion::StringBuilder builder(buf, sizeof(buf));
+
+  if (!ConvertToShortest(double_conversion::DoubleToStringConverter::EcmaScriptConverter(), val, builder))
+  {
+    itkGenericExceptionMacro(<< "Conversion failed for " << val);
+  }
+  return std::string(builder.Finalize());
+}
+
+} // namespace
 
 namespace itk
 {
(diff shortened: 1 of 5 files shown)


thewtex (approved):
Great to see the improved coverage :green_heart:

N-Dekker:
FYI, some interesting differences between the output of `std::ostringstream` and `itk::NumberToString<double>`  that I encountered while writing the unit tests (using NL = `std::numeric_limits<double>`):

input | stream.str() | itk::NumberToString<double>
---|-------------|-----------------------------------
-0.0 | -0 | 0
 NL::lowest() | -1.79769e+308 | -1.7976931348623157e+308
 NL::epsilon() | 2.22045e-16 | 2.220446049250313e-16
1e-6| 1e-06 | 0.000001
 1e20 | 1e+20 | 100000000000000000000
 NL::denorm_min() | 4.94066e-324 | 5e-324
 NL::min() | 2.22507e-308 | 2.2250738585072014e-308
 NL::quiet_NaN() | nan | NaN
 NL::infinity() | inf | Infinity

Modules/Core/Common/src/itkNumberToString.cxx
```diff
@@ -16,45 +16,66 @@
...
+FloatingPointNumberToString(const TValue val)
 {
-  char                                               buf[256];
-  const double_conversion::DoubleToStringConverter & converter =
-    double_conversion::DoubleToStringConverter::EcmaScriptConverter();
+  // Declare a buffer, large enough for strings like:
+  // "-100000000000000000000" (-1e20, either float or double, 23 chars)
+  // "-1.7976931348623157e+308" (-DBL_MAX, 25 chars)
+  // "-0.0000033333333333333333" (-3e-005/9.0, 26 chars)
+  char buf[32];
```
dzenanz, line 51: Did you consult [this SO topic](https://stackoverflow.com/questions/1701055/what-is-the-maximum-length-in-chars-needed-to-represent-any-double-value)? Can you explain in the comments why not e.g. 325?

dzenanz: You could add the other two cases from the comment and some of those edge cases from the SO answers as unit tests. That way we wouldn't have to guess how long of a buffer we need.

dzenanz:
Now I see your comment. Perhaps in the case of

input | stream.str() | itk::NumberToString
-- | -- | --
NL::denorm_min() | 4.94066e-324 | 5e-324

`itk::NumberToString` is less precise because of a "short" buffer?

Modules/Core/Common/src/itkNumberToString.cxx
N-Dekker: > Did you consult [this SO topic](https://stackoverflow.com/questions/1701055/what-is-the-maximum-length-in-chars-needed-to-represent-any-double-value)? Can you explain in the comments why not e.g. 325?

@dzenanz  Nice SO topic, thanks!  `325` would make sense if the decimal (C printf "%f") notation is used for any floating point value. If I understand SO correctly. But Google's `EcmaScriptConverter().ToShortest(val)` switches to scientific (exponential) notation when the decimal notation gets too long.

As far as I can see (having looked at various edge cases), `EcmaScriptConverter().ToShortest(val)` never exceeds 26 chars. The original `buf[256]` is certainly much larger than necessary. So I think `buf[26]` should be OK, but I made it `buf[32]`, just to be sure. And because `32` is a power of two, just like `256` in the original code. 😃 

Note that `ToShortest(val)` would have an assert failure on some of those edge cases if the buffer would be smaller than 26:
> Assertion failed: !is_finalized() && position_ + n < buffer_.length(), file ITK\Modules\ThirdParty\DoubleConversion\src\double-conversion\utils.h, line 288

Before extending the unit tests or the comments, do you agree that `buf[32]` is fine?

dzenanz: Can the plain notation (0.000000001) be requested? If yes, then we need a longer buffer. If not, 32 is a good choice.

N-Dekker: > Can the plain notation (0.000000001) be requested?

`itk::NumberToString` does not offer any way to change its floating point format. Internally it uses a `const` reference to the `static` DoubleToStringConverter object returned by `EcmaScriptConverter()`. The properties of this converter object cannot be changed. So the answer is _no_!  

> If yes, then we need a longer buffer. If not, 32 is a good choice.

Thanks 😃 

FYI, Below here the `NumberToString<double>{}(d)` output for `d` = `-1.0/9.0`, `-1.0/90.0`, `-1.0/900.0`, `-1.0/9000.0`, etc. Always below 26 chars:

    "-0.1111111111111111"
    "-0.011111111111111112"
    "-0.0011111111111111111"
    "-0.00011111111111111112"
    "-0.000011111111111111112"
    "-0.000001111111111111111"
    "-1.1111111111111111e-7"
    "-1.1111111111111112e-8"
    "-1.111111111111111e-9"
    "-1.1111111111111111e-10"
	...
    "-1.1111111111111113e-306"
    "-1.1111111111111112e-307"
    "-1.1111111111111113e-308"

dzenanz: Thanks for answers.

dzenanz:
If you wish to add some more unit tests, you can do so in a new PR specifically aimed at increasing test coverage. But that does not seem necessary.

N-Dekker:
> If you wish to add some more unit tests, you can do so in a new PR specifically aimed at increasing test coverage. But that does not seem necessary.

@dzenanz  Thanks, I think it isn't necessary either to add more unit tests. 😃 I did add them especially to ensure that the STYLE commit c00531fe23cea71f61685e50e4b6503ab8f1ae82 wouldn't change the existing behavior. And that should be fine, as the tests are passed now both before and after the STYLE commit.

BTW, the tests might eventually break if the floating point format of `itk::NumberToString` is ever going to be changed. Instead of calling `EcmaScriptConverter()`, `itk::NumberToString` might be changed to use a different `DoubleToStringConverter` object, having different formatting flags. https://github.com/google/double-conversion/blob/master/double-conversion/double-to-string.h#L113-L121  But that's not my proposal, for now...!

N-Dekker:
FYI, I just submitted a related issue: "Maximum string length DoubleToStringConverter::EcmaScriptConverter() ToShortest? 26 chars?", https://github.com/google/double-conversion/issues/146

N-Dekker:
Last weekend I submitted a related pull request to Google's [double-conversion](https://github.com/google/double-conversion) library, which aims to allow further simplifying the implementation of `itk::NumberToString` (as well as similar user code). Please have a look:

https://github.com/google/double-conversion/pull/159 _"Add `DoubleToStringConverter::ToShortestString` overload set"_

PR #2257, opened 2021-01-18
libpng Apple Silicon ARM64 support

thewtex:
Add ARM support to libpng to avoid linker errors with Clang.


dzenanz (approved):
Code looks good.

PR #2262, opened 2021-01-20
ENH: Generate a list of remote modules changed for release notes

thewtex:
Also ignore checkout of invalid hashes.


dzenanz (approved):

jhlegarreta (approved):
Have not tried, but looks OK. Thanks for doing this Matt.

PR #2264, opened 2021-01-22
ENH: Include DiffeomorphicDemonsRegistrationFilter wrapping

tbirdso:
ENH: Copied `DiffeomorphicDemonsRegistrationFilter.wrap` from Nonunit to PDEDeformable module so that it is built with ITK, and added Python test.

## PR Checklist
- [ ] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [ ] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [x] Added test (or behavior not changed)
- [ ] Updated API documentation (or API not changed)
- [ ] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [x] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [ ] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKExamples) for all new major features (if any)

Refer to the [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) for
further development details if necessary.

PR #2265, opened 2021-01-22
WIP: ENH: Use extern template and explicit instantiation to speed up tests

phcerdan:
Add CMake option `ITK_EXTERN_TEMPLATE_AND_EXPLICIT_INSTANTIATION`
to enable this option.
This is only useful if BUILD_TESTING is ON, and it is ignored in the
install tree.

The idea is to speed up compilation instantiating only once the
templates with the set of types used.

Similar instantiation of selected types are used for python wrappings.
The CMake machinery to generate bindings should be re-used here, but
I have done it manually for the selected classes as a proof-of-work.

Opening this PR to compare test times in CI. Setting the option to ON by
default for this purpose.

This option also enable third-parties using ITK to compile their targets
with: `target_compile_definition(target PUBLIC ITK_EXTERN_TEMPLATE_AND_EXPLICIT_INSTANTIATION)`
And provide their own set of files `itkImageSourceExternTemplate.h`
and the associated `itkImageSourceExternTemplate.cxx` to instantiate the types they use.

CI preliminary results:

```
        Description        |    Windows    |     Linux     |     MacOs     |
============================================================================
Without                    |   1h.59m      |   1h.21m      |    2h.7m      |
----------------------------------------------------------------------------
Image,ISource, ItoIFilter  |   1h.25m      |   1h.15m      |    2h.3m      |
----------------------------------------------------------------------------
+ IFileReader              |   xh.xxm      |   xh.xxm      |    xh.xm      |
----------------------------------------------------------------------------
```

Reading [this SO question](https://stackoverflow.com/questions/61477486/using-extern-template-with-third-party-header-only-library), we might need to wait for C++20 modules.

> Unfortunately, there’s no way to avoid these [class] instantiations. An
> explicit instantiation declaration of a class template doesn’t prevent
> (implicit) instantiation of that template; it merely prevents
> instantiating its non-inline, non-template member functions (which is
> often none of them!) because some other translation unit will supply the
> actual function symbols and object code.

## PR Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -620,7 +620,14 @@ if(BUILD_TESTING)
   add_subdirectory(Utilities/InstallTest)
 endif()
 
-
+# Enable extern template and explicit instantiation for classes used in tests
+# The extra header and source file needed for this feature are named with the suffix ExternTemplate.h|cxx
+# and placed in a subfolder named ExternTemplate inside the test folder of their module.
+# For example:
+#  itkImageExternTemplate.h and itkImageExternTemplate.cxx
+#  are placed in in Modules/Core/Common/test/ExternTemplate/
+option(ITK_EXTERN_TEMPLATE_AND_EXPLICIT_INSTANTIATION "Enable extern template and explicit instantiation for classes used in tests to speed up compilation." ON) # TODO turn it OFF by default, ON for CIpurposes.
+mark_as_advanced(ITK_EXTERN_TEMPLATE_AND_EXPLICIT_INSTANTIATION_FOR_TEST)
 
 #-----------------------------------------------------------------------------
 # The subdirectories added below this line should use only the public
--- a/Modules/Core/Common/include/itkImage.h
+++ b/Modules/Core/Common/include/itkImage.h
@@ -391,4 +391,8 @@ private:
 #  include "itkImage.hxx"
 #endif
 
+#ifdef ITK_EXTERN_TEMPLATE_AND_EXPLICIT_INSTANTIATION
+#  include "itkImageExternTemplate.h"
+#endif
+
 #endif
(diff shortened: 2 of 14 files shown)


blowekamp (commented):
Look at the export specification in the [MetaDataObject](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkMetaDataObject.h) explicit instantiate to see what is required to get the compiled template to work with shared libraries in different configurations and on different systems.

I am looking forward to see if there is a build time difference. In some cases like ITK Wrapping and SimpleITK, there can be improvements in the linking ( noticeable with relinking after small changes) phase of the build. However that is not easily timed separately.

dzenanz (commented):
Please report findings when you have completed the tests. Code looks fine.

phcerdan:
Well, first look at compilation time for test, picking two recently opened PRs (picked at random): #2249 | #2264 
```
Linux:
With ExterrnTemplate: 1h.15m
Without: 1h.21min | 1h.18m

MacOs:
With ExterrnTemplate: 2h.3m
Without: 2h.7min | 2h.5m

Windows:
With ExterrnTemplate: 1h.25m
Without: 1h.54min | 1h.59m
```
Linux and MacOs appear to have insignificant time differences, but in windows is a huge >25% reduction (with just 3 classes and not all the ImageTypes)?

phcerdan:
Ok, no time differences, after reading [this SO question](https://stackoverflow.com/questions/61477486/using-extern-template-with-third-party-header-only-library), we might need to wait for C++20 modules.

> Unfortunately, there’s no way to avoid these [class] instantiations. An
> explicit instantiation declaration of a class template doesn’t prevent
> (implicit) instantiation of that template; it merely prevents
> instantiating its non-inline, non-template member functions (which is
> often none of them!) because some other translation unit will supply the
> actual function symbols and object code.

Closing.

PR #2270, opened 2021-01-27
BUG: Wrap MetaDataObject for Array and std::vector value types

thewtex:
These are other commonly used types, e.g. when "NRRD_measurement frame"
is set.

Addresses: https://github.com/Project-MONAI/MONAI/issues/1451


aylward (approved):
LGTM!!!   Nice wrapping system, makes these changes seem intuitive.

dzenanz (approved):

Issue #350, opened 2018-12-23
Major performance issue TimeProbe, ResourceProbe constructor


N-Dekker:
Two weeks ago, Theo van Walsum (@tvanwalsum, Erasmus MC, Rotterdam) reported a major performance issue at [discourse.itk.org](https://discourse.itk.org/t/elasitx-performance-issue-with-itk-4-8/1467):  
> I recently upgraded my elastix registration software, and simultaneously upgrading itk from 4.8 to the 4.13. This greatly affected the registration timings (3 to 4-fold slower). After some testing, I found it this problem starts with itk 4.9.

While Theo encountered the issue with ITK 4.8, using VS2013, it is still there with [ITK 5 beta 3](https://github.com/InsightSoftwareConsortium/ITK/tree/v5.0b03), using VS2017.

It appears that this performance issue is caused by an enhancement of itk::ResourceProbe, introduced with ITK 4.9. This enhancement includes calling `GetSystemInformation()` in the default-constructor of `ResourceProbe`:  https://github.com/InsightSoftwareConsortium/ITK/blob/v4.9.0/Modules/Core/Common/include/itkResourceProbe.hxx#L51 Which causes a major performance drop of its derived classes, including `itk::TimeProbe`.

It can be reproduced by profiling the following example: 

      double meanValues[263];

      for (auto&& meanValue : meanValues)
      {
        itk::TimeProbe timeProbe;
        timeProbe.Start();
        timeProbe.Stop();
        meanValue = timeProbe.GetMean();
      }

On Visual Studio 2017 Version 15.9.4 64-bit Release, I observed that this code takes more than 25 seconds. When I manually remove the `GetSystemInformation()` call from the default-constructor of ResourceProbe (at https://github.com/InsightSoftwareConsortium/ITK/blob/v5.0b03/Modules/Core/Common/include/itkResourceProbe.hxx#L52), the very same code takes less than 0.001 seconds!

N-Dekker:
@hyunjaeKang I think this issue is related to an ITK contribution of yours, right? Can you please have a look? Specifically, do you think it would be OK to remove the `GetSystemInformation()` call from the default-constructor of `itk::ResourceProbe`?

N-Dekker:
Update: I'm preparing a fix for this issue... please wait!

dzenanz:
[...]

PR #351, opened 2018-12-24
PERF: Remove SystemInformation data from ResourceProbe, fix issue #350

N-Dekker:
ITK 4.9 added system info to `itk::ResourceProbe` (the base class of
`itk::TimeProbe` and `itk::MemoryProbe`), by calling `GetSystemInformation()`
in its constructor. This function call appeared very time consuming.
Using Visual Studio 2017 (Release configuration), it was observed that a single
`GetSystemInformation()` function call took more than 0.1 second.

This commit removes the `this->GetSystemInformation()` function call from
ResourceProbe, deprecates the member function, removes the related data
members, and only retrieves the system info when and where it is actually being
used: in `PrintSystemInformation` and `PrintJSONSystemInformation`.

Fixes issue #350, "Major performance issue TimeProbe, ResourceProbe
constructor", which was based on the analysis of an `elastix` example
case by Theo van Walsum (@tvanwalsum).

PR #2273, opened 2021-01-27
STYLE: Remove empty ResourceProbe legacy function GetSystemInformation()

N-Dekker:
Follow-up to:
"PERF: Remove SystemInformation data from ResourceProbe, fix issue #350"
Pull request https://github.com/InsightSoftwareConsortium/ITK/pull/351
Commit 747f3d143359cf15c3c4ed9e266125f96d198f8d, 2018-12-24

### DIFF
--- a/Modules/Core/Common/include/itkResourceProbe.hxx
+++ b/Modules/Core/Common/include/itkResourceProbe.hxx
@@ -593,16 +593,6 @@ ResourceProbe<ValueType, MeanType>::PrintJSONSystemInformation(std::ostream & os
   os << "  }";
 }
 
-
-// This protected member function that was introduced with ITK 4.8 has been deprecated
-// as of ITK 5.0. Please do not call or override this member function.
-#if !defined(ITK_LEGACY_REMOVE)
-template <typename ValueType, typename MeanType>
-void
-ResourceProbe<ValueType, MeanType>::GetSystemInformation()
-{}
-#endif
-
 } // end namespace itk
 
 #endif
--- a/Modules/Core/Common/include/itkResourceProbe.h
+++ b/Modules/Core/Common/include/itkResourceProbe.h
@@ -177,9 +177,6 @@ protected:
   void
   PrintJSONvar(std::ostream & os, const char * varName, T varValue, unsigned int indent = 4, bool comma = true);
 
-  /** Obsolete member function from ITK 4.8 - 4.13. Does not do anything anymore. */
-  itkLegacyMacro(virtual void GetSystemInformation());
-
   /** Cause the object to print itself out. */
   virtual void
   Print(std::ostream & os, Indent indent) const;


Modules/Core/Common/include/itkResourceProbe.hxx
```diff
@@ -558,16 +558,6 @@ ResourceProbe<ValueType, MeanType>::PrintJSONSystemInformation(std::ostream & os
   os << "  }";
 }
 
-
-// This protected member function that was introduced with ITK 4.8 has been deprecated
-// as of ITK 5.0. Please do not call or override this member function.
```
dzenanz, line 563: I think the policy is to remove deprecated things at the beginning of a new version. When starting 5.0 development, @hjmjohnson removed ITKv3 compatibility and everything else which was declared deprecated during 4.x. We declared this deprecated as of 5.0, so removing it in 5.2 already seems a bit fast.

N-Dekker: > We declared this deprecated as of 5.0, so removing it in 5.2 already seems a bit fast.

Well, it was deprecated more than two years ago. In another decade. It feels like a long time... long before the corona pandemic...  😢 

Technically, I would prefer to prevent any user from calling `GetSystemInformation()` (as it does nothing useful), as well as from overriding the function (as it is not being called by internally by `ResourceProbe` anymore). The sooner the better. But if you think otherwise, I'll leave it up to you.

hjmjohnson: @N-Dekker   I think we should start pushing for ITKv6.0 that only supports C++14 (or C++17), cmake 3.16+, python 3.6+ and make another round of deprecations.

:)

hjmjohnson:
@N-Dekker Does building with ITK_LEGACY_REMOVE and ITK_FUTURE_LEGACY_REMOVE avoid the problem you are having?

N-Dekker:
> @N-Dekker Does building with ITK_LEGACY_REMOVE and ITK_FUTURE_LEGACY_REMOVE avoid the problem you are having?

@hjmjohnson Thanks for asking Hans, I don't have a big problem with this empty (protected) `GetSystemInformation()` member function. I just proposed to remove it, because it isn't useful anymore. And it has been deprecated more than two years ago already.

But then Dženan (@dzenanz) explained that the function should still be there in ITKv5.2, so hopefully it can be removed before ITKv6! 

You may also just close this PR without merging, if you like.

hjmjohnson (approved):
We are past ITK 5.2 now.  I think this useless function can go.

dzenanz:
I vote to wait until 5.3.0 final is tagged before merging this.

By the way, PerformanceBenchmarking is [currently broken on unix](https://open.cdash.org/viewBuildError.php?buildid=7456319):
`Modules/Core/Common/include/itkFixedArray.h:266:29: error: 'constexpr const ValueType& itk::FixedArray<TValue, VLength>::operator[](unsigned int) const' cannot be overloaded`
from `PerformanceBenchmarkingInformation.cxx:21`.

N-Dekker:
> By the way, PerformanceBenchmarking is [currently broken on unix](https://open.cdash.org/viewBuildError.php?buildid=7456319):
> `Modules/Core/Common/include/itkFixedArray.h:266:29: error: 'constexpr const ValueType& itk::FixedArray<TValue, VLength>::operator[](unsigned int) const' cannot be overloaded`

Thanks @dzenanz But does it compile C++14? I see it also says:

> Modules/Core/Common/include/itkSize.h:481:27: error: 'MakeSize' function uses 'auto' type specifier without trailing return type

C++14 introduced that exactly! 😺

dzenanz:
I created a [PR](https://github.com/InsightSoftwareConsortium/ITKPerformanceBenchmarking/pull/92) to check whether that is the issue.

N-Dekker:
@dzenanz Now that ITK 5.3.0 is released, is it the right moment for this little PR?

dzenanz:
I am fine with it. What do the others think?

hjmjohnson:
@N-Dekker We are not at version 5.4.0.  Can you update this PR so it can be reviewed and finalized?

hjmjohnson:
@N-Dekker I think it is time to resurrect and clear this PR.  Would you mind resolving the conflicts and poking the PR for re-review?

hjmjohnson (approved):

dzenanz (approved):

PR #2278, opened 2021-02-01
symmetric eigen dimension

thewtex:
- STYLE: Add SymmetricEigenAnalysisImageFilter::SetOrderEigenValuesBy
- ENH: Set default dimension for SymmetricEigenAnalysisImageFilter

### DIFF
--- a/Modules/Filtering/ImageIntensity/include/itkSymmetricEigenAnalysisImageFilter.h
+++ b/Modules/Filtering/ImageIntensity/include/itkSymmetricEigenAnalysisImageFilter.h
@@ -204,9 +204,6 @@ extern ITKImageIntensity_EXPORT std::ostream &
  * OrderByMagnitude:  |lambda_1| < |lambda_2| < .....
  * DoNotOrder:        Default order of eigen values obtained after QL method
  *
- * The user of this class is explicitly supposed to set the dimension of the
- * 2D matrix using the SetDimension() method.
- *
  * \ingroup IntensityImageFilters  MultiThreaded  TensorObjects
  *
  * \ingroup ITKImageIntensity
@@ -283,7 +280,7 @@ public:
 #endif
 
 protected:
-  SymmetricEigenAnalysisImageFilter() = default;
+  SymmetricEigenAnalysisImageFilter() { this->SetDimension(TInputImage::ImageDimension); }
   ~SymmetricEigenAnalysisImageFilter() override = default;
 };
 
--- a/Modules/Filtering/ImageIntensity/include/itkSymmetricEigenAnalysisImageFilter.h
+++ b/Modules/Filtering/ImageIntensity/include/itkSymmetricEigenAnalysisImageFilter.h
@@ -106,6 +106,12 @@ public:
       m_Calculator.SetOrderEigenValues(false);
     }
   }
+  void
+  SetOrderEigenValuesBy(EigenValueOrderEnum order)
+  {
+    this->OrderEigenValuesBy(order);
+  }
+
 
 private:
   CalculatorType m_Calculator;
--- a/Modules/Filtering/ImageIntensity/include/itkSymmetricEigenAnalysisImageFilter.h
+++ b/Modules/Filtering/ImageIntensity/include/itkSymmetricEigenAnalysisImageFilter.h
@@ -111,6 +111,19 @@ public:
   {
     this->OrderEigenValuesBy(order);
   }
+  EigenValueOrderEnum
+  GetOrderEigenValuesBy() const
+  {
+    if (m_Calculator.GetOrderEigenMagnitudes())
+    {
+      return EigenValueOrderEnum::OrderByMagnitude;
+    }
+    if (m_Calculator.GetOrderEigenValues())
+    {
+      return EigenValueOrderEnum::OrderByValue;
+    }
+    return EigenValueOrderEnum::DoNotOrder;
+  }
 
 
 private:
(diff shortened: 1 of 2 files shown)


dzenanz (approved):

Modules/Filtering/ImageIntensity/include/itkSymmetricEigenAnalysisImageFilter.h
```diff
@@ -106,6 +106,12 @@ class SymmetricEigenAnalysisFunction
       m_Calculator.SetOrderEigenValues(false);
     }
   }
+  void
+  SetOrderEigenValuesBy(EigenValueOrderEnum order)
+  {
+    this->OrderEigenValuesBy(order);
+  }
+
```
Leengit, line 114: I do not know what motivates adding the `SetOrderEigenValuesBy` method as a synonym of the `OrderEigenValuesBy` method of the `SymmetricEigenAnalysisFunction` class, so I do not know if the same should be added for the other classes in this file ... or elsewhere in ITK:
```shell
$ fgrep -Rl OrderEigenValuesBy *
Modules/Filtering/ImageFeature/include/itkHessian3DToVesselnessMeasureImageFilter.hxx
Modules/Filtering/ImageIntensity/include/itkSymmetricEigenAnalysisImageFilter.h
Modules/Filtering/ImageIntensity/test/itkSymmetricEigenAnalysisImageFilterTest.cxx
```

jhlegarreta, line 114: Is this method tested in the test file?

Modules/Filtering/ImageIntensity/include/itkSymmetricEigenAnalysisImageFilter.h
```diff
@@ -277,7 +280,7 @@ class SymmetricEigenAnalysisImageFilter
 #endif
 
 protected:
-  SymmetricEigenAnalysisImageFilter() = default;
+  SymmetricEigenAnalysisImageFilter() { this->SetDimension(TInputImage::ImageDimension); }
```
Leengit, line 283: Several other ITK filter classes define and set
```cpp
static constexpr unsigned int ImageDimension = TInputImage::ImageDimension;
```
If that is useful in this case -- e.g., so that the `GetDimension` method need not call to `this->GetFunctor().GetDimension()` -- then we could leverage that into this edit too, using `Self::ImageDimension` instead of `TInputImage::ImageDimension`.

Leengit (approved):
Looks good, thank you.

phcerdan (approved):
ImageDimension is not the same than matrix dimension of the pixels for this filter.

I didn’t add the default provided in this patch because it wasn’t clear why a 2D image would have pixels as 2D matrices and a 3D image, 3D matrices. But if that is the common case, all good.

Modules/Filtering/ImageIntensity/include/itkSymmetricEigenAnalysisImageFilter.h
thewtex: @jhlegarreta good catch! :eagle: :eye: I added a `ITK_TEST_SET_GET_VALUE` check and the missing `Get` method.

jhlegarreta: Not sure if I see the additions in the file test beyond the change that was already present to adapt to the new constructor: maybe you did not commit those changes :thinking: ?

thewtex: Good idea to check. Looking into the details, the call in `itk::Hessian3DToVesselnessMeasureImageFilter` is internal, instead of the interface.

Modules/Filtering/ImageIntensity/include/itkSymmetricEigenAnalysisImageFilter.h
thewtex: If I follow, correctly the suggestion is to move the initialization down into the functor, which would be nice. However, in this case, looking into the details, the functor is no longer templated over the image type (it just has the pixel type), losing the information on the image dimension.

thewtex:
> I didn’t add the default provided in this patch because it wasn’t clear why a 2D image would have pixels as 2D matrices and a 3D image, 3D matrices. But if that is the common case, all good. 

Yes, I think this hits the use cases I can think of. But, if a different matrix dimension is desired, it can still be set after initialization. Without reading the class documentation, the current default of `0` was biting me with a segfault :shark: .

thewtex:
Thanks, all, for the reviews!

Modules/Filtering/ImageIntensity/test/itkSymmetricEigenAnalysisImageFilterTest.cxx
```diff
@@ -120,6 +121,8 @@ class SymmetricEigenAnalysisImageFilterHelper : public SymmetricEigenAnalysisIma
     filter->SetFunctor(filter->GetFunctor());
 
     filter->OrderEigenValuesBy(order);
+    filter->SetOrderEigenValuesBy(order);
+    // ITK_TEST_SET_GET_VALUE(order, filter->GetDimension());
```
jhlegarreta, line 125: Sorry to comment on an already merged PR, could not have a look before: is testing `GetOrderEigenValuesBy` here being added in a separate PR?

jhlegarreta, line 125: Sorry to comment on an already merged PR, could not have a look before: is testing `GetOrderEigenValuesBy` here being added in a separate PR?

jhlegarreta (commented):
Thanks for doing this Matt.

An inline comment.

Modules/Filtering/ImageIntensity/test/itkSymmetricEigenAnalysisImageFilterTest.cxx
thewtex: @jhlegarreta good catch, thanks for the review -- that line should be uncommented. I'll create a follow-up.

thewtex: PR: https://github.com/InsightSoftwareConsortium/ITK/pull/2283

PR #2279, opened 2021-02-02
BUG: Fix `itk::VectorGradientMagnitudeImageFilter` uninitialized memory defect

### LINKED ISSUE #2275: Valgrind defect detected in VectorGradientMagnitudeImageFilter with FixedArray
4 valgrind defect detected:

https://open.cdash.org/viewDynamicAnalysis.php?buildid=7011436

Sample:

```
{
   
   Memcheck:Cond
   fun:_ZNSt7__equalILb0EE5equalIPKfS3_EEbT_S4_T0_
   fun:_ZSt11__equal_auxIPKfS1_EbT_S2_T0_
   fun:_ZSt5equalIPKfS1_EbT_S2_T0_
   fun:_ZNK3itk10FixedArrayIfLj3EEeqERKS1_
   fun:_ZNK3itk10FixedArrayIfLj3EEneERKS1_
   fun:_ZN3itk34VectorGradientMagnitudeImageFilterINS_23RGBToVectorImageAdaptorINS_5ImageINS_8RGBPixelItEELj2EEEEEfNS2_IfLj2EEEE20SetDerivativeWeightsENS_10FixedArrayIfLj3EEE
   fun:_Z42itkVectorGradientMagnitudeImageFilterTest1iPPc
   fun:main
}
```

Using demangler.com I see the following is used:  `itk::VectorGradientMagnitudeImageFilter<itk::RGBToVectorImageAdaptor<itk::Image<itk::RGBPixel<unsigned short>, 2u> >, float, itk::Image<float, 2u> >::SetDerivativeWeights(itk::FixedArray<float, 3u>)` 

This type of defect is likely due to a value not fully being initialized.

jhlegarreta:
Fix `itk::VectorGradientMagnitudeImageFilter`  uninitialized memory defect.

Fixes #2275.

## PR Checklist

### DIFF
--- a/Modules/Filtering/ImageGradient/include/itkVectorGradientMagnitudeImageFilter.h
+++ b/Modules/Filtering/ImageGradient/include/itkVectorGradientMagnitudeImageFilter.h
@@ -162,7 +162,9 @@ public:
   using InputImagePointer = typename InputImageType::Pointer;
   using OutputImagePointer = typename OutputImageType::Pointer;
 
-  /** The dimensionality of the input and output images. */
+  /** The dimensionality of the input and output images. Dimensionality
+   * of the two images is assumed to be the same. */
+  static constexpr unsigned int InputImageDimension = TInputImage::ImageDimension;
   static constexpr unsigned int ImageDimension = TOutputImage::ImageDimension;
 
   /** Length of the vector pixel type of the input image. */
@@ -274,6 +276,7 @@ public:
 
 #ifdef ITK_USE_CONCEPT_CHECKING
   // Begin concept checking
+  itkConceptMacro(SameDimensionCheck, (Concept::SameDimension<InputImageDimension, ImageDimension>));
   itkConceptMacro(InputHasNumericTraitsCheck, (Concept::HasNumericTraits<typename InputPixelType::ValueType>));
   itkConceptMacro(RealTypeHasNumericTraitsCheck, (Concept::HasNumericTraits<RealType>));
   // End concept checking
--- a/Modules/Filtering/ImageGradient/include/itkVectorGradientMagnitudeImageFilter.h
+++ b/Modules/Filtering/ImageGradient/include/itkVectorGradientMagnitudeImageFilter.h
@@ -225,17 +225,22 @@ public:
   }
 #endif
 
-  using WeightsType = FixedArray<TRealType, VectorDimension>;
+  using ComponentWeightsType = FixedArray<TRealType, VectorDimension>;
+  using DerivativeWeightsType = FixedArray<TRealType, ImageDimension>;
+#if !defined(ITK_LEGACY_REMOVE)
+  using WeightsType ITK_DEPRECATED_MSG("Use DerivativeWeightsType or ComponentWeightsType instead.") =
+    ComponentWeightsType;
+#endif
 
   /** Directly Set/Get the array of weights used in the gradient calculations.
       Note that calling UseImageSpacingOn will clobber these values. */
-  itkSetMacro(DerivativeWeights, WeightsType);
-  itkGetConstReferenceMacro(DerivativeWeights, WeightsType);
+  itkSetMacro(DerivativeWeights, DerivativeWeightsType);
+  itkGetConstReferenceMacro(DerivativeWeights, DerivativeWeightsType);
 
   /** Set/Get the array of weightings for the different components of the
       vector.  Default values are 1.0. */
-  itkSetMacro(ComponentWeights, WeightsType);
-  itkGetConstReferenceMacro(ComponentWeights, WeightsType);
+  itkSetMacro(ComponentWeights, ComponentWeightsType);
+  itkGetConstReferenceMacro(ComponentWeights, ComponentWeightsType);
 
   /** Set/Get principle components calculation mode.  When this is set to TRUE/ON,
       the gradient calculation will involve a principle component analysis of
@@ -478,13 +483,13 @@ protected:
   }
 
   /** The weights used to scale derivatives during processing */
-  WeightsType m_DerivativeWeights;
+  DerivativeWeightsType m_DerivativeWeights;
 
   /** These weights are used to scale
       vector component values when they are combined to produce  a scalar.  The
       square root */
-  WeightsType m_ComponentWeights;
-  WeightsType m_SqrtComponentWeights;
+  ComponentWeightsType m_ComponentWeights;
+  ComponentWeightsType m_SqrtComponentWeights;
 
 private:
   bool m_UseImageSpacing;
--- a/Modules/Filtering/ImageGradient/test/itkVectorGradientMagnitudeImageFilterTest1.cxx
+++ b/Modules/Filtering/ImageGradient/test/itkVectorGradientMagnitudeImageFilterTest1.cxx
@@ -63,14 +63,14 @@ itkVectorGradientMagnitudeImageFilterTest1(int argc, char * argv[])
   auto useImageSpacing = static_cast<bool>(std::stoi(argv[3]));
   ITK_TEST_SET_GET_BOOLEAN(filter, UseImageSpacing, useImageSpacing);
 
-  auto                    derivativeWeightsValue = static_cast<FilterType::WeightsType::ValueType>(std::stod(argv[4]));
-  FilterType::WeightsType derivativeWeights;
+  auto derivativeWeightsValue = static_cast<FilterType::DerivativeWeightsType::ValueType>(std::stod(argv[4]));
+  FilterType::DerivativeWeightsType derivativeWeights;
   derivativeWeights.Fill(derivativeWeightsValue);
   filter->SetDerivativeWeights(derivativeWeights);
   ITK_TEST_SET_GET_VALUE(derivativeWeights, filter->GetDerivativeWeights());
 
-  auto                    componentWeightsValue = static_cast<FilterType::WeightsType::ValueType>(std::stod(argv[5]));
-  FilterType::WeightsType componentWeights;
+  auto componentWeightsValue = static_cast<FilterType::ComponentWeightsType::ValueType>(std::stod(argv[5]));
+  FilterType::ComponentWeightsType componentWeights;
   componentWeights.Fill(componentWeightsValue);
   filter->SetComponentWeights(componentWeights);
   ITK_TEST_SET_GET_VALUE(componentWeights, filter->GetComponentWeights());


jhlegarreta:
Added a concept check to ensure that the input and output image dimensionality images match. Although I did not dig into the details of the implementation, from what we investigated in the related issue, if not done and the user specifies template image parameters with different dimensionalities, the filter will crash or give some inconsistent result I'd dare to say.

Modules/Filtering/ImageGradient/include/itkVectorGradientMagnitudeImageFilter.h
```diff
@@ -225,7 +227,7 @@ class ITK_TEMPLATE_EXPORT VectorGradientMagnitudeImageFilter : public ImageToIma
   }
 #endif
 
-  using WeightsType = FixedArray<TRealType, VectorDimension>;
+  using WeightsType = FixedArray<TRealType, ImageDimension>;
```
phcerdan, line 230: This change seems fundamental, are you sure of this?
```
  /** Length of the vector pixel type of the input image. */
  static constexpr unsigned int VectorDimension = InputPixelType::Dimension;
```

Modules/Filtering/ImageGradient/include/itkVectorGradientMagnitudeImageFilter.hxx
```diff
@@ -69,7 +64,7 @@ VectorGradientMagnitudeImageFilter<TInputImage, TRealType, TOutputImage>::PrintS
   }
   os << std::endl;
   os << indent << "m_ComponentWeights = ";
-  for (i = 0; i < VectorDimension; i++)
+  for (i = 0; i < ImageDimension; i++)
```
phcerdan, line 67: Maybe change to `unsigned int i` here too.

And ImageDimension is not the same than VectorDimension.

Modules/Filtering/ImageGradient/include/itkVectorGradientMagnitudeImageFilter.hxx
```diff
@@ -155,7 +150,7 @@ VectorGradientMagnitudeImageFilter<TInputImage, TRealType, TOutputImage>::Before
   Superclass::BeforeThreadedGenerateData();
 
   // Calculate the square-roots of the component weights.
-  for (unsigned i = 0; i < VectorDimension; ++i)
+  for (unsigned i = 0; i < ImageDimension; ++i)
```
phcerdan, line 153: Again, ImageDimension is not the same than VectorDimension (i.e length/size of the pixel).

phcerdan (changes requested):
The assertion for equal dimension of input/output image seems ok.
But changing VectorDimension ( the length/size of the pixels type) for ImageDimension should be done carefully.

Modules/Filtering/ImageGradient/include/itkVectorGradientMagnitudeImageFilter.hxx
N-Dekker: If you're adjusting this line of code anyway, you may consider the prefix form `++i` instead of the postfix form (`i++`). Sutter & Alexandrescu - _C++ Coding Standards: 101 Rules, Guidelines, and Best Practices_ [28. Prefer the canonical form of ++ and --. Prefer calling the prefix forms](https://www.oreilly.com/library/view/c-coding-standards/0321113586/ch29.html)

Modules/Filtering/ImageGradient/include/itkVectorGradientMagnitudeImageFilter.h
phcerdan: I just realized that this change is what you were discussing with @blowekamp in the associated issue!

jhlegarreta: Pablo, thanks for having had a look. I'm marking as resolved then.

Modules/Filtering/ImageGradient/include/itkVectorGradientMagnitudeImageFilter.hxx
jhlegarreta: Pablo, thanks for having had a look. I'm marking as resolved following [this comment of yours](https://github.com/InsightSoftwareConsortium/ITK/pull/2279#discussion_r568480074).

Modules/Filtering/ImageGradient/include/itkVectorGradientMagnitudeImageFilter.hxx
jhlegarreta: Thanks both. Yes, definitely those would be welcomed changes. However, those are `STYLE` changes, and in this case I preferred to try to only address the bug in the PR so that it becomes clear where it came from. I will note down these to try to address them in a separate PR once we are sure the bug is resolved. Note that I already meant to do such changes in separate PRs as [noted here](https://github.com/InsightSoftwareConsortium/ITK/issues/2275#issuecomment-770134280).

jhlegarreta:
Following the CI build failures, looks like the baseline images need to be updated. Will do that later tonight.

blowekamp:
The changes look good to me. Thanks for not adding style changes to the bug fix.

I have not looked at the code closely, what is the explanation for the baseline images changing? Do we expect this change to effect the results of users?

issakomi:
```
- using WeightsType = FixedArray<TRealType, VectorDimension>;
+ using WeightsType = FixedArray<TRealType, ImageDimension>;
```

AFAIK, here was the issue with initialized values in that fixed array, detected by Valgrind.

But, is _m_ComponetsWeights_ really related to image dimension and not to vector dimension? Somehow suspicious, comment is still `Set/Get the array of weightings for the different components of the vector`.

May be we need two `WeightsType`s, one for image dimension and one for vector dimension?

Sorry if i am wrong, it is very theoretical, i use that filter in my app, but for 3D RGB images dimension, so both are 3, can not test.

jhlegarreta:
@issakomi Thanks for the work further investigating the issue :100: ! Makes sense to me; will give it a try later, see how the test behaves locally and will report back the results.

jhlegarreta:
Have gone through the affected loops/parts and the changes make sense to me.

The concerned tests are passing locally. Let's see if the failing `WatershedSegmentation1Test1` and `WatershedSegmentation1Test2` example tests do also pass.

blowekamp (approved):
Great work figuring this trickness and wrong sized arrays out!

It should be noted the this is technically, an incompatibly change of types. Perhaps there could be a:
```
#ifndef ITK_LEGACY_REMOVE
using WeightsType = ComponentWeightType;
#endif
```

I have not looked at it closely to determine if this would be the correct size of all cases. It may not be needed but if some thought can be put into it to prevent braking anyones code it would be good.

jhlegarreta:
@blowekamp thanks for the [note](https://github.com/InsightSoftwareConsortium/ITK/pull/2279#pullrequestreview-582318107). Had a look at it, and thought about it: I do not think there is a good solution. The internal variables `m_DerivativeWeights` and `m_ComponentWeights` have now different types, so adding the pre-processor proposed directive, if the `ITK_LEGACY_REMOVE` flag is `OFF`, `WeightsType` would be equivalent to `ComponentWeightsType` which has the former's parameters, and thus users have the `WeightsType` type available as before; however, the type of `m_DerivativeWeights` would be new, and we would break the API.

The only thing that I can think of to maintain backwards compatibility but **also keeping the memory defect around** if using the old API would be:

```
#ifndef ITK_LEGACY_REMOVE
  using DerivativeWeightsType = FixedArray<TRealType, VectorDimension>;
  using ComponentWeightsType = FixedArray<TRealType, VectorDimension>;
  using WeightsType = ComponentWeightsType;
#else
  using DerivativeWeightsType = FixedArray<TRealType, ImageDimension>;
  using ComponentWeightsType = FixedArray<TRealType, VectorDimension>;
#endif
```

Not sure if this makes sense/other insight is welcome.

blowekamp:
@jhlegarreta I suppose my assumption was with the "compatibility" was that  VectorDimension and ImageDimension were the same value, as that was the only case where the old code would be correct.  Thanks for considering the compatibility options. Please do what every you think is best, and lets get this fixed merged!

jhlegarreta:
Made the changes, force pushed.

phcerdan (approved):
Kudos for the bug fix @jhlegarreta!

Before it was wrong, the dimension of `ComponentsWeights` mismatched the algorithm in the Evaluate functions, for example here: https://github.com/InsightSoftwareConsortium/ITK/blob/3ecf6c87d18d51b124589d5a74389244d692109c/Modules/Filtering/ImageGradient/include/itkVectorGradientMagnitudeImageFilter.h#L431-L458

phcerdan:
I would add the following to try to not break API (but show a warning) if the compiler is modern enough:

```cpp
#ifndef ITK_LEGACY_REMOVE
  using DerivativeWeightsType = FixedArray<TRealType, VectorDimension>;
  using ComponentWeightsType = FixedArray<TRealType, VectorDimension>;
  using WeightsType = ComponentWeightsType;
#else
  using DerivativeWeightsType = FixedArray<TRealType, ImageDimension>;
  using ComponentWeightsType = FixedArray<TRealType, VectorDimension>;
  #if __cplusplus >= 201402L
  [[deprecated("Use DerivativeWeightsType or ComponentWeightsType instead")]]
  using WeightsType = ComponentWeightsType;
  #endif
#endif

```

jhlegarreta:
> Kudos for the bug fix @jhlegarreta!
Before it was wrong, the dimension of ComponentsWeights mismatched the algorithm in the Evaluate functions, for example here: 

Thanks Pablo. @issakomi, @blowekamp, @N-Dekker and yourself who took the time to review and propose possible fixes or questioned if a given solution was good made this possible.

jhlegarreta:
> I would add the following to try to not break API (but show a warning) if the compiler is modern enough:
> ```
> #ifndef ITK_LEGACY_REMOVE
>   using DerivativeWeightsType = FixedArray<TRealType, VectorDimension>;
>   using ComponentWeightsType = FixedArray<TRealType, VectorDimension>;
>   using WeightsType = ComponentWeightsType;
> #else
>   using DerivativeWeightsType = FixedArray<TRealType, ImageDimension>;
>   using ComponentWeightsType = FixedArray<TRealType, VectorDimension>;
>   // ----- suggestion here ----- deprecated attribute requires C++14
>   #if __cplusplus >= 201402L
>  [[deprecated("Use DerivativeWeightsType or ComponentWeightsType instead.")]]
>   using WeightsType = ComponentWeightsType;
>   #endif
>   // ---------
> #endif
> ```

@phcerdan I'd dare to say that the deprecation warning would need to to go in the `#ifndef` block, not the `#else` block, since the latter is supposed to be the fix/modern code, where we do not need/define `WeightsType`, right?

Also, I think the only thing that should go inside the  `#if __cplusplus >= 201402L` block would be the warning message, otherwise, the type will not be known if the compiler does not match the condition.

Forgive me if I'm wrong.

phcerdan:
> @phcerdan I'd dare to say that the deprecation warning would need to to go in the `#ifndef` block, not the `#else` block, since the latter is supposed to be the fix/modern code, where we do not need/define `WeightsType`, right?
> 
> Also, I think the only thing that should go inside the  `#if __cplusplus >= 201402L` block would be the warning message, otherwise, the type will not be known if the compiler does not match the condition.
> 
> Forgive me if I'm wrong.

You are right, forgive me, I am not very familiar with the `ITK_LEGACY_REMOVE`, I think this is your correction:
```cpp
#ifndef ITK_LEGACY_REMOVE
  using DerivativeWeightsType = FixedArray<TRealType, VectorDimension>;
  using ComponentWeightsType = FixedArray<TRealType, VectorDimension>;
  #if __cplusplus >= 201402L
  [[deprecated("Use DerivativeWeightsType or ComponentWeightsType instead.")]]
  #endif
  using WeightsType = ComponentWeightsType;
#else
  using DerivativeWeightsType = FixedArray<TRealType, ImageDimension>;
  using ComponentWeightsType = FixedArray<TRealType, VectorDimension>;
#endif
```

jhlegarreta:
> I think this is your correction:

Yes, that was what I meant. Thanks for putting it into code.

I'll do that tonight and will push force.

issakomi:
```
#ifndef ITK_LEGACY_REMOVE
  using DerivativeWeightsType = FixedArray<TRealType, VectorDimension>;
  using ComponentWeightsType = FixedArray<TRealType, VectorDimension>;
  #if __cplusplus >= 201402L
  [[deprecated("Use DerivativeWeightsType or ComponentWeightsType instead.")]]
  #endif
  using WeightsType = ComponentWeightsType;
#else
  using DerivativeWeightsType = FixedArray<TRealType, ImageDimension>;
  using ComponentWeightsType = FixedArray<TRealType, VectorDimension>;
#endif
```
Sorry to interfere, in that version, if ITK_LEGACY_REMOVE is OFF (which is default), it is exactly the same what is taken.

dzenanz:
`using ComponentWeightsType = FixedArray<TRealType, VectorDimension>;` can be outside of `#ifdef ITK_LEGACY_REMOVE` as it is the same in both branches.

phcerdan:
If I understood @issakomi, he is right, nothing is changed with the current code.

Double negations are hard for the brain, I think this is right:

```cpp
using ComponentWeightsType = FixedArray<TRealType, VectorDimension>;
using DerivativeWeightsType = FixedArray<TRealType, ImageDimension>; //<--- change this from  VectorDimension to ImageDimension
#if !defined(ITK_LEGACY_REMOVE) //<-- This would be the default code: ITK_LEGACY_REMOVE not defined
  #if __cplusplus >= 201402L
  [[deprecated("Use DerivativeWeightsType or ComponentWeightsType instead.")]]
  #endif
  using WeightsType = ComponentWeightsType;
#endif
// #else WeightsType is not defined
```

jhlegarreta:
@dzenanz 's [comment](https://github.com/InsightSoftwareConsortium/ITK/pull/2279#issuecomment-774248658) is fair wrt what I proposed.

@phcerdan @issakomi Not sure if I see it the same way:
- `ITK_LEGACY_REMOVE` is `OFF` [by default](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CMakeLists.txt#L331), so the legacy code blocks are not removed by default.
- The member variables and the corresponding `Set/Get` macro methods' prototypes are being changed, and hence the API is being broken. So the backwards compatibility needs to be provided not just to keep the old type around, but to define the member variables the same way acknowledging that the memory defect will be around in that case. So to me that means that we should necessarily have
```
  using DerivativeWeightsType = FixedArray<TRealType, VectorDimension>;
  using ComponentWeightsType = FixedArray<TRealType, VectorDimension>;
  using WeightsType = ComponentWeightsType;
```
when the old code is to be kept, which is when `#ifndef ITK_LEGACY_REMOVE` (or `#if !defined(ITK_LEGACY_REMOVE)`) turns out to be `False`, which is when ITK_LEGACY_REMOVE is `True`. (`Remove current legacy code completely` according to its definition)

And that I guess would allow users not to have `SetDerivativeWeights()/GetDerivativeWeights()` broken because the type has changed.

So I do not see how the proposed solution in [here](https://github.com/InsightSoftwareConsortium/ITK/pull/2279#issuecomment-774250997) is backwards compatible when the image and component/vector dimensions are different.

Do we want the fix to be backwards compatible ONLY when the image and component/vector dimensions are the same and break otherwise to have the defect fixed in any case?

issakomi:
> If I understood @issakomi

Sorry for bad structured sentence, you understood correctly, of course. I have updated the comment.

phcerdan:
>Do we want the fix to be backwards compatible ONLY when the image and component/vector dimensions are the same and break otherwise to have the defect fixed in any case?

I think yes! We break code only when is bugged.

We have to assume that existing code with `ImageDimension != VectorDimension` (i.e `DerivativeWeightsType!=ComponentsWeightType`) is bugged and we are glad to break it because it's good for users. So no backwards compatibility is needed for this.

> And I guess that would allow users not to have SetDerivativeWeights()/GetDerivativeWeights() broken because the type for m_ DerivativeWeights has changed.

It would only break in the case that is worth to break, `ImageDImension!=VectorDimension`. I don't think this case is common, or the bug wouldn't survived for this long. 

The only API break we are tying to avoid is with a non existing `WeightsType` typedef.

ITK_LEGACY_REMOVE=OFF, the default, means legacy code is alive. Again, double negatives are weird to think about it, at least for me.
```cpp
#if not remove_it:
// keep it
#endif
```
https://github.com/InsightSoftwareConsortium/ITK/blob/3ecf6c87d18d51b124589d5a74389244d692109c/Modules/Core/Common/include/itkMacro.h#L580-L594

The code means we keep `WeightsType` for now by default, and will remove it in the future (or when `ITK_LEGACY_REMOVE=ON`)

jhlegarreta:
Push forced with Pablo's proposal.

issakomi:
does it compile? For me it compiles (GCC 8.3.0) with
`using WeightsType [[deprecated("Use DerivativeWeightsType or ComponentWeightsType instead.")]] = ComponentWeightsType;
`
Let me check 2 minutes, may be i am wrong

issakomi:
```
[ 50%] Building CXX object CMakeFiles/test007.dir/test007.cpp.o
In file included from /home/r/tmp/test_rgb/src/test007.cpp:45:
/home/r/itkf/ITK/Modules/Filtering/ImageGradient/include/itkVectorGradientMagnitudeImageFilter.h:234:3: error: expected unqualified-id before ‘using’
   using WeightsType = ComponentWeightsType;
   ^~~~~
make[2]: *** [CMakeFiles/test007.dir/build.make:76: CMakeFiles/test007.dir/test007.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:83: CMakeFiles/test007.dir/all] Error 2
make: *** [Makefile:91: all] Error 2
```

Change to
```cpp
using ComponentWeightsType = FixedArray<TRealType, VectorDimension>;
using DerivativeWeightsType = FixedArray<TRealType, ImageDimension>;
  #if __cplusplus >= 201402L
  using WeightsType [[deprecated("Use DerivativeWeightsType or ComponentWeightsType instead.")]] = ComponentWeightsType;
  #else
  using WeightsType = ComponentWeightsType;
  #endif
#endif
```

phcerdan:
@issakomi thanks for testing it, you are right, the correct placement of attributes for `using` is before `=`.
```cpp
using WeightsType [[deprecated("Use DerivativeWeightsType or ComponentWeightsType instead.")]] = ComponentWeightsType; 
```

phcerdan:
Maybe for other PR, but we can add to itkMacro.h the following:

```cpp
#if __cplusplus >= 201402L
# define ITK_DEPRECATED [[deprecated]]
# define ITK_DEPRECATED_MSG(msg) [[deprecated(msg)]]
#else
# define ITK_DEPRECATED
# define ITK_DEPRECATED_MSG
#endif
```

And then just:
```cpp
using WeightsType ITK_DEPRECATE_MSG("Use DerivativeWeightsType or ComponentWeightsType instead.") = ComponentWeightsType; 
```
without extra `#ifdef.`

Modules/Filtering/ImageGradient/include/itkVectorGradientMagnitudeImageFilter.h
```diff
@@ -225,17 +227,24 @@ class ITK_TEMPLATE_EXPORT VectorGradientMagnitudeImageFilter : public ImageToIma
...
 #endif
 
-  using WeightsType = FixedArray<TRealType, VectorDimension>;
+  using ComponentWeightsType = FixedArray<TRealType, VectorDimension>;
+  using DerivativeWeightsType = FixedArray<TRealType, ImageDimension>;
+#if !defined(ITK_LEGACY_REMOVE)
+#  if __cplusplus >= 201402L
+  [[deprecated("Use DerivativeWeightsType or ComponentWeightsType instead.")]]
+#  endif
+  using WeightsType = ComponentWeightsType;
```
phcerdan, line 236: Remove the `#if __cplusplus` branch. It will be handled in future PR with a macro.

phcerdan (changes requested):

Modules/Filtering/ImageGradient/include/itkVectorGradientMagnitudeImageFilter.h
phcerdan: Or, if you wish, this is valid ITK code:
```cpp
  using WeightsType ITK_DEPRECATED_MSG("Use DerivativeWeightsType or ComponentWeightsType instead.") =
    ComponentWeightsType;
#endif
```

issakomi: Interesting, looks like here
```
#    if (__GNUC__ * 100 + __GNUC_MINOR__) >= 409 && __cplusplus > 201103L
#      define ITK_COMPILER_CXX_ATTRIBUTE_DEPRECATED 1
#    else
#      define ITK_COMPILER_CXX_ATTRIBUTE_DEPRECATED 0
#    endif
```
is assumed C++11, but i that `[[deprecated]]` thing is definitely C++14. Probably this part

```
...
#    if defined(ITK_COMPILER_CXX_ATTRIBUTE_DEPRECATED) && ITK_COMPILER_CXX_ATTRIBUTE_DEPRECATED
#      define ITK_DEPRECATED [[deprecated]]
#      define ITK_DEPRECATED_MSG(MSG) [[deprecated(MSG)]]
...
```
will cause problems sometimes. May be `__cplusplus > 201103L` above is a bug.

About compiler intrinsic things `__attribute__ ` for GCC and `__declspec` for Visual Studio, not sure, but i use them for address alignment and it was not easy to get it working, not sure how  it works with deprecated, but for alignment one requires prepend and another append.
This things are difficult to test, sometimes they work, sometimes not. BTW, many Visual Studio version have issues with value of ___cplusplus_, updated VS 2019 don't, but earlier versions reported 1998xx and ignored cmake.

jhlegarreta: Thanks both of you for the interest, time, and effort to propose a proper solution to the backwards compatibility issue. I've push forced. Let's see what the builds say.

issakomi: This is what i mean with prepend/append

```
#ifdef __GNUC__
#define DEPRECATED(func) func __attribute__ ((deprecated))
#elif defined(_MSC_VER)
#define DEPRECATED(func) __declspec(deprecated) func
...
```

jhlegarreta: CI builds seem not to hiccup till now/the code is compiling,  but it might with the larger set of compilers in the dashboard.

If I understand correctly, [the above](https://github.com/InsightSoftwareConsortium/ITK/pull/2279#discussion_r571468667) should go into [CMake](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/WriteCompilerDetectionHeader.cmake)? A patch set might be desirable there?

issakomi: @jhlegarreta i am not sure about syntax with `using` and don't know - does it cause build error or simply doesn't work sometimes, so actually i am not ready for PR, if somebody is sure, may be. It is complicated stuff, many combinations, IMHO.

jhlegarreta: So let's keep this in mind if at some point we witness related problems, and get this PR merged if builds are passing so as to keep moving forward. Thanks for all the useful thoughts !

phcerdan: @issakomi it is tricky indeed with the different placement of the attributes among compilers :(
Maybe is worth to have a look at the third party project for these kind of `ifdefs`, as a reference: https://github.com/nemequ/hedley/blob/master/hedley.h 

For the case of deprecated functions, we already have the `itkLegacyMacro`: https://github.com/InsightSoftwareConsortium/ITK/blob/3fd1ff784029f226e976c016b06068a9af0b5d4a/Modules/Core/Common/include/itkMacro.h#L601-L618 that handles the different placement.

It seems the placement is shared among compilers with `using` at least.

phcerdan (approved):
Neat, thanks @jhlegarreta!

Modules/Filtering/ImageGradient/include/itkVectorGradientMagnitudeImageFilter.h
issakomi: I have tested with GCC 8.3.0 and it works good with both C++11 and C++14. Also it seems to work with this GCC version both with prepend and append of __attribute__((deprecated)). Not sure about all compilers, specially older, but seems to be fine for now. Thank you.

blowekamp (approved):
Looks great! Thanks for all the extra work to maintain compatibility!

PR #2282, opened 2021-02-03
ENH: Wrap GradientDescentOptimizerv4 for Python

tbirdso:
## PR Checklist


dzenanz (approved):

thewtex (approved):
In general, I think we should avoiding wrapping these template aliases, since they are essentially duplicates. But, in this case we could make an exception since this is the most commonly used optimizer with the v4 registration framework.

PR #2284, opened 2021-02-03
ENH: Module configuration exec to import

thewtex:
This is a simplication that continues the integration of WrapITK into
ITK. The use of `exec` is a security faux pas, and the use of importlib
instead helps analysis tools like PyInstaller understand dependencies.
The use of WRAPITK_PYTHON_PATH has not been tested or supported since the
WrapITK integration and is removed. Base on Python test runtimes, this change is also
slightly (1%) faster.


dzenanz (commented):
LGTM

Wrapping/Generators/Python/itk/support/itkBase.py
```diff
@@ -246,13 +246,14 @@ def setup(self):
...
         self.old_path = sys.path
-        sys.path = [itkConfig.swig_lib, itkConfig.swig_py] + itkConfig.path + sys.path
+        sys.path = [itkConfig.swig_lib, itkConfig.swig_py] + sys.path
 
     def load(self, name: str):
         self.setup()
         try:
             import importlib
 
+            # import pdb; pdb.set_trace()
```
floryst, line 256: Remove commented debugging line

thewtex: :+1: good catch

hjmjohnson (approved):

PR #2285, opened 2021-02-06
ENH: Add ITK_DEPRECATED and ITK_DEPRECATED_MSG to itkMacro

phcerdan:
ITK_DEPRECATED can be used along ITK_LEGACY_REMOVE to warn the users
about deprecated features before its removal.

[[deprecated]] attribute is a C++14 standard feature, so it is only enabled if
the compiler is modern enough.

https://en.cppreference.com/w/cpp/language/attributes/deprecated

There are compiler specifics attributes prior standardization:
For example in [MSVC](https://docs.microsoft.com/en-us/cpp/cpp/deprecated-cpp?view=msvc-160): `__declspec(deprecated)` and `__declspec(deprecated("message"))`
In gcc: `__attribute__((deprecated("message")))`

So, we could make this feature available for older compilers redefining it per compiler.
However, I am not sure if the placement of the attribute/declspec is in the same position for all compilers, so I opted for only
enable it when the standard is C++14.

See [#2279](https://github.com/InsightSoftwareConsortium/ITK/pull/2279) for an usage example.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkMacro.h
+++ b/Modules/Core/Common/include/itkMacro.h
@@ -617,6 +617,14 @@ OutputWindowDisplayDebugText(const char *);
 #  endif
 #endif
 
+#if __cplusplus >= 201402L
+#  define ITK_DEPRECATED [[deprecated]]
+#  define ITK_DEPRECATED_MSG(msg) [[deprecated(msg)]]
+#else
+#  define ITK_DEPRECATED
+#  define ITK_DEPRECATED_MSG(msg)
+#endif
+
 // Macros to create runtime deprecation warning messages in function
 // bodies.  Example usage:
 //


jhlegarreta:
@phcerdan Thanks for doing this. I think it can be useful. A few thoughts:
- How would `ITK_DEPRECATED` be used?
- I know this requires work and testing, but if other compilers might have a different way of specifying deprecation, can we think of a general macro that could remain valid at least for the major compilers? Maybe that would require getting the standard statement (e.g. `using WeightsType = ComponentWeightsType;`) to be deprecated and the message as arguments for the macro so that the appropriate deprecation statement can be constructed? This would require being able to parse the expression, which involves an additional effort, and not sure if it would work for all or most use cases we can think of or if that would require having a different macro (e.g. `ITK_DEPRECATED_TYPE`, `ITK_DEPRECATED_STRUCT`, etc.) :pensive: ([here](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkMultiThreaderBase.h#L250) is the other use case I could find in the ITK code base).

And if we go for the latter, I think we should have them tested to ensure that a given statement is correctly parsed. Given that this is a general feature that programmers might want to be aware of, adding corresponding examples/explanations in a release note might be also welcome.

I know, lots of work :pray:. I might not be able to develop actively, so I would understand that you prefer to at least have the current solution merged, and leave the rest as an issue that can be tackled in the future.

phcerdan:
Thanks @jhlegarreta for your review.
Reading the failures in CI, I realized we already have a macro `ITK_DEPRECATED` and `ITK_DEPRECATED_MSG` thanks to CMake: https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/WriteCompilerDetectionHeader.cmake

in the build folder, the CMake generated file `itk_compiler_detection.h` contains:
```cpp
#  ifndef ITK_DEPRECATED
#    if defined(ITK_COMPILER_CXX_ATTRIBUTE_DEPRECATED) && ITK_COMPILER_CXX_ATTRIBUTE_DEPRECATED
#      define ITK_DEPRECATED [[deprecated]]
#      define ITK_DEPRECATED_MSG(MSG) [[deprecated(MSG)]]
#    elif ITK_COMPILER_IS_GNU || ITK_COMPILER_IS_Clang
#      define ITK_DEPRECATED __attribute__((__deprecated__))
#      define ITK_DEPRECATED_MSG(MSG) __attribute__((__deprecated__(MSG)))
#    elif ITK_COMPILER_IS_MSVC
#      define ITK_DEPRECATED __declspec(deprecated)
#      define ITK_DEPRECATED_MSG(MSG) __declspec(deprecated(MSG))
#    else
#      define ITK_DEPRECATED
#      define ITK_DEPRECATED_MSG(MSG)
#    endif
#  endif
```
For GNU: `Modules/Core/Common/compilers/ITK_COMPILER_INFO_GNU_CXX.h`
```cpp
#    if (__GNUC__ * 100 + __GNUC_MINOR__) >= 409 && __cplusplus > 201103L
#      define ITK_COMPILER_CXX_ATTRIBUTE_DEPRECATED 1
#    else
#      define ITK_COMPILER_CXX_ATTRIBUTE_DEPRECATED 0
#    endif

```

So your second point is addressed by CMake.
The first point, how to use attributes is a language thing, interested user can follow the link provided.
Closing for now, and going back to the PR #2279 for testing the macro.

A future PR might use use this feature in all (`ITK_LEGACY_REMOVE`) legacy code.

phcerdan:
Summary: Closing because we already have this feature.

PR #2294, opened 2021-02-11
Release hdf5 1.10.7

seanm:


seanm:
@dzenanz @thewtex @hjmjohnson I think this worked... building some variations on my machine now to test more...

dzenanz:
CircleCI complains about:
```text
CMake Error at Modules/ThirdParty/HDF5/src/itkhdf5/CMakeLists.txt:1 (cmake_minimum_required):
  CMake 3.12 or higher is required.  You are running version 3.11.2
```
ITK currently requires 3.10.2. With this patch, that requirement will go up in the default configuration. Should we bump minimum CMake version? If so, should we go for CMake 3.16.3 which comes with [Ubuntu 20.04](https://launchpad.net/ubuntu/focal/+source/cmake) or something else?

seanm:
hmm, I wonder if HDF really requires that.  Seems like a strange thing to increase in a minor version bump.  I'll see if I can find when/why that changed.

Also, I'm looking to be as conservative as possible here, I'd like to get this into the release branch for example.

seanm:
Hmm, their increase from 3.10 to 3.12 is part of a pretty big commit: https://github.com/HDFGroup/hdf5/commit/666205a55f01698a71484f504bc8eb9541800f25

The comment is only: "Updated CMake minimum version to 3.12 and added version checks for Windows features."

So up to you guys about bumping CMake version to match.  I guess otherwise the Apple Silicon issue could be fixed by just cherry picking the exact (small) commit I need.  I just figured it was a good occasion to also bump HDF5 but I did not anticipate this cmake version speedbump.

Modules/ThirdParty/HDF5/src/itkhdf5/CMakeLists.txt
```diff
@@ -1,14 +1,18 @@
-cmake_minimum_required (VERSION 3.10)
+cmake_minimum_required (VERSION 3.12)
 project (HDF5 C)
 # ITK --start
 if(POLICY CMP0083)
```
dzenanz, line 4: This ITK-specific addition can now be removed, given that the identical thing is added just below in the upstream code.

Modules/ThirdParty/HDF5/src/itkhdf5/CMakeLists.txt
```diff
@@ -39,9 +43,14 @@ endif ()
...
 # CMake version 3.14 added option --ignore-eol to compare files
 #     cmake -E compare_files --ignore-eol file1 file2
-#if(CMAKE_VERSION VERSION_LESS "3.14.0" AND WIN32)
-#    MESSAGE(FATAL_ERROR "Windows builds requires a minimum of CMake 3.14")
-#endif()
+set (CMAKE_IGNORE_EOL "--ignore-eol")
+if (CMAKE_VERSION VERSION_LESS "3.14.0")
+  set (CMAKE_IGNORE_EOL "")
+  if (WIN32)
+    message (FATAL_ERROR "Windows builds requires a minimum of CMake 3.14")
```
dzenanz, line 50: This also increases CMake version requirement.

Modules/ThirdParty/HDF5/src/itkhdf5/CMakeLists.txt
seanm: Good catch. It didn't generate a conflict, so I never looked that closely.  Should I add a new commit to remove that? Otherwise I'm not sure which it should be squashed into...

dzenanz: A new commit is fine, preferably at the end.

Modules/ThirdParty/HDF5/src/itkhdf5/CMakeLists.txt
seanm: 3.14 was released on 2019-03-14.  Yes, I think I remember that π release!

So, yeah, someone needs to decide about increasing min cmake or not.

Modules/ThirdParty/HDF5/src/itkhdf5/CMakeLists.txt
seanm: Perfect, that's the easiest too. :)  Done.

Modules/ThirdParty/HDF5/src/README.md
```diff
@@ -6,84 +6,6 @@ tree with only the library source code and [CMake] build system. It is not a
...
-```sh
-   $ git rev-parse --short=8 hdf5-upstream
-```
-
-to get the commit from which the `hdf5-upstream` branch must be started on the
-next update. Edit the `git branch hdf5-upstream` line above to record it, and
-commit this file.
-
-
+Update hdf5 from upstream using the `UpdateFromUpstream.sh` script one directory up.
```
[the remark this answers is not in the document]
thewtex: The HDF5 mangling is still required as far as I am aware, unless there are new developments in this update.

Modules/ThirdParty/HDF5/src/itkhdf5/CMakeLists.txt
thewtex: While I like :pie: :yum:, we should forgo the CMake minimum required bump until ITK 5.3. We have systems running Ubuntu 18.04, and we can resolve the minimum required versions on CI systems, etc. after the 5.2.0 release.

seanm: I'm a little out of the ITK loop...  Is 5.2 branched? What's master currently? What's release currently?

thewtex: As of this morning `master` and `release` were equivalent -- we are in the release candidate phase for 5.2.

thewtex:
Let's please wait until after 5.2.0 to merge due to the bump in CMake minimum required version.

dzenanz:
When rebasing/redoing this on top of 5.2.0 final, please also add a commit which explicitly bumps the version number in ITK to CMake 3.16.3.

dzenanz:
5.2.0 has been tagged! Time to get this PR up to date, including CMake version bump.

seanm:
I may have found another hdf5 issue to fix... and I'm waiting for upstream to merge my `override` additions... but yes, I haven't forgotten about this, and will do it.

thewtex:
Can this please be rebased for 5.2.1?

thewtex:
@seanm yes, great plan.

dzenanz:
CMake version was updated. Can you please continue work on updating HDF5?

seanm:
@dzenanz yes, I have not forgotten, I should have time next week...

seanm:
Closing in favour of: https://github.com/InsightSoftwareConsortium/ITK/pull/2665

PR #2299, opened 2021-02-12
Release fp except ub

seanm:


seanm:
@jcfr as you authored 6db502e61bfeb9998564b3005a6bd7b7b3bfa2c6 could you review?

@thewtex @dzenanz @hjmjohnson  FYI

I'm starting to realize that try_compile seems like an anti-pattern really. It assumes the machine doing the building is the same as the machine being compiled for.

I thought about trying to eliminate the use of try_compile entirely here, but thought this conservative change a better place to start a discussion from.

hjmjohnson (commented):
LGTM.

Modules/Core/Common/src/itkFloatingPointExceptions_unix_feenableexcept_using_fegetenv.cxx
```diff
@@ -87,11 +87,16 @@ itk_fedisableexcept(const fexcept_t excepts)
 
 // Implementation for Intel
 
-#  if (defined ITK_HAS_STRUCT_FENV_T_CONTROL)
+// Note: on Apple, ignore the result of the try_compiles that generate these
```
dzenanz, line 90: `try_compile` was a must 20 years ago (when ITK was largely designed) due to very varied compiler standards adherence. It is problematic for cross-compiling. And it makes configure step longer. Reducing our dependence on it is welcome.

dzenanz (approved):

Modules/Core/Common/src/itkFloatingPointExceptions_unix_feenableexcept_using_fegetenv.cxx
seanm: may give that a shot one day, but for now this small change is probably safer, especailly with 5.2 so close.

hjmjohnson (approved):

thewtex (approved):
Thanks @seanm !

PR #2311, opened 2021-02-14
ENH: Use strongly typed enums for `itk::NiftiImageIO` readable formats

jhlegarreta:
Use strongly typed enums to define the file formats readable by the
`itk::NiftiImageIO` class,

## PR Checklist

### DIFF
--- a/Modules/IO/NIFTI/include/itkNiftiImageIO.h
+++ b/Modules/IO/NIFTI/include/itkNiftiImageIO.h
@@ -28,28 +28,58 @@
 namespace itk
 {
 
-
-/** \class Analyze75Flavor
+/**\class NiftiImageIOEnums
+ * \brief
  * \ingroup ITKIONIFTI
- * Enum used to define way to treat legacy Analyze75 files
  */
-enum class Analyze75Flavor : uint8_t
+class NiftiImageIOEnums
 {
-  /** Behavior introduced in ITK4.0 by NIFTI reader interpreting Analyze files */
-  AnalyzeITK4 = 4,
-  /** Will ignore orientation code and negative pixel dimensions */
-  AnalyzeFSL = 3,
-  /** Will ignore orientation code and respect negative pixel dimensions */
-  AnalyzeSPM = 2,
-  /** Same as AnalyzeITK4 but will show warning about deprecated file format (Default)*/
-  AnalyzeITK4Warning = 1,
-  /** Reject Analyze files as potentially wrong  */
-  AnalyzeReject = 0
+public:
+  /** \class Analyze75Flavor
+   * \ingroup ITKIONIFTI
+   * Enum used to define the way to treat legacy Analyze75 files.
+   */
+  enum class Analyze75Flavor : uint8_t
+  {
+    /** Behavior introduced in ITK4.0 by NIFTI reader interpreting Analyze files */
+    AnalyzeITK4 = 4,
+    /** Will ignore orientation code and negative pixel dimensions */
+    AnalyzeFSL = 3,
+    /** Will ignore orientation code and respect negative pixel dimensions */
+    AnalyzeSPM = 2,
+    /** Same as AnalyzeITK4 but will show warning about deprecated file format (Default)*/
+    AnalyzeITK4Warning = 1,
+    /** Reject Analyze files as potentially wrong  */
+    AnalyzeReject = 0
+  };
+
+  /** \class NiftiFileEnum
+   * \ingroup ITKIONIFTI
+   * Enum used to define the possible file formats readable by this ImageIO implementation.
+   */
+  enum class NiftiFileEnum : int8_t
+  {
+    /** 2-file Nifti (consisting of .hdr and .img file). */
+    TwoFileNifti = 2,
+    /** 1-file Nifti (consisting of .nii file). */
+    OneFileNifti = 1,
+    /** Legacy Analyze 7.5 format (consisting of .hdr and .img file). */
+    Analyze75 = 0,
+    /** Some other file format, or file system error. */
+    OtherOrError = -1,
+  };
 };
+
+/** Backwards compatibility for enum values */
+#if !defined(ITK_LEGACY_REMOVE)
+using Analyze75Flavor = NiftiImageIOEnums::Analyze75Flavor;
+#endif
 
 /** Define how to print enumerations */
 extern ITKIONIFTI_EXPORT std::ostream &
-                         operator<<(std::ostream & out, const Analyze75Flavor value);
+                         operator<<(std::ostream & out, const NiftiImageIOEnums::Analyze75Flavor value);
+extern ITKIONIFTI_EXPORT std::ostream &
+                         operator<<(std::ostream & out, const NiftiImageIOEnums::NiftiFileEnum value);
 
 /**
  *\class NiftiImageIO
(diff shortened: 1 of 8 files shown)


seanm:
Looks broadly reasonable, if you give me a few days I can test it against my app, but right now I'm trying to get universal binaries working on Mac...

jhlegarreta:
> Looks broadly reasonable, if you give me a few days I can test it against my app, but right now I'm trying to get universal binaries working on Mac...

Sure @seanm. It will push force a patch set to try to fix the Python failure later today, but this is not by any means urgent to me. The critical point with this would be ensuring that the change is backwards-compatible (should be)/does not break anything for third-party applications.

Modules/IO/NIFTI/wrapping/itkNiftiImageIO.wrap
```diff
@@ -5,3 +5,4 @@ itk_wrap_include("itkNiftiImageIOFactory.h")
 itk_wrap_simple_class("itk::NiftiImageIO" POINTER)
 itk_wrap_simple_class("itk::NiftiImageIOFactory" POINTER)
 itk_wrap_simple_class("itk::Analyze75Flavor" ENUM)
+itk_wrap_simple_class("itk::NiftiFileEnum")
```
thewtex, line 8: Since `itk::NiftiImageIO` uses `itk::NiftiFileEnum`, we want the wrapping of `itk::NiftiFileEnum` to be declared before `itk::NiftiImageIO`.

thewtex (changes requested):
@jhlegarreta thanks for the improvement!

An inline comment that will hopefully address the failing tests.

Modules/IO/NIFTI/wrapping/itkNiftiImageIO.wrap
jhlegarreta: The same should apply to `itk::Analyze75Flavor` then, right? If so, I'll do that in a separate PR.

Having missed the `ENUM` specifier is causing the failure I think.

thewtex: > The same should apply to itk::Analyze75Flavor then, right? If so, I'll do that in a separate PR.

Makes sense.

> Having missed the ENUM specifier is causing the failure I think.

The `ENUM` is no longer required following the strongly typed upgrade.

jhlegarreta: > Since itk::NiftiImageIO uses itk::NiftiFileEnum, we want the wrapping of itk::NiftiFileEnum to be declared before itk::NiftiImageIO.

Done in b1d159b.

> The same should apply to itk::Analyze75Flavor then, right? If so, I'll do that in a separate PR.
> Makes sense.

Done in PR #2318.

thewtex:
@jhlegarreta we have to set up these classes for wrapping. There is a `class *Enums` then an inner `enum class * : uint8_t`. There is an example in GDCM:

https://github.com/InsightSoftwareConsortium/ITK/blob/58050aeb0cbe2619465d9bf374366b8e1a8467fb/Modules/IO/GDCM/include/itkGDCMImageIO.h#L49-L67

jhlegarreta:
> @jhlegarreta we have to set up these classes for wrapping. There is a class *Enums then an inner enum class * : uint8_t. There is an example in GDCM:

Done in  9645d86.

@hjmjohnson that required placing `Analyze75Flavor` into the `NiftiImageIOEnums` class. I have not seen any preprocessor directives for its backward compatibility, so not sure if this change will please everybody/how can be done otherwise to have the enums wrapped.

thewtex (approved):
Awesome, thank you @jhlegarreta !

Modules/IO/NIFTI/include/itkNiftiImageIO.h
```diff
@@ -82,28 +107,24 @@ class ITKIONIFTI_EXPORT NiftiImageIO : public ImageIOBase
...
-    TwoFileNifti = 2,
-    /** 1-file Nifti (consisting of .nii file). */
-    OneFileNifti = 1,
-    /** Legacy Analyze 7.5 format (consisting of .hdr and .img file). */
-    Analyze75 = 0,
-    /** Some other file format, or file system error. */
-    OtherOrError = -1,
-  };
-
+#if !defined(ITK_LEGACY_REMOVE)
```
dzenanz, line 110: A similar backwards compatibility block could be added to the namespace scope for `Analyze75Flavor`, see other comment.

Modules/IO/NIFTI/include/itkNiftiImageIO.h
```diff
@@ -28,28 +28,53 @@
...
+    /** 2-file Nifti (consisting of .hdr and .img file). */
+    TwoFileNifti = 2,
+    /** 1-file Nifti (consisting of .nii file). */
+    OneFileNifti = 1,
+    /** Legacy Analyze 7.5 format (consisting of .hdr and .img file). */
+    Analyze75 = 0,
+    /** Some other file format, or file system error. */
+    OtherOrError = -1,
+  };
 };
```
dzenanz, line 71:
```suggestion
};

#if !defined(ITK_LEGACY_REMOVE)
  using Analyze75Flavor = NiftiImageIOEnums::Analyze75Flavor;
#endif
```

Modules/IO/NIFTI/include/itkNiftiImageIO.h
jhlegarreta: Looks reasonable. Thanks for the suggestion @dzenanz ! If I am not mistaken, @seanm @vfonov @hjmjohnson might use this class on a regular basis, so commenting on this aspect would be appreciated as well.

hjmjohnson (approved):
This all looks reasonable and fortifies the codebase.  This is a much cleaner implementation that will eventually allow for better static checking of consistency.

Modules/IO/NIFTI/include/itkNiftiImageIO.h
jhlegarreta: Done in 807dcef.

Modules/IO/NIFTI/include/itkNiftiImageIO.h
jhlegarreta: Added in 807dcef. Thanks for the suggestions @dzenanz.

thewtex (approved):

hjmjohnson (approved):

jhlegarreta:
@seanm Please let us know if this works fine on your builds once you get the time to test it. Thanks.

seanm:
So finally had time to try and indeed it breaks my build, there's a whole bunch of errors like:

error: no member named 'Analyze75Flavor' in namespace 'itk'; did you mean 'itk::NiftiImageIOEnums::Analyze75Flavor'?
error: no type named 'FileType' in 'itk::NiftiImageIO'; did you mean 'SizeType'?
error: no member named 'Analyze75' in 'itk::NiftiImageIO'

etc. etc.

Was this change meant to be backwards compatible?

jhlegarreta:
Thanks for giving this a try @InsightSoftwareConsortium/community 

In principle, the change was designed to be backwards compatible: see [line 74](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/IO/NIFTI/include/itkNiftiImageIO.h#L74) concerning `Analyze75Flavor` and [line 115](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/IO/NIFTI/include/itkNiftiImageIO.h#L115) concerning `NiftiFileEnum`.

Are your builds available on the dashboard so that I can have a closer look at the exact files/lines causing problems?

As far as I can remember, my relevant CMake flag is set to `OFF`. Is this the case for your builds?

seanm:
Well, I wrapped my code in a bunch of `#if (ITK_VERSION_MAJOR > 5) || ((ITK_VERSION_MAJOR == 5) && (ITK_VERSION_MINOR >= 2))` spaghetti and got it compiling again.

It's so very verbose now, with things like:

```
if ((fileType == itk::NiftiImageIOEnums::NiftiFileEnum::OneFileNifti) ||
     (fileType == itk::NiftiImageIOEnums::NiftiFileEnum::TwoFileNifti))
```

4 levels deep for an simple enum. :( Why the new `NiftiImageIOEnums`?

jhlegarreta:
As far as I understand, the strongly typed enums require to be defined within a class to be correctly wrapped in Python (I believe @thewtex and @hjmjohnson could provide further details if necessary), and hence  the `NiftiImageIOEnums` class.

[The ITK v5 migration guide](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/ITK5MigrationGuide.md#strongly-typed-enumerations) does also provide some more details about this.

dzenanz:
@seanm do you have `ITK_LEGACY_REMOVE` ON? That would explain your errors. If you want your application to build with both ITKv4 and ITKv5, you should have legacy removed in v5 and kept in (the default) in v5 build.

seanm:
I do, yes.  It's ok, I've got stuff building again with those version checks.

jhlegarreta:
> I do, yes. It's ok, I've got stuff building again with those version checks.

Thanks for understanding @seanm, and sorry for the hassle.

PR #2322, opened 2021-02-16
STYLE: Use the static `FixedArray::Filled` to initialize array ivars

jhlegarreta:
Improve array ivar initialization in
`itk::VectorGradientMagnitudeImageFilter` constructor using the static
`Fixed::Fill(const ValueType &)` method.

Avoids static casts and unnecessary boilerplate code.

Also, take advantage of the commit to perform such initialization in the
header file.

## PR Checklist

### DIFF
--- a/Modules/Filtering/ImageGradient/include/itkVectorGradientMagnitudeImageFilter.hxx
+++ b/Modules/Filtering/ImageGradient/include/itkVectorGradientMagnitudeImageFilter.hxx
@@ -33,20 +33,9 @@ namespace itk
 template <typename TInputImage, typename TRealType, typename TOutputImage>
 VectorGradientMagnitudeImageFilter<TInputImage, TRealType, TOutputImage>::VectorGradientMagnitudeImageFilter()
 {
-  unsigned int i;
-
   m_UseImageSpacing = true;
   m_UsePrincipleComponents = true;
   m_RequestedNumberOfThreads = this->GetNumberOfWorkUnits();
-  for (i = 0; i < ImageDimension; i++)
-  {
-    m_DerivativeWeights[i] = static_cast<TRealType>(1.0);
-  }
-  for (i = 0; i < VectorDimension; i++)
-  {
-    m_ComponentWeights[i] = static_cast<TRealType>(1.0);
-    m_SqrtComponentWeights[i] = static_cast<TRealType>(1.0);
-  }
   this->DynamicMultiThreadingOn();
   this->ThreaderUpdateProgressOff();
 }
--- a/Modules/Filtering/ImageGradient/include/itkVectorGradientMagnitudeImageFilter.h
+++ b/Modules/Filtering/ImageGradient/include/itkVectorGradientMagnitudeImageFilter.h
@@ -486,13 +486,13 @@ protected:
   }
 
   /** The weights used to scale derivatives during processing */
-  DerivativeWeightsType m_DerivativeWeights;
+  DerivativeWeightsType m_DerivativeWeights = DerivativeWeightsType::Filled(1);
 
   /** These weights are used to scale
       vector component values when they are combined to produce  a scalar.  The
       square root */
-  ComponentWeightsType m_ComponentWeights;
-  ComponentWeightsType m_SqrtComponentWeights;
+  ComponentWeightsType m_ComponentWeights = ComponentWeightsType::Filled(1);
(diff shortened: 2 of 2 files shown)


jhlegarreta:
Related to [this comment](https://github.com/InsightSoftwareConsortium/ITK/pull/2279#discussion_r568455909)/PR #2279.

issakomi (approved):

phcerdan (approved):
:+1:

dzenanz (approved):

dzenanz:
The loop variable is now unused 😄

D:\a\1\s\Modules\Filtering\ImageGradient\include\itkVectorGradientMagnitudeImageFilter.hxx(36): warning C4101: 'i': unreferenced local variable

Modules/Filtering/ImageGradient/include/itkVectorGradientMagnitudeImageFilter.hxx
```diff
@@ -38,15 +38,9 @@ VectorGradientMagnitudeImageFilter<TInputImage, TRealType, TOutputImage>::Vector
...
-  for (i = 0; i < ImageDimension; i++)
-  {
-    m_DerivativeWeights[i] = static_cast<TRealType>(1.0);
-  }
-  for (i = 0; i < VectorDimension; i++)
-  {
-    m_ComponentWeights[i] = static_cast<TRealType>(1.0);
-    m_SqrtComponentWeights[i] = static_cast<TRealType>(1.0);
-  }
+  m_DerivativeWeights.Fill(static_cast<TRealType>(1.0));
```
N-Dekker, line 41: It's certainly an improvement, thanks @jhlegarreta. Can you still check if the brute force static cast is really necessary here? C++ Core Guidelines recommends to [avoid casts](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es48-avoid-casts)

jhlegarreta: Is there a way to do that without checking the compilers complaints in the dashboard after merging?

N-Dekker: @jhlegarreta Please also consider initializing the member variables in the member-initializer-list of the constructor. Using the static member function `itk::FixedArray::Filled(const ValueType &)`.

N-Dekker: > Is there a way to do that without checking the compilers complaints in the dashboard after merging?

@jhlegarreta  Hmmm, I'm afraid you never know when compilers will start complaining, sorry. 🤔  But I also expect code analyzers to complain about unnecessary casting, eventually...

The C++11 way of doing type conversion by curly braces, as in `TRealType{1.0}`, is more type safe than `static_cast<TRealType>(1.0)`.  So I would prefer `TRealType{1.0}` instead of `static_cast`.  Although I still think it should not be necessary in this case.

jhlegarreta: @N-Dekker not sure if I follow. Is [your first proposal](https://github.com/InsightSoftwareConsortium/ITK/pull/2322#discussion_r576910097) preferred over [your second](https://github.com/InsightSoftwareConsortium/ITK/pull/2322#discussion_r576922272) ? Wondering whether we can/it is worthwhile to initialize these these in the header file 🤔 .

N-Dekker: @jhlegarreta Actually the two proposals are independent 😃 I suggested to avoid `static_cast` (if possible), and I suggested to initialize the member variables by the static `FixedArray::Filled(const ValueType &)` member function (instead of the non-static `FixedArray::Fill(const ValueType &)`  that you originally had in mind). Technically you can either accept or reject either one independently. Although of course, I would very much like to see both of them. 😄  This one combines both: https://github.com/InsightSoftwareConsortium/ITK/pull/2322#discussion_r576910097 

C++11 in-class initialization in the header file is also fine by me.

jhlegarreta: :+1: Done in 0a8acb1.

N-Dekker (approved):
Wonderful! Thank you @jhlegarreta

PR #2332, opened 2021-02-23
STYLE: Remove virtual from itkGPUKernelClassMacro, use C++11 `= delete`

N-Dekker:
Deleted all special member functions of GPU kernel helper classes
defined by `itkGPUKernelClassMacro(kernel)`, by C++11 `= delete`.

Note that construction, destruction, copy, and move were previously
already disabled by declaring their special member functions `private`.
So this adjustment is just a style modernization, not an API change.

Made the destructor non-virtual, to avoid the definition of a virtual
table for these classes.

### DIFF
--- a/Modules/Core/Common/include/itkMacro.h
+++ b/Modules/Core/Common/include/itkMacro.h
@@ -1218,14 +1218,11 @@ compilers.
   kernel                                                                                                               \
   {                                                                                                                    \
   public:                                                                                                              \
+    ITK_DISALLOW_COPY_AND_MOVE(kernel);                                                                                \
+    kernel() = delete;                                                                                                 \
+    ~kernel() = delete;                                                                                                \
     static const char * GetOpenCLSource();                                                                             \
-                                                                                                                       \
-  private:                                                                                                             \
-    kernel();                                                                                                          \
-    virtual ~kernel();                                                                                                 \
-    kernel(const kernel &);                                                                                            \
-    void operator=(const kernel &);                                                                                    \
-  };
+  }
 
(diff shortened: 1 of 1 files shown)


Modules/Core/Common/include/itkMacro.h
```diff
@@ -1218,14 +1218,11 @@ compilers.
   kernel                                                                                                               \
   {                                                                                                                    \
   public:                                                                                                              \
+    ITK_DISALLOW_COPY_AND_MOVE(kernel);                                                                                \
+    kernel() = delete;                                                                                                 \
+    ~kernel() = delete;                                                                                                \
```
phcerdan, line 1223: delete on a non-virtual destructor? My alarms just switched on, can you expand on this? Where have you seen this pattern? Did you mean default?

phcerdan (changes requested):

Modules/Core/Common/include/itkMacro.h
phcerdan: if this class cannot be constructed, and not deleted, is there any better way to represent it?

phcerdan (commented):
-

phcerdan (approved):

Modules/Core/Common/include/itkMacro.h
phcerdan: Ok, I see this limits the usage to only the static function.

thewtex (approved):
Nice, thank you @N-Dekker !

dzenanz (approved):

PR #2346, opened 2021-02-28
BUG: memory leak in LSMImageIOTest

issakomi:
The PR closes memory leak in _itkLSMImageIOTest_ , s. [dashboard](https://open.cdash.org/viewDynamicAnalysis.php?buildid=7070588).
Also added _try/catch_ for `lsmImageIO->Write(buffer)`, because the function can throw exception.

### DIFF
--- a/Modules/IO/LSM/test/itkLSMImageIOTest.cxx
+++ b/Modules/IO/LSM/test/itkLSMImageIOTest.cxx
@@ -49,7 +49,6 @@ itkLSMImageIOTest(int argc, char * argv[])
 
   ITK_EXERCISE_BASIC_OBJECT_METHODS(lsmImageIO, LSMImageIO, TIFFImageIO);
 
-
   reader->SetImageIO(lsmImageIO);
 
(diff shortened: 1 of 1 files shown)


jhlegarreta (approved):
Thanks for doing this @issakomi :100:.  I take the blame on this.

For other PRs, usually style changes (remove empty lines that one might not be comfortable with) or improvements (e.g. surrounding the filter updates or IO statements with the `ITK_TRY_EXPECT_NO_EXCEPTION` macro) are preferred if done in separate PRs so that the bug can be isolated in the history and while looking at the cause when inspecting the builds, and also when having a look at the commit history.

Also, adding a commit message is desirable besides the subject (and PR message) to provide a more detailed explanation about the changes. This helps when inspecting he commit history (e.g. part of the PR message fits into that purpose).

Hope these make sense and are useful !

dzenanz (approved):

thewtex (approved):
@issakomi thank you!

PR #2347, opened 2021-03-01
BUG: Initialize member variables

jhlegarreta:
Initialize member variables.

Fixes:
```
UMC ==4407== Conditional jump or move depends on uninitialised value(s)
==4407==    at 0xC3A9316:
double_conversion::DoubleToStringConverter::ToShortestIeeeNumber(double,
double_conversion::StringBuilder*,
double_conversion::DoubleToStringConverter::DtoaMode) const
(double-to-string.cc:172)
==4407==    by 0x6835CE6:
double_conversion::DoubleToStringConverter::ToShortest(double,
double_conversion::StringBuilder*) const (double-to-string.h:166)
==4407==    by 0x68354F6: (anonymous
namespace)::ConvertToShortest(double_conversion::DoubleToStringConverter
const&, double, double_conversion::StringBuilder&)
(itkNumberToString.cxx:31)
==4407==    by 0x68355F8: std::string (anonymous
namespace)::FloatingPointNumberToString(double) (itkNumberToString.cxx:55)
==4407==    by 0x683555B: itk::NumberToString::operator()(double) const
(itkNumberToString.cxx:71)
==4407==    by 0x6834EC5: std::ostream& itk::operator<< (std::ostream&,
itk::Array const&) (itkArrayOutputSpecialization.cxx:37)
==4407==    by 0x250C46: itk::DirectedHausdorffDistanceImageFilter,
itk::Image >::PrintSelf(std::ostream&, itk::Indent) const
(itkDirectedHausdorffDistanceImageFilter.hxx:222)
==4407==    by 0x67FBB8D: itk::LightObject::Print(std::ostream&,
itk::Indent) const (itkLightObject.cxx:125)
==4407==    by 0x24E25F: itkDirectedHausdorffDistanceImageFilterTest(int,
char**) (itkDirectedHausdorffDistanceImageFilterTest.cxx:60)
==4407==    by 0x1D634B: main (ITKDistanceMapTestDriver.cxx:217)
```

and
```
UMC ==5645== Conditional jump or move depends on uninitialised value(s)
==5645==    at 0x927FDB0: std::ostreambuf_iterator > std::num_put >
>::_M_insert_int(std::ostreambuf_iterator >, std::ios_base&, char,
>unsigned long) const (in /usr/lib64/libstdc++.so.6.0.19)
==5645==    by 0x928004C: std::num_put >
>::do_put(std::ostreambuf_iterator >, std::ios_base&, char, unsigned long)
>const (in /usr/lib64/libstdc++.so.6.0.19)
==5645==    by 0x928C2ED: std::ostream& std::ostream::_M_insert(unsigned
long) (in /usr/lib64/libstdc++.so.6.0.19)
==5645==    by 0x24F2B6: std::ostream& itk::operator<< <2u>(std::ostream&,
itk::Size<2u> const&) (itkSize.h:412)
==5645==    by 0x2C6068: itk::ImagePCAShapeModelEstimator, itk::Image
>::PrintSelf(std::ostream&, itk::Indent) const
>(itkImagePCAShapeModelEstimator.hxx:52)
==5645==    by 0x6DBFB8D: itk::LightObject::Print(std::ostream&,
itk::Indent) const (itkLightObject.cxx:125)
==5645==    by 0x2C2F09: itkImagePCAShapeModelEstimatorTest(int, char**)
(itkImagePCAShapeModelEstimatorTest.cxx:141)
==5645==    by 0x20DA6B: main (ITKImageStatisticsTestDriver.cxx:272)
```

and
```
UMC ==6426== Conditional jump or move depends on uninitialised value(s)
==6426==    at 0x10E3942C: __printf_fp_l (in /usr/lib64/libc-2.17.so)
==6426==    by 0x10E38526: vfprintf (in /usr/lib64/libc-2.17.so)
==6426==    by 0x10E63178: vsnprintf (in /usr/lib64/libc-2.17.so)
==6426==    by 0x1095263C: ??? (in /usr/lib64/libstdc++.so.6.0.19)
==6426==    by 0x10958CA3: std::ostreambuf_iterator > std::num_put >
>::_M_insert_float(std::ostreambuf_iterator >, std::ios_base&, char, char,
>double) const (in /usr/lib64/libstdc++.so.6.0.19)
==6426==    by 0x10958F8F: std::num_put >
>::do_put(std::ostreambuf_iterator >, std::ios_base&, char, double) const
>(in /usr/lib64/libstdc++.so.6.0.19)
==6426==    by 0x10964B94: std::ostream& std::ostream::_M_insert(double)
(in /usr/lib64/libstdc++.so.6.0.19)
==6426==    by 0x738573: itk::MultiphaseSparseFiniteDifferenceImageFilter,
itk::Image, itk::Image, itk::ScalarChanAndVeseLevelSetFunction,
itk::Image, itk::ConstrainedRegionBasedLevelSetFunctionSharedData,
itk::Image, itk::ScalarChanAndVeseLevelSetFunctionData, itk::Image > > >,
unsigned int>::PrintSelf(std::ostream&, itk::Indent) const
(itkMultiphaseSparseFiniteDifferenceImageFilter.hxx:1421)
==6426==    by 0xE497B8D: itk::LightObject::Print(std::ostream&,
itk::Indent) const (itkLightObject.cxx:125)
==6426==    by 0x736F6D:
itkScalarChanAndVeseSparseLevelSetImageFilterTest1(int, char**)
(itkScalarChanAndVeseSparseLevelSetImageFilterTest1.cxx:54)
==6426==    by 0x331ADE: main (ITKReviewTestDriver.cxx:327)

UMC ==6426== Conditional jump or move depends on uninitialised value(s)
==6426==    at 0x10957DB0: std::ostreambuf_iterator > std::num_put >
>::_M_insert_int(std::ostreambuf_iterator >, std::ios_base&, char,
>unsigned long) const (in /usr/lib64/libstdc++.so.6.0.19)
==6426==    by 0x1095804C: std::num_put >
>::do_put(std::ostreambuf_iterator >, std::ios_base&, char, unsigned long)
>const (in /usr/lib64/libstdc++.so.6.0.19)
==6426==    by 0x109642ED: std::ostream& std::ostream::_M_insert(unsigned
long) (in /usr/lib64/libstdc++.so.6.0.19)
==6426==    by 0x7385BB: itk::MultiphaseSparseFiniteDifferenceImageFilter,
itk::Image, itk::Image, itk::ScalarChanAndVeseLevelSetFunction,
itk::Image, itk::ConstrainedRegionBasedLevelSetFunctionSharedData,
itk::Image, itk::ScalarChanAndVeseLevelSetFunctionData, itk::Image > > >,
unsigned int>::PrintSelf(std::ostream&, itk::Indent) const
(itkMultiphaseSparseFiniteDifferenceImageFilter.hxx:1422)
==6426==    by 0xE497B8D: itk::LightObject::Print(std::ostream&,
itk::Indent) const (itkLightObject.cxx:125)
==6426==    by 0x736F6D:
itkScalarChanAndVeseSparseLevelSetImageFilterTest1(int, char**)
(itkScalarChanAndVeseSparseLevelSetImageFilterTest1.cxx:54)
==6426==    by 0x331ADE: main (ITKReviewTestDriver.cxx:327
```
signaled at:
https://open.cdash.org/viewDynamicAnalysis.php?buildid=7070588

## PR Checklist

### DIFF
--- a/Modules/Filtering/ImageStatistics/include/itkImagePCAShapeModelEstimator.h
+++ b/Modules/Filtering/ImageStatistics/include/itkImagePCAShapeModelEstimator.h
@@ -202,7 +202,7 @@ private:
 
   VectorOfDoubleType m_EigenVectorNormalizedEnergy;
 
-  ImageSizeType m_InputImageSize;
+  ImageSizeType m_InputImageSize{ {} };
 
   unsigned int m_NumberOfPixels{ 0 };
 
--- a/Modules/Filtering/DistanceMap/include/itkDirectedHausdorffDistanceImageFilter.hxx
+++ b/Modules/Filtering/DistanceMap/include/itkDirectedHausdorffDistanceImageFilter.hxx
@@ -30,7 +30,6 @@ namespace itk
 {
 template <typename TInputImage1, typename TInputImage2>
 DirectedHausdorffDistanceImageFilter<TInputImage1, TInputImage2>::DirectedHausdorffDistanceImageFilter()
-  : m_MaxDistance(1)
 {
   // this filter requires two input images
   this->SetNumberOfRequiredInputs(2);
--- a/Modules/Nonunit/Review/include/itkMultiphaseSparseFiniteDifferenceImageFilter.h
+++ b/Modules/Nonunit/Review/include/itkMultiphaseSparseFiniteDifferenceImageFilter.h
@@ -541,8 +541,8 @@ protected:
 private:
   unsigned int m_CurrentFunctionIndex;
 
-  double       m_RMSSum;
-  unsigned int m_RMSCounter;
+  double       m_RMSSum{ 0. };
+  unsigned int m_RMSCounter{ 0 };
 
   /** This flag is true when methods need to check boundary conditions and
       false when methods do not need to check for boundary conditions. */


Modules/Filtering/DistanceMap/include/itkDirectedHausdorffDistanceImageFilter.h
```diff
@@ -185,7 +185,7 @@ class ITK_TEMPLATE_EXPORT DirectedHausdorffDistanceImageFilter : public ImageToI
   DistanceMapPointer m_DistanceMap;
 
   Array<RealType>       m_MaxDistance;
-  Array<IdentifierType> m_PixelCount;
+  Array<IdentifierType> m_PixelCount = Array<RealType>::Fill(NumericTraits<IdentifierType>::ZeroValue());
```
issakomi, line 188: They (_m_PixelCount_, _m_MaxDistance_, _m_Sum_) are initialized to size of number of threads  in _::BeforeThreadedGenerateData()_ .  Here they have zero size. I guess this can not compile too, AFAIK.

dzenanz, line 188: Maybe `Filled` instead of `Fill`?

jhlegarreta: Thanks for giving this a thought. `Filled` is not a member of `itk::Array`, only one of `itk::FixedArray`.

jhlegarreta: You are right; I had a look at `BeforeThreadedGenerateData()` but here I assumed that the same thinking could not be applied. At the same time, I questioned myself about the dynamic analysis warning if the array has no allocated size at this point/when calling it for a newly created class instance to print its members.  I will need to investigate more.

Leengit, line 188: I don't understand this.  This is in the class declaration, for a non-static member so the initialization here is the default for any constructor that doesn't explicitly give an initialization for this member in the `: member0(value0), member1(value1)` part.  But the old code with the zero-number-of-arguments constructor did so with that constructor; so I don't understand why this is better.  Also, the right hand side is invoking `Fill` without a `this` pointer available, which I don't understand since `Fill` is not a static method.

Modules/Filtering/ImageStatistics/include/itkImagePCAShapeModelEstimator.h
```diff
@@ -202,7 +202,7 @@ class ITK_TEMPLATE_EXPORT ImagePCAShapeModelEstimator : public ImageShapeModelEs
 
   VectorOfDoubleType m_EigenVectorNormalizedEnergy;
 
-  ImageSizeType m_InputImageSize;
+  ImageSizeType m_InputImageSize = NumericTraits<ImageSizeType>::ZeroValue();
```
Leengit, line 205: Or use:
```c++
ImageSizeType m_InputImageSize{ NumericTraits<ImageSizeType>::ZeroValue() };
```

Leengit (changes requested):

Modules/Filtering/DistanceMap/include/itkDirectedHausdorffDistanceImageFilter.h
Leengit: It appears that we are casting a pointer to the `Fill` method to be of type `Array<IdentifierType>::SizeValueType`.

Modules/Filtering/ImageStatistics/include/itkImagePCAShapeModelEstimator.h
jhlegarreta: Done in ddd0e27.

Modules/Filtering/DistanceMap/include/itkDirectedHausdorffDistanceImageFilter.h
jhlegarreta: Right, calling `Fill` was not correct. The old code is producing warnings in the dynamic analysis. As said before, I ignore why since the array has no allocated size. Explicitly initialized to a zero size in ddd0e27.

N-Dekker: What do you think about a simple `{}` initialization?

    Array<IdentifierType> m_PixelCount{};

issakomi: Sorry for noise.  I would recommend simply remove internal temporaries from print functions in all affected classes to get rid of Valgrind defects, that print function is IMHO not for debug, more to view parameters (defaults or set by a user). Alternatively, may be it could be possible use `if (GetSize() > 0)` (not tested, not sure) in print function for arrays that are _purposely_ not initialized in constructor, if e.g. size is assumed to be number of threads, like above arrays. There are tests using print function directly after New(), there is IMHO no useful information from internal temporaries at that point, even if they are initialized somehow.

jhlegarreta: Looks cleanest to me. Done in afee6cc; I confess that given that I do not use ITK on a regular basis I miss the best elegant ways to initialize some of ITK fundamental types.

dzenanz (approved):

Leengit (approved):
LGTM

issakomi:
I am trying with simple own test
```
#include "itkDirectedHausdorffDistanceImageFilter.h"
#include <iostream>

int main(int, char**)
{
  using ImageType = itk::Image<unsigned char, 2>;
  using FilterType = itk::DirectedHausdorffDistanceImageFilter<ImageType, ImageType>;
  typename FilterType::Pointer filter = FilterType::New();
  std::cout << filter << std::endl;
  return 0;
}
```
`
valgrind --leak-check=full ./test`

and Valgrind defects are still here, with `Array<IdentifierType> m_PixelCount{};`
Also with `Array<RealType> m_MaxDistance{};`
Only commenting both _m_MaxDistance_ and _m_PixelCount_ in PrintSelf() function removes Valgrind errors.
So i am afraid Valgrind errors will not go from dashboard.

BTW simple
```
#include <itkArray.h>
#include <iostream>

int main(int, char**)
{
  itk::Array<int> x;
  std::cout << x << std::endl;
  return 0;
}
```
doesn't produce any Valgrind error. Prints `[]` without any errors. I am not sure. I'll look again tomorrow, if errors will be on dashboard, it is interesting.

jhlegarreta:
@issakomi thanks for trying the fixes. Highly appreciated :100:.

I think removing the variables from the `PrintSelf` method is not the direction in which we wish to move: all member variables have to be initialized to known values to avoid cases where the user does not set them and they result in unexpected/undocumented behavior. Testing the print methods through the `ITK_EXERCISE_BASIC_OBJECT_METHODS` and the dynamic analysis are good tools to detect such omissions. Even if this leads us to greater, time consuming development efforts. The effort is worthwhile in the long run. Hope this makes sense :+1:,

If the current fix does not work, then other parts of the code where the same convention is used to initialize an `itk::Array` might suffer from the same issue I'd say. Worrying. And in any case, not sure why that happens, since the array has zero size, and had zero size even before the proposed fix :pensive (personally, when I looked at this I was more confused to see that no defect is detected for `m_MaxDistance`, which has size 1 and its value is not initialized, but that can be left for another battle/PR).

What if instead of the current fix the following is done in the constructor?
```
m_PixelCount.SetSize(this->GetNumberOfWorkUnits());
m_PixelCount.Fill(NumericTraits<IdentifierType>::ZeroValue());
```

issakomi:
```
  m_MaxDistance.SetSize(this->GetNumberOfWorkUnits());
  m_MaxDistance.Fill(NumericTraits<RealType>::ZeroValue());
  m_PixelCount.SetSize(this->GetNumberOfWorkUnits());
  m_PixelCount.Fill(NumericTraits<IdentifierType>::ZeroValue());
```

setting this in constructor (also removed `{}`  and `: m_MaxDistance(1)`) removes all Valgrind errors.

jhlegarreta:
Thanks again @issakomi :100:.

> setting this in constructor (also removed {} and : m_MaxDistance(1)) removes all Valgrind errors.

Nice. Not sure why other member variables of the same type and initialized with the empty braces are not throwing Valgrind defects, though :pensive:.

> Edit: only exactly like above, if you let : m_MaxDistance(1) and initialize only m_PixelCount - errors are still here. Ohh.

Worrying if they refer to `m_PixelCount`.

> Edit:
> Sorry, i don't think PrintSelf() it is a kind of in deep debug, IMHO it is just to see parameters, defaults or set by user, so i am not sure that such intermediate temporaries in PrintSelf() are really a big help, specially filled with zeros. I would prefer to remove internals like these two arrays from PrintSelf() and look closer first. But you decide.

The ITK SW Guides states that all member variables should be printed whether they are publicly exposed or not (section [C.21](https://itk.org/ItkSoftwareGuide.pdf)). I see your point; maybe others have better arguments than me to illustrate the need.

issakomi:
> The ITK SW Guides states that all member variables should be printed whether they are publicly exposed or not (section C.21).

OK, i see. Probably nothing bad can happen if move initialization from BeforeThreadedGenerateData() to constructor, i hope. Still don't fully understand why all the errors happen, with std::vector _m_Sum_ there are no problems, empty or not, only with vnl vector sometimes, i never use vnl in own code, so a little bit lazy to dig.

```
  ThreadIdType numberOfThreads = this->GetNumberOfWorkUnits();

  // Resize the thread temporaries
  m_MaxDistance.SetSize(numberOfThreads);
  m_PixelCount.SetSize(numberOfThreads);
  m_Sum.resize(numberOfThreads);

  // Initialize the temporaries
  m_MaxDistance.Fill(NumericTraits<RealType>::ZeroValue());
  m_PixelCount.Fill(0);
```

issakomi:
Wait... looks like it is possible to avoid Valgring detect only if remove `: m_MaxDistance(1)` here in .hxx
```
template <typename TInputImage1, typename TInputImage2>
DirectedHausdorffDistanceImageFilter<TInputImage1, TInputImage2>::DirectedHausdorffDistanceImageFilter()
  : m_MaxDistance(1)
```

MaxDistance: [6.93676334907487e-310]

That is what causing Valgrind error, at least in my tests, nothing else in .h, with `{}` or without. Only this one change seems to be enough. Current code doesn't remove error in _DirectedHausdorffDistanceImageFilter_

jhlegarreta:
>  looks like it is possible to avoid Valgring detect only if remove : m_MaxDistance(1) here in .hxx
> Edit: IMHO worth to try, size 1 is anyway overridden later in BeforeThreadedGenerateData()

Makes sense now ! Double checked the faulty line in the Valgrind message, and it was pointing to `m_MaxDistance`; so looks like my attempts at identifying the faulty variable were off by one line 😓 .

So, @issakomi :medal: for the relentless effort made on this.

Force pushed a commit removing the single element allocation call from the implementation file; in my mind makes more sense to have these variable not to have any elements when instantiating this class.

Thanks to all for the patience and for bearing with the process.

blowekamp:
Great work tracking this down!

I saw some arrays of number of threads. I glances at the code, and each thread is accessing it's id's index concurrently. This causes[ false sharing](https://en.wikipedia.org/wiki/False_sharing) of memory between threads and can be a significant performance hinderance. Additionally, it restricts the algorithm to use the older fixed threading model. It looks like this filter could use some updating and improvements.

jhlegarreta:
> I saw some arrays of number of threads. I glances at the code, and each thread is accessing it's id's index concurrently. This causes false sharing of memory between threads and can be a significant performance hinderance. Additionally, it restricts the algorithm to use the older fixed threading model. It looks like this filter could use some updating and improvements.

Thanks for casting light to this Brad ! Looks like that would deserve opening an issue to keep track of it?

dzenanz (approved):

thewtex (approved):

thewtex:
Merging for ITK 5.2rc03.

PR #2352, opened 2021-03-02
Reduce duplication in ContourExtractor2DImageFilter

dzenanz:
## PR Checklist

### DIFF
--- a/Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.h
+++ b/Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.h
@@ -24,6 +24,7 @@
 #include <unordered_map>
 #include <deque>
 #include <list>
+#include <atomic>
 
 namespace itk
 {
@@ -192,10 +193,13 @@ protected:
   void
   GenerateData() override;
 
-  /** Subroutine to handle the case that the supplied values are
-   * intensities to be compared to a contour value. */
+  /** Subroutine to create contours for a single label. */
   void
-  GenerateDataForIntensities();
+  CreateSingleContour(const InputImageType * image,
+                      InputRegionType        region,
+                      InputRealType          lowerIsovalue,
+                      InputRealType          upperIsovalue,
+                      SizeValueType          totalNumberOfPixels);
 
   /** Subroutine to handle the case that the supplied values are
    * labels, which are *not* compared to a contour value. */
@@ -215,19 +219,14 @@ private:
                              InputIndexType  fromIndex,
                              InputOffsetType toOffset);
 
-  void
-  AddSegment(const VertexType from, const VertexType to);
 
-  void
-  FillOutputs();
-
-  InputRealType   m_ContourValue;
-  bool            m_ReverseContourOrientation;
-  bool            m_VertexConnectHighPixels;
-  bool            m_LabelContours;
-  bool            m_UseCustomRegion;
-  InputRegionType m_RequestedRegion;
-  unsigned int    m_NumberOfContoursCreated;
+  InputRealType    m_ContourValue;
+  bool             m_ReverseContourOrientation;
+  bool             m_VertexConnectHighPixels;
+  bool             m_LabelContours;
+  bool             m_UseCustomRegion;
+  InputRegionType  m_RequestedRegion;
+  std::atomic_uint m_NumberOfContoursCreated;
 
   // Represent each contour as deque of vertices to facilitate addition of
   // nodes at beginning or end. At the end of the processing, we will copy
@@ -307,12 +306,15 @@ private:
   using VertexMapIterator = typename VertexToContourMap::iterator;
   using VertexContourRefPair = typename VertexToContourMap::value_type;
 
-  // The contours we find in the image are stored here
-  ContourContainer m_Contours;
+  void
+  AddSegment(const VertexType     from,
+             const VertexType     to,
+             ContourContainer &   contours,
+             VertexToContourMap & contourStarts,
+             VertexToContourMap & contourEnds);
 
-  // And indexed by their beginning and ending points here
-  VertexToContourMap m_ContourStarts;
-  VertexToContourMap m_ContourEnds;
+  void
+  FillOutputs(ContourContainer & contours, VertexToContourMap & contourStarts, VertexToContourMap & contourEnds);
 
   // The number of outputs we have allocated capacity for
   unsigned int m_NumberOutputsAllocated;
--- a/Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.h
+++ b/Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.h
@@ -196,7 +196,7 @@ protected:
   /** Subroutine to create contours for a single label. */
   void
   CreateSingleContour(const InputImageType * image,
-                      InputRegionType        region,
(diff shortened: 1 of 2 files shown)


Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.hxx
```diff
@@ -655,81 +525,83 @@ ContourExtractor2DImageFilter<TInputImage>::AddSegment(VertexType from, VertexTy
...
   --m_NumberLabelsRemaining;
-  if (m_NumberOutputsWritten + m_Contours.size() > m_NumberOutputsAllocated)
+  if (m_NumberOutputsWritten + contours.size() > m_NumberOutputsAllocated)
   {
     // We do not have enough capacity; increase capacity to what we need,
     // plus a guess of one contour for each unprocessed label.
-    m_NumberOutputsAllocated = m_NumberOutputsWritten + m_Contours.size() + m_NumberLabelsRemaining;
+    m_NumberOutputsAllocated = m_NumberOutputsWritten + contours.size() + m_NumberLabelsRemaining;
     this->SetNumberOfIndexedOutputs(m_NumberOutputsAllocated);
   }
```
Leengit, line 602: This should work for this single-threaded case, but be careful when we go multi-threaded.

Leengit (changes requested):
Looking good, though I have a few comments.  Thanks for doing this!

Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.h
```diff
@@ -307,12 +306,15 @@ class ITK_TEMPLATE_EXPORT ContourExtractor2DImageFilter
...
   using VertexContourRefPair = typename VertexToContourMap::value_type;
 
-  // The contours we find in the image are stored here
-  ContourContainer m_Contours;
+  void
+  AddSegment(const VertexType     from,
+             const VertexType     to,
+             ContourContainer &   contours,
+             VertexToContourMap & contourStarts,
+             VertexToContourMap & contourEnds);
```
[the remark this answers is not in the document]
Leengit: I am thinking it would be better to put them in a structure _not_ as pointers or references.

Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.hxx
```diff
@@ -655,81 +525,83 @@ ContourExtractor2DImageFilter<TInputImage>::AddSegment(VertexType from, VertexTy
...
     contour.push_front(from);
     contour.push_back(to);
     contour.m_ContourNumber = m_NumberOfContoursCreated++;
     // Add the contour to the end of the list and get a reference to it.
-    m_Contours.push_back(contour);
+    contours.push_back(contour);
 
     // recall that end() is an iterator to one past the back!
-    const auto newContour = --m_Contours.end();
+    const auto newContour = --contours.end();
```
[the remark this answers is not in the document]
dzenanz: It doesn't compiler. I guess an iterator is needed instead of a reference.

Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.h
```diff
@@ -215,19 +219,14 @@ class ITK_TEMPLATE_EXPORT ContourExtractor2DImageFilter
...
-  bool            m_UseCustomRegion;
-  InputRegionType m_RequestedRegion;
-  unsigned int    m_NumberOfContoursCreated;
+  InputRealType    m_ContourValue;
+  bool             m_ReverseContourOrientation;
+  bool             m_VertexConnectHighPixels;
+  bool             m_LabelContours;
+  bool             m_UseCustomRegion;
+  InputRegionType  m_RequestedRegion;
+  std::atomic_uint m_NumberOfContoursCreated;
```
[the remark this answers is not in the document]
dzenanz: If `m_NumberOfContoursCreated` is a local variable, the output is wrong.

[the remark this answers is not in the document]
dzenanz: Maybe not. Something else cause the difference in output.

Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.hxx
```diff
@@ -359,20 +349,27 @@ ContourExtractor2DImageFilter<TInputImage>::GenerateDataForLabels()
...
   }
 
-  for (InputPixelType label : allLabels)
+  m_Interpolate = false;
+
+  for (SizeValueType i = 0; i < allLabels.size(); i++)
   {
+    const InputPixelType label = allLabels[i];
+    const InputRealType  previousLabel = i > 0 ? allLabels[i - 1] : label - 1;
+    const InputRealType  followingLabel = i < allLabels.size() - 1 ? allLabels[i + 1] : label + 1;
```
Leengit, line 358: If `InputPixelType` is `unsigned char` then `label - 1` won't be less than `label` when `label` is 0; likewise, `label + 1` won't be greater than `label` when `label` is 255.  Maybe casting to `InputRealType` earlier would help.  

Even then there is the edge case that `InputPixelType` is `double` and `label` is 1e20; adding or subtracting 1 won't change it because of a lack of precision.  If we permit `label` to be `Inf` or `-Inf` that's trouble too.

I am thinking to not provide `lowerIsovalue` and `upperIsovalue` to `CreateSingleContour`, but instead supply `label`.  Then within `CreateSingleContour` use something like:
```c++
InputPixelType v0 = it.GetPixel(4);
InputPixelType v1 = it.GetPixel(5);
InputPixelType v2 = it.GetPixel(7);
InputPixelType v3 = it.GetPixel(8);
if (m_LabelContours)
{
    v0 = v0 == label;
    v1 = v1 == label;
    v2 = v2 == label;
    v3 = v3 == label;
}
unsigned char  squareCase = 0;
if (v0 > m_contourValue)
{
    squareCase += 1
}
if (v1 > m_contourValue)
{
    squareCase += 2
}
if (v2 > m_contourValue)
{
    squareCase += 4
}
if (v3 > m_contourValue)
{
    squareCase += 8
}
```

Leengit (changes requested):
Thank you.  I added one more comment.

Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.hxx
thewtex: Please address this in a follow-up patch.

thewtex:
Merging for ITK 5.2rc03.

Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.hxx
dzenanz: If `label` is 0 of type `unsigned char`, label-1 will be an expression of type signed short with value of `-1` due to C++ expression rules, I think. This works for the test, which has 0 and 255 as label values.

dzenanz: Updating this logic to cover more cases can be done after we iron out performance issues.

Leengit: I will take a stab at this issue of distinct labels.  @dzenanz

PR #2355, opened 2021-03-03
ENH: Add itk.transformread, itk.transformwrite

thewtex:
Like itk.imread, itk.imwrite, itk.meshread, itk.meshwrite, simple
functions to read and write transform files. Returns and accept a Python
list of Transform's.

Also fix wrapping for TransformBaseTemplate class, TransformFileWriter,
TransformBase.


Wrapping/Generators/Python/Tests/extras.py
```diff
@@ -234,6 +235,23 @@ def custom_callback(name, progress):
 image_series = itk.imread([image_series3d_filename, image_series3d_filename])
 assert image_series.GetImageDimension() == 3
 
+baseline_parameters = np.array([0.6563149 ,   0.58065837,  -0.48175367, -0.74079868, 0.37486398,  -0.55739959,  -0.14306664,   0.72271215, 0.67617978, -66., 69.,  32.])
```
N-Dekker, line 238: Are these (`0.6563149 , 0.58065837, -0.48175367,`...) arbitrary values?

Wrapping/Generators/Python/itk/support/itkExtras.py
```diff
@@ -773,6 +773,58 @@ def meshread(
...
+
+    reader = itk.TransformFileReaderTemplate[itk.D].New()
+    reader.SetFileName(str(filename))
+    reader.Update()
+
+    transforms = []
+    transform_list = reader.GetModifiableTransformList()
+    while not transform_list.empty():
+        transform = transform_list.pop()
+        transforms.append(itk.down_cast(transform))
```
N-Dekker, line 800: My Python knowledge is limited, but I just wonder, would it be possible to retrieve the transforms without modifying (i.e. without popping) the `transform_list`?  Something like the following:

    for transform in transform_list:
        transforms.append(itk.down_cast(transform))

Wrapping/Generators/Python/Tests/extras.py
thewtex: These are what are in the baseline input transform file.

Wrapping/Generators/Python/itk/support/itkExtras.py
thewtex: The transform_list is a C++-wrapped `std::list` and the `transforms` are a Python `list`. This syntax was required to procure the elements with SWIG's wrapping of the `std::list`, which leaves a little to be desired. It would be nice to not modify, but we discard the `transform_list` and `reader` at the end of the function, so it is not necessary.

Wrapping/Generators/Python/Tests/extras.py
N-Dekker: > These are what are in the baseline input transform file.

Thanks Matt, so those values are from Testing/Data/Input/LinearTransform.h5, right?

Wrapping/Generators/Python/itk/support/itkExtras.py
N-Dekker: Basically I would like to be sure that the order is preserved correctly when you do `transform_list.pop` and `transforms.append`?

Wrapping/Generators/Python/Tests/extras.py
thewtex: Correct :+1:

Wrapping/Generators/Python/itk/support/itkExtras.py
thewtex: Good point!

Checking the [docs](https://docs.python.org/3/tutorial/datastructures.html), it is at the end, so we have to reverse the result.

N-Dekker: @thewtex I'm glad I could be of help. But if `transform_list` is just a regular Python list, why not use the simple `for transform in transform_list`, as I suggested? 

    for transform in transform_list:
        transforms.append(itk.down_cast(transform))

?

thewtex: This is a good idea, and I tried it at first, but the SWIG wrapping is not functioning correctly to support it.  `transform_list` is a SWIG wrapped `std::list` that also tries to expose the Python [Abstract Base Clase (ABC)](https://docs.python.org/3/library/collections.abc.html) interface for a MutableSequence. However, when attempting to iterate over it in this way, it does not return the correctly typed `itk.TransformBaseTemplate` object that we can use.

N-Dekker: Thanks, Matt. Do you possibly also have a link to the API of the SWIG wrapped `std::list`?

Wrapping/Generators/Python/Tests/extras.py
N-Dekker: Ah, I see now, it's at https://data.kitware.com/#collection/57b5c9e58d777f126827f5a1/folder/57b76d6b8d777f10f269bc75

Wrapping/Generators/Python/itk/support/itkExtras.py
thewtex: Here's how to produce what is generated, this time for a `double`:

```python
import itk
l = itk.list[itk.D]()
dir(l)
```

The result:

```
['__abstractmethods__', '__bool__', '__class__', '__contains__',
'__delattr__', '__delitem__', '__delslice__', '__dict__', '__dir__',
'__doc__', '__eq__', '__format__', '__ge__', '__getattribute__',
'__getitem__', '__getslice__', '__gt__', '__hash__', '__iadd__', '__init__',
'__init_subclass__', '__iter__', '__le__', '__len__', '__lt__', '__module__',
'__ne__', '__new__', '__nonzero__', '__reduce__', '__reduce_ex__', '__repr__',
'__reversed__', '__setattr__', '__setitem__', '__setslice__', '__sizeof__',
'__slots__', '__str__', '__subclasshook__', '__swig_destroy__', '__weakref__',
'_abc_impl', 'append', 'assign', 'back', 'begin', 'clear', 'count', 'empty',
'end', 'erase', 'extend', 'front', 'get_allocator', 'index', 'insert',
'iterator', 'pop', 'pop_back', 'pop_front', 'push_back', 'push_front',
'rbegin', 'remove', 'rend', 'resize', 'reverse', 'size', 'swap', 'this',
'thisown']

```

N-Dekker: Ah, I see. `pop` seems somewhat special, in this "dir listing", because it isn't part of the C++ `std::list` 🤔 It's a pity that regular Python `for` iteration over the list doesn't work properly. While I don't expect the C++ `begin` to `end` to work well either, for the SWIG wrapping.

N-Dekker:
The current unit test only tests reading and writing a single transform, right? Could you possibly also add a unit test that reads and writes a list of (at least) two different transforms?

Wrapping/Generators/Python/Tests/extras.py
N-Dekker: Just double-checked the TransformParameters from LinearTransform.h5. According to the HDF Viewer: 

```
0.65631490118447
0.5806583745824385
-0.4817536741017158
-0.7407986817430222
0.37486398378429736
-0.5573995934598175
-0.14306664045479867
0.7227121458012518
0.676179776908723
-65.99999999999997
69.00000000000004
32.000000000000036

```

Which are _all close_ to the corresponding values of your `baseline_parameters`  👍

thewtex:
> The current unit test only tests reading and writing a single transform, right? Could you possibly also add a unit test that reads and writes a list of (at least) two different transforms?

@N-Dekker good idea, added :+1:

N-Dekker (approved):

thewtex:
Merging for 5.2rc03.

PR #2356, opened 2021-03-03
imread imwrite imageio

thewtex:
- ENH: imageio kwarg for imread, imwrite
- STYLE: black formatting on itkExtras.py


thewtex:
Merging for ITK 5.2rc03.

PR #2357, opened 2021-03-03
ENH: xarray_from_image .attr, image_from_xarray MetaDataDictionary support

thewtex:
Metadata is mapped to / from the itk.Image MetaDataDictionary and the
xarray.DataArray attrs. Exclude origin, spacing, direction as
appropriate because they are handled specially in itk and xarray.


jhlegarreta (approved):
Have not tested, but I assume the change is fair/necessary.

PR #2360, opened 2021-03-04
STYLE: Remove 6 no-op dynamic_casts (casting T* to T*) from Modules/Core

N-Dekker:
Removed six "no-op" dynamic_casts that just cast pointers to their own compile-time type from Modules/Core.

### DIFF
--- a/Modules/Core/Common/src/itkProcessObject.cxx
+++ b/Modules/Core/Common/src/itkProcessObject.cxx
@@ -91,7 +91,7 @@ ProcessObject::MakeOutput(const DataObjectIdentifierType & name)
     return this->MakeOutput(this->MakeIndexFromOutputName(name));
   }
 
-  return itkDynamicCastInDebugMode<DataObject *>(DataObject::New().GetPointer());
+  return DataObject::New().GetPointer();
 }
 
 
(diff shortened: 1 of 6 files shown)


thewtex (approved):

dzenanz (approved):
How did you identify these instances?

N-Dekker:
> How did you identify these instances?

@dzenanz Thanks for asking! And thanks to you both for approving the pull request! We're hunting dynamic_casts as they do _occasionally_ appear to cause troubles (segfaults) for ITKEstastix on MacOS. I found these particular "no-op" cases by doing a `#define dynamic_cast MyDynamicCast`, with the following definition of MyDynamicCast:

```
#include <type_traits>

template<class TPointer, class U>
TPointer
MyDynamicCast(U* source)
{
  using T = typename std::remove_pointer<TPointer>::type;
  static_assert(!std::is_same<typename std::remove_cv<T>::type, typename std::remove_cv<U>::type>::value,
      "no-op dynamic_cast<T*>!");
  return dynamic_cast<T*>(source);
}

template <class T, class U>
T &
MyDynamicCast(U& source)
{
  static_assert(!std::is_same<typename std::remove_cv<T>::type, typename std::remove_cv<U>::type>::value,
      "no-op dynamic_cast<T&>!");
  return dynamic_cast<T &>(source);
}

```

This redefinition of `dynamic_cast` triggers a compilation error whenever trying to cast to its own compile-time type. Disclaimer: `#define dynamic_cast` is non-Standard, as far as I know.  So it could do anything. 😸  

I must say I'm not against _all_ uses of `dynamic_cast`, but in cases where they can be avoided easily, I would rather have them removed.

PR #2365, opened 2021-03-04
STYLE: Remove 9 no-op dynamic_casts (casting T* to T*)

N-Dekker:
Removed nine "no-op" dynamic_casts that just cast pointers to their own
compile-time type.

Note: in those cases where `auto` was used to declare a variable
initialized by such a no-op `dynamic_cast`, this commit replaced
`auto` by the explicitly specified declaration type.


N-Dekker:
Follow-up to pull request #2360 (which specifically did Modules/Core).

PR #2368, opened 2021-03-04
STYLE: Remove unintended extra space from destructors and operators

N-Dekker:
It appears that clang-format has in some cases introduced an unintended
space character between the class name and the `::~`, in the definition
of a destructor. This typically seems to have happened when there
originally was a line break between the class name and the colon
characters.

Follow-up to:
"STYLE: Remove space between class and member names in C++ source files"
pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2096
commit 734d6f84d728b343baae5ef66643a5568a5a30c7

### DIFF
--- a/Modules/Core/Common/src/itkFileOutputWindow.cxx
+++ b/Modules/Core/Common/src/itkFileOutputWindow.cxx
@@ -30,7 +30,7 @@ FileOutputWindow::FileOutputWindow()
   m_FileName = "";
 }
 
-FileOutputWindow ::~FileOutputWindow()
+FileOutputWindow::~FileOutputWindow()
 {
   delete m_Stream;
   m_Stream = nullptr;
--- a/Modules/Core/Common/src/itkHexahedronCellTopology.cxx
+++ b/Modules/Core/Common/src/itkHexahedronCellTopology.cxx
@@ -33,5 +33,5 @@ const int HexahedronCellTopology ::m_Faces[6][4] = { { 0, 4, 7, 3 }, { 1, 2, 6, 
 
 HexahedronCellTopology::HexahedronCellTopology() = default;
 
-HexahedronCellTopology ::~HexahedronCellTopology() = default;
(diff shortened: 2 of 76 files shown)


dzenanz (approved):
Awesome!

N-Dekker:
> Awesome!

Thanks Dženan! I just looked at those destructors and saw some "space" for improvement  😸

N-Dekker:
@dzenanz I just added similar cosmetic fixes to `operator` definitions, in a separate commit. Please check! Do you like the two commits to be squashed into one?

N-Dekker:
> If this is a clang-format version 8.0 style bug, then we should defer to automation over perfection.

@hjmjohnson If I understand correctly, it isn't really a bug. When the class name and the colons were originally separated by a line-break, clang-format just does not have "the guts" to glue them together. It needs our help  😃 

Once the class name and the colons are attached without a space, clang-format does not take them apart anymore. As far as I see.

hjmjohnson:
@N-Dekker 

> > If this is a clang-format version 8.0 style bug, then we should defer to automation over perfection.
> 
> @hjmjohnson If I understand correctly, it isn't really a bug. When the class name and the colons were originally separated by a line-break, clang-format just does not have "the guts" to glue them together. It needs our help 😃
> 
> Once the class name and the colons are attached without a space, clang-format does not take them apart anymore. As far as I see.

@N-Dekker AWESOME!  Thank you for checking.  Then this is a fabulous change!

N-Dekker:
> @N-Dekker AWESOME! Thank you for checking. Then this is a fabulous change!

Thank you Hans! 

For the record, these Clang-Format issues appear related:
 - [[clang-format] [PR44189] Remove space between identifier and scope resolution operator unless the identifier is a macro](https://reviews.llvm.org/D72216)
 - [Bug 44189 - clang-format does not remove space between non-macro identifier and ::](https://bugs.llvm.org/show_bug.cgi?id=44189)

Both submitted by by Pablo Martin-Gomez.

PR #2374, opened 2021-03-05
Modernize the implementation of ExceptionObject

N-Dekker:
A major revision of one of my earliest contributions to ITK 😸  which aimed to make `itk::ExceptionObject` exception-safe, by introducing a rather complicated new internal data structure: 

Commit 8bfd356e80a8a68cf9211b4089dabaafd8eabd35, _"BUG: fixed issue 0007094: itk::ExceptionObject should never cause a crash. Removed ExceptionObject::UpdateWhat(), as discussed with Brad King and Bill Lorensen at [Insight-developers]"_, 28 May 2008.

This pull request proposes to very much simplify the implementation, while preserving its exception-safety.

### DIFF
--- a/Modules/Core/Common/include/itkExceptionObject.h
+++ b/Modules/Core/Common/include/itkExceptionObject.h
@@ -19,10 +19,9 @@
 #  error "Do not include itkExceptionObject.h directly,  include itkMacro.h instead."
 #else // itkExceptionObject_h
 
+#  include <memory> // For shared_ptr.
 #  include <string>
 #  include <stdexcept>
-
-#  include "itkSmartPointer.h"
 
 namespace itk
 {
@@ -50,10 +49,15 @@ class ITKCommon_EXPORT ExceptionObject : public std::exception
 public:
   static constexpr const char * const default_exception_message = "Generic ExceptionObject";
   using Superclass = std::exception;
-  /** Various types of constructors.  Note that these functions will be
-   * called when children are instantiated.  The default constructor and
-   * the copy constructor of ExceptionObject never throw an exception. */
-  ExceptionObject() noexcept;
+
+  /** Explicitly-defaulted default-constructor. Creates an empty exception object.
+   * \note The other five "special member functions" (copy-constructor,
+   * copy-assignment operator, move-constructor, move-assignment operator,
+   * and destructor) are defaulted implicitly, following the C++ "Rule of Zero".
+   * All of these special member functions are `noexcept`.
+   */
+  ExceptionObject() noexcept = default;
+
   explicit ExceptionObject(const char * file,
                            unsigned int lineNumber = 0,
                            const char * desc = "None",
@@ -62,14 +66,6 @@ public:
                            unsigned int lineNumber = 0,
                            std::string  desc = "None",
                            std::string  loc = "Unknown");
-  ExceptionObject(const ExceptionObject & orig) noexcept;
-
-  /** Virtual destructor needed for subclasses. Has to have empty throw(). */
-  ~ExceptionObject() noexcept override;
-
-  /** Assignment operator. */
-  ExceptionObject &
-  operator=(const ExceptionObject & orig) noexcept;
 
   /** Equivalence operator. */
   virtual bool
@@ -122,48 +118,9 @@ public:
   what() const noexcept override;
 
 private:
-  /** \class ReferenceCounterInterface
-   *
-   *  Exception data.  Location of the error and description of the error.
-   *
-   *  Class hierarchy
-   *
-   *
-   *           ReferenceCounterInterface (Register/UnRegister)
-   *                     ^
-   *                     |
-   *               ExceptionData       LightObject (Register/UnRegister)
-   *                     ^                  ^
-   *                     |                  |
-   *                   ReferenceCountedExceptionData (Register/UnRegister)
-   *
-   *
-   *
-   *  The ReferenceCounterInterface is an abstract class providing
-   *  the API interface expected by the SmartPointer. Its second derived
-   *  class, the ReferenceCountedExceptionData, double inherits from LightObject
-   *  and ExceptionData, and overloads the Register()/UnRegister() methods to
-   *  delegate them to its second parent, the LightObject.
-   *
-   * \ingroup ITKCommon
-   */
-  class ReferenceCounterInterface
-  {
-  public:
-    virtual void
-    Register() const = 0;
-
-    virtual void
-    UnRegister() const = 0;
-
-    ReferenceCounterInterface();
-    virtual ~ReferenceCounterInterface();
-  };
   class ExceptionData;
-  class ReferenceCountedExceptionData;
-  SmartPointer<const ReferenceCounterInterface> m_ExceptionData;
-  const ExceptionData *
-  GetExceptionData() const;
+
+  std::shared_ptr<const ExceptionData> m_ExceptionData;
 };
 
 /** Generic inserter operator for ExceptionObject and its subclasses. */
--- a/Modules/Core/Common/include/itkMetaProgrammingLibrary.h
+++ b/Modules/Core/Common/include/itkMetaProgrammingLibrary.h
@@ -19,6 +19,7 @@
 #define itkMetaProgrammingLibrary_h
 
 #include "itkMacro.h"
+#include "itkSmartPointer.h"
 
 namespace itk
 {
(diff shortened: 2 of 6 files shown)


Modules/Core/Common/include/itkExceptionObject.h
```diff
@@ -50,10 +49,15 @@ class ITKCommon_EXPORT ExceptionObject : public std::exception
...
+
+  /** Explicitly-defaulted default-constructor. Creates an empty exception object.
+   * \note The other five "special member functions" (copy-constructor,
+   * copy-assignment operator, move-constructor, move-assignment operator,
+   * and destructor) are defaulted implicitly, following the C++ "Rule of Zero".
+   * All of these special member functions are `noexcept`.
+   */
+  ExceptionObject() noexcept = default;
+
   explicit ExceptionObject(const char * file,
```
dzenanz, line 61: Since you are touching this, maybe this variant of the constructor is not necessary. `char *` is implicitly convertible into `std::string`, so the other constructor signature should be enough?

dzenanz (approved):
Great simplification!

Modules/Core/Common/include/itkExceptionObject.h
N-Dekker: @dzenanz For me personally, it would be fine to just remove the `const char *` `ExceptionObject` constructor overload, as you're suggesting. However, unlike the other commits of this PR, that would actually change the API!

This `itk::ExceptionObject` constructor explicitly supported "null" `char` pointers as arguments for `file`, `desc`, and `loc`:

https://github.com/InsightSoftwareConsortium/ITK/blob/v5.2rc03/Modules/Core/Common/src/itkExceptionObject.cxx#L133-L137

Would it be OK to you to drop "null" `char` pointer support for `file`, `desc`, and `loc`? And if so, how about `SetLocation` and `SetDescription`? 

    const char * nullCharPtr{ nullptr };

    itk::ExceptionObject exceptionObject(nullCharPtr, __LINE__, nullCharPtr, nullCharPtr);
    exceptionObject.SetLocation(nullCharPtr);
    exceptionObject.SetDescription(nullCharPtr);

Shall we just drop support for any of these three null-pointer "use cases"? You may also first merge this PR, and then reconsider the `const char *` issue afterwards 😸

dzenanz: Let's not change the API, especially since 5.2 final is so close.

Leengit (commented):
Looks good.

jhlegarreta:
@N-Dekker do you think this build warning is related to this PR?
https://open.cdash.org/viewBuildError.php?type=1&buildid=7093971

N-Dekker:
> @N-Dekker do you think this build warning is related to this PR?
> https://open.cdash.org/viewBuildError.php?type=1&buildid=7093971

    itkExceptionObjectGTest.cxx:29:9: warning: 'TestClass' has virtual functions but non-virtual destructor [-Wnon-virtual-dtor]

Thanks @jhlegarreta But I think it's not related to this PR. I think it's related to PR https://github.com/InsightSoftwareConsortium/ITK/pull/2380 ("BUG: Remove duplicate "itk::ERROR: itk::ERROR: " from itkExceptionMacro"), which introduced `TestClass`.

Unfortunately that was also a PR of mine 😸 I'll look at how to fix it! Maybe just add `virtual TestClass() = default;`

SimonRit:
@N-Dekker I have a new error on the RTK dashboard which seems related to this PR, see [here](https://my.cdash.org/viewBuildError.php?buildid=1982089). Is it obvious to you if this is a problem related to your change or an RTK issue?

N-Dekker:
> I have a new error on the RTK dashboard which seems related to this PR, see [here](https://my.cdash.org/viewBuildError.php?buildid=1982089).

Oh, thanks @SimonRit, I see:

> ExceptionObject::~ExceptionObject(void)" dÚjÓ dÚfini(e) dans cuda_compile_1_generated_rtkCudaBackProjectionImageFilter.cu.obj [D:\RTK-2015-Shared-ON\Modules\Remote\RTK\src\RTK.vcxproj]
> CrÚation de la bibliothÞque /.../RTK-2015-Shared-ON/lib/RelWithDebInfo/itkRTK-5.2.lib et de l'objet /.../RTK-2015-Shared-ON/lib/RelWithDebInfo/itkRTK-5.2.exp
> D:\RTK-2015-Shared-ON\bin\RelWithDebInfo\itkRTK-5.2.dll : fatal error LNK1169: un ou plusieurs symboles dÚfinis Ó diffÚrentes reprises ont ÚtÚ rencontrÚs [D:\RTK-2015-Shared-ON\Modules\Remote\RTK\src\RTK.vcxproj]

Looks like it's from Visual C++ 2015, right? I did not see this error before.

Could it be that it was an _incremental_ build? If so, could you possibly do a "clean build" instead?

SimonRit:
Yes, it's Visual c++ 2015. ITK's compilation is incremental indeed ([here](https://github.com/SimonRit/RTK/blob/dashboard/shiitake_build_itk.cmake)), I'll rerun it (then RTK's which is [not incremental](https://github.com/SimonRit/RTK/blob/dashboard/shiitake.cmake)). I'll keep you posted but that's a good lead, thanks.

LucasGandel:
@N-Dekker  @SimonRit 

For information, I reproduce the same error with a clean build (VS2017 compiler). I'll try to investigate a bit and keep you posted if I find anything interesting.

N-Dekker:
> I reproduce the same error with a clean build (VS2017 compiler). I'll try to investigate a bit and keep you posted if I find anything interesting.

Thanks @LucasGandel Just for my information, do you do a static (lib-only) or a shared (DLL) ITK build?

LucasGandel:
@N-Dekker I am building shared libraries.

Also, I am building RTK as an external module (outside ITK), but the dashboard is building it as a remote module (inside ITK) so it should not make any difference.

SimonRit:
Yes, on the dashboard, I can see this issue with shared libs, both with RTK
built in ITK or ouside ITK. You need to activate Cuda if you want to see
this issue.

N-Dekker:
> You need to activate Cuda if you want to see this issue.

But I don't want to see it  😭 Thanks anyway!

Would the only solution be to explicitly implement the `itk::ExceptionObject` destructor in its CXX file? I just don't understand, so many destructors in the world are compiler-generated, according the C++ "rule of zero". Why would this one be more troublesome?

blowekamp:
> Would the only solution be to explicitly implement the `itk::ExceptionObject` destructor in its CXX file? I just don't understand, so many destructors in the world are compiler-generated, according the C++ "rule of zero". Why would this one be more troublesome?

With shared libraries it is important that the destructor is in a compiled file and not implicitly instantiated so that there is only one virtual table for an object.

What does the C++ standards have to say about how symbols from shared libraries are resolved at runtime? System and implementation specific?

N-Dekker:
So far I haven't been able to reproduce the issue. I just did a VS2019 shared lib (DLL) build of ITK, and linked a test executable. Works fine for me! (I did not figure out how to enable Cuda on ITK!)

Obviously we could just add a non-inline destructor to `ExceptionObject`, if that fixes the link error. But then I still don't understand why the "rule of zero" _usually_ works fine, except for now!

SimonRit:
> So far I haven't been able to reproduce the issue.

Same here, I'm not able to provide a minimal example yet.

>  (I did not figure out how to enable Cuda on ITK!)

Set `Module_RTK` and `RTK_USE_CUDA` to ON (see [here](https://github.com/SimonRit/RTK/blob/master/INSTALLATION.md)).

> Obviously we could just add a non-inline destructor to `ExceptionObject`, if that fixes the link error.

I'll try to verify it later.

N-Dekker:
If `ExceptionObject` would really need to have a non-inline destructor, I think we should also explicitly define a move-constructor and move-assignment operator, because these two will then not be defined implicitly anymore. In fact we would then have to switch from "the rule of zero" to "the rule of five". https://en.cppreference.com/w/cpp/language/rule_of_three

But again, I would like to see a minimal example, showing the need for a non-inline destructor!

blowekamp:
Here is a repository with a similar demonstration:
https://github.com/blowekamp/PrivateDynamicCastTest

It should be a good start to reproduce this issue.

The class really needs an explicit virtual table in the compiled object file to be portable across systems and configurations.

blowekamp:
@N-Dekker As I said it is a compiler and system specific how and when the vtable is "anchored". After testing some time ago, assuming that the implementation anchors with destructor has proven to be portable and stable across the board. Thank you for pointing out that some compilers do it differently, and it would be nice if they were all the same. But please continue with what's known to work. These multiple vtable issues are difficult to reproduces as they require multiple shared libraries, and complex systems. Please use what is known to work, and should solve the presented problem.

PR #2382, opened 2021-03-08
PERF: Avoid Superfluous PointsLocator Updates

samuelgerber:
Only updated PointLocators in PointSetToPointSetMetricv4 when necessary.


dzenanz (approved):

thewtex (approved):
@samuelgerber thanks!

Issue #2362, opened 2021-03-04
ThirdParty\GDCM\src\gdcm\Source\Common\gdcmSystem.cxx(815,3): error C3861: 'gettimeofday': identifier not found 


al-sabr:
ITK 4.13 downloaded by CTK

![image](https://user-images.githubusercontent.com/26530975/109975940-7cf50500-7cfb-11eb-8fce-3844f4119d35.png)

2>  G:\programming\cpp\CTK\build\ITK\Modules\ThirdParty\GDCM\src\gdcm\Source\Common\gdcmSystem.cxx(815,3): error C3861: 'gettimeofday': identifier not found [G:\programming\cpp\CTK\build\ITK-build\Modules\ThirdParty\GDCM\src\gdcm\Source\Common\gdcmCommon.vcxproj]

al-sabr:
I added these lines in gdcmSystem.cxx in ITK 4.13

```
#include <chrono>

int gettimeofday(struct timeval* tp, struct timezone* tzp) {
   namespace sc = std::chrono;
   sc::system_clock::duration d = sc::system_clock::now().time_since_epoch();
   sc::seconds s = sc::duration_cast<sc::seconds>(d);
   tp->tv_sec = s.count();
   tp->tv_usec = sc::duration_cast<sc::microseconds>(d - s).count();

   return 0;
}
```

al-sabr:
\`\`\`

Build started...

1>------ Build started: Project: ITK, Configuration: Debug x64 ------

1>Build started 3/4/2021 3:14:30 PM.

1>Target ResolveProjectReferences:

1>Target InitializeBuildStatus:

1>  Touching "x64\\Debug\\ITK\\ITK.tlog\\unsuccessfulbuild".

1>Target CustomBuild:

1>  "The build of 'G:\\programming\\cpp\\CTK\\build\\CMakeFiles\\49db1edb128ae857bc2adce3b068477e\\ITK-patch.rule' depends on 'G:\\PROGRAMMING\\CPP\\CTK\\BUILD\\ITK-CMAKE\\SRC\\ITK-STAMP\\DEBUG\\ITK-UPDATE' which is produced by the build of 'G:\\programming\\cpp\\CTK\\build\\CMakeFiles\\49db1edb128ae857bc2adce3b068477e\\ITK-update.rule'. The items cannot be built in parallel."

[...]

PR #2383, opened 2021-03-08
Resolving compilation error of ITK within CTK.

al-sabr:
This pull request is resolving the following ticket.

https://github.com/InsightSoftwareConsortium/ITK/issues/2362


dzenanz:
`:\a\1\s\Modules\ThirdParty\GDCM\src\gdcm\Source\Common\gdcmSystem.cxx(920): error C2668: 'gdcm::gettimeofday': ambiguous call to overloaded function` in [this build](https://open.cdash.org/viewBuildError.php?buildid=7084102).

jhlegarreta:
Cross-referencing https://github.com/InsightSoftwareConsortium/ITK/issues/2362#issuecomment-822834280
> You can remove this pull request because the problem is not in code but CMAKE variable.

Thanks for the effort.

PR #2371, opened 2021-03-05
Parallelize ContourExtractor2DImageFilter

dzenanz:
## PR Checklist
- [x] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [ ] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [x] Added test (or behavior not changed)
- [ ] Updated API documentation (or API not changed)
- [ ] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [ ] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [ ] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKExamples) for all new major features (if any)

Refer to the [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) for
further development details if necessary.

PR #2384, opened 2021-03-08
PERF: Parallelize ContourExtractor2DImageFilter

Leengit:
This pull request addresses the same issue as Pull Request #2371 of @dzenanz but I am submitting this new pull request because I am cautious about submitting to his pull request, I've rebased to merged master-branch commits in the same code, and I have built on his code to a fair degree.  It is probably best to **_not_** merge both that pull request and this pull request.

Much of this code is co-authored with @dzenanz.

## PR Checklist
- [X] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKExamples) for all new major features (if any)

### DIFF
--- a/Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.hxx
+++ b/Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.hxx
@@ -20,12 +20,14 @@
 
 #include <algorithm>
 
-#include "itkConstantBoundaryCondition.h"
 #include "itkConstShapedNeighborhoodIterator.h"
-#include "itkTotalProgressReporter.h"
+#include "itkConstantBoundaryCondition.h"
 #include "itkContourExtractor2DImageFilter.h"
 #include "itkImageRegionIterator.h"
 #include "itkImageRegionRange.h"
+#include "itkMultiThreaderBase.h"
+#include "itkShapedNeighborhoodIterator.h"
+#include "itkTotalProgressReporter.h"
 
 namespace itk
 {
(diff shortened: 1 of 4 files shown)


Leengit:
Unfortunately, `ShapedImageNeighborhoodRange` with `ConstantBoundaryImageNeighborhoodPixelAccessPolicy` is like `ConstShapedNeighborhoodIterator` with `ConstantBoundaryCondition`. In both cases, the default “outside” pixel value is used when the pixel falls outside the image, but we want it to be used when the pixel falls outside the region.

dzenanz:
Maybe @N-Dekker is willing to extend `ShapedImageNeighborhoodRange` for this use case?

N-Dekker:
@Leengit @dzenanz Do I understand correctly that you want any pixel value outside the requested region to be treated as constant value zero, while iterating over the neighborhood of a pixel at the border of the region? Even when the requested region is in the middle of the image?

Leengit:
> @Leengit @dzenanz Do I understand correctly that you want any pixel value outside the requested region to be treated as constant value zero, while iterating over the neighborhood of a pixel at the border of the region? Even when the requested region is in the middle of the image?

Yes, exactly, except for the zero part.  I need to be able to set what value is returned when a pixel is outside of the region.

Leengit:
FWIW, there may be more than one region in play.  There's the region #1 that we've been discussing that determines when a pixel value should be ignored and instead the set constant is returned.  But, I also have a shaped iterator or range that is traversing a slightly different region #2.  Because of its radius and other parameters the region #2 exploration may end up requesting values for individual pixels outside of region #1, and it is those pixels that I want to get the set constant for.

Also, in some cases the pixels outside of region #1 might also be outside of the image.  I'd like to get the set constant in those cases too, rather than an exception or error.

N-Dekker:
> Yes, exactly, except for the zero part. I need to be able to set what value is returned when a pixel is outside of the region.

Honestly I was not aware of this use case, during the initial development of `ShapedImageNeighborhoodRange`. But I think it is quite well possible to add a new "pixel access policy", which takes the requested region into account.

By default, `ShapedImageNeighborhoodRange` does _not_ pass the requested region to the pixel access policy. Simply because it was never needed. But I think a new pixel access policy could have the region included as "pixel access parameter".

As far as I can see, `ShapedImageNeighborhoodRange` does not need to be adjusted. We just need a new policy! Working title: `RegionalConstantBoundaryImageNeighborhoodPixelAccessPolicy`  😄 

If the value for a pixel outside the region shouldn't always be zero, is the value still known at compile-time? If not, how or when is the value estimated?

Leengit:
I like your analysis ... all we need is a new policy!

> If the value for a pixel outside the region shouldn't always be zero, is the value still known at compile-time? If not, how or when is the value estimated?

Runtime.  This is being used by a superpixel calculating routine and this step starts with a region of labels.  The user can use uchar, int32_t, or float, etc. as labels.  Each time there is a clump of pixels with the same value, I want to draw a contour around the entirity of it.  I set the border-pixel value to be a value not used as a label by the user, so that border region is always recognized as "other".  (In particular, If the the label type is char and the user uses all 256 possible values as labels, the user gets an exception!)

N-Dekker:
@Leengit Still under construction, but please have a look:
https://github.com/N-Dekker/ITK/tree/RegionalConstantBoundaryImageNeighborhoodPixelAccessPolicy

The idea is that `itk::RegionalConstantBoundaryImageNeighborhoodPixelAccessPolicy<ImageType>` can be specified as template argument `TImageNeighborhoodPixelAccessPolicy` of the ShapedImageNeighborhoodRange. A range variable can then be declared as follows:

    itk::ShapedImageNeighborhoodRange<ImageType,
                                      itk::RegionalConstantBoundaryImageNeighborhoodPixelAccessPolicy<ImageType>>
      neighborhoodRange{ *image, location, offsets, { regionIndex, regionSize, constantValue } };


If you're in a hurry, you may just copy https://github.com/N-Dekker/ITK/blob/RegionalConstantBoundaryImageNeighborhoodPixelAccessPolicy/Modules/Core/Common/include/itkRegionalConstantBoundaryImageNeighborhoodPixelAccessPolicy.h and adjust it to your needs.

Leengit:
Thanks @N-Dekker.  I have to work on something else now, so I will probably just wait until your code is merged.

On a related note, if the `ShapedImageNeighborhoodRange` code is similar to the `ConstShapedNeighborhoodIterator` code then Issue #2387 could apply.  Hopefully not, but it might be worth adding a test for.

Leengit:
This is now ready for review.  By using (Shaped)RegionRange objects instead of (Shaped)RegionIterator objects we're using better code, and we bypass Issue #2387.

Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.hxx
```diff
@@ -38,56 +40,53 @@ ContourExtractor2DImageFilter<TInputImage>::ContourExtractor2DImageFilter()
...
 void
-ContourExtractor2DImageFilter<TInputImage>::CreateSingleContour(const InputImageType * image,
-                                                                InputRegionType        shrunkRegion,
-                                                                InputRealType          lowerIsovalue,
-                                                                InputRealType          upperIsovalue,
-                                                                SizeValueType          totalNumberOfPixels)
+ContourExtractor2DImageFilter<TInputImage>::CreateSingleContour(InputPixelType         label,
+                                                                const InputRegionType  shrunkRegion,
+                                                                SizeValueType          totalNumberOfPixels,
+                                                                ContourContainerType & ContoursOutput)
```
dzenanz, line 85: style: variable names should start with a lowercase letter

Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.hxx
```diff
@@ -280,29 +273,40 @@ template <typename TInputImage>
...
+  // itk::ConstantBoundaryImageNeighborhoodPixelAccessPolicy<InputImageType>) can support constant values outside of the
+  // region (rather than merely outside of the image).
+  if (input->GetLargestPossibleRegion() != inputRegion)
+  {
+    inputGC = InputImageType::New();
+    inputGC->SetRegions(inputRegion);
+    inputGC->Allocate();
+    const RegionRange      inputRange{ *inputGC, inputRegion };
+    const RegionConstRange originalRange{ *this->GetInput(), inputRegion };
+    std::copy(originalRange.cbegin(), originalRange.cend(), inputRange.begin());
```
dzenanz, line 292: I thought we would have to make a copy for each label's region. Just one copy is not too bad. Of course, solving this via boundary condition is preferred. But that can be left for a follow-up PR.

dzenanz (approved):

Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.hxx
Leengit: Will do.

Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.hxx
Leengit: Yes, just one copy overall ... because the problem case that the extended region addresses is when the label of interest is also present just outside the region of interest.  Since the label-specific bounding boxes guarantee that the label of interest won't be inside the region but outside the bounding box, we don't need to special-case the bounding box boundary.

Yes, a follow-up PR when we have a boundary condition for regions is what I am thinking too.

dzenanz (approved):
This could be merged as-is. Alternatively, the last two commits could be squashed into the other commits.

Leengit:
> This could be merged as-is. Alternatively, the last two commits could be squashed into the other commits.

I squashed them.

Leengit:
I rebased to the current upstream/master, 38dec0a2ce84ebb506f9061f3e5bbe3c77c07cd6.

PR #2385, opened 2021-03-09
ENH: Add option to access index in point set registration

samuelgerber:
This changes adds the class `PointSetToPointSetMetricWithIndexv4` to the `Metricsv4` module.
This class contains the previous implementation of everything in `PointSetToPointSetMetricv4` but changes the
pure virtual functions such that the index in addition to the point is passed to the functions. 
This avoids using the fixed points locator to find the index in subclasses that need access to it (see for example https://github.com/samuelgerber/ITKThinShellDemons/).

The pure virtual functions `GetLocalNeighborhoodXXX` are changed to pass the `PointIdentiferType` index of the point as well.

The `PointSetToPointSetMetricv4` inherits from `PointSetToPointSetMetricWithIndexv4` and passes the functions above to the original ones, such as to remain unchanged for classes inheriting from `PointSetToPointSetMetricv4`.

Thus future point set to point set metrics have the choice of which method to implement by choosing to inherit from 
`PointSetToPointSetMetricv4` or `PointSetToPointSetMetricWithIndexv4`.

## PR Checklist
- [] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)


samuelgerber:
The python wrapping fails with:
`itkEuclideanDistancePointSetToPointSetMetric: warning(4): ITK type not wrapped, or currently not known: itk::PointSetToPointSetMetricWithIndexv4`

I did not see PointSetToPointSetMetricv4 wrapped and I do not understand why PointSetToPointSetMetricWithIndexv4 would need to be wrapped. Can anybody clarify this for me? 
Thank you

dzenanz (commented):
`PointSetToPointSetMetricv4` inherits from it, so WithIndex variant needs to be wrapped too.

samuelgerber:
> `PointSetToPointSetMetricv4` inherits from it, so WithIndex variant needs to be wrapped too.

Yes, and EuclideanPointSetToPoinSetMetricv4 inherits from PointSetToPointSetMetricv4 and PointSetToPointSetMetricv4 is not wrapped either and did not cause any troubles before?

samuelgerber:
This functionality could also be added by adding those methods to PointSetToPointSetMetricv4.h and have them by default call the versions without index. 
A superclass seems like a cleaner solution though and does not require subclass requiring access to the index to override both options.

Modules/Registration/Metricsv4/wrapping/itkPointSetToPointSetMetricWithIndexv4.wrap
```diff
@@ -0,0 +1,12 @@
+itk_wrap_include("itkPointSet.h")
+itk_wrap_include("itkDefaultStaticMeshTraits.h")
+
+UNIQUE(types "${WRAP_ITK_SCALAR};D")
+
+itk_wrap_class("itk::PointSetToPointSetMetricWithIndexv4" POINTER_WITH_SUPERCLASSES)
```
dzenanz, line 6: `POINTER_WITH_SUPERCLASSES` or `POINTER_WITH_SUPERCLASS`?

dzenanz:
Maybe also rebase on current master, to resolve merge conflicts.

dzenanz (approved):
I assume this is ready for merging?

thewtex:
LGTM

@samuelgerber thanks for pushing this upstream.

@ntustison does this look good to you?

samuelgerber:
> I assume this is ready for merging?

I'd like to have @ntustison , or someone else familiar with the v4 framework, input on this before merging.
This functionality is not absolutely necessary but helpful for subcalsses in two instances for me so far. It could also be added in different ways.

ntustison:
Hey all, thanks for the consideration.  It looks fine to me but it would seem as though y'all are stretching the capabilities more than we have so I would tend to go with your judgement with respect to this pull request.  Thanks for all the work.

PR #2386, opened 2021-03-09
COMP: Fix image pointer casting error

jhlegarreta:
Fix image pointer casting error.

Fixes
```
In file included from
ITK/Modules/Core/FiniteDifference/include/itkDenseFiniteDifferenceImageFilter.h:194:0,
from
ITK/Modules/Filtering/AnisotropicSmoothing/include/itkAnisotropicDiffusionImageFilter.h:21,
from
ITK/Modules/Filtering/AnisotropicSmoothing/include/itkGradientAnisotropicDiffusionImageFilter.h:21,
from
src/Filtering/AnisotropicSmoothing/ComputePeronaMalikAnisotropicDiffusion/Code.cxx:22:
/.../ITKExamples-build/ITK/Modules/Core/FiniteDifference/include/itkDenseFiniteDifferenceImageFilter.hxx:
In instantiation of
'void itk::DenseFiniteDifferenceImageFilter<TInputImage, TOutputImage>::CopyInputToOutput()
[with TInputImage = itk::Image<unsigned char, 2>; TOutputImage = itk::Image<float, 2>]':
src/Filtering/AnisotropicSmoothing/ComputePeronaMalikAnisotropicDiffusion/Code.cxx:72:1:
required from here

ITK/Modules/Core/FiniteDifference/include/itkDenseFiniteDifferenceImageFilter.hxx:44:35:
error: conversion from 'itk::SmartPointer<itk::Image<float, 2> >::ObjectType*
{aka itk::Image<float, 2>*}' to non-scalar type
'itk::Image<unsigned char, 2>::Pointer
{aka itk::SmartPointer<itk::Image<unsigned char, 2> >}' requested

typename TInputImage::Pointer tempPtr =	output.GetPointer();
                              ^~~~~~~
```

raised at:
https://open.cdash.org/viewBuildError.php?buildid=7083938

## PR Checklist

### DIFF
--- a/Modules/Core/FiniteDifference/include/itkDenseFiniteDifferenceImageFilter.hxx
+++ b/Modules/Core/FiniteDifference/include/itkDenseFiniteDifferenceImageFilter.hxx
@@ -41,8 +41,9 @@ DenseFiniteDifferenceImageFilter<TInputImage, TOutputImage>::CopyInputToOutput()
   // Check if we are doing in-place filtering
   if (this->GetInPlace() && this->CanRunInPlace())
   {
-    typename TInputImage::Pointer tempPtr = output.GetPointer();
-    if (tempPtr && tempPtr->GetPixelContainer() == input->GetPixelContainer())
+    const void * const inputPixelContainer = input->GetPixelContainer();
+    const auto * const tempPtr = output.GetPointer();
+    if (tempPtr != nullptr && tempPtr->GetPixelContainer() == inputPixelContainer)
     {
       // the input and output container are the same - no need to copy
       return;


jhlegarreta:
Note that this was triggered by an `ITKExamples` build:
https://open.cdash.org/build/7083938

The same `blaster.kitware` machine builds `ITK` sources nightly; not sure why the error is not being reported in those.

jhlegarreta:
> For the record, the compilation error must have been introduced with pull request #2360 commit 5b3fc71 "STYLE: Remove 6 no-op dynamic_casts (casting T* to T*) from Modules/Core".

:+1: Thanks. That explains why this appeared only recently:
https://open.cdash.org/build/7077668
vs:
https://open.cdash.org/build/7079738

> @jhlegarreta Interesting :thinking: Do I understand correctly that the example tries to convert a itk::Image<float, 2> pointer to a itk::Image<unsigned char, 2> pointer type? That doesn't look right, does it?

Thanks for the heads-up @N-Dekker. Yes, you are right/the error description says so. Having a look at the [code triggering the error](https://github.com/InsightSoftwareConsortium/ITKExamples/blob/master/src/Filtering/AnisotropicSmoothing/ComputePeronaMalikAnisotropicDiffusion/Code.cxx#L22), those are indeed the pixel types defined for the input and output images (files are defined [here](https://github.com/InsightSoftwareConsortium/ITKExamples/blob/master/src/Filtering/AnisotropicSmoothing/ComputePeronaMalikAnisotropicDiffusion/CMakeLists.txt) and the [example can be visually inspected here](https://examples.itk.org/src/filtering/anisotropicsmoothing/computeperonamalikanisotropicdiffusion/documentation)). I might be missing something, but I do not see anything wrong in the cast from the moment that the class is templated over its input and output image types. But please correct me if I'm wrong.

Modules/Core/FiniteDifference/include/itkDenseFiniteDifferenceImageFilter.hxx
```diff
@@ -41,7 +41,7 @@ DenseFiniteDifferenceImageFilter<TInputImage, TOutputImage>::CopyInputToOutput()
   // Check if we are doing in-place filtering
   if (this->GetInPlace() && this->CanRunInPlace())
   {
-    typename TInputImage::Pointer tempPtr = output.GetPointer();
+    typename TInputImage::Pointer tempPtr = dynamic_cast<TInputImage *>(output.GetPointer());
```
N-Dekker, line 44: Please consider `auto tempPtr = output.GetPointer()` instead.

N-Dekker (changes requested):

Modules/Core/FiniteDifference/include/itkDenseFiniteDifferenceImageFilter.hxx
N-Dekker: Something like this:

    const void * const inputPixelContainer = input->GetPixelContainer();
    const auto * const tempPtr = output.GetPointer();

    if ((tempPtr != nullptr) && (tempPtr->GetPixelContainer() == inputPixelContainer))
    {
      // the input and output container are the same - no need to copy
      return;
    }

jhlegarreta: Thanks for the suggestion @N-Dekker : I went on to minimally modify the code in a3cca73. Let me know if you think that  `tempPtr` should absolutely be compared to `nullptr` (if not doing it could derive in unexpected situations in some edge cases?) and the convenience of adding the `inputPixelContainer` local variable.

N-Dekker: @jhlegarreta Does your minimal modification (just changing the declaration of `tempPtr`) compile well?  If so, it's fine by me.

I added the `inputPixelContainer` as a `void` pointer, because I was thinking that `tempPtr->GetPixelContainer() == input->GetPixelContainer()` would not compile, when `tempPtr` and `input` have different pixel types. Just like the comparison `intPtr == floatPtr` fails to compile, for an `int*` and a `float*`: https://godbolt.org/z/sj8snx

The `nullptr` is not _absolutely_ necessary. So again, if your minimal modification solves the problem, it's fine by me.

jhlegarreta: Thanks for the explanations @N-Dekker  !

It compiled for the CI builds (had also tested locally on Windows/MSVC), but just as the existing build error is only reported when ITK is built from the `ITKExamples` `Superbuild`, I prefer to protect myself and avoid further back and forth on this. So done in 57e5f81.

dzenanz (approved):

N-Dekker (approved):
Thank you Jon! 👍

Modules/Core/FiniteDifference/include/itkDenseFiniteDifferenceImageFilter.hxx
```diff
@@ -41,8 +41,9 @@ DenseFiniteDifferenceImageFilter<TInputImage, TOutputImage>::CopyInputToOutput()
   // Check if we are doing in-place filtering
   if (this->GetInPlace() && this->CanRunInPlace())
   {
-    typename TInputImage::Pointer tempPtr = output.GetPointer();
-    if (tempPtr && tempPtr->GetPixelContainer() == input->GetPixelContainer())
+    const void * const inputPixelContainer = input->GetPixelContainer();
+    const auto * const tempPtr = output.GetPointer();
```
Leengit, line 45: How about?:
```c++
const typename TInputImage::PixelContainer * const inputPixelContainer = input->GetPixelContainer();
const TOutputImage * const tempPtr = output.GetPointer();
```

Leengit (changes requested):

Modules/Core/FiniteDifference/include/itkDenseFiniteDifferenceImageFilter.hxx
dzenanz: `inputPixelContainer` would be typed pointer, and therefore the compile might complain about the comparison.

Leengit: Right, got it, the comparison would then be in trouble.  My religion says to use explicit types whenever possible, in this case to catch when return types in the first two lines change unexpectedly, so I might go with the following third line.  But you don't have to join my religion... please do as makes the most sense to you.
```c++
const typename TInputImage::PixelContainer * const inputPixelContainer = input->GetPixelContainer();
const TOutputImage * const tempPtr = output.GetPointer();
if (tempPtr != nullptr && (void *) tempPtr->GetPixelContainer() == (void *) inputPixelContainer)
```

Leengit (approved):

jhlegarreta:
@Leengit thanks for the proposal. If people think that the option you proposed is less error prone, strictly more appropriate and more readable, I would not mind changing it, but I do not have strong arguments for/against.

Modules/Core/FiniteDifference/include/itkDenseFiniteDifferenceImageFilter.hxx
N-Dekker: I'm fine with either `typename TInputImage::PixelContainer` (from @Leengit) or `auto` (as I originally proposed). But please please please no C-style casts. If possible, I strongly prefer no casts at all.

N-Dekker:
While I think the current code is just fine, here is yet another alternative:

    if (std::equal_to<const void *>{}(output->GetPixelContainer(), input->GetPixelContainer()))
    {
      // the input and output container are the same - no need to copy
      return;
    }


You see, `tempPtr` is not really necessary, as it is just `output.GetPointer()`, and `output` is a _smart_ pointer 😃  Moreover, it was checked a few lines before that `output` is not null! And `std::equal_to<const void *>` allows comparing `PixelContainer` pointers of different types.

jhlegarreta:
@N-Dekker I'm fine if what you are proposing is *the* solution, but am not willing to push force that immediately to later find another alternative. So I will wait for a couple of days before adopting that change and push forcing to leave time for discussion if people find downsides to it.

jamesobutler:
I ran into a compile issue with SimpleITK's current master building against ITK's current master seemingly due to the changes in https://github.com/InsightSoftwareConsortium/ITK/commit/5b3fc71e874323253216bb2d146b2a7f7c9029ea as well. The SimpleITK build appears to have succeeded after applying the current changes proposed in this PR.

I'm currently testing to upgrade the ITK version used in 3D Slicer.

issakomi:
> So I will wait for a couple of days before adopting that change and push forcing to leave time for discussion if people find downsides to it.

Sure, thank you for looking at the issue, i am waiting, can not build project with ITK master branch

Screenshot 2021-03-11 at 04 36 45

N-Dekker:
@jhlegarreta 

> So I will wait for a couple of days before adopting that change and push forcing to leave time for discussion if people find downsides to it.

Thanks for your patience  😃 Again, I think the current commit https://github.com/InsightSoftwareConsortium/ITK/pull/2386/commits/57e5f81869b966425c4b1947664a330f5d20180e is just fine. Given the comments from James (@jamesobutler) and Mihail (@issakomi), I would prefer to have it merged today already. Would that be OK to you as well?

We could always discuss possible further style improvements afterwards, right?

Modules/Core/FiniteDifference/include/itkDenseFiniteDifferenceImageFilter.hxx
jhlegarreta: Resolving following Niels' last comment. Thanks @N-Dekker and @Leengit for the discussion, insight and proposals.

jhlegarreta:
@jamesobutler and @issakomi thanks for the report and @jamesobutler for giving the fix a try.

> We could always discuss possible further style improvements afterwards, right?

@N-Dekker Please feel free to submit your proposal as a PR. Thanks for the effort to make the code even more modern and robust.

PR #2389, opened 2021-03-10
ENH: Support 32 bit integer IO in wrapping

thewtex:
Even though 32 bit integer is often not enabled in the wrapping for the
entire toolkit, add minimial support so we can alway read / write / cast
files with this type in their native type.


aylward (approved):
LGTM!

dzenanz (approved):

Issue #2250, opened 2021-01-14
Apple Silicon support


thewtex:
Build and test on the new Apple M1 ARM Silicon processor / system.

### Description

We will have to work with the newer Xcode / processor / binary system.

### Additional Information

This is a tracking issue to discuss issues / progress.

CC: @dzenanz @imikejackson

imikejackson:
Just submitted PR #2251 which detects the arm64 target and allows the compilation to proceed farther.

imikejackson:
Trying to compile ITK on Apple Si (13" MacBook Pro M1) gives the following compile error:

```FAILED: Modules/ThirdParty/PNG/src/itkpng/CMakeFiles/itkpng.dir/pngget.c.o 
[...]

PR #2391, opened 2021-03-10
COMP: removed hardcoded `-march=corei7` compilation flag

seanm:
This fixes building for x86_64 from an arm64 Mac.

In my view, a library like ITK should not be assuming what kinds of CPU the resulting executable will be run on. It can't know if the result will only be run on the building machine, or deployed to customers running all kinds of models of computer. Only the programmer building ITK knows this.

## PR Checklist

### DIFF
--- a/CMake/ITKSetStandardCompilerFlags.cmake
+++ b/CMake/ITKSetStandardCompilerFlags.cmake
@@ -240,10 +240,7 @@ function(check_compiler_optimization_flags c_optimization_flags_var cxx_optimiza
     # Check this list on both C and C++ compilers
     set(InstructionSetOptimizationFlags
        # https://gcc.gnu.org/onlinedocs/gcc-4.8.0/gcc/i386-and-x86_002d64-Options.html
-       # NOTE the corei7 release date was 2008
-       #-mtune=native # Tune the code for the computer used compile ITK, but allow running on generic cpu archetectures
        -mtune=generic # for reproducible results https://github.com/InsightSoftwareConsortium/ITK/issues/1939
-       -march=corei7 # Use ABI settings to support corei7 (circa 2008 ABI feature sets, core-avx circa 2013)
        )
   endif()
   set(c_and_cxx_flags ${InstructionSetOptimizationFlags})


seanm:
@hjmjohnson @imikejackson @dzenanz @thewtex @SimonRit @rmukh this fixes building for Rosetta on an arm64 Mac.  See also discussion in https://github.com/InsightSoftwareConsortium/ITK/issues/2250

Philosophically, I think it's wrong for a library like ITK to hardcode such a flag anyway, but I suspect some may disagree...

dzenanz:
I am fine with this change. Hans?

issakomi:
FYI, _-mtune=generic_ has different meaning for different GCC versions

` Therefore, if you upgrade to a newer version of GCC, code generation controlled by this option will change to reflect the processors that are most common at the time that version of GCC is released.`

From GCC 10.2 [documentation](https://gcc.gnu.org/onlinedocs/gcc-10.2.0/gcc/x86-Options.html#x86-Options)

issakomi:
BTW, i don't have optimization flag _-march=corei7_ in my M1 builds (both _amd64_ and _x86_64_). Probably it was removed by Cmake's [check_c_compiler_flag/check_cxx_compiler_flag](https://cmake.org/cmake/help/latest/module/CheckCCompilerFlag.html) in _ITKSetStandardCompilerFlags.cmake_.  But probably _x86_64_ can be  done on M1 in different ways, not sure is it always removed or not. Both builds work for me without problems. And [Rosetta](https://developer.apple.com/documentation/apple-silicon/about-the-rosetta-translation-environment) shouldn't have problems with corei7 instructions, AFAIK.

CMake/ITKSetStandardCompilerFlags.cmake
```diff
@@ -240,10 +240,7 @@ function(check_compiler_optimization_flags c_optimization_flags_var cxx_optimiza
     # Check this list on both C and C++ compilers
     set(InstructionSetOptimizationFlags
        # https://gcc.gnu.org/onlinedocs/gcc-4.8.0/gcc/i386-and-x86_002d64-Options.html
-       # NOTE the corei7 release date was 2008
-       #-mtune=native # Tune the code for the computer used compile ITK, but allow running on generic cpu archetectures
        -mtune=generic # for reproducible results https://github.com/InsightSoftwareConsortium/ITK/issues/1939
-       -march=corei7 # Use ABI settings to support corei7 (circa 2008 ABI feature sets, core-avx circa 2013)
```
dzenanz, line 246: Could these be exposed as advanced CMake configuration variables? Also, if sensible defaults could be chosen based on the host architecture that would be great.

seanm: Querying the host architecture is part of the problem. It just doesn't work when cross compiling.  If I'm on an arm computer and building for intel, it does not help to examine the host.

dzenanz: Cross-compiling is rare, and I think it is better to bother small number of people doing it with specifying a different option, than 99% of everyone else.

seanm:
>The -mtune=generic will prohibit the compiler from using any modern CPU features added in the past 2 decades.

Are you sure about that?  The gcc docs also say about `-mtune=generic`:

"Produce code optimized for the most common IA32/AMD64/EM64T processors. If you know the CPU on which your code will run, then you should use the corresponding -mtune or -march option instead of -mtune=generic. ***But, if you do not know exactly what CPU users of your application will have, then you should use this option.***" (emphasis mine)

I guess it depends what they mean by "most common IA32/AMD64/EM64T processors" :)

seanm:
@issakomi you should be able to repro the build error on an arm64 Mac if you set `CMAKE_OSX_ARCHITECTURES=x86_64` (only).

CMake/ITKSetStandardCompilerFlags.cmake
seanm: It's not rare on macOS. These days, *every* app is built as a [universal binary](https://en.wikipedia.org/wiki/Universal_binary) (i.e. containing intel & arm executable code). Whether you build your app on intel or arm, the other half of it is essentially cross-compiled.

Introspection of the host causes all kinds of problems. For example, HDF5 was doing a try-compile to check `sizeof(long double)`.  It's 16 on intel, but 8 on arm.  Assuming the answer is the same for both your intel half and arm half is incorrect.

seanm: ... and it's exactly to catch things like that `long double` example that I'm setting up a cdash submission that builds as x86_64-only but on arm64 hardware.  Since the code can still be run (in emulation) it will reveal such problems.

This `march=corei7` thing is the only remaining build issue.

seanm: Perhaps a compromise would be to wrap this in some kind of `if not Apple`?

issakomi:
> you should be able to repro the build error on an arm64 Mac if you set CMAKE_OSX_ARCHITECTURES=x86_64 (only)

I could reproduce with command line:
`mi@mis-Mac-mini tt % /Applications/CMake.app/Contents/bin/cmake -DBUILD_TESTING:BOOL=OFF -DCMAKE_OSX_ARCHITECTURES:STRING=x86_64 ../ITK`

```
      clang: error: the clang compiler does not support '-march=corei7'
      make[1]: *** [CMakeFiles/cmTC_3f278.dir/testCCompiler.c.o] Error 1
      make: *** [cmTC_3f278/fast] Error 2
```


I didn't have the error, i have done both builds separately, for _x86_64_ configured for XCode _only_ with above option, nothing else. _-mtune=corei7_ didn't appear, highly likely because i used _cmake-gui_ and set the option after first run. Then in XCode GUI selected schema ALL_BUILD and "Any Mac (Intel)".

Screenshot 2021-03-11 at 16 45 32

@dzenanz  Apple Silicon have changed a lot recently, now arm64 for Desktops and cross-compile are usual.

CMake/ITKSetStandardCompilerFlags.cmake
thewtex: To avoid issues with cross compiling and other architectures, we could white list the addition of this flag when the target architecture is an x86_64, i.e. do a `list(APPEND` to `InstructionSetOptimizationFlags` when [`CMAKE_SYSTEM_PROCESSOR`](https://cmake.org/cmake/help/v3.20/variable/CMAKE_SYSTEM_PROCESSOR.html) is `AMD64` or `x86_64`.

issakomi:
BTW (work directly from terminal as below)

```
#include <iostream>
int main(int, char**)
{
  const int x = sizeof(double);
  const int y = sizeof(long double);
  std::cout << x << " " << y << std::endl;
  return 0;
}
```

```
mi@mis-Mac-mini ~ % clang++ --target="x86_64-apple-macos10.12" -mtune=corei7 test0.cpp
mi@mis-Mac-mini ~ % ./a.out
8 16
```

```
mi@mis-Mac-mini ~ % clang++ --target="arm64-apple-macos11" test0.cpp
mi@mis-Mac-mini ~ % ./a.out                                         
8 8
```

CMake/ITKSetStandardCompilerFlags.cmake
issakomi: This works, but do **not** set CMAKE_OSX_ARCHITECTURES to x86_64, otherwise configuration fails (?)

`mi@mis-Mac-mini tt % /Applications/CMake.app/Contents/bin/cmake -DBUILD_TESTING:BOOL=OFF  -DCMAKE_C_COMPILER_TARGET:STRING="x86_64-apple-macos10.12" -DCMAKE_CXX_COMPILER_TARGET:STRING="x86_64-apple-macos10.12" ../ITK`

CMAKE_SYSTEM_PROCESSOR can be set or not, i don't see effect

issakomi: It is challenging, _x86_64_ should work with _-march=corei7_, IMHO, even if it should be better used conditionally or removed. There is a problem with Eigen, even if pass most important
```
-DCMAKE_CXX_COMPILER_TARGET:STRING="x86_64-apple-macos11"
-DCMAKE_C_COMPILER_TARGET:STRING="x86_64-apple-macos11"
```
and other params _itkExternal_Eigen3.cmake_  fails, sometimes provided target is used, but at some point
`/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang -x c -target arm64-apple-macos11.1`

thewtex: > itkExternal_Eigen3.cmake

Where is this? We may need to propagage `CMAKE_CXX_COMPILER_TARGET` for Eigen.

issakomi: The error is at line 58, the generation of Eigen3 failed with command above the line. I have tried to pass target, but finally i build own Eigen3. ITK configuration worked:

/Applications/CMake.app/Contents/bin/cmake -DBUILD_TESTING:BOOL=OFF -DCMAKE_BUILD_TYPE:STRING=Release -DCMAKE_CONFIGURATION_TYPES:STRING=Release -DCMAKE_SYSTEM_PROCESSOR:STRING=x86_64 -DCMAKE_OSX_ARCHITECTURES:STRING=x86_64 -DCMAKE_C_COMPILER_TARGET:STRING=x86_64-apple-macos10.14 -DCMAKE_CXX_COMPILER_TARGET:STRING=x86_64-apple-macos10.14 -DITK_USE_SYSTEM_EIGEN:BOOL=ON -DEigen3_DIR:STRING="/Users/mi/eigen/build"  -G Xcode ../ITK

Looks like related things were correct:

```
-- Performing Test CXX_HAS_WARNING-mtune_generic
-- Performing Test CXX_HAS_WARNING-mtune_generic - Success
-- Performing Test CXX_HAS_WARNING-march_corei7
-- Performing Test CXX_HAS_WARNING-march_corei7 - Success

-- Performing Test VXL_HAS_SSE2_HARDWARE_SUPPORT
-- Performing Test VXL_HAS_SSE2_HARDWARE_SUPPORT - Success

-- Looking for 128-bit float. [Checking long double...]
-- Looking for 128-bit float. Found long double.
```

But there are still very many questions, the post will be too long if i start to write about, if someone will play with this, look into cmake logs too, specially cmake's "try compile" is difficult, cmake_system_processer too, tried also to set in tool-chain file, but still don't understand, but seems to be not crucial if target triple is set. Nonetheless i could build one of my applications, but it is a little bit too early to declare x86_64 build on m1 as "supported", IMHO. Good thing is that arm64 native build seems to be OK.

issakomi:
FYI, s. new variable [CMAKE_APPLE_SILICON_PROCESSOR](https://cmake.org/cmake/help/latest/variable/CMAKE_APPLE_SILICON_PROCESSOR.html)

CMake/ITKSetStandardCompilerFlags.cmake
issakomi: @thewtex  This change in _itkExternal_Eigen3.cmake_ fixes Eigen3 failure, if [CMAKE_APPLE_SILICON_PROCESSOR](https://cmake.org/cmake/help/latest/variable/CMAKE_APPLE_SILICON_PROCESSOR.html) is used. More tests are required, but seems that other variables didn't help.

```
  execute_process(
    COMMAND
    ${CMAKE_COMMAND} ${_eigen3_source_dir}
+   "-DCMAKE_APPLE_SILICON_PROCESSOR=${CMAKE_APPLE_SILICON_PROCESSOR}"
    "-DCMAKE_INSTALL_PREFIX=${_eigen3_cmake_install_prefix}"
    "-DCMAKE_INSTALL_INCLUDEDIR=${_eigen3_cmake_install_includedir}"
    "-DCMAKE_INSTALL_DATADIR=${_eigen3_cmake_install_datadir}"
    "-DCMAKE_GENERATOR=${CMAKE_GENERATOR}"
    "-DCMAKE_SH=${CMAKE_SH}"
    "-DCMAKE_GENERATOR_TOOLSET=${CMAKE_GENERATOR_TOOLSET}"
    "-DCMAKE_SH=${CMAKE_SH}"
    "-DCMAKE_GENERATOR_PLATFORM=${CMAKE_GENERATOR_PLATFORM}"
    "-DCMAKE_GENERATOR_INSTANCE=${CMAKE_GENERATOR_INSTANCE}"
    "-DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}"
    "-DCMAKE_C_FLAGS=${CMAKE_C_FLAGS}"
    "-DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}"
    "-DCMAKE_CXX_FLAGS=${CMAKE_CXX_FLAGS}"
    ${_additional_external_project_args}
    WORKING_DIRECTORY ${_eigen3_build_dir}
    OUTPUT_VARIABLE ITKEigen3Config_STDOUT
    ERROR_VARIABLE ITKEigen3Config_STDERR
    )
```

dzenanz: Good analysis. Could you turn this into a patch @issakomi?

thewtex:
@seanm @issakomi @hjmjohnson please take a look at https://github.com/InsightSoftwareConsortium/ITK/pull/2459. This white list flag addition when we are compiling for x86_64. This way, we can keep the optimizations when possible. Testing on an M1 system, we no longer run into the configure error with 

> CMAKE_OSX_ARCHITECTURES=x86_64 (only)

seanm:
And with `CMAKE_OSX_ARCHITECTURES=x86_64;arm64` ?

issakomi:
> And with CMAKE_OSX_ARCHITECTURES=x86_64;arm64 ?

If Eigen3 configuration fails - may be skip passing compiler flags to external CMake generator? There is a comment, BTW
```
  # Nothing to link or compile, so no need to pass compiler flags.
  # However, generators and c,cxx compilers have to be explictily passed
  # for CMake configuration to work.
```
However they are passed, may be remove these [lines](https://github.com/InsightSoftwareConsortium/ITK/blob/4f64c6c962272571e5979a284cfcc40e09a2d53e/CMake/itkExternal_Eigen3.cmake#L51)

There is also typo _explictily_  (explicitly) in comment above.

thewtex:
> And with CMAKE_OSX_ARCHITECTURES=x86_64;arm64 ?

Configuration also succeeds.

PR #2393, opened 2021-03-11
COMP: Fix missing initializer warning

jhlegarreta:
Fix missing initializer warning.

Fixes:
```
/tmp/bld/ITK/Modules/Core/Common/include/itkImage.hxx: In instantiation of
'unsigned int itk::Image<TPixel,
VImageDimension>::GetNumberOfComponentsPerPixel() const [with TPixel =
itk::Index<4u>; unsigned int VImageDimension = 4u]':
bld/ITK/Modules/Core/Common/test/itkConstNeighborhoodIteratorTest.cxx:427:1:
required from here
[CTest: warning matched]
bld/ITK/Modules/Core/Common/include/itkImage.hxx:151:21: warning: missing
initializer for member 'itk::Index<4u>::m_InternalArray'
[-Wmissing-field-initializers]
   const PixelType p{};
                     ^
```

raised at:
https://open.cdash.org/viewBuildError.php?type=1&buildid=7088974
https://open.cdash.org/viewBuildError.php?type=1&buildid=7089276
https://open.cdash.org/viewBuildError.php?type=1&buildid=7089222

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkImage.hxx
+++ b/Modules/Core/Common/include/itkImage.hxx
@@ -148,7 +148,7 @@ Image<TPixel, VImageDimension>::GetNumberOfComponentsPerPixel() const
 {
   // use the GetLength() method which works with variable length arrays,
   // to make it work with as much pixel types as possible
-  const PixelType p{};
+  const auto p = PixelType();
   return NumericTraits<PixelType>::GetLength(p);
 }
 


dzenanz (approved):

N-Dekker:
Thanks for reporting this compiler warning, Jon. This warning appears specific for old versions of GCC (up to GCC 4.9.4). More recent versions don't produce the warning anymore. Would it be possible to just disable the warning for GCC 4?

 The original code appears entirely correct:

    const PixelType p{};

However, if it's absolutely necessary to address the old GCC warning, it could probably be rewritten to:

     const auto p = PixelType();

Leengit:
Would either of the following make the compiler happy?:
```c++
const PixelType p;
const PixelType p();
```

Modules/Core/Common/include/itkImage.hxx
```diff
@@ -148,7 +148,9 @@ Image<TPixel, VImageDimension>::GetNumberOfComponentsPerPixel() const
 {
   // use the GetLength() method which works with variable length arrays,
   // to make it work with as much pixel types as possible
-  const PixelType p{};
+  IndexType start;
+  start.Fill(0);
+  const PixelType p = this->GetPixel(start);
```
Leengit, line 153: Is the origin guaranteed to be a valid index into the image?  If not, `this->GetPixel(start)` may fail.

N-Dekker:
> Would either of the following make the compiler happy?:
> 
> ```c++
> const PixelType p;
> const PixelType p();
> ```

@Leengit   Note that the second line declares a function `p()`! Anyway, if you like, you may try them at https://godbolt.org/z/Kz9Tf8  😃

Leengit:
> > Would either of the following make the compiler happy?:
> > ```c++
> > const PixelType p;
> > const PixelType p();
> > ```
> 
> @Leengit Note that the second line declares a function `p()`! Anyway, if you like, you may try them at https://godbolt.org/z/Kz9Tf8 smiley

Thank you for the link... and the first one fails too.

Modules/Core/Common/include/itkImage.hxx
jhlegarreta: If I understand correctly, indices are in image space, hence I'd dare to say that this would work. I have some doubts where it would fail (i.e. the call itself or when trying to get the pixel) should the image be a null pointer. Might be a good enough reason to think of alternatives.

Leengit: Generally, the `region` argument to `Image::SetRegions()` need not encompass the origin and if that is what is in effect before the call to `Image::Allocate()` then the origin would be a problem.  I don't know enough context for these edits to know whether this problem applies in this case.

jhlegarreta:
Thanks for reviewing @N-Dekker !

> Thanks for reporting this compiler warning, Jon. This warning appears specific for old versions of GCC (up to GCC 4.9.4). More recent versions don't produce the warning anymore, on this line of code. Would it be possible to just disable the warning for GCC 4?

I do not have arguments to disable the warning. Maintainers of those sites (@blowekamp and folks at Kitware @dzenanz, @thewtex  @Leengit if I'm not mistaken) might have an opinion on this. If I'm not mistaken one of `AWS BATCH`/`CentOS Coverage`/`Valgrind` sites/builds was put up recently, so I ignore if that has an old version of GCC.

> The original code appears entirely correct:
const PixelType p{};

The compiler says that `p` is not initialized. I ignore how that is dealt with across different GCC compilers.

> However, if it's absolutely necessary to address the old GCC warning, it could probably be rewritten to:
 const auto p = PixelType();

If this is guaranteed to initialize the variable to a known value, I'd fine with the solution. I've tested on my Windows/MSVC machine and p is initialized to all 0's :tada: . But I do not have the means to try on a GCC compiler to ensure that such would be the case. Unless people prefer to disable the warnings, I'll push force what @N-Dekker proposed tonight.

Modules/Core/Common/include/itkImage.hxx
jhlegarreta: Thanks for the enlightening explanation @Leengit ! So enough reasons to go for another solution, probably what @N-Dekker proposed [here](https://github.com/InsightSoftwareConsortium/ITK/pull/2393#issuecomment-796582214).

N-Dekker:
Thanks, Jon! `const auto p = PixelType()` should work well, as the `Type()` syntax is standardized by C++03 (2003) to yield a "value-initialized" object. "Value-initialized" basically means either default-constructed or zero-initialized, depending on the type. With C++11, `Type{}` (with curly braces) also yields a value-initialized object, but unfortunately it appears to trigger those old GCC4 warnings.

jhlegarreta:
As always, thanks for the insight @N-Dekker . Force pushed. Let's see if the builds are happy.

N-Dekker (approved):
> As always, thanks for the insight @N-Dekker . Force pushed. Let's see if the builds are happy.

You're welcome @jhlegarreta I'm pretty sure that your pull request will even make the old GCC4 happy  😄  Approved!

PR #2396, opened 2021-03-11
STYLE: Use equal_to on pixel containers DenseFiniteDifferenceImageFilter

N-Dekker:
Follow-up to pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2386
commit 1c25350c93c33578b11c83bce501496c306ff3b1
"COMP: Fix image pointer casting error"

Also removed an unused `#include <list>`.

### DIFF
--- a/Modules/Core/FiniteDifference/include/itkDenseFiniteDifferenceImageFilter.hxx
+++ b/Modules/Core/FiniteDifference/include/itkDenseFiniteDifferenceImageFilter.hxx
@@ -19,10 +19,12 @@
 #define itkDenseFiniteDifferenceImageFilter_hxx
 #include "itkDenseFiniteDifferenceImageFilter.h"
 
-#include <list>
 #include "itkImageRegionIterator.h"
 #include "itkNumericTraits.h"
 #include "itkNeighborhoodAlgorithm.h"
+
+#include <functional> // For equal_to.
+
 
 namespace itk
 {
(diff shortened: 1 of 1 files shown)


N-Dekker:
This is just a very small style improvement. I believe a much greater improvement is possibly for the second part of this member function, `DenseFiniteDifferenceImageFilter::CopyInputToOutput()` (after the `return`).

Modules/Core/FiniteDifference/include/itkDenseFiniteDifferenceImageFilter.hxx
```diff
@@ -39,15 +41,11 @@ DenseFiniteDifferenceImageFilter<TInputImage, TOutputImage>::CopyInputToOutput()
...
-    if (tempPtr != nullptr && tempPtr->GetPixelContainer() == inputPixelContainer)
-    {
-      // the input and output container are the same - no need to copy
-      return;
-    }
+    // the input and output container are the same - no need to copy
+    return;
   }
 
   ImageRegionConstIterator<TInputImage> in(input, output->GetRequestedRegion());
```
N-Dekker, line 51: Looks like the nine lines of code that start here could be replaced by a single function call:

    ImageAlgorithm::Copy<TInputImage, TOutputImage>(
      input, output, input->GetRequestedRegion(), output->GetRequestedRegion());

`ImageAlgorithm::Copy` is very much optimized for speed. In fact, I observed that it is even significantly faster, than my personal favorite ITK5 range based iteration, like `std::copy(inputRange.cbegin(), inputRange.cend(), std::begin(outputRange))`.

dzenanz: If `ImageAlgorithm::Copy` supports image adaptors, it would be equivalent. I am not sure how important that is in this case.

dzenanz (approved):

Modules/Core/FiniteDifference/include/itkDenseFiniteDifferenceImageFilter.hxx
N-Dekker: @dzenanz 
> If ImageAlgorithm::Copy supports image adaptors, it would be equivalent.

I think `ImageAlgorithm::Copy` does support image adaptors. But maybe Bradley can confirm... @blowekamp ?

> I am not sure how important that is in this case.

I do believe using  `ImageAlgorithm::Copy` will make `DenseFiniteDifferenceImageFilter::CopyInputToOutput()` really much much faster.  But I don't know how many users actually call `DenseFiniteDifferenceImageFilter::CopyInputToOutput()` on a regular basis. If it isn't being used so much anymore, we might as well leave it alone. What do you think?

N-Dekker: By the way, is the original line of code OK, saying `in(input, output->GetRequestedRegion())`?  I don't know, but it looks like it should be `in(input, input->GetRequestedRegion())`: `input` instead of `output` 🤔 !

dzenanz: We usually limit processing to the requested region, which is set on the output. So it looks right.

thewtex (approved):

PR #2398, opened 2021-03-11
STYLE: C++11 inheriting constructors and Rule of Zero for ExceptionObject derived classes

N-Dekker:
`MemoryAllocationError`, `RangeError`, `InvalidArgumentError`, and
`IncompatibleOperandsError` may simply inherit the constructors from
their base class, `itk::ExceptionObject`.


dzenanz (approved):
Great!

N-Dekker:
FYI Added an extra commit, which removes the explicit destructor declarations and definitions from those classes.

dzenanz (approved):

Issue #2406, opened 2021-03-15
Wrapping Build Error


melrobin:
### Description
I receive a build error related to Python wrapping.

### Actual behavior
I receive several errors like the following before failing compilation.

[ 91%] Generating itkFastSymmetricForcesDemonsRegistrationFilterPython.cpp, ../../Generators/Python/itk/itkFastSymmetricForcesDemonsRegistrationFilterPython.py
/home/melrobin/packages/ITK/build/Wrapping/Typedefs/itkFastSymmetricForcesDemonsRegistrationFilter.i:172: Warning 401: Nothing known about base class 'itk::PDEDeformableRegistrationFilter< itk::Image< double,2 >,itk::Image< double,2 >,itk::Image< itk::Vector< double,3 >,2 > >'. Ignored.
/home/melrobin/packages/ITK/build/Wrapping/Typedefs/itkFastSymmetricForcesDemonsRegistrationFilter.i:172: Warning 401: Maybe you forgot to instantiate 'itk::PDEDeformableRegistrationFilter< itk::Image< double,2 >,itk::Image< double,2 >,itk::Image< itk::Vector< double,3 >,2 > >' using %template.

```none
```

### Versions
Development sources

### Environment
Fedora 33, gcc 10.2.1

dzenanz:
Confirmed. This happens when Python wrapping and `ITKReview` module are turned on.
```text
dzenan@corista:~/ITK-git-rel$ make ITKReviewPython
...
[ 96%] Built target PCRE
[ 96%] Built target swig
[ 96%] Generating itkFastSymmetricForcesDemonsRegistrationFilterPython.cpp, ../../Generators/Python/itk/itkFastSymmetricForcesDemonsRegistrationFilterPython.py
/home/dzenan/ITK-git-rel/Wrapping/Typedefs/itkFastSymmetricForcesDemonsRegistrationFilter.i:172: Warning 401: Nothing known about base class 'itk::PDEDeformableRegistrationFilter< itk::Image< float,2 >,itk::Image< float,2 >,itk::Image< itk::Vector< float,3 >,2 > >'. Ignored.
[...]

PR #2410, opened 2021-03-16
COMP: Wrap FastSymmetricForcesDemonsRegistrationFilter like its parent class

thewtex:
Addresses #2406


melrobin (approved):
Build completed with these changes.

PR #2412, opened 2021-03-16
WIP: BUG: Replace dynamic_cast by Convert, doing implicit type conversion

N-Dekker:
If possible, it is preferable not to use of `dynamic_cast` because of
performance costs, and sometimes segmentation faults (as observed on
MacOS).

`Conversion::Convert<T*>(ptr)` safely converts a pointer to `T*`. It
returns null if `ptr` is not implicitly convertible to `T*`.

Calls to this new conversion function replace `dynamic_cast` in 21 places.


dzenanz (commented):
Have you counted how many dynamic casts remain after this update? Did you examine each case and decided which one to change to this new conversion style?

N-Dekker:
> Have you counted how many dynamic casts remain after this update?

No, thanks for asking @dzenanz, did you already count?

> Did you examine each case and decided which one to change to this new conversion style?

These specific cases addressed by this PR appear to _either_ have done a "no-op" dynamic_cast (just casting to its own compile-time type), _or_ simply yield null, depending on the template instantiation.

Anyway, I have to double-check the PR myself. First I wanted to be sure that it would pass the CI tests 😃

dzenanz (approved):

N-Dekker:
I believe this pull request could improve the performance of various filters but I'm reluctant to continue, as I'm not really familiar with these filters.

hjmjohnson:
@N-Dekker   What should be done with this PR?

N-Dekker:
@hjmjohnson Thansk for asking! I have to admit I haven't yet decided whether to continue on this! It's kind of nice if these particular dynamic_casts can be avoided, and replaced by a simple `Convert` function that just does implicit conversion. But I'm not really sure if it's worth the extra maintenance. Theoretically it could yield a performance improvement, but I just don't know yet if it's noticeable.

Modules/Core/QuadEdgeMesh/include/itkQuadEdgeMeshFrontIterator.hxx
```diff
@@ -132,12 +133,12 @@ template <typename TMesh, typename TQE>
...
+  if (auto * edge = Experimental::Conversion::Convert<QEType *>(m_Mesh->GetEdge()))
   {
     return edge;
   }
   using QEDual = typename QEType::DualType;
-  if (auto * edge = dynamic_cast<QEDual *>(m_Mesh->GetEdge()))
+  if (auto * edge = Experimental::Conversion::Convert<QEDual *>(m_Mesh->GetEdge()))
   {
     return edge->GetRot();
   }
```
N-Dekker, line 144: Now that we use C++17, I think this can be implemented as follows:

```cpp
  auto * edge = m_Mesh->GetEdge(); 

  if constexpr (std::is_convertible_v<decltype(edge), QEType *>)
  {
    return edge;
  }
  using QEDual = typename QEType::DualType;
  if constexpr (std::is_convertible_v<decltype(edge), QEDual *>)
  {
    return edge->GetRot();
  }
```

This would remove the old `dynamic_cast` _and_ make the `if` statements entirely compile-time.

N-Dekker:
@hjmjohnson 

> Can we assume it is no longer a priority and close it?

Indeed! https://github.com/InsightSoftwareConsortium/ITK/pull/2412#discussion_r2414453832 suggests a better approach. So I'll just close this pull request.

PR #2417, opened 2021-03-17
COMP: Define ExceptionObject destructor out-of-line (doing Rule of Five)

N-Dekker:
Followed the C++ Rule of Five, instead of the Rule of Zero, for
`ExceptionObject`, and added the definition of its destructor to its CXX
file.

Aims to fix Visual C++ link errors (error LNK2005 "symbol was defined
more than once", fatal error LNK1169) reported by @SimonRit, and
confirmed by @LucasGandel, at "Modernize the implementation of
ExceptionObject", pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2374

### DIFF
--- a/Modules/Core/Common/include/itkExceptionObject.h
+++ b/Modules/Core/Common/include/itkExceptionObject.h
@@ -50,12 +50,7 @@ public:
   static constexpr const char * const default_exception_message = "Generic ExceptionObject";
   using Superclass = std::exception;
 
-  /** Explicitly-defaulted default-constructor. Creates an empty exception object.
-   * \note The other five "special member functions" (copy-constructor,
-   * copy-assignment operator, move-constructor, move-assignment operator,
-   * and destructor) are defaulted implicitly, following the C++ "Rule of Zero".
-   * All of these special member functions are `noexcept`.
-   */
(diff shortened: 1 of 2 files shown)


dzenanz (approved):

dzenanz:
@SimonRit @LucasGandel can you check whether this fixes the issue for you?

N-Dekker:
In general, I very much prefer "Rule of Zero", rather than "Rule of Five", so please only merge this pull request if it is really necessary, in order to fix those link errors 🤔 !

SimonRit:
Thanks a lot @N-Dekker. I'll run the tests and let you know, probably tomorrow. Like you, I wish I could understand what's going on but it may well be an issue with nvcc, the Cuda compiler. I haven't had time to prepare an example doing this with a file compiled by nvcc.

SimonRit (approved):
It does fix the problem, see [here](https://my.cdash.org/build/1983415). Thanks again!
Like you, I wish I could understand the issue better because I believe the problem is similar to other problems we have (discussed on discourse [here](https://discourse.itk.org/t/template-class-already-defined-when-using-itk-testdriver/3807/2)). In the meantime, I approve this change, thanks for doing it (and teaching me rules of digits).

thewtex (approved):

PR #2419, opened 2021-03-17
COMP: fix CMake Deprecation Warning (Compatibility with CMake < 2.8.12)

dzenanz:
This is just a convenience commit to get rid of the warnings for 5.2 final.
These libraries should be updated in a normal way once 5.3 development starts.

Text of the error messages:

CMake Deprecation Warning at Modules/ThirdParty/GoogleTest/src/itkgoogletest/CMakeLists.txt:4 (cmake_minimum_required):
  CMake.

CMake Deprecation Warning at Modules/ThirdParty/GoogleTest/src/itkgoogletest/googletest/CMakeLists.txt:56 (cmake_minimum_required):
  CMake.

CMake Deprecation Warning at Modules/ThirdParty/MINC/src/libminc/CMakeLists.txt:25 (CMAKE_MINIMUM_REQUIRED):
  CMake.


thewtex (approved):
LGTM!

PR #2420, opened 2021-03-18
COMP: Avoid creating a copy variable

jhlegarreta:
Avoid creating a copy variable.

Fixes:

```
[CTest: warning matched]
/Users/builder/externalModules/Core/Common/test/itkShapedImageNeighborhoodRangeGTest.cxx:547:21:
warning: loop variable 'pixel' of type 'const
itk::ShapedImageNeighborhoodRange<const itk::Image<unsigned char,
2>, itk::ZeroFluxNeumannImageNeighborhoodPixelAccessPolicy<const
itk::Image<unsigned char, 2> > >::PixelProxy<true, void>' creates
a copy from type 'const itk::ShapedImageNeighborhoodRange<const
itk::Image<unsigned char, 2>,
itk::ZeroFluxNeumannImageNeighborhoodPixelAccessPolicy<const
itk::Image<unsigned char, 2> > >::PixelProxy<true, void>'
[-Wrange-loop-analysis]
    for (const auto pixel : range)
                    ^
[CTest: warning matched]
/Users/builder/externalModules/Core/Common/test/itkShapedImageNeighborhoodRangeGTest.cxx:547:10:
note: use reference type 'const
itk::ShapedImageNeighborhoodRange<const
itk::Image<unsigned char, 2>,
itk::ZeroFluxNeumannImageNeighborhoodPixelAccessPolicy<const
itk::Image<unsigned char, 2> >
>::PixelProxy<true, void> &' to prevent
>copying
    for (const auto pixel : range)
        ^~~~~~~~~~~~~~~~~~
```

raised at:
https://open.cdash.org/viewBuildError.php?type=1&buildid=7103973

## PR Checklist

### DIFF
--- a/Modules/Core/Common/test/itkShapedImageNeighborhoodRangeGTest.cxx
+++ b/Modules/Core/Common/test/itkShapedImageNeighborhoodRangeGTest.cxx
@@ -544,7 +544,7 @@ TEST(ShapedImageNeighborhoodRange, IteratorRetrievesSamePixelValuesAsConstNeighb
 
     itk::SizeValueType i = 0;
 
-    for (const auto pixel : range)
+    for (const PixelType pixel : range)
     {
       EXPECT_EQ(pixel, constNeighborhoodIterator.GetPixel(i));
       ++i;


Modules/Core/Common/test/itkShapedImageNeighborhoodRangeGTest.cxx
```diff
@@ -544,7 +544,7 @@ TEST(ShapedImageNeighborhoodRange, IteratorRetrievesSamePixelValuesAsConstNeighb
 
     itk::SizeValueType i = 0;
 
-    for (const auto pixel : range)
+    for (const auto & pixel : range)
```
Leengit, line 547: The `auto` is setting the pixel type to a `PixelProxy`, which is great in some circumstances, but in this case I'd venture that a simple `PixelType` is what is intended.  Especially because `PixelType` is `unsigned char` here, I might instead go with
```c++
for (const PixelType pixel : range)
```

N-Dekker:
@jhlegarreta Thanks for your pull request! Do you possibly know why it was not encountered before? This line of code has been like that already from the very initial commit, 9 April 2018:

https://github.com/InsightSoftwareConsortium/ITK/blob/fd8b105a2fb0f51eb993145fb02db91226607789/Modules/Core/Common/test/itkShapedImageNeighborhoodRangeGTest.cxx#L339

Modules/Core/Common/test/itkShapedImageNeighborhoodRangeGTest.cxx
N-Dekker: In general I think the syntax `for (const auto value : range)` should be fully supported, especially for generic code. But if that causes warnings that need to be addressed, I'm OK with both `for (const PixelType pixel...)` and `for (const auto & pixel...)`.

jhlegarreta: Changed to `const PixelType pixel` in 4a75ae8.

jhlegarreta:
> @jhlegarreta Thanks for your pull request! Do you possibly know why it was not encountered before? This line of code has been like that already from the very initial commit, 9 April 2018:

I noticed that @N-Dekker, but do not have any idea why this has only been reported in the last few days. Maybe the build that started reporting it got an update (downgrade?) that triggered the warning.

N-Dekker (approved):
Thank you @jhlegarreta ! Approved!  I see now, I did follow the suggested approach of @Leengit already before, to spell out `PixelType` 😄

Leengit (approved):

dzenanz (approved):

PR #2424, opened 2021-03-19
ENH: update remote modules

dzenanz:
As remote modules are being actively modified right now, we should update them more frequently.


SimonRit (approved):
Thanks

Issue #2426, opened 2021-03-19
ContourExtractor2DImageFilter Valgrind defects


blowekamp:
The nightly valgrind analysis is detecting memory defects in the recently updated CountourExtractor2DImageFilter.

https://open.cdash.org/viewDynamicAnalysisFile.php?id=8334354

Sample issue:
```

==4696== Command: /tmp/bld/ITK-build/bin/ITKPathTestDriver itkContourExtractor2DImageFilterTest /tmp/bld/ITK-build/ExternalData/Modules/Filtering/Path/test/Input/ContourExtractor2DTest.tif
==4696== Parent PID: 55
==4696== 
UMR ==4696== Invalid read of size 8
==4696==    at 0x2012EC: std::pair const, std::_List_iterator >::ContourType> >::pair&, true>(itk::ContinuousIndex&, std::_List_iterator >::ContourType> const&) (stl_pair.h:312)
==4696==    by 0x1F2328: itk::ContourExtractor2DImageFilter >::AddSegment(itk::ContinuousIndex, itk::ContinuousIndex, itk::ContourExtractor2DImageFilter >::ContourData&) (itkContourExtractor2DImageFilter.hxx:545)
==4696==    by 0x1E8367: itk::ContourExtractor2DImageFilter >::CreateSingleContour(unsigned char, itk::Image const*, itk::ImageRegion<2u>, unsigned long, std::list >::ContourType, std::allocator >::ContourType> >&) (itkContourExtractor2DImageFilter.hxx:191)
==4696==    by 0x1E439A: itk::ContourExtractor2DImageFilter >::GenerateData() (itkContourExtractor2DImageFilter.hxx:72)
==4696==    by 0x679AE50: itk::ProcessObject::UpdateOutputData(itk::DataObject*) (itkProcessObject.cxx:1708)
==4696==    by 0x67BDB1D: itk::DataObject::UpdateOutputData() (itkDataObject.cxx:376)
==4696==    by 0x67BD7E6: itk::DataObject::Update() (itkDataObject.cxx:308)
==4696==    by 0x6799A5D: itk::ProcessObject::Update() (itkProcessObject.cxx:1286)
==4696==    by 0x1DA5A2: itkContourExtractor2DImageFilterTest(int, char**) (itkContourExtractor2DImageFilterTest.cxx:599)
==4696==    by 0x16D0F5: main (ITKPathTestDriver.cxx:202)
[...]

PR #2427, opened 2021-03-19
COMP: Address ContourExtractor2DImageFilter valgrind defects

Leengit:
Address Issue #2426.

## PR Checklist

### DIFF
--- a/Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.hxx
+++ b/Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.hxx
@@ -444,9 +444,9 @@ ContourExtractor2DImageFilter<TInputImage>::AddSegment(VertexType from, VertexTy
   {
     // We need to connect these two contours with the current arc. The act of
     // connecting the two contours will add the needed arc.
-    const ContourContainerIterator & tail(newTail->second);
+    const ContourContainerIterator tail(newTail->second);
     itkAssertOrThrowMacro((tail->front() == to), "End doesn't match Beginning");
-    const ContourContainerIterator & head(newHead->second);
+    const ContourContainerIterator head(newHead->second);
     itkAssertOrThrowMacro((head->back() == from), "Beginning doesn't match End");
     if (head == tail)
     {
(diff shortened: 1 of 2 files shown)


Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.hxx
```diff
@@ -486,7 +486,8 @@ ContourExtractor2DImageFilter<TInputImage>::AddSegment(VertexType from, VertexTy
         // Now remove the old end of 'head' from the ends map and add
         // the new end.
         contourData.m_ContourEnds.erase(newHead);
-        contourData.m_ContourEnds.insert(VertexToContourContainerIteratorMapKeyValuePair(head->back(), head));
+        const VertexToContourContainerIteratorMapKeyValuePair newEnd(head->back(), head);
+        contourData.m_ContourEnds.insert(newEnd);
```
blowekamp, line 490: Hello, Thank you for addressing the issue about the valgrind defects I created. 

I'm sure you PR is just fine the way it is... but I thought I'd give one of those extra suggestions :)
Consider:

`contourData.m_ContourEnds.emplace(head->back(), head)`

dzenanz (approved):
Looks good at a glance. Brad's suggestion sounds helpful.

Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.hxx
Leengit: Good, thank you.  I have now done that.

Leengit:
In addition to using `emplace`, with this latest push I have also changed a `std::unordered_map`  to a `std::map`.  Although the former, is _O(n)_ and the latter is _O(n lg n)_, only the latter guarantees that an iterator set before an erase or insert will remain valid afterwards, and we need that.  Maybe this is what valgrind was trying to tell us?

jhlegarreta:
@Leengit thanks for giving this a try. Looks like the class is giving a lot of work :pensive: .

I've been noticing that the test is also failing for some of the nightly builds, e.g.:
https://open.cdash.org/test/367698202

Just in case the information might be relevant, I had a look at the failure report: it looks like the  `AddSegment` method is triggering a segfault. The proposed changes to the implementation file in this PR happen in that method. You may want to have a look at the error report in the link aboce if you have not done so/if you get a chance. Maybe the information provided helps to better understand the root of the problem.

Leengit:
@jhlegarreta -- I am hoping that the switch from `std::unordered_map` to `std::map` will address the segmentation faults.  But if you notice otherwise, please submit a follow up issue.

jhlegarreta:
@Leengit Valgrind is reporting no defects now !! :fireworks: Thanks for having fixed this.

PR #2429, opened 2021-03-19
STYLE: Remove extra itk prefix from support Python modules

thewtex:
Since these are now in the `itk` package, they do not need the prefix.

itkTemplate is named template_class to avoid a naming conflict with the
*template* function in extras.

This is a follow-up to 22998bd2266c1680db5152302a2e82f8f9cb228f


dzenanz (approved):
LGTM. Perhaps mention in commit message "This is the follow-up to ..." and find Hans' commit which moved this stuff into a separate `itk.support` package.

dzenanz (approved):
Awesome!

Issue #2428, opened 2021-03-19
Add direct access to itk.array_from_vector_container, itk.array_view_from_vector_container, itk.vector_container_from_array


thewtex:
These functions are defined on `itk.PyVectorContainer`, but they should be exposed like `itk.array_from_image` by adding them to `itkExtras.py`.

Reference:

https://github.com/InsightSoftwareConsortium/ITK/blob/4c785c2176785e57048d67ec618c8712bcc99782/Wrapping/Generators/Python/itk/support/itkExtras.py#L196-L242

Note: we should skip the `CamelCase` versions.

tbirdso:
Should we continue to support a tuple or list for `ttype`? If so, for what purpose?

thewtex:
Yes -- because for all the other `itkTemplate` uses, a tuple is supported.

PR #2430, opened 2021-03-19
ENH: Expose numpy / vector container Python interface

tbirdso:
Add direct access to `itk.array_from_vector_container`, `itk.array_view_from_vector_container`, `itk.vector_container_from_array`.

Closes [Issue 2428](https://github.com/InsightSoftwareConsortium/ITK/issues/2428).

## PR Checklist

### DIFF
--- a/Wrapping/Generators/Python/itk/support/extras.py
+++ b/Wrapping/Generators/Python/itk/support/extras.py
@@ -61,6 +61,9 @@ __all__ = [
     "image_from_array",
     "GetImageViewFromArray",
     "image_view_from_array",
+    "array_from_vector_container",
+    "array_view_from_vector_container",
+    "vector_container_from_array",
     "GetArrayFromVnlVector",
     "array_from_vnl_vector",
     "GetArrayViewFromVnlVector",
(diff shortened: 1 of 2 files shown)


tbirdso:
It seems that inputs to `PyVectorContainer` are fixed with 64-bit integer indices.

From [itkPyVectorContainer](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Bridge/NumPy/include/itkPyVectorContainer.h#L52):
>   using VectorContainerType = VectorContainer<IdentifierType, TElement>;

From [itkIntTypes.h](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkIntTypes.h#L66):
>using IdentifierType = SizeValueType;
...
using IndexValueType = int64_t;

`VectorContainer` is wrapped for `unsigned char`, `unsigned int`, and `unsigned long long` index types even with `unsigned long long` wrapping deselected. 

@thewtex am I understanding this right, and is it intentional that the `numpy` interface is only wrapped for `itk::VectorContainer<unsigned long long, T>` inputs? If so, should the Python functional interface reflect this, and if not, can we adjust `PyVectorContainer` to receive generic `VectorContainer` templates?

thewtex:
@tbirdso on the Python side, there is only one type used for the index type, `Py_ssize_t`.

> can we adjust PyVectorContainer to receive generic VectorContainer templates?

Yes, this can be added.

thewtex:
Following integration of #2434 , we will need to add these to `extras.__all__`.

Wrapping/Generators/Python/itk/support/extras.py
```diff
@@ -385,6 +385,80 @@ def GetImageViewFromArray(arr, is_vector: bool = False, ttype=None):
...
+#  >>> itk.array_from_vector_container(x)
+# Traceback (most recent call last):
+#  File "<stdin>", line 1, in <module>
+#  File "C:\repos\ITK-build\Wrapping\Generators\Python\itk\support\itkExtras.py", line 328, in array_from_vector_container
+#    return itk.PyVectorContainer[keys[0]].array_view_from_vector_container(container)
+#  File "C:\repos\ITK-build\Wrapping\Generators\Python\itk\itkPyVectorContainerPython.py", line 622, in array_view_from_vector_container
+#    memview       = itkPyVectorContainerUC._array_view_from_vector_container(vector_container)
+# TypeError: in method 'itkPyVectorContainerUC__array_view_from_vector_container', argument 1 of type 'itkVectorContainerULLUC *
+
+
```
Leengit, line 419: Is this error still arising even with your fix?

tbirdso: Hi Lee, this error is the same as the one above where `PyVectorContainer` is only wrapped to accept `VectorContainer`s with index type `unsigned long long` (aka `Py_ssize_t`). Per discussion above the best plan is probably to expose the `numpy` Python interface in this PR with existing wrapping and then add additional wrapping to `PyVectorContainer` in a subsequent PR.

I'll go ahead and remove the inline comment as it is now outdated. The error still occurs but is not a blocking case for this PR.

Leengit: Got it.  It looks like VectorContainer needs wrapping changes too; `itkVectorContainer.wrap` wrapping supports `${ITKM_IT}` as its first template argument but not "ULL" too.

Leengit (approved):
LGTM.

tbirdso:
I went ahead and updated `itkPyVectorContainerTest` to rely on the functional interface. The test does not cover conversion from a 2D `numpy` array back to a container of non-scalars. When I attempt to convert an array of 3D points back to an `itk::VectorContainer` object in my local environment I get an error:

> \>>> arr2
array([[1., 2., 3.],
       [3., 3., 3.]], dtype=float32)
\>>> itk.PyVectorContainer[itk.Point[itk.F,3]].vector_container_from_array(arr2)
  File "<stdin>", line 1, in <module>
  File "C:\repos\ITK-build-dev\Wrapping\Generators\Python\itk\itkPyVectorContainerPython.py", line 479, in vector_container_from_array
    assert ndarr.ndim == 1 , "Only arrays of 1 dimension are supported."
AssertionError: Only arrays of 1 dimension are supported.

This behavior predates the current PR and is unrelated to the Python interface. Is this expected/known behavior? (@thewtex )

Wrapping/Generators/Python/itk/support/extras.py
tbirdso: Apologies, `PyVectorContainer` takes in `IdentiferType` which is equivalent to `uint64_t` and `Py_ssize_t` but may not be equivalent to `unsigned long long` depending on the system. It is correct to  wrap `VectorContainer` with `${ITKM_IT}`. I've updated my above comment accordingly.

What is the best way to discern system wrapping types? For instance, on my system `${ITKM_IT}` seems to wrap for `itk.ULL` but on other systems I understand it may wrap for `itk.UL`.

Leengit: I don't know how to get "UL" or "ULL" from `Py_ssize_t` or `IdentifierType`.  Maybe you can get at the parameter types of one of the functions:
```python
itk.PyVectorContainer[itk.F].array_view_from_vector_container
itk.PyVectorContainer[itk.F]._array_view_from_vector_container
```
etc., and use that.

tbirdso: Just put in [Draft PR 2443](https://github.com/InsightSoftwareConsortium/ITK/pull/2443) which tackles typing issues discussed here. Going to go ahead and resolve this thread but please feel free to add comments there.

Wrapping/Generators/Python/itk/support/extras.py
```diff
@@ -385,6 +388,69 @@ def GetImageViewFromArray(arr, is_vector: bool = False, ttype=None):
...
+        DataType = itk.template(container)[1][1]
+    keys = [k for k in itk.PyVectorContainer.keys() if k[0] == DataType]
+    if len(keys) == 0:
+        raise RuntimeError("No suitable template parameter can be found.")
+    # Create numpy array of the type of the input container
+    return itk.PyVectorContainer[keys[0]].array_view_from_vector_container(container)
+
+
+def vector_container_from_array(arr, ttype=None):
+    """Get an Array view with the content of the vector container"""
```
dzenanz, line 434: docstring is wrong due to copy-paste error

tbirdso: Good catch! Fixed.

thewtex (approved):
:100:

PR #2435, opened 2021-03-23
STYLE: Remove `;` from itkExceptionMacro PatchBasedDenoisingImageFilter

N-Dekker:
The semi-colons at the end of the arguments passed to
`itkExceptionMacro` calls in the implementation of
`PatchBasedDenoisingImageFilter` appear unnecessary.

### DIFF
--- a/Modules/Filtering/Denoising/include/itkPatchBasedDenoisingImageFilter.hxx
+++ b/Modules/Filtering/Denoising/include/itkPatchBasedDenoisingImageFilter.hxx
@@ -231,7 +231,7 @@ PatchBasedDenoisingImageFilter<TInputImage, TOutputImage>::Initialize()
     // throw an exception when this isn't the case
     itkExceptionMacro(<< "Patch is larger than the entire image (in at least one dimension)."
                       << "\nImage region: " << largestRegion << "\nPatch length (2*radius + 1): "
-                      << this->GetPatchDiameterInVoxels() << "\nUse a smaller patch for this image.\n";);
+                      << this->GetPatchDiameterInVoxels() << "\nUse a smaller patch for this image.\n");
   }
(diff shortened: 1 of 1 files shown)


dzenanz (approved):

thewtex (approved):

Modules/Filtering/Denoising/include/itkPatchBasedDenoisingImageFilter.hxx
```diff
@@ -390,7 +390,7 @@ PatchBasedDenoisingImageFilter<TInputImage, TOutputImage>::EnforceConstraints()
     if (m_ImageMax[ic] <= m_ImageMin[ic])
     {
       itkExceptionMacro(<< "Each image component must be nonconstant.  "
-                        << "Component " << ic << " has the constant value " << m_ImageMax[ic] << ".\n";);
+                        << "Component " << ic << " has the constant value " << m_ImageMax[ic] << ".\n");
```
N-Dekker, line 393: I don't think it's necessary for an exception message to end with "\n" either, but I wasn't sure either if removing the "\n"  would make some users unhappy..., what do you think?

dzenanz: You need to look at how formatted messages look on the console before deciding whether to remove `\n`.

N-Dekker: > You need to look at how formatted messages look on the console before deciding whether to remove \n.

OK, thanks @dzenanz, then I prefer not to decide, and leave the \n as they are, for now.  The semi-colons are obviously mistaken anyway!

PR #2438, opened 2021-03-24
ENH: Trigger PR labeler workflows from forks

jhlegarreta:
Change the `GitHub` webhook event to a `pull_request_target` event to
trigger the PR labeler workflows from forks.

See documentation:
https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target
and
https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/

## PR Checklist

### DIFF
--- a/.github/workflows/label-pr.yml
+++ b/.github/workflows/label-pr.yml
@@ -1,7 +1,7 @@
 name: Label PRs
 
 on:
-  - pull_request
+  - pull_request_target
 
 jobs:
   build:


jhlegarreta:
Let's see if this makes the trick.

dzenanz (approved):
It would be a great improvement if it worked.

dzenanz:
If this works, will it allows us to also [apply clang-format on PRs](https://github.com/InsightSoftwareConsortium/ITK/blob/master/.github/workflows/apply-clang-format.yml)?

jhlegarreta:
> If this works, will it allows us to also apply clang-format on PRs from forks?

Not 100% sure how the commands throughout the comments on a PR work, but chances are that it might.

jhlegarreta:
:sob: The Actions failed:
https://github.com/InsightSoftwareConsortium/ITK/runs/2188115680?check_suite_focus=true

The webhook event is not supported yet:
https://github.com/srvaroa/labeler/issues/13

:pensive:.

dzenanz:
Should we revert this PR?

jhlegarreta:
> Should we revert this PR?

Not sure; the workflow does not show the error in the PR/marks the check as passing, and hence it will not be considered as an unsuccessful check when merging the PRs (somewhat similar to what it was before).

Maybe we could track it as an issue? But even then, we purely depend on the author of the labeler action or someone else (I think I might not have the bandwidth to learn `go` and to try at this time) contributing provided that the feature gets merged into a release in google's `go-github` repository.

We could ask to the google folks if they will merge it, or propose the labeler action repo owner to switch to track the `go-github` `master` branch (the latter being probably not desired).

PR #2443, opened 2021-03-24
ENH: Add PyVectorContainer Wrappings to match VectorContainer

tbirdso:
Updated `PyVectorContainer` template arguments to allow conversion from `VectorContainer`s of index types other than `IdentifierType`. Wrapping, testing, and functional interface updates also included.

Note that `PyVectorContainer` still does not fully cover all possible `VectorContainer` inputs but this PR gets it much closer, wrapping for different index types as well as a variety of scalar, vector, and offset data types not previously covered.

This PR depends on [PR 2430](https://github.com/InsightSoftwareConsortium/ITK/pull/2430). Feedback welcome!

## PR Checklist
- [x] Updated test (or behavior not changed)
- [x] Added Python wrapping to existing files as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5

### DIFF
--- a/Modules/Bridge/NumPy/include/itkPyVectorContainer.hxx
+++ b/Modules/Bridge/NumPy/include/itkPyVectorContainer.hxx
@@ -24,9 +24,9 @@
 namespace itk
 {
 
-template <class TElement>
+template <typename TElementIdentifier, typename TElement>
 PyObject *
-PyVectorContainer<TElement>::_array_view_from_vector_container(VectorContainerType * vector)
+PyVectorContainer<TElementIdentifier, TElement>::_array_view_from_vector_container(VectorContainerType * vector)
 {
   PyObject * memoryView = NULL;
   Py_buffer  pyBuffer;
@@ -56,9 +56,9 @@ PyVectorContainer<TElement>::_array_view_from_vector_container(VectorContainerTy
   return memoryView;
 }
 
-template <class TElement>
-const typename PyVectorContainer<TElement>::VectorContainerType::Pointer
-PyVectorContainer<TElement>::_vector_container_from_array(PyObject * arr, PyObject * shape)
+template <typename TElementIdentifier, typename TElement>
+const typename PyVectorContainer<TElementIdentifier, TElement>::VectorContainerType::Pointer
+PyVectorContainer<TElementIdentifier, TElement>::_vector_container_from_array(PyObject * arr, PyObject * shape)
 {
   PyObject * obj = NULL;
   PyObject * shapeseq = NULL;
--- a/Modules/Bridge/NumPy/wrapping/itkPyVectorContainer.wrap
+++ b/Modules/Bridge/NumPy/wrapping/itkPyVectorContainer.wrap
@@ -1,19 +1,116 @@
+itk_wrap_include("set")
+
 itk_wrap_class("itk::PyVectorContainer")
-  UNIQUE(types "${WRAP_ITK_SCALAR};D;UI")
-  foreach(d ${ITK_WRAP_IMAGE_DIMS})
-    # Wrap itk.Point[itk.F, ${d}]
-    list(APPEND types PF${d})
+  # double is always needed by KernelTransform
+  UNIQUE(scalar_types "${WRAP_ITK_SCALAR};D")
+  UNIQUE(index_types "IT;UC")
+  foreach(tt ${index_types})
+    foreach(t ${scalar_types})
+      # These are wrapped below with "vectypes"
+      if(NOT t IN_LIST WRAP_ITK_USIGN_INT)
+        itk_wrap_template("${ITKM_${tt}}${ITKM_${t}}" "${ITKT_${tt}},${ITKT_${t}}")
+        set(PyVectorContainerTypes ${ITKM_${tt}}${ITKM_${t}})
+        configure_file( ${CMAKE_CURRENT_SOURCE_DIR}/PyVectorContainerBuffer.i.in
+                        ${CMAKE_CURRENT_BINARY_DIR}/PyVectorContainerBuffer.i.temp
+                        @ONLY)
+        file(READ ${CMAKE_CURRENT_BINARY_DIR}/PyVectorContainerBuffer.i.temp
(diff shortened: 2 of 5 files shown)


Modules/Bridge/NumPy/include/itkPyVectorContainer.hxx
```diff
@@ -24,9 +24,9 @@
 namespace itk
 {
 
-template <class TElement>
+template <class TElementIdentifier, class TElement>
```
Leengit, line 27:
```c++
template <typename TElementIdentifier, typename TElement>
```

Modules/Bridge/NumPy/include/itkPyVectorContainer.hxx
```diff
@@ -56,9 +56,9 @@ PyVectorContainer<TElement>::_array_view_from_vector_container(VectorContainerTy
   return memoryView;
 }
 
-template <class TElement>
-const typename PyVectorContainer<TElement>::VectorContainerType::Pointer
-PyVectorContainer<TElement>::_vector_container_from_array(PyObject * arr, PyObject * shape)
+template <class TElementIdentifier, class TElement>
```
Leengit, line 59:
```c++
template <typename TElementIdentifier, typename TElement>
```

Modules/Bridge/NumPy/wrapping/itkPyVectorContainer.wrap
```diff
@@ -1,19 +1,33 @@
 itk_wrap_class("itk::PyVectorContainer")
+  UNIQUE(identifiertypes "${WRAP_ITK_SCALAR};UC;UI;UL;ULL")
```
Leengit, line 2: I don't know but will ask anyway ... would using `${ITKM_IT}` instead of `UL` be more appropriate?  That way we get `UL` only if it is needed. 

I assume that we want `UC`, `UI`, and `ULL` regardless of what extra wrapping the user does or does not request.  If so, is the same true for `US`?

Modules/Bridge/NumPy/wrapping/itkPyVectorContainer.wrap
```diff
@@ -1,19 +1,33 @@
...
-    # Wrap itk.Point[itk.F, ${d}]
-    list(APPEND types PF${d})
+    # Wrap itk.Offset[${d}]
+    list(APPEND types O${d})
+    # Wrap itk.ContinuousIndex[itk.D,${d}]
+    list(APPEND types CID${d})
+    foreach(t ${real_types})
+      # Wrap itk.Point[${t}, ${d}]
+      list(APPEND types P${t}${d})
+      # Wrap itk.Vector[${t}, ${d}]
```
Leengit, line 14: If we want `itk.Vector` do we want `itk.CovariantVector` too?

Modules/Bridge/NumPy/wrapping/itkPyVectorContainer.wrap
thewtex: We do not want `WRAP_ITK_SCALAR` here -- only the types that were found to be used / wrapped in the toolkit.

tbirdso: Both good points -- the index type list should be `UC;UI;IT` to match the current wrappings for `VectorContainer` as closely as possible. Updating momentarily.

Modules/Bridge/NumPy/wrapping/itkPyVectorContainer.wrap
tbirdso: `VectorContainer` is not currently wrapped for `CovariantVector`, should we add that wrapping?

Leengit: I don't see wrapping for Vector or CovariantVector in `itkVectorContainer.wrap` so that is self consistent.  The only place I see wrapping for Vector in `itkPyVectorContainer.wrap` is the new lines 14 & 15 in this pull request:
```cmake
# Wrap itk.Vector[${t}, ${d}]
list(APPEND types V${t}${d})
```
I am not sure I understand the motivation there, so I am not sure whether CovariantVector is also needed.  If it is, the code would be something like:
```cmake
# Wrap itk.Vector[${t}, ${d}] and itk.CovariantVector[${t}, ${d}]
list(APPEND types V${t}${d})
list(APPEND types CV${t}${d})
```

thewtex: Let's avoid wrapping `CovariantVector` until it is needed.

tbirdso: Following up briefly, `itk::VectorContainer` is wrapped for `itk::Vector` in [`itkVectorContainer.wrap`](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/wrapping/itkVectorContainer.wrap#L16):
>  itk_wrap_template("${ITKM_IT}${ITKM_V${t}${d}}" "${ITKT_IT},${ITKT_V${t}${d}}")

The primary motivation for updating `PyVectorContainer` wrapping here is to improve ease of access to Python array representation of `VectorContainer` objects. `itk.array_from_vector_container` must translate an input `VectorContainer` to an equivalently wrapped `PyVectorContainer` prior to representation as a numpy Array. The goal is to wrap `PyVectorContainer` most nearly for common use cases where we might try to translate a `VectorContainer` to numpy.

tbirdso:
Previously saw [build warnings](https://open.cdash.org/viewBuildError.php?type=1&buildid=7120711) from attempting to wrap for template arguments beyond those currently wrapped for `VectorContainer`. I refactored `itkPyVectorContainer.wrap` to precisely follow the wrappings set out in `itkVectorContainer.wrap`. The result is functional but not pretty. I'm open to feedback on how to make the wrapping file more readable and extensible in the future, though any changes we make to `itkPyVectorContainer.wrap` should likely also propagate to `itkVectorContainer.wrap`.

thewtex (approved):
💯

PR #2422, opened 2021-03-18
BUG: Fix regression in PolygonSpatialObject::IsInsideInObjectSpace

rflanz:
This is a followup to the bugfix for issue #1082 where there was a
regression in the code on determining the inside/outsideness of points.

There are some cases where the algorithm still fails as there are two
boundary checks in the code which should be identical.
1. Test on adjacent segment pairs from start to finish
2. Test on segment pair from finish point to start point on closed
   polygons.

The previous bugfix fixed the code in item 1, but did not update the
check (which should be identical in item 2).

This should address this issue making both boundary on segments the same.

## PR Checklist
- [x] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [x] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [ ] Added test (or behavior not changed)
- [x] Updated API documentation (or API not changed)
- [x] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [x] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [ ] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples) for all new major features (if any)

Refer to the [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) for
further development details if necessary.

PR #2453, opened 2021-03-29
BUG: Fix regression in PolygonSpatialObject::IsInsideInObjectSpace

dzenanz:
This is a slightly cleaned-up clone of #2422.


dzenanz:
Force-push sets the intensity difference tolerance to zero (from default two) to make sure the test fails without the fix. If default inside value of 255 was used, this would not have been necessary. If someone doesn't mind making the change and regenerating the test baseline they are welcome to update my branch.

jhlegarreta (commented):
Maybe the commit could be drawn from the `master` branch once it gets merged? (not sure how the sync is handled when it is required)

dzenanz:
This commit merges cleanly into master, as PR #2454 shows.

jhlegarreta (approved):

dzenanz:
The failures are unrelated. Merging.

PR #2456, opened 2021-03-29
ENH: Add .process_object attribute to functional filters

thewtex:
Add an attribute on the functional filter interfaces that points back to
the associated itk.ProcessObject that the filter uses internally.


dzenanz (approved):
Why is this needed or desired?

thewtex:
This can be use, e.g.:

- To get the default values for the functional filter.
- Use the functional filter in streaming operations where `UpdateOutputInformation` on the ProcessObject defines the extent required of the input when a region is passed to the filter.

PR #2450, opened 2021-03-27
COMP: fix link error building Universal Binary on Intel Mac

seanm:
Without this, building Universal Binary on an Intel Mac (but not on an ARM Mac) gives link errors for undefined symbols like _itk_png_do_expand_palette_rgb8_neon and a few others.

Introspecting the CPU type of the build machine is fundamentally incompatible with the idea of Universal Binaries.

PR #2459, opened 2021-03-30
COMP: Constrain x86_64 optimization flags according to CMAKE_SYSTEM_PROCESSOR

thewtex:
CMAKE_SYSTEM_PROCESSOR defines the target compilation architecture. When
we want x86_64 optimizations, e.g. -march=corei7, only add these flags
when building for this architecture. On Linux/macOS (GCC/AppleClang),
CMAKE_SYSTEM_PROCESSOR is "x86_64" and on Windows (Visual Studio) it is
"AMD64".

### DIFF
--- a/CMake/ITKSetStandardCompilerFlags.cmake
+++ b/CMake/ITKSetStandardCompilerFlags.cmake
@@ -212,41 +212,43 @@ function(check_compiler_optimization_flags c_optimization_flags_var cxx_optimiza
   set(${c_optimization_flags_var} "" PARENT_SCOPE)
   set(${cxx_optimization_flags_var} "" PARENT_SCOPE)
 
-  if(MSVC)
-    check_avx_flags(InstructionSetOptimizationFlags)
-    if("${CMAKE_SIZEOF_VOID_P}" EQUAL "4")
-       list(APPEND InstructionSetOptimizationFlags
-            /arch:SSE /arch:SSE2)
-    endif()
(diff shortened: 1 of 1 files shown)


seanm:
Introspecting the system processor is antithetical to cross compiling. It'd be nice to see less of it, not more of it.  As you know, I just got rid of a problematic case the other day: https://github.com/InsightSoftwareConsortium/ITK/pull/2450

It's still setting `-march=corei7`, which, in my view, as I've said before, is undesirable:

- If you're building to deploy to the public, you may want to support older processors, which many people out there have
- If you're building for yourself only, you might want an even higher setting, like `-march=core-avx2`

It seems to be a choice that is best for no one, and best left to the person building the code.

thewtex:
> Introspecting the system processor is antithetical to cross compiling. It'd be nice to see less of it, not more of it. As you know, I just got rid of a problematic case the other day: #2450

Note that [CMAKE_SYSTEM_PROCESSOR]https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html) is not the host system processor. This improves cross compilation support -- it is not antithetical to cross compliing.

> It seems to be a choice that is best for no one, and best left to the person building the code.

We do want the person building the code to be able to choose their optimization flags, but we also want good, sensible defaults. Enabling SIMD support on x86_64 is a good default.

dzenanz (approved):

PR #2464, opened 2021-03-30
COMP: Fix CMake warning by setting policy 0120 to OLD

dzenanz:
We will also have to update VNL after https://github.com/vxl/vxl/pull/833 is merged or otherwise addressed.


seanm:
+1

PR #2449, opened 2021-03-26
ENH: Make itk::Matrix trivially copyable, following Rule of Zero

N-Dekker:
Defaulted the default-constructor of `itk::Matrix`, and added a
GoogleTest, to check that it does properly zero-initialize its elements.

Removed its user-defined copy-constructor and copy-assignment, following
the C++ "Rule of Zero". Tested that `itk::Matrix` has thereby become
"trivially copyable" for built-in element types.

PR #2468, opened 2021-03-31
COMP: No longer test `is_trivially_copyable` in itkMatrixGTest on GCC 4

N-Dekker:
Worked around:

> ITK/Modules/Core/Common/test/itkMatrixGTest.cxx:44:15: error:
> 'is_trivially_copyable' is not a member of 'std'

From CDash (https://open.cdash.org/viewBuildError.php?buildid=7131131):
> Site: AWS-BATCH-88ede920-9899-42aa-8b05-b6d829ac45d3
> Build Name: CentOS Coverage
> Build Time: 2021-03-30 04:52:57

Reported by Matt McCormick at pull request
"ENH: Make itk::Matrix trivially copyable, following Rule of Zero"
https://github.com/InsightSoftwareConsortium/ITK/pull/2449

### DIFF
--- a/Modules/Core/Common/test/itkMatrixGTest.cxx
+++ b/Modules/Core/Common/test/itkMatrixGTest.cxx
@@ -41,9 +41,15 @@ Expect_Matrix_default_constructor_zero_initializes_all_elements()
 } // namespace
 
 
+// GCC version 4 does not yet support C++11 `std::is_trivially_copyable`, as
+// GCC 4.8.5 produced an error message on an attempt to build ITK 5 from the
+// master branch (CentOS Coverage, 2021-03-30), saying:
+// > error: 'is_trivially_copyable' is not a member of 'std'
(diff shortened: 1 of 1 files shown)


dzenanz (approved):
I guess we will be able to drop this workaround when we drop GCC4 😄

thewtex (approved):
Thanks for the follow-up @N-Dekker !

N-Dekker:
Thanks for your approval @dzenanz and @thewtex 

For the record, the GCC4 compilation error is reproducible at https://godbolt.org/z/GoaYvn7Pv even when using GCC 4.9.4:
```
#include <type_traits>

int main()
{
    static_assert(std::is_trivially_copyable<int>(), "Fails to compile on GCC 4");
}
```
Output:
> source :9:19: error: 'is_trivially_copyable' is not a member of 'std'

PR #2472, opened 2021-03-31
COMP: Workaround Clang 3 error default initialization const MatrixGTest

N-Dekker:
Worked around a compiler bug encountered with Clang 3.8.1 and
Mac10.10-AppleClang (which appears fixed with more recent Clang
compiler versions), saying:

> error: default initialization of an object of const type
> 'const itk::Matrix' without a user-provided default constructor

Reported by Jon Haitz Legarreta Gorroño (@jhlegarreta) at pull request
"ENH: Make itk::Matrix trivially copyable, following Rule of Zero"
https://github.com/InsightSoftwareConsortium/ITK/pull/2449

### DIFF
--- a/Modules/Core/Common/test/itkMatrixGTest.cxx
+++ b/Modules/Core/Common/test/itkMatrixGTest.cxx
@@ -28,7 +28,15 @@ template <typename TMatrix>
 void
 Expect_Matrix_default_constructor_zero_initializes_all_elements()
 {
-  const TMatrix defaultConstructedMatrix;
+#ifndef __clang__
+  // Clang versions before 3.9.0 reject the `const` here (erroneously).
+  // Mac10.10-AppleClang-dbg-x86_64-static produced an error message on an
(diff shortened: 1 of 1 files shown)


dzenanz (approved):
Maybe we could also drop clang 3.x with ITK 5.3?

thewtex (approved):

jhlegarreta:
Thanks for doing this @N-Dekker :100:. It is not an easy task to keep all compilers happy.

N-Dekker:
> It is proving it hard to keep all compilers happy.

Indeed, Jon! Apparently this particular `const` keyword made the old clang 3.x unhappy. While removing this `const` could cause a warning from Visual C++, when Code Analysis is enabled, saying something like:

> warning C26496: The variable 'defaultConstructedMatrix' does not change after construction, mark it as const (con.4).

And I do think this warning is meaningful, so I'd rather keep the `const` keyword in there!

PR #2474, opened 2021-04-01
ENH: Add explicit `itk::Vector(const std::array &)` constructor

N-Dekker:
Added `itk::Vector` constructor, allowing to specify its elements by a
C++11 `std::array`. Added `itkVectorGTest` to test this constructor.

Follow-up to:
"ENH: Add explicit constructors Point and FixedArray for C++11 std::array"
pull request https://github.com/InsightSoftwareConsortium/ITK/pull/748
commit 8080e037bc76755c5c5ac8d62ae2daf980542227

### DIFF
--- a/Modules/Core/Common/include/itkVector.h
+++ b/Modules/Core/Common/include/itkVector.h
@@ -137,6 +137,11 @@ public:
     : BaseArray(r)
   {}
 
+  /** Explicit constructor for std::array. */
+  explicit Vector(const std::array<ValueType, NVectorDimension> & stdArray)
+    : BaseArray(stdArray)
+  {}
+
   /** Pass-through assignment operator for the Array base class. */
   template <typename TVectorValueType>
   Vector &
--- a/Modules/Core/Common/test/CMakeLists.txt
+++ b/Modules/Core/Common/test/CMakeLists.txt
@@ -626,6 +626,7 @@ set(ITKCommonGTests
       itkSizeGTest.cxx
       itkSmartPointerGTest.cxx
       itkVectorContainerGTest.cxx
+      itkVectorGTest.cxx
       itkWeakPointerGTest.cxx
       itkCommonTypeTraitsGTest.cxx
       itkMetaDataDictionaryGTest.cxx
--- a/Modules/Core/Common/test/itkVectorGTest.cxx
+++ b/Modules/Core/Common/test/itkVectorGTest.cxx
@@ -1,0 +1,56 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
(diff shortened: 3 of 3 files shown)


dzenanz (approved):

Modules/Core/Common/include/itkVector.h
```diff
@@ -137,6 +137,11 @@ class ITK_TEMPLATE_EXPORT Vector : public FixedArray<T, NVectorDimension>
     : BaseArray(r)
   {}
 
+  /** Explicit constructor for std::array. */
+  explicit Vector(const std::array<ValueType, NVectorDimension> & stdArray)
```
Leengit, line 141: Thank you for making and submitting this change.  Would you say more about why you specified the constructor as `explicit`?  In particular, if there is a use case where implicit use of this constructor is prone to errors, what is it? -- and perhaps add the explanation to `itkVector.h`.

Leengit:
Would the motivation for adding this constructor to `itk::Vector` apply also to `itk::CovariantVector`?

If, yes, then going a little farther out on that branch ... what about adding a function to mimic https://github.com/InsightSoftwareConsortium/ITK/blob/60c2920599013d20293432902cb9d775eb87d0b2/Modules/Core/Common/include/itkVariableLengthVector.h#L352 in `itk::VariableLengthVector`? -- the pair of arguments `datain` and `sz` would be replaced with a single `std::array` argument that supplied the data and (implicitly) the size.

Modules/Core/Common/include/itkVector.h
N-Dekker: >  Would you say more about why you specified the constructor as explicit? 

@Leengit Thanks for asking  😃 Suppose there is a function that has an `itk::Point` and an `itk::Vector` as argument.  For example:

    // Example
    Point<double> AddVector(Point<double>, Vector<double>);

If both `itk::Point` and an `itk::Vector`  would have an implicit constructor, converting from `std::array`, a user who would pass two arrays as arguments might accidentally pass the arrays in the wrong order. Because of the `explicit` keyword, the user must be explicit about which `std::array` has the point coordinates, and which array has the vector elements:  

    // User code:
    const auto point = AddVector(Point<double>(stdArray1), Vector<double>(stdArray2));

Conceptually, `itk::Vector` and `itk::Point` are more specific than the general purpose `std::array`. I think it's good practice to make conversion from a more general data type to a more specific type explicit. Do you agree?

Leengit:
> If, yes, then going a little farther out on that branch ... what about adding a `template <unsigned int NVectorDimension>` constructor to mimic

Hmm, a templated constructor might have adverse implications for Python wrapping....  That could cancel out any value for the possible templated constructor for `itk::VariableLengthVector` that I just asked about.

N-Dekker:
> Would the motivation for adding this constructor to `itk::Vector` apply also to `itk::CovariantVector`?

@Leengit I think so, yes. But only if users would want to have it, of course. I'm proposing to add `itk::Vector(const std::array &)` now because  I would like to use it myself  😺 

Regarding `VariableLengthVector` I'm not entirely sure, because `std::array` isn't _variable length_. But of course, if it might make any user happy, it could be considered.

Leengit (approved):

PR #2477, opened 2021-04-06
STYLE: Matrix constructors vnl_matrix/vnl_matrix_fixed explicit/implicit

N-Dekker:
Declared the `Matrix(const vnl_matrix &)` constructor `explicit`, while
removing the `explicit` keyword from `Matrix(const InternalMatrixType &)`,
as the latter one is the safest constructor of the two.
(Note that `InternalMatrixType = vnl_matrix_fixed<T, NRows, NColumns>`.)

Added the corresponding static asserts to itkMatrixGTest.

### DIFF
--- a/Modules/Core/Common/test/itkMatrixGTest.cxx
+++ b/Modules/Core/Common/test/itkMatrixGTest.cxx
@@ -19,7 +19,7 @@
 // First include the header file to be tested:
 #include "itkMatrix.h"
 #include <gtest/gtest.h>
-#include <type_traits> // For is_trivially_copyable.
+#include <type_traits> // For is_convertible and is_trivially_copyable.
 
 
 namespace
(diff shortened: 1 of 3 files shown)


Modules/Core/Common/test/itkMatrixGTest.cxx
```diff
@@ -54,9 +54,35 @@ Expect_GetIdentity_returns_identity_matrix()
...
+    vnl_matrix_fixed<typename TMatrix::ValueType, TMatrix::RowDimensions, TMatrix::ColumnDimensions>,
+    TMatrix>();
+}
+
 } // namespace
 
 
+static_assert((!vnl_matrix_is_convertible_to_itk_Matrix<itk::Matrix<float>>()) &&
+                (!vnl_matrix_is_convertible_to_itk_Matrix<itk::Matrix<double, 4, 5>>()),
+              "itk::Matrix prevents implicit conversion from vnl_matrix");
```
dzenanz, line 79: prevents -> doesn't prevent

dzenanz (approved):

Modules/Core/Common/test/itkMatrixGTest.cxx
N-Dekker: @dzenanz Well... it _does_ prevent implicit conversion from `vnl_matrix` 😸 But I think I understand what you mean, the static_assert message will appear when it doesn't 🤔  I'll make it "itk::Matrix _should_ prevent implicit conversion from vnl_matrix", hope that's OK to you!

And of course, I'll try to fix the CI failures!

dzenanz (approved):

Modules/Filtering/LabelMap/include/itkShapeLabelMapFilter.hxx
```diff
@@ -326,7 +326,7 @@ ShapeLabelMapFilter<TImage, TLabelImage>::ThreadedProcessLabelObject(LabelObject
   {
     principalMoments[i] = pm(i);
   }
-  MatrixType principalAxes = eigen.V.transpose();
+  MatrixType principalAxes(eigen.V.transpose());
```
N-Dekker, line 329: Note that strictly speaking, this pull request _might_ break _some_ user code, as it required the user to make the construction of an `itk::Matrix` from a ("non-fixed") `vnl_matrix` explicit. But the fix is usually straightforward, as you see here, using parentheses when declaring an `itk::Matrix` from a  `vnl_matrix`, instead of `=`.

hjmjohnson (approved):

PR #2480, opened 2021-04-12
WIP: ENH: Added type hints for wrapped Python classes and methods.

kian-weimer:
This is a method of generating IDE type hints for the dynamically generated ITK Python classes
and dictionaries. The type hints are generated through .pyi files in the itk-stubs directory. 

In this approach, the stub files tell the IDE that the dynamically generated itkTemplate (i.e. 
itk.ListObject) objects used to specify types for classes are actually classes of that name. This 
allows type hints to be used. 

Two methods are typed so that the `New()` method appears to return a generic ‘Typed’ object 
for that class and the dictionary class access appears to return the same class for which the 
`New()` method can be used. 

The `New()` appears to return a generic `Typed` class with the name 
`Typed<class_name>PROXY`. This class then defines generic type hints for all methods of the 
class. This way the user has an idea of what methods are available and parameters are 
needed, but may not know the exact type or dimension for some parameters or returned values. 
This class is not real and is only used to generically describe type hints for all methods of a 
class. 

The current code is a proof of concept using `itk::Statistics::ListSample` as an example. The 
methods and parameters for this class and its parent classes are typed.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/test/CMakeLists.txt
+++ b/Modules/Core/Common/test/CMakeLists.txt
@@ -119,6 +119,7 @@ itkMultiThreaderTypeFromEnvironmentTest.cxx
 itkMultiThreadingEnvironmentTest.cxx
 itkMultiThreaderParallelizeArrayTest.cxx
 itkMultithreadingTest.cxx
+itkMultiThreaderExceptionsTest.cxx
 
 itkMetaProgrammingLibraryTest.cxx
 itkPromoteType.cxx
@@ -573,6 +574,8 @@ itk_add_test(NAME itkRegionSplitterMultidimensionalTest COMMAND ITKCommon2TestDr
 itk_add_test(NAME itkMetaDataObjectTest COMMAND ITKCommon2TestDriver itkMetaDataObjectTest)
 
 itk_add_test(NAME itkMultithreadingTest COMMAND ITKCommon2TestDriver itkMultithreadingTest 100)
+
+itk_add_test(NAME itkMultiThreaderExceptionsTest COMMAND ITKCommon2TestDriver itkMultiThreaderExceptionsTest)
 
 if(ITK_BUILD_SHARED_LIBS AND ITK_DYNAMIC_LOADING)
   macro(BuildClientTestLibrary _name _type)
--- a/Modules/Core/Common/test/itkMultiThreaderExceptionsTest.cxx
+++ b/Modules/Core/Common/test/itkMultiThreaderExceptionsTest.cxx
@@ -1,0 +1,104 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+#include "itkImageSource.h"
+#include "itkTestingMacros.h"
+
+using ThreaderEnum = itk::MultiThreaderBase::ThreaderEnum;
+
+namespace itk
+{
+template <typename TOutputImage>
+class ITK_TEMPLATE_EXPORT DummyImageSource : public ImageSource<TOutputImage>
+{
+public:
+  ITK_DISALLOW_COPY_AND_MOVE(DummyImageSource);
+
+  /** Standard class type aliases. */
+  using Self = DummyImageSource;
+  using Superclass = ImageSource<TOutputImage>;
+  using Pointer = SmartPointer<Self>;
+  using ConstPointer = SmartPointer<const Self>;
+
+  /** Convenient type alias. */
+  using OutputImageRegionType = typename TOutputImage::RegionType;
+
+  /** Method for creation through the object factory. */
+  itkNewMacro(Self);
+
+  /** Run-time type information (and related methods). */
+  itkTypeMacro(DummyImageSource, ImageSource);
+
+  /** Index for which exception is thrown. */
+  itkSetMacro(ExceptionIndex, IndexValueType);
+
+protected:
+  DummyImageSource() = default;
+  ~DummyImageSource() override = default;
+  void
+  GenerateOutputInformation() override
+  {
+    TOutputImage * output = nullptr;
+    output = this->GetOutput(0);
+    typename TOutputImage::RegionType largestPossibleRegion;
+    largestPossibleRegion.SetSize({ 4 });
+    output->SetLargestPossibleRegion(largestPossibleRegion);
+  }
+
+  void
+  DynamicThreadedGenerateData(const OutputImageRegionType & outputRegionForThread) override
+  {
+    if (outputRegionForThread.GetIndex(0) == m_ExceptionIndex)
+    {
+      std::cout << "Exception launched" << std::endl;
+      itkGenericExceptionMacro(<< "Error");
+    }
+  }
+
+private:
+  IndexValueType m_ExceptionIndex{ 0 };
+};
+} // end namespace itk
+
+int
+itkMultiThreaderExceptionsTest(int, char *[])
+{
+  using OutputPixelType = float;
+  constexpr unsigned int Dimension = 1;
+
+  using OutputImageType = itk::Image<OutputPixelType, Dimension>;
+
+  std::set<ThreaderEnum> threadersToTest = { ThreaderEnum::Platform, ThreaderEnum::Pool };
+#ifdef ITK_USE_TBB
+  threadersToTest.insert(ThreaderEnum::TBB);
+#endif // ITK_USE_TBB
+  for (auto thType : threadersToTest)
+  {
+    itk::MultiThreaderBase::SetGlobalDefaultThreader(thType);
+    typename itk::DummyImageSource<OutputImageType>::Pointer dummySrc;
+    dummySrc = itk::DummyImageSource<OutputImageType>::New();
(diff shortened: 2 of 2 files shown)


Wrapping/Generators/Python/CMakeLists.txt
```diff
@@ -137,6 +137,10 @@ if(NOT EXTERNAL_WRAP_ITK_PROJECT)
     support/init_helpers
```
dzenanz, line 137: in commit message genetically -> generically

Wrapping/Generators/Python/itk-stubs/BaseClasses.pyi
```diff
@@ -0,0 +1,407 @@
...
+    @staticmethod
+    def SetGlobalReleaseDataFlag(flag: bool) -> None:
+        """Turn on/off a flag to control whether every object releases its data after being used by a filter.
+            Being a global flag, it controls the behavior of all DataObjects and ProcessObjects."""
+        ...
+
+    @staticmethod
+    def cast(Object) -> DataObject:
+        """Cast an object to itk.DataObject"""
+        ...
```
dzenanz, line 407: perhaps add a new line at the end of file

Wrapping/Generators/Python/itk-stubs/__init__.pyi
```diff
@@ -0,0 +1,121 @@
...
+
+    #! mv type should be TypedVectorPROXY once it exists as the actual type is unknown
+    def SetMeasurementVector(self, instanceId: itk.UL, mv: itk.Vector) -> None:
+        """Replace a measurement with a different measurement."""
+        ...
+
+    @staticmethod
+    def cast(Object) -> TypedListSamplePROXY:
+        """Cast an object to itk::Statistics::ListSample< TMeasurementVector >"""
+        ...
```
dzenanz, line 121: new line

dzenanz (commented):
This should be reviewed by Matt and/or Hans.

kian-weimer:
The commit message has been fixed and newlines added to the ends of files.

Wrapping/Generators/Python/itk-stubs/BaseClasses.pyi
```diff
@@ -0,0 +1,407 @@
+from itkConfig import ITK_GLOBAL_VERSION_STRING as __version__
+from itk.support.extras import *
+from itk.support.init_helpers import *
+from itk.support.types import * # itkCType, F, D, UC, US, UI, UL, SL, LD, ULL, SC, SS, SI, SLL, B
```
thewtex, line 4: We want to use explicit imports, not `import *` as a best practice. There may be types  that conflict with the classes defined here.

thewtex: The convention we used for `itk.support.types` is `import itk.support.types as itkt`.

thewtex (commented):
Cool!

Wrapping/Generators/Python/itk-stubs/BaseClasses.pyi
kian-weimer: Thanks for the feedback! In this update I returned it to the way that it appears in `__init__.py` so that the suggestions would match what is available.

kian-weimer:
The `TypehintGeneration.py` file has been added to help assist in the typing process. This file
has several functions that parse the help text for a class and generate useful Python pyi code
formatted as described above. The Doxygen documentation is parsed and used to add
descriptions to itk class methods when available.

While this parser works on the majority of classes. There are currently three problem areas:
1. It can only generate complete code for instantiatable classes. However, inherited method hints can be displayed for a class.
2. The c++ to python type conversion is not complete. Ideally, there would be a better method for doing this as my current approach isn't ideal. The typed itk classes also make this part difficult because they can end in many different ways. This is currently being worked around by a large block of auto-generated code but does not fix all issues.
3. Typed PROXY class hints have Unions for parameters with varying types. Currently, the types are gathered by running the parser for each type variation of the class. While this works, it's quite slow.

Hints for the `itk.Vector` class have been generated and added to the `__init__.pyi` file. Note
that most C types still don’t have conversions so they are not yet converted.

Wrapping/Generators/Python/CMakeLists.txt
```diff
@@ -171,11 +184,17 @@ if(NOT EXTERNAL_WRAP_ITK_PROJECT)
...
   foreach(_file ${ITK_PYTHON_SUPPORT_MODULES})
     set(install_tgt "${CMAKE_CURRENT_SOURCE_DIR}/itk/${_file}.py")
     list(APPEND ITK_WRAP_PYTHON_FILES ${ITK_WRAP_PYTHON_FILES} "${install_tgt}")
   endforeach()
+  foreach(_file ${ITK_PYTHON_STUB_MODULES})
+    set(install_tgt "${CMAKE_CURRENT_SOURCE_DIR}/${_file}.py")
+    list(APPEND ITK_WRAP_PYTHON_STUB_FILES ${ITK_WRAP_PYTHON_STUB_FILES} "${install_tgt}")
+  endforeach()
   WRAP_ITK_PYTHON_BINDINGS_INSTALL(/itk/support "ITKCommon" ${ITK_WRAP_PYTHON_FILES})
+  WRAP_ITK_PYTHON_BINDINGS_INSTALL(/itk-stubs "ITKCommon" ${ITK_WRAP_PYTHON_STUB_FILES})
```
thewtex, line 197: `/itk-stubs` -> `/itk` to install the files in the package at `itk/__init__.pyi` and `itk/BaseClasses.pyi` (assuming this is where we want these files).

thewtex (changes requested):
@kian-weimer outstanding work!!


> While this parser works on the majority of classes. There are currently three problem areas:

We could address many of these issues also with improving maintainability by incorporating the .pyi generation logic into *ITK/Wrapping/Generators/SwigInterface/igenerator.py*. C++ to Python types are known here. Also, we should use the existing type/class declarations as in *ITK/Wrapping/Generators/*Python/itk/support/types.py* instead of additional PROXY classes. Note:

```
image = itk.Image[itk.UC, 2].New()
isinstance(image, itk.Image) == True
```

There would need to be two stages of .pyi generation, though, since the documentation is not yet available when igenerator.py runs.

thewtex:
Adding @jcfr for ideas related to using this for generating Sphinx API docs.

kian-weimer:
@thewtex Thanks for the reply! Is there any documentation on the igenerator.py file that I could use for reference? I'm not quite sure how I could implement .pyi generation into this file yet.

If you would like, I can remove the use of `Typed_PROXY` classes and keep everything with the same name. The only problem is that something like `itk.Image` will still appear to be an image (in the typehints) before the `New()` method is called on it and thus there will be methods that appear to be available that aren't. By pretending to have the two classes we can differentiate the typehints shown between `itk.image` and the output of `itk.Image.New()`. If this isn't an issue I can make them one. 

I could add images of the difference if you'd like.

thewtex:
> @thewtex Thanks for the reply! Is there any documentation on the igenerator.py file that I could use for reference? I'm not quite sure how I could implement .pyi generation into this file yet

@kian-weimer igenerator.py is a script that takes the C++ abstract syntax tree and generates a Swig interface .i file. However, you could also generate the .pyi file from the same. They best way to understand how it works is to use `pdb` and step through the code.

> I could add images of the difference if you'd like

That would be great, thanks!

kian-weimer:
@thewtex I'll try walking through the code, thank you. 

Here are the images for hints with my current method: 
Typehints for `itk.Image`. 
image

Typehints for output of `itk.Image.New()` 
image

If I combine the .pyi file declarations of `Image` and `TypedImagePROXY` into one the following occurs:
image
image
(I didn't change all the classes so `ImageBase` still appears as a typed proxy)

In the images above, if we use that method that I'm taking but remove the proxy classes the `itkTemplate` objects, like `itk.Image` will display incorrect typehints.

thewtex:
> How would I initially call the code? 

The easiest way to call `igenerator.py` with the right arguments is to use ` -G Ninja` for the [ninja](https://ninja-build.org/) generator. Then, build with the `ninja` command. Then, search through the file `build.ninja` -- this contains all the command line commands used to generate files. Call the command line locally after inserting `import pdb; pdb.set_trace()` in igenerator.py.

>Here are the images for hints with my current method:

Thanks!

I wonder if we can address this by specifying, e.g. by specifying that `itk.Image.__init__` returns an `Image` class instead of an `Image` instance. See, for example:https://adamj.eu/tech/2021/05/16/python-type-hints-return-class-not-instance/

kian-weimer:
@thewtex @hjmjohnson 
I've begin implementing the type hinting into the igernerator.py file. 
Two .pyi files, `__init__.pyi` and `methods.pyi`, are now generated during build time. 
I've changed the formatting slightly so that instantiated classes appear as their actual name (i.e. `Image`) rather than proxies (i.e. `TypedImagePROXY`). 
Currently, hints are created for all non-operator methods passed through the `generate_method` method. This appears to be a problem because several methods (like Swap) are not actually wrapped. Is there any suggestion on how to avoid hinting these methods?
Variable type hints are for provided when possible, but not all C++ to python type conversions are known to me  yet. This will take more work to flush out. Currently types that cannot be converted just do not have type hints, which still makes the hinting useful. 
My old method of getting documentation for methods was not the most efficient, I just pulled information from the itk website. I think there should be a way to do this locally, but I have yet to find it. 
Any advice is appreciated!

kian-weimer:
Note that I've temporarily left my old code version (some commented out) in for reference. 
Below are some images of the current generated .pyi code format and the type hint results:

The `__init__.pyi` file contains the code for the `itkTemplate` before an actual object is created:
image

The `methods.txt` file has method hints for when an object is instantiated:
image

Suggestions for an `itkTemplate` object (before the `New()` method has been called):
image

Suggestions for an `itk.Image` object:
image

thewtex:
@kian-weimer looking like good progress :+1: 

> Currently, hints are created for all non-operator methods passed through the generate_method method. This appears to be a problem because several methods (like Swap) are not actually wrapped. Is there any suggestion on how to avoid hinting these methods?

We will want to re-use the logic in igenerator.py to include / exclude methods when possible and add exceptions when necessary.

dzenanz:
Did you close this PR on purpose?

PR #2484, opened 2021-04-14
Use non-deprecated ITK_DISALLOW_COPY_AND_MOVE in ResampleInPlace

dzenanz:
## PR Checklist
- [x] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [x] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [x] Added test (or behavior not changed)
- [ ] Updated API documentation (or API not changed)
- [ ] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [ ] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [ ] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples) for all new major features (if any)

Fixes:
```text
In file included from /Users/builder/externalModules/Core/Transform/test/itkResampleInPlaceImageFilterTest.cxx:19:
/Users/builder/externalModules/Core/Transform/include/itkResampleInPlaceImageFilter.h:100:3: error: static_assert failed "Replace deprecated ITK_DISALLOW_COPY_AND_ASSIGN with modern ITK_DISALLOW_COPY_AND_MOVE"
  ITK_DISALLOW_COPY_AND_ASSIGN(ResampleInPlaceImageFilter);
  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[CTest: warning matched] /Users/builder/externalModules/Core/Common/include/itkMacro.h:394:5: note: expanded from macro 'ITK_DISALLOW_COPY_AND_ASSIGN'
    static_assert(false, "Replace deprecated ITK_DISALLOW_COPY_AND_ASSIGN with modern ITK_DISALLOW_COPY_AND_MOVE")
    ^             ~~~~~
```

PR #2488, opened 2021-04-16
ENH: Add Python wrapping for `itk::ResampleInPlaceImageFilter`

jhlegarreta:
Add Python wrapping for `itk::ResampleInPlaceImageFilter`.

## PR Checklist

### DIFF
--- a/Modules/Core/Transform/wrapping/itkResampleInPlaceImageFilter.wrap
+++ b/Modules/Core/Transform/wrapping/itkResampleInPlaceImageFilter.wrap
@@ -1,0 +1,8 @@
+itk_wrap_filter_dims(d3 3)
+if(d3)
+  itk_wrap_class("itk::ResampleInPlaceImageFilter" POINTER)
+    foreach(t ${WRAP_ITK_SCALAR})
+      itk_wrap_template("${ITKM_I${t}${d3}}${ITKM_I${t}${d3}}" "${ITKT_I${t}${d3}},${ITKT_I${t}${d3}}")
+    endforeach()
+  itk_end_wrap_class()
+endif()


jhlegarreta:
Have not tested.

Cross-referencing https://github.com/InsightSoftwareConsortium/ITK/pull/2484#issuecomment-820746973.

jhlegarreta:
If I understand correctly from the error log, it looks like the current implementation does not support dealing with vector images, and thus the Python wrapping fails for such images.

Should we remove for now the vector image wrapping or shall I wait until that is implemented to rebase upon that?

Modules/Core/Transform/wrapping/itkResampleInPlaceImageFilter.wrap
```diff
@@ -0,0 +1,14 @@
+itk_wrap_class("itk::ResampleInPlaceImageFilter" POINTER)
+  itk_wrap_image_filter("${WRAP_ITK_SCALAR}" 2)
+  itk_wrap_image_filter("${WRAP_ITK_VECTOR_REAL}" 2)
+  itk_wrap_image_filter("${WRAP_ITK_COLOR}" 2)
+
+  # VectorImages
+  UNIQUE(to_types "UC;${WRAP_ITK_SCALAR}")
```
dzenanz, line 7: Looking at https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/ImageGrid/wrapping/itkRegionOfInterestImageFilter.wrap, we probably don't need special treatment of UC.

dzenanz:
I think the compile problem is that ResampeInPlace only works with Rigid3D transform, and if we try to instantiate it with 2D image there is some sort of undefined operation with vec2 and vec3.

dzenanz:
/home/vsts/work/1/s/Modules/Core/Transform/include/itkResampleInPlaceImageFilter.hxx:68:76: error: no match for 'operator-' (operand types are 'itk::Point<double, 2>::VectorType {aka itk::Vector<double, 2>}' and 'itk::Point<double, 3>::VectorType {aka itk::Vector<double, 3>}')

jhlegarreta:
Thanks for the review @dzenanz :100: !

Removed the treatment for `unsigned char`s, and restricted the wrapping to 3D images only in 0bb11ae.

Modules/Core/Transform/wrapping/itkResampleInPlaceImageFilter.wrap
jhlegarreta: Thanks @dzenanz. Removed in 0bb11ae.

jhlegarreta:
Reading the [CDash error trace](https://open.cdash.org/viewBuildError.php?buildid=7173262) I'd say that the filter does not yet support vector image types, so limiting it to scalars for now.

jhlegarreta:
Given the unrelated `macOS` failures, we cannot know if the Python wrapping for it would work; the rest are passing.

dzenanz (approved):

PR #2490, opened 2021-04-16
BUG: Avoid crash TranslationTransform::SetParameters when too few params

N-Dekker:
When the argument passed to `TranslationTransform` member function
`SetParameters` would contain less than `SpaceDimension` values,
the member function would have undefined behavior, which would typically
just cause a crash.

This commit avoids such undefined behavior by throwing an exception, in
this case. GoogleTest unit tests are included.

Follow-up to:
"BUG: Fix MatrixOffsetTransformBase SetFixedParameters if too few params"
Pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2118
Commit fcb85b3a1e4f78d3c9c3742f213e7acefb87314f

### DIFF
--- a/Modules/Core/Transform/include/itkTranslationTransform.hxx
+++ b/Modules/Core/Transform/include/itkTranslationTransform.hxx
@@ -46,6 +46,12 @@ template <typename TParametersValueType, unsigned int NDimensions>
 void
 TranslationTransform<TParametersValueType, NDimensions>::SetParameters(const ParametersType & parameters)
 {
+  if (parameters.Size() < SpaceDimension)
+  {
+    itkExceptionMacro(<< "Error setting parameters: parameters array size (" << parameters.Size()
+                      << ") is less than expected (SpaceDimension = " << SpaceDimension << ")");
+  }
+
   // Save parameters. Needed for proper operation of TransformUpdateParameters.
   if (&parameters != &(this->m_Parameters))
   {
--- a/Modules/Core/Transform/test/itkTranslationTransformGTest.cxx
+++ b/Modules/Core/Transform/test/itkTranslationTransformGTest.cxx
@@ -1,0 +1,80 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+// First include the header file to be tested:
+#include "itkTranslationTransform.h"
+
+#include <gtest/gtest.h>
+
+#include <algorithm> // For equal.
+#include <numeric>   // For iota.
+
+
+namespace
+{
+
+template <unsigned NDimensions>
+void
+Expect_SetParameters_throws_when_size_is_less_than_SpaceDimension()
+{
+  using TransformType = itk::TranslationTransform<double, NDimensions>;
+  const auto transform = TransformType::New();
+
+  for (unsigned size{}; size < TransformType::SpaceDimension; ++size)
+  {
+    const typename TransformType::ParametersType parameters(size, 0.0);
+    EXPECT_THROW(transform->SetParameters(parameters), itk::ExceptionObject);
+  }
+}
+
+
+template <unsigned NDimensions>
+void
+Expect_SetParameters_sets_translation_offset()
+{
+  using TransformType = itk::TranslationTransform<double, NDimensions>;
+  using ParametersType = typename TransformType::ParametersType;
+
+  const auto transform = TransformType::New();
+
+  // Check setting the offset to (0, 0, 0, ...).
+  transform->SetParameters(ParametersType(NDimensions, 0.0));
+  EXPECT_EQ(transform->GetOffset(), typename TransformType::OutputVectorType());
+
+  // Check setting the offset to (1, 2, 3, ...).
+  ParametersType parameters(NDimensions);
+  std::iota(parameters.begin(), parameters.end(), 1.0);
+  transform->SetParameters(parameters);
+  EXPECT_TRUE(std::equal(parameters.begin(), parameters.end(), transform->GetOffset().begin()));
+}
+
+} // namespace
+
+
+TEST(TranslationTransform, SetParametersThrowsWhenSizeIsLessThanSpaceDimension)
+{
+  Expect_SetParameters_throws_when_size_is_less_than_SpaceDimension<2>();
+  Expect_SetParameters_throws_when_size_is_less_than_SpaceDimension<3>();
+}
+
+
+TEST(TranslationTransform, SetParametersSetsOffset)
+{
+  Expect_SetParameters_sets_translation_offset<2>();
+  Expect_SetParameters_sets_translation_offset<3>();
+}
--- a/Modules/Core/Transform/test/CMakeLists.txt
+++ b/Modules/Core/Transform/test/CMakeLists.txt
@@ -189,5 +189,6 @@ set(ITKTransformGTests
   itkBSplineTransformGTest.cxx
   itkEuler3DTransformGTest.cxx
   itkMatrixOffsetTransformBaseGTest.cxx
+  itkTranslationTransformGTest.cxx
 )
 CreateGoogleTestDriver(ITKTransform "${ITKTransform-Test_LIBRARIES}" "${ITKTransformGTests}")


N-Dekker:
@lassoan This pull request aims to prevent a certain crash of 3D Slicer, similar to pull request #2118: In case a user tries to load a manually edited transform file, which has the wrong number of parameters. The pull request would produce an exception, that can be handled properly by 3D Slicer.

Modules/Core/Transform/test/itkTranslationTransformGTest.cxx
```diff
@@ -0,0 +1,80 @@
...
+{
+
+template <unsigned NDimensions>
+void
+Expect_SetParameters_throws_when_size_is_less_than_SpaceDimension()
+{
+  using TransformType = itk::TranslationTransform<double, NDimensions>;
+  const auto transform = TransformType::New();
+
+  for (unsigned size{}; size < TransformType::SpaceDimension; ++size)
```
lassoan, line 38: `unsigned size{}` is enough for the compiler but unnecessarily vague for the human reader. Why not use `unsigned int size=0` instead?

lassoan (approved):
This is a very useful fix, thank you.

This kind of check and exception throwing is already done for example in BSplineBaseTransform, so it should not be a problem starting to throw exception for this error.

It would be nice to add this fix to all similar SetParameters methods.

Modules/Core/Transform/test/itkTranslationTransformGTest.cxx
N-Dekker: > This is a very useful fix, thank you.
> It would be nice to add this fix to all similar SetParameters methods.

@lassoan  You're welcome. I think these checks are there already in quite a few `SetParameters` and `SetFixedParameters` methods. I just happen to encounter a missing one occasionally!

Modules/Core/Transform/include/itkTranslationTransform.hxx
```diff
@@ -46,6 +46,12 @@ template <typename TParametersValueType, unsigned int NDimensions>
 void
 TranslationTransform<TParametersValueType, NDimensions>::SetParameters(const ParametersType & parameters)
 {
+  if (parameters.Size() < SpaceDimension)
```
dzenanz, line 49: I assume that `parameters.Size() > SpaceDimension` is not a problem?

dzenanz (approved):

Modules/Core/Transform/test/itkTranslationTransformGTest.cxx
N-Dekker: > `unsigned size{}` is enough for the compiler but unnecessarily vague for the human reader. Why not use
> `unsigned int size=0` instead?

@lassoan  I think these are two questions in one. `unsigned int` versus just `unsigned`? The addition of `int` just seems like unnecessary extra typing to me. C++ does not support `unsigned float`. An `unsigned` is always an integer. The word `int` appears redundant to me, in this case.

The notation `unsigned size=0` (with the equal-sign) suggests that a conversion is taking place, from the constant signed `int` zero (`0`) to `unsigned`. In general, conversion from `signed` to `unsigned` may be unsafe. And in this particular case, I would not want an integer conversion anyway. I would just like to have a properly initialized `unsigned`. That is what I'm intending to express by `unsigned size{}`.

As far as I know, `unsigned size{}` is the modern C++ way to do it. Using C++11 "universal initialization". Of course, you may still not like it. But it may eventually be the syntax that gets the best support from C++ tools, like Clang-Tidy. 

I hope that answers your question  😃

Modules/Core/Transform/include/itkTranslationTransform.hxx
N-Dekker: > I assume that parameters.Size() > SpaceDimension is not a problem?

@dzenanz Thanks for asking. Honestly I don't know what to do about `parameters.Size() > SpaceDimension`. It was always accepted silently, it never caused a crash, or "undefined behavior". I just don't know if they might sometimes be used elsewhere.

Modules/Core/Transform/test/itkTranslationTransformGTest.cxx
dzenanz: `unsigned size{}` is a newish syntax, and less knowledgeable programmers may be confused whether this leaves `size` un-initialized, or zero-initialized. Would `unsigned size{0}` or `unsigned size{0u}` be type-safe enough for your taste @N-Dekker? These two formulations make it clear that `size` will be zero.

N-Dekker: @dzenanz I'm sorry but C++ programmers nowadays just _have_ to be familiar with the C++11 (2011) `{}` syntax.

Actually I would like to express: "for each `size` less than `SpaceDimension`, having the same type as `SpaceDimension`". Unfortunately there does not seem to be a more convenient syntax to say so.

    // Pseudo code:
    for( const auto size : RangeOfNumbersLessThan(SpaceDimension) ) 
    { // ...
    }

N-Dekker: By the way (for what it's worth), generally speaking, initialization by `0` might be more expensive than initialization by an empty `{}`. For example for `vnl_bignum`:

    vnl_bignum x{}; // Calls the fast default-constructor, implemented by C++11 =default.
    vnl_bignum x{0}; // Calls the much more complicated vnl_bignum(T) constructor.

Of course, that's not the issue for `unsigned`. It just shows that initialization by `0` may not always be the best choice,

Leengit: In current ITK code
```bash
egrep -R 'unsigned +int' | wc -l # gives 20709
egrep -R 'unsigned' | egrep -v 'unsigned +(char|int|short|long)' | wc -l # gives 6415
```
It looks like  `unsigned int` is 3-4 times more popular than a simple `unsigned`.

N-Dekker: > It looks like `unsigned int` is 3-4 times as popular as a simple `unsigned`.

Thanks for figuring that out, @Leengit Of course, this could be changed easily by a global `s/unsigned int/unsigned/`  😃 What do you think?

In the past, ITK did use `void` for an empty function parameter list, as in `Update(void)`. That was also changed a while ago. Times are changing. C++ is moving forward as well.

Anyway, I just hope that neither `unsigned` nor the empty `{}` initialization are show stoppers for this pull request!

dzenanz: You have 2 approvals and some suggestions. If you don't want to apply suggestions, go ahead and merge.

N-Dekker: @dzenanz OK, thanks! I just hope I explained well why I chose `unsigned` and the `{}` initializer. Even if I did not convince everyone. I do understand if some of you would like the `for` loop in a different way. Everyone has their preferences, and their good reasons.

It's a bit unfortunate that C++ leaves us with so many choices to make at such a low level of detail, that it's sometimes hard to see the forest from the trees... Anyway, I'll just merge now, thanks for all your feedback!

lassoan: The code must be optimized first and foremost for readability. This usually also results in simple, safe, maintainable code. Sometimes you need to balance between performance vs. simplicity (but this should be done only along code paths pinpointed by profiling). Compared to these, it does not matter whether new C++ features are used or not, or something is convenient for writer of the code (code is written once and read hundreds of times).

In this case of course performance does not matter (there are so much not-particularly-optimized code around it), but you never know how the compiler will optimize things until you actually run the code and profile it. Just for curiosity, I created a very basic test code (see at the bottom) and ran it on https://www.onlinegdb.com/online_c++_compiler. The result was that execution time was not statistically different (for the particular measurement set that I've compared, a=0 was tiny bit faster with mean 2.092508sec, compared to a{} with mean 2.096592). But of course the difference is practically not relevant at all, because even such extreme example resulted in statistically not different computation time.

In summary, you can safely work with the assumption that there is no performance difference between `unsigned int a{}` and `unsigned int a=0`, therefore you can optimize the code for readability without worrying about any performance impact.

Note that most people who read the ITK source code may not not be C++ programmers anymore, but Python programmers who need to look at the ITK source code because some details are not explained in the documentation. They are familiar with basic C/C++ but they are not interested in C++ syntax subtleties. I don't say we need to optimize readability for these people, just that nowadays it is not a realistic expectation for programmers to know C++ syntax.

```
#include <iostream>
#include <chrono>
int main()
{
    for (int repeat=0; repeat<30; ++repeat)
    {
        {
            int sum=0;
            auto start = std::chrono::high_resolution_clock::now();
            for (int i=0; i<2e8; ++i)
            {
                int a{};
                sum += a;
            }
            auto finish = std::chrono::high_resolution_clock::now();
            std::chrono::duration<double> elapsed = finish - start;
            std::cout << "Using {}: " << sum << ", time elapsed: " << elapsed.count() << std::endl;
        }
    
        {
            int sum=0;
            auto start = std::chrono::high_resolution_clock::now();
            for (int i=0; i<2e8; ++i)
            {
                int a=0;
                sum += a;
            }
            auto finish = std::chrono::high_resolution_clock::now();
            std::chrono::duration<double> elapsed = finish - start;
            std::cout << "Using =0: " << sum << ", time elapsed: " << elapsed.count() << std::endl;
        }
    }    
    return 0;
}
```

N-Dekker: Thanks for looking at `{}` versus `=0`, @lassoan I did not expect a remarkable difference for `int`, but in general, for example for the initialization of a `vnl_bignum` or an `Eigen::bfloat16`, `{}` could certainly be significantly faster than `=0`. Which makes sense, because `=0` triggers a call to a converting constructor, that may need some more time.

As C++ programmers, I think we really should get used to the `{}` initialization syntax. It's about time, as it was introduced 10 years ago already!

I did my very best to make the unit test that we're discussing here well readable to anyone who knows C++. I'm sorry if you think it still isn't readable enough. 😢 https://github.com/InsightSoftwareConsortium/ITK/blob/deb0af73c06faac66063ab117c97ab0daaf4c489/Modules/Core/Transform/test/itkTranslationTransformGTest.cxx#L31-L43

But anyway, I'm glad you like the `TranslationTransform::SetParameters` improvement that was actually the main part of this pull request 😃

lassoan: I don't think `a{}`-style initialization should be used for simple types, but I agree that in theory it can make a difference when large amount of complex data types are initialized.

Leengit:
I like the {} initialization syntax because it is a tad more type restrictive in some cases.  And, I am one of these programmers who wants strict type checking because I find that a complaint from a compiler is much more fun than spending hours hunting down some bug that the compiler could have caught.  But I have stepped into the realm of (programming) religion ... ask me why emacs is better than all the other editors if you really want to get me started.  :-)

PR #2498, opened 2021-04-21
clang python wrapping

thewtex:
- BUG: Add vnl_vector_from_array to extras __all__
- COMP: Do not add IPO to Python wrapping if not supported


dzenanz (approved):
Rebase on master to get a working MacCI.

PR #2506, opened 2021-04-23
BUG: Fix SpatialOrientationAdapter::FromDirectionCosines

blowekamp:
The old algorithm could generate an invalid combinations of axes. The new algorithm is greedy by selecting the element in the matrix with the greatest magnitude, and eleminating invalid elements for further selection.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/src/itkSpatialOrientationAdapter.cxx
+++ b/Modules/Core/Common/src/itkSpatialOrientationAdapter.cxx
@@ -32,47 +32,70 @@ namespace itk
 SpatialOrientationAdapter::OrientationType
 SpatialOrientationAdapter::FromDirectionCosines(const DirectionType & Dir)
 {
-  int      axes[9] = { 0, 0, 0, 0, 0, 0, 0, 0, 0 };
-  unsigned dominant_axis;
-
-  dominant_axis = Function::Max3(Dir[0][0], Dir[1][0], Dir[2][0]);
-  axes[dominant_axis] = Function::Sign(Dir[dominant_axis][0]);
-  dominant_axis = Function::Max3(Dir[0][1], Dir[1][1], Dir[2][1]);
-  axes[dominant_axis + 3] = Function::Sign(Dir[dominant_axis][1]);
-  dominant_axis = Function::Max3(Dir[0][2], Dir[1][2], Dir[2][2]);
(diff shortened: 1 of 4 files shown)


Modules/Core/Common/src/itkSpatialOrientationAdapter.cxx
```diff
@@ -32,47 +32,70 @@ namespace itk
 SpatialOrientationAdapter::OrientationType
 SpatialOrientationAdapter::FromDirectionCosines(const DirectionType & Dir)
 {
```
dzenanz, line 34: Commit message: The old algorithm could generate an invalid combinations of axes.

dzenanz (approved):
Thanks for adding the fix here too!

issakomi:
Interesting, just discovered strange part in _itkSpatialOrientationAdapter.cxx_  (not part of the PR)

```
// all terms must be defined, otherwise just punt
  if (terms[0] == SpatialOrientation::ITK_COORDINATE_UNKNOWN ||
      terms[1] == SpatialOrientation::ITK_COORDINATE_UNKNOWN || terms[2] == SpatialOrientation::ITK_COORDINATE_UNKNOWN)
  {
    return SpatialOrientation::ITK_COORDINATE_ORIENTATION_RIP;
  }
```
Why RIP? Why not  _ITK_COORDINATE_ORIENTATION_INVALID_ or may be RAI (fallback to identity matrix).
However it is _not_ a request to change this, the file is very critical, smallest issue may cause huge compatibility problems, personally i prefer to stay away.

blowekamp:
> Why RIP? Why not _ITK_COORDINATE_ORIENTATION_INVALID_ or may be RAI (fallback to identity matrix).
> However it is _not_ a request to change this, the file is very critical, smallest issue may cause huge compatibility problems, personally i prefer to stay away.

Yes, I think the intention was to fallback to the default with an identity matrix. For the failure case, that was added in the test, I two axis were given the same label. So it was a failure case not caught by the fallback. This could be changed to verify that the results a valid, but I don't see an easy way to test for the failure case in the bug.

However, with the new algorithm I do not see how unknown, or an invalid combination could be produced. Even with an invalid matrix of all zeros, it should produce a valid results.  So these lines might be able to be removed?

Issue #2510, opened 2021-04-26
ENH: StatisticsLabelMapFilter should add support for a lower-memory median calculation


Leengit:
### Description

Pull Request #2503 (for Issue #2482) fixes a bug where the calculated median was quite wrong in some situations.  The precise calculation that is now used can use as much additional memory as the input image (if most pixels have the same non-zero label) and sometimes this input image could be large.  This is a request that an option be added, e.g. `SetConserveMemory`, for the `itk::StatisticsLabelMapFilter` class (and, as appropriate, for its subclasses) and a supporting implementation be created that computes the precise median more efficiently memory-wise (though less efficiently time-wise).

Issue #2482, opened 2021-04-14
itk.LabelImageToStatisticsLabelMapFilter issue with median measurements in 16bit images in Python


jluethi:
### Description

Dear ITK Team
Thanks a lot for making this great library accessible from Python!
We're trying to use ITK in a 3D image processing workflow and use it to make measurements on single cells. For that purpose, one of the things we use is `itk.LabelImageToStatisticsLabelMapFilter` to make the measurements.
When using this with label images and 16bit intensity images, the median measurements have an unexpected behavior. Typically, most of our intensity values are < 1000, but our images sometimes contain some very bright pixels (either in the area to be measured or outside of it. It appears that the GetMedian measurement in this function is doing an 8bit rescaling of the 16bit image, thus being very sensitive to bright/hot pixels.
Is this an intended default? It's very unexpected and makes median measurements unusable for us, as they're just wrong now for the values we're interested in.
I'm providing a dummy-case below to show this issue. In that dummy case, I change 1 pixel value outside the region that is measured to a very high value and compare the measurements.

### Steps to Reproduce

```
# Python test case
import numpy as np
import itk
import copy

def calculate_features(lbl_img, int_img):
    filt = itk.LabelImageToStatisticsLabelMapFilter[
                type(lbl_img), type(int_img), itk.LabelMap.x3
            ].New()
    filt.SetInput1(lbl_img)
    filt.SetInput2(int_img)
    filt.SetComputeHistogram(False)
    filt.SetComputeFeretDiameter(True)
    filt.SetComputePerimeter(True)
    filt.Update()
    features = filt.GetOutput()
    df2 = dict()
    df2['mean'] = np.array([features[l].GetMean() for l in features.GetLabels()])
    df2['median'] = np.array([features[l].GetMedian() for l in features.GetLabels()])
    print('Means: {}'.format(df2['mean']))
    print('Medians: {}'.format(df2['median']))
    #return df2

shape = (50, 50, 1)
# Define a dummy label image
lbl_img_np = np.zeros(shape).astype('uint16')
lbl_img_np[5:10, 5:10, 0] = 1
lbl_img_np[15:20, 5:10, 0] = 2
lbl_img_np[25:30, 5:10, 0] = 3
lbl_img_np[5:10, 15:20, 0] = 4
lbl_img_np[15:20, 15:20, 0] = 5
lbl_img_np[25:30, 15:20, 0] = 6

# Define a dummy intensity image
[...]

PR #2514, opened 2021-04-26
Test and improve StatisticsLabelMapFilter's median computation

blowekamp:
## PR Checklist

### DIFF
--- a/Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.h
+++ b/Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.h
@@ -147,10 +147,15 @@ protected:
   PrintSelf(std::ostream & os, Indent indent) const override;
 
 private:
-  FeatureImagePixelType m_Minimum;
-  FeatureImagePixelType m_Maximum;
-  unsigned int          m_NumberOfBins;
-  bool                  m_ComputeHistogram;
+  FeatureImagePixelType m_Minimum{ NumericTraits<FeatureImagePixelType>::ZeroValue() };
+  FeatureImagePixelType m_Maximum{ NumericTraits<FeatureImagePixelType>::ZeroValue() };
+  // Set the number of bins to match the number of values for 8,16-bit integers.
+  static constexpr bool m_DefaultNumberOfBinsToNumberOfValues =
+    NumericTraits<typename Self::FeatureImagePixelType>::IsInteger && sizeof(typename Self::FeatureImagePixelType) <= 2;
+  unsigned int m_NumberOfBins{ m_DefaultNumberOfBinsToNumberOfValues
+                                 ? 1 << (8 * sizeof(typename Self::FeatureImagePixelType))
+                                 : 128 };
+  bool         m_ComputeHistogram{ true };
 }; // end of class
 } // end namespace itk
 
--- a/Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.hxx
+++ b/Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.hxx
@@ -177,8 +177,8 @@ StatisticsLabelMapFilter<TImage, TFeatureImage>::ThreadedProcessLabelObject(Labe
   }
 
   // the median
-  double median = 0;
-  double count = 0; // will not be fully set, so do not use later !
+  double median = 0.0;
+  double count = 0.0; // will not be fully set, so do not use later !
   for (SizeValueType i = 0; i < histogram->Size(); ++i)
   {
     count += histogram->GetFrequency(i);
@@ -186,6 +186,19 @@ StatisticsLabelMapFilter<TImage, TFeatureImage>::ThreadedProcessLabelObject(Labe
     if (count >= (totalFreq / 2))
     {
       median = histogram->GetMeasurementVector(i)[0];
+      // If there are an even number of elements average with the next bin with elements
+      if (labelObject->Size() % 2 == 0 && count == totalFreq / 2)
+      {
+        while (++i < histogram->Size())
+        {
+          if (histogram->GetFrequency(i) > 0)
+          {
+            median += histogram->GetMeasurementVector(i)[0];
+            median *= 0.5;
+            break;
+          }
+        }
+      }
       break;
     }
   }
--- a/Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.hxx
+++ b/Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.hxx
@@ -31,10 +31,6 @@ namespace itk
 template <typename TImage, typename TFeatureImage>
 StatisticsLabelMapFilter<TImage, TFeatureImage>::StatisticsLabelMapFilter()
 {
-  m_Minimum = NumericTraits<FeatureImagePixelType>::ZeroValue();
-  m_Maximum = NumericTraits<FeatureImagePixelType>::ZeroValue();
-  m_NumberOfBins = 128;
-  m_ComputeHistogram = true;
   this->SetNumberOfRequiredInputs(2);
 }
 
@@ -73,10 +69,22 @@ StatisticsLabelMapFilter<TImage, TFeatureImage>::ThreadedProcessLabelObject(Labe
   histogramSize.Fill(m_NumberOfBins);
 
   typename HistogramType::MeasurementVectorType featureImageMin(1);
-  featureImageMin.Fill(m_Minimum);
 
   typename HistogramType::MeasurementVectorType featureImageMax(1);
-  featureImageMax.Fill(m_Maximum);
+
+
+  if (NumericTraits<typename Self::FeatureImagePixelType>::IsInteger &&
+      m_NumberOfBins == 1 << (8 * sizeof(typename Self::FeatureImagePixelType)))
+  {
+    // Add padding so the center of bins are integers
+    featureImageMin.Fill(NumericTraits<typename Self::FeatureImagePixelType>::min() - 0.5);
+    featureImageMax.Fill(NumericTraits<typename Self::FeatureImagePixelType>::max() + 0.5);
+  }
+  else
+  {
+    featureImageMin.Fill(m_Minimum);
+    featureImageMax.Fill(m_Maximum);
+  }
 
   typename HistogramType::Pointer histogram = HistogramType::New();
   histogram->SetMeasurementVectorSize(1);
@@ -183,7 +191,7 @@ StatisticsLabelMapFilter<TImage, TFeatureImage>::ThreadedProcessLabelObject(Labe
   {
     count += histogram->GetFrequency(i);
 
-    if (count >= (totalFreq / 2))
+    if (count >= ((totalFreq + 1) / 2))
     {
       median = histogram->GetMeasurementVector(i)[0];
       // If there are an even number of elements average with the next bin with elements
--- a/Modules/Filtering/LabelMap/test/itkStatisticsLabelMapFilterGTest.cxx
+++ b/Modules/Filtering/LabelMap/test/itkStatisticsLabelMapFilterGTest.cxx
@@ -1,0 +1,251 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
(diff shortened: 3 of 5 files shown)


Leengit (commented):
This is great! -- thank you.  Does this test pass?

If it does then I am confused as to why this test passes but the example from Issue #2482 gives results that are much less precise.  Any thoughts on why that might be?  Differences between the two approaches that I notice are:

1. That issue has multiple labels, not just a single one.
2. That issue uses a 5 × 5 region for each label rather than 25 × 25, which should be noisier in the sense that the median is likely to deviate more from the mean.
3. Some of the tests here have more symmetry than the Issue #2482 test has.  For example one of the tests here has all pixel values equal to 0 except one each at 32000 and −32000; a bug that is manifest only when things are less symmetric might escape notice.

blowekamp:
> 2\. That issue uses a 5 × 5 region for each label rather than 25 × 25, which should be noisier in the sense that the median is likely to deviate more from the mean.

I am working on a couple more tweaks. There is no special handling for there there is an even number of elements. So, if there is a large gap in the middle the upper element is returned. Perhaps, that is occurring?

Modules/Filtering/LabelMap/test/itkStatisticsLabelMapFilterGTest.cxx
```diff
@@ -0,0 +1,251 @@
...
+      {
+        it.Set(rand() % randMax);
+        ++it;
+      }
+      return image;
+    }
+
+
+    static typename LabelImageType::Pointer
+    CreateLabelImage()
```
dzenanz, line 97: Why not make CreateImage templated, and pass image type?

blowekamp:
@Leengit I have added my improvements and bug fixes to the histogram based median calculation. One of your proposed exact median computation methods should still be added as an a flag enable alternate computation. As a number of labeled pixels get small your postposed computation method at some point be faster than the overhead of all the elements in the histogram. So it is a worth wild option for those dealing with small labels regions.

Additionally the LabelStatisticsImageFilter's median computation from a histogram has the same issues we have discovered here.

Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.hxx
```diff
@@ -177,15 +177,27 @@ StatisticsLabelMapFilter<TImage, TFeatureImage>::ThreadedProcessLabelObject(Labe
...
       median = histogram->GetMeasurementVector(i)[0];
+      // If there are an even number of elements average with the next bin with elements
+      if (labelObject->Size() % 2 == 0 && count == totalFreq / 2)
+      {
+        while (++i < histogram->Size())
+        {
+          if (histogram->GetFrequency(i) > 0)
+          {
+            median += histogram->GetMeasurementVector(i)[0];
+            median *= 0.5;
```
dzenanz, line 205: shouldn't there be `break;` here?

dzenanz (approved):

Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.hxx
blowekamp: good catch! Thanks!

Modules/Filtering/LabelMap/test/itkStatisticsLabelMapFilterGTest.cxx
blowekamp: The outer support class FixtureUtilities is templated so that they can share the same typedefs.

Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.hxx
blowekamp: I changed the test case to have 4 elements to cover this case.

dzenanz (approved):

Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.hxx
```diff
@@ -177,15 +185,28 @@ StatisticsLabelMapFilter<TImage, TFeatureImage>::ThreadedProcessLabelObject(Labe
...
   for (SizeValueType i = 0; i < histogram->Size(); ++i)
   {
     count += histogram->GetFrequency(i);
 
-    if (count >= (totalFreq / 2))
+    if (count >= ((totalFreq + 1) / 2))
     {
       median = histogram->GetMeasurementVector(i)[0];
+      // If there are an even number of elements average with the next bin with elements
+      if (labelObject->Size() % 2 == 0 && count == totalFreq / 2)
```
Leengit, line 198: Are `labelObject->Size()` and `totalFreq` equal?  If so, why are both forms present here?  Is it that they are different types?

Leengit (changes requested):

Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.h
```diff
@@ -147,10 +147,15 @@ class ITK_TEMPLATE_EXPORT StatisticsLabelMapFilter
...
-  FeatureImagePixelType m_Maximum;
-  unsigned int          m_NumberOfBins;
-  bool                  m_ComputeHistogram;
+  FeatureImagePixelType m_Minimum{ NumericTraits<FeatureImagePixelType>::ZeroValue() };
+  FeatureImagePixelType m_Maximum{ NumericTraits<FeatureImagePixelType>::ZeroValue() };
+  // Set the number of bins to match the number of values for 8,16-bit integers.
+  static constexpr bool m_DefaultNumberOfBinsToNumberOfValues =
+    NumericTraits<typename Self::FeatureImagePixelType>::IsInteger && sizeof(typename Self::FeatureImagePixelType) <= 2;
+  unsigned int m_NumberOfBins{ m_DefaultNumberOfBinsToNumberOfValues
+                                 ? 1 << (8 * sizeof(typename Self::FeatureImagePixelType))
```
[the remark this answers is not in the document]
blowekamp: Correct, which is why there is the additional condition `&&  sizeof(typename Self::FeatureImagePixelType) <= 2`.

The type of `m_NumberOfBins` likely should be corrected to match the histogram size type.

Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.hxx
```diff
@@ -177,15 +185,28 @@ StatisticsLabelMapFilter<TImage, TFeatureImage>::ThreadedProcessLabelObject(Labe
...
+      {
+        while (++i < histogram->Size())
+        {
+          if (histogram->GetFrequency(i) > 0)
+          {
+            median += histogram->GetMeasurementVector(i)[0];
+            median *= 0.5;
```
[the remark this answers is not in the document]
Leengit: I'm off by one, aren't I?  Sorry about that.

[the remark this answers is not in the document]
blowekamp: > I can't tell whether `totalFreq` is an integer or floating point type. If it is integer then in the case that there are 4 values total, we want to return the average of the values and at indexes 1 and 2; but it looks like we are first seeking the value at index 2 and then looking for the value after that.

`totalFreq` is an integer type.

The difference between the `index` and `count` is confusing and tripped this code up a bit. The thought case is covered by the `2D_even` test case and the correct results are returned.

> If totalFreq is a floating point type then, in the case of 4 values, we are first seeking the value for index 3, but we want to first seek for the value for index 1.

This is incorrect.

```
bin:     0 1 2 3 
count: 1 1 1 1
```
At bin/index 1 the count is 2. The first bin used for the median would be bin 1. 

Again this case is covered by a test.

Modules/Filtering/LabelMap/test/itkStatisticsLabelMapFilterGTest.cxx
```diff
@@ -0,0 +1,321 @@
...
+      return image;
+    }
+
+    static typename ImageType::Pointer
+    CreateImageRandom(PixelType randMax = 500, unsigned int randSeed = 0)
+    {
+      typename ImageType::Pointer image = ImageType::New();
+
+      typename ImageType::SizeType imageSize;
+      imageSize.Fill(25);
```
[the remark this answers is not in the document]
blowekamp: This is a good point. I addressed it by added a couple test cases with just 3 and for elements.

Modules/Filtering/LabelMap/test/itkStatisticsLabelMapFilterGTest.cxx
```diff
@@ -0,0 +1,321 @@
...
+      while (!it.IsAtEnd())
+      {
+        const typename ImageType::IndexType & idx = it.GetIndex();
+        values.push_back(image->GetPixel(idx));
+        ++it;
+      }
+
+      std::sort(values.begin(), values.end());
+
+      auto n1 = values.size() / 2;
```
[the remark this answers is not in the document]
blowekamp: I added an assert and some GTest ASSERT to ensure the size is not zero.

blowekamp:
@Leengit Thank you for the in-depth look at the code. I think I have resolved most of the points you brought up. Except the readability of multiple variable.

As earlier noted, there are other placed in ITK which compute a histogram based median. They contain similar computational inaccuracy or bugs. The next step after this PR is merged to to move that computation into a function in `itk::Statistics::Algorithms` and reuse it in filters such at the LabelStatisticsImageFilter.

Leengit (approved):
@blowekamp -- thanks for all the effort here!  Looks good.

Leengit:
@jluethi if building current ITK from source is something you can do easily enough, I would appreciate your feedback on whether this pull request solves your problem.  For pixels that are `unsigned short`, you would want to set the `number_of_bins` to 65536 or simply leave it as default.  If this implementation doesn't meet your needs, we can address them in a subsequent pull request.

jluethi:
@Leengit Are there instructions on how to build ITK from source somewhere? I'm not too familiar with C++, but happy follow instructions. I assume I will need to do more here than download the repository and pip install it, right?
My main question would be: How does it perform on my test case from https://github.com/InsightSoftwareConsortium/ITK/issues/2482? What values do I get compared to the true medians on my 16 bit images?

dzenanz:
General build instructions: https://itk.org/Wiki/ITK/Configuring_and_Building
Python-specific instructions: https://itk.org/Wiki/ITK/Python_Wrapping

Leengit:
My testing of @jluethi's test code is that it now computes the desired medians if one explicitly uses `filt.SetNumberOfBins(65536)`.  However for a reason having to do with base classes vs. subclasses, the change of the default value for `NumberOfBins`, which is also to 65536 in this case, doesn't happen to apply for this test code.  I am testing a quick fix for that and will submit a pull request pronto.  That should obviate the need for an explicit `filt.SetNumberOfBins(65536)` call.

jluethi:
I lost the overview a bit over which solution was chosen now and how well it performs. @Leengit Do I understand correctly that with what has been merged into master, using `filt.SetNumberOfBins(65536)` solves the issue I described above? But at the moment, this needs to be set explicitly and discussions about how to have median calculations exact by default are still ongoing?
Would this mean I can try and get a build based on the current master running & set `filt.SetNumberOfBins(65536)` => then that use case should be covered? And at some point, this may not be necessary anymore, once a more general solution has been found?

Leengit:
Yes, using what has already been merged into master: a 16-bit pixel type and `filt.SetNumberOfBins(65536)` will get an exact median calculation via a histogram.

As soon as Pull Request #2521 is merged into ITK, you won't have to call `filt.SetNumberOfBins(65536)`.  More generally, the default behavior when the pixel type is 8 or 16 bits in size will then be an exact median via a histogram.

At that point, it is only for types larger than 16 bits that the median calculation could show behavior similar to what you describe in Issue #2482.  To get an exact median calculation with these larger pixel types, there is a meaningful tradeoff between using the least amount of memory and using the least amount of time.  Issue #2510 asks that these two additional implementations be added and made optionally available to the user.

Leengit:
Pull Request #2521 is now merged.

PR #2530, opened 2021-05-13
Fix HDF5 installed find package

blowekamp:
Addresses instalation issue with ITK_USE_SYSTEM_HDF5 with a system HDF5 that does not provide a HDF5Config.cmake file. The wrong find_package signature was being generated in the ITKHDF5.cmake file.

This change is also needed in the release branch.


jhlegarreta (approved):

dzenanz (approved):

dzenanz:
Let's wait to see the nightly tests before making a release PR.

PR #2327, opened 2021-02-18
ENH: Increase coverage

jhlegarreta:
Improvements concerning testing and a few related enhancements:
- Fix an RTTI bug concerning a `Superclass` name.
- Remove a duplicate include file.
- Print all member variables in a number of classes.
- Use `itkBooleanMacro` to set the boolean ivar values to On/Off.
- Increase code coverage of a number of classes.

## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [X] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [X] Added test (or behavior not changed)
- [X] Updated API documentation (or API not changed)
- [X] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)

PR #2534, opened 2021-05-15
STYLE: Make test style more consistent

jhlegarreta:
Make test style more consistent:
- Use the ITK testing macros to reduce boilerplate code.
- Increase consistency in the missing argument check message and test
  ending messages.
- Define the constant expressions (e.g. the image dimensionality) at a
  single place at the beginning of the tests.
- Use the ITK C++ coding style (e.g. do not capitalize first letter in
  local variable names, etc.).

Take advantage of the commit to fix some typos.

## PR Checklist

### DIFF
--- a/Modules/Filtering/ImageGradient/test/itkVectorGradientMagnitudeImageFilterTest1.cxx
+++ b/Modules/Filtering/ImageGradient/test/itkVectorGradientMagnitudeImageFilterTest1.cxx
@@ -39,20 +39,22 @@ itkVectorGradientMagnitudeImageFilterTest1(int argc, char * argv[])
     return EXIT_FAILURE;
   }
 
+  constexpr unsigned int Dimension = 2;
+
   using RGBPixelType = itk::RGBPixel<unsigned short>;
-  using CharImageType = itk::Image<unsigned char, 2>;
-  using RGBImageType = itk::Image<RGBPixelType, 2>;
-  using AdaptorType = itk::RGBToVectorImageAdaptor<RGBImageType>;
-  using FilterType = itk::VectorGradientMagnitudeImageFilter<AdaptorType>;
-  using ReaderType = itk::ImageFileReader<RGBImageType>;
-  using RescaleFilterType = itk::RescaleIntensityImageFilter<FilterType::OutputImageType, CharImageType>;
-  using WriterType = itk::ImageFileWriter<CharImageType>;
+  using CharImageType = itk::Image<unsigned char, Dimension>;
+  using RGBImageType = itk::Image<RGBPixelType, Dimension>;
 
   // Create a reader and filter
+  using ReaderType = itk::ImageFileReader<RGBImageType>;
   ReaderType::Pointer reader = ReaderType::New();
   reader->SetFileName(argv[1]);
+
+  using AdaptorType = itk::RGBToVectorImageAdaptor<RGBImageType>;
   AdaptorType::Pointer adaptor = AdaptorType::New();
   adaptor->SetImage(reader->GetOutput());
+
+  using FilterType = itk::VectorGradientMagnitudeImageFilter<AdaptorType>;
   FilterType::Pointer filter = FilterType::New();
 
   ITK_EXERCISE_BASIC_OBJECT_METHODS(filter, VectorGradientMagnitudeImageFilter, ImageToImageFilter);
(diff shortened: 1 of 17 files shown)


jhlegarreta:
Note that this change does not mean to be exhaustive at all. I just came across these cases while I was working on #2327. Thanks.

Modules/Filtering/ImageGradient/test/itkVectorGradientMagnitudeImageFilterTest1.cxx
```diff
@@ -27,7 +27,7 @@
 int
 itkVectorGradientMagnitudeImageFilterTest1(int argc, char * argv[])
 {
-  if (argc != 7)
+  if (argc < 7)
   {
     std::cerr << "Missing Parameters." << std::endl;
```
Leengit, line 32: I don't know the answer to this question ... is there an ITK style for using `argc != 7` vs. `argc < 7`.  Since I don't know any reason why `argc > 7` would be desirable in this case, I'd think that an error would be appropriate in that case too.

Modules/Filtering/ImageGradient/test/itkVectorGradientMagnitudeImageFilterTest1.cxx
```diff
@@ -39,20 +39,22 @@ itkVectorGradientMagnitudeImageFilterTest1(int argc, char * argv[])
...
-  using AdaptorType = itk::RGBToVectorImageAdaptor<RGBImageType>;
-  using FilterType = itk::VectorGradientMagnitudeImageFilter<AdaptorType>;
-  using ReaderType = itk::ImageFileReader<RGBImageType>;
-  using RescaleFilterType = itk::RescaleIntensityImageFilter<FilterType::OutputImageType, CharImageType>;
-  using WriterType = itk::ImageFileWriter<CharImageType>;
+  using CharImageType = itk::Image<unsigned char, Dimension>;
+  using RGBImageType = itk::Image<RGBPixelType, Dimension>;
 
   // Create a reader and filter
+  using ReaderType = itk::ImageFileReader<RGBImageType>;
```
Leengit, line 49: I don't know the answer to this question ... is there an ITK style for the placement of these `using` statements?  Should they all be near the top of the scope or should they be delayed until their first use?  I've generally been doing the former, which is the opposite of the change here.

Modules/IO/MINC/test/itkMINCImageIOTest_2D.cxx
```diff
@@ -122,9 +123,8 @@ itkMINCImageIOTest_2D(int argc, char * argv[])
     }
     else
     {
-      std::cerr << "Incorrecte number of additional arguments " << std::endl;
-      std::cerr << "Usage: " << std::endl;
-      std::cerr << argv[0] << " inputfile outputfile [sum mx my ]" << std::endl;
+      std::cerr << "Incorrect number of additional arguments " << std::endl;
```
Leengit, line 126: Elsewhere you chnaged "argument" to "parameter".

dzenanz (approved):

Modules/Filtering/ImageGradient/test/itkVectorGradientMagnitudeImageFilterTest1.cxx
jhlegarreta: Thanks Lee. Changed back to `!=` in 79c65d7. You're right as per the ITK SW Guide (https://itk.org/ItkSoftwareGuide.pdf; section C.24, p. 370: C.24). It might have been myself that wrote that part, but at times I am not consistent myself wither. Sorry.

I do not mean to change all tests in this PR to that match convention as I am not meaning to be exhaustive. That can be done in a separate PR.

Modules/IO/MINC/test/itkMINCImageIOTest_2D.cxx
jhlegarreta: Changed in 79c65d7.

Modules/Filtering/ImageGradient/test/itkVectorGradientMagnitudeImageFilterTest1.cxx
jhlegarreta: The ITK SW Guide (https://itk.org/ItkSoftwareGuide.pdf section C.23.1, p. 368) contains the following passage:
```
Note that constant declarations (e.g. image dimensions, etc.) and type alias declarations (e.g. pixel
and image types, etc.) should be local to where they are used for the sake of readability.
```

Maybe it can be improved. Generally, I'd say that we prefer declaring aliases close to where they are used, except for those ones that are used extensively (e.g. the image types) in a test.

PR #2536, opened 2021-05-15
ENH: Increase `Common` operators' code coverage

jhlegarreta:
Increase `Common` operator classes' code coverage:
- Remove explicit calls to the Print method and rely on the basic
  method exercising macro call.
- Improve slightly the test style to make them more consistent (e.g.
  input argument checking, test finishing message, make the types and
  variables dwell closer to the place where they are instantiated, etc.)

## PR Checklist

### DIFF
--- a/Modules/Core/Common/test/itkAnnulusOperatorTest.cxx
+++ b/Modules/Core/Common/test/itkAnnulusOperatorTest.cxx
@@ -18,6 +18,7 @@
 
 #include "itkAnnulusOperator.h"
 #include "itkStdStreamStateSave.h"
+#include "itkTestingMacros.h"
 
 int
 itkAnnulusOperatorTest(int, char *[])
@@ -31,12 +32,18 @@ itkAnnulusOperatorTest(int, char *[])
   using PixelType = float;
   using OperatorType = itk::AnnulusOperator<PixelType, Dimension>;
 
-
   OperatorType normalizedAnnulus;
+
+  ITK_EXERCISE_BASIC_OBJECT_METHODS((&normalizedAnnulus), AnnulusOperator, NeighborhoodOperator);
+
+
   normalizedAnnulus.NormalizeOn();
   normalizedAnnulus.SetInnerRadius(3);
   normalizedAnnulus.SetThickness(2);
-  normalizedAnnulus.BrightCenterOff();
+
+  const bool brightCenter = false;
+  ITK_TEST_SET_GET_BOOLEAN((&normalizedAnnulus), BrightCenter, brightCenter);
+
   try
   {
     normalizedAnnulus.CreateOperator();
@@ -56,10 +63,6 @@ itkAnnulusOperatorTest(int, char *[])
     std::cout << "Unknown exception" << std::endl;
     return EXIT_FAILURE;
   }
-
-  std::cout << "Normalized annulus" << std::endl;
-  normalizedAnnulus.Print(std::cout);
-  std::cout << std::endl << std::endl;
 
   OperatorType::SizeType normalizedAnnulusSize;
   normalizedAnnulusSize = normalizedAnnulus.GetSize();
@@ -104,12 +107,30 @@ itkAnnulusOperatorTest(int, char *[])
   std::cout << std::endl << std::endl;
   std::cout << "Creating an annulus by specifying values";
   OperatorType annulus;
-  annulus.NormalizeOff();
-  annulus.SetInnerRadius(2);
-  annulus.SetThickness(1);
-  annulus.SetExteriorValue(1);
-  annulus.SetAnnulusValue(8);
-  annulus.SetInteriorValue(4);
+
+  const bool normalize = false;
+  ITK_TEST_SET_GET_BOOLEAN((&annulus), Normalize, normalize);
+
+  const double innerRadius = 2;
+  annulus.SetInnerRadius(innerRadius);
+  ITK_TEST_SET_GET_VALUE(innerRadius, annulus.GetInnerRadius());
+
+  const double thickness = 1;
+  annulus.SetThickness(thickness);
+  ITK_TEST_SET_GET_VALUE(thickness, annulus.GetThickness());
+
+  const typename OperatorType::PixelType exteriorValue = 1;
+  annulus.SetExteriorValue(exteriorValue);
+  ITK_TEST_SET_GET_VALUE(exteriorValue, annulus.GetExteriorValue());
+
+  const typename OperatorType::PixelType annulusValue = 8;
+  annulus.SetAnnulusValue(annulusValue);
+  ITK_TEST_SET_GET_VALUE(annulusValue, annulus.GetAnnulusValue());
+
+  const typename OperatorType::PixelType interiorValue = 4;
+  annulus.SetInteriorValue(interiorValue);
+  ITK_TEST_SET_GET_VALUE(interiorValue, annulus.GetInteriorValue());
+
   annulus.CreateOperator();
 
   OperatorType::SizeType annulusSize;
@@ -182,9 +203,13 @@ itkAnnulusOperatorTest(int, char *[])
   OperatorType::SpacingType spacing;
   spacing[0] = 0.5;
   spacing[1] = 0.25;
+
+  annulus.SetSpacing(spacing);
+  ITK_TEST_SET_GET_VALUE(spacing, annulus.GetSpacing());
+
   annulus.SetInnerRadius(2);
   annulus.SetThickness(1);
-  annulus.SetSpacing(spacing);
+
   annulus.CreateOperator();
 
   annulusSize = annulus.GetSize();
@@ -203,5 +228,7 @@ itkAnnulusOperatorTest(int, char *[])
     std::cout << std::endl;
   }
 
+
+  std::cout << "Test finished." << std::endl;
   return EXIT_SUCCESS;
 }
--- a/Modules/Core/Common/test/itkDerivativeOperatorTest.cxx
+++ b/Modules/Core/Common/test/itkDerivativeOperatorTest.cxx
@@ -19,6 +19,7 @@
 
 #include "itkMath.h"
 #include "itkDerivativeOperator.h"
+#include "itkTestingMacros.h"
 
 namespace itk
 {
@@ -63,33 +64,31 @@ itkDerivativeOperatorTest(int, char *[])
   constexpr unsigned int Dimension = 1;
   using PixelType = float;
 
+  // Exercise object basic methods
+  using DerivativeOperatorType = itk::DerivativeOperator<PixelType, Dimension>;
+
+  DerivativeOperatorType derivativeOperator;
+
+  ITK_EXERCISE_BASIC_OBJECT_METHODS((&derivativeOperator), DerivativeOperator, NeighborhoodOperator);
+
+
+  // Test other methods using the helper
   using OperatorType = itk::DerivativeOperatorTestHelper<PixelType, Dimension>;
 
   OperatorType op1;
 
-  op1.SetOrder(1);
-  if (op1.GetOrder() != 1)
-  {
-    std::cerr << "Set/GetOrder() failed" << std::endl;
-    return EXIT_FAILURE;
-  }
+  unsigned int order = 1;
+  op1.SetOrder(order);
+  ITK_TEST_SET_GET_VALUE(order, op1.GetOrder());
 
-  op1.SetOrder(2);
-  if (op1.GetOrder() != 2)
-  {
-    std::cerr << "Set/GetOrder() failed" << std::endl;
-    return EXIT_FAILURE;
-  }
+  order = 2;
+  op1.SetOrder(order);
+  ITK_TEST_SET_GET_VALUE(order, op1.GetOrder());
 
-  op1.SetOrder(1);
-  if (op1.GetOrder() != 1)
-  {
-    std::cerr << "Set/GetOrder() failed" << std::endl;
-    return EXIT_FAILURE;
-  }
+  order = 1;
+  op1.SetOrder(order);
+  ITK_TEST_SET_GET_VALUE(order, op1.GetOrder());
 
-  // Exercise PrintSelf
-  op1.Print(std::cout);
 
   OperatorType::CoefficientVector expected1;
   expected1.push_back(0.5);
@@ -99,6 +98,8 @@ itkDerivativeOperatorTest(int, char *[])
   // Check actual coefficient values
   if (!op1.CheckCoefficients(expected1))
   {
+    std::cerr << "Test failed!" << std::endl;
+    std::cerr << "Error in coefficients" << std::endl;
     return EXIT_FAILURE;
   }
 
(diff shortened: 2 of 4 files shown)


jhlegarreta:
This PR will be accompanied by several separate PRs where I'll be adding RTTI to the operators that are missing it, using the Set/Get macros to set/get the ivar values, etc.

jhlegarreta:
Not sure about this compiler failure:
```
Modules/Core/Common/test/itkGaussianDerivativeOperatorTest.cxx:47:36:
error: 'using GaussianOp = class itk::GaussianDerivativeOperator<double, 1u>
{aka class itk::GaussianDerivativeOperator<double, 1u>}' has no member named 'GetMaximumKernelWidth';
did you mean 'SetMaximumKernelWidth'?

   ITK_TEST_SET_GET_VALUE(width, op.GetMaximumKernelWidth());
                                                             ^
```

since the method exists in https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkGaussianOperator.h#L129

Modules/Core/Common/test/itkGaussianDerivativeOperatorTest.cxx
```diff
@@ -159,5 +177,13 @@ itkGaussianDerivativeOperatorTest(int argc, char * argv[])
...
-  return EXIT_FAILURE;
+  std::cout << "Test finished." << std::endl;
+  if (testStatus)
+  {
+    return EXIT_SUCCESS;
+  }
+  else
+  {
+    return EXIT_FAILURE;
+  }
```
Leengit, line 188: `return testStatus ? EXIT_SUCCESS : EXIT_FAILURE;`

Modules/Core/Common/test/itkNeighborhoodOperatorTest.cxx
```diff
@@ -77,45 +78,82 @@ println(const char * c)
 int
 itkNeighborhoodOperatorTest(int, char *[])
 {
-  println("Testing derivative operator");
+  std::cout << "Testing derivative operator" << std::endl;
   itk::DerivativeOperator<float, 3, vnl_vector<float>> d;
```
Leengit, line 82:
```c++
using Real = float;
constexpr unsigned int Dimension = 3;
itk::DerivativeOperator<Real, Dimension, vnl_vector<Real>> d;
```
... and use `Real` and/or `Dimension` in some line below as well.  (Or skip my `Dimension` suggestion because other values for Dimension are used later.)

Modules/Core/Common/test/itkNeighborhoodOperatorTest.cxx
```diff
@@ -77,45 +78,82 @@ println(const char * c)
...
+  direction = 0;
+  b.SetDirection(direction);
+  ITK_TEST_SET_GET_VALUE(direction, b.GetDirection());
+
   b.CreateDirectional();
-  b.Print(std::cout);
 
-  unsigned int i;
 
   double scales[3] = { 1.0 / 2.0, 1.0 / 4.0, 1.0 / 5.0 };
```
Leengit, line 142: `const double`

Leengit (commented):
Looks good.

Modules/Core/Common/test/itkNeighborhoodOperatorTest.cxx
```diff
@@ -77,45 +78,82 @@ println(const char * c)
 int
```
dzenanz, line 78: Why not remove now unused `println`?

dzenanz (commented):
Looks good

Modules/Core/Common/test/itkAnnulusOperatorTest.cxx
```diff
@@ -31,12 +32,18 @@ itkAnnulusOperatorTest(int, char *[])
...
+
+  ITK_EXERCISE_BASIC_OBJECT_METHODS((&normalizedAnnulus), AnnulusOperator, NeighborhoodOperator);
+
+
   normalizedAnnulus.NormalizeOn();
   normalizedAnnulus.SetInnerRadius(3);
   normalizedAnnulus.SetThickness(2);
-  normalizedAnnulus.BrightCenterOff();
+
+  auto brightCenter = false;
```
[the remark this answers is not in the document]
jhlegarreta: Changed in 8449d1a. Not sure how consistent we've been concerning this, and I'd say that the use of `auto` has gained popularity in the code base lately. Probably being clear about this in the ITK SW Guide would help (the testing section was written a while ago).

Even in the CI builds pass, not sure if the dashboard sites will complain since most of the setter methods do not have the `const` keyword.

I've got a separate branch to be submitted once this one gets merged transitioning to the Set/Get macro methods.

Modules/Core/Common/test/itkAnnulusOperatorTest.cxx
```diff
@@ -104,12 +107,30 @@ itkAnnulusOperatorTest(int, char *[])
...
   std::cout << "Creating an annulus by specifying values";
   OperatorType annulus;
-  annulus.NormalizeOff();
-  annulus.SetInnerRadius(2);
-  annulus.SetThickness(1);
-  annulus.SetExteriorValue(1);
-  annulus.SetAnnulusValue(8);
-  annulus.SetInteriorValue(4);
+
+  auto normalize = false;
```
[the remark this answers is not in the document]
jhlegarreta: Changed in 8449d1a.

Modules/Core/Common/test/itkAnnulusOperatorTest.cxx
```diff
@@ -104,12 +107,30 @@ itkAnnulusOperatorTest(int, char *[])
...
+  ITK_TEST_SET_GET_BOOLEAN((&annulus), Normalize, normalize);
+
+  double innerRadius = 2;
```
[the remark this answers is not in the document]
jhlegarreta: Changed in 8449d1a.

Modules/Core/Common/test/itkGaussianDerivativeOperatorTest.cxx
```diff
@@ -24,38 +24,59 @@ namespace
...
 
   using GaussianOp = itk::GaussianDerivativeOperator<double, 1>;
 
   std::cout << "Testing variance: " << variance << " error: " << error << " width: " << width << " order: " << order
-            << std::endl;
+            << " spacing: " << spacing << std::endl;
 
   GaussianOp op;
 
+  auto normalizeAcrossScale = false;
```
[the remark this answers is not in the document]
jhlegarreta: Changed in 8449d1a.

Modules/Core/Common/test/itkGaussianDerivativeOperatorTest.cxx
```diff
@@ -24,38 +24,59 @@ namespace
...
+  op.CreateDirectional();
 
   double total = std::accumulate(op.Begin(), op.End(), 0.0);
-
   std::cout << "total: " << total << std::endl;
 
-  if (order == 0 && std::abs(total - 1.0) > itk::NumericTraits<double>::epsilon() * 32)
+  std::cout.precision(16);
+
+  double epsilon = itk::NumericTraits<double>::epsilon() * 32;
```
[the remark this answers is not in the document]
jhlegarreta: Changed in 8449d1a.

Modules/Core/Common/test/itkGaussianDerivativeOperatorTest.cxx
```diff
@@ -81,27 +102,31 @@ itkGaussianDerivativeOperatorTest(int argc, char * argv[])
...
+    TestGaussianOperator(variance, error, width, order, spacing);
+
+    std::cout << "Test finished." << std::endl;
+    return EXIT_SUCCESS;
   }
   else if (argc > 1)
   {
-    std::cerr << "Usage: " << itkNameOfTestExecutableMacro(argv) << " [ variance error width order ]" << std::endl;
+    std::cerr << "Missing Parameters." << std::endl;
+    std::cerr << "Usage: " << itkNameOfTestExecutableMacro(argv) << " [ variance error width order spacing]"
```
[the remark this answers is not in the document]
jhlegarreta: Changed in 8449d1a.

Modules/Core/Common/test/itkNeighborhoodOperatorTest.cxx
jhlegarreta: Used an approach inspired by the suggested snippet in 8449d1a.

Modules/Core/Common/test/itkNeighborhoodOperatorTest.cxx
jhlegarreta: Changed in 8449d1a.

Modules/Core/Common/test/itkNeighborhoodOperatorTest.cxx
jhlegarreta: Left it behind inadvertently. Removed in 8449d1a. Thanks.

Modules/Core/Common/test/itkGaussianDerivativeOperatorTest.cxx
jhlegarreta: The ITK SW Guide discourages the use of the ternary operator.

issakomi:
> Not sure about this compiler failure:
> 
> ```
> Modules/Core/Common/test/itkGaussianDerivativeOperatorTest.cxx:47:36:
> error: 'using GaussianOp = class itk::GaussianDerivativeOperator<double, 1u>
> {aka class itk::GaussianDerivativeOperator<double, 1u>}' has no member named 'GetMaximumKernelWidth';
> did you mean 'SetMaximumKernelWidth'?
> 
>    ITK_TEST_SET_GET_VALUE(width, op.GetMaximumKernelWidth());
>                                                              ^
> ```
> 
> since the method exists in https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkGaussianOperator.h#L129

This might be a minor bug, comment in _itkGaussianDerivativeOperator.h_ mentions _Get_ kernel max width, but there is no function, only _Set_.
itk:: GaussianDerivativeOperator doesn't inherit methods from itk::GaussianOperator, BTW.
```
  /** Sets/Get a limit for growth of the kernel.  Small maximum error values with
   *  large variances will yield very large kernel sizes.  This value can be
   *  used to truncate a kernel in such instances.  A warning will be given on
   *  truncation of the kernel. */
  void
  SetMaximumKernelWidth(unsigned int n)
  {
    m_MaximumKernelWidth = n;
  }
```

Leengit:
Thank you for the edits.  Looks good to me, pending figuring out one way or another what is up with the GitHub actions failures.

jhlegarreta:
:facepalm: : https://github.com/InsightSoftwareConsortium/ITK/pull/2536#issuecomment-843134445 Thanks @issakomi !!

Have not looked into it in detail, but `itk::GaussianDerivativeOperator` could inherit also from `itk::DerivativeOperator` , isn't it? Not sure what the policy about multiple inheritance is. But if that should be done, a separate PR is welcome.

PRs #2544 and #2545 should address the failures. Will rebase on `master` once these get merged.

jhlegarreta:
d0d6804 has been rebased on `master`.

jhlegarreta:
Cross-referencing #2547.

Also, this has uncovered the interesting:
```
Modules/Core/Common/test/itkAnnulusOperatorTest.cxx:122:32: error: 'using PixelType = using PixelType = float' is protected within this context
```

and similar errors for the rest of the classes.

All the concerned classes declare a `using PixelType = typename Superclass::PixelType;` inside their `protected` section. However, they receive the pixel type as a template parameter. Hence, am I wrong when saying that we should prefer `using PixelType  = TPixel;` in the `public` section and the use `PixelType` everywhere in their methods?

dzenanz:
> `using PixelType = TPixel;` in the `public` section

Yes, I think this is correct.

By the way, almost each time you make "increase X coverage" PR some inconsistency in the library is discovered 😄

jhlegarreta:
> Yes, I think this is correct.

Will make a PR tonight.

> By the way, almost each time you make "increase X coverage" PR some inconsistency in the library is discovered :smile:

:+1:.

jhlegarreta:
>> Yes, I think this is correct.
> Will make a PR tonight.

PR #2550. Finally left `TPixel` type for the method parameter/return types as I was not sure how consistent are we being on this or whether specific criteria are applicable. I can add an additional commit if `PixelType` is preferred.

jhlegarreta:
Not sure what to think about the failures:
https://open.cdash.org/viewTest.php?onlyfailed&buildid=7251375

Detail:
https://open.cdash.org/test/410774528
https://open.cdash.org/test/410774620

I'd say that since `itk::DerivativeOperator` does not have its RTTI implementation, it's probably calling its parent's, and hence the `Name of Class` found using `GetNameOfClass` is the parent's (?).

If that hypothesis is true, #2556 should fix it.

Modules/Core/Common/test/itkDerivativeOperatorTest.cxx
```diff
@@ -99,6 +98,8 @@ itkDerivativeOperatorTest(int, char *[])
   // Check actual coefficient values
   if (!op1.CheckCoefficients(expected1))
   {
+    std::cerr << "Test failed!" << std::endl;
+    std::cerr << "Error in coefficients" << std::endl;
```
N-Dekker, line 102: Just a small comment here: `std::cerr` does flush automatically with each `<<` insertion. (For details, check: https://en.cppreference.com/w/cpp/io/cerr) This means that explicit flushing of the stream by `<< std::endl` is unnecessary. So in this case, I would suggest:

    std::cerr << "Test failed!\nError in coefficients\n";

The C++ Core Guidelines also suggest to avoid `std::endl`: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rio-endl

jhlegarreta: @N-Dekker thanks for the review. At some point we started preferring `std::endl` over `\n` a few years ago. Cannot remember the reasons, but it dates back to the `Gerrit` days, so I have been unable to find the related comments/discussion.

Also, although at times many of my own changes are not extensive throughout the whole code base, in this case I'm reluctant to change this only for this test: this looks to me like a major style change, involving the SW Guide as well, and virtually all test argument check blocks, and error blocks. Unless a sweeping change is made (and I'm well aware of the effort this involves), I'd sacrifice performance for consistency in this case.

jhlegarreta:
#2556 fixes these: https://github.com/InsightSoftwareConsortium/ITK/pull/2536#issuecomment-850078188

But at the same time, as noted in the comment, it worries me :worried: because failures were not consistent for all classes missing their RTTIs.

Also, @Leengit not sure whether there is room for improvement involving these comments: https://github.com/InsightSoftwareConsortium/ITK/pull/2536#discussion_r633501782
https://github.com/InsightSoftwareConsortium/ITK/pull/2547#issuecomment-845020549
And what I was forced to do because the pointers/variable types were not matching for the `DerivativeScalings`:
https://github.com/InsightSoftwareConsortium/ITK/pull/2536/files#diff-802c29f5c013b90694382e6078ba14f99dd3cae19d83f10dbcef6753d4459457R150

I'll push a separate PR switching to getter/setter macros, but I ignore whether the above is just the result of a poor design for the setter/getters regardless of using the macros or poor variable definition, and hence whether they should be addressed now.

dzenanz (approved):

dzenanz:
itkPyBufferMemoryLeakTest (Failed) is a fluke. Good to merge, I think.

PR #2550, opened 2021-05-20
ENH: Increase pixel type alias consistency in common operators

jhlegarreta:
Increase pixel type alias consistency in common operators:
- Make `PixelType` alias be public in `itk::AnnulusOperator`
- Prefer aliasing template pixel type in `common` operator classes

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkAnnulusOperator.h
+++ b/Modules/Core/Common/include/itkAnnulusOperator.h
@@ -77,6 +77,7 @@ public:
   using Superclass = NeighborhoodOperator<TPixel, TDimension, TAllocator>;
 
   /** Additional type aliases. */
+  using PixelType = typename Superclass::PixelType;
   using SizeType = typename Superclass::SizeType;
   using OffsetType = typename Superclass::OffsetType;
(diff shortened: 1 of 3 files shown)


Leengit (commented):
Is the use of of `typename Superclass::PixelType` in the replaced code supposed to be a way to tell the programmer that these classes inherit their type aliases from the superclass ... kind of paralleling how these classes inherit methods from the superclass?

Does the ITK style guide say anything about this?  If not, should it?

jhlegarreta:
@Leengit Not sure about the intention of using the Superclass type for the pixel, although it is exactly the same as the current classes'. Maybe it was more consistent with the rest of the aliases. I'd say that the SW Guide does not propose or encourage a practice here. If you feel like it should, please do a PR. If you think that the former solution should be used, I'd happy to delete the relevant commit in this PR.

Leengit (approved):
I see value in both the `typename Superclass::PixelType` and `TPixel` approaches.  I have a slight preference for the former, but nothing significant.  Either way ... looks good to me.

dzenanz (approved):

PR #2552, opened 2021-05-25
no global mp lock

thewtex:
- BUG: Do not create global multiprocessing RLock

This addresses:

> /opt/conda/lib/python3.6/multiprocessing/semaphore_tracker.py:143: UserWarning: semaphore_tracker: There appear to be 1 leaked semaphores to clean up at shutdown len(cache))

Reported in Project-MONAI/MONAI#1629

Here we adopt the approach in tqdm:

  tqdm/tqdm#617

to address the similar issues with PyTorch:

  tqdm/tqdm#611

but for our use case. A global set of locks is added to the class on
first use.

We also add a threading lock in addition to a multiprocessing
lock.

### DIFF
--- a/Wrapping/Generators/Python/itk/support/lazy.py
+++ b/Wrapping/Generators/Python/itk/support/lazy.py
@@ -23,18 +23,6 @@ from itk.support import base
 # use of the builtin set
 from builtins import set as _builtin_set
 
-# Need to use a recursive lock for thread ownership
-# within the given thread you can acquire a RLock as often as you like.
-# Other threads need to wait until this thread releases the resource again.
-from multiprocessing import RLock as _mp_RLock
-
-# A single lock is needed for all lazy loading.  This lock blocks
-# across all threads until this thread has completed all its imports
-# and dependancies.  The complex inter-relationship, and the recursive
-# nature of imports, makes a more fine-grained locking very difficult
-# to implement robustly.
-_gbl_lazy_load_lock: _mp_RLock = _mp_RLock()
-
 not_loaded: str = "not loaded"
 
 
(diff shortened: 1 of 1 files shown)


blowekamp:
I'm not sure what to do about the problems caused by mixing process forks and threads that seems to be common in python. Having thread locks indexed by the pid maybe more robust to prevent deadlocks when forking and a thread lock is held.

https://stackoverflow.com/questions/39890363/what-happens-when-a-thread-forks

thewtex:
@blowekamp good point. Since we do not need the thread lock, just the multiprocessing lock to address the pytorch / MONAI use case, I am commenting out the thread lock for now with a note to also investigate a per-pid thread lock.

Wrapping/Generators/Python/itk/support/lazy.py
```diff
@@ -45,6 +33,68 @@ def _lazy_itk_module_reconstructor(module_name, state):
 
     return lazy_module
 
+def _ThreadingRLock(*args, **kwargs):
```
dzenanz, line 36: Perhaps comment out the function, as its invocation is commented out at the moment?

dzenanz:
I modified a MONAI tutorial example to use ITKReader:
https://github.com/dzenanz/MONAI-tutorials/commit/c7a288007462423a8c1b6c1305111a5c82fffeaa
and it works without crashing both before and after the patch.

dzenanz (approved):

thewtex:
> Since we do not need the thread lock,

CI tells us that this is not true on macOS:

- https://open.cdash.org/viewTest.php?onlyfailed&buildid=7246416
- https://open.cdash.org/viewTest.php?onlyfailed&buildid=7246771

Since the test verifies the use of ThreadPool and tqdm has battle-tested this implementation, let's go with it.

PR #2554, opened 2021-05-26
BUG: add support for long long pixel types in PyBuffer

dzenanz:
Problem was reported on the forum:
https://discourse.itk.org/t/runtimeerror-no-suitable-template-parameter-can-be-found/4159

Behavior before this PR:
```text
(pyEnv) M:\a>python kaggle_watershed.py
Traceback (most recent call last):
  File "kaggle_watershed.py", line 24, in <module>
    dist_img, seg_img = apply_watershed(image)
  File "kaggle_watershed.py", line 19, in apply_watershed
    segmentation = itk.GetArrayFromImage(watershedOp.GetOutput())
  File "M:\Dev\ITK-py\Wrapping\Generators\Python\itk\support\extras.py", line 310, in GetArrayFromImage
    image_or_filter, "GetArrayFromImage", keep_axes, update, ttype
  File "M:\Dev\ITK-py\Wrapping\Generators\Python\itk\support\extras.py", line 296, in _GetArrayFromImage
    raise RuntimeError("No suitable template parameter can be found.")
RuntimeError: No suitable template parameter can be found.
```
and after:
```text
(pyEnv) M:\a>python kaggle_watershed.py

(pyEnv) M:\a>
```

The commit is based on `release` branch, as we will want to integrate it there too.

## PR Checklist
- [x] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [x] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [ ] Added test (or behavior not changed)
- [ ] Updated API documentation (or API not changed)
- [ ] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [x] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [ ] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples) for all new major features (if any)

PR #2557, opened 2021-05-28
BUG: add support for long long pixel types in PyBuffer

dzenanz:
Version of #2554 for release branch.


Leengit (approved):
This looks good to me.

Bigger picture than this PR is that on most (or all?) of our systems, the types `UI`, `UL`, and `ULL` collectively represent just the two `uint32_t` and `uint64_t`, so we are wrapping 2 sizes for the price of 3.  Ideally, we would somehow map `UL` to its equivalent type in those stituations where it is equal to `UI` or `ULL` -- instead of wrapping `UL` separately.  (Likewise for signed ints.)

PR #2560, opened 2021-05-30
DOC: Fix enum class name in documentation

jhlegarreta:
Fix enum class name in documentation.

## PR Checklist

### DIFF
--- a/Modules/Filtering/MathematicalMorphology/include/itkMathematicalMorphologyEnums.h
+++ b/Modules/Filtering/MathematicalMorphology/include/itkMathematicalMorphologyEnums.h
@@ -34,7 +34,7 @@ namespace itk
 class MathematicalMorphologyEnums
 {
 public:
-  /**\class AlgorithmType
+  /**\class Algorithm
    * \brief Algorithm or implementation used in the dilation/erosion operations.
    * \ingroup ITKMathematicalMorphology
    */


Modules/Filtering/MathematicalMorphology/include/itkMathematicalMorphologyEnums.h
```diff
@@ -34,7 +34,7 @@ namespace itk
 class MathematicalMorphologyEnums
 {
 public:
-  /**\class AlgorithmType
+  /**\class Algorithm
```
N-Dekker, line 37: Thanks for fixing this type error, @jhlegarreta Unfortunately there's another error in this line: it declares the enum class type as a `class`, but it is not a class! Did you try to see how your fix affects the doxygen HTML output?

In order to get proper doxygen HTML output, I would suggest to simply remove "`\class AlgorithmType`". Doxygen may automatically figure out the name of the enum class type. It appears unnecessary to mention the type name in the doxygen comment.

jhlegarreta: Unfortunately, I have not checked the doxygen HTML output. I'm not willing to build it locally, and if I am not mistaken we are lacking a CI that builds the documentation on the PRs.

I confess that I have not looked at the appropriate doxygen keyword for an `enum class`, and that I just followed what @hjmjohnson adopted in his first sweeping change towards the adoption of strongly typed enums in https://github.com/InsightSoftwareConsortium/ITK/commit/0dd31a62f851a0ae95ef4c6d6c85e531a4ebca59. I think the doxygen keywords have not been changed for any enum since then.

I fear that this should be done for all enum classes for the sake of consistency. Also, note that I submitted this yesterday:
https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/pull/155

Thus, if you are willing to check the doxygen output, and figure out the best way to document the strongly typed enums, I would encourage you to comment on the ITK SW Guide as well so that the potential change is correctly documented.

Thanks @N-Dekker .

N-Dekker: @jhlegarreta I see now, many `enum class` definitions in ITK have that misplaced `\class` doxygen command. So I understand you do not want to fix this just for the `Algorithm` enum class.

You may see the effect of adding `\class` to enum class definitions: they get listed here, with the "classes, structs, unions and interfaces": https://itk.org/Doxygen/html/annotated.html 

While they should instead be documented with the Member Enumeration Documentation, for example here https://itk.org/Doxygen/html/classitk_1_1MathematicalMorphologyEnums.html (You see, the documentation text "Algorithm or implementation used in the dilation/erosion operations." should be here!)

I think it was just a mistake to use `\class` for enum classes. But then, if you don't have time to look at it in more detail, just leave your PR the way it is right now! It obviously fixes a typo!

jhlegarreta: I do agree. An effort should be made to improve these: best would be to open an issue so that it can be tracked, and hopefully all `enum`s documentation gets fixed in a single commit and PR. Thanks for caring @N-Dekker.

jhlegarreta: @N-Dekker I'm resolving this. If you have the time to open an issue concerning the Doxygen class, please go ahead.

jhlegarreta:
The `ITK.macOS.Python` failure is unrelated to this patch set.

N-Dekker (approved):
> The ITK.macOS.Python failure is unrelated to this patch set.

OK, then I'll approve  😄 Thanks Jon!

thewtex (approved):
@jhlegarreta thanks!!

PR #2561, opened 2021-05-30
STYLE: Prefer aliasing template pixel type in `common` operator classes

jhlegarreta:
Prefer aliasing template pixel type in `common` operator classes over
aliasing the superclass pixel type, which is earlier set to the template
pixel type.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkBackwardDifferenceOperator.h
+++ b/Modules/Core/Common/include/itkBackwardDifferenceOperator.h
@@ -55,7 +55,7 @@ public:
   /** Run-time type information (and related methods). */
   itkTypeMacro(BackwardDifferenceOperator, NeighborhoodOperator);
 
-  using PixelType = typename Superclass::PixelType;
+  using PixelType = TPixel;
 
 protected:
   /** Necessary to work around a compiler bug in VC++. */
--- a/Modules/Core/Common/include/itkForwardDifferenceOperator.h
+++ b/Modules/Core/Common/include/itkForwardDifferenceOperator.h
@@ -54,7 +54,7 @@ public:
   /** Run-time type information (and related methods). */
   itkTypeMacro(ForwardDifferenceOperator, NeighborhoodOperator);
 
-  using PixelType = typename Superclass::PixelType;
+  using PixelType = TPixel;
 
 protected:
   /** Necessary to work around VC++ compiler bug. */
--- a/Modules/Core/Common/include/itkSobelOperator.h
+++ b/Modules/Core/Common/include/itkSobelOperator.h
@@ -135,7 +135,7 @@ protected:
   /** Alias support for coefficient vector type. Necessary to
    * work around compiler bug on VC++. */
   using CoefficientVector = typename Superclass::CoefficientVector;
-  using PixelType = typename Superclass::PixelType;
+  using PixelType = typename TPixel;
 
   /** Calculates operator coefficients. */
   CoefficientVector


jhlegarreta:
Left behind in 84f671c. I'm sorry.

Modules/Core/Common/include/itkBackwardDifferenceOperator.h
```diff
@@ -55,7 +55,7 @@ class ITK_TEMPLATE_EXPORT BackwardDifferenceOperator : public NeighborhoodOperat
   /** Run-time type information (and related methods). */
   itkTypeMacro(BackwardDifferenceOperator, NeighborhoodOperator);
 
-  using PixelType = typename Superclass::PixelType;
+  using PixelType = TPixel;
```
N-Dekker, line 58: It seems to me that this alias isn't really necessary, it is? Use code may use `BackwardDifferenceOperatorType::PixelType`, even without this alias, because it is inherited from the base class.

    class Base
    {
    public:
      using SomeType = int;
    };

    class Derived: public Base {};

    int main()
    {
      Derived::SomeType var{42};
    }

You see, in this example, it is unnecessary to add an `Derived::SomeType` alias.

Leengit (approved):
To me the highest priority is that we are consistent from file to file in ITK.  (We don't have to go make things consistent right now, but it should be an eventual goal.)  Within that, I can see value in any of the three possibilities presented.

1. explicitly inherit from `Superclass` -- this makes it obvious that we are being consistent with the superclass, and are not overriding it in anyway.
2. explicitly use something from the `Self` class, such as a template parameter -- this is straightfoward and explicitly shows that we are doing the obviously right thing, and are not using the `Superclass` to override whatever would be natural in the `Self` class.
3. implicitly use the value from `Superclass` by not having any line of code for this in `Self` -- eliminates uneccessary verbiage.

Another possibility that I think is overboard, but I'll mention it in case someone else thinks it is not: use the second option above and then have
```c++
static_assert(is_same<Superclass::PixelType, Self::PixelType>::value, "PixelType is inconsistent between Superclass and Self.");
```
... or something similar using ITK concepts.

jhlegarreta:
>  Use code may use BackwardDifferenceOperatorType::PixelType

True :+1: . I suspect that was the case as well in 84f671c.

So @N-Dekker I can remove all `PixelType` aliases from the `Operator` classes, including those in 84f671c. I'd appreciate if we all agreed on this.

> To me the highest priority is that we are consistent from file to file in ITK

I agree. I wish all these could be agreed upon and made clear in the SW Guide so that there is less place for disagreement and inconsistency, and in such a way that a convention can clearly be referenced.

That being said, I would very much appreciate if an effort was made to add style-wise suggestions to the SW Guide, updating the existing ones if necessary. That way, those of us that unfortunately do not use ITK on a daily basis, and do not even use C++ regularly, and are hence lacking the reflects to more clearly identify redundant code as the one @N-Dekker points, would have a guide to try to avoid introducing inconsistencies.

A few examples come into mind, besides the template parameter aliasing in this PR :
- Initialization lists, which were encouraged a couple of years ago, incorporated to the ITK SW Guide, and are now being discouraged *de facto* throughout the commits favour of brace initializers and brace list initialization in the header files
- The use of `auto`.
- The recent discussion about `std::endl` in https://github.com/InsightSoftwareConsortium/ITK/pull/2536#discussion_r641998361.
- etc.

Changing these in a single PR is unfeasible, but agreeing on a given practice and documenting it will help us close the gap and increasingly reduce inconsistencies.

I know these come largely with increasingly modern C++ practice, but if we do not keep an updated SW Guide, we run the risk of spending time suggesting these changes throughout the commits.

jhlegarreta:
So once #2536 gets hopefully merged, I'll remove all `PixelType` aliases from the child classes and push force here.

jhlegarreta:
And also probably the
```
  /** Typedef support for coefficient vector type.  Necessary to
   *  work around compiler bug on VC++. */
using CoefficientVector = typename Superclass::CoefficientVector;
```
aliases. Hopefully such compiler issues (e.g. commented in the very first commits https://github.com/InsightSoftwareConsortium/ITK/commit/bb3650475a2bd658b96fd75d3883d0e0c1fb8c1c) have now been resolved.

jhlegarreta:
Closing in favor of #2567. Thanks for the time spent in reviewing.

Modules/Core/Common/include/itkBackwardDifferenceOperator.h
jhlegarreta: See #2567.

PR #2563, opened 2021-05-31
ENH: Upgrade ITK from C++11 to C++14

N-Dekker:
Follow-up to commit babb6779b67a4b14f14cc5629da95105c8df4c83
"COMP: Enforce building ITK with C++11", Hans Johnson (@hjmjohnson), 19 January 2018

Discussion item: https://discourse.itk.org/t/may-we-start-using-c-14-at-the-master-branch/4172

### DIFF
--- a/CMake/itkCompilerChecks.cmake
+++ b/CMake/itkCompilerChecks.cmake
@@ -28,13 +28,13 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "Intel" AND
   message(FATAL_ERROR "Intel C++ (ICC) 14.0 or later is required.")
 endif ()
 
-# Make sure we have C++11 enabled.
-if(NOT ITK_IGNORE_CMAKE_CXX11_CHECKS)
+# Make sure we have C++14 enabled.
+if(NOT ITK_IGNORE_CMAKE_CXX14_CHECKS)
   # Needed to make sure libraries and executables not built by the
-  # itkModuleMacros still have the C++11 compiler flags enabled
+  # itkModuleMacros still have the C++14 compiler flags enabled
   # Wrap this in an escape hatch for unknown compilers
   if(NOT CMAKE_CXX_STANDARD)
-    set(CMAKE_CXX_STANDARD 11) # Supported values are ``11``, ``14``, and ``17``.
+    set(CMAKE_CXX_STANDARD 14) # Supported values are 14, 17, 20, and 23.
   endif()
   if(NOT CMAKE_CXX_STANDARD_REQUIRED)
     set(CMAKE_CXX_STANDARD_REQUIRED ON)
--- a/CMake/itkCompilerChecks.cmake
+++ b/CMake/itkCompilerChecks.cmake
@@ -1,31 +1,31 @@
-# Minimum compiler version check: GCC >= 4.8
+# Minimum compiler version check: GCC
 if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND
-    CMAKE_CXX_COMPILER_VERSION VERSION_LESS 4.8)
-  message(FATAL_ERROR "GCC 4.8 or later is required.")
+    CMAKE_CXX_COMPILER_VERSION VERSION_LESS 5.1)
+  message(FATAL_ERROR "GCC 5.1 or later is required.")
 endif ()
 
-# Minimum compiler version check: LLVM Clang >= 3.3
+# Minimum compiler version check: LLVM Clang
 if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND
-    CMAKE_CXX_COMPILER_VERSION VERSION_LESS 3.3)
-  message(FATAL_ERROR "LLVM Clang 3.3 or later is required.")
+    CMAKE_CXX_COMPILER_VERSION VERSION_LESS 3.4)
+  message(FATAL_ERROR "LLVM Clang 3.4 or later is required.")
 endif ()
 
-# Minimum compiler version check: Apple Clang >= 5.0 (Xcode 5.0)
+# Minimum compiler version check: Apple Clang >= 5.1 (Xcode 5.1)
 if (CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" AND
-    CMAKE_CXX_COMPILER_VERSION VERSION_LESS 5.0)
-  message(FATAL_ERROR "Apple Clang 5.0 or later is required.")
+    CMAKE_CXX_COMPILER_VERSION VERSION_LESS 5.1)
+  message(FATAL_ERROR "Apple Clang 5.1 or later is required.")
 endif ()
 
-# Minimum compiler version check: Microsoft C/C++ >= 19.0 (aka VS 2015 aka VS 14.0)
+# Minimum compiler version check: Microsoft C/C++ >= 19.10 (MSVC 14.1, Visual Studio 15 2017)
 if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC" AND
-    CMAKE_CXX_COMPILER_VERSION VERSION_LESS 19.0)
-  message(FATAL_ERROR "Microsoft Visual Studio 2015 or later is required.")
+    CMAKE_CXX_COMPILER_VERSION VERSION_LESS 19.10)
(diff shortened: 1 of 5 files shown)


dzenanz (approved):
I like rolling version updates like this.

jhlegarreta:
The requirements in the ITK SW Guide (section [Configuring and Building ITK](https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/blob/master/SoftwareGuide/Latex/Introduction/Installation.tex)) will need to be updated accordingly.

CMake/itkCompilerChecks.cmake
```diff
@@ -28,13 +28,13 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "Intel" AND
...
-if(NOT ITK_IGNORE_CMAKE_CXX11_CHECKS)
+# Make sure we have C++14 enabled.
+if(NOT ITK_IGNORE_CMAKE_CXX14_CHECKS)
   # Needed to make sure libraries and executables not built by the
-  # itkModuleMacros still have the C++11 compiler flags enabled
+  # itkModuleMacros still have the C++14 compiler flags enabled
   # Wrap this in an escape hatch for unknown compilers
   if(NOT CMAKE_CXX_STANDARD)
-    set(CMAKE_CXX_STANDARD 11) # Supported values are ``11``, ``14``, and ``17``.
+    set(CMAKE_CXX_STANDARD 17) # Supported values are 14, 17, 20, and 23.
```
dzenanz, line 39: Shouldn't this be 14 instead of 17?

dzenanz (commented):
Mostly looks good

CMake/itkCompilerChecks.cmake
N-Dekker: oops, typo!

jhlegarreta (commented):
I do not have a strong opinion on this, since I do not use ITK in my daily work (and henc eI have not voiced my opinion in discourse), but in the past I have seen some resistance to such changes in some contexts (e.g. people maintaining an application that depends on ITK like to see long time support).

Relevant to this, it may be worthwhile revisiting the deprecation cycle statement in https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/blob/master/SoftwareGuide/Latex/Introduction/Installation.tex#L27. I do not know if this change complies with that. If it does, it is a strong argument for it.

N-Dekker:
@jhlegarreta 
> in the past I have seen some resistance to such changes in some contexts (e.g. people maintaining an application that depends on ITK like to see long time support).

Thanks for your very relevant comment, Jon. Generally speaking there are always users who want to stick to the old standards, and those who prefer the latest and greatest. To me, a move to C++14 feels like a compromise, as it is quite an old standard revision already. C++17, which has quite some more interesting features, seems generally available for a while already. But then the relevant question is of course, whether there are still ITK users who would want to keep up-to-date with the latest ITK version, while requiring an older C++ version.

dzenanz:
VS2015 now has some [extended support](https://docs.microsoft.com/en-us/lifecycle/products/visual-studio-2015) until 2025, while mainstream support ended in March 2020 (more than a year ago). And since VS2017 and 2019 are compatible with VS2015's CRT, I think it is OK to drop support for VS2015 now.

Modules/Core/Common/include/itkMacro.h
```diff
@@ -188,14 +188,14 @@ namespace itk
 #  error "The SGI compiler is not supported under ITKv4 and above"
 #endif
 #if defined(__APPLE__)
-#  if defined(__clang__) && (__cplusplus < 201103L)
-#    error "Apple LLVM < 5.0 (clang < 3.3) is not supported under ITKv5"
+#  if defined(__clang__) && (__cplusplus < 201402L)
+#    error "Apple LLVM < 5.1 (clang < 3.4) is not supported under ITKv5.3"
```
N-Dekker, line 192: It appears that ITK.macOS.Python (Azure.Mac-1622555791099 Darwin-Build4705-PR2563-CXX_STANDARD-from-11-to-14-Python) fails here, because `defined(__clang__) && (__cplusplus < 201402L)`

At https://open.cdash.org/viewBuildError.php?buildid=7259248

Any suggestion? Is the compiler properly set to C++14 for ITK.macOS.Python?

CMake/itkCompilerChecks.cmake
```diff
@@ -1,31 +1,33 @@
...
+# Minimum compiler version check: LLVM Clang
 if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND
-    CMAKE_CXX_COMPILER_VERSION VERSION_LESS 3.3)
-  message(FATAL_ERROR "LLVM Clang 3.3 or later is required.")
+    CMAKE_CXX_COMPILER_VERSION VERSION_LESS 3.4)
+  message(FATAL_ERROR "LLVM Clang 3.4 or later is required.")
 endif ()
 
-# Minimum compiler version check: Apple Clang >= 5.0 (Xcode 5.0)
+# Minimum compiler version check: Apple Clang (Xcode)
```
seanm, line 13: @N-Dekker Please restore Xcode version in parens (to 5.1).  Xcode and AppleClang version numbers do not always match up (though they do here).

N-Dekker: @seanm Thanks, so what should it say exactly? `Apple Clang >= 5.1 (Xcode 5.1)` ?

seanm: Yes, perfect. Looks redundant, but on my Mac for example I have Xcode 11.3.1 but it has AppleClang 11.0.0.  People are more familiar with their Xcode version than the actual compiler version.

N-Dekker: Thanks @seanm. Addressed by amend + force-push https://github.com/InsightSoftwareConsortium/ITK/compare/5539f556732dca021e8c9aebb4b12a52b61c428f..a35dcbb8d8469663fd58ef23765380c056f13eb2

CMake/itkCompilerChecks.cmake
```diff
@@ -1,31 +1,33 @@
...
+# Minimum compiler version check: Apple Clang >= 5.1 (Xcode 5.1)
 if (CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" AND
-    CMAKE_CXX_COMPILER_VERSION VERSION_LESS 5.0)
-  message(FATAL_ERROR "Apple Clang 5.0 or later is required.")
+    CMAKE_CXX_COMPILER_VERSION VERSION_LESS 5.1)
+  message(FATAL_ERROR "Apple Clang 5.1 or later is required.")
 endif ()
 
 # Minimum compiler version check: Microsoft C/C++ >= 19.0 (aka VS 2015 aka VS 14.0)
+# Note that VS 2015 has implemented most C++14 features, although it still does not
```
dzenanz, line 20: @blowekamp do we agree we can drop v140 toolkit and move to v141 (VS2017)?

blowekamp: We can drop the v140 toolkit. Even VS2015 has been updated to use v141 by default now.

dzenanz: Awesome! @N-Dekker please bump to VS2017!

N-Dekker: > please bump to VS2017!

@dzenanz Amended and force-pushed, please check: https://github.com/InsightSoftwareConsortium/ITK/compare/a35dcbb8d8469663fd58ef23765380c056f13eb2..40a795f019427f94a06e1b295b0ab66502377dc2

dzenanz (approved):

dzenanz (approved):

Modules/Core/Common/include/itkMacro.h
thewtex: @N-Dekker this may need to be updated:

https://github.com/InsightSoftwareConsortium/ITK/blob/71bb6a99fe02fe80e8e39ed4e00b009b3c7444dd/Wrapping/Generators/CastXML/CMakeLists.txt#L220-L224

thewtex (approved):

dzenanz:
itkPyBufferMemoryLeakTest (Failed) is an unstable test. Merging!

PR #2567, opened 2021-06-02
STYLE: Avoid repeating parent aliases in operator classes

jhlegarreta:
Avoid repeating parent aliases in `Common` operator classes.

## PR Checklist


Modules/Core/Common/include/itkNeighborhoodOperator.h
```diff
@@ -87,6 +87,12 @@ class ITK_TEMPLATE_EXPORT NeighborhoodOperator : public Neighborhood<TPixel, VDi
   /** Slice iterator type alias support */
   using SliceIteratorType = SliceIterator<TPixel, Self>;
 
+  using PixelRealType = typename NumericTraits<TPixel>::RealType;
+
+  /** Typedef support for coefficient vector type. Necessary
```
dzenanz, line 92: typedef or alias? I thought we chose to use alias consistently.

dzenanz (approved):

Modules/Core/Common/include/itkNeighborhoodOperator.h
```diff
@@ -87,6 +87,12 @@ class ITK_TEMPLATE_EXPORT NeighborhoodOperator : public Neighborhood<TPixel, VDi
...
+   * to fix bug in the microsoft VC++ compiler. */
+  using CoefficientVector = std::vector<typename PixelRealType>;
```
dzenanz, line 94: Is CoefficientVector needed to be public? It was protected before.

Modules/Core/Common/include/itkGaussianDerivativeOperator.h
```diff
@@ -210,8 +210,6 @@ class ITK_TEMPLATE_EXPORT GaussianDerivativeOperator : public NeighborhoodOperat
   PrintSelf(std::ostream & os, Indent i) const override;
 
 protected:
-  using CoefficientVector = typename Superclass::CoefficientVector;
```
dzenanz, line 213: Keeping these is probably needed, as per https://stackoverflow.com/questions/1643035/propagating-typedef-from-based-to-derived-class-for-template

N-Dekker: > Keeping these is probably needed

If so, wouldn't a simple C++98 `using` just be fine? As in:

    using typename Superclass::CoefficientVector;

dzenanz: Yes, `using typename Superclass::CoefficientVector;` is sufficient and preferred.

Modules/Core/Common/include/itkNeighborhoodOperator.h
jhlegarreta: Good catch. Changed in d7c9182.

Modules/Core/Common/include/itkNeighborhoodOperator.h
jhlegarreta: You are right, but `PixelRealtype` was/is defined as public, and thus I moved it to the public section. And `PixelRealType` seems to be required in the public section so that the public method `ScaleCoefficients` is allowed to use it.

Modules/Core/Common/include/itkDerivativeOperator.h
```diff
@@ -102,9 +99,11 @@ class ITK_TEMPLATE_EXPORT DerivativeOperator : public NeighborhoodOperator<TPixe
   }
 
 protected:
-  /** Typedef support for coefficient vector type.  Necessary to
-   * work around compiler bug on VC++. */
-  using CoefficientVector = typename Superclass::CoefficientVector;
+  /** Type alias support for pixel real type.*/
+  using typename Superclass::PixelRealType;
```
dzenanz, line 103: If PixelRealType is public in superclass, and we make it protected here, will it cause problems in possible descendants?

jhlegarreta: Although no class inherits from `itk::DerivativeOperator`, I put it back to the public section.

Modules/Core/Common/include/itkNeighborhoodOperator.h
jhlegarreta: Finally, as it seems that the `CoefficientVector` is not being inherited I put it back to the protected section.

hjmjohnson (dismissed):
This is a good start.  Please cherry-pick the last two commits from https://github.com/hjmjohnson/ITK/tree/AvoidRepeatingParentAliasesInCommonOperators. does a more complete refactoring to keep the style and recommendations consistent.

jhlegarreta:
> This is a good start. Please cherry-pick the last two commits from https://github.com/hjmjohnson/ITK/tree/AvoidRepeatingParentAliasesInCommonOperators. does a more complete refactoring to keep the style and recommendations consistent.

Done. Thanks @hjmjohnson :1st_place_medal:.

Sorry for the long delay in doing this.

Modules/Core/Common/include/itkThreadedImageRegionPartitioner.h
```diff
@@ -60,7 +60,7 @@ class ITK_TEMPLATE_EXPORT ThreadedImageRegionPartitioner : public ThreadedDomain
   itkTypeMacro(ThreadedImageRegionPartitioner, ThreadedDomainPartitioner);
 
   /** Type of the object being threaded over */
-  using DomainType = typename Superclass::DomainType;
+  using typename Superclass::DomainType;
```
jhlegarreta, line 63: @hjmjohnson looks like the change here is causing problems on Windows when trying to define the `SizeType` and `IndexType` aliases a few lines below:
https://open.cdash.org/viewBuildError.php?buildid=7451857

Any guess?

I've also noticed that `using ImageRegionType = typename Self::DomainType;` defines yet another alias for the `DomainType`. Not sure both aliases meaning the same type are required.

N-Dekker:
I like very much this style improvement, cool!  😃 Have you been using some script, to replace `using X = Superclass::X` by `using X`? If so, I would like to try it on https://github.com/SuperElastix/elastix as well...  😺

Modules/Core/Common/include/itkThreadedImageRegionPartitioner.h
N-Dekker: @jhlegarreta I tried to build it locally on Windows, and got the same compilation errors, from VS2019. It appears that it can be solved by doing:

```
  using SizeType = typename ImageRegionType::SizeType;
  using IndexType = typename ImageRegionType::IndexType;

```

Instead of:
https://github.com/InsightSoftwareConsortium/ITK/blob/201537c60b2184b3febf5706d39dce71024cd769/Modules/Core/Common/include/itkThreadedImageRegionPartitioner.h#L68-L69

jhlegarreta: Thanks for having given this a try @N-Dekker :100: . I still wonder why two aliases (`DomainType` and `ImageRegionType`) are created to refer to the same type.

dzenanz (approved):
@N-Dekker Hans' changes are so extensive I can't imagine them being done manually 😄

hjmjohnson:
> I like very much this style improvement, cool! 😃 Have you been using some script, to replace `using X = Superclass::X` by `using X`? If so, I would like to try it on https://github.com/SuperElastix/elastix as well... 😺

Just some vim- macro magic.  I did it on the fly with a simple regular expression substitution macro.

Should be easy to do with sed as well. 's/using *\([^ ]*\) *= *typedef *Superclass::\1/using typename Superclass::\1/g'

Modules/Core/Common/include/itkThreadedImageRegionPartitioner.h
jhlegarreta: Went ahead and adopted the suggestion @N-Dekker proposed. Push-forced in 431d91c. I also realized that there is a comment saying that these types are deprecated (although no pre-processor guard is added). It is left for a separate exercise if anybody has the bandwidth to have a look

dzenanz (approved):

jhlegarreta:
The CDash builds are not happy with this change :worried:.

As for the failure in `itkMultiphaseDenseFiniteDifferenceImageFilterTest.cxx`, I think the superclass alias is missing the last template parameter ( `TFunction`) [here](
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Nonunit/Review/test/itkMultiphaseDenseFiniteDifferenceImageFilterTest.cxx#L41). Which seems to be a bug that surfaced thanks to the PR at least :hammer_and_wrench:.

I will propose it later today.

The rest of the failures to the reported 50 require more investigation.

dzenanz:
Oh, I just now see this. That's what happens when there are 30 new emails in the morning 😄

jhlegarreta:
@hjmjohnson I had a quick look yesterday to the rest of the compilation errors, but was not able to come up with a solution. Can you please have a look? Thanks.

dzenanz:
@seanm just submitted #2753.

jhlegarreta:
> @seanm just submitted #2753.

The fix for one of the files has already been merged in #2751. I missed the other file. Sorry and thanks @seanm.

seanm:
I've just rebased #2753.

Does the github-based CI have `Module_ITKReview` enabled anywhere?  I'm guessing not... it probably should...

dzenanz:
Review module is fairly small now. I think we can enable it for the CI. PR incoming.

PR #814, opened 2019-05-03
BUG: Update and enable ITKv3 Examples

blowekamp:
## PR Checklist
- [ ] [Makes breaking changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes)
- [ ] [Makes design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes)
- [ ] Adds the License notice to new files.
- [ ] Adds Python wrapping.
- [ ] Adds tests and baseline comparison (quantitative).
- [ ] [Adds test data](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/UploadBinaryData.md).
- [ ] Adds Examples to [ITKExamples](https://github.com/InsightSoftwareConsortium/ITKExamples)
- [ ] Adds Documentation.

Refer to the [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) for
further development details if necessary.

PR #2568, opened 2021-06-02
STYLE: Use macros to Set/Get ivars in `Common` operators

jhlegarreta:
Use macros to Set/Get ivars in `Common` operators.

Take advantage of the commit to leave the documentation of the ivars only
in their corresponding Set/Get methods, as dictated by the ITK SW Guide.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkLaplacianOperator.h
+++ b/Modules/Core/Common/include/itkLaplacianOperator.h
@@ -96,8 +96,8 @@ public:
     os << indent << "DerivativeScalings: " << m_DerivativeScalings << std::endl;
   }
 
-  /** Sets the weights that are applied to the derivative in each axial
-   *  direction when the kernel is computed.  These weights are all 1.0 by
+  /** Set/Get the weights that are applied to the derivative in each axial
+   *  direction when the kernel is computed. These weights are all 1.0 by
    *  default. This method must be called BEFORE CreateOperator */
   void
   SetDerivativeScalings(const double * s);
--- a/Modules/Core/Common/include/itkNeighborhoodOperator.h
+++ b/Modules/Core/Common/include/itkNeighborhoodOperator.h
@@ -89,7 +89,7 @@ public:
 
   using PixelRealType = typename NumericTraits<TPixel>::RealType;
 
-  /** Sets the dimensional direction of a directional operator. */
+  /** Set/Get the direction (dimension number) of a directional operator. */
   void
   SetDirection(const unsigned long direction)
   {
@@ -100,13 +100,7 @@ public:
     }
     m_Direction = direction;
   }
-
-  /** Returns the direction (dimension number) of a directional operator. */
-  unsigned long
-  GetDirection() const
-  {
-    return m_Direction;
-  }
+  itkGetConstNonVirtualMacro(Direction, unsigned long);
 
   /** Creates the operator with length only in the specified direction.
    * The radius of the operator will be 0 except along the axis on which
@@ -180,7 +174,6 @@ protected:
   }
 
 private:
-  /** Direction (dimension number) of the derivative. */
   unsigned long m_Direction{ 0 };
 };
 } // namespace itk
--- a/Modules/Core/Common/include/itkAnnulusOperator.h
+++ b/Modules/Core/Common/include/itkAnnulusOperator.h
@@ -96,11 +96,7 @@ public:
   {
     m_InnerRadius = r;
   }
-  double
-  GetInnerRadius() const
-  {
-    return m_InnerRadius;
-  }
+  itkGetConstNonVirtualMacro(InnerRadius, double);
 
   /** Set/Get the thickness of the annulus.  The outer radius of the
    * annulus is defined as r = InnerRadius + Thickness. Thickness is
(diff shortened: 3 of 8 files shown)


Modules/Core/Common/include/itkLaplacianOperator.h
```diff
@@ -97,8 +97,8 @@ class ITK_TEMPLATE_EXPORT LaplacianOperator : public NeighborhoodOperator<TPixel
     Superclass::PrintSelf(os, i.GetNextIndent());
   }
 
-  /** Sets the weights that are applied to the derivative in each axial
-   *  direction when the kernel is computed.  These weights are all 1.0 by
+  /** Set/Get the weights that are applied to the derivative in each axial
+   *  direction when the kernel is computed. These weights are all 1.0 by
    *  default. This method must be called BEFORE CreateOperator */
   void
   SetDerivativeScalings(const double * s);
```
jhlegarreta, line 104: Still not convinced by this and its getter method due to this:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/test/itkNeighborhoodOperatorTest.cxx#L150
If somebody happens to have a proposal, I'd consider it.

jhlegarreta:
Not sure about what the errors imply:
https://open.cdash.org/viewBuildError.php?buildid=7260537

Suggestions are welcome.

Modules/Core/Common/include/itkNeighborhoodOperator.h
```diff
@@ -87,19 +87,13 @@ class ITK_TEMPLATE_EXPORT NeighborhoodOperator : public Neighborhood<TPixel, VDi
...
     m_Direction = direction;
   }
-
-  /** Returns the direction (dimension number) of a directional operator. */
-  unsigned long
-  GetDirection() const
-  {
-    return m_Direction;
-  }
+  itkGetConstMacro(Direction, unsigned long);
```
dzenanz, line 96: This macro calls `this->Modified();`, which is declared in `itk::Object`. I guess that `itk::NeighborhoodOperator` does not derive from it.

jhlegarreta: Yep, I did not dig further, but that was what I thought when I left the keyboard. Thanks Dženan.

Maybe I'm mistaken and there are good reasons not to do so, but I'd say that all ITK objects should be able to use the Get/Set macros. `itk::NeighborhoodOperator`'s base class does not inherit from `itk::Object`: 
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkNeighborhood.h

Maybe this was a justified design decision. Hopefully people will be able to comment.

hjmjohnson: It may be a performance concern.  

itkFixedArray objects do not derive from itkObject for that reason.

N-Dekker: A `GetDirection()` member function defined by `itkGetConstMacro` would be _virtual_, which is probably not necessary. 

Makes me think, maybe we should have an `itkGetConstNonVirtualMacro` !

jhlegarreta:
Push forced to see the error reports back and see where we were at this level.

N-Dekker:
I'm sorry but I still don't like to have a `virtual` keyword added to those Get and Set member functions. Especially when they were originally non-virtual, and when they worked just fine without a `virtual` keyword. The only reason for adding `virtual` to an existing member function should be that some user _actually_ needs to override it.

C++ Core Guidelines [C.132: Don’t make a function virtual without reason](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-virtual) says on this topic:
> Redundant virtual increases run-time and object-code size. A virtual function can be overridden and is thus open to mistakes in a derived class...

We already spent a lot of time in the last few years, removing `virtual` from the ITK code base, and we're still not finished. Removal of `virtual` might break legacy user code, even while it is an improvement to new user code, which is why Bradley (@blowekamp) introduced a `ITK_ITERATOR_VIRTUAL` macro, for legacy support.

https://github.com/InsightSoftwareConsortium/ITK/blob/6e3d31d138ff48ae6e5f2e32fcb17c5f172a29ed/Modules/Core/Common/include/itkMacro.h#L1314-L1326

So please only make a member function `virtual` when it is actually needed!

jhlegarreta:
Thanks for the comment @N-Dekker . Then, as pointed in [this comment of yours](https://github.com/InsightSoftwareConsortium/ITK/pull/2568#discussion_r645549336) I'll mark it to `status:Blocked` until the proposed `itkGetConstNonVirtualMacro` gets added to the ITK code base.

hjmjohnson:
@jhlegarreta Is this still something that needs to be worked on?    If so, could you please rebase and make a game plan for what needs to be done?

jhlegarreta:
> @jhlegarreta Is this still something that needs to be worked on? If so, could you please rebase and make a game plan for what needs to be done?

Thanks for the heads-up Hans. Still on my ToDo list; have not had the time too look at my ITK-related ToDo list in a while. I do think this still deserves having a go. Not sure when I'll be able to take it up, though.

Also, wisest might be to wait until what is mentioned here is resolved/done: https://github.com/InsightSoftwareConsortium/ITK/pull/2568#discussion_r724308177

thewtex (dismissed):

jhlegarreta:
@N-Dekker [expressed concerns](https://github.com/InsightSoftwareConsortium/ITK/pull/2568#issuecomment-932973030) about this, but going over the files I do not see any of these members being `virtual`, and inspecting the commits I do not see where they were being declared as virtual. Sorry @hjmjohnson. @N-Dekker do you think that we should still wait till you have time to resolve https://github.com/InsightSoftwareConsortium/ITK/pull/2785 to merge this?

hjmjohnson:
Per the request to consolidate the scattered `virtual` / `override` / `final` design conversation around ITK Get/Set macros, I've opened **#6232 (Master tracking — virtual / override / final semantics on Get/Set macros for ITK 6.x)** as a single venue.

It pulls in the 2018 [Discourse #814](https://discourse.itk.org/t/removal-of-virtual-keywords-from-constneighborhooditerator/814) precedent (the same `final`-as-stress-test technique and the @phcerdan / @N-Dekker 18–24% HoughTransform performance gain), the 2026-05-07 final-keyword audit (17 base methods overridden across 19 files in upstream ITK), and a draft additive `itkVirtualSetMacro` family on `hjmjohnson:macro-optionally-virtual @ ffc70ed0b9`. Surfaces 7 binary/short-list questions — Q1 (default-emit `virtual`?) is the load-bearing one this PR is blocked on.

Suggest holding further design comments on this thread until #6232 has a maintainer-quorum answer on Q1; otherwise the conversation re-fragments. This thread can stay open as the implementation venue once direction is settled.

jhlegarreta:
Closing after offline discussion with @hjmjohnson so that work can make progress without this being a roadblock.

PR #2573, opened 2021-06-03
Remove WriteCompilerDetectionHeader

### LINKED ISSUE #2465: WriteCompilerDetectionHeader module is  removed in CMake 3.20
Some discussion about this is provided in #2373. While #2464 does get rid of the warning, we should stop using `WriteCompilerDetectionHeader`. This will require non-trivial effort.

dzenanz:
This is a follow-up to and partial reversal of #2464. Closes #2465.

We now require C++14, which includes all of the granular
C++11 features which used to be detected before.


dzenanz:
This PR requires #2647, so it is rebased on top of it.

hjmjohnson (approved):

PR #2578, opened 2021-06-04
ENH: Add numbers to the three `itkPyBufferMemoryLeakTest` print messages

N-Dekker:
Add numbers (1, 2, 3) to the print messages in itkPyBufferMemoryLeakTest,
to ease analysis of the output from a test failure.

### DIFF
--- a/Modules/Bridge/NumPy/wrapping/test/itkPyBufferMemoryLeakTest.py
+++ b/Modules/Bridge/NumPy/wrapping/test/itkPyBufferMemoryLeakTest.py
@@ -34,7 +34,7 @@ for i in range(n):
     M.append(resource.getrusage(resource.RUSAGE_SELF).ru_maxrss)
 
 if M[5] - M[4] > 1000:
-    print("Memory leak!")
+    print("Memory leak 1!")
     sys.exit(1)
 
 # creating new numpy volume each time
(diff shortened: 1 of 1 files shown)


N-Dekker:
@dzenanz Could the `itkPyBufferMemoryLeakTest` failures possibly be introduced by pull request #2554 "BUG: add support for long long pixel types in PyBuffer", merged by cd36c640a0a918d914c7fe102748c1ca00fa074a ?

dzenanz (approved):
Incremental build didn't work for me, I needed to manually rebuild something to avoid a clean build.

While it is possible that did it, I see no reason why it would do that. This same test used to be flaky on Mac before, if I remember correctly.

N-Dekker:
> While it is possible that PR did it, I see no reason why it would do that.

I don't know either. Would it matter that these `types` from itkPyBuffer and itkImageDuplicator have become out of sync, with PR  #2554 ?

https://github.com/InsightSoftwareConsortium/ITK/blob/cd36c640a0a918d914c7fe102748c1ca00fa074a/Modules/Bridge/NumPy/wrapping/itkPyBuffer.wrap#L5


https://github.com/InsightSoftwareConsortium/ITK/blob/cd36c640a0a918d914c7fe102748c1ca00fa074a/Modules/Core/Common/wrapping/itkImageDuplicator.wrap#L11-L13

hjmjohnson (approved):

dzenanz:
Even if it does not fix the pyBuffer memory leak, ImageDuplicator should be wrapped with LL types. -> Done in #2584.

PR #2579, opened 2021-06-05
STYLE: C++14 auto return type for ConnectedImageNeighborhoodShapeOffsets

N-Dekker:
Simplified `GenerateConnectedImageNeighborhoodShapeOffsets()` by using
C++14 automatic return type deduction.

### DIFF
--- a/Modules/Core/Common/include/itkConnectedImageNeighborhoodShape.h
+++ b/Modules/Core/Common/include/itkConnectedImageNeighborhoodShape.h
@@ -257,22 +257,12 @@ private:
   {
     return (includeCenterPixel ? 1 : 0) + CalculateNumberOfConnectedNeighbors(maximumCityblockDistance);
   }
-
-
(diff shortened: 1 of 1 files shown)


dzenanz (approved):
GenerateConnectedImageNeighborhoodShapeOffsets is no longer a friend of the class. Does that matter?

N-Dekker:
> GenerateConnectedImageNeighborhoodShapeOffsets is no longer a friend of the class.

Well spotted, @dzenanz The "generate offsets" function used to be a friend of the shape class, in order to call its private helper function, `Shape::CalculateNumberOfOffsets()`  (at compile-time!) Which was necessary to specify its return type (specifically, the size of the `std::array` of offsets that it returns).

With C++14 automatic return type deduction, it is no longer necessary for the generate function to call `Shape::CalculateNumberOfOffsets()`, not even at compile time. Which was why I "unfriended" the function from the class. 😃

PR #2580, opened 2021-06-05
STYLE: Use C++14 <type_traits> template aliases instead of nested types

N-Dekker:
Replaced uses of nested types from `<type_traits>` (from C++11) by the
corresponding C++14 template aliases. Removed the corresponding `typename`
keywords that were originally necessary. Did so by running the following
command, at the GNU bash prompt, from the root of the ITK source tree:

    find . \( -iname itk*.cxx -or -iname itk*.hxx -or -iname itk*.h \) -exec perl -pi -w -e 's/typename std::(\w+)<(.+)>::type/std::$1_t<$2>/g;' {} \;

### DIFF
--- a/Modules/Core/Common/include/itkImageBufferRange.h
+++ b/Modules/Core/Common/include/itkImageBufferRange.h
@@ -85,7 +85,7 @@ private:
   constexpr static bool SupportsDirectPixelAccess =
     std::is_same<PixelType, InternalPixelType>::value &&
     std::is_same<typename TImage::AccessorType, DefaultPixelAccessor<PixelType>>::value &&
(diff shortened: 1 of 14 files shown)


N-Dekker:
Original proposal of these C++14 template aliases:

Walter E. Brown, 2013-03-12, TransformationTraits Redux, http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3546.pdf
Walter E. Brown, 2013-04-18, TransformationTraits Redux, v2, http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3655.pdf

Related issue:

Loïc Joly, 15 October 2020, The "_t" and "_v" version of type traits should be used instead of "::type" and "::value", https://jira.sonarsource.com/browse/RSPEC-6020

dzenanz (approved):
Awesome!

Issue #2583, opened 2021-06-07
Explicit instantiation of itk::Image fails with 5.2 when pixel type does not support operator==


blowekamp:
With the additional of itk::Image::operator==, it is a *new* requirement for the Image's pixel type to support operator==. Previously, SimpleITK ( in some configurations ) had been explicitly instantiating the type: `itk::Image<std::deque<itk::LabelObjectLine<2u>, std::allocator<itk::LabelObjectLine<2u> > >, 1u>` because it was frequently used within the LabelMap module. With ITK 5.2 this results in a compilation error due to operator== not being supported. See SimpleITK/SimpleITK#1404 . While, in this case the explicit instantiation can just be removed, a more general solution in ITK would be good.

Do we want to explicitly require the pixel type to support the operator==, perhaps with a concept check?

Should the itk::Image::operator== be disabled when not supported by the pixel?

N-Dekker:
@blowekamp Sorry, when I proposed `itk::Image::operator==`, I did not realize that some pixel types might not supports `operator==`.

However, it seems quite reasonable to me to require that a pixel type does indeed support `operator==`. For [itk::LabelObjectLine](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/LabelMap/include/itkLabelObjectLine.h) it seems quite easy to add `operator==`, right?

blowekamp:
> However, it seems quite reasonable to me to require that a pixel type does indeed support `operator==`. For [itk::LabelObjectLine](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/LabelMap/include/itkLabelObjectLine.h) it seems quite easy to add `operator==`, right?

Component in ITK are frequently are used in interesting and unexpected ways.

That would address this particular case. If that approach is followed, I'd recommend adding some type of concept check to the itk::Image.

dzenanz:
A more meaningful error message from a concept check sound like a good solution.

PR #2585, opened 2021-06-07
COMP: Support non-EqualityComparable pixels Image template instantiation

N-Dekker:
The ability to instantiate `itk::Image<TPixel, VImageDimension>` for a
non-EqualityComparable pixel type was broken by:

"ENH: Add operator== and operator!= to itk::Image"
pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2188
commit 7f3337d27ee684aa4ab6ccbd294603926895ee82

Reported by Bradley Lowekamp:
issue https://github.com/InsightSoftwareConsortium/ITK/issues/2583
"Explicit instantiation of itk::Image fails with 5.2 when pixel type
does not support operator=="

This commit fixes the issue by defining the `operator==` and `operator!=`
friend functions of `itk::Image` as function _templates_, rather than as
non-templates.

### DIFF
--- a/Modules/Core/Common/include/itkImage.h
+++ b/Modules/Core/Common/include/itkImage.h
@@ -27,6 +27,8 @@
 #include "itkWeakPointer.h"
 #include "itkNeighborhoodAccessorFunctor.h"
 
+#include <type_traits> // For is_same
+
 namespace itk
 {
 /** \class Image
@@ -318,9 +320,15 @@ public:
   unsigned int
   GetNumberOfComponentsPerPixel() const override;
 
-  /** Returns (image1 == image2). */
-  friend bool
-  operator==(const Image & lhs, const Image & rhs)
+  /** Returns (image1 == image2).
+   * \note `operator==` and `operator!=` are defined as function templates
+   * (rather than as non-templates), just to allow template instantiation of
+   * `itk::Image` for non-EqualityComparable pixel types.
+   */
+  template <typename TEqualityComparable>
+  friend std::enable_if_t<std::is_same<TEqualityComparable, TPixel>::value, bool>
+  operator==(const Image<TEqualityComparable, VImageDimension> & lhs,
+             const Image<TEqualityComparable, VImageDimension> & rhs)
   {
     if ((lhs.GetBufferedRegion() != rhs.GetBufferedRegion()) || (lhs.m_Spacing != rhs.m_Spacing) ||
         (lhs.m_Origin != rhs.m_Origin) || (lhs.m_Direction != rhs.m_Direction) ||
@@ -349,16 +357,18 @@ public:
       return false;
     }
 
-    const TPixel * const lhsBufferPointer = lhsBuffer.GetBufferPointer();
-    const TPixel * const rhsBufferPointer = rhsBuffer.GetBufferPointer();
+    const TEqualityComparable * const lhsBufferPointer = lhsBuffer.GetBufferPointer();
+    const TEqualityComparable * const rhsBufferPointer = rhsBuffer.GetBufferPointer();
 
     return ((lhsBufferPointer == rhsBufferPointer) ||
             std::equal(lhsBufferPointer, lhsBufferPointer + bufferSize, rhsBufferPointer));
   }
 
   /** Returns (image1 != image2). */
-  friend bool
-  operator!=(const Image & lhs, const Image & rhs)
+  template <typename TEqualityComparable>
+  friend std::enable_if_t<std::is_same<TEqualityComparable, TPixel>::value, bool>
+  operator!=(const Image<TEqualityComparable, VImageDimension> & lhs,
+             const Image<TEqualityComparable, VImageDimension> & rhs)
   {
     return !(lhs == rhs);
   }
--- a/Modules/Core/Common/test/itkImageGTest.cxx
+++ b/Modules/Core/Common/test/itkImageGTest.cxx
@@ -133,8 +133,21 @@ Expect_unequal_when_pixel_values_differ()
   Expect_unequal(*image1, *image2);
 }
 
+
+// An example of a type that does not support `x == y`, for instances `x`, `y`.
+// (Neither does it support `x != y`).
+struct NonEqualityComparableType
+{
+  int data;
+};
+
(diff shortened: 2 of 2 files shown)


Leengit:
Nice.

Do we want the template to include a `NVImageDimension` parameter too, and then check it for equality with `VImageDimension` ... so that `image1==image2` will compile (and return `false`) if the two images have different image dimensions?

N-Dekker:
> Do we want the template to include a `NVImageDimension` parameter too, and then check it for equality with `VImageDimension` ... so that `image1==image2` will compile (and return `false`) if the two images have different image dimensions?

Nice suggestion @Leengit I think I would rather have a compilation error on `image1==image2`, when their dimensions differ.  But please let us know if you have a useful use case to support  `image1==image2`, when their dimensions differ.

dzenanz (approved):
If it compiles and works as intended, good.

N-Dekker:
> If it compiles and works as intended, good.

I must say, that Linux-warning from is non-ObVious 😸 to me: https://open.cdash.org/viewBuildError.php?type=1&buildid=7271820

> Modules/Core/Common/include/itkImage.hxx:79:1: warning: 'void itk::Image<TPixel, VImageDimension>::FillBuffer(const TPixel&) [with TPixel = {anonymous}::NonEqualityComparableType; unsigned int VImageDimension = 2]' defined but not used [-Wunused-function]

Any suggestion?

dzenanz:
> Any suggestion?

Sadly, no. It is just as cryptic to me.

N-Dekker:
@dzenanz Just fixed the Linux GCC warning, which said:  

> 'FillBuffer(const TPixel&)' defined but not used [-Wunused-function]

The fix (as I just force-pushed) was to simply use the function, en another unit test! 😄

dzenanz (approved):

Leengit (commented):
In `itkImage.h`, is this line correct?:
```c++
const TPixel * const lhsBufferPointer = lhsBuffer.GetBufferPointer();
```
I worry that `TPixel` is the pixel type of the right-hand side, but not necessarily of the left-hand side.

N-Dekker:
> In `itkImage.h`, is this line correct?:
> 
> ```c++
> const TPixel * const lhsBufferPointer = lhsBuffer.GetBufferPointer();
> ```
> 
> I worry that `TPixel` is the pixel type of the right-hand side, but not necessarily of the left-hand side.

Well observed @Leengit   Would you like it better if it would use `auto` instead?

Please keep in mind that the aim of this pull request is just to allow `itk::Image` template instantiations for non-EqualityComparable pixel types. It isn't meant to further enhance `operator==`.

Leengit:
> > In `itkImage.h`, is this line correct?:
> > ```c++
> > const TPixel * const lhsBufferPointer = lhsBuffer.GetBufferPointer();
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I worry that `TPixel` is the pixel type of the right-hand side, but not necessarily of the left-hand side.
> 
> Well observed @Leengit Would you like it better if it would use `auto` instead?
> 
> Please keep in mind that the aim of this pull request is just to allow `itk::Image` template instantiations for non-EqualityComparable pixel types. It isn't meant to further enhance `operator==`.

So that the compiler catches type errors for me, I usually use the actual type if I can determine it and if it isn't completely obvious -- e.g., because we just cast to it on the right-hand side.  In this case, I believe that that would be `TEqualityComparable`, but `auto` would work too.

N-Dekker:
> So that the compiler catches type errors for me, I usually use the actual type if I can determine it and if it isn't completely obvious -- e.g., because we just cast to it on the right-hand side. In this case, I believe that that would be `TEqualityComparable`, but `auto` would work too.

OK @Leengit If I amend the commit to use `TEqualityComparable`, will you approve the pull request for me?  😸

Leengit:
> > So that the compiler catches type errors for me, I usually use the actual type if I can determine it and if it isn't completely obvious -- e.g., because we just cast to it on the right-hand side. In this case, I believe that that would be `TEqualityComparable`, but `auto` would work too.
> 
> OK @Leengit If I amend the commit to use `TEqualityComparable`, will you approve the pull request for me? smile_cat

It's a deal.

N-Dekker:
> > OK @Leengit If I amend the commit to use `TEqualityComparable`, will you approve the pull request for me? smile_cat
> 
> It's a deal.

@Leengit Thanks! Please check the amend + force-push: https://github.com/InsightSoftwareConsortium/ITK/compare/4549ea34b6f124cce48c6dcf7ba33aea9476e4b9..b02c9cfdb9788cbd02388f3579e26903a1654f52  Hope you appreciate the `static_assert` that I added with the amend. 😃

Leengit:
> @Leengit Thanks! Please check the amend + force-push: https://github.com/InsightSoftwareConsortium/ITK/compare/4549ea34b6f124cce48c6dcf7ba33aea9476e4b9..b02c9cfdb9788cbd02388f3579e26903a1654f52 Hope you appreciate the `static_assert` that I added with the amend. smiley

The `TEqualityComparable` for `lhsBufferPointer` is good, thank you.

Is the thinking that the `static_assert` will make sure that `operator==` is only ever used the way it was before this pull request?  That seems reasonable until we have a need for and an implementation of a more general use of `operator==` for images.

Leengit (approved):

N-Dekker:
> Is the thinking that the `static_assert` will make sure that `operator==` is only ever used the way it was before this pull request?

Indeed  😃 

> That seems reasonable until we have a need for and an implementation of a more general use of `operator==` for images.

Yes, exactly. The static_assert may be removed once (if ever) we need `operator==` to compare two images of different types. And then of course some extra unit tests must be added. But for now, at least the template instantiation issue should be fixed.

N-Dekker:
Honestly this is not my favorite pull request, because I prefer the two parameters of `operator==` and `operator!=` (lhs, rhs)  to be declared exactly the same way. But as this pull request does fix issue #2583, I think it may be merged anyway.

blowekamp (approved):
Thanks for addressing the reported issue 👍

PR #2589, opened 2021-06-09
C++20 gcc-11 compatibility

darrent1974:
## PR Checklist

### DIFF
--- a/Modules/Core/Common/src/itkSmapsFileParser.cxx
+++ b/Modules/Core/Common/src/itkSmapsFileParser.cxx
@@ -300,7 +300,7 @@ ITKCommon_EXPORT std::istream &
 template <typename TFirstType>
 struct MapRecordPlusor
 {
-  MapRecordPlusor<TFirstType>(const char * token = "Size")
+  MapRecordPlusor(const char * token = "Size")
     : m_Token(token)
   {}
 
@@ -320,7 +320,7 @@ struct MapRecordPlusor
 template <typename TFirstType>
 struct MapRecordConditionalPlusor
 {
-  MapRecordConditionalPlusor<TFirstType>(const char * filter, const char * token = "Size")
+  MapRecordConditionalPlusor(const char * filter, const char * token = "Size")
     : m_Filter(filter)
     , m_Token(token)
   {}
--- a/Modules/ThirdParty/VNL/src/vxl/core/vnl/algo/vnl_svd_fixed.h
+++ b/Modules/ThirdParty/VNL/src/vxl/core/vnl/algo/vnl_svd_fixed.h
@@ -163,7 +163,7 @@ class vnl_svd_fixed
   bool valid_;        // false if the NETLIB call failed.
 
   // Disallow assignment.
-  vnl_svd_fixed<T,R,C>(vnl_svd_fixed<T,R,C> const &) { }
+  vnl_svd_fixed(vnl_svd_fixed<T,R,C> const &) { }
   vnl_svd_fixed<T,R,C>& operator=(vnl_svd_fixed<T,R,C> const &) { return *this; }
 };
 
--- a/Modules/ThirdParty/VNL/src/vxl/core/vnl/vnl_matrix.h
+++ b/Modules/ThirdParty/VNL/src/vxl/core/vnl/vnl_matrix.h
@@ -735,7 +735,7 @@ class VNL_EXPORT vnl_matrix
 //--------------------------------------------------------------------------------
 
  protected:
-  vnl_matrix<T>( unsigned ext_num_rows, unsigned ext_num_cols,
+  vnl_matrix( unsigned ext_num_rows, unsigned ext_num_cols,
       T * continuous_external_memory_block, bool manage_own_memory )
   : num_rows{ ext_num_rows}
   , num_cols{ ext_num_cols}
--- a/Modules/ThirdParty/VNL/src/vxl/core/vnl/vnl_vector.h
+++ b/Modules/ThirdParty/VNL/src/vxl/core/vnl/vnl_vector.h
@@ -459,7 +459,7 @@ class VNL_EXPORT vnl_vector
     set_data(datain, this->size(), LetArrayManageMemory);
   }
  protected:
-  vnl_vector<T>( size_t ext_num_elmts, T * extdata, bool manage_own_memory )
+  vnl_vector( size_t ext_num_elmts, T * extdata, bool manage_own_memory )
     : num_elmts{ ext_num_elmts }
     , data{ extdata }
     , m_LetArrayManageMemory{ manage_own_memory }


thewtex:
@darrent1974 thanks for the contribution :pray: 

After CI is green, can the VXL changes please be integrated upstream?

https://github.com/vxl/vxl

Modules/Core/Common/include/itkFixedArray.h
```diff
@@ -254,11 +254,19 @@ class ITK_TEMPLATE_EXPORT FixedArray
   bool
   operator==(const FixedArray & r) const;
 
+#if __cplusplus < 202002L
+  /** For C++ versions < C++20 define the != operator.
+   * For C++20 != is now a secondary operator, as such
+   * the compiler will now use a ! applied to the primary
+   * == operator, which makes the implementation redundant
+   * in this context.
+   */
```
N-Dekker, line 263: Very interesting @darrent1974 Which compile error did you get from GCC11 when compiling for C++20? I tried the following on gcc 11.1 `-std=c++20`, at https://godbolt.org/z/P7P9315vb but it does seem to compile fine:

```
class Foo
{
public:
    bool operator==(const Foo&) const { return true; }

    bool operator!=(const Foo& arg) const
    {
        return !(*this == arg);
    }
};

int main()
{
    return (Foo{} == Foo{}) && (Foo{} != Foo{});
}

```

Modules/Core/Common/src/itkSmapsFileParser.cxx
```diff
@@ -300,7 +300,7 @@ ITKCommon_EXPORT std::istream &
 template <typename TFirstType>
 struct MapRecordPlusor
 {
-  MapRecordPlusor<TFirstType>(const char * token = "Size")
+  MapRecordPlusor(const char * token = "Size")
```
N-Dekker, line 303: Thanks @darrent1974 Looks fine, of course, with any version of the C++ standard. I do observe a compilation error, indeed, when trying gcc 11.1 `-std=c++20`, at https://godbolt.org/z/14T4E7jGx

```
template <typename T> class Bar
{
public:
  Bar<T>() {}
};

```

It says:
```

source>:8:10: error: expected unqualified-id before ')' token
    8 |   Bar<T>() {}
      |          ^
```

Is that the error you encountered as well? If so, maybe you could mention the error message in the commit message. So that we can easily see (and verify) that it isn't just a cosmetic fix.

BTW I didn't yet know that this is a C++20 language change! Thanks for the information 😃

N-Dekker:
@darrent1974 Could you possibly reword the commit subject "COMP: Removed constructor template parameters" and "COMP: Removed constructor template parameter" a little bit, to clarify which one is which? At ITK we allow (or we suggest) up to 72 characters of the subject line of a commit. For example:

     COMP: Removed constructor template parameters from the VNL library

and:

     COMP: Removed constructor template parameters from itkSmapsFileParser

Modules/Core/Common/src/itkSmapsFileParser.cxx
darrent1974: Yes, that's the one I'll change it accordingly.

Modules/Core/Common/include/itkFixedArray.h
darrent1974: Hi @N-Dekker, forgive me for the commit messages. It's my first non-trivial pull-request to ITK and I was a bit uncertain as to how much should be said.

Regarding the changes to itkFixedArray and itkPoint wrt the == and != operators. It was quite subtle and took be a while to nut it out. I was getting "ambiguous overload" errors for the operators on these related classes. See the snippet at the bottom.

After a bit of digging around and googling I came across the following article [https://brevzin.github.io/c++/2019/07/28/comparisons-cpp20/](https://brevzin.github.io/c++/2019/07/28/comparisons-cpp20/) which explained some changes regarding comparison operators.

After digging down into the code and found that the == operators are identical for itkFixedArray and itkPoint which I results in the ambiguity. Given this, I decided (for C++20) to discard the definition of == for itkPoint as its parent is sufficient.

Also, given the changes regarding the now "secondary" != operator it seemed redundant to now define != explicitly on itkFixedArray as the compiler would nicely insert a !==.

Let me know if you want a better commit message.

/ITK/src/Modules/ThirdParty/GoogleTest/src/itkgoogletest/googletest/include/gtest/gtest.h:1527:11: error: ambiguous overload for 'operator==' (operand types are 'const itk::FixedArray<double, 2>' and 'const itk::Point<float, 2>')
 1527 |   if (lhs == rhs) {
      |       ~~~~^~~~~~
In file included from /ITK/src/Modules/Core/Common/include/itkMatrix.h:21,
                 from /ITK/src/Modules/Core/Transform/include/itkMatrixOffsetTransformBase.h:23,
                 from /ITK/src/Modules/Core/Transform/test/itkMatrixOffsetTransformBaseGTest.cxx:20:
/ITK/src/Modules/Core/Common/include/itkPoint.h:127:3: note: candidate: 'bool itk::Point<TCoordRep, NPointDimension>::operator==(const Self&) const [with TCoordRep = float; unsigned int NPointDimension = 2; itk::Point<TCoordRep, NPointDimension>::Self = itk::Point<float, 2>]' (reversed)
  127 |   operator==(const Self & pt) const
      |   ^~~~~~~~
In file included from /ITK/src/Modules/Core/Common/include/itkFixedArray.h:454,
                 from /ITK/src/Modules/Core/Common/include/itkNumericTraits.h:1986,
                 from /ITK/src/Modules/Core/Common/include/itkPoint.h:22,
                 from /ITK/src/Modules/Core/Common/include/itkMatrix.h:21,
                 from /ITK/src/Modules/Core/Transform/include/itkMatrixOffsetTransformBase.h:23,
                 from /ITK/src/Modules/Core/Transform/test/itkMatrixOffsetTransformBaseGTest.cxx:20:
/ITK/src/Modules/Core/Common/include/itkFixedArray.hxx:66:1: note: candidate: 'bool itk::FixedArray<TValue, VLength>::operator==(const itk::FixedArray<TValue, VLength>&) const [with TValue = double; unsigned int VLength = 2]'
   66 | FixedArray<TValue, VLength>::operator==(const FixedArray & r) const
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~

N-Dekker: > It's my first non-trivial pull-request to ITK and I was a bit uncertain as to how much should be said.

I think your pull-request is quite good already 😃 Now I'm still wondering, would it be necessary to have an `#if __cplusplus < 202002L` for each and every `operator!=`, as long as we still support C++14 and C++17. (We only just dropped C++11!)

Would the C++20 error also occur if  the `operator!=` overloads were defined as `friend` functions (inside the class definitions), instead of member functions? Like this:

    friend bool
    operator!=(const FixedArray & lhs, const FixedArray & rhs)
    {
      return !(lhs == rhs);
    }

In general, it appears preferable to define such operators as `friend` functions instead of member functions anyway.

However, if such `friend` functions can't fix the C++20 errors, another approach could be to add a macro, like `ITK_DEFINE_INEQUAL_OPERATOR_MEMBER_FUNCTION(className)`, which may be defined as "nothing" for C++20. Such a macro could be added to "ITK\Modules\Core\Common\include\itkMacro.h" What do you think?

darrent1974: I'm not really sure at the moment 😊

Firstly, I ONLY got the ambiguity error for the itkFixedArray/itkPoint classes for "==" operator, which makes me wonder if the error may have been localized to this particular instance. It seemed happy enough with the existing != operator definition. I got rid of it by choice as I thought it was no longer necessary due to the new handling of !=.

The implication I get from the several articles ([1](https://andreasfertig.blog/2020/06/cpp20-inside-of-a-spaceship/) [2](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0515r3.pdf)) I've now read on the new comparison operators is that it appears to negate the need to define "friend" operators due to the new reversible nature of the equality operator. 

So, does that potentially require all != operator definitions to be blanked for C++20 along the lines of a macro as you propose? I'm not familial enough with the ITK codebase overall to know how big a job this may be. 

Also, if my interpretation of the "friend" issue is accurate - could all friend definitions and any potential ordering operators <,> etc also be revisited to simplify for C++20. 

Or, will everything keep working as expected as is, with only the occasional spot fire popping up here and there due to edge cases.

N-Dekker:
@darrent1974 Can you please remove the "COMP: Modified ==/!= operators for C++20" commit from this pull request? I feel that it needs some further investigation! My pull request #2592 is inspired by your commit 😄  

The other two commits ("COMP: Removed constructor template parameters...") look fine to me now, so we may then quickly merge those two already.

Regarding  "COMP: Modified ==/!= operators for C++20", I'm not yet 100% sure if it's OK to just remove both `operator!=` and `operator==` from `itk::Point`. I understand that  `itk::Point` will then  still implicitly inherit those operators from `itk::FixedArray`. Would that allow comparisons between a point and any other `itk::FixedArray` _derived_ object (like `itk::Vector`)? If so, I'm just not 100% sure if that's OK, from the point of view of software design.

darrent1974:
@N-Dekker, I'm prepared to remove "COMP: Modified ==/!= operators for C++20". However, I'm going to submit it as a new draft pull request, In this new one I'll change the != behavior to use your new macro as per [#2592](https://github.com/InsightSoftwareConsortium/ITK/pull/2592)

However, I still believe it is valid in the context of C++20 == operators use a single parent == definition on itk::FixedArray that will be used in derived classes even for comparisons between derived objects. I do note that there is some precedent for using the following in derived classes:

 bool
 operator==(const Self & v) const
 {
   return Superclass::operator==(v);
 }

However, I'm not convinced if it is truly warranted unless a true specialization is required, but if this is required to be retained for consistency or other general style in ITK then I'm not fussed (unless it can be shown to introduce problems). 

I did some quick hacking with stripped down definitions of FixedArray, Point and Vector which illustrate some interesting behavior, wrt to operator reversibility and ambiguity. 

Have a look at [https://godbolt.org/z/dqE7zW59c](https://godbolt.org/z/dqE7zW59c)

I may not be able to make all the changes to the pull request in the next day or so. We've just come out of a COVID lockdown and I have a backlog of other things to attend to 😄

N-Dekker:
> I'm prepared to remove "COMP: Modified ==/!= operators for C++20".

Thanks @darrent1974 A little remark. I see you have added another commit that reverts this one. So that your pull request now consists on four commits. Could you possibly still remove both the "COMP: Modified ==/!=" commit and its revert commit, so that this commits would end up having only two commits? (The two commits about constructor template parameters)  If you don't have time to do so, shall I try to do it? It would make this pull request still easier to get merged.

> However, I'm going to submit it as a new draft pull request, In this new one I'll change the != behavior to use your new macro as per #2592

Sounds cool, but please double check my pull request #2592, as it does already use the macro in Common header files. So I think it overlaps (at least partially) with your  "COMP: Modified ==/!=" commit.

> Have a look at https://godbolt.org/z/dqE7zW59c

Interesting, apparently the original ITK code already supported `p == v` for an `itk::Point`and an `itk::Vector`. I'm not sure if that was originally intended!

> We've just come out of a COVID lockdown and I have a backlog of other things to attend to 😄

I'm glad to hear you also got out of the lockdown. Even if the extra freedom leaves you less time to do programming  ;-)

N-Dekker (approved):
@darrent1974  Thank you for the very first commits to pave the way from C++14 to C++20  🎉 These two "constructor template parameters" commits are clearly non-controversial. Hope we can soon also get the slightly more challenging operator==/operator!= problem fixed! 

@thewtex As you asked me to review: Yes, I do approve  😃

N-Dekker:
@darrent1974 I will try to get this little pull request of yours merged tomorrow. Unless someone else already merges today 😄

thewtex:
Thanks @N-Dekker @darrent1974 !

N-Dekker:
@darrent1974 Hope you're still enjoying the lockdown relaxations 😃  Can you possibly tell us whether the extensive removal of `operator!=` member functions (replacing them by ITK_UNEQUAL_OPERATOR_MEMBER_FUNCTION macro calls) at the master branch of ITK has helped to enable C++20 on gcc-11?

Do you think it is still necessary to remove `operator==` member functions from classes that are derived from `itk::FixedArray`, in order to support C++20?

darrent1974:
@N-Dekker Yes! I've been quite busy - we actually bought a new house a few days ago which has kept me occupied.

I've gotten back on this today and have attempted a rebuild with gcc-11 with all the ITK_UNEQUAL_OPERATOR_MEMBER_FUNCTION changes included.

I do still think that `operator==` member function should at least be removed from `itk::Point` to permit error-free compilation as with tests enabled, the following ambiguity error is still raised:

/ITK/src/Modules/ThirdParty/GoogleTest/src/itkgoogletest/googletest/include/gtest/gtest.h:1527:11: error: ambiguous overload for 'operator==' (operand types are 'const itk::FixedArray<double, 2>' and 'const itk::Point<float, 2>')
 1527 |   if (lhs == rhs) {

Additionally I'm getting a couple of compile errors in other sections relating to the use of ITK_UNEQUAL_OPERATOR_MEMBER_FUNCTION that you may want to investigate first as you've been working on it. I think both cases just need some extra conditional removal for C++20.

1. Doesn't like the dangling member template parameter defined above ITK_UNEQUAL_OPERATOR_MEMBER_FUNCTION, ie 
 /ITK/src/Modules/Core/Common/include/itkWeakPointer.h:80:42: error: expected unqualified-id before ';' token
   80 |   ITK_UNEQUAL_OPERATOR_MEMBER_FUNCTION(R);
   
2. When building tests, the following error is raised. 
/ITK/src/Modules/Core/Common/include/itkConstShapedNeighborhoodIterator.h:284:29: error: 'operator!=' has not been declared in 'using Superclass = class itk::NeighborhoodIterator<itk::Image<long int, 1>, itk::ZeroFluxNeumannBoundaryCondition<itk::Image<long int, 1>, itk::Image<long int, 1> > >'
  284 |   using Superclass::operator!=;


I think it may be best if I add a new WIP pull request for the proposed `itk::Point` change after the you have a look at the above two issues.

What do you think?

Darren

N-Dekker:
> @N-Dekker we actually bought a new house a few days ago

@darrent1974 Congratulations! I wish you all the best with your new house!

> gtest.h:1527:11: error: ambiguous overload for 'operator==' (operand types are 'const itk::FixedArray<double, 2>' and 'const itk::Point<float, 2>')
> 1527 | if (lhs == rhs) {

I guess this must be triggered by an `ASSERT_EQ(someFixedArray, somePoint)` or an `EXPECT_EQ(someFixedArray, somePoint)`, in one of the ITK GoogleTest unit tests, right?  Can you see which ITK unit test triggered this error? Could it be a workaround to pass the arguments in reverse order, `ASSERT_EQ(somePoint, someFixedArray)`?

>  1. Doesn't like the dangling member template parameter defined above ITK_UNEQUAL_OPERATOR_MEMBER_FUNCTION, ie
>        itkWeakPointer.h:80:42: error: expected unqualified-id before ';' token

Oops, my bad! Could this be fixed by restoring the original `WeakPointer::operator!=` definition? From https://github.com/InsightSoftwareConsortium/ITK/blob/v5.2.0/Modules/Core/Common/include/itkWeakPointer.h#L79-L84 Would GCC C++20 accept the original `WeakPointer::operator!=`?


>  2. When building tests, the following error is raised.
>        /Modules/Core/Common/include/itkConstShapedNeighborhoodIterator.h:284:29: error: 'operator!=' has not been declared in 'using Superclass = class itk::NeighborhoodIterator<itk::Image<long int, 1>, itk::ZeroFluxNeumannBoundaryCondition<itk::Image<long int, 1>, itk::Image<long int, 1> > >'
>        284 |   using Superclass::operator!=;

Would GCC C++20 be happy if this particular `using Superclass::operator!=` would not be there at all?


> I think it may be best if I add a new WIP pull request for the proposed `itk::Point` change after the you have a look at the above two issues.
> 
> What do you think?

Sounds like a plan  😃

darrent1974:
@N-Dekker, thanks for your wishes - it's quite an exciting time.

> Would GCC C++20 be happy if this particular using Superclass::operator!= would not be there at all?

Yes, as a test I commented it out - compiled fine.

> I guess this must be triggered by an `ASSERT_EQ(someFixedArray, somePoint) `or an `EXPECT_EQ(someFixedArray, somePoint)` in one of the ITK GoogleTest unit tests, right? Can you see which ITK unit test triggered this error? Could it be a workaround to pass the arguments in reverse order,` ASSERT_EQ(somePoint, someFixedArray)`?

Yes, it occurs in itkMatrixOffsetTransformBaseGTest.cxx in: 

```
// Checks the object created by MatrixOffsetTransformBase::New().
TEST(MatrixOffsetTransformBase, CheckNew)
{
  Check_New_MatrixOffsetTransformBase<float, 2>();
  Check_New_MatrixOffsetTransformBase<double, 3>();
}
```
With the error being thrown in Check_New_MatrixOffsetTransformBase() by the specific line:

```
  EXPECT_EQ(zeroFilledFixedArray, transformBase->GetCenter());
```

Unfortunately, in this case reversing the order of the arguments does not resolve the ambiguity error. 
Given the now reversible nature of primary operators such as `==` in C++20 I think such comparisons should now be coded in a way that they expected to be order neutral. 

At this point, to get this test to compile in C++20 would require the proposed `itk::Point` change.

thewtex:
Added to `release` for v5.2.1.

PR #2590, opened 2021-06-09
BUG: Support double colons in Changelog commit summary

### LINKED ISSUE #2558: Add support for colons in the changelog summary
As reported in:

https://discourse.itk.org/t/itk-5-2-0-has-been-released/4165/3

subjects with `::` get cut off in the summary.

The script used is here:

https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/Maintenance/AuthorsChangesSince.py

@N-Dekker

thewtex:
Closes #2558


dzenanz (commented):
LGTM

jhlegarreta (approved):
Have not tried, but looks good. Thanks Matt !

N-Dekker:
Have not tried, but does it also support a single colon in a commit message?

thewtex:
> Have not tried, but does it also support a single colon in a commit message?

Yes :-)

Utilities/Maintenance/AuthorsChangesSince.py
```diff
@@ -186,25 +186,25 @@ def formatted_current_author():
             prefix = line.split(":")[0].strip()
             commit = line.split(":")[-1]
             if prefix == "BUG":
-                description = line.split(":")[1]
+                description = line.partition(":")[2].rpartition(":")[0].strip()
```
N-Dekker, line 189: Does Python support writing helper functions? If so, I guess this may be a candidate! `make_description (const string & line)` 😸

N-Dekker: Can you possibly tell me what a line of the git log looks like? There's always a colon at the beginning and a colon at the end, right? Do you have an example?

Leengit:
What's the goal here?  For example, would
```
    ENH: Replace ":" with ";".   
```
(with leading and trailing spaces) be an acceptable commit message, and how do we want it parsed?  With this PR's changes, it gives
```python
prefix = 'ENH'
commit = '" with ";".   '
description = 'Replace "'
```

Leengit:
```python
prefix = line.split(":")[0].strip()
description = ":".join(line.split(":")[1:]).strip()
```
gives
```python
prefix = 'ENH'
description = 'Replace ":" with ";".'
```

N-Dekker:
@Leengit I just had a closer look. I'm not sure but it looks like the lines are produced by https://github.com/InsightSoftwareConsortium/ITK/blob/bf7600b317928ca7fd3fe997974ea1ebd34e635e/Utilities/Maintenance/AuthorsChangesSince.py#L221 

    git shortlog --format=%s:%h --topo-order --no-merges {git_revision}

For example:
>       BUG: "tmpImageIndex" not initialized to 0.:d829adebe8

So it appears that there is an extra colon inserted at the end. Does that clarify the matter?

Leengit:
Thank you @N-Dekker.  If I am understanding correctly, we want to strip off the trailing colon and hashtag.  For the likes of:
```
line = '   ENH: Replace ":" with ";".   :abcd123   '
```
we could use: 
```python
prefix = line.split(":")[0].strip()
description = ":".join(line.split(":")[1:-1]).strip()
hash = line.split(":")[-1].strip()
```
giving:
```
prefix = 'ENH'
description = 'Replace ":" with ";".'
hash = 'abcd123'
```

thewtex:
This implementation functions as desired, extracting the summary statement, without the prefix or the commit.

Utilities/Maintenance/AuthorsChangesSince.py
N-Dekker: > There is not a significant improvement by creating a function.

Well, personally I do think a `make_description` function could in this case improve readability and maintainability, as I see the same `description =` line of code repeated six times now. But I leave it up to you if you think otherwise.

N-Dekker (approved):
Thanks @thewtex ! I'm glad that you fixed this issue!

Leengit:
We could compute `description` just once, alongside `prefix` and `commit` before the tightest `if` statement begins.

Utilities/Maintenance/AuthorsChangesSince.py
thewtex: Maybe defining a function elsewhere could make sense when it is used in various places, but I do not think the additional indirection and complexity significantly improves readability and readability in this case.

thewtex:
> We could compute description just once, alongside prefix and commit before the tightest if statement begins.

Yes, I will do that.

PR #2591, opened 2021-06-10
COMP: Require Apple Clang 7.0.2 as minimum in itkCompilerChecks.cmake

N-Dekker:
Following commit https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/commit/f65f786d883a4de033da118908cf06aacb9e7c2b
"ENH: Update Xcode Apple Clang compiler for nightly dashboard"

The new version numbers are based on feedback from Sean McBride (@seanm), at:

Pull request "ENH: Update compiler versions for C++14"
https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/pull/156

Discussion topic "May we start using C++14 at the master branch?"
https://discourse.itk.org/t/may-we-start-using-c-14-at-the-master-branch/4172/23?u=niels_dekker

### DIFF
--- a/CMake/itkCompilerChecks.cmake
+++ b/CMake/itkCompilerChecks.cmake
@@ -10,10 +10,10 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND
   message(FATAL_ERROR "LLVM Clang 3.4 or later is required.")
 endif ()
 
-# Minimum compiler version check: Apple Clang >= 5.1 (Xcode 5.1)
+# Minimum compiler version check: Apple Clang >= 7.0.2 (Xcode 7.2.1)
 if (CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" AND
-    CMAKE_CXX_COMPILER_VERSION VERSION_LESS 5.1)
-  message(FATAL_ERROR "Apple Clang 5.1 or later is required.")
+    CMAKE_CXX_COMPILER_VERSION VERSION_LESS 7.0.2)
+  message(FATAL_ERROR "Apple Clang 7.0.2 or later is required.")
 endif ()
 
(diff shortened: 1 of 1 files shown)


CMake/itkCompilerChecks.cmake
```diff
@@ -10,10 +10,10 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND
   message(FATAL_ERROR "LLVM Clang 3.4 or later is required.")
 endif ()
 
-# Minimum compiler version check: Apple Clang >= 5.1 (Xcode 5.1)
+# Minimum compiler version check: Apple Clang >= 7.0.2 (Xcode 7.2.1)
```
dzenanz, line 13: The [release notes](https://developer.apple.com/library/archive/releasenotes/DeveloperTools/RN-Xcode/Chapters/Introduction.html#//apple_ref/doc/uid/TP40001051-CH1-SW876), under sub-heading "Apple LLVM Compiler Version 6.1" say "This updated compiler includes full support for the C++14 language standard". So it may not be necessary to bump to version 7.

thewtex (approved):
Thanks, @N-Dekker !

CMake/itkCompilerChecks.cmake
N-Dekker: @dzenanz Strictly speaking I think you're right, in the sense that some Apple Clang versions before 7.0.2 may already support C++14. But I understood from Sean (@seanm) that it should be sufficient for the latest ITK version to support Apple Clang >= 7.0.2.

Note that back in December 2019 ("before corona"), https://github.com/InsightSoftwareConsortium/ITK/blob/333bfc690dca71a63ea884b0ac42eaa63e62e2c3/Documentation/SupportedCompilers.md already said so:

> AppleClang 7.0.2 and later (from Xcode 7.2.1) should be supported

OK?

dzenanz (approved):

PR #2592, opened 2021-06-10
ENH: Add `ITK_UNEQUAL_OPERATOR_MEMBER_FUNCTION`

N-Dekker:
Follow-up to pull request #2589 "C++20 gcc-11 compatibility", by Darren Thompson (@darrent1974)

### DIFF
--- a/Modules/Core/Common/include/itkConstNeighborhoodIterator.h
+++ b/Modules/Core/Common/include/itkConstNeighborhoodIterator.h
@@ -388,14 +388,7 @@ public:
     return it.GetCenterPointer() == this->GetCenterPointer();
   }
 
-  /** Returns a boolean != comparison of the memory addresses of the center
-   * elements of two ConstNeighborhoodIterators of like pixel type and
-   * dimensionality.  The radii of the iterators are ignored. */
-  bool
-  operator!=(const Self & it) const
-  {
-    return it.GetCenterPointer() != this->GetCenterPointer();
-  }
+  ITK_UNEQUAL_OPERATOR_MEMBER_FUNCTION(Self);
 
   /** Returns a boolean < comparison of the memory addresses of the center
    * elements of two ConstNeighborhoodIterators of like pixel type and
--- a/Modules/Core/Common/include/itkMacro.h
+++ b/Modules/Core/Common/include/itkMacro.h
@@ -394,6 +394,20 @@ namespace itk
     static_assert(false, "Replace deprecated ITK_DISALLOW_COPY_AND_ASSIGN with modern ITK_DISALLOW_COPY_AND_MOVE")
 #endif
 
+// Note: The following macro, ITK_UNEQUAL_OPERATOR_MEMBER_FUNCTION(TypeName),
+// is only for internal use within the implementation of ITK. It may be
+// modified, renamed, or removed without backward compatibility support.
+#if __cplusplus < 202002L
+// For C++14 and C++17, this macro defines an operator!= member function that
+// just calls the corresponding operator== member function.
+#  define ITK_UNEQUAL_OPERATOR_MEMBER_FUNCTION(TypeName)                                                               \
+    bool operator!=(const TypeName & other) const { return !(this->operator==(other)); }                               \
+    ITK_MACROEND_NOOP_STATEMENT
+#else
+// With C++20, operator!= is defined implicitly, calling the corresponding operator==.
+#  define ITK_UNEQUAL_OPERATOR_MEMBER_FUNCTION(TypeName) ITK_MACROEND_NOOP_STATEMENT
+#endif
+
 /** Macro used to add standard methods to all classes, mainly type
  * information. */
 #define itkTypeMacro(thisClass, superclass)                                                                            \
--- a/Modules/Core/Common/include/itkVersor.hxx
+++ b/Modules/Core/Common/include/itkVersor.hxx
@@ -143,14 +143,6 @@ Versor<T>::operator/(const Self & v) const
 /** Comparison operator */
 template <typename T>
 bool
-Versor<T>::operator!=(const Self & v) const
-{
-  return !(*this == v);
-}
-
-/** Comparison operator */
-template <typename T>
-bool
 Versor<T>::operator==(const Self & v) const
 {
   // Evaluate the quaternion ratio between them
--- a/Modules/Core/Common/include/itkAutoPointer.h
+++ b/Modules/Core/Common/include/itkAutoPointer.h
@@ -159,12 +159,7 @@ public:
     return (void *)m_Pointer == (void *)r.m_Pointer;
   }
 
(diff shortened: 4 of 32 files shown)


Modules/Core/Common/include/itkConstNeighborhoodIterator.h
```diff
@@ -388,14 +388,7 @@ class ITK_TEMPLATE_EXPORT ConstNeighborhoodIterator
...
 
-  /** Returns a boolean != comparison of the memory addresses of the center
-   * elements of two ConstNeighborhoodIterators of like pixel type and
-   * dimensionality.  The radii of the iterators are ignored. */
-  bool
-  operator!=(const Self & it) const
-  {
-    return it.GetCenterPointer() != this->GetCenterPointer();
-  }
+  ITK_DEFINE_INEQUAL_OPERATOR_MEMBER_FUNCTION(Self);
```
dzenanz, line 391: Should we retain docstrings?

Modules/Core/Common/include/itkVersor.hxx
```diff
@@ -140,14 +140,6 @@ Versor<T>::operator/(const Self & v) const
   return result;
 }
 
-/** Comparison operator */
```
dzenanz, line 143: This is just removed, not replaced by the new macro. Is this intentional?

Modules/Core/Common/include/itkMacro.h
```diff
@@ -394,6 +394,15 @@ namespace itk
     static_assert(false, "Replace deprecated ITK_DISALLOW_COPY_AND_ASSIGN with modern ITK_DISALLOW_COPY_AND_MOVE")
 #endif
 
+#if __cplusplus < 202002L
+#  define ITK_DEFINE_INEQUAL_OPERATOR_MEMBER_FUNCTION(TypeName)                                                        \
```
dzenanz, line 398: I think the proper word is `unequal`, not `inequal`. See https://www.merriam-webster.com/words-at-play/unequal-vs-inequality-negative-prefix-usage.

Leengit: FWIW, https://en.cppreference.com/w/cpp/language/operator_comparison and https://www.cplusplus.com/doc/tutorial/operators/ say the operator's name is "not equal to".

Modules/Core/Common/include/itkMacro.h
```diff
@@ -394,6 +394,15 @@ namespace itk
...
+    bool operator!=(const Self & other) const { return !(this->operator==(other)); }                                   \
```
Leengit, line 399: With
```c++
bool operator!=(const Self & rhs) const { return !(*this == rhs); }
```
the macro might also work for `friend` equality operators.

Modules/Core/Common/include/itkConstNeighborhoodIterator.h
N-Dekker: Personally I would prefer to simply remove the (existing) documentation of `operator!=`, whenever it just matches the documentation of the corresponding `operator==`, without adding any extra information. As in this case:

https://github.com/InsightSoftwareConsortium/ITK/blob/v5.2.0/Modules/Core/Common/include/itkConstNeighborhoodIterator.h#L382-L384

Versus

https://github.com/InsightSoftwareConsortium/ITK/blob/v5.2.0/Modules/Core/Common/include/itkConstNeighborhoodIterator.h#L391-L393

I wouldn't mind adding a "standard" comment to the macro call, though, if you want to. For example:

     /** Defines operator!= simply by calling operator== */
     ITK_DEFINE_INEQUAL_OPERATOR_MEMBER_FUNCTION(Self);

What do you think?

Modules/Core/Common/include/itkVersor.hxx
N-Dekker: Thanks for reviewing so carefully, but yes, it's intended 😸 The macro call defines `operator!=`as a simple inline one-liner, in the `h` file. So the `operator!=` implementation may just be removed from the corresponding `hxx` file.

Modules/Core/Common/include/itkMacro.h
N-Dekker: Good idea @Leengit However, I want to just make one step at a time. This may be the next step.

My initial aim is just to make C++20 happy (PR #2589 by @darrent1974), while preserving the original (ITK 5.2.0) behavior as much as possible.

Ideally I would like to move entirely from "member function" equality operators to `friend` equality operator, but I don't know if it is feasible. It would help if an automatic tool like clang-tidy could do the job!

Modules/Core/Common/include/itkMacro.h
N-Dekker: @Leengit Thanks, but sorry, Dženan was first 😸 By the way, I also like `NONEQUAL` and `NON_EQUAL`. Maybe we should make the macro name user-configurable  ;-)

N-Dekker: @dzenanz Please check the new name   :-)

Modules/Core/Common/include/itkConstNeighborhoodIterator.h
dzenanz: > adding a "standard" comment to the macro

This is probably better.

dzenanz (approved):

N-Dekker:
I think the pull request is technically OK now, but I'm not 100% sure if the macro should be an official part of the ITK API (like `ITK_DISALLOW_COPY_AND_MOVE`), or just an internal implementation detail. Or maybe just considered a temporary workaround, aiming to offer preliminary C++20 support. What do you think @dzenanz  Shall I add a note that the macro is experimental?

dzenanz:
As this change is not exactly necessary, I am not sure we should complicate our lives with it. When C++20 becomes ITK's minimum (5-6 years from now sounds like a long time into the future), we could simply remove all of the `!=` operators which trivially call `==` operators. Meanwhile, keep current `!=` operators as-is, without adding this macro.

If we do add the macro, definitely mark it as internal-only, temporary, or something similar.

N-Dekker:
> When C++20 becomes ITK's minimum (5-6 years from now sounds like a long time into the future), we could simply remove all of the `!=` operators which trivially call `==` operators.

@dzenanz Thanks, but I think this change is quite nice already, as it removes a lot of boilerplate code (reducing the code by more than 160 lines). Indeed it would not be necessary anymore once C++20 becomes ITK's minimum, but that may still take a long time, as you say. Until then, more people will try to build ITK using C++20, even while C++14 is the minimum. For them the macro would be especially helpful  😄   

> If we do add the macro, definitely mark it as internal-only, temporary, or something similar.

I did, just by adding a note in the comment in itkMacro.h. Please check!

dzenanz (approved):

blowekamp (approved):
LGTM. I like the reduction in code bulk this achieves too.

PR #2594, opened 2021-06-14
BUG: Use consistent ITK_USE_CLANG_FORMAT variable name

thewtex:
Across ITKClangFormatSetup.cmake and ITKModuleClangFormat.cmake.


dzenanz (approved):

jhlegarreta (approved):

jhlegarreta:
Sorry for my late thought: now I'm wondering whether there will be some clash across the *two* variables, since they are named the same, but look as if they are defined in two different places. Shouldn't the variable be defined at a single place and be re-used by the other file?

thewtex:
They are in two places so the one file can be used independently from the other, e.g. ITK versus ITK external module build.

CMake/ITKClangFormatSetup.cmake
```diff
@@ -55,7 +55,7 @@ endif()
 if(CLANGFORMAT_EXECUTABLE)
   mark_as_advanced(CLANGFORMAT_EXECUTABLE)
 endif()
-if ( ITK_USE_CLANGFORMAT AND NOT EXISTS "${CLANGFORMAT_EXECUTABLE}")
+if ( ITK_USE_CLANG_FORMAT AND NOT EXISTS "${CLANGFORMAT_EXECUTABLE}")
```
N-Dekker, line 58: `CLANGFORMAT_EXECUTABLE`? Or `CLANG_FORMAT_EXECUTABLE`? 😺

thewtex: We could use `CLANG_FORMAT` consistently.

dzenanz (approved):

hjmjohnson (approved):

PR #2597, opened 2021-06-16
STYLE: Use `ITK_UNEQUAL_OPERATOR_MEMBER_FUNCTION` in Modules/Core

N-Dekker:
Replaced hand-written definitions of `operator!=` member functions by
`ITK_UNEQUAL_OPERATOR_MEMBER_FUNCTION` macro calls, in Modules/Core.

Did some minor adjustments to corresponding `operator==` member
functions, where necessary, and removed non-const operator overloads.

Follow-up to commit 0ad6ef1b558f162e83562a3d420f24aca38fa8d8
"STYLE: Use `ITK_UNEQUAL_OPERATOR_MEMBER_FUNCTION` in Core/Common"

PR #2600, opened 2021-06-17
STYLE: ITK_UNEQUAL_OPERATOR_MEMBER_FUNCTION when `==` just returns true

N-Dekker:
Replaced hand-written definitions of `operator!=` member functions by
`ITK_UNEQUAL_OPERATOR_MEMBER_FUNCTION` macro calls, and adjusted the
corresponding `operator==` member functions whenever those `==`
operators would simple just always return `true`. (These are member
functions of classes whose instances are all considered equal.)

Did find and replace in Visual Studio, using regular expressions like:

Find:

    bool\r?\n  operator!=\(const (.+) \&\) const\r?\n  {\r?\n    return false;\r?\n  }\r?\n\r?\n  bool\r?\n  operator==\(const (.+) \& other\) const\r?\n  {\r?\n    return !\(\*this != other\);\r?\n  }

Replace by:

    bool operator==(const $1 &) const { return true; }\r\n\r\ITK_UNEQUAL_OPERATOR_MEMBER_FUNCTION($1);

Follow-up to:
pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2597
commit 27f26685cc10be8bee74989a1c03639cd89705ce
"STYLE: Use ITK_UNEQUAL_OPERATOR_MEMBER_FUNCTION in Modules/Core"

### DIFF
--- a/Modules/Filtering/Convolution/include/itkMaskedFFTNormalizedCorrelationImageFilter.hxx
+++ b/Modules/Filtering/Convolution/include/itkMaskedFFTNormalizedCorrelationImageFilter.hxx
@@ -70,16 +70,12 @@ public:
   }
 
   bool
-  operator!=(const PostProcessCorrelation &) const
+  operator==(const PostProcessCorrelation &) const
   {
-    return false;
+    return true;
   }
 
-  bool
-  operator==(const PostProcessCorrelation & other) const
-  {
-    return !(*this != other);
-  }
+  ITK_UNEQUAL_OPERATOR_MEMBER_FUNCTION(PostProcessCorrelation);
 
   inline TImage
   operator()(const TImage & NCC, const TImage & denominator, const TImage & numberOfOverlapPixels) const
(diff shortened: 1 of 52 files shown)


dzenanz (approved):
Wow, there were so many instances of this. And with inverted logic!

N-Dekker:
> Wow, there were so many instances of this. And with inverted logic!

@dzenanz  Indeed! It looks like at that time the preference was to first define `operator!=`, and then define `operator==` in terms of `operator!=`. While C++20 does it exactly the other way, making  `operator==` a "primary", and   `operator!=` a "secondary" operator: https://brevzin.github.io/c++/2019/07/28/comparisons-cpp20/

When a class has an `operator==` that just returns true, maybe we should say it follows the _Animal Farm animal_ design pattern. You know, "All animals are equal..."  😸

Leengit (approved):

PR #2605, opened 2021-06-19
Use `ITK_UNEQUAL_OPERATOR_MEMBER_FUNCTION` in Numerics, Segmentation, Video 

N-Dekker:
And also used this macro in `UnsharpMaskingFunctor`, `SliceIterator`, and `TanHelper`.

Note: This should be the last "bulk PR" to add those `ITK_UNEQUAL_OPERATOR_MEMBER_FUNCTION` macro calls.

### DIFF
--- a/Modules/Core/Common/test/itkSimpleFilterWatcherTest.cxx
+++ b/Modules/Core/Common/test/itkSimpleFilterWatcherTest.cxx
@@ -31,20 +31,12 @@ public:
   TanHelper() = default;
   ~TanHelper() = default;
   bool
-  operator!=(const TanHelper & rhs) const
+  operator==(const TanHelper & rhs) const
   {
-    return this != &rhs;
+    return this == &rhs;
   }
 
-  bool
-  operator==(const TanHelper & other) const = delete;
-  /* NOTE: operator== NOT defined. It is not required
-   * as part of the defined specification for a Function.
-  bool operator==(const TanHelper & other) const
-  {
-    return !( *this != other );
-  }
-  */
+  ITK_UNEQUAL_OPERATOR_MEMBER_FUNCTION(TanHelper);
 
   inline TOutput
   operator()(const TInput & A) const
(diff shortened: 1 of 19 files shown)


Modules/Core/Common/test/itkSimpleFilterWatcherTest.cxx
```diff
@@ -31,20 +31,12 @@ class TanHelper
...
   bool
-  operator!=(const TanHelper & rhs) const
+  operator==(const TanHelper & rhs) const
   {
-    return this != &rhs;
+    return this == &rhs;
   }
 
-  bool
-  operator==(const TanHelper & other) const = delete;
```
N-Dekker, line 40: @hjmjohnson I see you added the `= delete` here almost three years ago, hope it's OK to you that this PR actually removes the `= delete`! I believe it's in general preferable for a class to define `operator==` (without `= delete`) whenever it defines `operator!=`. To simply make it an "equality comparable" type. (C++20 will automatically generate `operator!=` for a class that has `operator==`.)
https://github.com/InsightSoftwareConsortium/ITK/commit/dbcdf4bce2cf4e0e31526064f823a5ccd4454293#diff-8ff9e0510254bcf8e71ee15fa988a4eb4ea1f83b6a896bed25caa273108c8d6c

Leengit (approved):
Thank you.

dzenanz (approved):

Modules/Core/Common/test/itkSimpleFilterWatcherTest.cxx
hjmjohnson: @N-Dekker Your work here is great!  Thanks for updating and making the code consistent.  I don't know the complete history of why these were not implemented.  The "= delete" was a refactoring to explicitly remove use of functions that were previously not implemented.

Having proper implementations seems like a great idea.

Hans

hjmjohnson (approved):

PR #2606, opened 2021-06-21
BUG: Remove unused virtual operator!=, operator== from ColormapFunction

N-Dekker:
The `virtual` `ColormapFunction` member functions `operator!=` and
`operator==` appear incorrect, as they always just return `false` or
`true`, respectively, even while `ColormapFunction` has four values as
data members, and corresponding setters and getters.

The two member functions aren't used elsewhere in ITK, and they aren't
overridden by derived ITK classes. So instead of fixing them, it appears
preferable to just remove them.


dzenanz (approved):

thewtex (approved):

ntustison:
Sounds and looks good to me.

PR #2611, opened 2021-06-24
COMP: Remove MultiThreaderBase pixelCount

thewtex:
To address:

  /ITK/Modules/Core/Common/src/itkMultiThreaderBase.cxx:538:17: warning:
  variable 'pixelCount' set but not used [-Wunused-but-set-variable]
    SizeValueType pixelCount = 1;

### DIFF
--- a/Modules/Core/Common/src/itkMultiThreaderBase.cxx
+++ b/Modules/Core/Common/src/itkMultiThreaderBase.cxx
@@ -535,11 +535,6 @@ MultiThreaderBase::ParallelizeImageRegion(unsigned int                          
   }
   ProgressReporter progress(filter, 0, 1);
 
-  SizeValueType pixelCount = 1;
-  for (unsigned d = 0; d < dimension; ++d)
-  {
-    pixelCount *= size[d];
-  }
   struct RegionAndCallback rnc
   {
     funcP, dimension, index, size, 0, filter


Modules/Core/Common/src/itkMultiThreaderBase.cxx
```diff
@@ -535,11 +535,6 @@ MultiThreaderBase::ParallelizeImageRegion(unsigned int
...
   ProgressReporter progress(filter, 0, 1);
 
-  SizeValueType pixelCount = 1;
-  for (unsigned d = 0; d < dimension; ++d)
-  {
-    pixelCount *= size[d];
-  }
   struct RegionAndCallback rnc
   {
     funcP, dimension, index, size, 0, filter
```
N-Dekker, line 540: It appears that `pixelCount` was originally passed as initializer value to the `RegionAndCallback` object. Now it is just zero (`0`). With "Progress reporting revisionv5" https://github.com/InsightSoftwareConsortium/ITK/pull/1575 commit https://github.com/InsightSoftwareConsortium/ITK/pull/1575/commits/c8787b8ed8696d8ee6b0ddef0a377695921392f2 3 March 2020

If you think `pixelCount` may be removed here, I guess it should also be removed from the definition of `RegionAndCallback`, right?

https://github.com/InsightSoftwareConsortium/ITK/blob/v5.2.0/Modules/Core/Common/include/itkMultiThreaderBase.h#L444

thewtex: @dzenanz @blowekamp what is the intended approach here?

blowekamp: Hrmm... Just glancing through #1575 ( that is a big PR), I think a comment should have been left here as to why it is this way. It may be unfortunately used as a "magic" value to distinguished one reporting progress method vs another. Looking into how `RegionaAndCallback::pixelCount` is use ( or not used ) should clarify the purpose of the zero value or indicate it's not needed.

Either way the code is not used.

dzenanz (approved):
As the code is no longer used, it makes sense to remove it.

Modules/Core/Common/src/itkMultiThreaderBase.cxx
N-Dekker: Was the calculated `pixelCount` meant to be passed to RegionAndCallback, or not?

blowekamp: > Was the calculated `pixelCount` meant to be passed to RegionAndCallback, or not?

The current zero value is correct and working as far as I know. As I said if may be a "magic" value. 

The `pixelCount` computation can be removed.

N-Dekker:
> As the code is no longer used, it makes sense to remove it.

@dzenanz  OK, but then why not remove `pixelCount` from the RegionAndCallback struct definition, if it is always zero anyway?

blowekamp:
> > As the code is no longer used, it makes sense to remove it.
> 
> @dzenanz OK, but then why not remove `pixelCount` from the RegionAndCallback struct definition, if it is always zero anyway?

I see it only usage left here: https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/src/itkMultiThreaderBase.cxx#L569

Zero may intentionally be passed to the TotalProgressReporter so that all completed chunk could result in a progress event. Unfortunately, this feature is not documented in the TotalProgress reporter or it's usage :(

dzenanz:
> remove pixelCount from the RegionAndCallback

I guess we could and should do it. I guess the reason this was not cleaned up during progress refactoring is because that PR was getting really big (and probably the author getting tired of it 😄).

> Zero may intentionally be passed to the TotalProgressReporter so that all completed chunk could result in a progress event.

I guess we should update the documentation of `TotalProgressReporter` during removal of pixelCount from the RegionAndCallback.

PR #2618, opened 2021-06-25
COMP: Update GDCM

### LINKED ISSUE #2617: GCC 11.1.0 fails to compile current GDCM
Error:

```cpp
In file included from Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmImageChangePhotometricInterpretation.cxx:14:
Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmImageChangePhotometricInterpretation.h:67:64: error: ‘numeric_limits’ is not a member of ‘std’
   67 |   return v < 0 ? 0 : (v > std::numeric_limits<T>::max() ? std::numeric_limits<T>::max() : v);

```

It is fixed upstream: https://github.com/malaterre/GDCM/commit/4404b770be337bd0d5d3c2289abfd34426433db2

So, we just need to update the third party, PR incoming!

phcerdan:
Update GDCM to fix errors compiling with `gcc` 11

Fix #2617

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -52,13 +52,15 @@ Modules/Filtering/Denoising/include/itkPatchBasedDenoisingImageFilter.hxx hooks-
 Documentation/ReleaseNotes/* hooks-max-size=300000
 Utilities/Doxygen/doxygen.config.in hooks-max-size=150000
 
-Modules/ThirdParty/GDCM/src/gdcm/Source/DataDictionary/gdcmPrivateDefaultDicts.cxx hook-max-size=1021501
-Modules/ThirdParty/GDCM/src/gdcm/Utilities/gdcmopenjpeg/src/lib/openjp2/j2k.c hook-max-size=654063
-Modules/ThirdParty/GDCM/src/gdcm/Source/InformationObjectDefinition/Part3.xml hook-max-size=4336000
-Source/InformationObjectDefinition/Part3.xml hook-max-size=4336000
+Modules/ThirdParty/GDCM/src/gdcm/Source/DataDictionary/gdcmPrivateDefaultDicts.cxx hooks-max-size=1021501
+Modules/ThirdParty/GDCM/src/gdcm/Source/DataDictionary/gdcmDefaultDicts.cxx hooks-max-size=1200000
+Source/DataDictionary/gdcmDefaultDicts.cxx hooks-max-size=1200000
+Modules/ThirdParty/GDCM/src/gdcm/Utilities/gdcmopenjpeg/src/lib/openjp2/j2k.c hooks-max-size=654063
+Modules/ThirdParty/GDCM/src/gdcm/Source/InformationObjectDefinition/Part3.xml hooks-max-size=4336000
+Source/InformationObjectDefinition/Part3.xml hooks-max-size=4336000
 
 # Avoid ghostflow-check-master failures
-Modules/ThirdParty/HDF5/src/itkhdf5/src/H5C.c      hook-max-size=400000
-Modules/ThirdParty/HDF5/src/itkhdf5/src/H5Dchunk.c hook-max-size=400000
-Modules/ThirdParty/HDF5/src/itkhdf5/src/H5Shyper.c hook-max-size=500000
-Modules/ThirdParty/HDF5/src/itkhdf5/src/H5Tconv.c  hook-max-size=500000
+Modules/ThirdParty/HDF5/src/itkhdf5/src/H5C.c      hooks-max-size=400000
+Modules/ThirdParty/HDF5/src/itkhdf5/src/H5Dchunk.c hooks-max-size=400000
+Modules/ThirdParty/HDF5/src/itkhdf5/src/H5Shyper.c hooks-max-size=500000
+Modules/ThirdParty/HDF5/src/itkhdf5/src/H5Tconv.c  hooks-max-size=500000
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,7 +29,7 @@ foreach(pnew "") # Currently Empty
     cmake_policy(SET ${pnew} NEW)
   endif()
 endforeach()
-foreach(pold "CMP0120") # CMP0120 will require non-trivial effort to address
+foreach(pold "CMP0120" "CMP0091") # CMP0120 will require non-trivial effort to address
   if(POLICY ${pold})
     cmake_policy(SET ${pold} OLD)
   endif()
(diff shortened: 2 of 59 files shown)


dzenanz (commented):
LGTM, but:

ghostflow-check-master errors:

commit c9a27c3 creates blob d5b218569316b6c637edc33d253bc5251439a863 at Source/DataDictionary/gdcmDefaultDicts.cxx with size 1194766 bytes (1166.76 KiB) which is greater than the maximum size 1048576 bytes (1024.00 KiB). If the file is intended to be committed, set the hooks-max-size attribute on its path.
commit 26f747b creates blob d5b218569316b6c637edc33d253bc5251439a863 at Modules/ThirdParty/GDCM/src/gdcm/Source/DataDictionary/gdcmDefaultDicts.cxx with size 1194766 bytes (1166.76 KiB) which is greater than the maximum size 300000 bytes (292.97 KiB). If the file is intended to be committed, set the hooks-max-size attribute on its path.
commit 26f747b creates blob d5b218569316b6c637edc33d253bc5251439a863 at Modules/ThirdParty/GDCM/src/gdcm/Source/DataDictionary/gdcmDefaultDicts.cxx with size 1194766 bytes (1166.76 KiB) which is greater than the maximum size 300000 bytes (292.97 KiB). If the file is intended to be committed, set the hooks-max-size attribute on its path.
Please rewrite commits to fix the errors listed above (adding fixup commits will not resolve the errors) and force-push the branch again to update the merge request.

phcerdan:
I just modified .gitattributes for those paths [here](https://github.com/InsightSoftwareConsortium/ITK/pull/2618/commits/2d7d0488092692eae5a435f58fa1624b6857a357) ghostflow didn't seem to notice...

phcerdan:
Oh, not sure if it is hooks-max-size or hook-max-size, both of them can be found in gitattributes. I am going to try with the plural.

.gitattributes
```diff
@@ -53,6 +53,8 @@ Documentation/ReleaseNotes/* hooks-max-size=300000
 Utilities/Doxygen/doxygen.config.in hooks-max-size=150000
 
 Modules/ThirdParty/GDCM/src/gdcm/Source/DataDictionary/gdcmPrivateDefaultDicts.cxx hook-max-size=1021501
+Modules/ThirdParty/GDCM/src/gdcm/Source/DataDictionary/gdcmDefaultDicts.cxx hook-max-size=1200000
```
dzenanz, line 56: Should this be `hook-max-size` or `hooks-max-size`?

phcerdan: That is what I am testing, `hook-max-size` was used already in .gitattributes

phcerdan: Yeah, it's the plural, I am fixing all of them `hook`->`hooks`

However, still one ghostflow error that I think must be fixed upstream.

dzenanz: Option 1: merge with the known upstream problem, and fix in upstream later.
Option 2: fix in upstream first, then re-run the update here in ITK.
Option 3: ignore the upstream problem 😞

dzenanz (approved):

.gitattributes
phcerdan: I don't think is upstream, upstream doesn't know about hooks-max-size.

I am trying to modify the "local" to gdcm .gitattributes for that path, but doesn't seem to work.

dzenanz: The one you are modifying only applies after merging. The GDCM subtree before the merge only cares about subtree's `.gitattributes`, which come from [the upstream](https://github.com/malaterre/GDCM/blob/master/.gitattributes).

phcerdan: I see, however, ITK does modify that file on top of the upstream .gitattributes. See https://github.com/InsightSoftwareConsortium/ITK/blob/42719cad6626ca623b62b94f1e961ffeb69e4d2e/Modules/ThirdParty/GDCM/src/gdcm/.gitattributes

dzenanz: Maybe the paths there should not start with a `/`?

phcerdan: I tried with no effect.

dzenanz: I am out of ideas then. Maybe @bradking can provide some insight?

bradking: The ghostflow check is complaining about the *import* commit, not the after-merged file.  The import script itself needs to add a `.gitattributes` entry for the path.

bradking: In other words, the import script needs to pretend that the upstream content had the `.gitattributes` entry, but actually add it itself before creating the import commit.

phcerdan: Thanks @bradking, so if I understood correctly, we can maybe hack it in the UpdateFromUpstream.sh script:

```bash
extract_source () {
    git_archive
    pushd "${extractdir}/${name}-reduced"
    echo "* -whitespace" > .gitattributes
    echo "Source/DataDictionary/gdcmDefaultDicts.cxx hooks-max-size=1200000" > .gitattributes
    popd
}

```

dzenanz: > PR upstream adding the path to .gitattributes

I think that is the best solution, which we can do later if it will take a long time for that round-trip.

bradking: Setting the attribute in upstream will look strange because it has no need for this.

Several of our import scripts in ITK, VTK, CMake, etc. all have little attribute modification steps like this.  The `>>.gitattributes` line looks fine.

phcerdan: > The >>.gitattributes line looks fine.

I am missing then how to avoid the merge conflict. It seems a trivial one.. but maybe there is a merge strategy set up?

bradking: Don't add the attribute on the ITK side ahead of time.  Or, if it is there already, manually resolve the conflict once.  Future updates should merge cleanly.

phcerdan: Fixing the trivial merge conflict manually worked. Thanks Brad.

phcerdan:
Ok, a summary on how to solve ghostflow errors, kudos to @bradking for the hints.

For the after merge commit errors, those with the full path from ITK top directory, we have to add the patsh to the top [.gitattributes](https://github.com/phcerdan/ITK/blob/update_gdcm/.gitattributes).

For the import commits errors, those with the path relative to the ThirdParty being updated, we have to add those relative paths to the `UpdateFromUpstream.sh` script. This might generate a merge conflict when running the Update script, but just resolve manually (easy).

dzenanz (approved):

thewtex:
Backported to `release` for ITK v5.2.1.

PR #2633, opened 2021-07-06
ENH: Add MakePoint, MakeVector, MakeIndex, MakeSize function templates

N-Dekker:
Added a templated generalization of the overloaded `MakePoint`,
`MakeVector`, `MakeIndex`, and `MakeSize`, functions from
"itkGTestTypedefsAndConstructors.h", allowing them to be more widely
used.

Placed the function templates directly in the namespace `itk`.

### DIFF
--- a/Modules/Core/Common/include/itkPoint.h
+++ b/Modules/Core/Common/include/itkPoint.h
@@ -348,6 +348,26 @@ swap(Point<TCoordRep, NPointDimension> & a, Point<TCoordRep, NPointDimension> & 
   a.swap(b);
 }
 
+
+/** Makes a Point object, having the specified values as coordinates. */
+template <typename TValue, typename... TVariadic>
+auto
+MakePoint(const TValue firstValue, const TVariadic... otherValues)
+{
+  // Assert that the other values have the same type as the first value.
+  const auto assertSameType = [](const auto value) {
+    static_assert(std::is_same<decltype(value), const TValue>::value, "Each value must have the same type!");
+    return true;
+  };
+  const bool assertions[] = { true, assertSameType(otherValues)... };
+  (void)assertions;
+  (void)assertSameType;
+
+  constexpr unsigned                  dimension{ 1 + sizeof...(TVariadic) };
+  const std::array<TValue, dimension> stdArray{ { firstValue, otherValues... } };
+  return Point<TValue, dimension>{ stdArray };
+}
+
 } // end namespace itk
 
 #ifndef ITK_MANUAL_INSTANTIATION
--- a/Modules/Core/Common/include/itkSize.h
+++ b/Modules/Core/Common/include/itkSize.h
@@ -20,8 +20,8 @@
 
 #include "itkIntTypes.h"
 #include "itkMacro.h"
-#include <algorithm> // For copy_n.
-#include <type_traits>
+#include <algorithm>   // For copy_n.
+#include <type_traits> // For is_integral.
 #include <memory>
 
 namespace itk
@@ -470,6 +470,20 @@ swap(Size<VDimension> & one, Size<VDimension> & two)
 {
   std::swap(one.m_InternalArray, two.m_InternalArray);
 }
+
+
+/** Makes a Size object, having the specified size values. */
+template <typename... T>
+auto
+MakeSize(const T... values)
+{
+  const auto toValueType = [](const auto value) {
+    static_assert(std::is_integral<decltype(value)>::value, "Each value must have an integral type!");
+    return static_cast<SizeValueType>(value);
+  };
+  return Size<sizeof...(T)>{ { toValueType(values)... } };
+}
+
 
 // static constexpr definition explicitly needed in C++11
 template <unsigned int VDimension>
--- a/Modules/Core/Common/include/itkVector.h
+++ b/Modules/Core/Common/include/itkVector.h
@@ -328,6 +328,26 @@ swap(Vector<T, NVectorDimension> & a, Vector<T, NVectorDimension> & b)
   a.swap(b);
 }
 
+
+/** Makes a Vector object, having the specified values as coordinates. */
+template <typename TValue, typename... TVariadic>
+auto
+MakeVector(const TValue firstValue, const TVariadic... otherValues)
+{
+  // Assert that the other values have the same type as the first value.
+  const auto assertSameType = [](const auto value) {
+    static_assert(std::is_same<decltype(value), const TValue>::value, "Each value must have the same type!");
+    return true;
+  };
+  const bool assertions[] = { true, assertSameType(otherValues)... };
+  (void)assertions;
+  (void)assertSameType;
+
+  constexpr unsigned                  dimension{ 1 + sizeof...(TVariadic) };
+  const std::array<TValue, dimension> stdArray{ { firstValue, otherValues... } };
+  return Vector<TValue, dimension>{ stdArray };
+}
+
 } // end namespace itk
 
 #ifndef ITK_MANUAL_INSTANTIATION
--- a/Modules/Core/Common/include/itkIndex.h
+++ b/Modules/Core/Common/include/itkIndex.h
@@ -19,6 +19,7 @@
 #define itkIndex_h
 
 #include "itkOffset.h"
+#include <type_traits> // For is_integral.
 
 namespace itk
 {
@@ -567,6 +568,20 @@ swap(Index<VDimension> & one, Index<VDimension> & two)
 {
   std::swap(one.m_InternalArray, two.m_InternalArray);
 }
+
+
+/** Makes an Index object, having the specified index values. */
+template <typename... T>
+auto
+MakeIndex(const T... values)
+{
+  const auto toValueType = [](const auto value) {
+    static_assert(std::is_integral<decltype(value)>::value, "Each value must have an integral type!");
+    return static_cast<IndexValueType>(value);
+  };
+  return Index<sizeof...(T)>{ { toValueType(values)... } };
+}
+
 
 // static constexpr definition explicitly needed in C++11
 template <unsigned int VDimension>
--- a/Modules/Core/Common/test/itkIndexGTest.cxx
+++ b/Modules/Core/Common/test/itkIndexGTest.cxx
@@ -19,6 +19,7 @@
 // First include the header file to be tested:
 #include "itkIndex.h"
 #include <gtest/gtest.h>
+#include <initializer_list>
 #include <limits>
 
 namespace
@@ -55,4 +56,18 @@ TEST(Index, FilledReturnsIndexWithSpecifiedValueForEachElement)
   // Test for 1-D and 3-D.
   Expect_Filled_returns_Index_with_specified_value_for_each_element<1>();
   Expect_Filled_returns_Index_with_specified_value_for_each_element<3>();
+}
+
+
+TEST(Index, Make)
+{
+  static_assert((decltype(itk::MakeIndex(1, 1))::Dimension == 2) && (decltype(itk::MakeIndex(1, 1, 1))::Dimension == 3),
+                "The dimension of the created itk::Size should equal the number of arguments");
+
+  EXPECT_EQ(itk::MakeIndex(0, 0), itk::Index<2>());
+  EXPECT_EQ(itk::MakeIndex(0, 0, 0), itk::Index<3>());
+
+  const auto itkIndex = itk::MakeIndex(1, 2, 3, 4);
+  const auto values = { 1, 2, 3, 4 };
+  EXPECT_TRUE(std::equal(itkIndex.begin(), itkIndex.end(), values.begin(), values.end()));
 }
--- a/Modules/Core/Common/test/itkPointGTest.cxx
+++ b/Modules/Core/Common/test/itkPointGTest.cxx
@@ -20,8 +20,10 @@
 #include "itkPoint.h"
 #include <gtest/gtest.h>
 
-#include <iterator> // For begin and end.
-#include <numeric>  // For iota.
+#include <initializer_list>
+#include <iterator>    // For begin and end.
+#include <numeric>     // For iota.
+#include <type_traits> // For is_same.
 
 namespace
 {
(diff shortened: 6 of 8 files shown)


N-Dekker:
@blowekamp Do you agree that these function templates could replace the functions you wrote for "itkGTestTypedefsAndConstructors.h"? At https://github.com/InsightSoftwareConsortium/ITK/blob/v5.2.0/Modules/Core/TestKernel/include/itkGTestTypedefsAndConstructors.h

blowekamp:
> @blowekamp Do you agree that these function templates could replace the functions you wrote for "itkGTestTypedefsAndConstructors.h"? At https://github.com/InsightSoftwareConsortium/ITK/blob/v5.2.0/Modules/Core/TestKernel/include/itkGTestTypedefsAndConstructors.h

Yes, this much better and take advantage of the C+11 features that were not available when originally written. However, there are a couple concerns:
 - These implementations implicitly does explicit type conversion for the arguments, which might hide some unintended conversion issues?
 - Should these procedural methods be in sub-namespace? I believe I raised the same question with the addition of `itk::WriteImage`.

Having written this, the methods were originally written to provide consistent initialization between these ITK array-like data types. IMHO, it'd be "best" if there were some uniform constructions to these classes so these procedures were not needed. When compared to `std::make_unique`, that has some legitimate purpose to implicitly deduce the class template parameter for the procedural argument.

dzenanz (approved):

Modules/Core/Common/include/itkSize.h
```diff
@@ -471,6 +471,16 @@ swap(Size<VDimension> & one, Size<VDimension> & two)
   std::swap(one.m_InternalArray, two.m_InternalArray);
 }
 
+
+/** Makes a Size object, having the specified size values. */
+template <typename... T>
+constexpr auto
+MakeSize(const T... values)
+{
+  return Size<sizeof...(T)>{ { static_cast<SizeValueType>(values)... } };
```
N-Dekker, line 480: @blowekamp You wrote:
> These implementations implicitly does explicit type conversion for the arguments, which might hide some unintended conversion issues?

The `static_cast<SizeValueType>` here in `MakeSize` allows users to pass `int` values, as in `MakeSize(1, 2, 3)`.

We could remove this `static_cast<SizeValueType>`, but then the user would have to write something like `MakeSize(1U, 2U, 3U)`, to avoid signed-to-unsigned conversion. What do you think?

N-Dekker:
@blowekamp 
> Should these procedural methods be in sub-namespace? 

In my opinion, it is appropriate to have them in the same namespace as `itk::Point`, `itk::Vector`, `itk::Index`, and `itk::Size` themselves. As they are very much like an extra set of "constructors" for those types. However, we could temporarily place them in `itk::Experimental`, if you like that better... do you?

Modules/Core/Common/include/itkPoint.h
```diff
@@ -348,6 +348,17 @@ swap(Point<TCoordRep, NPointDimension> & a, Point<TCoordRep, NPointDimension> &
...
+/** Makes a Point object, having the specified values as coordinates. */
+template <typename TValue, typename... TVariadic>
+auto
+MakePoint(const TValue firstValue, const TVariadic... otherValues)
+{
+  constexpr unsigned                  dimension{ 1 + sizeof...(TVariadic) };
+  const std::array<TValue, dimension> stdArray{ { firstValue, otherValues... } };
+  return Point<TValue, dimension>{ stdArray };
+}
+
```
Leengit, line 361: I would prefer to not infer the type from the first argument.  We could instead require that the `TValue` type be explicitly supplied with something like the following.  In particular, the user would thus not be surprised that the supplied values are cast to the explicitly supplied type.
```c++
template <typename TValue, typename... TVariadic>
constexpr auto
MakePoint(const TVariadic... values)
{
  constexpr unsigned                  dimension{ sizeof...(TVariadic) };
  const std::array<TValue, dimension> stdArray{ { static_cast<TValue>(values)... } };
  return Point<TValue, dimension>{ stdArray };
}

// Example:
auto p = MakePoint<unsigned>(1, 2);
```

Leengit: If there's a programmer out there who wants to count the arguments too, we could _additionally_ supply something like:
```c++
template <typename TValue, unsigned int NDimension, typename... TVariadic>
auto
MakePoint(const TVariadic... values)
{
  const std::array<TValue, NDimension> stdArray{ { static_cast<TValue>(values)... } };
  return Point<TValue, NDimension>{ stdArray };
}

// Example #2
auto p = MakePoint<unsigned, 2>(1, 2);
```

dzenanz: That would make this convenience function rather inconvenient.

Leengit: > That would make this convenience function rather inconvenient.

Agreed to some extent ... our `MakePoint` would then become less like `std::make_pair` or `std::make_typle` which don't ask us to specify the types, and more like `std::make_unique`, which does ask us to specify the type.  Either way our `MakePoint` has the convenience of allowing us to specify a number of arguments equal to the dimension instead of a fixed number of arguments.

blowekamp: The other [implementation](https://github.com/InsightSoftwareConsortium/ITK/blob/v5.2.0/Modules/Core/TestKernel/include/itkGTestTypedefsAndConstructors.h#L49-L54) referenced the itk::ImageBase class for the dimension, and just itk::Point needed to get the component type from the itk::ImageBase class. 


Why is `std::initializer_list` bad for constructors of these objects? This use case here is manually programmed arguments which are not time critical.

N-Dekker: @Leengit @blowekamp @dzenanz Thanks for the discussion so far. These are indeed the fundamental design decisions behind this little pull request!

I would very much prefer to allow inferring the point type from the arguments of `MakePoint`, like `std::make_pair` and `std::make_tuple`. This appears most convenient to me. And it also seems unnecessarily redundant to me, having to specify the value type when it can be deduced automatically.

Note that the current pull request does _allow_ specifying the value type, if the user really wants to: 

    // Specifying <double> is allowed here, but not necessary.
    assert( itk::MakePoint<double>(1.0, 2.0) == itk::MakePoint(1.0, 2.0) );

However, to avoid any confusion I think `MakePoint` should require that all arguments have the same type.  I'm still looking into how to write such a requirement. (Same requirement should then apply to `MakeVector`, of course.)

Regarding `MakeIndex` and `MakeSize`, it might be nice to check if all arguments are integers. To allow `MakeSize(2, 3)`, but not `MakeSize(2.0F, 3.0F)`, and certainly not `MakeSize(0.5, 1.5)`.

Instead of variadic arguments, `std::initializer_list` could have been used, but then compile-time deduction of the Dimension would not be possible. So then, users would have to explicitly specify the Dimension, for example, by calling `itk::MakePoint<4>( { 1, 2, 3, 4 } )`. Which could be implemented as follows:

    template <unsigned NDimension, typename TValue>
    itk::Point<TValue, NDimension>
    MakePoint(const std::initializer_list<TValue> initializerList)
    {
      assert(initializerList.size() <= NDimension);
      Point<TValue, NDimension> point{};
      std::copy(initializerList.begin(), initializerList.end(), point.begin());
      return point;
    }

Personally I would prefer to support inferring the Dimension automatically, for `MakePoint`. However it might still be nice to have a `Point::Point(std::initializer_list<ScalarValue>)` constructor added to `itk::Point`. What do you think?

Leengit: > However, to avoid any confusion I think `MakePoint` should require that all arguments have the same type. I'm still looking into how to write such a requirement. (Same requirement should then apply to `MakeVector`, of course.)

If it is worth the time to try and debug ... I like the first recommended solution at [this stack overflow page](https://stackoverflow.com/questions/3703658/specifying-one-type-for-all-arguments-passed-to-variadic-function-or-variadic-te).  It would use a helper struct, which I have called `type_t`, and might replace our use of `auto` like this:
```c++
template <typename T1, typename...>
struct type_t
{
  using type = T1;
};

template <typename TValue, typename... TVariadic>
constexpr type_t<Point<TValue, 1 + sizeof...(TVariadic)>,
                 std::enable_if<is_same<TVariadic, TValue>::value>::type...>::type
MakePoint(const TValue firstValue, const TVariadic... otherValues)
{
   // The body from before
}
```
Notes: We may want `is_convertible` instead of `is_same`.  The `constexpr` in the return type may or may not be achievable, and can be omitted if it is not.  If we are still supporting C++11 we might have to define `type_t` differently -- see note at [implementation of `void_t`](https://en.cppreference.com/w/cpp/types/void_t) referring to [CWG1558](https://wg21.cmeerw.net/cwg/issue1558).

blowekamp: Just continuing to brain storm other options.... Most of the time these make methods will be used with respect to an Image which dictates the dimension and type. For those who prefer to be more type specifying, could we do some specialization so that the following would work: `MakePoint<ImageType>(1.0, 1.1, 1.2)`

>However, to avoid any confusion I think MakePoint should require that all arguments have the same type. I'm still looking into how to write such a requirement. (Same requirement should then apply to MakeVector, of course.)

agreed.

> Regarding MakeIndex and MakeSize, it might be nice to check if all arguments are integers. To allow MakeSize(2, 3), but not MakeSize(2.0F, 3.0F), and certainly not MakeSize(0.5, 1.5).
 agreed.

>Instead of variadic arguments, std::initializer_list could have been used, but then compile-time deduction of the Dimension would not be possible. So then, users would have to explicitly specify the Dimension, for example, by calling itk::MakePoint<4>( { 1, 2, 3, 4 } ). 

Thanks for the explanation. However, my suggestion was to add `initializer_list` to the constrictor of the these types so that we could do:
```
  SizeType sz({1,2,3});
  IndexType idx{{0,-1,2});
  PointType pt({1.1, 2.3, 4.0});
```

I don't think adding the following auto-typing style is to useful:
```
 auto pt = MakePoint(1.1, 2.3, 4.0);
```

N-Dekker: Thanks for your interesting suggestion @Leengit Did you already give it a try?

Personally I think SFINAE (`enable_if`) isn't really necessary here. SFINAE is especially useful to exclude certain template instantiations from the overload set. But in this pull request there is one `itk::MakePoint`. I think it would be clearer to just produce a compile-time assert failure, when the argument types don't match.

The following seems to work fine, inside the implementation of `MakePoint(const TValue firstValue, const TVariadic... otherValues)`:

    const auto assertSameType = [](const auto value) {
      static_assert(std::is_same<decltype(value), const TValue>::value, "Each value must have the same type!");
      return true;
    };
    const bool assertions[] = { true, assertSameType(otherValues)... };

Would that also be OK to you?

BTW, ITK has just dropped C++11 support at the master branch. C++14 features are available to us now 😃

Leengit: I have not tried the SFINAE version from stack overflow.  The `static_assert` approach looks good; the error message that we get to supply is a better bet than the SFINAE error message, which could be fairly cryptic.

N-Dekker: FYI Here an example where I found those `Make` functions especially useful: https://github.com/SuperElastix/elastix/blob/a505befa8cf59b97b8e54c5d672c704062066534/Core/Main/GTesting/itkTransformixFilterGTest.cxx#L428-L435 Specifically:

    itkTransform->SetTranslation(MakeVector(1.0, -2.0, 3.0));
    itkTransform->SetCenter(MakePoint(3.0, 2.0, 1.0));

dzenanz:
Is this ready for merging?

N-Dekker:
> Is this ready for merging?

Thanks for asking @dzenanz  As submitter of a PR, when I feel unconfident about a PR of mine, I mark it "draft" or "WIP". There appears no way in GitHub to explicitly approve my own PR. So if a PR of mine is _not_ "draft" or "WIP", then I _think_ it's ready.

So in this case, yes, I think it's ready. 😄

Modules/Core/Common/include/itkVector.h
```diff
@@ -328,6 +328,26 @@ swap(Vector<T, NVectorDimension> & a, Vector<T, NVectorDimension> & b)
...
+MakeVector(const TValue firstValue, const TVariadic... otherValues)
+{
+  // Assert that the other values have the same type as the first value.
+  const auto assertSameType = [](const auto value) {
+    static_assert(std::is_same<decltype(value), const TValue>::value, "Each value must have the same type!");
+    return true;
+  };
+  const bool assertions[] = { true, assertSameType(otherValues)... };
+  (void)assertions;
+  (void)assertSameType;
```
N-Dekker, line 344: I found it a bit disappointing that it took me so much code to simply just assert that all values have the same type, but for now I think it's good enough. Approved 👍

N-Dekker (commented):
Approved 👍

Modules/Core/Common/include/itkPoint.h
N-Dekker: Bradley (@blowekamp) wrote:
> Thanks for the explanation. However, my suggestion was to add `initializer_list` to the constrictor of the these types so that we could do:
> 
> ```
>   SizeType sz({1,2,3});
>   IndexType idx{{0,-1,2});
>   PointType pt({1.1, 2.3, 4.0});
> ```

For the record, `SizeType sz({1,2,3})` and `IndexType idx{{0,-1,2}}` do already compile well, assuming both have `Dimension` >= 3. For `PointType pt({1.1, 2.3, 4.0})` to compile, an extra `itk::Point` constructor should be added. Otherwise, just call `itk::MakePoint(1.1, 2.3, 4.0)`  😃

blowekamp: Additionally, have only 2/3 of these common types support the format can be problematic as they are not uniform. I apologized for not digging into the git history, and PR requests, but there has been some non-portable compiler issues with some of the those two constructors.   So, I am going to say: I believe they are not tested, so they are most likely slightly broken :)

N-Dekker: Well, even old C++98/03 compiler did support `SizeType sz = {{1,2,3}}` on ITK4, right? :)

PR #2638, opened 2021-07-08
Update KWSys to 2021-07-08 (d5a1dc68)

### LINKED ISSUE #2637: ITK's copy of KWSys is returning `0` as `std::string`, forbidden by C++23
I work on MSVC's C++ Standard Library implementation, and we regularly build ITK (along with many other open-source projects) to identify compiler/library bugs that would break your project, so we can fix them before shipping. This also allows us to provide advance notice of source-breaking changes - which is the case here.

The paper [P2166R1](https://wg21.link/P2166R1) "Prohibiting `basic_string` And `basic_string_view` Construction From `nullptr`" has been accepted for the upcoming C++23 Standard, and we recently implemented it by merging https://github.com/microsoft/STL/pull/1995 from our contributor @sam20908. Our open-source test pass then discovered the following code in ITK's copy of KWSys that is now disallowed by C++23:

https://github.com/InsightSoftwareConsortium/ITK/blob/bdad19f67c63c299025f3ea3e41b7bca0d26951f/Modules/ThirdParty/KWSys/src/KWSys/SystemTools.cxx#L4215-L4216
https://github.com/InsightSoftwareConsortium/ITK/blob/bdad19f67c63c299025f3ea3e41b7bca0d26951f/Modules/ThirdParty/KWSys/src/KWSys/SystemTools.cxx#L4241
https://github.com/InsightSoftwareConsortium/ITK/blob/bdad19f67c63c299025f3ea3e41b7bca0d26951f/Modules/ThirdParty/KWSys/src/KWSys/SystemTools.cxx#L4380
https://github.com/InsightSoftwareConsortium/ITK/blob/bdad19f67c63c299025f3ea3e41b7bca0d26951f/Modules/ThirdParty/KWSys/src/KWSys/SystemTools.cxx#L4387

These lines attempting to `return 0;` as `std::string` have always been undefined behavior according to C++98 through C++20 (as `string`'s constructor has a precondition that the pointer is non-null). In C++23, `basic_string` is gaining a constructor overload from `nullptr_t` that's deleted, to statically detect and reject attempted construction from `nullptr`. Here, since `0` is a null pointer constant, overload resolution is ambiguous between `const char *` and the deleted `nullptr` constructor, so it's an error (albeit with a less-clear message than if the argument were `nullptr`):

```
error C2440: 'return': cannot convert from 'int' to 'std::basic_string<char,std::char_traits<char>,std::allocator<char>>'
note: No constructor could take the source type, or constructor overload resolution was ambiguous
```

There are several possible ways to fix this code; `return "";`, `return {};`, or `return std::string{};` would all be valid ways of returning an empty string, that are backwards-compatible with earlier versions of C++.

I'm reporting this here because https://gitlab.kitware.com/utils/kwsys says:

> KWSys has no independent issue tracker.  After encountering an issue (bug) please submit a patch using the instructions for Contributing. Otherwise please report the issue to the tracker for the project that hosts the copy of KWSys in which the problem was found.

dzenanz:
Closes #2637.


Modules/Core/Common/src/itkDirectory.cxx
```diff
@@ -58,7 +58,7 @@ Directory::PrintSelf(std::ostream & os, Indent indent) const
 bool
 Directory::Load(const char * dir)
 {
-  return m_Internal->Load(dir);
+  return static_cast<bool>(m_Internal->Load(dir));
```
N-Dekker, line 61: It might be cleared to explicitly compare to the "success" status:

    return m_Internal->Load(dir) == itksys::Status::Success()

dzenanz: I tried this first! But there is no `operator==`.

thewtex (approved):

PR #2641, opened 2021-07-13
STYLE: Use MakePoint, MakeVector, MakeIndex, MakeSize from itk namespace

N-Dekker:
Replaced calls to the `Make` functions from the namespace
`itk::GTest::TypedefsAndConstructors` by calls to the function templates
from `itk`, that were added by:

pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2633
commit 8aed68490b733ba45b36b2e8e84e99980db45948
"ENH: Add MakePoint, MakeVector, MakeIndex, MakeSize function templates"

### DIFF
--- a/Modules/Core/Transform/test/itkBSplineTransformGTest.cxx
+++ b/Modules/Core/Transform/test/itkBSplineTransformGTest.cxx
@@ -111,14 +111,14 @@ TEST(ITKBSplineTransform, Copying_Clone)
 
   ASSERT_EQ(coeffImageArray.Size(), 2);
 
-  SizeType      imageSize = MakeSize(10, 10);
+  SizeType      imageSize = itk::MakeSize(10, 10);
   DirectionType imageDirection; // filled with zeros
   imageDirection(0, 1) = -1;
   imageDirection(1, 0) = 1;
 
-  VectorType imageSpacing = MakeVector(1.1, 1.2);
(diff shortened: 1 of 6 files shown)


N-Dekker:
@blowekamp Is it OK to you if we just use the new `itk::Make` function templates now, instead of the original `Make` functions (from `itk::GTest::TypedefsAndConstructors`) that you wrote before?

blowekamp (approved):
LGTM

dzenanz (approved):
Should we remove the duplicated methods/functions from `itk::GTest::TypedefsAndConstructors`?

dzenanz:
I guess end-users could access them, but `GTest` sub-namespace implied they are for tests. Not many new tests were written even within ITK. Maybe ask on the forum whether anyone used them? If not (I assume so), we could just remove them and add a not in the migration guide. If somebody did use them, I guess we could
```c++
#if defined(legacy)
  using itk::MakeIndex; // make it also available in itk::GTest namespace
```

N-Dekker:
> Maybe ask on the forum whether anyone used them? If not (I assume so), we could just remove them and add a not in the migration guide.

@dzenanz Shall we then first merge this single-commit PR, and have a new separate PR which removes the original `Make` functions? The new PR could then be discussed at discourse.

PR #2646, opened 2021-07-14
BUG: Resolve warnings with MakeIndex and MakePoint

tbirdso:
Resolves nightly exotic build warnings relating to `itk::Index` and `itk::Point` initialization by relying on `itk::MakeIndex` and `itk::MakePoint` helpers.

[Build information](https://open.cdash.org/viewBuildError.php?type=1&buildid=7339831):
Site: RogueResearch7
Build Name: Mac10.11-AppleClang-dbg-x86_64-static
Build Time: 2021-07-14 05:03:37

Warnings resolved:
> [CTest: warning matched] /Users/builder/externalModules/Filtering/LabelMap/test/itkStatisticsLabelMapFilterGTest.cxx:201:30: warning: suggest braces around initialization of subobject [-Wmissing-braces]
  image->SetPixel(IndexType{ 0, 0 }, 32000);
...

> CTest: warning matched] /Users/builder/externalModules/Video/Core/test/itkImageToVideoFilterTest.cxx:122:49: warning: suggest braces around initialization of subobject [-Wmissing-braces]
        ITK_TEST_EXPECT_EQUAL(frame->GetPixel({ idx[1], idx[2] }), it.Get());
...

> [CTest: warning matched] /Users/builder/externalModules/IO/NIFTI/test/itkNiftiImageIOTest12.cxx:60:43: warning: suggest braces around initialization of subobject [-Wmissing-braces]
  image->SetOrigin(ImageType::PointType({ -7.0, -13.0, -19.0 }));

## PR Checklist

### DIFF
--- a/Modules/Filtering/LabelMap/test/itkStatisticsLabelMapFilterGTest.cxx
+++ b/Modules/Filtering/LabelMap/test/itkStatisticsLabelMapFilterGTest.cxx
@@ -198,15 +198,15 @@ TEST_F(StatisticsLabelMapFixture, 2D_ones_with_outliers)
   image->FillBuffer(value);
 
   // Test with outliers outside the label.
-  image->SetPixel(IndexType{ 0, 0 }, 32000);
-  image->SetPixel(IndexType{ 0, 1 }, -32000);
+  image->SetPixel(itk::MakeIndex(0, 0), 32000);
+  image->SetPixel(itk::MakeIndex(0, 1), -32000);
 
 
   auto                  labelImage = Utils ::CreateLabelImage();
   Utils::LabelPixelType label = 1;
   labelImage->FillBuffer(label);
-  labelImage->SetPixel(IndexType{ 0, 0 }, 0);
-  labelImage->SetPixel(IndexType{ 0, 1 }, 0);
+  labelImage->SetPixel(itk::MakeIndex(0, 0), 0);
+  labelImage->SetPixel(itk::MakeIndex(0, 1), 0);
 
   Utils::LabelObjectType::ConstPointer labelObject = Utils::ComputeLabelObject(labelImage, image, label, 1 << 16);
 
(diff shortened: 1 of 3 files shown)


dzenanz (approved):

N-Dekker:
@tbirdso Thanks Thomas! I'm glad to see that `itk::MakeIndex` and `itk::MakePoint` can be useful this way, so quickly after being added to the master (pull #2633).  😃  

Could you possibly still mention the platform (Mac10.11-AppleClang) and the specific warning in the commit message?

> warning: suggest braces around initialization of subobject [-Wmissing-braces]

It looks like the warning could also be fixed by adding extra braces, but I'm fine with calling `MakeIndex` and `MakePoint`. 😃 

BTW, according to Clang bug report [Warning "missing braces around initializer" causing problems with std::array](https://bugs.llvm.org/show_bug.cgi?id=21629) Clang >= 6.0 should no longer produce this warning, in these cases.

dzenanz:
@N-Dekker it will be a while before we can require clang 6+.

N-Dekker:
> it will be a while before we can require clang 6+.

@dzenanz Well, until then I guess we would have to either add extra braces to `IndexType{1,2,3}`, by doing `IndexType{ {1,2,3} }` instead, or simply call `itk::MakeIndex(1,2,3)`.  😃 


BTW, for C++17, [clang 5](https://releases.llvm.org/5.0.0/tools/clang/docs/ReleaseNotes.html) should already be sufficient!

tbirdso:
@N-Dekker Thanks for adding these helpers in [PR 2633](https://github.com/InsightSoftwareConsortium/ITK/pull/2633)! I think they will be very useful 😃 

I've updated the commit message as you requested. Helpers were used here at @dzenanz 's suggestion, I agree that they'll help with readability and avoiding the braces warning going forward.

dzenanz:
@tbirdso you updated the PR description, not the commit message. Commit message is visible in git clients too, while PR description is only visible on GitHub.

seanm:
I've tried to turn that warning off:

https://open.cdash.org/build/7339873/notes

but it doesn't seem to work.  Not sure why.  If anyone knows how I should be specifying the flag in that file, that'd be great...

tbirdso:
@dzenanz  Commit message is now updated as well, I will amend my previous comment.

dzenanz:
> I've tried to turn that warning off

Now I remember you said that before 😄

N-Dekker:
> I've tried to turn that warning off:
> https://open.cdash.org/build/7339873/notes

@seanm  Note that the warning may still be useful when the aggregate (struct) has more than one data member.

Clang 6.0 did not entirely remove the warning. It only suppressed the warning for aggregates that have just one data member. As far as I understand!

Example, at https://godbolt.org/z/cheE68GxY :

```
struct MyAggregateStruct { int data[2]; int count; };

int MyFun(const MyAggregateStruct& ) { return 0; }

int main()
{
    return MyFun({1, 2});
}

```

Still yields the warning, as it should, even with clang 6+:

>:10:19: warning: suggest braces around initialization of subobject [-Wmissing-braces]
    return MyFun({1, 2});

N-Dekker (approved):
@tbirdso Thank you Tom!

seanm:
@dzenanz yes, your deja vu is correct.

@N-Dekker it's just on this old bot that I want to disable it.

thewtex (approved):
:green_heart:

Issue #1948, opened 2020-08-06
Improve Dask serializer registration


thewtex:
As suggested by @jakirkham in #1942, we can improve the registration of Dask serialization and deserialization functions, including making them available in `distribute`. This could be done with `NDArrayITKBase`, but it is probably worth re-visiting after we have serializer for common `itk.DataObject`'s.

As @jakirkham noted:

If one communicates objects that Dask already knows how to work with (like [NumPy arrays]( https://github.com/dask/distributed/blob/2.22.0/distributed/protocol/numpy.py )), then one will get optimized serialization for free. This may involve a little (or maybe no) code in ITK to make sure NumPy arrays are handled reasonable well and handed to Dask when functions return. The benefit here seems useful even outside the Dask context.

If one adds [pickle protocol 5]( https://docs.python.org/3/library/pickle.html#pickle-oob ) support to objects (possibly extending to older Python's with [`pickle5`]( https://github.com/pitrou/pickle5-backport ) or leveraging NumPy to do this indirectly ( https://github.com/numpy/numpy/pull/12091 )), then one can leverage Dask's own support for pickle protocol 5 ( https://github.com/dask/distributed/pull/3784 ) ( https://github.com/dask/distributed/pull/3849 ), which should yield a similar performance boost to Dask's own custom serialization (as the underlying principles are the same). This would require merely updating how pickling works in ITK. It also will work with anything else that supports pickle protocol 5.

Alternatively one can use [Dask's custom serialization]( https://distributed.dask.org/en/latest/serialization.html ) by following [this example on how to extend to custom types]( https://distributed.dask.org/en/latest/serialization.html#id3 ). The usual strategy here is to put these registration bits in its own module. This will soften the Distributed dependency and provide a single point for registering these functions. The latter is useful as one needs to [`import` this module in `distributed`]( https://github.com/dask/distributed/blob/2.22.0/distributed/protocol/__init__.py#L104-L113 ) for it to work. This is doable with a bit of work in ITK and Distributed, but not too difficult.

Any of these options seems fine. Possibly it's worth doing multiple or even all of them (this is what we did in RAPIDS https://github.com/rapidsai/cudf/pull/5139 ). Just trying to show what's possible and what it would entail to do 🙂

_Originally posted by @jakirkham in https://github.com/InsightSoftwareConsortium/ITK/pull/1942#issuecomment-668789613_

jakirkham:
[...]

Issue #1091, opened 2019-07-14
Support itk.Image use in Dask Array `map_blocks`


thewtex:
To avoid explicit type coersion in Dask Array `map_blocks`, consider adding

- [x] `dtype`
- [x] `shape`
- [x] `ndim`
- [ ] `__array_function__`

to `itk.Image`.

@jakirkham @mrocklin

thewtex:
`ndim`, `dtype`, `shape` added in #1780

jakirkham:
Nice! Thanks for the update Matt 😄

mrocklin:
Woot.  Have we tried this out in a workflow yet?  If not, it might be fun
to try something and throw it into a "Better ITK and Dask Integration"
blogpost.

thewtex:
@mrocklin great idea -- I'll see about following up on our initial post with some denoising...

GenevieveBuckley:
Is this something I can help along?

thewtex:
Hi @GenevieveBuckley,

Help on this would be amazing! :pray: 

We need to add pickle support to `itk.Image`.

The process may be a great example basis for a Dask Blog post on how to add required pickle support for data structures from native Python packages.

We need to get this test to pass:

```python
import itk
import pickle
import numpy as np

# Create an itk.Image
array = np.random.randint(0, 256, (8, 12)).astype(np.uint8)
image = itk.image_from_array(array)
image.SetSpacing([1.0, 2.0])
image.SetOrigin([11.0, 4.0])
theta = np.radians(30)
cosine = np.cos(theta)
sine = np.sin(theta)
rotation = np.array(((cosine, -sine), (sine, cosine)))
image.SetDirection(rotation)

# Verify serialization works with a np.ndarray
serialized = pickle.dumps(array)
deserialized = pickle.loads(serialized)
assert np.array_equal(array, deserialized)

# How to check for image consistency
image_copy = itk.image_duplicator(image)
compared = itk.comparison_image_filter(image, image_copy)
assert np.sum(compared) == 0

# We need to add support for this
serialized = pickle.dumps(image)
deserialized = pickle.loads(serialized)
compared = itk.comparison_image_filter(image, deserialized)
assert np.sum(compared) == 0
```

[...]

PR #2649, opened 2021-07-15
ENH: fix Python multi-processing hang on unix

### LINKED ISSUE #2069: ITK Python Pool Threader deadlocks with multiprocess module
ITK Python is deadlocking when using with heavy weight process in the `multiprocess` python module.

The following is simplified from  SimpleITK/SimpleITK#1239.

The below code produces the deadlock for everything we have testing except MacOS with Python 3.8.

ITK Python version 5.1.post1 was used. 

```
#!/usr/bin/env python

import multiprocessing as mp
import numpy as np
import os
import threading
import itk

num_images = 2


def f( fname ):
    image = itk.imread(fname)


    print("Resampling {} pid: {} {}...".format(fname, os.getpid(), threading.get_ident(),fname))
    resampled = itk.resample_image_filter(image, reference_image=image, use_reference_image=True)

    return image # resampled


if __name__ == '__main__':

    paths = []
    for i in range(num_images):
        image = sitk.Image(10,11,12, sitk.sitkUInt8)
        path = f'/tmp/image_{i}.nii'
        sitk.WriteImage(image, path)
        paths.append(path)

    #print("GlobalDefaultThreader: {}".format(sitk.ProcessObject_GetGlobalDefaultThreader()))
    #print("GlobalNumberOfThreads: {}".format(sitk.ProcessObject_GetGlobalDefaultNumberOfThreads()))

    print("main pid: {} {}".format(os.getpid(), threading.get_ident()))


    f(paths[0])

    with mp.Pool(1) as p:
        p.map(f, paths)
```

Setting the global default threader to "platform" does not produce a deadlock.
Setting the resample to only use 1 work unit does not produce a deadlock.

It may be interesting to test ITK 4.x with the threader explicitly set to the pool.

dzenanz:
Closes #2069.

_Initial (**old**) description:_

This turns a hang into:
```text
(pyEnv) dzenan@corista:~/nn_work_dir$ python itkMultiProcessingMONAI.py 
main pid: 2578395 140116265719616
Resampling image_0.nii pid: 2578395 140116265719616...
Resampling image_0.nii pid: 2578404 140116265719616...
Resampling image_1.nii pid: 2578404 140116265719616...
In ThreadPool::atfork_prepareIn ThreadPool::atfork_resumeTraceback (most recent call last):
  File "itkMultiProcessingMONAI.py", line 44, in <module>
    p.map(f, paths)
  File "/usr/lib/python3.8/multiprocessing/pool.py", line 364, in map
    return self._map_async(func, iterable, mapstar, chunksize).get()
  File "/usr/lib/python3.8/multiprocessing/pool.py", line 771, in get
    raise self._value
multiprocessing.pool.MaybeEncodingError: Error sending result: '[<itk.itkImagePython.itkImageUC3; proxy of <Swig Object of type 'itkImageUC3 *' at 0x7f6ee94955a0> >]'. Reason: 'TypeError("cannot pickle 'SwigPyObject' object")'
(pyEnv) dzenan@corista:~/nn_work_dir$
```

### DIFF
--- a/Modules/Core/Common/include/itkThreadPool.h
+++ b/Modules/Core/Common/include/itkThreadPool.h
@@ -124,13 +124,23 @@ auto result = pool->AddWork([](int param) { return param; }, 7);
   SetDoNotWaitForThreads(bool doNotWaitForThreads);
 
 protected:
-  /* We need access to the mutex in AddWork, and the variable is only
+  /** We need access to the mutex in AddWork, and the variable is only
    * visible in .cxx file, so this method returns it. */
   std::mutex &
   GetMutex();
 
   ThreadPool();
-  ~ThreadPool() override;
+
+  /** Stop the pool and release threads. To be called by the destructor and atfork. */
+  void
+  CleanUp();
+
+  ~ThreadPool() override { this->CleanUp(); }
+
+  static void
+  PrepareForFork();
+  static void
+  ResumeFromFork();
 
 private:
   /** Only used to synchronize the global variable across static libraries.*/
--- a/Modules/Core/Common/src/itkThreadPool.cxx
+++ b/Modules/Core/Common/src/itkThreadPool.cxx
@@ -77,6 +77,9 @@ ThreadPool::GetInstance()
       {
         new ThreadPool(); // constructor sets m_PimplGlobals->m_ThreadPoolInstance
       }
+#if defined(ITK_USE_PTHREADS)
+      pthread_atfork(ThreadPool::PrepareForFork, ThreadPool::ResumeFromFork, ThreadPool::ResumeFromFork);
+#endif
     }
   }
   return m_PimplGlobals->m_ThreadPoolInstance;
@@ -132,7 +135,8 @@ ThreadPool::GetNumberOfCurrentlyIdleThreads() const
   return int(m_Threads.size()) - int(m_WorkQueue.size()); // lousy approximation
 }
 
-ThreadPool::~ThreadPool()
+void
+ThreadPool::CleanUp()
 {
   {
     std::unique_lock<std::mutex> mutexHolder(m_PimplGlobals->m_Mutex);
@@ -154,6 +158,21 @@ ThreadPool::~ThreadPool()
   }
 }
 
+void
+ThreadPool::PrepareForFork()
+{
+  m_PimplGlobals->m_ThreadPoolInstance->CleanUp();
+}
+
+void
+ThreadPool::ResumeFromFork()
+{
+  ThreadPool * instance = m_PimplGlobals->m_ThreadPoolInstance.GetPointer();
+  ThreadIdType threadCount = instance->m_Threads.size();
+  instance->m_Threads.clear();
+  instance->m_Stopping = false;
+  instance->AddThreads(threadCount);
+}
 
 void
 ThreadPool::ThreadExecute()


Modules/Core/Common/src/itkThreadPool.cxx
```diff
@@ -154,6 +158,23 @@ ThreadPool::~ThreadPool()
   }
 }
 
+void
+ThreadPool::atfork_prepare()
+{
+  printf("In ThreadPool::atfork_prepare");
```
hjmjohnson, line 164: Debugging comments should be removed.

Modules/Core/Common/src/itkThreadPool.cxx
```diff
@@ -154,6 +158,23 @@ ThreadPool::~ThreadPool()
...
+  m_PimplGlobals->m_ThreadPoolInstance->CleanUp();
+}
+
+void
+ThreadPool::atfork_resume()
+{
+  printf("In ThreadPool::atfork_resume");
```
hjmjohnson, line 171: Debugging comments should be removed.

hjmjohnson (dismissed):

Modules/Core/Common/src/itkThreadPool.cxx
dzenanz: Once this PR gets to a working stage, I will remove them. Currently, they don't get printed.

Modules/Core/Common/include/itkThreadPool.h
```diff
@@ -124,13 +124,23 @@ auto result = pool->AddWork([](int param) { return param; }, 7);
...
-  ~ThreadPool() override;
+
+  /** Stop the pool and release threads. To be called by the destructor and atfork. */
+  void
+  CleanUp();
+
+  ~ThreadPool() override { this->CleanUp(); }
+
+  static void
+  atfork_prepare();
```
thewtex, line 141: Per style guidelines: `atfork_prepare` -> `PrepareForFork`

Modules/Core/Common/include/itkThreadPool.h
```diff
@@ -124,13 +124,23 @@ auto result = pool->AddWork([](int param) { return param; }, 7);
...
+  static void
+  atfork_resume();
```
thewtex, line 143: Per style guidelines: `atfork_resume` -> `ResumeFromFork`

Modules/Core/Common/include/itkThreadPool.h
dzenanz: Sure, I can do that. Do you have any advice about overcoming `multiprocessing.pool.MaybeEncodingError: Error sending result: '[<itk.itkImagePython.itkImageUC3; proxy of <Swig Object of type 'itkImageUC3 *' at 0x7f6ee94955a0> >]'. Reason: 'TypeError("cannot pickle 'SwigPyObject' object")'`? This is when invoking the [repro script](https://github.com/InsightSoftwareConsortium/ITK/issues/2069#issuecomment-754122517).

blowekamp:
After a fork, the PID changes and the old threads are no longer valid. A `GetThreadPool` method should keep track of the PID, and create a new thread pool when the PID changes.

blowekamp:
Am I understanding correctly that with this PR, when a fork occurs, the all threads are terminated in the parent, and new threads are recreated in both new processes? I would expect the parent keeps it's threads and the child process "abandons" the parents threads and keeps new ones.

dzenanz:
> parent keeps it's threads and the child process "abandons" the parents threads

That might be possible, but with more experimentation/effort. The only thing I know of which forks with ITK is Python multiprocessing. And I am trying to address a hang in this situation, with minimal effort. Stopping and cleaning up the thread pool before the fork greatly simplifies the reasoning.

Modules/Core/Common/include/itkThreadPool.h
thewtex: That issue is discussed in #1091 #1948

dzenanz:
I guess that #1948 is a blocker for this. Now waiting for that to be done before continuing work on this.

blowekamp:
@dzenanz I think if you take the test code from #2069, and just write the image instead of returning it, the pool issue should still be reproduced. This would separate this issue from #1948, to move part of it forward.

dzenanz:
@blowekamp you are right! Changing return to imwrite avoids pickling, thus allowing me to verify this fix works. I cleaned it up as per Hans' and Matt's suggestions. Please re-review.

thewtex (approved):
:tada:

thewtex:
Applied to the `release` branch

PR #2652, opened 2021-07-18
STYLE: Increase tests style consistency

jhlegarreta:
Increase tests style consistency:
- Use the `itkNameOfTestExecutableMacro` to get the test name and prevent
  the the test from crashing when `argv[0]` is null.
- Use the ITK testing macros when calling filter update methods to reduce
  boilerplate code.
- Increase consistency in the missing argument check message.
- Use `EXIT_SUCCESS` and `EXIT_FAILURE` to return from tests on
  success/failure.
- Use the `ITK_EXERCISE_BASIC_OBJECT_METHODS` macro instead of explicitly
  calling the `Print` method.

Take advantage of the commit to:
- Remove the argument check when no input parameters are required.
- Define the required types within the test body.
- Define the required types once the input argument check has passed.
- Remove duplicate calls to filter updates.
- Remove unused expected exception cases.
- Only enclose within the `ITK_TRY_EXPECT_NO_EXCEPTION` macro the filter
  update calls that may throw exceptions.
- Remove some unnecessary printed messages.
- Remove unnecessary comment lines devoid of relevant content.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/test/itkImageRandomIteratorTest2.cxx
+++ b/Modules/Core/Common/test/itkImageRandomIteratorTest2.cxx
@@ -22,6 +22,7 @@
 #include "itkImageFileWriter.h"
 #include "itkImageFileReader.h"
 #include "itkTestingComparisonImageFilter.h"
+#include "itkTestingMacros.h"
 
 
 int
@@ -29,9 +30,9 @@ itkImageRandomIteratorTest2(int argc, char * argv[])
 {
   if (argc < 2)
   {
-    std::cerr << "Missing arguments " << std::endl;
-    std::cerr << "Usage: " << std::endl;
-    std::cerr << argv[0] << "  outputImageFile" << std::endl;
+    std::cerr << "Missing parameters." << std::endl;
+    std::cerr << "Usage: " << itkNameOfTestExecutableMacro(argv);
+    std::cerr << "  outputImageFile" << std::endl;
     std::cerr << "[baselineImage  differenceImage]" << std::endl;
     return EXIT_FAILURE;
   }
@@ -103,15 +104,8 @@ itkImageRandomIteratorTest2(int argc, char * argv[])
 
     writer2->SetInput(difference->GetOutput());
 
-    try
-    {
-      writer2->Update();
-    }
-    catch (const itk::ExceptionObject & excp)
-    {
-      std::cerr << excp << std::endl;
-      return EXIT_FAILURE;
-    }
+    ITK_TRY_EXPECT_NO_EXCEPTION(writer2->Update());
+
 
     std::cout << "Number of pixels with differences = ";
     std::cout << difference->GetNumberOfPixelsWithDifferences() << std::endl;
--- a/Modules/IO/BMP/test/itkBMPImageIOTestPalette.cxx
+++ b/Modules/IO/BMP/test/itkBMPImageIOTestPalette.cxx
@@ -31,7 +31,9 @@ itkBMPImageIOTestPalette(int argc, char * argv[])
 {
   if (argc != 5)
   {
-    std::cerr << "Usage: " << argv[0] << " input"
+    std::cerr << "Missing parameters." << std::endl;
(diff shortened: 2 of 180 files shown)


jhlegarreta:
A few notes:
- Using the `itk::ImageWrite` method is out of the scope of this PR.
- I did not add the test ending printed messages. That can be done in a separate PR.
- I did not check all `Update` calls. That can be done in a separate PR.
- I did not check for the input argument name style in the parameter number check. That can be done in a separate PR.
- I did not check for comment message styles (except for a few exceptions). That can be done in a separate PR.

Modules/IO/BMP/test/itkBMPImageIOTestPalette.cxx
```diff
@@ -31,7 +31,9 @@ itkBMPImageIOTestPalette(int argc, char * argv[])
 {
   if (argc != 5)
   {
-    std::cerr << "Usage: " << argv[0] << " input"
+    std::cerr << "Missing parameters." << std::endl;
+    std::cerr << "Usage: " << itkNameOfTestExecutableMacro(argv);
+    std::Cerr << " input"
```
tbirdso, line 36: [CTest failure](https://open.cdash.org/viewBuildError.php?buildid=7347474) due to typo here:

> D:\a\1\s\Modules\IO\BMP\test\itkBMPImageIOTestPalette.cxx(36): error C2039: 'Cerr': is not a member of 'std'

tbirdso (commented):
Test macro standardization looks great! One typo to fix for CI.

Modules/IO/BMP/test/itkBMPImageIOTestPalette.cxx
jhlegarreta: @tbirdso thanks for the review ! Fixed in 76ef46b.

Modules/Core/Common/test/itkImageRandomIteratorTest2.cxx
```diff
@@ -22,16 +22,17 @@
...
 
 int
 itkImageRandomIteratorTest2(int argc, char * argv[])
 {
   if (argc < 2)
   {
-    std::cerr << "Missing arguments " << std::endl;
-    std::cerr << "Usage: " << std::endl;
-    std::cerr << argv[0] << "  outputImageFile" << std::endl;
+    std::cerr << "Missing parameters." << std::endl;
```
N-Dekker, line 33: The original term `arguments` sounds fine to me, why do you prefer `parameters` here? Although both terms are used interchangeably, "arguments" typically refer to the actual values passed to a function (as in this case), while "parameters" typically refer to the names listed in the function definition. https://developer.mozilla.org/en-US/docs/Glossary/Parameter

But I very much like your proposal to use `itkNameOfTestExecutableMacro(argv)`  😃

jhlegarreta: @N-Dekker It is not a matter of preference, it is what the SW Guide says:
https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/blob/master/SoftwareGuide/Latex/Appendices/CodingStyleGuide.tex#L2853
If you think that the term is technically inaccurate, please make a PR to the SW Guide, and a subsequent sweeping change on the ITK code base can follow in a PR separate to this one.

In another PR I brought up the idea of putting these checks and the end of test message+return statement (I believe the message is helpful, but it has been a matter of discussion) in two functions re-usable across the code to avoid these boilerplate statements and make things really consistent. But I do not have the bandwidth to implement that at this time.

tbirdso (approved):

dzenanz (approved):
Awesome!

Modules/Core/Common/test/itkImageRandomIteratorTest2.cxx
jhlegarreta: @N-Dekker I'm marking this as resolved. Please feel free to propose a PR as time permits. Thanks for the review :+1: .

PR #2656, opened 2021-07-20
COMP: directly use [[deprecated]] in itkLegacyMacro - C++14 allows that

dzenanz:
## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkMacro.h
+++ b/Modules/Core/Common/include/itkMacro.h
@@ -610,15 +610,8 @@ OutputWindowDisplayDebugText(const char *);
 //   Provide legacy methods with no warnings.
 #    define itkLegacyMacro(method) method
 #  else
-//   Setup compile-time warnings for uses of deprecated methods if
-//   possible on this compiler.
-#    if defined(__GNUC__) && !defined(__INTEL_COMPILER)
-#      define itkLegacyMacro(method) method __attribute__((deprecated))
-#    elif defined(_MSC_VER)
-#      define itkLegacyMacro(method) __declspec(deprecated) method
-#    else
-#      define itkLegacyMacro(method) method
-#    endif
+//   Request compile-time warnings for uses of deprecated methods.
+#    define itkLegacyMacro(method) [[deprecated]] method
 #  endif
 #endif
 


hjmjohnson (approved):

thewtex (approved):
:robot: :rocket:

Modules/Core/Common/include/itkMacro.h
```diff
@@ -610,15 +610,8 @@ OutputWindowDisplayDebugText(const char *);
...
-//   possible on this compiler.
-#    if defined(__GNUC__) && !defined(__INTEL_COMPILER)
-#      define itkLegacyMacro(method) method __attribute__((deprecated))
-#    elif defined(_MSC_VER)
-#      define itkLegacyMacro(method) __declspec(deprecated) method
-#    else
-#      define itkLegacyMacro(method) method
-#    endif
+//   Request compile-time warnings for uses of deprecated methods.
+#    define itkLegacyMacro(method) [[deprecated]] method
```
N-Dekker, line 614: You could even consider to add a message, `[[deprecated("reason")]]`, as I see at https://en.cppreference.com/w/cpp/language/attributes/deprecated  I think it might be useful for users to explicitly mention `ITK_LEGACY_REMOVE` in the message, for example:

     [[deprecated("This is a legacy method that may be removed by ITK_LEGACY_REMOVE")]]

N-Dekker (approved):
Wonderful @dzenanz!

Modules/Core/Common/include/itkMacro.h
dzenanz: This is a warning which will appear for library users. And instructions how to remove the method they are using would not be too helpful.

N-Dekker: > This is a warning which will appear for library users. And instructions how to remove the method they are using would not be too helpful.

OK, but do you have a better suggestion?

    [[deprecated("This is a legacy method that is only available as long as ITK_LEGACY_REMOVE=OFF")]]

?

dzenanz: That is better. Maybe also write that it will be removed with next major version of ITK? As this PR was merged, it will need to go into a follow-up.

N-Dekker: @dzenanz OK, what about the following? 

  "This is a legacy method. It is only available when ITK_LEGACY_REMOVE=OFF, and it may be removed with the next major ITK release"

I chose "may be removed", instead of "will be removed", as you never know for sure, right? 😺. Or would you prefer, "it is intended to be removed"?

dzenanz: > intended to be removed

This is the truth! Still leaves some wiggle room in case we decide to change mind. I like this the best.

PR #2666, opened 2021-07-27
COMP: Do not enable remote module examples when building Doxygen docs

thewtex:
With CMake 3.20.4, this causes some sort of cache / module enablement
recursion issue resulting in the backtrace:

  CMake Error at CMake/ITKModuleAPI.cmake:78 (message):
    No such module: "ITKIOPNG"
    CMake/ITKModuleAPI.cmake:31 (itk_module_load)
    CMake/ITKModuleAPI.cmake:129 (_itk_module_config_recurse)
    /home/kitware/Dashboards/Tests/ITK-build/CMakeTmp/ITKConfig.cmake:47 (itk_module_config)
    Modules/Remote/AnisotropicDiffusionLBR/examples/CMakeLists.txt:5 (find_packag

We do not need the remote module examples for the Doxygen build, so do
not enable them by default in this case.

### DIFF
--- a/CMake/ITKModuleMacros.cmake
+++ b/CMake/ITKModuleMacros.cmake
@@ -337,7 +337,7 @@ macro(itk_module_examples)
   cmake_dependent_option(Module_${itk-module}_BUILD_EXAMPLES
     "Build the examples for Module_${itk-module}"
     ON
-    "BUILD_EXAMPLES OR ITK_BUILD_EXAMPLES;Module_${itk-module}"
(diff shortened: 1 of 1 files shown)


seanm:
You've got a typo in the commit message: "some sore of"

Seems reasonable looking otherwise, but I don't know much about cmake, I'm probably not the best reviewer...

jhlegarreta (approved):

thewtex:
Addressing commit message typo and merging. Looking forward to having Doxygen updates.

PR #2668, opened 2021-07-30
ENH:Spatial object const correctness

aylward:
1) Extends itkSpatialObject and derived classes to include a method GetConstChildren() that is similar to GetChildren(), but is a const function and returns a list of const objects.

2) Corrects SpatialObject's MetaSceneConverter to accept const pointer to the spatialobject to be written.  The MetaSceneConverter then uses the new GetConstChildren() member function of spatial objects to preserve const-ness as it writes the objects.

3) Updates SurfaceSpatialObject to have get/set member functions for setting a SurfacePoint's Normal in world space.

4) Updates SpatialObjectWriter to accept a const pointer - used by MetaSceneConverter that was also updated.

5) Updated tests for SpatialObjects to verify use of Surface normals in world space.


aylward:
Will squash prior to commit, if approved.

aylward:
Build failures are due to dashboard being offline.

PR #2669, opened 2021-07-30
BUG: TubeSpatialObject didn't preserver Artery flag

aylward:
MetaVesselTubeConverter had a case mis-match on reading and writing "True" and "False" for Artery flag.  Fixed by making the argument test case insensitive.

This was already being tested in itkTubeTK, but the wrong baselines were used so incorrect results were being accepted.  Updates to itkTubeTK will include new baselines.


thewtex:
Rebased on `release`.

dzenanz (approved):

thewtex (approved):

PR #2671, opened 2021-07-30
BUG: Write spatial object color

aylward:
The color of a SpatialObject was not being passed to the MetaObject used to write it to disk.  As a result, color info was lost on write.

This is tested in itkTubeTK, but the baselines were incorrect.  Updated baselines in itkTubeTK will verify these changes.


aylward:
These commits will be squashed prior to merge.

thewtex:
Squashed and rebased on `release`

thewtex (approved):

PR #2679, opened 2021-08-06
COMP: Fix compilation for ITK_FUTURE_LEGACY=ON

hjmjohnson:
Preparing for future code builds with ITK_FUTURE_LEGACY=ON requires
that deprecated function names be replaced with new
names.  Replace SetUseImageSpacingOn() with SetUseImageSpacing(true)


dzenanz (approved):

thewtex (approved):
:rocket:

PR #2680, opened 2021-08-08
COMP: Update SpatialObjects to correct const-ness

aylward:
SpatialObjects can now return a const list of their children, and SpatialObjectWriter accepts a const object.

Furthermore, updated test...A spatial object point has a position in object space, but until it is added to a spatial object (which potentially exists in a hierarchy of objects with parent/child transforms) the point does not have a position in world space.  So, the test has been updated to add a point to an object before asking for that point's world space position.


dzenanz (approved):

aylward:
I am verifying these changes with TubeTK.  Please hold on merging.   Almost done...

aylward:
All have been verified.  

Squashing and then force pushing back to this branch.  And then cherry-picking onto release.

thewtex:
Awesome, thanks, @aylward ! I will integrate after CI finishes.

thewtex (approved):
Love the tests 💚

PR #2665, opened 2021-07-27
Update hdf5 to 1.10.7

seanm:
## PR Checklist
- [ ] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [ ] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [ ] Added test (or behavior not changed)
- [ ] Updated API documentation (or API not changed)
- [ ] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [ ] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [ ] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples) for all new major features (if any)

Refer to the [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) for
further development details if necessary.

PR #2683, opened 2021-08-09
Update HDF5 to 1.10.7

dzenanz:
This is a slightly improved variant of #2665.


dzenanz:
@bradking Ghostflow still complains despite increasing the limit:

commit 1f0411c creates blob 42b9acd74fb8803e677daf2b6830ebf5af58fad3 at Modules/ThirdParty/HDF5/src/itkhdf5/src/H5Shyper.c with size 518489 bytes (506.34 KiB) which is greater than the maximum size 500000 bytes (488.28 KiB). If the file is intended to be committed, set the hooks-max-size attribute on its path.

bradking:
There is a more-local `.gitattributes` overriding it in a subdirectory:

```console
$ git check-attr hooks-max-size -- Modules/ThirdParty/HDF5/src/itkhdf5/src/H5Shyper.c
Modules/ThirdParty/HDF5/src/itkhdf5/src/H5Shyper.c: hooks-max-size: 500000
$ grep H5Shyper.c Modules/ThirdParty/HDF5/.gitattributes
src/itkhdf5/src/H5Shyper.c   hooks-max-size=500000
```

seanm:
@dzenanz Did you run the mangling on any other OSes?  I only did macOS.

dzenanz:
No, I didn't.

dzenanz:
Maybe I could update Windows mangling, and you Linux?

seanm:
I don't have Windows nor Linux.  Only macOS and FreeBSD.

dzenanz:
It turns out that Linux produces the same list you got on MacOS, Sean.

dzenanz:
This is ready for review/merging.

seanm:
I'll be able to review later today... some fires in the office to deal with first. :)

Jon, alas, many of the warnings I fixed in upstream HDF5 post-date 1.10.7.  But getting to 1.10.7 is a necessary first step I think.  We can then consider going to the tip of the 1.10 branch.

jhlegarreta:
> many of the warnings I fixed in upstream HDF5 post-date 1.10.7. But getting to 1.10.7 is a necessary first step I think. We can then consider going to the tip of the 1.10 branch.

I see; I did not pay attention to the versions that integrated the fixes. My bad. In any case, this is another step forward.

dzenanz:
I am really itching to merge this. Can you please review @seanm?

seanm:
So do we want to now update to the tip of the 1.10 branch?  It would solve most (all?) of the compiler warnings, but I'm not sure how stable that branch is...

jhlegarreta:
I have had a look at the HDF5 releases and I am not sure if I understand their versioning. I ignore if the tip of 1.10 contains changes that are not in e.g. 1.12.x. Not sure if they plan to release a 1.10.8. But in any case, if the 1.10 tip will solve more/most/all compiler warnings, +1 to bring them in to `master`; then jumping to a new HDF5 release can be made later, and be incorporated on a new ITK minor release.

dzenanz:
+1 for going to tip of 1.10 now. Then, if there are no major version compatibility issues, go to 1.12.x series before starting ITK 5.3 release cycle.

PR #2687, opened 2021-08-12
COMP: Add missing template export macro to SLICImageFilter

thewtex:
Addresses visibility issues when building on macOS / static, e.g. with
the Python wrappings.

### DIFF
--- a/Modules/Segmentation/SuperPixel/include/itkSLICImageFilter.h
+++ b/Modules/Segmentation/SuperPixel/include/itkSLICImageFilter.h
@@ -58,7 +58,7 @@ namespace itk
  * \ingroup Segmentation ITKSuperPixel MultiThreading
  */
 template <typename TInputImage, typename TOutputImage, typename TDistancePixel = float>
-class SLICImageFilter : public ImageToImageFilter<TInputImage, TOutputImage>
+class ITK_TEMPLATE_EXPORT SLICImageFilter : public ImageToImageFilter<TInputImage, TOutputImage>
 {
 public:
   ITK_DISALLOW_COPY_AND_MOVE(SLICImageFilter);


Leengit:
The shell command
```bash
cd git/ITK ; egrep -R 'class.*: public ImageToImageFilter<' | fgrep -v ITK_TEMPLATE_EXPORT
```
gives 5 other possibilities that might need `ITK_TEMPLATE_EXPORT`:
```text
Modules/Filtering/BiasCorrection/include/itkN4BiasFieldCorrectionImageFilter.h:class N4BiasFieldCorrectionImageFilter : public ImageToImageFilter<TInputImage, TOutputImage>
Modules/Filtering/ImageFilterBase/test/itkImageToImageFilterTest.cxx:class ImageToImageFilterTestHelper : public ImageToImageFilter<TInputImage, TOutputImage>
Modules/Core/Transform/include/itkResampleInPlaceImageFilter.h:class ResampleInPlaceImageFilter : public ImageToImageFilter<TInputImage, TOutputImage>
Examples/Filtering/CompositeFilterExample.cxx:class CompositeExampleImageFilter : public ImageToImageFilter<TImage, TImage>
Wrapping/Generators/Python/PyUtils/itkPyImageFilter.h:class PyImageFilter : public ImageToImageFilter<TInputImage, TOutputImage>
```

Looking for all templated superclasses
```bash
egrep -R 'class.*: public [A-Za-z]+<' | fgrep -v ITK_TEMPLATE_EXPORT
```
gives 290 possibilities.

jhlegarreta:
> gives 5 other possibilities that might need ITK_TEMPLATE_EXPORT:

I'd dare to say that the macro should not be necessary/is not useful for the classes in the tests and the examples.

thewtex:
> gives 5 other possibilities that might need ITK_TEMPLATE_EXPORT:

@Leengit good catch, patch here: https://github.com/InsightSoftwareConsortium/ITK/pull/2688

As @jhlegarreta , we don't need the macro for tests, examples.

PR #2691, opened 2021-08-13
BUG: Initialize C++ standard for remote modules

thewtex:
We now default to C++14, which is not the default standard on most compilers. Addresses:

  Modules/Core/Common/include/itkMacro.h:192:6: error: "Apple LLVM < 5.1 (clang < 3.4) is not supported under ITKv5.3"

When building remote modules externally, which is triggered by too low of a C++ standard.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,14 +35,14 @@ foreach(pold "CMP0091")
   endif()
 endforeach()
 
+# ====
+include(${CMAKE_CURRENT_SOURCE_DIR}/CMake/ITKInitializeCXXStandard.cmake)
+include(${CMAKE_CURRENT_SOURCE_DIR}/CMake/ITKInitializeBuildType.cmake)
+
 # ==== Define language standard configurations requiring at least c++14 standard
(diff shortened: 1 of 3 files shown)


dzenanz (approved):

tbirdso (approved):

CMakeLists.txt
```diff
@@ -40,19 +40,8 @@ if(CMAKE_CXX_STANDARD EQUAL "98" OR CMAKE_CXX_STANDARD LESS "14")
...
-endif()
-if(NOT CMAKE_CXX_STANDARD_REQUIRED)
-  set(CMAKE_CXX_STANDARD_REQUIRED ON)
-endif()
-if(NOT CMAKE_CXX_EXTENSIONS)
-  set(CMAKE_CXX_EXTENSIONS OFF)
-endif()
-
 # ====
+include(${CMAKE_CURRENT_SOURCE_DIR}/CMake/ITKInitializeCXXStandard.cmake)
```
Leengit, line 44: Line 39 above checks the value of `CMAKE_CXX_STANDARD`, even though it is possible that it is first set here.  Even when Cmake allows that, it would be clearer and more future proof if we check the value after this include statement.

thewtex: I will adjust the order.

PR #2696, opened 2021-08-17
ENH: Split Hausdorff distance test

### LINKED ISSUE #2445: Split `itkHausdorffDistanceImageFilterTest` to the corresponding classes unit tests
Split `itkHausdorffDistanceImageFilterTest` to the corresponding classes unit tests

### Description

The classes ``itk::HausdorffDistanceImageFilter` and `itk::DirectedHausdorffDistanceImageFilter` are tested in the same [`itkHausdorffDistanceImageFilterTest.cxx](itk::https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/DistanceMap/test/itkHausdorffDistanceImageFilterTest.cxx) file. Each class should have its own test file.

The test's name is also `itkHausdorffDistanceImageFilterTest`, which is misleading considering that `itk::DirectedHausdorffDistanceImageFilter` is a separate class with its own header and implementation files, and is being tested under the mentioned test name.

### Expected behavior

The classes `itk::HausdorffDistanceImageFilter` and `itk::DirectedHausdorffDistanceImageFilter` should be tested in separate files: the part of the test relevant to `itk::DirectedHausdorffDistanceImageFilter` should be put into a separate test file.

### Actual behavior

The `itkHausdorffDistanceImageFilterTest.cxx` contains tests for both the `itk::HausdorffDistanceImageFilter` and the `itk::DirectedHausdorffDistanceImageFilter` classes. Each class should be tested in separate files.

It is also weird since the test does not include the `itk::HausdorffDistanceImageFilter` header file.

### Additional Information

The class `itk::DirectedHausdorffDistanceImageFilter` does have [its own test file](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/DistanceMap/test/itkDirectedHausdorffDistanceImageFilterTest.cxx) that uses some input image.

Best would be to rename the file to `itkDirectedHausdorffDistanceImageFilterTest2.cxx` and name the test with the piece of code stripped from the `itkHausdorffDistanceImageFilterTest.cxx` as `itkDirectedHausdorffDistanceImageFilterTest1.cxx`.

The existing `itkDirectedHausdorffDistanceImageFilterTest.cxx` was probably contributed without noticing that the relevant class was being exercised in  `itkHausdorffDistanceImageFilterTest.cxx`. Also, the former might not be providing meaningful image results despite being tested (the output was certainly not checked when contributing). So another possibility would be to use the synthetic data as in `itkHausdorffDistanceImageFilterTest.cxx` for the test, and if a meaningful image is found, or the output is meaningful for an actual image file, use that for the ITK SW Guide or ITK Sphinx examples.

ebrahimebrahim:
ENH: Split Hausdorff distance test

Resolves #2445

The test `itkHausdorffDistanceImageFilterTest` previously covered two classes, `HausdorffDistanceImageFilter` and `DirectedHausdorffDistanceImageFilter`.

The `DirectedHausdorffDistanceImageFilter` have been split off into a new test, `itkDirectedHausdorffDistanceImageFilterTest1`.

The test formerly named `itkDirectedHausdorffDistanceImageFilterTest` is now called `itkDirectedHausdorffDistanceImageFilterTest2`.

The latter test is sort of trivial, but covers slightly different aspects of `DirectedHausdorffDistanceImageFilter` and so it may be worth keeping. I simplified its code a bit, so that it doesn't appear to promise to test more than what it actually does.

As for the output file of `itkDirectedHausdorffDistanceImageFilterTest2`, it looks as it should; i.e. it looks like the input image. I think this filter is just meant to compute a value and it's not meant to modify the image. This addresses a concern raised at the end of #2445.

---

## PR Checklist

### DIFF
--- a/Modules/Filtering/DistanceMap/test/itkDirectedHausdorffDistanceImageFilterTest2.cxx
+++ b/Modules/Filtering/DistanceMap/test/itkDirectedHausdorffDistanceImageFilterTest2.cxx
@@ -28,11 +28,10 @@ using namespace itk::print_helper;
 int
 itkDirectedHausdorffDistanceImageFilterTest2(int argc, char * argv[])
 {
-  if (argc != 5)
+  if (argc != 3)
   {
     std::cerr << "Missing Parameters." << std::endl;
-    std::cerr << "Usage: " << itkNameOfTestExecutableMacro(argv)
-              << " inputFileName1 inputFileName2 outputFileName useImageSpacing" << std::endl;
+    std::cerr << "Usage: " << itkNameOfTestExecutableMacro(argv) << " inputFileName outputFileName" << std::endl;
     return EXIT_FAILURE;
   }
 
@@ -49,7 +48,7 @@ itkDirectedHausdorffDistanceImageFilterTest2(int argc, char * argv[])
 
   ITK_TRY_EXPECT_NO_EXCEPTION(reader1->Update());
 
-  reader2->SetFileName(argv[2]);
+  reader2->SetFileName(argv[1]);
 
   ITK_TRY_EXPECT_NO_EXCEPTION(reader2->Update());
 
@@ -68,8 +67,7 @@ itkDirectedHausdorffDistanceImageFilterTest2(int argc, char * argv[])
   filter->SetInput2(image2);
   ITK_TEST_SET_GET_VALUE(image2, filter->GetInput2());
 
-  auto useImageSpacing = static_cast<bool>(std::stoi(argv[4]));
-  ITK_TEST_SET_GET_BOOLEAN(filter, UseImageSpacing, useImageSpacing);
+  ITK_TEST_SET_GET_BOOLEAN(filter, UseImageSpacing, true);
 
   ITK_TRY_EXPECT_NO_EXCEPTION(filter->Update());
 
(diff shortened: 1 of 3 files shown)


Modules/Filtering/DistanceMap/test/itkDirectedHausdorffDistanceImageFilterTest2.cxx
```diff
@@ -49,7 +48,7 @@ itkDirectedHausdorffDistanceImageFilterTest2(int argc, char * argv[])
 
   ITK_TRY_EXPECT_NO_EXCEPTION(reader1->Update());
```
dzenanz, line 49: This is nor part of #2445, but you could further simplify the tests by using convenience function which was recently added. Example of how to do it: https://github.com/InsightSoftwareConsortium/ITKSphinxExamples/pull/307.

The style in tests:
```cpp
typename ImageType::Pointer image1 = nullptr;
ITK_TRY_EXPECT_NO_EXCEPTION(image1 = itk::ReadImage<ImageType>(argv[1]));
```

dzenanz (approved):
Assuming CI passes, this is good as-is. It can be further simplified by using `ReadImage` convenience function - this can be an additional commit.

thewtex (approved):
@ebrahimebrahim nice work!

Modules/Filtering/DistanceMap/test/itkDirectedHausdorffDistanceImageFilterTest2.cxx
dzenanz: This could be done in a follow-up PR, and might include using `ReadImage` in many other tests, not just the ones touched by this PR.

PR #2701, opened 2021-08-19
BUG: Fix matrix check rejecting NIFTI sform for small voxels

### LINKED ISSUE #2674: PR switching from qform to sform causing many downstream support issues, particularly for small voxel sizes
The PR here, https://github.com/InsightSoftwareConsortium/ITK/pull/1868

Is regularly triggering issues in various downstream projects, with users with (wrong? malformed? other?) NIFTI files which used to work with ITK-based tools and now do not.

https://github.com/Project-MONAI/MONAILabel/issues/35
https://github.com/CoBrALab/RABIES/issues/160
https://github.com/ANTsX/ANTs/issues/1213
https://github.com/SimpleITK/SimpleITK/issues/1433

I'm not sure what the solution to this is, but I'd like to propose a couple:

1. Change code to be a warning and fallthrough to qform reading
2. Provide a more informative error for the user to correct the generation of the invalid NIFTI file

cookpa:
Fixes #2674

NIFTI sform matrices were being called non-invertible if their determinant was <
10-5. Because the determinant scales with the product of the voxel spacings, any
image with voxels smaller than about 20 microns would not have its direction
cosines read from the sform. This caused errors when the header sform was valid,
but the qform transform was not available as a fallback.

Invertibility is now tested by checking the condition number of the singular
value decomposition. If this is above the float epsilon, the matrix will be
considered invertible.

An alternative solution would be to consider a matrix invertible whenever the
determinant is > 0, which is what itkMatrix does. The condition number test as
implemented is a bit more conservative, but should allow spacings down to a
nanometer or less.

## PR Checklist

### DIFF
--- a/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
+++ b/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
@@ -20,7 +20,6 @@
 #include "itkMetaDataObject.h"
 #include "itkSpatialOrientationAdapter.h"
 #include <nifti1_io.h>
-
 #include "itkNiftiImageIOConfigurePrivate.h"
 
 namespace itk
@@ -1759,14 +1758,15 @@ IsAffine(const mat44 & nifti_mat)
       return false;
     }
   }
-  // Get the inverse of this matrix:
-  // Make sure the matrix is invertible to begin with...
-  if (fabs(vnl_determinant(mat)) <= 1e-5)
+
+  const float condition = vnl_matrix_inverse<float>(mat.as_matrix()).well_condition();
+  // Check matrix is invertible by testing condition number of inverse
+  if (!(condition > std::numeric_limits<float>::epsilon()))
   {
     return false;
   }
 
-  // Calculate the inverse and seperate the inverse translation component
+  // Calculate the inverse and separate the inverse translation component
   // and the top 3x3 part of the inverse matrix
   const vnl_matrix_fixed<float, 4, 4> inv4x4Matrix = vnl_matrix_inverse<float>(mat.as_matrix()).as_matrix();
   const vnl_vector_fixed<float, 3>    inv4x4Translation(inv4x4Matrix[0][3], inv4x4Matrix[1][3], inv4x4Matrix[2][3]);


dzenanz (approved):
LGTM, but it would be good if someone else took a look too. Even better if someone else tried 😄

thewtex (approved):
@cookpa well done. Thanks for the contribution!

cookpa:
A simple test with some sample images here

https://github.com/cookpa/TestITKNIFTI

hjmjohnson (approved):
This is a much more comprehensive and well-thought-out condition.  Thank you for your contribution!

jhlegarreta:
I have only loosely followed the issue/PR. Thanks for the effort all. @cookpa wondering whether the test you wrote in the separate repository should be put into ITK/you would be willing to do that.

cookpa:
I'd be willing to write the test @jhlegarreta , is there an up to date guide on how to do that? I think I would need to upload the data files from that repo, because I had to manually edit the headers to remove the qform.

jhlegarreta:
:100: Thanks @cookpa. The [ITK SW Guide](https://itk.org/ItkSoftwareGuide.pdf) contains guidelines (sections 9.4 C.23) to create ITK tests. Essentially:
- The `main.cxx` you wrote should be placed in the [Modules/IO/NIFTI/test](https://github.com/cookpa/ITK/tree/master/Modules/IO/NIFTI/test) folder and be renamed to e.g. `itkNiftiSmallVoxelSizeSformTest.cxx`, `itkNiftiQformTFallBackTest.cxx` or any other appropriate name. The `main` function should be named the same without the file extension.
- Make the test quantitative comparing the read values and the expected values. Use the ITK testing macros (see section 9.4 in the ITK SW Guide) as necessary.
- Improve the input argument checking as stated in the pointed ITK SW Guide sections.
- Add the test entry/entries to the corresponding [CMakeLists.txt](https://github.com/cookpa/ITK/blob/master/Modules/IO/NIFTI/test/CMakeLists.txt).
- Add the test content links to the [corresponding folder](https://github.com/cookpa/ITK/tree/master/Modules/IO/NIFTI/test/Input). [Data.md](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/Data.md) and [UploadBinaryData.md](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/UploadBinaryData.md) provide instructions to generate the content links.

HTH.

cookpa:
Thanks @jhlegarreta - I have not forgotten this. I think this existing test code would work, if it was given files with the right headers:

https://github.com/cookpa/ITK/blob/master/Modules/IO/NIFTI/test/itkNiftiReadWriteDirectionTest.cxx

I'll look more into how to do that.

gdevenyi:
Are there any plans to merge this into a 5.2.x bugfix? I'm running into this in with a high-resolution 0.3mm isotropic image I'm working with.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -1759,14 +1758,15 @@ IsAffine(const mat44 & nifti_mat)
       return false;
     }
   }
-  // Get the inverse of this matrix:
-  // Make sure the matrix is invertible to begin with...
-  if (fabs(vnl_determinant(mat)) <= 1e-5)
+
+  const float condition = vnl_matrix_inverse<float>(mat.as_matrix()).well_condition();
+  // Check matrix is invertible by testing condition number of inverse
+  if (!(condition > std::numeric_limits<float>::epsilon()))
```
seanm, line 1764: Why ! > instead of <= ?

seanm:
The commit message is great, but alas my maths are too rusty... @vfonov maybe you'd care to take a look?

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
Leengit: `! >` and `<=` give different results for floating point NaNs.  With the former, if the condition is small or is a NaN then ...

seanm: Right! Pretty subtle though. Something like `if (!isfinite(condition) || (condition < epsilon))` might make the intention clearer.

cookpa: @seanm you're right, that is a nicer way to phrase it.

Leengit: The `! >` approach dates back forever and its purpose is easily recognizable by us oldtimers.  If we are nonetheless worried that it is opaque, that could be addressed in a comment.

If we instead go with the newer approach, instead of `isfinite` we might want to be checking `isnan` -- unless we are sure that a positive infinity should return `false` in our current context, or unless we are sure that positive infinity is impossible in our current context.

seanm: >The ! > approach dates back forever and its purpose is easily recognizable by us oldtimers

Maybe I'm an even older timer than you then, because my foggy old memory forgot about this trick. :)

gdevenyi:
Looks like there's still some corner cases here, https://github.com/InsightSoftwareConsortium/ITK/issues/3333

PR #2709, opened 2021-09-01
ENH: Replace GetNumberOfWeights() by static constexpr `NumberOfWeights`

N-Dekker:
Replaced the `GetNumberOfWeights()` member functions of both `BSplineBaseTransform` and `BSplineInterpolationWeightFunction` by static constexpr `NumberOfWeights` data members.

Allows using the number of weights at compile-time, for example to declare an `std::array` of those weights.

### DIFF
--- a/Documentation/ITK5MigrationGuide.md
+++ b/Documentation/ITK5MigrationGuide.md
@@ -440,6 +440,9 @@ now has signature
 `JacobianPositionType` is publicly exposed in `itk::Transform`.
 See commit [commit 212cae5](https://github.com/InsightSoftwareConsortium/ITK/commit/212cae522d8451ea089c41f8a151279e1dd17042) for details.
 
+With ITK 5.3, the `GetNumberOfWeights()` member functions of `itk::BSplineBaseTransform` and `itk::BSplineInterpolationWeightFunction`
+are replaced by static constexpr data members named `NumberOfWeights`.
+
 Consolidated Vector Filter
 --------------------------
 
--- a/Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h
+++ b/Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h
@@ -23,6 +23,7 @@
 #include "itkBSplineKernelFunction.h"
 #include "itkArray.h"
 #include "itkArray2D.h"
+#include "itkMath.h"
 
 namespace itk
 {
@@ -69,6 +70,9 @@ public:
   /** Spline order. */
   static constexpr unsigned int SplineOrder = VSplineOrder;
 
+  /** Number of weights. */
+  static constexpr unsigned int NumberOfWeights{ Math::UnsignedPower(VSplineOrder + 1, VSpaceDimension) };
+
   /** OutputType type alias support. */
   using WeightsType = Array<double>;
 
@@ -98,8 +102,10 @@ public:
   /** Get support region size. */
   itkGetConstMacro(SupportSize, SizeType);
 
+#if !defined(ITK_LEGACY_REMOVE)
   /** Get number of weights. */
-  itkGetConstMacro(NumberOfWeights, unsigned int);
+  itkLegacyMacro(unsigned int GetNumberOfWeights() const) { return Self::NumberOfWeights; }
+#endif
 
 protected:
   BSplineInterpolationWeightFunction();
@@ -108,9 +114,6 @@ protected:
   PrintSelf(std::ostream & os, Indent indent) const override;
 
 private:
-  /** Number of weights. */
-  unsigned int m_NumberOfWeights;
-
   /** Size of support region. */
   SizeType m_SupportSize;
 
--- a/Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.hxx
+++ b/Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.hxx
@@ -30,15 +30,11 @@ namespace itk
 template <typename TCoordRep, unsigned int VSpaceDimension, unsigned int VSplineOrder>
 BSplineInterpolationWeightFunction<TCoordRep, VSpaceDimension, VSplineOrder>::BSplineInterpolationWeightFunction()
 {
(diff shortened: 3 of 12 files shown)


jhlegarreta:
Thanks for doing this @N-Dekker . My knowledge about the rationale of the change is limited, but I would love to know about its advantages. Also, it looks like it might be a candidate for extension throughout the code, so I am wondering about the plans for it. The ITK SW Guide should be updated to cover the aspects related to this change.

Last but not least, coverage should be kept for the legacy implementation using the appropriate guards in the test files.

dzenanz (commented):
This change should be mentioned in the migration guide. Also, as Jon pointed out, ITK + tests should build with legacy both on and off.

N-Dekker:
Thanks for your comments @jhlegarreta and @dzenanz  At the moment, I'm mainly focused on BSpline transform, and I believe a significant performance improvement may be achieved by allowing the number of weights to be calculated at compile-time, and then (as a next step) by storing those weights in a compile-time fixed array (like `std::array`). 

Thanks for your suggestions. I will still add legacy-only tests for `GetNumberOfWeights()` with this pull request. Which migration guide document is it specifically?

jhlegarreta:
Thanks for the explanation.

> I believe a significant performance improvement may be achieved

Then maybe this would be a `PERF:` change (or `STYLE: ` with more reservations?).

N-Dekker:
@jhlegarreta 
> Then maybe this would be a `PERF:` change (or `STYLE: ` with more reservations?).

Well... I named this particular commit `ENH` as it adds a "new feature": the ability to use the number of weights at compile-time.  I think the real (significant) PERF improvement would come from a possible follow-up commit that would use a compile-time fixed array for weights. That would then clearly be a `PERF` commit to me.

However, if you insist, I can adjust the commit subject to either `STYLE` or `PERF`, no problem. Which one do you prefer?

dzenanz:
I prefer this to remain `ENH` type commit. Migration guide is here:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/ITK5MigrationGuide.md

N-Dekker:
> I prefer this to remain `ENH` type commit. 

Thanks @dzenanz  I think the commit does have both style and performance aspects to it, but in the first place, it should be seen as an enhancement.

> Migration guide is here:
> https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/ITK5MigrationGuide.md

Are you sure? Isn't that document specifically about migrating from ITK 4.x to ITK 5.0?  My intention for this commit would be to have it included with ITK 5.3.

hjmjohnson (commented):
Great changes!

dzenanz:
> Isn't that document specifically about migrating from ITK 4.x to ITK 5.0?

It is primarily about that, but it is also for migrating from 4.x to current 5.x.

N-Dekker:
@dzenanz @jhlegarreta The latest force-push aims to address both your comments on migration documentation and (legacy) test coverage. Please check: https://github.com/InsightSoftwareConsortium/ITK/compare/ad721095bba8b662f66597538531cee14cec1d6a..f3582c006de6ee741b8777c3126ec1531e4cc07a

dzenanz (approved):

jhlegarreta:
I'm late, sorry. Thanks for doing this. I'm still missing to add the corresponding explanation in the ITK SW Guide, even more so as I now realize that this affects also the `PrintSelf` method, e.g.
https://github.com/InsightSoftwareConsortium/ITK/pull/2709/files#diff-9eb7d4537b8504d16ce1a0f319f5cd47456bab934b18bc0370f9dc0fccd9071fR65

Note that other static members are not usually printed, e.g. `SplineOrder` for the affected class in this case
https://github.com/InsightSoftwareConsortium/ITK/pull/2709/files#diff-7014ac3a68b0fc3b484a271e57aef63e90f4f5157e81945b20dc6808e28973eaL70

So if we are to print them, it should be explicitly stated in the guide, and classes in ITK revised.

N-Dekker:
@jhlegarreta 
> I'm late, sorry. Thanks for doing this. 

Sorry for moving on so quickly, but there are a few more BSpline implementation improvements I have in mind, that I want to do step by step. I think it can be made significantly faster, using modern C++ (up to C++14).

> Note that other static members are not usually printed, e.g. `SplineOrder` for the affected class in this case
... 
> So if we are to print them, it should be explicitly stated in the guide, and classes in ITK revised.

I do agree that it isn't really necessary for `BSplineInterpolationWeightFunction::PrintSelf` to print "NumberOfWeights", as it is always constant anyway. Shall I just remove it? I guess that would be in accordance with the ITK SW Guide, right?

@dzenanz what do you think? I guess like what you said about "SupportSize" at https://github.com/InsightSoftwareConsortium/ITK/pull/2711#discussion_r701971465  Just remove the print statement, right? 😃

hjmjohnson:
Thanks @N-Dekker.  I would remove the print statement. It is constant, and not all that helpful when dynamically interrogating the class.

N-Dekker:
> Thanks @N-Dekker. I would remove the print statement. It is constant, and not all that helpful when dynamically interrogating the class.

OK, that's fine to me, @hjmjohnson  As @dzenanz  and I already agreed to remove "SupportSize" as well, there isn't anything left to print anymore then, from `BSplineInterpolationWeightFunction`! Shall I then just remove the entire `BSplineInterpolationWeightFunction::PrintSelf` override? Just removing the entire function definition?

jhlegarreta:
>  Just removing the entire function definition?

I'd say that every class' `PrintSelf` method should call at least the superclass `PrintSelf` method. The ITK SW Guide reads
```
The Superclass must always be printed.
```

section C.21

dzenanz:
By removing the method definition, superclass' variant will be invoked via inheritance. And yes, I am in favor of removing he print statement.

jhlegarreta:
> By removing the method definition, superclass' variant will be invoked via inheritance. And yes, I am in favor of removing he print statement.

True, sorry for the oversight.

PR #2710, opened 2021-09-02
ENH: Declare Filled and operator[] of Index, Size, FixedArray constexpr

N-Dekker:
Declared the `Filled(value)` and `operator[](pos)` member functions of
`itk::FixedArray`, `itk::Index`, and `itk::Size` constexpr.

### DIFF
--- a/Modules/Core/Common/include/itkFixedArray.h
+++ b/Modules/Core/Common/include/itkFixedArray.h
@@ -262,8 +262,8 @@ public:
 #    pragma GCC diagnostic ignored "-Warray-bounds"
 #  endif
 #endif
-  reference       operator[](unsigned int index) { return m_InternalArray[index]; }
-  const_reference operator[](unsigned int index) const { return m_InternalArray[index]; }
+  constexpr reference       operator[](unsigned int index) { return m_InternalArray[index]; }
+  constexpr const_reference operator[](unsigned int index) const { return m_InternalArray[index]; }
 #if defined(__GNUC__)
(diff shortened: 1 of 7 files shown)


Modules/Core/Common/include/itkFixedArray.h
```diff
@@ -262,8 +262,8 @@ class ITK_TEMPLATE_EXPORT FixedArray
 #    pragma GCC diagnostic ignored "-Warray-bounds"
 #  endif
 #endif
-  reference       operator[](unsigned int index) { return m_InternalArray[index]; }
-  const_reference operator[](unsigned int index) const { return m_InternalArray[index]; }
+  constexpr reference       operator[](unsigned int index) { return m_InternalArray[index]; }
```
dzenanz, line 265: Adding `constexpr` here is a bit confusing. But I guess required to make `Modules/Core/Common/test/itkFixedArrayGTest.cxx:235` possible at compile time.

dzenanz (approved):

Modules/Core/Common/include/itkFixedArray.h
N-Dekker: @dzenanz Indeed, the constexpr `operator[]` is used here, in order to test the result of `FixedArray::Filled(value)` at compile time.

It doesn't seem bad to declare `operator[]`  as well, as C++17 also does so for `std::array::operator[]`: https://en.cppreference.com/w/cpp/container/array/operator_at  😃

hjmjohnson (approved):
Very nice!

N-Dekker:
@dzenanz @hjmjohnson Thanks for your approval!

PR #2711, opened 2021-09-02
ENH: Replace BSpline GetSupportSize() by static constexpr `SupportSize`

N-Dekker:
Replaced the `GetSupportSize()` member function of
`BSplineInterpolationWeightFunction` by a static constexpr data member,
`SupportSize`.

Allows using this size at compile-time.

Follow-up to pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2709
commit 34f3db8d8ad0ba848cbbfee071edf327e09f6a96
"ENH: Replace GetNumberOfWeights() by static constexpr `NumberOfWeights`"

### DIFF
--- a/Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.hxx
+++ b/Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.hxx
@@ -30,15 +30,13 @@ namespace itk
 template <typename TCoordRep, unsigned int VSpaceDimension, unsigned int VSplineOrder>
 BSplineInterpolationWeightFunction<TCoordRep, VSpaceDimension, VSplineOrder>::BSplineInterpolationWeightFunction()
 {
-  // Initialize support region is a hypercube of length SplineOrder + 1
-  m_SupportSize.Fill(SplineOrder + 1);
-
   // Initialize offset to index lookup table
   m_OffsetToIndexTable.set_size(Self::NumberOfWeights, SpaceDimension);
 
-  unsigned int counter = 0;
+  constexpr auto supportSize = Self::SupportSize;
+  unsigned int   counter = 0;
 
-  for (const IndexType index : ZeroBasedIndexRange<VSpaceDimension>(m_SupportSize))
+  for (const IndexType index : ZeroBasedIndexRange<VSpaceDimension>(supportSize))
   {
     for (unsigned int j = 0; j < SpaceDimension; ++j)
     {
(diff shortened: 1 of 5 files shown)


Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.hxx
```diff
@@ -63,7 +61,9 @@ BSplineInterpolationWeightFunction<TCoordRep, VSpaceDimension, VSplineOrder>::Pr
   Superclass::PrintSelf(os, indent);
 
   os << indent << "NumberOfWeights: " << Self::NumberOfWeights << std::endl;
-  os << indent << "SupportSize: " << m_SupportSize << std::endl;
+
+  constexpr auto supportSize = Self::SupportSize;
+  os << indent << "SupportSize: " << supportSize << std::endl;
```
dzenanz, line 66: Why not print `Self::SupportSize` directly? Why print this constant at all?

dzenanz (approved):

Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.hxx
N-Dekker: > Why not print Self::SupportSize directly?

Good question @dzenanz. Linux/GCC rejected the code when Self::SupportSize was directly inserted into the output stream (os). It did complain about an "undefined reference to itk::BSplineInterpolationWeightFunction::SupportSize'", at https://open.cdash.org/viewBuildError.php?buildid=7435820 While it was just fine to the other platforms/compilers.

> Why print this constant at all?

Honestly I don't know. The SupportSize of a BSplineInterpolationWeightFunction was always just filled by `VSplineOrder + 1` during initialization.  The original `m_SupportSize` could never be modified by the user either. Would you agree to just remove this print statement?

dzenanz: Yes, removing the print statement is the best in my opinion.

N-Dekker: @dzenanz Just removed the print statement for SupportSize, with force-push https://github.com/InsightSoftwareConsortium/ITK/compare/eae5572cde584c2020fd0eb68035e45c2ab6acc5..4c58c6fa7f5b7f5eb0689de17b8e7e7613442144

dzenanz (approved):

hjmjohnson (approved):

N-Dekker:
Thanks for your feedback and approval of this PR, @dzenanz and @hjmjohnson !

PR #2712, opened 2021-09-03
PERF: Use FixedArray for BSplineInterpolationWeightFunction OutputType

N-Dekker:
Replaced `itk::Array` by `itk::FixedArray` as output type (array of
weights) of `itk::BSplineInterpolationWeightFunction`.

Observed a significant improvement of the runtime performance of a call to `itk::BSplineBaseTransform::TransformPoint`: the runtime duration appeared ~30% lower than before, from 1.5 sec. down to 1.0 sec. when calling `TransformPoint` 10'000'000 times. (Tested using Visual C++ 2019 Release).

### DIFF
--- a/Modules/Registration/Common/include/itkImageToImageMetric.hxx
+++ b/Modules/Registration/Common/include/itkImageToImageMetric.hxx
@@ -437,7 +437,6 @@ ImageToImageMetric<TFixedImage, TMovingImage>::MultiThreadingInitialize()
     this->m_BSplineTransformIndicesArray.SetSize(1, 1);
     this->m_BSplinePreTransformPointsArray.resize(1);
     this->m_WithinBSplineSupportRegionArray.resize(1);
-    this->m_BSplineTransformWeights.SetSize(1);
     this->m_BSplineTransformIndices.SetSize(1);
 
     delete[] this->m_ThreaderBSplineTransformWeights;
@@ -457,7 +456,6 @@ ImageToImageMetric<TFixedImage, TMovingImage>::MultiThreadingInitialize()
     }
     else
     {
-      this->m_BSplineTransformWeights.SetSize(this->m_NumBSplineWeights);
       this->m_BSplineTransformIndices.SetSize(this->m_NumBSplineWeights);
 
       this->m_ThreaderBSplineTransformWeights = new BSplineTransformWeightsType[m_NumberOfWorkUnits - 1];
@@ -465,7 +463,6 @@ ImageToImageMetric<TFixedImage, TMovingImage>::MultiThreadingInitialize()
 
       for (ThreadIdType ithread = 0; ithread < m_NumberOfWorkUnits - 1; ++ithread)
       {
-        this->m_ThreaderBSplineTransformWeights[ithread].SetSize(this->m_NumBSplineWeights);
         this->m_ThreaderBSplineTransformIndices[ithread].SetSize(this->m_NumBSplineWeights);
       }
     }
@@ -797,7 +794,7 @@ ImageToImageMetric<TFixedImage, TMovingImage>::PreComputeTransformValues()
   m_Transform->SetParameters(dummyParameters);
 
   // Cycle through each sampled fixed image point
-  BSplineTransformWeightsType    weights(m_NumBSplineWeights);
+  BSplineTransformWeightsType    weights;
   BSplineTransformIndexArrayType indices(m_NumBSplineWeights);
   bool                           valid;
   MovingImagePointType           mappedPoint;
--- a/Modules/Registration/Common/include/itkImageToImageMetric.hxx
+++ b/Modules/Registration/Common/include/itkImageToImageMetric.hxx
@@ -437,7 +437,6 @@ ImageToImageMetric<TFixedImage, TMovingImage>::MultiThreadingInitialize()
     this->m_BSplineTransformIndicesArray.SetSize(1, 1);
     this->m_BSplinePreTransformPointsArray.resize(1);
     this->m_WithinBSplineSupportRegionArray.resize(1);
-    this->m_BSplineTransformIndices.SetSize(1);
 
     delete[] this->m_ThreaderBSplineTransformWeights;
     this->m_ThreaderBSplineTransformWeights = nullptr;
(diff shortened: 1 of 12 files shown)


N-Dekker:
Does anyone have a suggestion?  https://open.cdash.org/viewBuildError.php?type=1&buildid=7436933 (Linux-Build5098-PR2712-BSplineInterpolationWeightFunction-FixedArray-Python) says:
> itkBSplineInterpolationWeightFunction: warning(4): ITK type not wrapped, or currently not known: itk::FunctionBase< itk::ContinuousIndex< double, 2 >, itk::FixedArray< double, 9 > >

> Wrapping/Typedefs/itkBSplineInterpolationWeightFunction.i:163: Warning 401: Maybe you forgot to instantiate 'itk::FunctionBase< itk::ContinuousIndex< double,3 >,itk::FixedArray< double,64 > >' using %template.

Modules/Registration/Common/include/itkImageToImageMetric.hxx
```diff
@@ -457,15 +456,13 @@ ImageToImageMetric<TFixedImage, TMovingImage>::MultiThreadingInitialize()
     }
     else
     {
-      this->m_BSplineTransformWeights.SetSize(this->m_NumBSplineWeights);
       this->m_BSplineTransformIndices.SetSize(this->m_NumBSplineWeights);
```
dzenanz, line 459: Could this variable receive the same constant treatment?

dzenanz (commented):
> itk::FixedArray< double, 9 >

I think fixed arrays etc are wrapped by default only up to dimension 4.

N-Dekker:
@dzenanz 
> I think fixed arrays etc are wrapped by default only up to dimension 4.

Would it be OK to wrap `FixedArray<double, 9>` and `FixedArray<double, 64>` as well, by default? If so, can you please tell me how to do that?

I _think_ that both `FunctionBase<ContinuousIndex<double, 2>, FixedArray<double, 9>>` and `FunctionBase<ContinuousIndex<double, 3>, FixedArray<double, 64>>` would then also need to be wrapped, but I guess that should be sufficient.

Modules/Registration/Common/include/itkImageToImageMetric.hxx
N-Dekker: > Could this variable receive the same constant treatment?

@dzenanz Indeed! I think that `m_BSplineTransformIndices` could also be declared as `FixedArray` (rather than `Array`), which would probably yield even more performance gain  😃  Would you want to have it included with the very same pull request, or would it be OK to do that in another PR?

dzenanz:
I am not sure whether anything needs to be done for `FunctionBase`, but if it does it should probably be done [here](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/wrapping/itkFunctionBase.wrap).

[`FixedArray.wrap`](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/wrapping/itkFixedArray.wrap) is incredibly simple, and I don't know where `WRAPPER_TEMPLATES` is used. https://itk.org/ITKSoftwareGuide/html/Book1/ITKSoftwareGuide-Book1ch9.html does not shine light on it.

Modules/Registration/Common/include/itkImageToImageMetric.hxx
dzenanz: I prefer it done as part of this PR.

thewtex:
`FixedArray` wrapping types are specified here:

https://github.com/InsightSoftwareConsortium/ITK/blob/34f3db8d8ad0ba848cbbfee071edf327e09f6a96/Wrapping/WrapITKTypes.cmake#L115-L152

Modules/Registration/Common/include/itkImageToImageMetric.hxx
N-Dekker: @dzenanz Added another commit to this PR, to declare the indices as FixedArray: "PERF: Use FixedArray for BSplineBaseTransform ParameterIndexArrayType". Indeed, even more PERF gain! 😃

N-Dekker:
> `FixedArray` wrapping types are specified here:

@thewtex Thanks Matt! Please check my attempt to wrap `FixedArray<double, 9>` and `FixedArray<double, 64>`, with this PR:
https://github.com/InsightSoftwareConsortium/ITK/blob/7c7945584767a48ad4370413d6e6effab44a985d/Wrapping/WrapITKTypes.cmake#L152-L155 

Unfortunately [Darwin-Build5152-PR2712-BSplineInterpolationWeightFunction-FixedArray-Python](https://open.cdash.org/viewBuildError.php?type=1&buildid=7438765) still says:

> itkBSplineInterpolationWeightFunction: warning(4): ITK type not wrapped, or currently not known: itk::FunctionBase< itk::ContinuousIndex< double, 2 >, itk::FixedArray< double, 9 > >
> itkBSplineInterpolationWeightFunction: warning(4): ITK type not wrapped, or currently not known: itk::FunctionBase< itk::ContinuousIndex< double, 3 >, itk::FixedArray< double, 64 > >

Do you have more suggestions?

dzenanz:
I guess we need to add something to https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/wrapping/itkFunctionBase.wrap. I am not quite certain what.

N-Dekker:
> I guess we need to add something to https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/wrapping/itkFunctionBase.wrap. I am not quite certain what.

@dzenanz Thanks! I'm not certain at all, but I gave it a try:

    itk_wrap_template("${ITKM_CID2}${ITKM_FAD9}" "${ITKT_CID2}, ${ITKT_FAD9}") 
    itk_wrap_template("${ITKM_CID3}${ITKM_FAD64}" "${ITKT_CID3}, ${ITKT_FAD64}") 

https://github.com/InsightSoftwareConsortium/ITK/blob/90b6738d33d61df3419af8e6bca6a043ea3acc5d/Modules/Core/Common/wrapping/itkFunctionBase.wrap#L29-L31

I'd be surprised if it would work directly, but we'll see.

Wrapping/WrapITKTypes.cmake
```diff
@@ -148,6 +148,13 @@ WRAP_TYPE("itk::FixedArray" "FA" "itkFixedArray.h")
     ADD_TEMPLATE("${ITKM_SC}${d}" "${ITKT_SC},${d}")
     ADD_TEMPLATE("${ITKM_B}${d}"  "${ITKT_B},${d}")
   endforeach()
+
+  # Wrap FixedArray for BSplineInterpolationWeightFunction:
+  ADD_TEMPLATE("${ITKM_D}9" "${ITKT_D},9")
+  ADD_TEMPLATE("${ITKM_D}16" "${ITKT_D},16")
+  ADD_TEMPLATE("${ITKM_UL}16" "${ITKT_UL},16")
```
dzenanz, line 155: Shouldn't D and F be enough? Can F be chosen for interpolation precision? What sense does it make to wrap UL?

N-Dekker: > Shouldn't D and F be enough? Can F be chosen for interpolation precision? What sense does it make to wrap UL?

@dzenanz Honestly I'm clueless. https://open.cdash.org/viewBuildError.php?type=1&buildid=7441714 says:

> itkBSplineBaseTransform: warning(4): ITK type not wrapped, or currently not known: itk::FixedArray< unsigned long, 16 >

Just doing trial and error now, but please tell me if you have a suggestion!

dzenanz: If the build system asks for it, add it 😄

That answers all my previous questions.

dzenanz (approved):

thewtex (approved):
@N-Dekker awesome work! :1st_place_medal: :snake: :ringed_planet: 

Merging for 5.3 RC1

CC @vicory @tbirdso

PR #2716, opened 2021-09-05
PERF: `BSplineKernelFunction::FastEvaluate(u)`: a faster way to evaluate BSplineKernelFunction

N-Dekker:
Added a static member function `BSplineKernelFunction::FastEvaluate(u)`. Equivalent to the existing `BSplineKernelFunction::Evaluate(u)` member function. Allows writing faster code, because it is non-virtual, and
because it does not require the creation of a `BSplineKernelFunction` object.

Removed the dynamically allocated `m_Kernel` data member from `BSplineInterpolationWeightFunction` and replaced the virtual function call `m_Kernel->Evaluate(x)` by a faster static member function call, to the newly added `BSplineKernelFunction::FastEvaluate` member function.

### DIFF
--- a/Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h
+++ b/Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h
@@ -20,7 +20,6 @@
 
 #include "itkFunctionBase.h"
 #include "itkContinuousIndex.h"
-#include "itkBSplineKernelFunction.h"
 #include "itkArray.h"
 #include "itkArray2D.h"
 #include "itkMath.h"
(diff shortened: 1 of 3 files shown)


dzenanz (approved):
Can you comment on the performance improvement @N-Dekker?

N-Dekker:
> Can you comment on the performance improvement @N-Dekker?

@dzenanz  Not yet, but I'm pretty sure that removing both the dynamic allocation of the kernel and the virtual function call does improve the performance.

thewtex (approved):

ntustison (approved):

Issue #2676, opened 2021-08-03
'Superclass' has not been declared


dzenanz:
It looks like #2653 introduced a regression failure in `itkLabelGeometryImageFilterTest`:
[open.cdash.org/viewBuildError.php?onlydeltap&buildid=7354628](open.cdash.org/viewBuildError.php?onlydeltap&buildid=7354628):
```text
In file included from /home/kitware/DashboardsITK/Modules/Nonunit/Review/test/itkLabelGeometryImageFilterTest.cxx:26:0:
/home/kitware/Dashboards/src/ITK/Modules/Nonunit/Review/test/itkLabelGeometryImageFilterTest.cxx: In function 'int LabelGeometryImageFilterTest(std::__cxx11::string, std::__cxx11::string, std::__cxx11::string, std::__cxx11::string, std::__cxx11::string)':
/home/kitware/DashboardsITK/Modules/Core/TestKernel/include/itkTestingMacros.h:69:51: error: 'Superclass' has not been declared
   std::cout << "Name of Superclass = " << object->Superclass::GetNameOfClass() << std::endl;                           \
                                                   ^
[CTest: warning matched] /home/kitware/DashboardsITK/Modules/Core/TestKernel/include/itkTestingMacros.h:69:51: note: in definition of macro 'ITK_EXERCISE_BASIC_OBJECT_METHODS'
   std::cout << "Name of Superclass = " << object->Superclass::GetNameOfClass() << std::endl;                           \
                                                   ^~~~~~~~~~
/home/kitware/DashboardsITK/Modules/Core/TestKernel/include/itkTestingMacros.h:79:28: error: 'Superclass' has not been declared
   if (!std::strcmp(object->Superclass::GetNameOfClass(), #SuperclassName))                                             \
                            ^
[CTest: warning matched] /home/kitware/DashboardsITK/Modules/Core/TestKernel/include/itkTestingMacros.h:79:28: note: in definition of macro 'ITK_EXERCISE_BASIC_OBJECT_METHODS'
   if (!std::strcmp(object->Superclass::GetNameOfClass(), #SuperclassName))                                             \
                            ^~~~~~~~~~
```
This was also [encountered earlier](https://github.com/InsightSoftwareConsortium/ITK/pull/2286#issuecomment-779382543), #2319 attempted to clarify the error message.

dzenanz:
[...]

PR #2720, opened 2021-09-08
COMP: Avoid Superclass type alias GCC compile errors

thewtex:
GCC is not resolving `object->Superclass` in some cases even though the
definition is available and other compilers are resolving the
definition.

xref https://github.com/InsightSoftwareConsortium/ITK/issues/2676

### DIFF
--- a/Modules/Core/TestKernel/include/itkTestingMacros.h
+++ b/Modules/Core/TestKernel/include/itkTestingMacros.h
@@ -62,7 +62,22 @@ namespace itk
 #  define TEST_SET_GET_BOOLEAN ITK_TEST_SET_GET_BOOLEAN
 #endif
 
-// object's Class must be specified to build on sun studio
+/* clang-format off */
+#if defined(__GNUC__)
+#define ITK_EXERCISE_BASIC_OBJECT_METHODS(object, ClassName, SuperclassName)                                           \
+  object->Print(std::cout);                                                                                            \
+  std::cout << "Name of Class = " << object->GetNameOfClass() << std::endl;                                            \
+  if (!std::strcmp(object->GetNameOfClass(), #ClassName))                                                              \
+  {                                                                                                                    \
+    std::cout << "Class name is correct" << std::endl;                                                                 \
+  }                                                                                                                    \
+  else                                                                                                                 \
+  {                                                                                                                    \
+    std::cerr << "Class name provided does not match object's NameOfClass" << std::endl;                               \
+    return EXIT_FAILURE;                                                                                               \
+  }                                                                                                                    \
+  ITK_MACROEND_NOOP_STATEMENT
+#else // not GCC
 #define ITK_EXERCISE_BASIC_OBJECT_METHODS(object, ClassName, SuperclassName)                                           \
   object->Print(std::cout);                                                                                            \
   std::cout << "Name of Class = " << object->GetNameOfClass() << std::endl;                                            \
@@ -86,6 +101,8 @@ namespace itk
     return EXIT_FAILURE;                                                                                               \
   }                                                                                                                    \
   ITK_MACROEND_NOOP_STATEMENT
+#endif // GCC
+/* clang-format on */
 
 #define ITK_TRY_EXPECT_EXCEPTION(command)                                                                              \
   try                                                                                                                  \


Modules/Core/TestKernel/include/itkTestingMacros.h
```diff
@@ -62,30 +62,45 @@ namespace itk
...
+      std::cout << "Class name is correct" << std::endl;                                                               \
+    }                                                                                                                  \
+    else                                                                                                               \
+    {                                                                                                                  \
+      std::cerr << "Class name provided does not match object's NameOfClass" << std::endl;                             \
+      return EXIT_FAILURE;                                                                                             \
+    }                                                                                                                  \
+    ITK_MACROEND_NOOP_STATEMENT
+#else // not GCC
+object->Print(std::cout);
```
jhlegarreta, line 81: Should the body of the `else` statement be indented? Are the line continuation backslashes missing in the body of the `else` statement?

jhlegarreta (commented):
Although I do not have a better proposal (sorry), I am not sure avoiding to print the superclass name is what we would like as a mid-/long-term fix.

I find particularly worrying that this is only happening for a few classes in the review module; the macro is used extensively throughout the code and the faulty statements are working well. So I think there is something more behind it that we are not being able to identify.

As for the proposed fix, if this is what we aim at for now, am I wrong when saying that the only statements that should need to be within the `if/else` are those related to the superclass? i.e. That the rest of the code is exactly the same, and we could avoid duplicating it?

Leengit:
In #2676, the `object` is a smart pointer rather than a C++ pointer, and perhaps that is the reason that `object->Superclass::GetNameOfClass()` is confusing to the compiler.  Would a simple switch to `(*object).Superclass::GetNameOfClass()` help?  Unfortunately I can't test out this or other solutions on my own system because the original code compiles just fine on my system.

Modules/Core/TestKernel/include/itkTestingMacros.h
thewtex: > Although I do not have a better proposal (sorry), I am not sure avoiding to print the superclass name is what we would like as a mid-/long-term fix.

Ideally we would find the bug in GCC. However, the condition that causes this is not obvious and compilers are complicated. Letting false-positive failures clutter the dashboard prevents long term progress by masking other issues.

> I find particularly worrying that this is only happening for a few classes in the review module; the macro is used extensively throughout the code and the faulty statements are working well. So I think there is something more behind it that we are not being able to identify.

Yes, we should leave #2676 open, but that should not prevent this from cluttering the dashboard. The check is covered by Clang, Visual Studio. We are potentially missing other problems that GCC could bring up.

> As for the proposed fix, if this is what we aim at for now, am I wrong when saying that the only statements that should need to be within the if/else are those related to the superclass? 

Yes, the superclass statements are removed.

> That the rest of the code is exactly the same, and we could avoid duplicating it?

There is not a clean way to avoid duplicating it that I am aware of because of the use of macros inside macros.

jhlegarreta: :+1: Sounds reasonable.

thewtex:
> Would a simple switch to (*object).Superclass::GetNameOfClass() help? 

This results in the same error with GCC 9.3.0.

Leengit (approved):
Looks good.

jhlegarreta:
Windows complaints might have to do with the comments in https://github.com/InsightSoftwareConsortium/ITK/pull/2720#discussion_r704398000:
https://open.cdash.org/viewBuildError.php?buildid=7444074

Leengit:
Grasping at straws, but ... in, e.g., `Modules/Nonunit/Review/include/itkLabelGeometryImageFilter.h`, should `itkTypeMacro` be invoked before `itkNewMacro`?  That might be helpful if, deep inside, `itkNewMacro` can print debug output that makes use of the `GetNameOfClass` method defined by `itkTypeMacro`.  (Yeah, the compiler should be figuring this out, even if out of order, ....)

Modules/Core/TestKernel/include/itkTestingMacros.h
```diff
@@ -62,30 +62,45 @@ namespace itk
 #  define TEST_SET_GET_BOOLEAN ITK_TEST_SET_GET_BOOLEAN
 #endif
 
+#if defined(__GNUC__)
 // object's Class must be specified to build on sun studio
```
dzenanz, line 66: I guess we can remove this comment now?

dzenanz (commented):
>  Would a simple switch to (*object).Superclass::GetNameOfClass() help?

I am trying this on my Linux box.

dzenanz:
Lee's suggestion does not work:
```text
In file included from /home/dzenan/Dashboards/Tests/ITK/Modules/Nonunit/Review/test/itkLabelGeometryImageFilterTest.cxx:26:
/home/dzenan/Dashboards/Tests/ITK/Modules/Nonunit/Review/test/itkLabelGeometryImageFilterTest.cxx: In function ‘int LabelGeometryImageFilterTest(std::string, std::string, std::string, std::string, std::string)’:
/home/dzenan/Dashboards/Tests/ITK/Modules/Core/TestKernel/include/itkTestingMacros.h:68:53: error: ‘Superclass’ has not been declared
   68 |   std::cout << "Name of Superclass = " << (*object).Superclass::GetNameOfClass() << std::endl;                         \
      |                                                     ^~~~~~~~~~
/home/dzenan/Dashboards/Tests/ITK/Modules/Core/TestKernel/include/itkTestingMacros.h:68:53: note: in definition of macro ‘ITK_EXERCISE_BASIC_OBJECT_METHODS’
   68 |   std::cout << "Name of Superclass = " << (*object).Superclass::GetNameOfClass() << std::endl;                         \
      |                                                     ^~~~~~~~~~
/home/dzenan/Dashboards/Tests/ITK/Modules/Core/TestKernel/include/itkTestingMacros.h:78:30: error: ‘Superclass’ has not been declared
   78 |   if (!std::strcmp((*object).Superclass::GetNameOfClass(), #SuperclassName))                                           \
      |                              ^~~~~~~~~~
/home/dzenan/Dashboards/Tests/ITK/Modules/Core/TestKernel/include/itkTestingMacros.h:78:30: note: in definition of macro ‘ITK_EXERCISE_BASIC_OBJECT_METHODS’
   78 |   if (!std::strcmp((*object).Superclass::GetNameOfClass(), #SuperclassName))                                           \
      |                              ^~~~~~~~~~
```

jhlegarreta:
> Grasping at straws, but ... in, e.g., `Modules/Nonunit/Review/include/itkLabelGeometryImageFilter.h`, should `itkTypeMacro` be invoked before `itkNewMacro`? That might be helpful if, deep inside, `itkNewMacro` can print debug output that makes use of the `GetNameOfClass` method defined by `itkTypeMacro`. (Yeah, the compiler should be figuring this out, even if out of order, ....)

I was tempted to push a PR with that, but I found many counter-examples (classes whose RTTI macro declaration comes after the `New` macro and whose tests exercise the basic object methods, and the compiler seems not to complain), e.g.:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkImportImageFilter.h#L65
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/test/itkImportImageTest.cxx#L44

Even in the `Review` module itself:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Nonunit/Review/include/itkAreaClosingImageFilter.h#L99
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Nonunit/Review/test/itkAreaClosingImageFilterTest.cxx#L52

Modules/Core/TestKernel/include/itkTestingMacros.h
thewtex: >Should the body of the else statement be indented? Are the line continuation backslashes missing in the body of the else statement?

`clang-format` did a bit too much formatting :-)

jhlegarreta (approved):
Not sure how to make the linter happy.

Thanks for doing this @thewtex.

dzenanz:
If we can't make clang-format happy, there is `/* clan-format: off */` (or something similar) to disable checking.

dzenanz (approved):

dzenanz (approved):

hjmjohnson (approved):

dyollb:
Maybe something like this would work:

```cpp
#if defined(__GNUC__)
#define ITK_EXERCISE_BASIC_OBJECT_METHODS(object, ClassName, SuperclassName)                                           \
  object->Print(std::cout);                                                                                            \
  std::cout << "Name of Class = " << object->GetNameOfClass() << std::endl;                                            \
  using self_t = std::remove_reference<decltype(*object)>::type; \
  using parent_t = typename self_t::Superclass; \
  std::cout << "Name of Superclass = " << dynamic_cast<parent_t&>(*object).GetNameOfClass() << std::endl; \
  ITK_MACROEND_NOOP_STATEMENT
```

PR #2721, opened 2021-09-08
PERF: Use FixedArray for table within BSplineInterpolationWeightFunction

N-Dekker:
Replaced `itk::Array2D` by `itk::FixedArray`, used as private `TableType`
of `BSplineInterpolationWeightFunction`. This is the type of a table,
used internally by its `Evaluate` member function.

Declared this table `const`, and initialized the table by an in-class
default-member-initializer. Defaulted the default-constructor.

Follow-up to pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2712
commit bc7c5dff2d6a29815bce50cbe7fd8bb5476a4a2b
"PERF: Use FixedArray for BSplineBaseTransform ParameterIndexArrayType"
commit 9bf745bda51044c90208c7b0776e87d7c86adb11
"PERF: Use FixedArray for BSplineInterpolationWeightFunction OutputType"

### DIFF
--- a/Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h
+++ b/Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h
@@ -22,6 +22,7 @@
 #include "itkContinuousIndex.h"
 #include "itkArray.h"
 #include "itkArray2D.h"
+#include "itkIndexRange.h"
 #include "itkMath.h"
 
 namespace itk
@@ -112,15 +113,21 @@ public:
 #endif
 
 protected:
-  BSplineInterpolationWeightFunction();
+  BSplineInterpolationWeightFunction() = default;
   ~BSplineInterpolationWeightFunction() override = default;
 
 private:
   /** Lookup table type. */
-  using TableType = Array2D<unsigned int>;
+  using TableType = FixedArray<IndexType, NumberOfWeights>;
 
   /** Table mapping linear offset to indices. */
-  TableType m_OffsetToIndexTable;
+  const TableType m_OffsetToIndexTable{ [] {
+    TableType     table;
+    // Note: Copied the constexpr value `SupportSize` to a temporary, `SizeType{ SupportSize }`, to prevent a GCC
+    // (Ubuntu 7.5.0-3ubuntu1~18.04) link error, "undefined reference to `SupportSize`".
+    std::copy_n(ZeroBasedIndexRange<SpaceDimension>(SizeType{ SupportSize }).cbegin(), NumberOfWeights, table.begin());
+    return table;
+  }() };
 };
 } // end namespace itk
 
--- a/Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.hxx
+++ b/Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.hxx
@@ -23,30 +23,9 @@
 #include "itkImage.h"
 #include "itkMatrix.h"
 #include "itkMath.h"
-#include "itkIndexRange.h"
 
 namespace itk
 {
-/** Constructor */
-template <typename TCoordRep, unsigned int VSpaceDimension, unsigned int VSplineOrder>
-BSplineInterpolationWeightFunction<TCoordRep, VSpaceDimension, VSplineOrder>::BSplineInterpolationWeightFunction()
-{
-  // Initialize offset to index lookup table
-  m_OffsetToIndexTable.set_size(Self::NumberOfWeights, SpaceDimension);
-
-  constexpr auto supportSize = Self::SupportSize;
-  unsigned int   counter = 0;
-
-  for (const IndexType index : ZeroBasedIndexRange<VSpaceDimension>(supportSize))
-  {
-    for (unsigned int j = 0; j < SpaceDimension; ++j)
-    {
-      m_OffsetToIndexTable[counter][j] = index[j];
-    }
-    ++counter;
-  }
-}
-
 /** Compute weights for interpolation at continuous index position */
 template <typename TCoordRep, unsigned int VSpaceDimension, unsigned int VSplineOrder>
 typename BSplineInterpolationWeightFunction<TCoordRep, VSpaceDimension, VSplineOrder>::WeightsType


Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h
```diff
@@ -112,15 +113,19 @@ class ITK_TEMPLATE_EXPORT BSplineInterpolationWeightFunction
...
 private:
   /** Lookup table type. */
-  using TableType = Array2D<unsigned int>;
+  using TableType = FixedArray<IndexType, NumberOfWeights>;
 
   /** Table mapping linear offset to indices. */
-  TableType m_OffsetToIndexTable;
+  const TableType m_OffsetToIndexTable{ [] {
+    TableType     table;
+    std::copy_n(ZeroBasedIndexRange<SpaceDimension>(SupportSize).cbegin(), NumberOfWeights, table.begin());
```
dzenanz, line 126: Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h:125:17: error: 'ZeroBasedIndexRange' was not declared in this scope

When an address of a `static constexpr symbol` is taken, it needs to be defined somewhere in the translation unit, hence: `itkBSplineInterpolationWeightFunctionTest.cxx:(.text+0x2539): undefined reference to `itk::BSplineInterpolationWeightFunction<double, 3u, 3u>::SupportSize'`

dzenanz (commented):
I guess this exposes a need for more follow-up, `Modules/Core/Transform/include/itkBSplineTransform.hxx:629:44: error: too many arguments to function call, expected single argument 'index', have 3 arguments`

Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h
N-Dekker: > Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h:125:17: error: 'ZeroBasedIndexRange' was not declared in this scope

Thanks @dzenanz This error message must have been from a previous build. With force-push https://github.com/InsightSoftwareConsortium/ITK/compare/4a6a1de47f3da0563d6f39a34729cfe42a0792a1..ffc90b13d0f38b82c188dc2404611236ebf5b90f I added the missing `#include "itkIndexRange.h"`, which should fix the error. And indeed, on other platforms, this PR now builds fine!

dzenanz (approved):

Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h
```diff
@@ -112,15 +113,21 @@ class ITK_TEMPLATE_EXPORT BSplineInterpolationWeightFunction
...
 private:
   /** Lookup table type. */
-  using TableType = Array2D<unsigned int>;
+  using TableType = FixedArray<IndexType, NumberOfWeights>;
 
   /** Table mapping linear offset to indices. */
-  TableType m_OffsetToIndexTable;
+  const TableType m_OffsetToIndexTable{ [] {
+    TableType     table;
+    // Note: Copied the constexpr value `SupportSize` to a temporary, `SizeType{ SupportSize }`, to prevent a GCC
```
dzenanz, line 126: Good solution!

dzenanz (approved):

Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h
N-Dekker: @dzenanz Thanks! It appears that GCC doesn't like to have a const-reference to a static `constexpr` data member. The syntax `T{x}` produces a copy that may be passed to a const-reference parameter, so that appears to solve the errors. (The copying itself is easily optimized by the compiler, I assume.) 

Note that this issue may appear more often with a constexpr static `itk::Size<N>` data member than with constexpr static data members of a built-in type, because `itk::Size<N>` is more likely to be passed to a const-reference parameter, while built-in types are more likely passed by-value. (Pass-by-value is not a problem, as far as I can see.)

N-Dekker:
@thewtex Would this PR still be in time for ITK 5.3? It would be nice to have it included in the same release as the other two BSpline "Use FixedArray for ..." commits, from pull request #2712, as they kinda belong together 😃

dzenanz:
There is a greater chance of that if you merge it now, or as soon as the CI passes 😄

Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h
Leengit: > It appears that GCC doesn't like to have a const-reference to a static constexpr data member. 

This can happen when the static constexpr member is declared within the class declaration but not defined outside of the class declaration, and then a reference is used.  In the associated hxx file define the member -- omit the `static` keyword in the hxx file and don't initialize the variable if it is already initialized within the class declaration.

N-Dekker: > In the associated hxx file define the member -- omit the `static` keyword in the hxx file and don't initialize the variable if it is already initialized within the class declaration.

Doing so might solve the problem, thanks @Leengit However, I feel that it's unfortunate if it is necessary to do it that way. The following is accepted, even by GCC 7.5: https://godbolt.org/z/zb79f7v1E as below:

```
int square(const long long& ref)
{
    return static_cast<int>(ref*ref);
}

int main()
{
    constexpr int i{42};
    return square(i);
}
```

You see, the `ref` parameter refers to a `long long` that isn't explicitly declared or defined anywhere! So do you think it's a compiler issue, or a language issue, that we need to work around those "undefined reference" errors?

Leengit: I don't claim to know what the GCC folks are thinking nor what the C++ language spec says on the matter, but I've always imagined that the thinking is something like the following.  The implementation of the likes of
```c++
static const int CrucialParameter = 3;
```
(or `constexpr` instead of `const`) within a class declaration is usually never allocated in memory.  Usually the compiler treats it much as
```c++
#define CrucialParameter 3
```
would be treated (though with class namespace scoping).  That works great unless one wants a pointer or reference to `CrucialParameter`.  The way to tell GCC about this added need is define the member too, sometime after the class declaration is closed:
```
const inst MyClass::CrucialParameter;
```

The code sample you supply doesn't includes a `static const` so I'm not sure that it applies to the present case.

N-Dekker:
> There is a greater chance of that if you merge it now, or as soon as the CI passes 😄

Thanks for the suggestion @dzenanz As you know I'd rather not merge my own PRs. But maybe, if this one gets one more approval, I'll just do it.  😸 

@Leengit I feel your comment https://github.com/InsightSoftwareConsortium/ITK/pull/2721#discussion_r705396530 is relevant, but it's not a show-stopper to this particular PR. OK?

hjmjohnson (approved):

Leengit:
> > There is a greater chance of that if you merge it now, or as soon as the CI passes smile
> 
> Thanks for the suggestion @dzenanz As you know I'd rather not merge my own PRs. But maybe, if this one gets one more approval, I'll just do it. smile_cat
> 
> @Leengit I feel your comment [#2721 (comment)](https://github.com/InsightSoftwareConsortium/ITK/pull/2721#discussion_r705396530) is relevant, but it's not a show-stopper to this particular PR. OK?

Not a show stopper.  Don't wait on account of me!

dzenanz:
I now see that Matt has tagged 5.3 RC1. This PR didn't make the cut 😞

Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h
N-Dekker: > That works great unless one wants a pointer or reference to `CrucialParameter`. The way to tell GCC about this added need is define the member too, sometime after the class declaration is closed:

@Leengit OK, but then, do you think ideally all "static constexpr" data should have a separate definition, apart from its initialization, in ITK? A quick search suggests that ITK now declares more than 2000 "static constexpr" constants already! (Quite some of them from ThirdParty, I have to admit.)

N-Dekker:
> I now see that Matt has tagged 5.3 RC1. This PR didn't make the cut 😞

@dzenanz @thewtex  So sad for this lonely little PR to be apart from the other two BSpline "Use FixedArray for ..." commits  😢 Any chance that RC2 will bring them together again?

Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h
Leengit: I suppose if we think it will happen often enough that a user of ITK will want a pointer or reference to `CrucialParameter` then we might want to proactively define it outside the class declaration.  I have no opinion on that possibility.

Except for that circumstance (and except for ITK's direct use of a pointer or reference), the optimization of not allocating memory is the desired outcome; we would deliberately not add the definition outside of the class declaration.

dzenanz:
RC2 will be cut from master, so they will be reunited 💕

thewtex:
Yes, 5.3 RC 2 will include these. Before then, there there are new dashboard errors to address https://open.cdash.org/viewBuildError.php?buildid=7449043 😸

Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h
```diff
@@ -112,15 +113,21 @@ class ITK_TEMPLATE_EXPORT BSplineInterpolationWeightFunction
...
+    // (Ubuntu 7.5.0-3ubuntu1~18.04) link error, "undefined reference to `SupportSize`".
+    std::copy_n(ZeroBasedIndexRange<SpaceDimension>(SizeType{ SupportSize }).cbegin(), NumberOfWeights, table.begin());
```
Leengit, line 128: If it isn't making sense to defined `SupportSize` outside the class declaration so that references and pointers to it are enabled -- perhaps because we don't want to allocate that memory just to support the reference -- similar can be accomplished via a local variable to this lambda.
```c++
const auto supportSize {Self::SupportSize};
```
As to `const` vs. `constexpr` here, I think the former might work out better -- the latter might be interpreted as a hint to the compiler that we don't want the memory allocated.

N-Dekker: @Leengit  Indeed, it appears that the link error could also be avoided by a local `const auto supportSize` within the lambda. But personally I don't really like to have an extra local variable. For now I think the syntax `SizeType{ SupportSize }` is just fine.

In general, it might become cumbersome to spell out the type name of the `constexpr` data member every time it is passed by reference. Instead _maybe_ a helper function could be useful:

    auto CopyToNonConstexprValue(T value) { return value; }

The value of the constexpr `SupportSize` could then be passed to a function like `ProcessSize(const SizeType&)`, as follows:

    ProcessSize( CopyToNonConstexprValue(SupportSize) );

Do you like that?   😺

Leengit: I fear that it will give the same errors that we are already / still seeing in cdash.  But if it doesn't then I like it.

N-Dekker: @Leengit Try it out at https://godbolt.org/z/YGTh9q1Po  😃  As follows (using GCC 10.3):

```
struct SizeType {};

struct Wrapper
{
  constexpr static SizeType ConstexprValue{};
};

void ProcessSize(const SizeType&) {}

template <typename T>
auto CopyToNonConstexprValue(T value) { return value; }

int main()
{
    ProcessSize(CopyToNonConstexprValue(Wrapper::ConstexprValue)); // OK
    ProcessSize(Wrapper::ConstexprValue); // <== error: undefined reference
}

```

N-Dekker: @Leengit  Related section from https://github.com/cplusplus/draft/releases/download/n4892/n4892.pdf (C++ Working Draft, 2021-06-18): 

**D.7 Redeclaration of static constexpr data members [depr.static.constexpr]**

1 For compatibility with prior revisions of C++, a constexpr static data member may be redundantly redeclared
outside the class with no initializer. This usage is deprecated.
[Example 1:
```
struct A {
static constexpr int n = 5; // definition (declaration in C++2014)
};
constexpr int A::n; // redundant declaration (definition in C++2014)

```
— end example]

Leengit: Looks like a good idea.  Is this for C++23?  Do they have an alternative way of making references and pointers work for class static constexpr members, or are they saying that the class declaration needs to be sufficient?

Leengit:
I didn't try your `CopyToNonConstexprValue` example but ... looks like it works!

Leengit:
> I didn't try your `CopyToNonConstexprValue` example but ... looks like it works!

Like the use of a local variable, this will require the existence of a (implicit or explicit) copy constructor.  The deprecated technique of redundantly declaring the member variable outside of the class declaration allows any constructor.

PR #2729, opened 2021-09-09
BUG: Improve dimension-dependent BSplineInterpolationWeightFunction wrapping

thewtex:
Avoid build errors when we wrap for 4-D images.

### DIFF
--- a/Modules/Core/Common/wrapping/itkFunctionBase.wrap
+++ b/Modules/Core/Common/wrapping/itkFunctionBase.wrap
@@ -27,8 +27,12 @@ itk_wrap_class("itk::FunctionBase" POINTER)
     itk_wrap_template("${ITKM_PD${d}}RGBAD" "${ITKT_PD${d}},itk::RGBAPixel< double >")
 
     # Required by BSplineInterpolationWeightFunction
-    itk_wrap_template("${ITKM_CID2}${ITKM_FAD9}" "${ITKT_CID2}, ${ITKT_FAD9}")
-    itk_wrap_template("${ITKM_CID3}${ITKM_FAD64}" "${ITKT_CID3}, ${ITKT_FAD64}")
+    # Wrapping for spline order 3, components = (SplineOrder + 1)^SpaceDimension
+    set(comp 1)
+    foreach(i RANGE 1 ${d})
+      math(EXPR comp "${comp}*4")
+    endforeach()
+    itk_wrap_template("${ITKM_CID${d}}${ITKM_FAD${comp}}" "${ITKT_CID${d}}, ${ITKT_FAD${comp}}")
 
   endforeach()
 
--- a/Wrapping/WrapITKTypes.cmake
+++ b/Wrapping/WrapITKTypes.cmake
@@ -150,11 +150,15 @@ WRAP_TYPE("itk::FixedArray" "FA" "itkFixedArray.h")
   endforeach()
 
   # Wrap FixedArray for BSplineInterpolationWeightFunction:
-  ADD_TEMPLATE("${ITKM_D}9" "${ITKT_D},9")
-  ADD_TEMPLATE("${ITKM_D}16" "${ITKT_D},16")
-  ADD_TEMPLATE("${ITKM_UL}16" "${ITKT_UL},16")
-  ADD_TEMPLATE("${ITKM_D}64" "${ITKT_D},64")
-  ADD_TEMPLATE("${ITKM_UL}64" "${ITKT_UL},64")
+  foreach(d ${ITK_WRAP_IMAGE_DIMS})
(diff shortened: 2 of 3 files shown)


Modules/Core/Common/wrapping/itkFunctionBase.wrap
```diff
@@ -28,7 +28,15 @@ itk_wrap_class("itk::FunctionBase" POINTER)
...
     # Required by BSplineInterpolationWeightFunction
     itk_wrap_template("${ITKM_CID2}${ITKM_FAD9}" "${ITKT_CID2}, ${ITKT_FAD9}")
-    itk_wrap_template("${ITKM_CID3}${ITKM_FAD64}" "${ITKT_CID3}, ${ITKT_FAD64}")
+    foreach(d ${ITK_WRAP_IMAGE_DIMS})
+      # Wrapping for spline order 3, components = (SplineOrder + 1)^SpaceDimension
+      set(comp 4)
+      math(EXPR stop "${d}-1")
+      foreach(i RANGE 1 ${stop})
+        math(EXPR comp "${comp}*4")
+      endforeach()
```
Leengit, line 37: `foreach(i RANGE 1 ${stop})` [has undefined behavior](https://cmake.org/cmake/help/latest/command/foreach.html) if `${stop}` is 0, which could happen if 1 is ever a supplied value in the `ITK_WRAP_IMAGE_DIMS` list.  It would be safer to use
```cmake
set(comp 1)
foreach(i RANGE 1 ${d})
  math(EXPR comp "${comp}*4")
endforeach()
```

Wrapping/WrapITKTypes.cmake
```diff
@@ -151,10 +151,16 @@ WRAP_TYPE("itk::FixedArray" "FA" "itkFixedArray.h")
...
-  ADD_TEMPLATE("${ITKM_UL}16" "${ITKT_UL},16")
-  ADD_TEMPLATE("${ITKM_D}64" "${ITKT_D},64")
-  ADD_TEMPLATE("${ITKM_UL}64" "${ITKT_UL},64")
+  foreach(d ${ITK_WRAP_IMAGE_DIMS})
+    # Wrapping for spline order 3, components = (SplineOrder + 1)^SpaceDimension
+    set(comp 4)
+    math(EXPR stop "${d}-1")
+    foreach(i RANGE 1 ${stop})
+      math(EXPR comp "${comp}*4")
+    endforeach()
```
Leengit, line 160: See previous comment.

Modules/Core/Common/wrapping/itkFunctionBase.wrap
```diff
@@ -28,7 +28,15 @@ itk_wrap_class("itk::FunctionBase" POINTER)
 
     # Required by BSplineInterpolationWeightFunction
     itk_wrap_template("${ITKM_CID2}${ITKM_FAD9}" "${ITKT_CID2}, ${ITKT_FAD9}")
```
dzenanz, line 30: Shouldn't this line be replaced by the below loop too?

Wrapping/WrapITKTypes.cmake
```diff
@@ -151,10 +151,16 @@ WRAP_TYPE("itk::FixedArray" "FA" "itkFixedArray.h")
 
   # Wrap FixedArray for BSplineInterpolationWeightFunction:
   ADD_TEMPLATE("${ITKM_D}9" "${ITKT_D},9")
```
dzenanz, line 153: Same here.

Modules/Core/Common/wrapping/itkFunctionBase.wrap
Leengit: Looks like the loop is for cubic splines.  Looks like this one line adds quadratic splines for 2-dimensional images.

Modules/Core/Common/wrapping/itkFunctionBase.wrap
thewtex: :+1:

Modules/Core/Common/wrapping/itkFunctionBase.wrap
thewtex: This was needed, but it turns out that there was a mistake in the wrapping of BSplineInterpolationWeightFunction SplineOrder.

N-Dekker:
@thewtex Just for my understanding, did my commit 9bf745bda51044c90208c7b0776e87d7c86adb11 (Use FixedArray for BSplineInterpolationWeightFunction OutputType, PR #2712) cause a problem, that you're addressing here? I thought the CI was green for PR #2712!

thewtex:
@N-Dekker yes, we did not see the failure in CI because CI does not build 4-D for efficiency.

Merging for the 5.3 RC 1 Python packages.

PR #2731, opened 2021-09-12
FastEvaluate(u) for both BSplineKernelFunction _and_ BSplineDerivativeKernelFunction 

N-Dekker:
Added static member function `FastEvaluate(u)` to `BSplineDerivativeKernelFunction`, just like the one added to `BSplineKernelFunction`.

Used the `FastEvaluate(u)`  member functions of both `BSplineKernelFunction` _and_ `BSplineDerivativeKernelFunction` inside the implementation of `MattesMutualInformationImageToImageMetric` and `MattesMutualInformationImageToImageMetricv4`.

Follow-up to pull request #2716 "PERF: `BSplineKernelFunction::FastEvaluate(u)`: a faster way to evaluate BSplineKernelFunction"

### DIFF
--- a/Modules/Registration/Common/include/itkMattesMutualInformationImageToImageMetric.h
+++ b/Modules/Registration/Common/include/itkMattesMutualInformationImageToImageMetric.h
@@ -323,10 +323,6 @@ private:
   PDFValueType  m_FixedImageBinSize{ 0.0 };
   PDFValueType  m_MovingImageBinSize{ 0.0 };
 
-  /** Cubic BSpline kernel for computing Parzen histograms. */
-  typename CubicBSplineFunctionType::Pointer           m_CubicBSplineKernel;
-  typename CubicBSplineDerivativeFunctionType::Pointer m_CubicBSplineDerivativeKernel;
-
   /** Helper array for storing the values of the JointPDF ratios. */
   using PRatioType = PDFValueType;
   using PRatioArrayType = Array2D<PRatioType>;
--- a/Modules/Registration/Metricsv4/include/itkMattesMutualInformationImageToImageMetricv4.h
+++ b/Modules/Registration/Metricsv4/include/itkMattesMutualInformationImageToImageMetricv4.h
@@ -255,10 +255,6 @@ protected:
   PDFValueType  m_FixedImageBinSize;
   PDFValueType  m_MovingImageBinSize;
 
-  /** Cubic BSpline kernel for computing Parzen histograms. */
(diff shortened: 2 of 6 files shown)


Modules/Registration/Common/include/itkMattesMutualInformationImageToImageMetric.h
```diff
@@ -323,10 +323,6 @@ class ITK_TEMPLATE_EXPORT MattesMutualInformationImageToImageMetric
   PDFValueType  m_FixedImageBinSize{ 0.0 };
   PDFValueType  m_MovingImageBinSize{ 0.0 };
 
-  /** Cubic BSpline kernel for computing Parzen histograms. */
-  typename CubicBSplineFunctionType::Pointer           m_CubicBSplineKernel;
```
dzenanz, line 327: Do we need to guard this under `legacy` ifdef?

Modules/Registration/Metricsv4/include/itkMattesMutualInformationImageToImageMetricv4.h
```diff
@@ -255,10 +255,6 @@ class ITK_TEMPLATE_EXPORT MattesMutualInformationImageToImageMetricv4
   PDFValueType  m_FixedImageBinSize;
   PDFValueType  m_MovingImageBinSize;
 
-  /** Cubic BSpline kernel for computing Parzen histograms. */
```
dzenanz, line 258: Perhaps merge this commit with the previous one?

dzenanz (commented):
Mostly looks good

Modules/Registration/Common/include/itkMattesMutualInformationImageToImageMetric.h
N-Dekker: Both "cubic BSpline kernel" data members of `MattesMutualInformationImageToImageMetric` were private, and neither of them had getters or setters. So I don't think there is any possible reason to guard them under legacy ifdef. 

The corresponding "cubic BSpline kernel" data members of the v4 version (`MattesMutualInformationImageToImageMetricv4`) were not part of the public interface either. These two were declared `protected`, but I think they should also have been declared `private` in the first place. They were only accessed by its friend class, `MattesMutualInformationImageToImageMetricv4GetValueAndDerivativeThreader`. So I think they can also be removed without legacy support. Is that OK to you?

dzenanz: If they were not public, no need to put them through legacy process.

dzenanz (approved):

PR #2737, opened 2021-09-14
BUG: Disable all floating point exceptions 

hjmjohnson:
itk::FloatingPointExceptions::Disable()

In this commit, a flag is added to disable all floating-point exceptions, rather than just two exceptions. This bug was identified in the following discussion thread: https://discourse.itk.org/t/itk-not-loading-nrrd-files-any-more/4224/10. This fix will resolve an exception that occurs when loading NRRD images.


hjmjohnson:
Supercedes #2736

thewtex (approved):

dzenanz (approved):

dzenanz:
Does the Windows variant need to be touched too?
https://github.com/InsightSoftwareConsortium/ITK/blob/acedfab214ef9ccd147e47f85bf5d13e3949e973/Modules/Core/Common/src/itkFloatingPointExceptions_win.cxx#L45-L46

PR #2739, opened 2021-09-15
PERF: Remove unnecessary `IdentityTransform::m_ZeroJacobian`

N-Dekker:
`m_ZeroJacobian` appeared to be used only as a source object to copy to
a `jacobian` parameter. As an `Array2D`, it was using dynamically
allocated memory.


hjmjohnson (approved):
LGTM.

PR #2742, opened 2021-09-17
COMP: Copy BSplineInterpolationWeightFunction::SupportSize within lambda

N-Dekker:
Copied the constexpr static data member `SupportSize` to a local
variable, to prevent Clang linker errors, like:

> Undefined symbols for architecture x86_64:
> "itk::BSplineInterpolationWeightFunction::SupportSize", referenced from:
> itk::BSplineInterpolationWeightFunction::'lambda'()::operator()() const in itkBSplineInterpolationWeightFunctionTest.cxx.o

From Mac10.13-AppleClang-dbg-x86_64-static https://open.cdash.org/viewBuildError.php?type=0&buildid=7457677, reported by Sean McBride (@seanm).

### DIFF
--- a/Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h
+++ b/Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h
@@ -123,9 +123,11 @@ private:
   /** Table mapping linear offset to indices. */
   const TableType m_OffsetToIndexTable{ [] {
     TableType     table;
-    // Note: Copied the constexpr value `SupportSize` to a temporary, `SizeType{ SupportSize }`, to prevent a GCC
-    // (Ubuntu 7.5.0-3ubuntu1~18.04) link error, "undefined reference to `SupportSize`".
-    std::copy_n(ZeroBasedIndexRange<SpaceDimension>(SizeType{ SupportSize }).cbegin(), NumberOfWeights, table.begin());
+    // Note: Copied the constexpr value `SupportSize` to a local variable, to prevent a GCC
+    // (Ubuntu 7.5.0-3ubuntu1~18.04) link error, "undefined reference to `SupportSize`", and Clang
+    // (Mac10.13-AppleClang-dbg-x86_64-static) "Undefined symbols for architecture x86_64".
+    const auto    supportSize = SupportSize;
+    std::copy_n(ZeroBasedIndexRange<SpaceDimension>(supportSize).cbegin(), NumberOfWeights, table.begin());
     return table;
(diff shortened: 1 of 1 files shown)


Leengit:
@N-Dekker thank you for your perseverance on this issue!  #2721

You know me ... I push to avoid `auto` and to use `{}` constructor syntax instead of assignment syntax, so I'd write
```c++
const SizeType supportSize{ SupportSize };
```
... but you can safely ignore me without even responding to this part.

seanm:
I'm not a big fan of `auto` either. I find it makes code faster/easier to write, but harder to read, and readability is more important in my books.

Anyway, I'll test if this patch fixes things.

N-Dekker:
> I push to avoid `auto` and to use `{}` constructor syntax instead of assignment syntax, so I'd write
> 
> ```c++
> const SizeType supportSize{ SupportSize };
> ```

Thanks for your comment @Leengit To me, the syntax `TypeX variableX{ value }`  suggests that a _conversion_ takes place from the type of `value` to `TypeX`. _Or_ that `value` is copied into the very first data member of an aggregate of type `TypeX`.  I would typically use this syntax when that's my intention. 

On the other hand, the syntax `auto variableX = value` communicates most clearly to me that both `variableX` and `value` have the same type, and that `variableX` will be an identical copy of `value`.

Hope that clarifies my motivation  😃

seanm:
I confirm it fixes the link error.

Leengit:
> Thanks for your comment @Leengit To me, the syntax `TypeX variableX{ value }` suggests that a _conversion_ takes place from the type of `value` to `TypeX`.

You responded, so that's on you.  :-)  C++ is more restrictive with types conversions with `{}` syntax than with `=` syntax.

> _Or_ that `value` is copied into the very first data member of an aggregate of type `TypeX`. I would typically use this syntax when that's my intention.

Yes, it looks like that.  But I am thinking that to actually achieve that you'd have to use `{ {SupportSize} }` or maybe `={SupportSize}`.

> On the other hand, the syntax `auto variableX = value` communicates most clearly to me that both `variableX` and `value` have the same type, and that `variableX` will be an identical copy of `value`.

Yes, I agree that `auto` communicates "same type" very clearly.

Leengit (approved):
Looks good -- thanks!

hjmjohnson (approved):

dzenanz (approved):

N-Dekker:
@hjmjohnson @dzenanz @Leengit  Thanks for your approvals!

PR #2747, opened 2021-09-20
PERF: Remove protected `itk::Transform` data member `m_DirectionChange`

N-Dekker:
Observed ~29% reduction of `sizeof(itk::Transform<double>)`, from 248
bytes (before this commit) down to 176 bytes, using Visual C++ 2019
64-bit Release.

Follow-up to commit c42fb3f468c1acfd9b8ebd2fb99fe1d07e5b7e6b
"BUG: Remove extra m_DirectionChange from Transform." by Matt McCormick (@thewtex),
13 April 2012

### DIFF
--- a/Modules/Core/Transform/include/itkTransform.h
+++ b/Modules/Core/Transform/include/itkTransform.h
@@ -590,8 +590,6 @@ protected:
   PreservationOfPrincipalDirectionDiffusionTensor3DReorientation(const InputDiffusionTensor3DType &,
                                                                  const InverseJacobianPositionType &) const;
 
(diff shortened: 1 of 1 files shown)


dzenanz (approved):

N-Dekker:
BTW, I think the type `itk::Transform::DirectionChangeMatrix` may also be removed (even though the type is `public`), from https://github.com/InsightSoftwareConsortium/ITK/blob/eec9fe67056bfb5c693a4915b86eac9271e1dc83/Modules/Core/Transform/include/itkTransform.h#L172

Anyway I think removing the data member `m_DirectionChange` is the most beneficial.

PR #2748, opened 2021-09-21
COMP: Fix missing initialization braces warning

jhlegarreta:
Fix missing initialization braces warning.

Fixes:
```
[CTest: warning matched]
/Users/builder/externalModules/Filtering/ImageFeature/test/itkLaplacianImageFilterTest.cxx:62:35:
warning: suggest braces around initialization of subobject [-Wmissing-braces]
  itk::Vector<float, 2> spacing({ 0.5, 5.0 });
                                  ^~~~~~~~
                                  {       }
[CTest: warning suppressed] 1 warning generated.
```

reported at:
https://open.cdash.org/viewBuildError.php?type=1&buildid=7468732

## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [X] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)

PR #2756, opened 2021-09-24
COMP: Fix missing initialization braces warning

jhlegarreta:
Fix missing initialization braces warning.

Fixes:
````
[CTest: warning matched]
/Users/builder/externalModules/Filtering/LabelMap/test/itkLabelImageToLabelMapFilterTest.cxx:29:37:
warning: suggest braces around initialization of subobject [-Wmissing-braces]
  p_image->SetRegions(itk::Size<3>{ 0, 0, 0 });
                                    ^~~~~~~
                                    {      }
```

and
```
[CTest: warning matched]
/Users/builder/externalModules/Filtering/ImageFeature/test/itkLaplacianImageFilterTest.cxx:62:36:
warning: suggest braces around initialization of subobject [-Wmissing-braces]
  itk::Vector<float, 2> spacin{ { 0.5, 5.0 } };
                                  ^~~~~~~~
                                  {       }
```

reported at:
https://open.cdash.org/viewBuildError.php?type=1&buildid=7475017

## PR Checklist

### DIFF
--- a/Modules/Filtering/ImageFeature/test/itkLaplacianImageFilterTest.cxx
+++ b/Modules/Filtering/ImageFeature/test/itkLaplacianImageFilterTest.cxx
@@ -59,7 +59,7 @@ itkLaplacianImageFilterTest(int argc, char * argv[])
   ITK_TRY_EXPECT_NO_EXCEPTION(reader->Update());
 
   // Assign a spacing other than [1,1] for testing
-  itk::Vector<float, 2> spacing{ { 0.5, 5.0 } };
+  itk::Vector<float, 2> spacing{ { { 0.5, 5.0 } } };
   reader->GetOutput()->SetSpacing(spacing);
 
   // Set up filter
--- a/Modules/Filtering/LabelMap/test/itkLabelImageToLabelMapFilterTest.cxx
+++ b/Modules/Filtering/LabelMap/test/itkLabelImageToLabelMapFilterTest.cxx
@@ -24,9 +24,13 @@ void
 zeroSizeCase()
 {
   // test filter with zero sized image
-  auto p_filter = itk::LabelImageToLabelMapFilter<itk::Image<unsigned char, 3>>::New();
-  auto p_image = itk::Image<unsigned char, 3>::New();
-  p_image->SetRegions(itk::Size<3>{ 0, 0, 0 });
+  constexpr unsigned int ImageDimension = 3;
+  using PixelType = unsigned char;
+  using ImageType = itk::Image<PixelType, ImageDimension>;
+  auto p_filter = itk::LabelImageToLabelMapFilter<ImageType>::New();
+  auto p_image = ImageType::New();
+
+  // The image region is empty by default: do not set any region to it and allocate memory
   p_image->Allocate(true);
 
   p_filter->SetInput(p_image);


jhlegarreta:
Looks like #2748 did not make one of the compilers happy.

I am not being able to configure ITK locally due to some CMake error, so I cannot test locally if this compiles (at least on Windows).

@seanm can you please check that this makes happy `RogueResearch7` ? Thanks.

Modules/Filtering/ImageFeature/test/itkLaplacianImageFilterTest.cxx
```diff
@@ -59,7 +59,7 @@ itkLaplacianImageFilterTest(int argc, char * argv[])
   ITK_TRY_EXPECT_NO_EXCEPTION(reader->Update());
 
   // Assign a spacing other than [1,1] for testing
-  itk::Vector<float, 2> spacing{ { 0.5, 5.0 } };
+  itk::Vector<float, 2> spacing(0.5);
```
tbirdso, line 62: This is changing the test parameters. Spacing should be non-uniform [0.5, 5.0].

tbirdso (changes requested):

seanm:
That warning is pretty useless, and I've tried to disable it, see here:

https://open.cdash.org/build/7475017/notes

note presence of `-Wno-missing-braces`.  I'm stumped why it doesn't work...

Modules/Filtering/ImageFeature/test/itkLaplacianImageFilterTest.cxx
N-Dekker: What about `itk::MakeVector(0.5f, 5.0f)`?  😃

jhlegarreta: :100: Good catch. Addressed in 3265872.

dzenanz (approved):
Code looks good, but this is about pleasing the compilers 😄

Modules/Filtering/LabelMap/test/itkLabelImageToLabelMapFilterTest.cxx
```diff
@@ -24,9 +24,13 @@ void
...
-  auto p_filter = itk::LabelImageToLabelMapFilter<itk::Image<unsigned char, 3>>::New();
-  auto p_image = itk::Image<unsigned char, 3>::New();
-  p_image->SetRegions(itk::Size<3>{ 0, 0, 0 });
+  constexpr unsigned int ImageDimension = 3;
+  using PixelType = unsigned char;
+  using ImageType = itk::Image<PixelType, ImageDimension>;
+  auto p_filter = itk::LabelImageToLabelMapFilter<ImageType>::New();
+  auto p_image = ImageType::New();
+
+  p_image->SetRegions(ImageType::SizeType::Filled(0));
```
N-Dekker, line 33: An image region is already empty by default: 

https://github.com/InsightSoftwareConsortium/ITK/blob/v5.3rc01/Modules/Core/Common/include/itkImageRegion.h#L365

So the entire `p_image->SetRegions` call may be removed  😃

jhlegarreta: :+1: Removed in 33d1d0d.

Modules/Filtering/ImageFeature/test/itkLaplacianImageFilterTest.cxx
jhlegarreta: > What about itk::MakeVector(0.5f, 5.0f)? smiley

@N-Dekker I personally do not have a preference other than minimizing the burden on the contributors. I barely code in C++, and my knowledge on modern C++ practice or efficiency aspects is limited. I would suggest appropriately documenting these in the class headers, or even better, the ITK SW Guide, so that we can always be aware of the best/most readable/efficient practices, taking possibly into account that compilers need to be happy, and stick to them when contributing, and let the other ways be tested in the test files.

So for now, I am leaving as it is unless Sean says the current solution does not satisfy the compiler.

jhlegarreta:
The `CircleCI` build is expected to fail due to the know issues in the `Review` module. Cross-referencing https://github.com/InsightSoftwareConsortium/ITK/pull/2567#issuecomment-925830778.

PR #2759, opened 2021-09-27
COMP: making it compile on VS2017

dzenanz:
The regression was introduced by 4f309803b9b3c0268679d658e22d2f751f9c7b6c.

WIP. Reduces number of compile failures on Visual Studio 2017 on My computer. I plan to get back to this later this week.

### DIFF
--- a/Modules/Core/Common/include/itkCellInterface.h
+++ b/Modules/Core/Common/include/itkCellInterface.h
@@ -54,27 +54,27 @@
 // classes deriving form CellInterface (excluded).
 #define itkCellInheritedTypedefs(superclassArg)                                                                        \
   using Superclass = superclassArg;                                                                                    \
-  using typename Superclass::PixelType;                                                                                \
-  using typename Superclass::CellType;                                                                                 \
-  using typename Superclass::CellAutoPointer;                                                                          \
-  using typename Superclass::CellConstAutoPointer;                                                                     \
-  using typename Superclass::CellRawPointer;                                                                           \
-  using typename Superclass::CellConstRawPointer;                                                                      \
-  using typename Superclass::CellTraits;                                                                               \
-  using typename Superclass::CoordRepType;                                                                             \
-  using typename Superclass::InterpolationWeightType;                                                                  \
-  using typename Superclass::PointIdentifier;                                                                          \
-  using typename Superclass::PointIdIterator;                                                                          \
-  using typename Superclass::PointIdConstIterator;                                                                     \
-  using typename Superclass::CellIdentifier;                                                                           \
-  using typename Superclass::CellFeatureIdentifier;                                                                    \
+  using PixelType = typename Superclass::PixelType;                                                                    \
+  using CellType = typename Superclass::CellType;                                                                      \
+  using CellAutoPointer = typename Superclass::CellAutoPointer;                                                        \
+  using CellConstAutoPointer = typename Superclass::CellConstAutoPointer;                                              \
+  using CellRawPointer = typename Superclass::CellRawPointer;                                                          \
+  using CellConstRawPointer = typename Superclass::CellConstRawPointer;                                                \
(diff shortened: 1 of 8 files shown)


N-Dekker:
@dzenanz FYI, I did an similar partial revert for VS2107 on elastix! https://github.com/SuperElastix/elastix/commit/fd523e5526dc8381772cefc30c25e6f9e218fa01 

Specifically to work around VS2107 compiler bug "Compile error when using "using declaration" referencing a base class type that refers to itself", EssentiaX - Reported March 12, 2019 [Fixed in version 16.2], https://developercommunity.visualstudio.com/t/486683

The compile error only seemed to appear when the type that the class was "using"  was actually a base class.

dzenanz:
I saw https://github.com/SuperElastix/elastix/commit/fd523e5526dc8381772cefc30c25e6f9e218fa01. I tried that first, but then realized that `using X = Superclass::X;` works in VS2017 too.

Modules/Core/Common/include/itkCellInterface.h
```diff
@@ -54,27 +54,27 @@
...
+  using CellConstRawPointer = typename Superclass::CellConstRawPointer;                                                \
+  using CellTraits = typename Superclass::CellTraits;                                                                  \
+  using CoordRepType = typename Superclass::CoordRepType;                                                              \
+  using InterpolationWeightType = typename Superclass::InterpolationWeightType;                                        \
+  using PointIdentifier = typename Superclass::PointIdentifier;                                                        \
+  using PointIdIterator = typename Superclass::PointIdIterator;                                                        \
+  using PointIdConstIterator = typename Superclass::PointIdConstIterator;                                              \
+  using CellIdentifier = typename Superclass::CellIdentifier;                                                          \
+  using CellFeatureIdentifier = typename Superclass::CellFeatureIdentifier;                                            \
```
N-Dekker, line 70: Are you sure all of these changes are necessary, for VS2017? I had the impression that the VS2017 error "C2886: symbol cannot be used in a member using-declaration" only occurred when the "using type" itself a base class of this class. (Compiler bug https://developercommunity.visualstudio.com/t/486683)

N-Dekker:
> I saw [SuperElastix/elastix@fd523e5](https://github.com/SuperElastix/elastix/commit/fd523e5526dc8381772cefc30c25e6f9e218fa01). I tried that first, but then realized that `using X = Superclass::X;` works in VS2017 too.

Indeed, VS2017 supports both C++11 `using` type aliases _and_  C++98 using-declarations. Only limited by compiler bugs  😸

Modules/Core/Common/include/itkCellInterface.h
dzenanz: You are more than welcome to take over this PR, and explore the minimum amount of changes which get it compile on VS2017.

N-Dekker: I'm having a look now, and it appears that there's another category of erroneous (!) VS2017 compile errors that we're encountering now after pull request #2567, saying:

> error C2244: unable to match function definition to an existing declaration

This VS2017 compile error may be reproduced by:

```
template <typename T>
class Base
{
public:
  using NestedType = T;
};

template <typename T>
class Derived : public Base<T>
{
  using Superclass = Base<T>;
  using typename Superclass::NestedType;
  NestedType MemberFunction();
};

template <typename T>
typename Derived<T>::NestedType
Derived<T>::MemberFunction()
{  // <==  error C2244: unable to match function definition to an existing declaration

  return NestedType{};
}
```

And this particular error may be solved by a _trailing return type_ instead:

```
template <typename T>
auto
Derived<T>::MemberFunction() -> NestedType
{  // OK!

  return NestedType{};
}
```

And in this particular case (having a nested dependent type as return type), I think it would actually be quite nice to use a trailing return type.

To be continued...

dzenanz:
Closing in favor of #2780.

PR #2659, opened 2021-07-21
ENH: Bump `MetaIO`

jhlegarreta:
Bump `MetaIO`.

Fixes #2602.

## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [X] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)

PR #2761, opened 2021-09-29
 MetaIO: Update to 2021-09-28 (5e87f062) 

dzenanz:
@codeling The script runs on Linux without major deletions but still does not seem to work properly.


bradking:
FYI, #2659 was not done with the proper history shape.  That's what caused the update here with the old script to fail.  I manually built a fixed history shape to recover and switch to the new update script.

dzenanz (commented):
I can't approve my own PR, but it looks good.

dzenanz:
And thank you, Brad!

jhlegarreta (approved):

PR #2765, opened 2021-10-01
COMP: Update CircleCI image to cimg/python:3.9

thewtex:
cimg are the new CircleCI images based on Ubuntu.

This may address:

> Error message was: SSL certificate problem: certificate has expired

on CDash submission.

Due to:

  https://scotthelme.co.uk/lets-encrypt-old-root-expiration/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/python:3.7-stretch
+      - image: cimg/python:3.9
     working_directory: ~/
     resource_class: medium
     branches:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -46,7 +46,8 @@ jobs:
       - run:
           name: Dependencies
           command: |
-            sudo apt-get install -y  rsync ninja-build ccache
+            sudo apt-get update
+            sudo apt-get install -y  rsync ninja-build ccache python-pip
```
N-Dekker, line 50: Honestly I don't know much about CircleCI, but it says:
> Package python-pip is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  python3-pip

At https://app.circleci.com/pipelines/github/InsightSoftwareConsortium/ITK/3465/workflows/74239fd7-7453-43de-af8a-efdb2ed086e4/jobs/10221/parallel-runs/0/steps/0-106

thewtex: Let's try python3-pip

PR #2766, opened 2021-10-01
ENH: Add 1D FFT classes from `ITKUltrasound` external module

tbirdso:
Moves 1D FFT class implementations from [`ITKUltrasound`](https://github.com/kitwaremedical/itkultrasound) into `ITK/Modules/Filtering/FFT`. These classes differ from existing base ITK FFT classes in that they perform forward/inverse FFT in a single direction in a given ND image, while other ITK FFT classes perform ND FFT in an ND image.

This change allows FFT classes to be extended with GPU acceleration in [`ITKVkFFTBackend`](https://github.com/InsightSoftwareConsortium/ITKVkFFTBackend) which will then act as a dependency for [`ITKUltrasound`](https://github.com/kitwaremedical/itkultrasound). See [Issue 14 in `ITKVkFFTBackend`](https://github.com/InsightSoftwareConsortium/ITKVkFFTBackend/issues/14).

## PR Checklist


Modules/Filtering/FFT/include/itkComplexToComplex1DFFTImageFilter.hxx
```diff
@@ -0,0 +1,170 @@
...
+ComplexToComplex1DFFTImageFilter<TInputImage, TOutputImage>::New()
+{
+  Pointer smartPtr = ObjectFactory<Self>::Create();
+
+#ifdef ITKUltrasound_USE_clFFT
+  if (smartPtr.IsNull())
+  {
+    smartPtr =
+      dynamic_cast<Self *>(OpenCLComplexToComplex1DFFTImageFilter<TInputImage, TOutputImage>::New().GetPointer());
+  }
```
Leengit, line 50: Thanks for doing this.

Please see the discussion in Pull Request #2744, which is about a change that is being deferred to a later version of ITK but it highlights an issue that is applicable here: in current ITK code generally, a non-null `MyClass::Pointer` created from `ObjectFactory<MyClass>::Create()` will require a call to `UnRegister()` before being returned, but such a smart pointer created from `MyClass::New().GetPointer()` will require that `UnRegister()` not be called.

tbirdso: Thank you @Leengit ! I will give this issue attention before marking the PR as ready for review.

tbirdso: @Leengit is it sufficient to call `UnRegister` immediately after `ObjectFactory<Self>::Create()` returns? i.e. [`itkComplexToComplex1DFFTImageFilter.hxx`](https://github.com/InsightSoftwareConsortium/ITK/pull/2766/files#diff-3d993d90d493d78a45746642e1032067aff037768f0a833c855c25d960699c8bR39-R45):

```  
Pointer smartPtr = ObjectFactory<Self>::Create();

if (smartPtr.IsNotNull())
{
  // Decrement ITK SmartPointer produced from object factory
  smartPtr->UnRegister();
}
```

tbirdso:
@thewtex We've discussed moving 1D FFT filter creation to use an object factory for creation based on OpenCL / FFTW / VNL backends; my understanding is that this will require additional development for setting up factories. 

Is it reasonable to stick with the existing dispatch methods that have been thoroughly tested in `ITKUltrasound` for the scope of this PR, and then add factory creations for both 1D and ND FFT filters in a subsequent PR? Or, is there a simpler way to do this without significant expansion to this PR?

Modules/Filtering/FFT/include/itkComplexToComplex1DFFTImageFilter.hxx
dzenanz: #2744 is a pending PR, which might not be merged soon. Write the code so it works now. That PR will need some update, and expansion now that these classes are being ported into ITK proper. On the positive side, there will be less changes in remote modules.

dzenanz (commented):
Is this a direct copy from ITKUltrasound? In case it isn't, preferably split this commit into two, the first one just copying the code, and the second doing the changes to make it fit here. Either way, please indicate the exact commit at which the files were taken.

SimonRit:
Thanks for moving this code to ITK's main repo. For your information, there is a class in RTK doing 1D filtering along lines, [rtk::FFTProjectionsConvolutionImageFilter](https://github.com/SimonRit/RTK/blob/master/include/rtkFFTProjectionsConvolutionImageFilter.h). As far as I understand after a quick review, RTK's implementation is less advanced because one cannot chose the filtering direction. But it has other features: zero padding and technicalities related to tomography such as data padding before filtering or windowing. Adapting RTK's code to the content of your PR would be quite some work with no obvious benefit other than maintainability (which is important of course).

But my main comment is that we have achieved something similar in a rather different fashion. We do a 1D filtering but with nD FFTs: the 1D kernel is treated as a nD image with all dimensions but the first set to 1; its FFT is multiplied by each line of the nD FFT of the input image:
https://github.com/SimonRit/RTK/blob/master/include/rtkFFTProjectionsConvolutionImageFilter.hxx#L163-L177
It would be good to test if this is less efficient than what is proposed here (I will try to find some time to do it once your PR is ready to be merged). But my feeling is that we have achieved the same thing with much less code... For example, we let ITK's code manage the different FFT options (Vnl, FFTW, etc.).

In any case, I find your PR useful as a starting point and I will consider this option as an alternative to RTK's implementation, thanks for doing it.

Modules/Filtering/FFT/include/itkComplexToComplex1DFFTImageFilter.hxx
tbirdso: Thanks @dzenanz, I meant to ask if the code as currently written with a call to `UnRegister` is compatible with the current state of the ITK code base not taking #2744 into account.

The plan is to eventually move to using the object factory to generate FFT classes based on library availability (FFTW, OpenCL) rather than the dispatch methods here, which are particularly affected by #2744.

tbirdso:
Interesting failures in [`PythonGetNameOfClass`](https://open.cdash.org/test/505443507):
> <itkTemplate itk::Forward1DFFTImageFilter\>: wrong class name: Forward1DFFTImageFilter
<itkTemplate itk::Inverse1DFFTImageFilter\>: wrong class name: Inverse1DFFTImageFilter

Will investigate.

tbirdso:
@dzenanz 
> Is this a direct copy from ITKUltrasound? In case it isn't, preferably split this commit into two, the first one just copying the code, and the second doing the changes to make it fit here. Either way, please indicate the exact commit at which the files were taken.

There are a few minor changes in regards to how we are handling construction with dispatch methods, removing OpenCL references (these will be moved into ITKVkFFTBackend), and moving test data to `data.kitware.com` to meet ITK's requirements on maximum test file sizes. I will work on splitting changes from the original files as you said.

tbirdso:
@SimonRit
> Thanks for moving this code to ITK's main repo. For your information, there is a class in RTK doing 1D filtering along lines, [rtk::FFTProjectionsConvolutionImageFilter](https://github.com/SimonRit/RTK/blob/master/include/rtkFFTProjectionsConvolutionImageFilter.h). As far as I understand after a quick review, RTK's implementation is less advanced because one cannot chose the filtering direction. But it has other features: zero padding and technicalities related to tomography such as data padding before filtering or windowing. Adapting RTK's code to the content of your PR would be quite some work with no obvious benefit other than maintainability (which is important of course).

Thanks for weighing in! As you said, each has its advantages. In the case of ITKUltrasound the 1D image filters have often been used for spectral analysis other than signal filtering in the frequency domain along transducer beam lines. 

[`itkFFTPadImageFilter`](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/FFT/include/itkFFTPadImageFilter.h) is useful for data padding prior to performing FFT. As this operates in ND it could be useful to add a 1D padding filter as a corollary.

> But my main comment is that we have achieved something similar in a rather different fashion. We do a 1D filtering but with nD FFTs: the 1D kernel is treated as a nD image with all dimensions but the first set to 1; its FFT is multiplied by each line of the nD FFT of the input image: https://github.com/SimonRit/RTK/blob/master/include/rtkFFTProjectionsConvolutionImageFilter.hxx#L163-L177 It would be good to test if this is less efficient than what is proposed here (I will try to find some time to do it once your PR is ready to be merged). But my feeling is that we have achieved the same thing with much less code... 

I'm not clear on the implications of performing filtering in one direction in an nD FFT image, but in this case the outcome of 1D forward/inverse FFT filters can be expanded beyond filtering.

> For example, we let ITK's code manage the different FFT options (Vnl, FFTW, etc.).

Yes, our goal in moving this code to ITK is to manage backend options (Vnl, FFTW, OpenCL, etc).

> In any case, I find your PR useful as a starting point and I will consider this option as an alternative to RTK's implementation, thanks for doing it.

Thank you! These changes do not replace any existing FFT functionality in ITK.

Perhaps @thewtex or @dzenanz would like to weigh in on the RTK filtering implementation as well?

dzenanz:
@tbirdso I am afraid I did not quite understand how RTK's implementation works, so I can't meaningfully comment.

SimonRit:
> @tbirdso I am afraid I did not quite understand how RTK's implementation works, so I can't meaningfully comment.

Sorry if I was not clear, I can try to explain. Taking the 1D FFT of each line of a nD image and multiplying it by the 1D FFT of a 1D kernel is equivalent to to taking the nD FFT of the image and multiplying each line by the 1D FFT of the 1D kernel. It can be written mathematically with delta functions, see e.g. equation (7) of [this paper](https://hal.archives-ouvertes.fr/hal-02486620/document). And to take the 1D FFT of the 1D kernel, I use an nD FFT with the size set to 1 but in one dimension, see e.g. [here](https://github.com/SimonRit/RTK/blob/master/include/rtkFFTRampImageFilter.hxx#L59-L89) for the ramp kernel (used a lot in tomography).

thewtex:
> Is it reasonable to stick with the existing dispatch methods that have been thoroughly tested in ITKUltrasound for the scope of this PR, and then add factory creations for both 1D and ND FFT filters in a subsequent PR? 

@tbirdso yes, sounds good.

thewtex:
> Interesting failures in PythonGetNameOfClass:

@tbirdso we likely need to add these classes to the exclude list, like the other FFT classes, here:

https://github.com/InsightSoftwareConsortium/ITK/blob/24c3ad91db624592318c886c8d81614e0662ea72/Wrapping/Generators/Python/Tests/getNameOfClass.py#L42

thewtex:
@SimonRit yes, it would be very interesting to know if these classes can be useful for your 1D convolution! I think we will only know after some benchmarking whether the current approach in `rtk::FFTProjectionsConvolutionImageFilter.` is faster, or whether we could use a compose pipeline of these 1D FFT filters, the FFT pad filter, a 1D windowing filter (I think we have one of those too that we could port(?)) . Either way, it would be cool to add a 1D FFT convolution filter to this module for use across many contexts.

tbirdso:
@dzenanz

> Is this a direct copy from ITKUltrasound? In case it isn't, preferably split this commit into two, the first one just copying the code, and the second doing the changes to make it fit here. Either way, please indicate the exact commit at which the files were taken.

I've split this into two commits with the first copying 1DFFT files and the second applying minor fixups to conform to ITK. There is one exception where including the full testing data files in the source tree violated the maximum test file size for ITK and caused a `ghostflow-check-master` failure. As a result SHA256 hash files for test data have been included in the base commit instead.

dzenanz (approved):

thewtex:
Thanks @tbirdso !

A rebase on `master` will help with the CircleCI build.

thewtex:
CC @LucasGandel

thewtex (approved):
:rocket:

PR #2770, opened 2021-10-03
COMP: Initialize CXX standard in ITK_USE_FILE

thewtex:
Code that builds against ITK may not be built with a sufficient or
compatible C++ standard.

Use ITKInitializeCXXStandard.cmake in client code, which calls

  include(${ITK_USE_FILE})

according to our existing documentation, which will set
CMAKE_CXX_STANDARD and CMAKE_CXX_STANDARD_REQUIRED if not already set.

To address build errors as discussed in:

  https://discourse.itk.org/t/hello-world-from-itk-examples-installation-not-compiling-for-itk-5-3/4441/8


jhlegarreta (approved):
Have not tried. Thanks Matt.

Tests are passing for `CircleCI`, it's just the build report that could not been sent.

thewtex:
CircleCI build addressed in #2765

PR #2778, opened 2021-10-04
COMP: fixed various Wunused-but-set-variable warnings from Clang 13

seanm:
Three solutions:
- removed the dead code.
- use the variable by logging it
- cast variable to void

## PR Checklist


seanm:
@thewtex @dzenanz @hjmjohnson I'm not at all sure these are the best fixes here. But figured I'd do something and we can discuss here.

Only someone that knows these classes can really know if the code is dead from an old refactoring, or if those variables should in fact be used.

Modules/Core/Common/test/itkPeriodicBoundaryConditionTest.cxx
```diff
@@ -308,9 +308,9 @@ itkPeriodicBoundaryConditionTest(int, char *[])
     for (auto & j : back)
     {
       const float d = testIter.GetPixel(j);
-      sum += d * d;
```
dzenanz, line 311: I guess this was here to make sure compiler does not optimize away the call on the previous line.

Modules/Core/Common/test/itkPeriodicBoundaryConditionTest.cxx
```diff
@@ -308,9 +308,9 @@ itkPeriodicBoundaryConditionTest(int, char *[])
...
     }
     ++testIter;
+    (void)sum;
```
dzenanz, line 313: I guess logging to console would work, preferably just once here instead of multiple times in the loop.

Modules/IO/MRC/src/itkMRCHeaderObject.cxx
```diff
@@ -110,52 +110,6 @@ MRCHeaderObject::SetHeader(const Header * buffer)
...
-    extendedHeaderBytes += 2;
-  }
-  if (this->m_Header.nreal & 512)
-  {
-    extendedHeaderBytes += 4;
-  }
-  if (this->m_Header.nreal & 1024)
-  {
-    extendedHeaderBytes += 2;
-  }
```
dzenanz, line 157: How many tests invoke this code? If just one, maybe figure out what `extendedHeaderBytes` should be and compare against it here? Just removing code does not feel right in this instance.

dzenanz (commented):
I made some suggestions where I felt different from your proposal.

Modules/Core/Common/test/itkPeriodicBoundaryConditionTest.cxx
seanm: OK, I now use this dummy sum and log it.

Modules/Core/Common/test/itkPeriodicBoundaryConditionTest.cxx
seanm: OK, I now use this dummy sum and log it.

Modules/IO/MRC/src/itkMRCHeaderObject.cxx
seanm: I threw in a null dereference and ran all tests.  The following hit the removed code:

```
	929 - itkMRCImageIOTest (ILLEGAL)
	930 - itkMRCImageIOTest2a (ILLEGAL)
	931 - itkMRCImageIOTest2b (ILLEGAL)
	932 - itkMRCImageIOTest2c (ILLEGAL)
```

Looking at git blame, the code I removed has been dead code for a very long time.

I agree just removing it is not ideal without understanding this code.  According to `git log` every commit for about a decade has been just stylistic-type stuff, so I'm not sure anyone actually knows this code. :)

jhlegarreta: Looks like these classes were originally contributed by @blowekamp and @dave3d , so maybe they still have some data that would require these code (?) or know the purpose of the relevant part.

blowekamp: @seanm I have encountered the below warning..... Look like I missed an e-mail 🤣

PR #2780, opened 2021-10-05
Workaround VS2017 compile errors using-declarations

N-Dekker:
Worked around three categories of VS2017 compile errors, caused by compiler bugs:

> error C2653: '...': is not a class or namespace name
> error C2886: '...': symbol cannot be used in a member using-declaration
> error C2244: unable to match function definition to an existing declaration

The errors were reported by @dzenanz at #2759 "COMP: making it compile on VS2017" , after the merge of pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2567 https://github.com/InsightSoftwareConsortium/ITK/commit/4f309803b9b3c0268679d658e22d2f751f9c7b6c "STYLE: Avoid repeating parent aliases"

### DIFF
--- a/Modules/Bridge/NumPy/include/itkPyBuffer.hxx
+++ b/Modules/Bridge/NumPy/include/itkPyBuffer.hxx
@@ -70,8 +70,9 @@ PyBuffer<TImage>::_GetArrayViewFromImage(ImageType * image)
 }
 
 template <class TImage>
-const typename PyBuffer<TImage>::OutputImagePointer
+auto
 PyBuffer<TImage>::_GetImageViewFromArray(PyObject * arr, PyObject * shape, PyObject * numOfComponent)
+  -> const OutputImagePointer
 {
   PyObject * shapeseq = NULL;
   PyObject * item = NULL;
--- a/Modules/Bridge/NumPy/include/itkPyVectorContainer.hxx
+++ b/Modules/Bridge/NumPy/include/itkPyVectorContainer.hxx
@@ -57,8 +57,9 @@ PyVectorContainer<TElementIdentifier, TElement>::_array_view_from_vector_contain
 }
 
 template <typename TElementIdentifier, typename TElement>
-const typename PyVectorContainer<TElementIdentifier, TElement>::VectorContainerType::Pointer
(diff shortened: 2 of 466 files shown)


dzenanz (approved):
Awesome Niels!

N-Dekker:
> Awesome Niels!

@dzenanz Thanks 😄 Honestly it was more work that I had expected, especially because I wanted to preserve as much as possible from pull request #2567. I did not realize before that trailing return types would help to work around those errors.

dzenanz:
> it was more work that I had expected

I realized that too, and since I was tight with time I made that draft PR with progress (in case someone else is bugged by it too) 😄

N-Dekker:
> ...since I was tight with time I made that draft PR with progress (in case someone else is bugged by it too)

@dzenanz  It bugged you? Does that mean you actually still use VS2017? Anyway, can you possibly confirm that this PR does indeed make VS2017 happy?

dzenanz:
I do have VS2017 installed, and a few years ago when I set up nightly build on my machine (Ryzenator) VS2017 was the latest and greatest. So it is still built nightly. I just tried building your branch, and it built successfully! I have switched to VS2019 for daily use long time ago.

N-Dekker:
Maybe it should be a style guideline to use a trailing return type for any class template member function definition that is defined outside its class template definition (so usually in a *.hxx file), whenever the return type is specified in terms of a dependent member type. What do you think?

dzenanz:
I agree. That would make specifying those types easier.

N-Dekker:
> I agree. That would make specifying those types easier.

An easier guideline could be something like: Whenever choosing between a trailing return type (as introduced with C++11) and the old (C++98) form of having the return type before the function name, prefer the form that is the shortest, and the simplest.

    // Preferred:
    int f()

    // Not preferred:
    auto f() -> int

    // Preferred:
    auto Template::MemberFun() -> MemberType

    // Not preferred:
    typename Template::MemberType Template::MemberFun()

OK?

Anyway, most importantly, the coding style should be compatible with VS2017  😸

dzenanz (approved):

thewtex:
Thanks, @N-Dekker !

PR #2627, opened 2021-07-02
BUG: Similarity3DTransform::SetScale should recompute m_Offset

N-Dekker:
Closes #2629 "Similarity3DTransform::SetScale should recompute m_Offset", which said:

> When `Similarity3DTransform::SetScale` is called, it does not compute
> `m_Offset`, which appears to cause an incorrect result from
> `similarity3DTransform.TransformPoint` when `TransformPoint` is called
> directly after `SetScale`.

Included a unit test (GoogleTest) to ensure that the issue is fixed.

PR #2783, opened 2021-10-06
BUG: VersorTransform did not update offset in SetParameters

zivy:


dzenanz (approved):
This is so reminiscent of #2627!

dzenanz:
Yes! 5.3.0 will be cut from master, but I think we are going to have another RC first. Bug fixes can always be included.

PR #2785, opened 2021-10-07
WIP: Remove `virtual` from Get/Set macro definitions

N-Dekker:
Declared Get/Set member functions non-virtual _by default_ when `ITK_FUTURE_LEGACY_REMOVE` is set.

Added `itkVirtual` and `itkNonVirtual` Get/Set macro's, to explicitly choose between virtual and non-virtual. 

Observed a significant reduction in the size of generated "ITK*.lib" files when declaring Get/Set member functions non-virtual by default. For example, "ITKTransformFactory-5.3.lib" went from 18.3 MB down to 17.5 MB, and "ITKCommon-5.3.lib" went from 8.1 MB down to 7.6 MB, using VS2019 (Release).


dzenanz (commented):
I agree with Brad that this is somewhat of a big change. If doing it now, make it `FUTURE_LEGACY`, which would graduate into `LEGACY` at time of 6.0. And, of course, wait until 5.3 final is out before merging it.

For virtual dispatches to work correctly, the setters must be invoked consistently (instead of setting the ivars directly). Because there are not that many instances where virtual setters are overridden, the consistency in calling the setters is absent. So there is a case for making the setters non-virtual by default. Also, this is an open source library which is open to contributions. If someone needs some class' functionality to be virtualized, they can propose a PR which makes some or all of the setters virtual.

dzenanz (commented):
LGTM

_commit 8f26661 creates blob 12398f35ac335d57dafd510e044b91bd28b4d29f at Modules/Core/Common/include/itkMacro.h with size 101090 bytes (98.72 KiB) which is greater than the maximum size 100000 bytes (97.66 KiB). If the file is intended to be committed, set the hooks-max-size attribute on its path._

Wow, `itkMacro.h` is 100KB! This might be the default limit, so we will need a special entry in `.gitattributes` for it.

hjmjohnson:
@N-Dekker   Now that we are at ITK 6.0, can we revisit this?  It seems like now is the time to put effort into evaluating this effort.

N-Dekker:
>  Now that we are at ITK 6.0, can we revisit this?

@hjmjohnson I have to think about it. Over the last few years (after submitting this PR) I did encounter some cases where Get and/or Set member functions were overridden. I don't really _like_ that, I think it's error prone. (What if someone overrides the Get, but forgets to override the corresponding Set?) And when a class has a specific data member allocated for a Get and Set, it seems wasteful to override it to bypass the allocated data. But having said so, I know now that this feature of overridable Get and Set is actually being used in practice. So I hesitate.  🤔

hjmjohnson:
im just revisiting old PRs

________________________________

dzenanz:
Maybe mark this PR as draft until you make up your mind?

hjmjohnson:
Per the request to consolidate the scattered `virtual` / `override` / `final` design conversation around ITK Get/Set macros, I've opened **#6232 (Master tracking — virtual / override / final semantics on Get/Set macros for ITK 6.x)** as a single venue.

It pulls in the 2018 [Discourse #814](https://discourse.itk.org/t/removal-of-virtual-keywords-from-constneighborhooditerator/814) precedent (the same `final`-as-stress-test technique and the @phcerdan / @N-Dekker 18–24% HoughTransform performance gain), the 2026-05-07 final-keyword audit (17 base methods overridden across 19 files in upstream ITK), and a draft additive `itkVirtualSetMacro` family on `hjmjohnson:macro-optionally-virtual @ ffc70ed0b9`. Surfaces 7 binary/short-list questions — Q1 (default-emit `virtual`?) is the load-bearing one this PR is blocked on.

Suggest holding further design comments on this thread until #6232 has a maintainer-quorum answer on Q1; otherwise the conversation re-fragments. This thread can stay open as the implementation venue once direction is settled.

hjmjohnson:
@N-Dekker Can the PR be closed as substantively resolved via 
- [X] https://github.com/InsightSoftwareConsortium/ITK/pull/6253
- [X] https://github.com/InsightSoftwareConsortium/ITK/pull/6415

N-Dekker:
@hjmjohnson 
> Can the PR be closed as substantively resolved...

Hereby!

PR #2788, opened 2021-10-08
BUG: fix a crash in QuadEdgeMeshFilter tests

dzenanz:
itkSquaredEdgeLengthDecimationQuadEdgeMeshFilterTest
itkQuadricDecimationQuadEdgeMeshFilterTest

were crashing in Debug mode on Visual Studio 2017.

The element is used for its type, so the actual access is optimized away in Release mode. In these tests, the cell/pixel data did not have an element with key `0`, so trying to access it caused a crash.


N-Dekker:
> tkSquaredEdgeLengthDecimationQuadEdgeMeshFilterTest
> itkQuadricDecimationQuadEdgeMeshFilterTest
> were crashing in Debug mode on Visual Studio 2017.

Nice catch! I guess this bug might cause a crash on any platform or compiler version, right?

jhlegarreta (approved):
Have not tried, but :100: for doing this Dženan.

Just to make sure, are commits 0b87d5e and 36d01ea required for the fix to work ? If so, would it be better to squash them into the same commit as the fix ?

PR #2796, opened 2021-10-09
WIP: ENH: Increase `itk::DCMTKFileReader` class coverage

jhlegarreta:
Add a new test to the `IODCMTK` module to increase the coverage of the
`itk::DCMTKFileReader::GetElement*` methods.

## PR Checklist


jhlegarreta:
A few notes:
- I know the module gets less use than `GDCM`, but if it is in the toolkit, it should get some more testing.
- I could not compile the `DCTMK` test module:
```
Build started...
1>------ Build started: Project: ITKIODCMTKTestDriver, Configuration: Debug x64 ------
1>itkDCMTKGetDicomTagsTest.cxx
1>C:\SDKs\ITK\itk-head\ITK\Modules\IO\DCMTK\include\itkDCMTKFileReader.h(27,10): fatal error C1083:
Cannot open include file: 'dcmtk/dcmdata/dcdict.h': No such file or directory
1>Done building project "ITKIODCMTKTestDriver.vcxproj" -- FAILED.
========== Build: 0 succeeded, 1 failed, 0 up-to-date, 0 skipped ==========
```
Using the built-in `DCMTK` (i.e. not a system `DCMTK`) I cannot see where the error stems from. I'll try to submit an experimental build to see if that helps other folks to try to reproduce the issue.

Thus, I could not compile and run the test to see its errors.

Not sure if the CIs have the module flag set to `On` so they may come up green.

- Not sure if the test name is the most appropriate one. Suggestions are welcome.
- I did not found non-empty tags in the used DICOM data to exercise other methods like, `GetElementCSorOB`, etc. Adding another test that uses another set of DICOM data could allow querying such elements/exercising those methods. Best would be to make the test quantitative for that.
- @thewtex In order to make the test quantitative, I'm wondering whether we have CMake-enabled ways to compare a potential txt output file with the read tags to a baseline with the expected tags. If not, do you have some suggestion?

dzenanz (approved):
Making the test quantitative would be good, but that can be a follow-up commit. This is good as-is.

jhlegarreta:
> Making the test quantitative would be good, but that can be a follow-up commit. This is good as-is.

Sounds reasonable. Note that I have not been able to compile the module, and hence nor I have tested it, and I would not merge this until it gets tested.

The [CI builds did not set the module to `ON`](https://dev.azure.com/itkrobotlinux/ITK.Linux/_build/results?buildId=6727&view=logs&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=f58be928-45a1-58e7-c1ad-04869565b3f4&l=61099), and thus, even if they have come green, it might well be the case that the test does not compile/pass:
```
2021-10-10T01:29:38.0772539Z 595: //Request building ITKIODCMTK (non-remote)
2021-10-10T01:29:38.0773102Z 595: Module_ITKIODCMTK:BOOL=OFF
```

jhlegarreta:
This is still relevant stale bot. I am willing to take up this as soon as #2817 gets resolved.

Modules/IO/DCMTK/test/itkDCMTKGetDicomTagsTest.cxx
```diff
@@ -0,0 +1,157 @@
...
+  if (fileReader.GetElementCS(group, element, strTarget, throwException) == EXIT_SUCCESS)
+  {
+    std::cout << strTarget << std::endl;
+  }
+
+  // Patient Weight
+  // Expect 68.039
+  group = 0x0010;
+  element = 0x1030;
+  if (fileReader.GetElementDS(group, element, 1, &strTarget, throwException) == EXIT_SUCCESS)
```
issakomi, line 92: &strTarget withput &

Modules/IO/DCMTK/test/itkDCMTKGetDicomTagsTest.cxx
```diff
@@ -0,0 +1,157 @@
...
+  {
+    std::cout << strTarget << std::endl;
+  }
+
+  // Echo Number
+  group = 0x0018;
+  element = 0x0086;
+  // Expect (12576)"1"
+  itk::int32_t int32Target;
+  if (fileReader.GetElementIS(group, element, int32Target, throwException) == EXIT_SUCCESS)
```
issakomi, line 102: VR IS is "integer string" not _int32_ (it would be VR SL)

Modules/IO/DCMTK/test/itkDCMTKGetDicomTagsTest.cxx
```diff
@@ -0,0 +1,157 @@
...
+  {
+    std::cout << int32Target << std::endl;
+  }
+
+  // Acquisition Duration
+  group = 0x0019;
+  element = 0x105a;
+  // Expect 1304791694
```
issakomi, line 110: Strange, my viewer and also _dcmdump_ show
`(0019,105a) FL 414273984                                #   4, 1 AcquisitionDuration`
not 1304791694. Both are somehow strange for float, there may be an issue in the file.
Also applicable to private tags in the file, it is correct to access private tags (odd group) using private creator, not just by group and element. But to simply increase coverage by using the somehow it is probably not important.

Modules/IO/DCMTK/test/itkDCMTKGetDicomTagsTest.cxx
issakomi: Change requested (s. above comment).

issakomi (changes requested):

Modules/IO/DCMTK/test/itkDCMTKGetDicomTagsTest.cxx
jhlegarreta: :100: thanks for the review @issakomi. Done. Push-forced in 73c450e.

Modules/IO/DCMTK/test/itkDCMTKGetDicomTagsTest.cxx
jhlegarreta: Not sure where I got the value from :-S. I probably used an online viewer, which might not be fully trustworthy.

I am not able to build the module locally, so I was relying on an online tool rather than installing locally some library/tool. But I am willing to fix it if anybody can point me to a dependable tool that I could use to potentially get the accurate values for more tags.

jhlegarreta: > Also applicable to all private tags in the test, it is correct to access private tags (odd group) using private creator, not just by group and element. 

Can you elaborate on this? If there is a better way to access the tags, I may have a look to also add them (?).

issakomi: > Can you elaborate on this? If there is a better way to access the tags, I may have a look to also add them (?).

Sorry, not in the near future. The test file has explicit transfer syntax, it works, a VR is known. Imagine some implicit file, VR can be taken only from a private dictionary, otherwise only OB could be used as workaround.

issakomi: This particular tag is also known for DCMTK (is in private dictionary), so it is OK.

issakomi (approved):

Modules/IO/DCMTK/test/itkDCMTKGetDicomTagsTest.cxx
issakomi: > I am not able to build the module locally, so I was relying on an online tool rather than installing locally some library/tool.

I have no problem with Linux build and internal DCMTK. I remember with external and at least static libraries there were problems too, something with missing libraries (they were there, but not found).

P.S.
I have to admit I am not experienced with DCMTK, actually I am testing it for PACS stuff, it seems to be clean and well maintained library.

jhlegarreta:
> Making the test quantitative would be good, but that can be a follow-up commit. This is good as-is.

I may give a go at this. If the module is turned on by the CIs, they'll tell about the tag contents. I'd think of two ways to make this quantitative:
- Write the expected tag values to a text file, provide it as the input to the test, read each line and compare each value with the value read from the DICOM files.
- Write the values read from the DICOM files to a text file, provide a text file with the expected tag values as a baseline to CMake, and let it do the comparison job.

If the latter is possible, it involves less code, and allows for improvements (e.g. reading/comparing more tags) without minimal changes to the test code.

@dzenanz Wondering whether the CMake `compare` is able to compare two text files.

issakomi:
DCMTK has own tests, BTW, there are some macros too e.g. OFCHECK_EQUAL
https://github.com/DCMTK/dcmtk/tree/master/dcmdata/tests

jhlegarreta:
> DCMTK has own tests

:+1: Here we are trying to make sure that **ITK's** `DCMTK` `IO` object reads what it is supposed to read.

> BTW, there are some macros too e.g. OFCHECK_EQUAL
https://github.com/DCMTK/dcmtk/tree/master/dcmdata/tests

:+1: Thanks Mihail.

dzenanz:
CMake can compare files ([syntax](https://stackoverflow.com/questions/54353120/how-to-compare-files-in-cmake)), but only binary comparison. It should work in this case, and it is less work to set up than read the file and compare ourselves. The downside is that error message will be "file are different", rather than more meaningful messages we can craft in the custom code. If you have time, you can go the custom comparison route. Otherwise let CMake compare the outputs.

P.S. I don't know how easy or hard it is to turn `CMake -E compare_files` into a test.

dzenanz:
Slicer should be able to read DICOM tags.

jhlegarreta:
> P.S. I don't know how easy or hard it is to turn CMake -E compare_files into a test.

Yes, that one I had found yesterday, but did not find out whether our `compare` argument does also accept text files.

issakomi:
FYI, some parts are too bad for testing and coverage, e.g. this [function](https://github.com/InsightSoftwareConsortium/ITK/blob/dca4a36dde8dc4b45efff4ef7df411e7010d0163/Modules/IO/DCMTK/src/itkDCMTKFileReader.cxx#L1064) (unused, AFAIK).
It takes _Bits Allocated_ (0x0028, 0x0100) with wrong name SamplesPerPixel and assumes that if _Bits Allocated_  is 8 or 16 it means that the image is scalar, otherwise RGB or VECTOR, it is catastrophic. RGB image has of course _Bits Allocated_  8 or 16 too and _Samples per Pixel_ (0x0028,0x0002) **3**. And e.g. 32 bits dose (RT) image is not a vector. Also _Photometric Interpretation_ should be checked. There were some very very old Arc Nema images with bitsallocated 24 for RGB in the wild, it is exception and invalid for DICOM, extremely rare, never seen one.

```cpp
IOPixelEnum
DCMTKFileReader::GetImagePixelType() const
{
  unsigned short SamplesPerPixel;
  if (this->GetElementUS(0x0028, 0x0100, SamplesPerPixel, false) != EXIT_SUCCESS)
  {
    return IOPixelEnum::UNKNOWNPIXELTYPE;
  }
  IOPixelEnum pixelType;
  switch (SamplesPerPixel)
  {
    case 8:
    case 16:
      pixelType = IOPixelEnum::SCALAR;
      break;
    case 24:
      pixelType = IOPixelEnum::RGB;
      break;
    default:
      pixelType = IOPixelEnum::VECTOR;
  }
  return pixelType;
}
```

jhlegarreta:
@issakomi Thanks for the investigation. I'd be more than happy to add more test cases to test the relevant methods if the appropriate DICOM data can be shared.

issakomi:
> @issakomi Thanks for the investigation. I'd be more than happy to add more test cases to test the relevant methods if the appropriate DICOM data can be shared.

I think i have mentioned a bug, better say forgotten garbage in _DCMTKFileReader_ class.
There are many test files for GDCM tests. But i think they should better tested manually first to get some knowledge of the class, e.g. we don't know does PALETTE_COLOR works or not.

The Test is [here](https://drive.google.com/file/d/1pNFjRuSqzlEIopaJSSZLIfxNIjlRWHF6/view?usp=sharing), it is complete minimal standalone test. Not intended to be included in test suite.

```cpp
#include "itkImage.h"
#include "itkImageFileReader.h"
#include "itkDCMTKImageIO.h"
#include "itkDCMTKFileReader.h"
#include <iostream>

int main(int, char **)
{
  using TImage = itk::Image<float, 3>;
  using ReaderType = itk::ImageFileReader<TImage>;
  auto imageIO = itk::DCMTKImageIO::New();
  imageIO->SetFileName("../data/raw-RGB.dcm");
  try
  {
    imageIO->ReadImageInformation();
  }
  catch (const itk::ExceptionObject & ex)
  {
    std::cout << ex.GetDescription() << std::endl;
  }
  itk::IOPixelEnum pixel_type1 = imageIO->GetPixelType();
  std::cout << "DCMTKImageIO reported Pixel Type  "
    << imageIO->GetPixelTypeAsString(pixel_type1)
    << std::endl;

  // GetImagePixelType() is unused garbage
  itk::DCMTKFileReader dcmtkFileReader;
  dcmtkFileReader.SetFileName("../data/raw-RGB.dcm");
  dcmtkFileReader.LoadFile();
  itk::IOPixelEnum pixel_type2 = dcmtkFileReader.GetImagePixelType();
  std::cout << "DCMTKFileReader reported Pixel Type  "
    << imageIO->GetPixelTypeAsString(pixel_type2)
    << std::endl;
  return 0;
}
```

```
DCMTKImageIO reported Pixel Type  rgb
DCMTKFileReader reported Pixel Type  scalar
```

Sorry, but i am out and will not do any more work on DCMTK in the near future.

Modules/IO/DCMTK/test/itkDCMTKGetDicomTagsTest.cxx
```diff
@@ -0,0 +1,215 @@
...
+  element = 0x1061;
+  if (fileReader.GetElementUI(group, element, strTarget, throwException) == EXIT_SUCCESS)
+  {
+    outputFile << strTarget << std::endl;
+    std::string expectedVal = "1.2.840.113619.2.133.1762890640.1886.1055165015.961";
+    ITK_TRY_EXPECT_EQUAL(strTarget, expectedVal);
+  }
+
+  // Image Pixel Type
+  IOPixelEnum pixelType = fileReader.GetImagePixelType();
```
issakomi, line 205: GetImagePixelType() is obviously too bad and should be fixed or even better removed from the class.

issakomi (changes requested):

Modules/IO/DCMTK/test/itkDCMTKGetDicomTagsTest.cxx
```diff
@@ -0,0 +1,215 @@
...
+    ITK_TRY_EXPECT_EQUAL(usTarget, expectedVal);
+  }
+
+  // Photometric Interpretation
+  group = 0x0028;
+  element = 0x0004;
+  if (fileReader.GetElementUS(group, element, strTarget, throwException) == EXIT_SUCCESS)
+  {
+    outputFile << strTarget << std::endl;
+    unsigned short expectedVal = "????";
```
issakomi, line 159: PhotometricInterpretation is VR CS (Code String), not US.

jhlegarreta: Sorry, my bad. Fixed in 172680e. Thanks for the review.

Modules/IO/DCMTK/test/itkDCMTKGetDicomTagsTest.cxx
jhlegarreta: The ideal thing would be to fix it. If `itk::DCMTKImageIO` (maybe through `itk::ImageIOBase`) or the `itk::DCMTKImageIO::ReadImageInformation` are doing a better job in this, then part of the code might be valid for here (haven't looked). If we do not find a way to fix it, then it should be deprecated (and yet be tested so that at least its behavior is consistent) prior to being removed.

Modules/IO/DCMTK/test/itkDCMTKGetDicomTagsTest.cxx
issakomi: Thank you. The line
`unsigned short expectedVal = "????";`
will not compile.

Modules/IO/DCMTK/test/itkDCMTKGetDicomTagsTest.cxx
issakomi: itk::DCMTKImageIO::ReadImageInformation is in fact OK, not ideal too, i don't like that SamplesPerPixel=4 is assumed to be RGBA, in fact it could be e.g. CMYK. But is really not important, currently there are no DICOM files with SamplesPerPixel=4 (CMYK, ARGB), they are deprecated long long ago. So in 99.999% ReadImageInformation work fine. There are more to test (manually first), e.g. how DCMKT handles YBR images, PALETTE_COLOR. Maybe later.

issakomi: > and yet be tested so that at least its behavior is consistent) prior to being removed

test are always good, but not including known completely broken stuff into test suite, sorry. The test files can not match too, sorry.

Modules/IO/DCMTK/test/itkDCMTKGetDicomTagsTest.cxx
jhlegarreta: I know. Thanks for noticing. Following https://github.com/InsightSoftwareConsortium/ITK/pull/2796#discussion_r1014594203 all values should be checked, but it will take before I am able to do that. Ideally I'd also remove the expected values and the comparisons from the test, and rely on CMake and the text files for quantitative checking purposes.

Modules/IO/DCMTK/test/itkDCMTKGetDicomTagsTest.cxx
jhlegarreta: Thanks for all this insight. Will see when I get the time to push this, as I am still being blocked by #2817 to be able to properly do this.

jhlegarreta: Now blocked by #3820.

hjmjohnson:
Closing as superseded by **#6189**, which carries this work forward to a buildable, runnable state on current `upstream/main`.

The new PR addresses everything in this thread:

- **The 3.5-year-old blocker (#3820)** is fixed in commit 1 of #6189: a narrowly-scoped `target_include_directories()` on `ITKIODCMTKTestDriver` that pulls in the DCMTK ExternalProject's include dirs without changing the IODCMTK module's public DEPENDS surface. Tests can now `#include "itkDCMTKFileReader.h"` directly.
- **All of @issakomi's specific concerns** are addressed:
  - The `unsigned short = "????"` compile error → rewritten
  - The `1304791694` wrong-by-transcription value → corrected to actual `414273984.0` (verified against the on-disk fixture with pydicom before authoring)
  - PhotometricInterpretation now read as `CS` string `"MONOCHROME2"` instead of `unsigned short`
  - The buggy `GetImagePixelType()` test was dropped per @malaterre's earlier feedback that the method itself is questionable
- **Structural problems** (3 duplicate `itk_add_test` names, comma-list-to-`argv[1]` mismatch, wholesale gersemi reformat) → all resolved
- **Test now passes locally on macOS** in 3.59s — `pre-commit run --all-files` exits 0

The two test invocations targeting `RGBDicomSeries/raw-RGB.dcm` and `RTDoseDicomSeries/32bitDoseRTImage.dcm` are dropped because those fixtures were never published to ITKTestingData. The `Input/DicomSeries/Image0075.dcm` invocation is preserved and exercises 13 distinct `GetElement*` accessors.

@jhlegarreta — your authorship of the original test concept is preserved via `Co-Authored-By:` trailers on the new commits. Thanks for opening this in 2021 even when the build couldn't yet support it. @issakomi — your review feedback shaped the corrected test; thank you.

Closing this PR; please follow the discussion on #6189.

PR #2811, opened 2021-10-13
Fix various Clang-Tidy warnings from the three itk*IOFactoryRegisterManager.h.in files

N-Dekker:


N-Dekker:
> In certain contexts, Clang plus `-flto` can be very aggressive to remove "what is not used," and I do not want this code to be discarded by the compiler.

Ah, I understand your concern, but I think it's OK. The most fundamental part is the construction of those `RegisterManager` objects, right? Those objects are still "non-constexpr" with this PR. (I only declared them "const", not "constexpr"). It's just the `RegisterRegisterList` C arrays of function pointers that are declared `constexpr` with this PR. I think that should be just fine. Right? As long as the `RegisterManager(list)` constructors get called, it should be fine  😃

thewtex:
> Because I think that they are just meant as internal implementation detail of the corresponding RegisterManager instance.

Yes

N-Dekker:
> > Because I think that they are just meant as internal implementation detail of the corresponding RegisterManager instance.
> 
> Yes

Thanks, because then the next improvement could be to make those `RegisterRegisterList` arrays non-global, local to the construction of  of the corresponding `RegisterManager` instance.  😃

PR #2826, opened 2021-10-19
STYLE: Use `auto` for declaration of variables initialized by `New()`

N-Dekker:
Replaced initializations of the form `T::Pointer var = T::New()`
(sometimes starting with `typename`) with `auto var = T::New()`, to
reduce code redundancy.

In accordance with C++ Core Guidelines, August 19, 2021:
"ES.11: Use `auto` to avoid redundant repetition of type names"
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-auto

Using Notepad++ v8.1.4, Find in Files (Regular expression):

    Find what:  typename (\w+)::Pointer[ ]+(\w+) = \1::New\(\);
    Replace with:  auto $2 = $1::New\(\);
    Filters: itk*.hxx;itk*.h;itk*.cxx

And then again without `typename`:

    Find what:  (\w+)::Pointer[ ]+(\w+) = \1::New\(\);

Follow-up to commit de713e7ac52f7815a35754de885795ff0a1c4981
"STYLE: Use auto for variable creation", Hans Johnson (@hjmjohnson), 9 February 2018.


SimonRit:
Dankjewel Niels! Can you quickly explain why, in your opinion, the new version has a "better" style than the older one? It is not obvious to me and it would be easier to adopt the new style if I'm convinced it's better.

Leengit:
This aging C++ programmer tries to avoid `auto` except where the explicit type is a huge mouthful such as with iterators or lambda expressions, or is operating system dependent; and I would be happy enough keeping `typename VariableVectorInterpolatorType::Pointer`.  
(Or, I'd use
```c++
using VariableVectorInterpolatorTypePointer = typename VariableVectorInterpolatorType::Pointer;
VariableVectorInterpolatorTypePointer interpolator{ VariableVectorInterpolatorType::New() };
```
)

My reasoning has to do with the fact that the compiler can catch a type mismatch in milliseconds (microseconds?) but it sometimes takes me hours to track down a bug caused by an unexpected type.

But this is not a show stopper -- unless you are persuaded too. :smile:

N-Dekker:
Thanks for your comment @Leengit 

> I would be happy enough keeping `typename VariableVectorInterpolatorType::Pointer`. (Or, I might use
> 
> ```c++
> using VariableVectorInterpolatorTypePointer = typename VariableVectorInterpolatorType::Pointer;
> VariableVectorInterpolatorTypePointer interpolator{ VariableVectorInterpolatorType::New() };
> ```
> )

Wow, so unfortunate to hear you prefer those two lines to a simple one-liner, using `auto`:

```c++
auto interpolator = VariableVectorInterpolatorType::New();
```

I can understand that using `auto` may be disputable for an arbitratry initialization, as in `auto answer = Compute(42);` However, the return type of `VariableVectorInterpolatorType::New()` should be obvious to any ITK developer. Redundantly spelling out the type name multiple times would more likely introduce typo's than prevent them, as far as I can see.

N-Dekker:
For the record, it appears that this pull request is in accordance with the ITK Coding Style (from the ITK Software Guide), and specifically the section **_The auto Keyword_**, added by Matt McCormick (@thewtex) commit https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/pull/59/commits/51e92270f316bea966d4da492579ef752a5ca97d which says:

> The `auto` keyword should be used to specify a type in the following cases:
> -  The type is duplicated on the left side of an assigment when it is mandated on the right side, e.g. when there is an explicit cast or initializing with \code{new} or ITK's `::New()`.
...

At https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/blob/v5.2.1/SoftwareGuide/Latex/Appendices/CodingStyleGuide.tex#L1507-L1511

So if you agree, please say so by approving this pull request  😃

PR #2828, opened 2021-10-20
ENH: Support np.asarray on itk.Matrix

thewtex:


HastingsGreer:
That's a good point! Going forward, image orientations in python will be much more accessible :)

PR #2847, opened 2021-10-28
ENH: Add UTF-8 support for ImageSeriesReader when reading series of i…

9a4gl:
On Windows ImageSeriesReader is not able to open images in folders that contain non-ASCII characters. As seen on many other places in ITK and GDCM code, we can use std::ifstream constructor that takes wchar_t* for filename, which is nonstandard but available only on Windows with Microsoft compiler. That way, if user pass utf8 filename any folder/filename can be opened.

This can also be related to:
https://discourse.itk.org/t/gdcmseriesfilenames-fails-to-read-folder-with-non-ascii-characters/2365

### DIFF
--- a/Modules/IO/ImageBase/include/itkImageFileReader.hxx
+++ b/Modules/IO/ImageBase/include/itkImageFileReader.hxx
@@ -275,7 +275,12 @@ ImageFileReader<TOutputImage, ConvertPixelTraits>::TestFileExistanceAndReadabili
 
   // Test if the file can be open for reading access.
   std::ifstream readTester;
+#ifdef _MSC_VER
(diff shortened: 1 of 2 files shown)


thewtex (commented):
@9a4gl thank you for the contribution!

As @dzenanz suggested, could a new test please be added that uses a filename with characters that causes the issue you encountered?

9a4gl:
> @9a4gl thank you for the contribution!
> 
> As @dzenanz suggested, could a new test please be added that uses a filename with characters that causes the issue you encountered?

Will try to find some time for that next week.

thewtex:
Merging this for inclusion in 5.3RC3, but #3028 was created to track a follow-up with test coverage.

Issue #2849, opened 2021-10-28
Remove MetaProgrammingLibrary and replace by C++14 STL equivalents


dzenanz:
Since we started requiring C++11 `itk::mpl` is fairly redundant. Even more so now that we require a minimum of C++14. It is desireable to completely remove this and rely on standard template library and language features directly.

Part of this is already accomplished, mostly by #2730, but also in part by d3e151f09d04830c8ef6ac2f2b104d247ffb99f3 and perhaps some other, earlier refactoring.

PR #2854, opened 2021-11-02
COMP: Fix bit shift overflow warning

blowekamp:
itkStatisticsLabelMapFilter.hxx(77): warning C4293: '<<': shift count
negative or too big, undefined behavior

## PR Checklist

### DIFF
--- a/Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.h
+++ b/Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.h
@@ -137,9 +137,10 @@ public:
   static constexpr unsigned int
   GetDefaultNumberOfBins()
   {
-    return NumericTraits<FeatureImagePixelType>::IsInteger && sizeof(FeatureImagePixelType) <= 2
-             ? 1 << (8 * sizeof(FeatureImagePixelType))
-             : 128;
+    constexpr size_t bitsShift = std::min(8 * sizeof(FeatureImagePixelType), 8 * sizeof(m_NumberOfBins) - 1);
+
+    return std::is_integral<FeatureImagePixelType>::value && sizeof(FeatureImagePixelType) <= 2 ? 1u << bitsShift
+                                                                                                : 128u;
   }
 
 protected:
--- a/Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.hxx
+++ b/Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.hxx
@@ -73,8 +73,9 @@ StatisticsLabelMapFilter<TImage, TFeatureImage>::ThreadedProcessLabelObject(Labe
   typename HistogramType::MeasurementVectorType featureImageMax(1);
 
 
-  if (NumericTraits<typename Self::FeatureImagePixelType>::IsInteger &&
-      m_NumberOfBins == 1 << (8 * sizeof(typename Self::FeatureImagePixelType)))
+  constexpr size_t bitsShift = std::min(8 * sizeof(FeatureImagePixelType), 8 * sizeof(m_NumberOfBins) - 1);
+  if (std::is_integral<FeatureImagePixelType>::value && sizeof(FeatureImagePixelType) <= 2 &&
+      m_NumberOfBins == 1 << bitsShift)
   {
     // Add padding so the center of bins are integers
     featureImageMin.Fill(NumericTraits<typename Self::FeatureImagePixelType>::min() - 0.5);


Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.h
```diff
@@ -137,9 +137,11 @@ class ITK_TEMPLATE_EXPORT StatisticsLabelMapFilter
   static constexpr unsigned int
   GetDefaultNumberOfBins()
   {
-    return NumericTraits<FeatureImagePixelType>::IsInteger && sizeof(FeatureImagePixelType) <= 2
-             ? 1 << (8 * sizeof(FeatureImagePixelType))
-             : 128;
+    constexpr unsigned short bitsShift =
+      std::min(8 * sizeof(typename Self::FeatureImagePixelType), sizeof(unsigned int) - 1);
```
Leengit, line 141: I believe that an `8` is missing; it should be:
```c++
std::min(8 * sizeof(typename Self::FeatureImagePixelType), 8 * sizeof(unsigned int) - 1);
```

Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.h
```diff
@@ -137,9 +137,11 @@ class ITK_TEMPLATE_EXPORT StatisticsLabelMapFilter
...
+
+    return NumericTraits<FeatureImagePixelType>::IsInteger && sizeof(FeatureImagePixelType) <= 2 ? 1u << bitsShift
+                                                                                                 : 128;
```
Leengit, line 144: It may be for style reasons only, but how about using `128u` for symmetry with `1u`?

Leengit (commented):
Thanks!  Two comments.

Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.h
blowekamp: ugh, that was sloppy.

seanm: Why the implicit downsize to `short`? ie why not `size_t` (what `sizeof` returns)?

blowekamp: done

Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.h
blowekamp: done

Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.hxx
```diff
@@ -73,8 +73,10 @@ StatisticsLabelMapFilter<TImage, TFeatureImage>::ThreadedProcessLabelObject(Labe
   typename HistogramType::MeasurementVectorType featureImageMax(1);
 
 
-  if (NumericTraits<typename Self::FeatureImagePixelType>::IsInteger &&
-      m_NumberOfBins == 1 << (8 * sizeof(typename Self::FeatureImagePixelType)))
+  constexpr size_t short bitsShift =
+    std::min(8 * sizeof(typename Self::FeatureImagePixelType), 8 * sizeof(unsigned int) - 1);
+  if (NumericTraits<typename Self::FeatureImagePixelType>::IsInteger && sizeof(FeatureImagePixelType) <= 2 &&
+      m_NumberOfBins == 1 << bitsShift)
```
Leengit, line 79: I assume that this pull request is needed because there is a system where `sizeof(unsigned int)` is only 2.  In such a case and only in such a case, this pull request uses a bitsShift of 15 (rather than 16 as previously) in the case that `sizeof(FeatureImagePixelType)` is also 2.  Unfortunately, the goal is to have 2<sup>16</sup> buckets in such a case, centered on the integer values.  Instead we will have only 2<sup>15</sup> buckets.  I guess the padding / centering is still appropriate....

However, I think the best solution would be to use 2<sup>16</sup> buckets in such a case.  Can that be accomplished on every system by using `unsigned long` in key places?

blowekamp: > I assume that this pull request is needed because there is a system where `sizeof(unsigned int)` is only 2.
No. It occurred with when the pixel type was 64-bits, and the size of m_NumberOfBins was only 32-bits.

Leengit: Correct me if I am wrong, but the problem is that in the old code
```c++
    return NumericTraits<FeatureImagePixelType>::IsInteger && sizeof(FeatureImagePixelType) <= 2
             ? 1 << (8 * sizeof(FeatureImagePixelType))
             : 128;
```
the `1 << (8 * sizeof(FeatureImagePixelType))` was causing a compiler warning even when code was choosing the `128` branch.

blowekamp: Correct.

Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.h
```diff
@@ -137,9 +137,11 @@ class ITK_TEMPLATE_EXPORT StatisticsLabelMapFilter
   static constexpr unsigned int
   GetDefaultNumberOfBins()
   {
-    return NumericTraits<FeatureImagePixelType>::IsInteger && sizeof(FeatureImagePixelType) <= 2
-             ? 1 << (8 * sizeof(FeatureImagePixelType))
-             : 128;
+    constexpr size_t bitsShift =
+      std::min(8 * sizeof(typename Self::FeatureImagePixelType), 8 * sizeof(unsigned int) - 1);
```
N-Dekker, line 141: Please consider using `CHAR_BIT`, instead of `8` (two times), according to https://clang.llvm.org/extra/clang-tidy/checks/readability-magic-numbers.html

Why do you use `typename Self::FeatureImagePixelType`, instead of simply `FeatureImagePixelType` (as on the next line)?

Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.h
```diff
@@ -137,9 +137,11 @@ class ITK_TEMPLATE_EXPORT StatisticsLabelMapFilter
...
+
+    return NumericTraits<FeatureImagePixelType>::IsInteger && sizeof(FeatureImagePixelType) <= 2 ? 1u << bitsShift
```
N-Dekker, line 143: Please consider using C++11 `std::is_integral<FeatureImagePixelType>::value`.

blowekamp: yes, done.

Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.h
blowekamp: no, yes. done.

dzenanz (approved):
Looks good on a glance.

Using STL functions directly is preferred, as we would like to remove redundancy from ITK, see #2849.

Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.hxx
```diff
@@ -73,8 +73,9 @@ StatisticsLabelMapFilter<TImage, TFeatureImage>::ThreadedProcessLabelObject(Labe
   typename HistogramType::MeasurementVectorType featureImageMax(1);
 
 
-  if (NumericTraits<typename Self::FeatureImagePixelType>::IsInteger &&
-      m_NumberOfBins == 1 << (8 * sizeof(typename Self::FeatureImagePixelType)))
+  constexpr size_t bitsShift = std::min(8 * sizeof(FeatureImagePixelType), 8 * sizeof(m_NumberOfBins) - 1);
+  if (std::is_integral<FeatureImagePixelType>::value && sizeof(FeatureImagePixelType) <= 2 &&
+      m_NumberOfBins == 1 << bitsShift)
```
Leengit, line 78: STYLE: Use `1u` here.

Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.h
```diff
@@ -137,9 +137,10 @@ class ITK_TEMPLATE_EXPORT StatisticsLabelMapFilter
   static constexpr unsigned int
   GetDefaultNumberOfBins()
   {
-    return NumericTraits<FeatureImagePixelType>::IsInteger && sizeof(FeatureImagePixelType) <= 2
-             ? 1 << (8 * sizeof(FeatureImagePixelType))
-             : 128;
+    constexpr size_t bitsShift = std::min(8 * sizeof(FeatureImagePixelType), 8 * sizeof(m_NumberOfBins) - 1);
+
+    return std::is_integral<FeatureImagePixelType>::value && sizeof(FeatureImagePixelType) <= 2 ? 1u << bitsShift
+                                                                                                : 128u;
```
Leengit, line 143: Since we are checking whether `sizeof(FeatureImagePixelType) <= 2`, this method could be re-written as:
```c++
  GetDefaultNumberOfBins()
  {
    return std::is_integral<FeatureImagePixelType>::value && sizeof(FeatureImagePixelType) <= 2
             ? 1u << 8 * std::min(2, sizeof(FeatureImagePixelType))
             : 128u;
  }
```

Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.hxx
Leengit, line 78: Since we are checking whether `sizeof(FeatureImagePixelType) <= 2`, these three lines of code could be:
```c++
    if (std::is_integral<FeatureImagePixelType>::value && sizeof(FeatureImagePixelType) <= 2 &&
        m_numberofbins == 1u << 8 * std::min(2, sizeof(FeatureImagePixelType)))
```

Leengit (approved):
I am suggesting alternatives even though your code is amply good as it is.  Thanks!

PR #2859, opened 2021-11-07
BUG: Fix constexpr linkage issue

hjmjohnson:
Reason: Prior to C++17 one must  provide the definition of the static
constexpr member as well as the declaration. The declaration and the
initializer go inside the class definition, but the member definition
has to be separate.

FAILED: bin/ITKCommon2TestDriver
  "itk::BSplineInterpolationWeightFunction<float, 2u, 1u>::SupportSize", referenced from:
      itk::BSplineInterpolationWeightFunction<float, 2u, 1u>::GetSupportSize() const in itkBSplineInterpolationWeightFunctionTest.cxx.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
[1081/1334] Building CXX object Modules/Registration/RegistrationMethodsv4/test/CMakeFiles/ITKRegistrationMethodsv4TestDriver.dir/itkSimpleImageRegistrationTestWithMaskAndSampling.cxx.o
ninja: build stopped: subcommand failed.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h
+++ b/Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h
@@ -87,6 +87,7 @@ public:
 
   /** The support region size: a hypercube of length SplineOrder + 1 */
   static constexpr SizeType SupportSize{ SizeType::Filled(VSplineOrder + 1) };
+  // Declaration here, definition must be outside of class until C++17, see .hxx file for linker definition
 
   /** Evaluate the weights at specified ContinuousIndex position.
    * Subclasses must provide this method. */
--- a/Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.hxx
+++ b/Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.hxx
@@ -26,6 +26,20 @@
 
 namespace itk
 {
+
+#if __cplusplus < 201703L
+// For compatibility with pre C++17 International Standards, a constexpr static
+// data member may be redundantly redeclared outside the class with no
+// initializer. This usage is deprecated.
+//
+// For C++14 and earlier, the definition of static constexpr must be outside of
+// class until C++17, see .h file for declaration & initialization
+
+template <typename TCoordRep, unsigned int VSpaceDimension, unsigned int VSplineOrder>
+constexpr typename BSplineInterpolationWeightFunction<TCoordRep, VSpaceDimension, VSplineOrder>::SizeType
+  BSplineInterpolationWeightFunction<TCoordRep, VSpaceDimension, VSplineOrder>::SupportSize;
+#endif
+
 /** Compute weights for interpolation at continuous index position */
 template <typename TCoordRep, unsigned int VSpaceDimension, unsigned int VSplineOrder>
 typename BSplineInterpolationWeightFunction<TCoordRep, VSpaceDimension, VSplineOrder>::WeightsType


dzenanz:
This has a bunch of [compile errors](https://open.cdash.org/viewBuildError.php?buildid=7557769) on the CI.

hjmjohnson:
> This has a bunch of [compile errors](https://open.cdash.org/viewBuildError.php?buildid=7557769) on the CI.

I saw that yesterday.  I have not been able to replicate the errors on Ubuntu 20.04 or Mac.  I'll need to dig in further later tonight.

Modules/Core/Common/test/itkBSplineInterpolationWeightFunctionTest.cxx
```diff
@@ -154,7 +164,7 @@ itkBSplineInterpolationWeightFunctionTest(int, char *[])
     IndexType startIndex1;
     IndexType startIndex2;
 
-    double testFailed = false;
+    bool testFailed = false;
```
N-Dekker, line 167: Please consider moving this adjustment into a commit, as it appears unrelated to the constexpr linkage issue. (Following the "One Commit. One Change" rule.) If you want to fix `double testFailed = false`, please do it on both places (line 167 _and_ line 70) .

Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h
```diff
@@ -87,6 +87,7 @@ class ITK_TEMPLATE_EXPORT BSplineInterpolationWeightFunction
 
   /** The support region size: a hypercube of length SplineOrder + 1 */
   static constexpr SizeType SupportSize{ SizeType::Filled(VSplineOrder + 1) };
+  // Declaration here, definition must be outside of class until C++17, see .hxx file for linker definition
```
N-Dekker, line 90: Thanks very much for addressing this problem @hjmjohnson I did not realize that my pull request #2711 would cause those link errors, sorry!

Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h
```diff
@@ -106,7 +107,7 @@ class ITK_TEMPLATE_EXPORT BSplineInterpolationWeightFunction
 
 #if !defined(ITK_LEGACY_REMOVE)
   /** Get support region size. */
-  itkLegacyMacro(SizeType GetSupportSize() const) { return Self::SupportSize; };
+  itkLegacyMacro(SizeType GetSupportSize() const) { return SizeType::Filled(VSplineOrder + 1); };
```
N-Dekker, line 110: @hjmjohnson Now that you have defined `SupportSize` in the proper C++14 compatible way, is it still necessary to adjust `GetSupportSize()` like you propose here? If technically possible, I still prefer a simple `return Self::SupportSize`.

Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.hxx
```diff
@@ -18,14 +18,19 @@
...
-#include "itkBSplineInterpolationWeightFunction.h"
 #include "itkBSplineKernelFunction.h"
 #include "itkImage.h"
 #include "itkMatrix.h"
 #include "itkMath.h"
 
 namespace itk
 {
+
+// Definition of static constexpr must be outside of class until C++17, see .h file for declaration & initialization
```
N-Dekker, line 29: Can you please add something like `#if __cplusplus < 201703L` here? With C++17 it is actually deprecated to define a static constexpr outside of the class definition!!!

In the [Working Draft of the C++ Standard](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/n4901.pdf) it says:
> For compatibility with prior C++ International Standards, a `constexpr` static data member may be redundantly
redeclared outside the class with no initializer. This usage is deprecated.

Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.hxx
```diff
@@ -18,14 +18,19 @@
 #ifndef itkBSplineInterpolationWeightFunction_hxx
 #define itkBSplineInterpolationWeightFunction_hxx
 
-#include "itkBSplineInterpolationWeightFunction.h"
```
N-Dekker, line 21: Why should `#include "itkBSplineInterpolationWeightFunction.h"` be removed from the corresponding hxx file?

hjmjohnson: I'll deal with this more uniformly in a different commit.  The issues is IDE warning based.  The recursive nature of the .h file including the .hxx file which in turn includes the .h file again causes the IDE to issue warnings about unterminated @ifdef's 

I don't see why the recurse calls should be there unless you want to #include the .hxx instead of the .h file, which I don't see as necessary.

N-Dekker: @hjmjohnson Wasn't the #include "itk....h" there in the hxx file in order to support `ITK_MANUAL_INSTANTIATION`? So that users could have a dedicated cxx file that would only #include hxx files, and then manually instantiate what they need?  I'm just asking.

hjmjohnson: I think that was the theory back in 2001, but I don't think that use case is in practice.  In the possible very rare case that it was used that way, we could simply include the .h file and the .hxx file in the cxx file.

dzenanz (approved):

PR #2862, opened 2021-11-09
ENH: Added template parameter for texture mask type

emmenlau:
Currently `ScalarImageToCooccurrenceMatrixFilter` assumes that the mask image would be the same type as the gray value image. This is not ideal because masks do not necessarily have to be of the same type. This limits the usability of the current design. What is more important is that forcing the mask to be of the same type can be problematic when working with floating-point types.

This PR extends `ScalarImageToCooccurrenceMatrixFilter` to add a template parameter for the mask image type, and uses the updated filter in `ScalarImageToTextureFeaturesFilter`.

This is my first contribution to ITK and I do not feel confident to say if the PR is breaking the API. It *does* add a new template parameter (with a default value) in front of an existing template parameter. So people that used both parameters before will not be able to build against this change.

Suggestions for how to change this PR would be very welcome!

## PR Checklist

### DIFF
--- a/Modules/Numerics/Statistics/include/itkScalarImageToCooccurrenceMatrixFilter.h
+++ b/Modules/Numerics/Statistics/include/itkScalarImageToCooccurrenceMatrixFilter.h
@@ -90,7 +90,9 @@ namespace Statistics
  * \ingroup ITKStatistics
  */
 
-template <typename TImageType, typename THistogramFrequencyContainer = DenseFrequencyContainer2>
+template <typename TImageType,
+          typename THistogramFrequencyContainer = DenseFrequencyContainer2,
+          typename TMaskImageType = TImageType>
 class ITK_TEMPLATE_EXPORT ScalarImageToCooccurrenceMatrixFilter : public ProcessObject
 {
 public:
@@ -118,6 +120,10 @@ public:
   using OffsetVector = VectorContainer<unsigned char, OffsetType>;
   using OffsetVectorPointer = typename OffsetVector::Pointer;
   using OffsetVectorConstPointer = typename OffsetVector::ConstPointer;
+  using MaskImageType = TMaskImageType;
+  using MaskPointer = typename MaskImageType::Pointer;
+  using MaskConstPointer = typename MaskImageType::ConstPointer;
+  using MaskPixelType = typename MaskImageType::PixelType;
 
   using MeasurementType = typename NumericTraits<PixelType>::RealType;
 
@@ -164,9 +170,9 @@ public:
 
   /** Method to set/get the mask image */
   void
-  SetMaskImage(const ImageType * image);
+  SetMaskImage(const MaskImageType * image);
 
-  const ImageType *
+  const MaskImageType *
   GetMaskImage() const;
 
   /** method to get the Histogram */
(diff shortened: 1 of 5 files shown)


Modules/Numerics/Statistics/include/itkScalarImageToCooccurrenceMatrixFilter.h
```diff
@@ -90,7 +90,9 @@ namespace Statistics
  * \ingroup ITKStatistics
  */
 
-template <typename TImageType, typename THistogramFrequencyContainer = DenseFrequencyContainer2>
+template <typename TImageType,
+          typename TMaskType = TImageType,
```
dzenanz, line 94: I understand that this is a more logical sequence of template parameters. But for painless backwards compatibility it is way better to put the new parameter last. That way other people will not need to change their code.

Modules/Numerics/Statistics/include/itkScalarImageToTextureFeaturesFilter.h
```diff
@@ -105,7 +105,9 @@ namespace Statistics
  * \endsphinx
  */
 
-template <typename TImageType, typename THistogramFrequencyContainer = DenseFrequencyContainer2>
+template <typename TImageType,
+          typename TMaskType = TImageType,
```
dzenanz, line 109: Change order, same as above.

dzenanz (approved):

blowekamp:
Looks like a very nice PR. Thank you for the contribution.

Given that we have `SetMaskImage` and `GetMaskImage` the expected typename is `MaskImageType`. Could you please update the `MaskType` to `MaskImageType`. Thank you.

emmenlau:
I've changed the `MaskType` to `MaskImageType` and `TMaskType` to `TMaskImageType`. Also, I've modified the test to actually use a different mask type than the image type, so test the new behavior. Everything builds and tests for me.

dzenanz:
There is now a [warning on Windows](https://open.cdash.org/viewBuildError.php?type=1&buildid=7564536): `itkScalarImageToCooccurrenceMatrixFilter.hxx(274): warning C4805: '!=': unsafe mix of type 'bool' and type 'unsigned char' in operation`. So your test expansion uncovered a problem with your patch 😄👍. `InsidePixelValue` should have the same type as mask image, not the intensity image. That should be an easy fix.

emmenlau:
> So your test expansion uncovered a problem with your patch smile+1. `InsidePixelValue` should have the same type as mask image, not the intensity image. That should be an easy fix.

Very well spotted! Fixed in latest version. While I was looking into this, I also saw that `region` is of type `ImageType::RegionType`  but used in the iterator for the image and the mask. Is that ok, or what is the clean way to deal with this? A static assert that ensures both image types have the same region type? Like
```
static_assert(std::is_same_v<typename ImageType::RegionType, typename MaskImageType::RegionType>, "error");
```

dzenanz:
If images have the same dimension, their region type will be the same. I think that same dimension is enforced by default somewhere in the processing pipeline.

Modules/Numerics/Statistics/include/itkScalarImageToCooccurrenceMatrixFilter.h
```diff
@@ -215,7 +221,7 @@ class ITK_TEMPLATE_EXPORT ScalarImageToCooccurrenceMatrixFilter : public Process
   MeasurementVectorType m_UpperBound;
   bool                  m_Normalize;
 
-  PixelType m_InsidePixelValue;
+  MaskPixelType m_InsidePixelValue;
```
dzenanz, line 224: On lines 184-185
```cpp
itkSetMacro(InsidePixelValue, PixelType);
itkGetConstMacro(InsidePixelValue, PixelType);
```
also needs to be changed to `MaskPixelType`.

dzenanz:
Do you plan to finish this PR @emmenlau?

emmenlau:
Could you please take a look again? Thanks for the review!

dzenanz (approved):

PR #2865, opened 2021-11-11
STYLE: Do not do `return;` directly after throwing an exception

N-Dekker:
Any return statement directly after a `throw` (possibly from a call to `itkExceptionMacro` or `itkGenericExceptionMacro`) is unreachable, by definition.

### DIFF
--- a/Modules/Bridge/VTK/include/itkVTKImageImport.hxx
+++ b/Modules/Bridge/VTK/include/itkVTKImageImport.hxx
@@ -125,7 +125,6 @@ VTKImageImport<TOutputImage>::PropagateRequestedRegion(DataObject * outputPtr)
   if (!output)
   {
     itkExceptionMacro(<< "Downcast from DataObject to my Image type failed.");
-    return;
   }
   Superclass::PropagateRequestedRegion(output);
   if (m_PropagateUpdateExtentCallback)
--- a/Modules/Core/Common/include/itkByteSwapper.hxx
+++ b/Modules/Core/Common/include/itkByteSwapper.hxx
@@ -129,7 +129,6 @@ ByteSwapper<T>::SwapRangeFromSystemToBigEndian(T * p, BufferSizeType num)
       return;
     default:
       itkGenericExceptionMacro(<< "Cannot swap number of bytes requested");
-      return;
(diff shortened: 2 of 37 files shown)


dzenanz (approved):

hjmjohnson (approved):
Nice work!  My favorite way to increase code coverage... Delete code :).

N-Dekker:
Interestingly, MSVC produces a warning for an unreachable `return returnValue;`, but not for a simple `return;` (returning nothing)! MSVC v19.29 VS16.11 at https://godbolt.org/z/cYM9KqGj6 having warning level 4 (`/W4`):

```
void ReturnNothing()
{
    throw Exception{};
    return;  // No warning!
}

void ReturnLambda()
{
    throw Exception{};
    return []{}(); // warning C4702: unreachable code
}

int ReturnInt()
{
    throw Exception{};
    return 1; // warning C4702: unreachable code
}
```

Leengit:
Thanks!!  Also,
  Modules/IO/TransformMINC/include/itkMINCTransformAdapter.h
  Modules/IO/NIFTI/test/itkNiftiImageIOTest.h
  Modules/Core/GPUCommon/include/itkGPUReduction.hxx

N-Dekker:
> Thanks!! Also,
> Modules/IO/TransformMINC/include/itkMINCTransformAdapter.h
> Modules/IO/NIFTI/test/itkNiftiImageIOTest.h
> Modules/Core/GPUCommon/include/itkGPUReduction.hxx

Nice catch @Leengit but I consider them a slightly different category: these are cases the unreachable return statement would return a non-void value. With this pull request I specifically looked at "empty" (void) return statements. The unreachable non-void return statements can be found, just by enabling MSVC warning C4702. Unfortunately it is a Warning Level 4 warning, while ITK typically builds at Warning Level 3. Moreover, it is explicitly disabled on a few places, for example at https://github.com/InsightSoftwareConsortium/ITK/blob/v5.3rc02/Modules/ThirdParty/DICOMParser/src/DICOMParser/DICOMCallback.h#L27  

The remaining unreachable lines of code could be dealt with in another pull request, of course.

Issue #983, opened 2019-06-01
Make ITK exceptions and their tests specific


jhlegarreta:
### Description

ITK **exceptions** are largely **non-specific**, i.e. a general `itkExceptionMacro` can loosely used to throw virtually any type of exception. Making exceptions be more specific, including the messages, would make the toolkit more consistent.

At the same time, the `TRY_EXPECT_EXCEPTION` macro is **non-specific**, meaning that it **expects the statement to throw an exception, whichever the exception is**. **When writing such a statement in a test, the programmer generally expects a specific exception to be caught**, i.e. an exception with a known message to be thrown. However, **if the code is not designed correctly or hides a bug, the statement may well catch an exception, but one that has been thrown elsewhere/is different from the one expected**. **The test or macro would work correctly**, because it does not check the exception type (or better, the source of the exception, or both).

PR #511 [uncovered such a situation](https://github.com/InsightSoftwareConsortium/ITK/pull/511#issuecomment-497877457).

### Impact analysis

Having non-specific exceptions provides a flexible framework, but may hide bugs. Also, if there is no way to check the exception type and/or the class that threw it, when expecting an exception, any other exception thrown in the call stack may be considered as a valid exception

### Expected behavior

Creating a specific set of exceptions and enhancing the `TRY_EXPECT_EXCEPTION` to accept the class and exception type would help in making the code more consistent, and robust.

e.g.
```
TRY_EXPECT_EXCEPTION(filter->Update(), FilterClass, ExceptionType);
```

### Actual behavior

Current exception checking:
```
TRY_EXPECT_EXCEPTION(filter->Update);
```

### Versions

`master`.

### Additional Information

None.

jhlegarreta:
This is still relevant and can help uncovering hidden bugs.

N-Dekker:
FYI, GoogleTest does have an `EXPECT_THROW(statement, expected_exception)` macro:
https://github.com/InsightSoftwareConsortium/ITK/blob/v5.3rc02/Modules/ThirdParty/GoogleTest/src/itkgoogletest/googletest/include/gtest/gtest.h#L1952

Being used in various ITK GTest unit tests, for example at:
https://github.com/InsightSoftwareConsortium/ITK/blob/v5.3rc02/Modules/Core/Transform/test/itkTranslationTransformGTest.cxx#L41

jhlegarreta:
[...]

PR #2869, opened 2021-11-12
COMP: Remove unreachable code

N-Dekker:
Removed code that would produce MSVC warning C4702, "unreachable code"
(if this warning would have been enabled), from Visual Studio 2019.

Explicitly deleted (`= delete`) the function template
`PreFillDirection()`, which also had unreachable code.

Most of these changes were already suggested by Lee Newberg (@Leengit) at
https://github.com/InsightSoftwareConsortium/ITK/pull/2865#issuecomment-966483880

Follow-up to pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2865
commit ad64a79f657ecbf59fec980ca37c700222008ebc
"STYLE: Do not do `return;` directly after throwing an exception"

### DIFF
--- a/Modules/Filtering/FFT/test/itkComplexToComplexFFTImageFilterTest.cxx
+++ b/Modules/Filtering/FFT/test/itkComplexToComplexFFTImageFilterTest.cxx
@@ -98,6 +98,16 @@ itkComplexToComplexFFTImageFilterTest(int argc, char * argv[])
   const char *      outputImageFileName = argv[2];
   const std::string pixelTypeString(argv[3]);
 
+  // Test streaming enumeration for ComplexToComplexFFTImageFilterEnums::TransformDirection elements
+  const std::set<itk::ComplexToComplexFFTImageFilterEnums::TransformDirection> allTransformDirection{
+    itk::ComplexToComplexFFTImageFilterEnums::TransformDirection::FORWARD,
+    itk::ComplexToComplexFFTImageFilterEnums::TransformDirection::INVERSE
+  };
+  for (const auto & ee : allTransformDirection)
+  {
+    std::cout << "STREAMED ENUM VALUE ComplexToComplexFFTImageFilterEnums::TransformDirection: " << ee << std::endl;
+  }
+
   itk::ImageIOBase::Pointer imageIO =
     itk::ImageIOFactory::CreateImageIO(inputImageFileName, itk::ImageIOFactory::IOFileModeEnum::ReadMode);
   imageIO->SetFileName(inputImageFileName);
@@ -116,7 +126,6 @@ itkComplexToComplexFFTImageFilterTest(int argc, char * argv[])
         std::cerr << "Unknown image dimension." << std::endl;
         return EXIT_FAILURE;
     }
-    return EXIT_SUCCESS;
   }
   else if (pixelTypeString.compare("double") == 0)
   {
@@ -130,23 +139,10 @@ itkComplexToComplexFFTImageFilterTest(int argc, char * argv[])
         std::cerr << "Unknown image dimension." << std::endl;
         return EXIT_FAILURE;
     }
-    return EXIT_SUCCESS;
   }
   else
   {
     std::cerr << "Unknown pixel type string." << std::endl;
     return EXIT_FAILURE;
   }
-
-  // Test streaming enumeration for ComplexToComplexFFTImageFilterEnums::TransformDirection elements
-  const std::set<itk::ComplexToComplexFFTImageFilterEnums::TransformDirection> allTransformDirection{
-    itk::ComplexToComplexFFTImageFilterEnums::TransformDirection::FORWARD,
-    itk::ComplexToComplexFFTImageFilterEnums::TransformDirection::INVERSE
-  };
-  for (const auto & ee : allTransformDirection)
-  {
-    std::cout << "STREAMED ENUM VALUE ComplexToComplexFFTImageFilterEnums::TransformDirection: " << ee << std::endl;
-  }
-
-  return EXIT_FAILURE;
 }
--- a/Modules/IO/NIFTI/test/itkNiftiImageIOTest.h
+++ b/Modules/IO/NIFTI/test/itkNiftiImageIOTest.h
@@ -54,16 +54,10 @@ constexpr unsigned char LEFT = 128;    /*Bit pattern 1 0 0  00000*/
 constexpr unsigned char ANTERIOR = 64; /*Bit pattern 0 1 0  00000*/
 constexpr unsigned char SUPERIOR = 32; /*Bit pattern 0 0 1  00000*/
 
+// The only specializations of this function template that may be called are given below.
 template <unsigned int TDimension>
 typename itk::ImageBase<TDimension>::DirectionType
-PreFillDirection()
-{
-  typename itk::ImageBase<TDimension>::DirectionType myDirection;
-  myDirection.Fill(0.0);
-  myDirection.SetIdentity();
-  itkGenericExceptionMacro("This template should never be used. Only valid values are given below.");
-  return myDirection;
-}
+PreFillDirection() = delete;
 
 template <>
 itk::ImageBase<1>::DirectionType
--- a/Modules/Core/GPUCommon/include/itkGPUReduction.hxx
+++ b/Modules/Core/GPUCommon/include/itkGPUReduction.hxx
@@ -116,7 +116,6 @@ GPUReduction<TElement>::GetReductionKernel(int whichKernel, int blockSize, int i
   if (whichKernel != 5 && whichKernel != 6)
   {
     itkExceptionMacro(<< "Reduction kernel undefined!");
-    return 0;
   }
 
   std::ostringstream defines;
--- a/Modules/IO/MINC/test/itkMINCImageIOTest_2D.cxx
+++ b/Modules/IO/MINC/test/itkMINCImageIOTest_2D.cxx
@@ -153,8 +153,4 @@ itkMINCImageIOTest_2D(int argc, char * argv[])
     std::cerr << excp << std::endl;
     return EXIT_FAILURE;
   }
-
-
-  std::cout << "Test finished." << std::endl;
-  return EXIT_SUCCESS;
 }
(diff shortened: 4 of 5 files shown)


Modules/IO/NIFTI/test/itkNiftiImageIOTest.h
```diff
@@ -54,16 +54,10 @@ constexpr unsigned char LEFT = 128;    /*Bit pattern 1 0 0  00000*/
...
 typename itk::ImageBase<TDimension>::DirectionType
-PreFillDirection()
-{
-  typename itk::ImageBase<TDimension>::DirectionType myDirection;
-  myDirection.Fill(0.0);
-  myDirection.SetIdentity();
-  itkGenericExceptionMacro("This template should never be used. Only valid values are given below.");
-  return myDirection;
-}
+PreFillDirection() = delete;
```
Leengit, line 60: I haven't seen this use of `delete` to incapacitate the non-specialized function template -- I like it!

Leengit (approved):
Looks good.

If it is in the purview of this pull request ... note that there are some `switch` statements here that are followed by code that is unreachable, because each `case` directly invokes `return`.  (How many compilers will figure out that this code is unreachable is a separate question.)

Modules/Filtering/FFT/test/itkComplexToComplexFFTImageFilterTest.cxx
```diff
@@ -137,16 +137,4 @@ itkComplexToComplexFFTImageFilterTest(int argc, char * argv[])
     std::cerr << "Unknown pixel type string." << std::endl;
     return EXIT_FAILURE;
   }
-
-  // Test streaming enumeration for ComplexToComplexFFTImageFilterEnums::TransformDirection elements
```
dzenanz, line 141: Should this block be moved before if starting on line 107? Instead of being removed.

dzenanz (approved):

Modules/Filtering/FFT/test/itkComplexToComplexFFTImageFilterTest.cxx
N-Dekker: @dzenanz Good question! I did not write the code, so honestly I don't know. But if I understand correctly, these ten lines of code just print two enum values. They don't really test any assertion. And then it does `return EXIT_FAILURE`. So whatever it printed, the test would always fail at this point. Personally, I think these lines of code may be removed.

dzenanz: They were probably there to increase code coverage. I wouldn't mind restoring them to operational status.

N-Dekker:
> If it is in the purview of this pull request ... note that there are some `switch` statements here that are followed by code that is unreachable, because each `case` directly invokes `return`. (How many compilers will figure out that this code is unreachable is a separate question.)

Good catch again, @Leengit I guess you refer to the `return EXIT_SUCCESS` statements in itkComplexToComplexFFTImageFilterTest.cxx, at:
- https://github.com/InsightSoftwareConsortium/ITK/blob/v5.3rc02/Modules/Filtering/FFT/test/itkComplexToComplexFFTImageFilterTest.cxx#L119
- https://github.com/InsightSoftwareConsortium/ITK/blob/v5.3rc02/Modules/Filtering/FFT/test/itkComplexToComplexFFTImageFilterTest.cxx#L133 

Right?

Modules/Filtering/FFT/test/itkComplexToComplexFFTImageFilterTest.cxx
N-Dekker: > I wouldn't mind restoring them to operational status.

As far as I can see, they never had operational status before  😸  Introduced with pull request #1471 (merged February last year), they were always unreachable: https://github.com/InsightSoftwareConsortium/ITK/blob/0dd31a62f851a0ae95ef4c6d6c85e531a4ebca59/Modules/Filtering/FFT/test/itkComplexToComplexFFTImageFilterTest.cxx#L147-L155

Anyway, if you want to, I can move these ten lines of code up, in order to make them reachable for the first time 😃

dzenanz: 😄 Yes, I would prefer that. And I assume @jhlegarreta would love it!

N-Dekker: If the aim is to test `operator<<`, I would prefer a test that inserts these two enum values into an `std::ostringstream`, instead of polluting the standard output! (And then doing test assertions on the result of `std::ostringstream`). But that's _way_ beyond the scope of this pull request. So I'll just move the printing code a few lines up for now.

jhlegarreta: Your assumption is correct @dzenanz :+1: . @N-Dekker thanks for these cleanup. Yes, please, move the block so that it can be reached and the coverage guaranteed.

N-Dekker: > Yes, please, move the block so that it can be reached and the coverage guaranteed.

OK, I will, but I would like to emphasize that proper unit testing is more that just code coverage. It should check, preferably in an automatic way, that the result of function calls is according to their specification, or expectation.

dzenanz: Of course. But even just invoking something can point out problems by failing to compile, or crashing at run time. Testing doesn't have to  be an all-or-nothing approach. If we try to make it that, I expect that we will be getting nothing a lot more often than all 😞

jhlegarreta: > OK, I will, but I would like to emphasize that proper unit testing is more that just code coverage. It should check, preferably in an automatic way, that the result of function calls is according to their specification, or expectation.

Agreed. Another example of the need you point is in #983. I have failed to find the time and peace of mind to try to propose a solution during this time.

An issue can be opened concerning the enums' streaming checking. Bug fixes and PRs lead to finding new room for improvement, but IMHO trying to address many issues at a time makes things less easier or PRs to stall (and I do have some of these :worried:). So we all try our best in here in the time we have.

jhlegarreta (approved):
Have not tried, but it looks :100:. Thanks for doing this @N-Dekker !

Leengit (approved):

dzenanz (approved):

N-Dekker:
> /azp run ITK.Windows

Thanks @dzenanz  For the record, the (apparently unrelated) test failure came from itkSPSAOptimizerTest

Site Name: | Azure.WIN-09DUQDFK1QA
-- | --
Build Name: | Windows_NT-Build6936-PR2869-Remove-unreachable-code
Total time: | 27m 52s 800ms
OS Name: | Windows
OS Platform: | AMD64
OS Release: | DataCenter Server
OS Version: | (Build 17763)

https://open.cdash.org/viewTest.php?onlyfailed&buildid=7569491

Test output:
> SPSAOptimizer Test 
>
> GetValue( 99.9999 -100) = 24000
...
> GetValue( 1.98369 -1.98687) = -9.99951
> GetValue( 1.98356 -1.98693) = -9.99951
> Solution        = (1.98362,-1.9869)
> StateOfConvergence in last iteration: 7.01606e-06
> NumberOfIterations: 54
> Stop condition: SPSAOptimizer: Below tolerance. Tolerance is 1e-05
> Test failed.

https://open.cdash.org/test/539433131

jhlegarreta:
@N-Dekker it looks like this change:
https://github.com/InsightSoftwareConsortium/ITK/pull/2869/files#diff-46b02345e1d860f304cc1ce37e97144bf95b06d6e8c6188d73956bac1bfa150cR60

is causing build failures for the following sites/builds:
Site: RogueResearch7
Build Name: Mac10.11-AppleClang-rel-x86_64-static 
Build Name: Mac10.11-AppleClang-dbg-x86_64-static 

e.g.
https://open.cdash.org/viewBuildError.php?buildid=7575447

N-Dekker:
> it looks like this change: https://github.com/InsightSoftwareConsortium/ITK/pull/2869/files#diff-46b02345e1d860f304cc1ce37e97144bf95b06d6e8c6188d73956bac1bfa150cR60
> 
> is causing build failures for the following sites/builds: 
> Site: RogueResearch7
> Build Name: Mac10.11-AppleClang-rel-x86_64-static
> Build Name: Mac10.11-AppleClang-dbg-x86_64-static
> 
> e.g. https://open.cdash.org/viewBuildError.php?buildid=7575447

Thanks @ jhlegarreta  I see now, it says things like:
```
/Users/builder/externalModules/IO/NIFTI/test/itkNiftiImageIOTest.cxx:26:1: error: redefinition of 'PreFillDirection'
PreFillDirection<1>()
^
[CTest: warning matched] /Users/builder/externalModules/IO/NIFTI/test/itkNiftiImageIOTest.h:64:1: note: previous definition is here
PreFillDirection<1>();
^
```

I never saw such an error before. Do you think it can be fixed by replacing the "deleted" (`= delete`) PreFillDirection function template by a dummy definition, like the following?

```
typename itk::ImageBase<TDimension>::DirectionType
PreFillDirection()
{
  itkGenericExceptionMacro("This template should never be used. Only valid values are given below.");
}
```

jhlegarreta:
> Do you think it can be fixed by replacing the "deleted" (= delete) PreFillDirection function template by a dummy definition, like the following?

Sorry Niels, I do not know of an elegant way yo circumvent this, and I do not have access to a macOS to be able to easily test other approaches.

Modules/IO/NIFTI/test/itkNiftiImageIOTest.h
```diff
@@ -54,16 +54,10 @@ constexpr unsigned char LEFT = 128;    /*Bit pattern 1 0 0  00000*/
...
 
 template <>
 itk::ImageBase<1>::DirectionType
```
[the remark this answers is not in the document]
N-Dekker: > Maybe if we have `typename` with every occurrence of `DirectionType`...

Thanks for the suggestion @Leengit But I believe `typename` just shouldn't be used on an explicit specialization of a function template. Because in that context, the types are no longer "dependent".

PR #2873, opened 2021-11-15
Use the faster `TransformPhysicalPointToIndex` overload

N-Dekker:
Replaced `image->TransformPhysicalPointToIndex(point, index)` calls by
`image->TransformPhysicalPointToIndex(point)`, which is slightly faster,
as it does not figure out whether or not the point is inside the image.

Follow-up to pull request #993
commit 0703516
"STYLE: Use image->TransformPhysicalPointToIndex(point), returning index"

### DIFF
--- a/Modules/Filtering/ImageFeature/test/itkHessianRecursiveGaussianFilterScaleSpaceTest.cxx
+++ b/Modules/Filtering/ImageFeature/test/itkHessianRecursiveGaussianFilterScaleSpaceTest.cxx
@@ -108,9 +108,7 @@ itkHessianRecursiveGaussianFilterScaleSpaceTest(int, char *[])
     PointType center;
     center.Fill(0.0);
 
-    IndexType centerIndex;
-
-    outputImage->TransformPhysicalPointToIndex(center, centerIndex);
+    IndexType centerIndex = outputImage->TransformPhysicalPointToIndex(center);
 
     // Irrespective of the scale, the Hxx component should be the same
     double centerHxx = outputImage->GetPixel(centerIndex)[0];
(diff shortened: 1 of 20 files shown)


Modules/Filtering/ImageFeature/test/itkHessianRecursiveGaussianFilterScaleSpaceTest.cxx
```diff
@@ -165,9 +163,7 @@ itkHessianRecursiveGaussianFilterScaleSpaceTest(int, char *[])
     PointType center;
     center.Fill(0.0);
 
-    IndexType centerIndex;
-
-    outputImage->TransformPhysicalPointToIndex(center, centerIndex);
+    IndexType centerIndex = outputImage->TransformPhysicalPointToIndex(center);
```
dzenanz, line 166: I am not sure how important performance is here, and other similar instances. But this improves style slightly, and saves a line of code.

dzenanz (approved):
The second is more of a style commit, than a performance one. Potentially split most of it into a separate style commit. Only calls inside large for loops (per pixel) count as real performance improvement!

hjmjohnson (approved):

N-Dekker:
> The second is more of a style commit, than a performance one. Potentially split most of it into a separate style commit. Only calls inside large for loops (per pixel) count as real performance improvement!

@dzenanz  I agree, it's mostly a style commit, it just _might_  improve the performance. 😄  The first commit might improve the performance of user code, as it is part of a filter.

Actually I was trying out the C++17 [[[nodiscard]]](https://en.cppreference.com/w/cpp/language/attributes/nodiscard) attribute on the older (potentially slower) overload, `ImageBase::TransformPhysicalPointToIndex(point, index)`. I think `[[nodiscard]]`  would make sense in this case.

Will adjust the commit message of the second commit, from PERF to STYLE.

N-Dekker:
I think this pull request is ready now, only CircleCi has failed, saying "Build timed out", and:
> Upgrade your pricing plan to take advantage of longer build times
> context deadline exceeded

https://app.circleci.com/pipelines/github/InsightSoftwareConsortium/ITK/3864/workflows/c431702d-7d38-400b-a113-07357a08f3b8/jobs/10620

dzenanz (approved):

PR #2876, opened 2021-11-16
BUG: fix filter progress overflowing due to bad logic in atomic add

### LINKED ISSUE #2874: Progress report overflows
### Description

Despite that it seems from looking at the code of 
void ProcessObject::IncrementProgress(float increment)
that it protects from overflow, it doesn't and it easily overflows, if, for example, we increment progress 11 times by 0.1.
But because of floating-point arithmetic, it also overflows if we increment progress 10 times by 0.1.
This is because m_Progress.fetch_add(integerIncrement) doesn't return updated value, but the value that was stored in m_Progress before fetch_add() (see the description of fetch_add()). But even if fetch_add() was implemented in such a way, so that it would return an updated value, the code still would be incorrect, because race condition between multiple threads may occur. 
### Steps to Reproduce

Here's the code that reproduces the problem:

```cpp
#include <itk/itkResampleImageFilter.h>
#include <gtest/gtest.h>

TEST(Progress, IncrementOverflow)
{
  auto filter = itk::ResampleImageFilter<itk::Image<double, 3>, itk::Image<double, 3>>::New();
  for (size_t i = 0; i < 11; ++i)
  {
    filter->IncrementProgress(0.1);
    std::cout << filter->GetProgress() << std::endl;
  }
}
```
### Expected behavior

The expected output of the above code would be:
```none
0.1
0.2
0.3
0.4
0.5
0.6
0.7
0.8
0.9
1.0
1.0
```
### Actual behavior

But the actual output is:
```none
0.1
0.2
0.3
0.4
0.5
0.6
0.7
0.8
0.9
1.25729e-08
0.1
```
As you can see, on the 10-th step it already overflowed (because of floating-point arithmetic), and on eleven step we get 0.1 instead of expected 1.0

### Reproducibility

100%.
### Versions

ITK 5.1.2 
But I looked at the latest master, and the code of IncrementProgress() is identical, so the issue is reproducible with the latest ITK.
### Environment

Windows 10

dzenanz:
fetch_add returns the value before the increment, not after.

Closes #2874.

### DIFF
--- a/Modules/Core/Common/src/itkProcessObject.cxx
+++ b/Modules/Core/Common/src/itkProcessObject.cxx
@@ -1157,9 +1157,9 @@ ProcessObject::IncrementProgress(float increment)
   // Clamp the value to be between 0 and 1.
   uint32_t integerIncrement = progressFloatToFixed(increment);
 
-  uint32_t oldProgress = m_Progress;
+  uint32_t oldProgress = m_Progress.fetch_add(integerIncrement);
 
-  uint32_t updatedProgress = m_Progress.fetch_add(integerIncrement);
+  uint32_t updatedProgress = m_Progress;
 
   // check if progress overflowed
   if (oldProgress > updatedProgress)
--- a/Modules/Core/Common/test/itkDataObjectAndProcessObjectTest.cxx
+++ b/Modules/Core/Common/test/itkDataObjectAndProcessObjectTest.cxx
@@ -171,6 +171,17 @@ itkDataObjectAndProcessObjectTest(int, char *[])
   process->UpdateProgress(0.0);
   ITK_TEST_SET_GET_VALUE(0.0, process->GetProgress());
 
+  // verify no progress overflow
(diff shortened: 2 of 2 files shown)


dzenanz:
I don't think [atomic_compare_exchange](https://en.cppreference.com/w/c/atomic/atomic_compare_exchange) is necessary here. But a more strict solution is welcome, should someone want to improve upon this.

hjmjohnson:
@dzenanz Could the test code in the issue be made into a formal test?

blowekamp:
> I don't think [atomic_compare_exchange](https://en.cppreference.com/w/c/atomic/atomic_compare_exchange) is necessary here. But a more strict solution is welcome, should someone want to improve upon this.

I think this is OK. When the function is exits the value should be clamped. However there is the possibility that while in the method another thread get's the invalid value.

blowekamp (approved):
Looks OK.

hjmjohnson (approved):

thewtex (approved):

PR #2878, opened 2021-11-16
DOC: Remove lingering Ultrasound dependency

tbirdso:
Follow up to #2866 removing lingering Ultrasound dependency for 1D FFT.

Unlike the classes addressed in #2866, `itk::ComplexToComplexProxy` is not wrapped for Python so the Doxygen group will only affect documentation and not load dependencies.

## PR Checklist


dzenanz (approved):

thewtex (approved):
:book: 

CircleCI false positive:

```
context deadline exceeded
```

PR #2879, opened 2021-11-16
COMP: No longer delete (`= delete`) PreFillDirection() function template

N-Dekker:
Pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2869 commit af32a779c4586304e52ddc27eb81656b949c9eeb ("COMP: Remove unreachable code") "deleted" the `PreFillDirection()` function template definition, by `= delete`. This appeared to cause compilation errors on the following sites/builds:

  Site: RogueResearch7
  Build Name: Mac10.11-AppleClang-rel-x86_64-static
  Build Name: Mac10.11-AppleClang-dbg-x86_64-static

As reported by Jon Haitz Legarreta Gorroño (@jhlegarreta) at https://github.com/InsightSoftwareConsortium/ITK/pull/2869#issuecomment-970306179

The error messages included:

    /Users/builder/externalModules/IO/NIFTI/test/itkNiftiImageIOTest.cxx:26:1: error: redefinition of 'PreFillDirection'
    PreFillDirection<1>()
    ^
    [CTest: warning matched] /Users/builder/externalModules/IO/NIFTI/test/itkNiftiImageIOTest.h:64:1: note: previous definition is here
    PreFillDirection<1>();
    ^

### DIFF
--- a/Modules/IO/NIFTI/test/itkNiftiImageIOTest.h
+++ b/Modules/IO/NIFTI/test/itkNiftiImageIOTest.h
@@ -54,23 +54,10 @@ constexpr unsigned char LEFT = 128;    /*Bit pattern 1 0 0  00000*/
 constexpr unsigned char ANTERIOR = 64; /*Bit pattern 0 1 0  00000*/
 constexpr unsigned char SUPERIOR = 32; /*Bit pattern 0 0 1  00000*/
 
-// The only specializations of this function template that may be called are given below.
+// Specializations of this function template are only implemented for 1D to 4D (defined in the cxx file).
 template <unsigned int TDimension>
 typename itk::ImageBase<TDimension>::DirectionType
-PreFillDirection() = delete;
-
-template <>
-itk::ImageBase<1>::DirectionType
-PreFillDirection<1>();
-template <>
-itk::ImageBase<2>::DirectionType
-PreFillDirection<2>();
-template <>
-itk::ImageBase<3>::DirectionType
-PreFillDirection<3>();
-template <>
-itk::ImageBase<4>::DirectionType
-PreFillDirection<4>();
+PreFillDirection();
 
 template <typename T>
 int


dzenanz (approved):

jhlegarreta (commented):
Can the exception be exercised in the corresponding test file, please?

N-Dekker:
I haven't been able to reproduce these compilation errors so far. But that doesn't mean that the errors are fake, of course  😄 

For what it's worth, I tried the following at https://godbolt.org/z/Yccnqnq4z

```
template <unsigned> struct Y{};

namespace itk
{
template <unsigned N> struct ImageBase
{
  using DirectionType = Y<N>;
};
}

typename itk::ImageBase<TDimension>::DirectionType
PreFillDirection() = delete;

template <>
itk::ImageBase<1>::DirectionType
PreFillDirection<1>();

template <>
itk::ImageBase<2>::DirectionType
PreFillDirection<2>();

template <>
itk::ImageBase<3>::DirectionType
PreFillDirection<3>();

template <>
itk::ImageBase<4>::DirectionType
PreFillDirection<4>();

template <>
itk::ImageBase<1>::DirectionType PreFillDirection<1>() { return {}; }
template <>
itk::ImageBase<2>::DirectionType PreFillDirection<2>() { return {}; }

int main()
{
    PreFillDirection<1>();
}
```

N-Dekker:
> Can the exception be exercised in the corresponding test file, please?

@jhlegarreta Thanks for your question. With the [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/48b950eff5d49690e0ebd61c21c14f5c1856c6de..173ae8bfa3815a7c1d9aaa01f4c26b7651e241a3) that I just did, the function template no longer has a function-body. So now it cannot be called, and it does not throw an exception anymore. Does that answer your question?

Note that `PreFillDirection()` is not part of the ITK API. It isn't even in the `itk` namespace! As far as I can see, it's just a little helper function template for a test, `itkNiftiImageIOTest3`.

Leengit (approved):
LGTM

jhlegarreta:
> @jhlegarreta Thanks for your question. With the force-push that I just did, the function template no longer has a function-body. So now it cannot be called, and it does not throw an exception anymore. Does that answer your question?

> Note that PreFillDirection() is not part of the ITK API. It isn't even in the itk namespace! As far as I can see, it's just a little helper function template for a test, itkNiftiImageIOTest3.

Reading [the related call](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/IO/NIFTI/test/itkNiftiImageIOTest.h#L292:L296), the accompanying explanation, and the use of the direction variable [here](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/IO/NIFTI/test/itkNiftiImageIOTest.h#L311), I am not that sure whether the behavior proposed is the expected one for the method.

dzenanz:
@seanm Can you check whether this fixes the compile error?

N-Dekker:
The old clang bug is reproducible at https://godbolt.org/z/W58nEWadM (as I wrote before), and the compiler error does indeed disappear when removing `= delete`, according this pull request  😃 

I wonder if it is necessary to still supports such an old compiler version (clang 3.8.1), but for now, this pull request does the job, apparently!

Modules/IO/NIFTI/test/itkNiftiImageIOTest.h
```diff
@@ -57,7 +57,7 @@ constexpr unsigned char SUPERIOR = 32; /*Bit pattern 0 0 1  00000*/
 // The only specializations of this function template that may be called are given below.
 template <unsigned int TDimension>
 typename itk::ImageBase<TDimension>::DirectionType
-PreFillDirection() = delete;
+PreFillDirection();
 
 template <>
```
Leengit, line 62: With the `= delete` removed above, it becomes an ordinary (template) declaration.  I am thinking that the four declarations of the template specializations that follow it do not do anything now and can be removed.  In particular, removal does not cause any problems when I modify your [example](https://godbolt.org/z/W58nEWadM) accordingly.

seanm:
Sorry, just seeing all this now.  Been a hectic few days.

> The old clang bug is reproducible at https://godbolt.org/z/W58nEWadM (as I wrote before), and the compiler error does indeed disappear when removing `= delete`, according this pull request 😃

I could try on the bot if you still want, but sounds unnecessary at this point...

> I wonder if it is necessary to still supports such an old compiler version (clang 3.8.1), but for now, this pull request does the job, apparently!

It's not that old, really. :) The Rogue7 bot is running OS X El Capitan, the last update of which was in July 2018. It's running the newest Xcode possible for that version of macOS: Xcode 8.2.1, released December 2016. That Xcode's compiler is AppleClang 8.0.

N-Dekker:
Thanks for your comment @seanm For the record: The Clang compilation error appears reported by jonathan.sauer, 2013-10-10: [Bug 17537 - Explicit specialization of deleted function template not possible](https://bugs.llvm.org/show_bug.cgi?id=17537) It appears fixed by Clang 3.9, as noted by Andrew Somerville, 2017-11-02.

Modules/IO/NIFTI/test/itkNiftiImageIOTest.h
N-Dekker: @Leengit It may be fine to remove the four declarations of the template specializations from my [example](https://godbolt.org/z/W58nEWadM) but I'm not sure if we can just remove them from the "itkNiftiImageIOTest.h", while having them defined in the corresponding cxx file, "itkNiftiImageIOTest.cxx", and being called from yet two other cxx files (itkNiftiImageIOTest3.cxx and itkNiftiImageIOTest7.cxx), which only "see" the header file, during compilation 🤷

seanm:
And according to the table here:

https://en.wikipedia.org/wiki/Xcode#Xcode_7.0_-_10.x_(since_Free_On-Device_Development)_2

AppleClang 8.0 is based on open source Clang 3.9 svn, but they probably branched before that bug was fixed.

dzenanz:
I was just about to merge this. I guess I can leave it up for another day or two until you figure out whether further modifications are desirable.

jhlegarreta (approved):

N-Dekker:
> I was just about to merge this. I guess I can leave it up for another day or two until you figure out whether further modifications are desirable.

Thanks @dzenanz. As far as I can see, it is sufficient to just remove `= delete` from the function template, in order to make the old (but not _that_ old) clang compiler happy.

However, it may be nice to still give the suggestion of @Leengit a try, from https://github.com/InsightSoftwareConsortium/ITK/pull/2879#discussion_r751658773 to also remove the declarations of the four explicit specializations. I'll try it out with my next force-push!

dzenanz:
@seanm, does this make your compiler happy? 😄

seanm:
I've confirmed this pacifies the Rogue7 bot.  Thanks all!

PR #2890, opened 2021-11-21
PERF: Use "magic statics" instead of Singleton in TimeStamp::Modified()

N-Dekker:
Avoids pointer indirection and read/write access to global data inside
this frequently called Core member function.

### DIFF
--- a/Modules/Core/Common/include/itkTimeStamp.h
+++ b/Modules/Core/Common/include/itkTimeStamp.h
@@ -31,7 +31,6 @@
 #include "itkMacro.h"
 #include "itkIntTypes.h"
 #include <atomic>
-#include "itkSingletonMacro.h"
 
 namespace itk
 {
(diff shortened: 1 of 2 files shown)


blowekamp:
The ITK singleton is needed to synchronize between ITK Python modules. This will not meet the requirements. @thewtex  Could provide further information.

N-Dekker:
> The ITK singleton is needed to synchronize between ITK Python modules. This will not meet the requirements. @thewtex Could provide further information.

Thanks @blowekamp Obviously this pull request doesn't work well so far, as https://open.cdash.org/test/546938925 says:

> assert met.GetMTime() < filt.GetMTime()

 I'm still trying to figure out how the ITK singleton appears to ensure that there is only one global "ModifiedTime" (`GlobalTimeStamp`), while just defining such a "ModifiedTime" as a `static` variable in a cxx file would not be good enough... 🤔  Any hint of clue, please?

blowekamp:
It has to do with ITK python building the ITK libraries static, but having each ITK module as a separate loadable (Shared library) python module. The there are multiple instances of these "global" variables on some system.

blowekamp:
> 
> I'm still trying to figure out how the ITK singleton appears to ensure that there is only one global "ModifiedTime" (`GlobalTimeStamp`), while just defining such a "ModifiedTime" as a `static` variable in a cxx file would not be good enough... 🤔 Any hint of clue, please?

There is some python code that is run to set the global singleton instance to the same object across the different libraries.

https://github.com/InsightSoftwareConsortium/ITK/blob/master/Wrapping/Generators/Python/CMakeLists.txt#L351

N-Dekker:
@blowekamp OK, I give it up for now! Thanks for your explanation!

thewtex:
Yes, to confirm, @blowekamp is right, the indirection here is required for synchronization across Python modules.

PR #2896, opened 2021-11-23
STYLE: Remove `TimeStamp` default-constructor, assignment (Rule of Zero)

N-Dekker:
Allowed the compiler to generate the default-constructor and assignment
operator of `TimeStamp` implicitly, followed C++ Core Guidelines,
August 19, 2021:

"If you can avoid defining default operations, do"
http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c20-if-you-can-avoid-defining-default-operations-do

Added compile-time checks of the type traits of the special member
functions of `TimeStamp`.

### DIFF
--- a/Modules/Core/Common/test/itkTimeStampTest.cxx
+++ b/Modules/Core/Common/test/itkTimeStampTest.cxx
@@ -16,10 +16,16 @@
  *
  *=========================================================================*/
 
-#include <iostream>
 #include "itkTimeStamp.h"
 #include "itkMultiThreaderBase.h"
 
+#include <iostream>
+#include <type_traits>
+
+static_assert(std::is_nothrow_default_constructible<itk::TimeStamp>::value, "Check TimeStamp default-constructibility");
+static_assert(std::is_trivially_copy_constructible<itk::TimeStamp>::value, "Check TimeStamp copy-constructibility");
+static_assert(std::is_trivially_copy_assignable<itk::TimeStamp>::value, "Check TimeStamp copy-assignability");
+static_assert(std::is_trivially_destructible<itk::TimeStamp>::value, "Check TimeStamp destructibility");
 
 // A helper struct for the test, the idea is to have one timestamp per thread.
 // To ease the writing of the test, we use  MultiThreaderBase::SingleMethodExecute
--- a/Modules/Core/Common/include/itkTimeStamp.h
+++ b/Modules/Core/Common/include/itkTimeStamp.h
@@ -70,10 +70,6 @@ public:
   static Self *
   New();
 
-  /** Constructor must remain public because classes instantiate
-   * TimeStamps implicitly in their construction.  */
-  TimeStamp() { m_ModifiedTime = 0; }
-
   /** Destoy this instance. */
   void
   Delete()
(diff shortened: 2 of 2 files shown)


Modules/Core/Common/test/itkTimeStampTest.cxx
```diff
@@ -16,10 +16,17 @@
...
 #include "itkMultiThreaderBase.h"
 
+#include <iostream>
+#include <type_traits>
+
+static_assert(std::is_nothrow_default_constructible<itk::TimeStamp>::value &&
+                std::is_trivially_copy_constructible<itk::TimeStamp>::value &&
+                std::is_trivially_copy_assignable<itk::TimeStamp>::value &&
+                std::is_trivially_destructible<itk::TimeStamp>::value,
+              "Check the special member functions of itk::TimeStamp.");
```
blowekamp, line 29: I'd prefer that these were separate asserts so that if they fail we will know which one fails.

blowekamp:
Nice clean up 👍

dzenanz (approved):

Modules/Core/Common/test/itkTimeStampTest.cxx
N-Dekker: > I'd prefer that these were separate asserts

@blowekamp  I agree, but I find it a bit cumbersome to write a separate message text for each of the static_assert's. 🤔 

    static_assert(std::is_nothrow_default_constructible<itk::TimeStamp>::value, "Check default-constructibility of TimeStamp");
    static_assert(std::is_trivially_copy_constructible<itk::TimeStamp>::value,
      "Check copy-constructibility of TimeStamp");
    static_assert(std::is_trivially_copy_assignable<itk::TimeStamp>::value,
      "Check copy-assignability of TimeStamp");
    static_assert(std::is_trivially_copy_assignable<itk::TimeStamp>::value,
      "Check copy-assignability of TimeStamp");
    static_assert(std::is_trivially_destructible<itk::TimeStamp>::value,
      "Check destructibility of TimeStamp");

Would you like it like that?

blowekamp: Yea, that is fine. It'd be easier if the C++17 signature could be used without the require text messaged 😁

N-Dekker: Maybe we could introduce an `itkStaticAssertMacro(condition)` that accepts a single argument  😃

N-Dekker: @blowekamp Please check the [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/700a62e995a7b9542ade2145728f7662ee3783ca..084f10d8a0723186cbbbcbf7e70809c2ddd8f6ee) I just did  😃

dzenanz: > Maybe we could introduce an itkStaticAssertMacro(condition) that accepts a single argument

Hopefully we will move to C++17 soon-ish 😄 so that will not be needed.

N-Dekker: @blowekamp @dzenanz I certainly wouldn't mind moving to C++17  😄

blowekamp (approved):
Thank you.

N-Dekker:
The CircleCI failure appears irrelevant to this pull request, as it says at [app.circleci.com](https://app.circleci.com/pipelines/github/InsightSoftwareConsortium/ITK/3919/workflows/25ba8731-da73-4215-9817-5444260af037/jobs/10675) :
> Build timed out
> Upgrade your pricing plan to take advantage of longer build times
> context deadline exceeded

All the other CI checks are performed successfully, so I think this pull request is ready.

dzenanz (approved):

PR #2897, opened 2021-11-23
STYLE: Use `unique_ptr` for `ImageToImageMetric` data members

N-Dekker:
Following C++ Core Guidelines, August 19, 2021:
"Use unique_ptr or shared_ptr to avoid forgetting to delete objects created using new"
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-smart

Defaulted (`= default`) its destructor.


dzenanz (approved):
Code looks good. Hopefully it works too 😄

dzenanz (approved):

PR #2900, opened 2021-11-24
ENH: add a new ContinuousIntegration build: LinuxLegacyRemoved

dzenanz:
It has happened several times that ITK fails to build
on one of the dashboard machines which has ITK_LEGACY_REMOVE=ON.
This could be avoided if we had a CI build without legacy.

## PR Checklist

### DIFF
--- a/Testing/ContinuousIntegration/AzurePipelinesLinux.yml
+++ b/Testing/ContinuousIntegration/AzurePipelinesLinux.yml
@@ -80,3 +80,78 @@ jobs:
         testRunTitle: 'CTest $(Agent.OS)'
       condition: succeededOrFailed()
       displayName: 'Publish test results'
+
+- job: LinuxLegacyRemoved
+  timeoutInMinutes: 0
+  cancelTimeoutInMinutes: 300
+  pool:
+    vmImage: ubuntu-18.04
+  steps:
+    - checkout: self
+      clean: true
+      fetchDepth: 5
+    - bash: |
+        set -x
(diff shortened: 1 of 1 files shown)


seanm:
Nice, thanks. That should keep my bots happier too.

Testing/ContinuousIntegration/AzurePipelinesLinuxLegacyRemoved.yml
```diff
@@ -0,0 +1,83 @@
+name: ITK.Linux
```
jhlegarreta, line 1: The name should be different so that the build can be distinguished from the regular `ITK.Linux`, maybe `ITK.Linux.LegacyRemove` or `ITK.Linux.LegacyRemoved` would honor its purpose.

dzenanz: Complete oversight on my part.

dzenanz:
This now seems to work fine. The distinction on the dashboard is `LegacyRemoved` suffix in the build name, e.g. [Linux-Build7081-PR2900-legacyRemovedCI**LegacyRemoved**](https://open.cdash.org/build/7601470) vs [Linux-Build7081-PR2900-legacyRemovedCI](https://open.cdash.org/build/7601462). In this case a bit non-obvious because this PR's branch name is `legacyRemovedCI` which is also part of the build name.

hjmjohnson (approved):

jhlegarreta (approved):
Looks good. Thanks for doing this @dzenanz !

PR #2155, opened 2020-12-07
COMP: Fix invalid wrap combinations

hjmjohnson:
Resolves #2152 

## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [X] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [X] Added test (or behavior not changed)
- [X] REMOVED  Python wrapping to INVALID COMBINATIONS

Issue #2152, opened 2020-12-05
IPO Python wrapping compiler warnings


hjmjohnson:
### Description

Recent compilation option changes (IPO turn on I think) now expose some compiler warnings.

### Steps to Reproduce

Build with cmake 3.18, python on Ubutu 18.04

### Expected behavior

Clean build with no warnings

### Actual behavior
See warnings at bottom issue in the "Additional Information section"

### Reproducibility

100% of time on Ubuntu 18.04 with cmake 3.18 (and IPO ?).

### Versions
ITK master branch as of 2020-12-05

### Environment
Ubuntu 18.04

/usr/bin/c++ --version 
c++ (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

cmake --version 
cmake version 3.18.4

### Additional Information
```
[377/649] Linking CXX shared module Wrapping/Generators/Python/itk/_ITKEigenPython.so
/home/johnsonhj/Dashboard/src/ITK/Modules/Numerics/Eigen/include/itkEigenAnalysis2DImageFilter.hxx: In member function ‘GenerateData’:
/home/johnsonhj/Dashboard/src/ITK/Modules/Core/Common/include/itkVector.hxx:109:25: warning: ‘MEM[(const ValueType &)&eigenVector].m_InternalArray[2]’ may be used uninitialized in this function [-Wmaybe-uninitialized]
     const RealValueType value = (*this)[i];
                         ^
/home/johnsonhj/Dashboard/src/ITK/Modules/Numerics/Eigen/include/itkEigenAnalysis2DImageFilter.hxx: In member function ‘GenerateData’:
/home/johnsonhj/Dashboard/src/ITK/Modules/Core/Common/include/itkVector.hxx:109:25: warning: ‘MEM[(const ValueType &)&eigenVector].m_InternalArray[2]’ may be used uninitialized in this function [-Wmaybe-uninitialized]
     const RealValueType value = (*this)[i];
                         ^
/home/johnsonhj/Dashboard/src/ITK/Modules/Core/Common/include/itkVector.hxx:109:25: warning: ‘MEM[(const ValueType &)&eigenVector].m_InternalArray[3]’ may be used uninitialized in this function [-Wmaybe-uninitialized]
[...]

PR #2902, opened 2021-11-25
COMP: Initialize variable to remove warnings

PranjalSahu:
Initializing variables v0, v1, v2 with 0 to remove variable may not be initialized warnings.
Also replacing usage of GetGaussCurvatureData with GetModifiableGaussCurvatureData in tests.

## PR Checklist

### DIFF
--- a/Modules/Core/Mesh/include/itkTriangleMeshCurvatureCalculator.hxx
+++ b/Modules/Core/Mesh/include/itkTriangleMeshCurvatureCalculator.hxx
@@ -85,6 +85,7 @@ template <typename TInputMesh>
 void
 TriangleMeshCurvatureCalculator<TInputMesh>::ComputeGaussCurvature(MeshConstPointer inputMesh)
 {
+  MeshPointType v0, v1, v2;
   MeshPointType e0, e1, e2;
   double        A, alpha0, alpha1, alpha2;
 
@@ -99,6 +100,10 @@ TriangleMeshCurvatureCalculator<TInputMesh>::ComputeGaussCurvature(MeshConstPoin
     dA[k] = 0.0;
   }
 
+  v0.Fill(0);
+  v1.Fill(0);
+  v2.Fill(0);
+
   CellsContainerConstPointer  outCells = inputMesh->GetCells();
   CellsContainerConstIterator cellsItr = outCells->Begin();
 
@@ -107,9 +112,9 @@ TriangleMeshCurvatureCalculator<TInputMesh>::ComputeGaussCurvature(MeshConstPoin
     CellType *               cellPointer = cellsItr.Value();
     MeshPointIdConstIterator point_ids = cellPointer->GetPointIds();
 
-    MeshPointType v0 = inputMesh->GetPoint(point_ids[0]);
-    MeshPointType v1 = inputMesh->GetPoint(point_ids[1]);
-    MeshPointType v2 = inputMesh->GetPoint(point_ids[2]);
+    v0 = inputMesh->GetPoint(point_ids[0]);
+    v1 = inputMesh->GetPoint(point_ids[1]);
+    v2 = inputMesh->GetPoint(point_ids[2]);
 
     // Edges
     e0[0] = v1[0];


N-Dekker:
Thanks you for your contribution @PranjalSahu 

Sorry for the nitpick, but please try to keep the first like of your commit message at 72 characters, at most. Otherwise Github will chop it off, as in "COMP: Use GetModifiableObjectMacro and initialize variable to remove …".

One more comment inline, please wait!

Modules/Core/Mesh/include/itkTriangleMeshCurvatureCalculator.hxx
```diff
@@ -107,9 +112,9 @@ TriangleMeshCurvatureCalculator<TInputMesh>::ComputeGaussCurvature(MeshConstPoin
     CellType *               cellPointer = cellsItr.Value();
     MeshPointIdConstIterator point_ids = cellPointer->GetPointIds();
 
-    MeshPointType v0 = inputMesh->GetPoint(point_ids[0]);
-    MeshPointType v1 = inputMesh->GetPoint(point_ids[1]);
-    MeshPointType v2 = inputMesh->GetPoint(point_ids[2]);
+    v0 = inputMesh->GetPoint(point_ids[0]);
+    v1 = inputMesh->GetPoint(point_ids[1]);
+    v2 = inputMesh->GetPoint(point_ids[2]);
```
N-Dekker, line 117: Did the original code have any warning? As far as I can see, it did already properly initialize v0, v1, and v2! So then, why should it be changed, in your opinion?

Modules/Core/Mesh/include/itkTriangleMeshCurvatureCalculator.hxx
```diff
@@ -85,6 +85,7 @@ template <typename TInputMesh>
 void
 TriangleMeshCurvatureCalculator<TInputMesh>::ComputeGaussCurvature(MeshConstPointer inputMesh)
 {
+  MeshPointType v0, v1, v2;
```
N-Dekker, line 88: Instead of calling `Fill(0)` for each of these three variables, they may be initialized by an empty initializer-list, `{}`:

    MeshPointType v0{};
    MeshPointType v1{};
    MeshPointType v2{};

Modules/Core/Mesh/include/itkTriangleMeshCurvatureCalculator.hxx
PranjalSahu: When I build original code with Python Wrapping I get following warnings:

```
Linking CXX shared module Wrapping/Generators/Python/itk/_ITKMeshPython.so
../Modules/Core/Mesh/include/itkTriangleMeshCurvatureCalculator.hxx: In member function ‘Compute’:
../Modules/Core/Mesh/include/itkTriangleMeshCurvatureCalculator.hxx:127:11: warning: ‘MEM[(ValueType &)&v1 + 8]’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  127 |     e0[2] -= v0[2];
      |           ^
../Modules/Core/Mesh/include/itkTriangleMeshCurvatureCalculator.hxx:88:21: note: ‘MEM[(ValueType &)&v1 + 8]’ was declared here
   88 |   MeshPointType v0, v1, v2;
      |                     ^
../Modules/Core/Mesh/include/itkTriangleMeshCurvatureCalculator.hxx:127:11: warning: ‘MEM[(ValueType &)&v0 + 8]’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  127 |     e0[2] -= v0[2];
      |           ^
../Modules/Core/Mesh/include/itkTriangleMeshCurvatureCalculator.hxx:88:17: note: ‘MEM[(ValueType &)&v0 + 8]’ was declared here
   88 |   MeshPointType v0, v1, v2;
   ```

However, no such warning is thrown when only building C++ code. The GetPoint method definition from PointSet class is shown below:

```
PointType point;
bool exist = m_PointsContainer->GetElementIfIndexExists(ptId, &point);
if (!exist)
{
    itkExceptionMacro("Point id doesn't exist: " << ptId);
}
return point;
```

So it takes care of the cases when the id does not exists by throwing exception but I don't know how this behaves with Python Wrapping.

Comment (N-Dekker) [reply Modules/Core/Mesh/include/itkTriangleMeshCurvatureCalculator.hxx:117]: @PranjalSahu  Interesting, thanks for your explanation! I _think_ the warning is a false alarm (although of course, we do not want to have warnings anyway).  I _think_  the compiler suspects that, because `GetElementIfIndexExists(ptId, &point)` might not assign anything to `point`,  the `return` statement might be executed while `point` is still uninitialized. Right?

File: Modules/Core/Mesh/test/itkTriangleMeshCurvatureCalculatorTest.cxx [hunk 3]
```diff
@@ -95,7 +95,7 @@ itkTriangleMeshCurvatureCalculatorTest(int argc, char * argv[])
   curvCalculator->SetCurvatureTypeToGaussian();
   curvCalculator->Compute();
 
-  gaussCurvatureData = curvCalculator->GetGaussCurvatureData();
+  gaussCurvatureData = curvCalculator->GetModifiableGaussCurvatureData();
```
N-Dekker, line 98: @PranjalSahu Please consider having two separate commits: one for the `Wmaybe-uninitialized` warning, and one for the `GetModifiable` calls. (Following the rule "One Commit. One Change".)

Modules/Core/Mesh/include/itkTriangleMeshCurvatureCalculator.hxx
PranjalSahu: Will test and reply here.

Modules/Core/Mesh/test/itkTriangleMeshCurvatureCalculatorTest.cxx
PranjalSahu: Yes will make two separate PRs.

Modules/Core/Mesh/include/itkTriangleMeshCurvatureCalculator.hxx
PranjalSahu: I have tested the change you suggested and it didn't work. And the change that I did also does not work. 

I also did one more test. At line 194 in [GetPoint(PointIdentifier ptId)](https://github.com/InsightSoftwareConsortium/ITK/blob/v5.3rc02/Modules/Core/Common/include/itkPointSet.hxx#L194) there is another exception for absence of m_PointsContainer. I commented all itkExceptionMacro and the warning went away. The problem is occurring wherever the return is itkExceptionMacro, which probably translates to null in Python. I believe it will need a major change to handle such warnings while doing Python wrapping.

N-Dekker: @PranjalSahu Hmmm, weird!

> The problem is occurring wherever the return is itkExceptionMacro, which probably translates to null in Python

I don't know that much about the Python wrapping, but a call to itkExceptionMacro is unconditionally translated to something like:
```
  {
    std::ostringstream exceptionDescriptionOutputStringStream; 
    exceptionDescriptionOutputStringStream << "ITK ERROR: " x;
    throw ::itk::ExceptionObject(
      std::string{ __FILE__ }, __LINE__, exceptionDescriptionOutputStringStream.str(), std::string{ITK_LOCATION });
  }
```

https://github.com/InsightSoftwareConsortium/ITK/blob/v5.3rc02/Modules/Core/Common/include/itkMacro.h#L525-L541

So really I have to admit I don't entirely understand what's going on!

Could you possibly try the following as well, as the last few lines of `PointSet::GetPoint(PointIdentifier ptId)`, instead of the adjustment I previously suggested? (Directly after the comment which says "Ask the container if the point identifier exists.")

```
  if (!m_PointsContainer->IndexExists(ptId))
  {
    itkExceptionMacro("Point id doesn't exist: " << ptId);
  }
  return (*m_PointsContainer)[static_cast<std::size_t>(ptId)];
```

You never know, maybe the compiler gets happy when there is an unconditional return statement at the end!

PranjalSahu: ok will test and reply here.

PranjalSahu: No this also does not work.

N-Dekker:
@PranjalSahu Those warnings you encountered look similar to issue  #2152 "IPO Python wrapping compiler warnings", which appears to be fixed by pull request #2155 "COMP: Fix invalid wrap combinations", by Hans (@hjmjohnson). Would a similar fix be applicable here?

PranjalSahu:
I think I got the problem. The code is meant to be run for 3D points so it is throwing warning for 2D wrap combinations (the 2nd index is going to be invalid there). I am checking it by compiling it only for 3D.

PranjalSahu:
Yes, it worked. I removed the wrapping for 2D and the warnings went away. Thanks for pointing to that PR :).

PR #2905, opened 2021-11-26
STYLE: Use keyword style for HeaderTest target_link_libraries

thewtex:
Use the more modern explicit keyword-style.

### DIFF
--- a/CMake/ITKModuleHeaderTest.cmake
+++ b/CMake/ITKModuleHeaderTest.cmake
@@ -85,7 +85,7 @@ macro( itk_module_headertest _name )
         ${_test_num}
         )
       add_executable( ${_test_name} ${_header_test_src} )
-      target_link_libraries( ${_test_name} ${${_name}_LIBRARIES} itksys )
+      target_link_libraries( ${_test_name} PUBLIC ${${_name}_LIBRARIES} itksys )
 
       add_dependencies(${_name}-all ${_test_name})
       math( EXPR _test_num "${_test_num} + 1" )


CMake/ITKModuleHeaderTest.cmake
```diff
@@ -85,7 +85,7 @@ macro( itk_module_headertest _name )
         ${_test_num}
         )
       add_executable( ${_test_name} ${_header_test_src} )
-      target_link_libraries( ${_test_name} ${${_name}_LIBRARIES} itksys )
+      target_link_libraries( ${_test_name} LINK_PUBLIC ${${_name}_LIBRARIES} itksys )
```
N-Dekker, line 88: @thewtex Interesting! But are you sure that it is more modern to use `LINK_PUBLIC`? https://cmake.org/cmake/help/latest/command/target_link_libraries.html says:

> Libraries for a Target and/or its Dependents (Legacy)
>
```
 target_link_libraries(<target>
                      <LINK_PRIVATE|LINK_PUBLIC> <lib>...
                     [<LINK_PRIVATE|LINK_PUBLIC> <lib>...]...)

```
>
> The LINK_PUBLIC and LINK_PRIVATE modes can be used to specify both the link dependencies and the link interface in one command.
>
> This signature is for compatibility only. Prefer the PUBLIC or PRIVATE keywords instead.

N-Dekker: Would `PRIVATE` also work well in this case? If so, I guess `PRIVATE` would be preferable, right?

thewtex: > This signature is for compatibility only. Prefer the PUBLIC or PRIVATE keywords instead.

I will change it from `LINK_PUBLIC` to `PUBLIC`, but the change is for a keyword vs no keyword at all.

> Would PRIVATE also work well in this case? If so, I guess PRIVATE would be preferable, right?

Since it is an executable, it does not matter.

N-Dekker: Ah, thanks Matt! I see now: the choice between `PRIVATE` and `PUBLIC` has no effect at all on the generated vcxproj file, for an executable. So indeed, it does not matter!

N-Dekker (approved):
Approved! Thanks again, Matt!

PR #2906, opened 2021-11-27
COMP: Read Only access to GaussCurvatureData

PranjalSahu:
To remove legacy code "itkGetObjectMacro should be replaced with itkGetModifiableObjectMacro or itkGetConstObjectMacro".

## PR Checklist

### DIFF
--- a/Modules/Core/Mesh/include/itkTriangleMeshCurvatureCalculator.h
+++ b/Modules/Core/Mesh/include/itkTriangleMeshCurvatureCalculator.h
@@ -107,7 +107,7 @@ public:
   itkSetObjectMacro(TriangleMesh, InputMeshType);
 
   /** Get Gauss curvature. */
-  itkGetModifiableObjectMacro(GaussCurvatureData, DoubleVectorContainer);
+  itkGetConstObjectMacro(GaussCurvatureData, DoubleVectorContainer);
 
   /** Set the curvature type to Guass. */
   void
--- a/Modules/Core/Mesh/test/itkTriangleMeshCurvatureCalculatorTest.cxx
+++ b/Modules/Core/Mesh/test/itkTriangleMeshCurvatureCalculatorTest.cxx
@@ -41,8 +41,6 @@ itkTriangleMeshCurvatureCalculatorTest(int argc, char * argv[])
   // Exercise basic object methods
   ITK_EXERCISE_BASIC_OBJECT_METHODS(curvCalculator, TriangleMeshCurvatureCalculator, Object);
 
-  using DoubleVectorContainerPointer = typename CurvatureCalculatorType::DoubleVectorContainerPointer;
-
   // Test for empty mesh, curvature data should be null.
   TriangleMeshType::Pointer triangleMesh = TriangleMeshType::New();
 
@@ -68,7 +66,7 @@ itkTriangleMeshCurvatureCalculatorTest(int argc, char * argv[])
   curvCalculator->Compute();
 
   // Output should be null for empty mesh.
-  DoubleVectorContainerPointer gaussCurvatureData = curvCalculator->GetModifiableGaussCurvatureData();
+  auto gaussCurvatureData = curvCalculator->GetGaussCurvatureData();
   ITK_TEST_EXPECT_TRUE(gaussCurvatureData == nullptr);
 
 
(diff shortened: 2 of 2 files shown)


Modules/Core/Mesh/test/itkTriangleMeshCurvatureCalculatorTest.cxx
```diff
@@ -95,7 +95,7 @@ itkTriangleMeshCurvatureCalculatorTest(int argc, char * argv[])
   curvCalculator->SetCurvatureTypeToGaussian();
   curvCalculator->Compute();
 
-  gaussCurvatureData = curvCalculator->GetGaussCurvatureData();
+  gaussCurvatureData = curvCalculator->GetModifiableGaussCurvatureData();
```
N-Dekker, line 98: @PranjalSahu Is it really necessary to get "modifiable access" to the curvature data via this `gaussCurvatureData` pointer?  If read-only access is sufficient, I think it would be preferable to keep calling the old `GetGaussCurvatureData()`. And then, just declare `gaussCurvatureData` as (smart) pointer-to-const.

N-Dekker: The following declaration of `gaussCurvatureData` compiles for me:

    using DoubleVectorContainer = typename CurvatureCalculatorType::DoubleVectorContainer;
    itk::SmartPointer<const DoubleVectorContainer> gaussCurvatureData = curvCalculator->GetGaussCurvatureData();


Or simply:

    auto gaussCurvatureData = curvCalculator->GetGaussCurvatureData();

When doing so (either way), it seems to me that the GetModifiable calls are no longer necessary!

PranjalSahu: Yes, the intention it to only support Read access. Will do the suggested changes and test it.

PranjalSahu: itkGetConstObjectMacro is used for access to Gauss curvature data and I have successfully compile it with ITK_FUTURE_LEGACY_REMOVE flag.

N-Dekker: Thanks for addressing my comments so carefully @PranjalSahu

Modules/Core/Mesh/include/itkTriangleMeshCurvatureCalculator.h
```diff
@@ -107,7 +107,7 @@ class ITK_TEMPLATE_EXPORT TriangleMeshCurvatureCalculator : public Object
   itkSetObjectMacro(TriangleMesh, InputMeshType);
 
   /** Get Gauss curvature. */
-  itkGetModifiableObjectMacro(GaussCurvatureData, DoubleVectorContainer);
+  itkGetConstObjectMacro(GaussCurvatureData, DoubleVectorContainer);
```
N-Dekker, line 110: If it all works fine with just a `itkGetConstObjectMacro` call here, it's fine to me as well. Just want to mention that `itkGetModifiableObjectMacro` defines _both_ a "modifiable" and a `const` getter, even when `ITK_FUTURE_LEGACY_REMOVE` is defined:

https://github.com/InsightSoftwareConsortium/ITK/blob/v5.3rc02/Modules/Core/Common/include/itkMacro.h#L1131-L1133 

So strictly speaking, you should still be able to have "const" access, by calling `GetGaussCurvatureData()`, even when `itkGetModifiableObjectMacro`  is being used. (Please correct me if I'm wrong.)

Having said so, `itkGetConstObjectMacro` is fine to me as well.

PranjalSahu: Yes, you are right. Both the access methods will be defined. But in the master code in the first Get call, we have called GetModifiableGaussCurvatureData and then later called GetGaussCurvatureData. And the compiler is throwing error because the pointer returned is of different types.

```
candidate function not viable: no known conversion from 'const
itk::TriangleMeshCurvatureCalculator<itk::Mesh<double, 3>>::DoubleVectorContainer *' 
(aka 'const VectorContainer<unsigned long, double> *') to 
'itk::SmartPointer<itk::VectorContainer<unsigned long, double>>' for 
1st argument operator=(SmartPointer r) noexcept
```

But we only wanted to give read access to the curvature data so itkGetConstObjectMacro is the right choice.

N-Dekker: @PranjalSahu I think I agree with you, but just to be sure: Is there a reasonable use case, retrieving a pointer to the curvature data from the calculator, and then wanting to modify that data "in-place" (rather than making a copy)?  

If such a use case doesn't seem to be there at the moment, your PR is fine to me.  As long as it passes all CI checks  😃

PranjalSahu:
Yes I don’t see such use case right now.
The obtained curvature data can be written to point data container of a
mesh by the user. Also the point data container can have custom data type
so we let the user decide how to use the obtained curvature data.

N-Dekker (approved):
@PranjalSahu Thanks you for addressing my comment! Approved!

dzenanz (approved):

PR #2907, opened 2021-11-27
Move `BSplineInterpolateImageFunction` member function definitions into class, use `unique_ptr`

N-Dekker:
Moved the `threadId` specific member function definitions of `BSplineInterpolateImageFunction` into its class template definition.

Used `unique_ptr` for threaded `BSplineInterpolateImageFunction` data

### DIFF
--- a/Modules/Core/ImageFunction/include/itkBSplineInterpolateImageFunction.h
+++ b/Modules/Core/ImageFunction/include/itkBSplineInterpolateImageFunction.h
@@ -28,14 +28,15 @@
 #ifndef itkBSplineInterpolateImageFunction_h
 #define itkBSplineInterpolateImageFunction_h
 
-#include <vector>
-
 #include "itkInterpolateImageFunction.h"
 #include "vnl/vnl_matrix.h"
 
 #include "itkBSplineDecompositionImageFilter.h"
 #include "itkConceptChecking.h"
 #include "itkCovariantVector.h"
+
+#include <memory> // For unique_ptr.
+#include <vector>
 
 namespace itk
 {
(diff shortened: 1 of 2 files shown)


Modules/Core/ImageFunction/include/itkBSplineInterpolateImageFunction.h
```diff
@@ -28,15 +28,16 @@
...
-#include <vector>
-
 #include "itkInterpolateImageFunction.h"
 #include "vnl/vnl_matrix.h"
 
 #include "itkBSplineDecompositionImageFilter.h"
 #include "itkConceptChecking.h"
 #include "itkCovariantVector.h"
 
+#include <memory> // For unique_ptr.
```
blowekamp, line 38: This comment is not necessary and should be removed. Imagine if every include did this. And then we had to maintain the accuracy of the comment.

N-Dekker: @blowekamp Is it acceptable to you to leave it this way for now, or is it a "show stopper" to you?

Personally I do find it helpful to comment why a certain `#include` is there, in those cases where it isn't entirely obvious.

I certainly would not do:

    // Obvious - not what I want!
    #include <any>    // For any.
    #include <tuple>  // For tuple.
    #include <vector> // For vector.

But for example, the following appears useful to me:

    // Useful comments, in my opinion:
    #include <memory>  // For unique_ptr.
    #include <numeric> // For inner_product.
    #include <utility> // For move.

dzenanz (approved):

PR #2910, opened 2021-11-30
ENH: Raw Pointer for Python Wrapping

PranjalSahu:
Minor changes for better code readability and for optimized Python wrapping.
Using Raw Pointer instead of Smart Pointer.

## PR Checklist


PranjalSahu:
@thewtex I didn't do the change to convert K and dA variables to Double vector container as in the loop we are accessing values at various indexes  in non-serial order and using GetElement/SetElement is going to be inefficient in such case (function call stack). Please let me know if there is better alternative.

PranjalSahu:
Should I create separate commits for these changes ?

dzenanz:
You can. Maybe into 3? "Use raw pointer", "Add testing", "Fix wrapping", not necessarily in this order.

Modules/Core/Mesh/include/itkTriangleMeshCurvatureCalculator.h
```diff
@@ -31,6 +31,7 @@
 
 #include <iostream>
 #include "itkMesh.h"
+#include "itkTriangleCell.h"
```
dzenanz, line 34: When you have an "**and**" in commit message, it is a good indication it should be split into two commit. There is no limit how small commits can be. I do realize this is nitpicking, so feel free to ignore this time.

dzenanz (approved):

Modules/Core/Mesh/include/itkTriangleMeshCurvatureCalculator.h
```diff
@@ -129,15 +131,12 @@ class ITK_TEMPLATE_EXPORT TriangleMeshCurvatureCalculator : public Object
 
   /** Discrete Gauss curvature (K) computation */
   void
-  GetGaussCurvature(MeshConstPointer input);
```
thewtex, line 132: We want to remove the smart pointer from the interface here.

Modules/Core/Mesh/include/itkTriangleMeshCurvatureCalculator.h
```diff
@@ -129,15 +131,12 @@ class ITK_TEMPLATE_EXPORT TriangleMeshCurvatureCalculator : public Object
...
-
-  void
-  ComputeGaussCurvature(MeshConstPointer input);
+  ComputeGaussCurvature(const InputMeshType * input);
 
 private:
   CurvaturesEnum               m_CurvatureType = CurvaturesEnum::GaussCurvature;
   DoubleVectorContainerPointer m_GaussCurvatureData;
-  MeshConstPointer             m_TriangleMesh;
```
thewtex, line 140: This still needs to be a smart pointer (we want to hold a reference to the object).

thewtex (changes requested):
@PranjalSahu thanks! Follow-up noted inline.

dzenanz (approved):

jhlegarreta:
@PranjalSahu one of these two commits causes a memory leak 59d95db 4aab01e:
https://open.cdash.org/viewDynamicAnalysis.php?buildid=7611627

PranjalSahu:
I will check it. It is probably due to the newly added check for Triangle
Cell. I have to see whether the memory allocated for K and dA are properly
released after the itkExceptionMacro is called.

PranjalSahu:
I think it is due to the TetrahedronCell inserted in the mesh. 
The documentation for [SetCell](https://itk.org/Doxygen/html/classitk_1_1Mesh.html#af48765961f72f75c8d2c6a1d12cd0124) in itk mesh says:

**Access routines to fill the Cells container (m_CellsContainer), and get information from it. If SetCell is used to overwrite a cell currently in the mesh, it is the caller's responsibility to release the memory for the cell currently at the CellIdentifier position prior to calling SetCell.**

So we need to the release the memory of the original cell present in the mesh before setting the new TetrahedronCell.
Testing it and will update here.

PR #2911, opened 2021-12-01
ENH: Increase `itk::ImageBase` coverage

jhlegarreta:
Increase `itk::ImageBase` coverage: test exceptions.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/test/itkImageBaseGTest.cxx
+++ b/Modules/Core/Common/test/itkImageBaseGTest.cxx
@@ -119,6 +119,44 @@ Check_New_ImageBase()
   EXPECT_TRUE(imageBase->GetInverseDirection().GetVnlMatrix().is_identity());
 }
 
+template <unsigned int VImageDimension>
+void
+CheckInvalidSpacingExceptions()
+{
+  using SpacingType = typename itk::ImageBase<VImageDimension>::SpacingType;
+  using DirectionType = typename itk::ImageBase<VImageDimension>::DirectionType;
+
+  const auto imageBase = itk::ImageBase<VImageDimension>::New();
+
+  // Test exceptions
+  const SpacingType initialSpacing = imageBase->GetSpacing();
+  const SpacingType negativeSpacing(-1.0);
+  const SpacingType zeroSpacing(0.0);
+
+#if !defined(ITK_LEGACY_REMOVE)
+  // Only a warning is displayed
+  imageBase->SetSpacing(negativeSpacing);
+  EXPECT_EQ(imageBase->GetSpacing(), negativeSpacing);
+
+  // Set the spacing value back to its default value
+  imageBase->SetSpacing(initialSpacing);
+#else
+  EXPECT_THROW(imageBase->SetSpacing(negativeSpacing), itk::ExceptionObject);
+  EXPECT_EQ(imageBase->GetSpacing(), initialSpacing);
+#endif
+
+  EXPECT_THROW(imageBase->SetSpacing(zeroSpacing), itk::ExceptionObject);
+  EXPECT_EQ(imageBase->GetSpacing(), initialSpacing);
+
+  const DirectionType initialDirection = imageBase->GetDirection();
+  const DirectionType zeroDirection{};
+
+  EXPECT_THROW(imageBase->SetDirection(zeroDirection), itk::ExceptionObject);
+
+  // The direction should be kept unmodified after the attempt
+  EXPECT_EQ(imageBase->GetDirection(), initialDirection);
+}
+
 } // end namespace
 
 
@@ -138,4 +176,10 @@ TEST(ImageBase, New)
 {
   Check_New_ImageBase<2>();
   Check_New_ImageBase<3>();
+}
+
+TEST(ImageBase, InvalidSpacingExceptions)
+{
+  CheckInvalidSpacingExceptions<2>();
+  CheckInvalidSpacingExceptions<3>();
 }


jhlegarreta:
I am not used to using `GTest` so I may need some iterations to succeed. Thanks for the understanding and patience.

N-Dekker:
Thanks for giving GTest a try, Jon! Cool! Looks promising  😃 

I find it a great advantage of GTest unit tests, that it eases testing one specific requirement (or property) per test function (rather than having a big "test main" full of checks). In this case, I would suggests a specific test function that checks what should happen with a negative spacing, one that checks the behavior for zero spacing, etc.

Of course, you could still combine it into a test function for invalid spacings. But I would prefer to keep it separate from the test function that specifically checks the result of `ImageBase::New()`.

So I would add something like:

    TEST(ImageBase, InvalidSpacingExceptions)
    {
      ...
    }


Hope that helps 😃

dzenanz:
You could look at some recent PR where Niels is adding GTests for an example of his preferred style 😄

jhlegarreta:
@N-Dekker please I do not mean to argue about these aspects, but I need to state why I used gtest in this case and my view given the prompt/suggestions. __Note that I do not want to engage myself in a discussion that is unrelated to the needs of the `itk::ImageBase` code__. If maintainers think that we should write new tests using gtest, I'd comply with that :100:, but that debate is best for discourse 💬 .

> Hope that helps 😃

It did help. Thanks !

While using gtest might be interesting, I only modified the `gtest` because the class at issue has no non-`gtest`. Sorry if it seemed something else. IMHO using `gtest` adds an additional dependency (with all its implications about support, documentation, life expectations of the framework, etc.). Maybe I have not used it enough, but I think we can write all our tests (including the needs mentioned in issue #983) without `gtest` without a considerable amount of work (but some work, yes). The current style across `gtest` and non-`gtest` tests is also different and makes it harder to switch between them/to read them. But __I do not mean to argue about it__.

> (...) it eases testing one specific requirement (or property) per test function (rather than having a big "test main" full of checks). In this case, 

Ultimately, it is in our hands to adopt such a style in non-`gtest` tests. Many of the current non-`gtest` tests are split into helper functions or other methods that do a particular job that fits best into separate methods.

One can also argue that the suggested approach to split virtually all methods into different functions introduces boiler-plate code, because for example, and in the simplest case, an image had to be created yet another time for the method that I added.

I'd say that once one gets used the non-`gtests`, provided that the style is consistent, the majority of them are easy to read. One of the keys being consistency.

Following that, I would argue about the reasons why the particular test modified in this PR the style does not stick to the ITK style (e.g. note the use of underscores in method names): is that the `gtest` style? Even if it were, is that documented in the ITK SW Guide? Ain't it more consistent to use the same style across `gtest` and non-`gtest` tests? So I am not that convinced by that argument, but again, __I do not mean to argue about it__.

jhlegarreta:
@hjmjohnson @thewtex @blowekamp @dzenanz @N-Dekker a few comments strictly related to the class/test below. I might be missing something fundamental. If that is the case, sorry for wasting your time all. Hopefully the test is able to capture such inconsistencies if they exist.

This check
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkImageBase.hxx#L75

is checking the ivar value, not the actual method parameter value. Is there any reason for that? It does not look quite right to me.

Similarly, and although for all practical purposes a negative or 0-spacing image will be unusable, reading the code, I'm wondering why setting a negative/0.0 spacing is still (seems to be?) allowed:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkImageBase.hxx#L90

That does not really match the message that the related exception in `ComputeIndexToPhysicalPointMatrices` conveys:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkImageBase.hxx#L165

Same applies to the direction:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkImageBase.hxx#L172

jhlegarreta:
As it detaches from https://github.com/InsightSoftwareConsortium/ITK/pull/2911#issuecomment-984207103, the test has captured what it looks to me an inconsistent behavior:
https://open.cdash.org/test/554052473

Comments are welcome.

thewtex:
@jhlegarreta thank you for these coverage improvements! :books: 

>  checking the ivar value, not the actual method parameter value. Is there any reason for that? It does not look quite right to me.

If you mean it is using `this->m_Spacing` instead of `this->GetSpacing()`, that should be ok and preferred since this is inside `::SetSpacing` -- if `GetSpacing` in the future had a side effect, there could be undesirable interactions between `GetSpacing` -> `SetSpacing`.


> is checking the ivar value, not the actual method parameter value. Is there any reason for that? It does not look quite right to me.

> Similarly, and although for all practical purposes a negative or 0-spacing image will be unusable, reading the code, I'm wondering why setting a negative/0.0 spacing is still (seems to be?) allowed:
master/Modules/Core/Common/include/itkImageBase.hxx#L90

While, yes, a negative spacing does not follow ITK's definition of spacing as a length and it would could invalid behavior in many algorithms, this was made into a warning instead of an error because feedback from the community suggested that it would break existing data/code that was using negative spacing. We could maybe change this from a warning to an error in ITK 6. A tracking issue here: https://github.com/InsightSoftwareConsortium/ITK/issues/2917

> That does not really match the message that the related exception in ComputeIndexToPhysicalPointMatrices conveys:
master/Modules/Core/Common/include/itkImageBase.hxx#L165

> Same applies to the direction:
master/Modules/Core/Common/include/itkImageBase.hxx#L172

0 spacing or 0 direction determinant are also both degenerate cases but have always not been supported.

jhlegarreta:
Thanks for the comments @thewtex. Looks like I was not clear enough:

> If you mean it is using this->m_Spacing instead of this->GetSpacing(), that should be ok and preferred since this is inside ::SetSpacing -- if GetSpacing in the future had a side effect, there could be undesirable interactions between GetSpacing -> SetSpacing.

I meant 
```
for (unsigned int i = 0; i < VImageDimension; ++i)
{
  if (spacing[i] < 0.0)
  {
  (...)
```

instead of
```
if (this->m_Spacing[i] < 0.0)
for (unsigned int i = 0; i < VImageDimension; ++i)
{
  if (this->m_Spacing[i] < 0.0)
  {
  (...)
```

But now reading #2917 I guess that explains it.

So I will modify the test such that it allows and checks 0 and negative spacing values, and once the strict enforcement is implemented the test will serve as a first heads-up.

N-Dekker:
@thewtex @jhlegarreta  It would make sense to me as well, to check the `spacing` parameter (rather than the previous value of the `ImageBase::Spacing` property), inside `ImageBase::SetSpacing(const SpacingType & spacing)`. 

Just my two cents.

dzenanz:
> This check
master/Modules/Core/Common/include/itkImageBase.hxx#L75
 is checking the ivar value, not the actual method parameter value.

I think this is a bug. The [original implementation](https://github.com/InsightSoftwareConsortium/ITK/commit/d447f0452bb5ea92a555e630d05b57da535bd3a9) botched it (sorry Brad). I agree with [Jon's suggestion](https://github.com/InsightSoftwareConsortium/ITK/pull/2911#issuecomment-984793594).

dzenanz:
PR incoming.

blowekamp:
> I think this is a bug. The [original implementation](https://github.com/InsightSoftwareConsortium/ITK/commit/d447f0452bb5ea92a555e630d05b57da535bd3a9) botched it (sorry Brad). I agree with [Jon's suggestion](https://github.com/InsightSoftwareConsortium/ITK/pull/2911#issuecomment-984793594).

I agree that it's incorrect. There was no test so I'm not surprised. Thanks Jon.

I'll also point out that this commit is noted as "blowekamp authored and Jon Haitz Legarreta Gorroño committed" so... I'm not sure about the history and origin of it 🤷

jhlegarreta:
> I'll also point out that this commit is noted as "blowekamp authored and Jon Haitz Legarreta Gorroño committed" so... I'm not sure about the history and origin of it shrug

Maybe it all stems from a fault of my own, an oversight of mine at the time that commit was merged, etc. so I'm sorry if that is the case :pensive:.

thewtex (approved):
:green_circle: :straight_ruler:

Modules/Core/Common/test/itkImageBaseGTest.cxx
```diff
@@ -119,6 +119,65 @@ Check_New_ImageBase()
...
+{
+  using SpacingValueType = typename itk::ImageBase<VImageDimension>::SpacingValueType;
+  using SpacingType = typename itk::ImageBase<VImageDimension>::SpacingType;
+  using DirectionType = typename itk::ImageBase<VImageDimension>::DirectionType;
+
+  const auto imageBase = itk::ImageBase<VImageDimension>::New();
+
+  // Test exceptions
+  SpacingValueType spacingValue = -1.0;
+  SpacingType      spacing;
```
N-Dekker, line 134: Just a little suggestion: instead of reusing the very same local `spacing` variable, I would use three `const` spacings, to further improve code readability:

    const SpacingType initialSpacing = imageBase->GetSpacing();
    const SpacingType negativeSpacing(-1.0);
    const SpacingType zeroSpacing(0.0);

jhlegarreta: Thanks for the review. Reads definitely more cleanly. Force pushed in  8c06b7a.

dzenanz (approved):
Awesome!

Modules/Core/Common/test/itkImageBaseGTest.cxx
```diff
@@ -119,6 +119,46 @@ Check_New_ImageBase()
...
+  EXPECT_EQ(imageBase->GetSpacing(), initialSpacing);
+
+  SpacingValueType directionValue = 0.0;
+  DirectionType    direction;
+  direction.Fill(directionValue);
+
+  EXPECT_THROW(imageBase->SetDirection(direction), itk::ExceptionObject);
+
+  // The direction should be kept unmodified after the attempt
+  EXPECT_TRUE(imageBase->GetDirection().GetVnlMatrix().is_identity());
```
N-Dekker, line 159: Thanks for following my suggestion @jhlegarreta  You might do the same for the `Direction` property, if you like. See below. But I leave it up to you.
```

  const DirectionType initialDirection = imageBase->GetDirection();
  const DirectionType zeroDirection{};

  EXPECT_THROW(imageBase->SetDirection(zeroDirection), itk::ExceptionObject);

  // The direction should be kept unmodified after the attempt
  EXPECT_EQ(imageBase->GetDirection(), initialDirection);

```

N-Dekker (approved):

Modules/Core/Common/test/itkImageBaseGTest.cxx
jhlegarreta: Done in  50fdb97.

dzenanz (approved):

Issue #2917, opened 2021-12-02
Enforce positive Spacing


thewtex:
### Description

While image pixel spacing is assumed to be a positive length unit in ITK, this had not been strictly enforced in the past, and some applications used negative spacing to flip data direction. If a flip is desired, this is supported in ITK's `Direction` instead of `Spacing`.

Change the warning here:

master/Modules/Core/Common/include/itkImageBase.hxx#L75

to an error.

### Impact analysis

Changes may be required in existing applications. This update would be associated with ITK 6.

### Expected behavior

An attempt to set negative spacing throws an error.

### Actual behavior

This is not enforced in code.

### Versions

ITK version 5.X and below.

blowekamp:
There exists image files with negative spacing. We must ensure that we are still able to read these files.

I think this is already done in the ImageFileReader:
https://github.com/InsightSoftwareConsortium/ITK/blob/6853d63728a181e437f45473deb53c111eb1a6b7/Modules/IO/ImageBase/include/itkImageFileReader.hxx#L220-L233

seanm:
I recall some previous ITK version (4 to 5?) did some changes with negative space and I had to carry a local patch for a while... I'd be interested to test any PR that comes along...

blowekamp:
Could this be implemented now with a "FUTURE_LEGACY_REMOVE" conditional?

dzenanz:
It was [originally implemented](https://github.com/InsightSoftwareConsortium/ITK/commit/d447f0452bb5ea92a555e630d05b57da535bd3a9) with `FUTURE_LEGACY_REMOVE`, and that has been transformed into `LEGACY_REMOVE` during [transition to 5.0](https://github.com/InsightSoftwareConsortium/ITK/commit/5c84e6e6b4768e596439c9879d86f51fa4bda51a). So when we switch to 6.0, just remove conditional and throw an exception unconditionally.

dzenanz:
The logic in my previous comment would hold, have the feature been working properly, see discussion in https://github.com/InsightSoftwareConsortium/ITK/pull/2911#issuecomment-984941366. Now I am not sure whether to remove it completely in 6.0, or keep it as legacy.

blowekamp:
[...]

PR #2918, opened 2021-12-02
BUG: negative spacing is not supported, but check did not enforce that

dzenanz:
Related to #2911 and #2917.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkImageBase.hxx
+++ b/Modules/Core/Common/include/itkImageBase.hxx
@@ -141,6 +141,12 @@ ImageBase<VImageDimension>::SetDirection(const DirectionType & direction)
 {
   bool modified = false;
 
+  if (vnl_determinant(direction.GetVnlMatrix()) == 0.0)
+  {
+    itkExceptionMacro("Bad direction, determinant is 0. Refusing to change direction from " << this->m_Direction
+                                                                                            << " to " << direction);
+  }
+
   for (unsigned int r = 0; r < VImageDimension; ++r)
   {
     for (unsigned int c = 0; c < VImageDimension; ++c)
@@ -170,11 +176,6 @@ ImageBase<VImageDimension>::ComputeIndexToPhysicalPointMatrices()
   for (unsigned int i = 0; i < VImageDimension; ++i)
   {
     scale[i][i] = this->m_Spacing[i];
-  }
-
-  if (vnl_determinant(this->m_Direction.GetVnlMatrix()) == 0.0)
-  {
-    itkExceptionMacro(<< "Bad direction, determinant is 0. Direction is " << this->m_Direction);
   }
 
   this->m_IndexToPhysicalPoint = this->m_Direction * scale;
--- a/Modules/Core/Common/include/itkImageBase.hxx
+++ b/Modules/Core/Common/include/itkImageBase.hxx
@@ -72,14 +72,23 @@ ImageBase<VImageDimension>::SetSpacing(const SpacingType & spacing)
 {
   for (unsigned int i = 0; i < VImageDimension; ++i)
   {
-    if (this->m_Spacing[i] < 0.0)
+    // Check for zero-valued spacing
+    if (spacing[i] == 0.0)
     {
+      itkExceptionMacro(
+        "Zero-valued spacing is not supported and may result in undefined behavior.\nRefusing to change spacing from "
+        << this->m_Spacing << " to " << spacing);
+    }
+
+    // Check for negative-valued spacing
+    if (spacing[i] < 0.0)
+    {
+      constexpr char message[] = "Negative spacing is not supported and may result in undefined behavior.\n";
 #if !defined(ITK_LEGACY_REMOVE)
-      itkWarningMacro("Negative spacing is not supported and may result in undefined behavior. Spacing is "
-                      << this->m_Spacing);
+      itkWarningMacro(<< message << "Proceeding to set spacing to " << spacing);
       break;
 #else
-      itkExceptionMacro("Negative spacing is not allowed: Spacing is " << this->m_Spacing);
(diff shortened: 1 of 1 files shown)


Modules/Core/Common/include/itkImageBase.hxx
```diff
@@ -72,14 +72,13 @@ ImageBase<VImageDimension>::SetSpacing(const SpacingType & spacing)
 {
   for (unsigned int i = 0; i < VImageDimension; ++i)
   {
-    if (this->m_Spacing[i] < 0.0)
+    if (spacing[i] < 0.0)
     {
 #if !defined(ITK_LEGACY_REMOVE)
-      itkWarningMacro("Negative spacing is not supported and may result in undefined behavior. Spacing is "
-                      << this->m_Spacing);
+      itkWarningMacro("Negative spacing is not supported and may result in undefined behavior. Spacing is " << spacing);
```
blowekamp, line 78: This message should say something like:

"Negative spacing is not supported and may result in undefined behavior. Proceeding to set spacing to " <<

Modules/Core/Common/include/itkImageBase.hxx
```diff
@@ -72,14 +72,13 @@ ImageBase<VImageDimension>::SetSpacing(const SpacingType & spacing)
...
       break;
 #else
-      itkExceptionMacro("Negative spacing is not allowed: Spacing is " << this->m_Spacing);
+      itkExceptionMacro("Negative spacing is not allowed: Spacing is " << spacing);
```
blowekamp, line 81: This message should say something like:

"Negative spacing is not supported and may result in undefined behavior. Refusing to change spacing to " << spacing << " from " << this->m_Spacing

blowekamp (commented):
Thanks for fixing the logic. The error messages need a update to be accurate and clear.

jhlegarreta:
@dzenanz thanks for doing this.

After reading Matt's and Brad's comments in #2917, and the need to at least read negative spacing images by third party applications, maybe commit d447f0452bb5ea92a555e630d05b57da535bd3a9 was thought to be a halfway compromise before the strict enforcement, as it is guarded by the `FUTURE_LEGACY_REMOVE` macro (but I still find the message inaccurate/contradicting the fact that setting the zero/negative spacing does actually take place).

For context, maybe some related discussion about zero/negative spacings are these (?):
https://itk.org/pipermail/community/2016-November/012275.html (newest reply from François)
https://discourse.itk.org/t/nifti-file-orientation-change-between-2019-08-26-and-itk-v5-1rc01/2594/12?
https://public.kitware.com/pipermail/insight-users/2016-October/054037.html

But reading negative (at least negative) spacing images (but immediately setting to an "equivalent" positive value) is still allowed, as Brad points in https://github.com/InsightSoftwareConsortium/ITK/issues/2917#issuecomment-984794713.

Thus, now I am not sure whether I unnecessarily opened a front with my PR and thoughts. Sorry, not the most lucid days.

Modules/Core/Common/include/itkImageBase.hxx
dzenanz: Done in force-push.

blowekamp:
> Thus, now I am not sure whether I unnecessarily opened a front with my PR and thoughts. Sorry, not the most lucid days.

Jon, I am glad your efforts have gotten this issue straightened out. I recall encountering it messing around with SimpleITK and setting the spacing and getting a confusing error message.

jhlegarreta (commented):
The message and the check make more sense to me :-).

I guess the block should be copied and adapted to check and warn about the same situation for a zero-valued spacing, isn't it? This would mean removing (or moving to `SetSpacing` and adapting) this check:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkImageBase.hxx#L163

At that point, `this->m_Spacing[i] == 0` should never evaluate to `true`.

And I guess a separate PR or another commit in this PR could do the same for zero-valued directions?
i.e. add a check here:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkImageBase.hxx#L131
As well as moving this other direction-related determinant check (also checking the method parameter instead of `m_Direction`) to `SetDirection` as well ?
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkImageBase.hxx#L170

dzenanz:
@jhlegarreta I have "Allow edits and access to secrets by maintainers" enabled by default, and it is active for this PR too. Feel free to add an additional commits, or amend this one. Also, you can make a separate PR, possibly later. As the changes in this PR are minimal, I am happy to abandon it in favor of a more comprehensive one.

jhlegarreta:
So the legacy remove build is doing its job :ok_hand:  and we have the first issue uncovered by it:
https://open.cdash.org/test/554773719

The exception seems to be thrown in here:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Registration/Metricsv4/test/itkImageToImageMetricv4Test.cxx#L597
So it looks likes some registration process does

jhlegarreta:
> @jhlegarreta I have "Allow edits and access to secrets by maintainers" enabled by default, and it is active for this PR too. Feel free to add an additional commits, or amend this one. Also, you can make a separate PR, possibly later. As the changes in this PR are minimal, I am happy to abandon it in favor of a more comprehensive one.

Done in 8df1ea4.

dzenanz:
I guess we should change the metric exception testing to trigger it by something other than negative spacing.

jhlegarreta:
> Not sure about the positive spacing values that would still raise the metric exception, but there are a few other displacement field image properties that might raise an exception:

> Still expect the legacy remove build failure.

Here it is #2923. Tested locally. Looks to be good enough to raise the exception.

dzenanz:
This now passes tests. It is ready for reviewing and merging.

blowekamp (approved):
Nice!

PR #2919, opened 2021-12-02
BUG: Wrong inverse matrix in MatrixOffsetTransformBase::ComputeInverseJacobianWithRespectToPosition

josempozo:
itk::MatrixOffsetTransformBase::ComputeInverseJacobianWithRespectToPosition was returning directly m_Matrix, the same as the (non-inverse) ComputeJacobianWithRespectToPosition. This has been corrected to return the inverse matrix.

This correction has required also the correction of a few function return and parameters' Types. 
Those corrected Types were inconsistent when instantiated with input and output of different dimensions.

In addition, the generation of the inverse transformation required declaring the same template class, but with the input and output dimensions interchanged, as friend class.

Apparently the bug was inadvertently introduced in 
Change-Id: Ib4d253de1a592693ed253a3e7e95ddbe21df76c4 (PERF: Improve MatrixOffset Jacobian computation performance)
and subsequently not fixed but got mixed with wrong JacobianTypes in
Change-Id: I3c99bd36f915f97d3e9488d5071cb272e2f7743e (PERF: Change type of Jacobian w.r. position to vnl_matrix_fixed).

### DIFF
--- a/Modules/Core/Transform/include/itkMatrixOffsetTransformBase.hxx
+++ b/Modules/Core/Transform/include/itkMatrixOffsetTransformBase.hxx
@@ -417,7 +417,8 @@ MatrixOffsetTransformBase<TParametersValueType, NInputDimensions, NOutputDimensi
 
 template <typename TParametersValueType, unsigned int NInputDimensions, unsigned int NOutputDimensions>
 bool
-MatrixOffsetTransformBase<TParametersValueType, NInputDimensions, NOutputDimensions>::GetInverse(Self * inverse) const
+MatrixOffsetTransformBase<TParametersValueType, NInputDimensions, NOutputDimensions>::GetInverse(
+  InverseTransformType * inverse) const
 {
   if (!inverse)
   {
@@ -446,7 +447,7 @@ typename MatrixOffsetTransformBase<TParametersValueType, NInputDimensions, NOutp
   InverseTransformBasePointer
   MatrixOffsetTransformBase<TParametersValueType, NInputDimensions, NOutputDimensions>::GetInverseTransform() const
 {
-  Pointer inv = New();
+  auto inv = InverseTransformType::New();
 
   return GetInverse(inv) ? inv.GetPointer() : nullptr;
 }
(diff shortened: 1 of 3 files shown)


N-Dekker:
@josempozo  Interesting, thanks, I'll have a look. Can you possibly also add a minimal unit test? (I mean, a test that is _just_ big enough to demonstrate the bug, and to check the fix.) Kind regards, Niels

josempozo:
> @josempozo Interesting, thanks, I'll have a look. Can you possibly also add a minimal unit test? (I mean, a test that is _just_ big enough to demonstrate the bug, and to check the fix.) Kind regards, Niels

@N-Dekker, thanks for the suggestion. I will try to make this test.

dzenanz (approved):
These commits should be squashed (can be done at merge time), especially given:

commit 13f91f1 has an invalid commit subject; the first line must be no longer than 78 characters.
commit 2b04e9b has an invalid commit subject; the first line must be no longer than 78 characters.

Modules/Core/Transform/include/itkMatrixOffsetTransformBase.hxx
```diff
@@ -446,7 +447,7 @@ typename MatrixOffsetTransformBase<TParametersValueType, NInputDimensions, NOutp
   InverseTransformBasePointer
   MatrixOffsetTransformBase<TParametersValueType, NInputDimensions, NOutputDimensions>::GetInverseTransform() const
 {
-  Pointer inv = New();
+  typename InverseTransformType::Pointer inv = InverseTransformType::New();
```
N-Dekker, line 450: (Minor issue) Please consider using `auto` here, following pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2826 "STYLE: Use auto for declaration of variables initialized by `New()`"

josempozo: I have now amended it using auto.

N-Dekker: Thanks @josempozo !

dzenanz:
I squashed the fix commits, and changed the message to use SHA1s instead of ChangeIDs (which are not longer used).

thewtex (approved):
Thank you for the contribution! :pray:

PR #2932, opened 2021-12-09
COMP: Use `itk::Math::abs` for floating point type argument

jhlegarreta:
Use `itk::Math::abs` instead of `abs` for floating point type argument.

Fixes:
```
[CTest: warning matched]
/Users/builder/externalModules/Core/Transform/test/itkAffineTransformTest.cxx:619:11:
warning: using integer absolute value function 'abs' when argument is of floating point type [-Wabsolute-value]
      if (abs(jaffInverseJacobianPosition[i][j] - matrix3invTruth[i][j]) > 1e-13)
          ^
[CTest: warning matched]
/Users/builder/externalModules/Core/Transform/test/itkAffineTransformTest.cxx:619:11:
note: use function 'std::abs' instead
      if (abs(jaffInverseJacobianPosition[i][j] - matrix3invTruth[i][j]) > 1e-13)
          ^~~
          std::abs
```

raised for example in:
https://open.cdash.org/viewBuildError.php?type=1&onlydeltap&buildid=7615457

## PR Checklist

### DIFF
--- a/Modules/Core/Transform/test/itkAffineTransformTest.cxx
+++ b/Modules/Core/Transform/test/itkAffineTransformTest.cxx
@@ -616,7 +616,7 @@ itkAffineTransformTest(int, char *[])
   {
     for (unsigned int j = 0; j < Affine3DType::MatrixType::ColumnDimensions; ++j)
     {
-      if (abs(jaffInverseJacobianPosition[i][j] - matrix3invTruth[i][j]) > 1e-13)
+      if (itk::Math::abs(jaffInverseJacobianPosition[i][j] - matrix3invTruth[i][j]) > 1e-13)
       {
         std::cout << "Failed ComputeInverseJacobianWithRespectToPosition." << std::endl
                   << "jaffInverseJacobianPosition: " << jaffInverseJacobianPosition << std::endl


dzenanz (approved):

jhlegarreta:
Introduced in 7c7233f.

@dzenanz @josempozo.

dzenanz:
I saw this morning that a few of Sean's machines produce this warning. Thanks for fixing it Jon!

seanm (approved):

Leengit:
```bash
fgrep -R 'abs(' ~/git/ITK | fgrep -v '::abs' | fgrep -v fabs | fgrep -v ThirdParty
```
finds more locations that might benefit from instead using `std::abs`.  Would it make sense to address these with this pull request?

seanm:
Would be nice to see this one merged to get cdash greener. This PR fixes a warning that was not there yesterday.

dzenanz:
You are welcome to make a follow-up PR @Leengit.

jhlegarreta:
> finds more locations that might benefit from instead using std::abs

There is also `std::fabs` so not sure whether we should always stick to `std::fabs` for non-integer types to avoid taking into account the floating point type precision:
https://en.cppreference.com/w/cpp/numeric/math/fabs

dzenanz (approved):

jhlegarreta:
#2934 includes this fix. Feel free to close it in favor or that one, or else, in case it is deemed more secure to merge this one to have the warning go, and later merge the former as an improvement, the former will need to be rebased and the conflict fixed.

dzenanz:
The other PR is ready to be merged, so let's close this one.

Issue #2928, opened 2021-12-07
Crash when use ImageFileReader read jpeg 


Hconk:
### Description
If compile with release version. When I use `ImageFileReader` read some broke jpeg image, it's crash in Update function. I expected throw an exception. 
But success in Debug.

### Steps to Reproduce
Build With Release. 
https://gist.github.com/Hconk/5b1bed75ebb3c62757277a1367d61edb

### Expected behavior
If image broke, expect throw an exception.

### Actual behavior
Crash at update. 

![image](https://user-images.githubusercontent.com/8470001/144960323-9cf583e8-8b78-4559-be58-218cb45e51e1.png)

![image](https://user-images.githubusercontent.com/8470001/144960311-6782e907-34bc-4cce-8fd1-28c1fdaf91f1.png)

### Reproducibility
Release Version. 

### Versions
c4960ee19c

### Environment
* CMake:  3.21.1
* OS: Centos 7.3 / Windows 10
* ITK Version:  5.2.1 / master branch
* compile: gcc10.2 / gcc 4.8 / vs2017

### Additional Information
Test data.
[Google Drive](https://drive.google.com/file/d/1rL4cmT8ZBqPEmyHny6MfByVg4wt_6MCQ/view?usp=sharing)

![211869_09365_55_55](https://user-images.githubusercontent.com/8470001/144959265-18556a05-b8c6-4ceb-b7e5-cae99ef917cd.jpg)

dzenanz:
I can't reproduce your error with a recent `master` 05ae9a5061555598bff524ff958fca66aa5b37db.
```c++
#include <itkJPEGImageIO.h>
#include <itkImageFileReader.h>

int main(int, char**) {
  constexpr unsigned int Dimension = 2;
  using PixelType = unsigned char;
  using ImageType = itk::Image<PixelType, Dimension>;
  itk::JPEGImageIO::Pointer io;
  io = itk::JPEGImageIO::New();
  itk::ImageFileReader<ImageType>::Pointer JPEGReader;
  JPEGReader = itk::ImageFileReader<ImageType>::New();
  JPEGReader->SetFileName("C:/a/211869_09365_55_55.jpg");
  JPEGReader->SetImageIO(io);
  try
  {
    JPEGReader->Update();
    std::cout << *JPEGReader->GetOutput();
  }
  catch (itk::ExceptionObject& err)
  {
    std::cerr << "ExceptionObject caught !" << std::endl;
[...]

PR #2933, opened 2021-12-09
BUG: fixed undefined behaviour with corrupted JPEG file

dzenanz:
See https://github.com/InsightSoftwareConsortium/ITK/issues/2928.


dzenanz (commented):
Mostly looks good.

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
```diff
@@ -36,7 +36,7 @@ extern "C"
   {
     /* cinfo->err really points to a itk_jpeg_error_mgr struct, so coerce pointer
      */
-    auto * myerr = reinterpret_cast<itk_jpeg_error_mgr *>(cinfo->err);
+    itk_jpeg_error_mgr * myerr = (itk_jpeg_error_mgr *)cinfo->err;
```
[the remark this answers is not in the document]
issakomi: I have imagined JPEG style here. OK.

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
```diff
@@ -321,6 +319,13 @@ JPEGImageIO::ReadImageInformation()
   // ability to scale an image on input).
   jpeg_calc_output_dimensions(&cinfo);
```
issakomi, line 320: Maybe call this _jpeg_calc_output_dimensions(&cinfo)_ in _Read_ function too, from the comment it looks like there may be situation where _cinfo.image_height_ etc. may be affected? And m_Dimensions are saved here from _cinfo_. Not 100% sure. Gimp code doesn't have that _jpeg_calc_output_dimensions_ call.

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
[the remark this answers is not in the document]
issakomi: I have reverted. But still don't know, IMHO C code were not so bad in `extern "C"` block.

jhlegarreta:
Have not followed closely the issue/source of the problem, but :100: to all involved for investigating this and proposing a fix.

Can the image file that demonstrates the issue be added to the test images and be provided to the test as another case so that to ensure that the issue does not leak back into the code ?

issakomi:
The image provided by @Hconk, from the issue mentioned above:
https://drive.google.com/file/d/1rL4cmT8ZBqPEmyHny6MfByVg4wt_6MCQ/view

issakomi:
@jhlegarreta Short video:
https://youtu.be/zJf3maxYaN0

dzenanz:
A [warning](https://open.cdash.org/viewBuildError.php?type=1&buildid=7616046): `Modules/IO/JPEG/src/itkJPEGImageIO.cxx:160:10: warning: variable 'ui' might be clobbered by 'longjmp' or 'vfork' [-Wclobbered]`

issakomi:
@dzenanz updated

issakomi:
@dzenanz It is probably GCC 4.8 compiler? Wait a little, i didn't try with 4.8. My 8.3 doesn't have warnings, there are could be more. Just a moment, please.

issakomi:
This one with "ui" is gone. There is one more warning with GCC 4.8, in WriteSlice, the last one, but i shall look closer later, no time now.

issakomi:
All tests look good, but there is still one warning from _WriteSlice_ function with GCC 4.8 (don't see with other compilers). Honestly it is a little bit too fast, i would prefer to test more.

jhlegarreta:
> Can the image file that demonstrates the issue or an appropriate one (i.e. having some compatible license) be added to the test images and be provided to the test as another case so that to ensure that the issue does not leak back into the code ?

Here it is: https://github.com/issakomi/ITK/pull/1

jhlegarreta:
@Hconk hope it is OK to use the image you shared. Let us know otherwise.

issakomi:
Thank you, i have merged the test. I am still working on the branch. It is WIP, please don't merge into ITK master just now.

The image is interesting. In fact if someone just cut in hex editor parts of a image, the JPEG can survive, i tried to produce corrupted images, there were only warnings. BTW, warnings are unfortunately invisible in current JPEG IO, messages are supressed.
`METHODDEF(void) itk_jpeg_output_message(j_common_ptr) {}`


This particular image has 2 issues

```
Corrupt JPEG data: premature end of data segment
Unsupported marker type 0x02
```
Don't know exactly, may be second causes the critical error, not 100% sure.

issakomi:
To demonstrate the bug with _setjmp_ wrapping, _Debian 10_, _GCC 8.3_

This program

```cpp
#include <cstdio>
#include <csetjmp>

namespace
{
bool wrap(jmp_buf env)
{
  if (setjmp(env))
  {
    return true;
  }
  return false;
}
}

int main(int, char**)
{
  jmp_buf env;
  volatile int x = 0;
//  if (wrap(env))
  if (setjmp(env))
  {
    ++x;
    printf("%d\n", x);
  }
  longjmp(env, 1);
  return 0;
}
```

with `if (setjmp(env))` will print incrementing number forever (caution, must terminate with Ctrl-C), jumping forever.

```
r@deb:~/Desktop$ ./a.out 
1
2
3
4
...
^C
```

with `if (wrap(env))`

```
r@deb:~/Desktop$ ./a.out 
Segmentation fault
```
Similar program

```cpp
#include <cstdio>
#include <csetjmp>

namespace
{
bool wrap(jmp_buf env)
{
  if (setjmp(env))
  {
    return true;
  }
  return false;
}
}

int main(int, char**)
{
  jmp_buf env;
  volatile int x = 0;
//  if (wrap(env))
  if (setjmp(env))
  {
    ++x;
    if (x <= 5)
    {
      printf("%d\n", x);
    }
    else
    {
      printf("return point 1\n");
      return 0;
    }
  }
  longjmp(env, 1);
  printf("return point 2\n");
  return 0;
}
```

with `if (setjmp(env))`

```
r@deb:~$ ./a.out 
1
2
3
4
5
return point 1
r@deb:~$
```

with `if (wrap(env))`

```
r@deb:~/$ ./a.out 
r@deb:~$ 
```

prints nothing somehow.

Hconk:
> @Hconk hope it is OK to use the image you shared. Let us know otherwise.

Yes，it's OK. I hava some similar images, If necessary, I will share it when I have time.

Hconk:
> Thank you, i have merged the test. I am still working on the branch. It is WIP, please don't merge into ITK master just now.
> 
> The image is interesting. In fact if someone just cut in hex editor parts of a image, JPEG library can survive, i tried to produce corrupted images, there were only warnings. BTW, warnings are unfortunately invisible in current JPEG IO, messages are supressed. `METHODDEF(void) itk_jpeg_output_message(j_common_ptr) {}`
> 
> This particular image has 2 issues
> 
> ```
> Corrupt JPEG data: premature end of data segment
> Unsupported marker type 0x02
> ```
> 
> Don't know exactly, maybe the second causes the critical error, not 100% sure.

I found easy way to produce corrupted jpeg file.  insert `FF 02` into `SOS(0xFFDA)`  field  will cause crash. Hope to bring some help to your work. 
Minimum	example
good image hex:
```
λ Desktop itk_demo → cat ./good.jpg| xxd -u
00000000: FFD8 FFE0 0010 4A46 4946 0001 0101 0060  ......JFIF.....`
00000010: 0060 0000 FFDB 0043 0001 0101 0101 0101  .`.....C........
00000020: 0101 0101 0101 0101 0101 0101 0101 0101  ................
00000030: 0101 0101 0101 0101 0101 0101 0101 0101  ................
00000040: 0101 0101 0101 0101 0101 0101 0101 0101  ................
00000050: 0101 0101 0101 0101 01FF C900 0B08 0001  ................
00000060: 0001 0101 1100 FFDA 0008 0101 0000 3F00  ..............?.
00000070: D2CF 20FF D9                             .. ..
```

corrupted image hex:
```
λ Desktop itk_demo → cat corrupted.jpg| xxd -u
00000000: FFD8 FFE0 0010 4A46 4946 0001 0101 0060  ......JFIF.....`
00000010: 0060 0000 FFDB 0043 0001 0101 0101 0101  .`.....C........
00000020: 0101 0101 0101 0101 0101 0101 0101 0101  ................
00000030: 0101 0101 0101 0101 0101 0101 0101 0101  ................
00000040: 0101 0101 0101 0101 0101 0101 0101 0101  ................
00000050: 0101 0101 0101 0101 01FF C900 0B08 0001  ................
00000060: 0001 0101 1100 FFDA 0008 0101 0000 3F00  ..............?.
00000070: D2CF 20FF 02FF D9                        .. ....
```

issakomi:
@Hconk thank you!

I have optionally enabled JPEG messages `#define JPEGIO_JPEG_MESSAGES 1` [here](https://github.com/issakomi/ITK/blob/2889f5c3d5e8187919bd543b867d0c0b9203ee46/Modules/IO/JPEG/src/itkJPEGImageIO.cxx#L25), maybe i shall have to  set default to _false_ later, not sure what ITK owners think. The messages are important, IMHO.

It is still **WIP**, there are some things i would like to check, specially in _WriteSlice_ functions.

Modules/IO/JPEG/include/itkJPEGImageIO.h
```diff
@@ -111,7 +111,7 @@ class ITKIOJPEG_EXPORT JPEGImageIO : public ImageIOBase
   PrintSelf(std::ostream & os, Indent indent) const override;
 
   void
-  WriteSlice(std::string & fileName, const void * buffer);
+  WriteSlice(std::string & fileName, const void * const buffer);
```
issakomi, line 114: This fixed warning with GCC 4.8 and possibly other compilers.
```
[ 78%] Building CXX object Modules/IO/JPEG/src/CMakeFiles/ITKIOJPEG.dir/itkJPEGImageIO.cxx.o
/home/rr/mitk/ITK/Modules/IO/JPEG/src/itkJPEGImageIO.cxx: In member function ‘void itk::JPEGImageIO::WriteSlice(std::string&, const void*)’:
/home/rr/mitk/ITK/Modules/IO/JPEG/src/itkJPEGImageIO.cxx:403:1: warning: argument ‘buffer’ might be clobbered by ‘longjmp’ or ‘vfork’ [-Wclobbered]
 JPEGImageIO::WriteSlice(std::string & fileName, const void *buffer)
 ^
```

issakomi:
Probably the PR is closer to be ready.

There are many more issues, but may be enough for this PR, can be done in separate PRs.

In particular in _Write_:
```cpp
  if (this->GetComponentType() != IOComponentEnum::UCHAR && this->GetComponentType() != IOComponentEnum::UINT)
  {
    itkExceptionMacro(<< "JPEG supports unsigned char/int only");
  }
```
It is bug, a kind of remnant from an ancient attempt to do something with JPEG 12 bits, probably. Of course `unsigned int` doesn't work, nowhere in pointer arithmetic is something related. I tried (without any hope) to save UINT 2D image and got garbage.

The same in constructor. 
```cpp
  // 12bits is not working right now, but this should be doable
#if BITS_IN_JSAMPLE == 8
  m_ComponentType = IOComponentEnum::UCHAR;
#else
  m_ComponentType = IOComponentEnum::UINT;
#endif
```
It makes no sense to allow other _BITS_IN_JSAMPLE_, it can not work. Better `#error` at this point, an attempt to use external 12 bits or more JPEG library here will end up in crashes.

Here
```cpp
  if (cinfo.image_width > 65536 || cinfo.image_height > 65536)
  {
    itkExceptionMacro(<< "JPEG : Image is too large for JPEG");
  }
```
maximum dimension is defined in _jmorecfg.h_
`#define JPEG_MAX_DIMENSION  65500L  /* a tad under 64K to prevent overflows */`
if using external JPEG library it may be set to different value (very unlikely, but theoretically possible)

_WriteSlice_ could be reviewed, IMHO.

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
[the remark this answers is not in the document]
issakomi: I would prefer to use C-style here like in all implementation i have seen. The function is a kind of extension of JPEG lib. From Wikipedia: `Some compilers will produce non-working code if a function pointer declared extern "C" points to a C++ function that is not declared extern "C".` There is also related discussion [Does a C++ cast strip the 'extern "C"' from a declaration?](https://stackoverflow.com/questions/39584003/does-a-c-cast-strip-the-extern-c-from-a-declaration). Honestly i am too lazy to read that discussion, IMHO, it is safer to avoid "low reward" experiments here.

issakomi:
macOS.Python  _itkPyBufferMemoryLeakTest (Failed)_ fails for all PR for some reason

issakomi (approved):

Modules/IO/JPEG/test/itkJPEGImageIODegenerateCasesTest.cxx
```diff
@@ -0,0 +1,52 @@
...
+  using ImageType = itk::Image<PixelType, Dimension>;
+
+  itk::JPEGImageIO::Pointer io = itk::JPEGImageIO::New();
+
+  itk::ImageFileReader<ImageType>::Pointer reader = itk::ImageFileReader<ImageType>::New();
+
+  reader->SetFileName(argv[1]);
+  reader->SetImageIO(io);
+
+  ITK_TRY_EXPECT_NO_EXCEPTION(reader->Update());
```
issakomi, line 47: @jhlegarreta  This particular file doesn't trigger exception, so the test works, other broken files can trigger exception, before entering the scan-lines _while_ loop. Important is to avoid segmentation fault or crash, exception is OK. Just FYI.

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
```diff
@@ -36,37 +42,31 @@ extern "C"
...
     longjmp(myerr->setjmp_buffer, 1);
   }
 
-  METHODDEF(void) itk_jpeg_output_message(j_common_ptr) {}
+  METHODDEF(void) itk_jpeg_output_message(j_common_ptr cinfo)
+  {
+#if (defined JPEGIO_JPEG_MESSAGES && JPEGIO_JPEG_MESSAGES == 1)
+    char buffer[JMSG_LENGTH_MAX + 1];
+    (*cinfo->err->format_message)(cinfo, buffer);
+    printf("%s\n", buffer);
```
jhlegarreta, line 60: I'd say that printing messages should be done through `itkDebugMacro` macro calls instead of `print`s. But have not checked/tested in which circumstances the message should be printed.

Modules/IO/JPEG/test/itkJPEGImageIODegenerateCasesTest.cxx
jhlegarreta: Hats off to the effort to investigate this @issakomi. I cannot investigate as closely. I can only point to a few things:
- I am not sure about the debug/release compilation of the CI builds, so it might well be that the image provided by @Hconk does not trigger any issue in our CI builds. However, many CDash builds are build in release mode, so any (appropriately) corrupted file without this fix should trigger the issue as I understand it.
- If prior to this patch set a given test file was generating a segmentation fault or crash of any nature, then that test file should be added to the test images, and ensure that it does not produce a segfault with this patch set, and is dealt with nicely, either being read if it can be read or generating an exception (and I'd say exception and not warning) saying that it cannot be read. That exception can be tested with the `ITK_TRY_EXPECT_EXCEPTION` macro.
- If the image provided by @Hconk is now being dealt with nicely and can be read without issues, the `ITK_TRY_EXPECT_NO_EXCEPTION` is OK.
- Otherwise, I do not see that this patch set adds new exceptions. If any of the existing ones were not being tested, and we would like them to be tested, images triggering them should be added. See here the exceptions that are not being exercised (any line in red is not exercised/tested):
https://open.cdash.org/viewCoverageFile.php?buildid=7620043&fileid=49347999

Hope this casts some light on what is desired to fully test these cases, and to ensure that the behavior is predictable and conforming to the expectations.

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
```diff
@@ -212,21 +208,24 @@ JPEGImageIO::Read(void * buffer)
...
 
   // read the bulk data
-  unsigned int remainingRows;
   while (cinfo.output_scanline < cinfo.output_height)
   {
-    remainingRows = cinfo.output_height - cinfo.output_scanline;
-    jpeg_read_scanlines(&cinfo, &row_pointers[cinfo.output_scanline], remainingRows);
+    if (setjmp(jerr.setjmp_buffer))
+    {
+      itkWarningMacro("JPEG error in the file " << this->GetFileName());
```
jhlegarreta, line 223: If this means that the file cannot be read, I think it should say so in the message and possibly trigger an exception. Maybe other maintainers can comment better on the warning vs. exception question.

jhlegarreta: If I understand correctly this situation is similar to this other one for PNG files:
https://github.com/InsightSoftwareConsortium/ITK/pull/2942/files#diff-1b5f81c245a7c32cc9d0efe02aac3fd4ca831e5b3c7a5cdaa63d575821b8ad58R359

And there an exception is being added.

Modules/IO/JPEG/test/itkJPEGImageIODegenerateCasesTest.cxx
issakomi: The issue with undefined behavior is fixed, incorrect _setjmp_ implementation caused it. It works in debug, release etc. the same way.

> See here the exceptions that are not being exercised (any line in red is not exercised/tested):
> https://open.cdash.org/viewCoverageFile.php?buildid=7620043&fileid=49347999

I can give you an image that will cover this part of code. This one, for example.
https://drive.google.com/file/d/1zGobe2UkAo1nEYMfEv5IV1TM5bhMYx7G/view?usp=sharing

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
issakomi: It is OK. At this point warning is sufficient, the image will be returned. There are different scenarios for different broken image, depends how JPEG library handles them, some causes a warning without jumping, some with, some will be refused, it depends.
The implementation is similar with Gimp and other, i think Gimp opened billions of JPEG images on all possible platforms and is reliable.

Modules/IO/JPEG/test/itkJPEGImageIODegenerateCasesTest.cxx
jhlegarreta: OK. Thanks. Then testing the exceptions would be best addressed in a separate PR. I will keep a note and propose a separate PR that uses the shared image (thanks for doing it) as time permits.

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
jhlegarreta: OK.

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
issakomi: IMHO, not, _itkDebugMacro_ is C++ code (`std::ostringstream itkmsg`), i would prefer to avoid C++ code here in `extern "C"`. Also name of function is line number are not informative in particular case.

issakomi: > But have not checked/tested in which circumstances the message should be printed.

The image used in test:
```
Corrupt JPEG data: premature end of data segment
Unsupported marker type 0x02
```
The second from the [post](https://github.com/InsightSoftwareConsortium/ITK/pull/2933#discussion_r767343167)
```
Corrupt JPEG data: 152 extraneous bytes before marker 0xda
Unsupported marker type 0x02
```

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
issakomi: In fact it is not important for me personally, for me is important to avoid crashes, they are imported into my app. If you like we can trigger exception here. But in fact, Gimp, Eye of Mate, also Google drive preview are showing thia image like we do now, probably the logic is "if possible to show partially, do it".

dzenanz:
I don't know whether this is ready for review and merge or not. If it is not, please convert it into a draft PR.

issakomi:
> I don't know whether this is ready for review and merge or not. If it is not, please convert it into a draft PR.

Yes, it is.

issakomi:
> I don't know whether this is ready for review and merge or not. If it is not, please convert it into a draft PR.

Yes, it is. There will be several more small PRs later, there are more issues unrelated to _setjmp_.

issakomi:
Just realized that in

```cpp
class JPEGFileWrapper
{
public:
  JPEGFileWrapper(const char * const fname, const char * const openMode)
    : m_FilePointer(nullptr)
  {
    m_FilePointer = fopen(fname, openMode);
  }

  virtual ~JPEGFileWrapper()
  {
    if (m_FilePointer != nullptr)
    {
      fclose(m_FilePointer);
    }
  }

  FILE * m_FilePointer;
};
```

`FILE * m_FilePointer;` probably would be better `FILE * volatile m_FilePointer;`

The class destructor may be triggered by _longjmp_. What do you think? Generally everything accessed during jumping should be _volatile_, AFAIK.

But not doing push now. I still have doubts: do you wish JPEG messages or not. If not, i shall change _JPEGIO_JPEG_MESSAGES_ to _0_

issakomi:
[itkPyBufferMemoryLeakTest](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Bridge/NumPy/wrapping/test/itkPyBufferMemoryLeakTest.py) with ITK.macOS.Python fails again. It happens with many PRs. Not related.

dzenanz (commented):
LGTM, but since I opened the PR I cannot formally approve it.

hjmjohnson (approved):
Approving based on @dzenanz LGTM statement :).

PR #2934, opened 2021-12-09
STYLE: Prefer itk::Math::abs over abs or fabs

hjmjohnson:
The itk::Math::abs is templated over more types than
std::abs (i.e. booleans and unsigned types), and therefore
provides more generality in templated code.

vim $(fgrep -R 'abs(' ../ITK | fgrep -v '::abs' | \
      fgrep -v fabs | fgrep -v ThirdParty |fgrep -v Remote | \
      fgrep -v cmake- |sed 's/:.*//g'|sort -u  )

Co-authored-by: Lee Newberg <lee.newberg@kitware.com>

This is a follow-up to #2932.  Prefer using itk::Math::abs which can be consistently applied to a large number of data types consistently.

## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) was made (or the changes have been approved)


jhlegarreta:
Thanks for doing this @hjmjohnson @Leengit. Looks like there are more appearances of `std::fabs` (at least) than the ones changed here, e.g.
https://github.com/InsightSoftwareConsortium/ITK/search?q=fabs

Were those missed by the regex?

Leengit:
From my Ubuntu 20.04 prompt I make all substitutions with:
```bash
cd ~/git/ITK
fgrep -Rl 'abs(' | egrep '\.(h|c)(xx|pp)?$' | fgrep -v ThirdParty |
  xargs sed -i -e 's/\([^A-Za-z:]\)\(::\)\?\([A-Za-z]\+::\)*\(f\|l\|ll\)\?abs\(f\|l\)\?(/\1itk::Math::abs(/g'
git diff --name-only | xargs clang-format-8 -i
```
I then selectively invoke `git restore` on those files that don't like the switch to `itk::Math::abs`.

I was instead substituting to `std::abs`, but am abandoning that in favor of this pull request.

hjmjohnson:
NOTE:  itk::Math::abs is delegated to vxl.  I made a recent change to vxl to use std::abs when available:   https://github.com/vxl/vxl/pull/874

jhlegarreta (approved):

dzenanz (approved):

jhlegarreta:
Looks like this change has generated failures in the `Ubuntu-g++-4.8.2-Release-System-VXL` build of `eldorado.kitware`:
https://open.cdash.org/viewBuildError.php?onlydeltap&buildid=7619302

Although I first thought that adding `#include <vnl/vnl_complex>` to `itkMath.h` would be the fix, maybe the issue is more the system `VXL` used. Not sure whether a new minimum `VXL` version should be bumped or how the change in this PR is meant to be backwards compatible otherwise.

hjmjohnson:
The minimum VXL version required will need to be bumped.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #2936, opened 2021-12-10
ENH: Spatial object to image filter preserve index

aylward:
The base class for SpatialObjectToImage filtering didn't use the index
of the reference image.  As a result, when, for example, rendering a
spatial object into an image, if the reference image had a non-zero
index, the generated image would not match the spatial extent of the
reference image.

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkSpatialObjectToImageFilter.h
+++ b/Modules/Core/SpatialObjects/include/itkSpatialObjectToImageFilter.h
@@ -50,6 +50,7 @@ public:
   using ConstPointer = SmartPointer<const Self>;
 
   using OutputImageType = TOutputImage;
+  using IndexType = typename OutputImageType::IndexType;
   using SizeType = typename OutputImageType::SizeType;
   using PointType = typename OutputImageType::PointType;
   using OutputImagePointer = typename OutputImageType::Pointer;
@@ -100,6 +101,7 @@ public:
     this->SetOrigin(refImage->GetOrigin());
     this->SetSpacing(refImage->GetSpacing());
     this->SetDirection(refImage->GetDirection());
+    this->SetIndex(refImage->GetLargestPossibleRegion().GetIndex());
     this->SetSize(refImage->GetLargestPossibleRegion().GetSize());
   }
 
@@ -161,6 +163,12 @@ public:
   virtual const double *
   GetOrigin() const;
 
+  /** The index of the output image. The index is the pixel
+   * coordinates of the image region.
+   * \sa GetSize() */
+  itkSetMacro(Index, IndexType);
+  itkGetConstMacro(Index, IndexType);
+
   /** The spatial object being transformed can be part of a hierarchy.
    * How deep in the hierarchy should we descend in generating the
    * image?  A ChildrenDepth of 0 means to only include the object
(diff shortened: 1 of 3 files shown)


aylward:
Will squash and merge, if approved.

Modules/Core/SpatialObjects/include/itkSpatialObjectToImageFilter.hxx
```diff
@@ -312,7 +311,27 @@ SpatialObjectToImageFilter<TInputSpatialObject, TOutputImage>::GenerateData()
...
     region.SetSize(size);
   }
-  region.SetIndex(index);
+
+  IndexType index;
+  index.Fill(0);
+  specified = false;
+  for (i = 0; i < OutputImageDimension; ++i)
+  {
+    if (m_Index[i] != 0)
```
dzenanz, line 320: This logic seems entirely unnecessary. If `SetIndex` has not been invoked, `m_Index` will be all zeroes because it was default initialized to that, and we don't need an additional `index` variable of all zeroes. If `m_Index` was set, we want to use it. So we can directly use `m_Index` in either case. So the entire set of changes for lines 314 to 334 should be replaced by a single line: `region.SetIndex(m_Index);`.

Modules/Core/SpatialObjects/test/itkSpatialObjectToImageFilterTest.cxx
```diff
@@ -70,6 +70,12 @@ itkSpatialObjectToImageFilterTest(int, char *[])
   imageFilter->SetChildrenDepth(childrenDepth);
   ITK_TEST_SET_GET_VALUE(childrenDepth, imageFilter->GetChildrenDepth());
 
+  ImageType::IndexType indx;
+  indx[0] = 10;
+  indx[1] = 10;
+  imageFilter->SetIndex(indx);
```
dzenanz, line 76: Doesn't this necessitate changes further down in the test?

Modules/Core/SpatialObjects/include/itkSpatialObjectToImageFilter.hxx
aylward: :) - you're right!   Got carried away with cut-n-paste.   Will fix.

Modules/Core/SpatialObjects/include/itkSpatialObjectToImageFilter.h
```diff
@@ -188,6 +196,7 @@ class ITK_TEMPLATE_EXPORT SpatialObjectToImageFilter : public ImageSource<TOutpu
   void
   GenerateData() override;
 
+  IndexType     m_Index;
   SizeType      m_Size;
```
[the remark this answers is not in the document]
aylward: That is a line that wasn't touching in this pull request, and there are many other member variables that would also need to be updated throughout this and many other classes to have a new standard of initializing member variables at declaration instead of in the constructor.   Right now it would be confusing to devs to have some variables initialized in the .h and others in the constructor - or potentially having constructor initializations being different than the .h.   So, for now, keeping as is.

Modules/Core/SpatialObjects/test/itkSpatialObjectToImageFilterTest.cxx
aylward: That is what I initially thought too, but changing the index of a region only changes the spatial extent covered by it, and the physical and index position of the "object" spanned by that region is unchanged.  So, the test was actually already verifying the usage of a region's index in the most thorough way (writing data in physical space and verifying it in index space), and yet no changes were needed.

dzenanz (approved):

PR #2941, opened 2021-12-11
COMP: Remove implicit conversions

hjmjohnson:
The implicit conversion for computing the P2 variable
using unsigned integers occurs if directly converting
to floating-point values.  This caused underflow
computation errors.

Explicitly convert values to consistently use
floating-point computations to avoid many implicit
type conversions between signed and unsigned values.


hjmjohnson:
Failing test itkPyBufferMemoryLeakTest seems unrelated to this change.

jhlegarreta (approved):

PR #2953, opened 2021-12-14
ENH: Python Wrapping for Array PixelType in Mesh

PranjalSahu:
Adding Python Wrapping for Mesh with Array PixelType. 
This is needed for the Osteoarthritis project where multiple attributes (variable length) will be stored 
for each point in the mesh. Alternative for this would be VariableLengthVector but we don't see any use case where direct vector computation would be performed on the pixel data. Wrapping for CellInterface is only done for Double.

This does not throw any warnings.

## PR Checklist


dzenanz (approved):
LGTM. Does this wrapping correctly enable the desired abilities in osteoarthritis code?

PranjalSahu:
I want to add small test case for this in Python. That will demonstrate the usage also.
Is this folder the right place for this ?
[Core/Common/Wrapping/test](https://github.com/InsightSoftwareConsortium/ITK/tree/master/Modules/Core/Common/wrapping/test)

dzenanz:
Yes, that is the right place for such a test.

hjmjohnson (approved):

jhlegarreta:
@PranjalSahu looks like warnings were introduced by this commit in the Windows Python builds:
https://dev.azure.com/itkrobotwindowpython/ITK.Windows.Python/_build?definitionId=1&_a=summary

PranjalSahu:
I am also checking the issue.

PranjalSahu:
So I looked at another build before that commit ([ITK.Windows.Python](https://dev.azure.com/itkrobotwindowpython/ITK.Windows.Python/_build/results?buildId=2721&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=77aad734-2057-5694-9ae2-ee1f5f26eae8&l=3394)) which was successful  and it also has the same warning message. So that's not the issue. It is someplace else. Checking !

PR #2958, opened 2021-12-15
Remove circular header includes

hjmjohnson:
STYLE: Remove recursive inclusion of .h files
    
The .h files include the .hxx files, which in
turn inclue the original .h file.
    
Break recursive inclusion of the .h files from the .hxx files.
    
This should avoid reading the .h file multiple times.

## PR Checklist


SimonRit:
I am in favor of this change but I believe that `ITKModuleHeaderTest.cmake` should be modified to avoid inclusion of `.hxx` files. See for example [this patch](https://github.com/SimonRit/RTK/commit/4f017a1581f668f5f5c33e0dfb7dee452fcff9e2).

dzenanz (approved):
LGTM. Only spelling corrections needed.

SimonRit (approved):
Thanks for taking into account my suggestion. LGTM. I now think that all `_hxx` guards in .hxx files can be removed since .hxx files will only be included within corresponding `_h` guards.

dzenanz:
@SimonRit unless `USE_MANUAL_INSTANTIATION` is defined, and then `.hxx` file needs to be explicitly added to some list of inclusions. Which might not be a big problem? It needs weighing against simplification of not needing include guards in `.hxx` files.

SimonRit:
I was under the impression that it would then come down to add guards to .cxx files... but clearly I don't understand the manual instantiation part of ITK so I fully trust you!

hjmjohnson:
Note.  Mac failure fixed in different PR.

PR #2965, opened 2021-12-15
COMP: Duplicate symbol name with #define

hjmjohnson:
The class name and the #define symbol must have
different values to avoid preprocessing
the name to an invalid value.


hjmjohnson:
This error only show up with Example builds on.

dzenanz (approved):

PR #2967, opened 2021-12-15
BUG: MeshCellVisitor Examples failing: #define used class name

aylward:
The compiler #ifndef/#define directive used to avoid redundant
definition of a class used a variable that was the
same as the class name in MeshCellVisitor and MeshCellVisitor2
examples.  As a result, the actual class name in the code was getting
overwritten by the compiler directive.

The errors reported by VS2019 were
C:\src\ITK\Examples\DataRepresentation\Mesh\MeshCellVisitor.cxx(100): error C2059: syntax error: ')'
and
C:\src\ITK\Examples\DataRepresentation\Mesh\MeshCellVisitor2.cxx(187): error C2059: syntax error: '('


dzenanz:
This is a duplicate of #2965.

PR #2975, opened 2021-12-17
BUG: updated JPEG IO

issakomi:
Commits are intentionally not squashed.

Follow-up of https://github.com/InsightSoftwareConsortium/ITK/pull/2933

### DIFF
--- a/Modules/IO/JPEG/src/itkJPEGImageIO.cxx
+++ b/Modules/IO/JPEG/src/itkJPEGImageIO.cxx
@@ -247,7 +247,7 @@ JPEGImageIO::JPEGImageIO()
 #if BITS_IN_JSAMPLE == 8
   m_ComponentType = IOComponentEnum::UCHAR;
 #else
-  m_ComponentType = IOComponentEnum::UINT;
+#  error
 #endif
   m_UseCompression = false;
   this->Self::SetQuality(95);
--- a/Modules/IO/JPEG/src/itkJPEGImageIO.cxx
+++ b/Modules/IO/JPEG/src/itkJPEGImageIO.cxx
@@ -435,20 +435,23 @@ JPEGImageIO::WriteSlice(std::string & fileName, const void * const buffer)
     itkExceptionMacro(<< "JPEG: too many components");
   }
 
+  auto * volatile row_pointers = new JSAMPROW[height];
+
   struct itk_jpeg_error_mgr   jerr;
   struct jpeg_compress_struct cinfo;
   cinfo.err = jpeg_std_error(&jerr.pub);
+
   // set the jump point
   if (setjmp(jerr.setjmp_buffer))
   {
     jpeg_destroy_compress(&cinfo);
+    delete[] row_pointers;
     itkExceptionMacro(<< "JPEG error, failed to write " << fileName);
   }
 
   jpeg_create_compress(&cinfo);
 
   // set the destination file
-  // struct jpeg_destination_mgr compressionDestination;
   jpeg_stdio_dest(&cinfo, fp);
 
   cinfo.image_width = static_cast<JDIMENSION>(width);
@@ -465,11 +468,12 @@ JPEGImageIO::WriteSlice(std::string & fileName, const void * const buffer)
       break;
     default:
       cinfo.in_color_space = JCS_UNKNOWN;
+      itkWarningMacro("Image may be saved incorrectly as JPEG");
       break;
   }
 
   // set the compression parameters
-  jpeg_set_defaults(&cinfo); // start with reasonable defaults
+  jpeg_set_defaults(&cinfo);
   jpeg_set_quality(&cinfo, this->GetQuality(), TRUE);
   if (m_Progressive)
   {
@@ -507,29 +511,26 @@ JPEGImageIO::WriteSlice(std::string & fileName, const void * const buffer)
   // start compression
   jpeg_start_compress(&cinfo, TRUE);
 
-  volatile const JSAMPLE * outPtr = ((const JSAMPLE *)buffer);
-
-  // write the data. in jpeg, the first row is the top row of the image
-  auto *    row_pointers = new JSAMPROW[height];
-  const int rowInc = numComp * width;
-  for (unsigned int ui = 0; ui < height; ++ui)
   {
-    row_pointers[ui] = const_cast<JSAMPROW>(outPtr);
-    outPtr = const_cast<JSAMPLE *>(outPtr) + rowInc;
+    const auto * ptr = static_cast<const JSAMPLE *>(buffer);
+    const auto   rowbytes = num_comp * width;
+    for (size_t ui = 0; ui < height; ++ui)
+    {
+      row_pointers[ui] = const_cast<JSAMPLE *>(ptr) + rowbytes * ui;
+    }
   }
-  jpeg_write_scanlines(&cinfo, row_pointers, height);
-
-  if (fflush(fp) == EOF)
+  while (cinfo.next_scanline < cinfo.image_height)
   {
-    delete[] row_pointers;
-    itkExceptionMacro(<< "JPEG : Out of disk space");
+    // usually one iteration
+    const auto remaining = cinfo.image_height - cinfo.next_scanline;
+    jpeg_write_scanlines(&cinfo, &row_pointers[cinfo.next_scanline], remaining);
   }
 
   // finish the compression
   jpeg_finish_compress(&cinfo);
 
-  // clean up and close the file
-  delete[] row_pointers;
+  // clean up
   jpeg_destroy_compress(&cinfo);
+  delete[] row_pointers;
 }
 } // end namespace itk
--- a/Modules/IO/JPEG/src/itkJPEGImageIO.cxx
+++ b/Modules/IO/JPEG/src/itkJPEGImageIO.cxx
@@ -406,11 +406,23 @@ JPEGImageIO::WriteSlice(std::string & fileName, const void * const buffer)
   // use this class so return will call close
   JPEGFileWrapper JPEGfp(fileName.c_str(), "wb");
   FILE *          fp = JPEGfp.m_FilePointer;
-
   if (!fp)
   {
     itkExceptionMacro("Unable to open file " << fileName << " for writing." << std::endl
                                              << "Reason: " << itksys::SystemTools::GetLastSystemError());
+  }
+
+  // set the information about image
+  const SizeValueType width = m_Dimensions[0];
+  const SizeValueType height = m_Dimensions[1];
+  if (width > JPEG_MAX_DIMENSION || height > JPEG_MAX_DIMENSION)
+  {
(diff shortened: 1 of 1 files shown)


Modules/IO/JPEG/src/itkJPEGImageIO.cxx
```diff
@@ -390,9 +397,9 @@ JPEGImageIO::Write(const void * buffer)
     itkExceptionMacro(<< "JPEG Writer can only write 2-dimensional images");
   }
 
-  if (this->GetComponentType() != IOComponentEnum::UCHAR && this->GetComponentType() != IOComponentEnum::UINT)
```
issakomi, line 393: This is **bug**. Tried _UINT_ 2D image as input, got garbage, of course.

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
```diff
@@ -427,31 +439,19 @@ JPEGImageIO::WriteSlice(std::string & fileName, const void * const buffer)
...
 
-  // set the information about image
-  const SizeValueType width = m_Dimensions[0];
-  const SizeValueType height = m_Dimensions[1];
-
   // The JPEG standard only supports images up to 64K*64K due to 16-bit fields
   // in SOF markers.
   cinfo.image_width = width;
   cinfo.image_height = height;
-  if (cinfo.image_width > 65536 || cinfo.image_height > 65536)
```
issakomi, line 438: This is wrong. JPEG defines JPEG_MAX_DIMENSION in jmorecfg.h
`#define JPEG_MAX_DIMENSION  65500L  /* a tad under 64K to prevent overflows */`
External/system library may have modified jmorecfg.h too.

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
```diff
@@ -427,31 +439,19 @@ JPEGImageIO::WriteSlice(std::string & fileName, const void * const buffer)
...
 
-  cinfo.input_components = this->GetNumberOfComponents();
-  const unsigned int numComp = this->GetNumberOfComponents();
-
-  // Maximum number of components (color channels) allowed in JPEG image.
-  // JPEG spec set this to 255. However ijg default it to 10.
-  if (cinfo.input_components > 255)
-  {
-    itkExceptionMacro(<< "JPEG : Too many components for JPEG");
-  }
```
issakomi, line 451: Useless, s. below.

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
```diff
@@ -247,7 +247,7 @@ JPEGImageIO::JPEGImageIO()
 #if BITS_IN_JSAMPLE == 8
   m_ComponentType = IOComponentEnum::UCHAR;
 #else
-  m_ComponentType = IOComponentEnum::UINT;
```
issakomi, line 250: Doesn't work at all.

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
```diff
@@ -205,6 +201,10 @@ JPEGImageIO::Read(void * buffer)
   // read the header
   jpeg_read_header(&cinfo, TRUE);
 
+  // jpeg_calc_output_dimensions used in ReadImageInformation,
+  // so has to used here too
+  jpeg_calc_output_dimensions(&cinfo);
+
```
issakomi, line 207: _jpeg_calc_output_dimensions_ is used in _ReadImageInformation_ before, have to be used here too to ensure same parameters.

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
```diff
@@ -334,14 +338,17 @@ JPEGImageIO::ReadImageInformation()
...
-    case 2:
-      m_PixelType = IOPixelEnum::VECTOR;
-      break;
     case 3:
       m_PixelType = IOPixelEnum::RGB;
       break;
     case 4:
+      // TODO
       m_PixelType = IOPixelEnum::RGBA;
+      itkWarningMacro("JPEG image may not be opened correct");
```
issakomi, line 347: 4 components JPEGs are unlikely, probably never RGBA. Mostly CMYK, rare, [here](https://data.kitware.com/api/v1/item/61bc65c34acac99f42e67a5c/download?contentDisposition=inline) is one example. Opened with this IO it looks more "pale" as in other viewers.

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
```diff
@@ -337,14 +337,17 @@ JPEGImageIO::ReadImageInformation()
...
-    case 2:
-      m_PixelType = IOPixelEnum::VECTOR;
-      break;
     case 3:
       m_PixelType = IOPixelEnum::RGB;
       break;
     case 4:
+      // TODO
       m_PixelType = IOPixelEnum::RGBA;
+      itkWarningMacro("JPEG image may be opened not correclty");
```
dzenanz: opened not correclty ->opened incorrectly

Same for line 250

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
```diff
@@ -508,29 +507,26 @@ JPEGImageIO::WriteSlice(std::string & fileName, const void * const buffer)
...
-  // write the data. in jpeg, the first row is the top row of the image
-  auto *    row_pointers = new JSAMPROW[height];
-  const int rowInc = numComp * width;
-  for (unsigned int ui = 0; ui < height; ++ui)
   {
-    row_pointers[ui] = const_cast<JSAMPROW>(outPtr);
-    outPtr = const_cast<JSAMPLE *>(outPtr) + rowInc;
+    const unsigned int rowbytes = numComp * width;
+    unsigned int       count_rows = 0;
+    while (cinfo.next_scanline < cinfo.image_height)
```
dzenanz, line 513: This should avoid allocating almost an entire image worth of temporary memory, correct? Just 1 temp scanline instead of all of them?

dzenanz (commented):
Looks good.

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
issakomi: Updated, now there is still _while_ loop, but almost always will run in 1 iteration, very close as it was before, but safer if the lib will not be able to write all lines at once. The reason for change was that there was possible memory leak, pointer was allocated after _setjmp_ point.

> This should avoid allocating almost an entire image
Allocation is not big: image height * size of pointer

P.S.
Previous version, with 1 line per call, was also correct, all reference apps doing so, even if JPEG recommends to do less calls. But probably now is better.

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
```diff
@@ -334,14 +338,17 @@ JPEGImageIO::ReadImageInformation()
...
-    case 2:
-      m_PixelType = IOPixelEnum::VECTOR;
-      break;
     case 3:
       m_PixelType = IOPixelEnum::RGB;
       break;
     case 4:
+      // TODO
       m_PixelType = IOPixelEnum::RGBA;
+      itkWarningMacro("JPEG image may be opened incorrectly");
```
issakomi, line 347: 4 components JPEGs are unlikely, probably never, RGBA. Mostly CMYK, rare, here is one example. Opened with this IO it looks more "pale" as in other viewers. Looks the same if opened as RBG, BTW. More work for correct conversion from CMYK is required (example in e.g. Gimp's jpeg-load.c) and will be very rarely used.

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
```diff
@@ -506,29 +504,26 @@ JPEGImageIO::WriteSlice(std::string & fileName, const void * const buffer)
...
+    const auto * ptr = static_cast<const JSAMPLE *>(buffer);
+    const auto   rowbytes = num_comp * width;
+    for (size_t ui = 0; ui < height; ++ui)
+    {
+      row_pointers[ui] = const_cast<JSAMPLE *>(ptr) + rowbytes * ui;
+    }
   }
-  jpeg_write_scanlines(&cinfo, row_pointers, height);
-
-  if (fflush(fp) == EOF)
```
issakomi, line 521: I have check, fflush is called [here](https://github.com/InsightSoftwareConsortium/ITK/blob/a89145bccda6a36f42cfdd45d3a6b27234ff54fe/Modules/ThirdParty/JPEG/src/itkjpeg/jdatadst.c#L173) in `term_destination (j_compress_ptr cinfo) `called by `jpeg_finish_compress(&cinfo);`

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
issakomi: Here is how Gimp handles this
```
  switch (cinfo.output_components)
    {
    case 1:
      image_type = GIMP_GRAY;
      layer_type = GIMP_GRAY_IMAGE;
      break;

    case 3:
      image_type = GIMP_RGB;
      layer_type = GIMP_RGB_IMAGE;
      break;

    case 4:
      if (cinfo.out_color_space == JCS_CMYK)
        {
          image_type = GIMP_RGB;
          layer_type = GIMP_RGB_IMAGE;
          break;
        }
      /*fallthrough*/

    default:
      g_message ("Don't know how to load JPEG images "
                 "with %d color channels, using colorspace %d (%d).",
                 cinfo.output_components, cinfo.out_color_space,
                 cinfo.jpeg_color_space);
      return -1;
      break;
    }
```

issakomi: CMYK is already done, will be added in separate PR. The files are not unusual, Photoshop has the option to export CMYK jpegs. Current "RGBA" part is bug.

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
```diff
@@ -247,7 +247,7 @@ JPEGImageIO::JPEGImageIO()
 #if BITS_IN_JSAMPLE == 8
   m_ComponentType = IOComponentEnum::UCHAR;
 #else
-  m_ComponentType = IOComponentEnum::UINT;
+#  error
```
dzenanz, line 249: We should probably add a [message](https://en.cppreference.com/w/cpp/preprocessor/error) here. Maybe "JPEG files with more than 8 bits per sample are not supported"?

dzenanz (approved):

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
issakomi: Done.

dzenanz (approved):

thewtex (approved):
@issakomi excellent, thank you!

PR #2977, opened 2021-12-18
COMP: seems all use of kwiml is no longer actually needed, removed it.

seanm:


seanm:
@hjmjohnson @dzenanz @thewtex I didn't yet `git rm` all of kwiml, but seems we can...

dzenanz (approved):

seanm:
Oh, I didn't mean for this to be merged so fast...

So I guess you concur that nothing needs kwiml?  Is that surprising or not?  I guess I can just delete the folder in ThirdParty too then?

dzenanz:
I don't remember hearing of KWIML before this, so I assume nothing needs it 😄

dzenanz:
Maybe if was needed before we could rely on C++11. But now we require C++14, so I guess try deleting it and if nothing breaks we will know we don't need it 😄

seanm:
@dzenanz see https://github.com/InsightSoftwareConsortium/ITK/pull/2982

PR #2980, opened 2021-12-20
BUG: GDCM IO: reset internal variables

issakomi:
It is safer to reset internal variables on _InternalReadImageInformation_,
because a user can re-use the IO.

Example of such re-using from this [post](https://discourse.itk.org/t/encountering-undefined-behavior-when-writing-dicom-images/4587?u=mihail.isakov)

```
	ImageIOType::Pointer dcmIOsingleFiles = ImageIOType::New();

	// Read files one by one
	// Read all headers of all DICOM files in fileNames
	for (std::string filename : fileNames) {
		dcmIOsingleFiles->SetFileName(filename);
		dcmIOsingleFiles->ReadImageInformation();
```

### DIFF
--- a/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
+++ b/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
@@ -441,6 +441,11 @@ GDCMImageIO::Read(void * pointer)
 void
 GDCMImageIO::InternalReadImageInformation()
 {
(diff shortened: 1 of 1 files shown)


issakomi:
Maybe more variable could be examined. Also many style changes can be done in the IO. If required, please do it yourself later, sorry. I don't use the IO and don't have time now.

dzenanz (approved):

PR #2982, opened 2021-12-20
Remove KWIML third party code

seanm:
Seems like it was only needed pre-C++11.


Modules/Bridge/VtkGlue/itk-module-init.cmake
```diff
@@ -52,7 +52,7 @@ set(_required_vtk_libraries
   )
 if(ITK_WRAP_PYTHON)
   list(APPEND _required_vtk_libraries ${_target_prefix}WrappingPythonCore
-    ${_target_prefix}CommonCore ${_target_prefix}CommonDataModel ${_target_prefix}kwiml ${_target_prefix}CommonExecutionModel)
+    ${_target_prefix}CommonCore ${_target_prefix}CommonDataModel ${_target_prefix}CommonExecutionModel)
```
dzenanz, line 55: This commit should be broken into two. First one to remove the remaining references to KWIML, the second one to remove KWIML itself.

dzenanz (approved):

Modules/Bridge/VtkGlue/itk-module-init.cmake
seanm: OK. But why?

dzenanz: Those are two logical changes, and it is preferable that each change has its own commit. In this case it is easy to separate them, but sometimes it is not. In spirit, this is similar to Third party updates: one commit to change the branch name or hash, one commit which adds the actual third party changes which can be big and usually crowd out all other changes.

seanm: Done.

dzenanz (approved):

thewtex (approved):
Great!

PR #2984, opened 2021-12-22
BUG: Runge-Kutta wrong TimeBounds implementation

josempozo:
The aimed normalization of the time apparently messed up the Runge-Kutta 
implementation resulting in a wrong algorithm. The implementation has been corrected to remove this bug. 

The existing test (itkTimeVaryingVelocityFieldIntegrationImageFilterTest) only tested for the integration of a homogeneous and constant velocity field, where the error could not be detected.
I have extended the test adding the case of a non-trivial (inhomogeneous and time-variaying) velocity field. This test is not passed by the previous implementation.

Added boolean flag to control the interpretation of LowerTimeBound and UpperTimeBound as rates 
in the velocity field time span. If true, it is equivalent to consider that the velocity field is defined in the normalized time interval [0, 1], ignoring the input origin and spacing in the temporal dimension.

This normalization was the previous behaviour. Thus, the default is true for backwards compatibility.

### DIFF
--- a/Modules/Filtering/DisplacementField/include/itkTimeVaryingVelocityFieldIntegrationImageFilter.h
+++ b/Modules/Filtering/DisplacementField/include/itkTimeVaryingVelocityFieldIntegrationImageFilter.h
@@ -144,6 +144,18 @@ public:
    */
   itkGetConstMacro(NumberOfIntegrationSteps, unsigned int);
 
+  /**
+   * Get/Set a flag to interpret LowerTimeBound and UpperTimeBound as rates
+   * in the velocity field time span. This is equivalent to consider that the velocity
+   * field is defined in the normalized time interval [0, 1], ignoring the input origin
+   * and spacing in the temporal dimension.
+   *
+   * The default is true for backwards compatibility.
+   */
+  itkSetMacro(TimeBoundsAsRates, bool);
+  itkGetConstMacro(TimeBoundsAsRates, bool);
+  itkBooleanMacro(TimeBoundsAsRates);
+
 protected:
   TimeVaryingVelocityFieldIntegrationImageFilter();
   ~TimeVaryingVelocityFieldIntegrationImageFilter() override = default;
@@ -174,6 +186,8 @@ protected:
   unsigned int m_NumberOfTimePoints;
 
   DisplacementFieldInterpolatorPointer m_DisplacementFieldInterpolator;
+
+  bool m_TimeBoundsAsRates{ true };
 
 private:
   VelocityFieldInterpolatorPointer m_VelocityFieldInterpolator;
--- a/Modules/Filtering/DisplacementField/include/itkTimeVaryingVelocityFieldIntegrationImageFilter.hxx
+++ b/Modules/Filtering/DisplacementField/include/itkTimeVaryingVelocityFieldIntegrationImageFilter.hxx
@@ -179,11 +179,6 @@ TimeVaryingVelocityFieldIntegrationImageFilter<TTimeVaryingVelocityField, TDispl
     }
   }
 
-  if (this->m_UpperTimeBound == this->m_LowerTimeBound)
-  {
-    return displacement;
-  }
-
   // Perform the integration
   // Need to know how to map the time dimension of the input image to the
   // assumed domain of [0,1].
--- a/Modules/Filtering/DisplacementField/include/itkTimeVaryingVelocityFieldIntegrationImageFilter.hxx
+++ b/Modules/Filtering/DisplacementField/include/itkTimeVaryingVelocityFieldIntegrationImageFilter.hxx
@@ -179,42 +179,40 @@ TimeVaryingVelocityFieldIntegrationImageFilter<TTimeVaryingVelocityField, TDispl
     }
   }
 
-  // Perform the integration
-  // Need to know how to map the time dimension of the input image to the
-  // assumed domain of [0,1].
+  // Perform the integration.
+  // With TimeBoundsAsRates On, we need to map the time dimension of the input image to the
+  // normalized domain of [0,1].
 
-
-  typename TimeVaryingVelocityFieldType::PointType spaceTimeOrigin = inputField->GetOrigin();
-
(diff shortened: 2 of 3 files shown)


Modules/Filtering/DisplacementField/test/itkTimeVaryingVelocityFieldIntegrationImageFilterTest.cxx
```diff
@@ -64,7 +64,7 @@ itkTimeVaryingVelocityFieldIntegrationImageFilterTest(int, char *[])
   index.Fill(0);
   VectorType displacement;
 
-  auto = IntegratorType::New();
+  auto inverseIntegrator = IntegratorType::New();
```
dzenanz, line 67: This commit should be squashed with the one which introduced the use of `auto` here.

Modules/Filtering/DisplacementField/test/itkTimeVaryingVelocityFieldIntegrationImageFilterTest.cxx
```diff
@@ -173,6 +173,10 @@ itkTimeVaryingVelocityFieldIntegrationImageFilterTest(int, char *[])
   integrator->SetInput(timeVaryingVelocityField);
   integrator->SetLowerTimeBound(0.2);
   integrator->SetUpperTimeBound(0.8);
+  /* This time bounds are meant to be absolute
```
dzenanz, line 176: This commit should be squashed with the one which introduced this option.

Modules/Filtering/DisplacementField/test/itkTimeVaryingVelocityFieldIntegrationImageFilterTest.cxx
```diff
@@ -123,7 +123,7 @@ itkTimeVaryingVelocityFieldIntegrationImageFilterTest(int, char *[])
   size[0] = 3;
   size[1] = 3;
   size[2] = 401;
-  size[3] = 41;
+  size[3] = 61;
```
dzenanz, line 126: Commit message should be updated to say "This a a modification to the time-varying test".

Modules/Filtering/DisplacementField/include/itkTimeVaryingVelocityFieldIntegrationImageFilter.h
```diff
@@ -144,6 +144,18 @@ class ITK_TEMPLATE_EXPORT TimeVaryingVelocityFieldIntegrationImageFilter
    */
   itkGetConstMacro(NumberOfIntegrationSteps, unsigned int);
 
+  /**
```
dzenanz, line 147: This commit should be squashed with the previous one.

Modules/Filtering/DisplacementField/include/itkTimeVaryingVelocityFieldIntegrationImageFilter.hxx
```diff
@@ -18,8 +18,6 @@
 #ifndef itkTimeVaryingVelocityFieldIntegrationImageFilter_hxx
 #define itkTimeVaryingVelocityFieldIntegrationImageFilter_hxx
 
-#include "itkTimeVaryingVelocityFieldIntegrationImageFilter.h"
```
dzenanz: This should be squashed with the first commit.

Modules/Filtering/DisplacementField/test/itkTimeVaryingVelocityFieldIntegrationImageFilterTest.cxx
```diff
@@ -99,5 +102,113 @@ itkTimeVaryingVelocityFieldIntegrationImageFilterTest(int, char *[])
     return EXIT_FAILURE;
   }
 
+
+  /* Second test with inhomogeneous and time-varying velocity field
```
dzenanz, line 106: This is an awesome improvement to the test!

dzenanz (approved):
Looks great. Some commits should be squashed. Thank you for this contribution.

Modules/Filtering/DisplacementField/test/itkTimeVaryingVelocityFieldIntegrationImageFilterTest.cxx
josempozo: I'm not sure how to change this message now.

But I could merge the commit with the previous one. Do you think this is a good idea?

Modules/Filtering/DisplacementField/test/itkTimeVaryingVelocityFieldIntegrationImageFilterTest.cxx
josempozo: Thanks.

Modules/Filtering/DisplacementField/test/itkTimeVaryingVelocityFieldIntegrationImageFilterTest.cxx
dzenanz: Here are some explanations of how to change a commit message:
https://stackoverflow.com/questions/1884474/change-old-commit-message-using-git-rebase
If you use some GUI tool such as TortoiseGit, it is even easier.

It is better not to merge, as making time normalization optional, and exercising that option are two different things, and it is best to have them as separate commits.

josempozo: Thanks for the pointer. I think that I have resolved it now. Let me know if it looks correct from your side.

dzenanz: You had two branches (one with original message, one with updated message) which were merged at the end, producing a strange looking history when linarized by GitHub. I picked the correct one and pushed it your branch. It is ready for merging now. Let's wait until CI finishes before doing that.

josempozo: Thanks for solving that merge.

PR #2939, opened 2021-12-11
STYLE: Use the `itkNameOfTestExecutableMacro` macro in tests

jhlegarreta:
- STYLE: Use the `itkNameOfTestExecutableMacro` macro in tests 
- STYLE: Use `argc` and `argv` to name test argument variables 

## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [X] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)

PR #2985, opened 2021-12-22
STYLE: Use the `itkNameOfTestExecutableMacro` macro in tests

jhlegarreta:
Use the `itkNameOfTestExecutableMacro` macro to get the test name
instead of hard-coding the test name.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/test/itkCMakeConfigurationTest.cxx
+++ b/Modules/Core/Common/test/itkCMakeConfigurationTest.cxx
@@ -35,6 +35,7 @@
 #include <sys/stat.h>
 #include <ctime>
 #include <cstring>
+#include "itkTestingMacros.h"
 
 void
 itkCMakeInformationPrintFile(const char * name, std::ostream & os)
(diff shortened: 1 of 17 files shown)


jhlegarreta:
Left behind in PR #2939.

A couple of additional notes:
- Not sure if this is the right move for `itkCMakeConfigurationTest`, as it does not depend on ITK. Insights are welcome.
- Some of the changes (added the `Missing parameters` message, changing `std::cout` to `std::cerr`) were only made in cases where the whole check was changed in order to restrict the number of changes to what it made sense as the necessary minimum. Making the style of the checking consistent beyond these can be done in a separate PR.

Modules/Core/Common/test/itkCMakeConfigurationTest.cxx
```diff
@@ -74,7 +75,7 @@ main(int argc, char * argv[])
 {
   if (argc != 2)
   {
-    std::cerr << "Usage: itkCMakeInformationTest <top-of-build-tree>\n";
+    std::cerr << "Usage: " << itkNameOfTestExecutableMacro(argv) << " <top-of-build-tree>\n";
```
dzenanz, line 78: If this does not compile, renaming into `itkCMakeConfigurationTest` should be done.

dzenanz (approved):

Modules/Core/Common/test/itkCMakeConfigurationTest.cxx
dzenanz: Oh, it is used here:
https://github.com/InsightSoftwareConsortium/ITK/blob/24c3ad91db624592318c886c8d81614e0662ea72/Modules/Core/Common/test/CMakeLists.txt#L557-L560
and it did not get broken. All OK!

thewtex (approved):
:green_heart:

PR #2983, opened 2021-12-21
ENH: Make Parameters empty, do not warn in Transform default-constructor

N-Dekker:
Defaulted (`= default`) the `Transform` default-constructor, ensuring
that it will make both its `Parameters` and its `FixedParameters` empty.

Allowed calling this default-constructor from a derived transform class
without getting a `itkWarningMacro`.

PR #2986, opened 2021-12-22
STYLE: Use `Transform` default-constructor instead of `Transform(0)`

N-Dekker:
Adjusted the default-constructor of `IdentityTransform`,
`MINCTransformAdapter`, and `ProjectTransform` to use the
default-constructor of `Transform`, instead of its
`Transform(NumberOfParametersType)` constructor, as in each of these
cases, both `Parameters` and `FixedParameters` should just be empty.

Follow-up to:
pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2983
commit 20249ece000a22347dd363df4cdeb6fccdedaae4
"ENH: Make Parameters empty, do not warn in Transform default-constructor"

### DIFF
--- a/Modules/Core/Transform/include/itkIdentityTransform.h
+++ b/Modules/Core/Transform/include/itkIdentityTransform.h
@@ -240,10 +240,7 @@ public:
   {}
 
 protected:
-  IdentityTransform()
-    : Transform<TParametersValueType, NDimensions, NDimensions>(0)
-  {}
-
+  IdentityTransform() = default;
   ~IdentityTransform() override = default;
 };
 } // end namespace itk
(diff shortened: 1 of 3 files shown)


dzenanz (approved):

thewtex (approved):
:santa:

N-Dekker:
> 🎅

Thanks Matt  By default, I would suggest that the default-constructor of `SantaClaus` should simply (and preferably implicitly) call the corresponding default-constructor its `BaseClaus`  🤶🤶🤶

thewtex:
@N-Dekker yes, those optimizations will help :santa: get all those :gift: 's delivered!

dzenanz (approved):

PR #2987, opened 2021-12-22
ENH: Adding Python Wrapping for SyNImageRegistrationMethod for Mesh

PranjalSahu:
Needed for ThinShellDemons Registration. Example usage are added in the ITKThinShellDemons repo.

## PR Checklist

### DIFF
--- a/Modules/Registration/RegistrationMethodsv4/wrapping/itkSyNImageRegistrationMethod.wrap
+++ b/Modules/Registration/RegistrationMethodsv4/wrapping/itkSyNImageRegistrationMethod.wrap
@@ -5,7 +5,7 @@ itk_wrap_class("itk::SyNImageRegistrationMethod" POINTER)
     foreach(t ${WRAP_ITK_REAL})
       itk_wrap_template("${ITKM_${t}}${d}${ITKM_${t}}${d}"
                         "itk::Image< ${ITKT_${t}}, ${d} >, itk::Image< ${ITKT_${t}}, ${d} >")
-      itk_wrap_template("${ITKM_${t}}${d}${ITKM_${t}}${d}BSOUDFT${ITKT_D}${d}"
+      itk_wrap_template("${ITKM_${t}}${d}${ITKM_${t}}${d}BSOUDFT${ITKM_D}${d}"
                         "itk::Image< ${ITKT_${t}}, ${d} >, itk::Image< ${ITKT_${t}}, ${d} >, itk::BSplineSmoothingOnUpdateDisplacementFieldTransform< ${ITKT_D}, ${d} >")
     endforeach()
   endforeach()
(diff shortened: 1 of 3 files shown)


Modules/Registration/RegistrationMethodsv4/wrapping/itkSyNImageRegistrationMethod.wrap
```diff
@@ -1,12 +1,15 @@
 itk_wrap_include("itkBSplineSmoothingOnUpdateDisplacementFieldTransform.h")
+itk_wrap_include("itkMesh.h")
 
 itk_wrap_class("itk::SyNImageRegistrationMethod" POINTER)
   foreach(d ${ITK_WRAP_DIMS})
     foreach(t ${WRAP_ITK_REAL})
       itk_wrap_template("${ITKM_${t}}${d}${ITKM_${t}}${d}"
                         "itk::Image< ${ITKT_${t}}, ${d} >, itk::Image< ${ITKT_${t}}, ${d} >")
-      itk_wrap_template("${ITKM_${t}}${d}${ITKM_${t}}${d}BSOUDFT${ITKT_D}${d}"
+      itk_wrap_template("${ITKM_${t}}${d}${ITKM_${t}}${d}BSOUDFT${ITKM_D}${d}"
```
dzenanz, line 9: This change looks like a bug fix. Maybe it warrants a separate commit, with the explanation what was broken?

thewtex (approved):
@PranjalSahu awesome!

dzenanz (approved):
Looks good on a glance.

Modules/Registration/RegistrationMethodsv4/wrapping/itkSyNImageRegistrationMethod.wrap
PranjalSahu: I can create separate commit. Actually nothing is broken. I discovered it while doing the Syn wrapping. It is change in naming only. If there is some code which is directly calling it by that name then it can fail. Instantiating using Class[TypeA, TypeB] will remain the same.

PranjalSahu:
Checking the issue.
It is throwing a warning but I am able to instantiate the class and it works properly.

`itkSyNImageRegistrationMethod: warning(4): ITK type not wrapped, or currently not known: itk::SmartPointer< const itk::Mesh< double, 2, itk::DefaultStaticMeshTraits< double, 2, 2, float, float, double > > >`

PranjalSahu:
So I looked at the code and I found this:

https://github.com/InsightSoftwareConsortium/ITK/blob/191e122087150c66d8352ebf1d0ecc7f1be76eff/Wrapping/Generators/SwigInterface/igenerator.py#L478

Do we need to add an entry for `itk::SmartPointer< const itk::Mesh.+ >"` as well ?
Let me try a build with this change.

dzenanz (approved):

PR #2988, opened 2021-12-22
JPEG IO: handle CMYK correctly

issakomi:
In JPEG IO 4-components JPEG files are assumed to be in _RBGA_ format. It is not correct, they are _CMYK_.
Photoshop has the option to export JPEG in _CMYK_ color-space. The files are rare, but not unusual. 
Note that _CMYK_ is out color-space, internally JPEG holds the data in _YCCK_ color-space.

Example file in [Wikipedia](https://en.wikipedia.org/wiki/File:Channel_digital_image_CMYK_color.jpg)

I have exported smaller photo from Photoshop, possible to use for test, s. [data.kitware.com](https://data.kitware.com/api/v1/item/61c093324acac99f420bbc9a/download?contentDisposition=inline).

Currently in my app JPEG IO opens the Wikipedia file like this:
al

With Slicer it is little tricky, Slicer uses black background, so the alpha channel is turned into black channel, to see the bug it is required to change the Slicer background to e.g. white. Paste following code into _Python Interactor_ 
```
sliceViewName = "Red"
view = slicer.app.layoutManager().sliceWidget(sliceViewName).sliceView()
view.setBackgroundColor(qt.QColor.fromRgb(255,255,255))
view.forceRender()
```
sl

The PR corrects this.
Also there is the option to load original channels as _VECTOR_ image (to examine channels and backward compatibility).


Modules/IO/JPEG/src/itkJPEGImageIO.cxx
```diff
@@ -207,30 +206,76 @@ JPEGImageIO::Read(void * buffer)
...
+
+    // finish the decompression step
+    jpeg_finish_decompress(&cinfo);
   }
+  else
+  {
+    JSAMPROW buf1[1];
+    while (cinfo.output_scanline < cinfo.output_height)
+    {
+      auto * volatile buf0 = new JSAMPLE[cinfo.output_width * 4];
```
dzenanz, line 244: Maybe `cinfo.output_components` could be used instead of `4`? Or add a comment `// CMYK has 4 components`?

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
```diff
@@ -207,30 +206,76 @@ JPEGImageIO::Read(void * buffer)
...
+          // even when an APP14 marker doesn't exist. This is the behavior
+          // of recent versions of PhotoShop as well.
+          const float K = buf1[0][4 * i + 3];
+          row_pointers[scanline][3 * i + 0] = static_cast<JSAMPLE>(buf1[0][4 * i + 0] * K / 255.0f);
+          row_pointers[scanline][3 * i + 1] = static_cast<JSAMPLE>(buf1[0][4 * i + 1] * K / 255.0f);
+          row_pointers[scanline][3 * i + 2] = static_cast<JSAMPLE>(buf1[0][4 * i + 2] * K / 255.0f);
+        }
+      }
+
+      delete[] buf0;
```
dzenanz, line 273: Could this allocation/deallocation be taken out of the loop, so there is only one allocation? If `cinfo.output_width` does not change from scanline to scanline, that should be done.

dzenanz (commented):
Looks good! Thanks for adding this feature.

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
issakomi: OK, it is clearer now, i hope. Yes, that `if (!(m_IsCMYK && m_CMYKtoRGB))` was worse readable.

Modules/IO/JPEG/src/itkJPEGImageIO.cxx
issakomi: OK.

dzenanz (approved):
Looks great!

issakomi:
I temporary converted to Draft again only because i clicked in a browser "Resolve conflict", this is only _CMakeLists.txt_ file with the test from https://github.com/InsightSoftwareConsortium/ITK/commit/4f770cb6bd2211eb32b928901826026c08740c6b. So there are 2 commits now, the second is `Merge branch 'master' into jpegcmyk`. Have to resolve it somehow.

dzenanz (approved):

issakomi:
I have found more files in the [post](https://github.com/haraldk/TwelveMonkeys/issues/571#issue-736193647), they are loading well, BTW.

PR #2992, opened 2021-12-24
BUG: Fix class names in `itkTypeMacro`

jhlegarreta:
Fix `Watersheds` module classes class names in `itkTypeMacro`.

## PR Checklist


jhlegarreta:
A few notes:
- The filenames do not match the class names, maybe because other base classes and their filenames have the same names used in these classes.
- `Segmenter`, `Boundary`, etc. seem too general, but they are within the `itk::watershed` namespace.
- I preferred to change the class names in the `itkTypeMacro` macro over changing the class names to `Watershed*` to avoid making a backwards incompatible/more extensive change, plus the above reason. But if we should do the latter, I will amend the commit.

dzenanz (approved):
This seems like the right choice. Further (possibly opposite changes) can be proposed in a new PR.

PR #2993, opened 2021-12-24
COMP: Use and move ITK_DISALLOW_COPY_AND_MOVE calls to public section

jhlegarreta:
Use the `ITK_DISALLOW_COPY_AND_MOVE` macro in the `Watersheds` module to
enhance consistency across the toolkit when disallowing the copy
constructor and the assign operator.

Move the `ITK_DISALLOW_COPY_AND_MOVE` calls to public section following
the discussion in
https://discourse.itk.org/t/noncopyable

Following the changes:
https://github.com/InsightSoftwareConsortium/ITK/commit/b0f183b04fd78062afafea26636b4dc597651b88
https://github.com/InsightSoftwareConsortium/ITK/commit/10429ae6b86c8d1f1578fe32c25180425f14a95d
https://github.com/InsightSoftwareConsortium/ITK/commit/7d177ca0c77aedb38cc8cf5561b785317617c315

## PR Checklist

### DIFF
--- a/Modules/Segmentation/Watersheds/include/itkWatershedBoundary.h
+++ b/Modules/Segmentation/Watersheds/include/itkWatershedBoundary.h
@@ -54,6 +54,8 @@ template <typename TScalar, unsigned int TDimension>
 class ITK_TEMPLATE_EXPORT Boundary : public DataObject
 {
 public:
+  ITK_DISALLOW_COPY_AND_MOVE(Boundary);
+
   /** The dimensionality of this boundary.  For example, if the boundary
    * of a set of planes, it has dimensionality 2.  If the boundary is
    * a set of lines, it has dimensionality 1.  Dimensionality is one less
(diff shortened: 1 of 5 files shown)


jhlegarreta:
@hjmjohnson @N-Dekker not sure if these were left behind on purpose, and thus if the PR should be discarded.

N-Dekker (approved):
Honestly it is not entirely obvious to me that the original intention was to entirely disallow copy and move, in these five cases. As in these cases, those copy constructors and assignment operators were declared `protected` (rather than `private`, which was the old C++98/C++03 way to prevent copying), and they were actually implemented (however, each of them by an empty function body).

Still I do agree with you that it's fine to use `ITK_DISALLOW_COPY_AND_MOVE` in these cases, especially because the CI tests have passed successfully  😃   

So thanks @jhlegarreta !

jhlegarreta:
> Honestly it is not entirely obvious to me that the original intention was to entirely disallow copy and move, in these five cases. As in these cases, those copy constructors and assignment operators were declared protected (rather than private, which was the old C++98/C++03 way to prevent copying), and they were actually implemented (however, each of them by an empty function body).

Thanks for the insight @N-Dekker :100: !

dzenanz (approved):

PR #2994, opened 2021-12-25
STYLE: Use `auto` for variables initialized by `New()` in Examples

jhlegarreta:
Replace initializations of the form `T::Pointer var = T::New()`
(sometimes starting with `typename`) with `auto var = T::New()` in `Examples` to
reduce code redundancy

In accordance with C++ Core Guidelines, August 19, 2021:
"ES.11: Use `auto` to avoid redundant repetition of type names"
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-auto

Following PR https://github.com/InsightSoftwareConsortium/ITK/pull/2826
and commit 828453d1bf61c487310d2d8c9570093e08798a40.

## PR Checklist

### DIFF
--- a/Examples/DataRepresentation/Image/Image1.cxx
+++ b/Examples/DataRepresentation/Image/Image1.cxx
@@ -62,7 +62,7 @@ main(int, char *[])
   // Software Guide : EndLatex
   //
   // Software Guide : BeginCodeSnippet
-  ImageType::Pointer image = ImageType::New();
+  auto image = ImageType::New();
   // Software Guide : EndCodeSnippet
 
 
--- a/Examples/DataRepresentation/Image/Image2.cxx
+++ b/Examples/DataRepresentation/Image/Image2.cxx
@@ -81,7 +81,7 @@ main(int, char * argv[])
   // Software Guide : EndLatex
 
   // Software Guide : BeginCodeSnippet
-  ReaderType::Pointer reader = ReaderType::New();
+  auto reader = ReaderType::New();
   // Software Guide : EndCodeSnippet
 
   // Software Guide : BeginLatex
--- a/Examples/DataRepresentation/Image/Image3.cxx
+++ b/Examples/DataRepresentation/Image/Image3.cxx
@@ -40,7 +40,7 @@ main(int, char *[])
   using ImageType = itk::Image<unsigned short, 3>;
 
   // Then the image object can be created
-  ImageType::Pointer image = ImageType::New();
+  auto image = ImageType::New();
 
   // The image region should be initialized
   const ImageType::SizeType size = {
--- a/Examples/DataRepresentation/Image/Image4.cxx
+++ b/Examples/DataRepresentation/Image/Image4.cxx
@@ -79,7 +79,7 @@ main(int, char *[])
 {
   constexpr unsigned int Dimension = 3;
   using ImageType = itk::Image<unsigned short, Dimension>;
-  ImageType::Pointer image = ImageType::New();
+  auto image = ImageType::New();
 
   const ImageType::SizeType size = {
     { 200, 200, 200 }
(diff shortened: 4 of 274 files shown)


jhlegarreta:
A few notes:

- I just did this because 5ab79bc4149007a151c7e46735a49cffd5967a19 was merged, and because it made no sense to keep the `Examples` inconsistent with the rest of the code.
- @N-Dekker I noticed that your regexes do not cover cases where the `New()` statement is on a separate line, or does it? If not, there might be places across the code where the substitution was not applied.

  I did not find the regex doing that job so for this patch set I did it manually.

- There are many SW Guide paragraphs that mention smart pointers as the return types just before the calls to `New()`, e.g.
```
//  A visitor implementation class can now be created using the normal
//  invocation to its \code{New()} method and assigning the result to a
//  \doxygen{SmartPointer}.
```

I think it is fair to leave them as they are without further explanation, since they are still accurate even if using `auto`.

If I'm not wrong the CIs do not build the examples, so I guess we will not know if I did not break anything until this gets merged and the regular builds report to CDash (did not build locally).

N-Dekker:
Very nice @jhlegarreta  Indeed, the commit that I did (828453d1bf61c487310d2d8c9570093e08798a40), did not deal with the situation of having the `New()` statement on a separate line.

Just some comment on your commit message:

I would suggest to have a _slightly_ different subject line, so that your commit can be distinguished more easily from mine. For example:

    STYLE: Use auto for variables initialized by `New()` in Examples

Or just:

    STYLE: Use auto for variables initialized by `New()` (follow-up)


The commit message says:

> Following commit 5ab79bc.

That particular commit appears to produce a warning on GitHub, saying: "This commit does not belong to any branch ..."  On the Git Bash command prompt, this particular commit cannot be found from a fresh ITK git clone, as I tried by doing `git show 5ab79bc4149007a151c7e46735a49cffd5967a19`. (It said "fatal: bad object".)


So I would suggest to _instead_ mention the corresponding commit at the master branch: `828453d1bf61c487310d2d8c9570093e08798a40` (commit 828453d1bf61c487310d2d8c9570093e08798a40).

Also I think it would be nice to mention the corresponding pull request in the commit message: `https://github.com/InsightSoftwareConsortium/ITK/pull/2826` (pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2826)
 
My two cents, Niels

jhlegarreta:
Done in 41cedd7.

Note that I essentially only looked at the very same PR you mentioned https://github.com/InsightSoftwareConsortium/ITK/pull/2826/commits, and as you can see the commit reference was accurate. The message you mention does appear on the web interface as well. For some reason it does not appear in the web history, but usually when a PR is merged from a fork, a new merge commit with the message `Merge pull request (...)` is added, and the original one is kept in the history. See the screenshots.

commit_auto_style_1

commit_auto_style_2

commit_auto_style_3

Not sure what's up with that commit, but it is done in any case.

N-Dekker (approved):
👍

N-Dekker:
@jhlegarreta  The commit you originally referred to (5ab79bc4149007a151c7e46735a49cffd5967a19) belongs to PR #2826, indeed, so I guess that indeed, it will be kept in the history at github.com/InsightSoftwareConsortium/ITK for a long time. However, it won't be available "offline" at local ITK clones of end users.  That's just why I prefer to mention the commit from the master branch (828453d1bf61c487310d2d8c9570093e08798a40), as that one is available at local ITK clones. And it may be available for an even longer time...  😃 

Thanks for adjusting the commit message!

dzenanz (approved):

Leengit:
We might be able to handle the multi-line statements with `New()` by first invoking `clang-format` with `ColumnLimit: 10000` (or `ColumnLimit: 0`).  Then they'll all be single line.  After the `auto` regular-expression edit, then run `clang-format` again with default parameters.  There could be side effects, but it might be worth a try to check it out.

PR #2999, opened 2021-12-27
Fix BuildInformation GetValue, GetDescription `key` parameter type, extend its test

N-Dekker:
Added a check that for each entry from the map retrieved by `GetMap()`,
both `GetValue(key)` and `GetDescription(key)` return the expected
string.

### DIFF
--- a/Modules/Core/Common/test/itkBuildInformationGTest.cxx
+++ b/Modules/Core/Common/test/itkBuildInformationGTest.cxx
@@ -49,10 +49,15 @@ TEST(ITKBuildInformation, InformationFeatures)
 
   for (auto mapEntry : localMap)
   {
+    const auto & key = mapEntry.first;
+
     std::cout << "--------------------------------------------------------------------" << std::endl
-              << "Key: " << mapEntry.first << std::endl
+              << "Key: " << key << std::endl
               << "\tValue: " << mapEntry.second.m_Value << std::endl
               << "\tDescription: " << mapEntry.second.m_Description << std::endl;
+
+    EXPECT_EQ(itk::BuildInformation::GetValue(key), mapEntry.second.m_Value);
+    EXPECT_EQ(itk::BuildInformation::GetDescription(key), mapEntry.second.m_Description);
   }
 
   EXPECT_EQ(itk::BuildInformation::GetValue("PROJECT_URL"), "http://www.itk.org");
--- a/Modules/Core/Common/include/itkBuildInformation.h
+++ b/Modules/Core/Common/include/itkBuildInformation.h
@@ -90,9 +90,9 @@ public:
   static const MapType &
   GetMap();
   static const MapValueType
-  GetValue(const MapKeyType &&);
+  GetValue(const MapKeyType &);
   static const MapValueDescriptionType
-  GetDescription(const MapKeyType &&);
+  GetDescription(const MapKeyType &);
   static const std::vector<MapKeyType>
   GetAllKeys();
 
--- a/Modules/Core/Common/src/itkBuildInformation.cxx.in
+++ b/Modules/Core/Common/src/itkBuildInformation.cxx.in
@@ -78,7 +78,7 @@ BuildInformation::GetMap()
 }
 
 const BuildInformation::MapValueType
-BuildInformation::GetValue(const MapKeyType &&key)
+BuildInformation::GetValue(const MapKeyType & key)
 {
   const MapType &localMap = BuildInformation::GetInstance()->GetMap();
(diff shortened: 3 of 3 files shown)


N-Dekker:
@hjmjohnson Can you possibly recall why these `GetValue` and `GetDescription` member functions were declared this way, having a const r-value reference `const MapKeyType &&` as parameter type?  You know, `const` r-value references are quite rare in C++, as r-value references are mostly used to "move" from, whereas `const` actually prevents doing such a "move".

It seems to me that an old-style `const MapKeyType &` (const l-value reference) would be just fine here.


https://github.com/InsightSoftwareConsortium/ITK/blob/a908033cacbbf7d23c12941cb69e39da336c8ee1/Modules/Core/Common/include/itkBuildInformation.h#L86-L87

Modules/Core/Common/test/itkBuildInformationGTest.cxx
```diff
@@ -49,10 +49,15 @@ TEST(ITKBuildInformation, InformationFeatures)
 
   for (auto mapEntry : localMap)
   {
+    const auto key = mapEntry.first;
+
```
dzenanz, line 53: Why not add `value` as a local variable `auto value = mapEntry.second;`? This discrepancy triggers my OCD 😄

N-Dekker: > Why not add value as a local variable

@dzenanz Thanks for the suggestions, but the thing that bothers me is the inability to pass an l-value `key` as argument to `GetValue` and `GetDescription`. So  `key` is the key thing of this pull request  😸

N-Dekker:
@dzenanz @hjmjohnson  As this pull request still needs to be adjusted anyway, would you agree that the `std::cout` statement that prints all map entries may just be removed, once each map entry is checked by those two `EXPECT_EQ` calls?   

https://github.com/InsightSoftwareConsortium/ITK/blob/9a0be4934e7656d8aa4ba98b7d8c6d293ae7074b/Modules/Core/Common/test/itkBuildInformationGTest.cxx#L50-L62

dzenanz:
The test seems somewhat tautological. But I guess many tests do. Since no one is probably checking that output, I am fine with removing it.

N-Dekker:
https://open.cdash.org/viewBuildError.php?buildid=7646720  (Azure.fv-az200-80, Windows_NT-Build7319...) says:

> itkBuildInformationGTest.cxx(59): error C2664: 'const itk::BuildInformation::MapValueType itk::BuildInformation::GetValue(const itk::BuildInformation::MapKeyType &&)': cannot convert argument 1 from '_Ty1' to 'const itk::BuildInformation::MapKeyType &&'

https://open.cdash.org/viewBuildError.php?buildid=7646718 (Azure.Mac-1640624256914, Darwin-Build7268...) says:
> itkBuildInformationGTest.cxx:59:47: error: rvalue reference to type 'const basic_string<...>' cannot bind to lvalue of type 'const basic_string<...>'

https://open.cdash.org/viewBuildError.php?buildid=7646711  (Azure.fv-az36-134, Linux-Build7338...) says:
> itkBuildInformationGTest.cxx:59:50: error: cannot bind rvalue reference of type 'const MapKeyType&& {aka const std::__cxx11::basic_string<char>&&}' to lvalue of type 'const std::__cxx11::basic_string<char>'

@hjmjohnson @dzenanz Do you agree that the signatures of `GetValue(const MapKeyType &&)` and `GetDescription(const MapKeyType &&)` should just have a single `&`?

dzenanz:
I don't think that `itk::BuildInformation` is a widely used class. You are one of the rare people who even looked at it. You want to show it some love? Sure, go ahead. I am fine with changes to this class, as long as everything compiles 😄

dzenanz (approved):

hjmjohnson:
> @hjmjohnson Can you possibly recall why these `GetValue` and `GetDescription` member functions were declared this way, having a const r-value reference `const MapKeyType &&` as parameter type? You know, `const` r-value references are quite rare in C++, as r-value references are mostly used to "move" from, whereas `const` actually prevents doing such a "move".
> 
> It seems to me that an old-style `const MapKeyType &` (const l-value reference) would be just fine here.
> 
> https://github.com/InsightSoftwareConsortium/ITK/blob/a908033cacbbf7d23c12941cb69e39da336c8ee1/Modules/Core/Common/include/itkBuildInformation.h#L86-L87

I probably made a mistake here.  A change would be welcomed.

hjmjohnson (approved):

Issue #2950, opened 2021-12-13
BUG: FFT object factories not registered by default in ITK remote module tests


tbirdso:
### Description

FFT image filters were recently updated to allow swappable backends at runtime via object factory methods. Projects consuming ITK and optional remote modules need only include `USEITK.cmake` to construct relevant factory manager files and initiate factory registration. However, ITK and its remote modules itself do not construct these files. As a result ITK factories are not available by default in ITK and remote module tests and currently must be manually included, which creates an additional burden on remote modules. `itk::ImageFileReader` and `itk::ImageFileWriter` typically handle the overhead of registering ImageIO default factories where needed, but no such overlying class exists for FFT filters.

I see three possible paths forward:
1) _Implement the proper framework to register default factories for ITK tests (and remote modules)_. Ideally we would treat ITK tests as the final project consumers of ITK itself and `include(UseITK.cmake)` for each test driver so that default factories are available in testing, which would most closely mirror the behavior of external projects that use ITK. My understanding is that ITK tests are compiled in a separate step similar to wrappings. I believe this is the best option, although the most intensive.
2) Manually register FFT factories in ITK and remote module tests. This is a stopgap that is already added to ITK FFT tests, and could be extended to remote modules. It requires no new build architecture but is a manual process that requires knowledge of underlying FFT requirements for each class. For instance, see segfault failures in [ITKMontage](https://open.cdash.org/viewTest.php?onlyfailed&buildid=7604228) tests where `itk::FFTPadImageFilter` init failed because no factory was registered for `itk::ForwardFFTImageFilter`.
3) Introduce an overhead class for FFTs similar to `itk::ImageFileReader` and `itk::ImageFileWriter` to handle default factory registrations where required. This is my least favorite option as it introduces more overhead specifically for FFT classes and is not easily extensible to any future classes that we wish to make depend on object factory registrations.

Thank you @dzenanz for performing initial investigation of the issue in [ITKMontage](https://open.cdash.org/viewTest.php?onlyfailed&buildid=7604228) and linking it back to FFT changes.

### Steps to Reproduce

Reproducing in ITKMontage:
1. Build ITK with ITKMontage enabled as a remote module, or separately with ITKMontage built against ITK (both in `RelWithDebInfo` configuration)
2. Run `ctest -C Release -R itkMontagePCMSynthetic_2cc`

### Expected behavior

Test passes. FFT factories are registered by default.

### Actual behavior

[...]

PR #3002, opened 2021-12-27
ENH: Add default FFT object factories to TestKernel initialization

tbirdso:
Addresses remote module test failures in #2950 by adding FFT factories to the list of default factories to be registered with ITKTestKernel.

`itkTestDriverIncludeRequiredIOFactories.h` and similar strings are renamed to `itkTestDriverIncludeRequiredFactories.h` to reflect that FFT factories and any other future backends managed with the object factory should be default registered here. Manual FFT factory registrations previously required for testing are removed.

ITKMontage tests pass when these changes are applied.

## PR Checklist

### DIFF
--- a/CMake/ITKModuleTest.cmake
+++ b/CMake/ITKModuleTest.cmake
@@ -56,7 +56,7 @@ EM_ASM(
   set(CMAKE_TESTDRIVER_AFTER_TESTMAIN "#include \"itkTestDriverAfterTest.inc\"${emscripten_after}")
   create_test_sourcelist(Tests ${KIT}TestDriver.cxx
     ${KitTests}
-    EXTRA_INCLUDE itkTestDriverIncludeRequiredIOFactories.h
+    EXTRA_INCLUDE itkTestDriverIncludeRequiredFactories.h
     FUNCTION  ProcessArgumentsAndRegisterRequiredFactories
     )
   add_executable(${KIT}TestDriver ${KIT}TestDriver.cxx ${Tests} ${ADDITIONAL_SRC})
--- a/Modules/Core/TestKernel/src/itkTestDriverIncludeRequiredFactories.cxx
+++ b/Modules/Core/TestKernel/src/itkTestDriverIncludeRequiredFactories.cxx
@@ -16,7 +16,7 @@
  *
  *=========================================================================*/
 
-#include "itkTestDriverIncludeRequiredIOFactories.h"
+#include "itkTestDriverIncludeRequiredFactories.h"
 
 // ImageIO
 #include "itkGDCMImageIOFactory.h"
@@ -41,12 +41,31 @@
 #include "itkOFFMeshIOFactory.h"
 #include "itkVTKPolyDataMeshIOFactory.h"
 
+// FFT
+#include "itkFFTImageFilterFactory.h"
+#include "itkVnlComplexToComplex1DFFTImageFilter.h"
+#include "itkVnlComplexToComplexFFTImageFilter.h"
+#include "itkVnlForward1DFFTImageFilter.h"
+#include "itkVnlForwardFFTImageFilter.h"
+#include "itkVnlHalfHermitianToRealInverseFFTImageFilter.h"
+#include "itkVnlInverse1DFFTImageFilter.h"
+#include "itkVnlInverseFFTImageFilter.h"
+#include "itkVnlRealToHalfHermitianForwardFFTImageFilter.h"
+
 #ifdef __EMSCRIPTEN__
 #  include <emscripten.h>
 #endif
 
 void
 RegisterRequiredFactories()
+{
+  RegisterRequiredIOFactories();
+
+  RegisterRequiredFFTFactories();
+}
+
+void
+RegisterRequiredIOFactories()
 {
   // ImageIO
   itk::ObjectFactoryBase::RegisterFactory(itk::MetaImageIOFactory::New());
@@ -68,6 +87,21 @@ RegisterRequiredFactories()
   itk::ObjectFactoryBase::RegisterFactory(itk::OBJMeshIOFactory::New());
   itk::ObjectFactoryBase::RegisterFactory(itk::OFFMeshIOFactory::New());
   itk::ObjectFactoryBase::RegisterFactory(itk::VTKPolyDataMeshIOFactory::New());
+}
+
+void
+RegisterRequiredFFTFactories()
+{
+  itk::ObjectFactoryBase::RegisterFactory(itk::FFTImageFilterFactory<itk::VnlComplexToComplex1DFFTImageFilter>::New());
+  itk::ObjectFactoryBase::RegisterFactory(itk::FFTImageFilterFactory<itk::VnlComplexToComplexFFTImageFilter>::New());
+  itk::ObjectFactoryBase::RegisterFactory(itk::FFTImageFilterFactory<itk::VnlForward1DFFTImageFilter>::New());
+  itk::ObjectFactoryBase::RegisterFactory(itk::FFTImageFilterFactory<itk::VnlForwardFFTImageFilter>::New());
+  itk::ObjectFactoryBase::RegisterFactory(
+    itk::FFTImageFilterFactory<itk::VnlHalfHermitianToRealInverseFFTImageFilter>::New());
+  itk::ObjectFactoryBase::RegisterFactory(itk::FFTImageFilterFactory<itk::VnlInverse1DFFTImageFilter>::New());
+  itk::ObjectFactoryBase::RegisterFactory(itk::FFTImageFilterFactory<itk::VnlInverseFFTImageFilter>::New());
+  itk::ObjectFactoryBase::RegisterFactory(
+    itk::FFTImageFilterFactory<itk::VnlRealToHalfHermitianForwardFFTImageFilter>::New());
 }
 
 void
--- a/Modules/Filtering/FFT/include/itkFFTImageFilterFactory.h
+++ b/Modules/Filtering/FFT/include/itkFFTImageFilterFactory.h
@@ -117,6 +117,9 @@ protected:
     OverrideFFTImageFilterType<typename FFTImageFilterTraits<TFFTImageFilter>::template InputPixelType<float>,
                                typename FFTImageFilterTraits<TFFTImageFilter>::template OutputPixelType<float>,
                                3>();
+    OverrideFFTImageFilterType<typename FFTImageFilterTraits<TFFTImageFilter>::template InputPixelType<float>,
+                               typename FFTImageFilterTraits<TFFTImageFilter>::template OutputPixelType<float>,
+                               4>();
 
     OverrideFFTImageFilterType<typename FFTImageFilterTraits<TFFTImageFilter>::template InputPixelType<double>,
                                typename FFTImageFilterTraits<TFFTImageFilter>::template OutputPixelType<double>,
@@ -127,6 +130,9 @@ protected:
     OverrideFFTImageFilterType<typename FFTImageFilterTraits<TFFTImageFilter>::template InputPixelType<double>,
                                typename FFTImageFilterTraits<TFFTImageFilter>::template OutputPixelType<double>,
                                3>();
+    OverrideFFTImageFilterType<typename FFTImageFilterTraits<TFFTImageFilter>::template InputPixelType<double>,
+                               typename FFTImageFilterTraits<TFFTImageFilter>::template OutputPixelType<double>,
+                               4>();
   }
 };
 
--- a/Modules/Filtering/FFT/include/itkFFTImageFilterFactory.h
+++ b/Modules/Filtering/FFT/include/itkFFTImageFilterFactory.h
@@ -19,6 +19,7 @@
 #ifndef itkFFTImageFilterFactory_h
 #define itkFFTImageFilterFactory_h
 
+#include "itkImage.h"
 #include "itkObjectFactoryBase.h"
 #include "itkVersion.h"
 
--- a/Modules/Segmentation/SuperPixel/test/itkSLICImageFilterGTest.cxx
+++ b/Modules/Segmentation/SuperPixel/test/itkSLICImageFilterGTest.cxx
@@ -23,7 +23,7 @@
 
 #include "itkCommand.h"
 
-#include "itkTestDriverIncludeRequiredIOFactories.h"
+#include "itkTestDriverIncludeRequiredFactories.h"
 #include "itkTestingHashImageFilter.h"
 
 namespace
--- a/Modules/Core/TestKernel/include/itkTestDriverIncludeRequiredFactories.h
+++ b/Modules/Core/TestKernel/include/itkTestDriverIncludeRequiredFactories.h
@@ -15,12 +15,26 @@
  *  limitations under the License.
  *
  *=========================================================================*/
+#ifndef itkTestDriverIncludeRequiredFactories_h
+#define itkTestDriverIncludeRequiredFactories_h
 
-// Legacy file for backwards compatibility
 
-#ifndef itkTestDriverIncludeRequiredIOFactories_h
-#define itkTestDriverIncludeRequiredIOFactories_h
+#include "itkTestDriverInclude.h"
 
-#include "itkTestDriverIncludeRequiredFactories.h"
+#ifdef __EMSCRIPTEN__
+#  include <emscripten.h>
+#endif
+
+void
+RegisterRequiredFactories();
+
+void
+RegisterRequiredIOFactories();
+
+void
+RegisterRequiredFFTFactories();
+
+void
+ProcessArgumentsAndRegisterRequiredFactories(int * ac, ArgumentStringType * av);
 
 #endif
--- a/Modules/Core/TestKernel/itk-module.cmake
+++ b/Modules/Core/TestKernel/itk-module.cmake
@@ -24,6 +24,7 @@ itk_module(ITKTestKernel
     ITKIOMeshGifti
     ITKIOMeshOBJ
     ITKIOMeshOFF
+    ITKFFT
   COMPILE_DEPENDS
     ITKKWSys
     ITKDoubleConversion
--- a/Modules/Core/TestKernel/src/CMakeLists.txt
+++ b/Modules/Core/TestKernel/src/CMakeLists.txt
@@ -10,7 +10,7 @@ if(NOT DO_NOT_INSTALL_ITK_TEST_DRIVER) # used only by vcpkg
 endif()
 
 set(ITKTestKernel_SRCS
-  itkTestDriverIncludeRequiredIOFactories.cxx
+  itkTestDriverIncludeRequiredFactories.cxx
   itkTestDriverInclude.cxx
   itkTestingExtractSliceImageFilter.cxx
   itkTestingHashImageFilter.cxx)
--- a/Modules/Core/TestKernel/src/itkTestDriver.cxx
+++ b/Modules/Core/TestKernel/src/itkTestDriver.cxx
@@ -26,7 +26,7 @@
  *
  *=========================================================================*/
 
-#include "itkTestDriverIncludeRequiredIOFactories.h"
+#include "itkTestDriverIncludeRequiredFactories.h"
 
 
 /* Select the environment variable holding the shared library runtime
--- a/Modules/Filtering/Convolution/test/itkFFTConvolutionImageFilterDeltaFunctionTest.cxx
+++ b/Modules/Filtering/Convolution/test/itkFFTConvolutionImageFilterDeltaFunctionTest.cxx
@@ -67,19 +67,6 @@ itkFFTConvolutionImageFilterDeltaFunctionTest(int argc, char * argv[])
   }
   deltaFunctionImage->SetPixel(middleIndex, 1);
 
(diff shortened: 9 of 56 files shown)


CMake/ITKModuleTest.cmake
```diff
@@ -65,7 +65,7 @@ EM_ASM(
 endmacro()
 
 
-macro(CreateTestDriver_SupportBuildInIOFactories KIT KIT_LIBS KitTests)
+macro(CreateTestDriver_SupportBuildInFactories KIT KIT_LIBS KitTests)
```
dzenanz, line 68: BuildIn -> BuiltIn

Modules/Core/TestKernel/src/itkTestDriverIncludeRequiredFactories.cxx
```diff
@@ -41,12 +41,42 @@
 #include "itkOFFMeshIOFactory.h"
 #include "itkVTKPolyDataMeshIOFactory.h"
 
+// FFT
```
dzenanz, line 44: Doesn't `itkTestDriverIncludeRequiredFactories.h` already include all of these? If so, the above list of IOs should be cleaned up too.

Modules/Core/TestKernel/src/itkTestDriverIncludeRequiredFactories.cxx
```diff
@@ -70,6 +100,34 @@ RegisterRequiredFactories()
   itk::ObjectFactoryBase::RegisterFactory(itk::VTKPolyDataMeshIOFactory::New());
 }
 
+void
+RegisterRequiredFFTFactories()
+{
+#if defined(ITK_USE_FFTWD) || defined(ITK_USE_FFTWF)
```
dzenanz, line 106: Can we avoid duplication with https://github.com/InsightSoftwareConsortium/ITK/blob/test-kernel-register-fft-factories/Modules/Core/TestKernel/include/itkTestDriverIncludeBuiltInFactories.h?

Modules/Segmentation/SuperPixel/test/itkSLICImageFilterGTest.cxx
```diff
@@ -23,7 +23,7 @@
 
 #include "itkCommand.h"
 
-#include "itkTestDriverIncludeRequiredIOFactories.h"
+#include "itkTestDriverIncludeRequiredFactories.h"
```
dzenanz, line 26: Will this change be required in external code, such as remote modules? If so, we could keep a file called `itkTestDriverIncludeRequiredIOFactories.h` with content `#include "itkTestDriverIncludeRequiredFactories.h"`.

dzenanz:
It looks like https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/FFT/include/itkFFTImageFilterFactory.h should `#include "itkImage.h"`.

Modules/Core/TestKernel/src/itkTestDriverIncludeRequiredFactories.cxx
tbirdso: `itkTestDriverIncludeRequiredFactories` is referenced in the [`CreateTestDriver`](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CMake/ITKModuleTest.cmake#L59) macro while `itkTestDriverIncludeBuildInFactories` is referenced in the [`CreateTestDriver_SupportBuildInFactories`](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CMake/ITKModuleTest.cmake#L113) macro. It looks like `BuildInFactories` was split out from `RequiredFactories` in [this commit](https://github.com/InsightSoftwareConsortium/ITK/commit/a2c7a16e45a35f3c3988b1dac4d349b0a754fc1f) 10 years ago to aid regression testing.

I'm not sure that I fully understand the implications of how `BuildInFactories` is being used here, it looks like it is only invoked in the `Nonunit/Review` and `Nonunit/IntegratedTest` modules. @dzenanz do you maybe know why these modules are using a slightly different testing structure?

If this is an old/outdated structure we can look into consolidating, but the scope of this PR should probably stay limited to adding default factories.

Modules/Core/TestKernel/src/itkTestDriverIncludeRequiredFactories.cxx
tbirdso: `itkTestDriverIncludeBuildInFactories.h` includes these while `itkTestDriverIncludeRequiredFactories.h` only declares the functions that are implemented in `itkTestDriverIncludeRequiredFactories.cxx`. Ideally we should keep factory registrations in the `.cxx` implementation file.

Modules/Segmentation/SuperPixel/test/itkSLICImageFilterGTest.cxx
tbirdso: All tests relying on the ITKTestKernel architecture already include `itkTestDriverIncludeRequiredIOFactories.h`, so this line will only appear in older tests not relying on the test kernel. Looking through ITK I'm only able to find this line in GTest and "v3" segmentation/registration tests that are built as separate executables. Are we still supporting these in ITK 5, or are remote modules expected to use the test kernel?

Modules/Filtering/FFT/include/itkFFTImageFilterFactory.h
```diff
@@ -116,6 +117,9 @@ class FFTImageFilterFactory : public itk::ObjectFactoryBase
     OverrideFFTImageFilterType<typename FFTImageFilterTraits<TFFTImageFilter>::template InputPixelType<float>,
                                typename FFTImageFilterTraits<TFFTImageFilter>::template OutputPixelType<float>,
                                3>();
+    OverrideFFTImageFilterType<typename FFTImageFilterTraits<TFFTImageFilter>::template InputPixelType<float>,
```
tbirdso, line 120: `ITKMontage` tests require FFT padding for a 4D image, while FFT factory overrides were previously only registered for 1D, 2D, and 3D images. After consideration I think it's reasonable to add a default override for Vnl and FFTW FFT implementations for 4D images as this is a reasonably commonly used ITK image dimension and appears to be supported by Vnl and FFTW backends. 

For higher dimensions we will require consuming projects to explicitly register FFT factory overrides for their needs. Other FFT implementations in remote modules such as ITKVkFFTBackend may not support 4D images, so I expect to revisit this in the near future to make 4D overrides optional.

Modules/Core/TestKernel/src/itkTestDriverIncludeRequiredFactories.cxx
dzenanz: I don't know why `Nonunit` tests have a different structure. Maybe a relic of old, maybe because they need it. They do testing which involves multiple modules, so they are different from unit tests.

> we can look into consolidating

We should note this as an issue, unless you are going to do this soon.

Modules/Segmentation/SuperPixel/test/itkSLICImageFilterGTest.cxx
dzenanz: "v3" registration is still fully supported.

Modules/Filtering/FFT/include/itkFFTImageFilterFactory.h
dzenanz: Sounds good.

dzenanz (approved):

Modules/Core/TestKernel/src/itkTestDriverIncludeRequiredFactories.cxx
tbirdso: > They do testing which involves multiple modules, so they are different from unit tests.

Ah okay, good to know.

> We should note this as an issue

Agreed, I will enter an issue to track this discussion separately.

tbirdso: See #3004

CMake/ITKModuleTest.cmake
tbirdso: Fixed

Modules/Segmentation/SuperPixel/test/itkSLICImageFilterGTest.cxx
tbirdso: Re-added `itkTestDriverIncludeRequiredIOFactories.h` for backwards compatibility

Modules/Core/TestKernel/include/itkTestDriverIncludeRequiredIOFactories.h
```diff
@@ -15,20 +15,12 @@
...
 
-#include "itkTestDriverInclude.h"
-
-#ifdef __EMSCRIPTEN__
-#  include <emscripten.h>
-#endif
-
-void
-RegisterRequiredFactories();
+#ifndef itkTestDriverIncludeRequiredIOFactories_h
```
dzenanz, line 21: I don't think that include guards are needed, as the included file will have them.

dzenanz (approved):

Modules/Core/TestKernel/include/itkTestDriverIncludeRequiredIOFactories.h
tbirdso: I thought the include guards were part of the style guidelines? I'll try without them

dzenanz: Include guards are needed to [prevent errors](https://www.geeksforgeeks.org/include-guards-in-c/), ITK's style guide just defines their style 😄

tbirdso: Removing the include guards fails the KWStyle commit check:

```
Modules/Core/TestKernel/include/itkTestDriverIncludeRequiredIOFactories.h:17: error: #ifndef not defined
pre-commit hook failure
-----------------------

KWStyle check failed.

Line numbers in the errors shown refer to the file:
Modules/Core/TestKernel/include/itkTestDriverIncludeRequiredIOFactories.h.kws
```

dzenanz: But maybe KWStyle might complain due to their lack. If it does, put them back in 😄

dzenanz: > Removing the include guards fails the KWStyle commit check

Then keep them. I wrote the above comment while you were writing yours 😄

tbirdso: Right, they're redundant here but not particularly messy. Will leave them in to make KWStyle happy.

Modules/Core/TestKernel/include/itkTestDriverIncludeBuiltInFactories.h
```diff
@@ -44,6 +44,28 @@
...
+#if defined(ITK_USE_FFTWD) || defined(ITK_USE_FFTWF)
+#  include "itkFFTWComplexToComplex1DFFTImageFilter.h"
+#  include "itkFFTWComplexToComplexFFTImageFilter.h"
+#  include "itkFFTWForward1DFFTImageFilter.h"
+#  include "itkFFTWForwardFFTImageFilter.h"
+#  include "itkFFTWHalfHermitianToRealInverseFFTImageFilter.h"
+#  include "itkFFTWInverse1DFFTImageFilter.h"
+#  include "itkFFTWInverseFFTImageFilter.h"
+#  include "itkFFTWRealToHalfHermitianForwardFFTImageFilter.h"
+#endif
```
thewtex, line 67: For test purposes, we only need the VNL implementations -- this can be removed.

Modules/Core/TestKernel/include/itkTestDriverIncludeBuiltInFactories.h
```diff
@@ -78,6 +100,31 @@ ProcessArgumentsAndRegisterBuiltInFactories(int * ac, ArgumentStringType * av)
...
+  itk::ObjectFactoryBase::RegisterFactory(itk::FFTImageFilterFactory<itk::FFTWComplexToComplexFFTImageFilter>::New());
+  itk::ObjectFactoryBase::RegisterFactory(itk::FFTImageFilterFactory<itk::FFTWForward1DFFTImageFilter>::New());
+  itk::ObjectFactoryBase::RegisterFactory(itk::FFTImageFilterFactory<itk::FFTWForwardFFTImageFilter>::New());
+  itk::ObjectFactoryBase::RegisterFactory(
+    itk::FFTImageFilterFactory<itk::FFTWHalfHermitianToRealInverseFFTImageFilter>::New());
+  itk::ObjectFactoryBase::RegisterFactory(itk::FFTImageFilterFactory<itk::FFTWInverse1DFFTImageFilter>::New());
+  itk::ObjectFactoryBase::RegisterFactory(itk::FFTImageFilterFactory<itk::FFTWInverseFFTImageFilter>::New());
+  itk::ObjectFactoryBase::RegisterFactory(
+    itk::FFTImageFilterFactory<itk::FFTWRealToHalfHermitianForwardFFTImageFilter>::New());
+#endif
```
thewtex, line 115: For test purposes, we only need the VNL implementations -- this can be removed.

Modules/Core/TestKernel/src/itkTestDriverIncludeRequiredFactories.cxx
```diff
@@ -41,12 +41,42 @@
...
+#  include "itkFFTWComplexToComplex1DFFTImageFilter.h"
+#  include "itkFFTWComplexToComplexFFTImageFilter.h"
+#  include "itkFFTWForward1DFFTImageFilter.h"
+#  include "itkFFTWForwardFFTImageFilter.h"
+#  include "itkFFTWHalfHermitianToRealInverseFFTImageFilter.h"
+#  include "itkFFTWInverse1DFFTImageFilter.h"
+#  include "itkFFTWInverseFFTImageFilter.h"
+#  include "itkFFTWRealToHalfHermitianForwardFFTImageFilter.h"
+#endif
```
thewtex, line 64: remove

Modules/Core/TestKernel/src/itkTestDriverIncludeRequiredFactories.cxx
```diff
@@ -70,6 +100,34 @@ RegisterRequiredFactories()
...
+  itk::ObjectFactoryBase::RegisterFactory(itk::FFTImageFilterFactory<itk::FFTWComplexToComplexFFTImageFilter>::New());
+  itk::ObjectFactoryBase::RegisterFactory(itk::FFTImageFilterFactory<itk::FFTWForward1DFFTImageFilter>::New());
+  itk::ObjectFactoryBase::RegisterFactory(itk::FFTImageFilterFactory<itk::FFTWForwardFFTImageFilter>::New());
+  itk::ObjectFactoryBase::RegisterFactory(
+    itk::FFTImageFilterFactory<itk::FFTWHalfHermitianToRealInverseFFTImageFilter>::New());
+  itk::ObjectFactoryBase::RegisterFactory(itk::FFTImageFilterFactory<itk::FFTWInverse1DFFTImageFilter>::New());
+  itk::ObjectFactoryBase::RegisterFactory(itk::FFTImageFilterFactory<itk::FFTWInverseFFTImageFilter>::New());
+  itk::ObjectFactoryBase::RegisterFactory(
+    itk::FFTImageFilterFactory<itk::FFTWRealToHalfHermitianForwardFFTImageFilter>::New());
+#endif
```
thewtex, line 117: remove

thewtex (changes requested):
@tbirdso looking good!

To improve the consistency of the factories registered by default by the test driver, just the VNL factories can be registered. If other factories are required by a test, e.g. FFTW, CUDA, VkFFT, they should be explicitly registered or used directly be the test.

thewtex:
> To improve the consistency of the factories registered by default by the test driver, just the VNL factories can be registered. If other factories are required by a test, e.g. FFTW, CUDA, VkFFT, they should be explicitly registered or used directly be the test.

@tbirdso @jhlegarreta this will help with #3008 , for example

Modules/Core/TestKernel/include/itkTestDriverIncludeBuiltInFactories.h
tbirdso: Thanks @thewtex !

When available, FFTW implementations are registered by default alongside VNL implementations when `UseITK.cmake` is included in a project consuming ITK (see [https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/FFT/itk-module.cmake#L11](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/FFT/itk-module.cmake#L11)). For the sake of consistency, isn't it better for the TestKernel act in the same way?

To me it seems that consistency is preserved if the TestKernel and any consuming project follow the same rules for default registrations:
- If FFTW is available, register FFTW factories;
- Always register VNL factories

We should also note that the principle purpose of these changes is to reduce boilerplate for FFT factory registrations in ITK tests. We do have tests that expect FFTW factory overrides if FFTW is available, such as [itkFFT1DImageFilterTest.cxx](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/FFT/test/itkFFT1DImageFilterTest.cxx#L109). Removing FFTW factories from default registration in the TestKernel would mean reintroducing that boilerplate or altering test coverage.

dzenanz: I agree with Tom, it is better to have this here, once, than many times over in different tests. FFTW is not built by default, so VNL backend will be exercised much more often.

tbirdso:
Rebased on `master`.

Modules/Core/TestKernel/src/itkTestDriverIncludeRequiredFactories.cxx
tbirdso: @thewtex Do you agree with above discussion that we should register FFTW factories by default to 1) align with behavior in consumer projects and 2) reduce test boilerplate?

Modules/Core/TestKernel/include/itkTestDriverIncludeBuiltInFactories.h
thewtex: This patch provide consistency with the previous Vnl and FFTW testing infrastructure, but that is not what I meant by consistency.

In unit testing, we want

- Tests that check "one thing"
- Consistency in what is tested

> When available, FFTW implementations are registered by default alongside VNL implementations when UseITK.cmake is included in a project consuming ITK (see https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/FFT/itk-module.cmake#L11). For the sake of consistency, isn't it better for the TestKernel act in the same way?

No, we do not want the same behavior for application projects and internal ITK unit testing. For deployed applications, we want the application to be able to load the best FFT backend available at runtime. For unit testing in the TestKernel, we want consistent behavior to occur in the unit tests.

Tests that require an FFT-based filter should be able to run the tests without additional instrumentation, e.g. registering and FFT backend. However, they should get the same backend from the TestKernel.  If the goal is to also test functionality with additional FFT backends, then those backend classes should be instantiated explicitly and explicitly tested in the tests. To test additional backends in unit tests, the factory mechanism should not be used. FFTW is not the only additional backend. There is VkFFT and hopefully more in the future. By explicitly using the FFT backend classes in the unit tests, the tested behavior is consistent and we get coverage. Registering FFT-backends in this way for FFTW cannot be completed with other FFT backends, which may be added in remote module builds that occur after ITK has been built. Also, the factory does not yet have a priority mechanism, and the FFT implementation may depend on the registration or other mechanism. It is better to be explicit with the registered methods.

Modules/Core/TestKernel/src/itkTestDriverIncludeRequiredFactories.cxx
thewtex: @tbirdso sorry, the discussion above was hidden, and I missed it. I followed up there. The goal here should be to provide a default, consistent FFT backend for tests. The behavior in consumer projects is should be different than unit tests, which are expected to have a consistent FFT backend. We can reduce test boilerplate in how the tests are written as much as possible, balancing readability and code coverage.

tbirdso: Thanks for the explanation above @thewtex, I agree with the motive to "test one thing" for each test and see how swapping in FFTW classes via the object factory can violate this idea. I'll move forward with removing default FFTW registration in TestKernel. 

This might require revisiting FFT tests in a couple of places where it is assumed that the "best" backend has been registered first so as to have highest priority, and adding a dedicated test to verify that the object factory is functioning properly with Vnl/FFTW swappable backends. If the scope looks like it will expand beyond small fixes I will add an issue to track and bump to a separate PR.

Modules/Core/TestKernel/include/itkTestDriverIncludeBuiltInFactories.h
tbirdso: Thanks for the explanation; I've pushed a WIP commit for removing default FFTW registrations and will squash into the relevant TestKernel commit if all tests pass.

tbirdso:
@thewtex Could this please be reviewed again?

thewtex (approved):
Looks great!

@tbirdso thanks for your patience and persistence.

PR #2286, opened 2021-02-08
ENH: Increase coverage

jhlegarreta:
Improvements concerning testing and a few related enhancements:
- Fix an RTTI bug concerning a `Superclass` name.
- Remove a duplicate include file.
- Print all member variables in a number of classes.
- Use `itkBooleanMacro` to set the boolean ivar values to On/Off.
- Increase code coverage of a number of classes.

## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [X] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [X] Added test (or behavior not changed)
- [X] Updated API documentation (or API not changed)
- [X] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)

Issue #2233, opened 2021-01-05
IO factory registration for Python modules


dzenanz:
### Description

Automatically register remote module IOs for regular image reading and writing in Python.

### Expected behavior

pip install itk-scanco
```py
image = itk.imread("file.isq") # image is successfully read
```

### Actual behavior

pip install itk-scanco
```py
ImageType = itk.Image[itk.ctype('signed short'), 3]
reader = itk.ImageFileReader[ImageType].New()
imageio = itk.ScancoImageIO.New()
reader.SetImageIO(imageio)
reader.SetFileName(file_name)
reader.Update()
image = reader.GetOutput() # image is successfully read
```

### Additional Information

Starting point for investigation:
https://github.com/InsightSoftwareConsortium/ITK/blob/7f36e8e2ff52f402a4c9377fbdc68c3ba018e6c9/Wrapping/Generators/Python/itk/support/itkExtras.py#L675-L678

thewtex:
Perhaps in one of the other `support` modules we could add a `register_remote_module_factories` function, e.g.

```
_REMOTE_FACTORIES_REGISTERED = False

def register_remote_module_factories():
	if _REMOTE_FACTORIES_REGISTERED:
		return
	
	imageio_factories = filter(lambda x: 'ImageIOFactory' in x, dir(itk))
    for factory in imageio_factories:
       # Do we have to worry about duplicate registration here?
       factory.RegisterOneFactory()
	_REMOTE_FACTORIES_REGISTERED = True
```

Then, call `register_remote_module_factories` from `itk.imread`, `itk.imwrite`.

blowekamp:
Another possibility is to use ITK's dynamic loading of shared libraries which contain the "itkLoad" method for factory registration. Here is a PR related to generating the required code for factories (ImageIO only?) with CMake:

https://github.com/InsightSoftwareConsortium/ITK/pull/134

thewtex:
> ITK's dynamic loading of shared libraries

For performance reasons, we do not want to load all possible shared libraries.

blowekamp:
> For performance reasons, we do not want to load all possible shared libraries.

If the installation structure is such that the libraries with dynamic "itkLoad" methods are in one path/location this could be overcome. However, this may not be the best solution for this problem, just presenting an option.

dzenanz:
Did we decide not to do this? When custom IO is required, it needs to be specified in the `imread` call.

thewtex:
[...]

PR #3003, opened 2021-12-28

jhlegarreta:
- Test the boolean ivars using the `ITK_TEST_SET_GET_BOOLEAN` macro, and
  rely on it to set the boolean ivar values.
- Test exceptions using the `ITK_TRY_EXPECT_EXCEPTION` macro.
- Convert existing hard-coded parameters to test input parameters and
  add parameters to the tests where appropriate so that different
  valued-tests can be added.
- Remove explicit calls to the `Print` method and rely on the basic
  method exercising macro call.
- Remove explicit calls to the `GetNameOfClass` and rely on the basic
  method exercising macro call.

## PR Checklist


dzenanz (approved):
This seems to uncover some [FFT failures](https://open.cdash.org/viewTest.php?onlyfailed&buildid=7647570). Maybe once #3002 is integrated, they will go away? Otherwise awesome!

jhlegarreta:
> This seems to uncover some FFT failures. Maybe once #3002 is integrated, they will go away?

I think they were more of my mistake than actual bugs: when registering the `FFT` factory, the superclass name of the class seems to be `ImageToImageFilter`, rather than the superclass name in the `itkTypeMacro` of the class at issue. @tbirdso might confirm if this behavior is the expected one.

tbirdso:
Thanks for doing this @jhlegarreta ! Yes, the superclass name for each distinct `FFT` backend implementation should be the superclass name in its `itkTypeMacro`, for example the hierarchy for a `VnlForwardFFTImageFilter` would be

`itk::ImageToImageFilter` -> `itk::ForwardFFTImageFilter` -> `itk::VnlForwardFFTImageFilter`

It looks like [FFT failures](https://open.cdash.org/viewTest.php?onlyfailed&buildid=7647570) stem from class name failures rather than superclass name failures at the moment:

> Class name provided does not match object's NameOfClass

My guess is that this because the object factory upcasts each backend implementation to the base FFT class rather than implementation class (Vnl, FFTW, etc). If an FFT test doesn't explicitly request a certain backend we usually don't worry about which backend is being used, so it would be appropriate to pass in to `ITK_EXERCISE_BASIC_OBJECT_METHODS` the name of the base FFT class with `ImageToImageFilter` as its superclass.

Modules/Filtering/FFT/test/itkComplexToComplex1DFFTImageFilterTest.cxx
```diff
@@ -110,12 +112,21 @@ itkComplexToComplex1DFFTImageFilterTest(int argc, char * argv[])
...
       return EXIT_FAILURE;
     }
 
+#ifndef ITK_FFT_FACTORY_REGISTER_MANAGER
+#  if defined(ITK_USE_FFTWD) || defined(ITK_USE_FFTWF)
+    ITK_EXERCISE_BASIC_OBJECT_METHODS(inverse, FFTWComplexToComplex1DFFTImageFilter, ImageToImageFilter);
+#  else
+    ITK_EXERCISE_BASIC_OBJECT_METHODS(inverse, VnlComplexToComplex1DFFTImageFilter, ImageToImageFilter);
+#  endif
+#endif
```
tbirdso, line 128: Following from the previous comment, `inverse` here is an `itk::ComplexToComplex1DFFTImageFilter` with some backend returned from the object factory. We could try downcasting to each backend and exercising object methods, but it's easier to just use the base FFT class:

```cpp
ITK_EXERCISE_BASIC_OBJECT_METHODS(inverse, ComplexToComplex1DFFTImageFilter, ImageToImageFilter);
```

Also note that `#ifndef ITK_FFT_FACTORY_REGISTER_MANAGER` isn't necessary here as `ITK_FFT_FACTORY_REGISTER_MANAGER` will never be defined for ITK tests. #3002 will streamline default factory registration in the test kernel.

tbirdso: After building and experimenting a bit more it looks like the solution may be a bit more complicated than I initially expected. For an object factory override of `ComplexToComplexFFTImageFilter` each of the following calls returns a different name:

```
inverseComplexFilter->GetNameOfClass() -->VnlComplexToComplexFFTImageFilter
inverseComplexFilter->Self::GetNameOfClass() -->ComplexToComplexFFTImageFilter
inverseComplexFilter->Superclass::GetNameOfClass() -->ImageToImageFilter
```

This behavior is expected but a bit opaque as `GetNameOfClass` is overridden via object factory generation. I wonder if it would be pragmatic to add an additional test macro that calls `->Self::GetNameOfClass()` rather than `->Superclass::GetNameOfClass()` so that we can check the base FFT name and ignore the specific implementation?

jhlegarreta: I was not being able to get a clear picture out of the previous explanations, but your last comment 
https://github.com/InsightSoftwareConsortium/ITK/pull/3003#discussion_r776032896 seems to match what I was seeing when testing locally.

I think the current commit b36f5a8 will pass, as I tested it locally.

I am not sure all this is due to factories or backends, so not sure how a new eventual macro should be named to test these cases. And I'm not sure whether we can ensure that in all cases the hierarchy will have at most 3 levels. Maybe downcasting could be used maybe to call those `itkTypeMacro`s that we chase here. Otherwise, if you think that a new macro is necessary/handy for these cases/can better test these in the future, +1 for it.

As for the comment about `#ifndef ITK_FFT_FACTORY_REGISTER_MANAGER` I really did not think much about it, only re-used what was in place when registering the factories.

In either case, IMHO whether testing through downcasting, adding a new macro, and removing the `#ifndef ITK_FFT_FACTORY_REGISTER_MANAGER` guard would be best addressed by subsequent, separate PRs if the current commit's CIs come green.

jhlegarreta: ~I think the current commit b36f5a8 will pass, as I tested it locally.~

Failed again :man_facepalming:, so I am commenting the `ITK_EXERCISE_BASIC_OBJECT_METHODS` call. I think best is to address it separately.

tbirdso: Documented thoughts and potential resolutions in #3008 , feel free to continue discussion there

Modules/Filtering/FFT/test/itkComplexToComplexFFTImageFilterTest.cxx
```diff
@@ -88,6 +88,10 @@ transformImage(const char * inputImageFileName, const char * outputImageFileName
 
   using ComplexFilterType = itk::ComplexToComplexFFTImageFilter<ComplexImageType>;
   auto inverseComplexFilter = ComplexFilterType::New();
+
+  ITK_EXERCISE_BASIC_OBJECT_METHODS(inverseComplexFilter, ComplexToComplexFFTImageFilter, ImageToImageFilter);
```
tbirdso, line 92: This line likely needs to be commented out as well until as resolution is reached for how to handle name checks on upcast classes per #3008

Modules/Filtering/FFT/test/itkComplexToComplex1DFFTImageFilterTest.cxx
jhlegarreta: :+1: Thanks @tbirdso.

Modules/Filtering/FFT/test/itkComplexToComplexFFTImageFilterTest.cxx
jhlegarreta: :+1: `itkVnlComplexToComplexFFTImageFilterTest.cxx` does it as well. Done in 38e9b6d. Let's see if this passes now. Thanks for giving time to this.

Modules/Filtering/FFT/test/itkComplexToComplex1DFFTImageFilterTest.cxx
tbirdso: FYI #3010 addresses the issue with name checking upcasted classes with `ITK_EXERCISE_BASIC_OBJECT_METHODS`. Fine to either uncomment here or push to a subsequent PR.

dzenanz: Rebasing on `master` and uncommenting the tests calls if preferred.

Modules/Core/Mesh/test/itkSimplexMeshAdaptTopologyFilterTest.cxx
```diff
@@ -19,10 +19,17 @@
...
+#include "itkTestingMacros.h"
 
 int
-itkSimplexMeshAdaptTopologyFilterTest(int, char *[])
+itkSimplexMeshAdaptTopologyFilterTest(int argc, char * argv[])
 {
+  if (argc != 3)
+  {
+    std::cerr << "Missing parameters." << std::endl;
+    std::cerr << "Usage: " << itkNameOfTestExecutableMacro(argv) << " threshold selectionMethod" << std::endl;
```
issakomi, line 30: Not a request, just a side note, FYI.

In _itkNameOfTestExecutableMacro_

```cpp
#define itkNameOfTestExecutableMacro(argv)                                                                             \
  [argv](const std::string & functionName) {                                                                           \
    return ((argv == nullptr) || (argv[0] == nullptr) || (argv[0][0] == '\0')) ? ("<" + functionName + " executable>") \
                                                                               : argv[0];                              \
  }(__func__)
```

the expression `((argv == nullptr) || (argv[0] == nullptr) || (argv[0][0] == '\0'))` will never evaluate to _true_. If _argv_ is there, _argv[0]_ is always name of the executable, isn't it?
Also for usage with intentional _nullptr_ (to print name of function) it is problematic, with GCC 8.3
`std::cout << tkNameOfTestExecutableMacro(nullptr)<< std::endl;`
or
`std::cout << itkNameOfTestExecutableMacro((char**)nullptr) << std::endl;`
doesn't compile, only
```cpp
	char ** x = nullptr;
	std::cout << itkNameOfTestExecutableMacro(x) << std::endl;
```
So if fact the macro is useless code
if someone wishes name of the executable
`std::cout << argv[0] << std::endl;` does exactly the same
or if someone wishes name of function
`std::cout << "<" __func__ << ">" << std::endl;`.

Modules/Filtering/FFT/test/itkComplexToComplex1DFFTImageFilterTest.cxx
jhlegarreta: Done in 5643fdb. The concerned tests are now passing locally. Thanks for the time given to this @tbirdso.

jhlegarreta:
Folks, the tests are passing on Windows, but not for Linux/macOS when uncommenting the `ITK_EXERCISE_BASIC_OBJECT_METHODS` statements in the relevant Vnl/FFTW backend-powered tests in commit 5643fdb:
https://dev.azure.com/itkrobotwindow/ITK.Windows/_build/results?buildId=7352&view=results

I had tested locally on a Windows machine before posting https://github.com/InsightSoftwareConsortium/ITK/pull/3003#discussion_r776416206, so I do not understand how these statements are interpreted differently across different compilers.

I do not have a Linux/macOS machine at hand to be able to try to investigate this. 

Also, note that in the https://github.com/InsightSoftwareConsortium/ITK/pull/2286#discussion_r575685116  I cross-referenced in the issue #3008, I was experiencing this very same behavior (passing on Windows/MSVC, other CIs reporting errors) for another class.

I have push forced again in ddd6563 the files commenting out the concerned statements.

I think best is to have it merged as it is, and investigate the issue separately. Otherwise, this is blocking changes to the other 90 files.

dzenanz (approved):

dzenanz:
Thanks for this large contribution!

tbirdso:
Thanks for testing @jhlegarreta , I agree with the decision to go ahead and merge. I'll investigate different behavior among platforms as this may relate to #2233 as well.

PR #3007, opened 2021-12-28
COMP: Allow setting image spacing via SpacingType::Filled

### LINKED ISSUE #3006: Cannot set image spacing via SpacingType::Filled
This code snippet does not compile:
```cpp
  using ImageType = itk::Image<short, 3>;
  auto image = ImageType::New();
  image->SetSpacing(ImageType::SpacingType::Filled(0.02));
```
producing this error message:
```text
1>tester.cpp
1>M:\Dev\Tester\tester.cpp(40,57): error C2664: 'void itk::ImageBase<3>::SetSpacing(const itk::Vector<double,3> &)': cannot convert argument 1 from 'itk::FixedArray<T,3>' to 'const itk::Vector<double,3> &'
1>        with
1>        [
1>            T=double
1>        ]
1>M:\Dev\Tester\tester.cpp(40,57): message : Reason: cannot convert from 'itk::FixedArray<T,3>' to 'const itk::Vector<double,3>'
1>        with
1>        [
1>            T=double
1>        ]
1>M:\Dev\Tester\tester.cpp(40,51): message : No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called
1>M:\CDev\ITK-git\Modules\Core\Common\include\itkImageBase.h(422,3): message : see declaration of 'itk::ImageBase<3>::SetSpacing'
```
This is mind boggling.

dzenanz:
Closes #3006.

### DIFF
--- a/Modules/Core/Common/include/itkVector.h
+++ b/Modules/Core/Common/include/itkVector.h
@@ -129,6 +129,10 @@ public:
   Vector(const Vector<TVectorValueType, NVectorDimension> & r)
     : BaseArray(r)
   {}
+  template <typename TVectorValueType>
+  Vector(const FixedArray<TVectorValueType, NVectorDimension> & r)
+    : BaseArray(r)
+  {}
   Vector(const ValueType r[Dimension])
     : BaseArray(r)
   {}


dzenanz:
This does not compile: `Modules/Core/Transform/include/itkBSplineTransformInitializer.hxx:186:34: error: ambiguous overload for 'operator-' (operand types are 'PointType {aka itk::Point<float, 2>}' and 'itk::BSplineTransformInitializer<itk::BSplineTransform<double, 2, 3>, itk::Image<float, 2> >::OriginType {aka itk::Point<double, 2>}')`.  I don't know whether the problem is in this PR or in itkBSplineTransformInitializer. `transformDomainOrigin` might need to be cast to `PointType` there. What do the others think? @N-Dekker @jhlegarreta

Modules/Core/Common/include/itkVector.h
```diff
@@ -129,6 +129,10 @@ class ITK_TEMPLATE_EXPORT Vector : public FixedArray<T, NVectorDimension>
   Vector(const Vector<TVectorValueType, NVectorDimension> & r)
     : BaseArray(r)
   {}
+  template <typename TVectorValueType>
+  Vector(const FixedArray<TVectorValueType, NVectorDimension> & r)
+    : BaseArray(r)
```
N-Dekker, line 134: If possible, I would declare this constructor `explicit`. Because I think the original idea was that every Vector is a FixedArray, but _not_ every FixedArray is a Vector.

jhlegarreta: Not sure where the error lies: do we test somewhere that the `-` operator can be applied to the basic underlying `itk::Point` types with different pixel types? As simple and trivial as it may seem when `float` and `double` are involved, maybe doing that in an `itkPointTest` may help in discarding that as the cause.

jhlegarreta: Or maybe it is closer to what Niels mentions, and the problem does not lie in the subtraction, but the assignation of the result to an `itk::VectorType`. Maybe it has to be done in a `for` loop to avoid the warning?

N-Dekker:
Instead of adding yet another static `Filled` member function, maybe we could also add a generic `MakeFilled` function template:

    template <typename TContainer, typename TValue>
    TContainer
    MakeFilled(const TValue & value)
    {
      TContainer result;
      std::fill(result.begin(), result.end(), value);
      return result;
    }

Which would support fixed ITK arrays, as well as `std::array`, for example:

    auto spacing = MakeFilled<SpacingType>(1.0);             // = { 1.0, 1.0 }, for a 2D spacing
    auto rgbPixel = MakeFilled<itk::RGBPixel<uint8_t>>(255); // = { 255, 255, 255 };
    auto stdArray = MakeFilled<std::array<int, 4>>(42);      // = { 42, 42, 42, 42 };

What do you think?

dzenanz:
Sounds good! Can you turn that into a PR? Feel free to [auto-close](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) this one from there.

Leengit:
> Instead of adding yet another static Filled member function, maybe we could also add a generic MakeFilled function template:

Good idea.

Instead of having it as a parameter, we might want to infer `TValue` as the type of `*result.begin()`.  Alternatively, we would continue to accept both parameters but do some SFINAE magic with `enable_if` to enforce that `TValue` is the same type as `*result.begin()`.  The advantage of this is that we won't get code duplication; we avoid having compiled units (and correspondingly multiple breakpoints for debugging) for `MakeFilled<TContainer, int>`, `MakeFilled<TContainer, float>`, and `MakeFilled<TContainer, double>` when the container contains doubles.

N-Dekker:
@Leengit Nice idea! Would you like to do the PR? Then I'll do the nitpicking 😸 

BTW, maybe it should also have a check that `TContainer().empty()` is false. For a container that is empty by default, it doesn't make much sense.

Leengit:
I created Issue #3230 and put the type manipulations in the code same there.  @N-Dekker if you figure out where to put it, I can do the nitpicking for you.

dzenanz:
Closing in favor of #3247.

Issue #3008, opened 2021-12-28
Handling `GetNameOfClass` for upcast instances


tbirdso:
## Motivation

ITK tests rely on the `ITK_EXERCISE_BASIC_OBJECT_METHODS` macro for unit testing individual classes, including checking that class and superclass names match expectations. Investigation in #3003 reveals that upcast instances returned from object factory instantiation may have inconsistent naming which may require more detailed test handling than existing macros allow.

## Details

FFT classes are a concrete example of how the ITK object factory can be used to dynamically swap in different backends for a given class interface at runtime. `itk::ComplexToComplexFFTImageFilter` is a base FFT class which defines an interface and relies on the object factory to implement a backend such as Vnl or FFTW for FFT computation. The class hierarchy looks like this:

> `itk::ImageToImageFilter` -> `itk::ComplexToComplexFFTImageFilter` -> `itk::VnlComplexToComplexFFTImageFilter`

If a filter is instantiated through the object factory:

```cpp
itk::ComplexToComplexFFTImageFilter::Pointer inverseComplexFilter = itk::ComplexToComplexFFTImageFilter::New();
```

Then several methods are available to determine the name of the class and its superclass:
```
inverseComplexFilter->GetNameOfClass() --> VnlComplexToComplexFFTImageFilter
inverseComplexFilter->Self::GetNameOfClass() --> ComplexToComplexFFTImageFilter
inverseComplexFilter->Superclass::GetNameOfClass() --> ImageToImageFilter
```

This demonstrates that `GetNameOfClass` is overridden even though the filter object is otherwise referenced as an instance of the `itk::ComplexToComplexFFTImageFilter` base class inline.

[`ITK_EXERCISE_BASIC_OBJECT_METHODS`](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/TestKernel/include/itkTestingMacros.h#L83) internally calls `GetNameOfClass` on the instance directly and fails if the returned value does not match the string passed into the macro. Generally this is useful for validating that the object has been constructed within its expected hierarchy, however object factory instantiation intentionally violates the standard hierarchy by providing a subclass instance that has been upcast to the base class. 

## Potential Resolutions

- Avoid overriding the class name directly: it seems a little strange to allow the base class knowledge of its subclass, but is only a weak coupling as it currently exists and would require nontrivial effort to rearchitect. Would not favor this option.
[...]

PR #3010, opened 2021-12-28
ENH: Query class name rather than subclass in test macro

tbirdso:
Addresses #3008 by weakening `ITK_EXERCISE_BASIC_OBJECT_METHODS` so that `object->GetNameOfClass()` no longer needs to match the user-defined string value for testing. `object->Self::GetNameOfClass()` is queried instead. This allows the check to succeed for subclasses which have been upcast to the expected class, which is a necessary function of object factory instantiation.

## PR Checklist


tbirdso:
All checks are passing with these changes, which makes me believe that querying the base name with `object->Self::GetNameOfClass()` more closely matches the original intent and existing usage of `ITK_EXERCISE_BASIC_OBJECT_METHODS` in checking the base name rather than the potential subclass name.

dzenanz (approved):

thewtex (approved):
:+1:

Issue #2959, opened 2021-12-15
ImageIOFactory read tiff image get error pixel type.


Hconk:
### Description
I use Photoshop/ImageJ open the test image ,  both get real pixel type of 16bit. But ITK TIFFIO get UCHAR( 8bit ).
![image](https://user-images.githubusercontent.com/8470001/146147676-e132b29d-5436-47b1-8141-eea64e929bac.png)
![image](https://user-images.githubusercontent.com/8470001/146147254-3f1d7d7b-8f95-44aa-8158-0f53fd5b718e.png)

### Steps to Reproduce
```C++
#include <itkImageFileReader.h>

int
main(int, char**)
{
  std::string filename = "../tiff_bug_1.tif";
  itk::ImageIOBase::Pointer imageIO = itk::ImageIOFactory::CreateImageIO(
    filename.c_str(), itk::ImageIOFactory::ReadMode);

  if (imageIO.IsNull()) {
    std::cout << "fail create Image io." << std::endl;
    return false;
  }

  imageIO->SetFileName(filename);
  imageIO->ReadImageInformation();

  const auto pixelType = imageIO->GetComponentType();
  std::cout << "image pixel type: " << pixelType << std::endl;
  std::cout << "image pixel type: "
            << imageIO->GetComponentTypeAsString(pixelType) << std::endl;
  std::cout << "image dim: " << imageIO->GetNumberOfDimensions() << std::endl;
  return 0;
}
```

### Expected behavior
Test image real pixel type is unsigned shot(16bit).

### Actual behavior
```
image pixel type: 1
image pixel type: unsigned_char
image dim: 2
```

### Reproducibility

everytime.

### Versions
ITK Version: 5.1.0/5.2.0/master latest

### Environment

CMake: 3.21.1
Compiler: gcc4.5/vs2017

### Additional Information
test data 1:
https://drive.google.com/file/d/1O3PCYt6nz1HKuvnZxtbI_U3fox582yAt/view?usp=sharing
sha256: 9201f1ac5888d6c3efbf3055fd6505c84e100383e4f5bb14d7237b1187f77186

test data 2:
https://drive.google.com/file/d/17I1ua6HOtIKV7jQId-10ioJCra6s6i9Z/view?usp=sharing
sha256: b693140a8fdd76765b5091732a0b7db902621b7873588819c783282d9598dd73

dzenanz:
[...]

PR #3013, opened 2021-12-29
BUG: TIFF IO: exception if orientation is not supported

issakomi:
Throw exception to avoid wrong interpretation of a image (`RGBA unsigned char` instead `SCALAR 16-bits`).

_Please note that is quick and minimal fix for the problem!_

S. https://github.com/InsightSoftwareConsortium/ITK/issues/2959

CC @blowekamp 

**Edit**:
In [ReadGenericImage](https://github.com/InsightSoftwareConsortium/ITK/blob/6682f865b39fc0fa16e6483536e31ace8780f875/Modules/IO/TIFF/src/itkTIFFImageIO.cxx#L1368) is similar block, not reached on _ReadImageInformation_ and _Read_ with test images (s. issue).

### DIFF
--- a/Modules/IO/TIFF/src/itkTIFFImageIO.cxx
+++ b/Modules/IO/TIFF/src/itkTIFFImageIO.cxx
@@ -488,6 +488,12 @@ TIFFImageIO::ReadImageInformation()
 
   if (!m_InternalImage->CanRead())
   {
+    if (!(m_InternalImage->m_Orientation == ORIENTATION_TOPLEFT ||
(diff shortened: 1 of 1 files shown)


blowekamp:
This does not look correct to me and there is not test images to demonstrate correctness or incorrectness in test.

The code path which goes through TIFFReadRGBAImageOriented supports these orientation.

issakomi:
There are test images in the discussion. Very clear.

issakomi:
S. https://github.com/InsightSoftwareConsortium/ITK/issues/2959#issuecomment-1002896569

PR #3014, opened 2021-12-29
Progress reporting improvements

### LINKED ISSUE #2875: Some filters report progress both from theader and from themselves
### Description
In ITK 5 there's a possibility to report progress from threader. It can be controlled by flag ThreaderUpdateProgressOff() or ThreaderUpdateProgressOn(). 
Right now, 4 filters report their progress both from themselves and from threader, and because of the bug with progress overflow  that I reported earlier (https://github.com/InsightSoftwareConsortium/ITK/issues/2874), the progress overflows: it reaches 100% once, then it goes to 0%, and then again to 100%.
Those filters are:
itk::ResampleImageFilter, itk::LabelMapToLabelImageFilter, itk::LabelShapeKeepNObjectsImageFilter, itk::LabelMapFilter.

Even after the bug with overflow will be fixed, it is still wrong to report progress from both threader and filter, because it will reach 100% before it is actually 100%, so in any case it is needed to add this->ThreaderUpdateProgressOff() in their constructors.

### Steps to Reproduce

Here's the code that reproduces the bug for itk::ResampleImageFilter. The bug is similar for the rest of the filters.
https://gist.github.com/ozavalistyi/9a234eacb36316d922c0a9571ee13db7
### Expected behavior
Progress should be reported up to 100% and then stop.
The output of the code above should not contain "ERROR: Progress overflowed".

### Actual behavior
Progress is reported up to 100% once, then it overflows and is reported from 0% to 100% again (
The output of the above code contains "ERROR: Progress overflowed".
The output is attached to gist, that I provided.

### Reproducibility
100%

### Versions

ITK 5.1.2
But I looked at the latest master, and those filters do not contain this->ThreaderUpdateProgressOff() in their constructors, so it is reproducible with the latest ITK also.
### Environment

Windows 10

dzenanz:
Related to #2875.

### DIFF
--- a/Modules/Core/Common/include/itkProgressAccumulator.h
+++ b/Modules/Core/Common/include/itkProgressAccumulator.h
@@ -138,6 +138,8 @@ private:
     // The tags for adding/removing observers to mini-pipeline filter
     unsigned long ProgressObserverTag;
     unsigned long StartObserverTag;
+
+    float AccumulatedProgress{ 0.0 };
   };
 
   /** A callback function that is called by the progressing filters */
--- a/Modules/Core/Common/include/itkSimpleFilterWatcher.h
+++ b/Modules/Core/Common/include/itkSimpleFilterWatcher.h
@@ -21,6 +21,7 @@
 #include "itkCommand.h"
 #include "itkProcessObject.h"
 #include "itkTimeProbe.h"
+#include <mutex>
 
 namespace itk
 {
@@ -186,6 +187,7 @@ protected:
   {
     if (m_Process)
     {
+      std::lock_guard<std::mutex> outputSerializer(m_ProgressOutput);
       m_Steps++;
       if (!m_Quiet)
       {
@@ -290,6 +292,7 @@ private:
   bool                        m_TestAbort{ false };
   std::string                 m_Comment;
   itk::ProcessObject::Pointer m_Process;
+  std::mutex                  m_ProgressOutput;
 
   using CommandType = SimpleMemberCommand<SimpleFilterWatcher>;
   CommandType::Pointer m_StartFilterCommand;
--- a/Modules/Core/Common/src/itkPoolMultiThreader.cxx
+++ b/Modules/Core/Common/src/itkPoolMultiThreader.cxx
@@ -205,7 +205,7 @@ PoolMultiThreader::ParallelizeArray(SizeValueType             firstIndex,
         do
         {
           status = m_ThreadInfoArray[i].Future.wait_for(threadCompletionPollingInterval);
-          if (filter)
+          if (filter && status == std::future_status::timeout)
           {
             filter->IncrementProgress(0);
           }
@@ -297,7 +297,7 @@ PoolMultiThreader::ParallelizeImageRegion(unsigned int         dimension,
           do
           {
             status = m_ThreadInfoArray[i].Future.wait_for(threadCompletionPollingInterval);
-            if (filter)
+            if (filter && status == std::future_status::timeout)
             {
               filter->IncrementProgress(0);
             }
--- a/Modules/Core/Common/src/itkProgressAccumulator.cxx
+++ b/Modules/Core/Common/src/itkProgressAccumulator.cxx
@@ -116,7 +116,13 @@ ProgressAccumulator::ReportProgress(Object * who, const EventObject & event)
     FilterRecordVector::iterator it;
     for (it = m_FilterRecord.begin(); it != m_FilterRecord.end(); ++it)
     {
-      m_AccumulatedProgress += it->Filter->GetProgress() * it->Weight;
+      float progress = it->Filter->GetProgress();
+      if (progress != it->AccumulatedProgress)
+      {
+        m_AccumulatedProgress += progress * it->Weight;
+        it->AccumulatedProgress = 0.0; // this filter is now active
+      }
+      // else skip this filter, as it has finished
     }
 
     // Update the progress of the client mini-pipeline filter
@@ -154,7 +160,8 @@ ProgressAccumulator::ReportProgress(Object * who, const EventObject & event)
         // and then reset this filter's progress.
         // It is not necessary to call UpdateProgress(0.0f) explicitly on the filter because this is done
         // automatically when the filter is restarted.
-        m_BaseAccumulatedProgress += it->Filter->GetProgress() * it->Weight;
+        it->AccumulatedProgress = it->Filter->GetProgress();
+        m_BaseAccumulatedProgress += it->AccumulatedProgress * it->Weight;
       }
     }
   }
--- a/Modules/Core/Common/src/itkProgressReporter.cxx
+++ b/Modules/Core/Common/src/itkProgressReporter.cxx
@@ -58,9 +58,15 @@ ProgressReporter::~ProgressReporter()
   // Only thread 0 should update progress.
   if (m_ThreadId == 0 && m_Filter)
   {
-    // Set the progress to the end of its current range.  The filter has
-    // finished.
-    m_Filter->UpdateProgress(m_InitialProgress + m_ProgressWeight);
+    // Set the progress to the end of its current range.
+    // Make sure it increases the progress, in case multiple reporters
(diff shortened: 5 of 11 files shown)


jhlegarreta (commented):
Have not tried, but wondering whether
> Now the progress fills up twice faster so it reaches 100% when only 50% of work is done.

is the expected behavior, or an expected limitation (e.g. if it cannot be set to 50% when half of the job has been done, and to be 100% only when the job has been completed), whether the wording is accurate or else whether it is a misunderstanding of mine.

Modules/Filtering/LabelMap/include/itkLabelMapToLabelImageFilter.hxx
```diff
@@ -29,6 +29,7 @@ template <typename TInputImage, typename TOutputImage>
 LabelMapToLabelImageFilter<TInputImage, TOutputImage>::LabelMapToLabelImageFilter()
 {
   m_OutputImage = nullptr;
+  this->ThreaderUpdateProgressOff();
```
blowekamp, line 32: The base class `LabelMapFilter` is also disabling this. Is this really needed?

Modules/Filtering/LabelMap/include/itkLabelShapeKeepNObjectsImageFilter.hxx
```diff
@@ -29,6 +29,7 @@ LabelShapeKeepNObjectsImageFilter<TInputImage>::LabelShapeKeepNObjectsImageFilte
   m_NumberOfObjects = 1;
   m_ReverseOrdering = false;
   m_Attribute = LabelObjectType::NUMBER_OF_PIXELS;
+  this->ThreaderUpdateProgressOff();
```
blowekamp, line 32: LabelShaeKeepNObjectsImageFilter is also derived from InPlaceLabelMap/LabelMapFilter

dzenanz:
Before 840bae7 progress would go from 0% to 100% twice, overflowing from 100% to 0% about halfway. After 840bae7, the progress fills up to 100% during first half of the work, and stays there the second half of the work. Now, the progress fills up to 50% during first half of the work, and fills up to 100% during second half of the work.

In all instances, the progress was usually being filled smoothly.

blowekamp (approved):
Looks good. I just would like to know why it is needed to set is multiple times in a class hierarchy of constructors

Modules/Filtering/LabelMap/include/itkLabelMapToLabelImageFilter.hxx
dzenanz: Uh-oh. I didn't look at inheritance hierarchy.

I checked several children of `LabelMapFilter`, they all handle progress reporting themselves. I will remove this duplication.

dzenanz:
@jhlegarreta I updated the commit message.

jhlegarreta (approved):
Thanks for the patch set and the commit message update, now it makes more sense to me.

After Brad's comment, I'm wondering whether setting/updating the threader update progress report flag should be disabled in child classes/whether that makes sense.

dzenanz:
I have a more comprehensive solution. I am testing it locally.

thewtex:
@dzenanz please start commit messages with capital letters for the ChangeLog.

dzenanz:
There is something wrong with progress updates the second time filter is run:
```text
1570: Trying filter->UpdateLargestPossibleRegion()
1570: -------- Start GradientMagnitudeRecursiveGaussianImageFilter "" Progress Quiet
1570: Filter took 0.0594145 seconds.
1570: -------- End GradientMagnitudeRecursiveGaussianImageFilter ""
1570:
1570: itk::ExceptionObject (000000D8CD2FE3C8)
1570: Location: "void __cdecl itk::SimpleFilterWatcher::EndFilter(void)"
1570: File: C:\Dev\ITK-git\Modules\Core\Common\include\itkSimpleFilterWatcher.h
1570: Line: 283
1570: Description: ITK ERROR: GradientMagnitudeRecursiveGaussianImageFilter(000000D8CD2FED70): Filter does not have progress.
1570:
1570:
1570:   In C:\Dev\ITK-git\Modules\Filtering\ImageGradient\test\itkGradientMagnitudeRecursiveGaussianFilterTest.cxx, line 217
1/1 Test #1570: itkGradientMagnitudeRecursiveGaussianFilterTest ...***Failed    0.47 sec
```

dzenanz:
> something wrong with progress updates the second time filter is run

This is the only problem left. How much of a problem is this? Does someone know how to solve it? I am somewhat tired of this PR. It started as quite a small fix.

dzenanz:
I found out the problem. It will take me some time to clean up the code. I will push the update tomorrow.

thewtex (approved):

Modules/Core/Common/include/itkProgressAccumulator.h
```diff
@@ -138,6 +138,8 @@ class ITKCommon_EXPORT ProgressAccumulator : public Object
     // The tags for adding/removing observers to mini-pipeline filter
     unsigned long ProgressObserverTag;
     unsigned long StartObserverTag;
+
+    bool AccumulatedProgress{ 0.0 };
```
blowekamp, line 142: I am getting an error here with conversion from double to bool.

dzenanz: I changed the variable name and how it is used in code, but forgot to change the type. And I was wondering why does debugger keep representing it as a float. Silly me! Thank you Brad.

blowekamp:
In testing this PR against SimpleITK I got a large number of FFT related test further inailures, and one Progress report test failing. It will need further investigation.

dzenanz:
Thanks for testing. Feel free to push to this this branch if you find some easy fixes.

dzenanz:
PythonLaplacianImageFilterTest deadlocks on my computer too. At least I can reproduce the failure.

dzenanz:
CI Linux Python:
```text
	209 - PythonGradientMagnitudeRecursiveGaussianImageFilterTest (Timeout)
	210 - itkGradientVectorFlowImageFilterPythonTest (Timeout)
	211 - PythonLaplacianImageFilterTest (Timeout)
	215 - PythonGeodesicActiveContourLeftVentricleTest (Timeout)
```

dzenanz:
I have removed the problematic commit from this PR. I will see if I can iron it out in a follow-up PR. I think this can be merged now.

dzenanz:
I removed all far-reaching code from this PR, and left all the non-problematic code here. It probably no longer fixes #2875, so I changed the description accordingly. I plan the real fix for a follow-up PR.

jhlegarreta (commented):
:+1: Thanks @dzenanz. Given that @blowekamp tested a previous version against SimpleITK, may be he would like to take that again, but I'm for merging it as the one-but-last build passed tests, and hope this will as well.

blowekamp:
Having to reset internal filter's progress when re-running seems like a red flag to me. I'd like to test with just re-running a filter in a standalone test to alleviate my concern. I'm not sure I'd have time tomorrow. There are a lot of changes here.

blowekamp (approved):
I tested a few cases that were of concerned to me, and they all behaved correctly 🥳 

Also SimpleITK tests seemed to be OK too (although there are other FFT issues).

Nice work!

dzenanz:
Are the FFT issues captured in #3048?

tbirdso:
> Also SimpleITK tests seemed to be OK too (although there are other FFT issues).

@blowekamp Could you please add @dzenanz  and myself to FFT discussion if there are other open issues?

blowekamp:
When I narrow in down in SimpleITK, I'll create an issue if needed. Here is a build with the failure FYI: https://open.cdash.org/viewTest.php?onlyfailed&buildid=7657537

tbirdso:
At a glance I'm guessing that the SimpleITK tests will need to add factory registrations for VNL FFT classes as the segfaulting tests all seem to require classes that depend on FFT filter instantiations. Whenever you get around to taking a closer look I'd recommend cross-referencing #3002 to see how this is being done in the ITK TestKernel as well as ITK GTests.

PR #3016, opened 2021-12-31
MINC 2021-12-23 (728d19e8)

### LINKED ISSUE #2202: Update libMINC's CMake code
### Description

Configuring ITK with CMake 3.19.2 I get the following warning:
```
CMake Deprecation Warning at Modules/ThirdParty/MINC/src/libminc/CMakeLists.txt:25 (CMAKE_MINIMUM_REQUIRED):
  CMake.

```

We need to update ITK's copy of libMINC, and maybe libMINC itself needs to be updated.

dzenanz:
Closes #2202.


dzenanz:
I guess I will wait until https://github.com/BIC-MNI/libminc/pull/116 is merged, then update again (that should take care of ghostflow check failure).

gdevenyi:
I merged the whitespace correction

Modules/ThirdParty/MINC/src/libminc/CMakeLists.txt
```diff
@@ -49,7 +49,9 @@ IF(NOT LIBMINC_EXTERNALLY_CONFIGURED)
 
 
   OPTION(LIBMINC_MINC1_SUPPORT           "Support minc1 file format, requires NETCDF" OFF)
+  OPTION(LIBMINC_BUILD_EZMINC            "Build C++ EMINC library based on libminc1" ON)
```
gdevenyi, line 52: The addition of this with default value ON may need to be switched off to keep this build the same after the update, since ITK isn't using the EZMinc layer.

dzenanz:
This is ready for merging. I would prefer an approval 😄

hjmjohnson (approved):

PR #3021, opened 2022-01-02
STYLE: Declare local `ImageRegion` variables `const`

N-Dekker:
Replaced lines of code like the following:

    RegionType region;
    region.SetIndex(index);
    region.SetSize(size);

With:

    const RegionType region(index, size);

Following C++ Core Guidelines, August 19, 2021, "By default, make objects immutable"
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rconst-immutable

### DIFF
--- a/Modules/Bridge/VTK/include/itkVTKImageExport.hxx
+++ b/Modules/Bridge/VTK/include/itkVTKImageExport.hxx
@@ -358,9 +358,7 @@ VTKImageExport<TInputImage>::PropagateUpdateExtentCallback(int * extent)
     size[i] = (extent[i * 2 + 1] - extent[i * 2]) + 1;
   }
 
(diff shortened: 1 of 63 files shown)


dzenanz (approved):
I assume that some tool reported opportunities to make Regions const, but the changes were manual?

N-Dekker:
> I assume that some tool reported opportunities to make Regions const, but the changes were manual?

Honestly I did it all manually  😆 I did a global search on substrings like "egion.SetIndex" and "egion.SetSize", and then checked if the corresponding region was just default-constructed a few lines earlier, and unmodified afterwards.

hjmjohnson (approved):
LOVE THIS!! Thank you.

PR #3022, opened 2022-01-02
BUG: Fix testing macro for unix platforms and uncomment FFT checks

tbirdso:
Follow-up to #3010 removing support to check base class names via `ITK_EXERCISE_BASIC_OBJECT_METHODS` when compiled with GCC. This follows from discussion in #2676 where GCC cannot resolve `object->Self` or `object->Superclass` while other compilers have no such issue.

Also uncomments checks in FFT tests relying on this change.

## PR Checklist

### DIFF
--- a/Modules/Core/TestKernel/include/itkTestingMacros.h
+++ b/Modules/Core/TestKernel/include/itkTestingMacros.h
@@ -67,15 +67,6 @@ namespace itk
 #define ITK_EXERCISE_BASIC_OBJECT_METHODS(object, ClassName, SuperclassName)                                           \
   object->Print(std::cout);                                                                                            \
   std::cout << "Name of Class = " << object->GetNameOfClass() << std::endl;                                            \
-  if (!std::strcmp(object->GetNameOfClass(), #ClassName))                                                              \
-  {                                                                                                                    \
-    std::cout << "Class name is correct" << std::endl;                                                                 \
-  }                                                                                                                    \
-  else                                                                                                                 \
-  {                                                                                                                    \
-    std::cerr << "Class name provided does not match object's NameOfClass" << std::endl;                               \
-    return EXIT_FAILURE;                                                                                               \
-  }                                                                                                                    \
   ITK_MACROEND_NOOP_STATEMENT
 #else // not GCC
 #define ITK_EXERCISE_BASIC_OBJECT_METHODS(object, ClassName, SuperclassName)                                           \
--- a/Modules/Filtering/FFT/test/itkComplexToComplex1DFFTImageFilterTest.cxx
+++ b/Modules/Filtering/FFT/test/itkComplexToComplex1DFFTImageFilterTest.cxx
@@ -119,7 +119,7 @@ itkComplexToComplex1DFFTImageFilterTest(int argc, char * argv[])
       return EXIT_FAILURE;
     }
 
-    // ITK_EXERCISE_BASIC_OBJECT_METHODS(inverse, ComplexToComplex1DFFTImageFilter, ImageToImageFilter);
+    ITK_EXERCISE_BASIC_OBJECT_METHODS(inverse, ComplexToComplex1DFFTImageFilter, ImageToImageFilter);
 
     return doTest<FFTInverseType>(argv[1], argv[2], argv[3]);
   }
--- a/Modules/Filtering/FFT/test/itkComplexToComplexFFTImageFilterTest.cxx
+++ b/Modules/Filtering/FFT/test/itkComplexToComplexFFTImageFilterTest.cxx
@@ -89,7 +89,7 @@ transformImage(const char * inputImageFileName, const char * outputImageFileName
   using ComplexFilterType = itk::ComplexToComplexFFTImageFilter<ComplexImageType>;
   auto inverseComplexFilter = ComplexFilterType::New();
 
-  // ITK_EXERCISE_BASIC_OBJECT_METHODS(inverseComplexFilter, ComplexToComplexFFTImageFilter, ImageToImageFilter);
+  ITK_EXERCISE_BASIC_OBJECT_METHODS(inverseComplexFilter, ComplexToComplexFFTImageFilter, ImageToImageFilter);
 
 
   inverseComplexFilter->SetInput(forwardFilter->GetOutput());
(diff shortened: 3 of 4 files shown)


tbirdso:
Linux failure for debugging: [https://open.cdash.org/viewBuildError.php?buildid=7655442](https://open.cdash.org/viewBuildError.php?buildid=7655442)

tbirdso:
`this->Self` appears in other files such as [`itkStreamingProcessObject.cxx`](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/src/itkStreamingProcessObject.cxx#L155) but `object->Self` does not seem to appear outside of its introduction in `itkTestingMacros.h`. Need to investigate further to understand why the public `Self` does not seem to be exposed on Linux.

tbirdso:
@dzenanz @Leengit would appreciate your input on the best path forward here. This PR aims to update the `ITK_EXERCISE_BASIC_OBJECT_METHODS` test macro so that upcast classes pass name checks, however compilation with GCC fails to resolve `object->Self` and `object->Superclass` as discussed in #2676.

I'm not finding much relevant discussion online on GCC failing to resolve public type aliases where MSVC succeeds. Is there speculation as to why this is happening? Is this a potential bug or undefined behavior in GCC?

In lieu of a proper fix, two paths forward are 1) removing name checks altogether so that the macro simply prints and returns for GCC, or 2) documenting the restriction that `ITK_EXERCISE_BASIC_OBJECT_METHODS` should not be used in cases where the instance under consideration may be upcast from a subclass (i.e. object factory instantiation) such that `object->GetNameOfClass()` has been overridden.

Leengit:
I don't see the fix here.  With respect to those past pull requests, I tried some permutations where `Superclass` was preceded earlier in the line with `typename` thinking that that was the issue, but I couldn't get that to work.  I did not try using the `template` keyword instead; that might be worth exploring because `Superclass` and `Self` are usually (always?) templates.

Alternatively, we might try putting the template parameters on the `using` statement itself.  That is, maybe changing the likes of
```c++
using Self = FFTWComplexToComplex1DFFTImageFilter;
```
to
```c++
using Self = FFTWComplexToComplex1DFFTImageFilter<TInputImage, TOutputImage>;
```
would be helpful?  I wish I had something stronger than these shots in the dark.

With regard to the two alternatives at the bottom of @tbirdso's comment, I guess that I have a slight leaning for Alternative 1 over Alternative 2.  The former removes the name checking, but only for one platform; and it means that if we can figure this out one day, we don't have to revert on the policy that Alternative 2 represents.

dzenanz (commented):
I also prefer option 1 (remove the checks for GCC).

jhlegarreta:
@tbirdso thanks for the effort. I am sorry to chime in here. Removing the print statement from the `GCC` case will significantly drop the coverage, as the `GetNameOfClass()` method provided by `itkTypeMacro`:
https://github.com/InsightSoftwareConsortium/ITK/blob/047fa6934ac32711d1d7b84fa6eb81c72db01960/Modules/Core/Common/include/itkMacro.h#L400

will not be called, as the only machine computing/reporting the coverage is a Linux machine, and thus I assume it is using `GCC`.

Can the std output print statement in
https://github.com/InsightSoftwareConsortium/ITK/pull/3022/files#diff-7910dc02a58a8effd0eb44c5d870bbd406c6a0dd35e72e9d640759c19451995bL69

be kept, please?

Printing the superclass could also be left.

I am struggling/have worked hard to raise coverage, and losing these would be a setback. IMO not losing a given coverage level makes it easier to improve it, to locate coverage improvement areas (which also means providing easier ways of testing, probably giving ways to uncover more easily any sort of bug), etc. Hope it makes sense.

A few examples:
- In the past 2 days, we've lost 50 lines from the 89.01% (129034 LOC tested) level: it may not seem much, but raising the coverage by 50 lines takes a non-negligible amount of time.
- Last year we achieved a 90% level, and in 15 days-time we went to 88%. That means ~2000 lines we lost, were added without coverage, or additions/deletions made the balance change in the way we would not want to.

Currently, we do not have good ways to see how a PR changes the coverage, or ways to automatically know which specific files are affected in the nightly coverage reports, so the task is time-consuming/hard once a change gets merged. At times, the line-by-line reports seem not to be accurate, but that inaccuracy is something we need to live with I guess.

Sorry for the lengthy explanation, and for having it done here (whereas it affects all PRs).

Thanks for understanding and the effort to figure this out @tbirdso, and reviewers !

dzenanz:
> object->Superclass::GetNameOfClass() returns a compiler error on GCC

We could print `object->Self::GetNameOfClass()` instead, mainly to increase coverage.

jhlegarreta:
> Sorry for not including you earlier @jhlegarreta , I'll try to do a better job of pulling you in on discussions relating to test infrastructure.

No worries @tbirdso. It is fine. Sometimes I cannot devote time to reviewing PRs.
 
> Thank you for all of your hard work to increase testing coverage, and for taking the time to explain here. I'm not very familiar with our coverage reporting, I see what I assume is the Linux machine you're talking about on CDash, is there somewhere I can go to see the coverage results?

:+1:

There is a `Coverage` category at the bottom of the dahsboard (the build is earlier in the scroll). You can click on the percentage to get the details of the coverage.

cdash_coverage


>> Can the std output print statement be kept, please?

> If you think it will provide more benefit than harm I'm open to keeping this print statement. My main concern is that on GCC we will be printing object->GetNameOfClass() while on other compilers we will be checking object->Self::GetNameOfClass(), which in the case of upcast instances will be two different strings representing different levels in the class hierarchy. This could lead to confusion from different compiler behaviors in any tests dealing with classes that rely on the object factory for instantiation, though this will likely remain a very small percentage of test scenarios.

Yes, I see the problem. However, 1) It is a known problem (maybe you can add a note to issue #3008); 2) I do not think that that many people will be checking the name of class or comparing across OS; 3) The issue, in reality, is not in the testing macro, but somewhere else.

So I'd definitely keep the std call (if you go back to the file's history, the call was added prior to the comparison). Say there are +1000 classes, and hopefully calls to the macro. I would not like to lose 1000 lines of coverage. Even losing 50 hurts.

>> Printing the superclass could also be left.

> I'm not sure that we have a viable way to do this on GCC right now, as object->Superclass::GetNameOfClass() returns a compiler error on GCC only (hence why it was previously removed).

OK, got it. That was not strictly necessary.

Thanks.

Modules/Core/TestKernel/include/itkTestingMacros.h
```diff
@@ -67,15 +67,6 @@ namespace itk
 #define ITK_EXERCISE_BASIC_OBJECT_METHODS(object, ClassName, SuperclassName)                                           \
   object->Print(std::cout);                                                                                            \
   std::cout << "Name of Class = " << object->GetNameOfClass() << std::endl;                                            \
```
tbirdso, line 69: @jhlegarreta I've added back the `std::cout` statement here. Is this sufficient for code coverage, i.e. can we get by without the explicit `std::strcmp` removed in the next line?

jhlegarreta: @tbirdso should be OK. Thanks.

jhlegarreta (approved):

PR #3030, opened 2022-01-04
Doxygen Generation Improvements

blowekamp:
Three things are done in the PR:
 - Use SVG file format for class graphs
 - Use mathjax javascript to render math latex on the client side
 - Replace static datetime stamp in footer with value dynamically set via javascript

The following changes reduce the HTML from 2.7GB to 884MB a 67%  reduction in size. Additionally, without the daily file modifications of the time stamps few files are modified each night, improving client side cache and enabling committing to gh-pages with little daily churn.

Sample GH repo hosting Doxygen:
https://github.com/blowekamp/ITKDoxygen
https://blowekamp.github.io/ITKDoxygen/

SimpleITK has been using these Doxygen changes on GH pages:
https://simpleitk.org/doxygen/latest/html/index.html

Additionally, the ITKDoxygen repo provides a Dockerfile to reproduce the Doxygen so that issue can better be reproduced and changes to the Doxygen generation system can be contributed by the community.

### DIFF
--- a/Documentation/Doxygen/DoxygenFooter.html
+++ b/Documentation/Doxygen/DoxygenFooter.html
@@ -21,7 +21,8 @@
   </small>
 </div>
 <address class="footer"><small>
-$generatedby &#160;<a href="http://www.doxygen.org/index.html">
+Generated on <span id="datetime">unknown</span> for $projectname by &#160;
+<a href="http://www.doxygen.org/index.html">
 <img class="footer" src="$relpath^doxygen.png" alt="doxygen"/>
 </a> $doxygenversion
 </small></address>
--- a/Documentation/Doxygen/DoxygenHeader.html
+++ b/Documentation/Doxygen/DoxygenHeader.html
@@ -14,6 +14,7 @@
 $treeview
 $search
 $mathjax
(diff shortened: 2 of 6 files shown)


hjmjohnson (approved):
These seem like good changes to me.

thewtex (approved):
> Reduce the HTML from 2.7GB to 884MB a 67% reduction in size

Awesome!!

dzenanz (approved):
Trying to search on https://blowekamp.github.io/ITKDoxygen/ resulted in a `search.php` file being downloaded on my computer. Will this problem affect the main deployment?

blowekamp:
> Trying to search on https://blowekamp.github.io/ITKDoxygen/ resulted in a `search.php` file being downloaded on my computer. Will this problem affect the main deployment?

I believe this is related to the 50MB+ search.idx (?) file I deleted which exceeded the GitHub max file size.  Its first job was to share the generated Doxygen HTML, and second as a demonstration of using gh pages and ci, but the later is not 100% there.

jhlegarreta (approved):
:100: Thanks @blowekamp !

PR #3035, opened 2022-01-05
COMP: fix warning triggered by setjmp

SimonRit:
The warning was triggered by GCC 7.4.1
 warning: variable ‘num_comp’ might be clobbered by ‘longjmp’ or ‘vfork’
 [-Wclobbered]

See
https://my.cdash.org/viewBuildError.php?type=1&buildid=2111692

## PR Checklist

### DIFF
--- a/Modules/IO/JPEG/src/itkJPEGImageIO.cxx
+++ b/Modules/IO/JPEG/src/itkJPEGImageIO.cxx
@@ -488,7 +488,7 @@ JPEGImageIO::WriteSlice(std::string & fileName, const void * const buffer)
   {
     itkExceptionMacro(<< "JPEG: image is too large");
   }
-  const int num_comp = this->GetNumberOfComponents();
+  const volatile int num_comp = this->GetNumberOfComponents();
   if (num_comp > MAX_COMPONENTS)
   {
     itkExceptionMacro(<< "JPEG: too many components");


SimonRit:
@issakomi Please review, I don't really understand this code... (I've never used `setjmp`).

issakomi:
> I don't really understand this code... (I've never used setjmp).

I think the idea of _volatile_ is to prevent writing variable to register, setjmp/longjmp (a kind of nonlocal _goto_) clobbers registers. _Never using_ it is a good practice, IMHO, at least with C++.

SimonRit:
Thanks for the explanation. I have added the const qualifier, as suggested.

N-Dekker:
Interesting! It seems to me that it might warn when a variable is declared _before_ a `setjmp`, and this variable is then being used _after_ the `setjmp`. Using C++ exceptions, constructors, destructors, and dynamic memory allocation in combination with low-level C `setjmp` calls might also be risky. 🤔  Anyway, I hope your `volatile` does the job, @SimonRit !

SimonRit:
> Anyway, I hope your `volatile` does the job, @SimonRit !

If getting rid of the warning is "the job", yes! Thanks for your comment.

dzenanz (approved):

thewtex (commented):
Great to see these improvements in our JPEG IO!

issakomi:
Interesting that GCC 7 warns about _num_comp_ and doesn't warn about nearby _width_ and _height_. There is
 `const auto rowbytes = num_comp * width;`
later. I believe that nothing bad can happen here, because that point is reached only once, if setjmp returns 0, otherwise the function exit. Theoretically jumping between _setjmp_ and _longjmp_ can happen many times, so the warnings make sense in some scenarios, by the 2nd or whatever jump _num_comp_ could be clobbered.

```cpp
#include <cstdio>
#include <csetjmp>

int main(int, char**)
{
  jmp_buf env;
  volatile int x = 0;
  if (setjmp(env))
  {
    ++x;
    if (x <= 5)
    {
      printf("%d\n", x);
    }
    else
    {
      printf("return point 1\n");
      return 0;
    }
  }
  longjmp(env, 1);
  printf("return point 2\n");
  return 0;
}
```
```

r@deb:~$ ./a.out 
1
2
3
4
5
return point 1
r@deb:~$
```

If some compiler warns about _height_ and _width_, they could be set _const volatile_ too, IMHO.

PR #3036, opened 2022-01-05
STYLE: Declare local zero-filled `Index` variables `constexpr`

N-Dekker:
Replaced lines of code which look like the following:

    IndexType index;
    index.Fill(0);

With:

    constexpr IndexType index = { { 0 } };

In accordance with C++ Core Guidelines, January 3, 2022:

"By default, make objects immutable"
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rconst-immutable

"Declare an object `const` or `constexpr` unless you want to modify its
value later on"
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-const

### DIFF
--- a/Modules/Core/Common/include/itkConstNeighborhoodIterator.hxx
+++ b/Modules/Core/Common/include/itkConstNeighborhoodIterator.hxx
@@ -210,9 +210,8 @@ template <typename TImage, typename TBoundaryCondition>
 ConstNeighborhoodIterator<TImage, TBoundaryCondition>::ConstNeighborhoodIterator()
 
 {
-  IndexType zeroIndex;
-  zeroIndex.Fill(0);
-  SizeType zeroSize;
+  constexpr IndexType zeroIndex = { { 0 } };
+  SizeType            zeroSize;
   zeroSize.Fill(0);
 
   m_Bound.Fill(0);
--- a/Modules/Core/Common/include/itkBresenhamLine.hxx
+++ b/Modules/Core/Common/include/itkBresenhamLine.hxx
@@ -53,12 +53,12 @@ BresenhamLine<VDimension>::BuildLine(LType Direction, IdentifierType length) -> 
 
   OffsetArray result(length);
 
-  IndexType m_CurrentImageIndex, StartIndex, LastIndex;
+  IndexType m_CurrentImageIndex, LastIndex;
 
   Direction.Normalize();
   // we are going to start at 0
   m_CurrentImageIndex.Fill(0);
-  StartIndex.Fill(0);
+  constexpr IndexType StartIndex = { { 0 } };
   for (unsigned i = 0; i < VDimension; ++i)
   {
     LastIndex[i] = (IndexValueType)(length * Direction[i]);
--- a/Modules/Core/Common/include/itkConstNeighborhoodIteratorWithOnlyIndex.hxx
+++ b/Modules/Core/Common/include/itkConstNeighborhoodIteratorWithOnlyIndex.hxx
@@ -129,8 +129,7 @@ ConstNeighborhoodIteratorWithOnlyIndex<TImage>::GetBoundingBoxAsImageRegion() co
 template <typename TImage>
 ConstNeighborhoodIteratorWithOnlyIndex<TImage>::ConstNeighborhoodIteratorWithOnlyIndex()
 {
-  IndexType zeroIndex;
-  zeroIndex.Fill(0);
+  constexpr IndexType zeroIndex = { { 0 } };
 
   SizeType zeroSize;
   zeroSize.Fill(0);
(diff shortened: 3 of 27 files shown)


blowekamp:
@N-Dekker I don't think I understand what this initialization is doing.

The itk::IndexType in an aggregate type. So for example to initialize a itk::Index<3> object something like `x = {{0, 0, 0}}` would be needed.  So what does `itk::Index<3> x = {{ 0 }}}` do?

thewtex (approved):
:+1:

Modules/Core/Common/include/itkConstNeighborhoodIterator.hxx
```diff
@@ -210,9 +210,8 @@ template <typename TImage, typename TBoundaryCondition>
 ConstNeighborhoodIterator<TImage, TBoundaryCondition>::ConstNeighborhoodIterator()
 
 {
-  IndexType zeroIndex;
-  zeroIndex.Fill(0);
-  SizeType zeroSize;
+  constexpr IndexType zeroIndex = { { 0 } };
+  SizeType            zeroSize;
```
dzenanz, line 214: Is a similar PR for `Size`s in the works? There should be less of those than zero indices. 😄

hjmjohnson (approved):
Very Nice.

dzenanz (approved):

Modules/Core/Common/include/itkConstNeighborhoodIterator.hxx
N-Dekker: > Is a similar PR for `Size`s in the works? There should be less of those than zero indices.

@dzenanz  Well, I can surely take a look at those as well. But first I'd like to see if this `Index` PR can make it onto the master... Wow, three approvals already  😄 

Indeed, zero-`Size`s appear far less popular than zero-`Index`es 😄😄

hjmjohnson: I'd really like to see:
```c++
constexpr RegionType                  region( {{0}}, {{128}} );

constexpr RegionType3D                  region( {{0}}, { {128, 256, 384} } );
```

:). if someone else wants to put the work in.

N-Dekker: @hjmjohnson Thanks for merging! You wrote:

>     constexpr RegionType3D region( {{0}}, { {128, 256, 384} } );

Nice suggestion! However, an object of type `itk::ImageRegion<N>`  cannot be constexpr, because it has virtual member functions!  For performance reasons, it might be worth considering to remove those virtual member functions. Eventually... but that may take some more time.

Now I just realized that  `itk::ImageRegion<N>` has another constructor, which only has a `Size` parameter (zero-initializing the Index automatically). So instead, we could write:

     const RegionType3D region( { {128, 256, 384} } );

Would you like that too?

blowekamp:
> Thanks for asking @blowekamp. `itk::Index<3> x = {{ 0 }}` zero-initializes _all_ the elements of the internal array of `x`. (All three of them, in this case.) This syntax works like that for a very long time already, for any aggregate type, even in C++98/03.

Thanks you. I'm still trying to understand this a little more. I found a couple slash dot posts. And it seems to come down to array initialization. For example:

```
int a[3] = {0};
```

is basically the same thing but without the aggregate class, just the C array member. This seems to be a feature of [array initialization](https://en.cppreference.com/w/c/language/array_initialization). From the linked reference, the following seems to be the key point used here: "All array elements that are not initialized explicitly are zero-initialized." So isn't the following be the same:
```
int a[3] = {};
```

I looked at the assembly generated by godbolt for a couple cases and it appears to be the same. One interesting case is:

```
int a[3] = {1};
```
which initializes an array to `{1,0,0}`.

From this deep dive it seems only the following code is needed:

`itk::Index<3> x = {{ }}`

But now that I have written all this... I recall there have been variety of non-compliant compiler warning/errors that have been encountered with this aggregate array initialization. Perhaps this {{0}} syntax is what was needed to work with all the compilers.

This post seems like it may be a bit of noise now, but I have written it and don't want to delete it. Thanks for making it to the end.

N-Dekker:
@blowekamp Thanks for having another look at C++ array/aggregate initialization!

> From this deep dive it seems only the following code is needed:
>
>      itk::Index<3> x = {{ }}

Indeed! Or even:
 
     itk::Index<3> x{};

Both `= {{ }}` and `{}` would have been OK to me as well. But `itk::Index<3> x = { { 0 } }` is also fine  😄

Anyway, my main aim was to declare those `Index` variables `constexpr`, which is a nice style improvement, and it _might_ even improve the performance somewhat.  😃 (No benchmark available, just hoping that it does!)

Modules/Core/Common/include/itkConstNeighborhoodIterator.hxx
hjmjohnson: Yes. 

Good point on constexpr.  const is what I should have requested.

Leengit: If it were a pointer `itk::ImageRegion<N> *` I could imagine why virtual member functions might not be known until runtime, but since it is not a pointer, what's the issue with virtual member functions, especially if those member functions aren't used in the constructor?  I need to go back to the textbooks....

PR #3038, opened 2022-01-05
DOC: Fix possessive "its" usage

### LINKED ISSUE #2997: Check and change typos `it's` -> `its`
The term `it's` can be found 171 times in ITK cxx, h, and hxx files. Check which of these instances should be changed to `its`. From https://github.com/InsightSoftwareConsortium/ITK/pull/2996#discussion_r775281256.

jamesobutler:
This closes #2997. I have reviewed and updated ITK files (not third-party sources) to update the correct usage of the possessive "Its". Note that I have not checked the inverse if any "its" usage should really be the contraction "it's", but this forgetting of an apostrophe for the contraction is seemingly not a common mistake.

Possessive "Its"
Correct: "Its class does something"

Contraction "It's" means "It is"
Correct: "It's correct to assume that something"

### DIFF
--- a/Documentation/ReleaseNotes/3.12.md
+++ b/Documentation/ReleaseNotes/3.12.md
@@ -1165,10 +1165,10 @@ Changes in this Release
         STYLE: according to kwstyle.
 
     Code/Common/itkImageBase
-        BUG: 8490: ImageBase::Initialized method modifies it's MTime requiring pipelines to re-executed. Added InitializeBufferedRegion method which does not call Modify as a protected method
+        BUG: 8490: ImageBase::Initialized method modifies its MTime requiring pipelines to re-executed. Added InitializeBufferedRegion method which does not call Modify as a protected method
 
     Code/Common/itkImageBase
-        BUG: 8490: ImageBase::Initialized method modifies it's MTime requiring pipelines to re-executed. Added InitializeBufferedRegion method which does not call Modify as a protected method
+        BUG: 8490: ImageBase::Initialized method modifies its MTime requiring pipelines to re-executed. Added InitializeBufferedRegion method which does not call Modify as a protected method
 
     Code/Common/itkImageRandomNonRepeatingConstIteratorWithIndex
         STYLE: according to kwstyle.
--- a/CMake/ITKModuleRemote.cmake
+++ b/CMake/ITKModuleRemote.cmake
@@ -178,7 +178,7 @@ function(itk_fetch_module _name _description)
   if(NOT DEFINED Module_${_name} )
     option(Module_${_name} "(Remote-${MODULE_COMPLIANCE_LEVEL}) ${_description}" OFF)
   else()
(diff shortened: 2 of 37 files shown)


seanm:
Nice!

But the third party changes, like to GDCM, should be dropped.

Modules/ThirdParty/GDCM/src/gdcm/Source/DataStructureAndEncodingDefinition/gdcmCP246ExplicitDataElement.cxx
```diff
@@ -26,7 +26,7 @@ VL CP246ExplicitDataElement::GetLength() const
     {
     assert( ValueField->GetLength().IsUndefined() );
     Value *p = ValueField;
-    // If this is a SQ we need to compute it's proper length
+    // If this is a SQ we need to compute its proper length
```
dzenanz, line 29: This is the only instance of changing third party code.

Documentation/ReleaseNotes/3.12.md
```diff
@@ -1165,10 +1165,10 @@ Changes in this Release
         STYLE: according to kwstyle.
 
     Code/Common/itkImageBase
-        BUG: 8490: ImageBase::Initialized method modifies it's MTime requiring pipelines to re-executed. Added InitializeBufferedRegion method which does not call Modify as a protected method
+        BUG: 8490: ImageBase::Initialized method modifies its MTime requiring pipelines to re-executed. Added InitializeBufferedRegion method which does not call Modify as a protected method
```
dzenanz, line 1168: I am not sure whether we should be updating old release notes.

hjmjohnson (approved):
Nice.  Thanks for this change.

jamesobutler:
> But the third party changes, like to GDCM, should be dropped.

Sorry, that one edit of a third party source snuck in there. I've updated the commit to remove it.

Modules/ThirdParty/GDCM/src/gdcm/Source/DataStructureAndEncodingDefinition/gdcmCP246ExplicitDataElement.cxx
hjmjohnson: @jamesobutler   Could you please revert the change to the GDCM file, `git commit --amend` and force push this PR again?

Documentation/ReleaseNotes/3.12.md
hjmjohnson: @dzenanz  I think we should update release notes for grammatical changes.

jamesobutler:
Also if this is more of `STYLE` commit I can change it to that. It is modifying text in documentation files and also comments in other files.

dzenanz:
I am fine with `DOC` prefix.

hjmjohnson:
@jamesobutler This definitely seems like a DOC issue.

Incorrect grammar should not be a matter of preference.

PR #3042, opened 2022-01-05
Gdcm update 2022 01 05

seanm:


seanm:
This will fix some of the few remaining `-Wreserved-id-macro` warnings.

dzenanz:
I guess we need to increase allowed file sizes [here](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/ThirdParty/GDCM/UpdateFromUpstream.sh#L44) and [here](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/ThirdParty/GDCM/src/gdcm/.gitattributes). While at it, we should probably clean up [main `.gitattributes`](https://github.com/InsightSoftwareConsortium/ITK/blob/v5.3rc02/.gitattributes#L55-L60).

seanm:
Which exactly to update though?  For example, one error message says `commit c1974cb creates blob 0e6f95c9ebff44cde44a50323e9549552ab44d2f at Modules/ThirdParty/GDCM/src/gdcm/Source/DataDictionary/privatedicts.xml with size 952489 bytes (930.17 KiB) which is greater than the maximum size 300000 bytes (292.97 KiB). If the file is intended to be committed, set the hooks-max-size attribute on its path.`  But where is this 300000 maximum?  It doesn't seem to be in any of 3 files you linked to.

dzenanz:
Changes to the files I mentioned should be done before triggering the update. 300000 might be the default limit.

seanm:
Something is still confusing though:

```
commit 21fbc51 creates blob 802c69d9f1a79607e3973ad9d49b48598ca983d0 at Source/InformationObjectDefinition/Part3.xml with size 2336006 bytes (2281.26 KiB) which is greater than the maximum size 1048576 bytes (1024.00 KiB). If the file is intended to be committed, set the hooks-max-size attribute on its path.
```

Where is this maximum size 1048576 bytes from?

The three files you mentioned have:

```
Source/InformationObjectDefinition/Part3.xml hooks-max-size=4336000
```

and 

```
/Source/InformationObjectDefinition/Part3.xml hooks.MaxObjectKiB=4096
```

Both those number are *already* > 2336006 bytes.

Modules/ThirdParty/GDCM/src/gdcm/Source/Common/gdcmSwapper.txx
```diff
@@ -172,7 +176,9 @@ namespace gdcm
     }
   template <> inline float SwapperDoOp::Swap<float>(float val)
     {
-    return static_cast<float>(Swap((uint32_t)val));
+    uint32_t temp;
+    memcpy(&temp, &val, sizeof(uint32_t));
```
issakomi, line 180: Failed on Linux
```
2022-01-05T22:29:31.5821948Z /home/vsts/work/1/s/Modules/ThirdParty/GDCM/src/gdcm/Source/Common/gdcmSwapper.txx: In static member function 'static T gdcm::SwapperDoOp::Swap(T) [with T = float]':
2022-01-05T22:29:31.5823350Z /home/vsts/work/1/s/Modules/ThirdParty/GDCM/src/gdcm/Source/Common/gdcmSwapper.txx:180:5: error: 'memcpy' was not declared in this scope
2022-01-05T22:29:31.5824395Z      memcpy(&temp, &val, sizeof(uint32_t));
2022-01-05T22:29:31.5850093Z      ^~~~~~
```

seanm: Yes, I noticed. I've already submitted a patch upstream.

issakomi: I am afraid that the patch will not work on Linux
```cpp
#if defined(_MSC_VER)

#include <cstring>
```
should be outside `#if defined(_MSC_VER)`

seanm: Fixed [here](https://github.com/malaterre/GDCM/pull/134)

Modules/ThirdParty/GDCM/src/gdcm/Source/Common/gdcmSwapper.txx
```diff
@@ -106,7 +106,9 @@ namespace gdcm
     }
   template <> inline float SwapperNoOp::Swap<float>(float val)
     {
-    return Swap((uint32_t)val);
+    uint32_t temp;
+    memcpy(&temp, &val, sizeof(uint32_t));
+    return static_cast<float>(Swap(temp));
```
issakomi, line 111: IMHO, not yet sure about it, **maybe i am wrong**, is this cast to float from uint32 OK?
Is it not required to copy bits back to float instead?
```cpp
   uint32_t temp;
   memcpy(&temp, &val, 4);
   temp = Swap(temp);
   memcpy(&val, &temp, 4);
   return val;
```

CC @N-Dekker, could you kindly take a look?

N-Dekker: @issakomi Thanks for asking me, but honestly I don't know what `SwapperNoOp::Swap<float>` is supposed to do. Does `NoOp` mean "do nothing", as in https://en.wikipedia.org/wiki/NOP_(code) ?

Quite recently (honestly) I became aware that a `union` should never be used to convert one type to another. It would cause _undefined behavior_. If such a bit-wise conversion is really necessary (as in this case), it can be done more properly by `std::memcpy`. (Or by `std::bit_cast`, if you're using C++20 : https://en.cppreference.com/w/cpp/numeric/bit_cast ) 

My two cents!

issakomi: > Does NoOp mean "do nothing"

Looks like it depends on _GDCM_WORDS_BIGENDIAN_

```cpp
#ifdef GDCM_WORDS_BIGENDIAN
class SwapperDoOp
{
public:
  template <typename T> static T Swap(T val) {return val;}
  template <typename T> static void SwapArray(T *, unsigned int ) {}
};
...
#else
class SwapperNoOp
{
public:
  template <typename T> static T Swap(T val) {return val;}
  template <typename T> static void SwapArray(T *, size_t ) {}
};
...
```

Yes, that solution with union is know as being _not always working_.

Thank you for the answer!
Maybe i look at it later again, actually i got a headache already. :)

N-Dekker: > actually i got a headache already. :)

Sorry to hear so @issakomi ! Stay away from the computer for a little while, and I'm sure you will soon get well again  :)

seanm: @issakomi thanks for your review... I'll take another look...

Leengit:
There's always the approach with reinterpret_cast
```c++
return *reinterpret_cast<float *>(&Swap(*reinterpret_cast<uint32_t *>(&val)));
```

Modules/ThirdParty/GDCM/src/gdcm/Source/Common/gdcmSwapper.txx
seanm: @issakomi you're right, my patch was wrong, I'll tag you in my fix to upstream GDCM...

seanm:
> There's always the approach with reinterpret_cast

Strictly speaking, that's UB too, which is why the memcpy dance is best.

Modules/ThirdParty/GDCM/src/gdcm/Source/Common/gdcmSwapper.txx
issakomi: Thank you. If fact i don't have a way to validate just now with some data set or GDCM example. As you find it right.

Leengit:
> > There's always the approach with reinterpret_cast
> 
> Strictly speaking, that's UB too, which is why the memcpy dance is best.

I'd like to understand this better ... can you recommend a web page that addresses this issue?

Modules/ThirdParty/GDCM/src/gdcm/Source/Common/gdcmSwapper.txx
seanm: BTW I've been playing with an old PowerPC G5 Mac running OpenBSD, which is how I've come across these issues...

seanm:
@Leengit see for example https://adriann.github.io/undefined_behavior.html

N-Dekker:
> ```c++
> return *reinterpret_cast<float *>(&Swap(*reinterpret_cast<uint32_t *>(&val)));
> ```

@Leengit  Definitely UB, but I guess it won't compile anyway, as it tries to take the address of a temporary 😸 I mean: the r-value returned by `Swap`! Or did you recently overload `operator&` to support taking the address of an r-value?  😉

Leengit:
As @issakomi  mentions, we need a second `memcpy`, yes?  As in:
```c++
template <>
inline float
SwapperNoOp::Swap<float>(float val)
{
  uint32_t temp;
  memcpy(&temp, &val, sizeof(uint32_t));
  temp = Swap(temp);
  memcpy(&val, &temp, sizeof(float));
  return val;
}
```

seanm:
> As @issakomi mentions, we need a second `memcpy`, yes?

https://github.com/malaterre/GDCM/pull/135

issakomi:
Just for completeness, looked what other libraries do, e.g. [boost 1.77.0](https://www.boost.org/doc/libs/1_77_0/libs/endian/doc/html/endian.html#conversion_reference)

> Other built-in types, such as bool, float, or unscoped enumerations, do not have the same property, which means that reversing their constituent bytes may produce an invalid value, leading to undefined behavior. These types are therefore disallowed in endian_reverse, but are still allowed in endian_reverse_inplace. Even if an object becomes invalid as a result of reversing its bytes, as long as its value is never read, there would be no undefined behavior.

And here in boost [FAQ](https://www.boost.org/doc/libs/1_77_0/libs/endian/doc/html/endian.html#overview_faq)

> Is there floating point support?
An attempt was made to support four-byte floats and eight-byte doubles, limited to IEEE 754 (also known as ISO/IEC/IEEE 60559) floating point and further limited to systems where floating point endianness does not differ from integer endianness. Even with those limitations, support for floating point types was not reliable and was removed. For example, simply reversing the endianness of a floating point number can result in a signaling-NAN.
Support for float and double has since been reinstated for endian_buffer, endian_arithmetic and the conversion functions that reverse endianness in place. The conversion functions that take and return by value still do not support floating point due to the above issues; reversing the bytes of a floating point number does not necessarily produce another valid floating point number.

STL does exactly the  [same](https://github.com/letoram/SDL2/blob/704d68008135015dedf90409f2f7e07bd895ce63/include/SDL_endian.h#L213) thing as in my example with _union_

Looks like there is probably no general good solution at all (IEEE 754 floating-point standard does not specify endianness), there are also very special implementations. If the conversion is required a programmer has to find a best solution that works in particular case.

Leengit:
A workable solution at the bottom that unfortunately would surely need changes to the code that calls it would be to not convert `float` to `float`.  Instead an _encoding_ `Swap` function (from actual floats to a byte array that is written to disk) would be `float` to `uint32_t` and _decoding_ `Swap` function (from a byte array back to floats) would be `uint32_t` to `float`.

seanm:
So this leaves figuring out the max sizes that @dzenanz and I were discussing...

seanm:
I guess I could just arbitrarily increase a few of them until CI is happy...

seanm:
Anyone know where `hooks.MaxObjectKiB` and `hooks-max-size` come from?  They don't seem to be a git thing.  Most search results are very kitware-specific.  It would help to have docs about how they work.

dzenanz:
Judging by this comment, `hooks-max-size` comes from ghostflow:
https://github.com/BRAINSia/BRAINSTools/blob/d2fd21fb2295c071593e2f0728f0a3ab0890a1b3/.gitattributes#L45-L46

dzenanz:
Blame leads to this commit: 3251408adaad66a33a5a6804d3e2752092faa25a by @thewtex. Maybe @bradking or @mathstuf can explain where `hooks-max-size` comes from?

mathstuf:
https://gitlab.kitware.com/utils/rust-ghostflow/-/blob/master/ghostflow-cli/doc/checks.md#check_size

seanm:
Thanks!  And this `hooks.MaxObjectKiB` thing? Anyone know where it's from?  Seems not to be from ghostflow, and most search results are very old...

N-Dekker:
FYI `hooks-max-size` has previously triggered error messages in a pull request that would extend "itkMacro.h", which was almost 100 KB already: https://github.com/InsightSoftwareConsortium/ITK/pull/2785#pullrequestreview-774973629  saying:

> commit https://github.com/InsightSoftwareConsortium/ITK/commit/8f26661f8463af1c0d533015670e72cfcd36860e creates blob 12398f35ac335d57dafd510e044b91bd28b4d29f at Modules/Core/Common/include/itkMacro.h with size 101090 bytes (98.72 KiB) which is greater than the maximum size 100000 bytes (97.66 KiB). If the file is intended to be committed, set the hooks-max-size attribute on its path.

Fortunately this could be fixed by removing a lot of white-space from itkMacro.h 😃 By clang-format `AlignEscapedNewlines`: commit 7829ef68adb9e45b92da69017235aa35c92f0c64

mathstuf:
`hooks.MaxObjectKiB` is from the `hooks` branch used for `pre-commit` checks. These can largely be removed these days IMO. It was renamed because the `KiB` meant anything wanting partial KiBs was dealing with floating point parsing, so just straight bytes was preferred.

seanm:
OK thanks.  So maybe this is what we need: https://github.com/malaterre/GDCM/pull/139/commits/ba4d34a0ebf488588f86e8ba83056aa16e473ce5

seanm:
OK that's now upstreamed, so maybe this will work: https://github.com/InsightSoftwareConsortium/ITK/pull/3215

PR #3049, opened 2022-01-06
BUG: Fix for incorrect number of cells in ConnectedRegionsMeshFilter output

PranjalSahu:
Fix for correct number of cells in the output mesh. A counter while inserting cell
is used. Earlier the same cell id as in original mesh is used and due to that while counting total number of cells it was
giving wrong result.

Also added a test in Python for testing number of connected components and count of cells in the
largest connected component. Serves the purpose of demonstrating the usage of ConnectedRegionsMeshFilter as well.

## PR Checklist

### DIFF
--- a/Modules/Core/Mesh/test/itkConnectedRegionsMeshFilterTest2.cxx
+++ b/Modules/Core/Mesh/test/itkConnectedRegionsMeshFilterTest2.cxx
@@ -1,0 +1,113 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+#include "itkImageFileReader.h"
+#include "itkConnectedRegionsMeshFilter.h"
+#include "itkTestingMacros.h"
+#include "itkBinaryThresholdImageFilter.h"
+#include "itkBinaryMask3DMeshSource.h"
+#include "itkMeshFileReader.h"
+#include <iostream>
+#include <string>
+
+int
+itkConnectedRegionsMeshFilterTest2(int argc, char * argv[])
+{
+
+  if (argc < 4)
+  {
+    std::cerr << "Usage: " << itkNameOfTestExecutableMacro(argv)
+              << " Input NumberOfConnectedComponents NumberOfCellsInLargestComponent" << std::endl;
+    return EXIT_FAILURE;
+  }
+
+  // Check if input file is a mesh file or image file by checking presence of .vtk
+  std::string fileName(argv[1]);
+  std::size_t found = fileName.find(".vtk");
+  bool        imageSource = true;
+
+  if (found != std::string::npos)
+  {
+    imageSource = false;
+  }
+
+  const unsigned int Dimension = 3;
+  using MeshType = itk::Mesh<float, Dimension>;
+  MeshType::Pointer mesh;
+
+  if (imageSource)
+  {
+    // Read 3D Image to create test mesh
+    using PixelType = unsigned char;
+    using ImageType = itk::Image<PixelType, Dimension>;
+
+    using ReaderType = itk::ImageFileReader<ImageType>;
+    auto reader = ReaderType::New();
+    reader->SetFileName(fileName);
+    ITK_TRY_EXPECT_NO_EXCEPTION(reader->Update());
+
+    auto inputImage = reader->GetOutput();
+
+    // Threshold the 3D image to get binary mask for mesh generation
+    using ThresholdFilterType = itk::BinaryThresholdImageFilter<ImageType, ImageType>;
+    auto thresholdFilter = ThresholdFilterType::New();
+    thresholdFilter->SetInput(inputImage);
+    thresholdFilter->SetLowerThreshold(0);
+    thresholdFilter->SetUpperThreshold(200);
+    thresholdFilter->SetOutsideValue(1);
+    thresholdFilter->SetInsideValue(0);
+    ITK_TRY_EXPECT_NO_EXCEPTION(thresholdFilter->Update());
+
+    auto outputImage = thresholdFilter->GetOutput();
+
+    // Get mesh from binary image
+    using MeshSourceType = itk::BinaryMask3DMeshSource<ImageType, MeshType>;
+
+    auto meshSource = MeshSourceType::New();
+    meshSource->SetInput(outputImage);
+    meshSource->SetObjectValue(1);
+    ITK_TRY_EXPECT_NO_EXCEPTION(meshSource->Update());
+    mesh = meshSource->GetOutput();
+  }
+  else
+  {
+    // Read the test mesh using MeshFileReader
+    using ReaderType = itk::MeshFileReader<MeshType>;
+    ReaderType::Pointer polyDataReader = ReaderType::New();
+    polyDataReader->SetFileName(fileName);
+    ITK_TRY_EXPECT_NO_EXCEPTION(polyDataReader->Update());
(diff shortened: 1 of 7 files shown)


dzenanz (approved):

jhlegarreta (commented):
Kudos to the effort to detect and investigate the bug @PranjalSahu. 

Although I like the Python test, I truly think it is more suited to the Sphinx Examples (with or without the checks). I guess this has not been discussed to date, but IMHO maintaining part of the testing in Python and another part in C++ makes things more difficult in many ways, and Python tests do not count towards the coverage, which is very unfortunate.

I would C&P the example to the Sphinx Examples repository, translate it into a C++ example as well, and add the C++ testing version here (providing the image and the expected values as inputs to the test).

Note that there is already a similar test:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Mesh/test/itkExtractMeshConnectedRegionsTest.cxx

The name is a little bit misleading since there is no use of any e.g. `MeshExtractionFilter` in that example. So I'd rename the existing test to `itkConnectedRegionsMeshFilterTest1.cxx` and name this one `itkConnectedRegionsMeshFilterTest2.cxx`.

I did not dare to mark this as `Request changes` because as said I guess Python vs. C++ testing has not been discussed, but the above reasoning makes most sense to me.

thewtex:
I think @PranjalSahu created the test in Python because that is his use case. (Nice job on creating the test!). @jhlegarreta I also see the value of adding C++ coverage. @jhlegarreta do you agree that if @PranjalSahu  also creates a C++ test as you suggested, we should then merge this?

jhlegarreta:
> @jhlegarreta do you agree that if @PranjalSahu also creates a C++ test as you suggested, we should then merge this?

Yes.

PranjalSahu:
I am adding the test in C++ also.

dzenanz (approved):

Modules/Core/Mesh/wrapping/test/itkConnectedRegionsMeshFilterTest.py
```diff
@@ -0,0 +1,56 @@
...
+# ==========================================================================*/
+import itk
+import sys
+
+filename = sys.argv[1]
+
+inputImage = itk.imread(filename, itk.UC)
+
+# Threshold the 3D image to get binary mask for mesh generation
+thresholdFilter = itk.BinaryThresholdImageFilter.IUC3.New()
```
dzenanz, line 26: Preferred Python naming uses `snake_case`, e.g. `threshold_filter`.

Modules/Core/Mesh/wrapping/test/itkConnectedRegionsMeshFilterTest.py
```diff
@@ -0,0 +1,56 @@
...
+thresholdFilter.SetInput(inputImage)
+thresholdFilter.SetLowerThreshold(0)
+thresholdFilter.SetUpperThreshold(200)
+thresholdFilter.SetOutsideValue(1)
+thresholdFilter.SetInsideValue(0)
+thresholdFilter.Update()
+
+outputImage = thresholdFilter.GetOutput()
```
dzenanz, line 34: outputImage -> output_image etc.

Modules/Core/Mesh/test/itkConnectedRegionsMeshFilterTest2.cxx
```diff
@@ -0,0 +1,82 @@
...
+#include "itkBinaryMask3DMeshSource.h"
+#include <iostream>
+
+/*
+ * Test the mesh connectivity class.
+ */
+int
+itkConnectedRegionsMeshFilterTest2(int, char * argv[])
+{
+
```
jhlegarreta, line 31: Please use the input argument check block, as recommended by the ITK SW Guide.

Modules/Core/Mesh/test/itkConnectedRegionsMeshFilterTest2.cxx
```diff
@@ -0,0 +1,82 @@
...
+  // Read 3D Image to create Test Mesh
+  using PixelType = unsigned char;
+  const unsigned int Dimension = 3;
```
jhlegarreta, line 34: Although a minor style detail, the dimension definition is usually immediately after the input argument check block, i.e. before any other type alias definition, etc.

Modules/Core/Mesh/test/itkConnectedRegionsMeshFilterTest2.cxx
```diff
@@ -0,0 +1,82 @@
...
+  auto mesh = meshSource->GetOutput();
+
+  // Check number of connected components in the mesh
+  using ConnectFilterType = itk::ConnectedRegionsMeshFilter<MeshType, MeshType>;
+  auto connectivityFilter = ConnectFilterType::New();
+  connectivityFilter->SetInput(mesh);
+  connectivityFilter->SetExtractionModeToAllRegions();
+  connectivityFilter->Update();
+
+  ITK_TEST_EXPECT_TRUE(connectivityFilter->GetNumberOfExtractedRegions() == 15);
```
jhlegarreta, line 73: Eventually, for another input mesh (e.g. if the filter is found not to be working for that particular mesh), we'd like to be able to check this dynamically, according to the value of an input parameter that is provided together with the image.

Modules/Core/Mesh/test/itkConnectedRegionsMeshFilterTest2.cxx
```diff
@@ -0,0 +1,82 @@
...
+
+  // Check the number of cells in the largest connected component
+  connectivityFilter->SetExtractionModeToLargestRegion();
+  connectivityFilter->Update();
+
+  ITK_TEST_EXPECT_TRUE(connectivityFilter->GetOutput()->GetNumberOfCells() == 2154);
```
jhlegarreta, line 79: Same comment as before concerning the expected number of extracted regions.

Modules/Core/Mesh/test/itkConnectedRegionsMeshFilterTest2.cxx
```diff
@@ -0,0 +1,82 @@
...
+#include "itkImageFileReader.h"
+#include "itkConnectedRegionsMeshFilter.h"
+#include "itkTestingMacros.h"
+#include "itkBinaryThresholdImageFilter.h"
+#include "itkBinaryMask3DMeshSource.h"
+#include <iostream>
+
+/*
+ * Test the mesh connectivity class.
+ */
```
jhlegarreta, line 27: The ITK SW Guide considers that the purpose of a test is described by its name, so we usually avoid to add a block to comment on what filters the file tests.

Modules/Core/Mesh/test/itkConnectedRegionsMeshFilterTest2.cxx
```diff
@@ -0,0 +1,82 @@
...
+
+  // Read 3D Image to create Test Mesh
+  using PixelType = unsigned char;
+  const unsigned int Dimension = 3;
+  using ImageType = itk::Image<PixelType, Dimension>;
+
+  using ReaderType = itk::ImageFileReader<ImageType>;
+  auto reader = ReaderType::New();
+  reader->SetFileName(argv[1]);
+  reader->Update();
```
jhlegarreta, line 40: As per the ITK SW Guide, we call any `Update` method with a `ITK_TRY_EXPECT_NO_EXCEPTION` macro.

jhlegarreta (commented):
Thanks for the effort :100: @PranjalSahu. 

A few comments to improve the C++ test style, and especially, to potentially open it for testing with other meshes while keeping it quantitative.

Modules/Core/Mesh/test/itkConnectedRegionsMeshFilterTest2.cxx
PranjalSahu: In the current test implementation, I am generating a mesh from a 3D image file. 
So its future application to test directly on other mesh is not present. 

What I can do is to check if the input file is not a .vtk then continue with the current approach and if the input file is a .vtk then directly work on that as a mesh source. Also I can make the number of cells, and connected components as input parameters to the test.

jhlegarreta (approved):
Thanks for the effort @PranjalSahu !

Modules/Core/Mesh/test/itkConnectedRegionsMeshFilterTest2.cxx
jhlegarreta: Even in the case where the input was another image file, the previous implementation was not allowing to test it. Thanks for considering mesh files as well.

dzenanz (approved):

PR #3052, opened 2022-01-06
Better progress reporting

### LINKED ISSUE #2875: Some filters report progress both from theader and from themselves

dzenanz:
This is a follow-up to #3014. Closes #2875.

### DIFF
--- a/Modules/Core/Common/include/itkProcessObject.h
+++ b/Modules/Core/Common/include/itkProcessObject.h
@@ -983,6 +983,9 @@ private:
   /** Friends of ProcessObject */
   friend class DataObject;
 
+  friend class ProgressReporter;
+  friend class TotalProgressReporter;
+
   friend class DataObjectConstIterator;
   friend class InputDataObjectConstIterator;
   friend class OutputDataObjectConstIterator;
--- a/Modules/Core/Common/src/itkProgressReporter.cxx
+++ b/Modules/Core/Common/src/itkProgressReporter.cxx
@@ -16,6 +16,7 @@
  *
  *=========================================================================*/
 #include "itkProgressReporter.h"
+#include "itkMultiThreaderBase.h"
 
 namespace itk
 {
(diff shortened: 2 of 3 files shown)


thewtex (approved):
:railway_track: :racing_car:

blowekamp:
> Brad can you test this with SimpleITK?

@dzenanz I quickly rebuild and ran SimpleITK's test suit and it appears to be OK.

Issue #3048, opened 2022-01-06
FFT Python wrappings don't compile


dzenanz:
### Description

I get errors when building Python wrapping:
```text
74>------ Build started: Project: ITKFFTCastXML, Configuration: RelWithDebInfo x64 ------
74>Generating ../../itkVnlComplexToComplex1DFFTImageFilter.xml
74>C:/Dev/ITK-py/Wrapping/itkVnlComplexToComplex1DFFTImageFilter.cxx(26,18): error G026DE12C: too few template arguments for class template 'VnlComplexToComplex1DFFTImageFilter'
74>    typedef itk::VnlComplexToComplex1DFFTImageFilter< itk::Image< std::complex< float >,2 > > itkVnlComplexToComplex1DFFTImageFilterICF2;
74>                 ^
74>C:/Dev/ITK-git/Modules/Remote/Ultrasound/include\itkVnlComplexToComplex1DFFTImageFilter.h:35:27: note: template is declared here
74>class ITK_TEMPLATE_EXPORT VnlComplexToComplex1DFFTImageFilter
74>                          ^
74>C:/Dev/ITK-py/Wrapping/itkVnlComplexToComplex1DFFTImageFilter.cxx(27,18): error G026DE12C: too few template arguments for class template 'VnlComplexToComplex1DFFTImageFilter'
74>    typedef itk::VnlComplexToComplex1DFFTImageFilter< itk::Image< std::complex< float >,2 > >::Pointer itkVnlComplexToComplex1DFFTImageFilterICF2_Pointer;
74>                 ^
74>C:/Dev/ITK-git/Modules/Remote/Ultrasound/include\itkVnlComplexToComplex1DFFTImageFilter.h:35:27: note: template is declared here
74>class ITK_TEMPLATE_EXPORT VnlComplexToComplex1DFFTImageFilter
74>                          ^
74>C:/Dev/ITK-py/Wrapping/itkVnlComplexToComplex1DFFTImageFilter.cxx(29,18): error G026DE12C: too few template arguments for class template 'VnlComplexToComplex1DFFTImageFilter'
74>    typedef itk::VnlComplexToComplex1DFFTImageFilter< itk::Image< std::complex< float >,3 > > itkVnlComplexToComplex1DFFTImageFilterICF3;
74>                 ^
[...]

Issue #2884, opened 2021-11-19
TriangleMeshToBinaryImageFilter not giving correct result after padding image and cropping output


nicktasios:
### Description

Due to a [bug](https://github.com/InsightSoftwareConsortium/ITK/issues/2416) with `TriangleMeshToBinaryImageFilter` where it can get corrupt near the edges, I tried padding the image and then cropping it by 1 pixel border. Unfortunately, that shifts the final results as you can see here:
<img width="410" alt="image" src="https://user-images.githubusercontent.com/91547393/142653607-b15aa58e-bda1-48a3-8c6f-14b59e998296.png">
 Green is with padding/cropping, and white is without.

### Steps to Reproduce

Run the following code
```python
t_mesh_filter.py                                                                        
import itk                                                                              
                                                                                        
if __name__ == '__main__':                                                              
    should_pad_and_crop = True                                                          
                                                                                        
    TCoordinate = itk.D                                                                 
    Dimension = 3                                                                       
    TMesh = itk.Mesh[TCoordinate, Dimension].New()                                      
    sphere = itk.RegularSphereMeshSource[TMesh].New()                                   
    sphere.SetResolution(4)                                                             
    sphere_mesh = sphere.GetOutput()                                                    
                                                                                        
    mesh_writer = itk.MeshFileWriter[TMesh].New()                                       
    mesh_writer.SetFileName( "sphere.vtk" )                                             
    mesh_writer.SetInput(sphere_mesh)                                                   
    mesh_writer.Update()                                                                
                                                                                        
    TMesh = itk.Mesh[itk.SS, Dimension].New()                                           
    mesh_reader = itk.MeshFileReader[TMesh].New()                                       
    mesh_reader.SetFileName("sphere.vtk")                                               
    mesh_reader.Update()                                                                
    sphere_mesh = mesh_reader.GetOutput()                                               
                                                                                        
    TPixel = itk.SS                                                                     
    TImage = itk.Image[TPixel, Dimension]                                               
                                                                                        
    image = itk.Image[TPixel, Dimension].New()                                          
    region = itk.ImageRegion[Dimension]()                                               
    region.SetSize([128, 128, 128])                                                     
    region.SetIndex([0, 0, 0])                                                          
    image.SetRegions(region)                                                            
    image.Allocate()                                                                    
    image.SetOrigin([-1.28, -1.28, -1.28])                                              
    image.SetSpacing([0.02, 0.02, 0.02])                                                
[...]

PR #3053, opened 2022-01-06
BUG: Condition FFT wrappings on complex types

tbirdso:
Addresses discussion in #3048 and #2884 surrounding scenario where a real type, i.e. `double`, may be enabled for ITK Python wrappings but its corresponding complex type, i.e. `std::complex<double>`, is not enabled. Adds checks to three FFT class templates to ensure that the classes are not wrapped if either real or complex pixel types are not available.

## PR Checklist


tbirdso:
FYI @dzenanz

dzenanz (commented):
Instead of this, we could make `ITK_WRAP_complex_float` a dependent option of `ITK_WRAP_float` in CMake. Analogously for double. Wrapping for complex does not make much sense without the base type. That would prevent needing this here, and in other places. A cleanup of `.wrap` files could be done then.

dzenanz:
Here is an example of how to make a dependent option:
https://github.com/InsightSoftwareConsortium/ITK/blob/v5.3rc02/CMakeLists.txt#L328-L330

tbirdso:
@dzenanz If I understand correctly, this fix addresses the opposite problem from what you're suggesting -- rather than worrying about the scenario where `ITK_WRAP_COMPLEX_DOUBLE=ON` but `ITK_WRAP_DOUBLE=OFF`, this is intended to address the failure where `ITK_WRAP_COMPLEX_DOUBLE=OFF` and `ITK_WRAP_DOUBLE=ON`. I think the latter case, where a user wants double wrappings but no `std::complex` wrappings, is an uncommon but reasonable scenario.

Making complex types dependent on real types is valid but I'm not sure it is required or warranted by this issue.

dzenanz (approved):

thewtex (approved):

thewtex:
:+1:

PR #3062, opened 2022-01-08
BUG: Fix superclass name in `itkTypeMacro`

jhlegarreta:
- BUG: Fix superclass name in `itkTypeMacro` 
- BUG: Use the superclass name in `itkTypeMacro` 

## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [X] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)

PR #3066, opened 2022-01-08
BUG: Do not print non-self member variables in `PrintSelf`

jhlegarreta:
Do not print non-self member variables in `PrintSelf`.

Fixes
```
  CurrentPosition: ITK test driver caught an ITK exception:

itk::ExceptionObject (0x55be11155e70)
Location: "unknown"
File: /home/vsts/work/1/s/Modules/Numerics/Optimizersv4/src/itkObjectToObjectOptimizerBase.cxx
Line: 201
Description: ITK ERROR: LBFGSBOptimizerv4(0x55be11150c00): m_Metric has not been assigned. Cannot get parameters.
```

raised in:
https://open.cdash.org/test/583682381

## PR Checklist

### DIFF
--- a/Modules/Numerics/Optimizersv4/src/itkLBFGSBOptimizerv4.cxx
+++ b/Modules/Numerics/Optimizersv4/src/itkLBFGSBOptimizerv4.cxx
@@ -71,7 +71,6 @@ LBFGSBOptimizerv4 ::PrintSelf(std::ostream & os, Indent indent) const
   Superclass::PrintSelf(os, indent);
 
   os << indent << "InitialPosition: " << m_InitialPosition << std::endl;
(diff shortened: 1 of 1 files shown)


jhlegarreta:
#3062 passing relies on this fix.

Also, note that this is the minimum set of changes required to fix the issue, but the `itk::LBFGSBOptimizerv4::PrintSelf` method is printing many non-self members. Probably other optimization classes are as well, and the issue might involve the related metrics or transforms (e.g. there is no `m_CurrentPosition` in any superclass or related metrics, except for a few specialized optimizers) . Might have been an effect of the non-v4 classes. It should be should be addressed in a separate PR.

dzenanz (approved):

PR #3071, opened 2022-01-09
BUG: Synchronize the object factory across Python modules

### LINKED ISSUE #2909: Use global, cross-binary singleton for itk::ObjectFactoryBase
Similar to the global, cross-binary singleton for the global timestamp, use a pointer for the object factory that is synchronized across Python module binaries on initialization.

After implementation, attempt a revert of https://github.com/InsightSoftwareConsortium/ITK/pull/2864/commits/35df104b3eaf2fbd7c765d1e786db804bdf5fbbd

tbirdso:
## Overview

Addresses #2909 to make the object factory behave as a singleton across Python modules.

## Background

The ITK object factory is used to provide alternate class implementations at runtime, such as adding support for new image/mesh/transform formats in `ImageRead()` and `ImageWrite()` or (more recently) swapping out backends for accelerated FFT computations. In theory "the object factory" should be treated as a global singleton within ITK, allowing any module or library to define class overrides that are then picked up at program execution by any other filter or class within ITK. In implementation "object factories" are defined and registered to a common `ObjectFactoryBase` object, which can then be accessed with `ObjectFactory<T>::Create()` to yield a subclass instance upcast to type `T`.

ITK Python wrappings introduce extra complexity into factory management where ITK is broken into individual modules such as ITKCommon, ITKFFT, ITKConvolution, etc. that are loaded on a lazy, as-needed basis. Along with this, Python modules maintain their own distinct versions of singleton objects with some behavioral differences across platforms, as we touched on with discussion in #2864. In practice both Windows and Linux were observed to initialize and maintain distinct, unsynchronized factory lists in their own individual modules.

This issue was previously confronted in [6770250](https://github.com/InsightSoftwareConsortium/ITK/commit/67702504e840a60bc968e2e09804879d9c6c5801), which added code for synchronizing global timestamps through `PyCapsule` objects in the Python C API, and then in [a66337e](https://github.com/InsightSoftwareConsortium/ITK/commit/a66337ec215c88f4900a2caf419b055483c42085), which extended synchronization so that all modules would point to a common map of singletons in the ITKCommon module (rather than a growing list of individual singleton references to synchronize). Unfortunately issues remained with object factory synchronization:
- Singleton synchronization executes after factory initialization, so each module initializes its own factory list before the singleton factory list is available;
- Object factory init caches a reference to the factory list, so each module continues to refer to its own individual list even though the common list is available;
- At exit each module attempts to destroy its referenced factory lists, so simply removing caching leads to memory management errors and segfaults.

## Changes

This modules introduces changes to appropriately synchronize object factory references across Python modules to behave as a singleton:
- `ObjectFactoryBase::Initialize()` is exposed and called immediately after `SingletonIndex` is synchronized. New functionality checks the cache against the global instance in ITKCommon and, if different, synchronizes the cache with the global instance before destroying the cache.
- The underlying factory list structure is made to inherit from `LightObject` to handle shared ownership among modules. In C++ projects the structure will have exactly one reference as a global singleton, while in Python the structure will have as many references as the number of modules with a synchronized object factory.
- Reverts [35df104](https://github.com/InsightSoftwareConsortium/ITK/commit/35df104b3eaf2fbd7c765d1e786db804bdf5fbbd) so that FFT factories are registered exactly once with C++ magic statics.

## Previous behavior

Object factory lists differ across modules. This can be verified by force loading all modules and then checking the list of factories in the ITKCommon module:
```py
> import itk
> itk.force_load()
> len(itk.ObjectFactoryBase.GetRegisteredFactories())
0
```

Unfortunately I'm not aware of a way to examine object factories in other modules from the console, as `itk.ObjectFactoryBase` always points to the object factory in ITKCommon. I've added more discussion in #2909 on debugging steps that can be used to trace through factories at load time.

## New behavior

Object factory lists are synchronized across modules.

```py
> import itk
> itk.force_load()
> len(itk.ObjectFactoryBase.GetRegisteredFactories())
38
```

`pdb` and `gdb` can be used to trace through and verify that each module references the same `SingletonIndex` and `ObjectFactoryBasePrivate` structures after load (see #2909).

## Other notes

- Closes #2909 .

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkObjectFactoryBase.h
+++ b/Modules/Core/Common/include/itkObjectFactoryBase.h
@@ -57,7 +57,7 @@ namespace itk
 
 // Forward reference because of private implementation
 class OverRideMap;
-struct ObjectFactoryBasePrivate;
+class ObjectFactoryBasePrivate;
 
 class ITKCommon_EXPORT ObjectFactoryBase : public Object
 {
@@ -237,6 +237,10 @@ public:
     (void)staticFactoryRegistration;
   }
 
+  /** Initialize the static members of ObjectFactoryBase.
+   *  RegisterInternal() and InitializeFactoryList() are called here. */
+  static void
+  Initialize();
 
 protected:
   void
@@ -277,11 +281,6 @@ private:
   /** Initialize the static list of Factories. */
   static void
   InitializeFactoryList();
-
-  /** Initialize the static members of ObjectFactoryBase.
-   *  RegisterInternal() and InitializeFactoryList() are called here. */
-  static void
-  Initialize();
 
   /** Register default factories which are not loaded at run time. */
   static void
--- a/Modules/Core/Common/src/itkObjectFactoryBase.cxx
+++ b/Modules/Core/Common/src/itkObjectFactoryBase.cxx
@@ -31,6 +31,7 @@
 #  include "itkDynamicLoader.h"
 #endif
 #include "itkDirectory.h"
+#include "itkLightObject.h"
 #include "itkSingleton.h"
 #include "itkVersion.h"
 #include <cstring>
@@ -60,7 +61,7 @@ SynchronizeList(FactoryListType * output, FactoryListType * input, bool internal
       int curr = 0;
       for (auto & i : *output)
       {
-        if (i->GetNameOfClass() == factory->GetNameOfClass())
+        if (typeid(*i) == typeid(*factory))
         {
           pos = curr;
           break; // factory already in internal factories.
@@ -99,8 +100,9 @@ namespace itk
  *
  */
 
-struct ObjectFactoryBasePrivate
+class ObjectFactoryBasePrivate : public LightObject
 {
+public:
   ~ObjectFactoryBasePrivate()
   {
     itk::ObjectFactoryBase::UnRegisterAllFactories();
@@ -126,11 +128,12 @@ struct ObjectFactoryBasePrivate
 ObjectFactoryBasePrivate *
 ObjectFactoryBase::GetPimplGlobalsPointer()
 {
-  if (m_PimplGlobals == nullptr)
+  static auto                deleteLambda = []() { m_PimplGlobals->UnRegister(); };
+  ObjectFactoryBasePrivate * globalInstance =
+    Singleton<ObjectFactoryBasePrivate>("ObjectFactoryBase", SynchronizeObjectFactoryBase, deleteLambda);
+  if (globalInstance != m_PimplGlobals)
   {
-    static auto deleteLambda = []() { delete m_PimplGlobals; };
-    m_PimplGlobals =
-      Singleton<ObjectFactoryBasePrivate>("ObjectFactoryBase", SynchronizeObjectFactoryBase, deleteLambda);
+    SynchronizeObjectFactoryBase(globalInstance);
   }
   return m_PimplGlobals;
 }
@@ -867,15 +870,20 @@ ObjectFactoryBase::SynchronizeObjectFactoryBase(void * objectFactoryBasePrivate)
   // We keep track of the previoulsy registered factory in `previousObjectFactoryBasePrivate`
   // but assign the new pointer to `m_PimplGlobals` so factories can be
   // registered directly with the new pointer.
-  ObjectFactoryBasePrivate * previousObjectFactoryBasePrivate;
-  previousObjectFactoryBasePrivate = GetPimplGlobalsPointer();
-
+  ObjectFactoryBasePrivate * previousObjectFactoryBasePrivate = m_PimplGlobals;
   m_PimplGlobals = static_cast<ObjectFactoryBasePrivate *>(objectFactoryBasePrivate);
+
   if (m_PimplGlobals && previousObjectFactoryBasePrivate)
   {
     SynchronizeList(m_PimplGlobals->m_InternalFactories, previousObjectFactoryBasePrivate->m_InternalFactories, true);
     SynchronizeList(
       m_PimplGlobals->m_RegisteredFactories, previousObjectFactoryBasePrivate->m_RegisteredFactories, false);
+  }
+
+  if (m_PimplGlobals && previousObjectFactoryBasePrivate && previousObjectFactoryBasePrivate != m_PimplGlobals)
+  {
+    m_PimplGlobals->Register();
+    previousObjectFactoryBasePrivate->UnRegister();
   }
 }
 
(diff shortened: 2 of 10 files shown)


Modules/Core/Common/include/itkObjectFactoryBase.h
```diff
@@ -57,7 +57,7 @@ namespace itk
 
 // Forward reference because of private implementation
 class OverRideMap;
-struct ObjectFactoryBasePrivate;
+class ObjectFactoryBasePrivate;
```
dzenanz, line 60: This commit is surprisingly small.

dzenanz (approved):
Code looks good.

Modules/Core/Common/include/itkObjectFactoryBase.h
tbirdso: Yes, tracing and debugging took much longer than writing 😬

Modules/Core/Common/src/itkObjectFactoryBase.cxx
```diff
@@ -60,7 +61,7 @@ SynchronizeList(FactoryListType * output, FactoryListType * input, bool internal
       int curr = 0;
       for (auto & i : *output)
       {
-        if (i->GetNameOfClass() == factory->GetNameOfClass())
+        if (typeid(*i) == typeid(*factory))
```
thewtex, line 64: Is this change necessary? If not, can it be removed? There may be issues in using typeid, see e.g. 

https://stackoverflow.com/questions/53467813/comparing-two-type-info-from-typeid-operator

Modules/Filtering/FFT/wrapping/test/FFTObjectFactoryTest.py
```diff
@@ -0,0 +1,50 @@
...
+except ImportError as e:
+  pass
+
+# Verify ITKCommon factory list is not empty after loading all modules
+assert len(itk.ObjectFactoryBase.GetRegisteredFactories()) > 0
+
+# Verify that FFT factories are in the list
+overrides_found = list()
+for factory in itk.ObjectFactoryBase.GetRegisteredFactories():
+  for override_name in factory.GetClassOverrideNames():
```
thewtex, line 45: Can the Python `black` executable be run on the file to get the standard style?

thewtex (changes requested):
Way to go @tbirdso !! :1st_place_medal: :champagne: :taco: 

A few items noticed inline.

Modules/Core/Common/src/itkObjectFactoryBase.cxx
tbirdso: This change (or something similar) is necessary for synchronizing templated factories that have the same name but override different classes; for example, `itk::FFTImageFilterFactory` is templated over an incomplete FFT subclass type to minimize the code maintenance burden.

By comparing only the results of `factory->GetNameOfClass()` we get the following behavior:
1. Python session initialized and `itk.force_load()` is called
2. ITKCommon is initialized with 0 factories in its internal factory list
3. ITKFFT is initialized with 8 Vnl FFT override factories in its internal factory list, all of type `FFTImageFilterFactory`
4. Python C API `SingletonIndex` synchronization occurs
5. ITKCommon's factory list is synchronized with ITKFFT's factory list. Since all 8 factories return the same value for `factory->GetNameOfClass()`, 7 factories are discarded and only 1 factory is retained in ITKCommon
6. ITKFFT's factory list is discarded

To avoid discarding factories during factory sync we need a comparison operation that can capture template specializations.

I looked into [`std::is_same`](https://en.cppreference.com/w/cpp/types/is_same), but my understanding is that determines comparison values at _compile time_ whereas we are looking for _runtime_ comparison in order to properly synchronize Python modules. Comparison of [`type_info`](https://en.cppreference.com/w/cpp/types/type_info) objects  (what is returned from `typeid()`) is more appropriate here.

Maybe I'm misunderstanding, but it looks like discussion in the Stack Overflow [thread](https://stackoverflow.com/questions/53467813/comparing-two-type-info-from-typeid-operator) you referenced arrives at the conclusion that some implementation details can vary but direct type comparison among two types via `typeid` will still succeed. The second response lays out these details of GCC's `typeid` implementation:
1. Compare if the two `type_info` are at the same address -- which could be an issue in Python binaries which may have different addresses for the same types, I'm not sure. But then it falls back to:
2. Compare the two class names (which includes template specialization) -- in practice this should always succeed (at least as long as the binaries were compiled with the same compiler). I have observed differences in how specialized FFT factory names are generated across MSVC and GCC compilers, but this won't affect comparison outcomes on a given platform.

Lastly, I would argue that even if there is a scenario where `typeid` fails it is still a preferable failure condition to `factory->GetNameOfClass()`.
- When comparing `factory->GetNameOfClass()` we tend towards false positives, which results in factories being prematurely discarded. This would lead to very confusing Python behavior where classes suddenly can't be instantiated and filters break.
- When comparing `typeid` we could tend towards false negatives, which results in duplicate factories being added to the list. In Python this could lead to a bloated factory list (i.e., maybe 100 instead of 10 factory objects) and potentially leave factories in an incorrect order so that if two factories override the same class, the wrong one is used to instantiate that class. However, this would be a much rarer case.

@thewtex Sorry for the long explanation, hope this helps make sense of why `typeid` is needed here. If there is a better structure to use for comparing type information at runtime, or if there are other issues with `typeid` I'm missing, I can look for alternatives. But, I think this is the best path forward for now.

dzenanz (approved):

Modules/Core/Common/src/itkObjectFactoryBase.cxx
thewtex: @tbirdso thanks for the detailed explanation! That makes sense.

`typeid` is brittle, so it has to be used with care. E.g. on macOS, the Run-Time Type Information (RTTI) has caused many frustrating, hard-to-debug, failures in `dynamic_cast`'s. It turns out that, in practice, with certain build settings, the way libraries are loaded on macOS can cause divergent RTTI without explicit attributes on the symbols across the binaries. For this reason, all the FFT classes must have the `ITK_TEMPLATE_EXPORT` macro, or factory registration / identification may fail on macOS. We may also run into issues when a project mixes compilers, compiler versions, or C++ standard versions.

However, in the absence of a better approach, let's use this one.

thewtex (approved):

Modules/Core/Common/src/itkObjectFactoryBase.cxx
tbirdso: Thanks for the explanation as well @thewtex 

>  It turns out that, in practice, with certain build settings, the way libraries are loaded on macOS can cause divergent RTTI without explicit attributes on the symbols across the binaries. For this reason, all the FFT classes must have the ITK_TEMPLATE_EXPORT macro, or factory registration / identification may fail on macOS.

I wasn't aware of this and will certainly keep this in mind for the future

> We may also run into issues when a project mixes compilers, compiler versions, or C++ standard versions.

This is one instance that came to mind for the false negative case above

> However, in the absence of a better approach, let's use this one.

👍

Issue #3057, opened 2022-01-07
`PythonFFTImageFilterTest` failing in Windows CI


tbirdso:
### Description

Windows CI is failing after #3002 due to issue in FFT test

[CDash results](https://open.cdash.org/test/582543193)

### Steps to Reproduce

[Windows CI on Azure](https://dev.azure.com/itkrobotwindowpython/ITK.Windows.Python/_build/results?buildId=2786&view=results)

### Expected behavior

`ComplexToRealImageFilter` is available and all CI passes

### Actual behavior

```
Running itkImageFileWriterICF2... done
Traceback (most recent call last):
  File "D:/a/1/s/Modules/Filtering/FFT/wrapping/test/FFTImageFilterTest.py", line 45, in <module>
    realFilter = itk.ComplexToRealImageFilter[ComplexImageType, ImageType].New(fftFilter)
  File "D:\a\1\s-build\Wrapping\Generators\Python\itk\support\lazy.py", line 130, in __getattribute__
    value = types.ModuleType.__getattribute__(self, attr)
AttributeError: module 'itk' has no attribute 'ComplexToRealImageFilter'
itkTestDriver: Process exited with return value: 1
```

### Reproducibility

Appears in `master` branch throughout several merges, but did not appear in CI checks for #3002 prior to merge and does not seem to appear locally

### Versions

`main` branch

### Environment

Windows CI

### Additional Information

#3002 which seems to have introduced the issue dealt with C++ TestDriver changes but not wrappings, so I'm a little confused as to how this error appeared from there. Will investigate.

tbirdso:
Haven't been able to recreate on my Windows setup after rebuilding against `master`. ` PythonFFTImageFilterTest` passes, I can instantiate `itk.ComplexToRealImageFilter` just fine, and [`itkComplexToRealImageFilter.wrap`](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/ImageIntensity/wrapping/itkComplexToRealImageFilter.wrap) has not been touched in years.

It would be helpful to have another set of eyes on this, maybe I'm missing something obvious. @dzenanz if you have a moment could you weigh in on this issue?

dzenanz:
Let's wait for nightly builds and see whether my computer also chokes on it. If so, I could try to fix it.

Oops, we don't have a nightly Windows Python build. Rebasing would probably require a massive rebuild. I could do that overnight.

tbirdso:
[...]

PR #3074, opened 2022-01-10
COMP: Enable filtering group for Windows Python CI

tbirdso:
Explicitly adds ITKImageIntensity as a test dependency of ITKFFT to resolve Windows Python failure where `itk.ComplexToRealImageFilter` is not found after building ITKCommon (see #3057).

## PR Checklist

### DIFF
--- a/Modules/Filtering/FFT/itk-module.cmake
+++ b/Modules/Filtering/FFT/itk-module.cmake
@@ -20,6 +20,7 @@ itk_module(ITKFFT
   TEST_DEPENDS
     ITKTestKernel
     ITKImageCompare
+    ITKImageIntensity
   FACTORY_NAMES
     ${_fft_backends}
   DESCRIPTION
--- a/Testing/ContinuousIntegration/AzurePipelinesWindowsPython.yml
+++ b/Testing/ContinuousIntegration/AzurePipelinesWindowsPython.yml
@@ -57,6 +57,7 @@ jobs:
           ITK_WRAP_PYTHON:BOOL=ON
           ITK_BUILD_DEFAULT_MODULES:BOOL=OFF
           ITKGroup_Core:BOOL=ON
+          Module_ITKImageIntensity:BOOL=ON
           Module_ITKBridgeNumPy:BOOL=ON
           Python3_FIND_REGISTRY=LAST
         ")


Testing/ContinuousIntegration/AzurePipelinesWindowsPython.yml
```diff
@@ -57,6 +57,7 @@ jobs:
           ITK_WRAP_PYTHON:BOOL=ON
           ITK_BUILD_DEFAULT_MODULES:BOOL=OFF
           ITKGroup_Core:BOOL=ON
+          ITKGroup_Filtering:BOOL=ON
```
dzenanz, line 60: You don't have to enable the whole group. You can do `Module_ITKFFT` only, if that is the point.

tbirdso: Fixed!

dzenanz (approved):

tbirdso:
@dzenanz I just noticed that Windows Python is not in the list of CI -- should we be running that for each push?

dzenanz:
We recently removed it from the list, because it was going over the Azure's time limit (6h). This test failure was a rare instance where that was a problem.

tbirdso:
> We recently removed it from the list, because it was going over the Azure's time limit (6h). This test failure was a rare instance where that was a problem.

It looks like it is still running for updates to `master` though. Was it only running over the time limit when run against unmerged PRs?

Would it be reasonable to try re-enable it here to assist in testing a possible fix, then disable it again before marking the PR as ready? I'm having trouble finding the list of pipelines to run, is that set somewhere in `.github` or `Testing`?

tbirdso:
I'm still working on rebuilding on my side, having issues with `ITKPyBasePython` not being found for import despite its `.pyd` binary being located at the correct path. I think this is an issue with a CMake setting on my side and not directly related to the FFT problem tracked here.

```py
>>> import itk
>>> itk.force_load()
  File "<stdin>", line 1, in <module>
  File "C:\repos\ITK-build-unstable\Wrapping\Generators\Python\itk\support\init_helpers.py", line 116, in force_load
    getattr(itk, k)
  File "C:\repos\ITK-build-unstable\Wrapping\Generators\Python\itk\support\lazy.py", line 137, in __getattribute__
    base.itk_load_swig_module(module, namespace)
  File "C:\repos\ITK-build-unstable\Wrapping\Generators\Python\itk\support\base.py", line 102, in itk_load_swig_module
    itk_load_swig_module(dep, namespace)
  File "C:\repos\ITK-build-unstable\Wrapping\Generators\Python\itk\support\base.py", line 110, in itk_load_swig_module
    l_module = loader.load(swig_module_name)
  File "C:\repos\ITK-build-unstable\Wrapping\Generators\Python\itk\support\base.py", line 259, in load
    l_spec.loader.exec_module(l_module)  # pytype: disable=attribute-error
  File "<frozen importlib._bootstrap_external>", line 783, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "C:\repos\ITK-build-unstable\Wrapping\Generators\Python\itk\support\..\ITKPyBasePython.py", line 13, in <module>
    from . import _ITKPyBasePython
ImportError: DLL load failed while importing _ITKPyBasePython: The specified module could not be found.
```

dzenanz:
It could be that it was exceeding the time limit only occasionally, but that was enough trouble to disable it. I think it needs to be enabled somewhere in Azure configuration, not by a change in this repository.

dzenanz:
The tests for `merge to master` might be doing an incremental build, so that's why they were not running over the time limit. I think it might be easier to merge this PR to see whether it resolves the problem, then to search through Azure configurations and fiddle with that.

tbirdso:
Even if the change here is small, testing by merging to `master` seems like it could get really messy, especially if the problem is not resolved and we need to try more small changes.

Let me try a little longer to build locally with the test's CMake configuration, if I haven't made progress we can revisit tomorrow?

tbirdso:
Reading back through [a failed run log](https://dev.azure.com/itkrobotwindowpython/a2cc6bea-a502-416d-adf3-3aa4d21ad7ff/_apis/build/builds/2799/logs/16), it actually looks like ITKFFT is already enabled as a dependency, so this change will have no effect.

```
2022-01-10T15:57:22.5006457Z -- Enabled ITKFFT, needed by [ITKTestKernel].
```

Back to the drawing board...

dzenanz (approved):

Modules/Filtering/FFT/itk-module.cmake
```diff
@@ -20,6 +20,7 @@ itk_module(ITKFFT
   TEST_DEPENDS
     ITKTestKernel
     ITKImageCompare
+    ITKImageIntensity
```
dzenanz, line 23: Nitpick: this commit should have a `COMP: ` prefix, as it is a compile problem and not an actual code bug.

tbirdso:
I believe I've identified the issue as a cascading dependency failure, where ITKImageIntensity was implicitly required by the Python FFT test but not built accordingly. I am still resolving an issue with building ITK Python modules on my local Windows machine and cannot yet verify that this resolves the Windows CI issue. However, if all other CI passes then these minor changes are not expected to break anything and the best path forward may be to go ahead and merge, then check incremental Windows CI from `master`.

Modules/Filtering/FFT/itk-module.cmake
tbirdso: Good catch, fixed.

dzenanz:
Your analysis in the issue description instills confidence in this solution.

thewtex (approved):

PR #2832, opened 2021-10-21
Hdf5 1 12 1

seanm:
## PR Checklist
- [x] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [x] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [ ] Added test (or behavior not changed)
- [ ] Updated API documentation (or API not changed)
- [ ] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [ ] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [ ] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples) for all new major features (if any)

Refer to the [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) for
further development details if necessary.

PR #3075, opened 2022-01-10
ENH: Update eigen3.4

### LINKED ISSUE #3045: Update eigen to 3.4
From the [look of the branch name](https://github.com/InsightSoftwareConsortium/ITK/blob/51dc8fee7da4e1a52d0190327316da512aac84f0/Modules/ThirdParty/Eigen3/UpdateFromUpstream.sh#L7), ITK master currently uses eigen 3.3.x.

3.4 was [released](https://eigen.tuxfamily.org/index.php?title=3.4) a few months ago.

In order to eventually fix all the [-Wreserved-id-macro warnings](https://open.cdash.org/viewBuildError.php?type=1&buildid=7660676) eigen will eventually need updating.  So I figure at least moving to the latest release will be a first step.

phcerdan:
Fixes #3045


seanm:
Looks reasonable (not that I'll pretend to actually have reviewed all the eigen changes!)

phcerdan:
It's ready. I had to do a manual merge (copy files from upstream manually) for the Eigen/src/Geometry module, it was keeping old versions instead of new after conflicts. Not sure why.

I tested against an external module using Eigen (ITKTotalVariation) and works ok.

dzenanz:
The [update script](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/ThirdParty/Eigen3/UpdateFromUpstream.sh) should be updated. Take a look at a [recent example](https://github.com/InsightSoftwareConsortium/ITK/commit/fc46ec6bcefa1e1154a6524cfcd7eb36857a51cd). The new version works better, produces way less conflicts. Especially it does not produce stupid, avoidable conflicts.

Modules/ThirdParty/Eigen3/UpdateFromUpstream.sh
```diff
@@ -4,7 +4,7 @@ thirdparty_module_name="Eigen3"
 
 # readonly repo="https://gitlab.kitware.com/third-party/eigen.git"
 upstream_git_url="https://gitlab.kitware.com/phcerdan/eigen.git"
-upstream_git_branch="for/itk-3.3-find_package"
+upstream_git_branch="for/itk-3.4"
```
dzenanz, line 7: Changes which are not part of Eigen's upstream changes should be in separate commits, such as changes to this file. A recent example:
https://github.com/InsightSoftwareConsortium/ITK/pull/2832

dzenanz (commented):
Mostly looks good.

phcerdan:
> The [update script](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/ThirdParty/Eigen3/UpdateFromUpstream.sh) should be updated. Take a look at a [recent example](https://github.com/InsightSoftwareConsortium/ITK/commit/fc46ec6bcefa1e1154a6524cfcd7eb36857a51cd). The new version works better, produces way less conflicts. Especially it does not produce stupid, avoidable conflicts.

I am confused. There is a commit explicitly changing my original UpdateFromStream from what you call the new version to old version: 06d395f4bbb @vfonov 

> Changes which are not part of Eigen's upstream changes should be in separate commits, such as changes to this file

Thanks, will do.

Modules/ThirdParty/Eigen3/UpdateFromUpstream.sh
```diff
@@ -7,12 +7,10 @@ shopt -s dotglob
 readonly name="Eigen3"
 readonly ownership="Eigen Upstream <kwrobot@kitware.com>"
 readonly subtree="Modules/ThirdParty/Eigen3/src/itkeigen"
-# readonly repo="https://github.com/eigenteam/eigen-git-mirror"
```
dzenanz: Leaving the official mirror here will probably save somebody some searching.

Modules/ThirdParty/Eigen3/UpdateFromUpstream.sh
```diff
@@ -7,12 +7,10 @@ shopt -s dotglob
...
 # readonly repo="https://gitlab.kitware.com/third-party/eigen.git"
 readonly repo="https://gitlab.kitware.com/phcerdan/eigen.git"
-readonly tag="for/itk-eigen-3.3"
+readonly tag="for/itk-3.4"
 readonly paths="
-Eigen/CMakeLists.txt
```
dzenanz: Why remove automated updates for this file?

dzenanz (commented):
More questions again :(

Modules/ThirdParty/Eigen3/UpdateFromUpstream.sh
phcerdan: That's was the non-official mirror when eigen was in mercurial. Now they have a proper git version here:
https://gitlab.com/libeigen/eigen.

I will put it there.

Modules/ThirdParty/Eigen3/UpdateFromUpstream.sh
phcerdan: They removed it upstream, the Update script was failing because it.

dzenanz (approved):

seanm:
Not too surprisingly this added some colour in cdash... I'll take a look at it...

phcerdan:
> Not too surprisingly this added some colour in cdash... I'll take a look at it...

Just Yellow or RED :fearful:? 

Thanks for reporting upstream, we can update when they fix it. This Update was on top of their current 3.4 branch.

phcerdan:
Fixing warning `Wused-but-marked-unused` at #3095

Issue #3080, opened 2022-01-11
`itkConnectedRegionsMeshFilterPythonTest` is failing in Windows CI


tbirdso:
### Description

`itkConnectedRegionsMeshFilterPythonTest` is failing in [recent Windows Python CI results](https://open.cdash.org/viewTest.php?onlyfailed&buildid=7672621).

Seems to have started after the recent commit [4bf1434](https://github.com/InsightSoftwareConsortium/ITK/commit/4bf1434852743edb885efa9f1e6237c7a703b555) added the test.

### Steps to Reproduce

1. Build ITK Python wrapping with the same CMake configuration as in [AzurePipelinesWindowsPython.yml](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Testing/ContinuousIntegration/AzurePipelinesWindowsPython.yml)
2. Run `ctest -C Release -R itkConnectedRegionsMeshFilterPythonTest`

### Expected behavior

Test passes

### Actual behavior

```
2022-01-11T15:44:34.6727167Z 184: Traceback (most recent call last):
2022-01-11T15:44:34.6800473Z 184:   File "D:/a/1/s/Modules/Core/Mesh/wrapping/test/itkConnectedRegionsMeshFilterTest.py", line 28, in <module>
2022-01-11T15:44:34.7029045Z 184:     threshold_filter = itk.BinaryThresholdImageFilter.IUC3IUC3.New()
2022-01-11T15:44:34.7057735Z 184:   File "D:\a\1\s-build\Wrapping\Generators\Python\itk\support\lazy.py", line 130, in __getattribute__
2022-01-11T15:44:34.7370988Z 184:     value = types.ModuleType.__getattribute__(self, attr)
2022-01-11T15:44:34.7422543Z 184: AttributeError: module 'itk' has no attribute 'BinaryThresholdImageFilter'
2022-01-11T15:44:34.7721303Z 184: itkTestDriver: Process exited with return value: 1
2022-01-11T15:44:34.9520603Z 58/71 Test #184: itkConnectedRegionsMeshFilterPythonTest .............***Failed    1.47 sec
```

### Reproducibility

100%

### Versions

`master` branch

### Environment

Windows only (passes on macOS and Linux)

### Additional Information

- Short-term resolution is likely the same as #3057, where `Module_ITKThresholding` needs to be explicitly turned `ON` for the module to be appropriately wrapped in Windows CI
[...]

PR #3081, opened 2022-01-11
ENH: Adding Module dependency for Windows Python CI

PranjalSahu:
For windows Python CI adding ITKThresholding module.

### DIFF
--- a/Testing/ContinuousIntegration/AzurePipelinesWindowsPython.yml
+++ b/Testing/ContinuousIntegration/AzurePipelinesWindowsPython.yml
@@ -57,6 +57,7 @@ jobs:
           ITK_WRAP_PYTHON:BOOL=ON
           ITK_BUILD_DEFAULT_MODULES:BOOL=OFF
           ITKGroup_Core:BOOL=ON
+          Module_ITKThresholding:BOOL=ON
           Module_ITKImageIntensity:BOOL=ON
           Module_ITKBridgeNumPy:BOOL=ON
           Python3_FIND_REGISTRY=LAST


dzenanz (approved):

Testing/ContinuousIntegration/AzurePipelinesWindowsPython.yml
```diff
@@ -57,6 +57,7 @@ jobs:
           ITK_WRAP_PYTHON:BOOL=ON
           ITK_BUILD_DEFAULT_MODULES:BOOL=OFF
           ITKGroup_Core:BOOL=ON
+          ITKThresholding:BOOL=ON
```
tbirdso, line 60: In my CMake GUI I see the variable `Module_ITKThresholding` rather than `ITKThresholding`. Does `ITKThresholding` work when you try it?

Note that Windows Python CI are not run for each merge request due to Azure time constraints, so we won't be able to verify this change in PR CI. The best we can do is to merge and see what happens (not preferred) or test locally (better).

See discussion in #3080.

tbirdso (commented):
Thank you for tackling this one @PranjalSahu ! Making a quick note to verify we are setting the correct CMake variable.

Testing/ContinuousIntegration/AzurePipelinesWindowsPython.yml
PranjalSahu: Yes, I started with Module_ITKThresholding but changed it later to ITKThresholding. Let me check again. I didn't find the option in GUI.

PranjalSahu: So in this file I see ITKThresholding, [itk-module.cmake](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/Convolution/itk-module.cmake)

tbirdso: Good thinking, but I believe these refer to different properties. In [`itk-module.cmake`](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/Convolution/itk-module.cmake#L10) we are using a list of strings to define ITK module dependencies for both cxx and Python projects, whereas the variable `Module_ITKThresholding` is a boolean that we use to control which Python modules are wrapped.

Maybe @dzenanz or @thewtex can shed more light on the CMake architecture?

tbirdso: Note that the variables to turn off/on Python wrapping generation for different modules won't appear in the dashboard unless `ITK_BUILD_DEFAULT_MODULES` is set to `OFF`, otherwise I believe all modules are wrapped by default.

PranjalSahu: So earlier, I learnt from @dzenanz  that remote modules use Module_ in front not the core ITK modules.
But your finding in windows GUI is surprising. So yes we should ask them !

tbirdso: > So earlier, I learnt from @dzenanz that remote modules use Module_ in front

This is true, remote modules can be enabled via `Module_HASI`, `Module_Ultrasound`, etc

> not the core ITK modules

I'm not sure about this -- you can see in [AzurePipelinesWindowsPython.yml](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Testing/ContinuousIntegration/AzurePipelinesWindowsPython.yml#L60) where `Module_ITKImageIntensity` is turned `ON` to enable the core `ITKImageIntensity` module. It looks like our naming convention is to use `Module_ITK<module>` for enabling wrappings for individual modules.

> But your finding in windows GUI is surprising. So yes we should ask them !

👍

PranjalSahu: I am changing it to Module_ITKThresholding and testing a  build in Linux with same options as Windows Python CI.

dzenanz: Built-in modules have the `ITK` prefix in the name, e.g. ITKThresholding, ITKImageIntensity. Remote modules don't have it, e.g. Montage, Ultrasound. For enabling via CMake configuration, syntax is `Module_ModuleName:BOOL=ON`, so it should be `Module_ITKThresholding` here. If we wanted to enable a remote module, we would use e.g. `Module_Montage`.

PranjalSahu: Ok I mis-understood !

dzenanz (approved):

tbirdso (approved):
💯 LGTM!

PR #3082, opened 2022-01-11
COMP: Added mangling for 3 HDF5 functions that were missed

seanm:
So these 3 functions exist or not depending on preprocessing:

- the first two are wrapped in #ifndef NDEBUG
- the last one in #if H5C_DO_TAGGING_SANITY_CHECKS

I guess to be fully correct, when generating the mangling, we need to do it for release and debug, on different OSes, and with different build options, and merge the results.


seanm:
@dzenanz @hjmjohnson

dzenanz (approved):

hjmjohnson (approved):
Thanks sean.  Should we merge old symbols from previous manglings and the vtk mangle file as well?

seanm:
> Should we merge old symbols from previous manglings and the vtk mangle file as well?

I haven't seen a need...

PR #3083, opened 2022-01-11
BUG: Do not refuse to read images of vectors with many components

dzenanz:
This was encountered when reading 31-component image
which is float on disk, but is being read as double.

https://github.com/dzenanz/ITKUltrasound/blob/398ec5213b3b7d01a769ca6f32b2b36959a30761/test/itkSpectra1DAveragingImageFilterTest.cxx#L155

## PR Checklist


blowekamp:
Approach looks good...

Modules/IO/ImageBase/include/itkConvertPixelBuffer.hxx
```diff
@@ -481,6 +490,33 @@ ConvertPixelBuffer<InputPixelType, OutputPixelType, OutputConvertTraits>::Conver
...
+    {
+      OutputConvertTraits::SetNthComponent(c, *outputData, static_cast<OutputComponentType>(*(inputData + c)));
+    }
+    for (int c = componentCount; c < outputNumberOfComponents; ++c)
+    {
+      OutputConvertTraits::SetNthComponent(c, *outputData, 0); // set the rest of components to zero
+    }
+
+    ++outputData;
+    inputData += inputNumberOfComponents;
```
tbirdso, line 516: It looks like these two lines are both advancing to the next respective input/output pixel. Just to confirm, is there a reason for the difference in notation?

tbirdso (commented):
Just one clarification, otherwise LGTM!

Modules/IO/ImageBase/include/itkConvertPixelBuffer.hxx
dzenanz: I was following the pattern established by other similar code in that file.

dzenanz:
I can add the test. The "offending" image is 1x1 pixel with 31 components, so quite small.

thewtex (approved):

thewtex:
:tada: :taco: :t-rex:

PR #3077, opened 2022-01-11
ENH: Increase coverage for miscellaneous classes

jhlegarreta:
- STYLE: Prefer returning `int` instead of `bool` in test method 
- ENH: Increase coverage for miscellaneous classes

## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [X] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [X] Added test (or behavior not changed)

PR #3086, opened 2022-01-12
BUG: Guarantee array contains enough elements for access in loop

jhlegarreta:
Guarantee array contains enough elements for access in loop.
Specifically, the array must contain at least two elements so that the
sorting can take place.

## PR Checklist

### DIFF
--- a/Modules/Filtering/Thresholding/include/itkThresholdLabelerImageFilter.hxx
+++ b/Modules/Filtering/Thresholding/include/itkThresholdLabelerImageFilter.hxx
@@ -47,9 +47,9 @@ ThresholdLabelerImageFilter<TInputImage, TOutputImage>::BeforeThreadedGenerateDa
 {
   auto size = static_cast<unsigned int>(m_Thresholds.size());
 
-  for (unsigned int i = 0; i < size - 1; ++i)
+  for (unsigned int i = 1; i < size; ++i)
   {
-    if (m_Thresholds[i] > m_Thresholds[i + 1])
+    if (m_Thresholds[i - 1] > m_Thresholds[i])
     {
       itkExceptionMacro(<< "Thresholds must be sorted.");
     }


jhlegarreta:
Identified in #3077 build failures.

seanm:
For your interest: clang `-fsanitize=unsigned-integer-overflow` can flag unsigned overflow at runtime.  It's sometimes deliberate, and it is well-defined, but sometimes nice to catch.

Leengit:
If we specifically want the exception for `size == 0` then we'd keep that code.  However, if the goal is to make the the loop behave properly when `size == 0` then we could instead use:
```c++
for (unsigned int i = 1; i < size; ++i)
{
  if (m_Thresholds[i - 1] > m_Thresholds[i])
  {
    itkExceptionMacro(<< "Thresholds must be sorted.");
  }
}
```
We might use this version regardless so that we have insurance in case the exception is removed in the future.

jhlegarreta:
@Leengit thanks for the suggestion. I added a commit that implements your suggestion. If the zero-size exception were not present, the exception message would be confusing, so I think it should be kept.

Leengit:
Thank you for the quick change.

I looked a little deeper and it appears that `size == 0` is reasonable in some situations:
https://github.com/InsightSoftwareConsortium/ITK/blob/6a1575d68cea6d94a946533c79f42a8ab6aa3f5b/Modules/Filtering/Thresholding/include/itkThresholdLabelerImageFilter.h#L83-L86
However, I cannot say one way or another whether that means that the exception we are adding with this  pull request is appropriate.

That file has another loop terminated with `i < size - 1`.  That's okay because the code just quoted here handles the `size == 0` case, but defensive programming could be useful there too:
https://github.com/InsightSoftwareConsortium/ITK/blob/6a1575d68cea6d94a946533c79f42a8ab6aa3f5b/Modules/Filtering/Thresholding/include/itkThresholdLabelerImageFilter.h#L91-L98

jhlegarreta:
Thanks for the investigation @Leengit. I clearly don't have the bandwidth to investigate this further, i.e. what would happen if the thresholds array had no elements. That should be checked and eventually have a test added in a separate PR. I dropped the zero-size exception commit.

Leengit (approved):
Thank you.

PR #3091, opened 2022-01-13
STYLE: Remove old FFT 1D implementation includes

thewtex:
These are no longer required since we use the factory mechanism.


tbirdso (approved):
👍 LGTM

dzenanz (approved):

PR #3094, opened 2022-01-14
COMP: add C++11 override to various HDF5 methods that indeed override

seanm:
This fixes numerous warnings about missing override keywords.

These are all already fixed in upstream HDF5 but were not fixed by our recent update to HDF5 1.12.1 because the fixes post-date that release (which is the current release).


dzenanz (commented):
We can update to a tip of corresponding branch, e.g. `https://github.com/HDFGroup/hdf5/commits/hdf5_1_12`. We need to change it here: https://github.com/InsightSoftwareConsortium/ITK/blob/103e0a9329f63d7564c37c7252afda6c33093f8d/Modules/ThirdParty/HDF5/UpdateFromUpstream.sh#L11
Does that include these override specifications?

seanm:
No, it's still not backported to the 1.12 branch.  I've pinged them, but, well, it's not done yet.

Besides, this way has the benefit of us basically being on a stable release.

thewtex (approved):
Nice, @seanm !

PR #3095, opened 2022-01-14
COMP: Disable Eigen3 warning Wused-but-marked-unused

phcerdan:
Triggered using clang with std>=17
The warning is correct, in Eigen/x/arch/SSE the functions are used.
However in other archs the functions might be unused.
This patch hacks CMake to ignore it while waiting for resolution upstream.
Reported here: https://gitlab.com/libeigen/eigen/-/issues/2416

Another option to explore suggested by @seanm:
If we assume that ITK would only use the SSE arch we could remove the unused attribute.

### DIFF
--- a/Modules/ThirdParty/Eigen3/src/itkeigen/CMakeLists.txt
+++ b/Modules/ThirdParty/Eigen3/src/itkeigen/CMakeLists.txt
@@ -758,6 +758,12 @@ if(ITK_USE_EIGEN_MPL2_ONLY)
   target_compile_definitions (eigen_internal INTERFACE "EIGEN_MPL2_ONLY")
 endif()
 
+# Hack to disable warning. To remove when/if fixed upstream.
+# See issue: https://gitlab.com/libeigen/eigen/-/issues/2416
+if(CMAKE_CXX_STANDARD GREATER_EQUAL 17 AND CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  target_compile_options(eigen_internal INTERFACE "-Wno-used-but-marked-unused")
+endif()
+
 # BUILD: go one directory before to localize the headers:
 #   #include <itkeigen/Eigen/x>
 # INSTALL: headers require pre-prend itkeigen/Eigen/X.


dzenanz (approved):

Modules/ThirdParty/Eigen3/src/itkeigen/CMakeLists.txt
```diff
@@ -758,6 +758,12 @@ if(ITK_USE_EIGEN_MPL2_ONLY)
   target_compile_definitions (eigen_internal INTERFACE "EIGEN_MPL2_ONLY")
 endif()
 
+# Hack to disable warning. To remove when/if fixed upstream.
+# See issue: https://gitlab.com/libeigen/eigen/-/issues/2416
+if(CMAKE_CXX_STANDARD GREATER_EQUAL 17 AND CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  target_compile_definitions (eigen_internal INTERFACE "-Wnoused-but-marked-unused")
```
seanm, line 764: shouldn't that have another dash: `-Wno-used-but-marked-unused`?

seanm: Also, will this cause 'unknown warning' warnings on older clangs?

phcerdan: > shouldn't that have another dash: -Wno-used-but-marked-unused?

Indeed, fixed.

> Also, will this cause 'unknown warning' warnings on older clangs?

The attribute was introduced with c++17, so I am guessing any older clang will fail to compile with CXX_STANDARD < 17 anyway.

seanm: Yeah, probably. If not, hopefully we'll have a fix from upstream by then.

seanm (approved):

dzenanz (approved):

thewtex (approved):
:green_heart:

seanm:
So seems this did not work?

https://open.cdash.org/viewBuildError.php?type=1&buildid=7683217

phcerdan:
Ok, I think compiler id `Clang` does not include `AppleClang` since this policy.
https://cmake.org/cmake/help/latest/policy/CMP0025.html

Do you have any machine with an upstream Clang to see if the warnings were solved there?

phcerdan:
So, if that warning is introduced in c++17, and we are compiling with c++14, I assumed it shouldn't trigger.

PR #3098, opened 2022-01-14
BUG: Provide consistent shape on Fortran-order np.ndarray views

thewtex:
The itk.Image shape should be consistent for Fortran-order np.ndarray views.

Check the .base attr recursively and its .f_contiguous flag status.

This only applies to GetImageViewFromArray -- GetImageFromArray creates
a copy that is C-order.


aylward:
Nice!  Looking forward to trying this with MONAI!

thewtex:
Merging for 5.3rc3

PR #3102, opened 2022-01-14
BUG: Initialize variables

jhlegarreta:
- BUG: Initialize `itk::ExhaustiveOptimizer` ivars 
- BUG: Initialize `itk::Array` variable values in test 

## PR Checklist

### DIFF
--- a/Modules/Registration/Common/test/itkMultiResolutionImageRegistrationMethodTest.cxx
+++ b/Modules/Registration/Common/test/itkMultiResolutionImageRegistrationMethodTest.cxx
@@ -135,7 +135,8 @@ itkMultiResolutionImageRegistrationMethodTest(int, char *[])
   registration->SetInitialTransformParameters(initialParameters);
   ITK_TEST_SET_GET_VALUE(initialParameters, registration->GetInitialTransformParameters());
 
-  ParametersType initialTransformParametersOfNextLevel(1);
+  typename ParametersType::ValueType initialTransformParametersOfNextLevelVal(0.0);
+  ParametersType                     initialTransformParametersOfNextLevel(1, initialTransformParametersOfNextLevelVal);
   registration->SetInitialTransformParametersOfNextLevel(initialTransformParametersOfNextLevel);
   ITK_TEST_SET_GET_VALUE(initialTransformParametersOfNextLevel,
                          registration->GetInitialTransformParametersOfNextLevel());
--- a/Modules/Numerics/Optimizers/include/itkExhaustiveOptimizer.h
+++ b/Modules/Numerics/Optimizers/include/itkExhaustiveOptimizer.h
@@ -154,9 +154,9 @@ protected:
 
   ParametersType m_CurrentIndex;
 
-  SizeValueType m_MaximumNumberOfIterations;
+  SizeValueType m_MaximumNumberOfIterations{ 1 };
 
-  MeasureType m_MaximumMetricValue;
+  MeasureType m_MaximumMetricValue{ itk::NumericTraits<MeasureType>::max() };
 
   MeasureType m_MinimumMetricValue;
 


jhlegarreta:
Uncovered in #3077.

A few notes:
- Did the minimal set of changes that I believe will make `Valgrind` happy, but there are more uninitialized variables in `itk::ExhaustiveOptimizer`, and its great-grandparent `itk::Optimizer`.
- Have not investigated deeply, but the maximum metric value seems to depend on the optimizer and the position, but given that the former is not known at initialization, I decided to set to the numeric traits' max value.

Even if these classes do not belong to the `v4` versions, they may need some more investigation and potentially some further fixes in a separate PR.

thewtex (approved):
Excellent @jhlegarreta !

Modules/Numerics/Optimizers/src/itkExhaustiveOptimizer.cxx
```diff
@@ -31,6 +31,8 @@ ExhaustiveOptimizer::ExhaustiveOptimizer()
   m_CurrentValue = 0;
   m_CurrentParameter = 0;
   m_CurrentIndex.Fill(0);
+  m_MaximumNumberOfIterations = 1;
+  m_MaximumMetricValue = itk::NumericTraits<MeasureType>::max();
```
N-Dekker, line 35: Please consider placing this in a constructor member initializer list:
 
     :
     m_MaximumNumberOfIterations(1),
     m_MaximumMetricValue(itk::NumericTraits<MeasureType>::max())

N-Dekker: Or (preferably) C++11 in-class default member initialization, in the header file:

Replace:
https://github.com/InsightSoftwareConsortium/ITK/blob/cde72b3e7d716542b11d2dbb402bbe5816cf731f/Modules/Numerics/Optimizers/include/itkExhaustiveOptimizer.h#L157-L159
With:

    SizeValueType m_MaximumNumberOfIterations{ 1 };

    MeasureType m_MaximumMetricValue{ itk::NumericTraits<MeasureType>::max() };

Modules/Registration/Common/test/itkMultiResolutionImageRegistrationMethodTest.cxx
```diff
@@ -135,7 +135,7 @@ itkMultiResolutionImageRegistrationMethodTest(int, char *[])
   registration->SetInitialTransformParameters(initialParameters);
   ITK_TEST_SET_GET_VALUE(initialParameters, registration->GetInitialTransformParameters());
 
-  ParametersType initialTransformParametersOfNextLevel(1);
+  ParametersType initialTransformParametersOfNextLevel = ParametersType(1);
```
N-Dekker, line 138: Which problem did you encounter with the original code, `ParametersType initialTransformParametersOfNextLevel(1)`?

N-Dekker:
Can you please mention in your commit subject line that it is about the optimizers? For example (if I understand correctly):

    BUG: Initialize ExhaustiveOptimizer member variables

Modules/Numerics/Optimizers/src/itkExhaustiveOptimizer.cxx
jhlegarreta: Done in f1a52f1.

Modules/Registration/Common/test/itkMultiResolutionImageRegistrationMethodTest.cxx
jhlegarreta: As mentioned in the commit message, Valgrind detected that the values were not initialized.

jhlegarreta:
> Can you please mention in your commit subject line that it is about the optimizers? For example (if I understand correctly):

>     BUG: Initialize ExhaustiveOptimizer member variables

Done.

Modules/Numerics/Optimizers/src/itkExhaustiveOptimizer.cxx
N-Dekker: @jhlegarreta Thank you, Jon! 👍

Modules/Registration/Common/test/itkMultiResolutionImageRegistrationMethodTest.cxx
N-Dekker: I'm sorry, but I still don't see a _semantic_ difference between the original:

    ParametersType initialTransformParametersOfNextLevel(1);

and the proposed:

    ParametersType initialTransformParametersOfNextLevel = ParametersType(1);


Both lines of code call the `explicit OptimizerParameters(SizeValueType dimension)` constructor, which does _not_ initialize its elements.

I think you might fix the Valgrind issue by using the `explicit OptimizerParameters(const SizeValueType dimension, const ValueType & value)` constructor instead. For example:

    ParametersType initialTransformParametersOfNextLevel(1, 2.0);

If you do so, the initial parameter value will be `2.0`.

jhlegarreta: You're right. Overlook from my side. Fixed in 2942c8d.

Modules/Registration/Common/test/itkMultiResolutionImageRegistrationMethodTest.cxx
```diff
@@ -135,7 +135,8 @@ itkMultiResolutionImageRegistrationMethodTest(int, char *[])
   registration->SetInitialTransformParameters(initialParameters);
   ITK_TEST_SET_GET_VALUE(initialParameters, registration->GetInitialTransformParameters());
 
-  ParametersType initialTransformParametersOfNextLevel(1);
+  typename ParametersType::ValueType initialTransformParametersOfNextLevelVal(0.0);
```
N-Dekker, line 138: `typename ParametersType::ValueType`...? Interesting! Before C++11, such a use of `typename` was disallowed, because it is outside of a template. With C++11 and C++14, it should just compile fine  😃 

Only a very _very_ old compiler may complain, like GCC 4.4 at https://godbolt.org/z/xsP94bndW saying _error: using 'typename' outside of template_ Fortunately we do not need to support GCC 4.4 anymore  😄  So no problem!

N-Dekker (approved):
Thank you, Jon! 👍

hjmjohnson (approved):

PR #3106, opened 2022-01-15
STYLE: Default (`= default`) special member functions of `Neighborhood`

N-Dekker:
Added in-class default member initializers, in accordance with C++ Core
Guidelines, January 3, 2022: "Prefer in-class initializers to member
initializers in constructors for constant initializers"
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-in-class-initializer

Defaulted its default constructor, copy-constructor, and copy-assignment
operator.

No longer optimized self-assignment (`neighborhoodX = neighborhoodX`).
Note that this commit still supports self-assignment correctly. It just
does not have optimized code anymore for this very specific use case. In
accordance with C++ Core Guidelines, January 3, 2022: "Make copy
assignment safe for self-assignment"
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c62-make-copy-assignment-safe-for-self-assignment

### DIFF
--- a/Modules/Core/Common/include/itkNeighborhood.h
+++ b/Modules/Core/Common/include/itkNeighborhood.h
@@ -96,28 +96,20 @@ public:
   using NeighborIndexType = SizeValueType;
 
   /** Default constructor. */
-  Neighborhood()
-  {
-    m_Radius.Fill(0);
-    m_Size.Fill(0);
-    for (DimensionValueType i = 0; i < VDimension; ++i)
-    {
-      m_StrideTable[i] = 0;
-    }
-  }
+  Neighborhood() = default;
 
   /** Default destructor. */
   virtual ~Neighborhood() = default;
 
   /** Copy constructor. */
-  Neighborhood(const Self & other);
+  Neighborhood(const Self &) = default;
 
   /** Move-constructor. */
   Neighborhood(Self &&) = default;
 
   /** Assignment operator. */
   Self &
-  operator=(const Self & other);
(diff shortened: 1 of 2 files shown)


hjmjohnson (approved):

thewtex (approved):

Modules/Core/Common/include/itkNeighborhood.h
```diff
@@ -96,28 +96,20 @@ class ITK_TEMPLATE_EXPORT Neighborhood
...
-  Neighborhood()
-  {
-    m_Radius.Fill(0);
-    m_Size.Fill(0);
-    for (DimensionValueType i = 0; i < VDimension; ++i)
-    {
-      m_StrideTable[i] = 0;
-    }
-  }
+  Neighborhood() = default;
```
dzenanz, line 99: Could we apply the rule of zero here? Declare no constructor, assignment or move operators?

dzenanz (approved):

Modules/Core/Common/include/itkNeighborhood.h
N-Dekker: @dzenanz  Thanks for your comment!

> Could we apply the rule of zero here? Declare no constructor, assignment or move operators?

Not entirely _zero_! We must still keep the destructor of `itk::Neighborhood`, as it is virtual. So because we have an explicitly declared destructor, I think it is reasonable to declare all of them (albeit `= default`).

PR #3108, opened 2022-01-16
DOC: Remove mistake from `ImageRegion::IsInside(region)` documentation

N-Dekker:
Reverted commit a872710c63ab223e78f45de8d3e4bc435172867d
"STYLE: Documenting the behavior of the IsInside() method when the argument region has zero size in any of its dimensions" by Luis Ibanez (@luisibanez), 18 December 2009

Unlike the original documentation, `ImageRegion::IsInside(region)` may in some cases return true, even when the argument region has zero size!

Examples:

    using RegionType = ImageRegion<2>;
    using IndexType = RegionType::IndexType;
    using SizeType = RegionType::SizeType;

    const RegionType currentRegion1(IndexType::Filled(-1), SizeType::Filled(2));
    const RegionType zeroSizedRegion1{};

    EXPECT_TRUE(currentRegion1.IsInside(zeroSizedRegion1)); // OK: returns true!

    const RegionType currentRegion2(SizeType::Filled(2));
    const RegionType zeroSizedRegion2(IndexType::Filled(1), SizeType{ { 0 } });

    EXPECT_TRUE(currentRegion2.IsInside(zeroSizedRegion2)); // OK: returns true!

Simplified the implementation of `ImageRegion::IsInside(const Self &)`, assuming that it may return true for a zero-sized region as argument, when the index (start corner position) of this zero-sized region is inside the current region.

### DIFF
--- a/Modules/Core/Common/include/itkImageRegion.h
+++ b/Modules/Core/Common/include/itkImageRegion.h
@@ -283,10 +283,7 @@ public:
     return true;
   }
 
-  /** Test if a region (the argument) is completely inside of this region. If
-   * the region that is passed as argument to this method, has a size of value
-   * zero, then it will not be considered to be inside of the current region,
-   * even its starting index is inside. */
+  /** Test if a region (the argument) is completely inside of this region. */
   bool
   IsInside(const Self & region) const
   {
--- a/Modules/Core/Common/include/itkImageRegion.h
+++ b/Modules/Core/Common/include/itkImageRegion.h
@@ -285,23 +285,18 @@ public:
 
   /** Test if a region (the argument) is completely inside of this region. */
   bool
-  IsInside(const Self & region) const
+  IsInside(const Self & otherRegion) const
   {
-    IndexType beginCorner = region.GetIndex();
+    const auto otherIndex = otherRegion.m_Index;
+    const auto otherSize = otherRegion.m_Size;
 
-    if (!this->IsInside(beginCorner))
-    {
-      return false;
-    }
-    IndexType        endCorner;
-    const SizeType & size = region.GetSize();
     for (unsigned int i = 0; i < ImageDimension; ++i)
     {
-      endCorner[i] = beginCorner[i] + static_cast<OffsetValueType>(size[i]) - 1;
-    }
-    if (!this->IsInside(endCorner))
-    {
-      return false;
+      if (otherIndex[i] < m_Index[i] || otherIndex[i] + static_cast<IndexValueType>(otherSize[i]) >
+                                          m_Index[i] + static_cast<IndexValueType>(m_Size[i]))
+      {
+        return false;
+      }
     }
     return true;
   }
--- a/Modules/Core/Common/test/itkImageRegionGTest.cxx
+++ b/Modules/Core/Common/test/itkImageRegionGTest.cxx
@@ -20,7 +20,6 @@
 #include "itkImageRegion.h"
 #include "itkIndexRange.h"
 #include <gtest/gtest.h>
-#include <algorithm>   // For count.
 #include <type_traits> // For remove_const_t and remove_reference_t.
 
 
@@ -32,29 +31,17 @@ TEST(ImageRegion, ZeroSizedRegionIsInside)
     using SizeType = typename RegionType::SizeType;
 
     auto paddedRegion = region;
-    paddedRegion.PadByRadius(1);
+    paddedRegion.PadByRadius(2);
+
+    // Extend the specified region (in 2D, by an extra column).
+    const RegionType extendedRegion(region.GetIndex(), region.GetSize() + SizeType::Filled(1));
 
     for (const auto & index : itk::ImageRegionIndexRange<RegionType::ImageDimension>(paddedRegion))
     {
       const RegionType zeroSizedRegion(index, SizeType{ { 0 } });
 
-      const auto offset = index - region.GetIndex();
-
-      if (std::count(offset.begin(), offset.end(), 0) == 0)
-      {
-        // In this case (the most common case) the one-sized region is considered inside this region if and only if its
-        // index is inside this region.
-        EXPECT_EQ(region.IsInside(zeroSizedRegion), region.IsInside(index));
-      }
-      else
-      {
-        // In this border case, at least one of the index values of the zero-sized region appears equal to the
-        // corresponding index value of this region. (Which means for 2D that the zero-sized region is on the left or
-        // upper border of this region.) For historical reasons (based on the original implementation of
-        // `itk::ImageRegion::IsInside` from 5 February 2002), the zero-sized region is then considered _outside_ of
-        // this region.
-        EXPECT_FALSE(region.IsInside(zeroSizedRegion));
-      }
+      // The zero-sized region is considered inside this region if and only if its index is inside the extended region.
+      EXPECT_EQ(region.IsInside(zeroSizedRegion), extendedRegion.IsInside(index));
     }
   };
 
--- a/Modules/Core/Common/test/itkImageRegionGTest.cxx
+++ b/Modules/Core/Common/test/itkImageRegionGTest.cxx
@@ -1,0 +1,90 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+// First include the header file to be tested:
+#include "itkImageRegion.h"
+#include "itkIndexRange.h"
+#include <gtest/gtest.h>
+#include <algorithm>   // For count.
+#include <type_traits> // For remove_const_t and remove_reference_t.
+
+
+// Tests when zero-sized regions are considered to be inside of a region.
+TEST(ImageRegion, ZeroSizedRegionIsInside)
+{
+  const auto check = [](const auto & region) {
+    using RegionType = std::remove_const_t<std::remove_reference_t<decltype(region)>>;
+    using SizeType = typename RegionType::SizeType;
+
+    auto paddedRegion = region;
+    paddedRegion.PadByRadius(1);
+
+    for (const auto & index : itk::ImageRegionIndexRange<RegionType::ImageDimension>(paddedRegion))
+    {
+      const RegionType zeroSizedRegion(index, SizeType{ { 0 } });
+
+      const auto offset = index - region.GetIndex();
+
+      if (std::count(offset.begin(), offset.end(), 0) == 0)
+      {
+        // In this case (the most common case) the one-sized region is considered inside this region if and only if its
+        // index is inside this region.
+        EXPECT_EQ(region.IsInside(zeroSizedRegion), region.IsInside(index));
+      }
+      else
+      {
+        // In this border case, at least one of the index values of the zero-sized region appears equal to the
+        // corresponding index value of this region. (Which means for 2D that the zero-sized region is on the left or
+        // upper border of this region.) For historical reasons (based on the original implementation of
+        // `itk::ImageRegion::IsInside` from 5 February 2002), the zero-sized region is then considered _outside_ of
+        // this region.
+        EXPECT_FALSE(region.IsInside(zeroSizedRegion));
+      }
+    }
+  };
+
+  // Check for a 2D and a 3D image region.
+  check(itk::ImageRegion<2>(itk::Size<2>::Filled(3)));
+  check(itk::ImageRegion<3>(itk::MakeIndex(-1, 0, 1), itk::MakeSize(2, 3, 4)));
+}
+
+
+// Tests that regions of size 1 are considered to be inside of a region, if and only if ("iff") their index is inside of
+// this region.
+TEST(ImageRegion, OneSizedRegionIsInsideIffItsIndexIsInside)
+{
+  const auto check = [](const auto & region) {
+    using RegionType = std::remove_const_t<std::remove_reference_t<decltype(region)>>;
+    using SizeType = typename RegionType::SizeType;
+
+    auto paddedRegion = region;
+    paddedRegion.PadByRadius(1);
+
+    for (const auto & index : itk::ImageRegionIndexRange<RegionType::ImageDimension>(paddedRegion))
+    {
+      const RegionType oneSizedRegion{ index, SizeType::Filled(1) };
+
+      // The one-sized region is inside this region if and only if its index is inside this region.
+      EXPECT_EQ(region.IsInside(oneSizedRegion), region.IsInside(index));
+    }
+  };
+
+  // Check for a 2D and a 3D image region.
+  check(itk::ImageRegion<2>(itk::Size<2>::Filled(3)));
+  check(itk::ImageRegion<3>(itk::MakeIndex(-1, 0, 1), itk::MakeSize(2, 3, 4)));
+}
--- a/Modules/Core/Common/test/CMakeLists.txt
+++ b/Modules/Core/Common/test/CMakeLists.txt
@@ -616,6 +616,7 @@ set(ITKCommonGTests
       itkImageBufferRangeGTest.cxx
       itkImageRegionRangeGTest.cxx
       itkImageIORegionGTest.cxx
+      itkImageRegionGTest.cxx
       itkIndexGTest.cxx
       itkIndexRangeGTest.cxx
       itkMatrixGTest.cxx


N-Dekker:
Discussion topic at discourse.itk.org: [What bool value should `imageRegion.IsInside(zeroSizedRegion)` return?](https://discourse.itk.org/t/what-bool-value-should-imageregion-isinside-zerosizedregion-return/4734)

dzenanz (approved):
I like this better.

Modules/Core/Common/include/itkImageRegion.h
```diff
@@ -285,23 +285,18 @@ class ITK_TEMPLATE_EXPORT ImageRegion final : public Region
 
   /** Test if a region (the argument) is completely inside of this region. */
   bool
-  IsInside(const Self & region) const
+  IsInside(const Self & otherRegion) const
   {
-    IndexType beginCorner = region.GetIndex();
+    const auto otherIndex = otherRegion.m_Index;
```
hjmjohnson, line 290: If this were "const auto & otherIndex  = otherRegion.m_index" then it would not need to make a copy of the information.  

Even better:  Remove these constant aliases, and just use "otherRegion.m_index[i]" below.

hjmjohnson (approved):
This is an improvement. I have a comment that can be ignored if you like.

Modules/Core/Common/include/itkImageRegion.h
dzenanz: I think the compiler is likely to optimize it away either way.

N-Dekker:
Please note that the second commit, "PERF: Simplify `ImageRegion::IsInside(const Self &)`" does introduce a very small behavior change, compared to the old code. The old code considered zero-sized regions to be _inside_ the current region when their index (their start corner position) would be inside the current region, _except_ when such an index would be at the left or top border of the current region!

The commit "PERF: Simplify `ImageRegion::IsInside(const Self &)`" does not make such an exception. With this commit, zero-sized regions are considered inside the current region whenever their index is inside the current region.

This little behavior change did not cause any test regression error.

Modules/Core/Common/include/itkImageRegion.h
```diff
@@ -283,28 +283,20 @@ class ITK_TEMPLATE_EXPORT ImageRegion final : public Region
...
-    const SizeType & size = region.GetSize();
     for (unsigned int i = 0; i < ImageDimension; ++i)
     {
-      endCorner[i] = beginCorner[i] + static_cast<OffsetValueType>(size[i]) - 1;
-    }
-    if (!this->IsInside(endCorner))
-    {
-      return false;
+      if ((otherIndex[i] < m_Index[i]) || ((otherIndex[i] + static_cast<IndexValueType>(otherSize[i])) >
+                                           (m_Index[i] + static_cast<IndexValueType>(m_Size[i]))))
```
Leengit, line 296: `+` has higher operator precedence than '<` and '>'; and `<` and `>` have higher precedence than `||`. So some parentheses can be removed here.  (Unless this is to influence `clang-format`??)

Please don't slow down the merge just because of this trivia.

N-Dekker: @Leengit Thanks for your suggestion. I just placed those extra parentheses there for clarity. But I'm willing to remove them if it is the preferred coding style nowadays. Does anyone else have an opinion on them?

Leengit:
If the outer region has an upper left corner `(0, 0)` and is size `(10, 10)` then its lower right corner is `(9, 9)`, right?  I didn't check the old code, but with the new code any region with corner `(x, y)` and size `(0, 0)` will be inside so long as `x, y ∈ {0, 1, ..., 10}`.  In particular, `(x, y) = (10, 10)` is inside.  Is this what we want?  Regardless of what we want it to be, if it doesn't already, the documentation should probably mention the behavior for this case.

In a more purely mathematical sense, I would think that if any dimension is zero then the region has no pixels; and all the pixels -- all zero of them -- are in the bigger region (even if it too has zero pixels).  So, we could return true in that case _regardless of the corner_.  But that's math not image analysis.  We should do what image analysis demands, and probably document any differences from this mathematical approach in the documentations.

Modules/Core/Common/include/itkImageRegion.h
N-Dekker: > I think the compiler is likely to optimize it away either way.

Possible. However, I do believe the commit as a whole could significantly improve the performance, as it does less comparisons than the original code. And it appears that `ImageRegion::IsInside` is called quite frequently.

Modules/Core/Common/include/itkImageRegion.h
dzenanz: Now that Lee mentioned it, I agree that there could be less parentheses.

N-Dekker: > I agree that there could be less parentheses.

How many do you want?  😸

N-Dekker:
> If the outer region has an upper left corner `(0, 0)` and is size `(10, 10)` then its lower right corner is `(9, 9)`, right? I didn't check the old code, but with the new code any region with corner `(x, y)` and size `(0, 0)` will be inside so long as `x, y ∈ {0, 1, ..., 10}`. In particular, `(x, y) = (10, 10)` is inside. Is this what we want?

That certainly is not what I expect! In your case, any region with index `(10, 10)` should be considered outside the outer region, according to both the original and the proposed new code.

Is the following what you mean?

    using RegionType = ImageRegion<2>;
    using IndexType = RegionType::IndexType;
    using SizeType = RegionType::SizeType;

    const RegionType outerRegion(IndexType{ { 0 } }, SizeType::Filled(10));
    const RegionType zeroSizedRegion(IndexType::Filled(10), SizeType{ { 0 } });

    EXPECT_FALSE(outerRegion.IsInside(zeroSizedRegion));

Leengit:
> EXPECT_FALSE(outerRegion.IsInside(zeroSizedRegion));

I don't know the use case for this code, so I don't know what the right answer should be for this zero-sized region, or for zero-sized regions generally.  The current code in this pull request will give `true`.  However, if that's wrong then, yes, your code is the correct test.

If the mathy approach described above is correct and we should be returning `true` whenever `otherRegion` has zero pixels, then the source code could be
```c++
bool
IsInside(const Self & otherRegion) const
{
  const auto otherIndex = otherRegion.m_Index;
  const auto otherSize = otherRegion.m_Size;

  // If otherRegion has zero pixels then every pixel -- all zero of them -- are inside the current region.
  for (unsigned int i = 0; i < ImageDimension; ++i)
  {
    if (otherSize[i] == 0)
    {
      return true;
    }
  }

  for (unsigned int i = 0; i < ImageDimension; ++i)
  {
    if (otherIndex[i] < m_Index[i] ||
        otherIndex[i] + static_cast<IndexValueType>(otherSize[i]) > m_Index[i] + static_cast<IndexValueType>(m_Size[i]))
    {
      return false;
    }
  }
  return true;
}
```

Leengit:
The comment deleted by this pull request
```c++
  /** Test if a region (the argument) is completely inside of this region. If
   * the region that is passed as argument to this method, has a size of value
   * zero, then it will not be considered to be inside of the current region,
   * even its starting index is inside. */
```
indicates the opposite of my mathy example: that is, if `otherRegion` has zero pixels then always return false.  If that is our goal then the code can go back to using one loop:
```c++
bool
IsInside(const Self & otherRegion) const
{
  const IndexType otherIndex = otherRegion.m_Index;
  const SizeType  otherSize = otherRegion.m_Size;

  // If otherRegion has zero pixels or is not wholly inside then return false.
  for (unsigned int i = 0; i < ImageDimension; ++i)
  {
    if (otherSize[i] == 0 || otherIndex[i] < m_Index[i] ||
        otherIndex[i] + static_cast<IndexValueType>(otherSize[i]) > m_Index[i] + static_cast<IndexValueType>(m_Size[i]))
    {
      return false;
    }
  }
  return true;
}
```

N-Dekker:
> > `EXPECT_FALSE(outerRegion.IsInside(zeroSizedRegion));`
> 
> I don't know the use case for this code, so I don't know what the right answer should be for this zero-sized region, or for zero-sized regions generally. The current code in this pull request will give `true`. However, if that's wrong then, yes, your code is the correct test.

I did not realize before that my proposed adjustment would consider a zero-sized region with index (10, 10) _inside_ your outer region of size 10x10, so thanks for the heads up, Lee! 😮  

So now I'm not entirely sure anymore.  I do like the _relative_ simplicity of having just a single `for` loop and a single `if` statement, as currently in this PR:

    if ((otherIndex[i] < m_Index[i]) || ((otherIndex[i] + static_cast<IndexValueType>(otherSize[i])) >
                                          (m_Index[i] + static_cast<IndexValueType>(m_Size[i]))))
    {
      return false;
    }

I don't really like to have extra code, just for the very specific case of a zero-sized region. Especially because in practice, the return value does not seem to matter much for a zero-sized region as argument. On the other hand, I don't want the result to be counterintuitive 🤔

The original code treated the left and the upper border positions of the outer region different from the rest of the outer region, when used as the index of a zero-sized region as argument. That did not seem intuitive either, did it?

N-Dekker:
> The comment deleted by this pull request
> 
> ```c++
>   /** Test if a region (the argument) is completely inside of this region. If
>    * the region that is passed as argument to this method, has a size of value
>    * zero, then it will not be considered to be inside of the current region,
>    * even its starting index is inside. */
> ```
> 
> indicates the opposite of my mathy example

@Leengit Indeed! That's my alternative for this pull request: #3110 "BUG: `region.IsInside(zeroSizedRegion)` should always return false"  😄

Modules/Core/Common/include/itkImageRegion.h
N-Dekker: @Leengit @dzenanz Now I see at https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/blob/36b553127ead40247889028306fc2fd566aa1e65/SoftwareGuide/Latex/Appendices/CodingStyleGuide.tex#L2203-L2209 

```
Unnecessary parentheses for grouping hinder reading the code.
The C++ precedence and associativity (the order in which the operands are
evaluated) of operators must be taken into account to avoid using unnecessary
parentheses.

As a general principle, these apply to condition expressions, and statements
where mathematical, logical or bitwise operations are performed.
```

C++ Core Guidelines (January 3, 2022) appears to encourage the use of parentheses only _slightly_ more (specifically for _bitwise_ logical operations): "If in doubt about operator precedence, parenthesize" https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-parens

OK, so (from now) I will (try to) change my habit regarding this style (by reducing the number of parentheses (that I will be using)).  😺

Leengit: Just to document: The bitwise operators (`&`, `|`, ...), like the logical operators (`&&`, `||`), have an operator precedence lower than the relational operators (`<`, `==`, ...)  and can be used for logical operations that don't need short circuiting.  However when the intent, like with arithmetic (`+`, `-`, ...), is to have the precedence be higher than the relational operators then parentheses are usually required.  I try to avoid the parentheses in the first use case, though sometimes people will use them anyway to make sure that folks are not confused with the precedence that would be appropriate for the second use case.

blowekamp:
I don't have time to read this whole thread. There is a mixing of documentation changes, behavioral changes, performance changes and style changes.

Please keep the current behavior, add test to ensure that it is preserved. It is very likely some code inside or outside ITK relies on the current behavior. Perhaps the ITK test still passes, because it is not correctly checking the results. Or code outside ITK, is not part to ITK regressions checks.

Again please keep the current behavior.

N-Dekker:
@blowekamp 
> I don't have time to read this whole thread
...
> Again please keep the current behavior.

Please take your time to read the issue, Bradley. The current version is both unnecessary slow and inconsistent, with respect to zero-sized regions. I think it must be fixed one way or the other.

I can imagine that we leave the old (slow and inconsistent) behavior behind an #if LEGACY, though.

blowekamp:
Correct me if I am wrong. You encountered a discrepancy between behavior and documentation. I support correcting the documentation to what the community has been using for over a decade. You have a theoretical ( not measured ) performance improvement, that is also not driven by improvement to a particular application.

I don't find it worth the risk for breaking the communities code, and waisting their time tracking down the bug for these theoretical reasons. People can move for other tool kits when frustrated with breaking and unpredictable changes in a toolkit.

And no mater what the file change is the behavior should be correctly documented in the methods doxygen string, and a test added for this special case.

Leengit:
To be clear the behavior of the old code is as follows.  With `InputRegion.IsInside(QueryRegion)` for an `InputRegion` that has index `(0, 0)` and size `(10, 10)` and a `QueryRegion` that has index `(x, y)` and size `(1, 1)`, the method returns `true` exactly when `x, y ∈ {0, 1, ..., 9}`.  That's normal.

If instead `QueryRegion` has index `(x, y)` and size `(0, 0)` then the method returns `true` exactly when `x, y ∈ {1, ..., 9}`.  That is, the top row and left column are now excluded.  I am thinking that the relevant question is whether it is likely that there is code out there that is depending on this behavior.

blowekamp:
For a paleontological perspective, when the documentation commit was made in 2009 was when I was working on the initial IO streaming making heavy use of regions and region splitters.

Thank you for the clear example @Leengit, looks like some good cases for some tests.

Modules/Core/Common/include/itkImageRegion.h
N-Dekker: OK! Thanks! Just removed those parentheses that aren't strictly necessary (and also did `git rebase master`), with my latest [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/165e9e74e49393eebb3001e91e770d8641589d8c..25ee2e960d70b8a7a2a133400d8b815605d6707b)

dzenanz (approved):

Leengit (approved):
I am happy with either this pull request or Pull Request #3110.  Either is an improvement.

dzenanz:
@N-Dekker since Jon Haitz is taking a break, I feel obliged to point this out: it would be good if you added the various examples from this PR's discussion as test cases in the code.

N-Dekker:
> it would be good if you added the various examples from this PR's discussion as test cases in the code.

@dzenanz  Just added a unit test, with my latest [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/25ee2e960d70b8a7a2a133400d8b815605d6707b..667820d94f28a89cd4c9a6e5df738e0c150dad42) Hope you like it  😃

dzenanz (approved):
I like it. Can we test some indices which are outside?

N-Dekker:
> I like it. Can we test some indices which are outside?

Thanks @dzenanz. Yes, we can test some indices outside the outer region, but then, what would you expect for a zero-sized region?  Or would it be OK to you to just do this test for _non-zero_-sized regions (and then, expect `IsInside` to always return false, of course.)

dzenanz:
The more tests, the better. Ideally test both zero-sized and non-zero-sized regions with start indices outside of the main region.

Modules/Core/Common/test/itkImageRegionGTest.cxx
```diff
@@ -0,0 +1,90 @@
...
+    auto paddedRegion = region;
+    paddedRegion.PadByRadius(1);
+
+    for (const auto & index : itk::ImageRegionIndexRange<RegionType::ImageDimension>(paddedRegion))
+    {
+      const RegionType zeroSizedRegion(index, SizeType{ { 0 } });
+
+      const auto offset = index - region.GetIndex();
+
+      if (std::count(offset.begin(), offset.end(), 0) == 0)
```
dzenanz, line 43: I am fine with eliminating this exception. I am certain this was not intentional. At first I didn't even understand where it came from 😄

N-Dekker: @dzenanz My intention is to first have a unit test that describes the original behavior, including the exceptional border cases, and then, when it has passed on the CI, propose the adjustment I had in mind to _both_ the implementation and the test. Please wait...  😃

Modules/Core/Common/test/itkImageRegionGTest.cxx
```diff
@@ -32,29 +31,17 @@ TEST(ImageRegion, ZeroSizedRegionIsInside)
     using SizeType = typename RegionType::SizeType;
 
     auto paddedRegion = region;
-    paddedRegion.PadByRadius(1);
+    paddedRegion.PadByRadius(2);
+
+    // Extend the specified region (in 2D, by an extra column).
+    const RegionType extendedRegion(region.GetIndex(), region.GetSize() + SizeType::Filled(1));
```
dzenanz, line 37: This feels like we have traded one unintended exception for another.

dzenanz (approved):

Modules/Core/Common/test/itkImageRegionGTest.cxx
N-Dekker: Indeed! I only realized that this PR would trade one unintended exception for another from the comment https://github.com/InsightSoftwareConsortium/ITK/pull/3108#issuecomment-1014808455 by @Leengit !

As an alternative, we could reconsider PR #3110 "BUG: region.IsInside(zeroSizedRegion) should always return false" 😸

Leengit: What say we go with PR #3110 instead?

dzenanz: We could. Please move the tests over there 😄

N-Dekker: PR #3110 already has a test, and it's very simple  😄 https://github.com/InsightSoftwareConsortium/ITK/blob/85268b9455bf22973b30fc8755f2a3b56df883be/Modules/Core/Common/test/itkImageRegionGTest.cxx#L24-L25

dzenanz: I meant copy the tests which do no involve zero-sized region.

N-Dekker: @dzenanz Just updated PR #3110 by copying the test for one-sized regions to its GTest. Note that the old-style itkImageRegionTest already had a few tests that involve non-zero-sized regions as well. They still pass successfully (without any adjustment) with either one of the two pull requests, #3108 or #3110.

dzenanz:
After the discussion here, I now prefer the other solution, PR #3110.

Modules/Core/Common/test/itkImageRegionGTest.cxx
N-Dekker: @dzenanz 
> This feels like we have traded one unintended exception for another.

Of course, the unintended exception of this PR could also be fixed by adding an extra check that the index of the inner region (argument) may not be beyond the end of the outer region.  As I mentioned at https://github.com/InsightSoftwareConsortium/ITK/pull/3108#issuecomment-1014906560 ("Another possibility is to add an extra check, to ensure that a zero-sized region is considered outside, when its index is outside"), including a possible implementation. With the extra check, IsInside(zeroSizedRegion) would be true, if and only if IsInside(zeroSizedRegion.GetIndex()) would be true. It would not affect the behavior of non-zero-sized regions.

dzenanz:
I just merged #3110, so I am closing this competing PR.

PR #3114, opened 2022-01-17
COMP: Disable Eigen3 warning in AppleClang

phcerdan:
Follow on: #3095
`Clang` does not include `AppleClang`: https://cmake.org/cmake/help/latest/policy/CMP0025.html


phcerdan:
Not really a solution, I think the problem is that the warning is used with older standards? Conversation in #3095

PR #3115, opened 2022-01-17
BUG: fixed simultaneous read/write access of same ivar by different t…

seanm:
…hreads

Made the ivar atomic, so at least the simultaneous access is well-defined.

Found by TSan.  A great many tests were failing because of this.

### DIFF
--- a/Modules/Core/Common/include/itkMultiThreaderBase.h
+++ b/Modules/Core/Common/include/itkMultiThreaderBase.h
@@ -35,6 +35,7 @@
 #include "itkImageRegion.h"
 #include "itkImageIORegion.h"
 #include "itkSingletonMacro.h"
+#include <atomic>
 #include <functional>
 #include <thread>
 #include "itkProgressReporter.h"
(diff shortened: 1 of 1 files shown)


dzenanz (approved):

seanm:
BTW I think this is a recent regression, but did not check history/blame.  TSan used to have only hundreds of failures, but now has thousands. I think most are due to this.

dzenanz:
Can you take a look at #3116 and try to figure out the hang in [a few filters](https://github.com/InsightSoftwareConsortium/ITK/pull/3014#issuecomment-1005249793)?
```text
	209 - PythonGradientMagnitudeRecursiveGaussianImageFilterTest (Timeout)
	210 - itkGradientVectorFlowImageFilterPythonTest (Timeout)
	211 - PythonLaplacianImageFilterTest (Timeout)
	215 - PythonGeodesicActiveContourLeftVentricleTest (Timeout)
```
It might not be even all of these now, but only two of them (Laplacian and one more).

hjmjohnson (approved):

PR #3120, opened 2022-01-17
COMP: Remove use of deprecated functions

hjmjohnson:
Make deprecation of macros clearer using the
'pragma GCC poison' mechansim.

Fix errant use of deprecated macro names.


jhlegarreta (approved):
:+1: Thanks Hans. I take the blame for still having used the old macro names in the test you fixed.

thewtex (approved):

dzenanz (approved):

PR #3121, opened 2022-01-17
ENH: Add `FFTDiscreteGaussianImageFilter`

tbirdso:
Changes:
- Implements `FFTDiscreteGaussianImageFilter` as a subclass of `DiscreteGaussianImageFilter` to perform convolution of an ND single-channel input image with a Gaussian kernel of fixed width in the Fourier domain
- Implements `FFTDiscreteGaussianImageFilterEnums` to select among kernel generation methods, where `COMBINED_OPERATORS` composes through vector multiplication the exact same ND kernel as used in `DiscreteGaussianImageFilter` in image format, while `IMAGE_SOURCE` relies on the (potentially faster) method of directly generating the kernel image through `itk::GaussianImageSource`
- Exposes protected methods for generating the kernel from error/width/spacing inputs in the base `DiscreteGaussianImageFilter` class so that `FFTDiscreteGaussianImageFilter` can use the same process for generating a kernel to specifications
- Adds `FFTDiscreteGaussianImageFilterFactory` for manually overriding `DiscreteGaussianImageFilter` for FFT acceleration benefits
- Updates test coverage, including a test comparing `FFTDiscreteGaussianImageFilter` output to `DiscreteGaussianImageFilter` output with the same parameters
- Includes Python wrapping

Input:
2th_cthead1

Test output screenshot:

fft-output

## PR Checklist
- [x] Updated API documentation (or API not changed)\


Modules/Filtering/Smoothing/test/itkFFTDiscreteGaussianImageFilterTest.cxx
```diff
@@ -0,0 +1,71 @@
...
+
+  auto writer = WriterType::New();
+  writer->SetFileName(argv[4]);
+  writer->SetInput(filter->GetOutput());
+  writer->UseCompressionOn();
+  writer->Update();
+
+  return EXIT_SUCCESS;
+
+  return EXIT_SUCCESS;
```
thewtex, line 70: duplicate

thewtex (commented):
Looking great!

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.h
```diff
@@ -0,0 +1,141 @@
...
+ * \sa Neighborhood
+ * \sa NeighborhoodOperator
+ * \sa RecursiveGaussianImageFilter
+ *
+ * \ingroup ImageEnhancement
+ * \ingroup ImageFeatureExtraction
+ * \ingroup ITKSmoothing
+ *
+ * \sphinx
+ * \sphinxexample{Filtering/Smoothing/SmoothWithRecursiveGaussian,Computes the smoothing with Gaussian kernel}
```
dzenanz, line 50: Does sphinx example `SmoothWithRecursiveGaussian` use `FFTDiscreteGaussianImageFilter`? If no, this line should be removed.

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.h
```diff
@@ -0,0 +1,141 @@
...
+  using Self = FFTDiscreteGaussianImageFilter;
+  using Superclass = DiscreteGaussianImageFilter<TInputImage, TOutputImage>;
+  using Pointer = SmartPointer<Self>;
+  using ConstPointer = SmartPointer<const Self>;
+
+  /** Method for creation through the object factory. */
+  itkNewMacro(Self);
+
+  /** Run-time type information (and related methods). */
+  itkTypeMacro(FFTDiscreteGaussianImageFilter, ImageToImageFilter);
```
dzenanz, line 70:
```suggestion
  itkTypeMacro(FFTDiscreteGaussianImageFilter, DiscreteGaussianImageFilter);
```

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.hxx
```diff
@@ -0,0 +1,208 @@
...
+  }
+  else
+  {
+    // Couldn't crop the region (requested region is outside the largest
+    // possible region).  Throw an exception.
+
+    // store what we tried to request (prior to trying to crop)
+    inputPtr->SetRequestedRegion(inputRequestedRegion);
+
+    // build an exception
```
dzenanz, line 84: `itkExceptionMacro` is a more convenient way to raise an exception, e.g.: `itkExceptionMacro(<< "Requested region " << inputRequestedRegion << " is (at least partially) outside the largest possible region " << inputPtr->GetLargestPossibleRegion());`

Modules/Filtering/Smoothing/test/itkFFTDiscreteGaussianImageFilterTest.cxx
```diff
@@ -0,0 +1,70 @@
...
+#include "itkTestingMacros.h"
+
+int
+itkFFTDiscreteGaussianImageFilterTest(int argc, char * argv[])
+{
+  if (argc != 5)
+  {
+    std::cerr << "Missing parameters." << std::endl;
+    std::cerr << "Usage:" << std::endl;
+    std::cerr << itkNameOfTestExecutableMacro(argv) << " inputFilename sigma kernelWidth outputFilename filterDimensionality" << std::endl;
```
dzenanz: This could be broken up into two lines, e.g. `std::cerr << " inputFilename sigma kernelWidth outputFilename filterDimensionality" << std::endl;`

Modules/Filtering/Smoothing/test/itkFFTDiscreteGaussianImageFilterTest.cxx
```diff
@@ -0,0 +1,70 @@
...
+
+#include "itkFFTDiscreteGaussianImageFilter.h"
+#include "itkImageFileReader.h"
+#include "itkImageFileWriter.h"
+#include "itkTestingMacros.h"
+
+int
+itkFFTDiscreteGaussianImageFilterTest(int argc, char * argv[])
+{
+  if (argc != 5)
```
dzenanz, line 28: this should be 6, not 5 (judging by parameter list below), or at least `argc<5`.

Modules/Filtering/Smoothing/test/itkFFTDiscreteGaussianImageFilterTest.cxx
```diff
@@ -0,0 +1,70 @@
...
+
+  filter->SetInput(reader->GetOutput());
+  filter->SetVariance(sigma);
+  filter->SetMaximumKernelWidth(kernelWidth);
+  filter->SetUseImageSpacing(true);
+  filter->Update();
+
+  using WriterType = itk::ImageFileWriter<ImageType>;
+
+  auto writer = WriterType::New();
```
dzenanz, line 62: Why not use `itk::WriteImage(filter->GetOutput(), argv[4, true);`? We are not trying to demonstrate streaming, or other advanced usage.

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.h
tbirdso: Ah, didn't catch this because the superclass is not actually used in `itkTypeMacro`:
[https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkMacro.h#L446](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkMacro.h#L446)

Can someone remind me of whether there is a reason for keeping the unused `superclass` field in `itkTypeMacro`? I feel like I've seen discussion on this before...

Either way the change looks good.

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.h
```diff
@@ -0,0 +1,141 @@
...
+  using ArrayType = typename Superclass::ArrayType;
+  using SigmaArrayType = typename Superclass::SigmaArrayType;
+  using ScalarRealType = typename Superclass::ScalarRealType;
+
+  /** Complex pixel types for intermediate operations */
+  using ComplexPixelType = Image<std::complex<RealPixelType>, ImageDimension>;
+
+  // Pad input region to kernel image size
+  void
+  GenerateInputRequestedRegion() override;
```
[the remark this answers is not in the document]
tbirdso: [`itk::DiscreteGaussianImageFilter::GenerateInputRequestedRegion`](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/Smoothing/include/itkDiscreteGaussianImageFilter.h#L303) is currently marked `public`, should this also be changed?

dzenanz (approved):

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.h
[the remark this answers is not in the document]
thewtex: yes, please

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.hxx
```diff
@@ -0,0 +1,202 @@
...
+                      << " is (at least partially) outside the largest possible region "
+                      << inputPtr->GetLargestPossibleRegion());
+  }
+}
+
+template <typename TInputImage, typename TOutputImage>
+void
+FFTDiscreteGaussianImageFilter<TInputImage, TOutputImage>::SetMaximumError(const ArrayType arr)
+{
+  itkWarningMacro("FFTDiscreteGaussianImageFilter uses a fixed-width Gaussian kernel and ignores maximum error");
```
tbirdso, line 91: @thewtex @dzenanz Before moving out of WIP I'd like to get your take on handling elements from `DiscreteGaussianImageFilter` in its subclass here.

`DiscreteGaussianImageFilter` defines two variables with setters/getters for constructing a kernel:
- `MaximumError` is the allowable error resulting from kernel truncation
- `MaximumKernelWidth` is the maximum diameter of the Gaussian kernel and overrides `MaximumError` if necessary
These values are used by `itk::GaussianOperator` to construct a Gaussian kernel operator.

`FFTDiscreteGaussianImageFilter`, however, needs to construct a Gaussian kernel _image_ rather than an operator in order to apply FFT operations. `itk::GaussianImageSource` only accounts for image metadata (including size) and does not take kernel error into consideration.

Are the following interface adjustments that I've made here reasonable? If not, do you have suggestions for what could be changed?
- `MaximumError` is ignored and a warning is issued if the user tries to set its value
- `MaximumKernelWidth` is redefined from a loose constraint that _could_ be the resulting kernel size, to a hard constraint that explicitly sets the kernel size.

One alternative would be to revert to using an `itk::GaussianOperator` and then manually transfer operator values to a kernel image, but we would lose benefits of `itk::GaussianImageSource` such as defining the kernel off-centered from a pixel grid.

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.hxx
```diff
@@ -0,0 +1,202 @@
...
+}
+
+template <typename TInputImage, typename TOutputImage>
+void
+FFTDiscreteGaussianImageFilter<TInputImage, TOutputImage>::SetInputBoundaryCondition(
+  const InputBoundaryConditionPointerType)
+{
+  itkWarningMacro("FFTDiscreteGaussianImageFilter ignores InputBoundaryCondition, use RealBoundaryCondition instead");
```
tbirdso, line 99: Same as above -- is it reasonable to ignore superclass interface elements such as `InputBoundaryCondition` if they are not needed for our implementation?

Modules/Filtering/Smoothing/test/itkFFTDiscreteGaussianImageFilterTest.cxx
```diff
@@ -0,0 +1,65 @@
...
+  reader->Update();
+
+  using FilterType = itk::FFTDiscreteGaussianImageFilter<ImageType, ImageType>;
+  auto filter = FilterType::New();
+  ITK_EXERCISE_BASIC_OBJECT_METHODS(filter, FFTDiscreteGaussianImageFilter, DiscreteGaussianImageFilter);
+
+  filter->SetInput(reader->GetOutput());
+  filter->SetVariance(sigma);
+  filter->SetMaximumKernelWidth(kernelWidth);
+  filter->SetUseImageSpacing(true);
```
tbirdso, line 57: `UseImageSpacing` is used to define the Gaussian kernel operator in `DiscreteGaussianImageFilter` but is not used here. I'm not completely clear on what this variable would mean in the context of creating and applying a Gaussian kernel image in FFT convolution.

tbirdso (commented):
@thewtex @dzenanz A few design questions for you before formally moving to the review stage.

Modules/Filtering/Smoothing/test/itkFFTDiscreteGaussianImageFilterTest.cxx
dzenanz: `UseImageSpacing` was added to filters which used to ignore image spacing, to allow both backwards compatible behavior while also allowing correct processing. It can be ignored nowadays, and always correctly apply image spacing.

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.hxx
dzenanz: Of course.

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.hxx
dzenanz: What you are doing is reasonable. Possibly upgrade those warnings into exceptions/errors. This is a new class, and if someone switches to using it they can remove (or comment out) setting those ignored parameters.

tbirdso: Thanks for the reassurance @dzenanz . My thinking behind making these warnings rather than errors is to allow smoothing to succeed if the user registers a factory to override `DiscreteGaussianImageFilter` with `FFTDiscreteGaussianImageFilter`, enabling the use of accelerated hardware. We're still "pretty close" to the original interface, so the user should be warned where parameters are unused but operations should be allowed to proceed.

Modules/Filtering/Smoothing/test/itkFFTDiscreteGaussianImageFilterTest.cxx
tbirdso: Sounds great. I will add another `itkWarningMacro` to notify the user that `UseImageSpacing` is always `true` for `FFTDiscreteGaussianImageFilter`.

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.hxx
thewtex: For an FFT Gaussian filter, `MaximumError` and `MaximumKernelWidth` could be used to determine how much to enlarge the `InputRequestedRegion`.

tbirdso: @thewtex I'm not sure I understand, we're currently using `MaximumKernelWidth` as the fixed size to enlarge the `InputRequestedRegion`. Is the suggestion to go back to using `MaximumError` with a `GaussianOperator` to determine the kernel size?

dzenanz: @thewtex we are already using `MaximumKernelWidth` to enlarge `InputRequestedRegion`: [`1f27fe4` (#3121) R62-R66](https://github.com/InsightSoftwareConsortium/ITK/pull/3121/commits/1f27fe425ed148b63dc71adb1faf148b5b3069a2#diff-35b3faf28f0ff02d401f61ba337a9a8b123e4449c85a21544894d946346d55ddR62-R66).

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.hxx
```diff
@@ -0,0 +1,205 @@
...
+    return;
+  }
+  else
+  {
+    // store what we tried to request (prior to trying to crop)
+    inputPtr->SetRequestedRegion(inputRequestedRegion);
+
+    // Couldn't crop the region (requested region is outside the largest
+    // possible region).  Throw an exception.
+    itkExceptionMacro(<< "Requested region " << inputRequestedRegion
```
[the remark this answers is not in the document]
tbirdso: Ah, I needed to set the requested region on the output image instead. Expanding on current functionality:
- `FFTDiscreteGaussianImageFilter::GenerateInputRequestedRegion()` calls `ImageToImageFilter::GenerateInputRequestedRegion();` which is responsible for propagating the output requested region to the input image;
- [`DataObject`](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/src/itkDataObject.cxx#L352) already handles cases where the output requested region is partially or fully outside the input image's largest possible region;
- `FFTDiscreteGaussianImageFilter::GenerateInputRequestedRegion()` simply pads the existing requested region by the kernel radius.

Given these facts, there are only three possible scenarios:
- The output requested region is partially or fully outside the largest possible region available in the input image. This causes an exception to be thrown in `itk::ImageToImageFilter`.
- The output requested region is exactly the same as the largest possible region. `itk::FFTDiscreteGaussianImageFilter` attempts to pad the input region by the kernel radius, then immediately crops it back to the largest possible region and compensates internally with a boundary condition.
- The output requested region is smaller than the largest possible region. `itk::FFTDiscreteGaussianImageFilter` pads the input region by the kernel radius and crops only as much as necessary to fall inside the largest possible region.

Therefore there is no scenario where the crop here will fail and the exception in line 73 will be thrown. It's impossible to write a test case covering this line.

@dzenanz @thewtex @Leengit  In this case, is it best to remove lines that we can't cover, or keep it _just in case_ some underlying functionality in other classes ever changes?

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.hxx
thewtex: > MaximumKernelWidth is the maximum diameter of the Gaussian kernel and overrides MaximumError if necessary

We should keep this behavior to determine radius to enlarge the InputRequestRegion. A `GaussianOperator` can be used to compute `GetRadius` as in the base class.

tbirdso: Got it, thanks for clarifying @thewtex . Will revise.

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.hxx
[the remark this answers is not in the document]
dzenanz: I vote for removing the exception. The way how the pipeline functions hasn't been changed in ages, and I don't foresee the reason to do that in the future, either.

tbirdso:
@thewtex @dzenanz Ping for re-review

Note that I've consolidated and exposed functionality from `itk::DiscreteGaussianImageFilter` for generating a kernel radius to error specifications with `itk::GaussianOperator` to avoid duplication in `itk::FFTDiscreteGaussianImageFilter`.

`UseImageSpacing` and `MaximumError` are now used for kernel generation.

Modules/Filtering/Smoothing/include/itkDiscreteGaussianImageFilter.h
```diff
@@ -147,6 +152,10 @@ class ITK_TEMPLATE_EXPORT DiscreteGaussianImageFilter : public ImageToImageFilte
   itkSetMacro(RealBoundaryCondition, RealBoundaryConditionPointerType);
   itkGetConstMacro(RealBoundaryCondition, RealBoundaryConditionPointerType);
 
+  /** Get the variance, optionally adjusted for pixel spacing */
+  ArrayType
+  GetImageSpacingVarianceArray() const;
```
tbirdso, line 157: Open to suggestions on a better name. 

This method returns the variance to use if image spacing is taken into consideration for the gaussian kernel. For each element,

` output[i] = variance[i] / ((image spacing [i]) ^2) `

tbirdso:
For looking forward, I've gone ahead and added a factory override mechanism so that developers can opt in to having classes like `itk::MultiResolutionPyramidImageFilter` rely on `FFTDiscreteGaussianImageFilter` rather than `DiscreteGaussianImageFilter` for blurring. We can combine this with GPU-accelerated FFT overrides to speed up blurring with very large images and/or very large kernels.

If more discussion is required around the factory mechanism I'm fine with splitting it off in a separate PR.

Modules/Filtering/Smoothing/include/itkDiscreteGaussianImageFilter.h
thewtex: We may not want to expose this.

Modules/Filtering/Smoothing/test/itkFFTDiscreteGaussianImageFilterFactoryTest.cxx
```diff
@@ -0,0 +1,79 @@
...
+  if (dynamic_cast<OverrideFilterType *>(baseFilter.GetPointer()) != nullptr)
+  {
+    std::cout << "Object factory instantiation succeeded before a factory was registered" << std::endl;
+    return EXIT_FAILURE;
+  }
+
+  // Register factory and verify instantiation override
+  using FactoryType = itk::FFTDiscreteGaussianImageFilterFactory;
+  FactoryType::Pointer overrideFactory = FactoryType::New();
+  itk::ObjectFactoryBase::RegisterFactory(overrideFactory);
```
thewtex, line 49: Cool!!

Modules/Filtering/Smoothing/include/itkDiscreteGaussianImageFilter.h
thewtex: As a `public` method at least.

thewtex (commented):
@tbirdso looking awesome! :sparkler: 

Tests should also be added to verify that we get the same result with the DiscreteGaussianImageFilter and the FFTDiscreteGaussianImageFilter.

dzenanz (commented):
Mostly looks good, some minor suggestions.

Modules/Filtering/Smoothing/include/itkDiscreteGaussianImageFilter.hxx
```diff
@@ -26,6 +26,61 @@
...
+typename DiscreteGaussianImageFilter<TInputImage, TOutputImage>::ArrayType
+DiscreteGaussianImageFilter<TInputImage, TOutputImage>::GetImageSpacingVarianceArray() const
+{
+  if (m_UseImageSpacing)
+  {
+    ArrayType pixelVariance;
+    // Adjusted variance = var / (spacing ^ 2)
+    for (unsigned int dim = 0; dim < ImageDimension; ++dim)
+    {
+      if (this->GetInput() == nullptr || this->GetInput()->GetSpacing()[dim] == 0.0)
```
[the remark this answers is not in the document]
tbirdso: Great catch. I've refactored so that the only condition for throwing an exception is that the input image has not been set.

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.hxx
```diff
@@ -0,0 +1,173 @@
...
+  auto           inputOrigin = localInput->GetOrigin();
+  for (size_t dim = 0; dim < ImageDimension; ++dim)
+  {
+    double radius = (kernelSize[dim] - 1) / 2;
+    mean[dim] = inputSpacing[dim] * radius + inputOrigin[dim]; // center pixel pos
+  }
+  kernelSource->SetMean(mean);
+  kernelSource->SetSigma(this->GetSigmaArray());
+
+  // Estimate kernel generation to be roughly 5% of effort
```
dzenanz, line 143: Change the comment

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.hxx
```diff
@@ -0,0 +1,172 @@
...
+  auto           inputOrigin = localInput->GetOrigin();
+  for (size_t dim = 0; dim < ImageDimension; ++dim)
+  {
+    double radius = (kernelSize[dim] - 1) / 2;
+    mean[dim] = inputSpacing[dim] * radius + inputOrigin[dim]; // center pixel pos
+  }
+  kernelSource->SetMean(mean);
+  kernelSource->SetSigma(this->GetSigmaArray());
+
+  progress->RegisterInternalFilter(kernelSource, 1.0f / numberOfStages);
```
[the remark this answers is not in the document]
tbirdso: Updated

Modules/Filtering/Smoothing/test/itkDiscreteGaussianImageFilterTest2.cxx
```diff
@@ -36,7 +42,9 @@ itkDiscreteGaussianImageFilterTestA(const char * inputFilename, float sigma, con
 
   auto filter = FilterType::New();
   filter->SetInput(reader->GetOutput());
-  filter->SetVariance(sigma);
+  filter->SetSigma(sigma);
+  filter->SetMaximumError(kernelError);
+  filter->SetMaximumKernelWidth(kernelWidth);
   filter->Update();
 
   using WriterType = itk::ImageFileWriter<ImageType>;
```
dzenanz, line 50: using WriteImage would reduce amount of code.

tbirdso:
@thewtex 's suggestion
> Tests should also be added to verify that we get the same result with the DiscreteGaussianImageFilter and the FFTDiscreteGaussianImageFilter.

was well-founded, because on more testing I am getting different images for `DiscreteGaussianImageFilter` and `FFTDiscreteGaussianImageFilter`.

`DiscreteGaussianImageFilterComparisonTest`:
discrete-blur-output

`FFTDiscreteGaussianImageFilterTest`:
fft-discrete-blur-output

Diff:
diff-output

It looks like the main issue here is that each filter relies on a different kernel source with what appears to be fundamentally different mechanisms for Gaussian kernel generation, leading to different kernel values from the same parameters.

`DiscreteGaussianImageFilter` creates and applies iterative 1D `itk::GaussianOperator`s which [rely on Bessel functions](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkGaussianOperator.hxx#L38) for defining kernel values. `FFTDiscreteGaussianImageFilter` requires an ND kernel image that it attempts to generate through `itk::GaussianImageSource` based on the kernel radii from `itk::GaussianOperator`s constructed to the user error specifications, however the image source relies on [`itk::GaussianSpatialFunction`](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkGaussianSpatialFunction.hxx#L36) to generate kernel elements which seems to produce very different values from `itk::GaussianOperator`.

@thewtex @dzenanz could you take a look and weigh in on this? Two potential options that I see for moving forward:
- Dig deeper into the differences in the `itk::GaussianOperator` and `itk::GaussianSpatialFunction` procedures to understand how parameters may need to be mapped between the two for similar outputs;
- Successively generate 1D `itk::GaussianOperators`, copy to `itk::VariableSizeMatrix` objects, cross-multiply to produce an ND kernel, iteratively copy elements to a kernel image, and update image metadata to match the input image. This would increase prep time as we may need to run 3- or 4-D matrix multiplication to get "large" kernels. However I would favor this approach as it retains a dependency on `itk::GaussianOperator` only and time for kernel generation was previously anticipated to be small compared to convolution.

tbirdso:
Thanks @thewtex . Do you know if we multiply operators to get a kernel anywhere else in ITK? I'm having trouble finding examples of how to do this efficiently.

thewtex:
@tbirdso nothing comes to bind offhand, but I agree -- taking a look to re-use any exsting functionality makes sense, but we can also use a manual implementation.

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.hxx
tbirdso: ✔️

Modules/Filtering/Smoothing/test/itkDiscreteGaussianImageFilterTest2.cxx
tbirdso: If possible I'd rather restrict changes to `itk::DiscreteGaussianImageFilter` and its test coverage to strictly necessary changes for this PR, such as exposing test parameters so that we can compare output more easily. We can consolidate test read/writes in a later PR.

tbirdso:
@dzenanz @thewtex Ping for re-review.

New changes:
- Can use `FFTDiscreteGaussianImageFilterEnums` to specify how `FFTDiscreteGaussianImageFilter` should generate its kernel image for FFT convolution
- If the kernel source is set to `FFTDiscreteGaussianImageFilterEnums::KernelSource::COMBINED_OPERATORS` then the filter will generate an empty image and manually perform vector multiplication to compose an ND Gaussian kernel image from N 1D directional kernels
- Tests are added to exercise different kernel methods and to compare output to `DiscreteGaussianImageFilter` when the kernel source is `COMBINED_OPERATORS`

thewtex (changes requested):

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.h
```diff
@@ -0,0 +1,178 @@
...
+   * results to the DiscreteGaussianImageFilter class.
+   *
+   * Generating an ND kernel image directly with
+   * GaussianImageSource relies on a slightly different
+   * spatial function and will produce different smoothing,
+   * but kernel generation will be faster.
+   */
+  enum class KernelSource : uint8_t
+  {
+    COMBINED_OPERATORS = 0,
```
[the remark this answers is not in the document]
tbirdso: Changed

Modules/Filtering/Smoothing/test/CMakeLists.txt
```diff
@@ -59,6 +64,45 @@ itk_add_test(NAME itkDiscreteGaussianImageFilterTest1a
...
+        ${ITK_TEST_OUTPUT_DIR}/FFTDiscreteGaussianImageFilterTestImageSourceOutput.mha
+        3.5 0.05 35 2 1)
+itk_add_test(NAME itkFFTDiscreteGaussianImageFilterKernelDimensionalityTest
+      COMMAND ITKSmoothingTestDriver
+      --compare DATA{Baseline/FFTDiscreteGaussianImageFilterTestKernelDimensionalityOutput.mha}
+                ${ITK_TEST_OUTPUT_DIR}/FFTDiscreteGaussianImageFilterTestKernelDimensionalityOutput.mha
+      itkFFTDiscreteGaussianImageFilterTest
+        DATA{${ITK_DATA_ROOT}/Input/2th_cthead1.png}
+        ${ITK_TEST_OUTPUT_DIR}/FFTDiscreteGaussianImageFilterTestKernelDimensionalityOutput.mha
+        1.2 0.25 11 1 0)
```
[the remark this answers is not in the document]
tbirdso: Expanded test coverage with 3D images and non-unit spacing + origin in most recent commit.

Modules/Filtering/Smoothing/test/itkDiscreteGaussianImageFilterTest2.cxx
```diff
@@ -53,30 +61,33 @@ itkDiscreteGaussianImageFilterTestA(const char * inputFilename, float sigma, con
...
   }
 
+  unsigned int vec_dim = std::stoi(argv[2]);
   if (vec_dim != 1 && vec_dim != 3)
   {
     std::cerr << "This test only supports 3-channel image or 1-channel image for demo! Exiting ... " << std::endl;
     return EXIT_FAILURE;
   }
 
+  float        sigma = (argc >= 6) ? std::stof(argv[5]) : 0.0;
```
dzenanz, line 89: I find it easier to spot inconsistencies if the same number appears in the condition and the index, e.g. `argc > 5 ... argv[5]`.

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.h
```diff
@@ -151,7 +155,12 @@ class ITK_TEMPLATE_EXPORT FFTDiscreteGaussianImageFilter : public DiscreteGaussi
   itkGetConstMacro(KernelSource, FFTDiscreteGaussianImageFilterEnums::KernelSource);
 
 protected:
-  FFTDiscreteGaussianImageFilter() = default;
+  FFTDiscreteGaussianImageFilter()
+  {
```
dzenanz, line 159: Do we need the body? I think both of these assignments can be made at the point of declaration.

tbirdso: At a glance I can't find any other examples of initializing filter members at the point of declaration, it seems to be standard practice to initialize in the constructor. Without distinct benefit I'd prefer to stick to the standard approach.

Modules/Filtering/Smoothing/test/itkDiscreteGaussianImageFilterTest2.cxx
tbirdso: Fair enough, will fix.

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.h
dzenanz: It was [not possible before C++11](https://stackoverflow.com/questions/24149924/has-the-new-c11-member-initialization-feature-at-declaration-made-initializati), and not all classes have been converted to the new syntax. But initialization at declaration is the preferred style, and should be used in new code.

tbirdso: 👍 Thanks for the link and background, it seems like there is benefit especially in the context of overridden constructors. I will make the change.

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.h
```diff
@@ -0,0 +1,189 @@
...
+  /** Typedef for convolution */
+  using ConvolutionImageFilterType = FFTConvolutionImageFilter<RealImageType, RealImageType, OutputImageType>;
+
+  /** Overridden accessors for unused parameters */
+
+  void
+  SetInputBoundaryCondition(const InputBoundaryConditionPointerType) override;
+
+  auto
+  GenerateKernelImage() -> RealImagePointerType;
```
thewtex, line 151: API should use raw pointers instead of smart pointers, primarily for wrapping simplification, ownership rules.

thewtex (changes requested):
Beautiful!

One minor issue noted inline.

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.h
tbirdso: Changes:
- Added `m_KernelImage` member so that the filter object has ownership of the kernel image, rather than previous method of passing ownership out of `GenerateKernelImage`;
- `GenerateKernelImage` is now a protected method and passes out a raw pointer reference to `m_KernelImage` after it has been initialized

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.h
```diff
@@ -49,7 +50,7 @@ class FFTDiscreteGaussianImageFilterEnums
    */
   enum class KernelSource : uint8_t
   {
-    COMBINED_OPERATORS = 0,
+    OPERATORS = 0,
```
dzenanz, line 53: Changes which are not related to increase of test coverage could be squashed into the previous commit.

Modules/Filtering/Smoothing/test/itkDiscreteGaussianImageFilterTest2.cxx
```diff
@@ -84,23 +86,33 @@ itkDiscreteGaussianImageFilterTest2(int argc, char * argv[])
     return EXIT_FAILURE;
   }
 
-  float        sigma = (argc >= 6) ? std::stof(argv[5]) : 0.0;
-  float        kernelError = (argc >= 7) ? std::stof(argv[6]) : 0.01;
-  unsigned int kernelWidth = (argc >= 8) ? static_cast<unsigned int>(std::stoi(argv[7])) : 32;
+  float        sigma = (argc > 5) ? std::stof(argv[5]) : 0.0;
```
dzenanz, line 89: Changes which are not related to increase of test coverage could be squashed into the previous commit.

Modules/Filtering/Smoothing/test/itkFFTDiscreteGaussianImageFilterTest.cxx
```diff
@@ -24,30 +24,17 @@
...
-    std::cerr << itkNameOfTestExecutableMacro(argv)
-              << " inputFilename outputFilename sigma kernelError kernelWidth filterDimensionality kernelSource"
-              << std::endl;
-    return EXIT_FAILURE;
-  }
-
-  using ScalarPixelType = float;
-  const size_t ImageDimension = 2;
-  using ImageType = itk::Image<ScalarPixelType, ImageDimension>;
+  float        sigma = (argc > 4) ? std::stof(argv[4]) : 0.0;
```
dzenanz, line 31: Changes which are not related to increase of test coverage could be squashed into the previous commit.

dzenanz (approved):
Looks good, just a few nitpicks.

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.h
tbirdso: Is it preferable to try to keep separate commits or reasonable to simply squash all into one? I tried to keep things separated but with fixups the distinctions between operational/coverage fixes have become blurry.

thewtex (approved):
LGTM!

Modules/Filtering/Smoothing/include/itkFFTDiscreteGaussianImageFilter.h
dzenanz: This is good as-is. While I would have preferred these few changes moved to the previous commit and left most of the changes in this commit as a separate commit, I assume you are already quite tired of this PR and it is fine to merge it if CI is green.

tbirdso: > I assume you are already quite tired of this PR and it is fine to merge it if CI is green

😃 

Thanks @dzenanz , if CI fails I will look at refactoring commits along with any necessary fixes, otherwise I'll merge as-is.

tbirdso: CI has passed, I will go ahead with the merge.

PR #3122, opened 2022-01-18
Update nullptr usage for c++11

hjmjohnson:
Minor changes to prefer to use nullptr


seanm:
Can I get rid of `-Wno-zero-as-null-pointer-constant` on my bots after this?

hjmjohnson:
> Can I get rid of `-Wno-zero-as-null-pointer-constant` on my bots after this?

I would hope so.

dzenanz (approved):

seanm:
I've removed `-Wno-zero-as-null-pointer-constant` on my TSan bot... let's see tomorrow...

seanm:
@hjmjohnson look like there are a number of such warnings: https://open.cdash.org/viewBuildError.php?type=1&buildid=7689549  want to tackle them?

seanm:
Actually, it looks like just two: itkThreadSupport.h:52 and itkThreadSupport.h:57.  I don't understand the comment about `NULL` being required for the second... it's C++ not C...

PR #3123, opened 2022-01-18
BUG: Protect itk::OutputWindow cerr usage with a mutex

seanm:
TSan otherwise complained when multiple threads use OutputWindow::DisplayText at the same time, such as in the itkSLICImageFilterTest3 test. This I don't really understand though, since C++11 cerr usage is supposed to be thread safe. But even if strictly thread-safe, output can get interleaved, and a mutex solves both that and the TSan complaint.


seanm:
This will probably be my last TSan PR.  I've been through every failure and many of them are the same root cause.  There are a few more issues I'll create though...

dzenanz (approved):

PR #3126, opened 2022-01-18
BUG: Do not let HDF5 set the CMAKE_INSTALL_PREFIX

thewtex:
Submitted upstream: https://github.com/HDFGroup/hdf5/pull/1380

### DIFF
--- a/Modules/ThirdParty/HDF5/src/itkhdf5/config/cmake_ext_mod/HDFMacros.cmake
+++ b/Modules/ThirdParty/HDF5/src/itkhdf5/config/cmake_ext_mod/HDFMacros.cmake
@@ -434,7 +434,9 @@ macro (HDF_DIR_PATHS package_prefix)
     endif ()
   endif ()
 
-  if (CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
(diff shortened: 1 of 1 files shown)


Modules/ThirdParty/HDF5/src/itkhdf5/config/cmake_ext_mod/HDFMacros.cmake
```diff
@@ -434,7 +434,7 @@ macro (HDF_DIR_PATHS package_prefix)
     endif ()
   endif ()
 
-  if (CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+  if (NOT ${package_prefix}_EXTERNALLY_CONFIGURED AND CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
```
hjmjohnson, line 437: Could you add 
```
# ITK --start. Do not let HDF5 set the CMAKE_INSTALL_PREFIX for externally configured projects.

# ITK --stop
```
Comments before and after this line so that we know it is a deviation from the upstream for ITK.  That makes resolving merge conflicts much much easier to manage in the future.

thewtex: Done.

Upstream should have this change, soon.

dzenanz (approved):

hjmjohnson (approved):

PR #3128, opened 2022-01-19
WIP: BUG: Install .pyi submodule files

thewtex:
Also avoid clobbering ITK_STUB_DIR, which is set in the top level
CMakeLists.txt

### DIFF
--- a/Wrapping/Generators/Python/CMakeLists.txt
+++ b/Wrapping/Generators/Python/CMakeLists.txt
@@ -289,6 +289,13 @@ macro(itk_setup_swig_python type base_name interface_file python_file cpp_file d
   )
 
   WRAP_ITK_PYTHON_BINDINGS_INSTALL(/itk "${WRAPPER_LIBRARY_NAME}" "${python_file}")
+  get_filename_component(python_file_basename "${python_file}" NAME)
+  # Each wrapped submodule has a Python stub file, except for pyBase
+  if(NOT "${WRAPPER_LIBRARY_NAME}" STREQUAL "ITKPyBase" AND NOT "${WRAPPER_LIBRARY_NAME}Python.py" STREQUAL "${python_file_basename}")
+    string(REPLACE "Python.py" ".pyi" python_stub_file "${python_file_basename}")
+    set(python_stub_file "${ITK_STUB_DIR}/${python_stub_file}")
+    WRAP_ITK_PYTHON_BINDINGS_INSTALL(/itk "${WRAPPER_LIBRARY_NAME}" "${python_stub_file}")
+  endif()
 
 endmacro()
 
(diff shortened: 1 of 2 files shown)


thewtex:
@hjmjohnson @kian-weimer this would add the .pyi files to the Python packages, but are they ready? I noticed that it looks like generation of an *itk/__init__.py* was intended, but it is not created. Looking at the contents of the submodule .pyi files, they create classes that inherit from bases classes, but their base classes are not imported -- does that still work? Many of the methods have a `""""""` empty documentation string -- is this a known issue?

kian-weimer:
@thewtex 

I'll take a look at your first two questions in a bit, I'm currently rebuilding ITK to look for the issues. 

The expected behavior is that there should be an __init__.pyi file generated within the itk-stubs directory that imports all of the other files. I'm not seeing it appear right away during the build so that may be a concern. 

I'm pretty sure that it does not matter if the base classes are imported in each file since they are all imported centrally in the __init__.pyi file. This may be a product of the static analyzer (and might not work if the code were more than type hints). When my build is finished I'll take a closer look to make sure. 

The empty doc strings are a known issue. There were some issues in finding a good method of pulling the doc strings into the methods before the first version of the wrapping was released. Currently there is no code to do this and the doc strings just serve as placeholders for the future.

kian-weimer:
It looks like the function call that generates the `__init__.pyi` file was removed in [this commit](https://github.com/InsightSoftwareConsortium/ITK/commit/4f33aab02ed37259cd2611dc0bb02b74f4c41154#diff-9450e80f05a976e103d0fe314f3219d0c99addeef090c9b456ac66f0eed8ae48). Without the call the `__init__.pyi` file is never generated which appears to prevent the type hinting from working. 

@hjmjohnson Was this on purpose?

hjmjohnson:
@kian-weimer   It does seem to have been a mistake.  I have not cleaned my build directory in a long time, so I do have the __init__.pyi file.

Does adding this call back in fix the problem?

thewtex:
I think we need to move the generation of `__init__.pyi` out of *igenerator.py*. Unlike suggested by the comment there, `igenerator.py` is executed in parallel, and all of the `*.pyi` files will not be generated at that time.

kian-weimer:
@thewtex I agree it would be better if the file generation was moved to a different location. In the previous implementation the entire file was recreated after each run of `igenerator.py` in the hopes that the final iteration would create the correct version. But that's not ideal. Do you have any suggestions for where the init generation step could be moved to?

kian-weimer:
@hjmjohnson I believe that adding the call will fix the problem (at least most of the time), but I'll rebuild again to be sure.

thewtex:
The `igenerator.py` calls running in parallel are going to possible clobber each other's output, and they may have the wrong output.

Instead, try accumulating the file's content in a CMake variable / global property, then use CMake `configure_file`.

hjmjohnson:
@thewtex   I think this change is no longer relevant after the massive wrapping modifications Kian and I did recently.  I am closing.  If I am in error, please reopen.

PR #3129, opened 2022-01-20
COMP: added override keyword on a dtor

seanm:
I have a patch pending for upstream HDF5 too.

### DIFF
--- a/Modules/ThirdParty/HDF5/src/itkhdf5/c++/src/H5Location.h
+++ b/Modules/ThirdParty/HDF5/src/itkhdf5/c++/src/H5Location.h
@@ -334,7 +334,7 @@ class H5_DLLCPP H5Location : public IdComponent {
 #endif // DOXYGEN_SHOULD_SKIP_THIS
 
     // Noop destructor.
-    virtual ~H5Location();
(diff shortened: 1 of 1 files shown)


dzenanz (approved):

N-Dekker:
Fine, thanks @seanm  For ITK, we did clang-tidy [modernize-use-override](https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-override.html), which meant that whenever `override` was added, the `virtual` keyword would be removed from that member function declaration. Did you consider removing `virtual` in this case as well?

seanm:
>Did you consider removing virtual in this case as well?

That is not the HDF5 convention.

thewtex (approved):

N-Dekker:
> > Did you consider removing virtual in this case as well?
> 
> That is not the HDF5 convention.

OK, I see. Thanks!

N-Dekker (approved):
Thanks Sean! I would suggest a more specific commit subject line (as long as the 72 chars limit allows us), for example: 

    COMP: Add override specifier to the HDF5 `H5Location` destructor

But it's OK to me anyway.

PR #3139, opened 2022-01-24
STYLE: Change unsigned to unsigned int.

Leengit:
This pull request was generated with:
```bash
cd ITK
fgrep -Rl 'unsigned' * | egrep '\.(h|c)(xx|pp|)$' | fgrep -v ThirdParty |
  xargs sed -i -r \
    -e 's/unsigned([^A-Za-z_])/unsigned int\1/g' \
    -e 's/unsigned int (char|short|long|int)/unsigned \1/g' \
    -e 's/is_unsigned int/is_unsigned/g' \
    -e 's/unsigned int\(/(unsigned int)(/g' \
    -e 's/maximum unsigned int value/maximum unsigned value/g'
git restore \
  Examples/RegistrationITKv4/ImageRegistration1.cxx \
  Modules/Core/Common/include/itkIntTypes.h \
  Modules/Core/Common/include/itkMacro.h \
  Modules/Core/Common/include/itkMath.h \
  Modules/Core/Common/include/itkNeighborhoodAlgorithm.hxx \
  Modules/IO/MRC/include/itkMRCHeaderObject.h \
  Modules/IO/MRC/src/itkMRCImageIO.cxx \
  Modules/IO/TIFF/src/itkTIFFImageIO.cxx \
  Modules/Numerics/Statistics/include/itkNormalVariateGenerator.h \
  Modules/Registration/Common/test/RegistrationITKv3/ImageRegistration1.cxx \
  Modules/Video/BridgeOpenCV/include/itkOpenCVVideoCapture.hxx
git diff --name-only | xargs clang-format-Linux -i
```

## PR Checklist


dzenanz (commented):
This style requires more typing. I prefer the opposite change: replacing `unsigned int` by `unsigned`. What do the others think? @jhlegarreta @blowekamp @hjmjohnson @thewtex

N-Dekker:
> This style requires more typing. I prefer the opposite change: replacing unsigned int by unsigned.

Thank you @Leengit, but like @dzenanz I also prefer the opposite change! Just `unsigned` is clear enough to me. We don't write `signed int` either, do we?

The use of  `unsigned int` is cumbersome when trying to do type conversion by `T{value}` syntax, for example:

    x = unsigned{ value }; // OK
    y = unsigned int{ value }; // Error!

Leengit:
In case it makes a difference: without this pull request there are 38632 lines of code that contain `unsigned` and, of those, there are 31815 lines of code that have `char`, `int`, `short`, or `long` immediately afterwards.

- So, by subtraction, there are 6817 lines of code that do not have one of those four types after `unsigned`.
- In contrast, there are 23125 lines that contain `unsigned int`.

Leengit:
> This style requires more typing. I prefer the opposite change: replacing `unsigned int` by `unsigned`.

The present pull request changes 224 files.  The opposite change would change 3437 files.

Leengit:
> We don't write `signed int` either, do we?

Much as we could shorten `unsigned int` to `unsigned`, we could shorten `signed int` to `signed`, but I'd recommend against it.  Yes, I know that's not what you meant.  But it was fun to check that my [compiler would accept](https://en.cppreference.com/w/cpp/language/types) `signed a = 0;`.

It is a matter of preference, of course.  I write my integer types as
(`static`)? (`const`|`constexpr`)? (`unsigned`)? (`char`|`short`|`int`|`long`|`long long`)
where the three leading parts are optional, but the last part is not.  In particular, I keep the `int` in `unsigned int` but I leave out the `int` in `short int`.

hjmjohnson:
I prefer fixed-width types https://en.cppreference.com/w/cpp/types/integer over fundamental types https://en.cppreference.com/w/cpp/language/types whenever possible.


In fundamental types, I prefer "signed int" and "unsigned int" primarily because I work a lot with new programmers, and they need every extra hint they can get.   I am routinely shocked by how many senior college CS students are unable to differentiate between different numerical data types.

Leengit:
> I prefer fixed-width types https://en.cppreference.com/w/cpp/types/integer over fundamental types https://en.cppreference.com/w/cpp/language/types whenever possible.

Agreed!  Dealing with platform specific `long` is a drag.

For the folks who like`signed int` better than `int`, it would be nice if there were `sint32_t`, etc. equivalent to `int32_t`, etc.  For the folks who like `unsigned` better than `unsigned int`, it would be nice if there were an `uns32_t`, etc.

N-Dekker:
By the way, I like `std::intmax_t` and `std::uintmax_t` very much: just get me the largest available integer type available, don't want to think about the number of bits!

Leengit:
So, where are we with this?  While I can see the appeal of both `unsigned` and `signed int`, the majority of our code uses `unsigned int` and `int`, respectively.  This pull request takes the 10% minority that is `unsigned` and makes it match the 90% majority that is `unsigned int`.

Leengit:
I just force-pushed to rebase to the current master branch.

seanm:
Haven't followed this discussion closely, but my own coding style is to use `int` and `unsigned int`.

blowekamp:
I am agnostic on this change/preference like our automatic tools to enforce such a guideline such as clang-tidy for KWStyle.

hjmjohnson (approved):

PR #3140, opened 2022-01-25
DOC: warning for calling GetImageViewFromArray on non-contiguous array

HastingsGreer:
There's a bit of a foot-gun in GetImageViewFromArray, where if you pass it a non-contiguous array, it makes a contiguous copy of that array and returns an image view of that copy. The copy is then stored in the .base member of the returned image, to keep it from being garbage collected. This behavior is used by GetImageFromArray, which should be able to get images from non-contiguous arrays. 

However, the .base member can be lost if the returned image is passed to, for example, a SetTransform method in itk, and then no longer referenced from python. This can cause surprise segfaults.

This pull request adds a warning to GetImageViewFromArray that is displayed when it is not actually returning a view, which can be suppressed with a keyword argument. 

Another possibility would be to move the handling of non-contiguous arrays to GetImageFromArray, and throw an error if a non-contiguous array is passed to GetImageViewFromArray. That would be a cleaner solution, but a breaking change.

### DIFF
--- a/Modules/Bridge/NumPy/wrapping/PyBuffer.i.in
+++ b/Modules/Bridge/NumPy/wrapping/PyBuffer.i.in
@@ -72,6 +72,10 @@
         inverts the indexing scheme, the Image representation will be the
         same for an array and its transpose. If flipping is desired, see
         *np.reshape*.
+
+        By default, a warning is issued if this function is called on a non-contiguous
+        array, since a copy is performed and care must be taken to keep a reference
+        to the copied array. This warning can be suppressed with need_contiguous=False
         """
 
         assert ndarr.ndim in (1, 2, 3, 4, 5), \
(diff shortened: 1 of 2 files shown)


thewtex:
@HastingsGreer thanks!! I like your approach.

Could a note in the docstring please be added?, perhaps *need_contiguous=False can be passed to suppress the warning that indicates a reference must be kept to the result when a non-contiguous array is passed.*

thewtex (commented):
:medal_sports: 

Sorry I missed this the first time, but we will want to propagate the new argument and doc up to:

https://github.com/InsightSoftwareConsortium/ITK/blob/e2c206afa8cfc441f948ecaea2b23591f284ce8d/Wrapping/Generators/Python/itk/support/extras.py#L401-L408

so it hits `help(itk.image_view_from_array)` and the `need_contiguous` propagates down into this interface.

dzenanz:
I guess this is not ready for merging yet?

hjmjohnson:
@HastingsGreer  Would you mind summarizing what work needs to be done for thie PR to be rejected/accepted?

dzenanz:
@HastingsGreer ping!

thewtex:
I rebased on`master` and added a commit to propagate the arg and documentation into extras.py.

thewtex (approved):

PR #3145, opened 2022-01-25
WIP: BUG: Fixed __init__.pyi generation and import issues for python stub files.

kian-weimer:
There are a few problems remaining with the Python Stub interface (.pyi) file generation.

1. `__init__.pyi` generation
The function call that generated this file was removed in an earlier commit.
Instead of recreating the old behavior, the `__init__.pyi` file generation was
moved into CMake configuration files. The previous implementation was unreliable
in parallel builds and would often be missing a few imports. The new implementation
should work better in a parallel build environment.

2. Missing imports for dependent .pyi files. 
In the original implementation Python hints were contained in just two files. To resolve parallelization issues the .pyi files were split into many separate files that were generated in parallel and then imported into a main `__init__.pyi` file. Some of these files are dependent on each other and some form of importing needs to be added. The solution may not be as simple as adding import statements due to the following:
   - Classes pretended with `_` cannot be imported in the static context. While this could work in an actual program, the IDE will not produce the type hints due to the underscore. Removing the underscore will reveal several 'fake' classes to the user. 
   - The import names may not be known. While some classes such `Array` are contained in a file with their same name`itkArray.pyi` others, like `Object`, are in files with other names (`ITKCommonBase.pyi`). It may be difficult to learn the correct file name to import during the execution of `igenerator.py` without maintaining a list.  

The final solution resolved these issues by splitting the Template and Proxy classes into two separate files, allowing the leading underscores to be removed from the proxies (Which is necessary for * imports to work). The `_proxies.pyi` then imports all proxy files making all Proxy classes available to other Template and Proxy classes through the same interface: `from ._proxies import <class_name>Proxy as _<class_name>Proxy`. This method allows for the imports to function as desired while not requiring the file name of each class to be known and without exposing the proxy classes to the user. 

These issues were mentioned in #3128.

### DIFF
--- a/Wrapping/Generators/SwigInterface/igenerator.py
+++ b/Wrapping/Generators/SwigInterface/igenerator.py
@@ -1752,45 +1752,6 @@ def write_class_pyi(
         pyiFile.write(interfaces_code)
 
 
-def write_common_init_interface_file(interface_dir_path: Path) -> None:
-    # This code may still be problematic due to possible parallel runs overwriting
-    # the file each time.  As long as the last write is successful, this code
-    # works.  Initial testing demonstrated that igenerator.py was never run
-    # in parallel, so that also makes this OK.
-    with open(interface_dir_path / "__init__.pyi", "w") as pyiIndexFile:
-        # This is a bit of a hack to re-write this file every time
-        # a new .pyi file is written.
-        # Keeping track of these generated files from CMake is
-        # very convoluted, and the cmake custom command dependencies
-        # caused circular dependencies.
-        #
-        # This __init__.pyi file needs a small preamble
-        # and then import all the auto-generated .pyi files
-        # statically.
-        preamble = """
-# Create/Clear .pyi file and add boilerplate import content
-# Python Header Interface File
-# imports as they appear in __init__.py
-from typing import Union, Any
-from itkConfig import ITK_GLOBAL_VERSION_STRING as __version__Ï
-
-## Must import using relative paths so that symbols are imported successfully
-from .support.extras import *
-from .support.types import *
-
-### All items below are written out statically during
-
-"""
-        pyiIndexFile.write(preamble)
-
-        # If we could keep track of the interface files in CMake, we could avoid
-        # rewriting the __init__.pyi file every time a new version is created.
-        for interface_file in Path(f"{interface_dir_path}").glob(f"i*.pyi"):
-            pyiIndexFile.write(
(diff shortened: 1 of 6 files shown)


thewtex (commented):
Looking good :+1:

kian-weimer:
Issue 2 in the description above should now be resolved with the recent commit. However, while finding this solution I've identified another issue that will need to be fixed: 

3. When designing this system, I mistakenly assumed that all instances of a class would be defined within one submodule and therefore the wrapping process for a given class would happen in just one run of `igenerator.py`. However, in some cases a class has a new overload defined for it within a different submodule. For example, fixed array is wrapped in both `itkShapeLabelObject.wrap` and `itkFixedArray.wrap` which means that it appears in two separate, potentially parallel runs of `igenerator.py`. This causes The class to appear in two separate .pyi files, thus preventing some of the class overloads to appear properly in the type hints. 

The solution to this issue may be rather complex, as the current code heavily relies on the (now disproved) fact that each file can be generated in just one run of `igenerator.py`. 

I'm now looking into solutions for this, which may include a slight restructuring of the `.wrap` files if it proves to be possible.

thewtex (changes requested):
@kian-weimer thank you for the fix!

A few clean-up issues noted inline.

Wrapping/Generators/SwigInterface/igenerator.py
```diff
@@ -2024,14 +2039,23 @@ def generate_swig_input(submoduleName, classes):
...
         )
 
-        write_class_pyi(
-            Path(f"{options.pyi_dir}/{classes[itk_class].submodule_name}.pyi"),
+        # write_class_pyi(
+        #     Path(f"{options.pyi_dir}/{classes[itk_class].submodule_name}.pyi"),
+        #     itk_class,
+        #     outputPYIHeaderFile.getvalue(),
+        #     outputPYIMethodFile.getvalue(),
+        # )
```
thewtex, line 2047: Remove commented code

thewtex (changes requested):

Wrapping/Generators/SwigInterface/igenerator.py
kian-weimer: Sorry! Just noticed that as well. The commits should be fixed.

thewtex (approved):

thewtex (approved):

thewtex:
Yes, please address the other issues in additional PR's.

PR #3146, opened 2022-01-26
STYLE: Replace `NumericTraits<double>::ZeroValue()` and `NumericTraits<double>::OneValue()` calls by `0.0` and `1.0`, respectively

N-Dekker:
Aims to improve code readability (assuming that most programmers know that the type of the floating point literals `0.0` and `1.0` is `double`, according to the C++ Standard).

### DIFF
--- a/Modules/Core/Common/include/itkVersor.hxx
+++ b/Modules/Core/Common/include/itkVersor.hxx
@@ -444,7 +444,7 @@ Versor<T>::Set(const VectorType & axis)
     throw exception;
   }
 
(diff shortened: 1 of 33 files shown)


seanm:
I like!

dzenanz (approved):
There are surprisingly many instances of these. How about doing the analogous thing for float, by using `0.0f` and `1.0f`?

dzenanz (approved):

hjmjohnson (approved):

PR #2970, opened 2021-12-16
COMP: Use itk.ULL in Windows Platform

PranjalSahu:
Fix for Python test in Windows platform.
itk.ULL is used in windows while itk.UL is used in Ubuntu and MacOS. Refer: [WrapBasicTypes.cmake](https://github.com/InsightSoftwareConsortium/ITK/blob/047fa6934ac32711d1d7b84fa6eb81c72db01960/Wrapping/WrapBasicTypes.cmake#L190)

In another test, a message is printed if the required wrapping is un-available, Refer: [itkPyVectorContainerTest](https://github.com/InsightSoftwareConsortium/ITK/blob/895fc718e814527aa156f7ff63ca8d5e15761d26/Modules/Bridge/NumPy/wrapping/test/itkPyVectorContainerTest.py#L44)

This is a follow-up to #2953 and #2957.

## PR Checklist
- [X] Added test (or behavior not changed)

PR #3149, opened 2022-01-31
WIP: Linux support for vector_container_from_array

PranjalSahu:
Currently, **itk.vector_container_from_array** does not work in Ubuntu due to the absence of itk.ULL key in 
the list of PyVectorContainer wrapping. It throws the error message
**"No suitable template parameter can be found"**

The correct key would be itk.UL.

Also, the test is not performed in Linux for itkPyVectorContainerTest
where we give the message "Insufficient wrapping to perform itkPyVectorContainerTest" and just skip the test.
Also refer https://github.com/InsightSoftwareConsortium/ITK/pull/2970.

Are there any issues with the approach in this PR that I am not aware of?

Opening for getting comments as we need this support for mesh serialization.


thewtex (approved):
:1st_place_medal:

dzenanz (approved):

PR #108, opened 2018-11-05
STYLE: Replace integer literals which are cast to bool.

hjmjohnson:
Finds and replaces integer literals which are cast to bool.

SRCDIR=/Users/johnsonhj/Dashboard/src/ITK #My local SRC
BLDDIR=/Users/johnsonhj/Dashboard/src/ITK-clang/ #My local BLD

cd /Users/johnsonhj/Dashboard/src/ITK-clang/
run-clang-tidy.py -checks=-*,modernize-use-bool-literals  -header-filter=.* -fix

PR #3152, opened 2022-02-02
ENH: Expose FFT filter dimension traits

tbirdso:
Moves FFT filter image dimensions into existing trait objects in order
to decouple FFT factory overrides from fixed dimension lists. This will
allow other, external FFT implementations that may not be valid for
higher-dimension overrides to reuse the FFT factory class.

Future changes may replace the explicit dimension lists with CMake
variables to avoid registering unnecessary factories.

Motivation for these changes comes from developing accelerated FFT filter overrides in the [ITKVkFFTBackend](https://github.com/InsightSoftwareConsortium/ITKVkFFTBackend) external module.

## PR Checklist

### DIFF
--- a/Modules/Filtering/FFT/include/itkFFTImageFilterFactory.h
+++ b/Modules/Filtering/FFT/include/itkFFTImageFilterFactory.h
@@ -116,11 +116,11 @@ protected:
   {
     OverrideFFTImageFilterType<typename FFTImageFilterTraits<TFFTImageFilter>::template InputPixelType<float>,
                                typename FFTImageFilterTraits<TFFTImageFilter>::template OutputPixelType<float>>(
-      std::integer_sequence<unsigned int, 4, 3, 2, 1>{});
+      typename FFTImageFilterTraits<TFFTImageFilter>::FilterDimensions{});
 
     OverrideFFTImageFilterType<typename FFTImageFilterTraits<TFFTImageFilter>::template InputPixelType<double>,
                                typename FFTImageFilterTraits<TFFTImageFilter>::template OutputPixelType<double>>(
-      std::integer_sequence<unsigned int, 4, 3, 2, 1>{});
+      typename FFTImageFilterTraits<TFFTImageFilter>::FilterDimensions{});
   }
 };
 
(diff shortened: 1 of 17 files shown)


thewtex (approved):

blowekamp:
Looks like a good step forward. Do you see the CMake options being a list like this or a start, stop range?

 It is a shame that the [std::make_integer_sequence ](https://en.cppreference.com/w/cpp/utility/integer_sequence) starts at 0, so we can't directly use it. The results of it could have additional performed on each integer element with a variadic template to perform addition. I could write a class to do that if you think it would help with the final cmake parameters needed.

tbirdso:
@blowekamp I haven't tried it yet but I think the simplest way would be to rely on the explicit list of dimensions specified by the user in `ITK_WRAP_IMAGE_DIMS`, or define a similar CMake variable to do the same. That way the logic for creating dimension lists would be handled in the CMake config step and the code you've already written (thanks!) for unpacking the `integer_sequence` would suffice for handling overrides at compile time.

Leengit:
I didn't dive deeply enough to understand these variadic templates ... apologies if this is a dumb question:

In
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/FFT/include/itkFFTImageFilterFactory.h#L98
and
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/FFT/include/itkFFTImageFilterFactory.h#L108
what is the advantage of having the likes of
```c++
const std::integer_sequence<unsigned int, D, ImageDimensions...> &
```
as an argument to the method?  Could the expression that appears as an argument be eliminated on line 98 making this a function of void?  On line #108, could the type that appears as an argument be placed with the other template parameters rather than as a function argument?

I am thinking that the usual arguments about partially templated methods do not apply.  Is it a matter of style / preference?

tbirdso:
@Leengit It looks like there was discussion around this in #3065 regarding how the recursive procedure is terminated, to me this falls under an "if it ain't broke" scenario. If you and @blowekamp decide that a different implementation would be advantageous it may be best to track discussion in a dedicated issue.

Leengit:
> how the recursive procedure is terminated

Of course.  That mine was a dumb question is now confirmed.  :-)

dzenanz (approved):

tbirdso:
@Leengit Not a dumb question, thanks for double-checking on something that wasn't immediately clear. If/when ITK moves to C++17 one day in the future, we could replace the current recursive implementation with something like a [fold expression](https://en.cppreference.com/w/cpp/language/fold) to greatly improve readability.

PR #2829, opened 2021-10-21
ENH: pickle serialization for itk images (issue 1091)

GenevieveBuckley:
This PR adds pickle support to ITK images. We want this so that ITK images can be serialized for use with Dask arrays.

Closes https://github.com/InsightSoftwareConsortium/ITK/issues/1091

cc @thewtex 

## PR Checklist
- [ ] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [ ] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [ ] Added test (or behavior not changed)
- [ ] Updated API documentation (or API not changed)
- [ ] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [ ] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [ ] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples) for all new major features (if any)

Refer to the [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) for
further development details if necessary.

PR #3154, opened 2022-02-03
WIP: Support for Mesh serialization

PranjalSahu:
This is a work in progress. A working prototype is presented here to get comments.

SetPointsUsingArray and GetPointsAsArray methods are added in PointSet.
Similarly, SetCellsUsingArray and GetCellsUsingArray are added in Mesh. 

The approach handles different types of meshes by passing its enum type as the first entry 
and then the point ids similar to the approach in [MeshFileWriter](https://github.com/InsightSoftwareConsortium/ITK/blob/94b80ea4bc4a7789f5be962967a7ed09bd187d34/Modules/IO/MeshBase/include/itkMeshFileWriter.hxx#L335-L401)

The functions take as argument 1D VectorContainer which can be easily converted to a numpy and reverse.
The NumPy array is passed as dict and __getstate__ and __setstate__ methods are added in python (refer https://github.com/InsightSoftwareConsortium/ITK/pull/2829).
This will be used to serialize and deserialize the mesh.

Currently working on the PointData and CellData. Should be similar to the current approach.
More Tests will be added to handle different cells and other corner cases. 
Also in the dict other metadata will be passed ex. mesh type etc. to reconstruct the correct type of mesh.

## PR Checklist


thewtex:
@PranjalSahu awesome! The data structure should follow:

  https://insightsoftwareconsortium.github.io/itk-wasm/api/Mesh.html

GenevieveBuckley:
I'm slightly confused about why I've been requested as a reviewer, I don't do a whole lot with meshes (maybe will do more in the near future, but not a lot right now). Can you elaborate on what you were hoping for from me, @thewtex ?

thewtex:
@GenevieveBuckley , no expectations, but since you did such an amazing job on the `Image` serialization :-), your eyes may see pickling issues here. But, it is more just a heads-up that we are building on the work that you did for other data structures.

Wrapping/Generators/Python/itk/support/extras.py
```diff
@@ -837,9 +840,62 @@ def image_from_dict(image_dict: Dict) -> "itkt.Image":
...
+    else:
+        point_data  = itk.array_from_vector_container(point_data)
+
+    cell_data = mesh.GetCellData()
+    if cell_data.Size() == 0:
+        cell_data = np.array([], 'float64')
+    else:
+        cell_data  = itk.array_from_vector_container(cell_data)
+
+    return dict(
```
GenevieveBuckley, line 885: There are a lot of empty strings in this output dictionary. It looks like a round trip serialization/de serialization would lose a fair bit of metadata from the original input.

GenevieveBuckley (commented):
Ok, I've had a look over it now. It seems like nice work overall.
I know it says above this is still a work in progress. Before it is merged, my two biggest suggestions are that:
1. The `itkMeshSetGetCellTest.py` test assertions need to be strengthened, and
2. The dictionary returned on serialization currently includes a lot of empty string values, so it looks like there would be a fair bit of metadata loss upon serialization/de-serialization. This should also be addressed (and it will require a few new test assertions for these as well)

I don't feel like I can provide feedback on the header files, so I haven't left comments there.

Wrapping/Generators/Python/itk/support/extras.py
PranjalSahu: @GenevieveBuckley Thank you for the comments. Yes, Matt pointed to me the wasm mesh data structure ([itk wasm](https://insightsoftwareconsortium.github.io/itk-wasm/api/Mesh.html)) and I will use the same keys here. The current code has some placeholders and will be replaced with the correct values. 

I wasn't sure about some of the metadata in [MeshType](https://insightsoftwareconsortium.github.io/itk-wasm/api/MeshType.html) ex. cellPixelComponents so didn't fill it yet.

PranjalSahu:
@thewtex I found that we have Object name functionality present for all objects. I am thinking of using that instead of using a new field for **name** in the **PointSet**. If you agree then I will remove the addition of new **name** field from this PR.

Also, I have not added the support for QuadEdge Mesh in the wasm data structure in this PR.

PranjalSahu:
Also I realized for code coverage I should instantiate all types of Cells while testing SetCellsAsArray.

Wrapping/Generators/Python/itk/support/helpers.py
```diff
@@ -178,6 +178,73 @@ def image_filter_wrapper(*args, **kwargs):
...
+    _pixelType_to_prefix = {
+        "Scalar": "",
+        "RGB": "RGB",
+        "RGBA": "RGBA",
+        "Offset": "O",
+        "Vector": "V",
+        "CovariantVector": "CV",
+        "SymmetricSecondRankTensor": "SSRT",
+        "FixedArray": "FA",
+        "Array": "A",
```
dzenanz, line 242: Should `VLV` go here? VariableLengthVector might need to be supported here, maybe elsewhere too.

dzenanz (commented):
Mostly looks good.

Modules/Core/Mesh/include/itkMesh.hxx
```diff
@@ -103,6 +103,160 @@ Mesh<TPixelType, VDimension, TMeshTraits>::SetCells(CellsContainer * cells)
...
+{
+  itkDebugMacro("setting Cells container to " << cells);
+
+  this->ReleaseCellsMemory();
+  unsigned long index = 0;
+  unsigned long cellId = 0;
+
+  while (index < cells->Size())
+  {
+    int currentCellType = static_cast<int>(cells->GetElement(index++));
```
dzenanz, line 209: Should we always cast to `int`? Not `CellIdentifier` or something else?

PranjalSahu: That element will always be a cell type which are limited in number (~ 10) so int is fine there.

The point ids in that cell are taken in this line 
`cellPointer->SetPointId(i, cells->GetElement(index++));`

And the cell is inserted at:
`this->m_CellsContainer->InsertElement(cellId++, cellPointer.ReleaseOwnership());`
where the cellId is unsigned long. I can replace it to CellIdentifier.

Modules/Core/Common/include/itkPointSet.h
```diff
@@ -128,6 +132,10 @@ class ITK_TEMPLATE_EXPORT PointSet : public DataObject
    * separated into. */
   itkGetConstMacro(MaximumNumberOfRegions, RegionType);
 
+  /** Set/Get the the name of PointSet object */
+  itkSetMacro(Name, std::string);
+  itkGetConstMacro(Name, std::string);
```
thewtex, line 137: Per discussion offline, can this be removed in favor of `ObjectName`?

Modules/Core/Common/include/itkPointSet.h
```diff
@@ -244,6 +255,9 @@ class ITK_TEMPLATE_EXPORT PointSet : public DataObject
   RegionType m_RequestedNumberOfRegions;
   RegionType m_BufferedRegion;
   RegionType m_RequestedRegion;
+
+  // Name of the PointSet
+  std::string m_Name;
```
thewtex, line 260: remove

Modules/Core/Common/include/itkPointSet.hxx
```diff
@@ -66,6 +66,41 @@ PointSet<TPixelType, VDimension, TMeshTraits>::SetPoints(PointsContainer * point
...
+PointSet<TPixelType, VDimension, TMeshTraits>::SetPointsUsingArray(PointsVectorContainer * points)
+{
+  itkDebugMacro("setting Points container to " << points);
+
+  if (points->Size() % PointDimension != 0)
+  {
+    itkExceptionMacro("Number of entries in given 1d array incompatible with the point dimension");
+  }
+
+  unsigned long int numberOfPoints = points->Size() / PointDimension;
```
thewtex, line 80: `unsigned long int` -> `SizeValueType` (`unsigned long int` is only 32 bits on Windows).

Modules/Core/Common/include/itkPointSet.hxx
```diff
@@ -66,6 +66,41 @@ PointSet<TPixelType, VDimension, TMeshTraits>::SetPoints(PointsContainer * point
...
+
+  if (!m_PointsContainer)
+  {
+    this->SetPoints(PointsContainer::New());
+  }
+  m_PointsContainer->Initialize();
+  m_PointsContainer->Reserve(numberOfPoints);
+
+  unsigned long index = 0;
+  unsigned long point_id = 0;
```
thewtex, line 90: `unsigned long` -> `SizeValueType`

`point_id` -> `pointId` per ITK variable naming conventions

Modules/Core/Common/include/itkPointSet.h
```diff
@@ -156,6 +164,9 @@ class ITK_TEMPLATE_EXPORT PointSet : public DataObject
   void
   SetPoints(PointsContainer *);
 
+  void
+  SetPointsUsingArray(PointsVectorContainer *);
+
```
thewtex, line 169: Do we need a `GetPointsArray` here?

Modules/Core/Mesh/include/itkMesh.h
```diff
@@ -331,6 +356,16 @@ class ITK_TEMPLATE_EXPORT Mesh : public PointSet<TPixelType, VDimension, TMeshTr
...
 
+  virtual void
+  SetCellsUsingArray(CellsVectorContainer *);
+
+  // Convenience method for inserting cells when they are all of same type
+  virtual void
+  SetCellsUsingArray(CellsVectorContainer *, int cellType);
+
+  virtual CellsVectorContainer *
+  GetCellsAsArray();
```
thewtex, line 367: Consider `GetCellsAsArray` -> `GetCellsArray`

thewtex (changes requested):
Looking great!!!

A few comments inline.

Could we also please add a test for serialization of a `PointSet`?

Modules/Core/Common/include/itkPointSet.h
PranjalSahu: No, output from GetPoints() can be converted to 2D numpy array using itk.array_from_vector_container

Modules/Core/Mesh/include/itkMesh.hxx
PranjalSahu: Used IdentifierType where appropriate.

Modules/Core/Common/include/itkPointSet.hxx
PranjalSahu: used IdentifierType where appropriate.

dzenanz (commented):
LGTM, assuming CI passes.

thewtex:
@PranjalSahu could you please rebase on `master`?

PranjalSahu:
yes working on it.

thewtex:
To investigate: [Wrap PyVectorContainer for `Point`](https://github.com/InsightSoftwareConsortium/ITK/blob/928c2ce89ab8abb0dc485bcbbdc6b19b61b67a66/Modules/Bridge/NumPy/wrapping/itkPyVectorContainer.wrap#L3) , [direct copy](https://github.com/InsightSoftwareConsortium/itk-wasm/pull/512/files), change Python test code to use `SetPoints` instead of `SetPointsArray`, etc.

PranjalSahu:
@thewtex SetPointsArray is replaced with SetPoints. There are two versions for it now one 
with argument type (**itkVectorContainerULPF3** *) and other with (**itkVectorContainerULF** *). 

The one with **itkVectorContainerULF** makes it easier to directly use NumPy arrays for setting
the points (no conversion to **itkPoint** needed). So the output from other libraries like scikit etc. (ex. Nx3 Numpy arrays) can be plugged directly.


I used the reinterpret_cast for this inside the newly added SetPoints:
`
auto * pointsPtr = reinterpret_cast<PointsContainer *>(points);
`

No need for **assign** operation is needed. Also, I found that for DynamicTraits it wasn't compiling as it 
does not support direct **assign** operation (map).

Sample usage in Python:
```
points_vc = itk.vector_container_from_array(arr.flatten())
mesh.SetPoints(points_vc)
```

Modules/Core/Common/wrapping/test/itkPointSetSerializationTest.py
```diff
@@ -0,0 +1,76 @@
...
+    v_point.SetElement(i, point)
+
+arr = itk.array_view_from_vector_container(v_point)
+points_vc = itk.vector_container_from_array(arr.flatten())
+
+point_set.SetPoints(points_vc)
+
+
+# Check the serialization of mesh
+serialize_deserialize = pickle.loads(pickle.dumps(point_set))
```
thewtex, line 54: :100:

thewtex (approved):
We have converged on an excellent outcome. @PranjalSahu thanks for your hard work. @GenevieveBuckley thanks for the review.

@PranjalSahu awesome to see the `PointSet` serialization here, too!

Modules/Core/Common/include/itkPointSet.hxx
```diff
@@ -66,6 +66,25 @@ PointSet<TPixelType, VDimension, TMeshTraits>::SetPoints(PointsContainer * point
...
+void
+PointSet<TPixelType, VDimension, TMeshTraits>::SetPoints(PointsVectorContainer * points)
+{
+
+  itkDebugMacro("setting Points container to " << points);
+  if (points->Size() % PointDimension != 0)
+  {
+    itkExceptionMacro("Number of entries in given 1d array incompatible with the point dimension");
+  }
+  auto * pointsPtr = reinterpret_cast<PointsContainer *>(points);
```
N-Dekker, line 82: I'm sorry but this `reinterpret_cast<PointsContainer *>(points)` very much looks like undefined behavior to me. How did you find out that it seemed to work at all?

PranjalSahu: This can be tested by using a 1D array. I can write a small test case here if it is not present already.
I am checking if this is covered under test or not.

Modules/Core/Mesh/wrapping/test/itkMeshSerializationTest.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+# Set Cells in the Mesh as Triangle Cell
+cells_array = np.zeros([NumberOfCells, Dimension], np.uint)
+
+# Insert point ids in that cell
+for i in range(NumberOfCells):
+    cells_array[i, :] = [i * 1, (i + 1) * 2, (i + 2) * 3]
+
+# Insert the cells in mesh by converting to 1D array
```
PranjalSahu, line 59: One similar test case is here which uses a 1D input array.

Modules/Core/Common/include/itkPointSet.hxx
N-Dekker: Thanks @PranjalSahu I'm glad to see that you already tested this code: https://github.com/InsightSoftwareConsortium/ITK/blob/9a5dade7f7e13fcef307f8224a01c2ce2b0ab13b/Modules/Core/Common/test/itkPointSetTest.cxx#L102-L127

Nevertheless, it's officially undefined behavior. Sorry 🤷. The C++ Standard does not guarantee that it works. In theory, it might do anything. And in practice, undefined behavior _does_ lead to trouble. I just asked about [a similar example](https://godbolt.org/z/ahPze3YPz) at the ACCU mailing list, and Jonathan Wakely (@jwakely) confirmed that it is indeed undefined behavior.

So I would prefer to see the `reinterpret_cast<PointsContainer *>` "trick" removed. But otherwise (at least), a note should be added that this function is unsafe.

PR #3155, opened 2022-02-03
COMP: Increase test coverage for `FFTDiscreteGaussianImageFilter`

tbirdso:
Adds test step for streaming `FFTDiscreteGaussianImageFilterEnums` which was previously [not covered](https://open.cdash.org/viewCoverageFile.php?buildid=7717739&fileid=49970264).

## PR Checklist

### DIFF
--- a/Modules/Filtering/Smoothing/test/itkFFTDiscreteGaussianImageFilterTest.cxx
+++ b/Modules/Filtering/Smoothing/test/itkFFTDiscreteGaussianImageFilterTest.cxx
@@ -95,6 +95,13 @@ itkFFTDiscreteGaussianImageFilterTestProcedure(int argc, char ** argv)
 
   itk::WriteImage(filter->GetOutput(), argv[3], true);
 
+  // Test streaming enumeration for FFTDiscreteGaussianImageFilterEnums::KernelSource elements
+  for (const auto val : { itk::FFTDiscreteGaussianImageFilterEnums::KernelSource::OPERATORS,
+                          itk::FFTDiscreteGaussianImageFilterEnums::KernelSource::IMAGE_SOURCE })
+  {
+    std::cout << "STREAMED ENUM VALUE FFTDiscreteGaussianImageFilterEnums::KernelSource: " << val << std::endl;
+  }
+
   return EXIT_SUCCESS;
 }
 


dzenanz (approved):

thewtex (approved):
Thanks for checking coverage! :book:

Modules/Filtering/Smoothing/test/itkFFTDiscreteGaussianImageFilterTest.cxx
```diff
@@ -95,6 +95,16 @@ itkFFTDiscreteGaussianImageFilterTestProcedure(int argc, char ** argv)
 
   itk::WriteImage(filter->GetOutput(), argv[3], true);
 
+  // Test streaming enumeration for FFTDiscreteGaussianImageFilterEnums::KernelSource elements
+  const std::set<itk::FFTDiscreteGaussianImageFilterEnums::KernelSource> allKernelSource{
+    itk::FFTDiscreteGaussianImageFilterEnums::KernelSource::OPERATORS,
+    itk::FFTDiscreteGaussianImageFilterEnums::KernelSource::IMAGE_SOURCE
+  };
+  for (const auto & ee : allKernelSource)
```
N-Dekker, line 103: What was the reason to construct an `std::set` of these enum values? I saw it before, I'm just wondering why. FYI, the following also works, creating an initializer-list "on the fly": 

    for (const auto enumValue : { 
        itk::FFTDiscreteGaussianImageFilterEnums::KernelSource::OPERATORS,
        itk::FFTDiscreteGaussianImageFilterEnums::KernelSource::IMAGE_SOURCE} )
    {
        ....
    }

In general, enum values do not need to be passed by "const reference" (`const auto &`), because they can be copied quickly, so I would rather leave the ampersand `&` out.

Just my two cents. HTH, Niels

tbirdso: I based my changes off of a previous implementation in ITK, I like yours better 😃 Updated.

N-Dekker (approved):
@tbirdso Thank you for addressing my comment!

tbirdso:
@N-Dekker I'm going to go ahead and merge this as CI is passing, please feel free to continue discussion here or move to a dedicated issue / discourse topic

tbirdso:
I'm not sure about how CMake determines Google Test coverage unfortunately.

jhlegarreta:
I do not think there is a difference from the point of view of the coverage whether it is written as a `GTest` or not. Have not looked at it, but I guess any executable generated from the `tests` folders is passed to `gcov` so that any hit gets counted.

There are cases, such the one pointed by @N-Dekker, that are supposedly covered by some test, but are somehow not being counted/not shown as hit in the dashboard. There are a few `GTest`-only covered classes/methods that are shown as covered. Knowing whether they are actually covered because they get called from somewhere else is not evident to me. Maybe one can rule out that by testing
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/Maintenance/computeCodeCoverageLocallyForOneTest.sh

with a `GTest` and seeing the report.

I have always assumed that statements called from `GTest` are being counted, and have granted that there are inconsistencies in the coverage such as the one pointed. Unfortunately, this makes it harder to investigate the statements that are not actually covered in order to raise the coverage.

I do not have a better explanation. I might be wrong in all of the above.

N-Dekker:
Thanks for your prompt reply, @jhlegarreta

Honestly I still don't understand the technical issue that appears to cause some (?) GTests from being excluded from code coverage analysis. To be continued, hopefully...  😃

jhlegarreta:
> Honestly I still don't understand the technical issue that appears to cause some (?) GTests from being excluded from code coverage analysis. To be continued, hopefully... smiley

Sorry if I was not clear enough: I'd dare to say that false negatives also happen for non-`GTest` tests. That is, lines that the dashboard says are not covered seem to be called from non-`GTest` tests. I do not have examples handy now/do not have time to look and provide some example now.

PR #3118, opened 2022-01-17
BUG: Initialize `itk::ExhaustiveOptimizer::m_MinimumMetricValue`

jhlegarreta:
Provide an initial value to the
`itk::ExhaustiveOptimizer::m_MinimumMetricValue` ivar.

Fixes:
```
UMC ==18288== Conditional jump or move depends on uninitialised value(s)
==18288==    at 0x663E1F8:
__printf_fp_l (printf_fp.c:387)
==18288==    by 0x6659639:
__vfprintf_internal (vfprintf-internal.c:1687)
==18288==    by 0x666E119:
__vsnprintf_internal (vsnprintf.c:114)
==18288==    by 0x64D4C3F:
??? (in /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.28)
==18288==    by 0x6506245:
std::ostreambuf_iterator > std::num_put > >::
_M_insert_float(std::ostreambuf_iterator >, std::ios_base&, char, char, double) const
(in /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.28)
==18288==    by 0x6514972:
std::ostream& std::ostream::_M_insert(double) (in /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.28)
==18288==    by 0x489F46D:
itk::ExhaustiveOptimizer::PrintSelf(std::ostream&, itk::Indent) const (itkExhaustiveOptimizer.cxx:229)
==18288==    by 0x53B4B19:
itk::LightObject::Print(std::ostream&, itk::Indent) const (itkLightObject.cxx:125)
==18288==    by 0x16536A:
itkExhaustiveOptimizerTest(int, char**) (itkExhaustiveOptimizerTest.cxx:158)
==18288==    by 0x12F2E7:
main (ITKOptimizersTestDriver.cxx:222)
```

Uncovered by the dynamic analysis in:
https://open.cdash.org/viewDynamicAnalysisFile.php?id=9074678

## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [X] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)

PR #3158, opened 2022-02-03
STYLE: In-class default member initializers for ExhaustiveOptimizer data

N-Dekker:
Follow-up to pull request #3102 and pull request #3118, both by Jon Haitz Legarreta Gorroño (@jhlegarreta).

### DIFF
--- a/Modules/Numerics/Optimizers/src/itkExhaustiveOptimizer.cxx
+++ b/Modules/Numerics/Optimizers/src/itkExhaustiveOptimizer.cxx
@@ -25,16 +25,6 @@ namespace itk
 ExhaustiveOptimizer::ExhaustiveOptimizer()
 {
   itkDebugMacro("Constructor");
-
-  m_StepLength = 1.0;
-  m_CurrentIteration = 0;
-  m_CurrentValue = 0;
-  m_CurrentParameter = 0;
-  m_CurrentIndex.Fill(0);
-  m_Stop = false;
-  m_NumberOfSteps.Fill(0);
-
-  m_StopConditionDescription.str("");
 }
 
 /**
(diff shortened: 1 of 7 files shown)


N-Dekker:
The empty `{}` default member initializers that I added with this pull request are not strictly necessary. They are just there to make it _absolutely clear_ that none of the data member were forgotten, anymore.

Those empty `{}` default member initializers would be really needed for types that do _not_ have a user-defined default-constructor. However, it may not be obvious at a glance whether or not types like `StepsType` and `ParametersType` have a user-defined default-constructor.

Modules/Numerics/Optimizers/src/itkExhaustiveOptimizer.cxx
```diff
@@ -25,16 +25,6 @@ namespace itk
 ExhaustiveOptimizer::ExhaustiveOptimizer()
 {
   itkDebugMacro("Constructor");
```
blowekamp, line 27: Is this still useful? Could we just use the default constructor?

N-Dekker: @blowekamp Good question. 😃  This `itkDebugMacro("Constructor")` certainly isn't useful to me. I do see them in a few other places in ITK, though. Shall we remove them all?

If we would remove the `itkDebugMacro("Constructor")` call, we could certainly do `ExhaustiveOptimizer::ExhaustiveOptimizer() = default`.

dzenanz: These sort of debug statements are useful during initial development, and occasionally during some later debugging. As this is quite a mature class by now, I vote we remove it and go with `= default;`.

dzenanz (approved):

dzenanz (approved):

hjmjohnson (approved):

PR #3020, opened 2022-01-02
BUG: fixed thread safety errors found by Thread Sanitizer

seanm:
std::vector<bool> has less thread safety than vectors of other types, because multiple boolean bits can be packed into the same byte(s). C++11 standard § 23.2.2, paragraph 2 describes this futher.

So instead, just use int instead of bool. This allows enough thread safety for the usage at hand, as evidenced by various tests now passing under thread sanitizer.

PR #3166, opened 2022-02-05
BUG: fixed threading bug, don't use vector<bool> from multiple threads

seanm:
vector<bool> is not thread safe due to the possibility of multiple bits being packed together in the same memory location.   Found by TSan.

### DIFF
--- a/Modules/Numerics/Optimizersv4/include/itkQuasiNewtonOptimizerv4.h
+++ b/Modules/Numerics/Optimizersv4/include/itkQuasiNewtonOptimizerv4.h
@@ -151,8 +151,11 @@ protected:
   /** The Hessian with local support */
   HessianArrayType m_HessianArray;
 
-  /** Valid flag for the Quasi-Newton steps */
-  std::vector<bool> m_NewtonStepValidFlags;
+  /** Valid flag for the Quasi-Newton steps.
+   * NB: although semantically boolean, vector<bool> is not thread safe due to the possibility of multiple bits being
+   * packed together in the same memory location.
+   */
+  std::vector<uint8_t> m_NewtonStepValidFlags;
 
   /** Estimate a Newton step */
   virtual void


seanm:
This failure did not occur a week or two ago when I last ran TSan, so it was presumably recently introduced.

Modules/Numerics/Optimizersv4/include/itkQuasiNewtonOptimizerv4.h
```diff
@@ -151,8 +151,11 @@ class ITK_TEMPLATE_EXPORT QuasiNewtonOptimizerv4Template
   /** The Hessian with local support */
   HessianArrayType m_HessianArray;
 
-  /** Valid flag for the Quasi-Newton steps */
-  std::vector<bool> m_NewtonStepValidFlags;
+  /** Valid flag for the Quasi-Newton steps.
+   * NB: although semantically boolean, vector<bool> is not thread safe due to the possibility of multiple bits being
+   * packed together in the same memory location.
+   */
+  std::vector<int> m_NewtonStepValidFlags;
```
N-Dekker, line 158: Good catch @seanm !

Please consider ` std::vector<uint8_t>`, as you did before, with pull request #3020 commit b66133ab53369bdf265348c0c31cfa6451b53934

BTW, if the use of `std::vector<"semantically boolean">` appears more often in ITK, I would suggest adding a type alias, for example `using Bool = uint8_t`. So that it can be documented in one place.

N-Dekker (approved):
@seanm Thanks for switching to `std::vector<uint8_t>`  👍  I still think it would be better to have a dedicated "semantically boolean" type (or type alias) added to ITK, which would then be documented on one place. (Rather than documenting the reason for `std::vector<uint8_t>` on multiple places, wherever it is used.) But for now I see it's most important that the threading bug is fixed.

PR #3173, opened 2022-02-08
Add a type alias, `BooleanStdVectorType`, for `std::vector<uint8_t>`

N-Dekker:
Follow-up to pull request #3166 commit 915b8e48d8ff7d9f5b943eb673826ad85bf74a8c "BUG: fixed threading bug, don't use vector<bool> from multiple threads", by Sean McBride (@seanm) (merged on 8 February 2022).

### DIFF
--- a/Modules/Core/Common/include/itkBooleanStdVector.h
+++ b/Modules/Core/Common/include/itkBooleanStdVector.h
@@ -1,0 +1,41 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+#ifndef itkBooleanStdVector_h
+#define itkBooleanStdVector_h
+
+#include "itkIntTypes.h"
+#include <vector>
+
+namespace itk
+{
+
+/** The type alias `BooleanStdVectorType` provides an alternative to `std::vector<bool>`. `std::vector<bool>` is not
+ * thread safe due to the possibility of multiple bits being packed together in the same memory location.
+ * BooleanStdVectorType does not have such a space optimization. BooleanStdVectorType is semantically like
+ * `std::vector<bool>`, but unlike `std::vector<bool>`, it does "avoid data races when the contents of the contained
+ * object in different elements in the same container [...] are modified concurrently", according to the C++ Standard,
+ * section [container.requirements.dataraces], "Container data races".
+ *
+ * \note BooleanStdVectorType is only intended to be used for the storage of Boolean values (true and false), not for
+ * arbitrary integer values.
+ */
+using BooleanStdVectorType = std::vector<uint8_t>;
+
+} // namespace itk
+
+#endif
--- a/Modules/Core/Common/test/itkBooleanStdVectorGTest.cxx
+++ b/Modules/Core/Common/test/itkBooleanStdVectorGTest.cxx
@@ -1,0 +1,43 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+// First include the header file to be tested:
+#include "itkBooleanStdVector.h"
+
+#include <type_traits>
+
+namespace
+{
+using ValueType = itk::BooleanStdVectorType::value_type;
+
+static_assert(sizeof(ValueType) <= sizeof(bool),
+              "The value type of BooleanStdVectorType should not be larger than the built-in C++ bool type.");
+static_assert(
+  ValueType{ false } == false,
+  "The value type of BooleanStdVectorType should be allowed as a conditional expression, evaluating to false.");
+static_assert(
+  ValueType{ true } == true,
+  "The value type of BooleanStdVectorType should be allowed as a conditional expression, evaluating to true.");
+static_assert(bool{ ValueType{ false } } == false,
+              "The value type of BooleanStdVectorType should support a lossless round-trip from bool value false.");
+static_assert(bool{ ValueType{ true } } == true,
+              "The value type of BooleanStdVectorType should support a lossless round-trip from bool value true.");
+static_assert(std::is_same<decltype(*itk::BooleanStdVectorType{}.data()), ValueType &>::value,
+              "BooleanStdVectorType should have a valid data() member function (unlike std::vector<bool>).");
+
+} // namespace
--- a/Modules/Core/Common/test/CMakeLists.txt
+++ b/Modules/Core/Common/test/CMakeLists.txt
@@ -605,6 +605,7 @@ endif()
 
 set(ITKCommonGTests
       itkAggregateTypesGTest.cxx
+      itkBooleanStdVectorGTest.cxx
       itkBuildInformationGTest.cxx
       itkConnectedImageNeighborhoodShapeGTest.cxx
       itkConstantBoundaryImageNeighborhoodPixelAccessPolicyGTest.cxx
--- a/Modules/Core/FiniteDifference/include/itkDenseFiniteDifferenceImageFilter.h
+++ b/Modules/Core/FiniteDifference/include/itkDenseFiniteDifferenceImageFilter.h
@@ -18,6 +18,7 @@
 #ifndef itkDenseFiniteDifferenceImageFilter_h
 #define itkDenseFiniteDifferenceImageFilter_h
 
+#include "itkBooleanStdVector.h"
 #include "itkFiniteDifferenceImageFilter.h"
 #include "itkMultiThreaderBase.h"
 
@@ -173,9 +174,7 @@ private:
     TimeStepType                       TimeStep;
     std::vector<TimeStepType>          TimeStepList;
 
-    // NB: although semantically boolean, vector<bool> is not thread safe due to the possibility of multiple bits being
-    // packed together in the same memory location.
-    std::vector<uint8_t> ValidTimeStepList;
+    BooleanStdVectorType ValidTimeStepList;
   };
 
   /** This callback method uses ImageSource::SplitRequestedRegion to acquire an
--- a/Modules/Core/FiniteDifference/include/itkFiniteDifferenceImageFilter.h
+++ b/Modules/Core/FiniteDifference/include/itkFiniteDifferenceImageFilter.h
@@ -18,6 +18,7 @@
 #ifndef itkFiniteDifferenceImageFilter_h
 #define itkFiniteDifferenceImageFilter_h
 
+#include "itkBooleanStdVector.h"
 #include "itkInPlaceImageFilter.h"
 #include "itkFiniteDifferenceFunction.h"
 
@@ -331,11 +332,11 @@ protected:
    * \param timeStepList The set of time changes compiled from all the threaded calls
    * to ThreadedGenerateData.
    * \param valid The set of flags indicating which of "timeStepList" elements are
-   *  valid. Although they are uint8_t, they should be treated like bools.
+   *  valid.
    *
    * The default is to return the minimum value in the list. */
   virtual TimeStepType
-  ResolveTimeStep(const std::vector<TimeStepType> & timeStepList, const std::vector<uint8_t> & valid) const;
+  ResolveTimeStep(const std::vector<TimeStepType> & timeStepList, const BooleanStdVectorType & valid) const;
 
   /** Set the number of elapsed iterations of the filter. */
   itkSetMacro(ElapsedIterations, IdentifierType);
--- a/Modules/Core/FiniteDifference/include/itkFiniteDifferenceImageFilter.hxx
+++ b/Modules/Core/FiniteDifference/include/itkFiniteDifferenceImageFilter.hxx
@@ -167,7 +167,7 @@ FiniteDifferenceImageFilter<TInputImage, TOutputImage>::GenerateInputRequestedRe
 template <typename TInputImage, typename TOutputImage>
 typename FiniteDifferenceImageFilter<TInputImage, TOutputImage>::TimeStepType
 FiniteDifferenceImageFilter<TInputImage, TOutputImage>::ResolveTimeStep(const std::vector<TimeStepType> & timeStepList,
-                                                                        const std::vector<uint8_t> &      valid) const
+                                                                        const BooleanStdVectorType &      valid) const
 {
   TimeStepType oMin = NumericTraits<TimeStepType>::ZeroValue();
   bool         flag = false;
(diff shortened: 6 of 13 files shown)


seanm:
Hmmm, I wonder if there isn't an even better way... like can one make a `vector` of `atomic_flag` or a `vector` of `std::atomic<bool>`?  That would probably be even more self-documenting if it worked.

N-Dekker:
> Hmmm, I wonder if there isn't an even better way... like can one make a `vector` of `atomic_flag` or a `vector` of `std::atomic<bool>`? That would probably be even more self-documenting if it worked.

Good question @seanm First of all, this PR is just meant as a style improvement. Your original commits already allowed multiple threads to simultaneously adjust different elements from one and the same vector. And thereby it fixed those TSan warnings, right? Do we need more than that? I'm asking, because making everything atomic isn't for free either. It also has a performance price. So there might be a trade off.

Modules/Core/Common/include/itkBooleanStdVector.h
```diff
@@ -0,0 +1,43 @@
...
+{
+
+/** The element type of `BooleanStdVectorType`. Should be treated like a Boolean type. */
+using BooleanType = uint8_t;
+
+static_assert(sizeof(BooleanType) <= sizeof(bool), "itk::BooleanType should not be larger than the built-in C++ bool");
+static_assert(!BooleanType{ false }, "itk::BooleanType{ false } should be allowed as a conditional expression.");
+static_assert(BooleanType{ true }, "itk::BooleanType{ true } should be allowed as a conditional expression.");
+static_assert(bool{ BooleanType{ true } } == true && bool{ BooleanType{ false } } == false,
+              "itk::BooleanType should support lossless roundtip from a bool value.");
```
Leengit, line 34: Let's also assert about equality and logical and bitwise operators.  And if any of these fail, let's document that.
```c++
static_assert(BooleanType{ false } == !BooleanType{ true }, "itk::BooleanType is bad");
static_assert(BooleanType{ true } == !BooleanType{ false }, "itk::BooleanType is bad");
static_assert(BooleanType{ false } == ~BooleanType{ true }, "itk::BooleanType is bad");
static_assert(BooleanType{ true } == ~BooleanType{ false }, "itk::BooleanType is bad");
static_assert((BooleanType{ true } & BooleanType{ true }) == BooleanType{ true }, "itk::BooleanType is bad");
static_assert((BooleanType{ true } & BooleanType{ false }) == BooleanType{ false }, "itk::BooleanType is bad");
static_assert((BooleanType{ false } & BooleanType{ true }) == BooleanType{ false }, "itk::BooleanType is bad");
static_assert((BooleanType{ false } & BooleanType{ false }) == BooleanType{ false }, "itk::BooleanType is bad");
static_assert((BooleanType{ true } | BooleanType{ true }) == BooleanType{ true }, "itk::BooleanType is bad");
static_assert((BooleanType{ true } | BooleanType{ false }) == BooleanType{ true }, "itk::BooleanType is bad");
static_assert((BooleanType{ false } | BooleanType{ true }) == BooleanType{ true }, "itk::BooleanType is bad");
static_assert((BooleanType{ false } | BooleanType{ false }) == BooleanType{ false }, "itk::BooleanType is bad");
static_assert((BooleanType{ true } ^ BooleanType{ true }) == BooleanType{ false }, "itk::BooleanType is bad");
static_assert((BooleanType{ true } ^ BooleanType{ false }) == BooleanType{ true }, "itk::BooleanType is bad");
static_assert((BooleanType{ false } ^ BooleanType{ true }) == BooleanType{ true }, "itk::BooleanType is bad");
static_assert((BooleanType{ false } ^ BooleanType{ false }) == BooleanType{ false }, "itk::BooleanType is bad");
```

dzenanz: Since you wrote these asserts, putting them in should be easy!

Modules/Core/Common/include/itkBooleanStdVector.h
```diff
@@ -0,0 +1,43 @@
...
+#define itkBooleanStdVector_h
+
+#include "itkIntTypes.h"
+#include <vector>
+
+namespace itk
+{
+
+/** The element type of `BooleanStdVectorType`. Should be treated like a Boolean type. */
+using BooleanType = uint8_t;
```
dzenanz, line 28: Why not put this in https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkIntTypes.h?

Modules/Core/Common/include/itkBooleanStdVector.h
```diff
@@ -0,0 +1,43 @@
...
+
+static_assert(sizeof(BooleanType) <= sizeof(bool), "itk::BooleanType should not be larger than the built-in C++ bool");
```
dzenanz, line 30: This could go into a separate (G)Test file.

seanm:
>First of all, this PR is just meant as a style improvement.

Yup, I get that.

>Your original commits already allowed multiple threads to simultaneously adjust different elements from one and the same vector. And thereby it fixed those TSan warnings, right?

Correct.

I'm not sure I'm convinced this change has much benefit.  It's many LOC for not much benefit IMHO.  It doesn't help a programmer to know when he shouldn't use a vector of bools (not that I see a way to achieve that). But I don't strongly oppose it.  But I'd argue at least for a change of name, something like `ThreadSafeBooleanVectorType` that indicates *why* a plain `std::vector<bool>` isn't used

Modules/Core/Common/include/itkBooleanStdVector.h
seanm: Why must "itk::BooleanType not be larger than the built-in C++ bool"?

dzenanz: We don't want to accidentally waste memory.

Leengit:
It probably won't work due to a conflicting specialization in the standard library, but just in case you want to wrestle with it ... perhaps we could disallow `std::vector<bool>`:
```c++
namespace std
{
template <>
class vector<bool>
{
  static constexpr auto once = static_assert(false, "Bad!");
}
} // namespace std
```
... or somehow remove the body and replace it with `= delete;`?

Modules/Core/Common/include/itkBooleanStdVector.h
seanm: If somehow there were a system where this size check did not hold, yes, you would "waste" a bit of memory.  But the alternative is what?  Using a vector of bool?  No, that would be UB and worse than memory waste.  It's not like you have an option smaller than `uint8_t`.

Leengit:
> It probably won't work ...

Never mind.  It would break user's code if they legitimately wanted to use `vector<bool>`, and our code too.

Modules/Core/Common/include/itkBooleanStdVector.h
N-Dekker: @Leengit Thanks for the suggestion, but sorry, I find it a bit overkill to have so many asserts on such a simple typedef, `using BooleanType = uint8_t`.

Leengit: I get that.  Perhaps we could have these in a unit test file rather than in the include file that everyone must wait to compile.

seanm: Personally, I think all these asserts are overkill.

Modules/Core/Common/include/itkBooleanStdVector.h
N-Dekker: > Why not put this in https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkIntTypes.h?

Well, I would like to avoid that ITK has its own Boolean type for general use. I would prefer to "in general" keep using the built-in `bool` type.  This `BooleanType` typedef is _only_ intended to define the element type of `BooleanStdVectorType`.

We could also drop this `BooleanType` typedef entirely, and only define `BooleanStdVectorType = std::vector<uint8_t>`. Would that be fine to you as well?

dzenanz: > drop this BooleanType typedef entirely

That makes more sense if it is only going to be used for `std::vector`.

N-Dekker:
> But I'd argue at least for a change of name, something like `ThreadSafeBooleanVectorType` that indicates _why_ a plain `std::vector<bool>` isn't used

@seanm I understand. But `ThreadSafeBooleanVectorType` may also be somewhat confusing, as you already mentioned that the elements of such an `vector<uint8_t>` are not yet atomic. We could also call it something `UncompressedBooleanStdVectorType`, or `UnoptimizedBooleanStdVectorType`, to indicate that it does not use the space optimization of `std::vector<bool>`. What do you think?


Personally I still think it would be OK to stick with the name `itk::BooleanStdVectorType`, as long as it is clearly documented _why_ we use it.

Modules/Core/Common/include/itkBooleanStdVector.h
Leengit: If all the asserts pass then they are overkill. :smile:  But if any fail then that means that someone who is hoping to get the `bool` behavior will instead get the `uint8_t` behavior.  At the very least, I'd want to document any different behavior in `itkBooleanStdVector.h` (even if the asserts live only in a unit test file).

seanm: >someone who is hoping to get the bool behavior will instead get the uint8_t behavior

The whole point is to get the uint8_t behaviour because the bool behaviour is not thread safe.

seanm:
>But `ThreadSafeBooleanVectorType` may also be somewhat confusing, as you already mentioned that the elements of such an `vector<uint8_t>` are not yet atomic.

They are atomic. Each element can be read or written without any other thread seeing a half-written value.

>We could also call it something `UncompressedBooleanStdVectorType`, or `UnoptimizedBooleanStdVectorType`, to indicate that it does not use the space optimization of `std::vector<bool>`. What do you think?

The raison d'être for this typedef is thread safety, not "compression" nor "optimization".

Modules/Core/Common/include/itkBooleanStdVector.h
N-Dekker: > If somehow there were a system where this size check did not hold

That's very hypothetical, as long as we use `uint8_t`.  This check `sizeof(BooleanType) <= sizeof(bool)` is just meant to ensure that we do not _accidentally_ switch to using a larger integer type.  

If it appears that we would somewhere in the future _want_ to use a larger integer type, we would have to adjust this static_assert, of course. But for now I would rather not waste extra memory on this type.

N-Dekker:
@seanm 
> They are atomic. Each element can be read or written without any other thread seeing a half-written value.

I'm sorry I don't think so. I don't believe `uint8_t` is atomic "out of the box". But for the current use cases, it doesn't seem necessary to use atomics. If I understand correctly.

> The raison d'être for this typedef is thread safety, not "compression" nor "optimization".

Well, for now, yes. But this `BooleanStdVectorType` typedef can be used anywhere where an uncompressed vector of booleans would be wanted. Even in single-threaded code, it would probably be faster to iterate over an uncompressed vector of booleans than over `std::vector<bool>`.

seanm:
> I'm sorry I don't think so. I don't believe `uint8_t` is atomic "out of the box". But for the current use cases, it doesn't seem necessary to use atomics. If I understand correctly.

Yeah, sorry, they're not atomic.  I mean to say that per C++11 "23.2.2 Container data races": "implementations are required to avoid data races when the contents of the contained object in different elements in the same sequence, excepting vector<bool>, are modified concurrently"

But accessing the *same* element still requires a mutex or other synchronization technique. And that's exactly the issue in #3031.

> > The raison d'être for this typedef is thread safety, not "compression" nor "optimization".
> 
> Well, for now, yes.

I think it makes more sense to name it for its actual purpose instead of some hypothetical other purpose.

Modules/Core/Common/include/itkBooleanStdVector.h
Leengit: On my system, `uint8_t` has the same behavior as `bool` for all these equality, logical, and bitwise operators.  If this is the case for all platforms then the `static_asserts` are overkill.  

(Note that the tests I wrote above involving `~` are buggy; they insist that `uint8_t` do something that `bool` does not.)

N-Dekker: @Leengit Thanks, we could indeed add such a note. But of course, the proposed `BooleanStdVectorType` is only intended to be used for Boolean values, false and true. It's just not intended to be used to store arbitrary 8-bits integer values.

N-Dekker: @Leengit I hope I sufficiently addressed your suggestion for a note, by a [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/dec13145e404b8b0448c575e9f16a260fbe19727..283d885183e788a00bc693f63f04750a696c4cef) which adds: 
```
 * \note BooleanStdVectorType is only intended to be used for the storage of Boolean values (true and false), not for
 * arbitrary integer values.
 */
```

N-Dekker:
> I mean to say that per C++11 "23.2.2 Container data races": "implementations are required to avoid data races when the contents of the contained object in different elements in the same sequence, excepting vector, are modified concurrently"

Thanks @seanm I just added a quote from this C++ Standard citation to the documentation: [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/dec13145e404b8b0448c575e9f16a260fbe19727..283d885183e788a00bc693f63f04750a696c4cef)

Modules/Core/Common/include/itkBooleanStdVector.h
Leengit: What I want to alert the reader to is ... sometimes people will write
```c++
bool flag0 = myInt0;
bool flag1 = myInt1;
```
when they mean
```c++
bool flag0 = myInt0 != 0;
bool flag1 = myInt1 != 0;
```
because the former is more concise and these two approaches have the exact same result.  In particular, one can follow either approach with
```c++
if (flag1 == flag2)
{
    // myInt0 and myInt1 were both zero or were both nonZero
};
```

However, if we put `unint8_t` in there instead of `bool` then the two approaches are no longer equivalent, and the `flag1 == flag2` test will not do what we are expecting under the concise approach when, say, `myInt0=10` and `myInt1=20`.

I think I've laid out my ideas here.  Whatever you think is best will work for me.

N-Dekker: > sometimes people will write
> 
> ```c++
> bool flag0 = myInt0;
> bool flag1 = myInt1;
> ```

But you won't write that, right? Wouldn't you do brace-initialization? As in:

    bool flag0 { myInt0 };
    bool flag1 { myInt1 };

Then the compiler would warn you against such a "narrowing conversion" 😸 

> I think I've laid out my ideas here. 

I agree that it's good to be aware of the fundamental differences between bool and uint8_t.

> Whatever you think is best will work for me.

Thanks @Leengit And if you do think so, please approve this pull request  😄

Leengit: You are correct that I would not use that concise approach.  If all programmers were like me, the world would be a different place.  (Yes, I know that that can be interpreted in multiple ways.)

I have reviewed and approved.  Thanks!

Modules/Core/Common/test/itkBooleanStdVectorGTest.cxx
```diff
@@ -0,0 +1,43 @@
...
+static_assert(sizeof(ValueType) <= sizeof(bool),
+              "The value type of BooleanStdVectorType should not be larger than the built-in C++ bool type.");
+static_assert(
+  ValueType{ false } == false,
+  "The value type of BooleanStdVectorType should be allowed as a conditional expression, evaluating to false.");
+static_assert(
+  ValueType{ true } == true,
+  "The value type of BooleanStdVectorType should be allowed as a conditional expression, evaluating to true.");
+static_assert(bool{ ValueType{ false } } == false,
+              "The value type of BooleanStdVectorType should support a lossless round-tip from bool value false.");
```
issakomi, line 37: > round-tip

Maybe _round-trip_?

blowekamp:
Please keep in mind that if you are doing `boolean_array[thead_id]` with concurrent read/write access there will be false sharing and significant performance penalties. This is a poor way to organize data for threaded algorithms. If there occurrences of memory organization, it should be considered to change the data layout for improve threading.

N-Dekker:
> Please keep in mind that if you are doing `boolean_array[thead_id]` with concurrent read/write access there will be false sharing and significant performance penalties.

Thanks @blowekamp But please note that this pull request does not introduce any false sharing or performance penalty. It just introduces a new typedef (`using` alias) for an `std::vector` specialization that we are using already in ITK, to represent a vector of Boolean values.

If you agree that this is a valid style improvement, please approve this PR!

dzenanz (approved):

PR #3188, opened 2022-02-10
STYLE: Let assignment operators return a non-const reference

N-Dekker:
Removed the "const" from `const Self &` return types of assignment
operators.

In accordance with C++ Core Guidelines, January 3, 2022: "Make copy
assignment non-virtual, take the parameter by const&, and return by
non-const&"
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c60-make-copy-assignment-non-virtual-take-the-parameter-by-const-and-return-by-non-const

### DIFF
--- a/Modules/Core/Common/include/itkArray.h
+++ b/Modules/Core/Common/include/itkArray.h
@@ -115,10 +115,10 @@ public:
   }
 
   /** Copy operator */
-  const Self &
+  Self &
   operator=(const Self & rhs);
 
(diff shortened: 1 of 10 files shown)


N-Dekker:
Note that many of the existing assignment operators in ITK already return a non-const `Self &`.

dzenanz (approved):

hjmjohnson (approved):

PR #3189, opened 2022-02-10
DOC: Improve/correct TransformGeometryFilter documentation.

zivy:


dzenanz (approved):

Modules/Core/Transform/include/itkTransformGeometryImageFilter.h
```diff
@@ -26,25 +26,28 @@
...
- * If all the transforms are rigid, there is a far superior way to achieve a similar result.
- * Updating image metadata in-place removes the accumulated resampling errors
- * as well as eliminating the expense of accessing the physical memory of the image.
- * We need to compose all the transforms beforehand to make use of this filter.
- *
- * \param \c RigidTransform Currently must be a VersorRigid3DTransform.
+ * \brief Modify an image's geometric meta-data, changing its "physical" extent.
+ *
+ * The TransformGeometryImageFilter "physically" changes the image in
+ * space using the inverse of the given transformation. The
```
blowekamp, line 32: Saying "inverse of the given transformation" will be confusing to most users. The high level view is that it performs the same transformation as the resample image filter, where each  pixels physical location ti transformed by provided transform. Please say what the filter does not what it does not not do.

blowekamp (commented):
Thank you for cleaning this documentation up.

Modules/Core/Transform/include/itkTransformGeometryImageFilter.h
zivy: done

dzenanz (approved):

blowekamp (approved):
Thank you for accommodating my suggestions.

PR #3191, opened 2022-02-11
COMP: added missing HDF5 symbol mangling

seanm:
Building my own app against both vtk master and itk master results in link errors because both the libraries fail to mangle these 7 symbols.

### DIFF
--- a/Modules/ThirdParty/HDF5/src/itkhdf5/src/itk_hdf5_mangle.h
+++ b/Modules/ThirdParty/HDF5/src/itkhdf5/src/itk_hdf5_mangle.h
@@ -382,6 +382,7 @@ e.g. remove mangling for H5Fget_info but leave it for H5Fget_info1 and H5Fget_in
 #define H5C__verify_cork_tag_test itk_H5C__verify_cork_tag_test
 #define H5C_cache_image_pending itk_H5C_cache_image_pending
 #define H5C_cache_image_status itk_H5C_cache_image_status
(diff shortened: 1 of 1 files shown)


N-Dekker (approved):
Thanks @seanm! Could it be that these HDF5 symbols were introduced with the upgrade from 1.10.7 to 1.12.1, pull request #2832 commit 621b4bd1617b3d56d6ba5ad749512d8a969b21f2 ?

Anyway, it's fine by me. 👍

hjmjohnson (approved):

dzenanz:
All 7 of these symbols were removed by [PR2832](https://github.com/InsightSoftwareConsortium/ITK/pull/2832/commits/f21b8f9eaf7d6b703d168e40de1b286a4b511f4e#diff-ff21a10153dc41fdf05a04941f76a83d7a4b232d3049863f7db64f522d61e58fL482).

seanm:
Yeah, it's hard to catch them all, because you really need to build all combinations of Debug + Release + Mac + UNIX + Window as some symbols exist only conditionally.

hjmjohnson:
PR2832 generated a new set of mangled symbols from a blank slate. Perhaps we should merge the previous manglings with the new manglings (and also merge the VTK identified manglings too).  It probably does not hurt to have a superset of required manglings.

seanm:
>It probably does not hurt to have a superset of required manglings.

Well, it sorta does, because it means you don't actually know how to generate the manglings from scratch and are declaring defeat. :)

PR #3197, opened 2022-02-14
DOC: VLV CanBeDivided struct divides, not multiplies

dzenanz:


Modules/Core/Common/include/itkVariableLengthVector.h
```diff
@@ -1152,7 +1152,7 @@ struct CanBeMultiplied
             mpl::And<mpl::IsNumber<TExpr1>, mpl::IsArray<TExpr2>>>
 {};
 
-/** Tells whether objects from two types can be multiplied.
+/** Tells whether objects from two types can be divided.
```
Leengit, line 1155: The text "divided" does not match the name of the struct `CanBeMultiplied`.  Also, there is no description for this pull request.

Leengit (changes requested):

Modules/Core/Common/include/itkVariableLengthVector.h
dzenanz: This is a docstring for the class which follows it.

I thought that a one-line PR does not need a description.

PR #3201, opened 2022-02-14
COMP: FEM module was not compiling after renames in 4b43536

dzenanz:
4b43536c8b22bb716a4f0b76a0582a99d6d8911d did a lot of renames.
This one was somehow overlooked.
It causes compile errors when Module_ITKFEMRegistration is enabled:
```
96>------ Build started: Project: ITKFEMRegistration, Configuration: Debug x64 ------
96>itkFEMRegistrationFilter.cxx
96>C:\Dev\ITK-git\Modules\Numerics\FEM\include\itkImageToRectilinearFEMObjectFilter.h(72,54): error C2065: 'VDimension': undeclared identifier
96>C:\Dev\ITK-git\Modules\Numerics\FEM\include\itkImageToRectilinearFEMObjectFilter.h(175): message : see reference to class template instantiation 'itk::fem::ImageToRectilinearFEMObjectFilter<TInputImage>' being compiled
96>C:\Dev\ITK-git\Modules\Numerics\FEM\include\itkImageToRectilinearFEMObjectFilter.h(72,1): error C2975: 'VDimension': invalid template argument for 'itk::fem::FEMObject', expected compile-time constant expression
96>C:\Dev\ITK-git\Modules\Numerics\FEM\include\itkFEMObject.h(74): message : see declaration of 'VDimension'
95>ITKIOHDF5.vcxproj -> C:\Dev\ITK-vs19\lib\Debug\ITKIOHDF5-5.3.lib
96>Done building project "ITKFEMRegistration.vcxproj" -- FAILED.
```

PR #3203, opened 2022-02-15
BUG: FEM module has incorrect public member name

Leengit:
Commit 4b43536c8b22bb716a4f0b76a0582a99d6d8911d "STYLE: Change template parameter name prefix N to V" intended to change template parameter names but mistakenly changed a member variable's name.  This commit fixes that.

I don't have permission to push to #3201, so this pull request is intended to replace it.


Leengit:
@dzenanz

thewtex (approved):

PR #3211, opened 2022-02-18
Replace `std::vector<bool>` with `std::unique_ptr<bool[]>`, remove `EnabledArrayType = std::vector<bool>`

N-Dekker:
Replaced `std::vector<bool>` by `std::unique_ptr<bool[]>` in two cases as it appears preferable to avoid using `std::vector<bool>`, when there is a convenient alternative.

In general, `std::vector<bool>` appears troublesome, as it may not prevent data races, and its support for random access may be suboptimal. On the other hand, an `std::unique_ptr<bool[]>` offers fast random access and prevents data races when accessing different elements in the same array.

Follow-up to pull request #3166 commit 915b8e48d8ff7d9f5b943eb673826ad85bf74a8c "BUG: fixed threading bug, don't use vector<bool> from multiple threads" by Sean McBride (merged on 8 February 2022).

Also removed an unused type alias, `EnabledArrayType = std::vector<bool>`.

### DIFF
--- a/Modules/Core/Common/include/itkAnnulusOperator.hxx
+++ b/Modules/Core/Common/include/itkAnnulusOperator.hxx
@@ -21,6 +21,8 @@
 #include "itkMath.h"
 #include "itkSphereSpatialFunction.h"
 
+#include <memory> // For make_unique.
+
 namespace itk
 {
 /** Create the operator */
@@ -105,7 +107,7 @@ AnnulusOperator<TPixel, TDimension, TAllocator>::GenerateCoefficients() -> Coeff
 
   const typename SizeType::SizeValueType w = this->Size();
 
-  std::vector<bool>              outside(w);
+  const auto                     outside = std::make_unique<bool[]>(w);
   CoefficientVector              coeffP(w);
   OffsetType                     offset;
   typename SphereType::InputType point;
--- a/Modules/Core/Common/test/itkTimeStampTest.cxx
+++ b/Modules/Core/Common/test/itkTimeStampTest.cxx
@@ -20,6 +20,7 @@
 #include "itkMultiThreaderBase.h"
 
 #include <iostream>
+#include <memory> // For make_unique.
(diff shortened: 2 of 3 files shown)


Leengit:
Ah, if that works then so might the following.  What you have done is fine, but in case it is useful in your tool chest:

```cpp
#include <iomanip>
#include <iostream>
#include <vector>

class itkBool
{
public:
  // Get value
  operator bool() const { return m_value; }
  // Set value
  itkBool(bool v = false)
    : m_value(v)
  {}
  // itkBool(const itkBool &) = default;
  itkBool &
  operator=(bool v)
  {
    m_value = v;
    return *this;
  }

private:
  bool m_value;
};

int
main()
{
  std::vector<itkBool> myVector;

  itkBool b1{ false }; // set
  itkBool b2{ b1 };    // get and set

  myVector.push_back(true); // set
  bool b = myVector[0];     // get

  myVector[0] = false; // set
  int i = myVector[0]; // get

  bool result0 = (true & myVector[0]); // get
  bool result1 = !myVector[0];         // get
}
```

N-Dekker:
@Leengit Thanks for you suggestion! In these two particular cases (this specific pull request), I think `std::unique_ptr<bool[]>` is preferable anyway, because it is more lightweight that any `std::vector`. Specifically, in these cases, no `push_back` or `resize` was being done, so the buffer was just allocated once. In those cases, the extra overhead of `std::vector` is not necessary.

In other cases, where a `std::vector` of Boolean-like element would be the most appropriate, we did choose to use `std::vector<uint8_t>`, which appears _good enough_ to replace `std::vector<bool>`. It might be nice to use something like your suggested `class itkBool` instead, as in `std::vector<itkBool>`. I just don't know if it's worth the extra maintenance. Assuming that `std::vector<uint8_t>` is already good enough for now.

hjmjohnson (approved):

PR #3212, opened 2022-02-19
COMP: Remove un-maintained wrapping interfaces

hjmjohnson:
REMOVED A LOT OF DEAD CODE.   Only wrapping with Python has worked since at least 2018.

## PR Checklist
- [X] REMOVED JAVA TCL PERL


hjmjohnson:
An ITK discourse topic is located at https://discourse.itk.org/t/remove-java-ruby-tcl-explicit-wrapping-targets/4818

dzenanz (commented):
Looks good at a glance. @thewtex and @blowekamp should pitch in.

hjmjohnson:
@dzenanz NOTE:  If we choose to accept this route of action (i.e. only support python wrapping, as has been the default practice for many many years), then many other simplifications to the wrapping infrastructure are also possible.

blowekamp:
I know some experiments have been done with explicit instantiation in recent years. Do anyone know I they experiments used this explicit instantiation wrapping infrastructure?

hjmjohnson:
@blowekamp Kent Williams did some preliminary work 10 years ago, then a student of mine did a bunch of work attempting to make it work about 4 years ago.  This has never really panned out as a useful exercise with any measurable benefits.

After a lot of effort, I am now convinced that explicit instantiation is not worth exploring further.  I have several huge code patch changes in a dormant branch that automate the process of performing explicit instantiation, but I could not decern any benefits from it.

hjmjohnson:
@thewtex  Matt, do you have any initial thoughts.  This PR represents a go-no-go situation for how we proceed with fixing (untangling) the python wrapping. 

Hans

thewtex (approved):
:+1: thanks, Hans!

thewtex:
> Do anyone know if the experiments used this explicit instantiation wrapping infrastructure?

I also did these experiments with the wrapping infrastructure. In practice, there was not a build speed-up. I do not think there is a reason to keep this code around.

I have work in progress for C++20 Module Header-Units, which I am more hopeful about because the compiler is able to re-use AST representations.

PR #3214, opened 2022-02-21
ENH: Introduce itk::ThreadSafeBooleanType for use with std::vector

Leengit:
`std::vector<bool>` packs multiple bits into a byte which hampers multithreaded operations.  This pull request introduces `itk::ThreadSafeBooleanType`, which has the storage size of `struct { bool value; }`.  `itk::ThreadSafeBooleanType` acts like `bool` except that specialized templates for `bool` will not be specialized for `itk::ThreadSafeBooleanType`.  In particular, `std::vector<itk::ThreadSafeBooleanType>` will not pack multiple bits into a single byte.

Additionally, `itk::ThreadSafeBooleanType` replaces `uint8_t` in the definition 
```cpp
using BooleanStdVectorType = std::vector<uint8_t>;
```
This means that, e.g., 
```cpp
uint_8 rhs = 3;
BooleanStdVectorType v;
v.push_back(rhs);
uint_8 result = v[0];
```
will now return `result=1` rather than `result=3`.

Thank you to @N-Dekker for good ideas.

## PR Checklist


dzenanz (approved):

N-Dekker:
Sorry @Leengit  but it still seems to me that it is a bit _overdone_, to introduce our own ITK Boolean class, as it appears that in practice, `std::vector<uint8_t>` (with a simple type alias `BooleanStdVectorType`) is good enough. As I commented before at https://github.com/InsightSoftwareConsortium/ITK/pull/3211#issuecomment-1045302342 But it's a nice C++ exercise.

Is your plan to then replace all `bool` in ITK by `itk::ThreadSafeBooleanType`?

dzenanz:
> replace all `bool`

That seems counterproductive. Adding another layer of abstraction has a computational cost. It makes sense only in multi-threaded array accesses.

N-Dekker:
> Adding another layer of abstraction has a computational cost. It makes sense only in multi-threaded array accesses.

It seems only relevant for `std::vector`. Other STL containers work just fine with `bool` as element type, including `std::array` and `std::deque`. And `std::unique_ptr<bool[]>` also works fine, of course.

Leengit:
Thank you for the feedback @N-Dekker.  I submitted the pull request because I thought it was a net win -- but if you don't agree then that influences my opinion.

It is not my plan to replace all `bool` in ITK, nor even the subset of all `bool` used as parameters to templates that might otherwise be specialized.  My goals are two-fold: (1) to provide the means for others who need this functionality, and (2) to get `result=1` in the above code snippet.

N-Dekker:
> I submitted the pull request because I thought it was a net win -- but if you don't agree then that influences my opinion.

I find it hard to estimate whether it is a net win or loss, honestly. In general, when the benefit is not so clear, I prefer not to add extra code.

> It is not my plan to replace all `bool` in ITK, nor even the subset of all `bool` used as parameters to templates that might otherwise be specialized. My goals are two-fold: (1) to provide the means for others who need this functionality, and (2) to get `result=1` in the above code snippet.

Thanks for your explanation @Leengit So can you promise that you won't replace all `bool` in ITK afterwards?  😃

Leengit:
> Thanks for your explanation @Leengit So can you promise that you won't replace all `bool` in ITK afterwards? smiley

Hmm.  Would I still be free to replace all but one?  :smiley:

dzenanz:
If you want to replace just one, you could a new commit to this PR, thus logically grouping that change here.

Modules/Core/Common/include/itkThreadSafeBooleanType.h
```diff
@@ -0,0 +1,63 @@
...
+/** The type alias `ThreadSafeBooleanType` provides an alternative to `bool`. `std::vector<bool>` is not thread safe due
+ * to the possibility of multiple bits being packed together in the same memory location.
+ * std::vector<ThreadSafeBooleanType> does not have such a space optimization. std::vector<ThreadSafeBooleanType> is
+ * semantically like `std::vector<bool>`, but unlike `std::vector<bool>`, it does "avoid data races when the contents of
+ * the contained object in different elements in the same container [...] are modified concurrently", according to the
+ * C++ Standard, section [container.requirements.dataraces], "Container data races".
+ *
+ */
+
+class ThreadSafeBooleanType
```
N-Dekker, line 33: The class isn't entirely thread-safe (it isn't atomic), which is OK, because it just works around the infamous `std::vector<bool>` problem. However, I then find the name of the class "ThreadSafeBooleanType" too pretentious. I would prefer to just call it `Boolean`.

Note that it may also yield a performance gain in single-threaded code, as random access to `std::vector<itk::Boolean>` would probably be faster than random access to `std::vector<bool>`

Modules/Core/Common/include/itkThreadSafeBooleanType.h
```diff
@@ -0,0 +1,63 @@
...
+  constexpr ThreadSafeBooleanType &
+  operator=(const ThreadSafeBooleanType & v) = default;
+
+  constexpr ThreadSafeBooleanType &
+  operator=(bool v)
+  {
+    m_value = v;
+    return *this;
+  }
+
```
N-Dekker, line 53: Is it really necessary to explicitly declare those two assignment operators? Usually the compiler would automatically generate assignment operators implicitly in such a case. See also "If you can avoid defining default operations, do" https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c20-if-you-can-avoid-defining-default-operations-do

Modules/Core/Common/include/itkThreadSafeBooleanType.h
```diff
@@ -0,0 +1,63 @@
...
+ * the contained object in different elements in the same container [...] are modified concurrently", according to the
+ * C++ Standard, section [container.requirements.dataraces], "Container data races".
+ *
+ */
+
+class ThreadSafeBooleanType
+{
+public:
+  // Constructors
+  constexpr ThreadSafeBooleanType(const ThreadSafeBooleanType &) = default;
```
N-Dekker, line 37: Is it really necessary to explicitly declare the copy-constructor? See also (just like for the assignment operators) "If you can avoid defining default operations, do" https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c20-if-you-can-avoid-defining-default-operations-do

Modules/Core/Common/include/itkBooleanStdVector.h
```diff
@@ -19,6 +19,7 @@
 #define itkBooleanStdVector_h
 
 #include "itkIntTypes.h"
```
N-Dekker, line 21: `#include "itkIntTypes.h"` was only there for `uint8_t`, so if you want to replace it by `ThreadSafeBooleanType`, you might as well remove `#include "itkIntTypes.h"` !

Modules/Core/Common/include/itkThreadSafeBooleanType.h
Leengit: Should we use `itk::BooleanType`?

Modules/Core/Common/include/itkBooleanStdVector.h
```diff
@@ -34,7 +35,8 @@ namespace itk
  * \note BooleanStdVectorType is only intended to be used for the storage of Boolean values (true and false), not for
  * arbitrary integer values.
  */
-using BooleanStdVectorType = std::vector<uint8_t>;
+
```
N-Dekker, line 38: The documentation (above here) is specifically about `BooleanStdVectorType`, so I'm not sure if it's handy to add an extra white line between the documentation and the definition of `BooleanStdVectorType`.

Modules/Core/Common/include/itkThreadSafeBooleanType.h
Leengit: Removing them does work on my platform.  I'll push and see what GitHub actions think.

Modules/Core/Common/include/itkThreadSafeBooleanType.h
Leengit: Removing it does work on my platform.  I'll push and see what GitHub actions think.

Modules/Core/Common/include/itkBooleanStdVector.h
Leengit: Good catch.  Thank you.

Modules/Core/Common/include/itkThreadSafeBooleanType.h
seanm: I like `ThreadSafeBooleanType`. It gives a good indication of what this is for. Just `itk::BooleanType` is so generic, it tells nothing.

Modules/Core/Common/include/itkBooleanStdVector.h
N-Dekker: Will Doxygen then still understand that they belong together?

Leengit: Will fix.

Modules/Core/Common/include/itkThreadSafeBooleanType.h
```diff
@@ -0,0 +1,63 @@
...
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+#ifndef itkThreadSafeBooleanType_h
+#define itkThreadSafeBooleanType_h
+
+namespace itk
+{
+
+/** The type alias `ThreadSafeBooleanType` provides an alternative to `bool`. `std::vector<bool>` is not thread safe due
```
N-Dekker: Actually it's not a type alias. It's a `class`! So maybe it should not end with `Type` anyway. I think in ITK, the postfix `Type` is (usually) only used for typedef's (type aliases), and `using` declarations thereof.

Modules/Core/Common/include/itkBooleanStdVector.h
N-Dekker: Almost running out of nitpicks  ;-)

Modules/Core/Common/include/itkThreadSafeBooleanType.h
Leengit: I just pushed it as `itk::Boolean` but we could still change it.  Let me know.

Modules/Core/Common/include/itkThreadSafeBooleanType.h
Leengit: Will fix.

Modules/Core/Common/include/itkBoolean.h
```diff
@@ -0,0 +1,49 @@
...
+ * semantically like `std::vector<bool>`, but unlike `std::vector<bool>`, it does "avoid data races when the contents of
+ * the contained object in different elements in the same container [...] are modified concurrently", according to the
+ * C++ Standard, section [container.requirements.dataraces], "Container data races".
+ *
+ */
+
+class Boolean
+{
+public:
+  constexpr Boolean(bool v = false)
```
N-Dekker, line 36: I see your proposed Boolean class is different from `bool`, as a declared local `bool` may be uninitialized, while your proposal would make a default-constructed `Boolean` false. That's an intended difference, right? Just to be sure!

    bool x; // uninitialized;  
    itk::Boolean y; // false

Leengit: It seems safer this way, but I am open to suggestions.

Leengit:
The most recent push includes Doxygen keywords.

Modules/Core/Common/include/itkBoolean.h
```diff
@@ -0,0 +1,54 @@
...
+    : m_value(v)
+  {}
+
+  // Get value
+  constexpr operator bool() const { return m_value; }
+
+private:
+  bool m_value = false;
```
N-Dekker, line 49: Maybe `m_Value`, instead of `m_value`?

Leengit: Will do.

Modules/Core/Common/include/itkBoolean.h
N-Dekker: > It seems safer this way, but I am open to suggestions.

It's OK to me. Just want to be sure that it's an intended difference between itk::Boolean and `bool`!

N-Dekker: Of course `{}` does the job for both of them:

    bool x2{}; // false  
    itk::Boolean y2{}; // false

N-Dekker: Most relevantly, both `std::vector<bool>(n)` and `std::vector<itk::Boolean>(n)` represent `n` elements, initialized by `false`. So that's also OK!

Modules/Core/Common/include/itkBoolean.h
```diff
@@ -0,0 +1,54 @@
...
+#ifndef itkBoolean_h
+#define itkBoolean_h
+
+namespace itk
+{
+
+/**
+ * \class Boolean
+ *
+ * \brief An alternatitve to bool for addtional thread safety
```
N-Dekker, line 27: Two more typo's here: "alternatitve" and "addtional". But no big deal, as you removed so many typo's already, recently. 😄

Leengit: Fixed with the most recent push.

dzenanz (approved):
This should be squashed upon, or before merging.

Modules/Core/Common/include/itkBoolean.h
```diff
@@ -0,0 +1,54 @@
...
+ * \ingroup ITKCommon
+ */
+class Boolean
+{
+public:
+  constexpr Boolean(bool v = false)
+    : m_Value(v)
+  {}
+
+  // Get value
```
N-Dekker, line 45: Maybe some more documentation for this operator, saying something like:

    /** Supports implicit conversion to `bool`. */

Modules/Core/Common/include/itkBoolean.h
```diff
@@ -0,0 +1,54 @@
...
+ * "avoid data races when the contents of the contained object in different elements in the same container [...] are
+ * modified concurrently", according to the C++ Standard, section [container.requirements.dataraces], "Container data
+ * races".
+ *
+ * \ingroup ITKCommon
+ */
+class Boolean
+{
+public:
+  constexpr Boolean(bool v = false)
```
N-Dekker, line 41: Maybe some documentation for this constructor, saying something like:

    /** Supports default-construction and implicit conversion from `bool`. */

N-Dekker (approved):
@Leengit  Thanks for addressing my comments. Please do squash the two commits, as already requested by @dzenanz Otherwise approved  👍

Modules/Core/Common/include/itkBoolean.h
N-Dekker: No show-stopper, but did you consider a separate (explicitly defaulted) default-constructor? As in:

    Boolean() = default;

    /** Supports implicit conversion from `bool`. */
    constexpr Boolean(bool v)
        ....

One of the advantages could be that a defaulted default-constructor is basically self-documenting (whereas a conversion operator may need some documentation).  The compiler takes care of "constexpr-ness", for defaulted member functions.

Such a _defaulted_ default-constructor would then use the in-class default member initialization that you already added (but did not yet use):

    bool m_Value = false;

Leengit:
The build indicates: `ITK type not wrapped, or currently not known: itk::Boolean`
(Longer version below)

Should we wrap `itk::Boolean`?  Alternatively, how would we go about telling the build that it is okay that we are not wrapping it?

[3082/5681] Generating ../../Typedefs/itkDenseFiniteDifferenceImageFilter.i, ../../Typedefs/itkFiniteDifferenceFunction.i, ../../Typedefs/itkFiniteDifferenceImageFilter.i, ../../Typedefs/itkDenseFiniteDifferenceImageFilterSwigInterface.h, ../../Typedefs/itkFiniteDifferenceFunctionSwigInterface.h, ../../Typedefs/itkFiniteDifferenceImageFilterSwigInterface.h, ../../Typedefs/itkDenseFiniteDifferenceImageFilter.idx, ../../Typedefs/itkFiniteDifferenceFunction.idx, ../../Typedefs/itkFiniteDifferenceImageFilter.idx, ../../Generators/Python/itk/Configuration/ITKFiniteDifference_snake_case.py, ../../Generators/Python/itk-stubs/itkDenseFiniteDifferenceImageFilterTemplate.pyi, ../../Generators/Python/itk-stubs/itkDenseFiniteDifferenceImageFilterProxy.pyi, ../../Generators/Python/itk-stubs/itkFiniteDifferenceFunctionTemplate.pyi, ../../Generators/Python/itk-stubs/itkFiniteDifferenceFunctionProxy.pyi, ../../Generators/Python/itk-stubs/itkFiniteDifferenceImageFilterTemplate.pyi, ../../Generators/Python/itk-stubs/itkFiniteDifferenceImageFilterProxy.pyi
itkFiniteDifferenceImageFilter: warning(4): ITK type not wrapped, or currently not known: itk::Boolean

Modules/Core/Common/include/itkBoolean.h
Leengit: Will do.

dzenanz:
> itkFiniteDifferenceImageFilter: warning(4): ITK type not wrapped, or currently not known: itk::Boolean

`FiniteDifferenceImageFilter` needs it, so we should wrap it.

dzenanz (approved):

Leengit:
This most recent push does the squash, incorporates @N-Dekker's most recent suggestions, and wraps `itk::Boolean`.  Let's see what the GitHub actions think.

N-Dekker:
Another suggestion for the class name could be `BoolHolder`, as it just _holds_ a `bool`. This could make it clearer to other users that it isn't intended to entirely replace `bool` for Boolean values in ITK's code base. It's just a little "holder class" to bypass the infamous `std::vector<bool>` specialization.

Just a suggestion, not a change request 😃

blowekamp:
This class does not address the [False Sharing](https://en.wikipedia.org/wiki/False_sharing) problem when indexing by threadID. An array indexed by threads is an anti-pattern.

One possible option, would be to pad each element so that the size is equal to the cache line of the processor. This already done in a few cases. However, these cases may have aggregated multiple arrays of threads IDs to a single array of padded struct type.

Leengit:
Would someone with the authority and knowledge of the GitHub incantation please re-launch the ITK.macOS.Python check?  The error message is:

##[error]We stopped hearing from agent Hosted Agent. Verify the agent machine is running and has a healthy network connection. Anything that terminates an agent process, starves it for CPU, or blocks its network access can cause this error. For more information, see: https://go.microsoft.com/fwlink/?linkid=846610

N-Dekker:
@Leengit It appears that your `itk::Boolean` supports implicit conversion to and from `int`. Is that your intention?  Example:

    int i = itk::Boolean();
    itk::Boolean b = int();

At: https://godbolt.org/z/h8G33xePd

Actually, conversions to and from any other arithmetic type appears supported as well now!

If you would like to prevent such conversions, you might add explicitly _deleted_ conversion member functions, like:

    /** Prevents implicit conversion from arithmetic types other than `bool`. */
    Boolean(int) = delete;

    /** Prevents implicit conversion to arithmetic types other than `bool`. */
    operator int() const = delete;

As you can see at https://godbolt.org/z/Wx4G33Yfo

dzenanz:
> itk::Boolean supports implicit conversion to and from int

I think bool does that, so this is desirable?

N-Dekker:
> I think bool does that, so this is desirable?

The currently proposed `itk::Boolean` is already more _safe_ than `bool`, as it is initialized (to `false`) automatically when it is declared:

    bool b; // Unsafe: uninitialized!
    itk::Boolean itkBoolean; // More safe: initialized to false

Preventing implicit conversion between `itk::Boolean`  and non-bool arithmetic types would make it even more safe.

No show-stopper, just something to be considered. Maybe for the next PR  😸

hjmjohnson (approved):

PR #3221, opened 2022-02-23
STYLE: Rename `ColorTable` template parameter `TPixel` to `TComponent`

N-Dekker:
Aims to avoid confusion, as the pixel type in this context is
`RGBPixel<TComponent>`, which is _not_ the template parameter of `ColorTable`.

### DIFF
--- a/Modules/Core/Common/include/itkColorTable.h
+++ b/Modules/Core/Common/include/itkColorTable.h
@@ -36,7 +36,7 @@ namespace itk
  * \ingroup ITKCommon
  */
 
-template <typename TPixel>
+template <typename TComponent>
 class ITK_TEMPLATE_EXPORT ColorTable : public Object
 {
 public:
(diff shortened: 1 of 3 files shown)


dzenanz (approved):

N-Dekker:
Thanks for approving this PR, @dzenanz and @hjmjohnson I think it's ready, too. FYI Here is where I think the pixel type and the component type get confused:

     TPixel zeroComponent(0); 
     pixel = colors->GetColorComponent(numberOfColors, rgb); 
     if (pixel != zeroComponent) 

At the current "master" revision: https://github.com/InsightSoftwareConsortium/ITK/blob/a837e7b5c7bce6face0956cd0ea1a3bfef899fed/Modules/Core/Common/test/itkColorTableTest.cxx#L110-L112

In this case, the type of `zeroComponent`  is `TPixel`, but the type of `pixel` is actually `itk::RGBPixel<TPixel>`! So the `if (pixel != zeroComponent)` looks somewhat funny! 

I'm intending to resolve this confusion by another PR, once this one is merged.

PR #3224, opened 2022-02-23
STYLE: Avoid `T var = elementValue` syntax to initialize a filled array

N-Dekker:
The C++ copy-initialization syntax `T var = value` suggests that the
value at the right side is implicitly converted to the target type `T`.

When the target type `T` is a `FixedArray<TValue, VLength>`, or one of
its derived types, `T var = elementValue` _fills_ the variable with the
specified `elementValue`, by assigning the `elementValue` to each of its
elements. In this case, the C++ direct-initialization syntax
`T var(elementValue)` appears more appropriate.

Occurrences of `T var = elementValue` were found by locally temporarily
declaring the corresponding constructors `explicit`.


Leengit (approved):
Looks good.

Issue #3222, opened 2022-02-23
WIP: BUG: `ContourSpatialObject<TDimension>::Update()` looks suspicious


N-Dekker:
There is a line of code in `ContourSpatialObject<TDimension>::Update()` that looks suspicious to me:

    newPoint = pnt[d] + i * step[d]; 

https://github.com/InsightSoftwareConsortium/ITK/blob/a2b39155dae42fb88d52c0d7d7d7e1f1a8c3e954/Modules/Core/SpatialObjects/include/itkContourSpatialObject.hxx#L217-L225

You see, `newPoint` is a `PointType`, and it gets filled by a scalar value, `pnt[d] + i * step[d]`, over and over again, inside the inner for loop.

I think the intention was to do: 

    newPoint[d] = pnt[d] + i * step[d]; 
 
But I'm not sure, because I did not study the code well enough yet! Any feedback or possible explanation is welcome!

_Note: This bug report is still in progress! More details soon, hopefully!_

dzenanz:
@aylward might have some comment.

aylward:
You're right!   Great catch!    Want to submit a pull request to fix it?

N-Dekker:
> Great catch! Want to submit a pull request to fix it?

Thanks @aylward but I don't know how to fix it properly. Is it really a serious bug?  Or is this part of the code not so important anyway?

If it is a serious bug, can you please tell me the intended behavior?

N-Dekker:
The apparent bug is part of the `case InterpolationMethodEnum::LINEAR_INTERPOLATION` section of a `switch (m_InterpolationMethod)`. It appears that this specific "case" is not tested, at least not by [itkContourSpatialObjectTest.cxx](https://github.com/InsightSoftwareConsortium/ITK/blob/a2b39155dae42fb88d52c0d7d7d7e1f1a8c3e954/Modules/Core/SpatialObjects/test/itkContourSpatialObjectTest.cxx). If it is indeed untested, and if it is also unused, I would prefer to just put some `itkExceptionMacro(<< "not yet defined")` call there, just like in `case InterpolationMethodEnum::BEZIER_INTERPOLATION`. Then postpone actual implementation until it is really needed. Would that be OK to you?

https://github.com/InsightSoftwareConsortium/ITK/blob/a2b39155dae42fb88d52c0d7d7d7e1f1a8c3e954/Modules/Core/SpatialObjects/include/itkContourSpatialObject.hxx#L183-L193

[...]

PR #3225, opened 2022-02-23
BUG: Quick-fix ContourSpatialObject::Update(), LINEAR_INTERPOLATION case

N-Dekker:
`ContourSpatialObject::Update()` appears to have an assignment to
`newPoint`, in the `case InterpolationMethodEnum::LINEAR_INTERPOLATION`
section, which was meant to assign just a single element `newPoint[d]`.

As was confirmed by Stephen Aylward (@aylward) at issue https://github.com/InsightSoftwareConsortium/ITK/issues/3222,
"`ContourSpatialObject<TDimension>::Update()` LINEAR_INTERPOLATION case
may need some adjustment".

Some more adjustment may still be needed, as `newPoint` now still gets
modified multiply times (rather than just once), before it eventually
gets used.


aylward (approved):
LGTM!

PR #3231, opened 2022-02-25
BUG: Fix code that treated `point1 - point2` as Point rather than Vector

N-Dekker:
An subtraction of the form `point1 - point2` yields a `Vector`, not a `Point`. Adjusted both `TransformToDisplacementFieldFilter::NonlinearThreadedGenerateData` and `RayCastInterpolateImageFunction::Evaluate` accordingly.

These two little flaws were found by locally (temporarily) declaring converting constructors of `Point` "explicit".

### DIFF
--- a/Modules/Core/ImageFunction/include/itkRayCastInterpolateImageFunction.hxx
+++ b/Modules/Core/ImageFunction/include/itkRayCastInterpolateImageFunction.hxx
@@ -1416,10 +1416,17 @@ RayCastInterpolateImageFunction<TInputImage, TCoordRep>::Evaluate(const PointTyp
   ray.ZeroState();
   ray.Initialise();
 
-  PointType origin = this->m_Image->GetOrigin();
-  for (unsigned int i = 0; i < origin.Length; ++i)
-    origin[i] -= 0.5 * this->m_Image->GetSpacing()[i];
-  ray.SetRay(point - origin, direction);
+  const PointType origin = this->m_Image->GetOrigin();
+  const auto      spacing = this->m_Image->GetSpacing();
+
+  PointType rayPosition = point;
+
+  for (unsigned int i = 0; i < PointType::Length; ++i)
+  {
+    rayPosition[i] -= origin[i] - 0.5 * spacing[i];
+  }
+
+  ray.SetRay(rayPosition, direction);
   ray.IntegrateAboveThreshold(integral, m_Threshold);
 
   return (static_cast<OutputType>(integral));
--- a/Modules/Filtering/DisplacementField/include/itkTransformToDisplacementFieldFilter.hxx
+++ b/Modules/Filtering/DisplacementField/include/itkTransformToDisplacementFieldFilter.hxx
@@ -185,7 +185,6 @@ TransformToDisplacementFieldFilter<TOutputImage, TParametersValueType>::Nonlinea
   // to an output pixel
   PointType outputPoint;       // Coordinates of output pixel
   PointType transformedPoint;  // Coordinates of transformed pixel
-  PointType displacementPoint; // the difference
   PixelType displacementPixel; // the difference, cast to pixel type
 
 
@@ -203,11 +202,11 @@ TransformToDisplacementFieldFilter<TOutputImage, TParametersValueType>::Nonlinea
       // Compute corresponding input pixel position
       transformedPoint = transform->TransformPoint(outputPoint);
 
-      displacementPoint = transformedPoint - outputPoint;
+      const typename PointType::VectorType displacementVector = transformedPoint - outputPoint;
       // Cast PointType -> PixelType
       for (IndexValueType idx = 0; idx < ImageDimension; ++idx)
       {
-        displacementPixel[idx] = static_cast<typename PixelType::ValueType>(displacementPoint[idx]);
+        displacementPixel[idx] = static_cast<typename PixelType::ValueType>(displacementVector[idx]);
       }
       outIt.Set(displacementPixel);
       ++outIt;


dzenanz (approved):

Modules/Core/ImageFunction/include/itkRayCastInterpolateImageFunction.hxx
```diff
@@ -1416,10 +1416,17 @@ RayCastInterpolateImageFunction<TInputImage, TCoordRep>::Evaluate(const PointTyp
...
-    origin[i] -= 0.5 * this->m_Image->GetSpacing()[i];
-  ray.SetRay(point - origin, direction);
+  const PointType origin = this->m_Image->GetOrigin();
+  const auto      spacing = this->m_Image->GetSpacing();
+
+  PointType rayPosition = point;
+
+  for (unsigned int i = 0; i < PointType::Length; ++i)
+  {
+    rayPosition[i] += 0.5 * spacing[i] - origin[i];
```
Leengit, line 1426: Subtracting a point `origin` from a vector `spacing` is a little weird mathematically.  How about instead subtracting a vector from a point:
```cpp
rayPosition[i] -= origin[i] - 0.5 * spacing[i];
```

Modules/Filtering/DisplacementField/include/itkTransformToDisplacementFieldFilter.hxx
```diff
@@ -203,11 +202,11 @@ TransformToDisplacementFieldFilter<TOutputImage, TParametersValueType>::Nonlinea
       // Compute corresponding input pixel position
       transformedPoint = transform->TransformPoint(outputPoint);
 
-      displacementPoint = transformedPoint - outputPoint;
+      const auto displacementVector = transformedPoint - outputPoint;
```
Leengit, line 205: The type isn't apparent from the right-hand side, so I'd put `VectorType` here instead of `auto`, where `using VectorType = ...` should be defined in the places where `PointType` is defined, and ultimately is `itk::Vector<double, 3>` in this case.

Modules/Core/ImageFunction/include/itkRayCastInterpolateImageFunction.hxx
```diff
@@ -1416,10 +1416,17 @@ RayCastInterpolateImageFunction<TInputImage, TCoordRep>::Evaluate(const PointTyp
   ray.ZeroState();
   ray.Initialise();
 
-  PointType origin = this->m_Image->GetOrigin();
-  for (unsigned int i = 0; i < origin.Length; ++i)
-    origin[i] -= 0.5 * this->m_Image->GetSpacing()[i];
-  ray.SetRay(point - origin, direction);
+  const PointType origin = this->m_Image->GetOrigin();
+  const auto      spacing = this->m_Image->GetSpacing();
```
Leengit, line 1420: The type isn't apparent from the right-hand side, so I'd put `SpacingType` here instead of `auto`.

Leengit (approved):
Looks good.  I made suggestions, but they are optional.

Modules/Core/ImageFunction/include/itkRayCastInterpolateImageFunction.hxx
N-Dekker: @Leengit Thanks for the suggestion. I wouldn't mind following your suggestion, but don't you agree that _in general_ the form

    a += b - c;

Is clearer than the equivalent:

    a -= c - b;

?

Modules/Core/ImageFunction/include/itkRayCastInterpolateImageFunction.hxx
N-Dekker: @Leengit Not entirely convinced. For a variable that is already named `spacing`, initialized by `GetSpacing()`, to explicitly specify that its type is `SpacingType` doesn't seem very informative to me.  🤔

Modules/Core/ImageFunction/include/itkRayCastInterpolateImageFunction.hxx
N-Dekker: @Leengit Update: amended the pull request according to your suggestion regarding `rayPosition[i] +/-= ...`

Modules/Filtering/DisplacementField/include/itkTransformToDisplacementFieldFilter.hxx
N-Dekker: @Leengit I would rather not add extra type aliases with this pull request. As a compromise, I just replaced this `auto` by `typename PointType::VectorType`, as I see that being used in similar cases, in the same source file. Hope it's OK to you!

hjmjohnson (approved):

Modules/Filtering/DisplacementField/include/itkTransformToDisplacementFieldFilter.hxx
Leengit: Looks good.

Leengit:
Perhaps for a later pull request, would it make sense to allow `point - vector`, but disallow `vector - point`?  Perhaps it would break external code, in which case temporarily setting this would enable us to find and pretty up our own code, but not much more.

N-Dekker:
> Perhaps for a later pull request, would it make sense to allow `point - vector`, but disallow `vector - point`?

@Leengit Thanks for the suggestion, but isn't that the current situation already?

Leengit:
>  isn't that the current situation already?

Because an early version of this pull request used `rayPosition[i] += 0.5 * spacing[i] - origin[i]`, I had assumed that the right-hand side compiles; I didn't actually check though.  I propose that that right-hand side shouldn't compile (modulo the setting of ITK_LEGACY_REMOVE) because it is `vector - point` rather than `point - vector`.

N-Dekker:
> Because an early version of this pull request used `rayPosition[i] += 0.5 * spacing[i] - origin[i]`, I had assumed that the right-hand side compiles

It did, definitely. But in this case, spacing[i] and origin[i] were just scalars (typically floating point values), not `Vector`  or `Point` objects. So it should just work fine  😄  Otherwise, please present a more detailed example to demonstrate your point.

Leengit:
Now I see, @N-Dekker; thank you for the clarification.

Issue #3241, opened 2022-03-01
Accidental itk::NiftiImageIO metadata behaviour change wrt `pixDim[0]`


seanm:
In f38b1dd404d5a888b063bdfc023195a69d5212d7 I accidentally changed some behaviour.  Specifically, the value of `pixdim[0]` in `itk::NiftiImageIO` metadata is different before and after the change.

Previous to the change, `SetImageIOMetadataFromNIfTI()` was populated from the `nifti_1_header` structure.  After the change, it is populated from the `nifti_image` structure.  The latter is created from the former with `nifti_convert_nhdr2nim()`.  Most of the fields used in `SetImageIOMetadataFromNIfTI()` have 1-to-1 correspondences in these two structures.  I expected all 8 elements of the `pixdim` array were the same in both structures, but in fact no; element 0 is special.  `nifti_convert_nhdr2nim()` unconditionally sets `pixdim[0]` to 0.0 (by virtue of allocating the structure with `calloc`).

In my own app, I was using `pixdim[0]` (retrieved from `itk::NiftiImageIO`) as the last parameter to `nifti_quatern_to_mat44()`.  I see now that I should be passing `qfac` as the last parameter.  But the `itk::MetaDataObject` doesn't include it, so my first proposal here is to start including it.

Then I got to thinking that me even needing to call `nifti_quatern_to_mat44()` in my app is really a layering violation (leaking ITK's use of niftilib) and that it'd be nice if `qto_xyz` were also part of the metadata, so I added that too.

There remains the question of whether to actually revert my accidental behaviour change. There are pros and cons.

- it's been changed since at least ITK 5.0a1, so people may already be expecting this new behaviour as much as others are expecting the old behaviour.
- `pixdim[0]` is now always 0.0 and returning something that never changes is kinda silly.
- everything in `SetImageIOMetadataFromNIfTI()` now comes from `nifti_image`, changing one field to be from `nifti_1_header` would be kinda weird. Now that I review all this, there are other fields that can be subtly different between the structs, even when they have identical names.  For example `nifti_convert_nhdr2nim()` does things like this: `nim->intent_p1 = FIXED_FLOAT( nhdr.intent_p1 ) ;`

Given that this change is now 4 years old, I'm inclined to keep this new/current behaviour but:
- document the behaviour change in some release notes (where?)
- add `qfac` to the metadata
- add `qto_xyz` to the metadata

seanm:
[...]

PR #3242, opened 2022-03-01
ENH: bug #3241: Add qfac and qto_xyz to itkNiftiImageIO metadata

seanm:
bug #3241: In f38b1dd I accidentally changed some behaviour. Specifically, the value of pixdim[0] in itk::NiftiImageIO metadata is different before and after the change.

Previous to the change, SetImageIOMetadataFromNIfTI() was populated from the nifti_1_header structure. After the change, it is populated from the nifti_image structure. The latter is created from the former with nifti_convert_nhdr2nim(). Most of the fields used in SetImageIOMetadataFromNIfTI() have 1-to-1 correspondences in these two structures. I expected all 8 elements of the pixdim array were the same in both structures, but in fact no; element 0 is special. nifti_convert_nhdr2nim() unconditionally sets pixdim[0] to 0.0 (by virtue of allocating the structure with calloc).

In my own app, I was using pixdim[0] (retrieved from itk::NiftiImageIO) as the last parameter to nifti_quatern_to_mat44(). I see now looking at niftilib itself that I should be passing qfac as the last parameter.  But the itk::MetaDataObject doesn't include it, so this commit adds qfac to the metadata dictionary.

Additionally, we now include the whole qto_xyz matrix, so that ITK clients that call nifti_quatern_to_mat44() don't even need to.  This plugs a leaky abstraction and allows my own app for example to not even know that ITK is implemented via niftilib, and never need to call down to it.

Note that the behaviour change of f38b1dd is not being reverted, because it's been four years now, no one else has noticed, and some people may be relying on this behaviour now.

### DIFF
--- a/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
+++ b/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
@@ -994,6 +994,20 @@ NiftiImageIO::SetImageIOMetadataFromNIfTI()
   std::ostringstream intent_name;
   intent_name << nim->intent_name;
   EncapsulateMetaData<std::string>(thisDic, "intent_name", intent_name.str());
+
+  // The below were added after ITK 5.3rc2.
+
+  EncapsulateMetaData<float>(thisDic, "qfac", nim->qfac);
+
+  std::vector<float> qto_xyz;
+  for (int row = 0; row < 4; ++row)
+  {
+    for (int column = 0; column < 4; ++column)
+    {
+      qto_xyz.push_back(nim->qto_xyz.m[row][column]);
+    }
+  }
+  EncapsulateMetaData<std::vector<float>>(thisDic, "qto_xyz", qto_xyz);
 }
 
 void


Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -994,6 +994,23 @@ NiftiImageIO::SetImageIOMetadataFromNIfTI()
...
   EncapsulateMetaData<std::string>(thisDic, "intent_name", intent_name.str());
+
+  // The below since ITK 5.3
+  std::ostringstream qfac;
+  qfac << nim->qfac;
+  EncapsulateMetaData<std::string>(thisDic, "qfac", qfac.str());
+
+  std::ostringstream qto_xyz;
+  qto_xyz <<
+  nim->qto_xyz.m[0][0] << " " << nim->qto_xyz.m[0][1] << " " <<
```
dzenanz, line 1005: Linter complains here. Please apply `clang-format`.

dzenanz (commented):
LGTM

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -994,6 +994,23 @@ NiftiImageIO::SetImageIOMetadataFromNIfTI()
   std::ostringstream intent_name;
   intent_name << nim->intent_name;
   EncapsulateMetaData<std::string>(thisDic, "intent_name", intent_name.str());
+
+  // The below since ITK 5.3
+  std::ostringstream qfac;
+  qfac << nim->qfac;
+  EncapsulateMetaData<std::string>(thisDic, "qfac", qfac.str());
```
N-Dekker, line 1001: If you would want to have lossless conversion of those float values to string, you might consider using `itk::NumberToString`, instead of `std::ostringstream`, for example:

    EncapsulateMetaData<std::string>(thisDic, "qfac",  NumberToString<float>{}(qfac));


Just my 2 cents, Niels

seanm: Lossless would indeed be better.  Searching the codebase, I see things like `itk::EncapsulateMetaData<double>(metaDict, "TestDouble", metaDataDouble);` elsewhere... I was mostly copying nearby code, I wonder why itkNiftiImageIO is only using strings...?

N-Dekker: @seanm I understand if you'd rather not directly change the way NiftiImageIO currently does float-to-string, if your intention is just to add some extra data. No problem, maybe lossless float-to-string conversion could still be added somewhere in the future.  😃

seanm: I'm just questioning why it needs to be a string at all.  Seems `itk::EncapsulateMetaData` allows other types too... Then again, everything else is already a string, so probably best to just follow the existing pattern.

dzenanz: @blowekamp might have some suggestions about metadata encapsulation.

hjmjohnson: @seanm I think it should be converted to a number if it is a number.  I don't know why it would have been converted to a string.  I also prefer `NumberToString{}(qfac)` as the most lossless option.

I am making an issue to track the necessary update to make these conversions lossless in the meta data storage.

seanm: So I just tried locally, and this works:

```
  EncapsulateMetaData<float>(thisDic, "qfac", nim->qfac);
```

but this does not:

```
  std::vector<float> qto_xyz;
  qto_xyz.push_back(nim->qto_xyz.m[0][0]);
  ...
  qto_xyz.push_back(nim->qto_xyz.m[3][3]);
  EncapsulateMetaData<std::vector>(thisDic, "qto_xyz", qto_xyz);
```

How should we store a 4x4 matrix of floats?  It does not seem to like std::vector.

dzenanz: Judging by [this](https://github.com/InsightSoftwareConsortium/ITK/blob/v5.3rc03/Modules/Core/Common/include/itkMetaDataObject.h#L53-L54), `std::vector` should be supported. But you probably need to give it's element type too `std::vector<float>`, like this: `EncapsulateMetaData<std::vector<float>>(thisDic, "qto_xyz", qto_xyz);`

seanm: Oh of course!  That works.  Can you tell C++ is not my mother tongue? :)

dzenanz: What is the type of `qto_xyz`? Maybe you could encapsulate it directly.

seanm: It's a niftilib struct, so not really appropriate to expose as such.

seanm:
I've updated the commit message, and this is ready to merge IMHO.

dzenanz (approved):

seanm:
But don't merge yet... I want to be 100% my own app works against this...

seanm:
OK, good to go, I had a minor panic there for a second, but there was a typo in my own code :)

hjmjohnson (approved):

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -994,6 +994,29 @@ NiftiImageIO::SetImageIOMetadataFromNIfTI()
...
+  qto_xyz.push_back(nim->qto_xyz.m[1][2]);
+  qto_xyz.push_back(nim->qto_xyz.m[1][3]);
+  qto_xyz.push_back(nim->qto_xyz.m[2][0]);
+  qto_xyz.push_back(nim->qto_xyz.m[2][1]);
+  qto_xyz.push_back(nim->qto_xyz.m[2][2]);
+  qto_xyz.push_back(nim->qto_xyz.m[2][3]);
+  qto_xyz.push_back(nim->qto_xyz.m[3][0]);
+  qto_xyz.push_back(nim->qto_xyz.m[3][1]);
+  qto_xyz.push_back(nim->qto_xyz.m[3][2]);
+  qto_xyz.push_back(nim->qto_xyz.m[3][3]);
```
N-Dekker, line 1018: It appears that these 16 lines of code may be refactored to the following:

```
  for (const auto & row : nim->qto_xyz.m)
  {
    for (const float value : row)
    {
      qto_xyz.push_back(value);
    }
  }

```
Right?

dzenanz: Or nested `for r`, `for c` loops.

dzenanz: For loops are more readable and less prone to errors. Let the compiler do the unrolling, if advantageous.

Leengit: If speed is somehow the issue then we might make use of the `std::vector<>(begin, end)` constructor, as in
```cpp
std::vector<float> qto_xyz(&nim->qto_xyz.m[0][0], &nim->qto_xyz.m[0][0] + 16);
```
Admittedly this is ugly and does assume something about the packing of the `nim->qto_xyz.m` object.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -994,6 +994,29 @@ NiftiImageIO::SetImageIOMetadataFromNIfTI()
...
+  EncapsulateMetaData<std::vector<float>>(thisDic, "qto_xyz", qto_xyz);
```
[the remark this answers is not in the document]
seanm: Encoding `mat44` itself I don't like, because it's a niftilib data structure, which I don't want to leak outside ITK.  Adding an `operator==` to it would also require changing niftilib, and it's C anyway.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
seanm: I think the nested r & c loops is best...

N-Dekker: I think it is OK to assume that `nim->qto_xyz.m` is a simple `float [4][4]` because that's just the way it is  😄 

Still I guess it would be preferable to store it as a 4x4 matrix, rather than an std::vector of 16 elements. Right?

Leengit: We could `static_assert` that `nim->qto_xyz.m` is of type `float [4][4]`.  But if speed is not critical here, the nested r & c loops do fine.

seanm: Speed is not *at all* critical here. It's dwarfed by the file i/o to actually read the nifti file.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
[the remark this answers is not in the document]
N-Dekker: `mat44` is defined in the global namespace (obviously, because it is C), so technically it is allowed to add an `operator==` to "our" global namespace, for example as follows:

```
bool
operator==(const mat44 & lhs, const mat44 & rhs)
{
  return std::equal(
    std::begin(lhs.m), std::end(lhs.m), std::begin(rhs.m), [](const auto & leftRow, const auto & rightRow) {
      return std::equal(std::begin(leftRow), std::end(leftRow), std::begin(rightRow));
    });
}
```

Another option might be to adjust `itk::MetaDataObject<MetaDataObjectType>` to allow types without an `operator==`.

I realize that it is my fault that `itk::MetaDataObject<MetaDataObjectType>`  requires `operator==` for the type to be stored: pull request #2246 commit 694bbc0a1f58bdb4564252e6f75d8611f4e2a5ed  😮  It could be adjusted to use std::memcmp, for types that don't have an `operator==`.  (@Leengit A SFINAE opportunity!!!)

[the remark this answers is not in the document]
N-Dekker: > Encoding mat44 itself I don't like, because it's a niftilib data structure

OK, but then, what would you think about encapsulating the matrix as a `itk::Matrix<float, 4, 4>`?

dzenanz (approved):
This is good as-is, but if you want to improve it further feel free to modify.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
Leengit: Though not needed for the present case ... 
It might be nice, it might be nice
To have
```cpp
template <typename TUnderlying, int VRows, int VColumns>
constexpr inline std::vector<TUnderlying>
CArrayToStdVector(const TUnderlying (&arr)[VRows][VColumns])
{
  return { &arr[0][0], &arr[0][0] + VRows * VColumns };
}
```
on your side.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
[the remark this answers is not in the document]
seanm: That doesn't make me recoil with C++ whiplash. :)  But it starts to be very different from the rest of the existing code in that function, which seems to prefer basic data structures.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
N-Dekker: Just made a pull request that would ease passing a C-style `float [4][4]` array directly to a `itk::Matrix`, please check: #3305  It would support the following:

    using MatrixType = Matrix<float, 4, 4>;
    EncapsulateMetaData<MatrixType>(thisDic, "qto_xyz", MatrixType(nim->qto_xyz.m));

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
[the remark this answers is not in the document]
seanm: The other thing I like about this PR as-is is that's its easy to cherry pick just this change.  For my own app, I'll be cherry-picking this into ITK 5.2 for a hot fix.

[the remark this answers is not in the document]
N-Dekker: OK, but do you agree that `itk::Matrix<float, 4, 4>` would be a proper data type to use when storing `nim->qto_xyz` into the `itk::MetaDataDictionary`?

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
N-Dekker: > ```c++
> template <typename TUnderlying, int VRows, int VColumns>
> constexpr inline std::vector<TUnderlying>
> CArrayToStdVector(const TUnderlying (&arr)[VRows][VColumns])
> {
>   return { &arr[0][0], &arr[0][0] + VRows * VColumns };
> }
> ```

@Leengit It might be nice indeed, but it's still not obvious to me why a fixed-size two-dimensional array (matrix) would need to be flattened to a 1-dimensional vector, before storing it into the dictionary. It seems more intuitive to me to have an `CArrayToStdArray(const TUnderlying (&arr)[VRows][VColumns])` function, returning an `std::array` of `std::array` elements.  (`std::array<std::array<TUnderlying, VColumns>, VRows>`). Right?

Leengit: > returning an `std::array` of `std::array` elements

That would be nice.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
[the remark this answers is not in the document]
blowekamp: The type used here should be one of the explicitly instantiated MetaDataObject types: 
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkMetaDataObject.h#L298-L321

I believe only itk::Matrix<float,3,3> is there, itk::Matrix<float,4,4> would need to be added.

N-Dekker (approved):
Thanks @seanm I gave up my resistance 😺  While I still think it would be nicer to store `nim->qto_xyz` as a 4x4 float matrix (either `mat44`, or `itk::Matrix`, or an `std::array` of std::array's), Bradley (@blowekamp) has just explained that neither of them are currently supported as MetaDataObject type. You might still consider a vector-of-vectors, `std::vector<std::vector<float>>`, but that's up to you.

PR #3247, opened 2022-03-03
Add `MakeFilled<TContainer>(value)`, and let FixedArray, Index, and Size use it

N-Dekker:


Modules/Core/Common/include/itkMakeFilled.h
```diff
@@ -0,0 +1,47 @@
...
+/** Yields a container of the specified type, filled by the specified value, assigned to each of its elements.
+ * Only intended for container types of a compile-time fixed size, including instances of FixedArray (and derived
+ * classes), Index, Offset, Size, and std::array. Moreover, their size (number of elements) should be greater than zero.
+ */
+template <typename TContainer>
+constexpr TContainer
+MakeFilled(typename TContainer::const_reference value)
+{
+  // Note: The seemingly redundant {} initialization is required for C++14, in order to declare MakeFilled constexpr.
+  TContainer container{};
```
hjmjohnson, line 32: Perhaps we could provide hints to future C++17 or C++20 code maintainters

```c++
#if __cplusplus >= 201703L
TContainer container;
#else
TContainer container{};
#endif
```

At some point, ITK will require C++17 or C++20, and it will help future developers to know these details.  This is certainly not necessary, but if you already know what the future code should look like, now might be the time to document that.

Modules/Core/Common/include/itkFixedArray.h
```diff
@@ -423,12 +424,7 @@ class ITK_TEMPLATE_EXPORT FixedArray
   static constexpr FixedArray
   Filled(const ValueType & value)
   {
-    FixedArray result{};
-    for (ValueType & element : result.m_InternalArray)
-    {
-      element = value;
-    }
-    return result;
+    return MakeFilled<FixedArray>(value);
```
hjmjohnson, line 427: NICE!

Leengit (approved):
Thank you.

hjmjohnson (approved):

Modules/Core/Common/include/itkMakeFilled.h
N-Dekker: Thanks for the suggestion @hjmjohnson However, I hesitate, because in general, I would rather avoid `#if` directives. And specifically in this case, I'm pretty sure that the redundant `{}` initialization will be optimized away with any Release build.

Would it not be sufficient that the comment explicitly mentions "C++14"? So that it can be easily found by a global search for "C++14", when upgrading to C++17?

For the sake of completeness, I just tried:

```
constexpr int Answer() {
    int uninitializedVariable;
    uninitializedVariable = 42;
    return uninitializedVariable;
}

```

At https://godbolt.org/z/an8hhbEeK  Interestingly, it says (clang 13.0.1):

> warning: uninitialized variable in a constexpr function is a C++20 extension [-Wc++20-extensions]

So it may still take a while before we can say goodbye to those `{}` !

Leengit:
Looks good.

dzenanz (approved):

PR #3250, opened 2022-03-04
STYLE: Remove `std::` prefix from types that work without it

Leengit:
This is accomplished with
```sh
cd git/ITK
find * -type f |
  egrep '\.(c|cc|cxx|cpp|h|hh|hxx|hpp)$' |
  fgrep -v ThirdParty |
  fgrep -v itkIntTypes.h |
  xargs sed -i -r -e \
    's/std::((size|ptrdiff|max_align|u?int(_fast|_least|max|ptr)?[0-9]*)_t)/\1/g'
```
Closes #3248.

## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [X] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [X] Added test (or behavior not changed)
- [X] Updated API documentation (or API not changed)
- [X] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [X] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [X] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples) for all new major features (if any)

PR #3252, opened 2022-03-06
STYLE: Replace type aliases with `using`-declarations, in itkIntTypes.h

N-Dekker:
Reduced code redundancy by replacing type alias declarations of the form
`using` _type_ `= std::`_type_ with the equivalent `using`-declarations.

### DIFF
--- a/Modules/Core/Common/include/itkIntTypes.h
+++ b/Modules/Core/Common/include/itkIntTypes.h
@@ -25,38 +25,38 @@
 
 namespace itk
 {
-using int8_t = std::int8_t;
-using uint8_t = std::uint8_t;
-using int16_t = std::int16_t;
-using uint16_t = std::uint16_t;
-using int32_t = std::int32_t;
-using uint32_t = std::uint32_t;
-using int64_t = std::int64_t;
-using uint64_t = std::uint64_t;
+using std::int8_t;
+using std::uint8_t;
+using std::int16_t;
+using std::uint16_t;
+using std::int32_t;
+using std::uint32_t;
+using std::int64_t;
+using std::uint64_t;
 
-using int_least8_t = std::int_least8_t;
-using uint_least8_t = std::uint_least8_t;
-using int_least16_t = std::int_least16_t;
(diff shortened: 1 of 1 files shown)


Leengit (approved):
Thank you

dzenanz (approved):

hjmjohnson (approved):

Leengit:
There are other cases, which could be found with
```sh
cd ~/git/ITK
egrep -R 'using +([A-Za-z0-9_]+) *= *([A-Za-z0-9_]+::)+\1;'
```
or found and fixed with
```sh
cd ~/git/ITK
find * -type f | egrep '\.(c|cc|cxx|cpp|h|hh|hxx|hpp)$' | fgrep -v /ThirdParty/ |
  xargs sed -i -r -e 's/using +([A-Za-z0-9_]+) *= *(([A-Za-z0-9_]+::)+\1;)/using \2/g'
```

Leengit:
I retract the previous comment.

Leengit:
As I discovered when I tried the above `sed` command, this shortened form is not good enough for the compiler in _class_ scope for class members.  That makes me worry about whether the long and short forms are exactly equivalent in namespace scope.  If they are not exactly equivalent, perhaps that could be documented here (and/or in the commit message).

N-Dekker:
> As I discovered when I tried the above `sed` command, this shortened form is not good enough for the compiler in _class_ scope for class members. That makes me worry about whether the long and short forms are exactly equivalent in namespace scope. If they are not exactly equivalent, perhaps that could be documented here (and/or in the commit message).

Thanks for your little investigation on this topic @Leengit Indeed, these specific kind of `using`-declarations (`using std::`_type_) cannot be in class scope, but they work just fine in namespace score,  😃 

Of course I specifically focused on "itkIntTypes.h" as it relates to your PR #3250 commit c173dfd803acf943d9ab80683ac4133fea29cdf5  I still think we should have such a `using`-declaration for each `std` type for which we dropped the `std::` prefix. At least, if we don't want to just _assume_ that the `std::` prefix may be dropped on any supported platform.

Issue #2925, opened 2021-12-04
Test default default cases for strongly typed enum serialization


jhlegarreta:
### Description

Test default default cases (corresponding to non-existing enum values) for strongly typed enum serialization.

### Expected behavior

Default cases should also be tested.

### Actual behavior

Default cases are not being tested because the set of enum values for testing are systematically built using enum values corresponding to existing cases, e.g.
```
  const std::set<itk::RecursiveGaussianImageFilterEnums::GaussianOrder> allGaussianOrder{
    itk::RecursiveGaussianImageFilterEnums::GaussianOrder::ZeroOrder,
    itk::RecursiveGaussianImageFilterEnums::GaussianOrder::FirstOrder,
    itk::RecursiveGaussianImageFilterEnums::GaussianOrder::SecondOrder
  };
  for (const auto & ee : allGaussianOrder)
  {
    std::cout << "STREAMED ENUM VALUE RecursiveGaussianImageFilterEnums::GaussianOrder: " << ee << std::endl;
  }
``` 

Corresponding to:
https://github.com/InsightSoftwareConsortium/ITK/blob/828453d1bf61c487310d2d8c9570093e08798a40/Modules/Filtering/Smoothing/test/itkRecursiveGaussianImageFiltersTest.cxx#L515

### Reproducibility

%100.

### Versions

ITK `master`.

### Additional Information

When sorting directories by the uncovered lines in ascending order:
https://open.cdash.org/viewCoverage.php?buildid=7606267&filtercount=1&field1=filename/string&compare1=63&value1=&status=-1

It becomes apparent that many directories/files are 1 line away from getting a 100% code coverage, and such lines correspond to default cases (corresponding to non-existing enum values) for strongly typed enum serialization.

See, for example, the `Modules/Filtering/Smoothing/src` directory, which is only missing the following lines:
https://open.cdash.org/viewCoverageFile.php?buildid=7606267&fileid=39050963

Although ideally we should also test for the actual printed values (cross-referencing #983), that can be done in a separate PR.

This would be an easy change and is an easy way to increase the coverage across the code base.

PR #3253, opened 2022-03-06
ENH: Test streaming `itk::SpatialOrientation` enum class types

jhlegarreta:
Test streaming `itk::SpatialOrientation` enum class types.

Spatial orientation enums were refactored and turned into strongly typed
enums in commit 182c018.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/test/itkSpatialOrientationTest.cxx
+++ b/Modules/Core/Common/test/itkSpatialOrientationTest.cxx
@@ -1,0 +1,114 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+#include "itkSpatialOrientation.h"
+#include <iostream>
+#include <set>
+
+
+int
+itkSpatialOrientationTest(int, char *[])
+{
+
+  // Test streaming enumeration for SpatialOrientationEnums::CoordinateTerms elements
+  const std::set<itk::SpatialOrientationEnums::CoordinateTerms> allCoordinateTerms{
+    itk::SpatialOrientationEnums::CoordinateTerms::ITK_COORDINATE_UNKNOWN,
+    itk::SpatialOrientationEnums::CoordinateTerms::ITK_COORDINATE_Right,
+    itk::SpatialOrientationEnums::CoordinateTerms::ITK_COORDINATE_Left,
+    itk::SpatialOrientationEnums::CoordinateTerms::ITK_COORDINATE_Posterior,
+    itk::SpatialOrientationEnums::CoordinateTerms::ITK_COORDINATE_Anterior,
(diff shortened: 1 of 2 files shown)


Modules/Core/Common/test/itkSpatialOrientationTest.cxx
```diff
@@ -0,0 +1,113 @@
...
+#include "itkSpatialOrientation.h"
+#include <set>
+
+
+int
+itkSpatialOrientationTest(int, char *[])
+{
+
+  // Test streaming enumeration for SpatialOrientationEnums::CoordinateTerms elements
+  const std::set<itk::SpatialOrientationEnums::CoordinateTerms> allCoordinateTerms{
```
N-Dekker, line 28: Is there a reason why those enum values stored in an `std::set`, before printing them? It appears to me that these values might as well be passed directly to the `for` loop, like the following:

```
for (const auto enumValue : { itk::SpatialOrientationEnums::CoordinateTerms::ITK_COORDINATE_UNKNOWN,
                              itk::SpatialOrientationEnums::CoordinateTerms::ITK_COORDINATE_Right,
                              // ...
                              itk::SpatialOrientationEnums::CoordinateTerms::ITK_COORDINATE_Superior })

```

Just my two cents, Niels

jhlegarreta: I know this has been requested and changed in https://github.com/InsightSoftwareConsortium/ITK/pull/3155#r798926614, but ALL the rest of the serializations have been done using an `std::set`. I am not against the way you propose, but that should be done in a separate PR, across the whole code, and the ITK SW Guide be changed accordingly: https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/pull/155.

Modules/Core/Common/test/itkSpatialOrientationTest.cxx
```diff
@@ -0,0 +1,114 @@
...
+    itk::SpatialOrientationEnums::CoordinateTerms::ITK_COORDINATE_Right,
+    itk::SpatialOrientationEnums::CoordinateTerms::ITK_COORDINATE_Left,
+    itk::SpatialOrientationEnums::CoordinateTerms::ITK_COORDINATE_Posterior,
+    itk::SpatialOrientationEnums::CoordinateTerms::ITK_COORDINATE_Anterior,
+    itk::SpatialOrientationEnums::CoordinateTerms::ITK_COORDINATE_Inferior,
+    itk::SpatialOrientationEnums::CoordinateTerms::ITK_COORDINATE_Superior
+  };
+  for (const auto & ee : allCoordinateTerms)
+  {
+    std::cout << "STREAMED ENUM VALUE SpatialOrientationEnums::CoordinateTerms: " << ee << std::endl;
```
N-Dekker, line 40: Is it really helpful to print all these enums to standard output? Or could they also be inserted to a string-stream, instead? For example as follows:

    std::ostringstream stringStream;
    stringStream << ee;
    ASSERT( ! stringStream.str().empty() );

Doing so might reduce the "noise" in the test output.

Modules/Core/Common/test/itkSpatialOrientationTest.cxx
N-Dekker: > I know this has been requested and changed in [#3155 (comment)](https://github.com/InsightSoftwareConsortium/ITK/pull/3155#r798926614),

So fortunately, I'm consistent with my comments  😺 Always trying to avoid redundant and unnecessary code.

Modules/Core/Common/test/itkSpatialOrientationTest.cxx
jhlegarreta: @N-Dekker
- The serialization test not being quantitative has been mentioned before (see the `Additional Information` in #2925). 
- As said, any change to these affects any other strongly typed enums in the code base, and there are a few dozens. The above issue can host any discussion and proposals to improve these. Changes should be done for all strongly typed enum serialization tests at once. So it is beyond the scope of this PR.

dzenanz:
Merging. Consistency improvements can be accomplished in another PR.

PR #3237, opened 2022-02-27
ENH: Declare converting `Point(v)` constructors `explicit`

N-Dekker:
Prevented implicit conversion of a value to `Point`. Aims to avoid
possible flaws, like the ones addressed by:

Pull request https://github.com/InsightSoftwareConsortium/ITK/pull/3225
commit 5b71d63a88cc80547e8584eae59e945fbbf51e5d
"BUG: Quick-fix ContourSpatialObject::Update(), LINEAR_INTERPOLATION case"

Pull request https://github.com/InsightSoftwareConsortium/ITK/pull/3228
commit 12a501cddc3f4c4c93190cb2e64bb2b9c1f558c0
"BUG: Fix `SetCenterInObjectSpace` calls in Registration test"

Pull request https://github.com/InsightSoftwareConsortium/ITK/pull/3231
commit e4841aaba2fd67dabda95189bc1a93da4fa146c3
"STYLE: Clarify estimation ray position `RayCastInterpolateImageFunction`"

Deleted `Point(nullptr_t)`, especially to avoid `PointType p = 0`.

PR #3255, opened 2022-03-07
Remove implicit conversion from a single scalar value to `itk::Vector`

N-Dekker:
Follow-up to commit a8909625948547989f97385fd0b06c9f331606eb "ENH: Make Vector construction from scalar value explicit" by Brian Helba (@brianhelba), January 26, 2015

Analog to: pull request #3237 commit 8825834406356a66705437b74c843a54a47c57c3 "ENH: Declare converting `Point(v)` constructors `explicit`"

### DIFF
--- a/Modules/Core/Common/include/itkVector.h
+++ b/Modules/Core/Common/include/itkVector.h
@@ -106,7 +106,7 @@ public:
    * \note The other five "special member functions" are defaulted implicitly, following the C++ "Rule of Zero". */
   Vector() = default;
 
-#if !defined(ITK_LEGACY_FUTURE_REMOVE)
+#if !defined(ITK_LEGACY_REMOVE)
   /** Constructor to initialize entire vector to one value.
    * \warning Not intended to convert a scalar value into
    * a Vector filled with that value.
@@ -116,6 +116,9 @@ public:
   /** Constructor to initialize entire vector to one value,
    * if explicitly invoked. */
   explicit Vector(const ValueType & r);
+
+  /** Prevents copy-initialization from `nullptr`, as well as from `0` (NULL). */
+  Vector(std::nullptr_t) = delete;
 #endif
 
   /** Pass-through constructor for the Array base class. */
--- a/Documentation/ITK5MigrationGuide.md
+++ b/Documentation/ITK5MigrationGuide.md
@@ -454,6 +454,12 @@ The namespece itself is only available in legacy mode. Some classes were renamed
 
 Enumeration member names (`ITK_COORDINATE_UNKNOWN`, `ITK_COORDINATE_Right`, `ITK_COORDINATE_PrimaryMinor`, `ITK_COORDINATE_ORIENTATION_RIP` etc) are unchanged.
 
+Implicit conversion of a single scalar value to a container (which would _fill_ the container by the
+scalar value) is discouraged. With ITK 5.3, when having `ITK_LEGACY_REMOVE=ON`, the constructors of
+`Point`, `RGBPixel`, `RGBAPixel`, and `Vector` that accept a single scalar value as argument are
+declared `explicit`. ITK 5.3 has included a preferable alternative to these constructors:
+`itk::MakeFilled<ContainerType>(value)`.
+
 Consolidated Vector Filter
 --------------------------
 
--- a/Modules/Core/SpatialObjects/include/itkSpatialObject.h
+++ b/Modules/Core/SpatialObjects/include/itkSpatialObject.h
@@ -318,7 +318,7 @@ public:
                             CovariantVectorType &        value,
                             unsigned int                 depth = 0,
                             const std::string &          name = "",
-                            const DerivativeOffsetType & offset = 1);
+                            const DerivativeOffsetType & offset = MakeFilled<DerivativeOffsetType>(1));
 
   /** Return the n-th order derivative value at the specified point. */
   virtual void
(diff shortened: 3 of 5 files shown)


Modules/Core/Common/include/itkVector.h
```diff
@@ -106,7 +106,7 @@ class ITK_TEMPLATE_EXPORT Vector : public FixedArray<T, VVectorDimension>
    * \note The other five "special member functions" are defaulted implicitly, following the C++ "Rule of Zero". */
   Vector() = default;
 
-#if !defined(ITK_LEGACY_FUTURE_REMOVE)
+#if !defined(ITK_LEGACY_REMOVE)
```
dzenanz, line 109: Do we need an entry in migration guide for this change? Or is this minor enough?

dzenanz (approved):

N-Dekker:
@dzenanz Do you have any suggestion how to fix https://open.cdash.org/viewBuildError.php?buildid=7780689 ?

> [Wrapping/Modules/ITKLevelSets/itkVectorThresholdSegmentationLevelSetImageFilterPython.cpp:3906](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Wrapping/Modules/ITKLevelSets/itkVectorThresholdSegmentationLevelSetImageFilterPython.cpp#L3906):44: error: no matching function for call to 'itk::VectorThresholdSegmentationLevelSetImageFilter<itk::Image<float, 2>, itk::Image<itk::Vector<float, 2>, 2>, float>::SetMean(const double&)'

       (arg1)->SetMean((double const &)*arg2);

I can't find that cpp file (itkVectorThresholdSegmentationLevelSetImageFilterPython.cpp), how is it generated?

dzenanz:
That file is automatically generated by SWIG, from `itkVectorThresholdSegmentationLevelSetImageFilter.wrap`.

dzenanz:
I guess that making the constructor explicit is preventing some `SetMean` call from compiling. The reported lines do not match my local file. I guess local python compilation is needed for hunting this down.

N-Dekker:
The error message does seem to make sense, because `SetMean` expects a Vector as argument, not just a single `double` value:

https://github.com/InsightSoftwareConsortium/ITK/blob/4f309803b9b3c0268679d658e22d2f751f9c7b6c/Modules/Segmentation/LevelSets/include/itkVectorThresholdSegmentationLevelSetImageFilter.h#L122

Of course, we could add an extra  `SetMean(double)` overload, which would _fill_ the Vector of mean values by the specified `double` value. But would that be desirable?

Leengit:
The build with wrapping ends up compiling a few versions of the templates that aren't otherwise built.  Hopefully the above bugs can be found with `cmake` flags like
```sh
-DITK_WRAP_complex_double:BOOL=ON DITK_WRAP_complex_float:BOOL=ON \
-DITK_WRAP_covariant_vector_double:BOOL=ON \
-DITK_WRAP_covariant_vector_float:BOOL=ON DITK_WRAP_double:BOOL=ON \
-DITK_WRAP_float:BOOL=ON DITK_WRAP_rgba_unsigned_char:BOOL=ON \
-DITK_WRAP_rgba_unsigned_short:BOOL=ON \
-DITK_WRAP_rgb_unsigned_char:BOOL=ON \
-DITK_WRAP_rgb_unsigned_short:BOOL=ON DITK_WRAP_signed_char:BOOL=ON \
-DITK_WRAP_signed_long_long:BOOL=ON DITK_WRAP_signed_short:BOOL=ON \
-DITK_WRAP_unsigned_char:BOOL=ON DITK_WRAP_unsigned_long_long:BOOL=ON \
-DITK_WRAP_unsigned_short:BOOL=ON DITK_WRAP_vector_double:BOOL=ON \
-DITK_WRAP_vector_float:BOOL=ON
```
... without having to wade through the auto-generated code used for wrapping if we also set `-DITK_WRAP_PYTHON:BOOL=OFF`.

N-Dekker:
> This still fails to compile. Matt suggested to revert in #3258.

@dzenanz  I just hope 8825834406356a66705437b74c843a54a47c57c3 doesn't get reverted (as suggested by PR #3258). But what can I do?

dzenanz:
Finding bugs exposed by the explicit constructor are a great attempt, which might work if there aren't too many of them 😄

N-Dekker:
I think this pull request is ready now 😃 The missing `#include "itkNumericTraitsVectorPixel.h"`, fixed with https://github.com/InsightSoftwareConsortium/ITK/pull/3255/commits/563256ac55dd3d6aa23b83242686fb0776514563 might be a serious bug, because `MeanVectorType` appears defined as `double`, instead of a `itk::Vector`, when the specific `NumericTraits<Vector>` specialization is not included. Problem solved!  😄

thewtex (approved):

dzenanz (approved):

hjmjohnson (approved):

hjmjohnson:
building AnisotropicDiffusionLBR now fails:

```c++
/Users/johnsonhj/Dashboard/src/AnisotropicDiffusionLBR/include/itkLinearAnisotropicDiffusionLBRImageFilter.hxx:480:27: error: reference to type 'const itk::Vector<double, 3>' could not bind to an rvalue of type 'double'
  m_NextImage->FillBuffer(0.);
                          ^~
```

NOTE:  ITK_FUTURE_LEGACY_REMOVE:BOOL=ON. Some of the remote modules probably will need some TLC .

N-Dekker:
Thanks for the notice, @hjmjohnson 

    m_NextImage->FillBuffer(0.);

What you would need here is a zero-initialized `itk::Vector`. In general, this can be achieved by the syntax `VectorType()`, or `VectorType{}`. (Assuming that there is a `VectorType` alias. Otherwise I guess something like `ImageType::PixelType`) But probably an empty pair of curly braces will do as well:

     m_NextImage->FillBuffer({});

HTH, Niels

hjmjohnson:
FYI:  https://github.com/InsightSoftwareConsortium/ITKAnisotropicDiffusionLBR/pull/51

PR #3200, opened 2022-02-14
WIP: BUG: Restructured python type hint generation procedure

kian-weimer:
This PR should resolve an issue discovered in PR #3145 

> 3\. When designing this system, I mistakenly assumed that all instances of a class would be defined within one submodule and therefore the wrapping process for a given class would happen in just one run of `igenerator.py`. However, in some cases a class has a new overload defined for it within a different submodule. For example, fixed array is wrapped in both `itkShapeLabelObject.wrap` and `itkFixedArray.wrap` which means that it appears in two separate, potentially parallel runs of `igenerator.py`. This causes The class to appear in two separate .pyi files, thus preventing some of the class overloads to appear properly in the type hints.

As stated above, a few inter-submodule dependencies were causing a few classes type hints to fail.
This commit restructures the generation process to allow for these dependencies.  
The generation occurs in two stages:
1. pickle files are saved for each class defined for each submodule in igenerator
	(Some classes are defined multiple times (once for each submodule))
2. Pickle files are merged and stub files are produced in pyi_generator
	(This runs once after all runs of igenerator have finished)

Currently pyi_generator runs before igenerator, this is a problem that is still being worked on.
A dependency needs to be added to cause pyi_generator to be ran last.
Comments have been added to the CMake files to help describe my current implementation

## PR Checklist
- [ ] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [ ] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [ ] Added test (or behavior not changed)
- [ ] Updated API documentation (or API not changed)
- [ ] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [ ] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [ ] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples) for all new major features (if any)

Refer to the [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) for
further development details if necessary.

Issue #2058, opened 2020-10-14
Doxygen documentation is broken


issakomi:
The problem was first mentioned here

https://discourse.itk.org/t/how-to-process-ct-images-one-by-one-and-save-these-images-of-one-series-in-memory/3489/11

but unfortunately is not limited to particular class, e.g. see screenshot

![Screenshot at 2020-10-14 14-53-31](https://user-images.githubusercontent.com/1494580/96002099-097c5500-0e39-11eb-9514-64815df1122c.png)

blowekamp:
A problem with the [itkgroup.pl](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/Doxygen/itkgroup.pl) might explain the problem.

issakomi:
BTW, locally build HTML Doxygen documentation doesn't have the issue (note, it doesn't contain Sphinx example links, "^" instead), here is [html.tar.gz](https://drive.google.com/file/d/1NiB0C_6SpUqKA8FfyEkfxr-hmN24ppsc/view?usp=sharing) , if someone wish to use the file, please download, it is 1.6GB and my free space on drive is limited, i shall delete the file in several days or so. 

Unfortunately i don't know how to invoke _itkgroup.pl_ correctly to test, it is not in _build_ folder, only in source. And my Doxygen knowledge in near zero. BTW, last change of that file was 7 years ago.

The related settings were
![Screenshot at 2020-10-14 19-56-48](https://user-images.githubusercontent.com/1494580/96027596-61768400-0e58-11eb-89c2-f5ceacd0a7d4.png)

blowekamp:
Then perhaps it is related the the version of Doxygen used for the build.

brad-t-moore:
Resurrecting this is it is a current issue: [https://itk.org/Doxygen/html/classitk_1_1SpatialObject.html](https://itk.org/Doxygen/html/classitk_1_1SpatialObject.html)

brad-t-moore:
@thewtex @dzenanz Have you all noticed this too?  I'd say many classes' API documentation is affected.

blowekamp:
PR #3183 should have addressed this issue. But the itk.org Doxygen hasn't been updated since Jan 18th. There is some additional info here: https://discourse.itk.org/t/doxygen-with-github-actions-and-pages/4790/1

Check the class in the liked alternate docs and let us know if it is OK there.

PR #3264, opened 2022-03-09
BUG: Provide module name with ITK_WRAP_DOC enabled

thewtex:
WRAPPER_MODULE_NAME is no longer available with the recent wrapping
macro refactoring.

Addresses:

```
CMake Error:
  Running

   '/usr/bin/ninja' '-C' '/work/ITK-cp39-cp39-manylinux2014_x64' '-t' 'recompact'

  failed with:

   ninja: error: build.ninja:56497: multiple rules generate Wrapping/Typedefs/_doc.i [-w dupbuild=err]
```

that occurs when building the Python packages when ITK_WRAP_DOC is
enabled.

### DIFF
--- a/Wrapping/Generators/CastXML/CMakeLists.txt
+++ b/Wrapping/Generators/CastXML/CMakeLists.txt
@@ -196,7 +196,7 @@ endmacro()
 macro(itk_end_wrap_submodule_castxml module)
   # write the wrap_*.cxx file
   #
-  # Global vars used: WRAPPER_INCLUDE_FILES WRAPPER_MODULE_NAME and WRAPPER_TYPEDEFS
+  # Global vars used: WRAPPER_INCLUDE_FILES and WRAPPER_TYPEDEFS
   # Global vars modified: none
 
   # Create the cxx file which will be given to castxml.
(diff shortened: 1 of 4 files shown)


dzenanz (approved):

dzenanz:
Is this related to #2058?

kian-weimer (approved):
It looks like the `WRAPPER_MODULE_NAME` variable was removed from the `itk_wrap_submodule` macro. This should fix it. We're working on more wrapping CMake refactors to try to clean things up. Will watch out for issues like this in the future.

hjmjohnson (approved):
@thewtex Sorry for the oversight.  The change that Kian and I are making will hopefully make the code easier to understand, and easier to manipulate in the future. 

The side effects of the macro nesting is hard to track.

hjmjohnson:
@thewtex See the next round of tested changes here: https://github.com/InsightSoftwareConsortium/ITK/pull/3200. We hope to have another PR to simplify the code tonight or tomorrow.  Then we can get back to understanding what needs to be done to get the type-hinting working.  

I think that the type-hinting code is mostly working, we just need to resolve the problem of multiple sources of overriding causing only the final implementation to be reported.

PR #3267, opened 2022-03-10
DOC: Do not end the subject line with a period

Leengit:
Remove the period at the end of the subject line in the bullet immediately following the bullet that indicates `Do not end the subject line with a period.`


N-Dekker (approved):
Fine by me, although I still find that "spacing not being considered" a bit hard to understand. What about "STYLE: Change template parameter name prefix N to V", as an example subject line? Wouldn't that be a clearer? At least, I like it better, as it is more straightforward! And it's yours  😃 

Approved anyway

Leengit:
> What about "STYLE: Change template parameter name prefix N to V", as an example subject line?

Done, with the most recent force push.

dzenanz (approved):

thewtex (approved):

PR #3302, opened 2022-03-14
ENH: Add `NumberToString<>` specialization, deducing its parameter type

N-Dekker:
Specialized `NumberToString<TValue>` for `TValue` = `void`, allowing users to convert a number of any type to string without having to specify the type of the number, passed as parameter.

Inspired by the `std::less<T>` function object, which has a similar `std::less<>` specialization: https://en.cppreference.com/w/cpp/utility/functional/less 

Aims to make `NumberToString` more appealing, especially for users who would rather not explicitly specify the type of every number they need to convert.  Otherwise those users might rather use `std::ostream` (`<<`) or `std::to_string`, which might be a pity, because `NumberToString` offers full precision (lossless) conversion.

### DIFF
--- a/Modules/Core/Common/test/itkNumberToStringGTest.cxx
+++ b/Modules/Core/Common/test/itkNumberToStringGTest.cxx
@@ -98,6 +98,28 @@ Test_decimal_notation_supports_up_to_twentyone_digits()
   }
 }
 
+
+template <typename TValue>
+void
+Test_default_specialization_of_NumberToString()
+{
+  using NumericLimitsType = std::numeric_limits<TValue>;
+
+  for (const TValue value : { NumericLimitsType::lowest(),
+                              NumericLimitsType::denorm_min(),
+                              TValue(),
+                              NumericLimitsType::epsilon(),
+                              NumericLimitsType::min(),
+                              NumericLimitsType::max(),
+                              NumericLimitsType::infinity(),
+                              NumericLimitsType::quiet_NaN() })
+  {
+    // Expect the same string from the default specialization `NumberToString<>` as from the TValue specific
+    // `NumberToString<TValue>`.
+    EXPECT_EQ(itk::NumberToString<>()(value), itk::NumberToString<TValue>()(value));
+  }
+}
+
 } // namespace
 
 
@@ -141,4 +163,15 @@ TEST(NumberToString, DecimalNotationUpTo21Digits)
 {
   Test_decimal_notation_supports_up_to_twentyone_digits<float>();
   Test_decimal_notation_supports_up_to_twentyone_digits<double>();
+}
+
+
+TEST(NumberToString, DefaultSpecialization)
+{
+  Test_default_specialization_of_NumberToString<int8_t>();
+  Test_default_specialization_of_NumberToString<uint8_t>();
+  Test_default_specialization_of_NumberToString<intmax_t>();
+  Test_default_specialization_of_NumberToString<uintmax_t>();
+  Test_default_specialization_of_NumberToString<float>();
+  Test_default_specialization_of_NumberToString<double>();
 }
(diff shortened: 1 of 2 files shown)


Leengit:
We might want to have only a declaration in `itkNumberToString.h` and have the definition in `itkNumberToString.hxx`.

N-Dekker:
> Instead of template specialization (for `void`, `float`, etc.) 

Thank you for the suggestion(s) @Leengit, but it is pretty much _essential_ that `NumberToString` has a specialization for `float`, in order to support lossless float-to-string conversion, as implemented by Google's double-conversion library! 

> we could use SFINAE

SFINAE is a wonderful mechanism but it's also a bit tricky, and it may be hard to understand, so I would be reluctant to use it in _production code_, unless it's obviously the best solution to a specific problem. Do you think the SFINAE approach would really be superior in this case, from a user perspective?

Leengit:
> Thank you for the suggestion(s) @Leengit, but it is pretty much _essential_ that `NumberToString` has a specialization for `float`, in order to support lossless float-to-string conversion, as implemented by Google's double-conversion library!

Although I see a `declaration` specialization for `float` I don't see a `definition` specialization for that.  If the latter exists somewhere then the suggested SFINAE approach is not taking advantage of it. :frowning_face:  If it doesn't exist then maybe the SFINAE test does pass this test. :smiley:

> Do you think the SFINAE approach would really be superior in this case, from a user perspective?

It's not a big deal either way.  I was motivated by not having distinct lines of code to implement the functions `itk::NumberToString<void>.operator()<float>` and `itk::NumberToString<float>.operator()` which are intended to produce the same result.  Having one of them call the other almost gives us completely shared lines of code, and is plenty good enough if your hesitancy for using SFINAE is stronger.

Leengit:
> Specifically for a function that just has one or two lines of code, and no extra external header dependencies, I think it is very well justifiable to just keep its function definition in the .h file, inside the class definition.

That's sensible.  I was motivated by the fact that the `itkNumberToString.hxx` file already exists and already has the non-specialized definition, so I was thinking to keep the definitions together.

N-Dekker:
> Although I see a `declaration` specialization for `float` I don't see a `definition` specialization for that. 

@Leengit FYI Here it is, in Modules/Core/Common/src/itkNumberToString.cxx:
https://github.com/InsightSoftwareConsortium/ITK/blob/8123eb3e91cc9a96d9fefba78e1e34a1a9842b4a/Modules/Core/Common/src/itkNumberToString.cxx#L74-L79

Modules/Core/Common/test/itkNumberToStringGTest.cxx
```diff
@@ -142,3 +164,14 @@ TEST(NumberToString, DecimalNotationUpTo21Digits)
...
+
+
+TEST(NumberToString, DefaultSpecialization)
+{
+  Test_default_specialization_of_NumberToString<int8_t>();
+  Test_default_specialization_of_NumberToString<uint8_t>();
+  Test_default_specialization_of_NumberToString<intmax_t>();
+  Test_default_specialization_of_NumberToString<uintmax_t>();
+  Test_default_specialization_of_NumberToString<float>();
+  Test_default_specialization_of_NumberToString<double>();
```
dzenanz, line 176: do we have an overload for `long double`?

dzenanz (approved):

Modules/Core/Common/test/itkNumberToStringGTest.cxx
dzenanz: I don't need it. I was just checking whether it is omitted on purpose.

dzenanz:
Maybe give Hans a few hours for a chance to review, and then merge?

hjmjohnson (approved):

PR #3305, opened 2022-03-16
ENH: Add explicit `Matrix(const RawDataArrayType &)` constructor

N-Dekker:
Allows constructing an `itk::Matrix` by passing a C-style array of
arrays, of the corresponding value type and dimensions.

### DIFF
--- a/Modules/Core/Common/test/itkMatrixGTest.cxx
+++ b/Modules/Core/Common/test/itkMatrixGTest.cxx
@@ -19,6 +19,7 @@
 // First include the header file to be tested:
 #include "itkMatrix.h"
 #include <gtest/gtest.h>
+#include <numeric>     // For iota.
 #include <type_traits> // For is_convertible and is_trivially_copyable.
 
 
@@ -71,6 +72,33 @@ vnl_matrix_fixed_is_convertible_to_itk_Matrix()
     TMatrix>();
 }
 
+
+template <typename TMatrix>
+void
+Expect_Matrix_is_constructible_from_raw_array_of_arrays()
+{
+  using ValueType = typename TMatrix::ValueType;
+
+  constexpr auto    numberOfRows = TMatrix::RowDimensions;
+  constexpr auto    numberOfColumns = TMatrix::ColumnDimensions;
+  constexpr auto    numberOfElements = numberOfRows * numberOfColumns;
+  ValueType         rawArray[numberOfRows][numberOfColumns];
+  ValueType * const beginOfRawArray = rawArray[0];
+
+  // Just ensure that each element of the raw array has a different value.
+  std::iota(beginOfRawArray, beginOfRawArray + numberOfElements, ValueType{ 1 });
+
+  // Construct an itk::Matrix from a raw C-style array-of-arrays.
+  const TMatrix matrix(rawArray);
+
+  for (unsigned int row = 0; row < numberOfRows; ++row)
+  {
+    for (unsigned int column = 0; column < numberOfColumns; ++column)
+    {
+      EXPECT_EQ(matrix(row, column), rawArray[row][column]);
+    }
+  }
+}
 } // namespace
 
 
@@ -107,4 +135,11 @@ TEST(Matrix, GetIdentity)
   Expect_GetIdentity_returns_identity_matrix<itk::Matrix<float>>();
   Expect_GetIdentity_returns_identity_matrix<itk::Matrix<double>>();
   Expect_GetIdentity_returns_identity_matrix<itk::Matrix<double, 2, 2>>();
+}
+
+
+TEST(Matrix, IsConstructibleFromRawArrayOfArrays)
+{
+  Expect_Matrix_is_constructible_from_raw_array_of_arrays<itk::Matrix<float>>();
+  Expect_Matrix_is_constructible_from_raw_array_of_arrays<itk::Matrix<double, 2, 3>>();
 }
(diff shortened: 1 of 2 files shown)


Modules/Core/Common/test/itkMatrixGTest.cxx
```diff
@@ -108,3 +135,10 @@ TEST(Matrix, GetIdentity)
   Expect_GetIdentity_returns_identity_matrix<itk::Matrix<double>>();
   Expect_GetIdentity_returns_identity_matrix<itk::Matrix<double, 2, 2>>();
 }
+
+
+TEST(Matrix, IsConstructibleFromRawArrayOfArrays)
+{
+  Expect_Matrix_is_constructible_from_raw_array_of_arrays<itk::Matrix<float>>();
```
dzenanz, line 142: What are the numbers of rows and columns in this case?

dzenanz (approved):

Modules/Core/Common/test/itkMatrixGTest.cxx
N-Dekker: > What are the numbers of rows and columns in this case?

https://github.com/InsightSoftwareConsortium/ITK/blob/0e1c2c3dcd96fe85cd5a6d190239af950acac851/Modules/Core/Common/include/itkMatrix.h#L50-L51

My intention was to test both a default and a non-default combination of dimensions. Rather than exhaustively test all possibilities  😺

N-Dekker:
Passed well on the regular C++ builds (ITK.Linux, ITK.Windows, ITK.macOS), but failed on the Python builds, saying:

> Wrapping/Typedefs/itkMatrix.i:76: Error: Syntax error in input(3). 

At https://open.cdash.org/viewBuildError.php?buildid=7798097

Any clue? Would `using RawDataArrayType = T[VRows][VColumns]` be a problem to SWIG? Would it prefer an old-style `typedef` for array types?

dzenanz:
> T[VRows][VColumns] a problem to SWIG

Probably. @thewtex will know, maybe even @PranjalSahu (he had a close encounter with Wrapping recently).

N-Dekker:
This line of code:

https://github.com/InsightSoftwareConsortium/ITK/blob/d9a729c91f9fa025ca2a3f38334ec770825b92c4/Modules/Core/Common/include/itkMatrix.h#L69

    using RawDataArrayType = T[VRows][VColumns]; 

Could be rewritten to "old-style":

    typedef T RawDataArrayType[VRows][VColumns]; 

Just wondering if that would fix the SWIG errors...

dzenanz:
I don't think old-style `typedef` would make any difference compared to `using`. I think that the problem comes from the actual type attempting to be wrapped.

N-Dekker:
My most recent [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/d9a729c91f9fa025ca2a3f38334ec770825b92c4..d2db189558b5469fd62495df00456863dedf9a5d) removed the typedef (`using` alias) for `T[VRows][VColumns]`, but the SWIG errors are still there 🤔 

At https://open.cdash.org/viewBuildError.php?buildid=7799641

> Wrapping/Typedefs/itkMatrix.i:76: Error: Syntax error in input(3) 

So I guess SWIG just does not like a reference to a C-style array as parameter (like the one of the proposed new explicit `itk::Matrix` constructor)... but I'm not sure. 🤷 

I still believe the pull request would be useful, but apparently it doesn't "swig".

N-Dekker:
@dzenanz Interesting... it looks like my last [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/d2db189558b5469fd62495df00456863dedf9a5d..3624d55069d7d67d177ab271665bd5937b076aa7) has worked around the SWIG errors!!!  😄 SWIG appears to like:

    template <typename TRow>
    explicit Matrix(const TRow (&rows)[VRows])

While SWIG did not like:

    explicit Matrix(const T (&arrayOfArrays)[VRows][VColumns])

Personally I would prefer the `const T (&arrayOfArrays)[VRows][VColumns]` version, but eventually I guess it's up to SWIG what to do 🤔 !

dzenanz (approved):

Leengit:
SWIG didn't like the non-templated code, but in this case there are multiple ways to template.  I don't know that they are any better, but for SWIG research purposes, did you happen to try either of
```cpp
  template <typename TMatrix>
  explicit Matrix(const TMatrix & matrix)
    : m_Matrix(&matrix[0][0])
  {
    static_assert(std::is_same<TMatrix, T[VRows][VColumns]>::value,
                  "The type of the matrix should correspond with this itk::Matrix instantiation.");
  }

  template <typename TElement>
  explicit Matrix(const TElement (&elements)[VRows][VColumns])
    : m_Matrix(&elements[0][0])
  {
    static_assert(std::is_same<TElement, T>::value,
                  "The type of an element should correspond with this itk::Matrix instantiation.");
  }
```

N-Dekker:
> SWIG didn't like the non-templated code, but in this case there are multiple ways to template. I don't know that they are any better, but for SWIG research purposes, did you happen to try either

No I did not. Thanks for the suggestion @Leengit. `Matrix(const TMatrix & )` does not really appeal to me, because it might be too greedy, in case of overload resolution. Although I guess you might SFINAE away non-array-of-array argument types. `Matrix(const TElement (&elements)[VRows][VColumns])` would certainly be better, but I'm afraid SWIG wouldn't like it. Although I'm not sure. So you might check, if you really want to know.

Leengit:
Thanks @N-Dekker.  Your code is a-okay, so there is no need to put additional work into these variations.  I was asking in case you already had the information.

Leengit (approved):

PR #3307, opened 2022-03-17
ENH: Let `MetaDataDictionary` support non-equality-comparable data

N-Dekker:
No longer require `operator==` for the types of values stored in `itk::MetaDataDictionary`.

Follow-up to:
pull request #2246
commit 694bbc0a1f58bdb4564252e6f75d8611f4e2a5ed
"ENH: Add operator==, operator!= to MetaDataObject and MetaDataDictionary"

### DIFF
--- a/Modules/Core/Common/include/itkMetaDataObject.h
+++ b/Modules/Core/Common/include/itkMetaDataObject.h
@@ -134,7 +134,7 @@ public:
   friend bool
   operator==(const Self & lhs, const Self & rhs)
   {
-    return lhs.m_MetaDataObjectValue == rhs.m_MetaDataObjectValue;
+    return EqualMetaDataObjects(&lhs.m_MetaDataObjectValue, &rhs.m_MetaDataObjectValue);
   }
 
   /** Returns (metaDataObject1 != metaDataObject2). */
@@ -149,6 +149,24 @@ protected:
   ~MetaDataObject() override = default;
 
 private:
+  /** Tells whether the specified pointers point to objects that compare equal.
+   * \note The template parameter and the trailing return type are there, just to enable SFINAE.*/
+  template <typename TMetaDataObject>
+  static auto
+  EqualMetaDataObjects(const TMetaDataObject * const lhs, const TMetaDataObject * const rhs) -> decltype(*lhs == *rhs)
+  {
+    static_assert(std::is_same<TMetaDataObject, MetaDataObjectType>::value, "Check the template argument.");
+    return *lhs == *rhs;
+  }
+
+  /** Overload for MetaDataObject types that do not have an `operator==`.*/
+  static bool
+  EqualMetaDataObjects(const void * const lhs, const void * const rhs)
+  {
+    return std::memcmp(lhs, rhs, sizeof(MetaDataObjectType)) == 0;
+  }
+
+
   /** Internal helper function used to implement operator== for MetaDataObjectBase. */
   bool
   Equal(const MetaDataObjectBase & metaDataObjectBase) const override
--- a/Modules/Core/Common/test/itkMetaDataDictionaryGTest.cxx
+++ b/Modules/Core/Common/test/itkMetaDataDictionaryGTest.cxx
@@ -286,7 +286,8 @@ TEST(MetaDataDictionary, Equal)
 
   const auto createMetaDataDictionary = [](const int value) {
     itk::MetaDataDictionary metaDataDictionary;
-    itk::EncapsulateMetaData(metaDataDictionary, "key", value);
+    itk::EncapsulateMetaData(metaDataDictionary, "key1", value);
+    itk::EncapsulateMetaData(metaDataDictionary, "key2", std::to_string(value));
     return metaDataDictionary;
   };
 
@@ -301,4 +302,34 @@ TEST(MetaDataDictionary, Equal)
   expectUnequal(metaDataDictionary1, metaDataDictionary2);
   expectUnequal(metaDataDictionary1, defaultMetaDataDictionary);
   expectUnequal(metaDataDictionary2, defaultMetaDataDictionary);
+}
+
+
+TEST(MetaDataDictionary, SupportsNonEqualityComparableData)
+{
+  // An example of a data type that is not "equality-comparable".
+  struct DataStruct
+  {
+    int data;
+  };
+
+  itk::MetaDataDictionary dictionary;
+
+  for (const int i : { 0, 1 })
+  {
+    const auto       key = "key_" + std::to_string(i);
+    const DataStruct inputDataObjectValue = { i };
+    itk::EncapsulateMetaData(dictionary, key, inputDataObjectValue);
+
+    // Now retrieve the encapsulated data, and check if it has the expected value:
+
+    const itk::MetaDataObjectBase * const base = dictionary.Get(key);
+    ASSERT_NE(base, nullptr);
+
+    const auto * const dataObject = dynamic_cast<const itk::MetaDataObject<DataStruct> *>(base);
+    ASSERT_NE(dataObject, nullptr);
+
+    const DataStruct & dataObjectValue = dataObject->GetMetaDataObjectValue();
+    EXPECT_EQ(dataObjectValue.data, inputDataObjectValue.data);
+  }
 }


N-Dekker:
@Leengit Please have a look at my SFINAE 😸 You know, I always try to avoid it, but this time I just could not resist the magic of _Substitution Failure Is Not An Error_  😄 What do you think?

dzenanz (approved):

N-Dekker:
Hmmm... 🤔  this pull request may not add much if we _only_ want to support the MetaDataObject template instantiations declared at https://github.com/InsightSoftwareConsortium/ITK/blob/24e7b452b4d607154e68bb6fdd2ef5bbb5c2d1b4/Modules/Core/Common/include/itkMetaDataObject.h#L298-L321  Do we?

Modules/Core/Common/include/itkMetaDataObject.h
```diff
@@ -134,7 +134,7 @@ class ITK_TEMPLATE_EXPORT MetaDataObject : public MetaDataObjectBase
   friend bool
   operator==(const Self & lhs, const Self & rhs)
   {
-    return lhs.m_MetaDataObjectValue == rhs.m_MetaDataObjectValue;
+    return EqualMetaDataObjects(&lhs.m_MetaDataObjectValue, &rhs.m_MetaDataObjectValue);
```
Leengit, line 137: I would consider passing these in by reference rather than as `const` pointers.  That mimics typical `operator==` convention.  Also, it discourages `nullptr` as an argument to `EqualMetaDataObjects`.

Modules/Core/Common/include/itkMetaDataObject.h
```diff
@@ -149,6 +149,24 @@ class ITK_TEMPLATE_EXPORT MetaDataObject : public MetaDataObjectBase
...
+   * \note The template parameter and the trailing return type are there, just to enable SFINAE.*/
+  template <typename TMetaDataObject>
+  static auto
+  EqualMetaDataObjects(const TMetaDataObject * const lhs, const TMetaDataObject * const rhs) -> decltype(*lhs == *rhs)
+  {
+    static_assert(std::is_same<TMetaDataObject, MetaDataObjectType>::value, "Check the template argument.");
+    return *lhs == *rhs;
+  }
+
+  /** Overload for MetaDataObject types that do not have an `operator==`.*/
```
Leengit, line 162: I fear that in the case that `operator==` does exist, this code will produce two versions of `EqualMetaDataObjects`.  If the testing code works on all platforms then those two versions are not considered redefinitions of each other, but it may be difficult to know which of the two the compiler is choosing.  If the testing is working then maybe my fears are unfounded.

If `EqualMetaDataObjects` were a class instead of a function then we could put the second version here as the default implementation and the first version here as a template specialization.

Modules/Core/Common/include/itkMetaDataObject.h
```diff
@@ -149,6 +149,24 @@ class ITK_TEMPLATE_EXPORT MetaDataObject : public MetaDataObjectBase
...
+  static bool
+  EqualMetaDataObjects(const void * const lhs, const void * const rhs)
+  {
+    return std::memcmp(lhs, rhs, sizeof(MetaDataObjectType)) == 0;
```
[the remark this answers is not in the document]
Leengit: Never mind.  I see that `MetaDataObjectType` is the template parameter of the class.  (Even though it does not start with `T`.)

Modules/Core/Common/include/itkMetaDataObject.h
N-Dekker: @Leengit  In general, I also prefer pass by reference rather than by pointer. But `EqualMetaDataObjects` is just an internal private helper function overload set, so it does not matter so much.  And in this case it is actually _essential_ that the parameters are pointer! The idea is that by providing an overload for void pointers, the compiler must do overload resolution. The overload for void pointers is less specific than the one for `const TMetaDataObject *`, so the compiler will always select the one for `const TMetaDataObject *`... except when the one for one for `const TMetaDataObject *` is SFINAE-ed away!

So only when SFINAE has _removed_ the `const TMetaDataObject *` specific overload, the compiler will select the void pointer overload to be called. OK?

SFINAE removes the `const TMetaDataObject *` specific overload when the expression `decltype(*lhs == *rh)` is invalid ("substitution failure") for that specific `TMetaDataObject` type.

Modules/Core/Common/include/itkMetaDataObject.h
N-Dekker: > If the testing code works on all platforms then those two versions are not considered redefinitions of each other, but it may be difficult to know which of the two the compiler is choosing. If the testing is working then maybe my fears are unfounded.

Just to be 100% sure, I did specifically extend the existing `MetaDataDictionary.Equal` unit tests to also check comparing dictionaries after storing an `std::string`. Because `std::string` is an obvious example for which `operator==` is very different from a bitwise comparison (memcmp). Luckily, the tests are passing successfully 😃

https://github.com/InsightSoftwareConsortium/ITK/blob/d84e49daa45429f3b2a6acc7efc25bfdb3a1129e/Modules/Core/Common/test/itkMetaDataDictionaryGTest.cxx#L290

N-Dekker:
This PR was intended to more easily allow storing simple data structs into the `itk::MetaDataDictionary`, like `mat44` from: https://github.com/InsightSoftwareConsortium/ITK/blob/f372040a21058ac1d2ae71407b1b8bb4c49bddb4/Modules/ThirdParty/NIFTI/src/nifti/niftilib/nifti1_io.h#L58-L60

The idea was that it would allow passing such a data struct directly to `EncapsulateMetaData`, instead of having to copy it to another type of object, like `std::vector`, before calling  `EncapsulateMetaData`, as in pull request #3242 commit 9eb4c1209d215eed244e5e0a1d8f15942a61cb76:

https://github.com/InsightSoftwareConsortium/ITK/blob/9eb4c1209d215eed244e5e0a1d8f15942a61cb76/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx#L1002-L1010

Those nine lines of code could then be replaced by one line, as follows:

    EncapsulateMetaData(thisDic, "qto_xyz", nim->qto_xyz);

Of course, if such simple data structs are not meant to be supported by `itk::MetaDataDictionary`, and if all supported types already have an `operator==`, then this PR may be abandoned anyway.

dzenanz:
This PR is probably useful, even if its immediate use no longer exists.

N-Dekker:
> This PR is probably useful, even if its immediate use no longer exists.

It would actually useful if we would add an instantiation of `MetaDataObject<mat44>` to the list of supported/exported instantiations, at:
 
https://github.com/InsightSoftwareConsortium/ITK/blob/24e7b452b4d607154e68bb6fdd2ef5bbb5c2d1b4/Modules/Core/Common/include/itkMetaDataObject.h#L298-L321

Is `itk::MetaDataDictionary` really _only_ supported for those exported instantiations? I guess `itk::MetaDataDictionary` would also work fine for non-exported types, at least for an internally used (non-exported) dictionary.

N-Dekker:
> Sure, it'd be nice if `itk::MetaDataDictionary` could support additional types, like say `itk::Matrix`, or 2D C arrays, or whatever, but not third party structs that might possibly change.

I think it would be feasible to add support for N-dimensional C-style arrays to `itk::MetaDataDictionary`. This PR currently certainly would not be sufficient. The implementation of `SetMetaDataObjectValue` would also need to be adjusted, in order to support C-style arrays: 

https://github.com/InsightSoftwareConsortium/ITK/blob/f84720ee0823964bd135de8eb973acc40c1e70e1/Modules/Core/Common/include/itkMetaDataObject.hxx#L55-L60

In pseudo code, it might be:

    if MetaDataObjectType is an array-type
      CopyArray(m_MetaDataObjectValue, newValue);
    else
      m_MetaDataObjectValue = newValue; // Simple assignment.

This would allow encapsulating "qto_xyz" as a 2-D C-style array, as follows (mind the `.m`, referring to the array data of `mat44`):

    EncapsulateMetaData(thisDic, "qto_xyz", nim->qto_xyz.m);

seanm:
> `EncapsulateMetaData(thisDic, "qto_xyz", nim->qto_xyz.m);`

Right; storing `nim->qto_xyz.m` I don't object to, but storing `nim->qto_xyz` I do object to.

N-Dekker:
Related PR: #3320 "ENH: Support C-style arrays as MetaDataObjectType for MetaDataDictionary"

dzenanz:
Should we close it, or keep at as draft indefinitely?

N-Dekker:
> Should we close it, or keep at as draft indefinitely?

Close-without-merge would be OK to me now. Note that the current implementation conflicts with 7299e5c59a2f63f5e3de7fcbc193259fc4a61873 (Support C-style arrays as MetaDataObjectType for MetaDataDictionary), but of course, this conflict could be resolved whenever necessary.

PranjalSahu:
I am facing an issue related to this in Python wrappings.
While converting ITK image to xarray we need the metadata. 

But for "**qto_xyz**" key the **dict** method fails for the Python object.
Refer: https://github.com/InsightSoftwareConsortium/ITK/issues/3520

I wanted to know what would be the best way to handle this in ITK Python.
From my understanding, we probably need Matrix44 wrapping to handle this similar 
to the issue https://github.com/InsightSoftwareConsortium/ITK/pull/2270

Issue #3310, opened 2022-03-17
Serialization Test Failures in Windows CI


tbirdso:
### Description

[CDash results](https://open.cdash.org/viewTest.php?onlyfailed&buildid=7800309) show failures in Mesh and Point Set serialization Python tests.

### Steps to Reproduce

`ctest -C Release -R MeshSerialization`

### Expected behavior

Test passes

### Actual behavior

```
Traceback (most recent call last):
  File "D:\a\1\s-build\Wrapping\Generators\Python\itk\support\template_class.py", line 525, in __getitem__
    this_item = self.__template__[key]
KeyError: (<itkCType unsigned long>, <class 'itk.itkPointPython.itkPointF3'>)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "D:/a/1/s/Modules/Core/Mesh/wrapping/test/itkMeshSerializationTest.py", line 36, in <module>
    v_point = itk.VectorContainer[itk.UL, PointType].New()
  File "D:\a\1\s-build\Wrapping\Generators\Python\itk\support\template_class.py", line 529, in __getitem__
    raise itk.TemplateTypeError(self, key)
itk.support.extras.TemplateTypeError: itk.VectorContainer is not wrapped for input type `itk.UL, itk.Point[itk.F,3]`.
```

### Reproducibility

100%

### Versions

main branch

### Environment

Windows 10, Python 3.8.5

### Additional Information

Issue stems from inconsistent `IdentifierType` wrappings across platforms where `ITKM_IT` may be replaced with `itk.UL` or `itk.ULL` depending on architecture. This leads to difficulty in designing tests for objects such as `VectorContainer` that require an `IdentifierType` key in their template.

tbirdso:
I missed the open PR, thanks @dzenanz !

tbirdso:
Thanks @PranjalSahu.

Probably unrelated to enums, but as a note on the previous resolution for posterity, a few ITK Python constructors have slightly different wrappings on Windows vs Linux due to different `IdentifierType` specifiers for `unsigned long` vs `unsigned long long` system architecture. `IdentityType` manglings are described in the [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) on page 223 under the wrapping table. A good way to approach this is what @PranjalSahu did in #3309 with an inline platform check during the test.

I'm not sure that `itk.UL` vs `itk.ULL` will account for differences in enum behavior, though.

[...]

PR #3316, opened 2022-03-20
ENH: Handle itk.ULL in _get_itk_pixelid for Windows

PranjalSahu:
Added conditional long type in itk.support.extras._get_itk_pixelid for Windows.
The failing MeshSerialization test is also modified.
Refer https://github.com/InsightSoftwareConsortium/ITK/issues/3310

### DIFF
--- a/Wrapping/Generators/Python/itk/support/extras.py
+++ b/Wrapping/Generators/Python/itk/support/extras.py
@@ -260,13 +260,18 @@ def _get_itk_pixelid(numpy_array_type):
     """Returns a ITK PixelID given a numpy array."""
 
     import itk
+    def _long_type():
+        if os.name == "nt":
+            return itk.ULL
+        else:
+            return itk.UL
 
     # This is a Mapping from numpy array types to itk pixel types.
     _np_itk = {
         np.uint8: itk.UC,
         np.uint16: itk.US,
         np.uint32: itk.UI,
-        np.uint64: itk.UL,
+        np.uint64: _long_type(),
         np.int8: itk.SC,
         np.int16: itk.SS,
         np.int32: itk.SI,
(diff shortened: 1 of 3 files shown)


Wrapping/Generators/Python/itk/support/extras.py
```diff
@@ -260,13 +260,18 @@ def _get_itk_pixelid(numpy_array_type):
     """Returns a ITK PixelID given a numpy array."""
 
     import itk
+    def _long_type():
+        if os.name == "nt":
+            return itk.ULL
+        else:
+            return itk.UL
```
tbirdso, line 267: Looking good! I wasn't sure about using `os.name` here at first, but on investigation we are doing the exact same platform check to define the width of `itk.UL` in [`types.py`](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Wrapping/Generators/Python/itk/support/types.py#L125). So, carry on 😄

tbirdso (commented):
Type changes look good! `itkMeshSerializationTest` still failed on my Windows machine, but it looks like a data pathing issue. Could you please add a commit changing the [test input data path](itkMeshSerializationTest) to `DATA{${ITK_DATA_ROOT}/Input/mushroom.vtk}`?

Wrapping/Generators/Python/itk/support/extras.py
PranjalSahu: Yes, that's what we are following everywhere.

PranjalSahu: Changed the pathname. I tested it locally with a different vtk file.

tbirdso (approved):

PR #3320, opened 2022-03-21
ENH: Support C-style arrays as MetaDataObjectType for MetaDataDictionary

N-Dekker:
Allows passing a N-dimensional C-style array as value, when calling
`itk::EncapsulateMetaData(dictionary, key, value)`.

### DIFF
--- a/Modules/Core/Common/include/itkMetaDataObject.h
+++ b/Modules/Core/Common/include/itkMetaDataObject.h
@@ -134,7 +134,7 @@ public:
   friend bool
   operator==(const Self & lhs, const Self & rhs)
   {
-    return lhs.m_MetaDataObjectValue == rhs.m_MetaDataObjectValue;
+    return Self::EqualValues(lhs.m_MetaDataObjectValue, rhs.m_MetaDataObjectValue);
   }
 
   /** Returns (metaDataObject1 != metaDataObject2). */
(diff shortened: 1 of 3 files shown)


Leengit (approved):
Looks good.

dzenanz (approved):

N-Dekker:
FYI: [This force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/b828da6287689e40f4716ef55fdd85fb4130235d..8ceb12de42426a717d136a95536041b1bbbf0df5) only adjusted the unit test and the commit text, not the actual implementation.

Modules/Core/Common/include/itkMetaDataObject.h
```diff
@@ -149,6 +149,52 @@ class ITK_TEMPLATE_EXPORT MetaDataObject : public MetaDataObjectBase
...
+  template <typename TValue>
+  static auto
+  Assign(TValue & target, const TValue & source) -> decltype(target = source)
+  {
+    return target = source;
+  }
+
+  /** `Assign` overload for C-style arrays (as well as arrays of arrays). */
+  template <typename TValue, size_t VNumberOfElements>
+  static void
```
Leengit, line 163: Should / can this return a reference to the left hand side?

N-Dekker: @Leengit Thanks for asking. The `Assign` overload for C-style arrays might indeed return a reference to the left hand side, but I would not say it _should_. Because it isn't really necessary. Note that it is just an internal private member function (template). Technically, it is not a template specialization of the other `Assign(target, source)`, it is just an overload. (If it wound have been a template specialization, it would be required to have the same return type as the generic function template. But that's not the case here.)

PR #3326, opened 2022-03-23
Add `itk::ConvertNumberToString(val)` convenience function template

N-Dekker:
Offers full precision conversion from number to string, just like an `itk::NumberToString` function object, but having a convenient function interface, like `std::to_string`.

### DIFF
--- a/Modules/Core/Common/include/itkNumberToString.h
+++ b/Modules/Core/Common/include/itkNumberToString.h
@@ -74,6 +74,18 @@ public:
   }
 };
 
+
+/** Converts the specified numeric value to a string, using a `NumberToString` function object. Produces a full
+ * precision (lossless) string representation, also for `float`and `double`. Otherwise it is similar to `std::to_string`
+ */
+template <typename TValue>
+std::string
+ConvertNumberToString(const TValue val)
+{
+  constexpr NumberToString<TValue> convert{};
+  return convert(val);
+}
+
 } // namespace itk
 
 #ifndef ITK_MANUAL_INSTANTIATION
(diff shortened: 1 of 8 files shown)


Modules/Core/Common/include/itkNumberToString.h
```diff
@@ -74,6 +74,18 @@ class NumberToString<void>
...
 
+
+/** Converts the specified numeric value to a string, using a `NumberToString` function object. Produces a full
+ * precision (lossless) string representation, also for `float`and `double`. Otherwise it is similar to `std::to_string`
+ */
+template <typename TValue>
+std::string
+ConvertNumberToString(const TValue val)
+{
+  constexpr NumberToString<TValue> convert{};
```
hjmjohnson, line 85: I have not thought the implications through carefully.  I am just bringing up that perhaps this is a case where a static constexpr might make sense?

https://stackoverflow.com/a/13867690

N-Dekker: > https://stackoverflow.com/a/13867690 (Does static constexpr variable inside a function make sense?)

@hjmjohnson Interesting link, thanks. I'm a bit surprised about it, as I would in general be reluctant to "claim" static memory in a function, unless it is _really_ necessary. Note that in this particular case, a `NumberToString<TValue>` is pretty much empty, and it is constructed and destructed in no-time, so I would think it's OK to leave it non-static anyway.

BTW, I would also be OK with just creating the `NumberToString<TValue>` function object as a temporary, on the fly, instead of declaring that empty little `convert` object:

    return NumberToString<TValue>()(val);

Or:

    return NumberToString<TValue>{}(val);

If you would like that.

dzenanz: I am convinced that the performance difference will be minimal, and that compiler will optimize away a lot of stuff either way.

dzenanz (approved):

hjmjohnson (approved):
sorry, I meant to approve this earlier.

dzenanz:
ITK.Windows failure is spurious.

PR #3327, opened 2022-03-23
STYLE: Encapsulate NiftiImage qto_xyz mat44 data as `Matrix<float,4,4>`

N-Dekker:
Using `itk::Matrix<float, 4, 4>` (instead of `std::vector<float>`) to store the "qto_xyz" data from m_NiftiImage, as this `Matrix` template instantiation more closely resembles the original niftilib `mat44` data structure.

Follow-up to pull request #3242 commit 9eb4c1209d215eed244e5e0a1d8f15942a61cb76 "ENH: bug #3241: Add qfac and qto_xyz to itkNiftiImageIO metadata" by Sean McBride (@seanm), which was merged on March 17, 2022.

### DIFF
--- a/Modules/Core/Common/include/itkMetaDataObject.h
+++ b/Modules/Core/Common/include/itkMetaDataObject.h
@@ -364,6 +364,7 @@ extern template class ITKCommon_EXPORT_EXPLICIT MetaDataObject<Array<char>>;
 extern template class ITKCommon_EXPORT_EXPLICIT MetaDataObject<Array<int>>;
 extern template class ITKCommon_EXPORT_EXPLICIT MetaDataObject<Array<float>>;
 extern template class ITKCommon_EXPORT_EXPLICIT MetaDataObject<Array<double>>;
+extern template class ITKCommon_EXPORT_EXPLICIT MetaDataObject<Matrix<float, 4, 4>>;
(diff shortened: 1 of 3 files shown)


dzenanz (approved):

seanm:
OK, give me a chance to try this in my app, want to make sure it actually works for me...

N-Dekker:
> OK, give me a chance to try this in my app, want to make sure it actually works for me...

Thanks @seanm  I'm not in a hurry  😃

If you want be sure than it does not get merged accidentally, you may do `Review changes  -> Request changes`, and then switch to `Review changes  -> Approve` once you're sure it works for you.

BTW, instead of using `Matrix<float,4,4>`, we might now also store the data of "qto_xyz" as a raw C-style array `float[4][4]`, if you would like that better. But I think using `Matrix<float,4,4>` is most convenient in this case, as `itk::Matrix<T,Rows,Columns>` is a regular, copyable C++ class type, supporting the most common matrix operations, and interaction with `vnl_matrix`

seanm:
OK, I've tested, this is good for me.

(Calling it "STYLE" seems a little light though, it is a behaviour and public API change, even though the old way didn't ship in a release.)

hjmjohnson (approved):

PR #3328, opened 2022-03-23
Simplify wrapping code

hjmjohnson:
Attempting to simplify wrapping codebase.  Minimize the variables that are modified in the global namespace.

## PR Checklist


thewtex (approved):

Modules/Bridge/NumPy/wrapping/CMakeLists.txt
```diff
@@ -2,6 +2,5 @@ if(ITK_WRAP_PYTHON)
   itk_wrap_module(ITKBridgeNumPy)
   set(WRAPPER_SUBMODULE_ORDER itkPyBuffer itkPyVectorContainer)
   set(WRAPPER_SWIG_LIBRARY_FILES ${WRAPPER_SWIG_LIBRARY_FILES} "${CMAKE_CURRENT_BINARY_DIR}/PyBuffer.i")
-  itk_auto_load_submodules()
-  itk_end_wrap_module()
+  itk_auto_load_and_end_wrap_submodules()
```
dzenanz, line 5: Should this type of change be added to migration guide?

hjmjohnson: The old code is still backward compatible.  Using itk_auto_load_and_end_wrap_submodules may be preferred.  If we deprecate the old macros, then the migration guide should be updated.

dzenanz: Should we update developer style guide section in the software guide?

hjmjohnson: ?? The wrapping code is such a mess, and it still has quite a few cleanups that we are still working on to simplify it.

At this point, I don't know what to recommend.  I think there is a reasonable chance that itk_auto_load_and_end_wrap_submodules() could be a temporary incremental improvement on our path to a simpler wrapping infrastructure.

PR #3329, opened 2022-03-24

jhlegarreta:
- ENH: Increase coverage for miscellaneous classes
- BUG: Fix gradient direction vector index 
- BUG: Fix Superclass name in RTTI macro
- STYLE: Prefer using `ITK_TRY_EXPECT_NO_EXCEPTION` macro in tests
- STYLE: Rename `itkImageToHistogramFilterTest3` to honor tested class

## PR Checklist


Modules/Filtering/DiffusionTensorImage/test/itkDiffusionTensor3DReconstructionImageFilterTest.cxx
```diff
@@ -118,7 +118,33 @@ itkDiffusionTensor3DReconstructionImageFilterTest(int argc, char * argv[])
       gradientDirection[2] = gradientDirections[i][2];
       tensorReconstructionFilter->AddGradientImage(gradientDirection, gradientImage);
       std::cout << "Gradient directions: " << gradientDirection << std::endl;
+
+      const TensorReconstructionImageFilterType::GradientDirectionType::element_type epsilon = 1e-3;
```
jhlegarreta, line 122: Not sure why such a low precision is required here, but it was the case when checking the expected and obtained values.

Modules/Filtering/DiffusionTensorImage/test/itkDiffusionTensor3DReconstructionImageFilterTest.cxx
```diff
@@ -118,7 +118,33 @@ itkDiffusionTensor3DReconstructionImageFilterTest(int argc, char * argv[])
...
+          std::cerr.precision(static_cast<int>(itk::Math::abs(std::log10(epsilon))));
+          std::cerr << "Test failed!" << std::endl;
+          std::cerr << "Error in gradientDirection [" << i << "]"
+                    << "[" << j << "]" << std::endl;
+          std::cerr << "Expected value " << gradientDirectionComponent << std::endl;
+          std::cerr << " differs from " << outputComponent;
+          std::cerr << " by more than " << epsilon << std::endl;
+          return EXIT_FAILURE;
+        }
+      }
```
jhlegarreta, line 141: Simplifying blocks like the one above, where a loop is required, through some testing macro would be great. Ideally, an `ITK_TEST_SET_GET_VALUE`-like being able to deal with ND arrays, including floating point comparisons, would make things easier in tests and more consistent.

Cross-referencing issue #1273.

Modules/Filtering/DiffusionTensorImage/include/itkDiffusionTensor3DReconstructionImageFilter.h
```diff
@@ -251,7 +251,7 @@ class ITK_TEMPLATE_EXPORT DiffusionTensor3DReconstructionImageFilter
     {
       itkExceptionMacro(<< "Gradient direction " << idx << "does not exist");
     }
-    return m_GradientDirectionContainer->ElementAt(idx + 1);
+    return m_GradientDirectionContainer->ElementAt(idx);
```
jhlegarreta, line 254: Wondering it the Slicer3D folks use this diffusion class at all, but this might be interesting in case they do.

thewtex (approved):
:green_heart:

Modules/Filtering/DiffusionTensorImage/test/itkDiffusionTensor3DReconstructionImageFilterTest.cxx
dzenanz: @blowekamp Does something like this exist in GTest? Or maybe some test helpers you wrote?

Modules/Registration/RegistrationMethodsv4/test/itkTimeVaryingBSplineVelocityFieldImageRegistrationTest.cxx
```diff
@@ -103,6 +104,10 @@ PerformTimeVaryingBSplineVelocityFieldImageRegistration(int argc, char * argv[])
   {
     learningRate = std::stod(argv[9]);
   }
+  if (argc >= 12)
+  {
+    convergenceWindowSize = static_cast<unsigned int>(std::stoi(argv[11]));
```
dzenanz, line 109: Why not directly use `std::stoul`?

dzenanz (approved):

jhlegarreta:
> @blowekamp Does something like this exist in GTest? Or maybe some test helpers you wrote?

I do not think GTest is the solution to this, sorry. That would involve translating either the whole test or this part, none of which are desirable IMO. It was more like a open-ended, wish list-like thought.

Modules/Filtering/DiffusionTensorImage/test/itkDiffusionTensor3DReconstructionImageFilterTest.cxx
blowekamp: This is the closes thing that I am aware of: https://github.com/InsightSoftwareConsortium/ITK/blob/dba14cdb283ba6c940a96e3f2199044c4dbff848/Modules/Core/TestKernel/include/itkGTestPredicate.h#L33-L34

dzenanz: So I correctly remembered that you wrote something similar a few years ago 😄

Modules/Registration/RegistrationMethodsv4/test/itkTimeVaryingBSplineVelocityFieldImageRegistrationTest.cxx
jhlegarreta: :+1: My bad; I did not know about `std::stoul`. Done in 417c01e. I will need to review the appearances in my previous PRs.

dzenanz (approved):

hjmjohnson (approved):

PR #3330, opened 2022-03-24
ENH: Add configuration to disable Python factory loading

tbirdso:
Allow user to disable default loading of object factory override methods in Python. This is primarily intended to assist in debug investigations where factory overrides are suspected of interfering with expected behavior.

Behavior can be configured in two lines:
```py
import itkConfig
itkConfig.DefaultFactoryLoading = False
```

With default factory loading disabled a developer is responsible for manually supplying any implementations of abstract classes that previously relied on factory instantiation. For example, `itk.ImageFileReader` relies on factory overrides of `ImageIO` to read files based on their extension, so with no factories registered the developer must supply a suitable `ImageIO` object, i.e. : 
```py
image = itk.imread('path/to/image.mha', imageio=itk.MetaImageIO.New())
```

The `PythonNoDefaultFactories` test is added to cover the scenario where `DefaultFactoryLoading` is set to `False`. `DefaultFactoryLoading` is `True` by default and factory loading is covered by other tests.

## PR Checklist


thewtex (approved):

thewtex:
@jcfr this, along with #3331 will allow running your test pipeline in Slicer without loading HDF5 IO

dzenanz (approved):

PR #3331, opened 2022-03-24
COMP: Do not use ITK_AUTOLOAD_PATH when ITK_WRAP_PYTHON is set

thewtex:
For the use case of Slicer using ITK Python wheels, ITK_AUTOLOAD_PATH
is set, and the Python wheel binaries will load the Slicer ITK binaries,
which will cause a crash.

Dynamic loading in Python is unnecessary and can be performed in a more
explicit and controlled manner by Python scripting, so disable this
feature by default when ITK_WRAP_PYTHON is set.

### DIFF
--- a/Modules/Core/Common/src/itkObjectFactoryBase.cxx
+++ b/Modules/Core/Common/src/itkObjectFactoryBase.cxx
@@ -262,7 +262,7 @@ ObjectFactoryBase::Initialize()
     m_PimplGlobals->m_Initialized = true;
     ObjectFactoryBase::InitializeFactoryList();
     ObjectFactoryBase::RegisterInternal();
-#ifdef ITK_DYNAMIC_LOADING
+#if defined(ITK_DYNAMIC_LOADING) && !defined(ITK_WRAPPING)
     ObjectFactoryBase::LoadDynamicFactories();
 #endif
   }


Modules/Core/Common/src/itkObjectFactoryBase.cxx
```diff
@@ -262,7 +262,7 @@ ObjectFactoryBase::Initialize()
     m_PimplGlobals->m_Initialized = true;
     ObjectFactoryBase::InitializeFactoryList();
     ObjectFactoryBase::RegisterInternal();
-#ifdef ITK_DYNAMIC_LOADING
+#if !defined(ITK_DYNAMIC_LOADING) && !defined(ITK_WRAPPING)
```
jcfr, line 265:
```suggestion
#if defined(ITK_DYNAMIC_LOADING) && !defined(ITK_WRAPPING)
```

This would match the following table

| `ITK_DYNAMIC_LOADING` | `ITK_WRAPPING`     | Call `LoadDynamicFactories` | 
|-----------------------|--------------------|-----------------------------|
| :x:                   | :x:                |  :no_entry:         |
| :x:                   | :heavy_check_mark: |  :no_entry:         |
| :heavy_check_mark:    | :x:                |  :ok:                |
| :heavy_check_mark:    | :heavy_check_mark: |  :no_entry:         |

jcfr (changes requested):

Modules/Core/Common/src/itkObjectFactoryBase.cxx
tbirdso: +1

thewtex: Yes!

tbirdso (approved):

jcfr (approved):

Issue #2802, opened 2021-10-11
Use ReadImage and WriteImage functions where appropriate


dzenanz:
Setting up a reader or a writer just to use them once in a straightforward way requires 5 lines of code, and it can be accomplished by using the appropriate convenience function. 

`itk::ReadImage` and `itk::WriteImage` are available since ITK 5.2.0. See https://github.com/InsightSoftwareConsortium/ITKSphinxExamples/pull/225
for an illustration of the changes required and https://github.com/InsightSoftwareConsortium/ITKSphinxExamples/issues/226 for further discussion.

This can be done piece-wise. A possible starting point are [bundled examples](https://github.com/InsightSoftwareConsortium/ITK/tree/master/Examples) which are used in the [software guide](https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/). @josetascon did a good job in [ITKSphinxExamples](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples/commits?author=josetascon).

sadhana-r:
I can work on this issue

dzenanz:
Sadhana's effort should move us along, but will probably not fully resolve this. I therefore assigned this to "backlog" milestone.

krupalbhat:
can i work on this issue?

dzenanz:
Yes! #5163 was merged, the torch can pass to you now. Perhaps first consider https://github.com/InsightSoftwareConsortium/ITK/pull/5163#discussion_r1932895575?

Whatever path you take, I suggest first doing the changes on a smaller set of files (~10 or so), get that reviewed, and then apply it to more files.

hjmjohnson:
Largely done. As of `main`, **41 of 45** files under `Examples/` use `itk::ReadImage` / `itk::WriteImage`; only the following 4 still use the legacy `ImageFileReader::New()` / `ImageFileWriter::New()` form:

- \`Examples/Filtering/DiffusionTensor3DReconstructionImageFilter.cxx\`
- \`Examples/Segmentation/LaplacianSegmentationLevelSetImageFilter.cxx\`
- \`Examples/RegistrationITKv4/DeformableRegistration1.cxx\`
- \`Examples/RegistrationITKv4/DeformableRegistration11.cxx\`

These four are non-trivial pipelines (multi-component DTI reader, registration with explicit deformation-field IO) where the convenience wrapper is less of a fit. Closing — any further per-file conversion can be tracked in a fresh, scoped issue if/when desired.

PR #2499, opened 2021-04-21
STYLE: Replace postfix by prefix increment in `for` loops in ITK/Modules

N-Dekker:
Applied to ITK/Modules, at the GNU bash prompt:

    find . \( -iname *.cxx -or -iname *.hxx -or -iname *.h \) -exec perl -pi -w -e 's/(\s+for \(.*; .+); (\w+)\+\+\)/$1; ++$2)/g;' {} \;

(Excluded ITK/Modules/ThirdParty.)

According to Herb Sutter, Andrei Alexandrescu - C++ Coding Standards:
101 Rules, Guidelines, and Best Practices: "Prefer the canonical form of
++ and --. Prefer calling the prefix forms"

Follow-up to:
pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2493
commit f81578bbb6242032b611d7e3e4dca3b7557fbbe0
"STYLE: Replace postfix by prefix increment in `for` loops in Core/Common"

PR #3339, opened 2022-03-25
BUG: Fix failing isAffine check due to insuffcient precision in calculations

### LINKED ISSUE #3333: Small-voxel NIFTI file still failing orthonormal test
I started running into the orthonormal issue (supposedly fixed by #2701) with a file on an earlier ITK version, so I updated to ITK5.3.0rc3 and found it actually is still triggered

### Description
fslhd dump of file, I think this is ok and valid.
```
filename        brain5_t1.nii.gz

sizeof_hdr      348
data_type       UINT16
dim0            3
dim1            549
dim2            746
dim3            611
dim4            0
dim5            0
dim6            0
dim7            0
vox_units       mm
time_units      Unknown
datatype        512
nbyper          2
bitpix          16
pixdim0         0.000000
pixdim1         0.300000
pixdim2         0.300000
pixdim3         0.300000
pixdim4         0.000000
pixdim5         0.000000
pixdim6         0.000000
pixdim7         0.000000
vox_offset      352
cal_max         0.000000
cal_min         0.000000
scl_slope       1.000000
scl_inter       0.000000
phase_dim       0
freq_dim        0
slice_dim       0
slice_name      Unknown
slice_code      0
slice_start     0
slice_end       0
slice_duration  0.000000
toffset         0.000000
intent          Unknown
intent_code     0
intent_name
intent_p1       0.000000
intent_p2       0.000000
intent_p3       0.000000
qform_name      Unknown
qform_code      0
qto_xyz:1       0.300000 0.000000 0.000000 0.000000 
qto_xyz:2       0.000000 0.300000 0.000000 0.000000 
qto_xyz:3       0.000000 0.000000 0.300000 0.000000 
qto_xyz:4       0.000000 0.000000 0.000000 1.000000 
qform_xorient   Left-to-Right
qform_yorient   Posterior-to-Anterior
qform_zorient   Inferior-to-Superior
sform_name      Unknown
sform_code      1
sto_xyz:1       0.300000 0.000000 0.000000 -81.665497 
sto_xyz:2       0.000000 0.300000 0.000000 -79.595703 
sto_xyz:3       0.000000 0.000000 0.300000 -26.607700 
sto_xyz:4       0.000000 0.000000 0.000000 1.000000 
sform_xorient   Left-to-Right
sform_yorient   Posterior-to-Anterior
sform_zorient   Inferior-to-Superior
file_type       NIFTI-1+
file_code       1
descrip         mnc2nii brain5_t1.mnc nifti/brain5_t1.nii
aux_file
```

But ANTs built against ITK5.3.0rc3 still triggers the orthonormal complaint
```
# Found a quick way to trigger the bug
antsApplyTransforms -d 3 -r brain5_t1.nii.gz
Exception caught during reference file reading 

itk::ExceptionObject (0x44efdb0)
Location: "unknown" 
File: /gpfs/fs0/project/m/mchakrav/gdevenyi/quarantine/2019b/ANTs/20220323/src/build/ITKv5/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
Line: 1980
Description: ITK ERROR: ITK only supports orthonormal direction cosines.  No orthonormal definition found!

 file brain5_t1.nii.gz
```

I also have a sister file of this one, with very similar properties, which does pass (?!)
```
filename        brain4_t1.nii.gz

sizeof_hdr      348
data_type       UINT16
dim0            3
dim1            569
dim2            740
dim3            615
dim4            0
dim5            0
dim6            0
dim7            0
vox_units       mm
time_units      Unknown
datatype        512
nbyper          2
bitpix          16
pixdim0         0.000000
pixdim1         0.300000
pixdim2         0.300000
pixdim3         0.300000
pixdim4         0.000000
pixdim5         0.000000
pixdim6         0.000000
pixdim7         0.000000
vox_offset      352
cal_max         0.000000
cal_min         0.000000
scl_slope       1.000000
scl_inter       0.000000
phase_dim       0
freq_dim        0
slice_dim       0
slice_name      Unknown
slice_code      0
slice_start     0
slice_end       0
slice_duration  0.000000
toffset         0.000000
intent          Unknown
intent_code     0
intent_name
intent_p1       0.000000
intent_p2       0.000000
intent_p3       0.000000
qform_name      Unknown
qform_code      0
qto_xyz:1       0.300000 0.000000 0.000000 0.000000 
qto_xyz:2       0.000000 0.300000 0.000000 0.000000 
qto_xyz:3       0.000000 0.000000 0.300000 0.000000 
qto_xyz:4       0.000000 0.000000 0.000000 1.000000 
qform_xorient   Left-to-Right
qform_yorient   Posterior-to-Anterior
qform_zorient   Inferior-to-Superior
sform_name      Unknown
sform_code      1
sto_xyz:1       0.300000 0.000000 0.000000 -83.675797 
sto_xyz:2       0.000000 0.300000 0.000000 -87.853699 
sto_xyz:3       0.000000 0.000000 0.300000 -84.445999 
sto_xyz:4       0.000000 0.000000 0.000000 1.000000 
sform_xorient   Left-to-Right
sform_yorient   Posterior-to-Anterior
sform_zorient   Inferior-to-Superior
file_type       NIFTI-1+
file_code       1
descrip         mnc2nii brain4_t1.mnc nifti/brain4_t1.nii
aux_file
```

```
antsApplyTransforms -d 3 -r brain4_t1.nii.gz --verbose
Using double precision for computations.
Reference image: brain4_t1.nii.gz
=============================================================================
The composite transform comprises the following transforms (in order): 
=============================================================================
WARNING: No transforms found, using identify transform
Default pixel value: 0
```

The only thing that differs between the files is the origin and number of slices in the various directions.

### Steps to Reproduce

1. Use ANTs 1c5ae7c42f1149d26505819a710b50e6867d2f31 to attempt to read these files

"working file" brain4, https://nextcloud.computecanada.ca/index.php/s/5GBCYiJZq7sp5Sz
"not working file" brain5, https://nextcloud.computecanada.ca/index.php/s/MxRA6KN3Jcy54NQ

### Expected behavior

Files read.

### Actual behavior

Orthonormality error

### Reproducibility
 
100%

### Versions

5.3.0rc3

### Environment

Centos 7, intel/2019u4 compiler

gdevenyi:
Fixes #3333

Co-authored-by: Vladimir S. FONOV <vladimir.fonov@gmail.com>

## PR Checklist

- [X]  Added test 

- NO Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- NO Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples) for all new major features (if any)


gdevenyi:
This fixed my issues with my high resolution human brains and rat brains I've been working with.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -1753,21 +1753,26 @@ Normalize(std::vector<double> & x)
 static bool
 IsAffine(const mat44 & nifti_mat)
 {
-  const vnl_matrix_fixed<float, 4, 4> mat(&(nifti_mat.m[0][0]));
+  vnl_matrix_fixed<double, 4, 4> mat;
+
+  for (int i = 0; i < 4; i++)
+    for (int j = 0; j < 4; j++)
```
N-Dekker, line 1759: Instead of `i++` and `j++`, please write `++i` and `++j`, following pull request #2499 commit 971aef211cad336ba75cb2da114fc86e873444e5 "STYLE: Replace postfix by prefix increment in `for` loops in ITK/Modules"

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -1753,21 +1753,26 @@ Normalize(std::vector<double> & x)
...
+      mat[i][j] = (double)nifti_mat.m[i][j];
```
N-Dekker, line 1760: Please either just remove the C-style `(double)` cast here (because the conversion from `float` to `double` can take place safely implicitly), or replace it with the C++11/14 `T{expression}` syntax, using curly braces:

    mat[i][j] = double{ nifti_mat.m[i][j] }; 

Following the C++ Core Guidelines, which discourages C-style `(T)expression` casts, at:
http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#prosafety-type-safety-profile

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -1753,21 +1753,26 @@ Normalize(std::vector<double> & x)
...
+
   // First make sure the bottom row meets the condition that it is (0, 0, 0, 1)
   {
-    float bottom_row_error = itk::Math::abs(mat[3][3] - 1.0f);
+    double bottom_row_error = std::fabs(mat[3][3] - 1.0);
     for (int i = 0; i < 3; ++i)
     {
-      bottom_row_error += itk::Math::abs(mat[3][i]);
+      bottom_row_error += fabs(mat[3][i]);
```
N-Dekker, line 1767: `fabs` should be `std::fabs`, just like three lines above, right?

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -1753,21 +1753,26 @@ Normalize(std::vector<double> & x)
...
     if (bottom_row_error > std::numeric_limits<float>::epsilon())
     {
       return false;
     }
   }
 
-  const float condition = vnl_matrix_inverse<float>(mat.as_matrix()).well_condition();
+  const double condition = vnl_matrix_inverse<double>(mat.as_matrix()).well_condition();
   // Check matrix is invertible by testing condition number of inverse
   if (!(condition > std::numeric_limits<float>::epsilon()))
```
N-Dekker, line 1777: Should `std::numeric_limits<float>` here be replaced with `std::numeric_limits<double>`? I'm just asking, I do not know it myself.

N-Dekker (approved):
@gdevenyi Thanks for addressing my commits, Gabriel!  👍 

You might still consider adjusting the subject line of the commit text to get it at _at most_ 72 chars, for example:

    BUG: Fix failure isAffine due to insufficient precision in calculations

Because currently GitHub breaks the last word of the subject line in two: "caluc……lations" at 
https://github.com/InsightSoftwareConsortium/ITK/pull/3339/commits/abc6b33b1b5bc241a55456382cef32be382ada94

But it's not a big deal to me, of course! Approved anyway!

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -1753,21 +1753,26 @@ Normalize(std::vector<double> & x)
...
+  vnl_matrix_fixed<double, 4, 4> mat;
+
+  for (int i = 0; i < 4; ++i)
+    for (int j = 0; j < 4; ++j)
+      mat[i][j] = double{ nifti_mat.m[i][j] };
+
   // First make sure the bottom row meets the condition that it is (0, 0, 0, 1)
   {
-    float bottom_row_error = itk::Math::abs(mat[3][3] - 1.0f);
+    double bottom_row_error = std::fabs(mat[3][3] - 1.0);
```
N-Dekker, line 1764: Using `std::fabs` is fine to me of course, but was there a specific reason to use it here, instead of `itk::Math::abs`?  Just curious!

Leengit:
Depending upon where I look in ITK, I see different usage for converting from `float` to `double` for a new local variable:
```c++
auto   xd0 = static_cast<double>(xf * xf);
auto   xd1 = double{ xf * xf };
double xd2 = xf * xf;
```
Which is preferred?

hjmjohnson:
> 

I prefer static_cast<double>( ).    It is verbose but very explicit about the desired behavior.

N-Dekker:
> Depending upon where I look in ITK, I see different usage for converting from `float` to `double` for a new local variable:
> 
> ```c++
> auto   xd0 = static_cast<double>(xf * xf);
> auto   xd1 = double{ xf * xf };
> double xd2 = xf * xf;
> ```
> 
> Which is preferred?

Good question 🤔 I would prefer _not_ to use `static_cast<double>(xf * xf)` in that particular case, as it might silently convert from integer to floating point. `static_cast<double>` is too "brute force", when you just want to convert one floating point type to another.

Leengit:
Does the answer change for potentially lossy casts such as between `signed` and `unsigned` or from `double` to `int`, etc?  (N.B. Although the compiler may not know it, hopefully we know that these losses won't happen given the context, or are acceptable.)

N-Dekker:
> Does the answer change for potentially lossy casts such as between `signed` and `unsigned` or from `double` to `int`, etc? (N.B. Although the compiler may not know it, hopefully we know that these losses won't happen given the context, or are acceptable.)

Sure! `static_cast` is _typically_ for potentially lossy casts.  On the other hand, `double{ x }` is lossless.

hjmjohnson:
> > Does the answer change for potentially lossy casts such as between `signed` and `unsigned` or from `double` to `int`, etc? (N.B. Although the compiler may not know it, hopefully we know that these losses won't happen given the context, or are acceptable.)
> 
> Sure! `static_cast` is _typically_ for potentially lossy casts. On the other hand, `double{ x }` is lossless (as checked by the compiler).

I answered giving my preference under the condition of converting from `float` to `double`, and assuming that the conversion was required.   I prefer explicit documentation that the developer understands that the conversion is necessary and desired.  I work with novice developers, and the explicit documentation of the desired behavior is welcome, even when it is redundnant.

Leengit:
And by its omission, am I to assume that you would not use the third form in either the lossy or lossless scenarios?

hjmjohnson:
> And by its omission, am I to assume that you would not use the third form in either the lossy or lossless scenarios?

Personally, I would not.  I know that it is a lossless conversion, but novice developers benefit from having the conversion made painfully obvious.

N-Dekker:
> And by its omission, am I to assume that you would not use the third form in either the lossy or lossless scenarios?

In this particular case, conversion from `float` to `double`:

https://github.com/InsightSoftwareConsortium/ITK/blob/abc6b33b1b5bc241a55456382cef32be382ada94/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx#L1756-L1760

In this case, implicit conversion `mat[i][j] = nifti_mat.m[i][j];`  would be OK to me as well. Implicit conversion is _quite safe_, in C++.

N-Dekker:
> I know that it is a lossless conversion, but novice developers benefit from having the conversion made painfully obvious.

Still, novice developers need to learn about `double{ x }` conversion, as well as about implicit conversion. I'm sorry to say so, but C++ has become quite an advanced and complicated language. We cannot hide that away from novice developers.

N-Dekker:
While I have already approved this PR, I think it would be worth adding a minimal unit test, that checks the fix. Something like this?

From @vfonov at https://github.com/InsightSoftwareConsortium/ITK/issues/3333#issuecomment-1078058122

```
0.300000012, 0, 0, -81.6654968
0, 0.300000012, 0, -79.5957031
0, 0, 0.300000012, -26.6077003
0, 0, 0, 1

```

Would that be the most simplified (minimal) test case?

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
gdevenyi: Because that's what @vfonov dropped in the original thread. Seems ITK uses the  Math::abs universally, so I'll fix that.

N-Dekker (approved):

vfonov:
> 

I made a simple test (also showing the limits of the original code from http://callumhay.blogspot.com/2010/10/decomposing-affine-transforms.html:

```

template<typename T> static bool
IsAffine(float _mat[4][4])
{
  vnl_matrix_fixed<T, 4, 4> mat;

  for(int i=0;i<4;i++)
  for(int j=0;j<4;j++)
    mat[i][j]=static_cast<T>(_mat[i][j]);

  // First make sure the bottom row meets the condition that it is (0, 0, 0, 1)
  {
    T bottom_row_error = std::fabs(mat[3][3] - 1.0);
    for (int i = 0; i < 3; ++i)
    {
      bottom_row_error += fabs(mat[3][i]);
    }
    if (bottom_row_error > std::numeric_limits<T>::epsilon())
    {
      return false;
    }
  }

  const T condition = vnl_matrix_inverse<T>(mat.as_matrix()).well_condition();
  // Check matrix is invertible by testing condition number of inverse
  if (!(condition > std::numeric_limits<T>::epsilon()))
  {
    return false;
  }

  // Calculate the inverse and separate the inverse translation component
  // and the top 3x3 part of the inverse matrix
  const vnl_matrix_fixed<T, 4, 4> inv4x4Matrix = vnl_matrix_inverse<T>(mat.as_matrix()).as_matrix();
  const vnl_vector_fixed<T, 3>    inv4x4Translation(inv4x4Matrix[0][3], inv4x4Matrix[1][3], inv4x4Matrix[2][3]);
  const vnl_matrix_fixed<T, 3, 3> inv4x4Top3x3 = inv4x4Matrix.extract(3, 3, 0, 0);

  // Grab just the top 3x3 matrix
  const vnl_matrix_fixed<T, 3, 3> top3x3Matrix = mat.extract(3, 3, 0, 0);
  const vnl_matrix_fixed<T, 3, 3> invTop3x3Matrix = vnl_matrix_inverse<T>(top3x3Matrix.as_matrix()).as_matrix();
  const vnl_vector_fixed<T, 3>    inv3x3Translation = -(invTop3x3Matrix * mat.get_column(3).extract(3));

  // Make sure we adhere to the conditions of a 4x4 invertible affine transform matrix
  const T     diff_matrix_array_one_norm = (inv4x4Top3x3 - invTop3x3Matrix).array_one_norm();
  const T     diff_vector_translation_one_norm = (inv4x4Translation - inv3x3Translation).one_norm();
  constexpr T normed_tolerance_matrix_close = 1e-2;
  return !((diff_matrix_array_one_norm > normed_tolerance_matrix_close) ||
           (diff_vector_translation_one_norm > normed_tolerance_matrix_close));
}
```

testing results:
```
  float case0[4][4]={
    0.300000012, 0, 0, -8.16654968,
    0, 0.300000012, 0, -7.95957031,
    0, 0, 0.300000012, -2.66077003,
    0, 0, 0, 1
  };

  float case1[4][4]={
    0.300000012, 0, 0, -81.6654968,
    0, 0.300000012, 0, -79.5957031,
    0, 0, 0.300000012, -26.6077003,
    0, 0, 0, 1
  };

  float case2[4][4]={
    0.003000012, 0, 0, -8166549.68,
    0, 0.003000012, 0, -7959570.31,
    0, 0, 0.003000012, -2660770.03,
    0, 0, 0, 1
  };

  std::cerr<<"Case0:"<<std::endl;
  std::cerr<<"float:"<<IsAffine<float>(case0)<<std::endl;
  std::cerr<<"double:"<<IsAffine<double>(case0)<<std::endl;

  std::cerr<<"Case1:"<<std::endl;
  std::cerr<<"float:"<<IsAffine<float>(case1)<<std::endl;
  std::cerr<<"double:"<<IsAffine<double>(case1)<<std::endl;

  std::cerr<<"Case2:"<<std::endl;
  std::cerr<<"float:"<<IsAffine<float>(case2)<<std::endl;
  std::cerr<<"double:"<<IsAffine<double>(case2)<<std::endl;
  ```
  
  Outputs:
  ```
Case0:
float:1
double:1
Case1:
float:0
double:1
Case2:
float:0
double:0
```

N-Dekker:
@vfonov Thanks for the test code!

So I see, case2 (below here) is considered non-affine by both `IsAffine<float>` _and_ `IsAffine<double>`. Is that mathematically correct?  Or is it just a limitation of the method? 

```
  float case2[4][4]={
    0.003000012, 0, 0, -8166549.68,
    0, 0.003000012, 0, -7959570.31,
    0, 0, 0.003000012, -2660770.03,
    0, 0, 0, 1
  };
```

Just curious. Obviously switching from `float` to `double` does improve the precision of IsAffine, so I'm fine with the proposed code change.

hjmjohnson:
> How about the following two; which do you prefer when `xf` is a `float`?
> 
> ```c++
> auto   xd1 = double{ xf * xf };
> double xd3{ xf * xf };
> ```

Very minor preference for the second option.

N-Dekker:
Just recently, I did a few contributions that should make conversions in ITK easier to read: by a new conversion function, `itk::ConvertNumberToString(val)` (PR #3326), as well as by declaring several converting constructors `explicit` (PR #3255, #3237, #3238). I hope that novice developers will benefit from them as well 😃

N-Dekker:
@Leengit @hjmjohnson If you have time, please also check this PR as a whole  😃 !  (And approve, if you think it's OK!)

Leengit (approved):
If you have a test case that works with `double` but fails with `float` it would be good to add that so that anyone who is tempted to change this back (for any reason) will pause and reflect.

gdevenyi:
What remains is I need to figure out how to add a test to exercise the function (modelled after @vfonov's helpful code)

vfonov:
just set all voxels to 0 in your nifti file that triggers this error.

gdevenyi:
> @gdevenyi Thank you for working on this nice change. As soon as the test case is added, please ping me and I'll review it. I'd approve the current changes if they had a test.

Thanks, just need to figure out how to add data and a test :+1:

vfonov:
you can also generate an empty ITK Image with given parameters and save it as a NIFTI file

vfonov:
hmm, looks like files generated by ITK don't have the header that triggers isAffine check.

cookpa:
@vfonov images generated by ITK will always contain a qform, which will be used when `IsAffine` returns false. In #3333, the offending data had a qform_code of 0, which triggers the error.

@gdevenyi, I found [this comment](https://github.com/InsightSoftwareConsortium/ITK/pull/2701#issuecomment-903145452) by @jhlegarreta very helpful when navigating the test data upload.

issakomi:
> Depending upon where I look in ITK, I see different usage for converting from `float` to `double` for a new local variable:
> 
> ```c++
> auto   xd0 = static_cast<double>(xf * xf);
> auto   xd1 = double{ xf * xf };
> double xd2 = xf * xf;
> ```
> 
> Which is preferred?

Is it not safer (assumed _xf_ is `float`)

`double xd3 = static_cast<double>xf * xf;`

With integers it definitely required to cast on the right side, otherwise there is an arithmetic overflow (CWE-190, overflow before widen).

```cpp
#include <iostream>
#include <limits.h>
int main(int, char **)
{
	int x = INT_MAX;
	long long y0 = x * 5;
	long long y1 = (long long)x * 5;
	std::cout << "y0=" << y0 << " y1=" << y1 << std::endl;
	return 0;
}
```

Output: `y0=2147483643 y1=10737418235`

I have to admit that i am not 100% about float/double, i also do like in suggestion. At least this quick tests shows:

```cpp
#include <iostream>
int main(int,char**)
{
	float x = 1e-22f;
	float y = 1e+22f;
	double a = static_cast<double>(y) / x;
	double b = static_cast<double>(y / x);
	double c = y / x;
	double d = double{ y / x };
	std::cout << a << " " << b << " " << c << " " << d << std::endl;
	return 0;
}
```

Output: `1e+44 inf inf inf`

Leengit:
> Is it not safer ...

Agreed 100%.  If overflow, underflow, or loss of precision is likely to be an issue than casting the inputs can be better than casting the outputs.  In that scenario my summary of the above responses to my questions, about what to do if there is to be a named intermediate variable, is
```cpp
double xd{ x };                     // Use if the compiler will know that this is lossless
auto   xd = double{ x };            // same functionality, but slight preference to avoid it

auto   xd = static_cast<double>(x); // Use if the compiler considers this lossy
double xd = x;                      // same functionality, but avoid it
```

Leengit:
> simple `for` loop from zero to _n_

For numeric types, especially basic types, I shy away from the default constructor because it used to be that explicit initialization was required.  Admittedly, that's mostly a prejudice at this point, not so much sound reasoning.

gdevenyi:
Hey friends, maybe you could move the casting discussion to ITK discourse or a separate issue? Thanks!

gdevenyi:
I have to learn to/write a test to exercise this so it doesn't regress in the future.

gdevenyi:
The test I was going to add is a simple loader of a reproducing image I have.

gdevenyi:
Test passes integration 🎉

N-Dekker:
> Test passes integration 🎉

Glad to hear so! Can you please explain, just for my understanding. I'm sorry my knowledge on the "old-style" ITK unit tests is limited. Do I understand correctly that:

1. you uploaded SmallVoxels_AffinePrecision.nii.gz?  
2. running itkNiftiImageIOTest2 with this image (SmallVoxels_AffinePrecision) would cause an "ITK ERROR: ITK only supports orthonormal direction cosines" exception, when compiled with the original itkNiftiImageIO.cxx ?
3. running itkNiftiImageIOTest2 with this image + the proposed itkNiftiImageIO.cxx fix would not have that exception anymore?

hjmjohnson (approved):

hjmjohnson (changes requested):
Actually on second thought I don't think this tests works properly.

I think that a list of images given the the data path are all downloaded, but I think that only the first file is actually processed as input.

I think that a completely new test needs to be made that only tests the new image.

gdevenyi:
@N-Dekker I added a replacement test, it fails if I back out my changes, and passes with the changes added.

gdevenyi:
@hjmjohnson I replaced the use of itkNiftiImageIOTest2 with my own test.

gdevenyi:
Did I miss some step to add a new datafile? I used `git data-upload` and the file is in my account on data.kitware.com
```
1153: Description:  Could not create IO object for reading file /home/vsts/work/1/s/Testing/Data/Input/SmallVoxels_AffinePrecision.nii.gz
1153: The file doesn't exist. 

```

gdevenyi:
Hopefully sorted, I had the absolute test data path instead of the relative one from the NIFTI IO directory.

Modules/IO/NIFTI/test/itkNiftiImageIOTest13.cxx
```diff
@@ -0,0 +1,52 @@
...
+  using PixelType = float;
+  using ImageType = itk::Image<PixelType, Dimension>;
+
+  try
+  {
+    using ReaderType = itk::ImageFileReader<ImageType>;
+    ReaderType::Pointer reader = ReaderType::New();
+    reader->SetFileName(argv[1]);
+    reader->Update();
+    ImageType::Pointer image = reader->GetOutput();
```
N-Dekker, line 42: Please consider using `itk::ReadImage` instead. Just one line of code should so it  😃

Modules/IO/NIFTI/test/itkNiftiImageIOTest13.cxx
```diff
@@ -0,0 +1,52 @@
...
+  }
+  catch (itk::ExceptionObject & err)
```
N-Dekker, line 44: Please catch by "const reference", if possible.

Modules/IO/NIFTI/test/itkNiftiImageIOTest13.cxx
N-Dekker: Maybe some extra checks here to see that the read image is as expected...? Or at least, check that it is a valid image?

gdevenyi: https://github.com/InsightSoftwareConsortium/ITK/issues/2802

Modules/IO/NIFTI/test/itkNiftiImageIOTest13.cxx
N-Dekker: > @hjmjohnson I replaced the use of itkNiftiImageIOTest2 with my own test.

Would it not be possible to re-use itkNiftiImageIOTest2 in this case (instead of adding an extra cxx file)?

N-Dekker: > That's far beyond my C++ experience to know what you're asking.

@gdevenyi  When you're programming C++, you should really know about the difference between "const reference" and non-const reference.

By default, it is recommended in C++ to catch exceptions by const reference:

    catch (const itk::ExceptionObject & err)

As a rule of thumb, only use a non-const reference, when you intend to modify the exception object that you have "caught". 

HTH, Niels

gdevenyi: Done.

Modules/IO/NIFTI/test/itkNiftiImageIOTest13.cxx
N-Dekker: > This test is to explicitly see if the NiftiIO isAffine check fails on the random image I had in my collection which was causing issues

I guess you could still add some checks that the read image is valid, for example: 

    if (image == nullptr)
    {
      std::cerr << "Error: The read image is null!\n";
      return EXIT_FAILURE;
    }

    if (image->GetBufferPointer() == nullptr)
    {
      std::cerr << "Error: the buffer of the read image is null!\n";
      return EXIT_FAILURE;
    }

    if (image->GetBufferedRegion().GetNumberOfPixels() == 0)
    {
      std::cerr << "Error: The read image has no pixels!\n";
      return EXIT_FAILURE;
    }

Right?

Modules/IO/NIFTI/test/CMakeLists.txt
```diff
@@ -96,3 +97,7 @@ itk_add_test(NAME itkNiftiReadWriteDirectionSmallVoxelTest
         DATA{Input/SmallVoxels_nosform.nii.gz}
         ${ITK_TEST_OUTPUT_DIR}
         )
+
+itk_add_test(NAME itkNiftiSmallVoxelsAffinePrecisionTest
+COMMAND ITKIONIFTITestDriver itkNiftiImageIOTest13
+      DATA{Input/SmallVoxels_AffinePrecision.nii.gz})
```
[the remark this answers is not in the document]
gdevenyi: Reverting the change, and running both tests.
```
$ ctest -R AffinePrecision --output-on-failure
Test project /home/gdevenyi/projects/itk-build
    Start 1157: itkNiftiSmallVoxelsAffinePrecisionTest
1/2 Test #1157: itkNiftiSmallVoxelsAffinePrecisionTest ............................***Failed    0.04 sec
ExceptionObject caught !

itk::ExceptionObject (0x563a086a0fa0)
Location: "unknown" 
File: /home/gdevenyi/projects/ITK/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
Line: 1988
Description: ITK ERROR: ITK only supports orthonormal direction cosines.  No orthonormal definition found!



    Start 1158: itkNiftiSmallVoxelsAffinePrecisionTest_fromitkNiftiImageIOTest2
2/2 Test #1158: itkNiftiSmallVoxelsAffinePrecisionTest_fromitkNiftiImageIOTest2 ...   Passed    0.03 sec

50% tests passed, 1 tests failed out of 2

Label Time Summary:
ITKIONIFTI    =   0.07 sec*proc (2 tests)

Total Test time (real) =   0.22 sec

The following tests FAILED:
        1157 - itkNiftiSmallVoxelsAffinePrecisionTest (Failed)
Errors while running CTest

```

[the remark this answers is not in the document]
gdevenyi: > So itkNiftiImageIOTest2 just cannot be used for SmallVoxels_AffinePrecision.nii.gz, for some unknown reason, right?

Correct.

The difference between the two sets of code appears to be
```
    itk::NiftiImageIO::Pointer io = itk::NiftiImageIO::New();
    auto                       imageReader = ImageReaderType::New();
    imageReader->SetImageIO(io);
```

Where the NiftiimageIO ImageIO is explicitly set.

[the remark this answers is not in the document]
N-Dekker: @gdevenyi Indeed, itkNiftiImageIOTest2 does explicitly set ImageIO, but it does also do a "read" without explicitly setting ImageIO. By calling `itk::IOTestHelper::ReadImage`: https://github.com/InsightSoftwareConsortium/ITK/blob/8c2c654f3deb8897f22cf9bbce5df3c6482f9a82/Modules/IO/NIFTI/test/itkNiftiImageIOTest2.cxx#L57

Anyway, it seems to me that the CMake code of itkNiftiIOShouldSucceed was already incorrect. It says: 
```
 itk_add_test(NAME itkNiftiIOShouldSucceed 
       COMMAND ITKIONIFTITestDriver itkNiftiImageIOTest2 
               ${ITK_TEST_OUTPUT_DIR} itkNiftiIOShouldSucceed true DATA{${ITK_DATA_ROOT}/Input/LittleEndian.hdr,LittleEndian.mhd,LittleEndian.img}) 
```
https://github.com/InsightSoftwareConsortium/ITK/blob/6fc6164252922314c4e83db70b6717291e7dd763/Modules/IO/NIFTI/test/CMakeLists.txt#L55-L57

The arguments of itk_add_test don't seem to match correctly with arg1, arg2, and prefix at https://github.com/InsightSoftwareConsortium/ITK/blob/8c2c654f3deb8897f22cf9bbce5df3c6482f9a82/Modules/IO/NIFTI/test/itkNiftiImageIOTest2.cxx#L32-L42 

When I run it locally, `arg1` == "itkNiftiIOShouldSucceed", `arg2` == "true", and `prefix` == "C:/X/bin/I/ITK/ExternalData/Testing/Data/Input/LittleEndian.hdr". While `arg1` should have been the boolean ("true" or "false"), and `arg2` should have been the input file name. Right? 

I think itkNiftiSmallVoxelsAffinePrecisionTest should be more like the following, _if_ you would still like to use the existing itkNiftiImageIOTest2:

```
itk_add_test(NAME itkNiftiSmallVoxelsAffinePrecisionTest
      COMMAND ITKIONIFTITestDriver itkNiftiImageIOTest2
              ${ITK_TEST_OUTPUT_DIR} true DATA{${ITK_DATA_ROOT}/Input/SmallVoxels_AffinePrecision.nii.gz} SomeSpecificPrefix)
```

I don't know what exactly the prefix ("SomeSpecificPrefix") should be. But it should be the last argument to `itk_add_test`.

My two cents, Niels

N-Dekker:
@gdevenyi Would a smaller test image file be possible? I just noticed that the file you added for this test, SmallVoxels_AffinePrecision.nii.gz is more than 1 MB! https://data.kitware.com/#item/627539e24acac99f42c8929a

gdevenyi:
> @gdevenyi Would a smaller test image file be possible? I just noticed that the file you added for this test, SmallVoxels_AffinePrecision.nii.gz is more than 1 MB! https://data.kitware.com/#item/627539e24acac99f42c8929a
> 
> If I understand correctly, the only part of the image that is relevant to this test is the 4x4 matrix at `m_NiftiImage->sto_xyz`, right? That's just 64 bytes.

I took the problematic naturalistic image I had, and zeroed all the voxels to make it maximally compressible. I don't know the NIFTI format well enough to successfully modify the file any further without accidentally breaking the important properties.

cookpa:
Does this work? I just copied the header to a smaller array using RNifti

[smallVoxels_affinePrecision_reduced.nii.gz](https://github.com/InsightSoftwareConsortium/ITK/files/8689659/smallVoxels_affinePrecision_reduced.nii.gz)

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -1753,10 +1753,15 @@ Normalize(std::vector<double> & x)
...
+  vnl_matrix_fixed<double, 4, 4> mat;
+
+  for (int i = 0; i < 4; ++i)
+    for (int j = 0; j < 4; ++j)
+      mat[i][j] = double{ nifti_mat.m[i][j] };
+
   // First make sure the bottom row meets the condition that it is (0, 0, 0, 1)
   {
-    float bottom_row_error = itk::Math::abs(mat[3][3] - 1.0f);
+    double bottom_row_error = itk::Math::abs(mat[3][3] - 1.0);
```
issakomi, line 1764: is _float_ epsilon several lines below (line 1769) intentional
```cpp
if (bottom_row_error > std::numeric_limits<float>::epsilon())
```
I thought that it might have been forgotten.

N-Dekker:
> Does this work? I just copied the header to a smaller array using RNifti
> 
> [smallVoxels_affinePrecision_reduced.nii.gz](https://github.com/InsightSoftwareConsortium/ITK/files/8689659/smallVoxels_affinePrecision_reduced.nii.gz)

@cookpa Thanks Philip. Your reduced nii.gz does trigger the "ITK ERROR: ITK only supports orthonormal direction cosines" when I try to read it by `itk::ReadImage`, using an ITK version _without_ Gabriel's (@gdevenyi) proposed fix. (ITK v5.3rc03, VS2019, Windows.)  And it's just 129 bytes!  😃

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
N-Dekker: > is _float_ epsilon several lines below (line 1769) intentional?

@issakomi Good question. But I wonder, isn't `std::numeric_limits<T>::epsilon()` rather arbitrary anyway? I just don't know, but if `mat[3][i]` must be close to zero (for `i` = 0, 1, 2), then those matrix elements might as well be compared against `std::numeric_limits<T>::min()`, which if much smaller than `std::numeric_limits<T>::epsilon()`.

Honestly I'm just wondering what is intended as well. I have no suggestion here.

cookpa: I think the bottom row should be precisely (0,0,0,1) - but I guess it's possible there might be some epsilon error when the inverse is calculated

vfonov: I think it was just forgotten.

N-Dekker: > there might be some epsilon error when the inverse is calculated

OK, but `std::numeric_limits<T>::epsilon()` isn't just _some_ epsilon error. It is specifically the difference between 1 and the least value greater than 1 that is representable by floating point type `T`. https://www.cplusplus.com/reference/cfloat/ That's why I wonder if this specific epsilon should also be used for values that should be close to zero (rather than close to one). But I just don't know, honestly.

gdevenyi: One further idea here, currently the test is made with the sum as each entry added. Should we instead test the individual entries?

Changing to `::min()`, or testing the individual entries will be substantially stricter than the existing implementation as well.

gdevenyi:
Thanks @cookpa and @N-Dekker. I've updated the test file.

N-Dekker:
>  I've updated the test file.

@gdevenyi That's really great, Gabriel, thanks! Can you possibly also remove the large (1.042 MB) version of SmallVoxels_AffinePrecision.nii.gz from https://data.kitware.com/#item/627539e24acac99f42c8929a ?

N-Dekker:
> > @gdevenyi That's really great, Gabriel, thanks! Can you possibly also remove the large (1.042 MB) version of SmallVoxels_AffinePrecision.nii.gz from https://data.kitware.com/#item/627539e24acac99f42c8929a ?
> 
> As far as I understood, I did.

But it's still there at https://data.kitware.com/#item/627539e24acac99f42c8929a !  Right? Isn't it possible to remove the file from https://data.kitware.com/#item/627539e24acac99f42c8929a ?

Modules/IO/NIFTI/test/itkNiftiImageIOTest13.cxx
N-Dekker: @gdevenyi Did you still consider adding the extra checks that I suggested here? In order to ensure that the reading does actually yield a valid image?

```
if (image == nullptr)
{
  std::cerr << "Error: The read image is null!\n";
  return EXIT_FAILURE;
}

if (image->GetBufferPointer() == nullptr)
{
  std::cerr << "Error: the buffer of the read image is null!\n";
  return EXIT_FAILURE;
}

if (image->GetBufferedRegion().GetNumberOfPixels() == 0)
{
  std::cerr << "Error: The read image has no pixels!\n";
  return EXIT_FAILURE;
}

```

gdevenyi:
@hjmjohnson I believe we're just waiting on you here. Can you please re-review?

Modules/IO/NIFTI/test/itkNiftiImageIOTest13.cxx
dzenanz: This can be condiserably abbreviated via `itk::ReadImage` and `ITK_TRY_EXPECT_NO_EXCEPTION`. We can do that in a follow-up PR. Let's merge this, as it has dragged on for quite a long time.

dzenanz (approved):

PR #3345, opened 2022-03-29
COMP: Allow setting DisplacementField with a VectorImage

HastingsGreer:
`itk::DisplacementFieldTransform` is typically set from an array of displacement vectors. Previously only inputs of `itk::Image[itk::Vector[<type>,<VectorDimension>],<ImageDimension>]` were accepted. 

This PR aims to expose a method for setting the transform from an input of type `itk::VectorImage[<type>,<ImageDimension>]` . 

The difference between an "image of vectors" and a "vector image" is that the former has compile-time fixed-length vector pixels whereas the latter has variable-length vector pixels which have been fixed at a certain length at runtime. 

## PR Checklist
- [ ] Added test (or behavior not changed)  <-- (added Python test, C++ test pending)

### DIFF
--- a/Modules/Filtering/DisplacementField/include/itkDisplacementFieldTransform.h
+++ b/Modules/Filtering/DisplacementField/include/itkDisplacementFieldTransform.h
@@ -21,6 +21,7 @@
 #include "itkTransform.h"
 
 #include "itkImage.h"
+#include "itkVectorImage.h"
 #include "itkMatrixOffsetTransformBase.h"
 #include "itkImageVectorOptimizerParametersHelper.h"
 #include "itkVectorInterpolateImageFunction.h"
@@ -157,6 +158,7 @@ public:
 
   /** Define the displacement field type and corresponding interpolator type. */
   using DisplacementFieldType = Image<OutputVectorType, Dimension>;
+  using VectorImageDisplacementFieldType = VectorImage<TParametersValueType, Dimension>;
   using DisplacementFieldPointer = typename DisplacementFieldType::Pointer;
   using DisplacementFieldConstPointer = typename DisplacementFieldType::ConstPointer;
 
@@ -181,6 +183,9 @@ public:
    * container. */
   virtual void
   SetDisplacementField(DisplacementFieldType * field);
+  virtual void
+               SetDisplacementField(VectorImageDisplacementFieldType * field);
+  virtual void SetDisplacementField(std::nullptr_t) = delete;
   itkGetModifiableObjectMacro(DisplacementField, DisplacementFieldType);
 
   /** Get/Set the inverse displacement field. This must be supplied by the user for
--- a/Modules/Filtering/DisplacementField/include/itkDisplacementFieldTransform.hxx
+++ b/Modules/Filtering/DisplacementField/include/itkDisplacementFieldTransform.hxx
@@ -24,6 +24,7 @@
 #include "itkImageRegionIteratorWithIndex.h"
 #include "vnl/algo/vnl_symmetric_eigensystem.h"
 #include "vnl/algo/vnl_matrix_inverse.h"
+#include "itkCastImageFilter.h"
 
 namespace itk
 {
@@ -344,6 +345,18 @@ DisplacementFieldTransform<TParametersValueType, VDimension>::SetDisplacementFie
 
 template <typename TParametersValueType, unsigned int VDimension>
 void
+DisplacementFieldTransform<TParametersValueType, VDimension>::SetDisplacementField(
+  VectorImageDisplacementFieldType * field)
+{
+  using CasterType = CastImageFilter<VectorImageDisplacementFieldType, DisplacementFieldType>;
+  auto caster = CasterType::New();
+  caster->SetInput(field);
+  caster->Update();
+  this->SetDisplacementField(caster->GetOutput());
+}
+
+template <typename TParametersValueType, unsigned int VDimension>
+void
 DisplacementFieldTransform<TParametersValueType, VDimension>::SetInverseDisplacementField(
   DisplacementFieldType * inverseField)
 {
@@ -477,7 +490,7 @@ DisplacementFieldTransform<TParametersValueType, VDimension>::SetFixedParameters
   }
   if (nullState)
   {
-    this->SetDisplacementField(nullptr);
+    this->SetDisplacementField(static_cast<DisplacementFieldType *>(nullptr));
     this->SetInverseDisplacementField(nullptr);
     return;
   }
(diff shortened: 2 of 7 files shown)


Modules/Filtering/DisplacementField/include/itkDisplacementFieldTransform.hxx
```diff
@@ -341,7 +341,12 @@ DisplacementFieldTransform<TParametersValueType, VDimension>::SetDisplacementFie
   }
   this->SetFixedParametersFromDisplacementField();
 }
-
+template <typename TParametersValueType, unsigned int VDimension>
+void
+DisplacementFieldTransform<TParametersValueType, VDimension>::SetDisplacementField(VectorImage<TParametersValueType, VDimension> * field)
+{
+  return;
```
dzenanz, line 348: Shouldn't this do something? I guess that delegating to the other method does not compile?

Modules/Filtering/DisplacementField/include/itkDisplacementFieldTransform.h
```diff
@@ -181,7 +183,9 @@ class ITK_TEMPLATE_EXPORT DisplacementFieldTransform : public Transform<TParamet
    * container. */
   virtual void
   SetDisplacementField(DisplacementFieldType * field);
-  SetDisplacementField(VectorImage<TParametersValueType, VDimension> * field);
+  virtual void
+               SetDisplacementField(VectorImageDisplacementFieldType * field);
+  virtual void SetDisplacementField(nullptr);
```
dzenanz, line 188: Did you meant `nullptr_t` instead of `nullptr`?

Modules/Filtering/DisplacementField/include/itkDisplacementFieldTransform.hxx
```diff
@@ -341,12 +342,25 @@ DisplacementFieldTransform<TParametersValueType, VDimension>::SetDisplacementFie
...
 void
-DisplacementFieldTransform<TParametersValueType, VDimension>::SetDisplacementField(VectorImage<TParametersValueType, VDimension> * field)
+DisplacementFieldTransform<TParametersValueType, VDimension>::SetDisplacementField(
+  VectorImageDisplacementFieldType * field)
+{
+  using CasterType = CastImageFilter<VectorImageDisplacementFieldType, DisplacementFieldType>;
+  auto caster = CasterType::New();
+  caster->SetInput(field);
+  caster->Update();
+  this->SetDisplacementField(caster->GetOutput());
```
dzenanz, line 355: This makes sense.

Modules/Filtering/DisplacementField/include/itkDisplacementFieldTransform.h
tbirdso: Thanks @dzenanz , this will help make the interface much cleaner

HastingsGreer:
There are several subclasses of itkDisplacementFieldTransform that override SetDisplacementField, in these classes our new overload is hidden:
```
/playpen-ssd/tgreer/ITK/Modules/Filtering/DisplacementField/include/itkDisplacementFieldTransform.hxx:348:1: warning: ‘void itk::DisplacementFieldTransform<TParametersValueType, VDimension>::SetDisplacementField(itk::DisplacementFieldTransform<TParametersValueType, VDimension>::VectorImageDisplacementFieldType*) [with TParametersValueType = double; unsigned int VDimension = 3; itk::DisplacementFieldTransform<TParametersValueType, VDimension>::VectorImageDisplacementFieldType = itk::VectorImage<double, 3>]’ was hidden [-Woverloaded-virtual]
  348 | DisplacementFieldTransform<TParametersValueType, VDimension>::SetDisplacementField(
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /playpen-ssd/tgreer/ITK/Modules/Filtering/DisplacementField/include/itkTimeVaryingBSplineVelocityFieldTransform.h:21,
```

Is this a concern?

jhlegarreta:
Sorry to chime in. @HastingsGreer @tbirdso thanks for doing this, and for having added a Python test 👌. In order to keep the code coverage, can a C++ test be added, please?

tbirdso:
Hi @jhlegarreta , I understand your concern. As these additions are time-sensitive it would be best if we can focus on getting the build passing for now, with changes covered effectively in Python. Can revisit C++ once stable.

Modules/Filtering/DisplacementField/wrapping/test/itkDisplacementFieldTransformTest.py
```diff
@@ -0,0 +1,60 @@
...
+# ==========================================================================*/
+import itk
+itk.auto_progress(2)
+
+ScalarType = itk.F
+VectorDimension = 2
+VectorType = itk.Vector[itk.F,Dimension]
+
+ImageDimension = 2
+ImageType = itk.Image[PixelType, ImageDimension]
```
hjmjohnson, line 26: Should PixelType be VectorType here?

Modules/Filtering/DisplacementField/wrapping/test/itkDisplacementFieldTransformTest.py
```diff
@@ -0,0 +1,60 @@
...
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ==========================================================================*/
+import itk
+itk.auto_progress(2)
+
+ScalarType = itk.F
+VectorDimension = 2
+VectorType = itk.Vector[itk.F,Dimension]
```
hjmjohnson, line 23: Should itk.F be Scalartype here?

Modules/Filtering/DisplacementField/wrapping/test/itkDisplacementFieldTransformTest.py
tbirdso: Yes, good catch. I included these fixes and a few others in the most recent fixup commit (and ran the Python file this time).

tbirdso:
Updated the PR description for clarity.

thewtex:
Instead of passing `nullptr` directly, can we:

```
DisplacementFieldType * empty{ nullptr };
this->SetDisplacementField(empty);

or

// Default construction is a nullptr
auto empty = DisplacementFieldType::New();
this->SetDisplacementField(empty);
```

to pass in a typed pointer?

tbirdso:
@thewtex Yes, I believe the first is equivalent to a `static_cast` and would succeed. I'm not sure that I agree that the second case would return a `nullptr` if `New()` is used? Either way, yes, there is still an interface available to pass in a null value.

thewtex:
:tada: 

Squashing and adding C++ coverage as suggested by @jhlegarreta

thewtex (approved):

PR #3347, opened 2022-03-30
ENH: Adding Support for Transform Serialization

PranjalSahu:
This PR adds support for Transform serialization both composite and individual.

- For individual transforms we rely on itk.template to get the right object type. Transforms like BSpline take additional parameters and those are encoded in the tuple obtained from itk.template.

- However, for transforms inside a composite transform, we have to rely on the GetTransformTypeAsString() to get the right object type as itk.template returns the object of type itk.TransformBase and the information is lost. 

- In a composite transform, BSpline can only take Spline order as 3 (we wrap for Spline order 3 only).
- There are certain transforms that have 2D and 3D inside the name and they don't template parameters so they are handled specially inside the method **special_transform_check.**
-  Most keys are arrays to support for composite transform and no special handling is done for individual or composite transform there. A sample json for Affine Transform:
```{'name': ['transform0'], 'numberOfTransforms': 1, 'dataType': ['float64'], 'inDimension': [3], 'outDimension': [3], 'transformName': ['AffineTransform'], 'transformType': ['D3'], 'parameters': [array([1., 0., 0., 0., 1., 0., 0., 0., 1., 0., 0., 0.])], 'fixedParameters': [array([0., 0., 0.])], 'numberOfParameters': [12], 'numberOfFixedParameters': [3]}```

- A sample json for Composite transform:
``` {'name': ['composite_transform', 'transform0', 'transform1'], 'numberOfTransforms': 2, 'dataType': ['float64', 'float64', 'float64'], 'inDimension': [3, 3, 3], 'outDimension': [3, 3, 3], 'transformName': ['CompositeTransform', 'AffineTransform', 'Rigid3DTransform'], 'transformType': ['D3', 'D33', 'D33'], 'parameters': [array([], dtype=float64), array([1., 0., 0., 0., 1., 0., 0., 0., 1., 0., 0., 0.]), array([1., 0., 0., 0., 1., 0., 0., 0., 1., 0., 0., 0.])], 'fixedParameters': [array([], dtype=float64), array([0., 0., 0.]), array([0., 0., 0.])], 'numberOfParameters': [0, 12, 12], 'numberOfFixedParameters': [0, 3, 3]}```

- We only consider the first transform to be of type composite, similar to the convention in 
https://github.com/InsightSoftwareConsortium/ITK/blob/da7f448799d057db798eed2ce9ccbaccde01d9f7/Modules/IO/TransformInsightLegacy/src/itkTxtTransformIO.cxx#L294

## PR Checklist

### CODE REVIEW

File: Modules/Core/Transform/wrapping/test/CMakeLists.txt [hunk 1]
```diff
@@ -0,0 +1,4 @@
+list(FIND ITK_WRAP_IMAGE_DIMS 3 wrap_3_index)
```
thewtex, line 1: `wrap_3_index` is unused

thewtex (changes requested):
Outstanding work @PranjalSahu !! :star2: :star_struck: 

I love what you have done.

A few requests:

- `dataType` -> `parametersValueType` to be more explicit in what this refers to, i.e. the transform parameters, and to have a connection with the existing naming `TParametersValue`
- Instead of a `dict` with entries that are `lists`, can we have a `list` where each element has a `dict` with entries `["name", "parametersValueType", "transformName", "transformType", "inDimension", "outDimension", "numberOfParameters", "numberOfFixedParameters"]`? If the transform to be serialized is a composite transform, then the component transforms are serialized into the list. If the transform to be serialized is a non-composite transform, then it is a single-entry list. This will help in multiple ways: 1) Individual list entries can be used independently. 2) There is not a risk for client code to create component lists without a consistent number of entries. 3) There is better compatibility with the TransformFileReader / TransformFileWriter.

Modules/Core/Transform/wrapping/test/itkTransformSerializationTest.py
```diff
@@ -0,0 +1,89 @@
...
+
+# Add the above created transforms in the composite transform
+for transform in transform_object_list:
+    composite_transform.AddTransform(transform)
+
+# Check the serialization of composite transform
+serialize_deserialize = pickle.loads(pickle.dumps(composite_transform))
+
+assert serialize_deserialize.GetObjectName() == composite_transform.GetObjectName()
+assert serialize_deserialize.GetNumberOfTransforms(), 4
```
dzenanz, line 65: Shouldn't this be 5?

dzenanz (approved):

PranjalSahu:
Also, we will replace the set parameters with direct copy instead of looping through elements.
I can create an issue for that.

PranjalSahu:
Checking the issue of why DisplacementFieldTransform is giving **AttributeError: module 'itk' has no attribute 'DisplacementFieldTransform'** in windows.
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Wrapping/Generators/Python/itk/support/lazy.py#L131

Is there any difference in how modules are loaded in itk python for windows and linux ?
@thewtex @tbirdso

tbirdso:
@PranjalSahu My guess is that we need to explicitly enable wrappings for the `Module_ITKDisplacementField` module as we did with ITKThresholding in [https://github.com/InsightSoftwareConsortium/ITK/pull/3081](https://github.com/InsightSoftwareConsortium/ITK/pull/3081)

PR #3352, opened 2022-03-31
ENH: Adding Module Dependency for Windows Python Test

PranjalSahu:
Needed for itkTransformSerializationTest.
Other transforms (AffineTransform, Rigid3DTransform, BSplineTransform, QuaternionRigidTransform) are inside 
core so not an issue.


thewtex (approved):
Thanks for the follow-up @PranjalSahu !

PR #3360, opened 2022-04-03
DOC: Use full URL logo in README

thewtex:
To resolve in PyPI package rendering.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-![ITK - The Insight Toolkit](Documentation/Art/itkBannerSmall.png)
+![ITK - The Insight Toolkit](https://raw.githubusercontent.com/InsightSoftwareConsortium/ITK/801370c025c7d296783481779a41c6d559c992c5/Documentation/Art/itkBannerSmall.png)
 
 ITK: The Insight Toolkit
 ========================


dzenanz (approved):

README.md
```diff
@@ -1,4 +1,4 @@
-![ITK - The Insight Toolkit](Documentation/Art/itkBannerSmall.png)
+![ITK - The Insight Toolkit](https://raw.githubusercontent.com/InsightSoftwareConsortium/ITK/801370c025c7d296783481779a41c6d559c992c5/Documentation/Art/itkBannerSmall.png)
```
jhlegarreta, line 1: Why not `https://raw.githubusercontent.com/InsightSoftwareConsortium/ITK/master/Documentation/Art/itkBannerSmall.png`?

If at some point the banner/logo changes, this would guarantee that the new is picked up. Renaming the `master` branch may also happen at some point, but IMO it would be easier to pick.

thewtex: Yes, this issue is renaming the file or changing the path. This path is used in the released Python packages -- if the file no longer exists at the referenced path in `master`, then it will 404 and there is no way to go back into released packages and change the path.

jhlegarreta (approved):

PR #3342, opened 2022-03-28
STYLE: Avoid old-style cast

Leengit:
Replace old-style cast `(int)a` or `int(a)` with new-style `static_cast<int>(a)`, etc.

Avoiding old-style casts on advice of:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es48-avoid-casts
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es49-if-you-must-use-a-cast-use-a-named-cast
https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines-pro-type-cstyle-cast.html

Thank you @N-Dekker for the references.

## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [X] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [X] Added test (or behavior not changed)
- [X] Updated API documentation (or API not changed)
- [X] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [X] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [X] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples) for all new major features (if any)

PR #3361, opened 2022-04-03
STYLE: Remove casts from `MersenneTwisterRandomVariateGenerator::hash`

N-Dekker:
Moved duplicate code into a lambda, replacing two unsafe C-style
pointer casts by a static_cast, and removing two unnecessary
`static_cast<unsigned int>` casts.

### DIFF
--- a/Modules/Core/Common/src/itkMersenneTwisterRandomVariateGenerator.cxx
+++ b/Modules/Core/Common/src/itkMersenneTwisterRandomVariateGenerator.cxx
@@ -89,31 +89,27 @@ MersenneTwisterRandomVariateGenerator::MersenneTwisterRandomVariateGenerator()
 MersenneTwisterRandomVariateGenerator::~MersenneTwisterRandomVariateGenerator() = default;
 
 MersenneTwisterRandomVariateGenerator::IntegerType
-MersenneTwisterRandomVariateGenerator ::hash(time_t t, clock_t c)
+MersenneTwisterRandomVariateGenerator::hash(const time_t t, const clock_t c)
 {
   itkInitGlobalsMacro(PimplGlobals);
   // Get an IntegerType from t and c
   // Better than IntegerType(x) in case x is floating point in [0,1]
   // Based on code by Lawrence Kirby: fred at genesis dot demon dot co dot uk
 
-  IntegerType h1 = 0;
-  auto *      p = (unsigned char *)&t;
+  const auto convert = [](const auto arg) {
+    IntegerType        h{ 0 };
+    const auto * const p = reinterpret_cast<const unsigned char *>(&arg);
 
-  const auto sizeOfT = static_cast<unsigned int>(sizeof(t));
-  for (unsigned int i = 0; i < sizeOfT; ++i)
-  {
-    h1 *= UCHAR_MAX + 2U;
-    h1 += p[i];
-  }
-  IntegerType h2 = 0;
-  p = (unsigned char *)&c;
+    for (size_t i = 0; i < sizeof(arg); ++i)
+    {
+      h *= UCHAR_MAX + 2U;
+      h += p[i];
+    }
+    return h;
+  };
 
-  const auto sizeOfC = static_cast<unsigned int>(sizeof(c));
-  for (unsigned int j = 0; j < sizeOfC; ++j)
-  {
-    h2 *= UCHAR_MAX + 2U;
-    h2 += p[j];
-  }
+  const IntegerType h1 = convert(t);
+  const IntegerType h2 = convert(c);
   return (h1 + m_PimplGlobals->m_StaticDiffer++) ^ h2;
 }
 


N-Dekker:
@Leengit  Another "spin-off" from your pull request #3342  😃

N-Dekker:
FYI, the original code as added by Daniel Blezek (@blezek), commit 352e56087fadc01f52ac9f076554555a0799cbc0, 6 October 2005, did not have those two casts from size_t to `unsigned int` either:  

https://github.com/InsightSoftwareConsortium/ITK/blob/352e56087fadc01f52ac9f076554555a0799cbc0/Code/Common/itkMersenneTwisterRandomVariateGenerator.h#L260-L273

N-Dekker:
Off-topic, here is an explanation of why `UCHAR_MAX + 2U` was used as factor used as factor for multiplication: [Ben Pfaff - How should I pick the value to pass to srand()?](https://benpfaff.org/writings/clc/random-seed.html) saying:
> The factor used for multiplication normally comes out as 257, a prime and therefore a good candidate.

Modules/Core/Common/src/itkMersenneTwisterRandomVariateGenerator.cxx
```diff
@@ -89,31 +89,28 @@ MersenneTwisterRandomVariateGenerator::MersenneTwisterRandomVariateGenerator()
...
-  const auto sizeOfC = static_cast<unsigned int>(sizeof(c));
-  for (unsigned int j = 0; j < sizeOfC; ++j)
-  {
-    h2 *= UCHAR_MAX + 2U;
-    h2 += p[j];
-  }
+  const auto convert = [](const auto arg) {
+    IntegerType        h = 0;
+    const void * const voidPtr = &arg;
+    auto * const       unsignedCharPtr = static_cast<const unsigned char *>(voidPtr);
```
dzenanz, line 102: Why not use reinterpret cast here?

Leengit: Let's document in the conversation here (and maybe in the code too) why it makes sense to do the pointer conversion via an intermediate `void *`.  Or if it is reasonable to avoid `void *`, as the original code does, then let's do that instead, e.g., via the `reinterpret_cast`.

Modules/Core/Common/src/itkMersenneTwisterRandomVariateGenerator.cxx
```diff
@@ -89,31 +89,28 @@ MersenneTwisterRandomVariateGenerator::MersenneTwisterRandomVariateGenerator()
...
-
-  const auto sizeOfC = static_cast<unsigned int>(sizeof(c));
-  for (unsigned int j = 0; j < sizeOfC; ++j)
-  {
-    h2 *= UCHAR_MAX + 2U;
-    h2 += p[j];
-  }
+  const auto convert = [](const auto arg) {
+    IntegerType        h = 0;
+    const void * const voidPtr = &arg;
```
Leengit, line 101: I prefer curly brace construction of these two variables.

Leengit (commented):
It's outside the purview of this pull request, but why does this code seem to ignore big endian vs. little endian considerations?  Is it because the hash code would work either way and we don't care that the class produces different hash codes on big vs. little endian platforms?

dzenanz:
> the hash code would work either way

Yes. The point is to produce pseudo-random numbers, not to produce the same sequence on each platform.

hjmjohnson:
I did think that this implementation produced the same sequence on all platforms given the same seed.

Hans

dzenanz:
You are probably right Hans. It should be the same, since summation is order-independent.

hjmjohnson:
Side note:  https://en.cppreference.com/w/cpp/numeric/random/mersenne_twister_engine. I was curious if the C++11 implementation could replace the custom implementation.  

It seems that @N-Dekker has dome some work to verify that the results are the same already:  Modules/Core/Common/test/itkMersenneTwisterRandomVariateGeneratorGTest.cxx

This is just a curiosity at this point.

Modules/Core/Common/src/itkMersenneTwisterRandomVariateGenerator.cxx
N-Dekker: >Why not use reinterpret cast here?

@dzenanz  Thanks for asking. I was told that the effect of reinterpret_cast is "implementation defined" (compiler dependent), and that it should be avoided whenever there is a reasonably safe alternative. Specifically, [C++ Standard Working Draft n4910](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/n4910.pdf) 7.6.1.10 Reinterpret cast [expr.reinterpret.cast] says:

> A pointer can be explicitly converted to any integral type large enough to hold all values of its type. The
mapping function is implementation-defined.

Implicit conversion of an address to a `const void ` pointer is entirely safe, though. And a cast from such a pointer to `const unsigned char *` is OK as well.

N-Dekker: > Let's document in the conversation here (and maybe in the code too)

Thanks for you suggestion @Leengit I wouldn't mind a line of comment here, but then, I would rather not see such a comment copy-and-pasted throughout the code base, whenever an address is converted to an `unsigned char *`. So do you think we need an extra helper function template, `itk::CastToUnsignedCharPtr(T*) -> unsigned char* `, if such conversion would take place more often ?

Leengit: My reading of http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/n4910.pdf 7.6.1.10, bullet 7 is that skipping the cast to `void *` is required to be equivalent to having used `void *` as an intermediate.

Either way, unfortunately C++ doesn't seem to guarantee that we can do what we want to do here via pointer manipulation.  (I think we should do it anyway.  An alternative is to look at `t & 0xFF`, `t >> 8 & 0xFF`, etc.)

Leengit: I suppose we could side-step the issue with something like
```cpp
  const auto convert = [](auto arg) {
    IntegerType h{ 0 };
    for (size_t i{ 0 }; i < sizeof(arg); ++i)
    {
      h *= UCHAR_MAX + 2U;
      h += (arg & 0xFF);
      arg >>= 8;
    }
    return h;
  };
```

N-Dekker: @dzenanz @Leengit You converted me to use reinterpret_cast  😸  With [this force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/8583dba865362e4c30a169848ed19329f67b8945..f68a49aa157ee3f8324da429f7ad076edfef8037)

dzenanz (approved):

Modules/Core/Common/src/itkMersenneTwisterRandomVariateGenerator.cxx
N-Dekker: Looks like this code by Lawrence Kirby is quite popular  😄 I encountered similar lines of code on various sites, including: 

  http://beagle.gel.ulaval.ca/refmanual/2.2.0/RandomizerMersenne_8cpp-source.html 
  https://groups.google.com/g/comp.lang.c/c/smOwxlZVbeA (post by Ben Pfaff)
  https://github.com/spcl/foMPI-NA/blob/master/MersenneTwister.h#L319-L340

Leengit (approved):
Looks good, thank you

Modules/Core/Common/src/itkMersenneTwisterRandomVariateGenerator.cxx
Leengit: > @Leengit Nice suggestion, `(arg & 0xFF)`, but I'm not entirely sure if it would work for any `clock_t` or `time_t` implementation. You?

If `clock_t` or `time_t` could ever be other than a `signed` or `unsigned` `int` (of any size including `char`), we'd be in trouble.  I can't say that I have surveyed every system, but I think there is a decent probability that this  approach would actually work.  (But your solution is  good, so there is no need to pursue this further.)

Leengit:
FWIW, I am still thinking that this hash function will produce different answers on big endian vs. little endian platforms.  And, I am still thinking that that is perfectly okay.

N-Dekker:
> FWIW, I am still thinking that this hash function will produce different answers on big endian vs. little endian platforms.

@Leengit Thanks for sharing this peculiar finding of yours!

> And, I am still thinking that that is perfectly okay.

I think I'm neutral. I don't really _like_ that the outcome of the hash function appears to depend on "endianness".  But on the other hand, this code by Lawrence Kirby been used on so many places already, for so many years (as I mentioned at https://github.com/InsightSoftwareConsortium/ITK/pull/3361#discussion_r841940681) that I feel reluctant to adjust this logic right now.

dzenanz:
Merge?

PR #3369, opened 2022-04-10
COMP: Fix Slicer vtkITK build on macOS

markasselin:
This commit addresses a Slicer macOS build error reported on Discourse
(see https://discourse.slicer.org/t/slicer-build-on-macos-monterey/21940)
and on GitHub (see comments on issue Slicer/Slicer#5944).

The `_POSIX_SOURCE` preprocessor identifier is defined in several places
including by Python and several ITK ThirdParty modules. It looks like
this definition is causing unintended side effects when building `vtkITK`.

When `itkMacro.h` is #included in `vtkITK`, `_POSIX_SOURCE` causes conditional
inclusion of code using `__assert_fail` and `__ASSERT_FUNCTION`, which are
only defined on Linux and not on macOS. This results in a compile error.

This commit fixes the error by checking for the `__linux` preprocessor
identifier instead of `__POSIX_SOURCE`.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkMacro.h
+++ b/Modules/Core/Common/include/itkMacro.h
@@ -776,7 +776,7 @@ compilers.
 
 #ifndef NDEBUG
 
-#  ifdef _POSIX_SOURCE
+#  ifdef __GLIBC__
 #    define itkAssertInDebugOrThrowInReleaseMacro(msg) __assert_fail(msg, __FILE__, __LINE__, __ASSERT_FUNCTION);
 #  else
 #    define itkAssertInDebugOrThrowInReleaseMacro(msg) itkGenericExceptionMacro(<< msg);


dzenanz (approved):

jcfr:
Since `__assert_fail` is a function provided by glibc [^1], testing for `__GLIBC__` [^2] is likely more specific and correct. Alternatively, a compile test could be implemented.

@thewtex In its current form, could this patch impact `itk-wasm` ?  (I don't expect it does but wanted to confirm)

[^1]: https://sourceware.org/git/?p=glibc.git;a=blob;f=assert/assert.c;h=133a183bc337075e06447e4c8e5cfba180aaf7bf;hb=HEAD#l97

[^2]: https://stackoverflow.com/questions/4266354/how-to-tell-if-glibc-is-used

thewtex:
> @thewtex In its current form, could this patch impact itk-wasm ? (I don't expect it does but wanted to confirm)

itk-wasm uses a clang-based toolchain.

I agree with @jcfr 's suggestion to check for `__GLIBC__` instead.

markasselin:
@jcfr @thewtex thanks for the input, I've updated the commit here and in the Slicer/ITK PR

dzenanz (approved):

jcfr (approved):

PR #3380, opened 2022-04-14
Add `bit_cast`, and use it for `_beginthreadex` calls

N-Dekker:
`bit_cast` helps to avoid undefined behavior when doing a bitwise cast.

### DIFF
--- a/Modules/Core/Common/src/itkPlatformMultiThreaderWindows.cxx
+++ b/Modules/Core/Common/src/itkPlatformMultiThreaderWindows.cxx
@@ -26,6 +26,7 @@
  *
  *=========================================================================*/
 #include "itkPlatformMultiThreader.h"
+#include "itkBitCast.h"
 #include "itkObjectFactory.h"
 #include "itksys/SystemTools.hxx"
 #include <cstdlib>
@@ -71,8 +72,8 @@ PlatformMultiThreader::MultipleMethodExecute()
     m_ThreadInfoArray[threadCount].UserData = m_MultipleData[threadCount];
     m_ThreadInfoArray[threadCount].NumberOfWorkUnits = m_NumberOfWorkUnits;
 
-    processId[threadCount] = (void *)_beginthreadex(
-      nullptr, 0, m_MultipleMethod[threadCount], &m_ThreadInfoArray[threadCount], 0, (unsigned int *)&threadId);
+    processId[threadCount] = bit_cast<HANDLE>(_beginthreadex(
+      nullptr, 0, m_MultipleMethod[threadCount], &m_ThreadInfoArray[threadCount], 0, (unsigned int *)&threadId));
 
     if (processId[threadCount] == nullptr)
     {
@@ -137,7 +138,7 @@ PlatformMultiThreader::SpawnThread(ThreadFunctionType f, void * UserData)
   // Using _beginthreadex on a PC
   //
   m_SpawnedThreadProcessID[id] =
-    (void *)_beginthreadex(nullptr, 0, f, &m_SpawnedThreadInfoArray[id], 0, (unsigned int *)&threadId);
+    bit_cast<HANDLE>(_beginthreadex(nullptr, 0, f, &m_SpawnedThreadInfoArray[id], 0, (unsigned int *)&threadId));
   if (m_SpawnedThreadProcessID[id] == nullptr)
   {
     itkExceptionMacro("Error in thread creation !!!");
@@ -175,9 +176,9 @@ ThreadProcessIdType
 PlatformMultiThreader::SpawnDispatchSingleMethodThread(PlatformMultiThreader::WorkUnitInfo * threadInfo)
 {
   // Using _beginthreadex on a PC
-  DWORD  threadId;
-  HANDLE threadHandle =
-    (HANDLE)_beginthreadex(nullptr, 0, this->SingleMethodProxy, threadInfo, 0, (unsigned int *)&threadId);
+  DWORD threadId;
+  auto  threadHandle =
+    bit_cast<HANDLE>(_beginthreadex(nullptr, 0, this->SingleMethodProxy, threadInfo, 0, (unsigned int *)&threadId));
   if (threadHandle == nullptr)
   {
     itkExceptionMacro("Error in thread creation !!!");
--- a/Modules/Core/Common/include/itkBitCast.h
+++ b/Modules/Core/Common/include/itkBitCast.h
@@ -1,0 +1,60 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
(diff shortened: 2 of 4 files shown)


Modules/Core/Common/src/itkPlatformMultiThreaderWindows.cxx
```diff
@@ -71,8 +72,8 @@ PlatformMultiThreader::MultipleMethodExecute()
     m_ThreadInfoArray[threadCount].UserData = m_MultipleData[threadCount];
     m_ThreadInfoArray[threadCount].NumberOfWorkUnits = m_NumberOfWorkUnits;
 
-    processId[threadCount] = (void *)_beginthreadex(
-      nullptr, 0, m_MultipleMethod[threadCount], &m_ThreadInfoArray[threadCount], 0, (unsigned int *)&threadId);
+    processId[threadCount] = bit_cast<HANDLE>(_beginthreadex(
```
dzenanz, line 75: I am quite skeptical of potential for undefined behavior here.

N-Dekker: > I am quite skeptical of potential for undefined behavior here.

@dzenanz Really? Why? A C-style cast of an integer to a pointer type obviously very much breaks through the C++ type system. Using such a pointer would _by definition_ cause undefined behavior, as far as I know.

dzenanz: `typedef void *HANDLE;` and `uintptr_t _beginthreadex(...)`. Which means that the above casts `uint *` to `void *`. The should be [well defined](https://stackoverflow.com/a/4810460/276168).

N-Dekker: `uintptr_t` is not a pointer type!  😱

dzenanz: That is surprising! Is integral -> pointer cast well defined? But I am afraid it is still somewhat of a moot point. The example which Microsoft's [Visual Studio documentation](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/beginthread-beginthreadex?view=msvc-170) provides does this: `HANDLE hThread = (HANDLE)_beginthreadex(...);`, followed by `WaitForSingleObject( hThread, INFINITE );`. That's what we do here too.

If you want to make changes, I would go the route of changing the definition of `HANDLE processId[ITK_MAX_THREADS];` into `uintptr_t processId[ITK_MAX_THREADS];` and removing the type casting entirely.

N-Dekker: > Is integral -> pointer cast well defined?

No. Using such a pointer afterwards is undefined behavior. (When its pointer value is the result of a C-style cast of an integer).

> The example which Microsoft's [Visual Studio documentation](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/beginthread-beginthreadex?view=msvc-170) provides does this: `HANDLE hThread = (HANDLE)_beginthreadex(...);`, followed by `WaitForSingleObject( hThread, INFINITE );`. 

Interesting, thanks for the link. Indeed, that's undefined behavior too. People from Visual Studio should have known. Of course you might decide to just trust the compiler, it _might_ just work. But it would clearly be preferable to use `bit_cast` in this case.

> If you want to make changes, I would go the route of changing the definition of `HANDLE processId[ITK_MAX_THREADS];` into `uintptr_t processId[ITK_MAX_THREADS];` and removing the type casting entirely.

That would certainly be nice, but them I guess we would also have to change:

https://github.com/InsightSoftwareConsortium/ITK/blob/a5989825db5980e913e03021f26bdf154f269a9d/Modules/Core/Common/include/itkThreadSupport.h#L66

Into:

    using ThreadProcessIdType = uintptr_t;

And make a lot more adjustments... Certainly, we would not be able to pass the result from `_beginthreadex(...)` directly to `WaitForSingleObject`. Unless we'd use ` bit_cast`,  of course  😸 

I still think the PR is currently OK. But the example from Microsoft's Visual Studio documentation should be fixed!

dzenanz (approved):

Modules/Core/Common/src/itkPlatformMultiThreaderWindows.cxx
dzenanz: > pass the result from _beginthreadex(...) directly to WaitForSingleObject

I didn't think of this issue. Approved!

N-Dekker:
@dzenanz I'm sorry, my claim that the `(HANDLE)` cast would cause undefined behavior was too strong. (Meaning I was wrong  😸.) As was just explained to me at the [#include <C++> discussion channel](https://www.includecpp.org/discord/), referring to http://eel.is/c++draft/expr.reinterpret.cast#4

> A pointer can be explicitly converted to any integral type large enough to hold all values of its type. The mapping function is implementation-defined.

In this particular case, the `(HANDLE)` cast would behave just like a `reinterpret_cast<HANDLE>`.

I still think that in this particular case,  `bit_cast<HANDLE>` is preferable to  `reinterpret_cast<HANDLE>`, because the behavior of C++20 `bit_cast` is defined by the C++ Standard, rather than just "implementation-defined".

In other cases, bit_cast does actually prevent undefined behavior, for example (https://godbolt.org/z/xearxvxM6):

    float f = 0.1f;
    int i1 = reinterpret_cast<int & >(f); // Undefined behavior!
    int i2 = bit_cast<int>(f); // OK

N-Dekker:
For the record, this pull request + discussion appears related to issue https://github.com/isocpp/CppCoreGuidelines/issues/1517 "Favor bit_cast over reinterpret_cast".

N-Dekker:
@Leengit Could you possibly have a look at this issue as well? It seems that I was too eager to start using bit_cast, rather than reinterpret_cast! You may find the discussion at issue https://github.com/isocpp/CppCoreGuidelines/issues/1517 "Favor bit_cast over reinterpret_cast" interesting.  😄 

Jason Turner has two short video's on this topic, however, they don't seem to apply specifically to the `(HANDLE)` cast that I was addressing with this pull request:
- [C++ Weekly - Ep 185 - Stop Using reinterpret_cast!](https://www.youtube.com/watch?v=L06nbZXD2D0)
- [C++ Weekly - Ep 265 - C++20's std::bit_cast](https://www.youtube.com/watch?v=crErQJMwz1g)

So I'm now considering to replace the `itk::bit_cast<HANDLE>` calls from this pull request by reinterpret_cast's. What do you think?

Leengit:
This is a new one for me and I haven't followed all your links, so I may end up wishing to eat my words in the next few hours or minutes but ... I am thinking that I would continue to use `reinterpret_cast` for objects of different sizes and for arrays and pointers.  I would use `bit_cast` only if the from and to objects are neither pointers nor arrays and are of the same size.
```cpp
  double * pDouble = nullptr;
  // this next line is good enough because it is to char* (assuming sizeof(char)==1) or void*
  char * pChar = reinterpret_cast<char *>(pDouble);

  float        f = 1.0f;
  unsigned int i1 = std::bit_cast<unsigned int>(f);        // yes
  unsigned int i2 = *reinterpret_cast<unsigned int *>(&f); // now deprecated
```

And if they are not trivially copyable then I don't know.

PR #3381, opened 2022-04-15
STYLE: Improve style in tests

jhlegarreta:
- COMP: Remove duplicate include file 
- STYLE: Use standard library exit codes in tests
- STYLE: Check test input arguments at the beginning
- STYLE: Remove uninformative/unnecessary print messages in tests
- STYLE: Prefer using `ITK_TRY_EXPECT_NO_EXCEPTION` macro in tests
- STYLE: Use the `itkNameOfTestExecutableMacro` macro in tests
- STYLE: Prefer removing unused input arguments in tests
- STYLE: Conform to ITK style in constant and variable naming in tests
- STYLE: Improve variable value testing information
- BUG: Fix variable type

## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [X] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)

PR #3382, opened 2022-04-16

jhlegarreta:
- DOC: Fix typo in method documentation
- ENH: Fix typo in direction value exception macro message
- ENH: Prefer using the `itkTypeMacroNoParent` macro for RTTI
- STYLE: Use the superclass name in `itkTypeMacro`
- BUG: Fix Superclass type alias and name in RTTI macro
- BUG: Fix Superclass name in RTTI macro
- ENH: Add `itkBooleanMacro` to `IFrameSafe` boolean ivar
- ENH: Add getter macro for boolean ivar
- ENH: Increase coverage for miscellaneous classes

## PR Checklist

### DIFF
--- a/Modules/Filtering/FastMarching/include/itkFastMarchingImageToNodePairContainerAdaptor.h
+++ b/Modules/Filtering/FastMarching/include/itkFastMarchingImageToNodePairContainerAdaptor.h
@@ -104,6 +104,7 @@ public:
   SetForbiddenImage(const ImageType * iImage);
 
   itkSetMacro(IsForbiddenImageBinaryMask, bool);
+  itkGetConstMacro(IsForbiddenImageBinaryMask, bool);
   itkBooleanMacro(IsForbiddenImageBinaryMask);
 
   /** \brief Get resulting Alive Points container*/
--- a/Modules/Filtering/FastMarching/include/itkFastMarchingImageToNodePairContainerAdaptor.h
+++ b/Modules/Filtering/FastMarching/include/itkFastMarchingImageToNodePairContainerAdaptor.h
@@ -98,7 +98,7 @@ public:
     is a binary mask; thus null values are considered as
     FastMarchingTraitsBase::Forbidden points
 
-    \li else (m_IsForbiddenImageBinaryMask is \c fasle) non null values
+    \li else (m_IsForbiddenImageBinaryMask is \c false) non null values
     represents FastMarchingTraitsBase::Forbidden points*/
   void
   SetForbiddenImage(const ImageType * iImage);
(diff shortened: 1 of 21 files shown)


jhlegarreta:
PR #3381 is expected to be merged prior to this.

Modules/Filtering/FastMarching/include/itkFastMarchingImageToNodePairContainerAdaptor.h
```diff
@@ -53,7 +53,7 @@ class ITK_TEMPLATE_EXPORT FastMarchingImageToNodePairContainerAdaptor : public O
   ITK_DISALLOW_COPY_AND_MOVE(FastMarchingImageToNodePairContainerAdaptor);
 
   using Self = FastMarchingImageToNodePairContainerAdaptor;
-  using Superclass = LightObject;
+  using Superclass = Object;
```
jhlegarreta, line 56: I guess this is the right inheritance. Let me know otherwise (i.e. if `LightObject` should be the parent class instead in https://github.com/InsightSoftwareConsortium/ITK/pull/3382/files#diff-15a472539026d9182f07806b46297e6a0fe90ab0012067872bad81ec3c62a638R50).

Modules/Filtering/FastMarching/include/itkFastMarchingImageToNodePairContainerAdaptor.h
```diff
@@ -98,12 +98,13 @@ class ITK_TEMPLATE_EXPORT FastMarchingImageToNodePairContainerAdaptor : public O
...
     FastMarchingTraitsBase::Forbidden points
 
-    \li else (m_IsForbiddenImageBinaryMask is \c fasle) non null values
+    \li else (m_IsForbiddenImageBinaryMask is \c false) non null values
     represents FastMarchingTraitsBase::Forbidden points*/
   void
   SetForbiddenImage(const ImageType * iImage);
 
   itkSetMacro(IsForbiddenImageBinaryMask, bool);
+  itkGetConstMacro(IsForbiddenImageBinaryMask, bool);
```
jhlegarreta, line 107: There is some inconsistency across the code base concerning the boolean getter macros: `itkGetConstReferenceMacro`, `itkGetConstMacro` and `itkGetMacro` all three are used. I guess we prefer one of the first two, but I'm not sure which of the two should be consistently used. Thoughts are welcome.

dzenanz: I think that const is preferred for simple types (int, bool etc) and const reference for others (std::string, itk::image etc). If modification is required, then `itkGetModifiableObjectMacro` should be used. And `itkGetMacro` should be replaced by one of the other 3, as appropriate.

dzenanz (approved):

Modules/Filtering/FastMarching/include/itkFastMarchingImageToNodePairContainerAdaptor.h
jhlegarreta: OK. I'll try to put this to the documentation of the macros or the ITK SW Guide.

As always, thanks for the review @dzenanz :100:.

Leengit: > I think that const is preferred for simple types (int, bool etc) and const reference for others (std::string, itk::image etc).

With a little C++ acrobatics involving templates and/or SFINAE, we might be able to create a macro that figures out at compile time whether to return by value or by reference.

PR #3383, opened 2022-04-16
BUG: Fix `itkXMLFileOutputWindowTest` regression

jhlegarreta:
Fix `itkXMLFileOutputWindowTest` regression: delete the contents of the
target file prior to writing new lines when no filename is given.

When no filename is provided, the `itk::XMLFileOutputWindow` class uses
a hard-coded filename to flush the buffer contents. If the CI builds do
not clean the target folder, as the append flag is set to `On`, lines
will be appended to the existing file, and the check on the written
lines will fail.

Removing the contents of the file is preferred over computing the
difference of the existing line count with respect to the new line
count, since the file will otherwise grow without bounds.

Fixes:
```
Error in numLinesRead == numLinesExpected
  In ITK/Modules/Core/Common/test/itkXMLFileOutputWindowTest.cxx, line 86
  lh: 47
  rh: 7
Expression is not equal
```

raised for example in:
https://open.cdash.org/test/669234154

## PR Checklist

### DIFF
--- a/Modules/Core/Common/test/itkXMLFileOutputWindowTest.cxx
+++ b/Modules/Core/Common/test/itkXMLFileOutputWindowTest.cxx
@@ -20,6 +20,16 @@
 #include "itkXMLFileOutputWindow.h"
 #include "itkTestingMacros.h"
 #include <fstream>
+#include <cstdio>
+#ifdef _WIN32
+#  include <direct.h>
+#  define cwd _getcwd
+#  define OS_SEP "\\"
+#else
+#  include "unistd.h"
+#  define cwd getcwd
+#  define OS_SEP "/"
+#endif
 
 
 int
@@ -46,6 +56,33 @@ itkXMLFileOutputWindowTest(int argc, char * argv[])
 
   logger->SetInstance(logger);
 
+  // If not input filename is provided, remove the contents in the existing file to avoid counting existing lines when
+  // contents are appended
+  if (argc == 1)
+  {
+    // In order to initialize the filename, some text needs to be written first
+    const char * regularText = "text";
+    logger->DisplayText(regularText);
+
+    // Get the filename
+    const char *      fileBaseName = logger->GetFileName();
+    const std::size_t size = 4096;
+    char              tmp[size];
+    char *            status = cwd(tmp, size);
+    if (!status)
+    {
+      std::cerr << "Test failed!" << std::endl;
+      std::cerr << "Error getting the current directory. Cannot delete the contents of the target file: "
+                << fileBaseName << std::endl;
+      return EXIT_FAILURE;
+    }
+    std::string fileName = tmp + std::string(OS_SEP) + std::string(fileBaseName);
+
+    // Delete the contents
+    std::ofstream ofs;
+    ofs.open(fileName, std::ofstream::out | std::ofstream::trunc);
+    ofs.close();
+  }
 
   // Check special cases
   const char * text = nullptr;


Modules/Core/Common/test/itkXMLFileOutputWindowTest.cxx
```diff
@@ -20,6 +20,16 @@
...
+#include <cstdio>
+#ifdef _WIN32
+#  include <direct.h>
+#  define cwd _getcwd
+#  define OS_SEP "\\"
+#else
+#  include "unistd.h"
+#  define cwd getcwd
+#  define OS_SEP "/"
+#endif
```
jhlegarreta, line 32: There may be some more elegant solution to get the working directory and the OS separator.

Modules/Core/Common/test/itkXMLFileOutputWindowTest.cxx
```diff
@@ -46,6 +56,33 @@ itkXMLFileOutputWindowTest(int argc, char * argv[])
...
+    // In order to initialize the filename, some text needs to be written first
+    const char * regularText = "text";
+    logger->DisplayText(regularText);
+
+    // Get the filename
+    const char *      fileBaseName = logger->GetFileName();
+    const std::size_t size = 4096;
+    char              tmp[size];
+    char *            status = cwd(tmp, size);
+    if (!status)
```
jhlegarreta, line 72: Also, although unlikely, if `getcwd` requires a larger buffer size, the error should be dealt with. Please feel free to improve this. Thanks.

Modules/Core/Common/test/itkXMLFileOutputWindowTest.cxx
```diff
@@ -46,6 +56,33 @@ itkXMLFileOutputWindowTest(int argc, char * argv[])
...
+      std::cerr << "Error getting the current directory. Cannot delete the contents of the target file: "
+                << fileBaseName << std::endl;
+      return EXIT_FAILURE;
+    }
+    std::string fileName = tmp + std::string(OS_SEP) + std::string(fileBaseName);
+
+    // Delete the contents
+    std::ofstream ofs;
+    ofs.open(fileName, std::ofstream::out | std::ofstream::trunc);
+    ofs.close();
```
jhlegarreta, line 84: I could not remove the file using `std::remove`. I ignore why. Please feel free to propose changes if you happen to know how deleting a file is effectively done.

Modules/Core/Common/test/itkXMLFileOutputWindowTest.cxx
```diff
@@ -46,6 +56,33 @@ itkXMLFileOutputWindowTest(int argc, char * argv[])
 
   logger->SetInstance(logger);
 
+  // If not input filename is provided, remove the contents in the existing file to avoid counting existing lines when
+  // contents are appended
+  if (argc == 1)
+  {
+    // In order to initialize the filename, some text needs to be written first
```
jhlegarreta, line 63: Does not look like an ideal requirement, but that is beyond the scope of this PR. Feel free to propose a PR for the class to be released from this constraint.

Modules/Core/Common/test/itkXMLFileOutputWindowTest.cxx
dzenanz: [std::remove](https://en.cppreference.com/w/cpp/algorithm/remove) deals with iterators, not files.

dzenanz (approved):
Looks fine to me.

dzenanz:
Merging now to reduce noise on the nightly dashboard.

Modules/Core/Common/test/itkXMLFileOutputWindowTest.cxx
jhlegarreta: :roll_eyes: I'd dare to say that it deals with files when the appropriate header is included (which was the case when I tried): https://en.cppreference.com/w/cpp/io/c/remove

PR #3384, opened 2022-04-18
COMP: Fix overly narrow cast

Leengit:
Fix so that the build no longer (correctly) complains that an `unsigned int` is cast to `double` in a context where narrowing casts are prohibited.

### DIFF
--- a/Modules/Registration/FEM/include/itkMIRegistrationFunction.hxx
+++ b/Modules/Registration/FEM/include/itkMIRegistrationFunction.hxx
@@ -376,8 +376,8 @@ MIRegistrationFunction<TFixedImage, TMovingImage, TDisplacementField>::ComputeUp
   double msigma = 0.0;
   double jointsigma = 0.0;
 
-  const double numsamplesB{ fixedSamplesB.size() };
-  const double numsamplesA{ fixedSamplesA.size() };
+  const double numsamplesB{ static_cast<double>(fixedSamplesB.size()) };
+  const double numsamplesA{ static_cast<double>(fixedSamplesA.size()) };
   double       nsamp = numsamplesB;
   //  if (maxf == minf && maxm == minm) return update;
   //    else std::cout << " b samps " << fixedSamplesB.size()


dzenanz (approved):

Modules/Registration/FEM/include/itkMIRegistrationFunction.hxx
```diff
@@ -376,8 +376,8 @@ MIRegistrationFunction<TFixedImage, TMovingImage, TDisplacementField>::ComputeUp
   double msigma = 0.0;
   double jointsigma = 0.0;
 
-  const double numsamplesB{ fixedSamplesB.size() };
-  const double numsamplesA{ fixedSamplesA.size() };
+  const double numsamplesB{ static_cast<double>(fixedSamplesB.size()) };
+  const double numsamplesA{ static_cast<double>(fixedSamplesA.size()) };
```
N-Dekker, line 380: Thanks @Leengit Just a note here, as it is merged already. These are typical (pretty much uncontroversial) use cases for the `auto` keyword, as in:

    const auto numsamplesA = static_cast<double>(fixedSamplesA.size());
    const auto numsamplesB = static_cast<double>(fixedSamplesB.size());

See also https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-auto.html#cast-expressions

Leengit: I am all for `auto` when some unwieldy type specifier would otherwise be used twice in the same statement, and for situations like iterators where the functionality is more important than the actual type.  If the ITK style guide is against me then I should also adjust to using `auto` in this situation where the type in question is short and basic.

I also prefer `{...}` rather than `=` in cases where I know there will be no narrowing.  That would change the suggestion to
```cpp
const auto numsamplesA{ static_cast<double>(fixedSamplesA.size()) };
const auto numsamplesB{ static_cast<double>(fixedSamplesB.size()) };
```

N-Dekker: > If the ITK style guide is against me...

@Leengit It certainly isn't against you personally, but...

> The 'auto' keyword should be used to specify a type in the following cases:
> - The type is duplicated on the left side of an initialization when it is mandated on the
	right side, e.g. when there is an explicit cast or initializing with 'new' or ITK's '::New()`.

https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/blob/dd10e331677cdfbb1c7768925db910b2a49ade20/SoftwareGuide/Latex/Appendices/CodingStyleGuide.tex#L1543-L1547

Issue #3368, opened 2022-04-07
PERF: Leverage speedup over requested region subset in `FFTConvolutionImageFilter`


tbirdso:
### Description

A performance discrepancy is observed when blurring with `FFTDiscreteGaussianImageFilter` in the frequency domain via GPU-based FFTs vs `DiscreteGaussianImageFilter` in physical space via convolution with a separable kernel on CPU.

### Impact analysis

In subplots below "Vk" represents the GPU implementation and "Vnl" represents the CPU implementation.

In the top row 3D cube images of variable "LargestPossibleRegionSize" are blurred. Filter performance follows a natural, smooth trend as expected.

In the bottom row 3D cube images of fixed "LargestPossibleRegion" size with variable "RequestedOutputRegion" sizes are blurred. GPU FFT convolution plateaus while CPU FFT shows a sharp performance increase over comparable input size in the top row. The result is that GPU FFT convolution offers minimal to no performance boost over CPU implementation for most image region subsets.

![blur-benchmark-plot](https://user-images.githubusercontent.com/40648863/162277293-bf22be7a-6325-43c0-8ffd-e60d61f38b88.png)

It is hypothesized that the performance difference is based on boundary handling in the respective filters. [FFTConvolutionImageFilter::GenerateInputRequestedRegion](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/Convolution/include/itkFFTConvolutionImageFilter.hxx#L50) always requests the largest possible input region, so every call processes the same region size irrespective of the requested output. This leads to the plateau observed in the lower row. By contrast, `DiscreteGaussianImageFilter` propagates the requested output region so that a subset of the input is actually processed. When the requested region is centered inside the input image, it is possible for boundary pixels to be obtained directly from the surrounding area in the input image rather than generated dynamically from the boundary condition. There is some evidence that the difference in boundary processing is a core issue here, setting the requested region index at (0,0,0) results in lessened CPU blurring performance.

### Expected behavior

GPU and CPU blurring performance should follow a similar trend when `OutputRequestedRegion` is reduced from `LargestPossibleRegion`.

### Actual behavior

CPU blurring performance exhibits significant speedup when processing image subsets while GPU blurring performance is independent of the requested region.

### Versions

ITK v5.3rc04

### Environment

Windows, CMake 3.22

### Additional Information

Will investigate and benchmark adjustments to how FFTConvolutionImageFilter sets its requested input region size.

[...]

PR #3387, opened 2022-04-20
ENH: Consider requested output region in FFT convolution

tbirdso:
Prior to this change, `itk::FFTConvolutionImageFilter` and
`itk::ConvolutionImageFilter` had the major difference where the former
would always run on the largest possible input image region whereas the
latter would run on a padded requested region. As a result spatial
convolution was much more performant when run on a small region of a
large image.

This change revises how `itk::FFTConvolutionImageFilter` processes input
and output image regions in order to perform convolution and discards
the assumption that the output requested region must always be the
largest possible region. The input image subregion is padded for the
kernel operator and then clipped of remaining image data so that forward
FFT is performed on a region exactly matching that which would be
processed under spatial convolution. The kernel image is likewise padded
to the match the input. The result of FFT multiplication and subsequent
inversion is then cropped to remove artifacts from padding, leaving only
the requested input size.

This approach has the advantage of scaling with the size of the
requested output region. FFT convolution continues to be disadvantaged
in comparison to spatial convolution where an output region is
iteratively expanded or shrunk, as under FFT convolution entirely new
complex images must be generated for any change in the output region.

Adds tests demonstrating functionality and equivalence to spatial
convolution.

This work is motivated by FFT acceleration with VkFFT applied to large image processing. The fundamental approach to subregion processing is demonstrated in the ITKVkFFTBackend [FFTConvolutionOverRegionOfInterest](https://github.com/InsightSoftwareConsortium/ITKVkFFTBackend/blob/master/example/FFTConvolutionOverRegionOfInterest.ipynb) example. Speedup is demonstrated below in the context of Vnl (default) FFT and VkFFT (GPU accelerated) backends. Benchmarks were collected via a script based off of the [GaussianBlurBenchmark](https://github.com/InsightSoftwareConsortium/ITKVkFFTBackend/blob/master/example/GaussianBlurBenchmark.ipynb) example. FFT performance over subregions (bottom row) is greatly improved, though more optimization is required to outperform the default spatial convolution multithreaded implementation over subregions.

Status quo: Vnl FFT, not responsive to requested region

conv-benchmark-pre-no-vkfft

Updated: Vnl FFT, responsive to requested region

conv-benchmark-post-no-vkfft

Status quo: Vk FFT, not responsive to requested region

conv-benchmark-pre-w-vkfft

Updated: Vk FFT, responsive to requested region

conv-benchmark-post-w-vkfft

## PR Checklist


tbirdso:
Mostly addresses #3368. More profiling work may be necessary to outperform multithreaded spatial convolution over image subregions on CPU. May be able to make up the difference in input image preparation optimizations, see TODO comments inline.

Modules/Filtering/Convolution/include/itkFFTConvolutionImageFilter.hxx
```diff
@@ -126,47 +147,160 @@ FFTConvolutionImageFilter<TInputImage, TKernelImage, TOutputImage, TInternalPrec
...
+  for (unsigned int dim = 0; dim < ImageDimension; ++dim)
+  {
+    signed int lower = static_cast<signed int>(kernelRadius[dim]) - (inputRequestedIndex[dim] - inputLargestIndex[dim]);
+    signed int upper =
+      static_cast<signed int>(kernelRadius[dim]) -
+      ((inputLargestIndex[dim] + inputLargestSize[dim]) - (inputRequestedIndex[dim] + inputRequestedSize[dim]));
+    lowerPad[dim] = (lower > 0 ? lower : 0);
+    upperPad[dim] = (upper > 0 ? upper : 0);
+
+    needsKernelPadding = (lower > 0 || upper > 0);
```
dzenanz, line 182: padding is overwritten here - only the last dimension matters
```suggestion
    needsKernelPadding = needsKernelPadding || lower > 0 || upper > 0;
```
or `needsKernelPadding ||= lower > 0 || upper > 0;` if it compiles.

Modules/Filtering/Convolution/include/itkFFTConvolutionImageFilter.hxx
```diff
@@ -126,47 +147,160 @@ FFTConvolutionImageFilter<TInputImage, TKernelImage, TOutputImage, TInternalPrec
...
+  else
+  {
+    kernelPaddedInput = input;
+  }
+
+  // Crop to region of interest to minimize FFT requested region size.
+  // This blocks forward FFT from requesting the entire input image
+  // when only a subregion (output requested region + kernel radius padding)
+  // is required for convolution to succeed.
+  // TODO: Improve readability by keeping the output index as-is rather than resetting to 0, if the filter allows it
```
dzenanz, line 209: Isn't this being done on lines 235-250?

Modules/Filtering/Convolution/test/itkFFTConvolutionImageFilterSubregionTest.cxx
```diff
@@ -0,0 +1,138 @@
...
+    kernelReader->SetFileName(argv[9]);
+    ITK_TRY_EXPECT_NO_EXCEPTION(kernelReader->Update());
+    kernelImage = kernelReader->GetOutput();
+  }
+  else
+  {
+    kernelImage = FFTGenerateGaussianKernel();
+  }
+
+  using ConvolutionFilterType = itk::FFTConvolutionImageFilter<ImageType>;
```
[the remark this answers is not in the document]
tbirdso: Great point, I will do this.

Modules/Filtering/Convolution/include/itkFFTConvolutionImageFilter.hxx
tbirdso: Good catch, I also just found this in testing. Fixed in latest commit.

Modules/Filtering/Convolution/include/itkFFTConvolutionImageFilter.hxx
tbirdso: Yes, I haven't yet looked into whether there is a way to preserve the image index with `itk::RegionOfInterestImageFilter` without adding an additional `itk::ChangeInformationImageFilter` to the pipeline. The performance impact is negligible either way, would be nice if we can do this by setting a parameter though.

Modules/Filtering/Convolution/include/itkFFTConvolutionImageFilter.h
```diff
@@ -181,18 +185,19 @@ class ITK_TEMPLATE_EXPORT FFTConvolutionImageFilter
...
-  InputSizeType
-  GetPadSize() const;
+  /** Get padding around the region of interest that results from FFT
+   *  factoring requirements. FFT typically requires that image side lengths
+   *  are factorable only by a fixed set of prime numbers (often 2, 3, and 5).
+   *  After the input image is padded for the kernel width and cropped to the
+   *  region of interest the result is then padded for FFT execution. This value
+   *  is reused for kernel padding and output cropping. */
+  InternalSizeType
+  GetFFTPadSize() const;
```
jhlegarreta, line 200: Sorry to chime in. Thanks for the effort @tbirdso :100:. Now a petition: can the two getter methods above be exercised in the test, please?

Sames applies to any other method added (if any).

Thanks.

tbirdso: Thank you for taking a look @jhlegarreta ! The two methods in question are used only internally, they are `protected` and not part of the public interface. I do not think that we typically test for these?

jhlegarreta: I had quickly read the class and thought they were `public`. There are many such `protected` methods that are not covered as you mention. I think that is unfortunate. IMO, we should exercise them. The only way I can think of is creating a subclass in the tests for that purpose. If you happen to have a better solution, I'd like to hear about it. Thanks.

dzenanz: Protected methods should be exercised indirectly, via the public methods which use them. I don't think we should specifically test them. The coverage can come naturally, via general testing of a class.

tbirdso:
As of writing there is exactly one [CDash test failure](https://open.cdash.org/test/674348727) remaining where a deconvolution test output differs slightly from the original baseline. I am summarizing my findings from tracing through both previous and updated filter behaviors here to 1) provide evidence for changing the baseline image to match updated behavior and 2) highlight what I think is an interesting case study in how small theoretical errors can propagate.

The issue lies in a bug in the previous implementation of `FFTConvolutionImageFilter` that is fixed in this update. Previously the filter always executed on the largest possible input region padded for 1) the kernel and 2) FFT factoring. The image pad size is calculated at [https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/Convolution/include/itkFFTConvolutionImageFilter.hxx#L390](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/Convolution/include/itkFFTConvolutionImageFilter.hxx#L390):
```
typename InputImageType::ConstPointer  inputImage = this->GetInput();
  InputSizeType                          inputSize = inputImage->GetLargestPossibleRegion().GetSize();
  typename KernelImageType::ConstPointer kernelImage = this->GetKernelImage();
  KernelSizeType                         kernelSize = kernelImage->GetLargestPossibleRegion().GetSize();

  InputSizeType padSize;
  for (unsigned int i = 0; i < ImageDimension; ++i)
  {
    padSize[i] = inputSize[i] + kernelSize[i];
    if (m_SizeGreatestPrimeFactor > 1)
    {
      while (Math::GreatestPrimeFactor(padSize[i]) > m_SizeGreatestPrimeFactor)
      {
        padSize[i]++;
      }
    }
  }
```
The issue lies in this line:
> padSize[i] = inputSize[i] + kernelSize[i];

When we perform spatial convolution we need to be able to sample by the kernel radius outside of the bounds of the requested region, so the padded image size is `2 * kernelRadius[i]`. However, if `kernelSize[i]` is odd then `kernelSize[i] = 2 * kernelRadius[i] + 1`, which means we are initially padding by one extra pixel. In many cases this is a negligible error, _however_ if the incorrect padding size does not meet FFT factoring conditions then significant extra padding can be subsequently added:
```
while (Math::GreatestPrimeFactor(padSize[i]) > m_SizeGreatestPrimeFactor)
      {
        padSize[i]++;
      }
```

[`itkInverseDeconvolutionImageFilterGaussianKernelTest`](https://open.cdash.org/testSummary.php?project=2&name=itkInverseDeconvolutionImageFilterGaussianKernelTest&date=2022-04-20) happens to have the correct conditions for this bug to appear. An input of size [250, 250] is padded for a kernel of size [51, 51]. In previous behavior this leads to a padding of [301, 301], which is not factorable by 2, 3, and 5, so it is expanded to an input size of [320, 320] for FFT to be carried out. Meanwhile, updated behavior correctly pads by the kernel radius to [300, 300], which is factorable by 2, 3, 5 and needs no additional padding for FFT. The difference in pipelines cascades from there, with kernels being padded to the respective input sizes of [320, 320] or [300, 300], followed by complex division for deconvolution, inverse FFT, and cropping to the requested output region.

All in all, this was an easy error to make and I thought it was interesting to trace through and see how an extra +1 in the wrong place can cause cascading failure. I will upload a corrected baseline image for this test.

Modules/Filtering/Convolution/include/itkFFTConvolutionImageFilter.h
tbirdso: While I agree with the sentiment of wanting to explicitly test protected methods, I'm uncertain that we have the resources to make this a common practice. Note that existing protected methods are shown to be covered via natural exercises in the existing [`itkFFTConvolutionImageFilter.h` coverage report](https://open.cdash.org/viewCoverageFile.php?buildid=7868666&fileid=49410952), which, while imperfect, has previously been noted as our standard for test coverage. Considering that we are not changing the `FFTConvolutionImageFilter` public interface I think it is okay to leave as-is for now.

tbirdso:
Fixed a build warning and squashed commits.

Leengit (commented):
I gave some in-line suggestions.

I am still pondering the odd vs. even kernel dimension size; I'll get back to you on that.

Modules/Filtering/Convolution/include/itkFFTConvolutionImageFilter.hxx
```diff
@@ -126,47 +147,154 @@ FFTConvolutionImageFilter<TInputImage, TKernelImage, TOutputImage, TInternalPrec
...
-  inputPadder->SetPadLowerBound(inputLowerBound);
+  PadSizeType    lowerPad, upperPad;
+  bool           needsKernelPadding = false;
+  KernelSizeType kernelRadius = this->GetKernelRadius();
+  for (unsigned int dim = 0; dim < ImageDimension; ++dim)
+  {
+    signed int lower = static_cast<signed int>(kernelRadius[dim]) - (inputRequestedIndex[dim] - inputLargestIndex[dim]);
+    signed int upper =
+      static_cast<signed int>(kernelRadius[dim]) -
+      ((inputLargestIndex[dim] + inputLargestSize[dim]) - (inputRequestedIndex[dim] + inputRequestedSize[dim]));
```
Leengit, line 178: On these lines and just above, my compiler says that an `int` plus an `unsigned int` is an `unsigned int`.  It looks like this code is instead depending upon that sum being an `int`.  However, the final assignment maybe reinterprets what has sometimes accidentally become a very large `unsigned int` back to what it should have been, a negative `int`; this should work but seems unsatisfactory.  Instead maybe declare the `*Size` variables and the `kernelRadius` variable as `InputIndexType`, by casting the right-hand side of their initializations if necessary.

Modules/Filtering/Convolution/include/itkFFTConvolutionImageFilter.hxx
```diff
@@ -43,11 +44,29 @@ template <typename TInputImage, typename TKernelImage, typename TOutputImage, ty
...
+
+    // Pad the output request region by the kernel radius.
+    // Note that FFT padding will always be generated by a boundary condition
+    // as an implementation detail, while pixels for kernel radius padding may be taken
+    // from the original image if they lies inside the image bounds.
+    inputRegion.PadByRadius(this->GetKernelRadius());
+
+    // Crop the output requested region to fit within the largest
+    // possible region.
+    typename InputImageType::Pointer inputPtr = const_cast<InputImageType *>(this->GetInput());
```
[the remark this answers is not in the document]
tbirdso: I will revisit this method. Note that the `GenerateInputRequestedRegion` implementation was mostly copied directly from [`itk::ConvolutionImageFilter`](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/Convolution/include/itkConvolutionImageFilter.hxx#L241).

Modules/Filtering/Convolution/include/itkFFTConvolutionImageFilter.hxx
tbirdso: Good catch!

Modules/Filtering/Convolution/include/itkFFTConvolutionImageFilter.hxx
```diff
@@ -126,47 +147,154 @@ FFTConvolutionImageFilter<TInputImage, TKernelImage, TOutputImage, TInternalPrec
...
-  auto inputPadder = InputPadFilterType::New();
-  inputPadder->SetBoundaryCondition(this->GetBoundaryCondition());
+  using PadSizeType = typename InputPadFilterType::SizeType;
 
-  InputSizeType inputLowerBound = this->GetPadLowerBound();
-  inputPadder->SetPadLowerBound(inputLowerBound);
+  PadSizeType    lowerPad, upperPad;
+  bool           needsKernelPadding = false;
+  KernelSizeType kernelRadius = this->GetKernelRadius();
+  for (unsigned int dim = 0; dim < ImageDimension; ++dim)
```
Leengit, line 173: What is the intent here for even and odd kernel sizes?  Is it that if the kernel size is `2r+1` then we are growing a range `[i,i+s)` to `[i-r,i+s+r)`, which means we need to pad if `i-r < inputLargestIndex` or if `i+s+r > inputLargestIndex+inputLargestSize`?

What happens if the kernel size is `2r`?  Do we grow the region the same, as if the kernel size were actually `2r+1`?  Or, for example, do we instead grow the upper limit by `r-1`?

tbirdso: We are padding by `r` on each side, meaning that the resulting region has size `size[dim] + 2 * r[dim]` in each direction regardless of whether the kernel is of size `2r` or `2r+1`. So, yes, we treat even-sized kernel images as if the kernel size was actually `2r+1` as you've said.

dzenanz (approved):

Modules/Filtering/Convolution/include/itkFFTConvolutionImageFilter.hxx
```diff
@@ -43,11 +44,29 @@ template <typename TInputImage, typename TKernelImage, typename TOutputImage, ty
...
+    // Note that FFT padding will always be generated by a boundary condition
+    // as an implementation detail, while pixels for kernel radius padding may be taken
+    // from the original image if they lies inside the image bounds.
+    inputRegion.PadByRadius(this->GetKernelRadius());
+
+    // Crop the output requested region to fit within the largest
+    // possible region.
+    typename InputImageType::Pointer inputPtr = const_cast<InputImageType *>(this->GetInput());
+    bool                             cropped = inputRegion.Crop(inputPtr->GetLargestPossibleRegion());
+    if (!cropped)
```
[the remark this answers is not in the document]
dzenanz: I also spent some minutes figuring out what goes on here. Variable `cropped` should be renamed into `wasOutside`.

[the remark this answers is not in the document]
tbirdso: I agree that this should be renamed. To me `wasOutside` implies the same as `cropped`, which is that cropping ran and removed some pixels outside of some common pixel region. I would favor something like `isRegionValid` to reflect the failure condition where there is no overlap between the requested and largest possible regions, leading to us failing out in the next line.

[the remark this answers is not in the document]
dzenanz: Maybe `wasCompletelyOutside`? `isRegionValid` says nothing about why.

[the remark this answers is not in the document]
tbirdso: `wasCompletelyOutside` is the inverse of the return case here where 0 == no overlap, 1 == some overlap and subsequent crop. How about `wasPartiallyInside`?

We only use this variable to conditionally throw an exception in the next line explicitly stating the reason that the region was invalid, so in this specific case I'm not sure how much it matters that the variable gives deeper insight as a redundancy for the error string.

dzenanz (approved):

Modules/Filtering/Convolution/include/itkFFTConvolutionImageFilter.hxx
tbirdso: After review I found that there is not a cast defined for casting from `InputSizeType` to `InputIndexType` as one is an `itk::Size` and the other is an `itk::Offset`. Attempting to cast directly returns a compiler failure. Instead I added two extra lines above to implicitly cast addition operations to the `int` type, which I think also helps with readability.

Leengit: One cast to `int` but not to `itk::Offset` (which is implemented at `int`)?

tbirdso: @Leengit I'm not sure I understand the question?

tbirdso:
Squashed changes

thewtex:
\azp run ITK.macOS.Python

thewtex (approved):
:1st_place_medal: :rocket:

blowekamp:
Looks like a very nice addition and improvement!

Looking through the test I don't see any tests which run a pipeline. They seem to mostly just call update on each filter.

Please consider adding a test which something like: ImageFileReader->PipelineMonitorImageFilter->FFTConvolutionImageFilter->StreamingImageFilter The [PipelineMonitorImageFilter ](https://itk.org/Doxygen/html/classitk_1_1PipelineMonitorImageFilter.html) save the requested regions and other pipeline communications to help verify correct execution of a pipeline.

tbirdso:
@blowekamp I've added an additional streaming test. Could you please verify that this covers the cases you had in mind?

Modules/Filtering/Convolution/test/itkConvolutionImageFilterStreamingTest.cxx
```diff
@@ -0,0 +1,173 @@
...
+  ITK_TRY_EXPECT_NO_EXCEPTION(streamer->Update());
+
+  // Verify the pipeline executed as expected
+  const unsigned int numUpdates = 1;
+  // Expect requested region propagates twice due to ConvolutionImageFilter having
+  // two inputs (Primary and KernelImage)
+  const unsigned int numOutputRequestedRegions = 2;
+  // Verify ConvolutionImageFilter upstream requests
+  ITK_TEST_EXPECT_EQUAL(inputMonitor->GetOutputRequestedRegions().size(), numOutputRequestedRegions);
+  ITK_TEST_EXPECT_EQUAL(inputMonitor->GetNumberOfUpdates(), numUpdates);
```
blowekamp, line 108: Don't we expect that the Convolution filter makes 10 different requested regions upstream? Causing the reader to read 10 regions. This looks like it just made one request and the inputs/reader read the whole image. Or maybe the reader can only read the whole image?

The reader's behavior is dependent on the type of input image, if the underlying ImageIO supports streaming. MetaIO should, as long as the image in not compressed.

tbirdso: Yes, good catch. I dug in and addressed two issues:
1. Tests were using `cthead1.png` and it appears streaming was not supported for PNG format. I have added `cthead1.mha` to ITKTestingData for this test, streaming appears to execute properly now.
2. After running into issues with the streamed requested region apparently exceeding the largest possible input region, I was able to track down where `itk::ConvolutionImageFilter` previously did not set the correct output region. Updated accordingly.

thewtex (approved):
:1st_place_medal: :rocket:

blowekamp:
Great! Thanks for adding the extra testing.

Leengit (approved):
:100:

PR #3392, opened 2022-04-22
STYLE: Remove unnecessary `using` type aliases from struct and enum definitions

N-Dekker:
In C, it might be useful to add such type aliases to `struct` and `enum` definitions (using `typedef`), but it is unnecessary in C++.

This pull request also removed duplicate lines of code from the definition of `ConditionVariableType`.

### DIFF
--- a/Modules/Compatibility/Deprecated/test/itkSpawnThreadTest.cxx
+++ b/Modules/Compatibility/Deprecated/test/itkSpawnThreadTest.cxx
@@ -24,7 +24,7 @@
 #include "itksys/SystemTools.hxx"
 #include "itkMutexLock.h"
 
-using SharedThreadData = struct
+struct SharedThreadData
 {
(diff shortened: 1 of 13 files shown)


Leengit (approved):

Leengit:
Maybe here too?:
https://github.com/InsightSoftwareConsortium/ITK/blob/ac684294e4c4d8387273d56ad97cdb5d98b270ee/Modules/IO/GE/src/Ge5xHdr.h#L60-L61
https://github.com/InsightSoftwareConsortium/ITK/blob/ac684294e4c4d8387273d56ad97cdb5d98b270ee/Modules/IO/LSM/src/itkLSMImageIO.cxx#L65-L66

dzenanz (approved):

hjmjohnson (approved):

PR #3394, opened 2022-04-25
ENH: Mark CMake variables as advanced

### LINKED ISSUE #3113: Mark third party CMake variables as advanced
We currently have a number of CMake variables related to 3rd party libraries polluting ITK's default CMake variable configuration options, which should be minimal. These should be marked as advanced with CMake's `mark_as_advanced` command in the ITK module CMake code. Variables include *EXPAT_** *NIFTI* and more.

dzenanz:
Closes #3113.


Leengit (approved):
Good, thank you.

PR #3398, opened 2022-04-28
STYLE: Clean-up code Windows version default-constructor `RealTimeClock`

N-Dekker:
Removed estimation of `FILETIME` for January 1, 1601 (`ft1`), as it is entirely
zero anyway, according to the Windows API specification.

Converted `FILETIME` to seconds without old-style LARGE_INTEGER/memcpy magic.

Removed redundant `this->` and `static_cast<__int64>`.

Used an empty pair of braces, `{}`, to zero-initialize SYSTEMTIME.

### DIFF
--- a/Modules/Core/Common/src/itkRealTimeClock.cxx
+++ b/Modules/Core/Common/src/itkRealTimeClock.cxx
@@ -37,45 +37,30 @@ RealTimeClock::RealTimeClock()
 
   m_Frequency = static_cast<FrequencyType>(frequency.QuadPart);
 
-  SYSTEMTIME st1;
-  SYSTEMTIME st2;
-  FILETIME   ft1;
-  FILETIME   ft2;
+  // Converts a FILETIME to the number of seconds since January 1, 1601.
+  const auto convertToSeconds = [](const FILETIME fileTime) {
+    const auto numberOfIntervals =
+      uint64_t{ fileTime.dwHighDateTime } * (uint64_t{ UINT32_MAX } + 1) + uint64_t{ fileTime.dwLowDateTime };
+    return static_cast<TimeStampType>(numberOfIntervals) / TimeStampType{ 1e7 };
+  };
 
-  ::memset(&st1, 0, sizeof(st1));
-  ::memset(&st2, 0, sizeof(st2));
+  SYSTEMTIME systemTime{};
+  systemTime.wYear = 1970;
+  systemTime.wMonth = 1;
+  systemTime.wDay = 1;
 
-  st1.wYear = 1601;
-  st1.wMonth = 1;
-  st1.wDay = 1;
+  FILETIME fileTime;
+  ::SystemTimeToFileTime(&systemTime, &fileTime);
 
-  st2.wYear = 1970;
-  st2.wMonth = 1;
-  st2.wDay = 1;
-
-  ::SystemTimeToFileTime(&st1, &ft1);
-  ::SystemTimeToFileTime(&st2, &ft2);
-
-  LARGE_INTEGER ui1;
-  LARGE_INTEGER ui2;
-
-  memcpy(&ui1, &ft1, sizeof(ui1));
-  memcpy(&ui2, &ft2, sizeof(ui2));
-
-  m_Difference = static_cast<TimeStampType>(ui2.QuadPart - ui1.QuadPart) / static_cast<TimeStampType>(1e7);
+  m_Difference = convertToSeconds(fileTime);
 
   FILETIME      currentTime;
-  LARGE_INTEGER intTime;
   LARGE_INTEGER tick;
 
   ::GetSystemTimeAsFileTime(&currentTime);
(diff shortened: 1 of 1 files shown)


Modules/Core/Common/src/itkRealTimeClock.cxx
```diff
@@ -35,7 +35,7 @@ RealTimeClock::RealTimeClock()
   LARGE_INTEGER frequency;
   ::QueryPerformanceFrequency(&frequency);
 
-  this->m_Frequency = static_cast<FrequencyType>(static_cast<__int64>(frequency.QuadPart));
+  m_Frequency = static_cast<FrequencyType>(frequency.QuadPart);
```
dzenanz, line 38: `this->` is sometimes required by GCC/clang, even if it compiles with MSVC.

dzenanz (approved):
Looks good on a glance.

Modules/Core/Common/src/itkRealTimeClock.cxx
```diff
@@ -35,51 +35,36 @@ RealTimeClock::RealTimeClock()
...
 
-  SYSTEMTIME st1;
-  SYSTEMTIME st2;
-  FILETIME   ft1;
-  FILETIME   ft2;
+  // Converts a FILETIME to the number of seconds since January 1, 1601.
+  const auto convertToSeconds = [](const FILETIME fileTime) {
+    const auto numberOfIntervals =
+      uint64_t{ fileTime.dwHighDateTime } * (uint64_t{ UINT32_MAX } + 1) + uint64_t{ fileTime.dwLowDateTime };
```
Leengit, line 43: Shift instead of multiplication by a power of 2?  Bitwise-or would even work.
```cpp
const auto numberOfIntervals = (uint64_t{ fileTime.dwHighDateTime } << 32) | uint64_t{ fileTime.dwLowDateTime };
```

Modules/Core/Common/src/itkRealTimeClock.cxx
N-Dekker: > `this->` is sometimes required by GCC/clang, even if it compiles with MSVC.

Thanks for your comment @dzenanz As far as I know, `this->` is only required for a class _template_, that is accessing a data member of one of its base classes. But please correct me if I'm wrong.

`RealTimeClock` is not a template, and it does not have a base class either. It just accesses its own data members. So I think all `this->` may be removed here.

dzenanz: And CI has succeeded, proving your point 😄

Modules/Core/Common/src/itkRealTimeClock.cxx
N-Dekker: > Shift instead of multiplication by a power of 2? Bitwise-or would even work.

I think you're correct, thanks @Leengit. But then again, it's a matter of style, and possibly code readability. Do we prefer multiplication + addition, or bit-shift + bitwise-or?

I'm in favor of multiplication + addition in this case, but if there are more people for bit-shift + bitwise-or, I will surrender  😃

dzenanz: Compiler might produce the same assembly either way. Even if it doesn't, a constructor is not a performance-critical part. I say good as-is.

N-Dekker: Actually my original plan was to do `const auto numberOfIntervals = bit_cast<uint64_t>(fileTime);` But then I thought it would be clearer (and safer) to use the arithmetic expression that I'm proposing here.

Leengit: I think that the result of `bit_cast` in this scenario might depend upon big endian vs. little endian; I wouldn't use it here.  As is works!

N-Dekker: > I think that the result of `bit_cast` in this scenario might depend upon big endian vs. little endian; I wouldn't use it here. As is works!

@Leengit OK, but the original code did memcpy from FILETIME to LARGE_INTEGER, which is basically what  `bit_cast` would do as well:
https://github.com/InsightSoftwareConsortium/ITK/blob/344e9bc58e69ed921f540c67d87dbb66fee4c466/Modules/Core/Common/src/itkRealTimeClock.cxx#L63

I was looking for a  `bit_cast` use case, by searching for "memcpy" in the code base. That's how I got here  😺

Leengit: I agree that `memcpy` and `bit_cast` are equivalent as far as endian goes.  I see at https://aakinshin.net/posts/datetime/ (search for BIGENDIAN within the page) that at least some people think endian matters in a context similar to ours.  

Though if it does matter then why haven't any of our tests failed on any system?  Regardless, the current approach is more reassuring.

Leengit (approved):

hjmjohnson (approved):

PR #3399, opened 2022-04-29
STYLE: Use `constexpr` and `bit_cast` within `GE4ImageIO::MvtSunf(int)`

N-Dekker:
And declare `sun_num` unsigned, to avoid unsigned-to-signed conversion.

### DIFF
--- a/Modules/IO/GE/src/itkGE4ImageIO.cxx
+++ b/Modules/IO/GE/src/itkGE4ImageIO.cxx
@@ -19,6 +19,7 @@
 #include "itkByteSwapper.h"
 #include "Ge4xHdr.h"
 #include "itksys/SystemTools.hxx"
+#include "itkBitCast.h"
 #include <iostream>
 #include <fstream>
 // From uiig library "The University of Iowa Imaging Group-UIIG"
@@ -344,11 +345,11 @@ GE4ImageIO::ReadHeader(const char * FileNameToRead)
 float
 GE4ImageIO ::MvtSunf(int numb)
 {
-#define signbit 020000000000U
-#define dmantissa 077777777U
-#define dexponent 0177U
-#define smantissa 037777777U
-#define smantlen 23U
+  constexpr auto signbit = 020000000000U;
+  constexpr auto dmantissa = 077777777U;
+  constexpr auto dexponent = 0177U;
+  constexpr auto smantissa = 037777777U;
+  constexpr auto smantlen = 23U;
   ByteSwapper<int>::SwapFromSystemToBigEndian(&numb);
   unsigned int dg_exp = (numb >> 24) & dexponent;
   unsigned int dg_sign = numb & signbit;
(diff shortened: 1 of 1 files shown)


Modules/IO/GE/src/itkGE4ImageIO.cxx
```diff
@@ -344,11 +345,11 @@ GE4ImageIO::ReadHeader(const char * FileNameToRead)
...
 {
-#define signbit 020000000000U
-#define dmantissa 077777777U
-#define dexponent 0177U
-#define smantissa 037777777U
-#define smantlen 23U
+  constexpr auto signbit = 020000000000U;
+  constexpr auto dmantissa = 077777777U;
+  constexpr auto dexponent = 0177U;
+  constexpr auto smantissa = 037777777U;
```
Leengit, line 351: I would rather see my 4-byte bit masks described in hexadecimal than in octal:
```cpp
  constexpr auto signbit   = 0x80000000U;
  constexpr auto dmantissa = 0x00FFFFFFU;
  constexpr auto dexponent = 0x7FU;
  constexpr auto smantissa = 0x007FFFFFU;
```

Also, does the trailing `U` count as a use of the type on the right-hand side, so that `auto` is appropriate on the left-hand side?  (Regardless, it is definitely a step up from the `#define`.)

Modules/IO/GE/src/itkGE4ImageIO.cxx
```diff
@@ -369,9 +370,7 @@ GE4ImageIO ::MvtSunf(int numb)
     sun_exp = 255;
   }
   dg_mantissa = dg_mantissa << 1;
-  int   sun_num = dg_sign | (sun_exp << smantlen) | ((dg_mantissa >> 9) & smantissa);
-  float x;
-  memcpy(&x, &sun_num, sizeof(x));
-  return x;
+  const unsigned int sun_num{ dg_sign | (sun_exp << smantlen) | ((dg_mantissa >> 9) & smantissa) };
```
Leengit, line 373: We could use multiplication and addition rather than shifts and bitwise-or, as in #3398.  (For those who didn't read #3398, ... that's a joke.  I prefer the latter.)

Leengit (approved):
Looks good, thank you.

Modules/IO/GE/src/itkGE4ImageIO.cxx
N-Dekker: > Personally, I would rather see my 4-byte bit masks described in hexadecimal than in octal

Thanks for the suggestions. Hexadecimal looks more common to me as well, but as it wasn't originally my code, I'd rather leave it octal for now.

>  Also, does the trailing `U` count as a use of the type on the right-hand side, so that auto is appropriate on the left-hand side?

I think so, yes.

hjmjohnson (approved):

PR #3400, opened 2022-04-30
Add const to first parameter of "SwapWrite" member functions, and other ByteSwapper style improvements

N-Dekker:
Declared the first parameter of its "SwapWrite" member functions pointer-to-const, and made various other `ByteSwapper` style improvements

### DIFF
--- a/Modules/Core/Common/include/itkByteSwapper.hxx
+++ b/Modules/Core/Common/include/itkByteSwapper.hxx
@@ -27,6 +27,7 @@
  *=========================================================================*/
 #ifndef itkByteSwapper_hxx
 #define itkByteSwapper_hxx
+#include <algorithm> // For swap.
 #include <memory>
 #include <cstring>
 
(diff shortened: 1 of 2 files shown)


Modules/Core/Common/include/itkByteSwapper.hxx
```diff
@@ -273,10 +274,8 @@ template <typename T>
 void
 ByteSwapper<T>::Swap2(void * pin)
 {
-  auto *               p = static_cast<unsigned short *>(pin);
-  const unsigned short h1 = (*p) << static_cast<short unsigned int>(8);
-  const unsigned short h2 = (*p) >> static_cast<short unsigned int>(8);
-  *p = h1 | h2;
```
N-Dekker, line 279: Note that the original code here would have undefined behavior if the pointer passed as argument would point to a type _other_ than signed or unsigned `short`.

hjmjohnson (approved):
LGTM.

PR #3402, opened 2022-04-30
STYLE: Use `bit_cast` to convert random seeds from uint32_t to int32_t

N-Dekker:
Instead of doing `*reinterpret_cast<const int32_t *>(&seed)`, which appears
more complicated.

### DIFF
--- a/Modules/Filtering/ImageNoise/include/itkAdditiveGaussianNoiseImageFilter.hxx
+++ b/Modules/Filtering/ImageNoise/include/itkAdditiveGaussianNoiseImageFilter.hxx
@@ -18,6 +18,7 @@
 #ifndef itkAdditiveGaussianNoiseImageFilter_hxx
 #define itkAdditiveGaussianNoiseImageFilter_hxx
 
+#include "itkBitCast.h"
 #include "itkImageScanlineIterator.h"
 #include "itkTotalProgressReporter.h"
 #include "itkNormalVariateGenerator.h"
@@ -53,7 +54,7 @@ AdditiveGaussianNoiseImageFilter<TInputImage, TOutputImage>::ThreadedGenerateDat
   typename Statistics::NormalVariateGenerator::Pointer randn = Statistics::NormalVariateGenerator::New();
   const uint32_t                                       seed = Self::Hash(this->GetSeed(), uint32_t(indSeed));
   // Convert the seed bit for bit to int32_t
-  randn->Initialize(*reinterpret_cast<const int32_t *>(&seed));
+  randn->Initialize(bit_cast<int32_t>(seed));
 
   // Define the portion of the input to walk for this thread, using
   // the CallCopyOutputRegionToInputRegion method allows for the input
--- a/Modules/Filtering/ImageNoise/include/itkShotNoiseImageFilter.hxx
+++ b/Modules/Filtering/ImageNoise/include/itkShotNoiseImageFilter.hxx
@@ -18,6 +18,7 @@
 #ifndef itkShotNoiseImageFilter_hxx
 #define itkShotNoiseImageFilter_hxx
 
+#include "itkBitCast.h"
 #include "itkMersenneTwisterRandomVariateGenerator.h"
 #include "itkImageScanlineIterator.h"
 #include "itkTotalProgressReporter.h"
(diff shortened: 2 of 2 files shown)


N-Dekker:
We might consider `static_cast<int32_t>(seed)` rather than `bit_cast<int32_t>(seed)`, in these two cases. With C++20, `static_cast<int32_t>(seed)` is guaranteed to behave exactly like  `std::bit_cast<int32_t>(seed)`: it will do the conversion "bit by bit".

With C++14, the behavior of `static_cast<int32_t>(seed)` is officially still "implementation-defined", in case of overflow. However, in practice it will just behave the same as `bit_cast<int32_t>(seed)` (but possibly faster).

For details, see the comments by Jonathan Wakely at https://github.com/isocpp/CppCoreGuidelines/issues/1517#issuecomment-1114564409

So what do you think in this case, `bit_cast` or `static_cast`?

Leengit:
To restate the obvious:

- The problem with `reinterpret_cast` is that, _**theoretically**_, a pointer to `uint32_t` might not be a legal pointer to `int32_t`; e.g., if for some unfathomable reason the compiler chose different rules for alignment such as the former has to have an address that is multiple of 4 but the latter has to be a multiple of 16.  I estimate the probability of this problem actually arising in practice as somewhere between zero and never.
- The problem with `static_cast` is that, _**theoretically**_, prior to C++20, a `uint32_t` value that is in excess of MAX_INT32, when cast to `int32_t`, could be set to, say, MAX_INT32, rather than having its bit pattern interpreted as a negative `int32_t`.  I estimate the probability of this problem actually arising in practice as a little bit higher, perhaps between 5\*never and 10\*never.
- The problem with `bit_cast` is that it might use something slower, like `memcpy`, if the compiler (or our substitute implementation) can't figure out that it should instead be implemented using, effectively, `reinterpret_cast` or `static_cast` (because on that system it is already verified that these implementations actually work).

I'd use `bit_cast`.  I can't imagine that this will ever be a performance issue.  If speed somehow becomes an issue in some circumstance then we can try to use some template magic with `bit_cast` to implement it as `static_cast` when that is appropriate (with a slew of static_asserts to verify that it is doing what we want).

N-Dekker:
Thanks for your reply, @Leengit First of all, avoiding undefined behavior is not merely a theoretical problem. The optimizer might do all kinds of clever optimizations, based on the _assumption_ that the program does not have undefined behavior. For example, GCC optimized away the `if (ptr == nullptr)` in the following code, just because it assumes that `int val = *ptr` has no undefined behavior:

```
int GetValue(int* ptr) {
    int val = *ptr;
    if (ptr == nullptr) return 0;
    return val;
}
```
https://godbolt.org/z/jo9693Was

However, you're right, apparently this particular reinterpret_cast does not have undefined behavior, as I just understood from https://github.com/isocpp/CppCoreGuidelines/issues/1517#issuecomment-1114830775 (A. Jiang) as well, referring to https://eel.is/c++draft/basic.lval#11.2

So apparently it's just a matter of style  😺 My preference is still to replace the reinterpret_cast, one way or the other, as reinterpret_cast is often frowned upon. With full C++20 support I would prefer static_cast. For now, `itk::bit_cast` would also be OK to me.

Leengit:
> The optimizer might do all kinds of clever optimizations, based on the assumption that the program does not have undefined behavior.

A concrete example related to the present case would be great if you have it!  I am thinking that in
```cpp
unsigned int
f(int i)
{
  unsigned int u = static_cast<unsigned int>(i);
  if (u > INT_MAX)
    return 0;
  return u;
}
```
there are 0 compilers that will optimize away the `if` statement on the premise that it is true only when the previous statement is undefined behavior.

N-Dekker:
> there are 0 compilers that will optimize away the `if` statement on the premise that it is true only when the previous statement is undefined behavior.

@Leengit But in your example, I don't see undefined behavior either. `static_cast<unsigned int>(i)` should be well defined, even when `i` itself is negative. Right?

Leengit:
> With C++14, the behavior of `static_cast<int32_t>(seed)` is officially still "implementation-defined", in case of overflow. 

@N-Dekker Are you making a distinction between "undefined behavior" and "implementation defined" behavior?

N-Dekker:
> Are you making a distinction between "undefined" behavior and "implementation-defined" behavior?

I should certainly do so 😺 So I hope I did.

    auto maxUint = static_cast<unsigned int>(-1); // OK: UINT_MAX
    auto i = static_cast<int>(maxUint); // Implementation-defined with C++14, just -1 with C++20
    auto f = reinterpret_cast<const float&>(maxUint); // Undefined behavior

Right?  https://godbolt.org/z/rTYedj61v

Leengit:
> I should certainly do so smiley_cat So I hope I did.

Thank you for the clarification.  I missed that the first time.

hjmjohnson (approved):

PR #3403, opened 2022-05-02
ENH: Increase `MeshIO` classes coverage

jhlegarreta:
- STYLE: Make style consistent with the ITK style
- STYLE: Use the `itkNameOfTestExecutableMacro` macro in tests
- DOC: Fix typos in documentation, comments and exception messages
- STYLE: Improve wording in `itk::GiftiMeshIO` exception messages
- ENH: Add boolean macros to `itk::MeshIOBase` data update ivars
- BUG: Avoid querying about null variable size
- BUG: Fix test driver name for `MeshOBJ` and `MeshOFF`
- ENH: Increase coverage for `MeshIO` classes

## PR Checklist

### DIFF
--- a/Modules/IO/MeshBase/include/itkMeshIOTestHelper.h
+++ b/Modules/IO/MeshBase/include/itkMeshIOTestHelper.h
@@ -1,0 +1,433 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         https://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+#ifndef itkMeshIOTestHelper_h
+#define itkMeshIOTestHelper_h
+
+#include "itkArray.h"
+#include "itkCommonEnums.h"
+#include "itkCovariantVector.h"
+#include "itkFixedArray.h"
+#include "itkDiffusionTensor3D.h"
+#include "itkMacro.h"
+#include "itkMatrix.h"
+#include "itkMeshIOBase.h"
+#include "itkRGBAPixel.h"
+#include "itkRGBPixel.h"
+#include "itkSymmetricSecondRankTensor.h"
+#include "itkVariableLengthVector.h"
+#include "itkVariableSizeMatrix.h"
+#include "itkVector.h"
+#include <complex>
+#include <iostream>
+
+
+// Define a local macro for variable to command testing to avoid including
+// itkTestingMacros.h, which causes link issues as the module hosting this
+// file is not a testing module.
+#define LOCAL_ITK_TEST_SET_GET_VALUE(variable, command)                 \
+  CLANG_PRAGMA_PUSH                                                     \
+  CLANG_SUPPRESS_Wfloat_equal if (variable != command) CLANG_PRAGMA_POP \
+  {                                                                     \
+    std::cerr << "Error in " << #command << std::endl;                  \
+    std::cerr << "  In " __FILE__ ", line " << __LINE__ << std::endl;   \
+    std::cerr << "Expected " << variable << std::endl;                  \
+    std::cerr << "but got  " << command << std::endl;                   \
+    return EXIT_FAILURE;                                                \
+  }                                                                     \
+  ITK_MACROEND_NOOP_STATEMENT
+
+
+template <typename TMeshIO>
+int
+TestBaseClassMethodsMeshIO(typename TMeshIO::Pointer meshIO)
+{
+  using FloatType = float;
+
+  FloatType floatValue = 1.0;
+  bool      usePointPixel = true;
+  meshIO->SetPixelType(floatValue, usePointPixel);
+  LOCAL_ITK_TEST_SET_GET_VALUE(1, meshIO->GetNumberOfPointPixelComponents());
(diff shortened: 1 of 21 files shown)


jhlegarreta:
For some reason, I was only able to run the tests from my IDE and not using `ctest` except for the `GiftiMeshIO` ones. Will need to see if they get executed by the CIs. Else, if anyone is willing to checkout the branch and have a look, that would be awesome.

Modules/IO/MeshBase/include/itkMeshIOTestHelper.h
```diff
@@ -0,0 +1,372 @@
...
+  {
+    std::cout << ext << std::endl;
+  }
+
+
+  return EXIT_SUCCESS;
+}
+
+void
+AllocateBuffer(void ** data, itk::IOComponentEnum componentType, itk::SizeValueType bufferSize)
```
jhlegarreta, line 323: Not sure this is the best thing to do, requiring the caller methods to remember to delete the pointers later. Suggestions are welcome.

jhlegarreta:
Not sure why the compilation is failing even after reading the error message. I got this working locally on my Win 7, MSVC 2019. Suggestions are welcome.

Modules/IO/MeshBase/include/itkMeshIOTestHelper.h
thewtex: @PranjalSahu what do you think?

thewtex:
To keep the builds happy, we may need to:

```
#include <iostream>
```

in itkMeshFileTestHelper.h

and remove `#include "itkTestingMacros.h"` from itkMeshFileTestHelper.h and use the expanded macros there for module dependency sake.

Thank you so much for working on this @jhlegarreta !

jhlegarreta:
@thewtex https://github.com/InsightSoftwareConsortium/ITK/pull/3403#issuecomment-1115090925 :+1: thanks. Will try to address part of the reported issues later tonight.

jhlegarreta:
Push-forced.

@thewtex:
- Removed the macros in 13398cf for the same reason.
- For the concerned file, I defined a local macro, instead of expanding it 80 times. Let me know if this is fine.

Modules/IO/MeshBase/include/itkMeshIOTestHelper.h
jhlegarreta: Definitely something better is required as the compiler is complaining about the deletion:
```
Modules/IO/MeshBYU/test/itkBYUMeshIOTest.cxx:169:12: warning: deleting 'void*' is undefined [-Wdelete-incomplete]   delete[] pointBuffer
             ^~~~~~~~~~~
```

e.g. https://open.cdash.org/viewBuildError.php?type=1&buildid=7889131

jhlegarreta:
> For some reason, I was only able to run the tests from my IDE and not using ctest except for the GiftiMeshIO ones. Will need to see if they get executed by the CIs. Else, if anyone is willing to checkout the branch and have a look, that would be awesome.

Probably related to d0a02cd, and me copying the test driver name for the rest of the test names added.

jhlegarreta:
Not sure why the `BYUMeshIO` test is timing out; I've got it passing locally. Same applies to the `GifiMeshIO` tests; they're passing locally on my Win 7, MSVC 2019. https://github.com/InsightSoftwareConsortium/ITK/pull/3403#discussion_r862562894 might be related to the latter at least, but I have not come up with a better solution defined in a single place and that can used by all tests.

jhlegarreta:
The seemingly random failures in tests/BYU test timeout might be due to multiple tests trying to write to/read from the same files in parallel (?), so I renamed the files.

Let's see if this makes at least the tests pass and leave only the compiler warnings.

As for the latter, in addition to my caution about the buffer allocation method, the CIs complain about deleting a `void *` pointer. If we are to get away from the helper method, and unless a better solution is found, reading/writing the cells might need to be done within each `switch` case, creating a different buffer type for each case, to immediately delete them after reading the data. This is a lot of code duplication/not elegant at all IMO.

jhlegarreta:
> Let's see if this makes at least the tests pass and leave only the compiler warnings.

Although I guess it is better that way, did not solve the failures :pensive:. Cannot debug this on my local machine as tests are passing.

thewtex:
> As for the latter, in addition to my caution about the buffer allocation method, the CIs complain about deleting a void * pointer. If we are to get away from the helper method, and unless a better solution is found, reading/writing the cells might need to be done within each switch case, creating a different buffer type for each case, to immediately delete them after reading the data. This is a lot of code duplication/not elegant at all IMO.

@jhlegarreta I agree. I think we would remove `ReadPoints` from these tests so the other test coverage improvements can be integrated. There is some indirect coverage when `itk::MeshFileReader` is used. It is better to have the other coverage improvements than no improvements.

dzenanz:
Running tests in parallel is accomplished by `ctest -j8`. Have you already tried this Jon?

jhlegarreta:
Sorry for the delay, I've been away from keyboard for the last couple of weeks.

> Running tests in parallel is accomplished by ctest -j8. Have you already tried this Jon?

I haven't. In any case, now (https://github.com/InsightSoftwareConsortium/ITK/pull/3403#issuecomment-1118072401) output files have different names, so my guess about failures being caused by different tests trying to write the same file no longer applies.

> @jhlegarreta I agree. I think we would remove ReadPoints from these tests so the other test coverage improvements can be integrated. There is some indirect coverage when itk::MeshFileReader is used. It is better to have the other coverage improvements than no improvements.

Rather than the discrepancy in the written/read data (note that I had commented the checks where this has been noticed), the CIs complained about deleting a `void *` pointer. I have transferred the buffer allocation blocks to each testing file, and will wait the build results before taking further steps. Will eventually open an issue to keep track of this code duplication.

Rebased on `master` and push-forced.

dzenanz (approved):

dzenanz:
Is this ready for merging, Jon?

jhlegarreta:
> Is this ready for merging, Jon?

Although there are a few things that I would have loved to avoid or improve (notably, the buffer allocation/deletion, and investigating why some round-trip tests are failing -I know adding `FIXME`s is undesired-, and why the `BYU` call to `WriteMeshInformation()` was timing out), tests were passing locally and I have no chance to look into these in another platform, so I'll open an issue to keep track of these.

Summarizing, unless someone else has the chance to look at the above, this is ready to go on my end.

PR #3405, opened 2022-05-02
xarray image name

thewtex:
- ENH: Transfer object name in xarray.DataArray conversion
- ENH: Refactor xarray support into separate module
- STYLE: Apply black to extras.py, xarray.py


Wrapping/Generators/Python/itk/support/xarray.py
```diff
@@ -0,0 +1,139 @@
...
+        pass
+
+
+__all__ = [
+    "xarray_from_image",
+    "image_from_xarray",
+]
+
+
+def xarray_from_image(l_image: "itkt.ImageOrImageSource") -> "xr.DataArray":
```
tbirdso, line 36: Were `xarray_from_image` and `image_from_xarray` previously defined somewhere else? I see where they are defined here but not where they were removed from.

thewtex: Yes, they were in extras.py -- they have been removed from there.

Wrapping/Generators/Python/itk/support/extras.py
```diff
@@ -667,7 +666,12 @@ def xarray_from_image(l_image: "itkt.ImageOrImageSource") -> "xr.DataArray":
     for key in metadata:
         if not key in ignore_keys:
             attrs[key] = metadata[key]
-    data_array = xr.DataArray(array_view, dims=dims, coords=coords, attrs=attrs)
+    name = "image"
+    if l_image.GetObjectName():
+        name = l_image.GetObjectName()
+    data_array = xr.DataArray(
+        array_view, name=name, dims=dims, coords=coords, attrs=attrs
+    )
```
tbirdso, line 674: These changes are in `xarray_from_image` which is redefined in `xarray.py`.

Wrapping/Generators/Python/itk/support/xarray.py
tbirdso: It looks like they are still present in this PR. Could you please take a look at `extras.py:627`?

thewtex: Ah, good catch! There was a copy in *Tests/extra.py*, which was removed, but not *extras.py* -- fixed now!

tbirdso: 😕 The copy I am seeing that still exists in the Github diff is at `itk/support/extras.py:627`. Please see my other comment below as well.

thewtex: I had some issues with the rebase -- just pushed an update.

Wrapping/Generators/Python/itk/support/extras.py
tbirdso: Fixed in recent push

Wrapping/Generators/Python/itk/support/xarray.py
```diff
@@ -0,0 +1,139 @@
...
+    import numpy as np
+
+    array_view = itk.array_view_from_image(l_image)
+    l_spacing = itk.spacing(l_image)
+    l_origin = itk.origin(l_image)
+    l_size = itk.size(l_image)
+    direction = np.flip(itk.array_from_matrix(l_image.GetDirection()))
+    image_dimension = l_image.GetImageDimension()
+
+    image_dims: Tuple[str, str, str] = ("x", "y", "z", "t")
```
PranjalSahu, line 56: Is the count correct here ?

thewtex: No -- good catch! Addressed.

PR #3414, opened 2022-05-04
BUG: XMLFileOutputWindowTest failed subsequent runs

hjmjohnson:
The XMLFileOutputWindowTest only passed on the iniital run
of the test.  Subsequent runs would append to the file and
cause incorrect file line counts.  Merged two test scenerios
into one test run, and explicitly define the testing environment.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/test/itkXMLFileOutputWindowTest.cxx
+++ b/Modules/Core/Common/test/itkXMLFileOutputWindowTest.cxx
@@ -21,97 +21,78 @@
 #include "itkTestingMacros.h"
 #include <fstream>
 #include <cstdio>
+#include "itksys/SystemTools.hxx"
 #ifdef _WIN32
 #  include <direct.h>
 #  define cwd _getcwd
-#  define OS_SEP "\\"
+static const std::string OS_SEP{ "\\" };
 #else
 #  include "unistd.h"
 #  define cwd getcwd
-#  define OS_SEP "/"
+static const std::string OS_SEP{ "/" };
 #endif
 
 
-int
-itkXMLFileOutputWindowTest(int argc, char * argv[])
+static bool
+DoTestXMLFileOutputWindow(std::string currentLoggerFilename, const unsigned int requiredLinesInFile)
 {
-  if (argc < 1)
+  if (currentLoggerFilename.empty())
   {
-    std::cerr << "Missing parameters." << std::endl;
-    std::cerr << "Usage: " << itkNameOfTestExecutableMacro(argv) << " [filename]" << std::endl;
-    return EXIT_FAILURE;
+    const std::string fileBaseName = []() -> std::string {
+      // If no input filename is provided, remove the default file to avoid counting existing lines when
+      // contents are appended.
+      itk::XMLFileOutputWindow::Pointer logger = itk::XMLFileOutputWindow::New();
+      // In order to initialize the filename, some text needs to be written first
+      const char * regularText = "text";
+      logger->DisplayText(regularText);
+
+      // Get the filename
+      return std::string{ logger->GetFileName() };
+    }();
+    const std::string cwd = itksys::SystemTools::GetCurrentWorkingDirectory();
+    currentLoggerFilename = cwd + OS_SEP + fileBaseName;
+    std::remove(currentLoggerFilename.c_str());
   }
+
 
   itk::XMLFileOutputWindow::Pointer logger = itk::XMLFileOutputWindow::New();
 
   ITK_EXERCISE_BASIC_OBJECT_METHODS(logger, XMLFileOutputWindow, FileOutputWindow);
 
-  if (argc > 1)
-  {
-    logger->SetFileName(argv[1]);
-  }
-
   logger->FlushOn();
   logger->AppendOn();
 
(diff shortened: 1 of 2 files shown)


Modules/Core/Common/test/itkXMLFileOutputWindowTest.cxx
```diff
@@ -21,97 +21,76 @@
...
 #  define cwd getcwd
-#  define OS_SEP "/"
+static const std::string OS_SEP{ "/" };
 #endif
 
 
-int
-itkXMLFileOutputWindowTest(int argc, char * argv[])
+static bool
+do_test(std::string current_logger_filename, const unsigned int required_lines_in_file)
```
jhlegarreta, line 37: Use medial case for method and variable names.

Modules/Core/Common/test/itkXMLFileOutputWindowTest.cxx
```diff
@@ -21,97 +21,76 @@
...
+    // If no input filename is provided, remove the default file to avoid counting existing lines when
+    // contents are appended.
+    itk::XMLFileOutputWindow::Pointer logger = itk::XMLFileOutputWindow::New();
+    // In order to initialize the filename, some text needs to be written first
+    const char * regularText = "text";
+    logger->DisplayText(regularText);
+
+    // Get the filename
+    const std::string fileBaseName = logger->GetFileName();
+    const std::string CWD = itksys::SystemTools::GetCurrentWorkingDirectory();
```
jhlegarreta, line 50: `CWD` to lowercase.

Modules/Core/Common/test/itkXMLFileOutputWindowTest.cxx
```diff
@@ -120,9 +99,39 @@ itkXMLFileOutputWindowTest(int argc, char * argv[])
...
+    std::cerr << "Usage: " << itkNameOfTestExecutableMacro(argv) << " <filename>" << std::endl;
+    return EXIT_FAILURE;
+  }
 
-  std::cout << "Test finished." << std::endl;
-  return EXIT_SUCCESS;
+  // First test with no filename given, to test autogenerating the filename
+  constexpr unsigned int numLinesExpected = 7;
+  bool                   status = do_test("", numLinesExpected);
+  const std::string      test1_msg{ status ? "TEST ONE PASSED" : "TEST ONE FAILED" };
```
jhlegarreta, line 124: Use medical case for variable naming (no underscores).

Modules/Core/Common/test/itkXMLFileOutputWindowTest.cxx
```diff
@@ -120,9 +99,39 @@ itkXMLFileOutputWindowTest(int argc, char * argv[])
...
+}
+
 
+int
+itkXMLFileOutputWindowTest(int argc, char * argv[])
+{
+  if (argc != 2)
+  {
+    std::cerr << "Missing parameters." << std::endl;
+    std::cerr << "Usage: " << itkNameOfTestExecutableMacro(argv) << " <filename>" << std::endl;
```
jhlegarreta, line 117: According to the ITK SW guide, we should use square brackets for optional arguments. However, in this case, the filename is not optional, so please remove the `<>`.

jhlegarreta (commented):
Thanks for doing this Hans.

Besides my style-related comments, as mentioned in https://github.com/InsightSoftwareConsortium/ITK/pull/3383#issuecomment-1117410726, I had attempted to use `std::remove` (by providing it the appropriate full path) in my local system, and the file was not being deleted. This was also happening on another test (see 94ee975). I ignore if your OS or the C++ standard you are using is successful at this, but it should be verified across OS and for the lowest C++ standard ITK supports.

hjmjohnson:
https://en.cppreference.com/w/cpp/io/c/remove
```
If the file is currently open by the current or another process, the behavior of this function is implementation-defined (in particular, POSIX systems unlink the file name, although the file system space is not reclaimed even if this was the last hard link to the file until the last running process closes the file, Windows does not allow the file to be deleted)
```

The file was previously open inside the 'logger' object.  That makes sense why the std::remove did not work previously.  

My version also had a similar problem, but it is now fixed. Would you be able to test the new solution on windows?

Modules/Core/Common/test/itkXMLFileOutputWindowTest.cxx
```diff
@@ -21,97 +21,78 @@
...
 #  define cwd getcwd
-#  define OS_SEP "/"
+static const std::string OS_SEP{ "/" };
 #endif
 
 
-int
-itkXMLFileOutputWindowTest(int argc, char * argv[])
+static bool
+DoTestXMLFileOutputWindow(std::string current_logger_filename, const unsigned int required_lines_in_file)
```
jhlegarreta, line 37: Medial case naming required here for the variables.

jhlegarreta (commented):
Works like a charm @hjmjohnson :100:.

Missing to address an in-line comment and to fix a typo in the commit message (scenerios: scenarios) (given that you'll need to amend the commit).

jhlegarreta (approved):
👌

PR #3416, opened 2022-05-06
ENH: Update TubeTK to support Write3DImagesAs4DImage in python

aylward:
Bump TubeTK version to support generation of 4D images in python.

TubeTK provides ITK classes that simplify reading/writing 4D images as lists of 3D images, even if ITK's Python API wasn't compiled with 4D image support (which is the default).   The TubeTK classes, however, accepted ITK image smartpointers as member function arguments, but to support python wrapping, they should instead accepted ITK image raw pointers.  This has been fixed in TubeTK in the tag given here.


thewtex (approved):
:100: :metal:

PR #2672, opened 2021-07-30
ENH: SurfaceSpatialObject templated over point type

aylward:
By templating over point type, derived surface classes can specify new point types with additional meta data.  This is the same design patter used for TubeSpatialObjects.

Also, updated the normal computation to work for 2D data, instead of just 3D data, and marked old code as Legacy.

PR #3417, opened 2022-05-07
ENH: SurfaceSpatialObject templated over point type

aylward:
By templating over point type, derived surface classes can specify new point types with additional meta data. This is the same design patter used for TubeSpatialObjects.

Also, updated the normal computation to work for 2D data, instead of just 3D data, and marked old code as Legacy.


aylward:
This replaces pr #2672

thewtex (approved):
:+1: :1234:

PR #3435, opened 2022-05-17
BUG: Prevent Image corruption in Dask serialization

thewtex:
Similar to: https://github.com/InsightSoftwareConsortium/ITK/commit/d244d4bf71be396e6d0007f87a92b2e926230adf


PranjalSahu:
@thewtex Did you see any such error cases for images?
I didn't do this change because haven't seen errors yet. But it is better to replace.

PranjalSahu (approved):

thewtex:
I did see the similar behavior with images, but I did not verify that this addressed the issue. However, if it was causing issues with the cell data, it should be the same with the image pixel buffer.

tbirdso (approved):

PR #3438, opened 2022-05-18
BUG: Copy to xarray by default

tbirdso:
Previously observed behavior where ITK data was released after ITK
Python filter went out of scope, resulting in cascading and difficult to
trace fatal memory r/w error on xarray access. This change makes data
copy the default behavior while exposing `in_place` option for cases
where operations are contained within a single block and behavior
benefits from in-place operations.

Addresses discussion in [https://github.com/spatial-image/multiscale-spatial-image/issues/44](https://github.com/spatial-image/multiscale-spatial-image/issues/44)

## PR Checklist

### DIFF
--- a/Wrapping/Generators/Python/itk/support/xarray.py
+++ b/Wrapping/Generators/Python/itk/support/xarray.py
@@ -34,12 +34,19 @@ __all__ = [
 ]
 
 
-def xarray_from_image(l_image: "itkt.ImageOrImageSource") -> "xr.DataArray":
+def xarray_from_image(l_image: "itkt.ImageOrImageSource", view:bool=False) -> "xr.DataArray":
     """Convert an itk.Image to an xarray.DataArray.
 
     Origin and spacing metadata is preserved in the xarray's coords. The
     Direction is set in the `direction` attribute.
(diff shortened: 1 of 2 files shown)


Wrapping/Generators/Python/itk/support/xarray.py
```diff
@@ -34,7 +34,7 @@
 ]
 
 
-def xarray_from_image(l_image: "itkt.ImageOrImageSource") -> "xr.DataArray":
+def xarray_from_image(l_image: "itkt.ImageOrImageSource", in_place:bool=False) -> "xr.DataArray":
```
thewtex, line 37: A more conventional name would be `view` here

thewtex (commented):
@tbirdso excellent!

Naming improvement suggested inline

Wrapping/Generators/Python/itk/support/xarray.py
tbirdso: Fixed

tbirdso:
Added a short test as well to verify that the "view" is valid in the same block.

thewtex (approved):
:100:

PR #3445, opened 2022-05-22
wasm dict members

thewtex:
- BUG: Add missing name to image_to_dict
- BUG: Remove extra dimension member from Mesh, PointSet dict's


thewtex:
@PranjalSahu sorry, I had an extra `dimension` in the itk-wasm docs.

PranjalSahu:
Also please check if the cellBufferSize entry is correct or not ?
I had doubts what exactly was needed there.  Currently it uses the size of vector_container.
https://github.com/InsightSoftwareConsortium/ITK/blob/28df2aff2020c1f28bb08b974aa54c92f2ac415a/Wrapping/Generators/Python/itk/support/extras.py#L839

thewtex:
> Also please check if the cellBufferSize entry is correct or not ?
> I had doubts what exactly was needed there. Currently it uses the size of vector_container.

:+1: yes, that's right.

PranjalSahu (approved):

PR #3458, opened 2022-06-02
STYLE: Call convenience member function, `ImageBase::SetRegions(region)`

N-Dekker:
STYLE: Call convenience member function, `ImageBase::SetRegions(region)`

Instead of doing something like:

    image->SetLargestPossibleRegion(region);
    image->SetBufferedRegion(region);
    image->SetRequestedRegion(region);

Did use Notepad++ v8.3.3 Regular expressions:

    Find what:
    ->SetLargestPossibleRegion\((.+)\);\r\n.+->SetBufferedRegion\(\1\);\r\n.+->SetRequestedRegion\(\1\);

    Replace with:
    ->SetRegions\($1\);

### DIFF
--- a/Modules/Core/Common/include/itkFloodFilledFunctionConditionalConstIterator.hxx
+++ b/Modules/Core/Common/include/itkFloodFilledFunctionConditionalConstIterator.hxx
@@ -82,9 +82,7 @@ FloodFilledFunctionConditionalConstIterator<TImage, TFunction>::InitializeIterat
   m_TemporaryPointer = TTempImage::New();
   typename TTempImage::RegionType tempRegion = this->m_Image->GetBufferedRegion();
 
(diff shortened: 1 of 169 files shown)


dzenanz (approved):
Since when does `SetRegions` exist?

N-Dekker:
> Since when does `SetRegions` exist?

Jim Miller (@millerjv) commit ea87880367b31ab9831041d56917c34cbafe6f93 said on 15 May 2012:

> moved the method SetRegions() out of the subclasses of ImageBase and put it directly in ImageBase.

So it has been around for a while  😄

dzenanz (approved):
There was a suspiciously small number of changes before this force-push 😄

N-Dekker:
> There was a suspiciously small number of changes before this force-push

Before the force-push, I had excluded the source files of the tests. I wanted to be sure that the original changes would pass the tests, before adjusting those tests!

Issue #3455, opened 2022-06-01
`MeshIO` classes Write methods fail/timeout and round-trip test fail


jhlegarreta:
### Description

Some of the `MeshIO` (`Modules/IO/Mesh*`) classes `Write*` methods fail/timeout or the round-trip (write data, read back and compare the read values to the expected ones) fail. 

### Steps to Reproduce

The failing tests are located in the `test` module of each `Modules/IO/Mesh*` module. The failing calls have been commented out in the tests and a `FIXME` tag was put in the comments while adding the tests for these classes in #3403.

### Expected behavior

Write methods work; tests do not timeout; round-trip checks pass.

### Actual behavior

Some of the write methods do not work; tests timeout (`MeshBYU`); round-trip checks do not pass.

### Reproducibility

%100.

### Versions

`master`.

### Environment

Issues were being reported in all platforms used by the CI builds. However, tests were passing on a Windows 7, MSVC 2019 machine.

### Additional Information

Cross-referencing https://github.com/InsightSoftwareConsortium/ITK/pull/3403#issuecomment-1141132531

Slightly related to this is the need to improve the buffer allocation method/buffer deletion, as noted in the referenced comment.

jhlegarreta:
The buffer allocation/de-allocation method introduced memory leaks :unamused::
https://open.cdash.org/viewDynamicAnalysis.php?buildid=7954824

I will push a WIP PR so that if @blowekamp or anyone that is able to run Valgrind on the branch and has the bandwidth can confirm whether the fix removes the leaks from the target test.

jhlegarreta:
> I will push a WIP PR so that if @blowekamp or anyone that is able to run Valgrind on the branch and has the bandwidth can confirm whether the fix removes the leaks from the target test.

PR #3459

PR #3460, opened 2022-06-05
BUG: Let MeshIOTestHelper AllocateBuffer return `std::shared_ptr<void>`

N-Dekker:
Let AllocateBuffer (from MeshIOTestHelper) return a `std::shared_ptr<void>`,
instead of producing a void pointer, to ensure that it is deleted correctly,
automatically.

Aims to fix Valgrind errors, saying:

> Mismatched free() / delete / delete []

As reported and discussed by Jon Haitz Legarreta Gorroño (@jhlegarreta) and Mihail Isakov (@issakomi) at pull request #3459

### DIFF
--- a/Modules/IO/MeshBase/include/itkMeshIOTestHelper.h
+++ b/Modules/IO/MeshBase/include/itkMeshIOTestHelper.h
@@ -375,59 +375,59 @@ TestBaseClassMethodsMeshIO(typename TMeshIO::Pointer meshIO)
   return EXIT_SUCCESS;
 }
 
-void
-AllocateBuffer(void ** data, itk::IOComponentEnum componentType, itk::SizeValueType bufferSize)
+namespace itk
 {
-  void * buffer = nullptr;
+namespace MeshIOTestHelper
+{
 
+template <typename T>
+std::shared_ptr<void>
+MakeSharedArray(const std::size_t bufferSize)
+{
+  return std::shared_ptr<void>(new T[bufferSize], std::default_delete<T[]>());
+}
+
+inline std::shared_ptr<void>
+AllocateBuffer(itk::IOComponentEnum componentType, itk::SizeValueType bufferSize)
+{
   switch (componentType)
   {
     case itk::IOComponentEnum::CHAR:
-      buffer = new char[bufferSize];
-      break;
+      return MakeSharedArray<char>(bufferSize);
     case itk::IOComponentEnum::UCHAR:
-      buffer = new unsigned char[bufferSize];
-      break;
+      return MakeSharedArray<unsigned char>(bufferSize);
     case itk::IOComponentEnum::USHORT:
-      buffer = new unsigned short[bufferSize];
-      break;
+      return MakeSharedArray<unsigned short>(bufferSize);
     case itk::IOComponentEnum::SHORT:
-      buffer = new short[bufferSize];
-      break;
+      return MakeSharedArray<short>(bufferSize);
     case itk::IOComponentEnum::UINT:
-      buffer = new unsigned int[bufferSize];
-      break;
+      return MakeSharedArray<unsigned int>(bufferSize);
     case itk::IOComponentEnum::INT:
-      buffer = new unsigned int[bufferSize];
-      break;
+      return MakeSharedArray<int>(bufferSize);
     case itk::IOComponentEnum::ULONG:
-      buffer = new unsigned long[bufferSize];
-      break;
+      return MakeSharedArray<unsigned long>(bufferSize);
     case itk::IOComponentEnum::LONG:
-      buffer = new long[bufferSize];
-      break;
+      return MakeSharedArray<long>(bufferSize);
     case itk::IOComponentEnum::LONGLONG:
-      buffer = new long long[bufferSize];
-      break;
+      return MakeSharedArray<long long>(bufferSize);
     case itk::IOComponentEnum::ULONGLONG:
-      buffer = new unsigned long long[bufferSize];
-      break;
+      return MakeSharedArray<unsigned long long>(bufferSize);
     case itk::IOComponentEnum::FLOAT:
-      buffer = new float[bufferSize];
-      break;
+      return MakeSharedArray<float>(bufferSize);
     case itk::IOComponentEnum::DOUBLE:
-      buffer = new double[bufferSize];
-      break;
+      return MakeSharedArray<double>(bufferSize);
     case itk::IOComponentEnum::LDOUBLE:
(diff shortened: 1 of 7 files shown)


N-Dekker:
Apparently Azure.Mac-1654425863237/Darwin-Build7250 cannot handle the code  😢 

https://open.cdash.org/viewBuildError.php?buildid=7958143 says:

>Modules/IO/MeshBase/include/itkMeshIOTestHelper.h:388:10: error: no matching conversion for functional-style cast from 'char *' to 'std::shared_ptr<char []>'

    return std::shared_ptr<T[]>(new T[bufferSize]);
           ^~~~~~~~~~~~~~~~~~~~~~~~

Very disappointing, as the code looks pretty clear to me.

Modules/IO/MeshBase/include/itkMeshIOTestHelper.h
```diff
@@ -375,59 +375,59 @@ TestBaseClassMethodsMeshIO(typename TMeshIO::Pointer meshIO)
   return EXIT_SUCCESS;
 }
 
-void
-AllocateBuffer(void ** data, itk::IOComponentEnum componentType, itk::SizeValueType bufferSize)
+namespace itk
+{
+namespace MeshIOTestHelper
```
jhlegarreta, line 380: Is there some reason why this method should be included in an additional namespace?

Not sure why I did not include in the `itk` namespace `TestBaseClassMethodsMeshIO` as well.

N-Dekker: > Is there some reason why this method should be included in an additional namespace?

Thanks for asking @jhlegarreta 

I did not want to keep `AllocateBuffer` in the global namespace, because I think it may easily cause naming conflicts with third party libraries (now or in the future). However, I did not want to place it _directly_ in the _root_ of the `itk` namespace (as in "itk::AllocateBuffer"), because it is not part of the public ITK interface. And "itk::AllocateBuffer" might also conflict with other parts of ITK  (now or in the future).  So to avoid any conflict, I chose `itk::MeshIOTestHelper::AllocateBuffer`. This very much clarifies where the function comes from. And it prevents any naming conflict. Is that OK to you?

dzenanz (approved):

Modules/IO/MeshBase/include/itkMeshIOTestHelper.h
jhlegarreta: @N-Dekker thanks for the explanation. Makes sense.

jhlegarreta:
So :100: for doing this @N-Dekker. If we are absolutely sure that this does not incur memory leaks, I'd be for merging it.

Leengit:
I am not understanding why
```cpp
return std::shared_ptr<void>(new T[bufferSize], std::default_delete<T[]>());
```
is a good idea.  That is, `shared_ptr<void>` acts like a `void *` which avoids type checking and, maybe, is not guaranteed to cast well to `int *` or `double *` or anything other than `char *`.  With the explicit deleter `std::default_delete<T[]>()`, which is necessary prior to C++17 for arrays, that code should work as
```cpp
return std::shared_ptr<T>(new T[bufferSize], std::default_delete<T[]>());
```

N-Dekker:
> That is, shared_ptr<void> acts like a void * which avoids type checking and, maybe, is not guaranteed to cast well to int * or double * or anything other than char *. 

Thanks for your comment @Leengit, but can you please explain? When a `void` pointer is initialized by `new T[bufferSize]`, casting that `void` pointer back to `T *` should work well, OK?

Leengit:
> When a `void` pointer is initialized by `new T[bufferSize]`, casting that `void` pointer back to `T *` should work well, OK?

Agreed, that works.

Because `shared_ptr<void>` and `void *` are less type safe than `shared_ptr<T>` and `T *`, I'd go with the latter pair.

N-Dekker:
> Because `shared_ptr<void>` and `void *` are less type safe than `shared_ptr<T>` and `T *`, I'd go with the latter pair.

I understand your point, but please note that the `AllocateBuffer` function is being used by MeshIO tests, which are _not_ templated. If I understand correctly, the runtime types of the buffers depends on input data files that are being read by the tests. So the specific type of a buffer is not yet known at compile time. Note that the tested MeshIO member functions that deal with these buffers also expect a void pointer to the buffer as input parameter.

This is why the original code had `AllocateBuffer` using a `void *`. This pull request just aims to solve the current memory problem (in the master branch), by making it a `shared_ptr<void>` instead.  Does that sound OK to you?

Leengit:
> This is why the original code had `AllocateBuffer` using a `void *`. This pull request just aims to solve the current memory problem (in the master branch), by making it a `shared_ptr<void>` instead. Does that sound OK to you?

That sounds very good to me -- solving the immediate problem is the way to go, even when it doesn't solve all other problems.  For that future pull request ... I'd use `shared_ptr<T>` to share a `T *`  if `T` is known at compile time.

N-Dekker:
>  For that future pull request ... I'd use `shared_ptr<T>` to share a `T *`  if `T` is known at compile time.

@Leengit  _If_ the buffer element type `T` would be known at compile time, I think I'd rather use `std::unique_ptr<T[]>` than `std::shared_ptr<T>`. `std::unique_ptr` is more lightweight than `std::shared_ptr`.

jhlegarreta:
As @issakomi said in https://github.com/InsightSoftwareConsortium/ITK/pull/3459#issuecomment-1151567968,  this fixed the memory leaks I introduced (thanks again both @issakomi and @N-Dekker :100:):
https://open.cdash.org/viewDynamicAnalysis.php?buildid=7966303

We are back down to two.

And the tests, together with this fix, have uncovered, through the `itkFreeSurferMeshIOTestBinaryTest` (`itkFreeSurferMeshIOTest`) test, a leak that has gone unnoticed so far in https://github.com/InsightSoftwareConsortium/ITK/blob/23da3dfa79152648f3d38050ee147df47f41ce0a/Modules/IO/MeshFreeSurfer/src/itkFreeSurferBinaryMeshIO.cxx#L479: 
https://open.cdash.org/viewDynamicAnalysisFile.php?id=9535345 

The leak in `itkMeshTest` predates PR #3403.
https://open.cdash.org/viewDynamicAnalysisFile.php?id=9535344

Maybe these two have to do with the round-trip tests in `MeshIO` classes not working as expected (cross-ref issue #3455).

jhlegarreta:
> Oh, absolutely. My bad. Will submit a PR later today. Thanks for having investigated this @issakomi 100.

PR #3467.

PR #3479, opened 2022-07-01
BUG: Do not use assert in Python wrappnig builds

### LINKED ISSUE #3473: Catching C++ itkAssertOrThrowMacro in Python
### Description
This happens generally in ITK, but a specific example is: the `itk-spcn / StructurePreservingColorNormalzationFilter` cannot handle all input images and the C++ code currently throws via an `itkAssertOrThrowMacro` when it discovers that it cannot handle an input image (e.g., because the RGB image is nonetheless too monochromatic). The C++ throw includes an informative message. However, this triggers a segmentation fault when used in Python (after `pip install itk-spcn` and `import itk`) and does not relay the informative message to the user.

This should be changed so that the Python user of this code gets the informative error message.

### Version
ITK sources as of May 27, 2022.

### Environment

Ubuntu.  Perhaps others.

thewtex:
Always throw an exception that can be interpreted by the Python
interpreter. Do not assert / crash.

Closes #3473


dzenanz (approved):

Leengit (approved):
For reasons that I don't have time to explore right now, my ITK build is not as successful as it should be.  (I haven't built ITK for a while so it is likely an issue with my set up rather than ITK.)

That said, the code here looks like it will solve #3473 that I submitted, so I am going to approve.

PR #2493, opened 2021-04-19
STYLE: Replace postfix by prefix increment in `for` loops in Core/Common

N-Dekker:
Followed common C++ guidelines, including:

Google C++ Style Guide: "Use the prefix form (++i) of the increment and
decrement operators unless you need postfix semantics."
https://google.github.io/styleguide/cppguide.html#Preincrement_and_Predecrement

Herb Sutter, Andrei Alexandrescu - C++ Coding Standards: 101 Rules,
Guidelines, and Best Practices: "Prefer the canonical form of ++ and --.
Prefer calling the prefix forms"
https://www.oreilly.com/library/view/c-coding-standards/0321113586/ch29.html

Applied at GNU bash, by the following command:

	find . \( -iname *.cxx -or -iname *.hxx -or -iname *.h \) -exec perl -pi -w -e 's/(\s+for \(.*; .+); (\w+)\+\+\)/$1; ++$2)/g;' {} \;

Manually replaced a few more postfix increments, which were not detected
by this regex command, as remarked by Mikhail Isakov (@issakomi).

PR #3484, opened 2022-07-11
Replace postfix with prefix increment and decrement

N-Dekker:
Followed common C++ guidelines, including:

Google C++ Style Guide: "Use the prefix form (++i) of the increment and
decrement operators unless you need postfix semantics."
https://google.github.io/styleguide/cppguide.html#Preincrement_and_Predecrement

Herb Sutter, Andrei Alexandrescu - C++ Coding Standards: 101 Rules,
Guidelines, and Best Practices: "Prefer the canonical form of ++ and --.
Prefer calling the prefix forms"
https://www.oreilly.com/library/view/c-coding-standards/0321113586/ch29.html

Follow-up to pull request #2493 commit f81578bbb6242032b611d7e3e4dca3b7557fbbe0 "STYLE: Replace postfix by prefix increment in `for` loops in Core/Common"

### DIFF
--- a/Modules/Compatibility/Deprecated/include/itkChildTreeIterator.hxx
+++ b/Modules/Compatibility/Deprecated/include/itkChildTreeIterator.hxx
@@ -106,7 +106,7 @@ template <typename TTreeType>
 auto
 ChildTreeIterator<TTreeType>::Next() -> const ValueType &
 {
-  m_ListPosition++;
+  ++m_ListPosition;
   this->m_Position = m_ParentNode->GetChild(m_ListPosition);
   if (this->m_Position == nullptr)
   {
--- a/Modules/Compatibility/Deprecated/include/itkInOrderTreeIterator.h
+++ b/Modules/Compatibility/Deprecated/include/itkInOrderTreeIterator.h
@@ -139,7 +139,7 @@ InOrderTreeIterator<TTreeType>::FindNextNode() const
     {
       return help;
     }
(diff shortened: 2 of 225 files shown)


Modules/Core/Common/include/itkConceptChecking.h
```diff
@@ -604,7 +604,7 @@ struct IncrementDecrementOperators
     constraints()
     {
       ++a;
-      a--;
+      --a;
```
dzenanz, line 607: It is funny this wasn't done in the first pass :D

dzenanz (approved):

Modules/Core/Common/include/itkConceptChecking.h
N-Dekker: Hmmm... I think I should not have modified `IncrementDecrementOperators::constraints()` anyway. I guess it's part of the constraints that both `a--` and `--a` are supported!

N-Dekker: Update: with the [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/2021f7ce8395e8471df58237408a48fcc167e890..917355e41da7b4969d73f2b5b078675651c87c90) that I just did, I undid those itkConceptChecking.h changes, and squashed the two commits.

dzenanz (approved):

hjmjohnson (approved):

PR #3494, opened 2022-07-22
BUG: Use absolute module path PyCapsule_Import

thewtex:
Addresses:

```
ImportError: PyCapsule_Import could not import module "_ITKCommonPython"
```

That can occur when unpickling for Dask, for example.

According to the `PyCapsule_Import` documentation, internally it calls
`PyImport_ImportModule`. According to the `PyImport_ImportModule`
documentation, this requires the absolute path -- we add the `itk.`
prefix.

In most other contexts, the `_ITKCommonPython` module is likely to
already be imported. Since `_ITKCommonPython` gets loaded via
PyImport_ImportModule, it is not registered with the ITK lazy loading
module system, so we add `from . import _ITKCommon`.

### DIFF
--- a/Wrapping/Generators/Python/PyBase/pyBase.i
+++ b/Wrapping/Generators/Python/PyBase/pyBase.i
@@ -7,6 +7,8 @@ import collections
 from sys import version_info as _version_info
 if _version_info < (3, 7, 0):
     raise RuntimeError("Python 3.7 or later required")
+
+from . import _ITKCommonPython
 %}
 
 //By including pyabc.i and using the -py3 command line option when calling SWIG,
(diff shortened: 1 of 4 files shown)


tbirdso (approved):

GenevieveBuckley (approved):
This looks good to me. Thanks for working on it!

I can't say for sure it will for sure fix the [dask example](https://github.com/dask/dask-blog/issues/138) (I wasn't able to build ITK last time I tried, so I can't test it now), but I certainly can't see this causing any problems for either ITK or dask.

hjmjohnson (approved):

PranjalSahu:
```python
import dask
import itk
from dask import delayed

if __name__== "__main__":
    from dask.distributed import Client, LocalCluster
    cluster = LocalCluster()
    client = Client(cluster)

    @delayed
    def performsum(m):
        return m.GetNumberOfPoints()

    m = itk.Mesh[itk.F, 3].New()
    a = performsum(m).compute()
    print(a)
```

Tested with a locally built wheel.
It works fine.

PR #3505, opened 2022-07-28
Replace all occurrences of `(&x)->Print` with `x.Print`

N-Dekker:
Aims to simplify the code.

### DIFF
--- a/Modules/Core/Common/include/itkEventObject.h
+++ b/Modules/Core/Common/include/itkEventObject.h
@@ -107,7 +107,7 @@ protected:
 inline std::ostream &
 operator<<(std::ostream & os, const EventObject & e)
 {
-  (&e)->Print(os);
+  e.Print(os);
   return os;
 }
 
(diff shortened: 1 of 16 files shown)


dzenanz (approved):
Have you tried to figure out why this syntax was used in the first place? To work around some compiler bug or limitation, or some other reason?

Also preferably squash the commits.

dzenanz:
Ah, I now see that first commit is for library code, and second for tests. Indifferent about squashing now.

N-Dekker:
> Have you tried to figure out why this syntax was used in the first place? To work around some compiler bug or limitation, or some other reason?

Good question! Honestly I have no idea. But obviously the dot operator (to access a member of an object) has been used extensively already in throughout the ITK source base. Specifically, `Print` calls of the form `x.Print(y)` occur many places already in ITK. So this PR would improve overall code consistency  😄 

My (uneducated) guess is that some contributors in the past might just have felt more comfortable with the arrow `->` than the dot. While I actually feel less conformable with the arrow, as it might be used on a pointer that could be null.

PR #3538, opened 2022-08-10
ENH: Add VTKPolyDataMeshIO GTest tests, testing lossless write and read

N-Dekker:
Added `VTKPolyDataMeshIO` GoogleTest, having the following unit tests:

 - LosslessWriteAndReadOfPoints
 - SupportWriteAndReadOfNaNCoordValues

### DIFF
--- a/Modules/IO/MeshVTK/test/CMakeLists.txt
+++ b/Modules/IO/MeshVTK/test/CMakeLists.txt
@@ -78,3 +78,9 @@ itk_add_test(NAME itkVTKPolyDataMeshIOTestInOutBinary
       ${ITK_TEST_OUTPUT_DIR}/binary2binaryvtk2mhaHeadMRVolume.mha
       0 1 1 1 1 72 72 2 2 1 1
 )
+
+set(ITKIOMeshVTKGTests
+  itkVTKPolyDataMeshIOGTest.cxx
+)
+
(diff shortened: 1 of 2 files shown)


N-Dekker:
The proposed tests write some .vtk files to the current directory (in order to read them back, afterwards). Is it OK for these tests to just leave those little .vtk files there, or should they be removed by the tests themselves, at the end?

dzenanz:
It would probably be best if we passed in the path to a temporary directory and wrote the files there. Passing a temp path from CMakeLists.txt file should be easy.

N-Dekker:
> It would probably be best if we passed in the path to a temporary directory and wrote the files there. Passing a temp path from CMakeLists.txt file should be easy.

Thanks @dzenanz. It isn't entirely obvious to me how to do it via CMake. Could the temporary directory be specified with the proposed CreateGoogleTestDriver, call? It now just says:

    CreateGoogleTestDriver(ITKIOMeshVTK "${ITKIOMeshVTK-Test_LIBRARIES}" "${ITKIOMeshVTKGTests}")

Currently, the vtk files created by these tests appear locally at the build tree, on my Windows system, at `<build>\ITK\Modules\IO\MeshVTK\test\`. So I don't think it will bother people. So we could also leave it the way it is right now. What do you think?

dzenanz (commented):
I meant use `${ITK_TEST_OUTPUT_DIR}` as an argument to the test executable, but I now realize that this is GTest, nor a regular test. Having the temporary files in the build folder are fine by me.

dzenanz (approved):

Issue #3539, opened 2022-08-10
MeshFileReader/VTKPolyDataMeshIO hangs forever on non-numeric chars in "POINTS" section (including "Infinity" and "NaN")


N-Dekker:
### Description

It appears that `MeshFileReader` hangs forever when trying to read an ASCII .vtk file that has non-numeric chars in "POINTS" section (using VTKPolyDataMeshIO). Even when the file was created by `MeshFileWriter`, using VTKPolyDataMeshIO as well. This may happen when some of the coordinates of the points has a NaN or an infinite value.

Note than `MeshFileReader` _does_ correctly process reading NaN and infinite values from a _binary_ .vtk file. So the issue only occurs with non-binary (ASCII) .vtk files.

### Steps to Reproduce

```
#include "itkMesh.h"
#include "itkMeshFilereader.h"
#include "itkMeshFileWriter.h"

int
main()
{
  using MeshType = itk::Mesh<int>;
  using NumericLimits = std::numeric_limits<float>;
  const auto inputMesh = MeshType::New();
  inputMesh->SetPoint(0, itk::MakePoint(42.0f, NumericLimits::infinity(), NumericLimits::quiet_NaN()));
  itk::WriteMesh(inputMesh, "fileName.vtk");

  auto outputMesh = itk::ReadMesh<MeshType>("fileName.vtk"); // <== hangs forever...!
}

```

VTK file created by itk::WriteMesh(inputMesh, "fileName.vtk"):

```
# vtk DataFile Version 2.0
File written by itkPolyDataMeshIO
ASCII
DATASET POLYDATA
POINTS 1 float
42 Infinity NaN

```

### Expected behavior

- MeshFileReader should be able to read any file by created by `MeshFileWriter`, so it should interpret "Infinity" as the floating point value infinity, and "NaN" as Not-A-Number.
- In other cases, when the "POINTS" section has non-numeric characters, MeshFileReader should throw an exception, rather than just hanging forever.

### Actual behavior

MeshFileReader hangs inside the `while` loop within `itk::VTKPolyDataMeshIO::ReadPointsBufferAsASCII`, at https://github.com/InsightSoftwareConsortium/ITK/blob/9ed9070c74ae00cb7c258bc550cf2dd6ae469c43/Modules/IO/MeshVTK/include/itkVTKPolyDataMeshIO.h#L201

### Reproducibility

100%

### Versions

[...]

PR #3542, opened 2022-08-10
BUG: VTKPolyDataMeshIO should not hang on inf and NaN in ASCII .vtk file

N-Dekker:
`MeshFileReader`/`VTKPolyDataMeshIO` did hang forever when trying to read an
ASCII .vtk file that contained non-numeric chars in "POINTS" section, for
example "Infinity" or "NaN".

With this commit, "inf" and "Infinity" will be interpreted as the floating point
value infinity, while "nan" and "NaN" will be interpreted as Not-A-Number. Other
strings of non-numeric characters in the list of point components (coordinates)
will cause the reader to throw an appropriate `itk::ExceptionObject`.

Addresses issue https://github.com/InsightSoftwareConsortium/ITK/issues/3539
"MeshFileReader/VTKPolyDataMeshIO hangs forever on non-numeric chars in "POINTS"
section (including "Infinity" and "NaN")"

### DIFF
--- a/Modules/IO/MeshVTK/include/itkVTKPolyDataMeshIO.h
+++ b/Modules/IO/MeshVTK/include/itkVTKPolyDataMeshIO.h
@@ -205,11 +205,7 @@ protected:
       if (line.find("POINTS") != std::string::npos)
       {
         /**  Load the point coordinates into the itk::Mesh */
-        SizeValueType numberOfComponents = this->m_NumberOfPoints * this->m_PointDimension;
-        for (SizeValueType ii = 0; ii < numberOfComponents; ++ii)
-        {
-          inputFile >> buffer[ii];
-        }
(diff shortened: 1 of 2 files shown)


dzenanz (approved):

N-Dekker:
> Is it worth it to go completely case insensitive to catch "Inf", "Nan", etc.?

@Leengit Good question! "Infinity" and "NaN" are essential because those are the strings produced by ITK's MeshFileWriter, using `itk::VTKPolyDataMeshIO` (which internally uses https://github.com/google/double-conversion). However, it appears that VTK prints infinity and Not-A-Number via printf-style "%g" formatting. Which yields "inf" and "nan" (both entirely lowercase).

I'm not aware of any other string representation of  infinity and Not-A-Number that is actually being used in .vtk files.

dzenanz (approved):

thewtex (approved):
:1234: :+1:

PR #3529, opened 2022-08-04
BUG: Allow nifti load to fail

hjmjohnson:
Ensure that attempts to write to invalid file paths causes error to be reported. Ref #1958.

## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [X] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [X] Added test (or behavior not changed)

PR #3548, opened 2022-08-14
BUG: fixed memory leaks in MINC IO

issakomi:
MINC IO does not free allocations before throwing an exception sometimes,
s. discussion https://github.com/InsightSoftwareConsortium/ITK/pull/3529

https://open.cdash.org/viewDynamicAnalysis.php?buildid=8097352

(reviewed _writing_ an image).

### DIFF
--- a/Modules/IO/MINC/src/itkMINCImageIO.cxx
+++ b/Modules/IO/MINC/src/itkMINCImageIO.cxx
@@ -26,6 +26,37 @@
 
 #include "itk_minc2.h"
 
+extern "C"
+{
+  void
+  MINCIOFreeTmpDimHandle(unsigned int size, midimhandle_t * ptr)
+  {
+    int          error = 0;
+    unsigned int x = 0;
+    if (!ptr)
+    {
+      /*
+       * Should never happen.
+       */
+#ifndef NDEBUG
+      printf("MINCIOFreeTmpDimHandle: ptr is null pointer");
+#endif
+      return;
+    }
+    for (; x < size; ++x)
+    {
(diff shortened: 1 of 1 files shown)


issakomi:
Don't know why 'Label PR' test fails, maybe something _github_ internal?

issakomi:
Have changed a little to avoid repeating code and tested exit points (to avoid double free, etc).

jhlegarreta (approved):
Thanks for doing this @issakomi. Have not tested but looks good.

Maybe, following for comment in the related issue:
> I have done trivial test, without using itkIOTestHelper.h, just read MHA and write MNC to wrong path, the leaks are there. If the path is correct -- there are no problems, 

Adding a test such as the one you did would be convenient.

gdevenyi:
@vfonov

issakomi:
> > I have done trivial test, without using itkIOTestHelper.h, just read MHA and write MNC to wrong path, the leaks are there. If the path is correct -- there are no problems,
> 
> Adding a test such as the one you did would be convenient.

it is really very simple [test](https://drive.google.com/file/d/1y8EcexH4ijOQ5-j0zIavvkNDKlumlJcJ/view?usp=sharing), to try `valgrind --leak-check=full ./test` without ITK's testing subsystem, already merged new ITK tests with intentional wrong path (which discovered [leaks](https://open.cdash.org/viewDynamicAnalysis.php?buildid=8097352)) are doing the same. Maybe later more tests could be done to trigger exceptions and increase coverage, e.g. 4D image.

dzenanz (approved):

vfonov:
> @vfonov

Looks ok

PR #3549, opened 2022-08-15
ENH: Support reading a mesh of non-3D points from a .vtk PolyData file

N-Dekker:
`itk::MeshFileWriter` does allow writing non-3D points to a .vtk file, so it
appears both reasonable and useful for `itk::MeshFileReader` to allow reading
those points back.

Implemented by setting the PointDimension within MeshFileReader::GenerateData(),
and tested by GTest `VTKPolyDataMeshIO.SupportsAnyPointDimensionGreaterThanZero`.

Related to issue https://github.com/slicersalt/ITKShape/issues/8 "Add 2D mesh support" by Matt McCormick (@thewtex).

### DIFF
--- a/Modules/IO/MeshBase/include/itkMeshFileReader.hxx
+++ b/Modules/IO/MeshBase/include/itkMeshFileReader.hxx
@@ -524,6 +524,8 @@ template <typename TOutputMesh, typename ConvertPointPixelTraits, typename Conve
 void
 MeshFileReader<TOutputMesh, ConvertPointPixelTraits, ConvertCellPixelTraits>::GenerateData()
 {
+  m_MeshIO->SetPointDimension(OutputPointDimension);
+
   typename TOutputMesh::Pointer output = this->GetOutput();
 
   // Test if the file exists and if it can be opened.
--- a/Modules/IO/MeshVTK/src/itkVTKPolyDataMeshIO.cxx
+++ b/Modules/IO/MeshVTK/src/itkVTKPolyDataMeshIO.cxx
@@ -343,7 +343,6 @@ VTKPolyDataMeshIO ::ReadMeshInformation()
 
       // Get number of Points
       ss >> this->m_NumberOfPoints;
-      this->m_PointDimension = 3; // vtk only support 3 dimensional points
 
       // Get point component type
       StringType pointType;
--- a/Modules/IO/MeshVTK/test/itkVTKPolyDataMeshIOGTest.cxx
+++ b/Modules/IO/MeshVTK/test/itkVTKPolyDataMeshIOGTest.cxx
@@ -33,6 +33,15 @@
 
 namespace
 {
+template <unsigned int VDimension>
+auto
+MakePointOfIncreasingCoordValues()
+{
+  itk::Point<float, VDimension> point;
+  std::iota(point.begin(), point.end(), 1.0f);
+  return point;
+}
+
 
 // Expects that VTKPolyDataMeshIO supports writing points, and then reading them back losslessly. (When a NaN coordinate
 // value is written, a NaN coordinate value is expected to be read back.)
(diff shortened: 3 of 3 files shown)


dzenanz (approved):

tbirdso (approved):
Looks great, thank you for putting this together @N-Dekker !

N-Dekker:
For the record, the initial "itkVTKPolyDataMeshIO.cxx" commit (9403e3b37a2250fe6fa2ef61de7c19cf014689ba, "ENH: Add mesh IO", 26 August 2011) already prevented reading non-3D points from a .vtk file: https://github.com/InsightSoftwareConsortium/ITK/blob/9403e3b37a2250fe6fa2ef61de7c19cf014689ba/Modules/IO/Mesh/src/itkVTKPolyDataMeshIO.cxx#L175

    this->m_PointDimension = 3; // vtk only support 3 dimensional points

Still it appeared quite easy to add non-3D support, just by removing that particular line of code from "itkVTKPolyDataMeshIO.cxx", and adding a `m_MeshIO->SetPointDimension(OutputPointDimension)` call to the `MeshFileReader`.

N-Dekker:
Linux-Build8777-PR3549-Reading-non-3D-VTKPolyDataLegacyRemoved https://open.cdash.org/viewBuildError.php?type=1&buildid=8100674 says:

```
In file included from Modules/Core/Common/include/itkQuadrilateralCell.h:171:0,
                 from Modules/Core/Common/include/itkHexahedronCell.h:21,
                 from Modules/IO/MeshBase/include/itkMeshFileReader.h:23,
                 from Modules/IO/MeshVTK/test/itkVTKPolyDataMeshIOGTest.cxx:20:
/.../s/Modules/Core/Common/include/itkQuadrilateralCell.hxx: In member function 'bool itk::QuadrilateralCell<TCellInterface>::EvaluatePosition(typename itk::QuadrilateralCell<TCellInterface>::Superclass::CoordRepType*, typename itk::QuadrilateralCell<TCellInterface>::Superclass::PointsContainer*, typename itk::QuadrilateralCell<TCellInterface>::Superclass::CoordRepType*, typename itk::QuadrilateralCell<TCellInterface>::Superclass::CoordRepType*, double*, typename itk::QuadrilateralCell<TCellInterface>::Superclass::InterpolationWeightType*) [with TCellInterface = itk::CellInterface<int, itk::CellTraitsInfo<1, float, float, long unsigned int, long unsigned int, long unsigned int, itk::Point<float, 1>, itk::VectorContainer<long unsigned int, itk::Point<float, 1> >, std::set<long unsigned int> > >]':

Modules/Core/Common/include/itkQuadrilateralCell.hxx:326:26: warning: iteration 1 invokes undefined behavior [-Waggressive-loop-optimizations]

         fcol[j] += pt[j] * weights[i];
                    ~~~~~~^~~~~~~~~~
Modules/Core/Common/include/itkQuadrilateralCell.hxx:324:34: note: within this loop

       for (unsigned int j = 0; j < CellDimension; ++j)
```

I think this warning is specifically for the unit test of 1D point support. The warning does make some sense, as `QuadrilateralCell` has `CellDimension = 2`, so access to `pt[j]` goes wrong when `pt` is a 1D point.

However, adding 1D support to `QuadrilateralCell` seems beyond the scope of this pull request. For now, I find it much more important to have tested 2D, and possibly 4D support for .vtk file reading. So for now, I'll just remove the 1D test from the unit test, with the next force-push.

dzenanz (approved):

N-Dekker:
Note that originally, `itk::VTKPolyDataMeshIO` did already "accidentally" support reading 2D points. In last months version of `ReadPointsBufferAsASCII`, at https://github.com/InsightSoftwareConsortium/ITK/blob/87f8716216a5e3199897a1f7f0754c9b3d0d1697/Modules/IO/MeshVTK/include/itkVTKPolyDataMeshIO.h#L207-L212

        /**  Load the point coordinates into the itk::Mesh */
        SizeValueType numberOfComponents = this->m_NumberOfPoints * this->m_PointDimension;
        for (SizeValueType ii = 0; ii < numberOfComponents; ++ii)
        {
          inputFile >> buffer[ii];
        }

This code would very much overestimate `numberOfComponents` for 2D, because `m_PointDimension` was set to 3, even for 2D.  But then, in practice, the `inputFile >> buffer[ii]` insertion would fail _silently_ once the last point coordinate was read. But then, all of its 2D point coordinates were usually correctly read already!

With pull request #3542 commit 678b97eaebed1a1fddcb77f64cfb5e293700edea (BUG: VTKPolyDataMeshIO should not hang on inf and NaN in ASCII .vtk file), error checking is added, so the insertion of components from the input file into the buffer will no longer _silently_ fail. Instead such a failure will now cause an itk::ExceptionObject to be thrown. Which is why `numberOfComponents` should not be overestimated anymore, in order to support 2D.

thewtex:
@N-Dekker thank you so much for working on this! :pray: 

Please see my comment in https://github.com/slicersalt/ITKShape/issues/8

Can you please verify the test outputs can be read with:

- ParaView
- 3D Slicer
- [meshio](https://pypi.org/project/meshio/)

?

N-Dekker:
Thanks for merging, @dzenanz I hope it's also fine to Matt (@thewtex) now. It's really useful for elastix/transformix to allow 2D points in a .vtk file.

thewtex:
> 3D Slicer did not complain, and it certainly did not crash, when I gave it a .vtk file, containing 2D points. I do not know whether it can do any further processing on them. But I can imagine that it likes 3D points better, given the name, "3D Slicer".

If it was loaded correctly, you should have been able to visualize the data.

There are many mesh libraries that use VTK or have their own implementation that follow the VTK standard. For other tools to use 2D .vtk files ITK produces, the test outputs will need to change.

For 2D and 4D meshes (including point sets), try the [itk-wasm `*.iwm` format](https://github.com/InsightSoftwareConsortium/itk-wasm).

PR #3551, opened 2022-08-16
ENH: Enable building HL library for the bundled HDF5

dzenanz:
This is needed by netCDF, which is needed for Zarr support.


thewtex (approved):
:+1:

PR #3557, opened 2022-08-17
Fix exception message

hjmjohnson:
The suggested member function does not exist.  Use one of the two
    available member functions to suggest the correct behavior for
    using the reference image information.


jhlegarreta (approved):

thewtex (approved):
:+1:

dzenanz (approved):

PR #3563, opened 2022-08-23
ENH: Add typehints to python wheel packages

hjmjohnson:
See https://peps.python.org/pep-0561/#stub-only-packages
Packaging type hints with package so use "itk", if making separate typehint
package use "itk-stubs"
set(ITK_STUB_BASENAME "itk")

The typehint files provide improved IDE introspection
capabilities for ITK.


thewtex (approved):
Thanks, Hans!

PR #3564, opened 2022-08-23
STYLE: Declare Ellipsoid SpatialFunction orientations as a fixed matrix

N-Dekker:
Declared the private (internal) EllipsoidInteriorExteriorSpatialFunction data
member `m_Orientations` as a `vnl_matrix_fixed`, instead of a raw pointer to
dynamically allocated memory.

Defaulted the `EllipsoidInteriorExteriorSpatialFunction` destructor, as it now
no longer needs to `delete` internal data pointers anymore.

Following C++ Core Guidelines, April 10, 2022, "Prefer scoped objects, don’t
heap-allocate unnecessarily",
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r5-prefer-scoped-objects-dont-heap-allocate-unnecessarily

PR #3565, opened 2022-08-23
STYLE: Declare WindowedSincInterpolateImageFunction data as fixed arrays

N-Dekker:
Declared the private (internal) WindowedSincInterpolateImageFunction data
members `m_WeightOffsetTable` and `m_WeightOffsetTable` as fixed-size C-style
arrays, instead of raw pointers to dynamically allocated memory.

Declared the private data members `m_WindowSize` and `m_OffsetTableSize` both
`static constexpr`, to ensure compile-time evaluation.

Defaulted the default-constructor and destructor of
`WindowedSincInterpolateImageFunction`, as they now no longer need to
explicitly manage the internal data.

Following C++ Core Guidelines, April 10, 2022, "Prefer scoped objects, don’t
heap-allocate unnecessarily",
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r5-prefer-scoped-objects-dont-heap-allocate-unnecessarily

Somewhat similar to pull request #3564 "STYLE: Declare Ellipsoid SpatialFunction orientations as a fixed matrix"

### DIFF
--- a/Modules/Core/ImageFunction/include/itkWindowedSincInterpolateImageFunction.h
+++ b/Modules/Core/ImageFunction/include/itkWindowedSincInterpolateImageFunction.h
@@ -21,6 +21,7 @@
 #include "itkConstNeighborhoodIterator.h"
 #include "itkZeroFluxNeumannBoundaryCondition.h"
 #include "itkInterpolateImageFunction.h"
+#include "itkMath.h"
 
 namespace itk
 {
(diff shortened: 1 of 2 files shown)


dzenanz (approved):

hjmjohnson (approved):

N-Dekker:
Thanks for your approval, @dzenanz, @hjmjohnson, and the 👍, @blowekamp ! I believe this pull request might improve the performance of `WindowedSincInterpolateImageFunction`, but I did not do any benchmark. My main aim is to avoid doing the memory allocation "manually", in cases where it can done "automatically" rather easily.

PR #3577, opened 2022-08-25
STYLE: Use unique_ptr to avoid the same delete[] on multiple exit paths

N-Dekker:
Did use `unique_ptr<T[]>` to cleanup code that had the very same `delete[]` on
multiple function exit paths, for example:

    catch (...)
    {
      delete[] buffer;
      throw;
    }
    delete[] buffer;

In this example, the catch clause can be removed entirely, when declaring the
buffer as a `unique_ptr<T[]>`.

---
Tip: When reviewing, you may want to hide whitespace changes: https://github.com/InsightSoftwareConsortium/ITK/pull/3577/files?w=1

### DIFF
--- a/Modules/Filtering/ImageGrid/include/itkBinShrinkImageFilter.hxx
+++ b/Modules/Filtering/ImageGrid/include/itkBinShrinkImageFilter.hxx
@@ -20,6 +20,8 @@
 
 #include "itkImageScanlineIterator.h"
 #include "itkTotalProgressReporter.h"
+
+#include <memory> // For unique_ptr.
 #include <numeric>
 #include <functional>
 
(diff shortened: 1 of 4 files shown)


N-Dekker:
@issakomi Hi Mihail! This pull request includes a small adjustment to your #3548 commit https://github.com/InsightSoftwareConsortium/ITK/commit/c5bc6a65409666039c4b56a95cde3b7f869afe74 "BUG: fixed memory leaks in MINC IO". Specifically, it proposes to remove the `try` and `catch` from the following code, as it proposes to declare both `start` and `count` as a `unique_ptr` :

```
  try
  {
    this->WriteImageInformation();
  }
  catch (const itk::ExceptionObject &)
  {
    delete[] start;
    delete[] count;
    throw;
  }

```
https://github.com/InsightSoftwareConsortium/ITK/blob/c5bc6a65409666039c4b56a95cde3b7f869afe74/Modules/IO/MINC/src/itkMINCImageIO.cxx#L1445-L1455

Please check!

dzenanz (approved):

issakomi:
Thank you!

PR #3590, opened 2022-08-30
Add `make_unique_for_overwrite` from C++20, and replace `new T[n]` with `make_unique_for_overwrite<T[]>(n)` calls

N-Dekker:
Replaced `new T[n]` with `make_unique_for_overwrite<T[]>(n)` in code that basically looks like: 

    auto * data = new T[n];
    Do things...
    delete [] data;

Doing so fixes some potential memory leaks, when the things being done before the `delete [] data` might throw an exception.

`make_unique_for_overwrite` is introduced with C++20: https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique It is hereby mimicked for dynamic array allocation as `itk::make_unique_for_overwrite`

### DIFF
--- a/Modules/Core/Common/include/itkMakeUniqueForOverwrite.h
+++ b/Modules/Core/Common/include/itkMakeUniqueForOverwrite.h
@@ -1,0 +1,81 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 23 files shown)


dzenanz (approved):

hjmjohnson:
@N-Dekker Very nice.  Thanks.

thewtex (approved):
:sunglasses:

blowekamp:
Great work to remove these manually manage function scoped resources. 👍

N-Dekker:
For the record, I have submitted a related pull request to the C++ Core Guidelines: https://github.com/isocpp/CppCoreGuidelines/pull/1969: _R.5 (scoped objects): Do not warn on a const `unique_ptr<T[]>`_

PR #3599, opened 2022-09-05
Attribute morphology consolidation

richardbeare:
STYLE: Documentation and removal of code hidden by macros

This PR consolidates documentation between parent and child classes - previously the child
classes had copies of the parent documentation.

The PAMI macro has been removed. Originally there were two versions
of the code from consecutive publications of the algorithm. The
non PAMI code was incorrect and retained for historical purposes only.

PERF: Used stl::stable_sort instead of standard sort of a class containing
both offset and pixel intensity. This slightly reduces RAM use.

Changes suggested as a result of:

@zivy SimpleITK/SimpleITK#1656

## PR Checklist
- [ ] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [ ] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [ ] Updated API documentation (or API not changed)

PR #3601, opened 2022-09-07
STYLE: Use unique_ptr and make_unique_for_overwrite in Review module

N-Dekker:
Suggested by Bradley Lowekamp (@blowekamp) at https://github.com/InsightSoftwareConsortium/ITK/pull/3599#issuecomment-1238166774

Follow-up to pull request https://github.com/InsightSoftwareConsortium/ITK/pull/3590 commit 15d09d6c756b46b4d1c296d8bb13b85f82654ef9
"STYLE: Replace `new T[n]` with `make_unique_for_overwrite` in cxx files"

### DIFF
--- a/Modules/Nonunit/Review/include/itkAttributeMorphologyBaseImageFilter.hxx
+++ b/Modules/Nonunit/Review/include/itkAttributeMorphologyBaseImageFilter.hxx
@@ -24,6 +24,7 @@
 #include "itkConnectedComponentAlgorithm.h"
 #include "itkNeighborhoodAlgorithm.h"
 #include "itkCastImageFilter.h"
+#include "itkMakeUniqueForOverwrite.h"
 
 /*
  * This code was contributed in the Insight Journal paper
@@ -97,12 +98,12 @@ AttributeMorphologyBaseImageFilter<TInputImage, TOutputImage, TAttribute, TFunct
 
   fit = faceList.begin();
 
-  m_SortPixels = new OffsetValueType[buffsize];
-  m_Parent = new OffsetValueType[buffsize];
+  m_SortPixels = make_unique_for_overwrite<OffsetValueType[]>(buffsize);
+  m_Parent = make_unique_for_overwrite<OffsetValueType[]>(buffsize);
 
   // This is a bit ugly, but I can't see an easy way around
-  m_Raw = new InputPixelType[buffsize];
-  m_AuxData = new AttributeType[buffsize];
+  m_Raw = make_unique_for_overwrite<InputPixelType[]>(buffsize);
+  m_AuxData = make_unique_for_overwrite<AttributeType[]>(buffsize);
 
   // copy the pixels to the sort buffer
   using CRegionIteratorType = ImageRegionConstIteratorWithIndex<TInputImage>;
@@ -120,7 +121,7 @@ AttributeMorphologyBaseImageFilter<TInputImage, TOutputImage, TAttribute, TFunct
     progress.CompletedPixel();
   }
   progress.CompletedPixel();
-  m_CompareOffset.buf = m_Raw;
+  m_CompareOffset.buf = m_Raw.get();
   std::stable_sort(&(m_SortPixels[0]), &(m_SortPixels[buffsize - 1]), m_CompareOffset);
   progress.CompletedPixel();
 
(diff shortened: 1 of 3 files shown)


Modules/Nonunit/Review/include/itkAttributeMorphologyBaseImageFilter.hxx
```diff
@@ -198,10 +199,10 @@ AttributeMorphologyBaseImageFilter<TInputImage, TOutputImage, TAttribute, TFunct
...
   }
 
-  delete[] m_Raw;
-  delete[] m_SortPixels;
-  delete[] m_Parent;
-  delete[] m_AuxData;
+  m_Raw.reset();
+  m_SortPixels.reset();
+  m_Parent.reset();
+  m_AuxData.reset();
```
N-Dekker, line 205: Note that these four `reset()` calls would not have been necessary if m_Raw, m_SortPixels, m_Parent, and m_AuxData would have been local variables, declared within the member function itself.

dzenanz (approved):

Modules/Nonunit/Review/include/itkAttributeMorphologyBaseImageFilter.hxx
hjmjohnson: @N-Dekker I think that the better way to handle this problem is to convert the instance variables into local variables for the GenerateData method.

It may require also converting some of the classes' private member functions into lambda functions inside of GenerateData.  (i.e. Criterion, Union )

NOTE:  Your simple .reset solution is an improvement, so I'd be happy to add it as an evolutionary step forward.

N-Dekker: > I think that the better way to handle this problem is to convert the instance variables into local variables for the GenerateData method.

@hjmjohnson I agree, but converting them into local variables doesn't appear entirely trivial. Some private member functions of AttributeMorphologyBaseImageFilter depend on them being  instance variables. While I believe those private member functions could be adjusted (probably converted to lambda's), I feel reluctant to do so, as I'm not familiar with this filter, and I'm not using it either.

Leengit: For that next pull request, can we get a member`.reset()` call invoked as part of the desctructor of a local variable of this `GenerateData` method?  That way, we wouldn't have to worry about what path is taken in exiting this method.

N-Dekker: > That way, we wouldn't have to worry about what path is taken in exiting this method.

@Leengit Thanks for your feedback. Honestly, I don't worry too much, because if the taken path does _not_ reach the four proposed `reset()` calls (typically because of an exception somewhere inside GenerateData), the memory is still owned by those four unique_ptrs (m_Raw, m_SortPixels, m_Parent, and m_AuxData), so it won't leak. The memory will then be freed by the next GenerateData call on that filter, or by the destructor of the filter.

Leengit: Agreed that the current pull request is a significant step in the right direction and is plenty good as is.

Giving some additional information for my idea in case we do revisit this ... I looked for a standard C++ template that would accept an arbitrary method rather than hardcoding `reset`, but I couldn't find one.  That we might have to code things ourselves makes things uglier; we might have to create something like
```c++
template <typename T> class ResetOnDestruction {
private:
  T &m_resetable;

public:
  ResetOnDestruction(ResetOnDestruction &) = delete;
  ResetOnDestruction(T &resetable) : m_resetable(resetable) {}
  ~ResetOnDestruction() { m_resetable.reset(); }
};
```
... which would then be used as in:
```c++
ResetOnDesctruction<decltype(m_SortPixels)> gc_SortPixels(m_SortPixels);
m_SortPixels = make_unique_for_overwrite<OffsetValueType[]>(buffsize);
```
... so that `m_SortPixels.reset()` is invoked when `gc_SortPixels` goes out of scope.

Leengit: > freed by the next GenerateData call on that filter

Indeed, much less terrible than a typical leak!  Still "if everyone did this", we could have a bunch of ITK objects holding onto extra memory they don't need at the moment.  This is not the highest priority in the overall scheme of things, but I'd venture that it's worth fixing if it catches someone's interest.

N-Dekker: I understand your point, @Leengit but I still think that the _ideal_ solution would be to declare those four unique_ptrs as local variables, instead of as member variables. Which would automatically cleanup the memory when exiting GenerateData, on each possible path.

N-Dekker: > I looked for a standard C++ template that would accept an arbitrary method rather than hardcoding reset, but I couldn't find one.

Could it be that you were looking for `unique_resource`?  https://en.cppreference.com/w/cpp/experimental/unique_resource

hjmjohnson (approved):

PR #3602, opened 2022-09-07
Each `UpdateOutputInformation()` member function should call `GetSource()` just once

N-Dekker:
When calling `GetSource()` twice on the same data object, debug output may appear duplicated, and it may cause unnecessary reference counting (both up and down) of the source object.

For clarification, the debug output may come from the `itkDebugMacro` call inside `DataObject::GetSource()`, and the reference counting comes from converting m_Source to a `SmartPointer`, here:
https://github.com/InsightSoftwareConsortium/ITK/blob/c1839700a9d28946d1313caf91795e4ce160eb13/Modules/Core/Common/src/itkDataObject.cxx#L247-L252

### DIFF
--- a/Modules/Core/Common/include/itkImageBase.hxx
+++ b/Modules/Core/Common/include/itkImageBase.hxx
@@ -213,9 +213,11 @@ template <unsigned int VImageDimension>
 void
 ImageBase<VImageDimension>::UpdateOutputInformation()
 {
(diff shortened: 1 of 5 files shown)


dzenanz (approved):

blowekamp:
LGTM

hjmjohnson (approved):

PR #3603, opened 2022-09-08
Add `DataObject::UpdateSource()`, replacing `GetSource()->Update()` calls

N-Dekker:
`DataObject::UpdateSource()` offers a more efficient, more concise alternative to the following common pattern:

    if (x->GetSource())
    {
      x->GetSource()->Update();
    }

Calling `DataObject::UpdateSource()` instead is more efficient, because it internally calls `DataObject::GetSource()` just once, instead of twice.

Replaced all (15) occurrences of the pattern with  `x->UpdateSource()` calls.

Follow-up to pull request #3602 "Each `UpdateOutputInformation()` member function should call `GetSource()` just once"

### DIFF
--- a/Modules/Core/Common/include/itkDataObject.h
+++ b/Modules/Core/Common/include/itkDataObject.h
@@ -333,6 +333,10 @@ public:
   DataObjectPointerArraySizeType
   GetSourceOutputIndex() const;
 
+  /** Does an Update() of its source, if it has one. Does nothing, otherwise. */
+  void
+  UpdateSource() const;
+
   /** Restore the data object to its initial state. This means releasing
    * memory. */
   virtual void
--- a/Modules/Core/Common/src/itkDataObject.cxx
+++ b/Modules/Core/Common/src/itkDataObject.cxx
@@ -268,6 +268,17 @@ DataObject::GetSourceOutputIndex() const
   return m_Source->MakeIndexFromOutputName(m_SourceOutputName);
 }
 
+void
+DataObject::UpdateSource() const
+{
+  const auto source = this->GetSource();
+
+  if (source)
+  {
+    source->Update();
+  }
+}
+
 //----------------------------------------------------------------------------
 void
 DataObject::PrintSelf(std::ostream & os, Indent indent) const
--- a/Modules/Registration/Common/include/itkCenteredTransformInitializer.hxx
+++ b/Modules/Registration/Common/include/itkCenteredTransformInitializer.hxx
@@ -40,14 +40,8 @@ CenteredTransformInitializer<TTransform, TFixedImage, TMovingImage>::InitializeT
   }
 
   // If images come from filters, then update those filters.
-  if (m_FixedImage->GetSource())
-  {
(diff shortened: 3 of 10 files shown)


dzenanz (approved):

blowekamp:
This change requires a little more thought, and I need to look into things a little closer.

The classes modified, and non-pipelined algorithms and don't follow the filter conventions...

N-Dekker:
> This change requires a little more thought, and I need to look into things a little closer.
> 
> The classes modified, and non-pipelined algorithms and don't follow the filter conventions...

Thanks for your feedback @blowekamp but do you mean that the following pattern, that currently occurs 15 times in ITK, does not follow the filter conventions?

    if (x->GetSource())
    {
      x->GetSource()->Update();
    }

blowekamp:
The `CenterTransformInitializer` object is not derived from the `ProcessingObject` base class. The class is not a filter, it can not be used as part of a pipeline. The "Initializer" algorithm has input, and tries to "update" those inputs even though it is not a filter or a pipeline object. If the algorithm developer is propagating execution to the input's of the algorithm, then  the class should be derived from the ProcessObject for proper execution so that the conventions of a filter are followed.

I suspect the other classes modified are not derived from ProcessObject either.

Adding another method to the DataObject, that is not suppose to be used with the conical ITK pipelines may contribute to confusion.

Typically "algorithm" classes operate in immediate mode and don't propagate execution up their inputs.

N-Dekker:
> Adding another method to the DataObject, that is not suppose to be used with the conical ITK pipelines may contribute to confusion.

Thanks for your explanation @blowekamp. Would it help to add a note to the documentation (doxygen) of the proposed `DataObject::UpdateSource()`, saying that its typical use cases are _outside_ the canonical ITK pipelines?

Otherwise, how would you suggest to fix those 15 `if (x->GetSource()) x->GetSource()->Update()` cases addressed by https://github.com/InsightSoftwareConsortium/ITK/pull/3603/commits/7a3af8102be943a9d0a94633e6fb527de4bcbb13 ?

N-Dekker:
I suspect that the pattern `if (x->GetSource()) x->GetSource()->Update()` is not thread-safe. Because the DataObject `x` does not "own" its source. It only has a _WeakPointer_ to the source:

https://github.com/InsightSoftwareConsortium/ITK/blob/c1839700a9d28946d1313caf91795e4ce160eb13/Modules/Core/Common/include/itkDataObject.h#L555

`x->GetSource()` returns a _SmartPointer_, which takes care of keeping the source "alive". However, it only does so until it goes out of scope. Or until the end of the statement, in this case. In between the two `x->GetSource()` statements, I guess the source may be removed by another thread. Which would break the `x->GetSource()->Update()` call. That's also why I would suggest to place the SmartPointer returned by x->GetSource() in a local variable that remains in scope when calling source->Update():


    const auto source = x->GetSource();

    if (source)
    {
      source->Update();
    }

hjmjohnson (approved):

PR #3613, opened 2022-09-12
STYLE: Declare `RGBGibbsPriorFilter` data members as `unique_ptr<T[]>`

N-Dekker:
Declared m_LabelStatus, m_Region, and m_RegionCount as `unique_ptr<T[]>`, which improved exception-safety. Defaulted the destructor of `RGBGibbsPriorFilter`.

### DIFF
--- a/Modules/Segmentation/MarkovRandomFieldsClassifiers/include/itkRGBGibbsPriorFilter.hxx
+++ b/Modules/Segmentation/MarkovRandomFieldsClassifiers/include/itkRGBGibbsPriorFilter.hxx
@@ -43,14 +43,6 @@ RGBGibbsPriorFilter<TInputImage, TClassifiedImage>::RGBGibbsPriorFilter()
   m_StartPoint.Fill(0);
 }
 
-template <typename TInputImage, typename TClassifiedImage>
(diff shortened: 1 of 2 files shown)


Modules/Segmentation/MarkovRandomFieldsClassifiers/include/itkRGBGibbsPriorFilter.hxx
```diff
@@ -623,11 +615,8 @@ RGBGibbsPriorFilter<TInputImage, TClassifiedImage>::RegionEraser()
 {
   const unsigned int size = m_ImageWidth * m_ImageHeight * m_ImageDepth;
 
-  delete[] m_Region;
-  m_Region = new unsigned short[size];
-
-  delete[] m_RegionCount;
-  m_RegionCount = new unsigned short[size];
```
N-Dekker, line 630: These original lines of code don't look exception-safe, because if either of the two `new` statements would fail (by throwing an `std::bad_alloc` exception), one of the raw pointers `m_Region` or `m_RegionCount` would remain "dangling", i.e., pointing to deallocated memory. The original `~RGBGibbsPriorFilter()` would then try to deallocated memory that was already deallocated, causing undefined behavior (typically crashing).

dzenanz (approved):

PR #3614, opened 2022-09-12
STYLE: Declare `ImageGaussianModelEstimator::m_Covariance` as unique_ptr

N-Dekker:
Also defaulted the destructor of `ImageGaussianModelEstimator`.


dzenanz (approved):
Code looks good, check/fix CI.

PR #3616, opened 2022-09-14
 Declare metric "per thread" data members as `unique_ptr<T[]>`, remove `mutable`

N-Dekker:
Declared registration metric "per thread" data members as `unique_ptr<T[]>` (similar to pull request #3613 and #3614), and removed `mutable` keywords from their declarations.

### DIFF
--- a/Modules/Registration/Common/include/itkMattesMutualInformationImageToImageMetric.h
+++ b/Modules/Registration/Common/include/itkMattesMutualInformationImageToImageMetric.h
@@ -360,7 +360,7 @@ private:
   // See
   //   https://thetweaker.wordpress.com/2010/05/05/stdvector-of-aligned-elements/
   //   https://connect.microsoft.com/VisualStudio/feedback/details/692988
-  mutable AlignedMMIMetricPerThreadStruct * m_MMIMetricPerThreadVariables;
+  AlignedMMIMetricPerThreadStruct * m_MMIMetricPerThreadVariables;
 #endif
 
   bool         m_UseExplicitPDFDerivatives{ true };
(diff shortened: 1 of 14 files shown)


hjmjohnson (approved):

ebrahimebrahim (approved):
my review should not mean much here as I'm no longer very familiar with the code, but it's always good to see raw pointers go away :+1: . IDK the original intention behind making those data members mutable

N-Dekker:
@ebrahimebrahim Thanks for having a look at this pull request, and for your approval! 😃 

>  IDK the original intention behind making those data members mutable

I don't know either, but I think it might be a misinterpretation of the meaning of `mutable` for a pointer: in this case (for example, `mutable AlignedMMIMetricPerThreadStruct * m_MMIMetricPerThreadVariables;`), the keyword applies to the pointer itself, _not_ to the value(s) that it points to. So `mutable` would only be necessary when the pointer itself would need to be assigned or reset in a `const` member function. On the other hand, the values that it points to can still be modified by both _const_ and non-const member functions, _even_ when the `mutable` keyword is not there!!! 

https://github.com/InsightSoftwareConsortium/ITK/blob/b03347bcdc4c2fb1d47f1fbcb1f7bc48c3f59089/Modules/Registration/Common/include/itkMattesMutualInformationImageToImageMetric.h#L363

ebrahimebrahim:
> the keyword applies to the pointer itself, _not_ to the value(s) that it points to. So `mutable` would only be necessary when the pointer itself would need to be assigned or reset in a `const` member function

Ah I was wondering if that's how it worked ... makes sense that it was a misinterpretation

PR #3619, opened 2022-09-15
Remove unreachable code from GE5ImageIO and use unique_ptr

N-Dekker:
`GE5ImageIO::ReadHeader` appeared to have unreachable code in `if`-clauses like this:

    buffer = new char[N];
    if (buffer == nullptr)
    {
      f.close();
      itkExceptionMacro("GE5ImageIO:Unable to allocate memory...!");
    }

This pull request proposes to remove those unreachable lines of code, and replace locally declared raw pointers by `unique_ptr`, in GE5ImageIO.

### DIFF
--- a/Modules/IO/GE/src/itkGE5ImageIO.cxx
+++ b/Modules/IO/GE/src/itkGE5ImageIO.cxx
@@ -255,22 +255,12 @@ GE5ImageIO::ReadHeader(const char * FileNameToRead)
   if (pixelHdrFlag)
   {
     buffer = new char[imageHdr.GENESIS_IH_img_l_exam];
-    if (buffer == nullptr)
-    {
-      f.close();
-      itkExceptionMacro("GE5ImageIO:Unable to allocate memory for exam header!");
-    }
     f.seekg(imageHdr.GENESIS_IH_img_p_exam, std::ios::beg);
     f.read(buffer, imageHdr.GENESIS_IH_img_l_exam);
   }
   else
   {
     buffer = new char[GENESIS_EX_HDR_LEN];
-    if (buffer == nullptr)
-    {
-      f.close();
-      itkExceptionMacro("GE5ImageIO:Unable to allocate memory for exam header!");
-    }
     f.seekg(GENESIS_EX_HDR_START, std::ios::beg);
     f.read(buffer, GENESIS_EX_HDR_LEN);
   }
@@ -314,22 +304,12 @@ GE5ImageIO::ReadHeader(const char * FileNameToRead)
   if (pixelHdrFlag)
   {
     buffer = new char[imageHdr.GENESIS_IH_img_l_series];
-    if (buffer == nullptr)
-    {
-      f.close();
-      itkExceptionMacro("GE5ImageIO:Unable to allocate memory for series header!");
-    }
     f.seekg(imageHdr.GENESIS_IH_img_p_series, std::ios::beg);
     f.read(buffer, imageHdr.GENESIS_IH_img_l_series);
   }
   else
   {
     buffer = new char[GENESIS_SE_HDR_LEN];
-    if (buffer == nullptr)
-    {
-      f.close();
-      itkExceptionMacro("GE5ImageIO:Unable to allocate memory for series header!");
-    }
     f.seekg(GENESIS_SE_HDR_START);
     f.read(buffer, GENESIS_SE_HDR_LEN);
   }
@@ -351,11 +331,6 @@ GE5ImageIO::ReadHeader(const char * FileNameToRead)
   if (pixelHdrFlag)
   {
     buffer = new char[imageHdr.GENESIS_IH_img_l_image];
-    if (buffer == nullptr)
-    {
-      f.close();
-      itkExceptionMacro("GE5ImageIO:Unable to allocate memory for MR header!");
-    }
     // Now seek to the MR header and read the data into the buffer.
     f.seekg(imageHdr.GENESIS_IH_img_p_image, std::ios::beg);
     f.read(buffer, imageHdr.GENESIS_IH_img_l_image);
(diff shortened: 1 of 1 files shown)


dzenanz (approved):

jhlegarreta:
@N-Dekker Thanks for all these efforts to avoid the use raw pointers.

In this case, the rationale and/or commit message looks to be inconsistent with the coverage report:
https://open.cdash.org/viewCoverageFile.php?buildid=8157810&fileid=52192360

The "unreachable code" is actually being reached.

Any ways to explain why that is happening if the code should be unreachable?

N-Dekker:
> In this case, the rationale and/or commit message looks to be inconsistent with the coverage report: https://open.cdash.org/viewCoverageFile.php?buildid=8157810&fileid=52192360
> 
> The "unreachable code" is actually being reached.
> 
> Any ways to explain why that is happening if the code should be unreachable?

Very interesting @jhlegarreta ! Those lines that say `itkExceptionMacro("GE5ImageIO:Unable to allocate memory`... can't _possibly_ be reached, because `new char[N]` never ever returns null, by definition! Could it be a bug in the Coverage analysis?

N-Dekker:
From a very old (but pretty famous) C++FAQ, by Marshall Cline:

> **Do I need to check for NULL after p = new Fred()?** 
> _No!_ (...)

https://www.cs.technion.ac.il/users/yechiel/c++-faq/new-never-returns-null.html

issakomi:
Just as a joke, not applicable.

> https://www.cs.technion.ac.il/users/yechiel/c++-faq/new-never-returns-null.html

```
Fred* p = new Fred();
if (p == NULL)    ← only needed if your compiler is from the Stone Age!
  throw std::bad_alloc();
```

Hmm, we don't know here what is class Fred,  may be e.g.:

```cpp
#include <iostream>

class Fred
{
public:
  void * operator new(size_t) /* throw() */
  {
    return nullptr;
  }
  void operator delete(void*)
  {
  }
};

int main(int, char**)
{
  Fred * f = new Fred;
  if (f == nullptr)
  {
    std::cout << "Fred is nullptr" << std::endl;
  }
  delete f;
  return 0;
}
```


```
r@deb2:~$ ./a.out 
Fred is nullptr
```

Works for me with and without `throw()`, without GCC 12.1.0 shows warning.

N-Dekker:
> Hmm, we don't know here what is class Fred, may be e.g.: [...]

@issakomi Very clever 👍, Mihail! Thanks! Now I just hope we can keep your `Fred` out of the ITK source tree  😸 

> Just FYI, related to bulk replacements with _make_unique_for_overwrite_ in general, not for the current PR.

Indeed, not relevant for this specific PR, but `itk::make_unique_for_overwrite` _does_ take user-defined `operator new` overloads into account, as I just checked:

    class Flintstone
    {
    public:
      void *
      operator new[](size_t)
      { return nullptr;
      }
      void
      operator delete[](void *)
      {}
    };


    TEST(MakeUniqueForOverwrite, UniqueFlintstoneIsNull)
    {
      const auto uniqueFlintstone = itk::make_unique_for_overwrite<Flintstone[]>(42);
      ASSERT_EQ(uniqueFlintstone, nullptr); // Passes successfully (no assert failure).
    }

So I think it works as expected. Right?

jhlegarreta:
>  Could it be a bug in the Coverage analysis?

IDK. I have observed at times coverage reports that seemed misleading (e.g. not reporting lines as covered when the test contained calls to them, and the file was contained in the `CMakeLists.txt` file).

In this case, a way to see if those lines are actually hit is to inspect the output of the corresponding test in the dashboard.

Also, I remember reading in other PRs compilers throwing warnings about unreachable code (maybe GCC 12.1.0 as @issakomi mentions?).

issakomi:
Тhe class Fred with a joke of course ... But it is good that scenario with overloaded new was checked, better check twice.

Screenshot 2022-09-16 at 00-43-33 CDash - View Coverage File
Maybe really a bug, strange that line `f.close();` is not marked as covered, how could it happen. Sorry, i don't know much about Coverage tool.

IMHO, i don't think compiler will warn about dead code, i have discovered several such bugs in my code with Synopsis Coverity analyzer, compiler said nothing.

```cpp
int main(int, char**)
{
	bool b = true;
	if (!b)
	{
		int a = 1;
		++a;
	}
}
```

```
r@deb2:~$ g++ test.cpp 
r@deb2:~$ g++ -Wall test.cpp
r@deb2:~$ g++ -O2 -Wall test.cpp 
r@deb2:~$ 
```
No warning.

N-Dekker:
> No warning.

Interesting... for what it's worth, the following does at least produce a warning from MSVC, saying:

> <source>(10): warning C4127: conditional expression is constant

```
int main()
{
	const bool b = true;
	int a = 1;
	if (b == false)
	{
		++a;
	}
	return a;
}

```

At https://godbolt.org/z/E69ea4Yer

N-Dekker:
For the record, the original `if ( buffer == NULL )` checks were introduced with commit https://github.com/InsightSoftwareConsortium/ITK/commit/fea55fac7a370e25d9cd246ba672657fdef1c976 in 2010. The very first edition of the C++ Standard (1998) already specified that expressions like `new char[N]` "only indicate failure by throwing
an exception", not by returning a null pointer.

Anyway, thanks for your comments @issakomi, @jhlegarreta and thanks for your approval @dzenanz I think this pull request is ready  😃

issakomi:
```cpp
        int* p = new(std::nothrow) int[100000000ul]; // non-throwing overload
        if (p == nullptr)
```
another alternative, but just for completeness, not applicable for the PR 
https://en.cppreference.com/w/cpp/memory/new/nothrow

hjmjohnson (approved):

N-Dekker:
Thanks for merging @hjmjohnson Of course, it would have been nice if any static analysis tool would have produced a warning on unreachable code! Moreover, I find it really misleading that ubuntu2004 Coverage 8 claimed that those unreachable lines of code were actually covered. At https://open.cdash.org/viewCoverageFile.php?buildid=8157810&fileid=52192360 as remarked by @jhlegarreta. Hope the Coverage tool will still be fixed!

PR #3620, opened 2022-09-15
Use `unique_ptr` internally in the implementation of `itk::Object`

N-Dekker:
- Declared the observers of `itk::Object` as list of `Observer` objects (instead of a list of raw pointers).
- Declared both `m_SubjectImplementation` and `m_MetaDataDictionary` as an `std::unique_ptr`.
- Declared `Observer::m_Event` as unique_ptr

### DIFF
--- a/Modules/Core/Common/src/itkObject.cxx
+++ b/Modules/Core/Common/src/itkObject.cxx
@@ -57,7 +57,7 @@ class ITKCommon_HIDDEN SubjectImplementation
 {
 public:
   SubjectImplementation() { m_Count = 0; }
-  ~SubjectImplementation();
+  ~SubjectImplementation() = default;
 
   unsigned long
   AddObserver(const EventObject & event, Command * cmd);
@@ -106,29 +106,20 @@ protected:
 
   template <typename TObject>
   void
-  InvokeEventRecursion(const EventObject & event, TObject * self, std::list<Observer *>::reverse_iterator & i);
+  InvokeEventRecursion(const EventObject & event, TObject * self, std::list<Observer>::reverse_iterator & i);
 
 private:
-  std::list<Observer *> m_Observers;
-  unsigned long         m_Count;
+  std::list<Observer> m_Observers;
+  unsigned long       m_Count;
 };
-
-SubjectImplementation::~SubjectImplementation()
-{
-  for (auto & observer : m_Observers)
-  {
-    delete observer;
-  }
-}
 
 unsigned long
 SubjectImplementation::AddObserver(const EventObject & event, Command * cmd)
 {
-  auto * ptr = new Observer(cmd, event.MakeObject(), m_Count);
-
-  m_Observers.push_back(ptr);
+  const unsigned long tag{ m_Count };
+  m_Observers.emplace_back(cmd, event.MakeObject(), tag);
   ++m_Count;
-  return ptr->m_Tag;
+  return tag;
 }
 
 void
@@ -136,9 +127,8 @@ SubjectImplementation::RemoveObserver(unsigned long tag)
 {
   for (auto i = m_Observers.begin(); i != m_Observers.end(); ++i)
   {
-    if ((*i)->m_Tag == tag)
+    if (i->m_Tag == tag)
     {
-      delete (*i);
       m_Observers.erase(i);
       m_ListModified = true;
       return;
@@ -149,10 +139,6 @@ SubjectImplementation::RemoveObserver(unsigned long tag)
 void
 SubjectImplementation::RemoveAllObservers()
 {
-  for (auto & observer : m_Observers)
-  {
-    delete observer;
-  }
   m_Observers.clear();
   m_ListModified = true;
 }
@@ -183,9 +169,9 @@ SubjectImplementation::InvokeEvent(const EventObject & event, const Object * sel
 
 template <typename TObject>
 void
-SubjectImplementation::InvokeEventRecursion(const EventObject &                       event,
-                                            TObject *                                 self,
-                                            std::list<Observer *>::reverse_iterator & i)
+SubjectImplementation::InvokeEventRecursion(const EventObject &                     event,
+                                            TObject *                               self,
+                                            std::list<Observer>::reverse_iterator & i)
 {
   // This method recursively visits the list of observers in reverse
   // order so that on the last recursion the first observer can be
@@ -200,20 +186,20 @@ SubjectImplementation::InvokeEventRecursion(const EventObject &                 
   {
 
     // save observer
-    const Observer * o = *i;
+    const Observer & o = *i;
 
     // Save its tag, before the observer could /possibly/ be removed.
-    const unsigned long tag{ o->m_Tag };
+    const unsigned long tag{ o.m_Tag };
 
-    if (o->m_Event->CheckEvent(&event))
+    if (o.m_Event->CheckEvent(&event))
     {
       InvokeEventRecursion(event, self, ++i);
 
-      const auto hasSameTag = [tag](const Observer * const observer) { return observer->m_Tag == tag; };
+      const auto hasSameTag = [tag](const Observer & observer) { return observer.m_Tag == tag; };
 
       if (!m_ListModified || std::any_of(m_Observers.begin(), m_Observers.end(), hasSameTag))
       {
-        o->m_Command->Execute(self, event);
+        o.m_Command->Execute(self, event);
       }
 
       return;
@@ -230,9 +216,9 @@ SubjectImplementation::GetCommand(unsigned long tag)
 {
   for (auto & observer : m_Observers)
   {
-    if (observer->m_Tag == tag)
+    if (observer.m_Tag == tag)
     {
-      return observer->m_Command;
+      return observer.m_Command;
     }
   }
   return nullptr;
@@ -241,9 +227,9 @@ SubjectImplementation::GetCommand(unsigned long tag)
 bool
 SubjectImplementation::HasObserver(const EventObject & event) const
 {
-  for (auto observer : m_Observers)
+  for (const auto & observer : m_Observers)
   {
-    const EventObject * e = observer->m_Event.get();
+    const EventObject * e = observer.m_Event.get();
     if (e->CheckEvent(&event))
     {
       return true;
@@ -260,10 +246,10 @@ SubjectImplementation::PrintObservers(std::ostream & os, Indent indent) const
     return false;
   }
 
-  for (auto observer : m_Observers)
+  for (const auto & observer : m_Observers)
   {
-    const EventObject * e = observer->m_Event.get();
-    const Command *     c = observer->m_Command;
+    const EventObject * e = observer.m_Event.get();
+    const Command *     c = observer.m_Command;
     os << indent << e->GetEventName() << "(" << c->GetNameOfClass();
     if (!c->GetObjectName().empty())
     {
--- a/Modules/Core/Common/src/itkObject.cxx
+++ b/Modules/Core/Common/src/itkObject.cxx
@@ -27,6 +27,7 @@
  *=========================================================================*/
 #include "itkCommand.h"
 #include <algorithm>
+#include <memory> // For unique_ptr.
 
 #include "itkSingleton.h"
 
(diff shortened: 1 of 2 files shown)


N-Dekker:
Any idea what the ITK.macOS.Python failure is about? https://open.cdash.org/viewBuildError.php?buildid=8158841 says:
> CMake Error at KWStyle-prefix/src/KWStyle-stamp/KWStyle-download-Release.cmake:49 (message):
>
> Command failed: 1
>
>  '/usr/local/Cellar/cmake/3.24.1/bin/cmake' '-P' '/.../s-build/KWStyle-prefix/tmp/KWStyle-gitclone.cmake'
>
> See also
>
>   /.../s-build/KWStyle-prefix/src/KWStyle-stamp/KWStyle-download-*.log

dzenanz:
Looking at Azure log:
```log
CMake Deprecation Warning at /Users/runner/work/1/ITK-dashboard/itk_common.cmake:96 (cmake_minimum_required):
  CMake.

  /Users/runner/work/1/ITK-dashboard/azure_dashboard.cmake:82 (include)
  /Users/runner/work/1/ITK-dashboard/dashboard.cmake:12 (include)
```
and then later on
```log
[527/5723] Building C object Modules/ThirdParty/VNL/src/vxl/v3p/netlib/CMakeFiles/itkv3p_netlib.dir/lapack/complex16/zlascl.c.o
[528/5723] Performing download step (git clone) for 'KWStyle'
FAILED: KWStyle-prefix/src/KWStyle-stamp/KWStyle-download /Users/runner/work/1/s-build/KWStyle-prefix/src/KWStyle-stamp/KWStyle-download 
cd /Users/runner/work/1/s-build && /usr/local/Cellar/cmake/3.24.1/bin/cmake -P /Users/runner/work/1/s-build/KWStyle-prefix/src/KWStyle-stamp/KWStyle-download-Release.cmake && /usr/local/Cellar/cmake/3.24.1/bin/cmake -E touch /Users/runner/work/1/s-build/KWStyle-prefix/src/KWStyle-stamp/KWStyle-download
CMake Error at KWStyle-prefix/src/KWStyle-stamp/KWStyle-download-Release.cmake:49 (message):
  Command failed: 1

   '/usr/local/Cellar/cmake/3.24.1/bin/cmake' '-P' '/Users/runner/work/1/s-build/KWStyle-prefix/tmp/KWStyle-gitclone.cmake'

  See also

    /Users/runner/work/1/s-build/KWStyle-prefix/src/KWStyle-stamp/KWStyle-download-*.log


[529/5723] Building C object Modules/ThirdParty/VNL/src/vxl/v3p/netlib/CMakeFiles/itkv3p_netlib.dir/lapack/complex16/zung2r.c.o
[530/5723] Building C object Modules/ThirdParty/VNL/src/vxl/v3p/netlib/CMakeFiles/itkv3p_netlib.dir/lapack/complex16/ztrevc.c.o
[531/5723] Building C object Modules/ThirdParty/VNL/src/vxl/v3p/netlib/CMakeFiles/itkv3p_netlib.dir/lapack/complex16/zlatrs.c.o
[532/5723] Building C object Modules/ThirdParty/VNL/src/vxl/v3p/netlib/CMakeFiles/itkv3p_netlib.dir/lapack/complex16/zlarfx.c.o
ninja: build stopped: subcommand failed.
Command exited with the value: 1
MakeCommand:/usr/local/Cellar/cmake/3.24.1/bin/cmake --build . --config "Release"
   1 Compiler errors
   1 Compiler warnings
Error(s) when building project
SetCTestConfiguration:BuildDirectory:/Users/runner/work/1/s-build
SetCTestConfiguration:SourceDirectory:/Users/runner/work/1/s
Test project /Users/runner/work/1/s-build
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 1
        Start   1: KWStyleExamplesTest
Could not find executable /Users/runner/work/1/s-build/KWStyle-build/KWStyle
Looked in the following places:
/Users/runner/work/1/s-build/KWStyle-build/KWStyle
/Users/runner/work/1/s-build/KWStyle-build/KWStyle
/Users/runner/work/1/s-build/KWStyle-build/Release/KWStyle
/Users/runner/work/1/s-build/KWStyle-build/Release/KWStyle
Release//Users/runner/work/1/s-build/KWStyle-build/KWStyle
Release//Users/runner/work/1/s-build/KWStyle-build/KWStyle
Users/runner/work/1/s-build/KWStyle-build/KWStyle
Users/runner/work/1/s-build/KWStyle-build/KWStyle
Users/runner/work/1/s-build/KWStyle-build/Release/KWStyle
Users/runner/work/1/s-build/KWStyle-build/Release/KWStyle
Release/Users/runner/work/1/s-build/KWStyle-build/KWStyle
Release/Users/runner/work/1/s-build/KWStyle-build/KWStyle

1: Test command:  "-xml" "/Users/runner/work/1/s/Utilities/KWStyle/ITK.kws.xml" "-v" "-o" "/Users/runner/work/1/s/Utilities/KWStyle/ITKOverwrite.txt" "-D" "/Users/runner/work/1/s-build/Utilities/KWStyle/ITKExamplesFiles.txt" "-gcc"
1: Working Directory: /Users/runner/work/1/s
Unable to find executable: /Users/runner/work/1/s-build/KWStyle-build/KWStyle
  1/291 Test   #1: KWStyleExamplesTest .....................................................***Not Run   0.00 sec
test 2
        Start   2: vcl_test_algorithm
```

dzenanz:
I don't know why would git clone from GitHub keep failing. https://github.com/InsightSoftwareConsortium/ITK/blob/v5.3rc04/Utilities/KWStyle/BuildKWStyle.cmake#L25

Modules/Core/Common/src/itkObject.cxx
```diff
@@ -112,42 +113,35 @@ class ITKCommon_HIDDEN SubjectImplementation
     bool                    m_Save;
   };
 
+  using ObserverListType = std::list<std::unique_ptr<const Observer>>;
+
   template <typename TObject>
   void
-  InvokeEventRecursion(const EventObject & event, TObject * self, std::list<Observer *>::reverse_iterator & i);
+  InvokeEventRecursion(const EventObject & event, TObject * self, ObserverListType::reverse_iterator & i);
```
dzenanz, line 120: Will this change in the public interface cause backwards compatibility issues?

N-Dekker: Good question, thanks @dzenanz  but no, I don't see any backwards compatibility issue here. `InvokeEventRecursion` is a protected member function of a "hidden" class (`ITKCommon_HIDDEN`) named "SubjectImplementation", which is defined inside itkObject.cxx. So `InvokeEventRecursion` is also declared and defined in itkObject.cxx, and it isn't being called outside itkObject.cxx. So it's really just an implementation detail. As far as I can see.

Modules/Core/Common/src/itkObject.cxx
```diff
@@ -219,7 +208,7 @@ SubjectImplementation::InvokeEventRecursion(const EventObject &
   {
 
     // save observer
-    const Observer * o = *i;
+    const auto & o = *i;
```
blowekamp, line 211: I'm not sure what the type is here? Reference to the unique pointer in the list?

The point of this patter is to save the pointer to the object, in case it was deleted/modified. Since the list still has the sole ownership I don't believe this works as intended any more. Set the prose above.

blowekamp (changes requested):
I am rather uncertain if the recursive function still works as intended as the list is the sole owner.

Modules/Core/Common/src/itkObject.cxx
blowekamp: Perhaps this should be maintained as a raw pointer, and the below "find" be modified to compare raw pointer of m_Observer items to this save raw pointer?

dzenanz: Is `InvokeEventRecursion` exercised in current tests? If not, now would be a good time to cover it by test(s).

blowekamp: There was a reasonable test added for this case:  f7307e26867076074a04b956633676964d5f7ffe It may be that valgrind will be required to detect the invalid memory access.

N-Dekker: Thanks @blowekamp The test that you provided, [CommandObserverObjectTest](https://github.com/InsightSoftwareConsortium/ITK/blob/2f2b29cbbe3fe91474933e0194db8d581d5b3247/Modules/Core/Common/test/itkCommandObserverObjectTest.cxx), is very helpful. I think it covers the internal helper function `InvokeEventRecursion`  well enough 👍 And it does pass successfully on this pull request  😄  

The pull request is not a design change. It just replaces existing `new` and `delete` commands by using `unique_ptr` instead. It does not change the behavior. IMhO, the use of  `unique_ptr` expresses more clearly (than when using raw pointers) that the list (`m_Observers`) has the sole ownership of the internal Observer objects.

Regarding the `std::find` call: https://github.com/InsightSoftwareConsortium/ITK/blob/2f2b29cbbe3fe91474933e0194db8d581d5b3247/Modules/Core/Common/src/itkObject.cxx#L228
This `std::find` call works just with a list of `unique_ptr`s  as with a list of raw pointers. (`std::unique_ptr` has an `operator==` which simply compares the internal raw pointers: https://en.cppreference.com/w/cpp/memory/unique_ptr/operator_cmp). 

You wrote about `const auto & o = *i`:
> I'm not sure what the type is here? Reference to the unique pointer in the list?

Indeed. Would you prefer it something like  `const ObserverPointerType & o = *i`? I still think this pull request ready to be merged right now. But if you prefer something like  `const ObserverPointerType & o = *i`, I can change that, of course. What do you think?

N-Dekker: > Perhaps this should be maintained as a raw pointer, and the below "find" be modified to compare raw pointer of m_Observer items to this save raw pointer?

Thanks @blowekamp This comment of yours inspired me to submit pull request #3629: "BUG: `Object::InvokeEvent` should not use a dangling Observer pointer"! Please check!

I think this PR (#3620) and PR #3629 can be processed in either order, but it's probably easiest now to first process PR #3629.

blowekamp (approved):
With #3629 merged this should be good to go.

N-Dekker:
> With #3629 merged this should be good to go.

Thanks @blowekamp, I'm actually preparing a revision! It appears that because we just abandoned Observer pointer comparison in favor of tag comparison, the list may now simply contain the observers themselves, rather than their pointers! Which further simplifies their memory management! 😃

blowekamp:
@N-Dekker If you are refactoring the whole thing, the recursive implementation with the search at each level is O(N**2) complexity. We should be able to use a better data structure to reduce the complexity, and improve the efficiency of the is the observer still there.  

Does the order of the observers are visited matter? Maybe.

dzenanz (approved):

N-Dekker:
> If you are refactoring the whole thing, the recursive implementation with the search at each level is O(N**2) complexity. We should be able to use a better data structure to reduce the complexity, and improve the efficiency of the is the observer still there.

Thanks @blowekamp but my focus is still on trying to avoid manual memory management (`new`/`delete`). A change from `std::list<Observer*>` to `std::list<Observer>` appears _low hanging fruit_, in that context. Switching to a different data structure sounds more challenging to me. Theoretically, other data structures may still be faster, but it really depends on the use case. How many observers does an observed object "usually" have? For a small number of observers, _O(N)_ may not be so much better than _O(1)_.

N-Dekker:
Ready to be merged.... although PR #3634 "STYLE: Remove `const` overload of `SubjectImplementation::AddObserver`" could also be merged first! (PR #3624 removes a few lines of code that are actually adjusted by this PR! These adjustments are no longer necessary when PR #3624 is merged beforehand.)

dzenanz:
I merged #3634. Now this branch has merge conflicts.

dzenanz (approved):

N-Dekker:
> I merged #3634. Now this branch has merge conflicts.

Thanks @dzenanz The conflicts should be fixed with the `rebase master` + force-push that I just did. 😃

N-Dekker:
OK, I think this one is ready to be merged, again! 😄 I was trying to sneak in another partially overlapping pull request, PR #3636 ("STYLE: Remove non-const overload of `Object::AddObserver`, add `mutable`") but it got stuck somehow, and it shouldn't stop this one (#3620) from getting merged anyway!

PR #3621, opened 2022-09-15
BUG: Open more NIFTI images successfully

### LINKED ISSUE #3514: Nifti file can not be opened after Jul 2020 commit
This Nifti [file](https://data.kitware.com/api/v1/file/62e79bc4bddec9d0c45a8d02/download) can not be opened after [ENH: Prefer to use sform over qform when both are set](https://github.com/InsightSoftwareConsortium/ITK/commit/34231b57021418fdb6afd4fcf5082a73b12969ed). The file was generated by [VINCI](https://vinci.sf.mpg.de/).

The issue was posted in the [Slicer forum](https://discourse.slicer.org/t/open-nifti-in-3d-slicer/24495).

In particular, this [block scope](https://github.com/InsightSoftwareConsortium/ITK/blob/8d3197081233d3a69a44a43dd706dcf7d6661275/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx#L1913-L1932), starting at the [line 1913](https://github.com/InsightSoftwareConsortium/ITK/blob/8d3197081233d3a69a44a43dd706dcf7d6661275/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx#L1913), prevents loading. The error message _"ITK only supports orthonormal direction cosines. No orthonormal definition found!"_ is not applicable.

Looks like that IO expects a header like this (a kind of IndexToPointMatrix)
```
…
pixdim 0.0 0.865759 0.865759 0.796 0.0 0.0 0.0 0.0
…
srow_x   0.865759 0.0      0.0      -54.975697
srow_y   0.0      0.865759 0.0      -54.975697
srow_z   0.0      0.0      0.796    -37.411999
```
but there is

```
…
pixdim 0.0 0.865759 0.865759 0.796 0.0 0.0 0.0 0.0
…
srow_x    1.0   0.0    0.0     -54.975697
srow_y    0.0   1.0    0.0     -54.975697
srow_z    0.0   0.0    1.0     -37.411999
```

The file worked with earlier releases, orientation and spacing were correct:
```
Spacing: [0.865759, 0.865759, 0.796]
Origin: [54.9757, 54.9757, -37.412]
Direction:
-1  0  0
 0 -1  0
 0  0  1
```
The file can be opened If the block is removed. Spacing and orientation will be like above again, _sform_ will be preferred.

Just FYI, please close the issue if not possible to fix.

dzenanz:
If sform and qform have different spacing, ITK refused to read the image. This commit fixes that, and qform information is used.

Co-authored-by: Mihail Isakov <mihail.isakov@googlemail.com>

Closes #3514.

## PR Checklist


gdevenyi:
Perhaps this should print a warning? Not sure it violates the NIFTI spec but it certainly sounds like a malformed NIFTI

issakomi:
> Perhaps this should print a warning? Not sure it violates the NIFTI spec but it certainly sounds like a malformed NIFTI

Here i tried to do with warning a month ago, BTW
https://github.com/issakomi/ITK/commit/aaaac5b8e89c87f1d295be56b78a38289be5e555#comments

blowekamp:
Thank you for adding data to test for this tricky case!

hjmjohnson (approved):
I'm going to approve this and defer to the community sentiment.

(Personally, I prefer failure to ambiguity, but I'm ok with this solution).

pieper:
I agree @hjmjohnson the ambiguity is unsettling.  With @lassoan we've discussed a "nifti patcher" that would let users input out-of-band knowledge (e.g. use the sform and ignore the qform) or explore options to fix imprecision (e.g. fix the sign of pixdim[0] if the head is flipped).  Do you think there's a home in ITK for that kind of tool?  If not an option would be for us to do something in python with nibabel.  We could expose either tool as a CLI with an interface module in Slicer.

dzenanz:
ITK is a library, and doesn't really come with tools. A "nifti patcher" could be an [example](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples).

dzenanz:
@hjmjohnson @pieper @lassoan should we print a warning to the console? Or is it better without a warning?

pieper:
I think warnings are appropriate anytime the data is ambiguous.

lassoan:
Printing a warning to the console is not great, because it is hard to capture it and show it to the user on the GUI. Instead it would be better if the IO classes would offer a message collection (list of messages that gather findings during loading of the data set; each message has a type - error, warning, info - and either a string or ID to specify the message text) that the application could display to the user.

dzenanz:
Adding that mechanism is completely feasible, but quite a bit of work. Even if we decide to do that sometime, it will not be a part of this PR. @lassoan do you prefer a warning or no warning for this case?

lassoan:
Logging a warning using itkWarningMacro could be useful (even though most users would not notice it).

If ITK had a mechanism to capture log messages for a specific object (similarly to [vtkErrorSink](https://github.com/Slicer/vtkAddon/blob/master/vtkErrorSink.h)) then log messages could be easily turned into a message collection.

dzenanz:
The force-push uses @issakomi's commit with a warning.

PR #3623, opened 2022-09-16
ENH: Test `itk::ImageFileReader` image IO to requested region mismatch

jhlegarreta:
- ENH: Test `itk::ImageFileReader` image IO to requested region mismatch
- STYLE: Break `ImageCompare` module test command into multiple lines

## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [X] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [X] Added test (or behavior not changed)

PR #3626, opened 2022-09-16
BUG: Fix `itk::STAPLEImageFilter` test argument order

jhlegarreta:
Fix `itk::STAPLEImageFilter` test argument order.

Set the `maximumIterations` and `confidenceWeight` arguments to the
wrong order in commit aa76dfb, which made it so that `maximumIterations`
was being set to 0, which in turn caused the code coverage to drop.

## PR Checklist


jhlegarreta:
Weirdly enough, this was not causing a regression error. Maybe the filter is not working as expected. To be investigated separately.

The PR will need to be rebased on `master` once PR #3623 gets merged.

PR #3629, opened 2022-09-17
BUG: `Object::InvokeEvent` should not use a dangling Observer pointer

N-Dekker:
`SubjectImplementation::InvokeEventRecursion` (an internal helper function of
`Object::InvokeEvent`) tried to "find" a *possibly* dangling pointer `o` in its
list of `Observer` pointers.

It appears more secure to check if an observer is still in the list by
comparing its tag to the tags of the other observers.

----

FYI A dangling `Observer` pointer `o` actually occurs when running the test function `testDeleteObserverDuringEvent()` from [CommandObserverObjectTest](https://github.com/InsightSoftwareConsortium/ITK/blob/2f2b29cbbe3fe91474933e0194db8d581d5b3247/Modules/Core/Common/test/itkCommandObserverObjectTest.cxx), while executing the following line of code:

https://github.com/InsightSoftwareConsortium/ITK/blob/2f2b29cbbe3fe91474933e0194db8d581d5b3247/Modules/Core/Common/test/itkCommandObserverObjectTest.cxx#L101

Which triggers calling the callback function `onUserRemove`:

https://github.com/InsightSoftwareConsortium/ITK/blob/2f2b29cbbe3fe91474933e0194db8d581d5b3247/Modules/Core/Common/test/itkCommandObserverObjectTest.cxx#L67-L72

### DIFF
--- a/Modules/Core/Common/src/itkObject.cxx
+++ b/Modules/Core/Common/src/itkObject.cxx
@@ -221,11 +221,16 @@ SubjectImplementation::InvokeEventRecursion(const EventObject &                 
     // save observer
     const Observer * o = *i;
 
+    // Save its tag, before the observer could /possibly/ be removed.
+    const unsigned long tag{ o->m_Tag };
+
     if (o->m_Event->CheckEvent(&event))
     {
       InvokeEventRecursion(event, self, ++i);
 
-      if (!m_ListModified || std::find(m_Observers.begin(), m_Observers.end(), o) != m_Observers.end())
+      const auto hasSameTag = [tag](const Observer * const observer) { return observer->m_Tag == tag; };
+
+      if (!m_ListModified || std::any_of(m_Observers.begin(), m_Observers.end(), hasSameTag))
       {
         o->m_Command->Execute(self, event);
       }


blowekamp (approved):
The change provide a good stepping block to correct behavior when changing to unique_ptr.

However, I object to labeling the old as having a "BUG". The dangling pointer is not dereferenced, it is check if it is still valid as is correct. All usages of raw pointers are not buggy and problematic.

blowekamp:
Also for reference, this case of deleting events during event's actually regularly occurs in SimpleITK because the "Delete" event of an object is used for free up some dependent resources/referenced, like other event objects.

N-Dekker:
> However, I object to labeling the old as having a "BUG".  The dangling pointer is not dereferenced, it is check if it is still valid as is correct.

Thanks for your comment, @blowekamp However, I still believe it is a bug. Even if the dangling pointer is not dereferenced! Please have another look at the original code from `SubjectImplementation::InvokeEventRecursion`:

https://github.com/InsightSoftwareConsortium/ITK/blob/a639acc22c41072d5f4cf201756db5b0add7457d/Modules/Core/Common/src/itkObject.cxx#L221-L231

The observer initially pointed to by `o` (line 222) may be removed during the recursion, so then `std::find` cannot find it anymore. Which is fine. But during the recursion, when that particular observer is removed, another observer may be _added_ as well, right? This newly added observer might coincidentally get exactly the same memory address as the one that was just removed.

When that happens, `std::find` will find the pointer `o`, but it's not pointing to the same observer anymore. I think `o->m_Command->Execute(self, event)` should then _not_ be called. Right?

blowekamp:
> When that happens, `std::find` will find the pointer `o`, but it's not pointing to the same observer anymore. I think `o->m_Command->Execute(self, event)` should then _not_ be called. Right?

I will concede that this sequence of events is not impossible. Good call. However I am skeptical that this is was you were originally call a bug. 🧐 😆 

This recursive function was a very elegant solution to a rather tricky problem encountered. 😀

N-Dekker:
> I will concede that this sequence of events is not impossible. Good call.

Thanks @blowekamp 

> This recursive function was a very elegant solution...

I believe so, but it will be even more elegant when using tags instead of dangling pointers  😸

PR #3634, opened 2022-09-18
STYLE: Remove `const` overload of `SubjectImplementation::AddObserver`

N-Dekker:
`SubjectImplementation` is an internal (hidden) helper class of `itk::Object`, which only uses the non-const overload of `SubjectImplementation::AddObserver`.

Removing this unused overload reduces maintenance.

----
Ubuntu2004 Coverage 8 appears to indicate that indeed, the `const` overload of `SubjectImplementation::AddObserver` was not covered: https://open.cdash.org/viewCoverageFile.php?buildid=8162114&fileid=52190897 (line 143-150) @jhlegarreta Can you please confirm that, Jon?

### DIFF
--- a/Modules/Core/Common/src/itkObject.cxx
+++ b/Modules/Core/Common/src/itkObject.cxx
@@ -67,9 +67,6 @@ public:
   unsigned long
   AddObserver(const EventObject & event, Command * cmd);
 
-  unsigned long
-  AddObserver(const EventObject & event, Command * cmd) const;
-
   void
   RemoveObserver(unsigned long tag);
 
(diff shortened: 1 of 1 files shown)


jhlegarreta:
> Ubuntu2004 Coverage 8 appears to indicate that indeed, the const overload of SubjectImplementation::AddObserver was not covered: https://open.cdash.org/viewCoverageFile.php?buildid=8162114&fileid=52190897 (line 143-150) @jhlegarreta Can you please confirm that, Jon?

I do not run coverage builds locally; always trust the dashboard reports, so if the dashboard says so...

dzenanz (approved):

PR #3638, opened 2022-09-20
Make `itk::Object` implementation more private, more hidden, more const

N-Dekker:
- Moved `Observer` into an unnamed namespace
- Made `SubjectImplementation` a private nested type of itk::Object
- Made protected `SubjectImplementation` members private
- Made iteration over observers "const"

### DIFF
--- a/Modules/Core/Common/include/itkObject.h
+++ b/Modules/Core/Common/include/itkObject.h
@@ -38,8 +38,6 @@
 
 namespace itk
 {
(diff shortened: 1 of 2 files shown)


dzenanz (approved):

N-Dekker:
I would suggest postponing this PR until after PR #3645 (COMP: Work around GCC <= 9.1 "error: invalid application of ‘sizeof’...") is merged.

PR #1628, opened 2020-02-17
STYLE: Clang tidy cleanups

hjmjohnson:
More clang-tidy identified cleanups found when building with remote modules.

PR #3641, opened 2022-09-21
Avoid `new` and `delete` for internal data members of ObjectFactoryBase

N-Dekker:


dzenanz (approved):

Modules/Core/Common/src/itkObjectFactoryBase.cxx
```diff
@@ -107,23 +100,18 @@ class ObjectFactoryBasePrivate : public LightObject
   ~ObjectFactoryBasePrivate() override
   {
     itk::ObjectFactoryBase::UnRegisterAllFactories();
-    if (m_InternalFactories)
+    for (auto & m_InternalFactorie : m_InternalFactories)
```
dzenanz, line 103: English spelling is `Factory`, not `Factorie` 😄

dzenanz (approved):
Did you figure out why heap allocation was used before?

Modules/Core/Common/src/itkObjectFactoryBase.cxx
N-Dekker: Indeed @dzenanz The term `m_InternalFactorie` appears to be introduced with a code _refactorieng_  ;-) PR #1628 commit 7f2c44cf534ac8e4a77eb0ec111c84b1c5d34be7: https://github.com/InsightSoftwareConsortium/ITK/blob/7f2c44cf534ac8e4a77eb0ec111c84b1c5d34be7/Modules/Core/Common/src/itkObjectFactoryBase.cxx#L109-L112

Anyway, I'd rather not include fixes of old typos with this PR, though. These can be fixed later, IMhO. Unless you insist...

Leengit: My check with
```shell
cd git/ITK
egrep -Ri 'factorie([^s]|$)'
```
shows the misspelling "factorie" occurs only in this variable in this file, on exactly two lines of code.

dzenanz: I thought you are introducing this typo in this commit, due to Dutch spelling being `Factorie` 😄

Of course, it can be fixed later. Maybe a new STYLE commit as part of this PR?

N-Dekker: @Leengit Thanks, but then you may also try `for (const auto & m_` for similar typo's. 😺  (Remember we use the `m_` prefix only for member variables.)  Anyway, I'd rather defer such typo fixes to another PR, someone else, possibly.... unless you click "Request changes"!

Leengit: FWIW,
```shell
egrep -Ri 'for ?\(auto ?& ?[a-z_]*ie[a-z_]* '
```
returns only the case we know about.

N-Dekker: OK, but `for (const auto & m_` does occur twice anyway.

Leengit: For the issue of whether the looping variable should properly begin with `m_`, I find a total of three instances with `egrep -Ri 'for ?\((const )?auto ?&? ?m_[a-z_]* ?'`:
```cpp
Modules/IO/GDCM/src/itkGDCMSeriesFileNames.cxx:    for (const auto & m_InputFileName : m_InputFileNames)
Modules/Core/Common/src/itkObjectFactoryBase.cxx:      for (auto & m_InternalFactorie : *m_InternalFactories)
Modules/Core/Mesh/include/itkConnectedRegionsMeshFilter.hxx:    for (const auto & m_RegionSize : m_RegionSizes)
```

N-Dekker:
> Did you figure out why heap allocation was used before?

@dzenanz Good question 🤔 It appears that `m_RegisteredFactories` was declared as a pointer already with the first version of ObjectFactoryBase, as committed by Bill Hoffman (@billhoffman), 11 April 2000: https://github.com/InsightSoftwareConsortium/ITK/blob/e31b986a15de8bef09dd351ecbf5926ff00c8c57/Code/Common/itkObjectFactoryBase.h#L163

You may ask Bill, but you know, at that time, raw pointers and manual memory management were very common! Everyone was using raw pointers around 2000! Maybe that could explain why...?

dzenanz:
STL was not consistently implemented in all the compilers, so that might be the reason.

dzenanz:
@Leengit @N-Dekker variable renames can go into another PR.

Modules/Core/Common/src/itkObjectFactoryBase.cxx
N-Dekker: @Leengit Would you possibly like to rename those three looping variables for us?  😃

Leengit: I've created Pull Request https://github.com/InsightSoftwareConsortium/ITK/pull/3646.

PR #3642, opened 2022-09-21
STYLE: Declare lambda's created for Singleton `const`, but non-static

N-Dekker:
The lifetime of lambda's passed as argument to the function `Singleton<T>` does not need to be extended beyond the function call, as these arguments directly get converted to `std::function` objects.


dzenanz (approved):

N-Dekker:
Please note that these particular lambda's are stateless: they have an empty capture list (`[]`). So there is no captured data to keep alive "statically" anyway.

Leengit:
These lambdas are pretty simple; could they be `constexpr`?  (I have no experience with that, so apologies if that's a stupid question.)

PR #3663, opened 2022-09-26
STYLE: Declare `m_VnlOptimizer` from Optimizers as unique_ptr

N-Dekker:
Also "defaulted" the destructors of those optimizer classes.


dzenanz (approved):

thewtex (approved):
🪨 🥇

PR #3665, opened 2022-09-26
STYLE: Make OverRideMap, ObjectFactoryBasePrivate private nested classes

N-Dekker:
Moved the forward-declaration of both `OverRideMap` and `ObjectFactoryBasePrivate` into the class definition of `ObjectFactoryBase`, as they are just part of the internal private implementation of `ObjectFactoryBase`.

Follow-up to commit 4c0526e4abd40b0157ad05645a8b32aab8d9e733, "STYLE: Make `SubjectImplementation` a private nested type of itk::Object".

### DIFF
--- a/Modules/Core/Common/include/itkObjectFactoryBase.h
+++ b/Modules/Core/Common/include/itkObjectFactoryBase.h
@@ -56,10 +56,6 @@ namespace itk
  * \ingroup ITKCommon
  */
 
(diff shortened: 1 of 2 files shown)


dzenanz:
I only now realize I asked for your review late yesterday afternoon, Brad. Sorry for merging so quickly.

PR #3670, opened 2022-09-29
STYLE: Declare BlockMatchingImageFilter data members as unique_ptr<T[]>

N-Dekker:


dzenanz (approved):

N-Dekker:
"First interaction / build (pull_request_target)" has failed! Any suggestion?

It says:
```
internal/modules/cjs/loader.js:1216
      throw new ERR_REQUIRE_ESM(filename, parentPath, packageJsonPath);
      ^

Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /node_modules/@octokit/graphql/index.js
require() of ES modules is not supported.
require() of /node_modules/@octokit/graphql/index.js from /node_modules/@actions/github/lib/github.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /node_modules/@octokit/graphql/package.json.
```

dzenanz:
The action probably broke again. Hopefully, someone (action author) will fix it soon, without us having to put effort into it. We can ignore this "failure" for the purposes of merging PRs.

N-Dekker:
> The action probably broke again. Hopefully, someone (action author) will fix it soon, without us having to put effort into it. We can ignore this "failure" for the purposes of merging PRs.

Thanks @dzenanz I see now, the same failure occurred with PR #3671, which I just submitted as well.

PR #3674, opened 2022-09-30
Remove unused local dynamically allocated `double[]` buffers from Segmentation module 

N-Dekker:
Might slightly speedup the code.

Note that those two local buffers were already there before `make_unique_for_overwrite` was introduced.


dzenanz (approved):

hjmjohnson:
@dzenanz   Do you know what the "First interaction" failing check is?

dzenanz:
It is failing because it needs a code update. I think it is on Matt's TODO list. We can ignore it, as it is not a real check.

PR #3678, opened 2022-10-05
Create SECURITY.md

JoliParris:
Hello❤️

## PR Checklist

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
+currently being supported with security updates.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 5.1.x   | :white_check_mark: |
+| 5.0.x   | :x:                |
+| 4.0.x   | :white_check_mark: |
+| < 4.0   | :x:                |
+
+## Reporting a Vulnerability
+
+Use this section to tell people how to report a vulnerability.
+
+Tell them where to go, how often they can expect to get an update on a
+reported vulnerability, what to expect if the vulnerability is accepted or
+declined, etc.


dzenanz (commented):
@blowekamp @hjmjohnson @jhlegarreta @thewtex I guess we should have a discussion about this, unless we had it already and I forgot about it.

If we add this document (it needs to be modified, suggested version is not what is done today), we should add it to the release checklist, as that is the time we will be updating supported versions.

jhlegarreta:
I am not sure about the trustworthiness of this PR. I encourage you to have a look at their activity.

> I guess we should have a discussion about this, unless we had it already and I forgot about it.

I do not remember having discussed this, and nor do I remember anyone bringing about this topic.

> If we add this document (it needs to be modified, suggested version is not what is done today), we should add it to the release checklist, as that is the time we will be updating supported versions.

Maybe I am very limited in my knowledge, but I do not see how security can affect ITK's codebase. I do not remember either having seen any PR relevant to this topic within the last 10+ years.

dzenanz:
If some image reader has a bug, we could be exposed to an attack via specially crafted images, similarly to that JPEG bug where viewing a specially crafted file causes a problem.

What this suggests is to say which versions we support, and how to receive reports of security vulnerabilities. We currently don't have any way to provide secrecy to such reports, which is meant to give us time to address it before public catches wind of it.

thewtex:
As @jhlegarreta noted, @JoliParris appears to not be a serious account.

This file adds clutter without much value. ITK is not a library with the same security sensitivities like other libraries involved in networking, etc. If there is a security related patch, which is extremely rare, a PR should be created, and it is added to the latest release. We do not have a mechanism for private disclosure of security patches.

jhlegarreta:
Closing following the comments above.

PR #3110, opened 2022-01-16
BUG: region.IsInside(zeroSizedRegion) should always return false

N-Dekker:
As documented by commit a872710c63ab223e78f45de8d3e4bc435172867d "STYLE: Documenting the behavior of the IsInside() method when the argument region has zero size in any of its dimensions", by Luis Ibanez (@luisibanez), 18 December 2009:

> Test if a region (the argument) is completely inside of this region. If the region that is passed as argument to this method, has a size of value zero, then it will not be considered to be inside of the current region, even its starting index is inside.

Included GoogleTest unit tests for both zero-sized and one-sized regions as argument.

---
This pull request competes with pull request #3108 "DOC: Remove mistake from `ImageRegion::IsInside(region)` documentation" I think there is a choice to be made! There is a disagreement between the code and the documentation. So now, should the code follow the documentation (this pull request, #3110), or should the documentation follow the code (pull request #3108)?

PR #3682, opened 2022-10-10
BUG: ImageIORegion::IsInside should return false for a zero-sized region

N-Dekker:
This commit ensures that `ImageIORegion::IsInside` returns false whenever its argument is an `ImageIORegion` that contains no pixels at all.

Following the adjustment of `itk::ImageRegion::IsInside(const Self &)` from pull request https://github.com/InsightSoftwareConsortium/ITK/pull/3110 commit 1295d239397f111bd008c1093d9ddc5a3867669a
"BUG: `region.IsInside(zeroSizedRegion)` should always return false", which was included with ITK v5.3rc04

See also "What bool value should `imageRegion.IsInside(zeroSizedRegion)` return?" at https://discourse.itk.org/t/what-bool-value-should-imageregion-isinside-zerosizedregion-return/4734

### DIFF
--- a/Modules/Core/Common/src/itkImageIORegion.cxx
+++ b/Modules/Core/Common/src/itkImageIORegion.cxx
@@ -190,25 +190,28 @@ ImageIORegion::IsInside(const IndexType & index) const
   return true;
 }
 
-/** Test if a region (the argument) is completely inside of this region */
-bool
-ImageIORegion::IsInside(const Self & region) const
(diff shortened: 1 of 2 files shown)


dzenanz (approved):
LGTM

JoliParris:
Hello Monde🌏

hjmjohnson (approved):

PR #3685, opened 2022-10-12
Fix nifti read sform case

### LINKED ISSUE #3676: Incorrect call when comparing matrices in nifti headers
### Description

In this code transpose is used when inverse is intended:

https://github.com/InsightSoftwareConsortium/ITK/blob/7357008cbc2880cda302f00efd1a39b02aeffe08/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx#L1978-L1982

It looks to me like the code was copy-pasted from above where transpose was intended but the code was not corrected:

https://github.com/InsightSoftwareConsortium/ITK/blob/7357008cbc2880cda302f00efd1a39b02aeffe08/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx#L1943-L1946

### Steps to Reproduce

I single stepped through while researching a different issue and realized this call was always false even when the matrices are identical.  On inspection I could see these are full index to physical transforms and the transpose of one is not equal to the other.

### Expected behavior

Code should implement intended behavior.

### Actual behavior

Call always fails.

### Reproducibility

Always.

### Versions

Since Jul 4, 2020 

The code in question comes from this commit:

https://github.com/InsightSoftwareConsortium/ITK/commit/34231b57021418fdb6afd4fcf5082a73b12969ed

### Environment

All

### Additional Information

N/A

hjmjohnson:
Resolves #3676 has been a long-hidden bug in a rare corner case with reading nifti images.

## PR Checklist


dzenanz:
Code looks good. The builds completed with warnings. Strangely, builds are not visible on the dashboard.

dzenanz:
Oh, builds are visible on [yesterday's dashboard](https://open.cdash.org/index.php?project=Insight&date=2022-10-11) 😄  Here are the [compile warnings](https://open.cdash.org/viewBuildError.php?type=1&buildid=8210276).

dzenanz (approved):

dzenanz (approved):

dzenanz:
@sjh26 @jamesobutler This should be cherry-picked in Slicer's fork, or Slicer's fork rebased on top of current master.

PR #3688, opened 2022-10-12
STYLE: Rename nested ObjectFactoryBase type OverRideMap to `OverrideMap`

N-Dekker:
The uppercase 'R' in the identifier `OverRideMap` appears unusual. Note that this nested type is just a private implementation detail, so this change does not affect of the API.

Also removed the internal type alias `StringOverMap`, which was only used once.


Leengit:
Perhaps also change `(Over|Under)Flow` to `\1flow`.

hjmjohnson (approved):

dzenanz (approved):

PR #3690, opened 2022-10-12
STYLE: Change Over/UnderFlow to Over/Underflow

Leengit:
"OverFlow" and "UnderFlow" should not have capital "F" characters.

### DIFF
--- a/Modules/Core/Common/test/CMakeLists.txt
+++ b/Modules/Core/Common/test/CMakeLists.txt
@@ -555,9 +555,9 @@ if("${CMAKE_SYSTEM_NAME}" MATCHES "Darwin|Linux")
   itk_add_test(NAME itkFloatingPointExceptionsTest2 COMMAND ITKCommon1TestDriver
     itkFloatingPointExceptionsTest ZeroDivByZero)
   itk_add_test(NAME itkFloatingPointExceptionsTest3 COMMAND ITKCommon1TestDriver
-    itkFloatingPointExceptionsTest FPOverFlow)
+    itkFloatingPointExceptionsTest FPOverflow)
   itk_add_test(NAME itkFloatingPointExceptionsTest4 COMMAND ITKCommon1TestDriver
-    itkFloatingPointExceptionsTest FPUnderFlow)
+    itkFloatingPointExceptionsTest FPUnderflow)
 
   set_tests_properties(
     itkFloatingPointExceptionsTest1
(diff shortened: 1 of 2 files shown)


dzenanz (approved):

hjmjohnson (approved):

Modules/Core/Common/test/CMakeLists.txt
```diff
@@ -555,9 +555,9 @@ if("${CMAKE_SYSTEM_NAME}" MATCHES "Darwin|Linux")
   itk_add_test(NAME itkFloatingPointExceptionsTest2 COMMAND ITKCommon1TestDriver
     itkFloatingPointExceptionsTest ZeroDivByZero)
   itk_add_test(NAME itkFloatingPointExceptionsTest3 COMMAND ITKCommon1TestDriver
-    itkFloatingPointExceptionsTest FPOverFlow)
+    itkFloatingPointExceptionsTest FPOverflow)
   itk_add_test(NAME itkFloatingPointExceptionsTest4 COMMAND ITKCommon1TestDriver
-    itkFloatingPointExceptionsTest FPUnderFlow)
+    itkFloatingPointExceptionsTest FPUnderflow)
```
N-Dekker, line 560: Why not just "Overflow" and "Underflow"? The prefix "FP" does not seem necessary, and it isn't used for "DivByZero" and "ZeroDivByZero" either.

N-Dekker (approved):
Thanks Lee, nice follow-up to PR #3688 (Rename nested ObjectFactoryBase type OverRideMap to `OverrideMap`). My comment on the "FP" prefix is just a suggestion, no show-stopper.

Modules/Core/Common/test/CMakeLists.txt
dzenanz: These tests are part of Common module, which has a lot of tests. FP prefix to the tests names is useful, at least to me.

Leengit: I'll leave it as is.

N-Dekker: Merged 👍  I guess the "FP" prefix stands for _floating point_, right? If so, then the prefix seems unnecessary, as the names of the tests `itkFloatingPointExceptionsTest3` and `itkFloatingPointExceptionsTest4` already make it clear to me that they are about floating points. Anyway, it's unrelated to the capital "F" in Over/UnderFlow, so I rest my case for now  😺

dzenanz: > itkFloatingPointExceptionsTest4

I didn't see this name 😄 I thought that FPUnderflow was test name.

PR #3696, opened 2022-10-19
DOC: updated comment for ITK_FUTURE_LEGACY_REMOVE

issakomi:
Changed _ITKv5_ to _ITKv6_ in cmake comment for ITK_FUTURE_LEGACY_REMOVE.


thewtex (approved):

issakomi:
Maybe
"Completely remove compilation of code which will become deprecated by default in a future major release"
or
"Completely remove compilation of code which will become deprecated by default in **the** future major release"

dzenanz (approved):

dzenanz:
I think the current wording is right. With each major release, we remove deprecated things, and promote "future deprecated" into currently deprecated. I don't remember whether we made some exceptions for things which were deprecated soon before a major version switch.

PR #3697, opened 2022-10-20
STYLE: removed duplicated messages

issakomi:
Removed _itkWarningMacro_ just before _itkExceptionMacro_ with exactly the same messages.
Looks rather unusual and highly likely not useful.

### DIFF
--- a/Modules/IO/ImageBase/src/itkImageIOBase.cxx
+++ b/Modules/IO/ImageBase/src/itkImageIOBase.cxx
@@ -108,7 +108,6 @@ ImageIOBase::SetDimensions(unsigned int i, SizeValueType dim)
 {
   if (i >= m_Dimensions.size())
   {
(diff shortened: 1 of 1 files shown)


thewtex (approved):
:+1:

N-Dekker:
Indeed, the `itkWarningMacro` calls don't seem very useful to me either.

For the record, these combined `itkWarningMacro`-`itkExceptionMacro` calls appear introduced by commit b5d2dbba8fff02a7febf84f76a47e435a3b108a8 "BUG: fixing bug 4560; add exceptions when indices are out of range...", Brad Davis, 2008-05-26.

N-Dekker (approved):
Thanks Mihail!

Just a minor suggestion: the subject line of your commit message might have been a little bit more specific (as long as it is <= 72 chars), for example:

    STYLE: Remove duplicated warning messages from ImageIOBase

But that's no show-stopper of course. Approved 👍

hjmjohnson (approved):

PR #3700, opened 2022-10-20
STYLE: Replace `reset(new T[n])` with make_unique_for_overwrite<T[]>(n)

N-Dekker:
Replaced `\.reset\(new (\w+)\[(.+)\]\);` with ` = make_unique_for_overwrite<$1[]>($2);`, using regular expressions in Visual Studio 2019.

Following C++ Core Guidelines, September 23, 2022, "Avoid calling `new` and `delete` explicitly", from http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-newdelete

### DIFF
--- a/Modules/Core/Common/include/itkNeighborhoodAllocator.h
+++ b/Modules/Core/Common/include/itkNeighborhoodAllocator.h
@@ -60,7 +60,7 @@ public:
   /** Defaulted destructor */
   ~NeighborhoodAllocator() = default;
 
-  /** Allocates memory using new() */
+  /** Allocates memory. */
   void
   Allocate(unsigned int n)
   {
@@ -68,7 +68,7 @@ public:
     m_ElementCount = n;
   }
 
-  /** Deallocates memory using delete[](). */
+  /** Deallocates memory. */
   void
   Deallocate()
   {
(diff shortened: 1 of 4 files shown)


dzenanz (approved):

Modules/Core/Common/include/itkNeighborhoodAllocator.h
```diff
@@ -63,7 +64,7 @@ class NeighborhoodAllocator
   void
   Allocate(unsigned int n)
   {
-    m_Data.reset(new TPixel[n]);
+    m_Data = make_unique_for_overwrite<TPixel[]>(n);
```
issakomi, line 67: The comments for _Allocate_ and _Deallocate_
`/** Allocates memory using new() */`
`/** Deallocates memory using delete[](). */`
are probably outdated, BTW.

N-Dekker: @issakomi Well... strictly speaking, those old comments are still correct, as `make_unique_for_overwrite` and `unique_ptr<T[]>` _do_ use `new` and `delete[]` (as they should). However, I think it should be considered an implementation detail of NeighborhoodAllocator. Moreover, removing them from the comments reduces the noise when searching for the remaining "naked" `new`s and `delete[]`s in the source code  😃 So thanks Mihail 👍 Please check the commit that I just added!

dzenanz (approved):

hjmjohnson (approved):

PR #3702, opened 2022-10-22
ENH: Add a test to check reading large `NIfTI` image subregions

### LINKED ISSUE #3625: Nifti volume of specific size + Shrink filter cause segmentation fault
### Description

The combination of NIFTI + Shrink filter causes a segmentation fault for some dimensions and sizes.

### Steps to Reproduce (MVE)

1. Generate a nifti volume of the size (1034, 1034, 1020)
    ```python
    import itk

    pixel_type = itk.US
    dimension = 3
    image_type = itk.Image[pixel_type, dimension]

    start = itk.Index[dimension]()
    start.Fill(0)

    size = itk.Size[dimension]()
    size[0] = 1034
    size[1] = 1034
    size[2] = 1020

    region = itk.ImageRegion[dimension]()
    region.SetIndex(start)
    region.SetSize(size)

    image = image_type.New(Regions=region)
    image.Allocate()

    itk.imwrite(image, "segmentation-fault-test.nii")
    ```

2. Create a pipeline with `ImageFileReader` + `ShrinkImageFilter` and update at the end
    ```python
    import itk
    from itk import ShrinkImageFilter

    image_path = "segmentation-fault-test"
    reader = itk.ImageFileReader.New(FileName=image_path)
    shrink = ShrinkImageFilter.New(reader)
    shrink.SetShrinkFactors(4)

    print("Update")
    shrink.Update()

    print("GetOutput")
    data = shrink.GetOutput()

    del shrink
    del reader
    print("Done")

   ```

3. Run it (for windows add `-Xfaulthandler` to the python command)

### Expected behavior

Script runs through without any errors and outputs all print messages.

### Actual behavior

A `segmentation fault` (linux) / `access violation` (windows) after the "Update" printout.

### Reproducibility

With the mentioned dimensions 100%.

### Versions/Environments

Linux:
- itk 5.1.2 and 5.2.1
- Python 3.8

Windows:
- itk 5.3rc4.post2
- Python 3.9

### Additional Information

- The dimensions matter, e.g.
  - `1034, 1034, 1020` (1090539120) crash
  - `4048, 2048, 132` (1094320128) works
  - `826, 843, 3314`(2307597852) crash
- Other voxel types uint8 (with `1034, 1034, 2040`) or float are also affected
- Seems to be related to the byte size in combination with the dimensions
- Other shrink factors (as `2`) are also affected
- `.nii.gz` is also affected
- `nrrd` seems to work FINE
- In some other workflow this causes a corruption of the volume without a seg fault (I try to add a picture later)

jhlegarreta:
Add a test to check reading large `NIfTI` image subregions.

Resolves #3625.

## PR Checklist

### DIFF
--- a/Modules/IO/NIFTI/test/itkNiftiLargeImageRegionReadTest.cxx
+++ b/Modules/IO/NIFTI/test/itkNiftiLargeImageRegionReadTest.cxx
@@ -1,0 +1,73 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         https://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+#include "itkImageFileReader.h"
+#include "itkImageFileWriter.h"
+#include "itkTestingMacros.h"
+
+
+int
+itkNiftiLargeImageRegionReadTest(int argc, char * argv[])
+{
+
+  if (argc != 2)
+  {
+    std::cerr << "Missing arguments" << std::endl;
+    std::cerr << "Usage: " << itkNameOfTestExecutableMacro(argv) << " outputFileName" << std::endl;
+    return EXIT_FAILURE;
+  }
+
+
+  constexpr int Dimension = 3;
+  using PixelType = unsigned short;
+
+  using ImageType = itk::Image<PixelType, Dimension>;
+
+  // Create a large image
+  ImageType::SizeType   size = { { 1034, 1034, 1020 } };
+  ImageType::RegionType region;
+  region.SetSize(size);
+
+  auto image = ImageType::New();
+  image->SetRegions(region);
+  image->Allocate(true);
+
+  const std::string fname{ argv[1] };
+  itk::WriteImage(image, fname);
+
+
+  using ReaderType = itk::ImageFileReader<ImageType>;
+  auto reader = ReaderType::New();
+
+  reader->SetFileName(fname);
+
+  // Ensure that a sufficiently large subregion, requiring allocating a pixel buffer larger than INT_MAX bytes, can be
+  // read, as nifti_clib did not support reading such large subregions previously.
+  region.SetSize({ { 1029, 1029, 1017 } });
+  reader->GetOutput()->SetRequestedRegion(region);
+
+  ITK_TRY_EXPECT_NO_EXCEPTION(reader->Update());
+
+  const auto output = reader->GetOutput();
+  ITK_TEST_EXPECT_TRUE(output != nullptr);
+  ITK_TEST_EXPECT_TRUE(output->GetBufferedRegion() == region);
+
+
+  std::cout << "Test finished." << std::endl;
+  return EXIT_SUCCESS;
+}
--- a/Modules/IO/NIFTI/test/CMakeLists.txt
+++ b/Modules/IO/NIFTI/test/CMakeLists.txt
@@ -13,6 +13,7 @@ itkNiftiImageIOTest10.cxx
 itkNiftiImageIOTest11.cxx
 itkNiftiImageIOTest12.cxx
 itkNiftiImageIOTest13.cxx
+itkNiftiLargeImageRegionReadTest.cxx
 itkNiftiReadAnalyzeTest.cxx
 itkNiftiReadWriteDirectionTest.cxx
 itkExtractSlice.cxx
@@ -114,6 +115,10 @@ itk_add_test(NAME itkNiftiQFormSFormDifferentSpacingTest
       itkNiftiImageIOTest13
       DATA{Input/ChickenEgg-zeros.nii.gz}
       )
+
+itk_add_test(NAME itkNiftiLargeImageRegionReadTest
+      COMMAND ITKIONIFTITestDriver itkNiftiLargeImageRegionReadTest
+      ${ITK_TEST_OUTPUT_DIR}/itkNiftiLargeImageRegionReadTest.nii.gz)
 
 itk_add_test(NAME itkNiftiWriteCoerceOrthogonalDirectionTest
         COMMAND ITKIONIFTITestDriver itkNiftiWriteCoerceOrthogonalDirectionTest


dzenanz (approved):

Modules/IO/NIFTI/test/itkNiftiLargeImageRegionReadTest.cxx
```diff
@@ -0,0 +1,68 @@
...
+
+
+int
+itkNiftiLargeImageRegionReadTest(int argc, char * argv[])
+{
+
+  if (argc != 2)
+  {
+    std::cerr << "Missing arguments" << std::endl;
+    std::cerr << "Usage: " << itkNameOfTestExecutableMacro(argv) << " outputFileName" << std::endl;
```
N-Dekker, line 31: Is it necessary to pass the file name as command-line argument?  I'm just wondering, because tests without command-line arguments can be tried out more easily within the IDE (Visual Studio). The file name is always "${ITK_TEST_OUTPUT_DIR}/itkNiftiLargeImageRegionReadTest.nii.gz", right?

dzenanz: With a command line argument, the same code could be reused for other testing. As the file is residing in nitfiIO's tests directory, that is less practical than ideal. But a command line argument gives us flexibility of where to put this large output file.

Modules/IO/NIFTI/test/itkNiftiLargeImageRegionReadTest.cxx
```diff
@@ -0,0 +1,68 @@
...
+  using ReaderType = itk::ImageFileReader<ImageType>;
+  auto reader = ReaderType::New();
+
+  reader->SetFileName(fname);
+
+  // Ensure that a sufficiently large subregion of an abritrary pixel type image can be read
+  region.SetSize({ { 1029, 1029, 1017 } });
+  reader->GetOutput()->SetRequestedRegion(region);
+
+  ITK_TRY_EXPECT_NO_EXCEPTION(reader->Update());
```
N-Dekker, line 63: Is there anything more you could possibly check? Only just checking that a function call throws no exception seems a bit minimal to me. I mean, an `Update()` implementation that just _does nothing at all_ would then also pass the test.

Modules/IO/NIFTI/test/itkNiftiLargeImageRegionReadTest.cxx
N-Dekker: > With a command line argument, the same code could be reused for other testing.

Thanks, but is this code really intended to be used for other tests as well? I thought it was specifically for large Nifti files.

Modules/IO/NIFTI/test/itkNiftiLargeImageRegionReadTest.cxx
N-Dekker: OK, thanks. I still find it interesting that even an Update() function that does _nothing at all_ would pass the test, but no problem.

Modules/IO/NIFTI/test/itkNiftiLargeImageRegionReadTest.cxx
```diff
@@ -0,0 +1,68 @@
...
+  const std::string fname{ argv[1] };
+  itk::WriteImage(image, fname);
+
+
+  using ReaderType = itk::ImageFileReader<ImageType>;
+  auto reader = ReaderType::New();
+
+  reader->SetFileName(fname);
+
+  // Ensure that a sufficiently large subregion of an abritrary pixel type image can be read
```
N-Dekker, line 59: Small typo here "abritrary" (no show-stopper to me). Instead of fixing the typo, I'd rather just shorten the sentence to:

>     // Ensure that a sufficiently large subregion of an image can be read.

But that's just my preference, no issue.

Modules/IO/NIFTI/test/itkNiftiLargeImageRegionReadTest.cxx
jhlegarreta: @N-Dekker I do not like the idea of writing the output data file to the binaries folder for many reasons; I could have required a path, but then it would be to the user to try to guess the filename in case they want to inspect it. Note that as pointed in the related issue, at some point it was necessary to inspect the output file.

Modules/IO/NIFTI/test/itkNiftiLargeImageRegionReadTest.cxx
jhlegarreta: I'm happy to take suggestions on what to check considering the problems pointed in the issue. Otherwise, I'm fine with the test as it is.

Modules/IO/NIFTI/test/itkNiftiLargeImageRegionReadTest.cxx
jhlegarreta: Re-worded. The word is no longer present.

dzenanz (approved):

Modules/IO/NIFTI/test/itkNiftiLargeImageRegionReadTest.cxx
N-Dekker: Thanks Jon, I was thinking of something like the following, _after_ the Update() call:

    const auto output = reader->GetOutput();
    ITK_TEST_EXPECT_TRUE(output != nullptr);
    ITK_TEST_EXPECT_TRUE(output->GetBufferedRegion() == region);

Does that look reasonable to you as well? To check that the Update() has indeed generated an image of the requested size.

jhlegarreta: Sounds reasonable. Done. Force-pushed. Thanks Niels.

dzenanz (approved):
This quantifiability is a clear improvement.

Modules/IO/NIFTI/test/itkNiftiLargeImageRegionReadTest.cxx
```diff
@@ -0,0 +1,72 @@
...
+  itk::WriteImage(image, fname);
+
+
+  using ReaderType = itk::ImageFileReader<ImageType>;
+  auto reader = ReaderType::New();
+
+  reader->SetFileName(fname);
+
+  // Ensure that a sufficiently large subregion can be read
+  region.SetSize({ { 1029, 1029, 1017 } });
```
N-Dekker, line 60: Sorry Jon, would it still be possible to add a comment that explains why specifically this size `{ 1029, 1029, 1017 }` was chosen? I believe these numbers are not entirely arbitrary. `{ 1000, 1000, 1000 }` would also be large, right? Then why not `{ 1000, 1000, 1000 }`?!?  I think that's because the reading needs to be tested a buffer of more than `INT_MAX` bytes, right? 

If so, I think the comment could be something like:

    // Ensure that a sufficiently large subregion can be read, which would need the allocation a pixel buffer larger than
    // INT_MAX bytes. Previous versions of the nifti_clib (before 19 September 2022) did not yet support reading such
    // large subregions.    
 

What do you think? I hope I'm not asking too much  😃 !

jhlegarreta: Done. Slightly reworded.

N-Dekker (approved):
Thanks, Jon. Approved 👍 👍 👍 !

N-Dekker:
FYI itkNiftiLargeImageRegionReadTest has entered the 4th position in to top 25 of most time consuming tests! At least on the [Windows build](https://dev.azure.com/itkrobotwindow/ITK.Windows/_build/results?buildId=9095&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=77aad734-2057-5694-9ae2-ee1f5f26eae8) of commit f423f2d7c7b8f0c33eb2a4bd07561aae63e89ad0. The average ITK test takes less than 0.3 sec, but this one takes more than 9 sec. Unfortunately I have no idea how to speed it up.

Here is the top 25 (out of in total 621 tests) of most time consuming tests:

Test name | Duration (sec)
-- | --
Test 2483:   itkKullbackLeiblerCompareHistogramImageToImageMetricTest | 9.24
Test 2531: itkSyNPointSetRegistrationTest | 9.17
Test 1896: itkRegionGrow2DTest | 9.06
Test 1274:   itkNiftiLargeImageRegionReadTest | 9.05
Test 2460:   itkGaussianExponentialDiffeomorphicTransformParametersAdaptorTest | 8.84
Test 665: itkImageFillBufferTest4.1 | 8.59
Test 2537:   itkTimeVaryingBSplineVelocityFieldImageRegistrationTest | 8.16
Test 2271:   itkDemonsImageToImageMetricv4RegistrationTest3 | 7.88
Test 909:   itkFastMarchingImageFilterTest_wm_multipleSeeds_NoTopo | 7.69
Test 2459:   itkBSplineExponentialDiffeomorphicTransformParametersAdaptorTest | 7.53
Test 2532:   itkBSplineSyNPointSetRegistrationTest | 7.43
Test 2335:   itkDiffeomorphicDemonsRegistrationFilterTest06 | 7.28
Test 2330:   itkDiffeomorphicDemonsRegistrationFilterTest01 | 7.18
Test 803:   itkTimeVaryingBSplineVelocityFieldTransformTest | 7.15
Test 2270:   itkDemonsImageToImageMetricv4RegistrationTest2 | 7.11
Test 722:   itkFFTConvolutionImageFilterTest4x5Mean | 6.86
Test 723:   itkFFTConvolutionImageFilterTest5x5Mean | 6.86
Test 2148:   itkGrayscaleMorphologicalClosingImageFilterTest2 | 6.74
Test 777:   itkPatchBasedDenoisingImageFilterDefaultTest | 6.54
Test 31: vnl_basic_operation_timings | 6.16
Test 2150:   itkGrayscaleMorphologicalOpeningImageFilterTest2 | 6.16
Test 693:   itkRelabelComponentImageFilterTest | 6.08
Test 2533:   itkTimeVaryingBSplineVelocityFieldPointSetRegistrationTest | 6.07
Test 692:   ConnectedComponentImageFilter.checkerboard_3D | 5.81
Test 690:   RelabelComponentImageFilter.big_zero | 5.54

dzenanz:
Most of that time is presumably spent compressing/decompressing using GZIP in single thread. There is probably not a lot we can do to speed it up, but that should not be a problem.

N-Dekker:
> There is probably not a lot we can do to speed it up, but that should not be a problem.

@dzenanz OK, thanks, I agree. Although I _do_ think it could be a problem when a unit test that takes a lot of time _unnecessarily_. (Both because it takes resources _and_ it slows down the development process.) With this particular test I think it cannot be done much faster anyway. The size of the image must be `> INT_MAX` bytes, so it will take some time anyway.

PR #3707, opened 2022-10-27
COMP: Fix uninitialized vector warnings

jhlegarreta:
Fix uninitialized vector warnings.

Fixes:
```
Modules/Numerics/Statistics/include/itkDistanceMetric.h:101:45:
warning: 'm' may be used uninitialized [-Wmaybe-uninitialized]
```

and
```
Modules/Numerics/Statistics/include/itkMeasurementVectorTraits.h:70:72:
warning: 'm' may be used uninitialized [-Wmaybe-uninitialized]
```

and
```
Modules/Numerics/Statistics/include/itkMembershipFunctionBase.h:97:45:
warning: 'm' may be used uninitialized [-Wmaybe-uninitialized]
```

and
```
Modules/Numerics/Statistics/include/itkSample.h:121:45:
warning: 'm' may be used uninitialized [-Wmaybe-uninitialized]
```

raised for example in:
https://open.cdash.org/viewBuildError.php?type=1&buildid=8240358

## PR Checklist

### DIFF
--- a/Modules/Numerics/Statistics/include/itkDistanceMetric.h
+++ b/Modules/Numerics/Statistics/include/itkDistanceMetric.h
@@ -96,7 +96,7 @@ public:
   SetMeasurementVectorSize(MeasurementVectorSizeType s)
   {
     // Test whether the vector type is resizable or not
-    MeasurementVectorType m;
+    MeasurementVectorType m{};
 
     if (MeasurementVectorTraits::IsResizable(m))
     {
--- a/Modules/Numerics/Statistics/include/itkMeasurementVectorTraits.h
+++ b/Modules/Numerics/Statistics/include/itkMeasurementVectorTraits.h
@@ -66,7 +66,7 @@ public:
     // If the default constructor creates a vector of
     // length zero, we assume that it is resizable,
     // otherwise that is a pretty useless measurement vector.
-    TVectorType             m;
+    TVectorType             m{};
     MeasurementVectorLength len = NumericTraits<TVectorType>::GetLength(m);
 
     return (len == 0);
--- a/Modules/Numerics/Statistics/include/itkMembershipFunctionBase.h
+++ b/Modules/Numerics/Statistics/include/itkMembershipFunctionBase.h
@@ -92,7 +92,7 @@ public:
   SetMeasurementVectorSize(MeasurementVectorSizeType s)
   {
     // Test whether the vector type is resizable or not
-    MeasurementVectorType m;
+    MeasurementVectorType m{};
 
     if (MeasurementVectorTraits::IsResizable(m))
     {
--- a/Modules/Numerics/Statistics/include/itkSample.h
+++ b/Modules/Numerics/Statistics/include/itkSample.h
@@ -116,7 +116,7 @@ public:
   SetMeasurementVectorSize(MeasurementVectorSizeType s)
   {
     // Test whether the vector type is resizable or not
-    MeasurementVectorType m;
+    MeasurementVectorType m{};
 
     if (MeasurementVectorTraits::IsResizable(m))
     {


dzenanz (approved):

jhlegarreta:
Converted to a draft, as I think I need some suggestion here to provide an appropriate fix.

- Simply using a list-based initialization, as proposed in the first commit, yields type conversion narrowing warnings:
  https://open.cdash.org/viewBuildError.php?buildid=8241864
- The vectors in some of these templates can be of any type, including `itk::Array`, `itk::FixedArray`, `itk::Point`, `itk::RGBPixel`, `itk::VariableLengthVector`, `itk::Vector`, or even `std::vector`.
  - Not all of them share the same initialization methods, e.g. `fill` vs. `Fill` vs. `MakeFilled`).
  - Not all of them have the `ValueType` alias (e.g. to do `MeasurementVectorType m{ static_cast<MeasurementVectorType::ValueType>(0.0) };`).

So for `itk::itkDistanceMetric.h` and `itk::MembershipFunctionBase` I could get away with it using a loop as the length is a member variable; however, such is not the case for `itk::MeasurementVectorTraits` or `itk::Sample`, where the vector can contain other more exotic types.

Not sure whether attempting to cast to all of the possible types in `itk::`MeasurementVectorTraits::IsResizable` to then initialize the vector properly is possible/the only way here, or how types can be handled in the case of `itk::Sample`.

@N-Dekker, @dzenanz, @thewtex, @blowekamp I would be very grateful if you could chime in here :-)

Modules/Numerics/Statistics/include/itkDistanceMetric.h
```diff
@@ -97,6 +97,10 @@ class ITK_TEMPLATE_EXPORT DistanceMetric : public FunctionBase<TVector, double>
   {
     // Test whether the vector type is resizable or not
     MeasurementVectorType m;
+    for (size_t i = 0; i < m_MeasurementVectorSize; ++i)
+    {
+      m[i] = 0;
+    }
```
N-Dekker, line 103: Thanks for asking, @jhlegarreta. Please consider list-initialization by an empty initializer list `{}`, instead of a `for` loop. So in this case, just: `MeasurementVectorType m{};`

N-Dekker:
@jhlegarreta I found a somewhat similar pull request of yours  😃: PR #2393 commit 7548a390d71dc25b80e792703ee3cfea17853a99 "COMP: Fix missing initializer warning", March 2021. At that time, the warnings were solved by value-initialization, using an empty pair parentheses `()`.  Similarly, you could also do:

    const auto m = MeasurementVectorType();

N-Dekker:
For the record, the warning appears to be introduced with GCC 11. Minimal example: https://godbolt.org/z/nsbGd9TnM

```
struct MyVector
{
  int data[42];
};

bool
IsResizable(const MyVector &)
{
  return false;
}

int
main()
{
  MyVector v;
  return IsResizable(v) ? 1 : 0;
}
```

Modules/Numerics/Statistics/include/itkDistanceMetric.h
jhlegarreta: :+1: Thanks Niels. Done in db2418b.

jhlegarreta:
> @jhlegarreta I found a somewhat similar pull request of yours smiley: PR https://github.com/InsightSoftwareConsortium/ITK/pull/2393 commit https://github.com/InsightSoftwareConsortium/ITK/commit/7548a390d71dc25b80e792703ee3cfea17853a99 "COMP: Fix missing initializer warning", March 2021.

I see that the fix also came after a few comments and unsuccessful attempts, and a suggestion of yours :-).

The loop-based change resulted here in segfaults, probably because the size member variable having a value that does not apply to the local variable :expressionless:.

I prefer not to introduce the `const` style change here. Let's see if this makes the CIs happy and fixes the warnings in the dashboard first. The style change may then be applied in another PR.

N-Dekker (approved):
Thanks, Jon!  Honestly I believe the compiler should _not_ have produced those warnings in the first place. I consider it a compiler bug. My example at https://github.com/InsightSoftwareConsortium/ITK/pull/3707#issuecomment-1295775822 obviously does not use an uninitialized value, and still GCC 11 produces the warning.

I guess the next generation compilers might produce a warning saying that the initializations that are added by this PR are unnecessary! 😺  

But for now, I understand that we need to address today's compiler warnings, so OK, approved  👍

dzenanz (approved):

issakomi:
> I believe the compiler should _not_ have produced those warnings in the first place. I consider it a compiler bug. My example at [#3707 (comment)](https://github.com/InsightSoftwareConsortium/ITK/pull/3707#issuecomment-1295775822) obviously does not use an uninitialized value, and still GCC 11 produces the warning.
> 
> I guess the next generation compilers might produce a warning saying that the initializations that are added by this PR are unnecessary! smiley_cat
> 
> But for now, I understand that we need to address today's compiler warnings, so OK, approved +1

@N-Dekker Thank you! BTW, gcc 12.2 does the same. Probably the reason is that the struct is passed  to the _IsResizable_ function and compiler doesn't check what happens there and gives -W**maybe**-uninitialized. There is also -Wuninitialized, could be triggered e.g.

```
#include <iostream>
struct MyVector
{
  int data[42];
};

int main(int, char **)
{
  MyVector v;
  int x = v.data[5] + 1;
  std::cout << x << std::endl;
  return 0;
}
```

Strange thing is that for the above example -Wall is required, IMHO, there is serious error here with an undefined result.

Unfortunately gcc 12.2 brings many new warnings to ITK build.

issakomi:
Another finding, for completeness, if MyVector is passed without `const` there is no warning

```
struct MyVector
{
  int data[42];
};

bool
IsResizable(MyVector &)
{
  return false;
}

int
main()
{
  MyVector v;
  return IsResizable(v) ? 1 : 0;
}
```

N-Dekker:
> if MyVector is passed without `const` there is no warning

Thanks Mihail, but that makes (some) sense... at least, it _would_ make sense if the function definition of `IsResizable(MyVector &)` would be elsewhere (e.g., hidden away in a separate library).

A _non-const_ parameter might still be assigned to inside the function (`IsResizable`, in our example), _before_ using its value. So then the warning that it "may be used uninitialized" would be less relevant.

PR #3552, opened 2022-08-16
COMP: Set CMake Policy 135 to NEW to suppress warnings in CMake 3.24

dzenanz:
Prevents warnings.

## PR Checklist
- [x] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [x] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)

PR #3709, opened 2022-10-29
COMP: Set `DOWNLOAD_EXTRACT_TIMESTAMP` to `TRUE` in wrapping modules

jhlegarreta:
Set `DOWNLOAD_EXTRACT_TIMESTAMP` to `TRUE`: the timestamps of the
extracted files will reflect the timestamps in the archive.

Configure/autoconf based projects like SWIG and PCRE have timestamped
files for their auto configuration dependencies. Thus, they depend on
the timestamps of the extracted files, and the external projects require
setting `DOWNLOAD_EXTRACT_TIMESTAMP` to `TRUE` to build robustly.

Set the `CMP0135` policy to `NEW` for the CastXML CMake-based project.

Suppresses CMake warnings in wrapping modules.

Fixes:
```
CMake Warning (dev) at /usr/local/share/cmake-3.24/Modules/ExternalProject.cmake:3074 (message):
  The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is
  not set.  The policy's OLD behavior will be used.  When using a URL
  download, the timestamps of extracted files should preferably be that of
  the time of extraction, otherwise code that depends on the extracted
  contents might not be rebuilt if the URL changes.  The OLD behavior
  preserves the timestamps from the archive instead, but this is usually not
  what you want.  Update your project to the NEW behavior or specify the
  DOWNLOAD_EXTRACT_TIMESTAMP option with a value of true to avoid this
  robustness issue.
  /usr/local/share/cmake-3.24/Modules/ExternalProject.cmake:4170 (_ep_add_download_command)
  Wrapping/Generators/SwigInterface/CMakeLists.txt:142 (ExternalProject_Add)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Found BISON: /usr/bin/bison (found version "3.5.1")
CMake Warning (dev) at /usr/local/share/cmake-3.24/Modules/ExternalProject.cmake:3074 (message):
  The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is
  not set.  The policy's OLD behavior will be used.  When using a URL
  download, the timestamps of extracted files should preferably be that of
  the time of extraction, otherwise code that depends on the extracted
  contents might not be rebuilt if the URL changes.  The OLD behavior
  preserves the timestamps from the archive instead, but this is usually not
  what you want.  Update your project to the NEW behavior or specify the
  DOWNLOAD_EXTRACT_TIMESTAMP option with a value of true to avoid this
  robustness issue.
  /usr/local/share/cmake-3.24/Modules/ExternalProject.cmake:4170 (_ep_add_download_command)
  Wrapping/Generators/SwigInterface/CMakeLists.txt:243 (ExternalProject_Add)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at /usr/local/share/cmake-3.24/Modules/ExternalProject.cmake:3074 (message):
  The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is
  not set.  The policy's OLD behavior will be used.  When using a URL
  download, the timestamps of extracted files should preferably be that of
  the time of extraction, otherwise code that depends on the extracted
  contents might not be rebuilt if the URL changes.  The OLD behavior
  preserves the timestamps from the archive instead, but this is usually not
  what you want.  Update your project to the NEW behavior or specify the
  DOWNLOAD_EXTRACT_TIMESTAMP option with a value of true to avoid this
  robustness issue.
  /usr/local/share/cmake-3.24/Modules/ExternalProject.cmake:4170 (_ep_add_download_command)
  Wrapping/Generators/CastXML/CMakeLists.txt:63 (ExternalProject_Add)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

raised for example in:
https://open.cdash.org/build/8245525/configure

Related documentation:
https://cmake.org/cmake/help/latest/module/ExternalProject.html?highlight=download_extract_timestamp
https://cmake.org/cmake/help/latest/policy/CMP0135.html

## PR Checklist

### DIFF
--- a/Wrapping/Generators/CastXML/CMakeLists.txt
+++ b/Wrapping/Generators/CastXML/CMakeLists.txt
@@ -60,6 +60,10 @@ else()
     if(ITK_BINARY_DIR)
       itk_download_attempt_check(CastXML)
     endif()
+    cmake_policy(PUSH)
+    if(POLICY CMP0135)
+      cmake_policy(SET CMP0135 NEW)
+    endif()
     ExternalProject_Add(castxml
       URL ${_castxml_url}
       URL_HASH SHA512=${_castxml_hash}
@@ -67,6 +71,7 @@ else()
       BUILD_COMMAND ""
       INSTALL_COMMAND "${CMAKE_COMMAND}" -E copy_directory "${CMAKE_CURRENT_BINARY_DIR}/castxml-prefix/src/castxml" "${CMAKE_CURRENT_BINARY_DIR}/castxml"
       )
+    cmake_policy(POP)
     set(CASTXML_EXECUTABLE ${castxml_ep})
     set(CASTXML_EXECUTABLE ${castxml_ep} CACHE FILEPATH "castxml executable." FORCE)
   # Build from source
--- a/Wrapping/Generators/SwigInterface/CMakeLists.txt
+++ b/Wrapping/Generators/SwigInterface/CMakeLists.txt
@@ -138,6 +138,11 @@ else()
               install
         )
       endif()
+      if(${CMAKE_VERSION} VERSION_LESS 3.24)
+        set(download_extract_timestamp_flag)
+      else()
+        set(download_extract_timestamp_flag DOWNLOAD_EXTRACT_TIMESTAMP TRUE)
+      endif()
       set(pcre_hash "abac4c4f9df9e61d7d7761a9c50843882611752e1df0842a54318f358c28f5953025eba2d78997d21ee690756b56cc9f1c04a5ed591dd60654cc78ba16d9ecfb")
       ExternalProject_Add(PCRE
         URL "https://data.kitware.com/api/v1/file/hashsum/sha512/${pcre_hash}/download"
@@ -148,6 +153,7 @@ else()
           --prefix=${CMAKE_CURRENT_BINARY_DIR}/PCRE
           --enable-shared=no
         ${extra_external_project_commands}
+        ${download_extract_timestamp_flag}
         )
 
       # swig uses bison find it by cmake and pass it down
@@ -240,6 +246,11 @@ message(STATUS \"Swig configure successfully completed.\")
         )
       endif()
 
+      if(${CMAKE_VERSION} VERSION_LESS 3.24)
+        set(download_extract_timestamp_flag)
+      else()
+        set(download_extract_timestamp_flag DOWNLOAD_EXTRACT_TIMESTAMP TRUE)
+      endif()
       ExternalProject_Add(swig
         URL "https://data.kitware.com/api/v1/file/hashsum/sha512/${swig_hash}/download"
         URL_HASH SHA512=${swig_hash}
@@ -247,6 +258,7 @@ message(STATUS \"Swig configure successfully completed.\")
           ${extra_swig_configure_env} ${CMAKE_COMMAND} -P "${_swig_configure_script}"
         ${extra_external_project_commands}
         DEPENDS PCRE
+        ${download_extract_timestamp_flag}
         )
     endif()
 


jhlegarreta:
- Not sure if there is a way to inherit them from
https://github.com/InsightSoftwareConsortium/ITK/commit/e3d6c94eb962ce8bbe4de2d83a1289dfa5cb6921#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR23

- I put the statements as close as possible to the lines triggering the warnings. Let me know if there is a better place if the above cannot be done.

jhlegarreta:
The warning is no longer present:

Previously:
`ITK.Linux.Python`
https://dev.azure.com/itkrobotlinuxpython/ITK.Linux.Python/_build/results?buildId=7740&view=logs&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=f58be928-45a1-58e7-c1ad-04869565b3f4&l=1145
`ITK.macOS.Python`: https://dev.azure.com/itkrobotmacospython/ITK.macOS.Python/_build/results?buildId=7812&view=logs&j=a5e52b91-c83f-5429-4a68-c246fc63a4f7&t=5add3200-7efa-59c0-0a45-43bce0c77c69&l=1124

Now:
`ITK.Linux.Python`: https://dev.azure.com/itkrobotlinuxpython/ITK.Linux.Python/_build/results?buildId=7741&view=logs&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=f58be928-45a1-58e7-c1ad-04869565b3f4&l=1144
`ITK.macOS.Python`: https://dev.azure.com/itkrobotmacospython/ITK.macOS.Python/_build/results?buildId=7813&view=logs&j=a5e52b91-c83f-5429-4a68-c246fc63a4f7&t=5add3200-7efa-59c0-0a45-43bce0c77c69&l=1123

N-Dekker:
Thanks for addressing this warning Jon. Just looking at https://cmake.org/cmake/help/latest/policy/CMP0135.html Is the _new_ CMake default behavior indeed preferable, for external projects CastXML and PCRE?

dzenanz (approved):

dzenanz:
As those are usually extracted only once, it should probably not matter?

jhlegarreta:
> Thanks for addressing this warning Jon. Just looking at https://cmake.org/cmake/help/latest/policy/CMP0135.html Is the new CMake default behavior indeed preferable, for external projects CastXML and PCRE?

TBH, I do not know what the impact of the `OLD` vs. `NEW` policy would be: i.e. if timestamps are checked somehow. Maybe to download again the contents depending on its value? Not sure. In any case, I believe we should have no warnings. If setting the `DOWNLOAD_EXTRACT_TIMESTAMP` variable is preferred over this, I'd happy to hear the rationale/suggestions. Also, then the change in PR #3552 should be maybe commented as to why it should apply only to specific parts.

blowekamp:
My tests are confirming that configure/autoconf based projects like SWIG and PCRE[2] depend on the time stamps of the extracted files and  the external project require setting DOWNLOAD_EXTRACT_TIMESTAMP to 0 to build robustly. This should be changed in this PR.

The reported build error is related to the `aclocal` (or specific version of it) executable missing.  Which is caused but autoconf being triggered due to timestamps. For this error to occur the system must have CMake >2.24, missing the required aclocal. Additionally the "extract time" for the extract files may vary based on the time it takes to extract the files from the archive, and the order of the files in the archive.

jhlegarreta:
:100: Thanks Brad. I've amended the commit.

N-Dekker:
DOWNLOAD_EXTRACT_TIMESTAMP is introduced with CMake 3.24: https://cmake.org/cmake/help/latest/release/3.24.html?highlight=download_extract_timestamp#modules Does that mean that `cmake_minimum_required` should then be adjusted as well?

https://github.com/InsightSoftwareConsortium/ITK/blob/d81a02cc23e08a4ddaa3a3c4468f2093cbf9cf5a/CMakeLists.txt#L13-L15

Wrapping/Generators/CastXML/CMakeLists.txt
```diff
@@ -66,6 +66,7 @@ else()
       CONFIGURE_COMMAND ""
       BUILD_COMMAND ""
       INSTALL_COMMAND "${CMAKE_COMMAND}" -E copy_directory "${CMAKE_CURRENT_BINARY_DIR}/castxml-prefix/src/castxml" "${CMAKE_CURRENT_BINARY_DIR}/castxml"
+      DOWNLOAD_EXTRACT_TIMESTAMP FALSE
```
blowekamp, line 69: CastXML is a CMake project, so the time stamp dependencies issues with configure/autoconf are not an issue.

blowekamp:
> DOWNLOAD_EXTRACT_TIMESTAMP is introduced with CMake 3.24

The minimum require version certainly should not be updated to 3.24. `ExternalProject_Add` appears to ignore unknown names arguments, however I am not sure this is behavior that should be relied upon?

The alternate would be conditionally assign `DOWNLOAD_EXTRACT_TIMESTAMP FALSE` to a variable and use that.

Wrapping/Generators/CastXML/CMakeLists.txt
jhlegarreta: There is a warning associated to that line:
```
Wrapping/Generators/CastXML/CMakeLists.txt:63 (ExternalProject_Add)
```

N-Dekker:
> The minimum require version certainly should not be updated to 3.24.

OK, but we now have `set(ITK_NEWEST_VALIDATED_POLICIES_VERSION "3.19.7")`, in the root CMakeLists. 

https://github.com/InsightSoftwareConsortium/ITK/blob/d81a02cc23e08a4ddaa3a3c4468f2093cbf9cf5a/CMakeLists.txt#L14

Shouldn't that already have preserved the old behavior, even when using CMake 3.24?

Wrapping/Generators/CastXML/CMakeLists.txt
blowekamp: Yes, for this project likely either behavior of the timestamp should be OK. However, I thought there was motivation to use the NEW CMP135 behavior where it could be used, like the castXML project. To do that the CMP135 could be set to new, and this line removed, or the value could be set to TRUE.

jhlegarreta:
> Shouldn't that already have preserved the old behavior, even when using CMake 3.24?

@N-Dekker According to
https://github.com/InsightSoftwareConsortium/ITK/blob/49150c8a701f629a122646f708a802f961d1ace7/CMakeLists.txt#L10

```
# CMake versions greater than the ITK_NEWEST_VALIDATED_POLICIES_VERSION policies will
# continue to generate policy warnings "CMake Warning (dev)...Policy CMP0XXX is not set:"
```

So I guess they will appear for 3.24.

> Yes, for this project likely either behavior of the timestamp should be OK. However, I thought there was motivation to use the NEW CMP135 behavior where it could be used, like the castXML project. To do that the CMP135 could be set to new, and this line removed, or the value could be set to TRUE.

> The minimum require version certainly should not be updated to 3.24. ExternalProject_Add appears to ignore unknown names arguments, however I am not sure this is behavior that should be relied upon?
> The alternate would be conditionally assign DOWNLOAD_EXTRACT_TIMESTAMP FALSE to a variable and use that.

@blowekamp Thanks for the suggestions and directions forward. Tried to do them in 6e20e06. Not sure what's wrong with it, though: the warnings appear back, and the variable contents are not properly parsed by `ExternalProject_Add`:
https://open.cdash.org/build/8248990/configure

Run out of ideas.

Wrapping/Generators/CastXML/CMakeLists.txt
```diff
@@ -60,6 +60,9 @@ else()
     if(ITK_BINARY_DIR)
       itk_download_attempt_check(CastXML)
     endif()
+    if(${CMAKE_VERSION} VERSION_LESS 3.24)
+      cmake_policy(SET CMP0135 NEW)
```
blowekamp, line 64: I saw this change after my comment:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/CMakeLists.txt#L21

jhlegarreta: Yep, https://github.com/InsightSoftwareConsortium/ITK/pull/3709#issuecomment-1295966013 The question there is how to inherit those values if we think they should apply to the wrapping projects.

mathstuf: This makes no sense. CMP0135 didn't exist until 3.24. I think you want `if (POLICY CMP0135)` to detect whether the policy exists or not.

You probably want a `cmake_policy(PUSH)` scope (with corresponding `POP` later).

jhlegarreta: > This makes no sense. CMP0135 didn't exist until 3.24.

Yes, Ben, you're right, I realized yesterday when I still saw the CI warning.

> I think you want if (POLICY CMP0135) to detect whether the policy exists or not.
> You probably want a cmake_policy(PUSH) scope (with corresponding POP later).

Can you elaborate on this please? As you see in the comment above, the policy is set for the root `CMakeLists.txt` so I am wondering whether your solution would allow us to inherit or apply existing policies to other subprojects such as the one here. According to the [documentation](https://cmake.org/cmake/help/latest/command/cmake_policy.html#cmake-policy-stack) we may be willing to use `block()/endblock()`?

mathstuf: Policies have their own scoping mechanism, so you'll want to push and pop any policies you set in something like this.

jhlegarreta: Sorry, Ben, maybe I was not clear in my question. ITK sets the following policies:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/CMakeLists.txt#L21

So if I understand well, according to the CMake [documentation](https://cmake.org/cmake/help/latest/command/cmake_policy.html#cmake-policy-stack), they are put on a stack. So the question is if doing the below
```
cmake_policy(PUSH)
ExternalProject_Add(castxml
  URL ${_castxml_url}
  URL_HASH SHA512=${_castxml_hash}
  CONFIGURE_COMMAND ""
  BUILD_COMMAND ""
  INSTALL_COMMAND "${CMAKE_COMMAND}" -E copy_directory "${CMAKE_CURRENT_BINARY_DIR}/castxml-prefix/src/castxml" "${CMAKE_CURRENT_BINARY_DIR}/castxml"
  )
cmake_policy(POP)
```

would apply the values of ALL policies in the stack (so `CMP0135` to `NEW` among them) to the project we are configuring within the `cmake_policy(PUSH)/cmake_policy(POP)` scope.

mathstuf: The policy would be set for the `ExternalProject_add` call which basically does some `add_custom_command` calls at heart. By the time `castxml` is running its configure step, the policy scope mentioned here is long gone and doesn't do anything.

jhlegarreta: OK, that confuses me further then, as you suggested using `cmake_policy(PUSH)/cmake_policy(POP)` in https://github.com/InsightSoftwareConsortium/ITK/pull/3709#discussion_r1027296475.

Should we just do ?
```
if(POLICY CMP0135)
      cmake_policy(SET CMP0135 NEW)
```

mathstuf: Yes, if you're setting policies not at top-level, use a scope:

```cmake
cmake_policy(PUSH)
if (POLICY CMP0135)
  cmake_policy(SET CMP0135 NEW)
endif ()
# Code that needs CMP0135
cmake_policy(POP)
```

It's just that the policy does *not* affect the project ExternalProject is building, only the prep work that EP itself does.

jhlegarreta: OK, thanks for the additional explanation Ben. Unless you tell the contrary, I think it is fine that the policy applies to the EP prep work, as the step that is triggering the warning is related to the moment the files are extracted.

I've amended the commit. Let's see if that makes the things work.

Wrapping/Generators/CastXML/CMakeLists.txt
jhlegarreta: Done. Resolving. Let's see what the CIs tell us.

thewtex (approved):
Outstanding, thank you so much for addressing this, @jhlegarreta !

blowekamp:
@jhlegarreta I apologize. It appears I had the logic of `DOWNLOAD_EXTRACT_TIMESTAMP` backwards. I corrected it in SimpleITK, but not here. 

 DOWNLOAD_EXTRACT_TIMESTAMP should be set to TRUE for SWIG and PCRE.

I have reread the documentation for the [CMP0135 ](https://cmake.org/cmake/help/latest/policy/CMP0135.html)and it's clarity should be improved.. Setting DOWNLOAD_EXTRACT_TIMESTAMP to TRUE is the OLD behavior. The OLD behaviors preserved the timestamp in the archive. This is required for SWIG and PCRE which have timed stamped files for autoconf dependencies.

@bradking FYI for CMake maintainer.

jhlegarreta:
@blowekamp It's fine, thanks Brad. I should have done a better job understanding as well. I've amended the commit, changed the subject and message body according to what you described. See if it now matches the appropriate behavior.

dzenanz:
@blowekamp For an open-source project like [CMake](https://gitlab.kitware.com/cmake/cmake/), if you think that documentation should be improved, the usual response/expectation is to propose the change you think would improve it. I took a look at the doc you pointed out, and I didn't an obvious improvement opportunity.

dzenanz (approved):

jhlegarreta:
Warnings are gone:
`ITK.Linux.Python`: https://dev.azure.com/itkrobotlinuxpython/ITK.Linux.Python/_build/results?buildId=7862&view=logs&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=f58be928-45a1-58e7-c1ad-04869565b3f4&l=1150
`ITK.macOS.Python`: https://dev.azure.com/itkrobotmacospython/ITK.macOS.Python/_build/results?buildId=7932&view=logs&j=a5e52b91-c83f-5429-4a68-c246fc63a4f7&t=5add3200-7efa-59c0-0a45-43bce0c77c69&l=1129
(the warnings appeared around those lines)

https://open.cdash.org/index.php?project=Insight#!#Experimental

Wrapping/Generators/CastXML/CMakeLists.txt
jhlegarreta: Resolving as changes have been addressed. Unresolve in case you still find that the issues were not addressed appropriately.

PR #3713, opened 2022-11-01
BUG: Remove backslash + indent from literals, to avoid unintended spaces

N-Dekker:
When a backspace at the end of a line of code was used as continuation character _inside_ a string literal, an indentation at the next line would unintentionally become part of the string itself.

This commit has removed the backslash, the line break, and the indentation from those string literals. Clang-Format has taken care of splitting the literals over multiple lines properly.

It affects a QuadEdgeMeshEulerOperatorDeleteCenterVertexFunction debug message, an exception message of Versor, and various exception messages from "Numerics/Statistics".


dzenanz (approved):

jhlegarreta:
Thanks for doing this Niels. I'd say this is a `STYLE` change, though: the basckslash/indent do not introduce any bug.

I know we'll need to wait for another round of CIs, but I'd be for changing the commit subject and PR prefix before merging this. Thanks.

N-Dekker:
> Thanks for doing this Niels. I'd say this is a `STYLE` change, though: the backslash/indent do not introduce any bug.

Thanks for your suggestion Jon, but this PR _does_ fix a few mistakes (these extra spaces are _erroneously_ there), and it _does_ change the behavior of the code. Because it changes debug output and exception messages. So I believe it's not just a matter of style. I think it's a bug, although I would agree that it's a _very little_ bug. Hope it's OK to you.

jhlegarreta:
> Thanks for your suggestion Jon, but this PR does fix a few mistakes (these extra spaces are erroneously there), and it does change the behavior of the code. Because it changes debug output and exception messages. So I believe it's not just a matter of style. I think it's a bug, although I would agree that it's a very little bug. Hope it's OK to you.

e.g. a typo in a word that is part of a message is also erroneously there, but does not introduce a problem with a result of the software. Here the situation is the same: the exception message or debug output might be incomplete or be difficult to read, but it does not affect any computation or result. So I am sorry, but I do not consider this a bug.

N-Dekker:
@jhlegarreta So do you really really want the commit subject line to be changed from "BUG" to "STYLE"? If you insist, I will adjust the commit text (albeit "under protest", because I think a STYLE commit should not change the runtime behavior).

jhlegarreta (approved):
Thanks Niels :+1:.

dzenanz:
> pull request distinguished between

That is a useful distinction.

@N-Dekker please merge when ready.

N-Dekker:
>  please merge when ready.

I always click the button "Ready for review" to indicate that it is ready. GitHub does not offer any other way for me to approve a PR that I submitted myself.

dzenanz:
I clicked "ready for review" button. I was impatient 😄

N-Dekker:
> I clicked "ready for review" button. I was impatient 😄

@dzenanz It isn't really the most exciting PR of the year, is it?  😸 😸

PR #3714, opened 2022-11-01
STYLE: Use `{}` instead of dummy as `NumericTraits::GetLength` argument

N-Dekker:
It appears unnecessary to use a local variable as function argument of `NumericTraits::GetLength(const T &)`, when this local variable is a "dummy", that is only just default-initialized, or initialized by `T()` or `{}`.

This commit removes such local "dummy" variables, and passes `{}` to those `NumericTraits::GetLength` calls that had such a dummy as argument.

Follow-up to pull request https://github.com/InsightSoftwareConsortium/ITK/pull/3710 commit be8cb4233b6510f436333213b7cac7c9ab7c07a7
"STYLE: Use `{}` as `MeasurementVectorTraits::IsResizable` argument"

### DIFF
--- a/Modules/Core/Common/include/itkImage.hxx
+++ b/Modules/Core/Common/include/itkImage.hxx
@@ -147,8 +147,7 @@ Image<TPixel, VImageDimension>::GetNumberOfComponentsPerPixel() const
 {
   // use the GetLength() method which works with variable length arrays,
   // to make it work with as much pixel types as possible
-  const auto p = PixelType();
-  return NumericTraits<PixelType>::GetLength(p);
+  return NumericTraits<PixelType>::GetLength({});
 }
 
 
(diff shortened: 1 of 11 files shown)


dzenanz (approved):

N-Dekker:
[Darwin-Build9051-PR3714-Use-empty-initializer-list-for-NumericTraits-GetLength](https://open.cdash.org/build/8250544/configure) says:

```
CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
-- Configuring incomplete, errors occurred!
See also "/Users/runner/work/1/s-build/CMakeFiles/CMakeOutput.log".

```

I guess that's unrelated to the code changes I'm proposing here... any suggestion?

PR #3715, opened 2022-11-02

jhlegarreta:
- Test exceptions using the `ITK_TRY_EXPECT_EXCEPTION` macro.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/test/itkNumericTraitsTest.cxx
+++ b/Modules/Core/Common/test/itkNumericTraitsTest.cxx
@@ -34,6 +34,7 @@
 #include "itkNumericTraitsRGBPixel.h"
 #include "itkNumericTraitsTensorPixel.h"
 #include "itkNumericTraitsVariableLengthVectorPixel.h"
+#include "itkTestingMacros.h"
 
 namespace
 {
@@ -799,6 +800,13 @@ itkNumericTraitsTest(int, char *[])
   CheckFixedArrayTraits(itk::Point<double, 5>());
   CheckFixedArrayTraits(itk::Point<long double, 5>());
 
+  using PointType = itk::Point<long double, 5>;
+  auto         pointPixel = PointType();
+  unsigned int pointSize = itk::NumericTraits<PointType>::GetLength(pointPixel) + 1;
+  ITK_TRY_EXPECT_EXCEPTION(itk::NumericTraits<PointType>::SetLength(pointPixel, pointSize));
+
+  pointSize = itk::NumericTraits<PointType>::GetLength(pointPixel) - 1;
+  ITK_TRY_EXPECT_EXCEPTION(itk::NumericTraits<PointType>::SetLength(pointPixel, pointSize));
 
   // itk::RGBPixel<char>()
   CheckFixedArrayTraits(itk::RGBPixel<char>());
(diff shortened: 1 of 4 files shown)


jhlegarreta:
@dzenanz @tbirdso @thewtex Not sure what's happening with the macOS build: it is unable to find Ninja. Suggestions are welcome.

tbirdso:
In the [macOS build log](https://dev.azure.com/itkrobotmacos/c947f891-8c0e-4d7f-9768-5685d97c2b9c/_apis/build/builds/9059/logs/15):

> 2022-11-02T12:34:09.0691690Z   WARNING: The script ninja is installed in '/Library/Frameworks/Python.framework/Versions/3.11/bin' which is not on PATH.

It looks like Python 3.11 is used by default but the build process is probably expecting 3.10?

jhlegarreta:
> It looks like Python 3.11 is used by default but the build process is probably expecting 3.10?

Maybe the all Azure builds have switched to Python 3.11 by default as it has been released last week.

However, IDK why it should affect `ITK.macOS` as it does not need Python.

tbirdso:
@jhlegarreta The Azure pipeline appears to currently use whatever Python is available by default for `pip install`-ing dependencies. Here's the full log entry for installing ninja:

```
2022-11-02T12:34:04.4121660Z + sudo python3 -m pip install ninja
2022-11-02T12:34:08.3273160Z WARNING: The directory '/Users/runner/Library/Caches/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you should use sudo's -H flag.
2022-11-02T12:34:08.9320560Z Collecting ninja
2022-11-02T12:34:08.9630950Z   Downloading ninja-1.10.2.4-py2.py3-none-macosx_10_9_universal2.macosx_10_9_x86_64.macosx_11_0_arm64.macosx_11_0_universal2.whl (231 kB)
2022-11-02T12:34:08.9732170Z      ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 231.5/231.5 kB 39.8 MB/s eta 0:00:00
2022-11-02T12:34:09.0498210Z Installing collected packages: ninja
2022-11-02T12:34:09.0691690Z   WARNING: The script ninja is installed in '/Library/Frameworks/Python.framework/Versions/3.11/bin' which is not on PATH.
2022-11-02T12:34:09.0694040Z   Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
2022-11-02T12:34:09.0742320Z Successfully installed ninja-1.10.2.4
```

I don't see any dependency install failures in the log, so either of these paths should be a viable fix:
1. Enforce Python == 3.10 prior to `pip install`-ing dependencies
2. Set `PATH=$PATH:/Library/Frameworks/Python.framework/Versions/3.11/bin` before building ITK

jhlegarreta:
:+1: thanks for the suggestions Tom.

Now, reading the CI configurations, all install `Ninja` through Python. Not sure why it only hit `ITK.macOS`, maybe it was the latest to trigger, and then the change came in.

I'd guess that we do not want to switch our CI builds to `Python 3.11` for now (unless we start a separate CI to test against the latest Python version, whatever it may be).

I'll let you discuss the above and I'll propose a PR to change all builds:
https://github.com/InsightSoftwareConsortium/ITK/tree/abc6a272cda280b2428bcc2ad709c0185dde6624/Testing/ContinuousIntegration

tbirdso:
> I'd guess that we do not want to switch our CI builds to Python 3.11 for now (unless we start a separate CI to test against the latest Python version, whatever it may be).

I agree that at a minimum we should be setting the Python version that pipelines are using so that we aren't hit with surprise upgrades. In Python workflows it appears we use the `UsePythonVersion` task, I imagine this could be added to cxx workflows as well. Maybe these could be standardized to Python 3.9 for now? https://github.com/InsightSoftwareConsortium/ITK/blob/abc6a272cda280b2428bcc2ad709c0185dde6624/Testing/ContinuousIntegration/AzurePipelinesLinuxPython.yml#L27

At some point we will want to build Python 3.11 packages for distribution. @thewtex and @jcfr may be able to share thoughts on that effort.

jhlegarreta:
> I agree that at a minimum we should be setting the Python version that pipelines are using so that we aren't hit with surprise upgrades. In Python workflows it appears we use the UsePythonVersion task, I imagine this could be added to cxx workflows as well. Maybe these could be standardized to Python 3.9 for now? 

:+1: PR #3716. Thanks Tom.

tbirdso:
@jhlegarreta Would you please rebase on https://github.com/InsightSoftwareConsortium/ITK@master following #3716?

jhlegarreta:
> @jhlegarreta Would you please rebase on https://github.com/InsightSoftwareConsortium/ITK@master following https://github.com/InsightSoftwareConsortium/ITK/pull/3716?

Done.

dzenanz (approved):

PR #3717, opened 2022-11-02
Remove backslash (+ indent) from string literals in tests

N-Dekker:
Follow-up to pull request #3713 "Remove backslash (+ indent) from string literals", but now applied to tests (`itk*Test*.cxx` files).

Removes unwanted (unintentional) spaces from test output messages, which were caused by an indentation after a backslash continuation character.


N-Dekker:
FYI, I did a separate commit for itkBSplineDecompositionImageFilterTest because it was slightly more complicated, and it could not be handled entirely by the regular expressions used for the first commit (which is the largest commit of the three).

dzenanz (approved):
It is hard to capture all cases of something in ITK with a regular expression 😄

jhlegarreta:
Thanks for doing this Niels. Rebasing on `master` will probably allow to make the `ITK.macOS` build happy.

PR #3718, opened 2022-11-02
ENH: Increase coverage for `itk::RandomImageSource`

jhlegarreta:
Increase coverage for `itk::RandomImageSource`:
- Exercise basic object methods using the `ITK_EXERCISE_BASIC_OBJECT_METHODS` macro.

Add the accompanying content data link file.

## PR Checklist


jhlegarreta:
Not sure whether `data.kitware` is working as expected. I was unable to get the content file using `git data-upload` because I was unable to authenticate (however, I was able to log in through the web and the API key was fine; set up for development a few times with the same result), and had to get it manually through the web. Now it looks like the content file is not being retrieved:
https://open.cdash.org/viewBuildError.php?buildid=8253637

Any guess?

Modules/Core/TestKernel/test/itkRandomImageSourceValuesTest.cxx
```diff
@@ -0,0 +1,58 @@
...
+
+  const ImageType::SizeType size{ { 10, 10 } };
+  randomImageSource->SetSize(size);
+
+  randomImageSource->SetMin(0.0);
+  randomImageSource->SetMax(20.0);
+
+  randomImageSource->SetNumberOfWorkUnits(1);
+
+  ITK_TRY_EXPECT_NO_EXCEPTION(randomImageSource->Update());
```
dzenanz, line 50: I would prefer if a known seed was explicitly set.

dzenanz (commented):
Did you upload the baseline into your private or public collection?

Modules/Core/TestKernel/test/itkRandomImageSourceValuesTest.cxx
jhlegarreta: There is no seed parameter: according to the [doc](https://github.com/InsightSoftwareConsortium/ITK/blob/f1211c2752fa3bf74ef5a81ea1b47bf151b54125/Modules/Core/TestKernel/include/itkRandomImageSource.h#L45), setting the number of threads/work units to 1 (I guess that is the default value as well, as the image source is also used in other tests) ensures reproducible results.

dzenanz (approved):

PR #3732, opened 2022-11-09
 Default `ChangeInformationImageFilter()` default-constructor, add GTest unit test

N-Dekker:


N-Dekker:
> What motivated you to add this test?

@dzenanz  Good question 😄 I felt triggered by the fact that some elastix code did not want to assume that the OutputDirection of a newly created ChangeInformationImageFilter is an identity matrix. As addressed by https://github.com/SuperElastix/elastix/pull/752 I think it should be considered safe to make that assumption.

thewtex (approved):
:green_heart:

Modules/Filtering/ImageGrid/test/itkChangeInformationImageFilterGTest.cxx
```diff
@@ -0,0 +1,57 @@
...
+    EXPECT_FALSE(newChangeInformationImageFilter->GetCenterImage());
+    EXPECT_FALSE(newChangeInformationImageFilter->GetUseReferenceImage());
+
+    EXPECT_TRUE(newChangeInformationImageFilter->GetOutputDirection().GetVnlMatrix().is_identity());
+
+    const auto expectAllEqualTo = [](const auto & range, const auto expectedValue) {
+      // Expects that all elements are equal to the expected value.
+      for (const auto element : range)
+      {
+        EXPECT_EQ(element, expectedValue);
```
blowekamp, line 46: This is an interesting way to do vector comparisons with nested lambda functions. 

One down side to this approach is that the produced GTest message will not be helpful in tracing the failed comparison.

Does the existing ITK_EXPECT_VECTOR_NEAR work?
Please consider improving and reusing GTest Predicates in itkGTestPredicate.h.

N-Dekker: Interesting suggestion, @blowekamp Thanks!

The existing ITK_EXPECT_VECTOR_NEAR does not _exactly_ do the same as the little `expectAllEqualTo` lambda I'm proposing here. You see `expectAllEqualTo` compares each element to one specific expected element value. And it expects them to be _exactly_ equal. I agree with you that the produced GTest messages from `expectAllEqualTo` may be less clear, although it won't be "rocket science" to figure out what went wrong (in case of a failure). And in practice I don't "expect"  it to _ever_ fail anyway, for this particular test.

We could add an extra macro, like `ITK_EXPECT_ALL_EQUAL_TO(rangeOfValues, expectedValue)`. But for the moment, I think the current PR is good enough. Right? Then maybe extend "itkGTestPredicate.h" later...

blowekamp: Yes, I agree that the current PR is fine.

I am just suggesting creating reusable gtest predicates, in stead of lost if little comparison code every where.

The ITK_EXPECT_VECTOR_NEAR could work good enough with a rms error of ~0, and a vector type constructed with the scalar for all values.

hjmjohnson (approved):

PR #3471, opened 2022-06-22
WIP: ENH: Add initial CiRun configuration for ARM build

thewtex:

PR #3737, opened 2022-11-11
DOC: sudo required in ARM wheel builds

thewtex:
May be related to manylinux image now being based on Alpine.


dzenanz (approved):

thewtex:
@tbirdso FYI, this is now required. However, I run into Eigen build errors.

Could you please take a look at rebasing https://github.com/InsightSoftwareConsortium/ITK/pull/3471 , get it green, so we can keep it green and not run into this?

tbirdso (approved):

tbirdso:
Yes, will take a look

PR #3742, opened 2022-11-15
ENH: Add check if optimizer supports scales

hjmjohnson:
Not all optimizers support scale weights.  The
itkImageRegistrationMethodv4 was unconditionally
setting the scales and causing a confusing warning message to be presented (identified in python, but also present in c++).  Providing an overridable member function that returns the support for weighting scales allow suppression of the warnings.

## PR Checklist
- [X] [API changes] Added new element to allow checking if optimizer supports scales.  Defaults to true.

### DIFF
--- a/Modules/Numerics/Optimizers/include/itkLBFGSBOptimizer.h
+++ b/Modules/Numerics/Optimizers/include/itkLBFGSBOptimizer.h
@@ -194,6 +194,13 @@ public:
   const std::string
   GetStopConditionDescription() const override;
 
+  /** Returns false unconditionally because LBFGSBOptimizer does not support using scales. */
+  bool
+  CanUseScales() const override
+  {
+    return false;
+  }
+
 protected:
   LBFGSBOptimizer();
   ~LBFGSBOptimizer() override;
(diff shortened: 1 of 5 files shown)


Modules/Numerics/Optimizers/include/itkLBFGSBOptimizer.h
```diff
@@ -194,6 +194,13 @@ class ITKOptimizers_EXPORT LBFGSBOptimizer : public SingleValuedNonLinearVnlOpti
   const std::string
   GetStopConditionDescription() const override;
 
+  /** Returns false unconditionally because LBFGSBOptimizer does not support using scales. */
+  bool
+  GetCanUseScales() const override
```
N-Dekker, line 199: Thanks Hans! BTW Why do you call it  **Get**CanUseScales()?  Why not just `CanUseScales()`?

hjmjohnson: Thanks for the input. Changed based on your suggestion.

dzenanz (approved):

thewtex (approved):

PR #3756, opened 2022-11-22
BUG: fixed memory leak in itkMeshTest

issakomi:
Fixed memory leak in _itkMeshTest_
https://open.cdash.org/viewDynamicAnalysisFile.php?id=9809009

Re-using of the mesh after this [line](https://github.com/InsightSoftwareConsortium/ITK/blob/ce57f309f4f4142b80367fd89a88f8dac9d6943a/Modules/Core/Mesh/test/itkMeshTest.cxx#L267) caused the memory leak. The cells data was not deleted.

The change is minimal even if it is difficult to see (I split into 2 local scopes, this shifted most lines, use "hide whitespace").


jhlegarreta (approved):

thewtex (approved):
@issakomi thank you!! 💚

PR #3765, opened 2022-11-24
ENH: Update remote modules for 2022-11-24

thewtex:


jhlegarreta (approved):
I'd have loved to see at least the subject in the commit message body.

jhlegarreta:
> @jhlegarreta what do you mean?

The commit message body is empty: there is only the subject. I know the subject is self-speaking in this case, but we tend to require or [strongly recommend](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#commit-messages) to spare some words in the message body. Do you agree?

thewtex:
Yes, in general a commit message body is helpful, but there is not too much to add the provides unavailable information in cases like this.

That said, I added the log from the script output.

dzenanz (approved):

jhlegarreta:
Thanks  Matt.

PR #3768, opened 2022-11-25
Fix new warning STL4037 raised by VS2022

dzenanz:
Example error message:

`M:\Dashboard\ITK\Modules\Core\Common\src\itkNumericTraits.cxx(77,85): warning C4996: 'std::complex<char>::complex': warning STL4037: The effect of instantiating the template std::complex for any type other than float, double, or long double is unspecified. You can define _SILENCE_NONFLOATING_COMPLEX_DEPRECATION_WARNING to suppress this warning. [M:\Dashboard\ITK-build\Modules\Core\Common\src\ITKCommon.vcxproj]`

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkNumericTraits.h
+++ b/Modules/Core/Common/include/itkNumericTraits.h
@@ -31,6 +31,7 @@
 
 #include <limits> // for std::numeric_limits
 #include <complex>
+#include <type_traits> // for std::is_floating_point
 
 namespace itk
 {
@@ -1979,6 +1980,12 @@ public:
     }
     m = NumericTraits<ValueType>::ZeroValue();
   }
+
+#if defined(ITK_LEGACY_REMOVE)
+  static_assert(std::is_floating_point<TComponent>::value,
+                "As per https://en.cppreference.com/w/cpp/numeric/complex the behavior is unspecified and may fail to "
+                "compile if TComponent is not float, double, or long double and undefined if T is not NumericType.");
+#endif // defined(ITK_LEGACY_REMOVE)
 };
 /// \endcond
 } // end namespace itk
--- a/Modules/Core/Common/test/itkNumericTraitsTest.cxx
+++ b/Modules/Core/Common/test/itkNumericTraitsTest.cxx
@@ -316,6 +316,7 @@ CheckAllSignedAndIntegerTraits()
   std::cout << "\tThis first one should fail" << std::endl << std::endl;
   didAllTestsPass &= !CheckSignedAndIntegerTraitsForComplexTypes<std::complex<ForcedFailureTestCase>>(
     "std::complex< ForcedFailureTestCase >");
+#if !defined(ITK_LEGACY_REMOVE)
   didAllTestsPass &= CheckSignedAndIntegerTraitsForComplexTypes<std::complex<UnknownTypeTestCase>>(
     "std::complex< UnknownTypeTestCase >");
   didAllTestsPass &= CheckSignedAndIntegerTraitsForComplexTypes<std::complex<char>>(" std::complex< char > ");
@@ -331,6 +332,7 @@ CheckAllSignedAndIntegerTraits()
   didAllTestsPass &=
     CheckSignedAndIntegerTraitsForComplexTypes<std::complex<unsigned long>>(" std::complex< unsigned long > ");
   didAllTestsPass &= CheckSignedAndIntegerTraitsForComplexTypes<std::complex<float>>(" std::complex< float > ");
+#endif // !defined(ITK_LEGACY_REMOVE)
   didAllTestsPass &= CheckSignedAndIntegerTraitsForComplexTypes<std::complex<double>>(" std::complex< double > ");
   didAllTestsPass &=
     CheckSignedAndIntegerTraitsForComplexTypes<std::complex<long double>>(" std::complex< long double > ");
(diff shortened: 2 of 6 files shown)


dzenanz:
After doing this, I realized it might have been better to have put integral complex types under legacy ifdefs. What do think? The same thing exists in VXL, at `Modules/ThirdParty/VNL/src/vxl/core/vnl/vnl_complex_traits.h`.

dzenanz:
I changed one of my builds to use VS2022 today: [Windows11-VS22x64-RelWithDebInfo-Favorite-Remotes](https://open.cdash.org/build/8296517).

Modules/Filtering/ImageFrequency/test/itkFrequencyFFTLayoutImageRegionIteratorWithIndexTest.cxx
```diff
@@ -263,7 +263,7 @@ itkFrequencyFFTLayoutImageRegionIteratorWithIndexTest(int, char *[])
   constexpr unsigned int Dimension = 3;
 
   using CharPixelType = char;
-  using FloatPixelType = char;
+  using FloatPixelType = float;
```
jhlegarreta, line 266: I'd dare to say that this change is unrelated to the `std::complex` warning. If that is the case, can it please be put into a separate commit so that the corresponding warning is clear, or even better, a separate PR so that it can be merged independently?

jhlegarreta (commented):
Looks good to me. I'd prefer the FFT change to be put into a separate commit/PR if it is unrelated to the `std::complex` warning. Thanks.

Modules/Core/Common/include/itkNumericTraits.h
```diff
@@ -1979,6 +1980,13 @@ class NumericTraits<std::complex<TComponent>>
     }
     m = NumericTraits<ValueType>::ZeroValue();
   }
+
+#ifdef ITK_USE_CONCEPT_CHECKING
+  static_assert(std::is_same<TComponent, float>::value || std::is_same<TComponent, double>::value ||
+                  std::is_same<TComponent, long double>::value,
+                "As per https://en.cppreference.com/w/cpp/numeric/complex the behavior is unspecified and may fail to "
+                "compile if TComponent is not float, double, or long double and undefined if T is not NumericType.");
```
N-Dekker, line 1988: OK, but I'd rather just do `static_assert(std::is_floating_point<TComponent>::value, "...")`

Modules/Core/Common/test/itkNumericTraitsTest.cxx
```diff
@@ -364,12 +350,17 @@ CheckIsComplexTraits()
                 "std::complex<float> does not have the correct IsComplex trait");
   static_assert(itk::NumericTraits<std::complex<double>>::IsComplex,
                 "std::complex<double> does not have the correct IsComplex trait");
+  static_assert(itk::NumericTraits<std::complex<long double>>::IsComplex,
+                "std::complex<long double> does not have the correct IsComplex trait");
```
N-Dekker, line 354: Should we really care about `long double` support? On MSVC, `long double` is basically just like `double`: 64-bits floating point.

Modules/Core/Common/include/itkNumericTraits.h
dzenanz: Good suggestion, will do.

Modules/Core/Common/test/itkNumericTraitsTest.cxx
dzenanz: It is rarely used nowadays indeed. But it is still an existing type, so we can keep supporting it.

Modules/Filtering/ImageFrequency/test/itkFrequencyFFTLayoutImageRegionIteratorWithIndexTest.cxx
dzenanz: I put it into a separate commit.

Modules/Core/Common/test/itkNumericTraitsTest.cxx
N-Dekker: OK, thanks.

N-Dekker:
For my understanding, it the idea to keep supporting `complex<int>`, _unless_ ITK_USE_CONCEPT_CHECKING is enabled?

dzenanz:
Correct. My [question](https://github.com/InsightSoftwareConsortium/ITK/pull/3768#issuecomment-1327869271) was should we switch from `CONCEPT_CHECKING` to `LEGACY`? Or maybe just outright remove it, as its behavior is not well defined per C++ standard.

N-Dekker:
@dzenanz Sorry, I accidentally edited your comment, instead of replying to it!

blowekamp:
I recall wanting to remove them a while ago because the non-real types were undefined. The only really issue would be the "One" and "Zero" symbols in the library. However, the `OneValue` and `ZeroValue` methods are the preferred approach now, so it is not as significant.

I agree that the complex number of integer usage should be LEGACY, and not dependent on concept checking.

dzenanz:
I updated the PR to use the standard `ITK_LEGACY_REMOVE` CMake option.

jhlegarreta:
If I understand well, concept checking will refuse to have integer values represented as complex numbers if left and merged, which may break some people's code. Even if it has unpredictable results, marking it as legacy code seems more appropriate to me, and concept checking adds another layer of protection that we eventually would be willing to remove anyways.

Modules/Core/Common/test/itkNumericTraitsTest.cxx
```diff
@@ -1273,6 +1281,7 @@ itkNumericTraitsTest(int, char *[])
   CheckFixedArrayTraits(std::complex<unsigned int>());
   CheckFixedArrayTraits(std::complex<long>());
   CheckFixedArrayTraits(std::complex<unsigned long>());
+#  endif // !ITK_USE_CONCEPT_CHECKING
```
jhlegarreta, line 1284: endif // !ITK_LEGACY_REMOVE


Or remove the comment altogether.

dzenanz:
I just created #3775 to keep track of my though about it.

dzenanz:
The latest force-push adds an entry in migration guide about this change.

PR #3781, opened 2022-12-02
WIP: BUG: Fixed issue with 4th component in FreeSurfer ASCII mesh

### LINKED ISSUE #3772: Memory leak in `itkFreeSurferMeshIOTestBinary`
### Description

The dashboard is reporting a memory leak in the `itkFreeSurferMeshIOTestBinary`  test:
https://open.cdash.org/viewDynamicAnalysisFile.php?id=9817770

The leak has been there ever since the test was added in:
https://github.com/InsightSoftwareConsortium/ITK/commit/0fcfaf1288928a76c3ef2b3fcc8b6e53246bfbba

### Steps to Reproduce

1. Have Valgrind or a dependable dynamic analysis tool.
2. Run the `itkFreeSurferMeshIOTestBinary` test: https://github.com/InsightSoftwareConsortium/ITK/blame/master/Modules/IO/MeshFreeSurfer/test/CMakeLists.txt#L18
3. Check the memory leak.

### Expected behavior

No memory leaks are present.

### Actual behavior

The following leak is reported:
```
UMR ==24113== Syscall param writev(vector[...]) points to uninitialised byte(s)
(...)
==24113==    by 0x50D38D2: itk::ByteSwapper::SwapWrite4Range(void const*, unsigned long, std::ostream*) (itkByteSwapper.hxx:367)
==24113==    by 0x50D07A7: itk::ByteSwapper::SwapWriteRangeFromSystemToBigEndian(unsigned int const*, int, std::ostream*) (itkByteSwapper.hxx:160)
==24113==    by 0x50D1FD9: void itk::FreeSurferBinaryMeshIO::WriteCells(unsigned int*, std::basic_ofstream >&) (itkFreeSurferBinaryMeshIO.h:147)
==24113==    by 0x50CF1E2: itk::FreeSurferBinaryMeshIO::WriteCells(void*) (itkFreeSurferBinaryMeshIO.cxx:481)
==24113==    by 0x144C45: int itkFreeSurferMeshIOTestHelper(itk::FreeSurferBinaryMeshIO::Pointer, itk::FreeSurferBinaryMeshIO::Pointer, char*, char*, char*, char*, bool, bool, bool, bool, bool, itk::FreeSurferBinaryMeshIO::SizeValueType, itk::FreeSurferBinaryMeshIO::SizeValueType, itk::FreeSurferBinaryMeshIO::SizeValueType, itk::FreeSurferBinaryMeshIO::SizeValueType) (itkFreeSurferMeshIOTest.cxx:118)
==24113==    by 0x13C881: itkFreeSurferMeshIOTest(int, char**) (itkFreeSurferMeshIOTest.cxx:210)
```

### Reproducibility

%100.

### Versions

`master`

### Environment

Any.

### Additional Information

- The leak in `FreeSurferMeshIO` is only present for the BINARY case in the dashboard.

- I am not persuaded that the problem lies in `itk::ByteSwapper` as the leak's call stack reports. The previous call points to `itk::FreeSurferBinaryMeshIO::WriteCells(void*) (itkFreeSurferBinaryMeshIO.cxx:481)`.

- The `itkBYUMeshIOTest`, `itkOFFMeshIOTest`, `itkVTKPolyDataMeshIOTest` and `itkOBJMeshIOTest` do the same thing: fist use the cell buffer to test the exception cases, then use it to write to the output file.

The only things that I can think of:
- The file is open within some methods but is not closed at the end of such methods and vice-versa (e.g. https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/IO/MeshFreeSurfer/src/itkFreeSurferBinaryMeshIO.cxx#L244). But the same happens in other MeshIO classes (e.g. https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/IO/MeshFreeSurfer/src/itkFreeSurferAsciiMeshIO.cxx#L91 does not call `CloseFile();` or still https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/IO/MeshOFF/src/itkOFFMeshIO.cxx#L98). However, the leak does not point to one of such methods, and he output stream is effectively closed in https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/IO/MeshFreeSurfer/src/itkFreeSurferBinaryMeshIO.cxx#L530
- The template argument for `make_unique_for_overwrite` is not the appropriate one. Not sure whether it should be the class template typename `T` everywhere.

Not sure if the MSVC memory leak detector is dependable, as it shows me a leak even if I remove all contents of the test and leave only the instantiation.
https://learn.microsoft.com/en-us/visualstudio/debugger/finding-memory-leaks-using-the-crt-library?view=vs-2022

I have been willing to resolve this since some time/unwilling to open an issue, but I have spent time trying to investigate this. Having the above issue with the MSVC leak detector I am not able to effectively test my guesses, and the apparently incoherent file opening/closing patterns across the files but the absence of more leaks adds some confusion.

Suggestions are welcome.

issakomi:
S. explanation https://github.com/InsightSoftwareConsortium/ITK/issues/3772#issuecomment-1334689836.

Fixed 4th element.

Tested with FreeSurfer 7.3.2, 2022/08/04


issakomi:
https://youtu.be/Pylza3ePZTs

issakomi:
Not sure how to handle arbitrary extension for binary meshes. Is it legal to have in constructor:

```cpp
  this->AddSupportedWriteExtension(".fsb");
  this->AddSupportedWriteExtension(".fcv");
```

and

```cpp
bool FreeSurferBinaryMeshIO::CanWriteFile(const char * fileName)
{
  return true;
}
```

or maybe additionally exclude some extensions supported by FreeSurfer:
```
These file formats are supported:
  ASCII:       .asc
  ICO:         .ico, .tri
  GEO:         .geo
  STL:         .stl
  VTK:         .vtk
  GIFTI:       .gii
  MGH surface-encoded 'volume': .mgh, .mgz
Freesurfer binary format assumed for all other extensions.
```
?

dzenanz:
> Is it legal

I think it is. @blowekamp do you agree?

BUT, that format will be used by almost all writes, instead of other formats. Let's say I want to create new mesh IO module which will support `*.dzz` format - it will have to fight with FreeSurfer for that extension. Not to mention other well known formats such as `*.obj`.

> Freesurfer binary format assumed for all other extensions.

This is fine for a FreeSurfer tool, not quite for a general purpose library such as ITK. I suggest to have just a few extensions for FreeSurfer, and return false for all other cases in `CanWriteFile`. I think that should allow users to manually set FreeSurfer IO if they want to write into a file with an arbitrary extension.

blowekamp:
> I think it is. @blowekamp do you agree?
> 
> BUT, that format will be used by almost all writes, instead of other formats. Let's say I want to create new mesh IO module which will support `*.dzz` format - it will have to fight with FreeSurfer for that extension. Not to mention other well known formats such as `*.obj`.

Does the MeshIO automatically determine the IO class for reading and writing like the ImageIO? If so accepting all files for CanWrite would cause priority difficulties depending on the order of the MeshIO registered. For example if *.vtk is supported by freesurfer and meshvtk, then the order the factories are registered will determine which meshIO is used. Worse if freesurfer cannot read the file, when it says it can it could take higher priority that IOs which can actually read the file.

I'd recommend just using the "supported" extensions for the CanWrite method.

If the freesurfer IO is wanted to be used for another format it can be explicitly test as the MeshIO for the reader/writer.

issakomi:
> If so accepting all files for CanWrite would cause priority difficulties depending on the order of the MeshIO registered.

Thanks. Yes. I will not change CanWrite.

jhlegarreta:
Sorry for the delay. Thanks for having investigated all issues and proposing a fix to them @issakomi :100:. I'm not an expert on FS file formats :pensive:.

For the current state of things:

- Am fine concerning @blowekamp's and @dzenanz's comments on the extensions. Have not used enough ITK in recent times so as to have an opinion on this.
- OK as for the `.fsa` warning, and the addition of `.asc`.
- However, I am not sure what is the best way (deprecation, etc.) to gradually remove the `.fsa` warning to only refuse to read/write them, as the code does not essentially change. @dzenanz @thewtex @blowekamp ?
- Should the [`label` argument](https://github.com/InsightSoftwareConsortium/ITK/pull/3781/files#diff-27aab81c47286b9c8caf6090f50d7a5b5d4c2de080653ce74b7047c49aaf9762R118) be an `int`, or should it be a `bool` (as initially suggested in the picture in https://github.com/InsightSoftwareConsortium/ITK/issues/3772#issuecomment-1334689836)?
- Whether it should be a `bool` or an `int`, it changes the API so a [migration guide entry](https://github.com/dzenanz/ITK/blob/05c5a8b798dc34c3cf721a18bc93b1fdc16b511f/Documentation/ITK5MigrationGuide.md) is necessary I guess. Not sure if something more would be required, as the [removal of the switch cases](https://github.com/InsightSoftwareConsortium/ITK/pull/3781/files#diff-cd0d27a112f2bd3777efa999cb6956829fd43129c90aa4ad2346e376b1d998e3L410) would also mean that if a third-party code contains such cases, they'd be affected (not sure if their code would work after Mihail's investigation in the related issue, though). I don't know how these changes should be dealt with :pensive:.

In the mid term:
- As for the `.fsb` extension being also weird for FS, not sure whether using e.g. `.stl`, `.vtk`, or `.gii` can then be considered in the tests instead of the `.fsb` extension.

In the longer run:
- @issakomi Many thanks as well for having investigated the mesh/cell vs. curvature/point data distinction (https://github.com/InsightSoftwareConsortium/ITK/issues/3772#issuecomment-1334689836). I guess we will need to e.g. add a flag to say that the file to be tested is a mesh vs. curvature file so that the corresponding methods are called for that type of file. Can be done in a separate PR. 
- Removing the write/read cycle from the test is fine for me if it makes things easier/allows to go forward at this state. Eventually we will be willing to test the methods somehow; can be done in a separate PR; better to have a working version for now.

dzenanz:
@hjmjohnson @seanm @gdevenyi do you know anyone from https://github.com/freesurfer/freesurfer? It would be good if someone more familiar with the file format would provide some feedback.

issakomi:
> Whether it should be a bool or an int, it changes the API so a [migration guide entry](https://github.com/dzenanz/ITK/blob/05c5a8b798dc34c3cf721a18bc93b1fdc16b511f/Documentation/ITK5MigrationGuide.md) is necessary I guess.

the [template](https://github.com/InsightSoftwareConsortium/ITK/blob/59456050a2161a4a6a43913d2a6f85219e0aa67a/Modules/IO/MeshFreeSurfer/include/itkFreeSurferAsciiMeshIO.h#L118) is _protected_, probably there is no API change from user perspective

> Not sure if something more would be required, as the [removal of the switch cases](https://github.com/InsightSoftwareConsortium/ITK/pull/3781/files#diff-cd0d27a112f2bd3777efa999cb6956829fd43129c90aa4ad2346e376b1d998e3L410) would also mean that if a third-party code contains such cases, they'd be affected (not sure if their code would work after Mihail's investigation in the related issue, though).

cell components are indices, _float_, _double_, _long double_ don't make sense and will crash FreeView.

issakomi:
> cell components are indices, float, double, long double don't make sense and will crash FreeView.

I have reverted this to simplify the discussion.

dzenanz (approved):

issakomi:
> Edit: We will also be willing to clarify the cell vs. point distinction, and the mentioned order in the documentation so that the class methods are called appropriately. Can definitely be done in a separate PR.

AFAIK, _Update_ seems to work. Probably it is the [recommended way](https://github.com/InsightSoftwareConsortium/ITK/blob/18a672368d7d5bd7778e5d61f25fc550ce1a5b4e/Modules/IO/MeshBase/include/itkMeshFileTestHelper.h#L250): just set file name, (input) for mesh reader/writer and _Update_, BTW.

issakomi:
The class is backward compatible with old buggy files with wrong 4th components e.g.
`0.000000  0.850651  0.525731  0.000000`
They still could be read.

Modules/IO/MeshFreeSurfer/src/itkFreeSurferAsciiMeshIO.cxx
```diff
@@ -147,6 +162,12 @@ FreeSurferAsciiMeshIO::ReadPoints(void * buffer)
 
   // Read points
   m_InputFile.precision(12);
+
+  // The 4th value of a point should be integer, usually 0.
+  // Here it is of type 'float' for backward compatibility
+  // with buggy files where the 4th value is 0.000000.
+  // Correctly written files with integer 4th value will
+  // work too.
   float value;
```
issakomi, line 171: Read the comment above why i have preserved `float value;` here.

dzenanz (approved):

jhlegarreta (approved):
OK, thanks @issakomi. For now, it's better to have this fixed, and to let the coverage improvement for later.

Edit: I've updated the PR message to avoid closing #3772 with this PR: the right PR will be the one that will fix the binary version.

issakomi:
> Edit: I've updated the PR message to avoid closing #3772 with this PR: the right PR will be the one that will fix the binary version.

What fix do you mean? The #3772 is the issue with the test, not with the binary mesh class.

issakomi:
I shall check `asc` extension once again. The problem is that there are other formats in FreeSurfer with the same extensions, generally any ASCII file with whatever inside can have "asc" extension. Bad. Probably it was the reason that ITK introduced own extensions. Other files (not mesh data) with asc extension can crash the ascii mesh class. Maybe i shall revert the part with "asc" extension and let it be as before or find a way to prevert a crash. It is very difficult with FreeSurfer and extensions.

jhlegarreta:
> What fix do you mean? The https://github.com/InsightSoftwareConsortium/ITK/issues/3772 is (or was) the issue with the test, not with the binary mesh class.

@issakomi you're right, my mistake. Added it back. Got confused by the focus on the ASCII files and the commit and PR messages not mentioning that the method calls in the test were modified to fix the issue.

jhlegarreta:
> I shall check asc extension once again. The problem is that there are other formats in FreeSurfer with the same extensions, generally any ASCII file with whatever inside can have "asc" extension. Bad. Probably it was the reason that ITK introduced own extensions. Other files (not mesh data) with asc extension can crash the ascii mesh class. Maybe i shall revert the part with "asc" extension and let it be as before or find a way to prevert a crash. It is very difficult with FreeSurfer and extensions. To load ASCII mesh into FreeSurfer they must have "asc" extension, otherwise FreeView crash, but we can not assume that any file with "asc" extension is a mesh and will work with ITK. Oh.

Maybe we should separate the fix to the memory leak in the test and fixing the ASCII format? That way at least the fix to the leak will be eventually merged without being blocked by the fix to the ASCII format.

jhlegarreta:
> If you like. I stuck a little with the PR. There are many things I don't like in both classes, e.g. switches over data types that can not work, this 'curvature data' stuff (where is cell type?). Also MeshIOBase could be a little bit bloated with cell/point pixel (?) types like RGBA, tensor, etc. But i don't want to step into this stuff and make so many big changes.

I think best is to split, have the memory leak issue resolved and merged, and then go to tackle the other issues incrementally. As you found the fix, I'd prefer if you would submit it to credit you as the author.

issakomi:
> I think best is to split, have the memory leak issue resolved and merged, and then go to tackle the other issues incrementally. As you found the fix, I'd prefer if you would submit it to credit you as the author.

S. https://github.com/InsightSoftwareConsortium/ITK/pull/3796

issakomi:
> S. https://github.com/InsightSoftwareConsortium/ITK/pull/3796

The Valgrind defect is gone:
https://open.cdash.org/index.php?project=Insight

I have updated this PR to **only** change 4th component now. Didn't change extension.
Replaced _sphere.fsa_ with _sphere.asc.fsa_, produces with FreeSurfer's `mris_convert sphere.fsb sphere.asc`, but old files with 4th component `0.000000` will work too.
I stuck somehow with FreeSurfer files and mesh IO in general, too many questionable stuff. So the PR is still draft.

jhlegarreta:
> The Valgrind defect is gone:
> https://open.cdash.org/index.php?project=Insight&date=2022-12-15

This is fantastic @issakomi ! A big thank you.

> I stuck somehow with FreeSurfer files and mesh IO in general, too many questionable stuff. So the PR is correct, but still draft, I am not sure what to do with the rest.

I see your feeling, but when we feel the changes needed are big, and we do not have time to properly do them all at once, or have a somewhat incomplete picture, if possible, it is better to split them and do small steps towards the solution.

e.g. you also did some changes to the binary FS mesh IO class in an earlier commit:
https://github.com/InsightSoftwareConsortium/ITK/commit/51832791689480af7dc236e8a95e6e073b890f30#diff-6484fbca670f039ecbb0541b42332f2b0b9844b719c42327b38d90c326335692

If you are more confident with these, can they be put to a separate PR, and make the appropriate changes to the test data/files so that the difference gets tested, fixed and merged independently?

Thanks for all this effort. I know tracking down some stuff can consume a lot of time and energy.

issakomi:
> e.g. you also did some changes to the binary FS mesh IO class in an earlier commit:

The only significant change was to modify _CanReadFile_ to use magic number instead of extension
```cpp
  if (fileTypeIdentifier == (-2 & 0x00ffffff))
  {
    // The input file is FreeSurfer binary surface file
    ok = true;
  }
  else if (fileTypeIdentifier == (-1 & 0x00ffffff))
  {
    // The input file is FreeSurfer binary curvature file
    ok = true;
  }
```

The problem is I think that a curvature file something very special. In ASCII format it looks like:

```
000 -11.15284 -100.95035 -0.40290 2.59229
001 -10.57401 -100.85500 -1.28304 2.75554
002 -10.89847 -101.10920 -1.06143 2.48846
003 -11.44057 -101.25616 -0.94590 2.29264
...
...
...
135503 -31.83741 2.20956 -27.58961 3.48273
135504 -16.78938 -65.03230 1.17469 1.45606
135505 -17.24210 -82.13521 -15.37917 2.26153
135506 -44.70148 43.66516 5.29382 2.99865
```

1st column is the index of the vertex
2nd, 3rd and 4th are coordinates
5th i don't know, a curvature probably :)

I am not sure that it can be used standalone at all, maybe together with other file. Or maybe there different kind of files.
I don't think that binary mesh class works, it reads number of points and cells (?), the numbers is different, how it could be? What kind of cell it should be? Only _Read/WritePointData_ methods are used.
So i decided to do no changes in binary class at.


I am playing with _FreeSurfer_. It is interesting. Shortly, it works like this (after installation, licensing and setting the environment)
Take **T1** MRI and do
`recon-all  -i ~/Downloads/DICOM/t1_gre_fsp_3d_sag__132750/00000001.dcm -subjid uih1 -all`
Give only one DICOM file, the series will be detected, after **_several hours_** there will be new subject (here **uih1**) with  skull strip (very high quality, BTW), segmentation, surfaces and tons of stuff. There are hundreds of tools (recon-all calls them), rather complicated. There are also videos and tutorials. I am not yet experienced with.

jhlegarreta:
> The only significant change in binary class was the modification of CanReadFile to use magic number instead of extension

:+1:, but we all agree that this is an `ENH:` improvement that we'd like to have: makes the FS binary mesh IO more robust, does not introduce breaking changes, etc. Or am I wrong?

> The problem is I think that a curvature file something very special. In ASCII format it looks like:

OK, but this is separate from the above binary format improvement.

> I am not sure that it can be used standalone at all. I don't think that binary mesh class works for such files (but works well with surfaces), not sure, it reads numbers of points and cells (?), the numbers are different, how it could be? What kind of cell it should be? Only Read/WritePointData methods are used.

I am not very familiar with mesh files. Maybe the curvature file is always accompanied by a vertex file? @PranjalSahu has been working on mesh files lately. Maybe he can cast some light on this.

> So i decided to do no changes in binary class at all.

I see, but we agree that the magic number change is independent of the curvature/point/cell data issue, right?

> There are hundreds of tools (recon-all calls them), rather complicated. There are also videos and tutorials.

I know FS does a lot of processing and generates a lot of files, and that it requires several hours of computation for a single subject. But these people have worked hard to develop the tool and make it available. I have not played with [FastSurfer](https://deep-mi.org/research/fastsurfer/), and IDK if it computes the same features/generates the exact same number of files, but might be an option if we want to get the output fast for experimenting with the files.

issakomi:
> Maybe the curvature file is always accompanied by a vertex file?

Yes, it can not be used standalone. There are just _float_ values, one per vertex. In fact there are different types: curvature, thickness, weights, etc. In Slicer's FreeSurfer importer plugin it is called _overlay_ file. Also binary and ASCII formats are different, binary has a kind of header: magic number, number of points, number of cells (unused), number of components (always 1) and a float array of values. I don't know how it is supposed to be used in binary mesh class standalone, it is not a mesh, probably there could be a method, something like _SetOverlayFile_. In ASCII class it is not available, the related methods (for point data) are empty. Generally ASCII files are not native, FreeSurfer doesn't produce them, actually they can be only created with `mris_convert` (`asc` extension signals that output should be in ASCII file).

jhlegarreta:
Looks like this is loosely related: https://github.com/InsightSoftwareConsortium/ITKIOMeshSTL/issues/44

PR #3783, opened 2022-12-04
STYLE: Do not suppress clang "-Wfloat-equal" inside `itkSetObjectMacro`

N-Dekker:
The macro definition does not do any floating point comparison.

### DIFF
--- a/Modules/Core/Common/include/itkMacro.h
+++ b/Modules/Core/Common/include/itkMacro.h
@@ -1089,7 +1089,6 @@ compilers.
 // clang-format on
 
 // clang-format off
-// NOTE: warning: comparing floating point with == or != is unsafe [-Wfloat-equal]
 /** Set pointer to object; uses Object reference counting methodology.
  * Creates method Set"name"() (e.g., SetPoints()). Note that using
  * smart pointers requires using real pointers when setting input,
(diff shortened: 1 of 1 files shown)


Modules/Core/Common/include/itkMacro.h
```diff
@@ -1098,14 +1097,11 @@ compilers.
   virtual void Set##name(type * _arg)                  \
   {                                                    \
     itkDebugMacro("setting " << #name " to " << _arg); \
-    CLANG_PRAGMA_PUSH                                  \
-    CLANG_SUPPRESS_Wfloat_equal                        \
     if (this->m_##name != _arg)                        \
```
dzenanz, line 1100: If `type` is float, this will be a floating point comparison. And this macro is used by remote modules and other external code, not just by ITK proper.

dzenanz (commented):
I am not sure whether we should do this.

Modules/Core/Common/include/itkMacro.h
N-Dekker: > If `type` is float, this will be a floating point comparison.

@dzenanz Thanks, but please double-check!  The argument `_arg` is declared as `type *`, so if `type` is float, then `_arg` is a _float pointer_, not a float.  The comparison `this->m_##name != _arg` _always_ does pointer comparison, with this specific macro.

dzenanz (approved):

Modules/Core/Common/include/itkMacro.h
dzenanz: I indeed failed to notice is it a float pointer 😄

PR #3792, opened 2022-12-08
STYLE: Replace `for` loops SetPointIds(PointIdConstIterator) by copy_n

N-Dekker:
Replaced manual `for` loops inside the `SetPointIds(PointIdConstIterator)` member functions of ten "Cel" types by equivalent `std::copy_n` calls. Addresses a few warnings from Clang Tidy (LLVM 15.0.4), saying:

> warning: use range-based for loop instead [modernize-loop-convert]

Note that in these cases, using `std::copy_n` appears even better than using range-based `for` loops!

### DIFF
--- a/Modules/Core/Common/include/itkHexahedronCell.hxx
+++ b/Modules/Core/Common/include/itkHexahedronCell.hxx
@@ -21,6 +21,8 @@
 #include "vnl/vnl_matrix_fixed.h"
 #include "vnl/algo/vnl_determinant.h"
 
+#include <algorithm> // For copy_n.
+
 namespace itk
 {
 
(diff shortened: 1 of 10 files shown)


dzenanz (approved):

thewtex (approved):
Nice!!

Is this a `PERF` instead of `STYLE`?

dzenanz:
> Is this a PERF instead of STYLE?

Doubtful. Compilers should be able to optimize a for loop nowadays.

N-Dekker:
> Is this a `PERF` instead of `STYLE`?

Thanks for asking Matt. std::copy_n _might_ be optimized for arrays of trivially copyable data, so this commit _might_ improve the performance. But I don't really expect a significant performance difference, honestly.

Honestly I was just trying to fix those warnings from clang-tidy, as they leaked into my local elastix compilation.  So you might also call it a `COMP`. 😃 But I considered it a matter of `STYLE`, as copy_n basically just does the same thing as the original code, but in a more modern way. Reducing the number of lines of code, while still being clear and readable, IMhO.

PR #3802, opened 2022-12-16
STYLE: Chain consecutive insertions (`<<`) to the same output stream

N-Dekker:
Avoiding unnecessarily mentioning one and the same output stream repetitively, when multiple consecutive statements insert into that one output stream.

Using Notepad++ v8.4.7 Regular Expressions, find `^([ ]+[^ ]+) << (.+);\r\n\1 << ` and replace with `$1 << $2 << `.


Modules/Core/Common/include/itkConstNeighborhoodIterator.hxx
```diff
@@ -580,8 +578,8 @@ ConstNeighborhoodIterator<TImage, TBoundaryCondition>::PrintSelf(std::ostream &
   {
     os << m_Region.GetSize()[i] << " ";
   }
-  os << "} }";
-  os << ", m_BeginIndex = { ";
+  os << "} }"
+     << ", m_BeginIndex = { ";
```
dzenanz, line 582: Why didn't this fit onto the previous line?

Modules/Core/Common/include/itkConicShellInteriorExteriorSpatialFunction.hxx
```diff
@@ -107,10 +107,10 @@ ConicShellInteriorExteriorSpatialFunction<VDimension, TInput>::PrintSelf(std::os
   }
   os << "]" << std::endl;
 
-  os << indent << "DistanceMin: " << m_DistanceMin << std::endl;
-  os << indent << "DistanceMax: " << m_DistanceMax << std::endl;
-  os << indent << "Epsilon: " << m_Epsilon << std::endl;
-  os << indent << "Polarity: " << m_Polarity << std::endl;
+  os << indent << "DistanceMin: " << m_DistanceMin << std::endl
+     << indent << "DistanceMax: " << m_DistanceMax << std::endl
```
dzenanz, line 111: This change looks pointless.

Modules/Core/Common/include/itkConstNeighborhoodIterator.hxx
```diff
@@ -568,9 +568,7 @@ ConstNeighborhoodIterator<TImage, TBoundaryCondition>::PrintSelf(std::ostream &
 {
   DimensionValueType i;
 
-  os << indent;
-  os << "ConstNeighborhoodIterator {this= " << this;
-  os << ", m_Region = { Start = {";
+  os << indent << "ConstNeighborhoodIterator {this= " << this << ", m_Region = { Start = {";
```
dzenanz, line 571: I like this!

Modules/Core/Common/include/itkConstNeighborhoodIterator.hxx
N-Dekker: Good question, but that's up to clang-tidy! Even when I remove the line-break, clang-tidy puts it back, in this case!

A follow-up PR could merge it together to `  os << "} }, m_BeginIndex = { "`, of course.  Clang-tidy would allow that on one single line.

Modules/Core/Common/include/itkConicShellInteriorExteriorSpatialFunction.hxx
N-Dekker: It's basically what the PR is doing: avoiding to mention the output stream variable repetitively. 🤷  

The idea of this PR is that rather than Pascal style `WriteLine(str)` statements (which print one line at a time), we could insert multiple lines into one and the same stream, by a single chain of `<<` insertions.

N-Dekker: But for the implementation of all those `PrintSelf` member functions, it might be even nicer to have some `PrintData` function that would write an arbitrary number of name-value pairs to the specified `os`, inserting the specified `indent` before each name-value-pair, e.g.: 

    PrintData(os, indent, { "Epsilon", m_Epsilon }, { "Polarity", m_Polarity }, { "DistanceMin", m_DistanceMin });

Could be implemented as a variadic template, possibly as a member function of `itk::LightObject`. What do you think?

dzenanz: This pattern appears a lot:
```cpp
os << indent << "DistanceMin: " << m_DistanceMin << std::endl;
```
we could have a macro which makes this less prone to errors:
```cpp
#define PrintMember(name) os << indent << #name ": " << m_##name << std::endl ITK_MACROEND_NOOP_STATEMENT
```
which would be invoked like this: `PrintMember(DistanceMin);`.

N-Dekker: Update: with my latest amend, I excluded the `os <<` insertions, which are typically from inside PrintSelf member functions. For PrintSelf a different approach may be more useful, as you suggested. Using a macro could be helpful, although I'd rather not hide away `os` and `indent` behind a macro definition. 

I'm thinking of introducing a `Printer` class, which would support chaining multiple `PrintMember` calls:

    Printer(os, indent).PrintMember("Epsilon", m_Epsilon).PrintMember("Polarity", m_Polarity);

And a `PrintMemberMacro(name)` macro, to shorten the above:

    Printer{os, indent}.PrintMemberMacro(Epsilon).PrintMemberMacro(Polarity);

Note that in some cases, the macro won't work anyway, for example `os << indent << "Pointer: " << static_cast<void *>(m_ImportPointer) << std::endl` at https://github.com/InsightSoftwareConsortium/ITK/blob/da36517a51d92cc38a98c544348115af124e5f8a/Modules/Core/Common/include/itkImportImageContainer.hxx#L216

dzenanz:
I would really like for somebody else to review this too.

Modules/Core/Common/include/itkConstNeighborhoodIterator.hxx
dzenanz: Your force-push removed this change :(

jhlegarreta (approved):
Of note is that going through the modified lines there are some seemingly unnecessary blank spaces (already present prior to this commit), e.g.
https://github.com/InsightSoftwareConsortium/ITK/pull/3802/commits/67ca856a8ced8b0ba481b5d81e2afaf192e90609#diff-40525bc7610ac6ba8f4dfedf16b2fad81460618d8f72bef57dee973cac2079fdR183
https://github.com/InsightSoftwareConsortium/ITK/pull/3802/commits/67ca856a8ced8b0ba481b5d81e2afaf192e90609#diff-f043933bf4ca24b86407d6d96654cedd3bfc7d20a9f4ac3f6c1574d3bccc9f22R114

or inconsistent use of them to align messages in consecutive lines, or the use of colon vs. equal, but that is an issue for a separate PR. The current commit/PR does a nice job for its purpose. Thanks @N-Dekker.

Modules/Core/Common/include/itkConstNeighborhoodIterator.hxx
jhlegarreta: @N-Dekker shall I merge this PR or are there some details that still need another round of force pushes?

N-Dekker: @jhlegarreta Thanks for asking. I think the PR is ready, although it has excluded output streams named `os`, as those are mostly from PrintSelf implementations. PrintSelf may be dealt with in a different way, in another PR, based on the discussion I had with Dženan (@dzenanz), starting at https://github.com/InsightSoftwareConsortium/ITK/pull/3802#discussion_r1051069125

Moreover, I'm not really sure if Dženan likes the idea of this PR. I mean, I'm proposing to adjust the following code:

    std::cout << line1 << std::endl;
    std::cout << line2 << std::endl;
  
Replacing these two statements with a single statement:

    std::cout << line1 << std::endl
              << line2 << std::endl;

I like it, because it does not unnecessarily repeat the name of the output stream. However, other developers may not like it, I don't know. 

It's fine to me if Dženan would like to wait a little longer, to see what other people think. Anyway, thanks for your support, Jon.

dzenanz:
I have minor dislike for these changes. But you liked them enough to go through the trouble of creating a PR, and another maintainer OKed it. No need to wait more - this can be merged. But I won't do it.

Modules/Core/Common/include/itkConstNeighborhoodIterator.hxx
jhlegarreta: OK. I'll merge it, but note that
- Unless this is automated, we will not be able to enforce this automatically. A script to do this automatically would be best.
- Please, go through the ITK SWG and do the corresponding changes to the coding style section. Also it is highly likely that the ITK SWG `Examples` were not included in this PR, so please do them in a separate PR.

jhlegarreta:
> I have minor dislike for these changes. But you liked them enough to go through the trouble of creating a PR, and another maintainer OKed it. No need to wait more - this can be merged. But I won't do it.

:+1: Cross-referencing https://github.com/InsightSoftwareConsortium/ITK/pull/3802#discussion_r1056361996.

Merging.

jhlegarreta:
A new reminder: https://github.com/InsightSoftwareConsortium/ITK/pull/3802#discussion_r1056361996. Can this:
- Be put to a script so that it can run any time/on remotes?
- Be applied to all the code base, including non-default modules and examples.
- Be stated in the ITK SW Guide.

Please.

Modules/Core/Common/include/itkConicShellInteriorExteriorSpatialFunction.hxx
jhlegarreta: This might not be easy but would be amazing :rocket:. Different styles might have been adopted to print e.g. matrices, etc., but we may reach a consensus. And it would definitely make the printed messages consistent: the state is now quite inconsistent, with whitespaces inserted manually (for some visual alignment of the code, and whose impact when printed is maybe unsure), `=` vs. `:`, etc. :pensive:.

N-Dekker:
> A new reminder: [#3802 (comment)](https://github.com/InsightSoftwareConsortium/ITK/pull/3802#discussion_r1056361996). Can this:
> 
>     * Be put to a script so that it can run any time/on remotes?
> 
>     * Be applied to all the code base, including non-default modules and examples.
> 
>     * Be stated in the ITK SW Guide.

Thanks for the reminder, Jon. It's still on my TODO list. But if you want it to be processed more quickly, please go ahead and make a proposal! Otherwise I'll just return to this topic when I have the time.

PR #3804, opened 2022-12-16
DOC: Add links to previous versions in the Doxygen footer

thewtex:
To be updated with each minor release.

### DIFF
--- a/Documentation/Doxygen/DoxygenFooter.html
+++ b/Documentation/Doxygen/DoxygenFooter.html
@@ -13,12 +13,19 @@
 <!--BEGIN !GENERATE_TREEVIEW-->
 <hr class="footer"/>
 <div class="footer" align="left">
-  <small>Tarballs of the nightly generated Doxygen documentation are available
+  <p><small>Tarballs of the nightly generated Doxygen documentation are available
     for the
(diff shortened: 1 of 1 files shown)


jhlegarreta (approved):
Wondering whether there is a way to do this automatically (e.g. to fill the list dynamically).

dzenanz (approved):

blowekamp:
@jhlegarreta  Take a looks at how the date and time are added:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/Doxygen/DoxygenFooter.html#L24
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/Doxygen/build_text.js.in

Something similar could be done for the version so that old Doxygen would be updated with the latest/releases automatically from javascript.

jhlegarreta:
> @jhlegarreta Take a looks at how the date and time are added:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/Doxygen/DoxygenFooter.html#L24
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/Doxygen/build_text.js.in

> Something similar could be done for the version so that old Doxygen would be updated with the latest/releases automatically from javascript.

:+1: Thanks Brad. Filed an issue so that it can be addressed as time permits/by anyone.

Issue #3654, opened 2022-09-25
Doxygen "unbalanced grouping commands" warnings


jhlegarreta:
### Description

The dashboard Doxygen build is showing a number of "unbalanced grouping commands" warnings:
https://open.cdash.org/viewBuildError.php?type=1&buildid=8176854

These issues are also being reported by the nightly ITKDoxygen build:
https://github.com/InsightSoftwareConsortium/ITKDoxygen/actions/runs/3120254526/jobs/5060690356#step:4:2789

### Expected information

The Doxygen build should be clean of warnings.

### Actual information

The groups seem to be present in the impacted classes' Doxygen pages, 

### Versions

`master`

### Additional Information

Related to PR https://github.com/InsightSoftwareConsortium/ITK/pull/3640#issuecomment-1253682434.

The relevant Doxygen source code class/method/line:
https://github.com/doxygen/doxygen/blob/ade61047308990ec679d1d59f0019a825e87451d/src/docgroup.cpp#L142

A few things I've looked into:
- Maybe the groups are not being closed properly in the `CMake` file `defgroup`s somewhere in here:
https://github.com/InsightSoftwareConsortium/ITK/blob/b870edf59bb0347116d245a445a2df8f5c6a8e56/CMake/ITKModuleDoxygen.cmake
https://github.com/InsightSoftwareConsortium/ITK/blob/d12f14e105500362a30264e683a4045eed21ba14/CMake/ITKGroups.cmake
Or else it's maybe a closing `*/` in here?
https://github.com/InsightSoftwareConsortium/ITK/blob/b870edf59bb0347116d245a445a2df8f5c6a8e56/Utilities/Doxygen/itkgroup.pl
e.g.
https://github.com/InsightSoftwareConsortium/ITK/blob/b870edf59bb0347116d245a445a2df8f5c6a8e56/Utilities/Doxygen/itkgroup.pl#L74

- Having a look at the first two warnings, which happen for [`itk::AutoPointer`](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkAutoPointer.h#L41) and [`itk::BackwardDifferenceOperator`](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkBackwardDifferenceOperator.h#L40) 
I don't see any hint in the group/module bar in Doxygen:
[...]

PR #3805, opened 2022-12-19
DOC: Fix unbalanced grouping commands Doxygen warnings

jhlegarreta:
Fix unbalanced grouping commands Doxygen warnings.

Fixes:
```
Modules/Core/Common/include/itkAutoPointer.h:261: warning: unbalanced grouping commands
```

warnings across classes.

The warning is raised when the
Utilities/Doxygen/itkdoxygen.pl

Doxygen Perl script processes the files at issue and starts a `/**@{` block that gets interrupted by an unexpected symbol (e.g. a brace) without being previously closed with the corresponding `/**@}*/` ending token.

Raised for example in:
https://open.cdash.org/viewBuildError.php?type=1&buildid=8344134

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkAutoPointer.h
+++ b/Modules/Core/Common/include/itkAutoPointer.h
@@ -254,6 +254,7 @@ TransferAutoPointer(TAutoPointerBase & pa, TAutoPointerDerived & pb)
     pb.ReleaseOwnership(); // pb Release Ownership and clears
   }
 }
+
 } // end namespace itk
 
 #endif
--- a/Modules/Core/Common/include/itkBSplineDerivativeKernelFunction.h
+++ b/Modules/Core/Common/include/itkBSplineDerivativeKernelFunction.h
@@ -186,6 +186,7 @@ private:
     itkGenericExceptionMacro("Evaluate not implemented for spline order " << SplineOrder);
   }
 };
+
 } // end namespace itk
 
 #endif
--- a/Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h
+++ b/Modules/Core/Common/include/itkBSplineInterpolationWeightFunction.h
@@ -122,6 +122,7 @@ protected:
   BSplineInterpolationWeightFunction() = default;
   ~BSplineInterpolationWeightFunction() override = default;
 };
+
 } // end namespace itk
 
 #ifndef ITK_MANUAL_INSTANTIATION
--- a/Modules/Core/Common/include/itkBSplineKernelFunction.h
+++ b/Modules/Core/Common/include/itkBSplineKernelFunction.h
@@ -180,6 +180,7 @@ private:
     itkGenericExceptionMacro("Evaluate not implemented for spline order " << SplineOrder);
   }
 };
+
 } // end namespace itk
 
 #endif
--- a/Modules/Core/Common/include/itkBackwardDifferenceOperator.h
+++ b/Modules/Core/Common/include/itkBackwardDifferenceOperator.h
@@ -73,6 +73,7 @@ protected:
     this->FillCenteredDirectional(coeff);
   }
 };
+
 } // namespace itk
 
 #ifndef ITK_MANUAL_INSTANTIATION
--- a/Modules/Core/Common/include/itkBufferedImageNeighborhoodPixelAccessPolicy.h
+++ b/Modules/Core/Common/include/itkBufferedImageNeighborhoodPixelAccessPolicy.h
@@ -126,6 +126,7 @@ public:
   {
     m_NeighborhoodAccessor.Set(imageBufferPointer + m_PixelIndexValue, pixelValue);
   }
+
 };
 
 } // namespace itk
--- a/Modules/Core/Common/include/itkCommonEnums.h
+++ b/Modules/Core/Common/include/itkCommonEnums.h
@@ -164,6 +164,7 @@ using IOFileModeType = CommonEnums::IOFileMode;
 using IOByteOrderType = CommonEnums::IOByteOrder;
 using CellGeometryType = CommonEnums::CellGeometry;
 #endif
+
 // Define how to print enumeration
 extern ITKCommon_EXPORT std::ostream &
                         operator<<(std::ostream & out, IOPixelEnum value);
--- a/Modules/Core/Common/include/itkConnectedImageNeighborhoodShape.h
+++ b/Modules/Core/Common/include/itkConnectedImageNeighborhoodShape.h
@@ -83,6 +83,7 @@ namespace itk
  * \ingroup ImageIterators
  * \ingroup ITKCommon
  */
+
 template <unsigned int VImageDimension>
 class ConnectedImageNeighborhoodShape
 {
@@ -247,8 +248,8 @@ private:
   {
     return (includeCenterPixel ? 1 : 0) + CalculateNumberOfConnectedNeighbors(maximumCityblockDistance);
   }
-};
 
+};
 
 /** Generates the offsets for a connected image neighborhood shape. */
 template <unsigned int VImageDimension, size_t VMaximumCityblockDistance, bool VIncludeCenterPixel>
--- a/Modules/Core/Common/include/itkConstantBoundaryImageNeighborhoodPixelAccessPolicy.h
+++ b/Modules/Core/Common/include/itkConstantBoundaryImageNeighborhoodPixelAccessPolicy.h
@@ -144,8 +144,8 @@ public:
       m_NeighborhoodAccessor.Set(imageBufferPointer + m_PixelIndexValue, pixelValue);
     }
   }
-};
 
+};
 
 } // namespace itk
 
--- a/Modules/Core/Common/include/itkContinuousIndex.h
+++ b/Modules/Core/Common/include/itkContinuousIndex.h
@@ -90,6 +90,7 @@ public:
       (*this)[i] = static_cast<TCoordinate>(index[i]);
     }
   }
+
 };
 } // namespace itk
 
--- a/Modules/Core/Common/include/itkCovariantVector.h
+++ b/Modules/Core/Common/include/itkCovariantVector.h
@@ -257,6 +257,7 @@ public:
       (*this)[i] = static_cast<T>(pa[i]);
     }
   }
+
 };
 
 /** Premultiply Operator for product of a vector and a scalar.
--- a/Modules/Core/Common/include/itkDefaultConvertPixelTraits.h
+++ b/Modules/Core/Common/include/itkDefaultConvertPixelTraits.h
@@ -77,6 +77,7 @@ public:
   {
     return pixel.GetScalarValue();
   }
+
 };
 
 #define ITK_DEFAULTCONVERTTRAITS_NATIVE_SPECIAL(type)           \
--- a/Modules/Core/Common/include/itkExtractImageFilterRegionCopier.h
+++ b/Modules/Core/Common/include/itkExtractImageFilterRegionCopier.h
@@ -155,6 +155,7 @@ public:
   {
     ImageRegionCopier<T1, T2>::operator()(destRegion, srcRegion);
   }
+
 };
 } // end namespace ImageToImageFilterDetail
 } // end namespace itk
--- a/Modules/Core/Common/include/itkFixedArray.h
+++ b/Modules/Core/Common/include/itkFixedArray.h
@@ -459,6 +459,7 @@ public:
   {
     return MakeFilled<FixedArray>(value);
   }
+
 };
 
 template <typename TValue, unsigned int VLength>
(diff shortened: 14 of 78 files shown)


jhlegarreta:
I checked the `/**@{` `/**@}*/` blocks for every file manually following https://github.com/InsightSoftwareConsortium/ITK/issues/3654#issuecomment-1353135951.

However, clang already warned me locally that it did not like the changes: it does not like changes in 61 files out of 79. 

@hjmjohnson @thewtex any suggestion on how we could work around this without having to recur to turning it on/off on each of these lines?

dzenanz (approved):

dzenanz:
It feels like we should update [clang format style](https://github.com/InsightSoftwareConsortium/ITK/blob/v5.3.0/.clang-format) to allow the changes proposed here. I am not sure whether that can accomplish it.

dzenanz:
I am running the build script with doxygen updated to 1.9.5, so we don't have to wait until tomorrow to see the updated output.

dzenanz:
The second suggestion would require us to move to a newer `clang-format` (14 or later). We are currently at 8.0.0. Moving to a newer one shouldn't be a big change. The first suggestion looks like it would work now. Can you try it?

dzenanz:
Doxygen 1.9.5 produces 100+ new warnings of style: Generating doc[Documentation/Doxygen/NeighborhoodIterators.dox:96](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/Doxygen/NeighborhoodIterators.dox#L96): warning: Unexpected token TK_HTMLTAG found as part of a title section. All warnings [here](https://open.cdash.org/viewBuildError.php?type=1&onlydeltap&buildid=8345942). @jhlegarreta should we keep this version?

jhlegarreta:
> Doxygen 1.9.5 produces 100+ new warnings of style: Generating doc[Documentation/Doxygen/NeighborhoodIterators.dox:96](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/Doxygen/NeighborhoodIterators.dox#L96): warning: Unexpected token TK_HTMLTAG found as part of a title section. All warnings [here](https://open.cdash.org/viewBuildError.php?type=1&onlydeltap&buildid=8345942). @jhlegarreta should we keep this version?

😓 I would keep the old version then. I would upgrade to the newer version once we have managed to fix the current warnings (which are still present in the new version).

dzenanz:
I just reverted doxygen to 1.8.16.

jhlegarreta:
> The first suggestion looks like it would work now. Can you try it?

Looks like our `clang-format` options prevent it from working as described in the thread: https://github.com/InsightSoftwareConsortium/ITK/actions/runs/3727206860/jobs/6321140027#step:4:86

@hjmjohnson Do you happen to know the `clang-format` option that should do the work for our current version?

dzenanz:
Adding `end of class` comment does trigger the described workaround:
```cpp
  /** Sets the value of the image buffer at the current index value to the
   * specified value.  */
  void
  SetPixelValue(InternalPixelType * const imageBufferPointer, const PixelType & pixelValue) const noexcept
  {
    m_NeighborhoodAccessor.Set(imageBufferPointer + m_PixelIndexValue, pixelValue);
  }

}; // end of class
```
This is an excerpt from ITK\Modules\Core\Common\include\itkBufferedImageNeighborhoodPixelAccessPolicy.h, near the end.

jhlegarreta:
https://github.com/InsightSoftwareConsortium/ITK/pull/3805#issuecomment-1358427743 OK, I was thinking about the namespace ending. I can do this, but I do not think we are willing to have this as a long-term solution.

@thewtex @blowekamp @hjmjohnson any thoughts?

thewtex:
Another approach could replace *itkdoxygen.pl* with a Python script :snake:

jhlegarreta:
> Another approach could replace itkdoxygen.pl with a Python script 🐍

The problem is rather clang-format disliking the insertion of blank spaces before the end of class closing brackets.

thewtex:
> The problem is rather clang-format disliking the insertion of a blank line

Can we remove the insertion of the blank line in the Python script migration?

jhlegarreta:
> Can we remove the insertion of the blank line in the Python script migration?

It should rather be *introduced*: `clang-format` dislikes it and it keeps trying to remove it.

I do not have the bandwidth to translate the script into Python.

Alternatives:
- If you feel like modifying the Perl script to add such a blank line and have this tested, please go ahead :arrow_forward:.
- If you feel like to translating the script to Python and introducing the blank line, please go ahead :arrow_forward:.
- Otherwise, pinging @hjmjohnson to see if he can allow the blank like without changing the script.

jhlegarreta:
Do you think the below approach plugged right after
https://github.com/InsightSoftwareConsortium/ITK/blob/23da3dfa79152648f3d38050ee147df47f41ce0a/Utilities/Doxygen/itkgroup.pl#L67

can be made to work?
```
if($line =~ /\};/ )
{
  // Take the current buffer $buffer (assuming it includes the current line `};` as well) and add a blank line
  // immediately before `};` (if the current buffer does not include the current line, then include it right after
  // the buffer end)

  (doTheAboveStuff)

  // (I think we do not care that much if }; is not the end of the class definition (i.e. even if a lot of blank lines
  // are inserted) as this is only to produce proper Doxygen code)

  (save the $buffer)
}
```

so that we can get the result in https://github.com/InsightSoftwareConsortium/ITK/pull/3805#issuecomment-1358427743 without needing to add `// end of class` and thus bypass clang's unhappyness about adding a blank line? Or maybe there is another way to achieve it

@thewtex @tbirdso maybe one of you is skilled at Perl to give that a try? Thanks.

dzenanz:
I still think that moving to latest stable `clang-format` (currently [15.0.7](https://github.com/llvm/llvm-project/releases)) is the best solution.

jhlegarreta:
https://github.com/InsightSoftwareConsortium/ITK/pull/3805#issuecomment-1408842616 OK, then +100 for upgrading. @hjmjohnson ?

hjmjohnson:
@jhlegarreta I trust the judgment of those who have invested time into this PR.  I will not have time to review it carefully.

jhlegarreta:
Cross-referencing https://github.com/InsightSoftwareConsortium/ITKClangFormatLinterAction/pull/3.

dzenanz:
Rebased

dzenanz:
This does reduce number of `warning: unbalanced grouping commands` from 92 down to 20.

jhlegarreta:
So I have revisited the comments here and the related issue (#3654):
- We did not make https://github.com/InsightSoftwareConsortium/ITK/pull/3805#issuecomment-1371375286 happen when the Perl script was translated to Python in PR #5157: if I read correctly,
https://github.com/InsightSoftwareConsortium/ITK/commit/a5db4a8a1057452e2a8d7532a1b5c75b7440bd9e#diff-41905fce6b73c41cc93104323e7d01cfec55393204c0069acc931745fa8438aeL56-L66

  https://github.com/InsightSoftwareConsortium/ITK/pull/5157/files#diff-56c509b541d6d68af2472e42febc63da9ec3e63b1d90bffd64a3e7bac3b975feR63-R68

  it was not done, unless the regular expression hits are different.

- The other possibility was transitioning to a newer `clang` (https://github.com/InsightSoftwareConsortium/ITK/pull/3805#issuecomment-1408842616). PR #5015 transitioned to `clang` 19, and 
https://github.com/InsightSoftwareConsortium/ITK/pull/5015/files#diff-1026e0038b722990204a42bed8a6f7c0ec2302aa79e3fad1959d62ba968edfa2R239, thus, using `Leave` (not introducing any new blank line, as opposed to `Always` cf. the `stackoverflow` answer posted in the link of that comment). So that looks OK to me, as now `clang` does not complain.

As for the discussion in issue #3654:
- The location of the `*@}` ending markers has not been changed by the Python script as far as I understand, so the underlying issue identified by @albert-github (https://github.com/InsightSoftwareConsortium/ITK/issues/3654#issuecomment-1353135951) seems not to have been addressed. According to https://github.com/InsightSoftwareConsortium/ITK/issues/3654#issuecomment-2610642614 and https://github.com/InsightSoftwareConsortium/ITK/issues/3654#issuecomment-2610662490, the purpose of the Python script should be reached by using other Doxygen means (disclaimer: I have not read the Python script except for the above part).
- As for the comment `Yes. We are currently using 1.9.4 and the latest release of Doxygen is 1.13.2. Time is likely better spent getting a newer version of Doxygen to work than maintaining the old one.` in https://github.com/InsightSoftwareConsortium/ITK/issues/3654#issuecomment-2610609494, updating the Doxygen version in `blaster` (which is the one that reports to CDash, not the version in https://github.com/InsightSoftwareConsortium/ITKDoxygen/commit/37cfce34712c7c0430235f52ad326d09fd11f9f4#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R18, which is 1.9.3) has not helped in removing the warnings.
- The remaining 20 warnings, are they because the files were not changed in this PR or because it is a demonstration that the underlying issue is not a simple addition of whitespaces?

So, these are my thoughts:
1. We can mark this as "Ready for review", ignore that the underlying issue has not been addressed (the use of the now Python script to do the job that the Doxygen `ALIASES` should do according to @albert-github), merge this PR, close issue #3654, and open a new issue to address the Python script vs. Doxygen `ALIASES` use. This would introduce a supposedly unnecessary change into the code, and one that does potentially not address the underlying issue (cf. the comment about the remaining 20 warnings).
1. We can wait until somebody has the bandwidth to address the Python script vs. Doxygen `ALIASES` use issue, at which point we would see whether it does the trick and could eventually close this PR and issue #3654.

My vote is for 1 for the sake of our mental well-being (which includes seeing less noise in the dashboard). But I am happy if others have the bandwidth to look into the `ALIASES` solution.

dzenanz:
I updated Doxygen on blaster to latest (1.13.2). Merging this PR sounds like a low-risk step in the right direction. Let's give people a few days to review (or maybe tackle option 2).

albert-github:
With the `ALIASES` solution from me you mean my proposal in https://github.com/InsightSoftwareConsortium/ITK/issues/4161#issuecomment-2629343940 ?

Still I see more in that approach than by means of using a script to convert the code based on some special white space in the code (as this probably will break again in the future / miss some special cases.). The note I made in #4161:
> Note: I think that running doxygen with -d preprocessor once with and once without the filter might help to identify where currently @{ ... @} is inserted so these files can be inspected (will be quite a bit of work though).

is probably still quite valid as it will be quite a bit of work in the already over stretched workloads.

dzenanz:
Doxygen warning log is swamped by `unbalanced grouping commands`. This PR will reduce that noticeably, and I think it would be advantageous to have that while waiting for someone to get around to fixing the grouping script. I agree that this is not maintainable in the long term.

jhlegarreta:
Re: https://github.com/InsightSoftwareConsortium/ITK/pull/3805#issuecomment-2639853070
> With the ALIASES solution from me you mean my proposal in https://github.com/InsightSoftwareConsortium/ITK/issues/4161#issuecomment-2629343940 ?

Yes.

> Still I see more in that approach than by means of using a script to convert the code based on some special white space in the code (as this probably will break again in the future / miss some special cases.). The note I made in https://github.com/InsightSoftwareConsortium/ITK/issues/4161:

> Note: I think that running doxygen with -d preprocessor once with and once without the filter might help to identify where currently @{ ... @} is inserted so these files can be inspected (will be quite a bit of work though).

> is probably still quite valid as it will be quite a bit of work in the already over stretched workloads.

I am open to any solution that is more robust than this PR. I think I made this clear. I do not have the bandwidth to try all avenues.

Re: https://github.com/InsightSoftwareConsortium/ITK/pull/3805#issuecomment-2639904282

> I am not in favor of this commit. Adding "magic" white space lines to avoid warnings is not a manageable and robust solution.

> The purpose to migrating the script to Python was that no one know Perl well enough to make the fix in the script. Now what the "grouping" script is in python it should be debuggable and fixable by the community.

There was a guess in https://github.com/InsightSoftwareConsortium/ITK/pull/3805#issuecomment-1407192615. Feel free to try.

albert-github:
I saw that in the output e.g. the mainpage was empty and its information should have been retrieved from the `MainPage.dox` file.
It looks like the problem also is a result from the filtering and as the `dox` files don't contain "C++" code but just comments, there is no need to run them through the "grouping filter".

Proposed patch: [diff.patch](https://github.com/user-attachments/files/18694127/diff.patch)

albert-github:
At all, in #4161 there were 2 suggestions to bypass the problems of the "grouping" script
1. Through `ALIASES`: `\itkGroupStart` / `\itkGroupEnd`
2. Through `\{` / `\}`

subsequently (no difference for doxygen) the used command identifier:
a. commands start with `\`
b. commands start with `@`


What is the consensus to use?

dzenanz:
> I saw that in the output e.g. the mainpage was empty and its information should have been retrieved from the `MainPage.dox` file. It looks like the problem also is a result from the filtering and as the `dox` files don't contain "C++" code but just comments, there is no need to run them through the "grouping filter".
> 
> Proposed patch: [diff.patch](https://github.com/user-attachments/files/18694127/diff.patch)

I created PR #5225 with the proposed changes.

PR #3813, opened 2022-12-20
STYLE: Replace `*x.begin()` with `x.front()`

N-Dekker:
The expression `x.front()` appears slightly more readable in these cases, because it does not have the `*` at the beginning.


dzenanz (approved):
Are there no more instances of this?

dzenanz (approved):

thewtex (approved):
Nice!!

PR #3818, opened 2022-12-23
COMP: Threads package is required by TBB

hjmjohnson:
```
CMake Error at /Users/johnsonhj/src/BT-release/BRAINSTools-Release-5.7.0/lib/cmake/TBB/TBBTargets.cmake:61 (set_target_properties):
  The link interface of target TBB::tbb contains:
    Threads::Threads
  but the target was not found.
```

In many builds "find_package(Threads)" is included from other third party libraries.  When requesting TBB, Threads should unconditionally be found and included.


dzenanz (approved):

thewtex (approved):
:+1: :1234:

PR #3840, opened 2023-01-01
ENH: Add test for the `itk::BoundedReciprocalImageFilter` class

jhlegarreta:
Add a test for the `itk::BoundedReciprocalImageFilter` class.

## PR Checklist


jhlegarreta:
Have not found a proper way to test the `BoundedReciprocal` functor alone. This would allow to test the filter's output against another instance that uses the `SetFunctor` method (which would take an instance of the `BoundedReciprocal` functor), thus freeing the filter testing from the [analytical comparison](https://github.com/InsightSoftwareConsortium/ITK/pull/3840/files#diff-ed4afdb18c31803cc935a8b11b18a186d56e33e330f22cb52f5a3b568bc60b02R71) (which would be done in the `BoundedReciprocal` functor test).

Otherwise, the current solution is enough to me/the above improvement can be made if somebody finds a way and the time to do it.

dzenanz (approved):

PR #3843, opened 2023-01-02
Fix MSVC warning C26495: `ZwQuerySystemInformation`, `m_MetaDataObjectValue` uninitialized

N-Dekker:
Fixed Visual Studio 2019 Code Analysis warnings from both `MetaDataObject` and `WindowsMemoryUsageObserver`, saying:
> warning C26495: Variable is uninitialized. Always initialize a member variable (type.6)

### DIFF
--- a/Modules/Core/Common/include/itkMemoryUsageObserver.h
+++ b/Modules/Core/Common/include/itkMemoryUsageObserver.h
@@ -74,9 +74,9 @@ protected:
   using PZwQuerySystemInformation = NTSTATUS(WINAPI *)(UINT, PVOID, ULONG, PULONG);
 
   // handle ntdll.dll library
-  HMODULE m_hNTLib;
+  HMODULE m_hNTLib{ nullptr };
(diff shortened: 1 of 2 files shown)


dzenanz (approved):

N-Dekker:
Note that MSVC/Code Analysis does not always produce this warning about uninitialized member variables, when it should. As I reported at developercommunity.visualstudio.com: [False negative (missing) /analyze warning C26495: Variable is uninitialized. Always initialize a member variable (type.6)](https://developercommunity.visualstudio.com/t/False-negative-missing-analyze-warnin/10239447). For a code example, see https://godbolt.org/z/c4GhdPMc8

So even when all of those "C26495" warnings would be fixed, there might still be some more uninitialized member variables...  So please upvote my MSVC bug report: https://developercommunity.visualstudio.com/t/False-negative-missing-analyze-warnin/10239447

thewtex (approved):
:100:

PR #3845, opened 2023-01-03
BUG: Fix uninitialized value `ImageRegistrationMethodv4::m_NumberOfLevels`

jhlegarreta:
Fix uninitialized value warning for
`itk::ImageRegistrationMethodv4::m_NumberOfLevels`.

Fixes:
```
UMC ==16463== Conditional jump or move depends on uninitialised value(s)
==16463==    at 0x9BD983: itk::ImageRegistrationMethodv4, itk::Image, itk::DisplacementFieldTransform, itk::Image, itk::PointSet > >::SetNumberOfLevels(unsigned long) (itkImageRegistrationMethodv4.hxx:881)
==16463==    by 0x9C44B6: itk::ImageRegistrationMethodv4, itk::Image, itk::DisplacementFieldTransform, itk::Image, itk::PointSet > >::ImageRegistrationMethodv4() (itkImageRegistrationMethodv4.hxx:103)
==16463==    by 0x9BBFA0: itk::SyNImageRegistrationMethod, itk::Image, itk::DisplacementFieldTransform, itk::Image, itk::PointSet > >::SyNImageRegistrationMethod() (itkSyNImageRegistrationMethod.hxx:41)
==16463==    by 0x9B5974: itk::SyNImageRegistrationMethod, itk::Image, itk::DisplacementFieldTransform, itk::Image, itk::PointSet > >::New() (itkSyNImageRegistrationMethod.h:84)
==16463==    by 0xA15CD4: itkSyNPointSetRegistrationTest(int, char**) (itkSyNPointSetRegistrationTest.cxx:132)
==16463==    by 0x317927: main (ITKRegistrationMethodsv4TestDriver.cxx:222)
```

and similar warnings triggered by registration testing code.

Visible in:
https://open.cdash.org/viewDynamicAnalysis.php?buildid=8374098

Introduced in d1c46d63be7f78d758720e1e0a463b73b30a4055.

Recover the comment highlighting the initialization to a 3-level
registration, and reword to make the need for the call (with a value
different from the current value) even more explicit.

## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [X] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)

PR #3851, opened 2023-01-04
STYLE: Add in-class `{}` member initializers to objects created by New()

N-Dekker:
For an object created by `New()`, the performance cost of adding a `{}` default member initializer to each of its non-static data members should be neglectable compared to the cost of allocating the object itself, as it is allocated on the heap.

In-class default member initialization effectively prevents Valgrind/Memcheck warnings like "Conditional jump or move depends on uninitialised value(s)", and Visual C++ Code Analysis warnings like "warning C26495: Variable is uninitialized. Always initialize a member variable (type.6)"

These cases are found using the regular expression `  \w.* m_\w[^{}=]+;`, excluding `static` data members and data members of a reference (`&`) type.


Modules/Bridge/VTK/include/itkVTKImageExport.h
```diff
@@ -128,14 +128,14 @@ class ITK_TEMPLATE_EXPORT VTKImageExport : public VTKImageExportBase
...
 private:
-  std::string m_ScalarTypeName;
-  int         m_WholeExtent[6];
-  int         m_DataExtent[6];
-  double      m_DataSpacing[3];
-  double      m_DataOrigin[3];
-  double      m_DataDirection[9];
-  float       m_FloatDataSpacing[3];
-  float       m_FloatDataOrigin[3];
+  std::string m_ScalarTypeName{};
```
dzenanz, line 131: I guess this is not needed for STL types?

Modules/Bridge/VtkGlue/include/itkImageToVTKImageFilter.h
```diff
@@ -105,8 +105,8 @@ class ITK_TEMPLATE_EXPORT ImageToVTKImageFilter : public ProcessObject
   ~ImageToVTKImageFilter() override;
 
 private:
-  ExporterFilterPointer m_Exporter;
-  vtkImageImport *      m_Importer;
+  ExporterFilterPointer m_Exporter{};
+  vtkImageImport *      m_Importer{};
```
dzenanz, line 109: I assume both of these initialize to nullptr?

dzenanz (commented):
Wouldn't this make it harder to spot a failure to really initialize these ivars in the constructor, or elsewhere?

Modules/Bridge/VTK/include/itkVTKImageExport.h
N-Dekker: > I guess this is not needed for STL types?

Strictly speaking, initialization by `{}` is not needed for STL containers, but it doesn't harm either. It just default-constructs them.

N-Dekker:
> Wouldn't this make it harder to spot a failure to really initialize these ivars in the constructor, or elsewhere?

Good question, but I'm not sure... Suppose a numeric data member is just zero-initialized by `{}`, while it should have been initialized to non-zero by the constructor, wouldn't that be easy to spot? Suppose a raw pointer is initialized as `nullptr` by `{}`, while it should have been initialized to a valid address in the constructor. Wouldn't that be easy to spot?

The use of uninitialized data isn't always spotted automatically, it may just have undefined behavior _silently_.

Modules/Bridge/VtkGlue/include/itkImageToVTKImageFilter.h
N-Dekker: > I assume both of these initialize to nullptr?

The short answer: yes 👍 

The long answer: `m_Importer` is a raw pointer, so it is initialized to nullptr, by `m_Importer{}`. m_Exporter is a ExporterFilterPointer, which is a `itk::SmartPointer<T>`. Which is default-constructed by `m_Exporter{}`.  The default-constructor of `itk::SmartPointer<T>` yields a smart pointer whose internal `m_Pointer` member is equal to nullptr.

N-Dekker:
Instead of a bulk commit like this, we could just keep chasing undefined member variables one by one, as in PR #3845 (by Jon)  #3843, and #3849 (by myself). Which is certainly useful, but also very time consuming. And we probably won't fix all of them that way, one by one. I think it's preferable to just initialize all data members _by default_. And only leave some particular data uninitialized in very special (performance critical) cases. What do you think?

dzenanz:
I have mildly positive attitude towards this PR. I would like some other people to pitch in.

N-Dekker:
ITK.Linux has some warning at https://open.cdash.org/viewBuildError.php?type=1&buildid=8377051, but it seems unrelated to this PR, I don't know:

> Modules/Core/Common/include/itkOctree.hxx:32:1: warning: const/copy propagation disabled: 49167 basic blocks and 82024 registers; increase --param max-gcse-memory above 504256752 [-Wdisabled-optimization]

N-Dekker:
FYI Here's a cheat sheet for the value initialized by `{}`:

  | Type | Value initialized by `{}` |
  |---|---|
  | numeric type (`int`, `unsigned`, `float`, etc.) | zero |
  | bool | false |
  | pointer type | nullptr |
  | class that has a user-provided default-constructor | default-constructed object |
  | array | all elements initialized (as if by `{}`) |
  | aggregate (struct) | all data members initialized (as if by `{}`) |

N-Dekker:
For the record, here is the discussion topic that I started on this subject: [Adding in-class `{}` default member initializers to data members of ITK classes](https://discourse.itk.org/t/adding-in-class-default-member-initializers-to-data-members-of-itk-classes/5613)

thewtex (approved):
👏 ✨

Modules/Compatibility/Deprecated/include/itkVectorResampleImageFilter.h
```diff
@@ -233,14 +233,14 @@ class ITK_TEMPLATE_EXPORT VectorResampleImageFilter : public ImageToImageFilter<
...
 private:
-  SizeType                m_Size;
-  TransformPointerType    m_Transform;
-  InterpolatorPointerType m_Interpolator;
-  PixelType               m_DefaultPixelValue;
-  SpacingType             m_OutputSpacing;
-  OriginPointType         m_OutputOrigin;
-  DirectionType           m_OutputDirection;
-  IndexType               m_OutputStartIndex;
+  SizeType                m_Size{};
```
jhlegarreta, line 236: Will these, the lists, the vectors or the arrays make unhappy Sean's AppleClang builds? I would dislike seeing new warnings because we are missing braces. I know that at this point the size of the size might not be known, but I ignore how AppleClang thinks.

jhlegarreta (commented):
Thanks for having taken the time to do this Niels.

A couple of in-line comments.

Please:
- Translate the gist (:ok_hand:) into a script file so that it dwells in the ITK Maintenance Utilities. Remotes may be grateful at some point.
- Apply this to non-default modules and examples (if not already done).
- Write this in the ITK SW guide.

Modules/Compatibility/Deprecated/include/itkVectorResampleImageFilter.h
N-Dekker: > Maybe I'm wrong and I have not had a look at the class of the m_Exporter pointer, but as you say it is a smart pointer, and these initialize themselves to the null pointer. Why is then this necessary? If it is, then it needs to be explained in the SW Guide; if it is an effect of the regex, then smart pointer cases should be revised/reverted IMO.

@jhlegarreta Indeed, for an `itk::SmartPointer`, the addition of an `{}` initializer has no effect at run-time. It's really a matter of style and consistency to add such an initializer to each and every data member. In my opinion, it will improve both robustness and code readability if we just add those initializers to _all_ data members, without having to distinguish between different types of data members.

Yes, I very much agree with you that this should be explained in the SW Guide.

Just to be sure, I double-checked for you: adding an empty pair of braces `{}` as initializer does _not_ introduce a clang warning (like "suggest braces around initialization of subobject [-Wmissing-braces]") 👍 Having tried the following code at https://godbolt.org/z/ofzxKzfan

```
struct Size
{
  int m_Data[3];
};

class MyClass
{
public:
  Size m_MissingBraces{1, 2}; // <== An old Clang warning here.
  Size m_ValueInitialized{};  // <== No Clang warning here  :-)
};

``` 
So that shouldn't be a problem 😃

jhlegarreta: OK. So I agree, the consistency and readability arguments are powerful in favor of this. Thanks Niels.

N-Dekker:
Thanks for your feedback @dzenanz 

> Give a few more days for others to comment/review, then merge?

Sure, that's fine to me! 

> But corresponding update should go into the software guide either concurrently with the merge of this, or very soon afterwards.

Actually, it already says:
> All member variables must be initialized when they are declared.

https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/blob/905c9bbc525adc73ff934d0ed24d06ef11151e22/SoftwareGuide/Latex/Appendices/CodingStyleGuide.tex#L1568

I just made a pull request to no longer exclude smart pointers. Instead, it proposes to exclude low level utility classes. Please check: https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/pull/192

Modules/Bridge/VtkGlue/include/itkImageToVTKImageFilter.h
jhlegarreta: Niels already replied and made the point clear.

The SW Guide PR has already been opened.

Resolving.

N-Dekker:
For your information, the last force-push just did another `git rebase master`, and solved the one little merge conflict.

PR #3859, opened 2023-01-10
DOC: Add emojis to the first-interaction message

thewtex:
Visually communicate our welcoming culture.


dzenanz (approved):

jhlegarreta (approved):
I'm usually reluctant to add emojis in our text-only Markdown or YML files, but I grant that they prettify otherwise daunting documents.

If they need to be added, I would prefer to use their text form, but not a requirement (other workflows already have them as pictures). So merge as necessary.

PR #3868, opened 2023-01-14
STYLE: Use braces for single line conditional loop body statements

jhlegarreta:
Use braces also for single line conditional loop body statements following the ITK SW style guidelines.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/test/itkImageAlgorithmCopyTest2.cxx
+++ b/Modules/Core/Common/test/itkImageAlgorithmCopyTest2.cxx
@@ -41,7 +41,9 @@ CheckBuffer(const TImageType * image, typename TImageType::PixelType p)
   while (!iter.IsAtEnd())
   {
     if (itk::Math::NotExactlyEquals(iter.Get(), p))
+    {
       return false;
+    }
     ++iter;
   }
   return true;


N-Dekker (approved):
OK of course, thanks Jon, but doesn't this occur on _many_ more places? For example

https://github.com/InsightSoftwareConsortium/ITK/blob/56ad54c2d262e147573137475fecc25adbe90a17/Modules/Core/Common/include/itkMath.h#L412-L415 

You may find them all by a regular expression like `  if \(.+\)\r\n [ ]+[^ {]` According to Notepad++ (on Windows), it has 265 hits in 109 files (using Filter = `itk*.h;itk*.hxx;itk*.cxx`)

jhlegarreta:
@N-Dekker Of course, it does. It was not meant to be a comprehensive PR at all; I just came across this while looking at other things, and I did not use a regex. However, I would very much appreciate a sweeping PR, but please contribute a script to the `Utilities` folder as well.

N-Dekker:
Thanks for explaining why you only addressed this particular case, Jon. I don't have time to address the other 200+ cases right now. I want to focus now on getting the initialization done (including PR #3851). If it is a priority to you, the regular expression that I posted (`  if \(.+\)\r\n [ ]+[^ {]`) may be of help to you. Just my two cent.

jhlegarreta:
>  If it is a priority to you, the regular expression that I posted (  if \(.+\)\r\n [ ]+[^ {]) may be of help to you. Just my two cent.

The regex is of great help to find them, and I'd very much like to have them changed, ideally with the regex and a script, but am also focusing on other changes now.

jhlegarreta:
Alternatively, we can set the state of this PR to draft or open an issue. What do you think @N-Dekker ?

N-Dekker:
> Alternatively, we can set the state of this PR to draft or open an issue. What do you think @N-Dekker ?

Looks like a good plan. By the way, I would just do the find-replace in the editor (Notepad++) or IDE (Visual Studio), so then I would not really have a script file. The main thing is to get the regular expression right. And it seems sufficient to me to just mention the regular expression in the commit message and/or the pull request. But of course, if you could write a script file (bash/bat/shell/python or even C++), feel free to do so. 

Note that the style improvement to add braces to single statements can be done with a find-replace in an editor, while for the in-class `{}` member initializers (PR #3851) I really needed some kind of scripting, because I did not want to "blindly" add those `{}` braces everywhere. The script was there to specifically look for classes that have an `itkNewMacro` macro call.

dzenanz:
Superseded by #3875.

PR #3878, opened 2023-01-20
STYLE: Default default-constructors of class templates in Core/Common

N-Dekker:
Defaulted default-constructors that just set all data members to zero or null, and those that only just call their base class.

### DIFF
--- a/Modules/Core/Common/include/itkBoundingBox.hxx
+++ b/Modules/Core/Common/include/itkBoundingBox.hxx
@@ -125,13 +125,6 @@ BoundingBox<TPointIdentifier, VPointDimension, TCoordRep, TPointsContainer>::Get
 }
 #endif
 
-/** */
-template <typename TPointIdentifier, unsigned int VPointDimension, typename TCoordRep, typename TPointsContainer>
-BoundingBox<TPointIdentifier, VPointDimension, TCoordRep, TPointsContainer>::BoundingBox()
-  : m_PointsContainer(nullptr)
-{
-  m_Bounds.Fill(NumericTraits<CoordRepType>::ZeroValue());
-}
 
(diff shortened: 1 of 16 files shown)


Modules/Core/Common/include/itkBoundingBox.hxx
```diff
@@ -125,13 +125,6 @@ BoundingBox<TPointIdentifier, VPointDimension, TCoordRep, TPointsContainer>::Get
 }
 #endif
 
-/** */
-template <typename TPointIdentifier, unsigned int VPointDimension, typename TCoordRep, typename TPointsContainer>
-BoundingBox<TPointIdentifier, VPointDimension, TCoordRep, TPointsContainer>::BoundingBox()
-  : m_PointsContainer(nullptr)
-{
-  m_Bounds.Fill(NumericTraits<CoordRepType>::ZeroValue());
```
dzenanz, line 133: Does `m_Bounds` still get initialized the same way?

N-Dekker: I think so, definitely. `m_Bounds` a FixedArray of `CoordRepType`, so it should be properly zero-initialized by its in-class `{}` initializer. Don't you agree?

dzenanz: It might be un-initialized? Best to check in the debugger.

N-Dekker: > It might be un-initialized? Best to check in the debugger.

Just submitted pull request #3881, to show that an empty `{}` properly initializes a FixedArray. Does that help?

dzenanz (approved):
Please rebase to fix conflicts.

PR #3881, opened 2023-01-20
ENH: Add GTest `FixedArray.ValueInitialized`

N-Dekker:
Tests that initialization by `{}` properly initializes a `FixedArray`.

### DIFF
--- a/Modules/Core/Common/test/itkFixedArrayGTest.cxx
+++ b/Modules/Core/Common/test/itkFixedArrayGTest.cxx
@@ -25,7 +25,8 @@
 #include <gtest/gtest.h>
 
 #include <array>
-#include <numeric> // For iota.
+#include <numeric>     // For iota.
+#include <type_traits> // For remove_reference_t.
 
 
 namespace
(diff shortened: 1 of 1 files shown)


Modules/Core/Common/test/itkFixedArrayGTest.cxx
```diff
@@ -324,3 +325,22 @@ TEST(FixedArray, StdMemberFunctionsWork)
...
+// Tests that each element of a value-initialized FixedArray is itself value-initialized. By definition,
+// "value-initialization" is performed when an object is constructed with an empty {} initializer.
+TEST(FixedArray, ValueInitialized)
+{
+  const auto expectEachElementValueInitialized = [](const auto & fixedArray) {
+    for (const auto & element : fixedArray)
+    {
+      using ValueType = std::remove_reference_t<decltype(element)>;
+
+      EXPECT_TRUE(element == ValueType{});
```
dzenanz: I would be more explicit here, and compare element to new lambda parameter `expectedValue`.

Modules/Core/Common/test/itkFixedArrayGTest.cxx
```diff
@@ -324,3 +325,22 @@ TEST(FixedArray, StdMemberFunctionsWork)
...
+    }
+  };
+
+  expectEachElementValueInitialized(itk::FixedArray<int, 3>{});
```
dzenanz, line 343: From here, pass 0, 0.0f, and nullptr as expectedValue.

N-Dekker: Sorry but I believe it's beyond the scope of this test to check that `T{}` equals 0 when T is an int, and that `T{}` equals `nullptr` when T is a pointer. That's a fundamental property of C++.

We could write such tests, like `EXPECT_EQ(int{}, 0)` and  `EXPECT_EQ(float{}, 0.0f)`, but I don't think they belong in a FixedArray test. They would fit better in a fundamental C++ feature test.

dzenanz (approved):

PR #147, opened 2018-11-07
WIP: Fix MINC read/write to convert from/to RAS coordinate system

gdevenyi:
Convert the coordinate system from RAS to LPS during MINC loading, and convert back on writing.

Also adds a setting to keep old configuration, so that MINC-based ITK tools (EZMinc) can maintain their original configuration and return sensible coordinate values in MINC-space.

Paging @vfonov

PR #3884, opened 2023-01-21
BUG: Fix reading of displacement field vectors from NIFTI

lassoan:
NIFTI image containing displacement field (intent = NIFTI_INTENT_DISPVECT) was read as scalar volume (all 3 vector components were set to the same value). Fixed by reading it the same way as the general-purpose vector image (intent = NIFTI_INTENT_VECTOR).

## PR Checklist

The changes already improve NIFTI compatibility, but we could potentially further improve correctness if we can answer these questions:
1. Should the displacement vectors be converted from RAS (native NIFTI) to LPS coordinate system (ITK convention)? The same way as the reader changes the image origin, spacing, axis directions to be in LPS coordinate system instead RAS.
2. NIFTI distinguishes between "vector" and "displacement vector" intent. Is there a way to keep this distinction in ITK? It would be nice if we read a displacement vector image and then we save it to file again to preserve the exact type.

### DIFF
--- a/Modules/IO/NIFTI/include/itkNiftiImageIO.h
+++ b/Modules/IO/NIFTI/include/itkNiftiImageIO.h
@@ -188,6 +188,37 @@ public:
   itkSetMacro(LegacyAnalyze75Mode, NiftiImageIOEnums::Analyze75Flavor);
   itkGetConstMacro(LegacyAnalyze75Mode, NiftiImageIOEnums::Analyze75Flavor);
 
+  /** Enable conversion of vector coordinates between RAS coordinate system (in NIFTI file) and
+   * LPS (ITK convention) when reading or writing a generic vector image file (intent code: 1007).
+   *
+   * This flag is disabled by default because vectors may store non-spatial information.
+   *
+   * By default an ITK vector image is written to NIFTI file as a generic vector image, unless
+   * "intent_code" field is set explicitly to set to "1006" (displacement vector).
+   */
+  itkSetMacro(ConvertRASVectors, bool);
+  itkGetConstMacro(ConvertRASVectors, bool);
+  itkBooleanMacro(ConvertRASVectors);
+
+  /** Enable conversion of vector coordinates between RAS coordinate system (in NIFTI file) and
+   * LPS (ITK convention) when reading or writing a "displacement vector" file (intent code: 1006).
+   *
+   * This flag is enabled by default. Disabling it may be useful to avoid unnecessary conversions
+   * in applications that uses RAS coordinate system.
+   *
+   * To write a vector image as displacement vector: set "intent_code" to "1006" in the metadata
+   * dictionary of the input image.
+   *
+   \code
+       itk::MetaDataDictionary& dictionary = image->GetMetaDataDictionary();
+       itk::EncapsulateMetaData<std::string>(dictionary, "intent_code", "1006");
+   \endcode
+   *
+   */
+  itkSetMacro(ConvertRASDisplacementVectors, bool);
+  itkGetConstMacro(ConvertRASDisplacementVectors, bool);
+  itkBooleanMacro(ConvertRASDisplacementVectors);
+
 protected:
   NiftiImageIO();
   ~NiftiImageIO() override;
@@ -237,6 +268,10 @@ private:
 
   double m_RescaleSlope{ 1.0 };
   double m_RescaleIntercept{ 0.0 };
+
+  bool m_ConvertRAS{ false };
+  bool m_ConvertRASVectors{ false };
+  bool m_ConvertRASDisplacementVectors{ true };
 
   IOComponentEnum m_OnDiskComponentType{ IOComponentEnum::UNKNOWNCOMPONENTTYPE };
 
--- a/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
+++ b/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
@@ -478,6 +478,9 @@ NiftiImageIO::PrintSelf(std::ostream & os, Indent indent) const
   os << indent << "RescaleIntercept: " << this->m_RescaleIntercept << std::endl;
   os << indent << "OnDiskComponentType: " << this->m_OnDiskComponentType << std::endl;
   os << indent << "LegacyAnalyze75Mode: " << this->m_LegacyAnalyze75Mode << std::endl;
+  os << indent << "ConvertRASVectors: " << m_ConvertRASVectors << std::endl;
+  os << indent << "ConvertRASDisplacementVectors: " << m_ConvertRASDisplacementVectors << std::endl;
+  os << indent << "ConvertRAS: " << m_ConvertRAS << std::endl;
 }
 
 bool
@@ -495,6 +498,9 @@ NiftiImageIO::MustRescale() const
          (itk::Math::abs(this->m_RescaleSlope - 1.0) > std::numeric_limits<double>::epsilon() ||
           itk::Math::abs(this->m_RescaleIntercept) > std::numeric_limits<double>::epsilon());
 }
+
+namespace
+{
 
 // Internal function to rescale pixel according to Rescale Slope/Intercept
 template <typename TBuffer>
@@ -520,6 +526,37 @@ CastCopy(float * to, const void * from, size_t pixelcount)
     to[i] = static_cast<float>(_from[i]);
   }
 }
+
+// Internal function to convert vectors between RAS and LPS coordinate systems.
+// Dimensions are CXYZT (ITK memory layout)
+template <typename TBuffer>
+void
+ConvertRASToFromLPS_CXYZT(TBuffer * buffer, size_t size)
+{
+  size_t numberOfVectors = size / 3;
+  for (size_t i = 0; i < numberOfVectors; ++i)
+  {
+    buffer[0] *= -1;
+    buffer[1] *= -1;
+    buffer += 3;
+  }
+}
+
+// Internal function to convert vectors between RAS and LPS coordinate systems.
+// Dimensions are XYZTC (NIFTI memory layout)
+template <typename TBuffer>
+void
+ConvertRASToFromLPS_XYZTC(TBuffer * buffer, size_t size)
+{
+  // Flip the sign of the first two components (L<->R, P<->A)
+  // and keep the third component (S) unchanged.
+  size_t numberOfComponents = size / 3 * 2;
+  for (size_t i = 0; i < numberOfComponents; ++i)
+  {
+    buffer[i] *= -1;
+  }
+}
+} // namespace
 
 void
 NiftiImageIO::Read(void * buffer)
@@ -791,6 +828,27 @@ NiftiImageIO::Read(void * buffer)
         }
     }
   }
+
+  if (this->m_ConvertRAS)
+  {
+    if (this->GetPixelType() != IOPixelEnum::VECTOR && this->GetPixelType() != IOPixelEnum::POINT)
+    {
+      itkExceptionMacro(<< "RAS conversion requires pixel to be 3-component vector or point. Current pixel type is "
+                        << numComponents << "-component " << this->GetPixelType() << ".");
+    }
+    switch (this->m_ComponentType)
+    {
+      case IOComponentEnum::FLOAT:
+        ConvertRASToFromLPS_CXYZT(static_cast<float *>(buffer), numElts * numComponents);
+        break;
+      case IOComponentEnum::DOUBLE:
+        ConvertRASToFromLPS_CXYZT(static_cast<double *>(buffer), numElts * numComponents);
+        break;
+      default:
+        itkExceptionMacro(<< "RAS conversion of datatype " << this->GetComponentTypeAsString(this->m_ComponentType)
+                          << " is not supported");
+    }
+  }
 }
 
 NiftiImageIOEnums::NiftiFileEnum
@@ -1043,7 +1101,8 @@ NiftiImageIO::ReadImageInformation()
   // Check the intent code, it is a vector image, or matrix image, then this is
   // not true.
   //
-  if (this->m_NiftiImage->intent_code == NIFTI_INTENT_VECTOR ||
+  if (this->m_NiftiImage->intent_code == NIFTI_INTENT_DISPVECT ||
+      this->m_NiftiImage->intent_code == NIFTI_INTENT_VECTOR ||
       this->m_NiftiImage->intent_code == NIFTI_INTENT_SYMMATRIX)
   {
     if (this->m_NiftiImage->dim[4] > 1)
@@ -1085,7 +1144,8 @@ NiftiImageIO::ReadImageInformation()
     this->SetNumberOfComponents(1);
   }
 
-  if (this->m_NiftiImage->intent_code == NIFTI_INTENT_VECTOR ||
+  if (this->m_NiftiImage->intent_code == NIFTI_INTENT_DISPVECT ||
+      this->m_NiftiImage->intent_code == NIFTI_INTENT_VECTOR ||
       this->m_NiftiImage->intent_code == NIFTI_INTENT_SYMMATRIX)
   {
     this->SetNumberOfComponents(this->m_NiftiImage->dim[5]);
@@ -1187,13 +1247,19 @@ NiftiImageIO::ReadImageInformation()
 
   // there are a wide variety of intents we ignore
   // but a few wee need to care about
+  this->m_ConvertRAS = false;
   switch (this->m_NiftiImage->intent_code)
   {
     case NIFTI_INTENT_SYMMATRIX:
       this->SetPixelType(IOPixelEnum::SYMMETRICSECONDRANKTENSOR);
       break;
+    case NIFTI_INTENT_DISPVECT:
+      this->SetPixelType(IOPixelEnum::VECTOR);
+      this->m_ConvertRAS = m_ConvertRASDisplacementVectors;
+      break;
     case NIFTI_INTENT_VECTOR:
       this->SetPixelType(IOPixelEnum::VECTOR);
+      this->m_ConvertRAS = m_ConvertRASVectors;
       break;
     case NIFTI_INTENT_NONE:
     case NIFTI_INTENT_CORREL:
@@ -1223,7 +1289,6 @@ NiftiImageIO::ReadImageInformation()
     case NIFTI_INTENT_LABEL:
     case NIFTI_INTENT_NEURONAME:
     case NIFTI_INTENT_GENMATRIX:
-    case NIFTI_INTENT_DISPVECT:
     case NIFTI_INTENT_POINTSET:
     case NIFTI_INTENT_TRIANGLE:
     case NIFTI_INTENT_QUATERNION:
@@ -1520,14 +1585,16 @@ NiftiImageIO::WriteImageInformation()
   }
   const unsigned int numComponents = this->GetNumberOfComponents();
 
+  MetaDataDictionary & thisDic = this->GetMetaDataDictionary();
+
   // TODO:  Also need to check for RGB images where numComponets=3
   if (numComponents > 1 && !(this->GetPixelType() == IOPixelEnum::COMPLEX && numComponents == 2) &&
       !(this->GetPixelType() == IOPixelEnum::RGB && numComponents == 3) &&
       !(this->GetPixelType() == IOPixelEnum::RGBA && numComponents == 4))
   {
-    this->m_NiftiImage->ndim = 5;   // This must be 5 for NIFTI_INTENT_VECTOR
+    this->m_NiftiImage->ndim = 5;   // This must be 5 for NIFTI_INTENT_VECTOR and NIFTI_INTENT_DISPVECT
                                     // images.
-    this->m_NiftiImage->dim[0] = 5; // This must be 5 for NIFTI_INTENT_VECTOR
+    this->m_NiftiImage->dim[0] = 5; // This must be 5 for NIFTI_INTENT_VECTOR and NIFTI_INTENT_DISPVECT
                                     // images.
     if (this->GetNumberOfDimensions() > 4)
     {
@@ -1543,7 +1610,21 @@ NiftiImageIO::WriteImageInformation()
     }
     else
     {
-      this->m_NiftiImage->intent_code = NIFTI_INTENT_VECTOR;
+      // Each voxel is a vector. Set intent code to NIFTI_INTENT_VECTOR (default)
+      // or NIFTI_INTENT_DISPVECT.
+      int         intentCode = NIFTI_INTENT_VECTOR;
+      std::string intentCodeStrInMetaData;
+      if (itk::ExposeMetaData<std::string>(thisDic, "intent_code", intentCodeStrInMetaData))
+      {
+        std::istringstream is(intentCodeStrInMetaData);
+        int                intentCodeInMetaData = -1;
+        is >> intentCodeInMetaData;
+        if (intentCodeInMetaData == NIFTI_INTENT_DISPVECT)
+        {
+          intentCode = NIFTI_INTENT_DISPVECT;
+        }
+      }
+      this->m_NiftiImage->intent_code = intentCode;
     }
     this->m_NiftiImage->nu = this->m_NiftiImage->dim[5] = this->GetNumberOfComponents();
     if (this->GetNumberOfDimensions() < 4)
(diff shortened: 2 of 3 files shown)


Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -1543,7 +1548,11 @@ NiftiImageIO::WriteImageInformation()
     }
     else
     {
+      // TODO: There is no way to tell if it is a displacement vector (NIFTI_INTENT_DISPVECT)
+      // or some other vector (NIFTI_INTENT_VECTOR), therefore we always use generic vector intent.
       this->m_NiftiImage->intent_code = NIFTI_INTENT_VECTOR;
+      // TODO: consider reorienting vector field values (NIFTI_INTENT_DISPVECT or NIFTI_INTENT_VECTOR)
+      // from ITK's LPS coordinate system to NIFTI's native RAS coordinate system.
```
N-Dekker, line 1555: Thanks @lassoan Would it be and idea to add such information to the MetaDataDictionary of the image? For example:

```
  // Assume LPS if unspecified.
  EncapsulateMetaData(
    metaDataDictionary, "CoordinateSystem", SpatialOrientationEnums::ValidCoordinateOrientations::ITK_COORDINATE_ORIENTATION_RAS);
  EncapsulateMetaData(metaDataDictionary, "Intent", "DisplacementField");

```

dzenanz: Of course, the writer then needs to make of the information in MetaDataDictionary.

dzenanz:
@lassoan I didn't deal much with DisplacementFields, but I assume that converting it to LPS coordinate system would be preferable. Or maybe have a setting in NiftiIO which controls it, and defaults to `ConvertToLPS`?

dzenanz (approved):
This PR looks good as is. The discussed improvements could be done in this PR, or a follow up PR.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
lassoan: Thanks for the suggestions!

MetaDataDictionary could work well for reading/writing the intent code. I'll add a commit to this PR to use these dictionary fields.

For the vector reorientation, I see two main options:
- Option A: ITK always converts to/from RAS if intent is "displacement vector" (does not reorient if intent is just generic "vector"). This would make all ITK-based tools handle "displacement vector" data sets as expected. Helper functions for converting between RAS/LPS coordinate system may be made available to users to manually deal with data sets that use generic "vector" intent for storing spatial data.
- Option B: Let the user reorient (regardless of "displacement vector" or "vector" intent), based on the MetadataDictionary field. This would be simpler for ITK but most likely users/applications would not be aware that flipping is necessary and they would just work incorrectly.

Would you have any preference?

lassoan:
> I didn't deal much with DisplacementFields, but I assume that converting it to LPS coordinate system would be preferable. Or maybe have a setting in NiftiIO which controls it, and defaults to ConvertToLPS

Yes - see the two options that I described above.

Probably Option A would be more consistent with what ITK does already (automatic conversion of image geometry to LPS).

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
dzenanz: Based on your description, I prefer A.

N-Dekker: I think I would prefer option A as well, unless some users would prefer to remain in RAS space...? If option A can be implemented, of course my suggested `("CoordinateSystem", ITK_COORDINATE_ORIENTATION_RAS)` dictionary entry would not be necessary anymore. I don't know if an `("Intent", "DisplacementField")` dictonary entry would then still be useful either. No problem.

lassoan: I was thinking about the automatic conversion to RAS, too. Slicer currently uses RAS internally, so after ITK reads a displacement field from a NRRD file, Slicer convert all the vectors to RAS. In case of NIFTI it would mean two conversions - RAS->LPS in ITK then LPS->RAS in Slicer. The time needed for conversion is usually not noticeable, so we could implement this simple, always-convert solution, and then if somebody runs into a performance issue then a bypass mechanism could be implemented.

lassoan:
Could you recommend a simple and efficient way to rescale vector components with (-1,-1,1) factors to switch between LPS/RAS?

dzenanz:
[MultiplyImageFilter](https://itk.org/Doxygen/html/classitk_1_1MultiplyImageFilter.html) sounds as the right solution.

lassoan:
Is there an example of using it within a filter?

The class already can do some rescaling but only for scalars. Maybe it could be just modified to allow applying different factors for each vector component?

N-Dekker:
@lassoan Something like this? 

    using VectorType = itk::Vector<float, 3>;
    using DisplacementFieldImageType = itk::Image<VectorType>;

    void
    ConvertCoordinates(DisplacementFieldImageType & displacementFieldImage)
    {
      for (VectorType & displacementVector : itk::ImageBufferRange<DisplacementFieldImageType>(displacementFieldImage))
      {
        displacementVector[0] *= -1;
        displacementVector[1] *= -1;
      }
    }

This range-based for-loop would iterate all the vectors in the displacement field image, and multiply the first and the second vector component "in place" by -1.

dzenanz:
[multiplyimagebyscalar](https://examples.itk.org/src/filtering/imageintensity/multiplyimagebyscalar/documentation) name was given to differentiate it from multiplying pixels of two images. If you replace scalar types by vector types, it might work. If not, you might also need to provide definition for vector multiplication. But the filter should take care of parallelizing the operation. @N-Dekker's suggestion is not parallel.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
gdevenyi: The ITK MINC reader needs this LPS/RAS conversion as it currently blindly assumes LPS, which is wrong, see #147

gdevenyi:
The ITK MINC reader has similar issues of blindly reading in as LPS without conversion, see #147. I would love to see a generic solution to this as its needed for that IO system as well.

lassoan:
Thanks for all the suggestions I'm still reviewing them to see which ones are the best applicable.

@gdevenyi This is indeed the same kind of issue as for MINC. It would be great to reuse the code that you implemented for MINC, but I find it hard (even if I copy-pasted the code) because the file formats are just too different and the IO classes are already way too complicated. Maybe by adding an adaptor layer outside could address this, but that would mean a much more complicated design, and I think it would only be needed for NIFTI and MINC, so I'm not sure if it is worth the effort.

lassoan:
I've added conversion of vectors between RAS/LPS coordinate systems. By default, conversion is enabled for displacement vector images and disabled for generic vector images. These defaults can be overridden by two member variables in the IO file.

I've also added the possibility of writing displacement vector images, by setting the "intent_code"="1006" in the metadata dictionary. I've kept the magic number stored as string in the metadata dictionary, as the NIFTI IO class already stored intent code like this and I wanted to preserve backward compatibility.

@dzenanz @N-Dekker it would be great if you could review.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -521,6 +523,36 @@ CastCopy(float * to, const void * from, size_t pixelcount)
   }
 }
 
+// Internal function to convert vectors between RAS and LPS coordinate systems.
+// Dimensions are CXYZT (ITK memory layout)
+template <typename TBuffer>
+void
+ConvertRASToFromLPS_CXYZT(TBuffer * buffer, size_t size)
+{
+  int numberOfVectors = size / 3;
```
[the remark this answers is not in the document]
lassoan: Good catch, I don't know why I wrote `int` there.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -521,6 +523,36 @@ CastCopy(float * to, const void * from, size_t pixelcount)
...
+  for (size_t i = 0; i < numberOfVectors; ++i)
+  {
+    buffer[0] = -buffer[0];
+    buffer[1] = -buffer[1];
+    buffer += 3;
```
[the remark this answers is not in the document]
lassoan: Makes sense. Should we use `*= -1` or `*= -1.0`?

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -521,6 +523,36 @@ CastCopy(float * to, const void * from, size_t pixelcount)
...
+
+// Internal function to convert vectors between RAS and LPS coordinate systems.
+// Dimensions are XYZTC (NIFTI memory layout)
+template <typename TBuffer>
+void
+ConvertRASToFromLPS_XYZTC(TBuffer * buffer, size_t size)
+{
+  // Flip the sign of the first two components (L<->R, P<->A)
+  // and keep the third component (S) unchanged.
+  int numberOfVectors = size / 3 * 2;
```
N-Dekker, line 549: In this context, I think `numberOfVectors` is not really the number of vectors, right? It seems to me that it is the number of components whose sign is to be flipped. If so, maybe rename it to `numberOfComponents`...?

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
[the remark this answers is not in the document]
dzenanz: `*= -1` is better. Using `-1.0` has type `double`, and the compiler might not properly optimize the expression if it happens to dealing with integral type (I am not sure whether this is possible here) or single-precision components.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -1543,7 +1605,21 @@ NiftiImageIO::WriteImageInformation()
     }
     else
     {
-      this->m_NiftiImage->intent_code = NIFTI_INTENT_VECTOR;
+      // Each voxel is a vector. Set intent code to NIFTI_INTENT_VECTOR (default)
+      // or NIFTI_INTENT_DISPVECT.
+      int         intentCode = NIFTI_INTENT_VECTOR;
+      std::string intentCodeStrInMetaData;
+      if (itk::ExposeMetaData<std::string>(thisDic, "intent_code", intentCodeStrInMetaData))
```
N-Dekker, line 1612: So "intent_code" was already there in the dictionary! Cool, I did not know! 👍

Modules/IO/NIFTI/include/itkNiftiImageIO.h
```diff
@@ -238,6 +269,10 @@ class ITKIONIFTI_EXPORT NiftiImageIO : public ImageIOBase
   double m_RescaleSlope{ 1.0 };
   double m_RescaleIntercept{ 0.0 };
 
+  bool m_ConvertRAS{ false };
+  bool m_ConvertRASVectors{ false };
+  bool m_ConvertRASDisplacementVectors{ true };
```
N-Dekker, line 274: No serious flaws found so far 😃 But I'm wondering, could we do without `m_ConvertRAS`? It seems to me that it isn't really a _property_ like m_ConvertRASVectors and m_ConvertRASDisplacementVectors. It should always just be equal to:

`(m_ConvertRASVectors && intent_code == NIFTI_INTENT_VECTOR) || (m_ConvertRASDisplacementVectors && intent_code == NIFTI_INTENT_DISPVECT) 
`
Right? Then maybe a little `private` member function would do, instead of `m_ConvertRAS`:
```

bool DoConvertRAS() const
{
  return (m_ConvertRASVectors && intent_code == NIFTI_INTENT_VECTOR) || (m_ConvertRASDisplacementVectors && intent_code == NIFTI_INTENT_DISPVECT);
}
```

Anyway, no problem, just a suggestion.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
lassoan: Correct. numberOfVectors=size/3. The numberOfVectors*2 is the number of components we update.

N-Dekker: Thanks! Don't forget the `int` here as well!

Modules/IO/NIFTI/include/itkNiftiImageIO.h
lassoan: We need to specify two behaviors: what to do if the file contains generic vector, and what to do if the file contains displacement vector.

We could just use one flag, but then overriding the default behavior would be more complicated: we would need to know what kind of data we read/write and change the flag accordingly. This is particularly inconvenient for reading, because we don't know the vector type of the file until we read the file header, so an extra pre-reading step could be necessary.

N-Dekker: OK, thank you for explaining. I was just wondering if we could do two flags, instead of three. But it's OK anyway!

lassoan: The proposed helper function would need to get the intent code and the source would be slightly different, depending on doing reading or writing (metadata dictionary or `m_NiftiImage->intent_code`). So, overall, I think the current implementation is a bit simpler. The third flag is private, so it does not add much complexity.

lassoan:
@N-Dekker Thank you very much for the review, I've addressed your comments and force-pushed. Please have a look.

Modules/IO/NIFTI/test/itkNiftiImageIOTest3.cxx
```diff
@@ -295,5 +300,10 @@ itkNiftiImageIOTest3(int argc, char * argv[])
   success |= TestImageOfVectors<float, 4, 4>(std::string("testVectorImage_float_4_4.nii.gz"));
   success |= TestImageOfVectors<double, 3, 3>(std::string("testVectorImage_double_3_3.nii.gz"));
 
+  // Test reading/writing as displacement field (NIFTI intent code = 1006)
+  success |=
+    TestImageOfVectors<double, 3, 1>(std::string("testDispacementImage_double_3_1.nii.gz"), std::string("1006"));
+  success |= TestImageOfVectors<float, 3, 1>(std::string("testDisplacementImage_float.nii.gz"), std::string("1006"));
```
N-Dekker, line 306: "testDisplacementImage_float.nii.gz", should that be "testDisplacementImage_float_3_1.nii.gz"? No problem to me, just wondering, because the other one has "_3_1" in the file name.

lassoan: Good catch. It should not affect functionality, but I'll rename for consistency.

Modules/IO/NIFTI/test/itkNiftiImageIOTest3.cxx
```diff
@@ -143,6 +143,11 @@ TestImageOfVectors(const std::string & fname)
       }
     }
   }
+  if (!intentCode.empty())
+  {
+    itk::MetaDataDictionary & dictionary = vi->GetMetaDataDictionary();
+    itk::EncapsulateMetaData<std::string>(dictionary, "intent_code", intentCode);
+  }
```
N-Dekker, line 150: The test sets the intent_code before writing, so shouldn't it then also check that after reading the file back, the intent_code is there in the read image?

N-Dekker:
The library code looks fine to me, just two questions on the test!

Modules/IO/NIFTI/include/itkNiftiImageIO.h
```diff
@@ -238,6 +269,10 @@ class ITKIONIFTI_EXPORT NiftiImageIO : public ImageIOBase
   double m_RescaleSlope{ 1.0 };
   double m_RescaleIntercept{ 0.0 };
 
+  bool m_ConvertRAS{ false };
```
jhlegarreta, line 272: Private members also need to get printed as per the ITK SW Guide:
https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/blob/master/SoftwareGuide/Latex/Appendices/CodingStyleGuide.tex#L3622

jhlegarreta (commented):
Thanks for doing all this Andras :100:.

An in-line comment about printing `m_ConvertRAS`.

Modules/IO/NIFTI/test/itkNiftiImageIOTest3.cxx
lassoan: We could defintely add a lot more testing, but that would take much more time. I would prefer to not put much more effort into this change at this point.

Modules/IO/NIFTI/include/itkNiftiImageIO.h
lassoan: OK, I'll print it.

lassoan:
I've force-pushed the requested changes (except adding more tests).

Modules/IO/NIFTI/test/itkNiftiImageIOTest3.cxx
N-Dekker: So then you just test that is it _possible_ to add an entry to the dictionary?

lassoan: The test performs basic checks (there is no crash or exception and the data that is written and read back is the same as the original) when reading and writing displacement fields in NIFTI file.
Probably the next test that someone should add is to compare with some reference NIFTI files. But some time ago I had to add test data and the process was so complicated that I would rather not invest time into going through that again at this point.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -521,6 +524,36 @@ CastCopy(float * to, const void * from, size_t pixelcount)
   }
 }
 
+// Internal function to convert vectors between RAS and LPS coordinate systems.
+// Dimensions are CXYZT (ITK memory layout)
```
jhlegarreta, line 528: If this documentation is relevant to the method/class, should be put in the method's declaration docstring; if it only relevant to those reading/maintaining the code, please move it to the method implementation body (e.g. before any statement).

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -521,6 +524,36 @@ CastCopy(float * to, const void * from, size_t pixelcount)
...
+  for (size_t i = 0; i < numberOfVectors; ++i)
+  {
+    buffer[0] *= -1;
+    buffer[1] *= -1;
+    buffer += 3;
+  }
+}
+
+// Internal function to convert vectors between RAS and LPS coordinate systems.
+// Dimensions are XYZTC (NIFTI memory layout)
```
jhlegarreta, line 543: Same as above.

Modules/IO/NIFTI/test/itkNiftiImageIOTest3.cxx
N-Dekker: I was expecting something like this, after `readback = itk::IOTestHelper::ReadImage<VectorImageType>(fname)`:
```
  if (!intentCode.empty())
  {
    const itk::MetaDataDictionary & dictionary = readback->GetMetaDataDictionary();

    std::string readIntentCode;
    if (itk::ExposeMetaData<std::string>(dictionary, "intent_code", readIntentCode)
    {
      if (readIntentCode != intentCode)
      {
        std::cout << "intent_code is different: " << readIntentCode << " != " << intentCode << std::endl;
        same = false;
      }

    }
    else
    {
      std::cout << "The read image should have an intent_code in its dictionary" << std::endl;
      same = false;
    }
  }
```

Isn't that the idea?

Modules/IO/NIFTI/include/itkNiftiImageIO.h
```diff
@@ -188,6 +188,37 @@ class ITKIONIFTI_EXPORT NiftiImageIO : public ImageIOBase
...
+  /** Enable conversion of vector coordinates between RAS coordinate system (in NIFTI file) and
+   * LPS (ITK convention) when reading or writing a "displacement vector" file (intent code: 1006).
+   *
+   * This flag is enabled by default. Disabling it may be useful to avoid unnecessary conversions
+   * in applications that uses RAS coordinate system.
+   *
+   * To write a vector image as displacement vector: set "intent_code" to "1006" in the metadata
+   * dictionary of the input image.
+   *
+   \code
```
jhlegarreta, line 212: Not sure if the preferred style in ITK (if Doxygen supports it) is to make this lines start with the asterisk as well.

jhlegarreta (commented):
Sorry :pray: for another review on the documentation Andras.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
lassoan: This is a local helper function as the others around it and all the others near the top of the file. It does not have a separate declaration, so I think the comment's location is appropriate. This is the convention that is already used in the short documentation of the other similar functions.

Modules/IO/NIFTI/include/itkNiftiImageIO.h
N-Dekker: @jhlegarreta  Code examples between \code and \endcode should _not_ have each code line start with `*`. That's OK!

lassoan: This is how `\code` is added in all .h files in ITK.

Modules/IO/NIFTI/test/itkNiftiImageIOTest3.cxx
lassoan: The main idea is to force to use the displacement field writing/reading branch of the code, which can be triggered by setting this intent code. But adding the suggested test is easy enough, so I've added it now.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -521,6 +524,36 @@ CastCopy(float * to, const void * from, size_t pixelcount)
...
+template <typename TBuffer>
+void
+ConvertRASToFromLPS_XYZTC(TBuffer * buffer, size_t size)
```
N-Dekker, line 546: @lassoan Can you please place both `ConvertRASToFromLPS_CXYZT` and `ConvertRASToFromLPS_XYZTC` inside an unnamed namespace? (Possibly nicely together in the same namespace.) That's just where any local helper function that is defined in a cxx file is supposed to be. Sorry it's not my fault, blame C++  😺 

Just a matter of adding right before their function definition:

    namespace
    {

and right afterwards:

    } // namespace

So the code itself does not need to move.

dzenanz (approved):

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
lassoan: I was surprised why all the existing helper functions there were not in an unnamed namespace, but I wanted to keep changes to existing code to the minimum. I've added the namespace now.

jhlegarreta (approved):

N-Dekker (approved):
Thanks for addressing my commits @lassoan. Your pull request looks fine to me now  👍

jhlegarreta:
Valgrind is showing some memory defects related to the NIFTI IO class. Not sure which is the change that triggered them, as they have only started appearing today (Valgrind builds have been green for 3 days after this branch was merged):
https://open.cdash.org/viewDynamicAnalysisFile.php?id=9949304

Valgrind points to
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx#L708

Which was not modified in this PR in reality.

The commits on Jan 26-28 do not seem to introduce related changes (unsure if #3852 has side-effects).

jhlegarreta:
https://github.com/InsightSoftwareConsortium/ITK/pull/3884#issuecomment-1408710644 Looks like it was a hiccup from Valgrind: no defects have been notified since 3 days ago. The comment can thus be ignored. Sorry for the noise.

lassoan:
Thanks for the update. I reviewed NIFTI IO class and noticed that the implementation was kind of fragile, because memory allocation happened very far from the place where the buffer is used; but I did not see any specific issue. So, I'm glad it turned out that there was no real failure.

jhlegarreta:
https://github.com/InsightSoftwareConsortium/ITK/pull/3884#issuecomment-1415981234 Thanks for the effort Andras. Maybe something to watch then; if noticed again, maybe we can open an issue.

blowekamp:
@jhlegarreta The Valgrind run I saw had many tests failing. This can indicate some type of abnormal execution of the tests. In such a case, the processes may be terminated prematurely ( due to some OS resource limitation encountered) where the memory will not be cleaned up correctly. This can trigger false positives of valgrind defects.

Therefore, if there are abnormal or inconsistent number of tests failing during valgrind then defects should not be considered actionable, just monitored to see if they are consistent.

PR #3887, opened 2023-01-26
BUG: SetRegions in image_from_dict

thewtex:
The Regions, BufferedRegion, LargestPossibleRegion, RequestedRegion, should all be the full image, given by the image_dict["size"]. In itk-wasm, it was observed that this could not be initialized.

### DIFF
--- a/Wrapping/Generators/Python/Tests/extras.py
+++ b/Wrapping/Generators/Python/Tests/extras.py
@@ -200,6 +200,12 @@ itk.meshwrite(mesh, sys.argv[5])
 itk.meshwrite(mesh, sys.argv[5], compression=True)
 
 # smoke test wasm / Python / NumPy conversion
+
+image_dict = itk.dict_from_image(image)
+image_back = itk.image_from_dict(image_dict)
+diff = itk.comparison_image_filter(image.astype(itk.F), image_back.astype(itk.F))
+assert np.sum(diff) == 0
+
 mesh_dict = itk.dict_from_mesh(mesh)
 mesh_back = itk.mesh_from_dict(mesh_dict)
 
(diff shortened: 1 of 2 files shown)


PranjalSahu (approved):
LGTM

Wrapping/Generators/Python/Tests/extras.py
```diff
@@ -200,6 +200,10 @@ def custom_callback(name, progress):
 itk.meshwrite(mesh, sys.argv[5], compression=True)
 
 # smoke test wasm / Python / NumPy conversion
+
+image_dict = itk.dict_from_image(image)
+image_back = itk.image_from_dict(image_dict)
```
tbirdso, line 205: For improved coverage would it make sense to add a couple of basic checks to validate that `image` and `image_back` metadata are equivalent?

dzenanz (approved):

hjmjohnson (approved):

Wrapping/Generators/Python/Tests/extras.py
thewtex: Yes, I will add a call to `itk.comparison_image_filter`.

tbirdso: Looks great, thanks.

On review my understanding is that [itk::Testing::ComparisonImageFilter](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/TestKernel/include/itkTestingComparisonImageFilter.h) directly compares image buffers and does not validate that metadata is equivalent between two images. For better coverage overall, is there a filter that does both?

Validating buffers will cover the `SetRegions` fix introduced in this PR, so for the purposes of this individual fix it is fine to move forward with `ComparisonImageFilter`.

thewtex: ComparisonImageFilter checks metadata per: https://github.com/InsightSoftwareConsortium/ITK/blob/d007ee8e78ab367d10a4fc116fceb4ba07604174/Modules/Core/TestKernel/include/itkTestingComparisonImageFilter.hxx#L48

tbirdso: I missed that, thanks!

PR #3896, opened 2023-01-31
STYLE: Use range-based `for` in MultipleLogOutput and LoggerManager

N-Dekker:
Also defaulted the default-constructor and destructor of MultipleLogOutput.

### DIFF
--- a/Modules/Core/Common/src/itkMultipleLogOutput.cxx
+++ b/Modules/Core/Common/src/itkMultipleLogOutput.cxx
@@ -20,15 +20,9 @@
 
 namespace itk
 {
-MultipleLogOutput::MultipleLogOutput()
-{
-  this->m_Output.clear();
-}
+MultipleLogOutput::MultipleLogOutput() = default;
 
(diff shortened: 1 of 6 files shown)


Modules/Core/Common/src/itkMultipleLogOutput.cxx
```diff
@@ -20,15 +20,9 @@
 
 namespace itk
 {
-MultipleLogOutput::MultipleLogOutput()
-{
-  this->m_Output.clear();
```
dzenanz, line 25: I assume this was a no-op?

dzenanz (approved):

Modules/Core/Common/src/itkMultipleLogOutput.cxx
N-Dekker: Indeed! m_Output is an `std::set`, so its default-constructor already creates an empty container.

dzenanz (approved):

thewtex (approved):
:100:

PR #3872, opened 2023-01-14
STYLE: Make `PrintSelf` implementation style consistent

jhlegarreta:
Make `PrintSelf`implementation style consistent following the ITK SWG
coding style guideline and the available ITK macros:
- Always print the superclass first.
- Print only the class instance variables.
- Use the `os << indent << "{ivarName}: " << m_ivarName << std::endl`
  recipe: do not print the leading `m_` of the ivar; use a colon after
  printing its name; leave a whitespace between the colon and the
  printed value of the ivar.
- For boolean ivars print `On`/`Off` according to their values using the
  ternary operator.
- Use `itkPrintSelfObjectMacro` to print objects inheriting from
  `itk::LightObject` that can be null.
- For pointers that do not inherit from `itk::LightObject` check whether
  they are non-null before attempting to print them and print the
  `(empty)` string otherwise.
- Rely on the self printing ability of the ITK classes to avoid
  boilerplate code when printing the corresponding ivars in other
  classes' `PrintSelf` methods.
- Take advantage of the output stream `<<` insertion operator overload
  for `std::vector` types in the `itk::print_helper` namespace defined
  in `itkPrintHelper.h` to print such types consistently.
- Cast custom types to their numerical values using
  `itk::NumericTraits::PrinType` so that they are printed as expected.
- Remove statements printing the hard-coded class name and its own
  pointer, e.g.
  ```
  os << indent << "ArrowSpatialObject(" << this << ")" << std::endl;
  ```
  `PrintSelf` does this without requiring the removed line.
- Save typing `this->` to point to the instance of the current class.

Apply the same principles to the `<<` operator overloads.

Overload the ostream operator for the `FixedImageSamplePoint` class used
by the `itk::ImageToImageMetric` class.

Overload the ostream operator for the `ImageVoxel` class used by the
`itk::DeformableSimplexMesh3DGradientConstraintForceFilter` class.

Overload the ostream operator for the `ImageVoxel` class used by the
`itk::DeformableSimplexMesh3DGradientConstraintForceFilter` class.
## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [X] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)

PR #3909, opened 2023-02-04
ENH: Use a macro to print numeric traits values of objects

jhlegarreta:
Use a macro to print numeric traits values of objects. Avoids boilerplate code.

Printed instances found using the regex:
```
os << indent << "(.*): " << static_cast<typename NumericTraits<(.*)>::PrintType>(m_(.*)) << std::endl;
```

and substituted using itkPrintSelfNumericTraitsMacro\($3\, \$2\);.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkMacro.h
+++ b/Modules/Core/Common/include/itkMacro.h
@@ -1271,6 +1271,14 @@ compilers.
   ITK_MACROEND_NOOP_STATEMENT
 
 
+// A useful macro in the PrintSelf method for printing member variables
+// that require to be casted to the corresponding numeric value so that
+// their numeric value gets printed.
+#define itkPrintSelfNumericTraitsMacro(name, type)                                                                    \
+  os << indent << #name << ": " << static_cast<typename NumericTraits<type>::PrintType>(this->m_##name) << std::endl; \
+  ITK_MACROEND_NOOP_STATEMENT
+
+
 /** Set a decorated output. This defines the Set"name"() and a Set"name"Output() method */
 #define itkSetDecoratedOutputMacro(name, type)                                                                       \
   virtual void Set##name##Output(const SimpleDataObjectDecorator<type> * _arg)                                       \


Modules/Core/Common/include/itkMacro.h
```diff
@@ -1271,6 +1271,14 @@ compilers.
   ITK_MACROEND_NOOP_STATEMENT
 
 
+// A useful macro in the PrintSelf method for printing member variables
+// that require to be casted to the corresponding numeric value so that
+// their numeric value gets printed.
+#define itkPrintSelfNumericTraitsMacro(name, type)                                                                    \
+  os << indent << #name << ": " << static_cast<typename NumericTraits<type>::PrintType>(this->m_##name) << std::endl; \
+  ITK_MACROEND_NOOP_STATEMENT
```
N-Dekker, line 1279: Thanks for this idea, Jon, but I would prefer to only make _limited_ use of macro's. In general, I think macro's make code harder to understand, and harder to debug. Specifically in this case I would at least suggest to take `os` and `indent` out of the macro definition, because it is unclear from the outside that a call to this macro "captures" these two variables.

Last December I already suggested a different approach for the implementation of those PrintSelf member functions (but I did not have time to make a PR in that direction: https://github.com/InsightSoftwareConsortium/ITK/pull/3802#discussion_r1053351196

jhlegarreta: Niels, it is still WIP, but here my few thoughts:
- After #3872, I think it becomes clear that having a macro saves typing/time and allows to print all members with a consistent style: I'd dare to say that most of the times what this macro attempts to replace has not been done because it requires a non-negligible amount of typing for what can be considered to be a low reward effort at first sight, but it pays off when debugging issues, as @blowekamp made it clear in https://github.com/InsightSoftwareConsortium/ITK/pull/3872#issuecomment-138553199.
- The macro follows the `itkPrintSelfObjectMacro` philosophy, and in the same way, the `os` and `indent` are not taken as parameters as they do not change. But that is the least of the issues for now.
 - The macro is very easy to understand in this case, as it is in the `itkPrintSelfObjectMacro` case.

issakomi:
BTW, almost always the cast to PrintType is useless, the difference is only for _char_ types (their print type is _int_) and for _long double_ for Sun Studio 32 bit only (cast to _double_ to avoid bug). Just FYI, there is the C++11 trick to avoid useless casts. But it really just FYI, not a request to implement.

```cpp
#include <type_traits>
#include <iostream>

template<typename T, typename U, std::enable_if_t<!std::is_same<T, U>::value, int> = 0>
T conditional_static_cast(U value) noexcept
{
  std::cout << "cast" << std::endl;
  return static_cast<T>(value);
}

template<typename T, typename U, std::enable_if_t<std::is_same<T, U>::value, int> = 0>
T conditional_static_cast(U value) noexcept
{
  std::cout << "no cast" << std::endl;
  return value;
}

int main(int,char**)
{
  unsigned int x{ 1U };
  unsigned long y{ 1UL };
  unsigned long long z{ 1ULL };
  size_t a = conditional_static_cast<size_t, unsigned int>(z);
  size_t b = conditional_static_cast<size_t, unsigned long>(x);
  size_t c = conditional_static_cast<size_t, unsigned long long>(y);
  return 0;
}
```

jhlegarreta:
Thanks for the comment @issakomi. As one never knows beforehand which `*Type` aliases may end up adopting such types, and given our extensive use of aliases, I think what you propose will be useful to take into account for the macro in case it can be accommodated into it, and probably another reason to use it in order to save typing.

issakomi:
Some care may be required for complex types in the macro, BTW,
```cpp
template <typename TComponent>
class NumericTraits<std::complex<TComponent>>
{
public:
  using Self = std::complex<TComponent>;
  // for backward compatibility
  using TheType = Self;
  using ValueType = TComponent;
  using PrintType = Self;
```

hjmjohnson:
Superceeded by https://github.com/InsightSoftwareConsortium/ITK/pull/6234

PR #3914, opened 2023-02-06
BUG: Support heterogeneous inputs/outputs for PyImageFilter

thewtex:
Addresses needs identified itk-cucim.


thewtex:
@grlee77 please take a look :pray:

grlee77 (approved):
Thanks! If I understand this correctly, instead of 5 supported pixel types matching at input and output, this will now generate all 25 possible input/output pixel type combinations.

dzenanz (approved):
@grlee77 yes, but other filters might want some different combination of types in the future. As PyImageFilter is quite generic, it makes sense to do all the combinations.

PR #3904, opened 2023-02-04
ENH: Remove `itkMath.h` duplicate include

jhlegarreta:
Remove `itkMath.h` duplicate include from
`itk::ScalarImageToRunLengthMatrixFilter.h`.

## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [X] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)

PR #3919, opened 2023-02-09
ENH: Remove `itkMath.h` duplicate include

jhlegarreta:
Remove `itkMath.h` duplicate include from
`itk::fem::FEMRegistrationFilter.h`.

Left behind in 48025bd985d17c53cb11ca4dc739cb83d4cc4802.

## PR Checklist


dzenanz (approved):

N-Dekker (approved):
Thanks Jon. Again, I think this is not an "ENH", instead, this is a "STYLE" commit. Also I would like the subject to be more specific. For example:

    STYLE: Remove duplicate #include "itkMath.h" in itkFEMRegistrationFilter

Anyway, no big deal of course. Approved anyway!

jhlegarreta:
> Thanks Jon. Again, I think this is not an "ENH", instead, this is a "STYLE" commit. Also I would like the subject to be more specific. For example:

The previous PR #3904 was marked as `ENH`. I'd say that this could be a `COMP` one (but my compiler does not flag duplicate includes so I cannot put the warning) or maybe a `PERF` one (even if probably the compiler will optimize this under the hood and the performance penalty would be unnoticeable anyways), but not a `STYLE` one.

Anyways, I made the subject more specific and changed the filename in the commit message which pointed to the header file instead of the implementation file.

thewtex (approved):
:+1:

N-Dekker:
> The previous PR #3904 was marked as `ENH`. I'd say that this could be a `COMP` one (but my compiler does not flag duplicate includes so I cannot put the warning) or maybe a `PERF` one (even if probably the compiler will optimize this under the hood and the performance penalty would be unnoticeable anyways), but not a `STYLE` one.

To me this is obviously just a code cleanup, so definitely `STYLE`. I already commented about it at https://github.com/InsightSoftwareConsortium/ITK/pull/3904#pullrequestreview-1284089534  `ENH` should be used when a new feature is added. `PERF` when a significant runtime performance improvement is expected. If we still want to use those commit labels anyway.

> I made the subject more specific and changed the filename in the commit message which pointed to the header file instead of the implementation file.

OK, thanks.

PR #3929, opened 2023-02-16
Default default-constructor of ConstantBoundaryCondition and other Core/Common class templates

N-Dekker:
Follow-up to pull request #3878 "STYLE: Default default-constructors of class templates in Core/Common"

### DIFF
--- a/Modules/Core/Common/include/itkConstNeighborhoodIterator.h
+++ b/Modules/Core/Common/include/itkConstNeighborhoodIterator.h
@@ -98,7 +98,7 @@ public:
   using ImageBoundaryConditionConstPointerType = const ImageBoundaryCondition<ImageType, OutputImageType> *;
 
   /** Default constructor */
-  ConstNeighborhoodIterator();
+  ConstNeighborhoodIterator() = default;
 
   /** Virtual destructor */
   ~ConstNeighborhoodIterator() override = default;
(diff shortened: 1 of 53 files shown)


dzenanz (approved):

Modules/Core/Common/test/itkCompensatedSummationTest2.cxx
```diff
@@ -71,7 +71,6 @@ class CompensatedSummationTest2Associate
     void
     ThreadedExecution(const DomainType & subdomain, const itk::ThreadIdType threadId) override
     {
-      itk::CompensatedSummation<double> compensatedSum;
```
N-Dekker, line 74: Interestingly, the commit which "defaulted" the default-constructor of `itk::CompensatedSummation` helped the compiler (clang) to find this unused variable! 😃  At https://open.cdash.org/viewBuildError.php?type=1&buildid=8482888

Modules/Core/Common/include/itkConstNeighborhoodIterator.h
```diff
@@ -612,11 +612,14 @@ class ITK_TEMPLATE_EXPORT ConstNeighborhoodIterator
...
 
+  /** Default boundary condition. */
+  TBoundaryCondition m_InternalBoundaryCondition{};
+
   /** Pointer to the actual boundary condition that will be used.
    * By default this points to m_BoundaryCondition, but
    * OverrideBoundaryCondition allows a user to point this variable an external
    * boundary condition.  */
-  ImageBoundaryConditionPointerType m_BoundaryCondition{};
+  ImageBoundaryConditionPointerType m_BoundaryCondition{ &m_InternalBoundaryCondition };
```
dzenanz, line 622: Is there some guarantee that `m_InternalBoundaryCondition` will be initialized before `m_BoundaryCondition`? Maybe because it was declared before? Or some other reason?

dzenanz (approved):
I love this simplification.

Modules/Core/Common/include/itkConstNeighborhoodIterator.h
N-Dekker: Strictly speaking, I think it doesn't matter if m_InternalBoundaryCondition is already initialized. Even before it is initialized, its address `&m_InternalBoundaryCondition` should already be OK. 

I just moved the m_InternalBoundaryCondition declaration upward, before using `&m_InternalBoundaryCondition`, to avoid any confusion. But I don't think it really matters. I think it's just OK  😃 Thanks for asking!

PR #3930, opened 2023-02-17
COMP: Exclude README, Utilities/Maintenance and other files from AZP CI

N-Dekker:
Excluded various files and directories from triggering a CI run, according to Azure Pipelines "trigger definition" at https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/trigger?view=azure-pipelines#trigger-batch-branches-paths-tags

These are files that are not tested by the CI anyway.

This pull request aims to avoid running the virtual machines of the CI unnecessarily, to save energy and to reduce the time waiting for the CI.

### DIFF
--- a/Testing/ContinuousIntegration/AzurePipelinesLinux.yml
+++ b/Testing/ContinuousIntegration/AzurePipelinesLinux.yml
@@ -5,6 +5,21 @@ trigger:
     include:
     - master
     - release*
+  paths:
+    exclude:
+    - '*.md'
+    - LICENSE
+    - NOTICE
+    - Documentation/*
+    - Utilities/Maintenance/*
+pr:
+  paths:
+    exclude:
+    - '*.md'
+    - LICENSE
+    - NOTICE
+    - Documentation/*
+    - Utilities/Maintenance/*
 variables:
   ExternalDataVersion: 5.3.0
 jobs:
--- a/Testing/ContinuousIntegration/AzurePipelinesLinuxPython.yml
+++ b/Testing/ContinuousIntegration/AzurePipelinesLinuxPython.yml
@@ -5,6 +5,21 @@ trigger:
     include:
     - master
     - release*
+  paths:
+    exclude:
+    - '*.md'
+    - LICENSE
+    - NOTICE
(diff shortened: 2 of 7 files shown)


dzenanz (approved):

N-Dekker:
Looks like it does not work 😢 Maybe I should have used `pr` (as well as `trigger`): https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/pr?view=azure-pipelines#pr-autocancel-branches-paths-drafts

dzenanz:
Maybe this needs to be in master before it has an effect?

N-Dekker:
> Maybe this needs to be in master before it has an effect?

Maybe, but I'm not sure. Usually I think the yml file of a branch _does_ affect the CI build of that particular branch. I'll still try to tweak the yml files a little bit more.

N-Dekker:
Is it possible to share one and the same list of excluded files between all (seven) AzurePipelines*.yml files? Current list of excluded files:

    - CODE_OF_CONDUCT.md
    - CONTRIBUTING.md
    - GettingStarted.md
    - GOVERNANCE.md
    - LICENSE
    - NOTICE
    - README.md
    - Documentation/*
    - Utilities/Maintenance/*

I would have liked "*.md", but that does not seem to be accepted by AZP. 

ITK appears to have a separate AzurePipelines*.yml file for each supported platform.

N-Dekker:
OK, I _think_ this should work, but I have no evidence that it does 🤷

N-Dekker:
For the record, you may also suppress the CI by adding the string `[skip ci]` to your commit message. However, if you do so in a pull request:

> You won't be able to merge the pull request if your repository is configured to require specific checks to pass first. To allow the pull request to be merged you can push a new commit to the pull request without the skip instruction in the commit message.

Source: https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs

dzenanz (approved):

dzenanz:
Maybe we can test whether this works by creating a new PR with that maintenance script you used recently?

PR #3917, opened 2023-02-08
STYLE: Add in-class `{}` initializers to classes with override = default

N-Dekker:
Added in-class `{}` default member initializers to data members of classes that have a destructor declared `override = default`.

Follow-up to pull request https://github.com/InsightSoftwareConsortium/ITK/pull/3913 commit 5c97b7605e5bec4e9443b57314b50d4e090a845e
"STYLE: Add in-class `{}` initializers to classes with virtual functions"

PR #3933, opened 2023-02-22
COMP: Fix uninitialized memory Valgrind defects

jhlegarreta:
- COMP: Fix itkParallelSparseFieldLevelSetImageFilter uninitialized memory 
- COMP: Fix `itk::Histogram` uninitialized memory
- STYLE: Remove an unnecessary static_cast from Histogram::PrintSelf
- STYLE: Prefer in-class `{}` member initializers
- STYLE: Make `PrintSelf` implementation style consistent
- STYLE: Remove duplicate statement in `itk::Histogram::Initialize`

## PR Checklist

### DIFF
--- a/Modules/Numerics/Statistics/include/itkHistogram.hxx
+++ b/Modules/Numerics/Statistics/include/itkHistogram.hxx
@@ -704,8 +704,7 @@ Histogram<TMeasurement, TFrequencyContainer>::PrintSelf(std::ostream & os, Inden
     }
   }
 
-  os << indent << "TempMeasurementVector: "
-     << static_cast<typename NumericTraits<MeasurementVectorType>::PrintType>(m_TempMeasurementVector) << std::endl;
+  os << indent << "TempMeasurementVector: " << m_TempMeasurementVector << std::endl;
   os << indent << "TempIndex: " << static_cast<typename NumericTraits<IndexType>::PrintType>(m_TempIndex) << std::endl;
   os << indent << "ClipBinsAtEnds: " << (m_ClipBinsAtEnds ? "On" : "Off") << std::endl;
 }
--- a/Modules/Numerics/Statistics/include/itkHistogram.hxx
+++ b/Modules/Numerics/Statistics/include/itkHistogram.hxx
@@ -176,8 +176,6 @@ Histogram<TMeasurement, TFrequencyContainer>::Initialize(const SizeType & size)
     this->m_OffsetTable[i + 1] = num;
   }
 
-  this->m_TempIndex.SetSize(this->GetMeasurementVectorSize());
-
   m_NumberOfInstances = num;
 
   // adjust the sizes of min max value containers
--- a/Modules/Numerics/Statistics/include/itkHistogram.hxx
+++ b/Modules/Numerics/Statistics/include/itkHistogram.hxx
@@ -196,7 +196,10 @@ Histogram<TMeasurement, TFrequencyContainer>::Initialize(const SizeType & size)
 
   // initialize auxiliary variables
   this->m_TempIndex.SetSize(this->GetMeasurementVectorSize());
+  m_TempIndex.Fill(0);
+
   this->m_TempMeasurementVector.SetSize(this->GetMeasurementVectorSize());
+  m_TempMeasurementVector.Fill(0);
 
   // initialize the frequency container
   m_FrequencyContainer->Initialize(this->m_OffsetTable[this->GetMeasurementVectorSize()]);
--- a/Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.h
+++ b/Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.h
@@ -124,7 +124,7 @@ public:
   Print(std::ostream & os, Indent indent) const;
 
 private:
-  char                      m_Pad1[128];
+  char                      m_Pad1[128]{};
   unsigned int              m_Size;
   RadiusType                m_Radius;
   std::vector<unsigned int> m_ArrayIndex;
@@ -133,7 +133,7 @@ private:
   /** An internal table for keeping track of stride lengths in a neighborhood,
    *  i.e. the memory offsets between pixels along each dimensional axis. */
   unsigned int m_StrideTable[Dimension];
-  char         m_Pad2[128];
+  char         m_Pad2[128]{};
 };
 
 /**
--- a/Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.h
+++ b/Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.h
@@ -125,14 +125,14 @@ public:
 
 private:
   char                      m_Pad1[128]{};
-  unsigned int              m_Size;
-  RadiusType                m_Radius;
-  std::vector<unsigned int> m_ArrayIndex;
-  std::vector<OffsetType>   m_NeighborhoodOffset;
+  unsigned int              m_Size{ 2 * Dimension };
+  RadiusType                m_Radius{};
+  std::vector<unsigned int> m_ArrayIndex{};
+  std::vector<OffsetType>   m_NeighborhoodOffset{};
 
   /** An internal table for keeping track of stride lengths in a neighborhood,
    *  i.e. the memory offsets between pixels along each dimensional axis. */
-  unsigned int m_StrideTable[Dimension];
+  unsigned int m_StrideTable[Dimension]{};
   char         m_Pad2[128]{};
 };
 
--- a/Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.hxx
+++ b/Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.hxx
@@ -82,7 +82,7 @@ ParallelSparseFieldCityBlockNeighborList<TNeighborhoodType>::Print(std::ostream 
 
   os << "ParallelSparseFieldCityBlockNeighborList: " << std::endl;
 
-  os << indent << "m_Pad1: " << m_Pad1 << std::endl;
+  os << indent << "Pad1: " << m_Pad1 << std::endl;
   os << indent << "Size: " << m_Size << std::endl;
   os << indent << "Radius: " << m_Radius << std::endl;
   os << indent << "ArrayIndex: " << m_ArrayIndex << std::endl;
--- a/Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.hxx
+++ b/Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.hxx
@@ -49,7 +49,6 @@ ParallelSparseFieldCityBlockNeighborList<TNeighborhoodType>::ParallelSparseField
   NeighborhoodType it(m_Radius, dummy_image, dummy_image->GetRequestedRegion());
   nCenter = it.Size() / 2;
 
-  m_Size = 2 * Dimension;
   m_ArrayIndex.reserve(m_Size);
   m_NeighborhoodOffset.reserve(m_Size);
 


jhlegarreta:
Not sure the second commit actually fixes the relevant issues.

dzenanz (approved):

Modules/Numerics/Statistics/include/itkHistogram.hxx
```diff
@@ -701,8 +701,7 @@ Histogram<TMeasurement, TFrequencyContainer>::PrintSelf(std::ostream & os, Inden
     }
   }
 
-  os << indent << "TempMeasurementVector: "
-     << static_cast<typename NumericTraits<MeasurementVectorType>::PrintType>(m_TempMeasurementVector) << std::endl;
+  os << indent << "TempMeasurementVector: " << m_TempMeasurementVector << std::endl;
```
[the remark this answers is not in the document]
dzenanz: Calling it a STYLE commit is fine, too.

N-Dekker:
The first commit "COMP: Fix itkParallelSparseFieldLevelSetImageFilter uninitialized memory" certainly looks fine to me 👍 

Apparently my previous bulk commits (PR #3851, #3913, #3917) don't address _all_ unitialized data member problems. That is because they only address classes that have virtual member functions or a `itkNewMacro`, `itkSimpleNewMacro`, or `itkCreateAnotherMacro` macro call.  `ParallelSparseFieldCityBlockNeighborList` appears to have neither of them!

Modules/Numerics/Statistics/include/itkHistogram.hxx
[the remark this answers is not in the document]
jhlegarreta: Done in 1e56f7c.

issakomi:
The size of array _m_TempMeasurementVector_ is set in _Initialize_ function at [the line 199](https://github.com/InsightSoftwareConsortium/ITK/blob/f0d342e0b81e690194cd3fe993316bca7570402b/Modules/Numerics/Statistics/include/itkHistogram.hxx#L199) in _itkHistogram.hxx_, this line is **the origin** of the uninitialized value(s). Probably _m_TempMeasurementVector_ is sometimes not used after, i guess, don't know exactly.
```cpp
  this->m_TempMeasurementVector.SetSize(this->GetMeasurementVectorSize());
```

The {} in [ .h file](https://github.com/InsightSoftwareConsortium/ITK/blob/f0d342e0b81e690194cd3fe993316bca7570402b/Modules/Numerics/Statistics/include/itkHistogram.h#L524) seems to have no effect. Values could be initialized with e.g. (after the line mentioned above).

```cpp
  for (size_t x = 0; x < m_TempMeasurementVector.size(); ++x)
  {
    m_TempMeasurementVector[x] = 0;
  }
```

It seems to fix the defect, but, honestly, i don't know is it a good idea or not. Just FYI. Not a suggestion.

BTW, another thing, s. Histogram.hxx
`this->m_TempIndex.SetSize(this->GetMeasurementVectorSize());`
is done two times in the same function, s. line 179 and 198.

issakomi:
Private variables _m_Pad1_ and _m_Pad2_ (char[128]) seem to be unused. Other `{}`-d variables are initialized in the constructor to different values.

N-Dekker:
>   for (size_t x = 0; x < m_TempMeasurementVector.size(); ++x)
  {
    m_TempMeasurementVector[x] = 0;
  }

FWIW, m_TempMeasurementVector seems to be an `itk::Array`, so `m_TempMeasurementVector.Fill(0)` could also work.  Anyway, good catch @issakomi !

N-Dekker:
> Private variables _m_Pad1_ and _m_Pad2_ (char[128]) seem to be unused.

If they are indeed unnecessary, please remove those m_Pad... members!

jhlegarreta:
Thanks for investigating the valgrind defects with `itk::Histogram` @issakomi :100:. I will try to review your comments and see if I can come up with a patch set as time permits.

> Private variables m_Pad1 and m_Pad2 (char[128]) seem to be unused.

Thanks. Will have a look at them and remove if appropriate. 

> Other {}-d variables are initialized in the constructor to different values.

Will see if I have the time to look if the initialization can be done in the header file.

dzenanz:
> to avoid "false sharing"

This is the more likely reason. Not printing the pad members sounds like the right thing. And then they might not need to be initialized?

I did only a glancing review initially. It is good that other people review as well.

jhlegarreta:
> . @jhlegarreta Do you really need the bits of m_Pad1 and m_Pad2 to be printed? Looking at PR https://github.com/InsightSoftwareConsortium/ITK/pull/3872 commit https://github.com/InsightSoftwareConsortium/ITK/commit/c47ed1c1b3e26051de655f7ad03e975fae93b73b

I'd say that it is not a matter of preference: if the person that designed the class thought/required a given variable to be a member, then the ITK standard is to have it printed as any other member (unless static). Note that I did not read the body of the methods to determine in which ways the variable is used when addressing the defect/when printing the variable.

The standard could be improved if we come up with a rationale as to why a given variable should not be printed, and a way for automatically marking such variables for the purpose you describe Niels (a no-op macro if that makes sense/serves the purpose).

In this case, and until we propose something, I can comment both lines to serve as a reminder, or else I can remove them.

> And then they might not need to be initialized?

I would keep the initialization to avoid issues. But as you wish.

dzenanz:
Initializing them seems safer, even if not necessary.

jhlegarreta:
A few notes:
- Have been thinking about the false sharing hypothesis: why use 2 variables instead of only 1?
- Have found that `itk::SparseFieldLevelSetImageFilter` has a similar class named `itk::SparseFieldCityBlockNeighborList` tha does not have the `m_Pad1`, `m_Pad2` variables: https://github.com/InsightSoftwareConsortium/ITK/blob/c44f9c1b39e627c5ef7fabec252232a92d570565/Modules/Segmentation/LevelSets/include/itkSparseFieldLevelSetImageFilter.h#L118

So maybe the variables were actually intended to be used. So for now, until we come up with more complete idea, I'll have them printed.

@issakomi @N-Dekker @dzenanz thanks for all the pointers.

issakomi:
FYI, this padding is there since the first 'parallel' version of the filter (2003). S. [here](https://github.com/InsightSoftwareConsortium/ITK/blob/f44771c712cc8eabad774e72734dc5f02a90c9fe/Code/BasicFilters/itkParallelSparseFieldLevelSetImageFilter.h#L601)
```
  struct ThreadData
  {
    char pad1 [128];
...
    char pad2 [128];
  };
  
  /** An array storing the individual structures for each thread. */
  ThreadData *m_Data;
```

Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.h
```diff
@@ -124,16 +124,16 @@ class ITK_TEMPLATE_EXPORT ParallelSparseFieldCityBlockNeighborList
...
 
 private:
-  char                      m_Pad1[128];
-  unsigned int              m_Size;
-  RadiusType                m_Radius;
-  std::vector<unsigned int> m_ArrayIndex;
-  std::vector<OffsetType>   m_NeighborhoodOffset;
+  char                      m_Pad1[128]{};
+  unsigned int              m_Size{ 2 * Dimension };
+  RadiusType                m_Radius{ 1 };
```
issakomi, line 129: The segfault may be related (not tested), better let in ctor as is. It is itk::Size from _Neighborhood_.
`using RadiusType = itk::Size<VDimension>;`

jhlegarreta: Thanks for the investigation @issakomi. Will amend the commit tonight.

jhlegarreta: Done in d741ad0.

dzenanz (approved):

Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.h
```diff
@@ -124,16 +124,16 @@ class ITK_TEMPLATE_EXPORT ParallelSparseFieldCityBlockNeighborList
   Print(std::ostream & os, Indent indent) const;
 
 private:
-  char                      m_Pad1[128];
-  unsigned int              m_Size;
-  RadiusType                m_Radius;
-  std::vector<unsigned int> m_ArrayIndex;
-  std::vector<OffsetType>   m_NeighborhoodOffset;
+  char                      m_Pad1[128]{};
```
N-Dekker, line 127: @jhlegarreta @dzenanz Please wait, before merging. I'm still considering to propose adjusting the ITK Coding Style Guide, to exclude padding bytes from `{}` initialization.

N-Dekker: Maybe we could make it clearer that these are just padding bytes that do not need to be initialized by renaming `m_Pad1` and `m_Pad2` to something like `m_UninitializedPaddingBytes1` and  `m_UninitializedPaddingBytes2`...? We could also possibly make it clearer by declaring them as arrays of `byte`, instead of arrays of `char`. C++17 defines an `std::byte` type that we can easily borrow: https://en.cppreference.com/w/cpp/types/byte

hjmjohnson: @N-Dekker   I agree with our last comment here.  the m_Pad1 is not really a state variable of the class.  It is a solution to a problem that happens to use a char array to enforce performance gains, but does not represent the state of the class.

These pad variables should have good documentation that they are excluded from the "standard rules" applied to member variables that represent the state of the object.

dzenanz: When did we move to C++14? Maybe it is time to move to C++17?

N-Dekker: - @dzenanz Check pull request #2563
C++17 has got some really great improvements. 😃

dzenanz: That was almost 2 years ago. I suggest making a forum topic about move to C++17, to see what others think about it. Perhaps describe implications on compiler support, and other similar considerations in the post.

dzenanz:
@jhlegarreta If this PR is ready for merging, please merge it. If not, please turn it into a draft.

jhlegarreta:
@dzenanz This is ready to be merged on my side; @N-Dekker was reluctant to initialize the `m_Pad1` and `m_Pad2` ivars, and @hjmjohnson also shared that. Niels assumed that those variables intended to avoid false sharing; we do not have any better hypothesis. I ignore if a steady decision has been made; if yes, it should be put in the ITK SW Guide. I am mostly away from keyboard these days, but I would very much like to have this merged as it will solve the Valgrind defects that are present since a month; changes can then be made in a separate PR (also to avoid needing to wait for the C++17 transition if that will affect how these variables are declared and printing the variables in the `PrintSelf` mehod), still ensuring that no new Valgrind defects are present.

N-Dekker:
> I would very much like to have this merged as it will solve the Valgrind defects that are present since a month

Thanks for you analysis, @jhlegarreta Just to recall (for those who did not remember), the undefined behavior appears caused by printing those padding bytes (while they were uninitialized):

https://github.com/InsightSoftwareConsortium/ITK/blob/ef40e98381289bf669be756deeb4d000e990991a/Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.hxx#L86-L93

I'm still in favor of simply removing those print statements. Your PR appears to assume that m_Pad1 and m_Pad2 are zero-terminated character strings. But they are not. They are just raw bytes. Even after adding `{}` initializers, printing them does not yield any useful information. I mean, `os << m_Pad1` always just prints an empty string.

jhlegarreta:
> Your PR appears to assume that m_Pad1 and m_Pad2 are zero-terminated character strings

My PR does no assumption on their utility and follows the current ITK SW Guide guidelines to print all non-static ivars.

> I mean, os << m_Pad1 always just prints an empty string,

Simply because they are not used elsewhere, and we are only guessing about their utility.

N-Dekker:
Just to explain what I mean by saying that PR (and specifically the print statements) appears to assume that m_Pad1 and m_Pad2 are always zero-terminated. If these arrays were filled with (only) _non-zero_ values, printing them by `os << m_Pad1` and `os << m_Pad2`  would still cause undefined behavior.

Example:

    char arr[3] = { '1', '2', '3' };
    std::cout << arr;  // Undefined behavior, arr is not zero-terminated!

Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.h
dzenanz: This PR has other improvements which are not contentious. I will merge this. @N-Dekker can then make a follow-up PR which just deals with Padding bytes.

dzenanz:
Padding bytes to be reassessed in a new PR. @N-Dekker if you do not plan to do that soon (maybe as part of C++17 PR which allows use of `alignas`), it would be good to create an issue so we don't forget about it. I think there are some other instances of padding bytes.

PR #3937, opened 2023-02-25
COMP: Move RayCastHelper from unnamed namespace to private class section

N-Dekker:
Put the RayCastInterpolateImageFunction helper class that is not meant to be exposed to the user inside RayCastInterpolateImageFunction, as a private nested type, instead of in an unnamed namespace. Header files (including *.hxx files) just shouldn't use an unnamed namespace.

Avoids a Clang-Tidy 17.0.0 warning saying:
> warning G253035D5: do not use unnamed namespaces in header files [google-build-namespaces]

Following C++ Core Guidelines, September 23, 2022 "Don’t use an unnamed (anonymous) namespace in a header",
http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf21-dont-use-an-unnamed-anonymous-namespace-in-a-header

### DIFF
--- a/Modules/Core/ImageFunction/include/itkRayCastInterpolateImageFunction.h
+++ b/Modules/Core/ImageFunction/include/itkRayCastInterpolateImageFunction.h
@@ -191,6 +191,9 @@ protected:
   InputPointType      m_FocalPoint{};
   double              m_Threshold{};
   InterpolatorPointer m_Interpolator{};
+
+private:
+  class RayCastHelper;
 };
 } // namespace itk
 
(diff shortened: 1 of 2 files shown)


hjmjohnson (approved):

dzenanz (approved):

N-Dekker:
I guess we could still shorten the identifier from "RayCastHelper" to just "Helper", once it is nested inside `RayCastInterpolateImageFunction`. But that would just be a minor style adjustment that could be done at any time afterwards.

PR #3946, opened 2023-03-04
STYLE: Construct local variables by passing their arguments directly

N-Dekker:
Replaced local variable initializations of the form `T var = T(args)` with the equivalent `T var(args)`, passing the `args` directly to the variable. (These two forms are semantically equivalent, but the latter one appears more to the point.)

Did so by Visual Studio 2022, Find and Replace, using regular expressions:

    Find what:  (\w+)[ ]+(\w+) = \1(\([^\)]+\);)
    Find what:  (\w+)[ ]+(\w+) = \1(\([^\)]+\)\);)
    Find what:  (\w+)[ ]+(\w+) = \1(\(.+,.+\);)
    Replace with:  $1 $2$3

### DIFF
--- a/Modules/Filtering/MathematicalMorphology/include/itkFlatStructuringElement.hxx
+++ b/Modules/Filtering/MathematicalMorphology/include/itkFlatStructuringElement.hxx
@@ -1105,8 +1105,8 @@ FlatStructuringElement<VDimension>::Annulus(RadiusType   radius,
 
   // Define the iterators for each ellipsoid
   using FloodIteratorType = FloodFilledSpatialFunctionConditionalIterator<ImageType, EllipsoidType>;
-  FloodIteratorType itEllipsoidOuter = FloodIteratorType(kernelImage, ellipsoidOuter, seed);
-  FloodIteratorType itEllipsoidInner = FloodIteratorType(kernelImage, ellipsoidInner, seed);
+  FloodIteratorType itEllipsoidOuter(kernelImage, ellipsoidOuter, seed);
+  FloodIteratorType itEllipsoidInner(kernelImage, ellipsoidInner, seed);
   itEllipsoidOuter.SetCenterInclusionStrategy();
   itEllipsoidInner.SetCenterInclusionStrategy();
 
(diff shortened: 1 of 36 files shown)


Modules/Filtering/MathematicalMorphology/include/itkFlatStructuringElement.hxx
```diff
@@ -1105,8 +1105,8 @@ FlatStructuringElement<VDimension>::Annulus(RadiusType   radius,
 
   // Define the iterators for each ellipsoid
   using FloodIteratorType = FloodFilledSpatialFunctionConditionalIterator<ImageType, EllipsoidType>;
-  FloodIteratorType itEllipsoidOuter = FloodIteratorType(kernelImage, ellipsoidOuter, seed);
-  FloodIteratorType itEllipsoidInner = FloodIteratorType(kernelImage, ellipsoidInner, seed);
+  FloodIteratorType itEllipsoidOuter(kernelImage, ellipsoidOuter, seed);
```
dzenanz, line 1108: This looks like it could have prompted this PR 😄

dzenanz (approved):

Modules/Filtering/MathematicalMorphology/include/itkFlatStructuringElement.hxx
N-Dekker: What do you mean? No problem but I don't really get the point of your comment! 😃 

    FloodIteratorType itEllipsoidOuter = FloodIteratorType(kernelImage, ellipsoidOuter, seed);

Is shortened by this PR to:

    FloodIteratorType itEllipsoidOuter(kernelImage, ellipsoidOuter, seed);

Which is nice, right?  😃

dzenanz: Yes, of all the changes are good. But why did you make this PR? The point of my comment: "This could have been the reason", and for good measure you fix all the other occurrences of the same type of annoyance.

N-Dekker: Ah, I see, thanks @dzenanz 

> But why did you make this PR?

I'm thinking of a PR to remove some unnecessary local _Iterator_ typedef's (or `using` aliases), especially when they are used only just once. Example:

    using IteratorType = ImageRegionConstIteratorWithIndex<ImageType>;
    IteratorType it(m_Image, region);

Could be shortened to: 

    ImageRegionConstIteratorWithIndex<ImageType> it(m_Image, region);

Which would help me when trying to find out quickly what kind of iterator `it` is.

However, before doing so, I thought I should first adjust those declarations of the form `IteratorType it = IteratorType(m_Image, region)`, to make sure  `IteratorType` is indeed only used once!

hjmjohnson (approved):

PR #3948, opened 2023-03-06
DOC: Consolidate release notes into Documentation/Releases

thewtex:
Add minimal README.


dzenanz (commented):
What is the advantage of this living in Releases instead of ReleaseNotes? Also, a few [ghostflow complaints](https://github.com/InsightSoftwareConsortium/ITK/runs/11780090834).

thewtex:
This brings them all into one, shorter directory (1.0.md was here already).

Addressing the trailing whitespace noted by ghostflow (not introduced in this commit), but the other size complaints will have to remain.

dzenanz (approved):

PR #3949, opened 2023-03-06
BUG:  Fix sobel m2 Apple Silicon mac failure

hjmjohnson:
## PR Checklist

API for SobelOperator now has a constraint that only signed values are allowed.  Unsigned values produce incorrect results.

 * Sobel Direction 1 coefficients (G_y) convolved with Neighborhood of 1's
      should equal 0
    
             G_y           *      N
      [ +1.   +2.   +1. ]    [ 1  1  1]
      [  0.    0.    0. ]  * [ 1  1  1]    =>   0
      [ -1.   -2.   -1. ]    [ 1  1  1]
    
    If the SobelOperator is templated over a pixel type
    that is not signed, the G_y coefficients above are
    typecast to the internal pixel type, this removing
    the sign, or introducing wrap errors.
    
    Test code:
        double d = -1.;
        unsigned char c = d;
        std::cout << +c << std::endl;
    
    On intel computers:
        prints 255
    On apple silicon computers:
        print 0
    
    NOTE: The sobel coeficients are not properly applied
          during the inner product for unsigned data types.
    
    Turn on concept checking for the SobelOperator to enforce
    signed neighborhood processing.

Previous tests for Sobel tested results with respect to underflow-generated images.  Those images did not look like typical sobel image results, so rescaled the signed values into valid PNG ranges for difference and display..  

@jhlegarreta  Updated the sobel tests you added in November to deal with signed outputs in a more meaningful way.
@seanm This change may address some of the mac m1 failures when building with Debug mode.


Modules/Core/Common/include/itkNeighborhoodInnerProduct.hxx
```diff
@@ -29,18 +29,15 @@ NeighborhoodInnerProduct<TImage, TOperator, TComputation>::Compute(const ConstNe
...
   using AccumulateRealType = typename NumericTraits<InputPixelRealType>::AccumulateType;
 
-  AccumulateRealType sum = NumericTraits<AccumulateRealType>::ZeroValue();
-
   using OutputPixelValueType = typename NumericTraits<OutputPixelType>::ValueType;
 
-  o_it = op.Begin();
   const typename OperatorType::ConstIterator op_end = op.End();
+  typename OperatorType::ConstIterator       o_it = op.Begin();
+  AccumulateRealType                         sum = NumericTraits<AccumulateRealType>::ZeroValue();
```
dzenanz, line 40: Would `AccumulateRealType sum{};` accomplish this?

Modules/Core/Common/test/itkSobelOperatorImageConvolutionTest.cxx
```diff
@@ -65,19 +68,98 @@ itkSobelOperatorImageConvolutionTest(int argc, char * argv[])
...
+
+    using OutputImageType = itk::Image<uint8_t, Dimension>;
+    // Assume min/max values are approximately +/- same magnitude so that the output images
+    // to be stored in uint8_t have an implied 0 at about pixel value 128.  Many web based viewers
+    // for the difference images in the testing outputs render better in this positive png range.
+    using RescaleIntensityType = itk::RescaleIntensityImageFilter<ImageType, OutputImageType>;
+    RescaleIntensityType::Pointer rescalerForVisualization = RescaleIntensityType::New();
+    rescalerForVisualization->SetInput(signedSobelImage);
+    rescalerForVisualization->SetOutputMinimum(0);
+    rescalerForVisualization->SetOutputMaximum(255);
```
dzenanz, line 158: Why change test output type from png to tiff, if we rescale to 0-255 range?

dzenanz (commented):
Looks good on a glance.

hjmjohnson:
Built on M2 mac with `-mtune=native` and `CMAKE_BUILD_TYPE=Debug`

CDASH PRE COMMIT:  https://open.cdash.org/viewTest.php?onlyfailed&buildid=8525763
CDASH POST COMMIT: https://open.cdash.org/viewTest.php?onlyfailed&buildid=8525768

Modules/Core/Common/test/itkSobelOperatorImageConvolutionTest.cxx
hjmjohnson: Originally I was writing signed sobel outputs, but only Slicer would display those outputs with reasonable visualization.  I wanted the outputs to display in cdash as a better looking (i.e. look like sobel should look), so at the last minute I updated to rescale the images.  

Since the images needed to be made new, I did not think that tiff would be any worse than png.

dzenanz: When I first saw TIFF, and commit message mentioning negative gradients, I thought negative values were going to be in the test images 😄

A bit surprising, that's all.

Modules/Core/Common/include/itkNeighborhoodInnerProduct.hxx
N-Dekker: In general I believe that any `T var = NumericTraits<T>::ZeroValue()` can be replaced safely with `T var{}`.

hjmjohnson: I'm working on a more universal patch to address this particular item.  I'd like to keep the universal change separate from the current patch set.

Note the old code used ZeroValue.

dzenanz (approved):

Modules/Core/Common/include/itkSobelOperator.hxx
```diff
@@ -76,7 +76,9 @@ template <typename TPixel, unsigned int VDimension, typename TAllocator>
 auto
 SobelOperator<TPixel, VDimension, TAllocator>::GenerateCoefficients() -> CoefficientVector
 {
-  std::vector<double> coeff;
+  CoefficientVector     coeff;
+  constexpr std::size_t coeff_size = (VDimension == 2) ? 9 : 27;
+  coeff.reserve(coeff_size);
```
N-Dekker, line 84: The commit message does not mention the `reserve`, is it a significant performance improvement?

N-Dekker: I guess `coeff_size` happens to be 3^N with N = `VDimension`, right? If so, you might consider `coeff_size = itk::Math::UnsignedPower(3, VDimension)`.

hjmjohnson: probably not significant given only 9 or 27 elements.  I forgot that I did that.  I think that the current code better documents the intended behaviors.

hjmjohnson: I added the comment, but I don't want to add an non-constexpr computation for the size until the Sobel operator works for more than 2D and 3D that it is currently hard coded for.

N-Dekker: `Math::UnsignedPower(3, VDimension)` _is_ constexpr  😃 But anyway, thanks!

BTW, if the Sobel  coefficients remain hard coded, it would be more efficient to directly return an initializer list of those coefficients, instead of storing them in a local variable:

    if (VDimension == 2 && this->GetDirection() == 0)
    {
      // No need to reserve or push_back   :-)
      return { -1.0, 0.0, 0.0, 1.0, 1.0, -2.0, -2.0, 0.0, 0.0, 2.0, 2.0, -1.0, -1.0, 0.0, 0.0, 1.0, 1.0 };
    }

But that's probably beyond this PR, no problem!

PR #3951, opened 2023-03-07
BUG: Casting negative numbers to unsigned char error

hjmjohnson:
Casting negative numbers to unsigned char is an undefined behavior.  Apple Silicon results in zero, Intel results in wrap around values.

ctest -V -R itkBSplineScatteredDataPointSetToImageFilterTest05

This test cast an image of Vector<double,3> pixels to and image of Vector<unsigned char, 3> pixels inconsistently across platforms.  Then the test was performed to compute differences of the output images.

The test was comparison was re-written to take the magnitude of the Vector<double,3> pixels and multiplies them into a range of approximately 0-255 for the (now) scalar output image.  This output image is easier to visualize and compute results.

## PR Checklist

Tests are not written robustly.  They depend on consistent behavior when casting from signed double to unsigned char.  That behavior is undefined, and is different on Apple Silicon vs Intel.


hjmjohnson:
I believe these other failing tests will need similar changes. 


itkSymmetricEigenAnalysisImageFilterTestDoNotOrder 
itkSymmetricEigenAnalysisImageFilterTestOrderByMagnitude 
itkSymmetricEigenAnalysisImageFilterTestOrderByValue

Images of floating point vector pixels are cast to unsigned char inconsistently.

hjmjohnson:
@dzenanz  I would like your feedback before I do similar changes to the other 3 failing tests on Apple Silicon.

Modules/Filtering/ImageGrid/test/itkBSplineScatteredDataPointSetToImageFilterTest5.cxx
```diff
@@ -115,19 +112,37 @@ itkBSplineScatteredDataPointSetToImageFilterTest5(int argc, char * argv[])
...
+
+  // phiLatticeImage contains pixels of type Vector<double, 3>
+  // with minmax values of (-5, 5) and magnitude max of about 5
+  auto myiterator =
+    itk::ImageRegionIterator<FilterType::PointDataImageType>(phiLatticeImage, phiLatticeImage->GetBufferedRegion());
+  auto myOutiterator = itk::ImageRegionIterator<OutputImageType>(outputImage, outputImage->GetBufferedRegion());
+  while (!myiterator.IsAtEnd())
+  {
+    // Convert vector image to magnitude and scale to use range of png values
+    constexpr double max_magnitude_value = 5.0;
```
dzenanz, line 135: If magnitude is <= 5, this looks good.

dzenanz (approved):

Modules/Filtering/ImageGrid/test/itkBSplineScatteredDataPointSetToImageFilterTest5.cxx
hjmjohnson: OK.  I think I'll dynamically compute the value.  The 5 is hard-coded here in an attempt to get the draft in front of you. The next version will dynamically compute max_magnitude value.  

I'll also add similar fixes in other failing tests. It might be a few days before I get them done.

Modules/Filtering/ImageGrid/test/itkBSplineScatteredDataPointSetToImageFilterTest5.cxx
```diff
@@ -115,19 +112,46 @@ itkBSplineScatteredDataPointSetToImageFilterTest5(int argc, char * argv[])
...
+  auto myiterator =
+    itk::ImageRegionIterator<FilterType::PointDataImageType>(phiLatticeImage, phiLatticeImage->GetBufferedRegion());
+  auto myOutiterator = itk::ImageRegionIterator<OutputImageType>(outputImage, outputImage->GetBufferedRegion());
 
+  float max_magnitude_value = 0.0;
+  while (!myiterator.IsAtEnd())
+  {
+    // Convert vector image to magnitude and scale to use range of png values
+    const auto magvalue = myiterator.Value().GetNorm();
+    myOutiterator.Set(magvalue);
```
dzenanz, line 138: Doesn't this create weird aliasing artifacts? E.g. if magnitudes are in [-4.5,4.4] range, they will get aliased to just five values (0,1,2,3,4). Later scaling to 0-255 will transform them into (0,56,113,170,226). If we want to avoid recomputing the norms twice, we need an intermediate float image. Also, using a single iterator is single-threaded. Maybe not important for a test?

Modules/Filtering/ImageGrid/test/itkBSplineScatteredDataPointSetToImageFilterTest5.cxx
hjmjohnson: I now dynamically compute the max value.

Modules/Filtering/ImageGrid/test/itkBSplineScatteredDataPointSetToImageFilterTest5.cxx
hjmjohnson: Good catch @dzenanz .  This computation in the test is very fast anyway.  I don't think it is worth doing performance testing.  The entire test only takes 150ms as is.

New patch coming shortly.

dzenanz (approved):
Looks good now. Thanks for fixing these tests!

hjmjohnson:
NOTE: M2 mac build completed successfully with these changes: https://open.cdash.org/viewTest.php?onlypassed&buildid=8530717

PR #3953, opened 2023-03-07
Remove unnecessary and duplicate iterator `GoToBegin()` calls from Modules/Filtering

N-Dekker:
When an iterator is just constructed, it is usually already at its begin position.


dzenanz (approved):
Looks good on a glance.

Was this operation completely manual? Or some combination of a clever search followed by manual changes?

N-Dekker:
> Looks good on a glance.
> Was this operation completely manual?

Thanks @dzenanz For this PR I looked for the regular expression `^  \w+\.GoToBegin\(\);` in .hxx files in ITK\Modules\Filtering. For each `<iterator>.GoToBegin()` call, I checked "manually" if that specific iterator was just constructor (to ensure that it was still at its begin).

hjmjohnson (approved):

PR #3957, opened 2023-03-09
STYLE: Replace `while (!outIt.IsAtEnd())` with `for` loops, in Filtering

N-Dekker:
Looked for iterations in "Modules/Filtering" of the following form:

    OutputIterator outIt(...);
    while (!outIt.IsAtEnd())
    {
      ...
      ++outIt;
    }

And replaced them with the corresponding `for` loops:

    for (OutputIterator outIt(...); !outIt.IsAtEnd(); ++outIt)
    {
      ...
    }

Such `for` loops appear preferable because they limits the scope of those `outIt` variables to where they are actually being used.

### DIFF
--- a/Modules/Filtering/DisplacementField/include/itkInverseDisplacementFieldImageFilter.hxx
+++ b/Modules/Filtering/DisplacementField/include/itkInverseDisplacementFieldImageFilter.hxx
@@ -218,8 +218,6 @@ InverseDisplacementFieldImageFilter<TInputImage, TOutputImage>::GenerateData()
 
   OutputImageRegionType region = outputPtr->GetRequestedRegion();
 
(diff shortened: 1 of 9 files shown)


dzenanz (approved):
I expected more than 10 instances of this.

N-Dekker:
> I expected more than 10 instances of this.

Well... there are many _similar_ cases not yet addressed by this PR. Some having different variable names, others are iterating by both `++` _and_ `NextLine()`. Or having multiple iterators e.g., `inIt` and `outIt`, iterating together in the same loop. But it's a lot of semi-manual work and it must be done very carefully, so with this PR I wanted to just do the `outIt` cases that are really obvious.

hjmjohnson (approved):

PR #3959, opened 2023-03-10
STYLE: Construct local variables by passing arguments directly (part 2)

N-Dekker:
Using the following regular expressions:

    Find what:  (\w[^ ]+)[ ]+(\w+) = \1(\([^\)]+\);)
    Find what:  (\w[^ ]+)[ ]+(\w+) = \1(\([^\)]+\)\);)
    Find what:  (\w[^ ]+)[ ]+(\w+) = \1(\(.+,.+\);)
    Find what:  (\w[^ ]+)[ ]+(\w+) =\r\n[ ]+\1(\(.+,.+\);)
    Find what:  (\w[^ <]+<.+>)[ ]+(\w+) =\r\n[ ]+\1(\([^\)]+\);)
    Find what:  (\w[^ <]+<.+>)[ ]+(\w+) =\r\n[ ]+\1(\(.+,.+\);)
    Replace with:  $1 $2$3

Follow-up to pull request https://github.com/InsightSoftwareConsortium/ITK/pull/3946
commit 9e5dda54257cedd49dae0e9439bd1dbead31fbba

### DIFF
--- a/Modules/Core/Common/test/itkFloodFillIteratorTest.cxx
+++ b/Modules/Core/Common/test/itkFloodFillIteratorTest.cxx
@@ -65,7 +65,7 @@ itkFloodFillIteratorTest(int, char *[])
   std::cout << "New sourceImage allocated" << std::endl;
 
   // Initialize the image to hold all 0's
-  itk::ImageRegionIterator<TImageType> it = itk::ImageRegionIterator<TImageType>(sourceImage, largestPossibleRegion);
+  itk::ImageRegionIterator<TImageType> it(sourceImage, largestPossibleRegion);
 
   for (it.GoToBegin(); !it.IsAtEnd(); ++it)
   {
--- a/Modules/Core/Common/test/itkFloodFilledSpatialFunctionTest.cxx
+++ b/Modules/Core/Common/test/itkFloodFilledSpatialFunctionTest.cxx
@@ -71,7 +71,7 @@ itkFloodFilledSpatialFunctionTest(int, char *[])
   for (int strat = 0; strat < 4; ++strat)
   {
     // Initialize the image to hold all 0's
-    itk::ImageRegionIterator<ImageType> it = itk::ImageRegionIterator<ImageType>(sourceImage, largestPossibleRegion);
+    itk::ImageRegionIterator<ImageType> it(sourceImage, largestPossibleRegion);
 
     for (it.GoToBegin(); !it.IsAtEnd(); ++it)
     {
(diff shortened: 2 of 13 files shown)


hjmjohnson:
@N-Dekker This is really great work.  Much cleaner codebase.

blowekamp:
Any particular reason to choose `()` over `{}` for initialization/construction here?

N-Dekker:
> Any particular reason to choose `()` over `{}` for initialization/construction here?

Good question @blowekamp 

My aim for this PR is just to avoid mentioning the same the type twice, instead of once, with each variable declaration. The `()` parentheses were there already in the original code, so the PR just changes from `T var = T(args)` to `T var(args)`.

If we want to have the `T var{ args }` curly braced syntax throughout the entire code base, I would prefer a separate PR to do so.

dzenanz:
Maybe do a follow-up PR to change to {} syntax soon after this is merged? I assume it would not be easy to find these instances later.

N-Dekker:
> Maybe do a follow-up PR to change to {} syntax soon after this is merged? I assume it would not be easy to find these instances later.

I guess it's not just _these_ instances, right? 

Also I wonder, isn't there some kind of tool to replace `(`args `)` with `{`args`}` all over the code base?

dzenanz:
> some kind of tool to replace (args ) with {args}

Can clang-tidy do it? Sounds like the kind of thing it might do.

N-Dekker:
> Can clang-tidy do it?

Sorry, I cannot find it at https://clang.llvm.org/extra/clang-tidy/checks/list.html

Also I wonder if any compiler or code analyzer produces a warning, when constructing an object by `T var(args)` (using parentheses, rather than curly braces), having multiple arguments.

N-Dekker:
Note that if we want to replace all instances of `T var(arg1, arg2)` with `T var{ arg1, arg2 }`, we should be careful not to change the semantics!  `T var{ arg1, arg2 }` might call a different `T` constructor than `T var(arg1, arg2)`! For example:

    std::vector<int> var1(3, 4); // size = 3, initial value of each element = 4.
    std::vector<int> var2{ 3, 4 }; // size = 2, initial element values 3 and 4.

N-Dekker:
FYI, Peter Sommerlad (@PeterSommerlad) informed me at the ACCU mailing list that [Cevelop](https://www.cevelop.com) does support converting the code from using `()` parentheses to `{}` curly braces for initialization.  It can produce a popup, saying "Replace with uniform variable initialization": 
Cevelop

It can then also convert the code to the new style. 

Not sure if it takes the exception to the guideline [Prefer the `{}`-initializer syntax](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es23-prefer-the--initializer-syntax) into account, which says:

> *Exception*  For containers, there is a tradition for using `{...}` for a list of elements and `(...)` for sizes

PR #3960, opened 2023-03-10
STYLE: Reduce scope ResampleImageFilter::NonlinearThreadedGenerateData

N-Dekker:
Reduced the scope of local `Point` and `ContinuousIndex` variables within `NonlinearThreadedGenerateData`. Note that these variables are very lightweight, so moving their declarations into a `for` loop won't harm the run-time performance.

### DIFF
--- a/Modules/Filtering/ImageGrid/include/itkResampleImageFilter.hxx
+++ b/Modules/Filtering/ImageGrid/include/itkResampleImageFilter.hxx
@@ -384,24 +384,21 @@ ResampleImageFilter<TInputImage, TOutputImage, TInterpolatorPrecisionType, TTran
   // Create an iterator that will walk the output region for this thread.
   using OutputIterator = ImageRegionIteratorWithIndex<TOutputImage>;
 
-  // Define a few indices that will be used to translate from an input pixel
-  // to an output pixel
-  OutputPointType outputPoint; // Coordinates of current output pixel
-  InputPointType  inputPoint;  // Coordinates of current input pixel
-
-  ContinuousInputIndexType inputIndex;
-
   using OutputType = typename InterpolatorType::OutputType;
 
   // Walk the output region
   for (OutputIterator outIt(outputPtr, outputRegionForThread); !outIt.IsAtEnd(); ++outIt)
   {
     // Determine the index of the current output pixel
+
+    OutputPointType outputPoint; // Coordinates of current output pixel
     outputPtr->TransformIndexToPhysicalPoint(outIt.GetIndex(), outputPoint);
 
     // Compute corresponding input pixel position
-    inputPoint = transformPtr->TransformPoint(outputPoint);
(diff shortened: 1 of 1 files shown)


dzenanz (approved):

blowekamp:
I remember spending a good bit of time improving the performance of that loop. Your change is only based on assumptions and not evidence :smile_cat:

N-Dekker:
> But my point is still true. It's a standard optimization guideline to to pull variables outside the loop.

@blowekamp Thanks, but Is that standard optimization guideline officially standardized?  😺 If so, I think it should make an exception for  variables of types that have only _defaulted_ constructors and destructors, only plain old data inside, _and_ no virtual member functions! Or more in general: variables of a [trivial type](https://en.cppreference.com/w/cpp/named_req/TrivialType).

N-Dekker:
Note that strictly speaking, the local `Point` variables (`inputPoint` and `outputPoint`) aren't really necessary. We might as well do something like this: 

    const bool isInsideInput = inputPtr->TransformPhysicalPointToContinuousIndex(
      transformPtr->TransformPoint(outputPtr->TransformIndexToPhysicalPoint(outIt.GetIndex())), inputIndex);

Instead of: https://github.com/InsightSoftwareConsortium/ITK/blob/31c713eb105eff4ba87a457a24a23ef4a6eb3071/Modules/Filtering/ImageGrid/include/itkResampleImageFilter.hxx#L400-L404

What do you think?

blowekamp:
Not sure how to make the transformation not virtual. A new template argument to the ResampleImageFilter would be cumbersome.

There is already the linear method which does some algorithmic optimization for improved performance. There might be an improved algorithm for specific for displacement fields that could be implemented similarly.

I think the one line is much less read-able. And if it is for performance it is just speculation without timings, and may not be consistent between compilers 😸

N-Dekker:
> Not sure how to make the transformation not virtual. A new template argument to the ResampleImageFilter would be cumbersome.

Thanks @blowekamp  I'm thinking, maybe for specific purposes, when the transform type is known at compile time, a `TransformSpecificResampleImageFilter<TTransform, TInputImage, TOutputImage>` could be worth trying. But I agree that it would be cumbersome to add a `TTransform` argument to the existing `itk::ResampleImageFilter` template. 

> I think the one line is much less read-able.

OK, then I won't propose the one-liner `TransformPhysicalPointToContinuousIndex(TransformPoint(TransformIndexToPhysicalPoint(GetIndex())))`.

Still I think the current PR is OK, just moving those three lightweight local variables into the `for` loop. Hope it's acceptable to you as well 😃

dzenanz:
Compiler will most likely produce the same code, due to optimizations. Or quite similar code, at least.

blowekamp:
> Compiler will most likely produce the same code, due to optimizations. Or quite similar code, at least.

Yes, it is likely fine. And even if those variable were intentionally move out of the loop for a measured performance benefit at some point in time, the compilers are not longer the same.

PR #3964, opened 2023-03-11
BUG: implicit conversion undefined behavior

hjmjohnson:
ITK/Modules/IO/TIFF/test/itkLargeTIFFImageWriteReadTest.cxx:143:5:

runtime error: implicit conversion from type 'int' of value 65536 (32-bit, signed) to type 'PixelType' (aka 'unsigned short') changed the value to 0 (16-bit, unsigned)

```
# using ubsan sanitizer flags:
 -fsanitize=undefined -fsanitize=integer -fsanitize=nullability -fno-sanitize-recover=undefined 
```

## PR Checklist
- [X] Updated/Added test (or behavior not changed)

### DIFF
--- a/Modules/IO/Meta/test/itkLargeMetaImageWriteReadTest.cxx
+++ b/Modules/IO/Meta/test/itkLargeMetaImageWriteReadTest.cxx
@@ -45,8 +45,6 @@ ActualTest(std::string filename, typename TImageType::SizeType size)
   typename ImageType::RegionType region;
   typename ImageType::IndexType  index;
 
-  PixelType pixelValue;
-
   itk::TimeProbesCollectorBase chronometer;
 
   { // begin write block
@@ -73,17 +71,14 @@ ActualTest(std::string filename, typename TImageType::SizeType size)
 
     std::cout << "Initializing pixel values " << std::endl;
 
-    IteratorType itr(image, region);
-    itr.GoToBegin();
-
-    pixelValue = itk::NumericTraits<PixelType>::ZeroValue();
-
+    PixelType pixelValue{};
     chronometer.Start("Initializing");
-    while (!itr.IsAtEnd())
+    for (IteratorType itr(image, region); !itr.IsAtEnd(); ++itr)
     {
       itr.Set(pixelValue);
-      ++pixelValue;
-      ++itr;
+      // NOTE: pixelValue for unsigned short will wrap around with large images.
+      pixelValue = (pixelValue != itk::NumericTraits<PixelType>::max()) ? pixelValue + static_cast<PixelType>(1)
+                                                                        : static_cast<PixelType>(0);
     }
     chronometer.Stop("Initializing");
 
@@ -125,16 +120,12 @@ ActualTest(std::string filename, typename TImageType::SizeType size)
 
   typename ImageType::ConstPointer readImage = reader->GetOutput();
 
-  ConstIteratorType ritr(readImage, region);
-
-  ritr.GoToBegin();
 
   std::cout << "Comparing the pixel values..." << std::endl;
 
-  pixelValue = itk::NumericTraits<PixelType>::ZeroValue();
-
+  PixelType pixelValue{};
   chronometer.Start("Compare");
-  while (!ritr.IsAtEnd())
+  for (ConstIteratorType ritr(readImage, region); !ritr.IsAtEnd(); ++ritr)
   {
     if (ritr.Get() != pixelValue)
     {
@@ -143,9 +134,9 @@ ActualTest(std::string filename, typename TImageType::SizeType size)
       std::cerr << "Read Image pixel value " << ritr.Get() << std::endl;
       return EXIT_FAILURE;
     }
-
-    ++pixelValue;
-    ++ritr;
+    // NOTE: pixelValue for unsigned short will wrap around with large images.
+    pixelValue = (pixelValue != itk::NumericTraits<PixelType>::max()) ? pixelValue + static_cast<PixelType>(1)
+                                                                      : static_cast<PixelType>(0);
   }
   chronometer.Stop("Compare");
 


Modules/IO/Meta/test/itkLargeMetaImageWriteReadTest.cxx
```diff
@@ -73,17 +71,13 @@ ActualTest(std::string filename, typename TImageType::SizeType size)
...
+    PixelType pixelValue = itk::NumericTraits<PixelType>::ZeroValue();
     chronometer.Start("Initializing");
-    while (!itr.IsAtEnd())
+    for (IteratorType itr(image, region); !itr.IsAtEnd(); ++itr)
     {
       itr.Set(pixelValue);
-      ++pixelValue;
-      ++itr;
+      // NOTE: pixelValue for unsigned short will wrap around with large images.
+      pixelValue = (pixelValue != itk::NumericTraits<PixelType>::max()) ? ++pixelValue : 0;
```
issakomi, line 80: It is likely that the GCC warning `operation on ... may be undefined [-Wsequence-point]` will not appear if the expression is

```cpp
pixelValue = (pixelValue != itk::NumericTraits<PixelType>::max()) ? pixelValue + 1 : 0;
```

or

```cpp
if (pixelValue != itk::NumericTraits<PixelType>::max())
{
  ++pixelValue;
}
else
{
  pixelValue = 0;
}
```

This warning happens with _-Wall_ argument, for some reason _GCC_ don't like expressions like:


```cpp
int main(int, char**)
{
	int x = 1;
	x = ++x;
	return 0;
}
```
```
r@deb2:~$ g++ -Wall test1.cpp 
test1.cpp: In function ‘int main(int, char**)’:
test1.cpp:4:11: warning: operation on ‘x’ may be undefined [-Wsequence-point]
    4 |         x = ++x;
      |         ~~^~~~~
```

blowekamp:
I may have written this code originally. I am not seeing where the warning/undefined behavior is coming from.

The PixelType/pixelValue is unsigned short, which has well-defined overflow behavior. However, the warning seems to indicate there is a conversion to int someplace? Where does this occur?

Modules/IO/Meta/test/itkLargeMetaImageWriteReadTest.cxx
```diff
@@ -73,17 +71,13 @@ ActualTest(std::string filename, typename TImageType::SizeType size)
 
     std::cout << "Initializing pixel values " << std::endl;
 
-    IteratorType itr(image, region);
-    itr.GoToBegin();
-
-    pixelValue = itk::NumericTraits<PixelType>::ZeroValue();
-
+    PixelType pixelValue = itk::NumericTraits<PixelType>::ZeroValue();
```
dzenanz, line 74: Should this be `PixelType pixelValue{};`?

dzenanz:
@blowekamp raises a good question.

issakomi:
I could not find any GCC `-fsanitize` option to detect this issue

```cpp
#include <iostream>
int main(int, char**)
{
	unsigned short x = 65535;
	++x;
	std::cout << x << std::endl;
	return 0;
}
```

but it works with Clang:

```
r@deb2:~$ clang++-14 -fsanitize=integer test2.cpp 
r@deb2:~$ ./a.out 
test2.cpp:5:2: runtime error: implicit conversion from type 'int' of value 65536 (32-bit, signed) to type 'unsigned short' changed the value to 0 (16-bit, unsigned)
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior test2.cpp:5:2 in 
0
```

Modules/IO/Meta/test/itkLargeMetaImageWriteReadTest.cxx
hjmjohnson: Sure. but somewhat unrelated.

hjmjohnson:
FYI:  This error surfaced on an M2 mac while attempting to find a way to verify that ITK was producing correct results.

It was using the stock apple clang compiler.

dzenanz:
People usually don't want wraparound to occur, hence the warning. Maybe there is a better way to tell the compiler we want wraparound? But as the is the code in a test, performance is not very important.

issakomi:
> Maybe there is a better way to tell the compiler we want wraparound?

Explicit cast?

```cpp
#include <iostream>
int main(int, char**)
{
	unsigned short x = 65535;
	x = static_cast<unsigned short>(x + 1);
	std::cout << x << std::endl;
	return 0;
}
```

```
r@deb2:~$ clang++-14 -fsanitize=integer test2.cpp 
r@deb2:~$ ./a.out 
0
```

Modules/IO/Meta/test/itkLargeMetaImageWriteReadTest.cxx
```diff
@@ -143,9 +133,8 @@ ActualTest(std::string filename, typename TImageType::SizeType size)
       std::cerr << "Read Image pixel value " << ritr.Get() << std::endl;
       return EXIT_FAILURE;
     }
-
-    ++pixelValue;
-    ++ritr;
+    // NOTE: pixelValue for unsigned short will wrap around with large images.
+    pixelValue = (pixelValue != itk::NumericTraits<PixelType>::max()) ? ++pixelValue : 0;
```
issakomi, line 137: The same GCC warning as at the line 80 before:

https://open.cdash.org/viewBuildError.php?type=1&buildid=8544205

issakomi:
> > ```c++
> > 	unsigned short x = 65535;
> > 	x = static_cast<unsigned short>(x + 1);
> > ```
> 
> The warning here make sense to me. The literals `65535` and `1` and of type `int` so warnings from conversions make since to me.... Did the original code have literal ints someplace?
> 
> It seems to me the code is getting more complicated and convoluted than it should be.


What warning? There are no warnings or sanitizer error with the code you included in the post.

dzenanz (approved):

Modules/IO/Meta/test/itkLargeMetaImageWriteReadTest.cxx
```diff
@@ -73,17 +71,14 @@ ActualTest(std::string filename, typename TImageType::SizeType size)
...
 
-    IteratorType itr(image, region);
-    itr.GoToBegin();
-
-    pixelValue = itk::NumericTraits<PixelType>::ZeroValue();
-
+    PixelType pixelValue{};
     chronometer.Start("Initializing");
-    while (!itr.IsAtEnd())
+    for (IteratorType itr(image, region); !itr.IsAtEnd(); ++itr)
```
N-Dekker, line 76: Please try not to mix style changes and bug fixes in one single commit. You know how much I like to see `while` loops being replaced with `for` loops, when using an iterator. 😃  But it seems off-topic here! It isn't the thing that causes undefined behavior, right?

PR #3969, opened 2023-03-17
ENH: Upgrade ITK from C++14 to C++17

N-Dekker:
Follow-up to pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2563 commit daec0fdd703b1bca16aeb07bc4e0a01101da950a "ENH: Upgrade ITK from C++11 to C++14"

Suggested by Dženan Zukić (@dzenanz) 😃 at https://github.com/InsightSoftwareConsortium/ITK/pull/3933#discussion_r1128770848 Still to be discussed.

More information about C++17:
- https://en.cppreference.com/w/cpp/17
- https://en.wikipedia.org/wiki/C%2B%2B17 
- https://isocpp.org/files/papers/p0636r0.html
- https://www.cppstories.com/2017/11/cpp17summary/
- etc...!

Discussion "Shall we start using C++17 at the master branch?" opened at https://discourse.itk.org/t/shall-we-start-using-c-17-at-the-master-branch/5828

PR #3897, opened 2023-01-31
COMP: Supress warning in MSVC for non-float complex

blowekamp:
## PR Checklist
- [ ] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [ ] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [ ] Added test (or behavior not changed)
- [ ] Updated API documentation (or API not changed)
- [ ] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [ ] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [ ] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples) for all new major features (if any)

Refer to the [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) for
further development details if necessary.

PR #3970, opened 2023-03-18
COMP: Place `#define _SILENCE_..._DEPRECATION_WARNING` before `#include`

N-Dekker:
Moved each `_SILENCE_NONFLOATING_COMPLEX_DEPRECATION_WARNING` macro definition upward, placing it before the very first `#include` statement.

Fixes Visual C++ 2022/C++17 warnings like:

> warning C4996: 'std::complex<char>::complex': warning STL4037: The effect of instantiating the template std::complex for any type other than float, double, or long double is unspecified. You can define _SILENCE_NONFLOATING_COMPLEX_DEPRECATION_WARNING to suppress this warning.

Follow-up to pull request https://github.com/InsightSoftwareConsortium/ITK/pull/3897 commit 3a9c76273e53f2da4b0e387066fceffd5b2d3e9d
"COMP: Supress warning in MSVC for non-float complex"

(Cherry-picked from pull request #3969.)


blowekamp (approved):
Thanks

N-Dekker:
> Thanks

You're welcome @blowekamp Just wondering, did your original PR #3897 commit 3a9c76273e53f2da4b0e387066fceffd5b2d3e9d already suppress the warnings for you? Because at my side, I kept seeing those warnings.

N-Dekker:
For the record, here a minimal example (https://godbolt.org/z/Ks51f1bjo): 

```
#include <complex>
// Doing the #define after the #include has no effect!
#define _SILENCE_NONFLOATING_COMPLEX_DEPRECATION_WARNING
std::complex<int> myComplex;
```

Using command-line options `/W3 /std:c++17 /Bv`, it has the following output:

> `<source>(8): warning C4996: 'std::complex<int>::complex': warning STL4037: The effect of instantiating the template std::complex for any type other than float, double, or long double is unspecified. You can define _SILENCE_NONFLOATING_COMPLEX_DEPRECATION_WARNING to suppress this warning.`
> Compiler Passes:
>   `C:\data\msvc\14.34.31931-Pre\bin\Hostx64\arm64\cl.exe:        Version 19.34.31931.0`

Place the `#define` before the `#include` to suppress the warning.  😃

dzenanz (approved):

N-Dekker:
I just realized that the warning is not specific to C++17, it also appears when compiling with `/std:c++14`. The warning appears to be added to MSVC very recently.

Modules/Core/Common/src/itkNumericTraits.cxx
```diff
@@ -15,6 +15,12 @@
  *  limitations under the License.
  *
  *=========================================================================*/
+
+#if !defined(ITK_LEGACY_REMOVE)
+// Supress MSVC warnings
```
N-Dekker, line 20: Shall I still do an amend/force-push to adjust the comment a little bit? It has a minor typo ("Supress"). No problem, of course.

dzenanz: Yes, please fix the misspelling.

PR #1904, opened 2020-07-02
BUG: Error converting from rotation axis and rotation angle to quaternion

huangjxbq:
BUG: 由旋转轴和旋转角度转化为四元数错误

PR #3984, opened 2023-03-30
BUG: Error converting from rotation axis and rotation angle to quater…

hjmjohnson:
A simple replacement for #1904 to fix the commit message.

### DIFF
--- a/Modules/Core/Transform/include/itkRigid3DPerspectiveTransform.hxx
+++ b/Modules/Core/Transform/include/itkRigid3DPerspectiveTransform.hxx
@@ -142,22 +142,8 @@ void
 Rigid3DPerspectiveTransform<TParametersValueType>::SetRotation(const Vector<TParametersValueType, 3> & axis,
                                                                double                                  angle)
 {
-  const double sinus = std::sin(angle / 2.0);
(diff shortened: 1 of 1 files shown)


dzenanz (approved):
Looks like a good simplification by removing redundancy. Could someone more versed in quaternion math than me also review this?

Modules/Core/Transform/include/itkRigid3DPerspectiveTransform.hxx
```diff
@@ -142,22 +142,8 @@ void
...
-  VnlQuaternionType q;
-
-  q[0] = cosinus;
-  q[1] = norm[0];
-  q[2] = norm[1];
-  q[3] = norm[2];
-
   VersorType v;
-  v.Set(q);
+  v.Set(axis, angle);
```
dzenanz, line 146: This code is invoked:
https://github.com/InsightSoftwareConsortium/ITK/blob/v5.3.0/Modules/Core/Common/include/itkVersor.hxx#L321-L345

hjmjohnson (commented):
@dzenanz Thanks for identifying the redundancy/ the patch seems to do exactly the same math, but with existing code rather than redundant code.

PR #3986, opened 2023-03-31
Declare lock_guard variables `const`, use C++17 class template argument deduction (CTAD)

N-Dekker:
Removed the template argument from `std::lock_guard<Mutex>` variable declarations, and declared all of them `const`.

Note that it does not appear necessary to replace C++11 `std::lock_guard` by C++17 `std::scoped_lock`, as `lock_guard` is still appreciated for its simplicity. For details, check Stack Overflow [std::lock_guard or std::scoped_lock?](https://stackoverflow.com/questions/43019598/stdlock-guard-or-stdscoped-lock) reply by Howard Hinnant at https://stackoverflow.com/a/60172828 (edited October 31, 2020), saying:
> For the common case that one needs to lock exactly one mutex, `std::lock_guard` has an API that is a little safer to use than `scoped_lock`.

### DIFF
--- a/Modules/Core/Common/include/itkLoggerThreadWrapper.hxx
+++ b/Modules/Core/Common/include/itkLoggerThreadWrapper.hxx
@@ -30,7 +30,7 @@ template <typename SimpleLoggerType>
 void
 LoggerThreadWrapper<SimpleLoggerType>::SetPriorityLevel(PriorityLevelEnum level)
 {
-  const std::lock_guard<std::mutex> lockGuard(m_Mutex);
+  const std::lock_guard lockGuard(m_Mutex);
   this->m_OperationQ.push(OperationEnum::SET_PRIORITY_LEVEL);
   this->m_LevelQ.push(level);
 }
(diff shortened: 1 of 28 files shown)


dzenanz (approved):

dzenanz:
Class template argument deduction is neat - we should use it more, now that we can 😄

hjmjohnson (approved):

dzenanz:
Uh-oh, this produced [20+ compile errors](https://open.cdash.org/viewBuildError.php?buildid=8602009) on clang.

N-Dekker:
Thanks @dzenanz This eventing I'm going to attend a C++ meetup here in Amsterdam. I'm interested to hear what they think about CTAD. Until now, it seemed such a cool feature!

dzenanz:
> Until now, it seemed such a cool feature!

I totally agree!

N-Dekker:
Before reverting this commit, I would still like to check if `scoped_lock` also triggers the warning that it "may not intend to support class template argument deduction. If not, we may switch from lock_guard to scoped_lock.

hjmjohnson:
> Thanks @dzenanz This eventing I'm going to attend a C++ meetup here in Amsterdam. I'm interested to hear what they think about CTAD. Until now, it seemed such a cool feature!

@N-Dekker My old brain thought it was cool, but did seem to make the code less clear (less redundant, but also less clear).

N-Dekker:
@dzenanz @hjmjohnson Please allow me a little bit of time to form an opinion on CTAD. I'll give you an answer in the morning 😺

N-Dekker:
> @dzenanz @hjmjohnson Please allow me a little bit of time to form an opinion on CTAD. I'll give you an answer in the morning 😺

Thanks for your patience 😃  Here is my proposal on how to deal with CTAD: https://github.com/InsightSoftwareConsortium/ITK/pull/3997#issuecomment-1498697783

PR #3992, opened 2023-04-03
STYLE: Replace `std::min` and `std::max` calls with C++17 `std::clamp`

N-Dekker:
The `clamp` algorithm was added to the Standard Library of C++17 as proposed by the paper `An algorithm to "clamp" a value between a pair of boundary values`, 2015-05-17, by Martin Moene (@martinmoene) and Niels Dekker (who happens to be the author of this commit 😸): https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4536.html

### DIFF
--- a/Modules/Core/Common/include/itkGaussianDerivativeOperator.h
+++ b/Modules/Core/Common/include/itkGaussianDerivativeOperator.h
@@ -174,7 +174,7 @@ public:
     constexpr double Min = 0.00001;
     const double     Max = 1.0 - Min;
 
-    m_MaximumError = std::max(Min, std::min(Max, maxerror));
+    m_MaximumError = std::clamp(maxerror, Min, Max);
   }
   double
   GetMaximumError()
--- a/Modules/Core/Common/include/itkTriangleHelper.hxx
+++ b/Modules/Core/Common/include/itkTriangleHelper.hxx
@@ -92,7 +92,7 @@ TriangleHelper<TPoint>::Cotangent(const PointType & iA, const PointType & iB, co
 
   CoordRepType bound(0.999999);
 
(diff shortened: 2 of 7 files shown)


N-Dekker:
`itk::Clamp<TInput, TOutput>` _might_ use `std::clamp` as well:


https://github.com/InsightSoftwareConsortium/ITK/blob/c642a68a7572ea683c11e488bdab9f19a8c1dd37/Modules/Filtering/ImageIntensity/include/itkClampImageFilter.h#L93-L108

Instead of the original code, I think it could just do:

    return static_cast<OutputType>(
      std::clamp(static_cast<double>(A), static_cast<double>(m_LowerBound), static_cast<double>(m_UpperBound)));

However, I don't really like to see so many type casts 🤔! So for the time being, I'd rather just leave `itk::Clamp<TInput, TOutput>` untouched.

dzenanz:
Do you know whether the internal cast to `double` there is a feature or a limitation? If limitation, we could just remove the cast altogether and simply replace by `return std::clamp(A, m_LowerBound, m_UpperBound);`

hjmjohnson (approved):

N-Dekker:
> Do you know whether the internal cast to `double` there is a feature or a limitation?

Thanks for asking @dzenanz. I think the intermediate conversion to double in `itk::Clamp<TInput, TOutput>` is there to avoid signed/unsigned comparison problems. The bounds (LowerBound, UpperBound) and the argument (`A`) may have different types. Whereas the three parameters of `std::clamp` all have the same type.

I think it may originally be intended as a feature, for example to allow clamping the pixels of a signed integer type between some unsigned integer values.

PR #3995, opened 2023-04-04
STYLE: Replace `sizeof(array) / sizeof(array[0])` with C++17 `std::size`

N-Dekker:
By Visual Studio, Replace in Files, using the following regular expression:

    Find what: sizeof\((\w+)\) / sizeof\(\1\[0\]\)
    Find what: sizeof\((.+)\) / sizeof\(\*\1\)
    Find what: sizeof\((\w+)\) / sizeof\(.*\)
    Replace with: std::size($1)

(The third regular expression, `sizeof\((\w+)\) / sizeof\(.*\)` was only applied to itkConvertBufferTest.cxx, as it cannot be generally applied).

`std::size(array)` appears to be a more readable and safer way to retrieve the number of elements of a C-style array. It is also `constexpr`.

### DIFF
--- a/Modules/Core/GPUCommon/src/itkOpenCLUtil.cxx
+++ b/Modules/Core/GPUCommon/src/itkOpenCLUtil.cxx
@@ -336,7 +336,7 @@ OpenCLCheckError(cl_int error, const char * filename, int lineno, const char * l
       "CL_INVALID_GLOBAL_WORK_SIZE",
     };
     // print error message
-    const int          errorCount = sizeof(errorString) / sizeof(errorString[0]);
+    constexpr int      errorCount = std::size(errorString);
     const int          index = -error;
     std::ostringstream errorMsg;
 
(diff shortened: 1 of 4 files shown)


dzenanz (approved):

N-Dekker:
FYI, While I did find these cases myself, there appears to be a ReSharper option to do such a find-and-replace:

https://blog.jetbrains.com/rscpp/2021/06/25/resharper-cpp-2021-2-eap-inline-function-refactoring/#modernizing_inspections saying:

> [...] new inspections that help you modernize your code:
>
>    `std::size` can be used. The C idiom for array size calculation is error-prone and works incorrectly when the `sizeof` operator is applied to a pointer instead of a built-in array. This happens, for example, with decaying function parameters (`void f(int arr[3])`) or dynamically allocated arrays (`int arr[]`). The new inspection suggests using std::size as a type safe alternative.

dzenanz:
If anyone has this tool, it would be good if they ran it on the codebase.

N-Dekker:
> If anyone has this tool, it would be good if they ran it on the codebase.

I have access to ReSharper C++, so I can give it a try, 😃

PR #3998, opened 2023-04-06
COMP: Add CTAD deduction guides to ImageBufferRange and ImageRegionRange

N-Dekker:
Addressed `-Wctad-maybe-unsupported` warnings from Mac10.15-AppleClang-rel (AppleClang 12.0.0.12000032), saying:

> warning: 'ImageBufferRange' may not intend to support class template argument deduction [-Wctad-maybe-unsupported]
> note: add a deduction guide to suppress this warning

### DIFF
--- a/Modules/Core/Common/include/itkImageRegionRange.h
+++ b/Modules/Core/Common/include/itkImageRegionRange.h
@@ -461,6 +461,11 @@ public:
   ~ImageRegionRange() = default;
 };
 
+
+// Deduction guide to avoid compiler warnings (-wctad-maybe-unsupported) when using class template argument deduction.
+template <typename TImage>
+ImageRegionRange(TImage &)->ImageRegionRange<TImage>;
+
 } // namespace itk
 
 #endif
--- a/Modules/Core/Common/include/itkImageBufferRange.h
+++ b/Modules/Core/Common/include/itkImageBufferRange.h
@@ -664,6 +664,12 @@ public:
   }
 };
 
+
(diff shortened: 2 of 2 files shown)


dzenanz (approved):

Modules/Core/Common/include/itkImageRegionRange.h
```diff
@@ -461,6 +461,11 @@ class ImageRegionRange final
   ~ImageRegionRange() = default;
 };
 
+
+// Deduction guide to avoid compiler warnings (-wctad-maybe-unsupported) when using class template argument deduction.
+template <typename TImage>
+ImageRegionRange(TImage &)->ImageRegionRange<TImage>;
+
```
N-Dekker, line 468: For the record, https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Dialect-Options.html#C_002b_002b-Dialect-Options appears to suggest working around those `-wctad-maybe-unsupported` warnings by adding a "dummy" deduction guide:

> -Wctad-maybe-unsupported (C++ and Objective-C++ only)

>    Warn when performing class template argument deduction (CTAD) on a type with no explicitly written deduction guides. This warning will point out cases where CTAD succeeded only because the compiler synthesized the implicit deduction guides, which might not be what the programmer intended. Certain style guides allow CTAD only on types that specifically "opt-in"; i.e., on types that are designed to support CTAD. This warning can be suppressed with the following pattern:

```
    struct allow_ctad_t; // any name works
    template <typename T> struct S {
      S(T) { }
    };
    // Guide with incomplete parameter type will never be considered.
    S(allow_ctad_t) -> S<void>;
```

PR #4000, opened 2023-04-06
Fix TransformPhysicalPointToIndex, TransformPhysicalPointToContinuousIndex  `[[nodiscard]]` warnings

N-Dekker:
Replacing `TransformPhysicalPointToIndex(point, index)` with `index = TransformPhysicalPointToIndex(point)` appeared to be a no-brainer, while replacing the `TransformPhysicalPointToContinuousIndex` calls was somewhat more challenging... but still no rocket science.

### DIFF
--- a/Modules/Compatibility/Deprecated/include/itkVectorResampleImageFilter.hxx
+++ b/Modules/Compatibility/Deprecated/include/itkVectorResampleImageFilter.hxx
@@ -114,7 +114,6 @@ VectorResampleImageFilter<TInputImage, TOutputImage, TInterpolatorPrecisionType>
   PointType inputPoint;  // Coordinates of current input pixel
 
   using ContinuousIndexType = ContinuousIndex<TInterpolatorPrecisionType, ImageDimension>;
-  ContinuousIndexType inputIndex;
 
   // Doc says this only works for VectorImage, but Image implementation says otherwise...
   const unsigned int numberOfComponents = this->GetInput()->GetNumberOfComponentsPerPixel();
@@ -131,7 +130,8 @@ VectorResampleImageFilter<TInputImage, TOutputImage, TInterpolatorPrecisionType>
 
     // Compute corresponding input pixel position
     inputPoint = m_Transform->TransformPoint(outputPoint);
-    inputPtr->TransformPhysicalPointToContinuousIndex(inputPoint, inputIndex);
+    const ContinuousIndexType inputIndex =
+      inputPtr->template TransformPhysicalPointToContinuousIndex<TInterpolatorPrecisionType>(inputPoint);
 
     // Evaluate input at right position and copy to the output
     if (m_Interpolator->IsInsideBuffer(inputIndex))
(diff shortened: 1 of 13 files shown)


Modules/Compatibility/Deprecated/include/itkVectorResampleImageFilter.hxx
```diff
@@ -114,7 +114,6 @@ VectorResampleImageFilter<TInputImage, TOutputImage, TInterpolatorPrecisionType>
   PointType inputPoint;  // Coordinates of current input pixel
```
dzenanz, line 114: In the commit messages: You might want to mention the performance benefit of avoiding the overload which does the bounds check, instead of motivating these changes by warning avoidance.

dzenanz (commented):
Mostly looks good.

Modules/Compatibility/Deprecated/include/itkVectorResampleImageFilter.hxx
dzenanz: But the warnings were added for performance reasons, mere days ago 😄

N-Dekker: OK, you're right, if we would just want to get rid of the warnings, cast to `(void)` might work as well 😈 

    // Cast away the info on whether the point is inside/outside.
    (void)TransformPhysicalPointToIndex(point, index);

Of course, modern C++17 programmers would use [std::ignore](https://en.cppreference.com/w/cpp/utility/tuple/ignore) instead:

    // Ignore whether the point is inside/outside.
    std::ignore( TransformPhysicalPointToIndex(point, index) );

Lovely 😄

dzenanz (approved):

Modules/Nonunit/Review/include/itkScalarChanAndVeseSparseLevelSetImageFilter.hxx
```diff
@@ -38,8 +38,7 @@ ScalarChanAndVeseSparseLevelSetImageFilter<TInputImage, TFeatureImage, TOutputIm
     InputPointType    origin = input->GetOrigin();
 
     // In the context of the global coordinates
-    FeatureIndexType start;
-    this->GetInput()->TransformPhysicalPointToIndex(origin, start);
+    FeatureIndexType start = this->GetInput()->TransformPhysicalPointToIndex(origin);
```
thewtex, line 41: It seems many of these could be marked `const` now?

thewtex (approved):
Thanks for addressing the warnings :green_heart: 

Possible follow-up noted inline.

Modules/Nonunit/Review/include/itkScalarChanAndVeseSparseLevelSetImageFilter.hxx
N-Dekker: Indeed, thanks Matt! Maybe for another PR? (I would like to have this one merged as soon as it's "good enough", because it fixes a few warnings at https://open.cdash.org/index.php?project=Insight)

PR #4002, opened 2023-04-06
(Not ready) Simpler `TransformPhysicalPointToContinuousIndex(point)` overloads

N-Dekker:
Added non-template overloads, which allow the user to write:

    index = image->TransformPhysicalPointToContinuousIndex(point);

Instead of calling one of the overload that was added with ITK 5.0, for example:

    index = image->template TransformPhysicalPointToContinuousIndex<typename ContinuousIndexType::ValueType>(point);

I assume that usually (or at least in _most_ common use cases), the user just wants TransformPhysicalPointToContinuousIndex to just yield a `ContinuousIndex<SpacePrecisionType, N>` (with `SpacePrecisionType` = `double`, by default). So in most use cases, the _ValueType_ of ContinuousIndex does not need to be a template argument of this function. Is that OK?

### DIFF
--- a/Modules/Core/Common/include/itkImageBase.h
+++ b/Modules/Core/Common/include/itkImageBase.h
@@ -509,6 +509,19 @@ public:
     return index;
   }
 
+
+  /** \brief Returns the continuous index from a physical point
+   * \note This specific overload is easier to use, because it does not have template arguments. It just uses
+   * `itk::SpacePrecisionType`, both for the coordinates of the point and the index values.
+   *
+   * \sa Transform */
+  [[nodiscard]] ContinuousIndex<SpacePrecisionType, VImageDimension>
+  TransformPhysicalPointToContinuousIndex(const PointType & point) const
+  {
+    return TransformPhysicalPointToContinuousIndex<SpacePrecisionType>(point);
+  }
+
+
   /** \brief Get the continuous index from a physical point
    *
    * Returns true if the resulting index is within the image, false otherwise.
--- a/Modules/Core/Common/include/itkImageAlgorithm.hxx
+++ b/Modules/Core/Common/include/itkImageAlgorithm.hxx
@@ -260,8 +260,7 @@ ImageAlgorithm::EnlargeRegionOverBox(const typename InputImageType::RegionType &
         outputPoint[d] = inputPoint[d];
       }
     }
-    outputCorners[count] =
-      outputImage->template TransformPhysicalPointToContinuousIndex<ContinuousIndexValueType>(outputPoint);
+    outputCorners[count] = outputImage->TransformPhysicalPointToContinuousIndex(outputPoint);
   }
 
   // Compute a rectangular region from the vector of corner indexes
--- a/Modules/Core/Common/include/itkPhasedArray3DSpecialCoordinatesImage.h
+++ b/Modules/Core/Common/include/itkPhasedArray3DSpecialCoordinatesImage.h
@@ -208,6 +208,18 @@ public:
     return index;
   }
 
+  /** \brief Returns the continuous index from a physical point
+   * \note This specific overload is easier to use, because it does not have template arguments. It just uses
+   * `itk::SpacePrecisionType`, both for the coordinates of the point and the index values.
+   *
+   * \sa Transform */
+  [[nodiscard]] ContinuousIndex<SpacePrecisionType, ImageDimension>
+  TransformPhysicalPointToContinuousIndex(const PointType & point) const
+  {
+    return TransformPhysicalPointToContinuousIndex<SpacePrecisionType>(point);
+  }
+
+
   /** \brief Get the continuous index from a physical point
    *
    * Returns true if the resulting index is within the image, false otherwise.
--- a/Modules/Core/ImageAdaptors/include/itkImageAdaptor.h
+++ b/Modules/Core/ImageAdaptors/include/itkImageAdaptor.h
@@ -389,6 +389,18 @@ public:
     return m_Image->template TransformPhysicalPointToContinuousIndex<TIndexRep>(point);
   }
 
+  /** \brief Returns the continuous index from a physical point
+   * \note This specific overload is easier to use, because it does not have template arguments. It just uses
+   * `itk::SpacePrecisionType`, both for the coordinates of the point and the index values.
+   *
+   * \sa Transform */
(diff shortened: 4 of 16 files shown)


dzenanz (approved):
I love it!

Modules/Core/Common/include/itkImageBase.h
```diff
@@ -509,6 +509,19 @@ class ITK_TEMPLATE_EXPORT ImageBase : public DataObject
...
+  /** \brief Returns the continuous index from a physical point
+   * \note This specific overload is easier to use, because it does not have template arguments. It just uses
+   * `itk::SpacePrecisionType`, both for the coordinates of the point and the index values.
+   *
+   * \sa Transform */
+  [[nodiscard]] ContinuousIndex<SpacePrecisionType, VImageDimension>
+  TransformPhysicalPointToContinuousIndex(const PointType & point) const
+  {
+    return TransformPhysicalPointToContinuousIndex<SpacePrecisionType>(point);
+  }
```
N-Dekker, line 522: Would it be useful to have two of such (simple, non-templated) overloads? One of `float` and one of `double`? Specifically:

    ContinuousIndex<float, VImageDimension>
    TransformPhysicalPointToContinuousIndex(const Point<float, VImageDimension> &) const;

    ContinuousIndex<double, VImageDimension>
    TransformPhysicalPointToContinuousIndex(const Point<double, VImageDimension> &) const;

So one from `Point<float, N>` to `ContinuousIndex<float, N>`, and one from `Point<double, N>` to `ContinuousIndex<double, N>`?  Overload resolution would then automatically choose the right one, no need for the user to specify template arguments. Right?

dzenanz: The float one is rarely used, I believe. If someone wants an unusual thing, they can spell out the template parameters.

N-Dekker: I'm wondering, could all those `image->template TransformPhysicalPointToContinuousIndex<ContinuousValueType>(point)` calls inside ITK itself then replaced by simple non-templated `image->TransformPhysicalPointToContinuousIndex(point)` calls?

dzenanz (approved):

dzenanz:
Is this ready for merging? It would fix a lot of compile errors on the dashboard.

N-Dekker:
> Is this ready for merging?

@dzenanz  No it's still a draft, sorry! PR #4005 is ready, though!

N-Dekker:
Is `ITK_USE_FLOAT_SPACE_PRECISION` still supported? As in:

https://github.com/InsightSoftwareConsortium/ITK/blob/1f8c0f8128b727d6d65ba2583661d489e7920f3b/Modules/Core/Common/include/itkFloatTypes.h#L26-L31

Maybe it's just me, but I got a lot of compile errors when I try to compile the current master with `ITK_USE_FLOAT_SPACE_PRECISION = ON`. Is it still tested elsewhere?

I'm asking here because `ITK_USE_FLOAT_SPACE_PRECISION = ON` would make it harder to use the simpler TransformPhysicalPointToContinuousIndex overload (proposed by this PR) inside ITK itself.

dzenanz:
`ITK_USE_FLOAT_SPACE_PRECISION` is still exposed in CMake, but I don't think it has been tested for a very long time. I think some components assume it is `double`. Maybe it is time to remove that setting and deprecate that header? What do @blowekamp and @hjmjohnson think about this?

N-Dekker:
If we can't have a simple, _easy to use_ alternative to the old "`IsInside` estimating" `TransformPhysicalPointToContinuousIndex(point, index)`, maybe just remove the `[[nodiscard]]` from that particular member function. I'm sorry, I had not expected so much trouble from `[[nodiscard]]`. I still think users should be aware of the fact that `TransformPhysicalPointToContinuousIndex(point, index)` is estimating `IsInside`. And I still think they should only use it when they really want to know whether the point is inside. But I just don't know how to make it easier. 🤷

N-Dekker:
> `ITK_USE_FLOAT_SPACE_PRECISION` is still exposed in CMake, but I don't think it has been tested for a very long time. I think some components assume it is `double`. Maybe it is time to remove that setting and deprecate that header? What do @blowekamp and @hjmjohnson think about this?

I think that as long as ITK supports `ITK_USE_FLOAT_SPACE_PRECISION` (allowing the user to choose between `SpacePrecisionType = float` and `SpacePrecisionType = double`) and as long as ContinuousIndex supports float, double, _and_ SpacePrecisionType as value type, a simple `TransformPhysicalPointToContinuousIndex(point)` overload that just returns `ContinuousIndex<SpacePrecisionType, N>` (proposed by this PR) is of very limited use.

hjmjohnson:
> I think that as long as ITK supports `ITK_USE_FLOAT_SPACE_PRECISION` (allowing the user to choose between `SpacePrecisionType = float`

I think we should test the ITK_USE_FLOAT_SPACE_PRECISION,  I could think of a customer who may use FLOAT_SPACE_PRECISION for the purpose of improving computational speed for their algorithm.  Removing this option **MAY** cause that customer great pain.

The above scenario is purely hypothetical .... just pointing out that removal may have unintended consequences for customers.

hjmjohnson:
FYI:  I just compiled successfully with ITK_USE_FLOAT_SPACE_PRECISION.  Test failures are as expected due to the limited numerical precision causing slight differences from the baselines.

> > I think that as long as ITK supports `ITK_USE_FLOAT_SPACE_PRECISION` (allowing the user to choose between `SpacePrecisionType = float`
> 
> I think we should test the ITK_USE_FLOAT_SPACE_PRECISION, I could think of a customer who may use FLOAT_SPACE_PRECISION for the purpose of improving computational speed for their algorithm. Removing this option **MAY** cause that customer great pain.
> 
> The above scenario is purely hypothetical .... just pointing out that removal may have unintended consequences for customers.

hjmjohnson:
Hi @N-Dekker — superseded by #6186, which is a clean rebase of your two commits on top of current `main` (the original branch was 4838 commits behind, and a direct cherry-pick reverted two years of style modernization on the same files).  Authorship is preserved via `GIT_AUTHOR_*`; @dzenanz's earlier guidance on the float-overload question is followed.  I'll close this PR once #6186 lands.

hjmjohnson:
Superceeded on upstream branch.

PR #4007, opened 2023-04-11
ENH: Add ITK_NO_DISCARD(message) macro to Image "Transform" members

dzenanz:
The new macro is applied only to the overloads which take two arguments and return the bool. This is the only scenario where it made sense to ignore the return value.

## PR Checklist


thewtex (approved):

Modules/Core/Common/include/itkMacro.h
```diff
@@ -1386,6 +1386,16 @@ itkDynamicCastInDebugMode(TSource x)
...
 
+#ifdef ITK_LEGACY_REMOVE
+#  if __cplusplus >= 202002L
+#    define ITK_NO_DISCARD(message) [[nodiscard(#    message)]]
+#  else
+#    define ITK_NO_DISCARD(message) [[nodiscard]]
+#  endif
+#else
+#  define ITK_NO_DISCARD(message)
+#endif
```
N-Dekker, line 1397: Thanks @dzenanz Some suggestions:
- For consistency with C++20 `[[nodiscard("Some message")]]` uses, I would suggest to remove the `#` and simply always add double-quotes to the actual message when _calling_ the macro.
- For consistency with the C++ attribute name, maybe just do `NODISCARD` without underscore.

Modules/Core/Common/include/itkImageBase.h
```diff
@@ -465,8 +465,8 @@ class ITK_TEMPLATE_EXPORT ImageBase : public DataObject
    *
    * \sa Transform */
   template <typename TCoordRep>
-  [[nodiscard]] bool
-  TransformPhysicalPointToIndex(const Point<TCoordRep, VImageDimension> & point, IndexType & index) const
+  ITK_NO_DISCARD(Call the overload which has the point as the only parameter and returns the index)
+  bool TransformPhysicalPointToIndex(const Point<TCoordRep, VImageDimension> & point, IndexType & index) const
```
N-Dekker, line 469: Please consider to leave the unconditional `[[nodiscard]]` (even for legacy-only) here. For `TransformPhysicalPointToIndex`, it's really a _no-brainer_ to replace it with the other overload. (It does not have the `template` difficulty that TransformPhysicalPointToContinuousIndex has.)

dzenanz: Even if replacement is easy, it requires code changes. The legacy flag's purpose is to allow this to be made gradually.

Modules/Core/Common/include/itkMacro.h
dzenanz: Applied the first two suggestions. I leave the third to you 😄

issakomi (approved):

Modules/Core/Common/include/itkMacro.h
N-Dekker: > I leave the third to you

Maybe just wait for ITK to upgrade to C++20 😸

Issue #3994, opened 2023-04-04
Nifti files with non-orthonormal directional cosines should be read and the error changed to a warning.


tomvars:
Hi all,

In this PR - https://github.com/InsightSoftwareConsortium/ITK/pull/1868/files ITK introduced an additional constraint on NiFTI files being read.

This has affected users across a variety of technologies which use ITK such as TorchIO, TotalSegmentator, Nvidia AAIA, HD-BET, nnDetection and even ITK-SNAP (with some files working on < v3.6 but not on v3.8 >)

Given so many users see this as unexpected behaviour could the developers investigate whether the constraint is still a reasonable one?
Note that in this issue this was discussed but the solution solved only one failure case related to small voxel sizes - https://github.com/InsightSoftwareConsortium/ITK/issues/2674 - this issue is still affecting users with larger voxel sizes across a variety of technologies

https://github.com/MIC-DKFZ/nnDetection/issues/24
https://github.com/SimpleITK/SimpleITK/issues/1433
https://github.com/fepegar/torchio/issues/669
https://github.com/wasserth/TotalSegmentator/issues/32

dzenanz:
People who use NIFTI a lot should weigh in. Candidates: @hjmjohnson @cookpa @vfonov @seanm. Maybe make the error/warning CMake configurable, with the default being a warning?

hjmjohnson:
ITK does not support non-orthogonal images,  to read the images that are non-orthogonal invalidates all the requirements about processing and analyzing images.

vfonov:
Maybe we should add a flag that forces Nifti reader to convert the non-confirming non-orthogonal matrix to the closest possible orthogonal?

dzenanz:
@tomvars please take a look at #4009.

cookpa:
I looked through the linked issues in the first post, but I didn't see any links to example data that reproduce the problem. Even the headers by themselves would be useful, these can be extracted with

head -c 352 image.nii > header.nii   

If the sform checks are too strict, it would be good to fix them. Otherwise everyone is just going to turn on this flag at compile time and the end users will likely not even know about it

vfonov:
[...]

PR #4009, opened 2023-04-12
ENH: Added permissive mode to NIFTI reader

vfonov:
Added permissive mode to NIFTI reader to allow reading files with non-orthogonal s-form.
Addresses #3994 

## PR Checklist


dzenanz (approved):

dzenanz:
It looks like you need to apply `clang-format`, see [complaints](https://github.com/InsightSoftwareConsortium/ITK/actions/runs/4679321376/jobs/8289184379?pr=4009).

CMakeLists.txt
```diff
@@ -338,6 +338,11 @@ if(NOT ("${ITK_NIFTI_IO_ANALYZE_FLAVOR}" STREQUAL "ITK4Warning" OR
 endif()
 mark_as_advanced(ITK_NIFTI_IO_ANALYZE_FLAVOR)
 
+# Allow non-orthogonal rotation matrix approximation in NIFTI files
+option(ITK_NIFTI_IO_SFORM_PERMISSIVE "Allow non-orthogonal rotation matrix in NIFTI sform" ON)
+mark_as_advanced(ITK_NIFTI_IO_SFORM_PERMISSIVE)
```
blowekamp, line 343: Should this be a compile time flag or a runtime environment variable?

issakomi: It looks like that he option changes the default behavior, not a final behavior:
`, m_SFORM_Permissive{ ITK_NIFTI_IO_SFORM_PERMISSIVE }`
m_SFORM_Permissive can be changed later `itkSetMacro(SFORM_Permissive, bool);
`

Also I can hardly believe that `#cmakedefine ITK_NIFTI_IO_SFORM_PERMISSIVE` **will work at all**.
```
#cmakedefine VAR ...
will be replaced with either
#define VAR ...
or
/* #undef VAR */
```

So even if the option is set to "ON", it will **not** set _m_SFORM_Permissive_ to _true_ here
`, m_SFORM_Permissive{ ITK_NIFTI_IO_SFORM_PERMISSIVE }`,
it will expand to `, m_SFORM_Permissive{}` and set it to _false_.

If it is not defined, the build will fail.

There is `#cmakedefine01`  s. [configure_file](https://cmake.org/cmake/help/latest/command/configure_file.html), this may work. But, IMHO, it would be better to review the whole thing. The option seems to only set the default, but it is not clear from the description.

vfonov: ok, I changed `#cmakedefine` to `#cmakedefine01` and updated option description.

CMakeLists.txt
```diff
@@ -338,6 +338,11 @@ if(NOT ("${ITK_NIFTI_IO_ANALYZE_FLAVOR}" STREQUAL "ITK4Warning" OR
 endif()
 mark_as_advanced(ITK_NIFTI_IO_ANALYZE_FLAVOR)
 
+# Allow non-orthogonal rotation matrix approximation in NIFTI files
+option(ITK_NIFTI_IO_SFORM_PERMISSIVE "Allow non-orthogonal rotation matrix in NIFTI sform by default" ON)
```
hjmjohnson, line 342: I am very opposed to this being the default compile time option.

Was OK with a runtime selection by the developer to explicitly request to ignore the fact that this behavior is not conformant with the rest of the ITK standard.

I think it is a very dangerous proposition to allow very hard to understand data assumption violations as the default.

hjmjohnson: Why is a compile time option needed?  The runtime behavior should be backward compatible with the historical behavior, and developers who wish to change the behavior have the option to set the desired new behavior at runtime.

PS: Thanks for putting this patch forward for discussion.

hjmjohnson (dismissed):

CMakeLists.txt
vfonov: So, I thought this way we will restore compatibility with the version before the strict checking was introduced.

blowekamp: I am also apposed to a compile time option.

Per has a "GlobalDefault" runtime variable would be more appropriate? With the default as Hans has described.

vfonov: so, what's the current recommendation?

dzenanz: ITK_NIFTI_IO_SFORM_PERMISSIVE defaulting to OFF? And whoever doesn't like that default, can change it at both configure time and run time.

dzenanz: Additional layer of complexity would be environment variable, similar to this logic: https://github.com/InsightSoftwareConsortium/ITK/blob/v5.3.0/Modules/Core/Common/src/itkMultiThreaderBase.cxx#L287-L360

dzenanz (approved):

Modules/IO/NIFTI/test/itkNiftiReadWriteDirectionTest.cxx
```diff
@@ -112,6 +160,30 @@ itkNiftiReadWriteDirectionTest(int argc, char * argv[])
...
+  // check if environment flag is used properly
+  // check without permissive option
+  itksys::SystemTools::PutEnv("ITK_NIFTI_SFORM_PERMISSIVE=NO");
+  ITK_TRY_EXPECT_EXCEPTION(itk::ReadImage<TestImageType>(argv[4]));
+  // check with permissive option
+  itksys::SystemTools::PutEnv("ITK_NIFTI_SFORM_PERMISSIVE=YES");
+  ITK_TRY_EXPECT_NO_EXCEPTION(itk::ReadImage<TestImageType>(argv[4]));
+  // check with permissive option, when nothing is specified
+  itksys::SystemTools::PutEnv("ITK_NIFTI_SFORM_PERMISSIVE");
+  ITK_TRY_EXPECT_NO_EXCEPTION(itk::ReadImage<TestImageType>(argv[4]));
```
issakomi, line 175: `itksys::SystemTools::PutEnv("ITK_NIFTI_SFORM_PERMISSIVE");` doesn't work

```
~$ export ITK_NIFTI_SFORM_PERMISSIVE
~$ env | grep ITK_NIFTI_SFORM_PERMISSIVE
~$
```
Probably at least empty value is required for environment variable.

BTW, FYI, another test _itkNiftiReadWriteDirectionSmallVoxelTest_ fails because number of arguments is wrong now.

dzenanz: `Export` and `PutEnv` might work differently?

issakomi: > `Export` and `PutEnv` might work differently?

Yes, _export_ doesn't override last value if there is no "="

```
r@deb2:~$ export SOME_VAR=1
r@deb2:~$ export SOME_VAR
r@deb2:~$ env | grep SOME_VAR
SOME_VAR=1
r@deb2:~$ 
```

PutEnv seems to unset the variable if there is no "=". Anyway, test works without there 2 lines.
```
bool SystemTools::PutEnv(const std::string& env)
{
  size_t pos = env.find('=');
  if (pos != std::string::npos) {
    std::string name = env.substr(0, pos);
    return setenv(name.c_str(), env.c_str() + pos + 1, 1) == 0;
  } else {
    return kwsysUnPutEnv(env) == 0;
  }
}
```

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -2089,15 +2103,48 @@ NiftiImageIO::SetImageIOOrientationFromNIfTI(unsigned short dims)
...
         }
-      } // sform is orthonormal
-    }   // sform not NIFTI_XFORM_UNKNOWN
+      }
+      else if (m_SFORM_Permissive) // sform is orthonormal
+      {
+        // Fix to deal with non-orthogonal matrixes
+        // this approach uses SVD decomposition
+        // maybe it is better to use nifti_make_orthog_mat44
+        // to be consistent with other software?
```
issakomi, line 2112: Unfortunately the block below doesn't produce orthogonal matrix. The test image _NonOrthoSform.nii.gz_ ends up with
```
  Direction: 
0.694449 0.264737 0.469327
-0.625039 -0.670148 -0.125412
0.356465 -0.693409 0.874073
```
The matrix is not orthogonal. I have tested the alternative you mentioned above and it works very good, `nifti_make_orthog_mat44` generates orthogonal matrices.

```cpp
      else if (m_SFORM_Permissive)
      {
        itkWarningMacro(<< this->GetFileName() << " has non-orthogonal sform, corrected");
        this->m_SFORM_Corrected = true;
        return nifti_make_orthog_mat44(this->m_NiftiImage->sto_xyz.m[0][0],
                                       this->m_NiftiImage->sto_xyz.m[0][1],
                                       this->m_NiftiImage->sto_xyz.m[0][2],
                                       this->m_NiftiImage->sto_xyz.m[1][0],
                                       this->m_NiftiImage->sto_xyz.m[1][1],
                                       this->m_NiftiImage->sto_xyz.m[1][2],
                                       this->m_NiftiImage->sto_xyz.m[2][0],
                                       this->m_NiftiImage->sto_xyz.m[2][1],
                                       this->m_NiftiImage->sto_xyz.m[2][2]);
      }
```

issakomi: Updated the above post (to trigger notification).

hjmjohnson (changes requested):
It appears that the logic for making an orthogonal matrix is not correct.  See https://github.com/InsightSoftwareConsortium/ITK/pull/4009#discussion_r1167992153

hjmjohnson:
@issakomi Would you please make a final review of the latest changes to ensure that they address your previously identified bug?

Thanks,
Hans

issakomi:
> @issakomi Would you please make a final review of the latest changes to ensure that they address your previously identified bug?

The bug is fixed. Thanks.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -2089,15 +2102,59 @@ NiftiImageIO::SetImageIOOrientationFromNIfTI(unsigned short dims)
           }();
           prefer_sform_over_qform = sform_and_qform_are_very_similar;
         }
-      } // sform is orthonormal
-    }   // sform not NIFTI_XFORM_UNKNOWN
+      }
+      else if (m_SFORM_Permissive) // sform is orthonormal
```
issakomi, line 2106: Very minor comment. IMHO, the comment `// sform is orthonormal` marks the end of the previous block. At this line it might be a little bit misleading, IMHO.

issakomi (approved):

dzenanz (approved):

PR #4010, opened 2023-04-12
PERF: Use faster TransformPhysicalPointToIndex/ContinuousIndex in Python

N-Dekker:
Replaced calls to TransformPhysicalPointToIndex(point, idx) and TransformPhysicalPointToContinuousIndex(point, idx) by the corresponding faster overloads.

Aims to address warnings at Windows_NT-Build3495-master-Python https://open.cdash.org/viewBuildError.php?type=1&buildid=8615447 saying:

> s-build\Wrapping\Modules\ITKCommon\itkImagePython.cpp: warning C4834:
> discarding return value of function with 'nodiscard' attribute

----

Please note: I don't really know how this Python wrapping code works, so I'm basically just _guessing_ how it should be fixed! So please review with extra care 😃!


dzenanz (commented):
LGTM, but I think that Matt should review this 😄

thewtex (approved):
LGTM thanks!

PR #3997, opened 2023-04-05
COMP: Remove CTAD because of clang "error: member reference base" bug

N-Dekker:
Removed a few cases of class template argument deduction, because of compile errors from [RogueResearch17/Mac10.15-AppleClang-rel](https://open.cdash.org/viewBuildError.php?buildid=8602009), saying:

> error: member reference base type 'ImageRegionRange' is not a structure or union

Caused by clang Bug 39663 "Class template argument deduction with deduced type fails when accessing member", reported by Baruch, 2018-11-14 at https://bugs.llvm.org/show_bug.cgi?id=39663 and fixed on 2020-10-08. The fix appears to be included with LLVM Clang 12.0.0.

This partially reverts pull request https://github.com/InsightSoftwareConsortium/ITK/pull/3989 commit 8ee4d9a410556cd1619c188e8e588710805b2da8 "STYLE: Use C++17 CTAD for `ImageRegionRange`"

PR #4012, opened 2023-04-14
COMP: Enable warning on C++17 class template argument deduction (CTAD)

N-Dekker:
Enabled a compiler warning from both GCC and Clang on possibly unintended use of CTAD.

Triggered by our discussions at
 - https://github.com/InsightSoftwareConsortium/ITK/pull/3997
 - https://github.com/InsightSoftwareConsortium/ITK/pull/3986

### DIFF
--- a/CMake/ITKSetStandardCompilerFlags.cmake
+++ b/CMake/ITKSetStandardCompilerFlags.cmake
@@ -122,6 +122,7 @@ function(check_compiler_warning_flags c_warning_flags_var cxx_warning_flags_var)
     -Wno-invalid-offsetof
     -Wno-undefined-var-template  # suppress invalid warning when explicitly instantiated in another translation unit
     -Woverloaded-virtual
+    -Wctad-maybe-unsupported
     -Wstrict-null-sentinel
   )
 ##-Wno-c++0x-static-nonintegral-init


dzenanz (approved):

N-Dekker:
Please check at PR #4013 if those warnings are really enabled at the CI, with this PR.

N-Dekker:
For the record, the GCC documentation on `-Wctad-maybe-unsupported`:

> Warn when performing class template argument deduction (CTAD) on a type with no explicitly written deduction guides. This warning will point out cases where CTAD succeeded only because the compiler synthesized the implicit deduction guides, which might not be what the programmer intended. Certain style guides allow CTAD only on types that specifically "opt-in"; i.e., on types that are designed to support CTAD. This warning can be suppressed with the following pattern:

```
    struct allow_ctad_t; // any name works
    template <typename T> struct S {
      S(T) { }
    };
    S(allow_ctad_t) -> S<void>; // guide with incomplete parameter type will never be considered

```
https://gcc.gnu.org/onlinedocs/gcc-12.2.0/gcc/C_002b_002b-Dialect-Options.html#index-Wctad-maybe-unsupported

thewtex (approved):
:+1:

N-Dekker:
This is our "safety belt" before we can start using CTAD  😃

PR #4013, opened 2023-04-14
WIP:  Class template argument deduction (CTAD) for ImageScanlineIterator, ImageScanlineConstIterator

N-Dekker:
Do not merge! Intended to check if the expected `-Wctad-maybe-unsupported` warnings will appear!


Modules/Core/Common/include/itkImageScanlineConstIterator.h
```diff
@@ -260,6 +260,12 @@ class ITK_TEMPLATE_EXPORT ImageScanlineConstIterator : public ImageConstIterator
   OffsetValueType m_SpanBeginOffset{}; // one pixel the beginning of the scanline
   OffsetValueType m_SpanEndOffset{};   // one pixel past the end of the scanline
 };
+
+// Deduction guide for class template argument deduction (CTAD).
+template <typename TImage>
+ImageScanlineConstIterator(SmartPointer<const TImage>, const typename TImage::RegionType &)
```
dzenanz, line 266: This type of deduction guide is needed for most classes and templated methods in ITK, in order to avoid this less comprehensible code:
https://github.com/InsightSoftwareConsortium/ITK/blob/v5.3.0/Modules/IO/ImageBase/include/itkImageFileWriter.h#L245-L254

Modules/Core/Common/include/itkImageAlgorithm.hxx
```diff
@@ -36,8 +36,8 @@ ImageAlgorithm::DispatchedCopy(const InputImageType *                       inIm
 {
   if (inRegion.GetSize()[0] == outRegion.GetSize()[0])
   {
-    itk::ImageScanlineConstIterator<InputImageType> it(inImage, inRegion);
-    itk::ImageScanlineIterator<OutputImageType>     ot(outImage, outRegion);
+    itk::ImageScanlineConstIterator it(inImage, inRegion);
```
dzenanz, line 39: I assume this can be done for other iterator types? It would save a lot of typing. And, of course, if someone wants to explicitly provide the image type, they can do so.

dzenanz (commented):
I like it.

Modules/Core/Common/include/itkImageAlgorithm.hxx
N-Dekker: Note that for the old ITK iterator types, a deduction guide is required, in order to support deduction from a `SmartPointer<ImageType>`. Whereas the deduction guides of ImageBufferRange and ImageRegionRange (added with PR #3998) are only there to suppress those warnings (`-Wctad-maybe-unsupported`).

Modules/Core/Common/include/itkImageScanlineConstIterator.h
dzenanz: Before we had that implementation with R-value references, I used to have two signatures of `WriteImage`, one with SmartPointer and one with RawPointer parameter (and one signature delegate to the other). I thought that CTAD could be used to write more elegant and comprehensible code which does not need to resort to R-value references.

N-Dekker: Apparently, there is a paper that proposes deduction guides for function templates: Mike Spertus, 2018-10-08, 
[Improving function templates with Class Template Argument Deduction](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1167r0.html)  Is that what you mean? Not yet included with C++17, obviously.

N-Dekker: Of course I'm still in favor of pass-by-reference, instead of pass-by-pointer, when the passed object must be valid (not "null") and the ownership is not "passed" to the function. (That is, when the lifetime of the object is not _extended_ by passing it to the function.)

dzenanz: I didn't know that proposal existed, but I want it 😄

For now, we could add CTAD guides for classes, where useful.

N-Dekker:
An itkStatisticsUniqueLabelMapFilterTest1 test failure from ITK.Linux/Linux-Build9966, at https://open.cdash.org/test/1042206954 It does not seem to be caused by the proposed changes of this PR, as far as I can see. Does anyone have a clue?

dzenanz:
That test fails occasionally for unknown reasons, probably something related to threading.

hjmjohnson (approved):
LGTM.

Modules/Core/Common/include/itkImageScanlineConstIterator.h
```diff
@@ -102,7 +102,7 @@ class ITK_TEMPLATE_EXPORT ImageScanlineConstIterator : public ImageConstIterator
 
   /** Constructor establishes an iterator to walk a particular image and a
    * particular region of that image. */
-  ImageScanlineConstIterator(const ImageType * ptr, const RegionType & region)
+  ImageScanlineConstIterator(const TImage * ptr, const RegionType & region)
```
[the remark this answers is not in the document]
N-Dekker: > good change. 
> Doing this universally and consistently through the package will be a nearly impossible challenge.

@hjmjohnson Well... we could just start by replacing the use of `ImageType` aliases by the corresponding `TImage` template arguments where they are used as parameter type of non-default constructors. (Which are typically the constructors of class templates that could support CTAD.)  I think that would be feasible  😃 Feel free to give it a try!

PR #4016, opened 2023-04-19
Use C++17 "constexpr if" in `ImageRegionRange` and `ImageHelper` instead of  `true_type`/`false_type` tag dispatching

N-Dekker:
Used "constexpr if" instead of overloading helper functions for `std::true_type` and `std::false_type` (tag dispatching), in order to simplify the code.

### DIFF
--- a/Modules/Core/Common/include/itkImageHelper.h
+++ b/Modules/Core/Common/include/itkImageHelper.h
@@ -18,8 +18,7 @@
 #ifndef itkImageHelper_h
 #define itkImageHelper_h
 
-#include "itkMacro.h"  // For ITK_TEMPLATE_EXPORT.
-#include <type_traits> // For true_type, false_type, and integral_constant.
+#include "itkMacro.h" // For ITK_TEMPLATE_EXPORT.
 
 namespace itk
 {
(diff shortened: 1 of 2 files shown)


Modules/Core/Common/include/itkImageRegionRange.h
```diff
@@ -146,52 +146,50 @@ class ImageRegionRange final
...
-    void Increment(std::true_type) noexcept
+    void
+    Increment() noexcept
     {
-      static_assert(VIndex < (ImageDimension - 1), "For a larger index value, the other overload should be picked");
-
       m_BufferIterator += m_OffsetTable[VIndex];
 
-      if (static_cast<SizeValueType>(++m_IterationOffset[VIndex]) >= m_IterationRegionSize[VIndex])
+      if constexpr (VIndex < (ImageDimension - 1))
```
dzenanz, line 154: Thanks for reminding me we can now use this!

Some other things which we can now use:
https://www.codingame.com/playgrounds/2205/7-features-of-c17-that-will-simplify-your-code/introduction

dzenanz (approved):

Modules/Core/Common/include/itkImageRegionRange.h
dzenanz: That article suggests that instances of `enable_if` are candidates for replacement with `if constexpr`.

N-Dekker: Thanks for your encouragement. In the meantime I found two cases of "old-style compile-time _if_" by std::true_type/std::false_type overloading: in [itkImageHelper.h](https://github.com/InsightSoftwareConsortium/ITK/blob/b7bfef30a8729118131db908e1c423aeb734c219/Modules/Core/Common/include/itkImageHelper.h). Preparing another commit for this PR to address those as well, please wait  😃  If you see other possible use cases for _constexpr if_, please just try it out!

dzenanz (approved):

Modules/Core/Common/include/itkImageHelper.h
```diff
@@ -62,73 +61,55 @@ class ImageHelper
 
   /** ComputeIndex with recursive templates */
   inline static void
-  ComputeIndex(const IndexType &     bufferedRegionIndex,
-               OffsetValueType       offset,
-               const OffsetValueType offsetTable[],
-               IndexType &           index)
+  ComputeIndex(const IndexType &                      bufferedRegionIndex,
+               OffsetValueType                        offset,
+               [[maybe_unused]] const OffsetValueType offsetTable[],
```
N-Dekker, line 66: Note that this pull request uses the C++17 attribute `[[maybe_unused]]`, to avoid warnings appearing at the `else` part of the "constexpr if".

thewtex (approved):
Clean! :sparkle:

PR #4018, opened 2023-04-19
Remove local ImageScanlineIterator type aliases, using C++17 CTAD, and extend `ImageScanlineConstIterator` deduction guide

N-Dekker:
Removed the local type aliases of `ImageScanlineIterator` and `ImageScanlineConstIterator` that were previously only used to declare a single local variable, and used C++17 class template argument deduction instead.

### DIFF
--- a/Modules/Core/Common/include/itkImageScanlineConstIterator.h
+++ b/Modules/Core/Common/include/itkImageScanlineConstIterator.h
@@ -19,6 +19,7 @@
 #define itkImageScanlineConstIterator_h
 
 #include "itkImageIterator.h"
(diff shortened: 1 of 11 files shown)


N-Dekker:
> I assume we can, and should, do the same simplification for other image iterator types?

Thanks for your approval, @dzenanz Of course, this is just a style improvement, I would not say "we should". But I do think it very much improves code readability. In general, when a typedef is used only once, the typedef may not be needed anyway. Moreover, I feel that CTAD makes it even easier to just use the original template name, instead of some helper typedef, when declaring a variable.

dzenanz (approved):

PR #4021, opened 2023-04-20
WIP: Fix compile errors on VS2017

dzenanz:
They were introduced around the time C++17 was enabled. See the list here:
https://open.cdash.org/viewBuildError.php?type=0&buildid=8592346

This only fixes a subset of compile errors, but after these changes compiler complains about something else. Not only do these changes seem unending, but they reduce code readability too. Is it time to drop VS2017 from the list of supported compilers?

### DIFF
--- a/Modules/Core/Common/include/itkConstShapedNeighborhoodIterator.h
+++ b/Modules/Core/Common/include/itkConstShapedNeighborhoodIterator.h
@@ -85,10 +85,10 @@ public:
 
   /** Inherit type alias from superclass */
   using typename Superclass::OffsetType;
-  using OffsetValueType = typename OffsetType::OffsetValueType;
+  using OffsetValueType = itk::OffsetValueType;
   using typename Superclass::RadiusType;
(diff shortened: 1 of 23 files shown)


thewtex (approved):

N-Dekker:
> Is it time to drop VS2017 from the list of supported compilers?

Would be fine to me to drop VS2017. Even though officially VS2017 _does_ support C++17. According to https://learn.microsoft.com/en-us/lifecycle/products/visual-studio-2017 its Mainstream End Date was more than a year ago

hjmjohnson:
@dzenanz  Yes!  I think it is time to drop VS2017 support for 5.4.0

dzenanz:
Closing this incomplete PR in favor of dropping support for VS2017. @N-Dekker we need to update the versions checks and documentation.

N-Dekker:
> @N-Dekker we need to update the versions checks and documentation.

@dzenanz Can you please make that a PR?  😃

PR #4026, opened 2023-04-24
STYLE: Use trailing return type instead of typename + dependent type

dzenanz:
This is a follow-up to b18668b639becf9e3dc6324ca68f915ab7bf0a1e and b37ac532060361739a2653bbbd2176dee0282d09.

It reapplied regex replace from b18668b639becf9e3dc6324ca68f915ab7bf0a1e.

## PR Checklist

### DIFF
--- a/Modules/Core/Transform/include/itkScaleTransform.hxx
+++ b/Modules/Core/Transform/include/itkScaleTransform.hxx
@@ -261,8 +261,8 @@ ScaleTransform<TParametersValueType, VDimension>::ComputeMatrix()
 
 // Back transform a point
 template <typename TParametersValueType, unsigned int VDimension>
-inline typename ScaleTransform<TParametersValueType, VDimension>::InputPointType
-ScaleTransform<TParametersValueType, VDimension>::BackTransform(const OutputPointType & point) const
+inline auto
+ScaleTransform<TParametersValueType, VDimension>::BackTransform(const OutputPointType & point) const -> InputPointType
 {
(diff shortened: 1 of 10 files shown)


N-Dekker:
Thanks @dzenanz Still don't know why I overlooked these when I originally ran the regex! Sorry! 🤷 

The regex certainly did overlook lines of code that started with `inline`. Can you please check if these are now also included? I mean:

    Find what: ^inline typename (\w+<.+>::)(.+)\r\n\1(.+)\r\n{\r\n

N-Dekker (approved):

Modules/Core/Transform/include/itkScaleTransform.hxx
```diff
@@ -261,8 +261,8 @@ ScaleTransform<TParametersValueType, VDimension>::ComputeMatrix()
 
 // Back transform a point
 template <typename TParametersValueType, unsigned int VDimension>
-inline typename ScaleTransform<TParametersValueType, VDimension>::InputPointType
-ScaleTransform<TParametersValueType, VDimension>::BackTransform(const OutputPointType & point) const
+auto
```
N-Dekker, line 264: Do we also want to remove all those `inline` keywords? I don't think they really matter, but the original author(s) seemed to care about them... 🤔

dzenanz: You are absolutely right. It was search and replace oversight. I will amend the commit

PR #4027, opened 2023-04-24
Replace ImageScanlineIterator `while` loops with `for` loops, move iterator declarations into `for` loop init-statements

N-Dekker:
Looked for iterations of the following form:

    ImageScanlineIterator outIt(output, outputRegionForThread);

    while (!outIt.IsAtEnd())
    {
      ...
      outIt.NextLine();
    }

And replaced them with the corresponding `for` loops.

Follow-up to pull request https://github.com/InsightSoftwareConsortium/ITK/pull/3957 commit 31c713eb105eff4ba87a457a24a23ef4a6eb3071 "STYLE: Replace `while (!outIt.IsAtEnd())` with `for` loops, in Filtering"

### DIFF
--- a/Modules/Filtering/DisplacementField/include/itkTransformToDisplacementFieldFilter.hxx
+++ b/Modules/Filtering/DisplacementField/include/itkTransformToDisplacementFieldFilter.hxx
@@ -177,9 +177,6 @@ TransformToDisplacementFieldFilter<TOutputImage, TParametersValueType>::Nonlinea
   OutputImageType *     output = this->GetOutput();
   const TransformType * transform = this->GetInput()->Get();
 
-  // Create an iterator that will walk the output region for this thread.
-  ImageScanlineIterator outIt(output, outputRegionForThread);
-
   // Define a few variables that will be used to translate from an input pixel
   // to an output pixel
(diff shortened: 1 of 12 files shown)


dzenanz (approved):
And this is just in the filtering module?

blowekamp:
The choice of using the while loop was not arbitrary. Several loops structures were tested across platforms and the one with the best performance which enabled vectorization of the inner loop was chosen. I don't recall the details, but this existing structure was not arbitrary at the time.

N-Dekker:
> The choice of using the while loop was not arbitrary. Several loops structures were tested across platforms and the one with the best performance which enabled vectorization of the inner loop was chosen. I don't recall the details, but this existing structure was not arbitrary at the time.

Thanks Bradley, I'm interested to see the details. But in general, I think it's just fine to use a `for` loop, when iterating by a `ImageScanlineIterator`. One of the advantages is that it reduces the scope of the iterator object to the loop itself. Moreover, the code appears easier to read when the initialization, the "end" test, and the "Next" statement of the iterator are brought together this way.

hjmjohnson (approved):

hjmjohnson (approved):

PR #4040, opened 2023-05-05
ENH: Add series_uid arg to imread() for dicom series selection

### LINKED ISSUE #4038: itk.imread() python function in extras.py should have series_uid as an optional argument
### Actual and Proposed Behavior

If you pass a directory name to the itk.imread() function, and if that directory contains one and only one DICOM series, then that series is read as an image.  Otherwise fails the function fails.

Three other options would be supported by having series_uid as an optional argument
1) If series_uid is a string and matches the name of a series_uid in the directory, that series should be read.
2) If series_uid is an integer, then the series_uid'th series in that directory should be read.
3) If series_uid isn't specified, then the first series in that directory should be read by default.

### Additional Information

Main challenge may be that a series_uid might not correspond to a single image.  In such a situation, which image should be read when given a series_uid is unclear.   ITK splits such series into multiple sub-series, so the series_uid that should be passed is munged - potentially causing unexpected behaviour.

aylward:
Typical usage:

`img = itk.imread(filename="./my_dicom_dir", series_uid="1.2.3.456.789.101112.1314")`

If filename is a directory of DICOM files, and series_uid is a string, it will read the DICOM series that matches this series_uid name.  If series_uid is an integer, it will read the series_uid'th DICOM series in the directory (using python notation, e.g., -1 is the last DICOM series in the directory). If unspecified, it will read the first series in the directory.

This does represent a change in the default behavior in that it used to raise an exception if a DICOM directory with more than one series was passed to imread, and now it will read the first series in that directory.

Closes #4038 

## PR Checklist
- Extends API of imread() in wrapping/generators/python/itk/support/extras.py to have series_uid as an argument
- No major design changes
- Tests added to wrapping/generators/testing/extras.py
- Updated API documentation

### DIFF
--- a/Wrapping/Generators/Python/Tests/CMakeLists.txt
+++ b/Wrapping/Generators/Python/Tests/CMakeLists.txt
@@ -63,6 +63,10 @@ if(ITK_WRAP_unsigned_char AND WRAP_2)
         ${ITK_TEST_OUTPUT_DIR}/extras-mushroom.vtk
         ${ITK_TEST_OUTPUT_DIR}
         ${ITK_TEST_OUTPUT_DIR}/LinearTransformWritten.h5
+        ${ExternalData_BINARY_ROOT}/Testing/Data/Input/DicomSeries
+      )
+    set_tests_properties( PythonExtrasTest PROPERTIES DEPENDS
+      PythonDICOMSeries
       )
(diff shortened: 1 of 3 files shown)


dzenanz (commented):
LGTM, but Matt should review.

thewtex (approved):
Awesome! :100:

PR #4041, opened 2023-05-08
COMP: gcc-13 compatability

ramen-newdals:
Using gcc-13 ITK does not compile as discussed in this form entry: (https://discourse.itk.org/t/building-itk-error-with-algorithim-module/5927/9). Was able to be fixed thanks to @dzenanz with the inclusion of  <cstdint> to itkMathematicalMorphologyEnums.h and specifying std::uint8_t on line 41.


dzenanz (approved):

hjmjohnson (approved):

N-Dekker:
FYI, I'm just proposing is a little style adjustment to this fix:

- #4387

PR #4042, opened 2023-05-08
COMP: Update maintenance script to compile with Visual Studio 2022

dzenanz:


dzenanz:
I was trying to apply it to Ultrasound remote module, but it did not catch [this change](https://github.com/KitwareMedical/ITKUltrasound/pull/225) which I made manually.

Also, I thought that PRs in Utilities directory should not trigger CI builds?

N-Dekker:
> Update maintenance script to compile with Visual Studio 2022

Is it VS2022 specific, or is it C++17 specific? I mean, in VS2022 you can still enable C++14, wouldn't that allow compiling the file?

dzenanz:
You are right. It is more C++17 related, than Visual Studio. I was to lazy to look up when that experimental stuff stopped being valid. What would be a better commit message?

N-Dekker (approved):
Thanks @dzenanz ! 👍

hjmjohnson (approved):

PR #4044, opened 2023-05-10
BUG: Add missing string value for NoTargets enum

blowekamp:
Fixed printing of default constructed FastMarchingUpwindGradient
containing: "INVALID VALUE FOR
itk::FastMarchingUpwindGradientImageFilterEnums::TargetCondition" for
m_TargetReachedMode member.

## PR Checklist

### DIFF
--- a/Modules/Filtering/FastMarching/test/itkFastMarchingUpwindGradientTest.cxx
+++ b/Modules/Filtering/FastMarching/test/itkFastMarchingUpwindGradientTest.cxx
@@ -388,7 +388,9 @@ itkFastMarchingUpwindGradientTest(int, char *[])
   };
   for (const auto & ee : allConditions)
   {
-    std::cout << "STREAMED ENUM VALUE FastMarchingUpwindGradientImageFilterEnums::TargetCondition: " << ee << std::endl;
+    std::ostringstream ss;
+    ss << ee;
+    ITK_TEST_EXPECT_TRUE(ss.str().find("INVALID") == std::string::npos);
   }
 
   if (passed)
--- a/Modules/Filtering/FastMarching/src/itkFastMarchingUpwindGradientImageFilter.cxx
+++ b/Modules/Filtering/FastMarching/src/itkFastMarchingUpwindGradientImageFilter.cxx
@@ -26,6 +26,8 @@ operator<<(std::ostream & out, const FastMarchingUpwindGradientImageFilterEnums:
   return out << [value] {
     switch (value)
     {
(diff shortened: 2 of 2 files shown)


dzenanz (approved):

Modules/Filtering/FastMarching/test/itkFastMarchingUpwindGradientTest.cxx
```diff
@@ -388,7 +388,9 @@ itkFastMarchingUpwindGradientTest(int, char *[])
   };
   for (const auto & ee : allConditions)
   {
-    std::cout << "STREAMED ENUM VALUE FastMarchingUpwindGradientImageFilterEnums::TargetCondition: " << ee << std::endl;
+    std::ostringstream ss;
+    ss << ee;
+    ITK_TEST_EXPECT_TRUE(ss.str().find("INVALID") == std::string::npos);
```
jhlegarreta, line 393: If this way to print the enum values is preferred over the other one, it should be put to the SW Guide.

jhlegarreta (approved):

Modules/Filtering/FastMarching/test/itkFastMarchingUpwindGradientTest.cxx
blowekamp: I used the output operator to render the enum  into a string so that the results could be validated. Not sure it's a different way to print enums that needs to documented. Just an approach for testing and validation.

jhlegarreta: I'm fine with the approach; I just think that if we now prefer this approach then we should put it into the SW Guide. Thanks Brad.

jhlegarreta:
BTW, can a commit message body be added? Minimally, copying the subject line and saying error reported at runtime, if possible.

Modules/Filtering/FastMarching/test/itkFastMarchingUpwindGradientTest.cxx
jhlegarreta: > Does the SWG contain recommendations for how to test enum to to string conversions? 

It does not I'm afraid, but would be good to add it.

> Otherwise this approach is specific to this use case and too specific for the SWG.

I think the approach is valid for all enums. Not sure why it shouldn't.

I do not mean it has to be done now, but I think it is good to have for the sake of consistency. Anyways, it is not a show-stopper for this PR to be merged.

PR #4047, opened 2023-05-17
COMP: Workaround itkPyCommand KWStyle check "error: namespace is wrong"

N-Dekker:
Worked around the KWStyle check failure ("error: namespace is wrong"), by moving the unnamed namespace of itkPyCommand.cxx inside the `itk` namespace. Apparently, KWStyle expects the first opened namespace in an ITK source file to be named `itk`.

----
Note: From a C++ point of view I think it's entirely correct to have an unnamed namespace at the begin of an ITK cxx file (even before opening the `itk` namespace). So I think the KWStyle check failure is undeserved. But well, that's why it's just a workaround 😺 The KWStyle check appears implemented at https://github.com/Kitware/KWStyle/blob/1173206c0e7f4bc70dc3af641daf6e33f4042772/kwsCheckNamespace.cxx

### DIFF
--- a/Wrapping/Generators/Python/PyUtils/itkPyCommand.cxx
+++ b/Wrapping/Generators/Python/PyUtils/itkPyCommand.cxx
@@ -18,6 +18,9 @@
 
 #include "itkPyCommand.h"
 
+namespace itk
+{
+
 namespace
 {
 // Wrapper to automatics obtain and release GIL
@@ -32,9 +35,6 @@ private:
   PyGILState_STATE m_GIL;
 };
 } // end anonymous namespace
-
-namespace itk
-{
 
 PyCommand::PyCommand()
 {


dzenanz (approved):

N-Dekker:
Instead of modifying itkPyCommand.cxx like this, I wouldn't mind if this particular KWStyle check would be removed (or at least: _skipped_). What do you think?

dzenanz:
I am totally fine with this modification.

thewtex:
@N-Dekker this can be avoided by overriding the namespace checkout for this file -- see the *Utilities/KWStyle/ITKOverwrite.txt* file.

N-Dekker:
Thanks for your comments so far, @dzenanz and @thewtex. Instead of this PR (4047), please consider:
- #4048

N-Dekker:
> I like this PR better than the other one.

Thanks @dzenanz I'm not opposed to this PR, but I like the other one (#4048) better 😸

thewtex:
There are valid argument for each, and I think either would be fine to merge. Personally, I also prefer #4048 -- yes, it is applied to all .cxx files, but this is a common enough issue in my experience.

N-Dekker:
> There are valid argument for each, and I think either would be fine to merge. Personally, I also prefer #4048

Thanks Matt. Actually I think the most ideal solution would be to extend the [KWStyle Namespace check](https://github.com/Kitware/KWStyle/blob/1173206c0e7f4bc70dc3af641daf6e33f4042772/kwsCheckNamespace.cxx) to allow top-level unnamed namespaces and [C++17 nested namespaces](http://www.nuonsoft.com/blog/2017/08/01/c17-nested-namespaces/) (as long as they are nested within the `itk` namespace). But that's quite some more work than just working around the warning (this PR), or disabling the warning for any cxx file (PR #4048)!

I'll reopen this PR now (Ready for review, so ready for merge). Although I still prefer #4048, I leave it up to the reviewers which one gets merged 😃

hjmjohnson:
Resolved in #4048

PR #4048, opened 2023-05-18
COMP: Disable KWStyle check "error: namespace is wrong" on *.cxx files

N-Dekker:
*.cxx files may have an unnamed namespace, instead of `namespace itk`, as their first namespace. For example Wrapping/Generators/Python/PyUtils/itkPyCommand.cxx This would cause undeserved failures of the `Namespace` check of KWStyle.

----
Aims to supersede (i.e. replace) pull request #4047:
- #4047

Based on a comment from @thewtex at https://github.com/InsightSoftwareConsortium/ITK/pull/4047#issuecomment-1552084915

### DIFF
--- a/Utilities/KWStyle/ITKOverwrite.txt
+++ b/Utilities/KWStyle/ITKOverwrite.txt
@@ -1,3 +1,2 @@
-test/.*\.cxx Namespace Disable
-Examples/.*\.cxx Namespace Disable
+.*\.cxx Namespace Disable
 QuickView.h Namespace Disable


Utilities/KWStyle/ITKOverwrite.txt
```diff
@@ -1,3 +1,2 @@
-test/.*\.cxx Namespace Disable
-Examples/.*\.cxx Namespace Disable
+.*\.cxx Namespace Disable
```
dzenanz, line 1: This disabling might be overly broad.

dzenanz: Yes, but I think that applies to all of ITK.

hjmjohnson: I think these namespace rules should only be enforced in the public API (Headers).

hjmjohnson (approved):

PR #4053, opened 2023-05-25
Gdcm update 2023 05 24

seanm:
Just sync from upstream.


dzenanz:
@malaterre should we ignore these [ghostflow warnings](https://github.com/InsightSoftwareConsortium/ITK/pull/4053/checks?check_run_id=13743018210)? Or can they be fixed?

seanm:
@dzenanz there were also warnings when I committed about some files sizes:

```
The path 'Modules/ThirdParty/GDCM/src/gdcm/Source/DataDictionary/privatedicts.xml' has size 1033 KiB, greater than allowed 1024 KiB.
The path 'Modules/ThirdParty/GDCM/src/gdcm/Source/InformationObjectDefinition/Part3.xml' has size 2282 KiB, greater than allowed 1024 KiB.
```

These messages don't make sense to me because the max size is already increased for those files in `Modules/ThirdParty/GDCM/src/gdcm/.gitattributes`

I had to commit bypassing hooks.

seanm:
> To avoid that, max size should be increased in GDCM repository already

What I'm saying is: it already is: https://github.com/malaterre/GDCM/blob/master/.gitattributes

dzenanz:
@bradking do you have an explanation for this?

bradking:
I can't reproduce the local hook failure.  I created a branch from commit `6adfd15ab92` and ran `./Modules/ThirdParty/GDCM/UpdateFromUpstream.sh`.  It produced a merge commit with content identical to the branch here.

Note that the ghostflow check failure in this PR is unrelated.  The encoding error should probably be fixed in GDCM upstream first.

seanm:
>I can't reproduce the local hook failure.

Weird. But I guess not a big deal. Like I said, I just bypassed the hook and committed manually. Maybe it only occurs with certain git versions or something...

seanm:
Those non-ASCII characters are not present in GDCM master.  @malaterre maybe you need to cherry pick something into the release branch?

dzenanz:
Let's give @malaterre a few more days to pitch in.

seanm:
I believe those files with non-ASCII chars aren't really even used in ITK...

dzenanz:
Hopefully, the next update will fix them.

dzenanz:
PR #4061 is a follow-up update.

PR #4049, opened 2023-05-18
COMP: Add explicit default copy constructor

blowekamp:
Addresses complation warning with clang-1316.0.21.2.5:

definition of implicit copy constructor for 'QualifiedIterator<true>' is deprecated because it has a user-declared copy assignment operator

## PR Checklist
- [ ] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [ ] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [ ] Added test (or behavior not changed)
- [ ] Updated API documentation (or API not changed)
- [ ] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [ ] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [ ] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples) for all new major features (if any)

Refer to the [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) for
further development details if necessary.

PR #4054, opened 2023-05-26
COMP: Rule of Zero, SFINAE for ImageBufferRange, ImageRegionRange iterators, update ShapedImageNeighborhoodRange documentation

N-Dekker:
Used SFINAE (`enable_if`) to implement non-const to const conversion for the iterator types of `ImageBufferRange` and `ImageRegionRange`. Followed the Rule of Zero for those types.

Prevents potential clang warnings, as reported before (pull request #4049) by Bradley Lowekamp (@blowekamp):

> definition of implicit copy constructor for 'QualifiedIterator<true>'
> is deprecated because it has a user-declared copy assignment operator

Documented that the iterator types of `ShapedImageNeighborhoodRange` now unconditionally follow the Rule of Zero, and did the same for the iterator types of `ImageBufferRange` and `ImageRegionRange`.

Follow-up to pull request #4049 commit df6aa66e40c6e4ce91c05df95f5364f3558ab43a "COMP: QualifiedIterator follow rule of zero", merged on 19 May, 2023

### DIFF
--- a/Modules/Core/Common/include/itkImageBufferRange.h
+++ b/Modules/Core/Common/include/itkImageBufferRange.h
@@ -23,7 +23,7 @@
 #include <cstddef>  // For ptrdiff_t.
 #include <iterator> // For random_access_iterator_tag.
 #include <limits>
-#include <type_traits> // For conditional, is_same, and is_const.
+#include <type_traits> // For conditional, enable_if, is_same, and is_const.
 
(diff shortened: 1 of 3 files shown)


blowekamp (approved):
Nice!

dzenanz (approved):

PR #4057, opened 2023-05-30
STYLE: Remove initial `<<` insertion from `itkExceptionMacro(<< "` calls

N-Dekker:
Let the C++ preprocessor append the specified literal string to the error message, instead of doing the equivalent `<<` insertion at run-time.

Aims to shorten the code a little bit, and improve consistency. The code base already had more than 750 `itkExceptionMacro` calls that passed a literal string directly, without having `<<` at its left side.


dzenanz (approved):

Modules/IO/GDCM/src/itkGDCMImageIO.cxx
```diff
@@ -1211,14 +1211,14 @@ GDCMImageIO::Write(const void * buffer)
       outpixeltype.SetPixelRepresentation(static_cast<unsigned short>(std::stoi(pixelRep.c_str())));
       if (this->GetNumberOfComponents() != 1)
       {
-        itkExceptionMacro(<< "Sorry Dave I can't do that");
+        itkExceptionMacro("Sorry Dave I can't do that");
```
issakomi, line 1214: “I'm sorry, Dave. I'm afraid I can't do that,” said HAL 9000
(c) 2001: A Space Odyssey

N-Dekker: Thanks for the reference, Mihail! Honestly I doubt if such an error message is very helpful to the user. In general, I'm not a fan of "funny" error messages. Unless they really help to communicate to the user what's going wrong. But anyway, that's beyond the scope of this pull request!

dzenanz: Maybe it is an error message which is not supposed to be reachable by the GDCM's users?

N-Dekker: Maybe... but then I guess the message only for someone named Dave 😸  Thanks for merging @dzenanz !

N-Dekker:
@N-Dekker 
> Aims to shorten the code a little bit

Oops, I see now:

    +1,418 −1,394

Meaning that the PR has actually _added_ more lines of code! Oops! I'm sorry  😸 !

PR #4058, opened 2023-05-31
COMP: Update itkExternal_Eigen3.cmake

ningfei:
Pass CMAKE_OSX_ARCHITECTURES to fix cross-compiling on Apple Silicon. Otherwise, compiler won't be set properly and cmake will fail to configure.


dzenanz (approved):

phcerdan (approved):
Thanks!

PR #4060, opened 2023-06-02

jhlegarreta:
- Use other testing macros as necessary to check for expected values.

## PR Checklist

### DIFF
--- a/Modules/Filtering/DisplacementField/test/itkTimeVaryingVelocityFieldIntegrationImageFilterTest.cxx
+++ b/Modules/Filtering/DisplacementField/test/itkTimeVaryingVelocityFieldIntegrationImageFilterTest.cxx
@@ -19,6 +19,7 @@
 #include "itkTimeVaryingVelocityFieldIntegrationImageFilter.h"
 #include "itkImportImageFilter.h"
 #include "itkTestingMacros.h"
+#include "itkVectorLinearInterpolateImageFunction.h"
 
 int
 itkTimeVaryingVelocityFieldIntegrationImageFilterTest(int argc, char * argv[])
@@ -71,6 +72,21 @@ itkTimeVaryingVelocityFieldIntegrationImageFilterTest(int argc, char * argv[])
 
   ITK_EXERCISE_BASIC_OBJECT_METHODS(integrator, TimeVaryingVelocityFieldIntegrationImageFilter, ImageToImageFilter);
 
+
+  auto velocityFieldInterpolator =
+    itk::VectorLinearInterpolateImageFunction<IntegratorType::TimeVaryingVelocityFieldType,
+                                              IntegratorType::ScalarType>::New();
+  integrator->SetVelocityFieldInterpolator(velocityFieldInterpolator);
+  ITK_TEST_SET_GET_VALUE(velocityFieldInterpolator, integrator->GetVelocityFieldInterpolator());
+
+  auto displacementFieldInterpolator =
+    itk::VectorLinearInterpolateImageFunction<IntegratorType::DisplacementFieldType, IntegratorType::ScalarType>::New();
+  integrator->SetDisplacementFieldInterpolator(displacementFieldInterpolator);
+  ITK_TEST_SET_GET_VALUE(displacementFieldInterpolator, integrator->GetDisplacementFieldInterpolator());
+
+  auto initialDiffeomorphism = IntegratorType::DisplacementFieldType::New();
+  integrator->SetInitialDiffeomorphism(initialDiffeomorphism);
+  ITK_TEST_SET_GET_VALUE(initialDiffeomorphism, integrator->GetInitialDiffeomorphism());
 
   auto lowerTimeBound = static_cast<typename IntegratorType::RealType>(std::stod(argv[1]));
   integrator->SetLowerTimeBound(lowerTimeBound);
--- a/Modules/Registration/Common/test/itkMatchCardinalityImageToImageMetricTest.cxx
+++ b/Modules/Registration/Common/test/itkMatchCardinalityImageToImageMetricTest.cxx
@@ -106,5 +106,13 @@ itkMatchCardinalityImageToImageMetricTest(int argc, char * argv[])
     }
   }
 
+
+  MetricType::ParametersType parameters = transform->GetParameters();
+  MetricType::DerivativeType derivative;
+  metric->GetDerivative(parameters, derivative);
+
+  MetricType::DerivativeType derivative1{};
+  ITK_TEST_EXPECT_EQUAL(derivative, derivative1);
+
   return EXIT_SUCCESS;
 }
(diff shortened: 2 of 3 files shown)


Modules/Registration/Common/test/itkMatchCardinalityImageToImageMetricTest.cxx
```diff
@@ -106,5 +106,13 @@ itkMatchCardinalityImageToImageMetricTest(int argc, char * argv[])
     }
   }
 
+
+  typename MetricType::ParametersType parameters = transform->GetParameters();
```
N-Dekker, line 110: Thanks Jon. Note that the `typename` keyword is not necessary here. Line 54 uses MetricType::ParametersType without the keyword. `typename` is only necessary for "dependent types" inside templates.

Modules/Registration/Common/test/itkMatchCardinalityImageToImageMetricTest.cxx
```diff
@@ -106,5 +106,13 @@ itkMatchCardinalityImageToImageMetricTest(int argc, char * argv[])
...
+  typename MetricType::DerivativeType derivative;
+  metric->GetDerivative(parameters, derivative);
+
+  typename MetricType::DerivativeType derivative1 = typename MetricType::DerivativeType{};
```
N-Dekker, line 114: In general, any variable declaration of the form `T var = T{}` could be simplified to just `T var{}`. So in this case:

    MetricType::DerivativeType derivative1{};

jhlegarreta: Done in [c38c358](https://github.com/InsightSoftwareConsortium/ITK/pull/4060/commits/c38c3582da50e0a463da7f8174709cc1ad1c2180).

Modules/Registration/Common/test/itkMatchCardinalityImageToImageMetricTest.cxx
N-Dekker: >  if I am not mistaken, in the past, OSX has complained about this

If so, then OSX must have been mistaken! A `typename` like this should never be necessary outside of templates.

Modules/Filtering/DisplacementField/test/itkTimeVaryingVelocityFieldIntegrationImageFilterTest.cxx
```diff
@@ -72,6 +73,22 @@ itkTimeVaryingVelocityFieldIntegrationImageFilterTest(int argc, char * argv[])
...
 
+  auto velocityFieldInterpolator =
+    itk::VectorLinearInterpolateImageFunction<typename IntegratorType::TimeVaryingVelocityFieldType,
+                                              typename IntegratorType::ScalarType>::New();
+  integrator->SetVelocityFieldInterpolator(velocityFieldInterpolator);
+  ITK_TEST_SET_GET_VALUE(velocityFieldInterpolator, integrator->GetVelocityFieldInterpolator());
+
+  auto displacementFieldInterpolator =
+    itk::VectorLinearInterpolateImageFunction<typename IntegratorType::DisplacementFieldType,
+                                              typename IntegratorType::ScalarType>::New();
```
N-Dekker, line 84: Please remove these four `typename` keywords as well. Sorry!

jhlegarreta: Done in [bf183be](https://github.com/InsightSoftwareConsortium/ITK/pull/4060/commits/bf183be9c4a49edec634ae6748f7e24e223cedee). Let's see what the dashboard says after merging, but plenty of cases are found across tests.

N-Dekker: FYI, from C++11 on (including C++14 and C++17), it appears _allowed_  to use `typename` keywords outside of a template. But it's still _unnecessary_ to do so, outside of a template.

Clang even introduced a warning saying:

> warning: 'typename' occurs outside of a template

Looking at https://godbolt.org/z/6YqjEPGG5 
```
struct Wrapper
{
    using Type = int;
};

typename Wrapper::Type i; // warning: 'typename' occurs outside of a template
```

So in fact it's really a matter of style, whether to use `typename` keywords in such cases.

N-Dekker: FYI the C++ standard tried to avoid the _need_ for such `typename` keywords as much as possible, looking at a paper like [Down with `typename`!](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0634r3.html)  There is also a follow-up paper, proposing to remove unnecessary `typename` keywords from the `std` library: [Down with `typename` in the Library!](https://open-std.org/JTC1/SC22/WG21/docs/papers/2020/p2150r0.html)

jhlegarreta: Thanks for all he information Niels. Unfortunately, it is not about a particular coding habit, but about what the dashboard compilers report or have been reporting.

N-Dekker: > Unfortunately, it is not about a particular coding habit, but about what the dashboard compilers report or have been reporting.

If so, it could be some compiler bug, of course. A compiler should never require a `typename` keyword outside of a template. So if you encounter such a compile error again, please share a link to the specific dashboard page!

jhlegarreta: Niels, the tests are full of such examples, even the files I modified contain examples; if CI and dashboard logs lived forever, you could probably navigate across similar coverage improvement PRs, compare push forced commits, dashboard states, etc. and see that in many cases the push force commits were done because a compiler complained about this.

N-Dekker: Do you mean this one? https://github.com/InsightSoftwareConsortium/ITK/compare/a612518c4fbec77e94112820183977e6aa6d18b4..a68d987c5b2703dde20a803c3036691668100c7a
```
  typename ConstantVelocityFieldTransformType::ConstantVelocityFieldInterpolatorPointer
    constantVelocityFieldInterpolator{};
```

That is actually inside a template.

dzenanz (approved):

Modules/Filtering/DisplacementField/test/itkTimeVaryingVelocityFieldIntegrationImageFilterTest.cxx
jhlegarreta: Just for the records, no new warnings have appeared after merging this, so not sure about past behavior, and `typename` instances outside templates. Anyways, thanks for the explanations @N-Dekker.

PR #4082, opened 2023-06-19
STYLE: Remove redundant checks in `ProcessObject::VerifyPreconditions()`

N-Dekker:
Removed redundant checks of required named inputs, which were performed already in the previous lines of code.

Partially reverts commit bf6f1065a07037fe27d290f72adf81c81ed408ec "BUG: VerifyPreconditions checks required named inputs", April 23, 2013.

----

Please compare the code that is removed by this PR with the code just above it:

https://github.com/InsightSoftwareConsortium/ITK/blob/bc1025b1c00d8cd99d98646c6a59072edf0488f9/Modules/Core/Common/src/itkProcessObject.cxx#L1332-L1341

Isn't it basically just doing the same check?

### DIFF
--- a/Modules/Core/Common/src/itkProcessObject.cxx
+++ b/Modules/Core/Common/src/itkProcessObject.cxx
@@ -1341,20 +1341,6 @@ ProcessObject::VerifyPreconditions() ITKv5_CONST
   }
 
   /**
-   * Verify the require named inputs.
-   */
-  auto i = m_RequiredInputNames.begin();
-  while (i != m_RequiredInputNames.end())
-  {
-    if (this->GetInput(*i) == nullptr)
-    {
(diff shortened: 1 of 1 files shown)


dzenanz:
@blowekamp please review.

blowekamp:
Yes, I agree it looks redundant. The linked change did not add a test. Could we test a couple cases of names inputs being null?

dzenanz:
Ping! Test, yes or no? If not, should we merge as is?

N-Dekker:
A test would be nice, but I haven't had time to write it so far. And honestly, I think the test should have been written before, when the original code (which is removed by this PR) was added! If necessary, of course. I'm fine by having this little clean-up merged now.

PR #4085, opened 2023-06-20
PERF: Use `std::move` inside `itkSetMacro`, declare parameter non-const

N-Dekker:
Avoids an unnecessary copy (possibly involving memory allocation) of the argument of a Set member function.

Note that `itkSetMacro` calls that define a Set member function which passes the argument by const-reference must explicitly specify the `const` keyword now.

### DIFF
--- a/Modules/IO/XML/include/itkDOMNode.h
+++ b/Modules/IO/XML/include/itkDOMNode.h
@@ -93,11 +93,11 @@ public:
   GetParent() const;
 
   /** Retrieve the tag name of this node. */
-  itkSetMacro(Name, std::string &);
+  itkSetMacro(Name, const std::string &);
   itkGetConstReferenceMacro(Name, std::string);
 
   /** Retrieve the special attribute "id" of this node. */
-  itkSetMacro(ID, std::string &);
+  itkSetMacro(ID, const std::string &);
   itkGetConstReferenceMacro(ID, std::string);
 
   /** Retrieve an attribute by key (return an empty string if not found). */
--- a/Modules/Core/Common/include/itkMacro.h
+++ b/Modules/Core/Common/include/itkMacro.h
@@ -966,17 +966,17 @@ compilers.
   itkSetDecoratedObjectInputMacro(name, type);         \
   itkGetDecoratedObjectInputMacro(name, type)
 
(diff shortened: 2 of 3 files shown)


dzenanz (approved):

Modules/IO/XML/include/itkDOMNode.h
```diff
@@ -93,11 +93,11 @@ class ITKIOXML_EXPORT DOMNode : public Object
   GetParent() const;
 
   /** Retrieve the tag name of this node. */
-  itkSetMacro(Name, std::string &);
+  itkSetMacro(Name, const std::string &);
```
blowekamp, line 96: These are the only use cases for of `itkSetMacro\.\*\&`. Wouldn't it now make more sense to just pass by value and not reference?

N-Dekker: @blowekamp I think so, yes: pass by value usually seems to be the right thing to do, for a `Set` member function that passes a "regular" copyable argument.

This appears to be a rare case where itkSetMacro was used to pass by const-reference. I just left it that way, to avoid breaking the ABI unnecessarily. However, it could be changed to pass by value in a follow-up commit, if anyone wants to.

Some motivation: historically, in C++ an input argument of a class types was usually passed by const-reference, to avoid copying. However, with a Set member functions, the argument might as well be passed by value, as its value can now be "moved" into the corresponding data member.

hjmjohnson: I would love to see a followup request made with a ITK v 6.0.0 release tag added.

Hans

N-Dekker:
Thanks for merging, Hans. Note that `itkSetMacro(Name, std::string)` may now compete with the dedicated string macro, `itkSetStringMacro(Name)`!

`itkSetStringMacro(Name)` creates two overloaded member functions, `SetName(const std::string &)` and `SetName(const char *)`, while `itkSetMacro(Name, std::string)` only creates a single one, `SetName(std::string)`, but I think that's usually just enough now!

Modules/IO/XML/include/itkDOMNode.h
N-Dekker: > I would love to see a followup request made with a ITK v 6.0.0 release tag added.

- @hjmjohnson Here it is, finally: pull request #5637  😇

Issue #4074, opened 2023-06-17
itkMacros less robust / break external code that also has an itk namespace


kislinsk:
### Description

In 24e7b452b4d607154e68bb6fdd2ef5bbb5c2d1b4, leading colons for namespaces have been removed with an automated search and replace. Unfortunately this also affected some macros and made them less robust resp. introduced a regression. We have a few projects that also have itk (sub-)namespaces and the lookup for classes like LightObject and ObjectFactory<T> used in macros like itkNewMacro() is now broken in these cases. It is not a compiler-specific issue as it happens on MSVC, gcc, and clang.

### Expected behavior

Macros always should be as robust as possible in particular regarding parenthesis or namespaces.

### Versions

Worked in ITK v5.2.1, broken in ITK v5.3.0.

### Environment

Windows, Ubuntu, macOS, MSVC, GCC, clang.

thewtex:
@kislinsk thanks for the report. Can you please verify that https://github.com/InsightSoftwareConsortium/ITK/pull/4087 addresses your use case?

kislinsk:
@thewtex Thanks, it works!

Not an issue for us but you may want to consider to do the same for the std namespace in macros.

thewtex:
@kislinsk thanks for the verification!

Good idea regarding `std` -- let's add that if it is found that it is needed.

PR #4087, opened 2023-06-23
COMP: Use full ::itk namespace in itkMacro.h

thewtex:
To address resolution in client code. #4074

### DIFF
--- a/Modules/Core/Common/include/itkMacro.h
+++ b/Modules/Core/Common/include/itkMacro.h
@@ -304,26 +304,26 @@ namespace itk
   itkCloneMacro(x);         \
   ITK_MACROEND_NOOP_STATEMENT
 
-#define itkSimpleNewMacro(x)                            \
-  static Pointer New()                                  \
-  {                                                     \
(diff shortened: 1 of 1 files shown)


dzenanz (approved):

Leengit (commented):
Maybe Lines 349 and 764 too.  Generally, we could use the trailing backslash as a hint that a macro is being defined.  With that approach, we can edit all ITK C++ files outside of `ThirdParty` with something like
```bash
find git/ITK/* -type f | \
    egrep -i -e '\.[it]?(cc|hh|[ch](\+\+|pp?|xx)?)$' | \
    fgrep -v ThirdParty | \
    xargs sed -i -r -e 's/(::)?itk::(.*)\\$/::itk::\2\\/g'
```
We'd need a subsequent call to `clang_format` to prettify the spacing in those macros.

Leengit:
For modification of all C++ macros, see 28f373aa4d84be3f48ca747b3ad2a241f49f6911.

thewtex:
Adding specifiers on lines 349 and 764 as suggested by @Leengit.

PR #4092, opened 2023-06-27
DOC: Remove duplicate "type" in alias documentation

jhlegarreta:
Remove duplicate "type" in alias documentation.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkNeighborhood.h
+++ b/Modules/Core/Common/include/itkNeighborhood.h
@@ -85,7 +85,7 @@ public:
   /** Offset type used to reference neighbor locations */
   using OffsetType = Offset<VDimension>;
 
-  /** External slice iterator type type alias support */
+  /** External slice iterator type alias support */
   using SliceIteratorType = SliceIterator<TPixel, Self>;
 
   /** Type used to refer to space dimensions */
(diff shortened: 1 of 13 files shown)


jhlegarreta:
Both ways may be correct; depends on how we read them? Would appreciate the input of some very strict voice on the technicalities or terms.

Maybe using "type type alias" was a sort of divergence of e.g. "ImageType type alias", appearances like the latter being much more frequent, over 400 hits.

hjmjohnson:
I think this was a global substitution of "type typedef" to "type type alias" .  Your changes are good.

hjmjohnson (approved):

N-Dekker (approved):
Good catch, Jon! 👍 

Just for your information, it appears that the C++ Standard hasn't really abandoned the term "typedef". It hasn't even deprecated the old term! Looking at https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/n4950.pdf it defines the meaning of an _alias_ declared by `using` in the section _The `typedef` specifier_:

> A _typedef-name_ can also be introduced by an _alias-declaration_. The identifier following the `using` keyword is not looked up; it becomes a _typedef-name_ and the optional attribute-specifier-seq following the identifier appertains to that _typedef-name_. Such a _typedef-name_ has the same semantics as if it were introduced by the `typedef` specifier.

Nevertheless I think it's just fine to use the term _type alias_ nowadays  👍

thewtex (approved):

PR #3664, opened 2022-09-26
STYLE: Remove space between class and member names (follow-up)

N-Dekker:
Removed spaces between class and member names, that were
accidentally introduced with pull request https://github.com/InsightSoftwareConsortium/ITK/pull/1191
commit 2074c2f9e5ed3f087d0e2059f8e0e8992fcad7ef "STYLE: Enforce ITK style
defined by .clang-format".

Using bash commands like:

    find . \( -iname *.cxx \) -exec perl -pi -w -e 's/([A-Z]\w+) \:\:(\w)/$1::$2/g;' {} \;

Follow-up to pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2096
commit 734d6f84d728b343baae5ef66643a5568a5a30c7 "STYLE: Remove space between
class and member names in C++ source files"

PR #4110, opened 2023-07-14
BUG: DCMTK reader wrongly rejects file with preamble

### LINKED ISSUE #4108: DCMTK reader fails for valid files
### Description

Valid dicom files are rejected.

### Steps to Reproduce

Use the itkDCMTKImageIO to read a file that has a dicom preamble and it is incorrectly rejected.

### Expected behavior

File should be accepted.

### Actual behavior

The `readNoPreambleDicom` implements a heuristic to see if the file starts with valid group/element tags.

https://github.com/InsightSoftwareConsortium/ITK/blame/master/Modules/IO/DCMTK/src/itkDCMTKImageIO.cxx#L231

But it fails when there is a preamble (which is often the case) there are 128 bytes and then the letters DICM.

https://dicom.nema.org/medical/dicom/current/output/html/part10.html#chapter_7

### Reproducibility

Always

### Versions

It appears to have been introduced by this commit:

https://github.com/InsightSoftwareConsortium/ITK/commit/19fa58feb02bfd45154e811d0e13510787e2adfb

### Environment

All platforms.

### Additional Information

I have a fix and will submit a PR.

pieper:
A previous optimization didn't take into account the premable of a dicom file.  This adds a check.

Fixes #4108

### DIFF
--- a/Modules/IO/DCMTK/src/itkDCMTKImageIO.cxx
+++ b/Modules/IO/DCMTK/src/itkDCMTKImageIO.cxx
@@ -65,7 +65,7 @@ DCMTKImageIO::DCMTKImageIO()
 }
 
 void
-DCMTKImageIO::SetLogLevel(LogLevelEnum level)
+DCMTKImageIO ::SetLogLevel(LogLevelEnum level)
 {
   switch (level)
   {
@@ -91,12 +91,12 @@ DCMTKImageIO::SetLogLevel(LogLevelEnum level)
       OFLog::configure(OFLogger::OFF_LOG_LEVEL);
       break;
     default:
-      itkExceptionMacro("Unknown DCMTK Logging constant " << static_cast<int>(level));
+      itkExceptionMacro(<< "Unknown DCMTK Logging constant " << static_cast<int>(level));
   }
 }
 
 DCMTKImageIO::LogLevelEnum
-DCMTKImageIO::GetLogLevel() const
+DCMTKImageIO ::GetLogLevel() const
 {
   dcmtk::log4cplus::Logger rootLogger = dcmtk::log4cplus::Logger::getRoot();
   switch (rootLogger.getLogLevel())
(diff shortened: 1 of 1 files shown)


pieper:
@hjmjohnson could you have a look at this?

It fixes https://github.com/InsightSoftwareConsortium/ITK/issues/4108

issakomi:
I can confirm that DCMTKIO's _CanReadFile_ function returns **false** for valid DICOM files with preamble. I have reproduced it just now.

jhlegarreta:
Thanks for doing this @pieper. I am willing to give this a :+1: when the following are addressed:
- We're trying to remove/have removed whitespaces before the scope resolution operator, e.g.
https://github.com/InsightSoftwareConsortium/ITK/pull/3664
so please remove them, e.g.
https://github.com/InsightSoftwareConsortium/ITK/pull/4110/files#diff-4f73091b3003136c994a445b6eda56d8ac3e9044936fbb78a0649164a8c31a73R68
- The output stream `<<` insertion operators were removed from `itkExceptionMacro` calls in: https://github.com/InsightSoftwareConsortium/ITK/pull/4057/files, so please remove them:
https://github.com/InsightSoftwareConsortium/ITK/pull/4110/files#diff-4f73091b3003136c994a445b6eda56d8ac3e9044936fbb78a0649164a8c31a73R94
If they should be added back, a separate commit and PR are required.
- Please document the method in the header file:
https://github.com/InsightSoftwareConsortium/ITK/pull/4110/files#diff-4f73091b3003136c994a445b6eda56d8ac3e9044936fbb78a0649164a8c31a73R135
- Adding a test that demonstrates the issue would be required to ensure that we do not run into this issue again.

In general, commits should only fall into one category, `BUG`, `DOC`, `STYLE`, etc. Helps reviewing, keeping changes of different nature separate, keeping a clean history, etc.

pieper:
Thanks for the review @jhlegarreta - I had tried to add just the one function but it looks like my checkout was against an older target.  I'll fix it.

Modules/IO/DCMTK/src/itkDCMTKImageIO.cxx
```diff
@@ -230,7 +259,7 @@ DCMTKImageIO::CanReadFile(const char * filename)
     }
```
thewtex, line 259: This could be removed following seek to 0

thewtex (changes requested):

thewtex:
#4501 has updates with @jhlegarreta requests addressed (note that `readPreampleDicom` is not in the header, so it is moved to an anonymous namespace instead). Also, replaced the file re-open, close with a seek.

thewtex:
Superceded by #4501

PR #4111, opened 2023-07-15
BUG: Get image geometry from GDCM Secondary Capture

### LINKED ISSUE #4109: GDCM ignores image geometry in secondary captures
### Description

Secondary capture images can optionally have ImageOrientationPatient and PixelSpacing fields, but GDCM currently forces the orientation to identity and looks for NominalScannedPixelSpacing instead of PixelSpacing, so otherwise valid images are loaded incorrectly.

There is more information in the discussion at this PR:

https://github.com/Slicer/Slicer/pull/7089

### Steps to Reproduce

The data available from the idc-open-data set on aws is the Visible Human Project male dataset converted to dicom by @dclunie.

You can get one file with this command:

`aws s3 cp --no-sign-request s3://idc-open-data/4aaf9181-fb6a-4a4c-bf49-d1eb9ed4a385/fbc89e1c-29c4-48cf-89fd-53b7c1dc1488.dcm .`

Or the whole dataset using this command (about 13 gigabytes).

`s5cmd --no-sign-request --endpoint-url https://s3.amazonaws.com cp 's3://idc-open-data/4aaf9181-fb6a-4a4c-bf49-d1eb9ed4a385/*' .`

### Expected behavior

The index to physical transform should be correctly populated.

### Actual behavior

The direction cosines are explicitly set of identity:

https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmImageHelper.cxx#L731-L740

The PixelSpacing is ignored, and the NominalScannedPixelSpacing is looked for, but is typically absent.

https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmImageHelper.cxx#L1265-L1273

### Reproducibility

Always

### Versions

Tested on current main branch,  but I suspect the code has been around for decades.

### Environment

Any

### Additional Information

I'll probably fix this, although it may not be a high priority if we fix https://github.com/InsightSoftwareConsortium/ITK/issues/4108 instead.

pieper:
Secondary capture images can have pixel spacing and image orientation, but the GDCM code explicitly ignored these values (whereas DCMTK code read them).

Per David Clunie (@dclunie) the editor of the standard, it was clarified in CP-586 that PixelSpacing is the correct place to store the best known value (e.g. one that has been corrected for distortion) while other fields like the NominalScannedPixelSpacing, which GDCM used before this commit, if stored at all should have the uncorrected values.

For these reasons this commit makes the default behavior of GDCM for SC match what it would do if SC images were MR or CT in terms of handling the image geometry, which is a better default than the previos behavior.

https://dicom.nema.org/medical/dicom/Final/cp586_ft.pdf

Fixes #4109

### DIFF
--- a/Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmImageHelper.cxx
+++ b/Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmImageHelper.cxx
@@ -729,7 +729,7 @@ std::vector<double> ImageHelper::GetDirectionCosinesValue(File const & f)
     }
 
   dircos.resize( 6 );
-  if( ms == MediaStorage::SecondaryCaptureImageStorage || !GetDirectionCosinesFromDataSet(ds, dircos) )
+  if( !GetDirectionCosinesFromDataSet(ds, dircos) )
     {
     dircos[0] = 1;
     dircos[1] = 0;
@@ -1269,7 +1269,8 @@ Tag ImageHelper::GetSpacingTagFromMediaStorage(MediaStorage const &ms)
   case MediaStorage::MultiframeTrueColorSecondaryCaptureImageStorage:
     // See PS 3.3-2008. Table C.8-25 SC IMAGE MODULE ATTRIBUTES
     // and Table C.8-25b SC MULTI-FRAME IMAGE MODULE ATTRIBUTES
-    t = Tag(0x0018,0x2010);
+    //t = Tag(0x0018,0x2010); Clarified in CP-586 2006-01-25 that PixelSpacing is correct
+    t = Tag(0x0028,0x0030);
     break;
   case MediaStorage::HardcopyGrayscaleImageStorage:
   case MediaStorage::HardcopyColorImageStorage:


issakomi:
IMHO, the idea to read IPP/IOP/PixelSpacing from the SecondaryCaptureImageStorage is good, there are many such data sets (they can be considered **Extended** Secondary Capture), but it should be done in GDCM or documented that the changes have to re-applied after every GDCM update.

------------------------------------

_ImagePosition (Patient)_ should be updated here, otherwise it will be 0,0,0 as before:
https://github.com/InsightSoftwareConsortium/ITK/blob/cdbbdf9441dab496196ba6843eb4a1c2d441e449/Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmImageHelper.cxx#L580

------------------------------------

The PR will **not** enable Pixel Spacing _(0x0028,0x0030)_ for `MediaStorage::SecondaryCaptureImageStorage`, actually it can  done by setting `ImageHelper::ForcePixelSpacing` to _true_, s. [here](https://github.com/InsightSoftwareConsortium/ITK/blob/cdbbdf9441dab496196ba6843eb4a1c2d441e449/Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmImageHelper.cxx#L1304C1-L1309C6). Change of `MediaStorage::MultiframeTrueColorSecondaryCaptureImageStorage` spacing is not related to `MediaStorage::SecondaryCaptureImageStorage`.

pieper:
> it should be done in GDCM or documented that the changes have to be re-applied after every GDCM update.

Very good point!  I wasn't sure how ITK interacts with the upstream repository and mostly wanted this change to go into Slicer's fork, but yes, coming up with the best conforming solution in GDCM makes good sense and then it can work its way into ITK and then Slicer.

> Probably it should work if both spacing attributes will be read for _SC Image_ and _(0028,0030)_ will have higher priority, if both are available.

Yes, I interpreted [CP 586](https://dicom.nema.org/medical/dicom/Final/cp586_ft.pdf) as saying that PixelSpacing is the preferred way to express the true value.  It might make sense to have a fallback to look for one of the other spacing tags if PixelSpacing is missing, but it felt like an obscure corner case and I didn't see any easy way to restructure the code since I'm not familiar with it.

I agree with your other points @issakomi  and would be happy to see a better implementation of this that covers more ases.  Let's see if @malaterre or others have suggestions on how best to approach this.

issakomi:
S. https://github.com/malaterre/GDCM/pull/158/files

thewtex:
Superceded by #4521

PR #4115, opened 2023-07-16
DOC: Avoid documenting ivar default values in SpatialFunction class

jhlegarreta:
Avoid documenting ivar default values in
`itk::ConicShellInteriorExteriorSpatialFunction` since:
- All ivar values should be initialized to some default value.
- When default values are changed, it is easy to forget about updating the documentation and making it inconsistent with the actual value in the code.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkConicShellInteriorExteriorSpatialFunction.h
+++ b/Modules/Core/Common/include/itkConicShellInteriorExteriorSpatialFunction.h
@@ -115,7 +115,7 @@ public:
 
   /** Set/Get direction along the gradient to search.
    * Set to true to use the direction that the gradient is pointing;
-   * set to false for the opposite direction. Default is Off. */
+   * set to false for the opposite direction. */
   itkGetConstMacro(Polarity, bool);
   itkSetMacro(Polarity, bool);
   itkBooleanMacro(Polarity);


jhlegarreta:
Not sure there is another way to show ivar default values through the Doxygen documentation, but I do think that we should avoid documenting explicitly for the above reasons.

Cross-referencing: https://github.com/InsightSoftwareConsortium/ITK/pull/4115

N-Dekker (approved):
Well... in general there is _always_ redundancy between documentation and implementation. And one may accidentally forget to keep the documentation consistent with the implementation. But in this particular case, the initialization is itself already included with the doxygen output at https://itk.org/Doxygen/html/classitk_1_1ConicShellInteriorExteriorSpatialFunction.html#aa0bf34461e6adf408abc7a0bf6e995c5 as it says:

    bool [...] m_Polarity { false }

So end users can find the default initial value of `Polarity` from the documentation this way.

Thanks Jon! 👍

jhlegarreta:
Looks like the badges in the `README` and the Azure build status are not in sync. Not sure why. Badges are all green

itk_badges_20230718

but Azure builds are not

itk_azure_20230718

Also, **badges are pointing at different commits across builds**, e.g. [f602b3c1](https://github.com/InsightSoftwareConsortium/ITK/commit/f602b3c1dcf854cb4e29f62a8964bbe3cee71522) for `ITK.Windows`, [3c20912e](https://github.com/InsightSoftwareConsortium/ITK/commit/3c20912e5567d547c71294af2b8afaceee920d64) for `ITK.macOS`.

@thewtex @tbirdso would you be able to investigate this, please? Thanks.

PR #4118, opened 2023-07-17
STYLE: GetGlobalDefaultSplitter replace call_once call with local static

N-Dekker:
Replaced a `call_once` call and two global variables with a single local static variable, to simplify the code, while still being thread-safe.

### DIFF
--- a/Modules/Core/Common/src/itkImageSourceCommon.cxx
+++ b/Modules/Core/Common/src/itkImageSourceCommon.cxx
@@ -18,23 +18,16 @@
 
 #include "itkImageRegionSplitterSlowDimension.h"
 #include "itkImageSourceCommon.h"
-#include <mutex>
 
 namespace itk
 {
 
-namespace
-{
-std::once_flag                   globalDefaultSplitterOnceFlag;
-ImageRegionSplitterBase::Pointer globalDefaultSplitter;
-} // namespace
-
 const ImageRegionSplitterBase *
 ImageSourceCommon::GetGlobalDefaultSplitter()
 {
-  std::call_once(globalDefaultSplitterOnceFlag,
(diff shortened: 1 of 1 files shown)


dzenanz (approved):

blowekamp:
The ITK Python libraries has some requirements for how global variable should be shared and initialized. I believe @thewtex knows this the best, perhaps get a review from him too?

N-Dekker:
> The ITK Python libraries has some requirements for how global variable should be shared and initialized. 

Thanks for your comment @blowekamp FYI I did mark the pull request "ready" anyway, because I don't really think the choice between calling `std::call_once` and using thread-safe local static variable initialization would affect Python library requirements. So I just hope the PR passes the review 😃

blowekamp:
The question I have is not that the new implementation has the same behavior as the old. The questions is if proper behavior for ITK Python should have an implementation for this variable using the custom initialization using the `itkInitGlobalsMacro`.

thewtex (changes requested):
@blowekamp is right -- this is not providing the required functionality for ITK Python, which is not just magic statics.

thewtex:
The Singleton functionality, used by ITK Python (although it could also be used in other contexts), uses a lamba; this style change (removing the lambda) is not compatible.

thewtex:
@N-Dekker I will create a PR to address the global default splitter properly.

thewtex:
Patch here: https://github.com/InsightSoftwareConsortium/ITK/pull/4122

N-Dekker:
Closed, as this pull request is superseded by:
- #4122 commit 0f5b4bd3593c3626c017fa1aa3acba0507d81247 by Matt (@thewtex)

PR #4121, opened 2023-07-21
STYLE: Remove redundant GetPrimaryOutput() calls from `ProcessObject`

N-Dekker:
Let ProcessObject member functions just call `GetPrimaryOutput()` once, and store the pointer in a local variable.

### DIFF
--- a/Modules/Core/Common/src/itkProcessObject.cxx
+++ b/Modules/Core/Common/src/itkProcessObject.cxx
@@ -1177,9 +1177,9 @@ ProcessObject::IncrementProgress(float increment)
 bool
 ProcessObject::GetReleaseDataFlag() const
 {
-  if (this->GetPrimaryOutput())
+  if (const auto primaryOutput = this->GetPrimaryOutput())
   {
-    return this->GetPrimaryOutput()->GetReleaseDataFlag();
+    return primaryOutput->GetReleaseDataFlag();
   }
   return false;
 }
@@ -1281,9 +1281,9 @@ ProcessObject::PrintSelf(std::ostream & os, Indent indent) const
 void
 ProcessObject::Update()
 {
-  if (this->GetPrimaryOutput())
+  if (const auto primaryOutput = this->GetPrimaryOutput())
   {
-    this->GetPrimaryOutput()->Update();
+    primaryOutput->Update();
   }
 }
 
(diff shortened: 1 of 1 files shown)


dzenanz (approved):

Modules/Core/Common/src/itkProcessObject.cxx
```diff
@@ -1177,9 +1177,9 @@ ProcessObject::IncrementProgress(float increment)
 bool
 ProcessObject::GetReleaseDataFlag() const
 {
-  if (this->GetPrimaryOutput())
+  if (const auto primaryOutput = this->GetPrimaryOutput())
```
N-Dekker, line 1180: This could also be written slightly differently, for example:

    if(const DataObject * const primaryOutput{ this->GetPrimaryOutput() })

Using `{ }` instead of `=` might be slightly more readable, to avoid the typical confusion between `=` and `==` in a conditional statement. On the other hand, I like `const auto primaryOutput =` for being short, no-nonsense. What do you think?

blowekamp: Yes, I agree an assignment in an if statement is confusing. My old school styling preference work be to have the variable declaration on a separate line with the initializer.

dzenanz: I agree with Brad, I like the initialization to be before `if`. But this is a suggestion.

N-Dekker: Thanks, but it seems to me that the original author did not want to spend extra lines of code on doing an initialization before the `if`. I guess that's why the original code just does:

    if (this->GetPrimaryOutput())
    {
      this->GetPrimaryOutput()->Update();
    }

My preference is still not to add extra lines of code. I would just like to get rid of the redundant function calls, that's all.

BTW, C++17 has introduced another possible syntax: `if(` _init-statement_ ﻿ `;` _condition_ `)`. In this case, that could look like:

    if (const auto primaryOutput = this->GetPrimaryOutput(); primaryOutput)

But in this particular case, I think we might as well just do:

    if (const auto primaryOutput = this->GetPrimaryOutput())

Which is still what I'm proposing now. Hope I made my point clearer.

PR #4122, opened 2023-07-21
BUG: GetGlobalDefaultSplitter to a global singleton

thewtex:
Ensure a shared instance across binaries via itk::Singleton.

### DIFF
--- a/Modules/Core/Common/src/itkImageSourceCommon.cxx
+++ b/Modules/Core/Common/src/itkImageSourceCommon.cxx
@@ -15,27 +15,29 @@
  *  limitations under the License.
  *
  *=========================================================================*/
-
 #include "itkImageRegionSplitterSlowDimension.h"
 #include "itkImageSourceCommon.h"
+#include "itkSingleton.h"
 #include <mutex>
 
 namespace itk
 {
 
-namespace
+struct ImageSourceCommonGlobals
 {
-std::once_flag                   globalDefaultSplitterOnceFlag;
-ImageRegionSplitterBase::Pointer globalDefaultSplitter;
-} // namespace
+  ImageSourceCommonGlobals() { m_GlobalDefaultSplitter = ImageRegionSplitterSlowDimension::New().GetPointer(); };
+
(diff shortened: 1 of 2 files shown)


blowekamp (approved):
LGTM

dzenanz (approved):

Modules/Core/Common/src/itkImageSourceCommon.cxx
```diff
@@ -15,27 +15,29 @@
...
-namespace
+struct ImageSourceCommonGlobals
 {
-std::once_flag                   globalDefaultSplitterOnceFlag;
-ImageRegionSplitterBase::Pointer globalDefaultSplitter;
-} // namespace
+  ImageSourceCommonGlobals() { m_GlobalDefaultSplitter = ImageRegionSplitterSlowDimension::New().GetPointer(); };
+
+  ImageRegionSplitterBase::Pointer m_GlobalDefaultSplitter;
+};
```
N-Dekker, line 31: Please consider using the C++ Rule of Zero: remove the user-defined default-constructor, and initialize m_GlobalDefaultSplitter at its declaration instead:

    struct ImageSourceCommonGlobals
    {
        ImageRegionSplitterBase::Pointer m_GlobalDefaultSplitter{ ImageRegionSplitterSlowDimension::New().GetPointer() };
    };

thewtex: @N-Dekker thanks for the review and tip. I tried this, but a number of the tests segfault -- the initialization order is important here.

N-Dekker: @thewtex Thanks for considering my suggestion to use the Rule of Zero. I don't really understand how such a code simplification could cause segfaults, do you? Of do you think those segfaults just happen coincidentally?

When having the proposed simplification, the initialization still takes place during the default-construction of the `ImageSourceCommonGlobals` struct. So as far as I can see, there is no reordering of the initialization.

PR #4144, opened 2023-08-08
STYLE: Use the `itkNameOfTestExecutableMacro` macro in tests

jhlegarreta:
Use the `itkNameOfTestExecutableMacro` macro to get the test name instead of hard-coding the test name.

Left behind in 1fa53fec51eee61f8d602505959640d50497838b and in successive related commits.

## PR Checklist

### DIFF
--- a/Modules/Bridge/VtkGlue/test/itkImageToVTKImageFilterRGBTest.cxx
+++ b/Modules/Bridge/VtkGlue/test/itkImageToVTKImageFilterRGBTest.cxx
@@ -20,6 +20,7 @@
 
 #include "itkImageFileReader.h"
 #include "itkRGBPixel.h"
+#include "itkTestingMacros.h"
 
 int
 itkImageToVTKImageFilterRGBTest(int argc, char * argv[])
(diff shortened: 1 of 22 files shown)


jhlegarreta:
There are a few other hits remaining, e.g.
https://github.com/InsightSoftwareConsortium/ITK/blob/0595670b96b3f79149bb422d325567fbeed053fe/Modules/IO/MRC/test/itkMRCImageIOTest.cxx#L287

@N-Dekker do you think we should also adopt the macro for these?

N-Dekker:
> https://github.com/InsightSoftwareConsortium/ITK/blob/0595670b96b3f79149bb422d325567fbeed053fe/Modules/IO/MRC/test/itkMRCImageIOTest.cxx#L287
>
> @N-Dekker do you think we should also adopt the macro for these?

Thanks for asking @jhlegarreta In this specific case, I don't think it happens _in practice_ that `argv[0]` is null. Because at that point, it is already checked that `argc` is greater than `2`. Which should _in practice_ mean that the number of arguments (from `argv`) is greater than two already, so `argv[0]` should be OK.

If you still think that `argv[0]` _might_ be null, it still does not seem very useful to write:

    std::string filePrefix = itkNameOfTestExecutableMacro(argv); // ???

It may do something like `filePrefix = "<" + functionName + " executable>"`. But that may not be a useful value for such a  `filePrefix` anyway.

dzenanz (approved):

N-Dekker (approved):
Thanks Jon! 👍

PR #4152, opened 2023-08-10
STYLE: ImageSourceCommonGlobals default member initializer, Rule of Zero

N-Dekker:
Added an in-class default member initializer to the data member of `ImageSourceCommonGlobals`. Removed its default-constructor, following the C++ "Rule of Zero".

----

Aims to clarify our discussion at https://github.com/InsightSoftwareConsortium/ITK/pull/4122#discussion_r1272163516

### DIFF
--- a/Modules/Core/Common/src/itkImageSourceCommon.cxx
+++ b/Modules/Core/Common/src/itkImageSourceCommon.cxx
@@ -25,9 +25,7 @@ namespace itk
 
 struct ImageSourceCommonGlobals
 {
-  ImageSourceCommonGlobals() { m_GlobalDefaultSplitter = ImageRegionSplitterSlowDimension::New().GetPointer(); };
-
-  ImageRegionSplitterBase::Pointer m_GlobalDefaultSplitter;
+  ImageRegionSplitterBase::Pointer m_GlobalDefaultSplitter{ ImageRegionSplitterSlowDimension::New().GetPointer() };
 };
 
 itkGetGlobalSimpleMacro(ImageSourceCommon, ImageSourceCommonGlobals, PimplGlobals);


dzenanz (approved):

N-Dekker:
@thewtex If you think this style improvement is OK, I'll mark it "Ready". I was hoping to see those segfaults that you mentioned at https://github.com/InsightSoftwareConsortium/ITK/pull/4122#discussion_r1272635505 but they don't seem to appear now. Could it be that you tried something slightly different to get those segfaults?

blowekamp:
The segfaults are likely related to shared library loading of ITK Python. Did you try that?

Does C++ provide any guarantees where the initialization gets executed in the previous implementation and the proposed?

Modules/Core/Common/src/itkImageSourceCommon.cxx
```diff
@@ -25,9 +25,7 @@ namespace itk
 
 struct ImageSourceCommonGlobals
 {
-  ImageSourceCommonGlobals() { m_GlobalDefaultSplitter = ImageRegionSplitterSlowDimension::New().GetPointer(); };
-
-  ImageRegionSplitterBase::Pointer m_GlobalDefaultSplitter;
+  ImageRegionSplitterBase::Pointer m_GlobalDefaultSplitter{ ImageRegionSplitterSlowDimension::New().GetPointer() };
```
N-Dekker, line 28: The original default-constructor of `ImageSourceCommonGlobals` does (implicitly) default-construct the smart pointer `m_GlobalDefaultSplitter` (_null_-initializing its pointer), and then assigns a new value from `ImageRegionSplitterSlowDimension::New()` to the pointer. My proposal does instead directly initialize `m_GlobalDefaultSplitter` to the new value. Semantically both should have the same effect. Unless some other thread tries to access the `ImageSourceCommonGlobals` object while it is still "under construction". Is that happening currently?

N-Dekker:
> The segfaults are likely related to shared library loading of ITK Python. Did you try that?

No, thanks for asking @blowekamp but I see that the CI has passed successfully, including ITK.Linux.Python and ITK.macOS.Python.  
 
> Does C++ provide any guarantees where the initialization gets executed in the previous
> implementation and the proposed?

I'm sorry I don't entirely understand your question. Maybe you can elaborate a bit? But anyway, this proposed style improvement should be just fine, from a C++ point of view. So I don't see how it could possibly trigger segfaults. However, I'm still studying ITK's global/singleton framework now. With the following pull request, I hope to improve stability, reducing the chance of segfaults, by adding the C++ `volatile` keyword:
- #4163

Please have a look  😃

thewtex:
> @thewtex If you think this style improvement is OK, I'll mark it "Ready". I was hoping to see those segfaults that you mentioned at #4122 (comment) but they don't seem to appear now. Could it be that you tried something slightly different to get those segfaults?

I will test with a clean build locally.

thewtex:
> I will test with a clean build locally.

With a clean build, the tests pass. This change is fine with me.

N-Dekker:
> With a clean build, the tests pass. This change is fine with me.

@thewtex Thanks Matt! 👍  I just marked my PR _ready_. Just for my understanding, is your local build fundamentally different with respect to the use of global data, compared to the CI builds (including ITK.Linux.Python and ITK.macOS.Python)? If so, can you please explain the difference?

thewtex:
As @blowekamp pointed out, shared and static builds can make a difference. I tried both locally.

thewtex (approved):

Issue #4153, opened 2023-08-10
ITK does not compile with VS2022 17.7.0


dzenanz:
### Steps to Reproduce

Update Visual Studio to:

Microsoft Visual Studio Community 2022 (64-bit) - Current
Version 17.7.0

### Expected behavior

ITK compiles without errors.

### Actual behavior

```log
214>itkPointGTest.cxx
214>M:\Dashboard\ITK\Modules\Core\Common\include\itkPoint.h(369,45): error C2187: syntax error: 'attribute specifier' was unexpected here
214>M:\Dashboard\ITK\Modules\Core\Common\test\itkPointGTest.cxx(63,57): message : see reference to function template instantiation 'auto itk::MakePoint<int,>(const TValue)' being compiled
214>        with
214>        [
214>            TValue=int
214>        ]
214>M:\Dashboard\ITK\Modules\Core\Common\include\itkPoint.h(367,1): error C2938: 'itk::MakePoint::<lambda_706d090a84c31981802e12f4e9361bdf>::<lambda_typedef_cdecl>' : Failed to specialize alias template
214>M:\Dashboard\ITK\Modules\Core\Common\include\itkPoint.h(367,1): message : see reference to alias template instantiation 'itk::MakePoint::<lambda_706d090a84c31981802e12f4e9361bdf>::<lambda_typedef_cdecl><_T1>' being compiled
214>M:\Dashboard\ITK\Modules\Core\Common\include\itkPoint.h(367,1): error C2938: 'itk::MakePoint::<lambda_706d090a84c31981802e12f4e9361bdf>::<lambda_typedef_vectorcall>' : Failed to specialize alias template
214>M:\Dashboard\ITK\Modules\Core\Common\include\itkPoint.h(367,1): message : see reference to alias template instantiation 'itk::MakePoint::<lambda_706d090a84c31981802e12f4e9361bdf>::<lambda_typedef_vectorcall><_T1>' being compiled
214>M:\Dashboard\ITK\Modules\Core\Common\include\itkPoint.h(367,1): error C2535: 'itk::MakePoint::<lambda_706d090a84c31981802e12f4e9361bdf>::operator unknown-type(void) noexcept const': member function already defined or declared
214>M:\Dashboard\ITK\Modules\Core\Common\include\itkPoint.h(367,1): message : see declaration of 'itk::MakePoint::<lambda_706d090a84c31981802e12f4e9361bdf>::operator unknown-type'
[...]

PR #4154, opened 2023-08-10
COMP: Adjust check that all arguments of MakePoint/Vector have same type

N-Dekker:
The original `assertSameType` lambda's appeared to trigger compile errors from MSVC, as reported by Dženan Zukić at https://github.com/InsightSoftwareConsortium/ITK/issues/4153 "ITK does not compile with VS2022 17.7.0", like:

> Modules\Core\Common\include\itkPoint.h(369,45): error C2187: syntax error: 'attribute specifier' was unexpected here
> Modules\Core\Common\test\itkPointGTest.cxx(63,57): message : see reference to function template instantiation 'auto itk::MakePoint<int,>(const TValue)' being compiled
> Modules\Core\Common\include\itkPoint.h(367,1): error C2938: 'itk::MakePoint::<lambda_706d090a84c31981802e12f4e9361bdf>::<lambda_typedef_cdecl>' : Failed to specialize alias template
> Modules\Core\Common\include\itkPoint.h(367,1): message : see reference to alias template instantiation 'itk::MakePoint::<lambda_706d090a84c31981802e12f4e9361bdf>::<lambda_typedef_cdecl><_T1>' being compiled

This appears to be caused by a compiler bug from MSVC Compiler Version 19.37.32822.

Instead of these `assertSameType` lambda, with this commit, both `MakePoint` and `MakeVector` make use of class template argument deduction (CTAD), as supported by `std::array`. The deduction guide of std::array ensures that all values have the same type.

### DIFF
--- a/Modules/Core/Common/include/itkPoint.h
+++ b/Modules/Core/Common/include/itkPoint.h
@@ -365,17 +365,11 @@ template <typename TValue, typename... TVariadic>
 auto
 MakePoint(const TValue firstValue, const TVariadic... otherValues)
 {
-  // Assert that the other values have the same type as the first value.
-  const auto assertSameType = [](const auto value) {
-    static_assert(std::is_same_v<decltype(value), const TValue>, "Each value must have the same type!");
-    return true;
-  };
-  const bool assertions[] = { true, assertSameType(otherValues)... };
-  (void)assertions;
-  (void)assertSameType;
+  constexpr unsigned int dimension{ 1 + sizeof...(TVariadic) };
 
-  constexpr unsigned int              dimension{ 1 + sizeof...(TVariadic) };
-  const std::array<TValue, dimension> stdArray{ { firstValue, otherValues... } };
+  // Note that the class template argument deduction guide of std::array ensures that all values have the same type.
+  const std::array stdArray{ firstValue, otherValues... };
+
   return Point<TValue, dimension>{ stdArray };
 }
 
--- a/Modules/Core/Common/include/itkVector.h
+++ b/Modules/Core/Common/include/itkVector.h
@@ -331,17 +331,11 @@ template <typename TValue, typename... TVariadic>
 auto
 MakeVector(const TValue firstValue, const TVariadic... otherValues)
 {
-  // Assert that the other values have the same type as the first value.
-  const auto assertSameType = [](const auto value) {
-    static_assert(std::is_same_v<decltype(value), const TValue>, "Each value must have the same type!");
-    return true;
-  };
-  const bool assertions[] = { true, assertSameType(otherValues)... };
-  (void)assertions;
-  (void)assertSameType;
+  constexpr unsigned int dimension{ 1 + sizeof...(TVariadic) };
 
-  constexpr unsigned int              dimension{ 1 + sizeof...(TVariadic) };
-  const std::array<TValue, dimension> stdArray{ { firstValue, otherValues... } };
+  // Note that the class template argument deduction guide of std::array ensures that all values have the same type.
+  const std::array stdArray{ firstValue, otherValues... };
+
   return Vector<TValue, dimension>{ stdArray };
 }
 


dzenanz (commented):
The code looks more elegant now. Did you test it with updated Visual Studio?

N-Dekker:
> The code looks more elegant now.

I think so too. Still I realize that the use of CTAD is somewhat tricky, even for `std::array`, which does have a deduction guide. I just did a [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/37dc2359a997a6e097c6501ea5108e505bca94e8..059ca0bf5cb208b05afd5f44a656efa46850150b), because I realized it might be risky for the dimension of the created `Point` or `Vector` to depend on the CTAD of `std::array`. Should be fine now! 

> Did you test it with updated Visual Studio?

Sure 😇

dzenanz (approved):

Modules/Core/Common/include/itkPoint.h
```diff
@@ -365,17 +365,11 @@ template <typename TValue, typename... TVariadic>
...
-  const bool assertions[] = { true, assertSameType(otherValues)... };
-  (void)assertions;
-  (void)assertSameType;
-
-  constexpr unsigned int              dimension{ 1 + sizeof...(TVariadic) };
-  const std::array<TValue, dimension> stdArray{ { firstValue, otherValues... } };
+  constexpr unsigned int dimension{ 1 + sizeof...(TVariadic) };
+
+  // Note that the class template argument deduction guide of std::array ensures that all values have the same type.
+  const std::array stdArray{ firstValue, otherValues... };
```
N-Dekker, line 371: For the record, class template argument deduction (CTAD) becomes kinda tricky here when `MakePoint` is called with a single argument which is itself an `std::array`. In that case, `otherValues...` is empty and the line `const std::array stdArray{ firstValue, otherValues... }` does a copy-construction!

Eventually I think everything will be fine anyway, as long as the dimension of the `Point` is not deduced by CTAD of the `std::array`! So in that case, `dimension` = 1, even when the deduced std::array has a size greater than one.

dzenanz: If you think it might be a problem in the future, you could make improvements. But probably first wait for the next nightly dashboards to confirm this fixes the compile issue.

N-Dekker: Thanks @dzenanz, no I don't see a problem with the merged commit. The _possible_ CTAD problem that I thought of should be prevented by the [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/37dc2359a997a6e097c6501ea5108e505bca94e8..059ca0bf5cb208b05afd5f44a656efa46850150b) that I did _before_ you merged.  😃

issakomi: It seems to be related:
https://open.cdash.org/viewBuildError.php?buildid=8907762

`The CXX compiler identification is AppleClang 10.0.0.10001145`

N-Dekker: Thanks @issakomi I see:

```
/Users/builder/external[Modules/Core/Common/include/itkPoint.h:371](https://github.com/InsightSoftwareConsortium/ITK/blob/89d3c3c0158806ab44a33d3c76b436537864398e/Modules/Core/Common/include/itkPoint.h#L371):20: error: no viable constructor or deduction guide for deduction of template arguments of 'array'
  const std::array stdArray{ firstValue, otherValues... };
                   ^
[CTest: warning matched] /Users/builder/external[Modules/Core/Common/test/itkPointGTest.cxx:63](https://github.com/InsightSoftwareConsortium/ITK/blob/89d3c3c0158806ab44a33d3c76b436537864398e/Modules/Core/Common/test/itkPointGTest.cxx#L63):46: note: in instantiation of function template specialization 'itk::MakePoint<int>' requested here
  static_assert(std::is_same_v<decltype(itk::MakePoint(0))::ValueType, int> &&
```


Looks similar to this AppleClang 10.0.0.10001044 issue reported by @andrewliao11: https://github.com/deepmind/open_spiel/issues/490#issuecomment-766549664

Not sure yet how to work around that one, do you have a suggestion?

issakomi: Not yet sure.
I don't have a better idea as a not elegant `#if #else` for Visual Studio and the rest. It seems that we have to workaround 2 bugs, one in VS 17.7 and another in Apple Clang 10.0
 
```
#include <array>

template<typename T, typename... Tv>
void f(T x, Tv... v)
{
#ifdef _MSC_VER
  const std::array a{x, v...};
#else
  const auto assertSameType = [](const auto value)
  {
    static_assert(std::is_same_v<decltype(value), const T>, "Each value must have the same type!");
    return true;
  };
  const bool assertions[] = { true, assertSameType(v)... };
  const std::array<T, sizeof...(Tv) + 1> a{x, v...};
#endif
}

int main()
{
  f(2, 0.1, 0.2); // must fail
  // f(42, 42, 42); // OK
  return 0;
}
```
[test 1](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:3,endLineNumber:11,positionColumn:3,positionLineNumber:11,selectionStartColumn:3,selectionStartLineNumber:11,startColumn:3,startLineNumber:11),source:'%23include+%3Carray%3E%0A%0Atemplate%3Ctypename+T,+typename...+Tv%3E%0Avoid+f(T+x,+Tv...+v)%0A%7B%0A++const+std::array+a%7Bx,+v...%7D%3B%0A%7D%0A%0Aint+main()%0A%7B%0A++f(2,+0.1,+0.2)%3B+//+must+fail%0A++//f(42,+42,+42)%3B+//+OK%0A++return+0%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:42.927110342840685,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:vcpp_v19_35_x64,deviceViewOpen:'1',filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'/std:c%2B%2B17',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x64+msvc+v19.35+(Editor+%231)',t:'0')),k:28.853837035975385,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+13.2',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x64+msvc+v19.35+(Compiler+%231)',t:'0')),k:28.219052621183923,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

[test 2](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:3,endLineNumber:18,positionColumn:3,positionLineNumber:18,selectionStartColumn:3,selectionStartLineNumber:18,startColumn:3,startLineNumber:18),source:'%23include+%3Carray%3E%0A%0Atemplate%3Ctypename+T,+typename...+Tv%3E%0Avoid+f(T+x,+Tv...+v)%0A%7B%0A++const+auto+assertSameType+%3D+%5B%5D(const+auto+value)%0A++%7B%0A++++static_assert(std::is_same_v%3Cdecltype(value),+const+T%3E,+%22Each+value+must+have+the+same+type!!%22)%3B%0A++++return+true%3B%0A++%7D%3B%0A++const+bool+assertions%5B%5D+%3D+%7B+true,+assertSameType(v)...+%7D%3B%0A++const+std::array%3CT,+sizeof...(Tv)+%2B+1%3E+a%7Bx,+v...%7D%3B%0A%7D%0A%0Aint+main()%0A%7B%0A++//f(2,+0.1,+0.2)%3B+//+must+fail%0A++f(42,+42,+42)%3B+//+OK%0A++return+0%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:42.927110342840685,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:g132,deviceViewOpen:'1',filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-std%3Dc%2B%2B17',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+13.2+(Editor+%231)',t:'0')),k:28.853837035975385,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+13.2',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+13.2+(Compiler+%231)',t:'0')),k:28.219052621183923,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

N-Dekker: Thanks, @issakomi Yet another attempt without non-elegant `#ifdef #else`, though... please have a look:
- #4157

issakomi: Thank you, great! Unfortunately I can not test _Apple Clang 10.0.0_ (macOS 10.13), it seems to have partial (or buggy) support for C++17, macOS High Sierra is not supported since Dec 2020, BTW. Clang 10 from _godbolt_ seems to be different and doesn't have the issue. Probably we have to commit and wait for dash-board results. If there will be a problem, maybe `#ifndef __APPLE__` could be better with the comment that it can be dropped together with _Apple Clang 10.0.0_ later. The actual solution in _master_ (to forward the check to _std::array_ deduction) is, IMHO, the best. It is pity that macOS 10.13 doesn't work well with.

dzenanz: Perhaps @seanm could test this?

N-Dekker: @dzenanz wrote:

> Perhaps @seanm could test this?

Hi Sean! Can you possibly have a look? It would already be helpful if you could just try out the following two lines of code on AppleClang 10.0.0 (macOS 10.13), when having enabled C++17:

    #include <array>
    std::array x{1, 2, 3};

It does compile at https://godbolt.org/z/W9cvd9x9a (clang 5.0.0). It should be able to automatically deduce the template arguments of `std::array` (`int` and `3`, in this case). But does it compile on AppleClang 10.0.0 (using `-std=c++17`)?

issakomi: This 

`# Custom compiler flags.`
`set(RR_FLAGS_CXX "${RR_FLAGS_COMMON} -std=c++14 -stdlib=libc++")`

https://open.cdash.org/build/8907762/notes#note0

looks strange.

BTW, e.g. Mac14.x-AppleClang-dbg-Universal set
`set(RR_FLAGS_CXX "${RR_FLAGS_COMMON} -std=c++20")`

Maybe we should ignore macOS builds with very special flags?

N-Dekker: The `-std=c++14` flag looks wrong indeed. Thanks Mihail! @seanm Sean, can you please make it `-std=c++17`, in /Users/builder/external/_scripts/builder22/ITK-AppleClang-rel-x86_64.cmake ?

PR #4157, opened 2023-08-14
COMP: Use conjunction_v to check argument types of MakePoint, MakeVector

N-Dekker:
The previous approach, using the class template argument deduction guide of std::array appeared to trigger compile errors from  AppleClang 10.0.0.10001044, as Mihail Isakov (@issakomi) remarked at https://github.com/InsightSoftwareConsortium/ITK/pull/4154#discussion_r1292840598, like:

> Modules/Core/Common/include/itkPoint.h:371:20: error: no viable constructor or deduction guide for deduction of template arguments of 'array'
>   const std::array stdArray{ firstValue, otherValues... };
>                   ^
> [CTest: warning matched] Modules/Core/Common/test/itkPointGTest.cxx:63:46: note: in instantiation of function template specialization 'itk::MakePoint<int>' requested here
>   static_assert(std::is_same_v<decltype(itk::MakePoint(0))::ValueType, int> &&

---

The proposed approach is very much inspired by the `std::conjunction` example at https://en.cppreference.com/w/cpp/types/conjunction#Example specifically:

```
template<typename T, typename... Ts>
constexpr bool all_types_are_same = std::conjunction_v<std::is_same<T, Ts>...>;
 
static_assert(all_types_are_same<int, int, int>);

```

### DIFF
--- a/Modules/Core/Common/include/itkPoint.h
+++ b/Modules/Core/Common/include/itkPoint.h
@@ -365,11 +365,11 @@ template <typename TValue, typename... TVariadic>
 auto
 MakePoint(const TValue firstValue, const TVariadic... otherValues)
 {
-  constexpr unsigned int dimension{ 1 + sizeof...(TVariadic) };
+  static_assert(std::conjunction_v<std::is_same<TVariadic, TValue>...>,
+                "The other values should have the same type as the first value.");
 
-  // Note that the class template argument deduction guide of std::array ensures that all values have the same type.
-  const std::array stdArray{ firstValue, otherValues... };
-
+  constexpr unsigned int              dimension{ 1 + sizeof...(TVariadic) };
+  const std::array<TValue, dimension> stdArray{ { firstValue, otherValues... } };
   return Point<TValue, dimension>{ stdArray };
 }
 
--- a/Modules/Core/Common/include/itkVector.h
+++ b/Modules/Core/Common/include/itkVector.h
@@ -331,11 +331,11 @@ template <typename TValue, typename... TVariadic>
 auto
 MakeVector(const TValue firstValue, const TVariadic... otherValues)
 {
-  constexpr unsigned int dimension{ 1 + sizeof...(TVariadic) };
+  static_assert(std::conjunction_v<std::is_same<TVariadic, TValue>...>,
+                "The other values should have the same type as the first value.");
 
-  // Note that the class template argument deduction guide of std::array ensures that all values have the same type.
-  const std::array stdArray{ firstValue, otherValues... };
-
+  constexpr unsigned int              dimension{ 1 + sizeof...(TVariadic) };
+  const std::array<TValue, dimension> stdArray{ { firstValue, otherValues... } };
   return Vector<TValue, dimension>{ stdArray };
 }
 


Modules/Core/Common/include/itkPoint.h
```diff
@@ -365,11 +365,11 @@ template <typename TValue, typename... TVariadic>
...
 {
-  constexpr unsigned int dimension{ 1 + sizeof...(TVariadic) };
-
-  // Note that the class template argument deduction guide of std::array ensures that all values have the same type.
-  const std::array stdArray{ firstValue, otherValues... };
+  static_assert(std::conjunction_v<std::is_same<TVariadic, TValue>...>,
+                "The other values should have the same type as the first value.");
 
+  constexpr unsigned int              dimension{ 1 + sizeof...(TVariadic) };
+  const std::array<TValue, dimension> stdArray{ { firstValue, otherValues... } };
```
issakomi, line 372: AFAIK, `{{` `}}` are not required since C++14, 
`double-braces required in C++11 prior to the CWG 1270 revision (not needed in C++11 after the revision and in C++14 and beyond)`
(from https://en.cppreference.com/w/cpp/container/array, S. comment in the "Example")

N-Dekker: Thanks @issakomi Indeed the double-braces are not required anymore here. But they are certainly still allowed. It seems to me that they emphasize two things:
  - The initialization is not a copy-construction
  - The declaration does not use class template argument deduction (CTAD)

The original line of code did use CTAD _and_ did allow copy-construction (if `firstValue` would itself happen be an std::array, and there would be no _other values_). Which would only work with a single pair of braces:

    const std::array stdArray{ firstValue, otherValues... };

So the double-braces are still somewhat meaningful. What do you think?

Modules/Core/Common/include/itkPoint.h
```diff
@@ -365,11 +365,11 @@ template <typename TValue, typename... TVariadic>
 auto
 MakePoint(const TValue firstValue, const TVariadic... otherValues)
 {
-  constexpr unsigned int dimension{ 1 + sizeof...(TVariadic) };
-
-  // Note that the class template argument deduction guide of std::array ensures that all values have the same type.
-  const std::array stdArray{ firstValue, otherValues... };
+  static_assert(std::conjunction_v<std::is_same<TVariadic, TValue>...>,
```
N-Dekker, line 368: Instead of `std::conjunction_v`, we could also use a C++17 [_fold expression_](https://en.cppreference.com/w/cpp/language/fold), for example:

    static_assert((std::is_same_v<TVariadic, TValue> && ...));

Modules/Core/Common/include/itkPoint.h
N-Dekker: @issakomi In general, "wrapping" and copy-construction may be confused more easily when using single braces:

    int x = 1;
    std::array<int, 1> a{x}; // "wraps" x as array element
    std::array<int, 1> b{a}; // copy-construction

When using double-braces, copy-construction won't compile:

    std::array<int, 1> c{ {x} };  // OK, also wraps" x as array element
    std::array<int, 1> d{ {a} };  // Error: copy-construction fails to compile

See https://godbolt.org/z/GvcWq1Mx1

Having said so, I can understand if people find the double-braces a bit (too) noisy.

issakomi: Thank you! Sure, but, IMHO, in _particular case_ (template) it would be array or arrays, not a copy-construction, double/single braces do the same, s. [test](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:30,endLineNumber:11,positionColumn:30,positionLineNumber:11,selectionStartColumn:30,selectionStartLineNumber:11,startColumn:30,startLineNumber:11),source:'%23include+%3Carray%3E%0A%0Atemplate%3Ctypename+T,+typename...+Tv%3E%0Avoid+f(T+x,+Tv...+v)%0A%7B%0A++const+std::array%3CT,+sizeof...(Tv)+%2B+1%3E+a%7B+%7Bx,+v...%7D+%7D%3B%0A%7D%0A%0Aint+main()%0A%7B%0A++std::array%3Cint,+2%3E+a%7B42,+42%7D%3B%0A++f(a,+a,+a)%3B%0A++return+0%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:42.927110342840685,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:g132,deviceViewOpen:'1',filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-std%3Dc%2B%2B17',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+13.2+(Editor+%231)',t:'0')),k:28.853837035975385,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+13.2',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+13.2+(Compiler+%231)',t:'0')),k:28.219052621183923,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4).

N-Dekker: FYI, A new Clang compiler warning, `-Wctad-selects-copy`, is being proposed to avoid accidentally doing a copy-construction when _wrapping_ was intended, while using CTAD: [[Clang][Sema] Add -Wctad-selects-copy to diagnose copy deduction candidate](https://reviews.llvm.org/D155475) by PoYao Chang (@poyaoc97).

This new Clang warning could serve as an alternative to the existing `-Wctad-maybe-unsupported`, which produced undeserved warnings when using `std::lock_guard` or `std::scoped_lock`, as I reported at
- https://github.com/llvm/llvm-project/issues/63288

N-Dekker:
It would be nice if we could get clearer why the AppleClang compile errors that are addressed by this PR occurred in the first place. Does AppleClang 10.0.0/macOS 10.13 reject _any_ attempt to do class template argument deduction (CTAD) for `std::array`? For example:

```cpp
#include <array>
std::array x{1, 2, 3};  // What does AppleClang 10.0.0 say? 

``` 
As I wrote at https://github.com/InsightSoftwareConsortium/ITK/pull/4154#discussion_r1298765302

Mihail (@issakomi) noticed before that Mac10.13-AppleClang-rel-x86_64 still seems to add a `-std=c++14` compiler flag: https://github.com/InsightSoftwareConsortium/ITK/pull/4154#discussion_r1298921875 Shouldn't that be `-std=c++17`? Would that solve the compile errors? I hope Sean (@seanm) can still have a look!

dzenanz:
@seanm Could you try `-std=c++17`, as Niels suggests in https://github.com/InsightSoftwareConsortium/ITK/pull/4157#issuecomment-1693014264?

dzenanz:
Ping @seanm.

seanm:
I've confirmed the compiler accepts the `-std=c++17` flag, and made the change to my script.  We'll see tonight...

seanm:
So the errors are still there: https://open.cdash.org/build/9062420/notes

We see the `-std=c++17` flag there now. Either it didn't work, or the way I'm passing `-std=c++17` is itself incorrect...

dzenanz:
Should we merge this now?

N-Dekker:
@seanm @dzenanz Thanks for your feedback

> Should we merge this now?

I would still like to know if the example at https://github.com/InsightSoftwareConsortium/ITK/pull/4157#issuecomment-1693014264 fails on AppleClang 10.0.0/macOS 10.13, when C++17 is enabled. Sean, can you still tell us?  

    #include <array>
    std::array x{1, 2, 3};  // What does AppleClang 10.0.0 say?

seanm:
```
builder22:Desktop builder$ cat test.cxx
#include <array>

int main ()
{
    std::array x{1, 2, 3};  // What does AppleClang 10.0.0 say?
    return 0;
}
```

```
builder22:Desktop builder$ xcrun clang --version
Apple LLVM version 10.0.0 (clang-1000.11.45.5)
Target: x86_64-apple-darwin17.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

```
builder22:Desktop builder$ xcrun clang -Weverything -fsyntax-only -std=c++17 test.cxx
test.cxx:5:5: warning: 'auto' type specifier is incompatible with C++98 [-Wc++98-compat]
    std::array x{1, 2, 3};  // What does AppleClang 10.0.0 say?
    ^~~
test.cxx:5:17: warning: generalized initializer lists are incompatible with C++98 [-Wc++98-compat]
    std::array x{1, 2, 3};  // What does AppleClang 10.0.0 say?
                ^
test.cxx:5:16: error: no viable constructor or deduction guide for deduction of template arguments of 'array'
    std::array x{1, 2, 3};  // What does AppleClang 10.0.0 say?
               ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/__tuple:223:64: note: candidate function template not viable: requires 0
      arguments, but 3 were provided
template <class _Tp, size_t _Size> struct _LIBCPP_TEMPLATE_VIS array;
                                                               ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/__tuple:223:64: note: candidate function template not viable: requires 1
      argument, but 3 were provided
2 warnings and 1 error generated.
```

N-Dekker:
Thanks @seanm I think we may ignore the warnings about being "incompatible with C++98". But the error is problemetic:

> error: no viable constructor or deduction guide for deduction of template arguments of 'array'

I guess this means that we cannot yet write such simple code, like `std::array x{1, 2, 3}`, as long as we need to support that particular platform. Because the included Standard C++ Library implementation does not yet have a deduction guide for `std::array`.

N-Dekker:
I still like `static_assert((std::is_same_v<TVariadic, TValue> && ...))` slightly better than using `conjunction_v`, as I said at https://github.com/InsightSoftwareConsortium/ITK/pull/4157#discussion_r1293658527 but that's just a matter of style. If this PR (using `conjunction_v`) solves those compile errors, I'm fine with having it merged as it is now.

dzenanz:
@seanm does this PR solve the compile errors?

seanm:
> @seanm does this PR solve the compile errors?

It does.  I vote to merge.

seanm (approved):

PR #4159, opened 2023-08-17
STYLE: Remove `ITK_X_ASSERT`, just use `assert` in `constexpr` functions

N-Dekker:
From C++14, the standard `assert` supports being called directly from inside a `constexpr` function, as was already mentioned in Andrzej's C++ blog entry https://akrzemi1.wordpress.com/2017/05/18/asserts-in-constexpr-functions The `ITK_X_ASSERT` workaround for GCC 4 is no longer needed, as ITK now requires at least GCC 7 (and also at least C++17).

This reverts commit 9ff8f9fcdef345a11a47320253d96438a60d5859 "COMP: Fix C++11 compatibility with assert in constexpr" by Bradley Lowekamp (@blowekamp), June 25, 2018.

### DIFF
--- a/Modules/Core/Common/include/itkConnectedImageNeighborhoodShape.h
+++ b/Modules/Core/Common/include/itkConnectedImageNeighborhoodShape.h
@@ -27,16 +27,6 @@
 #include <cstdint> // For uintmax_t
 #include <limits>
 
-// C++11 does not guarantee that assert can be used in constexpr
-// functions. This is a work-around for GCC 4.8, 4.9. Originating
-// from Andrzej's C++ blog:
-//  https://akrzemi1.wordpress.com/2017/05/18/asserts-in-constexpr-functions/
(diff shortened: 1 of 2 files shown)


dzenanz (approved):

blowekamp:
LGTM

N-Dekker:
Thanks for your approvals so far. FYI [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/39ede8758ba57f9d68410bf1ccb3e3bbff1e1083..e888f7bc2d1def98f80ad5e012f779004a2d1cd5) aims to fix compile errors at the Linux machines that appear introduced by the original PR, saying:

> error: there are no arguments to ‘assert’ that depend on a template parameter, so a declaration of ‘assert’ must be available [-fpermissive]

thewtex (approved):

PR #4163, opened 2023-08-20
STYLE: Use `volatile` to avoid global SingletonIndex being optimized out

N-Dekker:
Passing `initializedGlobalSingletonIndex` to a dummy function (`Unused`) may not always avoid being optimized out. Instead, the C++ `volatile` keyword is really meant to be used to avoid such optimization.

Added the C++17 `[[maybe_unused]]` attribute to avoid possible warnings.

Removed the claim that `Unused` ensures that its parameter is not optimized out.

### DIFF
--- a/Modules/Core/Common/include/itkSingleton.h
+++ b/Modules/Core/Common/include/itkSingleton.h
@@ -27,8 +27,7 @@
  *
  * This function is to be used to mark parameters as unused to suppress
  * compiler warning. It can be used when the parameter needs to be named
- * (i.e. itkNotUsed cannot be used) but is not always used. It ensures
- * that the parameter is not optimized out.
+ * (i.e. itkNotUsed cannot be used) but is not always used.
  */
 template <typename T>
 inline void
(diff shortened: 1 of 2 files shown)


N-Dekker:
Thanks for your approval @dzenanz I just had a look at the effect of calling `Unused(globalVariable)`, at https://godbolt.org/z/j1acbvheE (code below here), but I don't really see that it would prevent a global variable from being optimized out. (Looking at the generated assembly code, I see no effect, on an optimized compilation.)  Maybe it did before, on some specific compiler versions...?  Anyway, `volatile` does the job 😃    

```cpp
template <typename T>
inline void
Unused(const T &){};

namespace
{
  int Fun() { return 42; };
  int globalVariable = Fun();
}

int main()
{
  Unused(globalVariable);
}

```

N-Dekker:
FYI, C++20 has deprecated various uses of `volatile` (original proposal: [Deprecating volatile, JF Bastien, 2018-10-01](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1152r0.html)), but I just double-checked, the use case of this PR will _not_ be deprecated  👍

PR #118, opened 2018-11-06
ENH: Synchronize factories across modules in Python

hjmjohnson:
Replaces individual synchronization mechanisms across modules for each global
variables used in ITK by a central mechanism in which all these variables
are registered and which gives a single entry point to synchronize
all global variables.

Global variables are registered in a map, with a key that is their name and
a value that is a pair of their pointer and a pointer to a synchronization
function. The synchronization function will be called to synchronize
these global variables across the different modules. Different global
variables may have different synchronization functions: Some variables
will just override the previous value in all modules whereas some
others will need to actually synchronize their values.

Change-Id: Ib6ebc9b3f9acd93ce8319f00579f6bf4982ca6b8

PR #4164, opened 2023-08-20
STYLE: SingletonIndex does not need to store the unused `func` parameter

N-Dekker:
No longer stored the `func` parameter from `SingletonIndex::SetGlobalInstance`, and from `Singleton(globalName, func, deleteFunc)`, as it was never actually retrieved from `m_GlobalObjects` afterwards.

Deprecated the `SingletonIndex::SingletonData` type alias, as it is no longer used internally anymore, and it should not be public either.

----

Note that even with the initial commit a66337ec215c88f4900a2caf419b055483c42085 (pull request #118, Francois Budin (@fbudin69500) and Hans Johnson (@hjmjohnson), February 2019), the value of the `func` parameter was never retrieved from `SingletonIndex::m_GlobalObjects`, looking at:
https://github.com/InsightSoftwareConsortium/ITK/blob/a66337ec215c88f4900a2caf419b055483c42085/Modules/Core/Common/src/itkSingleton.cxx

https://github.com/InsightSoftwareConsortium/ITK/blob/a66337ec215c88f4900a2caf419b055483c42085/Modules/Core/Common/include/itkSingleton.h

### DIFF
--- a/Modules/Core/Common/include/itkSingleton.h
+++ b/Modules/Core/Common/include/itkSingleton.h
@@ -47,7 +47,12 @@ class ITKCommon_EXPORT SingletonIndex
 public:
   /** Standard class types. */
   using Self = SingletonIndex;
-  using SingletonData = std::map<std::string, std::tuple<void *, std::function<void(void *)>, std::function<void()>>>;
+
+#ifndef ITK_LEGACY_REMOVE
+  using SingletonData [[deprecated("The internal representation of the singleton data is private, and may not "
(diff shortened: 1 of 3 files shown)


N-Dekker:
Still "draft", just because I would like to have it processed _after_ another pull request #4162, because of possible merge conflicts between them.

dzenanz (approved):

hjmjohnson (approved):

PR #4168, opened 2023-08-23
STYLE: Replace `std::unique_lock` with `std::lock_guard` in ThreadPool

N-Dekker:
A `std::lock_guard` is more lightweight than a `std::unique_lock`, so for simple use cases, `std::lock_guard` appears preferable.

----

Note that `std::lock_guard` is already used extensively in within ITK (I counted 79 times), so this proposed change would be consistent with the existing code  😃

### DIFF
--- a/Modules/Core/Common/include/itkThreadPool.h
+++ b/Modules/Core/Common/include/itkThreadPool.h
@@ -93,7 +93,7 @@ auto result = pool->AddWork([](int param) { return param; }, 7);
 
     std::future<return_type> res = task->get_future();
     {
-      std::unique_lock<std::mutex> lock(this->GetMutex());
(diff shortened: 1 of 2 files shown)


dzenanz (approved):

N-Dekker:
@thewtex, @dzenanz Thanks for your approvals! Of course, this is really just a small style improvement, but I did notice that a `lock_guard` object is indeed smaller than a `unique_lock` 😃  At least at compile-time! At https://godbolt.org/z/xWc6vPsT3 as follows:

```cpp
#include <mutex>
static_assert(sizeof(std::lock_guard<std::mutex>) < sizeof(std::unique_lock<std::mutex>));
```

PR #4169, opened 2023-08-23
Pygccxml update to v2.4.0

thewtex:


dzenanz (approved):
Looks good on a glance. What is the main motivation for the update?

iMichka:
I think it's a good practice to be in sync to avoid having too much a gap the version ITK ships and pygccxml upstream. Doing version updates from time to time in smaller steps is probably better than massive updates each 5 years :).

Pygccxml adds official support for Python 3.11, 3.10, 3.9.0 (though not much has changed for pygccxml, so not a big change for ITK), and drops support for Python 3.6 and 3.5 (which are EOL).

I will make a few more updates to pygccxml (I have a few small fixes in mind but nothing that should concern ITK directly). But I do not plan to put too much energy into it as I have other projects to maintain that use more bandwith right now.

PR #4170, opened 2023-08-23
BUG: Fix xarray compute value check

thewtex:
The input are random integers from 0-255 -- a minimum of 1.0 will only occur occasionally.


tbirdso (approved):

dzenanz (approved):

Wrapping/Generators/Python/Tests/test_xarray.py
```diff
@@ -124,7 +124,7 @@
     data_array = itk.xarray_from_image(image, view=True)
     assert data_array.name == image.GetObjectName()
     # verify we can run a computation on xarray data
-    assert float(data_array.min().compute()) == 1.0
+    assert float(data_array.min().compute()) > -1.0
```
N-Dekker, line 127: Would the following still be OK?

    assert 0.0 <= float(data_array.min().compute()) <= 255.0 

Just curious... I just learned that Python allows an `a <= x <= b` syntax to denote "in-between" 😃

hjmjohnson: This seems wrong to me.  Why convert the integer to a float and then use integer based comparision?

```python
assert float(data_array.min().compute()) => 0.0

or 

assert int(data_array.min().compute()) => -1
```

thewtex: @hjmjohnson @N-Dekker will add both suggestions, thanks

jhlegarreta (approved):
Looks good to me. Thanks for doing this Matt.

N-Dekker (approved):
Thanks, Matt!

Wrapping/Generators/Python/Tests/test_xarray.py
```diff
@@ -124,7 +124,7 @@
     data_array = itk.xarray_from_image(image, view=True)
     assert data_array.name == image.GetObjectName()
     # verify we can run a computation on xarray data
-    assert float(data_array.min().compute()) == 1.0
+    assert 0 <= data_array.min().compute() <= 255
```
hjmjohnson, line 127: This smells much better :).

PR #4171, opened 2023-08-23
WIP (Not ready): STYLE: Replace once_flag of ThreadPoolGlobals with static local variable

N-Dekker:
A static local `bool` variable is more _lightweight_ than a `std::once_flag` data member, while its (static) initialization just as thread-safe as `std::call_once`.

### DIFF
--- a/Modules/Core/Common/src/itkThreadPool.cxx
+++ b/Modules/Core/Common/src/itkThreadPool.cxx
@@ -40,9 +40,6 @@ struct ThreadPoolGlobals
   // To lock on the various internal variables.
   std::mutex m_Mutex;
 
-  // To allow singleton creation of ThreadPool.
-  std::once_flag m_ThreadPoolOnceFlag;
-
   // The singleton instance of ThreadPool.
   ThreadPool::Pointer m_ThreadPoolInstance;
 
@@ -75,8 +72,8 @@ ThreadPool::GetInstance()
   // initialized.
   itkInitGlobalsMacro(PimplGlobals);
 
-  // Create a singleton ThreadPool.
-  std::call_once(m_PimplGlobals->m_ThreadPoolOnceFlag, []() {
+  // Create a ThreadPool singleton as part of the initialization of a static local variable, to ensure thread-safety.
+  [[maybe_unused]] static const volatile bool isCalledOnce = [] {
     m_PimplGlobals->m_ThreadPoolInstance = ObjectFactory<Self>::Create();
     if (m_PimplGlobals->m_ThreadPoolInstance.IsNull())
     {
@@ -85,7 +82,8 @@ ThreadPool::GetInstance()
 #if defined(ITK_USE_PTHREADS)
     pthread_atfork(ThreadPool::PrepareForFork, ThreadPool::ResumeFromFork, ThreadPool::ResumeFromFork);
 #endif
-  });
+    return true;
+  }();
 
   return m_PimplGlobals->m_ThreadPoolInstance;
 }
@@ -107,7 +105,7 @@ ThreadPool::SetDoNotWaitForThreads(bool doNotWaitForThreads)
 ThreadPool::ThreadPool()
 {
   // m_PimplGlobals->m_Mutex not needed to be acquired here because construction only occurs via GetInstance which is
-  // protected by call_once.
+  // protected by the thread-safe initialization of a static local variable, `isCalledOnce`.
 
   m_PimplGlobals->m_ThreadPoolInstance = this;        // threads need this
   m_PimplGlobals->m_ThreadPoolInstance->UnRegister(); // Remove extra reference


dzenanz (commented):
Code looks good, but it should be tested, preferably with Python multiprocessing via forking.

N-Dekker:
Wow, segfaults and timeouts!

https://open.cdash.org/test/1187815471 says:

> Test: PythonGetNameOfClass (Failed)
> Build: Linux-Build8919-PR4171-Replace-ThreadPoolGlobals-ThreadPoolOnceFlag-Python[ (Azure.fv-az71-143)](https://open.cdash.org/viewSite.php?siteid=118663) 
> ...
> Loading ITKVTK... done
> itkTestDriver: Process exception: Segmentation fault

https://open.cdash.org/test/1187815689 says:

> Test: PythonTemplateTest (Failed)
> Build: Linux-Build8919-PR4171-Replace-ThreadPoolGlobals-ThreadPoolOnceFlag-Python[ (Azure.fv-az71-143) ](https://open.cdash.org/viewSite.php?siteid=118663)on 2023-08-23 16:04:40
> Repository revision: 2e8ad9f6533ea6ea6063737c19250595af40a0fe
> Test Details: Completed (Timeout)

https://open.cdash.org/test/1187868304 says:

> Test: PythonExtrasTest (Failed)
> Build: Darwin-Build8992-PR4171-Replace-ThreadPoolGlobals-ThreadPoolOnceFlag-Python[ (Azure.Mac-1692806138896) ](https://open.cdash.org/viewSite.php?siteid=118666)on 2023-08-23 16:02:47
> Repository revision: 2e8ad9f6533ea6ea6063737c19250595af40a0fe
> Test Details: Completed (Failed)
...
> Traceback (most recent call last):
>  File "/Users/runner/work/1/s/Wrapping/Generators/Python/Tests/extras.py", line 328, in <module>
>    itk.transformwrite(transforms, sys.argv[7], compression=True)
>  File "/Users/runner/work/1/s-build/Wrapping/Generators/Python/itk/support/extras.py", line 1446, in transformwrite
    writer.Update()
> RuntimeError: thread::join failed: Invalid argument
> libc++abi: terminating with uncaught exception of type std::__1::system_error: thread::join failed: Invalid argument
> itkTestDriver: Process exception: Subprocess aborted

To be continued...

N-Dekker:
Does anyone have an explanation about what goes wrong here, on both ITK.Linux.Python and ITK.macOS.Python? Is there a fundamental reason why a static local variable initialization in an ITK core CXX file might not be thread-safe, when doing a Python wrapping? And if so, why would `std::call_once` then be any better?

dzenanz:
`std::call_once` might have some special treatment, while an ordinary local static variable does not.

dzenanz:
Static variables get constructed at initialization time, while `std::call_once` gets executed at run time.

blowekamp:
> Static variables get constructed at initialization time, while `std::call_once` gets executed at run time.

This introduces a race condition with other global static variable and their initializations. The `itk::ObjectFactory` depends on static initialization so it cannot be used during static initialization, so ITK objects derived from itk::Object can not be used during static initialization.

N-Dekker:
Do ITK.Linux.Python and ITK.macOS.Python possibly load the ITKCommon lib multiple times concurrently, getting multiple instances of ITKCommon at the same time?

blowekamp:
"Race condition" is not precise. It is more of a dependence on the order of the static initialization across the modules that is problematic and unspecified.

N-Dekker:
Sorry, I still don't get it! Looking at the code of this PR, the static local variable only gets initialized when ThreadPool::GetInstance() is called for the very first time. Which ensures that the lambda code only gets executed when ThreadPool::GetInstance() is called for the very first time. From C++11, static local variable initialization is officially thread-safe. So this should do exactly the same as the original `std::call_once` call: (It _should_ do the same, but it does not.)
 
https://github.com/InsightSoftwareConsortium/ITK/blob/2e8ad9f6533ea6ea6063737c19250595af40a0fe/Modules/Core/Common/src/itkThreadPool.cxx#L69-L89

dzenanz:
Maybe compiler does some special optimizations for local static variables, which it doesn't do for `std::call_once`?

issakomi:
> Maybe compiler does some special optimizations for local static variables, which it doesn't do for std::call_once?

AFAIK, it is possible that compiler with eg `-O3` can optimize away unused variable and doesn't run the lambda at all.

[test with -03](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:22,endLineNumber:4,positionColumn:22,positionLineNumber:4,selectionStartColumn:22,selectionStartLineNumber:4,startColumn:22,startLineNumber:4),source:'//+Type+your+code+here,+or+load+an+example.%0Avoid+f()%0A%7B%0A++++%5B%5Bmaybe_unused%5D%5D+static+bool+x+%3D+%5B%5D%7B+return+true%3B+%7D()%3B%0A%7D%0A%0Aint+main()%0A%7B%0A++++f()%3B%0A++++return+0%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:g132,deviceViewOpen:'1',filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-O3',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+13.2+(Editor+%231)',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

```
void f()
{
    [[maybe_unused]] static bool x = []{ return true; }();
}

int main()
{
    f();
    return 0;
}
```

```
f():
        ret
main:
        xor     eax, eax
        ret
```

P.S. I fact, it does not prove what exactly happens in the PR. Maybe something else.

issakomi:
> the static local variable only gets initialized when ThreadPool::GetInstance() is called for the very first time. Which ensures that the lambda code only gets executed when ThreadPool::GetInstance() is called for the very first time. 

Small demonstration that the above statement is correct

```cpp
#include <iostream>
#include <chrono>
#include <thread>

void f()
{
	static unsigned long long t0 = []()
	{
		return
			std::chrono::duration_cast<std::chrono::milliseconds>
				(std::chrono::system_clock::now().time_since_epoch()).count();
	}();
	std::cout << t0 << std::endl;
}

int main(int, char **)
{
	std::cout <<
		std::chrono::duration_cast<std::chrono::milliseconds>
			(std::chrono::system_clock::now().time_since_epoch()).count()
		<< std::endl;

	std::this_thread::sleep_for(std::chrono::milliseconds(3333));

	f();

	std::this_thread::sleep_for(std::chrono::milliseconds(1111));

	f();

	std::this_thread::sleep_for(std::chrono::milliseconds(1111));

	f();

	return 0;
}
```

```
r@deb2:~$ g++ s.cpp 
r@deb2:~$ ./a.out 
1692981873633
1692981876967
1692981876967
1692981876967
```

N-Dekker:
Thanks for having a look, @issakomi ! 

> maybe it would be better to leave the code as it is (once_flag), the reward is not high, IMHO?

This particular PR is just a style improvement. If it works, I think a static local variable is preferable to a public once_flag member in a global (file scope) data structure and a `call_once` call. But only if it works, of course  😺  

Moreover, I think it's worth knowing _in general_ whether we can safely assume that static local variable initialization is thread-safe. And if so, how to use it properly!

> AFAIK, it is possible that compiler with eg `-O3` can optimize away unused variable and doesn't run the lambda at all.

The compiler is only allowed to do optimizations that do not change the run-time behavior. (According to [the as-if rule](https://en.cppreference.com/w/cpp/language/as_if).)  So if the lambda doesn't do anything observable, it may indeed be optimized away. But if the lambda has an observable side effect, it _cannot_ be optimized away, e.g. (https://godbolt.org/z/7PPKe3Er4):

```cpp
[[maybe_unused]] static bool x = []{ std::cout << 42; return true; }();`
```

> the value is assigned (maybe re-initialized is correct term?) at the first run of the function.

Initialization is the correct term 😃  If a static local variable is initialized by a user-defined constructor, you may see that this constructor is called only once, by putting a print statement inside the constructor. 

But as both of you (@dzenanz @issakomi )  mention compiler optimization as a possible cause of the CI failures, I can still try to see what happens of optimization is suppressed by adding a `volatile` keyword. Please wait...

issakomi:
> But if the lambda has an observable side effect, it cannot be optimized away, e.g. (https://godbolt.org/z/7PPKe3Er4)

I run the above test with and without -O3 here is a lot of difference, specially with lambda. The f() in your example is not reduced to just 'ret', but reduced to printing '42'. Strictly, there is _no_ lambda (s. assembler).

> If it works, I think a static local variable is preferable to a public once_flag

IMHO, I still think it is a little bit a hack, to replace `call_once` with  initialization of a _unused_ local static variable, for the compiler it may be not clear what do we want.

issakomi:
> optimization is suppressed by adding a volatile keyword.

Maybe my 1st example is oversimplified, but volatile doesn't change anything

[test](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:38,endLineNumber:4,positionColumn:38,positionLineNumber:4,selectionStartColumn:38,selectionStartLineNumber:4,startColumn:38,startLineNumber:4),source:'//+Type+your+code+here,+or+load+an+example.%0Avoid+f()%0A%7B%0A++++%5B%5Bmaybe_unused%5D%5D+static+volatile+bool+x+%3D+%5B%5D%7B+return+true%3B+%7D()%3B%0A%7D%0A%0Aint+main()%0A%7B%0A++++f()%3B%0A++++return+0%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:g132,deviceViewOpen:'1',filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-O3',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+13.2+(Editor+%231)',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

AFAIK, `volatile` prevents _some_ optimization steps, related to read / write.

N-Dekker:
> IMHO, I still think it is a little bit a hack, to replace `call_once` with  initialization of a _unused_ local static variable

If you really _want_ to call `call_once`, I think that at least _in theory_, it might be preferable to keep the `once_flag` as a local variable, instead of having it as a `public` data member of globally shared data:

```cpp
ThreadPool::Pointer
ThreadPool::GetInstance()
{
  itkInitGlobalsMacro(PimplGlobals);

  static std::once_flag localOnceFlag{};

  std::call_once(localOnceFlag, []() {
    m_PimplGlobals->m_ThreadPoolInstance = ObjectFactory<Self>::Create();
    // ... some more things here...
  });

  return m_PimplGlobals->m_ThreadPoolInstance;
}
```  

Of course, it would then still depend on proper (thread-safe) support of static local variables...  🤔

issakomi:
> instead of having it as a public data member of globally shared data

Maybe make it _private_? In fact, I am still not 100% sure that the problem is the optimization, seems to be possible, but maybe something else.

N-Dekker:
> tried, it works (with `friend class ThreadPool;`).

Could you then make _all_ data members of PimplGlobals private? Would that compile? If so, please feel free to make it a pull request  😃

issakomi:
This (everything is _private_, except the ctor) works:

```cpp
struct ThreadPoolGlobals
{
  ThreadPoolGlobals() = default;

  friend class ThreadPool;

private:

  // To lock on the various internal variables.
  std::mutex m_Mutex;

  // To allow singleton creation of ThreadPool.
  std::once_flag m_ThreadPoolOnceFlag;

  // The singleton instance of ThreadPool.
  ThreadPool::Pointer m_ThreadPoolInstance;

#if defined(_WIN32) && defined(ITKCommon_EXPORTS)
  // ThreadPool's destructor is called during DllMain's DLL_PROCESS_DETACH.
  // Because ITKCommon-5.X.dll is usually being detached due to process termination,
  // lpvReserved is non-NULL meaning that "all threads in the process
  // except the current thread either have exited already or have been
  // explicitly terminated by a call to the ExitProcess function".
  // Therefore we must not wait for the condition_variable.
  std::atomic<bool> m_WaitForThreads{ false };
#else // In a static library, we have to wait.
  std::atomic<bool> m_WaitForThreads{ true };
#endif
};
```

The ctor has to be public, otherwise there is the error:

```
In file included from /home/r/itk/ITK/Modules/Core/Common/src/itkThreadPool.cxx:25:
/home/r/itk/ITK/Modules/Core/Common/include/itkSingleton.h: In instantiation of ‘T* itk::Singleton(const char*, std::function<void(void*)>, std::function<void()>) [with T = ThreadPoolGlobals]’:
/home/r/itk/ITK/Modules/Core/Common/src/itkThreadPool.cxx:66:1:   required from here
/home/r/itk/ITK/Modules/Core/Common/include/itkSingleton.h:121:16: error: ‘constexpr itk::ThreadPoolGlobals::ThreadPoolGlobals()’ is private within this context
  121 |     instance = new T;
      |                ^~~~~
/home/r/itk/ITK/Modules/Core/Common/src/itkThreadPool.cxx:42:3: note: declared private here
   42 |   ThreadPoolGlobals() = default;
```

I didn't try Python wrapping and test, probable they should work, the `ThreadPoolGlobals` struct in in itkThreadPool.cxx file, nobody else see it.

hjmjohnson:
@N-Dekker your own hypothesis in this thread was right: in a static build the Python modules each carry their own copy of ITKCommon, and that is exactly why the CI segfaults appeared. I reproduced it and opened #6705, which keeps your goal (drop `std::once_flag`) while leaving the guard where it has to live.

The question you asked — *"So this should do exactly the same as the original `std::call_once` call... why would `std::call_once` then be any better?"* — has a concrete answer, and it is not about optimization or static-initialization order.

<details>
<summary>Why <code>call_once</code> is better here: it is the guard's <em>storage</em>, not its semantics</summary>

`std::call_once` is not doing anything special. What matters is *where the flag lives*:

- `m_ThreadPoolOnceFlag` was a member of `ThreadPoolGlobals`, which is registered in the `SingletonIndex` and therefore **shared by every copy of ITKCommon in the process**.
- A function-local `static` has **per-binary storage**. Nothing can make it shared.

In a `BUILD_SHARED_LIBS=OFF` + `ITK_WRAP_PYTHON=ON` build, every `_ITK*Python` module statically links its own ITKCommon and then adopts the one shared index — `Wrapping/macro_files/itk_end_wrap_module.cmake` emits `itk::SingletonIndex::SetInstance(_ITKCommonPython_GetGlobalSingletonIndex())` for exactly that reason. So `isCalledOnce` is `false` again in the second module, the initializer re-runs against the *shared* globals, and you get a second `ThreadPool` plus a second `pthread_atfork` registration.

Your static local is perfectly thread-safe. It is just once-per-*module* rather than once-per-*process*. @blowekamp was pointing at the same thing with "dependence on the order of the static initialization across the modules".

</details>

<details>
<summary>Reproduction on current main</summary>

Fresh static + wrapped build of `main` with this PR's change applied, then a real `import itk`, two filters from two different wrapped modules, and `os.fork()`:

```
terminate called after throwing an instance of 'std::system_error'
  what():  Invalid argument
```

```
#11 itk::ThreadPool::CleanUp () from .../itk/_ITKSmoothingPython.abi3.so
#12 __run_prefork_handlers () at ./posix/register-atfork.c:141
#13 __libc_fork () at ./posix/fork.c:51
```

The failing prefork handler belongs to a *different* module's ITKCommon copy — the duplicate `pthread_atfork`. Unpatched `main` forks cleanly. Pinning a reference the way a live `PoolMultiThreader` does shows two distinct pools and worker threads going 5 → 9.

Worth noting why this hid so well: without a pinned reference the two `ThreadPool` pointers compare **equal**, because the allocator immediately reuses the block of the pool that the re-entered initializer just destroyed. Pointer comparison alone says "looks fine".

</details>

<details>
<summary>What #6705 does</summary>

- Keeps the guard inside `ThreadPoolGlobals`, but replaces the 4-byte `std::once_flag` with a 1-byte atomic guard plus the `std::mutex` already in that struct — so your size goal is met, and the warm path is an inlined atomic load instead of a call into `pthread_once`/`InitOnceExecuteOnce`.
- `std::atomic_flag` where C++20's `test()` exists, `std::atomic<bool>` otherwise, keyed off `__cpp_lib_atomic_flag_test` rather than `__cplusplus` (MSVC misreports the latter).
- Adds `itkThreadPoolSingletonSharingTest`: two `MODULE` libraries, each with its own static ITKCommon, sharing one `SingletonIndex`, asserting they see the same pool. It **fails on this PR's commit and passes on main**, which is the check that was missing in 2023.

On the `volatile` discussion with @issakomi — it is not needed once the guard is a real atomic that is actually read.

</details>

Happy to close this in favour of #6705, or to keep it open if you would rather take the atomic-guard approach here under your own authorship — your call, it was your goal and your diagnosis.

N-Dekker:
A rebase of this simple pull request still triggers CI test failures, even after (almost) 3 years:

- [ITK.Arm64 / ARMBUILD-Python (pull_request)](https://github.com/InsightSoftwareConsortium/ITK/actions/runs/31408046932/job/93519077744?pr=4171)
- [ITK.Linux.Python](https://dev.azure.com/itkrobotlinuxpython/59377e31-8b02-46ea-9d49-32b6bb4b2da6/_build/results?buildId=16820)
- [Linux-Build16820-PR4171-Replace-ThreadPoolGlobals-ThreadPoolOnceFlag-Python](https://open.cdash.org/builds/11480896)

Both PythonTemplateTest and PythonModuleToModuleTest have a timeout, according to https://open.cdash.org/builds/11480896/tests

These failures _might_ be caused by the way, loading, unloading, and reloading modules is supported by Python, as was described by Petr Viktorin (@encukou) at https://github.com/python/cpython/issues/78490#issuecomment-1093795735  already in 2018, referring to [PEP-489](https://peps.python.org/489) (Multi-phase extension module initialization). I was hoping that this old Python problem would be resolved by now, but apparently it is not  🤷 

So for now I give up, and close the pull request without merging. Maybe after 3 more years, we might give it another try  😄

PR #4172, opened 2023-08-23
STYLE: Replace std::unique_lock with std::lock_guard in Filter.hxx files

N-Dekker:
Removed explicit `unlock()` calls. Instead, just made sure that the `lock_guard` variables go out of scope as soon as other threads are allowed to merge data.

Follow-up to pull request https://github.com/InsightSoftwareConsortium/ITK/pull/4168 commit 1094cdde9ae725ad257d5d77308fffaa3dd6e0fe
"STYLE: Replace `std::unique_lock` with `std::lock_guard` in ThreadPool"

----

When reviewing, you might want to suppress whitespace changes: https://github.com/InsightSoftwareConsortium/ITK/pull/4172/files?w=1

### DIFF
--- a/Modules/Filtering/ImageStatistics/include/itkLabelOverlapMeasuresImageFilter.hxx
+++ b/Modules/Filtering/ImageStatistics/include/itkLabelOverlapMeasuresImageFilter.hxx
@@ -120,28 +120,23 @@ LabelOverlapMeasuresImageFilter<TLabelImage>::ThreadedStreamedGenerateData(const
   // local copy, this thread may do multiple merges.
   while (true)
   {
-
+    MapType tomerge{};
     {
-      std::unique_lock<std::mutex> lock(m_Mutex);
+      const std::lock_guard<std::mutex> lockGuard(m_Mutex);
 
       if (m_LabelSetMeasures.empty())
       {
         swap(m_LabelSetMeasures, localStatistics);
         break;
       }
-      else
-      {
-        // copy the output map to thread local storage
-        MapType tomerge;
-        swap(m_LabelSetMeasures, tomerge);
 
-        // allow other threads to merge data
-        lock.unlock();
+      // Move the data of the output map to the local `tomerge` and clear the output map.
+      swap(m_LabelSetMeasures, tomerge);
 
-        // Merge tomerge into localStatistics, locally
-        MergeMap(localStatistics, tomerge);
-      }
-    } // release lock
+    } // release lock, allow other threads to merge data
+
+    // Merge tomerge into localStatistics, locally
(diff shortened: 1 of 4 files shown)


Modules/Filtering/ImageStatistics/include/itkLabelOverlapMeasuresImageFilter.hxx
```diff
@@ -120,28 +120,23 @@ LabelOverlapMeasuresImageFilter<TLabelImage>::ThreadedStreamedGenerateData(const
...
+      const std::lock_guard<std::mutex> lockGuard(m_Mutex);
 
       if (m_LabelSetMeasures.empty())
       {
         swap(m_LabelSetMeasures, localStatistics);
         break;
       }
-      else
-      {
-        // copy the output map to thread local storage
```
N-Dekker, line 134: Note that this PR proposes to remove the term "thread local storage" here, as I think that that term has a specific meaning (https://en.wikipedia.org/wiki/Thread-local_storage) that is not entirely applicable to the local `tomerge` variable in this case.

N-Dekker:
@blowekamp Please have a look 😃 Note that this PR does _not_ fundamentally change the merge process that you introduced with 9c69fc7901421e3c47b47ed20c80f4c55b446303, 09943be533a660c3bddae463e955ce11c255f8e7, 234bbf38430595c49cbb195adaab6874e8d385b2, and 19fab83d6746cc27db35d22de2cd2429db1c8136. It just replaces std::unique_lock with std::lock_guard, and makes sure that the lock guards go out of scope as soon as other threads are allowed to access the data.

N-Dekker:
@blowekamp Thanks for your approval! Just out of curiosity, did you compare your concurrent map merging approach with a more naive approach that just does the merging directly to the output data member, while holding the lock? Looking at: 

https://github.com/InsightSoftwareConsortium/ITK/blob/3335e6943379d09cc8732d8a1b9b8ce064cadd96/Modules/Filtering/ImageStatistics/include/itkLabelOverlapMeasuresImageFilter.hxx#L119-L140

I see that it allows multiple threads to call `MergeMap` simultaneously, but then only one of them will be able to pass the result of the merge (`localStatistics`) to the output (`m_LabelSetMeasures`). The other threads have to just try it again, grabbing the data from `m_LabelSetMeasures` again and doing another merge. Right?

The naive approach I was thinking of just locks once for each thread, and merges directly into the output (`m_LabelSetMeasures`) while holding the lock:  

```cpp
  const std::lock_guard<std::mutex> lockGuard(m_Mutex);
  MergeMap(m_LabelSetMeasures, localStatistics);
  return;

``` 

Wouldn't that be at least as fast?

PR #4179, opened 2023-08-31
STYLE: Improve miscellaneous tests' style

jhlegarreta:
- STYLE: Declare variables close to where they are used in tests
- STYLE: Declare and use test image dimensionality as `constexpr`
- STYLE: Do not declare a test status variable that is only used once
- ENH: Allow `itkCellInterfaceTest` to run as long as possible
- STYLE: Use exception checking macros in itkVectorExpandImageFilterTest
- STYLE: Prefer setting the input once all ivars have been set in test

## PR Checklist

### DIFF
--- a/Modules/Core/Mesh/test/itkCellInterfaceTest.cxx
+++ b/Modules/Core/Mesh/test/itkCellInterfaceTest.cxx
@@ -142,91 +142,43 @@ TestCellInterface(std::string name, TCell * aCell)
 int
 itkCellInterfaceTest(int, char *[])
 {
-  int status;
+  int status = EXIT_SUCCESS;
 
   using VertexCellType = itk::VertexCell<CellInterfaceType>;
-  status = TestCellInterface("Vertex", new VertexCellType);
-  if (status != 0)
-  {
-    return EXIT_FAILURE;
-  }
+  status |= TestCellInterface("Vertex", new VertexCellType);
 
   using LineCellType = itk::LineCell<CellInterfaceType>;
-  status = TestCellInterface("Line", new LineCellType);
-  if (status != 0)
-  {
-    return EXIT_FAILURE;
-  }
+  status |= TestCellInterface("Line", new LineCellType);
 
   using PolyLineCellType = itk::PolyLineCell<CellInterfaceType>;
-  status = TestCellInterface("PolyLine", new PolyLineCellType);
-  if (status != 0)
-  {
-    return EXIT_FAILURE;
-  }
+  status |= TestCellInterface("PolyLine", new PolyLineCellType);
 
   using PolyLineCellType = itk::PolyLineCell<CellInterfaceType>;
-  status = TestCellInterface("PolyLineCellType with 7 vertices", new PolyLineCellType(7));
-  if (status != 0)
-  {
-    return EXIT_FAILURE;
-  }
+  status |= TestCellInterface("PolyLineCellType with 7 vertices", new PolyLineCellType(7));
 
   using TriangleCellType = itk::TriangleCell<CellInterfaceType>;
-  status = TestCellInterface("Triangle", new TriangleCellType);
-  if (status != 0)
-  {
-    return EXIT_FAILURE;
-  }
+  status |= TestCellInterface("Triangle", new TriangleCellType);
 
   using HexahedronCellType = itk::HexahedronCell<CellInterfaceType>;
-  status = TestCellInterface("HexahedronCell", new HexahedronCellType);
-  if (status != 0)
-  {
-    return EXIT_FAILURE;
-  }
+  status |= TestCellInterface("HexahedronCell", new HexahedronCellType);
 
   using TetrahedronCellType = itk::TetrahedronCell<CellInterfaceType>;
-  status = TestCellInterface("TetrahedronCell", new TetrahedronCellType);
-  if (status != 0)
-  {
-    return EXIT_FAILURE;
-  }
+  status |= TestCellInterface("TetrahedronCell", new TetrahedronCellType);
 
   using QuadraticEdgeCellType = itk::QuadraticEdgeCell<CellInterfaceType>;
-  status = TestCellInterface("QuadraticEdgeCell", new QuadraticEdgeCellType);
-  if (status != 0)
-  {
-    return EXIT_FAILURE;
-  }
+  status |= TestCellInterface("QuadraticEdgeCell", new QuadraticEdgeCellType);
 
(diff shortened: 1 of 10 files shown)


Modules/Filtering/ImageIntensity/test/itkVectorExpandImageFilterTest.cxx
```diff
@@ -272,57 +272,25 @@ itkVectorExpandImageFilterTest(int, char *[])
...
-    expander->SetInterpolator(interpolator);
-    testPassed = true;
-  }
+  std::cout << "Setting Interpolator to nullptr" << std::endl;
+  expander->SetInterpolator(nullptr);
+  ITK_TRY_EXPECT_EXCEPTION(expander->Update());
+
+
+  expander->ResetPipeline();
+  expander->SetInterpolator(interpolator);
```
jhlegarreta, line 291: It is weird that I had to add these two lines after the call to update the filter. The previous commit
https://github.com/InsightSoftwareConsortium/ITK/commit/10b5e6a506d3291438552c966833871530c6dc9a

produced a test failure:
https://open.cdash.org/viewTest.php?onlyfailed&buildid=8953052

It is unclear to me the reason of such failure, despite the test completing its instructions. Even more so this being a seemingly harmless style change. Any thoughts?

Modules/Filtering/ImageIntensity/test/itkVectorExpandImageFilterTest.cxx
```diff
@@ -272,57 +272,25 @@ itkVectorExpandImageFilterTest(int, char *[])
...
 
-  if (!testPassed)
-  {
-    std::cout << "Test failed." << std::endl;
-    return EXIT_FAILURE;
-  }
 
   std::cout << "Test passed." << std::endl;
-  return EXIT_SUCCESS;
+  return testPassed;
```
issakomi, line 295: `testPassed` is type _bool_, _true_ will turn into 1 (EXIT_SUCCESS is 0).
I think here should be `return EXIT_SUCCESS;` and `if (!testPassed)` block at line 275 should **not** be removed.

jhlegarreta: Thanks for the review @issakomi. Changed the variable type in ed49ad5. Not sure if I see why the block at line 275 should be removed. The subsequent statements only check exceptions (checking exceptions first, and thus, prior to that block in this case, is maybe cleaner in a test).

issakomi: This block should have returned EXIT_FAILURE if _testPassed_ is false. This is no longer applicable. But it would be possible to exit earlier, currently parts of the test runs after _testPassed_ is already false. But it is probably not critical.

Modules/Core/Mesh/test/itkCellInterfaceTest.cxx
```diff
@@ -142,91 +142,43 @@ TestCellInterface(std::string name, TCell * aCell)
 int
 itkCellInterfaceTest(int, char *[])
 {
-  int status;
+  int status = EXIT_FAILURE;
```
issakomi, line 145: This _status_ seems to be useless, if I see it correctly _TestCellInterface_ always return EXIT_SUCCESS. And _status_ is overwritten by each test. It is not a problem, just does not make much sense.

issakomi (approved):

Modules/Filtering/ImageIntensity/test/itkVectorExpandImageFilterTest.cxx
jhlegarreta: `testPassed` is never set to `EXIT_SUCCESS` (or `true` in the previous version) by any check once it has been set to `EXIT_FAILURE` by some check (or `false` in the previous version), so there is no risk of an intermediate check turning the value so that the test may report as if it is passing while some previous intermediate check set the flag to failing.

I do not think having removed this block, for example, https://github.com/InsightSoftwareConsortium/ITK/pull/4179/files#diff-7f8f7813216d20a0bc1de2689e2bf80c4994519b44f533e4a6228937ecb8a49bL266 is harmful in that way either; the subsequent checks are not influenced by the results of the checks up to that block.

But let me know if I'm still wrong.

In line with this change, running the test as long as possible even if intermediate checks fail may allow to identify multiple errors at a time. The current ITK SWG advocates for tests running as long as possible. The macros are a caveat to this, but it is the price to pay for a cleaner, more compact and readable test code under the current ITK infrastructure.

Modules/Core/Mesh/test/itkCellInterfaceTest.cxx
jhlegarreta: I missed that, sorry. I will amend the commit later today. Thanks.

Modules/Filtering/ImageIntensity/test/itkVectorExpandImageFilterTest.cxx
issakomi: In the first commit you have deleted the  block `if (!testPassed) return EXIT_FAILURE` and returned boolen testPassed. It is fixed . But parts of test still run after _testPassed_ is already set to 'failed'. It is not very critical and not important for me, honestly, just FYI.

jhlegarreta: I think I only changed the file in a single commit, right? Maybe you refer to the previous commit I amended following your https://github.com/InsightSoftwareConsortium/ITK/pull/4179#discussion_r1311655077. As for the concern in your https://github.com/InsightSoftwareConsortium/ITK/pull/4179#discussion_r1312955726 that is restated in https://github.com/InsightSoftwareConsortium/ITK/pull/4179#discussion_r1313025211, https://github.com/InsightSoftwareConsortium/ITK/pull/4179#discussion_r1312966812 acknowledges that the checks still run after `testPassed` is set to `EXIT_FAILURE` but provides the rationale for that, and highlights that an `EXIT_SUCCESS` is not changed to an `EXIT_FAILURE` value, which would be worrying.

issakomi: I didn't realize that the intention was to run the test "for as long as possible". Fine.

Modules/Core/Mesh/test/itkCellInterfaceTest.cxx
jhlegarreta: Done in 88858aa.

dzenanz (approved):

PR #4180, opened 2023-08-31
 Use `lock_guard<mutex>` in itkPlatformMultiThreaderWindows, ParallelSparseFieldLevelSetImageFilter, and PDEDeformable Registration

N-Dekker:
Following C++ Core Guidelines, April 13, 2023, [Use RAII, never plain `lock()`/`unlock()`](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#cp20-use-raii-never-plain-lockunlock)

Follow-up to pull request #4173 commit c4c3710d5fad0c1e4590896647d0e69f8c5a6c79 "STYLE: Use `lock_guard<mutex>` in `PlatformMultiThreader` classes"

### DIFF
--- a/Modules/Core/Common/src/itkPlatformMultiThreaderWindows.cxx
+++ b/Modules/Core/Common/src/itkPlatformMultiThreaderWindows.cxx
@@ -147,9 +147,10 @@ PlatformMultiThreader::TerminateThread(ThreadIdType WorkUnitID)
     return;
   }
 
-  m_SpawnedThreadActiveFlagLock[WorkUnitID]->lock();
-  m_SpawnedThreadActiveFlag[WorkUnitID] = 0;
-  m_SpawnedThreadActiveFlagLock[WorkUnitID]->unlock();
+  {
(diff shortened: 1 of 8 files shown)


N-Dekker:
@dzenanz @blowekamp  Thanks for your approval and your 👍   FYI, I found _one more_ possible `lock_guard<mutex>` use case, in `FFTWGlobalConfiguration::GetInstance()`:

https://github.com/InsightSoftwareConsortium/ITK/blob/3eb8fd7cdfbc5ac2d0c2e5e776848a4cbab3d7e1/Modules/Filtering/FFT/src/itkFFTWGlobalConfiguration.cxx#L151-L168 

This one is somewhat more interesting than the ones addressed so far by this PR, because it throws an exception between the `lock()` and the `unlock()` call. I _think_ that's a bug. (It might mean that the mutex would never be unlocked anymore! 🙀) Anyway, I can make that a separate PR.

N-Dekker:
> FYI, I found _one more_ possible `lock_guard<mutex>` use case, in `FFTWGlobalConfiguration::GetInstance()`

  ...
> Anyway, I can make that a separate PR.

Here it is, please have a look:
- #4181 

Both PRs can be processed independently.

PR #4181, opened 2023-09-01
BUG: Use `lock_guard<mutex>` in FFT configuration to deal with exception

N-Dekker:
When calling `lock()` and `unlock()` explicitly, throwing an exception between between those two calls would leave the mutex locked, which appears undesirable. Used `std::lock_guard` to solve this problem.

Following C++ Core Guidelines, April 13, 2023, "Use RAII, never plain `lock()`/`unlock()`" https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#cp20-use-raii-never-plain-lockunlock

Also replaced manual creation of an exception with an `itkGenericExceptionMacro` call.

### DIFF
--- a/Modules/Filtering/FFT/src/itkFFTWGlobalConfiguration.cxx
+++ b/Modules/Filtering/FFT/src/itkFFTWGlobalConfiguration.cxx
@@ -148,7 +148,7 @@ FFTWGlobalConfiguration::GetInstance()
   itkInitGlobalsMacro(PimplGlobals);
   if (!m_PimplGlobals->m_Instance)
   {
-    m_PimplGlobals->m_CreationLock.lock();
+    const std::lock_guard<std::mutex> lockGuard(m_PimplGlobals->m_CreationLock);
     // Need to make sure that during gaining access
     // to the lock that some other thread did not
     // initialize the singleton.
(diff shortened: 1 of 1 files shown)


Modules/Filtering/FFT/src/itkFFTWGlobalConfiguration.cxx
```diff
@@ -157,15 +157,9 @@ FFTWGlobalConfiguration::GetInstance()
       m_PimplGlobals->m_Instance = Self::New();
       if (!m_PimplGlobals->m_Instance)
       {
-        std::ostringstream message;
-        message << "itk::ERROR: "
-                << "FFTWGlobalConfiguration"
-                << " Valid FFTWGlobalConfiguration instance not created";
-        itk::ExceptionObject e_(__FILE__, __LINE__, message.str().c_str(), ITK_LOCATION);
-        throw e_; /* Explicit naming to work around Intel compiler bug.  */
+        itkExceptionMacro("Valid FFTWGlobalConfiguration instance not created");
```
blowekamp, line 160: This is a static method. The `itkExceptionMacro` using `this`, so this should not work if FFTW is enable in CI.

N-Dekker: Oops, I see now, after having enabled `ITK_USE_FFTWD` my local VS2022 build says:

> itkFFTWGlobalConfiguration.cxx(160,9): error C2355: 'this': can only be referenced inside non-static member functions or non-static data member initializers

Thanks @blowekamp Should be fixed with my next force-push!

Modules/Filtering/FFT/src/itkFFTWGlobalConfiguration.cxx
```diff
@@ -157,15 +157,9 @@ FFTWGlobalConfiguration::GetInstance()
       m_PimplGlobals->m_Instance = Self::New();
       if (!m_PimplGlobals->m_Instance)
       {
```
N-Dekker, line 159: Is it really necessary to manually check the result of `Self::New()` here? I mean, can't we just assume that Self::New() _itself_ throws an exception when it fails to create an FFTWGlobalConfiguration instance?

N-Dekker:
Ready for review, ready for merge, I think. Although I still wonder if the whole `if (!m_PimplGlobals->m_Instance)` can't just be removed, assuming that Self::New() throws an exception anyway, if it fails to create an FFTWGlobalConfiguration instance:

https://github.com/InsightSoftwareConsortium/ITK/blob/3a7336873418d9f975b32fe004767d85325a580b/Modules/Filtering/FFT/src/itkFFTWGlobalConfiguration.cxx#L157-L168

dzenanz (approved):

PR #4184, opened 2023-09-02
BUG: Use atomic bool to fix race found by thread sanitizer

seanm:


N-Dekker:
Interesting, thanks for addressing this issue, @seanm Can you please add some more information about the thread sanitizer? Are the results on a dashboard? Does it run regularly, automatically?

I think your proposed `std::atomic<bool>` approach might indeed solve the problem, but it seems to me that the proper fix would be to use `std::once_flag` and `std::call_once`. At least, that's the approach of ThreadPool:

https://github.com/InsightSoftwareConsortium/ITK/blob/3a7336873418d9f975b32fe004767d85325a580b/Modules/Core/Common/src/itkThreadPool.cxx#L44

https://github.com/InsightSoftwareConsortium/ITK/blob/3a7336873418d9f975b32fe004767d85325a580b/Modules/Core/Common/src/itkThreadPool.cxx#L79

@blowekamp ?

Modules/Core/Common/src/itkObjectFactoryBase.cxx
```diff
@@ -237,9 +238,9 @@ ObjectFactoryBase::Initialize()
 {
   itkInitGlobalsMacro(PimplGlobals);
 
-  if (!m_PimplGlobals->m_Initialized)
+  // Atomically set m_Initialized to true. If it was false before, enter the if.
+  if (m_PimplGlobals->m_Initialized.exchange(true) == false)
```
N-Dekker, line 242: I would prefer to keep the `if(!x)` syntax, rather than `if(x == false)`.

Modules/Core/Common/src/itkObjectFactoryBase.cxx
```diff
@@ -122,7 +123,7 @@ class ObjectFactoryBase::ObjectFactoryBasePrivate : public LightObject
 
   FactoryListType m_RegisteredFactories{};
   FactoryListType m_InternalFactories{};
-  bool            m_Initialized{ false };
+  std::atomic<bool> m_Initialized{ false };
   bool            m_StrictVersionChecking{ false };
```
N-Dekker, line 127: I'm sorry (it wasn't my preference) but the ITK formatting rules require the names of consecutively declared member variables to be aligned (assuming a fixed-size font). So the surrounding member variables should have an extra space, in order to pass clang-format checking on the CI.

Modules/Core/Common/src/itkObjectFactoryBase.cxx
dzenanz: The comment before this line explains the logic, so the comparison can have shorter form as Niels suggests.

blowekamp:
I have not looked (again) into the other initialization methods called, but I am assuming ( I recall correctly that) the methods complete the initialization. Is it problematic if other threads access this class before those methods have completed? When initialization is occurring should other thread be blocked until initialization in complete?

Consider if a lock is actually needed to block other threads until initialization is completed.

Modules/Core/Common/src/itkObjectFactoryBase.cxx
seanm: Done.

Modules/Core/Common/src/itkObjectFactoryBase.cxx
seanm: Hmm, OK, done.

The downside of that style is that it pollutes history, and will make the results of git blame less useful. Also makes merge conflicts more likely. Oh well.

seanm:
> Interesting, thanks for addressing this issue, @seanm Can you please add some more information about the thread sanitizer? Are the results on a dashboard? Does it run regularly, automatically?

Thread sanitizer is pretty slow, but I do have another faster bot now, so could look again at automating it for ITK.

In this case, I was running my own code with TSan, and hit this issue. It may or may not occur with some ITK unit test...

dzenanz (approved):

Modules/Core/Common/src/itkObjectFactoryBase.cxx
N-Dekker: > The downside of that style is that it pollutes history, and will make the results of git blame less useful. Also makes merge conflicts more likely. 

I agree, but for the time being, this is the ITK coding style 🤷 Maybe when we're going to upgrade the clang-format version, we might reconsider the formatting rules... I hope!

dzenanz: I also dislike declaration alignment.

N-Dekker (approved):
I believe this PR itself is OK, although we may still need to have another look at the comment by Bradley (@blowekamp): https://github.com/InsightSoftwareConsortium/ITK/pull/4184#issuecomment-1706456356 If necessary, that may be another PR.

Approved 👍

PR #4186, opened 2023-09-02
STYLE: Make non-macro comparison conditional blocks consistent in tests

jhlegarreta:
Make non-macro comparison conditional blocks consistent in tests: adhere to the ITK SWG guidelines:
- Report the messages to `std::cerr`.
- Make the printed messages be consistent.
- Declare and re-use epsilon/tolerance values.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/test/itkBSplineKernelFunctionTest.cxx
+++ b/Modules/Core/Common/test/itkBSplineKernelFunctionTest.cxx
@@ -63,26 +63,30 @@ itkBSplineKernelFunctionTest(int, char *[])
 
 
   // Testing the output of BSplineKernelFunction
-#define TEST_BSPLINE_KERNEL(ORDERNUM)                                                          \
-  {                                                                                            \
-    using FunctionType = itk::BSplineKernelFunction<ORDERNUM>;                                 \
-    auto function = FunctionType::New();                                                       \
-                                                                                               \
-    function->Print(std::cout);                                                                \
-    for (unsigned int j = 0; j < npoints; ++j)                                                 \
-    {                                                                                          \
-      double results = function->Evaluate(x[j]);                                               \
-      /* compare with external results */                                                      \
-      if (itk::Math::abs(results - b##ORDERNUM[j]) > 1e-6)                                     \
-      {                                                                                        \
-        std::cout << "Error with " << ORDERNUM << " order BSplineKernelFunction" << std::endl; \
-        std::cout << "Expected: " << b##ORDERNUM[j] << " but got " << results;                 \
-        std::cout << " at x = " << x[j] << std::endl;                                          \
-        std::cout << "Test failed" << std::endl;                                               \
-        return EXIT_FAILURE;                                                                   \
-      }                                                                                        \
-    }                                                                                          \
-  }                                                                                            \
+#define TEST_BSPLINE_KERNEL(ORDERNUM)                                               \
+  {                                                                                 \
+    using FunctionType = itk::BSplineKernelFunction<ORDERNUM>;                      \
+    auto function = FunctionType::New();                                            \
+                                                                                    \
+    function->Print(std::cout);                                                     \
+    const double epsilon = 1e-6;                                                    \
+    for (unsigned int j = 0; j < npoints; ++j)                                      \
+    {                                                                               \
+      double results = function->Evaluate(x[j]);                                    \
+      /* compare with external results */                                           \
+      if (itk::Math::abs(results - b##ORDERNUM[j]) > epsilon)                       \
+      {                                                                             \
+        std::cerr.precision(static_cast<int>(itk::Math::abs(std::log10(epsilon)))); \
+        std::cerr << "Test failed!" << std::endl;                                   \
+        std::cerr << "Error with " << ORDERNUM << " order BSplineKernelFunction ";  \
+        std::cerr << "at index [" << j << "] " << std::endl;                        \
+        std::cerr << "Expected value " << b##ORDERNUM[j] << std::endl;              \
+        std::cerr << " differs from " << results;                                   \
+        std::cerr << " by more than " << epsilon << std::endl;                      \
+        return EXIT_FAILURE;                                                        \
+      }                                                                             \
+    }                                                                               \
+  }                                                                                 \
   ITK_MACROEND_NOOP_STATEMENT
 
   TEST_BSPLINE_KERNEL(0);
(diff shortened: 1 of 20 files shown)


jhlegarreta:
I now realize that the blocks such as https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/blob/17227f362be0052881fc5e2f5a0657d80d9f188e/SoftwareGuide/Latex/Appendices/CodingStyleGuide.tex#L3596 may be done using a `ITK_TEST_EXPECT_TRUE` or a `ITK_TEST_EXPECT_EQUAL` macro, and this would save typing and all messages would be consistent. Left for a separate PR for anyone having the bandwidth. This PR makes the blocks consistent at least.

Modules/Core/Common/test/itkBSplineKernelFunctionTest.cxx
```diff
@@ -170,17 +181,19 @@ itkBSplineKernelFunctionTest(int, char *[])
     using FunctionType = itk::BSplineKernelFunction<SplineOrder - 1>;
     auto function = FunctionType::New();
 
-    for (double xx = -3.0; xx <= 3.0; xx += 0.1)
+    for (double xx = -3.0; xx <= 3.0; xx += epsilon)
```
issakomi, line 184: epsilon is 1e-6, not 0.1

Modules/Core/SpatialObjects/test/itkImageSpatialObjectTest.cxx
```diff
@@ -179,12 +187,17 @@ itkImageSpatialObjectTest(int, char *[])
   expectedDerivative[1] = 10;
   expectedDerivative[2] = 100;
   std::cout << "DerivativeAt() with interpolator ...";
-  if (itk::Math::abs(derivative[0] - expectedDerivative[0]) > 0.00001 ||
-      itk::Math::abs(derivative[1] - expectedDerivative[1]) > 0.00001 ||
-      itk::Math::abs(derivative[2] - expectedDerivative[2]) > 0.00001)
+  double epsilon = 0.00001;
```
issakomi, line 190: re-declaration (s. line 168)

jhlegarreta: Thanks. Had already spot that. Addressed in ff46ef4.

Modules/Core/Common/test/itkBSplineKernelFunctionTest.cxx
jhlegarreta: Should have been a few lines downwards. Thanks. Fixed in 542eb7b.

dzenanz:
Rebase to resolve conflicts?

jhlegarreta:
> Rebase to resolve conflicts?

Done in 76d446f.

issakomi:
```cpp
  std::cerr << "Test failed!" << std::endl;
  std::cerr << "Error in parameters at index [" << j << "]" << std::endl;
  std::cerr << "Expected value " << parameters3 << std::endl;
  std::cerr << " differs from " << pdash;
  std::cerr << " by more than " << epsilon << std::endl;
```

Just for the record, it is **not** recommended to put countless `std::endl`, leads to significantly larger code size and slows down execution (additional _flush()_ every time).

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#slio50-avoid-endl

Something like

```cpp
  std::cerr << "Test failed!\nError in parameters at index [" 
            << j << "]\nExpected value " << parameters3
            << "\n differs from " << pdash
            << " by more than " << epsilon << std::endl;
```
 is much more effective

AFAIK, std::cerr flushes always, std::endl can be replaced with '\n', but I personally use **one** std::endl at the end of a message.

It is not a request to change something, for me personally it is low priority.

jhlegarreta:
Thanks for the insight @issakomi. Niels has also raised his discomfort with the use of `std::endl` multiple times; most places that use them in tests could be simplified by the use of some macros (cross-referencing issue #1273). That way we would have a single place to substitute them. I have still this in my backlog. The ITK SWG should be changed accordingly at that point.

For the codebase itself, that should be part of a separate thread and potential PR.

jhlegarreta:
>  I will propose a patch set for the ITK SWG.

After think a little bit about it, and seeing that there is a non-negligible number of hits for `Expected: (...), but got: (...)`, the whole needs more thinking, and probably a macro so that all of this is simplified. So deferring to issue #1273.

issakomi:
> but I personally use one std::endl at the end of a message.

I have experimented a little at _godbolt_, even a single std::endl at the end makes a difference (but probably I will still use it for the time being with std::cout to be sure that a message appears immediately). There are also interesting effects sometimes, -O3 can make code bigger (probably faster?) and other interesting effects. Not trivial. But perhaps trusting C++ guidelines is the right thing to do.
Without optimization 1st and 2nd variants have 77 and 58 lines of assembler output.

issakomi:
> > I will propose a patch set for the ITK SWG.
> 
> After think a little bit about it, and seeing that there is a non-negligible number of hits for `Expected: (...), but got: (...)`, the whole needs more thinking, and probably a macro so that all of this is simplified. So deferring to issue #1273.

I am a little bit afraid that permanent re-writing of tests can introduce bugs, eg in this test there was a typo ... the `for` loop would run 6 million times instead of 60.

But may be for new tests.

jhlegarreta:
> I am a little bit afraid that permanent re-writing of tests can introduce bugs, eg in this test there was a typo ... the for loop would run 6 million times instead of 60.

I agree. Some tests are hard to read, which is also important when trying to increase the coverage, and thus improve the code, and contain code that can be simplified, as said previously. The style varies greatly as well, so the more blocks we can simplify, the less changes we will need to do in the long term, and the less risk we run to inadvertently introduce typos in tests. Also, if that _1e-6_ vs _0.1_ value had been assigned to and used through a variable, I would not have touched at it at all when changing the style. All in all, I am grateful for the reviews and the eyes that look at them.

As a side note, we do get reports for the slowest tests, but we have been lacking the time to use that information effectively to take action on a regular basis (although some contributors have used that information in the past). And as for the codebase,  
There have been efforts to measure and report the performance across versions in the past ([ITKPerformanceBenchmarking](https://github.com/InsightSoftwareConsortium/ITKPerformanceBenchmarking), [ITKPerformancePlot](https://github.com/jhlegarreta/ITKPerformancePlot) and its JS version [ITKPerformanceVisualization](https://github.com/InsightSoftwareConsortium/ITKPerformanceVisualization)), but we have not been able to capitalize the effort yet.

dzenanz (approved):

PR #4190, opened 2023-09-02
BUG: Do not use superclass template parameters in `itk::fem` RTTI

jhlegarreta:
Do not use the superclass template parameters in `itk::fem` classes RTTI macros.

## PR Checklist

### DIFF
--- a/Modules/Numerics/FEM/include/itkFEMSolverCrankNicolson.h
+++ b/Modules/Numerics/FEM/include/itkFEMSolverCrankNicolson.h
@@ -82,7 +82,7 @@ public:
   itkNewMacro(Self);
 
   /** Run-time type information (and related methods) */
-  itkTypeMacro(SolverCrankNicolson, Solver<TDimension>);
+  itkTypeMacro(SolverCrankNicolson, Solver);
 
   using Float = Element::Float;
 
--- a/Modules/Numerics/FEM/include/itkFEMSolverHyperbolic.h
+++ b/Modules/Numerics/FEM/include/itkFEMSolverHyperbolic.h
@@ -49,7 +49,7 @@ public:
   itkNewMacro(Self);
 
   /** Run-time type information (and related methods). */
-  itkTypeMacro(SolverHyperbolic, Solver<TDimension>);
+  itkTypeMacro(SolverHyperbolic, Solver);
 
   using Float = Element::Float;
 


jhlegarreta:
The module seems to be off by default, as there is a test for one of the classes:
https://github.com/InsightSoftwareConsortium/ITK/blob/409b54baa72db0ed2a9f4b4f7fd1eccda509614b/Modules/Numerics/FEM/test/itkFEMSolverHyperbolicTest.cxx#L200

But it is not shown in the dashrboard, e.g.
https://open.cdash.org/viewTest.php?onlypassed&buildid=8958988

Modules/Numerics/FEM/include/itkFEMSolverCrankNicolson.h
```diff
@@ -82,7 +82,7 @@ class ITK_TEMPLATE_EXPORT SolverCrankNicolson : public Solver<TDimension>
   itkNewMacro(Self);
 
   /** Run-time type information (and related methods) */
-  itkTypeMacro(SolverCrankNicolson, Solver<TDimension>);
+  itkTypeMacro(SolverCrankNicolson, Solver);
```
N-Dekker, line 85: Thanks Jon, but I'm wondering, does this code change ever affect the behavior of ITK? Either compile time or run-time? It seems to me that the second argument of `itkTypeMacro` is irrelevant, as it is always ignored anyway.

https://github.com/InsightSoftwareConsortium/ITK/blob/3a7336873418d9f975b32fe004767d85325a580b/Modules/Core/Common/include/itkMacro.h#L437-L439

jhlegarreta: @N-Dekker thanks for the insight. Your assumption seems correct to me. Maybe the with-parentclass-`itkTypeMacro` could be improved to provide the superclass name as well? Even if it is not formally tested, you will agree that we do not put the template parameters in the `itkTypeMacro`. Have not tried to find or counted the appearances, but I bet there is an overwhelming majority that does not put it.

I can change the commit prefix to `STYLE` if that makes more sense.

N-Dekker: > Maybe the with-parentclass-itkTypeMacro could be improved to provide the superclass name as well? 

Maybe, but then it would actually _need_ to have those the template parameters! For now I think it's clearer to keep the `Superclass` type definitions explicit, and outside of the `itkTypeMacro` definition. By the way, I noticed that even the very fist appearance of the `itkTypeMacro` definition did not use the superclass parameter. With commit 699b66cb04d410e555656828e8892107add38ccb Will Schroeder (@wschroed), June 27, 2001: https://github.com/InsightSoftwareConsortium/ITK/blob/699b66cb04d410e555656828e8892107add38ccb/Code/Common/itkMacro.h#L335-L337

So now, more than 22 years later, maybe we can conclude that the second parameter of `itkTypeMacro` just isn't necessary at all. We might then replace it by a simpler new macro that just has a single parameter (`thisClass`). What do you think?

> I can change the commit prefix to STYLE if that makes more sense.

Indeed, I would prefer STYLE, rather than BUG, but I leave it up to you.

jhlegarreta: Sorry Niels, deleted the branch inadvertently. Superseded by PR #4194.

> So now, more than 22 years later, maybe we can conclude that the second parameter of itkTypeMacro just isn't necessary at all. We might then replace it by a simpler new macro that just has a single parameter (thisClass). What do you think?

That is out of the scope of this PR. There is also a parentless macro that takes only the classname, so the above proposal, as it is, would bring confusion:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkMacro.h#L441

> Indeed, I would prefer STYLE, rather than BUG, but I leave it up to you.

Done in PR #4194.

PR #4199, opened 2023-09-03
STYLE: Use an unsigned integer to store the number of poles

jhlegarreta:
Use an unsigned integer instead of a signed to store the number of poles
in `itk::BSplineDecompositionImageFilter`: the number of poles cannot be
negative.

Take advantage of the commit to use `ITK_TEST_EXPECT_EQUAL` in the
corresponding check as an exact comparisons can be performed, ensures
that the lh and rh arguments have the exact same types at compile time,
and since it improves readability, and avoids boilerplate code.

## PR Checklist


N-Dekker (approved):
Thanks Jon!

PR #4209, opened 2023-09-07
STYLE: Remove unused parameter from `SetGlobalInstance` and `Singleton`

N-Dekker:
Added overloads of `SetGlobalInstance` and `Singleton` without the unused `func` parameter. Let the new `SetGlobalInstance` overload just return `void`, instead of `bool`. Deprecated the original overloads.

Follow-up to:

pull request https://github.com/InsightSoftwareConsortium/ITK/pull/4164 commit 6d1c4c7ab7f688bd0a4563bacbe788212a039769
"STYLE: SingletonIndex does not need to store the unused `func` parameter"

pull request https://github.com/InsightSoftwareConsortium/ITK/pull/4162 commit 6ec6328ed2383fb474ace88fbd10c3f1be6bfaba
"STYLE: Let `Singleton` assume that SetGlobalInstance always returns true"

### DIFF
--- a/Modules/Core/Common/include/itkSingleton.h
+++ b/Modules/Core/Common/include/itkSingleton.h
@@ -64,20 +64,29 @@ public:
   }
 
 
-  // Returns true.
-  //
   // It is assumed that the global will remain valid until the start
   // of globals being destroyed.
   template <typename T>
-  bool
+  void
+  SetGlobalInstance(const char * globalName, T * global, std::function<void()> deleteFunc)
+  {
+    this->SetGlobalInstancePrivate(globalName, global, deleteFunc);
+  }
+
+#ifndef ITK_FUTURE_LEGACY_REMOVE
+  template <typename T>
+  [[deprecated("Prefer calling the SetGlobalInstance(globalName, global, deleteFunc) overload (without the unused func "
+               "parameter)!")]] bool
   SetGlobalInstance(const char *                globalName,
                     T *                         global,
                     std::function<void(void *)> itkNotUsed(func),
                     std::function<void()>       deleteFunc)
   {
-    this->SetGlobalInstancePrivate(globalName, global, deleteFunc);
+    this->SetGlobalInstance(globalName, global, deleteFunc);
+    // Just returns true for backward compatibility (legacy only).
     return true;
   }
+#endif
 
   /** Set/Get the pointer to GlobalSingleton.
    * Note that SetGlobalSingleton is not concurrent thread safe. */
@@ -114,7 +123,7 @@ private:
 // A wrapper for a global variable registered in the singleton index.
 template <typename T>
 T *
-Singleton(const char * globalName, std::function<void(void *)> itkNotUsed(func), std::function<void()> deleteFunc)
+Singleton(const char * globalName, std::function<void()> deleteFunc)
 {
   static SingletonIndex * singletonIndex = SingletonIndex::GetInstance();
   Unused(singletonIndex);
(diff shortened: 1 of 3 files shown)


dzenanz (approved):

dzenanz:
We are probably going to do a major version bump soon, so this might be guarded under future legacy, instead of plain legacy.

N-Dekker:
> We are probably going to do a major version bump soon, so this might be guarded under future legacy, instead of plain legacy.

So then, would it still be OK to keep the proposed `[[deprecated]]` attributes?

dzenanz:
Absolutely!

N-Dekker:
@dzenanz Can you do a _Request changes_ to make it `ITK_FUTURE_LEGACY_REMOVE`?  😃

Modules/Core/Common/include/itkSingleton.h
```diff
@@ -114,18 +123,29 @@ class ITKCommon_EXPORT SingletonIndex
...
   {
     instance = new T;
-    SingletonIndex::GetInstance()->SetGlobalInstance<T>(globalName, instance, {}, deleteFunc);
+    SingletonIndex::GetInstance()->SetGlobalInstance<T>(globalName, instance, deleteFunc);
   }
   return instance;
 }
+
+
+#ifndef ITK_LEGACY_REMOVE
```
dzenanz, line 140:
```suggestion
#ifndef ITK_FUTURE_LEGACY_REMOVE
```

Modules/Core/Common/include/itkSingleton.h
```diff
@@ -64,20 +64,29 @@ class ITKCommon_EXPORT SingletonIndex
...
   // of globals being destroyed.
   template <typename T>
-  bool
+  void
+  SetGlobalInstance(const char * globalName, T * global, std::function<void()> deleteFunc)
+  {
+    this->SetGlobalInstancePrivate(globalName, global, deleteFunc);
+  }
+
+#ifndef ITK_LEGACY_REMOVE
```
dzenanz, line 76:
```suggestion
#ifndef ITK_FUTURE_LEGACY_REMOVE
```

dzenanz (commented):
Like this?

Modules/Core/Common/include/itkSingleton.h
N-Dekker: @dzenanz  Honestly my question was just, do you want to change your review status from "Approved" to "Request changes"? Or do you think the current PR is still OK anyway?

dzenanz: I encourage contributions, so I try not to make my nitpicks a barrier for merging. If people want to apply them, even better.

But I definitely misunderstood you here 😕 If a red symbol is what it takes to get the change, I will do it 😄

dzenanz (approved):
Squash upon merge?

N-Dekker:
@dzenanz GitHub automatically made you a co-author of the commits you suggested. Not sure if your co-authorship will be preserved when squashing. 😸 But I would like the commits to be squashed anyway 😄

dzenanz:
Please completely disregard my co-authorship. I prefer a single commit too!

N-Dekker:
I think the PR is ready now, but the CI says:
> **ghostflow-check-master** _Expected — Waiting for status to be reported_ 

Any clue?

dzenanz (approved):

dzenanz:
I overrode it.

PR #4225, opened 2023-09-21
STYLE: Remove private ObjectFactoryBase member `InitializeFactoryList()`

N-Dekker:
`ObjectFactoryBase::InitializeFactoryList()` has become unnecessary from pull request https://github.com/InsightSoftwareConsortium/ITK/pull/3641 commit 48d0cb0eed8a0ceb0f29115302f9972cc691ecb8 "STYLE: Remove pointer indirection from lists of ObjectFactoryBasePrivate"

PR #4226, opened 2023-09-21
BUG: Improve thead safety of ObjectFactoryBase::Initialize

blowekamp:
Block concurent threads until initializaiton is completed.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/src/itkObjectFactoryBase.cxx
+++ b/Modules/Core/Common/src/itkObjectFactoryBase.cxx
@@ -37,7 +37,7 @@
 #include "itksys/SystemTools.hxx"
 #include <cstring>
 #include <algorithm>
-#include <atomic>
+
 
 namespace
 {
@@ -121,10 +121,11 @@ public:
 
   ObjectFactoryBasePrivate() = default;
 
-  FactoryListType   m_RegisteredFactories{};
-  FactoryListType   m_InternalFactories{};
-  std::atomic<bool> m_Initialized{ false };
-  bool              m_StrictVersionChecking{ false };
+  FactoryListType m_RegisteredFactories{};
+  FactoryListType m_InternalFactories{};
+  bool            m_Initialized{ false };
+  std::mutex      m_Lock;
+  bool            m_StrictVersionChecking{ false };
 };
 
 auto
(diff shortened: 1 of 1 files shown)


dzenanz (approved):

Modules/Core/Common/src/itkObjectFactoryBase.cxx
```diff
@@ -237,14 +238,19 @@ ObjectFactoryBase::Initialize()
...
 
-  // Atomically set m_Initialized to true. If it was false before, enter the if.
-  if (!m_PimplGlobals->m_Initialized.exchange(true))
+  if (!m_PimplGlobals->m_Initialized)
   {
-    ObjectFactoryBase::InitializeFactoryList();
-    ObjectFactoryBase::RegisterInternal();
+    // only one thread initializing at a time, wait until initialization is completed.
+    std::lock_guard lockGuard(m_PimplGlobals->m_Lock);
+    if (!m_PimplGlobals->m_Initialized)
```
N-Dekker, line 245: Looks like you're doing _double-checked locking_, which is considered an _anti-pattern_ nowadays! If one thread sets `m_PimplGlobals->m_Initialized = true` while at the same time, another thread tries to evaluate `m_PimplGlobals->m_Initialized` outside the locked region, we have undefined behavior.

Modules/Core/Common/src/itkObjectFactoryBase.cxx
```diff
@@ -237,14 +238,19 @@ ObjectFactoryBase::Initialize()
...
+    {
+      ObjectFactoryBase::InitializeFactoryList();
```
N-Dekker, line 247: Please consider removing the `ObjectFactoryBase::InitializeFactoryList()` call, as in:
- #4225

Modules/Core/Common/src/itkObjectFactoryBase.cxx
blowekamp: What is the updated pattern? To ensure it's only initialized once and block other threads until it it initialed?

dzenanz: std::call_once?

blowekamp: My bigger concern now is if using `std::mutex` during static initialization is OK aka ( defined behavior ).

> evaluate m_PimplGlobals->m_Initialized outside the locked region, we have undefined behavior.

Not sure how undefined the undefined read behavior would be, maybe just not 100% synced between threads. It's will either be true or false, either will be OK.

N-Dekker: > std::call_once?

Thanks, @dzenanz. I just gave it a try, using std::call_once. Please check! (Still draft!)
- #4227

blowekamp:
4227 looks like a better approach.

PR #4227, opened 2023-09-21
WIP: Use call_once for dynamic loading ObjectFactoryBase::Initialize()

N-Dekker:
Removed `ObjectFactoryBasePrivate::m_Initialized`. Moved `RegisterInternal()`
call from `ObjectFactoryBase::Initialize()` to the default-constructor of
`ObjectFactoryBasePrivate`.

----

Also cherry-picked the commit from pull request #4225, _STYLE: Remove private ObjectFactoryBase member InitializeFactoryList()_. (This PR is not meant to supersede PR #4225)

### DIFF
--- a/Modules/Core/Common/include/itkObjectFactoryBase.h
+++ b/Modules/Core/Common/include/itkObjectFactoryBase.h
@@ -226,8 +226,7 @@ public:
     (void)staticFactoryRegistration;
   }
 
-  /** Initialize the static members of ObjectFactoryBase.
-   *  RegisterInternal() and InitializeFactoryList() are called here. */
+  /** Initialize the static members of ObjectFactoryBase. */
   static void
   Initialize();
 
--- a/Modules/Core/Common/include/itkObjectFactoryBase.h
+++ b/Modules/Core/Common/include/itkObjectFactoryBase.h
@@ -271,10 +271,6 @@ private:
 
   const std::unique_ptr<OverrideMap> m_OverrideMap;
 
-  /** Initialize the static list of Factories. */
-  static void
-  InitializeFactoryList();
-
   /** Register default factories which are not loaded at run time. */
   static void
   RegisterInternal();
--- a/Modules/Core/Common/src/itkObjectFactoryBase.cxx
+++ b/Modules/Core/Common/src/itkObjectFactoryBase.cxx
@@ -245,7 +245,7 @@ ObjectFactoryBase::Initialize()
 void
 ObjectFactoryBase::RegisterInternal()
 {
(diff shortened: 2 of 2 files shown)


dzenanz (approved):

N-Dekker:
Please note: this is really still in an experimental state. I still need to figure out exactly how things work!

issakomi:
A little off topic. Interesting what OpenGL factory is mentioned here :)
```
/**
 * Register any factories that are always present in ITK like
 * the OpenGL factory, currently this is not done.
 */
void
ObjectFactoryBase::RegisterInternal()
```

It looks like the class was a long time ago copied from VTK [vtkObjectFactory.cxx](https://github.com/Kitware/VTK/blob/eb72baaa1f33899d819670fa42bf4804da54fd08/Common/Core/vtkObjectFactory.cxx#L72).

N-Dekker:
> A little off topic. Interesting what OpenGL factory is mentioned [here](https://github.com/InsightSoftwareConsortium/ITK/blob/c3a8eab2ad465b549a35d744cc92fb157e04dd94/Modules/Core/Common/src/itkObjectFactoryBase.cxx#L252) :)
> 
> ```
> /**
>  * Register any factories that are always present in ITK like
>  * the OpenGL factory, currently this is not done.
>  */
> void
> ObjectFactoryBase::RegisterInternal()
> ```

Thanks for your input, @issakomi I'm now proposing to just remove RegisterInternal(), including the apparently VTK specific comment about "the OpenGL factory". Please check:
- #4228

N-Dekker:
I'm afraid I was mistaken when I was thinking that `call_once` might be useful for `ObjectFactoryBase::Initialize()`. The member function claims to be a "one time initialization method":

https://github.com/InsightSoftwareConsortium/ITK/blob/ecd7bdae8295e8850d543923df32592d173e3257/Modules/Core/Common/src/itkObjectFactoryBase.cxx#L223-L227

But in fact, it may be called _more_ than once, on the very same `ObjectFactoryBase` instance, by calling `ReHash`:

https://github.com/InsightSoftwareConsortium/ITK/blob/ecd7bdae8295e8850d543923df32592d173e3257/Modules/Core/Common/src/itkObjectFactoryBase.cxx#L457-L462

So I guess we may need to reconsider using a mutex (PR #4226?)... at least if we think we have serious thread-safety problem here.

N-Dekker:
Abandoned: this approach (using `call_once`) does not work, because of `ReHash` as I mentioned before at https://github.com/InsightSoftwareConsortium/ITK/pull/4227#issuecomment-1749069411

PR #4231, opened 2023-09-26
ENH: Make VectorContainer::size_type public, test nested `vector` types

N-Dekker:
Moved the `protected` section to the end of the `VectorContainer` definition, allowing it to use `size_type` from the `public` section.

Tested that all nested types of `std::vector` are directly accessible from `itk::VectorContainer`.


dzenanz (approved):

thewtex (approved):
:+1:

hjmjohnson (approved):

PR #4233, opened 2023-09-29
cmake format

thewtex:
- ENH: Add cmake-format config and script
- STYLE: Apply ./Utilities/Maintenance/cmake-format.bash --tracked


jhlegarreta:
Beautiful @thewtex :100:. A few questions/comments:

- Will this format any change made to the CMake files locally prior to a commit? If not, do you think it can be done within the scope of this PR?
- Will this trigger a GHA linter workflow for CMake files? If not, do you think it can be done within the scope of this PR?
- IIRC, in the early days we adopted/CMake used to use uppercases for its commands, but then switched to lowercases. Will this change them to lowercases as well ? If not, can that be done through the config file.
- Can we add a short description to the body of both commit messages?

CMake/CppcheckTargets.cmake
```diff
@@ -1,30 +1,21 @@
...
-# Original Author:
-# 2009-2010 Ryan Pavlik <rpavlik@iastate.edu> <abiryan@ryand.net>
-# http://academic.cleardefinition.com
+# Original Author: 2009-2010 Ryan Pavlik <rpavlik@iastate.edu> <abiryan@ryand.net> http://academic.cleardefinition.com
 # Iowa State University HCI Graduate Program/VRAC
 #
-# Addition:
-# 2011 Arnaud Gelas <arnaud_gelas@hms.harvard.edu>
-# Harvard Medical School
+# Addition: 2011 Arnaud Gelas <arnaud_gelas@hms.harvard.edu> Harvard Medical School
```
dzenanz, line 15: Do we want this soft of reformatting of the comments? Can we disable it?

CMake/CppcheckTargets.cmake
```diff
@@ -280,35 +255,24 @@ function(add_cppcheck _name)
...
-          ${CPPCHECK_EXECUTABLE}
-          ${CPPCHECK_QUIET_ARG}
-          ${CPPCHECK_TEMPLATE_ARG}
-          ${_cppcheck_args}
-          ${_files}
-        WORKING_DIRECTORY
-          "${CMAKE_CURRENT_SOURCE_DIR}"
-        COMMENT
-          "${_name}_cppcheck: Running cppcheck on target ${_name}..."
-       V ERBATIM)
```
dzenanz, line 311: `V ERBATIM` with space seems like a bug. Does it affect anything?

CMake/ExternalData.cmake
```diff
@@ -415,19 +405,18 @@ function(ExternalData_add_target target)
...
           COMMENT "Generating ${file}"
           OUTPUT "${file}"
-          COMMAND ${CMAKE_COMMAND} -Drelative_top=${CMAKE_BINARY_DIR}
-                                   -Dfile=${file} -Dname=${name}
-                                   -DExternalData_ACTION=local
-                                   -DExternalData_CONFIG=${config}
-                                   -P ${_ExternalData_SELF}
-          MAIN_DEPENDENCY "${name}"
-          )
+          COMMAND ${CMAKE_COMMAND} -Drelative_top=${CMAKE_BINARY_DIR} -Dfile=${file} -Dname=${name}
```
dzenanz, line 417: This was more readable previously. Can this be configured?

Modules/IO/TransformMINC/CMakeLists.txt
```diff
@@ -2,11 +2,8 @@ project(ITKIOTransformMINC)
 set(ITKIOTransformMINC_LIBRARIES ITKIOTransformMINC)
 
 if(NOT ITK_USE_SYSTEM_HDF5)
-  include_directories(
-    "${ITKHDF5_SOURCE_DIR}/src/itkhdf5"
-    "${ITKHDF5_SOURCE_DIR}/src/itkhdf5/src"
-    "${ITKHDF5_BINARY_DIR}/src/itkhdf5"
-    )
+  include_directories("${ITKHDF5_SOURCE_DIR}/src/itkhdf5" "${ITKHDF5_SOURCE_DIR}/src/itkhdf5/src"
```
dzenanz, line 5: This was more readable previously. Could we enforce one dir per line?

dzenanz (commented):
Generally seems helpful, but I would like the style options adjusted somewhat if possible.

CMake/CppcheckTargets.cmake
thewtex: We can try disabling comment re-formatting.

CMake/CppcheckTargets.cmake
thewtex: This is an unrelated issue: patch here: https://github.com/InsightSoftwareConsortium/ITK/pull/4234

Modules/IO/TransformMINC/CMakeLists.txt
thewtex: I think we can make one argument per line more relevant per configuration.

CMake/ExternalData.cmake
thewtex: I tweaked the configuration. Also, I excluded this file, which is tracked externally.

dzenanz (approved):
While some changes are still a mixed bag, this is overall a noticeable improvement.

jhlegarreta:
>> Will this format any change made to the CMake files locally prior to a commit? If not, do you think it can be done within the scope of this PR?
>> Will this trigger a GHA linter workflow for CMake files? If not, do you think it can be done within the scope of this PR?

> I will look at adding it to pre-commit hooks and GitHub Actions in follow-up work.

> It has support for pre-commit. I think we should consider adopting it, but that is a bigger effort.

Thanks Matt. I'll open the corresponding issue to keep track of it.

>> IIRC, in the early days we adopted/CMake used to use uppercases for its commands, but then switched to lowercases. Will this change them to lowercases as well ? If not, can that be done through the config file.

> Yes, it takes care of the command case, in addition to many other nagging items! 💅

Beautiful :100:.

>> Can we add a short description to the body of both commit messages?

> Yes, done. I also added our *.wrap for formatting.

Thanks Matt. You rock.

jhlegarreta (approved):

PR #4247, opened 2023-10-02
ENH: Add `MersenneTwisterRandomVariateGenerator::ResetNextSeed()`

N-Dekker:
Allows generating a reproducible sequence of pseudo-random numbers.

----

FYI, I would like to use this member function to write reproducible unit tests for elastix Random Sampler, which internally uses ITK's `ImageRandomConstIteratorWithIndex`. The problem is that `ImageRandomConstIteratorWithIndex` creates a new MersenneTwisterRandomVariateGenerator, and each `MersenneTwisterRandomVariateGenerator::New()` "get the next seed", by calling `GetNextSeed()`: https://github.com/InsightSoftwareConsortium/ITK/blob/75b99671a89a9ef6ba1ad8fda62ba3843ad61b54/Modules/Core/Common/src/itkMersenneTwisterRandomVariateGenerator.cxx#L65
The proposed `ResetNextSeed()` would allow multiple `MersenneTwisterRandomVariateGenerator` objects (created by `New()`) to generate the very same sequence of numbers.

### DIFF
--- a/Modules/Core/Common/include/itkMersenneTwisterRandomVariateGenerator.h
+++ b/Modules/Core/Common/include/itkMersenneTwisterRandomVariateGenerator.h
@@ -160,6 +160,11 @@ public:
   static Pointer
   GetInstance();
 
+  /** Resets the internal data that is used to calculate the next seed. (Does not reset the initial seed.) Allows
+   * generating a reproducible sequence of pseudo-random numbers. */
+  static void
+  ResetNextSeed();
+
   /** Length of state vector */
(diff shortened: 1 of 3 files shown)


N-Dekker:
@thewtex Would it still be feasible to include this PR with the upcoming release (assuming that it will be approved)? It appears a useful feature for both SuperElastix/elastix and ITKElastix.

dzenanz (approved):

thewtex:
@N-Dekker yes :+1:

PR #4250, opened 2023-10-05
DOC: Miscellaneous `ITKIOTransformMINC` doc improvements

jhlegarreta:
- DOC: Remove unnecessary `ITKIOTransformMINC` `README` files
- DOC: Improve `ITKIOTransformMINC` class documentation

## PR Checklist

### DIFF
--- a/Modules/IO/TransformMINC/README
+++ b/Modules/IO/TransformMINC/README
@@ -1,5 +1,0 @@
-ITK4-IOMINC
-===========
-
-TransformMINC module for ITK integration, to read and write MINC-style .xfm
-files
--- a/Modules/IO/TransformMINC/include/itkMINCTransformIO.h
+++ b/Modules/IO/TransformMINC/include/itkMINCTransformIO.h
@@ -32,7 +32,7 @@ namespace itk
 
 /** \class MINCTransformIOTemplate
  *
- * \brief Read and write transforms in Minc XFM Format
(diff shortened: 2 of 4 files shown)


jhlegarreta:
The `itk-module.cmake` description is not shown in the Doxygen documentation; happens as well to other modules, e.g. `ITKIOPhilipsREC`, maybe because they are not enabled by default.

blowekamp:
> The `itk-module.cmake` description is not shown in the Doxygen documentation; happens as well to other modules, e.g. `ITKIOPhilipsREC`, maybe because they are not enabled by default.

Yes, that is my understanding of it. I believe this is the configuration use:
https://open.cdash.org/build/9040972/notes#note0

There is also the [ITKDoxygen](https://github.com/InsightSoftwareConsortium/ITKDoxygen/tree/master) repo where a PR could be make to change the configuration for that separate build.

jhlegarreta:
> There is also the ITKDoxygen repo where a PR could be make to change the configuration for that separate build.

OK. IMO, all modules should be enabled for the Doxygen build.

blowekamp:
> OK. IMO, all modules should be enabled for the Doxygen build.

Yes, that would be nice. I think some issue include that some modules have additional dependencies, and other issues that can cause the full Doxygen build to stop working. That is if there is a small issue in any one of the modules then it all fails.

thewtex (approved):

dzenanz (approved):

PR #4254, opened 2023-10-05
ENH: Add `itkImageRandomConstIteratorWithIndexGTest` unit test

N-Dekker:
Tests that `ImageRandomConstIteratorWithIndex` is deterministic when the initial seed _and_ the next seed of the global `MersenneTwisterRandomVariateGenerator` are reset.

----
Follow-up to:
- PR https://github.com/InsightSoftwareConsortium/ITK/pull/4247

### DIFF
--- a/Modules/Core/Common/test/CMakeLists.txt
+++ b/Modules/Core/Common/test/CMakeLists.txt
@@ -1714,6 +1714,7 @@ set(ITKCommonGTests
     itkImageBufferRangeGTest.cxx
     itkImageRegionRangeGTest.cxx
     itkImageIORegionGTest.cxx
+    itkImageRandomConstIteratorWithIndexGTest.cxx
     itkImageRegionGTest.cxx
     itkIndexGTest.cxx
     itkIndexRangeGTest.cxx
--- a/Modules/Core/Common/test/itkImageRandomConstIteratorWithIndexGTest.cxx
+++ b/Modules/Core/Common/test/itkImageRandomConstIteratorWithIndexGTest.cxx
@@ -1,0 +1,76 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         https://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+// First include the header file to be tested:
+#include "itkImageRandomConstIteratorWithIndex.h"
+#include "itkImage.h"
+#include "itkImageBufferRange.h"
+#include "itkMersenneTwisterRandomVariateGenerator.h"
+
+#include <gtest/gtest.h> // For iota.
+
+#include <numeric>
+
+
+// Tests that ImageRandomConstIteratorWithIndex is deterministic, when the initial seed and the next seed of the global
+// MersenneTwisterRandomVariateGenerator are reset before creating the iterator.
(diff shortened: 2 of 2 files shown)


dzenanz (approved):

blowekamp:
Hrmm.... I have been wondering what the underlying problem that is being trying to address with these changes to the random generator. 

It looks like the goal is to get the ImageRandomConstIteratorWithIndex to be reproducible. I have made quite a few changes to some other random generators and usages to get reproduce able results. It appears this approach to to modify global variable to get the Iterator to produce the same path. I am not sure how this is going to scale to multiple threads using iterator concurrently.

It seems to me the iterator is missing the option to set a seed parameter. The most release work I can recall doing it modifications to the Noise filter to get them to be reproducible (given a fixed number to threads, and region splitting ):
https://github.com/InsightSoftwareConsortium/ITK/blob/ecd7bdae8295e8850d543923df32592d173e3257/Modules/Filtering/ImageNoise/include/itkSaltAndPepperNoiseImageFilter.hxx#L52-L55 So a seed+index hash may be useful for the iterator to generate reproducible results.

Let me know if you would like to discuss options further.

N-Dekker:
Thanks @blowekamp I think ITK's MersenneTwisterRandomVariateGenerator is very complicated, especially because of how it is using global variables. Looking at your example: 

https://github.com/InsightSoftwareConsortium/ITK/blob/ecd7bdae8295e8850d543923df32592d173e3257/Modules/Filtering/ImageNoise/include/itkSaltAndPepperNoiseImageFilter.hxx#L52-L53

Even a local `MersenneTwisterRandomVariateGenerator::New()` call affects the global state. It creates a `PimplGlobals` singleton (if it isn't there already), and it increments `m_PimplGlobals->m_StaticDiffer` which is shared between all MersenneTwisterRandomVariateGenerator variables. Two identical New() calls yield generators different seeds.

MersenneTwisterRandomVariateGenerator appears to be added to ITK in 2005 (352e56087fadc01f52ac9f076554555a0799cbc0  by Daniel Blezek, @blezek), when the C++ Standard Library didn't yet have the random number utilities that it offers nowadays.  Ideally I would now entirely replace MersenneTwisterRandomVariateGenerator with `std::mt19937` (from `<random>`), but doing so would cause lots of regression test failures 🤔

N-Dekker:
> That was by design.

Possible, but then I don't like the design! It makes it harder to write reproducible unit tests and reproducible image registrations. 

For new C++ code (assuming that it does not need to support legacy use cases), I would rather use the Standard C++ Library utilities for random number generation, from `<random>`.

blowekamp:
> Possible, but then I don't like the design! It makes it harder to write reproducible unit tests and reproducible image registrations.

If reproducible results are needed then the global and/or algo specific thread seeds are manually set. Not hard.

N-Dekker:
> If reproducible results are needed then the global and/or algo specific thread seeds are manually set. Not hard.

Before using `ImageRandomConstIteratorWithIndex` locally in a reproducible way, you'd have to do:

      MersenneTwisterRandomVariateGenerator::GetInstance()->SetSeed(initialSeed);
      MersenneTwisterRandomVariateGenerator::ResetNextSeed();

Right?

https://github.com/InsightSoftwareConsortium/ITK/blob/bd9e380bcde2792dd8687bce0a9501b1cdff099c/Modules/Core/Common/test/itkImageRandomConstIteratorWithIndexGTest.cxx#L53-L55

blowekamp:
> Before using `ImageRandomConstIteratorWithIndex` locally in a reproducible way, you'd have to do:

No. The ImageRegistrationMethodv4 was implemented with this iterator to be reproducible:

https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Registration/RegistrationMethodsv4/include/itkImageRegistrationMethodv4.hxx#L1004-L1013

dzenanz:
Brad, can you approve this PR or provide more actionable feedback? It is not clear to me whether this is good as-is or it needs more work.

N-Dekker:
I'm not in a hurry for this PR. Apparently there's a choice now between doing `MersenneTwisterRandomVariateGenerator::GetInstance()->SetSeed(initialSeed); ResetNextSeed();` _before_ constructing the iterator and calling `iterator.ReinitializeSeed(initialSeed)` afterwards. Under the hood, the `PimplGlobals` singleton gets modified in either way, but differently.

dzenanz:
@blowekamp should this be merged?

PR #4264, opened 2023-10-19
nifti 2023-09-11 (75180f77)

seanm:
Code extracted from:

    https://github.com/NIFTI-Imaging/nifti_clib.git

at commit 75180f776997bd161c5b231dfffb65e6385f5f93 (master).

### DIFF
--- a/Modules/ThirdParty/NIFTI/src/nifti/niftilib/nifti1_io.c
+++ b/Modules/ThirdParty/NIFTI/src/nifti/niftilib/nifti1_io.c
@@ -658,38 +658,38 @@ int nifti_update_dims_from_array( nifti_image * nim )
    nim->dx = nim->pixdim[1];
 
    /* if undefined, or less than 1, set to 1 */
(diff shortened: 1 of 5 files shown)


dzenanz (changes requested):
These updates should be done by running the appropriate update script, in this case https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/ThirdParty/NIFTI/UpdateFromUpstream.sh. Do you mind doing that and force-pushing?

seanm:
> These updates should be done by running the appropriate update script, in this case https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/ThirdParty/NIFTI/UpdateFromUpstream.sh. Do you mind doing that and force-pushing?

That's how I generated this patch.  Does it seem otherwise?

dzenanz:
It definitely seems otherwise. There should be at least two commits, update of nifti, and merge to master. Usually there is also a tag update commit. Your other PR, #4262, is a good example of that.

dzenanz:
Closing in favor of #4269.

seanm:
Weird. I swear I did the exact same here, but whatever. :)

PR #4266, opened 2023-10-23
Deprecate `Unused`, use C++17 `[[maybe_unused]]` instead

N-Dekker:
Two style commits:

 - Replace `Unused` calls with C++17 `[[maybe_unused]]` attribute
 - Deprecate (ITK_FUTURE_LEGACY_REMOVE) `Unused(const T &)`


Modules/Core/Common/src/itkObjectFactoryBase.cxx
```diff
@@ -832,8 +832,7 @@ ObjectFactoryBase::SynchronizeObjectFactoryBase(void * objectFactoryBasePrivate)
 std::list<ObjectFactoryBase *>
 ObjectFactoryBase::GetRegisteredFactories()
 {
-  //  static SingletonIndex * singletonIndex = SingletonIndex::GetInstance();
-  //  Unused(singletonIndex);
+  //  [[maybe_unused]] static SingletonIndex * singletonIndex = SingletonIndex::GetInstance();
```
N-Dekker, line 835: This commented-out declaration of `singletonIndex` was introduced with pull request #118 commit a66337ec215c88f4900a2caf419b055483c42085 "ENH: Synchronize factories across modules in Python", Francois Budin (@fbudin69500), February 2019:

https://github.com/InsightSoftwareConsortium/ITK/blob/a66337ec215c88f4900a2caf419b055483c42085/Modules/Core/Common/src/itkObjectFactoryBase.cxx#L925-L926

Maybe we might as well just remove those two commented-out lines of code. It's very hard to keep commented-out code up-to-date. What do you think?

dzenanz: Since you already did the work, perhaps let's keep it this time. But yes, we can occasionally clean up commented-out code.

dzenanz (approved):

thewtex (approved):

dzenanz (approved):

PR #3428, opened 2022-05-13
ENH: Use https instead of http when https works

Leengit:
`https` is more tamper-resistant and more private so we use it instead of `http` when it is available.

We detect which URLs work with `https` using
```shell
cd git/ITK
find . -type f | \
    fgrep -v sed-commands | fgrep -v .git/ | fgrep -v ThirdParty | \
    tr \\n \\0 | xargs -0 egrep -o 'http://[-._A-Za-z0-9]+' | cut -d : -f 2- | sort | uniq | \
    while read url
        do curl --connect-timeout 60 --output /dev/null --silent --head --fail "https${url#http}" && \
           echo "-e 's#${url}#https${url#http}#g' \\"
    done > sed-commands
```

We edit the ITK repository files by appending the output of the above command (except for the very last trailing backslash) to the command
```shell
cd git/ITK
find . -type f | \
    fgrep -v sed-commands | fgrep -v .git/ | fgrep -v ThirdParty | \
    tr \\n \\0 | xargs -0 sed -i -r \
```

PR #4273, opened 2023-10-30
Fix XML namespace of Natvis file, exclude Utilities/Debugger from CI

N-Dekker:
The ITK Debug Visualizers for Visual Studio do not seem to work when the XML namespace name starts with "https" (rather than just "http"). This was found by setting the option "Natvis diagnostic messages" to "Verbose", in either Visual Studio 2019 (version 16.11.31) or Visual Studio 2022 (version 17.7.6). It says:

> ITK.natvis(4,2): Fatal error: Expected element with namespace 'http://schemas.microsoft.com/vstudio/debugger/natvis/2010'.

Reverts the change of this particular file from pull request https://github.com/InsightSoftwareConsortium/ITK/pull/3428 commit 9cd0f2053f0cf181ec4412bf896301faf0cc2092 "ENH: Use https instead of http when https works", May 18, 2022.

-- With help from colleagues of mine at LKEB, Leiden University Medical Center.

Also excluded changes in "Utilities/Debugger" (like the proposed one in "ITK.natvis") from triggering the CI.

### DIFF
--- a/Utilities/Debugger/ITK.natvis
+++ b/Utilities/Debugger/ITK.natvis
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <!-- Documentation: https://msdn.microsoft.com/en-us/library/jj620914.aspx?f=255&MSPPError=-2147217396 -->
 <!-- e.g. copy into C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\Packages\Debugger\Visualizers\ITK.natvis -->
-<AutoVisualizer xmlns="https://schemas.microsoft.com/vstudio/debugger/natvis/2010">
+<AutoVisualizer xmlns="http://schemas.microsoft.com/vstudio/debugger/natvis/2010">
 
   <Type Name="itk::Size&lt;1&gt;">
     <DisplayString>[{m_InternalArray[0]}]</DisplayString>
--- a/Testing/ContinuousIntegration/AzurePipelinesLinux.yml
+++ b/Testing/ContinuousIntegration/AzurePipelinesLinux.yml
@@ -11,6 +11,7 @@ trigger:
     - LICENSE
     - NOTICE
     - Documentation/*
+    - Utilities/Debugger/*
     - Utilities/ITKv5Preparation/*
     - Utilities/Maintenance/*
 pr:
@@ -20,6 +21,7 @@ pr:
     - LICENSE
     - NOTICE
     - Documentation/*
+    - Utilities/Debugger/*
     - Utilities/ITKv5Preparation/*
     - Utilities/Maintenance/*
 variables:
--- a/Testing/ContinuousIntegration/AzurePipelinesLinuxPython.yml
+++ b/Testing/ContinuousIntegration/AzurePipelinesLinuxPython.yml
@@ -11,6 +11,7 @@ trigger:
     - LICENSE
     - NOTICE
     - Documentation/*
+    - Utilities/Debugger/*
     - Utilities/ITKv5Preparation/*
     - Utilities/Maintenance/*
 pr:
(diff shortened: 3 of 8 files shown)


dzenanz (approved):

Utilities/Debugger/ITK.natvis
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <!-- Documentation: https://msdn.microsoft.com/en-us/library/jj620914.aspx?f=255&MSPPError=-2147217396 -->
 <!-- e.g. copy into C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\Packages\Debugger\Visualizers\ITK.natvis -->
-<AutoVisualizer xmlns="https://schemas.microsoft.com/vstudio/debugger/natvis/2010">
+<AutoVisualizer xmlns="http://schemas.microsoft.com/vstudio/debugger/natvis/2010">
```
N-Dekker, line 4: I guess this fix should also be applied to the XML file "default.natstepfilter": 

https://github.com/InsightSoftwareConsortium/ITK/blob/3bbcb195fb3f8ed84d7e2081eced74ff1b0baf32/Utilities/Debugger/default.natstepfilter#L3 

But I don't really know yet. @Leengit Dit you intend to _also_ adjust those XML namespace names to use "https", when you submitted PR #3428 ?

N-Dekker: Update: officially, I think that the ".natstepfilter" file should just have "http", not "https", in the `xmlns` attribute of `StepFilter`. Looking at _Learn / Visual Studio / Debugging_ section [Customize C++ stepping behavior independent of Just My Code settings](https://learn.microsoft.com/en-us/visualstudio/debugger/just-my-code?view=vs-2022#BKMK_CPP_Customize_stepping_behavior):

> A _.natstepfilter_ file is an XML file with this syntax:
```

<?xml version="1.0" encoding="utf-8"?>
<StepFilter xmlns="http://schemas.microsoft.com/vstudio/debugger/natstepfilter/2010">
    <Function>
        <Name>FunctionSpec</Name>
        <Action>StepAction</Action>
    </Function>
    <Function>
        <Name>FunctionSpec</Name>
        <Module>ModuleSpec</Module>
        <Action>StepAction</Action>
    </Function>
</StepFilter>

``` 

However, it seems to me the Visual Studio debugger entirely ignores the xmlns attribute of `StepFilter` (as I just observed, using VS2022), so it may not matter so much anyway.  🤷  Still, for a _.natvis_ file, an incorrect xmlns attribute is a show-stopper.

Leengit: > But I don't really know yet. @Leengit Dit you intend to _also_ adjust those XML namespace names to use "https", when you submitted PR #3428 ?

I didn't differentiate contexts or file types.  Any http-style URL that worked via curl as a https-style URL was changed.  A few of those changes were manually undone: if a change proved to be a test breaker and couldn't be fixed by hand in some other way then it was undone.

dzenanz (approved):
Changing https -> http for `.natstepfilter` now would be good.

N-Dekker:
> LGTM.

Thanks @Leengit!

> The switch from http to https was made to add additional security where it was available...

As far as I understand, the security argument doesn't really apply to XML namespace names. The XML interpreter (Visual Studio, in this case) does not really access the internet location specified by the `xmlns` attribute. But please correct me if I'm wrong. 

> For the future ... perhaps the `yml` files can all `include` some new file that has their commonalities, rather than having so much duplication among them.

Sorry, I haven't been able to figure out how to do that: https://github.com/InsightSoftwareConsortium/ITK/pull/3930#issuecomment-1435715397  But if you can, please give it a try!

dzenanz (approved):

PR #4278, opened 2023-11-03
Add `Deref(T *)`, to ease dereferencing a pointer safely

N-Dekker:
`Deref(ptr)` throws an exception when its argument (`ptr`) is null.

A similar function template is being used internally in elastix for quite a while: https://github.com/SuperElastix/elastix/blob/f64965f09eb0903e7bf1da4a112d914c7ac2ce22/Common/elxDeref.h

### DIFF
--- a/Modules/Core/Common/include/itkDeref.h
+++ b/Modules/Core/Common/include/itkDeref.h
@@ -1,0 +1,59 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         https://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+#ifndef itkDeref_h
+#define itkDeref_h
+
+#include "itkMacro.h"
+#include <string>
+#include <typeinfo>
+
+namespace itk
+{
+
+/** \class DerefError
+ * Exception thrown when trying to dereference a null pointer.
+ * \ingroup ITKCommon
+ */
+class DerefError : public ExceptionObject
+{
+public:
+  // Inherit the constructors from its base class.
+  using ExceptionObject::ExceptionObject;
+
+  /** Runtime information support. */
+  itkTypeMacro(DerefError, ExceptionObject);
+};
+
+
+/** Dereferences the specified pointer, when the pointer is not null. Throws a `DerefError` exception when the pointer
+ * is null. Aims to avoid undefined behavior from accidentally dereferencing a null pointer.
+ */
+template <typename T>
+T &
+Deref(T * const ptr)
+{
+  if (ptr == nullptr)
+  {
+    itkSpecializedMessageExceptionMacro(
+      DerefError, "The pointer passed to `itk::Deref(T*)` is null! T's typeid name: `" << typeid(T).name() << '`');
+  }
+  return *ptr;
+}
+} // namespace itk
+
+#endif
--- a/Modules/IO/MeshVTK/test/itkVTKPolyDataMeshIOGTest.cxx
+++ b/Modules/IO/MeshVTK/test/itkVTKPolyDataMeshIOGTest.cxx
@@ -19,6 +19,7 @@
 // First include the header file to be tested:
 #include "itkMeshFileReader.h"
 
+#include "itkDeref.h"
 #include "itkMesh.h"
 #include "itkMeshFileWriter.h"
 #include "itkVTKPolyDataMeshIO.h"
@@ -53,9 +54,8 @@ Expect_lossless_writing_and_reading_of_points(const std::string &               
 {
   const auto inputMesh = TMesh::New();
 
-  const auto inputPoints = inputMesh->GetPoints();
-  ASSERT_NE(inputPoints, nullptr);
-  inputPoints->assign(points.cbegin(), points.cend());
+  auto & inputPoints = itk::Deref(inputMesh->GetPoints());
+  inputPoints.assign(points.cbegin(), points.cend());
 
   const auto writer = itk::MeshFileWriter<TMesh>::New();
   if (writeAsBinary)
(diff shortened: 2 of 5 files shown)


Modules/Core/Common/include/itkDeref.h
```diff
@@ -0,0 +1,62 @@
...
+/** Dereferences the specified pointer (either "raw" or "smart"), when the pointer is not null. Throws a `DerefError`
+ * exception when the pointer is null. Aims to avoid undefined behavior from accidentally dereferencing a null pointer.
+ */
+template <typename TPointer>
+decltype(auto)
+Deref(const TPointer & ptr)
+{
+  if (ptr == nullptr)
+  {
+    throw DerefError(__FILE__,
```
dzenanz, line 52: Is this file reported in the exception message, or the line where it is invoked?

dzenanz (commented):
Adding a follow-up commit with uses of this new functionality are highly welcome!

Modules/Core/Common/include/itkDeref.h
N-Dekker: Yes indeed, both of them (source file and line number) are reported in the exception message.

dzenanz: I meant, is file == ".../Modules/Core/Common/include/itkDeref.h", or file == ".../my/file/in/my/code/which/tried/to/deref/null.cpp"?

N-Dekker: The `itk::DerefError` exception message proposed by this PR will have something like "Modules/Core/Common/include/itkDeref.h" in there. Of course, it would have been nicer to have something like "my/file/in/my/code/which/tried/to/deref/null.cpp". I think that can be implemented most elegantly by C++20 [std::source_location](https://en.cppreference.com/w/cpp/utility/source_location) 😃. Until then, I think it's still nice for users to know the location where the `DerefError` is thrown, in "itkDeref.h", so that they can put a breakpoint at that location, when they are debugging the problem.

dzenanz (approved):

Modules/Core/Common/include/itkDeref.h
blowekamp: Did you consider using the "itkSpecializedMessageExceptionMacro?

N-Dekker: Thanks @blowekamp, please check my [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/0f35aa8bd59c3793e1e694c7f3b0ec8c679b0617..f881d08b438ae1fe102bfa5eeae849389362c31f)!

dzenanz (approved):

N-Dekker:
Update: I just removed `Deref` support for smart pointers, because there is _some_ risk that it might lead to bugs. Suppose a function `CreateObject()` would return a smart pointer to a newly created object. The following might then cause `ref` to refer to a _destructed_ object (after the line of code is executed):

    auto & ref = Deref( CreateObject() );

On the other hand, `Deref` can certainly be used safely on raw pointers, as a raw pointer does not control the lifetime of the object pointed to.

N-Dekker:
@dave3d Can you please explain how to fix the Spell Checking failures?

```
Run comment_spell_check \
  comment_spell_check \
    --miss \
    --prefix "gdcm" --exclude ThirdParty \
    --dict additional_dictionary.txt \
    --dict .github/workflows/itk_dict.txt \
    --suffix ".h" \
    --suffix ".hxx" \
    --suffix ".py" \
    Modules
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.9.18/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.9.18/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.9.18/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.9.18/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.9.18/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.9.18/x64/lib
    Modules/Core/Common/include/itkDeref.h, 28

    Modules/Core/Common/include/itkDeref.h, 43
DerefError:

2 misspellings found
``` 

At https://github.com/InsightSoftwareConsortium/ITK/actions/runs/6803791308/job/18499888354?pr=4278

dave3d:
> @dave3d Can you please explain how to fix the Spell Checking failures?
> 
> ```
> Run comment_spell_check \
>   comment_spell_check \
>     --miss \
>     --prefix "gdcm" --exclude ThirdParty \
>     --dict additional_dictionary.txt \
>     --dict .github/workflows/itk_dict.txt \
>     --suffix ".h" \
>     --suffix ".hxx" \
>     --suffix ".py" \
>     Modules
>   shell: /usr/bin/bash -e {0}
>   env:
>     pythonLocation: /opt/hostedtoolcache/Python/3.9.18/x64
>     PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.9.18/x64/lib/pkgconfig
>     Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.9.18/x64
>     Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.9.18/x64
>     Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.9.18/x64
>     LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.9.18/x64/lib
>     Modules/Core/Common/include/itkDeref.h, 28
> 
>     Modules/Core/Common/include/itkDeref.h, 43
> DerefError:
> 
> 2 misspellings found
> ```
> 
> At https://github.com/InsightSoftwareConsortium/ITK/actions/runs/6803791308/job/18499888354?pr=4278

DerefError is a new keyword that should be added to the spelling dictionary.  Add it to the file .github/workflows/itk_dict.txt.

N-Dekker:
Thanks for your feedback, @dzenanz, @dave3d Personally I find "Deref" clear enough. When I google on "deref", it's all about dereferencing: https://www.google.com/search?q=deref And as "DerefError" is the error that may occur specifically when calling "Deref", I would prefer to stick with "DerefError".  Hope that's acceptable to you both 🙏 

@dave3d would it be sufficient to add "Deref" to .github/workflows/itk_dict.txt in order to use "DerefError"? I guess it knows "Error" already, right?

dave3d:
> Thanks for your feedback, @dzenanz, @dave3d Personally I find "Deref" clear enough. When I google on "deref", it's all about dereferencing: https://www.google.com/search?q=deref And as "DerefError" is the error that may occur specifically when calling "Deref", I would prefer to stick with "DerefError". Hope that's acceptable to you both 🙏
> 
> @dave3d would it be sufficient to add "Deref" to .github/workflows/itk_dict.txt in order to use "DerefError"? I guess it knows "Error" already, right?

Yes, you can just add 'deref'.

Modules/IO/MeshVTK/test/itkVTKPolyDataMeshIOGTest.cxx
```diff
@@ -76,18 +76,16 @@ Expect_lossless_writing_and_reading_of_points(const std::string &
   reader->SetMeshIO(itk::VTKPolyDataMeshIO::New());
   reader->Update();
 
-  const auto outputMesh = reader->GetOutput();
-  ASSERT_NE(outputMesh, nullptr);
-  const auto outputPoints = outputMesh->GetPoints();
-  ASSERT_NE(outputPoints, nullptr);
+  const auto & outputMesh = itk::Deref(reader->GetOutput());
```
dzenanz, line 79: This is a good demonstration of how this is supposed to be used. I would like to see some more instances of use, better demonstrating a need for a new function and exception class.

N-Dekker: Thanks @dzenanz 

I use `elastix::Deref` extensively within elastix, e.g. [(source)](https://github.com/SuperElastix/elastix/blob/5f0940d2ec5498a6bc98f53c3eaa6a0a1ccd74b3/Common/ImageSamplers/itkImageRandomSampler.hxx#L56-L57):

    elastix::Deref(MersenneTwisterRandomVariateGenerator::GetInstance()).SetSeed(...);

and [(source)](https://github.com/SuperElastix/elastix/blob/5f0940d2ec5498a6bc98f53c3eaa6a0a1ccd74b3/Common/ImageSamplers/itkImageRandomSampler.hxx#L59-L61)

    MultiThreaderBase & multiThreader = elastix::Deref(this->ProcessObject::GetMultiThreader());
    multiThreader.SetSingleMethod(...);
    multiThreader.SingleMethodExecute();

Once `itk::Deref` is available, I would like to replace those `elastix::Deref` calls with `itk::Deref`. I feel that this function isn't really elastix specific. It could be useful for other ITK users as well. Does that answer your question?

PR #4289, opened 2023-11-06
ENH: Add `get()` member function to itk::SmartPointer

N-Dekker:
Eases code that uses both `itk` and `std` smart pointers. Also serves as a convenient shorter alternative to the somewhat lengthy "GetPointer()".

### DIFF
--- a/Modules/Core/Common/include/itkSmartPointer.h
+++ b/Modules/Core/Common/include/itkSmartPointer.h
@@ -134,6 +134,15 @@ public:
     return m_Pointer;
   }
 
+  /** Returns the stored (raw) pointer. Equivalent to `GetPointer()`, but then following the Standard C++ Library naming
+   * conversion (like `std::shared_ptr::get()`). */
+  ObjectType *
+  get() const noexcept
+  {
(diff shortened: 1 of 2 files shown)


dzenanz (approved):

Modules/Core/Common/include/itkSmartPointer.h
```diff
@@ -134,6 +134,15 @@ class SmartPointer
     return m_Pointer;
   }
 
+  /** Returns the stored (raw) pointer. Equivalent to `GetPointer()`, but then following the Standard C++ Library naming
+   * conversion (like `std::shared_ptr::get()`). */
+  ObjectType *
+  get() const noexcept
```
jhlegarreta, line 140: Why `get` instead of `Get`?

dzenanz: Because STL uses `get` (lowercase): https://en.cppreference.com/w/cpp/memory/shared_ptr/get

hjmjohnson (approved):

thewtex (approved):
Awesome!

PR #4290, opened 2023-11-06
STYLE: ResetNextSeed GTest should not use instance for GetNextSeed call

N-Dekker:
`MersenneTwisterRandomVariateGenerator::GetNextSeed()` is a static member function, so there is no need to call it via an instance of the generator, in the `MersenneTwisterRandomVariateGenerator.ResetNextSeed` unit test.

The commit also adds some documentation to the unit test.

Suggested by clang-tidy:
https://clang.llvm.org/extra/clang-tidy/checks/readability/static-accessed-through-instance.html

----

Obviously it was my own fault  😸 introduced with:
- https://github.com/InsightSoftwareConsortium/ITK/pull/4247

### DIFF
--- a/Modules/Core/Common/test/itkMersenneTwisterRandomVariateGeneratorGTest.cxx
+++ b/Modules/Core/Common/test/itkMersenneTwisterRandomVariateGeneratorGTest.cxx
@@ -64,14 +64,16 @@ TEST(MersenneTwisterRandomVariateGenerator, GetIntegerVariateReturnsSameAsStdMt1
 }
 
 
+// Tests that two GetNextSeed() calls return the very same seed value, when ResetNextSeed() is called before each of
+// those calls.
 TEST(MersenneTwisterRandomVariateGenerator, ResetNextSeed)
 {
-  const auto globalGenerator = MersenneTwisterRandomVariateGenerator::GetInstance();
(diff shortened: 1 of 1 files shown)


dzenanz (approved):

Modules/Core/Common/test/itkMersenneTwisterRandomVariateGeneratorGTest.cxx
```diff
@@ -64,14 +64,12 @@ TEST(MersenneTwisterRandomVariateGenerator, GetIntegerVariateReturnsSameAsStdMt1
 }
 
 
+// Tests that two GetNextSeed() calls return the same very seed, when ResetNextSeed() is called beforehand.
```
jhlegarreta, line 67: the same very seed : the very same seed

N-Dekker: Good catch, Jon! Fixed by [this force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/838e9be4aa8dd904258bf83680d56f49712dea0f..d871a24a825cfe6d60180f4ba6f1d8fa42fe3a9f.)

dzenanz (approved):

PR #4293, opened 2023-11-06
Gdcm 2023 11 06

seanm:


seanm:
@thewtex @dzenanz looks like nuking my git checkout worked!

Since @dzenanz updated GDCM just the other day, there's not much here except I added the .codespellrc config file.

dzenanz (approved):

thewtex (approved):
Woohoo!

PR #4301, opened 2023-11-08
ENH: ShapedImageNeighborhoodRange support C-array of offsets (by C++17)

N-Dekker:
Allowed the shape of a neighborhood to be specified by a C-array of offset, by using C++17 `std::data` and `std::size` internally, instead of `data()` and `size()` member functions.


dzenanz (approved):

thewtex (approved):
Cool!

PR #4323, opened 2023-11-13
ENH: Add test for `itk::XMLFilterWatcher`

jhlegarreta:
Add test for `itk::XMLFilterWatcher`: the watcher is applied to an image reading filter, and the progress updates written to the standard output by the filter are redirected to a text file through the CMake `--redirectOutput` option.

## PR Checklist


jhlegarreta:
A few notes:
- Did not add a message at the end of test about the test being finished because the whole test output is redirected to a file through CMake. Redirecting the output using CMake instead of redirecting the output from the code avoids boilerplate code.
- The test is not quantitative: comparing the files with the CMake `--compare` or the `--compare-MD5` options does not work as they both expect an image to apply an ITK Image IO class for reading the file. I ignore whether comparing TXT files can be achieved.

dzenanz (approved):

PR #4327, opened 2023-11-14
STYLE: Remove unreachable `break` statements after `return` statements

N-Dekker:
By doing Notepad++ Replace in Files, having:

    Find what:  return (.*);\r\n [ ]* break;
    Replace with:  return $1;
    Filters: itk*.h;itk*.hxx;itk*.cxx
    [v] Match case
    (*) Regular expression


dzenanz (approved):

thewtex (approved):
:100:

PR #4336, opened 2023-11-19
Data archive docs for 5.4

thewtex:


Utilities/Maintenance/SourceTarball.bash
```diff
@@ -36,27 +36,20 @@ return_pipe_status() {
...
-  if test "$md5sum" != "$2"; then
-    die "Object MD5/$2 is corrupt: $1"
-  fi
-}
-
 download_object() {
   algo="$1" ; hash="$2" ; path="$3"
   mkdir -p $(dirname "$path") &&
-  if curl -L "https://www.itk.org/files/ExternalData/$algo/$hash" -o "$path.tmp$$" 1>&2; then
+  if curl -L "http://127.0.01:8080/ipfs/$hash" -o "$path.tmp$$" 1>&2; then
```
phcerdan, line 52: Double checking: Is this local url correct?
How are we setting up that local url? is that address and port the default values of kubo?

phcerdan (approved):
Great docs!

One tiny question below.

Utilities/Maintenance/SourceTarball.bash
thewtex: Correct, that is the default value for `kubo` -- it can be started by running [ipfs-desktop](https://github.com/ipfs/ipfs-desktop/releases).

PR #4344, opened 2023-11-22
PERF: Make `ImageRegion` trivially copyable, remove inheritance (`ITK_FUTURE_LEGACY_REMOVE`)

N-Dekker:
Made all member functions of `ImageRegion` non-virtual, and removed its inheritance from `Region`.

The size of `ImageRegion<2>` was observed to go down from 40 bytes (before this commit) to 32 bytes (after this commit), on Visual C++ 2022 (x64). A run-time performance improvement of more than 25% on default-constructing and destructing a sequence of `ImageRegion` objects was observed, by benchmarking `std::make_unique<itk::ImageRegion<2>[]>(n)`.

### DIFF
--- a/Modules/Core/Common/include/itkImageRegion.h
+++ b/Modules/Core/Common/include/itkImageRegion.h
@@ -34,6 +34,15 @@
 #include "itkContinuousIndex.h"
 #include "itkMath.h"
 
+// Macro added to each `ImageRegion` member function that overrides a virtual member function of `Region`. In the
+// future, `ImageRegion` will no longer inherit from `Region`, so then those `ImageRegion` member functions will no
+// longer override.
+#ifdef ITK_FUTURE_LEGACY_REMOVE
+#  define itkRegionOverrideMacro // nothing (in the future)
+#else
+#  define itkRegionOverrideMacro override
+#endif
+
 namespace itk
 {
 // Forward declaration of ImageBase so it can be declared a friend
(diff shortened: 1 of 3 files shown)


dzenanz (commented):
Let's see how testing goes.

`itk::Region` itself is not much used. Did you consider making it trivially copyable? That way we could keep the inheritance.

N-Dekker:
> `itk::Region` itself is not much used. Did you consider making it trivially copyable? That way we could keep the inheritance.

Technically that might work as well, but it's a bit risky to keep the inheritance, when we make `itk::Region`  trivially copyable. You know, old books taught us to declare destructors of base classes `virtual`, to avoid undefined behavior when deleting a derived object through a pointer to its base.  We could mitigate such a risk by using `private` inheritance, instead of `public` inheritance. Or by declaring the base class destructor `protected`. But I guess that would beat the original purpose of the inheritance in the first place.  🤷 

My feeling is that the `itk::Region` base class might have been a nice idea originally, it might have some rare use cases, but `ImageRegion` is being used so extensively, also in performance critical code, that it's worth considering to drop the inheritance feature.

N-Dekker:
Interestingly, only when `ImageRegion` is made trivially copyable, compiler warnings about unused `ImageRegion` variables appear. At https://open.cdash.org/viewBuildError.php?type=1&buildid=9172873 (ITK.macOS/Darwin/Xcode_13.2.1/clang-1300.0.29.30):

> itkZeroFluxNeumannBoundaryCondition.hxx:97:14: warning: unused variable 'requestRegion' [-Wunused-variable]

A fix is included with this pull request.

blowekamp:
I would consider this a breaking API change, that is only suitable for a major revision.

N-Dekker:
> I would consider this a breaking API change, that is only suitable for a major revision.

Thanks Bradley, good point. This is indeed an API change. Which specific ITK revision could then include a change like this?

dzenanz:
Perhaps make it future legacy? So it becomes legacy in 6.0 (soon), and gets removed in 7.0.

N-Dekker:
Reverted status to "work in progress", because I would prefer to first see the related ("spin-off") pull request being processed:

- #4345

(Afterwards I would rebase this one.)

dzenanz (approved):

thewtex (approved):
:rocket: :fire: 

@N-Dekker very nice!

Note: the previously failing `first-interaction` check is now fixed via https://github.com/actions/first-interaction/pull/287

PR #4347, opened 2023-11-26
PERF: Replace `std::function` with template argument MultiThreaderBase ParallelizeImageRegion, ParallelizeImageRegionRestrictDirection

N-Dekker:
Declared the `funcP` parameter of `MultiThreaderBase` template member functions `ParallelizeImageRegion` and `ParallelizeImageRegionRestrictDirection` by a template argument, `TFunction`, instead of by `TemplatedThreadingFunctorType`. The `funcP` parameter is typically specified by a lambda. Using `TFunction` is more lightweight, as it avoids converting the lambda to an
`TemplatedThreadingFunctorType`. (`TemplatedThreadingFunctorType` is an alias of `std::function`.)

Deprecated (`ITK_FUTURE_LEGACY_REMOVE`) TemplatedThreadingFunctorType, as it is no longer used now.


blowekamp (changes requested):
The is no replacement documentation as to what this template parameter should be.

issakomi:
Finally this function (itkMultiThreaderBase.cxx) is called, `MultiThreaderBase::ThreadingFunctorType` is  `std::function<void(const IndexValueType index[], const SizeValueType size[])>;`

```cpp
void
MultiThreaderBase::ParallelizeImageRegion(unsigned int                            dimension,
                                          const IndexValueType                    index[],
                                          const SizeValueType                     size[],
                                          MultiThreaderBase::ThreadingFunctorType funcP,
                                          ProcessObject *                         filter)
{
  // This implementation simply delegates parallelization to the old interface
  // SetSingleMethod+SingleMethodExecute. This method is meant to be overloaded!
  if (!this->GetUpdateProgress())
  {
    filter = nullptr;
  }
  ProgressReporter progress(filter, 0, 1);

  struct RegionAndCallback rnc
  {
    funcP, dimension, index, size, filter
  };
  this->SetSingleMethod(&MultiThreaderBase::ParallelizeImageRegionHelper, &rnc);
  this->SingleMethodExecute();
}
```

Experimented a little:

```cpp
#include <memory>
#include <functional>
#include <iostream>
#include <cxxabi.h>

///////////////////////////////////////////////////////////
//
// To avoid memory leak with abi::__cxa_demangle
//
struct deleter
{
    void operator()(void * p) const { std::free(p); }
};
using ptr = std::unique_ptr<char, deleter>;
//
//
///////////////////////////////////////////////////////////

template <typename TFunction>
int func(TFunction f)
{
  int status; // Not checked
  ptr name(abi::__cxa_demangle(typeid(f).name(), nullptr, 0, &status));
  auto x = f(2);
  std::cout << name.get() << " " << x << std::endl;
  return x;
}

int main()
{
  auto                    l = [](int x) -> int { return x + 40; };
  std::function<int(int)> f = [](int x) -> int { return x + 40; };

  func(l);
  func(f);
  func([](int x) -> int { return x + 40; });
  
  return 0;
}
```

```
r@deb2:~$ g++ test.cpp 
r@deb2:~$ ./a.out 
main::{lambda(int)#1} 42
std::function<int (int)> 42
main::{lambda(int)#3} 42
```

issakomi (approved):

N-Dekker:
@issakomi Thanks for your approval, and thanks for sharing your experiment. I did not even know about `abi::__cxa_demangle` ! 

I think we may conclude from your experiment that:

1. The template parameter `TFunction` supports both lambda's and `std::function` objects
2. For each of the three `func(TFunction)` calls, the compiler does separate template instantiation, as `TFunction` is a different type for each instantiation.  In practice, this will allow the compiler to optimize the code specifically for each different call (for each instantiation, that is).
3. In any case, with any of the three `func` calls, the answer is always 42  😄 
 
Cool!

dzenanz:
> Proposed new text:

Perhaps add there references/links to some place where this is invoked, to serve as examples of how this is supposed to be called?

issakomi:
> The template parameter TFunction supports both lambda's and std::function objects

Looks like it is.
One more way:

```cpp
func([](int x) -> int { return x + 40; });
func< std::function<int(int)> >([](int x) -> int { return x + 40; });
```

```
main::{lambda(int)#3}
std::function<int (int)>
```

dzenanz (approved):

N-Dekker:
Thanks for reviewing and processing my pull request! For your information, I did a little benchmark and observed that replacing `std::function<void(const itk::ImageRegion&)>`  by a template argument `TFunction` significantly speeds up the runtime performance. Honestly my little benchmark was much simpler than those ParallelizeImageRegion member functions, but I think it still proves the point. https://gist.github.com/N-Dekker/e43033e03c86373864ed9992d9a3675c compares the following two functions:

```
  using FunctorType = std::function<void(const itk::ImageRegion<2>&)>;

  void CallStdFunction(FunctorType funcP, const itk::ImageRegion<2>& region)
  {
    funcP(region);
  }

  template <typename TFunction>
  void CallFunctionTemplate(TFunction funcP, const itk::ImageRegion<2>& region)
  {
    funcP(region);
  }
```
It appears that `CallFunctionTemplate` is about 3 times as fast as `CallStdFunction`, using Visual Sudio 2022 Release.

PR #4349, opened 2023-11-30
STYLE: Remove pre-C++17 definitions of static `constexpr` data members

N-Dekker:
According to C++17, "a `constexpr` static data member may be redundantly redeclared outside the class with no initializer", but then it says: "This usage is deprecated."

Cases found by Notepad++ regular expression `^constexpr .+::\w+.+;`

Follow-up to pull request #3985 commit 014ea433aaef9f931a24fba6a7a416b14473280a "STYLE: Remove pre-C++17 definition constexpr data member `SupportSize`"


dzenanz (approved):

hjmjohnson (approved):

thewtex (approved):
:100:

PR #4352, opened 2023-12-01
STYLE: Increase consistency in doc code block highlight language names

jhlegarreta:
Increase consistency in Markdown documentation code block highlight language names.

## PR Checklist


jhlegarreta:
Either way, it does not look like this makes any difference when having a look at the rendered documentation.

dzenanz (approved):
I guess that these are only longer-form specifications for the same thing.

jhlegarreta:
> I guess that these are only longer-form specifications for the same thing.

I am not sure Dženan; I do not know how they _should_ look like; e.g. either way, bash snippets are not highlighted at all, and I have not been able to find any official reference.

But at least we will be consistent, and if we discover the appropriate syntax, next time the search will be just for one word for each language.

hjmjohnson (approved):

PR #4360, opened 2023-12-03
STYLE: Update comments and parameter in the ImportImageContainer

### LINKED ISSUE #4359: itkImportImageContainer AllocateElements documentation doesn't seem right
This ( in [itkImportImageContainer.h](https://github.com/InsightSoftwareConsortium/ITK/blob/a6036a8b6f2e1f1f72c74094a1c22467cf1eec79/Modules/Core/Common/include/itkImportImageContainer.h#L164) ) doesn't seem right to me

```cpp
  /**
   * Allocates elements of the array.  If UseDefaultConstructor is true, then
   * the default constructor is used to initialize each element.  POD date types
   * initialize to zero.
   */
  virtual TElement *
  AllocateElements(ElementIdentifier size, bool UseDefaultConstructor = false) const;
```

_UseDefaultConstructor_ controls [this](https://github.com/InsightSoftwareConsortium/ITK/blob/a6036a8b6f2e1f1f72c74094a1c22467cf1eec79/Modules/Core/Common/include/itkImportImageContainer.hxx#L165):

```cpp
    if (UseDefaultConstructor)
    {
      data = new TElement[size](); // POD types initialized to 0, others use default constructor.
    }
    else
    {
      data = new TElement[size]; // Faster but uninitialized
    }
```

C++03 [value initialization](https://en.cppreference.com/w/cpp/language/value_initialization) with `()` is not about using a default constructor or not. A user-provided [default constructor](https://en.cppreference.com/w/cpp/language/default_constructor) **will be called in both cases**, of course.

The difference is for POD types (notation POD "plain old data" is, BTW, deprecated together with `std::is_pod`, actually it is `std::is_trivial<T>() && std::is_standard_layout<T>()` ). For such types, e.g. scalars, `()` will have effect of zero-initialization, the same as with C++11 `{}`.

```cpp
#include <iostream>

class A
{
public:
	A()
	{
		x = 0;
		std::cout << "A() ctor" << std::endl;
	}
	int x;
};

int main()
{
	A * a = new A[2];
	A * b = new A[2]();
	delete [] a;
	delete [] b;
	return 0;
}
```

```
r@deb2:~$ ./a.out 
A() ctor
A() ctor
A() ctor
A() ctor
```

It is e.g. related to RGB images
```
  using RGBPixelType = itk::RGBPixel<unsigned char>;
  using ImageType = itk::Image<RGBPixelType, 2>;
```

_itk::RGBPixel_ has a constructor `RGBPixel() { this->Fill(0); }` and
image->Allocate();
image->Allocate(true);
image->Allocate(false);
will make **no difference**, the default constructor will be always used (tested).

Again, for e.g. scalar types there is the effect of zero-initialization
`int * x = new int[100]();`

Any class with a use-provided default constructor, even the simplest one, is no longer _trivial_

```cpp
#include <iostream>
#include <type_traits>

class A
{
public:
  A() {}
  int x;
};

class B
{
public:
  int x;
};

class C
{
public:
  C() = default;
  int x;
};

int main()
{
  std::cout
    << std::is_standard_layout<A>() << " " << std::is_trivial<A>() << '\n'
    << std::is_standard_layout<B>() << " " << std::is_trivial<B>() << '\n'
    << std::is_standard_layout<C>() << " " << std::is_trivial<C>() << std::endl;
  return 0;
}
```

```
r@deb2:~$ ./a.out 
1 0
1 1
1 1
```

I would suggest to make STYLE changes, remove the part about the "default constructor" from the comment and rename the variable _UseDefaultConstructor_ to _UseValueInitialization_.

CC @N-Dekker 

**Edit:**

For completeness, in the above example  `A * a = new A[2]()` or `A * a = new A[2]{}` does **not** initialize `x`, no difference with `A * a = new A[2]`, but for B and C (trivial types)  `B * b = new B[2]()` or `B * b = new B[2]{}` does initialize `x` to 0. Tested with Valgrind.

issakomi:
Comments and the parameter _UseValueInitialization_ better describe the behavior of the
_ImportImageContainer_.
The same parameter in _VectorImage_ has also been updated.
Also removed comment from AllocateElements about 'new' and throwing
or not throwing an exception if allocation fails.
Closes #4359

### DIFF
--- a/Modules/Core/Common/include/itkImportImageContainer.h
+++ b/Modules/Core/Common/include/itkImportImageContainer.h
@@ -118,12 +118,12 @@ public:
    * container. However, in this particular case, Reserve as a Resize
    * semantics that is kept for backward compatibility reasons.
    *
-   * If UseDefaultConstructor is true, then * the default constructor is used
-   * to initialize each element.  POD date types initialize to zero.
+   * If UseValueInitialization is true, then POD types will be
+   * zero-initialized.
    *
    * \sa SetImportPointer() */
   void
-  Reserve(ElementIdentifier size, const bool UseDefaultConstructor = false);
+  Reserve(ElementIdentifier size, const bool UseValueInitialization = false);
 
   /** Tell the container to try to minimize its memory usage for
    * storage of the current number of elements.  If new memory is
@@ -162,12 +162,11 @@ protected:
   PrintSelf(std::ostream & os, Indent indent) const override;
 
   /**
-   * Allocates elements of the array.  If UseDefaultConstructor is true, then
-   * the default constructor is used to initialize each element.  POD date types
-   * initialize to zero.
+   * Allocates elements of the array.  If UseValueInitialization is true, then
+   * POD types will be zero-initialized.
    */
   virtual TElement *
-  AllocateElements(ElementIdentifier size, bool UseDefaultConstructor = false) const;
+  AllocateElements(ElementIdentifier size, bool UseValueInitialization = false) const;
 
   virtual void
   DeallocateManagedMemory();
--- a/Modules/Core/Common/include/itkImportImageContainer.hxx
+++ b/Modules/Core/Common/include/itkImportImageContainer.hxx
@@ -44,7 +44,7 @@ ImportImageContainer<TElementIdentifier, TElement>::~ImportImageContainer()
  */
 template <typename TElementIdentifier, typename TElement>
 void
-ImportImageContainer<TElementIdentifier, TElement>::Reserve(ElementIdentifier size, const bool UseDefaultConstructor)
+ImportImageContainer<TElementIdentifier, TElement>::Reserve(ElementIdentifier size, const bool UseValueInitialization)
 {
   // Reserve has a Resize semantics. We keep it that way for
   // backwards compatibility .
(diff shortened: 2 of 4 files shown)


Modules/Core/Common/include/itkImportImageContainer.h
```diff
@@ -162,12 +161,11 @@ class ITK_TEMPLATE_EXPORT ImportImageContainer : public Object
...
   /**
-   * Allocates elements of the array.  If UseDefaultConstructor is true, then
-   * the default constructor is used to initialize each element.  POD date types
-   * initialize to zero.
+   * Allocates elements of the array.  If UseInitValue is true, then POD
+   * types will be zero-initialized.
    */
   virtual TElement *
-  AllocateElements(ElementIdentifier size, bool UseDefaultConstructor = false) const;
+  AllocateElements(ElementIdentifier size, bool UseInitValue = false) const;
```
N-Dekker, line 168: Maybe `bool UseValueInitialization`? Or simply `bool ValueInitialize`? If we really want to be precise, I think we cannot avoid the technical C++ term "value-initialization" anyway. 🤷

N-Dekker:
Some very minor remarks (not a show-stopper), regarding the commit message:
- I'd rather say "parameter name" than "name of the variable"
- We have the convention to use the "imperative mood" in the subject line of a commit message (https://docs.itk.org/en/latest/contributing/index.html), so "STYLE: Update...", rather than "STYLE: Updated...". Again, no problem, just a convention.

issakomi:
> Good as-is, would be slightly better if Niels' comments were addressed. But that could be left for a follow-up PR.

Thanks! Yes, I shall address comments. I am setting temporary to draft.

dzenanz (approved):

N-Dekker (approved):
Thanks @issakomi, nice 👍  Also thanks for addressing my comments so carefully!

PR #4361, opened 2023-12-04
BUG: Wrap LBFGSOptimizerBaseHelperv4 before LBFGSOptimizerBasev4

thewtex:
LBFGSOptimizerBasev4 uses the LBFGSOptimizerBasev4 in its interface with the GetOptimizer method. Wrap LBFGSOptimizerBaseHelperv4 before LBFGSOptimizerBasev4 so LBFGSOptimizerBaseHelperv4 wrapper definition is available.

### DIFF
--- a/Modules/Numerics/Optimizersv4/wrapping/CMakeLists.txt
+++ b/Modules/Numerics/Optimizersv4/wrapping/CMakeLists.txt
@@ -1,8 +1,8 @@
 itk_wrap_module(ITKOptimizersv4)
 set(WRAPPER_SUBMODULE_ORDER
     itkVnlTypes
-    itkLBFGSOptimizerBasev4
     itkLBFGSOptimizerBaseHelperv4
+    itkLBFGSOptimizerBasev4
     itkLBFGSBOptimizerv4
     itkLBFGSOptimizerv4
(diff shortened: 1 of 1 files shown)


thewtex:
This is intended to address:

> ITKOptimizersv4Python: LNK2005 “public: class itk::LBFGSOptimizerBaseHelperv4 * __cdecl itk::LBFGSOptimizerBasev4::GetOptimizer(void)” already defined in itkLBFGSOptimizerBasev4Python.obj

reported: https://discourse.itk.org/t/itk-5-3-0-5-4rc-build-errors-shared-libs-python-wrappings-itkvtkglue-vs-2022-win11/6327

but I have not tested the impact on a shared library Windows build. It is possible that more is still needed to prevent implicit instantiation of LBFGSOptimizerBaseHelperv4. Regardless of whether additional steps are needed, this order should be changed.

dzenanz (approved):

sophonet:
Unfortunately, with this patch (changed order of itkLBFGSOptimizerBaseHelperv4 and itkLBFGSOptimiyerBasev4 in CMakeLists.txt) the error is still the same.

hjmjohnson (approved):

PR #4365, opened 2023-12-11
Use "constexpr if" in `MultiThreaderBase::ParallelizeImageRegionRestrictDirection`, and other style improvements

N-Dekker:
Style improvements `MultiThreaderBase::ParallelizeImageRegionRestrictDirection`:

 - Replaced an "if" with "constexpr if", allowing to simplify the initialization of `SplitDimension`
 
 - Removed the two `continue` statements, following [C++ Core Guidelines, "Minimize the use of `break`
and `continue` in loops"](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es77-minimize-the-use-of-break-and-continue-in-loops)
 
 - Replaced the local `splitRegion` variable with more lightweight `splitIndex` and `splitSize` variables
 
 - Replaced the default capture `[&]` of the local lambda with an explicit capture list, capturing `restrictedDirection` by value instead of by reference

### DIFF
--- a/Modules/Core/Common/include/itkMultiThreaderBase.h
+++ b/Modules/Core/Common/include/itkMultiThreaderBase.h
@@ -390,7 +390,7 @@ ITK_GCC_PRAGMA_DIAG_POP()
         SplitDimension,
         splitIndex.m_InternalArray,
         splitSize.m_InternalArray,
-        [&](const IndexValueType index[], const SizeValueType size[]) {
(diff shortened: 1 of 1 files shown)


dzenanz (approved):

hjmjohnson (approved):
Very nice!

N-Dekker:
Thanks, Hans!

PR #4367, opened 2023-12-12
ENH: `ImageRegion` support C++17 structured binding

N-Dekker:
Added `get()` member functions and specializations of `std::tuple_size` and `std::tuple_element` for ImageRegion, in order to support structured binding, for example:

    auto [index, size] = image.GetRequestedRegion();

### DIFF
--- a/Modules/Core/Common/include/itkImageRegion.h
+++ b/Modules/Core/Common/include/itkImageRegion.h
@@ -33,6 +33,8 @@
 #include "itkSize.h"
 #include "itkContinuousIndex.h"
 #include "itkMath.h"
+#include <type_traits> // For conditional and integral_constant.
+#include <utility>     // For tuple_element and tuple_size.
 
 // Macro added to each `ImageRegion` member function that overrides a virtual member function of `Region`. In the
 // future, `ImageRegion` will no longer inherit from `Region`, so then those `ImageRegion` member functions will no
@@ -357,6 +359,39 @@ public:
   SliceRegion
   Slice(const unsigned int dim) const;
 
+  /** Supports tuple-like access: `get<0>()` returns a reference to the index and `get<1>()` returns a reference to the
+   * size of the region. */
+  template <size_t VTupleIndex>
+  [[nodiscard]] auto &
+  get()
+  {
+    if constexpr (VTupleIndex == 0)
+    {
+      return m_Index;
+    }
+    else
+    {
+      static_assert(VTupleIndex == 1);
+      return m_Size;
+    }
+  }
+
+  /** Supports tuple-like access. Const overload. */
+  template <size_t VTupleIndex>
+  [[nodiscard]] const auto &
+  get() const
+  {
+    if constexpr (VTupleIndex == 0)
+    {
+      return m_Index;
+    }
+    else
+    {
+      static_assert(VTupleIndex == 1);
+      return m_Size;
+    }
+  }
+
 protected:
   /** Methods invoked by Print() to print information about the object
    * including superclasses. Typically not called by the user (use Print()
@@ -377,6 +412,35 @@ template <unsigned int VImageDimension>
 std::ostream &
 operator<<(std::ostream & os, const ImageRegion<VImageDimension> & region);
 } // end namespace itk
+
+
+namespace std
+{
+// NOLINTBEGIN(cert-dcl58-cpp)
+// Locally suppressed the following warning from Clang-Tidy (LLVM 17.0.1), as it appears undeserved.
+// > warning: modification of 'std' namespace can result in undefined behavior [cert-dcl58-cpp]
+
+/** `std::tuple_size` specialization, needed for ImageRegion to support C++ structured binding.
+ *
+ * Example, using structured binding to retrieve the index and size of a region:
+   \code
+    auto [index, size] = image.GetRequestedRegion();
+   \endcode
+ */
+template <unsigned int VImageDimension>
+struct tuple_size<itk::ImageRegion<VImageDimension>> : integral_constant<size_t, 2>
+{};
+
+/** `std::tuple_element` specialization, needed for ImageRegion to support C++ structured binding. */
+template <size_t VTupleIndex, unsigned int VImageDimension>
+struct tuple_element<VTupleIndex, itk::ImageRegion<VImageDimension>>
+  : conditional<VTupleIndex == 0, itk::Index<VImageDimension>, itk::Size<VImageDimension>>
+{
+  static_assert(VTupleIndex < tuple_size_v<itk::ImageRegion<VImageDimension>>);
+};
+
+// NOLINTEND(cert-dcl58-cpp)
+} // namespace std
 
 #undef itkRegionOverrideMacro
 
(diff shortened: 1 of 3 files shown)


dzenanz (approved):

N-Dekker:
@dzenanz Thanks for you approval! I think it would be pretty cool if `ImageRegion` would support structured binding. It feels a bit Pythonic 😃 

Unfortunately the code currently appears to trigger some undeserved clang-tidy warnings, saying _modification of 'std' namespace can result in undefined behavior [cert-dcl58-cpp]_, as I mentioned at https://github.com/llvm/llvm-project/issues/45454#issuecomment-1852420614  Even though the code appears correct, as confirmed by @Febbe at https://github.com/llvm/llvm-project/issues/45454#issuecomment-1852428326

Shall I locally disable this specific Clang-Tidy warning, by adding something like `//NOLINTNEXTLINE(cert-dcl58-cpp)` to those specific lines of code? Or shall we just leave it this way for now, as long as our CI is happy?

dzenanz:
Suppressing the warning sounds better to me.

Modules/Core/Common/include/itkImageRegion.h
```diff
@@ -378,6 +413,31 @@ std::ostream &
...
+
+/** `std::tuple_size` specialization, needed for ImageRegion to support C++ structured binding.
+ *
+ * Example, using structured binding to retrieve the index and size of a region:
+   \code
+    auto [index, size] = image.GetRequestedRegion();
+   \endcode
+ */
+template <unsigned int VImageDimension>
+struct tuple_size<itk::ImageRegion<VImageDimension>> : std::integral_constant<size_t, 2>
```
issakomi, line 428: I wonder if these `std::` are needed inside `namespace std`?

N-Dekker: @issakomi  Good catch 👍 These particular `std::` inside namespace `std` are not needed, and I don't even like them  😺 Will remove them with the next force-push

Modules/Core/Common/include/itkImageRegion.h
```diff
@@ -378,6 +413,35 @@ std::ostream &
 operator<<(std::ostream & os, const ImageRegion<VImageDimension> & region);
 } // end namespace itk
 
+
+namespace std
+{
+// NOLINTBEGIN(cert-dcl58-cpp)
+// Locally suppressed the following warning from Clang-Tidy (LLVM 17.0.1), as it appears undeserved.
+// > warning: modification of 'std' namespace can result in undefined behavior [cert-dcl58-cpp]
```
N-Dekker, line 421: @dave3d Hi Dave! Can you please tell me why Spell Checking Code Comments is complaining? It says at https://github.com/InsightSoftwareConsortium/ITK/actions/runs/7186016833/job/19570520532?pr=4367

>    Modules/Core/Common/include/itkImageRegion.h, 419
>
>   Modules/Core/Common/include/itkImageRegion.h, 421
>   Modules/Core/Common/include/itkImageRegion.h, 442
> 3 misspellings found

How can we see exactly which specific misspellings it refers to?

With this commit, I did already add `NOLINTBEGIN`, `NOLINTEND`, and `cert-dcl58-cpp` to ".github\workflows\itk_dict.txt". Shouldn't that already make the spell checker happy?

dave3d: I'm not sure.  I ran it on my computer and it passed OK.

dave3d: It seems like it could be using the old, non-updated dictionary?

N-Dekker: Thanks for your feedback @dave3d Indeed, the latest revision passes OK, after some trial and error! Apparently because I added `cert`, `dcl`, and `cpp` as three separate terms to the dictionary, instead of `cert-dcl58-cpp` as a whole! (This [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/2e9874271c6cfb4a4559d9bde61ac1cc6d7b053a..5cf540ef0759dbf37c97b8b4761b30423cc5815a) and this [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/5cf540ef0759dbf37c97b8b4761b30423cc5815a..c8c9983a697c97d5f032d57cd1656c15b6a64be4).)

Problem resolved 👍

N-Dekker:
@thewtex Although this is a rather small PR, I think it has a certain "cool factor", allowing users to do `auto [index, size] = region`, so maybe you can consider mentioning the feature explicitly in the next release notes  😃 At least, if it is still in time for ITK 5.4, of course!

Modules/Core/Common/include/itkImageRegion.h
N-Dekker: @dave3d Can you please have a look at the related CommentSpellCheck issues that I just submitted?
- https://github.com/SimpleITK/CommentSpellCheck/issues/53
- https://github.com/SimpleITK/CommentSpellCheck/issues/54

dave3d: OK, I can fix the code to do those things.

thewtex (approved):
@N-Dekker very, very cool! :sunglasses: 

Yes, I am making a mental note for the 5.4rc3 release notes.

issakomi:
I can run a scan today, it is OK to add the code like in `main`? BTW, I have added default ctor, without it I don't know how to use the class.

```
test.cpp: In function ‘int main()’:
test.cpp:26:16: error: no matching function for call to ‘String::String()’
   26 |         String s;
      |                ^
test.cpp:9:9: note: candidate: ‘String::String(const String&)’
    9 |         String(const String& arg)
      |         ^~~~~~
test.cpp:9:9: note:   candidate expects 1 argument, 0 provided
```

```cpp
#include <iostream>
#include <memory>
#include <cstring>

class String
{
public:
	String() {}
	String(const String& arg)
		:
		m_ptr(new char[strlen(arg.m_ptr.get()) + 1])
	{
		strcpy(m_ptr.get(), arg.m_ptr.get());
	}
	std::unique_ptr<char[]> m_ptr;
};

void Check(const String& arg)
{
	const auto copyOfString = arg;
	std::cout << copyOfString.m_ptr.get() << std::endl;
}

int main()
{
	String s;
	s.m_ptr = std::unique_ptr<char[]>(new char[4]{'a','b','c','\0'});
	Check(s);
	return 0;
}
```

hjmjohnson (approved):

dzenanz:
Did this PR cause these compile errors:
https://open.cdash.org/viewBuildError.php?buildid=9269183

issakomi:
> Did this PR cause these compile errors: https://open.cdash.org/viewBuildError.php?buildid=9269183


static_asserts are from #4373. BTW users add rather often `-DITK_FUTURE_LEGACY_REMOVE:BOOL=ON` too, even if it is for developers only, e.g. [here](https://discourse.itk.org/t/itkmathematicalmorphologyenums-h-compile-error-with-gcc-13-on-linux/6377).

Issue #513, opened 2019-02-16
Overload the stream insertion operator for types that are missing it


jhlegarreta:
### Description
Overload the stream insertion operator for types that are missing it.

### Expected behavior
To be able to print any data structure with a single line, e.g.:
```
os << indent << "MyStructure : " << m_MyStructure << std::endl;
```
### Actual behavior
Developer's are forced to iterate over such structures to get them printed (usually in the `PrintSelf` method), e.g.:
```
  while( it != m_CirclesList.end() )
    {
    os << indent << "[" << i << "]: " << *it << std::endl;
    ++it;
    ++i;
    }
```

### Additional information
Some fundamental ITK types (e.g. `IndexTypes`, `SizeType`, `SpacingType`, `OffsetType`,  `DirectionType`, `OffsetTable`, `RegionType`) either seem not to have such an overload or it is just because some filters (`itkImageConstIteratorWithIndex.h`, `itkImageIORegion.h`, `itkNeighborhood.h`,  `itkImportImageFilter.h`, `itkConstNeighborhoodIterator.h`, `itkImportImageFilter.hxx`, `itkDisplacementFieldJacobianDeterminantFilter.h`, `itkGaussianBlurImageFunction.h`, `itkEllipsoidInteriorExteriorSpatialFunction.h`) that declare ivars of such types are not using the correct types, since it looks weird that such fundamental types do not have the stream insertion operator overload.

Other more exotic type aliases (`GradientType`, `StrideTable`, `SeedsContainerType`, `WeightsType`, `ErrorArrayType`, `ExtentArrayType`) would also deserve some investigation, as well as other aliases that are not consistent with the toolkit convention for the intended use (e.g. `InputType` in `itkEllipsoidInteriorExteriorSpatialFunction.h`).

Finally, other data structures that are also missing such an overload are regular arrays, e.g.
```
double m_Sigma[ImageDimension2];
```
in `itkBSplineControlPointImageFunction.h` or `itkGaussianDerivativeImageFuncton.h`

or
```
double          m_Scale[NDimensions];
``` 
in `itkScalableAffineTransform.h`.

`std::list` types (e.g. `itkHoughTransform2DCirclesImageFilter.h` or `itkHoughTransform2DLinesImageFilter.h`).

Or `vnl_matrix` matrix (e.g. `itkBSplineScatteredDataPointSetToImageFilter.h`).

This is related to #512.

ChristianJacobsen:
[...]

PR #4368, opened 2023-12-14
BUG: Fix printing of values in MetaDataObject

### LINKED ISSUE #1454: MetaDataDictionary::Print only prints [UNKNOWN_PRINT_CHARACTERISTICS]
### Description

For any MetaDataObject calling `obj->Print(std::cout)` produces "[UNKNOWN_PRINT_CHARACTERISTICS]" on the console. 

https://github.com/InsightSoftwareConsortium/ITK/blob/60469760ee13f48b265ca7424e3e9fd652672710/Modules/Core/Common/include/itkMetaDataObject.hxx#L63-L69

### Expected behavior

For `itk::Object` base objects the `Print` method should be used while those which support `operator<<`, that method should be used.

It is expected to be able to iterate over the metadata dictionary and print the key and value of all types and get descriptive output.

phcerdan:
Also removes dead code related to METADATAPRINT

Fix #1454

## PR Checklist
- [NA] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5

### DIFF
--- a/Modules/Core/Common/include/itkMetaDataObject.hxx
+++ b/Modules/Core/Common/include/itkMetaDataObject.hxx
@@ -28,9 +28,79 @@
 #ifndef itkMetaDataObject_hxx
 #define itkMetaDataObject_hxx
 
+#include <type_traits>
+#include <iterator>
+
+template <typename T, typename = void>
+inline constexpr bool is_iterable_print_v = false;
+// Specialize for std::vector<T> and std::vector<std::vector<T>>
+template <typename T>
+inline constexpr bool is_iterable_print_v<std::vector<T>, std::void_t<>> = true;
+template <typename T>
+inline constexpr bool is_iterable_print_v<std::vector<std::vector<T>>, std::void_t<>> = true;
 
 namespace itk
 {
+template <typename TIterable>
+void
+printIterable(std::ostream & os, const TIterable & iterable)
+{
+  if constexpr (is_iterable_print_v<TIterable>)
+  {
+    os << "[";
+    auto begin = std::begin(iterable);
+    auto end = std::end(iterable);
+    for (auto it = begin; it != end; ++it)
+    {
+      if (it != begin)
+      {
+        os << ", ";
+      }
+      printIterable(os, *it);
+    }
+    os << "]";
+  }
+  else
+  {
+    // Handle non-iterable types
+    os << iterable;
+  }
+}
+
+template <typename MetaDataObjectType>
+void
+MetaDataObject<MetaDataObjectType>::PrintValue(std::ostream & os) const
+{
+  if constexpr (is_iterable_print_v<MetaDataObjectType>)
+  {
+    os << "[";
+    auto begin = std::begin(m_MetaDataObjectValue);
+    auto end = std::end(m_MetaDataObjectValue);
+    for (auto it = begin; it != end; ++it)
+    {
+      if (it != begin)
+      {
+        os << ", ";
+      }
+      printIterable(os, *it);
+    }
+    os << "]";
+  }
+  else
+  {
+    os << m_MetaDataObjectValue;
+  }
+}
+
+template <typename MetaDataObjectType>
+void
+MetaDataObject<MetaDataObjectType>::PrintSelf(std::ostream & os, Indent indent) const
+{
+  os << indent;
+  this->PrintValue(os);
+  os << std::endl;
+}
+
 template <typename MetaDataObjectType>
 const char *
 MetaDataObject<MetaDataObjectType>::GetMetaDataObjectTypeName() const
@@ -59,12 +129,6 @@ MetaDataObject<MetaDataObjectType>::SetMetaDataObjectValue(const MetaDataObjectT
   Self::Assign(m_MetaDataObjectValue, newValue);
 }
 
-template <typename MetaDataObjectType>
-void
-MetaDataObject<MetaDataObjectType>::Print(std::ostream & os) const
-{
-  Superclass::Print(os);
-}
 
 } // end namespace itk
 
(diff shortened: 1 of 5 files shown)


Modules/Core/Common/include/itkMetaDataObject.hxx
```diff
@@ -28,9 +28,79 @@
...
+template <typename T>
+inline constexpr bool is_iterable_print_v<std::vector<T>, std::void_t<>> = true;
+template <typename T>
+inline constexpr bool is_iterable_print_v<std::vector<std::vector<T>>, std::void_t<>> = true;
 
 namespace itk
 {
+template <typename TIterable>
+void
+printIterable(std::ostream & os, const TIterable & iterable)
```
jhlegarreta, line 46: Oh, this reminds me of https://github.com/InsightSoftwareConsortium/ITK/issues/513#issue-411127172

phcerdan: Maybe Iterable is not the best name here. I started the implementation to be general to any iterable,
but then I realized that there are quite a few iterables that are better printed as one element instead of a "list" of elements (std::string).

At the end I decided to only provide the specialization for `std::vector<T>` and `std::vector<std::vector<T>>` because those two are the only two cases explicitly instantiated that don't have `operator<<` defined.

jhlegarreta: OK. Still saves typing and provides consistency, so would be nice to introduce it across the code base as time permits. If an overload is ever implemented, only one line would need to be substituted for each match in that case.

N-Dekker, line 46: Interesting pull request, thanks @phcerdan 

Can you possibly hide `printIterable` (or whatever its name will be) from the public API? Because it isn't intended to be exposed directly to the end user, right? I guess it should be possible to make it a private (preferably `static`) member function of `MetaDataObject`. Otherwise, it might be possible to define it as a lambda, hidden locally inside `MetaDataObject::PrintValue`, but that seems more challenging to do.

Modules/Core/Common/include/itkMetaDataObject.hxx
```diff
@@ -28,9 +28,79 @@
...
+#include <type_traits>
+#include <iterator>
+
+template <typename T, typename = void>
+inline constexpr bool is_iterable_print_v = false;
+// Specialize for std::vector<T> and std::vector<std::vector<T>>
+template <typename T>
+inline constexpr bool is_iterable_print_v<std::vector<T>, std::void_t<>> = true;
+template <typename T>
+inline constexpr bool is_iterable_print_v<std::vector<std::vector<T>>, std::void_t<>> = true;
```
N-Dekker, line 40: Nice to see a use case for variable templates, thanks @phcerdan. Unfortunately "itkMetaDataObject.hxx" is usually `#include`'d (indirectly) by end users, and I guess it's not the intention to add `is_iterable_print_v` to the global namespace of the user, right? If you agree, you may possibly make `is_iterable_print_v` a private static member of `MetaDataObject`. Or otherwise, maybe it's simpler to just add overloaded private static `MetaDataObject::IsIterablePrint(const T&)` member functions. What do you think?

phcerdan: Good points @N-Dekker, I will make static private members

phcerdan: GCC doesn't like it... (moving the to static inline inside the class).
https://stackoverflow.com/questions/72190700/explicit-template-argument-list-not-allowed-with-g-but-compiles-with-clang

I will explore further.

phcerdan: It seems fixed in gcc14, but we cannot afford that.

N-Dekker: Another option I guess would be to have constexpr static member functions `IsIterablePrint(const T&)` overloaded (rather than "specialized") for `std::vector<T>` and `std::vector<std::vector<T>>`, right?

N-Dekker: Regarding the name of the boolean, maybe something like `ShouldBePrintedElementWise` ?

Modules/Core/Common/include/itkMetaDataObject.hxx
N-Dekker: If you can make it private, you might as well just name the function something like `PrintHelper`, rather than `printIterable`.  My 2 cent!

hjmjohnson (commented):
I am very supportive of these efforts!  Modern C++ can accomplish the desired task much more cleanly and robustly.

I appreciate @N-Dekker comments to fortify the API (avoid leaky API) and would like to see those comments addressed before finalizing this PR.

Modules/Core/Common/include/itkMetaDataObject.hxx
jhlegarreta: Sorry to chime in again Pablo. Went through the rest of the comments, especially those where Niels proposes this to be a private static method, and I realize that we will not be able to use this in other classes' `PrintSelf` method. So linked to https://github.com/InsightSoftwareConsortium/ITK/pull/4368/files#r1426050481, wouldn't it be better to make this method dwell in a common place where it can be used for that purpose?

phcerdan:
This PR had two features:
Fix printing of values in MetaDataObject. Fixed in #4814 
Allow printing of std::vector<std::vector<T>> and other iterables. This part is being tackled in #4831.

PR #4370, opened 2023-12-15
Add `MultiThreaderBase::SetSingleMethodAndExecute` and replace `SetSingleMethod`, `SingleMethodExecute` calls

N-Dekker:
A `SetSingleMethod` call is very commonly followed by a `SingleMethodExecute` call. The proposed `SetSingleMethodAndExecute` member function calls them both, as a convenience to the user.


hjmjohnson (approved):

N-Dekker:
Thanks for your approval @hjmjohnson  Looking at my PR again, I wonder, should it still be named a _Single_ method? 

It seems to me that the word "Single" was there, to clearly distinguish "Single" versus "Multiple" methods, but "Multiple method execution" is already deprecated: 
https://github.com/InsightSoftwareConsortium/ITK/blob/e186d876b3ce43b503eddfb0d8a3de448a932151/Modules/Core/Common/include/itkPlatformMultiThreader.h#L115 

So then, should we also deprecate the word "Single" in this context? Meaning:

- prefer `SetMethod` over `SetSingleMethod`
- prefer `MethodExecute` over `SingleMethodExecute`

?

hjmjohnson:
> Thanks for your approval @hjmjohnson Looking at my PR again, I wonder, should it still be named a _Single_ method?
> 
> It seems to me that the word "Single" was there, to clearly distinguish "Single" versus "Multiple" methods, but "Multiple method execution" is already deprecated:
> 
> https://github.com/InsightSoftwareConsortium/ITK/blob/e186d876b3ce43b503eddfb0d8a3de448a932151/Modules/Core/Common/include/itkPlatformMultiThreader.h#L115
> 
> So then, should we also deprecate the word "Single" in this context? Meaning:
> 
> * prefer `SetMethod` over `SetSingleMethod`
> * prefer `MethodExecute` over `SingleMethodExecute`
> 
> ?

@N-Dekker  I'm OK with creating a new set of preferred names without 'Single' and deprecating the 'Single' function variants.

N-Dekker:
Thanks for your comment @hjmjohnson  I just marked this pull request as "draft", as I'm now indeed considering to remove (or "FUTURE_LEGACY_REMOVE") the term "Single" from those member functions.

N-Dekker:
For the sake of time, I decided not to make a pull request (yet) to remove the word "Single" from `SetSingleMethod` and `SingleMethodExecute`. Instead, I just stick to the current pull request for now, to just add the convenience member function `SetSingleMethodAndExecute`.

dzenanz (approved):

thewtex (approved):

PR #4373, opened 2023-12-19
Replace `itkTypeMacroNoParent`, `itkTypeMacro` with `itkVirtualGetNameOfClassMacro`, `itkOverrideGetNameOfClassMacro`

N-Dekker:
Added two new macro's, `itkVirtualGetNameOfClassMacro` and `itkOverrideGetNameOfClassMacro`, replacing the old `itkTypeMacroNoParent` and `itkTypeMacro`.

The main aim is to be clearer about what those macro's do: add a virtual `GetNameOfClass()` member function and override it. Unlike `itkTypeMacro`, `itkOverrideGetNameOfClassMacro` does not have a `superclass` parameter, as it was not used anyway.

Note that originally `itkTypeMacro` did not use its `superclass` parameter either, looking at commit 699b66cb04d410e555656828e8892107add38ccb, Will Schroeder (@wschroed), June 27, 2001:
https://github.com/InsightSoftwareConsortium/ITK/blob/699b66cb04d410e555656828e8892107add38ccb/Code/Common/itkMacro.h#L331-L337

Deprecated (`ITK_FUTURE_LEGACY_REMOVE`) the old macro's.


dzenanz (approved):
Looks good on a glance.

issakomi:
> Deprecated (ITK_FUTURE_LEGACY_REMOVE) the old macro's.

Wouldn't it be better to keep the old macros for compatibility? The gain is minimal, but a lot of user code will be broken.

N-Dekker:
> > Deprecated (ITK_FUTURE_LEGACY_REMOVE) the old macro's.

> Wouldn't it be better to keep the old macros for compatibility? The gain is minimal, but a lot of user code will be broken.

Thanks @issakomi, but on the other hand, with `ITK_FUTURE_LEGACY_REMOVE`, the old macro's will still be around for a while. Otherwise, do we need a new macro to specify that some feature may be removed in the _far_ future? `ITK_FAR_FUTURE_LEGACY_REMOVE`? Or maybe `ITK_UNRECOMMENDED`? Informing the user that those old macro's are no longer preferable.

Anyway, I find the name `itkTypeMacroNoParent` hard to understand.  And I find it cumbersome that `itkTypeMacro(thisClass,superclass)` requires specifying the superclass, for nothing. So I think it's worth dropping the old macro's eventually.

dzenanz:
We could also keep this legacy macro around for more than one version, because it is so widespread.

issakomi:
IMHO, some code will just fail to compile and the user will have to search for solution frightened. Maybe give an advice what to do like here.

```
#if !defined(ITK_FUTURE_LEGACY_REMOVE)
#  define ITK_DISALLOW_COPY_AND_ASSIGN(TypeName) ITK_DISALLOW_COPY_AND_MOVE(TypeName)
#else
#  define ITK_DISALLOW_COPY_AND_ASSIGN(TypeName) \
    static_assert(false, "Replace deprecated ITK_DISALLOW_COPY_AND_ASSIGN with modern ITK_DISALLOW_COPY_AND_MOVE")
#endif
```

BTW, above code is also just renaming with no much gain, but at least with a hint.

N-Dekker:
@issakomi Please check this [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/910367ac51c718dc8c9008d755c2d0c1fbef0e47..6c4d6ac1f93afaaacdd02fd2266c5ef203e6b5fc), hoping to address your comment from https://github.com/InsightSoftwareConsortium/ITK/pull/4373#issuecomment-1862977656

N-Dekker:
@dzenanz If it's preferable to postpone the "future removal" of the old macro's, I would still very much like to see the new macro's being added. Hope that's still feasible!

dzenanz (approved):

dzenanz:
This PR is good as is. Just next time we do a round of deprecations, skip this change from advancing to "regular deprecated", and keep it as future legacy.

hjmjohnson (approved):

PR #4379, opened 2023-12-27
WIP: STYLE: Use ImageRegion `auto [index, size] = region` structured bindings

N-Dekker:
Replaced initializations of the form:

    SizeType size = region.GetSize();
    IndexType index = region.GetIndex();

By using structured binding of the form `auto [index, size] = region`, as was introduced by pull request https://github.com/InsightSoftwareConsortium/ITK/pull/4367 commit 72aa9a602bd92b665acc915d8df4bd00a27c98f0 "ENH: `ImageRegion` support C++17 structured binding"

Did so by Notepad++ Replace in Files, using regular expressions:

    Find what: ^([ ]+)const (.*)\w.+[ ]+(\w+) = (.+)\.GetIndex\(\);\r\n\1const \2\w+[ ]+(\w+) = \4\.GetSize\(\);$
    Replace with: $1const auto [$3, $5] = $4;

    Find what: ^([ ]+)const (.*)\w.+[ ]+(\w+) = (.+)\.GetSize\(\);\r\n\1const \2\w+[ ]+(\w+) = \4\.GetIndex\(\);$
    Replace with: $1const auto [$5, $3] = $4;

    Find what: ^([ ]+)(.*)\w.+[ ]+(\w+) = (.+)\.GetIndex\(\);\r\n\1\2\w+[ ]+(\w+) = \4\.GetSize\(\);$
    Replace with: $1auto [$3, $5] = $4;

    Find what: ^([ ]+)(.*)\w.+[ ]+(\w+) = (.+)\.GetSize\(\);\r\n\1\2\w+[ ]+(\w+) = \4\.GetIndex\(\);$
    Replace with: $1auto [$5, $3] = $4;

Manually added `const` whenever the values of the bindings were not modified, excluded `ImageIORegion`, and removed `Index` and `Size` type aliases that are no longer used after this commit.

----
When reviewing, it may be helpful to ignore whitespace changes: https://github.com/InsightSoftwareConsortium/ITK/pull/4379/files?w=1

### DIFF
--- a/Modules/Bridge/VTK/include/itkVTKImageExport.hxx
+++ b/Modules/Bridge/VTK/include/itkVTKImageExport.hxx
@@ -145,7 +145,7 @@ VTKImageExport<TInputImage>::WholeExtentCallback()
   }
 
   InputRegionType region = input->GetLargestPossibleRegion();
-  auto [index, size] = region;
+  const auto [index, size] = region;
 
   unsigned int i = 0;
   // Fill in the known portion of the extent.
@@ -386,7 +386,7 @@ VTKImageExport<TInputImage>::DataExtentCallback()
   }
 
   InputRegionType region = input->GetBufferedRegion();
-  auto [index, size] = region;
+  const auto [index, size] = region;
 
   unsigned int i = 0;
   for (; i < InputImageDimension; ++i)
--- a/Modules/Bridge/VTK/include/itkVTKImageExport.hxx
+++ b/Modules/Bridge/VTK/include/itkVTKImageExport.hxx
@@ -144,8 +144,7 @@ VTKImageExport<TInputImage>::WholeExtentCallback()
     itkExceptionMacro("Need to set an input");
   }
 
-  InputRegionType region = input->GetLargestPossibleRegion();
-  const auto [index, size] = region;
+  const auto [index, size] = input->GetLargestPossibleRegion();
 
   unsigned int i = 0;
   // Fill in the known portion of the extent.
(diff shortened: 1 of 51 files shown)


dzenanz (approved):
Looks good. Remove now unused type definitions to get rid of those warnings for a clean CI.

N-Dekker:
FYI, I'm considering to postpone this pull request, for the moment. I think it's a nice "proof of principle" on how to use structured bindings with `ImageRegion`. It looks nice. And it does significantly reduce the amount of boilerplate code. But it appears that the PR will trigger some warnings that I find hard to deal with! Like Microsoft's [lnt-accidental-copy](https://learn.microsoft.com/en-us/cpp/ide/lnt-accidental-copy), _"`auto` doesn't deduce references. A possibly unintended copy is being made"_ and Coverity's  "AUTO_CAUSES_COPY".

For example, the following line may trigger such a warning (lnt-accidental-copy, AUTO_CAUSES_COPY):

    const auto [index, size] = region; // copying

The warning would suggest doing the structured bindings "by reference" instead:

    const auto & [index, size] = region; // referencing

But I do not blindly want to follow their suggestion to replace copying with referencing. For a rather small object like `ImageRegion`, copying is pretty fast already, and it will be even faster when `ImageRegion` has become trivially copyable (pull request #4344). Whereas referencing potentially makes the code more complicated, as it will potentially introduce more dependencies between variables.

I'm considering a workaround for those warnings, but I'm not sure yet. 🤷

N-Dekker:
Basically I think this pull request is OK, but if it gets merged, we might still get into a discussion on `const auto [index, size] = region` (copy) versus `const auto & [index, size] = region` (reference). I do prefer copying, because an ImageRegion is quite a lightweight object, but my position in favor of copying will be stronger once `itk::ImageRegion` has become trivially-copyable. 

`itk::ImageRegion` should be trivially-copyable in the future: pull request #4344 commit 1f3bbb658698556ff0d437c4effb9c1e42acc7c2  So maybe we can reconsider using the `const auto [index, size] = region` syntax _in the future_! For the time being, I'll just close this pull request.

PR #4383, opened 2024-01-03
ENH: Add `itk::Copy(const T & original)`, which simply returns a copy

N-Dekker:
Intended to be used to make the act of copying explicit, and to avoid warnings like:

 - Microsoft Visual Studio 2022, IntelliSense Code Linter [lnt-accidental-copy] "`auto` doesn't deduce references. A possibly unintended copy is being made" (https://learn.microsoft.com/en-us/cpp/ide/lnt-accidental-copy)
 - Microsoft Visual Studio 2022, Warning C26820, "This is a potentially expensive copy operation. Consider using a reference unless a copy is required (p.9)" (https://learn.microsoft.com/en-us/cpp/code-quality/c26820?view=msvc-170)
 - Synopsys Coverity 2023.6.2, "CID 331225:  Performance inefficiencies  (AUTO_CAUSES_COPY)", "Using the `auto` keyword without an `&` causes the copy of an object...".
 - SonarSource static code analysis, "Unnecessary expensive copy should be avoided when using auto as a placeholder type", https://rules.sonarsource.com/cpp/RSPEC-6032/

### DIFF
--- a/Modules/Core/Common/include/itkCopy.h
+++ b/Modules/Core/Common/include/itkCopy.h
@@ -1,0 +1,59 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         https://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+#ifndef itkCopy_h
+#define itkCopy_h
+
+namespace itk
+{
+
+/** Returns a copy of its argument. Primarily used to make the act of copying (typically involving a copy-constructor
+ * call) explicit, and to avoid warnings like:
+ *
+ * - Microsoft Visual Studio 2022, IntelliSense Code Linter [lnt-accidental-copy] "`auto` doesn't deduce references. A
+ * possibly unintended copy is being made" (https://learn.microsoft.com/en-us/cpp/ide/lnt-accidental-copy)
+ * - Microsoft Visual Studio 2022, Warning C26820, "This is a potentially expensive copy operation. Consider using a
+ * reference unless a copy is required (p.9)" (https://learn.microsoft.com/en-us/cpp/code-quality/c26820?view=msvc-170)
+ * - Synopsys Coverity 2023.6.2, "CID 331225:  Performance inefficiencies  (AUTO_CAUSES_COPY)", "Using the `auto`
+ * keyword without an `&` causes the copy of an object...".
+ * - SonarSource static code analysis, "Unnecessary expensive copy should be avoided when using `auto` as a placeholder
+ * type", "Avoid this unnecessary copy by using a "const" reference. (cpp:S6032)",
+ * https://rules.sonarsource.com/cpp/RSPEC-6032/
+ *
+ * Example:
+   \code
+     const auto possiblyUnintendedCopy = image->GetRequestedRegion(); // Warning!
+     const auto intendedCopy = Copy(image->GetRequestedRegion());     // OK, no warning  :-)
+     const auto [index, size] = Copy(image->GetRequestedRegion());    // OK, no warning  :-)
+   \endcode
+ *
+ * \note In general, it is up to the programmer to choose between doing a copy and using a reference, of course. `Copy`
+ should only be used when doing a copy is preferred over using a reference.
+ */
+template <typename T>
+[[nodiscard]] constexpr T
+Copy(const T & original)
+{
+  // May call the copy-constructor of `T`.
+  return original;
+}
+
+} // namespace itk
+
+
+#endif // itkCopy_h
--- a/Modules/Core/Common/test/itkCopyGTest.cxx
+++ b/Modules/Core/Common/test/itkCopyGTest.cxx
@@ -1,0 +1,116 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         https://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+// First include the header file to be tested:
+#include "itkCopy.h"
+#include <gtest/gtest.h>
+
+#include <array>
+#include <cstdint> // For intmax_t.
+#include <random>
+#include <utility> // For pair.
+#include <vector>
+
+
+// Check compile-time evaluation:
+static_assert(itk::Copy(0) == 0);
+static_assert(itk::Copy(1) == 1);
+
+namespace
+{
+// Returns a (const) reference to the specified argument.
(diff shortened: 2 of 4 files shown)


dzenanz (approved):

Modules/Core/Common/include/itkCopy.h
```diff
@@ -0,0 +1,46 @@
...
+ *
+ *=========================================================================*/
+
+#ifndef itkCopy_h
+#define itkCopy_h
+
+namespace itk
+{
+
+/** Returns a copy of its argument. Primarily used to make the act of copying (typically involving a copy-constructor
```
dzenanz, line 25: Spelling error somewhere on this line?

N-Dekker: Maybe @dave3d can tell us? 🤷

Modules/Core/Common/test/itkCopyGTest.cxx
```diff
@@ -0,0 +1,67 @@
...
+  };
+
+  check(42);
+  check(std::vector<std::intmax_t>{ 1, 2, 3, 4 });
+  check(std::array<std::intmax_t, 4>{ 1, 2, 3, 4 });
+}
+
+
+// Tests that `itk::Copy` allows storing a copy of a value in a variable declared by `auto`, without a warning.
+TEST(Copy, AllowsUsingAutoKeywordCopyingWithoutWarnings)
```
N-Dekker, line 48: @issakomi Can you please tell us if the proposed `itk::Copy` would indeed silence those Coverity AUTO_CAUSES_COPY warnings? Specifically, is this `AllowsUsingAutoKeywordCopyingWithoutWarnings` unit test indeed warning-free? And do those AUTO_CAUSES_COPY appear when the `itk::Copy` calls in there are removed?

Modules/Core/Common/include/itkCopy.h
issakomi: Might be:
Coverity, Intellisense, Synopsys, lnt

Modules/Core/Common/test/itkCopyGTest.cxx
issakomi: I will try. BTW, I don't run a Coverity scan for the entire ITK. I scan my applications from time to time and some headers from ITK are also scanned if they are enabled somehow. I'll try to reproduce it. IMHO we could just ignore the warnings.

Modules/Core/Common/include/itkCopy.h
N-Dekker: Thanks @issakomi, Now I see, those four terms are mentioned in the CI 
output, at https://github.com/InsightSoftwareConsortium/ITK/actions/runs/7400527348/job/20134312236

```

    Modules/Core/Common/include/itkCopy.h, 25
    Modules/Core/Common/include/itkCopy.h, 25
    Modules/Core/Common/include/itkCopy.h, 25
    Modules/Core/Common/include/itkCopy.h, 25

Coverity:

Intellisense:

Synopsys:

lnt:

4 misspellings found
```

I hope Dave (@dave3d) can still make the CommentSpellCheck output more readable. I would think that `Modules/Core/Common/include/itkCopy.h, 25` is about line 25 in that source file, but apparently it's not, or is it? It's not clear to me either why those misspellings end with a colon (`:`). What that intentional?

Anyway, I'll add those to the dictionary file, https://github.com/InsightSoftwareConsortium/ITK/blob/d45127b5a3296b7f1721b57e0ffe8e0071d0c64f/.github/workflows/itk_dict.txt

Modules/Core/Common/test/itkCopyGTest.cxx
N-Dekker: @issakomi For a specific application, I think those auto-causes-copy warnings can usually be ignored safely. But I think there is a policy for ITK to address warnings as much as possible. And it's also harder to ignore these specific warnings now that they come from at least three different tools (Intellisense/Linter, MSVC compiler and Coverity).

N-Dekker: @issakomi I just extended the `AllowsUsingAutoKeywordCopyingWithoutWarnings` to include structured bindings. Hope you can also try that out on Coverty. Specifically:

```cpp
  auto originalPair =
    std::make_pair(std::vector<std::intmax_t>{ 1, 2, 3, 4 }, std::array<std::intmax_t, 4>{ 1, 2, 3, 4 });
  const auto [first, second] = itk::Copy(originalPair);
```

Does the `itk::Copy` call here suppress another AUTO_CAUSES_COPY warning? Would the AUTO_CAUSES_COPY warning occur without the `itk::Copy` call (when simply doing `const auto [first, second] = originalPair`)?

I did already observe that the `itk::Copy` call in the context of structured bindings suppresses such warnings from Intellisense/Linter.

Modules/Core/Common/include/itkCopy.h
dave3d: I'll check the code out today.  Also I plan to improve the spell checking output.  Probably next week.

Modules/Core/Common/test/itkCopyGTest.cxx
issakomi: > And it's also harder to ignore these specific warnings now that they come from at least three different tools

What I mean is that AUTO_CAUSES_COPY can be skipped if the object is small and trivial to copy. I have no experience with IntelliSense Code Linter. Coverity promises to suggest 'auto &' only if possible. The new _Copy_ function is a way to silence the warning, IMHO, it's the same thing as ignoring it.
BTW, maybe just create a test project for Coverity tests? Any opensource project can be scanned at https://scan.coverity.com. A full ITK scan will generate a ton of defects, mostly from third party modules.

Modules/Core/Common/include/itkCopy.h
N-Dekker: @dave3d Cool, thanks!

Modules/Core/Common/test/itkCopyGTest.cxx
issakomi: If necessary, please create a small test project for Coverity tests. Thank you.

N-Dekker: Thanks for doing those Coverity tests, @issakomi  In the mean time, I did further improve the unit tests. Especially by introducing a local function `GetReference` (just for the tests). Because I realized that those warnings appear specifically when there is a _reference_ at the right hand side (rhs). And that's exactly when `itk::Copy(rhs)` should be useful.

Thanks for suggesting my own Coverity tests, I think I'll do so later. But I think the MSVC warnings are quite similar, and I already have MSVC installed 😃.

N-Dekker:
For your information, a similar warning is produced by SonarSource, https://rules.sonarsource.com/cpp/RSPEC-6032/ as mentioned by Dvir Yitzchaki at the ACCU mailing list. Dvir posted an example at https://godbolt.org/z/ETdsYT3Ph

Modules/Core/Common/test/itkCopyGTest.cxx
issakomi: The reason for the suggestion of a special repository for _Coverity_ tests is that I just push a code somewhere in _main.cpp_ of my app and upload to _Coverity_ scan (scans are limited).

It is rather easy (github login can be used for _scan.coverity.com_), but a project have to be approved.

Short example on Linux:
 - download _Coverity_ archive, unpack in home directory
 - `export PATH=/home/r/cov-analysis-linux64-2023.6.2/bin:$PATH`
 - configure something like _mypoject/coverity-build_ and `cd mypoject/coverity-build`
 - `cov-build --dir cov-int make -j6`
 - `tar cf cov-int.tar cov-int`
 - `gzip cov-int.tar`
 - upload cov-int.tar.gz to scan portal (AFAIK, the name must always be _cov-int_)
 - wait for result of the scan

There are many option for automation, of course. Also a good idea may be to use "system" libraries  to reduce the number of defects (some parts of code can be skipped, but not really, the defects are not in the database, but seem to be still counted). There is also an option of "modeling" to suppress false-positives, but it seems to be a little over-complicated.

Probably commercial customers can do scan locally, not sure. Their portal might be a little better, IMHO, but the scan itself seems to be very good, not so nasty as many others and not so many false-positives (but there will be some of them, caution with silencing is a good idea, otherwise it is easy to make thing worse, as it happens very often with silencing of warnings from all the tools).

N-Dekker:
I think this PR is ready, only docs/readthedocs.org:itk is still "Pending — Read the Docs build is in progress!"

dzenanz (approved):

Issue #4388, opened 2024-01-09
Unable to load images with Umlauts in filepath (ITK / SimpleITK)


Pfleiderer-Adrian:
**Bug Description:**
I encountered an issue while attempting to load images using the ITK / SimpleITK framework in my Python script. The traceback indicates that the problem arises when attempting to create an IO object for reading an image file that contains Umlauts (ä, ü, ö) in its path. The relevant part of the traceback is as follows:

```
...
  File [...]\utils\prepare_data.py", line 202, in convert_mha_to_nii
    img = itk.imread(path_to_image)
  File "[...]\packages\lib\site-packages\itk\support\extras.py", line 1293, in imread
    reader = template_reader_type.New(**kwargs)
  File "[...]\packages\lib\site-packages\itk\support\template_class.py", line 661, in New
    return self._NewImageReader(
  File "[...]\packages\lib\site-packages\itk\support\template_class.py", line 160, in _NewImageReader
    raise RuntimeError(
RuntimeError: Could not create IO object for reading file "[ Path with Umlaut ]"
```

**Steps to Reproduce:**
Load a image file with the:
- ITK-Framework: itk.imread(**"[ Path with Umlaut ]"**)
- SimpleITK-Framework: sitk.ReadImage(**"[ Path with Umlaut ]"**)

**Expected Behavior:**
The ITK framework should be able to read images regardless of whether the file path contains Umlauts or not. A successful image loading operation is expected.
Alternatively, providing a more descriptive error message indicating that the path cannot be read due to special characters, such as Umlauts, would be helpful for users to understand and address the issue.

**Actual Behavior:**
The execution fails with a RuntimeError indicating that an IO object cannot be created for reading the image file.

**Additional Information:**
    Operating System: Windows 11
    ITK Version: 5.3.0
    Simple-ITK Version: 2.3.1
    Python Version: 3.9.0

**Note:**
The issue does not related to an incorrect path, as the error does not indicate a path error. Rather, it suggests that the ITK framework encounters difficulties creating an IO object when the file path contains Umlauts.

Thank you for your attention to this matter.

dzenanz:
This looks like a duplicate of #2693. Possibly also related: #540.

dzenanz:
> a simple check could be implemented

Contributions are always welcome.

Pfleiderer-Adrian:
> Contributions are always welcome.

I will implement that tomorrow. :)

[...]

PR #4390, opened 2024-01-10
Add checks for unsupported characters within image path and output path

Pfleiderer-Adrian:
## Description
I encountered an issue while attempting to load images using the ITK / SimpleITK framework in my Python script. The traceback indicates that the problem arises when attempting to create an IO object for reading an image file that contains Umlauts (ä, ü, ö) in its path. See: #4388

Providing a more descriptive error message indicating that the path cannot be read due to special characters, such as Umlauts, would be helpful for users to understand and address the issue.

## Changes
I have added a simple filepath check for illegal characters in the _NewImageReader and imwrite routines. Additionally, for the imwrite routine, a check was added to verify if the output directory exists.

## PR Checklist

### DIFF
--- a/Wrapping/Generators/Python/itk/support/extras.py
+++ b/Wrapping/Generators/Python/itk/support/extras.py
@@ -1261,6 +1261,27 @@ def imwrite(
     The writer is instantiated with the image type of the image in
     parameter (or, again, with the output image of the filter in parameter).
     """
+
+    # Check if output path exists
+    msg = ""
+    if not os.path.isdir(os.path.dirname(filename)):
+        msg += "\nThe output dir doesn't exist. \n" + f"Filename = {filename}"
+        raise RuntimeError(
+            f"Could not create IO object for writing file {filename}" + msg
+        )
+    # Check if path contains unsupported special characters
+    else:
+        try:
+            filename.encode("ascii")
+        except UnicodeEncodeError:
+            msg += (
+                "\nThe output path contains unsupported special characters for ascii encoding.\n"
+                + f"Filename = {filename}"
+            )
+            raise RuntimeError(
+                f"Could not create IO object for writing file {filename}" + msg
+            )
+
     import itk
 
     img = itk.output(image_or_filter)
--- a/Wrapping/Generators/Python/itk/support/template_class.py
+++ b/Wrapping/Generators/Python/itk/support/template_class.py
@@ -158,6 +158,15 @@ class itkTemplate(Mapping):
             msg = ""
             if not os.path.isfile(inputFileName):
                 msg += "\nThe file doesn't exist. \n" + f"Filename = {inputFileName}"
+            # Check if image path contains not supported special characters.
+            else:
+                try:
+                    inputFileName.encode("ascii")
+                except UnicodeEncodeError:
+                    msg += (
+                        "\nThe image path contains not supported special characters. \n"
+                        + f"Filename = {inputFileName}"
+                    )
             raise RuntimeError(
                 f"Could not create IO object for reading file {inputFileName}" + msg
             )


Wrapping/Generators/Python/itk/support/extras.py
```diff
@@ -1165,6 +1165,24 @@ def imwrite(
...
+        msg += "\nThe output dir doesn't exist. \n" + f"Filename = {filename}"
+        raise RuntimeError(
+            f"Could not create IO object for writing file {filename}" + msg
+        )
+    # Check if path contains unsupported special characters
+    else:
+        try:
+            filename.encode('ascii')
+        except UnicodeEncodeError:
+            msg += "\nThe output path contains not supported special characters. \n" + f"Filename = {filename}"
```
dzenanz, line 1181: not supported -> unsupported

Perhaps give user a hint about setting UTF-8 as the default code page?

Wrapping/Generators/Python/itk/support/template_class.py
```diff
@@ -157,6 +158,12 @@ def firstIfList(arg):
             msg = ""
             if not os.path.isfile(inputFileName):
                 msg += "\nThe file doesn't exist. \n" + f"Filename = {inputFileName}"
+            # Check if image path contains not supported special characters.
+            else:
+                try:
+                    inputFileName.encode('ascii')
+                except UnicodeEncodeError:
+                    msg += "\nThe image path contains not supported special characters. \n" + f"Filename = {inputFileName}"
```
dzenanz, line 165: Preferably give a hint about UTF8 code page.

dzenanz (commented):
Ah, this check in makes sense via Python interface. Mostly looks good.

blowekamp:
There may be something in the SWIG layer that can be down with the conversion. There is some information here:
https://www.swig.org/Doc3.0/Python.html#Python_nn77

That needs to be gone through.

jhlegarreta:
@Pfleiderer-Adrian thanks for having a look at this. A few comments:
- Commit 7304267 does not addess a bug, but a `STYLE` change.
- Please, add a meaningful commit message body in line with [ITK commit message practices](https://docs.itk.org/en/latest/contributing/index.html#commit-messages).

dzenanz:
The second commit should be squashed into the first one.

Wrapping/Generators/Python/itk/support/extras.py
```diff
@@ -1165,6 +1165,24 @@ def imwrite(
...
     parameter (or, again, with the output image of the filter in parameter).
     """
+
+    # Check if output path exists
+    msg = ""
+    if not os.path.isdir(os.path.dirname(filename)):
+        msg += "\nThe output dir doesn't exist. \n" + f"Filename = {filename}"
+        raise RuntimeError(
+            f"Could not create IO object for writing file {filename}" + msg
+        )
```
blowekamp, line 1175: This would be a useful check to occur in the C++ code.

Wrapping/Generators/Python/itk/support/extras.py
blowekamp, line 1181: It may be useful to place this check into a function.

blowekamp (commented):
A couple suggestions.

Wrapping/Generators/Python/itk/support/extras.py
dzenanz: I agree, but I assume that performing an equivalent check in C++ would be a lot harder. If the author has the will to explore this, it would be good.

Pfleiderer-Adrian:
Thank you for your feedback! I'll post an update this weekend. :)

N-Dekker:
Thank you for diving into this issue, @Pfleiderer-Adrian  Do I understand correctly that this issue (#4388) is only a problem for _Windows_ users, not for Linux users? (Specifically, only those Windows users that do not have UTF-8 support enabled _and_ selected a locale that does not support the filename?) I'm asking, because it feels too strict to _only_ accept ASCII characters from now, for all users.

So is it really the intention to reject any non-ASCII character in file names, for any user? Aren't we then throwing the baby out with the bath water?

Would it be an idea to _only_ do those checks _after_ trying to use `imageIO`, and _only_ when `imageIO` has really failed to do its job?

Wrapping/Generators/Python/itk/support/extras.py
hjmjohnson: Dont want to hold this up for a perfect solution.

hjmjohnson (changes requested):
This comment should be addressed before this is finalized.

https://github.com/InsightSoftwareConsortium/ITK/pull/4390#issuecomment-1888819346

Wrapping/Generators/Python/itk/support/extras.py
```diff
@@ -1261,6 +1261,27 @@ def imwrite(
     The writer is instantiated with the image type of the image in
     parameter (or, again, with the output image of the filter in parameter).
     """
+
+    # Check if output path exists
+    msg = ""
+    if not os.path.isdir(os.path.dirname(filename)):
```
dzenanz, line 1267: This will most likely interfere with Slicer's [MRML ID ImageIO](https://github.com/Slicer/Slicer/blob/126d5d26e3f0a21642bb795e2b797f7e2830c447/Libs/MRML/IDImageIO/itkMRMLIDImageIO.h#L34-L49). Scratch that - Slicer uses C++ version of ITK for "writing" temporary files.

Wrapping/Generators/Python/itk/support/template_class.py
```diff
@@ -158,6 +158,15 @@ def firstIfList(arg):
             msg = ""
             if not os.path.isfile(inputFileName):
                 msg += "\nThe file doesn't exist. \n" + f"Filename = {inputFileName}"
+            # Check if image path contains not supported special characters.
```
dzenanz, line 161: not supported -> unsupported

dzenanz (commented):
This makes the Python wrapping behave differently from C++.

hjmjohnson:
@Pfleiderer-Adrian thank you for surfacing this issue and starting the conversation — your PR put a real spotlight on the Windows non-ASCII-filename failure mode, and the reviewer discussion that followed (esp. @N-Dekker's pointer to [SuperElastix/elastix#1401](https://github.com/SuperElastix/elastix/pull/1401) by @codeling, and the @blowekamp / @dzenanz preference for fixing the C++ side rather than tightening the Python contract) shaped the takeover direction.

I've opened **#6231** as a successor that follows the elastix recipe: an 8-line UTF-8 active-code-page manifest auto-attached to every ITK test executable on Windows. With Windows 10 1903+, this lets the existing `const char *` filename APIs ingest UTF-8 byte sequences without code changes — no need to reject anything in the Python wrapper. You're credited as `Co-Authored-By:` on the PR commit.

A library-side follow-up (Tier B) will wire ITK's existing-but-unused `itkInternationalizationIOHelpers` (`i18n_open` / `i18n_fopen`) through `ImageIOBase` / PNG / JPEG so SimpleITK and Python users (where the manifest doesn't reach `python.exe`) also benefit. That'll be a separate PR, also co-authored with you.

Suggest closing this PR after #6231 merges. Thanks again for the contribution.

Pfleiderer-Adrian:
@hjmjohnson Thanks a lot for the detailed follow-up and for carrying this forward — really appreciate the co-author credit! The approach in #6231 sounds great to me. Happy to successfull close this PR.

PR #4398, opened 2024-01-14
COMP: Fix name of class assertion failure in FEM module

jhlegarreta:
Fix name of class assertion failure for
`itk::fem::FEMExceptionLinearSystemBounds`: use the correct class name.

Fixes:
```
M:\Dashboard\ITK\Modules\Numerics\FEM\include\itkFEMLinearSystemWrapper.h(586,3):
error C2607: static assertion failed
[M:\Dashboard\ITK-build\Modules\Numerics\FEM\src\ITKFEM.vcxproj]
```

Raised for example in:
https://open.cdash.org/viewBuildError.php?buildid=9299844

## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [X] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)

PR #4401, opened 2024-01-15
Add error message to static_assert in `GetNameOfClass()` macro, add `itkInternalGetNameOfClassImplementationMacro`

N-Dekker:
The added error message could be helpful when doing fixes like
- https://github.com/InsightSoftwareConsortium/ITK/pull/4398 (by @jhlegarreta)

### DIFF
--- a/Modules/Core/Common/include/itkMacro.h
+++ b/Modules/Core/Common/include/itkMacro.h
@@ -435,11 +435,13 @@ namespace itk
 
 
 // Internal macro (not part of the public ITK API), used to implement `GetNameOfClass()` member functions.
-#define itkInternalGetNameOfClassImplementationMacro(thisClass)                                              \
-  {                                                                                                          \
-    static_assert(std::is_same_v<thisClass, std::remove_const_t<std::remove_reference_t<decltype(*this)>>>); \
-    return #thisClass;                                                                                       \
-  }                                                                                                          \
+#define itkInternalGetNameOfClassImplementationMacro(thisClass)                                             \
+  {                                                                                                         \
+    static_assert(std::is_same_v<thisClass, std::remove_const_t<std::remove_reference_t<decltype(*this)>>>, \
+                  "The macro argument `" #thisClass                                                         \
+                  "` appears incorrect! It should correspond with the name of this class!");                \
+    return #thisClass;                                                                                      \
+  }                                                                                                         \
   ITK_MACROEND_NOOP_STATEMENT
 
 
--- a/Modules/Core/Common/include/itkMacro.h
+++ b/Modules/Core/Common/include/itkMacro.h
@@ -434,24 +434,23 @@ namespace itk
 #endif
 
 
-/** Macro's used to add `GetNameOfClass()` member functions to polymorphic ITK classes: `itkVirtualGetNameOfClassMacro`
- * adds a virtual `GetNameOfClass()` member function to the class definition, and `itkOverrideGetNameOfClassMacro` adds
- * a `GetNameOfClass()` override. */
-#define itkVirtualGetNameOfClassMacro(thisClass)                                                             \
-  virtual const char * GetNameOfClass() const                                                                \
+// Internal macro (not part of the public ITK API), used to implement `GetNameOfClass()` member functions.
+#define itkInternalGetNameOfClassImplementationMacro(thisClass)                                              \
   {                                                                                                          \
     static_assert(std::is_same_v<thisClass, std::remove_const_t<std::remove_reference_t<decltype(*this)>>>); \
     return #thisClass;                                                                                       \
   }                                                                                                          \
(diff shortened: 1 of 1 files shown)


Modules/Core/Common/include/itkMacro.h
```diff
@@ -434,24 +434,23 @@ namespace itk
...
 
-#define itkOverrideGetNameOfClassMacro(thisClass)                                                            \
-  const char * GetNameOfClass() const override                                                               \
-  {                                                                                                          \
-    static_assert(std::is_same_v<thisClass, std::remove_const_t<std::remove_reference_t<decltype(*this)>>>); \
-    return #thisClass;                                                                                       \
-  }                                                                                                          \
-  ITK_MACROEND_NOOP_STATEMENT
+
+/** Macro's used to add `GetNameOfClass()` member functions to polymorphic ITK classes: `itkVirtualGetNameOfClassMacro`
```
dzenanz, line 448: How would this affect the reported error messages?

Modules/Core/Common/include/itkMacro.h
```diff
@@ -434,24 +434,25 @@ namespace itk
...
 
-#define itkOverrideGetNameOfClassMacro(thisClass)                                                            \
-  const char * GetNameOfClass() const override                                                               \
-  {                                                                                                          \
-    static_assert(std::is_same_v<thisClass, std::remove_const_t<std::remove_reference_t<decltype(*this)>>>); \
-    return #thisClass;                                                                                       \
-  }                                                                                                          \
-  ITK_MACROEND_NOOP_STATEMENT
+#define itkOverrideGetNameOfClassMacro(thisClass) \
+  const char * GetNameOfClass() const override itkInternalGetNameOfClassImplementationMacro(thisClass)
```
SimonRit, line 455: Good idea! Should you add
```c++
#undef itkInternalGetNameOfClassImplementationMacro
```
after this line so that `itkInternalGetNameOfClassImplementationMacro` is indeed not in the public ITK API?

Modules/Core/Common/include/itkMacro.h
```diff
@@ -435,11 +435,13 @@ namespace itk
...
 // Internal macro (not part of the public ITK API), used to implement `GetNameOfClass()` member functions.
-#define itkInternalGetNameOfClassImplementationMacro(thisClass)                                              \
-  {                                                                                                          \
-    static_assert(std::is_same_v<thisClass, std::remove_const_t<std::remove_reference_t<decltype(*this)>>>); \
-    return #thisClass;                                                                                       \
-  }                                                                                                          \
+#define itkInternalGetNameOfClassImplementationMacro(thisClass)                                             \
+  {                                                                                                         \
+    static_assert(std::is_same_v<thisClass, std::remove_const_t<std::remove_reference_t<decltype(*this)>>>, \
+                  "The macro argument `" #thisClass                                                         \
```
dzenanz, line 441: Useful as is. It would be more useful if we added after "name of this class" its name (`std::remove_const_t<std::remove_reference_t<decltype(*this)>>`). Can this be accomplished?

Modules/Core/Common/include/itkMacro.h
N-Dekker: Thanks for the suggestions, @SimonRit  I tried the `#undef` at the end of itkMacro.h but MSVC then gave me lots of compile errors! Apparently, the internal macro still needs to be available in the header files that actually use the GetNameOfClass macro's. So it should not be "undefined" too early. OK?

SimonRit: Ok! I did not anticipate this but it makes sense. Sorry for the noise...

Modules/Core/Common/include/itkMacro.h
N-Dekker: Thanks for your comment @dzenanz. As far as I can see, with C++17, the (optional) message of a static_assert must be a simple literal string, so it cannot have something like `type(*this).name()`. https://en.cppreference.com/w/cpp/language/static_assert  But maybe with C++26... 😸  

Related C++ proposal: [user-generated `static_assert` messages](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2741r3.pdf) by Corentin Jabot

Modules/Core/Common/include/itkMacro.h
N-Dekker: Looking at the mistake already fixed by Jon (@jhlegarreta) PR #4398 commit 2999b59ebbf89ec1c38d99700f7b63922bd20a17, this PR would have added the following error message text:

    "The macro argument `FEMExceptionLinearSystem` appears incorrect! It should correspond with the name of this class!"

dzenanz (approved):

SimonRit (approved):

jhlegarreta (approved):

PR #4409, opened 2024-01-18
 WIP: Check that `Math::Round` template argument is an integer type

N-Dekker:
The `TReturn` template argument of `Math::Round` must be an integer type, according to https://github.com/InsightSoftwareConsortium/ITK/blob/66ab5b14c476d057ed46a4daba05b710681160e6/Modules/Core/Common/include/itkMath.h#L127

And indeed, `Math::Round` is internally optimized to produce an integer value. Otherwise, the user may consider calling `std::round` instead.

### DIFF
--- a/Modules/Core/Common/include/itkMath.h
+++ b/Modules/Core/Common/include/itkMath.h
@@ -33,6 +33,7 @@
 #include "itkMathDetail.h"
 #include "itkConceptChecking.h"
 #include <vnl/vnl_math.h>
+#include <type_traits>
 
 /* Only maintain backwards compatibility with old versions
  * of VXL back to the point where vnl_math:: was introduced
@@ -177,6 +178,10 @@ template <typename TReturn, typename TInput>
 inline TReturn
 Round(TInput x)
 {
+  static_assert(
+    sizeof(TReturn) > sizeof(int64_t) || std::is_integral_v<TReturn>,
+    "Math::Round template argument `TReturn` must be an integer type. If a floating point result is preferred, "
+    "consider using std::round (from <cmath>), or cast the result of a Math::Round call to a floating point type!");
   return RoundHalfIntegerUp<TReturn, TInput>(x);
 }
 
--- a/Modules/Filtering/ImageIntensity/include/itkRoundImageFilter.h
+++ b/Modules/Filtering/ImageIntensity/include/itkRoundImageFilter.h
@@ -20,6 +20,7 @@
 
 #include "itkUnaryGeneratorImageFilter.h"
 #include "itkMath.h"
+#include <type_traits> // For conditional_t and is_integral_v.
 
 namespace itk
 {
@@ -45,7 +46,18 @@ public:
   inline TOutput
   operator()(const TInput & A) const
   {
-    return itk::Math::Round<TOutput, TInput>(A);
+    if constexpr (sizeof(TOutput) <= sizeof(int64_t) && !std::is_integral_v<TOutput>)
+    {
+      using IntegerType = std::conditional_t<sizeof(TOutput) <= sizeof(int32_t), int32_t, int64_t>;
+
+      // The TReturn argument of Math::Round must be an integer type.
+      return static_cast<TOutput>(Math::Round<IntegerType>(A));
+    }
+    else
+    {
+      // Rare (exceptional) case: sizeof(TOutput) > sizeof(int64_t).
+      return Math::Round<TOutput>(A);
+    }
   }
 };
 } // namespace Functor
--- a/Modules/Filtering/Thresholding/include/itkHuangThresholdCalculator.hxx
+++ b/Modules/Filtering/Thresholding/include/itkHuangThresholdCalculator.hxx
@@ -21,6 +21,7 @@
 
 #include "itkMath.h"
 #include "itkProgressReporter.h"
+#include <cmath> // For round.
 
 namespace itk
 {
@@ -101,7 +102,7 @@ HuangThresholdCalculator<THistogram, TOutput>::GenerateData()
   for (InstanceIdentifier threshold = m_FirstBin; threshold < m_LastBin; ++threshold)
   {
     double entropy = 0.;
-    auto   mu = Math::Round<MeasurementType>(W[threshold] / S[threshold]);
+    auto   mu = static_cast<MeasurementType>(std::round(W[threshold] / S[threshold]));
 
     typename HistogramType::MeasurementVectorType v(1);
     v[0] = mu;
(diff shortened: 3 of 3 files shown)


dzenanz (approved):

blowekamp:
If I recall correctly one reason for our custom rounding method was that the behavior of round x.5 was different between implementations. There are 4 possibilities for rounding x.5 down, up, towards 0, away from 0.

Do you know if the std::round methods define this behavior now?

N-Dekker:
> Do you know if the std::round methods define this behavior now?

Thanks for asking @blowekamp According to https://en.cppreference.com/w/cpp/numeric/math/round, std::round is: _rounding halfway cases away from zero_

On the other side, Math::Round calls `RoundHalfIntegerUp`, which claims to round negative numbers towards zero:

https://github.com/InsightSoftwareConsortium/ITK/blob/66ab5b14c476d057ed46a4daba05b710681160e6/Modules/Core/Common/include/itkMath.h#L153-L155

I did not know about this, honestly. I find the behavior of std::round more intuitive. But then again, I'm not a banker  😺

blowekamp:
> I did not know about this, honestly. I find the behavior of std::round more intuitive. But then again, I'm not a banker 😺

Ahh... Now I recall one scenario where this really matter in ITK. For Images which have unit spacing, and an origin of say (0.5,) the center of the voxel can lie on the 0.5 boundary. With ITK's support for negative spacial locations it is important to always round in the positive direction to get consistent behavior across the 0 boundary.

It would be good to not change the behavior of the RoundImageFilter.

issakomi:
```cpp
#include <itkMath.h>
#include <iostream>
#include <cmath>

int main()
{
	double A = -1.5;
	std::cout << itk::Math::Round<double, double>(A) << " " << std::round(A) << std::endl;
	return 0;
}
```

```
r@deb2:~/tmp/test019/build$ ./test 
-1 -2
```

N-Dekker:
@blowekamp @issakomi Thanks for your feedback. With the force-pushes that I just did, the behavior of `RoundImageFilter` (specifically its `Functor::Round`) should be restored: rounding half up towards +infinity, using `Math::Round`. Please have a look!

I still think the compile-time check should be in there, as I encountered quite a few mistaken `Round<float>` and `Round<double>` calls, both in elastix (https://github.com/SuperElastix/elastix/pull/1020) and in ITK: https://github.com/InsightSoftwareConsortium/ITK/blob/c6fedbdb9ad501d879fba33c23cb088382d1dde5/Modules/Filtering/Denoising/include/itkPatchBasedDenoisingImageFilter.hxx#L45-L46

Note that the result of `Math::Detail::RoundHalfIntegerUp_base<TReturn>` appears very wrong when `TReturn` is a floating point type:

    RoundHalfIntegerUp_base<float>(0.75) returns 1.25
    RoundHalfIntegerUp_base<double>(0.75) returns 1.25
    RoundHalfIntegerUp_base<long double>(0.75) returns 1.25

Fortunately,  `Math::Round<TReturn>` only calls `RoundHalfIntegerUp_base<TReturn>` when `TReturn` is more than 64 bits.

Modules/Core/Common/include/itkMath.h
```diff
@@ -178,6 +179,10 @@ template <typename TReturn, typename TInput>
 inline TReturn
 Round(TInput x)
 {
+  static_assert(
+    sizeof(TReturn) > sizeof(int64_t) || std::is_integral_v<TReturn>,
+    "Math::Round template argument `TReturn` must be an integer type. If a floating point result is preferred, "
+    "consider using std::round (from <cmath>), or cast the result of a Math::Round call to a floating point type!");
```
issakomi, line 185: IMHO, it would be better to avoid suggestion of the _std::round_ ( _always_ rounds "to-nearest"), but itk::Math::Round actually rounds "upward". There are other functions, e.g. std::rint, std::nearbyint, they can be controlled via _fesetround_, but very often it has no effect, it depends on compiler flags related to floating point math, etc., so also bad. IMHO, just suggest to cast here should be safer.

Modules/Filtering/ImageIntensity/include/itkRoundImageFilter.h
```diff
@@ -47,7 +48,18 @@ class Round
   inline TOutput
   operator()(const TInput & A) const
   {
-    return itk::Math::Round<TOutput, TInput>(A);
```
issakomi, line 50: Isn't it safer to leave as is? IMHO, it would be nice to make Round work with floating output, but maybe not this time?

issakomi:
> ```
> RoundHalfIntegerUp_base<float>(0.75) returns 1.25
> RoundHalfIntegerUp_base<double>(0.75) returns 1.25
> RoundHalfIntegerUp_base<long double>(0.75) returns 1.25
> ```
> 
> Fortunately, Math::Round<TReturn> only calls RoundHalfIntegerUp_base<TReturn> when TReturn is more than 64 bits.

Yes, on Linux x86_64:

```cpp
#include <itkMath.h>
#include <iostream>

int main()
{
	std::cout << itk::Math::Round<long double>(0.75) << std::endl;
	return 0;
}
```

```
r@deb2:~/tmp/test019/build$ ./test 
1.25
```

Unfortunately, even if according to documentation TReturn must be an integer, this compiles without warnings. The size of `long double` is 16 bytes on Linux x86_64, so we are here. Bug. So _static_assert_ to prevent floating points type is a good idea.

N-Dekker:
This pull request has two commits that triggered discussion, but three other commits seem uncontroversial. Would it be an idea to just have the three "uncontroversial" commits merged, and postpone the other two? 

Specifically, the following three seem uncontroversial:

- https://github.com/InsightSoftwareConsortium/ITK/pull/4409/commits/5fb59a0b7c8e8125a3485f1e6210df6305cae1f6
- https://github.com/InsightSoftwareConsortium/ITK/pull/4409/commits/16147f5e3b95eb1faf3183155e935e3d67c49055
- https://github.com/InsightSoftwareConsortium/ITK/pull/4409/commits/8f23e8579703722f774857db904b39e74e89f806

Shall I cherry-pick those three into a new PR?

N-Dekker:
Back to draft: Let's first see if we can get agreement on the three less controversial commits:
- pull request #4430

N-Dekker:
> With the other PR merged, maybe it is time to reconsider this PR: update, close, or rebase?

Just did a rebase, but I would not mind having the remaining three commits postponed until after the next major release. Or if someone would like to make floating point support for `TResult` "official", and drop the old requirement that says that "`TReturn` must be an integer type", that could also be an option. Anyway, at the moment, it has no priority to me anymore. I did not know before that this PR would have triggered so much discussion and I don't want to be too pushy on this relatively small topic.

N-Dekker:
Closed for now, may be reopened/reconsidered after the next major ITK release, if anyone would like to  😃

PR #4413, opened 2024-01-19
STYLE: Replace old `if` by "constexpr if" in `itk::Math` implementation

N-Dekker:
Enforced compile-time evaluation of conditions, when rounding and when doing `CastWithRangeCheck`.


dzenanz (approved):

N-Dekker:
Officially it's "replace _with_", rather than "replace _by_", but you know what I mean  😸

PR #4430, opened 2024-01-25
Fix `Math::Round` calls that have a floating point type as template argument

N-Dekker:
The ~~three~~ two commits (out of five) from pull request #4409 that seem uncontroversial.

The idea is that _even_ if we do not (yet?) add a compile-time check against using a floating point type as `TReturn` template argument, we should still _avoid_ using a floating point type for `TReturn`! As documented:

https://github.com/InsightSoftwareConsortium/ITK/blob/66ab5b14c476d057ed46a4daba05b710681160e6/Modules/Core/Common/include/itkMath.h#L170-L172

Originally documented 15 years ago:

https://github.com/InsightSoftwareConsortium/ITK/blob/17269cbe9e620c308bf614a13f6acbbb4cc89e48/Code/Common/itkMath.h#L151-L153

### DIFF
--- a/Modules/Filtering/Denoising/include/itkPatchBasedDenoisingImageFilter.hxx
+++ b/Modules/Filtering/Denoising/include/itkPatchBasedDenoisingImageFilter.hxx
@@ -26,6 +26,7 @@
 #include "itkImageFileWriter.h"
 #include "itkGaussianOperator.h"
 #include "itkImageAlgorithm.h"
+#include "itkIntTypes.h"
 #include "itkVectorImageToImageAdaptor.h"
 #include "itkSpatialNeighborSubsampler.h"
 #include "itkMacro.h"
@@ -43,7 +44,7 @@ PatchBasedDenoisingImageFilter<TInputImage, TOutputImage>::PatchBasedDenoisingIm
   m_MinProbability(NumericTraits<RealValueType>::min() * 100)
   , // to avoid divide by zero
   m_SigmaUpdateDecimationFactor(
-    static_cast<unsigned int>(Math::Round<double>(1.0 / m_KernelBandwidthFractionPixelsForEstimation)))
+    static_cast<unsigned int>(Math::Round<int64_t>(1.0 / m_KernelBandwidthFractionPixelsForEstimation)))
   , m_NoiseSigma()
   , m_NoiseSigmaSquared()
   , m_SearchSpaceList(ListAdaptorType::New())
@@ -239,10 +240,10 @@ PatchBasedDenoisingImageFilter<TInputImage, TOutputImage>::Initialize()
 
   // For automatic sigma estimation, select every 'k'th pixel.
   m_SigmaUpdateDecimationFactor =
-    static_cast<unsigned int>(Math::Round<double>(1.0 / m_KernelBandwidthFractionPixelsForEstimation));
+    static_cast<unsigned int>(Math::Round<int64_t>(1.0 / m_KernelBandwidthFractionPixelsForEstimation));
   // For automatic sigma estimation, use at least 1% of pixels.
-  m_SigmaUpdateDecimationFactor = std::min(m_SigmaUpdateDecimationFactor,
-                                           static_cast<unsigned int>(Math::Round<double>(m_TotalNumberPixels / 100.0)));
+  m_SigmaUpdateDecimationFactor = std::min(
+    m_SigmaUpdateDecimationFactor, static_cast<unsigned int>(Math::Round<int64_t>(m_TotalNumberPixels / 100.0)));
   // For automatic sigma estimation, can't use more than 100% of pixels.
   m_SigmaUpdateDecimationFactor = std::max(m_SigmaUpdateDecimationFactor, 1u);
 
--- a/Modules/Filtering/ImageNoise/include/itkNoiseBaseImageFilter.hxx
+++ b/Modules/Filtering/ImageNoise/include/itkNoiseBaseImageFilter.hxx
@@ -54,7 +54,7 @@ NoiseBaseImageFilter<TInputImage, TOutputImage>::ClampCast(const double value) -
   {
     return NumericTraits<OutputImagePixelType>::NonpositiveMin();
   }
-  else if (NumericTraits<OutputImagePixelType>::is_integer)
+  else if constexpr (NumericTraits<OutputImagePixelType>::is_integer)
   {
     return Math::Round<OutputImagePixelType>(value);
   }


Modules/Filtering/Thresholding/include/itkHuangThresholdCalculator.hxx
```diff
@@ -102,7 +103,7 @@ HuangThresholdCalculator<THistogram, TOutput>::GenerateData()
   for (InstanceIdentifier threshold = m_FirstBin; threshold < m_LastBin; ++threshold)
   {
     double entropy = 0.;
-    auto   mu = Math::Round<MeasurementType>(W[threshold] / S[threshold]);
+    auto   mu = static_cast<MeasurementType>(std::round(W[threshold] / S[threshold]));
```
blowekamp, line 106: What type is the results of `W[theshold]`? If it can be signed then this may change results in some cases.

dzenanz (approved):

Modules/Filtering/Thresholding/include/itkHuangThresholdCalculator.hxx
issakomi: h

Modules/Filtering/Thresholding/include/itkHuangThresholdCalculator.hxx
```diff
@@ -22,6 +22,7 @@
 #include "itkMath.h"
 #include "itkProgressReporter.h"
 #include "itkMath.h"
+#include <cmath> // For round.
```
issakomi, line 25: "itkMath.h" is included twice, BTW

Modules/Filtering/Denoising/include/itkPatchBasedDenoisingImageFilter.hxx
```diff
@@ -26,6 +26,7 @@
 #include "itkImageFileWriter.h"
```
issakomi, line 26: Off-topic, but "itkImageFileWriter.h" is strange here.

Modules/Filtering/Thresholding/include/itkHuangThresholdCalculator.hxx
N-Dekker: I don't think people would _expect_ negative measurement values of the form `-n.5` (for a natural number `n`) to be rounded towards zero. Do you?

Modules/Filtering/Thresholding/include/itkHuangThresholdCalculator.hxx
N-Dekker: @issakomi Interesting... It appears quite common to `#include "itkMath.h"` twice! Just did a search for regular expression `#include "itkMath\.h"\r\n.*#include "itkMath\.h"`... 18 hits!  😸  Feel free to make that a new PR, if you like!

Modules/Filtering/Thresholding/include/itkHuangThresholdCalculator.hxx
issakomi: std::nearbyint (and std::rint) has 4 rounding modes:

```cpp
#include <iostream>
#include <cfenv>
#include <cmath>

namespace
{

void printrounding()
{
	std::cout << "\nRounding using ";
	switch (std::fegetround())
	{
	case FE_DOWNWARD:
		std::cout << "downward";
		break;
	case FE_TONEAREST:
		std::cout << "to-nearest";
		break;
	case FE_TOWARDZERO:
		std::cout << "toward-zero";
		break;
	case FE_UPWARD:
		std::cout << "upward";
		break;
	default:
		std::cout << "unknown";
		break;
	}
}

}

int main()
{
	// Default (FE_TONEAREST)
	printrounding();
	std::cout
		<< "\n -1.5 --> " << std::nearbyint(-1.5)
		<< "\n  1.5 --> " << std::nearbyint(1.5);

	// FE_TOWARDZERO
	std::fesetround(FE_TOWARDZERO);
	printrounding();
	std::cout
		<< "\n -1.5 --> " << std::nearbyint(-1.5)
		<< "\n  1.5 --> " << std::nearbyint(1.5);

	// FE_UPWARD
	std::fesetround(FE_UPWARD);
	printrounding();
	std::cout
		 << "\n -1.5 --> " << std::nearbyint(-1.5)
		 << "\n  1.5 --> " << std::nearbyint(1.5) << std::endl;

	return 0;
}
```

```
r@deb2:~/tmp$ g++ rounding.cpp 
r@deb2:~/tmp$ ./a.out 

Rounding using to-nearest
 -1.5 --> -2
  1.5 --> 2
Rounding using toward-zero
 -1.5 --> -1
  1.5 --> 1
Rounding using upward
 -1.5 --> -1
  1.5 --> 2
```

So some people expect this in some situations. Unfortunately setting a mode works very bad, from ITK build doesn't work for me in test. In particular case, IMHO, simple std::round (always to-nearest) may be OK. But it might be out of scope of the PR.

issakomi: > Unfortunately setting a mode works very bad,

Reproduced, cpp is from the above post, s. **-O3** flag. Probably this or similar was a reason to implement own _Round_.

```
r@deb2:~/tmp$ g++ rounding.cpp 
r@deb2:~/tmp$ ./a.out 

Rounding using to-nearest
 -1.5 --> -2
  1.5 --> 2
Rounding using toward-zero
 -1.5 --> -1
  1.5 --> 1
Rounding using upward
 -1.5 --> -1
  1.5 --> 2
```

```
r@deb2:~/tmp$ g++ -O3 rounding.cpp 
r@deb2:~/tmp$ ./a.out 

Rounding using to-nearest
 -1.5 --> -2
  1.5 --> 2
Rounding using toward-zero
 -1.5 --> -2
  1.5 --> 2
Rounding using upward
 -1.5 --> -2
  1.5 --> 2
```

issakomi (approved):
Please be aware that the result of the _HuangThresholdCalculator_ may be slightly different, but apparently this is acceptable (I hope).

Modules/Filtering/Thresholding/include/itkHuangThresholdCalculator.hxx
issakomi: Interesting. And the winner is [itkBSplineScatteredDataPointSetToImageFilter.hxx](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/ImageGrid/include/itkBSplineScatteredDataPointSetToImageFilter.hxx) with 3 `#include "itkMath.h"
`

issakomi: _itkMath.h_ already has `#include <cmath>`, BTW.

issakomi: S. #4436

N-Dekker: > _itkMath.h_ already has `#include <cmath>`, BTW.

OK, very well observed, but in general I would rather not depend on _indirectly_ included headers. Following the "Include What You Use" philosophy. (An indirectly_ included header might just be an internal implementation detail of the directly included header.)

N-Dekker:
@issakomi Thanks for your approval! I still wait for Bradley (@blowekamp) to approve, though. Otherwise maybe just do the other two commits, as these two seem even _less_ controversial to me 🤷

Modules/Filtering/Thresholding/include/itkHuangThresholdCalculator.hxx
blowekamp: > I don't think people would _expect_ negative measurement values of the form `-n.5` (for a natural number `n`) to be rounded towards zero. Do you?

Similarly I don't think most people would expect a bias when rounding around zero and with other numeric computations or bining around zero. Here is a figure:
https://upload.cppreference.com/mwiki/images/7/7c/math-round_away_zero.svg

blowekamp:
This is a subtle change in behavior I pointed out. It is hard to say if computation would be changed as a result of it. It would be good to maintain the same behavior. But my time is limited, and defer to the others to weigh the pro and cons of such a near "style" change.

N-Dekker:
@blowekamp Thanks for your reply. Is there any chance that you can still approve the other two commits of this PR?

N-Dekker:
@blowekamp Thanks! Please consider approving this PR, now that it has just the two most _uncontroversial_ commits! 😃

dzenanz (approved):

blowekamp (approved):

PR #4431, opened 2024-01-25
Use range-based `for` loop to iterate over children, add const to local variables in SpatialObject

N-Dekker:
Just a few coding style improvements.


dzenanz (approved):

aylward (approved):
Nice!

PR #4434, opened 2024-01-25
WIP: Add `SpatialObject::WorldSpaceContext` class, a faster alternative to SpatialObject::IsInsideInWorldSpace

N-Dekker:
`WorldSpaceContext::IsInsideSpatialObject` serves as a faster alternative to the
existing `SpatialObject::IsInsideInWorldSpace`. `WorldSpaceContext` has already
cached the ObjectToWorldTransformInverse transform during its construction.
`IsInsideSpatialObject` calls the faster
`SpatialObject::IsInsideInObjectSpace(const PointType &)` overload, instead of
the overload with three parameters.

Using VS2022 Release, `WorldSpaceContext::IsInsideSpatialObject` appeared more
than twice as fast as the equivalent `SpatialObject::IsInsideInWorldSpace` call.

----

Update: If the `IsInsideInWorldSpace(const PointType &)` overload proposed by pull request #4440 can make it onto the master branch, I will probably drop this `WorldSpaceContext` proposal (#4434), because then I think `WorldSpaceContext::IsInsideSpatialObject(point)` won't be much faster than the equivalent `SpatialObject::IsInsideInWorldSpace` call anymore.

### DIFF
--- a/Modules/Registration/Metricsv4/include/itkJointHistogramMutualInformationImageToImageMetricv4.hxx
+++ b/Modules/Registration/Metricsv4/include/itkJointHistogramMutualInformationImageToImageMetricv4.hxx
@@ -90,54 +90,62 @@ JointHistogramMutualInformationImageToImageMetricv4<TFixedImage,
    *  max and min for the fixed image. */
   itk::ImageRegionConstIteratorWithIndex<TFixedImage> fi(this->m_FixedImage, this->m_FixedImage->GetRequestedRegion());
 
-  /** \todo multi-thread me */
-  while (!fi.IsAtEnd())
   {
-    typename TFixedImage::PointType fixedSpacePhysicalPoint;
-    this->m_FixedImage->TransformIndexToPhysicalPoint(fi.GetIndex(), fixedSpacePhysicalPoint);
-    /* A null mask implies entire space is to be used.*/
-    if (this->m_FixedImageMask.IsNull() || this->m_FixedImageMask->IsInsideInWorldSpace(fixedSpacePhysicalPoint))
+    const typename Superclass::FixedImageMaskType::WorldSpaceContext worldSpaceContext(Superclass::m_FixedImageMask);
+
+    /** \todo multi-thread me */
+    while (!fi.IsAtEnd())
     {
-      const typename TFixedImage::PixelType currentValue = fi.Get();
-      // update the Fixed Image true min accordingly
-      if (currentValue < this->m_FixedImageTrueMin)
+      typename TFixedImage::PointType fixedSpacePhysicalPoint;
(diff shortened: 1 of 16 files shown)


Modules/Registration/Metricsv4/include/itkJointHistogramMutualInformationImageToImageMetricv4.hxx
```diff
@@ -90,54 +90,62 @@ JointHistogramMutualInformationImageToImageMetricv4<TFixedImage,
...
   {
-    typename TFixedImage::PointType fixedSpacePhysicalPoint;
-    this->m_FixedImage->TransformIndexToPhysicalPoint(fi.GetIndex(), fixedSpacePhysicalPoint);
-    /* A null mask implies entire space is to be used.*/
-    if (this->m_FixedImageMask.IsNull() || this->m_FixedImageMask->IsInsideInWorldSpace(fixedSpacePhysicalPoint))
+    const typename Superclass::FixedImageMaskType::WorldSpaceContext worldSpaceContext(Superclass::m_FixedImageMask);
+
+    /** \todo multi-thread me */
+    while (!fi.IsAtEnd())
     {
```
dzenanz, line 98: Should [MinimumMaximumImageCalculator](https://itk.org/Doxygen/html/classitk_1_1MinimumMaximumImageCalculator.html) be used here? And parallelize it? Though parallelization might not matter much in this cases, as this is almost certainly limited by memory throughput, not computation.

dzenanz (approved):

Modules/Registration/Metricsv4/include/itkJointHistogramMutualInformationImageToImageMetricv4.hxx
blowekamp: I recall optimizing the itk::MinimumMaximumImageFilter and was able to get surprising gains in the filter.

N-Dekker: As far as I can see, MinimumMaximumImageCalculator does not support masks: https://github.com/InsightSoftwareConsortium/ITK/blob/8010fafde3eb40a69db6ee9f49e547d0565da92b/Modules/Core/Common/include/itkMinimumMaximumImageCalculator.h

dzenanz: Ah, I didn't think about that 😄

N-Dekker:
This pull request (#4434, proposing `WorldSpaceContext`) is superseded by the following three pull requests, which should _together_ make any existing `spatialObject->IsInsideInWorldSpace(point)` just as fast as what _might have been_ achieved by `WorldSpaceContext`:
- #4435
- #4440
- #4441

PR #4435, opened 2024-01-26
PERF: Let SpatialObject directly access m_ObjectToWorldTransformInverse

N-Dekker:
Replaced `this->GetObjectToWorldTransformInverse()` calls with direct access to m_ObjectToWorldTransformInverse. A performance improvement greater than 6% was observed, from more than 0.74 sec. down to less than 0.69 sec, when calling `IsInsideInWorldSpace` 2^25 times (as tested on VS2019 Release).

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkSpatialObject.h
+++ b/Modules/Core/SpatialObjects/include/itkSpatialObject.h
@@ -317,18 +317,27 @@ public:
   itkSetMacro(DefaultOutsideValue, double);
   itkGetConstMacro(DefaultOutsideValue, double);
 
-  /** World space equivalent to ValueAtInObjectSpace */
+  /** World space equivalent to ValueAtInObjectSpace
+   * \note This member function assumes that the internal `ObjectToWorldTransformInverse` transform is up-to-date. This
+   * transform may be updated explicitly by calling `GetObjectToWorldTransformInverse()`, `Update()`, or
+   * `SetObjectToWorldTransform(transform)` */
   virtual bool
   ValueAtInWorldSpace(const PointType &   point,
                       double &            value,
                       unsigned int        depth = 0,
                       const std::string & name = "") const;
 
-  /** World space equivalent to IsInsideInObjectSpace */
+  /** World space equivalent to IsInsideInObjectSpace
+   * \note This member function assumes that the internal `ObjectToWorldTransformInverse` transform is up-to-date. This
+   * transform may be updated explicitly by calling `GetObjectToWorldTransformInverse()`, `Update()`, or
+   * `SetObjectToWorldTransform(transform)` */
   virtual bool
   IsInsideInWorldSpace(const PointType & point, unsigned int depth = 0, const std::string & name = "") const;
 
-  /** World space equivalent to IsEvaluableAtInObjectSpace */
+  /** World space equivalent to IsEvaluableAtInObjectSpace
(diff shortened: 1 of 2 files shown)


aylward:
Not ideal.

1) You are not guaranteed to have a valid inverse value unless you check the mtimes.
2) ITK's design pattern is not to directly access parent member vars for this reason.  So, the pattern used here isn't consistent with the rest of ITK.

How about instead still creating a function call to access it, but eliminate the test for mtimes in that function.  The function name should make it obvious that the variable is being assumed to be valid.   Perhaps

GetLastValidObjectToWorldTransformInverse()

or

GetObjectToWorldTransformInverseWithoutVerification()

or 

such.

aylward:
@dzenanz - what is the recommended pattern for this situation?

aylward:
Ah - I see this is all within the base class - not derived.   So my comment on derived class isn't valid, but it still holds that the inverse might not be valid without the mtime test.

s

N-Dekker:
Thanks for your feedback, @aylward `m_ObjectToWorldTransformInverse` is kept up-to-date with each `SetObjectToWorldTransform` call:

https://github.com/InsightSoftwareConsortium/ITK/blob/d4d7419a8ca3b41700dfe517a4185793242f923a/Modules/Core/SpatialObjects/include/itkSpatialObject.hxx#L525-L530 

As well as by `ProtectedComputeObjectToWorldTransform()` (which is called by SpatialObject::Update()):

https://github.com/InsightSoftwareConsortium/ITK/blob/d4d7419a8ca3b41700dfe517a4185793242f923a/Modules/Core/SpatialObjects/include/itkSpatialObject.hxx#L500-L509

That's why I _think_  `m_ObjectToWorldTransformInverse` is kept up-to-date well enough  🤷

N-Dekker:
My previous pull request proposes a different approach:
- https://github.com/InsightSoftwareConsortium/ITK/pull/4434

One of the ideas behind PR #4434 is to take those relatively expensive `GetObjectToWorldTransformInverse()` calls out of the pixel-by-pixel loop, and store a pointer to that inverse transform in a local instance of the proposed `WorldSpaceContext` class, _before_ going into the loop.

So this PR (#4435) kind-of competes with my other PR, #4434  😃 If `IsInsideInWorldSpace` would no longer call `GetObjectToWorldTransformInverse()`, the proposed `WorldSpaceContext` class would become less interesting. What do you think?

aylward:
This is outstanding!

I like the approach in #4434 - very nice!

Perhaps just add a comment to those worldspace functions that they assume that the inverse is up to date.    Then we'll cross our fingers and hope some third party hasn't done something that breaks.   As you said, it _should_ be up to date.

Thanks!

N-Dekker:
@aylward Thanks! If this PR passes the CI, I'll add a note to each of the four adjusted member functions (DerivativeAtInWorldSpace, IsInsideInWorldSpace, IsEvaluableAtInWorldSpace, ValueAtInWorldSpace) saying something like:

> This member function assumes that the internal `ObjectToWorldTransformInverse` transform is up-to-date. This transform may be updated explicitly by calling `GetObjectToWorldTransformInverse()`, `Update()`, or `SetObjectToWorldTransform(transform)`

OK?

aylward:
Perfect!

aylward (approved):

dzenanz (approved):

PR #4436, opened 2024-01-27
STYLE: remove duplicated itkMath.h headers

issakomi:
17 files had `#include "itkMath.h"` twice and one file 3 times.

### DIFF
--- a/Modules/Filtering/FastMarching/include/itkFastMarchingImageFilter.hxx
+++ b/Modules/Filtering/FastMarching/include/itkFastMarchingImageFilter.hxx
@@ -20,9 +20,7 @@
 
 #include "itkImageRegionIterator.h"
 #include "itkNumericTraits.h"
-#include "itkMath.h"
 #include <algorithm>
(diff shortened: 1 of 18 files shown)


Modules/Filtering/FastMarching/include/itkFastMarchingImageFilter.hxx
```diff
@@ -20,9 +20,7 @@
 
 #include "itkImageRegionIterator.h"
 #include "itkNumericTraits.h"
-#include "itkMath.h"
 #include <algorithm>
-#include "itkMath.h"
```
issakomi, line 25: Already included in itkFastMarchingImageFilter.h

dzenanz (approved):

N-Dekker:
FYI Such duplicate `#include`'s typically seem to be introduced by git merges. So it's not like people intentionally added the same `#include` twice.

N-Dekker (approved):
Thanks Mihail! Cool!

PR #4438, opened 2024-01-31
Documentation improvements on overriding `SpatialObject` member functions

N-Dekker:
Replaced SpatialObject examples of virtual functions that one may override, as `ValueAtInWorldSpace`,
`IsEvaluableAtInWorldSpace`, and `IsInsideInWorldSpace` may not need to be overridden. Actually it appears that none of the thirteen derived SpatialObject classes of ITK itself does override any of those three member functions!

Used the word "override" more consistently, and removed the term "pure virtual function".


aylward (approved):
Good catch!

PR #4441, opened 2024-02-01
PERF: Make `m_ObjectToWorldTransformInverse->TransformPoint` non-virtual

N-Dekker:
Suppressed the virtual call mechanism for
`m_ObjectToWorldTransformInverse->TransformPoint(point)` by explicit qualification with the `TransformType` scope, in the five `SpatialObject` "InWorldSpace" member functions.

A performance improvement of more than 13 % was observed, from more than 0.36 sec. (before this commit) to less than 0.31 sec. (after this commit) when calling IsInsideInWorldSpace(point) 2^25 times, using VS2022 Release.

---

Some background info:

`TransformType` is an alias for `AffineTransform<ScalarType, VDimension>`: https://github.com/InsightSoftwareConsortium/ITK/blob/2ca2351340a1a008beee2a2531c44117a9cfb19a/Modules/Core/SpatialObjects/include/itkSpatialObject.h#L99

`m_ObjectToWorldTransformInverse` is a `const` pointer, meaning that it always points to the very same `AffineTransform` instance, during its lifetime: https://github.com/InsightSoftwareConsortium/ITK/blob/2ca2351340a1a008beee2a2531c44117a9cfb19a/Modules/Core/SpatialObjects/include/itkSpatialObject.h#L724

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkSpatialObject.hxx
+++ b/Modules/Core/SpatialObjects/include/itkSpatialObject.hxx
@@ -138,7 +138,7 @@ SpatialObject<TDimension>::DerivativeAtInWorldSpace(const PointType &           
                                                     const std::string &          name,
                                                     const DerivativeOffsetType & offset)
 {
-  const PointType pnt = m_ObjectToWorldTransformInverse->TransformPoint(point);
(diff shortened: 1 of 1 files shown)


hjmjohnson (approved):

aylward:
Outstanding job!

PR #4442, opened 2024-02-01
STYLE: FUTURE_LEGACY_REMOVE virtual SpatialObject InWorldSpace functions

N-Dekker:
Made a first step towards removing the `virtual` keyword from four `SpatialObject` "InWorldSpace" member functions, in the future:

    ValueAtInWorldSpace
    IsInsideInWorldSpace
    IsEvaluableAtInWorldSpace
    DerivativeAtInWorldSpace

These member functions are not really meant to be overridden. Removing these `virtual` keyword may yield a run-time performance improvement.

Following C++ Core Guidelines, October 12, 2023, "Don’t make a function `virtual` without reason",
http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-virtual

---

Note that none of the thirteen derived SpatialObject classes of ITK does override any of those four member functions. So it appears that they really don't need to be declared `virtual`.

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkSpatialObject.h
+++ b/Modules/Core/SpatialObjects/include/itkSpatialObject.h
@@ -320,19 +320,29 @@ public:
   /** World space equivalent to ValueAtInObjectSpace
    * \note This member function assumes that the internal `ObjectToWorldTransformInverse` transform is up-to-date. This
    * transform may be updated explicitly by calling `GetObjectToWorldTransformInverse()`, `Update()`, or
-   * `SetObjectToWorldTransform(transform)` */
-  virtual bool
-  ValueAtInWorldSpace(const PointType &   point,
-                      double &            value,
-                      unsigned int        depth = 0,
-                      const std::string & name = "") const;
+   * `SetObjectToWorldTransform(transform)`
+   * \note This member function is not meant to be overridden. In the future, it may not be declared `virtual` anymore.
+   */
+#ifndef ITK_FUTURE_LEGACY_REMOVE
+  virtual
+#endif
+    bool
+    ValueAtInWorldSpace(const PointType &   point,
+                        double &            value,
+                        unsigned int        depth = 0,
+                        const std::string & name = "") const;
 
   /** World space equivalent to IsInsideInObjectSpace
    * \note This member function assumes that the internal `ObjectToWorldTransformInverse` transform is up-to-date. This
    * transform may be updated explicitly by calling `GetObjectToWorldTransformInverse()`, `Update()`, or
-   * `SetObjectToWorldTransform(transform)` */
-  virtual bool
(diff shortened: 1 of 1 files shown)


Modules/Core/SpatialObjects/include/itkSpatialObject.h
```diff
@@ -342,9 +352,14 @@ class ITK_TEMPLATE_EXPORT SpatialObject : public DataObject
   /** World space equivalent to IsEvaluableAtInObjectSpace
    * \note This member function assumes that the internal `ObjectToWorldTransformInverse` transform is up-to-date. This
    * transform may be updated explicitly by calling `GetObjectToWorldTransformInverse()`, `Update()`, or
-   * `SetObjectToWorldTransform(transform)` */
-  virtual bool
-  IsEvaluableAtInWorldSpace(const PointType & point, unsigned int depth = 0, const std::string & name = "") const;
+   * `SetObjectToWorldTransform(transform)`
+   * \note This member function is not meant to be overridden. In the future, it may not be declared `virtual` anymore.
```
issakomi, line 356: BTW, the statement (`This member function is not meant to be overridden.`) seems to conflict with the explanation above (lines 47-52):

```
 * <...> To
 * implement your own spatial object, you need to derive from the
 * following class, which requires overriding just a few virtual functions.
 * Examples of such functions are ValueAtInObjectSpace(),
 * IsInsideInObjectSpace(), and ComputeMyBoundingBox(), each of which has a
 * meaning specific to each particular object type.
```

Modules/Core/SpatialObjects/include/itkSpatialObject.h
```diff
@@ -320,19 +320,29 @@ class ITK_TEMPLATE_EXPORT SpatialObject : public DataObject
...
    * \note This member function assumes that the internal `ObjectToWorldTransformInverse` transform is up-to-date. This
    * transform may be updated explicitly by calling `GetObjectToWorldTransformInverse()`, `Update()`, or
-   * `SetObjectToWorldTransform(transform)` */
-  virtual bool
-  ValueAtInWorldSpace(const PointType &   point,
-                      double &            value,
-                      unsigned int        depth = 0,
-                      const std::string & name = "") const;
+   * `SetObjectToWorldTransform(transform)`
+   * \note This member function is not meant to be overridden. In the future, it may not be declared `virtual` anymore.
```
issakomi, line 324: BTW, the statement (`This member function is not meant to be overridden.`) seems to conflict with the explanation above (lines 47-52):

```
 * <...> To
 * implement your own spatial object, you need to derive from the
 * following class, which requires overriding just a few virtual functions.
 * Examples of such functions are ValueAtInObjectSpace(),
 * IsInsideInObjectSpace(), and ComputeMyBoundingBox(), each of which has a
 * meaning specific to each particular object type.
```

aylward (approved):

PR #4450, opened 2024-02-06
Let `ImageSpatialObject` update the image regions of its base class, let `ImageMaskSpatialObject` use these image regions

N-Dekker:
`SpatialObject` stores the image regions of the spatial object. It appears reasonable for its derived class `ImageSpatialObject` to keep them up-to-date, in accordance with its image.

A significant performance improvement of `ImageMaskSpatialObject` is then achieved by using the buffered region and the requested region as cached in its base class, instead of calling the virtual functions `GetBufferedRegion()` and `GetRequestedRegion()` from its image.

----

It's interesting to see that the common base class `itk::SpatialObject` already has those data members to cache the image regions (`m_LargestPossibleRegion`, `m_RequestedRegion`, `m_BufferedRegion`), but it looks like they weren't used anywhere, as far as I can see:

https://github.com/InsightSoftwareConsortium/ITK/blob/436423d798e63e72444dd4a9d8787ffd34c5eb60/Modules/Core/SpatialObjects/include/itkSpatialObject.h#L731-L733

For the record, those data members were added to `itk::SpatialObject` with commit 433987062ccaee80c9a7d2f6c1ff857fe96cf929 "ENH: Removed NDimensionalSpatialObject", Julien Jomier (@jjomier), 10 February 2003.

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkImageMaskSpatialObject.hxx
+++ b/Modules/Core/SpatialObjects/include/itkImageMaskSpatialObject.hxx
@@ -39,7 +39,7 @@ ImageMaskSpatialObject<TDimension, TPixel>::IsInsideInObjectSpace(const PointTyp
 
   const IndexType index = image->TransformPhysicalPointToIndex(point);
 
-  return image->GetBufferedRegion().IsInside(index) &&
+  return Superclass::GetBufferedRegion().IsInside(index) &&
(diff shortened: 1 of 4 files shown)


aylward (approved):
Nice specialization.

dzenanz (approved):

hjmjohnson (approved):

PR #4452, opened 2024-02-08
STYLE: Replace magic number SpatialObject `MaximumDepth = 9999999`

N-Dekker:
`MaximumDepth` should just be "as high as possible" (or as _deep_ as possible), in order for a function call like `IsInsideInObjectSpace(point, MaximumDepth)` to go _as deep as possible_.

----

This pull request replaces `9999999` with `std::numeric_limits<unsigned int>::max()`, which is typically `4294967295`  (assuming 32-bit int). That's obviously a lot more than `9999999`, but in practice the depth won't go _near_ either of those two numbers. So I consider this to be a style improvement, not an enhancement.

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkSpatialObject.h
+++ b/Modules/Core/SpatialObjects/include/itkSpatialObject.h
@@ -30,6 +30,7 @@
 #include "itkAffineTransform.h"
 #include "itkVectorContainer.h"
 #include "itkBoundingBox.h"
+#include <limits>
 
 namespace itk
 {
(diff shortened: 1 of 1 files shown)


aylward (approved):
LGTM!

dzenanz (approved):

issakomi:
A little off-topic, interesting, the iterator doesn’t seem to move [here](https://github.com/InsightSoftwareConsortium/ITK/blob/671d758ac3bd54574b997032d8f72d8467e0dcea/Modules/Core/SpatialObjects/include/itkSpatialObject.hxx#L466-L475). May be intention, not sure, a little bit unusual.

```cpp
void
SpatialObject<TDimension>::RemoveAllChildren(unsigned int depth)
{
  auto it = m_ChildrenList.begin();
  while (it != m_ChildrenList.end())
  {
    auto itPtr = *it;
    it = m_ChildrenList.erase(it);
    itPtr->SetParent(nullptr);
    if (depth > 0)
    {
      itPtr->RemoveAllChildren(depth - 1);
    }
  }

  this->Modified();
}
```

issakomi:
> @issakomi - from documentation: "The std::list::erase(it) function returns an iterator to the element that follows the last element erased, which can be used to adjust other iterator and pointer values that point to the list."

Thank you.

PR #4454, opened 2024-02-09
Add ValueInitializedIsZeroFilled tests, replace `MakeFilled<Self>(NumericTraits<T>::ZeroValue())` with `Self{}`

N-Dekker:
Tested that a "value-initialized" object of one of the `FixedArray` derived types, `CovariantVector` `DiffusionTensor3D`, `RGBAPixel`, `RGBPixel`, `SymmetricSecondRankTensor`, `Point`, or `Vector` is zero-filled.

Replaced `MakeFilled<Self>(NumericTraits<T>::ZeroValue())` with `Self{}` in the `ZeroValue()` implementation of `NumericTraits` specializations.

----

Used the syntax `T{}` to create "value-initialized" objects of each of those types, see also: https://en.cppreference.com/w/cpp/language/value_initialization

### DIFF
--- a/Modules/Core/Common/include/itkNumericTraitsCovariantVectorPixel.h
+++ b/Modules/Core/Common/include/itkNumericTraitsCovariantVectorPixel.h
@@ -116,7 +116,7 @@ public:
   static const Self
   ZeroValue()
   {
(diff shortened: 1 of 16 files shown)


dzenanz (approved):
LGTM

dzenanz (approved):

Issue #3163, opened 2022-02-05
Build error for wasm


engineer1109:
### Description

Build error for wasm when -s USE_PTHREADS=1

Tells:

In file included from /home/wjl/github/ITK/Modules/Core/Common/src/itkPlatformMultiThreader.cxx:40:
/home/wjl/github/ITK/Modules/Core/Common/src/itkPlatformMultiThreaderSingle.cxx:42:29: error: use of undeclared identifier 'm_GlobalMaximumNumberOfThreads'; did you mean 'm_MaximumNumberOfThreads'?

I changed  m_GlobalMaximumNumberOfThreads  to MultiThreaderBase::GetGlobalMaximumNumberOfThreads() will be ok.

### Actual behavior

```none
```

### Versions

Master , the lastest

### Environment

Linux 

emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 2.0.17
clang version 13.0.0 (/b/s/w/ir/cache/git/chromium.googlesource.com-external-github.com-llvm-llvm--project 3b677b81cec7b3c5132aee8fccc30252d87deb69)
Target: wasm32-unknown-emscripten
Thread model: posix

thewtex:
To use wasm, use [itk-wasm](http://insightsoftwareconsortium.github.io/itk-wasm). For parallelism, use the WebWorkerPool:

https://github.com/InsightSoftwareConsortium/itk-wasm/blob/12eb34db0fb1a2efdd71c1e5a60bf9808e6d2310/test/browser/core/WebWorkerPoolTest.js#L23-L56

PR #4456, opened 2024-02-09
COMP: Fix MultiThreaderSingle typos

fabian-philips:
The top of MultipleMethodExecute() in itkPlatformMultiThreaderSingle.cxx was outdated. I have copied two lines from itkPlatformMultiThreaderWindows.cxx in order to fix it. This problem was mentioned in as an issue but not yet fixed: https://github.com/InsightSoftwareConsortium/ITK/issues/3163.

## PR Checklist
- [x] Added test (or *behavior not changed*)
- [x] Updated API documentation (or *API not changed*)

### DIFF
--- a/Modules/Core/Common/src/itkPlatformMultiThreaderSingle.cxx
+++ b/Modules/Core/Common/src/itkPlatformMultiThreaderSingle.cxx
@@ -39,9 +39,9 @@ PlatformMultiThreader::MultipleMethodExecute()
   ThreadIdType thread_loop;
 
   // obey the global maximum number of threads limit
-  if (m_NumberOfWorkUnits > m_GlobalMaximumNumberOfThreads)
+  if (m_NumberOfWorkUnits > MultiThreaderBase::GetGlobalMaximumNumberOfThreads())
   {
-    m_NumberOfWorkUnits = m_GlobalMaximumNumberOfThreads;
+    m_NumberOfWorkUnits = MultiThreaderBase::GetGlobalMaximumNumberOfThreads();
   }
 
   for (thread_loop = 0; thread_loop < m_NumberOfWorkUnits; ++thread_loop)


dzenanz (approved):
LGTM

Modules/Core/Common/src/itkPlatformMultiThreaderSingle.cxx
```diff
@@ -39,9 +39,9 @@ PlatformMultiThreader::MultipleMethodExecute()
   ThreadIdType thread_loop;
 
   // obey the global maximum number of threads limit
-  if (m_NumberOfWorkUnits > m_GlobalMaximumNumberOfThreads)
+  if (m_NumberOfWorkUnits > MultiThreaderBase::GetGlobalMaximumNumberOfThreads())
   {
-    m_NumberOfWorkUnits = m_GlobalMaximumNumberOfThreads;
+    m_NumberOfWorkUnits = MultiThreaderBase::GetGlobalMaximumNumberOfThreads();
```
N-Dekker, line 44: For my understanding, isn't the GlobalMaximumNumberOfThreads always exactly `1`, for MultiThreaderSingle?

fabian-philips: Yes, that is correct, itkThreadSupports.h sets

constexpr size_t ITK_MAX_THREADS = 1;

in case of non-pthreads/win32threads (line 78). However, the proposed change does not change the style compared to the other variants - and the 1 from ITK_MAX_THREADS is retrieved via header functions only (same performance).

N-Dekker: Ah, thanks for explaining, @fabian-philips! I see now, the same GetGlobalMaximumNumberOfThreads() check in itkPlatformMultiThreaderPosix.cxx and itkPlatformMultiThreaderWindows.cxx, indeed!

N-Dekker (approved):
Thanks @fabian-philips!

I hope the code can later still be further simplified, I mean, instead of those four lines of code:

```
  if (m_NumberOfWorkUnits > m_GlobalMaximumNumberOfThreads)
  {
    m_NumberOfWorkUnits = m_GlobalMaximumNumberOfThreads;
  }

```
It seems nicer to me to just have a single line of code: `m_NumberOfWorkUnits = 1;` This file is specific to the single-threaded case, so it may just assume that the number of threads is `1`, right? Anyway, that may be something for another pull request  😃  For now: approved 👍

PR #3950, opened 2023-03-07
STYLE: Prefer C++11 zero initializer to ZeroValue

hjmjohnson:
Prefer to use standard C++ language initializers to the itk::NumericValues<>::ZeroValue() initializers.

## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [X] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)

PR #4460, opened 2024-02-12
Replace `NumericTraits<T>::ZeroValue()` with `T{}` or `0U`

N-Dekker:
For any type `T` that is supported by `NumericTraits<T>`, the expression `NumericTraits<T>::ZeroValue()` is just equivalent to `T{}`. `T{}` appears preferred, because it is more generic, and it may yield faster code than the corresponding `NumericTraits<T>::ZeroValue()` function call.

    Find what: itk::NumericTraits<(\w+)>::ZeroValue\(\)
    Find what: NumericTraits<(\w+)>::ZeroValue\(\)
    Replace with: $1{}
    Filters: itk*.hxx;itk*.h;itk*.cxx
    Directory: ITK\Modules
    [v] Match case
    (*) Regular expression

Excluded "itkNumericTraitsTest.cxx", because it purposely calls `NumericTraits::ZeroValue()` as part of the test.

Follow-up to pull request #4017 commit 5ac803e290fb2d5b642ed4e23622f9236b50fdeb "STYLE: Replace `T var{ NumericTraits<T>::ZeroValue() }` with `T var{}`"

Also follows up to:
- pull request #3950 by Hans Johnson (@hjmjohnson)

### DIFF
--- a/Modules/Numerics/Optimizersv4/include/itkGradientDescentLineSearchOptimizerv4.hxx
+++ b/Modules/Numerics/Optimizersv4/include/itkGradientDescentLineSearchOptimizerv4.hxx
@@ -28,7 +28,7 @@ GradientDescentLineSearchOptimizerv4Template<
 {
   this->m_MaximumLineSearchIterations = 20;
   this->m_LineSearchIterations = 0U;
-  this->m_LowerLimit = itk::NumericTraits<TInternalComputationValueType>::ZeroValue();
+  this->m_LowerLimit = TInternalComputationValueType{};
   this->m_UpperLimit = 5.0;
   this->m_Phi = 1.618034;
   this->m_Resphi = 2 - this->m_Phi;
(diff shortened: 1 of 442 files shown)


Modules/Numerics/Optimizersv4/include/itkGradientDescentLineSearchOptimizerv4.hxx
```diff
@@ -28,7 +28,7 @@ GradientDescentLineSearchOptimizerv4Template<
 {
   this->m_MaximumLineSearchIterations = 20;
   this->m_LineSearchIterations = NumericTraits<unsigned int>::ZeroValue();
```
issakomi, line 30: I think this is intentionally left out, I guess, `unsigned int{}` doesn't work, `(unsigned int){}` seems to compile, but probably `0U` is even better, IMHO.
P.S. There are not many similar situations (e.g. _unsigned char_, _long double_, etc), some are in itkFreeSurferAsciiMeshIO.cxx, AFAIK.

N-Dekker: @issakomi Please check the two commits that I added to address your comments. Also please consider approving this pull request 😃

dzenanz (approved):

issakomi (approved):

PR #4467, opened 2024-02-16
STYLE: Replace `itk::NumericTraits<unsigned int>::OneValue()` with `1`

N-Dekker:
Follow-up to pull request https://github.com/InsightSoftwareConsortium/ITK/pull/4460 commit 5d746cc0bc5efa6655580ee63c36cefde156a3c2
"STYLE: Replace `NumericTraits<unsigned int>::ZeroValue()` with `0U`"

### DIFF
--- a/Modules/IO/MeshBYU/src/itkBYUMeshIO.cxx
+++ b/Modules/IO/MeshBYU/src/itkBYUMeshIO.cxx
@@ -167,12 +167,12 @@ BYUMeshIO::ReadMeshInformation()
   // Set default point pixel component and point pixel type
   this->m_PointPixelComponentType = IOComponentEnum::FLOAT;
   this->m_PointPixelType = IOPixelEnum::SCALAR;
-  this->m_NumberOfPointPixelComponents = itk::NumericTraits<unsigned int>::OneValue();
+  this->m_NumberOfPointPixelComponents = 1;
 
   // Set default cell pixel component and point pixel type
   this->m_CellPixelComponentType = IOComponentEnum::FLOAT;
   this->m_CellPixelType = IOPixelEnum::SCALAR;
-  this->m_NumberOfCellPixelComponents = itk::NumericTraits<unsigned int>::OneValue();
+  this->m_NumberOfCellPixelComponents = 1;
 
   inputFile.close();
 }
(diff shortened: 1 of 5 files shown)


dzenanz (approved):
Does `1` vs `1u` make any difference?

N-Dekker:
> Does `1` vs `1u` make any difference?

In practice, when you assign the value (`1` or `1u`) to a variable of a built-in type, it does not make a difference at all. So it's just a matter of style 😃

blowekamp:
I am pretty sure at some point there was a compiler warning flag used that warned about conversions from int to unsigned int, and the draconian usage itk::NumericTraits was used to help determine where real conversion issues were occurring.

N-Dekker:
itkOBJMeshIO.cxx does `m_NumberOfCellPixelComponents = 3` without `u` as well.

N-Dekker:
If anyone insists on `1U`, that's OK to me as well, I can amend the commit if anyone wants to. I'm just _slightly_ in favor of `1`. Because of its simplicity 😃

N-Dekker:
> I am pretty sure at some point there was a compiler warning flag used that warned about conversions from int to unsigned int, and the draconian usage itk::NumericTraits was used to help determine where real conversion issues were occurring.

Thanks for explaining, @blowekamp. Nowadays I think the main reason for using itk::NumericTraits would be to support pixel types in a generic way. Specifically, `itk::NumericTraits<PixelType>::OneValue()` returns `1` when the pixel type is int, while it returns `{1, 1, 1}` for an RGB pixel type.

blowekamp:
> > I am pretty sure at some point there was a compiler warning flag used that warned about conversions from int to unsigned int, and the draconian usage itk::NumericTraits was used to help determine where real conversion issues were occurring.
> 
> Thanks for explaining, @blowekamp. Nowadays I think the main reason for using itk::NumericTraits would be to support pixel types in a generic way. Specifically, `itk::NumericTraits<PixelType>::OneValue()` returns `1` when the pixel type is `int`, while it returns `{1, 1, 1}` for an RGB pixel type.

The other case to consider is with itk::SizeType or std::std size type. The size of "long" and "long long" vary across systems and can change based on the word size for the architecture on some, but not other platforms.

N-Dekker:
> > Nowadays I think the main reason for using itk::NumericTraits would be to support pixel types in a generic way. Specifically, `itk::NumericTraits<PixelType>::OneValue()` returns `1` when the pixel type is `int`, while it returns `{1, 1, 1}` for an RGB pixel type.
> 
> The other case to consider is with itk::SizeType or std::std size type. The size of "long" and "long long" vary across systems and can change based on the word size for the architecture on some, but not other platforms.

OK, but for a built-in integer type (or a typedef or _alias_ of a built-in integer type), `std::numeric_limits<T>` typically offers everything one may need. So in that case, I would prefer `std::numeric_limits<T>`, rather than `itk::NumericTraits<T>`.

hjmjohnson (approved):

PR #4468, opened 2024-02-16
ENH: Add SameImageGridAs convenience method to ImageBase

### LINKED ISSUE #3536: Expose ImageToImageFilter::VerifyInputInformation() as a usable True/False function
### Description

When checking a large number of images before processing, say by reading just the metadata rather than the voxel data, it would be worthwhile to be able to check if they're all "in the same space" before doing expensive processing. 

Exposing the logic of `ImageToImageFilter::VerifyInputInformation()` as a function for using a test on a pair of images would be helpful in this regard.

### Expected behavior

`newFunction(img1, img2)` returns True if the images  would pass the same test in ImageToImage Filter (i.e. same origins, spacings, Directions etc)

### Actual behavior

NA

### Additional Information

Inspired by question here, https://discourse.itk.org/t/quick-test-for-are-images-in-the-same-space-aka-overalpping-simpleitk/5249/5

dzenanz:
Closes #3536.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkImageBase.h
+++ b/Modules/Core/Common/include/itkImageBase.h
@@ -768,6 +768,24 @@ public:
   bool
   VerifyRequestedRegion() override;
 
+  /** Checks whether the images' pixels with the same index occupy the same physical space.
+   * Compares the origin, spacing, and direction for equality within provided tolerances.
+   * There is no check for valid regions in between the images. */
+  bool
+  IsCongruentImageGeometry(const ImageBase * otherImage, double coordinateTolerance, double directionTolerance) const;
+
+  /** Check whether this image and the other image have the same grid in physical space.
+   * Compares largest possible regions for equality, and the origin, spacing,
+   * and direction cosines for equality within provided tolerances.
+   * See also: ImageToImageFilter, namely:
+   * https://github.com/InsightSoftwareConsortium/ITK/blob/v5.3.0/Modules/Core/Common/include/itkImageToImageFilter.h#L78-L92
+   * https://github.com/InsightSoftwareConsortium/ITK/blob/v5.3.0/Modules/Core/Common/src/itkImageToImageFilterCommon.cxx#L26-L27
+   */
+  bool
+  IsSameImageGeometryAs(const ImageBase * otherImage,
+                        double            coordinateTolerance = 1e-6,
+                        double            directionTolerance = 1e-6) const;
+
   /** INTERNAL This method is used internally by filters to copy meta-data from
    * the output to the input. Users should not have a need to use this method.
    *
--- a/Modules/Core/Common/include/itkImageBase.hxx
+++ b/Modules/Core/Common/include/itkImageBase.hxx
@@ -399,6 +399,38 @@ ImageBase<VImageDimension>::VerifyRequestedRegion()
 
 
 template <unsigned int VImageDimension>
+bool
+ImageBase<VImageDimension>::IsCongruentImageGeometry(const ImageBase * otherImage,
+                                                     double            coordinateTolerance,
+                                                     double            directionTolerance) const
+{
+  // check that the image occupy the same physical space, and that
+  // each index is at the same physical location
+
+  // tolerance for origin and spacing depends on the size of pixel
+  // tolerance for directions a fraction of the unit cube.
+  const SpacePrecisionType coordinateTol =
+    itk::Math::abs(coordinateTolerance * this->GetSpacing()[0]); // use first dimension spacing
+
+  return this->GetOrigin().GetVnlVector().is_equal(otherImage->GetOrigin().GetVnlVector(), coordinateTol) &&
+         this->GetSpacing().GetVnlVector().is_equal(otherImage->GetSpacing().GetVnlVector(), coordinateTol) &&
+         this->GetDirection().GetVnlMatrix().as_ref().is_equal(otherImage->GetDirection().GetVnlMatrix().as_ref(),
+                                                               directionTolerance);
+}
+
+
+template <unsigned int VImageDimension>
+bool
+ImageBase<VImageDimension>::IsSameImageGeometryAs(const ImageBase * otherImage,
+                                                  double            coordinateTolerance,
+                                                  double            directionTolerance) const
+{
+  return this->IsCongruentImageGeometry(otherImage, coordinateTolerance, directionTolerance) &&
+         this->GetLargestPossibleRegion() == otherImage->GetLargestPossibleRegion();
+}
+
+
+template <unsigned int VImageDimension>
 void
 ImageBase<VImageDimension>::SetBufferedRegion(const RegionType & region)
 {
--- a/Modules/Core/Common/include/itkImageToImageFilter.hxx
+++ b/Modules/Core/Common/include/itkImageToImageFilter.hxx
@@ -184,10 +184,7 @@ ImageToImageFilter<TInputImage, TOutputImage>::VerifyInputInformation() ITKv5_CO
       const SpacePrecisionType coordinateTol =
         itk::Math::abs(this->m_CoordinateTolerance * inputPtr1->GetSpacing()[0]); // use first dimension spacing
 
-      if (!inputPtr1->GetOrigin().GetVnlVector().is_equal(inputPtrN->GetOrigin().GetVnlVector(), coordinateTol) ||
-          !inputPtr1->GetSpacing().GetVnlVector().is_equal(inputPtrN->GetSpacing().GetVnlVector(), coordinateTol) ||
-          !inputPtr1->GetDirection().GetVnlMatrix().as_ref().is_equal(inputPtrN->GetDirection().GetVnlMatrix().as_ref(),
-                                                                      this->m_DirectionTolerance))
+      if (!inputPtr1->IsCongruentImageGeometry(inputPtrN, m_CoordinateTolerance, m_DirectionTolerance))
       {
         std::ostringstream originString, spacingString, directionString;
         if (!inputPtr1->GetOrigin().GetVnlVector().is_equal(inputPtrN->GetOrigin().GetVnlVector(), coordinateTol))


N-Dekker:
Interesting, thanks @dzenanz! Was there a specific reason to start working on this now? The [issue](https://github.com/InsightSoftwareConsortium/ITK/issues/3536) and [discussion](https://discourse.itk.org/t/quick-test-for-are-images-in-the-same-space-aka-overalpping-simpleitk/5249) were already a while ago (August 2022).

Recently I was actually thinking of a class that represents those five image properties together, `{ directionMatrix, index, size, spacing, originPoint }`. Internally in elastix, I already added a struct named `ImageDomain`, basically:
```cpp
struct ImageDomain
{
  using ImageBaseType = itk::ImageBase<VDimension>;

  using DirectionType = typename ImageBaseType::DirectionType;
  using IndexType = typename ImageBaseType::IndexType;
  using SizeType = typename ImageBaseType::SizeType;
  using SpacingType = typename ImageBaseType::SpacingType;
  using PointType = typename ImageBaseType::PointType;

  DirectionType direction;
  IndexType     index;
  SizeType      size;
  SpacingType   spacing;
  PointType     origin;
};
```

Including `operator==`, `operator!=`, and interface to `itk::Image`. So it would allow comparing image domains of two images by doing:

```cpp
    if ( ImageDomain(image1) == ImageDomain(image2) ) // equal image domains! 
```
And it would also allow putting the image domain properties into an `itk::Image`, e.g.:
```cpp
    imageDomain.ToImage(image);
```

https://github.com/SuperElastix/elastix/blob/0f86c8774fcb5ae362c8bdabc9bc6e13e9bfca40/Core/Main/GTesting/elxCoreMainGTestUtilities.h#L299-L318  

But I did't know if it would be of interest to other ITK users, so it's just internal to elastix now  🤷

Modules/Core/Common/include/itkImageBase.h
```diff
@@ -768,6 +768,19 @@ class ITK_TEMPLATE_EXPORT ImageBase : public DataObject
...
+  bool
+  SamePhysicalSpaceAs(const ImageBase * otherImage, double coordinateTolerance, double directionTolerance) const;
+
+  /** Check whether this image and the other image have the same grid in physical space.
+   * Compares largest possible regions for equality, and the origin, spacing,
+   * and direction cosines for equality within provided tolerances. */
+  bool
+  SameImageGridAs(const ImageBase * otherImage,
+                  double            coordinateTolerance = 1e-6,
+                  double            directionTolerance = 1e-6) const;
```
N-Dekker, line 782: Is this default of `1e-6` arbitrary? If so, what about using `DBL_EPSILON` instead, as a default? Personally I would prefer zero-tolerance (`= 0.0`) by default.

Modules/Core/Common/include/itkImageBase.hxx
```diff
@@ -398,6 +398,38 @@ ImageBase<VImageDimension>::VerifyRequestedRegion()
...
+  return this->GetOrigin().GetVnlVector().is_equal(otherImage->GetOrigin().GetVnlVector(), coordinateTol) &&
+         this->GetSpacing().GetVnlVector().is_equal(otherImage->GetSpacing().GetVnlVector(), coordinateTol) &&
+         this->GetDirection().GetVnlMatrix().as_ref().is_equal(otherImage->GetDirection().GetVnlMatrix().as_ref(),
+                                                               directionTolerance);
+}
+
+
+template <unsigned int VImageDimension>
+bool
+ImageBase<VImageDimension>::SameImageGridAs(const ImageBase * otherImage,
```
N-Dekker, line 424: Please consider declaring the `otherImage` parameter as a reference (`const ImageBase &`).

Modules/Core/Common/include/itkImageToImageFilter.hxx
```diff
@@ -184,10 +184,7 @@ ImageToImageFilter<TInputImage, TOutputImage>::VerifyInputInformation() ITKv5_CO
...
         itk::Math::abs(this->m_CoordinateTolerance * inputPtr1->GetSpacing()[0]); // use first dimension spacing
 
-      if (!inputPtr1->GetOrigin().GetVnlVector().is_equal(inputPtrN->GetOrigin().GetVnlVector(), coordinateTol) ||
-          !inputPtr1->GetSpacing().GetVnlVector().is_equal(inputPtrN->GetSpacing().GetVnlVector(), coordinateTol) ||
-          !inputPtr1->GetDirection().GetVnlMatrix().as_ref().is_equal(inputPtrN->GetDirection().GetVnlMatrix().as_ref(),
-                                                                      this->m_DirectionTolerance))
+      if (!inputPtr1->SamePhysicalSpaceAs(inputPtrN, m_CoordinateTolerance, m_DirectionTolerance))
       {
         std::ostringstream originString, spacingString, directionString;
         if (!inputPtr1->GetOrigin().GetVnlVector().is_equal(inputPtrN->GetOrigin().GetVnlVector(), coordinateTol))
```
N-Dekker, line 190: Maybe the redundancy of duplicate `is_equal` calls that was in the original code already should also be addressed. I mean, it now calls `is_equal` twice for each of the three properties, Origin, Spacing, and Direction. I think it would be better to store the intermediate `is_equal` results, to avoid duplicated `is_equal` calls: 

```cpp
      const bool equalOrigin = inputPtr1->GetOrigin().GetVnlVector().is_equal(...);
      const bool equalSpacing = inputPtr1->GetSpacing().GetVnlVector().is_equal(...);
      const bool equalDirection = inputPtr1->GetDirection().GetVnlMatrix().as_ref().is_equal(...);

      if (!equalOrigin || !equalSpacing || !equalDirection)
      {
        std::ostringstream originString, spacingString, directionString;
        if (!equalOrigin) {...}
        if (!equalSpacing) {...}
        if (!equalDirection) {...}
```

blowekamp:
Consider making the function name "IsSameImageGridAs", to make it clearer that this function returns a boolean and does not make a change to the image. And adding the "Is" prefix to the other image. I think the `Image *` is more consistent with the ITK style.

The Image class has member variable for the coordinate and direction tolerance. Perhaps it would make more since to use those values instead of a parameter. The might be needed if this functions were static.

Modules/Core/Common/include/itkImageToImageFilter.hxx
N-Dekker: I am more concerned about keeping the duplicated `is_equal` calls in sync. Each of the three conditions in `if (!equalOrigin || !equalSpacing || !equalDirection)` should exactly match the individual checks below there, `if (!equalOrigin)`, `if (!equalSpacing)`, and `if (!equalDirection)`.

N-Dekker:
@blowekamp 
> I think the `Image *` is more consistent with the ITK style.

Thanks for your comment, Bradley. `Image *` is certainly more consistent with the _old_ ITK style. But ITK's style is very much modernized in the last few years, starting with the release of ITK 5.0.0 and the upgrade to C++11. We have embraced modern C++ range iteration, range-based `for` loops, `{}` initializers, lambda's, etc., etc., and modernized source file formatting by clang-format. I would see the use of a C++ reference `Image &` in a case like this as another little step ahead.

N-Dekker:
> Having the support for tolerance during comparison would be highly desired, as different image formats might have different internal precisions.

Thanks for explaining @dzenanz. Do you actually have an example of such a use case? Having images of different image formats that should be considered to have the same grid or physical space, even while `image1.spacing == image2.spacing` would return false? I'm really interested to see such cases, to see where those rounding errors come from.

hjmjohnson:
> I don't have a case ready, but I do remember that comparison tolerance was introduced primarily because of that. E.g., load image.nrrd (with spacing/origin/orientation specified to 16 decimal places), save it as image.nii, then load image.nii, and it does not compare equal because some decimal places were discarded (or some similar situation). If you look at the commit message of the commit which added these tolerances (and associated PR discussion), you might find a case which demonstrates the problem.

@dzenanz Nifti (and DICOM) are the problems here.  They do not have the same level of precision as internal ITK structures (or NRRD) do.

dzenanz:
Both 32 vs 64 bit float, and floating vs decimal representations are potential problems. I think that NRRD/MetaIO writers were not writing full precision by default at some point, and that triggered renewed interest in `DoubleConversion` library.

blowekamp:
> @hjmjohnson
> 
> > @dzenanz Nifti (and DICOM) are the problems here. They do not have the same level of precision as internal ITK structures (or NRRD) do.
> 
> Thanks for the insights! So would it be 32-bit float precision versus 64-bit double precision? Or would it be more like a fixed-point decimal string representation versus floating point representation? Just for my understanding 😃

You are welcome to experiment with setting the GlobalDefaultTolerances to see how it effect the test suits and applications.

The value selection was not arbitrary. As I recall we did not even achieve float eps reading and writing floating point number, much less across different image format with different representations of the direction cosine matrix. Float eps is about 1e-7 and this default here is 1e-6, which was needed so to keep everything working. I do not recall if a specific cases were integrated into an ITK test or is was reported, and then relaxed.

The design was  to allow user/developer to set the tolerance is they which to enforce a pipeline to have higher precision. Hence why I am advocating to keep using the member variable that are set for these tolerances and not introduce additional parameters to the functions.

N-Dekker:
Is everybody happy about the term "ImageGrid" in this context? I'm just asking, because I'm just not sure! I mean, is it obvious to everyone that "ImageGrid" means those things together: largestPossibleImageRegion (`itk::Index<N>` + `itk::Size<N>`) + direction (`itk::Matrix<N, N>`) + origin (`Point<double, N>`) + spacing (`Vector<double, N>`)?

The documentation of `ImageBase` also mentions the term "the geometry of an image", is that basically the same as "ImageGrid"? It says: _The geometry of an image is defined by its position, orientation, spacing, and extent._ 

https://github.com/InsightSoftwareConsortium/ITK/blob/ef28b0f9f5024814ccd2c6285cd66e245ea244e2/Modules/Core/Common/include/itkImageBase.h#L57-L58

A colleague of mine considered the term "coordinate system" (or "image coordinate system") instead. While I myself introduced "ImageDomain", internally in elastix: https://github.com/SuperElastix/elastix/blob/0f86c8774fcb5ae362c8bdabc9bc6e13e9bfca40/Core/Main/GTesting/elxCoreMainGTestUtilities.h#L299-L318

What do you think? Any preference?

dzenanz:
`ImageGeometry` also sounds acceptable. `ImageGrid` is already used for a group of filters which modify it.

blowekamp:
> `ImageGeometry` also sounds acceptable. `ImageGrid` is already used for a group of filters which modify it.

I have used the term "congruent" before to describe these operations. To me "IsCongruent" describes the "SamePysicalSpace" operations, and "IsCongruenttImageGeometry"... Or something similar.

N-Dekker:
Interesting 😃 What about `HasCongruentImageGrid` and `HasCongruentPhysicalSpace`, maybe?

blowekamp:
> Interesting 😃 What about `HasCongruentImageGrid` and `HasCongruentPhysicalSpace`, maybe?

The first two places I looked for support of ITK convention was the itk::ImageRegion, and itk::SpacialObject. Both use "Is" as a prefix for boolean queries. Far from a complete audit, or a formal documentation of style preference.

A quick grep of Core/Common: "git grep  bool.\*Is -- Modules/Core/Common/ |wc" showed 144 of Is vs 5 of Has.

dzenanz:
> The Image class has member variable for the coordinate and direction tolerance.

Unfortunately, `ImageToImageFilter` is the one with [those tolerances](https://github.com/InsightSoftwareConsortium/ITK/blob/v5.4rc02/Modules/Core/Common/include/itkImageToImageFilter.h#L184-L217). I don't think we want that dependency, even if we could technically have it (I am not sure that we can).

N-Dekker:
Just wondering, should both member functions be public? I mean the one that compares _all_ of those properties (largestPossibleImageRegion + direction + origin + spacing) and the one that just compares three of them (direction + origin + spacing)?  Are both of these member functions useful for the end user?

dzenanz:
The one which compares everything will probably be used more often. But as the case of `ImageToImageFilter` shows, both are useful.

blowekamp:
> Just wondering, should both member functions be public? I mean the one that compares _all_ of those properties (largestPossibleImageRegion + direction + origin + spacing) and the one that just compares three of them (direction + origin + spacing)? Are both of these member functions useful for the end user?

Yes. The latter one is used in the ImageToImage filter. This is sufficient when there is another check the image regions being operated match.  This can be useful when one image is a sub region of another and the operation is some intersection of the regions.

N-Dekker:
> The latter one is used in the ImageToImage filter.

Well... ImageToImageFilter needs to know specifically, which of the three (direction, origin, spacing) are equal.  Otherwise ImageToImageFilter has to figure it out again. Sorry for repeating myself (from https://github.com/InsightSoftwareConsortium/ITK/pull/4468#discussion_r1494511479). I think it would be better if the direction+origin+spacing comparison function would tell specifically which of the properties are equal, and which ones are different. Maybe by returning a struct, `{ bool equalDirection, bool equalOrigin, bool equalSpacing }`?

dzenanz:
> returning a struct

I am afraid that would reduce utility of the function. Whoever needs to know which one is different, can do the comparison themselves (as is the case of `ImageToImageFilter`).

Modules/Core/Common/include/itkImageBase.hxx
```diff
@@ -398,6 +398,38 @@ ImageBase<VImageDimension>::VerifyRequestedRegion()
...
+                                                     double            coordinateTolerance,
+                                                     double            directionTolerance) const
+{
+  // check that the image occupy the same physical space, and that
+  // each index is at the same physical location
+
+  // tolerance for origin and spacing depends on the size of pixel
+  // tolerance for directions a fraction of the unit cube.
+  const SpacePrecisionType coordinateTol =
+    itk::Math::abs(coordinateTolerance * this->GetSpacing()[0]); // use first dimension spacing
```
N-Dekker, line 413: Is there a specific reason to only use the first spacing dimension, instead of all of them? Otherwise I would rather use the smallest of them (something like `*std::min_element(m_Spacing.begin(), m_Spacing.end())`)

dzenanz: Simplicity inherited from `ImageToImageFilter`'s code. Good for backwards compatibility.

N-Dekker: If backwards compatibility is the only reason, maybe `ITK_FUTURE_REMOVE_LEGACY` can be of help  😃

N-Dekker:
> Both 32 vs 64 bit float, and floating vs decimal representations are potential problems. I think that NRRD/MetaIO writers were not writing full precision by default at some point, and that triggered renewed interest in `DoubleConversion` library.

Ah, now I see, DICOM represents PixelSpacing values as Decimal String (DS) of maximum 16 characters (https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_6.2.html), whereas up to 25 characters may be needed to losslessly write a 64-bit `double` to string, using the  `DoubleConversion` library (https://github.com/google/double-conversion/issues/146). So DICOM may indeed introduce rounding errors that way  🤷

hjmjohnson:
@N-Dekker Exactly!  I recently ran into this problem and it was quite disturbing.  I need to write compliant dicom images, and I was using more than 16 characters.

N-Dekker:
@hjmjohnson I just looked at the possible rounding errors introduced by conversion from `double` to DICOM DecimalString (DS). It appears that a double-to-DecimalString conversion function is publicly available through the ITK interface, written by Mathieu Malaterre (@malaterre)  as part of GDCM:

https://github.com/InsightSoftwareConsortium/ITK/blob/adcc46bf9600b78bda422c49a8aca4611c1751d1/Modules/ThirdParty/GDCM/src/gdcm/Source/DataStructureAndEncodingDefinition/gdcmElement.h#L348-L349

I tried the following test:

```cpp
void
TestRoundTrip(const double inputValue)
{
  auto         roundTrippedValue = std::numeric_limits<double>::quiet_NaN();
  char         buf[32]{};
  // Convert input value to DICOM DecimalString:
  gdcm::x16printf(buf, 16, inputValue);
  std::istringstream inputStringStream(buf);
  inputStringStream >> roundTrippedValue;
  EXPECT_EQ(roundTrippedValue, inputValue)
    << "Relative difference: " << (std::abs((roundTrippedValue - inputValue) / inputValue));
}
```

The _highest_ relative difference between the input value and the round tripped value that I saw was `~6.7e-10`, which was for input value `-std::numeric_limits<double>::min()` (that is `-DBL_MIN`). Output:

```
  roundTrippedValue
    Which is: -2.22507386e-308
  inputValue
    Which is: -2.2250738585072014e-308
Relative difference: 6.7089844790757525e-10
```

(Well, I ignored the round trip for `DBL_MAX`, which had a relative difference was "infinity" 😸 )

Just for your information  😃

dzenanz:
We have a good discussion. Time for another round of clearly actionable feedback? Or is it time to move to approval?

Modules/Core/Common/include/itkImageBase.h
```diff
@@ -768,6 +768,19 @@ class ITK_TEMPLATE_EXPORT ImageBase : public DataObject
   bool
   VerifyRequestedRegion() override;
 
+  /** Check whether this image and the other image occupy the same physical space.
```
blowekamp, line 771: May be something like this:
"Checks whether the images' pixels with the same index occupy the same physical space.

Compares the origin, spacing, and direction for equality within provided tolerances. There is no
check for valid regions in between the images."

Modules/Core/Common/include/itkImageBase.h
```diff
@@ -768,6 +768,19 @@ class ITK_TEMPLATE_EXPORT ImageBase : public DataObject
...
+  bool
+  IsCongruentImageGeometry(const ImageBase * otherImage, double coordinateTolerance, double directionTolerance) const;
+
+  /** Check whether this image and the other image have the same grid in physical space.
+   * Compares largest possible regions for equality, and the origin, spacing,
+   * and direction cosines for equality within provided tolerances. */
+  bool
+  IsSameImageGeometryAs(const ImageBase * otherImage,
+                        double            coordinateTolerance = 1e-6,
+                        double            directionTolerance = 1e-6) const;
```
blowekamp, line 782: Perhaps we should just have no defaults here, and have a documentation link to the ImageToImage defaults?

blowekamp (approved):
Mostly documentation suggestions.

Modules/Core/Common/include/itkImageBase.h
dzenanz: That will be an extra step for common use cases. I think we should keep the defaults, but add a link to the `ImageToImage` defaults, when someone needs to learn about the tolerances.

blowekamp: Perhaps we could use the same global constant is [both](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/src/itkImageToImageFilterCommon.cxx#L26) places then?

Maybe add a common constant value in Macro.h?

dzenanz: That is a good idea. But I am already tired of this small PR. Can we do that in a different PR? Otherwise it would be good if someone took over this PR.

PR #4469, opened 2024-02-19
PERF: FUTURE: Default default-constructors of `RGBPixel` and `RGBAPixel`

N-Dekker:
Will make `image.Allocate()` and `image.Allocate(false)` calls much, much
faster, for RGB and RGBA images, when `ITK_FUTURE_LEGACY_REMOVE` is enabled.
An `Allocate()` call on a 1024x1024x1024 RGB image was observed to take more than
1.3 sec. before this commit, and less than 0.001 sec. after this commit, having
`ITK_FUTURE_LEGACY_REMOVE=ON`, using Visual C++ 2019 Release. Even
`image.Allocate(true)` (`initializePixels` = true) appears slightly faster now
with `ITK_FUTURE_LEGACY_REMOVE`.

Adjusted unit tests tests from itkCommonTypeTraitsGTest, because `RGBPixel` and
`RGBAPixel` become "trivial" types when they get defaulted default-constructors.

### DIFF
--- a/Modules/Core/Common/include/itkRGBAPixel.h
+++ b/Modules/Core/Common/include/itkRGBAPixel.h
@@ -78,7 +78,12 @@ public:
 
   /** Default-constructor.
    * \note The other five "special member functions" are defaulted implicitly, following the C++ "Rule of Zero". */
+#ifdef ITK_FUTURE_LEGACY_REMOVE
+  RGBAPixel() = default;
+#else
   RGBAPixel() { this->Fill(0); }
+#endif
+
   /** Pass-through constructor for the Array base class. */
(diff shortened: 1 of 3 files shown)


dzenanz (approved):
Thanks for explaining. I am on the fence of legacy vs future legacy.

N-Dekker:
>  I am on the fence of legacy vs future legacy.

Thanks @dzenanz. I am also on the fence of legacy vs future legacy.  😸 But only if someone is strongly in favor of LEGACY_REMOVE, instead of  FUTURE_LEGACY_REMOVE, I would still adjust the PR. @hjmjohnson Is FUTURE_LEGACY_REMOVE now OK to you?

PR #4473, opened 2024-02-21
PERF: FUTURE: Default default-constructor of `SymmetricSecondRankTensor`

N-Dekker:
Follow-up to pull request https://github.com/InsightSoftwareConsortium/ITK/pull/4469 commit 755cd10e5421d41f4004be101c888d5ba24be8e4
"PERF: FUTURE: Default default-constructors of `RGBPixel` and `RGBAPixel`"

### DIFF
--- a/Modules/Core/Common/include/itkSymmetricSecondRankTensor.h
+++ b/Modules/Core/Common/include/itkSymmetricSecondRankTensor.h
@@ -104,7 +104,11 @@ public:
 
   /** Default-constructor.
    * \note The other five "special member functions" are defaulted implicitly, following the C++ "Rule of Zero". */
+#ifdef ITK_FUTURE_LEGACY_REMOVE
+  SymmetricSecondRankTensor() = default;
+#else
   SymmetricSecondRankTensor() { this->Fill(0); }
+#endif
 
   SymmetricSecondRankTensor(const ComponentType & r) { this->Fill(r); }
 
--- a/Modules/Core/Common/test/itkCommonTypeTraitsGTest.cxx
+++ b/Modules/Core/Common/test/itkCommonTypeTraitsGTest.cxx
@@ -74,6 +74,13 @@ TEST(CommonTypeTraits, RGBPixelIsPOD)
   EXPECT_TRUE(std::is_trivial_v<T>);
   EXPECT_TRUE(std::is_standard_layout_v<T>);
 }
+
+TEST(CommonTypeTraits, SymmetricSecondRankTensorIsPOD)
+{
+  using T = itk::SymmetricSecondRankTensor<float, 3>;
+  EXPECT_TRUE(std::is_trivial_v<T>);
+  EXPECT_TRUE(std::is_standard_layout_v<T>);
+}
 #else
 TEST(CommonTypeTraits, RGBAPixelIsNotPOD)
 {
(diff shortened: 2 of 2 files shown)


Modules/Core/Common/include/itkSymmetricSecondRankTensor.h
```diff
@@ -104,7 +104,11 @@ class ITK_TEMPLATE_EXPORT SymmetricSecondRankTensor : public FixedArray<TCompone
 
   /** Default-constructor.
    * \note The other five "special member functions" are defaulted implicitly, following the C++ "Rule of Zero". */
+#ifdef itkSymmetricSecondRankTensor_h // TODO Must be: #ifdef ITK_FUTURE_LEGACY_REMOVE
```
N-Dekker, line 107: Note: the initial CI run is expected to fail, because of the incorrect `#ifdef`. I just want to see if defaulting `SymmetricSecondRankTensor()` would cause any test failures, other than the GTest `CommonTypeTraits.SymmetricSecondRankTensorIsNotPOD`.

blowekamp:
I assume this PR is similar to the RGBPixel improvement with the goal of improving the performance of the Image::Allocate method. Perhaps this is a case there the optimization should go into the Allocate method. In SimpleITK and when interfacing with other libraries and buffers, the Image container buffer is considered a raw buffer. Perhaps if certain conditions, c++ traits or itk PixelTraits then a special optimization in this method could occur where the buffer is initialized as some type of raw buffer, then case to the component type. 

These initialized objects can be an insidious thing to track down as the behavior is not deterministic and varies with debug and release builds. So I have some concern with the proposed future changes.

Modules/Core/Common/test/itkCommonTypeTraitsGTest.cxx
```diff
@@ -74,6 +74,13 @@ TEST(CommonTypeTraits, RGBPixelIsPOD)
   EXPECT_TRUE(std::is_trivial_v<T>);
   EXPECT_TRUE(std::is_standard_layout_v<T>);
 }
+
```
dzenanz, line 77:
```suggestion
#ifdef ITK_FUTURE_LEGACY_REMOVE
```

dzenanz: Oops, if is provided 20 lines earlier.

dzenanz (approved):

N-Dekker:
> I assume this PR is similar to the RGBPixel improvement with the goal of improving the performance of the Image::Allocate method.

Indeed, I have indeed seen huge ("future") performance improvements of `Image::Allocate` from the similar RGBPixel PR (#4469), and I would expect similar ("future") performance improvements for SymmetricSecondRankTensor, with this commit. But it's not just that. It's also a matter of consistency between pixel types:

```cpp
    TPixel pixel1; // May not be initialized, after this change 
    TPixel pixel2{}; // Properly zero-initialized.
    TPixel pixel3 = TPixel{}; // Properly initialized.
    TPixel pixel4 = TPixel(); // Properly initialized.
```	
As I mentioned at https://github.com/InsightSoftwareConsortium/ITK/pull/4469#discussion_r1496080480

Note that `std::make_unique<TPixel>()` and `std::make_shared<TPixel>()` will zero-initialize the created pixel(s) anyway, also with this PR.

Slightly off-topic, I think it would help if `itk::Image` would have an extra member function, `AllocateZeroInitialized()`, which would be equivalent to `Allocate(bool initializePixels)` with `true` as argument. It seems cleared to me to explicitly write `image.AllocateZeroInitialized()`, instead of `image.Allocate(true)`.  What do you think?

dzenanz (approved):

PR #4477, opened 2024-02-23
DOC: Update documentation of `GetNameOfClass` macro calls

N-Dekker:
- Follow-up to pull request https://github.com/InsightSoftwareConsortium/ITK/pull/4373

- commit 2c264ea2ef62e916c6ef947599ce659cc8fce5fe "STYLE: Replace itkTypeMacro calls with `itkOverrideGetNameOfClassMacro`"

- commit fa1f39a2b6149ad7d91c03a699de5dc85f6619a0 "STYLE: Replace itkTypeMacroNoParent with itkVirtualGetNameOfClassMacro"

----
🎉  Ready for review now: improved documentation of 1000+ `GetNameOfClass` macro calls!


dzenanz (approved):

Modules/Core/Common/include/itkCellInterface.h
```diff
@@ -461,7 +461,7 @@ class ITK_TEMPLATE_EXPORT CellInterface
 
 #endif
 
-  /** Standard part of every itk Object. */
+  /** Defines the virtual member function `GetNameOfClass()`, which returns the name of this class as a string. */
```
N-Dekker, line 464: Looking at my own proposal again, the words "as a string" seem somewhat superfluous 🤔

dzenanz: Does it return `std::string` or `char *`? 😄

N-Dekker: https://github.com/InsightSoftwareConsortium/ITK/blob/adcc46bf9600b78bda422c49a8aca4611c1751d1/Modules/Core/Common/include/itkMacro.h#L441-L442

So we could make it instead:
```cpp
  /** Defines `virtual const char * GetNameOfClass() const`, which returns the name of this class. */
```

Do you like that better?

(I would like it to fit on one line of code anyway.)

N-Dekker: And then of course, a similar line of documentation for itkOverrideGetNameOfClassMacro calls, like:

    /** Defines `const char * GetNameOfClass() const override`, which returns the name of this class. */

Right?

Modules/Core/Common/include/itkCellInterface.h
```diff
@@ -173,7 +173,7 @@ class ITK_TEMPLATE_EXPORT CellInterface
       return smartPtr;
     }
 
-    /** Run-time type information (and related methods).   */
+    /** Defines `const char * GetNameOfClass() const override`, returning the name of this class (or class template). */
```
N-Dekker, line 176: Of course, we can also shorten this to something like:
```cpp
  /** Defines `const char * GetNameOfClass() const override`. */
```
Because the same line of comment will be there in 1000+ places, maybe it's better to keep it short. For more details, people may just check the macro definitions themselves! What do you think?

dzenanz: Agreed.

dzenanz (approved):

blowekamp (commented):
The  `/** */` comment is for doxygen. Restating the functions signature in the doxygen string is redundant and does not contribute to describing the API, IMHO.

N-Dekker:
See also:
- #4498

N-Dekker:
Now that the doxygen output is OK again, it does indeed properly show the function signature of those `GetNameOfClass()` member functions, at https://itk.org/Doxygen/html/  Nevertheless, the current HTML documentation still seems outdated to me, as it just says something like _"Standard part of all itk objects"_: 

> `const char* itk::Object::GetNameOfClass() const [override] [virtual]`
> Standard part of all itk objects.

Would you like it if it would be replaced by the text _"Returns the name of this class"_? Or would you prefer no text at all (as the name of those member functions is already pretty clear)?

blowekamp:
Getting the function signature is a great improvement! Thank you.

Adding some prose would be nice. To me the interesting things about the method is that 1) Can be used polymorpically to return the name of the instantiated class 2) it return's the non-mangle and non-namespace qualified class name. To me the old "Run-time type information" says that to me, but I have been looked at that for far too long.

N-Dekker:
> Adding some prose would be nice. To me the interesting things about the method is that 1) Can be used polymorpically to return the name of the instantiated class 2) it return's the non-mangle and non-namespace qualified class name. To me the old "Run-time type information" says that to me, but I have been looked at that for far too long.

Thanks for the input @blowekamp Minor detail: `GetNameOfClass()` often returns the name of the _class template_, rather than the name of the class. For example, it may return `"Image"`, without its template arguments, rather than `"Image<int>"`.

Anyway, I would still like to keep the documentation _very short_ for most of those 1000+member functions. Maybe we can only just have some more extensive documentation at `LightObject::GetNameOfClass()`, as LightObject is kind of the root of the ITK class hierarchy. And then just have the other GetNameOfClass() member functions _refer_ to LightObject? So what about just `\see LightObject::GetNameOfClass()`, for any GetNameOfClass() other than the one of `LightObject`?

For example:

```cpp
    /** \see LightObject::GetNameOfClass() */
    itkVirtualGetNameOfClassMacro(CellInterface);

```

blowekamp:
Yes very good idea.

Actually, if I recall correctly with Doxygen if an overloaded method has no documentation strings, then the parent's documentation is used. Last I checked that was before the override specifier was added so I don't know the behavior with that additional specifier. I would hope it would be the same.

N-Dekker:
I really hope that this pull request can get merged without merge conflicts. I found that amending a commit like this (involving more than a thousand files) is quite time consuming! Just running git to do the commit locally takes a _long_ time!

hjmjohnson (approved):

PR #4479, opened 2024-02-24
Add and use `itk::ImageBase::AllocateInitialized()`

N-Dekker:
The existing three possible ways to call `Allocate(bool = false)` (either by `Allocate(true)`, `Allocate(false)`, or `Allocate()`) may accidentally be confused, especially by novice ITK users.

The proposed `AllocateZeroInitializedPixelBuffer()` is intended to reduce the chance of such a confusion, and improve code readability. It is equivalent to `Allocate(true)`.

----
For the record, "AllocateZeroInitializedPixelBuffer" was renamed to "AllocateInitialized", as suggested at https://github.com/InsightSoftwareConsortium/ITK/pull/4479#issuecomment-1964624698

### DIFF
--- a/Modules/Core/Common/include/itkImageBase.h
+++ b/Modules/Core/Common/include/itkImageBase.h
@@ -242,6 +242,14 @@ public:
   virtual void
   Allocate(bool initialize = false);
 
+  /** Allocates the pixel buffer of the image, zero-initializing its pixels. `AllocateInitialized()` is equivalent to
+   * `Allocate(true)`. It is just intended to make the code more readable. */
+  void
+  AllocateInitialized()
+  {
+    return this->Allocate(true);
+  }
+
   /** Set the region object that defines the size and starting index
    * for the largest possible region this image could represent.  This
    * is used in determining how much memory would be needed to load an
--- a/Modules/Core/Common/include/itkFloodFilledFunctionConditionalConstIterator.hxx
+++ b/Modules/Core/Common/include/itkFloodFilledFunctionConditionalConstIterator.hxx
@@ -83,7 +83,7 @@ FloodFilledFunctionConditionalConstIterator<TImage, TFunction>::InitializeIterat
   typename TTempImage::RegionType tempRegion = this->m_Image->GetBufferedRegion();
 
   m_TemporaryPointer->SetRegions(tempRegion);
-  m_TemporaryPointer->Allocate(true); // initialize buffer to zero
+  m_TemporaryPointer->AllocateInitialized();
 
   // Initialize the queue by adding the start index assuming one of
   // the m_Seeds is "inside" This might not be true, in which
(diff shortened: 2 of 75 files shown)


Modules/Core/Common/include/itkImageBase.h
```diff
@@ -242,6 +242,14 @@ class ITK_TEMPLATE_EXPORT ImageBase : public DataObject
   virtual void
   Allocate(bool initialize = false);
 
+  /** Allocates the pixel buffer of the image, zero-initializing its pixels. `AllocateZeroInitializedPixelBuffer()` is
+   * equivalent to `Allocate(true)`. It is just intended to make the code more readable. */
+  void
+  AllocateZeroInitializedPixelBuffer()
+  {
+    return this->Allocate(true);
```
issakomi, line 250: _Allocate_ in _ImageBase_ is just an empty function, BTW.

N-Dekker: Indeed, `ImageBase::Allocate(bool)` is empty. Ideally, it should have been a pure virtual function, but apparently declaring `Allocate(bool) = 0` would break backwards compatibility:

https://github.com/InsightSoftwareConsortium/ITK/blob/2c264ea2ef62e916c6ef947599ce659cc8fce5fe/Modules/Core/Common/include/itkImageBase.h#L235-L243

Maybe we should mark the (empty) function body of `ImageBase::Allocate(bool)`  `ITK_FUTURE_LEGACY_REMOVE`. Something to consider for ITK 6... 🤔 But that's beyond the scope of this PR anyway  🤷

N-Dekker: Obviously declaring `ImageBase::Allocate` _pure virtual_ would make `ImageBase` an abstract base class. So it would disallow creating `ImageBase` objects. Probably `ImageBase::New()` should then also be removed.

But currently it is still possible to create `ImageBase` objects. Such an object would just contain the image grid/size/geometry information, _without_ the pixel data. I can imagine that that _might_ be a useful feature. Actually I'm interested to hear if people actually use that feature! 😃

N-Dekker: Thanks @blowekamp I did not yet realize that ITKv4 registration would use images that way!

I think this resolves the conversation, but the pull request is still open  😃 Please approve, if you agree about adding an `ImageBase::AllocateZeroInitializedPixelBuffer()` member function!

blowekamp:
I find the member function name over verbose.

dzenanz:
`AllocateInitialized` for the name? Docstring should have sufficient explanation about default values.

N-Dekker:
> `AllocateInitialized` for the name?

Thanks for the suggestion @dzenanz. However, I would like the name to be self-explanatory. Until now it was a common practice to add a comment to actually _explain_ the meaning of `Allocate(true)` with each `Allocate(true)` call. For example: 

    m_TemporaryPointer->Allocate(true); // initialize buffer to zero

https://github.com/InsightSoftwareConsortium/ITK/blob/adcc46bf9600b78bda422c49a8aca4611c1751d1/Modules/Core/Common/include/itkFloodFilledFunctionConditionalConstIterator.hxx#L86

So do you (both) think the name `AllocateInitialized` is sufficiently self-explanatory? So that most people won't feel the need to look up the meaning, or add a comment like "initialize buffer to zero"?

blowekamp:
> So do you (both) think the name `AllocateInitialized` is sufficiently self-explanatory? So that most people won't feel the need to look up the meaning, or add a comment like "initialize buffer to zero"?

I think the name is sufficient 👍

dzenanz (approved):

PR #4483, opened 2024-02-28
STYLE: Introduce constants for the default tolerances

blowekamp:
Consolidates constants used for default tolerances into the itkMacro.h header file.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkMacro.h
+++ b/Modules/Core/Common/include/itkMacro.h
@@ -485,6 +485,7 @@ OutputWindowDisplayGenericOutputText(const char *);
 
 extern ITKCommon_EXPORT void
 OutputWindowDisplayDebugText(const char *);
+
 } // end namespace itk
 
 // The itkDebugStatement is to be used to protect code that is only used in the itkDebugMacro
--- a/Modules/Core/Common/include/itkImageBase.h
+++ b/Modules/Core/Common/include/itkImageBase.h
@@ -44,6 +44,15 @@
 namespace itk
 {
 
+/** The default tolerance when comparing the geometry of two images.
+ *
+ * The value was chosen based on precisions of file formats such as DICOM,
+ * to enable interoperability with images saved to file formats with higher precision.
+ */
+inline constexpr double DefaultImageCoordinateTolerance = 1e-6;
+inline constexpr double DefaultImageDirectionTolerance = 1e-6;
+
+
 /** \class ImageBase
  * \brief Base class for templated image classes.
  *
@@ -791,8 +800,8 @@ public:
    */
   bool
   IsSameImageGeometryAs(const ImageBase * otherImage,
-                        double            coordinateTolerance = 1e-6,
-                        double            directionTolerance = 1e-6) const;
+                        double            coordinateTolerance = DefaultImageCoordinateTolerance,
+                        double            directionTolerance = DefaultImageDirectionTolerance) const;
 
   /** INTERNAL This method is used internally by filters to copy meta-data from
    * the output to the input. Users should not have a need to use this method.
--- a/Modules/Core/Common/src/itkImageToImageFilterCommon.cxx
+++ b/Modules/Core/Common/src/itkImageToImageFilterCommon.cxx
@@ -23,8 +23,8 @@ namespace itk
 
 namespace
 {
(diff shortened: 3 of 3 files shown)


Modules/Core/Common/include/itkMacro.h
```diff
@@ -485,6 +485,15 @@ OutputWindowDisplayGenericOutputText(const char *);
...
 OutputWindowDisplayDebugText(const char *);
+
+/** The default tolerance when comparing the geometry of two images
+ *
+ * The value was chosen based on precisions of file formats such as DICOM,
+ * to enable interoperability with images saved to file formats with higher precision.
+ */
+const double itkDefaultCoordinateTolerance = 1e-6;
+const double itkDefaultDirectionTolerance = 1e-6;
+
```
N-Dekker, line 496: Thanks @blowekamp I very much appreciate having named constants for such "magic numbers". I'm not entirely sure about having them in "itkMacro.h" though, as they obviously aren't really `#define` macro's. Maybe have them in a separate file, "itkDefaultConstants.h" ? Or maybe just place them in "itkImageBase.h"?

By the way, I would rather drop the "itk" prefix from the identifiers, as they are already nicely defined inside the `itk` namespace.

N-Dekker:
Are the constants you are proposing here meant to be part of the public API? Just wondering. Personally I think it's fine to have them in the public API. In that case, you may even call it an enhancement (`ENH`), rather than just a style improvement (`STYLE`).  😃

Modules/Core/Common/include/itkMacro.h
dzenanz: I think that Niels is right. These constants feel as they belong in ImageBase, not here.

dzenanz (approved):
Even this is fine by me.

N-Dekker:
Looks like the ITK.Windows CI failure (itkConvolutionImageFilterStreamingTest) is unrelated: https://open.cdash.org/tests/1443991348 But I don't really know  🤷

blowekamp:
Regarding the name, I struggled to find other example of global constants in ITK, so I was not sure what naming convention to use. The closest I could find what itkCommonEnum, but they are just enum and not constant real numbers.

hjmjohnson (changes requested):
This breaks one definition rule. 

https://stackoverflow.com/questions/45987571/difference-between-constexpr-and-static-constexpr-global-variable

I think you need the C++17 inline constexpr if it is used in a header.
```c++
inline constexpr double itkDefaultCoordinateTolerance = 1e-6;
```

N-Dekker:
> I think you need the C++17 inline constexpr if it is used in a header.

Interesting, thanks @hjmjohnson  😃 Honesty I don't have experience with `inline constexpr`, do you? I do believe, in practice the original `const` works fine for global variables, also in a header. As initially proposed by Bradley:

```cpp
const double itkDefaultCoordinateTolerance = 1e-6;
```

With `const`, different cxx files might see a different `itkDefaultCoordinateTolerance` instance. At a different address, `&itkDefaultCoordinateTolerance`. But they would all have the same value anyway, `1e-6`.

With `inline constexpr` I assume that all cxx files would share one and the same `itkDefaultCoordinateTolerance` instance, at the same address, `&itkDefaultCoordinateTolerance`. This might certainly be a nice property. But I just don't know if it's necessary  🤷

blowekamp:
The `inline` specifier looks like a good hint for the compiler.

There appears to be concusses to move these definitions to the itkImageBase header. I assume in the itk namespace and not as a constant member variable.

I am still looking for what the names should be. Maybe `ImageDefaultCoordinateTolerance`?

N-Dekker:
> The `inline` specifier looks like a good hint for the compiler.

I guess the `inline` specifier might be slightly more work for the compiler, because it then has to make sure that the constant has one unique instance (at a unique address). But as long as it works, it's OK to me.

> I am still looking for what the names should be. Maybe `ImageDefaultCoordinateTolerance`?

Personally I would put the word "Default" at the begin (as in "DefaultImageCoordinateTolerance"), but I guess that's just a matter of taste! Either way is OK to me.

N-Dekker (approved):
Thanks @blowekamp 👍

dzenanz (approved):

hjmjohnson (approved):

PR #4485, opened 2024-02-28
Fixed parameter preservation after identity

hjmjohnson:
When setting a transform to represent an Identity mapping, it is crucial that the stationary component of the transform is preserved.
    
```python
tfm.SetCenter( [1,2,3] );
tfm.SetIdentity();
out=tfm.GetCenter();
assert( out == [1,2,3] );
```

## PR Checklist
files (if any)

### DIFF
--- a/Modules/Core/Transform/include/itkMatrixOffsetTransformBase.hxx
+++ b/Modules/Core/Transform/include/itkMatrixOffsetTransformBase.hxx
@@ -99,7 +99,9 @@ MatrixOffsetTransformBase<TParametersValueType, VInputDimension, VOutputDimensio
   m_MatrixMTime.Modified();
   m_Offset.Fill(OutputVectorValueType{});
   m_Translation.Fill(OutputVectorValueType{});
-  m_Center.Fill(InputPointValueType{});
+  // Fixed parameters must be preserved when setting the transform to identity
+  // the center is part of the stationary fixed parameters
+  // and should not be modified by SetIdentity. m_Center.Fill(InputPointValueType{});
   m_Singular = false;
   m_InverseMatrix.SetIdentity();
   m_InverseMatrixMTime = m_MatrixMTime;
--- a/Modules/Core/Transform/test/itkEuler3DTransformGTest.cxx
+++ b/Modules/Core/Transform/test/itkEuler3DTransformGTest.cxx
@@ -33,3 +33,18 @@ TEST(Euler3DTransform, SetFixedParametersThrowsWhenSizeIsLessThanInputSpaceDimen
     ASSERT_THROW(transform->SetFixedParameters(fixedParameters), itk::ExceptionObject);
   }
 }
+
+TEST(Euler3DTransform, TestSetGetCenterAfterSetIdentity)
+{
+  using EulerTransformType = itk::Euler3DTransform<>;
+
+  // Testing preservation of the center of rotation
+  auto                        eulerTransformWithCenter = EulerTransformType::New();
+  const itk::Point<double, 3> centerOfRotation({ 200, 400, 300 });
+  eulerTransformWithCenter->SetCenter(centerOfRotation);
+  EXPECT_EQ(eulerTransformWithCenter->GetCenter(), centerOfRotation);
+  eulerTransformWithCenter->SetIdentity();
+  // The center of rotation should be preserved when the transform is set to identity.
+  // Reseting a transform to identity should not affect the FixedParameters.
+  EXPECT_EQ(eulerTransformWithCenter->GetCenter(), centerOfRotation);
+}
--- a/Modules/Core/Transform/test/itkSimilarity2DTransformTest.cxx
+++ b/Modules/Core/Transform/test/itkSimilarity2DTransformTest.cxx
@@ -156,6 +156,7 @@ itkSimilarity2DTransformTest(int, char *[])
   }
 
   // 15 degrees in radians
+  transform->SetCenter({}); // Explicitly reset center to default values
   transform->SetIdentity();
   const double angle = 15.0 * std::atan(1.0f) / 45.0;
   const double sinth = std::sin(angle);


Modules/Core/Transform/include/itkMatrixOffsetTransformBase.hxx
```diff
@@ -103,7 +105,7 @@ MatrixOffsetTransformBase<TParametersValueType, VInputDimension, VOutputDimensio
   m_Singular = false;
   m_InverseMatrix.SetIdentity();
   m_InverseMatrixMTime = m_MatrixMTime;
-  this->Modified();
+  this->SetFixedParameters(tmp_fixed_parameters);
 }
```
N-Dekker, line 109: Are you sure it is no longer necessary to call `this->Modified()` here? Just wondering 🤷

Modules/Core/Transform/test/itkEuler3DTransformTest.cxx
```diff
@@ -34,10 +34,34 @@ itkEuler3DTransformTest(int, char *[])
   using EulerTransformType = itk::Euler3DTransform<double>;
   auto eulerTransform = EulerTransformType::New();
 
+  // Testing preservation of the center of rotation
```
N-Dekker, line 37: Did you consider using GoogleTest?  It's really convenient, as it allows you to specify what you "expect" in just a few lines of code:  

    eulerTransformWithCenter->SetCenter(centerOfRotation);
    eulerTransformWithCenter->SetIdentity();
    EXPECT_EQ(eulerTransformWithCenter->GetCenter(), centerOfRotation);

Anyway, maybe something to consider for the next time  😃

Modules/Core/Transform/include/itkMatrixOffsetTransformBase.hxx
hjmjohnson: Good catch!  That was a mistake!

Modules/Core/Transform/test/itkEuler3DTransformGTest.cxx
```diff
@@ -33,3 +33,20 @@ TEST(Euler3DTransform, SetFixedParametersThrowsWhenSizeIsLessThanInputSpaceDimen
     ASSERT_THROW(transform->SetFixedParameters(fixedParameters), itk::ExceptionObject);
   }
 }
+
+TEST(Euler3DTransform, TestSetGetCenterAfterSetIdentity)
+{
+  using EulerTransformType = itk::Euler3DTransform<>;
+
+  auto eulerTransform = EulerTransformType::New();
```
N-Dekker, line 41: Just a nitpick here: Looks like `eulerTransform` is unused, and may be removed!

Modules/Core/Transform/include/itkMatrixOffsetTransformBase.hxx
```diff
@@ -104,6 +106,7 @@ MatrixOffsetTransformBase<TParametersValueType, VInputDimension, VOutputDimensio
   m_InverseMatrix.SetIdentity();
   m_InverseMatrixMTime = m_MatrixMTime;
   this->Modified();
+  this->SetFixedParameters(tmp_fixed_parameters);
```
N-Dekker, line 109: For my understanding: Instead of restoring the original fixed parameters, wouldn't it be possible to just prevent them from being modified in the first place? 

Specifically, wouldn't it be sufficient to remove the existing `m_Center.Fill(InputPointValueType{})` call from `MatrixOffsetTransformBase::SetIdentity()`?

https://github.com/InsightSoftwareConsortium/ITK/blob/1b01d5ceb899f19b2ca104249770854dd8244eab/Modules/Core/Transform/include/itkMatrixOffsetTransformBase.hxx#L102

Modules/Core/Transform/test/itkEuler3DTransformGTest.cxx
```diff
@@ -33,3 +33,20 @@ TEST(Euler3DTransform, SetFixedParametersThrowsWhenSizeIsLessThanInputSpaceDimen
     ASSERT_THROW(transform->SetFixedParameters(fixedParameters), itk::ExceptionObject);
   }
 }
+
+TEST(Euler3DTransform, TestSetGetCenterAfterSetIdentity)
```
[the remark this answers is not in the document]
hjmjohnson: No problem.

Modules/Core/Transform/test/itkEuler3DTransformGTest.cxx
hjmjohnson: Not a nitpick.  It I'm trying to squeeze this between a huge pile of grant reviews.  I copy-n-pasted when moving to GTest.

hjmjohnson: Fixed.

Modules/Core/Transform/include/itkMatrixOffsetTransformBase.hxx
hjmjohnson: @N-Dekker   I tried that at first, but "parameters" are not part of the base class, they are part of the child classes, and some child classes started failing when I tried to do that. (the Scale*Transforms did not behave as expected).

I also tried to take an approach similar to the "GetInverse" function by replacing the m_Matrix with an identity matrix, but again the "Scale*Transforms" tests started failing.

Your suggestion would work for most transforms who's stationary components are the center of rotation, but as an abstract concept a transform may be defined with other fixedparameters other than center of rotation.  I did not want to identify those cases.

This was an attempt to enforce the minimal concepts that "FixedParameters" identify the stationary component of the transform, and should not be changed by setting to identity.

N-Dekker: Sorry, I still don't fully understand why some (old) unit tests would require changing FixedParameters during the execution of `SetIdentity()`. It seems to me that your proposed fix is in fact a workaround to this behavior. Maybe I can look at it closer later this week, but otherwise I'm at least glad that most of my comments were helpful to you 😃

hjmjohnson:
@N-Dekker FixedParameters should never change after being chosen, they represent the static component of the structure of the transform.  Similar to template parameters for a function in C++.   the "Parameters" represent the current state of the Transform (i.e. Not the structure of the transform).

The SetIdentity() function should change the state of the transform, not the structure of the transform.

N-Dekker:
Thanks for your clear explanation, @hjmjohnson Unfortunately I think my knowledge and understanding of `MatrixOffsetTransformBase` derived transforms just isn't sufficient yet  🤷 

> FixedParameters should never change after being chosen

`MatrixOffsetTransformBase::SetIdentity()` does zero-fill the `m_Center`. The values of `m_Center` are also part of FixedParameters, right? When you restore the previous values of FixedParameters, but you leave `m_Center` zero-filled, FixedParameters and `m_Center` will get out of sync, right? Isn't that a problem?

BTW, I won't stop this PR, I'm just trying to understand, because it seems a bit tricky to me!

thewtex (approved):

N-Dekker:
Could you possibly still explain where in the original `SetIdentity()` implementation the FixedParameters gets modified? Because I don't see anything like `m_FixedParameters = someNewValue` here:
https://github.com/InsightSoftwareConsortium/ITK/blob/13ccff6af85522d77aa0e5e19fa929988a9de49d/Modules/Core/Transform/include/itkMatrixOffsetTransformBase.hxx#L96-L107

hjmjohnson:
@N-Dekker The problem (at least for the Euler3DTransform) is explicitly that the m_Center is being filled with all zeros.  Initially I tried removing filling the center with all zeros, but that caused a family of "Scale*Transforms" to start failing regression tests.  The issue, I think, is that the behavior of the m_Center and m_Offset parameters can be manipulated by "SetFixedParameters" of the child classes.

My solution works under the principle that transforms can be properly configured with only "SetParamters" and "SetFixedParameters" from the child classes (as is done in the TransformIO mechansims).  Under this known behavior of all transforms, I concluded that restoring the fixed parameters would set the state of the class correct for whatever weird behavior a child class may choose to do.

Well.  ...... that was my rational for this approach which a) Passes all regression tests, and b) fixes the nasty conceptual bug I ran into when using the Eurler3DTransform with SetIdentity.

N-Dekker:
Thanks again Hans. I just did a rebuild of the latest master branch, having locally _removed_ the questionable m_Center zero-filling:

https://github.com/InsightSoftwareConsortium/ITK/blob/13ccff6af85522d77aa0e5e19fa929988a9de49d/Modules/Core/Transform/include/itkMatrixOffsetTransformBase.hxx#L102

I'll try the tests in a moment...

Modules/Core/Transform/include/itkMatrixOffsetTransformBase.hxx
```diff
@@ -99,7 +99,9 @@ MatrixOffsetTransformBase<TParametersValueType, VInputDimension, VOutputDimensio
   m_MatrixMTime.Modified();
   m_Offset.Fill(OutputVectorValueType{});
   m_Translation.Fill(OutputVectorValueType{});
-  m_Center.Fill(InputPointValueType{});
+  // Fixed parameters must be preserved when setting the transform to identity
+  // the center is part of the stationary fixed parameters
+  // and should not be modified by SetIdentity. m_Center.Fill(InputPointValueType{});
```
N-Dekker, line 104: Thanks very much for giving this approach a try, Hans! I think it's the right way to go. As far as I can see, the Scale*Transform tests still pass 👍 (Right? Please check!) Only [itkSimilarity2DTransformTest](https://open.cdash.org/test/1454097934) appears to fail, but I think itkSimilarity2DTransformTest itself is wrong! I'll have another look.

hjmjohnson: @N-Dekker Thanks. I won't be able to spend time digging into this until after next week.  I appreciate any assistance I can get.

2877 - itkSimilarity2DTransformTest (Failed)

N-Dekker: Technically, just adding `transform->SetCenter({})` to line 159 of itkSimilarity2DTransformTest.cxx should make the CI happy.

Conceptually, people should no longer assume that `transform->SetIdentity()` "automagically" resets the center. It's a breaking change. But I think it's a correct one.

N-Dekker (approved):
Thanks very much, Hans! 👍

N-Dekker:
I think this PR is ready for merge, but I would like to see one more approval! Especially because there is a tiny little chance that _some_ user may experience slightly different behavior, because with this PR, `MatrixOffsetTransformBase::SetIdentity()` no longer resets the center. Again, I think the PR is entirely correct, but one more approval would be nice!

PR #4489, opened 2024-03-01
Wrap AffineTransform for float parameters

dzenanz:
## PR Checklist

### DIFF
--- a/Modules/Numerics/Optimizersv4/include/itkRegistrationParameterScalesEstimator.hxx
+++ b/Modules/Numerics/Optimizersv4/include/itkRegistrationParameterScalesEstimator.hxx
@@ -60,7 +60,7 @@ RegistrationParameterScalesEstimator<TMetric>::EstimateMaximumStepSize() -> Floa
 
   for (SizeValueType d = 0; d < dim; ++d)
   {
-    minSpacing = std::min(minSpacing, spacing[d]);
+    minSpacing = std::min<FloatType>(minSpacing, spacing[d]);
   }
 
   return minSpacing;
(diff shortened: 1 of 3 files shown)


Modules/Numerics/Optimizersv4/include/itkRegistrationParameterScalesEstimator.hxx
```diff
@@ -60,7 +60,7 @@ RegistrationParameterScalesEstimator<TMetric>::EstimateMaximumStepSize() -> Floa
 
   for (SizeValueType d = 0; d < dim; ++d)
   {
-    minSpacing = std::min(minSpacing, spacing[d]);
```
N-Dekker, line 63: Oops, my fault! So I assumed with commit 93bef51dd113f7ceae68c4a6f0708ae2dc1ba7ce that `minSpacing` and `spacing[d]` would always have the same type, but apparently I was wrong! 

Your code fix is fine to me, thanks @dzenanz  You might still shorten the commit message a little bit, if you like. The first line of the compiler error message should be sufficient. Also note that the title seems to contradict with the content of the message: [/permissive-](https://learn.microsoft.com/en-us/cpp/build/reference/permissive-standards-conformance?view=msvc-170) means _non_-permissive!!! They call it "standards conformance mode".
 
Is permissive versus non-permissive really relevant to this compile error? I would expect `std::min(minSpacing, spacing[d])` to fail in any "compilation mode", when `minSpacing` and `spacing[d]` have different types.

thewtex (approved):

hjmjohnson (approved):

Modules/Numerics/Optimizersv4/include/itkRegistrationParameterScalesEstimator.hxx
dzenanz: Hans merged the PR before I could address your feedback about commit message 🙃

N-Dekker: OK, no problem. It would have been nice if you could have addressed my feedback, but it certainly wasn't a show-stopper to me 😃

PR #4492, opened 2024-03-04
ENH: Replaced a few remaining sprintf (comments only)

seanm:
This just prevents finding them when periodically checking that no new sprintf snuck in.

### DIFF
--- a/Modules/IO/GE/src/itkGE4ImageIO.cxx
+++ b/Modules/IO/GE/src/itkGE4ImageIO.cxx
@@ -191,7 +191,7 @@ GE4ImageIO::ReadHeader(const char * FileNameToRead)
     hdr->coordinateOrientation =
       itk::SpatialOrientationEnums::ValidCoordinateOrientations::ITK_COORDINATE_ORIENTATION_RSP;
   }
-  // RGEDEBUG(std::sprintf (debugbuf, "Plane = %d\n", hdr->imagePlane); cerr <<
+  // RGEDEBUG(std::snprintf (debugbuf, sizeof(debugbuf), "Plane = %d\n", hdr->imagePlane); cerr <<
   // debugbuf;)
 
   /* Get the Scan Matrix from the IMAGE Header */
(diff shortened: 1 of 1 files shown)


dzenanz (approved):

Modules/IO/GE/src/itkGE4ImageIO.cxx
```diff
@@ -191,7 +191,7 @@ GE4ImageIO::ReadHeader(const char * FileNameToRead)
     hdr->coordinateOrientation =
       itk::SpatialOrientationEnums::ValidCoordinateOrientations::ITK_COORDINATE_ORIENTATION_RSP;
   }
-  // RGEDEBUG(std::sprintf (debugbuf, "Plane = %d\n", hdr->imagePlane); cerr <<
```
N-Dekker, line 194: Thank you @seanm, but please consider simply removing those obsolete commented-out "RGEDEBUG" statements. I don't think they will ever be used again.

seanm: Are you sure they are obsolete?  I don't use that class, and have no way to know if those debug statements are useful or not. The more conservative thing for me to do is as I've done.

N-Dekker: Thanks for your reply, @seanm This RGEDEBUG statement was commented out more than 20 years ago, with commit  cd3675f57ac94f85fba534e973f3190d87730696. So I guess it won't be "uncommented" back in the next 20 years to come.

The `RGEDEBUG` macro looks obsolete to me anyway, but that may be beyond the scope of your pull request. Your PR is specifically about commented out RGEDEBUG statements, if I understand correctly.

seanm: >So I guess it won't be "uncommented" back in the next 20 years to come.

It might never be *committed* uncommented, but for all I know people use it every day *locally*.

>Your PR is specifically about...

I occasionally search my folder of third party code for "sprintf" in my mission to get rid of them all. I was tired of seeing this result in my searches. :)

blowekamp: I second removing these statements.

Modules/IO/GE/src/itkGE4ImageIO.cxx
```diff
@@ -191,7 +191,7 @@ GE4ImageIO::ReadHeader(const char * FileNameToRead)
...
+  // RGEDEBUG(std::snprintf (debugbuf, sizeof(debugbuf), "Plane = %d\n", hdr->imagePlane); cerr <<
   // debugbuf;)
```
N-Dekker, line 195: I very much appreciate that you are chasing those unsafe `sprintf` calls. However, if this code was actually being used, I think it would be preferable to insert the values directly into the output stream: 
```cpp
    std::cerr << "Plane = " << hdr->imagePlane << '\n';
```
Right? Anyway, that's probably beyond the scope of this PR!

N-Dekker:
> Brad and Niels, feel free to follow-up with your further suggested changes (removal, improvement).

Please check:
- #4506

PR #4493, opened 2024-03-05
Combine assignments to `basei[i]` in LinearInterpolateImageFunction, remove `&` from `distance` variables

N-Dekker:
Two little `LinearInterpolateImageFunction` coding style improvements.

### DIFF
--- a/Modules/Core/ImageFunction/include/itkLinearInterpolateImageFunction.h
+++ b/Modules/Core/ImageFunction/include/itkLinearInterpolateImageFunction.h
@@ -135,7 +135,7 @@ private:
     IndexType basei;
     basei[0] = std::max(Math::Floor<IndexValueType>(index[0]), this->m_StartIndex[0]);
 
-    const InternalComputationType & distance = index[0] - static_cast<InternalComputationType>(basei[0]);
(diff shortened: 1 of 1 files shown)


Modules/Core/ImageFunction/include/itkLinearInterpolateImageFunction.h
```diff
@@ -133,10 +133,9 @@ class ITK_TEMPLATE_EXPORT LinearInterpolateImageFunction : public InterpolateIma
   EvaluateOptimized(const Dispatch<1> &, const ContinuousIndexType & index) const
   {
     IndexType basei;
-    basei[0] = Math::Floor<IndexValueType>(index[0]);
-    basei[0] = std::max(basei[0], this->m_StartIndex[0]);
+    basei[0] = std::max(Math::Floor<IndexValueType>(index[0]), this->m_StartIndex[0]);
```
N-Dekker, line 136: Of course, replacing three `basei[0]` calls with one _might_ cause a performance improvement. I didn't check, but it certainly wouldn't harm  😃

hjmjohnson (approved):

PR #4495, opened 2024-03-07
Add `Array2D(numberOfRows, numberOfCols, initialValue)` constructor

N-Dekker:
Added a `Array2D(numberOfRows, numberOfCols, initialValue)` constructor, which just calls the corresponding `vnl_matrix` constructor. Added a few use cases, which aim to simplify resetting `ScheduleType` objects.

Updated the documentation of `Array2D`.


thewtex (approved):
:+1:

PR #4498, opened 2024-03-07
DOC: Add `GetNameOfClass` macro's to DOXYGEN_PREDEFINED

N-Dekker:
Currently, the itkVirtualGetNameOfClassMacro and itkOverrideGetNameOfClassMacro calls do not get expanded properly by Doxygen. This commit aims to fix the output from Doxygen at https://itk.org/Doxygen/html/


dzenanz (approved):
Code change looks good, so the only question is: does it work?

N-Dekker:
> Code change looks good, so the only question is: does it work?

I hope so  😺 Did not test it locally, but I think it's OK. Otherwise, at least it wouldn't harm, I guess.

blowekamp (commented):
LGTM

PR #4502, opened 2024-03-11
STYLE: Remove ` == true` from Boolean expressions

N-Dekker:
Following C++ Core Guidelines, February 15, 2024, "Don’t add redundant `==` or `!=` to conditions",
http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es87-dont-add-redundant--or--to-conditions

### DIFF
--- a/Modules/Core/Common/src/itkObjectFactoryBase.cxx
+++ b/Modules/Core/Common/src/itkObjectFactoryBase.cxx
@@ -78,7 +78,7 @@ SynchronizeList(FactoryListType & output, FactoryListType & input, bool internal
     }
     if (pos == -1)
     {
-      if (internal == true)
+      if (internal)
       {
         itk::ObjectFactoryBase::RegisterFactoryInternal(factory);
       }
--- a/Modules/Core/Common/src/itkTBBMultiThreader.cxx
+++ b/Modules/Core/Common/src/itkTBBMultiThreader.cxx
@@ -182,7 +182,7 @@ struct TBBImageRegionSplitter : public itk::ImageIORegion
   {
     // The following if statement is primarily used to ensure use of
     // is_splittable_in_proportion to avoid unused variable warning.
-    if (TBBImageRegionSplitter::is_splittable_in_proportion == true)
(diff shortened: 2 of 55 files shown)


dzenanz (approved):
Did you use some tool to find instances of this?

thewtex:
Windows failure:

```
The following tests FAILED:
	779 - itkConvolutionImageFilterStreamingValidTest (Failed)
```

test flake?

N-Dekker:
> Did you use some tool to find instances of this?

Well, I did it step by step, "heuristically", starting with a global search of just `m_UseImageSpacing == true`, then using regular expressions, `(  if \(\w+) == true\)`, `(  if \([^ ]+) == true\)`, normal search `== true)`, and `== true ||`, etc. Until I felt most relevant cases were addressed.

Looking back, I could have done just a brute force find `== true` and replace with (nothing). There are only a few  `== true` cases left now. Not sure if I should also address them.

N-Dekker:
Also not addressed:

https://github.com/InsightSoftwareConsortium/ITK/blob/e3189492ce6858035b68a2b89e4a4d2f1d4041ad/Modules/Filtering/DistanceMap/test/itkSignedDanielssonDistanceMapImageFilterTest11.cxx#L124

N-Dekker:
This one is not addressed, but maybe it should, what do you think?

https://github.com/InsightSoftwareConsortium/ITK/blob/e3189492ce6858035b68a2b89e4a4d2f1d4041ad/Modules/Core/Common/test/itkMetaProgrammingLibraryTest.cxx#L30

dzenanz:
Yes, you could also update that.

N-Dekker:
FYI While I think this pull request is very much in line with C++ Core Guidelines item "Don’t add redundant `==` or `!=` to conditions", I believe it would be helpful if the guideline would explicitly mention such `b == true` cases. So I also submitted a pull request to the C++ Core Guidelines 😃 :  

- https://github.com/isocpp/CppCoreGuidelines/pull/2185

PR #4501, opened 2024-03-11
BUG: DCMTK reader wrongly rejects file with preamble

thewtex:
A previous optimization didn't take into account the preamble of a dicom file.  This adds a check.

Fixes #4108

PR #4522, opened 2024-03-17
Fix 32bits dicom read crash

lassoan:
Fix crash in ITK DICOM reader when attempting to read an unusual but valid DICOM file, which has 32 bits allocated.
Fixes the error reported at https://discourse.slicer.org/t/slicer-crashes-while-trying-to-load-dicom-files/34931/3

## PR Checklist

### DIFF
--- a/Modules/IO/DCMTK/src/itkDCMTKImageIO.cxx
+++ b/Modules/IO/DCMTK/src/itkDCMTKImageIO.cxx
@@ -28,6 +28,7 @@
 #include "dcmtk/dcmimgle/dcmimage.h"
 #include "dcmtk/dcmjpeg/djdecode.h"
 #include "dcmtk/dcmjpls/djdecode.h"
+#include "dcmtk/dcmdata/dcmetinf.h"
 #include "dcmtk/dcmdata/dcrledrg.h"
 #include "dcmtk/oflog/oflog.h"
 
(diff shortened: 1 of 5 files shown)


Modules/IO/DCMTK/src/itkDCMTKImageIO.cxx
```diff
@@ -28,25 +28,108 @@
...
+  file.seekg(0, std::ios::beg);
+  std::vector<char> buffer(DCM_PreambleLen + DCM_MagicLen);
+  file.read(&buffer[0], buffer.size());
+  if (!file)
+  {
+    return false;
+  }
+  const char * magicReadFromFile = &buffer[DCM_PreambleLen];
+  return strncmp(magicReadFromFile, DCM_Magic, DCM_MagicLen) == 0;
+}
```
N-Dekker, line 56: - Thanks @lassoan Just wondering, how does your pull request relate to #4501 ? Didn't pull request 4501 fix the problem already?
More in detail, isPreambleDicom is just a style improvement to readPreambleDicom, right?

lassoan: Yes, my first commit is just a style improvement (and a little performance improvement) over #4501. There is no functional difference.

#4501 did not fix the issue because the application first tries to read the DICOM image using GDCM and if it fails then it uses DCMTK. However, GDCM did not return with an error but it crashed the application, so the application did not have a chance to use DCMTK.

thewtex (changes requested):
Very nice @lassoan !

Could you please:

- split the GDCM change into a patch and PR against `malaterre/GDCM` for the `release` branch
- Add a test here with the file reported on the Slicer Discourse for both ITKIOGDCM and ITKIODCMTK to ensure we do not crash

?

lassoan:
> * split the GDCM change into a patch and PR against `malaterre/GDCM` for the `release` branch

malaterre/GDCM: I got a message when I committed that I would need to send something to sourceforge. The link did not work, so I hoped that I didn't need to deal with it. It seems that GDCM is on github now - https://github.com/malaterre/GDCM - do you mean I should send a pull request there?

> * Add a test here with the file reported on the Slicer Discourse for both ITKIOGDCM and ITKIODCMTK to ensure we do not crash

I've asked for permission for using a slice of the data set as test data: https://discourse.slicer.org/t/slicer-crashes-while-trying-to-load-dicom-files/34931/5

thewtex:
> https://github.com/malaterre/GDCM - do you mean I should send a pull request there?

Yes, please. We integrate patches from there into ITK if accepted.

> I've asked for permission for using a slice of the data set as test data: https://discourse.slicer.org/t/slicer-crashes-while-trying-to-load-dicom-files/34931/5

Most excellent.

Modules/IO/DCMTK/src/itkDCMTKImageIO.cxx
```diff
@@ -28,25 +28,108 @@
...
+      return false;
+    }
+  } while (groupNo == 2);
+
+#if defined(NDEBUG)
+  std::ostringstream itkmsg;
+  itkmsg << "No DICOM magic number found, but the file appears to be DICOM without a preamble.\n"
+         << "Proceeding without caution.";
+  itk::OutputWindowDisplayDebugText(itkmsg.str().c_str());
+#endif
```
N-Dekker, line 131: Is this extra debug output really necessary? If so, is it really meant to be `#if defined(NDEBUG)`?  Or maybe `#ifndef NDEBUG`? (NDEBUG = _not_ debug)

lassoan: Good catch. I agree, it should be `#ifndef NDEBUG` as everywhere else. It was probably just a typo back then when it was implemented (I just moved the code to a new location within the file).

lassoan: `Proceeding without caution` comment does not make sense either. The DICOM tag detection mechanism is pretty safe there and the test is anyway just a quick test to check if the file looks like DICOM (if the file is actually not valid DICOM then reading will fail).

N-Dekker: Thanks, glad I could be of help 😃 Honestly I always found the name of the macro `NDEBUG` quite confusing myself. I guess we'd have to ask the C standard committee for a better name  😸

dzenanz: I guess by the time they wanted to add that feature, `DEBUG` macro was quite widely used, with perhaps different meaning, and they didn't want to break all that, so they came up with `NDEBUG`, which nobody used 😄

N-Dekker: BTW The use of `std::ostringstream` appears unnecessary in this case. I would rather pass the literal string to OutputWindowDisplayDebugText directly:

```cpp
  itk::OutputWindowDisplayDebugText(
    "No DICOM magic number found, but the file appears to be DICOM without a preamble.\n"
    "Proceeding without caution.");
```

lassoan:
Submitted a PR to GDCM:

https://github.com/malaterre/GDCM/pull/168

Modules/IO/DCMTK/src/itkDCMTKImageIO.cxx
N-Dekker: @lassoan Thanks for using `itkDebugMacro` instead  👍

thewtex:
Thank you @lassoan !

LGTM.

Path to integration:

1. Merge #4521 
2. I will create another PR that integrates malaterre/GDCM#168 via our subtree merge process.
3. Rebase this PR

lassoan:
Thank you! Let me know if you need any help from me.

lassoan:
The GDCM pull request has been merged.

hjmjohnson:
@lassoan Thanks!

thewtex:
Both the GDCM patch for this branch and the GDCM patch for #4521 have been integrated into GDCM, and these are integrated int #4521 -> after #4521 has been merged, this branch can be rebased and merged.

thewtex (approved):

lassoan:
Thank you @thewtex!

dzenanz:
From the dashboard:
```log
/Users/builder/external/ITK/Modules/IO/GDCM/src/itkGDCMImageIO.cxx:191:3: error: invalid use of 'this' outside of a non-static member function
  itkDebugMacro(<< "No DICOM magic number found, but the file appears to be DICOM without a preamble.");
  ^
```
Probably use `itkWarningMacro` instead of `itkDebugMacro`?

SimonRit:
It believe it will be the same problem for [itkWarningMacro](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkMacro.h#L528) (use of this->GetNameOfClass()). I was about to submit the following patch:
```
-  itkDebugMacro(<< "No DICOM magic number found, but the file appears to be DICOM without a preamble.");
+#ifndef NDEBUG
+  itkGenericOutputMacro("No DICOM magic number found, but the file appears to be DICOM without a preamble.");
+#endif
```
which compiles, what do you think?

dzenanz:
I would prefer to remove the debug statement entirely.

SimonRit:
Sounds good to me, that should also compile!

dzenanz:
I have not started any work. I am giving other people a chance to fix this 😄

SimonRit:
Nope, I haven't done more than testing the patch I suggested. Go ahead!

seanm:
@lassoan do you have a fix in the works?

dzenanz:
@seanm There is #4544.

Issue #394, opened 2019-01-10
A different directory for GTest cxx files?


N-Dekker:
[ITK\Modules\Core\Common\test](https://github.com/InsightSoftwareConsortium/ITK/tree/v5.0b03/Modules/Core/Common/test) now has more than 200 source files. While I think it's great to have such a large number of tests, I feel it becomes a bit cumbersome to find a specific test file in this directory. It seems to me that it it would be more convenient to place the GTest cxx files of the Core module into their own "gtest" subdirectory:

    ITK\Modules\Core\Common\gtest

Currently, the ITK Core has 11 GTest cxx files. I expect (and hope) that this number will still grow very much, in years to come.

Do you guys agree that it would be useful to move the GTest cxx to such a "gtest" subdirectory?

Of course, this issue could be applicable to other ITK modules as well. But the Core module appears to have the most GTest cxx files, and it also appears to have the most tests in one (almost) flat subdirectory.

dzenanz:
As this is more of a discussion than an issue, perhaps post this question on the [forum](https://discourse.itk.org/)?

N-Dekker:
Please discuss at https://discourse.itk.org/t/a-different-directory-for-gtest-cxx-files/1514

PR #4523, opened 2024-03-18
ENH: FlatStructuringElement and ShapedNeighborhoodIterator Interop

dzenanz:
Allow interoperability of FlatStructuringElement and ShapedNeighborhoodIterator to allow compilation of the following code:

```cpp
using SE2Type = itk::FlatStructuringElement<2>;
SE2Type::RadiusType r2 = SE2Type::RadiusType::Filled(5);
SE2Type k2 = SE2Type::Ball(r2);

itk::ShapedNeighborhoodIterator<TImage> iterator(r2, image, image->GetBufferedRegion());
iterator.CreateActiveListFromNeighborhood(k2); // this now compiles

```

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkConstShapedNeighborhoodIterator.h
+++ b/Modules/Core/Common/include/itkConstShapedNeighborhoodIterator.h
@@ -374,8 +374,15 @@ public:
   /** Add non-zero neighborhood offsets to the active list. The
    * radius of the neighborhood must match the radius of the shaped
    * iterator */
+  template <typename TNeighborPixel>
   void
-  CreateActiveListFromNeighborhood(const NeighborhoodType &);
+  CreateActiveListFromNeighborhood(const Neighborhood<TNeighborPixel, Self::Dimension> &);
+  void
+  CreateActiveListFromNeighborhood(const NeighborhoodType & neighborhood)
+  {
+    // just delegate to the templated version
+    this->CreateActiveListFromNeighborhood<PixelType>(neighborhood);
+  }
 
   /** Reimplements the operator++ method so that only active pixel locations
    * are updated. */
--- a/Modules/Core/Common/test/itkShapedNeighborhoodIteratorTest.cxx
+++ b/Modules/Core/Common/test/itkShapedNeighborhoodIteratorTest.cxx
@@ -18,6 +18,7 @@
 
 #include "itkNeighborhoodIteratorTestCommon.hxx"
 #include "itkShapedNeighborhoodIterator.h"
+#include "itkFlatStructuringElement.h"
 
 int
 itkShapedNeighborhoodIteratorTest(int, char *[])
@@ -190,6 +191,18 @@ itkShapedNeighborhoodIteratorTest(int, char *[])
     std::cout << it.GetPixel(off) << std::endl;
   }
   std::cout << it.GetPixel(off) << std::endl;
+
+  println("Testing interoperability with FlatStructuringElement");
+  using StructuringElementType = itk::FlatStructuringElement<4>;
+  const StructuringElementType::RadiusType       radius2 = StructuringElementType::RadiusType::Filled(2);
+  const StructuringElementType                   kernel = StructuringElementType::Cross(radius2);
+  itk::ShapedNeighborhoodIterator<TestImageType> iterator(radius2, img, img->GetBufferedRegion());
+  iterator.CreateActiveListFromNeighborhood(kernel);
+  iterator.SetLocation(loc);
+  for (ci = iterator.Begin(); ci != iterator.End(); ++ci)
+  {
+    std::cout << ci.GetNeighborhoodIndex() << " -> " << ci.GetNeighborhoodOffset() << " = " << ci.Get() << std::endl;
+  }
 
   println("testing operator=");
   itk::ShapedNeighborhoodIterator<TestImageType> oeIt;
--- a/Modules/Core/Common/include/itkConstShapedNeighborhoodIterator.hxx
+++ b/Modules/Core/Common/include/itkConstShapedNeighborhoodIterator.hxx
@@ -113,9 +113,10 @@ ConstShapedNeighborhoodIterator<TImage, TBoundaryCondition>::DeactivateIndex(Nei
 }
 
 template <typename TImage, typename TBoundaryCondition>
+template <typename TNeighborPixel>
 void
 ConstShapedNeighborhoodIterator<TImage, TBoundaryCondition>::CreateActiveListFromNeighborhood(
-  const NeighborhoodType & neighborhood)
+  const Neighborhood<TNeighborPixel, Self::Dimension> & neighborhood)
 {
   if (this->GetRadius() != neighborhood.GetRadius())
   {
@@ -123,8 +124,8 @@ ConstShapedNeighborhoodIterator<TImage, TBoundaryCondition>::CreateActiveListFro
                                                           << ") does not equal radius of neighborhood("
                                                           << neighborhood.GetRadius() << ')');
   }
-  typename NeighborhoodType::ConstIterator nit;
-  NeighborIndexType                        idx = 0;
+  typename Neighborhood<TNeighborPixel, Self::Dimension>::ConstIterator nit;
+  NeighborIndexType                                                     idx = 0;
   for (nit = neighborhood.Begin(); nit != neighborhood.End(); ++nit, ++idx)
   {
     if (*nit)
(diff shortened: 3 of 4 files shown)


thewtex (approved):

Modules/Core/Common/include/itkConstShapedNeighborhoodIterator.h
```diff
@@ -374,8 +374,15 @@ class ITK_TEMPLATE_EXPORT ConstShapedNeighborhoodIterator : private Neighborhood
...
+  template <typename NeighborPixelType>
   void
-  CreateActiveListFromNeighborhood(const NeighborhoodType &);
+  CreateActiveListFromNeighborhood(const Neighborhood<NeighborPixelType, Self::Dimension> &);
+  void
+  CreateActiveListFromNeighborhood(const NeighborhoodType & neighborhood)
+  {
+    // just delegate to the templated version
+    this->CreateActiveListFromNeighborhood<>(neighborhood);
+  }
```
N-Dekker, line 385: Thanks @dzenanz. Would it be possible to just remove the old `CreateActiveListFromNeighborhood(const NeighborhoodType &)` now, as it seems totally redundant when we get your more generic version?

Modules/Core/Common/include/itkConstShapedNeighborhoodIterator.h
```diff
@@ -374,8 +374,15 @@ class ITK_TEMPLATE_EXPORT ConstShapedNeighborhoodIterator : private Neighborhood
   /** Add non-zero neighborhood offsets to the active list. The
    * radius of the neighborhood must match the radius of the shaped
    * iterator */
+  template <typename NeighborPixelType>
```
N-Dekker, line 377: Nitpick: ITK's naming convention has always been to start of typename template argument names by `T`. So I would suggest something like `TNeighborPixel`.  (Unless you would like to abandon this old naming convention, of course.)

Modules/Core/Common/test/itkShapedNeighborhoodIteratorTest.cxx
```diff
@@ -191,6 +192,21 @@ itkShapedNeighborhoodIteratorTest(int, char *[])
   }
   std::cout << it.GetPixel(off) << std::endl;
 
+  println("Testing interoperability with FlatStructuringElement");
```
N-Dekker, line 195: Did you also consider making it a GoogleTest unit test? Or is there a reason why GoogleTest would be less appropriate here? You know I'm a fan of GoogleTest. 😃 I think it was a great move forward for ITK 5, to start using GTest.

Modules/Core/Common/test/itkShapedNeighborhoodIteratorTest.cxx
```diff
@@ -191,6 +192,21 @@ itkShapedNeighborhoodIteratorTest(int, char *[])
...
+  using SE2Type = itk::FlatStructuringElement<4>;
```
N-Dekker, line 196: What does "SE2Type" mean? Maybe rename it to `StructuringElementType`...?

Please also consider more meaningful variable names instead of those 1-letter identifiers. For example, `radius` instead of `r`, and `structuringElement` instead of `k`.

Modules/Core/Common/test/itkShapedNeighborhoodIteratorTest.cxx
```diff
@@ -191,6 +192,21 @@ itkShapedNeighborhoodIteratorTest(int, char *[])
...
+  SE2Type::RadiusType r;
+  r.Fill(2);
+  SE2Type k;
+  k = SE2Type::Cross(r);
```
N-Dekker, line 200: Please consider initialization, instead of declaration + assignment, for those two local variables:

```cpp
    const SE2Type::RadiusType r = SE2Type::RadiusType::Filled(2);
    const SE2Type k = SE2Type::Cross(r);
```
You may then even declare them `const`. 😺

Modules/Core/Common/test/itkShapedNeighborhoodIteratorTest.cxx
```diff
@@ -191,6 +192,21 @@ itkShapedNeighborhoodIteratorTest(int, char *[])
...
+  itk::ShapedNeighborhoodIterator<TestImageType> iterator(r, img, img->GetBufferedRegion());
+  iterator.CreateActiveListFromNeighborhood(k); // this now compiles
```
N-Dekker, line 202: The comment "this now compiles" is meaningful now, in the context of this pull request, but it may no longer be understood in a few years. Personally I would just drop it. But you could also elaborate a bit on why this compiles (now).

Modules/Core/Common/include/itkConstShapedNeighborhoodIterator.h
dzenanz: I thought about it too, but I am not sure that all the calls that worked before would continue working. The template variant relies on template argument being properly deduced, and that sometimes does not work.

Modules/Core/Common/test/itkShapedNeighborhoodIteratorTest.cxx
dzenanz: Less effort was required to extend this test, than to translate this into a GTest.

Modules/Core/Common/test/itkShapedNeighborhoodIteratorTest.cxx
dzenanz: I was concerned with things working, I totally forgot about style 😄

Modules/Core/Common/test/itkShapedNeighborhoodIteratorTest.cxx
N-Dekker: OK, but then, how can we make it easier for people to use GTest?  🤷 

I think GTest unit tests are especially useful, because they help us to describe and validate the requirements, assertions, and expectations of our components. A GTest unit test is usually also much shorter than a traditional ITK test function, making it easier to review and understand.

N-Dekker: We could simply add dummy GTest.cxx files for all filters, if that helps. 🤷

dzenanz: A big part of the reason why extending existing tests is easier is not having to think too much about syntax, as it can be easily gleaned from the existing code. And existence of some type and variable instances makes additions easier. Dummy GTests might help a bit with getting started, but might be confusing at other times, especially to newcomers, and would definitely increase clutter. So adding them would be a mixed bag.

Modules/Core/Common/include/itkConstShapedNeighborhoodIterator.h
N-Dekker: I tried it locally, and it appeared to work just fine, after removing the old `CreateActiveListFromNeighborhood(const NeighborhoodType &)` overload.

dzenanz: There is not a lot of code inside ITK which exercises this. I would rather be cautious.

Modules/Core/Common/test/itkShapedNeighborhoodIteratorTest.cxx
N-Dekker: - For me personally, it would help to have separate directory for GTests, next to each "test"directory, e.g., "Modules\Core\Common\gtest", next to "Modules\Core\Common\test". I asked about this before (#394), but was a very long time ago. (More than 5 years ago, actually.) Would it be worth reconsidering?

dzenanz: There is going to be non-trivial amount of changes required. If you are willing to do them, I am fine with that addition.

N-Dekker: Thanks @dzenanz I meant just moving our existing GTest.cxx files to separate "gtest" directories (so that's not really an addition, more like a restyling). Would that be OK to you now? Because five years ago, you didn't really like the idea: https://discourse.itk.org/t/a-different-directory-for-gtest-cxx-files/1514/2 I hope you changed your mind, in the mean time 😺

dzenanz: I don't mind it now 😄

N-Dekker (approved):
Thanks @dzenanz Hope the old `CreateActiveListFromNeighborhood(const NeighborhoodType &)` overload will become `ITK_FUTURE_LEGACY_REMOVE` in the near future  😸

PR #4528, opened 2024-03-19
ENH: Add GoogleTest unit tests for GradientImageFilter

N-Dekker:
Tests the output for a uniform input image and for an input image that has a constant gradient.


thewtex (approved):
🥇

PR #4529, opened 2024-03-19
Add inverse capability to KernelTransform

dzenanz:
## PR Checklist

### DIFF
--- a/Modules/Core/Transform/include/itkKernelTransform.hxx
+++ b/Modules/Core/Transform/include/itkKernelTransform.hxx
@@ -509,6 +509,43 @@ KernelTransform<TParametersValueType, VDimension>::GetFixedParameters() const ->
 
 
 template <typename TParametersValueType, unsigned int VDimension>
+bool
+KernelTransform<TParametersValueType, VDimension>::GetInverse(Self * inverseTransform) const
+{
+  if (!inverseTransform)
+  {
+    return false;
+  }
+
+  auto deepCopyLandmarks = [](const typename PointSetType::ConstPointer source,
+                              typename PointSetType::Pointer            destination) {
+    typename PointSetType::PointsContainer::ConstIterator sourceIt = source->GetPoints()->Begin();
+
+    typename PointSetType::PointIdentifier i{ 0 };
+    while (sourceIt != source->GetPoints()->End())
+    {
+      destination->SetPoint(i, sourceIt->Value());
+      ++i;
+      ++sourceIt;
+    }
+  };
+
+  // make a deep copy of the source and target landmarks
+  typename PointSetType::Pointer sourceLandmarks = PointSetType::New();
+  typename PointSetType::Pointer targetLandmarks = PointSetType::New();
+  deepCopyLandmarks(this->GetSourceLandmarks(), sourceLandmarks);
+  deepCopyLandmarks(this->GetTargetLandmarks(), targetLandmarks);
+
+  // inversion comes from reversing the landmarks
+  inverseTransform->SetSourceLandmarks(targetLandmarks);
+  inverseTransform->SetTargetLandmarks(sourceLandmarks);
+  inverseTransform->SetStiffness(this->GetStiffness());
+  inverseTransform->ComputeWMatrix();
+
+  return true;
+}
+
+template <typename TParametersValueType, unsigned int VDimension>
 void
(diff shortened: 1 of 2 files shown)


thewtex (changes requested):
Please add a test.

Modules/Core/Transform/include/itkKernelTransform.hxx
```diff
@@ -508,6 +508,43 @@ KernelTransform<TParametersValueType, VDimension>::GetFixedParameters() const ->
...
+    {
+      destination->SetPoint(i, sourceIt->Value());
+      ++i;
+      ++sourceIt;
+    }
+  };
+
+  // make a deep copy of the source and target landmarks
+  typename PointSetType::Pointer sourceLandmarks = PointSetType::New();
+  typename PointSetType::Pointer targetLandmarks = PointSetType::New();
```
N-Dekker, line 535: Please consider using `auto` here. For the return type of `Type::New()` we have agreed that `auto` is appropriate: https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/blob/d1562dd6dbc929ba18fdf8ce8aa86e8337597d7e/SoftwareGuide/Latex/Appendices/CodingStyleGuide.tex#L1558-L1562

Modules/Core/Transform/include/itkKernelTransform.hxx
```diff
@@ -508,6 +508,43 @@ KernelTransform<TParametersValueType, VDimension>::GetFixedParameters() const ->
...
+  deepCopyLandmarks(this->GetSourceLandmarks(), sourceLandmarks);
+  deepCopyLandmarks(this->GetTargetLandmarks(), targetLandmarks);
```
N-Dekker, line 537: Please consider directly copying the internal STL containers of the landmarks:
```cpp

  sourceLandmarks->GetPoints()->CastToSTLContainer() =
    m_SourceLandmarks->GetPoints()->CastToSTLConstContainer();
  targetLandmarks->GetPoints()->CastToSTLContainer() =
    m_TargetLandmarks->GetPoints()->CastToSTLConstContainer();

```

Modules/Core/Transform/include/itkKernelTransform.hxx
N-Dekker: `const auto` would be even nicer  😇 

```cpp
    const auto sourceLandmarks = PointSetType::New();
    const auto targetLandmarks = PointSetType::New();
```

dzenanz:
Just as I was making a pull request, I realized that a test is needed. I don't really have time to make a meaningful test (aside from the use in my own program). I also see that https://github.com/InsightSoftwareConsortium/ITK/blob/17a221c0c17dad59755e08c3df618125247334aa/Modules/Core/Transform/test/itkTestTransformGetInverse.cxx#L119 is crashing, probably meaning the inversion method I added is not thread-safe. It would be good if someone took over this PR (I am eying you Niels 😄). Otherwise it is likely going to fester for a long time.

Modules/Core/Transform/include/itkKernelTransform.hxx
```diff
@@ -508,6 +508,43 @@ KernelTransform<TParametersValueType, VDimension>::GetFixedParameters() const ->
...
+
+  // inversion comes from reversing the landmarks
+  inverseTransform->SetSourceLandmarks(targetLandmarks);
+  inverseTransform->SetTargetLandmarks(sourceLandmarks);
+  inverseTransform->SetStiffness(this->GetStiffness());
```
N-Dekker, line 542: I think it's OK for a class to directly access its own data members, without calling Get and Set. So more like:

```cpp
  inverseTransform->m_SourceLandmarks = targetLandmarks;
  inverseTransform->m_TargetLandmarks = sourceLandmarks;
  inverseTransform->m_Stiffness = m_Stiffness;
```

Right?

N-Dekker:
> It would be good if someone took over this PR

I can ask next week if it may be of interest to elastix. But otherwise this PR does not have a high priority to me. Does it have priority to you? If so, then I think you should try to make it ready, including a simple little test 🤷  Otherwise, just postpone until the summer  🌞 holidays  😸

dzenanz:
It is not of high priority to me either. So it would be good if you added it to your TODO list for summer holidays 😄

dzenanz:
Superseded by #6678. Thank you for thoughtful suggestions Niels.

PR #4533, opened 2024-03-26
WIP (Not ready): Add `GradientImageFilter::OverrideBoundaryCondition(std::unique_ptr<BoundaryConditionType>)`

N-Dekker:
Deprecated the old `GradientImageFilter` member function `OverrideBoundaryCondition(BoundaryConditionType *)` in favor of `OverrideBoundaryCondition(std::unique_ptr<BoundaryConditionType>)`.

Aims to make it more clear that the filter takes the ownership of the specified `BoundaryCondition` object.

Triggered by a comment from Bradley Lowekamp (@blowekamp) at https://github.com/InsightSoftwareConsortium/ITK/pull/4512#discussion_r1523141237

### DIFF
--- a/Modules/Filtering/ImageGradient/include/itkGradientImageFilter.h
+++ b/Modules/Filtering/ImageGradient/include/itkGradientImageFilter.h
@@ -147,7 +147,7 @@ public:
 
   /** Allows to change the default boundary condition */
   void
-  OverrideBoundaryCondition(std::unique_ptr<BoundaryConditionType> boundaryCondition);
+  OverrideBoundaryCondition(std::unique_ptr<BoundaryConditionType> && boundaryCondition);
 
 #ifdef ITK_USE_CONCEPT_CHECKING
   // Begin concept checking
--- a/Modules/Filtering/ImageGradient/include/itkGradientImageFilter.h
+++ b/Modules/Filtering/ImageGradient/include/itkGradientImageFilter.h
@@ -100,6 +100,9 @@ public:
   using CovariantVectorType = CovariantVector<OutputValueType, Self::OutputImageDimension>;
   using OutputImageRegionType = typename OutputImageType::RegionType;
 
+  /** Helper typedef for the parameter type of OverrideBoundaryCondition. */
+  using BoundaryConditionType = ImageBoundaryCondition<TInputImage>;
+
   /** GradientImageFilter needs a larger input requested region than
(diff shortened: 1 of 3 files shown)


N-Dekker:
For the record, I started a discussion on how unique_ptr parameters are supported by SWIG, at:  
- https://github.com/swig/swig/discussions/2854

hjmjohnson:
@N-Dekker — I've opened #6714, which carries this idea forward. Thank you for it, and apologies it sat blocked this long: the diagnosis here was correct, and the design you proposed is the one that landed.

**The SWIG blocker is gone.** ITK now pins SWIG 4.3.0, which ships `std_unique_ptr.i`. I verified the whole chain rather than assuming it: castxml emits the method, pygccxml resolves the argument as `::std::unique_ptr<...>` (not `?unknown?`, so `igenerator.py`'s `skip_method` no longer drops it), the `%unique_ptr` typemap binds, and the generated C++ compiles, links, and passes a runtime ownership test — the exact step that failed in 2024.

One trap worth recording for anyone doing this elsewhere: `%unique_ptr` must be given ITK's *mangled* alias (`itkImageBoundaryConditionIF2`), not the C++ spelling. The C++ spelling fails **silently** — SWIG falls back to the default typemap with no error or warning.

<details>
<summary>What #6714 changes relative to this PR</summary>

- Same core idea, but the new method is named `SetBoundaryCondition` and the existing `OverrideBoundaryCondition(T *)` is deprecated under `ITK_FUTURE_LEGACY_REMOVE` rather than changed in place, so existing callers keep both compiling and their current ownership semantics.
- Widened to the whole family. `GradientImageFilter` was the *only* one of six classes whose identically-signed `OverrideBoundaryCondition` took ownership; the other five now say explicitly in their docs that they do not. The rule: `OverrideBoundaryCondition(T *)` never takes ownership, and any owning API uses a different name **and** a `unique_ptr` parameter.
- Adds `GetBoundaryCondition` / `ResetBoundaryCondition`, GTests that observe ownership via a destruction counter, a Python test for the wrapped behavior, and a migration-guide entry.

</details>

I'd suggest closing this PR in favor of #6714, but that's your call — I haven't automated it, and there's no `Closes` keyword in the new commit.

hjmjohnson:
@N-Dekker Do you have any concerns with #6714 updates to this PR?

hjmjohnson:
Superseded by #6714

PR #4534, opened 2024-03-26
Add helper function, `Transform::InvertTransform`, simplifying `GetInverseTransform()`

N-Dekker:
- Added a static, protected helper function, `Transform::InvertTransform`
- Let derived `GetInverseTransform()` member functions just call `Transform::InvertTransform`

Aims to simplify the code, and reduce code duplication.

### DIFF
--- a/Modules/Core/Transform/include/itkCompositeTransform.hxx
+++ b/Modules/Core/Transform/include/itkCompositeTransform.hxx
@@ -446,17 +446,7 @@ template <typename TParametersValueType, unsigned int VDimension>
 auto
 CompositeTransform<TParametersValueType, VDimension>::GetInverseTransform() const -> InverseTransformBasePointer
 {
-  /* This method can't be defined in Superclass because of the call to New() */
-  Pointer inverseTransform = New();
-
-  if (this->GetInverse(inverseTransform))
-  {
-    return inverseTransform.GetPointer();
-  }
-  else
-  {
-    return nullptr;
-  }
+  return Superclass::InvertTransform(*this);
 }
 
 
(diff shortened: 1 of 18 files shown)


dzenanz (approved):

Modules/Core/Transform/include/itkCompositeTransform.hxx
```diff
@@ -446,17 +446,7 @@ template <typename TParametersValueType, unsigned int VDimension>
...
-
-  if (this->GetInverse(inverseTransform))
-  {
-    return inverseTransform.GetPointer();
-  }
-  else
-  {
-    return nullptr;
-  }
+  return Superclass::InvertTransform(*this);
```
N-Dekker, line 449: Looks like the original intention was already to implement `GetInverseTransform()` in its `Superclass`, as I understand from this comment, added by Michael Stauffer, commit d9de55c13c0632cc13fbb0f60f2b8f08eae1de0c (2012), saying `/* This method can't be defined in Superclass because of the call to New() */`

Modules/Core/Transform/include/itkv3Rigid3DTransform.h
```diff
@@ -124,8 +124,7 @@ class ITK_TEMPLATE_EXPORT Rigid3DTransform : public itk::Rigid3DTransform<TParam
   InverseTransformBasePointer
   GetInverseTransform() const override
   {
-    Pointer inv = New();
-    return this->GetInverse(inv) ? inv.GetPointer() : nullptr;
+    return Superclass::InvertTransform(*this);
   }
```
N-Dekker, line 128: Interestingly, this one is implemented inline, in the "h" file, whereas the other `GetInverseTransform()` member functions are implemented outside of their class definition, in their "hxx" file. Personally it's fine by me to have it inline, in the "h" file.

PR #4538, opened 2024-03-27
 `FUTURE_LEGACY_REMOVE` BoundaryConditionPointerType, replace with `BoundaryConditionType *`

N-Dekker:
`BoundaryConditionPointerType` is just a type alias for  `BoundaryConditionType *`. It appears much clearer to directly use `BoundaryConditionType *`, to explicitly make clear when using a _raw_ pointer, rather than a _smart_ pointer.


dzenanz (approved):

thewtex (approved):
:+1:

PR #4539, opened 2024-03-27
HDF5ImageIO: Remove `this->` when accessing its own data, use unique_ptr

N-Dekker:
Following C++ Core Guidelines, ["Use unique_ptr or shared_ptr to avoid forgetting to delete objects created using new"](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-smart)


dzenanz (approved):
LGTM, assuming you fix the compile error 😄 

Edit: you beat me by less than a minute.

thewtex (approved):

Issue #4536, opened 2024-03-26
Upgrade SWIG from version 4.0.2 to version 4.1.0 or higher, please!


N-Dekker:
- Pull request https://github.com/InsightSoftwareConsortium/ITK/pull/4533 currently fails on both [ITK.Linux.Python](https://open.cdash.org/viewBuildError.php?buildid=9492071) and [ITK.macOS.Python](https://open.cdash.org/viewBuildError.php?buildid=9492209), and it seems to be caused by the lack of `std::unique_ptr` support by SWIG 4.0.2. Which is the current SWIG version on the ITK master branch:

https://github.com/InsightSoftwareConsortium/ITK/blob/ac1ebe6dcf02f022f02cc8748fd14f9b9dc9266c/Wrapping/Generators/SwigInterface/CMakeLists.txt#L17-L18

`std::unique_ptr` is an essential utility to prevent memory leaks, as well as multiple `delete`'s on one pointer (which would cause undefined behavior, possibly leading to crashes). It is also useful as function parameter type, to indicate that the function takes ownership of the object that is passed as argument.

SWIG 4.1.0 (released in 2022) appears to be the first version to support `std::unique_ptr`: https://sourceforge.net/p/swig/news/2022/10/swig-410-released/

The most recent release is version 4.2.1. That version appears to have even more modern C++ support: https://swig.org

Can you please upgrade SWIG? At least to version 4.1.0, but the latest version appears even better  😃

----

For the record, SimpleITK is currently at [SWIG version 4.1.1](https://sourceforge.net/p/swig/news/2022/11/-swig-411-released/): https://github.com/SimpleITK/SimpleITK/blob/8e29d4902cdbc91b03578fa654b06d868b08f03c/SuperBuild/External_Swig.cmake#L22

Issue #3955, opened 2023-03-08
Create SWIG binaries for macOS


thewtex:
Similar to the Windows binaries and Linux x86_64 binaries, create SWIG binaries for macOS x86_64, ARM for the Python builds for speed.

They should be added to ITK/Wrapping/Generators/SwigInterface/CMakeLists.txt.

They should use statically linked PCRE.

They should be built with OSX SDK compatibility of 11.0.

Issue #3364, opened 2022-04-06
PyEval_InitThreads is deprecated in Python 3.9


dzenanz:
This occurs many times when compiling Python wrappings. Example message:
```text
122>vnl_vector_refPython.cpp
122>C:\Dev\ITK-py\Wrapping\Modules\ITKCommon\vnl_vector_refPython.cpp(7653,3): warning C4996: 'PyEval_InitThreads': deprecated in 3.9
```
Tested on `master` branch, but probably affects all versions.

dzenanz:
It looks like we only need to use a newer version of SWIG: https://github.com/swig/swig/issues/1996.

dzenanz:
SWIG changed the build system to CMake. https://github.com/InsightSoftwareConsortium/ITK/blob/master/Wrapping/Generators/SwigInterface/CMakeLists.txt will need to be rewritten. I expect serious simplification and probable unification of Win and *nix versions.

dzenanz:
We could maybe bundle it as a third-party library and update it more frequently using the existing infrastructure?

PR #4541, opened 2024-03-27
ENH: Bump SWIG to 2024-03-26-master

thewtex:
- Supports more modern C++
- More Pythonic interfaces, e.g. Python Iterator Protocol
- Python Limited ABI

There were duplicate definition errors due to double inclusion of

  %template(vectoritkImageXX, std::vector< itkImageXX >)

via itkImage_ext.i in PCAShapeSignedDistanceFunction. Remove
itkImage_ext.i from the latter but still provide the required typemaps
by adding DECL_PYTHON_STD_VEC_RAW_TO_SMARTPTR_TYPEMAP in
itkPCAShapeSignedDistanceFunction.wrap.

Remove -py3 because it is no longer supported (the default).

Remove "ios_base*" config templates because they are no longer emitted.

Bump PCRE. SWIG and PCRE are now built via CMake instead of autotools.

Closes https://github.com/InsightSoftwareConsortium/ITK/issues/4536
Closes https://github.com/InsightSoftwareConsortium/ITK/issues/3364
Closes https://github.com/InsightSoftwareConsortium/ITK/issues/3955

### DIFF
--- a/Modules/Segmentation/SignedDistanceFunction/wrapping/itkPCAShapeSignedDistanceFunction.wrap
+++ b/Modules/Segmentation/SignedDistanceFunction/wrapping/itkPCAShapeSignedDistanceFunction.wrap
@@ -8,6 +8,7 @@ foreach(d ${ITK_WRAP_IMAGE_DIMS})
   # `double` because of OptimizerParameters is hardcoded as templated
   # over `double` in parent class `ShapeSignedDistanceFunction`.
   foreach(t ${types})
+    string(APPEND ITK_WRAP_PYTHON_SWIG_EXT "DECL_PYTHON_STD_VEC_RAW_TO_SMARTPTR_TYPEMAP(itkImage${ITKM_${t}}${d}, itkImage${ITKM_${t}}${d}_Pointer)\n")
     itk_wrap_template("${ITKM_D}${d}${ITKM_I${t}${d}}" "${ITKT_D},${d},${ITKT_I${t}${d}}")
   endforeach()
 endforeach()
--- a/Modules/Bridge/NumPy/include/itkPyBuffer.hxx
+++ b/Modules/Bridge/NumPy/include/itkPyBuffer.hxx
@@ -130,7 +130,7 @@ PyBuffer<TImage>::_GetImageViewFromArray(PyObject * arr, PyObject * shape, PyObj
   {
     PyErr_SetString(PyExc_RuntimeError, "Size mismatch of image and Buffer.");
     PyBuffer_Release(&pyBuffer);
-    Py_DECREF(shapeseq);
+    SWIG_Py_DECREF(shapeseq);
     return nullptr;
   }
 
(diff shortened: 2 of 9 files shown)


thewtex:
@dzenanz note how to support `std::vector< ImageType >` on class APIs.

thewtex:
Closes #4536

N-Dekker:
Wow, that's a big step forward! Thanks Matt!

Still wondering, is there a specific reason to take the revision of 2024-03-26-master, rather than the last official release, https://github.com/swig/swig/releases/tag/v4.2.1  ?

thewtex:
> Still wondering, is there a specific reason to take the revision of 2024-03-26-master, rather than the last official release, https://github.com/swig/swig/releases/tag/v4.2.1 ?

Yes, we want this patch: https://github.com/swig/swig/commit/21aa6e8f9e4c7d1b0249c46140b7ce5397d7dc52

dzenanz:
Does this touch #3955?

Modules/Segmentation/SignedDistanceFunction/wrapping/itkPCAShapeSignedDistanceFunction.wrap
```diff
@@ -8,6 +8,7 @@ foreach(d ${ITK_WRAP_IMAGE_DIMS})
   # `double` because of OptimizerParameters is hardcoded as templated
   # over `double` in parent class `ShapeSignedDistanceFunction`.
   foreach(t ${types})
+    string(APPEND ITK_WRAP_PYTHON_SWIG_EXT "DECL_PYTHON_STD_VEC_RAW_TO_SMARTPTR_TYPEMAP(itkImage${ITKM_${t}}${d}, itkImage${ITKM_${t}}${d}_Pointer)\n")
```
dzenanz, line 11: Is this expected to work in remote modules? Should I replace [the old way](https://github.com/dzenanz/ANTsWasm/commit/f7e72cf5bb2a11837cb18aa13166cd2a56652221) with this style?

dzenanz:
MacOS Python build error:
```log
CMake Error at /usr/local/Cellar/cmake/3.28.1/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find BISON: Found unsuitable version "2.3", but required is at
  least "3.5" (found /usr/bin/bison)
  /usr/local/Cellar/cmake/3.28.1/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:598 (_FPHSA_FAILURE_MESSAGE)
  /usr/local/Cellar/cmake/3.28.1/share/cmake/Modules/FindBISON.cmake:306 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:91 (find_package)
```

dzenanz (approved):
Code looks good. Maybe we need a newer MacOS virtual machine image?

dzenanz:
I assume we need to bump the version here:
https://github.com/InsightSoftwareConsortium/ITK/blob/fad58494684c15c71525e127ba5fc6318dd33831/Testing/ContinuousIntegration/AzurePipelinesMacOS.yml#L34

thewtex:
> Does this touch https://github.com/InsightSoftwareConsortium/ITK/issues/3955?

I will add macOS binaries.

Modules/Segmentation/SignedDistanceFunction/wrapping/itkPCAShapeSignedDistanceFunction.wrap
thewtex: Yes

thewtex:
> I assume we need to bump the version here:

No, we want to keep testing against the older version, and bumping would not help with the build.

dzenanz (approved):

PR #4548, opened 2024-03-28
STYLE: Do not declare CumulativeGaussianCostFunction data as pointer

N-Dekker:
It appears unnecessary to manually manage the memory pointed to by its members, m_OriginalDataArray and m_MeasurePointer.

Defaulted its default-constructor and its destructor.


dzenanz (approved):
Looks good on a glance.

thewtex (approved):

PR #4558, opened 2024-04-03
Fixes and improvements for the Python Stable ABI for Windows

thewtex:


Wrapping/macro_files/itk_end_wrap_module.cmake
```diff
@@ -396,6 +396,11 @@ ${DO_NOT_WAIT_FOR_THREADS_CALLS}
 
   # build all the c++ files from this module in a common lib
   if(NOT TARGET ${lib})
+    set(development_component "Development.Module")
+    if(ITK_USE_PYTHON_LIMITED_API)
+      set(development_component "Development.SABIModule")
```
phcerdan, line 401: Is this override correct or you were thinking on appending? `set(development_component "${development_component} Development.SABIModule")`?

Just saying because in `CMake/ITKSetPython3Vars.cmake` both are added:
`find_package(Python3 ${PYTHON_REQUIRED_VERSION} COMPONENTS Interpreter Development.Module ${SKBUILD_SABI_COMPONENT})`

phcerdan (approved):
Looking good! One question/check, just in case.

Wrapping/macro_files/itk_end_wrap_module.cmake
thewtex: Yes, good eye. It is intended. Further down, we only link to either the `Development.Module` or the `Development.SABIModule`. We do not want to link to both so we use the constraint here. We also use `REQUIRED` here to throw an error when we do not have what we know we need.

In "ITKSetPython3Vars.cmake", the intention is to find all the components that might be needed. It also does not use `REQUIRED` because this will cause false-positive issues in some setups.

PR #4560, opened 2024-04-04
STYLE: Put local matrix variables in FEM on the stack, remove new/delete

N-Dekker:
Note that the type of these local matrices is `MatrixType = vnl_matrix<Float>`, so such a matrix is just empty, when default-constructed.

Following C++ Core Guidelines, February 15, 2024, ["Prefer scoped objects, don’t heap-allocate unnecessarily"](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r5-prefer-scoped-objects-dont-heap-allocate-unnecessarily)


dzenanz (approved):
Looks good on a glance.

N-Dekker:
> Looks good on a glance.

Thanks @dzenanz I think this PR might even improve the performance of the FEM module, but I did not have time to do bench-marking. It's a style improvement anyway 😇

thewtex (approved):

PR #4561, opened 2024-04-04
Use unique_ptr for the "pimpl pointers" of GiplImageIO and MINCImageIO

N-Dekker:
Both "pimpl-like" pointers were already `private` anyway.

- Following C++ Core Guidelines, ["Use unique_ptr or shared_ptr to avoid forgetting to delete objects created using new"](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-smart)
- Follow-up to pull request https://github.com/InsightSoftwareConsortium/ITK/pull/4539

@thewtex This PR obviously does not affect the public API. It just aims to make the code more secure, by removing two more cases of manual memory management. Would it still be in time for ITK 5.4.0?

### DIFF
--- a/Modules/IO/GIPL/include/itkGiplImageIO.h
+++ b/Modules/IO/GIPL/include/itkGiplImageIO.h
@@ -23,6 +23,7 @@
 #include <fstream>
 #include "itkImageIOBase.h"
 #include <cstdio>
+#include <memory> // For unique_ptr.
 
 namespace itk
 {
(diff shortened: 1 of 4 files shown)


dzenanz (approved):
Looks good on a glance.

thewtex (approved):
Thanks, @N-Dekker 

Yes, this should be OK for the release since it is not touching the public API and *should* be fine. Please keep an eye on the Valgrind Nightly dashboard build.

hjmjohnson (approved):

PR #4562, opened 2024-04-05
WIP: Use std filesystem path

blowekamp:
Add std::filesystem::path to ImageIOBase for eventual wchar support for paths.

This proposes an approach to remove direct access to m_FileName, with the eventual end goal to use GetFilePath and "legacy" GetFileName methods.
    
    Add Set/Get FilePath member functions to ImageIOBase, with the future
    plan to replace the m_FileName member variable with these access
    member functions. The m_FileName is directly access by derived ImageIO
    classes.
    
    The m_FileName is now "constant" to disallow derived classes from
    directly writing to the variable.  The SetFilePath method updates the
    legacy const m_FileName to maintain compatibility.

### DIFF
--- a/Modules/Core/Common/include/itkPathType.h
+++ b/Modules/Core/Common/include/itkPathType.h
@@ -1,0 +1,29 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         https://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+#ifndef itkPathType_h
+#define itkPathType_h
+
+#include <filesystem>
+
+namespace itk
+{
+using PathType = std::filesystem::path;
+}
+
+#endif
--- a/Modules/IO/ImageBase/include/itkImageIOBase.h
+++ b/Modules/IO/ImageBase/include/itkImageIOBase.h
@@ -20,6 +20,7 @@
 #include "ITKIOImageBaseExport.h"
 
 #include "itkIOConfigure.h"
+#include "itkPathType.h"
 
 #include "itkLightProcessObject.h"
 #include "itkIndent.h"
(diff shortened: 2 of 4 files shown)


Modules/IO/ImageBase/include/itkImageIOBase.h
```diff
@@ -710,8 +741,10 @@ class ITKIOImageBase_EXPORT ImageIOBase : public LightProcessObject
   /** Does the ImageIOBase object have enough info to be of use? */
   bool m_Initialized{};
 
+#ifndef ITK_FUTURE_LEGACY_REMOVE
   /** Filename to read */
-  std::string m_FileName{};
+  const std::string m_FileName{};
+#endif
```
N-Dekker, line 747: It would be preferable to make `m_FileName` private, instead of protected. Then the workaround (`const_cast<std::string &>`) would not be necessary at all. ImageIOBase has a GetFileName and a SetFileName anyway, so direct access to `m_FileName` should not be necessary.

Modules/Core/Common/include/itkPathType.h
```diff
@@ -0,0 +1,29 @@
...
+ *=========================================================================*/
+
+#ifndef itkPathType_h
+#define itkPathType_h
+
+#include <filesystem>
+
+namespace itk
+{
+using PathType = std::filesystem::path;
```
dzenanz, line 26: What is the need for this indirection?

Modules/IO/JPEG/include/itkJPEGImageIO.h
```diff
@@ -117,7 +117,7 @@ class ITKIOJPEG_EXPORT JPEGImageIO : public ImageIOBase
   PrintSelf(std::ostream & os, Indent indent) const override;
 
   void
-  WriteSlice(std::string & fileName, const void * const buffer);
+  WriteSlice(const std::string & fileName, const void * const buffer);
```
N-Dekker, line 120: This really looks like a "constness bug" that should be fixed anyway. Right? (Independent of using std::filesystem.) Maybe even for ITK 5.4?

dzenanz (commented):
This work is quite welcome! But perhaps wait with this potentially disruptive change until 5.4 is released?

Modules/IO/JPEG/include/itkJPEGImageIO.h
dzenanz: I second the motion to make this a separate PR which can be merged immediately.

N-Dekker:
@blowekamp Very interesting, thanks! I definitely like having a separate type for file paths, rather than just "std::string", for improved type safety. I'm also curious to see if it does improve non-ASCII character support in file names, on Windows.

Modules/Core/Common/include/itkPathType.h
blowekamp: I am adding a similar type in SimpleITK to transition from std::string to std::filesystem::path which is being conditioned to the preprocessor (maybe).  It also made wrapping with SWIG a little easier. 

It is shorter?

N-Dekker: Personally I would prefer to just use `std::filesystem::path`. (Unless it would need to be configurable, as in SimpleITK.) ITK already has so many typedefs (or type aliases, if you like). When doing code analysis, I often lose time, searching for the definition of a typedef. It doesn't always make the code more readable to have all those typedefs.

Modules/IO/JPEG/include/itkJPEGImageIO.h
blowekamp: done, and done.

seanm:
It looks like adopting `std::filesystem` would require increasing ITK's minimum macOS to 10.15:

https://stackoverflow.com/questions/58667853/does-use-of-c17-stdfilesystem-require-macos-10-15-xcode-11-1

Even though C++17 compilers exist on older version of macOS, the STL shipped on older versions does not support `std::filesystem`, from what I gather.

That would suck for me, I still support back to 10.13.

thewtex:
> I still support back to 10.13.

Our Python packages also require 10.11.

blowekamp:
std::filesystem is not supported with the currently support OSX SDK

PR #4566, opened 2024-04-09
DOC: ITK 5.4 RC 3 release documentation updates

thewtex:
More specifics on macOS package builds, Linux ARM builds.

### DIFF
--- a/Documentation/Maintenance/Release.md
+++ b/Documentation/Maintenance/Release.md
@@ -10,6 +10,7 @@ Current Maintainers
 The current ITK maintainers with a trusted GPG key are:
 
   * Matt McCormick (@thewtex) <matt.mccormick@kitware.com>
+  * Pablo Hernandez-Cerdan <pablo.hernandez.cerdan@outlook.com>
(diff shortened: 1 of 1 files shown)


phcerdan:
Looks good, what issues do we find in Linux ARM?
First time I hear about podman :1st_place_medal:

thewtex:
podman is more community-oriented than docker. Newer, but also getting better because of that :-)

On ARM Linux:

> cc1plus: fatal error: /work/ITK-cp311-cp311-manylinux_2_28_aarch64/Wrapping/Generators/CastXML/castxml/share/castxml/empty.cpp: No such file or directory

which is odd :thinking: . I am looking at bumping castxml.

thewtex:
> On ARM Linux:

>> cc1plus: fatal error: /work/ITK-cp311-cp311-manylinux_2_28_aarch64/Wrapping/Generators/CastXML/castxml/share/castxml/empty.cpp: No such file or directory

This and other build errors resolved in https://github.com/InsightSoftwareConsortium/ITK/pull/4573 . And move to the newer toolchain (the old one crashed when built with The Beast :bear: ).

PR #4572, opened 2024-04-11
ENH: Add macOS ARM Python CI

thewtex:
Also remove unused Python package installs.


hjmjohnson (approved):
This looks good to me.

PR #4573, opened 2024-04-11
Castxml linux arm

thewtex:


hjmjohnson (approved):
I don't see any problems with this.

PR #4574, opened 2024-04-11
secondary capture spacing

thewtex:
- ENH: Add more tests for DICOM Secondary Capture Spacing
- BUG: Check for Secondary Capture spacing following DICOM Part 3 Sect A.8.1.3


pieper (approved):
Looks correct as far as I can see (didn't build and test)

hjmjohnson (approved):

PR #4580, opened 2024-04-12
COMP: Use the c++ 17 common [[fallthrough]] attribute

hjmjohnson:
[[fallthrough]](C++17)	indicates that the fall through from the previous case label is intentional and should not be diagnosed by a compiler that warns on fall-through (attribute specifier)

Remove complicated #ifdef macro logic to enforce compiler specific behaviors.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkMacro.h
+++ b/Modules/Core/Common/include/itkMacro.h
@@ -264,20 +264,10 @@ namespace itk
 #  define itkExposeEnumValue(name) static_cast<int>(name)
 #endif
 
-// Use "ITK_FALLTHROUGH;" to annotate deliberate fall-through in switches,
-// use it analogously to "break;".  The trailing semi-colon is required.
-#if defined(__GNUC__) && !defined(__INTEL_COMPILER)
-#  if (__GNUC__ >= 7)
-#    define ITK_FALLTHROUGH __attribute__((fallthrough))
-#  endif
-#elif defined(__has_warning)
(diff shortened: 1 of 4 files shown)


N-Dekker (approved):
Thanks, Hans! Interestingly GitHub thinks that `[[fallthrough]](C++17)` links to https://github.com/InsightSoftwareConsortium/ITK/pull/C++17   😸  Anyway, no problem, approved 👍

thewtex (approved):

dzenanz (approved):

blowekamp:
This change looks good. How close are we to 5.4? Do we have another RC? Are we planning some kind of new feature freeze soon?

This seems like using a new C++17 feature. How certain are we that all compilers support this feature correctly?

hjmjohnson:
@blowekamp  I do think all compilers that support c++17 do support this.  It is not an onerous addition to the language. Many compilers have used similar non-standard methods.  c++17 standardized the mechanism.

I have not looked at "all compilers", but the common ones have definitive support for this feature.

thewtex:
> How close are we to 5.4? Do we have another RC? 

ITK 5.4RC3 was tagged recently, but an issue was discovered immediately -- 5.4RC4 will be tagged and announced with [the fix](https://github.com/InsightSoftwareConsortium/ITK/pull/4574) next week.

We should not add major new features until 5.4.0, but this should be fine -- we have had a C++17 requirement for a while, and it is passing all CI.

blowekamp:
Yes, this change "should" be fine. Let us get it in for 5.4rc4?

PR #4581, opened 2024-04-12
STYLE: Prefer c++17 [[maybe_unused]] attribute over (void)

hjmjohnson:
The use of (void)varname was a mechanism to silence compiler warnings prior to universal language support in c++17 for the [[maybe_unused]] attribute specifier.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/CMake/itkCheckHasFenvtStructMember.cxx
+++ b/Modules/Core/Common/CMake/itkCheckHasFenvtStructMember.cxx
@@ -21,15 +21,14 @@
 int
 main()
 {
-  fenv_t fenv;
+  [[maybe_unused]] fenv_t fenv;
 #if defined(ITK_CHECK_FENV_T_CONTROL)
-  (void)sizeof(fenv.__control);
+  [[maybe_unused]] const auto tempSize = sizeof(fenv.__control);
 #elif defined(ITK_CHECK_FENV_T_CONTROL_WORD)
-  (void)sizeof(fenv.__control_word);
+  [[maybe_unused]] const auto tempSize = sizeof(fenv.__control_word);
 #elif defined(ITK_CHECK_FENV_T_CW)
-  (void)sizeof(fenv.__cw);
+  [[maybe_unused]] const auto tempSize = sizeof(fenv.__cw);
 #else
-  (void)fenv;
 #  error \
     "Unknown fenv_t struct member test: Make sure to specify a compile definition of the form -DITK_CHECK_FENV_T_xxx"
 #endif
--- a/Modules/Core/Common/include/itkObjectFactoryBase.h
+++ b/Modules/Core/Common/include/itkObjectFactoryBase.h
@@ -218,12 +218,10 @@ public:
     {};
 
     // Factory registration, made thread-safe by "magic statics" (as introduced with C++11).
-    static const FactoryRegistration staticFactoryRegistration = [] {
+    [[maybe_unused]] static const FactoryRegistration staticFactoryRegistration = [] {
       RegisterFactoryInternal(TFactory::New());
       return FactoryRegistration{};
     }();
-
-    (void)staticFactoryRegistration;
   }
 
   /** Initialize the static members of ObjectFactoryBase. */
--- a/Modules/Core/Common/include/itkSingletonMacro.h
+++ b/Modules/Core/Common/include/itkSingletonMacro.h
@@ -25,11 +25,10 @@
 #ifndef itkSingletonMacro_h
 #define itkSingletonMacro_h
 
-#define itkInitGlobalsMacro(VarName)                       \
-  {                                                        \
-    static auto * staticGlobals = Get##VarName##Pointer(); \
-    (void)staticGlobals;                                   \
-  }                                                        \
+#define itkInitGlobalsMacro(VarName)                                        \
+  {                                                                         \
+    [[maybe_unused]] static auto * staticGlobals = Get##VarName##Pointer(); \
+  }                                                                         \
   ITK_MACROEND_NOOP_STATEMENT
 
 #define itkGetGlobalDeclarationMacro(Type, VarName) static Type * Get##VarName##Pointer()
--- a/Modules/Core/QuadEdgeMesh/test/itkQuadEdgeMeshEulerOperatorCreateCenterVertexTest.cxx
+++ b/Modules/Core/QuadEdgeMesh/test/itkQuadEdgeMeshEulerOperatorCreateCenterVertexTest.cxx
@@ -74,7 +74,7 @@ itkQuadEdgeMeshEulerOperatorCreateCenterVertexTest(int, char *[])
   std::cout << "OK" << std::endl;
 #endif
 
-  (void)createCenterVertex->GetNameOfClass();
+  [[maybe_unused]] const std::string name_of_class = createCenterVertex->GetNameOfClass();
 
   createCenterVertex->SetInput(mesh);
 #ifndef NDEBUG
--- a/Modules/Core/QuadEdgeMesh/test/itkQuadEdgeMeshEulerOperatorCreateCenterVertexTest.cxx
+++ b/Modules/Core/QuadEdgeMesh/test/itkQuadEdgeMeshEulerOperatorCreateCenterVertexTest.cxx
@@ -18,6 +18,7 @@
 
 #include "itkQuadEdgeMeshEulerOperatorCreateCenterVertexFunction.h"
 #include "itkQuadEdgeMeshEulerOperatorsTestHelper.h"
+#include "itkTestingMacros.h"
 
 int
 itkQuadEdgeMeshEulerOperatorCreateCenterVertexTest(int, char *[])
@@ -73,8 +74,8 @@ itkQuadEdgeMeshEulerOperatorCreateCenterVertexTest(int, char *[])
   }
   std::cout << "OK" << std::endl;
 #endif
-
-  [[maybe_unused]] const std::string name_of_class = createCenterVertex->GetNameOfClass();
+  ITK_TEST_EXPECT_EQUAL(std::string_view("QuadEdgeMeshEulerOperatorCreateCenterVertexFunction"),
+                        std::string_view(createCenterVertex->GetNameOfClass()));
 
   createCenterVertex->SetInput(mesh);
 #ifndef NDEBUG
--- a/Modules/Filtering/FFT/include/itkFFTWCommon.h
+++ b/Modules/Filtering/FFT/include/itkFFTWCommon.h
@@ -135,8 +135,6 @@ public:
 #  ifndef ITK_USE_CUFFTW
     const std::lock_guard<FFTWGlobalConfiguration::MutexType> lockGuard(FFTWGlobalConfiguration::GetLockMutex());
     fftwf_plan_with_nthreads(threads);
-#  else
-    (void)threads;
 #  endif
     // don't add FFTW_WISDOM_ONLY if the plan rigor is FFTW_ESTIMATE
     // because FFTW_ESTIMATE guarantee to not destroy the input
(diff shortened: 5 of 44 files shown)


Modules/Compatibility/Deprecated/include/itkTreeIteratorBase.hxx
```diff
@@ -124,7 +124,7 @@ TreeIteratorBase<TTreeType>::Add(ValueType element)
 
 template <typename TTreeType>
 bool
-TreeIteratorBase<TTreeType>::Add(int itkNotUsed(childPosition), ValueType element)
+TreeIteratorBase<TTreeType>::Add([[maybe_unused]] int childPosition, ValueType element)
```
N-Dekker, line 127: @hjmjohnson Sorry, I'm not in favor of replacing `itkNotUsed(param)` with `[[maybe_unused]]`.  `itkNotUsed(param)` is stronger than "maybe unused". `itkNotUsed` means _definitely_ not used at all. The compiler ensures that such a parameter is not used.

N-Dekker (changes requested):

Modules/Compatibility/Deprecated/include/itkTreeIteratorBase.hxx
N-Dekker: I do like `[[maybe_unused]]` as replacement of `(void)varname`, but not as a replacement for unused parameters. If a parameter is unused, I would either use `itkNotUsed(param)`, or simply leave the parameter unnamed.

blowekamp: I recall some places `(void)` was used to ensure that a symbol was not optimized away by the compiler. Does `[[maybe_unused]]` do the same thing?

N-Dekker: @blowekamp Can you please elaborate? Is the use of `(void)` to prevent optimization compiler-specific? If so, does it still work like that, nowadays? Is there any documentation on this behavior?

Modules/Core/Common/include/itkObjectFactoryBase.h
```diff
@@ -218,12 +218,10 @@ class ITKCommon_EXPORT ObjectFactoryBase : public Object
     {};
 
     // Factory registration, made thread-safe by "magic statics" (as introduced with C++11).
-    static const FactoryRegistration staticFactoryRegistration = [] {
+    [[maybe_unused]] static const FactoryRegistration staticFactoryRegistration = [] {
       RegisterFactoryInternal(TFactory::New());
       return FactoryRegistration{};
     }();
-
-    (void)staticFactoryRegistration;
```
blowekamp, line 226: This may be a case ( or a previous implementation of this ) or something similar where the compiler may have optimized away the variable used to hold results of initialization, but that variable is never directly accessed.

Modules/Compatibility/Deprecated/include/itkTreeIteratorBase.hxx
hjmjohnson: @blowekamp  I found only one place where this occurs in a test.  It appears that the compiler does preserve the variable when [[maybe_unused]] is called.

blowekamp: Thank you for investigating my concern. I was not able to reproduce the scenario with godbolt either.

thewtex (approved):

Modules/Core/Common/CMake/itkCheckHasFenvtStructMember.cxx
```diff
@@ -21,15 +21,14 @@
 int
 main()
 {
-  fenv_t fenv;
+  [[maybe_unused]] fenv_t fenv;
 #if defined(ITK_CHECK_FENV_T_CONTROL)
-  (void)sizeof(fenv.__control);
+  [[maybe_unused]] const auto tempSize = sizeof(fenv.__control);
```
N-Dekker, line 26: Basically this just tests that `fenv.__control` is there, right? Maybe then it would be nicer to make that a static_assert:

    static_assert( sizeof(fenv.__control) > 0 );

Just suggestion.

Modules/Core/Common/CMake/itkCheckHasFenvtStructMember.cxx
```diff
@@ -21,15 +21,14 @@
...
-  (void)sizeof(fenv.__control_word);
+  [[maybe_unused]] const auto tempSize = sizeof(fenv.__control_word);
 #elif defined(ITK_CHECK_FENV_T_CW)
-  (void)sizeof(fenv.__cw);
+  [[maybe_unused]] const auto tempSize = sizeof(fenv.__cw);
 #else
-  (void)fenv;
 #  error \
     "Unknown fenv_t struct member test: Make sure to specify a compile definition of the form -DITK_CHECK_FENV_T_xxx"
 #endif
```
N-Dekker, line 34: It seems unnecessarily to me to prevent a warning here, as it is directly followed by a compiler error anyway! So it won't solve any CI failures. Anyway, no problem.

Modules/Core/QuadEdgeMesh/test/itkQuadEdgeMeshEulerOperatorCreateCenterVertexTest.cxx
```diff
@@ -74,7 +74,7 @@ itkQuadEdgeMeshEulerOperatorCreateCenterVertexTest(int, char *[])
   std::cout << "OK" << std::endl;
 #endif
 
-  (void)createCenterVertex->GetNameOfClass();
+  [[maybe_unused]] const std::string name_of_class = createCenterVertex->GetNameOfClass();
```
N-Dekker, line 77: - Maybe do an `ITK_TEST_EXPECT_EQUAL` here as well, following pull request #4583 ?

N-Dekker (changes requested):

Modules/Filtering/FFT/include/itkFFTWCommon.h
```diff
@@ -129,14 +129,12 @@ class Proxy<float>
                ComplexType * in,
                PixelType *   out,
                unsigned int  flags,
-               int           threads = 1,
+               int           itkNotused(threads) = 1,
                bool          canDestroyInput = false)
   {
 #  ifndef ITK_USE_CUFFTW
     const std::lock_guard<FFTWGlobalConfiguration::MutexType> lockGuard(FFTWGlobalConfiguration::GetLockMutex());
     fftwf_plan_with_nthreads(threads);
```
N-Dekker, line 137: Requested changes because the `threads` parameter is actually used here!!! So please do `[[maybe_unused]] threads`, instead of `itkNotused(threads)`!

Modules/Filtering/FFT/include/itkFFTWCommon.h
```diff
@@ -228,15 +226,13 @@ class Proxy<float>
...
                ComplexType * out,
                unsigned int  flags,
-               int           threads = 1,
+               int           itkNotUsed(threads) = 1,
                bool          canDestroyInput = false)
   {
     //
 #  ifndef ITK_USE_CUFFTW
     const std::lock_guard<FFTWGlobalConfiguration::MutexType> lockGuard(FFTWGlobalConfiguration::GetLockMutex());
     fftwf_plan_with_nthreads(threads);
```
N-Dekker, line 235: Requested changes (2) because the `threads` parameter is actually used here! Please use `[[maybe_unused]]` in this case!

Modules/Filtering/FFT/include/itkFFTWCommon.h
```diff
@@ -331,14 +327,12 @@ class Proxy<float>
            ComplexType * out,
            int           sign,
            unsigned int  flags,
-           int           threads = 1,
+           int           itkNotUsed(threads) = 1,
            bool          canDestroyInput = false)
   {
 #  ifndef ITK_USE_CUFFTW
     const std::lock_guard<FFTWGlobalConfiguration::MutexType> lockGuard(FFTWGlobalConfiguration::GetLockMutex());
     fftwf_plan_with_nthreads(threads);
```
N-Dekker, line 335: Requested changes (3) because the `threads` parameter is actually used here! Please use `[[maybe_unused]]` in this case!

Modules/Filtering/FFT/include/itkFFTWCommonExtended.h
```diff
@@ -62,110 +62,119 @@ class ComplexToComplexProxy<float>
   using Self = ComplexToComplexProxy<float>;
 
   static PlanType
-  Plan_dft_c2r_1d(int n, ComplexType * in, PixelType * out, unsigned int flags, int threads = 1)
+  Plan_dft_c2r_1d(int n, ComplexType * in, PixelType * out, unsigned int flags, itkNotUsed(threads) = 1)
   {
 #  ifndef ITK_USE_CUFFTW
     const std::lock_guard<FFTWGlobalConfiguration::MutexType> lockGuard(FFTWGlobalConfiguration::GetLockMutex());
     fftwf_plan_with_nthreads(threads);
```
N-Dekker, line 69: @hjmjohnson  Thanks for using `[[maybe_unused]]` for `threads`, rather than `itkNotUsed` 👍 Can you please check each and every `threads` parameter, to make sure that all of them get a `[[maybe_unused]]`, rather than `itkNotUsed`, if they _might_ be used?

hjmjohnson: Argg.  I had a typo in my regex that looked for these.  I thought I did fix all of them.  Sorry, I'll take care of this soon.

Modules/Core/Common/CMake/itkCheckHasFenvtStructMember.cxx
N-Dekker: @hjmjohnson Just curious, what did you think of my suggestion to do `static_assert( sizeof(fenv.__control) > 0 )`, instead of introducing an unused `tempSize` variable?

Modules/Core/QuadEdgeMesh/test/itkQuadEdgeMeshEulerOperatorCreateCenterVertexTest.cxx
N-Dekker: Instead of introducing an unused variable, `name_of_class`, what about making it a "real" test? At least you could do something like:

    ITK_TEST_EXPECT_TRUE(createCenterVertex->GetNameOfClass());

Which tests that GetNameOfClass() returns a valid character string, not a null pointer.

hjmjohnson:
@thewtex -- This PR is getting old, and is one that will likely have merge-conflicts.  Is this one that can be merged at this point?

Modules/Core/Common/include/itkObjectFactoryBase.h
```diff
@@ -218,12 +218,10 @@ class ITKCommon_EXPORT ObjectFactoryBase : public Object
     {};
 
     // Factory registration, made thread-safe by "magic statics" (as introduced with C++11).
-    static const FactoryRegistration staticFactoryRegistration = [] {
+    [[maybe_unused]] static const FactoryRegistration staticFactoryRegistration = [] {
```
blowekamp, line 221: This is a case I have potential concerns about.

Modules/Core/Common/include/itkSingletonMacro.h
```diff
@@ -25,11 +25,10 @@
 #ifndef itkSingletonMacro_h
 #define itkSingletonMacro_h
 
-#define itkInitGlobalsMacro(VarName)                       \
-  {                                                        \
-    static auto * staticGlobals = Get##VarName##Pointer(); \
-    (void)staticGlobals;                                   \
```
blowekamp, line 31: Another case.

blowekamp (changes requested):
I noted  the two cases I saw, where the void maybe used to prevent a variable from being optimized away. Although I was not able reproduce this scenario, I am hesitant to merged those changes right before the release.

If those two cases were separated into another PR, I think this PR is OK, if @thewtex is OK with it and the release schedual.

hjmjohnson:
@blowekamp I'll wait until next month after the release.  Hopefully, there won't be too many merge conflicts.

hjmjohnson:
Making a draft until until after 5.4 release is complete.

thewtex:
Thank you for saving this improvement for after v5.4.0 :+1:

seanm:
Isn't `[[maybe_unused]]`` C++17, not C++11?

thewtex:
@hjmjohnson v5.4.0 -- please merge when you see fit.

hjmjohnson:
> Isn't `[[maybe_unused]]` C++17, not C++11?

Updated comments.

thewtex (approved):
:1234:

PR #4585, opened 2024-04-15
STYLE: Use transformType in transform dict representation

thewtex:
This is to be more consistent with `pixelType`, etc. for Image, Mesh,
PolyData representations.


dzenanz (approved):

PranjalSahu (approved):
LGTM

PR #4583, opened 2024-04-13
STYLE: Use `ITK_TEST_EXPECT_EQUAL` to compare class names in QEMesh

jhlegarreta:
Prefer using `ITK_TEST_EXPECT_EQUAL` to compare class names in `QuadEdgeMesh` tests.

## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [X] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
-

PR #4576, opened 2024-04-11
TEST: Add test for correct value of GetNameOfClass()

hjmjohnson:
The result of GetNameOfClass() was ignored in the tests. The value is now checked to ensure that the correct values are used.

PR #4587, opened 2024-04-15
STYLE: Replace `(const std::string)` casts with C++17 `std::string_view`

N-Dekker:
Prevented potential static analysis warnings, like Clang-Tidy:

> C-style casts are discouraged; use static_cast [google-readability-casting]

And JetBrains ReSharper:

> C-style cast is used instead of a C++ cast

Note that `std::string_view` is more lightweight than `std::string`.

- Follow-up to pull request #4583

### DIFF
--- a/Modules/Core/QuadEdgeMesh/test/itkQuadEdgeMeshEulerOperatorDeleteCenterVertexTest.cxx
+++ b/Modules/Core/QuadEdgeMesh/test/itkQuadEdgeMeshEulerOperatorDeleteCenterVertexTest.cxx
@@ -83,8 +83,8 @@ itkQuadEdgeMeshEulerOperatorDeleteCenterVertexTest(int argc, char * argv[])
   }
   std::cout << "OK" << std::endl;
 
-  ITK_TEST_EXPECT_EQUAL((const std::string) "QuadEdgeMeshEulerOperatorDeleteCenterVertexFunction",
-                        (const std::string)deleteCenterVertex->GetNameOfClass());
+  ITK_TEST_EXPECT_EQUAL(std::string_view("QuadEdgeMeshEulerOperatorDeleteCenterVertexFunction"),
+                        std::string_view(deleteCenterVertex->GetNameOfClass()));
 
   deleteCenterVertex->SetInput(mesh);
   std::cout << "     "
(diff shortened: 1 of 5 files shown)


dzenanz (approved):

dzenanz:
> GoogleTest has a few macro's specifically for string comparison

Make a competing PR using those? 😄

N-Dekker:
Note that `std::string_view` only provides read-only to the data, so in that sense, it's similar to `const std::string &`.

thewtex (approved):

jhlegarreta (approved):

Modules/Core/QuadEdgeMesh/test/itkQuadEdgeMeshEulerOperatorDeleteCenterVertexTest.cxx
```diff
@@ -83,8 +83,8 @@ itkQuadEdgeMeshEulerOperatorDeleteCenterVertexTest(int argc, char * argv[])
   }
   std::cout << "OK" << std::endl;
 
-  ITK_TEST_EXPECT_EQUAL((const std::string) "QuadEdgeMeshEulerOperatorDeleteCenterVertexFunction",
-                        (const std::string)deleteCenterVertex->GetNameOfClass());
+  ITK_TEST_EXPECT_EQUAL(std::string_view("QuadEdgeMeshEulerOperatorDeleteCenterVertexFunction"),
+                        std::string_view(deleteCenterVertex->GetNameOfClass()));
```
N-Dekker, line 87: Thanks for merging, @hjmjohnson! For the record, none of the recent versions of this test (pull request #4576, #4583, #4587) properly handle the possibility of `deleteCenterVertex->GetNameOfClass()` accidentally returning a null. I was hoping that the `std::string_view(const char*)` constructor would support a null char pointer as argument, but unfortunately it does not. Neither does the originally used `std::string(const char*)` constructor. It's undefined behavior 🤷 

If you just want to ensure that this `GetNameOfClass()` override has code coverage, you might as well just checks that the returned pointer is not null. For example as follows:

    ITK_TEST_EXPECT_TRUE(deleteCenterVertex->GetNameOfClass());

Instead of doing the string comparison check.

dzenanz: Is `GetNameOfClass()` ever expected to return a `NULL` pointer? So that problem might be moot.

N-Dekker: > Is `GetNameOfClass()` ever expected to return a `NULL` pointer?

No, but unit tests allow us to express what we expect, and what we do not expect 😸 

Obviously it's easy for anyone to write a malicious `GetNameOfClass()` override that _does_ return NULL. 😸 😸

PR #4589, opened 2024-04-15
BUG: Fix Py_LIMITED_API minor version spec

thewtex:
This is in hex, we want 11 in decimal.

Ref:

  https://docs.python.org/3/c-api/apiabiversion.html#c.PY_VERSION_HEX


phcerdan (approved):
Good catch :eyes:

PR #4594, opened 2024-04-17
COMP: Add testing of IsCongruentImageGeometry IsSameImageGeometryAs

blowekamp:
## PR Checklist

### DIFF
--- a/Modules/Core/Common/test/itkImageGTest.cxx
+++ b/Modules/Core/Common/test/itkImageGTest.cxx
@@ -18,6 +18,7 @@
 
 // First include the header file to be tested:
 #include "itkImage.h"
+#include "itkImageDuplicator.h"
 #include <gtest/gtest.h>
 
 namespace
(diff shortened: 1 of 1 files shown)


hjmjohnson (approved):

N-Dekker:
Very happy to see another GoogleTest unit test in ITK, thanks @blowekamp ! Shall I still review in detail?

Modules/Core/Common/test/itkImageGTest.cxx
```diff
@@ -18,6 +18,7 @@
 
 // First include the header file to be tested:
 #include "itkImage.h"
+#include "itkImageDuplicator.h"
```
N-Dekker, line 21: @blowekamp  Looks like this `#include "itkImageDuplicator.h"` was unnecessary, right?

Modules/Core/Common/test/itkImageGTest.cxx
```diff
@@ -204,3 +205,45 @@ TEST(Image, FillBufferOfNonEqualityComparableType)
     EXPECT_EQ(image->GetPixel({}).data, i);
   }
 }
+
+template <typename ImageType>
+typename ImageType::Pointer generate_image(std::array<typename ImageType::SizeValueType, 2> size)
```
N-Dekker, line 210: @blowekamp  Why did you use `std::array<typename ImageType::SizeValueType, 2>`, instead of `ImageType::SizeType`? `itk::Size<VDimension>` is an aggregate struct, that supports being initialized by `{ 2, 2 }`

blowekamp: It was a struggle to be able to use the initializer list syntax with many of the type/methods in ITK i.e. SizeType, PointType, VectorType. I didn't realize the initializer list would work here as expected.

I'll make a follow up patch with your suggestions.

blowekamp: I was really hoping `image->SetSpacing({1.0,1.0})' would work.

dzenanz: Take a look at my struggles: #3006.

N-Dekker: For what it's worth, the syntax `image->SetSpacing({ { 1.0, 2.0 } })` might work if we would decide to remove the `explicit` keyword from the `Vector(const std::array &)` constructor:

https://github.com/InsightSoftwareConsortium/ITK/blob/7f1c5846087f463ecb7875fb36053096911df92e/Modules/Core/Common/include/itkVector.h#L138

However, removing the `explicit` keyword would be an API change, so I would not do that until after the release of ITK 5.4.0. It _might_ potentially introduce some ambiguities.

N-Dekker: To mitigate the struggle, we also added a `itk::MakeVector` helper function (pull request #2633 commit 8aed68490b733ba45b36b2e8e84e99980db45948, included with ITK 5.3), which automagically deduces the scalar value type and the dimension:

    image->SetSpacing( itk::MakeVector(1.0, 2.0) );

PR #4595, opened 2024-04-18
BUG: Set NIFTI header xyzt_units field correctly in output

cookpa:
Previously, the correct xyzt code was assigned to the 'xyz_units' variable, but time gets masked out because it's only for space units, resulting in xyzt_units=2.

Fix by setting 'time_units' to NIFTI_UNITS_SEC.

This produces images with xyzt_units == 10, which is correct for ITK output (NIFTI_UNITS_MM | NIFTI_UNITS_SEC)

## PR Checklist
- [*] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [*] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [*] Added test (or behavior not changed)
- [*] Updated API documentation (or API not changed)
- [*] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)

### DIFF
--- a/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
+++ b/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
@@ -1309,7 +1309,7 @@ NiftiImageIO::ReadImageInformation()
   //
   // set up the dimension stuff
   double spacingscale = 1.0; // default to mm
-  switch (XYZT_TO_SPACE(this->m_NiftiImage->xyz_units))
+  switch (this->m_NiftiImage->xyz_units)
   {
     case NIFTI_UNITS_METER:
       spacingscale = 1e3;
@@ -1322,7 +1322,7 @@ NiftiImageIO::ReadImageInformation()
       break;
   }
   double timingscale = 1.0; // Default to seconds
-  switch (XYZT_TO_TIME(this->m_NiftiImage->xyz_units))
+  switch (this->m_NiftiImage->time_units)
   {
     case NIFTI_UNITS_SEC:
       timingscale = 1.0;
@@ -1393,7 +1393,7 @@ NiftiImageIO::ReadImageInformation()
   EncapsulateMetaData<std::string>(thisDic, ITK_InputFilterName, classname);
 
   // set the image orientation
-  this->SetImageIOOrientationFromNIfTI(dims);
+  this->SetImageIOOrientationFromNIfTI(dims, spacingscale, timingscale);
 
   // Set the metadata.
   this->SetImageIOMetadataFromNIfTI();
@@ -1503,13 +1503,14 @@ NiftiImageIO::WriteImageInformation()
   //                   dim[2] is required for 2-D volumes,
   //                   dim[3] for 3-D volumes, etc.
   this->m_NiftiImage->nvox = 1;
-  // Spacial dims in ITK are given in mm.
+  // Spatial dims in ITK are given in mm.
   // If 4D assume 4thD is in SECONDS, for all of ITK.
   // NOTE: Due to an ambiguity in the nifti specification, some developers
   // external tools believe that the time units must be set, even if there
-  // is only one dataset.  Having the time specified for a purly spatial
+  // is only one dataset.  Having the time specified for a purely spatial
   // image has no consequence, so go ahead and set it to seconds.
-  this->m_NiftiImage->xyz_units = static_cast<int>(NIFTI_UNITS_MM | NIFTI_UNITS_SEC);
+  this->m_NiftiImage->xyz_units = static_cast<int>(NIFTI_UNITS_MM);
+  this->m_NiftiImage->time_units = static_cast<int>(NIFTI_UNITS_SEC);
   this->m_NiftiImage->dim[7] = this->m_NiftiImage->nw = 1;
   this->m_NiftiImage->dim[6] = this->m_NiftiImage->nv = 1;
   this->m_NiftiImage->dim[5] = this->m_NiftiImage->nu = 1;
@@ -1537,6 +1538,9 @@ NiftiImageIO::WriteImageInformation()
     case 4:
       this->m_NiftiImage->dim[4] = this->m_NiftiImage->nt = this->GetDimensions(3);
       this->m_NiftiImage->pixdim[4] = this->m_NiftiImage->dt = static_cast<float>(this->GetSpacing(3));
+      // Add time origin here because it is not set with the spatial origin in
+      // SetNIfTIOrientationFromImageIO
+      this->m_NiftiImage->toffset = static_cast<float>(this->GetOrigin(3));
       this->m_NiftiImage->nvox *= this->m_NiftiImage->dim[4];
       [[fallthrough]];
     case 3:
@@ -1861,7 +1865,7 @@ IsAffine(const mat44 & nifti_mat)
 } // namespace
 
 void
-NiftiImageIO::SetImageIOOrientationFromNIfTI(unsigned short dims)
+NiftiImageIO::SetImageIOOrientationFromNIfTI(unsigned short dims, double spacingscale, double timingscale)
 {
   typedef SpatialOrientationAdapter OrientAdapterType;
   // in the case of an Analyze75 file, use old analyze orient method.
@@ -1929,7 +1933,7 @@ NiftiImageIO::SetImageIOOrientationFromNIfTI(unsigned short dims)
     // so check it first.
     // Commonly the 4x4 double precision dicom information is stored in
     // the 4x4 single precision sform fields, and that original representation
-    // is converted (with lossy conversoin) into the qform representation.
+    // is converted (with lossy conversion) into the qform representation.
     const bool qform_sform_are_similar = [=]() -> bool {
       vnl_matrix_fixed<float, 4, 4> sto_xyz{ &(this->m_NiftiImage->sto_xyz.m[0][0]) };
       vnl_matrix_fixed<float, 4, 4> qto_xyz{ &(this->m_NiftiImage->qto_xyz.m[0][0]) };
@@ -2122,14 +2126,18 @@ NiftiImageIO::SetImageIOOrientationFromNIfTI(unsigned short dims)
 
   //
   // set origin
-  m_Origin[0] = -theMat.m[0][3];
+  m_Origin[0] = -theMat.m[0][3] * spacingscale;
   if (dims > 1)
   {
-    m_Origin[1] = -theMat.m[1][3];
+    m_Origin[1] = -theMat.m[1][3] * spacingscale;
   }
   if (dims > 2)
   {
-    m_Origin[2] = theMat.m[2][3];
+    m_Origin[2] = theMat.m[2][3] * spacingscale;
+  }
+  if (dims > 3)
+  {
+    m_Origin[3] = this->m_NiftiImage->toffset * timingscale;
   }
 
   const int           max_defined_orientation_dims = (dims > 3) ? 3 : dims;
@@ -2364,7 +2372,7 @@ NiftiImageIO::Write(const void * buffer)
                                         // so will destructor of the image that really owns it.
     if (nifti_write_status)
     {
-      itkExceptionMacro("ERROR: nifti library failed to write image" << this->GetFileName());
+      itkExceptionMacro("ERROR: nifti library failed to write image: " << this->GetFileName());
     }
   }
   else /// Image intent is vector image
@@ -2469,7 +2477,7 @@ NiftiImageIO::Write(const void * buffer)
     this->m_NiftiImage->data = nullptr; // if left pointing to data buffer
     if (nifti_write_status)
     {
-      itkExceptionMacro("ERROR: nifti library failed to write image" << this->GetFileName());
+      itkExceptionMacro("ERROR: nifti library failed to write image: " << this->GetFileName());
     }
   }
 }
--- a/Modules/IO/NIFTI/test/itkNiftiImageIOTest14.cxx
+++ b/Modules/IO/NIFTI/test/itkNiftiImageIOTest14.cxx
@@ -1,0 +1,171 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         https://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+#include "itkNiftiImageIOTest.h"
+#include "itkMath.h"
+#include "itkTestingMacros.h"
+
+// Test scaling of different spatial or temporal units
+// ITK uses mm for space and s for time. NIFTI allows other storage units
+// These should be converted to mm and s on output
+//
+// NIFTI xyzt_units field should be 10 (mm, sec) for all output
+int
+itkNiftiImageIOTest14(int argc, char * argv[])
+{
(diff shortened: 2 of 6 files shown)


hjmjohnson (approved):

Modules/IO/NIFTI/test/itkNiftiImageIOTest14.cxx
```diff
@@ -0,0 +1,104 @@
...
+    if (xyzt_units != "10")
+    {
+      std::cerr << "xyzt_units not set correctly in metadata" << std::endl;
+      return EXIT_FAILURE;
+    }
+  }
+  catch (...)
+  {
+    itk::IOTestHelper::Remove(filename);
+    return EXIT_FAILURE;
```
N-Dekker, line 100: Thanks for your PR, @cookpa Looking at these lines of code, I was wondering, should this file ("xyzt_unit_check.nii") _only_ be removed in case of a failure (`EXIT_FAILURE`)? Or also in the EXIT_SUCCESS case?

Modules/IO/NIFTI/test/itkNiftiImageIOTest14.cxx
```diff
@@ -0,0 +1,104 @@
...
+  ImageType::Pointer image;
+
+  // fill out an image and write it as nifti.
+  const char * filename = "xyzt_unit_check.nii";
+  using OutputImageType = itk::Image<PixelType, Dimension>;
+  typename OutputImageType::RegionType region;
+  typename OutputImageType::IndexType  start;
+  start[0] = 0;
+  start[1] = 0;
+  start[2] = 0;
```
N-Dekker, line 52: Obviously, if the start index is all zero, it does not need to be set, explicitly. By default, a region starts at index zero.

Modules/IO/NIFTI/test/itkNiftiImageIOTest14.cxx
```diff
@@ -0,0 +1,104 @@
...
+  size[2] = 4;
+  region.SetSize(size);
+  region.SetIndex(start);
+  auto outputimage = OutputImageType::New();
+  outputimage->SetRegions(region);
+  outputimage->Allocate();
+
+  using WriterType = itk::ImageFileWriter<OutputImageType>;
+  auto                       writer = WriterType::New();
+  itk::NiftiImageIO::Pointer niftiImageIO(itk::NiftiImageIO::New());
```
N-Dekker, line 65: Just like `auto writer = WriterType::New()`, you may of course also do `auto niftiImageIO = itk::NiftiImageIO::New()`. Just a suggestion.

Modules/IO/NIFTI/test/itkNiftiImageIOTest14.cxx
cookpa: You're right, I'll fix this so it always gets removed

Modules/IO/NIFTI/test/itkNiftiImageIOTest14.cxx
cookpa: Thanks, I took this from another test. Is it OK if I fix the original also in this PR?

https://github.com/InsightSoftwareConsortium/ITK/blob/7f1c5846087f463ecb7875fb36053096911df92e/Modules/IO/NIFTI/test/itkNiftiImageIOTest5.cxx#L146-L148

N-Dekker: >  Is it OK if I fix the original also in this PR?

I think so, yes 😃 My preference would be to fix the original in a separate commit. (Rather than squashing everything into one commit.) But it could be part of the same PR, sure. 👍

Modules/IO/NIFTI/test/itkNiftiImageIOTest14.cxx
```diff
@@ -0,0 +1,104 @@
...
+
+// Test writing NIFTI xyzt_units field - should be 10 (mm, sec)
+int
+itkNiftiImageIOTest14(int argc, char * argv[])
+{
+
+  if (argc > 1)
+  {
+    char * testdir = *++argv;
+    itksys::SystemTools::ChangeDirectory(testdir);
```
N-Dekker, line 30: FWIW, I would rather just do `testdir = argv[1]`, without the pointer increment (`++argv`).

cookpa: Thanks I will change this, also don't need the following `success` variable

Modules/IO/NIFTI/test/itkNiftiImageIOTest14.cxx
cookpa: Thanks, I updated with your other feedback, will add a separate commit for the other files

cookpa: Right you are, I removed start also. Just testing now

Modules/IO/NIFTI/test/itkNiftiImageIOTest14.cxx
```diff
@@ -0,0 +1,109 @@
...
+  const char * filename = "xyzt_unit_check.nii";
+  using OutputImageType = itk::Image<PixelType, Dimension>;
+  typename OutputImageType::RegionType region;
+  typename OutputImageType::IndexType  start;
+  typename OutputImageType::SizeType   size;
+  size[0] = 8;
+  size[1] = 8;
+  size[2] = 4;
+  region.SetSize(size);
+  region.SetIndex(start);
```
N-Dekker, line 54: So again, I would suggest to remove the `region.SetIndex(start)` call. But if you _really_ want that call it like that, you should make sure that `start` is properly initialized. (You can do so, just by adding a pair of curly braces to its declaration: `typename OutputImageType::IndexType start{};`) But I would suggest to simply remove the `start` variable.

cookpa: Thanks, I just want a small blank image as I'm only testing the header. I got rid of start entirely and also changed Allocate() to AllocateInitialized(), does this look better?

```
  // fill out an image and write it as nifti.
  const char * filename = "xyzt_unit_check.nii";
  using OutputImageType = itk::Image<PixelType, Dimension>;
  typename OutputImageType::RegionType region;
  typename OutputImageType::SizeType   size;
  size[0] = 8;
  size[1] = 8;
  size[2] = 4;
  region.SetSize(size);
  auto outputimage = OutputImageType::New();
  outputimage->SetRegions(region);
  outputimage->AllocateInitialized();
```

N-Dekker: Thanks @cookpa I guess with Allocate(), "xyzt_unit_check.nii" would just contain arbitrary pixel values, which would probably be OK to your test. It's just the C++ defines read-access to uninitialized data as "undefined behavior", meaning that _anything_ might have happened.

BTW, `itk::Size<N>` is a simple aggregate struct, which supports list-initialization, for example as follows:

    typename OutputImageType::SizeType   size{ 8, 8, 4 };

So that you don't really need to do `size[0] = 8; size[1] = 8; size[2] = 4;`. But that's just a matter of style.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -1503,13 +1503,14 @@ NiftiImageIO::WriteImageInformation()
...
+  // Spatial dims in ITK are given in mm.
   // If 4D assume 4thD is in SECONDS, for all of ITK.
   // NOTE: Due to an ambiguity in the nifti specification, some developers
   // external tools believe that the time units must be set, even if there
-  // is only one dataset.  Having the time specified for a purly spatial
+  // is only one dataset.  Having the time specified for a purely spatial
   // image has no consequence, so go ahead and set it to seconds.
-  this->m_NiftiImage->xyz_units = static_cast<int>(NIFTI_UNITS_MM | NIFTI_UNITS_SEC);
+  this->m_NiftiImage->xyz_units = static_cast<int>(NIFTI_UNITS_MM);
+  this->m_NiftiImage->time_units = static_cast<int>(NIFTI_UNITS_SEC);
```
N-Dekker, line 1513: Just for my understanding, your pull request should ensure that `NiftiImageIO::SetImageIOMetadataFromNIfTI()` will store "10" in its dictionary, after it does `xyzt_units << SPACE_TIME_TO_XYZT(nim->xyz_units, nim->time_units)`, right? At 
https://github.com/InsightSoftwareConsortium/ITK/blob/7f1c5846087f463ecb7875fb36053096911df92e/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx#L948-L949

cookpa: I believe it should store whatever the metadata was in the original NIFTI file. But when the image is read, any the spacing and time should be adjusted to MM and SEC in the ITK image. Then when we write, the image should have units MM and SEC (xyzt_units=10) regardless of what the input was.

I think the metadata is being written correctly now because when a file is written, it has xyzt_units=10 and not 2. But I just noticed the time scaling may not be happening correctly:

https://github.com/InsightSoftwareConsortium/ITK/blob/7f1c5846087f463ecb7875fb36053096911df92e/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx#L1325-L1336

I'm not sure why it's even calling the macros here, `nim->xyz_units` and `nim->time_units` are already masked versions of the header's `xyzt_units`.

I'll extend the tests to catch this. I will need to make a new test image with ms time units.

N-Dekker: @cookpa Cool! By the way, if you would still like to extend the tests as part of this pull request, you might want to click "Convert to draft", to indicate that it is still in progress. Otherwise, your PR might accidentally get merged before you are finished!

cookpa:
Quick update, I found some further problems with nonstandard units. Here are two headers with the same physical space, except one is in units of micron | milliseconds (xyzt_units = 19) and the other is in the usual mm | s (xyzt_units = 10).

```
 Spacing [0.4, 0.4, 0.4, 1.2]				      |	 Spacing [0.4, 0.4, 0.4, 1200]
 Origin [-0.5, -0.5, 0.5, 0]				      |	 Origin [-500, -500, 500, 0]
 Direction 							 Direction 
1 0 0 0								1 0 0 0
0 -1 0 0							0 -1 0 0
0 0 1 0								0 0 1 0
0 0 0 1								0 0 0 1

 Size : [8, 8, 4, 5]						 Size : [8, 8, 4, 5]

  Image Dimensions   : [8, 8, 4, 5]				  Image Dimensions   : [8, 8, 4, 5]
  Bounding Box       : {[-0.5 -0.5 0.5 0], [2.7 2.7 2.1 6]}   |	  Bounding Box       : {[-500 -500 500 0], [-496.8 -496.8 501
  Voxel Spacing      : [0.4, 0.4, 0.4, 1.2]		      |	  Voxel Spacing      : [0.4, 0.4, 0.4, 1200]
  Intensity Range    : [0, 0]					  Intensity Range    : [0, 0]
  Mean Intensity     : 0					  Mean Intensity     : 0
  Direction Cos Mtx. : 						  Direction Cos Mtx. : 
1 0 0 0								1 0 0 0
0 -1 0 0							0 -1 0 0
0 0 1 0								0 0 1 0
0 0 0 1								0 0 0 1

  Voxel->RAS x-form  : 						  Voxel->RAS x-form  : 
  Image Metadata: 						  Image Metadata: 
    ITK_original_direction of unsupported type N3itk6MatrixId	    ITK_original_direction of unsupported type N3itk6MatrixId
    ITK_original_spacing of unsupported type NSt3__16vectorId	    ITK_original_spacing of unsupported type NSt3__16vectorId
    ITK_sform_corrected = NO					    ITK_sform_corrected = NO
    bitpix = 64							    bitpix = 64
    cal_max = 0							    cal_max = 0
    cal_min = 0							    cal_min = 0
    datatype = 64						    datatype = 64
    dim[0] = 4							    dim[0] = 4
    dim[1] = 8							    dim[1] = 8
    dim[2] = 8							    dim[2] = 8
    dim[3] = 4							    dim[3] = 4
    dim[4] = 5							    dim[4] = 5
    dim[5] = 1							    dim[5] = 1
    dim[6] = 1							    dim[6] = 1
    dim[7] = 1							    dim[7] = 1
    dim_info = 0						    dim_info = 0
    intent_code = 0						    intent_code = 0
    intent_p1 = 0						    intent_p1 = 0
    intent_p2 = 0						    intent_p2 = 0
    intent_p3 = 0						    intent_p3 = 0
    nifti_type = 1						    nifti_type = 1
    pixdim[0] = 0						    pixdim[0] = 0
    pixdim[1] = 0.4					      |	    pixdim[1] = 400
    pixdim[2] = 0.4					      |	    pixdim[2] = 400
    pixdim[3] = 0.4					      |	    pixdim[3] = 400
    pixdim[4] = 1.2					      |	    pixdim[4] = 1200
    pixdim[5] = 0						    pixdim[5] = 0
    pixdim[6] = 0						    pixdim[6] = 0
    pixdim[7] = 0						    pixdim[7] = 0
    qfac = -1							    qfac = -1
    qform_code = 1						    qform_code = 1
    qform_code_name = NIFTI_XFORM_SCANNER_ANAT			    qform_code_name = NIFTI_XFORM_SCANNER_ANAT
    qoffset_x = 0.5					      |	    qoffset_x = 500
    qoffset_y = 0.5					      |	    qoffset_y = 500
    qoffset_z = 0.5					      |	    qoffset_z = 500
    qto_xyz of unsupported type N3itk6MatrixIfLj4ELj4EEE	    qto_xyz of unsupported type N3itk6MatrixIfLj4ELj4EEE
    quatern_b = 0						    quatern_b = 0
    quatern_c = 1						    quatern_c = 1
    quatern_d = 0						    quatern_d = 0
    scl_inter = 0						    scl_inter = 0
    scl_slope = 1						    scl_slope = 1
    sform_code = 1						    sform_code = 1
    sform_code_name = NIFTI_XFORM_SCANNER_ANAT			    sform_code_name = NIFTI_XFORM_SCANNER_ANAT
    slice_code = 0						    slice_code = 0
    slice_duration = 0						    slice_duration = 0
    slice_end = 0						    slice_end = 0
    slice_start = 0						    slice_start = 0
    srow_x = -0.4 0 0 0.5				      |	    srow_x = -400 0 -0 500
    srow_y = 0 0.4 0 0.5				      |	    srow_y = 0 400 -0 500
    srow_z = 0 0 0.4 0.5				      |	    srow_z = 0 0 400 500
    toffset = 0							    toffset = 0
    vox_offset = 352						    vox_offset = 352
    xyzt_units = 10					      |	    xyzt_units = 19
```

Correctly scaled:

* Spacing [x,y,z]

Incorrectly scaled:

* Spacing [t]
* Bounding Box
* Origin

To do for me:

* Fix scaling where necessary
* Test images in memory have equivalent spacing in 4D, bounding box, origin
* Verify that the image on the right above is correctly written to disk with quantities in mm | s units (xyzt_units=10).

cookpa:
I think this is good to go now. I made some more changes to ensure that both space and time get scaled into mm and seconds respectively, in both the pixel dimensions and the origin fields. The xyzt_units should now be read and written correctly.

Modules/IO/NIFTI/test/itkNiftiImageIOTest14.cxx
cookpa: I ended up not using the code to create a blank image, because I added test images to capture more failures. Now I test this by writing one of them back out and checking that the units are correct.

Modules/IO/NIFTI/test/itkNiftiImageIOTest14.cxx
cookpa: I ended up not using the code to create a blank image, because I added test images to capture more failures. Now I test this by writing one of them back out and checking that the units are correct.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
cookpa: Now updated to use this->m_NiftiImage->xyz_units for space and (this->m_NiftiImage->time_units for time

cookpa:
Apologies, will look at this today

PR #4599, opened 2024-04-19
WIP: Add static member, `itk::Image::CreateInitialized(const RegionType &)`

N-Dekker:
`itk::Image::CreateInitialized` aims to simplify creating an image with an allocated buffer of zero-initialized pixels.

Replaced more than sixty cases of code like:

```cpp
    auto image = ImageType::New();
    image->SetRegions(region);
    image->AllocateInitialized();
```
with the following single statement:

```cpp
    auto image = ImageType::CreateInitialized(region);
```

in ITK test files (`"itk*Test*.cxx"`).

### DIFF
--- a/Modules/Core/Common/include/itkImage.h
+++ b/Modules/Core/Common/include/itkImage.h
@@ -189,6 +189,16 @@ public:
   void
   Allocate(bool initializePixels = false) override;
 
+  /** Creates an image with the specified region. Allocates its pixel buffer, zero-initializing its pixels. */
+  static Pointer
+  CreateInitialized(const RegionType & region)
+  {
+    const auto image = Image::New();
+    image->SetRegions(region);
+    image->AllocateInitialized();
+    return image;
+  }
+
   /** Restore the data object to its initial state. This means releasing
    * memory. */
   void
--- a/Modules/Filtering/ImageGrid/test/itkSliceBySliceImageFilterTest.cxx
+++ b/Modules/Filtering/ImageGrid/test/itkSliceBySliceImageFilterTest.cxx
@@ -165,13 +165,11 @@ itkSliceBySliceImageFilterTest(int argc, char * argv[])
   // spacing. We are setting the input image to have a non-zero
   // starting index.
   //
-  auto image = ImageType::New();
-  {
+  auto image = ImageType::CreateInitialized([&reader] {
     ImageType::RegionType region = reader->GetOutput()->GetLargestPossibleRegion();
     region.SetIndex(0, 10);
-    image->SetRegions(region);
-    image->AllocateInitialized();
-  }
+    return region;
+  }());
 
   ImageType::SpacingType spacing;
   ImageType::PointType   origin;
--- a/Modules/Core/Common/test/itkAdaptorComparisonTest.cxx
+++ b/Modules/Core/Common/test/itkAdaptorComparisonTest.cxx
@@ -195,11 +195,9 @@ itkAdaptorComparisonTest(int, char *[])
   region.SetSize(size);
   region.SetIndex(index);
 
-  auto scalar_image = ScalarImageType::New();
-  auto vector_image = VectorImageType::New();
+  auto scalar_image = ScalarImageType::CreateInitialized(region);
 
-  scalar_image->SetRegions(region);
-  scalar_image->AllocateInitialized();
+  auto vector_image = VectorImageType::New();
 
   vector_image->SetRegions(region);
   vector_image->Allocate();
--- a/Modules/Core/Common/test/itkBSplineInterpolationWeightFunctionTest.cxx
+++ b/Modules/Core/Common/test/itkBSplineInterpolationWeightFunctionTest.cxx
@@ -275,13 +275,11 @@ itkBSplineInterpolationWeightFunctionTest(int, char *[])
     auto kernel = KernelType::New();
 
     using ImageType = itk::Image<char, SpaceDimension>;
-    auto                  image = ImageType::New();
     ImageType::RegionType region;
     region.SetIndex(startIndex);
     region.SetSize(size);
 
-    image->SetRegions(region);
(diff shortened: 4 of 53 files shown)


dzenanz (approved):

Modules/Core/Common/include/itkImage.h
```diff
@@ -189,6 +189,16 @@ class ITK_TEMPLATE_EXPORT Image : public ImageBase<VImageDimension>
   void
   Allocate(bool initializePixels = false) override;
 
+  /** Creates an image at the specified region. Allocates its pixel buffer, zero-initializing its pixels. */
```
dzenanz, line 192:
```suggestion
  /** Creates an image with the specified region. Allocates its pixel buffer, zero-initializing its pixels. */
```

N-Dekker: Thanks @dzenanz, that's fine _with_ me  😃

Modules/Filtering/ImageGrid/test/itkSliceBySliceImageFilterTest.cxx
```diff
@@ -165,13 +165,11 @@ itkSliceBySliceImageFilterTest(int argc, char * argv[])
...
-  auto image = ImageType::New();
-  {
+  auto image = ImageType::CreateInitialized([&reader] {
     ImageType::RegionType region = reader->GetOutput()->GetLargestPossibleRegion();
     region.SetIndex(0, 10);
-    image->SetRegions(region);
-    image->AllocateInitialized();
-  }
+    return region;
+  }());
```
N-Dekker, line 172: FYI, this was probably the most interesting use case that I found. In the original code, the variable `region` was already kept inside a very small scope, between `{` and `}`. The proposed pull request places the `region` variable inside a lambda, in order to keep its scope small.

blowekamp:
Are there any similarly named methods to "Obj::CreateInitialized"? What about something like `MakeImage`? We already have similarly name methods for other classes?

Note: I'm not suggesting preference, but starting a conversion about what may be the best and most consistent interface.

N-Dekker:
> Are there any similarly named methods to "Obj::CreateInitialized"? What about something like `MakeImage`?

Thanks for asking @blowekamp I agree that it's good to discuss the name.

- Of course, the currently proposed `itk::Image::CreateInitialized` is very much inspired by the name `itk::ImageBase::AllocateInitialized()`, which was the outcome of our discussion at #4479
- `itk::Image::CreateInitialized` is very much meant as an alternative to `itk::Image::New()`. Having it as a static member function of `itk::Image` makes it easier to migrate code from using `ImageType::New` to using `ImageType::CreateInitialized`.
- Another option could be to overload `Image::New()`, so that the new static member function would be `itk::Image::New(const RegionType& )`, with the region as extra argument. But I wasn't sure if overloading `Image::New` could cause ambiguities or make things unclear 🤷 
-  The name "CreateInitialized" eases adding more functions in the future like "CreateUninitialized" (allocating the buffer, but leaving the pixels uninitialized) or "CreateFilled" (doing a FillBuffer), if there is any interest.
- A more verbose name is also possible, like "CreateAndSetRegionsAndAllocateInitialized". Doesn't sound very appealing, though! 
- If we'd go for a free function in the `itk` namespace instead, like `itk::MakeImage`, we would still need to make a choice about its template argument(s). One template argument, `itk::MakeImage<TImage>` or two, `itk::MakeImage<TPixel, NDimension>`? Both have their pro's and cons.

What do you think?

blowekamp:
Good points. We also need to consider other image type such as VectorImage, and LabelMap and even module image types like RLE image. Perhaps `itk::MakeImage<TImage>` could trivially be made to work with all of them?

N-Dekker:
> Good points. We also need to consider other image type such as VectorImage, and LabelMap and even module image types like RLE image. Perhaps `itk::MakeImage<TImage>` could trivially be made to work with all of them?

VectorImage would need an extra parameter to specify the vector length. So I think it would need to have its own static member function `VectorImage::CreateInitialized(region, vectorLength)` (or its own free function, `itk::MakeVectorImage<TVectorImage>(region, vectorLength)`).

Modules/Core/Common/include/itkImage.h
N-Dekker: Addressed "with" [this force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/519ab58942d4c2baafc03a68ae74d561a5e36d30..a8ac2f1edb5b2a0ceb0e5a356de87bb505b4766b)

N-Dekker:
@blowekamp I guess a static `Image::CreateInitialized(const RegionType &)` member function is automatically wrapped by SWIG, for each supported pixel type and image dimension. On the other hand I guess a free function template like `itk::MakeImage<TImage>` would need to be "SWIG-ed" separately for each supported image type. Right?

Just something to be taken into account. I'm not _against_ a free function template like that, but I still prefer the `Image::CreateInitialized(const RegionType &)` member function that I'm currently proposing 😃

dzenanz:
Why request a review? This PR is 2 years old, and in a draft state.

hjmjohnson:
> Why request a review? This PR is 2 years old, and in a draft state.

I was trying to clear the "Approved" state since this is 2 years old and stale.  I was not trying to re-request review, but clear the "Approved" flag state on a WIP that can not be merged.  I did not accomplish removing the "Approved"

dzenanz:
> I did not accomplish removing the "Approved"

To me, it looks like you did. On the right panel, there is a yellow dot instead of green check mark.

N-Dekker:
I still like `itk::Image::CreateInitialized(const RegionType &)`, but we could not get to a consensus. 🤷So then I just forgot about it!

hjmjohnson:
Superceeded by https://github.com/InsightSoftwareConsortium/ITK/pull/6680

PR #4601, opened 2024-04-20
COMP: Add GDCM system version check for SetSecondaryCaptureImagePlaneModule

thewtex:
This has not been released in GDCM, yet. Avoid build errors for a system
GDCM. There will be a difference in behavior, which will be flagged with
failing tests.

Co-authored-by: Mihail Isakov <mihail.isakov@gmail.com>


thewtex:
Verified with system GDCM on Debian Bullseye.

issakomi:
Thank you! BTW, there is another call on line 284 (in 'Debug' mode).

issakomi (approved):

PR #4602, opened 2024-04-21
BUG: Provide an API for SpatialObjectProperties that works with python

aylward:
The dictionary used by SpatialObjectProperties is queried using a function that returns the dictionary value in a variable that is passed by reference: https://itk.org/Doxygen/html/classitk_1_1SpatialObjectProperty.html#aff2b1c659a8e77bdb6afeef0bab037fb

    bool itk::SpatialObjectProperty::GetTagScalarValue ( const std::string & tag, double & value ) const

This function is not correctly wrapped by SWIG, and cannot be used from within python.

Here is a snippet of code that demonstrates the issue:

    In [1]: import itk
    In [2]: so = itk.TubeSpatialObject[3].New()
    In [3]: so.GetProperty().SetTagScalarValue("foo",1) In [4]: ret = 0.0
    In [5]: so.GetProperty().GetTagScalarValue("foo",ret) 
    ---------------------------------------------------------------------------
    TypeError                                 Traceback (most recent call last)
    Cell In[5], line 1
    ----> 1 so.GetProperty().GetTagScalarValue("foo",ret)
    TypeError: in method 'itkSpatialObjectProperty_GetTagScalarValue', argument 3 of type 'double &'

This commit adds new functions that can be correctly wrapped.  Old functions are unmodified to preserve backward compatibility.

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkSpatialObjectProperty.h
+++ b/Modules/Core/SpatialObjects/include/itkSpatialObjectProperty.h
@@ -55,6 +55,12 @@ public:
     m_Color = color;
   }
 
+  ColorType
+  Color()
+  {
+    return m_Color;
+  }
+
   ColorType &
   GetColor()
   {
@@ -74,21 +80,41 @@ public:
   SetRed(double r);
   double
   GetRed() const;
+  double
+  Red() const
+  {
+    return GetRed();
+  }
 
   void
   SetGreen(double g);
   double
   GetGreen() const;
+  double
+  Green() const
+  {
+    return GetGreen();
+  }
 
   void
   SetBlue(double b);
   double
   GetBlue() const;
+  double
+  Blue() const
+  {
+    return GetBlue();
+  }
 
   void
   SetAlpha(double a);
   double
   GetAlpha() const;
+  double
+  Alpha() const
+  {
+    return GetAlpha();
+  }
 
   void
   SetName(const std::string & name)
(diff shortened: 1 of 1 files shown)


Modules/Core/SpatialObjects/include/itkSpatialObjectProperty.h
```diff
@@ -55,6 +55,12 @@ class ITKSpatialObjects_EXPORT SpatialObjectProperty
     m_Color = color;
   }
 
+  ColorType
+  Color()
+  {
+    return m_Color;
+  }
```
N-Dekker, line 62: Please declare Color() `const` (just like the other new member functions proposed by this pull request).

Modules/Core/SpatialObjects/include/itkSpatialObjectProperty.h
```diff
@@ -74,21 +80,41 @@ class ITKSpatialObjects_EXPORT SpatialObjectProperty
   SetRed(double r);
   double
   GetRed() const;
+  double
+  Red() const
+  {
+    return GetRed();
+  }
```
N-Dekker, line 87: `Red()` looks redundant to me. Doesn't the existing `GetRed()` get "swigged" properly?

N-Dekker:
Thanks for your pull request @aylward It would be very disappointing if SWIG would still not support pass by reference or return by reference. I still hope that it can be enabled somehow. For pass by reference, StackOverflow [Swig: How to wrap double& (double passed by reference)?](https://stackoverflow.com/questions/3470641/swig-how-to-wrap-double-double-passed-by-reference) suggested `%include <typemaps.i>` and some ` %apply` statement. Did you already try that out?

aylward:
@N-Dekker - I saw the same stackoverflow post, but I didn't pursue it because it changes the signature of the function when wrapping - causing the wrapped API to not match the C++ API.   I didn't care for that break in philosophy compared to the rest of ITK's wrapping.

I am going to close this pull request and make another that contains only changes for the necessary Get statements.  This pull request had other changes to make a consistent API (offering non-get versions of every function, even if the original wrapped as-is).   Instead I'll reduce to the minimal changes for simplicity.   I'll also mark the non-standard return by reference as functions to be deprecated.   Stay tuned...

Modules/Core/SpatialObjects/include/itkSpatialObjectProperty.h
```diff
@@ -55,6 +55,12 @@ class ITKSpatialObjects_EXPORT SpatialObjectProperty
...
   }
 
+  ColorType
+  Color()
+  {
+    return m_Color;
+  }
+
   ColorType &
   GetColor()
```
N-Dekker, line 65: @aylward Is a

Modules/Core/SpatialObjects/include/itkSpatialObjectProperty.h
```diff
@@ -108,25 +134,60 @@ class ITKSpatialObjects_EXPORT SpatialObjectProperty
     return m_Name;
   }
 
+  std::string
+  Name() const
+  {
+    return m_Name;
+  }
```
N-Dekker, line 141: According to https://www.swig.org/Doc4.2/SWIGDocumentation.html SWIG _does_ support `const std::string &`, but honestly it's not clear to me how 🤷  It says:

> The `%naturalvar` feature was introduced in SWIG-1.3.28, prior to which it was necessary to manually apply the const reference typemaps, eg `%apply const std::string & { std::string * }`, but this example would also apply the typemaps to methods taking a std::string pointer.

And:

> The naturalvar behavior can also be turned on as a global setting via the `-naturalvar` commandline option or the module mode option, `%module(naturalvar=1)`.

Would "the naturalvar behavior" be of help to us?

Modules/Core/SpatialObjects/include/itkSpatialObjectProperty.h
aylward: See comments used to close this commit.

N-Dekker: @aylward I did see your comments, I'm just trying to understand the issue. Apparently returning a `Color` reference seems to work fine, while some tweaks may be necessary to return an `std::string` reference. I did not know about that before.

PR #4603, opened 2024-04-21
STYLE: Revert C++14 workaround when using the constexpr `SupportSize`

N-Dekker:
C++17 allows using a static constexpr member variable like `SupportSize`, without being "redeclared". The workaround of copying `SupportSize` into a local variable is no longer necessary.

Reverts pull request https://github.com/InsightSoftwareConsortium/ITK/pull/2742 commit 27e4815bb816c2cebe3abd0b9aca85afda8101c6
"COMP: Copy BSplineInterpolationWeightFunction::SupportSize within lambda"

PR #4605, opened 2024-04-22
ENH: Declare `front()` and `back()` of Index, Offset, and Size constexpr

N-Dekker:
Follow-up to pull request https://github.com/InsightSoftwareConsortium/ITK/pull/3236 commit 47bce264791a53853020a3911dd94d060cfbd655
"ENH: Declare begin(), end() of FixedArray, Index, Offset, Size constexpr"

### DIFF
--- a/Modules/Core/Common/include/itkIndex.h
+++ b/Modules/Core/Common/include/itkIndex.h
@@ -445,25 +445,25 @@ public:
     return m_InternalArray[pos];
   }
 
-  reference
+  constexpr reference
   front()
   {
     return *begin();
   }
 
-  const_reference
+  constexpr const_reference
   front() const
   {
     return *begin();
   }
 
-  reference
+  constexpr reference
   back()
   {
     return VDimension ? *(end() - 1) : *end();
   }
 
-  const_reference
+  constexpr const_reference
   back() const
   {
     return VDimension ? *(end() - 1) : *end();
--- a/Modules/Core/Common/include/itkOffset.h
+++ b/Modules/Core/Common/include/itkOffset.h
@@ -397,25 +397,25 @@ public:
     return m_InternalArray[pos];
   }
 
-  reference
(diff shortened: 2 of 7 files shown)


dzenanz (approved):

N-Dekker:
@thewtex Would this PR still be feasible for v5.4.0? It just adds `constexpr` keywords to `front()` and `back()` member functions.

blowekamp:
> @thewtex Would this PR still be feasible for v5.4.0? It just adds `constexpr` keywords to `front()` and `back()` member functions.

Out of my curiosity, is this for a particular issue? Has it given any measurable performance benefit?

N-Dekker:
> Out of my curiosity, is this for a particular issue? Has it given any measurable performance benefit?

This PR would pave the way to adding constexpr to `itk::Index` iteration, by `itk::IndexRange`. Which would allow creating the OffsetToIndexTable of BSplineInterpolationWeightFunction at compile-time, instead of at run-time. I haven't yet been able to benchmark, but it may certainly benefit its performance.

Note that `itk::Index<N>` is similar to `std::array<itk::IndexValueType, N>`, whose `front()` and `back()` are also marked "constexpr": https://en.cppreference.com/w/cpp/container/array/back

thewtex (approved):
@N-Dekker this is awesome!

I think we should save it for ITK 6 to avoid any unforeseen issues.

N-Dekker:
> I think we should save it for ITK 6 to avoid any unforeseen issues.

@thewtex  OK, thanks. Marked "draft" to avoid that this PR might accidentally be merged before the release of ITK 5.4!

N-Dekker:
I think this pull request may also be merged now, as v5.4.0 has been tagged (https://github.com/InsightSoftwareConsortium/ITK/pull/4603#issuecomment-2120498131)

Issue #4607, opened 2024-04-23
Address GCC 13.2 compilation issues


hjmjohnson:
### Description

The forth coming LTR Ubuntu 24.04 (on April 25, 2024) is distributed with GCC 13.2 compiler.

ITK should build cleanly with this new standard compiler.

### Steps to Reproduce

1. clone itk
2. build in release mode (with Examples turned on)

### Expected behavior

no build warnings
no failing tests

### Actual behavior

3 test failures
```
	1012 - itkGDCMLegacyMultiFrameTest (Failed)
	1021 - itkGDCMImageReadWriteTest_MultiFrameMRIZSpacing (Failed)
	1023 - itkGDCM_ComplianceTest_singlebit (Failed)
```

199+ compiler warnings

```
```

### Reproducibility
all the time

### Versions
ITK master branch on 2024-04-23

git rev-parse --short HEAD
```
3235b3675c
```

### Environment

Ubuntu 24.04 
GCC 13.2 compiler

### Additional Information

```
242 |   alignas(SizeValueType) SizeValueType m_InternalArray[VDimension];
      |                                        ^~~~~~~~~~~~~~~
In member function ‘constexpr const itk::Size<VDimension>::value_type& itk::Size<VDimension>::operator[](size_type) const [with unsigned int VDimension = 2]’,
    inlined from ‘itk::Neighborhood<TPixel, VDimension, TAllocator>::SizeValueType itk::Neighborhood<TPixel, VDimension, TAllocator>::GetSize(DimensionValueType) const [with TPixel = float; unsigned int VDimension = 2; TAllocator = itk::NeighborhoodAllocator<float>]’ at [Modules/Core/Common/include/itkNeighborhood.h:147](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkNeighborhood.h#L147):18,
    inlined from ‘void itk::NeighborhoodOperator<TPixel, VDimension, TAllocator>::FillCenteredDirectional(const CoefficientVector&) [with TPixel = float; unsigned int VDimension = 2; TAllocator = itk::NeighborhoodAllocator<float>]’ at [Modules/Core/Common/include/itkNeighborhoodOperator.hxx:107](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkNeighborhoodOperator.hxx#L107):43,
[...]

PR #4608, opened 2024-04-23
COMP:  Ubuntu2404 fixes issue 4607

hjmjohnson:
##

Fixes for 100's of warnings presented by the gcc 13.2 compiler under the upcoming Ubuntu 24.04  LTS release.

This does not fix all of the issues, but is for comparison against the master branch build 

https://open.cdash.org/builds/9565813

## Additional information
Related to issue #4607 

These new compiler warnings are due to increased default security settings in Ubuntu 24.04.

https://ubuntu.com/blog/whats-new-in-security-for-ubuntu-24-04-lts#:~:text=Overall%2C%20the%20vast%20range%20of,most%20secure%20release%20to%20date.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkExtractImageFilter.hxx
+++ b/Modules/Core/Common/include/itkExtractImageFilter.hxx
@@ -119,11 +119,7 @@ ExtractImageFilter<TInputImage, TOutputImage>::GenerateOutputInformation()
   outputPtr->SetLargestPossibleRegion(m_OutputImageRegion);
 
   // Set the output spacing and origin
-  const ImageBase<InputImageDimension> * phyData;
-
-  phyData = dynamic_cast<const ImageBase<InputImageDimension> *>(this->GetInput());
-
-  if (phyData)
+  if (this->GetInput())
   {
     // Copy what we can from the image from spacing and origin of the input
     // This logic needs to be augmented with logic that select which
--- a/Modules/Core/Common/include/itkExtractImageFilter.hxx
+++ b/Modules/Core/Common/include/itkExtractImageFilter.hxx
@@ -62,7 +62,6 @@ ExtractImageFilter<TInputImage, TOutputImage>::SetExtractionRegion(InputImageReg
                 "InputImageDimension must be greater than OutputImageDimension");
   m_ExtractionRegion = extractRegion;
 
-  unsigned int        nonzeroSizeCount = 0;
   InputImageSizeType  inputSize = extractRegion.GetSize();
   OutputImageSizeType outputSize;
   outputSize.Fill(0);
@@ -73,12 +72,16 @@ ExtractImageFilter<TInputImage, TOutputImage>::SetExtractionRegion(InputImageReg
    * check to see if the number of non-zero entries in the extraction region
    * matches the number of dimensions in the output image.
    */
+  unsigned int nonzeroSizeCount = 0;
   for (unsigned int i = 0; i < InputImageDimension; ++i)
   {
     if (inputSize[i])
     {
-      outputSize[nonzeroSizeCount] = inputSize[i];
-      outputIndex[nonzeroSizeCount] = extractRegion.GetIndex()[i];
+      if (nonzeroSizeCount < OutputImageDimension)
+      {
+        outputSize[nonzeroSizeCount] = inputSize[i];
+        outputIndex[nonzeroSizeCount] = extractRegion.GetIndex()[i];
+      }
       ++nonzeroSizeCount;
     }
   }
--- a/Modules/Core/Common/include/itkNeighborhood.h
+++ b/Modules/Core/Common/include/itkNeighborhood.h
@@ -136,7 +136,7 @@ public:
   SizeValueType
   GetRadius(DimensionValueType n) const
   {
-    return m_Radius[n];
+    return m_Radius.at(n);
   }
 
   /** Returns the size (total length) of the neighborhood along
@@ -144,7 +144,7 @@ public:
   SizeValueType
   GetSize(DimensionValueType n) const
   {
-    return m_Size[n];
+    return m_Size.at(n);
   }
 
   /** Returns the size (total length of sides) of the neighborhood. */
--- a/Modules/Core/Common/test/itkShapedImageNeighborhoodRangeGTest.cxx
+++ b/Modules/Core/Common/test/itkShapedImageNeighborhoodRangeGTest.cxx
@@ -366,11 +366,14 @@ TEST(ShapedImageNeighborhoodRange, IteratorsCanBePassedToStdReverseCopy)
   itk::ShapedImageNeighborhoodRange<ImageType> range{ *image, location, offsets };
 
   const unsigned int numberOfNeighborhoodPixels = 3;
+  ASSERT_EQ(numberOfNeighborhoodPixels, range.size());
 
   const std::vector<PixelType> stdVector(range.begin(), range.end());
-  std::vector<PixelType>       reversedStdVector1(numberOfNeighborhoodPixels);
-  std::vector<PixelType>       reversedStdVector2(numberOfNeighborhoodPixels);
-  std::vector<PixelType>       reversedStdVector3(numberOfNeighborhoodPixels);
+  ASSERT_EQ(stdVector.size(), numberOfNeighborhoodPixels);
+
+  std::vector<PixelType> reversedStdVector1(numberOfNeighborhoodPixels);
+  std::vector<PixelType> reversedStdVector2(numberOfNeighborhoodPixels);
+  std::vector<PixelType> reversedStdVector3(numberOfNeighborhoodPixels);
 
   // Checks bidirectionality of the ShapedImageNeighborhoodRange iterators!
   std::reverse_copy(stdVector.cbegin(), stdVector.cend(), reversedStdVector1.begin());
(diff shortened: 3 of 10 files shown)


Modules/Core/Common/include/itkNeighborhood.h
```diff
@@ -136,6 +136,11 @@ class ITK_TEMPLATE_EXPORT Neighborhood
   SizeValueType
   GetRadius(DimensionValueType n) const
   {
+    if (n >= VDimension)
+    {
+      itkExceptionMacro(<< " Can not get radius for dimension " << n << " greater than dimensionality of neighborhood "
+                        << VDimension);
+    }
     return m_Radius[n];
   }
```
hjmjohnson, line 145: Compiler warnings like:

[Modules/Core/Common/include/itkSize.h:356](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/include/itkSize.h#L356):85: warning: array subscript 2 is above array bounds of ‘const itk::Size<2>::SizeValueType [2]’ {aka ‘const long unsigned int [2]’} [-Warray-bounds=]

Modules/Core/QuadEdgeMesh/test/itkQuadEdgeMeshEulerOperatorSplitEdgeTest.cxx
```diff
@@ -42,7 +42,7 @@ itkQuadEdgeMeshEulerOperatorSplitEdgeTest(int, char *[])
   auto splitEdge = SplitEdge::New();
   std::cout << "     "
             << "Test No Mesh Input";
-  if (splitEdge->Evaluate((QEType *)1))
+  if (splitEdge->Evaluate((QEType *){}))
```
dzenanz, line 45: Is default initializer the same as `1`?

hjmjohnson: Not according to the compiler:

```
ITK_master/Modules/Core/QuadEdgeMesh/include/itkQuadEdge.h:245:15: warning: array subscript 0 is outside array bounds of ‘QEType [0]’ [-Warray-bounds=]
```

It should be a nullptr I think.

Modules/Core/Common/include/itkExtractImageFilter.hxx
```diff
@@ -75,7 +75,7 @@ ExtractImageFilter<TInputImage, TOutputImage>::SetExtractionRegion(InputImageReg
    */
   for (unsigned int i = 0; i < InputImageDimension; ++i)
   {
-    if (inputSize[i])
+    if (inputSize[i] and nonzeroSizeCount < OutputImageDimension)
```
N-Dekker, line 78: Looking at https://github.com/InsightSoftwareConsortium/ITK/blob/9d1a93dfb8edb10bce000f9695f9e626b3ed2a26/Modules/Core/Common/include/itkExtractImageFilter.hxx#L86-L92

It appears that the originally intended behavior was to throw an exception when nonzeroSizeCount gets too big. However, your proposed solution would "silently" prevent nonzeroSizeCount from getting too big. That's different behavior! I don't know which behavior is preferred 🤷  

BTW (nitpick alarm 😸): Most C++ programmers use `&&`, instead of `and`!

Modules/Core/Common/test/itkShapedImageNeighborhoodRangeGTest.cxx
```diff
@@ -914,9 +914,11 @@ TEST(ShapedImageNeighborhoodRange, ProvidesReverseIterators)
   const unsigned int numberOfNeighborhoodPixels = 3;
 
   const std::vector<PixelType> stdVector(range.begin(), range.end());
-  std::vector<PixelType>       reversedStdVector1(numberOfNeighborhoodPixels);
-  std::vector<PixelType>       reversedStdVector2(numberOfNeighborhoodPixels);
-  std::vector<PixelType>       reversedStdVector3(numberOfNeighborhoodPixels);
+
+  assert(stdVector.size() == numberOfNeighborhoodPixels);
```
N-Dekker, line 918: Good catch  👍  Please make that `ASSERT_EQ(stdVector.size(), numberOfNeighborhoodPixels)`.  ASSERT_EQ is specific to GoogleTest. It ends the unit test _directly_ at that point, even in Release mode, and reports a test failure.

Modules/Core/Common/include/itkExtractImageFilter.hxx
dzenanz: VS [chokes](https://open.cdash.org/viewBuildError.php?buildid=9565895) on `and`:
D:\a\1\s\Modules\Core\Common\include\itkExtractImageFilter.hxx(78): error C2065: 'and': undeclared identifier

hjmjohnson: my mistake.  jumping between python an C++.  I was doing this in the background while waiting for my python data analysis to complete.  Sorry.

hjmjohnson: @N-Dekker THANKS!
Fixed to 
```c++
  for (unsigned int i = 0; i < InputImageDimension && nonzeroSizeCount < OutputImageDimension; ++i)
  {
    if (inputSize[i])
    {
      outputSize[nonzeroSizeCount] = inputSize[i];
      outputIndex[nonzeroSizeCount] = extractRegion.GetIndex()[i];
      ++nonzeroSizeCount;
    }
  }
```

Modules/Core/Common/test/itkShapedImageNeighborhoodRangeGTest.cxx
hjmjohnson: Done!

Modules/Core/Common/include/itkExtractImageFilter.hxx
N-Dekker: You're welcome, but the point I was trying to make was that the original code would throw an exception if nonzeroSizeCount != OutputImageDimension. Wasn't that the intended behavior when nonzeroSizeCount gets too big?

N-Dekker: So I would suggest, instead:
```cpp
  for (unsigned int i = 0; i < InputImageDimension; ++i)
  {
    if (inputSize[i])
    {
      if (nonzeroSizeCount < OutputImageDimension)
      {
        outputSize[nonzeroSizeCount] = inputSize[i];
        outputIndex[nonzeroSizeCount] = extractRegion.GetIndex()[i];
      }
      ++nonzeroSizeCount;
    }
  }
```
In order to get into the `if` clause of `if (nonzeroSizeCount != OutputImageDimension)`, to throw an exception, when nonzeroSizeCount gets too big:

https://github.com/InsightSoftwareConsortium/ITK/blob/9d1a93dfb8edb10bce000f9695f9e626b3ed2a26/Modules/Core/Common/include/itkExtractImageFilter.hxx#L86-L92

hjmjohnson: @N-Dekker I looked at this again, and your solution is exactly what is needed. 

Thank you for your careful inspection.  Update to PR coming soon.

Modules/Core/Common/test/itkShapedImageNeighborhoodRangeGTest.cxx
```diff
@@ -350,7 +350,7 @@ TEST(ShapedImageNeighborhoodRange, IteratorsCanBePassedToStdVectorConstructor)
 // second argument of std::reverse (which requires bidirectional iterators).
 TEST(ShapedImageNeighborhoodRange, IteratorsCanBePassedToStdReverseCopy)
 {
-  using PixelType = unsigned char;
+  using PixelType = unsigned short;
```
hjmjohnson, line 353: FYI:  Initially I instrumented this with gcc pragma commands to ignore the warning, but the problem seems to be a false positive warning due to assuming that the range is a string.   Instead of suppressing the false positive due to processsing unsigned char, just change the type to unsigned short for this simple test case.


I looked carefully at the code in a debugger, and it appears that the behavior is correct for these containers.

hjmjohnson:
This allows for clean building on the latest Ubuntu 24.04 release with GCC 13.2.

blowekamp (approved):
Thank you for addressing these warnings. A suggestion to consider, that would reduce code size.

Modules/Core/Common/include/itkNeighborhood.h
```diff
@@ -136,6 +136,11 @@ class ITK_TEMPLATE_EXPORT Neighborhood
   SizeValueType
   GetRadius(DimensionValueType n) const
   {
+    if (n >= VDimension)
+    {
+      itkExceptionMacro(<< " Can not get radius for dimension " << n << " greater than dimensionality of neighborhood "
+                        << VDimension);
+    }
     return m_Radius[n];
```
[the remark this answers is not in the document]
hjmjohnson: Are you suggesting 
```c++
return m_Radius.at(n);
```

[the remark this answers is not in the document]
blowekamp: Yes. Using the code below seems better than adding it to all these methods. 

https://github.com/InsightSoftwareConsortium/ITK/blob/f408b77a57095a44964187af840353809f3644b0/Modules/Core/Common/include/itkSize.h#L359
https://github.com/InsightSoftwareConsortium/ITK/blob/f408b77a57095a44964187af840353809f3644b0/Modules/Core/Common/include/itkSize.h#L410

thewtex (approved):
Thank you for addressing these @hjmjohnson !

hjmjohnson:
@thewtex   I assume we want 5.4.0 to build cleanly on Ubuntu 24.04.  This is a lot of small changes, but I think they are appropriate for 5.4.0.

I'll let you merge, or change the milestone.  If you change the milestone, please convert to a draft.

PR #4609, opened 2024-04-24
COMP: Update GDCM system version check for SetSecondaryCaptureImagePlaneModule

issakomi:
Follow up https://github.com/InsightSoftwareConsortium/ITK/pull/4601
GDCM master branch is currently at version **3.1.0**.
Updated check to include versions 3.1+.

P.S.
Maybe GDCM will change its approach in the future and not insist on `SecondaryCaptureImagePlaneModule=true` for _reading_ at all.  ImagePlaneModule for SC is a part of the standard now. But it is still there. The variable could be useful for _writing_ in the future.

### DIFF
--- a/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
+++ b/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
@@ -281,8 +281,9 @@ GDCMImageIO::Read(void * pointer)
 
   itkAssertInDebugAndIgnoreInReleaseMacro(gdcm::ImageHelper::GetForceRescaleInterceptSlope());
 // Only available in newer versions
-#if (!defined(ITK_USE_SYSTEM_GDCM) || \
-     ((GDCM_MAJOR_VERSION == 3 && GDCM_MINOR_VERSION == 0 && GDCM_BUILD_VERSION > 23) || GDCM_MAJOR_VERSION > 3))
+#if (!defined(ITK_USE_SYSTEM_GDCM) ||                                                    \
+     ((GDCM_MAJOR_VERSION == 3 && GDCM_MINOR_VERSION == 0 && GDCM_BUILD_VERSION > 23) || \
+      (GDCM_MAJOR_VERSION == 3 && GDCM_MINOR_VERSION > 0) || GDCM_MAJOR_VERSION > 3))
   // Secondary capture image orientation patient and image position patient support
   itkAssertInDebugAndIgnoreInReleaseMacro(gdcm::ImageHelper::GetSecondaryCaptureImagePlaneModule());
 #endif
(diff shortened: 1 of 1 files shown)


Modules/IO/GDCM/src/itkGDCMImageIO.cxx
```diff
@@ -281,8 +281,9 @@ GDCMImageIO::Read(void * pointer)
 
   itkAssertInDebugAndIgnoreInReleaseMacro(gdcm::ImageHelper::GetForceRescaleInterceptSlope());
 // Only available in newer versions
-#if (!defined(ITK_USE_SYSTEM_GDCM) || \
-     ((GDCM_MAJOR_VERSION == 3 && GDCM_MINOR_VERSION == 0 && GDCM_BUILD_VERSION > 23) || GDCM_MAJOR_VERSION > 3))
+#if (!defined(ITK_USE_SYSTEM_GDCM) ||                                                    \
+     ((GDCM_MAJOR_VERSION == 3 && GDCM_MINOR_VERSION == 0 && GDCM_BUILD_VERSION > 23) || \
+      (GDCM_MAJOR_VERSION == 3 && GDCM_MINOR_VERSION >= 1) || GDCM_MAJOR_VERSION > 3))
```
N-Dekker, line 286: I think it would be slightly more elegant to do `> 0`  instead of `>= 1`, for the sake of consistency. No show-stopper, of course!

Modules/IO/GDCM/src/itkGDCMImageIO.cxx
```diff
@@ -455,8 +456,9 @@ GDCMImageIO::InternalReadImageInformation()
   // In general this should be relatively safe to assume
   gdcm::ImageHelper::SetForceRescaleInterceptSlope(true);
 // Only available in newer versions
-#if (!defined(ITK_USE_SYSTEM_GDCM) || \
-     ((GDCM_MAJOR_VERSION == 3 && GDCM_MINOR_VERSION == 0 && GDCM_BUILD_VERSION > 23) || GDCM_MAJOR_VERSION > 3))
+#if (!defined(ITK_USE_SYSTEM_GDCM) ||                                                    \
+     ((GDCM_MAJOR_VERSION == 3 && GDCM_MINOR_VERSION == 0 && GDCM_BUILD_VERSION > 23) || \
+      (GDCM_MAJOR_VERSION == 3 && GDCM_MINOR_VERSION >= 1) || GDCM_MAJOR_VERSION > 3))
```
N-Dekker, line 461: Same here: `> 0` appears slightly more elegant, but that's no show-stopper.

N-Dekker (approved):
Thanks @issakomi !  Approved 👍

Modules/IO/GDCM/src/itkGDCMImageIO.cxx
issakomi: Updated to `> 0`

PR #4610, opened 2024-04-24
BUG: AuthorsChangesSince.py support both single and multiline

thewtex:
Versions can be available in the *.remote.cmake files of the form:

  GIT_TAG abc123

and

  GIT_TAG
  abc123)

in the history. Support both.


hjmjohnson (approved):
This looks like a reasonable improvement to the utility script.

PR #4617, opened 2024-04-26
ENH: Optionally limit an ImageMask to a specific pixel value

aylward:
Nominally an ImageMaskSpatialObject considers any non-zero pixel to indicate that the object is present at that point.

With this change, it is possible to specify a MaskValue and to enable the use of that MaskValue (via the UseMaskValue functions) so that a specific pixel value is needed to indicate the presence of that object.

This is particularly useful when a single mask image defines multiple
objects via different values.   That single image can be assigned
to multiple ImageMaskSpatialObjects, each using a different MaskValue.  This reduces memory requirements, etc.

A new set of tests were added to confirm proper function.

Default behaviour is unchanged.

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkImageMaskSpatialObject.h
+++ b/Modules/Core/SpatialObjects/include/itkImageMaskSpatialObject.h
@@ -33,9 +33,10 @@ namespace itk
  * Image Registration Metrics.
  *
  * \note The bounding box of an image mask is defined in such a way that
- * any point whose nearest pixel has a non-zero value is inside the
- * bounding box. When all the pixels of an image are zero, the bounding box
- * of the image mask is empty, and its bounds are all zero.
+ * any point whose nearest pixel has a non-zero value (or matches the
+ * specified mask value) is inside the bounding box. When all the pixels of
+ * an image are zero, the bounding box of the image mask is empty, and
+ * its bounds are all zero.
  *
  * \sa ImageSpatialObject SpatialObject CompositeSpatialObject
  * \ingroup ITKSpatialObjects
@@ -74,13 +75,25 @@ public:
 
   /** Test whether a point is inside the object.
    *
-   * A point is inside the image mask when the value of its nearest pixel is non-zero.
+   * A point is inside the image mask when the value of its nearest pixel is
+   * non-zero or matches the Mask_Value value.
    *
    * For computational speed purposes, it is faster if the method does not  check the name of the class and the
    * current depth.
    */
   bool
   IsInsideInObjectSpace(const PointType & point) const override;
+
+  /** Specify the value in the mask image that defines the object.
+   * You must also call SetUseMaskValue(true) to enable the use
+   * of the mask value, otherwise the value is ignored and any
+   * non-zero value in the mask is used to define the object. */
+  itkSetMacro(MaskValue, PixelType);
+  itkGetConstReferenceMacro(MaskValue, PixelType);
+
+  itkBooleanMacro(UseMaskValue);
+  itkSetMacro(UseMaskValue, bool);
+  itkGetConstReferenceMacro(UseMaskValue, bool);
 
   /* Avoid hiding the overload that supports depth and name arguments */
   using Superclass::IsInsideInObjectSpace;
@@ -118,6 +131,10 @@ protected:
 
   typename LightObject::Pointer
   InternalClone() const override;
+
+private:
+  bool      m_UseMaskValue{};
+  PixelType m_MaskValue{};
 };
 } // end of namespace itk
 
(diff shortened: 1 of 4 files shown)


Modules/Core/SpatialObjects/include/itkImageMaskSpatialObject.hxx
```diff
@@ -29,6 +30,9 @@ template <unsigned int TDimension, typename TPixel>
 ImageMaskSpatialObject<TDimension, TPixel>::ImageMaskSpatialObject()
 {
   this->SetTypeName("ImageMaskSpatialObject");
+
+  m_UseMaskValue = false;
```
hjmjohnson, line 34: These assignments are not needed here. Initialize at the time of definition.

Modules/Core/SpatialObjects/include/itkImageMaskSpatialObject.h
```diff
@@ -118,6 +131,10 @@ class ITK_TEMPLATE_EXPORT ImageMaskSpatialObject : public ImageSpatialObject<TDi
 
   typename LightObject::Pointer
   InternalClone() const override;
+
+private:
+  bool      m_UseMaskValue{};
+  PixelType m_MaskValue{};
```
hjmjohnson, line 137: bool      m_UseMaskValue{false};
  PixelType m_MaskValue{NumericTraits<PixelType>::OneValue()};

Modules/Core/SpatialObjects/include/itkImageMaskSpatialObject.hxx
```diff
@@ -39,6 +43,12 @@ ImageMaskSpatialObject<TDimension, TPixel>::IsInsideInObjectSpace(const PointTyp
 
   const IndexType index = image->TransformPhysicalPointToIndex(point);
 
+  if (this->m_UseMaskValue == true)
+  {
+    return Superclass::GetBufferedRegion().IsInside(index) &&
+           Math::ExactlyEquals(image->GetPixel(index), this->m_MaskValue);
+  }
+
   return Superclass::GetBufferedRegion().IsInside(index) && Math::NotExactlyEquals(image->GetPixel(index), PixelType{});
```
hjmjohnson, line 52: Avoid unnecessary computations when outside the image, or m_useMaskValue == false.  Utilize early termination on false conditions in compound conditional.


```c++
constexpr auto background_indicator_zero = PixelType{};

const bool is_inside = Superclass::GetBufferedRegion().IsInside(index);
return  is_inside && 
  (
  (this->m_useMaskValue == false) &&  Math::NotExactlyEquals(image->GetPixel(index),background_indicator_zero)
) && 
Math::ExactlyEquals(image->GetPixel(index), this->m_MaskValue)
;
```

Modules/Core/SpatialObjects/include/itkImageMaskSpatialObject.hxx
```diff
@@ -123,14 +133,19 @@ ImageMaskSpatialObject<TDimension, TPixel>::InternalClone() const
...
 }
 
 template <unsigned int TDimension, typename TPixel>
 auto
 ImageMaskSpatialObject<TDimension, TPixel>::ComputeMyBoundingBoxInIndexSpace() const -> RegionType
 {
-  const ImagePointer imagePointer = this->GetImage();
+  ImagePointer imagePointer = this->GetImage();
+  bool         useMaskValue = this->GetUseMaskValue();
+  PixelType    maskValue = this->GetMaskValue();
```
hjmjohnson, line 149:
```c++
  const bool         useMaskValue = this->GetUseMaskValue();
  const PixelType    maskValue = this->GetMaskValue();
```

hjmjohnson (changes requested):
Recommendations to improve the readability and performance

Modules/Core/SpatialObjects/include/itkImageMaskSpatialObject.h
aylward: Done.

Minor tweak...
PixelType m_MaskValue{NumbericTraits<PixelType>::OneValue()};

Modules/Core/SpatialObjects/include/itkImageMaskSpatialObject.hxx
aylward: Done. Added const to all.

Modules/Core/SpatialObjects/include/itkImageMaskSpatialObject.hxx
aylward: Thanks!   Had to tweak the logic a bit...

  return is_inside && ((m_UseMaskValue == true && 
      Math::ExactlyEquals(image->GetPixel(index), this->m_MaskValue))
      || (m_UseMaskValue == false &&
      Math::NotExactlyEquals(image->GetPixel(index), background_zero)));

If you can think of a faster way, let me know.

thewtex (approved):
:cherry_blossom:

hjmjohnson (approved):

PR #4622, opened 2024-04-29
COMP: Add type aliases of ptrdiff_t and size_t to the `itk` namespace

N-Dekker:
Officially, we may not assume that `ptrdiff_t` and `size_t` are defined in the global namespace.

----

Motivation: Both `ptrdiff_t` and `size_t` are already commonly being used inside the `itk` namespace, without specifying their `std` namespace:

https://github.com/InsightSoftwareConsortium/ITK/blob/487da6d086b341808147c71e8dd1eb1f57371ea6/Modules/IO/ImageBase/include/itkConvertPixelBuffer.h#L64

https://github.com/InsightSoftwareConsortium/ITK/blob/487da6d086b341808147c71e8dd1eb1f57371ea6/Modules/Video/IO/src/itkFileListVideoIO.cxx#L64

It just compiles, so far! However, compilers may become more picky, as those types are not guaranteed to remain available in the global namespace.

### DIFF
--- a/Modules/Core/Common/include/itkIntTypes.h
+++ b/Modules/Core/Common/include/itkIntTypes.h
@@ -58,6 +58,9 @@ using std::uintmax_t;
 using std::intptr_t;
 using std::uintptr_t;
 
+using std::size_t;
+using std::ptrdiff_t;
+
 
 #if defined(ITK_USE_64BITS_IDS) && ((ULLONG_MAX != ULONG_MAX) || (LLONG_MAX != LONG_MAX))
 


dzenanz (approved):

hjmjohnson (commented):
This seems like a good solution to being backward compatible.

N-Dekker:
@thewtex FYI, I think this pull request can be merged safely for ITK 5.4. On the other hand, it isn't very urgent to me, _personally_. It _might_ _become_ urgent, once compilers start to complain about using `size_t` without specifying the namespace `std`.

dzenanz:
This looks like a low-risk addition, and merging it sooner rather than later is better in my opinion.

thewtex (approved):
@N-Dekker thanks for the patch :+1: This does seem to be low risk; however, it is also possible to cause issues since there may be naming conflicts. Since this is not addressing a known issue, it should be merged after v5.4.0.

N-Dekker:
> This does seem to be low risk; however, it is also possible to cause issues since there may be naming conflicts. Since this is not addressing a known issue, it should be merged after v5.4.0.

@thewtex OK, no problem. Marked "draft"to avoid accidentally being merged before the release of v5.4.0. 👍

N-Dekker:
I think this pull request may also be merged now, as v5.4.0 has been tagged: https://github.com/InsightSoftwareConsortium/ITK/pull/4603#issuecomment-2120498131

PR #4626, opened 2024-04-29
COMP: Use default copy, move for construc, assign `RealTimeStamp`

jhlegarreta:
Use the compiler-proved default implementations for `itk::RealTimeStamp` copy constructor, copy assignment, move constructor, and move assignment functions.

As noted in [1], the C++ standard deprecated the implicit generation of copy and assignment operators.

Fixes:
```
[CTest: warning matched] /Users/builder/externalModules/Core/Common/include/itkRealTimeStamp.h:56:3:
 warning: definition of implicit copy assignment operator for 'RealTimeStamp' is deprecated because it has a user-declared destructor [-Wdeprecated]
  ~RealTimeStamp();
  ^
[CTest: warning matched] /Users/builder/externalModules/Core/Common/include/itkDataObject.h:452:3:
 note: in implicit copy assignment operator for 'itk::RealTimeStamp' first required here
  itkSetMacro(RealTimeStamp, RealTimeStamp);
  ^
[CTest: warning matched] /Users/builder/externalModules/Core/Common/include/itkMacro.h:992:22:
 note: expanded from macro 'itkSetMacro'
      this->m_##name = std::move(_arg);                        \
                     ^
```

And other similar warnings stemming from `itk::RealTimeStamp` that have been appearing consistently in some macOS site builds in the dashboard: https://open.cdash.org/viewBuildError.php?type=1&buildid=9579479

[1] https://learn.microsoft.com/bs-latn-ba/cpp/error-messages/compiler-warnings/c5267?view=msvc-150#remarks

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkRealTimeStamp.h
+++ b/Modules/Core/Common/include/itkRealTimeStamp.h
@@ -55,6 +55,13 @@ public:
   /** Destructor */
   ~RealTimeStamp();
 
+  RealTimeStamp(const RealTimeStamp &) = default;
+  RealTimeStamp &
+  operator=(const RealTimeStamp &) = default;
+  RealTimeStamp(RealTimeStamp &&) = default;
+  RealTimeStamp &
+  operator=(RealTimeStamp &&) = default;
+
   /** Native type used to represent the time in different time units. */
   using TimeRepresentationType = RealTimeInterval::TimeRepresentationType;
 


dzenanz:
Perhaps the constructor needs to be removed and explicitly or implicitly defaulted?

N-Dekker (approved):
> the C++ standard deprecated the implicit generation of copy and assignment operators.

Just for the sake of completeness, _in general_, the "rule of zero" is still allowed (even though it assumes implicit generation of copy and assignment operators). But when following the rule of zero, the destructor should also be "implicit".

Thanks Jon, approved  👍

Modules/Core/Common/include/itkRealTimeStamp.h
```diff
@@ -55,6 +55,13 @@ class ITKCommon_EXPORT RealTimeStamp
   /** Destructor */
   ~RealTimeStamp();
 
+  RealTimeStamp(const RealTimeStamp &) = default;
```
dzenanz, line 58: The default constructor can be removed too, as [these assignments](https://github.com/InsightSoftwareConsortium/ITK/blob/f408b77a57095a44964187af840353809f3644b0/Modules/Core/Common/src/itkRealTimeStamp.cxx#L71-L72) can be made at the point of declaration. Then we could have rule of zero, right?

jhlegarreta: Not sure, what about this
https://github.com/InsightSoftwareConsortium/ITK/blob/f408b77a57095a44964187af840353809f3644b0/Modules/Core/Common/src/itkRealTimeStamp.cxx#L80?

dzenanz: I guess we can't have any constructors if we want "rule of zero". Not a show stopper, but we could default the default constructor too.

thewtex (approved):

Modules/Core/Common/include/itkRealTimeStamp.h
N-Dekker: @dzenanz Indeed. I believe the warnings might also be fixed by just removing the user-defined destructor, `~RealTimeStamp()`. Optionally, the default constructor might indeed be defaulted by `=default`. 

Doing so would not change the API, but it would be a very small _ABI_ change (regarding the binary interface), though. Because it would make those member functions _inline_  (header-only), while they are now still defined in the CXX file. No big deal to me, just something to take into account.

The current version of the PR does not change the API, it does not even change the _ABI_. As far as i can see  😃

jhlegarreta: Thanks for the comments. Leaving rule of zero for a separate PR when a new major ITK version will be considered. Resolving. Ready to be merged.

PR #4627, opened 2024-04-29
COMP: Use default copy and move construct and assign in `itk::Region`

jhlegarreta:
Use the compiler-proved default implementations for `itk::Region` copy constructor, copy assignment, move constructor, and move assignment functions.

As noted in [1], the C++ standard deprecated the implicit generation of copy and assignment operators.

Fixes:
```
[CTest: warning matched] /Users/builder/externalModules/Core/Common/include/itkRegion.h:91:11:
 warning: definition of implicit copy assignment operator for 'Region' is deprecated because it has a user-declared destructor [-Wdeprecated]
  virtual ~Region() = default;
          ^
[CTest: warning matched] /Users/builder/externalModules/Core/Common/include/itkImageIORegion.h:117:30:
 note: in implicit copy assignment operator for 'itk::Region' first required here
  operator=(Self &&) = default;
                             ^
[CTest: warning matched] /Users/builder/externalModules/IO/ImageBase/include/itkImageIOBase.h:228:3:
 note: in implicit move assignment operator for 'itk::ImageIORegion' first required here
  itkSetMacro(IORegion, ImageIORegion);
  ^
[CTest: warning matched] /Users/builder/externalModules/Core/Common/include/itkMacro.h:992:22:
 note: expanded from macro 'itkSetMacro'
      this->m_##name = std::move(_arg);                        \
                     ^
```

And other similar warnings stemming from `itk::Region` that have been appearing consistently in some macOS site builds in the dashboard: https://open.cdash.org/viewBuildError.php?type=1&buildid=9579479

[1] https://learn.microsoft.com/bs-latn-ba/cpp/error-messages/compiler-warnings/c5267?view=msvc-150#remarks

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkRegion.h
+++ b/Modules/Core/Common/include/itkRegion.h
@@ -90,6 +90,13 @@ public:
   Region() = default;
   virtual ~Region() = default;
 
+  Region(const Region &) = default;
+  Region &
+  operator=(const Region &) = default;
+  Region(Region &&) = default;
+  Region &
+  operator=(Region &&) = default;
+
 protected:
   /** Methods invoked by Print() to print information about the object
    * including superclasses. Typically not called by the user (use Print()


jhlegarreta:
@seanm Hopefully this will remove some of the 199 warnings on RogueResearch22.

seanm:
Cool, thanks!  I defer to C++ experts on the correctness of this change, but seems reasonable to me!

N-Dekker:
I'm sorry I think `Region` should remain copyable, as long as it is being used as base class of `ImageRegion` 🤷 

- For the record, in the future,  `ImageRegion` hopefully no longer inherits from `Region`: pull request #4344

jhlegarreta:
https://github.com/InsightSoftwareConsortium/ITK/pull/4627#issuecomment-2085575522 :+1: Done. Thanks Niels.

jhlegarreta:
Builds are failing. Run out of ideas.

N-Dekker (approved):
Does it indeed fix the warnings? I guess so, but you never know... anyway, approved, of course. 👍 Thanks Jon!

Modules/Core/Common/include/itkRegion.h
```diff
@@ -90,6 +90,13 @@ class ITKCommon_EXPORT Region
   Region() = default;
   virtual ~Region() = default;
 
+  Region(const Region &) = default;
```
dzenanz, line 93: Would "rule of zero" work here, but removing all of them?

thewtex (approved):

Modules/Core/Common/include/itkRegion.h
N-Dekker: @dzenanz  Good question... technically, the rule of zero would probably work as well, for `Region` but then we would also have to remove `~Region()`, which is now declared `virtual`. And you know, `Region` is still being used as base class. And a base class without virtual destructor, while still having (other) virtual functions looks like a code smell... at least it might cause compiler warnings  😸

N-Dekker: So the short answer is: no, I don't think "rule of zero" works for `itk::Region`. 🤷  The current PR should work fine, though 👍 And it's not even an API change, because the added (defaulted) member functions behave exactly as in the original code.

jhlegarreta: Thanks for the comments. Resolving. Ready to be merged.

PR #4628, opened 2024-04-30
STYLE: Replace SetSize/Index calls in tests with `region{ index, size }`

N-Dekker:
Replaced code like:

    RegionType region;
    region.SetSize(size);
    region.SetIndex(index);

With the shorter equivalent, `RegionType region{ index, size };`

    Find what: ^( [ ]+)(\w.*Region[^ ]*)([ ]+)(\w+);\r\n\1\4\.SetSize\((\w+)\);\r\n\1\4\.SetIndex\((\w+)\);
    Replace with: $1$2$3$4{ $6, $5 };
    Find what: ^( [ ]+)(\w.*Region[^ ]*)([ ]+)(\w+);\r\n\1\4\.SetIndex\((\w+)\);\r\n\1\4\.SetSize\((\w+)\);
    Replace with: $1$2$3$4{ $5, $6 };
    Filters: itk*Test*.cxx
    [v] Match case
    (*) Regular expression

### DIFF
--- a/Modules/Core/Common/test/itkAbortProcessObjectTest.cxx
+++ b/Modules/Core/Common/test/itkAbortProcessObjectTest.cxx
@@ -66,9 +66,7 @@ itkAbortProcessObjectTest(int, char *[])
   // fill in an image
   const ShortImage::IndexType index = { { 0, 0 } };
   const ShortImage::SizeType  size = { { 100, 100 } };
-  ShortImage::RegionType      region;
-  region.SetSize(size);
-  region.SetIndex(index);
+  ShortImage::RegionType      region{ index, size };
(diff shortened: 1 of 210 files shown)


dzenanz (approved):

dzenanz (approved):

N-Dekker:
@thewtex This pull request is only a style improvement of the unit tests. So I think it won't harm to have it merged with ITK 5.4. I like this PR very much, but it's not very urgent, of course. What do you think?  🤷

thewtex (approved):
@N-Dekker thanks!

Yes, in this case given the nature of the changes and since CI is passing, this should be safe for v5.4.0.

PR #4630, opened 2024-04-30
BUG: Mark LabelGeometryImageFilter as deprecated

blowekamp:
The LabelGeometryImageFilter has known computational inefficiencies and bugs such as some attributes are not computed with respect to image geometry.

This class is currently in the Review module and is planned to be moved to the Compatibility group under the Deprecated module.

## PR Checklist

### DIFF
--- a/Modules/Nonunit/Review/include/itkLabelGeometryImageFilter.h
+++ b/Modules/Nonunit/Review/include/itkLabelGeometryImageFilter.h
@@ -66,6 +66,8 @@ namespace itk
  *  by Padfield D., Miller J
  *  https://www.insight-journal.org/browse/publication/301
  *
+ * This class contains computational inefficiencies and bugs such as some attributes are not computed with respect to
+ * image geometry, consider using these supported alternatives:
  * \sa LabelStatisticsImageFilter
  * \sa LabelImageToShapeLabelMapFilter
  *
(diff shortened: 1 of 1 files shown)


blowekamp:
The intention is to produce a warning for the current up common release so that with 5.0 the class can be moved to the deprecated module.

dzenanz (approved):
Looks good. You mean to be moved to `Deprecated` module in 6.0, not 5.0?

dzenanz:
Maybe because it is off by default?

dzenanz:
Perhaps we should turn it on by default (low effort), of finish migration of content out of it and into other corresponding modules (high effort)?

blowekamp:
> Perhaps we should turn it on by default (low effort), of finish migration of content out of it and into other corresponding modules (high effort)?

I think turning it on by default may encourage more usage, and limit options available to the remaining classes. I Think the options for the classes in the Review module should be 1) Integrate in ITK proper after changes, 2) make a remote module 3) move to the Deprecate module.

Modules/Nonunit/Review/include/itkLabelGeometryImageFilter.h
```diff
@@ -66,6 +66,8 @@ namespace itk
  *  by Padfield D., Miller J
  *  https://www.insight-journal.org/browse/publication/301
  *
+ * This class contains computational inefficiencies and bugs such as some attributes are not computed with respect to
+ * image geometry, consider using these supported alternatives:
```
thewtex, line 70: :+1:

thewtex:
> This filter is only in ITKReview, so perhaps less notification is needed?

ITKReview is relied on in practice, but it would be helpful to have it marked as deprecated if you would like to move to ITKDeprecated in ITK 6. Please go ahead and merge.

Issue #4267, opened 2023-10-23
Feature Request: Make empty `itk.Image` and related data serializable (pickleable) for use with `dask.distributed`


tbirdso:
### Description

ITK supports Python's pickle serialization technique for fully buffered `itk.Image` objects (see https://github.com/InsightSoftwareConsortium/ITK/pull/2829), but not for advanced use cases of `itk.Image`.

It would be good to support the following cases:
1. Unbuffered `itk.Image` describing a relationship between an oriented spatial bounding box and an implied underlying voxel array. I.e., `image.GetBufferedRegion()` is zero and `image.GetLargestPossibleRegion()` is greater than zero.
2. Partially buffered `itk.Image` describing a subimage. I.e., `image.GetBufferedRegion()` is greater than zero and is strictly less than `image.GetLargestPossibleRegion()`.
3. Directly support serialization of related `itk.Image` header objects such as `itk.ImageRegion` and `itk.Matrix`.

### Steps to Reproduce

For case 1:

```py
import itk
import pickle

reader = itk.ImageFileReader[itk.Image[itk.F,3]].New()
reader.SetFileName('path/to/image.mha')
reader.UpdateOutputInformation() # update without buffering

result = pickle.dumps(reader.GetOutput())
```

### Expected behavior

Image is serialized to a bytestring and can be deserialized with `pickle.loads(result)` with no loss of information

### Actual behavior

Serialization fails because there is not a voxel buffer to convert to NumPy:

```
>>> pickle.dumps(im2)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\venvs\venv-itk\lib\site-packages\itk\itkImagePython.py", line 7297, in __getstate__
    state = itk.dict_from_image(self)
  File "C:\venvs\venv-itk\lib\site-packages\itk\support\extras.py", line 791, in dict_from_image
    pixel_arr = itk.array_from_image(image)
  File "C:\venvs\venv-itk\lib\site-packages\itk\support\extras.py", line 335, in GetArrayFromImage
    return _GetArrayFromImage(
  File "C:\venvs\venv-itk\lib\site-packages\itk\support\extras.py", line 317, in _GetArrayFromImage
    return templatedFunction(img, keep_axes, update)
  File "C:\venvs\venv-itk\lib\site-packages\itk\itkPyBufferPython.py", line 3839, in GetArrayFromImage
[...]

PR #4633, opened 2024-05-01
BUG: Add bufferedRegion to dict_from_image representation

thewtex:
Optional specification of the BufferedRegion, which is a subset of the
LargestPossibleRegion. Addresses #4267.

The LargestPossibleRegion's Index is assumed to be trivial (zeros).

Also update the NumPy bridge to correctly use the BufferedRegion instead
of the LargestPossibleRegion for the array shape.


dzenanz (approved):
Looks good on a glance.

Issue #4629, opened 2024-04-30
Detailed CMake options are not marked as advanced.


blowekamp:
### Description

The following CMake options/varaibles are not marked as advanced.
                                 
 - [x] COREFOUNDATION_LIBRARY          */Applications/Xcode_14.2.app/Contents/Develop
 - [x] GDCM_USE_COREFOUNDATION_LIBRAR  *ON                                           
  - [x] Module_FastBilateral            *OFF                                          
 - [x] OPJ_USE_DSYMUTIL                *OFF  

### Steps to Reproduce

The above where included in the `ccmake` GUI on Mac OS X initial configuration

### Expected behavior

These options seem like they should be marked advanced.

thewtex:
xref: https://github.com/malaterre/GDCM/pull/172

PR #4636, opened 2024-05-02
COMP: Fix for gcc13.2 compiler test failures

hjmjohnson:
Three tests fail in ITK for Release and RelWithDebInfo builds on Ubuntu 24.04 with GCC 13.2:
  itkGDCMLegacyMultiFrameTest (Failed)
  itkGDCMImageReadWriteTest_MultiFrameMRIZSpacing (Failed)
  itkGDCM_ComplianceTest_singlebit (Failed)

(Note: Debug builds do not fail these tests)

The initial review of the code did not indicate a problem, but it looks like perhaps an over-optimization that removes variables and causes the incorrect behavior to occur.

The function "FindDataElement" had duplicate code from "GetDataElement" so made "FindDataElement" use the "GetDataElement" working function.

## PR Checklist

### DIFF
--- a/Modules/ThirdParty/GDCM/src/gdcm/Source/DataStructureAndEncodingDefinition/gdcmDataSet.h
+++ b/Modules/ThirdParty/GDCM/src/gdcm/Source/DataStructureAndEncodingDefinition/gdcmDataSet.h
@@ -209,13 +209,9 @@ public:
 
   // DUMB: this only search within the level of the current DataSet
   bool FindDataElement(const Tag &t) const {
-    const DataElement r(t);
-    //ConstIterator it = DES.find(r);
-    if( DES.find(r) != DES.end() )
-      {
-      return true;
-      }
-    return false;
+    const auto it = GetDataElement(t);
+    // Return if tag is found
+    return it != GetDEEnd();
     }
 
   // WARNING:


hjmjohnson:
@malaterre   I don't have proof, but I feel that this is a compiler optimization bug where the logic of the function is incorrectly compiled away.  In Release and RelWithDebInfo compilations the function always returns false and does not seem to perform the DES.find() function.  

Given the commented out line that explicitly assigned an iterator, it feels like perhaps this has been a problem in the past as well.

My workaround allows the test to pass reliably in all compilation modes and seems like a reasonable approach to solving the same problem. 

If you agree that this is an acceptable approach, I'll work on making and upstream GDCM patch.

thewtex (approved):
LGTM

hjmjohnson:
@thewtex I don't know how to deal with the clang-format failure.  It would require a lot of changes to that file that are unrelated.

thewtex:
@hjmjohnson we should ignore the clang-format warning for this PR, editing the GDCM sources. After the v5.4.0 I will make improvements to the linting infrastructure. Ideally, we get this into upstream GDCM before merging, as you suggested, but we can merge as-is if there is insufficient time before the release.

hjmjohnson:
https://github.com/malaterre/GDCM/pull/173

hjmjohnson:
@thewtex   @malaterre  merged upstream.  https://github.com/malaterre/GDCM/pull/173.  

Matt,  I'll let you determine the best approach.  

1). add this patch now, and then update GDCM after 5.4.0
2) update GDCM now, close this PR without merging.

Modules/ThirdParty/GDCM/src/gdcm/Source/DataStructureAndEncodingDefinition/gdcmDataSet.h
```diff
@@ -209,13 +209,9 @@ class GDCM_EXPORT DataSet
...
   // DUMB: this only search within the level of the current DataSet
   bool FindDataElement(const Tag &t) const {
-    const DataElement r(t);
-    //ConstIterator it = DES.find(r);
-    if( DES.find(r) != DES.end() )
-      {
-      return true;
-      }
-    return false;
+    const auto it = GetDataElement(t);
```
issakomi, line 212: BTW,  this will highly likely trigger _copy_ of the DataElement (the function returns `const DataElement&`, but this `const auto =` forces copy of the object, IMO). It is not a request to change, the issue is very odd. cc @N-Dekker

thewtex:
@hjmjohnson @malaterre awesome, thank you for getting these issues resolved!

We should update GDCM from upstream in ITK. We want to:

- Pull in these improvements.
- Have our test suite run against the updates in GDCM re: spacing support, ensuring the same experience with vendored GDCM and system GDCM.
- Pull in CMake fixes re: #4629 (there may more needed)

@hjmjohnson if you have the cycles and interest, please go ahead. Otherwise, I will start the update Monday.

thewtex:
Integrated in https://github.com/InsightSoftwareConsortium/ITK/pull/4647

Modules/ThirdParty/GDCM/src/gdcm/Source/DataStructureAndEncodingDefinition/gdcmDataSet.h
N-Dekker: @issakomi I just had a few days off, sorry for my late response! I'm not very familiar with this code. I would expect that it would search in the DataElementSet for a DataElement with the specified tag. But instead, it locally creates a DataElement with the specified tag, using default values for its VL and VR. It then tries to find this local DataElement in the DataElementSet. So _I think_ it can only find a DataElement with those default values for VL and VR. Is that OK? 

Anyway, I agree with you that a "copy" might be skipped by doing `return (GetDataElement(t) != GetDEEnd())`.

issakomi: I have done some test (Linux, GCC 12).  These _GetDataElement_ and _FindDataElement_ rely on _DES.find_. DES is `std::set<DataElement>`. Interestingly DES.find doesn't seem to use DataElement's [operator==](https://github.com/InsightSoftwareConsortium/ITK/blob/28de86169f65ad8a9800b6cadc535e7f54c65a55/Modules/ThirdParty/GDCM/src/gdcm/Source/DataStructureAndEncodingDefinition/gdcmDataElement.h#L172-L187). The function works (both old and new variants) with GCC 12, but I still don't fully understand ...

issakomi: It seems that comparing for _find_ is something like `!(x < y) && !(y < x)`
without `operator==`, only with `operator<`. The last one uses `Tag`. But what was the issue with GCC 13 is still not clear.

N-Dekker: @issakomi Thanks for this little research! 👍

issakomi: Interesting that GDCM almost always does something like
```
  if( !subds2.FindDataElement(tps) ) return false;
  const DataElement &de = subds2.GetDataElement( tps );
```
FindDataElement does in fact 1st GetDataElement and 2nd compares returned one with invalid DataElement(Tag(0xffff, 0xffff)). So far 1 _find_ too much (and it is not cheap: "Complexity: logarithmic in the size of the container")
It could be e.g. or something
```
  const DataElement &de = subds2.GetDataElement( tps );
  if (de == DataElement(Tag(0xffff, 0xffff)) return false;
```
There are dozens or hundreds of such double _find_ s in e.g. gdcmImageHelper. Sometimes 3 times, e.g. in [itkGDCMImageIO](https://github.com/InsightSoftwareConsortium/ITK/blob/28de86169f65ad8a9800b6cadc535e7f54c65a55/Modules/IO/GDCM/src/itkGDCMImageIO.cxx#L650-L652)
Just FYI. I will not start PRs for this.

PR #4638, opened 2024-05-02
BUG: Support passing a single transform to transform write

thewtex:
This is a common use case. Without this patch, a segfault results.

### DIFF
--- a/Wrapping/Generators/Python/itk/support/extras.py
+++ b/Wrapping/Generators/Python/itk/support/extras.py
@@ -1453,7 +1453,7 @@ def transformread(filename: fileiotype) -> List["itkt.TransformBase"]:
 
 
 def transformwrite(
-    transforms: List["itkt.TransformBase"],
+    transforms: Union["itkt.TransformBase", List["itkt.TransformBase"]],
     filename: fileiotype,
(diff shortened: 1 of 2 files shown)


hjmjohnson (approved):
YES!  Thank you.  The old behavior is annoying when writing transforms in python.

Wrapping/Generators/Python/itk/support/extras.py
```diff
@@ -1462,8 +1462,8 @@ def transformwrite(
     Parameters
     ----------
 
-    transforms: list of itk.TransformBaseTemplate[itk.D]
-        Python list of the transforms to write.
+    transforms: single transform or list of transforms
+        Single transform Python list of the itk.TransformBaseTemplate[itk.D]'s to write.
```
dzenanz, line 1466: Nitpick: missing "or"

dzenanz (approved):

Wrapping/Generators/Python/itk/support/extras.py
thewtex: 👍 fixed

dzenanz (approved):

PR #4642, opened 2024-05-03
BUG: Fix path to write itkNiftiImageIOTest14 data

cookpa:
Fixes a test failure introduced by #4595 

@jhlegarreta, this works on my machine, but then so did the previous test. According to the error message here

https://open.cdash.org/tests/1540861560

```
** ERROR (nifti_image_write_engine): cannot open output file '/a_blatantly_obvious/bad_file_path/that/should/never/exist/on/the/computer/xyzt_units_output_check.nii'
Correctly caught exception for attempting to write to an invalid file.
Caught an exception: 

itk::ImageFileReaderException (0x6000037a4620)
Location: "unknown" 
File: /Users/runner/work/ITK/ITK/Modules/IO/ImageBase/include/itkImageFileReader.hxx
Line: 135
Description:  Could not create IO object for reading file xyzt_units_output_check.nii
The file doesn't exist. 
Filename = xyzt_units_output_check.nii


 /Users/runner/work/ITK/ITK/Modules/IO/ImageBase/include/itkIOTestHelper.h 54
Exception caught while reading image back in
```

It looks like a previous test somehow set the working directory to `/a_blatantly_obvious/bad_file_path/that/should/never/exist/on/the/computer/`. But I could not locate where this might have happened.

My mistake was to not include the test directory (passed to the test as `${ITK_TEST_OUTPUT_DIR}`) in my output file name. I fixed the test to write to the full path.

## PR Checklist

### DIFF
--- a/Modules/IO/NIFTI/test/itkNiftiImageIOTest14.cxx
+++ b/Modules/IO/NIFTI/test/itkNiftiImageIOTest14.cxx
@@ -90,7 +90,7 @@ itkNiftiImageIOTest14(int argc, char * argv[])
     }
   }
 
-  const char * output_test_fn = "xyzt_units_output_check.nii";
+  const std::string output_test_fn = std::string(test_dir) + "/xyzt_units_output_check.nii";
 
   // set the origin of time to a non-zero value
   auto newOrigin = ImageType::PointType();
@@ -100,15 +100,16 @@ itkNiftiImageIOTest14(int argc, char * argv[])
   newOrigin[3] = 2.0;
   test_image->SetOrigin(newOrigin);
 
-  itk::IOTestHelper::WriteImage<ImageType, itk::NiftiImageIO>(test_image, std::string(output_test_fn));
+  itk::IOTestHelper::WriteImage<ImageType, itk::NiftiImageIO>(test_image, output_test_fn);
 
-  bool fileHasCorrectXYZTTUnits = false;
-  bool fileHasCorrectTimeOrigin = false;
+  bool metadataHasCorrectXYZTTUnits = false;
+  bool metadataHasCorrectToffset = false;
+  bool imageHasCorrectTimeOrigin = false;
 
   try
   {
     // read the image back in
-    ImageType::Pointer image_from_disk = itk::IOTestHelper::ReadImage<ImageType>(std::string(output_test_fn));
+    ImageType::Pointer image_from_disk = itk::IOTestHelper::ReadImage<ImageType>(output_test_fn);
 
     // check the metadata for xyzt_units and toffset
     itk::MetaDataDictionary & dictionary = image_from_disk->GetMetaDataDictionary();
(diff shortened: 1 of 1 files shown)


thewtex (approved):
@cookpa thanks for the follow-up 🙏

dzenanz (approved):

cookpa:
Actually I realize this might still not work

```
const std::string output_test_fn = std::string(test_dir) + "xyzt_units_output_check.nii";
```

Because it assumes a path separator is at the end of test_dir. Other tests use

```
itksys::SystemTools::ChangeDirectory(testdir);
```

I think I will do that as well to avoid the potential for other problems with the bad working directory

thewtex:
If it is possible to pass in full paths to the test as args, that generally avoids issues.

cookpa:
@thewtex is it OK for me to use a forward slash, like

```
  const std::string output_test_fn = std::string(test_dir) + "/xyzt_units_output_check.nii";
```

dzenanz:
Yes, Windows supports forward slashes in file paths.

cookpa:
One unrelated change in this last commit: while running in verbose mode to look into the path issue, I noticed the metadata wasn't being checked correctly. The metadata is correct - it just would not have failed the test if it was wrong.

I also figured out that the `/a_blatantly_obvious/bad_file_path` message is actually not an error, it's done by `itk::IOTestHelper::WriteImage` to test that an exception is thrown with a bad path.

cookpa:
One more change, I added a line to throw the exception if the nifti file cannot be read in. That might shed more light on the failure.

hjmjohnson:
NOTE:  "/a_blatantly_obvious/bad_file_path/that/should/never/exist/on/the/computer/" is purposfully not supposed to exist. It is a path that is used to test proper behavior when a path does not exist.

hjmjohnson (approved):

Modules/IO/NIFTI/test/itkNiftiImageIOTest14.cxx
```diff
@@ -90,7 +90,7 @@ itkNiftiImageIOTest14(int argc, char * argv[])
     }
   }
 
-  const char * output_test_fn = "xyzt_units_output_check.nii";
+  const std::string output_test_fn = std::string(test_dir) + "/xyzt_units_output_check.nii";
```
jhlegarreta, line 93: Probably cleaner to use: `std::filesystem::path::preferred_separator` or a `std::filesystem::path` object to concatenate paths and then get the `std::string` from it. I think both require C++17, which we do.

jhlegarreta (commented):
Thanks for fixing this so quickly @cookpa.

Left an inline comment.

Modules/IO/NIFTI/test/itkNiftiImageIOTest14.cxx
seanm: `std::filesystem` isn't really used in ITK yet is it? Seems too close to the 5.4 release to start using it. Yes, C++17 is already required, but using `std::filesystem` also makes requirements of the system' STL library...

jhlegarreta: >  but using std::filesystem also makes requirements of the system' STL library

Not sure if that would pose a problem. But fair enough, the issue may be considered after 5.4, as PR #4562 makes progress.

Feel free to resolve the conversation and merge the PR then.

cookpa: Thanks for these comments, good to know the future plan but will leave as-is for now then

cookpa:
I think this is ready now, I can rebase if needed but I think admins can squash and merge directly now

Modules/IO/NIFTI/test/itkNiftiImageIOTest14.cxx
seanm: >Not sure if that would pose a problem. But fair enough, the issue may be considered after 5.4, as PR https://github.com/InsightSoftwareConsortium/ITK/pull/4562 makes progress.

In fact, I looked into it, and I think it would force dropping support for macOS <= 10.14, which would be a shame:

https://stackoverflow.com/questions/58667853/does-use-of-c17-stdfilesystem-require-macos-10-15-xcode-11-1

But indeed, that's for https://github.com/InsightSoftwareConsortium/ITK/pull/4562 now...

blowekamp: > Probably cleaner to use: `std::filesystem::path::preferred_separator` or a `std::filesystem::path` object to concatenate paths and then get the `std::string` from it. I think both require C++17, which we do.

itksys::SystemTools::JoinPath is the current way.

cookpa: Thanks @blowekamp I will update

hjmjohnson: Thanks for looking into this issue.

cookpa:
Hi @thewtex I was going to make the [path change as suggested](https://github.com/InsightSoftwareConsortium/ITK/pull/4642#discussion_r1591180629) by @blowekamp but I see you've got other things going on with this PR. I think it will work as-is, so happy to leave it be if that makes things easier.

thewtex:
Hi @cookpa , yes if you could please make that improvement in a follow-up PR, it will be appreciated. I will merge this now so we can get clean builds for other PRs.

PR #4644, opened 2024-05-04
COMP: restored old spelling of CLANG_SUPPRESS_Wfloat_equal

seanm:
Brought back #define CLANG_SUPPRESS_Wfloat_equal for compatibility with third party code using this as public API.

While here, also prefixed some other defines with ITK to indicate they are from ITK.


SimonRit (approved):
Thanks!

thewtex:
Applying clang-format.

thewtex (approved):

PR #4651, opened 2024-05-08
ENH: Add pre-commit configuration file to run clang-format

blowekamp:
The "pre-commit" be installed congenitally or with the `Utilities/GitSetup/setup-precommit` installation script.

This version of clang-tidy can be run on all files by `pre-commit run -a`.


hjmjohnson (approved):
@blowekamp  I love this idea!!!  I have found that pre-commit is much more reliable than our custom built tools.    It makes it much easier to keep the codebase clean than does our current system.

After this initial commit is in place, I'd be happy to propose a set of other changes that will make maintenance a lot easier for new developers.

blowekamp:
> @blowekamp I love this idea!!! I have found that pre-commit is much more reliable than our custom built tools. It makes it much easier to keep the codebase clean than does our current system.
> 
> After this initial commit is in place, I'd be happy to propose a set of other changes that will make maintenance a lot easier for new developers.

You may be interested in what I did in SimpleITK:
https://github.com/SimpleITK/SimpleITK/blob/master/.pre-commit-config.yaml

Utilities/GitSetup/setup-precommit
```diff
@@ -0,0 +1,56 @@
...
+
+die() {
+	echo 1>&2 "$@" ; exit 1
+}
+
+# Make sure we are inside the repository.
+cd "${BASH_SOURCE%/*}" &&
+
+# check if python executable exists and is at least MIN_PYTHON_VERSION
+if ! command -v python3 &> /dev/null; then
```
thewtex, line 15: On Windows Git Bash, this is `command -v python3` is going succeed even if Python is not installed or Python is installed but python.exe is not in the `PATH`. `python3` is a non-functional stub that suggests installing Python via the Windows store.

blowekamp: Do you have a suggestion to what would work here?

thewtex: I looked into using `uv` briefly, but it complained about not being able to find `python` -- it will require more digging or investigation into other approaches.

thewtex: This may work for a local Windows python:

https://github.com/pypa/hatch/blob/7986421e531d37155cfabd615546c4cdc658e385/src/hatch/python/distributions.py#L43-L46

thewtex: I have heard that `uv` has plans to distributed python, too.

blowekamp: There are two things here: 1) is python3 installed and working and 2) If it's not can the ITK configuration automagically install and use a version of python.

Perhaps we can only address #1 in this initial PR?

thewtex: Yes, only 1) can be addressed first, but the commit hook should be optional when python3 is not available (hook setup should or execution should not fail).

seanm: automagically installing python seems frought, there are so many OSes and ways to install python...

thewtex: It is not a system install, just a local binary, similar to the commit hooks themselves.

hjmjohnson:
@blowekamp Thanks. Superceeded by https://github.com/InsightSoftwareConsortium/ITK/pull/5032 and https://github.com/InsightSoftwareConsortium/ITK/pull/5013

PR #4654, opened 2024-05-10
STYLE: Remove `this->MakeOutput(0)` calls from constructors in Core

N-Dekker:
In those cases, `MakeOutput(0)` just did `OutputType::New()` anyway. This commit avoids unnecessary casts and calls to virtual functions.

Following C++ Core Guidelines, February 15, 2024, "Don’t call virtual functions in constructors and destructors",
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-ctor-virtual

----

Intended for _after_ the release of ITK v5.4.0

### DIFF
--- a/Modules/Core/Common/include/itkImageSource.hxx
+++ b/Modules/Core/Common/include/itkImageSource.hxx
@@ -39,11 +39,10 @@ namespace itk
 template <typename TOutputImage>
 ImageSource<TOutputImage>::ImageSource()
 {
-  // Create the output. We use static_cast<> here because we know the default
-  // output must be of type TOutputImage
-  const typename TOutputImage::Pointer output = static_cast<TOutputImage *>(this->MakeOutput(0).GetPointer());
   this->ProcessObject::SetNumberOfRequiredOutputs(1);
-  this->ProcessObject::SetNthOutput(0, output.GetPointer());
+
+  // Equivalent to SetNthOutput(0, MakeOutput(0)); in this case, calling MakeOutput is not necessary.
+  this->ProcessObject::SetNthOutput(0, TOutputImage::New());
 
   // Set the default behavior of an image source to NOT release its
   // output bulk data prior to GenerateData() in case that bulk data
--- a/Modules/Core/Mesh/include/itkImageToMeshFilter.hxx
+++ b/Modules/Core/Mesh/include/itkImageToMeshFilter.hxx
@@ -27,11 +27,10 @@ template <typename TInputImage, typename TOutputMesh>
 ImageToMeshFilter<TInputImage, TOutputMesh>::ImageToMeshFilter()
 {
   this->ProcessObject::SetNumberOfRequiredInputs(1);
-
-  const OutputMeshPointer output = dynamic_cast<OutputMeshType *>(this->MakeOutput(0).GetPointer());
-
   this->ProcessObject::SetNumberOfRequiredOutputs(1);
-  this->ProcessObject::SetNthOutput(0, output.GetPointer());
+
+  // Equivalent to SetNthOutput(0, MakeOutput(0)); in this case, calling MakeOutput is not necessary.
+  this->ProcessObject::SetNthOutput(0, OutputMeshType::New());
 }
 
 /**
--- a/Modules/Core/Mesh/include/itkMeshSource.hxx
+++ b/Modules/Core/Mesh/include/itkMeshSource.hxx
@@ -26,12 +26,10 @@ template <typename TOutputMesh>
 MeshSource<TOutputMesh>::MeshSource()
 
 {
-  // Create the output. We use static_cast<> here because we know the default
-  // output must be of type TOutputMesh
-  const OutputMeshPointer output = static_cast<TOutputMesh *>(this->MakeOutput(0).GetPointer());
-
   this->ProcessObject::SetNumberOfRequiredOutputs(1);
-  this->ProcessObject::SetNthOutput(0, output.GetPointer());
+
+  // Equivalent to SetNthOutput(0, MakeOutput(0)); in this case, calling MakeOutput is not necessary.
+  this->ProcessObject::SetNthOutput(0, TOutputMesh::New());
 }
 
 template <typename TOutputMesh>


Modules/Core/Common/include/itkImageSource.hxx
```diff
@@ -39,11 +39,8 @@ namespace itk
 template <typename TOutputImage>
 ImageSource<TOutputImage>::ImageSource()
 {
-  // Create the output. We use static_cast<> here because we know the default
-  // output must be of type TOutputImage
-  typename TOutputImage::Pointer output = static_cast<TOutputImage *>(this->MakeOutput(0).GetPointer());
   this->ProcessObject::SetNumberOfRequiredOutputs(1);
-  this->ProcessObject::SetNthOutput(0, output.GetPointer());
+  this->ProcessObject::SetNthOutput(0, TOutputImage::New().GetPointer());
```
blowekamp, line 43: Please add comment like: "Generally MakeOutput(n) should be used, but because a virtual method should not be called in a constructor the output is directly constructed."

It is important to note how things should be done, and why they are not done that way for this case. There have been bugs before related to not using MakeOutput.

N-Dekker: Thanks for the suggestion @blowekamp

> There have been bugs before related to not using MakeOutput.

Can you possibly still recall what such a bug looked like? Just for my understanding!

blowekamp: Don't recall the specifics... There are cases when a Process object can have it's outputs removed, and then during execution it needs to make them again.

Maybe f0d570ecfce3d7ca2f099f3f4e18c9e343fa4fc6?

N-Dekker: Interesting! In general, I think `this->MakeOutput(i)` calls are not necessary, inside a constructor of a class, when the `MakeOutput(i)` override of this class unconditionally just does the same `OutputType::New()` call for any index value `i`.

blowekamp: I agree they are not necessary. I am just suggestion adding documentation as to why the canonical way of creating a new output is not being used.

N-Dekker: I think I understand your point, thanks! I would just like to get clearer in which particular cases `this->MakeOutput(i)` would actually be useful, in a constructor.

Is `x->MakeOutput(i)` also intended to be called from outside? I mean, would an ITK _end_ user actually call `x->MakeOutput(i)`?

blowekamp: Hrmm... This might undo fixes made here: https://github.com/InsightSoftwareConsortium/ITK/pull/961

N-Dekker: Sorry, what do you mean? Your pull request #961 (merged on May 30, 2019) was about different source files (itkDataObject.h, itkProcessObject.cxx, itkDataObjectAndProcessObjectTest.cxx). Whereas this PR (PR #4654) only does internal style changes to the constructors of `ImageSource`, `MeshSource`, and `ImageToMeshFilter`. I think this pull request is innocent  😇

N-Dekker:
I think this pull request may also be merged now, as v5.4.0 has been tagged (https://github.com/InsightSoftwareConsortium/ITK/pull/4603#issuecomment-2120498131)

Modules/Core/Common/include/itkImageSource.hxx
N-Dekker: Oh, sorry Bradley, what do you want there? Something like

    // Basically equivalent to SetNthOutput(0, MakeOutput(0)).
    this->ProcessObject::SetNthOutput(0, TOutputImage::New().GetPointer())

?

I hope it does not need to be longer than one line of comment. This pull request is intended to simplify the code. The idea is: Just call `TOutput::New()` if that does the job!

blowekamp: What do you mean "As the idea would be to use New(), rather that MakeOutput, in multiple places in the source tree"?

N-Dekker: Well, this pull request already proposes to replace three `MakeOutput` calls with `TOutput::New()`. I don't think it's nice to have an extensive multi-line comment copy-and-pasted multiple times.

N-Dekker:
Wow, the beautiful unit test has failed! At https://open.cdash.org/tests/1563467945

> Error in !shrink->HasObserver(itk::StartEvent())

N-Dekker:
- Marked "draft", as I'm now considering another approach, calling `ProcessObject::MakeRequiredOutputs(*this, 1)` instead, following pull request #4688

blowekamp:
@N-Dekker Can this PR be closed, with #4688 addressing it?

N-Dekker:
> @N-Dekker Can this PR be closed, with #4688 addressing it?

Good question @blowekamp 😃 This PR specifically addresses cases where there is exactly 1 required output. PR #4688 does not yet address those cases. So it's still open 🤷

hjmjohnson:
@N-Dekker I took a fresh look at where this PR stands. It now conflicts with `main` (the touched lines gained `const`), and since `ProcessObject::MakeRequiredOutputs` landed via #4688, rewriting these three constructors as `ProcessObject::MakeRequiredOutputs(*this, 1);` would seem to resolve everything at once — it avoids the virtual call in the constructor while still going through `MakeOutput`, addressing @blowekamp's request to keep the canonical creation path. Would you prefer to update this PR to that approach, keep the direct `TOutput::New()` form, or close in favor of a fresh PR?

<details>
<summary>Status details</summary>

- Draft since June 2024, when you noted you were considering `MakeRequiredOutputs(*this, 1)` following #4688.
- `MakeRequiredOutputs` dispatches statically (`sourceObject.TSourceObject::MakeOutput(i)`), so it satisfies the C++ Core Guidelines rule this PR targets, and it also subsumes the `SetNumberOfRequiredOutputs(1)` call — one line per constructor, no explanatory comment needed.
- Eight files on `main` already use it (Statistics, DistanceMap, Watersheds, BlockMatching, EigenAnalysis2D).
- Behavior-neutral for these three classes: each class's own `MakeOutput` override unconditionally returns `TOutput::New()`, and constructors never dispatched to subclass overrides anyway.
- Follow-on: ~30 more single-output `this->MakeOutput(0)` constructor sites remain across Statistics, Registration, Path, Video, etc. Whatever idiom lands here in Core sets the precedent for sweeping those.

</details>

N-Dekker:
@hjmjohnson Thanks, I think the approach proposed by this PR is still fine (directly using `TOutput::New()`).

~~I'll try to fix the conflicts!~~ I think I fixed the conflicts (with this [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/91205663edce2c100faec1317fcf1cb33f5d7763..c4033b0f2d9eaaae126f5823670f2a8bd97729d8))

Rationale: A single `TOutput::New()` call appears clearer to me than `ProcessObject::MakeRequiredOutputs(*this, 1)`. It feels a bit overdone to me to call "MakeRequiredOutputs" (suggesting _multiple_ outputs) for just a single output.  And calling MakeOutput is also unnecessary, when `TOutput::New()` just does the job.

However, I have to admit that the gain of this PR is only small  🤷 It's just a matter of style!

hjmjohnson:
@N-Dekker Is it OK if I add another commit (for consistency) to addres :  ~30 more single-output this->MakeOutput(0) constructor sites remain across Statistics, Registration, Path, Video, etc. Whatever idiom lands here in Core sets the precedent for sweeping those.

N-Dekker:
> Is it OK if I add another commit (for consistency) to addres : ~30 more single-output this->MakeOutput(0) constructor sites remain across Statistics, Registration, Path, Video, etc. Whatever idiom lands here in Core sets the precedent for sweeping those.

Can you please do such a follow-up in a separate PR (although you may cherry-pick this PR's commit). Thanks, Hans.

Maybe some people still prefer the original code, _or_ calling `ProcessObject::MakeRequiredOutputs(*this, 1)`, so let's first see this `TOutput::New()` approach passes the review.

Modules/Core/Common/include/itkImageSource.hxx
```diff
@@ -39,11 +39,10 @@ namespace itk
...
 ImageSource<TOutputImage>::ImageSource()
 {
-  // Create the output. We use static_cast<> here because we know the default
-  // output must be of type TOutputImage
-  const typename TOutputImage::Pointer output = static_cast<TOutputImage *>(this->MakeOutput(0).GetPointer());
   this->ProcessObject::SetNumberOfRequiredOutputs(1);
-  this->ProcessObject::SetNthOutput(0, output.GetPointer());
+
+  // Equivalent to SetNthOutput(0, MakeOutput(0)); in this case, calling MakeOutput is not necessary.
+  this->ProcessObject::SetNthOutput(0, TOutputImage::New());
```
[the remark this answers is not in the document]
hjmjohnson: comment is ok.

PR #4659, opened 2024-05-13
BUG: Use unique filenames for Nifti tests

thewtex:
When the tests are run in parallel they can try to write to or remove
the same file from different processes, causing intermittent failures.

Also add `const` where it is missing.

Also use full paths as tests inputs, instead of changing directories,
which is more explicit.

### DIFF
--- a/Modules/IO/NIFTI/test/itkNiftiImageIOTest14.cxx
+++ b/Modules/IO/NIFTI/test/itkNiftiImageIOTest14.cxx
@@ -32,19 +32,18 @@ itkNiftiImageIOTest14(int argc, char * argv[])
   // images should have the same size, spacing, and origin, but may have different units
   if (argc != 4)
   {
-    std::cerr << "Usage: " << itkNameOfTestExecutableMacro(argv) << " test_dir ref_image test_image" << std::endl;
+    std::cerr << "Usage: " << itkNameOfTestExecutableMacro(argv) << " output_test_fn ref_image test_image" << std::endl;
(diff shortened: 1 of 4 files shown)


thewtex:
@cookpa please take a look

jhlegarreta (approved):
Not sure if the output NIfTI files can be compressed (`*.nii.gz`) directly to save space, but approving.

cookpa:
Thanks for tracking this down @thewtex . The fixes look good to me.

thewtex:
> NIfTI files can be compressed (*.nii.gz) directly to save space, but approving.

Yes, this works locally -- updating.

Modules/IO/NIFTI/test/itkNiftiImageIOTest14.cxx
```diff
@@ -162,8 +159,6 @@ itkNiftiImageIOTest14(int argc, char * argv[])
     throw;
   }
 
-  itk::IOTestHelper::Remove(output_test_fn.c_str());
-
```
N-Dekker, line 166: @thewtex Are you sure that the created output file should not be removed afterwards?

thewtex: It is better to keep it for inspection.

hjmjohnson (approved):

N-Dekker (approved):

hjmjohnson:
@thewtex THANK YOU!

PR #4661, opened 2024-05-13
COMP: Fix the WDocumentation warnings for enum class.

andrei-sandor:
Used with clang

Replaced \class Doxygen annotations with \enum to avoid warnings

### DIFF
--- a/Modules/Core/Common/ITKKWStyleOverwrite.txt
+++ b/Modules/Core/Common/ITKKWStyleOverwrite.txt
@@ -33,3 +33,14 @@ itkMultiThreaderBase.h Comments Disable
 itkSingleton\.cxx Namespace Disable
 itkExceptionObject\.h IfNDefDefine Disable
 itkConceptChecking\.h Template Disable
+itkCommonEnums\.h Comments Disable
+itkExtractImageFilter\.h Comments Disable
+itkFloatingPointExceptions\.h Comments Disable
+itkFrustumSpatialFunction\.h Comments Disable
+itkGaussianDerivativeOperator\.h Comments Disable
+itkLoggerBase\.h Comments Disable
+itkLoggerThreadWrapper\.h Comments Disable
+itkMultiThreaderBase\.h Comments Disable
+itkObjectStore\.h Comments Disable
+itkSpatialOrientation\.h Comments Disable
+itkSymmetricEigenAnalysis\.h Comments Disable
--- a/Modules/Core/Common/include/itkCommonEnums.h
+++ b/Modules/Core/Common/include/itkCommonEnums.h
@@ -40,7 +40,6 @@ class CommonEnums
 public:
   // Used in Input/Output for Images/Mesh/Transform types
   /**
-   * \class IOPixel
    * \ingroup ITKCommon
    * Enums used to manipulate the point/cell pixel type. The pixel type provides
    * context for automatic data conversions (for instance, RGB to
@@ -67,7 +66,6 @@ public:
   };
 
   /**
-   * \class IOComponent
    * \ingroup ITKCommon
    * Enums used to manipulate the component type. The component type
    * refers to the actual storage class associated with either a
@@ -92,7 +90,6 @@ public:
   };
 
   /**
-   * \class IOFile
    * \ingroup ITKCommon
    * Enums used to specify write style: whether binary or ASCII. Some
    * subclasses use this, some ignore it.
@@ -108,7 +105,6 @@ public:
   };
 
   /**
-   * \class IOFileMode
    * \ingroup ITKCommon
    * Mode in which the files is intended to be used
    */
@@ -119,7 +115,6 @@ public:
   };
 
   /**
-   * \class IOByteOrder
    * \ingroup ITKCommon
    * Enums used to specify byte order; whether Big Endian or Little Endian.
    * Some subclasses use this, some ignore it.
@@ -132,7 +127,6 @@ public:
   };
 
   /**
-   * \class CellGeometry
    * \ingroup ITKCommon
    * Enums used to specify cell type */
   enum class CellGeometry : uint8_t
@@ -190,7 +184,7 @@ extern ITKCommon_EXPORT std::ostream &
 class MeshEnums
 {
 public:
-  /** \class MeshClassCellsAllocationMethod
+  /**
    * \ingroup ITKCommon
    * Enum defining the possible methods used to allocate memory for
    * the Cells */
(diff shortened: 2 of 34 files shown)


N-Dekker:
Thanks @andrei-sandor !  Cool!

I'm interested to hear if these changes will affect the Doxygen output, for example at https://itk.org/Doxygen/html/classitk_1_1CommonEnums.html 

The proposed changes look fine to me, but they caused some (undeserved) KWStyle errors. Looking at https://open.cdash.org/tests/1552684315

> /home/vsts/work/1/s/Modules/Core/Common/include/itkCommonEnums.h:49: error: comment doesn't have \class
> /home/vsts/work/1/s/Modules/Core/Common/include/itkCommonEnums.h:76: error: comment doesn't have \class
> /home/vsts/work/1/s/Modules/Core/Common/include/itkCommonEnums.h:100: error: comment doesn't have \class

 Do you (or anyone) know how to address those as well?

dzenanz:
Maybe [this condition](https://github.com/Kitware/KWStyle/blob/6971e89ec4c763601edf1d0ea0d6f71d42fbca5d/kwsCheckComments.cxx#L197) needs to be updated to consider `enum` in addition to `class`? Or we just need to update [style overwrite rules](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/ITKKWStyleOverwrite.txt) to add appropriate exceptions - probably `Comments` or `NameOfClass` from [this list](https://github.com/Kitware/KWStyle/blob/6971e89ec4c763601edf1d0ea0d6f71d42fbca5d/kwsStyle.cxx#L176-L179).

Modules/Core/Common/include/itkCommonEnums.h
```diff
@@ -132,7 +132,7 @@ class CommonEnums
   };
 
   /**
-   * \class CellGeometry
+   * \enum CellGeometry
    * \ingroup ITKCommon
    * Enums used to specify cell type */
   enum class CellGeometry : uint8_t
```
N-Dekker, line 138: Ideally I think Doxygen should be able to automatically "defer" that `CellGeometry` is an enum, just by looking at the code. So then the whole line saying `* \... CellGeometry`  might simply be removed. But that may be beyond the scope of this pull request, so for now replacing `\class` with `\enum` looks fine to me.

seanm: Indeed. We could do this instead. But it would be a massive change touching many files.

Is there some reason ITK is explicitly using `\class`?  Did someone think it was required?  Is it a stylistic preference to be explicit? Or...?

seanm: Simply allowing doxygen to infer that the comments are attached to the enum below (as @N-Dekker suggested) seems best to me, as it will also eliminate needing to mess with KWStyle.  What do you think @dzenanz @thewtex ?

dzenanz: If it works, that's fine with me.

andrei-sandor: After removing the \enum, the following warning is produced `Modules/Core/Common/include/itkFloatingPointExceptions.h:37: error: comment doesn't have \class`. What are the possible options? Should a modification be done like here [style overwrite rules](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/ITKKWStyleOverwrite.txt)? Or maybe here [this condition](https://github.com/Kitware/KWStyle/blob/6971e89ec4c763601edf1d0ea0d6f71d42fbca5d/kwsCheckComments.cxx#L197) (adding an enum might be more complex)?

seanm: Anyone have any thoughts on how should we proceed?

thewtex: @andrei-sandor @seanm I think a style overwrite rule is appropriate in this case.

Thank you for your continued contributions :pray:

Modules/Core/Common/include/itkCommonEnums.h
```diff
@@ -270,17 +262,12 @@ extern ITKCommon_EXPORT std::ostream &
...
-   *  Position at which the new factory will be registered in the
-   *  internal factory container.
-   */
-  enum class InsertionPosition : uint8_t
-  {
-    INSERT_AT_FRONT,
-    INSERT_AT_BACK,
-    INSERT_AT_POSITION
-  };
+  *  \ingroup ITKCommon * Position at which the new factory will be registered in the * internal factory container.* /
```
seanm, line 265: Looks like you accidentally smooshed this into 1 line?

Modules/Core/Common/include/itkCommonEnums.h
seanm: Is there a way to run KWStyle locally on all ITK files, from our local computers?  (So that we don't have to depend on github CI.)

dzenanz: Usually `make KWStyle` will work.

seanm (approved):

seanm:
@dzenanz @N-Dekker this is ready for review now

dzenanz (approved):

Modules/Core/Common/ITKKWStyleOverwrite.txt
```diff
@@ -33,3 +33,14 @@ itkMultiThreaderBase.h Comments Disable
 itkSingleton\.cxx Namespace Disable
 itkExceptionObject\.h IfNDefDefine Disable
 itkConceptChecking\.h Template Disable
+itkCommonEnums\.h Comments Disable
```
N-Dekker, line 36: Thanks for your great work! Do I understand correctly that `Comments Disable` might as well just be applied to each and every source file? Because it just isn't a useful warning anyway?

dzenanz: @thewtex and @blowekamp might have some comments about this.

thewtex: The use of `\class ClassName` was a style decision, I believe. We could add these `Comments Disable` for the enum class files. Or, we could change our style, but we probably want to do it consistently throughout and then disable the option globally with the KWStyle rules: https://github.com/InsightSoftwareConsortium/ITK/blob/68619448ee9a6303f0b5958c4c7700c1fd5f78b3/Utilities/KWStyle/ITK.kws.xml#L12

dzenanz:
Should this be merged, or reviewed by someone else too?

seanm:
I think it's ready to merge, but I personally don't know much about the KWStyle suppression stuff...

dzenanz (approved):

PR #4664, opened 2024-05-15
WparamReturn

andrei-sandor:
Contains the following fixes

- Remove parameters in documentation that are not in the class
- Remove \return when no return statement
- Remove \param when no parameter
- Fix small typo

### DIFF
--- a/Modules/Core/Common/include/itkHexahedronCell.h
+++ b/Modules/Core/Common/include/itkHexahedronCell.h
@@ -31,11 +31,6 @@ namespace itk
  *
  * HexahedronCell represents a hexahedron, more precisely, a cuboid, for a Mesh.
  *
- * \tparam TPixelType The type associated with a point, cell, or boundary
- * for use in storing its data.
- *
- * \tparam TCellTraits Type information of mesh containing cell.
- *
  * \todo When reviewing this class, the documentation of the  template
  * parameters MUST be fixed.
  *
--- a/Modules/Core/Common/include/itkIsNumber.h
+++ b/Modules/Core/Common/include/itkIsNumber.h
@@ -28,7 +28,7 @@ namespace itk
 namespace mpl
 {
 /** Tells whether a type is a number.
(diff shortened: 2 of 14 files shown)


dzenanz (approved):

andrei-sandor:
> It looks like you need to apply clang-format to some of the changed files.

Error: Code is inconsistent with ITK Coding Style. Add the *action:ApplyClangFormat* PR label to correct.

This is the message generated by clang-format linter. Is it possible to type action:ApplyClangFormat in a comment here? I am working with a different version of clang-format which gives different formatting changes than here.

dzenanz:
I added that label (in GitHub UI on the right), but the action triggered (`Apply clang-format to PR / clang-format (pull_request)` in the checks list) [crashes](https://github.com/InsightSoftwareConsortium/ITK/actions/runs/9097226451/job/25004550885?pr=4664). ITK uses `clang-format 8.0.3`.

I forgot who created clang-format action, @thewtex or someone else (@jhlegarreta?).

@blowekamp has a parallel PR #4651 to update clang-format used to the latest version.

jhlegarreta:
> I forgot who created clang-format action, @thewtex or someone else (@jhlegarreta?).

I don't think it was me. Cannot remember if adding the label or calling something like `clang-format` as a comment to PRs has ever worked.

dzenanz:
Now that you mentioned it, I think it used to work for PRs made from branches in this repository, but not from branches in forks. But I think that the action itself might be broken now.

thewtex:
I created the Apply Clang Format Action. At the time it was created, it was limited in its ability to apply to PR's from forks because of permission functionality in GitHub Actions. Since then, I think GitHub has added the capability to indicate that an Action can push to a PR, etc. If someone wants to look into the adding the config required, please do. I plan to spend time helping with the pre-commit hook / action setup as a replacement.

seanm (approved):

thewtex (approved):

PR #4665, opened 2024-05-15
Wdeprecated

andrei-sandor:
Added ITK_DEPRECATED when \deprecated is used in a comment. They must both come together. Applied to structs, classes and functions

### DIFF
--- a/Modules/Core/Common/include/itkImage.h
+++ b/Modules/Core/Common/include/itkImage.h
@@ -171,7 +171,7 @@ public:
    * example usage:
    * using OutputImageType = typename ImageType::template Rebind< float >::Type;
    *
-   * \deprecated Use RebindImageType instead
+   * Deprecated: Use RebindImageType instead
    */
   template <typename UPixelType, unsigned int VUImageDimension = VImageDimension>
   struct Rebind
--- a/Modules/Core/Common/include/itkVector.h
+++ b/Modules/Core/Common/include/itkVector.h
@@ -110,7 +110,7 @@ public:
   /** Constructor to initialize entire vector to one value.
    * \warning Not intended to convert a scalar value into
    * a Vector filled with that value.
-   * \deprecated */
+   * Deprecated */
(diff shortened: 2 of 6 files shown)


Modules/Core/Common/include/itkSingleton.h
```diff
@@ -32,8 +32,7 @@
  * (i.e. itkNotUsed cannot be used) but is not always used.
  */
 template <typename T>
-[[deprecated("Preferably use the C++ `[[maybe_unused]]` attribute instead!")]] inline void
```
dzenanz, line 35: This commit seems to be fighting against clang-format, see:
https://github.com/InsightSoftwareConsortium/ITK/actions/runs/9097194889/job/25004442774?pr=4665

seanm:
So the issue here is that `-Wdocumentation` warns if a function/class is documented with a `\deprecated` but isn't also tagged `[[deprecated]]` for the compiler. Indeed, it generally makes sense that you should have both or neither.

But there are some tricky cases here:

This struct is "deprecated":

```c++
/**
   * \brief A structure which enable changing any image class' pixel
   * type to another.
   *
   *
   * \sa Image::Rebind
   * \deprecated Use template alias RebindImageType instead
   */
  template <typename UPixelType, unsigned int VUImageDimension = VImageDimension>
  struct Rebind
  {
    using Type = itk::VectorImage<UPixelType, VUImageDimension>;
  };
```

but its replacement references the deprecated struct:

```c++
using RebindImageType = typename Rebind<UPixelType, VUImageDimension>::Type;
```

If the struct is tagged `[[deprecated]]` then using it above triggers a warning about using a deprecated class.

-----

What do you all think of just changing the `\deprecated` in these cases to just `Deprecated:`?

seanm:
@andrei-sandor please amend the commit messge, as "COMP: Added ITK_DEPRECATED when \deprecated there" is no longer true.

thewtex (approved):
:1234:

PR #4667, opened 2024-05-15
COMP: Fixed Cpp Warnings

andrei-sandor:
Fixed the following warnings

-Wimplicit-fallthrough
-Wzero-as-null-pointer-constant
-Wdeprecated-dynamic-exception-spec

### DIFF
--- a/Modules/Core/QuadEdgeMesh/test/itkQuadEdgeMeshEulerOperatorSplitEdgeTest.cxx
+++ b/Modules/Core/QuadEdgeMesh/test/itkQuadEdgeMeshEulerOperatorSplitEdgeTest.cxx
@@ -42,7 +42,7 @@ itkQuadEdgeMeshEulerOperatorSplitEdgeTest(int, char *[])
   auto splitEdge = SplitEdge::New();
   std::cout << "     "
             << "Test No Mesh Input";
-  if (splitEdge->Evaluate((QEType *)0))
+  if (splitEdge->Evaluate((QEType *)nullptr))
   {
     std::cout << "FAILED." << std::endl;
     return EXIT_FAILURE;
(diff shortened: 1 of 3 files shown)


Modules/Core/Common/include/itkThreadSupport.h
```diff
@@ -51,7 +51,7 @@ using FastMutexType = pthread_mutex_t;
 using ThreadFunctionType = void * (*)(void *);
 using ThreadProcessIdType = pthread_t;
 #  if !defined(ITK_FUTURE_LEGACY_REMOVE)
-constexpr ThreadProcessIdType ITK_DEFAULT_THREAD_ID = 0;
+constexpr ThreadProcessIdType ITK_DEFAULT_THREAD_ID = nullptr;
```
dzenanz, line 54: I am uncomfortable with this change. Where does the warning appear?

andrei-sandor: The warning appears on MacOS. pthread_t is a pointer. However, on Linux, it is an int. We will remove the change.

dzenanz (approved):

thewtex (approved):
@andrei-sandor thank you!

Modules/Core/Common/include/itkThreadSupport.h
N-Dekker: @andrei-sandor @dzenanz If the warning still bothers you, I would suggest using `{}` instead. `{}` should be universal, so it should work well for both pointers and integers:

    constexpr ThreadProcessIdType ITK_DEFAULT_THREAD_ID{};

andrei-sandor:
> @andrei-sandor thank you!

@dzenanz @thewtex. I thank you. It was a pleasure to write PRs for ITK. I learned a lot form the code and from your valuable comments.

PR #4670, opened 2024-05-15
COMP: suppress 5 clang warnings from eigen headers

seanm:
These 4 warnings occur nowhere else in the ITK codebase, so suppressing them here allows them to be enabled on CI.


dzenanz (approved):

seanm:
Added a 5th suppression here for `-Wzero-as-null-pointer-constant`.

PR #4672, opened 2024-05-16
COMP: Fix Unitialized scenarios

andrei-sandor:
This is to fix the -Wconditional-uninitialized warning from clang. The compiler produces this warning in the following scenarios.

1. Variable is declared. Used inside a try clause and processed after the try/catch clause.
2. Variable is declared and used inside a if statement

For scenario 1, for itkANTSNeighborhoodCorrelationImageToImageMetricv4GetValueAndDerivativeThreader.hxxTo, the compiler complains about movingImageValue. This variable is declared and used inside the try clause. After the try clause, it is used. The compiler doesn't seem to catch that it is used in the try clause. To fix scenario 1, the code after the try/catch clause is brought inside the try statement. The two blocks depend on pointIsValid which is used inside the if statements. Hence, everything is found inside the try and the compiler doesn't complain about uninitialized while respecting the logic of the if statements and respecting the catch clause.

To fix scenario 2, the variable will be set to 0.0 which seems to be the easiest fix. Since the variable is processed inside if statements, setting to 0.0 is the easiest fix to solve the uninitialized problem.

### DIFF
--- a/Modules/Filtering/ImageStatistics/test/itkLabelStatisticsImageFilterTest.cxx
+++ b/Modules/Filtering/ImageStatistics/test/itkLabelStatisticsImageFilterTest.cxx
@@ -93,7 +93,7 @@ itkLabelStatisticsImageFilterTest(int argc, char * argv[])
   using RegionType = FilterType::RegionType;
   using LabelPixelType = FilterType::LabelPixelType;
 
-  LabelPixelType labelValue;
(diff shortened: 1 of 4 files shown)


dzenanz (approved):
Looks good, but clang-format needs to be applied.

thewtex (approved):

PR #4673, opened 2024-05-16
Wunused macros

andrei-sandor:
COMP: Fix unused macros related to MSVC and OS

This is to fix the -Wunused-macros generated by clang for macros related to MSVC and Microsoft.

For itkNumericTraitsTest.cxx, the _MSC_VER is added to respect the comment.

For itkTestDriver.cxx, ITK_TEST_DRIVER_PATH_SLASH / is related to non Windows machine and there is none in the code and hence it is removed. Now, for ITK_TEST_DRIVER_PATH_SLASH \\, it is not found in the entire code. This is the reason why it is removed.

------------------------------------------------------------------------------------------------

 COMP: Fix unused general macros

This is to fix the -Wunused-macros warning from clang. This is about fixing general macros used in the code.

In the commit, the macros are removed. However, I am wondering if we should comment them since someone might use them in the future.

------------------------------------------------------------------------------------------------

 COMP: Fixed unused legacy remove Macros

This is a commit to solve the Wunused-macros warning generated by clang related to ITK_LEGACY_REMOVE macro.

The change is to create an ifnded macro with it even if ITK_LEGACY_REMOVE is not used in the file. In itkConfigue.h, ITK_LEGACY_REMOVE is defined there if the code is in the legacy remove scenario or not. We need to keep ITK_LEGACY_REMOVE to create a flag that the code is legacy.


Modules/IO/BioRad/src/itkBioRadImageIO.cxx
```diff
@@ -31,9 +31,6 @@
 #include "itkMakeUniqueForOverwrite.h"
 
 #define BIORAD_HEADER_LENGTH 76
-#define BIORAD_NOTE_LENGTH 96
```
dzenanz, line 34: I would prefer to comment these out. While this code is unlikely to change, having this commented out is better than removed.

seanm: For this file only, or all similar cases in this same commit?

dzenanz: For all the changes in this commit.

Modules/IO/GIPL/src/itkGiplImageIO.cxx
```diff
@@ -40,35 +40,34 @@ class GiplImageIOInternals
...
-#define OBLIQUE_TOMO 12
+// #define GIPL_C_SHORT 144
+// #define GIPL_C_INT 160
+// #define GIPL_C_FLOAT 192
+// #define GIPL_C_DOUBLE 193
+// #define GIPL_SURFACE 200
+// #define GIPL_POLYGON 201
+//
+// /*  ORIENTATION DEFINITIONS (flag1)  */
+//
```
seanm, line 51: These 3 lines don't need to have `//` added

Modules/Core/Common/test/itkSTLThreadTest.cxx
```diff
@@ -15,8 +15,9 @@
  *  limitations under the License.
  *
  *=========================================================================*/
-
-#define ITK_LEGACY_TEST // so deprecation warnings are not triggered by this test
+#ifndef ITK_LEGACY_REMOVE
+#  define ITK_LEGACY_TEST
+#endif
```
seanm, line 20: The explanatory comment should be kept.

Modules/IO/GIPL/src/itkGiplImageIO.cxx
```diff
@@ -40,35 +40,34 @@ class GiplImageIOInternals
...
 
 /*  FORMAT DEFINITIONS  */
 
-#define FORMAT_GIPL 0
-#define FORMAT_GIPL_STRING "Gipl"
-#define FORMAT_MAYO 1
-#define FORMAT_MAYO_STRING "Mayo"
-#define FORMAT_NM_IGE 2
-#define FORMAT_NM_IGE_STRING "Starcam"
-
```
seanm, line 71: Sorry to be so nitpicky, but this empty line should be brought back I guess, as presumably these defines are grouped in some manner.

seanm:
/azp.run ITK.Linux

seanm (approved):

hjmjohnson (approved):

PR #4674, opened 2024-05-16
COMP: Fix ununsed code ( if(false) )

andrei-sandor:
This is a comment to fix the -Wunreachable-code warning generated by clang.

This fix simply removes if (false) clauses. An investigation was done to see why these clause exist. This change was done 10 years ago. The clauses at that time were if(0).

### DIFF
--- a/Modules/Registration/Metricsv4/test/itkJointHistogramMutualInformationImageToImageRegistrationTest.cxx
+++ b/Modules/Registration/Metricsv4/test/itkJointHistogramMutualInformationImageToImageRegistrationTest.cxx
@@ -156,6 +156,7 @@ itkJointHistogramMutualInformationImageToImageRegistrationTest(int argc, char * 
   std::cout << argc << std::endl;
   unsigned int numberOfIterations = 10;
   unsigned int numberOfDisplacementIterations = 10;
+  bool         useScalesEstimator = true;
   if (argc >= 5)
   {
     numberOfIterations = std::stoi(argv[4]);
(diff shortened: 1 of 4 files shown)


dzenanz (commented):
This pattern is used to be able to quickly change the code to test a feature.

The proper fix here would be to expose a new command-line parameter and use it in the if clause. Bonus points if that new command line parameter is exercised in corresponding test specification, e.g. https://github.com/InsightSoftwareConsortium/ITK/blob/v5.4rc03/Modules/Registration/Metricsv4/test/CMakeLists.txt#L113-L123.

andrei-sandor:
For the last commit, we would like to know what better name would it be to replace foo. Here, foo represents the variable false.

Modules/Registration/Metricsv4/test/itkMultiGradientImageToImageMetricv4RegistrationTest.cxx
```diff
@@ -61,6 +61,10 @@ itkMultiGradientImageToImageMetricv4RegistrationTest(int argc, char * argv[])
   {
     numberOfIterations = std::stoi(argv[4]);
   }
+  if (argc >= 6)
+  {
+    foo = std::stoi(argv[5]);
```
dzenanz, line 66: `foo` should be rename to `useSubsampling`.

Modules/Registration/Metricsv4/test/itkJointHistogramMutualInformationImageToImageRegistrationTest.cxx
```diff
@@ -164,6 +164,10 @@ itkJointHistogramMutualInformationImageToImageRegistrationTest(int argc, char *
   {
     numberOfDisplacementIterations = std::stoi(argv[5]);
   }
+  if (argc >= 7)
```
dzenanz, line 167: You need to declare the variables, e.g.
```c++
bool useScalesEstimator = true;
if (argc >= 7)
...
```

thewtex (approved):
@andrei-sandor thank you!

PR #4675, opened 2024-05-16
BUG: Wrap SymmetricEigenAnalysisImageFilter with CovariantVector Image output

### LINKED ISSUE #4520: SymmetricEigenAnalysisImageFilter incorrect python wrapping
The python implementation of SymmetricEigenAnalysisImageFilter is returning 6 values per pixel instead of 3, and more than 3 of the 6 values is non-zero. Since there are only 3 eigenvalues for a 3D Hessian (3x3) eigenvalue decomposition, this functionality must be incorrect.

I asked about this on the forum and the comment was that it was probably wrapped incorrectly.
"Maybe the problem is in [wrapping specification](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/ImageIntensity/wrapping/itkSymmetricEigenAnalysisImageFilter.wrap#L9)? [The test](https://github.com/InsightSoftwareConsortium/ITK/blob/be79ceb0a9343c02dba310f5faee371941f6fa40/Modules/Filtering/ImageIntensity/test/itkSymmetricEigenAnalysisImageFilterTest.cxx#L319) instantiates the filter with <SSRT<3>, FixedArray<3> >, whereas the wrapping instantiates with <SSRT<3>, SSRT<3> >."

https://discourse.itk.org/t/computing-hessian-eigenvalues-in-python-how-to-interpret-output-of-symmetriceigenanalysisimagefilter/6517

### Steps to Reproduce

1. Use a test image (ideally a CT scan) with size >= 100^3
2. Run this code:
```py
import itk
#Load the sample image, ideally a CT scan
hessian_image = itk.hessian_recursive_gaussian_image_filter(
            itk_image, sigma=2.0
        )
#hessian_image size = (6, 188, 217, 259)
eigen_values_image = itk.SymmetricEigenAnalysisImageFilter(hessian_image)
#eigen_values_image size = (6, 188, 217, 259)

eigen_values_array = itk.GetArrayFromImage(eigen_values_image, keep_axes=True)
print( eigen_values_array[:,100,100,100] )
```
### Expected behavior

You should see an array printed out that has 6 values. Similar to this, but the numbers obviously will not be the same
array([ -3.0928779 , 1.86561006, 0. , -11.97670084,
6.26783045, 0. ])

### Actual behavior

```none
```

### Reproducibility

Unknown
### Versions

5.3.0

### Environment

python:
4.9.18 (main, Feb 13 2024, 10:56:47)
[GCC 12.2.0]

thewtex:
The default template parameter output, an Image of
SymmetricSecondRankTensor, is not correct or valid. The output of this
filter is an image of eigenvectors. Wrap with an Image of
CovariantVector along with base classes. Also remove the invalid default
template parameter output.

Closes #4520

### DIFF
--- a/Modules/Core/Common/wrapping/itkImageToImageFilterB.wrap
+++ b/Modules/Core/Common/wrapping/itkImageToImageFilterB.wrap
@@ -1,3 +1,4 @@
+itk_wrap_include("itkSymmetricSecondRankTensor.h")
 itk_wrap_include("itkPhasedArray3DSpecialCoordinatesImage.h")
 itk_wrap_class("itk::ImageToImageFilter" POINTER)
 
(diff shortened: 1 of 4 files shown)


dzenanz:
@chatti if you built the Python wrapping from source, this should be easy to test.

aylward (approved):

PR #4676, opened 2024-05-16
Wmissing variable declaration

andrei-sandor:
COMP: Fix missing declarations of unused variables.

This is to fix the -Wmissing-variable-declarations warning of clang.

This is simply related that the variable in question is not used in the file

---------------------------------------------------------------------------

 COMP: Fix missing variables declaration (static)

This is a fix about -Wmissing-variable-declarations warning of clang

This is about the static case. A lot of variables were marked as warning due to non declaration. The static keyword was added when it makes sense that the variable is used internally.

### DIFF
--- a/Examples/RegistrationITKv4/DeformableRegistration17.cxx
+++ b/Examples/RegistrationITKv4/DeformableRegistration17.cxx
@@ -101,9 +101,11 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "itkResampleImageFilter.h"
 #include "itkDisplacementFieldTransform.h"
 
-
+namespace
+{
 unsigned int RmsCounter = 0;
 double       MaxRmsE[4] = { 0.8, 0.75, 0.4, 0.2 };
+} // namespace
 
 //  The following section of code implements a Command observer
 //  that will monitor the evolution of the registration process.
--- a/Modules/Filtering/Path/test/itkContourExtractor2DImageFilterTest.cxx
+++ b/Modules/Filtering/Path/test/itkContourExtractor2DImageFilterTest.cxx
@@ -55,15 +55,16 @@ itkContourExtractor2DImageFilterTestNamespace::MyVertexType _bottom_left[] = {
   itkContourExtractor2DImageFilterTestNamespace::MyVertexType(0, 7.5),
   itkContourExtractor2DImageFilterTestNamespace::MyVertexType(0.5, 8)
 };
-itkContourExtractor2DImageFilterTestNamespace::MyVertexListType bottom_left(_bottom_left, _bottom_left + 2);
+static itkContourExtractor2DImageFilterTestNamespace::MyVertexListType bottom_left(_bottom_left, _bottom_left + 2);
 
-itkContourExtractor2DImageFilterTestNamespace::MyVertexType _rev_bottom_left[] = {
+static itkContourExtractor2DImageFilterTestNamespace::MyVertexType _rev_bottom_left[] = {
   itkContourExtractor2DImageFilterTestNamespace::MyVertexType(0.5, 8),
   itkContourExtractor2DImageFilterTestNamespace::MyVertexType(0, 7.5)
 };
-itkContourExtractor2DImageFilterTestNamespace::MyVertexListType rev_bottom_left(_rev_bottom_left, _rev_bottom_left + 2);
+static itkContourExtractor2DImageFilterTestNamespace::MyVertexListType rev_bottom_left(_rev_bottom_left,
+                                                                                       _rev_bottom_left + 2);
 
-itkContourExtractor2DImageFilterTestNamespace::MyVertexType _row2_col4[] = {
+static itkContourExtractor2DImageFilterTestNamespace::MyVertexType _row2_col4[] = {
   itkContourExtractor2DImageFilterTestNamespace::MyVertexType(11, 6.5),
   itkContourExtractor2DImageFilterTestNamespace::MyVertexType(10, 6.5),
   itkContourExtractor2DImageFilterTestNamespace::MyVertexType(9.5, 6),
@@ -74,9 +75,9 @@ itkContourExtractor2DImageFilterTestNamespace::MyVertexType _row2_col4[] = {
   itkContourExtractor2DImageFilterTestNamespace::MyVertexType(11.5, 6),
   itkContourExtractor2DImageFilterTestNamespace::MyVertexType(11, 6.5)
 };
-itkContourExtractor2DImageFilterTestNamespace::MyVertexListType row2_col4(_row2_col4, _row2_col4 + 9);
+static itkContourExtractor2DImageFilterTestNamespace::MyVertexListType row2_col4(_row2_col4, _row2_col4 + 9);
 
-itkContourExtractor2DImageFilterTestNamespace::MyVertexType _rev_row2_col4[] = {
+static itkContourExtractor2DImageFilterTestNamespace::MyVertexType _rev_row2_col4[] = {
   itkContourExtractor2DImageFilterTestNamespace::MyVertexType(11, 6.5),
   itkContourExtractor2DImageFilterTestNamespace::MyVertexType(11.5, 6),
   itkContourExtractor2DImageFilterTestNamespace::MyVertexType(11.5, 5),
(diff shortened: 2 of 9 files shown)


Examples/RegistrationITKv4/DeformableRegistration17.cxx
```diff
@@ -102,8 +102,8 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "itkDisplacementFieldTransform.h"
 
 
-unsigned int RmsCounter = 0;
-double       MaxRmsE[4] = { 0.8, 0.75, 0.4, 0.2 };
+static unsigned int RmsCounter = 0;
```
dzenanz, line 105: Maybe use the newer `inline` instead of `static`? See this [StackOverflow](https://stackoverflow.com/a/52887822/276168) answer.

seanm: I'm no C++ expert, but aren't inline variables about having external linkage? Here this global variable is not needed outside this translation unit, so external linkage is not desired.

N-Dekker: Thanks @andrei-sandor, but I think it's still preferred in C++ to place those file-scope variables in an unnamed namespace, instead of declaring them `static`:
```cpp
namespace
{
unsigned int RmsCounter = 0;
double       MaxRmsE[4] = { 0.8, 0.75, 0.4, 0.2 };
} // namespace

```

It is no longer officially deprecated to use `static` here (https://stackoverflow.com/questions/4726570/deprecation-of-the-static-keyword-no-more), but it might still cause warnings from static analysis tools, if you do so.

seanm: I don't really see how one is better/worse than the other, checking with `nm` the result is the same.  But if this is ITK's required style, we can change to that...

dzenanz (approved):
You can apply Niels' suggestion if you wish, but the proper formatting has to be followed to satisfy the CI.

Examples/RegistrationITKv4/DeformableRegistration17.cxx
blowekamp: I generally use the unnamed name space as it easily gets the correct linkage for variable, functions, and classes definitions.

seanm:
> ...proper formatting has to be followed to satisfy the CI.

Yeah, we haven't found a way to install old clang-format version 8, so we keep skipping the commit hooks.  And newer clang-format makes unrelated changes.  There's no magic "Do: reformat" command like with VTK's gitlab?

dzenanz:
Normally, ITK builds the correct version of `clang-format`, and puts it e.g. here: `C:\Dev\ITK-vs22\ClangFormat-prefix\src\ClangFormat\clang-format.exe`. You should be able to copy that executable into a more suitable location, if you like. But also pre-commit hooks should be using it?

seanm:
> Normally, ITK builds the correct version of `clang-format`, and puts it e.g. here: `C:\Dev\ITK-vs22\ClangFormat-prefix\src\ClangFormat\clang-format.exe`. You should be able to copy that executable into a more suitable location, if you like. But also pre-commit hooks should be using it?

I have `ITK_USE_CLANG_FORMAT=ON` but I don't find any such thing built. That gets built in the bin directory?, or do you have to `make install`?

```
sean@hobgoblin ITK-bin % find . -iname "*clang*"
./KWStyle/Utilities/boost/config/compiler/clang_version.hpp
./KWStyle/Utilities/boost/config/compiler/clang.hpp
./KWStyle/Utilities/boost/predef/compiler/clang.h
./ITKClangFormatConfig.cmake
```

dzenanz:
Here is the output on my Linux build:
```log
dzenan@corista:~/ITK-git-rel$ find . -iname "clang*"
./temp/clang-format-Linux
./Wrapping/Generators/CastXML/castxml/share/castxml/clang
./Wrapping/Generators/CastXML/castxml-prefix/src/castxml/share/castxml/clang
./clang-format-Linux
./ClangFormat-prefix
./ClangFormat-prefix/src/ClangFormat-stamp
./ClangFormat-prefix/src/ClangFormat-stamp/ClangFormat-update
./ClangFormat-prefix/src/ClangFormat-stamp/ClangFormat-download-err.log
./ClangFormat-prefix/src/ClangFormat-stamp/ClangFormat-download-Release-impl.cmake
./ClangFormat-prefix/src/ClangFormat-stamp/ClangFormat-patch
./ClangFormat-prefix/src/ClangFormat-stamp/ClangFormat-mkdir
./ClangFormat-prefix/src/ClangFormat-stamp/ClangFormat-done
./ClangFormat-prefix/src/ClangFormat-stamp/ClangFormat-install
./ClangFormat-prefix/src/ClangFormat-stamp/ClangFormat-download-out.log
./ClangFormat-prefix/src/ClangFormat-stamp/ClangFormat-configure
./ClangFormat-prefix/src/ClangFormat-stamp/ClangFormat-patch-info.txt
./ClangFormat-prefix/src/ClangFormat-stamp/ClangFormat-download
./ClangFormat-prefix/src/ClangFormat-stamp/ClangFormat-build
./ClangFormat-prefix/src/ClangFormat-stamp/ClangFormat-urlinfo.txt
./ClangFormat-prefix/src/ClangFormat-stamp/ClangFormat-update-info.txt
./ClangFormat-prefix/src/ClangFormat-stamp/ClangFormat-download-Release.cmake
./ClangFormat-prefix/src/ClangFormat
./ClangFormat-prefix/src/ClangFormat/clang-format
./ClangFormat-prefix/src/ClangFormat-build
./ClangFormat-prefix/tmp/ClangFormat-cfgcmd.txt
./ClangFormat-prefix/tmp/ClangFormat-mkdirs.cmake
./KWStyle/Utilities/boost/config/compiler/clang.hpp
./KWStyle/Utilities/boost/predef/compiler/clang.h
./CMakeFiles/ClangFormat.dir
./CMakeFiles/ClangFormat-complete
dzenan@corista:~/ITK-git-rel$
```

dzenanz:
Yes, it gets build in the bin directory, under name `clang-format-Linux` (see above listing). It should be similar for Mac.

seanm:
> Yes, it gets build in the bin directory, under name `clang-format-Linux` (see above listing). It should be similar for Mac.

I've just built with `VERBOSE=1 make` and tried searching for keywords like `clang` and `format` but not found any evidence of it trying to build that. Do you have an exact string I might search for to see what's (not) happening?

dzenanz:
Target is called `ClangFormat`, so maybe try `make ClangFormat`? Here are the relevant entries from my cache:
Screenshot 2024-05-17 16 52 02

seanm:
```
sean@hobgoblin ITK-big-bin % make ClangFormat
make: *** No rule to make target `ClangFormat'.  Stop.
```

But grepping `ClangFormat` in the ITK codebase I found this URL: https://data.kitware.com/api/v1/file/5d274e88877dfcc902effc47/download which appears to be the Mac clang-format8 executable!  so I just downloaded that, copied it somewhere, did `chmod +x` and then `git config clangFormat.binary /Users/sean/Downloads/clang-format8` and presto the IK git hooks are now happy!

Examples/RegistrationITKv4/DeformableRegistration17.cxx
N-Dekker: Thanks for the clang ticket @seanm! It describes the issue perfectly 💯

Modules/Video/Core/test/itkTemporalProcessObjectTest.cxx
```diff
@@ -194,7 +194,7 @@ class CallRecord
  * Static list of CallRecord items representing the stack trace of
  * calls to GenerateData and TemporalStreamingGenerateData
  */
-std::vector<CallRecord> m_CallStack;
+static std::vector<CallRecord> m_CallStack;
```
seanm, line 197: It's orthogonal to this change, but this `m_CallStack` appears to be a plain old global variable, not a class instance variable.  Should it have the `m_` prefix?

Modules/Filtering/Path/test/itkContourExtractor2DImageFilterTest.cxx
andrei-sandor, line 1: For this one, I decided to keep it static since the variable that is static is used right after and there is no static to that one. I didn't want to let inside a big namespace to complicate more. Is it okay like this or I should create a large namespace?

Modules/Video/Core/test/itkTemporalProcessObjectTest.cxx
hjmjohnson: It should not have the "m_" prefix.  Out of scope for this PR.

hjmjohnson (approved):
@andrei-sandor Minor change needed.  New static variable should have the m_ removed from the variable name.  Will produce a followup PR to remove the m_

andrei-sandor:
@hjmjohnson Yes sure, that's good for me.

PR #4679, opened 2024-05-17
COMP: Fix -Wextra-semi Warnings for Macro

andrei-sandor:
This is to see if the following changes regarding the -Wextra-semi (semi = semicolon) warning of clang can be done throughout the repository.

This is for a macro that is not function-like. It should not contain a semicolon. It is a method that is declared and defined in a header so no semi colon is needed.

Are these following changes accepted to proceed with the other cases in the repository?

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -155,6 +155,7 @@ StatementMacros:
   - INTEL_PRAGMA_WARN_PUSH
   - INTEL_PRAGMA_WARN_POP
   - INTEL_SUPPRESS_warning_1292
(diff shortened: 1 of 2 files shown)


seanm:
This is just a first example of removing the semi-colon for a non-function-like macro (as opposed to a function-like macro).

If there are no objections to this pattern, we'll apply it for other such macros.

thewtex:
:+1: looks good, thanks!

thewtex (approved):

PR #4680, opened 2024-05-17
COMP: Fix nullptr issues with nullptr or supression

andrei-sandor:
itkSmartPointerGTest.cxx has a fix with warning supressions

The other file is a simple nullptr fix.


andrei-sandor:
I will create a new PR for the fix in a new branch

PR #4682, opened 2024-05-17
COMP: remove suppression of -Winconsistent-missing-override

seanm:
I tested with AppleClang 10, the oldest AppleClang submitting to cdash, and this false positive seems to have been fixed long ago.

### DIFF
--- a/Modules/Numerics/Statistics/include/itkSubsample.h
+++ b/Modules/Numerics/Statistics/include/itkSubsample.h
@@ -75,27 +75,12 @@ public:
    * object */
   using InstanceIdentifierHolder = std::vector<InstanceIdentifier>;
 
-// Disable clang warning false positive.
-// <https://llvm.org/bugs/show_bug.cgi?id=22582>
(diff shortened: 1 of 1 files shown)


dzenanz (approved):

seanm:
@dzenanz that "ITK.Linux.Python" bot seems to have failed pretty severely, any idea why?

dzenanz:
If a run fails, it is usually the most useful to look at the [dashboard](https://open.cdash.org/index.php?project=Insight). Search for a build which contains the PR number in its name (in this case `4682`).

seanm:
> If a run fails, it is usually the most useful to look at the [dashboard](https://open.cdash.org/index.php?project=Insight). Search for a build which contains the PR number in its name (in this case `4682`).

Good tip, I've noted in it my notes file.

Seems it was a transient failure and is green now.  (There seems to have a lot of these transient failures...)

thewtex (approved):

thewtex:
@seanm thanks!

PR #4650, opened 2024-05-08
PERF: Avoid redundant searches for a data element in a `gdcm::DataSet`

N-Dekker:
The original code requested `gdcm::DataSet` to search for one and the same data element three times in a row, by calling both FindDataElement and GetDataElement for the very same tag. With this commit, the search is only performed once.

Issue found by Mihail Isakov (@issakomi):
https://github.com/InsightSoftwareConsortium/ITK/pull/4636#discussion_r1593789757

----

For the record, the redundant searches appear introduced with commit 168c457e840c3ca9e7c0b6d4df4a25252a5ddc7a, June 5, 2014: https://github.com/InsightSoftwareConsortium/ITK/blob/168c457e840c3ca9e7c0b6d4df4a25252a5ddc7a/Modules/IO/GDCM/src/itkGDCMImageIO.cxx#L384-L387

PR #4685, opened 2024-05-20
ENH: allow calling RemoveObserver() from a const itkObject

tm9k1:
Exposes a const version of `RemoveObserver() API` just like the already existing `AddObserver()` API [here](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Core/Common/src/itkObject.cxx#L467) . 

Helps with application architecture when building observer design pattern with itkObject as subject.

@mentions: @thewtex - I apologize for mentioning you without any sort of prior discussion. I'm a new contributor, and would be very grateful for some feedback on this PR!

## PR Checklist

Thanks for reviewing this change!

### DIFF
--- a/Modules/Core/Common/include/itkObject.h
+++ b/Modules/Core/Common/include/itkObject.h
@@ -201,7 +201,7 @@ public:
 
   /** Remove the observer with this tag value. */
   void
-  RemoveObserver(unsigned long tag);
+  RemoveObserver(unsigned long tag) const;
 
   /** Remove all observers . */
   void
--- a/Modules/Core/Common/src/itkObject.cxx
+++ b/Modules/Core/Common/src/itkObject.cxx
@@ -493,7 +493,7 @@ Object::GetCommand(unsigned long tag)
 }
 
 void
-Object::RemoveObserver(unsigned long tag)
+Object::RemoveObserver(unsigned long tag) const
 {
   if (this->m_SubjectImplementation)
   {
--- a/Modules/Filtering/ImageGrid/test/itkBasicArchitectureTest.cxx
+++ b/Modules/Filtering/ImageGrid/test/itkBasicArchitectureTest.cxx
@@ -21,6 +21,7 @@
 #include "itkShrinkImageFilter.h"
 #include "itkCommand.h"
 #include "itkTextOutput.h"
+#include "itkTestingMacros.h"
 
 // The following three classes are used to support callbacks
 // on the shrink filter in the pipeline that follows later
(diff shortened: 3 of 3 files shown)


dzenanz (approved):

dzenanz:
You need to fix compile errors, of course.

thewtex:
@tm9k1 thank you for the contribution. :pray:  Please also exercise the addition in the test suite.

tm9k1:
Thanks so much for the prompt review! I've added a couple lines within the basic architecture test suite to include testing of the const version of RemoveObserver. 

Given that I just wanted to contribute in-passing, I haven't looked much into the way of working of the team here beyond your feedback and a quick look over the general contribution guidelines. So, please do let me know if I missed some important requirements for the patch, including the test.

Modules/Core/Common/src/itkObject.cxx
```diff
@@ -494,6 +494,13 @@ Object::GetCommand(unsigned long tag)
 
 void
 Object::RemoveObserver(unsigned long tag)
+{
+  const auto & thisAsConst = *this;
+  return thisAsConst.RemoveObserver(tag);
```
N-Dekker, line 499: Looks fine to me. You might as well do it by one line of code, using C++17 `std::as_const`:

    return std::as_const(*this).RemoveObserver(tag);

But I see, `Object::AddObserver` also had that local variable, `thisAsConst`.  My fault: commit f2783a9aa0157b50e272545edbcad9c3029195ff  😺 (ITK did not yet use C++17, when I did that commit 🤷) Anyway, no problem.

Modules/Filtering/ImageGrid/test/itkBasicArchitectureTest.cxx
```diff
@@ -165,7 +165,11 @@ itkBasicArchitectureTest(int, char *[])
   shrink->Update();
 
   // test RemoveObserver code
-  shrink->RemoveObserver(tag);
+  shrink->RemoveObserver(tag1);
+
+  // test RemoveObserver const code
+  const auto & constShrink = *shrink;
+  constShrink.RemoveObserver(tag2);
```
N-Dekker, line 172: Would like to see an "assert" after the `RemoveObserver` call, to test the effect of the `RemoveObserver` call.

tm9k1: no worries, @N-Dekker . I'll add in the assert to get the ball rolling!

Modules/Core/Common/include/itkObject.h
```diff
@@ -202,6 +202,8 @@ class ITKCommon_EXPORT Object : public LightObject
   /** Remove the observer with this tag value. */
   void
   RemoveObserver(unsigned long tag);
+  void
+  RemoveObserver(unsigned long tag) const;
```
N-Dekker, line 206: Instead of adding an extra RemoveObserver overload, can't we simply add a `const` keyword to the existing RemoveObserver member function?

N-Dekker: I think AddObserver has a non-const overload, just for historical reasons. For RemoveObserver, it should be fine to just a const-keyword.

tm9k1: ah, just looked into the link you mentioned above: https://github.com/InsightSoftwareConsortium/ITK/commit/f2783a9aa0157b50e272545edbcad9c3029195ff#diff-0d6ffb294d56f361c4f3b5f07a0aa40f4e046e3049cb107fe009a2b16b3c1bf6R154-R158 . Thanks for sharing!

tm9k1:
for some reason the macOS build step failed after 45 minutes.. re-triggered.

Modules/Filtering/ImageGrid/test/itkBasicArchitectureTest.cxx
N-Dekker: Thanks Piyush! Your test looks fine to me now 👍 

Honestly I had to double-check the logic of `HasObserver` to see that it's OK! Technically, each `itk::StartEvent()` call creates a _new_ event object, so initially I thought that `HasObserver(itk::StartEvent())` would always return false. But now I see, `HasObserver` checks the runtime-type of its argument, so it makes sense to pass it a newly created object  👍 

By the way, for a new unit test, I usually recommend using GoogleTest, but I see, we don't have any GTest unit test yet for `itk::Object`, as there is no "itkObjectGTest.cxx".  So it's OK to me now to use the old style "itkBasicArchitectureTest.cxx". Especially now that it does Arrange, Act, Assert! 😃

N-Dekker:
> for some reason the macOS build step failed after 45 minutes.. re-triggered.

@tm9k1 Just for your information, you may also retrigger a single build by adding a comment of the form `/azp run <build machine>`. For example: https://github.com/InsightSoftwareConsortium/ITK/pull/4650#issuecomment-2101219120 But in your case: `/azp run ITK.macOS.Arm64`

Modules/Filtering/ImageGrid/test/itkBasicArchitectureTest.cxx
tm9k1: > Technically, each itk::StartEvent() call creates a new event object, so initially I thought that HasObserver(itk::StartEvent()) would always return false. But now I see, HasObserver checks the runtime-type of its argument, so it makes sense to pass it a newly created object 👍

Thanks for the deep dive, Niels. I should confess I didn't look into the impl and just went ahead according to the signature - it expected a `const EventObject &` , so sending in a new one _should be fine_ 😬

N-Dekker: You might have checked that `HasObserver(itk::StartEvent())` still returns true _before_ calling `RemoveObserver(itk::StartEvent())` (while it should return false afterwards), but I think it's good enough already now. I don't want to be over-asking  😺

Modules/Core/Common/src/itkObject.cxx
N-Dekker: Conversation resolved: this comment of mine is no longer relevant now that you removed the non-const overload. removed

Modules/Core/Common/include/itkObject.h
blowekamp: I was just surprised these are not virtual methods, as they are not it should not matter if any derived class overrode the method.

N-Dekker: This conversation is also resolved now that you removed the non-const overload. 👍

N-Dekker: @blowekamp _If_ `RemoveObserver` would have been `virtual`, adding a `const` keyword would have been a possibly breaking API change. But now that it's not virtual, adding this `const` wouldn't be a problem. That's what you mean, right?

N-Dekker:
Wow, the beautiful unit test has failed! At https://open.cdash.org/tests/1563467945

> Error in !shrink->HasObserver(itk::StartEvent())

Modules/Filtering/ImageGrid/test/itkBasicArchitectureTest.cxx
tm9k1: I've added extra checks after the `AddObserver` API calls, thanks for pointing that out! :D

tm9k1:
Thanks for looking into the issue, Niels. I can confirm that your diagnosis was correct. I now remove the AnyEvent observer and then the endEvent observer as const.

N-Dekker (approved):
> Sorry for the inconvenience!

No problem! Approved, again! 😄

N-Dekker:
@tm9k1 Thank you very much for your contribution, and for addressing our comments so carefully, Piyush!

PR #4686, opened 2024-05-21
DOC: Update .zenodo

thewtex:
@phcerdan we should have done this before the tag, but it will take effect for v5.4.1.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -19,14 +19,14 @@
       "name": "King, Brad"
     },
     {
-      "affiliation": "Mayo Clinic, Mayo Clinic Graduate School for Biomedical Sciences",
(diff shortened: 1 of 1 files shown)


.zenodo.json
```diff
@@ -45,6 +45,19 @@
     {
       "name": "Cates, Josh"
     },
+    {
+      "name": "Dekker, Niels"
```
dzenanz, line 49: @N-Dekker I think that affiliation should be added here 😄

thewtex:
@dzenanz it is generated by this script: https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/Maintenance/UpdateZenodo.py ; the order is purposeful.

phcerdan (approved):

PR #4688, opened 2024-05-21
Add and use `ProcessObject::MakeRequiredOutputs`

N-Dekker:
- Intended to supersede pull request #4687

### DIFF
--- a/Modules/Filtering/DistanceMap/include/itkDanielssonDistanceMapImageFilter.hxx
+++ b/Modules/Filtering/DistanceMap/include/itkDanielssonDistanceMapImageFilter.hxx
@@ -30,16 +30,8 @@ namespace itk
 template <typename TInputImage, typename TOutputImage, typename TVoronoiImage>
 DanielssonDistanceMapImageFilter<TInputImage, TOutputImage, TVoronoiImage>::DanielssonDistanceMapImageFilter()
 {
-  this->SetNumberOfRequiredOutputs(3);
-
-  // distance map
-  this->SetNthOutput(0, this->MakeOutput(0));
-
-  // voronoi map
-  this->SetNthOutput(1, this->MakeOutput(1));
-
-  // distance vectors
-  this->SetNthOutput(2, this->MakeOutput(2));
+  // Make the outputs (distance map, voronoi map, distance vectors).
+  ProcessObject::MakeRequiredOutputs(*this, 3);
 
   m_SquaredDistance = false;
   m_InputIsBinary = false;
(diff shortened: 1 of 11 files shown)


hjmjohnson:
@blowekamp Brad would you mind double checking that this looks OK?

Modules/Filtering/DistanceMap/include/itkDanielssonDistanceMapImageFilter.hxx
```diff
@@ -30,16 +30,8 @@ namespace itk
 template <typename TInputImage, typename TOutputImage, typename TVoronoiImage>
 DanielssonDistanceMapImageFilter<TInputImage, TOutputImage, TVoronoiImage>::DanielssonDistanceMapImageFilter()
 {
-  this->SetNumberOfRequiredOutputs(3);
-
-  // distance map
-  this->SetNthOutput(0, this->MakeOutput(0));
```
blowekamp, line 36: I see an overridden DanielssonDistanceMapImageFilter::MakeOutput. The overrides created different type depending on the idx passed. This change no longer creates the correct type in the output.

Yes the polymorphic call in a constructor is not correct. And likely should be `DanielssonDistanceMapImageFilter::MakeOutput()`

blowekamp (changes requested):
There are some cases there the correct derive MakeOutput is no longer called.

Modules/Filtering/DistanceMap/include/itkDanielssonDistanceMapImageFilter.hxx
N-Dekker: Thanks for your comments @blowekamp The function call that I'm proposing with this PR, `ProcessObject::MakeRequiredOutputs(*this, 3)` still calls `DanielssonDistanceMapImageFilter::MakeOutput` for each index `idx` = 0, 1, 2.

Looking at the proposed new static member function `ProcessObject::MakeRequiredOutputs`:

https://github.com/InsightSoftwareConsortium/ITK/blob/ddbed4adfe208e276e61b6ecbbe795c3307d1b3a/Modules/Core/Common/include/itkProcessObject.h#L931-L941

`sourceObject.TSourceObject::MakeOutput(i)` directly calls `DanielssonDistanceMapImageFilter::MakeOutput` (because `TSourceObject` is deduced as `DanielssonDistanceMapImageFilter`, in this case. So I think the behavior will remain the same as before. What do you think?

blowekamp: Thank you for the explanation. I noticed that last time I looked at this PR, but missed it this time 😁

PR #4554, opened 2024-04-02
FUTURE_LEGACY_REMOVE InternalPixelPointerType, EdgeListPointerType, InputBoundaryConditionPointerType

N-Dekker:
Deprecated the following three typedefs of raw pointer types:

    - ImageAdaptor::InternalPixelPointerType
    - QuadEdgeMesh::EdgeListPointerType
    - DiscreteGaussianImageFilter::InputBoundaryConditionPointerType

It appears clearer to just use the `Type *` syntax, to explicitly make clear when using a raw pointer, rather than a smart pointer.

These cases were found by a search of the regular expression `^  using \w+PointerType = \w+Type \*;$` in all "itk*.h" files.

- Follow-up to PR https://github.com/InsightSoftwareConsortium/ITK/pull/4538

PR #4693, opened 2024-05-23
STYLE: Remove legacy content from "itkMultiThreader.h"

N-Dekker:
"itkMultiThreader.h" was deprecated already from ITK 5.0.

### DIFF
--- a/Modules/Compatibility/Deprecated/include/itkMultiThreader.h
+++ b/Modules/Compatibility/Deprecated/include/itkMultiThreader.h
@@ -19,30 +19,6 @@
 #ifndef itkMultiThreader_h
 #define itkMultiThreader_h
 
-#if !defined(ITK_LEGACY_REMOVE)
-#  include "itkMultiThreaderBase.h"
-#  include "itkPlatformMultiThreader.h"
-namespace itk
-{
-/** Since ITK 5.0 MultiThreader has been split into a class hierarchy.
- *  Most of the time you will want to replace it by MultiThreaderBase.
- *
- *  Additionally, call this->DynamicMultiThreadingOff(); prior to Update()
- *  (for example in constructor) if any of the following is true:
- *  - Your filter needs a constant, in-advance known number of threads
- *  - Your filter uses threadId parameter in ThreadedGenerateData()
- *  - Your filter uses a custom region splitting method */
-using MultiThreader = MultiThreaderBase;
-
-
-/** Replace it by PlatformMultiThreader if any of the following is true:
- *  - Your filter uses cross-thread synchronization e.g. itkBarrier
- *  - Your filter uses MultipleMethodExecute()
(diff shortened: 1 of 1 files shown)


dzenanz:
@thewtex Is the next version number 5.5, or 6.0?

dzenanz (commented):
If we begin 6.0 development soon, this entire file should be deleted.

Until then, I would prefer to leave this file as-is.

N-Dekker:
Of course, this specific code change isn't very important to me. I'm just wondering now what we should do about all the "legacy support code" in general, now that 5.4 is tagged.

Clearly it's OK now to mark things "future legacy":
- pull request #4554 (merged by Matt after having tagged 5.4 tag)

As well as to move things from "future legacy" to "legacy only":
- pull request #4691 (also merged by Matt _after_ 5.4)

So then it seems to make sense to me to also _drop_ old "legacy only" things. But again, I'm just wondering 🤷

dzenanz:
> drop old "legacy only" things

This should be done consistently (for all legacy things), once we declare the begin of 6.0 development.

dzenanz:
Something akin to this: 01580b37063ab3eeaa60a9f18068844b4d40e9f5.

N-Dekker:
Thanks @dzenanz Does that mean that all legacy things are treated the same way? Being: let them be there until ITK 6, and then drop them all? (I can imagine that some legacy things are more bothering than others. For example, I would find it a pity if legacy support of #4691 cannot yet be dropped with ITK 6)

dzenanz:
With a new major version start, we generally remove all the code which is currently legacy, and convert all future legacy into current legacy. Which is what you already did for a few of your favorites 😄

blowekamp:
Are we going to have some kinds of "ITKv5_COMPATIBILITY" mode?

dzenanz:
Would we need anything beyond legacy enabled?

seanm:
Isn't "STYLE" inappropriate?  This is a breaking change, not something merely stylistic.

thewtex:
The next feature release is ITK 6, and it is in progress.

Removing legacy content should be ENH instead of STYLE as @seanm mentioned.

We do not need to automatically or always remove "legacy" and update "future legacy" to "legacy". This can be handled on a case-by-case basis. If the code is still widely used, there is not a clear, straightforward path to migration, and the code is not causing issues, it can be kept.

If deprecated code is removed, and entry should be created in an ITK 6 migration guide here: https://github.com/InsightSoftwareConsortium/ITK/tree/master/Documentation/docs/migration_guides

thewtex:
Clarifications and initialization started in #4696 #4695

dzenanz:
Time to close this PR and take out the deprecation removal axe? 😄

dzenanz:
Closing in favor of #4729.

Issue #3037, opened 2022-01-05
itkTestTransformGetInverse test fails under TSan


seanm:
According to bc1bcae24435fd76c430180e709a50e920eba6f4, `itkTestTransformGetInverse` was added to demonstrate a threading bug.

Seems it was never fixed though?

The test looks to me to be deliberately using `GetInverse` on a transform shared between threads.

The root of the weridness, to me, comes from [here](https://github.com/InsightSoftwareConsortium/ITK/blob/9da82ba3016b360f593f69699a92b3c19a94595b/Modules/Core/Transform/include/itkMatrixOffsetTransformBase.hxx#L482) where a getter method mutates the object.

Full TSan report:

```
==================
WARNING: ThreadSanitizer: data race (pid=18275)
  Write of size 8 at 0x7b08000588c0 by thread T2:
    #0 itk::MatrixOffsetTransformBase<double, 3u, 3u>::GetFixedParameters() const itkMatrixOffsetTransformBase.hxx:482 (ITKTransformTestDriver:x86_64+0x10003d32e)
    #1 itk::MatrixOffsetTransformBase<double, 3u, 3u>::GetInverse(itk::MatrixOffsetTransformBase<double, 3u, 3u>*) const itkMatrixOffsetTransformBase.hxx:427 (ITKTransformTestDriver:x86_64+0x1000427f4)
    #2 itk::AffineTransform<double, 3u>::GetInverse(itk::AffineTransform<double, 3u>*) const itkAffineTransform.hxx:277 (ITKTransformTestDriver:x86_64+0x100037500)
    #3 void* TestGetInverseThreadFunction<itk::AffineTransform<double, 3u> >(void*) itkTestTransformGetInverse.cxx:72 (ITKTransformTestDriver:x86_64+0x100270fcd)
    #4 decltype(std::__1::forward<void* (*&)(void*)>(fp)(std::__1::forward<itk::PoolMultiThreader::ThreadPoolInfoStruct*&>(fp0))) std::__1::__invoke<void* (*&)(void*), itk::PoolMultiThreader::ThreadPoolInfoStruct*&>(void* (*&&&)(void*), itk::PoolMultiThreader::ThreadPoolInfoStruct*&&&) type_traits:4361 (ITKTransformTestDriver:x86_64+0x1003634f7)
[...]

PR #4698, opened 2024-05-30
BUG: added 2 missing lines

seanm:


seanm:
@thewtex here are the 2 missing lines you requested in #3037.  I have no idea what this does or fixes though, so could you propose a proper commit message and I'll amend the commit...

dzenanz:
I remember there was some discussion about this not long ago. @N-Dekker @hjmjohnson

dzenanz:
This is the PR where that somewhat related discussion happened: #4485.

thewtex:
@seanm per #3037 , those lines are meant to go at the start of 

```
template <typename TScalar, unsigned int NInputDimensions,
          unsigned int NOutputDimensions>
MatrixOffsetTransformBase<TScalar, NInputDimensions, NOutputDimensions>
::MatrixOffsetTransformBase(const MatrixType & matrix, const OutputVectorType & offset)
{
// here
```

thewtex:
And it may not be correct following the discussion around:

https://github.com/InsightSoftwareConsortium/ITK/pull/4485#issuecomment-1975185951

seanm:
> And it may not be correct

So shall I just close/discard this?

N-Dekker:
> Ideally, Hans and/or Niels will review this

OK, thanks @dzenanz I'll have a look!  (I'm back 😃!)

N-Dekker:
I'm still looking for an actual relevant use case for the `MatrixOffsetTransformBase(const MatrixType &, const OutputVectorType &)` constructor. It's protected, and it isn't being used by `MatrixOffsetTransformBase::New()`, obviously. Is it being tested, directly or indirectly?

Modules/Core/Transform/include/itkMatrixOffsetTransformBase.hxx
```diff
@@ -48,6 +48,8 @@ MatrixOffsetTransformBase<TParametersValueType, VInputDimension, VOutputDimensio
   m_MatrixMTime.Modified();
   std::copy_n(offset.begin(), VOutputDimension, m_Translation.begin());
   this->ComputeMatrixParameters();
```
N-Dekker, line 50: Slightly off-topic, maybe, but the `this->ComputeMatrixParameters()` call does nothing at all: https://github.com/InsightSoftwareConsortium/ITK/blob/07f8459bdf913647dab52108d8ff284d05eb4fe4/Modules/Core/Transform/include/itkMatrixOffsetTransformBase.hxx#L678-L684

(I was expecting `ComputeMatrixParameters()` to interact with `m_FixedParameters`, but obviously, it does not!)

N-Dekker:
I think the unit test with this PR might be as follows:

```cpp
TEST(MatrixOffsetTransformBase, CreateWithMatrixAndOffset)
{
  class DerivedTransform : public itk::MatrixOffsetTransformBase<>
  {
  public:
    ITK_DISALLOW_COPY_AND_MOVE(DerivedTransform);

    static auto
    Create()
    {
      // Indirectly call the `MatrixOffsetTransformBase(const MatrixType &, const OutputVectorType &)` constructor.
      const itk::SmartPointer<DerivedTransform> ptr = new DerivedTransform(MatrixType(), OutputVectorType());
      ptr->UnRegister();
      return ptr;
    }

  private:
    // Inherit the constructors of MatrixOffsetTransformBase:
    using itk::MatrixOffsetTransformBase<>::MatrixOffsetTransformBase;
  };

  const auto                                  transform = DerivedTransform::Create();
  const DerivedTransform::FixedParametersType expectedFixedParameters(DerivedTransform::InputSpaceDimension, 0);

  EXPECT_EQ(transform->GetFixedParameters(), expectedFixedParameters);
}

```

The last line is the essential one here: `EXPECT_EQ(transform->GetFixedParameters(), expectedFixedParameters)` checks that `m_FixedParameters` has the expected number of zero-values. But moreover, it checks that `GetFixedParameters()` does not crash! Would that be helpful to you? (The test could be added to Core/Transform/test/itkMatrixOffsetTransformBaseGTest.cxx)

Modules/Core/Transform/include/itkMatrixOffsetTransformBase.hxx
N-Dekker: If this `MatrixOffsetTransformBase(const MatrixType &, const OutputVectorType &)` constructor is really meant to be used, I guess it should do the same as the default-constructor, so it should then also do `m_InverseMatrixMTime = m_MatrixMTime`, just like the default-constructor:

https://github.com/InsightSoftwareConsortium/ITK/blob/fb4b463b38640c874202808311ff6e06afb1237e/Modules/Core/Transform/include/itkMatrixOffsetTransformBase.hxx#L30-L38

(But it does not need to call `ComputeMatrixParameters()` anyway.)

hjmjohnson:
@N-Dekker Added test that you suggested.

thewtex (approved):
🙌 💯 

@seanm @N-Dekker @hjmjohnson awesome teamwork!

N-Dekker:
Honestly, if this specific `MatrixOffsetTransformBase(const MatrixType &, const OutputVectorType &)` constructor isn't really useful, I would rather see it being removed. Would be willing to sacrifice the beautiful 😸 unit test that I proposed, with this PR.

hjmjohnson:
\azp ITK.Linux

Modules/Core/Transform/include/itkMatrixOffsetTransformBase.h
```diff
@@ -507,7 +507,10 @@ class ITK_TEMPLATE_EXPORT MatrixOffsetTransformBase
    * to values specified by the caller.  If the arguments are
    * omitted, then the MatrixOffsetTransformBase is initialized to an identity
    * transformation in the appropriate number of dimensions. */
-  MatrixOffsetTransformBase(const MatrixType & matrix, const OutputVectorType & offset);
+#if !defined(ITK_LEGACY_REMOVE)
+  [[deprecated("Removed unsued constructor")]] MatrixOffsetTransformBase(const MatrixType &       matrix,
```
N-Dekker, line 511: Can you please fix this little typo here, "unsued"? Although I know it's not _that_ bad, we won't get _sued_  😸

N-Dekker:
@hjmjohnson There are a few similar unused constructors in derived classes (11!) that should also be deprecated, when `MatrixOffsetTransformBase(const MatrixType &, const OutputVectorType &)` is deprecated:

    AffineTransform(const MatrixType & matrix, const OutputVectorType & offset);
    ComposeScaleSkewVersor3DTransform(const MatrixType & matrix, const OutputVectorType & offset);
    FixedCenterOfRotationAffineTransform(const MatrixType & matrix, const OutputVectorType & offset);
    QuaternionRigidTransform(const MatrixType & matrix, const OutputVectorType & offset);
    Rigid3DTransform(const MatrixType & matrix, const OutputVectorType & offset);
    ScalableAffineTransform(const MatrixType & matrix, const OutputVectorType & offset);
    ScaleSkewVersor3DTransform(const MatrixType & matrix, const OutputVectorType & offset);
    ScaleVersor3DTransform(const MatrixType & matrix, const OutputVectorType & offset);
    Similarity3DTransform(const MatrixType & matrix, const OutputVectorType & offset);
    VersorRigid3DTransform(const MatrixType & matrix, const OutputVectorType & offset);
    VersorTransform(const MatrixType & matrix, const OutputVectorType & offset);

These 11 constructors all directly or indirectly call `MatrixOffsetTransformBase(const MatrixType &, const OutputVectorType &)`. But none of them are tested, it appears that they aren't even compiled!!! At least not at the CI. And they are all protected, so they just won't be used anywhere, I guess.

hjmjohnson:
@N-Dekker Thanks.  I have implemented your additional suggestions.

Modules/Core/Transform/include/itkMatrixOffsetTransformBase.h
N-Dekker: Thanks for fixing the typo 👍

PR #4699, opened 2024-05-30
-Wextra-semi

andrei-sandor:
This is part 2 of if the following changes regarding the -Wextra-semi (semi = semicolon) warning of clang can be done throughout the repository. https://github.com/InsightSoftwareConsortium/ITK/pull/4679 

COMP: Remove ; from itkBooleanMacro

COMP: Remove ; from other itk macros 

COMP: Updated itkLegacyMacro Definition. We decided to change the definition of itkLegacyMacro to not have two ; if the if defined(ITK_LEGACY_REMOVE) branch is hit.

COMP: Remove ITK_ASSOCIATE ; or struct;
    Should we remove the semicolon from the struct or from ITK_ASSOCIATE?

COMP: Remove ; to function cases 

COMP: Remove cases with ; to override cases 

COMP: Remove ; IPL cases

### DIFF
--- a/Modules/Core/Common/include/itkProcessObject.h
+++ b/Modules/Core/Common/include/itkProcessObject.h
@@ -299,7 +299,7 @@ public:
   itkGetConstReferenceMacro(AbortGenerateData, bool);
 
   /** \brief Turn on and off the AbortGenerateData flag. */
-  itkBooleanMacro(AbortGenerateData);
+  itkBooleanMacro(AbortGenerateData)
 
   /** \deprecated
    * Set the execution progress of a process object. The progress is
@@ -488,7 +488,7 @@ public:
    * utilization. Default value is on. */
   itkSetMacro(ReleaseDataBeforeUpdateFlag, bool);
   itkGetConstReferenceMacro(ReleaseDataBeforeUpdateFlag, bool);
-  itkBooleanMacro(ReleaseDataBeforeUpdateFlag);
+  itkBooleanMacro(ReleaseDataBeforeUpdateFlag)
 
   /** Get/Set the number of work units to create when executing. */
   itkSetClampMacro(NumberOfWorkUnits, ThreadIdType, 1, ITK_MAX_THREADS);
@@ -893,7 +893,7 @@ protected:
    * updated in derived filters.
    */
   itkGetConstMacro(ThreaderUpdateProgress, bool);
-  itkBooleanMacro(ThreaderUpdateProgress);
+  itkBooleanMacro(ThreaderUpdateProgress)
   virtual void
   SetThreaderUpdateProgress(bool arg);
 
--- a/Modules/Core/Common/include/itkProcessObject.h
+++ b/Modules/Core/Common/include/itkProcessObject.h
@@ -495,9 +495,15 @@ public:
   itkGetConstReferenceMacro(NumberOfWorkUnits, ThreadIdType);
 
 #if !defined(ITK_LEGACY_REMOVE) || defined(ITKV4_COMPATIBILITY)
-  itkLegacyMacro(void SetNumberOfThreads(ThreadIdType count)) { this->SetNumberOfWorkUnits(count); }
+  itkLegacyMacro(void SetNumberOfThreads(ThreadIdType count))
+  {
+    this->SetNumberOfWorkUnits(count);
+  }
 
-  itkLegacyMacro(ThreadIdType GetNumberOfThreads() const) { return this->GetNumberOfWorkUnits(); }
+  itkLegacyMacro(ThreadIdType GetNumberOfThreads() const)
+  {
+    return this->GetNumberOfWorkUnits();
+  }
 #endif // !ITK_LEGACY_REMOVE
 
   /** Return the multithreader used by this class. */
@@ -905,7 +911,7 @@ protected:
   progressFixedToFloat(uint32_t fixed)
   {
     return static_cast<double>(fixed) / static_cast<double>(std::numeric_limits<uint32_t>::max());
-  };
+  }
 
   /**
    * Internal method convert floating point progress [0.0, 1.0] to internal integer representation. Values outside the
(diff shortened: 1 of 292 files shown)


Modules/IO/IPL/include/itkIPLFileNameList.h
```diff
@@ -93,21 +93,14 @@ class IPLFileSortInfo
...
-  IPLSetMacroDeclaration(ImageNumber, int);
-  IPLGetMacroDeclaration(ImageNumber, int);
-  IPLSetMacroDeclaration(Data, void *);
-  IPLGetMacroDeclaration(Data, const void *);
+  IPLSetMacroDeclaration(ImageFileName, std::string) IPLGetMacroDeclaration(ImageFileName, std::string)
+    IPLSetMacroDeclaration(SliceLocation, float) IPLGetMacroDeclaration(SliceLocation, float)
+      IPLSetMacroDeclaration(SliceOffset, int) IPLGetMacroDeclaration(SliceOffset, int)
+        IPLSetMacroDeclaration(EchoNumber, int) IPLGetMacroDeclaration(EchoNumber, int)
+          IPLSetMacroDeclaration(ImageNumber, int) IPLGetMacroDeclaration(ImageNumber, int)
+            IPLSetMacroDeclaration(Data, void *) IPLGetMacroDeclaration(Data, const void *)
```
seanm, line 101: clang-format screwed this up, you need to add this macro to that list in the .clang-format file to get it to format correctly.

andrei-sandor: Yes, sorry about that. I fixed the file.

Modules/IO/IPL/include/itkIPLFileNameList.h
```diff
@@ -93,21 +93,15 @@ class IPLFileSortInfo
...
+  IPLGetMacroDeclaration(ImageFileName, std::string) IPLSetMacroDeclaration(SliceLocation, float)
+  IPLGetMacroDeclaration(SliceLocation, float) IPLSetMacroDeclaration(SliceOffset, int)
+  IPLGetMacroDeclaration(SliceOffset, int) IPLSetMacroDeclaration(EchoNumber, int)
+  IPLGetMacroDeclaration(EchoNumber, int) IPLSetMacroDeclaration(ImageNumber, int)
+  IPLGetMacroDeclaration(ImageNumber, int) IPLSetMacroDeclaration(Data, void *)
+  IPLGetMacroDeclaration(Data, const void *)
 
-private:
-  std::string  m_ImageFileName{};
+    private : std::string m_ImageFileName{};
```
seanm, line 104: Formatting wrong here too. Hopefully running clang-format will fix it, since you added these macros to `StatementMacros`.

Modules/Core/Common/include/itkProcessObject.h
```diff
@@ -893,7 +893,7 @@ class ITKCommon_EXPORT ProcessObject : public Object
    * updated in derived filters.
    */
   itkGetConstMacro(ThreaderUpdateProgress, bool);
-  itkBooleanMacro(ThreaderUpdateProgress);
+  itkBooleanMacro(ThreaderUpdateProgress)
```
hjmjohnson, line 902: Look at the ITK_MACRO_NOOP mechanism.

We want the ; colon here to keep the code consistent.  It may be that the we need to update the itkMacro.h to get rid of these.

hjmjohnson (changes requested):

Modules/Core/Common/include/itkProcessObject.h
hjmjohnson: NOTE:  ALSO ENSURE THAT "ITK_LEGACY_REMOVE" and ITK_FUTURE_LEGACY_REMOVE is turned on when running this flag for testing.

seanm: >We want the ; colon here to keep the code consistent.

You want the semi colon there? Even though you wouldn't if the macro's equivalent was written out by hand?

Example, `itkBooleanMacro(ThreaderUpdateProgress)` written manually would be:

```
  virtual void ThreaderUpdateProgressOn() { this->SetThreaderUpdateProgress(true); }
  virtual void ThreaderUpdateProgressOff() { this->SetThreaderUpdateProgress(false); }
```

which you wouldn't generally write as:

```
  virtual void ThreaderUpdateProgressOn() { this->SetThreaderUpdateProgress(true); }
  virtual void ThreaderUpdateProgressOff() { this->SetThreaderUpdateProgress(false); };
```

As this macro isn't a function-like macro, it seems odd to end it with a semi.

Modules/IO/IPL/include/itkIPLFileNameList.h
hjmjohnson: I would really prefer to NOT remove the semi-colons.  It makes clang-formatting very challenging and tedious.  Additionally, many modern IDE's complain about the lack of ; at the end of what is interpreted as a statement.

I would prefer that the macros are always terminated by a statement, or a NOOP statement as is proposed in #4706

hjmjohnson (changes requested):

Modules/IO/IPL/include/itkIPLFileNameList.h
seanm: >It makes clang-formatting very challenging and tedious.

How so?  Are you perhaps not aware of clang-format's [statementmacros](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#statementmacros) setting?

>Additionally, many modern IDE's complain about the lack of ; at the end of what is interpreted as a statement.

I've never seen that, could you give an example?

hjmjohnson:
Closing.  I think this work is superceeded with updates made in the master branch that elminiates the need to remove the ";" from macros.

seanm:
Some of the commits here are still good, but we'll make a new PR.

PR #4701, opened 2024-05-31
COMP: Test NumPy pre-releases (NumPy 2.0 RC) in CI

thewtex:
Added to the macOS, Linux Python CI builds.

Re: #4700


dzenanz:
Maybe add it to one more, e.g. Windows or Linux?

thewtex:
> Maybe add it to one more, e.g. Windows or Linux?

I added it to Linux, too.

dzenanz (approved):

Issue #4235, opened 2023-10-01
Add support for pre-commit project pre-commit hooks


thewtex:
### Description

Extend our pre-commit hook capabilities by adding support for the [pre-commit](https://pre-commit.com/) project's pre-commit hooks.

### Impact analysis

This will give use access to a number of community maintained pre-commit hooks. [List of available hooks](https://pre-commit.com/hooks.html).

A few examples that we may want to enable:

- clang-format
- cppcheck
- cpplint
- include-what-you-use
- cmake-format
- cmake-lint
- ruff
- ruff-format

### Expected behavior

The ./Utilities/SetupForDevelopment.sh script configures and enables the pre-commit project's pre-commit hooks.

### Actual behavior

We currently only have our bash-based pre-commit hooks.

### Versions

Target: ITK 6.

### Environment

Should be cross-platform.

### Additional Information

Note that this requires Python on the development system the `pre-commit` package.

Our current hooks setup could check for presence of Python or the `pre-commit` command. If so, we set up the pre-commit hook and support daisy-chaining the current hooks and the pre-commit hooks.

Then, incrementally enable pre-commit hooks we want to enable in its configuration.

thewtex:
To set up the python environment, we could possibly use [micromamba](https://mamba.readthedocs.io/en/latest/user_guide/micromamba.html).

jhlegarreta:
Applies to `ITKSphinxExamples` as well :+1:, and maybe applicable to `itk-wasm` as well.

thewtex:
Even lighter that micromamba is [uv](https://astral.sh/blog/uv).

blowekamp:
This has been done in SimpleITK:
https://github.com/SimpleITK/SimpleITK/blob/master/Utilities/GitSetup/setup-precommit
https://github.com/SimpleITK/SimpleITK/pull/2069

The approach taker where was to use the pre-commit project as the primary master of the hooks, and call the KW hooks from it.

thewtex:
> The approach taken here was to use the pre-commit project as the primary master of the hooks, and call the KW hooks from it.

Cool! We could use that + a pre-commit script to install `uv` if needed since `uv` is much faster and `python3` is not available by default in Git Bash on Windows.

thewtex:
[pixi](https://pixi.sh) is worth investigating to bootstrap access to a local python.

PR #4707, opened 2024-06-03
ENH: Remove Clang Format Linter Action

thewtex:
This never worked from fork PR pull requests due to GitHub permission
limitations. While it may be possible to fix this with recently added
GitHub Action permission features, a future replacement is planned based
on the `pre-commit` hooks, #4235 and related.


dzenanz (approved):

hjmjohnson (approved):

dzenanz:
We might want to revert this, given #4712.

PR #4714, opened 2024-06-10
COMP: Fix python wrapping build state to match python version

hjmjohnson:
If initially configured with python >= 3.11 before selecting a python version <3.11 as the python environment, then the state of the ITK_USE_PYTHON_LIMITED_API was incorrectly cached to retain the compiler settings desired for building with the original 3.11 python version.

Py_buffer is only included as part of the limited
Python API in 3.11 or greater versions.

Compilation would fail due to building with
-DPy_LIMITED_API to restrict the Python interfaces available.

## PR Checklist


thewtex (approved):
Lgtm

PR #4715, opened 2024-06-10
ENH: Remove ITKv4 compatibility support

hjmjohnson:
ITKv4 compile time API support is removed.

## PR Checklist
- [x]API changes to remove ITKv4 support that is no longer selectable (https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- x ] Added test (or behavior not changed)


CMakeLists.txt
```diff
@@ -325,20 +325,16 @@ cmake_dependent_option(
...
+
 mark_as_advanced(
   ITK_LEGACY_SILENT
   ITK_LEGACY_REMOVE
   ITK_FUTURE_LEGACY_REMOVE
-  ITKV4_COMPATIBILITY)
+  )
+
+if(ITKV4_COMPATIBILITY)
+  message(FATAL_ERROR "ITKV4_COMPATIBILITY is removed starting in ITK version 4.0, this option is no longer valid.")
```
blowekamp, line 336: I believe this should be "version 4.0" -> "version 6.0"

Modules/Core/Common/include/itkImageSource.hxx
```diff
@@ -45,11 +45,7 @@ ImageSource<TOutputImage>::ImageSource()
   this->ProcessObject::SetNumberOfRequiredOutputs(1);
   this->ProcessObject::SetNthOutput(0, output.GetPointer());
 
-#if defined(ITKV4_COMPATIBILITY)
-  m_DynamicMultiThreading = false;
-#else
   m_DynamicMultiThreading = true;
```
blowekamp, line 48: This could probably be moved to an inline initializer now.

Modules/Core/Common/src/itkPoolMultiThreader.cxx
```diff
@@ -86,12 +86,10 @@ PoolMultiThreader::PoolMultiThreader()
   }
 
   ThreadIdType defaultThreads = std::max(1u, GetGlobalDefaultNumberOfThreads());
-#if !defined(ITKV4_COMPATIBILITY)
   if (defaultThreads > 1) // one work unit for only one thread
```
blowekamp, line 89: This is not the number of threads any, it's the number of work units only?

Modules/Core/Common/src/itkTBBMultiThreader.cxx
```diff
@@ -47,14 +47,10 @@ namespace itk
 TBBMultiThreader::TBBMultiThreader()
 {
   ThreadIdType defaultThreads = std::max(1u, GetGlobalDefaultNumberOfThreads());
```
blowekamp, line 49: Is the name default threads legacy?

blowekamp (approved):
Looks good. Minor comments.

Modules/Core/Common/include/itkImageSource.hxx
hjmjohnson: Fixed in different PR.

Modules/Core/Common/src/itkPoolMultiThreader.cxx
hjmjohnson: I agree, but this is outside the scope of efforts in this PR.

PR #4719, opened 2024-06-11
STYLE: Add `constexpr` to `if (Dimension ...)` statements

N-Dekker:
Made it clearer that the condition of these `if` statements can be evaluated at compile-time.

----

Cases found by the regular expression ` if \(\w*Dimension `. Manually checked if the conditions could really be evaluated at compile-time.

### DIFF
--- a/Modules/Core/Common/include/itkCrossHelper.h
+++ b/Modules/Core/Common/include/itkCrossHelper.h
@@ -51,13 +51,13 @@ public:
   {
     VectorType oCross;
 
-    if (Dimension > 2)
+    if constexpr (Dimension > 2)
     {
       oCross[0] = iU[1] * iV[2] - iV[1] * iU[2];
       oCross[1] = iV[0] * iU[2] - iU[0] * iV[2];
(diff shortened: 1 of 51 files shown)


dzenanz (approved):

N-Dekker:
Interestingly, when having added those `constexpr` keywords, GCC appears to "discover" some "variable set but not used" cases, as it produces some relevant `-Wunused-but-set-variable` warnings that it did not produce before. At https://open.cdash.org/viewBuildError.php?type=1&buildid=9680212

> itkSobelOperator.hxx: warning: variable 'center' set but not used
> itkJensenHavrdaCharvatTsallisPointSetMetricTest.cxx: warning: variable 'metricValues2D' set but not used
> itkJensenHavrdaCharvatTsallisPointSetMetricTest.cxx: warning: variable 'metricValues3D' set but not used

Addressed by the [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/b42351a739231122d9e978d211badf060f224f99..2b4f53f48163563ba51205c7f92460dea8129282) that I just did.

hjmjohnson (approved):
Very nice!

PR #4724, opened 2024-06-14
COMP: -Wextra-semi new fix proposed

andrei-sandor:
Decided to remove semicolon from struct and not from where the function-like macro is used. Is this a fix that is acceptable? Can we proceed with similar fixes?

### DIFF
--- a/Modules/Core/Common/include/itkCellInterface.h
+++ b/Modules/Core/Common/include/itkCellInterface.h
@@ -39,7 +39,8 @@
     {                                                                                                          \
       v->VisitFromCell(cellid, this);                                                                          \
     }                                                                                                          \
-  }
+  }                                                                                                            \
+  ITK_MACROEND_NOOP_STATEMENT
 
 // Define a macro for the common type alias required by the
 // classes deriving form CellInterface (included).
--- a/Modules/Core/Common/include/itkMacro.h
+++ b/Modules/Core/Common/include/itkMacro.h
@@ -696,7 +696,7 @@ OutputWindowDisplayDebugText(const char *);
   struct newtype : public oldtype                                         \
   {                                                                       \
     char _StructPadding[mincachesize - (sizeof(oldtype) % mincachesize)]; \
-  };
+  }
 
 //
 // itkAlignedTypedef is a macro which creates a new type to make a
--- a/Modules/Core/Common/include/itkNumericTraits.h
+++ b/Modules/Core/Common/include/itkNumericTraits.h
@@ -27,7 +27,8 @@
   static constexpr ValueType min(ValueType) { return std::numeric_limits<ValueType>::min(); } \
   static constexpr ValueType max(ValueType) { return std::numeric_limits<ValueType>::max(); } \
   static constexpr ValueType min() { return std::numeric_limits<ValueType>::min(); }          \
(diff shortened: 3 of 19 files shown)


seanm:
@hjmjohnson is this pattern fine, or do you even want a superfluous `static_assert` in this kind of case?  We'd like to be sure what you'll accept before repeating the pattern.

hjmjohnson (approved):

seanm:
OK, so as this is fine, we will update this PR to apply the same pattern...

andrei-sandor:
@hjmjohnson We have committed new changes. However, we would like to know if we can proceed with adding ITK_MACROEND_NOOP_STATEMENT to macros like itkBooleanMacro. I know that in the old PR, you mention that it is not needed, but we saw some cases where the NOOP statement is not used.

Modules/Core/Common/include/itkTriangleCell.h
```diff
@@ -188,9 +188,9 @@ class ITK_TEMPLATE_EXPORT TriangleCell
   // This was observed in at least gcc 4.8 and 5.4.0, and
   // AppleClang 7.0.2 and 8.0.0. Probably others too.
   // "= default" doesn't gain us much, so just don't use it here.
-  ~TriangleCell() override{};
+  ~TriangleCell() override{}
 #else
-  ~TriangleCell() override = default;
+  ~TriangleCell() override = default
```
seanm, line 193: This ; after default should be necessary, I think you maybe removed it in error. Your compiler probably isn't getting into this #else.

Modules/Core/Transform/include/itkTransform.h
```diff
@@ -595,9 +595,9 @@ class ITK_TEMPLATE_EXPORT Transform : public TransformBaseTemplate<TParametersVa
   // This was observed in at least gcc 4.8 and 5.4.0, and
   // AppleClang 7.0.2 and 8.0.0. Probably others too.
   // "= default" doesn't gain us much, so just don't use it here.
-  ~Transform() override{};
+  ~Transform() override{}
 #else
-  ~Transform() override = default;
+  ~Transform() override = default
```
seanm, line 600: here too

hjmjohnson:
> @hjmjohnson We have committed new changes. However, we would like to know if we can proceed with adding ITK_MACROEND_NOOP_STATEMENT to macros like itkBooleanMacro. I know that in the old PR, you mention that it is not needed, but we saw some cases where the NOOP statement is not used.

Please do use the `ITK_MACROEND_NOOP_STATEMENT` at the end of 

```c++
itkBooleanMacro 
itkGPUKernelMacro
itkForLoopAssignmentMacro
itkForLoopRoundingAndAssignmentMacro
```

and others that would otherwise complain about unnecessary ';'

Thank you!

hjmjohnson:
@andrei-sandor @seanm Is this ready for review?

hjmjohnson (approved):
This is great!

seanm (approved):

seanm:
> @andrei-sandor @seanm Is this ready for review?

It is ready for merge in our opinion. There are still more `-Wextra-semi` to fix, but may as well get this next tranche in.

seanm:
Would be nice to merge this, so that it doesn't conflict again...

PR #4725, opened 2024-06-14
Default copy and move of iterators, using `ITK_DEFAULT_COPY_AND_MOVE`

N-Dekker:
Replaced hand-written copy-constructors and assignment operators of four iterators with the equivalent `ITK_DEFAULT_COPY_AND_MOVE` macro calls.

----

- Follow-up to pull request #4652

### DIFF
--- a/Modules/Core/Common/include/itkConstNeighborhoodIteratorWithOnlyIndex.h
+++ b/Modules/Core/Common/include/itkConstNeighborhoodIteratorWithOnlyIndex.h
@@ -60,6 +60,8 @@ template <typename TImage>
 class ITK_TEMPLATE_EXPORT ConstNeighborhoodIteratorWithOnlyIndex : public Neighborhood<char, TImage::ImageDimension>
 {
 public:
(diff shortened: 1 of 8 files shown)


dzenanz (approved):
Looks great on a glance!

N-Dekker:
I did look carefully at those hand-written copy-constructors and assignment operators to check that they really just do "the default thing". But honestly an extra glance wouldn't harm  😃 

Note that some copy functions of other iterators seem to do non-default things, for example: 

- the copy-constructor of ConstNeighborhoodIterator: https://github.com/InsightSoftwareConsortium/ITK/blob/efe3dce79945bd476ef3275e959c373d14e1090a/Modules/Core/Common/include/itkConstNeighborhoodIterator.hxx#L240
- the copy-constructor of ImageConstIteratorWithIndex: https://github.com/InsightSoftwareConsortium/ITK/blob/efe3dce79945bd476ef3275e959c373d14e1090a/Modules/Core/Common/include/itkImageConstIteratorWithIndex.hxx#L46

I did not include those seemingly non-default-behaving copy-functions with this pull request.

hjmjohnson (approved):

PR #4728, opened 2024-06-18
PERF: Add fast `noexcept` "move semantics" to `Array2D`

N-Dekker:
Added a `noexcept` move-constructor and move-assignment operator to `Array2D`, which both _move_ the data of the `Array2D`.

Previously, an attempt to "move" an `Array2D` would have done an expensive copy.

### DIFF
--- a/Modules/Core/Common/include/itkArray2D.h
+++ b/Modules/Core/Common/include/itkArray2D.h
@@ -60,12 +60,32 @@ public:
   /** Copy-constructor. */
   Array2D(const Self & array);
 
+  /** Move-constructor.
+   * \note This move-constructor is `noexcept`, even while the move-constructor of its base class (`vnl_matrix`) is not
+   * `noexcept`, because unlike `vnl_matrix`, `Array2D` always manages its own memory. */
+  Array2D(Self && array) noexcept
+    : vnl_matrix<TValue>(std::move(array))
+  {
+    // Note: GCC <= 9.5 does not yet support "defaulting" (`= default`) this `noexcept` move-constructor.
+  }
+
   /** Converting constructor. Implicitly converts the specified matrix to an Array2D. */
   Array2D(const VnlMatrixType & matrix);
 
   /** Copy-assignment operator. */
   Self &
   operator=(const Self & array);
+
+  /** Move-assignment operator.
+   * \note This move-assignment operator is `noexcept`, even while the move-assignment operator of its base class
+   * (`vnl_matrix`) is not `noexcept`, because unlike `vnl_matrix`, `Array2D` always manages its own memory. */
+  Self &
+  operator=(Self && array) noexcept
+  {
+    // Note: GCC <= 9.5 does not yet support "defaulting" (`= default`) this `noexcept` move-assignment operator.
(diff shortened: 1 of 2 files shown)


dzenanz (approved):

Modules/Core/Common/include/itkArray2D.h
```diff
@@ -60,13 +60,33 @@ class ITK_TEMPLATE_EXPORT Array2D : public vnl_matrix<TValue>
   /** Copy-constructor. */
   Array2D(const Self & array);
 
+  /** Move-constructor.
+   * \note This move-constructor is `noexcept`, even while the move-constructor of its base class (`vnl_matrix`) is not
+   * `noexcept`, because unlike `vnl_matrix`, `Array2D` always manages its own memory. */
+  Array2D(Self && array) noexcept
+    : vnl_matrix<TValue>(std::move(array))
```
blowekamp, line 67: What is the state of array after the move?

hjmjohnson: array is an 'r-value' reference, so it should be ok as it is a temporary object.  As such it is explicitly expected to be invalid after this call.

hjmjohnson (approved):

Modules/Core/Common/include/itkArray2D.h
N-Dekker: > What is the state of array after the move?

@blowekamp I hope that the unit tests answer this question sufficiently:
 
https://github.com/InsightSoftwareConsortium/ITK/blob/5610e233c6a6d5fc2fa9a16259c379ef41c6cdec/Modules/Core/Common/test/itkArray2DGTest.cxx#L78-L80

So after the move, the right-hand-side (the "moved-from" array) is empty.

N-Dekker: > As such it is explicitly expected to be invalid after this call.

[Just nitpicking] In this particular case, the moved-from object (the right-hand-side after the move) isn't really "invalid", it's just empty. In general, I think a moved-from object should still have a "valid" value. At least, it should be properly "destructible".

N-Dekker: So in general, I think end-users should not care too much about the state of a moved-from object. It should just be destructible!

thewtex (approved):

N-Dekker:
Example in the current code, from https://github.com/InsightSoftwareConsortium/ITK/blob/abb592882a3f4665bdfd7ff4ce835bd3862b835d/Modules/Registration/Common/include/itkMultiResolutionPyramidImageFilter.hxx#L62

```cpp
    m_Schedule = ScheduleType(m_NumberOfLevels, ImageDimension, 0);
```

`ScheduleType` is defined as `itk::Array2D<unsigned int>`, so `ScheduleType(m_NumberOfLevels, ImageDimension, 0)` yields a temporary `Array2D` object (an r-value). With this pull request, the statement above here will do a fast `noexcept` move-assignment. (Until now, it would do a rather expensive copy.)

Probably not the best example, because such a schedule object is usually quite small anyway. But you get the picture  😸

PR #4730, opened 2024-06-19
PERF: Add fast `noexcept` "move semantics" to `Array`

N-Dekker:
- Follow-up to pull request https://github.com/InsightSoftwareConsortium/ITK/pull/4728 commit 9979c7d7c9e08173fb8e0768ba806b34e71ca013
"PERF: Add fast `noexcept` "move semantics" to `Array2D`"

### DIFF
--- a/Modules/Core/Common/include/itkArray.h
+++ b/Modules/Core/Common/include/itkArray.h
@@ -62,6 +62,15 @@ public:
    *  setting for memory management. */
   Array(const Array &);
 
+  /** Move-constructor.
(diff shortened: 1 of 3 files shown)


dzenanz (approved):
Looks good on a quick glance.

N-Dekker:
It appears more complicated for `Array` than it was for `Array2D`, to implement a fast noexcept move-assignment operator. Because unlike `Array2D`, `Array` _does_ have a `m_LetArrayManageMemory` flag. While its base class `vnl_vector` _also_ has a `m_LetArrayManageMemory` flag. These two flags may have different (Boolean) values, for one and the same object. To be continued...!

PR #3470, opened 2022-06-20
ENH: Removing TransformType key from Transform serialization

PranjalSahu:
- Removed TransformType key from Transform dict.
- Added test for BSpline Transform for testing transformation output before and after serialization.
- Removed the first Composite transform from the dict array. However, there is one issue with this approach. The dictionary is a list of transforms present inside the composite transform and the name for composite transform cannot be transferred here.  We need to decide if that will be needed. 
- Changed parameter setting from loop to one function call. Construct a new itkOptimizerParameters object by converting NumPy array to a list.

## PR Checklist
- [ ] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [ ] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [X] Added test (or behavior not changed)
- [ ] Updated API documentation (or API not changed)
- [ ] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [ ] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [ ] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples) for all new major features (if any)

Refer to the [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) for
further development details if necessary.

Sample Json for single Transform object:
```yaml
[
{'name': '', 
'parametersValueType': 'float64', 
'inputDimension': 2, 
'outputDimension': 2, 
'transformName': 'BSplineTransform', 
'inputSpaceName': '2D_VectorSpace', 
'outputSpaceName': '2D_VectorSpace', 
'parameters': array([ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,
        0.,  2.,  2.,  5.,  5.,  2.,  2.,  0.,  0.,  0.,  0.,  0.,  0.,
        5.,  5., 10., 10.,  5.,  5.,  0.,  0.,  0.,  0.,  0.,  0.,  2.,
        2.,  5.,  5.,  2.,  2.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,
[...]

PR #4734, opened 2024-06-19
BUG: Move InputSpaceName, OutputSpaceName from Transform to Transform…

thewtex:
…Base

These were added to support features of the OME-Zarr transform proposals. However, they need to be available in itk::TransformBase instead of itk::Transform for access in TransformIO classes.

### DIFF
--- a/Modules/Core/Transform/include/itkTransformBase.h
+++ b/Modules/Core/Transform/include/itkTransformBase.h
@@ -90,6 +90,19 @@ public:
    *  therefore we use here a large capacity integer. */
   using NumberOfParametersType = IdentifierType;
 
+  /* For storing the  name of InputSpace */
+  itkSetMacro(InputSpaceName, std::string);
+  itkGetConstReferenceMacro(InputSpaceName, std::string);
+
+  /** For storing the  name of InputSpace/OutputSpace.
+
+  InputSpaceName, OutputSpaceName provide identifiers for the world spaces
+  that the transform applied to and the direction of the spatial transformation.
+  The direction of the transform goes from the input space to output space.
+  Typical values include the names of an atlas or a dataset. */
+  itkSetMacro(OutputSpaceName, std::string);
+  itkGetConstReferenceMacro(OutputSpaceName, std::string);
+
   /** Return the number of parameters that completely define the Transform  */
   virtual NumberOfParametersType
   GetNumberOfParameters() const = 0;
@@ -178,6 +191,10 @@ protected:
   TransformBaseTemplate() = default;
   ~TransformBaseTemplate() override = default;
 #endif
+
+private:
+  std::string m_InputSpaceName{};
+  std::string m_OutputSpaceName{};
 };
 
 /** This helps to meet backward compatibility */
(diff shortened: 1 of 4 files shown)


PranjalSahu (approved):
LGTM

N-Dekker:
Thanks Matt! It looks fine to me, to move those two data members from `itk::Transform<TParametersValueType, VInputDimension, VOutputDimension>` to `itk::TransformBase`. Especially because they don't depend on `VInputDimension` or `VOutputDimension` anyway. 👍 By the way, I would call it an `ENH`, rather than a `BUG`.

Of course I would have liked to see unit tests of the form:

```cpp
    itk::TransformBase & transformBase = ...; // Arrange
    transformBase.SetInputSpaceName(exampleSpaceName); // Act
    EXPECT_EQ(transformBase.GetInputSpaceName(), exampleSpaceName); // Assert
```

Following the AAA pattern: Arrange-Act-Assert. Maybe to consider for a follow-up....?

Modules/Core/Transform/include/itkTransformBase.h
```diff
@@ -178,6 +191,9 @@ class TransformBaseTemplate : public Object
   TransformBaseTemplate() = default;
   ~TransformBaseTemplate() override = default;
 #endif
+
+  std::string m_InputSpaceName{};
+  std::string m_OutputSpaceName{};
```
N-Dekker, line 196: Please declare them `private`!

N-Dekker (changes requested):
One little request! Please declare those two data members private!  They were originally `private` already!

blowekamp:
Any consideration for putting these into the meta-data dictionary over member variables? Would we get serialization for free with any transform IO's if we did that?

Modules/Core/Transform/include/itkTransformBase.h
N-Dekker: @hjmjohnson Why did you mark this as "resolved"?

N-Dekker:
> Any consideration for putting these into the meta-data dictionary over member variables?

I guess, the question could be, are these "space names" fundamental general properties of any transform, or are they very much application specific? (In the latter case, I guess it would be better to have them as meta-data dictionary entries, rather than member variables.)

Having said so, the decision was made already two years ago, by adding them to `itk::Transform`. So maybe it's best to just leave them as member variables now  🤷

blowekamp:
> Having said so, the decision was made already almost two years ago, by adding them to `itk::Transform`: pull request #3470 commit [0575485](https://github.com/InsightSoftwareConsortium/ITK/commit/05754850fdeba5971600d3628221cf293c7d4852). So maybe it's best to just leave them as member variables now 🤷

They could still have setter and getter member functions even if they are in the meta-data dictionary. It is more of a question if having them in the dictionary would make serialization easier.

Yes, this question is incidental to the intent of this PR.

thewtex:
Thanks for the reviews.

@N-Dekker I added the extra coverage requested.

Regarding metadata vs member, I think member is more appropriate in this case: this is an explicit property that we want all tooling / IO to eventually interpret and the explicit interface is helpful. In practice it is easier to work with, including for IO, when the explicit interface is available.

N-Dekker (approved):
Thanks for addressing my comments, Matt! Approved 👍

PR #4735, opened 2024-06-20
COMP: macOS-11 Azure CI environment EOL

hjmjohnson:
The macOS-11 environment is deprecated and scheduled to be removed on June 28th, 2024.

Update to macos-12 (also Intel) because macos-14 images are ARM and the ARM testing is done with GitHub Actions.

### DIFF
--- a/Testing/ContinuousIntegration/AzurePipelinesMacOS.yml
+++ b/Testing/ContinuousIntegration/AzurePipelinesMacOS.yml
@@ -31,7 +31,8 @@ jobs:
   timeoutInMinutes: 0
   cancelTimeoutInMinutes: 300
   pool:
-    vmImage: macos-11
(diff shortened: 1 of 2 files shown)


dzenanz (changes requested):
This changes both to `macos-13`?

hjmjohnson:
> This changes both to `macos-13`?

Yes.  Comment fixed.  macos-12 had a false positive compiler warning about linking during python wrapping.
`ld: warning: -undefined dynamic_lookup may not work with chained fixups`

macos-13 is also an intel based deployment where the compilers are fixed. 

After suggestion by @thewtex we moved from 11 to 13 to avoid a lot of CI headaches.

blowekamp (approved):
Thanks.

dzenanz (approved):

thewtex (approved):
:100:

PR #4740, opened 2024-06-21
`Array2D`, `CompensatedSummation`, `Versor`: remove hand-written copy-constructor and assignment

N-Dekker:
- Replaced the hand-written copy-constructor and copy-assignment of `Array2D` and `CompensatedSummation` with `= default`.
- Simply removed them from `Versor`, following the "Rule of Zero"

The hand-written copy-constructors were found with regular expression `^(\w+)<.+>::\1\(const Self`

### DIFF
--- a/Modules/Core/Common/include/itkArray2D.h
+++ b/Modules/Core/Common/include/itkArray2D.h
@@ -58,7 +58,7 @@ public:
   Array2D(unsigned int numberOfRows, unsigned int numberOfCols, const TValue & initialValue);
 
   /** Copy-constructor. */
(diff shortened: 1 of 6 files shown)


dzenanz (approved):

N-Dekker:
For the record, applying the "Rule of Zero" to `itk::Versor` has the interesting "side effect" that it makes an instance like `Versor<double>` a [trivial class](https://en.cppreference.com/w/cpp/language/classes#Trivial_class). This may in some cases yield a performance improvement.

PR #4752, opened 2024-06-25
PERF: Add fast "move semantics" to SpatialObjectPoint classes

N-Dekker:
Replaced hand-written copy-constructors and assignment operators by `ITK_DEFAULT_COPY_AND_MOVE` macro calls, in `SpatialObjectPoint` and its derived classes.

Will make the (compiler-generated) move-assignment operator of these classes `noexcept`, as tested.

----

- For the record, the `ITK_DEFAULT_COPY_AND_MOVE` macro was introduced with ITK 5.4, by pull request #4652


aylward (approved):
Nice update to make these classes more ITK-ish!

hjmjohnson (approved):

PR #4754, opened 2024-06-25
COMP: Consolodate compiler version ICE work-arounds

hjmjohnson:
Implement compiler version-specific conditional code selection to facilitate identifying which compilers fail to compile the valid code.

This is an intermediate refactoring so that once
all the supported compilers work with the simpler
code, we can remove the conditional code.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkMacro.h
+++ b/Modules/Core/Common/include/itkMacro.h
@@ -1387,6 +1387,17 @@ itkDynamicCastInDebugMode(TSource x)
 #endif
 }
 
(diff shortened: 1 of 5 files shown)


seanm:
It so happens I was working on a similar cleanup of old compiler workarounds. I should share it here in a PR. I was waiting to test it on my oldest bot, with oldest AppleClang...

I think this workaround can just be removed at this point... I'll be back shortly...

hjmjohnson:
@seanm is working on a more complete solution.

PR #4757, opened 2024-06-25
Fix algorithm usage

hjmjohnson:
ENH: Fix algorithm to better match data types

Update stl algorithm based on suggestions which explictly warns against the anti-pattern used for toupper and tolower string manipulations.

https://en.cppreference.com/w/cpp/string/byte/toupper with std algorithms

## PR Checklist

### DIFF
--- a/Modules/IO/XML/src/itkStringTools.cxx
+++ b/Modules/IO/XML/src/itkStringTools.cxx
@@ -18,7 +18,7 @@
 
 #include "itkStringTools.h"
 
-#include <algorithm> // std::transform()
+#include <algorithm> // std::for_each()
 #include <cctype>    // std::toupper(), std::tolower()
 
 namespace itk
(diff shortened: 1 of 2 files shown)


Modules/IO/XML/src/itkStringTools.cxx
```diff
@@ -85,17 +84,17 @@ StringTools::TrimRight(std::string & str, const std::string & dislike)
 std::string &
 StringTools::ToUpperCase(std::string & str)
 {
-  // explicit cast needed to resolve ambiguity
-  std::transform(str.begin(), str.end(), str.begin(), (int (*)(int))std::toupper);
+  // https://en.cppreference.com/w/cpp/string/byte/toupper with std algorithms
+  std::for_each(str.begin(), str.end(), [](unsigned char c) -> int { return std::tolower(c); });
```
dzenanz, line 88: This should be `toupper`, not `tolower`.

dzenanz (commented):
This second commit has many style changes, it would be good if they were separated out into their own commit.

dzenanz (approved):

thewtex (approved):

dzenanz (approved):

PR #4760, opened 2024-06-26
Remove/cleanup old compiler workarounds

seanm:


hjmjohnson (approved):

Modules/Core/Common/include/itkMacro.h
```diff
@@ -175,38 +175,38 @@ namespace itk
...
+#  error "The SGI compiler is not supported"
 #endif
 #if defined(__APPLE__)
 #  if defined(__clang__) && (__cplusplus < 201703L)
-#    error "Apple LLVM compiling with a standard less than C++17 is not supported under ITKv5.4"
+#    error "Apple LLVM compiling with a standard less than C++17 is not supported"
 #  endif
 #elif defined(__clang__) && (__clang_major__ < 5)
-#  error "Clang < 5 is not supported under ITKv5.4"
+#  error "Clang < 5 is not supported"
```
dzenanz, line 206: Isn't clang 7 the minimum now?

dzenanz (approved):

Modules/Core/Common/include/itkMacro.h
dzenanz: Or is that AppleClang?

dzenanz: @N-Dekker should we update this number? We might need to update it elsewhere, too.

dzenanz: Possible update can be done in a follow-up PR.

seanm: Ideally, we should have cdash submissions with the oldest of each supported compiler.

hjmjohnson: All good comments, but I don't see anything that needs immediate action.  I will resolve this issue and merge these valuable improvements as a good starting point.

Modules/Core/Common/include/itkImageRegion.h
```diff
@@ -424,11 +424,11 @@ operator<<(std::ostream & os, const ImageRegion<VImageDimension> & region);
...
-#if defined(__clang__)
-#  pragma GCC diagnostic push
-// Old Clang compiler versions (before Clang 7.0.0) produced some unimportant warnings, like: "warning: 'tuple_size'
-// defined as a struct template here but previously declared as a class template"
-#  pragma GCC diagnostic ignored "-Wmismatched-tags"
+#if defined(__clang__) && defined(__apple_build_version__) && (__clang_major__ <= 10)
+#  pragma clang diagnostic push
+// Old AppleClang 10.0.0 (Xcode 10.1, newest on macOS 10.13) produced some unimportant warnings, like:
+// "warning: 'tuple_size' defined as a struct template here but previously declared as a class template"
+#  pragma clang diagnostic ignored "-Wmismatched-tags"
```
N-Dekker, line 431: @seanm With this commit, the warning is _only_ disabled on Apple. So with this commit, it is no longer disabled on other platforms (while running clang). Right? Is that intended?

It seems to me that the warning would also occur on Linux (running LLVM/Clang): https://discourse.llvm.org/t/lots-of-warnings-building/69731

seanm: The comment, and the commit message that introduced it, referred to warnings on a Mac, so it seemed to me this was referring to AppleClang.

Clang and AppleClang have different version numbers, so trying to handle both at once (if that's what it was trying to do), is incorrect.

I presume the warning could occur with open source Clang too.  It doesn't on github CI though.  If anyone can repro, we can know what versions to suppress.

seanm: do you have a snippet that repros the warning?  Could then find the compiler version with https://godbolt.org/ ...

N-Dekker: @seanm Sorry, I haven't been able to reproduce the warning myself anymore 🤷  It should be something like:

```cpp
#include <type_traits>
#include <utility>

class MyClass{};

namespace std
{
template <>
struct tuple_size<MyClass> : integral_constant<size_t, 1>
{};    
}

```

But to my surprise, it does not appear at https://godbolt.org/z/c95jeGz6x running clang 6.0.1 `-Wmismatched-tags -std=c++17`.

seanm: So really we shouldn't even be testing the compiler, but the std library version.

Your example does work on my old Mac BTW:

```
builder22:~ builder$ xcrun clang --version
Apple LLVM version 10.0.0 (clang-1000.11.45.5)
Target: x86_64-apple-darwin17.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
builder22:~ builder$ 
builder22:~ builder$ xcrun clang -fsyntax-only -Weverything ~/Desktop/test.cxx 
/Users/builder/Desktop/test.cxx:12:5: warning: 'tuple_size' defined as a struct template here but previously declared as a class template
      [-Wmismatched-tags]
    struct tuple_size<MyClass> : integral_constant<size_t, 1>
    ^
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/__tuple:25:22: note: did you mean struct here?
template <class _Tp> class _LIBCPP_TEMPLATE_VIS tuple_size;
                     ^
1 warning generated.
```

N-Dekker: > So really we shouldn't even be testing the compiler, but the std library version.

Indeed! Well, actually it seems like a combination of the std library version and the compiler version! Suppose we do the MyClass specialization by `class tuple_size<MyClass>` instead: https://godbolt.org/z/EqEf4cdx3 Then Clang (including Clang version 7.1.0) may produce a similar warning (but having the terms "struct" and "class" reverted): 

```
<source>:12:1: warning: 'tuple_size' defined as a class template here but previously declared as a struct template [-Wmismatched-tags]
class tuple_size<MyClass> : integral_constant<size_t, 1>
^
/opt/compiler-explorer/gcc-9.3.0/lib/gcc/x86_64-linux-gnu/9.3.0/../../../../include/c++/9.3.0/utility:84:5: note: did you mean class here?
    struct tuple_size;
    ^
```
However, this warning is no longer there with Clang version >= 8.0.0!

PR #4765, opened 2024-06-29
DOC: Add note about ITK release and remote module compatibility

jhlegarreta:
Add a note about the need to update the remote modules when a new ITK mjor/minor version is released.

## PR Checklist

### DIFF
--- a/Documentation/Maintenance/Release.md
+++ b/Documentation/Maintenance/Release.md
@@ -398,6 +398,11 @@ git push upstream $version
 Add any new remote modules to nightly builds. Some builds may be difficult to add
 due to third-party dependencies.
 
+When a new ITK version is released, experience has shown that the remote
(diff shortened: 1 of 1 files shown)


hjmjohnson (approved):

SimonRit:
I don't understand why remote modules should be bumped into a new major version for a minor version change of ITK. That wouldn't be my choice by default... Can you clarify why incompatibility of the binaries across ITK minor versions requires this?

thewtex:
While we attempt to maintain API compatibility between ITK minor releases, the ABI generally changes. This can be due to features and bug fixes in the ITK source code, but there are also usually build system changes that cause ABI changes.

thewtex (approved):

blowekamp:
Why is this preferred over more specific pinning of the package install requirements to something like "itk~=5.4.0"?

PR #4766, opened 2024-07-01
COMP: Set CastXML flag based on CMAKE_CXX_STANDARD

thewtex:
We may bump the standard version in the future or remote modules may use a different standard version.


dzenanz (approved):
How did you discover this?

thewtex:
> How did you run into this?

Using C++20 with ITK-Wasm.

PR #4767, opened 2024-07-02
BUG: Propogate MetaIOVersion to child objects in scene I/O

aylward:
The ITK I/O methods for SpatialObjectRead and SpatialObjectWrite have functions for setting the MetaIO version to use when reading/writing files.  See the MetaIO commit: https://github.com/Kitware/MetaIO/commit/eb952d5b2dbbccc27a99cc1b4cb27410f5267981

However, in multi-object MetaIO files (i.e., metaScene files) the user specification of the MetaIO version was not propagated to the readers used for the components objects of the file - resulting in warnings and errors.

This update and a corresponding MetaIO update (https://github.com/Kitware/MetaIO/pull/129) address this error.

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkMetaSceneConverter.h
+++ b/Modules/Core/SpatialObjects/include/itkMetaSceneConverter.h
@@ -144,6 +144,7 @@ private:
   {
     auto converter = TConverter::New();
     // needed just for Image & ImageMask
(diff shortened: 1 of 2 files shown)


dzenanz (approved):
Looks good on a glance.

Should MetaIO be updated before this is merged?

Issue #4748, opened 2024-06-23
Transition remote modules configuration from `setup.py` to the modern `pyproject.toml`


jhlegarreta:
### Description

Transition remote modules configuration from `setup.py` to the modern `pyproject.toml`.

### Impact analysis

`pyproject.toml` is the modern way to package projects using Python code:
https://packaging.python.org/en/latest/guides/writing-pyproject-toml/
https://packaging.python.org/en/latest/guides/modernize-setup-py-project/#modernize-setup-py-project

### Expected behavior

Transition all remote module's project configuration to a `pyproject.toml` file.

### Actual behavior

Almost all remote module's project configuration dwells in a `setup.py` file.

### Versions

N/A.

### Environment

N/A.

### Additional Information

Matt has contributed with a script that would help in the transition:
https://github.com/InsightSoftwareConsortium/ITKModuleTemplate/pull/161

Also, there is this tool Matt has used to update the remote module CI config to ITK 5.4:
https://github.com/asottile/all-repos

Ongoing work:
https://github.com/thewtex/ITK/commit/9e9d5131dcde8f46505c838218df68ce24701cd0

Remotes and PRs:

TODO:

- [ ] https://github.com/SCIInstitute/ITKCleaver: https://github.com/SCIInstitute/ITKCleaver/pull/56
- [ ] https://github.com/InsightSoftwareConsortium/scifio-imageio : Forked from scifio, the latter being ahead: needs to be managed

DONE:
- [x] https://github.com/KitwareMedical/ITKHigherOrderAccurateGradient: Duplicate of ISC, the latter forked from KitwareMedical and being ahead; needs to be managed - Deleted 2025-02-03
- [x] https://github.com/KitwareMedical/ITKContourInterpolation: PR https://github.com/KitwareMedical/ITKContourInterpolation/pull/122
- [x] https://github.com/KitwareMedical/ITKUltrasound PR https://github.com/KitwareMedical/ITKUltrasound/pull/239
- [x] https://github.com/KitwareMedical/MultipleImageIterator: No Python packaging
- [x] https://github.com/KitwareMedical/ITKIOScanco: https://github.com/KitwareMedical/ITKIOScanco/pull/85
- [x] https://github.com/InsightSoftwareConsortium/ITKLabelErodeDilate: https://github.com/InsightSoftwareConsortium/ITKLabelErodeDilate/pull/39
[...]

PR #4769, opened 2024-07-04
DOC: Clarify further remote module version bumps with ITK minor releases

jhlegarreta:
Clarify further remote module major version bumps with ITK minor releases.

## PR Checklist
- [X] No [major design changes]

### DIFF
--- a/Documentation/Maintenance/Release.md
+++ b/Documentation/Maintenance/Release.md
@@ -414,9 +414,12 @@ due to third-party dependencies.
 
 When a new ITK version is released, experience has shown that the remote
 module packages are not binary compatible across ITK major or minor releases.
-Thus, remote module versions should be bumped into a new major version when
-requiring a new ITK major or minor version. The process described below
-ensures appropriate versioning and compatibility of the remotes with ITK.
+While we attempt to maintain API compatibility between ITK minor releases, the
+ABI generally changes. This can be due to features and bug fixes in the ITK
+source code, but there are also usually build system changes that cause ABI
+changes. Thus, remote module versions should be bumped into a new major
+version when requiring a new ITK major or minor version. The process described
+below ensures appropriate versioning and compatibility of the remotes with ITK.
 
 Update Remote Modules
 ---------------------


jhlegarreta:
@SimonRit following the thread in PR #4765 and Matt's clarification.

dzenanz (approved):

blowekamp:
Why is this preferred over more specific pinning of the package install requirements to something like "itk~=5.4.0"?

SimonRit:
Thanks for integrating Matt's comment to the documentation. I still don't really understand why the major version of the remote module should be bumped but not the major version of ITK but this is most likely due to a lack of knowledge on my side.

blowekamp:
My suspicion is that this line is not correct:
https://github.com/InsightSoftwareConsortium/ITKIsotropicWavelets/blob/master/setup.py#L57

and should be something like "itk~=5.4.0" to only allow the ITK patch version to very.

jhlegarreta:
> My suspicion is that this line is not correct:
> https://github.com/InsightSoftwareConsortium/ITKIsotropicWavelets/blob/master/setup.py#L57
> and should be something like "itk~=5.4.0" to only allow the ITK patch version to very.

https://github.com/InsightSoftwareConsortium/ITKIsotropicWavelets/pull/159 and https://github.com/InsightSoftwareConsortium/ITK/issues/4748#issuecomment-2189851985.

Documentation/Maintenance/Release.md
```diff
@@ -414,9 +414,12 @@ due to third-party dependencies.
 
 When a new ITK version is released, experience has shown that the remote
 module packages are not binary compatible across ITK major or minor releases.
-Thus, remote module versions should be bumped into a new major version when
-requiring a new ITK major or minor version. The process described below
-ensures appropriate versioning and compatibility of the remotes with ITK.
+While we attempt to maintain API compatibility between ITK minor releases, the
+ABI generally changes. This can be due to features and bug fixes in the ITK
+source code, but there are also usually build system changes that cause ABI
+changes. Thus, remote module versions should be bumped into a new major
```
blowekamp, line 420: So, While ITK's main Python packages do not follow semantic versioning, it is recommended that remote module bump the major version when the minor version of ITK change to follow semantic versioning best practices.

Is that clearer

thewtex:
Perhaps it is clearer to have remote module package versions follow their own semantic versioning path, i.e. bump the minor version when bumping the `itk` version they build against ?. However, have the dependency on `itk-X.X.*` indicate the ABI compatibility in this explanation.

SimonRit:
> Perhaps it is clearer to have remote module package versions follow their own semantic versioning path, i.e. bump the minor version when bumping the `itk` version they build against ?. However, have the dependency on `itk-X.X.*` indicate the ABI compatibility in this explanation.

This has my preference but it might be because I didn't get the explanation...

thewtex:
I initially thought a remote module update should be considered a breaking change because it forces an `itk` version update, but this seems too stringent.

jhlegarreta:
Folks, please, go ahead and edit the change and push force so that it reflects better what the policy should be so that the policy and the underlying rationale are clear onwards.

jhlegarreta:
The below will need to be changed in accordance to what is adopted here:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/Maintenance/UpdateRequiredITKVersionInRemoteModules.sh#L26-L27

as well as the logic in the script.

PR #4775, opened 2024-07-12
ENH: Provide GetSpacing/GetOrigin calls that wrap for python

aylward:
itk::SpatialObjectToImageFilter provided GetSpacing() and GetOrigin() calls that returned double * objects.  That return type is not well supported by SWIG for python wrapping.

Herein we provide GetSpacingVector() and GetOriginPoint() functions that return VectorType and PointType objects from the itk::Image class definition which are nicely wrapped by SWIG for python.

See: https://discourse.itk.org/t/image-spacing-value-interactions-in-itk-python-wrapping/6707


Modules/Core/SpatialObjects/include/itkSpatialObjectToImageFilter.hxx
```diff
@@ -164,6 +166,17 @@ SpatialObjectToImageFilter<TInputSpatialObject, TOutputImage>::GetSpacing() cons
   return m_Spacing;
 }
 
+template <typename TInputSpatialObject, typename TOutputImage>
+const typename TOutputImage::SpacingType &
+SpatialObjectToImageFilter<TInputSpatialObject, TOutputImage>::GetSpacingVector() const
+{
+  for (unsigned int i = 0; i < OutputImageDimension; ++i)
+  {
+    m_SpacingVector[i] = m_Spacing[i];
```
dzenanz, line 175: If this assignment is done in the getter, what is the point of also having it as an ivar? This could be a local variable in the getter instead. Same remark for the other one.

dzenanz (changes requested):

Modules/Core/SpatialObjects/include/itkSpatialObjectToImageFilter.hxx
aylward: I was trying to avoid returning a reference to a local var and to make the minimal changes to the code to preserve backward compatibility.   What do you recommend?

dzenanz: Aha, so we can't return spacing by value? If so, this is fine.

dzenanz (approved):
If we add an extra ivar, we might want to print it in `PrintSelf` method.

aylward:
> If we add an extra ivar, we might want to print it in `PrintSelf` method.

Great point - will amend this commit!

Modules/Core/SpatialObjects/include/itkSpatialObjectToImageFilter.hxx
aylward: Yeah - this is still using a very old (the original) ITK API where spacing was double * - which doesn't wrap for python, hence this PR.  For ITK v6.0, I'd like to do a major update to SpatialObjects to follow ITK's current API, but it will break backward compatibility.

thewtex (approved):
Nice!

:+1: for making breaking changes to improve the API with ITK 6

Issue #4788, opened 2024-07-24
ITK's SpatialOrientation and OrientImageFilter have opposite definitions of the DICOM standard


blowekamp:
The definitions of the coordinate terms in SpatialOrientationEnums is as follows:
```
Coordinate orientation codes have a place-value organization such that an ImageDimension-al sequence of subcodes says both which varies fastest through which varies slowest, but also which end of the frame of reference is considered zero for each of the coordinates. For example, 'RIP' means Right to Left varies fastest, then Inferior to Superior, and Posterior to Anterior varies the slowest.
```

This is the opposite of the definitions for the DICOM standard:
https://www.slicer.org/wiki/Coordinate_systems

It is interpreted that the preferred way to describe coordinates is in the "TO" direction or the direction of an axes is increasing. See the following DICOM section: C.7.6.2.1.1

ITK is defined as having a LPS coordinate system. However the SpatialOrientation enums and the OrientImageFilters call the RAI coordinates. This is confusing.

UPDATED:

I propose moving `itk::OrientImageFilter` to the deprecated module. Users can using this filter by simply enabling this module as it will be available until least ITK 7.0 ( unplanned ).

The itk::SpatialOrientationEnums and itk::SpatialOrientationAdapters usage should be discouraged. Usage where appropriate will be marked with "ITK_FUTURE_LEGACY_REMOVE", which provides no clear time frame for removal.

The core improvement is the introduction of the "DICOMOrientation" class which contains enum classes to describe spatial orientation as interpreted from DICOM with ITK having "LPS" coordinate. The DICOMOrientation class include the functionality of the legacy SpatialOrientationAdaptor, but with the addition of interpreting string representations. Additionally, this new class provides compatibly with conversion from the legacy SpatialOrientationEnums.

 The [DICOMOrientImageFilter](https://github.com/InsightSoftwareConsortium/ITKSimpleITKFilters/blob/master/include/) from the SimpleITKFilters remote module is introduced to the "filter user" front end api, to clarify the change to the new DICOM based coordinate descriptions.

blowekamp:
@dzenanz @lassoan @thewtex Care to chime in on this proposal?

dzenanz:
I like it, but there is a lot work there. We should strive to keep the functionality on the same level, including Python accessibility. Also, there are a few examples which should be updated to demonstrate the new recommended syntax and filter names. If you are willing to do this, I am all for it!

dzenanz:
[...]

PR #4790, opened 2024-07-29
WIP: Deprecate OrientImageFilter

blowekamp:
The anatomical direction enums used by this class are the oposite of the DICOM specification. Depricated to be replaced by the DICOMOrientImageFilter for consistent orientation notation.

Relates to #4788

## PR Checklist


seanm:
Isn't `OrientImageFilter` a general class?  Why should it be tied to DICOM?

Looking at my own code, I see we use `OrientImageFilter` when reading MINC and NIfTI files too...

issakomi:
> The anatomical direction enums used by this class are the oposite of the DICOM specification

Could you, please, send a link where 3-letter enums are defined in DICOM standard?

I think it is about terminology (`from` vs `to` notation), ITK space is the same as DICOM space.

RAI-codes are widely used, e.g. in MetaIO headers, in ITK-Snap, I use them too.
I know that Slicer uses LPS-terminology (but, AFAIK, works perfectly with actual orientation filter and enums) and in SimpleITK already exists [DICOMOrientImageFilter](https://simpleitk.org/doxygen/latest/html/classitk_1_1simple_1_1DICOMOrientImageFilter.html) with LPS notation.

MetaImage header:
meta

ITK-Snap:
snap

My app (I heavily use actual enum internally):
aliza

gdevenyi:
Perhaps this is a side discussion but @issakomi I would argue that both those screenshots incorrectly describe what the orientation means.

I think it is far more clear to describe these orientations as describing the positive direction, i.e. RAS = +X+Y+Z, the direction of increasing coordinate (index) value is Right(X), Anterior(Y), Superior(Z)

http://www.grahamwideman.com/gw/brain/orientation/orientterms.htm

seanm:
I have observed that some people understand RAI to mean the origin is at the right-anterior-inferior corner, and others understand it to mean that the axes point *towards* the right, anterior, and inferior.  :(

issakomi:
> Perhaps this is a side discussion but @issakomi I would argue that both those screenshots incorrectly describe what the orientation means.

I didn't introduced these enums. It is breaking change. I am out of the discussion. Do whatever you want.

P.S. BTW, your MINC IO is completely brocken and loads with flipped axes.

gdevenyi:
> > Perhaps this is a side discussion but @issakomi I would argue that both those screenshots incorrectly describe what the orientation means.
> 
> I didn't introduced these enums. It is breaking change. I am out of the discussion. Do whatever you want.
> 
> P.S. BTW, your MINC IO is completely brocken and loads with flipped axes.

Not mine. And I'm well aware: https://github.com/InsightSoftwareConsortium/ITK/pull/147

issakomi:
> Not mine. And I'm well aware: #147

OK

hjmjohnson:
The genesis of the OrientImageFilter was to assist with dealing with Analyze (pre-nifti) naming and orientation.  The publicly available documentation available was often fragmented and incomplete.  

I would prefer to keep this OrientImageFilter for historical reasons, and because some users still depend upon it.  Improving the documentation, and adding a documentation warning would be my preferred approach.  

I would also be supportive of minimizing the use of this class internal to ITK.

blowekamp:
> I would prefer to keep this OrientImageFilter for historical reasons, and because some users still depend upon it. Improving the documentation, and adding a documentation warning would be my preferred approach.
> 
> I would also be supportive of minimizing the use of this class internal to ITK.

Please also see: PR #4791

The proposal to to replace it with the DICOMOrientImageFilter and the DICOMOrientation enums and keep legacy conversion from the SpatialOrientation enums to the DICOMOrientation class ( which has build it conversion between ITK direction matrix, enums, legacy enums, and strings (DICOM).

By moving the class to deprecated it allows users to migrate in the next 5+ years. And the tricky part of chasing the enums to the DICOM part has an even longer time period.

It is incredible confusing having documentation describing ITK as having DICOM compatible LPS coordinate systems, but the build in tools describe it as "RAI". I am hoping to use ITKv6 as an opportunity to improve the situation.

issakomi:
> I think it is far more clear to describe these orientations as describing the positive direction, i.e. RAS = +X+Y+Z, the direction of increasing coordinate (index) value is Right(X), Anterior(Y), Superior(Z)

@gdevenyi Thanks. I have added actually both for the next release and there is (was) explanation "right. to left", etc.

o2

I am afraid v5 will be the last version I use, for a long time.

seanm:
>The proposal to to replace it with the DICOMOrientImageFilter and the DICOMOrientation enums

I'm still not understanding what DICOM has to do with anything...

I've just read over itkOrientImageFilter.h and I don't see anything about DICOM.  It looks like a general structured grid reorientation class to me. Such a thing is useful. I use it in code that doesn't deal with DICOM files at all.

blowekamp:
Unfortunately, this conversation has been split over 3 topics. Here issue #4788 and PR #4791. I am closing this PR, to help focus the conversation on issue #4788.

dzenanz:
> I'm still not understanding what DICOM has to do with anything...

ITK used "origin" codes (RAI) instead of "to" codes (LPS). The "to" codes are used by DICOM standard, and are more widespread. To distinguish the "to" codes from "origin" codes, the "to" convention needs some name. In the other PR, Brad suggested that name be "DICOM", as in DICOMOrientation etc. Of course, orientation is general, and most medical image formats support orientation metadata, not just DICOM image format. Hopefully this explanation is helpful.

PR #3144, opened 2022-01-25
STYLE: Change `constexpr static` to `static constexpr`.

Leengit:
This pull request changes the 23 instances of `constexpr static` to be like the 2322 instances of `static constexpr`.
```bash
cd ITK
fgrep -Rl 'constexpr static' *
  | egrep '\.(c|h|cc)(xx|pp|)$'
  | fgrep -v ThirdParty
  | xargs sed -i -r \
    -e 's/([^A-Za-z0-9_])constexpr static([^A-Za-z0-9_])/\1static constexpr\2/g'
```
## PR Checklist
- [X] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [X] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [X] Added test (or behavior not changed)
- [X] Updated API documentation (or API not changed)
- [X] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [X] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [X] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples) for all new major features (if any)

PR #4791, opened 2024-07-31
ENH: Add DICOMOrientation class from remote module

blowekamp:
The class provides anatomical orientation based enums for describing coordinate systems. In ITK the physical coordinate systems is the same as the DICOM standard. DICOM has the following specification:

If Anatomical Orientation Type (0010,2210) is absent or has a value of BIPED, the x-axis is increasing to the left hand side of the patient. The y-axis is increasing to the posterior side of the patient. The z-axis is increasing toward the head of the patient.

Which describes the direction of increasing. The above coordinate system is described as "LPS" with this enumeration and corresponds to an ITK image with an identity direction cosine matrix.

See also #4788, #4788

## PR Checklist


Modules/Core/Common/include/itkDICOMOrientation.h
```diff
@@ -0,0 +1,274 @@
...
+
+  CoordinateEnum
+  GetCoordinateTerm(CoordinateMajornessTermsEnum cmt) const;
+
+public:
+#if 0
+// When this is enabled the following error occurs:
+//ITK/Modules/Core/Common/include/itkDICOMOrientation.h:93:11: error: enumerator value is not a constant expression
+//    RIP = ITK_ORIENTATIONENUM(CoordinateEnum::Right, CoordinateEnum::Inferior, CoordinateEnum::Posterior),
+
```
blowekamp, line 88: @N-Dekker Any idea why the function does not work for setting enum values?

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -1886,32 +1885,46 @@ NiftiImageIO::SetImageIOOrientationFromNIfTI(unsigned short dims, double spacing
         this->GetLegacyAnalyze75Mode() != NiftiImageIOEnums::Analyze75Flavor::AnalyzeITK4 &&
         this->GetLegacyAnalyze75Mode() != NiftiImageIOEnums::Analyze75Flavor::AnalyzeITK4Warning)
     { // only do this for Analyze file format
-      SpatialOrientationAdapter::OrientationType orient;
+      DICOMOrientation orient(DICOMOrientation::OrientationEnum::INVALID);
       switch (this->m_NiftiImage->analyze75_orient)
       {
         case a75_transverse_unflipped:
-          orient = SpatialOrientationEnums::ValidCoordinateOrientations::ITK_COORDINATE_ORIENTATION_RPI;
+          orient = DICOMOrientation(DICOMOrientation::CoordinateEnum::RightToLeft,
```
dzenanz, line 1892: In these long and and long-ish statements, it might make sense to `using CE = DICOMOrientation::CoordinateEnum`, then use `CE::RightToLeft, CE::PosteriorToAnterior, CE::InferiorToSuperior` on a single line. This applies to many places in the code, including previous commits.

dzenanz (commented):
Looking good so far.

dzenanz:
One warning, [Modules/IO/GE/src/itkGE5ImageIO.cxx:191](https://github.com/InsightSoftwareConsortium/ITK/blob/4d8bf5c1ed2201da5695c85bab01863d07b3089b/Modules/IO/GE/src/itkGE5ImageIO.cxx#L191):44: warning: 'void* memset(void*, int, size_t)' clearing an object of non-trivial type 'struct GEImageHeader'; use assignment or value-initialization instead [-Wclass-memaccess]

issakomi:
> In ITK the physical coordinate systems is the same as the DICOM standard.

It is not the same.
DICOM defines only right-handed [orientations](https://dicom.innolitics.com/ciods/rt-dose/image-plane/00200037). "Image Orientation (Patient) Attribute" defines "x" and "y" (6 values), "z" is _right-handed_ cross-product.

`The Patient-Based Coordinate System is a right handed system, i.e., the vector cross product of a unit vector along the positive x-axis and a unit vector along the positive y-axis is equal to a unit vector along the positive z-axis.`

ITK defines  "x", "y" **and** "z", so e.g.
```
-1  0  0
 0 -1  0
 0  0 -1
```
exists in ITK and is not applicable to DICOM definition.

issakomi:
In fact I also think that something should be done with ITK's "from" orientation codes, it seems that surprisingly many users don't understand them at all. But how should they if there is no clear documentation and, even worse, at many places there are already "to" notations.
May be e.g. `ITK_COORDINATE_ORIENTATION_RAI` could be renamed to `ITK_COORDINATE_ORIENTATION_NOTATION_FROM_RAI` and be equal to  `ITK_COORDINATE_ORIENTATION_NOTATION_TO_LPS` or something like that.
It will allow preserve AnatomicalOrientation field of Meta images (document properly!). Also the same for some methods returning them (clarify "to" and "from" in the name). And no need to kill OrientImageFilter.

Modules/Core/Common/include/itkDICOMOrientation.h
```diff
@@ -0,0 +1,280 @@
...
+private:
+  enum class CoordinateMajornessTermsEnum : uint8_t
+  {
+    PrimaryMinor = 0,
+    SecondaryMinor = 8,
+    TertiaryMinor = 16
+  };
+
+  constexpr static uint32_t
+  toOrientation(const CoordinateEnum primary, const CoordinateEnum secondary, const CoordinateEnum tertiary)
```
N-Dekker, line 79: It may be just a matter of style, but it appears that "static constexpr" would be preferable here, instead of  "constexpr static".  See  "static constexpr unsigned long is C++’s “lovely little old French whittling knife”, by Arthur O’Dwyer: https://quuxplusone.github.io/blog/2021/04/03/static-constexpr-whittling-knife/

Modules/Core/Common/include/itkDICOMOrientation.h
N-Dekker: @blowekamp Reproduced as follows: https://godbolt.org/z/hYocrbeb5

```cpp
class MyClass
{
private:
  static constexpr int
  Func42()
  {
    return 42;
  }

  enum class EnumClass
  {
    EnumConst42 = Func42()
  };
};

```

Output from GCC 14.2:
```
<source>:15:25: error: 'static constexpr int MyClass::Func42()' called in a constant expression before its definition is complete
   15 |     EnumConst42 = Func42()
      |                   ~~~~~~^~
<source>:15:26: error: enumerator value for 'EnumConst42' is not an integer constant
   15 |     EnumConst42 = Func42()
```

N-Dekker: @blowekamp Instead of a static member function, `toOrientation(primary, secondary, tertiary)`, would it be an option for you to add a [variable template](https://en.cppreference.com/w/cpp/language/variable_template) as static member variable to `DICOMOrientation`? For example, `m_ToOrientation`, as follows:

```cpp
template <CoordinateEnum primary, CoordinateEnum secondary, CoordinateEnum tertiary>
static constexpr uint32_t m_ToOrientation =
  (static_cast<uint32_t>(primary) << static_cast<uint8_t>(CoordinateMajornessTermsEnum::PrimaryMinor)) +
  (static_cast<uint32_t>(secondary) << static_cast<uint8_t>(CoordinateMajornessTermsEnum::SecondaryMinor)) +
  (static_cast<uint32_t>(tertiary) << static_cast<uint8_t>(CoordinateMajornessTermsEnum::TertiaryMinor));
```

To be used as follows:

```cpp
enum class OrientationEnum : uint32_t
{
  RIP = m_ToOrientation<CoordinateEnum::LeftToRight, CoordinateEnum::SuperiorToInferior, CoordinateEnum::AnteriorToPosterior>,
  LIP = m_ToOrientation<CoordinateEnum::RightToLeft, CoordinateEnum::SuperiorToInferior, CoordinateEnum::AnteriorToPosterior>,
  ... // etc. etc....
};


```

blowekamp: Done

Modules/Core/Common/include/itkDICOMOrientation.h
N-Dekker: -  Related to pull request #3144 by Lee Newberg (@Leengit)

Modules/Core/Common/include/itkDICOMOrientation.h
blowekamp: This seems to work.

Modules/Core/Common/include/itkDICOMOrientation.h
blowekamp: done

Modules/Core/Common/test/CMakeLists.txt
```diff
@@ -1758,7 +1758,9 @@ set(ITKCommonGTests
     itkWeakPointerGTest.cxx
     itkCommonTypeTraitsGTest.cxx
     itkMetaDataDictionaryGTest.cxx
-    itkSpatialOrientationAdaptorGTest.cxx)
+    itkSpatialOrientationAdaptorGTest.cxx
+    itkDICOMOrientionGTest.cxx
```
lassoan, line 1762:
```suggestion
    itkDICOMOrientationGTest.cxx
```

.github/workflows/itk_dict.txt
```diff
@@ -715,6 +715,7 @@ orthonormality
 opencl
 openjpeg
 oper
+osterior
```
lassoan, line 718: This is word does not belong to the dictionary, because it should be rejected. It would be better to avoid using this word in the source code or disable spellchecking at the single place where this word is intentionally used.

Modules/Filtering/ImageGrid/include/itkDICOMOrientImageFilter.h
```diff
@@ -0,0 +1,197 @@
...
+ * given fastest moving index first, second fastest second, third fastest third, where the label's direction indicates
+ * increasing values.
+ *
+ * An ITK image with an identity direction cosine matrix is in LPS (Left, Posterior, Superior) orientation as defined by
+ * the DICOM standard.
+ *
+ * \f[
+ * LPS = \begin{Bmatrix}
+ * from\ right\ to\ \textbf{L}eft \\
+ * from\ anterior\ towards\ \textbf{P}osterior  \\
```
blowekamp, line 53: @dzenanz It was added to the dictionary due to the mark up of P.

blowekamp:
The work has evolved into another PR. I didn't expect the continued comments on this WIP, some were applied to the other PR still under development.

seanm:
> The work has evolved into another PR. I didn't expect the continued comments on this WIP, some were applied to the other PR still under development.

So wait, this one is dead?  Should we only look at https://github.com/InsightSoftwareConsortium/ITK/issues/4788 ?  I'm confused with all these different PRs...

blowekamp:
> So wait, this one is dead? Should we only look at #4788 ? I'm confused with all these different PRs...

The original issue #4788 is the discussion. I have not yet directly asked for feedback on the other PR, its still approaching being finished and incorporating the feed back I have heard. 

There was strong opposition for using the specific "DICOM" term ( and apparently overly specific/inappropriate association), and the opposition to replacing the OritenImageFilter class, a new PR with a updated title, and an updates description seemed appropriate to help reset everyone for the updated goals.

dzenanz:
Should this PR be closed now?

blowekamp:
See PR #4804 and #4895

PR #4799, opened 2024-08-08
STYLE: Remove redundant non-const `Object::AddObserver` overload

N-Dekker:
The `const` overload of `Object::AddObserver` can serve both "const" and "non-const" instances of and references to an `itk::Object`, so the non-const overload should not be necessary.

Note: this is an ABI change, and it might affect the SWIG output, but it should be acceptable for ITK 6.

----

Related pull requests:
- #3636
- #3669
- #4685 by Piyush Aggarwal (@tm9k1)

### DIFF
--- a/Wrapping/Generators/Python/PyBase/pyBase.i
+++ b/Wrapping/Generators/Python/PyBase/pyBase.i
@@ -322,18 +322,10 @@ str = str
 
     %pythonprepend itkObject::AddObserver %{
         import itk
-        if len(args) == 3 and not issubclass(args[2].__class__, itk.Command) and callable(args[2]):
-            args = list(args)
+        if not issubclass(cmd.__class__, itk.Command) and callable(cmd):
             pycommand = itk.PyCommand.New()
-            pycommand.SetCommandCallable( args[2] )
-            args[2] = pycommand
-            args = tuple(args)
-        elif len(args) == 2 and not issubclass(args[1].__class__, itk.Command) and callable(args[1]):
-            args = list(args)
-            pycommand = itk.PyCommand.New()
-            pycommand.SetCommandCallable( args[1] )
-            args[1] = pycommand
-            args = tuple(args)
+            pycommand.SetCommandCallable( cmd )
+            cmd = pycommand
     %}
 
 %enddef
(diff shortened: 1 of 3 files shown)


Wrapping/Generators/Python/PyBase/pyBase.i
```diff
@@ -322,18 +322,10 @@ str = str
...
-            args[2] = pycommand
-            args = tuple(args)
-        elif len(args) == 2 and not issubclass(args[1].__class__, itk.Command) and callable(args[1]):
-            args = list(args)
-            pycommand = itk.PyCommand.New()
-            pycommand.SetCommandCallable( args[1] )
-            args[1] = pycommand
-            args = tuple(args)
+            pycommand.SetCommandCallable( cmd )
+            cmd = pycommand
```
N-Dekker, line 328: With this commit, the pythonprepend code is very much simplified, as the last parameter of `AddObserver` now has a name, "cmd". Rather than either `args[2]` or `args[3]`.

blowekamp: Does this function cause any problems: `Object::AddObserver(const EventObject & event, std::function<void(const EventObject &)> function) const`?

N-Dekker: @blowekamp It looks like the `AddObserver(const EventObject & event, std::function<void(const EventObject &)> function)` overload does not cause any problem, when looking at the CI. Do you think it _should_ cause problems?

dzenanz:
Is this ready for merging? Making the method more general could happen in a follow-up PR, if needed.

Wrapping/Generators/Python/PyBase/pyBase.i
N-Dekker: Is the Python wrapping supposed to support passing an `std::function<void(const EventObject &)>` as argument? Or is the `std::function<void(const EventObject &)>` support specific to the C++ interface?

_If_ the Python wrapping is supposed to support passing an `std::function<void(const EventObject &)>` as argument, _in Python_, does this pull request affect such support one way or another? Honestly that isn't clear to me. I just look at the CI, and the PR appears to pass the tests just fine.

dzenanz (approved):

PR #4800, opened 2024-08-12
DOC: Error in specific DOI.

zivy:
Address error in DOI which was updated in the large change of insight-journal links.


thewtex (approved):
💯

PR #4804, opened 2024-08-14
Add AnatomicalOrientation class

blowekamp:
This PR is a WIP, and more details will be added here when ready for discussion and review.

See #4788 for this issue and discussion of these changes.

### DIFF
--- a/Modules/Core/Common/include/itkAnatomicalOrientation.h
+++ b/Modules/Core/Common/include/itkAnatomicalOrientation.h
@@ -62,7 +62,7 @@ public:
     SuperiorToInferior = 9,  ///< foot
   };
 
-private:
+protected:
   enum class CoordinateMajornessTermsEnum : uint8_t
   {
     PrimaryMinor = 0,
@@ -80,7 +80,7 @@ private:
   GetCoordinateTerm(CoordinateMajornessTermsEnum cmt) const;
 
 public:
-  enum class ToEnum : uint32_t
+  enum class PositiveEnum : uint32_t
 
   {
     INVALID = 0,
@@ -236,7 +236,7 @@ public:
                              CoordinateEnum::RightToLeft>
   };
 
-  enum class FromEnum : uint32_t
+  enum class NegativeEnum : uint32_t
 
   {
     INVALID = 0,
@@ -400,7 +400,7 @@ public:
    */
   AnatomicalOrientation(CoordinateEnum primary, CoordinateEnum secondary, CoordinateEnum tertiary);
 
-  AnatomicalOrientation(ToEnum orientation)
+  AnatomicalOrientation(PositiveEnum orientation)
     : m_Value(orientation)
   {}
 
@@ -416,30 +416,33 @@ public:
   AnatomicalOrientation(LegacyOrientationType legacyOrientation);
 #endif
 
-  AnatomicalOrientation(FromEnum orientation);
+  AnatomicalOrientation(NegativeEnum orientation);
 
   explicit AnatomicalOrientation(const DirectionType & d)
-    : m_Value(DirectionCosinesToOrientation(d))
+    : m_Value(ConvertDirectionToPositiveEnum(d))
   {}
 
-  operator ToEnum() const { return m_Value; }
-
-  std::string
-  GetAsToStringEncoding() const;
-
-  std::string
-  GetAsFromStringEncoding() const;
+  static AnatomicalOrientation
+  CreateFromPositiveStringEncoding(std::string str);
 
   static AnatomicalOrientation
-  CreateFromToStringEncoding(std::string str);
+  CreateFromNegativeStringEncoding(std::string str);
 
-  static AnatomicalOrientation
-  CreateFromFromStringEncoding(std::string str);
+  operator PositiveEnum() const { return m_Value; }
 
-  /** An involution to convert between "to" and "from" single character encoding strings.
+
+  std::string
+  GetAsPositiveStringEncoding() const;
+
+  std::string
+  GetAsNegativeStringEncoding() const;
+
+  /** An involution to convert between "positive" and "negative" single character encoding strings.
    *
    * For example the string "RAS" is converted to "LPI" and vice versa.
    *
+   * The input maybe upper or lower case, while the output is always upper case.
+   * There is not check that the input is a valid encoding.
    *
    * */
   static std::string
@@ -448,13 +451,19 @@ public:
   DirectionType
   GetAsDirection() const
   {
-    return OrientationToDirectionCosines(m_Value);
+    return ConvertPositiveEnumToDirection(m_Value);
   }
 
-  ToEnum
-  GetAsOrientation() const
+  PositiveEnum
+  GetAsPositiveOrientation() const
   {
     return m_Value;
+  }
+
+  NegativeEnum
+  GetAsNegativeOrientation() const
+  {
+    return NegativeEnum(uint32_t(this->m_Value));
   }
 
   CoordinateEnum
@@ -488,34 +497,33 @@ public:
     return (static_cast<uint8_t>(a) & AxisField) == (static_cast<uint8_t>(b) & AxisField);
   }
 
-  /** \brief Return the closest orientation for a direction cosine matrix. */
-  static ToEnum
-  DirectionCosinesToOrientation(const DirectionType & dir);
-
-  /** \brief Return the direction cosine matrix for a orientation. */
-  static DirectionType OrientationToDirectionCosines(ToEnum);
-
 
   friend ITKCommon_EXPORT std::ostream &
-                          operator<<(std::ostream & out, ToEnum value);
+                          operator<<(std::ostream & out, PositiveEnum value);
+
+protected:
+  /** \brief Return the direction cosine matrix for a orientation. */
+  static DirectionType ConvertPositiveEnumToDirection(PositiveEnum);
 
 
-private:
-  // Private methods to create the maps, these will only be called once.
+  /** \brief Return the closest orientation for a direction cosine matrix. */
+  static PositiveEnum
+  ConvertDirectionToPositiveEnum(const DirectionType & dir);
+
 
   /** \brief Return the global instance of the map from orientation enum to strings.
    *
    * The implementation uses a function static local variable so the global is created only if needed, only once.
    */
-  static const std::map<ToEnum, std::string> &
+  static const std::map<PositiveEnum, std::string> &
   GetCodeToString();
 
   /** \brief Return the global instance of the map from string to orientation enum.
    */
-  static const std::map<std::string, ToEnum> &
+  static const std::map<std::string, PositiveEnum> &
   GetStringToCode();
 
-  ToEnum m_Value;
+  PositiveEnum m_Value;
 };
 
 
@@ -523,7 +531,7 @@ ITKCommon_EXPORT std::ostream &
                  operator<<(std::ostream & out, typename AnatomicalOrientation::CoordinateEnum value);
 
 ITKCommon_EXPORT std::ostream &
-                 operator<<(std::ostream & out, typename AnatomicalOrientation::ToEnum value);
+                 operator<<(std::ostream & out, typename AnatomicalOrientation::PositiveEnum value);
 
 ITKCommon_EXPORT std::ostream &
                  operator<<(std::ostream & out, const AnatomicalOrientation & orientation);
--- a/Modules/Core/Common/include/itkAnatomicalOrientation.h
+++ b/Modules/Core/Common/include/itkAnatomicalOrientation.h
@@ -30,13 +30,23 @@ namespace itk
 {
 
 /** \class AnatomicalOrientation
- * \brief A class to represent anatomical orientations and convert between conventions.
+ * \brief Representations of anatomical orientations and methods to convert between conventions.
  *
  *  Defines patient specific anatomical names to the XYZ axes of a 3D image.
  *
- * \todo update the details of the enum names in the docs
- *  Instances hold the patient orientation enum and allow conversion to and from enums, string and direction cosine
- * matrices. Conversions from a direction cosine matrix is approximated with the orientation of the closes axes.
+ * A class instances holds the patient orientation stored as a enumerated type. Conversion to different representations
+ * such as strings and direction cosine matrices are supported.
+ *
+ * The use of unambiguous anatomical orientation names such as "RightToLeft" is preferred, where "Right" is the
+ * "from direction" and the negative direction of the coordinates, while "Left" is the "to direction" and the positive
+ * direction. The following is an unambiguous construction of an AnatomicalOrientation object:
+ *
+ * \code
+ *  AnatomicalOrientation(AnatomicalOrientation::CoordinateEnum::RightToLeft,
+ *                        AnatomicalOrientation::CoordinateEnum::AnteriorToPosterior,
+ *                        AnatomicalOrientation::CoordinateEnum::InferiorToSuperior);
+ * \endcode
+ *
  *
  * \ingroup ITKCommon
  */
@@ -51,15 +61,17 @@ public:
   using LegacyOrientationType = SpatialOrientationEnums::ValidCoordinateOrientations;
 #endif
 
+  // Anatomical names for an axis.
+  //
   enum class CoordinateEnum : uint8_t
   {
     UNKNOWN = 0,
     RightToLeft = 2, ///< 0b0010
     LeftToRight = 3,
-    PosteriorToAnterior = 4, ///< front - 0b0100
-    AnteriorToPosterior = 5, ///< back
-    InferiorToSuperior = 8,  ///< head - 0b1000
-    SuperiorToInferior = 9,  ///< foot
+    PosteriorToAnterior = 4, ///< to front - 0b0100
+    AnteriorToPosterior = 5, ///< to back
+    InferiorToSuperior = 8,  ///< to head - 0b1000
+    SuperiorToInferior = 9,  ///< to foot
   };
 
 protected:
@@ -80,6 +92,7 @@ protected:
   GetCoordinateTerm(CoordinateMajornessTermsEnum cmt) const;
 
 public:
+  // Enumerated acronyms based on the positive or "To" direction of the anatomical coordinates.
   enum class PositiveEnum : uint32_t
 
   {
@@ -236,6 +249,7 @@ public:
                              CoordinateEnum::RightToLeft>
   };
 
+  // Enumerated acronyms based on the negative or "From" direction of the anatomical coordinates.
   enum class NegativeEnum : uint32_t
 
   {
@@ -416,24 +430,30 @@ public:
   AnatomicalOrientation(LegacyOrientationType legacyOrientation);
 #endif
 
-  AnatomicalOrientation(NegativeEnum orientation);
-
+  /** Conversion for a Direction Cosine Matrix to the closest anatomical orientation. Any partial axis rotations are
+   *  rounded to the nearest axis, and lost in this conversion.
+   */
   explicit AnatomicalOrientation(const DirectionType & d)
     : m_Value(ConvertDirectionToPositiveEnum(d))
   {}
 
+  /** Creates a AnatomicalOrientation from a string with the PositiveEnum encoding. The string is case-insensitive. If
+   * the string is not a valid encoding then the orientation is set to INVALID.
+   */
   static AnatomicalOrientation
   CreateFromPositiveStringEncoding(std::string str);
 
+  // Same as CreateFromPositiveStringEncoding but for the NegativeEnum encoding.
   static AnatomicalOrientation
   CreateFromNegativeStringEncoding(std::string str);
 
   operator PositiveEnum() const { return m_Value; }
 
-
+  /** Returns the PositiveEnum encoding as a string. The string is always upper case. */
   std::string
   GetAsPositiveStringEncoding() const;
 
+  /** Returns the NegativeEnum encoding as a string. The string is always upper case. */
   std::string
   GetAsNegativeStringEncoding() const;
 
@@ -442,12 +462,13 @@ public:
    * For example the string "RAS" is converted to "LPI" and vice versa.
    *
    * The input maybe upper or lower case, while the output is always upper case.
-   * There is not check that the input is a valid encoding.
+   * There is no check that the input is a valid encoding.
    *
    * */
   static std::string
   ConvertStringEncoding(std::string str);
 
+  /** \brief Return the direction cosine matrix for the orientation. */
   DirectionType
   GetAsDirection() const
   {
@@ -526,12 +547,18 @@ protected:
   PositiveEnum m_Value;
 };
 
-
+/** Outputs unambiguous anatomical orientation names such as "right-to-left". */
 ITKCommon_EXPORT std::ostream &
                  operator<<(std::ostream & out, typename AnatomicalOrientation::CoordinateEnum value);
 
+/** Outputs the PositiveEnum encoding as a string such as "LPS". */
 ITKCommon_EXPORT std::ostream &
                  operator<<(std::ostream & out, typename AnatomicalOrientation::PositiveEnum value);
+
+
+/** Outputs the NegativeEnum encoding as a string such as "RAI" */
+ITKCommon_EXPORT std::ostream &
+                 operator<<(std::ostream & out, typename AnatomicalOrientation::NegativeEnum value);
 
 ITKCommon_EXPORT std::ostream &
                  operator<<(std::ostream & out, const AnatomicalOrientation & orientation);
--- a/Modules/Core/Common/include/itkAnatomicalOrientation.h
+++ b/Modules/Core/Common/include/itkAnatomicalOrientation.h
@@ -1,0 +1,535 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         https://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+#ifndef itkAnatomicalOrientation_h
+#define itkAnatomicalOrientation_h
+
+#include "ITKCommonExport.h"
+#include "itkImageBase.h"
+#ifndef ITK_FUTURE_LEGACY_REMOVE
+#  include "itkSpatialOrientation.h"
+#endif
+#include <map>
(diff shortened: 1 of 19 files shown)


dzenanz (commented):
Looks good so far!

gdevenyi:
I'm wondering if some of the wrongness in #147 can be solved through the application of this class..

issakomi:
> I'm wondering if some of the wrongness in #147 can be solved through the application of this class..

I am afraid that nothing can solve the issue with MINC, orient filters re-order pixels, but _left_ is still _left_ etc, axis are not flipped. Even more difficult that MINC files **written** by ITK incorporate that flip, probably no way to distinguish. Maybe MINC **3** could be the solution? Maybe without dependency on HDF, if possible?

gdevenyi:
> Maybe MINC **3** could be the solution? Maybe without dependency on HDF, if possible?

HDF5 is the reason MINC exists (to properly address some of the huge deficiencies of NIFTI).

I'll just have to finish fixing the loader to read the voxels properly then.

blowekamp:
Please move this conversation about MINC to #147 or another place on topic.

Modules/Core/Common/include/itkAnatomicalOrientation.h
```diff
@@ -0,0 +1,365 @@
...
+    (static_cast<uint32_t>(VTertiary) << static_cast<uint8_t>(CoordinateMajornessTermsEnum::TertiaryMinor));
+
+  CoordinateEnum
+  GetCoordinateTerm(CoordinateMajornessTermsEnum cmt) const;
+
+public:
+#define ITK_ORIENTATION_ENUM(P, S, T) m_OrientationValue<P, S, T>
+
+
+  enum class ToOrientationEnum : uint32_t
```
blowekamp, line 94: @dzenanz What do you think of this name "ToOrientationEnum", locally I also have a "FromOrientationEnum"? 

Making it `itk::AnatomicalOrientation::ToOrientationEnum`. You earlier suggestions to use a "typedef" (using) make since too, but the enum name seems a little redundent.

The previous was [itk::SpatialOrientation::ValidCoordinateOrientationFlags](https://itk.org/Doxygen52/html/namespaceitk_1_1SpatialOrientation.html#a8240a59ae2e7cae9e3bad5a52ea3496e).  Also the enums were of the longer form "[ITK_COORDINATE_ORIENTATION_RIP](https://itk.org/Doxygen52/html/namespaceitk_1_1SpatialOrientation.html#a8240a59ae2e7cae9e3bad5a52ea3496eaa5d3197482c4335a63a1ad99d6a9edee)" to add even more length.

dzenanz: Since they are inside `AnatomicalOrientation` class, we could remove `Orientation` from the name, so have `ToEnum` and `FromEnum`? The previous super-long names were quite unwieldy.

Modules/IO/PhilipsREC/src/itkPhilipsRECImageIO.cxx
```diff
@@ -29,7 +29,7 @@
 #include "itkIOCommon.h"
 #include "itkByteSwapper.h"
 #include "itkMetaDataObject.h"
-#include "itkSpatialOrientationAdapter.h"
+#include "itkDICOMOrientation.h"
```
seanm, line 32: The idea is to rename this without DICOM, correct?

blowekamp: Thanks. I don't think this module is enabled by default.

blowekamp: Done

Modules/Core/Common/test/itkAnatomicalOrientationGTest.cxx
```diff
@@ -0,0 +1,249 @@
...
+  static_assert(int(OE::PIR) == int(FromOE::ASL));
+  static_assert(int(OE::ASL) == int(FromOE::PIR));
+
+  itk::AnatomicalOrientation itk_rai(FromOE::RAI);
+
+  EXPECT_EQ(itk_rai, itk::AnatomicalOrientation(OE::LPS));
+  EXPECT_EQ(itk_rai.GetAsOrientation(), OE::LPS);
+  EXPECT_EQ(itk_rai.GetAsString(), "LPS");
+  const std::array<CE, 3> expected_terms = { { CE::LeftToRight, CE::AnteriorToPosterior, CE::InferiorToSuperior } };
+  EXPECT_EQ(itk_rai.GetTerms(), expected_terms);
```
blowekamp, line 227: The expected result is in correctly defined here. The following is the output from the [test failure](https://open.cdash.org/tests/1681790452):

/Users/runner/work/1/s/Modules/Core/Common/test/itkAnatomicalOrientationGTest.cxx:227: Failure
  itk_rai.GetTerms()
    Which is: { right-to-left, anterior-to-posterior, inferior-to-superior }
  expected_terms
    Which is: { left-to-right, anterior-to-posterior, inferior-to-superior }

blowekamp:
@dzenanz Almost there. There is currently both the ToEnum, and FromEnum. The class "AnatomicalOrientation" can be seen to hold the representation of the orientation, and provide conversion with the Enum's, direction matrix, and perhaps the preferred handling it as 3 components.

There is still the ToEnum string representation left over from the original DICOMOrientation class. That is `AnatomicalOreination("LPS")` construction and `AnatomicalOrientation(OE::LPS).ToString() == "LPS"`. Anatomicalorientation::operator<< produce something like "left-to-right, anterior-to-posterior, inferior-to-superior" now, which is unambiguous. Perhaps the string character tripplets should just be removed? However the string representation is convent with a Python interface.

dzenanz:
> AnatomicalOreination("LPS") construction and AnatomicalOrientation(OE::LPS).ToString() == "LPS"

I would like to have this. This convenience did not exist within ITK proper before, so there is no need for some sort of backwards compatibility. I wonder whether we should have a method to convert "To" string from a "From" string and vice versa. E.g. if an application has "RAI" string as a "From" code, it can call something like `AnatomicalOrientation::ConvertToFrom("RAI")` and get "LPS". Such method would be symmetric, as the same operation would be needed to go from "From" to "To", and from "To" to "From" convention.

blowekamp:
So your suggestion is to keep the ambiguous "ToEnum" character representation, and just add a method to convert To/From character representations?

The concern here is really when `AnatomicalOrientation("RIA")` would be used, it is not clear if the string is being interpreted as a ToEnum or a FromEnum.

dzenanz:
If you want, you can remove the string constructor. Is there another similarly convenient way to initialize it? Something like:
```cpp
auto identityAO = AnatomicalOrientation::ToEnum("LPS");
AnatomicalOrientation ao(identityAO);
```

dzenanz:
More generally,:
```cpp
std::string toCode = "LPS";
auto identityAO = AnatomicalOrientation::ToEnum(toCode);
AnatomicalOrientation ao(identityAO);
```

Modules/Core/Common/include/itkAnatomicalOrientation.h
```diff
@@ -0,0 +1,533 @@
...
+
+
+#ifndef ITK_FUTURE_LEGACY_REMOVE
+  /** \brief Conversion from Legacy SpatialOrientation
+   *
+   * @param legacyOrientation
+   */
+  AnatomicalOrientation(LegacyOrientationType legacyOrientation)
+#  ifdef ITK_FUTURE_LEGACY_REMOVE
+    [deprecated("Use the AnatomicalOrientation::FromEnum type instead.")]
```
issakomi, line 426: This `#ifdef ITK_FUTURE_LEGACY_REMOVE` is unreachable inside `#ifndef ITK_FUTURE_LEGACY_REMOVE` block.

`[[deprecated]]`, not `[deprecated]`

For a function the syntax is e.g. `[[deprecated]] void f();`

Just FYI, with this changes or without -- you know my opinion about all this stuff. I got a notification and looked, I will unsubscribe from the PR to avoid further notifications, no need to respond.

dzenanz:
What is the status of this PR? I would like it to be finished/merged before it becomes stale, as PR revival usually involves extra effort.

Modules/Filtering/ImageGrid/include/itkOrientImageFilter.h
```diff
@@ -308,19 +309,13 @@ class ITK_TEMPLATE_EXPORT OrientImageFilter : public ImageToImageFilter<TInputIm
...
   GetMajorAxisFromPatientRelativeDirectionCosine(double x, double y, double z);
 
-  CoordinateOrientationCode m_GivenCoordinateOrientation{
-    SpatialOrientationEnums::ValidCoordinateOrientations::ITK_COORDINATE_ORIENTATION_RIP
-  };
-  CoordinateOrientationCode m_DesiredCoordinateOrientation{
-    SpatialOrientationEnums::ValidCoordinateOrientations::ITK_COORDINATE_ORIENTATION_RIP
-  };
-  bool m_UseImageDirection{ false };
+  CoordinateOrientationCode m_GivenCoordinateOrientation{ AnatomicalOrientation::ToEnum::LPS };
```
dzenanz, line 312: RIP in From notation should be LSA in To notation, not LPS?

blowekamp:
@dzenanz Shall we have the SpatialOrientation filter changes be in another PR?

dzenanz:
I am fine with that. And that might even be better, to not hold up this PR.

Modules/Core/Common/include/itkAnatomicalOrientation.h
blowekamp: Thank you . I have addressed this.

Modules/Filtering/ImageGrid/include/itkOrientImageFilter.h
blowekamp: No longer including filter in this PR.

blowekamp:
When Legacy Remove is enabled deprecated warnings are generated causing one of the ITK Linux builds to failed. What is the expected behavior here?

Modules/Core/Common/include/itkAnatomicalOrientation.h
```diff
@@ -0,0 +1,543 @@
...
+  {
+    UNKNOWN = 0,
+#ifdef ITK_AMBIGUOUS_TO_COORDINATE_ENUMS
+    Left = 2,
+    Right = 3,
+    Anterior = 4,
+    Posterior = 5,
+    Superior = 8,
+    Inferior = 9,
+#endif
```
N-Dekker, line 64: Are these enumerators meant for FUTURE_REMOVE?

blowekamp: I don't particularly care. As noted by the conditional these terms are ambiguous, and the opposite of the the current SpatialOrientationEnums::[CoordinateTerms](https://itk.org/Doxygen/html/group__ITKCommon.html#gaeebc5f41f756b247ec3e5179b03102f9). The terms below like "RightToLeft" a clear an unambiguous and should be used.

Whatever is easiest for the community with these terms is likely best.

dzenanz (approved):

Modules/Core/Common/include/itkAnatomicalOrientation.h
```diff
@@ -0,0 +1,543 @@
...
+  GetAsToStringEncoding() const;
+
+  std::string
+  GetAsFromStringEncoding() const;
+
+  static AnatomicalOrientation
+  CreateFromToStringEncoding(std::string str);
+
+  static AnatomicalOrientation
+  CreateFromFromStringEncoding(std::string str);
```
N-Dekker, line 445: "FromFrom", is that intended? It looks a bit funny 🤷

blowekamp: Yes. There is "ToEnum" and "FromEnum" types defined, the the string representation are called "To String Encoding" and "From String Encoding". Then "Create From X" was the prefix for these explicit constructor-like methods.

Do you have another suggestion?

N-Dekker: > Do you have another suggestion?

Sorry, no! I haven't entirely studied your proposal, I just wanted to be sure that "FromFrom" is not a typo! (Other users might also find it confusing, I don't know.)

By the way I find it hard to review a PR of 1000+ changed lines of code and 10+ commits. (Especially when it comes to _approving_ the entire PR.) I don't know about this one specifically, but in general I would suggest to try to keep PRs a bit smaller. (And then do follow-up PRs, if necessary.) Anyway, that's just a general comment, it may be too late for this specific PR!

blowekamp: Yes, this PR has gotten a big unwieldy and likely should be squashed into just a couple commits. But due to the requests, I wanted to keep the change set until it was close, which I guess it is now.

I'd recommended just reviewing the new itkAnatomicalOrientation.h file closely.

blowekamp:
Current warning is due to testing of the deprecated methods:
```
2024-09-05T14:28:44.5388532Z /home/vsts/work/1/s/Modules/Core/Common/test/itkAnatomicalOrientationGTest.cxx:244:73: warning: 'itk::AnatomicalOrientation::AnatomicalOrientation(itk::AnatomicalOrientation::LegacyOrientationType)' is deprecated: Use the AnatomicalOrientation::FromEnum type instead. [-Wdeprecated-declarations]
2024-09-05T14:28:44.5389330Z   244 |   itk::AnatomicalOrientation itk_rai(SOE::ITK_COORDINATE_ORIENTATION_RAI);
2024-09-05T14:28:44.5389692Z       |                                                                         ^
2024-09-05T14:28:44.5390384Z In file included from /home/vsts/work/1/s/Modules/Core/Common/test/itkAnatomicalOrientationGTest.cxx:22:
2024-09-05T14:28:44.5390862Z /home/vsts/work/1/s/Modules/Core/Common/include/itkAnatomicalOrientation.h:424:3: note: declared here
2024-09-05T14:28:44.5391276Z   424 |   AnatomicalOrientation(LegacyOrientationType legacyOrientation);
```

Anyone know other locations in ITK which test deprecated methods with out warnings?

dzenanz:
This might be the answer you are looking for: https://github.com/InsightSoftwareConsortium/ITK/blob/fb61d2cadc33bf50d1f0ec5b842b70a1a7021d0e/Modules/Core/Common/test/itkSTLThreadTest.cxx#L18-L20

dzenanz:
Taking a look at the definition in `itkMacro.h`, it might not.

Modules/Core/Common/test/itkAnatomicalOrientationGTest.cxx
```diff
@@ -19,6 +19,9 @@
 #include <gtest/gtest.h>
 
 #include "itkGTest.h"
+#ifndef ITK_FUTURE_LEGACY_REMOVE
```
dzenanz, line 22: Shouldn't this be `#ifdef`, not `#ifndef`?

blowekamp: If ITK_FUTURE_LEGACY_REMOVE is true then the method with the depreciation is removed, along with the test of it.

dzenanz (approved):
This is primarily rebase and squash? With removal of `#ifdef ITK_AMBIGUOUS_TO_COORDINATE_ENUMS`?

blowekamp:
> This is primarily rebase and squash? With removal of `#ifdef ITK_AMBIGUOUS_TO_COORDINATE_ENUMS`?

Yes

dzenanz:
Mark as "ready for review"?

blowekamp:
The Documentation still needs work.

Modules/Core/Common/include/itkAnatomicalOrientation.h
thewtex: Reading this class and the implementation for the first time, everything is quite good, but I was also confused by the `GetAsToStringEncoding`, `GetAsFromStringEncoding`, `CreateFromToStringEncoding`, `CreateFromFromStringEncoding`. Possible improvement: `GetPositiveStringEncoding`, `GetNegativeStringEncoding`, CreateFromPositiveStringEncoding`, `CreateFromNegativeStringEncoding`.

thewtex (commented):
Very nice!!

Inline comment regarding method naming.

Wish list, could come later: `*Long*` methods to work with the names as used in `PrintSelf`.

Modules/Core/Common/include/itkAnatomicalOrientation.h
blowekamp: That seems like an improved convention, it would also follow to rename "ToEnum"->"PositiveEnum" and "FromEnum"->"NegativeEnum".

Modules/Core/Common/include/itkAnatomicalOrientation.h
blowekamp: I have removed the conditional.

blowekamp:
> Wish list, could come later: `*Long*` methods to work with the names as used in `PrintSelf`.

The class has an "operator<<" so it work sells as a ivar for ITK classes derived from LightObject with a PrintSelf methods. Could you please explain this further or be more specific.

dzenanz (approved):

thewtex:
> The class has an "operator<<" so it work sells as a ivar for ITK classes derived from LightObject with a PrintSelf methods.

Yes, that is nice.

> Could you please explain this further or be more specific.

I was thinking:

```cxx
  static const std::map<PositiveEnum, std::string> &
  GetCodeToLongString();
// returns "left-to-right", for example

  static const std::map<std::string, PositiveEnum> &
  GetStringToCode()
// has "left-to-right", etc. keys
```

dzenanz:
Brad, do you plan to get around to Matt's suggestion soon? If not, I think we should merge this PR as-is and tackle the long string codes later.

blowekamp:
There was a complication with Matt suggestion, as I recall it may have been that should happen with the individual axes enums and not the three-pul. 

There is a todo to improve the documentation. Given the confusion around these term, it is perhaps equally importation as the code ( even it the code now appears to be unambiguous in its terminology.)

But sure, merge it.

dzenanz:
I am not trying to pressure you into merging. I am merely concerned about this big and important PR getting stale with time.

blowekamp:
It has a been rebased to renew the freshness, and help reduce potential conflicts.

Modules/Core/Common/include/itkAnatomicalOrientation.h
```diff
@@ -30,13 +30,23 @@ namespace itk
...
 /** \class AnatomicalOrientation
- * \brief A class to represent anatomical orientations and convert between conventions.
+ * \brief Representations of anatomical orientations and methods to convert between conventions.
  *
  *  Defines patient specific anatomical names to the XYZ axes of a 3D image.
  *
- * \todo update the details of the enum names in the docs
- *  Instances hold the patient orientation enum and allow conversion to and from enums, string and direction cosine
- * matrices. Conversions from a direction cosine matrix is approximated with the orientation of the closes axes.
+ * A class instances holds the patient orientation stored as a enumerated type. Conversion to different representations
```
dzenanz, line 37: a enumerated -> an enumerated

Modules/Core/Common/include/itkAnatomicalOrientation.h
```diff
@@ -0,0 +1,584 @@
...
+              (static_cast<uint32_t>(tertiary) << static_cast<uint8_t>(CoordinateMajornessTermsEnum::TertiaryMinor))))
+  {}
+
+
+  constexpr AnatomicalOrientation(PositiveEnum toOrientation)
+    : m_Value(toOrientation)
+  {}
+
+
+  constexpr AnatomicalOrientation(NegativeEnum fromOrientation)
```
N-Dekker, line 432: Please consider declaring those constructors `explicit` (to avoid implicit type conversion).

Modules/Core/Common/include/itkAnatomicalOrientation.h
```diff
@@ -0,0 +1,584 @@
...
+    : m_Value(PositiveEnum(uint32_t(fromOrientation)))
```
N-Dekker, line 433: Please use `static_cast<T>(x)` instead of `T(x)`, for built-in types and enum types.

Modules/Core/Common/include/itkAnatomicalOrientation.h
N-Dekker: OK, so you did consider 😃

Modules/Core/Common/include/itkAnatomicalOrientation.h
blowekamp: done

dzenanz (approved):

thewtex (approved):
:champagne:

blowekamp:
I'm planning on merging this tomorrow unless there are further comments.

N-Dekker:
@blowekamp Shall I still have a look before the merge?

Modules/Core/Common/src/itkAnatomicalOrientation.cxx
```diff
@@ -0,0 +1,328 @@
...
+  CoordinateEnum terms[3] = { CoordinateEnum::UNKNOWN, CoordinateEnum::UNKNOWN, CoordinateEnum::UNKNOWN };
+
+  std::multimap<double, std::pair<unsigned, unsigned>> value_to_idx;
+  for (unsigned int c = 0; c < 3; ++c)
+  {
+    for (unsigned int r = 0; r < 3; ++r)
+    {
+      value_to_idx.emplace(std::abs(dir[c][r]), std::make_pair(c, r));
+    }
+  }
```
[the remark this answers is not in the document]
blowekamp: The math or the general input and output of the method?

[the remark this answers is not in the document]
N-Dekker: So you are creating this multi-map, in order to find the direction matrix element that has the highest absolute value, right?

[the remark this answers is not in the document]
blowekamp: Yes, that appears to be what is going on. As the comment says it was mostly a copy of the existing working method in  itk::SpatialObjectAdaptor::FromDirectionCosines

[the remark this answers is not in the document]
N-Dekker: OK, thanks, and the math itself? How does it work? Is it a known approach?

dzenanz:
🎆 🎊 🥳

N-Dekker:
@blowekamp There appear some compile errors at the CI (master branch), if you know how to fix them, please do 😃 At https://open.cdash.org/viewBuildError.php?buildid=9981860
```

/Users/builder/externalModules/IO/PhilipsREC/src/itkPhilipsRECImageIO.cxx:713:61: error: no member named 'FromEnum' in 'itk::AnatomicalOrientation'
  AnatomicalOrientation coord_orient(AnatomicalOrientation::FromEnum::RSA);
                                     ~~~~~~~~~~~~~~~~~~~~~~~^
[CTest: warning matched] /Users/builder/externalModules/IO/PhilipsREC/src/itkPhilipsRECImageIO.cxx:713:37: warning: parentheses were disambiguated as a function declaration [-Wvexing-parse]
  AnatomicalOrientation coord_orient(AnatomicalOrientation::FromEnum::RSA);
                                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[CTest: warning matched] /Users/builder/externalModules/IO/PhilipsREC/src/itkPhilipsRECImageIO.cxx:713:38: note: add a pair of parentheses to declare a variable
  AnatomicalOrientation coord_orient(AnatomicalOrientation::FromEnum::RSA);
                                     ^
      
 
```

blowekamp:
Follow up: #4899

jhlegarreta:
Seems like there's an ITK Windows Python warning introduced by this PR:
```
D:\a\1\s-build\Wrapping\Modules\ITKCommon\itkAnatomicalOrientationPython.cpp(3971):
 warning C4996: 'itk::AnatomicalOrientation::AnatomicalOrientation':
 Use the AnatomicalOrientation::FromEnum type instead.
```

e.g.
https://dev.azure.com/itkrobotwindowpython/ITK.Windows.Python/_build/results?buildId=4113&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=77aad734-2057-5694-9ae2-ee1f5f26eae8&l=4288

Not sure how the below block could be handled better to avoid the warning:
https://github.com/InsightSoftwareConsortium/ITK/pull/4804/files#diff-0271cfd0418afac76ae9a99eb3c59923eac04cdfc032bbf51c3bc59c88ca1c81R441-R443

blowekamp:
> [](https://open.cdash.org/viewBuildError.php?type=1&buildid=9988068)
> 
> Not sure how the below block could be handled better to avoid the warning: https://github.com/InsightSoftwareConsortium/ITK/pull/4804/files#diff-0271cfd0418afac76ae9a99eb3c59923eac04cdfc032bbf51c3bc59c88ca1c81R441-R443

Neither am I. Is the build/wrapping defining ITK_LEGACY_SILENT when compiling?

N-Dekker:
@blowekamp Please note that the warning still uses the term `FromEnum`. Didn't you replace that by Positive/Negative Enum?

blowekamp:
> Please note that the warning still uses the term `FromEnum`. Didn't you replace that by Positive/Negative Enum?

Yes, good catch the text in the warning should be updated as you recommend. But the issue is that the warning should not be occurring.

blowekamp:
Perhaps the bigger problem on the nightly build:
```
/Users/builder/external[Modules/IO/PhilipsREC/src/itkPhilipsRECImageIO.cxx:721](https://github.com/InsightSoftwareConsortium/ITK/blob/0883d0c3b7967b9319ad9ac3a3d15fae576e382c/Modules/IO/PhilipsREC/src/itkPhilipsRECImageIO.cxx#L721):20: error: no viable overloaded '='
      coord_orient = SpatialOrientationEnums::ValidCoordinateOrientations::ITK_COORDINATE_ORIENTATION_RAI;
      ~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[CTest: warning matched] /Users/builder/external[Modules/Core/Common/include/itkAnatomicalOrientation.h:53](https://github.com/InsightSoftwareConsortium/ITK/blob/0883d0c3b7967b9319ad9ac3a3d15fae576e382c/Modules/Core/Common/include/itkAnatomicalOrientation.h#L53):24: note: candidate function (the implicit copy assignment operator) not viable: no known conversion from 'itk::SpatialOrientationEnums::ValidCoordinateOrientations' to 'const itk::AnatomicalOrientation' for 1st argument
class ITKCommon_EXPORT AnatomicalOrientation
                       ^
```

@seanm may be compiling with LEGACY_REMOVE enabled so this conversion function is removed.

dzenanz:
Most of `Ryzenator`'s and `Corista`'s builds have `ITK_LEGACY_REMOVE:BOOL=ON`.

blowekamp:
> Most of `Ryzenator`'s and `Corista`'s builds have `ITK_LEGACY_REMOVE:BOOL=ON`.

But they don't have PhilipsREC module enabled?

dzenanz:
> But they don't have PhilipsREC module enabled?

Ah, that is a non-default module, which I indeed did not turn on.

PR #4806, opened 2024-08-16
dicom structured report prefix

thewtex:
The dsr2html tool uses a hardcoded literal string as a hyperlink prefix for refrenced composite objects (images, waveforms, etc.) in rendering the Key Object Selection documents.

Replace this literal with a string argument that will be passed by a calling application / function.

Pushed upstream: DCMTK/dcmtk#108


jadh4v (approved):
LGTM

PR #4811, opened 2024-08-20
COMP: Get ITK_USE_FLOAT_SPACE_PRECISION=ON to compile again

thewtex:

PR #4813, opened 2024-08-21
COMP: Bump ITKTotalVariation remote module

thewtex:
CMake fixes to address CMake IMPORTED target:

  The link interface of target "proxTV" contains:

    OpenMP::OpenMP_CXX

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target. * An ALIAS target is missing.


dzenanz (approved):

dzenanz:
#4811 has now been merged. A plain rebase should be sufficient for a green CI.

dzenanz (approved):

dzenanz:
Also merged to release and master.

PR #4772, opened 2024-07-10
ENH: Add indent to image's matrix printing

blowekamp:
Add Matrix::PrintSelf method which uses ITK's indent object to allow proper indenting of nested objects.

Fixes left aligned matrix values when printing the Image classes.

## PR Checklist
- [ ] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [ ] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [ ] Added test (or behavior not changed)
- [ ] Updated API documentation (or API not changed)
- [ ] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [ ] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [ ] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples) for all new major features (if any)

Refer to the [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) for
further development details if necessary.

PR #4814, opened 2024-08-21
BUG: Fix printing of values in MetaDataObject

### LINKED ISSUE #1454: MetaDataDictionary::Print only prints [UNKNOWN_PRINT_CHARACTERISTICS]

blowekamp:
Also removes dead code related to METADATAPRINT

Does not currently add printing for std::vector.

Fix #1454
Near duplicate of: #4368
Co-Author: Pablo Hernandez-Cerdan <pablo.hernandez.cerdan@outlook.com>

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkMetaDataObject.hxx
+++ b/Modules/Core/Common/include/itkMetaDataObject.hxx
@@ -59,11 +59,45 @@ MetaDataObject<MetaDataObjectType>::SetMetaDataObjectValue(const MetaDataObjectT
   Self::Assign(m_MetaDataObjectValue, newValue);
 }
 
+namespace
+{
+template <class T, class = void>
+struct has_Print : std::false_type
+{};
+
+template <class T>
+struct has_Print<T, std::void_t<decltype(std::declval<T>().Print(std::declval<std::ostream &>()))>> : std::true_type
+{};
+
+template <class T, class = void>
+struct has_output_operator : std::false_type
+{};
+
+template <class T>
+struct has_output_operator<T, std::void_t<decltype(std::declval<std::ostream &>() << std::declval<T>())>>
+  : std::true_type
+{};
+} // namespace
+
 template <typename MetaDataObjectType>
 void
 MetaDataObject<MetaDataObjectType>::Print(std::ostream & os) const
 {
-  Superclass::Print(os);
+  // future c++20 feature
+  // constexpr bool hasPrint = false; requires( const &MetaDataObjectType obj ) { obj.Print(os); };
+
+  if constexpr (has_Print<MetaDataObjectType>::value)
+  {
+    m_MetaDataObjectValue.Print(os);
+  }
+  else if constexpr (has_output_operator<MetaDataObjectType>::value)
+  {
+    os << m_MetaDataObjectValue;
+  }
+  else
+  {
+    Superclass::Print(os);
+  }
 }
 
 } // end namespace itk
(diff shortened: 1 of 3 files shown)


dzenanz (approved):
Code looks elegant. Hopefully it works!

blowekamp:
This also does not make use of itk::Matrix::PrintSelf. The LightObject class provides a standard Print, and PrintSelf method ( with indent). The MetaDataObject class is derived from the LightObject, while the MetaDataDictionary is not.

dzenanz:
Do the image directions and other matrices get printed correctly or as `[UNKNOWN PRINT CHARACTERISTICS]`?

blowekamp:
The itk::Matrix and itk::Array classes get printed with their "operator<<". The new Matrix::PrintSelf (#4772) is not used.

std::vector still get printed with `[UNKNOWN PRINT CHARACTERISTICS]`

phcerdan:
Looks good and thanks for the follow up.
Why aren't we adding support for std::vector or other iterables?
Seems like a good opportunity.

phcerdan (approved):
Looks good!
In the future, we could look up on how to print iterables in a general way.

dzenanz:
Could [this operator](https://github.com/InsightSoftwareConsortium/ITK/blob/v5.4.0/Modules/Core/Common/include/itkPrintHelper.h#L44) be used?

blowekamp:
> Could [this operator](https://github.com/InsightSoftwareConsortium/ITK/blob/v5.4.0/Modules/Core/Common/include/itkPrintHelper.h#L44) be used?

I ran into issue with the two approaches I tried to use that overloaded operator. Perhaps this can be a second PR?

Modules/Core/Common/include/itkMetaDataObject.hxx
```diff
@@ -59,11 +59,45 @@ MetaDataObject<MetaDataObjectType>::SetMetaDataObjectValue(const MetaDataObjectT
   Self::Assign(m_MetaDataObjectValue, newValue);
 }
 
+namespace
+{
```
N-Dekker, line 63: Please don't open an unnamed namespace in a header file (*.h, *.hxx)

dzenanz: Looks like a good excuse for an immediate follow-up? 😄 Is there also some chance for you tackle `std::vector` printing?

blowekamp: @N-Dekker What is the alternative suggestion?  Should move the classes to ITK's meta programming sub namespace?

N-Dekker: > Should move the classes to ITK's meta programming sub namespace?

Good question... is the intention to have `has_Print` and `has_output_operator` as part of the public interface, for end-users? Then they could go into the meta programming sub-namespace. Or they could just remain in the root of the `itk` namespace. But then they should still be moved from the ".hxx" file into a ".h" file

Otherwise, is the intention to make clear to users that `has_Print` and `has_output_operator` are just implementation details? Then we could place them into some implementation-detail-namespace. We could for example name "itkMetaDataObjectImplementationDetail"

A third option: If you want to entirely hide them away from the end-users, you could probably make them privately nested types of MetaDataObject.

PR #4815, opened 2024-08-22
Document that Iterator(image, region) constructors initialize at the begin, add GTest unit test

N-Dekker:
Documented that `Iterator(image, region)` constructors initialize the iterator at the beginning of the region. Added a GoogleTest unit test to checks this added specification.

### DIFF
--- a/Modules/Core/Common/test/itkImageRegionIteratorGTest.cxx
+++ b/Modules/Core/Common/test/itkImageRegionIteratorGTest.cxx
@@ -1,0 +1,136 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         https://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+// First include the header files to be tested:
+#include "itkImageConstIterator.h"
+#include "itkImageConstIteratorWithIndex.h"
+#include "itkImageConstIteratorWithOnlyIndex.h"
+#include "itkImageIterator.h"
+#include "itkImageIteratorWithIndex.h"
+#include "itkImageLinearConstIteratorWithIndex.h"
+#include "itkImageLinearIteratorWithIndex.h"
+#include "itkImageRegionConstIterator.h"
+#include "itkImageRegionConstIteratorWithIndex.h"
+#include "itkImageRegionConstIteratorWithOnlyIndex.h"
+#include "itkImageRegionExclusionConstIteratorWithIndex.h"
+#include "itkImageRegionExclusionIteratorWithIndex.h"
+#include "itkImageRegionIterator.h"
+#include "itkImageRegionIteratorWithIndex.h"
+#include "itkImageRegionReverseConstIterator.h"
+#include "itkImageRegionReverseIterator.h"
+#include "itkImageReverseConstIterator.h"
+#include "itkImageReverseIterator.h"
+#include "itkImageScanlineConstIterator.h"
+#include "itkImageScanlineIterator.h"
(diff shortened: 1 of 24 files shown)


dzenanz (approved):

Modules/Core/Common/test/itkImageRegionIteratorGTest.cxx
```diff
@@ -0,0 +1,136 @@
...
+// Has_IsAtBegin tells whether the specified iterator type has an `IsAtBegin()` member function, returning a `bool`. It
+// is inspired by "How to detect whether there is a specific member variable in class?", answered by Andy Prowl, Jan 25,
+// 2013 at https://stackoverflow.com/a/14523787 (the `has_id` example).
+template <typename TIterator, typename = bool>
+struct Has_IsAtBegin : std::false_type
+{};
+
+template <typename TIterator>
+struct Has_IsAtBegin<TIterator, decltype(std::declval<TIterator>().IsAtBegin())> : std::true_type
+{};
```
N-Dekker, line 58: - Fun fact: I found this "has member" trick (`Has_IsAtBegin`) kind-of independent of `has_Print` and `has_output_operator`, from Bradley's (@blowekamp) pull requests, #4814 and #4821

Modules/Core/Common/test/itkImageRegionIteratorGTest.cxx
```diff
@@ -0,0 +1,136 @@
...
+CheckIteratorsConstructedAtBegin()
+{
+  (CheckConstructedAtBegin<TIteratorTemplate<itk::Image<int>>>(), ...);
+  (CheckConstructedAtBegin<TIteratorTemplate<itk::Image<double, 3>>>(), ...);
+}
+} // namespace
+
+
+// Checks that an iterator that is just constructed by `IteratorType(image, region)` is at the begin.
+TEST(ImageRegionIterator, IsConstructedAtBegin)
```
blowekamp, line 112: The is some 💪 use of variadict templates here. However with the nested dispatching of the functions how does test report a failure? Does it indicate the type? Is some Information needed the the "EXPECT" statements?

blowekamp (approved):
LGTM. Just a commented on a small thought.

Modules/Core/Common/test/itkImageRegionIteratorGTest.cxx
```diff
@@ -0,0 +1,136 @@
...
+  for (const itk::IndexValueType indexValue : { 0, 1 })
+  {
+    for (const auto sizeValue : { itk::SizeValueType{ 1 }, imageSizeValue - 1 })
+    {
+      const RegionType imageRegion(IndexType::Filled(indexValue), SizeType::Filled(sizeValue));
+
+      const TIterator iterator(image, imageRegion);
+      TIterator       iteratorThatGoesToBegin = iterator;
+      iteratorThatGoesToBegin.GoToBegin();
+      EXPECT_EQ(iterator, iteratorThatGoesToBegin);
```
N-Dekker, line 88: @blowekamp  Re: https://github.com/InsightSoftwareConsortium/ITK/pull/4815#discussion_r1734596667
> However with the nested dispatching of the functions how does test report a failure? Does it indicate the type? Is some Information needed the the "EXPECT" statements?

Indeed, the "EXPECT" statements currently do not report which particular iterator type has failed, if there would be any failure at all. Would you like that? In practice, none of them fail, and they probably never will. So it might be a bit hypothetical. But we could add the name of the iterator type (just in case of a failure) as follows: 

```cpp
     EXPECT_EQ(iterator, iteratorThatGoesToBegin) << "IteratorType: " << typeid(TIterator).name();
```
In the rare (unlikely) case of such a test failure, it would then print, with the report of the test failure:

    IteratorType: class itk::ImageScanlineIterator<class itk::Image<double,3> >

Would you like that?

Modules/Core/Common/test/itkImageRegionIteratorGTest.cxx
N-Dekker: Thanks, here is my reply: https://github.com/InsightSoftwareConsortium/ITK/pull/4815#discussion_r1734695241

Modules/Core/Common/test/itkImageRegionIteratorGTest.cxx
N-Dekker: Otherwise we could do a feature request to https://github.com/google/googletest to ask them to print the type names of the arguments, in case of a test failure! 😺

N-Dekker:
- For the record, this pull request is meant to support style commits like pull request #3953 commit 5a60ed00adbc108e5c81bd2ddfbc988f28eaf722

PR #4816, opened 2024-08-22
More -Wextra-semi fixes

seanm:


Modules/IO/MeshBase/include/itkMeshConvertPixelTraits.h
```diff
@@ -170,15 +170,15 @@ ITK_DEFAULTCONVERTTRAITS_NATIVE_SPECIAL(double);
...
+  }
 
 
 // Define traits for Offset<> from dimensions 1 to 5
-ITK_MESH_DEFAULTCONVERTTRAITS_OFFSET_TYPE(1)
-ITK_MESH_DEFAULTCONVERTTRAITS_OFFSET_TYPE(2)
-ITK_MESH_DEFAULTCONVERTTRAITS_OFFSET_TYPE(3)
-ITK_MESH_DEFAULTCONVERTTRAITS_OFFSET_TYPE(4)
-ITK_MESH_DEFAULTCONVERTTRAITS_OFFSET_TYPE(5)
+ITK_MESH_DEFAULTCONVERTTRAITS_OFFSET_TYPE(1);
```
dzenanz, line 177: Here, semicolons are added. Is this intentional?

dzenanz (approved):

Modules/IO/MeshBase/include/itkMeshConvertPixelTraits.h
seanm: Yes. My understanding is that ITK prefers that even non-function-like macros should require trailing semis (by means of that static_assert hack).  Not a fan of that style personally, but going with it.

thewtex (approved):
@seanm thanks :1st_place_medal:

seanm:
BTW I unsuppressed -Wextra-semi on one of my cdash submissions yesterday, and no warnings today!  So we are now -Wextra-semi-clean!

PR #4819, opened 2024-08-23
Fix clang -Wtautological-pointer-compare warnings

seanm:
Warning was:

```
/Modules/Registration/GPUPDEDeformable/include/itkGPUPDEDeformableRegistrationFilter.hxx:102:7: warning: comparison of array 'this->m_GPUSmoothingKernels' not equal to a null pointer is always true [-Wtautological-pointer-compare]
  if (m_GPUSmoothingKernels != nullptr)
      ^~~~~~~~~~~~~~~~~~~~~    ~~~~~~~
```

### DIFF
--- a/Modules/Registration/GPUPDEDeformable/include/itkGPUPDEDeformableRegistrationFilter.hxx
+++ b/Modules/Registration/GPUPDEDeformable/include/itkGPUPDEDeformableRegistrationFilter.hxx
@@ -90,21 +90,15 @@ GPUPDEDeformableRegistrationFilter<TFixedImage, TMovingImage, TDisplacementField
   }
 
   os << indent << "SmoothingKernels: ";
-  if (m_SmoothingKernels != nullptr)
+  for (unsigned d = 0; d < ImageDimension; ++d)
   {
-    for (unsigned d = 0; d < ImageDimension; ++d)
-    {
-      os << indent.GetNextIndent() << m_SmoothingKernels[d] << std::endl;
-    }
+    os << indent.GetNextIndent() << m_SmoothingKernels[d] << std::endl;
   }
 
   os << indent << "GPUSmoothingKernels: ";
-  if (m_GPUSmoothingKernels != nullptr)
+  for (unsigned d = 0; d < ImageDimension; ++d)
   {
-    for (unsigned d = 0; d < ImageDimension; ++d)
-    {
-      os << indent.GetNextIndent() << m_GPUSmoothingKernels[d] << std::endl;
-    }
+    os << indent.GetNextIndent() << m_GPUSmoothingKernels[d] << std::endl;
   }
 
   os << indent << "UpdateFieldSmoothingKernelSizes: ";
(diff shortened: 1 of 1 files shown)


seanm:
I'm not sure if clang is correct here... maybe maybe it's only tautological on macOS and not always?

dzenanz:
This might be tautological within some unit test, but `PrintSelf` can be invoked by user code, after those pointers have been set, right?

Modules/Registration/GPUPDEDeformable/include/itkGPUPDEDeformableRegistrationFilter.hxx
```diff
@@ -90,7 +90,6 @@ GPUPDEDeformableRegistrationFilter<TFixedImage, TMovingImage, TDisplacementField
   }
 
   os << indent << "SmoothingKernels: ";
-  if (m_SmoothingKernels != nullptr)
   {
```
jhlegarreta, line 93: Sorry to chime in. These braces are then unnecessary.

seanm: Yeah, but the point is I'm not even sure this change is correct at all. I'm not convinced these things can never be null.

dzenanz: Top-level variables are statically allocated at the point of declaration:
https://github.com/InsightSoftwareConsortium/ITK/blob/v5.4.0/Modules/Registration/GPUPDEDeformable/include/itkGPUPDEDeformableRegistrationFilter.h#L222-L229

dzenanz: But non-null check should go 4 lines down, here: `os << indent.GetNextIndent() << m_SmoothingKernels[d] << std::endl;`

seanm: Oh, I think I see. So `m_SmoothingKernels` indeed cannot be null, but `m_SmoothingKernels[x]` might be, for some `x`?

dzenanz: Exactly.

seanm: Might be. There's no check currently, so it would be crashing already I suppose.

dzenanz (approved):
We might want to remove the now unnecessary block braces `{}`.

Modules/Registration/GPUPDEDeformable/include/itkGPUPDEDeformableRegistrationFilter.hxx
dzenanz: Printing a null pointer results in something like `0x00000000`.

dzenanz (approved):

N-Dekker (approved):
Thanks 👍

PR #4821, opened 2024-08-26
STYLE: move details to MetaDataObjectDetail

blowekamp:
Move meta programming details from unnamed namespace to separate file and header.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkMetaDataObject.hxx
+++ b/Modules/Core/Common/include/itkMetaDataObject.hxx
@@ -28,6 +28,7 @@
 #ifndef itkMetaDataObject_hxx
 #define itkMetaDataObject_hxx
 
+#include "itkMetaDataObjectDetail.h"
 
 namespace itk
 {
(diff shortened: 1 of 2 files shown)


N-Dekker (approved):
Thanks @blowekamp I believe that the removal of this unnamed namespace from the hxx file is necessary because of the "one definition rule": multiple cxx files may share that same (one) definition of `itk::MetaDataObject`, whereas an unnamed namespace is different for each cxx file. But I did not ask the guru's  😸 

Instead of moving the details to a different .h file, it would also be OK to me to have them in itkMetaDataObject.h or itkMetaDataObject.hxx (as long as they are in the "details" namespace), but your current PR is also fine to me. 👍

blowekamp:
> Thanks @blowekamp I believe that the removal of this unnamed namespace from the hxx file is necessary because of the "one definition rule": multiple cxx files may share that same (one) definition of `itk::MetaDataObject`, whereas an unnamed namespace is different for each cxx file. But I did not ask the guru's 😸

Many complications to make this a clear issue template classes, unnamed namespace or local linkage, and only used as constexpr so the symbols may not really exist in the compilation unit.


> Instead of moving the details to a different .h file, it would also be OK to me to have them in itkMetaDataObject.h or itkMetaDataObject.hxx (as long as they are in the "details" namespace), but your current PR is also fine to me. 👍

I may have miss understood what you suggested in parts of #4814. But it's already done, and I followed the model of itkMathDetail.h, and others. We are not too consistent with how these detail name spaces are defined.

PR #4828, opened 2024-08-29
STYLE: Remove unused "itkFrequencyImageRegion*IteratorWithIndex.h" files

N-Dekker:
Both `FrequencyImageRegionIteratorWithIndex` and `FrequencyImageRegionConstIteratorWithIndex` appear unused and untested. Attempts to use them would trigger compile errors.

An attempt to use `FrequencyImageRegionIteratorWithIndex` triggered:

    error C3210: 'ImageRegionIteratorWithIndex<itk::Image<int,2> >': a member using-declaration can only be applied to a base class member

An attempt to use `FrequencyImageRegionConstIteratorWithIndex` triggered:

    error C2679: binary '=': no operator found which takes a right-hand operand of type 'const itk::Point<T,3>' (or there is no acceptable conversion)

This commit removes their header files. It also removes the suggestion to use `FrequencyImageRegionIteratorWithIndex` from the documentation of UnaryFrequencyDomainFilter.


phcerdan (changes requested):
They are used in the IsotropicWavelets module. We should move the tests from there here. 

These filters are useful, prefer a fix rather than removal.

N-Dekker:
@phcerdan  Thanks for your very quick response, Pablo!

> They are used in the IsotropicWavelets module. We should move the tests from there here.

Sorry, but how can they possibly be used when they cannot compile at all? Even the very first revision of those two iterators (six years ago, e16d0bc4eec7b3ee7ab63c3469077d54390df5c6) would have those compile errors, as far as I can see. 

The other six iterator types from Modules/Filtering/ImageFrequency compile fine, it's just these two. So I suspect that these two are not really necessary. 🤷

phcerdan:
Thanks @N-Dekker, I am out of office and away from keyboard to give a detailed answer about the history of the filters. But if you don't want to work on a fix, let me do it when I am back.

Thanks

PR #4835, opened 2024-09-06
Declare ByteSwapper SystemIsBigEndian(), SystemIsLittleEndian() `constexpr`

N-Dekker:
Declared ByteSwapper member functions SystemIsBigEndian(), SystemIsBE(), SystemIsLittleEndian(), and SystemIsLE() `constexpr`.

Replaced `if` with `if constexpr` whenever the condition only has calls to those member functions.


dzenanz (approved):
I love it.

N-Dekker:
> I love it.

Thanks @dzenanz! I was looking at [this ITKIOOMEZarrNGFF code of yours](https://github.com/InsightSoftwareConsortium/ITKIOOMEZarrNGFF/blob/7b52663070bb210d81bf50959409b511b31c53d3/src/itkOMEZarrNGFFImageIO.cxx#L281), thinking that that _if_ might have been _constexpr_...

```cpp
    if (ByteSwapper<int>::SystemIsBigEndian())
    {
      dtype = ">";
    }
```

By the way, [std::endian](https://en.cppreference.com/w/cpp/types/endian) is constexpr-friendly, but only available from C++20, I see.

dzenanz:
Feel free to make the change! After this one gets integrated, of course. Maybe even after this change makes it into a release, so we can use it for Python packages?

seanm:
> By the way, [std::endian](https://en.cppreference.com/w/cpp/types/endian) is constexpr-friendly, but only available from C++20, I see.

Just in time for basically no CPUs being Big Endian anymore. :)

blowekamp:
> > By the way, [std::endian](https://en.cppreference.com/w/cpp/types/endian) is constexpr-friendly, but only available from C++20, I see.
> 
> Just in time for basically no CPUs being Big Endian anymore. :)

Except big-endian is the default network byte order, and Java byte order... but still not the CPU or system.

PR #4841, opened 2024-09-09
Replace `#ifdef`s in ByteSwapper with `if constexpr`, mention C++20 `std::endian` in documentation

N-Dekker:
- Follow-up to pull request #4835

Note: when reviewing, it may be helpful to ignore white-space changes: https://github.com/InsightSoftwareConsortium/ITK/pull/4841/files?w=1

### DIFF
--- a/Modules/Core/Common/include/itkByteSwapper.hxx
+++ b/Modules/Core/Common/include/itkByteSwapper.hxx
@@ -39,202 +39,170 @@ namespace itk
 //------Big Endian methods----------------------------------------------
 
 // Use different swap methods based on type
-#ifdef CMAKE_WORDS_BIGENDIAN
+
 template <typename T>
 void
-ByteSwapper<T>::SwapFromSystemToBigEndian(T *)
-{}
-#else
-template <typename T>
-void
-ByteSwapper<T>::SwapFromSystemToBigEndian(T * p)
+ByteSwapper<T>::SwapFromSystemToBigEndian([[maybe_unused]] T * p)
 {
-  switch (sizeof(T))
+  if constexpr (!m_SystemIsBigEndian)
   {
-    case 1:
-      return;
-    case 2:
-      Self::Swap2(p);
-      return;
-    case 4:
-      Self::Swap4(p);
-      return;
-    case 8:
-      Self::Swap8(p);
-      return;
-    default:
-      itkGenericExceptionMacro("Cannot swap number of bytes requested");
+    switch (sizeof(T))
+    {
+      case 1:
+        return;
+      case 2:
+        Self::Swap2(p);
+        return;
+      case 4:
+        Self::Swap4(p);
+        return;
+      case 8:
+        Self::Swap8(p);
+        return;
+      default:
+        itkGenericExceptionMacro("Cannot swap number of bytes requested");
+    }
   }
 }
 
-#endif
-
-#ifdef CMAKE_WORDS_BIGENDIAN
 template <typename T>
 void
-ByteSwapper<T>::SwapRangeFromSystemToBigEndian(T *, BufferSizeType)
+ByteSwapper<T>::SwapRangeFromSystemToBigEndian([[maybe_unused]] T * p, [[maybe_unused]] BufferSizeType num)
 {
-  // nothing needs to be done here...
-}
-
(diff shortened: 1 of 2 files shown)


dzenanz (approved):

N-Dekker:
Still wondering, shouldn't `ByteSwapper` use `ITK_WORDS_BIGENDIAN`, rather than `CMAKE_WORDS_BIGENDIAN`?

Both of them are defined by `<bin>/ITK/Modules/Core/Common/itkConfigure.h`:
https://github.com/InsightSoftwareConsortium/ITK/blob/f8d8cf94c9c316cc6a7c4a28d86105951b5ecb06/Modules/Core/Common/src/itkConfigure.h.in#L27-L41

hjmjohnson (approved):
Nice!  Much cleaner and easier to understand.

Modules/Core/Common/include/itkByteSwapper.hxx
```diff
@@ -39,202 +39,170 @@ namespace itk
...
-#ifdef CMAKE_WORDS_BIGENDIAN
-template <typename T>
-void
-ByteSwapper<T>::SwapFromSystemToBigEndian(T *)
-{}
-#else
+
 template <typename T>
 void
 ByteSwapper<T>::SwapFromSystemToBigEndian(T * p)
```
N-Dekker, line 45: We may need to add `[[maybe_unused]]` here now!

seanm:
I do still have a working PowerPC G5 Mac, running recent OpenBSD, which I could try with to see if it breaks anything...

Modules/Core/Common/include/itkByteSwapper.hxx
```diff
@@ -39,202 +39,170 @@ namespace itk
...
-      return;
-    case 4:
-      Self::Swap4Range(p, num);
-      return;
-    case 8:
-      Self::Swap8Range(p, num);
-      return;
-    default:
-      itkGenericExceptionMacro("Cannot swap number of bytes requested");
+    switch (sizeof(T))
```
blowekamp, line 74: I was initially going to comment that this switch statement is duplicated code and could be factored into a function... But isn't sizeof(T) known at compile time? We should be able to do a better Swap range function to avoid duplication.

Modules/Core/Common/include/itkByteSwapper.hxx
N-Dekker: Indeed, the original PR did cause warnings, at https://open.cdash.org/viewBuildError.php?type=1&buildid=9892277, like:

> itkByteSwapper.hxx:153:55: warning: parameter 'p' set but not used [-Wunused-but-set-parameter]

These should all be prevented now, with [the force-push that I just did](https://github.com/InsightSoftwareConsortium/ITK/compare/3d4f5251cb6ad032a21b4be00ad551230233c3ab..a38ef793f4e7dbdd964f7d57c1bfc1e571e9670d).

Modules/Core/Common/include/itkByteSwapper.hxx
N-Dekker: @blowekamp It looks like the existing `Swap2`, `Swap4`, and `Swap8` could be combined into some "SwapN" member function, like:

```cpp
void
ByteSwapper<T>::SwapN(void * ptr)
{
  auto * p = static_cast<char *>(ptr);

  for (size_t i{}; i < (N / 2); ++i)
  {
    std::swap(p[i], p[(N - 1) - i]);
  }
}
```
Would you like that? Anyway, maybe nice for a follow-up pull request  😃 

For comparison, the old `Swap8` implementation is at: https://github.com/InsightSoftwareConsortium/ITK/blob/e9b3d24f782a42f5586169e048b8d289f869d78a/Modules/Core/Common/include/itkByteSwapper.hxx#L350-L361

thewtex (approved):
:sparkler:

N-Dekker:
I think this pull request is ready.  Bradley (@blowekamp) made a nice suggestion to further reduce code duplication in ByteSwapper, https://github.com/InsightSoftwareConsortium/ITK/pull/4841#discussion_r1751777344 but I would prefer to make that a separate PR, when this one is merged.

PR #4843, opened 2024-09-09
BUG: Remove SpatialObjectProperty's writable string methods from SWIG

### LINKED ISSUE #4779: Wrapping of member functions that are passed and/or return std:string goes wonky
The issues are demonstrated in https://github.com/InsightSoftwareConsortium/ITK/pull/4776

1) SWIG seems to be converting some member functions that are passed two std::strings as args to new versions of functions that returns a std::string * (rather than returning a void as defined by the C++ code). 

This documentation, though old, might be relevant...
https://www.swig.org/Doc1.3/Arguments.html#Arguments_nn2

That documented feature would cause a function in C++ that returns a void to wrap to a python function that returns a value - hence causing the SetTagValueString to appear to return a value in python.

2) Additionally,  SWIG converts return by value to return by pointer:
https://www.swig.org/Doc1.3/SWIG.html#SWIG_nn23 

So,  that is where the std::string * references are coming from.

**Summary**

I suspect there is a bug in our SWIG wrapping specification for these reasons and because new SWIG should handled std::string fine and these wonkynesses shouldn't be happening.

blowekamp:
SWIG's binding for writable std::string objects is a "std::string *" which is not implemented.

closes #4779 

 replaces #4776

## PR Checklist

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkSpatialObjectProperty.h
+++ b/Modules/Core/SpatialObjects/include/itkSpatialObjectProperty.h
@@ -96,11 +96,13 @@ public:
     m_Name = name;
   }
 
+#if !defined(ITK_WRAPPING_PARSER)
   std::string &
   GetName()
   {
     return m_Name;
   }
+#endif
 
   const std::string &
   GetName() const
@@ -123,15 +125,13 @@ public:
     return value;
   }
 
+
+#if !defined(ITK_WRAPPING_PARSER)
   bool
   GetTagStringValue(const std::string & tag, std::string & value) const;
+#endif
   std::string
-  GetTagStringValue(const std::string & tag) const
-  {
-    std::string value = "";
-    this->GetTagStringValue(tag, value);
-    return value;
-  }
+  GetTagStringValue(const std::string & tag) const;
 
 
   std::map<std::string, double> &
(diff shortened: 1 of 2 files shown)


blowekamp:
Seems to work:
image

dzenanz (commented):
LGTM, but I defer to @aylward

aylward (approved):

Modules/Core/SpatialObjects/include/itkSpatialObjectProperty.h
```diff
@@ -123,15 +125,13 @@ class ITKSpatialObjects_EXPORT SpatialObjectProperty
     return value;
   }
 
+
+#if !defined(SWIG) && !defined(ITK_WRAPPING_PARSER)
```
blowekamp, line 129: I think this should only be dependent on ITK_WRAPPING_PARSER

blowekamp:
@aylward @thewtex  Does this fix need to be applied to the release branch too?

thewtex:
@blowekamp thanks for following up with a patch on this!

> @aylward @thewtex Does this fix need to be applied to the release branch too?

Yes, done! https://github.com/InsightSoftwareConsortium/ITK/commit/efe773f555a567079b07a47eb769d10ff7b56671

PR #4845, opened 2024-09-10
BUG: Fix itkFrequencyImageRegionIterator

phcerdan:
Test it via recently added gtest (see #4827)

Closes #4828

## PR Checklist

### DIFF
--- a/Modules/Filtering/ImageFrequency/include/itkFrequencyImageRegionConstIteratorWithIndex.h
+++ b/Modules/Filtering/ImageFrequency/include/itkFrequencyImageRegionConstIteratorWithIndex.h
@@ -174,8 +174,11 @@ private:
   void
   Init()
   {
-    this->m_FrequencyOrigin = this->m_Image->GetOrigin();
-    this->m_FrequencySpacing = this->m_Image->GetSpacing();
+    for (unsigned int dim = 0; dim < ImageType::ImageDimension; ++dim)
+    {
+      this->m_FrequencyOrigin[dim] = this->m_Image->GetOrigin()[dim];
+      this->m_FrequencySpacing[dim] = this->m_Image->GetSpacing()[dim];
+    }
   }
 
   FrequencyType m_FrequencyOrigin;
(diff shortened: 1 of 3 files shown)


Modules/Filtering/ImageFrequency/include/itkFrequencyImageRegionConstIteratorWithIndex.h
```diff
@@ -174,8 +174,10 @@ class ITK_TEMPLATE_EXPORT FrequencyImageRegionConstIteratorWithIndex : public Im
   void
   Init()
   {
-    this->m_FrequencyOrigin = this->m_Image->GetOrigin();
-    this->m_FrequencySpacing = this->m_Image->GetSpacing();
+    for (unsigned int dim = 0; dim < ImageType::ImageDimension; ++dim) {
```
dzenanz, line 177: Was this set of changes necessary?

dzenanz (approved):
LGTM

Modules/Filtering/ImageFrequency/include/itkFrequencyImageRegionConstIteratorWithIndex.h
phcerdan: Yeah, different types

```cpp
Modules/Filtering/ImageFrequency/include/itkFrequencyImageRegionConstIteratorWithIndex.h:177:29:
 error: no match for ‘operator=’ (operand types are ‘itk::FrequencyImageRegionConstIteratorWithIndex<itk::Image<double, 3> >::FrequencyType’ {aka ‘itk::Vector<double, 3>’} and
  ‘const itk::ImageBase<3>::PointType’ {aka ‘const itk::Point<double, 3>’})
  177 |     this->m_FrequencyOrigin = this->m_Image->GetOrigin();
  ```

N-Dekker:
Thanks @phcerdan but are those two iterators really useful? Given the fact that these two did not compile in the six years of their existence, why not remove them by now?

If they are really useful (next to all the other iterator types that we have already in ITK), can you please add a test or example to show their usefulness?

Modules/Filtering/ImageFrequency/include/itkFrequencyImageRegionConstIteratorWithIndex.h
N-Dekker: Is there a specific reason why `m_FrequencyOrigin` and `m_Image->GetOrigin()` have different types? I mean, why not declare  `m_FrequencyOrigin` as an `itk::Point` as well?

phcerdan: I don't exactly remember the reasons, but I think Vector has more math operations operators than Point.

All the other frequency iterators are also itk::Vector. If you feel like changing it, open an issue and let's discuss it, but it is out of the scope of this fix.

phcerdan:
Done

Issue #4839, opened 2024-09-07
Support specification of Nifti sform tolerance via environmental variable


thewtex:
Per discussion with @effigies , it would be helpful to enable specification of a higher nifti sform tolerance via environmental variable.

There are cases where nifti headers, perhaps edited manually, that may have unintentional small shear, it would be helpful to specify an override via environmental variable a tolerance for orthogonality.

This would be used in docker images that use FSL, ANTs, etc.

Possibly related?

- https://github.com/InsightSoftwareConsortium/ITK/blob/fb61d2cadc33bf50d1f0ec5b842b70a1a7021d0e/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx#L1989-L2010

Todo:

- [ ] find a test case that demonstrates the issue
- [ ] create a patch (if needed) to enable specifying the related tolerance via environmental variable
- [ ] look into updated associated warnings to suggest setting the environmental variable if appropriate

CC @hjmjohnson

effigies:
Just adding a note that I'm having trouble finding records of when we were hitting this. My memory was that there were small shears causing ITK to reject images, but it could also be differences between the sform column norms and `pixdim[1:4]`, which I believe is what this particular code is handling.

Will keep my eyes out for the next one. When we've encountered it, it's generally been due to a failure to validate and update headers before passing to ANTs, so it's possible we won't trigger the case again until the next refactor leaves a hole for a messy image to slip through.

cookpa:
I think this is already available via #4009 

https://github.com/InsightSoftwareConsortium/ITK/blob/fb61d2cadc33bf50d1f0ec5b842b70a1a7021d0e/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx#L2080-L2123

This works for me:

```
export ITK_NIFTI_SFORM_PERMISSIVE=1
PrintHeader bad_sform.nii
```

cookpa:
BTW, the expected values of `ITK_NIFTI_SFORM_PERMISSIVE` are true/false, on/off, yes/no. It cannot be toggled with 1/0. Is this ITK convention, or something to fix?

cookpa:
Sorry for not reading carefully before, I see the idea here is to add a numeric tolerance rather than the boolean we currently have - this seems like a good idea.

I had a go at creating a test image. However, I noticed that the permissive mode introduces a flip. 

It's very possible I'm doing something wrong here, so here's R code to create the dummy images, starting from a template

```R
[...]

PR #4847, opened 2024-09-12
BUG: permissive cosine computation from non-orthogonal sform had reve…

cookpa:
…rsed SVD components

U and V were reversed in the reconstruction of M, according to definitions here

https://vxl.github.io/doc/release/core/vnl/html/classvnl__svd.html

As discussed in #4839 

cc: @vfonov @hjmjohnson 

## PR Checklist

### DIFF
--- a/Modules/IO/NIFTI/test/itkNiftiReadWriteDirectionTest.cxx
+++ b/Modules/IO/NIFTI/test/itkNiftiReadWriteDirectionTest.cxx
@@ -17,6 +17,7 @@
  *=========================================================================*/
 #include <iostream>
 #include <fstream>
+#include "itkMath.h"
 #include "itkNiftiImageIO.h"
 #include "itkImageFileReader.h"
 #include "itkImageFileWriter.h"
@@ -79,8 +80,8 @@ itkNiftiReadWriteDirectionTest(int argc, char * argv[])
   {
     std::cerr << "Missing Parameters." << std::endl;
     std::cerr << "Usage: " << itkNameOfTestExecutableMacro(argv)
-              << "imageWithBothQAndSForms imageWithNoQform imageWithNoSform  imageWithNonOrthoSform testOutputDir"
-              << std::endl;
+              << "imageWithBothQAndSForms imageWithNoQform imageWithNoSform imageWithNonOrthoSform testOutputDir "
+              << " [permissive_sform_threshold_deg=8.0]" << std::endl;
     std::cerr << "6 arguments required, received " << argc << std::endl;
     for (int i = 0; i < argc; ++i)
     {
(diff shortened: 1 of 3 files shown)


cookpa:
I will see if I can modify the existing tests to catch this

vfonov:
oops, looks like my typo

Modules/IO/NIFTI/test/itkNiftiReadWriteDirectionTest.cxx
```diff
@@ -205,17 +205,45 @@ itkNiftiReadWriteDirectionTest(int argc, char * argv[])
...
+  // Compute R_relative = R1 * R2^-1
+  auto rRelative = inputImageNoQformDirection * nonOrthoDirectionInv;
+
+  // Calculate the trace of the relative rotation matrix
+  double trace = rRelative[0][0] + rRelative[1][1] + rRelative[2][2];
+
+  // Calculate the angle of rotation between the two matrices
+  double angle = std::acos((trace - 1.0) / 2.0);
+
+  if (angle > 0.15)
```
jhlegarreta, line 239: Is this particular value chosen for some reason? If it is an arbitrary value, even if not used elsewhere, it should probably be defined as a variable and a comment be added to make it clear that it is a sensitive value. Better than that it would be to define a variable in the corresponding class, as commented in the issue.

Also, please print the value of the threshold in the error message.

jhlegarreta (commented):
Thanks for doing this @cookpa :100:.

An inline comment.

Modules/IO/NIFTI/test/itkNiftiReadWriteDirectionTest.cxx
cookpa: It's an arbitrary value, somewhat larger than the actual error from permissive mode (~0.1) but small enough to catch the bug that is fixed in this PR. The threshold will depend on the input data. I agree that if we switch from permissive mode on/off to a numeric threshold, it would make sense to se that in the tests. Meanwhile, I'll make it a local variable with a comment

cookpa: Alternatively I could make it an argument to the script, if that is OK? `itkNiftiReadWriteDirectionTest` is called a couple of times with different images, it might make sense to use different thresholds

jhlegarreta: Maybe an optional one? If we eventually transition towards a CMake variable or a class ivar, the test can be modified to adopt that one if not provided as an input argument?

Modules/IO/NIFTI/test/itkNiftiReadWriteDirectionTest.cxx
```diff
@@ -205,17 +205,45 @@ itkNiftiReadWriteDirectionTest(int argc, char * argv[])
...
+  itksys::SystemTools::PutEnv("ITK_NIFTI_SFORM_PERMISSIVE=NO");
+
   // check the resulting rotation matrix is orthogonal
   if (!CheckRotation<TestImageType>(inputImageNonOrthoSform))
   {
     std::cerr << "Rotation matrix after correcting is not orthgonal" << std::endl;
     return EXIT_FAILURE;
   }
 
+  // check similarity between direction without shear (argv[2]) and direction with shear (argv[4])
```
jhlegarreta, line 221: Minor detail: start comment sentence with capitals.

Modules/IO/NIFTI/test/itkNiftiReadWriteDirectionTest.cxx
cookpa: I made the threshold an argument to the script, with a default.

jhlegarreta: Thanks @cookpa. Out of curiosity, why are the values one magnitude larger than the threshold that you used previously (re https://github.com/InsightSoftwareConsortium/ITK/pull/4847#discussion_r1759588013)?

cookpa: I changed the threshold to degrees when making it a command line parameter, I figured it would be easier

jhlegarreta: :+1: Oh, should have figured out. Thanks Philip.

cookpa: No problem, thanks for your suggestions

dzenanz (commented):
Is this ready for merging?

thewtex (approved):
@cookpa well done!

Issue #4848, opened 2024-09-13
`PointSet::SetPoints(PointsVectorContainer *)` overload leads to undefined behavior


N-Dekker:
### Description

`PointSet::SetPoints(PointsVectorContainer * points)`  internally casts its argument to a `PointsContainer *`:

https://github.com/InsightSoftwareConsortium/ITK/blob/9a5dade7f7e13fcef307f8224a01c2ce2b0ab13b/Modules/Core/Common/include/itkPointSet.hxx#L76

This leads to undefined behavior, when the `PointsVectorContainer` object is being used as a `PointsContainer` object.

### Steps to Reproduce

The test code already has undefined behavior, even if it might just work on the currently tested platforms:

https://github.com/InsightSoftwareConsortium/ITK/blob/9a5dade7f7e13fcef307f8224a01c2ce2b0ab13b/Modules/Core/Common/test/itkPointSetTest.cxx#L102-L127

### Versions

- The `PointSet::SetPoints(PointsVectorContainer *)` overload  is introduced by pull request #3154 commit 6827360fa6a06fbe073911d7f8554c2c3368721d "ENH: Adding SetPoints in PointSet for 1D input array", merged on Mar 16, 2022, and included with both [v5.3.0](https://github.com/InsightSoftwareConsortium/ITK/releases/tag/v5.3.0) and [v5.4.0](https://github.com/InsightSoftwareConsortium/ITK/releases/tag/v5.4.0).

### Additional Information

Discussed before at https://github.com/InsightSoftwareConsortium/ITK/pull/3154#discussion_r1755213148, with @PranjalSahu

ebrahimebrahim:
Note that the problem has moved to here since #4867:

https://github.com/InsightSoftwareConsortium/ITK/blob/ec7569203a2f9a5c8febba3fad89b1dde5c0cd36/Modules/Core/Common/include/itkPointSetBase.hxx#L72-L73

N-Dekker:
Would it be OK if we would just deprecate, and eventually remove this particular `SetPoints(PointsVectorContainer *)` overload? @PranjalSahu What do you think?

PranjalSahu:
Yes, it can be deprecated. But for the future, we should think of a way for Python to allow direct point setting without creating a new copy. Because there are use cases where the meshes can be in GB size and for performing some simple filtering, a new copy would be too much.

PR #4850, opened 2024-09-17
Add `SetPointsByCoordinates` to PointSet, add warning to `SetPoints(PointsVectorContainer *)`

N-Dekker:
- Added a warning to `PointSet::SetPoints(PointsVectorContainer *)`, mitigating issue #4848 
- Added `PointSet::SetPointsByCoordinates` as a safe alternative.

### DIFF
--- a/Modules/Core/Common/include/itkPointSet.hxx
+++ b/Modules/Core/Common/include/itkPointSet.hxx
@@ -73,6 +73,8 @@ PointSet<TPixelType, VDimension, TMeshTraits>::SetPoints(PointsVectorContainer *
   {
     itkExceptionMacro("Number of entries in given 1d array incompatible with the point dimension");
   }
+
+  // Note: this cast is unsafe. It may lead to undefined behavior.
   auto * pointsPtr = reinterpret_cast<PointsContainer *>(points);
 
   m_PointsContainer = pointsPtr;
--- a/Modules/Core/Common/include/itkPointSet.hxx
+++ b/Modules/Core/Common/include/itkPointSet.hxx
@@ -81,6 +81,56 @@ PointSet<TPixelType, VDimension, TMeshTraits>::SetPoints(PointsVectorContainer *
   this->Modified();
 }
 
+
+template <typename TPixelType, unsigned int VDimension, typename TMeshTraits>
+void
+PointSet<TPixelType, VDimension, TMeshTraits>::SetPointsByCoordinates(const std::vector<CoordRepType> & coordinates)
+{
+  itkDebugMacro("Setting the points to the specified coordinates");
+
+  const size_t numberOfCoordinates = coordinates.size();
+
+  if (numberOfCoordinates % PointDimension != 0)
+  {
+    itkExceptionMacro("Number of specified coordinates incompatible with the point dimension");
+  }
+
+  const size_t numberOfPoints = numberOfCoordinates / PointDimension;
+
+  if (m_PointsContainer == nullptr)
+  {
+    m_PointsContainer = PointsContainer::New();
+  }
+
+  using STLContainerType = typename PointsContainer::STLContainerType;
+
+  STLContainerType & points = m_PointsContainer->CastToSTLContainer();
+  points.clear();
+
+  if constexpr (std::is_same_v<STLContainerType, std::vector<PointType>>)
+  {
+    // STLContainerType is either an std::vector or an std::map. Only when it is an std::vector, it should be resized.
+    // std::map does not have a resize function.
+    points.resize(numberOfPoints);
+  }
+  else
+  {
+    static_assert(std::is_same_v<STLContainerType, std::map<PointIdentifier, PointType>>);
+  }
+
+  auto coordinateIterator = coordinates.cbegin();
+
+  for (PointIdentifier pointIdentifier{}; pointIdentifier < numberOfPoints; ++pointIdentifier)
+  {
+    PointType & point = points[pointIdentifier];
+    std::copy_n(coordinateIterator, PointDimension, point.begin());
+    coordinateIterator += PointDimension;
+  }
+
+  this->Modified();
+}
+
+
 template <typename TPixelType, unsigned int VDimension, typename TMeshTraits>
 auto
 PointSet<TPixelType, VDimension, TMeshTraits>::GetPoints() -> PointsContainer *
--- a/Modules/Core/Common/include/itkPointSet.h
+++ b/Modules/Core/Common/include/itkPointSet.h
@@ -169,6 +169,10 @@ public:
   void
   SetPoints(PointsVectorContainer *);
 
+  /** Sets the points by specifying its coordinates. */
+  void
+  SetPointsByCoordinates(const std::vector<CoordRepType> & coordinates);
+
   /** Get the points container. */
   PointsContainer *
   GetPoints();
--- a/Modules/Core/Common/include/itkPointSet.h
+++ b/Modules/Core/Common/include/itkPointSet.h
@@ -164,7 +164,8 @@ public:
   void
   SetPoints(PointsContainer *);
 
-  /** Set the points container using a 1D vector. */
+  /** Set the points container using a 1D vector.
+  \warning This member function is unsafe. It may just work, but it may also lead to undefined behavior. */
   void
   SetPoints(PointsVectorContainer *);
 
--- a/Modules/Core/Common/test/CMakeLists.txt
+++ b/Modules/Core/Common/test/CMakeLists.txt
@@ -1748,6 +1748,7 @@ set(ITKCommonGTests
     itkOffsetGTest.cxx
     itkOptimizerParametersGTest.cxx
     itkPointGTest.cxx
+    itkPointSetGTest.cxx
     itkRGBAPixelGTest.cxx
     itkRGBPixelGTest.cxx
     itkShapedImageNeighborhoodRangeGTest.cxx
--- a/Modules/Core/Common/test/itkPointSetGTest.cxx
+++ b/Modules/Core/Common/test/itkPointSetGTest.cxx
@@ -1,0 +1,82 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         https://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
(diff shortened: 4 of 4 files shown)


N-Dekker:
> Perhaps write a Python test to demonstrate that?

I'm sorry I'm not very good in Python tests. 🤷  But it looks like SWIG supports passing a Python tuple or list as argument, when the parameter type is `std::vector`.

PranjalSahu:
> If I remember correctly, the main point of the old method was convenient usage from Python. @PranjalSahu is that correct?
> 
> @N-Dekker how would use of the proposed method look like from Python? Perhaps write a Python test to demonstrate that?

Yes, another library supported such functionality of directly passing a 1-D vector of points in any dimension, so we wanted to have a similar feature here. It would be better to find an alternative to **reinterpret_cast** instead of adding a new method.

If there are no alternatives then yes we can add a new method.

Also did @N-Dekker did you come across any platform where this method is breaking?

N-Dekker:
> did you come across any platform where this method is breaking?

Honestly, I did not. But it's really quite a coincidence that the `reinterpret_cast<PointsContainer *>(points)` just seems to work so far. As you know, it internally "reinterprets" the bits of an `std::vector<CoordRepType>` as an `std::vector<PointType>`.  Suppose the `std::vector` would internally have a `m_NumberOfElements` member variable.  This member would then have the same value after the "reinterpretation", as before. But that would be wrong, of course.  The number of elements of `std::vector<PointType>` is less than the number of elements of `std::vector<CoordRepType>`.

In general, undefined behavior should be avoided because the compiler (especially the optimizer) may mess up things as it is allowed to assume that there is no undefined behavior.

PranjalSahu:
ok it will be great if we could create a scenario where it fails!

My understanding is that it is converting the pointer type of the continuous memory block where the 1D vector of 1D points of length 3N is stored, to a vector of length N of 3D Points. Essentially the same as Numpy reshape operator (1 x 3N  -> 3xN).

This is needed because only 1D vectors are supported while converting the Numpy vector to VectorContainer and vice-versa.

Modules/Core/Common/include/itkPointSet.hxx
```diff
@@ -73,12 +73,79 @@ PointSet<TPixelType, VDimension, TMeshTraits>::SetPoints(PointsVectorContainer *
...
+  if ((m_PointsContainer != nullptr) && (m_PointsContainer->size() == numberOfPoints))
+  {
+    const auto & points = m_PointsContainer->CastToSTLConstContainer();
+
+    const bool equalCoordinates = [&points, &coordinates] {
+      auto coordinateIterator = coordinates.cbegin();
+
+      for (const auto & point : points)
+      {
+        if (!std::equal(point.cbegin(), point.cend(), coordinateIterator))
```
N-Dekker, line 109: Interesting errors at https://open.cdash.org/viewBuildError.php?buildid=9909255, please have a look:

```
/.../s/Modules/Core/Common/include/itkPointSet.hxx: In instantiation of 'void itk::PointSet<TPixelType, VDimension, TMeshTraits>::SetPointsByCoordinates(const std::vector<typename TMeshTraits::CoordRepType>&) [with TPixelType = double; unsigned int VDimension = 2; TMeshTraits = itk::QuadEdgeMeshTraits<double, 2, bool, bool, float, float>; typename TMeshTraits::CoordRepType = float]':
Wrapping/Modules/ITKQuadEdgeMesh/itkQuadEdgeMeshBasePython.cpp:11546:97:   required from here

Modules/Core/Common/include/itkPointSet.hxx:109:31: error: 'const struct std::pair<const long unsigned int, itk::QuadEdgeMeshPoint<float, 2, itk::GeometricalQuadEdge<long unsigned int, long unsigned int, bool, bool, true> > >' has no member named 'cbegin'
```

It appears that `PointsContainer` may be an `itk::MapContainer`, rather than just an `itk::VectorContainer`! 

https://github.com/InsightSoftwareConsortium/ITK/blob/9a5dade7f7e13fcef307f8224a01c2ce2b0ab13b/Modules/Core/QuadEdgeMesh/include/itkQuadEdgeMeshTraits.h#L87 

Obviously `reinterpret_cast<PointsContainer *>(points)` won't work when `PointsContainer` is an instance of `itk::MapContainer`. But my pull request should also still be adjusted to support `itk::MapContainer`, apparently 🤷

N-Dekker:
@PranjalSahu Please try the following:

```cpp
#include "itkPointSet.h"
#include "../../QuadEdgeMesh/include/itkQuadEdgeMeshTraits.h"
#include <gtest/gtest.h>

TEST(PointSet, SetPoints)
{
  static constexpr auto PointDimension = 2;

  using MeshTraits = itk::QuadEdgeMeshTraits<double, PointDimension, bool, bool>;
  using PointSetType = itk::PointSet<double, PointDimension, MeshTraits>;
  using CoordRepType = PointSetType::CoordRepType;

  auto pointSet = PointSetType::New();
  auto pointsVectorContainer = PointSetType::PointsVectorContainer::New();
  auto numberOfPoints = 42;
  pointsVectorContainer->CastToSTLContainer() = std::vector<CoordRepType>(numberOfPoints * PointDimension);
  pointSet->SetPoints(pointsVectorContainer);
  auto points = pointSet->GetPoints();

  EXPECT_EQ(points->size(), numberOfPoints);
}
```

It appears that after calling your `SetPoints(PointsVectorContainer *)` overload, the point-set has got an arbitrary number of points. While it should have been 42! 🤷 (Don't blame me for the MeshTraits 😇 )

PranjalSahu:
Ok yes, it was not designed to be used with such complex point types.
So yes, either we can create a new method as you suggest or add a comment to not use it with complex Point types.

N-Dekker:
By the way, I proposed a different name for the added member function, "SetPointsByCoordinates", because it is fundamentally different from the two existing SetPoints overloads. Those overloads both "share" the ownership of the object passed as argument (using ITK's reference counting mechanism).  Whereas "SetPointsByCoordinates" copies the coordinates from the specified argument.

Personally I think the name "SetPointsByCoordinates" is fine, but it could also be named, for example, "SetPointsFromCoordinates" or "CopyCoordinatesToPoints". A matter of taste, I guess  🤷

PranjalSahu:
Yes, that was also one of the requirements. Because the number of points could be large, we wanted to avoid copying the data.

N-Dekker:
> Yes, that was also one of the requirements. Because the number of points could be large, we wanted to avoid copying the data.

How large could the number of points possibly be, in your application? Can you please give the `SetPointsByCoordinates` member function that I'm proposing a chance?

Undefined behavior is just not an option, for any professional application or any serious open source project.

PranjalSahu:
We will have to add tests for Python wrapping also.

N-Dekker:
> We will have to add tests for Python wrapping also.

OK, thanks for reminding me, @PranjalSahu, but Python tests can also be added in a follow-up pull request. The given PR does not break any existing use case. While it does add a valid C++ use case, already.

(I like to keep pull requests small, so that they can be reviewed easily, and merged smoothly.)

Modules/Core/Common/include/itkPointSet.hxx
```diff
@@ -81,6 +81,50 @@ PointSet<TPixelType, VDimension, TMeshTraits>::SetPoints(PointsVectorContainer *
...
+  }
+
+  using STLContainerType = typename PointsContainer::STLContainerType;
+
+  STLContainerType & points = m_PointsContainer->CastToSTLContainer();
+  points.clear();
+
+  if constexpr (std::is_same_v<STLContainerType, std::vector<PointType>>)
+  {
+    points.resize(numberOfPoints);
```
dzenanz, line 112: Why is this inside an `if`? Isn't it always needed?

N-Dekker: Good question... as far as I observed, the STLContainerType of the PointsContainer is either an std::vector or an std::map. If it is an std::map, the expression `points[pointIdentifier]` just works well without `points.resize(numberOfPoints)`. And actually, std::map does not have a resize member function.

Should there be an extra line of comment, explaining why?

dzenanz:
```suggestion
    points.resize(numberOfPoints);  // std::map does not have a resize function
```

N-Dekker: Yes indeed, something like that, thanks. I'll do a little amend + force-push.

N-Dekker: Ideally the code should be more like:

```
if constexpr ( `points.resize(numberOfPoints)` is a valid expression )
{
  points.resize(numberOfPoints);
}
```

But I found it easier to just write `if constexpr (std::is_same_v<STLContainerType, std::vector<PointType>>)`  😃

N-Dekker: Please check my [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/37e1b83e065f07264e55450a55d02ba5db376738..9cb3f7c7ffdb76c1efcd9077e35aaf1324a70935)  😃

N-Dekker: By the way, I could still add an `else` clause, having `static_assert(std::is_same_v<STLContainerType, std::map>)`. If anyone would like that!

dzenanz (approved):

Modules/Core/Common/include/itkPointSet.hxx
N-Dekker: @dzenanz Here it is, please have a look: [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/9cb3f7c7ffdb76c1efcd9077e35aaf1324a70935..60d183c0d6e12034f61dcfaa52e95e90bca8290e)

dzenanz (approved):

blowekamp (approved):
This looks like a good and clean addition.

It only adds comments about an unsafe methods. These additional request for testing and compatibility ( python api?) could be done if the unsafe methods is marked for deprecation or removed in future work.

PranjalSahu:
> > We will have to add tests for Python wrapping also.
> 
> OK, thanks for reminding me, @PranjalSahu, but Python tests can also be added in a follow-up pull request. The given PR does not break any existing use case. While it does add a valid C++ use case, already.
> 
> (I like to keep pull requests small, so that they can be reviewed easily, and merged smoothly.)
> 
> PS My first impression is that passing an `np.array([1.0, 2.0, 3.0])` argument to a Python wrapped C++ function whose parameter type is `const std::vector<double>&` just works well. 👍 (I just tried it for [ITKElastix](https://github.com/InsightSoftwareConsortium/ITKElastix).)

Before removing this it would be great if the Python wrapping could be tested and merged. 
Because the only purpose to add the vectorcontainer support was for Python wrapping.

N-Dekker:
Is there a Python wheel for Windows available, based on the latest commit, including this pull request? I would like to try out the new `SetPointsByCoordinates` member function in Python.

dzenanz:
A Python wheel is not built automatically. You need a local build with [wrapping enabled](https://vtk.org/Wiki/ITK/Python_Wrapping). Then, copy `C:\Dev\ITK-py22\Wrapping\Generators\Python\WrapITK.pth` to `your_venv/Lib/site-packages`. After restarting Python interpreter ITK should be available.

PR #4855, opened 2024-09-20
STYLE: Clean-up ByteSwapper, using private `SwapBytes` helper function

N-Dekker:
Aims to reduce code duplication. Inspired by a suggestion from Bradley (@blowekamp) at https://github.com/InsightSoftwareConsortium/ITK/pull/4841/files#r1751777344

### DIFF
--- a/Modules/Core/Common/include/itkByteSwapper.hxx
+++ b/Modules/Core/Common/include/itkByteSwapper.hxx
@@ -28,7 +28,8 @@
 #ifndef itkByteSwapper_hxx
 #define itkByteSwapper_hxx
 #include "itkMakeUniqueForOverwrite.h"
-#include <algorithm> // For swap.
+#include <algorithm> // For for_each_n and swap.
+#include <cstddef>   // For byte.
 #include <memory>
 #include <cstring>
 
@@ -44,24 +45,9 @@ template <typename T>
 void
 ByteSwapper<T>::SwapFromSystemToBigEndian([[maybe_unused]] T * p)
 {
-  if constexpr (!m_SystemIsBigEndian)
+  if constexpr (!m_SystemIsBigEndian && sizeof(T) > 1)
   {
-    switch (sizeof(T))
-    {
-      case 1:
-        return;
-      case 2:
-        Self::Swap2(p);
-        return;
-      case 4:
-        Self::Swap4(p);
-        return;
-      case 8:
-        Self::Swap8(p);
-        return;
-      default:
-        itkGenericExceptionMacro("Cannot swap number of bytes requested");
-    }
+    SwapBytes(*p);
   }
 }
 
@@ -69,24 +55,9 @@ template <typename T>
 void
 ByteSwapper<T>::SwapRangeFromSystemToBigEndian([[maybe_unused]] T * p, [[maybe_unused]] BufferSizeType num)
 {
-  if constexpr (!m_SystemIsBigEndian)
+  if constexpr (!m_SystemIsBigEndian && sizeof(T) > 1)
   {
-    switch (sizeof(T))
-    {
-      case 1:
-        return;
-      case 2:
-        Self::Swap2Range(p, num);
-        return;
-      case 4:
-        Self::Swap4Range(p, num);
-        return;
-      case 8:
-        Self::Swap8Range(p, num);
-        return;
-      default:
-        itkGenericExceptionMacro("Cannot swap number of bytes requested");
-    }
+    std::for_each_n(p, num, [](T & element) { SwapBytes(element); });
   }
 }
 
@@ -127,24 +98,9 @@ template <typename T>
 void
 ByteSwapper<T>::SwapFromSystemToLittleEndian([[maybe_unused]] T * p)
 {
-  if constexpr (m_SystemIsBigEndian)
+  if constexpr (m_SystemIsBigEndian && sizeof(T) > 1)
   {
-    switch (sizeof(T))
-    {
-      case 1:
-        return;
-      case 2:
-        Self::Swap2(p);
-        return;
-      case 4:
-        Self::Swap4(p);
-        return;
-      case 8:
-        Self::Swap8(p);
-        return;
-      default:
-        itkGenericExceptionMacro("Cannot swap number of bytes requested");
-    }
+    SwapBytes(*p);
   }
 }
 
@@ -152,24 +108,9 @@ template <typename T>
 void
 ByteSwapper<T>::SwapRangeFromSystemToLittleEndian([[maybe_unused]] T * p, [[maybe_unused]] BufferSizeType num)
 {
-  if constexpr (m_SystemIsBigEndian)
+  if constexpr (m_SystemIsBigEndian && sizeof(T) > 1)
   {
-    switch (sizeof(T))
-    {
-      case 1:
-        return;
-      case 2:
-        Self::Swap2Range(p, num);
-        return;
-      case 4:
-        Self::Swap4Range(p, num);
-        return;
-      case 8:
-        Self::Swap8Range(p, num);
-        return;
-      default:
-        itkGenericExceptionMacro("Cannot swap number of bytes requested");
-    }
+    std::for_each_n(p, num, [](T & element) { SwapBytes(element); });
   }
 }
 
@@ -369,6 +310,33 @@ ByteSwapper<T>::SwapWrite8Range(const void * ptr, BufferSizeType num, OStreamTyp
     }
   }
 }
+
+
+// The following member function is private:
+
+template <typename T>
+void
+ByteSwapper<T>::SwapBytes(T & value)
+{
+  static constexpr size_t numberOfBytes = sizeof(T);
+
+  // Historically (from ITK v1.2.0, March 2003) the following number of bytes are supported:
+  if constexpr (numberOfBytes == 2 || numberOfBytes == 4 || numberOfBytes == 8)
+  {
+    // When the value is an integer, the following code is equivalent to `value = std::byteswap(value)`, in C++26:
+    auto * const bytes = reinterpret_cast<std::byte *>(&value);
+
+    for (size_t i{}; i < (numberOfBytes / 2); ++i)
+    {
+      std::swap(bytes[i], bytes[(numberOfBytes - 1) - i]);
+    }
+  }
+  else
+  {
+    itkGenericExceptionMacro("Cannot swap number of bytes requested");
+  }
+}
+
 } // end namespace itk
 
 #endif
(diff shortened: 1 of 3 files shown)


Modules/Core/Common/include/itkByteSwapper.hxx
```diff
@@ -369,6 +310,33 @@ ByteSwapper<T>::SwapWrite8Range(const void * ptr, BufferSizeType num, OStreamTyp
...
+    auto * const bytes = reinterpret_cast<std::byte *>(&value);
+
+    for (size_t i{}; i < (numberOfBytes / 2); ++i)
+    {
+      std::swap(bytes[i], bytes[(numberOfBytes - 1) - i]);
+    }
+  }
+  else
+  {
+    itkGenericExceptionMacro("Cannot swap number of bytes requested");
```
blowekamp, line 336: This is a run-time exception for a "constexpr" conditional?

blowekamp:
Looking like a nice cleanup!

Modules/Core/Common/include/itkByteSwapper.hxx
N-Dekker: Indeed! Only the _condition_ of an `if constexpr (condition) { ... }` must be evaluated at compile-time. Anything inside the `if` or `else` clause may be evaluated at run-time.

blowekamp: Yes. I think it would be an improvement and help detect errors in code.

Modules/Core/Common/include/itkByteSwapper.hxx
```diff
@@ -44,49 +45,19 @@ template <typename T>
...
-      case 4:
-        Self::Swap4Range(p, num);
-        return;
-      case 8:
-        Self::Swap8Range(p, num);
-        return;
-      default:
-        itkGenericExceptionMacro("Cannot swap number of bytes requested");
-    }
+    std::for_each_n(p, num, [](T & element) { SwapBytes(element); });
```
jhlegarreta, line 60: Forgive me if this is not possible, but this block is the same as in https://github.com/InsightSoftwareConsortium/ITK/pull/4855/files#diff-097dc3a747bd453b8fcfb3f846d65f65931572c23656e478ea40e591a4576bbcR113, so wondering if a private method can be created to be call from both functions.

Modules/Core/Common/include/itkByteSwapper.hxx
N-Dekker: Honestly it would be fine to me to also support other number of bytes. Why not support `numberOfBytes == 16`, for example? For example, when a user wants to swap [128-bit integers](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fint128.html)...!

I think the original `itkGenericExceptionMacro("Cannot swap number of bytes requested")` was there just because we only had three of those helper functions: `Swap2`, `Swap4`, and `Swap8`. Instead, we can now support just _any_ number of bytes! Would you like that?

dzenanz: That should only work on powers of 2, right? I think we have a way of doing that at compile time (in constexpr manner).

Modules/Core/Common/include/itkByteSwapper.hxx
N-Dekker: Sure, I can make a private `SwapRange` helper function to avoid having duplicate `std::for_each_n` call. 👍

dzenanz (approved):
Good even as-is.

Modules/Core/Common/include/itkByteSwapper.hxx
dzenanz: Additional level of indirection to save duplication of one line/function call might be counterproductive to understanding the code.

Modules/Core/Common/include/itkByteSwapper.hxx
N-Dekker: > That should only work on powers of 2, right?

@dzenanz Well... the currently supported numbers 2, 4, 8 suggest so. But maybe that's just because those are the sizes (number of bytes) of the most commonly used scalar types. 🤷  If anyone wants to swap a different number of bytes, no problem to me!

Instead of checking the size, we could also check [is_scalar](https://en.cppreference.com/w/cpp/types/is_scalar). Requiring that only scalars (integers, floating points, enum's, raw pointers etc.) are being byte-swapped. Would anyone like that? 

> I think we have a way of doing that at compile time (in constexpr manner).

Something like `static_assert(sqrt(numberOfBytes) * sqrt(numberOfBytes) == numberOfBytes)`? C++17 `std::sqrt` isn't yet constexpr, unfortunately 🤷

dzenanz: Maybe do `static_assert` for 2, 4, and 8 now, and think about what we want to do, if anything, in some follow-up PR?

Modules/Core/Common/include/itkByteSwapper.hxx
N-Dekker: > Additional level of indirection to save duplication of one line/function call might be counterproductive to understanding the code.

Thanks to clang-format that it is just one line! It might as well have been three lines of code:

```cpp
std::for_each_n(p, num, [](T & element)
  {
    SwapBytes(element);
  });
```

@jhlegarreta I see your point, and yes, technically a helper function is possible, but is it still acceptable for you to leave it as-is?

jhlegarreta: Acceptable.

dzenanz (approved):

Modules/Core/Common/include/itkByteSwapper.hxx
N-Dekker: @blowekamp wrote:

> I think it would be an improvement and help detect errors in code.

Indeed! `static_assert(numberOfBytes == 2 || numberOfBytes == 4 || numberOfBytes == 8)` has just helped us to find a bug! In the CI, https://open.cdash.org/viewBuildError.php?buildid=9915418 

```
itkByteSwapper.hxx: In instantiation of 'static void itk::ByteSwapper<T>::SwapBytes(T&) [with T = long double]':
itkByteSwapper.hxx:60:56:   required from 'static void itk::ByteSwapper<T>::SwapRangeFromSystemToBigEndian(T*, itk::ByteSwapper<T>::BufferSizeType) [with T = long double; itk::ByteSwapper<T>::BufferSizeType = long unsigned int]'
IO/MeshBase/include/itkMeshIOBase.h:679:65:   required from 'void itk::MeshIOBase::WriteBufferAsBinary(TInput*, std::ofstream&, itk::MeshIOBase::SizeValueType) [with TOutput = float; TInput = long double; std::ofstream = std::basic_ofstream<char>; itk::MeshIOBase::SizeValueType = long unsigned int]'
IO/MeshOFF/src/itkOFFMeshIO.cxx:582:119:   required from here

itkByteSwapper.hxx:324:58: error: static assertion failed
```

So in this case, `T = long double`! And on Linux, `long double` is more than `8` bytes!!!

How shall we fix this? Apparently ByteSwapper never worked before for such _long_ double's.

dzenanz: > Apparently ByteSwapper never worked before for such long double's.

It would have been a run-time error. The equivalent solution now would be to remove the [LDOUBLE case](https://github.com/InsightSoftwareConsortium/ITK/blob/v5.4.0/Modules/IO/MeshOFF/src/itkOFFMeshIO.cxx#L580-L585), so at run time [this exception](https://github.com/InsightSoftwareConsortium/ITK/blob/v5.4.0/Modules/IO/MeshOFF/src/itkOFFMeshIO.cxx#L588) is hit.

N-Dekker: @dzenanz  OK, so are you indeed proposing to just remove the LDOUBLE case for now? From https://github.com/InsightSoftwareConsortium/ITK/blob/311b7060ef39e371f3cd209ec135284ff5fde735/Modules/IO/MeshOFF/src/itkOFFMeshIO.cxx#L580-L585

dzenanz: Indeed I am, as I believe that would be a backwards compatible change.

I never worked on a big-endian system. Also, isn't IEEE 754 big endian? ISA implementation-defined, as per https://en.wikipedia.org/wiki/Endianness#Floating_point

N-Dekker: By the way, it appears that C++26 [std::byteswap](https://en.cppreference.com/w/cpp/numeric/byteswap) only supports integer types. Maybe byte-swapping is controversial for floating points?!?

N-Dekker: Still broken, now it's itkVTKPolyDataMeshIO, at https://open.cdash.org/viewBuildError.php?type=1&buildid=9915657:

> in instantiation of function template specialization 'itk::VTKPolyDataMeshIO::ReadPointsBufferAsBINARY<long double>' requested here

      CASE_INVOKE_BY_TYPE(ReadPointsBufferAsBINARY, inputFile)
                          ^

So would you also want us to drop `long double` support from `itk::VTKPolyDataMeshIO`?

dzenanz: Unsurprisingly, this endian swapping only affects IO classes.

As the other float types are being swapped, why not swap `long double` as well? Just add size of the `long double` to the list of the allowed sizes in the assert? @blowekamp what do you think?

blowekamp: I quick google said 128-bit float may be an abnormal case:
https://www.talospace.com/2018/12/the-saga-of-power-isa-128-bit-long.html

So... Just leave it as a run time exception of now?

N-Dekker: It would be fine to me to restore the `itkGenericExceptionMacro("Cannot swap number of bytes requested")` that I originally proposed, and remove the compile-time static_assert(numberOfBytes is 2 or 4 or 8).

dzenanz: That should leave support for `long double` if there is no need for endian swapping (the usual case).

dzenanz (approved):

N-Dekker:
Ready! We may of course reconsider/revisit `long double` support later, as well as a compile-time static_assert on the size or type of the byte-swapped argument. This is just a STYLE pull request anyway.

thewtex (approved):
Great!

N-Dekker:
> Great!

@thewtex Thanks, Matt! Please feel free to merge it!

PR #4856, opened 2024-09-20
Make the index type of VectorContainer `itk::SizeValueType` by default

N-Dekker:
For most use cases, it appears that the index type of `itk::VectorContainer` should just be `itk::SizeValueType` (or `itk::IdentifierType`, but that's just an alias of `itk::SizeValueType`).

This pull request allows users to write `itk::VectorContainer<ElementType>` as a shorthand for `itk::VectorContainer<itk::SizeValueType, ElementType>`.

### DIFF
--- a/Modules/Core/Common/include/itkVectorContainer.h
+++ b/Modules/Core/Common/include/itkVectorContainer.h
@@ -587,7 +587,7 @@ template <typename TElement>
 auto
 MakeVectorContainer(std::vector<TElement> stdVector)
 {
-  auto vectorContainer = VectorContainer<SizeValueType, TElement>::New();
+  auto vectorContainer = VectorContainer<TElement>::New();
   vectorContainer->CastToSTLContainer() = std::move(stdVector);
   return vectorContainer;
 }
--- a/Modules/Core/Common/include/itkVectorContainer.h
+++ b/Modules/Core/Common/include/itkVectorContainer.h
@@ -21,10 +21,13 @@
 #include "itkObject.h"
 #include "itkObjectFactory.h"
 
+#include <type_traits> // For is_void_v.
 #include <utility>
 #include <vector>
 
 namespace itk
+{
+namespace detail
 {
 /** \class VectorContainer
  *  \brief Define a front-end to the STL "vector" container that conforms to the
@@ -561,6 +564,22 @@ protected:
     , VectorType(first, last)
   {}
 };
+} // namespace detail
+
+
+/** Alias template, allowing to use `itk::VectorContainer<TElement>` without having to explicitly specify its
+ * `ElementIdentifier` type.
+ *
+ * The template parameters `T1` and `T2` allow specifying the index type and the element type, as follows:
+ *
+ * \tparam T1 The index type OR (when `T2` is `void`) the element type.
+ *
+ * \tparam T2 The element type OR `void`. When `T2` is `void`, the element type is specified by the first template
+ * argument (T1), and the index type will be `SizeValueType`.
+ */
+template <typename T1, typename T2 = void>
+using VectorContainer = detail::VectorContainer<std::conditional_t<std::is_void_v<T2>, SizeValueType, T1>,
+                                                std::conditional_t<std::is_void_v<T2>, T1, T2>>;
 
 
 /** Makes a VectorContainer that has a copy of the specified `std::vector`. */
--- a/Modules/Core/SpatialObjects/include/itkEllipseSpatialObject.h
+++ b/Modules/Core/SpatialObjects/include/itkEllipseSpatialObject.h
@@ -49,7 +49,7 @@ public:
   using typename Superclass::PointType;
   using typename Superclass::TransformType;
   using typename Superclass::BoundingBoxType;
-  using PointContainerType = VectorContainer<IdentifierType, PointType>;
+  using PointContainerType = VectorContainer<PointType>;
   using PointContainerPointer = SmartPointer<PointContainerType>;
 
   using ArrayType = FixedArray<double, TDimension>;
--- a/Modules/Core/SpatialObjects/include/itkMeshSpatialObject.h
+++ b/Modules/Core/SpatialObjects/include/itkMeshSpatialObject.h
@@ -53,7 +53,7 @@ public:
   using typename Superclass::PointType;
   using typename Superclass::BoundingBoxType;
 
-  using PointContainerType = VectorContainer<IdentifierType, PointType>;
+  using PointContainerType = VectorContainer<PointType>;
   using PointContainerPointer = typename PointContainerType::Pointer;
 
   /** Method for creation through the object factory. */
--- a/Modules/Core/SpatialObjects/include/itkTubeSpatialObject.h
+++ b/Modules/Core/SpatialObjects/include/itkTubeSpatialObject.h
@@ -57,7 +57,7 @@ public:
   using typename Superclass::PointType;
   using typename Superclass::TransformType;
   using typename Superclass::SpatialObjectPointType;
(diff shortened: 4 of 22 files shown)


Modules/Core/SpatialObjects/include/itkEllipseSpatialObject.h
```diff
@@ -100,13 +100,25 @@ class ITK_TEMPLATE_EXPORT EllipseSpatialObject : public SpatialObject<TDimension
...
 
 #if !defined(ITK_LEGACY_REMOVE)
-  itkLegacyMacro(void SetRadius(double radius)) { this->SetRadiusInObjectSpace(radius); }
-
-  itkLegacyMacro(void SetRadius(ArrayType radii)) { this->SetRadiusInObjectSpace(radii); }
-
-  itkLegacyMacro(ArrayType GetRadius() const) { return this->GetRadiusInObjectSpace(); }
-
-  itkLegacyMacro(void SetRadiiInObjectSpace(ArrayType radii)) { this->SetRadiusInObjectSpace(radii); }
+  itkLegacyMacro(void SetRadius(double radius))
```
dzenanz, line 103: This set of changes seems unrelated.

Modules/Core/SpatialObjects/include/itkMeshSpatialObject.h
```diff
@@ -96,7 +96,10 @@ class ITK_TEMPLATE_EXPORT MeshSpatialObject : public SpatialObject<TMesh::PointD
 
 #if !defined(ITK_LEGACY_REMOVE)
   /** \deprecated Return the type of pixel used */
-  itkLegacyMacro(const char * GetPixelTypeName()) { return m_PixelType.c_str(); }
+  itkLegacyMacro(const char * GetPixelTypeName())
```
dzenanz, line 99: Unrelated. Preferably group with the other into a separate commit.

Modules/Core/SpatialObjects/include/itkTubeSpatialObject.h
```diff
@@ -85,7 +85,10 @@ class ITK_TEMPLATE_EXPORT TubeSpatialObject : public PointBasedSpatialObject<TDi
 
 #if !defined(ITK_LEGACY_REMOVE)
   /** Compute the tangents and normals of the centerline of the tube. */
-  itkLegacyMacro(bool ComputeTangentAndNormals()) { return ComputeTangentsAndNormals(); }
+  itkLegacyMacro(bool ComputeTangentAndNormals())
```
dzenanz, line 88: Another unrelated change.

dzenanz (commented):
Mostly looks good.

Modules/Core/SpatialObjects/include/itkEllipseSpatialObject.h
N-Dekker: Thanks! Honestly I have no idea why clang-format decided to make those unrelated changes  🤷

Modules/Core/Common/include/itkVectorContainer.h
```diff
@@ -44,10 +48,10 @@ namespace itk
  * \sphinxexample{Core/Common/IterateOnAVectorContainer,Iterate On A Vector Container}
  * \endsphinx
  */
-template <typename TElementIdentifier, typename TElement>
+template <typename T1, typename T2 = void>
 class ITK_TEMPLATE_EXPORT VectorContainer
   : public Object
-  , private std::vector<TElement>
+  , private std::vector<std::conditional_t<std::is_void_v<T2>, T1, T2>>
```
N-Dekker, line 54: Unfortunately this "trick" does not work as smoothly as I was hoping for.... The compiler still considers `VectorContainer<ElementType>` and `VectorContainer<SizeValueType, ElementType>` to be different types. Which disallows interchanging one with the other. (For example, assigning a pointer to another.) Not really convenient 🤔

Modules/Core/SpatialObjects/include/itkEllipseSpatialObject.h
N-Dekker: - @dzenanz Related pull request, doing those unrelated changes: #4857

N-Dekker: - Resolved by pull request #4857

Modules/Core/SpatialObjects/include/itkTubeSpatialObject.h
N-Dekker: - Resolved (again) by pull request #4857

Modules/Core/SpatialObjects/include/itkMeshSpatialObject.h
N-Dekker: - Resolved (again) by pull request #4857

dzenanz (approved):

Modules/Core/Common/include/itkVectorContainer.h
```diff
@@ -21,11 +21,14 @@
...
 #include "itkObjectFactory.h"
 
+#include <type_traits> // For is_void_v.
 #include <utility>
 #include <vector>
 
 namespace itk
 {
+namespace ClassTemplate
+{
```
N-Dekker, line 31: This pull request places the original class template `VectorContainer` in a nested namespace, `itk::TemplateClass`, so that it will still be available to users who _really need_ to access the original class. Usually, users may just use `itk::VectorContainer` (as before), which will then be an alias to `itk::TemplateClass::VectorContainer`.

This approach is a bit of a workaround. Maybe we can _eventually_ just define ``itk::VectorContainer<TElement>` as a template class, without the `TElement` parameter. But that would be a breaking change. 🤷

Modules/Core/Common/include/itkVectorContainer.h
N-Dekker: Resolved by making `itk::VectorContainer` an _alias_ _template_ of `itk::ClassTemplate::VectorContainer`!

Modules/Core/Common/include/itkVectorContainer.h
```diff
@@ -561,14 +564,30 @@ class ITK_TEMPLATE_EXPORT VectorContainer
...
+ * The template parameters `T1` and `T2` allow specifying the index type and the element type, as follows:
+ *
+ * \tparam T1 The index type OR (when `T2` is `void`) the element type.
+ *
+ * \tparam T2 The element type OR `void`. When `T2` is `void`, the element type is specified by the first template
+ * argument (T1), and the index type will be `SizeValueType`.
+ */
+template <typename T1, typename T2 = void>
+using VectorContainer = ClassTemplate::VectorContainer<std::conditional_t<std::is_void_v<T2>, SizeValueType, T1>,
+                                                       std::conditional_t<std::is_void_v<T2>, T1, T2>>;
```
thewtex, line 582: Cool!

Modules/Core/Common/include/itkVectorContainer.h
thewtex: I think we should use `itk::detail::` here instead of creating a new namespace.

thewtex (changes requested):
@N-Dekker awesome!!

Naming comment inline.

Modules/Core/Common/include/itkVectorContainer.h
N-Dekker: Thanks for the suggestion @thewtex. Can you please explain a little bit? What is the general rule for _when_ to place something in `itk::detail::`?

dzenanz (approved):

thewtex (approved):
Excellent!! :sparkles:

PR #4860, opened 2024-09-24
ENH: Add "itkPointSetTest.py" unittest, testing `SetPointsByCoordinates`

N-Dekker:
Specifically tested `SetPointsByCoordinates` with a NumPy array as input argument, as was suggested by @PranjalSahu.

- Follow-up to pull request https://github.com/InsightSoftwareConsortium/ITK/pull/4850 commit f43b1b8977fa05d169c14b4c6f44ce35992a3959
ENH: Add `PointSet::SetPointsByCoordinates(coordinates)`

### DIFF
--- a/Modules/Core/Common/wrapping/test/itkPointSetTest.py
+++ b/Modules/Core/Common/wrapping/test/itkPointSetTest.py
@@ -1,0 +1,52 @@
+# ==========================================================================
+#
+#   Copyright NumFOCUS
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#          https://www.apache.org/licenses/LICENSE-2.0.txt
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ==========================================================================*/
+
+import itk
+import unittest
+import numpy as np
+
+class PointSetTestCase(unittest.TestCase):
+    """Tests itk.PointSet"""
+
+    def test_SetPointsByCoordinates_from_NumPy_array(self) -> None:
+        """Tests the SetPointsByCoordinates method, having a NumPy array as input"""
+
+        # The pixel type is irrelevant for this test, just use unsigned char.
+        PixelType = itk.UC
+
+        # Test the most common dimensions, and a few different numbers as number of points.
+        for dimension in [2, 3]:
+            for number_of_points in [1, 4, 9]:
+                # Create a NumPy array of random coordinates between -1.0 and 1.0
+                np_array = 2.0 * np.random.rand(number_of_points * dimension) - 1.0
+
+                point_set = itk.PointSet[PixelType, dimension].New()
+                point_set.SetPointsByCoordinates(np_array)
+
+                points = point_set.GetPoints()
+
+                self.assertEqual(points.Size(), number_of_points)
+
+                for i in range(number_of_points):
+                    point = points.GetElement(i)
+                    for j in range(dimension):
+                        self.assertAlmostEqual(point[j], np_array[i * dimension + j])
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/Modules/Core/Common/wrapping/test/CMakeLists.txt
+++ b/Modules/Core/Common/wrapping/test/CMakeLists.txt
@@ -49,6 +49,11 @@ if(ITK_WRAP_PYTHON)
     ${CMAKE_CURRENT_SOURCE_DIR}/itkPointSetSerializationTest.py)
   itk_python_add_test(
     NAME
+    itkPointSetTest
+    COMMAND
+    ${CMAKE_CURRENT_SOURCE_DIR}/itkPointSetTest.py)
+  itk_python_add_test(
+    NAME
     itkSpatialOrientationAdapterTest
     COMMAND
     ${CMAKE_CURRENT_SOURCE_DIR}/itkSpatialOrientationAdapterTest.py)


dzenanz:
/azp.run ITK.macOS.Arm64

dzenanz (approved):

N-Dekker:
> /azp run ITK.macOS.Arm64

@dzenanz FYI "ITK.macOS.Arm64"does not listen to `/azp run`, because it's not using Azure Pipelines!

N-Dekker:
Just checked, the test has actually run, and passed successfully: https://open.cdash.org/tests/1733705147

> Test: itkPointSetTest (Passed)

Command Line:

    /home/vsts/work/1/s-build/Wrapping/Generators/Python/itk/itkTestDriver "--add-before-env" "PYTHONPATH" "/home/vsts/work/1/s-build/Wrapping/Generators/Python" "--add-before-env" "PYTHONPATH" "/home/vsts/work/1/s-build/Wrapping/Generators/Python/itk" "--add-before-libpath" "/home/vsts/work/1/s-build/Wrapping/Generators/Python/itk" "/opt/hostedtoolcache/Python/3.9.20/x64/bin/python3" "/home/vsts/work/1/s/Modules/Core/Common/wrapping/test/itkPointSetTest.py"

👍

PranjalSahu:
Instead of Unsigned Char, can we instead use float ?

PranjalSahu:
We should also replace all the older tests using SetPoints with the new method to ensure everything else is working.

Modules/Core/Common/wrapping/test/itkPointSetTest.py
```diff
@@ -0,0 +1,52 @@
...
+import numpy as np
+
+class PointSetTestCase(unittest.TestCase):
+    """Tests itk.PointSet"""
+
+    def test_SetPointsByCoordinates_from_NumPy_array(self) -> None:
+        """Tests the SetPointsByCoordinates method, having a NumPy array as input"""
+
+        # The pixel type is irrelevant for this test, just use unsigned char.
+        PixelType = itk.UC
```
N-Dekker, line 30: > Instead of Unsigned Char, can we instead use float ?  Because that will be the most used datatype.

@PranjalSahu Thanks for the suggestion, but isn't UC a common pixel type as well? We could extend the test to try any possible pixel type, but it's irrelevant to the functionality of `SetPointsByCoordinates` anyway. `SetPointsByCoordinates` is just about the points of a PointSet. It's not about the pixels.

Note that this is _not_ about the coordinate type. The coordinate type is defined by the "MeshTraits" of a PointSet type. (By default, it is 32-bit float, independent of the pixel type.)

N-Dekker:
> We should also replace all the older tests using SetPoints with the new method to ensure everything else is working.

@PranjalSahu Please feel free to make another pull request to do so. (Otherwise I can also give it a try, once this pull request is merged.)

Modules/Core/Common/wrapping/test/itkPointSetTest.py
N-Dekker: For the record, by default, `itk.PointSet` supports double, float, short, unsigned char, unsigned int, itk::Array, itk::Matrix, itk::Point, and itk::Vector as pixel type, according to `itk.PointSet.GetTypes()`:

```python
>>> print(itk.PointSet.GetTypes())
<itkTemplate itk::PointSet>
Options:
  [<itkCType double>, 2]
  [<itkCType double>, 3]
  [<itkCType float>, 2]
  [<itkCType float>, 3]
  [<itkCType signed short>, 2]
  [<itkCType signed short>, 3]
  [<itkCType unsigned char>, 2]
  [<itkCType unsigned char>, 3]
  [<itkCType unsigned int>, 2, <class 'itk.itkDefaultStaticMeshTraitsPython.itkDefaultStaticMeshTraitsUI22FFUI'>]
  [<itkCType unsigned int>, 3, <class 'itk.itkDefaultStaticMeshTraitsPython.itkDefaultStaticMeshTraitsUI33FFUI'>]
  [<class 'itk.itkArrayPython.itkArrayD'>, 2]
  [<class 'itk.itkArrayPython.itkArrayD'>, 3]
  [<class 'itk.itkMatrixPython.itkMatrixD22'>, 2]
  [<class 'itk.itkMatrixPython.itkMatrixD33'>, 3]
  [<class 'itk.itkPointPython.itkPointD2'>, 2, <class 'itk.itkDefaultStaticMeshTraitsPython.itkDefaultStaticMeshTraitsD22DDD'>]
  [<class 'itk.itkPointPython.itkPointD3'>, 3, <class 'itk.itkDefaultStaticMeshTraitsPython.itkDefaultStaticMeshTraitsD33DDD'>]
  [<class 'itk.itkVectorPython.itkVectorF3'>, 3]
None

```

PranjalSahu (approved):
LGTM

Modules/Core/Common/wrapping/test/itkPointSetTest.py
```diff
@@ -0,0 +1,52 @@
...
+                point_set.SetPointsByCoordinates(np_array)
+
+                points = point_set.GetPoints()
+
+                self.assertEqual(points.Size(), number_of_points)
+
+                for i in range(number_of_points):
+                    point = points.GetElement(i)
+                    for j in range(dimension):
+                        self.assertAlmostEqual(point[j], np_array[i * dimension + j])
```
thewtex, line 48: It is better to use [`numpy` `assert_allclose`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose), which is must faster and has a shape check.

thewtex (changes requested):
@N-Dekker thanks!

Request inline.

Modules/Core/Common/wrapping/test/itkPointSetTest.py
N-Dekker: Thanks for the suggestion @thewtex Can you possibly type out exactly how you want it to be?

N-Dekker: @thewtex Also, do you think the test should be faster than it is right now?

dzenanz: Maybe `np.assert_allclose(points, np_array.T)`? Does that compile and work?

N-Dekker: @dzenanz Thanks for the suggestion, but no, unfortunately:
```python
np.assert_allclose(points, np_array.T)
```
> module 'numpy' has no attribute 'assert_allclose'
----

```python
np.testing.assert_allclose(points, np_array)
```
> TypeError: unsupported operand type(s) for -: 'itkVectorContainerULLPF2' and 'float'
----

```python
np.testing.assert_allclose(point_set, np_array)
```
> TypeError: unsupported operand type(s) for -: 'itkPointSetUC2' and 'float'

dzenanz: Maybe not for the whole array, but for a single point? And iterate over points.

N-Dekker: > Maybe not for the whole array, but for a single point?

Maybe... but I don't know how. So it would be helpful if Matt (@thewtex) could tell us, exactly. Otherwise we just keep guessing 🤷

thewtex: @N-Dekker 

```python
from numpy.testing import assert_allclose

pointset_dict = itk.dict_from_pointset(point_set)
assert_allclose(pointset_dict['points'], np_array))
```

N-Dekker: Thanks @thewtex, but sorry, that does not work for me either. See below. Did you locally try this particular unit test with your suggested adjustment? Would it be acceptable to you to just merge the pull request as it is right now, and then afterwards possibly make a new PR to further optimize the test, for example by using `numpy.testing.assert_allclose`?

```
ERROR: test_SetPointsByCoordinates_from_NumPy_array (__main__.PointSetTestCase.test_SetPointsByCoordinates_from_NumPy_array)
Tests the SetPointsByCoordinates method, having a NumPy array as input
----------------------------------------------------------------------
  File "itkPointSetTest.py", line 46, in test_SetPointsByCoordinates_from_NumPy_array
    pointset_dict = itk.dict_from_pointset(point_set)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\cedekker\.conda\envs\ITKSwigPy311\Lib\site-packages\itk\support\extras.py", line 957, in dict_from_pointset
    pixel_type, mangle, pixel_type_components = wasm_type_from_pointset_type(pointset)
                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\cedekker\.conda\envs\ITKSwigPy311\Lib\site-packages\itk\support\helpers.py", line 393, in wasm_type_from_pointset_type
    return pixelType, python_to_js(mangle), pixel_type_components
                      ^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\cedekker\.conda\envs\ITKSwigPy311\Lib\site-packages\itk\support\helpers.py", line 212, in python_to_js
    return _python_to_js[name]
           ~~~~~~~~~~~~~^^^^^^
KeyError: None
```

thewtex: > Thanks @thewtex, but sorry, that does not work for me either. See below. Did you locally try this particular unit test with your suggested adjustment? 

No

> Would it be acceptable to you to just merge the pull request as it is right now, and then afterwards possibly make a new PR to further optimize the test, for example by using numpy.testing.assert_allclose?

No, this is an important performance and usage fix, and it is also important to verify that the proposed change is correct in terms of `shape`.

I will investigate locally.

N-Dekker: > it is also important to verify that the proposed change is correct in terms of `shape`.

@thewtex Regarding the `shape`, please note that it is important that the argument of `SetPointsByCoordinates` is a "flat" (1-D) array of coordinate values, when doing

```python
point_set.SetPointsByCoordinates(np_array)
```

The `SetPointsByCoordinates` member function is meant to be an alternative to the existing unsafe `SetPoints` overload  (issue #4848), which also expects a "flat" array of coordinate values as argument.

So the PointSet and the array of coordinates (`np_array`) have different shapes, but that is _intentionally_. So I'm sorry, I don't think `numpy.testing.assert_allclose` can be used for this unit test. 🤷

dzenanz: It might, with reshaping. Reshaping in numpy is easy.

N-Dekker: I'm sorry, but this is just a single unit test. It is not meant to be a tutorial on how to reshape a numpy array.

thewtex: > The SetPointsByCoordinates member function is meant to be an alternative to the existing unsafe SetPoints overload (issue #4848), which also expects a "flat" array of coordinate values as argument.

> So the PointSet and the array of coordinates (np_array) have different shapes, but that is intentionally.

The intension and/or the related support functions may need improvements :-). NumPy arrays representing points should not be flat -- they should have a dimension that corresponds to the spatial dimension.

N-Dekker:
- While the merged pull request #4896 already extended the existing serialization tests in Python to use  `SetPointsByCoordinates`, I still think it would be useful to specifically test `SetPointsByCoordinates` in isolation, as I'm proposing with this unit test.

thewtex:
@N-Dekker Pull requests do not seem to be available for your branch, but a few improvements are added to your branch here:

https://github.com/thewtex/ITK/tree/itkPointSetTest-Python-unittest

which adds support for passing a *n_points x dim* to `SetPointsByCoordinates` (which is what I think people expect from a Python interface). It also adds a `GetPointsByCoordinates` that returns the same. And verifies with `numpy.testing.assert_allclose`.

N-Dekker:
> https://github.com/thewtex/ITK/tree/itkPointSetTest-Python-unittest
> 
> which adds support for passing a _n_points x dim_ to `SetPointsByCoordinates` (which is what I think people expect from a Python interface). It also adds a `GetPointsByCoordinates` that returns the same. And verifies with `numpy.testing.assert_allclose`.

Thanks Matt! I think it's very nice to add support for 2D (number of points x dimension) numpy arrays. The approach of using `%rename` and `%extend swig_name` is new to me, cool!

I'm glad to see your proposal still preserves the original support of flat arrays of coordinates as was added by PR #4850 and included with ITK v6.0a01. (I think that was an essential feature, in order to provide an alternative to the problematical `SetPoints` overload.) So your proposal looks like an enhancement on top of  PR #4850 and #4860 (this one), right? As such, I would rather see this one (PR #4860) being merged, so that we can then review your proposal as a follow-up PR. Would that be OK to you?

thewtex:
:+1: yes, @N-Dekker sounds like a good plan.

thewtex (approved):

PR #4862, opened 2024-09-26
Deprecate `ByteSwapper::OStreamType`, fix "SwapWriteRange" member functions

N-Dekker:
The type alias `OStreamType` just served as a workaround to an old MSVC bug. This pull request deprecates (`ITK_FUTURE_LEGACY_REMOVE`) the type alias.

This pull request also has a fix for the behavior of `SwapWriteRangeFromSystemToBigEndian` and `SwapWriteRangeFromSystemToLittleEndian`, for when `sizeof(T) == 1`. The original code did nothing, in that case. With this pull request, those member functions will in that case just write the specified range to the output stream.

### DIFF
--- a/Modules/Core/Common/include/itkByteSwapper.hxx
+++ b/Modules/Core/Common/include/itkByteSwapper.hxx
@@ -72,20 +72,7 @@ ByteSwapper<T>::SwapWriteRangeFromSystemToBigEndian(const T * p, int num, std::o
   }
   else
   {
-    switch (sizeof(T))
-    {
-      case 2:
-        Self::SwapWrite2Range(p, num, fp);
-        return;
-      case 4:
-        Self::SwapWrite4Range(p, num, fp);
-        return;
-      case 8:
-        Self::SwapWrite8Range(p, num, fp);
-        return;
-      default:
-        itkGenericExceptionMacro("Cannot swap number of bytes requested");
-    }
+    SwapWriteRange(p, num, *fp);
   }
 }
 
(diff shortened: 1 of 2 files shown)


dzenanz (approved):

Modules/Core/Common/include/itkByteSwapper.hxx
```diff
@@ -63,31 +63,16 @@ ByteSwapper<T>::SwapRangeFromSystemToBigEndian([[maybe_unused]] T * p, [[maybe_u
...
-      case 4:
-        Self::SwapWrite4Range(p, num, fp);
-        return;
-      case 8:
-        Self::SwapWrite8Range(p, num, fp);
-        return;
-      default:
-        itkGenericExceptionMacro("Cannot swap number of bytes requested");
-    }
+    SwapWriteRange(p, num, *fp);
```
N-Dekker, line 75: - Just like pull request #4855, this code change is again inspired by the suggestion from Bradley (@blowekamp) at https://github.com/InsightSoftwareConsortium/ITK/pull/4841/files#r1751777344

Modules/Core/Common/include/itkByteSwapper.hxx
```diff
@@ -337,6 +307,27 @@ ByteSwapper<T>::SwapBytes(T & value)
...
+    std::copy_n(buffer, chunkSize, chunk.get());
+    std::for_each_n(chunk.get(), chunkSize, [](T & element) { SwapBytes(element); });
+
+    outputStream.write(reinterpret_cast<const char *>(chunk.get()),
+                       static_cast<std::streamsize>(chunkSize * sizeof(T)));
+    buffer += chunkSize;
+    numberOfElements -= chunkSize;
+    chunkSize = std::min(numberOfElements, chunkSize);
+  }
+}
```
N-Dekker, line 329: For comparison, down here is the existing `ByteSwapper<T>::SwapWrite8Range`, from the current master revision. (The existing `SwapWrite8Range` is very similar to `SwapWrite2Range` and `SwapWrite4Range` as well).
https://github.com/InsightSoftwareConsortium/ITK/blob/3b04ad1b377c6d5d77a2c1f454c363eab8197340/Modules/Core/Common/include/itkByteSwapper.hxx#L289-L312

dzenanz (approved):

blowekamp (approved):
The changes look OK. I do have a concern that we don't have a way to test the code.

N-Dekker:
Thanks to you both for your approval, @dzenanz, @blowekamp 

> I do have a concern that we don't have a way to test the code.

Well, I assume that this code is tested _indirectly_, because `SwapWriteRangeFromSystemToBigEndian` and `SwapWriteRangeFromSystemToLittleEndian` are both being called by various IO modules, which are being tested.

Does that take away your concern? Otherwise press _Request changes_ 😺

PR #4866, opened 2024-10-01
COMP: Fix compilation issue when using RelabelComponentImageFilter

### LINKED ISSUE #4865: Compilation error while using itk::RelabelComponentImageFilte::GetSizeOfObjectInPhysicalUnits
### Description

Using the `itkRelabelComponentImageFilter` filter `GetSizeOfObjectInPhysicalUnits` with signed integer image causes a compilation error due comparison of integer expressions of different signedness.

### Steps to Reproduce

1. Install & build ITK (master)
2. Build [test snippet](https://gist.github.com/kenavolic/acd6a544e58a995f395812c6e404b9de)

### Expected behavior

Should be able to use the class method

### Actual behavior

Compilation error:
```shell
In file included from /home/anon/Dev/test_itk/main.cpp:2:
/home/anon/Dev/ITK/Modules/Segmentation/ConnectedComponents/include/itkRelabelComponentImageFilter.h: In instantiation of ‘float itk::RelabelComponentImageFilter<TInputImage, TOutputImage>::GetSizeOfObjectInPhysicalUnits(itk::RelabelComponentImageFilter<TInputImage, TOutputImage>::LabelType) const [with TInputImage = itk::Image<short int, 3>; TOutputImage = itk::Image<short int, 3>; itk::RelabelComponentImageFilter<TInputImage, TOutputImage>::LabelType = short int]’:
/home/anon/Dev/test_itk/main.cpp:13:46:   required from here
/home/anon/Dev/ITK/Modules/Segmentation/ConnectedComponents/include/itkRelabelComponentImageFilter.h:228:24: error: comparison of integer expressions of different signedness: ‘itk::RelabelComponentImageFilter<itk::Image<short int, 3>, itk::Image<short int, 3> >::LabelType’ {aka ‘short int’} and ‘const SizeValueType’ {aka ‘const long unsigned int’} [-Werror=sign-compare]
  228 |     if (obj > 0 && obj <= m_NumberOfObjects)
```

### Reproducibility

Always (under the same compilation flags)

### Versions

Current developpement version and older ones (detected on ITK5.2)

### Environment

* Ubuntu 18.04
* gcc 10.3
* cmake 3.25

### Additional Information

[Link to discourse discussion](https://discourse.itk.org/t/itk-getsizeofobjectinphysicalunits-signature/7237)

kenavolic:
* Update parameter type to reflect underlying container indexing

Fixes #4865

## PR Checklist

There is a public API change as the parameter type is changed.

### DIFF
--- a/Modules/Segmentation/ConnectedComponents/include/itkRelabelComponentImageFilter.h
+++ b/Modules/Segmentation/ConnectedComponents/include/itkRelabelComponentImageFilter.h
@@ -209,7 +209,7 @@ public:
   ObjectSizeType
   GetSizeOfObjectInPixels(LabelType obj) const
   {
-    if (obj > 0 && obj <= m_NumberOfObjects)
+    if (obj > 0 && static_cast<SizeValueType>(obj) <= m_NumberOfObjects)
     {
       return m_SizeOfObjectsInPixels[obj - 1];
     }
@@ -225,7 +225,7 @@ public:
   float
   GetSizeOfObjectInPhysicalUnits(LabelType obj) const
   {
-    if (obj > 0 && obj <= m_NumberOfObjects)
+    if (obj > 0 && static_cast<SizeValueType>(obj) <= m_NumberOfObjects)
     {
       return m_SizeOfObjectsInPhysicalUnits[obj - 1];
     }


dzenanz (approved):

Modules/Segmentation/ConnectedComponents/include/itkRelabelComponentImageFilter.h
```diff
@@ -207,7 +207,7 @@ class ITK_TEMPLATE_EXPORT RelabelComponentImageFilter : public InPlaceImageFilte
    * valid after the filter has executed.  Size of the background
    * (object #0) is not calculated.  */
   ObjectSizeType
-  GetSizeOfObjectInPixels(LabelType obj) const
+  GetSizeOfObjectInPixels(SizeValueType obj) const
   {
     if (obj > 0 && obj <= m_NumberOfObjects)
```
N-Dekker, line 212: Another solution might be to cast `obj` to `SizeValueType`, but then only when `obj` is greater than zero. In code:
```cpp
    GetSizeOfObjectInPixels(LabelType obj) const
    {
      if (obj > 0 && static_cast<SizeValueType>(obj) <= m_NumberOfObjects)
```

Doing so would avoid integer overflow when the passed argument is in fact a negative number. (I do not know if that is a concern to anyone. Just something to consider. Would people ever call something like `GetSizeOfObjectInPixels(-1)` anyway?)

kenavolic: I guess this would work and has the advantage to not break the public API.  From a strict language point of view, it is a bit confusing to enable `LabelType` parameter when accessing an array-like container by index. I'll consider this solution as a patch in order to keep the public API stable.

N-Dekker: > If it makes no sense to use the class with signed pixel type, it should be enforced at compile-time

I agree. But for your use case, it _does_ make sense to have a signed pixel type, right?

kenavolic: > > If it makes no sense to use the class with signed pixel type, it should be enforced at compile-time
> 
> I agree. But for your use case, it _does_ make sense to have a signed pixel type, right?

Yes

N-Dekker: Does your image also have negative pixel values? And if so, how does (or how _should_) the filter deal with them?

N-Dekker: For the record, the parameter type of `GetSizeOfObjectInPixels` was actually changed before, from `unsigned long` to `LabelType`, with commit 1bd69fc364bcced99bf48a3739e5acd3719b6e04 Luis Ibanez (@luisibanez), April 28, 2009: _COMP: Fixing warnings about casting._

kenavolic: > Does your image also have negative pixel values? And if so, how does (or how _should_) the filter deal with them?

Sorry for my poor explanations. We use legacy code that use signed short images in a processing pipeline where images have only positive value...

N-Dekker: No problem! So it still appears useful to allow signed pixel types.

The compile error you reported in issue #4865 is only an _error_ because you compiled with GCC option `-Werror`, right? Otherwise I guess it would just be a warning. Anyway, we also try to avoid any warning 😇 

I would like it if you would amend (+ force-push) your commit to use `static_cast<SizeValueType>`, and restore the original API. But I leave it up to you. 👍

blowekamp:
So SimpleITK has this filter instantiated with signed pixel types:

https://github.com/SimpleITK/SimpleITK/blob/master/Code/BasicFilters/json/RelabelComponentImageFilter.json#L7

But only returns std::vectors for these attributes and does not expose  any `get(index)` methods.

If I recall correctly the original Watershed filter relied on signed pixel types and this filter.

Modules/Segmentation/ConnectedComponents/include/itkRelabelComponentImageFilter.h
kenavolic: Done

dzenanz (approved):

N-Dekker (approved):
Thanks @kenavolic ! 👍

N-Dekker:
For the record, the issue was originally discussed at https://discourse.itk.org/t/itk-getsizeofobjectinphysicalunits-signature/7237

PR #4867, opened 2024-10-02
ENH: Add `itk::PointSetBase` as abstract base class of `itk::PointSet`

N-Dekker:
Allows an algorithm to process the points of a `PointSet` object, without having to know its `PixelType` at compile-time. The algorithm may then just use a pointer or reference to its abstract base class, `PointSetBase`. Doing so may reduce the "code bloat" caused by template instantiations, and ease wrapping the algorithm for Python.

With this commit, its `PixelType` and its `PointData` container are still defined inside `itk::PointSet`, while the other properties (its points and its region information) are moved to its base class, `PointSetBase`.


Modules/Core/Common/wrapping/itkPointSet.wrap
```diff
@@ -28,3 +28,12 @@ if(has_d_3)
   itk_wrap_template("VF33" "itk::Vector< float, 3 >, 3")
 endif()
 itk_end_wrap_class()
+
+itk_wrap_class("itk::PointSetBase" POINTER)
```
dzenanz, line 32: Shouldn't this be at the top of this file, before `PointSet`?

N-Dekker: Thanks, @dzenanz I just moved the `itk_wrap_class` statement upward, as you suggested, and placed another one in "itkQuadEdgeMeshBase.wrap", please check: [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/353ef12ef7e736b20b85191ab0c73c0fb283a46d..c0a21a187284906e68af346ad076ebb7945e1a07)

dzenanz (approved):

dzenanz:
> still preferable to wrap the base class before the derived class

Yes.

N-Dekker:
FYI, my intention is to pave the way for adding functions like `SetFixedPointSet(const PointSetBaseType * pointSet)` and `SetMovingPointSet(const PointSetBaseType * pointSet)` to the `itk::ElastixRegistrationMethod` of [elastix](https://github.com/SuperElastix/elastix).  These functions might then support `PointSet` objects of any `PixelType` or `PointData`.

Modules/Core/Common/wrapping/itkPointSet.wrap
N-Dekker: For the record, the answer is here: https://github.com/InsightSoftwareConsortium/ITK/pull/4867#issuecomment-2393606971 😺 

Which resolves the conversation, I guess  😇

dzenanz:
I was wondering what was the main purpose of this refactoring. 😄

thewtex (approved):
Awesome! :rocket:

Modules/Core/QuadEdgeMesh/wrapping/itkQuadEdgeMeshBase.wrap
```diff
@@ -19,6 +19,12 @@ foreach(d ${ITK_WRAP_IMAGE_DIMS})
 endforeach()
 itk_end_wrap_class()
 
+itk_wrap_class("itk::PointSetBase" POINTER)
+foreach(d ${ITK_WRAP_IMAGE_DIMS})
+  itk_wrap_template("MC${ITKM_IT}QEMPF${d}" "itk::MapContainer< ${ITKT_IT}, itk::QuadEdgeMeshPoint< float, ${d} > >")
+endforeach()
+itk_end_wrap_class()
+
```
N-Dekker, line 27: Even though this PR successfully passed the CI, it now looks like it made ITK.Windows.Python unhappy! It says at https://open.cdash.org/viewBuildError.php?type=1&buildid=9948358

```
itkQuadEdgeMeshBase: warning(4): ITK type not wrapped, or currently not known: itk::PointSetBase< itk::MapContainer< unsigned long long, itk::QuadEdgeMeshPoint< float, 2, itk::GeometricalQuadEdge< unsigned long long, unsigned long long, bool, bool > > > >
itkQuadEdgeMeshBase: warning(4): ITK type not wrapped, or currently not known: itk::MapContainer< unsigned long long, itk::QuadEdgeMeshPoint< float, 2 > >
itkQuadEdgeMeshBase: warning(4): ITK type not wrapped, or currently not known: itk::QuadEdgeMeshPoint< float, 2 >
```

Any idea?

dzenanz: Isn't `ITKT_IT` specifier for `IdentifierType`, which should in turn be `unsigned long long` on MSVC?

dzenanz: After a quick look, nothing jumps out as wrong. Best way to debug this is to set up a local Python build, and then iteratively tweak the wrap file and build.

N-Dekker: > Isn't `ITKT_IT` specifier for `IdentifierType`, which should in turn be `unsigned long long` on MSVC?

Thanks @dzenanz, looks like you're right 👍  I still wonder why `itk::IdentifierType` isn't just spelled out in the wrap files. "${ITKT_IT}" isn't very readable, is it?

dzenanz: All of those `ITKT_IT`-style specifiers are abbreviated. The list is here: https://itk.org/ITKSoftwareGuide/html/Book1/ITKSoftwareGuide-Book1ch9.html#x55-1520009.5

See also a PR which fixes `ITKT_IT`: https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/pull/215

N-Dekker: Thanks @dzenanz I still find it hard to grasp all those `${ITKA_B}`...`${ITKY_Z}` abbreviations, but maybe it's just me  🤷

dzenanz: I guess that `PointSetBase` should be wrapped with equivalent parameters to `QuadEdgeMeshTraits`. I guess it is not so right now?

N-Dekker: For every `PointSet` wrapping, there should be a `PointSetBase` wrapping. (But `PointSetBase` does not need as many wrappings as `PointSet`, because `PointSetBase` does not need to have a separate wrapping for each PixelType.)

N-Dekker: Should we also in this case use `${ITKT_F}` instead of `float`? But then, is `${ITKT_F}` anything more than just an "abbreviation" for `float`?

thewtex: @N-Dekker thanks for the follow-up.

Yes, this is likely related to the difference between the definition of `itk::IdentifierType` on Windows and other platforms. On Windows it resolves to `unsigned long long` to ensure 64-bit, on Unix `unsigned long`, which is already always 64-bit.

PR #4869, opened 2024-10-04
Remove unnecessary PyBuffer_Release calls, and other "Bridge/NumPy"  style improvements

N-Dekker:
According to https://docs.python.org/3.12/c-api/buffer.html#c.PyObject_GetBuffer, `PyBuffer_Release` should only be called after a _successful_ `PyObject_GetBuffer` call.

Also fixed a few MSVC 2022 warnings:
> warning C4189: 'res': local variable is initialized but not referenced
> warning C4389: '!=': signed/unsigned mismatch


dzenanz (approved):

thewtex (approved):
:+1:

PR #4863, opened 2024-09-27
Replace `memset` calls with `{}`, and `NULL` with `nullptr`, in "Bridge/NumPy"

N-Dekker:
Small style improvements, modernizing the code of "Bridge/NumPy"

PR #4870, opened 2024-10-06
Remove zero-filling memset calls and replace them with `()` or `{}`

N-Dekker:
- Replaced memset calls with `{}` assignment or initialization
- Replaced memset calls after `new T` with just `()` (doing `new T()` instead)
- Removed some `memset` calls that had no effect at all
---
- Follow-up to pull request #4863 commit aaff35cc345d3a587cfb3626c8cb93d72e2b7640

### DIFF
--- a/Modules/IO/VTK/src/itkVTKImageIO.cxx
+++ b/Modules/IO/VTK/src/itkVTKImageIO.cxx
@@ -858,9 +858,7 @@ VTKImageIO::WriteSymmetricTensorBufferAsBinary(std::ostream &                 os
   std::streamsize bytesRemaining = num;
   const SizeType  componentSize = this->GetComponentSize();
   SizeType        pixelSize;
-  char            zero[1024];
-
-  memset(zero, 0, 1024);
+  constexpr char  zero[1024]{};
 
   switch (this->GetNumberOfComponents())
   {
--- a/Modules/Core/Common/include/itkOffset.h
+++ b/Modules/Core/Common/include/itkOffset.h
@@ -447,9 +447,7 @@ template <unsigned int VDimension>
 Offset<VDimension>
 Offset<VDimension>::GetBasisOffset(unsigned int dim)
 {
-  Self ind;
-
-  memset(ind.m_InternalArray, 0, sizeof(OffsetValueType) * VDimension);
+  Self ind{};
   ind.m_InternalArray[dim] = 1;
   return ind;
 }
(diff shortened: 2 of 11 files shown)


blowekamp:
It doesn't seem that long ago  to me ( LOL, it was)  that I did a cross compiler performance test to demonstrate that memset was significantly faster than these default initialization. Do you have a link to recent performance comparisons across platforms and compilers?

Modules/IO/VTK/src/itkVTKImageIO.cxx
```diff
@@ -858,9 +858,7 @@ VTKImageIO::WriteSymmetricTensorBufferAsBinary(std::ostream &                 os
   std::streamsize bytesRemaining = num;
   const SizeType  componentSize = this->GetComponentSize();
   SizeType        pixelSize;
-  char            zero[1024];
-
-  memset(zero, 0, 1024);
+  constexpr char  zero[1024]{};
```
N-Dekker, line 861: @blowekamp Thanks for sharing your considerations regarding the possible performance effect of this commit. Why do you think `T var{}` might be slower than `T var; memset(&var, 0, sizeof(T))`?

In this particular case,  `constexpr char zero[1024]{}` is probably _faster_ than memset, because the `{}` initialization may be performed at compile-time. (We could even _enforce_ compile-time evaluation by making the variable `static constexpr`, but I doubt if that's worth it, in this particular case.)

blowekamp: As I recall my previous performance experiments were likely on large image buffers where a significant difference was measured. As I have seen a few of these changes preferring the initialization, I was querying in general how does the performance compare.

Quickly scanning these cases, it does not appear to be many large buffers in performance critical code. I don't think the are are carefully optimized cases.

The using the default initialize is cleaner, and more standardize code. In general, I'm in favor of it.

N-Dekker:
> It doesn't seem that long ago to me ( LOL, it was) that I did a cross compiler performance test to demonstrate that memset was significantly faster than these default initialization. Do you have a link to recent performance comparisons across platforms and compilers?

Thanks for asking, @blowekamp  I just tried:

```cpp
struct Pair
{
    int first;
    double second;
};

void ClearPair(Pair& arg)
{
    arg = {};
}

```

It produced the following ASM, at https://godbolt.org/z/vTKdTGb1x

```asm
ClearPair(Pair&):
        pxor    xmm0, xmm0
        movups  XMMWORD PTR [rdi], xmm0
        ret
```

The I replaced `arg = {}` with `memset(&arg, 0, sizeof(Pair))`. It produced exactly the same ASM: https://godbolt.org/z/s9788ex7G  😃

blowekamp:
There are a lot of different compilers. I see difference with some MSVC outputs.

Modules/IO/VTK/src/itkVTKImageIO.cxx
blowekamp: Thank you for the discussion and deep dive.

blowekamp (approved):

dzenanz (approved):

PR #4872, opened 2024-10-08
STYLE: Remove `const` from `const std::string` return types

N-Dekker:
Removing this `const` enables move semantics, which may improve the performance.

Following C++ Core Guidelines, Oct 3, 2024, Don’t return `const T`: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-return-const

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkSpatialObject.h
+++ b/Modules/Core/SpatialObjects/include/itkSpatialObject.h
@@ -147,7 +147,7 @@ public:
   itkSetMacro(TypeName, std::string);
 
   /** Get the typename of the SpatialObject */
-  virtual const std::string
+  virtual std::string
   GetTypeName() const
   {
     return m_TypeName;
--- a/Modules/IO/SpatialObjects/src/itkPolygonGroupSpatialObjectXMLFile.cxx
+++ b/Modules/IO/SpatialObjects/src/itkPolygonGroupSpatialObjectXMLFile.cxx
@@ -162,7 +162,7 @@ template <>
 class NumericTraits<std::string>
 {
 public:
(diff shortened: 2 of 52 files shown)


dzenanz (approved):
Does this affect wrapping?

N-Dekker:
> Does this affect wrapping?

I would not think so. SWIG certainly supports non-const `std::string` as return type. As in https://www.swig.org/Doc2.0/Library.html#Library_std_string :
```

%module example
%include "std_string.i"

std::string foo();
```

hjmjohnson (approved):

N-Dekker:
Off-topic, but please notice that this pull request is successfully passing the CI (after a few force-pushes), even though its PR branch includes the `itk::PointSetBase` commit 4b92507c6de3555ab201ddd9ff2758e713ca5bc4, which has broken the CI at the ITK master branch! As discussed at https://github.com/InsightSoftwareConsortium/ITK/pull/4867#discussion_r1790748146

Modules/Core/SpatialObjects/include/itkSpatialObject.h
```diff
@@ -147,7 +147,7 @@ class ITK_TEMPLATE_EXPORT SpatialObject : public DataObject
   itkSetMacro(TypeName, std::string);
 
   /** Get the typename of the SpatialObject */
-  virtual const std::string
+  virtual std::string
```
N-Dekker, line 150: I just noticed that _theoretically_ this might break legacy user code! But only if users would have overridden this virtual member function, in their legacy application.  Would some users really have overridden `SpatialObject::GetTypeName()` ?

PR #4883, opened 2024-10-17
Remove some local variables from PyBuffer, declare others const, remove buffer pointer casts

N-Dekker:
A few more coding style improvements of `itk::PyBuffer`.

### DIFF
--- a/Modules/Bridge/NumPy/include/itkPyBuffer.hxx
+++ b/Modules/Bridge/NumPy/include/itkPyBuffer.hxx
@@ -20,7 +20,8 @@
 
 
 #include "itkImportImageContainer.h"
-#include <memory> // For unique_ptr.
+#include <algorithm> // For reverse.
+#include <memory>    // For unique_ptr.
 
 namespace itk
 {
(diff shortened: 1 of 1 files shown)


Modules/Bridge/NumPy/include/itkPyBuffer.hxx
```diff
@@ -136,7 +125,7 @@ PyBuffer<TImage>::_GetImageViewFromArray(PyObject * arr, PyObject * shape, PyObj
   importer->SetImportPointer(data, numberOfPixels, importImageFilterWillOwnTheBuffer);
 
   OutputImagePointer output = TImage::New();
-  output->SetRegions(region);
+  output->SetRegions(size);
```
dzenanz, line 111: You seem to be mixing up `region` and `size`.

N-Dekker: Me? `itk::Image` _does_ support specifying its regions by their `itk::Size`! 😃 
https://github.com/InsightSoftwareConsortium/ITK/blob/a434e3f71829f23cd748a1694bc64455aecceabd/Modules/Core/Common/include/itkImageBase.h#L340 

If the index position of a region is just (0, 0), only its size matters.

dzenanz (approved):

PR #4887, opened 2024-10-19
Replace `Fill(T{})` with `{}` initialization, replace `T var; var.Fill` with `auto var = T::Filled` in tests

N-Dekker:
Replaced code of the form

    Type var;
    var.Fill(ElementType{});

with `Type var{};`

And specifically when having a local index or size variable, replaced code of the form

    T var;
    var.Fill(x);

with `auto var = T::Filled(x);`

Following C++ Core Guidelines, Oct 3, 2024, "Always initialize an object",
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-always

- Follow-up to pull request #4881

### DIFF
--- a/Modules/Bridge/VtkGlue/test/itkVtkMedianFilterTest.cxx
+++ b/Modules/Bridge/VtkGlue/test/itkVtkMedianFilterTest.cxx
@@ -50,9 +50,8 @@ itkVtkMedianFilterTest(int argc, char * argv[])
   reader->SetFileName(inputFilename);
 
   // Create and setup a median filter
-  auto                      medianFilter = FilterType::New();
-  FilterType::InputSizeType radius;
-  radius.Fill(2);
+  auto medianFilter = FilterType::New();
+  auto radius = FilterType::InputSizeType::Filled(2);
   if (argc > 2)
   {
     radius.Fill(std::stoi(argv[2]));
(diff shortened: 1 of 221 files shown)


N-Dekker:
@jhlegarreta I hope you like this PR! The adjustments might potentially prevent code analysis warnings about uninitialized variables. I added you as reviewer as you are very much involved with the tests of ITK.

jhlegarreta (approved):
Looks good. Would be good to apply all of these to the ITK SW Guide (where appropriate, or the `Fill` method is not explicitly demonstrated), the ITK Sphinx Examples, and to the remotes.

N-Dekker:
Thanks @jhlegarreta I agree that it should then also be applied to the ITK SW Guide.

In general, I'm proposing for variable initializations:

 1. To zero-fill a variable, use `{}` 
 2. To fill a variable with arbitrary (non-zero) fill-value
  a. do  `auto var = T::Filled(fillValue)` when the type is `itk::Size<N>` or `itk::Index<N>`
  b. do `auto var = itk::MakeFilled<T>(fillValue)` otherwise

Would that also be OK to you?

jhlegarreta:
> Would that also be OK to you?

Go for it if you have the bandwidth. Please, add the necessary explanation to the SWG and the necessary ITK Sphinx examples.

N-Dekker:
> Please, add the necessary explanation to the SWG and the necessary ITK Sphinx examples.

Thanks for suggestion Jon. 👍  My intention is to first update the ITK source tree (ITK/Modules/itk*.*) accordingly, and then adjust the SWG and examples. But I'm not yet finished with the ITK source tree, I have to admit  🤷

dzenanz (approved):

thewtex (approved):
Thanks @N-Dekker !

N-Dekker:
> Would be good to apply all of these to the ITK SW Guide (where appropriate, or the `Fill` method is not explicitly demonstrated), the ITK Sphinx Examples, and to the remotes.

@jhlegarreta Here is one for the ITK Sphinx Examples, to begin with:
- https://github.com/InsightSoftwareConsortium/ITKSphinxExamples/pull/428
Please have a look!

PR #4888, opened 2024-10-20
BUG: PyBuffer _GetImageViewFromArray should reject non-contiguous buffer

N-Dekker:
The C++ function `PyBuffer::_GetImageViewFromArray` cannot support a non-contiguous Python buffer as input, because the buffer is imported into an `itk::Image`, via `itk::ImportImageContainer::SetImportPointer`.

Note that the Python function `GetImageViewFromArray` (from Modules/Bridge/NumPy/wrapping/PyBuffer.i.in) _does_ support a non-contiguous NumPy array as input, by calling `numpy.ascontiguousarray`:

https://github.com/InsightSoftwareConsortium/ITK/blob/d36dfc65c7c3734872dc1bfedf16ce388e4ef33c/Modules/Bridge/NumPy/wrapping/PyBuffer.i.in#L90-L91

### DIFF
--- a/Modules/Bridge/NumPy/include/itkPyBuffer.hxx
+++ b/Modules/Bridge/NumPy/include/itkPyBuffer.hxx
@@ -59,9 +59,9 @@ PyBuffer<TImage>::_GetImageViewFromArray(PyObject * arr, PyObject * shape, PyObj
 
   SizeType size;
 
-  if (PyObject_GetBuffer(arr, &pyBuffer, PyBUF_ND | PyBUF_ANY_CONTIGUOUS) == -1)
+  if (PyObject_GetBuffer(arr, &pyBuffer, PyBUF_ANY_CONTIGUOUS) == -1)
   {
-    PyErr_SetString(PyExc_RuntimeError, "Cannot get an instance of NumPy array.");
+    PyErr_SetString(PyExc_RuntimeError, "Cannot get a contiguous buffer from the specified NumPy array.");
     return nullptr;
   }
 
@@ -84,8 +84,6 @@ PyBuffer<TImage>::_GetImageViewFromArray(PyObject * arr, PyObject * shape, PyObj
 
   const SizeValueType numberOfPixels = size.CalculateProductOfElements();
 
-  const bool isFortranContiguous = pyBuffer.strides != nullptr && pyBuffer.itemsize == pyBuffer.strides[0];
-
   const size_t len = numberOfPixels * numberOfComponents * sizeof(ComponentType);
   if (bufferLength < 0 || static_cast<size_t>(bufferLength) != len)
   {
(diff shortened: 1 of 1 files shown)


dzenanz (approved):
LGTM

Modules/Bridge/NumPy/include/itkPyBuffer.hxx
```diff
@@ -59,9 +59,9 @@ PyBuffer<TImage>::_GetImageViewFromArray(PyObject * arr, PyObject * shape, PyObj
 
   SizeType size;
 
-  if (PyObject_GetBuffer(arr, &pyBuffer, PyBUF_ND | PyBUF_ANY_CONTIGUOUS) == -1)
+  if (PyObject_GetBuffer(arr, &pyBuffer, PyBUF_ANY_CONTIGUOUS) == -1)
```
N-Dekker, line 62: - For the record, Matt (@thewtex) added the `PyBUF_ND` flag with https://github.com/InsightSoftwareConsortium/ITKBridgeNumPy/pull/14, Feb 24, 2017.

Modules/Bridge/NumPy/include/itkPyBuffer.hxx
```diff
@@ -94,9 +92,19 @@ PyBuffer<TImage>::_GetImageViewFromArray(PyObject * arr, PyObject * shape, PyObj
...
+    if (PyBuffer_IsContiguous(&pyBuffer, 'F') == 1)
+    {
+      // The buffer is Fortran contiguous (and not C-contiguous), so reverse the size elements.
+      std::reverse(size.begin(), size.end());
+    }
+    else
+    {
+      // The buffer is neither Fortran nor C-contiguous. This is unlikely to happen, because PyBUF_ANY_CONTIGUOUS was
+      // specified as argument, when retrieving the buffer. But if it _does_ happen, it's obviously an error.
+      PyErr_SetString(PyExc_RuntimeError, "The buffer should be contiguous, but it is not!");
```
thewtex, line 106: Here `return nullptr;` should be added to "throw" the CPython error.

thewtex (changes requested):
I did not intend `_GetImageViewFromArray` to be called in C++ outside of the Python binding path, but this change does not hurt.

See comment inline regarding throwing the error.

Modules/Bridge/NumPy/include/itkPyBuffer.hxx
N-Dekker: Good catch, thanks @thewtex ! Should be adressed by the [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/bd31ed47cc00047e3f2d28c587ae2acc5dd1116b..b769b81f648ecabf80a481f31ee44813ea4ea4dd) that I just did, please check!

thewtex (approved):

blowekamp:
> Presumably, byte order should be taken care of when reading/writing from disk, so there should be no issues with it once in memory.

It is reasonable to think that, but numpy does have representation to support non-native byte orders: https://numpy.org/doc/stable/reference/generated/numpy.dtype.byteorder.html 

And this have been encountered in the real world reading OME ZARR files generated by popular java libraries.

thewtex:
> And this have been encountered in the real world reading OME ZARR files generated by popular java libraries.

Are there public samples of these we could use in tests?

blowekamp:
> > And this have been encountered in the real world reading OME ZARR files generated by popular java libraries.
> 
> @blowekamp Are there public samples of these we could use in tests?

The  [raw2ometiff](https://github.com/glencoesoftware/raw2ometiff) or the GUI [NGFF-Converter](https://github.com/glencoesoftware/NGFF-Converter) tools produces big endian of data type more that 2 bytes.

PR #4889, opened 2024-10-21
BUG: Replace `SetPoints` with `SetPointsByCoordinates` in wrapping tests

N-Dekker:
When the points are specified by a flat VectorContainer of coordinates, passing those points to `SetPoints` may lead to undefined behavior. `SetPointsByCoordinates` is in that case preferred.

- Partially addresses issue #4848
- Related to pull request #4860

Aims to address a comment by @PranjalSahu at https://github.com/InsightSoftwareConsortium/ITK/pull/4860#issuecomment-2375249386 saying:
> We should also replace all the older tests using SetPoints with the new method to ensure everything else is working.

### DIFF
--- a/Modules/Core/Common/wrapping/test/itkPointSetSerializationTest.py
+++ b/Modules/Core/Common/wrapping/test/itkPointSetSerializationTest.py
@@ -52,7 +52,7 @@ for i in range(NumberOfPoints):
 arr = itk.array_view_from_vector_container(v_point)
 points_vc = itk.vector_container_from_array(arr.flatten())
 
-point_set.SetPoints(points_vc)
+point_set.SetPointsByCoordinates(points_vc)
 
 
 # Check the serialization of mesh
--- a/Modules/Core/Mesh/wrapping/test/itkMeshSerializationTest.py
+++ b/Modules/Core/Mesh/wrapping/test/itkMeshSerializationTest.py
@@ -50,7 +50,7 @@ for i in range(NumberOfPoints):
 arr = itk.array_view_from_vector_container(v_point)
 points_vc = itk.vector_container_from_array(arr.flatten())
 
-mesh.SetPoints(points_vc)
+mesh.SetPointsByCoordinates(points_vc)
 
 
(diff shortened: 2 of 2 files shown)


PranjalSahu (approved):
LGTM
The OAI methods project needs to be updated for newer version of ITK.
Here, [    itk_mesh.SetPoints(itk.vector_container_from_array(points_numpy))](https://github.com/uncbiag/OAI_analysis_2/blob/728fc41a66ae3c159ed41bde8365792d0fd8b617/oai_analysis/mesh_processing.py#L58)

Modules/Core/Common/wrapping/test/itkPointSetSerializationTest.py
```diff
@@ -52,7 +52,7 @@
 arr = itk.array_view_from_vector_container(v_point)
 points_vc = itk.vector_container_from_array(arr.flatten())
 
-point_set.SetPoints(points_vc)
```
thewtex, line 55: Can we make this "in addition" instead of "instead of" to improve coverage, i.e.,

```
point_set.SetPoints(points_vc)
point_set.SetPointsByCoordinates(points_vc)
```

N-Dekker: @thewtex I see your point; this pull request would take away code coverage from `PointSet::SetPoints(PointsVectorContainer *)`. But then again, I think we should eventually deprecate this specific `SetPoints` overload anyway, to avoid undefined behavior (issue #4848).

If we still want to have this specific `SetPoints` overload around for a while, I would prefer a separate isolated test for this specific overload.

thewtex: @N-Dekker yes, that is a good approach to both indicate that the code is unsafe and keep the coverage.

N-Dekker:
- Note: this pull request might be superseded by pull request #4896

N-Dekker:
- This pull request _is_ now superseded by pull request #4896  😃

PR #4884, opened 2024-10-17
STYLE: Replace `Fill(0.)` and `Fill(0u)` with `{}` initializer for local variables in tests

N-Dekker:
Replaced `Fill(0.)`, `Fill(0.0)`, `Fill(0.0f)`, and `Fill(0u)` calls with `{}` initializers.

- Follow-up to pull request #4881

PR #4891, opened 2024-10-22
STYLE: Replace Fill calls with `auto var = itk::MakeFilled<T>` in tests

N-Dekker:
Replaced code of the form

    T var;
    var.Fill(x);

with `auto var = itk::MakeFilled<T>(x);`

Following C++ Core Guidelines, Oct 3, 2024, "Always initialize an object", https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-always

    Find what: ^( [ ]+)([^ ].*)[ ]+(\w+);[\r\n]+\1\3\.Fill\(
    Replace with: $1auto $3 = itk::MakeFilled<$2>\(
    Filters: itk*Test*.cxx
    [v] Match case
    (*) Regular expression

Follow-up to
- pull request https://github.com/InsightSoftwareConsortium/ITK/pull/4881
- pull request https://github.com/InsightSoftwareConsortium/ITK/pull/4884
- pull request https://github.com/InsightSoftwareConsortium/ITK/pull/4887

### DIFF
--- a/Modules/Core/Common/test/itkAdaptorComparisonTest.cxx
+++ b/Modules/Core/Common/test/itkAdaptorComparisonTest.cxx
@@ -204,8 +204,7 @@ itkAdaptorComparisonTest(int, char *[])
   vector_image->SetRegions(region);
   vector_image->Allocate();
 
-  VectorImageType::PixelType initialVectorValue;
-  initialVectorValue.Fill(1.2345); // arbitrary value;
+  auto initialVectorValue = itk::MakeFilled<VectorImageType::PixelType>(1.2345); // arbitrary value;
   vector_image->FillBuffer(initialVectorValue);
 
   // Time trials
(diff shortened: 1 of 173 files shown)


dzenanz (approved):
Looks good on a glance.

N-Dekker:
Honestly I find the syntax `auto var = itk::MakeFilled<T>(x)` a bit verbose. For zero-filling I would certainly prefer the syntax `T var{}`. But I still think `auto var = itk::MakeFilled<T>(x)` is very much preferable to declaring the variable uninitialized and calling `Fill` afterwards.

As a follow-up, the variable might in some cases be declared constexpr:
```cpp
    // Assuming that the fill value x is also a compile-time constant:
    constexpr auto var = itk::MakeFilled<T>(x); 
```
😃

PR #4894, opened 2024-10-23
ENH: Add STL-style iterators and size() member function to `itk::Matrix`

N-Dekker:
Forwarded iterators and `size()` from its internal `m_Matrix`. This commit enables calling `itk::MakeFilled<MatrixType>(fillValue)` to create a filled matrix.

### DIFF
--- a/Modules/Core/Common/include/itkMatrix.h
+++ b/Modules/Core/Common/include/itkMatrix.h
@@ -59,6 +59,24 @@ public:
   using ValueType = T;
   using ComponentType = T;
 
+  /** A pointer to the ValueType. */
+  using pointer = ValueType *;
+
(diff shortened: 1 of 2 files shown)


dzenanz (approved):

dzenanz:
What is the use case for filled matrix? Usually we want an identity matrix.

N-Dekker:
> What is the use case for filled matrix? Usually we want an identity matrix.

- Good question 🤔  The CI failures from pull request #4891 _do_ show that these filled matrix are actually being used. Or at least, they are being tested 😃 

Moreover, I think an STL-style interface to itk::Matrix is nice to have, anyway  😃

thewtex:
@N-Dekker awesome!

PR #4899, opened 2024-10-24
COMP: Fix AnatomicalOrientation Enum

blowekamp:
Addresses compilation errors on optional module.

### DIFF
--- a/Modules/IO/PhilipsREC/src/itkPhilipsRECImageIO.cxx
+++ b/Modules/IO/PhilipsREC/src/itkPhilipsRECImageIO.cxx
@@ -710,7 +710,7 @@ PhilipsRECImageIO::ReadImageInformation()
   AffineMatrix spacing;
   spacing.SetIdentity();
 
-  AnatomicalOrientation coord_orient(AnatomicalOrientation::FromEnum::RSA);
+  AnatomicalOrientation coord_orient(AnatomicalOrientation::NegativeEnum::RSA);
 
   switch (par.sliceorient)
   {


dzenanz (approved):

thewtex (approved):
🚀 @blowekamp thanks for the quick follow-up

Modules/IO/PhilipsREC/src/itkPhilipsRECImageIO.cxx
```diff
@@ -710,7 +710,7 @@ PhilipsRECImageIO::ReadImageInformation()
   AffineMatrix spacing;
   spacing.SetIdentity();
 
-  AnatomicalOrientation coord_orient(AnatomicalOrientation::FromEnum::RSA);
+  AnatomicalOrientation coord_orient(AnatomicalOrientation::NegativeEnum::RSA);
```
N-Dekker, line 713: Thanks @blowekamp! Just for my understanding, why `NegativeEnum`, why not `PositiveEnum` ? They both have an `RSA` constant, right?

blowekamp: They were originally ToEnum, and FromEnum, then renamed to PositiveEnum and NegativeEnum. That change was applied here. There should be a description of these two ambiguous notations in the AnatomicalOrientation class. Please read it and let me know if you have specific suggestions to improve clarity.

This particular value is not used, but is require as AnatomicalOrientation is not default constructible ( perhaps INVALID) would be a reasonable default?

PR #4903, opened 2024-10-25
STYLE: Remove unused "one-filled" `direction` matrices from tests

N-Dekker:
These matrices were just filled with `1`, but never used afterwards.

This commit prevents "unused variable" warnings, which would occur when the matrices would be filled during their initialization, instead of by calling the `Fill` member function.

----

- Addresses warnings from https://open.cdash.org/viewBuildError.php?type=1&buildid=9983771 (triggered by pull request #4891) saying _warning: unused variable 'direction' [-Wunused-variable]_


N-Dekker:
@dzenanz This still reminds me of your comment at https://github.com/InsightSoftwareConsortium/ITK/pull/4894#issuecomment-2432376027:
> What is the use case for filled matrix?

I'm not _entirely_ sure if these "one-filled" matrices answer your question...  😺

dzenanz (approved):

PR #4909, opened 2024-10-28
ENH: Suppress irrelevant TIFF warnings

jcfr:
_Submitted on behalf of @lassoan by cherry-picking commit from [lassoan/ITK@tiff_unknown_tag_warning_suppress/](https://github.com/lassoan/ITK/commits/tiff_unknown_tag_warning_suppress/)_

LibTIFF logs hundreds of messages like "TIFFReadDirectory: Warning, Unknown field with tag 65280 (0xff00) encountered." when reading a volume from a TIFF file list that contains custom tags. Presence of custom tags is harmless and it should not flood the console or application log with warnings, because message display can significantly slow down image loading and excessive amount of irrelevant messages makes it difficult to notice important messages.

For background, see https://github.com/SlicerMorph/SlicerMorph/issues/320

## PR Checklist

### DIFF
--- a/Modules/IO/TIFF/src/itkTIFFImageIO.cxx
+++ b/Modules/IO/TIFF/src/itkTIFFImageIO.cxx
@@ -24,8 +24,33 @@
 
 #include "itk_tiff.h"
 
+#include "itksys/SystemTools.hxx"
+
 namespace itk
 {
+
+TIFFErrorHandler _ITKTIFFsavedWarningHandler = nullptr;
+
+// This warning handler suppresses log messages starting with "Unknown field with tag" because these kind of messages
+// are harmless (they just mean that the TIFF file contains custom tags) and they would flood the console when reading images
+// that contains hundreds of frames, hiding actual errors or relevant warnings.
+static void
+ITKTIFFTagNotFoundSuppressWarningHandler(const char * module, const char * fmt, va_list ap)
+{
+  if (fmt != nullptr)
+  {
+    if (itksys::SystemTools::StringStartsWith(fmt, "Unknown field with tag"))
+    {
+      // ignore unknown field warnings
+      return;
+    }
+  }
+  // call the saved warning handler
+  if (_ITKTIFFsavedWarningHandler != nullptr)
+  {
+    _ITKTIFFsavedWarningHandler(module, fmt, ap);
+  }
+}
 
 bool
 TIFFImageIO::CanReadFile(const char * file)
@@ -40,16 +65,18 @@ TIFFImageIO::CanReadFile(const char * file)
   }
 
   // Now check if this is a valid TIFF image
-  TIFFErrorHandler save = TIFFSetErrorHandler(nullptr);
-  int              res = m_InternalImage->Open(file);
-  if (res)
+  TIFFErrorHandler savedErrorHandler = TIFFSetErrorHandler(nullptr);
+  TIFFErrorHandler savedWarningHandler = TIFFSetWarningHandler(ITKTIFFTagNotFoundSuppressWarningHandler);
+  _ITKTIFFsavedWarningHandler = savedWarningHandler;
+  bool success = (m_InternalImage->Open(file) != 0);
+  if (!success)
   {
-    TIFFSetErrorHandler(save);
-    return true;
+    m_InternalImage->Clean();
   }
-  m_InternalImage->Clean();
-  TIFFSetErrorHandler(save);
-  return false;
+  TIFFSetErrorHandler(savedErrorHandler);
+  TIFFSetWarningHandler(savedWarningHandler);
+  _ITKTIFFsavedWarningHandler = nullptr;
+  return success;
 }
 
 void
@@ -152,6 +179,9 @@ TIFFImageIO::ReadVolume(void * buffer)
   const size_t width{ m_InternalImage->m_Width };
   const size_t height{ m_InternalImage->m_Height };
 
+  TIFFErrorHandler savedWarningHandler = TIFFSetWarningHandler(ITKTIFFTagNotFoundSuppressWarningHandler);
+  _ITKTIFFsavedWarningHandler = savedWarningHandler;
+
   for (uint16_t page = 0; page < m_InternalImage->m_NumberOfPages; ++page)
   {
     if (m_InternalImage->m_IgnoredSubFiles > 0)
@@ -175,6 +205,9 @@ TIFFImageIO::ReadVolume(void * buffer)
 
     TIFFReadDirectory(m_InternalImage->m_Image);
   }
+
+  TIFFSetWarningHandler(savedWarningHandler);
+  _ITKTIFFsavedWarningHandler = nullptr;
 }
 
 void


jcfr:
@lassoan After testing, I realized the `Clean()` function should be cleaned only in case of error. If not, there will be a seg fault 

```
--- a/Modules/IO/TIFF/src/itkTIFFImageIO.cxx
+++ b/Modules/IO/TIFF/src/itkTIFFImageIO.cxx
@@ -69,7 +69,7 @@ TIFFImageIO::CanReadFile(const char * file)
   TIFFErrorHandler savedWarningHandler = TIFFSetWarningHandler(ITKTIFFTagNotFoundSuppressWarningHandler);
   _ITKTIFFsavedWarningHandler = savedWarningHandler;
   bool success = (m_InternalImage->Open(file) != 0);
-  if (success)
+  if (!success)
   {
     m_InternalImage->Clean();
   }

```

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
```diff
@@ -40,16 +65,18 @@ TIFFImageIO::CanReadFile(const char * file)
...
-  int              res = m_InternalImage->Open(file);
-  if (res)
-  {
-    TIFFSetErrorHandler(save);
-    return true;
-  }
-  m_InternalImage->Clean();
-  TIFFSetErrorHandler(save);
-  return false;
+  TIFFErrorHandler savedErrorHandler = TIFFSetErrorHandler(nullptr);
```
blowekamp, line 68: So the problem with this is that this changes globals in TIFF and non-thread safe way.

The error messages should also go through the ITK messaging system and not straight to stdout/err the way the TIFF error handler does

Starting with libtiff 4.5 introduces error handing that is connected to the TIFF handle ( re-entrant ) and thus not global. 

https://libtiff.gitlab.io/libtiff/functions/TIFFError.html#c.TIFFErrorExtR

blowekamp (changes requested):
Unfortunately, we need to upgrade libtiff to address the warning output issues in a way that is safe.

blowekamp:
Yes, I think this was my WIP:


+ int _itkTIFFErrorHandlerExtR([[maybe_unused]] TIFF *tif, [[maybe_unused]] void *user_data, const char *module, const char *fmt, va_list ap)
+ {
+ 
+   if (::itk::Object::GetGlobalWarningDisplay())
+   {
+     char out[256];
+     vsnprintf(out, 256, fmt, ap);
+     std::ostringstream itkmsg;
+     itkmsg << "WARNING: libtiff(" << (module?module:"") << ") message: " << out << std::endl;
+     ::itk::OutputWindowDisplayWarningText(itkmsg.str().c_str());
+   }
+   return 1;
+ }
+ 
+ int _itkTIFFWarningHandlerExtR(TIFF *tif, void *user_data, const char *module, const char *fmt, va_list ap)
+ {
+   return _itkTIFFErrorHandlerExtR(tif, user_data, module, fmt, ap);
+ }
+ 
  int
--TIFFReaderInternal::Open(const char * filename)
++TIFFReaderInternal::Open(const char * filename, bool silent)
  {
    this->Clean();
    struct stat fs;
@@@ -42,7 -68,7 +69,22 @@@
  #endif
    }
  
--  this->m_Image = TIFFOpen(filename, "r");
++#if defined  TIFFLIB_AT_LEAST
++  std::unique_ptr<TIFFOpenOptions,  tiff_open_options_free> options(TIFFOpenOptionsAlloc());
++  if (silent)
++  {
++    TIFFOpenOptionsSetErrorHandlerExtR(options.get(), nullptr, nullptr);
++    TIFFOpenOptionsSetWarningHandlerExtR(options.get(), nullptr, nullptr);
++  }
++  else
++  {
++    TIFFOpenOptionsSetErrorHandlerExtR(options.get(), _itkTIFFErrorHandlerExtR, nullptr);
++    TIFFOpenOptionsSetWarningHandlerExtR(options.get(), _itkTIFFWarningHandlerExtR, nullptr);
++  }
++  this->m_Image = TIFFOpenExt(filename, "r", options.get());


Does Slicer have a more uptodate version of libtiff in the super build?

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
jcfr: > So the problem with this is that this changes globals in TIFF and non-thread safe way.

I see ... the current use of `TIFFSetErrorHandler` is already problematic then ?

How about using a Schwartz Counter to ensure the handler is installed on library loading ?

jcfr:
> Does Slicer have a more uptodate version of libtiff in the super build?

No. We rely on ITK to support TIFF image loading.

> re: wip

I think instead of using `OutputWindowDisplayWarningText` we should not report those at all, or if we do, it should not happen by default.

lassoan:
Thanks a lot @blowekamp, this fix is definitely better than my original proposal. It would still need the filtering of the useless unknown field warning.

To suppress the warning we could add the same message parsing as in my original proposal, but perhaps a simpler and more robust solution would be to suppress this specific warning in libtiff. Hundreds of other projects struggle with the same issue and a [fix is available](https://gitlab.com/libtiff/libtiff/-/merge_requests/463) but libtiff maintainers have been reluctant to merge it. I've added a note, but I'm not sure if this will make the maintainers reconsider. If they don't merge the fix then we could fork libtiff and merge the fix and then we would not need to implement the message filtering.

lassoan:
@jcfr thanks a lot for testing and the fix. I experienced crash, too, but I was not sure if it was because I have not rebuilt everything and wanted to do a clean build.

blowekamp:
> It would still need the filtering of the useless unknown field warning.

There are other ImageIO's that produce a large number of warnings too. Including DICOM with the MONOCHROME tag and MCRImageIO, and others.

There are a couple of other options to consider here.
 1. After this warning message is place into the ITK messaging system, applications can filter the outputs with a custom ITK logger.
 2. There is an existing itk::ProcessObject::GlobalWarningDisplay  property. This could also be used.
 3. Another option to consider it to add a property to ImageIO::Base WarningDisplay (or something similar) to suppress all warnings from IO.

These are some other options to consider.

lassoan:
Ideally, we would only want to suppress this specific warning message from libtiff.

> After this warning message is place into the ITK messaging system, applications can filter the outputs with a custom ITK logger.

Parsing all ITK messages all the time in a custom ITK logger, just to get rid of these warnings would not be very efficient.

> - There is an existing itk::ProcessObject::GlobalWarningDisplay property. This could also be used.
> - Another option to consider it to add a property to ImageIO::Base WarningDisplay (or something similar) to suppress all warnings from IO.

We would not want to suppress all messages from libtiff.

Overall, adding filtering in the TIFF logger redirection would or fixing libtiff (forking or pushing through [this fix](https://gitlab.com/libtiff/libtiff/-/merge_requests/463)) would work better, as they would suppress the irrelevant messages, but nothing else; and they would be simple and efficient.

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
```diff
@@ -152,6 +179,9 @@ TIFFImageIO::ReadVolume(void * buffer)
   const size_t width{ m_InternalImage->m_Width };
   const size_t height{ m_InternalImage->m_Height };
 
+  TIFFErrorHandler savedWarningHandler = TIFFSetWarningHandler(ITKTIFFTagNotFoundSuppressWarningHandler);
```
blowekamp, line 182: Setting this seems to stop the warning being produced during my testing.

blowekamp:
> Parsing all ITK messages all the time in a custom ITK logger, just to get rid of these warnings would not be very efficient.

How many warning messages is your application getting? 🤨

lassoan:
It depends. Usually not a lot. Log files are typically not longer than a few MB. Still, parsing every string, especially if it is done in the main thread might be noticeable.

It would be also nice to avoid parsing, because the message might change in the future. So, the best would be to fix libtiff.

If libtiff maintainers don't merge the fix, do you think ITK might merge the fix to its libtiff?

blowekamp:
> It depends. Usually not a lot. Log files are typically not longer than a few MB. Still, parsing every string, especially if it is done in the main thread might be noticeable.

I am rather skeptical that is would be measurable.  I had been planning to improve the logging/warning message system in ITK to aid in filtering warning and improve compatibility with other loggers. Primarily allowing the lower to compose the message by giving it the components of filename, line number, warning/error/level and message.


> 
> It would be also nice to avoid parsing, because the message might change in the future. So, the best would be to fix libtiff.

Most of the time the message is rather annoying. However, it can be useful with something like an OME-TIFF which has large amount of structure that will not be correctly handled. There are other microscopy TIFF-sh file that can have other necearry information in the TIFF tags to load the image correctly too, although I have not encountered that recently.

> 
> If libtiff maintainers don't merge the fix, do you think ITK might merge the fix to its libtiff?

ITK's libtiff is very old, and in need of an update which may require some non-trivial effort. 

For your application is there a couple common tags that are being reported? Another option may be to just register these tag that are annoying, that don't effect the reading of the images. For example the [ImageJ TIFF TAGS.](https://discourse.itk.org/t/unknown-tags-reading-tiff/1719/2?u=blowekamp)

dzenanz:
I believe that @lassoan's application is 3D Slicer. Is that right Andras?

dzenanz:
- [x] That [PR in libTIFF](https://gitlab.com/libtiff/libtiff/-/merge_requests/463) just got merged.

lassoan:
Yes, latest libtiff has now the option for turning off this warning!
I'll try to build ITK with the latest libtiff and see how it goes.

blowekamp:
> Yes, latest libtiff has now the option for turning off this warning!
> I'll try to build ITK with the latest libtiff and see how it goes.

I have been testing with a 4.6.0 version locally and it has mostly been OK. The issue it that it depends on several other libraries for compression and that I don't readily have. So some if the ITK TIFF compression tests fail.

lassoan:
I've backported the fix from libtiff to ITK's version: https://github.com/InsightSoftwareConsortium/ITK/pull/4920
(it was much easier to do this than upgrade ITK's libtiff)

dzenanz:
Should this be closed now, given that #4917 and #4920 were merged?

hjmjohnson:
I think is now out of date now that that https://github.com/InsightSoftwareConsortium/ITK/pull/4917 and https://github.com/InsightSoftwareConsortium/ITK/pull/4920 and #4961 have been merged.

PR #4912, opened 2024-10-29
ENH: Add `PointSet.GraftDoesShallowCopy` GoogleTest unit test

N-Dekker:
Tests that `PointSet::Graft` just copies the *pointers* to the points and the data. It does not do a "deep copy".

### DIFF
--- a/Modules/Core/Common/test/itkPointSetGTest.cxx
+++ b/Modules/Core/Common/test/itkPointSetGTest.cxx
@@ -18,6 +18,7 @@
 
 // First include the header file to be tested:
 #include "itkPointSet.h"
+#include "itkDeref.h"
 #include "../../QuadEdgeMesh/include/itkQuadEdgeMeshTraits.h"
 #include <gtest/gtest.h>
 #include <algorithm> // For equal.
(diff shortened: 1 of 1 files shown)


Modules/Core/Common/test/itkPointSetGTest.cxx
```diff
@@ -80,3 +81,35 @@ TEST(PointSet, SetPointsByCoordinates)
   TestSetPointsByCoordinates(*itk::PointSet<int>::New());
   TestSetPointsByCoordinates(*itk::PointSet<double, 2, itk::QuadEdgeMeshTraits<double, 2, bool, bool>>::New());
 }
+
+
+// Tests that `PointSet::Graft` just copies the *pointers* to the points and the data. It does not do a "deep copy".
+TEST(PointSet, GraftDoesShallowCopy)
+{
+  const auto check = [](const auto & pointSet) {
+    const auto clone = pointSet.Clone();
```
dzenanz, line 90: This tests not only that Graft does not do a deep copy, but that Clone does not do a deep copy either.

N-Dekker: Indeed @dzenanz Note that `pointSet.Clone()` does not even do a _shallow_ copy! It just creates an empty `PointSet` that has the same `PixelType` and `PointType` as the original. Is that OK?

Obviously the test just checks the behavior that `PointSet::Graft` always had. It seems useful to me to support deep copying, but adjusting `PointSet::Graft` to do so would possibly break legacy code. 🤷

dzenanz (approved):

Modules/Core/Common/test/itkPointSetGTest.cxx
N-Dekker: FYI, I'm considering to add deep copy functionality to PointSet and PointSetBase:
https://github.com/InsightSoftwareConsortium/ITK/compare/master...N-Dekker:ITK:PointSetBase-Clone-MakeDeepCopy

thewtex (approved):
:100:

PR #4913, opened 2024-10-29
STYLE: Rename _GetImageViewFromArray to _get_image_view_from_contiguous_array

N-Dekker:
Aims to make the difference clearer between the C++ `itk::PyBuffer` member
function (which only supports contiguous arrays as input) and the Python method
`itkPyBuffer.GetImageViewFromArray` (which supports *both* contiguous and
non-contiguous arrays).

The new name of the member function is more Pythonic (using snake case), but it
is also more in accordance with the C++ standard (which reserves identifiers
that begin with an underscore, followed by an uppercase letter).
- Follow-up to pull request #4888

### DIFF
--- a/Modules/Bridge/NumPy/include/itkPyBuffer.h
+++ b/Modules/Bridge/NumPy/include/itkPyBuffer.h
@@ -73,10 +73,11 @@ public:
   _GetArrayViewFromImage(ImageType * image);
 
   /**
-   * Get an ITK image from a Python array
+   * Get an ITK image from a contiguous Python array. Internal helper function for the implementation of
+   * `itkPyBuffer.GetImageViewFromArray`.
    */
   static OutputImagePointer
-  _GetImageViewFromArray(PyObject * arr, PyObject * shape, PyObject * numOfComponent);
+  _get_image_view_from_contiguous_array(PyObject * arr, PyObject * shape, PyObject * numOfComponent);
 };
 
 } // namespace itk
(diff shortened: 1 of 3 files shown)


dzenanz (approved):

thewtex (commented):
@N-Dekker in Python, they way to indicate that a function should be treated as "private", i.e. it is not intended to be called directly, its interface may change without notice, etc. is to start with an underscore. This is the standard Python convention. We may be making the code more C++ standard friendly by removing the underscore, but the Python convention functionality is more important in this case. We do not want to make this a function that people think they can just call directly. Reference-counting issues, e.g.

https://github.com/InsightSoftwareConsortium/ITK/blob/4b7886a816511cd07b31d14cb825620187d22f7d/Modules/Bridge/NumPy/wrapping/PyBuffer.i.in#L109

make it a foot-gun. And the interface is not intended for general consumption.

N-Dekker:
@thewtex Thank you for explaining! Now I see at https://docs.python.org/3.13/tutorial/classes.html#private-variables

> there is a convention that is followed by most Python code: a name prefixed with an underscore (e.g. `_spam`) should be treated as a non-public part of the API (whether it is a function, a method or a data member). It should be considered an implementation detail and subject to change without notice.

However, the C++ standard specifically reserves identifiers beginning with an underscore, followed by an _uppercase_ letter. So then, would it be OK to you to change the proposed name to underscore + _lowercase_ letter (`_getImageViewFromContiguousArray`)? 

Of course, we can also go all the way Pythonic, and use snake case: `_get_image_view_from_contiguous_array`. That's also fine by me. As long as we can avoid challenging the C++ standard  😇 What do you think?

thewtex:
> _get_image_view_from_contiguous_array

@N-Dekker yes, this is compatible with both Python and C++ :+1:

dzenanz (approved):

thewtex (approved):
:1st_place_medal:

PR #4895, opened 2024-10-23
Use AnatomicalOrientation in  OrientImageFilter

blowekamp:
Preserve compatibility with SpatialOrientation enums with implicit converstion. Additionally both the PositiveEnum and NegativeEnum are supported.

## PR Checklist
- [ ] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [ ] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [ ] Added test (or behavior not changed)
- [ ] Updated API documentation (or API not changed)
- [ ] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [ ] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [ ] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples) for all new major features (if any)

Refer to the [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) for
further development details if necessary.

PR #4918, opened 2024-10-30
Pixi python-exe task

thewtex:
Call `pixi run python-exe` to run a python interpreter with the locally built `itk` Python package.

Facilitates development and debugging.

I used this in testing / debugging #4895.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -106,6 +106,12 @@ cmd = "ctest -j3 --test-dir build-python"
 description = "Test ITK Python"
 depends-on = ["build-python"]
 
+[tool.pixi.feature.python.tasks.python-exe]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -106,6 +106,12 @@ cmd = "ctest -j3 --test-dir build-python"
 description = "Test ITK Python"
 depends-on = ["build-python"]
 
+[tool.pixi.feature.python.tasks.python-exe]
+cmd = '''cp ./build-python/Wrapping/Generators/Python/WrapITK.pth $(python -c "import site; print(next(p for p in site.getsitepackages() if \'site-packages\' in p))") &&
```
jhlegarreta, line 110: Will this path stay the same across systems (OS, local vs remote, etc.)?

jhlegarreta (approved):
Left a question inline.

pyproject.toml
thewtex: It should. I tested on Linux.

dzenanz (approved):
LGTM

hjmjohnson (approved):

PR #4923, opened 2024-10-31
ENH: Add Clone() and MakeDeepCopy() to PointSetBase

N-Dekker:
`PointSetBase::MakeDeepCopy()` allows making a "deep copy" of a point set (rather than `PointSet::Graft`, which does a "shallow copy", or `Clone()`, which does not copy at all).

----

Background: for elastix (ITK/Python interface), we would like to be able to make a "deep copy" of a point set, including its points and point data. Unfortunately `PointSet::Clone()` doesn't do it. It does not copy anything at all!!! An alternative solution would be to adjust `PointSet::Clone()`, so that it would do the deep copy. Would that be better? What do you think?

### DIFF
--- a/Modules/Core/Common/include/itkPointSetBase.h
+++ b/Modules/Core/Common/include/itkPointSetBase.h
@@ -67,6 +67,8 @@ public:
   /** \see LightObject::GetNameOfClass() */
   itkOverrideGetNameOfClassMacro(PointSetBase);
 
+  itkCloneMacro(Self);
+
   /** Convenient type alias obtained from TPointsContainer template parameter. */
   using PointType = typename TPointsContainer::Element;
   using CoordRepType = typename PointType::CoordRepType;
@@ -190,6 +192,25 @@ public:
 
   itkGetConstMacro(BufferedRegion, RegionType);
 
+
+  /** Returns a "deep copy" of this point set, having copied its points and point data. */
+  Pointer
+  MakeDeepCopy() const
+  {
+    const auto clone = this->Clone();
+    clone->Graft(this);
+
+    if (m_PointsContainer)
+    {
+      // Make a new copy of the points, detached from the original one.
+      const auto points = TPointsContainer::New();
+      points->CastToSTLContainer() = m_PointsContainer->CastToSTLConstContainer();
+      clone->m_PointsContainer = points;
+    }
+    clone->DetachPointData();
+    return clone;
+  }
+
 protected:
   /** Default-constructor, to be used by derived classes. */
   PointSetBase() = default;
@@ -215,6 +236,10 @@ protected:
   RegionType m_RequestedNumberOfRegions{};
   RegionType m_BufferedRegion{ -1 };
   RegionType m_RequestedRegion{ -1 };
+
+private:
+  virtual void
+  DetachPointData() = 0;
 };
 } // end namespace itk
 
--- a/Modules/Core/Common/include/itkPointSet.h
+++ b/Modules/Core/Common/include/itkPointSet.h
@@ -162,6 +162,18 @@ protected:
   void
   PrintSelf(std::ostream & os, Indent indent) const override;
 
+private:
+  void
+  DetachPointData() override
+  {
+    if (m_PointDataContainer)
+    {
+      // Make a new copy of the point data, detached from the original one.
+      const auto pointData = PointDataContainer::New();
+      pointData->CastToSTLContainer() = m_PointDataContainer->CastToSTLConstContainer();
+      m_PointDataContainer = pointData;
+    }
+  }
 }; // End Class: PointSet
 } // end namespace itk
 
--- a/Modules/Core/Common/test/itkPointSetGTest.cxx
+++ b/Modules/Core/Common/test/itkPointSetGTest.cxx
@@ -73,6 +73,42 @@ TestSetPointsByCoordinates(TPointSet & pointSet)
     }
   }
 }
+
+template <typename TPointSet>
+void
+TestMakeDeepCopy(const TPointSet & pointSet)
+{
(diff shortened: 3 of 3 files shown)


Modules/Core/Common/include/itkPointSetBase.h
```diff
@@ -190,6 +192,25 @@ class ITK_TEMPLATE_EXPORT PointSetBase : public DataObject
...
 
+
+  /** Returns a "deep copy" of this point set, having copied its points and point data. */
+  Pointer
+  MakeDeepCopy() const
+  {
+    const auto clone = this->Clone();
+    clone->Graft(this);
+
+    if (m_PointsContainer)
```
dzenanz, line 203: Why not put this into a method `DetachPoints`? The could here would then read:
```cpp
clone->DetachPoints();
clone->DetachPointData();
```
Which is clearer.

blowekamp:
We already have "Clone" and "CreateAnother"... This PR in now introducing a "MakeDeepCopy" method... 

Any guidelines on what each of these method should do and how they are different? Or do we just create new methods with new name as to not "break" anything?

dzenanz:
Maybe implement `virtual LightObject::Pointer InternalClone() const;`? And make `Clone()` function do deep copies in all classes? Do we know why `Clone()` doesn't do a deep copy?

Modules/Core/Common/include/itkPointSetBase.h
N-Dekker: Well observed, @dzenanz I added `DetachPointData()` as a virtual member function of `PointSetBase` is because it has to be implemented by its derived `PointSet` class (as an `override`). `PointSetBase` cannot detach its point data directly, because it is managed by its derived class. On the other hand, `PointSetBase` _is_ able to detach its own _points_, though, so it does not need the virtual/override mechanism.

With this in mind, would you still like it if we would add a _non_-virtual private `DetachPoints()` member function to `PointSetBase`?

blowekamp:
> Maybe implement `virtual LightObject::Pointer InternalClone() const;`? And make `Clone()` function do deep copies in all classes? Do we know why `Clone()` doesn't do a deep copy?

It does in some, like the transform. It would be manual implementation in each class and hard to keep up today.

N-Dekker:
> Do we know why Clone() doesn't do a deep copy?

Probably copying "nothing" is just the default. ITK classes get a `Clone()` member function almost for free. It's up to the authors of the ITK class to let it actually copy its data. I guess 🤷

Modules/Core/Common/include/itkPointSetBase.h
dzenanz: non-virtual **proteced** DetachPoints() member function to PointSetBase would be my preference

When looking at this code at first, I thought you are copying points twice. Then I realized once is for `points`, the other time for `point data`.

dzenanz:
Maybe time to change pointset so clone works in an intuitive way? And `Image::Clone()` does what DuplicateImage does? With a new major version number, we have more leeway for breaking changes. @thewtex what do you think?

Modules/Core/Common/include/itkPointSetBase.h
N-Dekker: I would prefer to only make a member _protected_ when it is actually necessary to be used by a derived class. I find the `private` keyword much clearer, in order to communicate that the member is _for internal use only_. We could initially make `DetachPoints()` `private`, and then maybe later change it to `protected`, but only when it is really needed for derived classes. Would that be OK to you?

thewtex:
I see and agree with all of the points made:

- `::Clone()` is expected to do a deep copy in accordance with its name and the behavior of Transform's and SpatialObject's.
- `InternalClone()` where the copy should be implemented following usage in `itkCloneMacro`
- We want to avoid many methods with similar names but different behavior for simplicity and to avoid confusion.
- We should be careful to avoid silent, backwards-incompatible changes in behavior, which can cause confusing and hard-to-identify breakage downstream.

What do you all think about emitting a warning in new deep copy code that could be silenced with `ITK_FUTURE_LEGACY_REMOVE` and possibly another more specific override?

N-Dekker:
Thanks @thewtex I don't think the original "copy nothing" version of `PointSet::Clone()` would have been used much. I guess users who want to have another PointSet object of the same type would just call `CreateAnother()` instead.

Moreover, user code that _did_ in the past mistakenly call `Clone()` when they would just want to have another object of the same type would still get what they want, after pull request #4925: another object of the same type. (Moreover: it would then also have copied the data.) 

- So I'm still in favor of pull request #4925 as it is now. But of course, we should then clearly inform the users about the improvement, at discourse as well as in the Release Notes.

N-Dekker:
@thewtex Thanks! What do you think of the following text, to be added to itk_6_migration_guide.md with pull request #4925? Is it OK? (I'm trying to avoid triggering the CI over such textual changes.)

```
Class changes
-------------

The `Clone()` member function of `itk::PointSet` now does a "deep copy" of its
data, creating a new instance that has a copy of the points, of the point data
and the region information properties of the original PointSet object. With
previous ITK versions, `PointSet::Clone()` did not copy any data. (It previously
just created a default-constructed PointSet object, like
`PointSet::CreateAnother()` does.)

```

dzenanz:
I remember that we exclude some non-code changes from triggering the CI.

thewtex:
> What do you think of the following text, to be added to itk_6_migration_guide.md with pull request #4925? Is it OK?

@N-Dekker looks great, thanks!

dzenanz:
Should this PR be closed, now that #4925 has been merged?

N-Dekker:
- Closed, because it is superseded by pull request #4925, which is merged

PR #4924, opened 2024-11-01
Replace `var.Fill(x)` with `MakeFilled<T>(x)`

N-Dekker:
- Follow-up to pull request #4891
- Following C++ Core Guidelines, Oct 3, 2024, ["Always initialize an object"](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-always)


dzenanz (approved):

N-Dekker:
Note to self: After this PR, I hope I have some time to do follow-ups on the ITK SW Guide, examples, etc., as suggested by Jon at https://github.com/InsightSoftwareConsortium/ITK/pull/4887#pullrequestreview-2380285121

Modules/Segmentation/LevelSetsv4/include/itkUpdateWhitakerSparseLevelSet.hxx
```diff
@@ -860,8 +853,7 @@ UpdateWhitakerSparseLevelSet<VDimension, TLevelSetValueType, TEquationContainer>
 {
   ZeroFluxNeumannBoundaryCondition<LabelImageType> spNBC;
 
-  typename NeighborhoodIteratorType::RadiusType radius;
-  radius.Fill(1);
+  auto radius = MakeFilled<typename NeighborhoodIteratorType::RadiusType>(1);
```
hjmjohnson, line 856: Should these now be "const auto" ?  

I'm approving, and think this PR should not be held up to add const, but it might be another benefit of using MakeFilled.

hjmjohnson (approved):

Modules/Segmentation/LevelSetsv4/include/itkUpdateWhitakerSparseLevelSet.hxx
N-Dekker: > Should these now be "const auto" ?

Good question @hjmjohnson Actually maybe even "constexpr auto"!!! However, it's a bit risky to just blindly add const or constexpr to all such cases, as those variables might need to be modified later on.  So I'm hoping that this can be achieved my a follow-up PR, using some intelligent code analysis tool, which checks that those variables aren't modified after their initialization 😃  

Thanks for your approval!

PR #4932, opened 2024-11-05
STYLE: Replace `T var; var = x` with `T var = x` for scalars in tests

N-Dekker:
Following C++ Core Guidelines, Oct 3, 2024, "Always initialize an object", https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-always

Using Notepad++, Replace in Files, doing:

    Find what: ^( [ ]+)([a-z][a-z ]*[a-z])([ ]+)(\w+);[\r\n]+\1\4\ =
    Replace with: $1$2$3$4 =
    Filters: itk*Test*.*
    [v] Match case
    (*) Regular expression

PR #4939, opened 2024-11-07
Small style improvements `MINCImageIO`, `FileOutputWindow`, `QuadEdgeMesh`

N-Dekker:
I guess these three commits are rather trivial. They just address some little cases that I noticed when doing larger pull requests, specifically:

- #4932
- #4937


Modules/Core/Common/src/itkFileOutputWindow.cxx
```diff
@@ -33,7 +33,6 @@ FileOutputWindow::FileOutputWindow()
 FileOutputWindow::~FileOutputWindow()
 {
   delete m_Stream;
-  m_Stream = nullptr;
```
dzenanz, line 36: This sets the pointer to be null, to indicate that the pointer is no longer valid. It does not attempt to access the object which was pointed at. If I remember correctly, this is a well established pattern.

Modules/IO/MINC/src/itkMINCImageIO.cxx
vfonov, line 1: I was trying to make sign flip clearly visible.

Modules/Core/Common/src/itkFileOutputWindow.cxx
N-Dekker: @dzenanz It's an old discussion, actually... [Is it worth setting pointers to NULL in a destructor?](https://stackoverflow.com/questions/3060006/is-it-worth-setting-pointers-to-null-in-a-destructor)

When the destructor is finished, the FileOutputWindow object is gone, and so is its m_Stream data member 🤷

blowekamp: But why are we not using a smart pointer? 🐱

Modules/IO/MINC/src/itkMINCImageIO.cxx
N-Dekker: OK, thanks @vfonov Of course I can revert my proposed adjustment, if you want to. But still... what would you think of `_sep *= -1.0` (instead of your original `_sep = -_sep`)?

Modules/Core/Common/src/itkFileOutputWindow.cxx
N-Dekker: Thanks @blowekamp  A smart pointer would been much better. It's just because m_Stream isn't private (it's protected) that I did not do so immediately. If you're very strict, you _might_ say that changing a `protected` data member is an API change. (On the other hand, the _P_ in _API_ stands for _Public_, not _Protected_ 😸.)

dzenanz: Doesn't API=Application Programming Interface?

Modules/IO/MINC/src/itkMINCImageIO.cxx
vfonov: I don't mind

hjmjohnson (approved):
This is good to me with or without the sign flip clarification change.

dzenanz (approved):

Modules/IO/MINC/src/itkMINCImageIO.cxx
```diff
@@ -1136,9 +1136,8 @@ MINCImageIO::WriteImageInformation()
             miget_dimension_size(m_MINCPImpl->m_MincApparentDims[j], &_sz);
 
             _start = _start + (_sz - 1) * _sep;
-            _sep = -_sep;
 
-            miset_dimension_separation(m_MINCPImpl->m_MincApparentDims[j], _sep);
+            miset_dimension_separation(m_MINCPImpl->m_MincApparentDims[j], -_sep);
```
N-Dekker, line 1140: Just to clarify: Of course, this is just a very minor improvement to me. I prefer not to unnecessarily modify variables (if possible), just like I prefer const over non-const variables. So I do prefer not to modify `_sep` at this point, and to pass `-_sep` as argument. But the _actual_ reason why this particular case bothered me was because it fooled me when I was searching for lines of code of the form `T * p; p = x` (pull request #4937). The regular expression I used (`^( [ ]+)([^ ].*\*)([ ]+)(\w+);[\r\n]+\1\4\ = `) accidentally matched with those two lines of code:

     _start = _start + (_sz - 1) * _sep;
     _sep = -_sep;

So it treated the first line of code as a declaration of a pointer, named `_sep`! Which was silly, of course! But it bothered me, nevertheless!

blowekamp: Perhaps adding a comment "to make sign flip clearly visible" would be appropriate?

N-Dekker: @blowekamp Do you mean the following?
```cpp
_start = _start + (_sz - 1) * _sep;
// to make sign flip clearly visible
_sep = -_sep;
```

That line of comment does keep my regular expression for `T * p; p = x` away for a little while. But then later, someone may think of a clever regular expression to bypass the comment. Which is not unreasonable to me:

```cpp
int * ptr;
// Let's make the pointer null!
ptr = nullptr;
``` 
Would ideally be rewritten to:
```cpp
// Let's make the pointer null!
int * ptr = nullptr;
```

Anyway, the main question is, when `_sep` is a variable, is the meaning of passing `-_sep` as argument clear enough to anyone? I think it is. 🤷

vfonov: this code flips sign of the step , it needs to adjust the start coordinate and then do actual sign and then store it in the internal variables.

N-Dekker: Thanks @vfonov With the change I'm proposing here, the local `_sep` variable simply keeps original value, as retrieved by calling the _get_ function `miget_dimension_separation`. We could rename the variable to `originalSeparator` to make that clearer. And then pass `-originalSeparator` to the _set_ function `miset_dimension_separation`:

    miset_dimension_separation(m_MINCPImpl->m_MincApparentDims[j], -originalSeparator);

 Would that be clear enough to you?

N-Dekker: Of course, if you are opposed to this style change, feel free to press "Request changes". No problem.

N-Dekker:
Converted back to draft, because there appears no consensus on my proposed removal of `_sep = -_sep`. Specifically, Vladimir does not seem to like it. So I'll reconsider that code change.

Modules/IO/MINC/src/itkMINCImageIO.cxx
N-Dekker: @hjmjohnson In this particular case I think I just give up. The removal of `_sep = -_sep` would only be a small code improvement (in my opinion) anyway.

More generally speaking, I would prefer to only add an extra variable when the variable is used _multiple times_ afterwards. A variable that is only used once _usually_ doesn't appeal very much to me. Sometimes a well-named variable can slightly improve readability, but it also adds up to the number of lines of code, which might then become more complicated (depending on the scope of all those variables). But it's partially just a matter of taste 🤷 

Thanks for asking, Hans! 😃

N-Dekker:
Update: Just dropped the commit _STYLE: Remove `_sep = -_sep` from `MINCImageIO::WriteImageInformation()`_, in order to ease getting this pull request processed.

dzenanz (approved):

PR #4943, opened 2024-11-10
STYLE: Use `std::abs`, instead of doing `if (x < 0) x = -x` manually

N-Dekker:
Aims to make the code clearer. Cases found by regular expressions like ` (\w+) = -\1;` and ` \w+ \*= -1;`

Removed old comments from inside two `DistanceToLine` member functions, saying that they are "trying to avoid an expensive fabs". `std::abs` isn't that expensive anymore nowadays.

### DIFF
--- a/Modules/Core/Common/include/itkMath.h
+++ b/Modules/Core/Common/include/itkMath.h
@@ -329,11 +329,7 @@ FloatAlmostEqual(T                                        x1,
     return false;
   }
 
-  typename Detail::FloatIEEE<T>::IntType ulps = FloatDifferenceULP(x1, x2);
-  if (ulps < 0)
-  {
-    ulps = -ulps;
-  }
+  typename Detail::FloatIEEE<T>::IntType ulps = std::abs(FloatDifferenceULP(x1, x2));
   return ulps <= maxUlps;
 }
 
--- a/Modules/Core/Common/include/itkTriangleCell.hxx
+++ b/Modules/Core/Common/include/itkTriangleCell.hxx
@@ -20,6 +20,7 @@
 #include "vnl/algo/vnl_determinant.h"
 
 #include <algorithm> // For copy_n.
+#include <cmath>     // For abs.
 
 namespace itk
 {
@@ -237,12 +238,7 @@ TriangleCell<TCellInterface>::DistanceToLine(PointType   x,
     denom += static_cast<double>(v21[i] * v21[i]);
   }
 
-  // trying to avoid an expensive fabs
-  double tolerance = 1.e-05 * num;
-  if (tolerance < 0.0)
-  {
-    tolerance = -tolerance;
-  }
+  double tolerance = std::abs(1.e-05 * num);
   if ((-tolerance < denom) && (denom < tolerance)) // numerically bad!
   {
     closestPoint = p1; // arbitrary, point is (numerically) far away
--- a/Modules/Core/TestKernel/include/itkTestingComparisonImageFilter.hxx
+++ b/Modules/Core/TestKernel/include/itkTestingComparisonImageFilter.hxx
@@ -24,6 +24,8 @@
 #include "itkNeighborhoodAlgorithm.h"
 #include "itkProgressReporter.h"
 
+#include <cmath> // For abs.
+
 namespace itk
 {
 namespace Testing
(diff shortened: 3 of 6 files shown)


Modules/Core/TestKernel/include/itkTestingComparisonImageFilter.hxx
```diff
@@ -165,12 +167,8 @@ ComparisonImageFilter<TInputImage, TOutputImage>::ThreadedGenerateData(const Out
...
 
         //  Assume a good match - so test center pixel first, for speed
-        RealType difference = static_cast<RealType>(t) - test.GetCenterPixel();
-        if (NumericTraits<RealType>::IsNegative(difference))
-        {
-          difference = -difference;
-        }
-        auto minimumDifference = static_cast<OutputPixelType>(difference);
+        RealType difference = std::abs(static_cast<RealType>(t) - test.GetCenterPixel());
+        auto     minimumDifference = static_cast<OutputPixelType>(difference);
```
N-Dekker, line 171: @thewtex Regarding our discussion at [discourse](https://discourse.itk.org/t/how-to-assert-that-images-are-equal-in-python-unit-test/7302) So yes, `ComparisonImageFilter` _does_ take the absolute value! (This pull request just aims to make that clearer. It does not change the behavior, as far as I can see.)

dzenanz (approved):

Modules/Core/Common/include/itkTriangleCell.hxx
```diff
@@ -237,12 +238,7 @@ TriangleCell<TCellInterface>::DistanceToLine(PointType   x,
     denom += static_cast<double>(v21[i] * v21[i]);
   }
 
-  // trying to avoid an expensive fabs
-  double tolerance = 1.e-05 * num;
-  if (tolerance < 0.0)
-  {
-    tolerance = -tolerance;
-  }
+  double tolerance = std::abs(1.e-05 * num);
```
N-Dekker, line 241: The comment "trying to avoid an expensive fabs" appears to be introduced by commit cc69c19e6094ce426c9abdcf0a9756b650b11c93 "ENH: Added EvaluatePosition()...", Julien Jomier (@jjomier), Nov 16, 2004. `fabs` might have been expensive twenty years ago. But nowadays, it appears to be the other way around! When compiler optimization is enabled, `fabs` appears to be compiled to just a single machine instruction! While the manual approach (`if (x < 0)...`) appears to yield quite a few more machine instructions! Example at https://godbolt.org/z/M8e6x8ssh

```cpp
double Call_fabs(double x)
{
    return fabs(x);
}

double Call_std_abs(double x)
{
    return std::abs(x);
}

double ManuallyEstimateAbs(double x)
{
    if (x < 0)
    {
        return -x;
    }
    return x;
}
```

Output:
```asm
Call_fabs(double):
        andpd   xmm0, XMMWORD PTR .LC0[rip]
        ret
Call_std_abs(double):
        andpd   xmm0, XMMWORD PTR .LC0[rip]
        ret
ManuallyEstimateAbs(double):
        pxor    xmm2, xmm2
        movapd  xmm3, xmm0
        movapd  xmm1, xmm0
        xorpd   xmm1, XMMWORD PTR .LC2[rip]
        cmpltsd xmm0, xmm2
        andpd   xmm1, xmm0
        andnpd  xmm0, xmm3
        orpd    xmm0, xmm1
        ret
.LC0:
        .long   -1
        .long   2147483647
        .long   0
        .long   0
.LC2:
        .long   0
        .long   -2147483648
        .long   0
        .long   0
```

dzenanz (approved):

hjmjohnson (approved):

Modules/Core/Common/include/itkMath.h
```diff
@@ -329,11 +329,7 @@ FloatAlmostEqual(T                                        x1,
     return false;
   }
 
-  typename Detail::FloatIEEE<T>::IntType ulps = FloatDifferenceULP(x1, x2);
```
thewtex, line 332: Are you sure that this has the same behavior, in particular with `-0` and `0`?

Modules/Core/TestKernel/include/itkTestingComparisonImageFilter.hxx
thewtex: @N-Dekker great!

Modules/Core/Common/include/itkMath.h
N-Dekker: Thanks for bringing 0 versus -0 to the attention, Matt. In this particular case, `FloatDifferenceULP` returns an integer type. For integer types 0 and -0 are bit-wise equal, in C++. So behavior change, in this particular case.

thewtex (approved):
:100:

PR #4952, opened 2024-11-15
Replace `T var; var = x` with `T var = x` for `Region`, `Index`, and `Size` variables

N-Dekker:
- Follow-up to pull request #4941


dzenanz (approved):

thewtex (approved):
:100:

PR #4954, opened 2024-11-15
STYLE: Minor style and readability changes

hjmjohnson:
Added explicit [[fallthrough]] annotations to make code easier to understand.

Added space in diagnostic messages to make errors easier to read.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkImageToImageFilter.hxx
+++ b/Modules/Core/Common/include/itkImageToImageFilter.hxx
@@ -191,7 +191,7 @@ ImageToImageFilter<TInputImage, TOutputImage>::VerifyInputInformation() const
         {
           originString.setf(std::ios::scientific);
           originString.precision(7);
-          originString << "InputImage Origin: " << inputPtr1->GetOrigin() << ", InputImage" << it.GetName()
+          originString << "InputImage Origin: " << inputPtr1->GetOrigin() << ", InputImage " << it.GetName()
                        << " Origin: " << inputPtrN->GetOrigin() << std::endl;
           originString << "\tTolerance: " << coordinateTol << std::endl;
         }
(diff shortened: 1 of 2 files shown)


Modules/Core/Common/include/itkImageToImageFilter.hxx
```diff
@@ -208,7 +208,7 @@ ImageToImageFilter<TInputImage, TOutputImage>::VerifyInputInformation() const
         {
           directionString.setf(std::ios::scientific);
           directionString.precision(7);
-          directionString << "InputImage Direction: " << inputPtr1->GetDirection() << ", InputImage" << it.GetName()
+          directionString << "InputImage Direction: " << inputPtr1->GetDirection() << ", InputImage " << it.GetName()
```
N-Dekker, line 211: For the sake of completeness, you could make similar changes to `ImageSink<TInputImage>::VerifyInputInformation()`, if you want to 🤷

Modules/IO/GDCM/src/itkGDCMImageIO.cxx
```diff
@@ -635,12 +635,19 @@ GDCMImageIO::InternalReadImageInformation()
   switch (ms)
   {
     case gdcm::MediaStorage::HardcopyGrayscaleImageStorage:
+      [[fallthrough]];
     case gdcm::MediaStorage::GEPrivate3DModelStorage:
+      [[fallthrough]];
     case gdcm::MediaStorage::Philips3D:
+      [[fallthrough]];
```
N-Dekker, line 642: @hjmjohnson Just wondering, did the original code cause any warning?

dzenanz (approved):
Good even as-is.

Modules/IO/GDCM/src/itkGDCMImageIO.cxx
N-Dekker: The C++ core guideline is to use `[[fallthrough]]` only for non-empty cases: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es78-dont-rely-on-implicit-fallthrough-in-switch-statements

It says:

> Multiple case labels of a single statement is OK:

```cpp
switch (x) {
case 'a':
case 'b':
case 'f':
    do_something(x);
    break;
}
```

N-Dekker (approved):
Thank you very much for addressing my comments, Hans! 👍 👍

dzenanz (approved):

Modules/IO/GDCM/src/itkGDCMImageIO.cxx
seanm: Yeah, I wouldn't put any `[[fallthrough]]` on empty cases either.

I don't think the old way would have had any warnings, at least not from clang's `-Wimplicit-fallthrough`.

I just noticed I don't have that flag enabled on my bots... I will check if there are none and enable it...

PR #4955, opened 2024-11-15
ENH: Fix dicom reading slices

hjmjohnson:
Dicom files that do not have explicit "SliceThickness" listed can still be used.  The actual "spacing between slices" is computed from the stack of slices when multiple dicoms are provided such that the initial setting of the "SliceThickness" placeholder is not used.  (NOTE: SliceThickness is used when a single slice is read into a 3D image).

This fixes an issue when a stack of 124 dicom slices does not have "SliceThickness" fields, but does have "ImagePositionPatient" for each slice. 

The previous implementation would throw an exception prior to getting to the code that explicitly computes the slice space from the "ImagePositionPatient" values.

The new code now allows ITK to make 3D nifti volumes that are compatible with those created by the dcm2niix package.

## PR Checklist

### DIFF
--- a/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
+++ b/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
@@ -57,6 +57,7 @@
 #include "gdcmDirectionCosines.h"
 
 #include <fstream>
+#include <itkImageBase.h>
 #include <sstream>
 
 namespace itk
@@ -681,7 +682,12 @@ GDCMImageIO::InternalReadImageInformation()
       const double * sp = image.GetSpacing();
       spacing[0] = sp[0];
       spacing[1] = sp[1];
-      spacing[2] = sp[2];
+      // A 2D dicom slice may not have an explicit interslice provided per
+      // file.  interslice distances can be computed after all slices are read.
+      // set to non-zero value here to avoid prematurely throwing an exception
+      // before the interslice thickness can be computed.
+      const auto abs_spacing_2 = std::abs(sp[2]); // Spacing may be negative at this point, will be fixed below
+      spacing[2] = (abs_spacing_2 < itk::DefaultImageCoordinateTolerance) ? 1.0 : sp[2];
     }
     break;
   }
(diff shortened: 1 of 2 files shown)


hjmjohnson:
Local mac testing:
```txt
100% tests passed, 0 tests failed out of 3011
```

dzenanz (approved):

thewtex (approved):
The Windows libtiff errors seem to be unrelated:

```
itktiff-6.0.lib(tif_luv.c.obj) : error LNK2005: __ucrt_int_to_float already defined in itktiff-6.0.lib(tif_aux.c.obj)
itktiff-6.0.lib(tif_color.c.obj) : error LNK2005: __ucrt_int_to_float already defined in itktiff-6.0.lib(tif_aux.c.obj)
bin\ITKAnisotropicSmoothingTestDriver.exe : fatal error LNK1169: one or more multiply defined symbols found
```

jhlegarreta (commented):
Maybe the comment style can be improved ?

Not sure if we should say inter-slice or interslice.

I know this is hard, but can a test be added that demonstrates the enhancement ?

hjmjohnson:
@jhlegarreta  I appreciate your comments, but I don't have the bandwidth to make this perfect.  I'm going to add it as is.

seanm:
I just tracked down 916a40cb3cebf817f91b69ef80a428b346c46fef as the cause of one of our app's unit test failures.

With the new less-strict tolerance check, `ITK_non_uniform_sampling_deviation` is no longer added to the metadata dictionary, and our test was expecting it to be there.

We use the value of `ITK_non_uniform_sampling_deviation` to decide for ourselves if it is above/below an acceptable tolerance. But 916a40cb3cebf817f91b69ef80a428b346c46fef has now hardcoded a tolerance, making it impossible for us to choose ourselves.

It's not clear to me from the commit message what bug this commit was fixing... Yes, `AlmostEquals()` will return false for "slices that are only 1e-15 different for 1.5mm slices", but so what?  Then `ITK_non_uniform_sampling_deviation` will contain a tiny number and callers can decide if they care. They can also just ignore `ITK_non_uniform_sampling_deviation`.

Where does this magic 1e-6 (`DefaultImageCoordinateTolerance`) number come from? What are its units?  millimetres? metres? inches? How can we assume this number is appropriate for every kind of DICOM file under the sun?

PR #4961, opened 2024-11-19
Update libtiff

blowekamp:
## PR Checklist


Modules/ThirdParty/TIFF/UpdateFromUpstream.sh
```diff
@@ -8,26 +8,61 @@ readonly name="tiff"
...
 port/dummy.c
 port/getopt.c
 port/libport.h
+port/libport_config.h.cmake.in
 "
 
 extract_source () {
     git_archive
     pushd "${extractdir}/${name}-reduced"
+    src/itktiff/libtiff/CMakeLists.txt
```
dzenanz: What is this line supposed to accomplish?

dzenanz (approved):
Generally looks good. Do you plan to do an update to 4.7.0 now, or is that left for later?

blowekamp:
> Generally looks good. Do you plan to do an update to 4.7.0 now, or is that left for later?

I think it will be done in this PR.

VTK's 4.6.0 changes were used and were committed in the same sub-tree as upstream. However this changes get clobbered when the updating to 4.7.0.  I believe that the merge of the VTK changes need to be "flattened" so that they are in the main ITK tree.

@bradking How do you suggest fixing the history so that updating to 4.7.0 will not remove the VTK changes?

bradking:
I don't think we should import into ITK from VTK's third-party repo.  I'll rebuild this to make the net changes directly.

blowekamp:
> I don't think we should import into ITK from VTK's third-party repo. I'll rebuild this to make the net changes directly.

I did a "git cherry-pick -m 1 [7408285](https://github.com/InsightSoftwareConsortium/ITK/pull/4961/commits/74082851256f83e450cccae72a716324d9f87680)" and that seemed to "flatten the merge into a single parent commit.

bradking:
@blowekamp I rebuilt your PR to update directly from upstream tiff 4.7.0 and apply ITK-specific build system edits.

bradking:
@blowekamp if there are any problems in CI, you can just edit the last commit or add new commits.

This needs to be merged with a real merge commit and no rebase or squash.  Then future updates with `Modules/ThirdParty/TIFF/UpdateFromUpstream.sh` should merge upstream changes into our local edits.

blowekamp:
> @blowekamp if there are any problems in CI, you can just edit the last commit or add new commits.
> 
> This needs to be merged with a real merge commit and no rebase or squash. Then future updates with `Modules/ThirdParty/TIFF/UpdateFromUpstream.sh` should merge upstream changes into our local edits.

Amazing work!

How was the name mangling generated? Does it still need to be run on other systems?

Also It looks like libtiff.def needs to be converted to a "cmake.in" to support `@MANGLE_PREFIX@`?

Also the libtiff.map file has the unmagled names but it looks like the file is not used and can be removed.

bradking:
`itk_tiff_mangle.h.in` has comments at the top (left from long ago) that describe how to generate the mangling.  We don't actually replace `@MANGLE_PREFIX@` with anything besides `itk_tiff`, but that pattern is used throughout ITK's third-party library mangling so I kept it.  For ` libtiff.def` I think we can leave it hard-coded.

I'll rebuild the branch with `libtiff.map` not imported, but we might as well let CI finish first to see if there are any other problems.

blowekamp:
> `itk_tiff_mangle.h.in` has comments at the top (left from long ago) that describe how to generate the mangling. 

Some have been adamant that these command must be run on all platforms to ensure all symbols have been identified.  So what platform(s) was it run on?

>We don't actually replace `@MANGLE_PREFIX@` with anything besides `itk_tiff`, but that pattern is used throughout ITK's third-party library mangling so I kept it. For ` libtiff.def` I think we can leave it hard-coded.

This is the commit that introduced it:
 a3afcf6fcd6f2a02cbaf928114a2d00b64e63149

My understanding is that 3D Slicer has multiple version of ITK in the runtime. ( ITK Python for pip and local build?) and there were problems with conflicting symbols. Additionally, missing mangled symbols with third partly libraries have caused problems.

bradking:
Okay, I pushed an update to drop the `.map` file and convert the `.def` file to use `@MANGLE_PREFIX@`.

The mangling currently accounts for Windows, macOS, and Linux.

seanm:
> Some have been adamant that these command must be run on all platforms to ensure all symbols have been identified.

I've seen this in the past, with HDF5 I think, which both VTK and ITK use, where the mangling was done only with some OSes and some symbols where missed, and then both VTK and ITK have the same symbol and you get a link error.

95% of symbols are of course shared between OSes, but that 5% has bitten us in the past.

blowekamp:
@bradking Thanks for all the work!

I checked the symbols an my Mac ARM and nothing were missing. I did have to modify the commands to make the symbols unique and sort them with case insensitivity ( but there were still order difference). I also build from an installed version of ITK successfully too.

Modules/ThirdParty/TIFF/src/itktiff/CMakeLists.txt
```diff
@@ -1 +1,254 @@
...
+
+# Standard installation paths
+include(GNUInstallDirs)
+# Documentation install directory (default to cmake project docdir)
+set(LIBTIFF_DOCDIR "${CMAKE_INSTALL_DOCDIR}")
+
+if(FALSE) # XXX(ITK): ITK handles testing
+# CTest testing
+enable_testing()
+endif()
```
N-Dekker, line 108: I guess this `if(FALSE) ... endif()` may be removed, right?

bradking: The idea is to have two new lines that suppress logic in between them, instead of removing blocks of code, in order to make it easier to resolve conflicts when updating from upstream in the future.

blowekamp: You can find the documentation for updating third party libraries here: 
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/docs/contributing/updating_third_party.md
This should give you an overview of the process.

N-Dekker: Thanks @blowekamp Do I understand correctly that two out of the five commits of this PR are scripted? So when reviewing, the focus should mostly be on the three "manual" commits, right? (The ones that are labeled "ENH".)

bradking: The commit titled "tiff: 2024-09-11" is generated by the script as a snapshot of the upstream version.  The merge commit was scripted as a subtree merge, though some conflict resolution was done manually.  Yes, the other three commits are the manual steps.

Modules/ThirdParty/TIFF/src/itktiff/cmake/InternalCodecs.cmake
```diff
@@ -23,6 +23,16 @@
...
+if(TRUE) # XXX(ITK): hard-code settings
+  set(ccitt ON)
+  set(packbits ON)
+  set(lzw ON)
+  set(lze ON)
+  set(thunder ON)
+  set(next ON)
+  set(logluv ON)
+  set(mdi ON)
+endif()
```
N-Dekker, line 35: Just wondering, would it be helpful to move those hard-coded ITK-specific settings to an ITK specific CMake file, for example https://github.com/InsightSoftwareConsortium/ITK/blob/256bb57baf9c7a5691cd229eb808ec97f5dc4109/Modules/ThirdParty/TIFF/src/CMakeLists.txt ?  (To avoid modifying libtiff's "InternalCodecs.cmake".)

Not a request, just a question.

N-Dekker: By the way, I don't see any show-stopper, so if the PR is ready, don't wait for me!

bradking: Possibly, but putting it here keeps it more local w.r.t. the options it is hard-coding.  That will make it easier to update for upstream changes in the future.

blowekamp: @N-Dekker That approach has been done in HDF5:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/ThirdParty/HDF5/src/CMakeLists.txt#L15

I'm not sure how much difference it makes, I don't have much experience with maintaining third-party libraries compared to @bradking.

hjmjohnson (approved):
@bradking @blowekamp  THANK YOU FOR YOUR HARD WORK!!!

Hans

Modules/ThirdParty/TIFF/src/itktiff/cmake/InternalCodecs.cmake
bradking: IIRC I did it that way for HDF5 because we didn't have [CMP0077](https://cmake.org/cmake/help/latest/policy/CMP0077.html) at the time and the options are sprinkled more broadly in the HDF5 source.

Issue #3136, opened 2022-01-22
Replace pthreads and Win32 threads with C++11 threads?


seanm:
I see ITK master uses both pthreads and Win32 threads.

Does this just pre-date C++11 threads?  Or is there some need for platform-specific code here?  i.e. could we switch to C++11 threads?

thewtex:
We do use C++11 threads by default in `master`.

hjmjohnson:
@thewtex I am assuming that the answer to @seanm 's question is "YES" we should purge the system of pthread and Win32 threads in favor of C++11 therads.

Only supporting and exposing one threading mechanism.

dzenanz:
> We do use C++11 threads by default in master.

Yes, ThreadPool uses C++11 threads.

hjmjohnson:
@dzenanz I'm asking for a little more clarification here so that I can determine the scope of the work I am considering.

A) Are Win32 & pthreads already deprecated and replaced by C++11 threads in the default builds on all platforms from on the master branch?  That would imply some tedious but simple housecleaning.  I'm willing to have a @kian-weimer take a stab at cleaning that up.

B) If, however, there are critical path dependencies or backward-compatible special builds that need to be maintained, then that task is probably too large to take on without a longer-term strategy in place.

thewtex:
The implementations are not the default, but they are around for backwards compatibility and testing. Yes, we should purge them, but that should occur in ITK 6.

C++11 threads are the current default for C++ builds.

We also support TBB threading, which is the default for Python package builds in ITK 5.3.

In the future, the best default could be #2831

dzenanz:
@hjmjohnson Purging native threads is a large and finnicky undertaking.

blowekamp:
With the prior v5.2 release there were still some issues and instabilities with the default ITK thread pools which cause SimpleITK to set the older platform multithreads as default. There have been various improvements and fixes to the default pool threader that are coming in v5.3 that needs some to to see if they address the previous issues.

IMHO it is premature to remove the older threading model as the threading pools are still maturing. Introducing a new C++11 option would be another option that would need testing, and time to detect issues before it was mature.

PR #4968, opened 2024-11-20
Preliminary cleanup towards removing pthreads (issue #3136)

seanm:
I was looking again at removing pthreads, and this is some preliminary cleanup.


Modules/Core/Common/include/itkThreadSupport.h
```diff
@@ -22,16 +22,6 @@
 
 #include "itkConfigure.h" // For ITK_USE_WIN32_THREADS, ITK_USE_PTHREADS, etc.
 
-// This implementation uses a routine called SignalObjectAndWait()
-// which is only defined on WinNT 4.0 or greater systems.  We need to
-// define this symbol in order to get the prototype for the
-// routine. This needs to be done before we load any system headers.
-#ifdef ITK_USE_WIN32_THREADS
-#  ifndef _WIN32_WINNT
-#    define _WIN32_WINNT 0x0501 // TBB 4.4 requires WinXP (0x0501 or greater)
```
dzenanz, line 31: I am not sure about removing this. Have you tried compiling with TBB on Windows?

dzenanz (commented):
Mostly looks good.

Modules/Core/Common/include/itkThreadSupport.h
seanm: I haven't, nor do I have Windows to try.

From the comment, it seems obsolete... Also the comment starts: "This implementation uses a routine called SignalObjectAndWait()" but I don't actually find that function anywhere in ITK... 

But I know little about Windows development so I defer to others...

Modules/Core/Common/src/itkPlatformMultiThreaderPosix.cxx
N-Dekker, line 1: Thanks @seanm, cool 👍. Just for my understanding, is the aim to eventually also remove itkPlatformMultiThreaderPosix.cxx ?

Modules/Core/Common/src/itkPlatformMultiThreaderPosix.cxx
```diff
@@ -65,21 +64,15 @@ PlatformMultiThreader::MultipleMethodExecute()
...
-#  ifndef __CYGWIN__
-  pthread_attr_setscope(&attr, PTHREAD_SCOPE_PROCESS);
-#  endif
   for (ThreadIdType thread_loop = 1; thread_loop < m_NumberOfWorkUnits; ++thread_loop)
   {
     m_ThreadInfoArray[thread_loop].UserData = m_MultipleData[thread_loop];
     m_ThreadInfoArray[thread_loop].NumberOfWorkUnits = m_NumberOfWorkUnits;
     int threadError = pthread_create(&(process_id[thread_loop]),
-                                     &attr,
+                                     nullptr,
```
N-Dekker, line 75: Just for my understanding: Is this only a style improvement, passing null instead of a default initialized `attr`? Is it still necessary to do so, given that we want to remove pthreads anyway?

seanm:
> I'll repeat my suggestion from #3136. I would recommend creating a new MultiThreader class with C++11 threads then when mature it replace the pthreads, which could then be marked deprecated.

C++11 threads are a pretty thin wrapper over pthreads, so I don't think much "maturing" will be needed.

Also, `itkPlatformMultiThreaderPosix` *already* uses `std::mutex` and friends.

dzenanz:
The first time around it failed with
```log
ITKWatersheds                       =   5.31 sec*proc (33 tests)
KWStyle                             =   7.64 sec*proc (1 test)
RUNS_LONG                           = 305.41 sec*proc (20 tests)

Total Test time (real) = 502.46 sec
SetCTestConfiguration:BuildDirectory:D:/a/1/s-build
SetCTestConfiguration:SourceDirectory:D:/a/1/s
SetCTestConfigurationFromCMakeVariable:DropMethod:CTEST_DROP_METHOD
SetCTestConfiguration:DropMethod:https
SetCTestConfigurationFromCMakeVariable:DropSite:CTEST_DROP_SITE
SetCTestConfiguration:DropSite:open.cdash.org
SetCTestConfigurationFromCMakeVariable:DropLocation:CTEST_DROP_LOCATION
SetCTestConfiguration:DropLocation:/submit.php?project=Insight
	Add file: D:/a/1/ITK-dashboard/dashboard.cmake
	Add file: D:/a/1/ITK-dashboard/itk_common.cmake
Submit files
   SubmitURL: https://open.cdash.org/submit.php?project=Insight
  Set CURLOPT_SSLVERSION to CURL_SSLVERSION_TLSv1_2
  Set CURLOPT_SSL_VERIFYPEER to on
   Upload file: D:/a/1/s-build/Testing/20241120-2338/Update.xml to https://open.cdash.org/submit.php?project=Insight&FileName=Azure.fv-az688-191___Windows_NT-Build12226-PR4968-issue3136___20241120-2338-Experimental___XML___Update.xml&build=Windows_NT-Build12226-PR4968-issue3136&site=Azure.fv-az688-191&stamp=20241120-2338-Experimental&MD5=b28ed7ae55dfc823ddf29b6193bbc445 Size: 652
   Error when uploading file: D:/a/1/s-build/Testing/20241120-2338/Update.xml
   Error message was: schannel: next InitializeSecurityContext failed: CRYPT_E_REVOCATION_OFFLINE (0x80092013) - The revocation function was unable to check revocation because the revocation server was offline.
   Problems when submitting via HTTP
##[error]Cmd.exe exited with code '-1'.
Finishing: Build and test
```

Modules/Core/Common/include/itkThreadSupport.h
seanm: Does github CI have TBB + Windows?  If not, anyone able to try?

blowekamp: In may be in the pixi builds:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/pixi.lock#L890

dzenanz: This compiles locally, and multi-threading tests pass.

dzenanz: > In may be in the pixi builds: [`master`/pixi.lock#L890](https://github.com/InsightSoftwareConsortium/ITK/blob/master/pixi.lock?rgh-link-date=2024-11-28T13%3A28%3A24Z#L890)

`ITK_USE_TBB` needs to be turned on too.

dzenanz: I don't think TBB is turned on:
https://github.com/InsightSoftwareConsortium/ITK/blob/c9419ca6b98dc6328153f69a04a27b99d4a7d27d/pyproject.toml#L31-L32

PR #4970, opened 2024-11-21
Modernize initialization of local iterators, replacing `IteratorType it; it = <expression>`

N-Dekker:
- Follow-up to pull request #4964


hjmjohnson (approved):
Example update PR would be nice for this too.

PR #4973, opened 2024-11-21
STYLE: Replace `T var; var = x` with `T var = x`

N-Dekker:

    Find what: ^( [ ]+)([^ ][^=\r\n]*)([ ]+)(\w+);[\r\n]+\1\4\ =
    Replace with: $1$2$3$4 =
    Filters: itk*.* !+\test
    Directory: D:\src\ITK\Modules
    [v] Match case
    (*) Regular expression

Excluded a few cases where the syntax `T var; var = x` appears intentional (in ConceptChecking, ImageKmeansModelEstimator, MahalanobisDistanceThresholdImageFunction, QuasiNewtonOptimizerv4Template, and VariableSizeMatrix).

- Following pull request #4970 commit 89b5bd0e24dd1cb6dd268a32094f1b337bf25b39 "STYLE: Replace `Iterator it; it = x` with `Iterator it = x`" and other similar commits for different types of variables.

----

FYI, Just like in previous PRs, this regular expression also encountered some "false positives", for example, it matches the following lines of code, accidentally: 

    delete m_FaceSet;
    m_FaceSet = new IndexSetType();

Which is why I started by creating PRs that just deal with specific types of variables. To ease reviewability, and to avoid such "false positives".

### DIFF
--- a/Modules/Filtering/ImageIntensity/include/itkAdaptImageFilter.h
+++ b/Modules/Filtering/ImageIntensity/include/itkAdaptImageFilter.h
@@ -155,9 +155,7 @@ public:
   void
   SetAccessor(AccessorType & accessor)
   {
-    FunctorType functor;
-
-    functor = this->GetFunctor();
+    FunctorType functor = this->GetFunctor();
     if (accessor != functor.GetAccessor())
     {
       functor.SetAccessor(accessor);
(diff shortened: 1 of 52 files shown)


dzenanz (approved):

Modules/Filtering/ImageIntensity/include/itkAdaptImageFilter.h
```diff
@@ -155,9 +155,7 @@ class AdaptImageFilter
   void
   SetAccessor(AccessorType & accessor)
   {
-    FunctorType functor;
-
-    functor = this->GetFunctor();
+    FunctorType functor = this->GetFunctor();
```
N-Dekker, line 158: For the record, here is where `AdaptImageFilter` will use the copy-constructor of `itk::Functor::AccessorFunctor` (instead of the assignment operator), which was originally deprecated, according to https://open.cdash.org/viewBuildError.php?type=1&buildid=10041267

hjmjohnson (approved):

PR #4976, opened 2024-11-21
Remove `p->Initialize()` calls directly after `p = T::New()`, in tests and example

N-Dekker:
- Follow-up to pull request #4972

### DIFF
--- a/Examples/DataRepresentation/Path/PolyLineParametricPath1.cxx
+++ b/Examples/DataRepresentation/Path/PolyLineParametricPath1.cxx
@@ -86,7 +86,6 @@ main(int argc, char * argv[])
   // Software Guide : BeginCodeSnippet
   ImageType::ConstPointer image = reader->GetOutput();
   auto                    path = PathType::New();
-  path->Initialize();
 
   using ContinuousIndexType = PathType::ContinuousIndexType;
   ContinuousIndexType cindex;


dzenanz (approved):

Modules/Numerics/Statistics/test/itkWeightedCentroidKdTreeGeneratorTest9.cxx
```diff
@@ -39,7 +39,6 @@ itkWeightedCentroidKdTreeGeneratorTest9(int argc, char * argv[])
   using NumberGeneratorType = itk::Statistics::MersenneTwisterRandomVariateGenerator;
 
   auto randomNumberGenerator = NumberGeneratorType::New();
-  randomNumberGenerator->Initialize();
```
N-Dekker, line 42: For most types, `p->Initialize()` has no effect at all, when the object is just returned by `T::New()`. However, `MersenneTwisterRandomVariateGenerator::Initialize()` changes the seed, so it _does_ have an effect. So I'm not sure if it's OK to remove this particular `randomNumberGenerator->Initialize()` call. Or if it matters at all 🤔

dzenanz: If New return object which will have a different seed each time, then this removal is fine. If New always starts with the same seed, then this removal of Initialize is a semantic change.

N-Dekker: I find it difficult to find out exactly what `MersenneTwisterRandomVariateGenerator::New()` does with its seed.  But it looks like the Initialize() call makes things even less predictable. I think I'll revert this removal, as my intention was just to removed "no op" Initialize() calls from the code.

hjmjohnson (approved):

PR #4979, opened 2024-11-21
Add libjpeg turbo

blowekamp:
## PR Checklist


dzenanz (approved):
LGTM. Is the plan to gradually switch to jpeg-turbo exclusively?

Thanks for these bundled library updates. What motivated it?

blowekamp:
> LGTM. Is the plan to gradually switch to jpeg-turbo exclusively?

Certainly having it as a flag give the option to rollback. But I am not sure it's needed. I think this is an easier upgrade that HDF5. 

> Thanks for these bundled library updates. What motivated it?

ITK's third parties need to be kept up to date. I figure it's better to put the effort in updates in ITK rather than creating additional external projects for SimpleITK.

These changes see to work well. I would like a decision if we should have libjpeg-turbo as an option, or if it should just be a replacement. After that I can redo the merge if needed, then get started on the mangling.

dzenanz:
Having it as an option sounds better to me, at least for now. Removing old code should be easy, when needed.

bradking (approved):
The import script and history structure LGTM.

blowekamp:
> Having it as an option sounds better to me, at least for now. Removing old code should be easy, when needed.

yes, I am a little bit on the fence about what the best choice is. the libjpeg-turbo is a fork of IJG's libjpeg, the is a common source code history, and much common code between the two. Replacing the library in the same directory may keep a cleaner history.

dzenanz:
> Having it as an option sounds better to me, at least for now. Removing old code should be easy, when needed.

This is a very mild opinion. I don't usually deal with JPEG images.

seanm:
I don't deal with JPEG (via ITK) either, so don't have a strong opinion. 

Does libjpeg-turbo have different minimum requirements than what's in ITK currently?

thewtex (approved):
:sparkler: thank you @blowekamp !

blowekamp:
> I don't deal with JPEG (via ITK) either, so don't have a strong opinion.

With some work it may be possible to get GDCM to work with this library, and remove the duplicated JPEG library.

> 
> Does libjpeg-turbo have different minimum requirements than what's in ITK currently?

I am not sure which requirements you are referring too.

seanm:
> With some work it may be possible to get GDCM to work with this library, and remove the duplicated JPEG library.

Indeed GDCM's included libjpeg is old, which I raised here:

https://sourceforge.net/p/gdcm/bugs/546/

@malaterre not sure if you ever saw the above ticket?  I could possibly update gdcm's libjpeg, but we'd need to decide which...

blowekamp:
I took a hack at it locally, and I think it's possible.

The old libjpeg needed some patches for lossless compression which I believe libjpeg-turbo supports.

PR #4977, opened 2024-11-21
Prefer local initialization to assignment

hjmjohnson:
Prefer initialization to assignment.  Make variables scoped more locally.

## PR Checklist
- [x] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [x] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)

PR #4981, opened 2024-11-22
STYLE: Remove extra processing return quickly when result is known

hjmjohnson:
Simplify the code by returning as soon as the condition is known to be false.
Remove unnecessary "ok" temporary variable.

## PR Checklist

### DIFF
--- a/Modules/Segmentation/Voronoi/include/itkVoronoiSegmentationRGBImageFilter.hxx
+++ b/Modules/Segmentation/Voronoi/include/itkVoronoiSegmentationRGBImageFilter.hxx
@@ -177,32 +177,22 @@ VoronoiSegmentationRGBImageFilter<TInputImage, TOutputImage>::TestHomogeneity(In
     }
   }
 
-  bool ok = true;
   j = 0;
-  double savem, savev;
-  while (ok && (j < 3))
+  while ((j < 3))
   {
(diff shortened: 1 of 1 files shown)


Modules/Segmentation/Voronoi/include/itkVoronoiSegmentationRGBImageFilter.hxx
```diff
@@ -177,32 +177,22 @@ VoronoiSegmentationRGBImageFilter<TInputImage, TOutputImage>::TestHomogeneity(In
     }
   }
 
-  bool ok = true;
   j = 0;
-  double savem, savev;
-  while (ok && (j < 3))
+  while ((j < 3))
```
N-Dekker, line 181: Thanks for making this a separate PR 👍  The extra pair of parentheses in `while ((j < 3))` still looks a bit funny to me, but no problem. As I wrote before: https://github.com/InsightSoftwareConsortium/ITK/pull/4977#discussion_r1852876529

N-Dekker (approved):
Approved (👍), but please consider a more specific commit message header, for example:

    STYLE: Return quickly when result is known, in Voronoi TestHomogeneity

Also, you may of course remove those extra parentheses, if you agree.

PR #4985, opened 2024-11-24
PERF: readability container size empty

hjmjohnson:
The emptiness of a container should be checked using the empty() method instead of the size() method. It is not guaranteed that size() is a constant-time function, and it is generally more efficient and also shows clearer intent to use empty(). Furthermore some containers may implement the empty() method but not implement the size() method. Using empty() whenever possible makes it easier to switch to another container in the future.

cd ${BLDDIR}
run-clang-tidy.py -extra-arg=-D__clang__ -checks=-*,readability-container-size-empty  -header-filter=.* -fix

### DIFF
--- a/Modules/Filtering/LabelMap/test/itkBinaryReconstructionLabelMapFilterTest.cxx
+++ b/Modules/Filtering/LabelMap/test/itkBinaryReconstructionLabelMapFilterTest.cxx
@@ -97,7 +97,7 @@ itkBinaryReconstructionLabelMapFilterTest(int argc, char * argv[])
                       obtainedAttributeSet.end(),
                       std::inserter(diff, diff.begin()));
 
-  if (diff.size() != 0)
+  if (!diff.empty())
   {
     std::cerr << "Error" << std::endl;
     std::cerr << " Obtained attribute set differs from expected atribute set" << std::endl;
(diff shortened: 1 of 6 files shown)


N-Dekker:
> It is not guaranteed that size() is a constant-time function

Actually, it _is_. For all `std` container types, `size()` has a constant-time complexity. So I don't expect any PERF improvement from this pull request. It's just a STYLE improvement to me. But I like it anyway  👍

hjmjohnson:
> > It is not guaranteed that size() is a constant-time function
> 
> Actually, it _is_. For all `std` container types, `size()` has a constant-time complexity. So I don't expect any PERF improvement from this pull request. It's just a STYLE improvement to me. But I like it anyway 👍
> 
> Originally, Microsoft implemented `empty()` by calling `size()`, so `empty()` _might_ then even be slower than `size()` 😱 Fortunately I was able to convince them not to do so anymore 😇: [microsoft/STL#1836](https://github.com/microsoft/STL/pull/1836)

FYI:  The message is copied from the clang-tidy documentation page.  I took it on faith that the documentation was correct. i trust your experiences more than the documentation :).

N-Dekker:
> The message is copied from the clang-tidy documentation page. I took it on faith that the documentation was correct. i trust your experiences more than the documentation :).

Ah, thanks Hans, I see:
- https://clang.llvm.org/extra/clang-tidy/checks/readability/container-size-empty.html
- https://github.com/llvm/llvm-project/blob/18abc7e0c5b34e9e7bbe0893a4a5281c0937f7d8/clang-tools-extra/clang-tidy/readability/ContainerSizeEmptyCheck.h#L21-L22

Maybe I should ask them for a clarification 🤔

dzenanz (approved):

dzenanz:
Done. I am not sure how much weight my approving review will carry 😄

PR #4995, opened 2024-11-27
STYLE: Rename `TCoordRep` template parameters to `TCoordinate`

N-Dekker:
Aims to make template parameter names for coordinate types clearer.

Following the ITKSoftwareGuide/CodingStyleGuide guideline on Naming Conventions, saying, "Names are generally spelled out; use of abbreviations is discouraged".

----

I think `CoordRep` was an abbreviation of "coordinate representation". But if I understand correctly, that is a specific mathematical term, for example used in https://en.wikipedia.org/wiki/Coordinate_vector as:

> We can mechanize the above transformation by defining a function ..., called the standard representation of V with respect to B, that takes every vector to its coordinate representation

It seems to me that ITK just used the term "representation" here as an equivalent to _type_. (Please correct me if I'm wrong.) But then it appears redundant for a template parameter named "TCoordRep", as its initial `T` already specifies that it is a type.

### DIFF
--- a/Modules/Core/Common/include/itkBoundingBox.h
+++ b/Modules/Core/Common/include/itkBoundingBox.h
@@ -55,7 +55,7 @@ namespace itk
  * Template parameters for BoundingBox:
  *
  * \tparam TPointIdentifier The type used to access a particular point (i.e., a point's id)
- * \tparam TCoordRep Numerical type with which to represent each coordinate value.
+ * \tparam TCoordinate Numerical type with which to represent each coordinate value.
  * \tparam VPointDimension Geometric dimension of space.
  *
  * \ingroup DataRepresentation
@@ -65,8 +65,8 @@ namespace itk
  */
 template <typename TPointIdentifier = IdentifierType,
           unsigned int VPointDimension = 3,
-          typename TCoordRep = float,
-          typename TPointsContainer = VectorContainer<TPointIdentifier, Point<TCoordRep, VPointDimension>>>
+          typename TCoordinate = float,
+          typename TPointsContainer = VectorContainer<TPointIdentifier, Point<TCoordinate, VPointDimension>>>
 class ITK_TEMPLATE_EXPORT BoundingBox : public Object
 {
 public:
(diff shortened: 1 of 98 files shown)


Modules/Core/Common/include/itkBoundingBox.h
```diff
@@ -89,7 +89,7 @@ class ITK_TEMPLATE_EXPORT BoundingBox : public Object
 
   /** Hold on to the type information specified by the template parameters. */
   using PointIdentifier = TPointIdentifier;
-  using CoordRepType = TCoordRep;
+  using CoordRepType = TCoordinate;
```
dzenanz, line 92: This is a bit funny, but as it is part of the public interface it is better not to change it. Too many headaches for little gain.

dzenanz (approved):

Modules/Core/Common/include/itkBoundingBox.h
N-Dekker: @dzenanz Indeed 🤔  I'm considering a follow-up PR that would add lines like `using CoordinateType = TCoordRep`. But I understand that the old `CoordRepType` aliases must still be preserved for quite a while.

hjmjohnson (approved):

jhlegarreta:
I know I am late to this and this is maybe better for a separate PR for whoever is willing to take it: what is the difference between `TCoordinateType` and `TCoordinate` : i.e. `template <typename TCoordinateType>` and `template <typename TCoordinate>`?:

https://github.com/InsightSoftwareConsortium/ITK/pull/4995/files#diff-536756753812270d2e0131d51a69dfbee949d4eb1a6f5b32018fb9746a54dad0R283
https://github.com/InsightSoftwareConsortium/ITK/pull/4995/files#diff-8797f3597c4d35086f34c4ce9ae595dd4673e56b7f3a7eeccfd44ee0e0d0b828R458

 If none, then we should pick the one that sticks best to ITK's policy.

N-Dekker:
Well spotted @jhlegarreta. I believe the initial `T` already indicates that the template parameter is a type. Which is in accordance with our Style Guide: https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/blob/1d04a6042af5dcf8019bcdce2a3e412b870d1e4a/SoftwareGuide/Latex/Appendices/CodingStyleGuide.tex#L1080-L1091

The Style Guide has `TPixel` as an example for pixel type. So `TCoordinate` should also be fine.

There might be incidental case where "Type" is added to the parameter name. That might be by mistake. Or it might be to avoid a naming conflict. I did not check  🤷

PR #4996, opened 2024-11-27
STYLE: One declaration per line for readability

hjmjohnson:
Multiple declarations in a single statement reduce readability.

Detect local variable declaration statements and update to have only one statement per declaration.

The clang-tidy readability-isolate-declaration check with fix options automatically identified these changes.

This is a pre-cursor to the larger work to be done in 4991

As a stand-alone commit, this has very little value to the toolkit, but it provides a significant improvement for identifying places where updates need to be considered.   Regular-expression tools work much better on this simplified code layout.

NOTE:  This is unmodified output of the clang-tidy check and fixes.

## PR Checklist
- [x] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [x] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)

PR #4997, opened 2024-11-27
Deprecate `CoordRepType` and replace it with `CoordinateType`

N-Dekker:
"CoordinateType" appears more readable than "CoordRepType".

    Find what: ( [ ]+)using CoordRepType = (\w+);
    Replace with: $1using CoordinateType = $2;\r\n$1using CoordRepType = CoordinateType;
    Filters: itk*.h
    Directory: D:\src\ITK\Modules
    [v] Match case
    (*) Regular expression

- Follow-up to pull request #4995

### DIFF
--- a/Modules/Core/Common/include/itkBoundingBox.h
+++ b/Modules/Core/Common/include/itkBoundingBox.h
@@ -89,7 +89,8 @@ public:
 
   /** Hold on to the type information specified by the template parameters. */
   using PointIdentifier = TPointIdentifier;
-  using CoordRepType = TCoordinate;
+  using CoordinateType = TCoordinate;
+  using CoordRepType = CoordinateType;
   using PointsContainer = TPointsContainer;
   using PointsContainerPointer = typename PointsContainer::Pointer;
   using PointsContainerConstPointer = typename PointsContainer::ConstPointer;
--- a/Modules/Core/Common/include/itkBoundingBox.h
+++ b/Modules/Core/Common/include/itkBoundingBox.h
@@ -95,8 +95,8 @@ public:
   using PointsContainerPointer = typename PointsContainer::Pointer;
   using PointsContainerConstPointer = typename PointsContainer::ConstPointer;
 
-  using PointType = Point<CoordRepType, VPointDimension>;
-  using BoundsArrayType = FixedArray<CoordRepType, VPointDimension * 2>;
+  using PointType = Point<CoordinateType, VPointDimension>;
+  using BoundsArrayType = FixedArray<CoordinateType, VPointDimension * 2>;
 
   /** Hold on to the dimensions specified by the template parameters. */
   static constexpr unsigned int PointDimension = VPointDimension;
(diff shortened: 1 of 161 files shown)


dzenanz (commented):
I would prefer if you introduced deprecation for `CoordRepType` in this PR. Pick legacy or future legacy.

dzenanz (approved):

Modules/Core/Common/include/itkBoundingBox.h
```diff
@@ -90,7 +90,10 @@ class ITK_TEMPLATE_EXPORT BoundingBox : public Object
   /** Hold on to the type information specified by the template parameters. */
   using PointIdentifier = TPointIdentifier;
   using CoordinateType = TCoordinate;
-  using CoordRepType = CoordinateType;
+#ifndef ITK_FUTURE_REMOVE_LEGACY
+  using CoordRepType [[deprecated("ITK 6 discourages using `CoordRepType`. Please use `CoordinateType` instead!")]] =
+    CoordinateType;
+#endif
```
N-Dekker, line 96: I'm not sure but I'm afraid that some users might get annoyed by this deprecated-warning. Especially those users who have intentionally enabled legacy support. What do you think?

N-Dekker: I'm thinking, maybe we should have a FUTURE_DEPRECATED macro, roughly as follows:

```cpp
#ifdef ITK_LEGACY_REMOVE
#define FUTURE_DEPRECATED [[deprecated]]
#else
#define FUTURE_DEPRECATED
#endif
```
To be used as follows:

```cpp
    using CoordRepType FUTURE_DEPRECATED = CoordinateType;
```

So that the warning only appears when users enable LEGACY_REMOVE. What do you think?

dzenanz: There are compiler flags to control warnings.

I now notice: is the correct ifdef `ITK_FUTURE_LEGACY_REMOVE`? You used `ITK_FUTURE_REMOVE_LEGACY`.

dzenanz: That `FUTURE_DEPRECATED` macro sounds good to me.

N-Dekker: On the other hand, I just hope that this PR can be processed quickly, before Hans (@hjmjohnson) submits another mega-PR  😸 His PR #4996 already caused some conflicts locally, when I was preparing this PR.

dzenanz (approved):

jhlegarreta:
I assume this
https://open.cdash.org/viewBuildError.php?onlydeltap&buildid=10058584

is related to this PR.

jhlegarreta:
Haven't had a look/do not have time to investigate now, but my first thought was whether the `Examples` were included in your set of changes.

dzenanz:
@seanm might provide more info about the build configuration.

N-Dekker:
Sorry I overlooked the Examples when I did this PR. I can have another look in a few hours from now.

PR #5010, opened 2024-12-04
COMP: Slow down removal of the `CoordRepType` aliases

N-Dekker:
For now, only remove those aliases when explicitly requested by defining `ITK_REMOVE_COORD_REP_TYPES`. The aim is to put them in `#ifndef ITK_FUTURE_LEGACY_REMOVE` with ITK 6.1.

### DIFF
--- a/Modules/Core/Common/include/itkBoundingBox.h
+++ b/Modules/Core/Common/include/itkBoundingBox.h
@@ -90,7 +90,7 @@ public:
   /** Hold on to the type information specified by the template parameters. */
   using PointIdentifier = TPointIdentifier;
   using CoordinateType = TCoordinate;
-#ifndef ITK_FUTURE_LEGACY_REMOVE
+#ifndef ITK_REMOVE_COORD_REP_TYPES
   using CoordRepType ITK_FUTURE_DEPRECATED(
     "ITK 6 discourages using `CoordRepType`. Please use `CoordinateType` instead!") = CoordinateType;
 #endif
--- a/Modules/Core/Common/include/itkCellInterface.h
+++ b/Modules/Core/Common/include/itkCellInterface.h
@@ -110,7 +110,7 @@ public:
 
   /** Save type information for this cell. */
   using CoordinateType = typename CellTraits::CoordinateType;
-#ifndef ITK_FUTURE_LEGACY_REMOVE
+#ifndef ITK_REMOVE_COORD_REP_TYPES
   using CoordRepType ITK_FUTURE_DEPRECATED(
     "ITK 6 discourages using `CoordRepType`. Please use `CoordinateType` instead!") = CoordinateType;
 #endif
@@ -530,7 +530,7 @@ class ITK_TEMPLATE_EXPORT CellTraitsInfo
 public:
   static constexpr unsigned int PointDimension = VPointDimension;
   using CoordinateType = TCoordinate;
-#ifndef ITK_FUTURE_LEGACY_REMOVE
+#ifndef ITK_REMOVE_COORD_REP_TYPES
   using CoordRepType ITK_FUTURE_DEPRECATED(
     "ITK 6 discourages using `CoordRepType`. Please use `CoordinateType` instead!") = CoordinateType;
 #endif
--- a/Modules/Core/Common/include/itkContinuousIndex.h
+++ b/Modules/Core/Common/include/itkContinuousIndex.h
@@ -57,7 +57,7 @@ public:
    * as a data element held in an Point.   */
   using ValueType = TCoordinate;
   using CoordinateType = TCoordinate;
-#ifndef ITK_FUTURE_LEGACY_REMOVE
+#ifndef ITK_REMOVE_COORD_REP_TYPES
   using CoordRepType ITK_FUTURE_DEPRECATED(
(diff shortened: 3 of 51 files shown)


Modules/Core/Common/include/itkBoundingBox.h
```diff
@@ -90,7 +90,7 @@ class ITK_TEMPLATE_EXPORT BoundingBox : public Object
   /** Hold on to the type information specified by the template parameters. */
   using PointIdentifier = TPointIdentifier;
   using CoordinateType = TCoordinate;
-#ifndef ITK_FUTURE_LEGACY_REMOVE
+#ifndef ITK_REMOVE_COORD_REP_TYPES
```
hjmjohnson, line 93: Can an advanced option be created for turning this on?
```cmake
# The ITK_REMOVE_COORD_REP_TYPES option is a transient flag to assist with API changes
# It is a flag expected to be removed after ITK version 6.0 release.  It is not intended to be part of the public API.
option(ITK_REMOVE_COORD_REP_TYPES OFF)
mark_as_advanced(ITK_REMOVE_COORD_REP_TYPES)
'''

blowekamp: >  Because when users need to support _both_ ITK 6.0 and ITK 5.4 in one and the same project, I would rather not see them "littering" their code with `#if ITK_VERSION_MAJOR < 6 CoordRepType #else CoordinateType `.

I think the mistake is trying to support 5.4 and 6.0 with FUTURE_LEGACY_REMOVE. You should be able to compile ITK 6.0 with LEGACY_REMOVE, and be much closer to support 5.4 too.

N-Dekker: @blowekamp Thanks, but then if you compile user code with ITK 6.0 + `LEGACY_REMOVE=ON`, you might get warnings, when you still use those deprecated `CoordRepType` aliases in your user code. Is that OK? (Again, I'd rather not see users writing `#if ITK_VERSION_MAJOR < 6 CoordRepType #else CoordinateType` in their code.)

blowekamp: Can you compile with LEGACY_SILENT until support for 5 is dropped?

N-Dekker: The `CoordRepType` are currently marked by `ITK_FUTURE_DEPRECATED(message)`, which evaluates to `[[deprecated(message)]]` when `LEGACY_REMOVE` is ON:

https://github.com/InsightSoftwareConsortium/ITK/blob/6a704aaadef717113aa80e7e021c408c086b3677/Modules/Core/Common/include/itkMacro.h#L1378-L1382

N-Dekker: I'm still 100% behind renaming CoordRepType to CoordinateType, I just hope that it doesn't annoy anyone  😇 !

N-Dekker: > Another option would be to back port the new typedef to release-5.4 branch.

Thanks @blowekamp  I think it would indeed be nice if we could cherry-pick these two for the release-5.4 branch:
- fbe9b046b21d9a1353be13512344cdc839fd15b4
- 5ea1a9a8dfc322cc4ae4c18b09136d27c554b31b

These two would just make the new CoordinateType aliases available to ITK 5.4 users, without any deprecate or removal. What do you think?

hjmjohnson:
@N-Dekker Progress comes at a cost.  I am 100% behind the suggestions here, but want to be cautious about expending too much effort in finding the perfect solution for all corner cases.

I like Brad's suggestion of backporting a few new typealias to 5.4 and then being more aggressive in keeping 6.0 clean.

I remember the days of nested #ifdefs littering the toolbox for many different platforms.  It was a nightmare to find corner case bugs.

Modules/Core/Common/include/itkBoundingBox.h
N-Dekker: - Please have a look at PR #5014

dzenanz (changes requested):
I like #5014 a lot better.

N-Dekker:
OK, thanks, so this pull request seems unnecessary, especially now that PR #5014 made it only the release branch 😃 

@dzenanz Is it OK for the main/master branches of remote modules (like ITKMontage) to assume that they support the latest revision of either the release branch or the master branch of ITK? Or should they still support a specific ITK tag?

If we can assume that the new CoordinateType aliases of ITK are available to the main branches of remote modules, I think it's very easy to upgrade them, by globally replacing "CoordRepType" with "CoordinateType" (case sensitive, not "whole word only"), in the remote modules.

dzenanz:
Yes, we usually assume that remote modules support the latest ITK master.

N-Dekker:
But maybe remote modules also define their own `<RemoteModule>::CoordRepType` types, in their public interface...? Then I'm afraid those should still get the same treatment as the ones in ITK 🤷 As in:

https://github.com/InsightSoftwareConsortium/ITK/blob/814a24c84d3bf4bdc42ce8f07276997fabdb45d3/Modules/Core/Common/include/itkPoint.h#L63-L67

dzenanz:
Inheriting classes in remote modules is less common. Plus, you would have to deal with changes in main ITK repo, so you should be aware of the renames. I don't think it is worth the effort.

PR #5016, opened 2024-12-05
Remove "Type" from `TCoordinateType` template parameters, remove initial "T" from `TCoordinateType` alias

N-Dekker:
The postfix `Type` is unnecessary for a template parameter `TCoordinateType`, because its initial `T` already indicates that it is a type.

On the other hand, a type alias (typedef) "typically" has a `Type` postfix, but should _not_ have a `T` initial.

Triggered by a comment from Jon (@jhlegarreta) at https://github.com/InsightSoftwareConsortium/ITK/pull/4995#issuecomment-2504978482


hjmjohnson (approved):
LGTM

jhlegarreta (approved):
Thanks for doing this, Niels.

PR #5020, opened 2024-12-06
Add experimental HTJ2K TransferSyntax for reading to GDCM

blowekamp:
Relates to #3983

The experimental support for HTJ2K has been merged upstream into GDCM. Update ITK to the latest GDCM release tag include this feature.

Adding it to ITK will allow the ITK community to better test the new feature.

## PR Checklist
- [ ] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [ ] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [ ] Added test (or behavior not changed)
- [ ] Updated API documentation (or API not changed)
- [ ] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [ ] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [ ] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples) for all new major features (if any)

Refer to the [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) for
further development details if necessary.

PR #5022, opened 2024-12-06
GDCM use itkopenjpeg

blowekamp:
Configure GDCM to use the ITK's vendored OpenJPEG. The GDCM's version is 2.3.0 while ITK contains 2.5.2.

Also it appears there were some unintentional reuse of CMake cache variable between the two instances, obscuring a missing file used to CMake try compile.

## PR Checklist


dzenanz (approved):
This is a subset of #5020, correct?

blowekamp:
> This is a subset of #5020, correct?

This PR cherry-picked on the one commit which enable GDCM to use itkopenjpeg, and also addresses a bug in itkOpenJPEG build system.

dzenanz (approved):

hjmjohnson (approved):

seanm:
In case it benefits others:

Since my app does not build with CMake, I have to list libraries I link against myself. This change resulted in the file `litkgdcmopenjp2.a` going away and thus causing me link errors. I just have to link against `litkopenjpeg.a` now.

seanm:
Another thing I noticed, probably since this PR: when installing ITK, there is `include/ITK` as before, with zillions of headers, but now there's also `include/itkopenjpeg-2.5`, with only 1 header.  Is that deliberate?  Why not put it in `include/ITK` with the other thousands of files?

seanm:
@blowekamp I'm curious: why didn't you instead (or also) update openjpeg within GDCM itself?

---

This change has caused me another issue... Our internal CI does some builds of our app against nightly versions of all our upstream, including a variation where we build ITK master with GDCM master using `ITK_USE_SYSTEM_GDCM=1` and `GDCM_DIR=...`.  (i.e. a very 'bleeding edge' build).

When we do that, it seems that `litkgdcmopenjp2.a` is still created, and so is `libitkopenjpeg.a`. Is that expected?  (It explains why I didn't notice all this as soon as this PR was merged, as my app was still linking with `litkgdcmopenjp2.a`, and only in the hackathon yesterday when I was building with `ITK_USE_SYSTEM_GDCM=0` did I notice all this).

dzenanz:
If I remember correctly, he did open some PRs to GDCM.

blowekamp:
> When we do that, it seems that `litkgdcmopenjp2.a` is still created, and so is `libitkopenjpeg.a`. Is that expected? (It explains why I didn't notice all this as soon as this PR was merged, as my app was still linking with `litkgdcmopenjp2.a`, and only in the hackathon yesterday when I was building with `ITK_USE_SYSTEM_GDCM=0` did I notice all this).

Yes this is expected, but maybe not the best behavior. ITK still uses ibitkopenjpeg.a for some ITK ImageIO modules. But the upstream GDCM reverts to having the GDCM vendored openjpeg. That is two copies of the same library in your binary!
 Ideally, there should be an upstream GDCM option to use a system OpenJPEG, which could be configured to ITK's OpenJPEG.

> If I remember correctly, he did open some PRs to GDCM.

That is quite related. GDCM upstream has been updated to a current openjpeg version. This PR reconfigured ITK to use ITK's thirdparty openjpeg with GDCM. By avoiding having two mangled openjpeg libraries the binary size should be smaller and maybe more efficient. There are other libraries that we also have duplicates in ITK alone that would be nice to remove too.

I had been thinking about removing the [gdcmopenjpeg directory] (https://github.com/InsightSoftwareConsortium/ITK/tree/master/Modules/ThirdParty/GDCM/src/gdcm/Utilities/gdcmopenjpeg) for the update scrips and repository to remove unused/duplicate source code.

PR #5025, opened 2024-12-09
 STYLE: Prefer explicit const designation

hjmjohnson:
Implements detection of local variables which could be declared as const but
are not. Declaring variables as const is recommended by many coding
guidelines, such as: ES.25 from the C++ Core Guidelines.

cd ${BLDDIR}
run-clang-tidy.py -extra-arg=-D__clang__ -checks=-*,misc-const-correctness -header-filter=.* -fix

## PR Checklist
- [x] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [x] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)

PR #5038, opened 2024-12-11
STYLE: Use range-based loops from C++11

hjmjohnson:
Only 1 commit ahead of #5025 

C++11 Range based for loops can be used in
    
Used as a more readable equivalent to the traditional for loop operating over a
range of values, such as all elements in a container, in the forward direction..

Range based loopes are more explicit for only computing the
end location once for containers.

## PR Checklist

### DIFF
--- a/Modules/Core/ImageFunction/include/itkRayCastInterpolateImageFunction.hxx
+++ b/Modules/Core/ImageFunction/include/itkRayCastInterpolateImageFunction.hxx
@@ -1003,29 +1003,29 @@ RayCastInterpolateImageFunction<TInputImage, TCoordinate>::RayCastHelper::Reset(
 
   else
   {
-    for (int i = 0; i < 3; ++i)
+    for (double & it : m_RayVoxelStartPosition)
     {
-      m_RayVoxelStartPosition[i] = 0.;
+      it = 0.;
     }
-    for (int i = 0; i < 3; ++i)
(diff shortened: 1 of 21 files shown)


Modules/Core/ImageFunction/include/itkRayCastInterpolateImageFunction.hxx
```diff
@@ -1003,29 +1003,29 @@ RayCastInterpolateImageFunction<TInputImage, TCoordinate>::RayCastHelper::Reset(
 
   else
   {
-    for (int i = 0; i < 3; ++i)
+    for (double & i : m_RayVoxelStartPosition)
     {
-      m_RayVoxelStartPosition[i] = 0.;
+      i = 0.;
```
thewtex, line 1008: For these and similar, `i` has a different mean than the typical index integer counter, so it could be called `it`, `iter`,`position`, etc.

thewtex (changes requested):

Modules/Core/ImageFunction/include/itkRayCastInterpolateImageFunction.hxx
hjmjohnson: It was easier and quicker than I anticipated.  Done!

thewtex (approved):
Thanks, Hans!

dzenanz (approved):

PR #5039, opened 2024-12-11
STYLE: Make prototype match definition names

hjmjohnson:
Enforce consistency in large projects, where it often happens that a definition of a function is refactored, changing the parameter names, but its declaration in the header file is not updated. With this check, we can easily find and correct such inconsistencies, keeping declaration and definition always in sync.

Unnamed parameters are allowed and are not taken into account when comparing function declarations.

## PR Checklist


thewtex (approved):
:1st_place_medal:

dzenanz (approved):

seanm:
Did you do this manually? clang-tidy has a check for this IIRC.

PR #5040, opened 2024-12-11
STYLE: Use auto for variable type matches

hjmjohnson:
This check is responsible for using the auto type specifier for variable declarations to improve code readability and maintainability.

The auto type specifier will only be introduced in situations where the variable type matches the type of the initializer expression. In other words auto should deduce the same type that was originally spelled in the source

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkMacro.h
+++ b/Modules/Core/Common/include/itkMacro.h
@@ -343,7 +343,7 @@ namespace itk
 #define itkFactorylessNewMacro(x) \
   static Pointer New()            \
   {                               \
-    x *     rawPtr = new x();     \
+    auto *  rawPtr = new x();     \
     Pointer smartPtr = rawPtr;    \
     rawPtr->UnRegister();         \
     return smartPtr;              \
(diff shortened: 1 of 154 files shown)


Modules/Core/Common/include/itkMacro.h
```diff
@@ -862,7 +862,7 @@ compilers.
 // clang-format off
 /** Set a decorated input. This defines the Set"name"() and a Set"name"Input() method */
 #define itkSetDecoratedInputMacro(name, type)                                                                       \
-  virtual void Set##name##Input(const SimpleDataObjectDecorator<type> * _arg)                                       \
+  virtual void Set## name## Input(const SimpleDataObjectDecorator<type> * _arg)                                       \
```
dzenanz, line 865: Is this space after `##` expected and/or desired?

dzenanz (approved):

Modules/Core/Common/include/itkMacro.h
hjmjohnson: reverted.  I don't know why this got auto-formatted as a change.  clang-format does not seem to care which way it is formatted.

PR #5044, opened 2024-12-11
ENH: Move Discrete Gaussian, Hessian ImageFunction's out of Review

thewtex:
Mature at this point.

These need to go into the ITKImageIntensity module instead of
ITKImageFunction because of their dependency on
itk::RescaleIntensityImageFilter.


dzenanz (approved):

N-Dekker:
Looks like these ImageFunction's have been in Review for more than 13 years already, right? Is there a specific reason to move them out of Review right now?

dzenanz:
We would like to completely eliminate `Review` module, by putting its contents into appropriate other modules. We decided to do it incrementally, as people had time and enthusiasm for it. And this is another step on that path.

N-Dekker:
> We would like to completely eliminate `Review` module, by putting its contents into appropriate other modules. We decided to do it incrementally, as people had time and enthusiasm for it. And this is another step on that path.

Ah, thanks for explaining @dzenanz! So then, should they still be reviewed anyway? Could it be that some of them are obsolete now, and may just be removed?

dzenanz:
The typical behavior was to turn on review module, because it had many frequently used classes. By now, we have moved most of them to proper places, and perhaps removed some. We probably should not remove any more, but part of moving the class to appropriate place is evaluating whether it is needed, and if unclear, cross-link the possibly redundant classes via "See also" in the documentation. Ideally, also add/improve testing when doing the move.

hjmjohnson (approved):

PR #5045, opened 2024-12-11
pointsetbase set get numpy arrays

thewtex:
- STYLE: Use Python in Common python test names
- ENH: Shaped NumPy array input to SetPointsByCoordinates

### DIFF
--- a/Modules/Core/Common/include/itkPointSetBase.h
+++ b/Modules/Core/Common/include/itkPointSetBase.h
@@ -119,6 +119,13 @@ public:
   void
   Initialize() override;
 
+  /** Point dimension. The dimension of a point is fixed at compile-time. */
+  static constexpr unsigned int
+  GetPointDimension()
+  {
+    return PointDimension;
+  }
+
   /** Get the number of points in the points container. */
   PointIdentifier
   GetNumberOfPoints() const;
--- a/Modules/Core/Common/wrapping/test/itkPointSetTest.py
+++ b/Modules/Core/Common/wrapping/test/itkPointSetTest.py
@@ -20,6 +20,7 @@ import itk
 import unittest
 import numpy as np
 
+
 class PointSetTestCase(unittest.TestCase):
     """Tests itk.PointSet"""
 
@@ -32,20 +33,23 @@ class PointSetTestCase(unittest.TestCase):
         # Test the most common dimensions, and a few different numbers as number of points.
         for dimension in [2, 3]:
             for number_of_points in [1, 4, 9]:
+                point_set = itk.PointSet[PixelType, dimension].New()
+
                 # Create a NumPy array of random coordinates between -1.0 and 1.0
                 np_array = 2.0 * np.random.rand(number_of_points * dimension) - 1.0
 
-                point_set = itk.PointSet[PixelType, dimension].New()
+                # Set with flat array
                 point_set.SetPointsByCoordinates(np_array)
 
-                points = point_set.GetPoints()
+                # Set with reshaped array
+                np_array = np.reshape(np_array, (number_of_points, dimension))
+                point_set.SetPointsByCoordinates(np_array)
 
-                self.assertEqual(points.Size(), number_of_points)
+                points = point_set.GetPointsByCoordinates()
 
-                for i in range(number_of_points):
-                    point = points.GetElement(i)
-                    for j in range(dimension):
-                        self.assertAlmostEqual(point[j], np_array[i * dimension + j])
+                self.assertEqual(points.shape[0], number_of_points)
+
+                np.testing.assert_allclose(np_array, points)
 
 
 if __name__ == "__main__":
--- a/Wrapping/Generators/Python/PyBase/pyBase.i
+++ b/Wrapping/Generators/Python/PyBase/pyBase.i
@@ -685,6 +685,32 @@ str = str
   }
 %enddef
 
+%define DECL_PYTHON_POINTSETBASE_CLASS(swig_name)
+    %rename(__SetPointsByCoordinates_orig__) swig_name::SetPointsByCoordinates;
+    %extend swig_name {
+        %pythoncode %{
+            def SetPointsByCoordinates(self, points):
+                """Set the points of the pointset by providing their coordinates
+                as a NumPy array."""
+                import numpy as np
+                if hasattr(points, 'ndim') and points.ndim != 1:
+                    self.__SetPointsByCoordinates_orig__(points.ravel())
+                else:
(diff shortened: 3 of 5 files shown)


dzenanz (approved):
LGTM

hjmjohnson (approved):

Modules/Core/Common/include/itkPointSetBase.h
```diff
@@ -119,6 +119,13 @@ class ITK_TEMPLATE_EXPORT PointSetBase : public DataObject
   void
   Initialize() override;
 
+  /** Point dimension. The dimension of a point is fixed at compile-time. */
+  static constexpr unsigned int
+  GetPointDimension()
+  {
+    return PointDimension;
+  }
+
```
N-Dekker, line 128: This member function seems redundant, as `PointDimension` is already a public member. Are `static constexpr` public data members (like `PointDimension`) not made available already, by SWIG?

Modules/Core/Common/wrapping/test/itkPointSetTest.py
```diff
@@ -32,20 +33,23 @@ def test_SetPointsByCoordinates_from_NumPy_array(self) -> None:
...
-                points = point_set.GetPoints()
+                points = point_set.GetPointsByCoordinates()
 
-                self.assertEqual(points.Size(), number_of_points)
+                self.assertEqual(points.shape[0], number_of_points)
 
-                for i in range(number_of_points):
-                    point = points.GetElement(i)
-                    for j in range(dimension):
-                        self.assertAlmostEqual(point[j], np_array[i * dimension + j])
```
N-Dekker, line 48: I think this is a good test for the flat array case. I think it should not have been removed.

Wrapping/Generators/Python/PyBase/pyBase.i
```diff
@@ -685,6 +685,32 @@ str = str
...
+                as a NumPy array."""
+                import numpy as np
+                if hasattr(points, 'ndim') and points.ndim != 1:
+                    self.__SetPointsByCoordinates_orig__(points.ravel())
+                else:
+                    self.__SetPointsByCoordinates_orig__(points)
+
+            def GetPointsByCoordinates(self):
+                """Get the points of the pointset by providing their coordinates
+                as a NumPy array. The array will have a shape of (n_points, dimension)."""
```
N-Dekker, line 703: I'm not sure about the name "GetPointsByCoordinates". Shouldn't it be something like "GetPointsAsNumpyArray"? Or "converts_points_to_np_array"?

Modules/Core/Common/include/itkPointSetBase.h
thewtex: @N-Dekker correct -- Python only has access to the run-time methods.

Modules/Core/Common/wrapping/test/itkPointSetTest.py
thewtex: We do retain testing for the flat array input. IT has the same output as the shaped array, and it is tested with the more detailed and efficient method.

N-Dekker: Thanks Matt, but in order to properly test for the flat array input, the asserts should be placed _directly_ after the `SetPointsByCoordinates(<flat array>)` call. Not just after the `SetPointsByCoordinates(<shaped array>)` call.

Wrapping/Generators/Python/PyBase/pyBase.i
N-Dekker: OK, but we could then also rename the new `SetPointsByCoordinates`. My original intention was to express that `SetPointsByCoordinates` sets the points by specifying a simple sequence of coordinates. For the newly added functionality, `SetPointsByNumpyArray` (or `set_points_by_np_array`) might be a more appropriate name, right?

Modules/Core/Common/include/itkPointSetBase.h
N-Dekker: Thanks Matt. It's a pity though, if SWIG cannot wrap them automatically. So just to be sure, I asked the question at swig/discussions as well: 

- https://github.com/swig/swig/discussions/3088

thewtex: On a high level, as a runtime language, Python has access to runtime information.

Wrapping/Generators/Python/PyBase/pyBase.i
thewtex: Isn't the API getting unnecessarily complicated?

I thought that you discouraged the use of `SetPoints`?

N-Dekker: > Isn't the API getting unnecessarily complicated?

When two functions do different things, I think it is clearer to also use different names for those functions. The original `SetPoints(PointsContainer *)` (which has been there for many, many years already, including ITK 1, 2, 3, 4, 5, and 6 alpha 1) _shares_ the container between the caller and the `PointSet`. The new `SetPointsByCoordinates(const std::vector<CoordinateType> &)` _copies_ the coordinates from the caller into the container of the `PointSet`.

> I thought that you discouraged the use of `SetPoints`?

The original `SetPoints(PointsContainer *)` is OK to me. It's OK to "share" your container with a point set, if that's what you want. However, the extra `SetPoints(PointsVectorContainer *)` overload that was added with ITK v5.3.0 appears problematic, as I explained at issue #4848 If I understand correctly, the extra `SetPoints(PointsVectorContainer *)` overload was _specifically_ introduced so support flat coordinate arrays. But unfortunately it may lead to undefined behavior. The new `SetPointsByCoordinates(const std::vector<CoordinateType> &)` was just meant to provide a safe alternative to the problematic `SetPoints(PointsVectorContainer *)` overload of ITK >= v5.3.0.

Is that a helpful explanation to you?

thewtex: @N-Dekker I understand what you mean. However, I do not see the need for another method, which over-complicates the API. It is standard practice in object-oriented APIs to have the same method accept two different types.

Modules/Core/Common/include/itkPointSetBase.h
N-Dekker: The reply from William S Fulton (@wsfulton) at https://github.com/swig/swig/discussions/3088#discussioncomment-11550037 suggests that it _is_ possible to directly get the value of a static constexpr data member of a template class, in Python.  (So then the `GetPointDimension()` member function that you added should not be necessary.) However, I haven't yet been able to figure out how to make it work. 🤷 To be continued, hopefully...

PR #5046, opened 2024-12-11
COMP: Limit PR overload of macOS CI

hjmjohnson:
Many PR's are held up due to
```txt
Queued — Waiting to run this check...
```
Sometimes for more than a day.

Maximum concurrent macOS jobs is 5, so limit the number of macOS builds per PR request. https://docs.github.com/en/actions/administering-github-actions/usage-limits-billing-and-administration

https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners macos-13 is an intel build (from the pixi.yml)  -- non python build

REMOVE macos-14 is an arm15 build (from pixi.yml) --redundant with Azure Pipeline build REMOVE macos-14 is an arm14 build (from macos-arm.yml) -- Python build Redundant with Azure Pipeline build REMOVE macos-14-Python is an arm14 build (from macos-arm.yml) -- Python build redundant with Azure Pipeline build

## PR Checklist


hjmjohnson:
MY OPINIONS REPRESENTED BELOW:  I'd like to hear others opinions.

There was a huge back-log of PR's not being evaluated due to the congestion caused by only 5 macOS builds allowed concurrently.

This PR may be too aggressive, but we should identify what needs to be covered and which runners should do the covering.

We had 6 macOS builds for every PR.  
```txt
Azure:
     ITK.macOS              (macos-??, ??)
     ITK.macOS.Python (macos-??, ??)
pixi.yml:
    ITK.Pixi / Pixi-Cxx (macos-13) (pull_request)  (macos-13, intel based)
    ITK.Pixi / Pixi-Cxx (macos-14) (pull_request)  (macos-14, arm based)
macos-arm.yml
     ITK.macOS.Arm64 / macOS-Python (macos-14, arm based)
     ITK.macOS.Arm64 / macOS               (macos-14, arm based)
```
I think we need one arm-based macOS-Python build and one intel-based macOS build.  That should be sufficient for PR coverage, in my opinion.  

If we get to a single GitHub actions Mac build per PR in, that would allow multiple PR's to be evaluated in parallel.

dzenanz (approved):
It is good that you identified the problem. Suggested solution sounds reasonable to me.

blowekamp:
Any consideration made for the time of each build? What is the most time efficient build?

Would it be useful to keep one of these builds just on the master branch?

thewtex:
There has been a lot of activity lately, which is outstanding!

In the longer term, CI will start to catch up with the surges.

I agree that we could remove some Mac builds.

We should prefer Intel Mac builds since they are slower and have been phased out by Apple.

I think we should keep the GitHub Actions Mac ARM build that submits to the dashboard and the Pixi one. The dashboard report is nice. The Pixi one tests the conda-forge toolchain, which is also good to have coverage.

I prefer to remove the Intel Mac from Pixi instead.

hjmjohnson:
> In the longer term, CI will start to catch up with the surges.

When we hit these issues during the surges, it greatly reduces the magnitude of the surge.  I've run into this several times over the past month.

hjmjohnson:
> Any consideration made for the time of each build? What is the most time efficient build?

I think the Azure builds are faster.

hjmjohnson:
@thewtex @dzenanz @blowekamp Do you know that the Azure pipelines are covering?  I don't know how those pipelines are instrumented.

hjmjohnson:
@blowekamp @thewtex @dzenanz Thanks for your feedback.  I think I incorporated the requested changes to pixi.

It's good to come to consensus quickly :).

thewtex:
> I think the Azure builds are faster.

My understanding is that they are running on the same infrastructure. However, I set up separate Azure Pipelines accounts per-build, so they are getting more runners.

thewtex:
The Azure builds are running `macos-13` (Intel Mac).

I would prefer to bring back both GitHub Actions macos-14 C++, Python builds. And remove both the macos-13 and macos-14 Pixi builds. We will be reducing conda-forge coverage, but it may be a necessary tradeoff for the CI speedup.

thewtex:
We could also limit macos-14 (ARM) Python builds to `master`...

PR #5047, opened 2024-12-11
COMP: Remove circular include dependency

hjmjohnson:
Reduce the number of files that need to be read by the pre-processor and remove the circular include dependancies.

The itkExceptionObject.h file was allowed to be included *only* by itkMacro.h which always included the file.

Moving the contents of itkExceptionObject.h to
itkMacro.h simplifies the dependency complexity.

itk::ExceptionObject is different from other classes because it depends on macros at the top of itkMacro.h for it's definition, but also is needed for many macros at the bottom of itkMacro.h.

ITK/Modules/Core/Common/include/itkExceptionObject.h:22:12: warning: circular header file dependency detected while including 'itkMacro.h', please check the include path [misc-header-include-cycle]
   22 | #  include "itkMacro.h"
      |            ^

## PR Checklist


Modules/Core/Common/include/itkMacro.h
```diff
@@ -1341,16 +1343,231 @@ compilers.
...
+} // end namespace itk
+// ====== End itkExceptionObject declaration
 
 /** itkDynamicCastInDebugMode
  * Use static_cast in Release builds, and dynamic_cast in Debug
  *
  * Note: this must come after:
- *
- *   #include "itkExceptionObject.h"
+ *  defining itkExceptionObject.h above
```
thewtex, line 1570: Can we add a historical note here that itkExceptionObject.h was previously separate file but itkMacro.h should only be #include's regardless?

hjmjohnson: yes.  I'll add that historical note at the start of the location where the old include used to be.

blowekamp:
I think smashing the file into Macro.h will make it less manageable. The circular dependency had been addresses with preprocessor commands. I am not clear what issue this is trying to address.  What effect will it have on Doxygen documentation?

hjmjohnson:
@blowekamp I respectfully disagree.  I have fought with those header guards and managed the convoluted circular many times.  They break static analysis tools that just give up when the circular dependency is encountered.   IDE's can fail to identify the circular inclusions, or how to handle them.  

It's probably not much of a speed increase, but it requires 3 file reads by the pre-processor rather than 1. (for every object file compilation that uses itkMacro.h). 

To me this is the better choice.

hjmjohnson:
@blowekamp I will try a different approach, but I think it will likely not work out.

blowekamp:
> @blowekamp I will try a different approach, but I think it will likely not work out.

I just did. Please see above and give it a try with your tools.

dzenanz (approved):
Looks reasonable to me.

dzenanz (approved):

hjmjohnson:
@blowekamp Your solution was what I thought about last night!  I've take your patch and modified to remove the circular dependency.  

I think the old code was overly complicated because:
1) the include was in the middle of the itkMacro.h file
2) the header guards tried to use itkExceptionMacro_h for 2 purposes (header guard and include control.

The current patch builds on macos-14 with python turned on.  It builds without errors, and it removes circular dependencies.

blowekamp (approved):
Thanks for taking a second stab the improvement.

thewtex (approved):

PR #5057, opened 2024-12-12
STYLE: clang-tidy bugprone-implicit-widening-of-multiplication-result

hjmjohnson:
The check diagnoses instances where a result of a multiplication is implicitly widened, and suggests (with fix-it) to either silence the code by making widening explicit, or to perform the multiplication in a wider type, to avoid the widening afterwards.

This is fine in general, but if width * height overflows, you end up wrapping back to the beginning of base instead of processing the entire requested buffer.

Indeed, this only matters for pretty large buffers (4GB+), but that can happen very easily for example in image processing, where for that to happen you “only” need a ~269MPix image.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkSymmetricEigenAnalysis.hxx
+++ b/Modules/Core/Common/include/itkSymmetricEigenAnalysis.hxx
@@ -18,6 +18,8 @@
 #ifndef itkSymmetricEigenAnalysis_hxx
 #define itkSymmetricEigenAnalysis_hxx
 
+#include <cstddef>
+
 #include "itkMath.h"
 #include "itkMakeUniqueForOverwrite.h"
 
@@ -60,7 +62,7 @@ SymmetricEigenAnalysis<TMatrix, TVector, TEigenMatrix>::ComputeEigenValuesLegacy
   const auto workArea1 = std::make_unique<double[]>(m_Dimension);
 
   // Copy the input matrix
-  const auto inputMatrix = make_unique_for_overwrite<double[]>(m_Dimension * m_Dimension);
+  const auto inputMatrix = make_unique_for_overwrite<double[]>(static_cast<size_t>(m_Dimension * m_Dimension));
   const auto dVector = make_unique_for_overwrite<double[]>(m_Dimension);
 
   unsigned int k = 0;
@@ -94,10 +96,10 @@ SymmetricEigenAnalysis<TMatrix, TVector, TEigenMatrix>::ComputeEigenValuesAndVec
   TEigenMatrix &  EigenVectors) const
 {
   const auto workArea1 = std::make_unique<double[]>(m_Dimension);
-  const auto workArea2 = std::make_unique<double[]>(m_Dimension * m_Dimension);
+  const auto workArea2 = std::make_unique<double[]>(static_cast<size_t>(m_Dimension * m_Dimension));
 
   // Copy the input matrix
-  const auto inputMatrix = make_unique_for_overwrite<double[]>(m_Dimension * m_Dimension);
+  const auto inputMatrix = make_unique_for_overwrite<double[]>(static_cast<size_t>(m_Dimension * m_Dimension));
   const auto dVector = make_unique_for_overwrite<double[]>(m_Dimension);
 
   unsigned int k = 0;
--- a/Modules/Core/Common/test/itkImportImageTest.cxx
+++ b/Modules/Core/Common/test/itkImportImageTest.cxx
@@ -16,6 +16,7 @@
  *
  *=========================================================================*/
 
+#include <cstddef>
 #include <iostream>
 #include "itkImageRegionIterator.h"
 #include "itkShrinkImageFilter.h"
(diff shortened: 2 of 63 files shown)


hjmjohnson:
I think that explicitly marking what was implicitly occurring is generally a good thing. 

I have mixed feelings about being even more verbose by changing automated fixups that clang-tidy provided with something like:
```c++
int i = 70000;
int j = 90000;
int row_size = 32000;
const itk::SizeValueType computedSize = static_cast<itk::SizeValueType>( i + (j*row_size));

auto value = myArray[ computedSize ];
```

Modules/Core/Common/include/itkSymmetricEigenAnalysis.hxx
```diff
@@ -18,6 +18,8 @@
 #ifndef itkSymmetricEigenAnalysis_hxx
 #define itkSymmetricEigenAnalysis_hxx
 
+#include <cstddef>
```
dzenanz, line 21: We might not need this, as these types should be defined in `itkIntTypes.h`.

dzenanz (commented):
I have mixed feelings about this PR.

blowekamp (commented):
I don't think using "unsigned long" would be a universally correct type.

hjmjohnson:
> I don't think using "unsigned long" would be a universally correct type.

"unsigned long" is not the type used.  clang-tidy computes the implicit type that is used, and puts that implicit type in the static_cast<>().

example clang-tidy conversion that are intended to help identify bugs (potentially buggy code shown):

```c++
/*
clang-format -i x.cxx; rm a.out; c++ -std=c++11 x.cxx ; ./a.out 16 44
~/local/llvm/bin/clang-tidy -fix --header-filter=".*"  --checks="bugprone*" x.cxx
*/

#include <iostream>
#include <cstdint>

template <typename TYPEA, typename TYPEB, typename TYPEC>
static auto
foo(TYPEA arga, TYPEB argb, TYPEC val) -> TYPEC
{
  std::cout << static_cast<long int>(arga) << " * " << static_cast<long int>(argb) << " = "
            << static_cast<long int>(val) << '\n';
  return val;
}

auto
main(int argc, char * argv[]) -> int
{
  const uint8_t a = std::stoi(argv[1]);
  const uint8_t b = std::stoi(argv[2]);

  // Original Code
  // uint16_t c = a * b;
  const uint8_t c = a * b;
  std::cout << static_cast<long int>(a) << " * " << static_cast<long int>(b) << " = " << static_cast<long int>(c)
            << '\n';
  std::cout << "=====" << '\n';

  // clang-tidy bugprone-implicit-widening-of-multiplication-result conversion suggestions:
  // NOTE: It does not fix the problem, but it makes the implicit conversion to unsigned_int more clear.
  foo<uint16_t, uint16_t, long int>(a, b, a * b);
  foo<uint16_t, uint16_t, unsigned long int>(a, b, a * b);
}
```
```txt
x.cxx:33:43: warning: performing an implicit widening conversion to type 'long' of a multiplication performed in type 'int' [bugprone-implicit-widening-of-multiplication-result]
   33 |   foo<uint16_t, uint16_t, long int>(a, b, a * b);
      |                                           ^
x.cxx:33:43: note: make conversion explicit to silence this warning
    6 |   foo<uint16_t, uint16_t, long int>(a, b, a * b);
      |                                           ^~~~~
      |                                           static_cast<long>( )
x.cxx:33:43: note: perform multiplication in a wider type
   33 |   foo<uint16_t, uint16_t, long int>(a, b, a * b);
      |                                           ^
      |                                           static_cast<long>( )
x.cxx:34:52: warning: performing an implicit widening conversion to type 'unsigned long' of a multiplication performed in type 'int' [bugprone-implicit-widening-of-multiplication-result]
   34 |   foo<uint16_t, uint16_t, unsigned long int>(a, b, a * b);
      |                                                    ^
x.cxx:34:52: note: make conversion explicit to silence this warning
   34 |   foo<uint16_t, uint16_t, unsigned long int>(a, b, a * b);
      |                                                    ^~~~~
      |                                                    static_cast<unsigned long>( )
x.cxx:34:52: note: perform multiplication in a wider type
   34 |   foo<uint16_t, uint16_t, unsigned long int>(a, b, a * b);
      |                                                    ^
      |                                                    static_cast<long>( )
```

```txt
16 * 44 = 192
=====
16 * 44 = 704
16 * 44 = 704
```

Modules/Core/Common/test/itkImportImageTest.cxx
```diff
@@ -26,7 +27,7 @@ int
 itkImportImageTest(int, char *[])
 {
   // Create a C-array to hold an image
-  auto * rawImage = new short[8 * 12];
+  auto * rawImage = new short[static_cast<unsigned long>(8 * 12)];
```
blowekamp, line 30: I am seeing unsigned long being used in a few places, which should not be a universally correct type.

hjmjohnson:
Not worth the benefits for making the type conversions explicit.

PR #5053, opened 2024-12-12
COMP: Fix compile error due to missing itk:: prefix for MakeFilled

dzenanz:
Sample error message:

`T:\Dashboard\ITK\Modules\Nonunit\Review\test\itkImageFunctionTest.cxx(137,17): error C2065: 'MakeFilled': undeclared identifier [T:\Dashboard\ITK-build\Modules\Nonunit\Review\test\ITKReviewTestDriver.vcxproj]`

## PR Checklist
- [x] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [x] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)

PR #5059, opened 2024-12-13
STYLE: Remove redundant idoms for setting origin/spacing

hjmjohnson:
The default origin is {0.0,...} and spacing is {1.0,...},
do not reset those values explicitly in tests.

When testing the Get/Set functions, use values other
then the default values.

Prefer initializations of regions with index and size
over explicit assignment calls of SetIndex() and SetSize()
immediately after declaration.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/test/itkImageTest.cxx
+++ b/Modules/Core/Common/test/itkImageTest.cxx
@@ -61,15 +61,12 @@ itkImageTest(int, char *[])
   image->GetSource();
   image->DisconnectPipeline();
 
-  auto                 spacing = itk::MakeFilled<Image::SpacingType>(1.0);
-  auto                 origin = itk::MakeFilled<Image::PointType>(1.0);
+
   Image::DirectionType direction;
   direction[0][0] = .5;
   direction[0][1] = .7;
   direction[1][0] = .7;
   direction[1][1] = .5;
-  image->SetSpacing(spacing);
-  image->SetOrigin(origin);
   image->SetDirection(direction);
 
   double dspacing[Image::ImageDimension] = { 2.0, 2.0 };
@@ -113,16 +110,17 @@ itkImageTest(int, char *[])
   }
 
   std::cout << "Test GetSmallestRegionContainingRegion." << std::endl;
+  // reset image spacing default values
+  auto spacing = itk::MakeFilled<Image::SpacingType>(1.0);
   image->SetSpacing(spacing);
-  origin.Fill(1.2);
+  auto origin = itk::MakeFilled<Image::PointType>(1.2);
   image->SetOrigin(origin);
   direction.SetIdentity();
   image->SetDirection(direction);
-  Image::RegionType      region;
+
   const Image::IndexType index{};
   auto                   size = Image::SizeType::Filled(4);
-  region.SetIndex(index);
-  region.SetSize(size);
+  Image::RegionType      region{ index, size };
   image->SetRegions(region);
 
   auto                   imageRef = Image::New();
@@ -154,23 +152,19 @@ itkImageTest(int, char *[])
   }
 
   using Image3D = itk::Image<float, 3>;
-  auto                     volume = Image3D::New();
-  auto                     spacingVol = itk::MakeFilled<Image3D::SpacingType>(1);
-  const Image3D::PointType originVol{};
-  Image3D::DirectionType   directionVol;
+  auto volume = Image3D::New();
+
+  Image3D::DirectionType directionVol;
   directionVol.SetIdentity();
-  volume->SetSpacing(spacingVol);
-  volume->SetOrigin(originVol);
   volume->SetDirection(directionVol);
 
-  Image3D::RegionType      cuboid;
+
   const Image3D::IndexType indexCuboid{};
   Image3D::SizeType        sizeCuboid;
   sizeCuboid[0] = 1;
   sizeCuboid[1] = 2;
   sizeCuboid[2] = 3;
-  cuboid.SetIndex(indexCuboid);
-  cuboid.SetSize(sizeCuboid);
+  Image3D::RegionType cuboid{ indexCuboid, sizeCuboid };
   volume->SetRegions(cuboid);
 
   using ProjectionTransformType = TestTransform<Image3D::ImageDimension>;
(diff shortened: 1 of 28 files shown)


hjmjohnson:
@N-Dekker This is based on your comment https://github.com/InsightSoftwareConsortium/ITK/pull/5053#discussion_r1882461374 from #5053

N-Dekker:
Thanks Hans! I agree that it's indeed usually better to not call `SetOrigin({0.0, 0.0, ...})` or `SetSpacing({1.0, 1.0, ...})` unnecessarily. It should be fine to rely on the defaults.

Didn't check your PR in detail yet But of course, it _might sometimes_ be useful to do such calls in a unit test, if it's specifically the purpose of the test to check them. Maybe a test would check that the origin remain zero after calling `SetOrigin({0.0, 0.0, ...})`. That might make sense.

Regarding the commit message: I would call it a "STYLE" commit, as it aims to improve coding style, without changing the behavior. I would prefer to use "ENH" for enhancements: newly added features.

Modules/Core/Common/test/itkImageTest.cxx
```diff
@@ -113,16 +110,17 @@ itkImageTest(int, char *[])
...
   image->SetOrigin(origin);
   direction.SetIdentity();
   image->SetDirection(direction);
-  Image::RegionType      region;
+
   const Image::IndexType index{};
   auto                   size = Image::SizeType::Filled(4);
-  region.SetIndex(index);
-  region.SetSize(size);
+  Image::RegionType      region{ index, size };
```
N-Dekker, line 123: Preferably do "One Commit One Change", so preferably don't change region initialization here.

hjmjohnson: I agree this is preferable, but ... that would triple the time necessary to make these changes.  If that level of granularity is required in this type of style commit, the second refactorings will never happen.

I'll add region changes to commit message.

hjmjohnson: FYI:  I tried to make two commits, but due to the proximity of changes in the code, it was hard to separate into two patches that independently worked.

hjmjohnson:
@dzenanz  @N-Dekker  If you approve of this set of code changes, I'll modify the commit message.

hjmjohnson:
NOTE:  All checks passed before the commit message update.

Modules/Core/Common/test/itkImageTest.cxx
N-Dekker: > I agree this is preferable, but ... that would triple the time necessary to make these changes.

OK, thanks, but are you in a hurry? Should all these style changes be applied _as soon as possible_? You know I like style improvements, but the amount is a bit overwhelming to me these days. Can you please explain, is there so much time pressure for those style changes right now?

You know, a style commit may help to prevent future bugs, but it may also introduce new bugs, when it is processed too quickly and carelessly.

dzenanz (approved):

Modules/Core/Common/test/itkImageTest.cxx
dzenanz: I suspect that Hans will have to get back to usual teaching responsibilities in January, so he wants to accomplish as much as possible before then.

Modules/Core/Common/test/itkImageTest.cxx
```diff
@@ -61,15 +61,12 @@ itkImageTest(int, char *[])
...
-  auto                 spacing = itk::MakeFilled<Image::SpacingType>(1.0);
-  auto                 origin = itk::MakeFilled<Image::PointType>(1.0);
+
   Image::DirectionType direction;
   direction[0][0] = .5;
   direction[0][1] = .7;
   direction[1][0] = .7;
   direction[1][1] = .5;
-  image->SetSpacing(spacing);
-  image->SetOrigin(origin);
```
N-Dekker, line 72: @hjmjohnson  This is a behavior change! In this particular case, `image->SetOrigin(origin)` set the origin to `{1.0, 1.0}`, originally. Was it your intention to change it to the default origin `{0.0, 0.0}` ?

hjmjohnson: This was a BAD test.  The value of origin had no-effect on what is tested.    The test does not change what it checks no-matter what the value of origin is.  It only exercised SetOrigin, but did not verify that the behavior was correct.

The testing of SetOrigin with values like "1234.0" and verifying that the behavior is correct is done elsewhere.

With regards to this test, the "SetOrigin" is pointless because it is untested and unused.

N-Dekker: OK, thanks Hans, I agree, a unit test should not just "exercise" a function, it should also check the effect of such a function call. Otherwise I agree that the function call is pointless.

Modules/Filtering/DisplacementField/test/itkInverseDisplacementFieldImageFilterTest.cxx
```diff
@@ -161,6 +146,16 @@ itkInverseDisplacementFieldImageFilterTest(int argc, char * argv[])
     ++it;
   }
 
+  // Test non-default values
+  auto spacing = itk::MakeFilled<DisplacementFieldType::SpacingType>(9876.0);
```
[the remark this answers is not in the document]
N-Dekker: Another option might be to use some "extreme" values from `std::numeric_limits<T>`.  (Again, this is just a suggestion.)

Modules/Core/Common/test/itkImageTest.cxx
N-Dekker: Thanks Dženan. @hjmjohnson Can you possibly still elaborate on this? I mean, is there indeed a time pressure for you to make as many style commits as possible, before January?

hjmjohnson: Yes,  I have a few slow days between teaching and research obligations.  I'll be going silent again starting next week until at least March 2025.

Additionally, this Pre-6.0 time period seems like a great time to make big steps, and take some risks.  Only 2 days left until next scheduled tag.

Modules/Filtering/DisplacementField/test/itkInverseDisplacementFieldImageFilterTest.cxx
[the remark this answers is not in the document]
hjmjohnson: I was looking into the "magic-numbers" clang-tidy, but had not tested that yet.

I did pick 9876 and 1234 as semi-obvious "extreme values".  I agree it would be better to have them named.  I'll do that when I look at the magic-numbers changes.

PR #5060, opened 2024-12-13
STYLE: Prefer python39+ coding style recommendations

hjmjohnson:
Only python3.9 and greater are supported, so remove
conditional code supporting <python3.9 statements.

Prefer native-supported type hints rather than
separate imports from the typing library.

Remove the redundant "r" designation for the default
file open behavior.

## PR Checklist

### DIFF
--- a/Modules/Filtering/Smoothing/wrapping/test/MedianImageFilterTest.py
+++ b/Modules/Filtering/Smoothing/wrapping/test/MedianImageFilterTest.py
@@ -25,11 +25,8 @@ import itk.support.types as itkt
 from sys import argv, version_info
 import warnings
 
-if version_info >= (3, 8):
-    from typing import Sequence, TypeVar, get_type_hints, get_args, get_origin, Union
-else:
-    from typing import Sequence, TypeVar, Union
-
+from typing import TypeVar, get_type_hints, get_args, get_origin, Union
+from collections.abc import Sequence
 
 try:
     from numpy.typing import ArrayLike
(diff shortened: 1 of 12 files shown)


dzenanz (approved):

hjmjohnson:
only mac builds have a single python test failure:
```
  File "/Users/runner/work/1/s/Modules/Filtering/Smoothing/wrapping/test/MedianImageFilterTest.py", line 69, in <module>
    assert Sequence[int] in get_args(radius_hints)
AssertionError
```

Double checking to ensure that this is a transient issue.

Modules/Filtering/Smoothing/wrapping/test/MedianImageFilterTest.py
```diff
@@ -25,11 +25,8 @@
 from sys import argv, version_info
 import warnings
 
-if version_info >= (3, 8):
-    from typing import Sequence, TypeVar, get_type_hints, get_args, get_origin, Union
-else:
-    from typing import Sequence, TypeVar, Union
-
+from typing import TypeVar, get_type_hints, get_args, get_origin, Union
```
thewtex, line 28: Does keeping the `Sequence` import from `typing` instead of `collections.abc` address the failure?

hjmjohnson:
I verified that this is a regression on Mac with Python builds.

It seems like there is a problem with the expected structure of type hints.  Will look at this soon.

Modules/Filtering/Smoothing/wrapping/test/MedianImageFilterTest.py
hjmjohnson: @thewtex That would be one way, but Sequence is deprecated in typing in 3.9 and the recommended import is from collections.abc.

I've fixed this by importing from collections.abc.  See the new patchset.

Wrapping/Generators/SwigInterface/igenerator.py
```diff
@@ -1242,7 +1241,8 @@ def generate_process_object_snake_case_functions(self, typedefs):
                 self.outputFile.write(
                     f"""from itk.support import helpers
 import itk.support.types as itkt
-from typing import Sequence, Tuple, Union
+from typing import Union
+from collections.abc import Sequence
```
hjmjohnson, line 1245: @thewtex   This fixes and is compatibible with non-deprecated 3.9 feature set.

Wrapping/Generators/SwigInterface/igenerator.py
```diff
@@ -438,9 +437,9 @@ def get_arg_type(decls, arg_type, for_snake_case_hints=True):
     elif arg_type_str.startswith("itk::FlatStructuringElement<"):
         return lib + "FlatStructuringElement"
     elif arg_type_str.startswith("itk::RGBPixel<"):
-        return "Tuple[int, int, int]"
+        return "tuple[int, int, int]"
     elif arg_type_str.startswith("itk::RGBAPixel<"):
-        return "Tuple[int, int, int, int]"
+        return "tuple[int, int, int, int]"
```
hjmjohnson, line 442: This allows us to remove 'from typing import Tuple' that is deprecated in 3.9.

PR #5063, opened 2024-12-14
FEEDBACK NEEDED:  STYLE: performance-avoid-endl clang-tidy recommendation

hjmjohnson:
Checks for uses of std::endl on streams and suggests using the newline character '\n' instead.

Using std::endl on streams can be less efficient than using the newline character '\n' because std::endl performs two operations: it writes a newline character to the output stream and then flushes the stream buffer.  Writing a single newline character using '\n' does not trigger a flush, which can improve performance. In addition, flushing the stream buffer can cause additional overhead when working with streams that are buffered.

This has a side effect of making the code more compact as well, which sometimes makes Examples more readable.
## PR Checklist


hjmjohnson:
I agree that there are times when flushing is essential. Perhaps the last element of all PrintSelf should be flushed, but we are flushing the buffer every few bytes in many cases.  This excessive flushing can slow down the progress of the application.  

When doing registrations with lots of diagnostics, flushing can significantly slow down the registration process.  I think that flushing should be optional in more places (i.e. the default) and programmers can flush when they identify needing the output immediately.  

I'm putting this in draft mode to gather more input.

thewtex:
To make the code easier to write and maintain, I would prefer to default to `std::endl` and substitute with `\n` in places where performance may be an issue.

hjmjohnson:
@blowekamp @thewtex Thanks for your feedback.  It is greatly appreciated.

blowekamp:
From a style perspective I think keeping std::ends is good in Examples and Tests. I am currious what the performance gain would be with this change in methods like PrintSelf.

PR #5064, opened 2024-12-14
BUG: Fix ExposeMetaData<std::vector<double>> im MetaImageIO

### LINKED ISSUE #5062: Wrong metadata "ITK_original_spacing" (MetaIO)
E.g. we take some 3D image with spacing [1.0, 2.0, 3.0], load it in ITK and save **MHA** file.
The saved image has incorrect metadata values in _ITK_original_spacing_:
`ITK_original_spacing = 2 3 3`

Spacing is correct, just this additional _ITK_original_spacing_ tag is incorrect

S. [itkImageFileReader.hxx](https://github.com/InsightSoftwareConsortium/ITK/blob/6775839a3ad65ad6e1b11c1c674f9c327f11e9f9/Modules/IO/ImageBase/include/itkImageFileReader.hxx#L218-L219)

```cpp
  EncapsulateMetaData<std::vector<double>>(
    thisDic, "ITK_original_spacing", std::vector<double>(spacing, spacing + TOutputImage::ImageDimension));
```

[Test](https://drive.google.com/file/d/1sWaNUdJodzdpWKqPm79Ha_peZKqCvbrb/view?usp=sharing)

P.S. I don't know where the problem happens, above lines from itkImageFileReader.hxx seem to be OK, not sure.

issakomi:
Indexes were incorrectly accessed with pre-increment instead of post-increment

```cpp
#include <iostream>
#include <vector>

int main()
{
  std::vector<int> v;
  v.push_back(1);
  v.push_back(2);
  v.push_back(3);

  {
    unsigned int i = 0;
    while (i < v.size() - 1)
    {
      std::cout << v[++i] << ' '; // wrong
    }
    std::cout << v[i] << std::endl;
  }

  {
    unsigned int i = 0;
    while (i < v.size() - 1)
    {
      std::cout << v[i++] << ' ';
    }
    std::cout << v[i] << std::endl;
  }

  return 0;
}
```

```
r@deb2:~$ ./a.out 
2 3 3
1 2 3
```

Closes #5062

### DIFF
--- a/Modules/IO/Meta/src/itkMetaImageIO.cxx
+++ b/Modules/IO/Meta/src/itkMetaImageIO.cxx
@@ -637,7 +637,7 @@ MetaImageIO::WriteImageInformation()
       unsigned int i = 0;
       while (i < vval.size() - 1)
       {
-        strs << vval[++i] << ' ';
+        strs << vval[i++] << ' ';
       }
       strs << vval[i];
     }


issakomi:
BTW, `i < v.size() - 1` is a little bit dangerous, what if _size_ is 0. I have tested, segfault. Probably it is unlikely to happen, just FYI.

issakomi:
BTW, in my code I would prefer something like

```cpp
  {
    const size_t vsize = v.size();
    if (vsize > 0)
    {
        for (size_t i = 0; i < vsize; ++i)
        {
          if (i > 0)
          {
            std::cout << 'x';
          }
          std::cout << v.at(i);
        }
    }
    std::cout << std::endl;
  }
```

Otherwise there can be issues if vector size is 0 or 1. Maybe there is something is _std_ from C++42 for the purpose?

Modules/IO/Meta/src/itkMetaImageIO.cxx
```diff
@@ -637,7 +637,7 @@ MetaImageIO::WriteImageInformation()
       unsigned int i = 0;
       while (i < vval.size() - 1)
       {
-        strs << vval[++i] << ' ';
+        strs << vval[i++] << ' ';
```
hjmjohnson, line 640: I prefer a function call like:
```c++
// Function to join strings with a delimiter similar to python's   ' '.join([1, 2, 3 ])
void join(const std::vector<std::string>& elements, const std::string& delimiter,  std::ostringstream strs) {
    for (size_t i = 0; i < elements.size(); ++i) {
        strs << elements[i];
        if (i != elements.size() - 1) {
            strs << delimiter;
        }
    }
}
```

This avoids the pre-post increment relationship to the [] operator.

hjmjohnson (changes requested):
I think this can be further improved using a for loop for clarity

dzenanz (approved):
Good as-is. It can be further refactored in a follow-up commit or PR.

hjmjohnson (approved):
Agree @dzenanz  Thanks for the insight.

issakomi:
@thewtex  Maybe this commit could be included in upcoming 5.4 bugfix release?

blowekamp:
> @thewtex Maybe this commit could be included in the upcoming 5.4 bugfix release?

+1

PR #5065, opened 2024-12-14
COMP: fixed some clang -Wformat-nonliteral warnings

seanm:
Not sure if this works with gcc...

### DIFF
--- a/Modules/Core/Common/include/itkMacro.h
+++ b/Modules/Core/Common/include/itkMacro.h
@@ -135,6 +135,13 @@ namespace itk
 #define CLANG_PRAGMA_POP ITK_CLANG_PRAGMA_POP
 #define CLANG_SUPPRESS_Wfloat_equal ITK_GCC_SUPPRESS_Wfloat_equal
 
+// For GCC/Clang, to decorate printf-like functions allowing for compiler checks of format strings.
+#if defined(__GNUC__)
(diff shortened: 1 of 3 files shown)


hjmjohnson:
@seanm This seems to work on all platforms!

seanm:
> @seanm This seems to work on all platforms!

Do you know if the CI here has gcc with -Wformat-literal enabled?

hjmjohnson:
@seanm  I don't think so

```bash
git grep format-literal did not have any responses.
```

seanm:
Where/how can we add warnings flags to CI?

dzenanz:
I believe you need to set them [here](https://github.com/InsightSoftwareConsortium/ITK/blob/v6.0a01/Testing/ContinuousIntegration/AzurePipelinesMacOS.yml#L74-L78) and [here](https://github.com/InsightSoftwareConsortium/ITK/blob/v6.0a01/.github/workflows/macos-arm.yml#L64-L69).

seanm:
Thanks. But I don't see any `-W` in either of those. Are no warning flags being set already? Just the compiler's default is used?

dzenanz:
CMake's defaults, probably, which might be slightly different from compiler's defaults.

seanm:
I forgot about this PR... IMHO can merge after review...

hjmjohnson (approved):

hjmjohnson (approved):

PR #5087, opened 2024-12-17
ENH: Make default-constructors of RGBPixel and RGBAPixel `constexpr`

N-Dekker:
Tested by means of `CheckConstexprBeginAndEndOfContainer()`. Note that these default-constructors were already `constexpr` _implicitly_, as implied by `= default`, when `ITK_FUTURE_LEGACY_REMOVE` would be enabled.

### DIFF
--- a/Modules/Core/Common/include/itkRGBAPixel.h
+++ b/Modules/Core/Common/include/itkRGBAPixel.h
@@ -81,7 +81,13 @@ public:
 #ifdef ITK_FUTURE_LEGACY_REMOVE
   RGBAPixel() = default;
 #else
-  RGBAPixel() { this->Fill(0); }
+  constexpr RGBAPixel()
+    : Superclass(Superclass())
+  {
+    // `: Superclass(Superclass())` is a workaround for an old compiler bug. A simple `: Superclass()` triggered
+    // warnings from GCC 9.4.0 saying: "warning: '<anonymous>' may be used uninitialized in this function
+    // [-Wmaybe-uninitialized]".
+  }
 #endif
 
   /** Pass-through constructor for the Array base class. */
--- a/.github/workflows/itk_dict.txt
+++ b/.github/workflows/itk_dict.txt
@@ -226,6 +226,7 @@ Wachowiak
 Wanlin
 Wdeprecated
 Wi
+Wmaybe
 Wtautological
 Xu
 Y'CbCr
--- a/Modules/Core/Common/include/itkRGBPixel.h
+++ b/Modules/Core/Common/include/itkRGBPixel.h
@@ -80,7 +80,13 @@ public:
 #ifdef ITK_FUTURE_LEGACY_REMOVE
   RGBPixel() = default;
 #else
-  RGBPixel() { this->Fill(0); }
+  constexpr RGBPixel()
+    : Superclass(Superclass())
+  {
+    // `: Superclass(Superclass())` is a workaround for an old compiler bug. A simple `: Superclass()` triggered
+    // warnings from GCC 9.4.0 saying: "warning: '<anonymous>' may be used uninitialized in this function
+    // [-Wmaybe-uninitialized]".
+  }
 #endif
 
 #if defined(ITK_LEGACY_REMOVE)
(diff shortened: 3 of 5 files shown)


dzenanz (approved):

dzenanz:
I use TortoiseGit, which implicitly uses plain push. I forgot what review-push does, but I am sure it could be found in the documentation.

thewtex (approved):
🚀 

@N-Dekker awesome!!

Yes, both ways work for creating the PR.

Modules/Core/Common/include/itkRGBAPixel.h
```diff
@@ -81,7 +81,9 @@ class ITK_TEMPLATE_EXPORT RGBAPixel : public FixedArray<TComponent, 4>
 #ifdef ITK_FUTURE_LEGACY_REMOVE
   RGBAPixel() = default;
 #else
-  RGBAPixel() { this->Fill(0); }
+  constexpr RGBAPixel()
+    : Superclass()
```
N-Dekker, line 85: With this [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/7544cbd5a0302f9d53c60bf536e595288c7bafc4..6464b887f18b156562c1c21d6146b1af7f1beb2a), replacing `{}` with `()`, I'm hoping to address the warnings at https://open.cdash.org/viewBuildError.php?type=1&buildid=10096694 saying:

> In file included from [Modules/Core/Common/test/itkNumericTraitsTest.cxx:42](https://github.com/InsightSoftwareConsortium/ITK/blob/7544cbd5a0302f9d53c60bf536e595288c7bafc4/Modules/Core/Common/test/itkNumericTraitsTest.cxx#L42):
/.../s/Modules/Core/Common/include/itkNumericTraitsRGBPixel.h: In function 'void numeric_traits_test::CheckVariableLengthArrayTraits(const T&) [with T = itk::RGBPixel<char>]':

[Modules/Core/Common/include/itkNumericTraitsRGBPixel.h:118](https://github.com/InsightSoftwareConsortium/ITK/blob/7544cbd5a0302f9d53c60bf536e595288c7bafc4/Modules/Core/Common/include/itkNumericTraitsRGBPixel.h#L118):17: warning: '<anonymous>' may be used uninitialized in this function [-Wmaybe-uninitialized]

hjmjohnson:
@N-Dekker  I only use `git push`. The `git review-push` looks like it does extra work that is probably no longer needed.

Modules/Core/Common/include/itkRGBAPixel.h
N-Dekker: It appears that with the old [Linux GCC 9.4.0](https://open.cdash.org/builds/10097099), the `constexpr` default-constructors that I wrote just don't properly zero-initialize the components, as they should. I'll think of another workaround...

N-Dekker: Another attempt, initializing by the seemingly redundant expression `Superclass(Superclass())`: [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/6464b887f18b156562c1c21d6146b1af7f1beb2a..07ee85ae52a882bdc634cf93fd3304e026d6bac0)

dzenanz: Now might be a reasonable time to reconsider which compiler versions we support in ITK 6?

hjmjohnson:
I think that "".github/workflows/itk_dict.txt"" needs to have "Wmaybe" added.

Modules/Core/Common/include/itkRGBAPixel.h
N-Dekker: @dzenanz It looks like this particular bug (as I reproduced with the example code at https://github.com/InsightSoftwareConsortium/ITK/pull/5087#discussion_r1889314541) is fixed with GCC 12.1 (which was released in May 2022): https://godbolt.org/z/of6Tdbzcj  GCC 11 still appears to have the compiler bug. I'm not sure if we can already drop GCC 11 now. Ubuntu-22.04 still comes with GCC 11.4 🤷 

Having said so, I think we have a workaround for this one now 😃

hjmjohnson:
I created an issue to track how this may be done automatically:  https://github.com/InsightSoftwareConsortium/ITK/issues/5092

PR #5088, opened 2024-12-17
BUG: GenerateImageSource sets Size from ReferenceImage

thewtex:
We want the same behavior as SetOutputParametersFromImage.

Without this, a crash occurs on Update().

### DIFF
--- a/Modules/Filtering/ImageSources/include/itkGenerateImageSource.hxx
+++ b/Modules/Filtering/ImageSources/include/itkGenerateImageSource.hxx
@@ -63,6 +63,9 @@ GenerateImageSource<TOutputImage>::GenerateOutputInformation()
       outputPtr->SetSpacing(referenceImage->GetSpacing());
       outputPtr->SetOrigin(referenceImage->GetOrigin());
       outputPtr->SetDirection(referenceImage->GetDirection());
+      this->m_Direction = referenceImage->GetDirection();
+      this->m_StartIndex = referenceImage->GetLargestPossibleRegion().GetIndex();
+      this->m_Size = referenceImage->GetLargestPossibleRegion().GetSize();
     }
     else
     {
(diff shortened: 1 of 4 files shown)


Modules/Filtering/ImageSources/include/itkGenerateImageSource.hxx
```diff
@@ -63,6 +63,9 @@ GenerateImageSource<TOutputImage>::GenerateOutputInformation()
       outputPtr->SetSpacing(referenceImage->GetSpacing());
       outputPtr->SetOrigin(referenceImage->GetOrigin());
       outputPtr->SetDirection(referenceImage->GetDirection());
+      this->SetDirection(referenceImage->GetDirection());
+      this->SetStartIndex(referenceImage->GetLargestPossibleRegion().GetIndex());
+      this->SetSize(referenceImage->GetLargestPossibleRegion().GetSize());
```
blowekamp, line 68: Calling this method will modify the m_Time of the filter. This may cause problems with executing the pipeline.

Perhaps the GridImageSource use of ["GetSize](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/ImageSources/include/itkGridImageSource.hxx#L56)" is incorrect? This would be tedious to check all derived classes are not accessing these parameters.

Alternatively, directly setting the Ivars, m_Direction, ..., would not modify the MTime.

blowekamp (commented):
Good find. Nice test. I think it could use a tweak though.

dzenanz:
For posterity, this was reported on the forum:
https://discourse.itk.org/t/using-set-setuse-referenceimage-for-gridimagesource-in-itk-python/7353

Modules/Filtering/ImageSources/include/itkGenerateImageSource.hxx
thewtex: Good point. I'll set with the ivar.

hjmjohnson:
@thewtex Looks like a fix missed the clang-format fix.

hjmjohnson (changes requested):
clang-format linting failed.

thewtex:
Yes, the clang-format linter CI failure is unrelated. Fix being tested in https://github.com/InsightSoftwareConsortium/ITK/pull/5102

PR #4940, opened 2024-11-07
Hdf5 1.14.5

blowekamp:
## PR Checklist
- [ ] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [ ] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [ ] Added test (or behavior not changed)
- [ ] Updated API documentation (or API not changed)
- [ ] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [ ] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [ ] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples) for all new major features (if any)

Refer to the [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) for
further development details if necessary.

PR #5097, opened 2024-12-19
ENH: Update CMake minimum support to 3.22.1

blowekamp:
The update to HDF5 #4940, added a required CMake minimum version of 3.18. Update the top-level requirement to match.

While possible to build ITK without HDF5, this is not a common configuration. The non-minimum requirement at the top level can be considered a bug. 

Note: Ubuntu 20.04 contains cmake 2.16.3, but standard support end April 2025 ( likely before the ITK 6.0 final release).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,7 @@
 # CMake versions greater than the ITK_NEWEST_VALIDATED_POLICIES_VERSION policies will
 # continue to generate policy warnings "CMake Warning (dev)...Policy CMP0XXX is not set:"
 #
-set(ITK_OLDEST_VALIDATED_POLICIES_VERSION "3.16.3")
+set(ITK_OLDEST_VALIDATED_POLICIES_VERSION "3.22.1")
 set(ITK_NEWEST_VALIDATED_POLICIES_VERSION "3.29.0")
 cmake_minimum_required(VERSION ${ITK_OLDEST_VALIDATED_POLICIES_VERSION}...${ITK_NEWEST_VALIDATED_POLICIES_VERSION}
                        FATAL_ERROR)


CMakeLists.txt
```diff
@@ -10,7 +10,7 @@
 # CMake versions greater than the ITK_NEWEST_VALIDATED_POLICIES_VERSION policies will
 # continue to generate policy warnings "CMake Warning (dev)...Policy CMP0XXX is not set:"
 #
-set(ITK_OLDEST_VALIDATED_POLICIES_VERSION "3.16.3")
+set(ITK_OLDEST_VALIDATED_POLICIES_VERSION "3.18.0")
```
dzenanz, line 13: Should we instead consider bumping up to 3.22.1, which was [shipped with ubuntu 22.04](https://launchpad.net/ubuntu/jammy/+source/cmake)?

hjmjohnson: I agree with @dzenanz.  It gives us the longest runway without an additional change.

thewtex: +1

CMakeLists.txt
```diff
@@ -10,7 +10,7 @@
...
 set(ITK_NEWEST_VALIDATED_POLICIES_VERSION "3.29.0")
```
hjmjohnson, line 14: FYI: homebrew mac version of 3.31.2 seems to work fine too.

hjmjohnson (changes requested):

dzenanz (approved):

CMakeLists.txt
dzenanz: We can update the highest verified version in a separate PR. That update needs some testing.

PR #5098, opened 2024-12-19
COMP: Address non-const operations when TBB enabled in test

blowekamp:
Addresses the following build error:
https://open.cdash.org/viewBuildError.php?buildid=10099800

Do we not a TBB build in CI?


dzenanz (approved):

dzenanz:
I guess we need to add `-DModule_ITKTBB:BOOL=ON` [here](https://github.com/InsightSoftwareConsortium/ITK/blob/e917d9f08801a403ff5b5ea218fa76df37cfb288/pyproject.toml#L35-L39), or somewhere else in that file.

dzenanz (approved):

dzenanz:
I added some MacOS nightly builds recently. The one called `MacOS-Rel-Favorite-Remotes` already has TBB enabled, and I just also added it to `MacOS-Rel-Python`.

PR #5099, opened 2024-12-20
WIP: Exploring the behavior of SetSpacing

hjmjohnson:
The SpatialObjectToImageFilter::SetSpacing has a specialized behavior when zeros are passed for the spacing size.

It is unclear what the intended behavior is.

Expanded the testsuite to demonstrate that sending all zeros as the spacing values to update results in no changes to the state of itkSpatialObjectToImageFilter, but does trigger a Modified status on the Object.

## PR Checklist

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkSpatialObjectToImageFilter.hxx
+++ b/Modules/Core/SpatialObjects/include/itkSpatialObjectToImageFilter.hxx
@@ -84,9 +84,8 @@ template <typename TInputSpatialObject, typename TOutputImage>
 void
 SpatialObjectToImageFilter<TInputSpatialObject, TOutputImage>::SetSpacing(const SpacingType & spacing)
 {
-  unsigned int i;
-
-  for (i = 0; i < TOutputImage::ImageDimension; ++i)
+  unsigned int i = 0;
+  for (; i < TOutputImage::ImageDimension; ++i)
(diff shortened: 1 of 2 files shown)


hjmjohnson:
@aylward @jjomier The behavior of the SetSpacing for the itkSpatialObjectToImageFilter.hxx has a strange behavior.

If the value of the new spacing is zero, the old value is preserved.  This seems strange to me.  I would have guessed that requesting a zero-sized spacing would (a) throw an exception or (b) set the spacing value to 1.0, or (c) actually set the value to zero.

Silently not doing what is requested seems like the wrong answer.

I would like to mirror the behavior from itkImageBase.h
```cxx
   if (spacing[i] == 0.0) 
   {
      itkExceptionMacro(
        "Zero-valued spacing is not supported and may result in undefined behavior.\nRefusing to change spacing from "
        << this->m_Spacing << " to " << spacing);
    }
```

Do you forsee an issue with this approach?

hjmjohnson:
See possible solution in :  https://github.com/InsightSoftwareConsortium/ITK/pull/5101

hjmjohnson:
A solution is implemented in https://github.com/InsightSoftwareConsortium/ITK/pull/5101.  The more I review the proposed solution, the more I'm convinced it is the correct way forward.

PR #5100, opened 2024-12-20
BUG: Consistency of enforcing m_ShrinkFactors > 1

hjmjohnson:
The behavior of setting individual m_ShrinkFactor elements should be the same as setting from an array of shrinkFactors.

## PR Checklist

### DIFF
--- a/Modules/Filtering/ImageGrid/include/itkShrinkImageFilter.hxx
+++ b/Modules/Filtering/ImageGrid/include/itkShrinkImageFilter.hxx
@@ -98,7 +98,14 @@ ShrinkImageFilter<TInputImage, TOutputImage>::SetShrinkFactor(unsigned int i, un
   }
 
   this->Modified();
-  m_ShrinkFactors[i] = factor;
+  if (m_ShrinkFactors[i] < 1)
+  {
+    m_ShrinkFactors[i] = 1;
+  }
+  else
+  {
+    m_ShrinkFactors[i] = factor;
+  }
 }
 
 


dzenanz (approved):

Modules/Filtering/ImageGrid/include/itkShrinkImageFilter.hxx
```diff
@@ -98,7 +98,14 @@ ShrinkImageFilter<TInputImage, TOutputImage>::SetShrinkFactor(unsigned int i, un
...
   this->Modified();
-  m_ShrinkFactors[i] = factor;
+  if (m_ShrinkFactors[i] < 1)
+  {
+    m_ShrinkFactors[i] = 1;
+  }
+  else
+  {
+    m_ShrinkFactors[i] = factor;
+  }
```
N-Dekker, line 108: @hjmjohnson Did you test the new behavior? It seems strange that you ignore the `factor` parameter when the original `m_ShrinkFactors[i]` value is less than `1`.

Instead of the if-else, I would expect something like `m_ShrinkFactors[i] = std::max(1, factor)`. But that is different from the new behavior introduced by PR #5100, of course.

N-Dekker: @hjmjohnson Sorry but do you agree that your PR #5100 has introduced a new bug? At https://github.com/InsightSoftwareConsortium/ITK/blob/74296d2ddcae8c7e5a16e5bdc35d774e8252b4bc/Modules/Filtering/ImageGrid/include/itkShrinkImageFilter.hxx#L83-L99

PR #5101, opened 2024-12-20
STYLE: Update localization for loop inspired code localizations

hjmjohnson:
STYLE: Prefer local initialization

Declare variables when they can be initialized immediatly.

Localize loop variable initialization using
regex to find common patterns to minimize:

REPLACE: (^\s*)([a-z]*) ([a-z])\;\n(\s*)for \(\3 \= ([0-9]+)\;
   WITH: $4for ($2 $3= $5;

SEARCH:  for \([a-zA-Z]+ =
         Manually review and replace loop indices

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkMacro.h
+++ b/Modules/Core/Common/include/itkMacro.h
@@ -862,7 +862,7 @@ compilers.
 // clang-format off
 /** Set a decorated input. This defines the Set"name"() and a Set"name"Input() method */
 #define itkSetDecoratedInputMacro(name, type)                                                                       \
-  virtual void Set##name##Input(const SimpleDataObjectDecorator<type> * _arg)                                       \
+  virtual void Set## name## Input(const SimpleDataObjectDecorator<type> * _arg)                                       \
   {                                                                                                                 \
     itkDebugMacro("setting input " #name " to " << _arg);                                                           \
     if (_arg != itkDynamicCastInDebugMode<SimpleDataObjectDecorator<type> *>(this->ProcessObject::GetInput(#name))) \
@@ -871,8 +871,8 @@ compilers.
       this->Modified();                                                                                             \
     }                                                                                                               \
   }                                                                                                                 \
-  virtual void Set##name(const SimpleDataObjectDecorator<type> * _arg) { this->Set##name##Input(_arg); }            \
-  virtual void Set##name(const type & _arg)                                                                         \
+  virtual void Set## name(const SimpleDataObjectDecorator<type> * _arg) { this->Set## name## Input(_arg); }            \
+  virtual void Set## name(const type & _arg)                                                                         \
   {                                                                                                                 \
     using DecoratorType = SimpleDataObjectDecorator<type>;                                                          \
     itkDebugMacro("setting input " #name " to " << _arg);                                                           \
@@ -887,7 +887,7 @@ compilers.
     ITK_GCC_PRAGMA_POP                                                                                              \
     auto newInput = DecoratorType::New();                                                                           \
     newInput->Set(_arg);                                                                                            \
-    this->Set##name##Input(newInput);                                                                               \
+    this->Set## name## Input(newInput);                                                                               \
   }                                                                                                                 \
   ITK_MACROEND_NOOP_STATEMENT
 // clang-format on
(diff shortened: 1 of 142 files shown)


Modules/Core/SpatialObjects/include/itkSpatialObjectToImageFilter.hxx
```diff
@@ -84,25 +84,18 @@ template <typename TInputSpatialObject, typename TOutputImage>
...
   {
-    for (i = 0; i < TOutputImage::ImageDimension; ++i)
+    this->Modified();
+    for (unsigned int i = 0; i < TOutputImage::ImageDimension; ++i)
     {
-      if (spacing[i] != 0)
+      if (spacing[i] <= 0)
       {
-        m_Spacing[i] = spacing[i];
+        itkExceptionMacro("Zero-valued spacing and negative spacings are not supported and may result in undefined "
```
blowekamp, line 94: In the ImageBase class, the exception only occur with LEGACY_REMOVE. The conversion I have this having the DataObject block certain settings is that it may prevent users from loading and correcting existing data. 

It may be a better approach to negative spacing be a warning in the DataObject, but the ProcessObject's have an exception when the inputs have these invalid values. I don't know if the Spatial object have a base ProcessObject like the Image filters do. Just my 2 cents.

Modules/Core/ImageFunction/include/itkGaussianBlurImageFunction.hxx
```diff
@@ -78,22 +78,10 @@ template <typename TInputImage, typename TOutput>
 void
 GaussianBlurImageFunction<TInputImage, TOutput>::SetSigma(const double * sigma)
 {
-  unsigned int i;
-
-  for (i = 0; i < Self::ImageDimension; ++i)
+  if (ContainerCopyWithCheck(m_Sigma, sigma, Self::ImageDimension))
```
dzenanz, line 81: How did you identify these instances? Local search for `this->Modified()` has ~1k hits. Most of them are scalar variables, but this might be a good way to identify the vector ones which would benefit from this refactoring.

Modules/Core/SpatialObjects/test/itkSpatialObjectToImageFilterTest.cxx
```diff
@@ -83,31 +83,69 @@ itkSpatialObjectToImageFilterTest(int, char *[])
   // Testing spacing
   std::cout << "Testing Spacing: ";
 
-  float  spacingFloat[2];
-  double spacingDouble[2];
-
-  for (unsigned int i = 0; i < 2; ++i)
+  constexpr float  floatCheckValue = 1.5;
```
dzenanz, line 86: Remove from commit message: `# Please enter the commit message for your changes. Lines
starting`

Modules/Core/Common/include/itkMacro.h
```diff
@@ -1200,33 +1200,73 @@ compilers.
   virtual void name##Off() { this->Set##name(false); } \
   ITK_MACROEND_NOOP_STATEMENT
 
+
```
dzenanz, line 1203: In commit message `redunancy` -> `redundancy`.

Modules/Core/ImageFunction/include/itkGaussianBlurImageFunction.hxx
hjmjohnson: I only looked for the setter functions that had this particular signature where the ContainerCopyWithCheck and ContainerFillWithCheck could be used as a replacement.

Manual effort. :(

Modules/Core/Common/include/itkMacro.h
hjmjohnson: Fixed.  I'm looking for a decent commit message spell/grammar checker.  It's hard to find one that will be universal and not too obtrusive.

hjmjohnson: --- Now I'm downright embarrassed.  I've been using Vim for decades, and I did know that recent versions have spell-checking built in.

I added `autocmd FileType gitcommit setlocal spell` to my ~/.vimrc and now the commit messages are spell checked.  


I have a note to add that gem somewhere in the documentation for developers in the future.

Modules/Core/SpatialObjects/include/itkSpatialObjectToImageFilter.hxx
hjmjohnson: itkSpatialObjectToImageFilter does derive from ProcessObject, so I think this implementation of throwing an exception here does conform to your recommendation.

This is a bug fix, and I don't know what the correct behavior should be without the Exception.

I appreciate your thoughts here.  I don't see how to easily implement a warning in the DataObject.

dzenanz (approved):
LGTM

hjmjohnson:
Minor changes needed to get Windows types aligned for "long" vs "int_64" comparisons.

seanm:
@hjmjohnson I think this maybe broke the build with `ITK_USE_GPU=ON`:

```
/ITK/Modules/Filtering/GPUSmoothing/include/itkGPUDiscreteGaussianImageFilter.hxx:233:12: error: use of undeclared identifier 'i'
  233 |       for (i = 1; i < filterDimensionality - 1; ++i)
      |            ^
```

PR #5110, opened 2024-12-22
STYLE: Prefer compile-time constexpr

hjmjohnson:
The values of m_Factor* are known at compile-
time, so compute them only at compile-time.

## PR Checklist

### DIFF
--- a/Modules/Core/ImageFunction/include/itkWindowedSincInterpolateImageFunction.h
+++ b/Modules/Core/ImageFunction/include/itkWindowedSincInterpolateImageFunction.h
@@ -46,7 +46,7 @@ public:
 
 private:
   /** Equal to \f$ \frac{\pi}{2 m} \f$ */
-  static const double m_Factor;
+  static constexpr double m_Factor = itk::Math::pi / (2 * VRadius);
 };
 
 /**
@@ -68,7 +68,7 @@ public:
 
 private:
   /** Equal to \f$ \frac{\pi}{m} \f$ */
-  static const double m_Factor;
+  static constexpr double m_Factor = itk::Math::pi / VRadius;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


Modules/Core/ImageFunction/include/itkWindowedSincInterpolateImageFunction.h
```diff
@@ -46,7 +46,7 @@ class ITK_TEMPLATE_EXPORT CosineWindowFunction
 
 private:
   /** Equal to \f$ \frac{\pi}{2 m} \f$ */
-  static const double m_Factor;
+  static constexpr double m_Factor = itk::Math::pi / (2 * VRadius);
```
N-Dekker, line 49: Your proposed change looks OK to me, but as it appears that `m_Factor` is only used inside `operator()`, I think it should just be a local (`static constexpr`) variable of `operator()`.

Modules/Core/ImageFunction/include/itkWindowedSincInterpolateImageFunction.h
```diff
@@ -141,10 +141,10 @@ class ITK_TEMPLATE_EXPORT BlackmanWindowFunction
 
 private:
   /** Equal to \f$ \frac{\pi}{m} \f$ */
-  static const double m_Factor1;
+  static constexpr double m_Factor1 = itk::Math::pi / VRadius;
 
   /** Equal to \f$ \frac{2 \pi}{m} \f$  */
-  static const double m_Factor2;
+  static constexpr double m_Factor2 = 2.0 * itk::Math::pi / VRadius;
```
N-Dekker, line 147: Same here: your proposed change is OK to me, but I think these factors should be made local to the corresponding `operator()`.

dzenanz (approved):

Modules/Core/ImageFunction/include/itkWindowedSincInterpolateImageFunction.h
hjmjohnson: Done.

Modules/Core/ImageFunction/include/itkWindowedSincInterpolateImageFunction.h
hjmjohnson: done.

Issue #4243, opened 2023-10-01
Add a GHA workflow to perform static code analysis


jhlegarreta:
### Description

Add a GHA workflow to perform static code analysis, using e.g. `cppcheck` or clang tools.

### Impact analysis

A static code analysis tools would allow to e.g. identify dangerous patterns, unreachable code, etc.

### Expected behavior

Improve the code base by passing all static syntax checks.

### Actual behavior

No static analysis is performed other than the analysis done locally by some maintainers and contributors.

### Versions

`master`.

### Environment

Any.

### Additional Information

-

PR #5116, opened 2025-01-03
Fixed various issues found by cppcheck 2.16.0

seanm:
Over the Christmas holidays, I played with pointing cppcheck at ITK.

There are already lots of commits here, so although I'm not finished, I thought I'd share this work so far.

### DIFF
--- a/Modules/Core/Common/include/itkConceptChecking.h
+++ b/Modules/Core/Common/include/itkConceptChecking.h
@@ -702,6 +702,7 @@ struct HasNumericTraits
     void
     constraints()
     {
+      // This is here to cause compile-time failures, if a provided custom type is missing some methods.
       Detail::UniqueType<typename NumericTraits<T>::ValueType>();
       Detail::UniqueType<typename NumericTraits<T>::PrintType>();
       Detail::UniqueType<typename NumericTraits<T>::AbsType>();
@@ -709,6 +710,7 @@ struct HasNumericTraits
       Detail::UniqueType<typename NumericTraits<T>::RealType>();
       Detail::UniqueType<typename NumericTraits<T>::ScalarRealType>();
       Detail::UniqueType<typename NumericTraits<T>::FloatType>();
+
       T a{};
 
       // Test these methods that take an instance of T to
--- a/Modules/Core/Common/include/itkNumericTraits.h
+++ b/Modules/Core/Common/include/itkNumericTraits.h
@@ -308,14 +308,14 @@ public:
     return !val;
   }
   static constexpr bool
-  IsNegative(bool val)
+  IsNegative(bool itkNotUsed(val))
   {
-    return val ? false : false;
+    return false;
   }
   static constexpr bool
-  IsNonnegative(bool val)
+  IsNonnegative(bool itkNotUsed(val))
   {
-    return val ? true : true;
+    return true;
   }
   static constexpr bool IsSigned = false;
   static constexpr bool IsInteger = true;
@@ -640,14 +640,14 @@ public:
     return val == Zero;
   }
   static constexpr bool
-  IsNegative(unsigned char val)
+  IsNegative(unsigned char itkNotUsed(val))
   {
-    return val ? false : false;
+    return false;
   }
   static constexpr bool
-  IsNonnegative(unsigned char val)
+  IsNonnegative(unsigned char itkNotUsed(val))
   {
-    return val ? true : true;
+    return true;
   }
   static constexpr bool IsSigned = false;
   static constexpr bool IsInteger = true;
@@ -845,14 +845,14 @@ public:
     return val == Zero;
   }
   static constexpr bool
-  IsNegative(unsigned short val)
+  IsNegative(unsigned short itkNotUsed(val))
   {
-    return val ? false : false;
+    return false;
   }
   static constexpr bool
-  IsNonnegative(unsigned short val)
+  IsNonnegative(unsigned short itkNotUsed(val))
   {
-    return val ? true : true;
+    return true;
   }
   static constexpr bool IsSigned = false;
   static constexpr bool IsInteger = true;
@@ -1069,14 +1069,14 @@ public:
     return val == Zero;
   }
   static constexpr bool
-  IsNegative(unsigned int val)
+  IsNegative(unsigned int itkNotUsed(val))
   {
-    return val ? false : false;
+    return false;
   }
   static constexpr bool
-  IsNonnegative(unsigned int val)
+  IsNonnegative(unsigned int itkNotUsed(val))
   {
-    return val ? true : true;
+    return true;
   }
   static constexpr bool IsSigned = false;
   static constexpr bool IsInteger = true;
--- a/Modules/Core/Common/include/itkResourceProbe.hxx
+++ b/Modules/Core/Common/include/itkResourceProbe.hxx
@@ -125,7 +125,7 @@ template <typename ValueType, typename MeanType>
 void
 ResourceProbe<ValueType, MeanType>::Start()
 {
-  this->m_NumberOfStarts++;
+  ++(this->m_NumberOfStarts);
   this->m_StartValue = this->GetInstantValue();
 }
 
@@ -143,7 +143,7 @@ ResourceProbe<ValueType, MeanType>::Stop()
   this->UpdateMinimumMaximumMeasuredValue(probevalue);
   this->m_TotalValue += probevalue;
   this->m_ProbeValueList.push_back(probevalue);
-  this->m_NumberOfStops++;
+  ++(this->m_NumberOfStops);
   this->m_NumberOfIteration = static_cast<CountType>(this->m_ProbeValueList.size());
 }
 
--- a/Modules/Core/Common/include/itkSmapsFileParser.h
+++ b/Modules/Core/Common/include/itkSmapsFileParser.h
@@ -291,14 +291,14 @@ template <typename TSmapsDataType>
 class ITK_TEMPLATE_EXPORT SmapsFileParser : public MapFileParser<TSmapsDataType>
 {
 public:
-  virtual ~SmapsFileParser() = default;
+  virtual ~SmapsFileParser() override = default;
   /** Load and parse the smaps file pointed by smapsFileLocation.
    *  If smapsFileLocation is empty, load the file located at
    *  "/proc/" + PID + "/smaps"
    *  Throw an exception is the file can't be opened.
    */
   virtual void
-  ReadFile(const std::string & mapFileLocation = "");
+  ReadFile(const std::string & mapFileLocation = "") override;
 };
 
 /** \class VMMapFileParser
@@ -310,12 +310,12 @@ template <typename TVMMapDataType>
 class ITK_TEMPLATE_EXPORT VMMapFileParser : public MapFileParser<TVMMapDataType>
 {
 public:
-  virtual ~VMMapFileParser() = default;
+  virtual ~VMMapFileParser() override = default;
   /** If no vmmap file, create one using "vmmap pid" command
    *  Throw an exception is the file can't be created/opened.
    */
   virtual void
-  ReadFile(const std::string & mapFileLocation = "");
+  ReadFile(const std::string & mapFileLocation = "") override;
 };
 } // end namespace itk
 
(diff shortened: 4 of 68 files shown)


Modules/Core/Common/include/itkConceptChecking.h
```diff
@@ -702,13 +702,6 @@ struct HasNumericTraits
     void
     constraints()
     {
-      Detail::UniqueType<typename NumericTraits<T>::ValueType>();
```
dzenanz, line 705: This is here to cause compile-time failures, if a provided custom type is missing some methods.

Modules/Core/ImageFunction/include/itkLinearInterpolateImageFunction.hxx
```diff
@@ -51,8 +51,6 @@ LinearInterpolateImageFunction<TInputImage, TCoordinate>::EvaluateUnoptimized(co
 
   // When RealType is VariableLengthVector, 'value' will be resized properly
   // below when it's assigned again.
-  Concept::Detail::UniqueType<typename NumericTraits<RealType>::ScalarRealType>();
```
dzenanz, line 54: I am not sure what is this supposed to assert, so I don't know whether it is fine to remove it. This seems to be the only instance of its use?

Modules/Core/Mesh/test/itkParametricSpaceToImageSpaceMeshFilterTest.cxx
```diff
@@ -29,7 +29,6 @@ struct helper
 template <unsigned int VDimension>
 struct helper<itk::Index<VDimension>>
 {
-  static constexpr unsigned int Dimension = VDimension;
```
dzenanz, line 32: The rest of these removals are fine.

dzenanz (commented):
Mostly looks good. Some requests inline.

thewtex:
Thanks @seanm !

Modules/Core/Common/include/itkConceptChecking.h
seanm: Restored with comment.

dzenanz (approved):

Modules/Core/Common/include/itkHexahedronCell.hxx
```diff
@@ -452,7 +452,7 @@ HexahedronCell<TCellInterface>::EvaluatePosition(CoordinateType *          x,
     else if ((itk::Math::abs(pcoords[0]) > ITK_DIVERGED) || (itk::Math::abs(pcoords[1]) > ITK_DIVERGED) ||
              (itk::Math::abs(pcoords[2]) > ITK_DIVERGED))
     {
-      return -1;
+      return false;
```
N-Dekker, line 455: This looks like a major bug fix, thanks! Did it not cause any test failures?

Would the return value of `EvaluatePosition` of both "Cell types" be important to any one? If so, maybe it should be separate bug report + pull request.

seanm: I confess I did not dig deep.  I've just stuck an assert(0) in the 3 places the comment "(S.Hirschberg 11.12.2001)" occurs, and no test enters any of those 3 branches. So there is no test coverage for this change.

I probably should amend the commit message to say that behaviour is changing here, from returning true to returning false. Hopefully this swap is correct. I chose false because the 3rd place already has false.

I made this change as a separate commit, I could make a separate PR, but not sure what benefit that gives...

seanm: OK I've dropped the commit.

N-Dekker: Thanks @seanm Sorry I don't want to discourage you. Finding bugs like this is really a major benefit from running code analysis tools, like you're doing. But then it's not always obvious how to deal with such bugs. Sometimes it's good enough to just make a note about them, without fixing them directly.  Ideally you would create a bug report, a unit test and a fix, but that's not always worth the effort 🤷

Modules/Core/Common/include/itkSmapsFileParser.h
```diff
@@ -291,14 +291,14 @@ template <typename TSmapsDataType>
 class ITK_TEMPLATE_EXPORT SmapsFileParser : public MapFileParser<TSmapsDataType>
 {
 public:
-  virtual ~SmapsFileParser() = default;
+  virtual ~SmapsFileParser() override = default;
```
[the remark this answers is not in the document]
seanm: Not so remarkable, no software is perfect. I guess cppcheck found a few cases clang-tidy missed.

cppcheck does not do any automated code changes, it was all hand edits by me reading warning messages.

Modules/Core/Common/include/itkResourceProbe.hxx
```diff
@@ -143,7 +143,7 @@ ResourceProbe<ValueType, MeanType>::Stop()
   this->UpdateMinimumMaximumMeasuredValue(probevalue);
   this->m_TotalValue += probevalue;
   this->m_ProbeValueList.push_back(probevalue);
-  this->m_NumberOfStops++;
+  ++(this->m_NumberOfStops);
```
N-Dekker, line 146: OK to me, thanks, but we might eventually consider to just remove this `this->`, because it just points to its very own data member, `ResourceProbe::m_NumberOfStops`. Then we can also remove the parentheses that you're adding here  😃

Modules/Segmentation/MarkovRandomFieldsClassifiers/include/itkRGBGibbsPriorFilter.hxx
```diff
@@ -626,7 +626,7 @@ RGBGibbsPriorFilter<TInputImage, TClassifiedImage>::RegionEraser()
   while (!labelledImageIt.IsAtEnd())
   {
     j = 0;
-    while ((m_Region[i] != valid_region_counter[j]) && (j < k))
+    while ((j < k) && (m_Region[i] != valid_region_counter[j]))
```
N-Dekker, line 629: Cool, good catch 👍

Modules/IO/JPEG2000/src/itkJPEG2000ImageIO.cxx
```diff
@@ -360,10 +360,7 @@ JPEG2000ImageIO::ReadImageInformation()
     this->m_Internal->m_Dinfo = nullptr;
   }
 
-  if (l_image)
-  {
-    opj_image_destroy(l_image);
-  }
+  opj_image_destroy(l_image);
```
N-Dekker, line 363: Cool, looks like the `if (!l_image)` at line 261 is indeed sufficient 👍

seanm: Thanks for going over everything carefully!  I didn't trust cppcheck blindly, but more eyes always helps!

Modules/Filtering/DisplacementField/include/itkTimeVaryingVelocityFieldIntegrationImageFilter.hxx
```diff
@@ -64,11 +64,11 @@ TimeVaryingVelocityFieldIntegrationImageFilter<TTimeVaryingVelocityField,
 {
   const TimeVaryingVelocityFieldType * input = this->GetInput();
   DisplacementFieldType *              output = this->GetOutput();
-  this->m_NumberOfTimePoints = input->GetLargestPossibleRegion().GetSize()[OutputImageDimension];
   if (!input || !output)
   {
     return;
   }
+  this->m_NumberOfTimePoints = input->GetLargestPossibleRegion().GetSize()[OutputImageDimension];
```
N-Dekker, line 71: As you said:
> a slight behaviour change now: `m_NumberOfTimePoints` is no longer set if `input` is provided and `output` is not.

Of course, if you want to be 100% sure that there is no behavior change, you can do, instead:

```cpp
if (input)
{
  this->m_NumberOfTimePoints = input->Get...
} 
if (!input || !output)
{
  return;
}
```

But I agree with you that it probably doesn't matter.

Modules/IO/TransformInsightLegacy/test/itkIOTransformTxtTest.cxx
```diff
@@ -201,7 +201,10 @@ int
 secondTest(const std::string & outputDirectory)
 {
   std::filebuf fb;
-  fb.open((outputDirectory + "IllegalTransform.txt"), std::ios::out);
+  if (!fb.open((outputDirectory + "IllegalTransform.txt"), std::ios::out))
+  {
+    return EXIT_FAILURE;
+  }
   std::ostream os(&fb);
```
N-Dekker: Looks like this code would have been easier with _just_ `std::ofstream` (instead of `std::filebuf` + `std::ostream`) 🤔 

But no problem, your proposed adjustment look OK to me. 👍

Modules/Core/Common/include/itkNumericTraits.h
```diff
@@ -308,14 +308,14 @@ class NumericTraits<bool> : public std::numeric_limits<bool>
     return !val;
   }
   static constexpr bool
-  IsNegative(bool val)
+  IsNegative([[maybe_unused]] bool val)
   {
-    return val ? false : false;
+    return false;
```
N-Dekker, line 313: Good catch, thanks. Minor detail: in this case I would rather not use `maybe_unused`, as it suggests that the argument is _maybe_ unused. Suggesting that it _might_ be used in some scenario's.  However, in this case the argument is _never_ used anymore. So then I would just remove the argument name.

dzenanz: Or use macro `itkNotUsed` instead, e.g. `IsNegative(bool itkNotUsed(val))`

N-Dekker: If you still want to preserve the argument name, you could use `itkNotUsed`:

    IsNegative(bool itkNotUsed(val))

But again, I would just remove the name `val` in this case.

N-Dekker (approved):
Thanks @seanm I leave it up to you if you still want to address my suggestions to remove those `virtual` keywords that became unnecessary when using `override`, and/or to remove those `val` parameter names (instead of `[[maybe_unused]]`). Approved one way or the other :+1:

Modules/Core/Common/include/itkSmapsFileParser.h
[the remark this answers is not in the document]
seanm: Since clang-tidy can auto-remove the virtual keywords, let's leave this PR as-is, someone will run clang-tidy again one day.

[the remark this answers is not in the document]
seanm: I should add: I've seen cppcheck give false positives with this warning. i.e. it says "you should put override here", I do, then it doesn't compile.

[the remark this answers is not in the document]
N-Dekker: @seanm Interesting! Do you still remember where cppcheck gave those false positives? Might be of interest to Daniel Marjamäki, the author of https://github.com/danmar/cppcheck !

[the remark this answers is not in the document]
seanm: I'm building up a cppcheck suppression list for ITK. Reporting bugs to them I'll maybe do after, I've done so in the past. There are a lot of false positives generally in these types of tools of course, it's par for the course.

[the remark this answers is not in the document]
N-Dekker: OK, but I assume that cppcheck has found a corresponding virtual member function in a base class, before it suggests to add `override`, right?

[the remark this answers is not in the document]
seanm: > OK, but I assume that cppcheck has found a corresponding virtual member function in a base class, before it suggests to add `override`, right?

You would think so, but I have not investigated. My goal is to:

1. fix true positives where I can
2. create a suppression file suppressing everything else

This way we can have 0 warnings and maybe someone that knows how CI work can add a cppcheck job.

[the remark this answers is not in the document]
dzenanz: > someone that knows how CI work can add a cppcheck job

Could you turn this into an issue, possibly after you add suppression file?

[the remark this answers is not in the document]
seanm: Searching existing ones, looks similar to: https://github.com/InsightSoftwareConsortium/ITK/issues/4243

PR #5125, opened 2025-01-07
2nd round of cppcheck warning fixes

seanm:


seanm:
@N-Dekker may interest you...

There are lots more of the 'use const references' warnings, but in public API, not sure if that's small enough a change to just do?  It won't affect existing callers after all...

Modules/Filtering/ImageIntensity/include/itkMaskNegatedImageFilter.h
```diff
@@ -93,13 +93,6 @@ class MaskNegatedInput
     return TPixelType{};
   }
 
-  template <typename TValue>
```
dzenanz, line 96: This was probably added to support `VariableLengthVector`s. I would not remove it.

Modules/IO/XML/src/itkDOMNode.cxx
```diff
@@ -569,7 +569,7 @@ DOMNode::Find(const std::string & path)
     if (pos != std::string::npos)
     {
       const std::string s2 = s.substr(pos + 1);
-      s = s.substr(0, pos);
+      s.resize(pos);
```
dzenanz, line 572: Cool refactoring!

Modules/Filtering/ImageIntensity/include/itkMaskNegatedImageFilter.h
seanm: Restored.

dzenanz (approved):
LGTM

Modules/Core/Common/test/itkFixedArrayTest.cxx
```diff
@@ -31,7 +31,7 @@ Set_c_Array(int x[3])
 }
 
 void
-Print_Array(itk::FixedArray<int, 3> x, std::ostream & os)
+Print_Array(const itk::FixedArray<int, 3> & x, std::ostream & os)
```
N-Dekker, line 34: In this particular case, either way is fine by me. `itk::FixedArray<int, 3>` can be copied very quickly, as it is trivially copy-able. So the original pass-by-value does not involve dynamic memory allocation or the execution of a hand-written copy-constructor.

In general, pass-by-value (rather than pass-by-const-reference) has the advantage that it is more thread-safe. With pass-by-const-reference, the caller is responsible for keeping the object alive and unmodified, while the function is running. So pass-by-value is often simpler.

However, when copying is more expensive (as with an arbitrary `std::vector`), it's often beneficial to go for pass-by-const-reference. Then again, sometimes move-semantics is even better...

So as always, each approach has its pro's and con's  🤔

Modules/Core/Common/test/itkMathRoundTest.cxx
```diff
@@ -20,7 +20,7 @@
 #include "itkIndex.h"
 
 bool
-math_test_helper(std::string str, bool test)
+math_test_helper(const std::string & str, bool test)
```
N-Dekker, line 23: In this particular case, I would expect any performance gain from using `const std::string &`, because `math_test_helper` is only called with a "literal" as argument. So if it's just about performance, I would use either `const char *` or (preferably) `std::string_view` as parameter type.

I'm interested to know if cppcheck allows passing `std::string_view` _by value_, because it's generally suggested to pass `std::string_view` by value.

seanm: cppcheck categorizes this under "performance", but this is just test code, so performance isn't an issue anyway.

N-Dekker: OK, so this proposed change is only meant to make cppcheck happy, right?

N-Dekker: - This is cool: https://github.com/danmar/cppcheck/pull/3817

So yes, cppcheck _allows_ passing `std::string_view` by value  😃

Modules/Core/Common/test/itkFixedArrayTest.cxx
N-Dekker: Once we would have cppcheck on the CI, would it still be possible to _allow_ pass-by-value (rather than pass-by-const-reference) in specific cases? For object types that can be copied very quickly. Or for specific functions?

Because as I wrote before, there _are_ cases where pass-by-value is preferable.

Modules/Core/Common/test/itkMathRoundTest.cxx
seanm: > OK, so this proposed change is only meant to make cppcheck happy, right?

Yes. :)

Modules/Core/Common/test/itkFixedArrayTest.cxx
seanm: I don't think it warns about every single pass-by-value under the sun, only for some classes where it usually makes sense, like std::string.

N-Dekker: Indeed I see now that cppcheck does not warn for every pass-by-value (under the 🌞). Looks like it does not warn when you pass an empty struct by value.  However, it does warn for `itk::FixedArray`. Looks like it even warns for very small ones, for example  `itk::FixedArray<int, 1>` and `itk::FixedArray<char, 1>`. While passing those two by-value is probably _faster_ than passing them by reference!

I'm not opposed to this specific code change, it's just a test anyway. But I just doubt if we should _always_ enforce pass-by-const-reference whenever cppcheck tells us to do so.

N-Dekker: FYI, I just posted at https://sourceforge.net/p/cppcheck/discussion/general/thread/cce6de4bfd/ "How to deal with passByValue (CWE: 398) warnings?"

Modules/Core/Common/test/itkMathRoundTest.cxx
seanm: I didn't use `ITK_CPPCHECK_TEST`.  I used `CMAKE_EXPORT_COMPILE_COMMANDS=ON` then invoked with:

`cppcheck --project=compile_commands.json --platform=unspecified --enable=style -q --library=qt --library=posix --library=gnu --library=bsd --library=windows --check-level=exhaustive --suppressions-list=../ITK/ITKcppcheckSuppressions.txt --template='{id},{file}:{line},{severity},{message}' -j22`

seanm: >I just checked: replacing std::string with std::string_view does indeed make cppcheck happy, in this case :)

OK, changed.

Modules/Core/Mesh/test/itkCellInterfaceTest.cxx
```diff
@@ -44,7 +44,7 @@ using CellAutoPointer = CellType::CellAutoPointer;
 
 template <typename TCell>
 int
-TestCellInterface(std::string name, TCell * aCell)
+TestCellInterface(const std::string & name, TCell * aCell)
```
N-Dekker, line 47: Of course, the `name` parameter type of this `TestCellInterface` could also be `string_view`.

Modules/Core/QuadEdgeMesh/test/itkQuadEdgeMeshCellInterfaceTest.cxx
```diff
@@ -66,7 +66,7 @@ class CustomQEPolyVisitor
 // Test the cell interface
 template <typename TCell>
 int
-TestCellInterface(std::string name, TCell * aCell)
+TestCellInterface(const std::string & name, TCell * aCell)
```
N-Dekker, line 69: ...and the `name` parameter of _this_ TestCellInterface may also be `string_view`. (I guess it should also be placed in a different namespace, to avoid name conflicts with the _other_ TestCellInterface function template, but apparently it "just" worked. That's probably beyond the scope of this PR anyway.)

Modules/Core/QuadEdgeMesh/test/itkQuadEdgeMeshCellInterfaceTest.cxx
```diff
@@ -174,7 +174,7 @@ TestCellInterface(std::string name, TCell * aCell)
 
 template <typename TCell>
 int
-TestQECellInterface(std::string name, TCell * aCell)
+TestQECellInterface(const std::string & name, TCell * aCell)
```
N-Dekker, line 177: string_view

N-Dekker (approved):
Approved!  The pass-by-const-references introduced with this PR are OK to me, it's just that I feel that cppcheck is too "pushy" with these passByValue warnings.

I would have preferred a few more string_view, rather than `const std::string &`, but I leave it up to you.

 👍

Modules/Core/QuadEdgeMesh/test/itkQuadEdgeMeshCellInterfaceTest.cxx
seanm: Done.

Modules/Core/QuadEdgeMesh/test/itkQuadEdgeMeshCellInterfaceTest.cxx
seanm: Done.

PR #5128, opened 2025-01-08
BUG: -1 was being returned in a function that returns bool

seanm:
Found by cppcheck.

The nearby comment occurs elsewhere in the codebase, and there it was changed to `return false` so I did the same here.


thewtex (approved):
@seanm fantastic!!

I have seen odd warnings in these files from some GCC versions in some situations for some time -- hopefully this addresses it!

CI failures are unrelated -- `data.kitware.com` was choking.

PR #5129, opened 2025-01-09
BUG: Remove empty tests.

zivy:
Two of the GDCM tests were empty, did not test anything and were always successful (itkGDCMSeriesMissingDicomTagTest.cxx, itkGDCMSeriesReadImageWriteTest.cxx).


blowekamp:
I looked through the history... many changes to this empty file were made... but this is the change that marked the actual test as "legacy" then it was eventually removed.

e3155393905a1fa802531fffa277778fd313117e

Not sure if it did something useful.

dzenanz (approved):
LGTM

seanm (approved):

N-Dekker (approved):
Thanks @zivy Well spotted 👍 

Just one remark: this is not really a bug fix, in my opinion. It's a clean-up action, which is a style improvement. (When an ITK user is interested to see the latest bug fixes that may affect their code, this is not one of them.)

So I would prefer a commit message subject line line:

    STYLE: Remove empty GDCM tests

(I would also add the term "GDCM", to make the subject line a bit more specific.) See also the section "Commit Messages" in https://docs.itk.org/en/latest/contributing/index.html#commit-messages

zivy:
@N-Dekker Just changed the commit message subject line and force pushed. Surprisingly the PR subject line on GitHub remains the same after this change.

dzenanz (approved):
There is a difference between PR name, PR description, and commit message(s). PR name/description is taken from the first commit when it is opened, and does not get updated with additional (force) pushes.

Issue #3662, opened 2022-09-26
Save scientific work citations to a BibTeX file for ease of use in Doxygen


jhlegarreta:
### Description

Citations to scientific works should be stored in a BibTeX file and Doxygen's `\cite` should be used to cite them in the documentation. Currently, there is a variety of syntaxes that are used to cite works across the documentation, ranging from no syntax highlight at all, code-style indentation, or paragraph, e.g.:
https://itk.org/Doxygen/html/classitk_1_1STAPLEImageFilter.html
https://itk.org/Doxygen/html/classitk_1_1ScalarChanAndVeseSparseLevelSetImageFilter.html
https://itk.org/Doxygen/html/classitk_1_1BSplineScatteredDataPointSetToImageFilter.html

### Expected information

Citatons should be hosted in a single BibTeX file so that they can easily be cross-referenced and reused across the documentation.

### Actual information

Citation styles are inconsistent and the same work might be cited differently/citations cannot be re-used.

### Versions

`master`

### Additional Information

Doxygen `\cite` command documentation:
https://doxygen.nl/manual/commands.html#cmdcite

The IJ works' bibtex files can actually be retrieved from the IJ website, so that would save some work. Maybe an script that gathers all references could be developed to have them all in a single file. The file could be dynamically updated e.g. every month.

IJ repository: https://github.com/InsightSoftwareConsortium/InsightJournal

zivy:
A limitation of the html version generated bibliography is that the style is fixed and even when the doi is available as part of the bib file it is not included in the output. This functionality is a [longstanding feature request](https://github.com/doxygen/doxygen/issues/5450) in doxygen.

jhlegarreta:
I see. Anyways, I believe consistency is important, so even if the DOI is not displayed by Doxygen, having all references in a BibTeX file would still be an improvement: easier to maintain, consistent style when being displayed, etc. Maybe the Doxygen folks can afford addressing the DOI issue at some point in the future.

zivy:
Agreed. Using the bibtex format forces us to provide more complete and consistent citation information. Some of the existing textual references were partial and ambiguous, required some searching to pinpoint the referenced paper.

albert-github:
One thing I saw is that when the "doi" is in the url there is a link to the "doi" see e.g. https://doc.cgal.org/latest/Arrangement_on_surface_2/citelist.html (ind the link is forwarded again to springer) where the bib file entry is:
```
[...]

PR #5149, opened 2025-01-23
Converting from raw in code citations to bibtex (1/N).

zivy:
First PR out of N, converting raw inline citations with various styles to using a bibliography database and the doxygen cite command. Will take multiple PRs to fully complete #3662.


zivy:
@dave3d is there a way to have the spellchecker ignore the bibliography tags which are used by the doxygen \cite command? Other than listing all the tags in the [itk_dict.txt](https://github.com/InsightSoftwareConsortium/ITK/blob/master/.github/workflows/itk_dict.txt) file.

dave3d:
> @dave3d is there a way to have the spellchecker ignore the bibliography tags which are used by the doxygen \cite command? Other than listing all the tags in the [itk_dict.txt](https://github.com/InsightSoftwareConsortium/ITK/blob/master/.github/workflows/itk_dict.txt) file.

So do you mean don't spell check the word after the "\cite" command?  I'll have a look at it.

zivy:
Yes. If there's the "\cite" directive ignore the next word which is the bibtex tag for the reference. Otherwise we'll have to add these exceptions for every reference because the tags are arbitrary and generally not "real" words. Thanks.

dzenanz:
Already great progress!

jhlegarreta (commented):
Looks great :heart: :rocket:. Thanks for doing this @zivy.

Two minor comments:
- Remove the dot from the subject message.
- Use the imperative mode.

Some comments that would require more work, as consistency would be important in this file:
- Use the same style for names, e.g. First Name Last Name and etc.
- Use brackets consistency for the field values. Use double brackets for all titles so that the cases are respected.
- Do not add whitespaces, indentations or add them systematically and consistently.
- Add first all articles, the books, then bookchapters, etc.: alphabetically.
- Sort the entries alphabetically within each group.
- Sort the field for each entry type the same way.
- Define and document the keys, e.g. lastnameYYYY and how collisions are resolved.

I know this is a lot of work, but the earlier is done, the easier it is to do for the upcoming cases as it sets a clear example.

Have you used some tool to gather this, if yes, please describe it in the commit message.

Is it still a draft or is it ready to be merged according to you @zivy  ?

hjmjohnson:
@zivy  Very nice. Perhaps just bulk adding these citations to the itk dictionary would be the easiest.

zivy:
@jhlegarreta I modified the commit message per your request (first two points). I don't have a style preference and all the variations found in the bib file were obtained from the online source of each of the publications, so there is inherent variability. I think this is ready for merge, but it won't pass the spellchecking due to the use of bibtex tags.

Sadly this is a tedious manual process because the inline citations are free-form, sometimes incomplete and sometimes partially incorrect (detecting all of them by assuming the year will be listed so searching for what looks like a year in all *.h files).

@hjmjohnson Will do a bulk dictionary update if @dave3d doesn't add this functionality in the next couple of days.

For now I think this PR can be merged.

jhlegarreta:
> @jhlegarreta I modified the commit message per your request (first two points). I don't have a style preference and all the variations found in the bib file were obtained from the online source of each of the publications, so there is inherent variability. I think this is ready for merge, but it won't pass the spellchecking due to the use of bibtex tags.

Sadly this is a tedious manual process because the inline citations are free-form, sometimes incomplete and sometimes partially incorrect (detecting all of them by assuming the year will be listed so searching for what looks like a year in all *.h files).

I know. I understand. Thanks for this work @zivy :100:.

jhlegarreta:
Sorry, closed inadvertently.

zivy:
@jhlegarreta It's been a long day. I am stepping away from the computer before I do more harm than good 😪

dzenanz (approved):

zivy:
Some additional updates:
1. Per @albert-github's [comment](https://github.com/InsightSoftwareConsortium/ITK/issues/3662#issuecomment-2614421337), populated the url field in the bibtex file using the doi information. This enables hyperlink creation using the default/current doxygen setup.
2.  Per @jhlegarreta comment above, used the [bibtex-tidy](https://github.com/FlamingTempura/bibtex-tidy) tool, the bibliography file has a more consistent look-and-feel (even though this doesn't matter in terms of functionality). Not added as a pre-commit hook because it [currently does not support](https://github.com/FlamingTempura/bibtex-tidy/issues/388) a check only setup.

jhlegarreta:
> Per @jhlegarreta comment above, used the bibtex-tidy tool, the bibliography file has a more consistent look-and-feel (even though this doesn't matter in terms of functionality). Not added as a pre-commit hook because it currently does not support a check only setup.

Simply great @zivy :100:. Thanks for doing that !

Utilities/Doxygen/DoxygenConfig.cmake
```diff
@@ -164,6 +164,7 @@ set(DOXYGEN_GRAPHICAL_HIERARCHY "NO")
 set(DOXYGEN_DOT_IMAGE_FORMAT "svg")
 set(DOXYGEN_DOT_GRAPH_MAX_NODES "150")
 set(DOXYGEN_DOT_MULTI_TARGETS "YES")
+set(DOXYGEN_CITE_BIB_FILES "${ITK_SOURCE_DIR}/Documentation/Doxygen/doxygen.bib")
```
dzenanz, line 167: This line should probably be added to https://github.com/InsightSoftwareConsortium/ITK/blob/master/Wrapping/DoxygenConfig.cmake too.

.github/workflows/spell-check-comments.yml
```diff
@@ -29,6 +29,7 @@ jobs:
           --miss \
           --prefix "gdcm" --exclude ThirdParty \
           --dict additional_dictionary.txt \
+          --bibtex Documentation/Doxygen/doxygen.bib \
```
dzenanz, line 32: I guess this was the "magic" needed to avoid the spell check errors?

dzenanz:
This caused a few dozen new [warnings](https://open.cdash.org/viewBuildError.php?type=1&buildid=10182794) of the style: Parsing file `Modules/Registration/Metricsv4/include/itkMattesMutualInformationImageToImageMetricv4.h:44: warning: \cite command has no label` and `Modules/Registration/Common/include/itkMattesMutualInformationImageToImageMetric.h:61: warning: unable to resolve reference to 'mattes2001' for \cite command`

albert-github:
I already made one issue #5202 with a proposed patch and I'm currently looking into the other problems.

albert-github:
I've also pushed a proposed patch for future doxygen versions: https://github.com/doxygen/doxygen/pull/11384

Issue #3703, opened 2022-10-22
IO ImageBase tests not exercised


jhlegarreta:
### Description

The tests [`itkLargeImageWriteConvertReadTest`](https://github.com/InsightSoftwareConsortium/ITK/blob/23da3dfa79152648f3d38050ee147df47f41ce0a/Modules/IO/ImageBase/test/itkLargeImageWriteConvertReadTest.cxx) and [`itkLargeImageWriteReadTest`](https://github.com/InsightSoftwareConsortium/ITK/blob/23da3dfa79152648f3d38050ee147df47f41ce0a/Modules/IO/ImageBase/test/itkLargeImageWriteReadTest.cxx) are not being exercised as the corresponding `itk_add_test` command are not present in the relevant [CMakeLists.txt](
https://github.com/InsightSoftwareConsortium/ITK/blob/23da3dfa79152648f3d38050ee147df47f41ce0a/Modules/IO/ImageBase/test/CMakeLists.txt) file, despite the files being listed:
https://github.com/InsightSoftwareConsortium/ITK/blob/23da3dfa79152648f3d38050ee147df47f41ce0a/Modules/IO/ImageBase/test/CMakeLists.txt#L12
and
https://github.com/InsightSoftwareConsortium/ITK/blob/23da3dfa79152648f3d38050ee147df47f41ce0a/Modules/IO/ImageBase/test/CMakeLists.txt#L13

Thus, the corresponding executables are not created.

### Steps to Reproduce

1. Get ITK
2. Configure and build
3. Execute
```
ctest -V -R itkLargeImageWriteConvertReadTest -C Debug
```
and
```
ctest -V -R itkLargeImageWriteReadTest-C Debug
```

CTest will report the message:
```
No tests were found!!!
```

### Expected behavior

All listed tests should be exercised.

### Actual behavior

Tests are not being exercised.

### Reproducibility

100%.

### Versions

`master`.

### Environment

Any.

### Additional Information

None

dzenanz:
[...]

PR #5151, opened 2025-01-23
TEST: Fix unexercised tests for large image read/write

jadh4v:
## PR Checklist

Closes: https://github.com/InsightSoftwareConsortium/ITK/issues/3703

### DIFF
--- a/Modules/IO/ImageBase/test/CMakeLists.txt
+++ b/Modules/IO/ImageBase/test/CMakeLists.txt
@@ -40,6 +40,50 @@ set(ITKIOImageBaseTests
     itkImageFileReaderManyComponentVectorTest.cxx)
 
 createtestdriver(ITKIOImageBase "${ITKIOImageBase-Test_LIBRARIES}" "${ITKIOImageBaseTests}")
+
+if("${ITK_COMPUTER_MEMORY_SIZE}" GREATER 16)
+  itk_add_test(
+    NAME
+    itkLargeImageWriteConvertReadTest
+    COMMAND
+    ITKIOImageBaseTestDriver
+    itkLargeImageWriteConvertReadTest
+    ${ITK_TEST_OUTPUT_DIR}/itkLargeImageWriteConvertReadTest.mha
+    30000L)
+  set_tests_properties(itkLargeImageWriteConvertReadTest
+    PROPERTIES
+    RESOURCE_LOCK
+    MEMORY_SIZE)
+
+  itk_add_test(
+    NAME
+    itkLargeImageWriteReadTest_2D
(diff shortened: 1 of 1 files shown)


dzenanz:
ITK doesn't recognize `TEST: ` prefix, so you should use `ENH: ` in the commit message. Otherwise looks good.

dzenanz (approved):

jadh4v:
> ITK doesn't recognize `TEST: ` prefix, so you should use `ENH: ` in the commit message. Otherwise looks good.

Updated the commit msg to use ENH

Modules/IO/ImageBase/test/CMakeLists.txt
```diff
@@ -40,6 +40,35 @@ set(ITKIOImageBaseTests
...
+    30000L)
+  itk_add_test(
+    NAME
+    itkLargeImageWriteReadTest_3D
+    COMMAND
+    ITKIOImageBaseTestDriver
+    itkLargeImageWriteReadTest
+    ${ITK_TEST_OUTPUT_DIR}/itkLargeImageWriteReadTest_3D.mha
+    30000L
+    4L)
```
blowekamp, line 69: It looks like each of these tests needs 5 GB of memory and can not be run concurrently on memory constrained system. Please add something like the following for each memory intensive test:
```
   set_tests_properties(itkComposeBigVectorImageFilterTest
            PROPERTIES
            RESOURCE_LOCK
            MEMORY_SIZE)
```
This will ensure only one test with "MEMORY_SIZE" resource constrained is run at a time.

dzenanz (approved):

dzenanz:
CI runs into: `Error: No space left on device`. We should maybe raise the memory available limit so that these tests do not run in the CI.

dzenanz:
Maybe a better solution would be to remove temporary files (at least for tests creating large temporaries) immediately after the test is successfully finished. But this is a lot of additional work, and could be its own issue.

jadh4v:
> No space left on device

I have increased the memory limit to 16GB. I think this should suffice for now? Removing the temp files, as you said can be a separate issue to track and could be applied across all tests.

jhlegarreta:
> Maybe a better solution would be to remove temporary files (at least for tests creating large temporaries) immediately after the test is successfully finished. But this is a lot of additional work, and could be its own issue.

It `ctest` supported this option, it would avoid the additional work.

dzenanz:
@bradking Does CTest support [this](https://github.com/InsightSoftwareConsortium/ITK/pull/5151#issuecomment-2613586621)? I suspect this would be interesting to other projects besides ITK.

PR #5154, opened 2025-01-23
BUG: SAGITAL_PLANE -> SAGITTAL_PLANE, though with legacy support

### LINKED ISSUE #3923: Identifier `SAGITAL_PLANE` is misspelled
The identifier `SAGITAL_PLANE` might also be misspelled 🤔 That was beyond the scope of this pull request:

_Originally posted by @N-Dekker in https://github.com/InsightSoftwareConsortium/ITK/pull/3921#discussion_r1103808432_

Leengit:
Closes #3923.

Replace the misspelling `SAGITAL_PLANE` with `SAGITTAL_PLANE`.

Note that both spellings are accepted if _not_ `ITK_LEGACY_REMOVE`. 

However, invocation of `std::ostream &operator<<(std::ostream & out, const OctreeEnums::Octree value)` with either spelling will output a string with the correct spelling: `"itk::OctreeEnums::Octree::SAGITTAL_PLANE"`

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkCommonEnums.h
+++ b/Modules/Core/Common/include/itkCommonEnums.h
@@ -212,7 +212,10 @@ public:
   enum class Octree : uint8_t
   {
     UNKNOWN_PLANE,   ///< The plane is Unknown
-    SAGITAL_PLANE,   ///< The plane is Sagittal
+    SAGITTAL_PLANE,  ///< The plane is Sagittal
+#if !defined(ITK_LEGACY_REMOVE)
+    SAGITAL_PLANE = SAGITTAL_PLANE, ///< Support misspelling
+#endif
     CORONAL_PLANE,   ///< The plane is Coronal
     TRANSVERSE_PLANE ///< The plane is Transverse
   };
--- a/Modules/Core/Common/include/itkCommonEnums.h
+++ b/Modules/Core/Common/include/itkCommonEnums.h
@@ -214,7 +214,7 @@ public:
     UNKNOWN_PLANE,   ///< The plane is Unknown
     SAGITTAL_PLANE,  ///< The plane is Sagittal
 #if !defined(ITK_LEGACY_REMOVE)
-    SAGITAL_PLANE = SAGITTAL_PLANE, ///< Support misspelling
+    SAGITAL_PLANE [[deprecated("Use SAGITTAL_PLANE instead")]] = SAGITTAL_PLANE, ///< Support misspelling
 #endif
     CORONAL_PLANE,   ///< The plane is Coronal
     TRANSVERSE_PLANE ///< The plane is Transverse
--- a/Modules/Core/Common/include/itkOctree.h
+++ b/Modules/Core/Common/include/itkOctree.h
@@ -50,6 +50,7 @@ public:
   // We need to expose the enum values at the class level
   // for backwards compatibility
   static constexpr OctreeEnum UNKNOWN_PLANE = OctreeEnum::UNKNOWN_PLANE;
+  static constexpr OctreeEnum SAGITTAL_PLANE = OctreeEnum::SAGITTAL_PLANE;
   static constexpr OctreeEnum SAGITAL_PLANE = OctreeEnum::SAGITAL_PLANE;
   static constexpr OctreeEnum CORONAL_PLANE = OctreeEnum::CORONAL_PLANE;
   static constexpr OctreeEnum TRANSVERSE_PLANE = OctreeEnum::TRANSVERSE_PLANE;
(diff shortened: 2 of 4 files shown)


dzenanz (approved):

Modules/Core/Common/include/itkCommonEnums.h
```diff
@@ -212,7 +212,10 @@ class OctreeEnums
   enum class Octree : uint8_t
   {
     UNKNOWN_PLANE,   ///< The plane is Unknown
-    SAGITAL_PLANE,   ///< The plane is Sagittal
+    SAGITTAL_PLANE,  ///< The plane is Sagittal
+#if !defined(ITK_LEGACY_REMOVE)
+    SAGITAL_PLANE = SAGITTAL_PLANE, ///< Support misspelling
```
blowekamp, line 217: There is a [[deprecated]] attribute that can be used on enums, and may be appropriate here.

Leengit: I added it.

dzenanz (approved):

hjmjohnson:
There is a dashboard issue that might be tricky to work around: 
 https://open.cdash.org/viewBuildError.php?type=1&buildid=10163799

dzenanz:
Could we fix this via warning suppression mechanism?

Leengit:
The dashboard is showing the warnings for deprecation when we haven't set `ITK_LEGACY_REMOVE`... we allow the deprecated spelling and get the warning.  Should I removed `[[deprecated]]` under the theory that the user didn't set `ITK_LEGACY_REMOVE` and therefore doesn't want the deprecation warnings?

dzenanz:
> Could we fix this via warning suppression mechanism?

I found an example of this:
https://github.com/InsightSoftwareConsortium/ITK/blob/524df2a9d1e8990743784afc971fabccce5b29d0/.circleci/config.yml#L74

dzenanz:
> remove [[deprecated]]

This is the easiest solution. @blowekamp what do you think?

blowekamp:
> Could we fix this via warning suppression mechanism?

The `ITK_FUTURE_DEPRECATED` may work here.

Recently in  itkAnatomicalOrientation I also had to add an ITK_WRAPPING. That may need to be added to the itkMacro.
https://github.com/InsightSoftwareConsortium/ITK/blob/5a6d1e9bf8e865acc7932793bb2933948cc1da13/Modules/Core/Common/include/itkAnatomicalOrientation.h#L436-L445

dzenanz:
> added to the itkMacro

If you add it to `itkMacro.h`, and change the `AnatomicalOrientation` to show how the new macro is supposed to be used, I am convinced that Lee can use it in this PR.

blowekamp:
> > added to the itkMacro
> 
> If you add it to `itkMacro.h`, and change the `AnatomicalOrientation` to show how the new macro is supposed to be used, I am convinced that Lee can use it in this PR.

@Leengit Please give the macro defined here a try: https://github.com/InsightSoftwareConsortium/ITK/blob/1090d31e0360022b4414d0c3cbf27927a7d55901/Modules/Core/Common/include/itkMacro.h#L1383-L1389

If warning persists with wrapping, then the macro may need to be modified to handle when ITK_WRAPPING is defined.

Leengit:
I am thinking that the optimal behavior is

1. if `ITK_LEGACY_REMOVE` IS NOT defined then the misspelling is available.  The misspelling should not produce a warning when building 3D Slicer itself, but it would be nice to have a [[deprecated]] warning in code that invokes 3D Slicer.
2. if `ITK_LEGACY_REMOVE` IS defined then the misspelling is not available.  It is a fatal error to use the misspelling.

It looks like use of `ITK_FUTURE_DEPRECATED` would
3. allow the misspelling even though `ITK_LEGACY_REMOVE` IS defined, producing a [[deprecated]] warning, but not a fatal error.

So am I hesitant to use `ITK_FUTURE_DEPRECATED`.  Or is my ideal something I should be thinking about differently?

Leengit:
I just force-pushed changes that I hope have the following behavior.

1. When `ITK_LEGACY_REMOVE` is set then the misspellings are not recognized and cause fatal errors.
2. When `ITK_LEGACY_REMOVE` is not set the misspellings are defined in terms of the proper spellings (and never in terms of other objects that are misspelled) and thus 3D Slicer should build without warnings or errors.
3. When `ITK_LEGACY_REMOVE` is not set the misspellings are marked as `[[deprecated]]`, so other software that uses the misspellings will get a deprecation warning.

Leengit:
As I discuss above, I was able to achieve a balancing of priorities ... but I acknowledge that there are other ways to balance the priorities.  If we should go forward with the current implementation, please approve.  Otherwise, let's continue the discussion here.  Thanks!

And/or if you think the failure of `itkFFTConvolutionImageFilterStreamingTest` is somehow related to this PR, please let me know.

dzenanz (approved):

PR #5162, opened 2025-01-23
STYLE: Add historical bulk changes to git blame ignore

### LINKED ISSUE #2623: Add historical bulk changes to .git-blame-ignore-revs
This file already exists and contains some bulk revisions to the codebase, but it may not have been systematically filled.

ebrahimebrahim:
Close #2623

In case this needs to be re-done in the future, I used the following script in a python environment with GitPython installed:

```py
import git
from pathlib import Path

def find_style_commits_with_large_changes(repo : git.Repo, prefix="STYLE", file_change_threshold=1000) -> list[str]:
    matching_commits = []
    for commit in repo.iter_commits():
        if commit.message.strip().startswith(prefix):
            num_files_changed = len(commit.stats.files)
            if num_files_changed >= file_change_threshold:
                matching_commits.append(commit.hexsha)

    return matching_commits


def get_commits_already_ignored(blame_ignore_revs_path) -> list[str]:
    return [
        line.strip()
        for line in blame_ignore_revs_path.read_text().split('\n')
        if line and not line.startswith('#')
    ]

if __name__ == "__main__":
    repo = git.Repo('.')
    style_commits = find_style_commits_with_large_changes(repo)
    blame_ignore_revs_path = Path("./.git-blame-ignore-revs")
    commits_already_ignored = get_commits_already_ignored(blame_ignore_revs_path)
    commits_to_add = [commit_hash for commit_hash in style_commits if commit_hash not in commits_already_ignored]
    entries_to_add : list[str] = []
    for commit_hash in commits_to_add:
        commit_message_title = repo.commit(commit_hash).message.split('\n')[0]
        entries_to_add.append(f"# {commit_message_title}\n{commit_hash}")
    text_to_add = '\n'.join(entries_to_add)
    with blame_ignore_revs_path.open('a') as file:
        file.write(text_to_add)
```

The `file_change_threshold` was ~~100~~1000, so the commits being added to the git-blame-ignore are those commits that both start with "STYLE" and that affect at least ~~100~~1000 files.

---

## PR Checklist

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -96,3 +96,35 @@ fbbc7aa38cf23d0966b1dbd0f9825797e84c2502
 f3899ce8c6d53fc077c4e0bfa2c6f1fe56166b57
 # STYLE: Initial run of cmake-format
 efa749515db840b54f7329b261d3ec4c3a575f83
+# STYLE: Prefer explicit const designation
+375ee8c47a03792d0d3d5447c6b4c4be6b9f2fb4
+# STYLE: Replace itkTypeMacro calls with `itkOverrideGetNameOfClassMacro`
+2c264ea2ef62e916c6ef947599ce659cc8fce5fe
+# STYLE: Remove recursive inclusion of .h files
+f84720ee0823964bd135de8eb973acc40c1e70e1
+# STYLE: Use `auto` for declaration of variables initialized by `New()`
+828453d1bf61c487310d2d8c9570093e08798a40
+# STYLE: Replace postfix by prefix increment in `for` loops in ITK/Modules
+971aef211cad336ba75cb2da114fc86e873444e5
+# STYLE: Call ITK_DISALLOW_COPY_AND_MOVE, not ITK_DISALLOW_COPY_AND_ASSIGN
+4c0982ddec474b1701d695839c78c7e888659ef4
+# STYLE: Pefer = default to explicitly trivial implementations
+5142ed7bb5e4e5042734fc47bf556a2342be6cec
+# STYLE: Prefer constexpr for const numeric literals
+605cea4735634acdc3222d1e0f1d3fcb2db05e26
+# STYLE: Prefer C++11 type alias over typedef
+66e5d6b3bcc28f1a85b702086b6cedc8cab6723b
+# STYLE: Provide a well documented macro for repeated code
+4a3257e293a9682b2d8b9063f72bddcae98d70dc
+# STYLE: Use "typename" for template parameters
+86bbb9d092d21bb309b0260c0b4082d9af966b67
+# STYLE: ITK_EXPORT had no use
+b788601abbeebd72e5504b10a75a9f417f666c35
+# STYLE:  Removed the MSVS6 compiler pragma 4786
+f1bc82756aaf853c50b6c2b5cfb4dfd88bc4824f
+# STYLE:  Converted module names to use camel case
+540228ef42fb667b3255f8610baa5cd9d7b6760d
+# STYLE: Change for development tool compliance
+13f99f81510fd76b0409b98dac6a4670436e007a
+# STYLE: Change to hxx extension for dev tools
+2fccd272862ab7cc408d643c8db2e1cb5d098fcf


Leengit (approved):
LGTM

N-Dekker:
I'm not sure about the whole git-blame-ignore-revs feature. When you want to find out what particular commit changed the code, why would you want to _always_ ignore commits that affect more than 100 files?

Would it not be possible for such a commit to introduce a significant change? I have seen large commits that include some manual changes, possibly introducing unexpected changes.

ebrahimebrahim:
@N-Dekker It's not about the number of files really, it's more about the fact that they are STYLE changes

> Would it not be possible for such a commit to introduce a significant change? I have seen large commits that include some manual changes, possibly introducing unexpected changes.

By definition if a change is marked as STYLE then the author is expressing that the change is not significant. These changes are largely not relevant to what one is looking for during a blame.

Also, this is all optional because the user can make git blame behave as they wish (`--ignore-revs-file` versus `--no-ignore-revs`), disabling in case one is actually interested in blaming through style changes

N-Dekker:
> It's not about the number of files really, it's more about the fact that they are STYLE changes

@ebrahimebrahim Suppose you want to find out when a particular bug is introduced. In general it might also be introduced by a STYLE commit, right? So you still want to see STYLE commits in a blame.

> this is all optional because the user can make git blame behave as they wish (--ignore-revs-file versus --no-ignore-revs)

Very interesting, thanks. I think I want to use that option: `--no-ignore-revs`! Do you happen to know if it can be enabled for TortoiseGit and the GitHub website?

For example, when I look at https://github.com/InsightSoftwareConsortium/ITK/blame/master/Modules/Core/Common/include/itkObject.h and press `[Blame]`, it says:
> (i) Ignoring revisions in [.git-blame-ignore-revs](https://github.com/InsightSoftwareConsortium/ITK/blob/master/.git-blame-ignore-revs).

In the TortoiseGitBlame settings, I don't see how to switch off ignoring revisions either 🤷

ebrahimebrahim:
> Do you happen to know if it can be enabled for TortoiseGit and the GitHub website?

I'm not familiar with tortoisegit but based on [this issue](https://gitlab.com/tortoisegit/tortoisegit/-/issues/3916) it looks like the `--no-ignore-revs` is the default behavior as they haven't added the "feature" yet -- idk if that info is up to date

As for github I don't see any way to turn it off :/

> In general it might also be introduced by a STYLE commit, right? So you still want to see STYLE commits in a blame.

I guess it comes down to what we feel ought to be the default behavior. If we want the default to be to not ignore style commits then one approach may be to rename the .git-blame-ignore-revs file so that github cannot find it -- then users who want to ignore style commits can explicitly use `--ignore-revs-file` to point to the renamed file. My sense is that the default behavior ought to be to ignore style commits, since they are largely red herrings when seeking out what caused a certain change or broke something, but it's up for debate

.git-blame-ignore-revs
```diff
@@ -96,3 +96,179 @@ fbbc7aa38cf23d0966b1dbd0f9825797e84c2502
 f3899ce8c6d53fc077c4e0bfa2c6f1fe56166b57
 # STYLE: Initial run of cmake-format
 efa749515db840b54f7329b261d3ec4c3a575f83
+# STYLE: Prefer local initialization
+e23a4dbd1d19c3fecd221a25e6f4c5825dc21e00
+# STYLE: Variable can be made constexpr
+e1acaf5aa8d23c82594fd1a7cc9282b3c9786b84
```
N-Dekker, line 102: Sorry for repeating myself, but I believe that a commit that makes the initialization of variables more local (e23a4dbd1d19c3fecd221a25e6f4c5825dc21e00) or adds more constexpr (e1acaf5aa8d23c82594fd1a7cc9282b3c9786b84) might unexpectedly (probably unintendedly) cause some behavior change, and might be relevant when doing a `git blame`. 

I would rathernot ignore anything other than formatting changes (by Python black, clang-format, etc).

dzenanz: That is certainly possible as Ebrahim pointed out, but is usually undesirable. We want to make the default more convenient for most people.

ebrahimebrahim: Hmm maybe the issue is a liberal interpretation of what really counts as a style commit. I agree that those two commits one may well want to include in a blame. I wouldn't have called those "STYLE".

I'll wait to see how if this discussion continues but I'm open to the option renaming `.git-blame-ignore-revs` to something that wouldn't be used by default by github's blame.

Leengit: Perhaps remove those two commits from the list, under the theory that they shouldn't have been labeled as STYLE?

N-Dekker: > We want to make the default more convenient for most people.

Why do people use `git blame` in the first place? I typically use it to find exactly when a specific feature or bug was introduced. A bug might also be introduced, accidentally, by a STYLE commit. My use case typically involves multiple "blame prior to change XXX/blame previous version" steps, iteratively, to find the relevant code change. 

Of course, I may not be like "most people", I don't know. 🤷

dzenanz: The problem is, with a mature codebase like ITK, a lot of the steps in git blame tracing can be style commits. And when something is hard, no one likes doing it. Style commits are, by definition, not likely to cause problems, but modify a lot of files, and therefore feature in blame tracing prominently. Unless they can be automatically ignored.

thewtex:
I think 100 is too small here -- 1000 files is closer to a "bulk change" in a project as large as ITK.

N-Dekker:
> Suppose I switch to Matt's suggested "bulk style changes" threshold of 1000 files instead of 100 files -- would that do for you?

Thank you for asking, @ebrahimebrahim. As I'm not a fan of ignoring commits in the first place, I would certainly prefer a threshold of 1000 files, instead of 100. But of course, any threshold is arbitrary. _If_ I would want to ignore commits at all, I would only want to ignore those that just do source code formatting adjustments (using clang-format, python black, etc).

Having said so, I very much appreciate your effort. 👍  I think I'll leave it up to the other ITK-ers how to proceed on this pull request.

dzenanz:
@ebrahimebrahim Let's raise the limit to 1000, and manually add/exclude the few identified exceptions.

dzenanz (approved):

Issue #4773, opened 2024-07-10
GlobalDefaultNumberOfThreads should not be greater than GlobalMaximumNumberOfThreads


N-Dekker:
### Description

Users typically call `MultiThreaderBase::SetGlobalMaximumNumberOfThreads` when they want to reduce (limit) the number of threads used by a process. However, when a user sets the `GlobalMaximumNumberOfThreads` while leaving the `GlobalDefaultNumberOfThreads` unspecified, this "default number" may end up becoming larger than the "maximum number"!

### Steps to Reproduce

```cpp
#include <itkMultiThreaderBase.h>
#include <iostream>

int main()
{
  itk::MultiThreaderBase::SetGlobalMaximumNumberOfThreads(3);

  std::cout
    << "Maximum: " << itk::MultiThreaderBase::GetGlobalMaximumNumberOfThreads() << '\n'
    << "Default: " << itk::MultiThreaderBase::GetGlobalDefaultNumberOfThreads() << '\n';
}

```
### Expected behavior

With this example, both Maximum and Default should end up being `3` (at least, if the machine supports having 3 simultaneous threads).

### Actual behavior

Output on my (LKEB/[LUMC](https://www.lumc.nl)) multi-core machine:
> Maximum: 3
> Default: 12

### Versions

ITK 5.4.0

### Environment

Windows 

### Additional Information

`GlobalDefaultNumberOfThreads` is being used when constructing the ThreadPool:

https://github.com/InsightSoftwareConsortium/ITK/blob/01e133c6b550786c63fe2886fe4252ecefa19567/Modules/Core/Common/src/itkThreadPool.cxx#L114 

So the number of threads held by the thread pool appears to also occasionally exceed `GlobalMaximumNumberOfThreads`!

---

- Triggered by research by Sebastian (@Svdvoort) (related to ITKElastix issue https://github.com/InsightSoftwareConsortium/ITKElastix/issues/160) and by Gonzalo Mosquera Rojas and Juancito van Leeuwen.

thewtex:
Addressed in #5283

PR #5191, opened 2025-01-27
BUG: fixed test that made bad assumptions of GetGlobalDefaultNumberOf…

seanm:
…Threads()

### DIFF
--- a/Modules/Core/Common/test/itkDataObjectAndProcessObjectTest.cxx
+++ b/Modules/Core/Common/test/itkDataObjectAndProcessObjectTest.cxx
@@ -228,8 +228,8 @@ itkDataObjectAndProcessObjectTest(int, char *[])
   ITK_TEST_SET_GET_VALUE(true, process->GetReleaseDataBeforeUpdateFlag());
 
   ITK_TEST_EXPECT_TRUE(itk::MultiThreaderBase::GetGlobalDefaultNumberOfThreads() <= process->GetNumberOfWorkUnits());
-  process->SetNumberOfWorkUnits(11);
-  ITK_TEST_SET_GET_VALUE(11, process->GetNumberOfWorkUnits());
+  process->SetNumberOfWorkUnits(ITK_MAX_THREADS - 1);
+  ITK_TEST_SET_GET_VALUE(ITK_MAX_THREADS - 1, process->GetNumberOfWorkUnits());
   process->SetNumberOfWorkUnits(0);
   ITK_TEST_SET_GET_VALUE(1, process->GetNumberOfWorkUnits());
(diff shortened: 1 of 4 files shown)


dzenanz (approved):

seanm:
@dzenanz how would I solve the case with the `ITK_TEST_SET_GET_VALUE` macro?  I'm not sure what to use instead...

Modules/Core/Common/test/itkDataObjectAndProcessObjectTest.cxx
```diff
@@ -229,7 +229,7 @@ itkDataObjectAndProcessObjectTest(int, char *[])
 
   ITK_TEST_EXPECT_TRUE(itk::MultiThreaderBase::GetGlobalDefaultNumberOfThreads() <= process->GetNumberOfWorkUnits());
   process->SetNumberOfWorkUnits(11);
-  ITK_TEST_SET_GET_VALUE(11, process->GetNumberOfWorkUnits());
+  ITK_TEST_SET_GET_VALUE(11, process->GetNumberOfWorkUnits()); // same...
```
dzenanz, line 232: Instead of 11, maybe use `ITK_MAX_THREADS / 2`? Or `ITK_MAX_THREADS - 1`?

N-Dekker:
Thanks @seanm !

- For the record, a slightly related issue: #4773

dzenanz (approved):

seanm:
There's one more failing test that I hope to figure out and add here...

seanm:
OK, I don't get it... need help :)

It's this code:

https://github.com/InsightSoftwareConsortium/ITK/blob/5df209ffdd2b351939631219de14c782754691ca/Modules/Core/Common/test/itkMultiThreaderExceptionsTest.cxx#L102

The test fails with `Failed to catch expected exception`.

The call to `dummySrc->SetNumberOfWorkUnits(4)` will be clamped down from 4 to 2 (my `ITK_DEFAULT_MAX_THREADS`), so I thought changing the hardedcoded `for` loop from 4 would fix it, but alas no (though that change maybe necessary too, but insufficient).

dzenanz:
Nothing jumps at me. I guess some digging is needed into `DummyImageSource`.

seanm:
I took a wild guess: there was another `4` in the code. Changing that fixed it.

seanm:
If this looks good to y'all, would be nice to merge, as it will green up cdash a bit.

dzenanz (approved):

seanm:
Seems this broke the build, depending on build options: https://open.cdash.org/viewBuildError.php?buildid=10179364

fix: https://github.com/InsightSoftwareConsortium/ITK/pull/5200

PR #5215, opened 2025-02-05
DOC: Convert from raw in-code citations to bibtex (3/N)

zivy:
Third PR out of N, converting raw inline citations with various styles to using a bibliography database and the doxygen cite command.

Bibtex file was formatted using bibtex-tidy
(https://github.com/FlamingTempura/bibtex-tidy). This tool can be used as a pre-commit hook but currently does not have an option for just performing compliance checking so not added to .pre-commit-config.yaml file.


dzenanz (approved):
Looks good on a glance.

jhlegarreta (approved):
@zivy Thanks for continuing with this monumental effort :100:.

PR #5198, opened 2025-01-29
Move empty member function definitions from "itk*.hxx" to "itk*.h"

N-Dekker:
Moved the member function definitions of member functions that have an empty
parameter list and an empty member function definition from their "itk*.hxx"
file to the corresponding "itk*.h" file, and into their class definition.

Defaulted the default-constructor of `PyImageFilter`.

Found by regular expression `::.+\(\)\r\n{}` in "itk*.hxx" files.

Suggested by Denis Shamonin (@dpshamonin).

PR #5221, opened 2025-02-06
STYLE: Move empty `const` member functions from .hxx to .h

N-Dekker:
Moved the member function definitions of `const`member functions that have an empty member function definition from their "itk*.hxx" file to the corresponding "itk*.h" file, and into their class definition.

Found by regular expression `^..[^:,].+\) const\r\n{}` in "itk*.hxx" files.

- Follow-up to pull request https://github.com/InsightSoftwareConsortium/ITK/pull/5198 commit b94770d8746133a441df1796e3d1628c6935adfc
"STYLE: Move empty member functions with empty param list from .hxx to .h"

### DIFF
--- a/Modules/Registration/Common/include/itkEuclideanDistancePointMetric.h
+++ b/Modules/Registration/Common/include/itkEuclideanDistancePointMetric.h
@@ -90,7 +90,9 @@ public:
 
   /** Get the derivatives of the match measure. */
   void
-  GetDerivative(const TransformParametersType & parameters, DerivativeType & Derivative) const override;
+  GetDerivative(const TransformParametersType & itkNotUsed(parameters),
+                DerivativeType &                itkNotUsed(derivative)) const override
+  {}
 
   /**  Get the match measure, i.e. the value for single valued optimizers. */
   MeasureType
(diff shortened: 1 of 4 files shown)


Modules/Registration/Common/include/itkEuclideanDistancePointMetric.h
```diff
@@ -90,7 +90,9 @@ class ITK_TEMPLATE_EXPORT EuclideanDistancePointMetric
 
   /** Get the derivatives of the match measure. */
   void
-  GetDerivative(const TransformParametersType & parameters, DerivativeType & Derivative) const override;
+  GetDerivative(const TransformParametersType & itkNotUsed(parameters),
+                DerivativeType &                itkNotUsed(derivative)) const override
+  {}
```
N-Dekker, line 95: It appears that `EuclideanDistancePointMetric::GetDerivative` has always been empty, looking at commit 06e94a207e0b9788dc7269720eb8c5ba39155997 (10 Oct 2004):
https://github.com/InsightSoftwareConsortium/ITK/blob/06e94a207e0b9788dc7269720eb8c5ba39155997/Code/Algorithms/itkEuclideanDistancePointMetric.txx#L145-L150

Just curious, was it still meant to be implemented? Or was it intentionally left empty?

PR #5214, opened 2025-02-05
STYLE: Use itk::ReadImage and itk::WriteImage in Example/Statistics-Correction

krupalbhat:
This commit addresses the review comments related to issue #2802.In this commit Examples related to statistics has been changed to use itk::ReadImage and itk::WriteImage.

PR #5163, opened 2025-01-23
STYLE: Use itk::ReadImage and itk::WriteImage in Examples/Segmentations, SpatialObjects and Statistics

sadhana-r:
See #2802.

## PR Checklist
- [x] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [x] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)

PR #5223, opened 2025-02-06
STYLE: Use itk::ReadImage and itk::WriteImage in Example/Statistics

dzenanz:
This commit addresses the review comments from PR #5163 related to issue #2802. In this commit Examples related to statistics has been changed to use itk::ReadImage and itk::WriteImage

Part of #2802. This is a re-imagining of #5214. Closes #5214.

## PR Checklist

### DIFF
--- a/Examples/Statistics/BayesianClassifier.cxx
+++ b/Examples/Statistics/BayesianClassifier.cxx
@@ -67,15 +67,21 @@ main(int argc, char * argv[])
     return EXIT_FAILURE;
   }
 
-  // input parameters
-  const char * membershipImageFileName = argv[1];
-  const char * labelMapImageFileName = argv[2];
 
   // setup reader
   constexpr unsigned int Dimension = 2;
   using InputPixelType = float;
   using InputImageType = itk::VectorImage<InputPixelType, Dimension>;
-  const auto input = itk::ReadImage<InputImageType>(membershipImageFileName);
+  InputImageType::Pointer input;
+  try
+  {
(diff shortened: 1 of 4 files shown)


Examples/Statistics/BayesianClassifier.cxx
```diff
@@ -67,15 +67,21 @@ main(int argc, char * argv[])
...
   // setup reader
   constexpr unsigned int Dimension = 2;
   using InputPixelType = float;
   using InputImageType = itk::VectorImage<InputPixelType, Dimension>;
-  const auto input = itk::ReadImage<InputImageType>(membershipImageFileName);
+  InputImageType::Pointer input;
+  try
+  {
+    input = itk::ReadImage<InputImageType>(argv[1]);
+  }
```
N-Dekker, line 79: In general it appears preferable to "try-catch" on a higher level, rather than adding a "try-catch" to each statement that might throw an exception.  So in this case, I would rather have the `try` to the begin of the `main` function, and the `catch` at the end, just before `return EXIT_SUCCESS`. 

See also CppCoreGuidelines [Don’t try to catch every exception in every function](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#e17-dont-try-to-catch-every-exception-in-every-function)

dzenanz: I completely agree with this. I updated accordingly the 4 files touched here.

N-Dekker (approved):
Thanks @dzenanz and @krupalbhat  👍

PR #5225, opened 2025-02-06
DOC: Main page was empty

dzenanz:
The problem is a result of filtering and as the dox files don't contain "C++" code but just comments, there is no need to run them through the "grouping filter". Raised here: https://github.com/InsightSoftwareConsortium/ITK/pull/3805#issuecomment-2640584317.

## PR Checklist


Utilities/Doxygen/DoxygenConfig.cmake
```diff
@@ -52,7 +52,9 @@ set(DOXYGEN_EXAMPLE_PATH "${DOXYGEN_TEST_DIRS}")
 set(DOXYGEN_EXAMPLE_PATTERNS "*.cxx")
 set(DOXYGEN_EXAMPLE_RECURSIVE "YES")
 set(DOXYGEN_IMAGE_PATH "${ITK_SOURCE_DIR}/Documentation/Art" "${ITK_BINARY_DIR}/Utilities/Doxygen/Modules")
-set(DOXYGEN_INPUT_FILTER "${ITK_DOXYGEN_INPUT_FILTER}")
+set(DOXYGEN_FILTER_PATTERNS
+    "*.h=${ITK_DOXYGEN_INPUT_FILTER}"
+    "*.cxx=${ITK_DOXYGEN_INPUT_FILTER}")
```
dzenanz, line 57: @albert-github why apply this to `*.cxx`, but not to `*.hxx`? Also, why use either? Aren't docstring taken only from `*.h` files?

albert-github: I didn't see any `hxx` file used when inspecting the output list and also the `FILE_PATTERNS          = *.h *.dox` doesn't mention them nor does `EXAMPLE_PATTERNS       = *.cxx`, but there is no objection to add them (but at the moment it won't have any effect).
It might even be that the `.cxx` isn't necessary.

hjmjohnson (approved):

PR #5231, opened 2025-02-07
ENH: Add support for VectorImages to Projection filters

blowekamp:
Update the ProjectionImageFilter base class to propagate the number of components to the output image.

Add multi-component support to max and min projection filters by performing operation on per commonents with initialization support for VariableLengthVectors.

## PR Checklist

### DIFF
--- a/Modules/Filtering/ImageStatistics/include/itkMinimumProjectionImageFilter.h
+++ b/Modules/Filtering/ImageStatistics/include/itkMinimumProjectionImageFilter.h
@@ -54,13 +54,39 @@ public:
   inline void
   Initialize()
   {
-    m_Minimum = NumericTraits<TInputPixel>::max();
+    if constexpr (std::is_same<TInputPixel, typename NumericTraits<TInputPixel>::ValueType>::value)
+    {
+      m_Minimum = NumericTraits<TInputPixel>::max();
+    }
+    else
+    {
+      m_Minimum = TInputPixel();
+      m_Minimum.Fill(NumericTraits<typename TInputPixel::ValueType>::max());
+    }
   }
 
   inline void
   operator()(const TInputPixel & input)
   {
-    m_Minimum = std::min(m_Minimum, input);
+
+    if constexpr (std::is_same<TInputPixel, typename NumericTraits<TInputPixel>::ValueType>::value)
+    {
+      m_Minimum = std::min(m_Minimum, input);
+    }
+    else
+    {
+      if (itk::NumericTraits<TInputPixel>::GetLength(m_Minimum) == 0)
+      {
+        m_Minimum = input;
+      }
+      else
+      {
+        for (unsigned int i = 0; i < itk::NumericTraits<TInputPixel>::GetLength(m_Minimum); ++i)
+        {
+          m_Minimum[i] = std::min(m_Minimum[i], input[i]);
+        }
+      }
+    }
   }
 
   inline TInputPixel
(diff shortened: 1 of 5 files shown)


Modules/Filtering/ImageStatistics/include/itkMinimumProjectionImageFilter.h
```diff
@@ -54,13 +54,38 @@ class MinimumAccumulator
...
   Initialize()
   {
-    m_Minimum = NumericTraits<TInputPixel>::max();
+    if constexpr (std::is_same<TInputPixel, typename NumericTraits<TInputPixel>::ValueType>::value)
+    {
+      m_Minimum = NumericTraits<TInputPixel>::max();
+    }
+    else
+    {
+      m_Minimum = TInputPixel();
```
dzenanz, line 63: Will this work correctly for constant-sized vectors such as RGB/RGBA?

Modules/Filtering/ImageStatistics/test/itkMaximumProjectionImageFilterTest4.cxx
```diff
@@ -0,0 +1,69 @@
...
+    return EXIT_FAILURE;
+  }
+
+  // Legacy compat with older MetaImages
+  itk::MetaImageIO::SetDefaultDoublePrecision(6);
+  const int dim = std::stoi(argv[1]);
+
+  using PixelType = unsigned char;
+
+  using ImageType = itk::VectorImage<PixelType, 3>;
```
dzenanz, line 44: Does regression test succeed if this is `using ImageType = itk::Image<itk::Vector<PixelType, 3>, 3>;`

Modules/Filtering/ImageStatistics/include/itkMinimumProjectionImageFilter.h
blowekamp: This is a good question. I was focused on SimpleITK VectorImage support... I am searching for the best way to initialize this for all vector-like types... Any ideas?

dzenanz: After this, set this to max component-wise via a for loop.

dzenanz:
```cpp
if (itk::NumericTraits<TInputPixel>::GetLength(m_Minimum) > 0)
{
  for ...
}
```

dzenanz: In fact, we don't even need the `if` - just the for loop is enough. If will devolve into a no-op if length of the vector is zero.

blowekamp: Yea, I have looked for a portable fill method long enough.

dzenanz: Does `Type::Filled()` work for VLVs? If so, it could be used here.

blowekamp: It probably works with enough types.

Modules/Filtering/ImageStatistics/test/itkMaximumProjectionImageFilterTest4.cxx
blowekamp: I gave a go with a local change for Vector and RGBPixel types, and it passed.

dzenanz (approved):

hjmjohnson (approved):

PR #5240, opened 2025-02-11
ENH: Let `itk::Array` support class template argument deduction (CTAD)

N-Dekker:
Added a deduction guide to avoid `-wctad-maybe-unsupported` compiler warnings.
Included a compile-time unit test.

### DIFF
--- a/Modules/Core/Common/include/itkArray.h
+++ b/Modules/Core/Common/include/itkArray.h
@@ -78,22 +78,25 @@ public:
    * the array does not manage the memory of the buffer. It merely points to
    * that location and it is the user's responsibility to delete it.
    * If "LetArrayManageMemory" is true, then this class will free the
-   * memory when this object is destroyed. */
-  Array(ValueType * datain, SizeValueType sz, bool LetArrayManageMemory = false);
+   * memory when this object is destroyed.
+   * \note This constructor supports class template argument deduction (CTAD). */
+  Array(TValue * datain, SizeValueType sz, bool LetArrayManageMemory = false);
 
 #if defined(ITK_LEGACY_REMOVE)
   /** Constructor that initializes array with contents from a user supplied
    * const buffer. The pointer to the buffer and the length is specified. By default,
    * the array does a deep copy of the const pointer data, so the array class also
-   * manages memory. */
-  Array(const ValueType * datain, SizeValueType sz);
+   * manages memory.
+   * \note This constructor supports class template argument deduction (CTAD). */
+  Array(const TValue * datain, SizeValueType sz);
 
 #else // defined ( ITK_LEGACY_REMOVE )
   /** Constructor that initializes array with contents from a user supplied
    * buffer. The pointer to the buffer and the length is specified. The array
    * does a deep copy of the const pointer data, so the array class also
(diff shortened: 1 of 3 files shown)


N-Dekker:
FYI, This pull request is inspired by `ConvertToDynamicArray` in elastix, at https://github.com/SuperElastix/elastix/blob/ca4593cb9836fc8c1d1177cbe5b8daef0e0e2ded/Common/itkNDImageTemplate.h#L240-L247 which does:

```cpp
  static TOut
  ConvertToDynamicArray(const TIn & in)
  {
    return TOut(in.data(), VDimension);
  }
```

I was thinking of rewriting it to:

```cpp
  static auto
  ConvertToDynamicArray(const TIn & in)
  {
    return itk::Array(in.data(), VDimension);
  }
```

But then it would be helpful if the template argument `TValue` of `itk::Array<TValue>` would be deduced automatically.

blowekamp:
Interesting, would this approach work with std::vectors as outputs too?

thewtex (approved):
💯

dzenanz (approved):

N-Dekker:
@blowekamp Although `std::vector<T>` does not have a `std::vector(const T *, size_t)` constructor, it is "intended" to support class template argument deduction (CTAD). So for example the following creates a `std::vector<Foo>`, allowing `Foo` to be deduced automatically, as tested at https://godbolt.org/z/7rPaq57W4

```cpp
#include <vector>
#include <type_traits>

struct Foo{};
Foo foo;
std::vector vec(42, foo); // CTAD without a warning :-)

static_assert(std::is_same_v<decltype(vec), std::vector<Foo>>); // OK!

```

blowekamp:
This compiled with gcc 14.2 on godbolt, do you know it's portable:

```
auto square(void) {
    int a[4] = {1,2,3,4};
    std::vector vec(a, a+4);
    static_assert(std::is_same_v<decltype(vec), std::vector<int>>);
   return vec;
}
```

Sorry for the tangent, about this C++ feature.

N-Dekker:
@blowekamp Interesting! I guess the `std::vector<T>` implementation of gcc has an extra constructor, `vector(const T * begin, const T * end)`! I'm not sure if it's standard-conforming to allow CTAD that way. It certainly isn't portable, because MSVC does not like it!

https://godbolt.org/z/9foEr3abT says:

```
<source>(9): error C2955: 'std::vector': use of class template requires template argument list
Z:/compilers/msvc/14.41.33923-14.41.33923.0/include\vector(429): note: see declaration of 'std::vector'
<source>(9): error C2514: 'std::vector': class template cannot be constructed
Z:/compilers/msvc/14.41.33923-14.41.33923.0/include\vector(429): note: see declaration of 'std::vector'
<source>(10): error C3203: 'vector': unspecialized class template can't be used as a template argument for template parameter '<unnamed-symbol>', expected a real type
<source>(10): error C2607: static assertion failed
<source>(11): error C2440: 'return': cannot convert from 'std::vector' to 'std::vector'
<source>(11): note: Cannot copy construct class 'std::vector' due to ambiguous copy constructors or no available copy constructor
Compiler returned: 2
```

PR #5241, opened 2025-02-11
ENH: Add value_type to FixedArray

N-Dekker:
Eases writing generic code.

### DIFF
--- a/Modules/Core/Common/include/itkFixedArray.h
+++ b/Modules/Core/Common/include/itkFixedArray.h
@@ -181,6 +181,9 @@ public:
     ConstIterator m_Iterator;
   };
 
+  /** The type of an element. */
+  using value_type = TValue;
+
   /** A pointer to the ValueType. */
   using pointer = ValueType *;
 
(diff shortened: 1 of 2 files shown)


dzenanz (approved):
In what context did you need this?

N-Dekker:
> In what context did you need this?

I was trying to simplify an old generic function in elastix, that converts a FixedArray, an `itk::Size` object, or an `itk::Index`, to a "dynamic" `itk::Array`. In https://github.com/SuperElastix/elastix/blob/4ae8c89bc34729d03b7776f633099aecd7f5bc3f/Common/itkNDImageTemplate.h#L238-L248

```cpp
  template <typename TIn, typename TOut>
  static TOut
  ConvertToDynamicArray(const TIn & in)
  {
    TOut out(VDimension);

    for (unsigned int i = 0; i < VDimension; ++i)
    {
      out[i] = in[i];
    }
    return out;
  }

```

I think this function should return an `itk::Array<TIn::value_type>`. `itk::Size` and `itk::Index` do have a value_type, but `itk::FixedArray` doesn't!

thewtex (approved):
:+1:

PR #5243, opened 2025-02-13
BUG: Do not run pre-commit GHA on push

thewtex:
Fails the `no-commit-to-branch` check for merges, but should be
unnecessary.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -2,8 +2,6 @@ name: pre-commit
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


blowekamp (approved):

dzenanz (approved):

jhlegarreta (commented):
Will the checks run for forced pushes?

thewtex:
> Will the checks run for forced pushes?

We should not have any forced pushes regardless!

`push` in this context means when a PR has been merged to `master`.

PR #3325, opened 2022-03-22
STYLE: Use ReadImage and WriteImage in a few iterator examples

PaulHax:
## PR Checklist
- [x] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [x] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [x] Added test (or behavior not changed)
- [x] Updated API documentation (or API not changed)
- [x] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [x] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [x] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples) for all new major features (if any)

Refactors 3 iterator examples to use the itk:ReadImage and itk:WriteImage convenience functions as directed by #2802

PR #2173, opened 2020-12-11
STYLE: Use the WriteImage convenience function anywhere in Modules/Core

N-Dekker:
Replaced explicit creation of `itk::ImageFileWriter` objects by calling
the new `itk::WriteImage` convenience function. Aims to reduces the
amount of boilerplate code.

PR #5246, opened 2025-02-15
STYLE: Enclose the majority of the main function in a try-catch block for Examples/SpatialObjects and Examples/RegistrationITKV4

krupalbhat:
This commit addresses the review comments from PR #5223 related to issue #2802 . In this commit, examples related to SpatialObjects and RegistrationITKV4 have been enclosed with try catch block.

## PR Checklist

### DIFF
--- a/Examples/RegistrationITKv4/MultiResImageRegistration1.cxx
+++ b/Examples/RegistrationITKv4/MultiResImageRegistration1.cxx
@@ -313,10 +313,10 @@ public:
 private:
   unsigned int m_CumulativeIterationIndex{ 0 };
 };
-
-
+namespace
+{
 int
-main(int argc, const char * argv[])
+ExampleMain(int argc, const char * const argv[])
 {
   if (argc < 4)
   {
@@ -481,19 +481,12 @@ main(int argc, const char * argv[])
   //
   //  Software Guide : EndLatex
 
-  try
-  {
-    registration->Update();
-    std::cout << "Optimizer stop condition: "
-              << registration->GetOptimizer()->GetStopConditionDescription()
-              << std::endl;
-  }
-  catch (const itk::ExceptionObject & err)
-  {
-    std::cout << "ExceptionObject caught !" << std::endl;
-    std::cout << err << std::endl;
-    return EXIT_FAILURE;
-  }
+
+  registration->Update();
+  std::cout << "Optimizer stop condition: "
+            << registration->GetOptimizer()->GetStopConditionDescription()
+            << std::endl;
+
 
   ParametersType finalParameters = transform->GetParameters();
 
@@ -665,4 +658,26 @@ main(int argc, const char * argv[])
 
 
   return EXIT_SUCCESS;
+}
+} // namespace
+int
+main(int argc, char * argv[])
+{
+  try
+  {
+    return ExampleMain(argc, argv);
+  }
+  catch (const itk::ExceptionObject & exceptionObject)
+  {
+    std::cerr << "ITK exception caught:\n" << exceptionObject << '\n';
+  }
+  catch (const std::exception & stdException)
+  {
+    std::cerr << "std exception caught:\n" << stdException.what() << '\n';
+  }
+  catch (...)
+  {
+    std::cerr << "Unhandled exception!\n";
+  }
+  return EXIT_FAILURE;
 }
--- a/Examples/RegistrationITKv4/MultiStageImageRegistration1.cxx
+++ b/Examples/RegistrationITKv4/MultiStageImageRegistration1.cxx
@@ -170,9 +170,10 @@ public:
 private:
   unsigned int m_CumulativeIterationIndex{ 0 };
 };
-
+namespace
+{
 int
-main(int argc, char * argv[])
+ExampleMain(int argc, const char * const argv[])
 {
   if (argc < 4)
   {
@@ -393,20 +394,13 @@ main(int argc, char * argv[])
   //  Software Guide : EndLatex
 
   // Software Guide : BeginCodeSnippet
-  try
-  {
-    transRegistration->Update();
-    std::cout
-      << "Optimizer stop condition: "
-      << transRegistration->GetOptimizer()->GetStopConditionDescription()
-      << std::endl;
-  }
-  catch (const itk::ExceptionObject & err)
-  {
-    std::cout << "ExceptionObject caught !" << std::endl;
-    std::cout << err << std::endl;
-    return EXIT_FAILURE;
-  }
+
+  transRegistration->Update();
+  std::cout
+    << "Optimizer stop condition: "
+    << transRegistration->GetOptimizer()->GetStopConditionDescription()
+    << std::endl;
+
 
   compositeTransform->AddTransform(
     transRegistration->GetModifiableTransform());
@@ -741,20 +735,12 @@ main(int argc, char * argv[])
   //  Software Guide : EndLatex
 
   // Software Guide : BeginCodeSnippet
-  try
-  {
-    affineRegistration->Update();
-    std::cout
-      << "Optimizer stop condition: "
-      << affineRegistration->GetOptimizer()->GetStopConditionDescription()
-      << std::endl;
-  }
-  catch (const itk::ExceptionObject & err)
-  {
-    std::cout << "ExceptionObject caught !" << std::endl;
-    std::cout << err << std::endl;
-    return EXIT_FAILURE;
-  }
+
+  affineRegistration->Update();
+  std::cout
+    << "Optimizer stop condition: "
+    << affineRegistration->GetOptimizer()->GetStopConditionDescription()
+    << std::endl;
 
   compositeTransform->AddTransform(
     affineRegistration->GetModifiableTransform());
@@ -931,15 +917,9 @@ main(int argc, char * argv[])
   // Before registration
   using TransformType = itk::IdentityTransform<double, Dimension>;
   TransformType::Pointer identityTransform;
-  try
-  {
-    identityTransform = TransformType::New();
-  }
-  catch (const itk::ExceptionObject & err)
-  {
-    err.Print(std::cerr);
-    return EXIT_FAILURE;
-  }
+
+  identityTransform = TransformType::New();
+
   identityTransform->SetIdentity();
   resample->SetTransform(identityTransform);
 
@@ -956,4 +936,28 @@ main(int argc, char * argv[])
   }
 
   return EXIT_SUCCESS;
+}
+} // namespace
+
+
+int
+main(int argc, char * argv[])
+{
+  try
+  {
+    return ExampleMain(argc, argv);
+  }
+  catch (const itk::ExceptionObject & exceptionObject)
+  {
+    std::cerr << "ITK exception caught:\n" << exceptionObject << '\n';
+  }
+  catch (const std::exception & stdException)
+  {
+    std::cerr << "std exception caught:\n" << stdException.what() << '\n';
+  }
+  catch (...)
+  {
+    std::cerr << "Unhandled exception!\n";
+  }
+  return EXIT_FAILURE;
 }
--- a/Examples/RegistrationITKv4/MultiResImageRegistration2.cxx
+++ b/Examples/RegistrationITKv4/MultiResImageRegistration2.cxx
@@ -166,8 +166,11 @@ public:
   }
 };
 
+namespace
+{
+
 int
-main(int argc, char * argv[])
+ExampleMain(int argc, const char * const argv[])
 {
   if (argc < 4)
   {
@@ -438,19 +441,12 @@ main(int argc, char * argv[])
   registration->AddObserver(itk::IterationEvent(), command);
   registration->SetNumberOfLevels(3);
 
-  try
-  {
-    registration->Update();
-    std::cout << "Optimizer stop condition: "
-              << registration->GetOptimizer()->GetStopConditionDescription()
-              << std::endl;
-  }
-  catch (const itk::ExceptionObject & err)
-  {
-    std::cout << "ExceptionObject caught !" << std::endl;
-    std::cout << err << std::endl;
-    return EXIT_FAILURE;
-  }
+
+  registration->Update();
+  std::cout << "Optimizer stop condition: "
+            << registration->GetOptimizer()->GetStopConditionDescription()
+            << std::endl;
+
 
   std::cout << "Optimizer Stopping Condition = "
             << optimizer->GetStopCondition() << std::endl;
@@ -622,4 +618,27 @@ main(int argc, char * argv[])
   }
 
   return EXIT_SUCCESS;
+}
+} // namespace
+
+int
+main(int argc, char * argv[])
+{
+  try
+  {
+    return ExampleMain(argc, argv);
+  }
+  catch (const itk::ExceptionObject & exceptionObject)
+  {
+    std::cerr << "ITK exception caught:\n" << exceptionObject << '\n';
+  }
+  catch (const std::exception & stdException)
+  {
+    std::cerr << "std exception caught:\n" << stdException.what() << '\n';
+  }
+  catch (...)
+  {
+    std::cerr << "Unhandled exception!\n";
+  }
+  return EXIT_FAILURE;
 }
--- a/Examples/RegistrationITKv4/MultiResImageRegistration3.cxx
+++ b/Examples/RegistrationITKv4/MultiResImageRegistration3.cxx
@@ -134,9 +134,10 @@ public:
   }
 };
 
-
+namespace
+{
 int
-main(int argc, char * argv[])
+ExampleMain(int argc, const char * const argv[])
 {
   if (argc < 4)
   {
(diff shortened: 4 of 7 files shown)


jhlegarreta (commented):
I do not think including dozens of lines of code, including type aliases, regular assignments, and other statements that we know do not throw ITK exceptions, within `try`/`catch` blocks would be a desirable change.

krupalbhat:
How would you prefer the error handling here? Should I keep try/catch only around the parts that might throw ITK exceptions? Initially, it was like that, but later it was changed to wrap the full main block based on review comments.

jhlegarreta:
> How would you prefer the error handling here? Should I keep try/catch only around the parts that might throw ITK exceptions? Initially, it was like that, but later it was changed to wrap the full main block based on review comments.

I made my comment based on the ITK guidelins (f I recall correctly, this is mentioned explicitly) and many style changes that have reduced the scope of such `try`/`catch` blocks. Prior to my comment, I had read this as well https://github.com/InsightSoftwareConsortium/ITK/pull/5163#discussion_r1932895575.

And now I see this other one https://github.com/InsightSoftwareConsortium/ITK/pull/5223#discussion_r1945410950.

I do not think using a single `try`/`catch` block for the entire body is better, but adding one for every ITK operation is not good either. IMO it should be picked where to place them.

hjmjohnson (dismissed):
These examples are an important part of the software guide.  The extra indentation will cause errors in generating the software guide.

These examples are to demonstrate major ITK design principles, and they are not intended to be robust applications.  While the error handling is good programming practice for production applications, they diminish the core demonstration of ITK features. 

I appreciate your submission, but I think that the benefits of keeping these examples focused only on ITK features outweighs the benefits of the improved error checking.

krupalbhat:
Thank you for the clarification. Should I go ahead and close this PR? I also need some guidance—there are still some examples that can be updated to use itk::ReadImage and itk::WriteImage. In those cases, should I add local try-catch blocks for exception handling, or is it better to not add any exception handling?"

N-Dekker:
Personally I like the direction of the PR. It's in general recommended to prevent exceptions from leaving the `main` function. Statistic analysis tools even produce a warning for such cases: "Warning: An exception may be thrown in function 'main' which should not throw exceptions" (JetBrains ReSharper) or https://clang.llvm.org/extra/clang-tidy/checks/bugprone/exception-escape.html (Clang-Tidy).

I also think it's a good idea to show how to catch those exceptions in the examples of ITK.

It's fine to me to place the entire body of a `main` function inside a try-catch block. If the extra indentation feels cumbersome, the inner part of the code might be moved to a helper function.

```cpp
namespace
{
// Note: ExampleMain might throw an exception!
int
ExampleMain(const int argc, const char * const * const argv)
{
  (The original code from main should go here!)
}
} // namespace

int
main(int argc, char * argv[])
{
  try
  {
    return ExampleMain(argc, argv);
  }
  catch (const itk::ExceptionObject & exceptionObject)
  {
    std::cerr << "ITK exception caught:\n" << exceptionObject << '\n';
  }
  catch (const std::exception & stdException)
  {
    std::cerr << "std exception caught:\n" << stdException.what() << '\n';
  }
  return EXIT_FAILURE;
}

```

My 2 cents, Niels

krupalbhat:
> Generally looks good. Thank you for the effort.
> 
> > commit [a422c27](https://github.com/InsightSoftwareConsortium/ITK/commit/a422c27057da1ea5b9e17ce12741992902ee79f6) has an invalid commit subject; the first line must be no longer than 78 characters.
> 
> It would be good if you amended the commit to change the commit message according to the above. If not, I can do it for you next week.

Thank you for the clarification. I have amended the commit message.

dzenanz (approved):

jhlegarreta (commented):
I very much dislike the change and agree with Hans https://github.com/InsightSoftwareConsortium/ITK/pull/5246#pullrequestreview-2619727407. So I cannot approve this change.

dzenanz:
Can you propose an alternative? Or pick one example most in need of change, and show how to change it?

jhlegarreta:
> Can you propose an alternative? Or pick one example most in need of change, and show how to change it?

The issue #2802 has been previously addressed changing just the concerned block and without requiring to enclose any additional part of the code in `try/catch`blocks or using `ITK_TRY_EXPECT_NO_EXCEPTION` to hold the statements, e.g.:
https://github.com/InsightSoftwareConsortium/ITK/pull/2160/files
https://github.com/InsightSoftwareConsortium/ITK/pull/2173/files
https://github.com/InsightSoftwareConsortium/ITK/pull/2457/files
https://github.com/InsightSoftwareConsortium/ITK/pull/3325/files

So again, re https://github.com/InsightSoftwareConsortium/ITK/pull/5223#discussion_r1945410950, I do not see the need to enclose parts that will never trigger an exception into a `try/catch` block.

hjmjohnson:
FYI:  I am not opposed to this set of conmits.  I appreciate the community efforts.  I'm on the fence about the need to make example code "perfect in all ways".

From my teaching, my tendency is to make examples as focused and simple as possible, even if it is not best-practices complete all around.   With that said, I am conflicted about where these examples fall on that continuum of completeness.  

I realize that the code is re-flowed, but that does affect the code in the software guide, and may make the code longer (more lines).  Perhaps we can "de-indent" the examples in the software guide.

blowekamp:
WIth optimization (such as registration and deep learning), it can be frustrating when processing runs for a long time and the some error occurs and results are not saved.

In the current change the exception handling is not useful. It may give a more descriptive output? Another utility of the exception handling may be to save the intermediate results?

dzenanz:
The main benefit of enclosing the entire "main" into a try-catch block is not having to think about what statements (and even constructors) might throw. Not matter where the exception occurs, some error message is produced.

For indentation problems, [Niels' suggestion](https://github.com/InsightSoftwareConsortium/ITK/pull/5246#issuecomment-2669253563) seems good. I could not find any indentation settings specific to try-catch block in [clang-format](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Examples/.clang-format). The closest is [BraceWrappingBeforeCatch](https://github.com/InsightSoftwareConsortium/ITK/blob/32a2a6de17ffb7c8319ab38dbe61bd3b7c171f00/Examples/.clang-format#L114).

N-Dekker:
Thanks @dzenanz For the sake of completeness, regarding my suggestion at https://github.com/InsightSoftwareConsortium/ITK/pull/5246#issuecomment-2669253563: Strictly speaking, the `catch (const itk::ExceptionObject &)` isn't necessary. A general `catch (const std::exception &)` catches exceptions from ITK as well (because `itk::ExceptionObject` is derived from `std::exception`).

dzenanz:
That is correct, but ITK exception is richer, usually containing the file/line where the error occured.

krupalbhat:
How should i proceed for now?

dzenanz:
> Another utility of the exception handling may be to save the intermediate results?

Adding that feature would be another step.

For now, do we agree that [Niels' suggestion](https://github.com/InsightSoftwareConsortium/ITK/pull/5246#issuecomment-2669253563) is good enough? I like and use this pattern a lot, see e.g. this [montage example](https://github.com/InsightSoftwareConsortium/ITKMontage/blob/v0.8.2/examples/CompleteMontage.cxx#L599-L627).

dzenanz:
@hjmjohnson, @blowekamp do you agree with the course of action?

hjmjohnson:
> @hjmjohnson, @blowekamp do you agree with the course of action?

I like @N-Dekker  suggestion *a lot*!  It addresses all my concerns, and it decouples the demonstration of exception handling from the demonstration of the code.

As an added bonus, I can see how this could accelerate copy-n-paste process for others to use these examples as starting points in their own code.

dzenanz:
@jhlegarreta and @blowekamp, do you have any objections to Niels' suggestion?

krupalbhat:
can i start working on it??

N-Dekker:
Thanks for your perseverance, @krupalbhat 👍 

Instead of merging the master branch into your branch, I would suggest to use `git rebase master` (on your fork, on your branch). I think that's the most common way to sync a branch on a fork with the master. Thereby it avoids the extra "merge branch" commit.

-- my 2 cents

krupalbhat:
Sure. Thanks a lot for your suggestion.

dzenanz:
Due to lack of response, I assume that Jon and Brad don't have objections. I guess you can proceed Krupal.

krupalbhat:
Thanks. I will try to complete by this weekend.

jhlegarreta:
> Due to lack of response, I assume that Jon and Brad don't have objections.

https://github.com/InsightSoftwareConsortium/ITK/pull/5246#issuecomment-2693058873: I still dislike the approach but will not oppose if others think that this is an improvement.

krupalbhat:
I have applied the suggested approach by renaming `main()` to `ExampleMain()` and using the following `main()` template across the examples:
```cpp
int
main(int argc, const char * argv[])
{
  try
  {
    return ExampleMain(argc, argv);
  }
  catch (const itk::ExceptionObject & exceptionObject)
  {
    std::cerr << "ITK exception caught:\n" << exceptionObject << '\n';
  }
  catch (const std::exception & stdException)
  {
    std::cerr << "std exception caught:\n" << stdException.what() << '\n';
  }
  catch (...)
  {
    assert(!"Unhandled exception!");
  }
  return EXIT_FAILURE;
}
```
Since this main function is now identical across all examples, would it be better to avoid this redundancy using a macro? Please let me know if there's a preferred approach.

dzenanz:
Each example should be self contained, so macro is the wrong solution. Also, I would like `assert(!"Unhandled exception!");` be changed into `std::cerr << "Unhandled exception!";`

krupalbhat:
Then, is it okay to write the same `main` function in all examples?  
Sure, I will change the last error statement to `std::cerr << "Unhandled exception!\n";`

dzenanz:
Yes.

Written on cellphone, excuse my brevity.

N-Dekker:
@dzenanz 
> I would like `assert(!"Unhandled exception!");` be changed into `std::cerr << "Unhandled exception!";`

Just explaining, I chose to have an assert failure on `catch (...)`, for two reasons:
- I think our code convention assumes that an exception is always derived from `std::exception`. So an exception that is _not_ derived from `std::exception` indicates that there is a bug. (While an exception caught by `catch (const std::exception & stdException)` is caused by an exceptional runtime situation, like out-of-memory, out-of-resources, or file I/O failures.) An assert failure is the typical way to report a bug.
- `std::cerr << ...` might itself throw an exception (exceptionally, of course). While an `assert` does not throw. Of course, you could add an extra try-catch around the `std::cerr << ...` statement, but that looks a bit ugly 🤷 

My 2 cent, Niels

dzenanz (approved):

dzenanz:
/azp.run ITK.macOS

N-Dekker:
> Since this main function is now identical across all examples, would it be better to avoid this redundancy using a macro? 

Dženan already answered to this question, but _if_ we would still feel the need to avoid such redundancy, I would rather have a template function, instead of a macro, to catch all exceptions. For example `ExecuteAndCatchAll<TFunc>(TFunc)`, to be used as follows:

```cpp
int
main(int argc, char ** argv)
{
  ExecuteAndCatchAll([argc, argv] { return ExampleMain(argc, argv); });
} // end main

```
 
Possibly implemented as follows:

```cpp
template <typename TFunc>
int
ExecuteAndCatchAll(TFunc func) noexcept
{
  try
  {
    try
    {
      return func();
    }
    catch (const itk::ExceptionObject & exceptionObject)
    {
      std::cerr << "ITK exception caught:\n" << exceptionObject << '\n';
    }
    catch (const std::exception & stdException)
    {
      std::cerr << "std exception caught:\n" << stdException.what() << '\n';
    }
  }
  catch (...)
  {
    // Maybe `assert(!"Unhandled exception!")`, because exceptions _must_ be derived from std::exception!
  }
  return EXIT_FAILURE;
}

```

krupalbhat:
The macOS CI/CD failed while Linux and Windows passed. Not sure how to fix this, any suggestions?

Examples/RegistrationITKv4/MultiResImageRegistration1.cxx
```diff
@@ -666,3 +660,25 @@ main(int argc, const char * argv[])
...
+    std::cerr << "ITK exception caught:\n" << exceptionObject << '\n';
+  }
+  catch (const std::exception & stdException)
+  {
+    std::cerr << "std exception caught:\n" << stdException.what() << '\n';
+  }
+  catch (...)
+  {
+    std::cerr << "Unhandled exception!\n";
+  }
```
N-Dekker, line 682: I hesitate because this approach still may not keep [clang-tidy bugprone-exception-escape](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/exception-escape.html) silent. At least _in theory_, `std::cerr << "..."` might still throw an exception.

This could be solved by using a C function like `fputs("...", stderr)`, instead of `std::cerr`. Or by adding an extra `try` `catch(...)`.

On the other hand, we could also decide to just ignore such clang-tidy checks.

dzenanz:
The error messages are [here](https://open.cdash.org/viewBuildError.php?buildid=10286752). They are: `Examples/RegistrationITKv4/MultiResImageRegistration2.cxx:634:12: error: no matching function for call to 'ExampleMain'` and `Examples/RegistrationITKv4/MultiResImageRegistration3.cxx:387:12: error: no matching function for call to 'ExampleMain'`. I think the problem is that the main's argv is `const`, while ExampleMain's argv is not const.

Examples/RegistrationITKv4/MultiResImageRegistration1.cxx
dzenanz: Are there any known situations in which `std::cerr << "...";` throws an exception?

dzenanz: To use `fputs` we would need to `#include <stdio.h>`. Inconvenient. I am not sure we want to teach this practice in examples.

dzenanz: I am not regularly running clang-tidy, so it is easy for me to ignore its objections 😄 I vote for this option.

N-Dekker:
> I think the problem is that the main's argv is `const`, while ExampleMain's argv is not const.

Indeed. For historical reasons, the parameters of the `main` function of a C++ program are usually non-const.

krupalbhat:
> > I think the problem is that the main's argv is `const`, while ExampleMain's argv is not const.
> 
> Indeed. For historical reasons, the parameters of the `main` function of a C++ program are usually non-const.
> 
> PS If you still want to make them const (which could be nice), please check that it's according to the C++ standard. 😃 ExampleMain may have `const` parameters, of course, that's no problem!

Yeah, In most of the examples they are using char* argv[]. I will stick to that

N-Dekker:
> In most of the examples they are using char* argv[]. I will stick to that

For what it's worth, here is a related C++ Standard issue: [CWG issue 1003 "Acceptable definitions of main"](https://cplusplus.github.io/CWG/issues/1003.html), Daniel Krügler, 2009-11-17. So it now says:

 An implementation shall allow both
- a function of `()` returning int and
- a function of (`int`, pointer to pointer to `char`) returning in

Still not really clear to me whether such a "pointer to pointer to `char`" may be `const` 🤷

Examples/RegistrationITKv4/MultiResImageRegistration1.cxx
N-Dekker: Thanks @dzenanz I'm neutral, if that's also an option  😺

krupalbhat:
> > I think the problem is that the main's argv is `const`, while ExampleMain's argv is not const.
> 
> Indeed. For historical reasons, the parameters of the `main` function of a C++ program are usually non-const.
> 
> PS If you still want to make them const (which could be nice), please check that it's according to the C++ standard. 😃 ExampleMain may have `const` parameters, of course, that's no problem!
I have done this, but it is still failing on macOS. Should I change ExampleMain to int ExampleMain(int argc, char* argv[])?

N-Dekker:
> Should I change ExampleMain to int ExampleMain(int argc, char* argv[])?

Not necessary! `int ExampleMain(int argc, const char* const argv[])` should work!

Examples/RegistrationITKv4/MultiStageImageRegistration1.cxx
```diff
@@ -954,6 +944,29 @@ main(int argc, char * argv[])
   {
     itk::WriteImage(caster->GetOutput(), argv[6]);
   }
+}
+return EXIT_SUCCESS;
```
N-Dekker, line 948: I think this `}` at line 947 is a typo! 

See also https://open.cdash.org/viewBuildError.php?buildid=10298799 saying:
> [Examples/RegistrationITKv4/MultiStageImageRegistration1.cxx:948](https://github.com/InsightSoftwareConsortium/ITK/blob/5a9a52a7d055a57a7dab107d9efe1a5fa82f894c/Examples/RegistrationITKv4/MultiStageImageRegistration1.cxx#L948):1: error: expected unqualified-id

N-Dekker:
FYI Here is an explanation for why a `char * *` argument does not convert implicitly to `const char * *`:

[FAQ  Why am I getting an error converting a `Foo**` → `const Foo**`](https://isocpp.org/wiki/faq/const-correctness#constptrptr-conversion)

It's a bit tricky, because usually `non-const` does convert implicitly not `const`, as you would expect.

Examples/RegistrationITKv4/MultiStageImageRegistration1.cxx
krupalbhat: Thanks a lot. Its working now!!

N-Dekker (approved):
Thank you @krupalbhat 

I found it a bit difficult to review the changes, because it has all those textual reformatting. But it looks fine to me now 👍

krupalbhat:
> Thank you @krupalbhat
> 
> I found it a bit difficult to review the changes, because it has all those textual reformatting. But it looks fine to me now 👍

Ok. I will fix those

N-Dekker (approved):

dzenanz (approved):

dzenanz:
@hjmjohnson Can you review this again?

dzenanz:
Thank you for your persistence Krupal!

krupalbhat:
Thanks a lot for your constant support, @dzenanz  and @N-Dekker! I’ve learned a lot from this. I can proceed the same way for other examples, right?

dzenanz:
Yes!

Issue #5245, opened 2025-02-15
1 test fails: NumberToString.DecimalNotationUpTo21Digits


yurivict:
```
272/3050 Testing: NumberToString.DecimalNotationUpTo21Digits
272/3050 Test: NumberToString.DecimalNotationUpTo21Digits
Command: "/usr/ports/science/InsightToolkit/work/.build/bin/ITKCommonGTestDriver" "--gtest_filter=NumberToString.DecimalNotationUpTo21Digits" "--gtest_also_run_disabled_tests"
Directory: /usr/ports/science/InsightToolkit/work/.build/Modules/Core/Common/test
"NumberToString.DecimalNotationUpTo21Digits" start time: Feb 14 12:43 PST
Output:
----------------------------------------------------------
Running main() from /wrkdirs/usr/ports/devel/googletest/work/googletest-1.16.0/googletest/src/gtest_main.cc
Note: Google Test filter = NumberToString.DecimalNotationUpTo21Digits
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from NumberToString
[ RUN      ] NumberToString.DecimalNotationUpTo21Digits
/usr/ports/science/InsightToolkit/work/ITK-5.4.2/Modules/Core/Common/test/itkNumberToStringGTest.cxx:99: Failure
Expected equality of these values:
  numberToString(power_of_ten)
    Which is: "100000010000"
  '1' + std::string(exponent, '0')
    Which is: "100000000000"
Floating point type: float

/usr/ports/science/InsightToolkit/work/ITK-5.4.2/Modules/Core/Common/test/itkNumberToStringGTest.cxx:100: Failure
Expected equality of these values:
  numberToString(-power_of_ten)
    Which is: "-100000010000"
  "-1" + std::string(exponent, '0')
    Which is: "-100000000000"
Floating point type: float

/usr/ports/science/InsightToolkit/work/ITK-5.4.2/Modules/Core/Common/test/itkNumberToStringGTest.cxx:108: Failure
Expected equality of these values:
  numberToString(power_of_ten)
    Which is: "0.000009999999999999999"
  "0." + std::string(-1 - exponent, '0') + '1'
    Which is: "0.00001"
Floating point type: double

/usr/ports/science/InsightToolkit/work/ITK-5.4.2/Modules/Core/Common/test/itkNumberToStringGTest.cxx:109: Failure
Expected equality of these values:
[...]

PR #5248, opened 2025-02-17
BUG: Replace `pow` with `Math::UnsignedPower` in NumberToString GTest

N-Dekker:
Aims to address test failures reported by @yurivict, at https://github.com/InsightSoftwareConsortium/ITK/issues/5245, saying:

```
/usr/ports/science/InsightToolkit/work/ITK-5.4.2/Modules/Core/Common/test/itkNumberToStringGTest.cxx:99: Failure
  numberToString(power_of_ten)
    Which is: "100000010000"
  '1' + std::string(exponent, '0')
    Which is: "100000000000"
Floating point type: float

...

/usr/ports/science/InsightToolkit/work/ITK-5.4.2/Modules/Core/Common/test/itkNumberToStringGTest.cxx:108: Failure
  numberToString(power_of_ten)
    Which is: "0.000009999999999999999"
  "0." + std::string(-1 - exponent, '0') + '1'
    Which is: "0.00001"
Floating point type: double
```

Using clang-19 on FreeBSD 14.2.

Sean McBride (@seanm) confirmed that this commit would solve the issue on his FreeBSD VM.

### DIFF
--- a/Modules/Core/Common/test/itkNumberToStringGTest.cxx
+++ b/Modules/Core/Common/test/itkNumberToStringGTest.cxx
@@ -18,8 +18,8 @@
 
 // First include the header file to be tested:
 #include "itkNumberToString.h"
+#include "itkMath.h"
 #include <gtest/gtest.h>
-#include <cmath> // For std::pow.
 
 namespace
 {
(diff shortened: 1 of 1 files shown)


thewtex (approved):
:100: thanks @N-Dekker

PR #5249, opened 2025-02-17
STYLE: Replace `std::pow(x, 2)` calls with `Math::sqr(x)`

N-Dekker:
`Math::sqr(x)` appears slightly more readable than `std::pow(x, 2)`.

### DIFF
--- a/Modules/Core/Common/include/itkFiniteCylinderSpatialFunction.hxx
+++ b/Modules/Core/Common/include/itkFiniteCylinderSpatialFunction.hxx
@@ -19,6 +19,7 @@
 #define itkFiniteCylinderSpatialFunction_hxx
 
 #include "itkFloatingPointExceptions.h"
+#include "itkMath.h"
 #include <cmath>
 
 namespace itk
@@ -102,7 +103,7 @@ FiniteCylinderSpatialFunction<VDimension, TInput>::Evaluate(const InputType & po
   }
 
   if (itk::Math::abs(distanceFromCenter) <= (halfAxisLength) &&
-      m_Radius >= std::sqrt(std::pow(pointVector.GetNorm(), 2.0) - std::pow(distanceFromCenter, 2.0)))
+      m_Radius >= std::sqrt(Math::sqr(pointVector.GetNorm()) - Math::sqr(distanceFromCenter)))
   {
     return 1;
   }
--- a/Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.hxx
+++ b/Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.hxx
@@ -128,7 +128,7 @@ StatisticsLabelMapFilter<TImage, TFeatureImage>::ThreadedProcessLabelObject(Labe
 
     // increase the sums
     sum += v;
-    sum2 += std::pow(static_cast<double>(v), 2);
+    sum2 += Math::sqr(static_cast<double>(v));
     sum3 += std::pow(static_cast<double>(v), 3);
     sum4 += std::pow(static_cast<double>(v), 4);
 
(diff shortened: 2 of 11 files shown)


issakomi:
BTW, `sqr` sometimes means  _square root_, e.g. in [VBA](https://support.microsoft.com/en-us/office/sqr-function-b0189a2a-57ba-4718-aef8-13e5d1ecf7e3), [Delphi](http://www.delphibasics.co.uk/RTL.php?Name=sqr).

seanm:
I had actually started writing a comment wondering why we saved 1 letter and didn't just call it `sqrt` like C/C++ do, then I realized it's not square root just before clicking "Comment".

Renaming it `square` might be wise...

N-Dekker:
Thanks @issakomi and @seanm I don't like the name `sqr` either. But it's taken directly from the [vxl](https://github.com/vxl/vxl/blob/ea82b8e562ddcf161cd8bb97041d1979cc4060a2/core/vnl/vnl_math.h#L694) library:

https://github.com/InsightSoftwareConsortium/ITK/blob/018e0eab2bae8b377fd0ee3af83fdcbf22a36f93/Modules/Core/Common/include/itkMath.h#L822

`itk::Math::sqr` was added to ITK almost nine years ago, with commit 2133446772e6a688e25cbfb1b74bb709103fe77f, from 2016.

If it would be implemented in ITK instead, I would prefer it to be named `Square`, following ITK's "Pascal Case" naming convention.

Anyway, my aim for this PR is to avoid calling `std::pow` unnecessarily, as it appears somewhat troublesome, as reported by issue #5245

I feel that a new square function for ITK is beyond the scope of this PR. (The new square function might be constexpr, it might be written as a function template, it would need extra unit tests, deprecating the current `sqr`, considering what to do about `Math::cube`, etc.) Do you agree that it would be an acceptable improvement to _at least_ replace `std::pow(x, 2)` calls with the old `Math::sqr(x)`, and then possibly _later_ introduce a new square function?

Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.hxx
```diff
@@ -128,7 +128,7 @@ StatisticsLabelMapFilter<TImage, TFeatureImage>::ThreadedProcessLabelObject(Labe
 
     // increase the sums
     sum += v;
-    sum2 += std::pow(static_cast<double>(v), 2);
+    sum2 += Math::sqr(static_cast<double>(v));
```
seanm, line 131: This one is arguably better left as it way, to be more similar to the next 2 lines... but I don't feel strongly about it.

seanm (approved):

Modules/Filtering/LabelMap/include/itkStatisticsLabelMapFilter.hxx
N-Dekker: I see your point, thanks, but on the other hand, the next line (`sum3 += std::pow(static_cast<double>(v), 3)`) might as well use `Math::cube` instead  😸

Modules/Core/Common/include/itkFiniteCylinderSpatialFunction.hxx
```diff
@@ -102,7 +103,7 @@ FiniteCylinderSpatialFunction<VDimension, TInput>::Evaluate(const InputType & po
   }
 
   if (itk::Math::abs(distanceFromCenter) <= (halfAxisLength) &&
-      m_Radius >= std::sqrt(std::pow(pointVector.GetNorm(), 2.0) - std::pow(distanceFromCenter, 2.0)))
+      m_Radius >= std::sqrt(Math::sqr(pointVector.GetNorm()) - Math::sqr(distanceFromCenter)))
```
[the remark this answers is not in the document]
N-Dekker: Note that `distanceFromCenter` is in fact the dot product of two vectors: 
https://github.com/InsightSoftwareConsortium/ITK/blob/e9af5c5e7745d59ccaa3ff2fd1eff4275ba427d0/Modules/Core/Common/include/itkFiniteCylinderSpatialFunction.hxx#L96-L97

So I'm just wondering if the square of `distanceFromCenter` was originally intended. (But I will leave it "squared" like this, because I don't know the original intention.)

dzenanz (approved):

dzenanz:
This is a step in the right direction. If someone has time, this might be improved further.

PR #5252, opened 2025-02-21
STYLE: Replace `std::pow(2, dim)` calls with `1 << dim`

N-Dekker:
Replaced `std::pow(2.0, x)` with `1ULL << x` where `x` is either `ImageDimension` or `ImageDimension + 1`. Declared a few more variables `constexpr`.

In general, a left-shift may be faster than an `std::pow` call.


thewtex (approved):
:1st_place_medal:

PR #5265, opened 2025-03-06
Remove a few local `stringstream` variables from ResourceProbe and ExceptionObject 

N-Dekker:
Three small style improvements.

### DIFF
--- a/Modules/Core/Common/include/itkResourceProbe.hxx
+++ b/Modules/Core/Common/include/itkResourceProbe.hxx
@@ -427,8 +427,6 @@ template <typename ValueType, typename MeanType>
 void
 ResourceProbe<ValueType, MeanType>::JSONReport(std::ostream & os)
 {
-  const std::stringstream ss;
-
   ValueType ratioOfMeanToMinimum;
   if (Math::ExactlyEquals(this->GetMinimum(), 0.0))
   {
--- a/Modules/Core/Common/include/itkResourceProbe.hxx
+++ b/Modules/Core/Common/include/itkResourceProbe.hxx
@@ -583,10 +583,9 @@ ResourceProbe<ValueType, MeanType>::PrintJSONSystemInformation(std::ostream & os
   PrintJSONvar(os, "Bitness", (systeminfo.Is64Bits() ? "64 bit" : "32 bit"), 6, false);
   os << "    },\n";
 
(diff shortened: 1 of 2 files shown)


Modules/Core/Common/include/itkResourceProbe.hxx
```diff
@@ -427,8 +427,6 @@ template <typename ValueType, typename MeanType>
 void
 ResourceProbe<ValueType, MeanType>::JSONReport(std::ostream & os)
 {
-  const std::stringstream ss;
```
N-Dekker, line 430: @hjmjohnson  Another case that I found because you added those const keywords (PR https://github.com/InsightSoftwareConsortium/ITK/pull/5035 commit https://github.com/InsightSoftwareConsortium/ITK/commit/375ee8c47a03792d0d3d5447c6b4c4be6b9f2fb4)! A local const stringstream looks suspicious, doesn't it?!? When it was still non-const, this unused variable could remain under the radar, for more than seven years, actually!

Modules/Core/Common/src/itkExceptionObject.cxx
```diff
@@ -41,11 +41,8 @@ class ExceptionObject::ExceptionData
     , m_File(std::move(file))
     , m_Line(line)
   {
-    std::ostringstream loc;
-
-    loc << ':' << m_Line << ":\n";
     m_What = m_File;
-    m_What += loc.str();
+    m_What += ':' + std::to_string(m_Line) + ":\n";
     m_What += m_Description;
```
N-Dekker, line 46: Note that historically, the `what()` text of an `itk::ExceptionObject` includes the file name and the line number, but _not_ the location, `m_Location` (typically the name of the throwing function). I think it would be nice to add this location to the `what()` text. What do you think? Maybe in another pull request...

Modules/Core/Common/include/itkResourceProbe.hxx
hjmjohnson: Great!  Linting tools often have un-intended benefits.

Modules/Core/Common/src/itkExceptionObject.cxx
hjmjohnson: Great idea!

hjmjohnson (approved):

PR #5266, opened 2025-03-07
ENH: Add location (function) to text returned by ExceptionObject::what()

N-Dekker:
- Triggered by encouraging feedback from Hans (@hjmjohnson) and Dženan (@dzenanz) at https://github.com/InsightSoftwareConsortium/ITK/pull/5265#discussion_r1983786150

---

With this pull request, a call like `itkGenericExceptionMacro("Oops, something horribly wrong!")` will create a "what text" like the following:

```
D:\MyProject\MySource.cxx:42: in 'void MyObject::DoMyThing(int)':
ITK ERROR: Oops, something horribly wrong!

```
So it adds something like ` in 'void MyObject::DoMyThing(int)':` to the first line.

### DIFF
--- a/Modules/Core/Common/src/itkExceptionObject.cxx
+++ b/Modules/Core/Common/src/itkExceptionObject.cxx
@@ -42,7 +42,12 @@ public:
     , m_Line(line)
   {
     m_What = m_File;
-    m_What += ':' + std::to_string(m_Line) + ":\n";
+    m_What += ':' + std::to_string(m_Line) + ':';
+    if (!m_Location.empty())
+    {
+      m_What += " in '" + m_Location + "':";
+    }
+    m_What += '\n';
     m_What += m_Description;
   }
 
--- a/Modules/Core/Common/test/itkExceptionObjectGTest.cxx
+++ b/Modules/Core/Common/test/itkExceptionObjectGTest.cxx
@@ -84,18 +84,48 @@ TEST(ExceptionObject, TestDescriptionFromSpecializedExceptionMacro)
 
 
 // Tests that `ExceptionObject::what()` returns the expected concatenation of
-// file name, line number, and description.
+// file name, line number, location and description.
 TEST(ExceptionObject, TestWhat)
 {
-  const itk::ExceptionObject exceptionObject(__FILE__, __LINE__, "test description");
+  {
+    // Test with empty location.
+    const itk::ExceptionObject exceptionObject(__FILE__, __LINE__, "test description", "");
 
-  const char * const what = exceptionObject.what();
-  const char * const file = exceptionObject.GetFile();
-  const char * const description = exceptionObject.GetDescription();
+    const char * const what = exceptionObject.what();
+    const char * const file = exceptionObject.GetFile();
+    const char * const description = exceptionObject.GetDescription();
 
-  ASSERT_NE(what, nullptr);
-  ASSERT_NE(file, nullptr);
-  ASSERT_NE(description, nullptr);
+    ASSERT_NE(what, nullptr);
+    ASSERT_NE(file, nullptr);
+    ASSERT_NE(description, nullptr);
 
-  EXPECT_EQ(what, file + (":" + std::to_string(exceptionObject.GetLine()) + ":\n") + description);
+    EXPECT_EQ(what, file + (":" + std::to_string(exceptionObject.GetLine()) + ":\n") + description);
+  }
+  {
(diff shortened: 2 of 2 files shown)


Modules/Core/Common/include/itkExceptionObject.h
```diff
@@ -131,6 +131,8 @@ class ITKCommon_EXPORT ExceptionObject : public std::exception
   what() const noexcept override;
 
 private:
+  static constexpr const char * m_DefaultLocation = "Unknown";
```
blowekamp, line 134: Please move this initialization to the CXX file. This will help ensure that the string only appears in the itkExceptionObject.o file and does not get duplicated anyplace else.

N-Dekker: @blowekamp Thanks, but does it really matter much? Moving `m_DefaultLocation = "Unknown"` to the CXX file would make it non-constexpr, right? So then wouldn't it be an extra burden for the linker?

"itkExceptionObject.h" already has another `static constexpr` literal string (`default_exception_message`), so I assumed that it would be ok  🤷 At https://github.com/InsightSoftwareConsortium/ITK/blob/ef37881d317b003121b9ff9c050affb6fa44b5e8/Modules/Core/Common/include/itkExceptionObject.h#L53

 `m_DefaultLocation` would replace two equal literal strings ("Unknown") in "itkExceptionObject.h", at https://github.com/InsightSoftwareConsortium/ITK/blob/ef37881d317b003121b9ff9c050affb6fa44b5e8/Modules/Core/Common/include/itkExceptionObject.h#L62-L66 So in a sense it would reduce the number of literal strings coming from the header file.

N-Dekker: By the way, it would be much easier if we would use an empty string as default, for the location. Then of course, a static `m_DefaultLocation` would not be necessary anymore. Because "empty" is a natural default state for a string. Would you like that as well?

Modules/Core/Common/src/itkExceptionObject.cxx
```diff
@@ -42,7 +42,12 @@ class ExceptionObject::ExceptionData
     , m_Line(line)
   {
     m_What = m_File;
-    m_What += ':' + std::to_string(m_Line) + ":\n";
+    m_What += ':' + std::to_string(m_Line) + ':';
+    if (!m_Location.empty() && m_Location != m_DefaultLocation)
+    {
+      m_What += " in '" + m_Location + "':";
+    }
```
N-Dekker, line 49: Now it avoids appending `" in 'Unknown':"`, when the location is `"Unknown"`. OK?  (We can also remove the condition `m_Location != m_DefaultLocation`, and then just get rid of `m_DefaultLocation` 🤷 )

Modules/Core/Common/include/itkExceptionObject.h
blowekamp: I don't know how each compiler handles the `static constexpr const char *` type and where the string will exist in object files. That is will it only occur in the itkExceptionObject.o file it will it occur in any compilation unit that uses this class? It certainly is not explicit like my suggestion. Perhaps you have more unsightly into this?

However quickly searching around my .o files for "Generic ExceptionObject" did not find it, but did find may occurrences of "Unknown", but that is not quite a unique string.

>So in a sense it would reduce the number of (duplicate) literal strings coming from the header file.

To my knowledge compilers already know its duplicate and only have one occourace in the data segment.

Thanks for the discussion.

N-Dekker: I see now, there are eight more `loc = "Unknown"` default parameters in ITK, in various other header files...

Modules/Core/Common/src/itkExceptionObject.cxx
N-Dekker: Just removed the condition `m_Location != m_DefaultLocation`!

Modules/Core/Common/include/itkExceptionObject.h
N-Dekker: So I realize now that the "Unknown" issue is beyond the scope of this pull request. Thanks Bradley! PR adjusted, please check!

blowekamp: Most file I would not make this comment about reducing duplicated strings on. But this file is used just about every where so it can have an impact.

dzenanz (approved):

dzenanz (approved):

thewtex (approved):
:mag:

PR #5269, opened 2025-03-08
STYLE: Replace "Unknown" default for exception location with `{}`

N-Dekker:
It is preferable not duplicate a literal string so many times (following [the DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)). Moreover, it appears sufficient to use an empty string value (`{}`), to indicate that the location of an exception is unspecified.

Note that an unspecified location may be "unknown", but instead, it may also be "irrelevant", or "hidden". So it may depend on the application.

- Triggered by a discussion with Bradley (@blowekamp) at https://github.com/InsightSoftwareConsortium/ITK/pull/5266#discussion_r1985469317

### DIFF
--- a/Modules/Core/Common/include/itkExceptionObject.h
+++ b/Modules/Core/Common/include/itkExceptionObject.h
@@ -59,7 +59,7 @@ public:
   explicit ExceptionObject(std::string  file,
                            unsigned int lineNumber = 0,
                            std::string  desc = "None",
-                           std::string  loc = "Unknown");
+                           std::string  loc = {});
 
   /** Copy-constructor. */
   ExceptionObject(const ExceptionObject &) noexcept = default;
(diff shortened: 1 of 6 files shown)


dzenanz (approved):

N-Dekker:
- Note: this pull request adjusts some code that is also adjusted (or removed) by pull request #5270.
 
So maybe first do the other pull request: #5270!

dzenanz (approved):

blowekamp:
> Note that this was without any compiler optimization. I guess an optimizer might produce exactly the same assembly code in either way. 🤷

I tried with optimization "-O3" and the code was the same for both.

N-Dekker:
> Should documentation be added that when the location is ""/{} that it will be rendered into a string as "Unknown"?

When the location is an empty string, it will not be printed! `ExceptionObject::Print` does not print an empty location. Look here:

https://github.com/InsightSoftwareConsortium/ITK/blob/c66eac0b42dfbf63101a42ee4fa74412cbd946cd/Modules/Core/Common/src/itkExceptionObject.cxx#L187-L190

I think that's just fine. Printing `"Location: Unknown"` isn't very helpful. OK?

So strictly speaking, this pull request might cause a small behavior change, as it may no longer print `"Location: Unknown"`. But in practice, the location is usually set by `ITK_LOCATION`, which is a _non-empty_ string. (For example, when using the `itkExceptionMacro`.)

PR #5270, opened 2025-03-08
STYLE: Remove Exception constructor overloads for `char *` arguments

N-Dekker:
Internally, `ExceptionObject` stores its file name, description, and location as `std::string`, so the corresponding constructor parameters might as well just be of type `std::string`.

Whenever such a parameter was originally declared `const std::string &`, this commit has removed the `const` and the `&`, in order to efficiently "move" the parameters from derived exception classes to their base class.

Note that this commit drops the support for null pointers as arguments.

### DIFF
--- a/Modules/Core/Common/include/itkDataObject.h
+++ b/Modules/Core/Common/include/itkDataObject.h
@@ -59,10 +59,7 @@ public:
   ~DataObjectError() noexcept override = default;
 
   /** Constructor. Needed to ensure the exception object can be copied. */
-  DataObjectError(const char * file, unsigned int lineNumber);
-
-  /** Constructor. Needed to ensure the exception object can be copied. */
-  DataObjectError(const std::string & file, unsigned int lineNumber);
+  DataObjectError(std::string file, unsigned int lineNumber);
 
   /** Copy constructor. Needed to ensure the exception object can be copied. */
   DataObjectError(const DataObjectError & orig) noexcept;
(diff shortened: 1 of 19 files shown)


dzenanz (approved):

hjmjohnson (approved):

N-Dekker:
Thanks for your approval, @hjmjohnson, @dzenanz I guess those redundant overloads for `const char *` were there originally for performance reasons, while doing C++98/03 compilation. With C++11, this has become obsolete, as it added fast move semantics to `std::string`.

As I noted, this commit drops the support for null pointers as arguments. For example:

```cpp
    // Exception description = nullptr. No longer supported with this commit.   
    throw itk::ExceptionObject(__FILE__, __LINE__, nullptr);
```

I think that's OK, because why would anyone want to use `nullptr` as description? (Meant as a rhetorical question 😺!) The original implementation treated `nullptr` just like an empty string. But then it's clearer to use literal `""` or a default-constructed `std::string`:

```cpp
    // Still supported with this commit, no problem. 
    throw itk::ExceptionObject(__FILE__, __LINE__, "");
    throw itk::ExceptionObject(__FILE__, __LINE__, std::string());
    throw itk::ExceptionObject(__FILE__, __LINE__, std::string{});
```
OK? Or do we still need to have legacy support (`#ifndef ITK_LEGACY_REMOVE`...) for null pointers as arguments?

thewtex (approved):
:rocket: :man_astronaut:

thewtex:
> those redundant overloads

+1 for just needing to support modern C++, `std::string`, sensible types

thewtex:
> those redundant overloads

+1 for just needing to support modern C++, `std::string`, sensible types

PR #5276, opened 2025-03-15
STYLE: Use override statements for C++11

hjmjohnson:
Describe function overrides using the override keyword from C++11.

-----
https://stackoverflow.com/questions/39932391/virtual-override-or-both-c
When you override a function you don't technically need to write either virtual
or override.

The original base class declaration needs the keyword virtual to mark it as
virtual.

In the derived class the function is virtual by way of having the ¹same type as
the base class function.

However, an override can help avoid bugs by producing a compilation error when
the intended override isn't technically an override. For instance, the function
type isn't exactly like the base class function. Or that a maintenance of the
base class changes that function's type, e.g. adding a defaulted argument.

In the same way, a virtual keyword in the derived class can make such a bug
more subtle by ensuring that the function is still virtual in the further
derived classes.

So the general advice is,

Use virtual for the base class function declaration.  This is technically
necessary.

Use override (only) for a derived class' override.  This helps maintenance.
-----
Remove 'virtual' is implied when 'override' is specified, so remove the
redundant specification.

cd ${BLDDIR}
run-clang-tidy.py -extra-arg=-D__clang__ -checks=-*,modernize-use-override  -header-filter=.* -fix

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkSmapsFileParser.h
+++ b/Modules/Core/Common/include/itkSmapsFileParser.h
@@ -291,13 +291,13 @@ template <typename TSmapsDataType>
 class ITK_TEMPLATE_EXPORT SmapsFileParser : public MapFileParser<TSmapsDataType>
 {
 public:
-  virtual ~SmapsFileParser() override = default;
+  ~SmapsFileParser() override = default;
   /** Load and parse the smaps file pointed by smapsFileLocation.
    *  If smapsFileLocation is empty, load the file located at
    *  "/proc/" + PID + "/smaps"
    *  Throw an exception is the file can't be opened.
    */
-  virtual void
+  void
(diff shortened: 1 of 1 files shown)


dzenanz (approved):

N-Dekker:
Thanks Hans!

Note that C++ Core Guidelines also has a guideline, saying [Virtual functions should specify exactly one of `virtual`, `override`, or `final`](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-override). So instead of copy-and-pasting an entire StackOverflow post, you might as well simply refer to that guideline, https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-override

thewtex (approved):
:100:

PR #5277, opened 2025-03-16
STYLE: Flip signs by `x *= -1`, instead of by `x = -x`

N-Dekker:
Avoided specifying the same expression twice (`<expression> = -<expression>`), and avoided evaluating that expression twice.

Using the following regular expression in Visual Studio:

    Find what:   (.+) = -\1;
    Replace with:   $1 *= -1;

### DIFF
--- a/Modules/Core/Common/include/itkVariableSizeMatrix.hxx
+++ b/Modules/Core/Common/include/itkVariableSizeMatrix.hxx
@@ -180,7 +180,7 @@ VariableSizeMatrix<T>::operator-()
   {
     for (unsigned int c = 0; c < this->Cols(); ++c)
     {
-      m_Matrix(r, c) = -m_Matrix(r, c);
+      m_Matrix(r, c) *= -1;
     }
   }
   return *this;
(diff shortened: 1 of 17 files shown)


dzenanz (approved):
And this is for readability only? It is not expected to have a performance impact?

Modules/Core/Common/include/itkVariableSizeMatrix.hxx
```diff
@@ -180,7 +180,7 @@ VariableSizeMatrix<T>::operator-()
   {
     for (unsigned int c = 0; c < this->Cols(); ++c)
     {
-      m_Matrix(r, c) = -m_Matrix(r, c);
+      m_Matrix(r, c) *= -1;
```
N-Dekker, line 183: @dzenanz This change _might_ have a positive performance impact, albeit a very, very small one (avoiding to call `m_Matrix(r, c)` twice).

For a local variable `var`, I don't expect any performance difference between `var = -var` and `van *= -1`.

In general, doing `<expression> *= -1` certainly won't harm  😃

N-Dekker: Honestly, for a variable `var`, the statement  `var = -var` doesn't bother me so much. I was triggered by statements like `direction[j][i] = -direction[j][i]`. (Where it takes an access operator or function call.) With this PR, I'm proposing to just use the form `<expression> *= -1` everywhere, for the sake of consistency,

PR #455, opened 2019-01-29
Meta data dictionary move and copy on write

blowekamp:

PR #5280, opened 2025-03-17
Avoid manual memory management `ImageSeriesReader::m_MetaDataDictionaryArray`

N-Dekker:
No longer manually allocated the MetaDataDictionary objects of `ImageSeriesReader` on the heap. Stored them in an internal `std::vector` instead.

Declared `ImageSeriesReader::m_MetaDataDictionaryArray` private, just like `ImageSeriesWriter::m_MetaDataDictionaryArray`.

### DIFF
--- a/Modules/IO/ImageBase/include/itkImageSeriesReader.hxx
+++ b/Modules/IO/ImageBase/include/itkImageSeriesReader.hxx
@@ -30,22 +30,6 @@
 
 namespace itk
 {
-// Destructor
-template <typename TOutputImage>
-ImageSeriesReader<TOutputImage>::~ImageSeriesReader()
-{
-  // Clear the eventual previous content of the MetaDictionary array
-  if (!m_MetaDataDictionaryArray.empty())
-  {
-    for (auto & i : m_MetaDataDictionaryArray)
-    {
-      // each element is a raw pointer, delete them.
-      delete i;
-    }
-  }
-  m_MetaDataDictionaryArray.clear();
-}
-
 template <typename TOutputImage>
 void
 ImageSeriesReader<TOutputImage>::PrintSelf(std::ostream & os, Indent indent) const
@@ -102,15 +86,8 @@ ImageSeriesReader<TOutputImage>::GenerateOutputInformation()
   const std::string key("ITK_ImageOrigin");
 
   // Clear the previous content of the MetaDictionary array
-  if (!m_MetaDataDictionaryArray.empty())
-  {
-    for (auto & i : m_MetaDataDictionaryArray)
-    {
-      // each element is a raw pointer, delete them.
-      delete i;
-    }
-  }
   m_MetaDataDictionaryArray.clear();
+  m_InternalMetaDataDictionaries.clear();
 
   const auto numberOfFiles = static_cast<int>(m_FileNames.size());
   if (numberOfFiles == 0)
@@ -297,6 +274,8 @@ ImageSeriesReader<TOutputImage>::GenerateData()
   double                             maxSpacingDeviation = 0.0;
   bool                               prevSliceIsValid = false;
 
+  m_InternalMetaDataDictionaries.reserve(static_cast<size_t>(numberOfFiles));
+
   for (int i = 0; i != numberOfFiles; ++i)
   {
     if (TOutputImage::ImageDimension != this->m_NumberOfDimensionsInImage)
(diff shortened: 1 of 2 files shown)


dzenanz (approved):

Modules/IO/ImageBase/include/itkImageSeriesReader.hxx
```diff
@@ -457,17 +434,20 @@ ImageSeriesReader<TOutputImage>::GenerateData()
...
-      *newDictionary = reader->GetImageIO()->GetMetaDataDictionary();
+      MetaDataDictionary newDictionary = reader->GetImageIO()->GetMetaDataDictionary();
       if (nonUniformSampling)
       {
         // slice-specific information
-        EncapsulateMetaData<double>(*newDictionary, "ITK_non_uniform_sampling_deviation", spacingDeviation);
+        EncapsulateMetaData<double>(newDictionary, "ITK_non_uniform_sampling_deviation", spacingDeviation);
       }
-      m_MetaDataDictionaryArray.push_back(newDictionary);
+      m_InternalMetaDataDictionaries.push_back(newDictionary);
```
blowekamp, line 443: Copying the dictionary is a non-trivial task. The prior implementation used pointer to ensure extra copying was not done. This now needs to manually be done.

This should be "emplace_back" to reduce copying the dictionary.  Also when the vector is resize, it will copy the dictionaries. The vector should reserve the expected number of dict ( aka number of slices.

blowekamp (commented):
If I am not mistaken this introduces extra copies to the dictionary.

Modules/IO/ImageBase/include/itkImageSeriesReader.hxx
N-Dekker: @blowekamp Thanks for expressing your concern, but I think you are mistaken. Copying a dictionary is actually pretty fast, now that it has "copy on write". For more than six years already: 

- Look at pull request #455 commit 327de92e76d7f228fbca6aaf6c8e679f6638d6ce (especially look at the author 😺)!

No problem, I can add a `vector::reserve` call, and a `std::move`, but I don't expect a major performance change 🤷

N-Dekker: @blowekamp Please check if your concern is taken away by this [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/b12865273b8fe1be5094684cdae4995ecc0a46bc..d7150cdf6f9fae17da67a0eedaf07505f2e081b2)

blowekamp: Ahh... Thanks for the link, lol. And thank you for using the move semantics which is fine, I agree it will not make much difference given the "brilliant" implement in the DataDictionary 🤣 


Side details discussion: I do find your C++ changes interesting and frequently learn things from them. Thank you. I suppose resizing and vector where the objects support move semantics is not too expensive, even without lazy copying.

blowekamp (approved):

Modules/IO/ImageBase/include/itkImageSeriesReader.hxx
```diff
@@ -457,17 +436,23 @@ ImageSeriesReader<TOutputImage>::GenerateData()
...
-      *newDictionary = reader->GetImageIO()->GetMetaDataDictionary();
+      MetaDataDictionary newDictionary = reader->GetImageIO()->GetMetaDataDictionary();
       if (nonUniformSampling)
       {
         // slice-specific information
-        EncapsulateMetaData<double>(*newDictionary, "ITK_non_uniform_sampling_deviation", spacingDeviation);
+        EncapsulateMetaData<double>(newDictionary, "ITK_non_uniform_sampling_deviation", spacingDeviation);
       }
-      m_MetaDataDictionaryArray.push_back(newDictionary);
+      m_InternalMetaDataDictionaries.push_back(std::move(newDictionary));
```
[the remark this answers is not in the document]
blowekamp: I think the usage of emplace_back would have been to remove the local variable... something like:
```
m_InternalMetaDataDictionaries.emplace_back(reader->GetImageIO()->GetMetaDataDictionary());
```

My thought had been to use emplace_back to avoid extra copies of the objects being created. For this particular object it doesn't master. And your implementation is nearly equivalent.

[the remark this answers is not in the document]
N-Dekker: > I think the usage of emplace_back would have been to remove the local variable... something like:
> 
> ```
> m_InternalMetaDataDictionaries.emplace_back(reader->GetImageIO()->GetMetaDataDictionary());
> ```

OK, but then it still wouldn't be much different from using push_back, as follows:
```
m_InternalMetaDataDictionaries.push_back(reader->GetImageIO()->GetMetaDataDictionary());
```

In either way (whether "emplacing" or "pushing"), it would always have to make a copy of the dictionary from the ImageIO object. 🤷 Fortunately copying a dictionary is very fast now, because of Copy On Write (your PR #455).

PR #5283, opened 2025-03-18
Global max threads during initialization

thewtex:
Clamp GlobalDefault threads with GlobalMaximum with initialization

 We must call GetGlobalDefaultNumberOfThreads() when clamping the value
 in SetGlobalMaximumNumberOfThreads, which has the side effect of
 initializing static variables so the desired clamping occurs.

This issue will only be observed before a threader like the
itk::PoolMultiThreader is instantiated because
GetGlobalDefaultNumberOfThreads() is called in its constructor.

Addresses #4773.

### DIFF
--- a/Modules/Core/Common/src/itkMultiThreaderBase.cxx
+++ b/Modules/Core/Common/src/itkMultiThreaderBase.cxx
@@ -217,8 +217,8 @@ MultiThreaderBase::SetGlobalMaximumNumberOfThreads(ThreadIdType val)
   m_PimplGlobals->m_GlobalMaximumNumberOfThreads = std::clamp<ThreadIdType>(val, 1, ITK_MAX_THREADS);
 
   // If necessary reset the default to be used from now on.
-  m_PimplGlobals->m_GlobalDefaultNumberOfThreads =
-    std::min(m_PimplGlobals->m_GlobalDefaultNumberOfThreads, m_PimplGlobals->m_GlobalMaximumNumberOfThreads);
+  MultiThreaderBase::SetGlobalDefaultNumberOfThreads(
+    std::min(Self::GetGlobalDefaultNumberOfThreads(), Self::GetGlobalMaximumNumberOfThreads()));
 }
 
 ThreadIdType
--- a/Modules/Core/Common/test/itkMultiThreaderBaseTest.cxx
+++ b/Modules/Core/Common/test/itkMultiThreaderBaseTest.cxx
@@ -103,6 +103,12 @@ SetAndVerify(int number)
 int
 itkMultiThreaderBaseTest(int argc, char * argv[])
 {
+  bool result = true;
+
+  itk::MultiThreaderBase::SetGlobalMaximumNumberOfThreads(1);
+  result &= itk::MultiThreaderBase::GetGlobalMaximumNumberOfThreads() == 1;
+  result &= itk::MultiThreaderBase::GetGlobalDefaultNumberOfThreads() == 1;
+
   // Choose a number of threads.
   int numberOfThreads = 10;
   if (argc > 1)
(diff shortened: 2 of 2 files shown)


Modules/Core/Common/test/itkMultiThreaderBaseTest.cxx
```diff
@@ -103,6 +103,12 @@ SetAndVerify(int number)
 int
 itkMultiThreaderBaseTest(int argc, char * argv[])
 {
+  bool result = true;
+
+  itk::MultiThreaderBase::SetGlobalMaximumNumberOfThreads(1);
+  result &= itk::MultiThreaderBase::GetGlobalMaximumNumberOfThreads() == 1;
```
dzenanz, line 109: Did this fail before this patch?

dzenanz (approved):
Looks good on a glance.

blowekamp (approved):

Modules/Core/Common/src/itkMultiThreaderBase.cxx
```diff
@@ -217,8 +217,7 @@ MultiThreaderBase::SetGlobalMaximumNumberOfThreads(ThreadIdType val)
   m_PimplGlobals->m_GlobalMaximumNumberOfThreads = std::clamp<ThreadIdType>(val, 1, ITK_MAX_THREADS);
 
   // If necessary reset the default to be used from now on.
-  m_PimplGlobals->m_GlobalDefaultNumberOfThreads =
-    std::min(m_PimplGlobals->m_GlobalDefaultNumberOfThreads, m_PimplGlobals->m_GlobalMaximumNumberOfThreads);
+  MultiThreaderBase::SetGlobalDefaultNumberOfThreads(std::min(GetGlobalMaximumNumberOfThreads(), m_PimplGlobals->m_GlobalMaximumNumberOfThreads));
```
N-Dekker, line 220: Thanks Matt, but is this as you intended? 

    std::min(GetGlobalMaximumNumberOfThreads(), m_PimplGlobals->m_GlobalMaximumNumberOfThreads))

That's just `min(Maximum, Maximum)`, right?

N-Dekker: So it unconditionally does `Default = Maximum`?

blowekamp: That is a good point. Perhaps we should reuse the setter method too:
```
Self::SetGlobalDefaultNumberOfThreads(Self::GetGlobalDefaultNumberOfThreads())
```

N-Dekker: Just for my understanding, what is the difference between the `GlobalDefault` and the `GlobalMaximum`, conceptually? When would a user set the `GlobalDefault`, and when would a user set the `GlobalMaximum`?

It seems to me that the `GlobalDefault` is the most important one, as in is used as the number of threads of the thread pool.

Modules/Core/Common/test/itkMultiThreaderBaseTest.cxx
thewtex: Yes

Modules/Core/Common/src/itkMultiThreaderBase.cxx
thewtex: Oops, yes it should be min(Default, Maximum).

When would a user set the GlobalDefault, and when would a user set the GlobalMaximum?

The Default defines the default value and the maximum defines a limit.

N-Dekker: > The Default defines the default value and the maximum defines a limit.

OK, thanks, but isn't the number of threads of the thread pool all that matters?

thewtex: @N-Dekker what do you mean?

N-Dekker: I mean, does ITK really need two global variables to specify the number of threads? Wouldn't it be sufficient to just have one GlobalNumberOfThreads variable?

thewtex: I think the intention is to support a limit on the number of threads that can be requested that is separate from the number of threads used by a filter.

thewtex: @N-Dekker an example use case may help:

- The maximum number of threads is 128, the number of cores on a many-core system.
- This is the maximum number of threads, but the default number of threads may be set to 12, which will result in better utilization of resources when processing a 256x256 image or when a simple filter is applied.
- Yet, more threads could be used where there is larger data or a more computationally complex filter.

hjmjohnson (approved):

PR #5284, opened 2025-03-19
ENH: Reset conj-grad learning rate for multi-start optimization

cookpa:
When used as a local optimizer in multi-start optimization, itkConjugateGradientLineSearchOptimizerv4 gives inconsistent results because its learningRate is carried over from the end of one optimization to the beginning of the next.

Added a function to store the initial learning rate when SetLearningRate is called - this is then restored on the first iteration when StartOptimization is called.

Added a test to check that repeated optimizations with the same starting parameters converge to the same result.

## PR Checklist

### DIFF
--- a/Modules/Numerics/Optimizersv4/include/itkConjugateGradientLineSearchOptimizerv4.h
+++ b/Modules/Numerics/Optimizersv4/include/itkConjugateGradientLineSearchOptimizerv4.h
@@ -78,6 +78,28 @@ public:
   void
   StartOptimization(bool doOnlyInitialization = false) override;
 
+  /**
+   * Set the learning rate for optimization. It is overridden by
+   * automatic learning rate estimation if enabled via a ScalesEstimator.
+   * The learning rate is updated during optimization according to the golden
+   * section line search, even if no ScalesEstimator is set.
+   *
+   * The learning rate set by this method is stored and re-used in subsequent
+   * calls to StartOptimization() unless overridden by automatic learning rate
+   * estimation.
+   */
+  void
+  SetLearningRate(TInternalComputationValueType learningRate) override;
+
+  /**
+   * Set/Get the initial learning rate, which is used in the first iteration of
+   * optimization unless overridden by automatic learning rate estimation.
+   * Setting the learning rate via SetLearningRate() will also change the initial
+   * learning rate for future optimizations.
+   */
+  itkSetMacro(InitialLearningRate, TInternalComputationValueType);
+  itkGetConstReferenceMacro(InitialLearningRate, TInternalComputationValueType);
+
 protected:
   /** Advance one Step following the gradient direction.
    * Includes transform update. */
@@ -94,8 +116,9 @@ protected:
   PrintSelf(std::ostream & os, Indent indent) const override;
 
 private:
-  DerivativeType m_LastGradient{};
-  DerivativeType m_ConjugateGradient{};
+  DerivativeType                m_LastGradient{};
+  DerivativeType                m_ConjugateGradient{};
+  TInternalComputationValueType m_InitialLearningRate{ 1 };
 };
 
 /** This helps to meet backward compatibility */
--- a/Modules/Numerics/Optimizersv4/include/itkConjugateGradientLineSearchOptimizerv4.hxx
+++ b/Modules/Numerics/Optimizersv4/include/itkConjugateGradientLineSearchOptimizerv4.hxx
@@ -38,6 +38,7 @@ void
 ConjugateGradientLineSearchOptimizerv4Template<TInternalComputationValueType>::StartOptimization(
   bool doOnlyInitialization)
 {
+  this->m_LearningRate = this->m_InitialLearningRate;
   this->m_ConjugateGradient.SetSize(this->m_Metric->GetNumberOfParameters());
   this->m_ConjugateGradient.Fill(TInternalComputationValueType{});
   this->m_LastGradient.SetSize(this->m_Metric->GetNumberOfParameters());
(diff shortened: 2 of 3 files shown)


cookpa:
As discussed here https://discourse.itk.org/t/unexpected-behaviour-in-multistartoptimizerv4/7466/1 with @dzenanz and others.

@ntustison this may affect ANTs in more than one place. It should improve antsAI by doing the multi-start optimizations more consistently, but I'm not sure if there's other places where this could hurt, eg if ANTs implicitly relies on the learning rate being carried over to new optimizations.

Modules/Numerics/Optimizersv4/include/itkConjugateGradientLineSearchOptimizerv4.h
```diff
@@ -94,8 +97,9 @@ class ITK_TEMPLATE_EXPORT ConjugateGradientLineSearchOptimizerv4Template
   PrintSelf(std::ostream & os, Indent indent) const override;
 
 private:
-  DerivativeType m_LastGradient{};
-  DerivativeType m_ConjugateGradient{};
+  DerivativeType                m_LastGradient{};
+  DerivativeType                m_ConjugateGradient{};
+  TInternalComputationValueType m_InitialLearningRate{};
```
N-Dekker, line 102: Would it be an option to directly initialize `m_InitialLearningRate` to `1` here? As follows:

    TInternalComputationValueType m_InitialLearningRate{ 1 };

jhlegarreta: Please, add a getter macro to `m_InitialLearningRate`. Exercise it in the test.

Modules/Numerics/Optimizersv4/include/itkConjugateGradientLineSearchOptimizerv4.hxx
```diff
@@ -101,6 +106,18 @@ ConjugateGradientLineSearchOptimizerv4Template<TInternalComputationValueType>::A
   this->InvokeEvent(IterationEvent());
 }
 
+/**
+ * Set the initial learning rate for optimization.
+ */
```
jhlegarreta, line 111: Please, move the documentation to the header file. Also, since getter and setter method docs come and are displayed in pairs, please modify the doc for this.

jhlegarreta (changes requested):
Thanks for doing this @cookpa.

A few comments inline.

Modules/Numerics/Optimizersv4/include/itkConjugateGradientLineSearchOptimizerv4.hxx
```diff
@@ -38,6 +38,12 @@ void
 ConjugateGradientLineSearchOptimizerv4Template<TInternalComputationValueType>::StartOptimization(
   bool doOnlyInitialization)
 {
+  // Initialize the learning rate if it has not been set
+  if (this->m_InitialLearningRate == NumericTraits<TInternalComputationValueType>::ZeroValue())
+  {
+    this->m_InitialLearningRate = NumericTraits<TInternalComputationValueType>::OneValue();
+  }
```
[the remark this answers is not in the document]
cookpa: Thanks, changed to initialize in the definition and removed the extra code

Modules/Numerics/Optimizersv4/include/itkConjugateGradientLineSearchOptimizerv4.h
cookpa: Thanks, I added set / get macros for m_InitialLearningRate, and exercise them in the tests.

Modules/Numerics/Optimizersv4/include/itkConjugateGradientLineSearchOptimizerv4.hxx
cookpa: Docstrings now moved to the definition

Modules/Numerics/Optimizersv4/include/itkConjugateGradientLineSearchOptimizerv4.h
```diff
@@ -78,6 +78,41 @@ class ITK_TEMPLATE_EXPORT ConjugateGradientLineSearchOptimizerv4Template
...
+   * automatic learning rate estimation if enabled via a ScalesEstimator.
+   * The learning rate is updated during optimization according to the golden
+   * section line search, even if no ScalesEstimator is set.
+   *
+   * The learning rate set by this method is stored and re-used in subsequent
+   * calls to StartOptimization() unless overridden by automatic learning rate
+   * estimation.
+   */
+  const TInternalComputationValueType &
+  GetLearningRate() const override;
```
jhlegarreta, line 105: Why do we need a getter method if it is defined in the base class? https://github.com/InsightSoftwareConsortium/ITK/blob/25d316237131813a37ff9c6eb4c634e6b55b2202/Modules/Numerics/Optimizersv4/include/itkGradientDescentOptimizerv4.h#L109-L114

jhlegarreta (commented):
Thanks for addressing the previous comments Philip.

I have an inline question.

Modules/Numerics/Optimizersv4/include/itkConjugateGradientLineSearchOptimizerv4.h
cookpa: The method is not really needed but I didn't know another way to make the Set / Get docstrings consistent as requested. Is there a better way to accomplish this?

jhlegarreta: Please, remove the method then. The comment about the Set/Get docstring only applies to cases where both methods are present.

cookpa: It's removed

jhlegarreta (approved):

hjmjohnson (approved):
This makes a lot of sense to me.  If you reset the initial state to all zeros again, one should reset the learning rate to it's staring position as well.

PR #5287, opened 2025-03-21
Deterministic multithreading usage of itkMersenneTwisterRandomVariate…

michalmeszaroswork:
…Generator.cxx

When using ITK library along with SuperElastix
https://github.com/SuperElastix/elastix
by running multiple elastix::ELASTIX objects in different threads such as :

const auto runElastix = [](){
elastix::ELASTIX elastixFilter;
int error = elastixFilter.RegisterImages(
                    fixedImg,
                    movingImg,
                    propertiesMap,
                    outDirPath,
                    false,
                    false,
                    nullptr,
                    nullptr);
}

    std::vector<std::thread> threads;
    for (int i = 0; i <= 1; ++i) {
        threads.emplace_back(runElastix);
    }

    for (auto &t : threads) {
        t.join();
    }

they share the same instance of MersenneTwisterRandomVariateGenerator, which cause that results are nondeterministic. This simple fix stabilize results along multiple threads and does not seems to break anything in ITK.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/src/itkMersenneTwisterRandomVariateGenerator.cxx
+++ b/Modules/Core/Common/src/itkMersenneTwisterRandomVariateGenerator.cxx
@@ -34,7 +34,6 @@ struct MersenneTwisterGlobals
   MersenneTwisterGlobals() = default;
   ~MersenneTwisterGlobals() = default;
 
-  MersenneTwisterRandomVariateGenerator::Pointer                  m_StaticInstance{};
   std::mutex                                                      m_StaticInstanceMutex{};
   std::atomic<MersenneTwisterRandomVariateGenerator::IntegerType> m_StaticDiffer{};
 };
@@ -74,13 +73,14 @@ MersenneTwisterRandomVariateGenerator::GetInstance()
   itkInitGlobalsMacro(PimplGlobals);
   const std::lock_guard<std::mutex> lockGuard(m_PimplGlobals->m_StaticInstanceMutex);
 
-  if (!m_PimplGlobals->m_StaticInstance)
+  thread_local MersenneTwisterRandomVariateGenerator::Pointer threadSafeInstance{};
+  if (!threadSafeInstance)
   {
-    m_PimplGlobals->m_StaticInstance = MersenneTwisterRandomVariateGenerator::CreateInstance();
-    m_PimplGlobals->m_StaticInstance->SetSeed();
+    threadSafeInstance = MersenneTwisterRandomVariateGenerator::CreateInstance();
+    threadSafeInstance->SetSeed();
   }
 
-  return m_PimplGlobals->m_StaticInstance;
+  return threadSafeInstance;
 }
 
 


dzenanz:
This breaks the singleton mechanics, e.g. setting a seed from one thread will now only apply that seed to that thread. Niels, does this change break any tests in Elastix?

N-Dekker:
Interesting, thanks @michalmeszaroswork! I just gave it a try: https://github.com/SuperElastix/elastix/tree/michalmeszaroswork-patch-1 (should take an hour or so to pass the test).

blowekamp:
Nice job identifying the dependency onto global random.  However, change the behavior of this may not be desirable by all even if it does not immediately "break" things. Let us consider another option.

The alternative is that each algorithm initializes its own random number generate ( either seeded by a user provided value, clock or the next seed from the global ).  Using an algorithm specific generator removes the dependency on the global ( even a thread local is still a global and could cause issues is some cases, especially with certain thread pool or task based threading models.)


> This breaks the singleton mechanics, e.g. setting a seed from one thread will now only apply that seed to that thread. Niels, does this change break any tests in Elastix?

Modules/Core/Common/src/itkMersenneTwisterRandomVariateGenerator.cxx
```diff
@@ -74,13 +73,14 @@ MersenneTwisterRandomVariateGenerator::GetInstance()
...
+  if (!threadSafeInstance)
   {
-    m_PimplGlobals->m_StaticInstance = MersenneTwisterRandomVariateGenerator::CreateInstance();
-    m_PimplGlobals->m_StaticInstance->SetSeed();
+    threadSafeInstance = MersenneTwisterRandomVariateGenerator::CreateInstance();
+    threadSafeInstance->SetSeed();
   }
 
-  return m_PimplGlobals->m_StaticInstance;
+  return threadSafeInstance;
```
N-Dekker, line 83: Nitpick: I would name it `threadLocalInstance`, rather than `threadSafeInstance`. The original instance `m_StaticInstance` was thread-safe already, but your essential change is to make it _thread local_.

Then I would suggest to have it initialized during its declaration, so that it can be `const`:
```cpp
  const thread_local MersenneTwisterRandomVariateGenerator::Pointer threadLocalInstance = [] {
    MersenneTwisterRandomVariateGenerator::Pointer instance = MersenneTwisterRandomVariateGenerator::CreateInstance();
    instance->SetSeed();
    return instance;
  }();

  return threadLocalInstance;
```

N-Dekker:
> The alternative is that each algorithm initializes its own random number generate ( either seeded by a user provided value, clock or the next seed from the global ). Using an algorithm specific generator removes the dependency on the global ( even a thread local is still a global and could cause issues is some cases, especially with certain thread pool or task based threading models.)

Interesting, @blowekamp . I guess for elastix we should then replace or rewrite ITK's random iterators, right? elastix's `ImageRandomSampler` currently still uses ITK's `ImageRandomConstIteratorWithIndex`, for example, which does `MersenneTwisterRandomVariateGenerator::New()` during its initialization. `MersenneTwisterRandomVariateGenerator::New()` calls `GetNextSeed()`, adjusting the global (shared) state. 🤷

blowekamp:
> > The alternative is that each algorithm initializes its own random number generate ( either seeded by a user provided value, clock or the next seed from the global ). Using an algorithm specific generator removes the dependency on the global ( even a thread local is still a global and could cause issues is some cases, especially with certain thread pool or task based threading models.)
> 
> Interesting, @blowekamp . I guess for elastix we should then replace or rewrite ITK's random iterators, right? elastix's `ImageRandomSampler` currently still uses ITK's `ImageRandomConstIteratorWithIndex`, for example, which does `MersenneTwisterRandomVariateGenerator::New()` during its initialization. `MersenneTwisterRandomVariateGenerator::New()` calls `GetNextSeed()`, adjusting the global (shared) state. 🤷

So the initial issue is that if you run the algorithm (elastix) concurrently in multiple threads without setting the seed, then the behavior is randomized? Are there seeds available to set per algorithm or the only option is setting a global seed?

Have you looked ImageRegistrationMethodv4? A seed can be set for each instance of the class, and it uses it own MersenneTwisterRandomVariateGenerator to generate seeds for each of it's components.

N-Dekker:
@blowekamp Even when each thread has its own instance of `MersenneTwisterRandomVariateGenerator`, they still share the one static variable `m_StaticDiffer`: https://github.com/InsightSoftwareConsortium/ITK/blob/db104239cf898dac4b46a89ccaaa4fe9a271a441/Modules/Core/Common/src/itkMersenneTwisterRandomVariateGenerator.cxx#L39

I guess that might still cause nondeterministic behavior, right? Even if it is "thread-safe" (atomic), it may still affect the behavior when one thread does `m_StaticDiffer++` before or after another thread. 🤷

blowekamp:
> I guess that might still cause nondeterministic behavior, right? Even if it is "thread-safe" (atomic), it may still affect the behavior when one thread does `m_StaticDiffer++` before or after another thread. 🤷

My understanding is that m_StaticDiffer is use to set the seeds sequentially. My recommendations above are to explicitly set the seeds for the algorithm, and not rely on the global state.

N-Dekker:
> My understanding is that m_StaticDiffer is use to set the seeds sequentially. My recommendations above are to explicitly set the seeds for the algorithm, and not rely on the global state.

Thanks @blowekamp, I see now, two different generators produce the same random number sequence, when they have the same initial seed:

```cpp
using itk::Statistics::MersenneTwisterRandomVariateGenerator;

const auto generator1 = MersenneTwisterRandomVariateGenerator::New();
const auto generator2 = MersenneTwisterRandomVariateGenerator::New();

generator1->SetSeed(42);
generator2->SetSeed(42);

std::cout << generator1->GetIntegerVariate() << '\n'
          << generator2->GetIntegerVariate() << '\n'
          << generator1->GetIntegerVariate() << '\n'
          << generator2->GetIntegerVariate() << '\n'
          << generator1->GetIntegerVariate() << '\n'
          << generator2->GetIntegerVariate() << '\n';
```

Output:
> 1608637542
> 1608637542
> 3421126067
> 3421126067
> 4083286876
> 4083286876

👍 

I think it's a bit unfortunate that the global `m_StaticDiffer` is still modified during the creation of those generators. Maybe it would be nice to have an extra `MersenneTwisterRandomVariateGenerator::Create` function that would not modify the global data. What do you think?

blowekamp:
> I think it's a bit unfortunate that the global `m_StaticDiffer` is still modified during the creation of those generators. Maybe it would be nice to have an extra `MersenneTwisterRandomVariateGenerator::Create` function that would not modify the global data. What do you think?

Yes. That makes sense to me.

dzenanz:
From the discussion here, it looks like this PR is not a good solution. Have we identified an alternative solution? If so, who should propose a PR? Or should this PR be abandoned?

N-Dekker:
> Have we identified an alternative solution? If so, who should propose a PR? Or should this PR be abandoned?

I'm still looking into [elastix](https://github.com/SuperElastix/elastix), trying to replace all those `GetInstance()` calls, but it's not so easy 🤷 I will comment here again when it works (_if_ it works).

N-Dekker:
@michalmeszaroswork Thanks again for your work on this topic. Do you agree that with the recent merge of pull request https://github.com/SuperElastix/elastix/pull/1323, your ITK pull request is no longer necessary, for elastix?

hjmjohnson:
Fixes made in https://github.com/SuperElastix/elastix/pull/1323 make this PR unnecessary.

PR #5289, opened 2025-03-26
Remove magic numbers `0xffffffff`, `4294967295`, and `4294967296` from MersenneTwisterRandomVariateGenerator

N-Dekker:
Three style improvements within `MersenneTwisterRandomVariateGenerator` member functions:

- Removed unnecessary `& 0xffffffffUL` from `Initialize`
- Replaced multiplication by `1 / 4294967296.0` with division by `double{ 1ULL << 32 }` in `GetVariateWithOpenUpperRange` and `GetVariateWithOpenRange` 
- Replaced `4294967295` with `numeric_limits<uint32_t>::max()` in `GetVariateWithClosedRange`

Following C++ Core Guidelines, [Avoid “magic constants”](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-magic)

### DIFF
--- a/Modules/Core/Common/include/itkMersenneTwisterRandomVariateGenerator.h
+++ b/Modules/Core/Common/include/itkMersenneTwisterRandomVariateGenerator.h
@@ -345,10 +345,10 @@ MersenneTwisterRandomVariateGenerator::Initialize(const IntegerType seed)
   IntegerType * s = state;
   IntegerType * r = state;
 
-  *s++ = seed & 0xffffffffUL;
+  *s++ = seed;
   for (IntegerType i = 1; i < MersenneTwisterRandomVariateGenerator::StateVectorLength; ++i)
   {
-    *s++ = (1812433253UL * (*r ^ (*r >> 30)) + i) & 0xffffffffUL;
+    *s++ = uint32_t{ (uint32_t{ 1812433253 } * (*r ^ (*r >> 30)) + i) };
     ++r;
   }
   reload();
--- a/Modules/Core/Common/include/itkMersenneTwisterRandomVariateGenerator.h
+++ b/Modules/Core/Common/include/itkMersenneTwisterRandomVariateGenerator.h
@@ -29,6 +29,7 @@
 #include <mutex>
 #include <climits>
 #include <ctime>
+#include <limits>
 
 namespace itk
 {
@@ -430,7 +431,7 @@ MersenneTwisterRandomVariateGenerator::GetIntegerVariate()
 inline double
 MersenneTwisterRandomVariateGenerator::GetVariateWithClosedRange()
 {
-  return static_cast<double>(GetIntegerVariate()) * (1.0 / 4294967295.0);
+  return static_cast<double>(GetIntegerVariate()) * (1.0 / double{ std::numeric_limits<uint32_t>::max() });
 }
 
 /** Get a random variate in the range [0, n] */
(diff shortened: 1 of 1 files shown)


Modules/Core/Common/include/itkMersenneTwisterRandomVariateGenerator.h
```diff
@@ -430,7 +431,7 @@ MersenneTwisterRandomVariateGenerator::GetIntegerVariate()
 inline double
 MersenneTwisterRandomVariateGenerator::GetVariateWithClosedRange()
 {
-  return static_cast<double>(GetIntegerVariate()) * (1.0 / 4294967295.0);
+  return static_cast<double>(GetIntegerVariate()) * (1.0 / double{ std::numeric_limits<uint32_t>::max() });
```
hjmjohnson, line 434: perhaps another minor optimization (probably done already by the compiler):

```c++
static constexpr double rangeScaler = (1.0 / double{ std::numeric_limits<uint32_t>::max() });
return static_cast<double>(GetIntegerVariate())*rangeScalar;
```

N-Dekker:
> My only very very minor concern is that this is a re-implementation of some upstream code somewhere that we might be deviating from. We are clearly NOT tracking with the upstream code anyway, and it is a very stable implementation.

I see your point, thanks. However, we obviously integrated `MersenneTwisterRandomVariateGenerator` so much into ITK that I think it's our responsibility to keep the code up-to-date with modern coding standards. This PR uses some modern C++ features that were not yet available when the generator was added to ITK. For example, using C++11 conversion like `double{ x }` and `uint32_t{ x }`, that are guaranteed lossless. Also, at the time when the generator was added, `numeric_limits<T>::max()` wasn't yet `constexpr`.

I think the original code used the expression `x & 0xffffffffUL` just to ensure that the result fits in 32 bits. But this can now be ensures _at compile-time_ by the syntax `uint32_t{ x }`.

Anyway, I think it's already OK to you  😃

blowekamp:
> They are well tested in the test suite and show no changes in results.

I am not sure that we have any tests that would be sensitive to the distribution or reproducibility of the generated random sequence.

My experience with modify random number generators has been poor. Changes that at first glance appear to be equivalent have produce different random-ness results.

Modules/Core/Common/include/itkMersenneTwisterRandomVariateGenerator.h
```diff
@@ -345,10 +346,10 @@ MersenneTwisterRandomVariateGenerator::Initialize(const IntegerType seed)
   IntegerType * s = state;
   IntegerType * r = state;
 
-  *s++ = seed & 0xffffffffUL;
+  *s++ = seed;
   for (IntegerType i = 1; i < MersenneTwisterRandomVariateGenerator::StateVectorLength; ++i)
   {
-    *s++ = (1812433253UL * (*r ^ (*r >> 30)) + i) & 0xffffffffUL;
+    *s++ = uint32_t{ (uint32_t{ 1812433253 } * (*r ^ (*r >> 30)) + i) };
```
N-Dekker, line 352: Note that the effect of the original `UL` suffix is platform dependent. On Linux, it's typically 64 bits, and on Windows 32 bits. _Probably_ it doesn't matter, but it's much clearer to explicitly convert to `uint32`.

dzenanz (approved):

dzenanz:
Should we merge this, or is there something else to do?

N-Dekker:
> Should we merge this, or is there something else to do?

I think it's ready. Further future improvements may be possible, but for now I think this should be OK.

PR #5296, opened 2025-04-03
STYLE: Let MersenneTwisterRandomVariateGenerator pass integers by value

N-Dekker:
Following C++ Core Guidelines, [For “in” parameters, pass cheaply-copied types by value and others by reference to `const`](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-in)

### DIFF
--- a/Modules/Core/Common/include/itkMersenneTwisterRandomVariateGenerator.h
+++ b/Modules/Core/Common/include/itkMersenneTwisterRandomVariateGenerator.h
@@ -204,7 +204,7 @@ public:
 
   /** Get an integer variate in [0, n] for n < 2^32 */
   IntegerType
-  GetIntegerVariate(const IntegerType & n);
+  GetIntegerVariate(const IntegerType n);
 
   /** Access to 53-bit random numbers (capacity of IEEE double precision)
    * in the range [0,1) */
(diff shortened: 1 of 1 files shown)


dzenanz (approved):

Modules/Core/Common/include/itkMersenneTwisterRandomVariateGenerator.h
```diff
@@ -204,7 +204,7 @@ class ITKCommon_EXPORT MersenneTwisterRandomVariateGenerator : public RandomVari
 
   /** Get an integer variate in [0, n] for n < 2^32 */
   IntegerType
-  GetIntegerVariate(const IntegerType & n);
+  GetIntegerVariate(const IntegerType n);
```
N-Dekker, line 207: I'm still thinking, shouldn't we explicitly use `uint32_t`, instead of "IntegerType". It appears essential to MersenneTwisterRandomVariateGenerator that the integer type is a 32-bit unsigned integer. 🤷

dzenanz: While we might, it is a potentially breaking interface change, whereas the current proposal is not.

N-Dekker: OK, thanks!

In the future, I hope we will use [std::mersenne_twister_engine](https://en.cppreference.com/w/cpp/numeric/random/mersenne_twister_engine) anyway  😃

PR #5086, opened 2024-12-17
Add constexpr to compile time const variables

hjmjohnson:
IDE can identify these cases, but code changes are implemented manually

If statement with compile-time constant can be replaced with "if constexpr"
    

```
    REPLACE: (^ *)const *([a-zA-Z0-9_:]* [0-9a-zA-Z_]* = *[\+\-e0-9\.]+;)
       WITH: $1constexpr $2
    
    REPLACE: (^ *)const (\w* \w*\{\}\;)
       WITH: $1constexpr $2
```

## PR Checklist
- [x] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [x] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)

PR #5299, opened 2025-04-10
COMP: Remove constexpr from check on runtime value in JPEGImageIO

WangWeiLin-MV:
Remove constexpr for runtime value `cinfo`.

Fixed a Visual C++ x86 (32-bit) error:
```log
error C2131: expression did not evaluate to a constant
```

### DIFF
--- a/Modules/IO/JPEG/src/itkJPEGImageIO.cxx
+++ b/Modules/IO/JPEG/src/itkJPEGImageIO.cxx
@@ -364,11 +364,14 @@ JPEGImageIO::ReadImageInformation()
 
   // jpeg_calc_output_dimensions to calculate cinfo.output_components
   jpeg_calc_output_dimensions(&cinfo);
-  if constexpr (sizeof(void *) < 8 && (static_cast<unsigned long long>(cinfo.output_width) * cinfo.output_height *
-                                       cinfo.output_components) > 0xffffffff)
+  if constexpr (sizeof(void *) < 8)
   {
-    jpeg_destroy_decompress(&cinfo);
-    itkExceptionMacro("JPEG image is too big " << this->GetFileName());
+    if ((static_cast<unsigned long long>(cinfo.output_width) * cinfo.output_height * cinfo.output_components) >
+        0xffffffff)
+    {
+      jpeg_destroy_decompress(&cinfo);
+      itkExceptionMacro("JPEG image is too big " << this->GetFileName());
+    }
   }
 
   // pull out the width/height


Modules/IO/JPEG/src/itkJPEGImageIO.cxx
```diff
@@ -364,8 +364,8 @@ JPEGImageIO::ReadImageInformation()
 
   // jpeg_calc_output_dimensions to calculate cinfo.output_components
   jpeg_calc_output_dimensions(&cinfo);
-  if constexpr (sizeof(void *) < 8 && (static_cast<unsigned long long>(cinfo.output_width) * cinfo.output_height *
```
dzenanz, line 367: Which compiler complains about this?

issakomi: I think any compiler on x86 (32 bit) will complain. On x86_64 the second part of the expression (cinfo) is unreachable.

N-Dekker: - Interesting... the `constexpr` was added by pull request #5086 commit e1e2cd10d926c1bc366edf3fd29c2f95f47a9d83 (by Hans, @hjmjohnson), on Dec 18, 2024. So apparently 32-bit compilation is broken for a while already!

A `constexpr`-preserving alternative could be:

```cpp
  if constexpr (sizeof(void *) < 8)
  {
    if ((static_cast<unsigned long long>(cinfo.output_width) * cinfo.output_height * cinfo.output_components) >
        0xffffffff)
    {
      jpeg_destroy_decompress(&cinfo);
      itkExceptionMacro("JPEG image is too big " << this->GetFileName());
    }
  }
```

But that's just a minor suggestion 🤷  Thanks @WangWeiLin-MV !

My two cents, Niels

N-Dekker (approved):
Thanks again @WangWeiLin-MV!  Some more minor suggestions:

When a commit addresses run-time problems, the commit message typically starts with "BUG: ". But in your case, it addresses _compile errors_, so then it's more appropriate to start with "COMP: "

It's helpful to us to make the commit subject line more specific, for example:

    COMP: Remove constexpr from check on runtime value in JPEGImageIO

(The commit subject line may be up to 72 characters.)

It's helpful to mention the compile error message in the commit text. For example:

    Fixed a Visual C++ x84 error, saying
    > error C2131: expression did not evaluate to a constant

Approved anyway  😃

WangWeiLin-MV:
@N-Dekker Thanks for your suggestions, updated.

N-Dekker:
> @N-Dekker Thanks for your suggestions, updated.

You're welcome! If you think it's ready now, please mark your pull request "Ready for review"!

N-Dekker (approved):
🎉

dzenanz (approved):

PR #5308, opened 2025-04-15
BUG: GDCM destroy JPEG2000 structures on error

blowekamp:
When an unsupported JPEG2000 encoding was encountered, resources need to be release. This addresses issue with repeated reading problematic files allocated threads until system limits were reached resulting in deadlock.

## PR Checklist

### DIFF
--- a/Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmJPEG2000Codec.cxx
+++ b/Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmJPEG2000Codec.cxx
@@ -1484,7 +1484,8 @@ bool JPEG2000Codec::GetHeaderInfo(const char * dummy_buffer, size_t buf_size, Tr
     &image) ? true : false;
   if(!bResult)
   {
-  opj_stream_destroy(cio);
+    opj_destroy_codec(dinfo);
+    opj_stream_destroy(cio);
     return false;
   }
   //image = opj_decode(dinfo, cio);
@@ -1527,6 +1528,8 @@ bool JPEG2000Codec::GetHeaderInfo(const char * dummy_buffer, size_t buf_size, Tr
 
   if( !check_comp_valid( image ) )
     {
+    opj_destroy_codec(dinfo);
+    opj_stream_destroy(cio);
     gdcmErrorMacro( "Invalid test failed" );
     return false;
     }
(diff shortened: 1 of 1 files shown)


dzenanz (approved):

Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmJPEG2000Codec.cxx
```diff
@@ -1484,7 +1484,8 @@ bool JPEG2000Codec::GetHeaderInfo(const char * dummy_buffer, size_t buf_size, Tr
     &image) ? true : false;
   if(!bResult)
   {
-  opj_stream_destroy(cio);
+    opj_destroy_codec(dinfo);
+    opj_stream_destroy(cio);
     return false;
```
N-Dekker, line 1489: Should these structures be "destroyed" with each and every return or exit (possibly by an exception) out of the function? Then maybe the C++ idiom [RAII](https://en.cppreference.com/w/cpp/language/raii) would be preferable.  For example, `dinfo` and could be declared as `unique_ptr` with a custom deleter function, being `opj_destroy_codec` or `opj_stream_destroy`, respectively.

blowekamp: In a bug fix to a third party library, I believe it is appropriate to follow existing conventions and not introduce new approaches to maintain uniformity in the third party library.

N-Dekker: Thanks for your reply, @blowekamp However, GDCM already follows the RAII idiom, and it already uses quite a few `unique_ptr`s, so my suggestion isn't _that_ new, I think. Your current approach may not be exception safe. (And GDCM _does_ use exceptions.)

N-Dekker: I think the preferable solution would be to declare both `dinfo` and `cio` as a `unique_ptr` (with custom deleter). But another approach could be to just add two local variables that takes care destroying `dinfo` and `cio`:


```cpp
// Destroys dinfo when going out of scope
const std::unique_ptr<opj_codec_t, decltype(&opj_destroy_codec)> dinfoDestroyer(dinfo, &opj_destroy_codec);

// Destroys cio when going out of scope
const std::unique_ptr<opj_stream_t, decltype(&opj_stream_destroy)> cioDestroyer(cio, &opj_stream_destroy);


```

blowekamp: But that is not how the library is currently using the JPEG2000 library, and this is a bug fix not a refactor of the third party library.

blowekamp:
We have not heard from upstream in 2 weeks. Should be consider merging?

dzenanz:
Yes!

PR #5316, opened 2025-04-18
STYLE: Remove `this->` on access to private data ResourceProbe

N-Dekker:
Replaced `this->m_` with `m_` in "itkResourceProbe.hxx". Removed parentheses
from `++(m_NumberOfStarts)` and `++(m_NumberOfStops)`.

Following [section "Accessing Members" of ITK's Coding Style Guide](https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/blob/v5.4.3/SoftwareGuide/Latex/Appendices/CodingStyleGuide.tex#L1708)

### DIFF
--- a/Modules/Core/Common/include/itkResourceProbe.hxx
+++ b/Modules/Core/Common/include/itkResourceProbe.hxx
@@ -91,17 +91,17 @@ template <typename ValueType, typename MeanType>
 void
 ResourceProbe<ValueType, MeanType>::Reset()
 {
-  this->m_TotalValue = ValueType{};
-  this->m_StartValue = ValueType{};
-  this->m_MinimumValue = NumericTraits<ValueType>::max();
-  this->m_MaximumValue = NumericTraits<ValueType>::NonpositiveMin();
-  this->m_StandardDeviation = ValueType{};
+  m_TotalValue = ValueType{};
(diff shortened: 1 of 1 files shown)


dzenanz (approved):
The reason I sometimes had to add `this->` was that GCC threw compile errors without it. Hopefully this PR doesn't run into that issue.

N-Dekker:
> The reason I sometimes had to add `this->` was that GCC threw compile errors without it.

@dzenanz Do you remember in what kind of scenario those GCC compile errors appeared? I think it should be OK to just remove the `this->` from access to data members, _unless_ it's inside a template class, accessing the non-private data of its base class. In this case (this PR), it's just access to its own private data, so that should be just fine.

By the way, _if_ it would be access to the non-private data of its base class, I think it would be clearer to do it by `Superclass::m_Data` (instead of `this->m_Data`). Right?

dzenanz:
Yes, this issue arises in templated classes (sometimes). You might be completely right.

N-Dekker:
> Yes, this issue arises in templated classes (sometimes). You might be completely right.

Thanks @dzenanz For the sake of completeness, it appears that those GCC compile errors only appear when _both_ the derived _and_ the base class is templated. _And_ the derived tried to access a non-private data member from its base.

PR #5318, opened 2025-04-22
MersenneTwisterRandomVariateGenerator: Simplify `GetNextSeed()`, and other style improvements

N-Dekker:
Five little style improvements in the implementation of MersenneTwisterRandomVariateGenerator.

### DIFF
--- a/Modules/Core/Common/src/itkMersenneTwisterRandomVariateGenerator.cxx
+++ b/Modules/Core/Common/src/itkMersenneTwisterRandomVariateGenerator.cxx
@@ -262,10 +262,9 @@ MersenneTwisterRandomVariateGenerator::PrintSelf(std::ostream & os, Indent inden
   // Print state vector contents
   os << indent << "State vector: " << m_State << std::endl;
   os << indent;
-  const IntegerType * s = m_State;
-  int                 i = StateVectorLength;
-  for (; i--; os << *s++ << '\t')
+  for (const IntegerType stateValue : m_State)
   {
+    os << stateValue << '\t';
   }
   os << std::endl;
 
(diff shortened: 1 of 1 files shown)


dzenanz (approved):

Modules/Core/Common/src/itkMersenneTwisterRandomVariateGenerator.cxx
```diff
@@ -175,11 +171,11 @@ MersenneTwisterRandomVariateGenerator::Initialize(const IntegerType seed)
...
-  IntegerType * r = m_State;
+  IntegerType *       s = m_State;
+  const IntegerType * r = m_State;
 
   *s++ = seed;
-  for (IntegerType i = 1; i < MersenneTwisterRandomVariateGenerator::StateVectorLength; ++i)
+  for (IntegerType i = 1; i < StateVectorLength; ++i)
   {
     *s++ = uint32_t{ (uint32_t{ 1812433253 } * (*r ^ (*r >> 30)) + i) };
     ++r;
```
N-Dekker, line 181: Honestly it's not clear to me why it uses two pointers, `s` and `r`, instead of one. 🤷 Both pointers always just point to the same element of `m_State`.

dzenanz: Maybe kick out `r` pointer and leave only `s`?

N-Dekker: I just hesitated because the original C++ implementation appeared to have both an `s` and an `r`. At least, if this is the original implementation: https://www.cs.rpi.edu/academics/courses/fall12/ds/hw/05_multi_linked_lists/MersenneTwister.h

	register uint32 *s = state;
	register uint32 *r = state;
	register int i = 1;
	*s++ = seed & 0xffffffffUL;
	for( ; i < N; ++i )
	{
		*s++ = ( 1812433253UL * ( *r ^ (*r >> 30) ) + i ) & 0xffffffffUL;
		r++;
	}

But we did already make quite some style improvements compared to that one, anyway.

PR #5320, opened 2025-04-23
Add `constexpr` to `if` when comparing dimensions, remove a few unnecessary `static_cast`'s

N-Dekker:
A few style commits, using the fact that dimensions are defined as `constexpr unsigned int`.


dzenanz (approved):

hjmjohnson (approved):
Very nice!

PR #5324, opened 2025-04-24
STYLE: Replace `static_cast<T>(Dimension)` with `T{ Dimension }`

N-Dekker:
Replaced `static_cast<(\w+)>\(([A-Za-z]*Dimension)\)` with `$1{ $2 }`, using regular expressions.

Excluded [`static_cast<double>(SizeAccumulateDimension)`](https://github.com/InsightSoftwareConsortium/ITK/blob/3c574f611d983e3c1f763d630a48a30c87940cd9/Modules/Filtering/ImageStatistics/include/itkAccumulateImageFilter.hxx#L162), as it is not a compile-time constant expression.

Following C++ Core Guidelines, ["Avoid casts"](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es48-avoid-casts)

----

Unlike `static_cast<T>(Dimension)`, `T{ Dimension }` checks at compile-time that the dimension can be converted losslessly, for example: https://godbolt.org/z/Eoc9fhnnz

```cpp
    static constexpr unsigned int SmallDimension{ 2 };
    static constexpr unsigned int LargeDimension{ UINT_MAX };

    return static_cast<int>(SmallDimension); // OK
    return static_cast<int>(LargeDimension); // overflow, but no error!
    return int{ SmallDimension }; // OK, this conversion is safe.
    return int{ LargeDimension }; // Compile error!

```

### DIFF
--- a/Modules/Core/Common/include/itkConstNeighborhoodIterator.hxx
+++ b/Modules/Core/Common/include/itkConstNeighborhoodIterator.hxx
@@ -185,7 +185,7 @@ ConstNeighborhoodIterator<TImage, TBoundaryCondition>::ComputeInternalIndex(cons
 {
   OffsetType ans;
   auto       r = static_cast<unsigned long>(n);
-  for (long i = static_cast<long>(Dimension) - 1; i >= 0; --i)
+  for (long i = long{ Dimension } - 1; i >= 0; --i)
   {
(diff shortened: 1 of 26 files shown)


dzenanz (approved):

thewtex (approved):
:100:

PR #5325, opened 2025-04-24
Allow initializing MersenneTwisterRandomVariateGenerator without mutex locking 

N-Dekker:
Allowed `New()`, `GetInstance()`, and the default-constructor of `MersenneTwisterRandomVariateGenerator` to initialize a
generator object without locking its `m_InstanceMutex`.

Implemented by adding two private member functions: `InitializeWithoutMutexLocking(IntegerType)` and `CreateRandomSeed()`.

### DIFF
--- a/Modules/Core/Common/include/itkMersenneTwisterRandomVariateGenerator.h
+++ b/Modules/Core/Common/include/itkMersenneTwisterRandomVariateGenerator.h
@@ -330,6 +330,10 @@ private:
     return hash(time(nullptr), clock());
   }
 
+  /** Internal method to initialize a generator object without mutex locking. */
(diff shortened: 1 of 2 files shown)


N-Dekker:
> Why do we want avoidance of mutex locking in the constructor? Constructor is not performance critical function.

Well, I think mutex locking should only be done when necessary.  Why would you want to lock the mutex of an object that is still under construction? 🤷 I find it clearer to only use locking when necessary.

dzenanz (approved):

dzenanz (approved):

thewtex (approved):
:+1:

PR #5331, opened 2025-04-28
MersenneTwisterRandomVariateGenerator: Remove Initialize and SetSeed overloads, remove commented out member functions

N-Dekker:
The original overload sets of `Initialize` and `SetSeed` made things unnecessarily compilated: `Initialize()` was implemented by `SetSeed()`. `SetSeed()` was then implemented by calling the other `SetSeed` overload. Which would call the other `Initialize` overload. Schematically: `Initialize()` &rarr; `SetSeed()` &rarr; `SetSeed(seed)` &rarr; `Initialize(seed)`.

With this pull request, there will be just one `SetSeed(seed)`, and one `Initialize(seed)`. `SetSeed` will just call `Initialize`.

Also removed the commented out member functions `load` and `save`.

Aim: simplifying the code.

### DIFF
--- a/Modules/Core/Common/include/itkMersenneTwisterRandomVariateGenerator.h
+++ b/Modules/Core/Common/include/itkMersenneTwisterRandomVariateGenerator.h
@@ -166,16 +166,9 @@ public:
   /** Length of state vector */
   static constexpr IntegerType StateVectorLength = 624;
 
-  /** Initialize with a simple IntegerType */
+  /** Sets the seed and initializes the internal state of this generator. */
   void
-  Initialize(const IntegerType seed);
-
-  /* Initialize with clock time */
-  void
-  Initialize()
-  {
-    SetSeed();
-  }
+  Initialize(const IntegerType seed = Self::CreateRandomSeed());
 
   /** Get a random variate in the range [0, 1] */
   double
@@ -260,24 +253,17 @@ public:
     return GetVariate();
   }
 
-  /** Re-seeding functions with same behavior as initializers
+  /** Re-seeding function with same behavior as Initialize.
    *
    * \note This method is thread-safe.
    */
-  /** @ITKStartGrouping */
   void
-  SetSeed(const IntegerType oneSeed)
+  SetSeed(const IntegerType seed = Self::CreateRandomSeed())
   {
-    // Seed the generator with a simple IntegerType
-    Initialize(oneSeed);
+    // Seed the generator.
+    Initialize(seed);
   }
 
-  void
-  SetSeed()
-  {
-    SetSeed(Self::CreateRandomSeed());
-  }
-  /** @ITKEndGrouping */
   /** Return the current seed
    *
    * \note This method is thread-safe.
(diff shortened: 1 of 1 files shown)


thewtex (approved):
Thanks @N-Dekker :pray:

Modules/Core/Common/include/itkMersenneTwisterRandomVariateGenerator.h
```diff
@@ -296,12 +282,6 @@ class ITKCommon_EXPORT MersenneTwisterRandomVariateGenerator : public RandomVari
   static IntegerType
   GetNextSeed();
 
-  /*
-  // Saving and loading generator state
-  void save( IntegerType* saveArray ) const;  // to array of size SAVE
-  void load( IntegerType *const loadArray );  // from such array
-  */
```
N-Dekker, line 303: For the record, the original `MTRand` implementation _did_ implement `save` and `load`, at https://www.cs.rpi.edu/academics/courses/fall12/ds/hw/05_multi_linked_lists/MersenneTwister.h

```cpp
inline void MTRand::save( uint32* saveArray ) const
{
	register const uint32 *s = state;
	register uint32 *sa = saveArray;
	register int i = N;
	for( ; i--; *sa++ = *s++ ) {}
	*sa = left;
}

inline void MTRand::load( uint32 *const loadArray )
{
	register uint32 *s = state;
	register uint32 *la = loadArray;
	register int i = N;
	for( ; i--; *s++ = *la++ ) {}
	left = *la;
	pNext = &state[N-left];
}

```

However, they were never implemented in ITK's `MersenneTwisterRandomVariateGenerator`, so I guess they were not really necessary so far. 🤷

thewtex: I wonder if we should hove to the C++ STL Mersenne Twister? https://cplusplus.com/reference/random/mt19937/

dzenanz: Oh, it is available since C++11. I wouldn't mind that being done, but I cannot devote resources towards that refactoring. @N-Dekker what about you?

N-Dekker: I think it would indeed be nice to move to the C++ STL Mersenne Twister. But initially I think we should then still preserve the interface of MersenneTwisterRandomVariateGenerator, and just have it implemented internally in terms of `std::mt19937`. The logic of having a global instance and a globally shared `m_StaticDiffer` is not in `std::mt19937`. But `std::mt19937` can certainly do the bit twisting magic for us  😃 As in our `MersenneTwisterRandomVariateGenerator::reload()`: https://github.com/InsightSoftwareConsortium/ITK/blob/7370ffd1f07c9f455f1dedad8d0dbc41c71bb7d5/Modules/Core/Common/src/itkMersenneTwisterRandomVariateGenerator.cxx#L212-L220

dzenanz: That is a reasonable step towards using it directly.

PR #5335, opened 2025-04-29
Remove `MersenneTwisterRandomVariateGenerator::` prefixes within its implementation

N-Dekker:
A minor style improvement, aiming to improve code readability.

### DIFF
--- a/Modules/Core/Common/src/itkMersenneTwisterRandomVariateGenerator.cxx
+++ b/Modules/Core/Common/src/itkMersenneTwisterRandomVariateGenerator.cxx
@@ -81,7 +81,7 @@ MersenneTwisterRandomVariateGenerator::Pointer
 MersenneTwisterRandomVariateGenerator::CreateInstance()
 {
   // Try the factory first
-  MersenneTwisterRandomVariateGenerator::Pointer obj = ObjectFactory<Self>::Create();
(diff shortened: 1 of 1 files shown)


dzenanz (approved):

Modules/Core/Common/src/itkMersenneTwisterRandomVariateGenerator.cxx
```diff
@@ -96,9 +96,9 @@ MersenneTwisterRandomVariateGenerator::CreateInstance()
 MersenneTwisterRandomVariateGenerator::Pointer
 MersenneTwisterRandomVariateGenerator::New()
 {
-  MersenneTwisterRandomVariateGenerator::Pointer obj = MersenneTwisterRandomVariateGenerator::CreateInstance();
+  MersenneTwisterRandomVariateGenerator::Pointer obj = Self::CreateInstance();
 
-  obj->InitializeWithoutMutexLocking(MersenneTwisterRandomVariateGenerator::GetNextSeed());
+  obj->InitializeWithoutMutexLocking(Self::GetNextSeed());
```
N-Dekker: Technically, it isn't even necessary to add `Self::` to internal calls to its own static member functions. But this is the way we like it, right?  😇

PR #5336, opened 2025-04-30
Add `DefaultSeed = 121212` to MersenneTwisterRandomVariateGenerator

N-Dekker:
Added `MersenneTwisterRandomVariateGenerator::DefaultSeed`

It appears useful to have a named constant for this value, for ITK as well as for [elastix](https://github.com/SuperElastix/elastix).

Following C++ Core Guidelines, [Avoid “magic constants”; use symbolic constants](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es45-avoid-magic-constants-use-symbolic-constants)

### DIFF
--- a/Modules/Core/Common/test/itkMersenneTwisterRandomVariateGeneratorGTest.cxx
+++ b/Modules/Core/Common/test/itkMersenneTwisterRandomVariateGeneratorGTest.cxx
@@ -25,6 +25,29 @@
 using itk::Statistics::MersenneTwisterRandomVariateGenerator;
 
 
+// Check that the default seed is not just zero!
+static_assert(MersenneTwisterRandomVariateGenerator::DefaultSeed != 0);
+
+
+// Tests that DefaultSeed is the seed of a default-constructed generator.
+TEST(MersenneTwisterRandomVariateGenerator, DefaultSeed)
+{
+  // Derived generator class whose `New()` simply calls the default-constructor.
+  class DerivedGenerator : public MersenneTwisterRandomVariateGenerator
+  {
+  public:
+    ITK_DISALLOW_COPY_AND_MOVE(DerivedGenerator);
+    itkSimpleNewMacro(DerivedGenerator);
+
+  protected:
+    DerivedGenerator() = default;
+    ~DerivedGenerator() override = default;
+  };
+
+  EXPECT_EQ(DerivedGenerator::New()->GetSeed(), MersenneTwisterRandomVariateGenerator::DefaultSeed);
+}
+
+
 // Tests that GetIntegerVariate() conforms with the C++11 requirement for std::mt19937,
 // when the ITK generator uses the default seed of std::mt19937:
 // "The 10000th consecutive invocation of a default-constructed object of type mt19937
--- a/.github/workflows/itk_dict.txt
+++ b/.github/workflows/itk_dict.txt
@@ -618,6 +618,7 @@ memcpy
 memmove
 memset
 merchantability
+mersenne
 meshpixperelt
 metacharacters
 metricb
--- a/Modules/Core/Common/include/itkMersenneTwisterRandomVariateGenerator.h
+++ b/Modules/Core/Common/include/itkMersenneTwisterRandomVariateGenerator.h
@@ -163,6 +163,15 @@ public:
   static void
   ResetNextSeed();
 
+  /** The initial seed of a default-constructed generator.
+   *
+   * \note The global generator retrieved by `GetInstance()`, and local generators created by `New()` may all have a
+   * different (non-default) seeds.
+   * \note The Mersenne Twister random number engine of the Standard C++ Library `std::mersenne_twister_engine`
+   * (`std::mt19937`) has a different default seed, `default_seed == 5489`.
+   */
+  static constexpr IntegerType DefaultSeed = 121212;
+
   /** Length of state vector */
   static constexpr IntegerType StateVectorLength = 624;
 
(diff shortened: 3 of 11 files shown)


Modules/Core/Common/test/itkMersenneTwisterRandomVariateGeneratorGTest.cxx
```diff
@@ -25,6 +25,33 @@
...
+  {
+  public:
+    ITK_DISALLOW_COPY_AND_MOVE(DefaultConstructableGenerator);
+
+    DefaultConstructableGenerator() = default;
+
+    ~DefaultConstructableGenerator() override
+    {
+      // Bypass ITK's reference counting mechanism, just for the test.
+      LightObject::m_ReferenceCount = 0;
```
dzenanz, line 47: Do we even need to fiddle with this?

N-Dekker: > Do we even need to fiddle with this?

I think so, yes. The destructor of `LightObject` wants `m_ReferenceCount` to be zero, otherwise it will produce a warning: https://github.com/InsightSoftwareConsortium/ITK/blob/7370ffd1f07c9f455f1dedad8d0dbc41c71bb7d5/Modules/Core/Common/src/itkLightObject.cxx#L181

blowekamp: The LightObject is not suppose to support stack based allocation. Some type of dynamic allocation of the object should likely be done to be conforming, and not aggressive manipulation of the IVARs of the class.

N-Dekker: OK, thanks @blowekamp but it's usually perfectly fine to put a random number generator on the stack. Why not?

N-Dekker: FYI, I'm trying to make the behavior of multithreading usage of the elastix library deterministic, as response to #5287 by @michalmeszaroswork: https://github.com/SuperElastix/elastix/tree/RandomVariateGenerator-deterministic-multithreading Instead of `New()`, I feel it's better to directly use the default-constructor of MersenneTwisterRandomVariateGenerator in this case, because the default-constructor does not modify the global data of `MersenneTwisterGlobals`. The `New()` member function of MersenneTwisterRandomVariateGenerator is a bit irregular. It isn't implemented in the common ITK way, by `itkNewMacro(x)`. Instead, it's here: https://github.com/InsightSoftwareConsortium/ITK/blob/7370ffd1f07c9f455f1dedad8d0dbc41c71bb7d5/Modules/Core/Common/src/itkMersenneTwisterRandomVariateGenerator.cxx#L96-L103 

On the other hand, the default-constructor just uses a default seed:
https://github.com/InsightSoftwareConsortium/ITK/blob/7370ffd1f07c9f455f1dedad8d0dbc41c71bb7d5/Modules/Core/Common/src/itkMersenneTwisterRandomVariateGenerator.cxx#L129-L132

N-Dekker: How would you write a unit test that checks that the default-constructor of MersenneTwisterRandomVariateGenerator sets the seed to `DefaultSeed`?

dzenanz: Have you tried to construct an instance via `MersenneTwisterRandomVariateGenerator::New()`, then use `GetSeed()` or `GetNextSeed()`?

blowekamp: Can you add a `itkSimpleNewMacro(x)`, use the default destructor and then do `DefaultConstructableGenerator::New().GetSeed()`?

dzenanz (approved):

Modules/Core/Common/test/itkMersenneTwisterRandomVariateGeneratorGTest.cxx
N-Dekker: > Can you add a `itkSimpleNewMacro(x)`, use the default destructor and then do `DefaultConstructableGenerator::New()->GetSeed()`?

@blowekamp Thanks! Just [force-pushed](https://github.com/InsightSoftwareConsortium/ITK/compare/86a08bdc3e90e89c6f8276ede57361eb645b5949..bc0f677ffc3dac6f940332d94f67109e1c8079dd), please check! (I also renamed the helper class `DerivedGenerator`, because now, it is no longer "default-constructable".) 

@dzenanz  I tried, but `MersenneTwisterRandomVariateGenerator
::New()->GetSeed()` usually does _not_ return `121212` ! It returns a different number every time you call it!

blowekamp: Looks good. Glad my suggestion worked out.

dzenanz:
This should be uppercase, as Mersenne is a person's name: `Run comment_spell_check \
file: Modules/Core/Common/include/itkMersenneTwisterRandomVariateGenerator.h  line: 166  word: mersenne`

dzenanz:
> dictionary have both lowercase "mersenne" and uppercase "Mersenne"

Maybe. Maybe just lowercase should suffice?

PR #5340, opened 2025-05-02
DOC: Add table to describe MetaIO/MetaImage ElementType tag

N-Dekker:
Explicitly specified the number of bits of integer and floating point types.

Aims to clarify possible confusion about `MET_LONG` and `MET_ULONG`, as the size of the C++ type `long` is platform specific.

### DIFF
--- a/Documentation/docs/learn/metaio.md
+++ b/Documentation/docs/learn/metaio.md
@@ -1086,8 +1086,44 @@ In addition to the above tags, `MetaImage` provides the following tags:
     -   Description: Physical size of each voxel.
 -   `ElementType`
     -   Type: `MET_STRING`
-    -   Description: One of enum type: `MET_UCHAR`, `MET_CHAR`… See
-        `metaTypes.h`.
+    -   Description: One of enum type: `MET_UCHAR`, `MET_CHAR`… as described in the following table:
+
+        | ElementType tag      | Description |
+        | -------------------- | ----------- |
+        | MET_NONE             | none |
+        | MET_ASCII_CHAR       | ASCII character |
+        | MET_CHAR             | 8-bit signed integer |
+        | MET_UCHAR            | 8-bit unsigned integer |
+        | MET_SHORT            | 16-bit signed integer |
+        | MET_USHORT           | 16-bit unsigned integer |
+        | MET_INT              | 32-bit signed integer |
+        | MET_UINT             | 32-bit unsigned integer |
+        | MET_LONG             | 32-bit signed integer |
+        | MET_ULONG            | 32-bit unsigned integer |
+        | MET_LONG_LONG        | 64-bit signed integer |
+        | MET_ULONG_LONG       | 64-bit unsigned integer |
+        | MET_FLOAT            | 32-bit single precision floating point |
+        | MET_DOUBLE           | 64-bit double precision floating point |
+        | MET_STRING           | string of 8-bit characters |
+        | MET_CHAR_ARRAY       | array of 8-bit signed integers |
+        | MET_UCHAR_ARRAY      | array of 8-bit unsigned integers |
+        | MET_SHORT_ARRAY      | array of 16-bit signed integers |
+        | MET_USHORT_ARRAY     | array of 16-bit unsigned integers |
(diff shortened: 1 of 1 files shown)


Documentation/docs/learn/metaio.md
```diff
@@ -1086,8 +1086,40 @@ In addition to the above tags, `MetaImage` provides the following tags:
...
+        | MET_NONE             | none |
+        | MET_ASCII_CHAR       | ASCII character |
+        | MET_CHAR             | 8-bit signed integer |
+        | MET_UCHAR            | 8-bit unsigned integer |
+        | MET_SHORT            | 16-bit signed integer |
+        | MET_USHORT           | 16-bit unsigned integer |
+        | MET_INT              | 32-bit signed integer |
+        | MET_UINT             | 32-bit unsigned integer |
+        | MET_LONG             | 32-bit signed integer |
+        | MET_ULONG            | 32-bit unsigned integer |
```
N-Dekker, line 1102: 👉 Please note: `MET_LONG` and `MET_ULONG` are both 32-bit! 👈

dzenanz (approved):

Documentation/docs/learn/metaio.md
thewtex: Always? Only on Windows? Can this information be added to the description?

N-Dekker: > Always? Only on Windows?

Always! Not only on Windows! I'm glad if you are surprised, because I felt surprised as well 😺 `MET_LONG` and `MET_ULONG` are both _always_ 4 bytes (= 32 bits), on _any_ platform, Windows, Linux, MacOS!

The sizes (number of bytes) of all supported element value types are specified by the constant array `MET_ValueTypeSize[MET_NUM_VALUE_TYPES]`:

https://github.com/InsightSoftwareConsortium/ITK/blob/f07971092f0e17c69d5f50f9fadcfccbc45928db/Modules/ThirdParty/MetaIO/src/MetaIO/src/metaTypes.h#L98-L99

`MET_ValueTypeSize[MET_LONG]` and `MET_ValueTypeSize[MET_ULONG]` both have the value `4` (bytes). It's hard-coded, independent of `sizeof(long)`.

> Can this information be added to the description?

Sure, but there is nothing platform-specific about the table anyway. 🤷 So there is no extra information. Other than: _Please note: this description applies to any platform, always!_ Would you want a note like that added to the description?

N-Dekker: Note that MET_LONG_TYPE and MET_ULONG_TYPE are a bit redundant, because MET_INT_TYPE and MET_UINT_TYPE are also 32-bit: https://github.com/InsightSoftwareConsortium/ITK/blob/f07971092f0e17c69d5f50f9fadcfccbc45928db/Modules/ThirdParty/MetaIO/src/MetaIO/src/metaTypes.h#L46-L49

But I think it has always been that way 🤷

thewtex: Thanks for your attention to this, Neils 🙏 

> Please note: this description applies to any platform, always!

Yes, adding that would help inform the reader 👍

N-Dekker: OK, thanks, please check my [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/f21e5332c43193719e94c38631b1761fdc3c0f70..d34745fd77715dc41978d00d6ae09e6326eb7430)!

dzenanz (approved):

Documentation/docs/learn/metaio.md
thewtex: @N-Dekker beautiful :100:

thewtex (approved):

Documentation/docs/learn/metaio.md
N-Dekker: Thanks! Personally I think both MET_LONG_TYPE and MET_ULONG_TYPE should be dropped, or at least deprecated, because they are redundant, and they just cause confusion. Right? Anyway, that's beyond the scope of this pull request!

PR #5344, opened 2025-05-12
BUG: Allow rescaling int to uint of same size

### LINKED ISSUE #5290: Reading Dicom: Rescaling does not allow pixel format change
### Description

I have a dicom image stored as short, which has a rescaling into the "unsigned short" domain that cannot be loaded with ITK.

```
Bits allocated 16
Bits stored 16
High Bit 15
Pixel Representation 1

Rescale Intercept 32768
Rescale Slope 1
Rescale Type US
```

When `itk::GDCMImageIO::ReadImageInformation`is called, there is an exception raised `"Pixel type larger than output type"`.

### Steps to Reproduce

High level snippet:
```c++
    using ReaderType = itk::ImageFileReader<itk::Image<uint16_t, 2>>;
    typename ReaderType::Pointer reader = ReaderType::New();
    reader->SetFileName(a_path);
    reader->SetImageIO(itk::GDCMImageIO::New());
    reader->Update();
```

Minimal code snippet:
```c++
auto a_imageIO = itk::GDCMImageIO::New();
a_imageIO->SetFileName(dicomPath.c_str()); // this file has the tags listed above
a_imageIO->ReadImageInformation(); // This raises the exception and refuse to load
```

This is also reproducible with the code taken directly from `void
GDCMImageIO::InternalReadImageInformation()` https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/IO/GDCM/src/itkGDCMImageIO.cxx#L538
```c++
       // snippet with code taken from "void GDCMImageIO::InternalReadImageInformation()"
        gdcm::ImageReader reader;
        reader.SetFileName(dicomPath.c_str());
        reader.Read();
        const gdcm::Image& image = reader.GetImage();
        const gdcm::PixelFormat& pixeltype = image.GetPixelFormat(); // This is "INT16" (value 5)

       // note this is slightly changed from GDCMImageIO::InternalReadImageInformation (where it's conditionally done)
        const float rescaleSlope = 1.;  // This part comes from the dicom tags reader, but is simplified here for ease of reproducibility
        const float rescaleIntercept = 32768.;  // This part comes from the dicom tags reader, but is simplified here for ease of reproducibility
        gdcm::Rescaler r;
        r.SetIntercept(rescaleIntercept );
        r.SetSlope(rescaleSlope );
        r.SetPixelFormat(pixeltype);
        auto outputpt = r.ComputeInterceptSlopePixelType(); // This is "UINT16" (value 4)
        // Note: end of the "changed" part, the following condition is in current ITK master branch

        if (pixeltype > outputpt) { // INT16 > UINT16, despite the fact that after rescaling result will fit in UINT16...
            itkAssertInDebugOrThrowInReleaseMacro("Pixel type larger than output type")
        }
```

### Expected behavior

Load image correctly without any exception. Like in other readers such as VTK or MicroDicom.

### Actual behavior

Exception raised `"Pixel type larger than output type"`.
Printed logs:
```
"<redacted>\\InsightToolkit-5.2.1\\Modules\\IO\\GDCM\\src\\itkGDCMImageIO.cxx:524:\nITK ERROR: Pixel type larger than output type"
```

### Reproducibility

All the time given the problematic input image.

### Versions

Currently testing with 5.2.1, but I see the problematic check still done in master branch.

### Environment

Windows 10, MSVC

### Additional Information

I do not see any reason for this check to be done, it means if my stored data is int32 and I have a rescale/slope that project the result into a int8 range itk raises an error? I see nothing wrong with that...
```
   if (pixeltype > outputpt) {
        itkAssertInDebugOrThrowInReleaseMacro("Pixel type larger than output type")
    }
```
Would there be any workaround other than removing the rescaling from the tags, load the image, and then reapply it manually afterward?

PtiLuky:
Dicomfile with signed int data (pixel rep=1) can be decoded as uint with the correct intercept. The assertion blocking this behavior has been changed to a more meaningful check.

Closes #5290

With a dicom image stored as short, which has a rescaling into the "unsigned short".

```
Bits allocated 16
Bits stored 16
High Bit 15
Pixel Representation 1

Rescale Intercept 32768
Rescale Slope 1
Rescale Type US
```

The following code was not working before this commit, and now works correctly:

```cpp
#include <iostream>


#include <itkImageFileReader.h>
#include <itkGDCMImageIO.h>

int
main(int argc, char ** argv)
{
  const auto path = "<myfilepath>";

  using ReaderType = itk::ImageFileReader<itk::Image<uint16_t, 2>>;
  typename ReaderType::Pointer reader = ReaderType::New();
  reader->SetFileName(path);
  reader->SetImageIO(itk::GDCMImageIO::New());
  reader->Update();

  return 0;
}
```

### DIFF
--- a/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
+++ b/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
@@ -537,7 +537,7 @@ GDCMImageIO::InternalReadImageInformation()
     switch (outputpt)
     {
       // Default comparison uses the > operator and the order in ScalarType enum
-      // INT types needs to be adjusted so signed input is allowed to produce unsigned output
+      // INT types need to be adjusted so signed input is allowed to produce unsigned output
       case gdcm::PixelFormat::UINT8:
         ptLarger = pixeltype > gdcm::PixelFormat::INT8;
         break;
--- a/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
+++ b/Modules/IO/GDCM/src/itkGDCMImageIO.cxx
@@ -533,7 +533,32 @@ GDCMImageIO::InternalReadImageInformation()
       outputpt = r.ComputeInterceptSlopePixelType();
     }
 
-    if (pixeltype > outputpt)
+    bool ptLarger = false;
+    switch (outputpt)
+    {
+      // Default comparison uses the > operator and the order in ScalarType enum
+      // INT types needs to be adjusted so signed input is allowed to produce unsigned output
+      case gdcm::PixelFormat::UINT8:
+        ptLarger = pixeltype > gdcm::PixelFormat::INT8;
+        break;
+      case gdcm::PixelFormat::UINT12:
+        ptLarger = pixeltype > gdcm::PixelFormat::INT12;
+        break;
+      case gdcm::PixelFormat::UINT16:
+        ptLarger = pixeltype > gdcm::PixelFormat::INT16;
+        break;
+      case gdcm::PixelFormat::UINT32:
+        ptLarger = pixeltype > gdcm::PixelFormat::INT32;
+        break;
+      case gdcm::PixelFormat::UINT64:
+        ptLarger = pixeltype > gdcm::PixelFormat::INT64;
+        break;
+      default:
+        ptLarger = pixeltype > outputpt;
+        break;
+    }
+
+    if (ptLarger)
     {
       itkAssertInDebugOrThrowInReleaseMacro("Pixel type larger than output type")
     }
--- a/Modules/IO/GDCM/test/Baseline/itkGDCMImageIOTestIntToUint.mha.cid
+++ b/Modules/IO/GDCM/test/Baseline/itkGDCMImageIOTestIntToUint.mha.cid
@@ -1,0 +1,1 @@
+bafkreifqaycilgy52jjjytkjbfrfaioc57bfjnfhfuuynozr3g3xmv6apu
--- a/Modules/IO/GDCM/test/CMakeLists.txt
+++ b/Modules/IO/GDCM/test/CMakeLists.txt
@@ -252,6 +252,19 @@ itk_add_test(
 
 itk_add_test(
   NAME
+  itkGDCMImageIOIntToUintTest
+  COMMAND
+  ITKIOGDCMTestDriver
+  --compare
+  DATA{Baseline/itkGDCMImageIOTestIntToUint.mha}
+  ${ITK_TEST_OUTPUT_DIR}/itkGDCMImageIOTestIntToUint.mha
+  itkGDCMImageReadWriteTest
+  DATA{Input/030658_001.dcm}
+  ${ITK_TEST_OUTPUT_DIR}/itkGDCMImageIOTestIntToUint.mha
+  scalar)
+
+itk_add_test(
+  NAME
   itkGDCMImageIO32bitsStoredTest
   COMMAND
   ITKIOGDCMTestDriver
--- a/Modules/IO/GDCM/test/Input/030658_001.dcm.cid
+++ b/Modules/IO/GDCM/test/Input/030658_001.dcm.cid
@@ -1,0 +1,1 @@
+bafkreieyej3y53xaqg6dcedbv5pa6y5m25rdrg42wdqm43oh4ijjrlpl4a


dzenanz:
Is there any chance you can use test file from @issakomi (provided [here](https://github.com/InsightSoftwareConsortium/ITK/issues/5290#issuecomment-2763043645)) as a unit test?

PtiLuky:
> Is there any chance you can use test file from @issakomi (provided [here](https://github.com/InsightSoftwareConsortium/ITK/issues/5290#issuecomment-2763043645)) as a unit test?

Yes sure. May I get some help about updating the file? I see the binary data are indirectly referenced and I cannot find how to upload this file.  

I found the following links https://discourse.itk.org/t/new-script-for-uploading-binary-testing-data/1484 which gives me an alias "data-upload" that doesn't seem to exist for me, and https://docs.itk.org/en/latest/contributing/upload_binary_data.html that wasn't clear about what to do.  
If possible I would really prefer being able to upload the file without having to create many additional accounts (and if possible if there is a way to do so without installing Node.js locally would be preferred...).

dzenanz:
From the docs (your second link) going with "Option A) Upload Via the Web Interface" is the easier path. I have done that, here is the `.cid` file:
[030658_001.zip](https://github.com/user-attachments/files/20210781/030658_001.zip)
that you are supposed to add to git tracking.

thewtex:
Hi @PtiLuky ,

We just upgraded the upload tool (it was offline yesterday):

  https://content-link-upload.itk.org

You should be able to use it easily without creating any new accounts. This is what @dzenanz 
 used, but please let us know if you have any issues using it yourself.

PtiLuky:
Thanks for your help @dzenanz and @thewtex , I could then add the file and a unit test for the current fix.

Locally I could pass the test but I had to manually copy the .dcm file to the `<...>/build/<...>/Input` directory since this was not done by itself, not sure if I did something wrong when adding the .cid...

I am not fully confident yet, so I first want to see if the test can pass there and then I will rebase.

@thewtex thanks for the information, when I tried the first link of the doc (the link you also provided), I had an error so I did not retry later!

Modules/IO/GDCM/test/itkGDCMImageIOIntToUintTest.cxx
```diff
@@ -0,0 +1,77 @@
...
+ *
+ *=========================================================================*/
+
+#include "itkGDCMImageIO.h"
+#include "itkImageFileReader.h"
+#include "itkTestingMacros.h"
+
+// Specific ImageIO test
+
+/** This test verifies itkGDCMImageIO can read
```
dzenanz, line 25: Could `itkGDCMImageReadWriteTest.cxx` have been used here instead of writing a new test? That could also be used to regress the read image against a baseline in MetaImage or NRRD format.

Modules/IO/GDCM/test/itkGDCMImageIOIntToUintTest.cxx
```diff
@@ -0,0 +1,77 @@
...
+  {
+    std::cerr << "Cannot read file " << argv[1] << std::endl;
+    return EXIT_FAILURE;
+  }
+
+  auto reader = ReaderType::New();
+  reader->SetFileName(argv[1]);
+  reader->SetImageIO(dcmImageIO);
+
+  try
```
dzenanz, line 56: This try-catch block could (and should) be replaced by `ITK_TRY_EXPECT_NO_EXCEPTION(reader->Update());`

Modules/IO/GDCM/test/itkGDCMImageIOIntToUintTest.cxx
```diff
@@ -0,0 +1,77 @@
...
+
+  InputImageType::SizeType extentSize = reader->GetOutput()->GetLargestPossibleRegion().GetSize();
+  std::cout << "Read image dimensions: (" << extentSize[0] << ", " << extentSize[1] << ", " << extentSize[2] << ')'
+            << std::endl;
+  if (extentSize[0] == 0 || extentSize[1] == 0 || extentSize[2] == 0)
+  {
+    std::cerr << "File read but empty " << std::endl;
+    return EXIT_FAILURE;
+  }
+
```
dzenanz, line 75: Consider adding `itk::WriteImage(reader->GetOutput(), argv[2], false);` here, to enable regression testing. Before being added as test baseline, that image should be compressed, either by temporarily using `true` flag to writing here, or by using Slicer to open, and then save as compressed.

dzenanz (commented):
This mostly looks good. I think it is preferable to split into two separate commits: fix itself, and test addition. The test addition should have prefix `ENH: ` to comply with style.

dzenanz:
You might need to re-run CMake configure and generate steps to have the test data appear where it should.

dzenanz (approved):

Modules/IO/GDCM/test/itkGDCMImageIOIntToUintTest.cxx
PtiLuky: Yes the demo image is a very simple image with some flat colors, I'm not surprised compression is efficient there!
image

hjmjohnson (approved):

PR #5346, opened 2025-05-13
FUTURE REMOVE `MersenneTwisterRandomVariateGenerator::Initialize`

N-Dekker:
The MersenneTwisterRandomVariateGenerator member function `Initialize` behaves just like `SetSeed`, but `SetSeed` is more commonly used, and clearer.


N-Dekker:
Thanks for your approval @dzenanz 👍 

By the way, I find the name `MersenneTwisterRandomVariateGenerator::Initialize` especially confusing, because it seems to suggest that a newly created `MersenneTwisterRandomVariateGenerator` is still uninitialized! But actually, an instance created by  `MersenneTwisterRandomVariateGenerator::New()` can be used directly, without calling `Initialize`!

PR #5337, opened 2025-05-01
STYLE: Replace `static_cast<T>(X)` with `T{ X }` for any "all caps" `X`

N-Dekker:
Replaced `static_cast<(\w+)>\(([A-Z_][A-Z_]+)\)` with `$1{ $2 }`. According to common C++ naming conventions, these all "all caps" identifiers refer to constants, specified by preprocessor macro's, so these conversions are evaluated at compile-time.

- Follow-up to pull request https://github.com/InsightSoftwareConsortium/ITK/pull/5324 commit 31859d3eb25b95f52eef5ff1ca16c01058579edb

PR #5348, opened 2025-05-14
STYLE: Replace `rand()` calls with `std::mt19937` in ImageAdaptor test

N-Dekker:
Modernized the code. Also improved the style, by removing `static_cast<float>`s.

### DIFF
--- a/Modules/Core/ImageAdaptors/test/itkComplexConjugateImageAdaptorTest.cxx
+++ b/Modules/Core/ImageAdaptors/test/itkComplexConjugateImageAdaptorTest.cxx
@@ -18,6 +18,7 @@
 
 #include "itkComplexConjugateImageAdaptor.h"
 #include "itkImageRegionIterator.h"
+#include <random>
 
 
 using PixelType = std::complex<float>;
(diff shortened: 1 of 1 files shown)


dzenanz (approved):

Modules/Core/ImageAdaptors/test/itkComplexConjugateImageAdaptorTest.cxx
```diff
@@ -34,13 +35,13 @@ itkComplexConjugateImageAdaptorTest(int, char *[])
...
-  srand(50L);
+  std::mt19937                          randomNumberEngine;
+  std::uniform_real_distribution<float> randomNumberDistribution;
 
   itk::ImageRegionIterator<ImageType> iter(image, region);
   for (iter.GoToBegin(); !iter.IsAtEnd(); ++iter)
   {
-    auto            randMax = static_cast<float>(RAND_MAX);
-    const PixelType pixel(static_cast<float>(rand()) / randMax, static_cast<float>(rand()) / randMax);
+    const PixelType pixel(randomNumberDistribution(randomNumberEngine), randomNumberDistribution(randomNumberEngine));
```
N-Dekker, line 44: For the record, this pull request was inspired by a discussion we had before about the use of `static_cast<float>(RAND_MAX)`, at https://github.com/InsightSoftwareConsortium/ITK/pull/5337#issuecomment-2847162171 It appeared that `RAND_MAX` is platform dependent, so this `static_cast<float>` may or may not be lossless. Which made me want to get rid of `RAND_MAX` 👹 !

PR #5349, opened 2025-05-14
Replace `auto var = T{...}` with `T var{...}`

N-Dekker:
Two little style commits:
 - Replace `auto var = T{}` with `T var{}`
 - Replace `auto var = T{ x }` with `T var{ x }`

Note that these proposed changes are not at all in disagreement  with clang-tidy [modernize-use-auto](https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-auto.html). 😃


dzenanz (approved):

thewtex (approved):
:100:

PR #5350, opened 2025-05-14
BUG: Fix random number in `RGBGibbsPriorFilter::GibbsTotalEnergy`

N-Dekker:
The original division of `rand()` by `32768.0` was probably meant to yield a number between `[0, 1>`. However, `rand()` returns an integer between `[​0​, RAND_MAX]`, and `RAND_MAX` may not be `32767`. In practice, when using GCC, `RAND_MAX` may be as large as `2147483647`.

The "magic number" `32768.0` was introduced on 12 Jan 2002 already, with commit 074c74f036a7faea5555a6125a2c740d1afe7caa.


N-Dekker:
@chentingpc Hi Ting Chen! Could it be that this is originally your code? If so, maybe you can have a look at this pull request! 😃

dzenanz (approved):
Looks good on a glance.

thewtex (approved):
@N-Dekker , wow, nice find!

PR #5353, opened 2025-05-16
STYLE: Replace `std::rand()` with `std::mt19937` in IndexRangeGTest

N-Dekker:
It is recommended to use C++11's random number generation facilities (like `std::mt19937`) to replace `rand()`, according to
https://en.cppreference.com/w/cpp/numeric/random/rand


dzenanz (approved):
Looks good on a glance. Rebasing on `master` should fix the CI failures.

dzenanz (approved):

hjmjohnson (approved):

PR #5355, opened 2025-05-19
Fix reusing TIFFImageIO with series reader

blowekamp:
Address a bug where incorrect file can be read when reusing the TIFFImageIO.


Modules/IO/ImageBase/test/CMakeLists.txt
```diff
@@ -933,7 +933,8 @@ itk_add_test(
   COMMAND
   itkUnicodeIOTest)
 
-set(ITKIOImageBaseGTests itkWriteImageFunctionGTest.cxx)
```
dzenanz, line 936: Perhaps change commit message to say "ENH: Add a test which exposes a bug in ImageIO with ReverseOrder"

blowekamp: When I wrote the test, I didn't realize it was only a TIFFIO bug. I'm going to move the test over to the TIFF directory too.

Modules/IO/TIFF/test/itkImageSeriesReaderReverse.cxx
```diff
@@ -0,0 +1,124 @@
...
+  // Test reading with ReverseOrder=True
+  auto reader = itk::ImageSeriesReader<ImageType>::New();
+  reader->SetImageIO(imageIO);
+  reader->SetFileNames(filePaths);
+  reader->SetReverseOrder(true);
+  reader->Update();
+  ImageType::Pointer reversed_img = reader->GetOutput();
+
+
+  // Print the value at 0,0 of each slice for the two images
```
dzenanz, line 111:
```suggestion
  // Print the value at 0,0 of each slice
```

dzenanz: Comment on line 73 says it is 3 images.

dzenanz (approved):
Mostly looks good.

blowekamp:
@thewtex This patch is about ready to go. The bug fix 5bccae4a2336faedeba6b6033471f1b22d6e73db, should be ported to the release-5.4 branch. As I recall there were some CI issues if the PR was based on that branch. So it will need to be cherry picked.

thewtex:
@blowekamp thanks, I added to release-5.4. I anticipate a 5.4.4 release in the coming weeks

Issue #4794, opened 2024-08-02
Behaviour change between ITK 5.3 and 5.4 with DICOM spacing


seanm:
Further to some discussions in https://github.com/InsightSoftwareConsortium/ITK/pull/4647 and https://github.com/InsightSoftwareConsortium/ITK/pull/4521, we still have a situation that isn't quite clear to us, and since those are merged already, we thought we'd create a new ticket.

Basically, there is a change in behaviour between ITK 5.3 and 5.4 that we are trying to understand if it's a bug or deliberate & correct.

If we query a DICOM file for its image spacing information, we get different results depending on:
- using `itk::ImageSeriesReader` vs using `itk::GDCMImageIO`
- having 1 DICOM file in the folder vs having multiple

A small C++ test case is attached: [inconsistency_in_spacing.zip](https://github.com/user-attachments/files/16473741/inconsistency_in_spacing.zip)

It uses a DICOM file named `D_CLUNIE_CT1_J2KR1.dcm`, from the GDCM test suite I think, but from David Clunie's collection ultimately I suppose.

In ITK 5.3:

Number of files | ImageSeriesReader | GDCMImageIO
---|---|---
Single | [0.661468, 0.661468, 1] | [0.661468, 0.661468, 1]
Multiple | [0.661468, 0.661468, 1] | [0.661468, 0.661468, 1]

Notice everything agrees: the z spacing is 1.

In ITK 5.4 / master:

Number of files | ImageSeriesReader | GDCMImageIO
---|---|---
Single | [0.661468, 0.661468, 5] | [0.661468, 0.661468, 5]
Multiple | [0.661468, 0.661468, 1] | [0.661468, 0.661468, 5]

Notice now everything does not agree.

The change from 1 to 5 is presumably deliberate from https://github.com/InsightSoftwareConsortium/ITK/pull/4521.

But then shouldn't that lone 1 (in the bottom left) also be 5?

Thanks.

issakomi:
[...]

PR #5357, opened 2025-05-20
BUG: Fix case where DICOM series order has negative spacing

blowekamp:
If the spacing between slices is negative, then direction cosine matrix was not correct with reguards to the sign.

## PR Checklist

### DIFF
--- a/Modules/IO/ImageBase/include/itkImageSeriesReader.hxx
+++ b/Modules/IO/ImageBase/include/itkImageSeriesReader.hxx
@@ -172,8 +172,8 @@ ImageSeriesReader<TOutputImage>::GenerateOutputInformation()
     // Override the position if there is an ITK_ImageOrigin
     ExposeMetaData<Array<SpacingScalarType>>(lastReader->GetImageIO()->GetMetaDataDictionary(), key, positionN);
 
-    // Compute and set the inter slice spacing
-    // and last (usually third) axis of direction
+    // Compute and set the inter-slice spacing
+    // and last (usually third) axis of a direction
     Vector<SpacingScalarType, TOutputImage::ImageDimension> dirN;
     for (unsigned int j = 0; j < TOutputImage::ImageDimension; ++j)
     {
@@ -187,9 +187,33 @@ ImageSeriesReader<TOutputImage>::GenerateOutputInformation()
     }
     else
     {
+      // always positive spacing, corrected in the direction
       spacing[this->m_NumberOfDimensionsInImage] = dirNnorm / (numberOfFiles - 1);
       this->m_SpacingDefined = true;
-      if (!m_ForceOrthogonalDirection)
+      if (m_ForceOrthogonalDirection)
+      {
+        // DICOM files only specify two direction vectors, and the third is derived as orthogonal to those. Hence only
+        // when loading DICOM files, the third direction is forced to be orthogonal from the DICOM ImageIO. However,
+        // with the forced positive spacing the sign of the direction must be checked.
+
+
+        // Compute dot product between dirN and the current direction column vector
+        SpacingScalarType dotProduct = 0.0;
+        for (unsigned int j = 0; j < TOutputImage::ImageDimension; ++j)
+        {
+          dotProduct += dirN[j] * direction[j][this->m_NumberOfDimensionsInImage];
+        }
+
+        // If dot product is negative, flip the sign of the entire direction column vector
+        if (dotProduct < 0.0)
+        {
+          for (unsigned int j = 0; j < TOutputImage::ImageDimension; ++j)
+          {
+            direction[j][this->m_NumberOfDimensionsInImage] *= -1.0;
+          }
+        }
+      }
+      else
       {
         for (unsigned int j = 0; j < TOutputImage::ImageDimension; ++j)
         {
--- a/Documentation/docs/releases/5.4.6.md
+++ b/Documentation/docs/releases/5.4.6.md
@@ -226,4 +226,3 @@ We remain dedicated to supporting current users through:
 #### Enhancements
 
 - MINC 2025-02-24 (3b8d9c7e) ([a32b73adbc](https://github.com/InsightSoftwareConsortium/ITK/commit/a32b73adbc))
-
--- a/Modules/IO/GDCM/test/CMakeLists.txt
+++ b/Modules/IO/GDCM/test/CMakeLists.txt
@@ -448,3 +448,28 @@ addcompliancetest(raw-YBR_FULL_422)
 addcompliancetest(RLE-RGB)
 addcompliancetest(HTJ2K-YBR_ICT Lily_full.mha)
 addcompliancetest(HTJ2K-YBR_RCT Lily_full.mha)
+
+set(ITKGDCMImageIOGTests itkGDCMImageIOGTest.cxx)
+creategoogletestdriver(ITKGDCMImageIO "${ITKIOGDCM-Test_LIBRARIES}" "${ITKGDCMImageIOGTests}")
+
+ExternalData_Expand_Arguments(
+  ITKData
+  _DICOM_SERIES_INPUT
+  "DATA{${ITK_DATA_ROOT}/Input/DicomSeries/,REGEX:Image[0-9]+.dcm}"
+)
+target_compile_definitions(
+  ITKGDCMImageIOGTestDriver
+  PRIVATE
+    "DICOM_SERIES_INPUT=${_DICOM_SERIES_INPUT}"
+  PRIVATE
+    "ITK_TEST_OUTPUT_DIR=${ITK_TEST_OUTPUT_DIR}"
+)
+
+set_property(
+  TARGET
+    ITKGDCMImageIOGTestDriver
+  APPEND
+  PROPERTY
+    DEPENDS
+      ITKData
+)
--- a/Modules/IO/GDCM/test/CMakeLists.txt
+++ b/Modules/IO/GDCM/test/CMakeLists.txt
@@ -449,7 +449,11 @@ addcompliancetest(RLE-RGB)
 addcompliancetest(HTJ2K-YBR_ICT Lily_full.mha)
 addcompliancetest(HTJ2K-YBR_RCT Lily_full.mha)
 
-set(ITKGDCMImageIOGTests itkGDCMImageIOGTest.cxx)
+set(
+  ITKGDCMImageIOGTests
+  itkGDCMImageIOGTest.cxx
+  itkGDCMSeriesDirectionGTest.cxx
+)
 creategoogletestdriver(ITKGDCMImageIO "${ITKIOGDCM-Test_LIBRARIES}" "${ITKGDCMImageIOGTests}")
 
 ExternalData_Expand_Arguments(
--- a/Modules/IO/GDCM/test/itkGDCMImageIOGTest.cxx
+++ b/Modules/IO/GDCM/test/itkGDCMImageIOGTest.cxx
@@ -1,0 +1,339 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         https://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+#include "itkImage.h"
+#include "itkImageFileWriter.h"
+#include "itkGDCMImageIO.h"
+#include "itkGDCMSeriesFileNames.h"
+#include "itkMetaDataObject.h"
+#include "itkGTest.h"
+#include "itksys/SystemTools.hxx"
+#include "itkImageSeriesReader.h"
+#include <string>
+#include <vector>
+
+
+#define _STRING(s) #s
+#define TOSTRING(s) std::string(_STRING(s))
+
+namespace
+{
+
+struct ITKGDCMImageIO : public ::testing::Test
+{
+  void
+  SetUp() override
+  {
+    itksys::SystemTools::MakeDirectory(m_TempDir);
+  }
+
+  void
+  TearDown() override
+  {
+    itksys::SystemTools::RemoveADirectory(m_TempDir);
+  }
+
+  const std::string m_TempDir{ TOSTRING(ITK_TEST_OUTPUT_DIR) + "/ITKGDCMImageIO" };
+  const std::string m_DicomSeriesInput{ TOSTRING(DICOM_SERIES_INPUT) };
+};
+
+class ITKGDCMSeriesTestData : public ITKGDCMImageIO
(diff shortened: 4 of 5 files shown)


blowekamp:
relates to SimpleITK/SimpleITK#2292

dzenanz (approved):
Looks good after a quick look.

blowekamp:
I update the test case to use the sample MRI series in ITK test data, and use the ReverseOrder flag. The test is compare the pixel values at the physical locations match.

dzenanz:
Maybe we could have both of those tests?

blowekamp:
> Maybe we could have both of those tests?

The DICOM tags were not correct in the prior example test case where the DICOM files were manually created. And oddly the X,Y spacing was being swapped for some reason likely related to the TAGs. Also this test case for the file has a non identify cosine matrix, which helps  verifies the correct column was multiplied and not a row by -1.

N-Dekker:
Thanks @blowekamp I'm just trying to understand the issue. It reminds me of an ITKElastix issue reported by @thewtex:
- https://github.com/InsightSoftwareConsortium/ITKElastix/issues/291

Which appeared related to `ImageFileReader::GenerateOutputInformation()`: 
https://github.com/InsightSoftwareConsortium/ITK/blob/b375cefdc7fcd761a5581359e35dfecdc5f9d615/Modules/IO/ImageBase/include/itkImageFileReader.hxx#L222-L235

blowekamp:
The original issue was reported in SimpleITK here:
https://github.com/SimpleITK/SimpleITK/issues/2292

The cause appeared to be that the z-slice direction was negative. This was more easily reproduce by enabling the "ReverseOrder" flag.

It may very well be likely related to this change too:
https://github.com/InsightSoftwareConsortium/ITK/blob/b375cefdc7fcd761a5581359e35dfecdc5f9d615/Modules/IO/ImageBase/include/itkImageFileReader.hxx#L222-L235

It's not clear to me if the original DICOM reported in SimpleITK had negative spacing. The "ReverseOrder" reproduction of the issue, has negative inter-slice distance.  There is certainly areas that need further investigation.

The principle used in the test, that for DICOM if the "reverse order" flag is set then the physical location of the voxel should be maintained, should be correct. But DICOM is complicated.

issakomi:
For the the change related to _SpacingBetweenSlices_ s.
https://github.com/InsightSoftwareConsortium/ITK/issues/4794#issuecomment-2266073027
https://github.com/InsightSoftwareConsortium/ITK/issues/4794#issuecomment-2266616251
This may lead to issues if SpacingBetweenSlices is negative, surprisingly it is not uncommon (invalid from DICOM point of view, except for ancient NM storage).

DICOM doesn't force any order of slices. For classic series it is easy to sort files (ascending along slice normal, if IPP/IOP are available). This is done [in itkGDCMSeriesFileNames.cxx](https://github.com/InsightSoftwareConsortium/ITK/blob/624f745601b44d99c0d97b50eb846e1054f6fbc1/Modules/IO/GDCM/src/itkGDCMSeriesFileNames.cxx#L141).

For enhanced multi-frame instances is it not really done in GDCM, but if the order is descending (very very often) GDCM returns negative spacing and the code in [tkImageFileReader.hxx](https://github.com/InsightSoftwareConsortium/ITK/blob/b375cefdc7fcd761a5581359e35dfecdc5f9d615/Modules/IO/ImageBase/include/itkImageFileReader.hxx#L222-L235) (posted above) is triggered and so far the image is correct, but has left-handed matrix.

P.S. I don't use this stuff in my app, so it is just FYI.

issakomi:
> For enhanced multi-frame instances is it not really done in GDCM, but if the order is descending (very very often) GDCM returns negative spacing and the code in [tkImageFileReader.hxx](https://github.com/InsightSoftwareConsortium/ITK/blob/b375cefdc7fcd761a5581359e35dfecdc5f9d615/Modules/IO/ImageBase/include/itkImageFileReader.hxx#L222-L235) (posted above) is triggered and so far the image is correct, but has left-handed matrix.
> 

@N-Dekker To demonstrate how GDCM returns negative spacing and the code in [itkImageFileReader.hxx](https://github.com/InsightSoftwareConsortium/ITK/blob/b375cefdc7fcd761a5581359e35dfecdc5f9d615/Modules/IO/ImageBase/include/itkImageFileReader.hxx#L222-L235) is triggered:

The test file can be e.g. [this one](https://data.kitware.com/api/v1/file/5c193fcd8d777f072bd9a4e0/download), used for regular testing.

Add e.g. `std::cout << sp[0] << ' ' << sp[1] << ' ' << sp[2] << std::endl;` before `return` in [gdcmImageHelper.cxx GetSpacingValueFromSequence](https://github.com/InsightSoftwareConsortium/ITK/blob/624f745601b44d99c0d97b50eb846e1054f6fbc1/Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmImageHelper.cxx#L317)
`1 1 -1.33`

or it is possible to load the file, save  as MHA and look at the line `ITK_original_spacing = 1 1 -1.33` (it is important to use recent ITK, there was a [bug](https://github.com/InsightSoftwareConsortium/ITK/commit/087a1c4fb0abe5fae279bf0dc285d4fb962424c6) in ITK_original_spacing).

blowekamp:
This patch does not directly deal with the DICOM tags, and specific DICOM information. It only correct a sign change in the DirectionCosine matrix when detected to maintain consistent physical spacing when the order is reversed. The specifics of multi-frame DICOM are not handled here.

If there is a specific DICOM series that should be tested, we can manually do that to ensure consistency.

Otherwise, I think this patch should be good?

Modules/IO/ImageBase/include/itkImageSeriesReader.hxx
```diff
@@ -189,7 +189,14 @@ ImageSeriesReader<TOutputImage>::GenerateOutputInformation()
     {
       spacing[this->m_NumberOfDimensionsInImage] = dirNnorm / (numberOfFiles - 1);
       this->m_SpacingDefined = true;
-      if (!m_ForceOrthogonalDirection)
+      if (m_ForceOrthogonalDirection)
+      {
+        for (unsigned int j = 0; j < TOutputImage::ImageDimension; ++j)
+        {
+          direction[j][this->m_NumberOfDimensionsInImage] *= std::signbit(dirN[j]) ? -1.0 : 1.0;
```
thewtex: In general, I do not think this is necessarily 1.0, -1.0.

blowekamp: The test case added the ImageSeriesReader::ReverseOrderOn option. Due to the way spacing is computed for DICOM it is forced to be positive with this flag, while the direction cosine matrix is the same .

This change enable an DICOM series loaded with ReverseOrderOn to occupy the same physical space as when it it off. This seem expected and correct behavior to me.

There is the open issue, getting a test DICOM series which natively has negative spacing in it.

thewtex: With `ImageSeriesReader::ReverseOrderOn`, this direction value is not necessarily `-1.0 : 1.0`.

blowekamp: Please clarify what you mean, the operation here is "*=" so only a sign correction is done.

EDIT: Updating test case  to use non-unit spacing.

blowekamp:
> Thanks @blowekamp I'm just trying to understand the issue. It reminds me of an ITKElastix issue reported by @thewtex:
> 
> * [GroupwiseRegistration notebook: ERROR: the direction cosines matrix of the fixed image is invalid! ITKElastix#291](https://github.com/InsightSoftwareConsortium/ITKElastix/issues/291)
> 
> Which appeared related to `ImageFileReader::GenerateOutputInformation()`:
> 
> https://github.com/InsightSoftwareConsortium/ITK/blob/b375cefdc7fcd761a5581359e35dfecdc5f9d615/Modules/IO/ImageBase/include/itkImageFileReader.hxx#L222-L235

I just tries this patch with this dataset referenced. I used the file names in the order presented by `GDCMSeriesFileNames`:

    -00-slice009.dcm
    -00-slice008.dcm
    -00-slice007.dcm
    -00-slice006.dcm
    -00-slice005.dcm
    -00-slice004.dcm
    -00-slice003.dcm
    -00-slice002.dcm
    -00-slice001.dcm
    -00-slice000.dcm

When read in this order with the patch the direction(2,2) == -1.0, same as reported in the issue. When ReverseOrder is enable then direction(2,2,) == 1.0 and the physical geometry is the same. For both cases the space is `[0.976562, 0.976562, 2]`. 

This behavior is consistent and follow the expectations of the patch.

Modules/IO/ImageBase/include/itkImageSeriesReader.hxx
blowekamp: The existing test using the DICOM series data in ITK had a non-trivial direction cosine matrix, and was preserved.

Also I added more testing based on the discussion in the SimpleITK issue and explicitly create a DICOM series with negative spacing.

thewtex: If the image does not have an orthogonal orientation with respect to the z-plane, this will not produce a correct result.

thewtex: This may fix the reported issue / test case, but it will break other use cases.

blowekamp: > If the image does not have an orthogonal orientation with respect to the z-plane, this will not produce a correct result.

Only changing the sign of the column vectors does not affect the orthogonality of the column vectors. As I understand this statement, it does not appear to be correct.


In ITK the spacing has to be positive, if the sign of the spacing is changed the the sign of the column vector should be changed to match. < added note: the code doesn't match this description let me check it >

Please be more specific and explanatory with your concerns.

blowekamp: I have update the code, to match that provided description.

Modules/IO/ImageBase/include/itkImageSeriesReader.hxx
```diff
@@ -187,9 +187,33 @@ ImageSeriesReader<TOutputImage>::GenerateOutputInformation()
...
+        {
+          dotProduct += dirN[j] * direction[j][this->m_NumberOfDimensionsInImage];
+        }
+
+        // If dot product is negative, flip the sign of the entire direction column vector
+        if (dotProduct < 0.0)
+        {
+          for (unsigned int j = 0; j < TOutputImage::ImageDimension; ++j)
+          {
+            direction[j][this->m_NumberOfDimensionsInImage] *= -1.0;
```
thewtex, line 212: This assumes that the image orientation is orthogonal to the z-plane. This is not always the case, e.g. some DICOMs. The solution should look more like:

https://github.com/InsightSoftwareConsortium/ITK/blob/31744be819a302cce05d52cb42c44d5de7139abe/Modules/IO/ImageBase/include/itkImageSeriesReader.hxx#L194-L197

thewtex (dismissed):
Please see the inline comments.

Please also add a test where the image is not orthogonal to the z-plane, i.e. the third direction component is neither [0,0,1] or [0,0,-1].

Modules/IO/ImageBase/include/itkImageSeriesReader.hxx
blowekamp: The "direction" matrix helps transform index-space to xyz space. The third column by construction in GDCM is one of the orthogonal directions to the imaging planes:
https://github.com/InsightSoftwareConsortium/ITK/blob/main/Modules/IO/GDCM/src/itkGDCMImageIO.cxx#L734-L743

This is the case where the wrong direction was computed due to the spacing sign, or the order of the images, and its is being corrected. The non-orthogonal imaging is a separate case.

thewtex: I understand that the third direction is orthogonal to the first two. That was not my point. And I was not referring to non-orthogonal imaging.

dzenanz: The axis that needs flipping might not be the third, but rather second or first. Is that your point Matt?

thewtex: It is not necessarily a simple flip of one value.

hjmjohnson:
@thewtex — the new test commit (`fcecee9e`) addresses both of your top-level requests directly. Quick visible summary:

- **"Add a test where the input does not have any 0's and 1's in the direction matrix"** → `GDCMSeriesDirection.Oblique_NonAxisAligned_Reversal` writes synthesized DICOM with row/col orientations derived from a non-trivial composed rotation (no zero/one elements), confirms forward and reversed file orderings give identical physical-point lookups for every voxel.
- **"Check that the resulting direction matrix's third vector equals the cross product of the first two"** → `GDCMSeriesDirection.ThirdColumnEqualsCrossProduct_Oblique` computes `c0 × c1` and asserts it is parallel to `c2` (sign-agnostic — ITK accepts left-handed direction matrices, det = ±1).

<details>
<summary>Full coverage map (17 gtest cases) and regression-detector check</summary>

| Test | Source / contract |
|---|---|
| `SimpleITK2292_BottomToTop_OriginAndDirectionAgree` / `SimpleITK2292_TopToBottom_OriginAndDirectionAgree` | [SimpleITK#2292](https://github.com/SimpleITK/SimpleITK/issues/2292) reproducer |
| `PhysicalPointInvariance_AxialReversal` / `PhysicalPointInvariance_ReverseOrderFlag` | Voxel-level invariance under file-order reversal |
| `Sagittal_AxisAligned_Reversal` / `Coronal_AxisAligned_Reversal` | Non-axial patient frames |
| `Oblique_NonAxisAligned_Reversal` | thewtex request #1 (no 0/1 entries) |
| `ThirdColumnEqualsCrossProduct_Oblique` | thewtex request #2 (cross-product check) |
| `ForceOrthogonalDirectionOff_UsesDirectIPPDifference` | Pin legacy code path |
| `ForceOrthogonalOnOff_AgreeOnAxisAligned` | On/off symmetry |
| `TwoSlices_MinimumValidCase` / `ManySlices_FineSpacing` | Boundary cases |
| `RealDicomSeries_PhysicalPointInvariant` | ITK's bundled `Input/DicomSeries` data |
| `ITKElastix291_TimeSeriesShapedDirection` | [ITKElastix#291](https://github.com/InsightSoftwareConsortium/ITKElastix/issues/291) contract |
| `OriginEqualsFirstListedSliceIPP` | Patch must not move origin |
| `SpacingAlwaysPositive` | Invariant |
| `RandomizedStress_PhysicalPointInvariance` | 8 random oblique trials, deterministic seed |

Reverting just the patched block in `itkImageSeriesReader.hxx` while keeping the suite causes **12 of the 17 tests to fail** (locally verified), so the suite is a faithful regression detector — not just exercises that happen to pass against the patched code.

</details>

<details>
<summary>Note on handedness</summary>

The patch produces left-handed direction matrices (det = −1) for canonical axial top-to-bottom DICOM where the IPP step is in −Z. ITK accepts this and the tests verify orthonormality with `|det| = 1` rather than `det = +1`. Worth flagging in case anyone reviewing thinks det should always be +1 — it's not, and the unpatched code already produces det = −1 in many DICOM cases.

</details>

dzenanz (commented):
I have taken a reasonable look at the first commit. It seems fine. I have not taken a good look at the tests in the second commit. Matt should review too.

hjmjohnson:
@dzenanz Thanks
@thewtex If you approve, would you please also merge?

thewtex (approved):
@hjmjohnson thank you!

PR #5367, opened 2025-05-28
WIP: COMP: Set CMP0177 in wrapping

thewtex:
This policy, added in CMake 3.31, changes the behavior of install(DESTINATION per

  https://cmake.org/cmake/help/latest/policy/CMP0177.html

Also remove the extra "/" in DESTINATION.

I am testing this in ITKPythonPackage.


thewtex:
Superceded by 11ea27f

PR #5379, opened 2025-06-04
Add cmake-format to pre-commit

blowekamp:
Apply automatic formatting for CMake file for style consistency via pre-commit.

Previous work done in 788fbc48, created a shell script: Utilities/Maintenance/cmake-format.bash and a yaml configuration file. Unfortunately, the pre-commit does not appear to understand the yaml ( perhaps some missing dependency?), so the json copy of the configuration was needed.

TODO what to do with previous work? Update to use json? Remove the old scrip?

What about the gitattribute properties style properties? Should they be removed? Ideally pre-commit could be updated to use git attributes, but currently we have a lot of file filtering configuration in the pre-commit hooks.


CMakeLists.txt
```diff
@@ -262,22 +260,15 @@ endif()
 # The default path when not wrapping.  Restore standard build location
 # if python wrapping is turned on, and then turned off.
 if(NOT CMAKE_LIBRARY_OUTPUT_DIRECTORY)
-  set(NO_WRAP_CMAKE_LIBRARY_OUTPUT_DIRECTORY
-      ${ITK_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR}
-      CACHE PATH "Shared library directory")
+  set(NO_WRAP_CMAKE_LIBRARY_OUTPUT_DIRECTORY ${ITK_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR} CACHE PATH
+                                                                                             "Shared library directory")
```
dzenanz: Interesting result

CMakeLists.txt
```diff
@@ -245,9 +245,7 @@ endif()
 # Because GoogleTest sets CMAKE_DEBUG_POSTFIX CMake CACHE variable to "d" when it is
 # built in debug, we preemptively set it to an empty string to avoid having a postfix
 # added to the ITK library names.
-set(CMAKE_DEBUG_POSTFIX
-    ""
-    CACHE STRING "Generate debug library name with a postfix.")
+set(CMAKE_DEBUG_POSTFIX "" CACHE STRING "Generate debug library name with a postfix.")
```
dzenanz: I like new style better.

dzenanz (approved):
LGTM. Is this applied recursively to all CMake files? I see that only top-level file is affected.

thewtex (approved):
Cool!

+1 for retiring the old script and just using the pre-commit hook along with its configuration. It will be easier to apply continuously.

hjmjohnson (changes requested):
See #5417  for updated approach as suggested by @blowekamp.

hjmjohnson:
@blowekamp Thanks for initiating this PR.  Based on suggestions, it was addressed with a  different tool.

PR #5380, opened 2025-06-04
nifti update

hjmjohnson:
ENH: Update to latest nifti ThirdParty codebase

- **COMP: Ignore kwstyle failures in ThirdParty**
- **nifti 2025-06-04 (943aacb4)**

## PR Checklist

### DIFF
--- a/Modules/ThirdParty/NIFTI/src/nifti/niftilib/nifti1_tool.c
+++ b/Modules/ThirdParty/NIFTI/src/nifti/niftilib/nifti1_tool.c
@@ -165,6 +165,7 @@ static int  g_debug = 1;
 #include <limits.h>
 #include <string.h>
 #include <stdint.h>
+#include <stddef.h>
 
 #include "nifti1_io.h"
 #include "nifti1_tool.h"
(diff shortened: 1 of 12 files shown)


dzenanz:
This PR adds 3 more CMake variables: `FSLSTYLE_NAME_CONFLICTS`, `FSLSTYLE_PIGZ_SUPPORT`, and `FSLSTYLE_REJECT_COMPLEX`. Should we keep them exposed in ITK's configuration, or always set them to some concrete values?

Modules/ThirdParty/NIFTI/src/nifti/niftilib/nifti1_tool.c
```diff
@@ -689,15 +690,15 @@ int verify_opts( nt_opts * opts, char * prog )
...
    char * cp;
-   int    len, remain = (int)sizeof(opts->command);  /* max command len */
-   int    c, ac;
+   size_t len, c, remain = sizeof(opts->command);  /* max command len */
+   int    ac;
    int    has_space;  /* arguments containing space must be quoted */
    int    skip = 0;   /* counter to skip some of the arguments     */
 
    /* get the first argument separately */
    len = snprintf( opts->command, sizeof(opts->command),
```
issakomi, line 699: FYI, _snprintf_ can return negative number, but `len` is now unsigned _size_t_

seanm: I've never seen snprintf return a negative, but indeed the macOS man page says vaguely "These functions return a negative value if an error occurs" but does not say what kind of error could cause this... other sources say "If an encoding error occurs, a negative number is returned."  nifticlib has a lot of ASCII-only assumptions already, so in practice I think this is nearly impossible.

But I could do a code review of all snprintf and use signed variables and check the error checking...

seanm: https://github.com/InsightSoftwareConsortium/nifti_clib/pull/19

PR #5385, opened 2025-06-06
BUG: Python version mis-match during build and test

hjmjohnson:
```txt
grep "Found Python" ITK.macOS.Python.logs

-- Found Python3: /Users/runner/hostedtoolcache/Python/3.9.22/x64/bin/python3 (found suitable version "3.9.22", required range is "3.9...3.999") found components: Interpreter Development.Module
-- Found Python: /usr/local/Frameworks/Python.framework/Versions/3.13/bin/python3.13 (found version "3.13.3") found components: Interpreter
Linking CXX shared module Wrapping/Generators/Python/itk/_ITKPyBasePython.cpython-313-darwin.so
```

NOTE:  Python3.13 is found after Python 3.9 is set.  The issue is that python3.9 is used to run the tests, but the shared libraries are built with 3.13.


hjmjohnson:
superceeds #5384

thewtex:
Thanks, @hjmjohnson ! I backported this following #5383

Issue #5358, opened 2025-05-21
Output of itk::DisplacementFieldJacobianDeterminantFilter depends on direction matrix


cookpa:
### Description

Jacobian determinants from itk::DisplacementFieldJacobianDeterminantFilter differ from those derived from itkDisplacementFieldTransform.hxx if the displacement field image has non-identity direction matrix.

### Steps to Reproduce

Here's a link to ITK code that computes Jacobian two ways: by using the itk::DisplacementFieldJacobianDeterminantFilter, and by iterating over each voxel calling `transform->ComputeJacobianWithRespectToPosition` on a displacement field transform object.

https://github.com/cookpa/antsJacobianExample/tree/master/itk_jacobian

The warp fields I tested on were from ANTs, and are available here

https://upenn.box.com/s/voe2agpd30serkj4kn0fvt9bit8pg5nx

By integrating the Jacobian over each label in the fixed space, I computed the expected volume of the deformed moving image. For an identity direction matrix, the two methods are similar.

But for the case where the direction matrix is diag(-1,-1,1), the determinant filter is very different.
 

### Expected behavior

Determinants are similar, if not exactly the same. They should both approximate the volume change within a region due to the warp.

### Actual behavior

Differences of a few percent for oblique direction matrices, large differences for D = diag(-1,-1, 1).

### Reproducibility

Consistent on my test data.

### Versions

v5.4.3

### Environment

Mac OS Intel
Python 3.11
Apple clang version 16.0.0 (clang-1600.0.26.6)

### Additional Information

Code at https://github.com/cookpa/antsJacobianExample

The box link above contains data and code.

hjmjohnson:
Resolved by merged PR #5390 ("BUG: Jacobian determinant filter results depended on index space orientation", merged 2026-05-05). PR title is an exact match for this issue. Closing manually — the PR used 'Fixing' rather than a canonical 'Fixes' keyword so GitHub did not auto-close.

PR #5390, opened 2025-06-09
BUG: Jacobian determinant filter results depended on index space orientation

cookpa:
Fixing #5358 

I updated the ITK test to check the determinants after permuting or flipping the index space, and updating the header to compensate. I think I've done this correctly, at least I can now get the same answer as with an identity transform.

I did some more tests in some sample data I made for debugging ANTs jacobians (linked in the issue). 

This PR makes the filter more comparable to what you would get by iterating over the displacement field voxels and calling the `ComputeJacobianWithRespectToPosition` function at each point. It's not exactly the same, because the finite difference sampling scheme is not the same, although the filter allows users to set custom weights, so maybe it could be made more similar. 

One other thing (which I can fix): in the test, the seeming typo variable name "dispacementfield" was used throughout one function, I thought it might be intentional, so I didn't change it. But happy to update if needed.

### DIFF
--- a/Modules/Filtering/DisplacementField/include/itkDisplacementFieldJacobianDeterminantFilter.h
+++ b/Modules/Filtering/DisplacementField/include/itkDisplacementFieldJacobianDeterminantFilter.h
@@ -81,14 +81,9 @@ namespace itk
  * C array of TRealValue type.
  *
  * \par Constraints
- * We use vnl_det for determinant computation, which only supports square
- * matrices. So the vector dimension of the input image values must be equal
- * to the image dimensions, which is trivially true for a deformation field
- * that maps an n-dimensional space onto itself.
-
- * Currently, dimensions up to and including 4 are supported. This
- * limitation comes from the presence of vnl_det() functions for matrices of
- * dimension up to 4x4.
+ * The vector dimension of the input image values must equal the image
+ * dimensions, which is trivially true for a deformation field that maps
+ * an n-dimensional space onto itself.
  *
  * The template parameter TRealType must be floating point (float or double) or
  * a user-defined "real" numerical type with arithmetic operations defined
@@ -150,6 +145,10 @@ public:
   using RealType = TRealType;
   using RealVectorType = Vector<TRealType, InputPixelType::Dimension>;
   using RealVectorImageType = Image<RealVectorType, TInputImage::ImageDimension>;
+
+  /** Matrix type used to store the input direction matrix and the gradient matrix from
+      which the Jacobian determinant is computed. */
+  using InternalMatrixType = vnl_matrix_fixed<TRealType, ImageDimension, ImageDimension>;
 
   /** Type of the iterator that will be used to move through the image.  Also
       the type which will be passed to the evaluate function */
@@ -271,6 +270,8 @@ private:
   typename ImageBaseType::ConstPointer m_RealValuedInputImage{};
 
   RadiusType m_NeighborhoodRadius{};
+
+  InternalMatrixType m_InputDirection{};
 };
 } // end namespace itk
 
--- a/Modules/Filtering/DisplacementField/include/itkDisplacementFieldJacobianDeterminantFilter.h
+++ b/Modules/Filtering/DisplacementField/include/itkDisplacementFieldJacobianDeterminantFilter.h
@@ -21,7 +21,8 @@
 #include "itkNeighborhoodIterator.h"
 #include "itkImageToImageFilter.h"
 #include "itkVector.h"
-#include "vnl/vnl_matrix.h"
+#include "vnl/vnl_matrix_fixed.h"
+#include "vnl/vnl_vector_fixed.h"
 #include "vnl/vnl_det.h"
 
 namespace itk
@@ -146,10 +147,6 @@ public:
   using RealVectorType = Vector<TRealType, InputPixelType::Dimension>;
   using RealVectorImageType = Image<RealVectorType, TInputImage::ImageDimension>;
 
-  /** Matrix type used to store the input direction matrix and the gradient matrix from
-      which the Jacobian determinant is computed. */
-  using InternalMatrixType = vnl_matrix_fixed<TRealType, ImageDimension, ImageDimension>;
-
   /** Type of the iterator that will be used to move through the image.  Also
       the type which will be passed to the evaluate function */
   using ConstNeighborhoodIteratorType = ConstNeighborhoodIterator<RealVectorImageType>;
@@ -263,6 +260,10 @@ protected:
   WeightsType m_HalfDerivativeWeights{};
 
 private:
+  /** Matrix type used to store the input direction matrix and the gradient matrix from
+      which the Jacobian determinant is computed. */
+  using InternalMatrixType = vnl_matrix_fixed<TRealType, ImageDimension, ImageDimension>;
+
   bool m_UseImageSpacing{ true };
 
   ThreadIdType m_RequestedNumberOfWorkUnits{};
--- a/Modules/Filtering/DisplacementField/test/itkDisplacementFieldJacobianDeterminantFilterTest.cxx
+++ b/Modules/Filtering/DisplacementField/test/itkDisplacementFieldJacobianDeterminantFilterTest.cxx
@@ -18,8 +18,10 @@
 
 #include <iostream>
 #include "itkDisplacementFieldJacobianDeterminantFilter.h"
+#include "itkFlipImageFilter.h"
 #include "itkImageRegionIteratorWithIndex.h"
 #include "itkNullImageToImageFilterDriver.hxx"
+#include "itkPermuteAxesImageFilter.h"
 #include "itkStdStreamStateSave.h"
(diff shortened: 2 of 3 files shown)


Modules/Filtering/DisplacementField/test/itkDisplacementFieldJacobianDeterminantFilterTest.cxx
```diff
@@ -118,7 +120,113 @@ TestDisplacementJacobianDeterminantValue()
...
   {
-    std::cout << "Test passed." << std::endl;
+    std::cout << "Determinant value test passed" << std::endl;
+  }
+
+  // Now test that the determinant is consistent if the voxel space is different, but the physical space is the same.
+
+  // Define physical point to test
+  itk::Point<double, 2> physPt;
+  dispacementfield->TransformIndexToPhysicalPoint(index, physPt);
```
dzenanz: `dispacementfield` is clearly a typo. While fixing, also turn it into camelCase: `displacementField`.

Modules/Filtering/DisplacementField/test/itkDisplacementFieldJacobianDeterminantFilterTest.cxx
```diff
@@ -118,7 +120,113 @@ TestDisplacementJacobianDeterminantValue()
...
+    std::cerr.precision(static_cast<int>(itk::Math::abs(std::log10(epsilon))));
+    std::cerr << "Test failed " << std::endl;
+    std::cerr << "Error in pixel value at physical point " << physPt << std::endl;
+    std::cerr << "Expected value " << detOriginal << std::endl;
+    std::cerr << " differs from " << expectedJacobianDeterminant;
+    std::cerr << " by more than " << epsilon << std::endl;
+    testPassed = false;
+  }
+
+  using PermuteFilterType = itk::PermuteAxesImageFilter<FieldType>;
```
dzenanz, line 159: I think this long-ish code of permutations can be replaced by a few invocations of `OrientImageFilter`, which updates direction according to permutations to retain same physical space.

dzenanz (commented):
Thanks for tackling this. I did not have a good look. I still have some remarks.

cookpa:
Thanks, I'll update on both of these points

Modules/Filtering/DisplacementField/test/itkDisplacementFieldJacobianDeterminantFilterTest.cxx
cookpa: Hi, I tried to do this but it looks like OrientImageFilter requires 3D input.

I could change the test data to 3D but I think it would change the ground truth in the original test.

dzenanz: Ah, that did not occur to me.

blowekamp:
@hjmjohnson Why are we merging main into branches. I thought our standard work flow was to rebase, to avoid an ugly commit history to extraneous merges?

hjmjohnson:
> @hjmjohnson Why are we merging main into branches. I thought our standard work flow was to rebase, to avoid an ugly commit history to extraneous merges?

It was a mistake.  It tried using "gh pr update-branch" which did not do what I wanted.

I have re-pushed the changes without the merge.

Modules/Filtering/DisplacementField/include/itkDisplacementFieldJacobianDeterminantFilter.h
```diff
@@ -151,6 +151,10 @@ class ITK_TEMPLATE_EXPORT DisplacementFieldJacobianDeterminantFilter
   using RealVectorType = Vector<TRealType, InputPixelType::Dimension>;
   using RealVectorImageType = Image<RealVectorType, TInputImage::ImageDimension>;
 
+  /** Matrix type used to store the input direction matrix and the gradient matrix from
+      which the Jacobian determinant is computed. */
+  using InternalMatrixType = vnl_matrix_fixed<TRealType, ImageDimension, ImageDimension>;
+
```
N-Dekker: Please move this typedef (`InternalMatrixType`) to the private section of the class (below here), if it is not intended to be part of the public interface.

(If it _is_ part of the public interface, it should be tested 🤷)

cookpa: Thanks, this is just used internally so you're quite right, it should be private.

Also, I updated the imports to specifically include vnl_matrix_fixed.h, which is used here, and vnl_vector_fixed.h, which is used in the hxx file

N-Dekker: Thanks @cookpa 👍

hjmjohnson:
Rebased onto current `upstream/main` (was 1914 commits behind) and force-pushed to refresh CI on a current base. Conflict was a trivial `#include` ordering merge in the test file; nothing semantic changed. Local build + targeted test (`itkDisplacementFieldJacobianDeterminantFilterTest`) pass on the rebased tip `77ae8409`.

@greptileai review this PR — it's been quiet for a few months and I'd like a fresh draft pass before nudging for approval.

cc @dzenanz @N-Dekker — your last threads on this PR were resolved with thanks; if you're satisfied with the current state, would either of you mind clicking the green Approve button so this can land? It's a real bug fix (orientation-dependent Jacobian) closing #5358 and the author has been responsive.

<details>
<summary>Rebase details</summary>

- Was at: `eb8ccb55` (Jan 29, 2026), 1914 commits behind `upstream/main`
- Now at: `77ae8409` rebased onto `c237a1b7` (May 1, 2026)
- 4 commits preserved: `ENH: Test...`, `BUG: Jacobians not in physical space`, `STYLE: Fix variable name typo`, `STYLE: Move InternalMatrixType to private...`
- Conflict resolution: `Modules/Filtering/DisplacementField/test/itkDisplacementFieldJacobianDeterminantFilterTest.cxx` lines 21-25 — both branches added new `#include` lines in the same alphabetical neighborhood. Kept both.
- Local: `pre-commit run --all-files` clean; `ctest -R itkDisplacementFieldJacobianDeterminantFilterTest` passes in 1.16 sec.

</details>

hjmjohnson:
@greptileai — addressed all four P2 findings on rebased commit `b3aeeca6`:

| # | Status | Where |
|---|---|---|
| 1 | ✅ Fixed | Added `#include "vnl/algo/vnl_determinant.h"` to `.hxx` (folded into the BUG: commit) |
| 2 | ✅ Fixed | Stale `vnl_det` 4D-cap doc comment removed; the `Constraints` paragraph now describes only the current vector-dimension == image-dimension constraint (folded into the BUG: commit) |
| 3 | ✅ Fixed | Swapped error labels in the new physical-point sub-test (folded into the ENH: test commit) |
| 4 | ❌ Kept as-is | The `1e-13` tolerance in the new permute/flip sub-test is intentional. The implementation produces bit-identical determinants under axis permutation/flip with the physical-space gradient fix — the tight tolerance documents that bit-exactness IS the contract, not just current behavior. Loosening to `1e-5` would weaken a passing regression guard. If you'd prefer a more idiomatic comparison, `EXPECT_EQ` (in a future GTest conversion) would express the same intent more clearly. |

Local pre-commit `--all-files` clean; targeted test passes locally on the rewritten history.

hjmjohnson:
@cookpa — taking advantage of `maintainerCanModify` to push a one-line `-Wshadow` fix that was failing both `ARMBUILD-Ubuntu-24.04-arm` and `ITK.Linux` (CDash: warn=2 err=0 test_fail=0 test_pass=3486; both stem from the same warning). The lambda's local `auto filter = FilterType::New()` at `Modules/Filtering/DisplacementField/test/itkDisplacementFieldJacobianDeterminantFilterTest.cxx:135` shadowed the outer-scope `filter` declared earlier in the same function. Renamed the lambda local to `innerFilter`, fixup'd into `BUG: Jacobians not in physical space` (now `d650b088`) so the commit chain stays clean.

Verified locally: `ITKDisplacementFieldTestDriver` rebuilds clean (no `-Wshadow` warning) and `ctest -R itkDisplacementFieldJacobianDeterminantFilterTest` passes. Pre-commit clean across the rewritten tree.

Push: `b3aeeca6ee` → `9d08d3512a` via `--force-with-lease`. Happy to revert if you'd rather take it.

cookpa:
This all looks great, thanks @hjmjohnson

hjmjohnson:
@ntustison Would you mind having a look at this PR?  Most of the work was done by @cookpa, but I used agentic-coding agents to bring it up to date and address the reviewers' comments.  

It builds cleanly, but could use a domain expert for a second set of eyes.

Thanks,
Hans

ntustison:
@hjmjohnson .  Yeah, no problem.  I'll get to it today and try to finish tomorrow.  Thanks for all this work.

ntustison (approved):
Thanks again @hjmjohnson and @cookpa .  Everything looks good to me.

hjmjohnson (approved):

PR #5391, opened 2025-06-12
 Make `GetComponentTypeAsString` and `GetPixelTypeAsString` functions static 

N-Dekker:
Follow-up to commit 2cedb035e9c35a22236c4b64de7cd621c1415b92, "ENH: Make string/type conversion functions static in ImageIOBase", David Doria, May 10, 2011.

Adjusted the calls to those member functions, following [clang-tidy static-accessed-through-instance](https://clang.llvm.org/extra/clang-tidy/checks/readability/static-accessed-through-instance.html
) ("Checks for member expressions that access static members through instances").


thewtex (approved):
Thanks, Neils!

PR #5392, opened 2025-06-12
STYLE: Assign initial values to variables

hjmjohnson:
Checks whether there are local variables that are declared without an initial value. These may lead to unexpected behaviour if there is a code path that reads the variable before assigning to it.

Only integers, booleans, floats, doubles, and pointers are checked. The fix option initializes all detected values to zero. An exception is float and double types, which are initialized to NaN.

cd
run-clang-tidy.py -extra-arg=-D__clang__ -checks=-*,cppcoreguidelines-init-variables  -header-filter=.* -fix

Initialize non-initialized variables to default values (e.g., NaN, nullptr, zero, etc.). Manually review and prefer initialization to assignment of default values.  Place variable initialization in a smaller scope.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkGaussianDerivativeOperator.hxx
+++ b/Modules/Core/Common/include/itkGaussianDerivativeOperator.hxx
@@ -41,14 +41,10 @@ GaussianDerivativeOperator<TPixel, VDimension, TAllocator>::GenerateCoefficients
 
 
   // Calculate scale-space normalization factor for derivatives
-  double norm;
+  double norm = 1.0;
   if (m_NormalizeAcrossScale)
   {
     norm = std::pow(m_Variance, m_Order / 2.0);
-  }
-  else
-  {
-    norm = 1.0;
   }
 
   // additional normalization for spacing
@@ -166,7 +162,7 @@ template <typename TPixel, unsigned int VDimension, typename TAllocator>
 double
 GaussianDerivativeOperator<TPixel, VDimension, TAllocator>::ModifiedBesselI0(double y)
 {
-  double accumulator;
+  double accumulator = NAN;
 
   const double d = itk::Math::abs(y);
   if (d < 3.75)
@@ -195,7 +191,7 @@ template <typename TPixel, unsigned int VDimension, typename TAllocator>
 double
 GaussianDerivativeOperator<TPixel, VDimension, TAllocator>::ModifiedBesselI1(double y)
 {
-  double       accumulator;
+  double       accumulator = NAN;
   const double d = itk::Math::abs(y);
   if (d < 3.75)
   {
@@ -235,7 +231,7 @@ GaussianDerivativeOperator<TPixel, VDimension, TAllocator>::ModifiedBesselI(int 
                                                                                                  // placeholder
   }
 
-  double accumulator;
+  double accumulator = NAN;
   if (y == 0.0)
   {
     return 0.0;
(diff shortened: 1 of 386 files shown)


Modules/Core/Common/include/itkSymmetricSecondRankTensor.hxx
```diff
@@ -197,17 +197,8 @@ template <typename T, unsigned int VDimension>
...
-  {
-    k = row * Dimension + col - row * (row + 1) / 2;
-  }
-  else
-  {
-    k = col * Dimension + row - col * (col + 1) / 2;
-  }
-
+  unsigned int k =
+    (row < col) ? row * Dimension + col - row * (row + 1) / 2 : col * Dimension + row - col * (col + 1) / 2;
```
dzenanz, line 201: Isn't this contrary to ITK style guide?

hjmjohnson: In this case, the style guide conflicts with "prefer-initialization to assignment".   

It could also be initialized with a lambda function, but that seems more convoluted.  

To me, this seems like the most straightforward way to initialize.

hjmjohnson: In some cases, this will also allow variables to be declared as const (and sometimes constexpr), as will be updated in a future PR..

Modules/Core/Common/include/itkGaussianDerivativeOperator.hxx
```diff
@@ -41,15 +41,11 @@ GaussianDerivativeOperator<TPixel, VDimension, TAllocator>::GenerateCoefficients
...
-  double norm;
+  double norm = 1.0;
   if (m_NormalizeAcrossScale)
   {
     norm = std::pow(m_Variance, m_Order / 2.0);
   }
-  else
-  {
-    norm = 1.0;
-  }
```
N-Dekker, line 52: Thanks! Honestly I would prefer:

    double norm = m_NormalizeAcrossScale ? std::pow(m_Variance, m_Order / 2.0) : 1.0;

Rather than first initializing by `1.0` and then resetting to `std::pow(m_Variance, m_Order / 2.0)`.

Modules/Filtering/ImageIntensity/include/itkHistogramMatchingImageFilter.hxx
```diff
@@ -337,8 +337,8 @@ HistogramMatchingImageFilter<TInputImage, TOutputImage, THistogramMeasurement>::
         break;
       }
     }
-
-    double mappedValue;
+    // Linear interpolate from point[j] and point[j+1].
+    double mappedValue = m_QuantileTable[1][j - 1] + (srcValue - m_QuantileTable[0][j - 1]) * m_Gradients[j - 1];
```
issakomi, line 341: The value of **j** can be `0`, isn't it? Look at `break` on line 337 and `if (j == 0)` on line 342. If so, there is a regression, s.  `[j - 1]` on line 341. This may be related (not 100% sure) to new [Valgrind defects](https://open.cdash.org/builds/10467500/dynamic_analysis) in itkHistogramMatchingImageFilterTest.

issakomi: v

Please fix

dzenanz: @hjmjohnson Can you address this please?

blowekamp: I think there is a separate NIFT valgrind defect on the dash board too. This one has so many related defects is hard to see how many others there are. issakomi has done a good job identifying the issue so it should be straightforward to address, or atleast refer the offending change.

PR #5393, opened 2025-06-12
 No longer declare `ImageIOBase::GetComponentTypeInfo()` deprecated, add GoogleTest unit test 

N-Dekker:
`ImageIOBase::GetComponentTypeInfo()` was declared "DEPRECATED" by commit 6ff5fc7f01148bd11070c5ecfe0d56f643f62565, Kent Williams, 9 Feb 2011. However, it is still there after all those years, and it still appears useful.

Added a GoogleTest unit test to test this member function.

### DIFF
--- a/Modules/IO/ImageBase/include/itkImageIOBase.h
+++ b/Modules/IO/ImageBase/include/itkImageIOBase.h
@@ -250,9 +250,6 @@ public:
   itkGetEnumMacro(ComponentType, IOComponentEnum);
 
   /** get the type_info for the current pixel component type.
-   * This function is DEPRECATED and only provided for backwards
-   * compatibility.  There is no use for this method that can't
-   * be satisfied by calling GetComponentType.
    */
   virtual const std::type_info &
   GetComponentTypeInfo() const;
--- a/Modules/IO/ImageBase/test/CMakeLists.txt
+++ b/Modules/IO/ImageBase/test/CMakeLists.txt
@@ -977,6 +977,8 @@ itk_add_test(
   COMMAND
   itkUnicodeIOTest)
 
-set(ITKIOImageBaseGTests itkWriteImageFunctionGTest.cxx
+set(ITKIOImageBaseGTests
+  itkImageIOBaseGTest.cxx
+  itkWriteImageFunctionGTest.cxx
 )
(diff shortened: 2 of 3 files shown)


Modules/IO/ImageBase/include/itkImageIOBase.h
```diff
@@ -250,9 +250,6 @@ class ITKIOImageBase_EXPORT ImageIOBase : public LightProcessObject
   itkGetEnumMacro(ComponentType, IOComponentEnum);
 
   /** get the type_info for the current pixel component type.
-   * This function is DEPRECATED and only provided for backwards
-   * compatibility.  There is no use for this method that can't
-   * be satisfied by calling GetComponentType.
    */
   virtual const std::type_info &
   GetComponentTypeInfo() const;
```
dzenanz, line 255: This does not seem to be used by anything within the main repository. Do you have some code which uses it? If not, we could indeed remove this method now, as it has been deprecated for 14 years.

N-Dekker: Good question @dzenanz. I recently used `GetComponentTypeInfo()` for a unit test of `ImageFileCastWriter`, which is specific to SuperElastix/elastix: https://github.com/SuperElastix/elastix/blob/0380b5324aa6b4ea8262d763cb159ee7ba3a86f4/Common/GTesting/itkImageFileCastWriterGTest.cxx#L142

We recently added support for "fixed-width type names", like "int8" and "float64", for output pixel types. The elastix unit test checks that the "ComponentTypeInfo" of a written image file _matches_ the user-specified fixed-width type (meaning that they have the same "type traits").

Honestly I don't think `GetComponentTypeInfo()` is perfect. But for this elastix unit test to compile, we need to be sure that the member function is there. Otherwise I guess I just need to adjust the elastix unit test  🤷‍♂️

N-Dekker: @dzenanz OK, I just stopped using `GetComponentTypeInfo()` in elastix: https://github.com/SuperElastix/elastix/pull/1336 https://github.com/SuperElastix/elastix/commit/3119f35b1d12d107eea7a1e8fb7a0909b3de3e84

N-Dekker:
- Closed, as it is superseded by pull request #5408 commit c4112c01536e1e21c745fa2097d9149a23482bf9

PR #5395, opened 2025-06-14
Prefer Initialization to Assignment in constructors.

hjmjohnson:
Use member initializations in the constructor body which can be converted into
member initializers of the constructor instead. This not only improves the
readability of the code but also positively affects its performance.
Class-member assignments inside a control statement or following the first
control statement are ignored.

This check implements C.49 from the C++ Core Guidelines.

Please note, that this check does not enforce rule C.48 from the C++ Core
Guidelines. For that purpose see check modernize-use-default-member-init.

Followed up by the natural simplifications:

Converts a default constructor’s member initializers into the new
default member initializers in C++11. Other member initializers that match the
default member initializer are removed. This can reduce repeated code or allow
use of ‘= default’.

## PR Checklist


thewtex (approved):

Modules/Core/Common/include/itkAutoPointer.h
```diff
@@ -57,11 +57,11 @@ class AutoPointer
...
   explicit AutoPointer(AutoPointer & p)
-  {
-    m_IsOwner = p.IsOwner();          // Ownership can only be taken from
-                                      // another owner
-    m_Pointer = p.ReleaseOwnership(); // release ownership if appropriate
-  }
+    // Ownership can only be taken from another owner
+    : m_Pointer(p.ReleaseOwnership())
+    // release ownership if appropriate
+    , m_IsOwner(p.IsOwner())
```
issakomi, line 63: Why itkAutoPointerTest [leaks memory](https://open.cdash.org/viewDynamicAnalysisFile.php?id=11305171) after this change? Is it possible that the order of operations was important (1st: take ownership, 2nd: release) ? What was wrong with the previous ctor?

blowekamp: I think you are correct, p.ReleaseOwnership() changes the value of p.IsOwner()

blowekamp: I'll make a PR....

PR #5400, opened 2025-06-16
BUG: fix an infinite loop in itk::VoronoiDiagram2DGenerator

ILoveGoulash:
The loop exit condition was present in the code, but never wired and removed via https://github.com/InsightSoftwareConsortium/ITK/commit/cd97879e6bfec6cd9a33a3dd374d68cf4ff4303e

## PR Checklist

### DIFF
--- a/Modules/Segmentation/Voronoi/include/itkVoronoiDiagram2DGenerator.hxx
+++ b/Modules/Segmentation/Voronoi/include/itkVoronoiDiagram2DGenerator.hxx
@@ -274,8 +274,10 @@ VoronoiDiagram2DGenerator<TCoordinate>::ConstructDiagram()
     buildEdges.push_back(curr);
     EdgeInfo front = curr;
     EdgeInfo back = curr;
-    while (!(rawEdges[i].empty()))
+    auto     maxStop = static_cast<int>(rawEdges[i].size());
+    while (!(rawEdges[i].empty()) && (maxStop > 0))
     {
+      --maxStop;
       curr = rawEdges[i].front();
       rawEdges[i].pop_front();
       unsigned char frontbnd = Pointonbnd(front[0]);


dzenanz:
@bradking Is ghostflow acting up? Commit [62457de](https://github.com/InsightSoftwareConsortium/ITK/commit/62457dec8703b4506501660fd1520abec4397b8a) has a very short commit message.

ILoveGoulash:
Nah, the full "http://..." link is probably what triggers it, GitHub prettifies it on display.

Modules/Segmentation/Voronoi/include/itkVoronoiDiagram2DGenerator.hxx
```diff
@@ -274,8 +274,10 @@ VoronoiDiagram2DGenerator<TCoordinate>::ConstructDiagram()
     buildEdges.push_back(curr);
     EdgeInfo front = curr;
     EdgeInfo back = curr;
-    while (!(rawEdges[i].empty()))
+    auto     maxStop = static_cast<int>(rawEdges[i].size());
```
hjmjohnson: Why is there a static_cast to a signed int here?  What if there are more edges than will fit in the positive range of an integer?

I think maxStop could just be the same type as size() ( likely size_t ).

ILoveGoulash: The `> 0` condition can't work on an unsigned. It'd make sense to cast it to a bigger type or use `size_t` and check for `!= SIZE_MAX`, though. Which one do you prefer?

N-Dekker: @ILoveGoulash Thank you for your pull request. 👍  Specifically regarding your comment here, I think you're mistaken! When `u` is an unsigned number, the condition `u > 0` just works fine! (It is equivalent to the condition `u != 0`.)

HTH, Niels

hjmjohnson:
@ILoveGoulash Thank you for your contribution.  There are just a few final actions that need to be finalized so that we can incorporate this into the main tree.

Modules/Segmentation/Voronoi/include/itkVoronoiDiagram2DGenerator.hxx
ILoveGoulash: You're right. I'll try to find the time tomorrow to fix that.

ILoveGoulash:
Should be okay now.

hjmjohnson:
@ILoveGoulash I rebased and removed the merge conflict, then updated the commit message for the regression test to be more descriptive.

Modules/Segmentation/Voronoi/test/itkVoronoiDiagram2DInfiniteLoopTest.cxx
```diff
@@ -0,0 +1,48 @@
...
+
+#include "itkVoronoiDiagram2DGenerator.h"
+#include "itkTestingMacros.h"
+
+int
+itkVoronoiDiagram2DInfiniteLoopTest(int argc, char * argv[])
+{
+  if (argc != 1)
+  {
+    std::cerr << "Takes no parameters, but " << argc << "parameters given" << std::endl;
```
N-Dekker, line 27: When the test has no parameters, preferably make it a GoogleTest. GoogleTest unit tests are typically more "to the point", avoiding unnecessary verbosity.

Modules/Segmentation/Voronoi/include/itkVoronoiDiagram2DGenerator.hxx
```diff
@@ -273,8 +273,10 @@ VoronoiDiagram2DGenerator<TCoordinate>::ConstructDiagram()
     buildEdges.push_back(curr);
     EdgeInfo front = curr;
     EdgeInfo back = curr;
-    while (!(rawEdges[i].empty()))
+    auto     maxStop = rawEdges[i].size();
+    while (!(rawEdges[i].empty()) && (maxStop != 0))
```
N-Dekker, line 277: Do I understand correctly that the check `!(rawEdges[i].empty())` is redundant now? I mean, could it just say:

```cpp
    while ( maxStop > 0 )
```

If so, I would prefer such a simplification.

N-Dekker: Couldn't simply be a `for` loop?

```cpp
    for ( auto maxStop = rawEdges[i].size(); maxStop != 0; --maxStop ) 
```

Maybe then rename `maxStop` to `j`:
```cpp
    for ( auto j = rawEdges[i].size(); j != 0; --j ) 
```

hjmjohnson: @N-Dekker   I've looked into this some more.  The problem is more difficult than the requested change you proposed. I think the issue is that we are running into a corner case of the algorithm that is not handled. 

This PR adds a test that induces an infinite loop.  The initially requested change breaks the infinite loop but changes the algorithm's behavior.  (other test starts failing).

I haven't figured out what needs to be done, but it likely requires careful consideration of the implemented algorithm and more careful identification of corner cases.

The crux of the problem is removing items from the list for processing, then adding them back to the list if none of the described cases are met (i.e. an infinite loop occurs).

hjmjohnson:
Superseded by #6017 (merged to main) and #6052 (backported to release-5.4). Thank you @ILoveGoulash for the original report and fix — it directly informed the final implementation.

PR #5404, opened 2025-06-16
STYLE: Use C++17 [[nodiscard]] attribute

hjmjohnson:
Adds [[nodiscard]] attributes (introduced in C++17) to member functions in
order to highlight at compile time which return values should not be ignored.CONTRIBUTING (CONTRIBUTING.md) guide. 

## PR Checklist


Examples/DataRepresentation/Image/ImageAdaptor3.cxx
```diff
@@ -73,7 +73,7 @@ class VectorPixelAccessor
 
   VectorPixelAccessor &
   operator=(const VectorPixelAccessor & vpa) = default;
-  ExternalType
```
dzenanz, line 76: How was it decided which getters to decorate with this?

seanm: I'm curious about that as well.

This was a huge change. (As a random aside, I often find ITK PRs are merged too quickly, before there's any time to even notice they exist, but I digress...)

Personally, I find ignored return values to be most often bugs, and so I'd rather see compilers warn about unused return values by default, without needing to add a bazillion decorations like this.  I'm not aware of any current compiler that has this option though, but clang-tidy can warn about all ignored return values, and deliberate ignoring of return values can be suppressed with a void cast like:

`(void)malloc(1000)`

Might be interesting to see how many warnings clang-tidy would have for that...

Modules/Core/Common/test/itkMatrixTest.cxx
```diff
@@ -332,10 +332,10 @@ itkMatrixTest(int, char *[])
 
   using LargeMatrixType = itk::Matrix<NumericType, 7, 7>;
   auto matrixBad = itk::MakeFilled<LargeMatrixType>(2.0);
-  ITK_TRY_EXPECT_EXCEPTION(matrixBad.GetInverse());
+  ITK_TRY_EXPECT_EXCEPTION([[maybe_unused]] auto inverse = matrixBad.GetInverse());
```
N-Dekker, line 335: FYI, the modern way to explicitly ignore a return value is by `std::ignore`. In this case:

```cpp
    ITK_TRY_EXPECT_EXCEPTION(std::ignore = matrixBad.GetInverse());
```

As is suggested in the [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es48-avoid-casts) This approach (using `std::ignore` to ignore return values) is officially supported from supported since C++26, but I guess it usually just works.

Modules/Core/Common/include/itkSymmetricEigenAnalysis.h
```diff
@@ -510,7 +510,7 @@ class ITK_TEMPLATE_EXPORT SymmetricEigenAnalysis
    * `TMatrix` (which is the case when `TMatrix` = `itk::Array2D`).
    */
   template <typename QMatrix = TMatrix>
-  auto
+  [[nodiscard]] [[nodiscard]] auto
```
N-Dekker, line 513: Did clang-tidy add those duplicated `[[nodiscard]] [[nodiscard]]` here, automatically?

- To be addressed by pull request #5506

PR #5407, opened 2025-06-16
STYLE: Use C++17 nested namespace syntax

hjmjohnson:
Replace nested namespaces such as 'namespace a { namespace b { ... } }' and changes to the more concise syntax introduced in C++17: 'namespace a::b { ... }'.

Inline namespaces are not modified.

Disabled KWStyle Namespace checking that is incapable of supporting the c++17 nested namespace formatting.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkBinaryOperationConcept.h
+++ b/Modules/Core/Common/include/itkBinaryOperationConcept.h
@@ -22,11 +22,8 @@
 #include "itkPromoteType.h"
 
 /// \cond HIDE_META_PROGRAMMING
-namespace itk
-{
-namespace Details
-{
-namespace op
+
+namespace itk::Details::op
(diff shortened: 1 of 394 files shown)


hjmjohnson:
This is a style proposal to make use of the C++17 formatted nested namespace conventions.  I think that this compact format is much easier to read and maintain. It does require disabling the KWStyle Namespace checking, because that checker does not support nested namespaces.  

Perhaps it would be easy to update KWStyle to accept "itk *[:]+.*" instead of the exact "itk" string.

dzenanz (approved):
I like this.

thewtex (approved):

seanm:
Do you make this change with a linter tool? Manually? Custom script?

PR #5410, opened 2025-06-17
ENH: Fixed width type enums (INT8, UINT64, FLOAT64, ...) for IOComponent

N-Dekker:
Added enumerators to `IOComponent` that explicitly specify the number of bits of the types they refer to:

    UINT8, INT8, UINT16, INT16, UINT32, INT32, UINT64, INT64, FLOAT32, FLOAT64

Tested that `MapPixelType<TPixel>::CType` maps to the expected enumerator.

`INT8` is clearer than `CHAR`, because the C++ type `char` may be either signed or unsigned. (`INT8` is mapped to `std::int8_t`, which is always signed, by definition.) `INT32` and `INT64` are clearer than `LONG`, because `long` may be either 32-bit or 64-bit, depending on the platform and the compiler. `FLOAT32` is clearer than `FLOAT`, because on other languages (notable Python), the standard `float` type may be 64 bits.


hjmjohnson:
Windows failure is due to Github removing support for windows-2019

hjmjohnson (approved):

seanm:
This change introduced -Wduplicate-enum warnings into ITK, and there were none before.

Any objection to wrapping the enum in a pragma push/pop disabling that warning around that enum?

PR #5414, opened 2025-06-18
Fix valgrind warnings

hjmjohnson:
Fix the two Valgrind-identified failures.

The incorrect manual refactoring of itkHistogramMatchingImageFilter.hxx in  1f374cfa4797cd580e1bbd1d5fbf59d996253297 changed the logic in a way
That would allow for uninitialized memory reads.

## PR Checklist

### DIFF
--- a/Modules/Filtering/ImageIntensity/include/itkHistogramMatchingImageFilter.hxx
+++ b/Modules/Filtering/ImageIntensity/include/itkHistogramMatchingImageFilter.hxx
@@ -337,18 +337,21 @@ HistogramMatchingImageFilter<TInputImage, TOutputImage, THistogramMeasurement>::
         break;
       }
     }
-    // Linear interpolate from point[j] and point[j+1].
-    double mappedValue = m_QuantileTable[1][j - 1] + (srcValue - m_QuantileTable[0][j - 1]) * m_Gradients[j - 1];
-    if (j == 0)
-    {
-      // Linear interpolate from min to point[0]
-      mappedValue = m_ReferenceMinValue + (srcValue - m_SourceMinValue) * m_LowerGradient;
-    }
-    else if (j == m_NumberOfMatchPoints + 2)
-    {
-      // Linear interpolate from point[m_NumberOfMatchPoints+1] to max
-      mappedValue = m_ReferenceMaxValue + (srcValue - m_SourceMaxValue) * m_UpperGradient;
-    }
+
+    const double mappedValue = [this, j, srcValue]() -> double {
+      if (j == 0)
+      {
+        // Linearly interpolate from min to point[0]
+        return m_ReferenceMinValue + (srcValue - m_SourceMinValue) * m_LowerGradient;
+      }
+      if (j == m_NumberOfMatchPoints + 2)
+      {
+        // Linearly interpolate from point[m_NumberOfMatchPoints+1] to max
+        return m_ReferenceMaxValue + (srcValue - m_SourceMaxValue) * m_UpperGradient;
+      }
+      // Linearly interpolate from point[j] and point[j+1].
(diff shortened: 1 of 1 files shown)


thewtex (approved):
:1st_place_medal:

hjmjohnson:
@blowekamp There might still be a memory leak with itkAutoPointer.h, but I am unable to replicate that memory leak with valgrind on my computer.  

I am walking through the debugger manually, and it seems to be working correctly, and valgrind is not reporting a leak in my local build.  

I'll have a separate PR that makes debugging this a little easier, but does not change anything in the base class.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -395,13 +395,27 @@ NiftiImageIO::GenerateStreamableReadRegionFromRequestedRegion(const ImageIORegio
 // conversions between proxy and nifti_image pointer and arrow syntax (e.g., m_NiftiImage->data).
 class NiftiImageIO::NiftiImageProxy
 {
-  nifti_image * m_ptr;
```
blowekamp: This class is becoming very close to std::unique_ptr with a custom deleter.

blowekamp (approved):
Thank you for address these defects. The histogramMatching change looks good.

The NIFTI changes look like it close to a RAII pattern, but not there. I think it's likely that the dashboard issue is addressed, but other hidden resource issues may exist. I can look into using the unique_ptr today if you make this PR just the histogram changes. Otherwise, these changes are in the right direction to fix the exposed issues on the dashboard.

blowekamp:
> @blowekamp There might still be a memory leak with itkAutoPointer.h, but I am unable to replicate that memory leak with valgrind on my computer.
> 
> I am walking through the debugger manually, and it seems to be working correctly, and valgrind is not reporting a leak in my local build.
> 
> I'll have a separate PR that makes debugging this a little easier, but does not change anything in the base class.

I don't see the AutoPointer failure on the current valgrind dashboard, so I think it was resolved.
https://open.cdash.org/builds/10478407/dynamic_analysis

hjmjohnson:
> I don't see the AutoPointer failure on the current valgrind dashboard, so I think it was resolved.
> https://open.cdash.org/builds/10478407/dynamic_analysis

Thanks Brad!  This is good news.  I spent a few hours trying to replicate the problem and manually step through trying to replicate the problem.  All looks good.!

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -395,13 +395,27 @@ NiftiImageIO::GenerateStreamableReadRegionFromRequestedRegion(const ImageIORegio
...
+public:
```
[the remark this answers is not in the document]
dzenanz: Copilot beat me to this suggestion.

dzenanz (commented):
Looks good on a glance.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
hjmjohnson: @blowekamp   I am going to abandon my work on the Nifti memory leaks.  I look forward to seeing your more elegant solution.

I was using a sledge hammer to squash the memory leak.  Once I found the problem, I just kept pounding out all the possible paths where I could find that the problem might occur.

IF YOU DO NOT HAVE RESOURCES TO FIX THE PROBLEM ELEGANTLY .... We may want to commit this ugly fix that seems to work, and then revisit a more elegant solution later.  I will leave it to you to determine the best path for squashing this bug.

hjmjohnson: See #5415

Issue #5415, opened 2025-06-18
Valgrind memory leak with itkNiftiImageIO.cxx


hjmjohnson:
### Description
BUG: Add method for freeing `nifti_image` memory to improve
 safety

Introduced `free_image_from_nifti_clib` in `NiftiImageProxy` to encapsulate
and consistently perform memory cleanup. Replaced direct `nifti_image_free`
calls throughout the codebase for improved maintainability and safety.

Every exit path must destory the nifti_c library calloc'ed memory that is
held by the itkNiftiImageIO->m_NiftiImage handle.

These changes resolve a valgrind memory leak bug.

 Modules/IO/NIFTI/src/itkNiftiImageIO.cxx

### Steps to Reproduce

```c++
  // Force NiftiIO as IO object.
  const itk::NiftiImageIO::Pointer                   NiftiIO = itk::NiftiImageIO::New();
  const itk::ImageFileWriter<TestImageType>::Pointer writer = itk::ImageFileWriter<TestImageType>::New();
  writer->SetImageIO(NiftiIO);
  writer->SetFileName(testFilename);
  writer->SetInput(inputImageNoSform);
  ITK_TRY_EXPECT_NO_EXCEPTION(writer->Update());

  // This time it should read from the newly written "sform" code in the image, which should
  // be the same as reading from qform of the original image
  const itk::ImageFileReader<TestImageType>::Pointer reader = itk::ImageFileReader<TestImageType>::New();
  reader->SetImageIO(NiftiIO);
  reader->SetFileName(testFilename);
  ITK_TRY_EXPECT_NO_EXCEPTION(reader->Update());
  const TestImageType::Pointer reReadImage = reader->GetOutput();
```

### Expected behavior
No memory leak

### Actual behavior

Memory leak of 'nim' object held by NiftiImageIO object shared between the read and the write held by the 

NiftiImageIO::NiftiImageProxy class.

The 'nim' object generated after calling 'nifti_simple_init_nim' is not always deleted when done with it.
```
  // fill out the image header.
  if (this->m_NiftiImage == nullptr)
  {
    this->m_NiftiImage = nifti_simple_init_nim();
  }
``



### Reproducibility
always

### Versions
I think this is a long-standing memory leak.  I think it is suppressed on some of the dashboards that we had previously depended upon.

PR #5416, opened 2025-06-18
Use unique_pointer with nifti deleter

blowekamp:
The prior implementation had two members holding pointer, without clear or consitent practices with keeping the pointers valid or manually deleting them. This caused memory leaks detected by valgrind.

Refactor to use one RIAA unique_pointer for the nifti_image object. Modify the nifti anonymous structure to allow for forward declaration.

Addresses #5415.

## PR Checklist

### DIFF
--- a/Modules/IO/NIFTI/include/itkNiftiImageIO.h
+++ b/Modules/IO/NIFTI/include/itkNiftiImageIO.h
@@ -25,6 +25,7 @@
 #include <memory>
 #include "itkImageIOBase.h"
 
+
 namespace itk
 {
 
@@ -251,6 +252,11 @@ protected:
   }
 
 private:
+  // This proxy class has a nifti_image as a smart pointer to encapsulate access to the header.
+  class NiftiImageProxy;
+
+  std::unique_ptr<NiftiImageProxy> m_Holder;
+
   // Try to use the Q and S form codes from MetaDataDictionary if they are specified
   // there, otherwise default to the backwards compatible values from earlier
   // versions of ITK. The qform guess would probably been better to have
@@ -274,16 +280,6 @@ private:
 
   void
   SetImageIOMetadataFromNIfTI();
-
-  // This proxy class provides a nifti_image pointer interface to the internal implementation
-  // of itk::NiftiImageIO, while hiding the niftilib interface from the external ITK interface.
-  class NiftiImageProxy;
-
-  // Note that it is essential that m_NiftiImageHolder is defined before m_NiftiImage, to ensure that
-  // m_NiftiImage can directly get a proxy from m_NiftiImageHolder during NiftiImageIO construction.
-  const std::unique_ptr<NiftiImageProxy> m_NiftiImageHolder;
-
-  NiftiImageProxy & m_NiftiImage;
 
   double m_RescaleSlope{ 1.0 };
   double m_RescaleIntercept{ 0.0 };
--- a/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
+++ b/Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
@@ -15,6 +15,7 @@
  *  limitations under the License.
  *
  *=========================================================================*/
+
 #include "itkNiftiImageIO.h"
 #include "itkIOCommon.h"
 #include "itkMetaDataObject.h"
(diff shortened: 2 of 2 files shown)


Modules/IO/NIFTI/include/itkNiftiImageIO.h
```diff
@@ -25,6 +25,10 @@
 #include <memory>
 #include "itkImageIOBase.h"
 
+
```
dzenanz, line 28: Spelling errors in the commit message:
- with not clear -> without clear
- anonomouse -> anonymous

Modules/IO/NIFTI/include/itkNiftiImageIO.h
```diff
@@ -25,6 +25,10 @@
...
+// Forward declaration of the nifti_image structure
+struct nifti_image_struct;
```
dzenanz: Could the existing `nifti_image` typedef be used here?

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -391,33 +391,19 @@ NiftiImageIO::GenerateStreamableReadRegionFromRequestedRegion(const ImageIORegio
...
-  {}
-
-  operator nifti_image *() { return m_ptr; }
-
-  nifti_image *
-  operator->()
+  if (ptr)
   {
-    return m_ptr;
+    nifti_image_free(ptr);
```
dzenanz, line 401: Shouldn't we also have `ptr = nullptr;` after this call (which boils down to `free(ptr)`)?

dzenanz (commented):
Mostly looks good, some remarks inline.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
blowekamp: This class method is defining the deleter used for the unique_pointer.  `ptr` is a parameter of the method and not permeant. The value of pointer is handled by unique_pointer.

I may be miss understanding what you are commenting on.

Modules/IO/NIFTI/include/itkNiftiImageIO.h
blowekamp: We cannot include the Nifti header here. The original nifti_image type was a typedef of an anonymous struct.

I do not know of a better solution.

Modules/ThirdParty/NIFTI/src/nifti/niftilib/nifti1_io.h
```diff
@@ -109,7 +109,7 @@ typedef enum _analyze75_orient_code {
            nifti1 format definition; it is used to implement one API
            for reading/writing formats in the nifti1 format.
  */
-typedef struct {                /*!< Image storage struct **/
+typedef struct nifti_image_struct {                /*!< Image storage struct **/
```
hjmjohnson: @blowekamp When this is approved, let's make sure to update the upstream to add this named structure.

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -391,33 +391,19 @@ NiftiImageIO::GenerateStreamableReadRegionFromRequestedRegion(const ImageIORegio
...
-public:
-  NiftiImageProxy(nifti_image * ptr)
-    : m_ptr(ptr)
-  {}
-
-  operator nifti_image *() { return m_ptr; }
-
-  nifti_image *
-  operator->()
+  if (ptr)
```
seanm: No need for this NULL check, like free(NULL), nifti_image_free(NULL) does nothing.

Modules/ThirdParty/NIFTI/src/nifti/niftilib/nifti1_io.h
seanm: First, I don't think this PR should change niftilib itself at all.  That should be separate.

But anyway, I think better for ITK to just use `struct nifti_image` than to change upstream. It would complicate regression testing different versions of ITK and different versions of niftilib.

seanm (changes requested):

Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
hjmjohnson: +1

Modules/ThirdParty/NIFTI/src/nifti/niftilib/nifti1_io.h
hjmjohnson: After reading @seanm comment.  I think you could use:

```c++
using nifti_image_struct = struct nifti_image; 
```
to avoid modifying the upstream.

hjmjohnson: Perhaps that does not work with the "foward declaration" ??

blowekamp: Yes, continuing to support non-ITK NIFTI libraries is import.

Anything similar to that does not work, be cause of the anonymous structure in NIFTI.

blowekamp:
OK, redid with another layer of encapsulation. It was a little hurried and I'm not going to be able to return to this until Monday. Someone else is welcomed to continue this work.

hjmjohnson (approved):

seanm (approved):
OK cool, so you were able to not change nifticlib itself, great!

Modules/ThirdParty/NIFTI/src/nifti/niftilib/nifti1_io.h
hjmjohnson: Thanks for looking into it.

PR #5419, opened 2025-06-18
COMP: Use value initialization to zero-initialize thread id variable

jhlegarreta:
Use value initialization to zero-initialize thread id variable.

Fixes:
```
In file included from
 Core/Common/src/itkPlatformMultiThreader.cxx:35:
[CTest: warning matched]
 Core/Common/src/itkPlatformMultiThreaderPosix.cxx:172:28:
 warning: zero as null pointer constant [-Wzero-as-null-pointer-constant]
  172 |   pthread_t threadHandle = 0;
      |                            ^
      |                            nullptr
```

raised for example in:
https://open.cdash.org/viewBuildError.php?type=1&buildid=10477917

and
```
ITK/Modules/Core/Common/include/itkThreadSupport.h:45:55:
 warning: zero as null pointer constant [-Wzero-as-null-pointer-constant]
   45 | constexpr ThreadProcessIdType ITK_DEFAULT_THREAD_ID = 0;
      |                                                       ^
      |                                                       nullptr
```

reported by a local build.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkThreadSupport.h
+++ b/Modules/Core/Common/include/itkThreadSupport.h
@@ -42,7 +42,7 @@ using FastMutexType = pthread_mutex_t;
 using ThreadFunctionType = void * (*)(void *);
 using ThreadProcessIdType = pthread_t;
 #  if !defined(ITK_FUTURE_LEGACY_REMOVE)
-constexpr ThreadProcessIdType ITK_DEFAULT_THREAD_ID = 0;
+constexpr ThreadProcessIdType ITK_DEFAULT_THREAD_ID = {};
 #  endif
 using ITK_THREAD_RETURN_TYPE = void *;
 /** @ITKEndGrouping */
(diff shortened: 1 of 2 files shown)


thewtex (approved):
:100:

seanm:
I've finally caught up on all my ITK github emails, and see only now that you changed this to *not* use nullptr.  This new way is probably fine.

IIRC this was the last -Wzero-as-null-pointer-constant in ITK!

seanm:
Tried locally, I see just one more (though of course I did not build every possible module or permutation):

```
ITK/Modules/Core/Common/include/itkThreadSupport.h:45:55: warning: zero as null pointer constant [-Wzero-as-null-pointer-constant]
   45 | constexpr ThreadProcessIdType ITK_DEFAULT_THREAD_ID = 0;
      |                                                       ^
      |                                                       nullptr
```

Want to amend this PR to fix that one too?

dzenanz (approved):

hjmjohnson:
> IIRC `pthread_t` is not a pointer on all platforms, I remember fixing this myself once, then having to back out of it...
> 
> I have a branch where I started on #3136 which would fix the issue in a different way.

I have the same recollection.  I've had to back out that change in the past as well. 

It would be great to have the proper #ifdef initializaitons for when it is needed to be 0 and when it should be nullptr.

hjmjohnson (approved):

PR #5422, opened 2025-06-19
COMP: suppress Clang -Wduplicate-enum

seanm:


dzenanz (approved):

dzenanz:
Interesting style change by clang-format.

seanm:
> Interesting style change by clang-format.

Mmmm, which style change?  I added more #if branches, so things got reindented...

Modules/Core/Common/include/itkMacro.h
```diff
@@ -121,12 +121,22 @@ namespace itk
...
+#    define ITK_CLANG_SUPPRESS_Wzero_as_null_pointer_constant \
+      ITK_PRAGMA(clang diagnostic ignored "-Wzero-as-null-pointer-constant")
+#  else
+#    define ITK_CLANG_SUPPRESS_Wzero_as_null_pointer_constant
+#  endif
+#  if __has_warning("-Wduplicate-enum")
+#    define ITK_CLANG_SUPPRESS_Wduplicate_enum ITK_PRAGMA(clang diagnostic ignored "-Wduplicate-enum")
+#  else
+#    define ITK_CLANG_SUPPRESS_Wduplicate_enum
+#  endif
```
N-Dekker, line 134: I see that the `if __has_warning` is necessary to avoid another warning (_warning: unknown warning group '-W...', ignored [-Wunknown-warning-option]_). Unfortunately it's a bit cumbersome to have 5 lines of boilerplate code to define such a macro (`ITK_CLANG_SUPPRESS_Wduplicate_enum`), for just one warning. Would it be possible to make macro definitions like this more concise? And would it be possible to just specify the string "-Wduplicate-enum" once, instead of twice?  I'm just asking. 🤷 No show stopper.

seanm:
> I meant what clang-format wants by its complaint in pre-commit check.

*maybe* I've fixed it with my change to the clang-format config file...

N-Dekker (approved):
Thanks @seanm 

If you ever find a way to define macro's like `ITK_CLANG_SUPPRESS_Wduplicate_enum` in a more concise manner, please let us know. Approved anyway  👍

N-Dekker:
> By the way, it appears that an enumerator like `INT8 = CHAR` would pass without a warning, even without your pull request

To further elaborate:

```cpp
  enum class MyEnum
  {
    UNKNOWN,
    CHAR,
    INT8 = CHAR,  // Intended duplicate value, no problem, no warning!
    WONDERFUL = 42,  // OK, new enum value
    DUPLICATE_ZERO = 0  // <== Warning! Duplicate value!
  };
```

See https://godbolt.org/z/78r883836

Issue #5369, opened 2025-05-28
Proposal to Update ITK's Minimum macOS Deployment Target to 10.15 (Catalina)


thewtex:
#### **1. Apple’s Deprecation of Older macOS Versions**

- macOS 10.15 (Catalina) loses **security support on June 30, 2025**, creating vulnerabilities for users on older systems[^2].
- Current ITK Python builds default to macOS 10.9 (Intel), macOS 10.11 (ARM), which are no longer receiving security updates[^2][^7].

#### **2. C++17 and SDK Compatibility**

- Full C++17 support (e.g., `<filesystem>`, `<variant>`) requires **macOS 10.15 or newer**[^4][^8][^10].
- Legacy deployment targets (e.g., 10.9–10.13) force workarounds like `<experimental/filesystem>`, increasing maintenance complexity[^4][^5].
- Modern compilers (Xcode 15+) and SDKs increasingly enforce stricter compatibility checks, causing build failures on older targets[^6][^9].

#### **3. Compiler Optimizations**

- Newer macOS SDKs enable advanced optimizations (e.g., autovectorization, LTO) that are restricted on older deployment targets.

#### **4. Maintenance Burden**

- Third-party projects (e.g., PyTorch) are dropping support for macOS <10.15, creating compatibility gaps[^1][^6][^10].

---

### **Proposed Changes**

1. **Update Minimum Deployment Target**
    - Set `MACOSX_DEPLOYMENT_TARGET` to **10.15** for ITK 6.x in Python wheels.
    - Document 10.15 as the new baseline for macOS support.
2. **CI/CD Pipeline Updates**
    - Update macOS runners with `macos-15`+ runners[^1][^9] We may need self-hosted runners for Intel support (#5326).
3. **C++17 Modernization**
    - Replace deprecated headers (e.g., `<experimental/filesystem>`) with standard C++17 equivalents.

---

### **Impact Mitigation**

- **LTS Branch**: Maintain ITK 5.x with macOS 10.9 support for legacy workflows [^1][^3].

---

### **Supporting Data**

| Key Point | Source |
| :-- | :-- |
| macOS 10.15 security EOL: June 2025 | [^2][^7] |
| C++17 `<filesystem>` requires 10.15 | [^4][^8][^10] |
| Xcode 15+ enforces 10.13+ targets | [^6][^9] |

---

### **Request for Feedback**

1. Are there critical use cases requiring macOS <10.15?
[...]

PR #5427, opened 2025-06-20
COMP: removed workarounds for old Apple stuff

seanm:
- removed workaround for Xcode 10
- emit #error for AppleClang less than Xcode 12.4
- removed workaround for broken fail() added in 1fcd00774bf65728cc4c9e2e7f10750a48334ca2 in 2003. Hopefully the workaround is no longer needed. I'm not sure, but all tests pass on my arm64 macOS 15.5.

Related to issue #5369.

### DIFF
--- a/Modules/IO/ImageBase/src/itkImageIOBase.cxx
+++ b/Modules/IO/ImageBase/src/itkImageIOBase.cxx
@@ -303,12 +303,7 @@ ImageIOBase::ReadBufferAsBinary(std::istream & is, void * buffer, ImageIOBase::S
 
   const std::streamsize numberOfBytesRead = is.gcount();
 
-#ifdef __APPLE_CC__
-  // fail() is broken in the Mac. It returns true when reaches eof().
-  if (numberOfBytesRead != numberOfBytesToBeRead)
-#else
   if ((numberOfBytesRead != numberOfBytesToBeRead) || is.fail())
-#endif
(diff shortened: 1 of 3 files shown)


seanm:
Not sure about the 1 commit for the `fail()` failure...

Modules/IO/ImageBase/src/itkImageIOBase.cxx
```diff
@@ -303,7 +303,7 @@ ImageIOBase::ReadBufferAsBinary(std::istream & is, void * buffer, ImageIOBase::S
 
   const std::streamsize numberOfBytesRead = is.gcount();
 
-#ifdef __APPLE_CC__
+#if 0 // was: #ifdef __APPLE_CC__
```
dzenanz: I would rather remove the entire ifdef, leaving only the `#else` part.

seanm: Thing is, I don't know if this workaround is still necessary or not, so I thought if removing it breaks something, leaving this breadcrumb would help someone find the bug again.

Maybe I should just restore this part...

dzenanz: This looks like code which is exercised a lot, and if this breaks it CI should fail.

thewtex (approved):

PR #5435, opened 2025-06-23
BUG: Use --replace-all for branch.main.rebase config

thewtex:
Some configurations may encounter:

 warning: branch.main.rebase has multiple values
 error: cannot overwrite multiple values with a single value
       Use a regexp, --add or --replace-all to change branch.main.rebase.


dzenanz (approved):
Looks good on a glance.

PR #5442, opened 2025-06-26
BUG: Use launcher for local Python pre-commit hooks

thewtex:
Python may not be available on Windows systems. Use a launcher script to run local Python pre-commit hooks with the Python executable we generate when bootstrapping our pre-commit hooks. The launcher script uses the local Python, which we know will be available and have a sufficient version. It also allows adjusting the path with Windows vs Unix.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -102,7 +102,7 @@ repos:
       stages: [pre-commit]
     - id: kw-commit-msg
       name: 'kw commit-msg'
-      entry: 'Utilities/Hooks/kw-commit-msg.py'
+      entry: 'Utilities/Hooks/run-pixi-python.sh Utilities/Hooks/kw-commit-msg.py'
       language: system
       stages: [commit-msg]
       exclude: "\\/ThirdParty\\/"
--- a/Utilities/Hooks/run-pixi-python.sh
+++ b/Utilities/Hooks/run-pixi-python.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+# Cross-platform launcher for Pixi-Python launched local Python pre-commit hooks
+
+# Determine Python executable path based on OS
+if [[ "$OSTYPE" == "msys" || "$OSTYPE" == "cygwin" || "$OSTYPE" == "mingw" || "$OSTYPE" == "win32" ]]; then
+    PYTHON_EXE="./.pixi/envs/pre-commit/python.exe"
+else
+    PYTHON_EXE="./.pixi/envs/pre-commit/bin/python"
+fi
+
+# Verify Python executable exists
+if [[ ! -x "$PYTHON_EXE" ]]; then
+    echo "Error: Python executable not found at $PYTHON_EXE" >&2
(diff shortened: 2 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -102,7 +102,7 @@ repos:
       stages: [pre-commit]
     - id: kw-commit-msg
       name: 'kw commit-msg'
-      entry: 'Utilities/Hooks/kw-commit-msg.py'
+      entry: 'Utilities/Hooks/run-pixi-python.sh Utilities/Hooks/kw-commit-msg.py'
       language: system
```
blowekamp, line 106: I think the only change need it to change language: system -> language: python. The pre-commit environment already know the python environment. This is consistent with the reference hooks: https://github.com/pre-commit/pre-commit-hooks/blob/main/.pre-commit-hooks.yaml

blowekamp (commented):
Please consider the simpler approach use by the reference hooks.

.pre-commit-config.yaml
dzenanz: I tried this, and it seems to work.

N-Dekker: Just wondering, would it be sufficient to just explicitly mention `python`? As in:

      entry: 'python Utilities/Hooks/kw-commit-msg.py'

At least, the computers I tried (Windows) know how to run python from a command-line!

dzenanz: I opened a PR with this suggestion: #5443.

thewtex: >  system -> language: python

> just explicitly mention python

These are not going to work as robustly across all systems. The reference pre-commit python hooks set up their own Python environment. In general, we can not assume that Python is installed and available in the PATH.

thewtex: image

dzenanz (approved):

N-Dekker:
Shall I still try it out on Windows?

Utilities/Hooks/run-pixi-python.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+# Cross-platform launcher for Pixi-Python launched local Python pre-commit hooks
+
+# Determine Python executable path based on OS
+if [[ "$OSTYPE" == "msys" || "$OSTYPE" == "cygwin" || "$OSTYPE" == "mingw" || "$OSTYPE" == "win32"]]; then
```
N-Dekker: Looks like there should be a space before the first closing bracket: `"win32"]]` should be `"win32" ]]`

hjmjohnson: !! Great catch.

thewtex:
@N-Dekker thanks for testing 🙏 . I am not sure why I did not receive that error. Could you please test again?

N-Dekker (approved):
Tested again (on Windows), looks like it works now! (No error message anymore, when doing a commit of my own, afterwards.)  Thanks Matt!

PR #5450, opened 2025-07-03
Let `IOComponent::CHAR` represents `signed char`

N-Dekker:
- Added a note about `IOComponent::CHAR` representing `signed char`
- Let file readers interpret the `CHAR` enum as `signed char` (instead of just plain `char`)
- Adjusted tests accordingly

Triggered by a discussion with Bradley (@blowekamp) at https://discourse.itk.org/t/long-double-support-for-imageio-meshio-pixels/7570/11 and his comment at https://github.com/InsightSoftwareConsortium/ITK/pull/5449#issuecomment-3032091149

### DIFF
--- a/Modules/Core/Common/include/itkCommonEnums.h
+++ b/Modules/Core/Common/include/itkCommonEnums.h
@@ -71,6 +71,8 @@ public:
    * Enums used to manipulate the component type. The component type
    * refers to the actual storage class associated with either a
    * SCALAR pixel type or elements of a compound pixel.
+   *
+   * \note The enum `CHAR` represents `signed char` (not just plain `char`).
    */
   ITK_CLANG_PRAGMA_PUSH
   ITK_CLANG_SUPPRESS_Wduplicate_enum
--- a/Modules/Core/TestKernel/src/itkTestDriverInclude.cxx
+++ b/Modules/Core/TestKernel/src/itkTestDriverInclude.cxx
@@ -828,7 +828,7 @@ HashTestImage(const char * testImageFilename, const std::vector<std::string> & b
   switch (componentType)
   {
     case itk::IOComponentEnum::CHAR:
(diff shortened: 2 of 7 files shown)


blowekamp (approved):
Thank you for this change. This  change now matches how I have seen this enum type used for ImageIO. Getting this cleanup up and documented is a good step forward.

N-Dekker:
For the record, `signed char` was already mapped to `IOComponentEnum::CHAR` before, by calling `IMAGEIOBASE_TYPEMAP(signed char, IOComponentEnum::CHAR)`, at: https://github.com/InsightSoftwareConsortium/ITK/blob/0dd31a62f851a0ae95ef4c6d6c85e531a4ebca59/Modules/IO/ImageBase/include/itkImageIOBase.h#L900

This logic was syntactically restyled (but not altered, sematically) into `std::is_same_v<TComponent, signed char> ? IOComponentEnum::CHAR` by pull request #5438, at: https://github.com/InsightSoftwareConsortium/ITK/blob/09305ce7c3509139ab0ca12047f8e6793b01db74/Modules/IO/ImageBase/include/itkImageIOBase.h#L611

N-Dekker:
As far as I can see, `signed char` and `std::int8_t` denote the very same time, on all supported platforms: 

    static_assert(std::is_same_v<std::int8_t, signed char>);

On the other hand, `signed char` and `char` are always distinct types, by definition, even when `char` is a signed type: 

    static_assert(! std::is_same_v<signed char, char>);

https://godbolt.org/z/3dGcx4vdP

N-Dekker:
> I just saw that GCC has the `-funsigned-char` and `-fsigned-char` it might be an interesting experiment to see if ITK works with both states.

@blowekamp The test failures of my experiment show that ITK currently does _not_ work with `-funsigned-char` (letting the default `char` be an unsigned type)
- pull request #5455
Note that the test failures are partially caused by the tests, rather than by the library!

The failures can typically be fixed by explicitly using `signed char`, instead of just plain `char`. So this pull request (#5450) is a step in the right direction, I think  😇

PR #5451, opened 2025-07-03
STYLE: Deprecate `UnknownType` from both ImageIOBase and MeshIOBase

N-Dekker:
The empty nested `UnknownType` classes are no longer being used.


dzenanz (approved):
LGTM

dzenanz:
Are these types maybe needed for WASM @thewtex?

Issue #5456, opened 2025-07-08
HDF5 library installed file names in Debug vs Release


seanm:
If I build with BUILD_SHARED_LIBS=NO, CMAKE_BUILD_TYPE=Release, and `make install` I get HDFs libraries with files names:

- libitkhdf5_cpp.a
- libitkhdf5_hl_cpp.a
- libitkhdf5_hl.a
- libitkhdf5.a

Whereas if I build as CMAKE_BUILD_TYPE=Debug I get:

- libitkhdf5_cpp-static-6.0_debug.a
- libitkhdf5_hl_cpp_debug.a
- libitkhdf5_hl-static-6.0_debug.a
- libitkhdf5-static-6.0_debug.a

Notice 3 differences:

- the addition of `_debug`
- the addition of `-static` on 3 of 4
- the addition of `6.0` on 3 of 4

Is all this deliberate?  Is it ITK doing it or HDF5?  No other files have these differences between Debug and Release, only these 4.

Also, probably unrelated, I also just noticed while writing this, that 3 other files have `5.0` in their name instead of `6.0`.  Is that expected?

- libITKgiftiio-5.0.a
- libITKniftiio-5.0.a
- libITKznz-5.0.a

This is all with current master.

blowekamp:
As I recall there was a large amount of CMake changes from upstream HDF5 with the most recent update: #4940

There maybe a variable to controls this behavior?

blowekamp:
Regarding the nifti file names it's likely related to this line:
https://github.com/InsightSoftwareConsortium/ITK/blame/09305ce7c3509139ab0ca12047f8e6793b01db74/Modules/ThirdParty/NIFTI/src/CMakeLists.txt#L1

Not sure why that line is there.

seanm:
> Regarding the nifti file names it's likely related to this line: https://github.com/InsightSoftwareConsortium/ITK/blame/09305ce7c3509139ab0ca12047f8e6793b01db74/Modules/ThirdParty/NIFTI/src/CMakeLists.txt#L1

Confirmed. And same in `ITK/Modules/ThirdParty/GIFTI/src/CMakeLists.txt`

@hjmjohnson as you changed that line last, know if maybe it can be done another way, to get the current version from elsewhere instead of re-stating it multiple times?

dzenanz:
Was this fixed by #5457?

seanm:
[...]

PR #5457, opened 2025-07-09
Issue #5456: added missing refereneces to 4th HDF5 library

seanm:


blowekamp:
Please make this 2 PRs. One for Nifti/Gifti, and the other for HDF5.

blowekamp:
I think we can merge the Nifti/Gifti one more easily than the HDF5.

hjmjohnson (approved):
This looks like the correct implementation to add a missing set of libraries. to CMake.

hjmjohnson (approved):
It will be nice to have access to the high level HDF5 capabilities:

```txt
hdf5_hl_cpp is the C++ high-level HDF5 library component that provides simplified interfaces for working with HDF5 (Hierarchical Data Format version 5) files. It is part of the HDF5 C++ high-level API, layered on top of:
	•	hdf5: the core HDF5 C library
	•	hdf5_hl: the high-level HDF5 C library
	•	hdf5_cpp: the C++ API for HDF5
```

blowekamp:
> > There may be concern with this library being enabled
> 
> What do you mean by "enabled"? The library was already being installed before this change (though with a weird filename).

OK, good to know that this didn't add an additional library. I didn't look at it closely.

PR #5458, opened 2025-07-09
COMP: fixed warnings about unneeded semi-colons

seanm:


Modules/Core/Common/include/itkEventObject.h
```diff
@@ -144,7 +144,7 @@ operator<<(std::ostream & os, const EventObject & e)
 
 #define itkEventMacroDefinition(classname, super)                            \
   classname::classname(const classname & s)                                  \
-    : super(s) {};                                                           \
+    : super(s) {}                                                            \
```
N-Dekker: Looks like the pre-commit hook wants to have a new line, just for the `{}` 🤷

seanm: ffs. :) ok changed.

Modules/Bridge/VTK/include/itkVTKImageImport.h
```diff
@@ -30,7 +31,8 @@
       this->m_##name = _arg;                        \
       this->Modified();                             \
     }                                               \
-  }
+  }                                                 \
+  ITK_MACROEND_NOOP_STATEMENT
```
N-Dekker, line 35: Off-topic, but it looks like the macro that is fixed by this pull request (`itkSetMacro2`) isn't really necessary, is it? It's quite similar to the definition of `itkSetMacro`, at https://github.com/InsightSoftwareConsortium/ITK/blob/f9d5d97915bad80b95efacae3e8428352901247b/Modules/Core/Common/include/itkMacro.h#L1012-L1025

thewtex (approved):

PR #5455, opened 2025-07-08
WIP: [Do not merge] Change the default `char` type to unsigned

N-Dekker:
Just as an experiment!

Triggered by @blowekamp at https://github.com/InsightSoftwareConsortium/ITK/pull/5450#issuecomment-3048788622

PR #5465, opened 2025-07-11
STYLE: Replace `char` with `signed char` as itk::Image argument in tests

N-Dekker:
Replaced `itk::Image<char` with `itk::Image<signed char`, in all `"itk*Test*.cxx"` source files. Removed pixel type alias `InputDataType = char` from itkAntiAliasBinaryImageFilterTest, and directly used `signed char` instead.

The use of `char` as `TPixel` argument in tests appears to triggers test failures, when compiling with an unsigned default `char` type (GCC option `-funsigned-char` or MSVC option `/J`), for example:

    [ RUN      ] FrequencyIterators.Even2D
    Unequal images, with 29 unequal pixels
    test/itkFrequencyIteratorsGTest.cxx:260: Failure
    Value of: fullAndHermitian
    Actual: false
    Expected: true

It appears clearer to explicitly specify that the pixel type should be signed, by using `signed char` instead.

----

- Those test failures appeared with an experimental pull request: #5455 
- This pull request is inspired by a comment from Bradley (@blowekamp) at https://github.com/InsightSoftwareConsortium/ITK/pull/5450#issuecomment-3048788622
- Various coding guidelines recommend using `signed char`, rather than plain `char`, for numeric values, for example https://wiki.sei.cmu.edu/confluence/display/c/INT07-C.+Use+only+explicitly+signed+or+unsigned+char+type+for+numeric+values and https://rules.sonarsource.com/cpp/tag/confusing/RSPEC-810/

### DIFF
--- a/Modules/Core/Common/test/itkBSplineInterpolationWeightFunctionTest.cxx
+++ b/Modules/Core/Common/test/itkBSplineInterpolationWeightFunctionTest.cxx
@@ -274,7 +274,7 @@ itkBSplineInterpolationWeightFunctionTest(int, char *[])
     using KernelType = itk::BSplineKernelFunction<SplineOrder>;
     auto kernel = KernelType::New();
 
-    using ImageType = itk::Image<char, SpaceDimension>;
+    using ImageType = itk::Image<signed char, SpaceDimension>;
     auto                        image = ImageType::New();
     const ImageType::RegionType region{ startIndex, size };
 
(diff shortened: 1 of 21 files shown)


blowekamp (approved):
Looks good. With the IO changes, the IO related tests should clearly be signed char. In the level sets the char images are used for mask/label, in this case I am not sure the sign-ness was intended, but I don' think it matters.

PR #5466, opened 2025-07-11
Remove unnecessary ByteSwapper Swap function calls for char types

N-Dekker:
When a buffer consists of elements of size `sizeof(char)`, the Swap member functions of `ByteSwapper` don't do anything.

### DIFF
--- a/Modules/IO/BMP/src/itkBMPImageIO.cxx
+++ b/Modules/IO/BMP/src/itkBMPImageIO.cxx
@@ -610,29 +610,9 @@ BMPImageIO::SwapBytesIfNecessary(void * buffer, SizeValueType numberOfPixels)
   switch (m_ComponentType)
   {
     case IOComponentEnum::CHAR:
-    {
-      if (m_ByteOrder == IOByteOrderEnum::LittleEndian)
-      {
-        ByteSwapper<char>::SwapRangeFromSystemToLittleEndian(static_cast<char *>(buffer), numberOfPixels);
-      }
-      else if (m_ByteOrder == IOByteOrderEnum::BigEndian)
-      {
-        ByteSwapper<char>::SwapRangeFromSystemToBigEndian(static_cast<char *>(buffer), numberOfPixels);
-      }
-      break;
-    }
     case IOComponentEnum::UCHAR:
     {
-      if (m_ByteOrder == IOByteOrderEnum::LittleEndian)
-      {
-        ByteSwapper<unsigned char>::SwapRangeFromSystemToLittleEndian(static_cast<unsigned char *>(buffer),
-                                                                      numberOfPixels);
-      }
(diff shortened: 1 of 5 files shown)


Modules/IO/BMP/src/itkBMPImageIO.cxx
```diff
@@ -610,29 +610,9 @@ BMPImageIO::SwapBytesIfNecessary(void * buffer, SizeValueType numberOfPixels)
   switch (m_ComponentType)
   {
     case IOComponentEnum::CHAR:
```
hjmjohnson, line 612: Please add [[fallthrough]] to explicitly indicate that this is intended to fall through to break in the next case.

Modules/IO/PhilipsREC/src/itkPhilipsRECImageIO.cxx
```diff
@@ -299,10 +299,8 @@ PhilipsRECImageIO::SwapBytesIfNecessary(void * buffer, SizeValueType numberOfPix
     switch (this->m_ComponentType)
     {
       case IOComponentEnum::CHAR:
```
hjmjohnson, line 301: [[fallthrough]]		here too please.

Modules/IO/PhilipsREC/src/itkPhilipsRECImageIO.cxx
```diff
@@ -338,10 +336,8 @@ PhilipsRECImageIO::SwapBytesIfNecessary(void * buffer, SizeValueType numberOfPix
...
```
hjmjohnson, line 338: [[fallthrough]]

Modules/IO/BMP/src/itkBMPImageIO.cxx
N-Dekker: @hjmjohnson Thanks for the suggestion, Hans, but I believe that in this particular "case" the `[[fallthrough]]` attribute is not necessary. Because `case IOComponentEnum::CHAR` doesn't itself have any specific statement for its own case:

```cpp
    case IOComponentEnum::CHAR:
    case IOComponentEnum::UCHAR:
    {
      // ...
      break;
    }
```
Please look at the example at https://en.cppreference.com/w/cpp/language/attributes/fallthrough.html You see, it doesn't suggest a `[[fallthrough]]` between `case 1` and `case 2` either. Because there is no statement between `case 1` and `case 2` anyway:

```cpp
    switch (n)
    {
        case 1:
        case 2:
            g();
            [[fallthrough]];
        case 3: // no warning on fallthrough
            h();
```

Did you still encounter compiler warnings for such "empty" cases? If so, that might be a compiler bug  🤔

hjmjohnson (approved):

PR #5469, opened 2025-07-15
Remove byte swapping for pixel components of size 1 from Bruker2dseqImageIO and VTKImageIO

N-Dekker:
- Removed unnecessary `BYTE_SWAP` calls from `Bruker2dseqImageIO::SwapBytesIfNecessary`
- Removed expensive `MACRO(char)` calls from `VTKImageIO::Write`

- Follow-up to pull request #5466

### DIFF
--- a/Modules/IO/Bruker/src/itkBruker2dseqImageIO.cxx
+++ b/Modules/IO/Bruker/src/itkBruker2dseqImageIO.cxx
@@ -361,10 +361,8 @@ Bruker2dseqImageIO::SwapBytesIfNecessary(void * buff, SizeValueType components)
     switch (this->m_OnDiskComponentType)
     {
       case IOComponentEnum::CHAR:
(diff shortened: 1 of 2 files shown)


N-Dekker:
@thewtex Thanks for requesting a review from Copilot. How would you want us to process such a review? In this particular case, I did consider its suggestion, but doesn't seem very useful to me. Is it OK if I just ignore it, then?

thewtex:
@N-Dekker I think they Copilot reviews can be handled similar to other reviews. Considering the suggestions is a good approach, and the suggestions (especially the "[nitpick]"'s) do not necessarily need to be followed. It is helpful for others reviewers to know that you have had time to take a look by making a short reply, e.g. "Current style preferred", and clicking the "Resolve conversation" button.

Thanks for this patch! I'll merge.

thewtex (approved):

PR #5472, opened 2025-07-15
DOC: Add information to NRRD and GDCM IOs.

zivy:
Add details on how to set metadata information that is specific to the NRRD and DICOM file formats.

### DIFF
--- a/Modules/IO/GDCM/include/itkGDCMImageIO.h
+++ b/Modules/IO/GDCM/include/itkGDCMImageIO.h
@@ -78,6 +78,26 @@ extern ITKIOGDCM_EXPORT std::ostream &
  * The compressors supported include "JPEG2000" (default), and
  * "JPEG". The compression level parameter is not supported.
  *
+ * DICOM tags are represented as strings in the metadata
+ * dictionary. The string format is "XXXX,XXXX", DICOM group number followed
+ * by element number, both are hexadecimals. The separator character
+ * is either a pipe "|" or a comma ",".
+ *
+ * \warning As the metadata dictionary uses the DICOM tag strings as
+ * keys it is possible to have multiple entries representing the same
+ * DICOM tag with different values. The last one encountered will be
+ * used when writing the image to file. For example, the patient name tag
+ * "0010,0010" and "0010|0010" may both be in the dictionary with different values
+ * due to the different separator character. Similarly,
+ * the series description tag may appear multiple times as "0008,103e",
+ * "0008,103E", "0008|103e", or "0008|103E". The strings differ in
+ * the separator character and lower or upper case letters in the
+ * hexadecimal numbers. Note that when read from file, letters in
+ * the hexadecimal numbers are always set to lower case and the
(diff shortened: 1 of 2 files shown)


Modules/IO/GDCM/include/itkGDCMImageIO.h
```diff
@@ -78,6 +78,25 @@ extern ITKIOGDCM_EXPORT std::ostream &
  * The compressors supported include "JPEG2000" (default), and
  * "JPEG". The compression level parameter is not supported.
  *
+ * DICOM tags are represented as strings in the metadata
+ * dictionary. The string format is "XXXX,XXXX", DICOM group number followed
+ * by element number, both are hexadecimals. The separator character
+ * is either a pipe "|" or a comma ",".
```
N-Dekker, line 84: Thanks @zivy Do I understand correctly that a pipe "|" is preferable? GDCMImageIOTest2 says that tags are "always read back with the pipe separator": https://github.com/InsightSoftwareConsortium/ITK/blob/cd1fc6499f0dc5e560ec177db4b6b53e286d6659/Modules/IO/GDCM/test/itkGDCMImageIOTest2.cxx#L172-L174

If the pipe is indeed preferable, can you please add this to the documentation as well? For example by extending the last sentence of the warning, below:

```
                                                   [...] To ensure consistency,
 * it is best to always use lower case letters when explicitly adding
 * tags to the metadata dictionary, and use pipe "|" as separator.
```

zivy: @N-Dekker, nice catch. Thanks.
Will update the PR and force push.

Modules/IO/GDCM/include/itkGDCMImageIO.h
```diff
@@ -78,6 +78,26 @@ extern ITKIOGDCM_EXPORT std::ostream &
...
+ * DICOM tag with different values. The last one encountered will be
+ * used when writing the image to file. For example, the patient name tag
+ * "0010,0010" and "0010|0010" may both be in the dictionary with different values
+ * due to the different separator character. Similarly,
+ * the series description tag may appear multiple times as "0008,103e",
+ * "0008,103E", "0008|103e", or "0008|103E". The strings differ in
+ * the separator character and lower or upper case letters in the
+ * hexadecimal numbers. Note that when read from file, letters in
+ * the hexadecimal numbers are always set to lower case and the
+ * separator character is a pipe. To ensure consistency,
```
N-Dekker, line 97: I think you mean here: `the separator character is` _always set to_ `a pipe`. Right? No change needed, I'm just checking!

zivy: yes, that is correct.

zivy:
@N-Dekker, please approve and merge this PR. Thanks.

N-Dekker (approved):
Approved, but I'm not familiar with the NRRD file format, so it would still be nice if someone else can also have a look.

blowekamp (approved):

PR #5475, opened 2025-07-23
DOC: Spell precommit

dave3d:
## PR Checklist

This PR address a spell checking pre-commit to ITK.  So any time code is committed to an ITK repo, the SimpleITK/CommentSpellCheck (CSC) script is run on the new code.  CSC checks the spelling of all comments in new file.

Previously CSC was run on ITK as a Github Action on the ITK repository.  Moving to a pre-commit runs CSC on the local developer's system allowing for faster feedback and reducing Github server usage.  Therefore, this PR removes .github/workflows/spell-check-comments.yml.

Also this PR includes many spelling fixes to the C++ files in the ITK code base.  These bugs had not been address before because our main concern has been the spelling in the header files, which are used to create the documentation.  With this new PR, the .cxx are also checked, so we had to fixed all the bugs already in ITK.

The .cxx fixes are split into multiple commits for organizational sake.  All the commits in this PR can be squashed into one, if that is preferred.

UPDATE:  I also meant to say that the version of CSC has been upgraded to the latest 0.4.3.  This version uses a pure Python spelling backend, unlike the previous version.  That previous version depended on the C enchant library, which was not available on all systems.

### DIFF
--- a/Modules/Filtering/BinaryMathematicalMorphology/test/itkBinaryDilateImageFilterTest.cxx
+++ b/Modules/Filtering/BinaryMathematicalMorphology/test/itkBinaryDilateImageFilterTest.cxx
@@ -62,7 +62,7 @@ itkBinaryDilateImageFilterTest(int, char *[])
   inputImage->SetRegions(region);
   inputImage->Allocate();
 
-  // Declare Iterator types apropriated for each image
+  // Declare Iterator types appropriate for each image
   using myIteratorType = itk::ImageRegionIterator<myImageType>;
 
   // Create one iterator for image (this is a light object)
--- a/Modules/Segmentation/ConnectedComponents/test/itkThresholdMaximumConnectedComponentsImageFilterTest.cxx
+++ b/Modules/Segmentation/ConnectedComponents/test/itkThresholdMaximumConnectedComponentsImageFilterTest.cxx
@@ -23,7 +23,7 @@
  * works for the Insight Journal:
  *
  * Urish KL, August J, Huard J. "Unsupervised segmentation for myofiber
- * counting in immunoflourescent microscopy images". Insight Journal.
+ * counting in immunofluorescent microscopy images". Insight Journal.
  * ISC/NA-MIC/MICCAI Workshop on Open-Source Software (2005)
  * https://doi.org/10.54294/h1vbsl
  */
--- a/Documentation/Doxygen/Registration.dox
+++ b/Documentation/Doxygen/Registration.dox
@@ -39,7 +39,7 @@ elements. They are:
 \li \b Optimizer: the method used to find the \e Transform parameters that optimize the \e Metric.
 
 A particular registration method is defined by selecting specific implementations of each one of these basic elements.
-In order to determine the registration method appropriated for a particular problem, is will be useful to answer the following questions:
+In order to determine the registration method appropriate for a particular problem, is will be useful to answer the following questions:
 
 
 
--- a/Examples/Filtering/DigitallyReconstructedRadiograph1.cxx
+++ b/Examples/Filtering/DigitallyReconstructedRadiograph1.cxx
@@ -625,7 +625,7 @@ main(int argc, char * argv[])
 
   // Software Guide : BeginLatex
   //
-  // Having initialised the interpolator we pass the object to the
+  // Having initialized the interpolator we pass the object to the
   // resample filter.
   //
   // Software Guide : EndLatex
--- a/Modules/Core/Common/CMake/itkCheckHasGNUAttributeAligned.cxx
+++ b/Modules/Core/Common/CMake/itkCheckHasGNUAttributeAligned.cxx
@@ -22,7 +22,7 @@ namespace itk
 }
 
 //
-// Check if the compoler support the GNU attribute extension for
+// Check if the compiler support the GNU attribute extension for
 // alignment, and does not contain a bug which causes internal
 // compiler segfault.
 //
--- a/Modules/Core/Common/src/itkFloatingPointExceptions_win.cxx
+++ b/Modules/Core/Common/src/itkFloatingPointExceptions_win.cxx
@@ -56,7 +56,7 @@ FloatingPointExceptions::HasFloatingPointExceptionsSupport()
 
 #else // defined(_MSC_VER)
 
-// MinGW has troubles include'ing float.h.
+// MinGW has troubles include-ing float.h.
 
 void
 FloatingPointExceptions::Enable()
--- a/Modules/Core/Common/src/itkImageRegionSplitterMultidimensional.cxx
+++ b/Modules/Core/Common/src/itkImageRegionSplitterMultidimensional.cxx
@@ -114,7 +114,7 @@ ImageRegionSplitterMultidimensional::ComputeSplits(unsigned int         dim,
                                                    const SizeValueType  regionSize[],
                                                    unsigned int         splits[])
 {
-  // size of each splited region
+  // size of each split region
   std::vector<double> splitRegionSize(dim); // Note: stack allocation preferred
   unsigned int        numberOfPieces = 1;
 
--- a/Modules/Core/Common/src/itkMultiThreaderBase.cxx
+++ b/Modules/Core/Common/src/itkMultiThreaderBase.cxx
@@ -66,7 +66,7 @@ struct MultiThreaderBaseGlobals
 
   // GlobalDefaultThreaderTypeIsInitialized is used only in this
   // file to ensure that the ITK_GLOBAL_DEFAULT_THREADER or
-  // ITK_USE_THREADPOOL environmenal variables are
+  // ITK_USE_THREADPOOL environmental variables are
   // only used as a fall back option.  If the SetGlobalDefaultThreader
   // API is ever used by the developer, the developers choice is
   // respected over the environmental variable.
--- a/Modules/Core/Common/test/itkImageIteratorWithIndexTest.cxx
+++ b/Modules/Core/Common/test/itkImageIteratorWithIndexTest.cxx
@@ -87,7 +87,7 @@ public:
     while (!it.IsAtEnd())
     {
       const PixelType value = it.Get();
-      if (itk::Math::NotExactlyEquals(value, it.Get())) // check repeatibility
+      if (itk::Math::NotExactlyEquals(value, it.Get())) // check repeatability
       {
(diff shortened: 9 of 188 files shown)


Modules/Filtering/BinaryMathematicalMorphology/test/itkBinaryDilateImageFilterTest.cxx
```diff
@@ -62,7 +62,7 @@ itkBinaryDilateImageFilterTest(int, char *[])
   inputImage->SetRegions(region);
   inputImage->Allocate();
 
-  // Declare Iterator types apropriated for each image
+  // Declare Iterator types appropriated for each image
```
dzenanz: This should probably be appropriate, not appropriate**d** (extra 'd'). This appears in many places.

Modules/Segmentation/ConnectedComponents/test/itkThresholdMaximumConnectedComponentsImageFilterTest.cxx
```diff
@@ -23,7 +23,7 @@
  * works for the Insight Journal:
  *
  * Urish KL, August J, Huard J. "Unsupervised segmentation for myofiber
- * counting in immunoflourescent microscopy images". Insight Journal.
+ * counting in immunofluorescent microscopy images". Insight Journal.
```
dzenanz, line 26: The paper on the website has this, correct, spelling.

.github/workflows/spell-check-comments.yml
```diff
@@ -1,50 +0,0 @@
-name: Spell Checking Code Comments
```
dzenanz, line 1: `git commit -n` bypasses pre-commit hooks, right? I vote we keep this. It's overhead is small.

blowekamp: With the move to pre-commit, it should be run as part of that existing CI.

zivy: @dzenanz I agree with you (I've bypassed my own local pre-commit hooks in the past and was saved from myself by the CI). 

Having said that, I believe the commit message from @dave3d is incorrect. The spell checker will be run by the CI. I see that it was added to the .pre-commit-config.yaml, so will run when the [pre-commit](https://github.com/InsightSoftwareConsortium/ITK/blob/f71cfd8945374cf0d71d5cb4d805682f69cf94b5/.github/workflows/pre-commit.yml#L1) action is run. Am I missing something?

dzenanz: Ahh, I forgot that.

Modules/Segmentation/ConnectedComponents/test/itkThresholdMaximumConnectedComponentsImageFilterTest.cxx
dzenanz: This was a soft of interesting note, not an objection.

Modules/Filtering/BinaryMathematicalMorphology/test/itkBinaryDilateImageFilterTest.cxx
dave3d: Oops.  Fixed it.

dzenanz (approved):

dzenanz (approved):

.github/workflows/itk_dict.txt
```diff
@@ -42,12 +46,20 @@ Creade
 CreateListOfSamplesWithIDs
 Crofton's
 Cuadra
+Cygwin
 Cz
 DDataFrameobject
+Daniesson
```
dzenanz: This looks like a misspelling.

.github/workflows/itk_dict.txt
```diff
@@ -64,45 +76,61 @@ Elmaghraby
 Englewood
 Eun
 Evol
+ExhaustiveOptv
```
dzenanz: Why not rename [this](https://github.com/InsightSoftwareConsortium/ITK/blob/d0b07068a98e28d71d483aee7c9236adb70913ed/Modules/Numerics/Optimizersv4/test/itkExhaustiveOptimizerv4Test.cxx#L43), like many other things? It appears in a test, and is therefore not part of the public API.

.github/workflows/itk_dict.txt
```diff
@@ -64,45 +76,61 @@ Elmaghraby
...
 Hauke
 Heapsort
 Heibel
+Helminen
 Hideaki
 Hierarchique
 Hilden
 Hiraki
 Hirschberg
+Hmmm
```
dzenanz: This is not a necessary part of a [comment here](https://github.com/InsightSoftwareConsortium/ITK/blob/d0b07068a98e28d71d483aee7c9236adb70913ed/Modules/IO/MINC/src/itkMINCImageIO.cxx#L1162), and could be removed instead of added to the dictionary.

.github/workflows/itk_dict.txt
dave3d: Should I exclude all tests from the spell checking?

dzenanz: That would only make sense as a temporary measure. This PR is already big. You could disable spell check in tests, and add an issue to enable it. Then tackle it when you (or someone else) has enthusiasm and time to spare.

hjmjohnson (changes requested):
There are a few words that have been added to the dictionary that do not represent words (see @dzenanz comments).  I think those words need to be removed from the dictionary to avoid future false negative checks.

@dave3d Thank you for addressing these spelling errors.  It is GREAT to have these corrections put into place.  After removing the non-words from the dictionary, I think this is ready to merge.

.github/workflows/itk_dict.txt
hjmjohnson: Fixed in upcomming force push.

.github/workflows/itk_dict.txt
```diff
@@ -248,135 +327,226 @@ anisotropies
...
 coeffP
-composedness
+commandline
+comp
 complainty
+composedness
 computable
+confocal
 connectedness
+consec
```
dzenanz, line 446: maybe rename this to consecutive, and remove `consec` from the dictionary?

.github/workflows/itk_dict.txt
```diff
@@ -248,135 +327,226 @@ anisotropies
...
 denoiser
 denormal
 deriv
+desc
 descr
 detections
+devicetype
+devs
 dgr
+dhbo
```
dzenanz, line 519: these 3, and even `dgr` above maybe should not be in the dictionary?

.github/workflows/itk_dict.txt
```diff
@@ -426,42 +606,77 @@ extrapolators
...
+fegetexcept
+fem
 femobject
-fh
+femparray
+fenv
+feraiseexcept
+fesetenv
+fexcept
+fffffff
```
dzenanz, line 634: do we want this in dictionary?

.github/workflows/itk_dict.txt
```diff
@@ -681,160 +982,238 @@ normalised
...
 nxy
-nxz
 nxyz
+nxz
 ny
 nyz
 nz
 nzelts
+oXooooooooooooooooooo
+objectif
```
dzenanz, line 1008: this looks like a misspelling

.github/workflows/itk_dict.txt
```diff
@@ -681,160 +982,238 @@ normalised
...
 octant
 octants
-oindex
+offcentre
 ofstream
+oindex
+oitr
 oldnrm
+onedegree
+oooooooooooo
```
dzenanz, line 1019: we probably should not have these ooos in the dictionary

.github/workflows/itk_dict.txt
```diff
@@ -681,160 +982,238 @@ normalised
...
+optimisation
+optimization
 optimizersv
+orientable
 orienter
+orig
 orthant
 orthantwise
 ortho
+orthog
```
dzenanz, line 1037: we should rename the variable and remove this from dictionary

.github/workflows/itk_dict.txt
```diff
@@ -844,97 +1223,159 @@ sigmad
...
 tcol
+td
+tdo
+tech
+teletype
 textheight
 textit
 texttt
 textwidth
+tfrequency
```
dzenanz, line 1316: Does `tFrequency` work? It feels like we should not be adding this to the dictionary.

.github/workflows/itk_dict.txt
```diff
@@ -844,97 +1223,159 @@ sigmad
...
 tgz
+thetamin
```
dzenanz, line 1318: Does `thetaMin` pass spell check? What about `thetaPlus`?

.github/workflows/itk_dict.txt
```diff
@@ -844,97 +1223,159 @@ sigmad
...
+thetaplus
 thisDic
 threadIDs
 threadsafe
 thresholder
+thruview
```
dzenanz, line 1324: maybe rename to `thruView`?

.github/workflows/itk_dict.txt
```diff
@@ -951,59 +1392,92 @@ upsampler
...
+vop
+vox
 voxelwise
+vp
+vr
+vsize
 vv
 vxl
 vxl's
+waitid
```
dzenanz, line 1426: `waitId`, with uppercase i?

.github/workflows/itk_dict.txt
```diff
@@ -951,59 +1392,92 @@ upsampler
...
 wopen
 workgroup
 worksize
+writability
 wstring
+wunlink
 wv
 xarray
 xfff
+xffff
```
dzenanz, line 1454: both this and the "word" above feel like they should not be in the dictionary

dzenanz (commented):
I went though the new additions to the dictionary, and pointed out the ones which struck me. You don't have to accept all suggestions.

.github/workflows/itk_dict.txt
hjmjohnson: @dzenanz I don't disagree, but determining which of these should be changed became a slippery slope with no end in sight.  I would support any of these changes that you want to contribute, but I feel the current benefits of what currently exists far outweigh waiting until this is perfect.

dzenanz: OK, then merge once CI pass, but leave these unresolved. That way, somebody (potentially me) can tackle these in the future.

PR #5476, opened 2025-07-30
ENH: Support 5D files in NRRD IO

lassoan:
Several modern imaging devices and processing algorithms now creates 5D data sets, including:
- time sequence of 3D displacement fields (result of deformable registration of 4D images)
- time sequence of 3D velocity fields (acquired by 4D MRI flow images)
- time sequence of 3D color (RGB or RGBA) images (result of colorized volumes provided by AI segmentation)

NRRD IO can now read and write all these data types, and in general any 5D arrays (2 or 3 spatial dimensions + scalar/vector pixel type + optional list axis).

## PR Checklist
- [x] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/main/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved) - added a new option, by default previous behavior is preserved
- [x] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/main/CONTRIBUTING.md#design-changes) were made (or the changes have been approved) - no major design changes
- [x] Added test (or behavior not changed) - added test
- [x] Updated API documentation (or API not changed) - API documentation added
- [x] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/main/Utilities/KWStyle/ITKHeader.h) to new files (if any) - added license
- [x] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5 - only added new testing files
- [x] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples) for all new major features (if any) - existing file reading examples still apply, special use cases covered by tests

@hjmjohnson
@blowekamp

### DIFF
--- a/Modules/IO/NRRD/test/itkNrrd5dVectorImageReadWriteTest.cxx
+++ b/Modules/IO/NRRD/test/itkNrrd5dVectorImageReadWriteTest.cxx
@@ -1,0 +1,643 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         https://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+#include <fstream>
+#include "itkExtractImageFilter.h"
+#include "itkImageDuplicator.h"
+#include "itkImageFileReader.h"
+#include "itkImageFileWriter.h"
+#include "itkJoinSeriesImageFilter.h"
+#include "itkNrrdImageIO.h"
+#include "itkPermuteAxesImageFilter.h"
+#include "itkTestingMacros.h"
+
+// Test writing of time-varying displacement field into NRRD file.
+
+// Generate a sequence of 2D or 3D displacement field images
+template <class ImageType>
+void
+GenerateImageSequence(int numberOfTimePoints, std::vector<typename ImageType::Pointer> & imageList)
+{
+  typename ImageType::SizeType      size;
+  typename ImageType::IndexType     start;
+  typename ImageType::RegionType    region;
+  typename ImageType::DirectionType inDirection; // 3x3 matrix
+  typename ImageType::PointType     inOrigin;    // 3D vector
+  typename ImageType::SpacingType   inSpacing;   // 3D vector
+  // Region
+  size[0] = 8;
+  size[1] = 12;
+  if (ImageType::SizeType::Dimension > 2)
+  {
+    size[2] = 10;
+  }
+  start.Fill(0);
+  region.SetSize(size);
+  region.SetIndex(start);
+  // Origin
+  inOrigin[0] = 10.0;
+  inOrigin[1] = 20.0;
+  if (ImageType::PointType::Dimension > 2)
+  {
+    inOrigin[2] = 15.0;
+  }
+  // Direction (matrix indices: column, row)
+  inDirection[0][0] = 0.94;
+  inDirection[0][1] = 0.24;
+  inDirection[1][0] = -0.10;
+  inDirection[1][1] = 0.87;
+  if (ImageType::DirectionType::RowDimensions > 2)
+  {
+    inDirection[0][2] = -0.24;
+    inDirection[1][2] = 0.49;
+    inDirection[2][0] = 0.33;
+    inDirection[2][1] = -0.43;
+    inDirection[2][2] = 0.84;
+  }
+  // Spacing
+  inSpacing[0] = 1.2;
+  inSpacing[1] = 2.3;
+  if (ImageType::SpacingType::Dimension > 2)
+  {
+    inSpacing[2] = 4.1;
+  }
+  for (int timePoint = 0; timePoint < numberOfTimePoints; ++timePoint)
+  {
+    typename ImageType::Pointer image = ImageType::New();
+    image->SetRegions(region);
+    image->Allocate();
+
+    // Fill each image with a different constant value for testing
+    image->FillBuffer(typename ImageType::PixelType(timePoint * 2));
+
+    // Set common geometry
+    image->SetOrigin(inOrigin);
+    image->SetDirection(inDirection);
+    image->SetSpacing(inSpacing);
+    // Add to list
+    imageList.push_back(image);
+  }
+}
+
+//----------------------------------------------------------------------------
+template <class PixelType, int SpaceDimension>
+void
+WriteImageSequenceFile(std::string                                                            filename,
+                       std::vector<typename itk::Image<PixelType, SpaceDimension>::Pointer> & imageList,
+                       bool                                                                   listAxisFastest = false)
+{
+  typedef itk::Image<PixelType, SpaceDimension>     ImageType;
+  typedef itk::Image<PixelType, SpaceDimension + 1> ImageSequenceType;
+
+  typedef itk::JoinSeriesImageFilter<ImageType, ImageSequenceType> JoinImageFilterType;
+  typename JoinImageFilterType::Pointer                            joinImageFilter = JoinImageFilterType::New();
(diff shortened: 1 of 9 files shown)


blowekamp:
Thank you for the clear description of the "AxesReorder" option. I am going to download this PR, examine it, and take it for a spin.

I frequently have to do similar operations for microscopy images, I have had a different philosophy on how to handle these case. After I examine the PR, I'll have a longer post discussing this approach vs others.

Modules/IO/NRRD/test/itkNrrd5dVectorImageReadWriteTest.cxx
```diff
@@ -0,0 +1,652 @@
...
+      duplicator->Update();
+      imageList.push_back(duplicator->GetOutput());
+    }
+  }
+}
+
+//----------------------------------------------------------------------------
+template <class PixelType, int SpaceDimension>
+void
+ReadImageFile(std::string filename, typename itk::VectorImage<PixelType, SpaceDimension>::Pointer & image)
```
dzenanz, line 349: Could this function be replaced by regular ITK function? E.g. `image = itk::ReadImage<itk::VectorImage<PixelType, SpaceDimension>>(filename);`

Modules/IO/NRRD/test/itkNrrd5dVectorImageReadWriteTest.cxx
```diff
@@ -0,0 +1,652 @@
...
+  typedef itk::Image<PixelType, SpaceDimension + 1> ImageSequenceType;
+
+  typedef itk::JoinSeriesImageFilter<ImageType, ImageSequenceType> JoinImageFilterType;
+  typename JoinImageFilterType::Pointer                            joinImageFilter = JoinImageFilterType::New();
+  for (auto image : imageList)
+  {
+    joinImageFilter->PushBackInput(image);
+  }
+
+  ImageSequenceType::Pointer joinedImage;
```
blowekamp: "typename" is needed

Modules/IO/NRRD/test/itkNrrd5dVectorImageReadWriteTest.cxx
```diff
@@ -0,0 +1,652 @@
...
+    auto origin1 = joinedImage->GetOrigin();
+    std::cout << origin1 << std::endl;
+
+    listAxis = 0;
+    using PermuteFilterType = itk::PermuteAxesImageFilter<ImageSequenceType>;
+    typename PermuteFilterType::Pointer permuteFilter = PermuteFilterType::New();
+    permuteFilter->SetInput(joinImageFilter->GetOutput());
+
+    // Define new axis order: e.g., swap axis 0 and 2 (X <-> Z)
+    itk::FixedArray<unsigned int, ImageSequenceType::ImageDimension> order;
```
blowekamp: "typename" is needed

blowekamp (commented):
Noting a couple changes I needed to compile.

blowekamp:
I am still digging into NRRDIO. There are a couple things that are not as I expected:
 - There is not support for streaming IO. I thought that at least reading a region of interest was supported.
 -  The existing behavior already does permutations of the data, when there is one range/"multi-component" axis ( this is only when the range dimension is not the fastest, which I don't think I have encountered)
 - The meta-data reported by the ImageIO, appears to be modified from what is in the file. Specifically, I am now verifying how the "kind" array values may change.

For microscopy workflows, I usually determine the axis from the metadata e.g. CXYZ, XYZCT, and the size of each dimension. Then determine which sub-region ( slice, channels etc. ) is needed for the analisys, then perform the needed axis permutations and conversion to scalar or vector image as needed.

Some prior microscopy work with NRRD had all the dimensions were of the "domain" kind despite the order being XYZC. In this particular case each channel was being loaded independently. So the UseScalarPixel approach would appropriate.

As a generalized solution to the which axis is a multi-component issue, it may be better if the ImageIO presented the image in the order of storage ( this would simplify the consistency of meta-data and stream regions translations.) Then it could be the responsibility of the ImageFileReader to permit the axis to the ITK ImageType with the multi-component as the fastest axis. The "pixelAxisIndex" or pixelComponentAxis? could be a property of the ImageIOBase class...

lassoan:
> There is not support for streaming IO. I thought that at least reading a region of interest was supported.

Fully agree, it would be really nice. It is not related to the work being done here, though and Csaba and I already spent way too much time with this, but we may get back to it in the future.

> The existing behavior already does permutations of the data, when there is one range/"multi-component" axis ( this is only when the range dimension is not the fastest, which I don't think I have encountered)

Yes, it does, and probably for some use cases this reordering strategy works well, but for us it resulted in complex and inefficient code (due to several unnecessary voxel reorderings).

It is clear that different use cases need different reordering strategy, that's why I made it configurable.
- The default option is the same as before. I think it is a bad one, but it's OK, we can keep it for backward compatibility.
- I added our preferred option (use the first non-list-type axis as pixel component), which works best if the image contains a time series of 3D scalar or vector volumes.
- The third option leaves all data in domain axes. I added this so that the user can get the data without reordering. I think it still does some reordering (domain axes come first, followed by other axes), so it might make sense to add another option that loads everything as domain axes and let the user figure out everything (how to extract pixel data and metadata for each axis).

> The meta-data reported by the ImageIO, appears to be modified from what is in the file. Specifically, I am now verifying how the "kind" array values may change.

As far as I remember, "kinds" does not change, but as axes are reordered, the indexes are changed.

> For microscopy workflows, I usually determine the axis from the metadata e.g. CXYZ, XYZCT, and the size of each dimension. Then determine which sub-region ( slice, channels etc. ) is needed for the analisys, then perform the needed axis permutations and conversion to scalar or vector image as needed.

This sounds good. If the nrrd metadata is already set correctly, then the `T` dimension will be `list` type and the newly added strategy should work well. But if your needs are different then it could make sense to make this logic available for users in ITK by adding a new axis reordering strategy.

> Some prior microscopy work with NRRD had all the dimensions were of the "domain" kind despite the order being XYZC. In this particular case each channel was being loaded independently. So the UseScalarPixel approach would appropriate.

Yes, that's exactly the point. You can use the AxisReorder member to choose how you want to get this array.

> As a generalized solution to the which axis is a multi-component issue, it may be better if the ImageIO presented the image in the order of storage ( this would simplify the consistency of meta-data and stream regions translations.) Then it could be the responsibility of the ImageFileReader to permit the axis to the ITK ImageType with the multi-component as the fastest axis. The "pixelAxisIndex" or pixelComponentAxis? could be a property of the ImageIOBase class...

I agree, it could be an additional strategy to override the axis kinds by manually specifying `PixelAxisIndex` and `ListAxisIndex` values.

It should be easy to add more strategies, because:
1. AxisReorder is an enum
2. All the axis determinations are done in a single function, `GetAxisOrderForFileReading` ,so when you add a new strategy then you may only need to make changes there.

Modules/IO/NRRD/test/itkNrrd5dVectorImageReadWriteTest.cxx
lassoan: Yes, sure, but would then be 6 lines instead of 1 - making the test a little bit harder to read.

lassoan:
I agree that it would be nice to generalize some of what we have implemented for NRRD and I would support you as much as I can with test data, use cases, reviews, etc. Since this generalization work is an internal redesign with minimal public API change (e.g., the axis reordering strategy may be moved to one level higher), it does not have to interfere with the integration of this PR. We cannot afford a long delay with getting this feature in Slicer and we don't want to put you under time pressure, so we would prefer proceeding with merging of this PR.

> NrrdImageIO only supports one non-domain kind, and treats "list" as a multi-component ( this may have been close to Csaba initial change?)

We haven't limited the number of additional axes to in the NrrdImageIO, but since we only need up to 5D now, we did not include >5D files in the tests and we only claim that up to 5D is supported. This potential future improvement should not prevent merging of this PR.

> NrrdImageIO does not report in the ITK metadata what kinds are in the file, it removes some dimensions. The reported metadata should be what is in the file.

All axis information were supposed to be preserved (the pixel component axis does not have an axis number, so "pixel" is used as index), but `NRRD_kinds[pixel]` was indeed was missing. I've added it now.

lassoan:
We would need this feature for SlicerHeart project, which funded its development. We can wait days but not weeks. If ITK upstream did not take it then we would merge it into Slicer's fork, which would be a missed opportunity for ITK and a maintenance headache for the Slicer fork.

@blowekamp let us know how you suggest to proceed. Thanks!

lassoan:
Thank you all for the reviews and getting this merged. I will be happy to help if we get to generalize this to other file types.

PR #5477, opened 2025-07-31
ENH: Modernize Exception Messages and  OutputWindow API 

blowekamp:
This modernization of ITK's logging and exception handling system provides optional context information and improved performance while maintaining backward compatibility.

Key Changes:
1. Enhanced OutputWindow API with Context-Aware Logging

- Add new context-aware versions of OutputWindow display functions that capture file, line, class name, and object address
- Enable customizable formatting of context information in OutputWindow subclasses
- Update itkDebugMacro, itkWarningMacro, and itkGenericOutputMacro  to use new context-aware API
- Maintain backward compatibility through conditional compilation (ITK_FUTURE_LEGACY_REMOVE)

2. Improved Exception Infrastructure

- Add new ExceptionObject constructor that accepts class name and object pointer for enhanced context
- Introduce itkExceptionStringMacro for direct string exceptions without ostringstream overhead
- Update specialized exception macros to eliminate unnecessary ostringstream usage for better performance
- Maintain exception formatting to include class name and object address when available

Many logging frameworks include their own formatting of messages and errors. Provide the message text separate from the message level and context to allow application to better customize output.

## PR Checklist
- [ ] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/main/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [ ] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/main/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [ ] Added test (or behavior not changed)
- [ ] Updated API documentation (or API not changed)
- [ ] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/main/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [ ] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [ ] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples) for all new major features (if any)

Refer to the [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) for
further development details if necessary.

PR #5480, opened 2025-08-15
PERF: Replace itkExceptionMacro("...") w/ itkExceptionStringMacro("...")

N-Dekker:
Replaced any `itkExceptionMacro` call with `itkExceptionStringMacro`, whenever its argument is just a literal string (possibly taking multiple lines of code).

    Find what: itkExceptionMacro(\([ \r\n]*"[^<;]*"\);)
    Replace with: itkExceptionStringMacro$1
    Directory: D:\ITK\Modules
    [v] Match case
    (*) Regular expression [v] . matches newline

- Follow-up to pull request #5477 commit 0907661ad4e70a52f9d4a019f5f858d3e485a92e "PERF: Use string macro for exceptions", Bradley Lowekamp (@blowekamp), Aug 11 2025.

### DIFF
--- a/Modules/Video/IO/src/itkFileListVideoIO.cxx
+++ b/Modules/Video/IO/src/itkFileListVideoIO.cxx
@@ -394,8 +394,8 @@ FileListVideoIO::OpenReader()
   }
   else
   {
-    itkExceptionMacro("FileListVideoIO doesn't support reading from unknown "
-                      "ReadType");
+    itkExceptionStringMacro("FileListVideoIO doesn't support reading from unknown "
+                            "ReadType");
   }
 }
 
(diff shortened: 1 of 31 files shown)


Modules/Video/IO/src/itkFileListVideoIO.cxx
```diff
@@ -394,8 +394,8 @@ FileListVideoIO::OpenReader()
   }
   else
   {
-    itkExceptionMacro("FileListVideoIO doesn't support reading from unknown "
-                      "ReadType");
+    itkExceptionStringMacro("FileListVideoIO doesn't support reading from unknown "
+                            "ReadType");
```
dzenanz, line 398: Could this be put on the previous line?

blowekamp: It likely got split by style formatting at some point.

For an ambitious person it may be worth automatically joining all these spit string into a single line. Then letting the formatting automatically split them base of current formatting policies.

N-Dekker: Yes it could. No problem to clang-format, it seems. Thanks @dzenanz 

However, I would suggest to have such literal concatenations in a separate PR, once this one is processed. I suspect that there are quite a few more cases like that.

dzenanz: OK, sounds good.

PR #5492, opened 2025-08-21
COMP: Add extra check for itk::FixedArray wrapping

darrent1974:
The FixedArray wrapping for BSplineInterpolationWeightFunction computes dimensions for wrapping templates, in some cases the computed dimension corresponds to a dimension where a wrapping template has already been generated. A check has been added to detect and prevent duplicated templates from being generated.

## PR Checklist
- [] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/main/Utilities/KWStyle/ITKHeader.h) to new files (if any)


Wrapping/WrapITKTypes.cmake
```diff
@@ -158,8 +158,17 @@ foreach(d ${ITK_WRAP_IMAGE_DIMS})
...
   endforeach()
-  add_template("${ITKM_D}${comp}" "${ITKT_D},${comp}")
-  add_template("${ITKM_UL}${comp}" "${ITKT_UL},${comp}")
+
+  # Check if computed comp dimension has already wrapped above
+  # by searching for it in ITK_WRAP_IMAGE_DIMS.
+  # If not, wrap it.
+  list(FIND ITK_WRAP_IMAGE_DIMS ${comp} INDEX)
+
+  if(INDEX LESS 0)
```
N-Dekker: Can you please use a more specific variable name, rather than just `INDEX`? If I understand correctly, the variable is "global" during the CMake run, so a short name like `INDEX` may cause naming conflicts, sooner or later. Right? 

Maybe something like `ITK_FIXED_ARRAY_COMP_INDEX`? OK?

darrent1974: Yeah, no problem

darrent1974:
No, not in compilation. However, for us, the warnings are output in python when ITK libs are loading. They've been annoying us for quite some time in our production environment. I believe you will only bump into this issue if you are wrapping images with dimensions of a multiple of 4. We happen to wrap dim 4 images in our build of ITK.

SimonRit (approved):
Looks good to me. I have had many similar wrapping warnings in the past and they are quite annoying. One alternative implementation to tackle this kind of issues which you might consider is to use the [remove macro](https://github.com/InsightSoftwareConsortium/ITK/blob/main/Wrapping/CMakeUtilityFunctions.cmake).

dzenanz (approved):
LGTM

N-Dekker (approved):
Thanks @darrent1974 !

By the way, I would suggest to squash the two commits before merging.

I like the suggestion of @SimonRit to use a helper macro, but that's no show-stopper to me, for now. Approved 👍

Issue #5326, opened 2025-04-25
Proposal to Deprecate macOS Intel Support in ITK 6 (Retain in ITK 5)


thewtex:
### **Proposal**

Deprecate macOS Intel (x86_64) support in ITK 6, while maintaining it in ITK 5 as a long-term support (LTS) release. This aligns with Apple’s phased sunsetting of Intel Macs and industry-wide shifts toward Apple Silicon optimization.

---

### **Rationale**

**1. Apple’s Official Transition Timeline**

- Apple completed its transition to Apple Silicon in June 2023 with the Intel Mac Pro’s discontinuation[^4].
- macOS 17 (2026) is projected to be Apple Silicon–only, ending major OS updates for Intel Macs[^2][^3].
- Security updates for Intel Macs will continue until ~2027 for newer models (e.g., 2019 iMac, Mac Pro) but lack feature parity[^2][^5].

**2. Third-Party Software Trends**

- Major developers (e.g., Adobe, Microsoft) now prioritize Apple Silicon optimizations, with some dropping Intel support entirely[^2][^5].
- ARM64 macOS now dominates developer toolchains, with Xcode and CI/CD pipelines increasingly excluding Intel[^5].

**3. ITK Maintenance Burden**

- Supporting legacy Intel macOS complicates builds, testing, and CI/CD for ARM64-specific optimizations.
- macOS Intel GitHub Action Runners (`macos-13`) are slower than the ARM counterparts.
- Deprecating Intel in ITK 6 allows focused resources on Apple Silicon performance.

---

### **Proposed Implementation**

- **ITK 6**:
    - Remove Intel macOS from build matrices.
    - Document deprecation in release notes, directing users to ITK 5 for legacy workflows.
- **ITK 5**:
    - Maintain Intel macOS support as an LTS branch until 2027 (aligned with Apple’s security update window).
    - Backport critical fixes only.

---

### **Community Impact Considerations**

- **Affected Users**: Researchers/clinicians relying on older Intel Macs for specialized hardware (e.g., GPUs, PCIe devices).
- **Mitigation**: Provide migration guides for Apple Silicon or Linux/Windows alternatives.
- **Timeline**: Target deprecation for ITK 6.0 (Q1 2026), coinciding with macOS 17’s expected release[^3].

---

### **Supporting Data**

| Key Fact | Source |
| :-- | :-- |
| macOS 17 (2026) likely Apple Silicon–only | [^2][^3] |
[...]

PR #5497, opened 2025-08-27
Use GHA hosted mac-15 to cross build for intel

blowekamp:
Relates to  #5326 and ##5369

## PR Checklist

### DIFF
--- a/.github/workflows/macos-arm.yml
+++ b/.github/workflows/macos-arm.yml
@@ -29,65 +29,97 @@ concurrency:
 
 env:
     ExternalDataVersion: 5.4.0
+    CMakeVersion: "4.0.1"
 
 jobs:
-    macOS-Python:
-        runs-on: macos-14
-        timeout-minutes: 0
-        steps:
-            - name: Checkout
-              uses: actions/checkout@v4
-              with:
-                fetch-depth: 5
-                clean: true
+  build:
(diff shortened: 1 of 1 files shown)


dzenanz (approved):
LGTM, but it would be good if someone else also reviews this.

blowekamp:
This is a WIP. To test the configuration without extra resources, the PR temporarily removed the OSX ITK python build. For a final solution it needs to be determined how best to allocate the free GH runners that we have.

thewtex:
@blowekamp thanks for the investigation and efforts.

If we keep the existing macos-15 arm build, and add an macos-15 intel build, maybe it will not bee too slow?

blowekamp:
> @blowekamp thanks for the investigation and efforts.
> 
> If we keep the existing macos-15 arm build, and add an macos-15 intel build, c++ only, maybe it will not be too slow?

Done. Refactored to use a matrix build following the parameterization done with some SimpleITK GHA build, please let me know of another exemplar GHA configuration should be used.

blowekamp:
The workflow.name added is "macos-x86_64-rosetta" but this is not used in the CDash name or site. With the system arch being clear on the dashboard, but not this target arch it will not be clear why this build is different. They may be a need to update a script in the dashboard branch to use this varible.

PR #4392, opened 2024-01-12
COMP: Skip Python-Wrapping of ITKVtkGlue

fabian-philips:
Python wrapping of VTKGlue Bridge crashed during compilation and native bridge exists. Discussed here https://discourse.itk.org/t/itk-5-3-0-5-4rc-build-errors-shared-libs-python-wrappings-itkvtkglue-vs-2022-win11.

## PR Checklist
- [x] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [x] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/master/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [x] Added test (or behavior not changed)
- [x] Updated API documentation (or API not changed)

Refer to the [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) for
further development details if necessary.

PR #5502, opened 2025-09-01
Bring back itkVtkGlue to Python wrapping

AlexyPellegrini:
## Brief

Python wrapping of itkVtkGlue has been removed by https://github.com/InsightSoftwareConsortium/ITK/pull/4392. This PR fixes the build issue and enables it again.

## Issue

The XML generations failed because the response file referencing the include directories for castxml was not generated correctly. All modules used the same `.castxml.inc` instead of their own `<modulename>.castxml.inc`. Once this is fixed, the module wraps correctly.

## Why is it needed

Even though [numpy enables to share an itk::Image and a vtkImageData](https://docs.itk.org/en/latest/learn/python_quick_start.html#itk-and-vtk), it does not support pipelining. Pipelining is important because some VTK and ITK filters support requesting only a subextent of a resource, enabling better performances especially for readers of formats meant to be streamed, such as OME-Zarr.

```py
image = itk.imread(sys.argv[1], imageio=itk.OMEZarrNGFFImageIO.New())
itktovtk = itk.ImageToVTKImageFilter[type(image)].New()
itktovtk.SetInput(image)
reslice = vtkImageReslice() # vtkImageReslice may query a subextent
reslice.SetInputConnection(itktovtk.GetImporter().GetOutputPort())
#...
```

## PR Checklist

---

FYI @finetjul

### DIFF
--- a/Modules/Bridge/VtkGlue/itk-module-init.cmake
+++ b/Modules/Bridge/VtkGlue/itk-module-init.cmake
@@ -60,6 +60,8 @@ set(
   _required_vtk_libraries
   ${_target_prefix}IOImage
   ${_target_prefix}ImagingSources
+  ${_target_prefix}kwiml
+  ${_target_prefix}vtksys
 )
 if(ITK_WRAP_PYTHON)
   list(
--- a/Modules/Bridge/VtkGlue/wrapping/CMakeLists.txt
+++ b/Modules/Bridge/VtkGlue/wrapping/CMakeLists.txt
@@ -1,45 +1,36 @@
 # Forward include directories to CastXML
-if(NOT VTK_RENDERING_BACKEND STREQUAL "None")
-  foreach(_vtk_lib IN LISTS ITKVtkGlue_VTK_LIBRARIES)
-    get_target_property(
-      _vtk_lib_include_dirs
-      ${_vtk_lib}
-      INTERFACE_INCLUDE_DIRECTORIES
-    )
-    if(_vtk_lib_include_dirs)
-      include_directories(${_vtk_lib_include_dirs})
-    endif()
-  endforeach()
-endif()
+foreach(_vtk_lib IN LISTS ITKVtkGlue_VTK_LIBRARIES)
+  get_target_property(
+    _vtk_lib_include_dirs
+    ${_vtk_lib}
+    INTERFACE_INCLUDE_DIRECTORIES
+  )
+  if(_vtk_lib_include_dirs)
+    include_directories(${_vtk_lib_include_dirs})
+  endif()
+endforeach()
 
-# Python provides native conversion between ITK and VTK
-if(ITK_WRAP_PYTHON)
+itk_wrap_module(ITKVtkGlue)
+if("${VTK_VERSION}" VERSION_LESS 7.0.0)
   message(
-    STATUS
-    "ITK_WRAP_PYTHON is turned on - skipping ITKVtkGlue, since native version is available https://docs.itk.org/en/latest/learn/python_quick_start.html#itk-and-vtk."
+    WARNING
+    "The ITKVtkGlue module can only built with Python 3 and VTK >= 7."
   )
 else()
-  itk_wrap_module(ITKVtkGlue)
-  if("${VTK_VERSION}" VERSION_LESS 7.0.0)
-    message(
-      WARNING
-      "The ITKVtkGlue module can only built with Python 3 and VTK >= 7.\n"
-      "Only the c++ version of the module will be built."
-    )
-  else()
-    list(
-      APPEND
-      WRAPPER_SWIG_LIBRARY_FILES
-      "${CMAKE_CURRENT_SOURCE_DIR}/VtkGlue.i"
-    )
-    set(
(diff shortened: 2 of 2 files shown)


dzenanz:
If that is a lot of work, maybe wait for a review by Matt.

Wrapping/macro_files/itk_auto_load_submodules.cmake
```diff
@@ -114,7 +114,7 @@ function(generate_castxml_commandline_flags)
   endforeach()
 
   #Write compile definitions and include paths to file.  @CONFIG_CASTXML_INC_CONTENTS@ expanded in configure_file
-  set(castxml_inc_file "${WRAPPER_LIBRARY_OUTPUT_DIR}/castxml_inputs/${library_name}.castxml.inc")
+  set(castxml_inc_file "${WRAPPER_LIBRARY_OUTPUT_DIR}/castxml_inputs/${WRAPPER_LIBRARY_NAME}.castxml.inc")
```
thewtex: Can a further explanation be provided as to why this is changed and why the change does not impact other libraries and modules?

AlexyPellegrini: ofc,

- Why this is changed?
https://github.com/InsightSoftwareConsortium/ITK/pull/3200 reworked the Python wrapping code, and is the reason why itkVtkGlue couldn't be wrapped anymore.  
If you take a look [at this line](https://github.com/InsightSoftwareConsortium/ITK/pull/3200/files#diff-f27632aa929473319cb26fb10a755a615ed7dbb6ed8dd8ee9254f48013812b97L129) you will see that the original code was in the `itk_wrap_module` macro directly, so it had access to `library_name` parameter. When the code has been moved to a different macro `library_name` became empty (because macro arguments are not real variables, so they do not outlive the macro body). 
Using `WRAPPER_LIBRARY_NAME` **actually reverts back to the original behavior**, where each module has its own include directories list, named `<module-name>.castxml.inc`.

- Why the change does not impact other libraries and modules?
Each module will have its own file, otherwise no behavior changed!

Note that this also fixes a subtle issue about incremental builds, because enabling or disabling a module with python wrapping caused all module wrapping to be rebuilt, because the `.castxml.inc` was modified, so all xml regenerated, and so on. With this change, the old behavior is restored, so already build modules shouldn't modify their own `castxml.inc`, so they won't be rebuilt for nothing!

Let me know if something is still unclear!

AlexyPellegrini:
> If this is patch set is meant to be included in the release branch, make `release` branch the target. Otherwise please base in on a newer commit to avoid merge conflicts.

Indeed, I rebased on main!

Wrapping/macro_files/itk_auto_load_submodules.cmake
dzenanz: The information here, especially about incremental builds, deserves to be in the commit message.

AlexyPellegrini: I updated the commit description!

dzenanz (approved):
LGTM

hjmjohnson (approved):

dzenanz:
Thank you Alexy!

dzenanz:
The point of the limited API is compatibility of later versions with 3.11. We could make options `ITK_USE_PYTHON_LIMITED_API` and `Module_ITKVtkGlue` mutually exclusive. Can you make that PR?

dzenanz:
There are more errors, such as [this](https://open.cdash.org/viewBuildError.php?buildid=10664951): `LINK : fatal error LNK1181: cannot open input file 'vtkkwiml.lib' [T:\Dashboard\ITK-build\Modules\Bridge\VtkGlue\ITKVtkGlueHeaderTest1.vcxproj]`. It also affects VTK glue test driver:
```log
Build started at 13:46...
1>------ Build started: Project: ITKVtkGlueTestDriver, Configuration: Debug x64 ------
1>LINK : fatal error LNK1104: cannot open file 'vtkkwiml.lib'
1>Done building project "ITKVtkGlueTestDriver.vcxproj" -- FAILED.
========== Build: 0 succeeded, 1 failed, 62 up-to-date, 0 skipped ==========
========== Build completed at 13:46 and took 01.438 seconds ==========
```
but not `ITKVtkGlue` target. Indeed, there is no `vtkkwiml.lib` anywhere on my computer.

Modules/Bridge/VtkGlue/itk-module-init.cmake
```diff
@@ -60,6 +60,8 @@ set(
   _required_vtk_libraries
   ${_target_prefix}IOImage
   ${_target_prefix}ImagingSources
+  ${_target_prefix}kwiml
```
dzenanz, line 63: We might need special handling for this, like we do for `vtksys`. But critical version might be different from 8.90.0. @AlexyPellegrini

dzenanz: The critical version is probably much newer, as [5e87806](https://gitlab.kitware.com/vtk/vtk/-/commit/5e878068c23f6b87a398dbe8000424971dcb7afa), which is affected, is only 6 months old.

PR #5504, opened 2025-09-01
COMP: Avoid PySequence_Fast_GET_ITEM - ITK 5

thewtex:
No longer available in Python 3.14.

Use PySequence_GetItem. This returns a reference. In the use cases
(image dimensions), the performance impact is negligible.


dzenanz (approved):
Looks good on a glance.

PR #5509, opened 2025-09-04
COMP: Update CastXML binaries to 2025.09.03

thewtex:
Better compiler support.

Use filenames consistent with the CastXMLSuperbuild build artifacts.

Use SHA256 hashes, which are offered by the GitHub Releases interface.

On Linux and Macos, these binaries require newer glibc and Macos
versions. These are to support updated GitHub Actions and manylinux
build environments.


dzenanz (approved):
Looks good on a glance.

hjmjohnson (approved):
Thanks to Matt and Brad for working on CastXML and keeping it up to date.

PR #5513, opened 2025-09-05
STYLE: Remove accidentally duplicated words from comments

N-Dekker:
Found by the regular expression ` (\w\w+) \1 `. Avoided removing intended duplicates, like "long long", "really really", and "very very". Skipped comments that may need other adjustments than just removing a duplicate word.

### DIFF
--- a/Modules/Filtering/ImageGrid/include/itkResampleImageFilter.h
+++ b/Modules/Filtering/ImageGrid/include/itkResampleImageFilter.h
@@ -220,7 +220,7 @@ public:
   itkGetConstReferenceMacro(Size, SizeType);
   /** @ITKEndGrouping */
   /** Get/Set the pixel value when a transformed pixel is outside of the
-   * image.  The default default pixel value is 0. */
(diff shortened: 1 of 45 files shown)


dzenanz (approved):

thewtex (approved):
💯

Modules/Filtering/ImageGrid/include/itkResampleImageFilter.h
```diff
@@ -220,7 +220,7 @@ class ITK_TEMPLATE_EXPORT ResampleImageFilter : public ImageToImageFilter<TInput
   itkGetConstReferenceMacro(Size, SizeType);
   /** @ITKEndGrouping */
   /** Get/Set the pixel value when a transformed pixel is outside of the
-   * image.  The default default pixel value is 0. */
+   * image.  The default pixel value is 0. */
```
N-Dekker, line 223: Maybe this was intended: the default of `DefaultPixelValue` is zero. So maybe that's the _default default_ pixel value 🤷

PR #5517, opened 2025-09-08
COMP: Disable limited Python API for itkVtkGlue module wrapping

AlexyPellegrini:
This is my proposition to fix the build issue introduced by #5502.
In short, it disables Python stable "limited" API when wrapping itkVtkGlue module wrapping.
Other modules are not affected by this change.

WDYT @dzenanz @hjmjohnson @thewtex 

## PR Checklist

### DIFF
--- a/Modules/Bridge/VtkGlue/wrapping/CMakeLists.txt
+++ b/Modules/Bridge/VtkGlue/wrapping/CMakeLists.txt
@@ -16,6 +16,12 @@ if("${VTK_VERSION}" VERSION_LESS 7.0.0)
     WARNING
     "The ITKVtkGlue module can only built with Python 3 and VTK >= 7."
   )
+elseif(ITK_USE_PYTHON_LIMITED_API)
+  message(
+    FATAL_ERROR
+    "The ITKVtkGlue module can only built without Python limited API due to VTK limitations."
+    "Please set `ITK_USE_PYTHON_LIMITED_API` to `FALSE`."
+  )
 else()
   list(
     APPEND
--- a/Wrapping/macro_files/itk_end_wrap_module.cmake
+++ b/Wrapping/macro_files/itk_end_wrap_module.cmake
@@ -489,7 +489,6 @@ ${DO_NOT_WAIT_FOR_THREADS_CALLS}
     ${use_python_limited_api_default}
     CACHE BOOL
     "Use Python's limited API for Python minor version compatibility."
-    FORCE
   )
   mark_as_advanced(ITK_USE_PYTHON_LIMITED_API)
   unset(use_python_limited_api_default)


dzenanz (approved):
Looks good on a glance.

Modules/Bridge/VtkGlue/wrapping/CMakeLists.txt
```diff
@@ -32,5 +32,6 @@ else()
     ${ITK_LIBRARIES}
     ${ITKVtkGlue_VTK_LIBRARIES}
   )
+  set(ITK_USE_PYTHON_LIMITED_API FALSE)
```
thewtex: It should not silently override. Instead, send a error communicating the incompatible state and suggest a workaround: the person building explicitly sets ITK_USE_PYTHON_LIMITED_API to FALSE.

Wrapping/CMakeLists.txt
```diff
@@ -15,7 +15,6 @@
 #   limitations under the License.
 #
 #==========================================================================*/
-cmake_minimum_required(VERSION 3.16.3...3.19.7 FATAL_ERROR)
```
thewtex: Why is this removed?

dzenanz: I thought commit message explained it.

thewtex: Ah, I see the it in the commit message now -- I looked in the PR description.

dzenanz: I also had the same question ... then I saw the answer in the commit message 😄

Modules/Bridge/VtkGlue/wrapping/CMakeLists.txt
AlexyPellegrini: If we do this, does it mean VtkGlue will not be available in ITK packages uploaded to pypi?
Otherwise, this is possible, we just need to fix the definition of ITK_USE_PYTHON_LIMITED_API as it is always forced to its default value with current code. And also update the CI jobs that enable this module and request limited API.

thewtex: VtkGlue is not available in the ITK packages uploaded to pypi regardless. There were packages uploaded to PyPI via https://github.com/InsightSoftwareConsortium/ITKVtkGlue, but it is not maintained.

The setting should inform the builder that they have the option to disable the python limited ABI. This will mean that the python bindings only work with the built major.minor version instead of Python 3.11 or greater.

AlexyPellegrini: Ok, it makes sense. I will update the PR then.

thewtex (approved):
@AlexyPellegrini thanks

dzenanz (approved):

PR #5531, opened 2025-09-22
ENH: itkexpat: Add script to import from upstream 

bradking:
Re-import expat 2.7.2 from upstream using our modern history structure.
This matches what was last manually imported by #5488.


hjmjohnson:
Awesome! Thanks @bradking

bradking:
Do: check

bradking:
I've pushed an update that passes ghostflow-director's size check.  I'd added the `.gitattributes` file in the wrong commit.

PR #5533, opened 2025-09-24
ENH: Update Eigen to 3.4.1 revision

hjmjohnson:
- **ENH: Update to 3.4.1 Eigen release**
- **Eigen3 2025-09-24 (e0545642)**

## PR Checklist


hjmjohnson:
> 3.4.1 release? I don't see such a thing here: https://gitlab.com/libeigen/eigen/-/releases
> 
> Wasn't ITK using eigen master? Is this release strictly newer, or from a different branch?

@seanm I took the tagged version from 2025-09-23.  It seems to contain cherry-picked content from the master branch.

The master branch is currently undergoing rapid infrastructure changes, which involve numerous significant updates to the build environment and version management.  Some of those changes in 5.0.0 have made it quite difficult to reconcile with ITK's build customizations.

I think that 3.4.1 includes the algorithmic updates without also requiring adopting the new build and versioning paradigms.  

```
commit b66188b5dfd147265bfa9ec47595ca0db72d21f5 (tag: 3.4.1, origin/3.4, hj/3.4)
Author: Antonio Sanchez <cantonios@google.com>
Date:   2025-09-23 15:22:29 -0500
```

Once the 5.0.0 infrastructure becomes more stable, it will make more sense to put in the effort to coordinate with ITK.

dzenanz:
Didn't we last update Eigen in July? So updating to 3.4.1 seems like a step backwards. @phcerdan can you comment about the effort required to update to 5.0?

phcerdan:
image

Not sure if this is a good strategy @hjmjohnson. We are following eigen master because the lack of releases of eigen. They have tagged 3.4.1, but doesn't seem like a full release yet.

> The master branch is currently undergoing rapid infrastructure changes, which involve numerous significant updates to the build environment and version management. Some of those changes in 5.0.0 have made it quite difficult to reconcile with ITK's build customizations.

Don't fully get this, can you provide some examples? In ITK we are basically using the header only version of Eigen. If you want extra components, or compile extra things, or have full control on the eigen version you want to include, you would need to go and use a external eigen (fully controlled bu you). This is also highly recommended if you need to share eigen with other components, even though you can do it with the ITK internal version too.

I think depending on extremely scarce releases (the last release 3.4.0 was 2021) isn't working for the "internal" ITK eigen. @seanm has been solving warnings directly upstream, and we just update our head to theirs. If you need a specific Eigen for your case, just go external eigen.

Does that make sense for your case?

PR #5538, opened 2025-09-30
Use `std::array` instead of `std::vector` for local variables

N-Dekker:
When an std::vector is initialized by a compile-time constant, specifying its size, and the std::vector isn't resized afterwards, it appears clearer to use std::array instead. (And it may produce faster code, of course, when using std::array.) 

Cases found by the following regular expression (using Visual Studio, Find in Files...):

     std::vector<[^=]+>[^=]+\([1-9][0-9]*\);


thewtex (approved):
🚀

PR #5554, opened 2025-10-14
COMP: Speedup CI free-disk-space stage

hjmjohnson:
Freeing disk space for local builds sometimes takes
more than 10 minutes to complete.  Merged several
improvements from various forks and PRs related to the
long-abandoned jlumbroso/free-disk-space action.

The BRAINSia/free-disk-space GitHub action variant is about 20% 
faster when deleting all the packages.  The ITK build requires deleting almost all the packages
to build ITK.  Analysis of how much space is needed to build
ITK is included inline in the comments, along with approximate
space-saving estimates for each selectable file to be removed.

If all the .o and .a files are removed after building, and before testing, the
amount of required disk space needed for successful CI runs is reduced by
about 7GB.

This effort reduced the time of the disk cleaning stage from 10+ minutes to 3.5 minutes.

### DIFF
--- a/.github/workflows/pixi.yml
+++ b/.github/workflows/pixi.yml
@@ -44,9 +44,40 @@ jobs:
             fetch-depth: 5
             clean: true
 
+        - name: Disk space reporting BEFORE (optimize free-disk-space)
+          run: |
+            echo "****** df -h ${{ runner.temp }}"
+            df -h "${{ runner.temp }}"
+            echo "****** df -h"
+            df -h
+            echo "****** df -hi /"
+            df -h /
+
         - name: Free Disk Space (Ubuntu)
           if: matrix.os == 'ubuntu-22.04'
-          uses: jlumbroso/free-disk-space@main
+          uses: BRAINSia/free-disk-space@v2
+          with:
+            removalmode: "rmz"
+
+            swap-storage:   "true"
+            haskell:        "true"
+            dotnet:         "true"
+            docker-images:  "false" # Takes too long
+            tool-cache:     "true"
+            android:        "false" # Takes too long
+            large-packages: "true"  # Takes too long to remove apt-get packages
+            mandb:          "true"  # Speeds up future apt-get installs (disables man page generation), this CI does not use apt-get
+            # For ITK run with find
+            # TIME_REPORT: =============================================================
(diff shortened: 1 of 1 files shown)


blowekamp:
Thank you for working in this. I have some related thoughts to share.

On some build for SimpleITK which run out of disk space, I delete all the *.o files before running the tests.

 I also wonder if there are CMake options to help clean up disk spaces after a test or a test group. For some other recent testing I have done, I use a temporary directory of the large image output, which gets erased if the tests pass.

.github/workflows/pixi.yml
```diff
@@ -44,9 +44,54 @@ jobs:
...
+            #
+            # POST disk cleanup
+            # Filesystem      Size  Used Avail Use% Mounted on
+            # /dev/root        73G   23G   51G  31% /         <--- cleanup (51-18) 33GB of memory
+            #
+            # AFTER running all tests
+            # ****** df -h /home/runner/work/_temp
+            # Filesystem      Size  Used Avail Use% Mounted on
+            # /dev/root        73G   65G  8.1G  89% /         <--- ITK Used 42.9GB to build and run tests
+            #                                                 Need to cleanup at least (43-18)=25GB of free space
```
blowekamp: I like the reporting of dispose usage added. I am not sure there is a need for this lengthy report here?

hjmjohnson:
> How much did this speed up free space reclaiming compared to the old way?

It looks like about 8 minutes instead of 10 minutes.  I like @blowekamp idea of removing .o files before running the test, which could then allow for significant speedup by clearing less space to start with.  Let me give that a try.

hjmjohnson:
@thewtex @dzenanz @blowekamp Thanks for the comment. I have now been able to reduce the disk cleaning phase from 10 minutes to 3.5minutes.  I also found that the pixie stages were not independent.

    ```bash
    pixi run configure # --> runs configure
    pixi run build         # --> runs configure, runs build
    pixi run test           # --> runs configure, runs build, runs test
    ```
    
    ```bash
    pixi run configure         # --> runs configure
    pixi run build_noconfigure # --> runs build
    pixi run test_nobuild      # --> runs test
    ``` 

Which resulted in the configure phase running 3 times (45 seconds for 2nd and 3rd time), and the build phase being run twice (probably only a few seconds in the 2nd and 3rd run).

dzenanz (approved):

PR #5555, opened 2025-10-15
ENH: Improve initialization in Path classes

blowekamp:
Update the Path and ParametricPath object to perform initialization of member variable inline, enabling the default implementation of the constructor.

## PR Checklist


Modules/Filtering/Path/include/itkPath.h
```diff
@@ -126,9 +126,9 @@ class ITK_TEMPLATE_EXPORT Path : public DataObject
   itkGetConstMacro(ZeroIndex, IndexType);
 
 private:
-  // These "constants" are initialized in the constructor
-  OffsetType m_ZeroOffset{}; // = 0 for all dimensions
-  IndexType  m_ZeroIndex{};  // = 0 for all dimensions
+  // These constants are initialized at compile time
+  const OffsetType m_ZeroOffset{}; // = 0 for all dimensions
+  const IndexType  m_ZeroIndex{};  // = 0 for all dimensions
```
N-Dekker: Even though those data members are `const`, they are still initialized in the (defaulted) constructor. If you want them to be initialized at compile time, declare them `static constexpr`.

N-Dekker (changes requested):
Please either revert to the original comment, `These "constants" are initialized in the constructor`, or (preferably, if possible) declare them `static constexpr`.

N-Dekker:
@blowekamp Thanks for your PR! Can you please make the subject line a little bit more specific? (In order to easily find it back when looking at the commit history.) For example:

    ENH: Use inline initialization and default constructors in Path

Modules/Filtering/Path/include/itkPath.h
N-Dekker: Eventually I hope we can just remove those data members, as well as those member functions, Path::GetZeroOffset() and Path::GetZeroIndex(). I think a member function that always just returns a zero-initialized Offset, or a zero-initialized Index isn't really necessary, right?  🤷 Anyway, that may be another pull request  😺

blowekamp: Let me see if I have this correct: The Index<> and Offset<> types are both POD.  So POD member variable with explicit default initialization, will not get initialized to zeros. But a static constexpr POD member variable will get initialized to zero with a compile file value?

Or are you saying that with this proposed close those variable are not initialize. Please be specific with your recommended changes.

N-Dekker: `m_ZeroOffset{}` and `m_ZeroIndex{}` are properly zero-initialized either way, no problem. But currently this initialization takes place at run-time. When they are declared `static constexpr`, the initialization will take place at compile-time.

For this commit, I recommend adding `static constexpr`, instead of `const`.

Maybe for a follow-up pull request, remove Path::GetZeroOffset(), Path::GetZeroIndex(), m_ZeroOffset, and m_ZeroIndex. But that may be beyond the scope of this pull request.

Modules/Filtering/Path/include/itkPath.hxx
```diff
@@ -36,25 +36,15 @@
...
 
 template <typename TInput, typename TOutput, unsigned int VDimension>
 void
 Path<TInput, TOutput, VDimension>::PrintSelf(std::ostream & os, Indent indent) const
 {
   this->Superclass::PrintSelf(os, indent);
 
-  os << indent << "ZeroOffset: " << m_ZeroOffset << std::endl;
-  os << indent << "ZeroIndex: " << m_ZeroIndex << std::endl;
+  os << indent << "ZeroOffset: " << GetZeroOffset() << std::endl;
```
dzenanz: Since these are no longer member variables, we don't need to print them. Good even as-is.

blowekamp: Looks like we can now remove the whole Path.hxx file.

PR #5556, opened 2025-10-16
Remove `this->GetZeroIndex()` and `this->GetZeroOffset()` calls from Filtering/Path

N-Dekker:
It appears unnecessary to call a virtual member function to just get a zero-initialized `Index` or  `Offset` object.

- Triggered by pull request #5555 by @blowekamp

### DIFF
--- a/Modules/Filtering/Path/include/itkChainCodePath.hxx
+++ b/Modules/Filtering/Path/include/itkChainCodePath.hxx
@@ -85,7 +85,6 @@ ChainCodePath<VDimension>::IncrementInput(InputType & input) const -> OffsetType
 
 template <unsigned int VDimension>
 ChainCodePath<VDimension>::ChainCodePath()
-  : m_Start(this->GetZeroIndex())
 {}
 
 template <unsigned int VDimension>
--- a/Modules/Filtering/Path/include/itkChainCodePath.h
+++ b/Modules/Filtering/Path/include/itkChainCodePath.h
@@ -148,7 +148,7 @@ public:
   void
   Initialize() override
   {
-    m_Start = this->GetZeroIndex();
+    m_Start = {};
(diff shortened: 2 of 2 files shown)


Modules/Filtering/Path/include/itkChainCodePath.hxx
```diff
@@ -85,7 +85,6 @@ ChainCodePath<VDimension>::IncrementInput(InputType & input) const -> OffsetType
 
 template <unsigned int VDimension>
 ChainCodePath<VDimension>::ChainCodePath()
-  : m_Start(this->GetZeroIndex())
```
dzenanz, line 88: I assume that `m_Start` is zero-initialized somewhere else? After this change, this constructor will not explicitly initialize anything, so should it explicitly exist?

N-Dekker: m_Start is indeed zero-initialized somewhere else: https://github.com/InsightSoftwareConsortium/ITK/blob/f46b4c93fec770370f74ab5280dc5cf8bfb2a384/Modules/Filtering/Path/include/itkChainCodePath.h#L162

And this constructor might as well be defaulted (`= default`). I did not want to do so with this pull request, because I thought maybe Bradley (@blowekamp) was about to do so already 😄

N-Dekker:
I hope this pull request can be merged before #5555, otherwise the commit messages (which state that GetZeroOffset() and GetZeroIndex() are virtual functions) must be adjusted!!!

dzenanz:
I have no objections to that 😄

N-Dekker:
> I have no objections to that 😄

I see, #5555 is still recovering from build errors, so #5556 is now ahead of #5555, in the CI process, fortunately  😸

PR #5562, opened 2025-10-17
STYLE: Increase code coverage, make intent clear

hjmjohnson:
Improve code coverage by refactoring termination path order.

Improve code intent with explicit variable for determining if a
a cluster is found.

## PR Checklist

### DIFF
--- a/Modules/Segmentation/SuperPixel/include/itkSLICImageFilter.hxx
+++ b/Modules/Segmentation/SuperPixel/include/itkSLICImageFilter.hxx
@@ -454,44 +454,45 @@ SLICImageFilter<TInputImage, TOutputImage, TDistancePixel>::ThreadedConnectivity
     idx[d] = Math::RoundHalfIntegerUp<IndexValueType>(cluster[numberOfComponents + d]);
   }
 
-
-  if (outputImage->GetPixel(idx) != clusterIndex)
+  bool clusterFound = outputImage->GetPixel(idx) == clusterIndex;
+  if (!clusterFound) // Search for cluster in in SearchLabels
   {
     itkDebugMacro("Searching for cluster: " << clusterIndex << " near idx: " << idx);
-
     searchLabelIt.SetLocation(idx);
-    size_t n = 0;
-    for (; n < searchLabelIt.Size(); ++n)
+    const size_t numSearchLabels = searchLabelIt.Size();
+    size_t       n = 0;
+    while (!clusterFound && n < numSearchLabels)
     {
       if (searchLabelIt.GetPixel(n) == clusterIndex)
       {
         idx = searchLabelIt.GetIndex(n);
-
+        clusterFound = true;
         itkDebugMacro("Non-Center does  match Id. @: " << idx << " for: " << clusterIndex);
-        break;
+      }
+      ++n;
+    }
+  }
+
+  if (clusterFound)
+  {
+    this->RelabelConnectedRegion(idx, clusterIndex, clusterIndex, indexStack);
+
+    if (indexStack.size() < minSuperSize)
+    {
+      // std::cout << "\tLabel is too small: " << indexStack.size() << std::endl;
+      // The connected Superpixel is too small, so demark the marker image
+      for (size_t indexStackDelabel = 0; indexStackDelabel < indexStack.size(); ++indexStackDelabel)
+      {
+        m_MarkerImage->SetPixel(indexStack[indexStackDelabel], 0);
       }
     }
-
-    if (n >= searchLabelIt.Size())
-    {
+  }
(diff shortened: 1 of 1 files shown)


hjmjohnson:
The primary purpose for this change is self-serving :).

I am using ITK as a talking point in my teaching of how
code coverage can help drive good software development
process, and I noticed that no module had 100% code coverage.

I then noticed that the code coverage exposed code that
existed in an if statement that was impossible to be true
(i.e. dead code).

So I am practicing what I preach to students: Employ
the Boy Scout principle that "encourages leaving a place
better than you found it".

So ... my good deed for the day :).

Modules/Segmentation/SuperPixel/include/itkSLICImageFilter.hxx
```diff
@@ -460,25 +460,16 @@ SLICImageFilter<TInputImage, TOutputImage, TDistancePixel>::ThreadedConnectivity
...
       if (searchLabelIt.GetPixel(n) == clusterIndex)
       {
         idx = searchLabelIt.GetIndex(n);
-
         itkDebugMacro("Non-Center does  match Id. @: " << idx << " for: " << clusterIndex);
         break;
       }
     }
-
-    if (n >= searchLabelIt.Size())
```
SimonRit, line 475: I don't see why this unreachable, you can reach n==searchLabelIt.Size() is the condition line 466 is never true, can't you? Or do you know this never happens? In the latter case, I would suggest adding a comment for other developers looking at this code. And I think a while loop would be more appropriate then rather than a for loop with a break.

dzenanz: I agree. If cluster is not found, n will be == searchLabelIt.Size()

blowekamp: As I recall this is the intent. It is possible that the cluster could be a donut or a C, and the region is missed with this size of the a neighborhood iterator.  However, not being able to create an example of this case does not mean it does not exist.

Modules/Filtering/ImageLabel/include/itkBinaryContourImageFilter.hxx
```diff
@@ -48,11 +48,10 @@ BinaryContourImageFilter<TInputImage, TOutputImage>::GenerateInputRequestedRegio
 
   // We need all the input.
   const InputImagePointer input = const_cast<InputImageType *>(this->GetInput());
-  if (!input)
+  if (input)
```
blowekamp, line 51: The original code pattern is used in many place in ITK where the "fail early" principle is followed for this and similar methods, where additional code exists after the check. 

These types of checks are not used is proper ITK code, however improper dynamic_casts, and assigning of inputs with base class methods they can cause an unexpected type in the pipelines inputs or outputs. The types of the inputs and output are not fully type safe and run time checks needed. I would not classify this as an "impossible" case.

This change does improve coverage,  so its good.

Modules/Segmentation/SuperPixel/include/itkSLICImageFilter.hxx
hjmjohnson: THANK YOU FOR SEEING THIS!! That is my mistake.  Let me pay closer attention and try to resolve this with a while loop that also increases code coverage.

hjmjohnson: I refactored the code to make it more transparent in identifying a cluster to process.

SimonRit (approved):
LGTM

Modules/Segmentation/SuperPixel/include/itkSLICImageFilter.hxx
```diff
@@ -454,44 +454,42 @@ SLICImageFilter<TInputImage, TOutputImage, TDistancePixel>::ThreadedConnectivity
...
-      m_MarkerImage->SetPixel(indexStack[indexStackDelabel], 0);
+      // std::cout << "\tLabel is too small: " << indexStack.size() << std::endl;
+      // The connected Superpixel is too small, so demark the marker image
+      for (size_t indexStackDelabel = 0; indexStackDelabel < indexStack.size(); ++indexStackDelabel)
+      {
+        m_MarkerImage->SetPixel(indexStack[indexStackDelabel], 0);
+      }
     }
   }
+#if defined(DEBUG)
```
dzenanz, line 490: shouldn't this be inside `else(clusterFound)` block?

hjmjohnson: Oh, yes!  I did not test in debug mode.

PR #5566, opened 2025-10-22
COMP: Improve code consistency with constexpr

hjmjohnson:
Prefer `static constexpr` over `constexpr static` because
it is more common and considered slightly clearer.

Since C++11, the brace form {} is preferred in most cases because it provides
safer, more consistent, and less surprising behavior.

{} (brace initialization) disallows implicit conversions that lose information.

## PR Checklist


Modules/Core/Common/include/itkLaplacianOperator.hxx
```diff
@@ -66,7 +66,7 @@ LaplacianOperator<TPixel, VDimension, TAllocator>::GenerateCoefficients() -> Coe
 {
   // Here we set the radius to 1's, here the
   // operator is 3x3 for 2D, 3x3x3 for 3D.
-  constexpr auto r = SizeType::Filled(1);
+  static constexpr auto r{ SizeType::Filled(1) };
```
blowekamp, line 69: Here is a case where the filled array may result in a larger binary. The explicit request for the compiler initialize data structure may increase the size of libraries.

N-Dekker: @blowekamp In practice I think the compiler should be able to inline it all away. Unless in some case when the address of the variable is taken. For example in this case, by having a pointer `p = &r`. When `r` is static, `p = &r` should always point to the same address during the run of the program. When `r` is non-static, `p = &r` may have a different address, each time it is evaluated for another instance of `r`.  However, I don't think it's common to take the address of a constexpr variable  🤷 

But of course, your main point should be measurable: does this change indeed result in a larger binary? Please feel free to do this experiment  😃

N-Dekker:
FYI, this talk convinced me to prefer `static constexpr` over `constexpr`, for local variables:

[C++ Weekly With Jason Turner- Episode 312 - Stop Using `constexpr` (And Use This Instead!)](https://www.youtube.com/watch?v=4pKtPWcl1Go). 

It's from 2022, so it does apply to C++17 as well, but it might be that there are new recommendations, for example for C++20 or later, I just don't know 🤷.

N-Dekker:
Here is where I found plain `constexpr` problemetic:

```cpp
constexpr unsigned int Dimension{ 3 };

auto myLambda = []
{
  return Dimension; // error: 'Dimension' cannot be implicitly captured
};
```

At https://godbolt.org/z/dz9jqbc7d 

MSVC does not allow using a constexpr variable within a lambda, when it is declared outside the lambda!

The error can be fixed, simply by making it `static`! 

```cpp
static constexpr unsigned int Dimension{ 3 };

auto myLambda = []
{
  return Dimension; // OK! Wonderful!
};
```

Modules/Core/Common/include/itkLaplacianOperator.hxx
blowekamp: I tested the following code in godbolt, the array was optimized away when both ones and ret were integer types. 
There appears to be no difference when optimization is enable with or with out the constexpr qualifier.
```
int main()
{
  // Static constexpr array of 1s
  static constexpr std::array<int, 10> ones = {1, 1, 1, 1, 1, 1, 1, 1, 1, 1};
  
  unsigned int ret = 0;
  for (auto n: ones)
    ret += n;
  return ret;

```

hjmjohnson:
I've been reviewing this some more. 

Context                                      |  Recommended form      | Comment
Inside class/struct                    |  static constexpr              | Shared compile-time constant
Inside namespace/global.        |  constexpr                        | Already has static linkage by default
Inside function/block scope     | constexpr                        | Local compile-time constant

=======
In most cases of constexpr (i.e. tests) we are using constexpr inside of the test function or block scope.
Use just constexpr	Why:  In function scope, all variables already have automatic storage duration (per call). 

static would give them static storage (shared, persistent), which is unnecessary unless you explicitly want that for other reasons (like needing a consistent memory address).  The non-static constexpr arrays *may* require copying the .rodata to the stack, and if an element of the local array is returned, it would have been a reference to an element on the stack, thus invalid for returning.  I did not find any instances of this behavior in our code.  

=======
Inside a class definition:   --> Use static constexpr (so that every instance does not treat this differently
Class data members that are constexpr must also be static unless they are inline (C++17+) or initialized per instance.
	•	Non-static data members belong to each instance — but constexpr implies compile-time constant, so it makes sense to make it static.
	•	Effect:
size and start become compile-time constants shared by all instances.
region can then use them in its member initializer list.

=======

After reviewing and thinking more deeply about this, I think my original rational was wrong regarding the need for static.

PR #5569, opened 2025-10-22
COMP: Prefer brace initialization

hjmjohnson:
Since C++11, the brace form {} is preferred in most cases because it provides
safer, more consistent, and less surprising behavior.

{} (brace initialization) disallows implicit conversions that lose information.

from request in #5566

## PR Checklist


Modules/Core/Common/include/itkDerivativeOperator.h
```diff
@@ -39,7 +39,7 @@ namespace itk
          using DerivativeOperatorType = itk::DerivativeOperator<float, 2>;
          DerivativeOperatorType derivativeOperator;
          derivativeOperator.SetDirection(0); // X dimension
-         constexpr auto radius = itk::Size<2>::Filled(1); // A radius of 1 in both dimensions is a 3x3 operator
+         constexpr  auto radius{ itk::Size<2>::Filled(1) }; // A radius of 1 in both dimensions is a 3x3 operator
```
N-Dekker: Sorry, I'm still not a fan of using brace-initialization for variables declared by `auto`. I feel it's so common to write `auto x2 = x1`, and it's just fine. So then, what about multi-line initializations? For example, the lambda at https://github.com/InsightSoftwareConsortium/ITK/blob/3279b33f97a3baeeb6999bfe22edf4331afc5bc5/Modules/Core/Common/src/itkAnatomicalOrientation.cxx#L41

```cpp
  auto enumAsChar = [](CoordinateEnum coord) -> char {
    switch (coord)
    {
      ....
    }
  };

```
Is that still OK to you? Or would you prefer the following?

```cpp
  auto enumAsChar{ [](CoordinateEnum coord) -> char {
    switch (coord)
    {
      ...
    }
  } };
```

seanm: Personally I'm not a fan.  `type var = thing;` is nicer to my eyes, likely because it's just been that way forever.

hjmjohnson:
@seanm @N-Dekker can you please make the changes you want fixed and I'll merge them.  I won't be able to make line-by-line changes based on preferences.  

This was done primarily with manual regex substitutions with manual fixes.  I'll give it a few days for you to make the fixes you want.

hjmjohnson:
My reading of the following

-1) The rules for {} initialization are simpler, more general, less ambiguous, and safer than for other forms of initialization.
-2) Use = only when you are sure that there can be no narrowing conversions. 
-3) For built-in arithmetic types, use = only with auto.```





1) My primary motivation for these changes
2) I read this as "= is allowed when you can verify narrowing conversions can not occur", but {} is still preferred even though they are functionally equivalent in this case.
3) I read this as "allow using = only with auto", but {} is still preferred even though they are functionally equivalent.

=====
Simple decisions about what to use:  {} initialization whenever you can,  sometimes it is functionally equivalent to =, but it is never functionally less safe or less ambiguous than =.


Regarding the core-guidelines:   These guidelines have been evolving over many years, so the use of " = " is in core-guideline examples for different topics written by different recommenders.  Perhaps some of the examples were written before this core-guideline came into being.  That is to say that the core-guidelines are not universally enforced in the core-guidelines documents :).

hjmjohnson:
"Prefer" is a recommendation, not a mandate.  I do not think we need to retro-actively make all the changes at once (or even ever).  I would not go searching for changes like "auto var = 7" as that is allowed as having the same saftely level as "auto var{7};", but for new code, or as we refactor code, I would prefer allowing for changes like "auto var =7; -> auto var{7};" when it is part of some larger efforts.

These are my opinions; I appreciate your opinions as well.  I hope we can come to an agreed-upon compromise.

dzenanz (approved):

PR #5579, opened 2025-10-30
COMP: Silence alloc-size-larger-than warning

hjmjohnson:
Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.hxx:751:49:
  warning: argument 1 value '18446744073709551615' exceeds maximum object size 9223372036854775807 [-Walloc-size-larger-than=]
  |   m_Data[ThreadId].m_LoadTransferBufferLayers = new LayerListType[bufferLayerSize];
  |

Limiting storage size for computing the bufferLayerSize assists with enforcing that it is never possible to allocate too much memory.

This agressive warning recognizes that:

```
void func( size_t var)
{
  size_t allocate_size = 2 * var +1; // Could overflow
  int a = new int[allocate_size];
}
```

This scenerio can never happen because effecively we are calling

```
unsigned char num_layers = ImageDimension;
func(num_layers); // i.e. num_layers can never be greater than 255
```

By reducing the storage size for computing the allocate_size to the smallest size needed to handle the largest value of `2 * var + 1` we can avoid warning.

## PR Checklist

### DIFF
--- a/Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.h
+++ b/Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.h
@@ -301,7 +301,11 @@ public:
   using StatusType = signed char;
 
   /** Type used for storing the number of layers */
-  using LayerCountType = signed char; /* purposefully small integer same as StatusType for backward compatibility */
+  using LayerCountType = StatusType; /* purposefully the same as StatusType
+                                        for backward compatibility. This value
+                                        is compared with up/down layer values
+                                        that can be signed.  Max layer count of
+                                        127 is enough. */
 
   /** The type of the image used to index status information.  Necessary for
    *  the internals of the algorithm. */


dzenanz (approved):

Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.hxx
```diff
@@ -733,10 +733,10 @@ ParallelSparseFieldLevelSetImageFilter<TInputImage, TOutputImage>::ThreadedAlloc
   m_Data[ThreadId].m_Semaphore[0] = 0;
   m_Data[ThreadId].m_Semaphore[1] = 0;
 
-  const std::size_t bufferLayerSize = 2 * m_NumberOfLayers + 1;
+  const unsigned short bufferLayerSize = 2 * m_NumberOfLayers + 1;
```
N-Dekker: Interesting problem, Hans 🤔 ! Would `unsigned int` also be OK, rather than `unsigned short`? (In general `int` and `unsigned int` are typically recommended as default integer types.)

By the way, I noticed that `ParallelSparseFieldLevelSetImageFilter::m_NumberOfLayers` is declared as `StatusType`. That seems a bit peculiar.  Shouldn't it also simply be `unsigned int`? (The similar `SparseFieldLevelSetImageFilter::m_NumberOfLayers` is already  `unsigned int`!) https://github.com/InsightSoftwareConsortium/ITK/blob/aaaf3ad4a3910d3da1f5ad26973d9f04802f0d22/Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.h#L396 versus https://github.com/InsightSoftwareConsortium/ITK/blob/aaaf3ad4a3910d3da1f5ad26973d9f04802f0d22/Modules/Segmentation/LevelSets/include/itkSparseFieldLevelSetImageFilter.h#L497

Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.hxx
```diff
@@ -785,15 +786,16 @@ ParallelSparseFieldLevelSetImageFilter<TInputImage, TOutputImage>::ThreadedAlloc
   // for the Downlists
   m_Data[ThreadId].m_InterNeighborNodeTransferBufferLayers[1] = new LayerPointerType *[m_NumberOfLayers + 1];
 
-  for (unsigned int i = 0; i < static_cast<unsigned int>(m_NumberOfLayers) + 1; ++i)
+  const LayerSizeType one_plus_number_of_layers = static_cast<LayerSizeType>(m_NumberOfLayers + 1);
```
dzenanz: Do we need this static cast here now?

dzenanz (approved):

Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.hxx
hjmjohnson: Removed.

Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.h
```diff
@@ -393,7 +393,8 @@ class ITK_TEMPLATE_EXPORT ParallelSparseFieldLevelSetImageFilter
    * consist of m_NumberOfLayers layers on both sides of a single active layer.
    * This active layer is the interface of interest, i.e. the zero
    * level set. */
-  StatusType m_NumberOfLayers{};
+  using LayerCountType = unsigned char;
+  LayerCountType m_NumberOfLayers{};
```
N-Dekker, line 403: Thanks Hans. I see, StatusType is _signed_ char, so did you intentionally switch to _unsigned_ char? Anyway, I think the macro calls should then be adjusted as well, as they now still say:

    itkSetMacro(NumberOfLayers, StatusType);
    itkGetConstMacro(NumberOfLayers, StatusType);

 A switch from `signed char` to `unsigned char` would be a small API change, of course, but no problem to me.  

Is there a specific reason for using an 8-bit char? Should `NumberOfLayers` always remain below `256`? Honestly I don't know what the filter is about 🤷

Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.hxx
hjmjohnson: Changed as requested.

Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.h
dzenanz: Number of layers is proportional to dimension, something like 2*dim+1.

N-Dekker: So then, it seems logical to have the same type for `NumberOfLayers` and `ImageDimension`, right? Just like in the non-parallel SparseFieldLevelSetImageFilter: https://github.com/InsightSoftwareConsortium/ITK/blob/aaaf3ad4a3910d3da1f5ad26973d9f04802f0d22/Modules/Segmentation/LevelSets/include/itkSparseFieldLevelSetImageFilter.h#L497

Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.h
```diff
@@ -301,7 +301,11 @@ class ITK_TEMPLATE_EXPORT ParallelSparseFieldLevelSetImageFilter
   using StatusType = signed char;
 
   /** Type used for storing the number of layers */
-  using LayerCountType = signed char; /* purposefully small integer same as StatusType for backward compatibility */
+  using LayerCountType = StatusType; /* purposefully the same as StatusType
```
dzenanz, line 304: The commit message does not correspond to this change.

Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.h
hjmjohnson: This is not my code, and I don't use this feature.  I'm trying to resolve a compiler warning, not evaluate the algorithm :).  

I tried making this unsigned int as a test, and that is going to require a lot of algorithmic considerations.  Making both StatusType and LayerCountType the same type avoids other compiler errors when these two concepts are used together.  I think it is common to only have a few layers (1 or 2) in each level set.  Even 127 seems to be an enormous number of layers.  

I reverted to :

```c++
using LayerCountType = StatusType;
```
and added a comment.

Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.h
hjmjohnson: OH:  commit message.... not "comment message".  Perhaps I did a rebase squash wrong.

hjmjohnson: I screwed up the squash commits.  Comment is now fixed.

Modules/Segmentation/LevelSets/include/itkParallelSparseFieldLevelSetImageFilter.h
hjmjohnson: @N-Dekker Conceptually, I agree with you, but that would require a much deeper dive into the codebase to ensure that it is implemented correctly.   I do not have the time to take that deep dive at the moment.  I think this PR is the minimal improvements that remove the compiler warning and safely keep the same behavior.

PR #5583, opened 2025-10-31
ITK_GTEST_EXERCISE_BASIC_OBJECT_METHODS: a macro for GoogleTest based ITK tests

N-Dekker:
The existing `ITK_EXERCISE_BASIC_OBJECT_METHODS` appears hard to used in GoogleTest based tests, because it may return `EXIT_FAILURE`, whereas GoogleTest based unit tests always return `void`.

A call to the proposed new `ITK_GTEST_EXERCISE_BASIC_OBJECT_METHODS` never causes the test to return. Internally it calls the GoogleTest specific macro's `EXPECT_FALSE` and `EXPECT_STREQ`.

First use case: `itkHausdorffDistanceImageFilterGTest`

### DIFF
--- a/Modules/Filtering/DistanceMap/test/itkHausdorffDistanceImageFilterGTest.cxx
+++ b/Modules/Filtering/DistanceMap/test/itkHausdorffDistanceImageFilterGTest.cxx
@@ -19,9 +19,7 @@
 // First include the header file to be tested:
 #include "itkHausdorffDistanceImageFilter.h"
 
-#include "itkTestingMacros.h"
-
-#include <gtest/gtest.h>
+#include "itkGTest.h"
 
 TEST(HausdorffDistanceImageFilter, Test)
 {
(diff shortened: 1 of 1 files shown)


Modules/Core/TestKernel/include/itkGTest.h
```diff
@@ -32,6 +32,17 @@
...
+#define ITK_GTEST_EXERCISE_BASIC_OBJECT_METHODS(object, ClassName, SuperclassName) \
+  [object] {                                                                       \
+    std::ostringstream outputStreamForPrintExercise;                               \
+    object->Print(outputStreamForPrintExercise);                                   \
+    EXPECT_FALSE(outputStreamForPrintExercise.str().empty());                      \
+    EXPECT_STREQ(object->Self::GetNameOfClass(), #ClassName);                      \
+    EXPECT_STREQ(object->Superclass::GetNameOfClass(), #SuperclassName);           \
+  }()
+
+
```
N-Dekker, line 45: For comparison, here is the existing `ITK_EXERCISE_BASIC_OBJECT_METHODS` macro: https://github.com/InsightSoftwareConsortium/ITK/blob/82caa0a1ea7ab2955d13c895ded9602b42777fb1/Modules/Core/TestKernel/include/itkTestingMacros.h#L53-L84

dzenanz (approved):

PR #5586, opened 2025-10-31
MISC: Fix minor warnings

hjmjohnson:
## PR Checklist

### DIFF
--- a/Modules/Core/SpatialObjects/include/itkSurfaceSpatialObject.hxx
+++ b/Modules/Core/SpatialObjects/include/itkSurfaceSpatialObject.hxx
@@ -101,10 +101,8 @@ SurfaceSpatialObject<TDimension, TSurfacePointType>::ComputeNormals()
       identifier[1] = 0;
       identifier[2] = 0;
 
-      float max[3];
-      max[0] = 99999999;
-      max[1] = 99999999;
-      max[2] = 99999999;
+      constexpr float max_float = std::numeric_limits<float>::max();
+      float           max[3] = { max_float, max_float, max_float };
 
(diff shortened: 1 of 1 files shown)


Modules/Core/SpatialObjects/include/itkSurfaceSpatialObject.hxx
```diff
@@ -101,10 +101,7 @@ SurfaceSpatialObject<TDimension, TSurfacePointType>::ComputeNormals()
       identifier[1] = 0;
       identifier[2] = 0;
 
-      float max[3];
-      max[0] = 99999999;
-      max[1] = 99999999;
-      max[2] = 99999999;
+      float max[3] = { 99999999.0, 99999999.0, 99999999.0 };
```
issakomi: I am afraid that 32 bits float can not represent either 99999999 or 99999999.0. It will be rounded to 1e+8. This may not be a problem, it just looks bad.

hjmjohnson: I dont' know this code, but it should probably be numeric_limits<float>::max

issakomi: IMHO, I don't know this code too, probably safer leave as is, std::numeric_limits<float>::max()  is 3.40282e+38

hjmjohnson: I read the code more carefully and verified that the variable max[3] is used to identify the maximum values in an array.  Initializing to std::numeric_limits::max() is a better initializer.

PR #5588, opened 2025-11-01
COMP: Refactor ccache configuration into standalone CMake module

hjmjohnson:
ccache has more utility than only wrapped versions.  since ccache
version 4.0 (december of 2020), support for SWIG is built in.  There is
no need for ccache-swig support after ccache version 4.0, so the logic
could be simplified.

Moved ccache setup from `Wrapping/Generators/CMakeLists.txt` to a
dedicated `itkCCacheSupport.cmake` module for better modularity and
reusability. Updated root CMakeLists.txt to include the new module.

This significantly speeds up clean builds

## PR Checklist


thewtex (approved):
👍

PR #5589, opened 2025-11-02
STYLE: Use `ImageRegionRange` in DistanceMap tests

N-Dekker:
Replaced hand-written `while` loops (which were using the old `ImageRegionIterator`) with modern range-based `std` algorithm calls.

### DIFF
--- a/Modules/Filtering/DistanceMap/test/itkContourDirectedMeanDistanceImageFilterTest.cxx
+++ b/Modules/Filtering/DistanceMap/test/itkContourDirectedMeanDistanceImageFilterTest.cxx
@@ -17,8 +17,12 @@
  *=========================================================================*/
 
 #include "itkContourDirectedMeanDistanceImageFilter.h"
+#include "itkImageRegionRange.h"
 #include "itkSimpleFilterWatcher.h"
 #include "itkTestingMacros.h"
+
+#include <algorithm> // For fill.
+#include <numeric>   // For iota.
 
 int
 itkContourDirectedMeanDistanceImageFilterTest(int, char *[])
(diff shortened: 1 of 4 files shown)


Modules/Filtering/DistanceMap/test/itkContourDirectedMeanDistanceImageFilterTest.cxx
```diff
@@ -54,20 +58,11 @@ itkContourDirectedMeanDistanceImageFilterTest(int, char *[])
...
 
-  itk::ImageRegionIterator<Image1Type> it1(image1, region1);
-  Pixel1Type                           count{};
-  while (!it1.IsAtEnd())
-  {
-    it1.Set(++count);
-    ++it1;
-  }
+  const itk::ImageRegionRange<Image1Type> imageRegionRange1(*image1, region1);
+  std::iota(imageRegionRange1.begin(), imageRegionRange1.end(), Pixel1Type{ 1 });
```
N-Dekker, line 62: With C++20, this line could become `std::ranges::iota(imageRegionRange1, Pixel1Type{ 1 })`  😃

Modules/Filtering/DistanceMap/test/itkContourDirectedMeanDistanceImageFilterTest.cxx
```diff
@@ -54,20 +58,11 @@ itkContourDirectedMeanDistanceImageFilterTest(int, char *[])
...
 
-  itk::ImageRegionIterator<Image2Type> it2(image2, region2);
-  while (!it2.IsAtEnd())
-  {
-    it2.Set(7.2);
-    ++it2;
-  }
+  const itk::ImageRegionRange<Image2Type> imageRegionRange2(*image2, region2);
+  std::fill(imageRegionRange2.begin(), imageRegionRange2.end(), Pixel2Type{ 7.2 });
```
N-Dekker, line 65: Similarly, with C++20, this line could become `std::ranges::fill(imageRegionRange2, Pixel2Type{ 7.2 })` 😃

Modules/Filtering/DistanceMap/test/itkContourDirectedMeanDistanceImageFilterTest.cxx
hjmjohnson: Soon :).

Modules/Filtering/DistanceMap/test/itkContourDirectedMeanDistanceImageFilterTest.cxx
hjmjohnson: Now that we are in beta phase of ITKv6, perhaps we should target c++20 for ITKv7.

PR #5593, opened 2025-11-03
COMP: Setting up the mac pixi numpy environment

hjmjohnson:
pixi python environments were failing due to failures importing
numpy.  

On macOS, conda-forge NumPy links against OpenBLAS and the Fortran
runtime; the dylib is @rpath/libgfortran.5.dylib (package:
libgfortran5). If you installed NumPy via pip inside a Pixi env, or
you’re importing from outside the Pixi env, the loader can’t find that
dylib.

```
python -c "import numpy; numpy.show_config()"
```

## PR Checklist


hjmjohnson:
Matt thanks.  I moved it when trying to figure out how to make the python environment work with numpy.   I fogot to move it back when I figured out that openblas and gfortan need to be installed.

PR #5600, opened 2025-11-04
COMP: Migrate macos runners from macos-13 to macos-15-intel

thewtex:
macos-13 (an Intel runner) will not be available after December 2025.

### DIFF
--- a/.github/workflows/clang-format-linter.yml
+++ b/.github/workflows/clang-format-linter.yml
@@ -10,8 +10,8 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v4
-    - uses: InsightSoftwareConsortium/ITKClangFormatLinterAction@master
+    - uses: actions/checkout@v5
+    - uses: InsightSoftwareConsortium/ITKClangFormatLinterAction@main
       with:
         error-message: 'Code is inconsistent with ITK Coding Style. Add the *action:ApplyClangFormat* PR label to correct.'
         itk-branch: release-5.4


thewtex:
On Azure Pipelines `macos-15` is also an Intel runner.

.github/workflows/clang-format-linter.yml
```diff
@@ -10,8 +10,8 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v4
-    - uses: InsightSoftwareConsortium/ITKClangFormatLinterAction@master
+    - uses: actions/checkout@v5
+    - uses: InsightSoftwareConsortium/ITKClangFormatLinterAction@main
```
dzenanz, line 14: We will need to update remote module CIs with this rename.

SimonRit: Yes. FYI, we have [removed this linter](https://github.com/RTKConsortium/RTK/pull/843) on RTK and [now use pre-commit](https://github.com/RTKConsortium/RTK/pull/815) to more comprehensively check style (CMake, C++, Python).

Testing/ContinuousIntegration/AzurePipelinesMacOS.yml
```diff
@@ -31,8 +31,7 @@ jobs:
   timeoutInMinutes: 0
   cancelTimeoutInMinutes: 300
   pool:
-    # https://github.com/actions/runner-images?tab=readme-ov-file#available-images
-    vmImage: macos-13
+    vmImage: macos-15
```
N-Dekker, line 34: Thanks Matt! Just wondering, is there a specific reason not to use `macos-14` on Azure Pipelines?

thewtex: It will be easier to compare intel / arm with our macos 15 arm builds.

N-Dekker: https://github.com/actions/runner-images?tab=readme-ov-file#available-images mentions two macos-14 versions:

- `macos-14-large` | [macOS-14](https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md)
- `macos-14` or `macos-14-xlarge` | [macOS-14-arm64](https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md)

Are both of them inconvenient to use on Azure Pipelines?

thewtex: It is important to keep in mind:

- The GitHub Actions runner images are not the same as the Azure Pipelines images
- We do not have access to all images for free / open source use
- We should limit the number of builds to avoid long queues

.github/workflows/pixi.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       timeout-minutes: 0
       strategy:
         matrix:
-          os: [ubuntu-22.04, windows-2022, macos-13, macos-14]
+          os: [ubuntu-22.04, windows-2022, macos-15-intel, macos-14]
```
N-Dekker, line 35: For the record, `macos-14-large` is also still intel.

According to [GitHub Actions: macOS 13 runner image is closing down](https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/#what-you-need-to-do): 


_For users that require the x86_64 (Intel) architecture, we have added a new label to support standard runner users. Jobs can be migrated to one of the following labels:
– `macos-15-intel` (new)
– `macos-14-large`
– `macos-latest-large` or `macos-15-large`_

thewtex: For free / open source usage, access to `-large` images is not available.

PR #5602, opened 2025-11-04
Remove unnecessary `resize(0)` and `clear()` calls, replace `resize(0)` with `clear()`

N-Dekker:
Four little style commits.

### DIFF
--- a/Modules/IO/PhilipsREC/src/itkPhilipsPAR.cxx
+++ b/Modules/IO/PhilipsREC/src/itkPhilipsPAR.cxx
@@ -403,7 +403,7 @@ PhilipsPAR::GetLineNumber(std::string file, int lineNum)
   if (file != this->m_FileName)
   {
     this->m_FileName = file;
-    this->m_PARFileLines.resize(0);
+    this->m_PARFileLines.clear();
 
     // Try to read the text file.
     std::ifstream local_InputStream;
--- a/Modules/Core/Common/include/itkColorTable.hxx
+++ b/Modules/Core/Common/include/itkColorTable.hxx
@@ -32,8 +32,8 @@ template <typename TComponent>
 void
 ColorTable<TComponent>::DeleteColors()
 {
-  m_Color.resize(0);
-  m_ColorName.resize(0);
+  m_Color.clear();
+  m_ColorName.clear();
 }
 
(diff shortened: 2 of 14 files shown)


Modules/IO/PhilipsREC/src/itkPhilipsPAR.cxx
```diff
@@ -384,9 +384,7 @@ GetImageInformationDefinitionV42(std::string file, int lineNum, PhilipsPAR * phi
 
 PhilipsPAR::PhilipsPAR()
   : m_FileName("")
-{
-  this->m_PARFileLines.resize(0);
-}
+{}
```
dzenanz, line 387: This constructor can now be defaulted, right?

Modules/Segmentation/Classifiers/include/itkClassifierBase.hxx
```diff
@@ -25,9 +25,7 @@ namespace itk
 template <typename TDataContainer>
 ClassifierBase<TDataContainer>::ClassifierBase()
   : m_DecisionRule(nullptr)
-{
-  m_MembershipFunctions.resize(0);
-}
+{}
```
dzenanz, line 28: This constructor can now be defaulted, right?

Modules/IO/PhilipsREC/src/itkPhilipsPAR.cxx
N-Dekker: Yes indeed, thanks @dzenanz. Looks like it could be a nice follow-up PR😃 

For this PR (5602), it looks clearer to me to just focus on `resize(0)` and `clear()`.  Especially because the `PhilipsPAR` constructor is not _entirely_ empty, after the removal of `resize(0)`. And ITK has a few more default-constructors that may be defaulted as well, for example PatchBasedDenoisingBaseImageFilter(), and ProjectedIterativeDeconvolutionImageFilter().

hjmjohnson (approved):

Modules/Segmentation/Classifiers/include/itkClassifierBase.hxx
N-Dekker: - Addressed by pull request #5621  😃

PR #5606, opened 2025-11-06
ITK_GTEST_SET_GET_BOOLEAN: another macro for GoogleTest based ITK tests

N-Dekker:
Proposing a new macro, `ITK_GTEST_SET_GET_BOOLEAN`, similar to the existing `ITK_TEST_SET_GET_BOOLEAN`, but now specifically for GoogleTest based ITK tests.

First use case: `itkContourDirectedMeanDistanceImageFilterGTest`

- Follow-up to pull request #5583

### DIFF
--- a/Modules/Filtering/DistanceMap/test/CMakeLists.txt
+++ b/Modules/Filtering/DistanceMap/test/CMakeLists.txt
@@ -9,7 +9,6 @@ set(
   itkSignedDanielssonDistanceMapImageFilterTest1.cxx
   itkSignedDanielssonDistanceMapImageFilterTest2.cxx
   itkContourMeanDistanceImageFilterTest.cxx
-  itkContourDirectedMeanDistanceImageFilterTest.cxx
   itkFastChamferDistanceImageFilterTest.cxx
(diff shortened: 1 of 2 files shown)


Modules/Core/TestKernel/include/itkGTest.h
```diff
@@ -43,6 +43,20 @@
...
+  EXPECT_TRUE(object->Get##variable());                    \
+  object->Set##variable(false);                            \
+  EXPECT_FALSE(object->Get##variable());                   \
+  object->variable##On();                                  \
+  EXPECT_TRUE(object->Get##variable());                    \
+  object->variable##Off();                                 \
+  EXPECT_FALSE(object->Get##variable());                   \
+  object->Set##variable(value)
+
+
```
N-Dekker, line 59: For comparison, the existing `ITK_TEST_SET_GET_BOOLEAN` macro is defined here: https://github.com/InsightSoftwareConsortium/ITK/blob/ce2052c6d3ff8eb783ae62bbfa373a44f4da731b/Modules/Core/TestKernel/include/itkTestingMacros.h#L227-L257

PR #5608, opened 2025-11-07
STYLE: Use `generate` in tests, assigning values from vnl_sample_uniform

N-Dekker:
Did find code like:

    v[0] = static_cast<T>(vnl_sample_uniform(a, b));
    v[1] = static_cast<T>(vnl_sample_uniform(a, b));
    v[2] = static_cast<T>(vnl_sample_uniform(a, b));

And replaced it with the equivalent `std::generate` call:

    std::generate(v.begin(), v.end(), [] {
      return static_cast<T>(vnl_sample_uniform(a, b));
    });

Aims to reduce code redundancy.

----

Motivation: in a follow-up I would like those `vnl_sample_uniform` calls to be replaced with modern C++ random number generation. This pull request should pave the way.

### DIFF
--- a/Modules/Core/Common/test/itkPointSetTest.cxx
+++ b/Modules/Core/Common/test/itkPointSetTest.cxx
@@ -19,6 +19,9 @@
 #include "itkPointSet.h"
 #include "vnl/vnl_sample.h"
 #include "itkTestingMacros.h"
+
+#include <algorithm> // For generate.
+#include <iterator>  // For begin and end.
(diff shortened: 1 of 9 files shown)


dzenanz (approved):

N-Dekker:
> @N-Dekker It would be good to start deprecating the exposure of vnl capabilities when better alternatives exist.

- @hjmjohnson Here is a small step in that direction: pull request #5611

PR #5611, opened 2025-11-08
ENH: Replace "vnl/vnl_sample.h" with `<random>`, in Core tests

N-Dekker:
Replaced code like `randomNumber = vnl_sample_uniform(min, max)` with something like:

    std::mt19937                 randomNumberEngine{};
    std::uniform_distribution<T> randomNumberDistribution(min, max);

    randomNumber = randomNumberDistribution(randomNumberEngine);

`vnl_sample_uniform` is discouraged, especially because it has a global state (static data). This problem is addressed by declaring random number engines as local (non-static) variables, local to the tests that use them.

----

Obviously, this modernization yields more lines of code than before. While in general, I like it better when modernization yields _less_ code! But I can't help it 🤷 It appears fundamental to modern random number generation to declare separate "engine" and "distribution" variables, instead of just doing a single library function call.

### DIFF
--- a/Modules/Core/Common/test/itkPointSetTest.cxx
+++ b/Modules/Core/Common/test/itkPointSetTest.cxx
@@ -17,12 +17,12 @@
  *=========================================================================*/
 
 #include "itkPointSet.h"
-#include "vnl/vnl_sample.h"
 #include "itkTestingMacros.h"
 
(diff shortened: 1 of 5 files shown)


hjmjohnson (approved):
Lovely!!!

PR #5618, opened 2025-11-10
ENH: Convert IsoContourDistanceImageFilter test to GoogleTest

N-Dekker:
- Follow-up to pull request #5614

### DIFF
--- a/Modules/Filtering/DistanceMap/test/CMakeLists.txt
+++ b/Modules/Filtering/DistanceMap/test/CMakeLists.txt
@@ -9,7 +9,6 @@ set(
   itkSignedDanielssonDistanceMapImageFilterTest2.cxx
   itkSignedMaurerDistanceMapImageFilterTest.cxx
   itkApproximateSignedDistanceMapImageFilterTest.cxx
-  itkIsoContourDistanceImageFilterTest.cxx
 )
 
 createtestdriver(ITKDistanceMap "${ITKDistanceMap-Test_LIBRARIES}" "${ITKDistanceMapTests}")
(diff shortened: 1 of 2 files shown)


dzenanz (approved):

hjmjohnson (approved):

Modules/Filtering/DistanceMap/test/CMakeLists.txt
```diff
@@ -9,7 +9,6 @@ set(
   itkSignedDanielssonDistanceMapImageFilterTest2.cxx
   itkSignedMaurerDistanceMapImageFilterTest.cxx
   itkApproximateSignedDistanceMapImageFilterTest.cxx
-  itkIsoContourDistanceImageFilterTest.cxx
```
N-Dekker, line 12: With this commit, 10 out of 18 DistanceMap tests are converted to GoogleTest. The remaining 8 tests seem harder to convert to GoogleTest, because most of them do file I/O, using data files from `${ITK_DATA_ROOT}`. The data file names cannot be passed directly to GoogleTest unit tests, but maybe via a CMake configure file ("itkDataRoot.h.in"), a compile option (`/D ITK_DATA_ROOT=/path/to/data`), or an environment variable...

dzenanz: Let's keep classic testing style for them, then 😄

N-Dekker: OK... for the time being, at least  `;-)`

PR #5622, opened 2025-11-11
WIP: Replace "vnl_sample.h" with `<random>` in ConnectedComponent tests

N-Dekker:
- Another follow-up to pull request #5611

### DIFF
--- a/Modules/Segmentation/ConnectedComponents/test/itkConnectedComponentImageFilterTest.cxx
+++ b/Modules/Segmentation/ConnectedComponents/test/itkConnectedComponentImageFilterTest.cxx
@@ -22,9 +22,9 @@
 #include "itkImageFileReader.h"
 #include "itkImageFileWriter.h"
 #include "itkSimpleFilterWatcher.h"
-#include "vnl/vnl_sample.h"
 #include "itkTestingMacros.h"
 #include <algorithm> // For generate.
+#include <random>    // For mt19937.
 
 int
 itkConnectedComponentImageFilterTest(int argc, char * argv[])
@@ -124,12 +124,20 @@ itkConnectedComponentImageFilterTest(int argc, char * argv[])
 
   std::vector<RGBPixelType> colormap;
   colormap.resize(numObjects + 1);
-  vnl_sample_reseed(1031571);
+
+  using RGBComponentType = RGBPixelType::ComponentType;
+  constexpr auto maxRGBComponentValue = std::numeric_limits<RGBComponentType>::max();
+
+  constexpr std::mt19937::result_type randomSeed{ 1031571 };
+  std::mt19937                        randomNumberEngine(randomSeed);
+  std::uniform_int_distribution<>     randomNumberDistribution(maxRGBComponentValue / 3, maxRGBComponentValue);
+
   for (auto & i : colormap)
   {
     RGBPixelType px;
-    std::generate(
-      px.begin(), px.end(), [] { return static_cast<unsigned char>(255 * vnl_sample_uniform(0.3333, 1.0)); });
+    std::generate(px.begin(), px.end(), [&randomNumberEngine, &randomNumberDistribution] {
+      return static_cast<RGBComponentType>(randomNumberDistribution(randomNumberEngine));
+    });
 
     i = px;
   }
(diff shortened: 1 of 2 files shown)


dzenanz (approved):
Looks good on a glance.

N-Dekker:
There appeared test failures at https://open.cdash.org/viewTest.php?onlyfailed&buildid=10817138 on the following four tests:

1. itkConnectedComponentImageFilterTest
2. itkConnectedComponentImageFilterTest2
3. itkConnectedComponentImageFilterTest3
4. itkConnectedComponentImageFilterTestRGB

I'm sorry, I don't really know how to fix them, but I guess we would have to change the ground truth images (somewhere in `${ITK_DATA_ROOT}`, in order to make this work. For example, for itkConnectedComponentImageFilterTestRGB: https://github.com/InsightSoftwareConsortium/ITK/blob/6568876c1b5eb3ee398dd69bd1a0e960dcca8da6/Modules/Segmentation/ConnectedComponents/test/CMakeLists.txt#L44-L55

Does anyone have suggestion?

Modules/Segmentation/ConnectedComponents/test/itkConnectedComponentImageFilterTest.cxx
```diff
@@ -124,12 +124,19 @@ itkConnectedComponentImageFilterTest(int argc, char * argv[])
 
   std::vector<RGBPixelType> colormap;
   colormap.resize(numObjects + 1);
-  vnl_sample_reseed(1031571);
```
dzenanz, line 127: Have you tried setting this seed to std's randomNumberEngine?

hjmjohnson: @N-Dekker Before we add new baseline images, we will want to ensure that a consistent seed value is set and stable across all platforms.

Modules/Segmentation/ConnectedComponents/test/itkConnectedComponentImageFilterTest.cxx
```diff
@@ -124,12 +124,19 @@ itkConnectedComponentImageFilterTest(int argc, char * argv[])
...
+
+  using RGBComponentType = RGBPixelType::ComponentType;
+  constexpr auto maxRGBComponentValue = std::numeric_limits<RGBComponentType>::max();
+
+  std::mt19937                    randomNumberEngine{};
```
hjmjohnson: @N-Dekker does {} initialize the seed to a constant value?  

I think 
```c++
constexpr auto random_seed = 1031571;
std::mt19937 randomNumberEngine{random_seed};
```
would make this easier to interpret.

Modules/Segmentation/ConnectedComponents/test/itkConnectedComponentImageFilterTest.cxx
N-Dekker: OK, I can try, but I would be surprised if "vnl_sample" and `std::mt19937` would produce _exactly_ the same sequence of random numbers... we'll see.

Modules/Segmentation/ConnectedComponents/test/itkConnectedComponentImageFilterTest.cxx
N-Dekker: The default constructor of `std::mt19937` always uses the same seed: `default_seed = 5489`. That's different from `1031571`, obviously. So we'll see if using `1031571` will make the CI happy...

hjmjohnson: That's great to know, but not obvious.  I would have guessed the default was to use an unknown changing seed unless explicitly specified.  I do think that setting the seed explicitly (with what ever value you want, even 5489) would be more meaningful.

N-Dekker: When the C++ standard specifies that the default-constructor of `std::mt19937` always uses the same seed, `default_seed`, I think we should be allowed to just trust the C++ standard.

We could make it explicit:

    std::mt19937 randomNumberEngine(std::mt19937::default_seed);

And then add to our unit tests:

    static_assert(std::mt19937::default_seed == 5489, "Check on any platform!");

But that all seems overdone to me. If we want to have _non-deterministic_ random numbers, we should use `std::random_device`.

Modules/Segmentation/ConnectedComponents/test/itkConnectedComponentImageFilterTest.cxx
hjmjohnson: They won't! (I've verified that previously).  It is only important that after this change that the baselines are a new stable value.

PR #5641, opened 2025-11-18
Replace iterative `Set(number)` calls with `FillBuffer(number)`, in tests

N-Dekker:
Replaced manual loops like

```cpp
  IteratorType it(image, image->GetBufferedRegion());
  while (!it.IsAtEnd())
  {
    it.Set(x);
    ++it;
  }
```

With the equivalent `image->FillBuffer(x)` call.

### DIFF
--- a/Modules/Registration/Common/test/itkMattesMutualInformationImageToImageMetricTest.cxx
+++ b/Modules/Registration/Common/test/itkMattesMutualInformationImageToImageMetricTest.cxx
@@ -132,7 +132,9 @@ TestMattesMetricWithAffineTransform(TInterpolator * interpolator,
   auto imgMovingMask = MovingImageType::New();
   imgMovingMask->CopyInformation(imgMoving);
   imgMovingMask->SetRegions(region);
-  imgMovingMask->AllocateInitialized();
+  imgMovingMask->Allocate();
+  // Set all moving mask voxels to 1
+  imgMovingMask->FillBuffer(1);
 
   auto imgFixedMask = FixedImageType::New();
   imgFixedMask->CopyInformation(imgFixed);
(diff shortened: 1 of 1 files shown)


Modules/Registration/Common/test/itkMattesMutualInformationImageToImageMetricTest.cxx
```diff
@@ -143,31 +145,20 @@ TestMattesMetricWithAffineTransform(TInterpolator * interpolator,
   { // Set up a mask that only has every 10th voxel listed is used in
```
dzenanz: Does this comment need updating? Is it every 17th?

dzenanz (approved):

dzenanz (approved):

Modules/Registration/Common/test/itkMattesMutualInformationImageToImageMetricTest.cxx
```diff
@@ -132,42 +132,33 @@ TestMattesMetricWithAffineTransform(TInterpolator * interpolator,
...
   auto imgFixedMask = FixedImageType::New();
   imgFixedMask->CopyInformation(imgFixed);
   imgFixedMask->SetRegions(region);
   imgFixedMask->AllocateInitialized();
 
   int NumberFixedImageMaskVoxels = 0;
-  { // Set up a mask that only has every 10th voxel listed is used in
+  { // Set up a mask that only has every 17th voxel listed is used in
     // fixed image region.
     // This should result in only about 588 samples
```
N-Dekker, line 147: Well spotted, @dzenanz, thanks! By the way, I checked that the number of "about 588 samples" is still OK, for an image of 100 x 100 pixels. I mean, `100 x 100 / 17` is around `588` 😃

PR #5644, opened 2025-11-19
STYLE: Replace FillImage with AllocateInitialized() calls, in tests

N-Dekker:
Reduced amount of duplicate code.

### DIFF
--- a/Modules/Registration/FEM/test/itkFEMFiniteDifferenceFunctionLoadTest.cxx
+++ b/Modules/Registration/FEM/test/itkFEMFiniteDifferenceFunctionLoadTest.cxx
@@ -54,19 +54,6 @@ using DemonsRegistrationMetricType =
   itk::DemonsRegistrationFunction<InputImageType, InputImageType, DeformationFieldImageType>;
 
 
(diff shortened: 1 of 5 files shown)


Modules/Registration/FEM/test/itkFEMFiniteDifferenceFunctionLoadTest.cxx
```diff
@@ -54,19 +54,6 @@ using DemonsRegistrationMetricType =
...
-template <typename TImage>
-void
-FillImage(TImage * image, typename TImage::PixelType value)
-{
-  using Iterator = itk::ImageRegionIteratorWithIndex<TImage>;
-  Iterator it(image, image->GetBufferedRegion());
-  for (it.GoToBegin(); !it.IsAtEnd(); ++it)
-  {
-    it.Set(value);
-  }
```
N-Dekker, line 67: Initially I thought this should be implemented, simply by calling `image->FillBuffer(value)`. But then it appeared that the little local helper function was _only_ called to do zero-filling! So then, `image->AllocateInitialized()` appears preferable! 😺

PR #5647, opened 2025-11-20
BUG: `EXPECT_EQ` in `while` loop in BSplineTransformGTest unreachable

N-Dekker:
The condition `!iter1.IsAtEnd() && iter2.IsAtEnd()` is usually false, because both iterators are expected to iterate over the same (buffered) region.

It should be sufficient to just check if the first iterator is not at the end. And then, it should be safe to _assert_ that the second iterator is not at the end either.

### DIFF
--- a/Modules/Core/Transform/test/itkBSplineTransformGTest.cxx
+++ b/Modules/Core/Transform/test/itkBSplineTransformGTest.cxx
@@ -71,9 +71,9 @@ bspline_eq(const itk::BSplineTransform<TParametersValueType, VDimension, VSpline
     ImageConstIterator iter1(coeffImage1, coeffImage1->GetBufferedRegion());
     ImageConstIterator iter2(coeffImage2, coeffImage2->GetBufferedRegion());
 
-
-    while (!iter1.IsAtEnd() && iter2.IsAtEnd())
+    while (!iter1.IsAtEnd())
     {
+      ASSERT_FALSE(iter2.IsAtEnd());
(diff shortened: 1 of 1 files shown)


Modules/Core/Transform/test/itkBSplineTransformGTest.cxx
```diff
@@ -71,9 +71,9 @@ bspline_eq(const itk::BSplineTransform<TParametersValueType, VDimension, VSpline
...
     ImageConstIterator iter2(coeffImage2, coeffImage2->GetBufferedRegion());
 
-
-    while (!iter1.IsAtEnd() && iter2.IsAtEnd())
+    while (!iter1.IsAtEnd())
     {
+      ASSERT_FALSE(iter2.IsAtEnd());
       EXPECT_EQ(iter1.Get(), iter2.Get()) << description << " Expected value at: " << iter1.GetIndex();
       ++iter1;
       ++iter2;
```
N-Dekker, line 79: With the original code, the body of this `while` loop is never executed. So it did not have full code coverage. Does that answer your question, @blowekamp ?

I think the original code intended to do  `! iter1.IsAtEnd() && ! iter2.IsAtEnd()`. (So a ` ! ` was missing.) But it should be sufficient to just check the first iterator (and then assert that the second one is also still OK).

blowekamp: Ahh.. yes. I missed the missing `!` in the logic reshuffle.

hjmjohnson (approved):

PR #5650, opened 2025-11-21
Use ImageBufferRange in ImagePCADecompositionCalculatorTest, ImageAlgorithmCopyTest2 and BSplineTransformGTest

N-Dekker:
Modernized various tests by replacing manual iterations with `std` algorithm calls, using `itk::ImageBufferRange`.

### DIFF
--- a/Modules/Filtering/ImageStatistics/test/itkImagePCADecompositionCalculatorTest.cxx
+++ b/Modules/Filtering/ImageStatistics/test/itkImagePCADecompositionCalculatorTest.cxx
@@ -21,7 +21,11 @@
 #include "itkTextOutput.h"
 
 #include "itkImagePCADecompositionCalculator.h"
+#include "itkImageBufferRange.h"
 #include "itkTestingMacros.h"
+
+#include <algorithm> // For copy.
+#include <iterator>
 
 // class to support progress feedback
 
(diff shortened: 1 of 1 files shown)


Modules/Filtering/ImageStatistics/test/itkImagePCADecompositionCalculatorTest.cxx
```diff
@@ -337,11 +341,10 @@ itkImagePCADecompositionCalculatorTest(int, char *[])
...
     std::cout << '[';
-    InputImageIterator basisImage_it(basis_it, basis_it->GetBufferedRegion());
-    for (basisImage_it.GoToBegin(); !basisImage_it.IsAtEnd(); ++basisImage_it)
-    {
-      std::cout << basisImage_it.Get() << ' ';
-    }
+    const itk::ImageBufferRange<const InputImageType> imageBufferRange(*basis_it);
+    std::copy(imageBufferRange.cbegin(),
+              imageBufferRange.cend(),
+              std::ostream_iterator<InputImageType::PixelType>(std::cout, " "));
```
dzenanz, line 347: Why change from character (`' '`) to string (`" "`)?

N-Dekker: Good question... Just to make sure the right `ostream_iterator` constructor is chosen!

`ostream_iterator( ostream_type& stream, const CharT* delim )` in https://en.cppreference.com/w/cpp/iterator/ostream_iterator/ostream_iterator.html

N-Dekker: But maybe we should propose an extra overload of `std::ostream_iterator::ostream_iterator` to support a single `char` as delimiter! 😸

PR #5651, opened 2025-11-21
Call FillBuffer on inputImage (A, B, or C) in ImageIntensity tests

N-Dekker:
Replaced manual `while (!it.IsAtEnd())` loops.

Found by `([ ]+)while \(!(it)\.IsAtEnd\(\)\)\r\n\1{\r\n\1  \2\.Set\((\w+)\);\r\n\1  \+\+\2;\r\n\1}`

PR #5655, opened 2025-11-25
STYLE: Replace iterations with FillBuffer calls in ImageIntensity tests

N-Dekker:
- Follow-up to pull request #5651

### DIFF
--- a/Modules/Filtering/ImageIntensity/test/itkVectorToRGBImageAdaptorTest.cxx
+++ b/Modules/Filtering/ImageIntensity/test/itkVectorToRGBImageAdaptorTest.cxx
@@ -81,12 +81,7 @@ itkVectorToRGBImageAdaptorTest(int, char *[])
   vector[2] = 1.4;
 
   // Initializing all the pixel in the image
-  it1.GoToBegin();
-  while (!it1.IsAtEnd())
-  {
-    it1.Set(vector);
-    ++it1;
-  }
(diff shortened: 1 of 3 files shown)


Modules/Filtering/ImageIntensity/test/itkVectorToRGBImageAdaptorTest.cxx
```diff
@@ -81,12 +81,7 @@ itkVectorToRGBImageAdaptorTest(int, char *[])
   vector[2] = 1.4;
 
   // Initializing all the pixel in the image
-  it1.GoToBegin();
-  while (!it1.IsAtEnd())
-  {
-    it1.Set(vector);
-    ++it1;
-  }
+  image->FillBuffer(vector);
```
N-Dekker, line 84: Note that there is another `while` loop a little bit further down here, that I did not touch with this pull request: https://github.com/InsightSoftwareConsortium/ITK/blob/07f5450a715c3f17e4fe0f140724f59cd02e5432/Modules/Filtering/ImageIntensity/test/itkVectorToRGBImageAdaptorTest.cxx#L107-L124 

I _think_ the original intention was to test that `RGBIteratorType it2(adaptor, adaptor->GetRequestedRegion())` works well with an adaptor, so then, replacing this with a `FillBuffer` call might not be according to the original intention 🤷

PR #5656, opened 2025-11-25
Use `ImageRegionRange` with `std::fill` in Filtering and Registration tests

N-Dekker:
Replaced hand-written `while` loops with more modern `std::fill` calls.

- Follow-up to pull request #5589

### DIFF
--- a/Modules/Filtering/ImageIntensity/test/itkMatrixIndexSelectionImageFilterTest.cxx
+++ b/Modules/Filtering/ImageIntensity/test/itkMatrixIndexSelectionImageFilterTest.cxx
@@ -16,10 +16,13 @@
  *
  *=========================================================================*/
 
-#include <fstream>
 #include "itkMatrixIndexSelectionImageFilter.h"
 #include "itkImageFileWriter.h"
+#include "itkImageRegionRange.h"
 #include "itkTestingMacros.h"
+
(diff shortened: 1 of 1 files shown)


Modules/Filtering/ImageFilterBase/test/itkMaskNeighborhoodOperatorImageFilterTest.cxx
```diff
@@ -63,24 +66,14 @@ itkMaskNeighborhoodOperatorImageFilterTest(int argc, char * argv[])
...
   {
-    itk::ImageRegionIterator<MaskImageType> it(mask1, region);
-    it.GoToBegin();
-    while (!it.IsAtEnd())
-    {
-      it.Set(1);
-      ++it;
-    }
+    const itk::ImageRegionRange<MaskImageType> imageRegionRange(*mask1, region);
+    std::fill(imageRegionRange.begin(), imageRegionRange.end(), 1);
```
N-Dekker, line 70: Note that `image->FillBuffer(fillValue)` is preferable to `std::fill(imageRegionRange.begin(), imageRegionRange.end(), fillValue)`, when the region is `image->GetBufferedRegion()`. But in this case, the region is smaller, so  `image->FillBuffer(fillValue)` cannot be used.

PR #5659, opened 2025-11-25
Simplify code by replacing calls to `begin()` and `rbegin()`

N-Dekker:
Two small style commits.

### DIFF
--- a/Modules/Core/Common/include/itkImageRegionRange.h
+++ b/Modules/Core/Common/include/itkImageRegionRange.h
@@ -397,7 +397,7 @@ public:
     endRegionIndex.back() += m_IterationRegionSize.back();
 
     OffsetType iterationOffset{ {} }; // Initialize offsets to 0, by aggregate initializer
-    *(iterationOffset.rbegin()) = m_IterationRegionSize.back();
+    iterationOffset.back() = m_IterationRegionSize.back();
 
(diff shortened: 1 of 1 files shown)


Modules/Registration/FEM/include/itkFEMRegistrationFilter.hxx
```diff
@@ -1508,7 +1508,7 @@ FEMRegistrationFilter<TMovingImage, TFixedImage, TFemObject>::PrintSelf(std::ost
   os << indent << "LandmarkArray: ";
   for (const auto & elem : m_LandmarkArray)
   {
-    os << indent.GetNextIndent() << "[" << &elem - &*(m_LandmarkArray.begin()) << "]: " << *elem << std::endl;
+    os << indent.GetNextIndent() << "[" << &elem - m_LandmarkArray.data() << "]: " << *elem << std::endl;
```
N-Dekker, line 1511: For the record, `SpatialObject::PrintSelf` does something similar, at https://github.com/InsightSoftwareConsortium/ITK/blob/4dcb75c155f81b0bd74d875d7ee13a8ca45c5ce3/Modules/Core/SpatialObjects/include/itkSpatialObject.hxx#L377

However, `&*(m_ChildrenList.begin())` cannot be replaced with `m_ChildrenList.data()`, because `m_ChildrenList` does not have a `data()` member function.

hjmjohnson (approved):

PR #5661, opened 2025-11-26
BUG: Fix CastImageFilter for VariableLengthVector

SimonRit:
The `Get` function of `VectorImage` iterators calls the `VariableLengthVector` constructor parameterized by a const pointer and a vector length (see `DefaultVectorPixelAccessor::Get`). The `m_Data` pointer of the constructed VariableLengthVector which was returned was then pointing to the returned pixel data. This led to an unexpected behavior in `CastImageFilter::DynamicThreadedGenerateDataDispatched` where modifications to the local variable
```
OutputPixelType value{ outputIt.Get() };
```
of `CastImageFilter::DynamicThreadedGenerateDataDispatched` were actually modifying the pixel of the image at the iterator position. The problem might occur elsewhere so it is safer to copy the data when a VariableLengthVector is constructed from a const pointer despite the potential loss in performances.

A `CastImageFilter` test has been added to illustrate the issue: when a `VectorImage` was casted to an `Image< Vector <`, the first pixel of the region was taking the value of the last pixel of the region. The test illustrated this by creating an image filled with 0 except the first pixel. The casted image was completely full of 0 before bug correction.

## PR Checklist

### DIFF
--- a/Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
+++ b/Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
@@ -145,12 +145,12 @@ CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatche
   ImageScanlineConstIterator inputIt(inputPtr, inputRegionForThread);
   ImageScanlineIterator      outputIt(outputPtr, outputRegionForThread);
 
-  OutputPixelType value{ outputIt.Get() };
   while (!inputIt.IsAtEnd())
   {
     while (!inputIt.IsAtEndOfLine())
     {
       const InputPixelType & inputPixel = inputIt.Get();
+      OutputPixelType        value{ outputIt.Get() };
       for (unsigned int k = 0; k < componentsPerPixel; ++k)
       {
         value[k] = static_cast<typename OutputPixelType::ValueType>(inputPixel[k]);
--- a/Modules/Filtering/ImageFilterBase/test/itkCastImageFilterTest.cxx
+++ b/Modules/Filtering/ImageFilterBase/test/itkCastImageFilterTest.cxx
@@ -400,6 +400,50 @@ TestVectorImageCast2()
 }
 
 
+bool
+TestVectorImageCast3()
+{
+  // This function casts an Image<Vector<float, 3>, 3>
+  // to a VectorImage<float, 3>
+  std::cout << "Casting from an Image<Vector<float,3>, 3> to VectorImage<float, 3> ... ";
+
+  using VectorFloatImageType = itk::Image<itk::Vector<float, 3>, 3>;
+  using FloatVectorImageType = itk::VectorImage<float, 3>;
+
+  // Create a 1x3 image of 2D vectors
+  auto image = VectorFloatImageType::New();
+
+  constexpr itk::Size<3>    size{ { 1, 1, 4 } };
+  const itk::ImageRegion<3> region{ size };
+  image->SetRegions(region);
+  image->AllocateInitialized();
+  VectorFloatImageType::PixelType vec{ { 1.3, 5.3, 6.4 } };
+  // Only the first pixel will be the vector (1.3, 5.3, 6.4)
+  image->SetPixel(itk::Index<3>(), vec);
+
+  using CastImageFilterType = itk::CastImageFilter<VectorFloatImageType, FloatVectorImageType>;
+  auto castImageFilter = CastImageFilterType::New();
+  castImageFilter->SetInput(image);
+  castImageFilter->SetNumberOfWorkUnits(1);
+  castImageFilter->Update();
+
+  bool success = true;
+  if (std::memcmp(castImageFilter->GetOutput()->GetBufferPointer(),
+                  image->GetBufferPointer(),
+                  sizeof(float) * size.CalculateProductOfElements()) == 0)
+  {
+    std::cout << "[PASSED]" << std::endl;
+  }
+  else
+  {
+    std::cout << "[FAILED]" << std::endl;
+    success = false;
+  }
+
+  return success;
+}
+
+
 int
 itkCastImageFilterTest(int, char *[])
 {
@@ -431,6 +475,7 @@ itkCastImageFilterTest(int, char *[])
   success &= TestCastFrom<double>();
   success &= TestVectorImageCast1();
   success &= TestVectorImageCast2();
+  success &= TestVectorImageCast3();
 
   std::cout << std::endl;
   if (!success)


Modules/Core/Common/include/itkVariableLengthVector.hxx
```diff
@@ -45,9 +45,20 @@ VariableLengthVector<TValue>::VariableLengthVector(ValueType * datain, unsigned
 template <typename TValue>
 VariableLengthVector<TValue>::VariableLengthVector(const ValueType * datain, unsigned int sz, bool LetArrayManageMemory)
   : m_LetArrayManageMemory(LetArrayManageMemory)
-  , m_Data(const_cast<ValueType *>(datain))
   , m_NumElements(sz)
-{}
+{
+  if (m_NumElements != 0)
+  {
+    m_Data = this->AllocateElements(m_NumElements);
```
hjmjohnson: How does this interact with "LetArrayManageMemory"?  It seems like m_LetArrayManageMemory should always be "true" for this case.

SimonRit: Good point, I oversaw this. So this is not a good fix then... I suggest modifying the iterator `Get` function then.

SimonRit:
I'm a bit lost and happy to get some advice. I hope the issue is clearly explained in the current commit log. The current proposal is not adequate as pointed out by [@hjmjohnson's comment](https://github.com/InsightSoftwareConsortium/ITK/pull/5661#discussion_r2566992705). I see two potential alternatives:
- modify `DefaultVectorPixelAccessor::Get` and create a local copy which would have a computional cost as pointed out in [this comment](https://github.com/InsightSoftwareConsortium/ITK/blob/main/Modules/Core/Common/include/itkDefaultVectorPixelAccessor.h#L80-L82).
- modify the `CastImageFilter` and replace [this line](https://github.com/InsightSoftwareConsortium/ITK/blob/main/Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx#L148)
```
OutputPixelType value{ outputIt.Get() };
``` 
with a copy
```
OutputPixelType value;
value = outputIt.Get();
```
I'm concerned that people actually do not expect a difference in behavior between the two so I would lean towards the first suggestion. Any thought?

SimonRit:
What I suggest is to revert @blowekamp commit 6a603279d6fd2c4a914a028e272a135c5bf45325... What do you think @blowekamp?

Modules/Filtering/ImageFilterBase/test/itkCastImageFilterTest.cxx
```diff
@@ -400,6 +400,52 @@ TestVectorImageCast2()
...
+  constexpr itk::Size<3>    size{ { 1, 1, 4 } };
+  const itk::ImageRegion<3> region{ size };
+  image->SetRegions(region);
+  image->AllocateInitialized();
+  VectorFloatImageType::PixelType vec;
+
+  // Only the first pixel will be the vector (1.3, 5.3, 6.4)
+  vec[0] = 1.3;
+  vec[1] = 5.3;
+  vec[2] = 6.4;
```
N-Dekker: Please consider doing `itk::Vector<float, 3> vec{ { 1.3, 5.3, 6.4 } };`

Modules/Filtering/ImageFilterBase/test/itkCastImageFilterTest.cxx
```diff
@@ -400,6 +400,52 @@ TestVectorImageCast2()
 }
 
 
+bool
+TestVectorImageCast3()
+{
+  // This function casts a VectorImage<float, 2>
+  // to an Image<Vector<unsigned char, 2>, 2>
```
N-Dekker: I guess `unsigned char` should be `float` here, right? And `2` should be `3`?

Modules/Filtering/ImageFilterBase/test/itkCastImageFilterTest.cxx
```diff
@@ -400,6 +400,52 @@ TestVectorImageCast2()
...
+  image->SetPixel(itk::Index<3>(), vec);
+
+  using CastImageFilterType = itk::CastImageFilter<VectorFloatImageType, FloatVectorImageType>;
+  auto castImageFilter = CastImageFilterType::New();
+  castImageFilter->SetInput(image);
+  castImageFilter->SetNumberOfWorkUnits(1);
+  castImageFilter->Update();
+
+  bool success = true;
+  if (std::memcmp(castImageFilter->GetOutput()->GetBufferPointer(), image->GetBufferPointer(), 4) == 0)
```
N-Dekker: Why are only the first `4` bytes compared? I guess it's because it is `sizeof(float)`, right? If so, please replace `4` with  `sizeof(float)`.

Modules/Filtering/ImageFilterBase/test/itkCastImageFilterTest.cxx
SimonRit: Sure addressed (for the new function only, revising the style of other functions in the test was not in the scope of this PR)

Modules/Filtering/ImageFilterBase/test/itkCastImageFilterTest.cxx
SimonRit: Mistake on my side, the whole thing is now compared with
```
std::memcmp(castImageFilter->GetOutput()->GetBufferPointer(),
                          image->GetBufferPointer(),
                          sizeof(float) * size.CalculateProductOfElements())`
```

Modules/Filtering/ImageFilterBase/test/itkCastImageFilterTest.cxx
SimonRit: Fixed, copy pasting error obviously. Thanks!

dzenanz:
The compilers should have gotten better over the last 10 years, they should be better at applying return value optimization. We could try reverting [this](https://github.com/InsightSoftwareConsortium/ITK/commit/6a603279d6fd2c4a914a028e272a135c5bf45325#diff-db56c92ca65ec1565694bc024edef4500cb13ffb63079d633ec640334e42e9e8L80-L82), and measure performance. ~100x performance difference would be painfully obvious. I doubt it will have such an impact.

blowekamp:
Looking at the CastImageFilter:
https://github.com/InsightSoftwareConsortium/ITK/blob/a9e3442997bcbfb2dc6361dc700c291964b720cf/Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx#L148

That first line use to copy of the VariableLengthVector breaking the reference to the data buffer, but some how it changes to retain a reference to the data buffer.



> The compilers should have gotten better over the last 10 years, they should be better at applying return value optimization. We could try reverting [this](https://github.com/InsightSoftwareConsortium/ITK/commit/6a603279d6fd2c4a914a028e272a135c5bf45325#diff-db56c92ca65ec1565694bc024edef4500cb13ffb63079d633ec640334e42e9e8L80-L82), and measure performance. ~100x performance difference would be painfully obvious. I doubt it will have such an impact.

Per pixel memory allocation with 50+ cores was historically rather slow. Likely small memory concurrent mallocs have improved, but iterative per-pixel memory allocation is still a HUGE performance issue.

blowekamp:
The change to add support to VariableLengthVector to the CastImageFilter occurred here:     87c3fb170816a3f4ec51320671ab475afb7bb399


We can quickly address the issue with the filter with the following change:
```
@@ -145,7 +146,7 @@ CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatche
   ImageScanlineConstIterator inputIt(inputPtr, inputRegionForThread);
   ImageScanlineIterator      outputIt(outputPtr, outputRegionForThread);
 
-  OutputPixelType value{ outputIt.Get() };
+  itk::NumericTraits<typename OutputPixelType::ValueType> value[componentsPerPixel];

```
Please let me know if you can add it to this PR, with the nice test or if I should create a separate PR.

If I have this correct conceptually, an ImageIterator's "Get" method should return a copy to the pixel, while the "Value" method should return a reference. In this case the PixelAccessor for the VectorImage is returning  a (modifiable) reference and it is getting modified in the filter. This fix will be more involved, if needed.

SimonRit:
> The change to add support to VariableLengthVector to the CastImageFilter occurred here: [87c3fb1](https://github.com/InsightSoftwareConsortium/ITK/commit/87c3fb170816a3f4ec51320671ab475afb7bb399)
> 
> We can quickly address the issue with the filter with the following change:
> 
> ```
> @@ -145,7 +146,7 @@ CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatche
>    ImageScanlineConstIterator inputIt(inputPtr, inputRegionForThread);
>    ImageScanlineIterator      outputIt(outputPtr, outputRegionForThread);
>  
> -  OutputPixelType value{ outputIt.Get() };
> +  itk::NumericTraits<typename OutputPixelType::ValueType> value[componentsPerPixel];
> ```
> 
> Please let me know if you can add it to this PR, with the nice test or if I should create a separate PR.
> 
> If I have this correct conceptually, an ImageIterator's "Get" method should return a copy to the pixel, while the "Value" method should return a reference. In this case the PixelAccessor for the VectorImage is returning a (modifiable) reference and it is getting modified in the filter. This fix will be more involved, if needed.

Done in the latest force-push. If you agree, let's merge this PR to fix CastImageFilter and then I'll open an issue with the minimal example above. BTW, `GradientRecursiveGaussianImageFilter` was failing with my previous "solution" so the filter might suffer from this iterator problem too.

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
```diff
@@ -145,7 +145,7 @@ CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatche
   ImageScanlineConstIterator inputIt(inputPtr, inputRegionForThread);
   ImageScanlineIterator      outputIt(outputPtr, outputRegionForThread);
 
-  OutputPixelType value{ outputIt.Get() };
+  itk::NumericTraits<typename OutputPixelType::ValueType> value[componentsPerPixel];
```
blowekamp: Sorry I posted the wrong diff, this is what I had working:
```
+  const auto &outputValue{outputIt.Get()};
+  auto value{ outputValue };
```
It was not an elegant way to force a copy, and I began looking at other constructs.

SimonRit:
@blowekamp Your suggestion was not compiling so I force-pushed a different version, see changes [here](https://github.com/InsightSoftwareConsortium/ITK/compare/b819ed6ffeb10d37ca8a1e14e7aeb5e734d6e6c2..380a7c19a12e564b34e6b1a443edc2c4979b11b0).

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
SimonRit: Sorry I only saw this after pushing yet another version. Let me know which version you prefer, they're all fine to me.

blowekamp:
@SimonRit I timed a couple options and this seems clean and perfrom the best:
```
@@ -145,12 +159,12 @@ CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatche
   ImageScanlineConstIterator inputIt(inputPtr, inputRegionForThread);
   ImageScanlineIterator      outputIt(outputPtr, outputRegionForThread);
 
-  OutputPixelType value{ outputIt.Get() };
   while (!inputIt.IsAtEnd())
   {
     while (!inputIt.IsAtEndOfLine())
     {
       const InputPixelType & inputPixel = inputIt.Get();
+      OutputPixelType value{outputIt.Get()};
       for (unsigned int k = 0; k < componentsPerPixel; ++k)
       {
         value[k] = static_cast<typename OutputPixelType::ValueType>(inputPixel[k]);
```

Sorry I didn't have a working diff to start.

blowekamp:
I updated the branch with the recommendation.

dzenanz (approved):

blowekamp:
We should consider porting the fix to the 5.4 branch.

PR #5662, opened 2025-11-27
STYLE: Do `auto size = SizeType::Filled(x)` in tests

N-Dekker:
Searched for code like:

    SizeType size;

    size[0] = x;
    size[1] = x;
    size[2] = x;

And replaced it with the equivalent `auto size = SizeType::Filled(x);`.

Using Notepad++, Replace in Files, doing:

  Find what: `^([ ]+)(.*SizeType)[ ]+size;[\r\n]+\1size\[0\] = (\w+);\r\n\1size\[1\] = \3;\r\n\1size\[2\] = \3;$`
  Find what: `^([ ]+)(.*SizeType)[ ]+size;[\r\n]+\1size\[0\] = (\w+);\r\n\1size\[1\] = \3;$`
  Replace with: `$1size = $2::Filled\($3\);`
  Filters: `itk*Test*.cxx`
  (*) Regular expression

PR #5664, opened 2025-11-28
STYLE: Use `SizeType size{ x, y, z }` brace-initialization in tests

N-Dekker:
Searched for code like:

    SizeType size;

    size[0] = x;
    size[1] = y;
    size[2] = z;

And replaced it with the equivalent `SizeType size{ x, y, z };`.

  Find what: `^([ ]+)(.*SizeType)[ ]+size;[\r\n]+\1size\[0\] = (\w+);\r\n\1size\[1\] = (\w+);\r\n\1size\[2\] = (\w+);$`
  Replace with: `$1$2 size{ $3, $4, $5 };`
  Find what: `^([ ]+)(.*SizeType)[ ]+size;[\r\n]+\1size\[0\] = (\w+);\r\n\1size\[1\] = (\w+);$`
  Replace with: `$1$2 size{ $3, $4 };`
  Filters: `itk*Test*.cxx`
  (*) Regular expression

- Follow-up to pull request #5662

### DIFF
--- a/Modules/Core/Common/test/itkImageLinearIteratorTest.cxx
+++ b/Modules/Core/Common/test/itkImageLinearIteratorTest.cxx
@@ -146,10 +146,7 @@ itkImageLinearIteratorTest(int, char *[])
     start[1] = 20;
     start[2] = 30;
 
-    ImageType::SizeType size;
-    size[0] = 2;
-    size[1] = 3;
-    size[2] = 4;
+    ImageType::SizeType size{ 2, 3, 4 };
 
     const ImageType::RegionType region{ start, size };
 
@@ -225,10 +222,7 @@ itkImageLinearIteratorTest(int, char *[])
     start[1] = 12;
     start[2] = 14;
 
-    ImageType::SizeType size;
-    size[0] = 11;
-    size[1] = 12;
-    size[2] = 13;
+    ImageType::SizeType size{ 11, 12, 13 };
 
     const ImageType::RegionType region{ start, size };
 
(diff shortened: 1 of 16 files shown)


Modules/Core/Common/test/itkImageLinearIteratorTest.cxx
```diff
@@ -146,10 +146,7 @@ itkImageLinearIteratorTest(int, char *[])
     start[1] = 20;
     start[2] = 30;
 
-    ImageType::SizeType size;
-    size[0] = 2;
-    size[1] = 3;
-    size[2] = 4;
+    ImageType::SizeType size{ 2, 3, 4 };
```
N-Dekker, line 149: We could also add an extra pair of braces:

    ImageType::SizeType size{ { 2, 3, 4 } };

Or use `=`:

    ImageType::SizeType size = { 2, 3, 4 };

Or both (extra braces and `=`):

    ImageType::SizeType size = { { 2, 3, 4 } };

Is all fine 🤷

hjmjohnson: @N-Dekker  I prefer what you have done.

N-Dekker: Thanks Hans! For `itk::Size`, a single pair of braces works fine, so I'm proposing to do it that way, indeed:
```cpp
ImageType::SizeType size{ 2, 3, 4 };
```

Note that for `itk::Vector`, an extra pair of braces or parentheses would still be necessary:

```cpp
itk::Vector<double, 3> itkVector1{ { 1.0, 2.0, 3.0 } }; // OK
itk::Vector<double, 3> itkVector2({ 1.0, 2.0, 3.0 }); // OK

itk::Vector<double, 3> itkVector3{ 1.0, 2.0, 3.0 }; // Compile error! (Using MSVC 2022.)
```

That's because `itk::Vector<T, N>` is not a simple struct like  `itk::Size<N>`. `itkVector1` and `itkVector2` are initialized by calling an explicit constructor of `itk::Vector<T, N>`.

Issue #5666, opened 2025-11-28
DataObjectError print to output stream should also print info about its DataObject


N-Dekker:
### Description

When an error message appears, saying something like `Requested region is (at least partially) outside the largest possible region`, it would be very helpful to also print some information about those regions. Or more in general, print some information about the data object for which the error occurred.

Looking at the implementation of `DataObjectError::PrintSelf`, it appears that the _intention_ is indeed to print such information, as it does `m_DataObject->PrintSelf`: https://github.com/InsightSoftwareConsortium/ITK/blob/a9e3442997bcbfb2dc6361dc700c291964b720cf/Modules/Core/Common/src/itkDataObject.cxx#L73-L87

However, `DataObjectError::PrintSelf` is never called! When someone prints a `DataObjectError`, for example by `std::cerr << error`, it will call `operator<<(std::ostream &, const ExceptionObject &)`, which calls `Print`, not `PrintSelf`!!! Here

https://github.com/InsightSoftwareConsortium/ITK/blob/a9e3442997bcbfb2dc6361dc700c291964b720cf/Modules/Core/Common/include/itkExceptionObject.h#L152-L155

### Steps to Reproduce

```cpp
  const auto           image = itk::Image<int>::New();
  itk::DataObjectError error(__FILE__, __LINE__);
  error.SetDescription("My description");
  error.SetDataObject(image);
  std::cerr << error << std::endl;
```
### Expected output

```
itk::DataObjectError (0000004659AFEE28)
File: path/to/source/file
Line: 42
Description: My description

Data object:
  RTTI typeinfo:   class itk::Image<int,2>
  Reference Count: 1
  Modified Time: 1
  Debug: Off
  Object Name:
  Observers:
    none
  Source: (none)
  Source output name: (none)
  Release Data: Off
  Data Released: False
  Global Release Data: Off
  PipelineMTime: 0
  UpdateMTime: 0
  RealTimeStamp: 0 seconds
  LargestPossibleRegion:
    Dimension: 2
    Index: [0, 0]
    Size: [0, 0]
  BufferedRegion:
    Dimension: 2
    Index: [0, 0]
    Size: [0, 0]
  RequestedRegion:
    Dimension: 2
    Index: [0, 0]
    Size: [0, 0]
  etc. etc....

```
### Actual output

```
itk::DataObjectError (0000004659AFEE28)
File: path/to/source/file
Line: 42
Description: My description
```
So no information at all about the data object!

[...]

PR #5667, opened 2025-11-28
BUG: Let DataObjectError override `ExceptionObject::Print`

N-Dekker:
Instead of introducing a new virtual member function, `DataObjectError::PrintSelf`.

It is essential that DataObjectError overrides `ExceptionObject::Print`, because that is the member function that is called, when doing `std::cerr << error` for a DataObjectError.

- Fixes issue #5666


dzenanz (approved):
LGTM. Thinking of how this went on for so long, I realized I never used `DataObjectError`. I guess most code uses `itk::ExceptionObject`.

Issue #5668, opened 2025-11-29
Undefined behavior of iterators Get for VariableLengthVector


SimonRit:
### Description

The `Get` function of an iterator on an `itk::Image` of `itk::VariableLengthVector` returns an `itk::VariableLengthVector` which data points to a pixel of the `itk::Image`. This is because the `DefaultVectorPixelAccessor::Get` calls the `VariableLengthVector` constructor parameterized by a `const pointer`. This leads to an unexpected behavior for the user.

### Steps to Reproduce

This is a minimal example reproducing the issue:
```
#include <itkVectorImage.h>
#include <itkImageRegionIterator.h>

int
main(int argc, char * argv[])
{
  using ImageType = itk::VectorImage<float,2>;
  auto img = ImageType::New();
  img->SetVectorLength(3);
  ImageType::SizeType size = { { 2, 2 } };
  img->SetRegions(size);
  img->AllocateInitialized();
  itk::ImageRegionIterator<ImageType> it(img, img->GetLargestPossibleRegion());
  auto v { it.Get() };
  float c = 0.;
  while(!it.IsAtEnd())
  {
    v.Fill(c++);
    it.Set(v);
    ++it;
  }
  it.GoToBegin();
  while(!it.IsAtEnd())
  {
    std::cout << it.Get() << std::endl;
    ++it;
  }
  return EXIT_SUCCESS;
}
```

### Expected behavior

It should display
```
[0, 0, 0]
[1, 1, 1]
[2, 2, 2]
[3, 3, 3]
```

### Actual behavior

It actually displays 
```
[3, 3, 3]
[1, 1, 1]
[2, 2, 2]
[3, 3, 3]
```

### Reproducibility

100%

### Versions

All versions since 6a603279d6fd2c4a914a028e272a135c5bf45325.

### Environment

All OSs.

PR #5669, opened 2025-12-01
Performance comparison of iterators and ranges to copy vector images

blowekamp:
## PR Checklist

### DIFF
--- a/Modules/Filtering/ImageFilterBase/test/itkImageIterationPerformanceTest.cxx
+++ b/Modules/Filtering/ImageFilterBase/test/itkImageIterationPerformanceTest.cxx
@@ -1,0 +1,416 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         https://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+#include <iostream>
+#include "itkImage.h"
+#include "itkVectorImage.h"
+#include "itkVector.h"
+#include "itkRGBPixel.h"
+#include "itkImageRegionRange.h"
+#include "itkImageScanlineIterator.h"
+#include "itkImageScanlineConstIterator.h"
+#include "itkImageRegionIteratorWithIndex.h"
+#include "itkImageRegionConstIterator.h"
+#include "itkTimeProbe.h"
+#include "itkNumericTraits.h"
+#include "itkTestingMacros.h"
+#include <iomanip>
+
+
+// Helper function to initialize an image with random values
+template <typename TImage>
+typename TImage::Pointer
+CreateAndInitializeImage(const typename TImage::SizeType & size, unsigned int numberOfComponentsPerPixel = 0)
+{
+  auto                        image = TImage::New();
+  typename TImage::RegionType region{ size };
+  image->SetRegions(region);
+  if (numberOfComponentsPerPixel > 0)
+  {
+    image->SetNumberOfComponentsPerPixel(numberOfComponentsPerPixel);
+  }
+  image->Allocate();
+
+  // Initialize with simple pattern (pixel index-based)
+  using PixelType = typename TImage::PixelType;
+  unsigned int count = 0;
+
+  const unsigned int length = image->GetNumberOfComponentsPerPixel();
+
+  itk::ImageRegionIteratorWithIndex<TImage> it(image, region);
+  for (it.GoToBegin(); !it.IsAtEnd(); ++it)
+  {
+    PixelType pixel{ it.Get() };
+    for (unsigned int k = 0; k < length; ++k)
+    {
+      pixel[k] = static_cast<typename itk::NumericTraits<PixelType>::ValueType>(count + k);
+    }
+    it.Set(pixel);
+    ++count;
+  }
+
+  return image;
+}
+
+
+// Method 1: ImageScanlineIterator approach
+template <typename TInputImage, typename TOutputImage>
+void
+CopyScanlineIterator(const TInputImage * inputPtr, TOutputImage * outputPtr)
+{
+  using InputPixelType = typename TInputImage::PixelType;
+  using OutputPixelType = typename TOutputImage::PixelType;
+  using ImageScanlineConstIterator = itk::ImageScanlineConstIterator<TInputImage>;
+  using ImageScanlineIterator = itk::ImageScanlineIterator<TOutputImage>;
+
+  const typename TOutputImage::RegionType outputRegion = outputPtr->GetRequestedRegion();
+  typename TInputImage::RegionType        inputRegion = outputRegion;
+
+  ImageScanlineConstIterator inputIt(inputPtr, inputRegion);
+  ImageScanlineIterator      outputIt(outputPtr, outputRegion);
+
+  const unsigned int componentsPerPixel = inputPtr->GetNumberOfComponentsPerPixel();
+  while (!inputIt.IsAtEnd())
+  {
+    while (!inputIt.IsAtEndOfLine())
+    {
+      const InputPixelType & inputPixel = inputIt.Get();
+      OutputPixelType        value(outputIt.Get());
+      for (unsigned int k = 0; k < componentsPerPixel; ++k)
+      {
+        value[k] = static_cast<typename OutputPixelType::ValueType>(inputPixel[k]);
+      }
+      outputIt.Set(value);
+
+      ++inputIt;
+      ++outputIt;
+    }
+    inputIt.NextLine();
+    outputIt.NextLine();
+  }
+}
+
+
+// Method 1b: ImageScanlineIterator approach using NumericTraits::GetLength()
+template <typename TInputImage, typename TOutputImage>
+void
+CopyScanlineIteratorNumericTraits(const TInputImage * inputPtr, TOutputImage * outputPtr)
+{
+  using InputPixelType = typename TInputImage::PixelType;
+  using OutputPixelType = typename TOutputImage::PixelType;
+  using ImageScanlineConstIterator = itk::ImageScanlineConstIterator<TInputImage>;
+  using ImageScanlineIterator = itk::ImageScanlineIterator<TOutputImage>;
+
+  const typename TOutputImage::RegionType outputRegion = outputPtr->GetRequestedRegion();
+  typename TInputImage::RegionType        inputRegion = outputRegion;
+
+  ImageScanlineConstIterator inputIt(inputPtr, inputRegion);
+  ImageScanlineIterator      outputIt(outputPtr, outputRegion);
+
+  unsigned int componentsPerPixel = itk::NumericTraits<OutputPixelType>::GetLength(outputIt.Get());
+  while (!inputIt.IsAtEnd())
+  {
+    while (!inputIt.IsAtEndOfLine())
+    {
+      const InputPixelType & inputPixel = inputIt.Get();
+
+      OutputPixelType value{ outputIt.Get() };
+      for (unsigned int k = 0; k < componentsPerPixel; ++k)
+      {
+        value[k] = static_cast<typename OutputPixelType::ValueType>(inputPixel[k]);
+      }
+      outputIt.Set(value);
+
+      ++inputIt;
+      ++outputIt;
+    }
+    inputIt.NextLine();
+    outputIt.NextLine();
+  }
+}
+
+
+// Method 2: ImageRegionRange approach
+template <typename TInputImage, typename TOutputImage>
+void
+CopyImageRegionRange(const TInputImage * inputPtr, TOutputImage * outputPtr)
+{
+  using InputPixelType = typename TInputImage::PixelType;
+  using OutputPixelType = typename TOutputImage::PixelType;
+
+  const typename TOutputImage::RegionType outputRegion = outputPtr->GetRequestedRegion();
+  typename TInputImage::RegionType        inputRegion = outputRegion;
+
+  auto inputRange = itk::ImageRegionRange<const TInputImage>(*inputPtr, inputRegion);
+  auto outputRange = itk::ImageRegionRange<TOutputImage>(*outputPtr, outputRegion);
+
+  auto       inputIt = inputRange.begin();
+  auto       outputIt = outputRange.begin();
+  const auto inputEnd = inputRange.end();
+
+  const unsigned int componentsPerPixel = inputPtr->GetNumberOfComponentsPerPixel();
+  while (inputIt != inputEnd)
+  {
+    const InputPixelType & inputPixel = *inputIt;
+    OutputPixelType        outputPixel{ *outputIt };
+    for (unsigned int k = 0; k < componentsPerPixel; ++k)
+    {
+      outputPixel[k] = static_cast<typename OutputPixelType::ValueType>(inputPixel[k]);
+    }
+    *outputIt = outputPixel;
+
+    ++inputIt;
+    ++outputIt;
+  }
+}
+
+
+// Method 2b: ImageRegionRange approach using NumericTraits::GetLength()
+template <typename TInputImage, typename TOutputImage>
+void
+CopyImageRegionRangeNumericTraits(const TInputImage * inputPtr, TOutputImage * outputPtr)
+{
+  using InputPixelType = typename TInputImage::PixelType;
+  using OutputPixelType = typename TOutputImage::PixelType;
+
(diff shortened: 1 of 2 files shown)


N-Dekker:
Thanks @blowekamp, interesting, I'll have a look!

In general, I wonder, should a performance benchmark be part of the regular unit test set? Or should it have its own project? Specifically, should your benchmark be a part of https://github.com/InsightSoftwareConsortium/ITKPerformanceBenchmarking ? 

If your benchmark would stay here in InsightSoftwareConsortium/ITK I would suggest to make it a GoogleTest! (Although maybe a benchmark suite like https://github.com/google/benchmark would even be nicer.)

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
```diff
@@ -163,6 +167,32 @@ CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatche
...
+  auto       inputIt = inputRange.begin();
+  auto       outputIt = outputRange.begin();
+  const auto inputEnd = inputRange.end();
+
+
+  OutputPixelType    outputPixel{ *outputIt };
+  const unsigned int componentsPerPixel = NumericTraits<OutputPixelType>::GetLength(outputPixel);
+  while (inputIt != inputEnd)
+  {
+    InputPixelType inputPixel = *inputIt;
```
N-Dekker, line 184: Why not `const InputPixelType & inputPixel`? (The original code also has `const InputPixelType & inputPixel`.)

Modules/Filtering/ImageFilterBase/test/itkImageIterationPerformanceTest.cxx
```diff
@@ -0,0 +1,382 @@
...
+  auto outputRange = itk::ImageRegionRange<TOutputImage>(*outputPtr, outputRegion);
+
+  auto       inputIt = inputRange.begin();
+  auto       outputIt = outputRange.begin();
+  const auto inputEnd = inputRange.end();
+
+  const unsigned int componentsPerPixel = inputPtr->GetNumberOfComponentsPerPixel();
+  while (inputIt != inputEnd)
+  {
+    InputPixelType  inputPixel = *inputIt;
```
N-Dekker: Again, please consider `const InputPixelType & inputPixel` here!

Modules/Filtering/ImageFilterBase/test/itkImageIterationPerformanceTest.cxx
```diff
@@ -0,0 +1,382 @@
...
+    OutputPixelType outputPixel(componentsPerPixel);
+    for (unsigned int k = 0; k < componentsPerPixel; ++k)
+    {
+      outputPixel[k] = static_cast<typename OutputPixelType::ValueType>(inputPixel[k]);
+    }
+    *outputIt = outputPixel;
```
N-Dekker, line 175: Please consider "moving" `outputPixel` into `*outputIt`. This is especially relevant when OutputPixelType is a VariableLengthVector, which has an expensive copy-constructor and a fast move-constructor.

blowekamp: Why? The outputIterator points to the memory in the pixel buffer. Assigning to the iterator should use the memory in the buffer and not allocate on a per pixel basis. Are there temporaries that are creating in this assignment operation?

N-Dekker: OK, sorry, my bad! Indeed, in this particular case (when copying pixels between `VectorImage` instances), move semantics doesn't really help!

Modules/Filtering/ImageFilterBase/test/itkImageIterationPerformanceTest.cxx
```diff
@@ -0,0 +1,382 @@
...
+
+  auto       inputIt = inputRange.begin();
+  auto       outputIt = outputRange.begin();
+  const auto inputEnd = inputRange.end();
+
+  const unsigned int componentsPerPixel = inputPtr->GetNumberOfComponentsPerPixel();
+  while (inputIt != inputEnd)
+  {
+    InputPixelType  inputPixel = *inputIt;
+    OutputPixelType outputPixel(componentsPerPixel);
```
N-Dekker: `OutputPixelType outputPixel(componentsPerPixel)` is expensive for `VariableLengthVector`. Please consider moving this declaration outside the `while (inputIt != inputEnd)` loop (just like you proposed for CastImageFilter).

Modules/Filtering/ImageFilterBase/test/itkImageIterationPerformanceTest.cxx
```diff
@@ -0,0 +1,382 @@
...
+  std::cout << "Results match:         " << (identical ? "YES" : "NO") << std::endl;
+}
+
+
+int
+itkImageIterationPerformanceTest(int, char *[])
+{
+  std::cout << "CTEST_FULL_OUTPUT" << std::endl;
+  std::cout << "ITK Image Iteration Performance Test" << std::endl;
+  std::cout << "=====================================" << std::endl;
```
N-Dekker: If you want this to become a regular unit test, please limit the amount of output! The ITK tests already produce an excessive amount of output, in my opinion.

Modules/Filtering/ImageFilterBase/test/itkImageIterationPerformanceTest.cxx
```diff
@@ -0,0 +1,382 @@
...
+
+  // Test different image sizes and types
+  constexpr unsigned int Dimension = 3;
```
N-Dekker, line 369: Please consider testing 2D as well. Maybe do one small 2D and 1 large 3D.

Modules/Filtering/ImageFilterBase/test/itkImageIterationPerformanceTest.cxx
```diff
@@ -0,0 +1,382 @@
...
+
+  // Small image for quick testing
+  itk::Size<Dimension> smallSize;
+  smallSize.Fill(64);
```
N-Dekker: Please: `constexpr auto smallSize = itk::Size<Dimension>::Filled(64);`

Modules/Filtering/ImageFilterBase/test/itkImageIterationPerformanceTest.cxx
```diff
@@ -0,0 +1,382 @@
...
+
+  // Medium image
+  itk::Size<Dimension> mediumSize;
+  mediumSize.Fill(256);
```
N-Dekker: Please: `constexpr auto mediumSize = itk::Size<Dimension>::Filled(256);`

Modules/Filtering/ImageFilterBase/test/itkImageIterationPerformanceTest.cxx
```diff
@@ -0,0 +1,382 @@
...
+
+  // Large image for realistic performance testing
+  itk::Size<Dimension> largeSize;
+  largeSize.Fill(512);
```
N-Dekker: Please: `constexpr auto largeSize = itk::Size<Dimension>::Filled(512);`

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
```diff
@@ -163,6 +167,32 @@ CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatche
...
+
+  auto inputRange = ImageRegionRange<const TInputImage>(*inputPtr, inputRegionForThread);
+  auto outputRange = ImageRegionRange<TOutputImage>(*outputPtr, outputRegionForThread);
+
+  auto       inputIt = inputRange.begin();
+  auto       outputIt = outputRange.begin();
+  const auto inputEnd = inputRange.end();
+
+
+  OutputPixelType    outputPixel{ *outputIt };
```
N-Dekker, line 180: At this point, the value of `*outputIt` may not yet be initialized, right? The output pixel values may still be uninitialized at this point. If so, better just declare `OutputPixelType outputPixel;` without retrieving the value from `*outputIt`.

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
blowekamp: Yes. You reference code times better on my system.

Modules/Filtering/ImageFilterBase/test/itkImageIterationPerformanceTest.cxx
N-Dekker: I just ran the last part of your benchmark:

```
--- Test Suite 4: VectorImage<float> to VectorImage<double> ---

=== Large Image (512^3) ===
Image size: 512x512x512
Scanline Iterator:     243.729 Mpixels/sec (0.550685 seconds)
ImageRegionRange:      35.8598 Mpixels/sec (3.74284 seconds)
Scanline NumericTraits:245.443 Mpixels/sec (0.546839 seconds)
Range NumericTraits:   290.972 Mpixels/sec (0.461273 seconds)
Results match:         YES
```

That looks very bad for ImageRegionRange. However, after moving the declaration of `outputPixel` outside the `while (inputIt != inputEnd)` loop, it said:

```
=== Large Image (512^3) ===
Image size: 512x512x512
Scanline Iterator:     246.817 Mpixels/sec (0.543794 seconds)
ImageRegionRange:      175.812 Mpixels/sec (0.763417 seconds)
Scanline NumericTraits:246.84 Mpixels/sec (0.543744 seconds)
Range NumericTraits:   293.25 Mpixels/sec (0.45769 seconds)
Results match:         YES
```

So now ImageRegionRange still doesn't seem as fast as Scanline Iterator, but it's only 0.76 versus 0.54 seconds now.

blowekamp: Yes, it is expensive. For this case in my posted results it's the reason why it's 10x slower when then the output is a VectorImage.

Interestingly, when `OutputPixelType outputPixel(componentsPerPixel)` is outside the loop it's ~2x slower.  AND if it's `OutputPixelType outputPixel{ *outputIt};` the same issue of unexpected modified output as reported in #5668 occurs.

Modules/Filtering/ImageFilterBase/test/itkImageIterationPerformanceTest.cxx
blowekamp: The additional full output is intentional so that the performance results from the CI can be seen.

I don't know what the long term utility of this performance comparison will be. Currently some systems ran out of memory. So a couple tweaks are needed for that... likely command line the size of the image and reduce number of images in memory.

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
blowekamp: You are correct that the values in outputPixel are uninitialized. However, for the VectorImage cast, the outputPixel will point to the data in the output Image's buffer. So this removed the need to allocation memory, reduced the memory locations used ( atleast for variable length vectors. )

dzenanz (commented):
As image region range seems more performant than scanline iterator, do you mind issuing a separate PR which updates the cast image filter with the refactoring?

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
blowekamp: This is also the "BUG" cast where the reference to the first pixel is maintained and will be assigned all values in the range.

dzenanz:
Correspondingly, `CastImageFilter` should prefer the more performant `Value()` method, if we need to degrade the performance of `Get()` to avoid the buggy behavior.

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
SimonRit: Are you implying that #5661 didn't fix a bug? To be honest, the time it took me to understand the origin of this bug is a clear indication to me that this is not the expected behavior of the Get function of an iterator. And I guess this is not consistent with how it behaves with scalar pixel.

blowekamp: The merged patch fixed the issue in the case filter. But this new code block with the ImageRangeIterator exhibits the same behavior as the original CastImageFilter.

Some tricks have been needed to get operations on the VectorImageFilter to perform reasonably. But they have been this way for over 10 years, and crippling filters by making them run 10x slower may not be a reasonable option.

SimonRit: I don't think it's reasonable too and I don't have the solution. But it does not mean it's not a bug. BTW, if you implement the (slow) version, a recursive Gaussian test fails which probably indicates a bug there too.

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
```diff
@@ -163,6 +167,32 @@ CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatche
     inputIt.NextLine();
     outputIt.NextLine();
   }
+#else
+
+  auto inputRange = ImageRegionRange<const TInputImage>(*inputPtr, inputRegionForThread);
+  auto outputRange = ImageRegionRange<TOutputImage>(*outputPtr, outputRegionForThread);
```
N-Dekker, line 173: Nitpick: technically this is OK, of course, but in ITK we don't really do AAA ("auto almost always") everywhere. (We just follow [clang-tidy modernize-use-auto](https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-auto.html), but that's still not AAA everywhere). So I would just do: 

```cpp
    ImageRegionRange<const TInputImage> inputRange(*inputPtr, inputRegionForThread);
    ImageRegionRange<TOutputImage>      outputRange(*outputPtr, outputRegionForThread);
```

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
```diff
@@ -139,18 +140,21 @@ CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatche
 
   this->CallCopyOutputRegionToInputRegion(inputRegionForThread, outputRegionForThread);
 
-  const unsigned int componentsPerPixel = outputPtr->GetNumberOfComponentsPerPixel();
 
+#if 0
```
N-Dekker, line 144: I guess the `#if 0` code may be removed when this PR is ready, right? Otherwise `if constexpr (false)` might be nicer... 😺 But I would prefer the code to just be removed, eventually.

Modules/Filtering/ImageFilterBase/test/itkImageIterationPerformanceTest.cxx
```diff
@@ -0,0 +1,416 @@
...
+    size = std::stoi(argv[1]);
+  }
+  else
+  {
+    if (argc < 2)
+    {
+      std::cerr << "Missing parameters." << std::endl;
+      std::cerr << "Usage: " << itkNameOfTestExecutableMacro(argv) << " imageSize" << std::endl;
+      return EXIT_FAILURE;
+    }
```
N-Dekker, line 384: When I run ITKImageFilterBaseTestDriver, it now says:
```
Available tests:
  0. itkNeighborhoodOperatorImageFilterTest
  1. itkImageToImageFilterTest
  2. itkVectorNeighborhoodOperatorImageFilterTest
  3. itkMaskNeighborhoodOperatorImageFilterTest
  4. itkCastImageFilterTest
  5. itkImageIterationPerformanceTest
To run a test, enter the test number: 5
CTEST_FULL_OUTPUT
Missing parameters.
Usage: <itkImageIterationPerformanceTest executable> imageSize

D:\X\Bin\ITK\bin\Release\ITKImageFilterBaseTestDriver.exe (process 46316) exited with code 1 (0x1).
Press any key to close this window . . .
```

Please consider a default `imageSize`, so that the test can be run that way as well...  🙏 (Of course I like GoogleTest even better!)

blowekamp: I'm just running `./bin/ITKImageFilterBaseTestDriver itkImageIterationPerformanceTest 512` easy peasy.

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
blowekamp: I made PR #5670 with just the change to the cast filter. 

This change below fails the testing due to the aliasing/reference of the first outputPixel, just like #5668. I'll update that issue with the Image Range iterator to hopefully focus the discussion of what the correct behavior should be.

Modules/Filtering/ImageFilterBase/test/itkImageIterationPerformanceTest.cxx
N-Dekker: OK, easy peasy. 👍 Still I think a default imageSize would make it even easier peasier  😄

blowekamp:
The benchmark code is being moved to InsightSoftwareConsortium/ITKPerformanceBenchmarking#105

And change to the cast image filter was completed in  #5670

PR #5670, opened 2025-12-03
PERF: Use tuned ImageRangeRegion copy in CastImageFilter

blowekamp:
Based on performance testing, across converting between Image of Vector and VectorImage, this loop was the best performing. Key features of the improved performane:
 - Uses NumericTraits::GetLength over Image::GetNumberOfComponents, the former may be constant, while the latter is virtual
 - Uses const InputPixelType & inputPixel
 - OutputPixelType value{ outputIt.Get() } initialized to a reference in the output image bufffer and does not perform memory allocation for variable length vectors.

## PR Checklist

### DIFF
--- a/Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
+++ b/Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
@@ -20,6 +20,7 @@
 
 #include "itkProgressReporter.h"
 #include "itkImageAlgorithm.h"
+#include "itkImageRegionRange.h"
 
 namespace itk
 {
@@ -139,29 +140,30 @@ CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatche
 
   this->CallCopyOutputRegionToInputRegion(inputRegionForThread, outputRegionForThread);
 
-  const unsigned int componentsPerPixel = outputPtr->GetNumberOfComponentsPerPixel();
+  ImageRegionRange<const TInputImage> inputRange(*inputPtr, inputRegionForThread);
+  ImageRegionRange<TOutputImage>      outputRange(*outputPtr, outputRegionForThread);
 
-  // Define the iterators
-  ImageScanlineConstIterator inputIt(inputPtr, inputRegionForThread);
-  ImageScanlineIterator      outputIt(outputPtr, outputRegionForThread);
+  auto       inputIt = inputRange.begin();
+  auto       outputIt = outputRange.begin();
+  const auto inputEnd = inputRange.end();
 
-  while (!inputIt.IsAtEnd())
+  // Note: This loop has been timed for performance with conversions between image of vectors and VectorImages and other
+  // combinations. The following was evaluated to be the best performance usage of iterators. Important considerations:
+  //  - Usage of NumericTraits::GetLength() is sometimes consant vs virutal method GetNumberOfComponentsPerPixel()
+  //  - The construction of inputPixel and outputPixel for VectorImages both reference the internal buffer and don't
+  //  require memory allocations.
+  const unsigned int componentsPerPixel = itk::NumericTraits<OutputPixelType>::GetLength(*outputIt);
+  while (inputIt != inputEnd)
   {
-    while (!inputIt.IsAtEndOfLine())
+    const InputPixelType & inputPixel = *inputIt;
+    OutputPixelType        outputPixel{ *outputIt };
+    for (unsigned int k = 0; k < componentsPerPixel; ++k)
     {
-      const InputPixelType & inputPixel = inputIt.Get();
-      OutputPixelType        value{ outputIt.Get() };
-      for (unsigned int k = 0; k < componentsPerPixel; ++k)
-      {
-        value[k] = static_cast<typename OutputPixelType::ValueType>(inputPixel[k]);
-      }
-      outputIt.Set(value);
-
-      ++inputIt;
-      ++outputIt;
+      outputPixel[k] = static_cast<typename OutputPixelType::ValueType>(inputPixel[k]);
     }
-    inputIt.NextLine();
-    outputIt.NextLine();
+    *outputIt = outputPixel;
+    ++inputIt;
+    ++outputIt;
   }
 }
 


Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
```diff
@@ -139,29 +139,28 @@ CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatche
...
+
+  // Note: this loop has been timed for performance with converstions betwee image of vectors and VectorImages and other
+  // combinations. The following is the most efficient usage of iterators. Important considerations are the some times
+  // constant GetLength, the const & to input pixel and free from memory allocations with variable length vectors.
+  const unsigned int componentsPerPixel = itk::NumericTraits<OutputPixelType>::GetLength(*outputIt);
+  while (inputIt != inputEnd)
   {
-    while (!inputIt.IsAtEndOfLine())
+    const InputPixelType & inputPixel = *inputIt;
+    OutputPixelType        outputPixel{ *outputIt };
```
SimonRit, line 159: One alternative to this is to replace this by
```
OutputPixelType outputPixel;
outputPixel = *outputIt;
```
outside the while loop. Does it improve performance?

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
```diff
@@ -139,29 +139,28 @@ CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatche
...
-
-  while (!inputIt.IsAtEnd())
+  auto inputRange = itk::ImageRegionRange<const TInputImage>(*inputPtr, inputRegionForThread);
+  auto outputRange = itk::ImageRegionRange<TOutputImage>(*outputPtr, outputRegionForThread);
+
+  auto       inputIt = inputRange.begin();
+  auto       outputIt = outputRange.begin();
+  const auto inputEnd = inputRange.end();
+
+  // Note: this loop has been timed for performance with converstions betwee image of vectors and VectorImages and other
```
SimonRit: Some typos:
converstions -> conversions
betwee -> between

SimonRit (approved):
LGTM, just typos and a question.

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
hjmjohnson: I am probably missing some larger context behavior, but why do we need a temporary local variable for outputPixel?  It seems that line 164 could be removed, and line 159 could be an alias to the current memory.
```c++
OutputPixelType & outputPixel  = *outputIt;
```

blowekamp: I encourage both of you to compile and try experiments with  PR #5669

@SimonRit Outside the loop it is a little slower ~2x for vector pixel output. It has odd statements to force a copy.
@hjmjohnson That does not compile.

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
blowekamp: rewrote doc.

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
N-Dekker: @hjmjohnson FYI, for VectorImage, `*outputIt` returns a _proxy_ to the current output pixel. Hope that clarifies why `OutputPixelType & outputPixel  = *outputIt` won't compile for VectorImage!

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
```diff
@@ -139,29 +140,30 @@ CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatche
 
   this->CallCopyOutputRegionToInputRegion(inputRegionForThread, outputRegionForThread);
```
N-Dekker, line 141: For my understanding, do inputRegionForThread and outputRegionForThread always have exactly the same number of pixels (as returned by `ImageRegion::GetNumberOfPixels()`), after calling `CallCopyOutputRegionToInputRegion`?

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
```diff
@@ -139,29 +140,30 @@ CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatche
...
+  // Note: This loop has been timed for performance with conversions between image of vectors and VectorImages and other
+  // combinations. The following was evaluated to be the best performance usage of iterators. Important considerations:
+  //  - Usage of NumericTraits::GetLength() is sometimes consant vs virutal method GetNumberOfComponentsPerPixel()
+  //  - The construction of inputPixel and outputPixel for VectorImages both reference the internal buffer and don't
+  //  require memory allocations.
+  const unsigned int componentsPerPixel = itk::NumericTraits<OutputPixelType>::GetLength(*outputIt);
+  while (inputIt != inputEnd)
   {
-    while (!inputIt.IsAtEndOfLine())
+    const InputPixelType & inputPixel = *inputIt;
```
N-Dekker, line 158: You can make it a beautiful range-based for loop, if you like  😃:

```cpp
      for (const InputPixelType & inputPixel : inputRange)
```

Then you may remove the variables `inputIt` and `inputEnd`.

blowekamp: How close are we to being able to do:

`for ( auto &[inputPixel, outputPixel] : zip(inputRange, outputRange))`

😈

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
dzenanz: If the filter has some "radius" (frequent), then the input region is slightly larger (I believe). Each filter decides how to increase the region during propagation from output to input.

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
```diff
@@ -139,29 +140,30 @@ CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatche
...
-      }
-      outputIt.Set(value);
-
-      ++inputIt;
-      ++outputIt;
+      outputPixel[k] = static_cast<typename OutputPixelType::ValueType>(inputPixel[k]);
     }
-    inputIt.NextLine();
-    outputIt.NextLine();
+    *outputIt = outputPixel;
```
dzenanz, line 164: Do we need this statement, given that `outputPixel` is a reference to `*outputIt`?

blowekamp: OutputPixel is only a reference for the VariableLengthVector, there is a good change the compiler is figuring this out the assignment get's optimized away. For fixed vectors ( or scalars if they worked here) the assignment is still needed.

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
N-Dekker: Looking at an older version, it looks like it assumes that the output region has the same size as the input region. Because it just checks inputIt.IsAtEnd() and inputIt.IsAtEndOfLine(), not outputIt.IsAtEnd() or outputIt.IsAtEndOfLine(): https://github.com/InsightSoftwareConsortium/ITK/blob/921c5bab8f0803224d6b7cd7c5ab14aefcfe89ff/Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx#L145-L165

That's OK to me, as long as that assumption is OK  😃 So then, in the new code, it's also OK to just check `inputRange.end()` and ignore `outputRange.end()`.

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
N-Dekker: I didn't try, but such a zip might be possible. Very cute 😸 However... `zip(inputRange, outputRange)` will check both inputRange.end() and outputRange.end(). It may not be necessary to check both, when we assume that input and output region have the same size anyway. So then, `zip(inputRange, outputRange)` may be unnecessary slow!

---

Update: The input and output region do indeed have the same number of pixels, as implied by the reply from @dzenanz at https://github.com/InsightSoftwareConsortium/ITK/pull/5670/files/2e8c70a4499a2e87c8e0f65d25a355baa82f3bb9#r2586937371 So then, it is indeed unnecessary to check _both_ inputRange.end() _and_ outputRange.end(). So in this case `zip` is not necessary. Just a range-based `for` loop over `inputRange` should be fine  😃

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
dzenanz: Cast image filter does not need a larger input region, so it does not override the default `CallCopyOutputRegionToInputRegion` implementation, and has the same input and output regions.

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
blowekamp: Using the range-based for loop in did not give the same performance is all cases in the test code.  🤷

N-Dekker: > Using the range-based for loop in did not give the same performance is all cases in the test code.

Interesting! Do you mean that my range-based `for` suggestion is in some cases _significantly_ slower than your current PR? If so, do you have a clue why?

N-Dekker: By the way, the range-based `for` loop can also be written as:

    for (const InputPixelType & inputPixel : ImageRegionRange(*inputPtr, inputRegionForThread))

So then the `inputRange` variable may be removed as well!

blowekamp: It appears measurable slower in the performance testing code. I don't know why.

PR #5672, opened 2025-12-05
DOC: BSplineTransform does not have ImageConstPointer type alias

dzenanz:
## PR Checklist

### DIFF
--- a/Modules/Core/Transform/include/itkBSplineTransform.h
+++ b/Modules/Core/Transform/include/itkBSplineTransform.h
@@ -82,7 +82,7 @@ namespace itk
  * are the same buffered region. The following illustrates the API:
  * \code
 
-   TransformType::ImageConstPointer images[2];
+   TransformType::ImagePointer images[2];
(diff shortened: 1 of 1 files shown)


Modules/Core/Transform/include/itkBSplineTransform.h
```diff
@@ -82,7 +82,7 @@ namespace itk
  * are the same buffered region. The following illustrates the API:
  * \code
 
-   TransformType::ImageConstPointer images[2];
+   TransformType::ImagePointer images[2];
 
    // Fill the images up with values
```
N-Dekker, line 88: Good catch @dzenanz.

Looking at the rest of the example, calling `transform->SetCoefficientImages(images)`, I wonder if `BSplineTransform::SetCoefficientImages` shouldn't actually support "const" images! `BSplineTransform` doesn't modify those images, right?

Anyway, I think that's beyond the scope of this PR.  😇

PR #5679, opened 2025-12-10
ENH: Use ImageRegionRange iterator in ImageAlgorithm::Copy

blowekamp:
## PR Checklist

### DIFF
--- a/Modules/Filtering/ImageGrid/include/itkPasteImageFilter.hxx
+++ b/Modules/Filtering/ImageGrid/include/itkPasteImageFilter.hxx
@@ -21,6 +21,7 @@
 #include "itkObjectFactory.h"
 #include "itkTotalProgressReporter.h"
 #include "itkImageAlgorithm.h"
+#include "itkImageScanlineIterator.h"
 
 namespace itk
 {


dzenanz (approved):

N-Dekker:
Cool, thanks @blowekamp 👍

Modules/Core/Common/include/itkImageAlgorithm.hxx
```diff
@@ -35,33 +34,17 @@ ImageAlgorithm::DispatchedCopy(const InputImageType *                       inIm
                                const typename OutputImageType::RegionType & outRegion,
                                FalseType)
 {
-  if (inRegion.GetSize()[0] == outRegion.GetSize()[0])
```
N-Dekker, line 38: The original code distinguishes between whether or not `inRegion.GetSize()[0] == outRegion.GetSize()[0]`. So it does _not_ assume that inRegion and outRegion always have the same multi-dimensional size, right?

Still, the original code _does_ assume that `outRegion` has at least as many pixels as `inRegion`, right?

I hope we can simply assume that `inRegion.GetSize() == outRegion.GetSize()`!

blowekamp: Good insight. If a check for this was needed I'd place it before the dispatch, as it should be done for all implementation. I don't see it in the header file either.

The documentation for the Copy function does show iterating until the end of the input region.

dzenanz: There was an `if` which used ScanlineIterator, and an implicit else, which used the regular iterator.

N-Dekker: Looks like all the tests passed successfully, so then I think it's just fine  😃

thewtex (approved):

PR #5680, opened 2025-12-10
BUG: Avoid copying uninitialized pixels in CastImageFilter

N-Dekker:
For a regular output pixel type like `RGBPixel<T>` and `itk::Vector<T>`, the expression `OutputPixelType outputPixel{ *outputIt }` had undefined behavior, within `CastImageFilter::DynamicThreadedGenerateDataDispatched`, because at that point, the pixel pointed to by `outputIt` was typically still uninitialized. For such output pixel types, `outputPixel` should be declared as a reference (`OutputPixelType &`) instead.

The original `OutputPixelType outputPixel{ *outputIt }` initialization is only OK when the output pixel type is a `VariableLengthVector<T>`.

### DIFF
--- a/Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
+++ b/Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
@@ -21,6 +21,7 @@
 #include "itkProgressReporter.h"
 #include "itkImageAlgorithm.h"
 #include "itkImageRegionRange.h"
+#include "itkVariableLengthVector.h"
 
 namespace itk
 {
@@ -156,12 +157,19 @@ CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatche
   while (inputIt != inputEnd)
   {
     const InputPixelType & inputPixel = *inputIt;
-    OutputPixelType        outputPixel{ *outputIt };
+
+    using OutputPixelValueType = typename OutputPixelType::ValueType;
+
+    constexpr bool isVariableLengthVector = std::is_same_v<OutputPixelType, VariableLengthVector<OutputPixelValueType>>;
+
+    // If the output pixel type is a VariableLengthVector, it behaves as a "reference" to the internal data. Otherwise
+    // declare outputPixel as a reference, `OutputPixelType &`, to allow it to access the internal buffer directly.
+    std::conditional_t<isVariableLengthVector, OutputPixelType, OutputPixelType &> outputPixel{ *outputIt };
+
     for (unsigned int k = 0; k < componentsPerPixel; ++k)
     {
-      outputPixel[k] = static_cast<typename OutputPixelType::ValueType>(inputPixel[k]);
+      outputPixel[k] = static_cast<OutputPixelValueType>(inputPixel[k]);
     }
-    *outputIt = outputPixel;
     ++inputIt;
     ++outputIt;
   }


N-Dekker:
> Please add this "optimization" to the benchmark test to determine if it's worth the code.

I can add this to the benchmark, but it's not an "optimization". It's really a bug fix!  When `v1` is an uninitialized `itk::Vector`, `itk::Vector v2{ v1 }` really is a bug. It has undefined behavior.

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
```diff
@@ -156,12 +156,19 @@ CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatche
...
     const InputPixelType & inputPixel = *inputIt;
-    OutputPixelType        outputPixel{ *outputIt };
+
+    using OutputPixelValueType = typename OutputPixelType::ValueType;
+
+    constexpr bool isVariableLengthVector = std::is_same_v<OutputPixelType, VariableLengthVector<OutputPixelValueType>>;
+
+    // If the output pixel type is a VariableLengthVector, it behaves as a "reference" to the internal data. Otherwise
+    // declare outputPixel as a reference, `OutputPixelType &`, to allow it to access the internal buffer directly.
+    std::conditional_t<isVariableLengthVector, OutputPixelType, OutputPixelType &> outputPixel{ *outputIt };
```
N-Dekker, line 167: Please check if this is the best criterion for choosing between:

1. `OutputPixelType outputPixel{ *outputIt };`  // Without C++ reference.
2. `OutputPixelType & outputPixel{ *outputIt };` // _With_ C++ reference.

Another possibility might be: check if `TOutputImage` is a `VectorImage`.

N-Dekker: I assume that CastImageFilter supports  `itk::VectorImage<T>` and `itk::Image<itk::Vector<T>>`, but _not_ `itk::Image<itk::VariableLengthVector<T>>`, right? A regular `itk::Image` whose (internal) pixel type is `itk::VariableLengthVector` seems strange to me, and it certainly isn't tested, for CastImageFilter.

N-Dekker: Not to mention `itk::VectorImage<itk::VariableLengthVector<T>>`... that would be _very_ strange to me  😸

blowekamp: Given that point,  it's certainly safer to check for VectorImage than the pixel type. 

The case for the Vector pixel type having "undefined" behavior for this situation, I have been struggling with. I suppose if the output pixel type was something like a std::vector ( or a variable length vector), there would be a major problem with the pixel being a non-trivial object.

However, for these trivial object I am not sure/convinced that assigning ( and not observing) the indeterminate value is really undefined behavior.
https://en.cppreference.com/w/cpp/language/default_initialization.html#Read_from_an_indeterminate_byte

N-Dekker: Thanks for the link, @blowekamp Looks like `itk::Vector<float>` is not an "uninitialized-friendly type", so I guess the following is applicable here:

"If an evaluation produces an indeterminate value, the behavior is [undefined](https://en.cppreference.com/w/cpp/language/ub.html)." 🤷

N-Dekker: > Given that point, it's certainly safer to check for VectorImage than the pixel type.

So would you prefer checking OutputImageType == VectorImage, instead of the current PR, checking OutputPixelType == VariableLengthVector ?

blowekamp: If we are going to have code to specially handle the Vector Images on output, we could just be clear and direct with it:
```
if constexpr ( IsOutputVectorImage ) {...} else {...}
```
IMHO it clearer than doing meta programming for the type.

I am wondering if we could explicit enable the `ImageIterator::Value` method to work like the reference we are trying to get the following to work:
```
 (*outputIt)[k] = static_cast<OutputPixelValueType>(inputPixel[k]);
```

N-Dekker: Sorry, I prefer the currently proposed:

```cpp
std::conditional_t<isVariableLengthVector, OutputPixelType, OutputPixelType &> outputPixel{ *outputIt };
```

Rather than:

```cpp
if constexpr (IsOutputVectorImage)
{
  OutputPixelType outputPixel{ *outputIt };
}
else
{
  OutputPixelType & outputPixel{ *outputIt };
}
```

I don't like too much code duplication. So I just stick with the current PR 🤷

blowekamp: Here at line 103 you proposed two differ loops:
https://github.com/InsightSoftwareConsortium/ITKPerformanceBenchmarking/pull/109/files

A similar things can be done with the Ranged iterator. The needs for the two loops and setting them are different, by as well just make two. We are conviently using the range for loop so it's rather small the duplication. Each loop can do it's one thing well.

N-Dekker: > I am wondering if we could explicit enable the ImageIterator::Value method to work like the reference we are trying to get the following to work:

```cpp
 (*outputIt)[k] = static_cast<OutputPixelValueType>(inputPixel[k]);
```
For VectorImage, `*outputIt` returns an instance of `PixelProxy`. For the end-user, `PixelProxy` only has an assignment operator to assign a pixel value to the proxy, and a conversion operator to retrieve the pixel value. Technically, we could manually still add other member functions from the pixel type to the proxy type, but I'm afraid it would cause feature creep. Moreover, it would be inefficient to call such member functions via the proxy class, instead of directly on the pixel type.  Sorry 🤷

N-Dekker: > Here at line 103 you proposed two differ loops:
> https://github.com/InsightSoftwareConsortium/ITKPerformanceBenchmarking/pull/109/files

You know I felt very bad 😞  when I did so! 

You see, the "if" in https://github.com/InsightSoftwareConsortium/ITKPerformanceBenchmarking/pull/109 has a `outputIt.Get()` call at the beginning, whereas the "else" has `outputIt.Set(value)` at the end: 

```cpp
      if constexpr (isVariableLengthVector<OutputPixelType>)
      {
        OutputPixelType value(outputIt.Get());
        for (unsigned int k = 0; k < componentsPerPixel; ++k)
        {
          value[k] = static_cast<typename OutputPixelType::ValueType>(inputPixel[k]);
        }
      }
      else
      {
        OutputPixelType value;
        for (unsigned int k = 0; k < componentsPerPixel; ++k)
        {
          value[k] = static_cast<typename OutputPixelType::ValueType>(inputPixel[k]);
        }
        outputIt.Set(value);
      }
```  
I did not know how to replace this `if constexpr` with a simple `std::conditional_t`  🤷

thewtex (approved):

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
```diff
@@ -156,12 +156,19 @@ CastImageFilter<TInputImage, TOutputImage>::DynamicThreadedGenerateDataDispatche
   while (inputIt != inputEnd)
   {
     const InputPixelType & inputPixel = *inputIt;
-    OutputPixelType        outputPixel{ *outputIt };
+
+    using OutputPixelValueType = typename OutputPixelType::ValueType;
+
+    constexpr bool isVariableLengthVector = std::is_same_v<OutputPixelType, VariableLengthVector<OutputPixelValueType>>;
```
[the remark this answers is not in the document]
N-Dekker: OK, thanks, that makes sense!

Modules/Filtering/ImageFilterBase/include/itkCastImageFilter.hxx
blowekamp: This does not contain this discussed ( and I thought agreed upon ) change to detect VectorImage instead of VariableLengthVector. This theoretical case of Image of VariableLengthVetor would not work. The optimization where the VariableLengthVector reference the vector image buffers is done here:
https://github.com/InsightSoftwareConsortium/ITK/blob/27244a03ced315691085ecc9ed3b80d4d948b981/Modules/Core/Common/include/itkDefaultVectorPixelAccessor.h#L78

This would not apply with other Image types, and the check should be based on the image type not the pixel type.

PR #5682, opened 2025-12-11
ENH: Add `std::vector`-like interface to VariableLengthVector

N-Dekker:
Added the following member functions:

     empty()
     size()
     at(size_type)
     front()
     back()
     data()
     begin()
     end()
     cbegin()
     cend()
     rbegin()
     rend()
     crbegin()
     crend()

As well as STL-style nested type aliases.

Aims to ease using `VariableLengthVector` in generic code.

### DIFF
--- a/Modules/Core/Common/include/itkVariableLengthVector.h
+++ b/Modules/Core/Common/include/itkVariableLengthVector.h
@@ -20,6 +20,7 @@
 
 #include <cassert>
 #include <algorithm>
+#include <iterator>
 #include <type_traits>
 #include "itkNumericTraits.h"
 #include "itkMetaProgrammingLibrary.h"
@@ -313,6 +314,17 @@ public:
 
   /** Typedef used to indicate the number of elements in the vector */
   using ElementIdentifier = unsigned int;
+
+  // Nested types from C++ Standard section "Container requirements" [container.reqmts], and from `std::vector`:
+  using value_type = TValue;
+  using reference = TValue &;
+  using const_reference = const TValue &;
+  using iterator = TValue *;
+  using const_iterator = const TValue *;
+  using difference_type = ptrdiff_t;
+  using size_type = size_t;
+  using reverse_iterator = std::reverse_iterator<iterator>;
+  using const_reverse_iterator = std::reverse_iterator<const_iterator>;
 
   /** Default constructor. It is created with an empty array
    *  it has to be allocated later by assignment, \c SetSize() or \c Reserve().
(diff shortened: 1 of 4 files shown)


Modules/Core/Common/test/itkRangeGTestUtilities.h
```diff
@@ -156,6 +165,55 @@ class RangeGTestUtilities
...
+    const auto distance = std::distance(std::begin(container), std::end(container));
+    ASSERT_GE(distance, 0);
+    EXPECT_EQ(static_cast<size_t>(distance), std::size(container));
+  }
+
+
+  // Expects that `distance(&front, &back) + 1` is equal to `size`, for the specified (non-empty) container.
+  template <typename TContainer>
+  static void
+  ExpectDistanceFromFrontToBackPlusOnePlusEqualsSize(const TContainer & container)
```
dzenanz: This name has too many "plus" words in it.

N-Dekker: Thanks @dzenanz. Addressed by [this force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/7794a31f6b8a10a2f9f09a362c1b7203a7dd6f8c..d660d09951a15a0f690f970692f7e7c30e6c1600). 😃

Modules/Core/Common/test/itkRangeGTestUtilities.h
```diff
@@ -156,6 +165,55 @@ class RangeGTestUtilities
...
+  }
+
+
+  // Expects that `distance(&front, &back) + 1` is equal to `size`, for the specified (non-empty) container.
+  template <typename TContainer>
+  static void
+  ExpectDistanceFromFrontToBackPlusOneEqualsSize(const TContainer & container)
+  {
+    ASSERT_FALSE(std::empty(container));
+    const auto distance = std::distance(&container.front(), &container.back());
```
N-Dekker, line 211: Note to self: it's a bit overdone, calling `std::distance` to estimate the distance between two raw pointers. We might as well just do:

    const auto distance = &container.back() - &container.front();

However, some coding guidelines don't like pointer arithmetic, so it might still be recommended to call `std::distance`  😸

Modules/Core/Common/include/itkVariableLengthVector.h
```diff
@@ -314,6 +315,17 @@ class ITK_TEMPLATE_EXPORT VariableLengthVector
   /** Typedef used to indicate the number of elements in the vector */
   using ElementIdentifier = unsigned int;
 
+  // Nested types from C++ Standard section "Container requirements" [container.reqmts], and from `std::vector`:
+  using value_type = TValue;
+  using reference = TValue &;
+  using const_reference = const TValue &;
+  using iterator = TValue *;
+  using const_iterator = const TValue *;
+  using difference_type = itk::ptrdiff_t;
```
N-Dekker: Note to self: `itk::` is not really necessary here, as we are already in the `itk` namespace. That's no show-stopper, though 😃

N-Dekker: OK, removed the unnecessary `itk::` from `ptrdiff_t` and rebased main: [force-pushed](https://github.com/InsightSoftwareConsortium/ITK/compare/d660d09951a15a0f690f970692f7e7c30e6c1600..67f2c16a95f5c0b22e871c79da048539bd610c89)

PR #5685, opened 2025-12-12
STYLE: Add `[[nodiscard]]` to non-const data(), begin(), end(), etc.

N-Dekker:
The clang-tidy [`modernize-use-nodiscard`](https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-nodiscard.html) check only considers _const_ member functions, whereas this commit specifically addresses non-const member functions.

- Follow-up to pull request #5404 by @hjmjohnson

### DIFF
--- a/Modules/Core/Common/include/itkFixedArray.h
+++ b/Modules/Core/Common/include/itkFixedArray.h
@@ -324,7 +324,7 @@ public:
     return m_InternalArray;
   }
 
-  ValueType *
+  [[nodiscard]] ValueType *
   data()
   {
     return m_InternalArray;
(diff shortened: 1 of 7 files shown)


hjmjohnson (approved):
Very nice.  Thanks for following up.

Modules/Core/Common/include/itkFixedArray.h
```diff
@@ -324,7 +324,7 @@ class ITK_TEMPLATE_EXPORT FixedArray
     return m_InternalArray;
   }
 
-  ValueType *
+  [[nodiscard]] ValueType *
```
dzenanz, line 327: Does PointType inherit this from FixedArray? Or does it already have `[[nodiscard]]` decorators?

dzenanz (approved):

Modules/Core/Common/include/itkFixedArray.h
N-Dekker: `itk::Point` inherits the `data()` member functions from `FixedArray` 👍

PR #5688, opened 2025-12-14
Use initializer lists, constexpr-if and switch in SobelOperator::GenerateCoefficients()

N-Dekker:
Replaced groups of consecutive `CoefficientVector::push_back` calls with `return` statements that simply initialize the CoefficientVector by an initializer list of those coefficients.

Used constexpr-`if` and `switch` blocks to avoid calling `this->GetDirection()` repeatedly during one and the same
execution of the function.

### DIFF
--- a/Modules/Core/Common/include/itkSobelOperator.hxx
+++ b/Modules/Core/Common/include/itkSobelOperator.hxx
@@ -28,42 +28,39 @@ SobelOperator<TPixel, VDimension, TAllocator>::Fill(const CoefficientVector & co
 {
   this->InitializeToZero();
 
-  if constexpr (VDimension == 2 || VDimension == 3)
+  // Note that this code is only good for 2d and 3d operators.  Places the
+  // coefficients in the exact center of the neighborhood
+  const unsigned int center = this->GetCenterNeighborhoodIndex();
+
+  if constexpr (VDimension == 2)
   {
-    // Note that this code is only good for 2d and 3d operators.  Places the
-    // coefficients in the exact center of the neighborhood
-    const unsigned int center = this->GetCenterNeighborhoodIndex();
-
-    if constexpr (VDimension == 3)
+    unsigned int coeff_index = 0;
+    for (int y = -1; y <= 1; ++y)
     {
-      unsigned int coeff_index = 0;
-      for (int z = -1; z <= 1; ++z)
+      for (int x = -1; x <= 1; ++x)
       {
-        for (int y = -1; y <= 1; ++y)
-        {
-          for (int x = -1; x <= 1; ++x)
-          {
-            const int pos = center + z * this->GetStride(2) + y * this->GetStride(1) + x * this->GetStride(0);
+        const int pos = center + y * this->GetStride(1) + x * this->GetStride(0);
+        // Note, The following line copies the double precision
+        // coefficients of SobelOperator to the pixel type
+        // of the neighborhood operator which may not support
+        // negative numbers, or floating point numbers.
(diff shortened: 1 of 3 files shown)


Modules/Core/Common/include/itkSobelOperator.hxx
```diff
@@ -82,136 +82,30 @@ template <typename TPixel, unsigned int VDimension, typename TAllocator>
...
+      case 1:
+        return { -1, -3, -1, 0, 0, 0, 1, 3, 1, -3, -6, -3, 0, 0, 0, 3, 6, 3, -1, -3, -1, 0, 0, 0, 1, 3, 1 };
+      case 2:
+        return { -1, -3, -1, -3, -6, -3, -1, -3, -1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 3, 1, 3, 6, 3, 1, 3, 1 };
+    }
   }
-
-  return coeff;
+  itkExceptionStringMacro(
+    "The ND version of the Sobel operator has not been implemented. Currently only 2D and 3D versions are available.");
```
blowekamp: Should this a compile time error or a static assert? Perhaps it is already impossible for this string to exist in compiler output?

dzenanz: A static assert on `VDimension` sounds appealing.

N-Dekker: `SobelOperator::Fill` also only supports 2D and 3D: https://github.com/InsightSoftwareConsortium/ITK/blob/07f5450a715c3f17e4fe0f140724f59cd02e5432/Modules/Core/Common/include/itkSobelOperator.hxx#L27-L31

So I guess the static_assert should then be at class level, right? `SobelOperator` should not be instantiated at all, for >= 4D or <= 1D, right?

N-Dekker: @blowekamp @dzenanz Added commit: [ENH: static_assert SobelOperator can only be instantiated for 2D or 3D](https://github.com/InsightSoftwareConsortium/ITK/pull/5688/commits/9cb95d0372fc347ce374ccf91c7457c4bef8a1f3). Please check!

N-Dekker: With this [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/9cb95d0372fc347ce374ccf91c7457c4bef8a1f3..35d6a3d28198e3927484c9d1fb2f03d40cecfc0a), I removed the test for 4D. It triggered a compile error at https://open.cdash.org/viewBuildError.php?buildid=10897987 saying:

```
itkSobelOperator.h(105): error C2338: static_assert failed: 'The ND version of the Sobel operator has not been implemented. Currently only 2D and 3D versions are available.'
```

hjmjohnson:
@N-Dekker, @dzenanz  Building with wrapping is erroring for me when building python packages:  I think we may need to restrict the dimensions manually.  

"In file included from /Users/johnsonhj/Dashboard/src/ITKPP-bld/build/ITK-3.9-macos_arm64/Wrapping/castxml_inputs/itkSobelOperator.cxx:16:
/Users/johnsonhj/Dashboard/src/ITKPP-bld/build/ITK-source/ITK/Modules/Core/Common/include/itkSobelOperator.h:105:5: error: static assertion failed due to requirement '4U == 2 || 4U == 3': The ND version of the Sobel operator has not been implemented. Currently only 2D and 3D versions are available.
  105 |     VDimension == 2 || VDimension == 3,
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 "

PR #5692, opened 2025-12-15
Deprecate `MultiThreaderType` aliases, use `MultiThreaderBase` directly

N-Dekker:
`ProcessObject::MultiThreaderType` and `ImageToImageMetric::MultiThreaderType` are always just aliases of `itk::MultiThreaderBase`, the common base class of PlatformMultiThreader, PoolMultiThreader, and TBBMultiThreader. So it appears clearer to just use `MultiThreaderBase` directly.

Deprecated those aliases, `ProcessObject::MultiThreaderType` and `ImageToImageMetric::MultiThreaderType`.

### DIFF
--- a/Modules/Registration/Common/include/itkImageToImageMetric.h
+++ b/Modules/Registration/Common/include/itkImageToImageMetric.h
@@ -349,7 +349,11 @@ public:
   itkGetConstReferenceMacro(UseCachingOfBSplineWeights, bool);
   itkBooleanMacro(UseCachingOfBSplineWeights);
   /** @ITKEndGrouping */
-  using MultiThreaderType = MultiThreaderBase;
+
+#ifndef ITK_FUTURE_LEGACY_REMOVE
+  using MultiThreaderType [[deprecated("Please use `itk::MultiThreaderBase` directly.")]] = MultiThreaderBase;
+#endif
+
   /** Get the Threader. */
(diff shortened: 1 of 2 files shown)


Modules/Registration/Common/include/itkImageToImageMetric.h
```diff
@@ -600,15 +604,15 @@ class ITK_TEMPLATE_EXPORT ImageToImageMetric : public SingleValuedCostFunction
...
+   * MultiThreaderBase::WorkUnitInfo info type
    * Do not allow inheritance for objects that are intended for static_cast<void *>
    * \ingroup ITKRegistrationCommon
    */
   class MultiThreaderWorkUnitInfoImageToImageMetricWrapper final
   {
   public:
     MultiThreaderWorkUnitInfoImageToImageMetricWrapper(const void * workunitInfoAsVoid)
-      : m_WorkUnitInfo(static_cast<const typename MultiThreaderType::WorkUnitInfo *>(workunitInfoAsVoid))
+      : m_WorkUnitInfo(static_cast<const MultiThreaderBase::WorkUnitInfo *>(workunitInfoAsVoid))
```
N-Dekker, line 615: Note that the `typename` keyword is not necessary for `MultiThreaderBase::WorkUnitInfo`   😃

PR #5693, opened 2025-12-16
Add `VariableLengthVector(length, value)` constructor

N-Dekker:
Proposed a `VariableLengthVector(unsigned int length, const TValue & value)` constructor that creates a `VariableLengthVector` of the specified `length`, filled with the specified `value` for each element.

Added a use case: simplifying `NumericTraits<VariableLengthVector<T>>`.

`Array` and `OptimizerParameters` already have such a constructor.

- Follow-up to pull request #2248

### DIFF
--- a/Modules/Core/Common/include/itkNumericTraitsVariableLengthVectorPixel.h
+++ b/Modules/Core/Common/include/itkNumericTraitsVariableLengthVectorPixel.h
@@ -95,45 +95,31 @@ public:
   static const Self
   max(const Self & a)
   {
-    Self b(a.Size());
-
-    b.Fill(NumericTraits<T>::max());
-    return b;
+    return Self(a.Size(), NumericTraits<T>::max());
   }
 
   static const Self
   min(const Self & a)
   {
-    Self b(a.Size());
-
-    b.Fill(NumericTraits<T>::min());
-    return b;
+    return Self(a.Size(), NumericTraits<T>::min());
   }
 
   static const Self
   ZeroValue(const Self & a)
   {
-    Self b(a.Size());
-
-    b.Fill(T{});
-    return b;
+    return Self(a.Size(), T{});
   }
 
   static const Self
   OneValue(const Self & a)
   {
-    Self b(a.Size());
-
-    b.Fill(NumericTraits<T>::OneValue());
-    return b;
+    return Self(a.Size(), NumericTraits<T>::OneValue());
   }
 
   static const Self
   NonpositiveMin(const Self & a)
   {
-    Self b(a.Size());
-    b.Fill(NumericTraits<T>::NonpositiveMin());
-    return b;
+    return Self(a.Size(), NumericTraits<T>::NonpositiveMin());
   }
 
   static bool


dzenanz (approved):
Looks good on a glance.

Modules/Core/Common/include/itkVariableLengthVector.h
```diff
@@ -344,6 +345,17 @@ class ITK_TEMPLATE_EXPORT VariableLengthVector
...
+  /** Constructor with size and initial value for each element. */
+  explicit VariableLengthVector(unsigned int length, const TValue & value)
+    : m_LetArrayManageMemory{ true }
+    , m_Data{ [length, &value] {
+      auto data = make_unique_for_overwrite<TValue[]>(length);
+      std::fill_n(data.get(), length, value);
+      return data.release();
+    }() }
+    , m_NumElements{ length }
+  {}
```
blowekamp: This implementation looks reasonable.... but, I looks at the other implementation in the hxx file. Why is this in the header and not hxx file? And is this implementation is rather different that `VariableLengthVector(unsigned int)` which uses `Reserve`.  Might be good to keep these is sync and not diverge too much?

Might even make since to do:
```
explicit VariableLengthVector(unsigned int length, const TValue & value) : VariableLengthVector(length) {
std::fill_n(m_Data, length, value);
}
```

blowekamp (approved):
Looks fine.

Modules/Core/Common/include/itkVariableLengthVector.h
N-Dekker: Thanks for your comments @blowekamp Is it preferable to move the implementation of this little constructor to the hxx file? VariableLengthVector does have _some_ of its member function implementations in its h file, already:

https://github.com/InsightSoftwareConsortium/ITK/blob/10965478ebc991d93650d7ef3d4ca10bcfcab6b3/Modules/Core/Common/include/itkVariableLengthVector.h#L402-L406

N-Dekker: Your suggestion is nice, but it may not be exception-safe: 
```
explicit VariableLengthVector(unsigned int length, const TValue & value) : VariableLengthVector(length) {
std::fill_n(m_Data, length, value);
}
```

When `std::fill_n(m_Data, length, value)` throws an exception, is the data still deallocated? 

----

Update: I just checked: your suggestion does _not_ leak, when the body (the `std::fill_n` call) throws an exception. 👍

blowekamp: There is some extra complication in the `VariableLengthVector(unsigned int)` construction... but there seems to be a lot of room for improvement in this class. 

Thank you for working on it.

N-Dekker: @blowekamp Addressing your comment, please wait...

N-Dekker: > There is some extra complication in the `VariableLengthVector(unsigned int)` construction... but there seems to be a lot of room for improvement in this class.

Indeed! I think `VariableLengthVector(unsigned int)` should call `AllocateElements`, instead of `Reserve`, right? Or simply just do `new TValue[length]`!

blowekamp: The utility of `AllocateElements` wrapping a `new` to throw an exception which requires additional memory allocation to perform `ostringstream` operations is certainly IMHO, questionable.

thewtex (approved):

Modules/Core/Common/include/itkVariableLengthVector.h
blowekamp: I think the changes to the existing constructor is another PR.

PR #5705, opened 2025-12-20
Use constructors with `(length, value)` and `AllocateInitialized()`, remove `Fill(value)` calls in tests

N-Dekker:
Replaced code like:

    ArrayType var(length);
    var.Fill(value);

with:

    ArrayType var(length, value);

    Find what: ^( [ ]+)([^ \n].* )(\w+)\(([^,\n]+)\);[\r\n]+\1\3\.Fill\(([^,\n]+)\);$
    Replace with: $1$2$3\($4, $5\);
    Filters: itk*Test*.cxx
    [v] Match case
    (*) Regular expression

In general, the `ArrayType(length)` constructor may not initialize the internal data (which may be unsafe), whereas `ArrayType(length, value)` properly "fills" the internal data.

### DIFF
--- a/Modules/Core/Common/test/itkImageBufferRangeGTest.cxx
+++ b/Modules/Core/Common/test/itkImageBufferRangeGTest.cxx
@@ -536,8 +536,7 @@ TEST(ImageBufferRange, SupportsVectorImage)
   image->SetRegions(imageSize);
   image->SetVectorLength(vectorLength);
   image->AllocateInitialized();
-  PixelType fillPixelValue(vectorLength);
-  fillPixelValue.Fill(42);
+  PixelType fillPixelValue(vectorLength, 42);
   image->FillBuffer(fillPixelValue);
 
   using RangeType = ImageBufferRange<ImageType>;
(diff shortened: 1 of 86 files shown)


Modules/Filtering/Smoothing/test/itkSmoothingRecursiveGaussianImageFilterOnImageAdaptorTest.cxx
```diff
@@ -58,23 +58,11 @@ itkSmoothingRecursiveGaussianImageFilterOnImageAdaptorTest(int, char *[])
...
   using myIteratorType = itk::ImageRegionIteratorWithIndex<myImageType>;
 
-  // Create one iterator for the Input Image A (this is a light object)
-  myIteratorType it(inputImage, inputImage->GetRequestedRegion());
-
-  // Initialize the content of Image A
-  while (!it.IsAtEnd())
-  {
-    myImageType::PixelType p(numberOfComponents);
-    p.Fill(0.0);
```
N-Dekker, line 73: These two lines matched the regular expression (`^( [ ]+)([^ \n].* )(\w+)\(([^,\n]+)\);[\r\n]+\1\3\.Fill\(([^,\n]+)\);$`), but they could not be replaced with `myImageType::PixelType p(numberOfComponents, 0.0);`! Doing so caused a compile error at https://open.cdash.org/viewBuildError.php?buildid=10910594 saying:

    error: no matching constructor for initialization of 'myImageType::PixelType' (aka 'itk::RGBPixel<float>')

Fortunately, the entire `while` loop appears unnecessary, when using `AllocateInitialized()`  😃

PR #5710, opened 2025-12-23
Remove AllocateElements from VariableLengthVector

N-Dekker:
`VariableLengthVector::AllocateElements` just converted any possible exception from `new TValue[size]` to an ITK exception:
https://github.com/InsightSoftwareConsortium/ITK/blob/d66197a727f5b9c1d8e4859dfc52710f5f982603/Modules/Core/Common/include/itkVariableLengthVector.hxx#L204-L216

It appears unnecessary for VariableLengthVector to do so. In many other cases, ITK _does_ allow non-ITK exceptions to occur, so end-users have to deal with them anyway.

This pull request was triggered by a comment by Bradley (@blowekamp) at https://github.com/InsightSoftwareConsortium/ITK/pull/5693#discussion_r2623942602 saying:
> The utility of `AllocateElements` wrapping a `new` to throw an exception which requires additional memory allocation to perform `ostringstream` operations is certainly IMHO, questionable.

PR #5713, opened 2026-01-02
BUG: Make `VariableLengthVector::m_Data` null when its length is zero

N-Dekker:
Originally, a zero-length VariableLengthVector might still have its `m_Data` pointing to the heap, as it might have its memory allocated by `new TValue[length]`, with `length == 0`.

A memory leak in `VariableLengthVector(unsigned int length)` was then introduced by pull request https://github.com/InsightSoftwareConsortium/ITK/pull/5710 commit bb79801dcf43e99ca11b7d47cbbc88c1a549889a
"STYLE: Replace AllocateElements calls within VariableLengthVector" forgetting to `delete` memory that was allocated by `new TValue[0]`. As was reported by Mihail Isakov (@issakomi) at https://github.com/InsightSoftwareConsortium/ITK/pull/5710#issuecomment-3690435714, referring to the Valgrind results at https://open.cdash.org/builds/10917349/dynamic_analysis

With this commit, m_Data should always be null when the VariableLengthVector is empty and manages its own memory (m_LetArrayManageMemory is `true`).

### DIFF
--- a/Modules/Core/Common/include/itkVariableLengthVector.hxx
+++ b/Modules/Core/Common/include/itkVariableLengthVector.hxx
@@ -29,18 +29,15 @@ namespace itk
 
 template <typename TValue>
 VariableLengthVector<TValue>::VariableLengthVector(unsigned int length)
-  : m_Data(new TValue[length])
+  : m_Data(length == 0 ? nullptr : new TValue[length])
   , m_NumElements(length)
-{
-  // postcondition(s)
-  itkAssertInDebugAndIgnoreInReleaseMacro(m_Data != nullptr);
-}
+{}
 
 template <typename TValue>
 VariableLengthVector<TValue>::VariableLengthVector(unsigned int length, const TValue & value)
   : VariableLengthVector(length)
 {
-  std::fill_n(m_Data, length, value);
+  Fill(value);
 }
 
 template <typename TValue>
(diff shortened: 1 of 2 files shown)


Modules/Core/Common/include/itkVariableLengthVector.hxx
```diff
@@ -293,10 +292,11 @@ template <typename TValue>
 void
 VariableLengthVector<TValue>::Fill(const TValue & v)
 {
-  itkAssertInDebugAndIgnoreInReleaseMacro(m_NumElements == 0 || (m_NumElements > 0 && m_Data != nullptr));
-  // VC++ version of std::fill_n() expects the output iterator to be valid
-  // instead of expecting the range [OutIt, OutIt+n) to be valid.
```
N-Dekker, line 298: Removed this comment, as it does not apply to currently supported versions of VC++ anymore, as far as I can see.

David Cole (@dlrdave) mentioned in https://review.source.kitware.com/%23/c/20307/, Oct 24, 2015:
> You cannot call std::fill_n with a nullptr for Debug VS 2013.
>
> The implementation of fill_n looks like this:
>
    template<class _OutIt,
      class _Diff,
      class _Ty> inline
      _OutIt fill_n(_OutIt _Dest, _Diff _Count, const _Ty& _Val)
      {  // copy _Val _Count times through [_Dest, ...)
      _DEBUG_POINTER(_Dest);
      return (_Fill_n(_Dest, _Count, _Val,
        _Is_checked(_Dest)));
      }
>
> The _DEBUG_POINTER line is called unconditionally, and will throw this assert if the pointer is nullptr.

However, this appears obsolete now, as VS 2019 and VS2022 no longer have those `_DEBUG_POINTER` macro calls. I don't see them in https://github.com/microsoft/STL/ either.

Honestly I don't know if the C++ standard now officially allows passing a null pointer as iterator to `fill_n`, or any other STL algorithm. 🤷

dzenanz (approved):
LGTM

PR #5714, opened 2026-01-05
BUG: Fix duplicate output files in Convolution streaming tests

blowekamp:
Made output filenames unique for FFT convolution streaming tests to prevent file collisions when tests run in parallel. Each test now uses a distinct output filename with 'Frequency' suffix for FFT variants.

## PR Checklist

### DIFF
--- a/Modules/Filtering/Convolution/test/CMakeLists.txt
+++ b/Modules/Filtering/Convolution/test/CMakeLists.txt
@@ -641,11 +641,11 @@ itk_add_test(
     ITKConvolutionTestDriver
     --compare
     DATA{Baseline/itkConvolutionImageFilterStreamingTestOutput.mha}
-    ${ITK_TEST_OUTPUT_DIR}/itkConvolutionImageFilterStreamingTestOutput.mha
+    ${ITK_TEST_OUTPUT_DIR}/itkConvolutionImageFilterStreamingValidTestOutput.mha
     itkConvolutionImageFilterStreamingTest
     spatial
     DATA{Input/cthead1.mha}
-    ${ITK_TEST_OUTPUT_DIR}/itkConvolutionImageFilterStreamingTestOutput.mha
+    ${ITK_TEST_OUTPUT_DIR}/itkConvolutionImageFilterStreamingValidTestOutput.mha
     50
     75
     150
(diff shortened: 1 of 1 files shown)


N-Dekker:
Thanks @blowekamp 

Is there any reason for those tests to share the same output directory? Or would it be better if each test would have its own output directory? Something like `${ITK_TEST_OUTPUT_DIR}/<test_name>` ?

Modules/Filtering/Convolution/test/CMakeLists.txt
```diff
@@ -625,11 +625,11 @@ itk_add_test(
     ITKConvolutionTestDriver
     --compare
     DATA{Baseline/itkConvolutionImageFilterStreamingTestOutput.mha}
-    ${ITK_TEST_OUTPUT_DIR}/itkFFTConvolutionImageFilterStreamingTestOutput.mha
+    ${ITK_TEST_OUTPUT_DIR}/itkFFTConvolutionImageFilterStreamingTestOutputFrequency.mha
```
N-Dekker: Would it be possible to keep the logic of `<test-name>` + "Output.mha"?  When we do it like that, each test will have its own unique output file name. For itkFFTConvolutionImageFilterStreamingTest, it would then still be "itkFFTConvolutionImageFilterStreamingTestOutput.mha" (as it was before).

itkFFTConvolutionImageFilterStreamingValidTest would then have "itkFFTConvolutionImageFilterStreamingValidTestOutput.mha"

Or do you have a specific reason to add the word "Frequency"? (It already has "FFT" as prefix!)

blowekamp: I just use GH Copilot locally to make these changes. But there appears to be an option "frequency" below and it appears to enable additional logic in Modules/Filtering/Convolution/test/itkConvolutionImageFilterStreamingTest.cxx. This seems like a reasonable additional term for the length filename.

N-Dekker: Maybe Copilot does not understand that each test name is already unique. So it appends an extra word that it can find in the neighborhood. Maybe it's hallucinating 🤷

Modules/Filtering/Convolution/test/CMakeLists.txt
```diff
@@ -658,11 +658,11 @@ itk_add_test(
     ITKConvolutionTestDriver
     --compare
     DATA{Baseline/itkConvolutionImageFilterStreamingTestOutput.mha}
-    ${ITK_TEST_OUTPUT_DIR}/itkFFTConvolutionImageFilterStreamingTestOutput.mha
+    ${ITK_TEST_OUTPUT_DIR}/itkFFTConvolutionImageFilterStreamingValidTestOutputFrequency.mha
     itkConvolutionImageFilterStreamingTest
     frequency
```
blowekamp, line 663: @N-Dekker Here is case where the test name is "itkFFTConvolutionImageFilterStreamingValidTest" not the "valid" but here the "frequency" command line argument is used. So there is some  pre-existing inconsistencies that could be corrected to make the <test-name> + "Output.mha" work. But later "valid" is used for the region mode.

N-Dekker (approved):
Thank you @blowekamp 👍

PR #5718, opened 2026-01-06
Make 3D SobelOperator consistent with 2D, add UseLegacyCoefficients, add ND support

N-Dekker:
The old kernel coefficients for 3D appeared inconsistent with the coefficients for 2D. This pull request allows generating these coefficients consistently, inspired by the AI generated function `make_nd_sobel_kernels`, presented by Hans Johnson (@hjmjohnson) at https://github.com/InsightSoftwareConsortium/ITK/issues/5702#issue-3748203291

The new coefficients are more commonly used by other toolkits, including SciPy's `scipy.ndimage.sobel`, and correspond with https://en.wikipedia.org/wiki/Sobel_operator

This PR removes the reference to "An Isotropic 3x3x3 Volume Gradient Operator", Irwin Sobel, 1995, as it cannot be found anymore.

Added a `UseLegacyCoefficients` option to SobelOperator, which allows the user to specify whether or not `SobelOperator::GenerateCoefficients()` should still produce the old coefficients for a 3D kernel. Enabled by default.

Removed the `static_assert` from `SobelOperator` that allowed only 2D and 3D. Added checks for 1D and 4D to the unit test.

- Superseding pull request #5715
- Addresses issue #5702

### DIFF
--- a/Modules/Core/Common/include/itkSobelOperator.h
+++ b/Modules/Core/Common/include/itkSobelOperator.h
@@ -112,10 +112,6 @@ public:
 
   itkOverrideGetNameOfClassMacro(SobelOperator);
 
-  static_assert(
-    VDimension == 2 || VDimension == 3,
-    "The ND version of the Sobel operator has not been implemented. Currently only 2D and 3D versions are available.");
-
   /** Creates the operator with length only in the specified direction.
    * The radius of the operator will be 0 except along the axis on which
    * the operator will work.
--- a/Modules/Core/Common/test/itkSobelOperatorGTest.cxx
+++ b/Modules/Core/Common/test/itkSobelOperatorGTest.cxx
@@ -78,10 +78,28 @@ TEST(SobelOperator, CheckKernelCoordinates)
 {
   for (const bool useLegacyCoordinates : { false, true })
   {
+    CheckKernelCoordinates<1>(useLegacyCoordinates, { ExpectedKernelType{ -1, 0, 1 } });
     CheckKernelCoordinates<2>(
       useLegacyCoordinates,
       { ExpectedKernelType{ -1, 0, 1, -2, 0, 2, -1, 0, 1 }, ExpectedKernelType{ -1, -2, -1, 0, 0, 0, 1, 2, 1 } });
+    CheckKernelCoordinates<4>(
+      useLegacyCoordinates,
+      { ExpectedKernelType{ -1, 0, 1, -2, 0, 2, -1, 0, 1, -2, 0, 2, -4, 0, 4, -2, 0, 2, -1, 0, 1, -2, 0, 2, -1, 0, 1,
+                            -2, 0, 2, -4, 0, 4, -2, 0, 2, -4, 0, 4, -8, 0, 8, -4, 0, 4, -2, 0, 2, -4, 0, 4, -2, 0, 2,
+                            -1, 0, 1, -2, 0, 2, -1, 0, 1, -2, 0, 2, -4, 0, 4, -2, 0, 2, -1, 0, 1, -2, 0, 2, -1, 0, 1 },
+        ExpectedKernelType{ -1, -2, -1, 0, 0, 0, 1, 2, 1, -2, -4, -2, 0, 0, 0, 2, 4, 2, -1, -2, -1, 0, 0, 0, 1, 2, 1,
+                            -2, -4, -2, 0, 0, 0, 2, 4, 2, -4, -8, -4, 0, 0, 0, 4, 8, 4, -2, -4, -2, 0, 0, 0, 2, 4, 2,
+                            -1, -2, -1, 0, 0, 0, 1, 2, 1, -2, -4, -2, 0, 0, 0, 2, 4, 2, -1, -2, -1, 0, 0, 0, 1, 2, 1 },
+        ExpectedKernelType{ -1, -2, -1, -2, -4, -2, -1, -2, -1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 2, 1, 2, 4, 2, 1, 2, 1,
+                            -2, -4, -2, -4, -8, -4, -2, -4, -2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2, 4, 2, 4, 8, 4, 2, 4, 2,
+                            -1, -2, -1, -2, -4, -2, -1, -2, -1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 2, 1, 2, 4, 2, 1, 2, 1 },
+        ExpectedKernelType{ -1, -2, -1, -2, -4, -2, -1, -2, -1, -2, -4, -2, -4, -8, -4, -2, -4, -2, -1, -2, -1,
+                            -2, -4, -2, -1, -2, -1, 0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,
+                            0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  1,  2,  1,  2,  4,  2,  1,  2,  1,
+                            2,  4,  2,  4,  8,  4,  2,  4,  2,  1,  2,  1,  2,  4,  2,  1,  2,  1 } });
   }
+
+  // For 3D, the legacy and the non-legacy coordinates are different.
   CheckKernelCoordinates<3>(
     true,
     { ExpectedKernelType{ -1, 0, 1, -3, 0, 3, -1, 0, 1, -3, 0, 3, -6, 0, 6, -3, 0, 3, -1, 0, 1, -3, 0, 3, -1, 0, 1 },


Modules/Core/Common/include/itkSobelOperator.hxx
```diff
@@ -74,32 +74,11 @@ template <typename TPixel, unsigned int VDimension, typename TAllocator>
...
-      case 1:
-        return { -1, -2, -1, 0, 0, 0, 1, 2, 1 };
-    }
-  }
-  if constexpr (VDimension == 3)
-  {
-    switch (direction)
-    {
-      case 0:
-        return { -1, 0, 1, -2, 0, 2, -1, 0, 1, -2, 0, 2, -4, 0, 4, -2, 0, 2, -1, 0, 1, -2, 0, 2, -1, 0, 1 };
```
dzenanz: Existing tests didn't pass with this change, right?

Modules/Core/Common/include/itkSobelOperator.hxx
```diff
@@ -74,32 +74,11 @@ template <typename TPixel, unsigned int VDimension, typename TAllocator>
...
-      case 1:
-        return { -1, -2, -1, 0, 0, 0, 1, 2, 1, -2, -4, -2, 0, 0, 0, 2, 4, 2, -1, -2, -1, 0, 0, 0, 1, 2, 1 };
-      case 2:
-        return { -1, -2, -1, -2, -4, -2, -1, -2, -1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 2, 1, 2, 4, 2, 1, 2, 1 };
-    }
-  }
-  itkExceptionMacro("The direction value (" << direction << ") should be less than the dimensionality (" << VDimension
-                                            << ").");
+  return CoefficientVector(kernel.cbegin(), kernel.cend());
```
dzenanz: ... and now they pass?

N-Dekker: - The old tests for 3D kernel coefficients (at the current main) fail for both PR #5715 and for this one (#5718). That's indeed the problem. The old 3D coefficients of `itk::SobelOperator` appear uncommon (compared to other toolkits), _and_ difficult to generalize to ND.

However, it would still be nice to find the historical Sobel article that proposed the old 3D coefficients!  "An Isotropic 3x3x3 Volume Gradient Operator", Irwin Sobel, Hewlett-Packard Laboratories, 1995. It must be _somewhere_!

N-Dekker: You see, this PR also adjusts the test for 3D: https://github.com/InsightSoftwareConsortium/ITK/pull/5718/files#diff-f70fdc8e2da55314b48a7a478f0b4cecd816052c2f3122af5b907bf2a22152c8

issakomi: > However, it would still be nice to find the historical Sobel article that proposed the old 3D coefficients! "An Isotropic 3x3x3 Volume Gradient Operator", Irwin Sobel, Hewlett-Packard Laboratories, 1995. It must be somewhere!

There might be something useful in this document:

[History_and_Definition_of_Sobel_Operator.pdf](https://github.com/user-attachments/files/24459734/History_and_Definition_of_Sobel_Operator.pdf)

N-Dekker: Thanks @issakomi Unfortunately [History_and_Definition_of_Sobel_Operator.pdf](https://github.com/user-attachments/files/24459734/History_and_Definition_of_Sobel_Operator.pdf) does not mention a 3D extension of the Sobel Operator  🤷

Modules/Core/Common/include/itkSobelOperator.h
```diff
@@ -134,6 +132,72 @@ class ITK_TEMPLATE_EXPORT SobelOperator : public NeighborhoodOperator<TPixel, VD
...
   void
   Fill(const CoefficientVector & coeff) override;
+
+private:
+  // Create N-dimensional Sobel kernels (one per axis).
+  // Each kernel is size 3^N, stored in row-major order with the last dimension varying fastest.
+  // Standard Sobel definition: derivative = [-1, 0, 1], smoothing = [1, 2, 1].
+  // Kernel for axis a is: K_a(x) = d[x_a] * Π_{j != a} s[x_j], with x_j ∈ {-1,0,1}.
+  inline std::vector<std::vector<int>>
+  make_nd_sobel_kernels(std::size_t N)
```
hjmjohnson: If this function were templated on N (aka VDimension), it could also be constexpr.

Then the test for N == 0 could also be a static assert instead of a runtime test.

N-Dekker: Thanks Hans! I just gave `constexpr` a try: [WIP: Make make_nd_sobel_kernels constexpr](https://github.com/InsightSoftwareConsortium/ITK/pull/5718/commits/e4e7c2287c92102c01eb633f1e871cf085ea89f7) (updated link)

Note that in this context, make_nd_sobel_kernels would just be an internal helper function of `itk::SobelOperator<TPixel, VDimension, TAllocator>`, so it already "knows" VDimension, and there is no need to pass it as parameter. VDimension can never be zero, so the assert isn't really necessary anyway.

Still: this is all WIP. Especially because I don't really _understand_ the AI generated code, do you?

N-Dekker: In the mean time, I think I found an AI hallucination! Looking at the original AI generated code at https://github.com/InsightSoftwareConsortium/ITK/issues/5702#issue-3748203291  This part appears useless:


```cpp
    // Precompute strides for row-major flattening (last dim fastest)
    std::vector<std::size_t> stride(N, 1);
    for (std::size_t i = 1; i < N; ++i) {
        stride[N - 1 - i] = stride[N - i] * 3;
    }
```

The rest of the code does not make use of those strides! So it's unnecessary! I think I'll remove this part of code with the next force-push!

N-Dekker:
FYI My intention is to squash most of these WIP commits before making the PR ready for review. (But before doing so, the PR should also offer a backward compatible legacy option, at least for 3D.)

Modules/Core/Common/include/itkSobelOperator.hxx
```diff
@@ -74,32 +77,68 @@ template <typename TPixel, unsigned int VDimension, typename TAllocator>
...
+            }
+            kernels[axis][linear] = val;
+          }
+        }
+      }
+      return kernels;
+    }();
+
+    const auto & kernel = sobelKernels[VDimension - 1 - direction];
+    return CoefficientVector(kernel.cbegin(), kernel.cend());
```
N-Dekker, line 138: Why not simply `kernel = sobelKernels[direction]`? I still don't understand why the kernels returned by the AI generated function from https://github.com/InsightSoftwareConsortium/ITK/issues/5702#issue-3748203291 are in the reverse order, compared by ITK 🤷

Modules/Core/Common/include/itkSobelOperator.h
```diff
@@ -124,6 +132,21 @@ class ITK_TEMPLATE_EXPORT SobelOperator : public NeighborhoodOperator<TPixel, VD
...
+  {
+    m_UseLegacyCoefficients = useLegacyCoefficients;
+  }
+
+  /** Tells whether or not `GenerateCoefficients()` returns the legacy coordinate values, compatible with ITK <= 5.4. */
+  bool
+  IsUsingLegacyCoefficients() const
+  {
+    return m_UseLegacyCoefficients;
+  }
```
N-Dekker: Note that itkSetMacro(

Modules/Core/Common/include/itkSobelOperator.h
```diff
@@ -124,6 +132,21 @@ class ITK_TEMPLATE_EXPORT SobelOperator : public NeighborhoodOperator<TPixel, VD
    * \sa CreateDirectional \sa Fill */
   // virtual void CreateToRadius(const unsigned long);
 
+  /** Allows specifying whether or not `GenerateCoefficients()` should return the legacy coordinate values, compatible
+   * with ITK <= 5.4. */
+  void
+  UseLegacyCoefficients(const bool useLegacyCoefficients)
+  {
+    m_UseLegacyCoefficients = useLegacyCoefficients;
+  }
```
N-Dekker, line 137: Note that `itkSetMacro(UseLegacyCoefficients, bool)` does not compile, because `itk::SobelOperator` is not derived from `itk::Object`, and it does not support `this->Modified()`.

The macro (itkSetMacro) would make it `SetUseLegacyCoefficients(bool _arg)` instead. But I think `UseLegacyCoefficients(const bool useLegacyCoefficients)` is more readable. OK?

What about the other member function, `IsUsingLegacyCoefficients()`? Is that clear enough?

dzenanz:
Then merge. Unless you are expecting reviews from more people.

N-Dekker:
For the record:

Below here some nice pictures of the kernels of a 3D Sobel operator, as posted by [Sven Fritsch](https://stackoverflow.com/users/12285738/sven-fritsch) at https://stackoverflow.com/a/66698671. Their values appear to correspond with the proposed new coordinates for a 3D `itk::SobelOperator` from this pull request.

> Here the sobel filters in [X direction](https://i.sstatic.net/73lM4.png), [Y direction](https://i.sstatic.net/gtW6o.png) and [Z direction](https://i.sstatic.net/11ekC.png).

73lM4 1

----
gtW6o 1

----
11ekC 1

dzenanz:
As far as I am concerned, you can merge it.

N-Dekker:
Thanks @dzenanz You know I generally prefer not to merge my own pull requests. I feel that it's more rewarding when someone merges my pull request, instead of having to merge it myself. In this particular case, it's not a very large PR (+142, −37 lines of code). It resolves a feature request by Hans and it fixes compile errors encountered by Bradley, so I think it's very relevant. But if nobody feels like merging the PR, I think I'll do so by the end of the week  🤷

PR #5725, opened 2026-01-08
ENH: Update connected component testing baselines

CavRiley:
## Description

This PR updates the baseline images for ConnectedComponent tests following the changes in PR #5622, which replaced `vnl_sample.h` with `<random>`. The switch to the standard library random number generator produced different random sequences, causing the existing baseline images to fail validation.

## Related Issues/PRs

Follow-up to #5622 

## PR Checklist

## Testing

Baseline images were regenerated by running the ConnectedComponent tests with the updated random number generator. All tests now pass with the new baselines.

### DIFF
--- a/Testing/Data/Baseline/BasicFilters/ConnectedComponentImageFilterTest.1.png.cid
+++ b/Testing/Data/Baseline/BasicFilters/ConnectedComponentImageFilterTest.1.png.cid
@@ -1,1 +1,1 @@
-bafkreidtm3ujuq2se3t4evejuuibd2cuk2xbv72yeqemnwo3syno72vjqq
+bafkreicqlhg6xztvfqmgynbf3xb6qhbghwfzvkugt6nwxn2kloiwnfalp4
--- a/Modules/Segmentation/ConnectedComponents/test/itkConnectedComponentImageFilterTest.cxx
+++ b/Modules/Segmentation/ConnectedComponents/test/itkConnectedComponentImageFilterTest.cxx
@@ -22,9 +22,9 @@
 #include "itkImageFileReader.h"
 #include "itkImageFileWriter.h"
 #include "itkSimpleFilterWatcher.h"
-#include "vnl/vnl_sample.h"
 #include "itkTestingMacros.h"
 #include <algorithm> // For generate.
+#include <random>    // For mt19937.
 
 int
 itkConnectedComponentImageFilterTest(int argc, char * argv[])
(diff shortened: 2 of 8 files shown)


Testing/Data/Baseline/BasicFilters/ConnectedComponentImageFilterTest.1.png.cid
```diff
@@ -1 +1 @@
-bafkreidtm3ujuq2se3t4evejuuibd2cuk2xbv72yeqemnwo3syno72vjqq
+bafkreicqlhg6xztvfqmgynbf3xb6qhbghwfzvkugt6nwxn2kloiwnfalp4
```
N-Dekker, line 1: Really cool, Cavan! How do you do that? I haven't yet figured out how to add testing data to ITK. 🤷

dzenanz: He recently updated software guide: https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/pull/230

blowekamp:
What were the .1 alternate bases lines generated?

They have some ### times been different results from different platforms, when the results are not always consistent.

dzenanz:
Ah, well noticed Brad. The baselines should be replaced, not added.

N-Dekker (approved):
Thanks @CavRiley ! Cool!

Approved, but please consider squashing the two commits 😃

blowekamp:
> > What were the .1 alternate bases lines generated?
> > They have some ### times been different results from different platforms, when the results are not always consistent.
> 
> The `.1` alternate baselines were regenerated to account for platform-specific differences in the test outputs.
> 
> I generated the primary `.png` baseline on a Mac machine and the `.1` alternative on a Linux machine. During testing, I noticed that the Mac-generated baseline did not fix the test failures on Linux, which is why the `.1` alternative was changed.
> 
> Let me know if this answered your question!

Thank you for making this contribution!

Do you have any insight why different results are occuring? Is the same random image being generated on the two platforms? Or is the filter just producing different results?

N-Dekker:
> I agree with @N-Dekker in that I would expect the same number sequence on each platform, but my best guess is that `std::uniform_int_distribution` may produce differing values based on platform.

Wow, that's a pity! Thanks @CavRiley! I did not realize before that the result of `std::uniform_int_distribution` is platform-dependent. But now I see: https://www.reddit.com/r/cpp/comments/7i21sn/til_uniform_int_distribution_is_not_portable/

So the random number engine produces the same sequence on all platforms, but the "distribution" does not!

Before our PR, did all platforms share the same baselines?

PR #5729, opened 2026-01-09
COMP: Restore reference ImageAlgorithm::Copy

blowekamp:
Only use the generic ImageRange Copy method for known Image and VectorImage image types. When a mix of image types or another image type such as RLEImage is used fall back to the original reference implementation.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkImageAlgorithm.h
+++ b/Modules/Core/Common/include/itkImageAlgorithm.h
@@ -73,7 +73,9 @@ struct ImageAlgorithm
        const typename InputImageType::RegionType &  inRegion,
        const typename OutputImageType::RegionType & outRegion)
   {
-    ImageAlgorithm::DispatchedCopy(inImage, outImage, inRegion, outRegion);
+    // There are overloaded method of Copy for Image and VectorImage,
+    // this case may be an unknow image type, or mixed image types, so use the reference implementation.
+    ImageAlgorithm::ReferenceCopy(inImage, outImage, inRegion, outRegion);
   }
 
   /// \cond HIDE_SPECIALIZATION_DOCUMENTATION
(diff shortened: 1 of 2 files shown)


Modules/Core/Common/include/itkImageAlgorithm.h
```diff
@@ -73,7 +73,9 @@ struct ImageAlgorithm
        const typename InputImageType::RegionType &  inRegion,
        const typename OutputImageType::RegionType & outRegion)
   {
-    ImageAlgorithm::DispatchedCopy(inImage, outImage, inRegion, outRegion);
+    // There are overloaded method of Copy for Image and VectorImage,
+    // this case may be an unknow image type, or mixed image types, so use the reference implementation.
+    ImageAlgorithm::ReferenceCopy(inImage, outImage, inRegion, outRegion);
```
dzenanz, line 78: Couldn't we do some `if constexpr` conditions on the inImage/outImage types, and in simple cases use the optimized dispatched copy variant?

dzenanz (approved):
Did you check whether this PR gets RLEImage compiling again? If so, looks good as-is. Inline suggestion optional.

Modules/Core/Common/include/itkImageAlgorithm.h
blowekamp: We already have overloaded `ImageAlgorithm::Copy` when the input and output image types are either `itk::image` or `itk::VectorImage` while being the same dimension. So for those cases the optimize DispatchedCopy will still be used.

So what case not covered could still be optimized?  Maybe when they are different dimensions but still a known image type?

dzenanz: Ah, then further optimization is not necessary.

PR #5731, opened 2026-01-16
RecursiveGaussian filters: Use std::array instead of C-array for m_SmoothingFilters, use std::generate and range-based `for`

N-Dekker:
Aims to improve the consistency between the implementations of the five recursive Gaussian filters.

- Follow-up to pull request #5730

### DIFF
--- a/Modules/Filtering/ImageFeature/include/itkHessianRecursiveGaussianImageFilter.hxx
+++ b/Modules/Filtering/ImageFeature/include/itkHessianRecursiveGaussianImageFilter.hxx
@@ -80,11 +80,9 @@ template <typename TInputImage, typename TOutputImage>
 void
 HessianRecursiveGaussianImageFilter<TInputImage, TOutputImage>::SetSigma(RealType sigma)
 {
-  const unsigned int numberOfSmoothingFilters = NumberOfSmoothingFilters;
-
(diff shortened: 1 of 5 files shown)


Modules/Filtering/ImageFeature/include/itkHessianRecursiveGaussianImageFilter.hxx
```diff
@@ -80,12 +80,8 @@ template <typename TInputImage, typename TOutputImage>
...
 HessianRecursiveGaussianImageFilter<TInputImage, TOutputImage>::SetSigma(RealType sigma)
 {
-  const unsigned int numberOfSmoothingFilters = NumberOfSmoothingFilters;
-
-  for (unsigned int i = 0; i < numberOfSmoothingFilters; ++i)
-  {
-    m_SmoothingFilters[i]->SetSigma(sigma);
-  }
+  std::for_each(
+    m_SmoothingFilters.cbegin(), m_SmoothingFilters.cend(), [sigma](const auto & filter) { filter->SetSigma(sigma); });
```
N-Dekker: Sorry, I changed my mind. I think in cases like these, a simple range-based `for` loop would be preferable, instead of calling `std::for_each`. I feel that a simple range-based `for` loop is easier to write, easier to debug, and more readable!

PR #5733, opened 2026-01-16
Address missing dependencies

blowekamp:
Addresses missing module and linking dependencies encountered during refactoring the module system.

These bug fixes can be merged as a separate PR from #5721

### DIFF
--- a/Modules/Core/Common/test/CMakeLists.txt
+++ b/Modules/Core/Common/test/CMakeLists.txt
@@ -220,7 +220,7 @@ target_link_options(
     "$<$<AND:$<C_COMPILER_ID:AppleClang>,$<VERSION_GREATER_EQUAL:$<C_COMPILER_VERSION>,15.0>>:LINKER:-no_warn_duplicate_libraries>"
 )
 itk_module_target_label(itkMathTest)
-target_link_libraries(itkMathTest LINK_PUBLIC ${ITKCommon_LIBRARIES})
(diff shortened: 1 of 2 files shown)


blowekamp:
@hjmjohnson @thewtex @dzenanz  Could some one else explain what they think "COMPILE_DEPENDS" should be doing in terms of linking and include paths for the modules library, and the module's interface of include and libraries for other modules using it?

There are clear cases where a module's libraries has derived itk::LightObjects but are only COMPILE_DEPENDS on ITKCommon. So it's not clear to me the intended difference from just "DEPENDS".

dzenanz (approved):
I believe that COMPILE_DEPENDS are modules needed for compiling, but not for running or header-including during development of applications. If module B does not publicly expose any symbols from header-only module A, then A can be B's compile dependency, otherwise it needs to be a full dependency. If A is not header-only, then it needs to be a full dependency.

Indeed, not many modules go there. Most are either DEPENDS or TEST_DEPENDS.

blowekamp:
> @blowekamp yes, it is for more efficient linking and to help limit library size.
> 
> I don't recall those ITKCommon details.
> 
> Please ping once ready, I will cut a release so we can get folks testing the updates. Really wonderful!

We will have to keep an eye on the library size. To my understanding most of the current ITK module system was designed for library files with library paths, which don't allow CMake to do proper dependency analysis.  I believe when CMake library targets are used CMake is able analyze dependencies to minimize the libraries which occurs on the command line. So I think this feature may not be relevant with the proposed modern CMake interfaces.

PR #5735, opened 2026-01-20
ENH: Add UseLegacyOperatorCoefficients to SobelEdgeDetectionImageFilter

N-Dekker:
Allowed switching off `UseLegacyCoefficients` from the Sobel operators used by SobelEdgeDetectionImageFilter.

Included GoogleTest unit tests.

Follow-up to pull request https://github.com/InsightSoftwareConsortium/ITK/pull/5718 commit 43801b28e751986614565e7266103624750cd44a
"ENH: Make 3D SobelOperator consistent with 2D, add UseLegacyCoefficients"

### DIFF
--- a/Modules/Filtering/ImageFeature/test/itkSobelEdgeDetectionImageFilterGTest.cxx
+++ b/Modules/Filtering/ImageFeature/test/itkSobelEdgeDetectionImageFilterGTest.cxx
@@ -1,0 +1,86 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         https://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+// First include the header file to be tested:
+#include "itkSobelEdgeDetectionImageFilter.h"
+
+#include "itkDeref.h"
+#include "itkImage.h"
+#include "itkImageBufferRange.h"
+
+#include <gtest/gtest.h>
+
+#include <cmath> // For sqrt.
+#include <vector>
+
+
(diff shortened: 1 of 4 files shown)


Modules/Filtering/ImageFeature/test/itkSobelEdgeDetectionImageFilterGTest.cxx
```diff
@@ -0,0 +1,84 @@
...
+    const auto filter = itk::SobelEdgeDetectionImageFilter<InputImageType, OutputImageType>::New();
+    filter->SetInput(inputImage);
+    filter->SetUseLegacyOperatorCoefficients(useLegacyCoefficients);
+    filter->Update();
+    const OutputImageType & outputImage = itk::Deref(filter->GetOutput());
+
+    const itk::ImageBufferRange outputImageBufferRange(outputImage);
+
+    const auto expectedPixelValues =
+      sqrtOfValues(useLegacyCoefficients ? std::vector<double>{ 3,  18, 3,  18, 36, 18, 3, 18, 3,  18, 36, 18, 36, 0,
```
dzenanz, line 76: Maybe add a comment about why we expect square root of coefficients as pixel values?

N-Dekker: Well... it comes from the fact that the `SobelEdgeDetectionImageFilter::::GenerateData()` runs `SqrtImageFilter` as final step of its mini-pipeline:

https://github.com/InsightSoftwareConsortium/ITK/blob/be57162b219711b8ab33641572afa73d0e37a9d2/Modules/Filtering/ImageFeature/include/itkSobelEdgeDetectionImageFilter.hxx#L135-L140

That's not really a mathematical explanation, it's just what it does  🤷 

Maybe there is a  mathematical explanation, though. Do you have one?

dzenanz: I guess the theoretical math has square roots all over, but it is computationally efficient to do the root just once, at the end. So a comment here like `the square root is computed by the filter after convolving the coefficients with the pixels` somewhere in the test would be helpful for someone just looking at the test (like me right now) without having recently looked at the filter implementation 😄

N-Dekker: @dzenanz Please check if this addresses your comment: https://github.com/InsightSoftwareConsortium/ITK/compare/c08bb732a2dfea3ffa732b37944a848b0ffdc63c..e6d4997c5553cdef6815628bd33e4823aa617514

PR #5740, opened 2026-01-23
ENH: Add example project which used ITK Modules and FetchContent

blowekamp:
This is a standalone directly that can be build separately to try out this combination of feature to build a project/executable with ITK.

By use FetchContent for ITK, along with specific ITK module interface library linking enable to build only the targets needed in ITK for a project. While the initial configuration time is longer due to downloading and configuring ITK, the CMake project structure is simpler than a Superbuild, and the build is more efficient.

## PR Checklist

### DIFF
--- a/Examples/HelloFetchITK/CMakeLists.txt
+++ b/Examples/HelloFetchITK/CMakeLists.txt
@@ -1,0 +1,14 @@
+cmake_minimum_required(VERSION 3.22.1...3.29.0 FATAL_ERROR)
+
+# This project is designed to be built outside the Insight source tree.
+project(HelloFetchITK)
+
+# Use an existing ITK installation if specified with ITK_DIR, or fetch and configure ITK.
+include(ITKFetchContents.cmake)
+
+itk_generate_factory_registration()
+
+add_executable(HelloFetchITK FetchWorld.cxx)
+
+# If ITK was fetched and configure, only the required modules are built.
+target_link_libraries(HelloFetchITK ITK::ITKCommonModule)
--- a/Examples/HelloFetchITK/ITKFetchContents.cmake
+++ b/Examples/HelloFetchITK/ITKFetchContents.cmake
@@ -1,0 +1,40 @@
+#-----------------------------------------------------------------------------
+# Get and build ITK using FetchContent
+
+include(FetchContent)
+
+# Set ITK Git repository and tag
+set(ITK_GIT_REPOSITORY "https://github.com/InsightSoftwareConsortium/ITK.git")
+
+set(ITK_GIT_TAG "main")
+
+# Set ITK build options
+set(ITK_BUILD_DEFAULT_MODULES ON)
+set(ITK_USE_KWSTYLE OFF)
+set(BUILD_TESTING OFF)
+set(BUILD_EXAMPLES OFF)
+
+FetchContent_Declare(
+  ITK
+  GIT_REPOSITORY "${ITK_GIT_REPOSITORY}"
+  GIT_TAG "${ITK_GIT_TAG}"
+  EXCLUDE_FROM_ALL
+  FIND_PACKAGE_ARGS
+    NAMES
+    ITK
+)
+
+FetchContent_MakeAvailable(ITK)
+
+# Check if FetchContent used find_package() or fetched from source
+FetchContent_GetProperties(ITK)
+if(ITK_SOURCE_DIR)
+  message(STATUS "ITK fetched from repository and built from source")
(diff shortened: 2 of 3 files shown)


Examples/HelloFetchITK/ITKFetchContents.cmake
```diff
@@ -0,0 +1,48 @@
+#-----------------------------------------------------------------------------
+# Get and build ITK using FetchContent
+
+include(FetchContent)
+
+# Set ITK Git repository and tag
+set(ITK_GIT_REPOSITORY "https://github.com/blowekamp/ITK.git")
```
hjmjohnson: @blowekamp This should probably default to InsightSoftwareConsortium rather than blowekamp organization.

Examples/HelloFetchITK/ITKFetchContents.cmake
```diff
@@ -0,0 +1,48 @@
...
+
+set(ITK_GIT_TAG "cmake_interface_module")
```
hjmjohnson: OK for DRAFT pr, but should this default to "main" or "v5.4.5" or "latest"? as the default?

Examples/HelloFetchITK/ITKFetchContents.cmake
```diff
@@ -0,0 +1,48 @@
...
+
+# Set ITK build options
+set(ITK_BUILD_DEFAULT_MODULES ON)
```
hjmjohnson, line 12: These variables should be conditionally set if not provided on the command line.

if(NOT BUILD_TESTING)
  set(BUILD_TESTING OFF)
endif()

I'm also a little worried that one will not be able use the variable BUILD_TESTING for ITK as off and the BUILD_TESTING for the local project to ON.

Examples/HelloFetchITK/ITKFetchContents.cmake
blowekamp: Yes, it is dependent PR #5721 for the modular interface libraries.

Examples/HelloFetchITK/ITKFetchContents.cmake
blowekamp: This is a minimal working example of the concept. Those variable could be turn on here, and since the targets are not needed then it would not be built. 

You can find a complete application example here:
https://github.com/SimpleITK/SimpleITK/blob/main/CMake/sitkITKFetchContent.cmake

Examples/HelloFetchITK/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+
+# Use an existing ITK installation if specified with ITK_DIR, or fetch and configure ITK.
+include(ITKFetchContents.cmake)
+
+itk_generate_factory_registration()
+
+add_executable(HelloFetchITK FetchWorld.cxx)
+
+# If ITK was fetched and configure, only the required modules are built.
+target_link_libraries(HelloFetchITK ITK::ITKCommonModule)
```
dzenanz, line 14: Seeing this here, I am now thinking that `Module` suffix is going to get repetitive quickly. Can you remind me why not use `ITK::ITKCommon` as the naming pattern? I guess that second ITK is needed to distinguish built-in modules vs. remotes (e.g. `ITK::Montage`).

Examples/HelloFetchITK/ITKFetchContents.cmake
blowekamp: At least during development, it may be best to use a variable that points to the current git hash... This may get tricky with source distributions and keeping this directory independent of the ITK CMAkE environment.

So...likely "main" is the best option.

Examples/HelloFetchITK/CMakeLists.txt
blowekamp: I encourage you to look at an ITKTargets.cmake file to see the library interfaces.

Let's ignore the namespace for now. `ITKCommon` references the compiled library.  However, not every module has a library, and not every module that has a library is named the same as the module. And some modules have multiple libraries.

So the `{itk-module}Module` interface library refers to all the includes, libraries and dependecies of an ITK module. Even when a module does not have a library is still has the includes, and a list of other `{itk-module}Modules` that it depends on. For example a recent change was to use `ITKZLIBModule` for the zlib library. This may reference an imported `ZLIB::ZLIB` target or our third-party `itkzlib` library.

You can see more usage of the new ITK interface pattern in SimpleITK too: https://github.com/SimpleITK/SimpleITK/blob/main/Code/Common/CMakeLists.txt

Issue #5756, opened 2026-02-03
Python pixi build fails with INTERFACE_INCLUDE_DIRECTORIES error


thewtex:
### Description

```
-- Configuring done (15.2s)
CMake Error in Modules/Bridge/NumPy/CMakeLists.txt:
  Target "ITKBridgeNumPyModule" INTERFACE_INCLUDE_DIRECTORIES property
  contains path:

    "/home/matt/src/ITK/.pixi/envs/python/include/python3.13"

  which is prefixed in the source directory.
```

### Steps to Reproduce

On `main`:

```sh
pixi run test-python
```

### Expected behavior

Should be allowed.

### Versions

On `main` following #5721 .

### Additional Information

@blowekamp how do you think this should be addressed?

PR #5765, opened 2026-02-03
Fix BridgeNumpy include and improve module GENEX INCLUDE variables

blowekamp:
Adresses #5756 

- Use generator expression for Python include path for build interface
- Do not add generator expression to module system includes
- Allow modules to override GENEX INCLUDE variables

### DIFF
--- a/Modules/Bridge/NumPy/CMakeLists.txt
+++ b/Modules/Bridge/NumPy/CMakeLists.txt
@@ -1,7 +1,10 @@
 cmake_minimum_required(VERSION 3.16.3)
 project(ITKBridgeNumPy)
 
-set(ITKBridgeNumPy_SYSTEM_INCLUDE_DIRS "${Python3_INCLUDE_DIRS}")
+set(
+  ITKBridgeNumPy_GENEX_SYSTEM_INCLUDE_DIRS
+  "$<BUILD_INTERFACE:${Python3_INCLUDE_DIRS}>"
+)
 
 if(NOT ITK_SOURCE_DIR)
   find_package(ITK REQUIRED)


CMake/ITKModuleMacros.cmake
```diff
@@ -244,7 +244,6 @@ endif()
   endif()
 
   # Prepare include directories with generator expressions for use in targets
-  set(${itk-module}_GENEX_INCLUDE_DIRS "")
```
dzenanz, line 247: These look like removals. Is that intended?

blowekamp: Yes! It allows module code to set these variable like what is done in the BridgeNumpy module. Currently modules can set `${itk-module}-INCLUDE_DIRS` so these variable are now similar

CMake/ITKModuleMacros.cmake
```diff
@@ -254,16 +253,11 @@ endif()
     "$<INSTALL_INTERFACE:$<INSTALL_PREFIX>/${${itk-module}_INSTALL_INCLUDE_DIR}>"
   )
 
-  # Prepare system include directories with generator expressions
-  set(${itk-module}_SYSTEM_GENEX_INCLUDE_DIRS "")
+  # Sytem include directories are assumed to be external dependencies that are not installed,
```
thewtex: System

CMake/ITKModuleMacros.cmake
```diff
@@ -254,16 +253,11 @@ endif()
...
+  # and thus do not have separte install interface paths.
```
thewtex: separate

blowekamp: done

PR #5768, opened 2026-02-05
In the future, no longer use legacy Sobel operator coefficients by default

N-Dekker:
When `ITK_FUTURE_LEGACY_REMOVE` is enabled, ITK will use the new and more consistent Sobel operator coordinates by default (instead of the legacy coordinates), with this pull request. This may affect results for 3D applications that are using SobelOperator, SobelEdgeDetectionImageFilter, or GradientDifferenceImageToImageMetric.

- Follow-up to pull request #5718 commit 43801b28e751986614565e7266103624750cd44a "ENH: Make 3D SobelOperator consistent with 2D, add UseLegacyCoefficients"

### DIFF
--- a/Modules/Core/Common/include/itkSobelOperator.h
+++ b/Modules/Core/Common/include/itkSobelOperator.h
@@ -158,7 +158,11 @@ protected:
   Fill(const CoefficientVector & coeff) override;
 
 private:
+#ifdef ITK_FUTURE_LEGACY_REMOVE
+  bool m_UseLegacyCoefficients{ false };
+#else
   bool m_UseLegacyCoefficients{ true };
+#endif
 };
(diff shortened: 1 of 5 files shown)


blowekamp:
This is probably fine, but I had a thought....

This change just changes the default, it does not produce any warnings or information to the developer or user that the setting is depricated. I am not sure if the end goal here is to keep the legacy option forever or eventually remove it? Or just assist in migrating to the new option.

Maybe in the "FUTURE" the option for this setting is removed? ( API warning when it's used? )
Maybe a warning should be produced (once) when the legacy option is enabled? (Respecting LEGACY_SILENT)

Just a thought, unfortunately it may add some unwelcome complication to this state variable.

PR #5771, opened 2026-02-05
Fix DCMTK support for ITK:: namespace in CMake

blowekamp:
Closes SimpleITK/SimpleITK#2513 

Addresses the following error:
```
The link interface of target "ITKDCMTKModule" contains: ITK::dcmdata but the target was not found.
```


dzenanz (approved):
Looks good on a glance. Thank you Brad.

thewtex (approved):

PR #5780, opened 2026-02-07
ENH: removed threading-unfriendly static var that didn't do much

seanm:


Modules/IO/NIFTI/src/itkNiftiImageIO.cxx
```diff
@@ -1072,14 +1071,9 @@ NiftiImageIO::ReadImageInformation()
   }
 
   m_Holder->ptr.reset(nifti_image_read(this->GetFileName(), false));
-  static std::string prev;
-  if (prev != this->GetFileName())
```
dzenanz, line 1076: Shouldn't this be equivalent to:
```suggestion
  if (this->GetFileName() != "")
```

N-Dekker: It looks like the original intention was to avoid dumping the header of one and the same file more than one, consecutively. It wasn't meant to just test that the file name is non-empty.

.cmake_targets.txt
N-Dekker, line 1: @hjmjohnson Is this large ".cmake_targets.txt" file really necessary here, or was it just added accidentally?

dzenanz: Aside from this, PR looks good.

Modules/IO/NIFTI/test/itkNiftiImageIOTest.cxx
```diff
@@ -144,6 +396,11 @@ itkNiftiImageIOTest(int argc, char * argv[])
     for (int imagenameindex = 1; imagenameindex < argc; ++imagenameindex)
     {
       auto fileName = std::string(argv[imagenameindex]);
+#if defined(ITK_USE_VERY_VERBOSE_NIFTI_DEBUGGING)
+      {
+        DumpNiftiHeader(fileName);
+      }
+#endif
```
N-Dekker, line 394: Thanks Hans, but if the header dumping isn't really useful anymore, I would prefer to simply remove it. It's hard to maintain code that is always `#ifdef`-ed out.  🤷 People who really want to see it back can always check the git history!

seanm: Not sure I agree. It can be handy to enable when you are debugging.

hjmjohnson: @N-Dekker The ifdef'ed code only needs to be maintained by a developer doing debugging.  If it is in the git history, that indicates you know where to look for it.  If the decision is to remove or always use, then my vote is to turn on verbose debugging (which will spam the nightly logs).  This code has been very useful to debug challenging NIFTI binary compatibility issues.

PR #5787, opened 2026-02-10
Add `ComputeIndex()` member function to ImageConstIterator

N-Dekker:
Provides an alternative to `ImageConstIterator::GetIndex()`, making it more clear that potentially expensive computation is involved.

### DIFF
--- a/Modules/Filtering/DisplacementField/include/itkInvertDisplacementFieldImageFilter.hxx
+++ b/Modules/Filtering/DisplacementField/include/itkInvertDisplacementFieldImageFilter.hxx
@@ -194,7 +194,7 @@ InvertDisplacementFieldImageFilter<TInputImage, TOutputImage>::DynamicThreadedGe
       }
       update = ItI.Get() + update * this->m_Epsilon;
       ItI.Set(update);
-      typename DisplacementFieldType::IndexType index = ItI.GetIndex();
+      typename DisplacementFieldType::IndexType index = ItI.ComputeIndex();
       if (this->m_EnforceBoundaryCondition)
       {
         for (unsigned int d = 0; d < ImageDimension; ++d)
--- a/Modules/Filtering/DisplacementField/include/itkIterativeInverseDisplacementFieldImageFilter.hxx
+++ b/Modules/Filtering/DisplacementField/include/itkIterativeInverseDisplacementFieldImageFilter.hxx
@@ -98,7 +98,7 @@ IterativeInverseDisplacementFieldImageFilter<TInputImage, TOutputImage>::Generat
     while (!OutputIt.IsAtEnd())
     {
       // get the output image index
-      const OutputImageIndexType index = OutputIt.GetIndex();
+      const OutputImageIndexType index = OutputIt.ComputeIndex();
       OutputImagePointType       originalPoint;
       outputPtr->TransformIndexToPhysicalPoint(index, originalPoint);
 
--- a/Modules/Filtering/ImageFeature/include/itkHoughTransform2DLinesImageFilter.hxx
+++ b/Modules/Filtering/ImageFeature/include/itkHoughTransform2DLinesImageFilter.hxx
@@ -282,8 +282,8 @@ HoughTransform2DLinesImageFilter<TInputPixelType, TOutputPixelType>::GetLines() 
           // Create the line.
           LineType::LinePointListType list; // Insert two points per line.
 
-          const double radius = it_input.GetIndex()[0];
-          const double teta = ((it_input.GetIndex()[1]) * 2 * Math::pi / this->GetAngleResolution()) - Math::pi;
+          const double radius = it_input.ComputeIndex()[0];
+          const double teta = ((it_input.ComputeIndex()[1]) * 2 * Math::pi / this->GetAngleResolution()) - Math::pi;
           const double Vx = radius * std::cos(teta);
           const double Vy = radius * std::sin(teta);
           const double norm = std::sqrt(Vx * Vx + Vy * Vy);
@@ -326,8 +326,8 @@ HoughTransform2DLinesImageFilter<TInputPixelType, TOutputPixelType>::GetLines() 
           {
             for (double length = 0; length < m_DiscRadius; length += 1)
             {
-              index[0] = static_cast<IndexValueType>(it_input.GetIndex()[0] + length * std::cos(angle));
-              index[1] = static_cast<IndexValueType>(it_input.GetIndex()[1] + length * std::sin(angle));
+              index[0] = static_cast<IndexValueType>(it_input.ComputeIndex()[0] + length * std::cos(angle));
+              index[1] = static_cast<IndexValueType>(it_input.ComputeIndex()[1] + length * std::sin(angle));
               if (postProcessImage->GetBufferedRegion().IsInside(index))
               {
                 postProcessImage->SetPixel(index, 0);
--- a/Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.hxx
+++ b/Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.hxx
@@ -352,10 +352,10 @@ ContourExtractor2DImageFilter<TInputImage>::GenerateDataForLabels()
     for (inputIt.GoToBegin(); !inputIt.IsAtEnd(); ++inputIt)
     {
       BoundingBoxType & bbox = labelBoundingBoxes[inputIt.Get()];
-      bbox.min[0] = std::min(bbox.min[0], inputIt.GetIndex()[0]);
-      bbox.min[1] = std::min(bbox.min[1], inputIt.GetIndex()[1]);
-      bbox.max[0] = std::max(bbox.max[0], inputIt.GetIndex()[0]);
-      bbox.max[1] = std::max(bbox.max[1], inputIt.GetIndex()[1]);
+      bbox.min[0] = std::min(bbox.min[0], inputIt.ComputeIndex()[0]);
+      bbox.min[1] = std::min(bbox.min[1], inputIt.ComputeIndex()[1]);
+      bbox.max[0] = std::max(bbox.max[0], inputIt.ComputeIndex()[0]);
+      bbox.max[1] = std::max(bbox.max[1], inputIt.ComputeIndex()[1]);
     }
     // Build the extended regions from the bounding boxes
     for (const InputPixelType label : allLabels)
--- a/Modules/Segmentation/Watersheds/include/itkWatershedSegmenter.hxx
+++ b/Modules/Segmentation/Watersheds/include/itkWatershedSegmenter.hxx
@@ -358,13 +358,13 @@ Segmenter<TInputImage>::CollectBoundaryInformation(flat_region_table_t & flatReg
             flr.bounds_min = flrt_it->second.bounds_min;
             flr.min_label = *(flrt_it->second.min_label_ptr);
             flr.value = flrt_it->second.value;
-            flr.offset_list.push_back(face->ComputeOffset(faceIt.GetIndex()));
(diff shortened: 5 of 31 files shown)


Modules/Core/Common/include/itkImageConstIterator.h
```diff
@@ -295,16 +295,24 @@ class ITK_TEMPLATE_EXPORT ImageConstIterator
...
   }
 
-  /** Get the index. This provides a read only reference to the index.
-   * This causes the index to be calculated from pointer arithmetic and is
-   * therefore an expensive operation.
+  /** Computes the index. Internally calls ImageBase::ComputeIndex, which may be a relatively expensive operation.
    * \sa SetIndex */
   [[nodiscard]] IndexType
-  GetIndex() const
+  ComputeIndex() const
```
dzenanz, line 301: What is the use-case for this? All the other iterators have `GetIndex()`, so someone calling this method instead seems unlikely. It increases the library interface. I feel that updating the documentation of the existing method would have been sufficient.

N-Dekker: I think users expect a `Get` member function to be fast. And users don't get triggered to use "IteratorWithIndex", when the corresponding "Iterator" class (_without_ index data member) already provides a `GetIndex()` member function.

dzenanz: We should mention `IteratorWithIndex` in the docs of `GetIndex()` in iterators without index classes. That might be better at educating users than adding another method to "slow" classes.

N-Dekker: Look here, HoughTransform2DLinesImageFilter calls `it_input.GetIndex()` multiple times on the very same position:

https://github.com/InsightSoftwareConsortium/ITK/blob/6c9a7bec1c46dc3469fd83894694cdb405b039b1/Modules/Filtering/ImageFeature/include/itkHoughTransform2DLinesImageFilter.hxx#L285-L286
https://github.com/InsightSoftwareConsortium/ITK/blob/6c9a7bec1c46dc3469fd83894694cdb405b039b1/Modules/Filtering/ImageFeature/include/itkHoughTransform2DLinesImageFilter.hxx#L329-L330

When the member function would be named `ComputeIndex()`, users would be more likely to do the same computation just once.

dzenanz: It is impossible to prevent users from making mistakes. And I don't think adding a method (without removing GetIndex) would help.

N-Dekker: In the following case, LevelSetNeighborhoodExtractor, the index is computed for each and every pixel in the region. But it's not clear from the use case that a computation takes place. In this case, "IteratorWithIndex" would probably be much faster. (It currently still uses an iterator without index data member.)

https://github.com/InsightSoftwareConsortium/ITK/blob/6c9a7bec1c46dc3469fd83894694cdb405b039b1/Modules/Segmentation/LevelSets/include/itkLevelSetNeighborhoodExtractor.hxx#L142

N-Dekker: > I don't think adding a method (without removing GetIndex) would help.

I think there are valid use cases for `ComputeIndex()`. It should only be avoided to be called "for each pixel". And it should be avoided to be called multiple times on the _same_ location.

Eventually it might be preferable to remove GetIndex. But first we need to have an alternative, for those valid use cases!

dzenanz: This change should be discussed with other maintainers. I assume there are multiple places which take iterator as a parameter, and then call `GetIndex()` on it. Should those places be changed to call `ComputeIndex()`? That seems like a lot of work, for not a lot of benefit.

N-Dekker: Posted at discourse.itk.org: [**Proposing `ImageConstIterator::ComputeIndex()`, a clearer alternative to GetIndex()**](https://discourse.itk.org/t/proposing-imageconstiterator-computeindex-a-clearer-alternative-to-getindex/7711)

Modules/Filtering/DisplacementField/include/itkInvertDisplacementFieldImageFilter.hxx
```diff
@@ -194,7 +194,7 @@ InvertDisplacementFieldImageFilter<TInputImage, TOutputImage>::DynamicThreadedGe
       }
       update = ItI.Get() + update * this->m_Epsilon;
       ItI.Set(update);
-      typename DisplacementFieldType::IndexType index = ItI.GetIndex();
+      typename DisplacementFieldType::IndexType index = ItI.ComputeIndex();
       if (this->m_EnforceBoundaryCondition)
```
N-Dekker, line 198: In this case, it might be better to declare `ItI` as `ImageRegionIteratorWithIndex`!

Modules/Filtering/DisplacementField/include/itkIterativeInverseDisplacementFieldImageFilter.hxx
```diff
@@ -98,7 +98,7 @@ IterativeInverseDisplacementFieldImageFilter<TInputImage, TOutputImage>::Generat
     while (!OutputIt.IsAtEnd())
     {
       // get the output image index
-      const OutputImageIndexType index = OutputIt.GetIndex();
+      const OutputImageIndexType index = OutputIt.ComputeIndex();
```
N-Dekker, line 101: In this case, it might be better to declare `OutputIt` as an iterator "WithIndex".

Modules/Filtering/ImageFeature/include/itkHoughTransform2DLinesImageFilter.hxx
```diff
@@ -282,8 +282,8 @@ HoughTransform2DLinesImageFilter<TInputPixelType, TOutputPixelType>::GetLines()
           // Create the line.
           LineType::LinePointListType list; // Insert two points per line.
 
-          const double radius = it_input.GetIndex()[0];
-          const double teta = ((it_input.GetIndex()[1]) * 2 * Math::pi / this->GetAngleResolution()) - Math::pi;
+          const double radius = it_input.ComputeIndex()[0];
+          const double teta = ((it_input.ComputeIndex()[1]) * 2 * Math::pi / this->GetAngleResolution()) - Math::pi;
```
N-Dekker, line 286: In this case, it would be faster to retrieve the index just one, instead of twice.

Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.hxx
```diff
@@ -352,10 +352,10 @@ ContourExtractor2DImageFilter<TInputImage>::GenerateDataForLabels()
...
     {
       BoundingBoxType & bbox = labelBoundingBoxes[inputIt.Get()];
-      bbox.min[0] = std::min(bbox.min[0], inputIt.GetIndex()[0]);
-      bbox.min[1] = std::min(bbox.min[1], inputIt.GetIndex()[1]);
-      bbox.max[0] = std::max(bbox.max[0], inputIt.GetIndex()[0]);
-      bbox.max[1] = std::max(bbox.max[1], inputIt.GetIndex()[1]);
+      bbox.min[0] = std::min(bbox.min[0], inputIt.ComputeIndex()[0]);
+      bbox.min[1] = std::min(bbox.min[1], inputIt.ComputeIndex()[1]);
+      bbox.max[0] = std::max(bbox.max[0], inputIt.ComputeIndex()[0]);
+      bbox.max[1] = std::max(bbox.max[1], inputIt.ComputeIndex()[1]);
```
N-Dekker, line 358: Here I think the index of `inputIt` should be retrieved only once, instead of 4x. But moreover,  `inputIt` should probably be declared as an iterator "WithIndex".

dzenanz:
Possibly also take the first 2 non-contentious commits to the separate PR, which should then be merged quickly.

Modules/Filtering/ImageFeature/include/itkHoughTransform2DLinesImageFilter.hxx
N-Dekker: To be considered for a follow-up pull request!

Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.hxx
N-Dekker: To be considered for a follow-up pull request!

Modules/Filtering/DisplacementField/include/itkIterativeInverseDisplacementFieldImageFilter.hxx
N-Dekker: To be considered for a follow-up pull request!

Modules/Filtering/DisplacementField/include/itkInvertDisplacementFieldImageFilter.hxx
N-Dekker: To be considered for a follow-up pull request!

Modules/Filtering/ImageFeature/include/itkHoughTransform2DLinesImageFilter.hxx
```diff
@@ -326,8 +326,8 @@ HoughTransform2DLinesImageFilter<TInputPixelType, TOutputPixelType>::GetLines()
           {
             for (double length = 0; length < m_DiscRadius; length += 1)
             {
-              index[0] = static_cast<IndexValueType>(it_input.GetIndex()[0] + length * std::cos(angle));
-              index[1] = static_cast<IndexValueType>(it_input.GetIndex()[1] + length * std::sin(angle));
+              index[0] = static_cast<IndexValueType>(it_input.ComputeIndex()[0] + length * std::cos(angle));
+              index[1] = static_cast<IndexValueType>(it_input.ComputeIndex()[1] + length * std::sin(angle));
```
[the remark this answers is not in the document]
N-Dekker: This pull request just replaces GetIndex() calls with ComputeIndex(), to raise the awareness of its computational cost. Other code adjustments are beyond the scope of this pull request. I want to keep this pull request easy to process, so that it can be merged smoothly.

Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.hxx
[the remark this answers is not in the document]
N-Dekker: Again, I consider such code adjustments (caching the result of `ComputeIndex()`) beyond the scope of this pull request. But I would like to have another look once this pull request is merged.

Modules/Core/Common/include/itkImageConstIterator.h
```diff
@@ -295,16 +295,24 @@ class ITK_TEMPLATE_EXPORT ImageConstIterator
...
   {
```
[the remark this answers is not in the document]
N-Dekker: `ComputeIndex()` already has full code coverage, because it is called by `GetIndex()`. However, I would like to directly test `ComputeIndex()` with the follow-up that I'm considering, once this PR is merged.

Modules/Segmentation/Watersheds/include/itkWatershedSegmenter.hxx
```diff
@@ -820,8 +820,8 @@ Segmenter<TInputImage>::GradientDescent(InputImageTypePointer img, ImageRegionTy
   {
     if (it.Get() == NULL_LABEL)
     {
-      valueIt.SetLocation(it.GetIndex());
-      labelIt.SetLocation(it.GetIndex());
+      valueIt.SetLocation(it.ComputeIndex());
+      labelIt.SetLocation(it.ComputeIndex());
```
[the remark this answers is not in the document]
N-Dekker: Copilot, you are repeating yourself.

hjmjohnson (approved):

PR #5789, opened 2026-02-10
Address namespace issue with GoogleTest libraries

blowekamp:
- Add depreciation warning for the removed GTest::GTest and GTest::Main targets.
- Correct the namespace to GTest in ITKTargets.

### DIFF
--- a/Modules/ThirdParty/GoogleTest/src/CMakeLists.txt
+++ b/Modules/ThirdParty/GoogleTest/src/CMakeLists.txt
@@ -68,12 +68,14 @@ if(NOT ITK_LEGACY_REMOVE)
   )
 endif()
 
-itk_module_target(gtest NO_INSTALL)
-itk_module_target(gtest_main NO_INSTALL)
+itk_module_target(gtest NO_INSTALL NAMESPACE GTest::)
+itk_module_target(gtest_main NO_INSTALL NAMESPACE GTest::)
 
-mark_as_advanced(gtest_build_samples)
-mark_as_advanced(gtest_build_tests)
-mark_as_advanced(gtest_disable_pthreads)
-mark_as_advanced(gtest_hide_internal_symbols)
-mark_as_advanced(INSTALL_GTEST)
-mark_as_advanced(GTEST_HAS_ABSL)
+mark_as_advanced(
+  gtest_build_samples
+  gtest_build_tests
+  gtest_disable_pthreads
+  gtest_hide_internal_symbols
+  INSTALL_GTEST
+  GTEST_HAS_ABSL
+)


Modules/ThirdParty/GoogleTest/src/CMakeLists.txt
```diff
@@ -47,12 +44,12 @@ if (NOT ITK_LEGACY_REMOVE)
...
+itk_module_target(gtest_main NO_INSTALL NAMESPACE GTest::)
 
-mark_as_advanced(gtest_build_samples)
-mark_as_advanced(gtest_build_tests)
-mark_as_advanced(gtest_disable_pthreads)
-mark_as_advanced(gtest_hide_internal_symbols)
-mark_as_advanced(INSTALL_GTEST)
-mark_as_advanced(GTEST_HAS_ABSL)
+mark_as_advanced(gtest_build_samples
+ gtest_build_tests
```
dzenanz:
```suggestion
gtest_build_tests
```

Modules/ThirdParty/GoogleTest/src/CMakeLists.txt
```diff
@@ -29,15 +29,27 @@ endif()
...
-
-mark_as_advanced(gtest_build_samples)
-mark_as_advanced(gtest_build_tests)
-mark_as_advanced(gtest_disable_pthreads)
-mark_as_advanced(gtest_hide_internal_symbols)
-mark_as_advanced(INSTALL_GTEST)
-mark_as_advanced(GTEST_HAS_ABSL)
+if (NOT ITK_LEGACY_REMOVE)
+  add_library(GTest::GTest INTERFACE IMPORTED)
+  add_library(GTest::Main INTERFACE IMPORTED)
```
N-Dekker, line 52: This would not work for the elastix ITK6 build, at least for now, because elastix always builds with ITK_LEGACY_REMOVE 🤔

Do you propose all those changes, just to make elastix happy? I hope not  😇 If it's just for elastix, we might as well simply adjust elastix with some dedicated CMake code, `if ITK6 ...`

blowekamp: Just the first commit: ec0835979d6f466102c2ffbd287f41043057f013 was needed for your request.
The condition could be changed to ITK_FUTURE_LEGACY_REMOVE if that is prefered.

The rest is needed to fix "ITK::gtest" to be "GTest::gtest" ( and similar to gtest_main) in the ITKTargets.cmake file for exporting in the build directory.

Modules/ThirdParty/GoogleTest/src/CMakeLists.txt
blowekamp: Comprehensively addressed with pre-commit changes in #5790

Modules/ThirdParty/GoogleTest/src/CMakeLists.txt
blowekamp: As noted in the PR  #5791, this alias was marked deprecated in CMake 3.20, and remove in 4.1.0. Given that I think the "ITK_LEGACY_REMOVE" is appropriate. Please let me know your preference for this code block

N-Dekker: I'm sorry, "ITK_LEGACY_REMOVE" won't work for elastix, because _for the time being_ we want to be able to build elastix with "ITK_LEGACY_REMOVE", both with ITK 5.4 and ITK 6.0

I'm now trying the following in elastix, at https://github.com/SuperElastix/elastix/compare/main...Use-ITK-v6.0b02

```cmake
if (ITK_VERSION VERSION_LESS_EQUAL 5.4.5)
  target_link_libraries(CommonGTest
    GTest::GTest
    GTest::Main
    ${ITK_LIBRARIES}
    elastix_lib
  )
else()
  target_link_libraries(CommonGTest
    GTest::gtest
    GTest::gtest_main
    ${ITK_LIBRARIES}
    elastix_lib
  )
endif()

```

Basically just a simple `if (ITK <= 5.4.5)`. I know, it isn't really a "sexy" piece of code, but it should work that way, right?

blowekamp: I am looking to move the changes of this PR forward. Do you have any recommendations of this PR?

For Elastix, if there is another patch release on 5.4 this would not work. I believe the logic "<6.0" would be better.

blowekamp: Changes to this code can be made as needed after it's merged.

hjmjohnson: I think we should move forward with this fix, and address the older issues with a separate PR.

PR #5795, opened 2026-02-12
ENH: Replace 'char' with int8_t when numeric is intended

hjmjohnson:
The 'char' type can be signed or unsigned on different platforms
and is generally used for ASCII character storage.  To conistently
represent 8-bit signed values on differnt platforms the int8_t
type is recommended.

## PR Checklist

### DIFF
--- a/Modules/Filtering/ImageIntensity/test/itkClampImageFilterTest.cxx
+++ b/Modules/Filtering/ImageIntensity/test/itkClampImageFilterTest.cxx
@@ -150,7 +150,7 @@ bool
 TestClampFrom()
 {
   const bool success =
-    TestClampFromTo<TInputPixelType, char>() && TestClampFromTo<TInputPixelType, unsigned char>() &&
+    TestClampFromTo<TInputPixelType, int8_t>() && TestClampFromTo<TInputPixelType, unsigned char>() &&
     TestClampFromTo<TInputPixelType, short>() && TestClampFromTo<TInputPixelType, unsigned short>() &&
     TestClampFromTo<TInputPixelType, int>() && TestClampFromTo<TInputPixelType, unsigned int>() &&
     TestClampFromTo<TInputPixelType, long>() && TestClampFromTo<TInputPixelType, unsigned long>() &&
@@ -263,7 +263,7 @@ template <typename TInputPixelType>
 bool
 TestClampFromWithCustomBounds()
 {
-  const bool success = TestClampFromToWithCustomBounds<TInputPixelType, char>() &&
+  const bool success = TestClampFromToWithCustomBounds<TInputPixelType, int8_t>() &&
                        TestClampFromToWithCustomBounds<TInputPixelType, unsigned char>() &&
                        TestClampFromToWithCustomBounds<TInputPixelType, short>() &&
                        TestClampFromToWithCustomBounds<TInputPixelType, unsigned short>() &&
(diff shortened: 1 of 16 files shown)


Modules/Filtering/ImageIntensity/test/itkClampImageFilterTest.cxx
```diff
@@ -290,12 +290,12 @@ itkClampImageFilterTest(int, char *[])
   auto filter = FilterType::New();
   ITK_EXERCISE_BASIC_OBJECT_METHODS(filter, ClampImageFilter, UnaryFunctorImageFilter);
 
-  const bool success = TestClampFrom<char>() && TestClampFrom<unsigned char>() && TestClampFrom<short>() &&
+  const bool success = TestClampFrom<int8_t>() && TestClampFrom<unsigned char>() && TestClampFrom<short>() &&
```
N-Dekker, line 293: In this context, I would rather use `signed char`, because it also has `unsigned char`. (Although in practice `int8_t` is probably just an alias of `signed char`.)

Modules/IO/NRRD/test/itkNrrdImageIOTest.cxx
```diff
@@ -47,7 +47,7 @@ itkNrrdImageIOTest(int argc, char * argv[])
   int           ret = EXIT_SUCCESS;
 
   ret += itkNrrdImageIOTestReadWriteTest<unsigned char, 2>(std::string(argv[1]), sz, inputFile);
-  ret += itkNrrdImageIOTestReadWriteTest<char, 2>(std::string(argv[1]), sz, inputFile);
+  ret += itkNrrdImageIOTestReadWriteTest<int8_t, 2>(std::string(argv[1]), sz, inputFile);
   ret += itkNrrdImageIOTestReadWriteTest<unsigned short, 2>(std::string(argv[1]), sz, inputFile);
```
N-Dekker, line 51: Here I would also rather use `signed char`, because of the context. (No show-stopper, though.)

blowekamp: What is interesting about this code is that ( if I am correct ) `int8_t` comes from the itk namespace, through argument dependent look up. If this was an example I might request to be explicit,  but this is fine!

N-Dekker: @blowekamp In practice, `int8_t` usually compiles without `std::`, even outside the `itk` namespace. For example at https://godbolt.org/z/a49qez1vz

```cpp
#include <cstdint>

constexpr int8_t x{}; // Usually compiles in C++, even without `std::`.
```

However, the C++ standard does _not_ guarantee that it compiles, without `std::`. So we placed integer aliases like `int8_t` in the `itk` namespace, just to ensure platform independence.

Modules/Filtering/ImageIntensity/test/itkClampImageFilterTest.cxx
hjmjohnson: See the comment below for the rationale for this partial change here.  

I had also implemented
```c+++
TestClampFrom<int8_t>() 
TestClampFrom<uint8_t>() 
TestClampFrom<signed char>() 
TestClampFrom<unsigned char>() 
```
At one point, and perhaps that is what we should put in the final test.

My personal preference is for only
```c+++
TestClampFrom<int8_t>() 
TestClampFrom<uint8_t>() 
```

Modules/IO/NRRD/test/itkNrrdImageIOTest.cxx
hjmjohnson: @blowekamp Thanks.  
FYI:  I previously had a temporary 'poison pill' in the itk::Image class that threw a compiler error if <char> was used as a pixel type.  Thank you for reminding me that I needed to remove that.

N-Dekker:
Thanks for explaining, Hans.

> * Allow 'unsigned char' and 'signed char' when the intent is bytes (i.e., for bit encodings, or enumerations).

When the intent is just raw bytes, it might be preferable to use [std::byte](https://en.cppreference.com/w/cpp/types/byte.html), as introduced with C++17 😇

PR #5797, opened 2026-02-12
ENH: Add constexpr overloads for `itk::Math::abs`

hjmjohnson:
Include integral and floating-point support

Refactored `itk::Math::abs` c++ constexpr,
supporting both integral and floating-point types. Improved compile-time
evaluation with `constexpr`. Enhanced test coverage for `itk::Math::abs`.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/include/itkMath.h
+++ b/Modules/Core/Common/include/itkMath.h
@@ -29,7 +29,9 @@
 #define itkMath_h
 
 #include <cassert>
+#include <complex>
 #include <cmath>
+#include <type_traits>
 #include "itkMathDetail.h"
 #include "itkConceptChecking.h"
 #include <vnl/vnl_math.h>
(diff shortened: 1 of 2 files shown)


Modules/Core/Common/include/itkMath.h
```diff
@@ -825,55 +826,49 @@ using vnl_math::squared_magnitude;
...
+constexpr bool
+abs(bool x) noexcept
 {
   return x;
 }
-inline unsigned long
-abs(unsigned long x)
+/* must use trailing return type to avoid build error for types
+ * where std::make_unsigned_t<T> is not defined (i.e. double, float, etc) */
+template <typename T, typename = std::enable_if_t<std::is_integral_v<T>>>
```
N-Dekker: In general, I find SFINAE (`std::enable_if_t`) quite tricky. Instead of SFINAE + overloading, did you consider having just a single `abs(T x)` function, that supports both integers and floats? Using `if constexpr (std::is_integral_v<T>)` and `if constexpr (std::is_floating_point_v<T>)`?

Modules/Core/Common/include/itkMath.h
```diff
@@ -825,55 +826,49 @@ using vnl_math::squared_magnitude;
...
+
+constexpr bool
+abs(bool x) noexcept
 {
   return x;
 }
-inline unsigned long
-abs(unsigned long x)
+/* must use trailing return type to avoid build error for types
+ * where std::make_unsigned_t<T> is not defined (i.e. double, float, etc) */
```
N-Dekker: This comment appears outdated. The proposed `abs(T x)` does _not_ use a trailing return type.

Modules/Core/Common/include/itkMath.h
hjmjohnson: I don't think that would be the correct approach.  We can not guarantee that T will be limited to integer and floating-point types.  If 'T' is another type, the proposed template may not behave correctly.

blowekamp: Consider:

```
{
if constexpr (std::enable_if_t<std::is_integral_v<T>>)
{...}
elseif constexpr (std::enable_if_t<std::is_floating_point_v<T>
{...}
else
{static_assert(...);}
}
```

N-Dekker: @blowekamp Indeed, that's what I had in mind! But without the `std::enable_if_t`, of course  😄

blowekamp: I haven't finished my first cup of coffee yet 🤣 

I was able to refactor SimpleITK to remove all the enable_if/SFINAE with these constexpr statement, it's really an extremely powerful tool. Might might a good project from someone (student level?/AI?) in ITK to make a similar change in ITK. It's so much more readable!

hjmjohnson: @blowekamp  Thanks Brad.  I'll try that out.

Modules/Core/Common/include/itkMath.h
```diff
@@ -825,55 +826,48 @@ using vnl_math::squared_magnitude;
...
+
+template <typename T, typename = std::enable_if_t<std::is_floating_point_v<T>>>
+constexpr T
+abs(T x) noexcept
 {
-  return x;
+  return x < 0 ? -x : x;
 }
 
 using std::abs;
```
N-Dekker: Just for my understanding, why do we do `using std::abs` here?  I guess that means that for `int`, `long`, `long long`, `float`, `double`, and `long double`, `itk::Math::abs(x)` will still call `std::abs`, right? Is that still intended? Looking at:
- https://en.cppreference.com/w/cpp/numeric/math/abs
- https://en.cppreference.com/w/cpp/numeric/math/fabs.html

Maybe that's OK, I'm just asking  😇

Modules/Core/Common/include/itkMath.h
hjmjohnson: @blowekamp Thanks for teachingme that approach.  I had not considered using the final 'else' branch to invalidate the template.  Seems obvious, now.  

I agree, this is much cleaner.  I just hired two undergraduates to work in my lab.  They are not ready for ITK programming yet, but perhaps they could take on some refactoring tasks starting next month.

Modules/Core/Common/include/itkMath.h
hjmjohnson: I don't have a good answer.  It was code originally copied from vnl. 

 I'm trying:
 
 ```c++
#if defined(ITK_FUTURE_LEGACY_REMOVE)
// Explicitly calling the std::abs for types needed in the ITK namespace (i.e. std::complex)
template <typename T>
constexpr auto
abs(std::complex<T> x) noexcept
{
  return std::abs(x);
}
#else
// The following is needed for backward compatibility with pre ITKv6.0 versions of ITK
// for itk::Math::abs<std::complex<double>>'
// and itk::Math::abs<std::complex<float>>'
using std::abs;
#endif
```

Comment (N-Dekker) [reply Modules/Core/Common/include/itkMath.h]: I see now, the original `itk::Math::abs` (before your PR) did not implement int, long, long long, float, double, and long double. It depended on `std::abs` to do so.

Now that you explicitly implemented `itk::Math::abs` for all integer and floating point types (with this PR), I think `using std::abs` is no longer necessary.

File: Modules/Core/Common/include/itkMath.h [hunk 4]
```diff
@@ -822,58 +823,60 @@ using vnl_math::sqr;
...
-// long long - target type will have width of at least 64 bits. (since C++11)
-inline unsigned long long
-abs(unsigned long long x)
+/**
+ * @brief Returns the absolute value of a number.
+ *
+ * This function provides a c++17 implementation of the vnl_math::abs absolute value functionality.
+ * It preserves backward compatibility with vnl_math::abs that was originally used in ITK.
+ *
+ * NOTE: itk::abs preserves backward compatibility with vnl_math::abs that was
```
dzenanz: Paragraph somewhat redundant, should be merged into the previous one.

Modules/Core/Common/include/itkMath.h
```diff
@@ -822,58 +823,60 @@ using vnl_math::sqr;
...
+        return static_cast<std::make_unsigned_t<T>>(x < 0 ? static_cast<unsigned int>(-static_cast<int>(x))
+                                                          : static_cast<unsigned int>(x));
+      }
+      else
+      {
+        return static_cast<std::make_unsigned_t<T>>(x < 0 ? -x : x);
+      }
+    }
+  }
+  else if constexpr (std::is_floating_point_v<T>)
```
dzenanz: The `if constexpr` is so much more readable than SFINAE.

Modules/Core/Common/include/itkMath.h
```diff
@@ -823,60 +825,104 @@ using vnl_math::cube;
...
+ * Where std::abs returns the absolute value in the same type as the input, itk::safe_abs
+ * returns the absolute value in the unsigned equivalent of the input type.
+ *
+ * @tparam T The type of the input number.
+ * @param x The input number.
+ * @return The absolute value of the input number as an unsigned type for integer types.
+ */
+template <typename T>
+constexpr auto
+safe_abs(T x) noexcept
```
N-Dekker, line 843: @hjmjohnson I understand that an ITK specific function name is helpful, to avoid confusion with `std::abs`, thanks. Can we possibly still reconsider the name of the ITK function? Why did you choose "safe_abs"? Would a different name also be OK to you? According to the Naming Conventions of our [CodingStyleGuide](https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/blob/main/SoftwareGuide/Latex/Appendices/CodingStyleGuide.tex): [_Global functions and class methods, either static or class members, are named beginning with a capital letter_](https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/blob/71baa24051ba86a079f3829ebc9791aa63f1e151/SoftwareGuide/Latex/Appendices/CodingStyleGuide.tex#L735). "itk::Math::Round" follows this naming convention: https://github.com/InsightSoftwareConsortium/ITK/blob/3cc3f4a1e6347e0c46a6ef1e49bbcabba2005fd9/Modules/Core/Common/include/itkMath.h#L177

So then, what do you think of "itk::Math::Abs", instead of "itk::Math::safe_abs"?

(Our style guide also says, that "names are generally spelled out; use of abbreviations is discouraged". So you might also consider "itk::Math::Absolute"  😇 )

blowekamp: Also the signature does not indicate the return type, so perhaps incorporations "Unsigned" into the name or `uabs` of you want C name style would be useful?

PR #5800, opened 2026-02-13
COMP: Do not use itkWarningMacro inside a non-itk::Object

blowekamp:
itkWarning macro requires methods from the itk::Object. This code block was non-functioning. It is replaced with direct calls to the appropriate warnings methods.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/src/itkNumericLocale.cxx
+++ b/Modules/Core/Common/src/itkNumericLocale.cxx
@@ -18,6 +18,7 @@
 
 #include "itkNumericLocale.h"
 #include "itkConfigurePrivate.h"
+#include "itkObject.h"
 
 #include <locale.h>
 #include <cstring>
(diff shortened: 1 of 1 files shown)


Modules/Core/Common/src/itkNumericLocale.cxx
```diff
@@ -132,10 +133,15 @@ NumericLocale::NumericLocale()
   if (currentLocale && std::strcmp(currentLocale, "C") != 0)
   {
     // Issue warning only once per instance
-    itkWarningMacro("LC_NUMERIC locale is '" << currentLocale << "' (not 'C'). "
-                                             << "Thread-safe locale functions not available. "
-                                             << "Locale-dependent number parsing may cause issues. "
-                                             << "Please manage locale at application level.");
+    if (::itk::Object::GetGlobalWarningDisplay())
```
hjmjohnson, line 136: I think that the 

```c++
#define itkGenericOutputMacro(x)           
```
could also be used here.

blowekamp: That would call OutputWindowDisplayGenericOutputText vs OutputWindowDisplayWarningText. An `itkGenericWarningMacro(x)` is what would be needed.

hjmjohnson: OK.  I think this one-off case should stay as you have it. Thanks for looking into this.

PR #5798, opened 2026-02-13
add constexpr for math operations

hjmjohnson:
- **ENH: Simplify `itk::Math` prime-related functions with `constexpr`**
- **ENH: Convert itkMathTest to itkMathGTest

## PR Checklist
- [x] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/main/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [x] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/main/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [x] Added test (or behavior not changed)

PR #5802, opened 2026-02-13
STYLE: Use itk::Math::Absolute() instead of abs to communicate subtle implementation differences

hjmjohnson:
- **STYLE: Use itk::Math::Absolute() name directly**

Depends on :
 - [x] #5798 

## PR Checklist


Examples/RegistrationITKv4/ImageRegistration11.cxx
```diff
@@ -85,7 +85,7 @@ class CommandIterationUpdate : public itk::Command
     }
     const double currentValue = optimizer->GetValue();
     // Only print out when the Metric value changes
-    if (itk::Math::abs(m_LastMetricValue - currentValue) > 1e-7)
+    if (itk::Math::safe_abs(m_LastMetricValue - currentValue) > 1e-7)
```
N-Dekker: Maybe for a follow-up pull request: simply use `std::abs` in cases like this! Or even better: `std::fabs`, because it is compared with a floating point number (`1e-7`) anyway:

```cpp
    if (std::fabs(m_LastMetricValue - currentValue) > 1e-7)
```

Examples/RegistrationITKv4/ImageRegistration14.cxx
```diff
@@ -76,7 +76,7 @@ class CommandIterationUpdate : public itk::Command
     }
     const double currentValue = optimizer->GetValue();
     // Only print out when the Metric value changes
-    if (itk::Math::abs(m_LastMetricValue - currentValue) > 1e-7)
+    if (itk::Math::safe_abs(m_LastMetricValue - currentValue) > 1e-7)
```
N-Dekker: Please also consider my comment at https://github.com/InsightSoftwareConsortium/ITK/pull/5797/changes/3cc3f4a1e6347e0c46a6ef1e49bbcabba2005fd9#r2807350877 I think it would be nice to rename  "itk::Math::safe_abs" to "itk::Math::Abs" or "itk::Math::Absolute", following our naming conventions.

hjmjohnson: :). I was trying to convey that this version was a 'safer' version of typical abs functions because of the corner case (i.e., taking the absolute value of the minimum of signed integers).

PR #5803, opened 2026-02-13
Deprecate `ImageConstIterator::GetIndex()`, use ComputeIndex() in tests and examples

N-Dekker:
- Follow-up to pull request #5787

### DIFF
--- a/Documentation/docs/migration_guides/itk_6_migration_guide.md
+++ b/Documentation/docs/migration_guides/itk_6_migration_guide.md
@@ -172,6 +172,15 @@ reoriented_image = itk.orient_image_filter(
 )
 ```
 
+`GetIndex()` replaced with `ComputeIndex()`, for iterators without index
+------------------------------------------------------------------------
+`ImageConstIterator::GetIndex()` is marked to be removed in the future. This iterator does not have an index,
+internally, so its `GetIndex()` member function does a potentially expensive computation. For an iterator that is
+derived from `ImageConstIterator`, please use its newly introduced `ComputeIndex()` member function instead. Or use an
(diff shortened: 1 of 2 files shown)


hjmjohnson (approved):
@N-Dekker This looks good as is. Adding a note to this change in the migration guide now will be useful in the future when the GetIndex is removed in  a few years form now.

N-Dekker:
> @N-Dekker This looks good as is. Adding a note to this change in the migration guide now will be useful in the future when the GetIndex is removed in a few years form now.

Thanks, Hans. I'm adding the following note, please check:

---

`GetIndex()` replaced with `ComputeIndex()`, for iterators without index
------------------------------------------------------------------------
`ImageConstIterator::GetIndex()` is marked to be removed in the future. This iterator does not have an index,
internally, so its `GetIndex()` member function does a potentially expensive computation. For an iterator that is
derived from `ImageConstIterator`, please use its newly introduced `ComputeIndex()` member function instead. Or use an
iterator _with_ index (like `ImageIteratorWithIndex`).

blowekamp:
> @blowekamp Thanks, I think you're presumption is correct!
> 
> Is https://open.cdash.org/viewBuildError.php?buildid=11052824 built using `ITK_FUTURE_LEGACY_REMOVE=ON` ?
> 
> * ITK_FUTURE_LEGACY_REMOVE will remove ImageConstIterator::GetIndex()
> * ITK_LEGACY_REMOVE will only produce a warning ("deprecated") for each ImageConstIterator::GetIndex() call

I think that is a reasonable presumption. You'd have to dig into the configuration log and files on the dashboard to confirm.

PR #5804, opened 2026-02-14
STYLE: Remove redundant `typename` in type aliases

hjmjohnson:
Refactor code to remove redundant `typename` specifiers in type aliases
where unnecessary. Simplifies type definitions for improved readability
and consistency while adhering to modern C++ practices.

'typename; is redundant before non-dependent names as of c++17

## PR Checklist

### DIFF
--- a/Examples/Filtering/SpatialObjectToImage1.cxx
+++ b/Examples/Filtering/SpatialObjectToImage1.cxx
@@ -186,9 +186,9 @@ main(int argc, char * argv[])
   // Software Guide : BeginCodeSnippet
   ellipse->SetRadiusInObjectSpace(size[0] * 0.2 * spacing[0]);
 
-  typename TubeType::PointType         point;
-  typename TubeType::TubePointType     tubePoint;
-  typename TubeType::TubePointListType tubePointList;
+  TubeType::PointType         point;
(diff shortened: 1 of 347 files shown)


Examples/Filtering/SpatialObjectToImage1.cxx
```diff
@@ -186,9 +186,9 @@ main(int argc, char * argv[])
   // Software Guide : BeginCodeSnippet
   ellipse->SetRadiusInObjectSpace(size[0] * 0.2 * spacing[0]);
 
-  typename TubeType::PointType         point;
-  typename TubeType::TubePointType     tubePoint;
-  typename TubeType::TubePointListType tubePointList;
+  TubeType::PointType         point;
```
dzenanz, line 189: Can we enable this warning on some of the CI images? Otherwise this is likely to creep back in, and require occasional cleanup.

hjmjohnson: I don't think that is an easy task.  We would need to instrument clang-tidy in a way that was not extremely verbose.

clang-tidy is very useful and almost always accurate, but not 100% correct.

blowekamp:
Thank for keeping an eye on the cdash! Those should have been addressed here: #5827

PR #5806, opened 2026-02-15
Use RegionConstIteratorWithIndex in ContourExtractor2DImageFilter, deprecate nested RegionIterator aliases

N-Dekker:
- Replaced using `RegionConstIterator` with using `ImageRegionConstIteratorWithIndex` in `ContourExtractor2DImageFilter::GenerateDataForLabels()`,  to avoid potentially expensive index computation with each iteration.
- Removed a `GoToBegin()` call, following pull request #5736
- Deprecated (FUTURE_LEGACY_REMOVE) the type aliases `RegionIterator` and `RegionConstIterator` that were nested withing `ContourExtractor2DImageFilter`, as they are no longer being used.


thewtex (approved):
🚀

PR #5808, opened 2026-02-17
PERF: Make IndexRange iterators _much_ faster, by m_Position data member

N-Dekker:
Added a numeric data member to the iterator types of IndexRange, denoting the position of an iterator relative to the begin of its IndexRange. Used to speed up its comparison operators.

The following code became more than 4x as fast, using Visual Studio 2026 (Release build). For a region size of around 100'000 x 10'000, the duration went from ~1.9 sec (before this commit) down to ~0.4 sec (after this commit).

    for (const auto & index: ImageRegionIndexRange<2>(region))
    {
      result += index[0] - index[1];
    }

### DIFF
--- a/Modules/Core/Common/include/itkIndexRange.h
+++ b/Modules/Core/Common/include/itkIndexRange.h
@@ -120,6 +120,8 @@ public:
     constexpr const_iterator &
     operator++() noexcept
     {
+      ++m_Position;
+
       for (unsigned int i = 0; i < (VDimension - 1); ++i)
       {
         auto & indexValue = m_Index[i];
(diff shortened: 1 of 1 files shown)


Modules/Core/Common/include/itkIndexRange.h
```diff
@@ -206,20 +212,7 @@ class IndexRange final
...
-          return true;
-        }
-        if (difference > 0)
-        {
-          break;
-        }
-      }
-      return false;
+      return lhs.m_Position < rhs.m_Position;
     }
```
N-Dekker, line 216: While the main aim of this PR is to speed up `operator==`, it also very much simplifies this `operator<` 😃

PR #5809, opened 2026-02-17
BUG: `Math::Absolute(-0.0)` should return "plus zero" (`+0.0`)

N-Dekker:
Adjusted `Absolute(T)` to ensure that `Absolute(-0.0)` returns `+0.0`.

Added a unit test, `AbsoluteMinusZeroReturnsPlusZero`.

### DIFF
--- a/Modules/Core/Common/include/itkMath.h
+++ b/Modules/Core/Common/include/itkMath.h
@@ -937,7 +937,9 @@ Absolute(T x) noexcept
 #if __cplusplus >= 202302L
     return std::abs(x);
 #else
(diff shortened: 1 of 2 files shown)


hjmjohnson (commented):
@N-Dekker Thank you!  If this fails/does not fail, I think the test is good.  The fix is good, also.  I'll keep an eye on this today, hopefully, and resolve this corner case soon.

N-Dekker:
@hjmjohnson No worries, I'm preparing a fix (to be proposed). 😇

PR #5812, opened 2026-02-17
Add `itk::MakeIndexRange` functions and use them

N-Dekker:
There are two kinds of IndexRange's:
1. `ZeroBasedIndexRange` is optimized for an image region that has a start index of all zeros.
2. `ImageRegionIndexRange` supports any arbitrary image region, having any arbitrary start index.

The proposed overload set of `MakeIndexRange` functions aims to ease creating the most appropriate IndexRange type for each use case:

 - `MakeIndexRange(size)` returns a `ZeroBasedIndexRange`
 - `MakeIndexRange(imageRegion)` returns an `ImageRegionIndexRange`
 - `MakeIndexRange(index, size)` returns an `ImageRegionIndexRange`

This pull request also proposes to start using `MakeIndexRange` in various tests, filters and other components, to improve code readability.

### DIFF
--- a/Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.h
+++ b/Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.h
@@ -145,7 +145,11 @@ public:
   using InputRealType = typename NumericTraits<InputPixelType>::RealType;
 
   /** Ranges and iterators for regions */
-  using RegionIndexRange = ImageRegionIndexRange<InputImageType::ImageDimension>;
+#ifndef ITK_FUTURE_LEGACY_REMOVE
+  using RegionIndexRange ITK_FUTURE_DEPRECATED(
+    "Please use `itk::ImageRegionIndexRange` or `itk::MakeIndexRange` directly!") =
+    ImageRegionIndexRange<InputImageType::ImageDimension>;
+#endif
   using RegionRange = ImageRegionRange<InputImageType>;
   using RegionConstRange = ImageRegionRange<const InputImageType>;
 
--- a/Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.hxx
+++ b/Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.hxx
@@ -111,7 +111,7 @@ ContourExtractor2DImageFilter<TInputImage>::CreateSingleContour(InputPixelType  
   using NeighborhoodRange = itk::ShapedImageNeighborhoodRange<const InputImageType, Policy>;
   NeighborhoodRange neighborhoodRange{ *input, InputIndexType(), offsets, m_UnusedLabel };
 
-  for (const InputIndexType index : RegionIndexRange{ usableRegion })
+  for (const InputIndexType index : MakeIndexRange(usableRegion))
   {
     neighborhoodRange.SetLocation(index);
     // There are sixteen different possible square types, diagrammed below.


Modules/Registration/RegistrationMethodsv4/include/itkTimeVaryingBSplineVelocityFieldImageRegistrationMethod.hxx
```diff
@@ -508,12 +508,8 @@ TimeVaryingBSplineVelocityFieldImageRegistrationMethod<
         fixedDisplacementField->GetRequestedRegion().GetIndex();
       typename DisplacementFieldType::SizeType fixedDomainSize = fixedDisplacementField->GetRequestedRegion().GetSize();
 
-      ImageRegionConstIteratorWithOnlyIndex<DisplacementFieldType> ItF(fixedDisplacementField,
-                                                                       fixedDisplacementField->GetRequestedRegion());
-      for (ItF.GoToBegin(); !ItF.IsAtEnd(); ++ItF)
+      for (const auto & index : MakeIndexRange(fixedDisplacementField->GetRequestedRegion()))
       {
-        typename DisplacementFieldType::IndexType index = ItF.GetIndex();
-
```
N-Dekker, line 516: This is maybe my favorite part of the PR, replacing four lines of code that initialized and used the old `ImageRegionConstIteratorWithOnlyIndex` with a single line, starting a range-based `for` loop, simply by doing

```cpp
for (const auto & index : MakeIndexRange(fixedDisplacementField->GetRequestedRegion()))
```

Hope you all like it as well 😃

Modules/Filtering/Path/include/itkContourExtractor2DImageFilter.h
```diff
@@ -145,7 +145,11 @@ class ITK_TEMPLATE_EXPORT ContourExtractor2DImageFilter
   using InputRealType = typename NumericTraits<InputPixelType>::RealType;
 
   /** Ranges and iterators for regions */
-  using RegionIndexRange = ImageRegionIndexRange<InputImageType::ImageDimension>;
+#ifndef ITK_FUTURE_LEGACY_REMOVE
+  using RegionIndexRange ITK_FUTURE_DEPRECATED(
+    "Please use `itk::ImageRegionIndexRange` or `itk::MakeIndexRange` directly!") =
+    ImageRegionIndexRange<InputImageType::ImageDimension>;
+#endif
```
N-Dekker, line 152: It appears that this nested `RegionIndexRange` type alias was only used internally by the filter itself, so now it is no longer necessary!

dzenanz (approved):

Modules/Core/Common/include/itkIndexRange.h
```diff
@@ -474,6 +473,30 @@ using ImageRegionIndexRange = IndexRange<VDimension, false>;
...
+}
+
+/* Creates a range of indices for the image region specified by its index and size. */
+template <unsigned int VDimension>
+[[nodiscard]] auto
+MakeIndexRange(const Index<VDimension> & imageRegionIndex, const Size<VDimension> & imageRegionSize)
+{
+  return ImageRegionIndexRange<VDimension>(ImageRegion<VDimension>{ imageRegionIndex, imageRegionSize });
+}
+
```
N-Dekker, line 499: Instead of these simple `MakeIndexRange` functions, I _think_ we might as well add similar functionality by using CTAD (class template argument deduction). But we had discussion about CTAD before. It might sometimes be risky to use CTAD (looking at [Arthur O'Dwyer's blog](https://quuxplusone.github.io/blog/tags/#class-template-argument-deduction)). And it might sometimes cause compiler warnings (_warning: 'T' may not intend to support class template argument deduction [-Wctad-maybe-unsupported]_). So maybe we should just stick with such simple `Make` functions 🤷

N-Dekker:
[macOS-Build22156645989-MakeIndexRange](https://open.cdash.org/builds/11056860) fails, saying:

```
Error in CheckBuffer(image5.GetPointer(), std::vector<float>(10, 3.14))
  In /Users/runner/work/ITK/ITK/Modules/Core/Common/test/itkImageAlgorithmCopyTest2.cxx, line 120
Expected true
  but got  0
```

Does anyone have a clue? 🤷

PR #5814, opened 2026-02-19
Use an iterator 'WithIndex' when the index is retrieved during each iteration

N-Dekker:
When the index is retrieved during _each_ iteration, an iterator _with index_ (like `ImageRegionIteratorWithIndex`) is much faster than an iterator that _computes_ the index (like `ImageRegionIterator`). 

This pull request proposes to use an iterator _with index_, instead of an iterator that _computes_ the index, in such cases.

It aims to speed up the following filters:
- AccumulateImageFilter
- BinomialBlurImageFilter
- DifferenceOfGaussiansGradientImageFilter
- HardConnectedComponentImageFilter
- InvertDisplacementFieldImageFilter
- IterativeInverseDisplacementFieldImageFilter
- LabelGeometryImageFilter
- MirrorPadImageFilter
- SpatialFunctionImageEvaluatorFilter

As well as:
- GaussianImageSource
- ScalarChanAndVeseLevelSetFunction

This pull request also proposes to deprecate `IterativeInverseDisplacementFieldImageFilter::OutputIterator`, as it is no longer used by now.

### DIFF
--- a/Modules/Filtering/DisplacementField/include/itkIterativeInverseDisplacementFieldImageFilter.h
+++ b/Modules/Filtering/DisplacementField/include/itkIterativeInverseDisplacementFieldImageFilter.h
@@ -85,8 +85,10 @@ public:
 
   using InputConstIterator = ImageRegionConstIterator<InputImageType>;
   using InputIterator = ImageRegionIterator<InputImageType>;
-  using OutputIterator = ImageRegionIterator<OutputImageType>;
-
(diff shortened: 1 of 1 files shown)


N-Dekker:
For the record, I observed that `ImageRegionConstIterator` (without index) can take _twice_ as much time as `ImageRegionConstIteratorWithIndex`, when retrieving the pixel value _and_ index with each iteration. As I mentioned at https://github.com/SuperElastix/elastix/pull/1410#issuecomment-3905217222

PR #5819, opened 2026-02-21
STYLE: Simplify `bool` assignments by removing `if` and `else`

N-Dekker:
Replaced code like `if (x) b = true; else b = false;` with `b = x;`

    Find what: ^([ ]+ )if \((.+)\)\r\n\1{\r\n\1  (.+) = true;\r\n\1}\r\n\1else\r\n\1{\r\n\1  \3 = false;\r\n\1}
    Replace with: $1$3 = $2;
    Filters: itk*.*
    [v] Match case
    (*) Regular expression

Improved code readability by ensuring that the expression at the right hand side of each assignment is a Boolean expression, rather than an integer expression.

### DIFF
--- a/Modules/IO/MeshOBJ/src/itkOBJMeshIO.cxx
+++ b/Modules/IO/MeshOBJ/src/itkOBJMeshIO.cxx
@@ -167,25 +167,11 @@ OBJMeshIO::ReadMeshInformation()
 
   this->m_PointDimension = 3;
 
-  // If number of points is not equal zero, update points
-  if (this->m_NumberOfPoints)
-  {
-    this->m_UpdatePoints = true;
-  }
-  else
-  {
-    this->m_UpdatePoints = false;
-  }
+  // If number of points is greater than zero, update points
+  this->m_UpdatePoints = this->m_NumberOfPoints > 0;
 
   // If number of cells is not equal zero, update points
-  if (this->m_NumberOfCells)
-  {
-    this->m_UpdateCells = true;
-  }
-  else
-  {
-    this->m_UpdateCells = false;
-  }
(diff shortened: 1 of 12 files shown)


Modules/IO/MeshOBJ/src/itkOBJMeshIO.cxx
```diff
@@ -167,25 +167,11 @@ OBJMeshIO::ReadMeshInformation()
...
-  if (this->m_NumberOfPoints)
-  {
-    this->m_UpdatePoints = true;
-  }
-  else
-  {
-    this->m_UpdatePoints = false;
-  }
+  // If number of points is greater than zero, update points
+  this->m_UpdatePoints = this->m_NumberOfPoints > 0;
```
N-Dekker, line 171: > 8-fold reduction in number of lines. But it takes a little bit more time to comprehend what happens. I wonder whether that tradeoff is worth it.

@dzenanz I think it's worth it  😇  With my latest [force-push](https://github.com/InsightSoftwareConsortium/ITK/compare/67cae79271b3783ae9903f9f6a9ab507dbab681e..2a1173843323fb02d78ac0d90d0af7ecd5d6a96e) I improved the readability by ensuring that the right hand side of the assignment is a Boolean expression, so that it no longer relies on implicit integer-to-bool conversion. Does this make the PR more acceptable to you?

I think it's quite clear to any C++ programmer that `this->m_NumberOfPoints > 0` is a `bool` expression, right?

hjmjohnson: My opinion is that the new code is easier to read.

PR #5826, opened 2026-02-24
STYLE: Remove GoToBegin() calls from Iterator constructors

N-Dekker:
The `Iterator(image, region)` constructors of ImageConstIteratorWithIndex and ImageConstIteratorWithOnlyIndex already initialize themselves at the begin, so it is not necessary for them to call GoToBegin().

Properly initialized `m_Remaining` in their member initializer lists.

### DIFF
--- a/Modules/Core/Common/include/itkImageConstIteratorWithIndex.hxx
+++ b/Modules/Core/Common/include/itkImageConstIteratorWithIndex.hxx
@@ -51,6 +51,7 @@ ImageConstIteratorWithIndex<TImage>::ImageConstIteratorWithIndex(const TImage * 
   : m_Image(ptr)
   , m_BeginIndex(region.GetIndex())
   , m_Region(region)
+  , m_Remaining(region.GetNumberOfPixels() > 0)
 {
   const InternalPixelType * buffer = m_Image->GetBufferPointer();
   m_PositionIndex = m_BeginIndex;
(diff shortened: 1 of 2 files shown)


dzenanz (approved):

Modules/Core/Common/include/itkImageConstIteratorWithIndex.hxx
```diff
@@ -51,6 +51,7 @@ ImageConstIteratorWithIndex<TImage>::ImageConstIteratorWithIndex(const TImage *
   : m_Image(ptr)
   , m_BeginIndex(region.GetIndex())
   , m_Region(region)
+  , m_Remaining(region.GetNumberOfPixels() > 0)
```
blowekamp, line 54: Back in the 32-bit ( or long = 32-bit) days we may have intentionally avoided computing the total number pixels in certain cases similar to this, to avoid potential overflow during multiplication. I suppose this is a non issue now.

N-Dekker: @blowekamp I think I got what you mean, but well... the use of `GetNumberOfPixels() > 0` to check if the region has any pixels was introduced in ImageConstIteratorWithIndex already back in 2003, with commit 76072d177439710d8e3b93b2d22b849320bde4d0  Those were the dying days of 32-bit, I guess 😺

PR #5831, opened 2026-02-24
Refactor third-party Eigen3 CMake code

blowekamp:
Make Eigen3 a follow standard ITK 3rd party library conventions.

Removes Eigen3 as an external project. Converted to use ITK module macros to install and export the eigen_internal target. Removes the CMake code to install auxiliary cmake and configuration file for target. The target is now imported through ITKTargets, and not a separate export and configuration file.

Preserve Eigen3::Eigen project exports.

Updated ITK/eigen repo then used sub-tree merge script to update.

## PR Checklist

### DIFF
--- a/Modules/ThirdParty/Eigen3/src/itkeigen/CMakeLists.txt
+++ b/Modules/ThirdParty/Eigen3/src/itkeigen/CMakeLists.txt
@@ -882,7 +882,7 @@ install ( FILES ${CMAKE_CURRENT_BINARY_DIR}/Eigen3Config.cmake
 # Install files (used for both eigen_external and eigen_internal)
 install(
   DIRECTORY   "${CMAKE_CURRENT_SOURCE_DIR}/Eigen/"
-  DESTINATION "${INCLUDE_INSTALL_DIR}/itkeigen/Eigen"
+  DESTINATION "${ITK3P_INSTALL_INCLUDE_DIR}/itkeigen/Eigen"
   PATTERN "*.txt" EXCLUDE)
 
 ######################### eigen_internal #####################################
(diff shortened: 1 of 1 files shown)


blowekamp:
@seanm  @hjmjohnson @phcerdan I know Eigen3 is used in ITKTotalVariations and itkSymmetricEigenAnalysis.h. I thought there was more interest in using Eigen in the ITK community. Are there other existing use cases?

phcerdan:
If you are touching third-party code (eigen code), better to make those changes in our third-party clone (https://github.com/InsightSoftwareConsortium/eigen) rather than ITK proper. This approach is really forced in VTK (ghostflow-check-main checks that). In ITK we are more flexible, but still recommended.

You are destroying with this all the efforts to make ITK modules to be able to reuse Eigen as-is, but honestly, I think is the right call. Nobody is doing it, and the extra complexity doesn't pay off.

Given that we are not allowing this anymore, I would strongly suggest to mangle symbols (and header guards) for our internal eigen to avoid conflicts with external ones (other libraries that use Eigen externally). VTK does this in a clean way.

phcerdan:
And yes, the idea was to take advantage of Eigen to replace VNL. I honestly think that codex/claude can nail this nowadays. It's more verbose than complex. A bit tricky, but our regressions tests should really help with this.

Modules/ThirdParty/Eigen3/src/itkeigen/CMakeLists.txt
```diff
@@ -882,7 +882,7 @@ install ( FILES ${CMAKE_CURRENT_BINARY_DIR}/Eigen3Config.cmake
 # Install files (used for both eigen_external and eigen_internal)
```
phcerdan, line 882: third-party modification should be done in our eigen clone: https://github.com/InsightSoftwareConsortium/eigen

blowekamp: We can review here. We can use ITK's infrastructure to build and test the changes before pushing upstream to the fork.

blowekamp:
Wrapping TotalVariation works with InsightSoftwareConsortium/ITKTotalVariation#57. 

I enabled #5925 in an ITK build and that work too.

I pushed the CMake changes for eigen to https://github.com/InsightSoftwareConsortium/eigen/tree/itk_module_install and did ran the update script.

This PR should be about ready to go!

Modules/ThirdParty/Eigen3/src/itkeigen/CMakeLists.txt
blowekamp: done

PR #5835, opened 2026-02-25
STYLE: remove redundant control fllow

hjmjohnson:
The clang-tidy check readability-redundant-control-flow looks for procedures
(functions returning no value) with return statements at the end of the
function. Such return statements are redundant.

Loop statements (for, while, do while) are checked for redundant continue
statements at the end of the loop body.

## PR Checklist

### DIFF
--- a/Modules/Core/Common/src/itkObject.cxx
+++ b/Modules/Core/Common/src/itkObject.cxx
@@ -209,8 +209,6 @@ Object::SubjectImplementation::InvokeEventRecursion(const EventObject &         
 
     ++i;
   }
-
-  return;
 }
 
 Command *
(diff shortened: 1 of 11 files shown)


N-Dekker (approved):
Thanks Hans!

Just a nitpick: the subject of the commit message, "STYLE: remove redundant control fllow" has a redundant letter `l` in "fllow". The commit text suggests that `continue` might also be removed, but that's not the case. The commit only removes `return` statements. So I would prefer a commit subject line like:

    STYLE: Remove redundant `return;` statements

Just a suggestion, not a show-stopper.

thewtex (approved):

PR #5838, opened 2026-02-26
COMP: Promote ITK header paths for name mangling of export.h

hjmjohnson:
When linking against ITK's expat library, the export.h must be found from the ITK-supplied expat.h so that name mangling is properly enforced.

The source layout changes in 8ec19eb1d6778a879b62a2d7006d9b12dbfd7abb allowed/usr/include/expat.h (i.e., without name mangling) to be found first, resulting in link errors due to name mangling.

Restored previous behavior for header path ordering.

This resolves an issue with building SlicerExecutionModel against the ITK v6.0b02+ versions.

## PR Checklist


hjmjohnson:
@blowekamp PS:  I'm inclined to agree with you regarding the use of itk_expat.h, but I am worried about backward compatibility as well.  Perhaps the addition of this path should be in an "ITK_FUTURE_LEGACY_REMOVE" block.

Modules/ThirdParty/Expat/CMakeLists.txt
```diff
@@ -21,6 +21,18 @@ else()
...
+  if(NOT ITK_FUTURE_LEGACY_REMOVE)
+    list(
+      APPEND
+      ITKExpat_INCLUDE_DIRS
+      # The include is needed to promote finding itk version of
+      # "#include <expat.h>" so that the linkage stage can find the
+      # correct symbol names for backward compatibiltiy with
+      # behaviors of ITKv5 and earlier.
+      ${ITKExpat_SOURCE_DIR}/src/itkexpat/lib
+    )
```
blowekamp: With the ITK_FUTURE_LEGACY_REMOVE this include will be here until ITK 9.0 and no warning will be generated to know that they were using expat in a potentially problematic way.

An alternative would be to generate a "expat.h" file in  `${ITKExpat_BINARY_DIR}/src/itkexpat` that includes the correct `itk_expat.h" and produces a compiler warning to inform of the issue, the generation of the file can be controlled, and the warning by LEGACY SILENT.

hjmjohnson: @blowekamp That is a good suggestion.  I'll try to work it up over the next few days.

Modules/ThirdParty/Expat/CMakeLists.txt
```diff
@@ -43,6 +43,18 @@ else()
...
+if(NOT ITK_USE_SYSTEM_EXPAT AND NOT ITK_FUTURE_LEGACY_REMOVE)
+  # Configure a file with the name <expat.h> as was the default in ITKv5 and earlier
+  # so that "#include <expat.h>" finds the itk mangled symbols during linkage stage.
+  # This is only needed to preserve behaviors of ITKv5 and earlier. ITKv6 and greater
+  # should #include "itk_expat.h" to get the correctly mangled symbols.
+  configure_file(itkv5_expat.h.in ${ITKExpat_BINARY_DIR}/expat.h)
+else()
+  # Remove file if configuration changes occur
+  if(EXISTS ${ITKExpat_BINARY_DIR}/expat.h)
+    file(REMOVE ${ITKExpat_BINARY_DIR}/expat.h)
```
blowekamp: Per cmake doc's no error is omitted if the file does not exist, so the conditional is not needed.

seanm:
@hjmjohnson I just git bisected a build failure to bfc8e07d360d40e07c1636fb09832a896de1d44f:

```
/Users/sean/external/ITK/Modules/ThirdParty/GDCM/src/gdcm/Utilities/gdcm_expat.h:20:11: fatal error: 'itk_expat.h' file not found
   20 | # include "itk_expat.h"
      |           ^~~~~~~~~~~~~
1 error generated.
```

Not yet sure what build param is different with this build vs my cdash submissions...

PR #5844, opened 2026-02-27
COMP: Fix IntensityImageIteratorType is not a class or namespace name

dzenanz:
The error message was: `T:\Dashboard\ITK\Modules\Nonunit\Review\include\itkLabelGeometryImageFilter.hxx(266,14): error C2653: 'IntensityImageIteratorType': is not a class or namespace name`

## PR Checklist


Modules/Nonunit/Review/include/itkLabelGeometryImageFilter.hxx
```diff
@@ -273,7 +271,7 @@ LabelGeometryImageFilter<TLabelImage, TIntensityImage>::GenerateData()
       auto mapIt = m_LabelGeometryMapper.find(label);
 
       value = static_cast<RealType>(it.Get());
-      index = it.GetIndex();
+      auto index = it.GetIndex();
```
N-Dekker: Preferably `const auto &` (to avoid warnings like "`auto` doesn't deduce references" and "Using the `auto` keyword without an `&` causes the copy of an object"):

    const auto & index = it.GetIndex();

PR #5847, opened 2026-03-01
STYLE: Remove ImageRegionConstIteratorWithIndex from OpenCVVideoCapture

N-Dekker:
The declared local iterator variable `itkIter` appears unused.

----
 This local variable was introduced with commit 48aba9a7a771ce71ff63d9522cf0bde0c4ad1a02 (9 May 2011), but even then, it was unused: https://github.com/InsightSoftwareConsortium/ITK/blob/48aba9a7a771ce71ff63d9522cf0bde0c4ad1a02/Modules/VideoIO/OpenCV/include/itkOpenCVVideoCapture.txx#L152-L153
@gabehart Do you possibly still have a clue why the variable was introduced in the first place? I hope you agree with its removal, as hereby proposed!

### DIFF
--- a/Modules/Video/BridgeOpenCV/include/itkOpenCVVideoCapture.hxx
+++ b/Modules/Video/BridgeOpenCV/include/itkOpenCVVideoCapture.hxx
@@ -19,7 +19,6 @@
 #define itkOpenCVVideoCapture_hxx
 
 #include "itkNumericTraits.h"
-#include "itkImageRegionConstIteratorWithIndex.h"
 
 #include "opencv2/core/version.hpp"
 #if !defined(CV_VERSION_EPOCH) // OpenCV >= 3.0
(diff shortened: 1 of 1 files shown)


Modules/Video/BridgeOpenCV/include/itkOpenCVVideoCapture.hxx
```diff
@@ -164,7 +163,6 @@ OpenCVVideoCapture<TVideoStream>::retrieve(cv::Mat & image, int itkNotUsed(chann
   int                          matrixType = CV_MAKETYPE(depth, channels);
 
   // Copy the pixels -- There is probably a faster way to do this
-  ImageRegionConstIteratorWithIndex<FrameType> itkIter(frame, frame->GetLargestPossibleRegion());
```
N-Dekker, line 167: Should we then also remove this line of comment? It said:

    // Copy the pixels -- There is probably a faster way to do this

But obviously, the declaration of an iterator does not "copy the pixels".

PR #5842, opened 2026-02-26
Refactor wrapping to use target properties and address bug building remote modules against an ITK build directory

blowekamp:
Use generator expression to convert target property into properly formatted string for include file. Change to file with GENERATE option to support generator expressions.

There were redundant calls to configure the inc file for each wrapping file. An additional prefix is added to the inc file to create multiple.

This addresses a bug where remote modules could not be wrapped against a build ITK due to missing include paths and definitions.

COMPATIBILITY NOTE: Removes support for directory level properties for includes and defines. 

## PR Checklist
- [ ] No [API changes](https://github.com/InsightSoftwareConsortium/ITK/blob/main/CONTRIBUTING.md#breaking-changes) were made (or the changes have been approved)
- [ ] No [major design changes](https://github.com/InsightSoftwareConsortium/ITK/blob/main/CONTRIBUTING.md#design-changes) were made (or the changes have been approved)
- [ ] Added test (or behavior not changed)
- [ ] Updated API documentation (or API not changed)
- [ ] Added [license](https://github.com/InsightSoftwareConsortium/ITK/blob/main/Utilities/KWStyle/ITKHeader.h) to new files (if any)
- [ ] Added Python wrapping to new files (if any) as described in [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) Section 9.5
- [ ] Added [ITK examples](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples) for all new major features (if any)

Refer to the [ITK Software Guide](https://itk.org/ItkSoftwareGuide.pdf) for
further development details if necessary.

PR #5854, opened 2026-03-03
Windows Python CMake configuration fixes

thewtex:
A few fixes for the CMake Python configuration to address the Windows ITKPythonPackage build.

### DIFF
--- a/Wrapping/macro_files/itk_end_wrap_module.cmake
+++ b/Wrapping/macro_files/itk_end_wrap_module.cmake
@@ -527,12 +527,13 @@ ${DO_NOT_WAIT_FOR_THREADS_CALLS}
       )
     else()
       if(MSVC)
-        # Disables 'conversion from 'type1' to 'type2', possible loss of data warnings
+        # /wd4244: Disables 'conversion from 'type1' to 'type2', possible loss of data warnings
+        # /wd4996: Disables deprecated declaration warnings in generated wrapper code
         set_target_properties(
           ${lib}
           PROPERTIES
             COMPILE_FLAGS
-              "/wd4244"
+              "/wd4244 /wd4996"
         )
       endif()
       target_link_directories(${lib} PUBLIC ${Python3_LIBRARY_DIRS})


Modules/Core/Common/src/CMakeLists.txt
```diff
@@ -159,12 +159,17 @@ set(
...
+  else()
+    set(_itk_python_target Python3::Module)
+  endif()
   list(APPEND ITKCommon_SRCS itkPyCommand.cxx)
   set_source_files_properties(
     itkPyCommand.cxx
     PROPERTIES
       INCLUDE_DIRECTORIES
-        "$<TARGET_PROPERTY:Python3::Module,INTERFACE_INCLUDE_DIRECTORIES>"
+        "$<TARGET_PROPERTY:${_itk_python_target},INTERFACE_INCLUDE_DIRECTORIES>"
```
blowekamp, line 172: Setting this property is actually no longer needed, as the include path for python from the Python::Module interface library is applies to all of Common.

Modules/Core/Common/src/CMakeLists.txt
```diff
@@ -255,7 +260,7 @@ if(UNIX)
   endif()
 endif()
 if(ITK_WRAP_PYTHON)
-  target_link_libraries(ITKCommon PRIVATE Python3::Module)
+  target_link_libraries(ITKCommon PRIVATE ${_itk_python_target})
```
blowekamp, line 263: Using the correct interface seems right here.

 For PR #5842, we are having issues related to this. Since itkPy*.h and hxx file include Python.h in there public interface, then cast_xml need the include path. So technically this should be public linkage, but this would cause every module to link again Python which is not needed. I have been debating an about a few solution, which include linking to "Python::Module" in a couple different places.

Wrapping/macro_files/itk_end_wrap_module.cmake
```diff
@@ -547,6 +547,7 @@ ${DO_NOT_WAIT_FOR_THREADS_CALLS}
               "/wd4244"
         )
       endif()
+      target_link_directories(${lib} PUBLIC ${Python3_LIBRARY_DIRS})
```
blowekamp, line 539: If the Python::Module/SABIModule interface library is used for linking it should automatically include a full path or a lib directory. Maybe the underlying cause of this need is a raw relative path of PYTHON_LIB being used some place?

Modules/Core/Common/src/CMakeLists.txt
thewtex: 👍 I'll try removing it.

Modules/Core/Common/src/CMakeLists.txt
thewtex: Maybe we just need to manually plug in the dependencies into the castxml call?

Wrapping/macro_files/itk_end_wrap_module.cmake
thewtex: Unfortunately, I do not expect the CMake Python modules to set the full path consistently.

Modules/Core/Common/src/CMakeLists.txt
blowekamp: The latest update add this to wrapping/CMakeLists.txt: ```list(APPEND WRAPPER_LIBRARY_LINK_LIBRARIES Python3::Module)``` which I think is pretty close to that.

Modules/Core/Common/src/CMakeLists.txt
thewtex: Was needed.

PR #5855, opened 2026-03-03
PERF: Replace ComputeIndex with IndexRange in MathematicalMorphology

N-Dekker:
Replaced potentially expensive iterative `dumbImg->ComputeIndex(it)` calls with efficient `IndexRange` iteration.

Removed the dummy image (`dumbImg`).

----

Follow-up to commit a8fa92c562fcfdd10a22b4c253b433a3d9c426b4 "BUG: Image iterator was used to iterate over the indexes outside the image. Only the indexes are accessed here, not the pixels, so the image iterator was not the right thing to use.", Gaëtan Lehmann, May 8, 2009. @glehmann I hope it's OK to you!

### DIFF
--- a/Modules/Filtering/MathematicalMorphology/include/itkAnchorOpenCloseImageFilter.hxx
+++ b/Modules/Filtering/MathematicalMorphology/include/itkAnchorOpenCloseImageFilter.hxx
@@ -19,9 +19,10 @@
 #define itkAnchorOpenCloseImageFilter_hxx
 
 #include "itkNeighborhoodAlgorithm.h"
-#include "itkImageRegionConstIteratorWithIndex.h"
 #include "itkAnchorUtilities.h"
 #include "itkImageRegionIterator.h"
+#include "itkIndexRange.h"
+
 namespace itk
 {
 template <typename TImage, typename TKernel, typename TCompare1, typename TCompare2>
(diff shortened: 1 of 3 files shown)


glehmann:
I'm always 100% OK with someone fixing the mistakes I made :)
This one feels like a life ago!

Modules/Filtering/MathematicalMorphology/include/itkAnchorOpenCloseImageFilter.hxx
```diff
@@ -175,38 +176,23 @@ AnchorOpenCloseImageFilter<TImage, TKernel, TCompare1, TCompare2>::DoFaceOpen(
...
-  // the image.
-  // I can't find any cleaner way, so we use a dumb image, not even allocated,
-  // to iterate
-  // over all the indexes inside the region.
-  //
-  // using ItType = ImageRegionConstIteratorWithIndex<TImage>;
-  // ItType it(input, face);
-
-  auto dumbImg = TImage::New();
-  dumbImg->SetRegions(face);
```
N-Dekker, line 190: > I'm always 100% OK with someone fixing the mistakes I made :)

Thanks @glehmann ! I think this wasn't really a mistake anyway, because when you did your commit a8fa92c562fcfdd10a22b4c253b433a3d9c426b4 (2009), ITK did not yet have a convenient way to iterate over the indexes inside a region _without_ an image. I proposed `itk::IndexRange` specifically for that use case, but only in 2019; it is only included with ITK from version 5.0.

thewtex (approved):
🚀 🔥

PR #5874, opened 2026-03-06
ENH: Convert 12 ITKCommon CTests to GoogleTest framework

hjmjohnson:
## Summary

Converts no-argument CTests in `Modules/Core/Common/test/` to GoogleTest,
each in a separate commit. Adds the `ITK_GTEST_SET_GET_VALUE` macro to
`itkGTest.h` for use in subsequent conversions.

### Converted tests

| Old CTest | New GTest file | Key assertions added |
|-----------|---------------|----------------------|
| `itkVersionTest` | `itkVersionGTest` | `EXPECT_EQ` for major/minor/patch versions |
| `itkDataTypeTest` | `itkDataTypeGTest` | `EXPECT_EQ` per vector element |
| `itkEventObjectTest` | `itkEventObjectGTest` | `EXPECT_TRUE/FALSE` for derivation/matching, `EXPECT_STREQ` for name |
| `itkModifiedTimeTest` | `itkModifiedTimeGTest` | `EXPECT_GT/GE` for time ordering |
| `itkByteSwapTest` | `itkByteSwapGTest` | `EXPECT_EQ` per type after double-swap round-trip |
| `itkIntTypesTest` | `itkIntTypesGTest` | `EXPECT_TRUE` for size/signedness per type group |
| `itkMathRoundTest` + `itkMathRoundTest2` + `itkMathRoundProfileTest1` | `itkMathRoundGTest` | `EXPECT_TRUE` on templated round/floor/ceil correctness |
| `itkMathCastWithRangeCheckTest` | `itkMathCastWithRangeCheckGTest` | `EXPECT_THROW` for overflow, `EXPECT_TRUE` per target type |
| `itkTimeStampTest` | `itkTimeStampGTest` | `EXPECT_LE`/`EXPECT_EQ` for thread-safe monotonic timestamps |
| `itkBoundingBoxTest` | `itkBoundingBoxGTest` | `EXPECT_EQ`/`EXPECT_NEAR` for bounds, center, diagonal, corners |
| `itkBoundaryConditionTest` | `itkBoundaryConditionGTest` | `EXPECT_EQ` for all 15 boundary neighborhood pixel values |

### Approach

- All diagnostic `std::cout` output preserved
- Expected values determined by running original tests against the build
- Tests split into logical sub-cases within each file
- Old `Test.cxx` files removed; entries removed from `ITKCommon1Tests` list and `itk_add_test()` calls
- New `GTest.cxx` files added to `ITKCommonGTests` list

## Test plan

- [x] Each GTest file builds without errors
- [x] Each GTest runs and passes (`--gtest_filter=<TestSuite>*`)
- [x] CMakeLists.txt updated correctly (removed from driver list, added to GTest list)
- [x] Full CI green

### DIFF
--- a/Modules/Core/Common/test/CMakeLists.txt
+++ b/Modules/Core/Common/test/CMakeLists.txt
@@ -30,7 +30,6 @@ set(
   itkImageSliceIteratorTest.cxx
   itkRGBPixelTest.cxx
   itkLightObjectTest.cxx
-  itkBoundaryConditionTest.cxx
   itkSparseImageTest.cxx
   itkSimpleFilterWatcherTest.cxx
   itkSymmetricEllipsoidInteriorExteriorSpatialFunctionTest.cxx
@@ -443,12 +442,6 @@ itk_add_test(
   COMMAND
     ITKCommon2TestDriver
     itkAutoPointerTest
-)
-itk_add_test(
-  NAME itkBoundaryConditionTest
-  COMMAND
-    ITKCommon1TestDriver
-    itkBoundaryConditionTest
 )
 itk_add_test(
   NAME itkBSplineInterpolationWeightFunctionTest
@@ -1800,6 +1793,7 @@ set(
   itkCommonTypeTraitsGTest.cxx
   itkMathGTest.cxx
   itkDataTypeGTest.cxx
+  itkBoundaryConditionGTest.cxx
   itkBoundingBoxGTest.cxx
   itkTimeStampGTest.cxx
   itkEventObjectGTest.cxx
--- a/Modules/Core/Common/test/itkBoundaryConditionGTest.cxx
+++ b/Modules/Core/Common/test/itkBoundaryConditionGTest.cxx
@@ -19,6 +19,8 @@
 #include "itkImageRegionIterator.h"
 #include "itkConstantBoundaryCondition.h"
 #include "itkNeighborhoodIterator.h"
+#include "itkGTest.h"
+
 namespace
 {
 void
@@ -26,7 +28,6 @@ println(const char * c)
 {
   std::cout << std::endl << c << std::endl;
 }
-} // namespace
 
 template <typename TPixel>
 void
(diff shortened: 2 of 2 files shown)


Modules/Core/Common/test/itkVersionGTest.cxx
```diff
@@ -17,34 +17,39 @@
...
-int
-itkVersionTest(int, char *[])
+TEST(Version, BasicObjectMethods)
 {
-  constexpr int testPassStatus{ EXIT_SUCCESS };
-
   const itk::Version::Pointer version = itk::Version::New();
+  EXPECT_STREQ(version->GetNameOfClass(), "Version");
+  EXPECT_NE(version.GetPointer(), nullptr);
+}
```
dzenanz, line 29: `ITK_EXERCISE_BASIC_OBJECT_METHODS` is missing from this code block.

dzenanz (commented):
I will re-visit this PR once it is marked "ready for review".

Modules/Core/Common/test/itkVersionGTest.cxx
blowekamp: The macro should likely be converted into a GTest Predicate and placed into itkGTestPredicate.h

N-Dekker: @blowekamp Here it is, already  😺 
https://github.com/InsightSoftwareConsortium/ITK/blob/9b60aa251278e93d7330bdcab9ae42b4209b0105/Modules/Core/TestKernel/include/itkGTest.h#L36

blowekamp: Apparently it's not been used enough and was missing from the context Claude was using ( and my human one too).

Modules/Core/TestKernel/include/itkGTest.h
```diff
@@ -42,6 +42,13 @@
     EXPECT_STREQ(object->Superclass::GetNameOfClass(), #SuperclassName);           \
   }()
 
+#define ITK_GTEST_SET_GET_VALUE(variable, command)                                                                     \
+  ITK_GCC_PRAGMA_PUSH                                                                                                  \
+  ITK_GCC_SUPPRESS_Wfloat_equal                                                                                        \
+  EXPECT_TRUE(variable != command) << "Error in " << #command << "  In " __FILE__ ", line " << __LINE__ << "Expected " \
+                                   << variable << "but got  " << command;                                              \
+  ITK_GCC_PRAGMA_POP                                                                                                   \
+  ITK_MACROEND_NOOP_STATEMENT
```
N-Dekker: Sorry but I don't think this macro (`ITK_GTEST_SET_GET_VALUE`) is really necessary for GoogleTest based unit test. I think it's preferable to just use an equivalent standard GoogleTest macro. For example:

Old style:

    ITK_TEST_SET_GET_VALUE(innerRadius, annulus.GetInnerRadius());

Equivalent GoogleTest macro call:
 
    EXPECT_EQ(annulus.GetInnerRadius(), innerRadius);

(I prefer the "expected value" at the right side, for human readability.)

Modules/Core/Common/test/itkVersionGTest.cxx
N-Dekker: @blowekamp Please use it! And tell Claude about it!  😸

Modules/Core/TestKernel/include/itkGTest.h
hjmjohnson: @N-Dekker I was trying to minimize the code changes so that it was easier to review with the least amount of changes at this point.  

My initial testing with claude-code made massive changes that were not trackable or reviewable due to too many simultaneous "improvements".   I actually agreed with many of the claude-code improvments, and they did modern best practices, but the new files did not represent the old files and were impossible to compare without a lot of work.  

The prompt I created explicitly is trying to do the MINIMUM amount of work to get google tests while maximizing the similarity to the old code.

Modules/Core/Common/test/itkIntTypesGTest.cxx
```diff
@@ -98,61 +99,56 @@ CheckType(size_t size, bool exactSize, bool issigned, const char * name, T * = n
...
-    return EXIT_SUCCESS;
-  }
+TEST(IntTypes, MaxAndPtrTypes)
+{
+  EXPECT_TRUE(CHECKTYPE(itk::intmax_t, 4, false, true));
+  EXPECT_TRUE(CHECKTYPE(itk::uintmax_t, 4, false, false));
 
-  return EXIT_FAILURE;
+  EXPECT_TRUE(CHECKTYPE(itk::intptr_t, sizeof(void *), false, true));
+  EXPECT_TRUE(CHECKTYPE(itk::uintptr_t, sizeof(void *), false, false));
```
dzenanz, line 153: I reviewed the first few commits, and they look fine.

PR #5897, opened 2026-03-08
ENH: Convert itkFixedArrayTest to GTest

hjmjohnson:


Modules/Core/Common/test/itkFixedArrayGTest.cxx
```diff
@@ -377,3 +377,62 @@ TEST(FixedArray, CheckFrontAndBack)
   itk::RangeGTestUtilities::CheckFrontAndBack(itk::FixedArray<int>({ 0, 1, 2 }));
   itk::RangeGTestUtilities::CheckFrontAndBack(itk::FixedArray<double>::Filled(std::numeric_limits<double>::max()));
 }
+
```
dzenanz, line 380: This conversion substantially cuts down content, primarily regarding ostream operators.

hjmjohnson: @dzenanz I am trying to fine-tune, I have requests to remove duplication and minimize redundant code in the GTest verions, and keep it exactly the same.

Changes to the claude instructions that resulted in removing some of these ostream operators:

- Preserve all `std::cout` diagnostic output from the original 
+ Preserve all `std::cout` diagnostic output from the original **only** if they contain functions that would otherwise not be called.  Otherwise, attempt to minimize the redundant console outputs

NOTE:  On failure the gtests would print the same diagnostics to the output, it's part of the GTest macros.

I was trying to have two benefits here:
   1) Reduce the noise from passing tests in the enourmous testing outputs

Hans

dzenanz: If this code removal was intentional, then OK. Otherwise please amend the PR. I leave it up to you to merge.

PR #5906, opened 2026-03-08
ENH: Convert itkImportImageTest to GTest

hjmjohnson:

### DIFF
--- a/Modules/Core/Common/test/itkImportImageGTest.cxx
+++ b/Modules/Core/Common/test/itkImportImageGTest.cxx
@@ -25,7 +25,7 @@
 #include <iostream>
 
 
-TEST(ImportImageFilter, ImportAndShrink)
+TEST(ImportImageFilter, ConvertedLegacyTest)
 {
   // Create a C-array to hold an image
   auto * rawImage = new short[8 * 12];
@@ -83,6 +83,10 @@ TEST(ImportImageFilter, ImportAndShrink)
   shrink->SetShrinkFactors(2);
   EXPECT_NO_THROW(shrink->Update());
 
+  //
+  // The rest of this code determines whether the shrink code produced
+  // the image we expected.
+  //
   const ShortImage::RegionType         requestedRegion = shrink->GetOutput()->GetRequestedRegion();
   itk::ImageRegionIterator<ShortImage> iterator2(shrink->GetOutput(), requestedRegion);
 


Modules/Core/Common/test/itkImportImageTest.cxx
```diff
@@ -16,43 +16,41 @@
...
   auto basicImport = ImportImageFilter::New();
-
-  ITK_EXERCISE_BASIC_OBJECT_METHODS(basicImport, ImportImageFilter, ImageSource);
+  ITK_GTEST_EXERCISE_BASIC_OBJECT_METHODS(basicImport, ImportImageFilter, ImageSource);
 
   ShortImage::Pointer                             image;
   constexpr itk::ImageRegion<Dimension>::SizeType size = { { 8, 12 } };
   itk::ImageRegion<Dimension>                     region = { size };
-  // local scope to make sure that imported data is not deleted with ImportImageFilter
-  // but with the ImportImageContainer is creates.
```
N-Dekker, line 50: These lines of comment still seem to make sense: explaining why a local scope was opened. I would prefer if the PR would _only just_ convert the test to GTest, without doing other adjustments. Is it possible to tell Claude to do just one thing at a time? ("One Commit One Change" philosophy.) "One Commit One Change" is especially relevant for bulk commits and bulk PRs.

hjmjohnson: I added this rule to the covert-to-gtest skill:

```
## Core Philosophy: Mechanical Conversion Only

> "One Commit One Change" — N-Dekker, ITK reviewer

This conversion must be **strictly mechanical**. The GTest file should behave
identically to the old test file. Do **not**:
- Remove existing comments (especially scope-explaining comments)
- Add `[[maybe_unused]]` unless a variable is genuinely never read
- Refactor, simplify, or restructure logic beyond what conversion requires
- Add diagnostic output that was not in the original

The reviewer standard: "Please just do Test-to-GoogleTest conversion."
```

hjmjohnson: [convert-to-gtest.md](https://github.com/user-attachments/files/25888263/convert-to-gtest.md)

N-Dekker: Thanks Hans! Hope it will help Claude to stick to its task! 👍

Modules/Core/Common/test/itkImportImageGTest.cxx
```diff
@@ -16,43 +16,42 @@
...
-int
-itkImportImageTest(int, char *[])
+TEST(ImportImageFilter, ImportAndShrink)
 {
   // Create a C-array to hold an image
   auto * rawImage = new short[8 * 12];
   for (unsigned int i = 0; i < 8 * 12; ++i)
   {
-    rawImage[i] = i;
+    rawImage[i] = static_cast<short>(i);
```
N-Dekker: Did Claude also introduce this `static_cast<short>`? I see that it converts a (typically) 32-bit unsigned to a  (typically) 16-bit signed, but I think this could have been detected by traditional code analysis tools (clang tidy) or compiler warnings. Do we want to fix all those conversions?

By the way, I believe it would be better not to use `static_cast` here. In general, I believe unnecessary casting should be avoided. Instead, the variable `i` should be declared as `short`.

Even better: use `std::iota` to fill the buffer from 0 to n-1. But that's far beyond the score of this PR!  😇 

(I also think this `static_cast<short>` is beyond the scope of CTest-to-GTest conversion, of course.)

hjmjohnson: Fix is on it's way to change to short.

Modules/Core/Common/test/itkImportImageGTest.cxx
```diff
@@ -16,43 +16,42 @@
...
-#include "itkTestingMacros.h"
+#include "itkGTest.h"
+#include "itkMath.h"
+
+#include <iostream>
+
 
-int
-itkImportImageTest(int, char *[])
+TEST(ImportImageFilter, ImportAndShrink)
```
N-Dekker: Maybe we should consistently name a GTest unit test that is converted from CTest "ConvertedLegacyTest". So in this case:

    TEST(ImportImageFilter, ConvertedLegacyTest)

But of course, the proposed test name is OK if this test is specifically testing import-and-shrink 🤷

Modules/Core/Common/test/itkImportImageTest.cxx
```diff
@@ -76,50 +75,24 @@ itkImportImageTest(int, char *[])
...
     itk::ShrinkImageFilter<ImportImageFilter::OutputImageType, ShortImage>::New();
 
   shrink->SetInput(image);
-  shrink->SetShrinkFactors(2); // Also tested with factors 3 and 4, with 12x12 image
-
-  ITK_TRY_EXPECT_NO_EXCEPTION(shrink->Update());
-
-  //
-  // The rest of this code determines whether the shrink code produced
-  // the image we expected.
```
N-Dekker, line 90: It would have been nice if this comment could have been preserved.

Modules/Core/Common/test/itkImportImageGTest.cxx
hjmjohnson: Done.

Modules/Core/Common/test/itkImportImageTest.cxx
hjmjohnson: Done

PR #5907, opened 2026-03-08
ENH: Convert Heaviside step function tests to itkHeavisideStepFunctionGTest

hjmjohnson:


Modules/Core/Common/test/itkAtanRegularizedHeavisideStepFunctionTest.cxx
```diff
@@ -60,6 +66,7 @@ itkAtanRegularizedHeavisideStepFunctionTest1(int, char *[])
     const OutputType df = functionBase0->EvaluateDerivative(ix);
     std::cout << ix << ' ' << f << ' ' << df << std::endl;
   }
-
-  return EXIT_SUCCESS;
+  // At x=0 with epsilon=0.5: Evaluate = 0.5, EvaluateDerivative = 2/pi ≈ 0.63662
+  EXPECT_NEAR(functionBase0->Evaluate(0.0), 0.5, 1e-10);
+  EXPECT_NEAR(functionBase0->EvaluateDerivative(0.0), 2.0 / itk::Math::pi, 1e-4);
```
N-Dekker: Did Claude generate these EXPECT_NEAR calls, just by converting the old test? I'm surprised, because there doesn't seem to be a similar check in the old test!

Modules/Core/Common/test/itkHeavisideStepFunctionGTest.cxx
```diff
@@ -42,7 +50,14 @@ itkHeavisideStepFunctionTest1(int, char *[])
...
     std::cout << ix << ' ' << f << ' ' << df << std::endl;
+    if (ix < 0.0)
+    {
+      EXPECT_EQ(f, 0.0);
+      EXPECT_EQ(df, 0.0);
+    }
+    else
+    {
+      EXPECT_EQ(f, 1.0);
+    }
```
N-Dekker: Did Claude generate this `if` ... `else` automatically? I don't see anything similar in the original test.

N-Dekker:
This one produces more lines of code than the original test (lines changed: 56 additions & 45 deletions). In general, I would expect (and prefer) _less_ lines of code, when converting an old test to GTest  🤷

Modules/Core/Common/test/itkAtanRegularizedHeavisideStepFunctionTest.cxx
hjmjohnson: Yes. Claude did add some new tests, especially in cases where functions were run to create diagnostic outputs, but the outputs were never tested.

Modules/Core/Common/test/itkHeavisideStepFunctionGTest.cxx
hjmjohnson: Yes.

N-Dekker: Thanks @hjmjohnson Interesting! However, these extra checks are very much beyond the scope of Test-to-GoogleTest conversion. It is hard to review and it breaks the One Commit One Change principle.

N-Dekker (dismissed):
Please don't add extra checks with this pull request. Please just do Test-to-GoogleTest conversion, with this PR.

Modules/Core/Common/test/itkSinRegularizedHeavisideStepFunctionGTest.cxx
```diff
@@ -17,10 +17,12 @@
...
 #include "itkSinRegularizedHeavisideStepFunction.h"
-#include "itkTestingMacros.h"
+#include "itkGTest.h"
 
-int
-itkSinRegularizedHeavisideStepFunctionTest1(int, char *[])
+#include <iostream>
+
+// From itkSinRegularizedHeavisideStepFunctionTest1.cxx
+TEST(HeavisideStepFunction, SinRegularized)
```
N-Dekker: Very much improved, thanks Hans  😃 

Here I would suggest:

    TEST(SinRegularizedHeavisideStepFunction, ConvertedLegacyTest)

The problem I see with the current PR is that it has `TEST(HeavisideStepFunction, ...)` in multiple CXX files. So the name of the test suite "HeavisideStepFunction" is re-used. Until now, I thing we had the convention to have a unique test suite name for each CXX file.

blowekamp: Yes. At some point we should consider ways to merge multiple test cxx/gtest into a single file. This will improve compilation time and reduce disk space and cache usage.

hjmjohnson: This was originally a single file, but I had to manually split it to address the request that minimal changes be made, which is the source of the problem that @N-Dekker identified.

The next PR update is going to go back to merging these three into one test.

N-Dekker: Thanks Hans.

These three tests test three different classes from three different h + hxx files? So then why should they all be tested in the same test suite (and eventually in the same cxx test file)?

This is not meant to be a rhetorical question. I did introduce a cxx test file myself that tests multiple classes in one and the same test suite ([itkImageIteratorsGTest.cxx](https://github.com/InsightSoftwareConsortium/ITK/blob/7c30a367924cbb3dd8ac3e7c880e7d4c7857f47a/Modules/Core/Common/test/itkImageIteratorsGTest.cxx))

thewtex (approved):

PR #5908, opened 2026-03-08
STYLE: Remove redundant virtual statements for C++11

hjmjohnson:
Use virtual for the base class function declaration.  This is technically
necessary.

Remove 'virtual' is implied when 'override' is specified, so remove the
redundant specification.

cd ${BLDDIR}
run-clang-tidy.py -extra-arg=-D__clang__ -checks=-*,modernize-use-override  -header-filter=.* -fix

## PR Checklist

### DIFF
--- a/Modules/Segmentation/LevelSetsv4/include/itkLevelSetQuadEdgeMesh.h
+++ b/Modules/Segmentation/LevelSetsv4/include/itkLevelSetQuadEdgeMesh.h
@@ -64,23 +64,23 @@ public:
   itkGetModifiableObjectMacro(Mesh, MeshType);
 
   /** Returns the value of the level set function at a given location iP */
-  virtual OutputType
+  OutputType
   Evaluate(const InputType & iP) const override;
 
   /** Returns the image gradient of the level set function at a given location iP
    * \todo to be implemented */
-  virtual GradientType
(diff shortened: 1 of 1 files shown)


Modules/Segmentation/LevelSetsv4/include/itkLevelSetQuadEdgeMesh.h
```diff
@@ -64,23 +64,23 @@ class ITK_TEMPLATE_EXPORT LevelSetQuadEdgeMesh
   itkGetModifiableObjectMacro(Mesh, MeshType);
 
   /** Returns the value of the level set function at a given location iP */
-  virtual OutputType
+  OutputType
   Evaluate(const InputType & iP) const override;
```
N-Dekker, line 68: Thanks Hans! Note that the commit subject suggests that it adds good old C++11 `override` specifiers. But instead, it _removes_ virtual keywords. So I would suggest a different subject line, like:

    STYLE: Remove redundant `virtual` keywords from overrides

dzenanz: The body of the commit message is still contradictory 😄

Issue #1261, opened 2019-09-18
clang-8 scan-build report


gdevenyi:
Following some recent cleanup work with ANTs, I ran clang-8's scan-build against ITK master today.

The complete scan-build report can be downloaded from:
https://drive.google.com/open?id=1eZ0iTtLXTyZC6TXoRFXgH8vo1TWaE2vV

Skimming the report, most of the reported issues are in the ThirdParty modules. I have already reported a extensive buglist to the HDF5 folks.

There are a few bugs in Numerics and Core

gdevenyi:
Copy and paste from the index.html, for the non-thirdparty lines:

Bug Group | Bug Type | File ▾ | Function/Method | Line | Path Length |  
-- | -- | -- | -- | -- | -- | --
Logic error | Called C++ object pointer is null | scratch/ITK-build/KWStyle/Utilities/boost/xpressive/detail/dynamic/sequence.hpp | operator+= | 77 | 26 | View Report
Logic error | Called C++ object pointer is null | scratch/ITK-build/KWStyle/Utilities/boost/xpressive/detail/dynamic/sequence.hpp | operator+= | 77 | 18 | View Report
Logic error | Called C++ object pointer is null | scratch/ITK-build/KWStyle/Utilities/boost/xpressive/detail/dynamic/sequence.hpp | operator+= | 77 | 16 | View Report
Dead store | Dead assignment | scratch/ITK-build/KWStyle/kwsCheckIndent.cxx | CheckIndent | 224 | 1 | View Report
Dead store | Dead assignment | scratch/ITK-build/KWStyle/kwsCheckMemberFunctions.cxx | FindMemberFunction | 308 | 1 | View Report
Dead store | Dead assignment | scratch/ITK-build/KWStyle/kwsCheckOperator.cxx | FindOperator | 242 | 1 | View Report
Dead store | Dead assignment | scratch/ITK-build/Modules/ThirdParty/HDF5/src/itkhdf5/H5Tinit.c | itk_H5T__init_native | 967 | 1 | View Report
Memory error | Use-after-free | scratch/ants-build/ITKv5/Modules/Core/Common/include/itkSmartPointer.h | Register | 212 | 17 | View Report
Memory error | Use-after-free | scratch/ants-build/ITKv5/Modules/Core/Common/include/itkSmartPointer.h | UnRegister | 221 | 18 | View Report
[...]

PR #5909, opened 2026-03-09
BUG: Fix divide-by-zero and integer truncation in CalculateFitError

hjmjohnson:
## Summary

Fixes two bugs in `CumulativeGaussianCostFunction::CalculateFitError()` detected by clang-22 `scan-build` (`core.DivideZero` checker, line 55):

**Bug 1 — Divide by zero:** If `m_OriginalDataArray` has zero elements, `numberOfElements` is 0 and the division `1 / numberOfElements` is undefined behavior. The existing size-mismatch early-return guard now also covers this case.

**Bug 2 — Integer truncation (silent wrong result):** The expression `(1 / numberOfElements) * fitError` performs integer division first. For any `numberOfElements > 1`, `1 / numberOfElements` truncates to `0`, making `CalculateFitError()` always return `0.0` — silently reporting a perfect fit regardless of actual error. The corrected RMSE formula is `sqrt(fitError / numberOfElements)`.

## Test plan

- [x] Existing `itkCumulativeGaussianOptimizerTest` continues to pass
- [x] Verify `CalculateFitError` returns non-zero for arrays that differ

Detected via clang-22 `analyze-build` — see https://github.com/InsightSoftwareConsortium/ITK/issues/1261#issuecomment-4023451241

### DIFF
--- a/Modules/Numerics/Optimizers/src/itkCumulativeGaussianCostFunction.cxx
+++ b/Modules/Numerics/Optimizers/src/itkCumulativeGaussianCostFunction.cxx
@@ -43,7 +43,7 @@ CumulativeGaussianCostFunction::CalculateFitError(MeasureType * setTestArray)
   // Use root mean square error as a measure of fit quality.
   const unsigned int numberOfElements = m_OriginalDataArray.GetNumberOfElements();
 
-  if (numberOfElements != setTestArray->GetNumberOfElements())
+  if (numberOfElements != setTestArray->GetNumberOfElements() || numberOfElements == 0)
   {
     return 1;
   }
@@ -52,7 +52,7 @@ CumulativeGaussianCostFunction::CalculateFitError(MeasureType * setTestArray)
   {
     fitError += Math::sqr(setTestArray->get(i) - m_OriginalDataArray.get(i));
   }
-  return (std::sqrt((1 / numberOfElements) * fitError));
+  return std::sqrt(fitError / static_cast<double>(numberOfElements));
 }
 
 double


dzenanz (approved):

Modules/Numerics/Optimizers/src/itkCumulativeGaussianCostFunction.cxx
```diff
@@ -52,7 +52,7 @@ CumulativeGaussianCostFunction::CalculateFitError(MeasureType * setTestArray)
   {
     fitError += Math::sqr(setTestArray->get(i) - m_OriginalDataArray.get(i));
   }
-  return (std::sqrt((1 / numberOfElements) * fitError));
+  return std::sqrt(fitError / numberOfElements);
```
blowekamp: Please change this to` sqrt(1.0/numberOfElements)*fixError`, this change looks numerically unstable to me.

blowekamp (changes requested):

Modules/Numerics/Optimizers/src/itkCumulativeGaussianCostFunction.cxx
hjmjohnson: I went with 

```c++
  return std::sqrt(fitError / static_cast<double>(numberOfElements));
```
to explicitly state what was implicitly done, and avoid the extra math operation.

dzenanz (approved):

Modules/Numerics/Optimizers/src/itkCumulativeGaussianCostFunction.cxx
dzenanz: I don't understand either why the existing formula would be numerically unstable. Also, your suggestion Brad takes `fitError` outside of the square root, changing the formula.

blowekamp (approved):
Thank you for responding to my comment.

I mistakenly saw a small number being divided by a large number, but preserving the operation is fine.

PR #5910, opened 2026-03-09
PERF: Add ccache/sccache support to GitHub Actions CI workflows

hjmjohnson:
## Summary

- Add `actions/cache@v4` to `pixi.yml` (ubuntu, windows, macos) and `arm.yml` (ubuntu-arm, macos rosetta, macos Python) to cache compiled objects between CI runs
- Add `configure-ci` pixi task in `pyproject.toml` that passes `-DCMAKE_C_COMPILER_LAUNCHER=ccache` and `-DCMAKE_CXX_COMPILER_LAUNCHER=ccache` to CMake; `pixi.yml` uses this task instead of `configure`
- `arm.yml` injects the same cmake launcher variables via `dashboard_cache` so the CTest dashboard script picks them up at configure time
- Cache sized at 11G to accommodate the full ITK object file footprint; keyed on OS + `pixi.lock` hash (pixi workflow) or OS + matrix name (arm workflow) to avoid stale cache on toolchain upgrades

## Test plan

- [x] Verify CI runs complete successfully on first run (cold cache — expect no speedup)
- [x] Verify subsequent CI runs show ccache hit rate > 0% in the ccache-action summary
- [x] Confirm no build failures introduced by the compiler launcher on all three OS platforms in `pixi.yml`
- [x] Confirm arm workflow builds succeed with ccache launcher on ubuntu-arm, macos rosetta, and macos Python matrix entries

### DIFF
--- a/.github/workflows/arm.yml
+++ b/.github/workflows/arm.yml
@@ -94,6 +94,7 @@ jobs:
           echo "CCACHE_BASEDIR=${GITHUB_WORKSPACE}" >> "$GITHUB_ENV"
           echo "CCACHE_COMPILERCHECK=content" >> "$GITHUB_ENV"
           echo "CCACHE_NOHASHDIR=true" >> "$GITHUB_ENV"
+          echo "CCACHE_NODIRECT=1" >> "$GITHUB_ENV"
           echo "CCACHE_DIR=${{ runner.temp }}/ccache" >> "$GITHUB_ENV"
           echo "CCACHE_MAXSIZE=11G" >> "$GITHUB_ENV"
           if [ "$RUNNER_OS" == "Linux" ]; then
--- a/.github/workflows/pixi.yml
+++ b/.github/workflows/pixi.yml
@@ -53,6 +53,7 @@ jobs:
             echo "CCACHE_BASEDIR=${GITHUB_WORKSPACE}" >> "$GITHUB_ENV"
             echo "CCACHE_COMPILERCHECK=content" >> "$GITHUB_ENV"
             echo "CCACHE_NOHASHDIR=true" >> "$GITHUB_ENV"
+            echo "CCACHE_NODIRECT=1" >> "$GITHUB_ENV"
             echo "CCACHE_DIR=${{ runner.temp }}/ccache" >> "$GITHUB_ENV"
             echo "CCACHE_MAXSIZE=11G" >> "$GITHUB_ENV"
             if [ "$RUNNER_OS" == "Linux" ]; then
--- a/Testing/ContinuousIntegration/AzurePipelinesBatch.yml
+++ b/Testing/ContinuousIntegration/AzurePipelinesBatch.yml
@@ -16,6 +16,7 @@ variables:
   CCACHE_BASEDIR: $(Build.SourcesDirectory)
   CCACHE_COMPILERCHECK: content
   CCACHE_NOHASHDIR: 'true'
+  CCACHE_NODIRECT: '1'
 jobs:
   - job: Windows
     timeoutInMinutes: 0
--- a/Testing/ContinuousIntegration/AzurePipelinesLinux.yml
+++ b/Testing/ContinuousIntegration/AzurePipelinesLinux.yml
@@ -32,6 +32,7 @@ variables:
   CCACHE_BASEDIR: $(Build.SourcesDirectory)
   CCACHE_COMPILERCHECK: content
   CCACHE_NOHASHDIR: 'true'
+  CCACHE_NODIRECT: '1'
 jobs:
 - job: Linux
   timeoutInMinutes: 0
--- a/Testing/ContinuousIntegration/AzurePipelinesLinuxPython.yml
+++ b/Testing/ContinuousIntegration/AzurePipelinesLinuxPython.yml
@@ -32,6 +32,7 @@ variables:
   CCACHE_BASEDIR: $(Build.SourcesDirectory)
   CCACHE_COMPILERCHECK: content
   CCACHE_NOHASHDIR: 'true'
+  CCACHE_NODIRECT: '1'
 jobs:
 - job: Linux
   timeoutInMinutes: 0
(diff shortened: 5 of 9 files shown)


dzenanz (commented):
Looks good.

Testing/ContinuousIntegration/AzurePipelinesBatch.yml
```diff
@@ -86,6 +99,7 @@ jobs:
         displayName: 'Build and test'
         env:
           CTEST_OUTPUT_ON_FAILURE: 1
+          CCACHE_MAXSIZE: 11G
```
dzenanz, line 109: Commit message claims 5G cache.

hjmjohnson: Fixed comment.

dzenanz (approved):

.github/workflows/arm.yml
```diff
@@ -88,6 +88,13 @@ jobs:
           fetch-depth: 5
           clean: true
 
+      - name: Set up ccache
+        uses: hendrikmuhs/ccache-action@v1.2
+        with:
+          key: ${{ matrix.os }}-${{ matrix.name }}-ccache
```
blowekamp: The documentation of the hash key is here:
https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching

The has key is matched from the beginning to the end, so it should start with the less specific an go to the more specific. I suggest the following:

ccache-v1-${{ matrix.os }}-${{ matrix.name }}-${{branch-name}}

Additionally with the raw action/cache there is a restore key field that can take multiple lines. This should have the above first then `ccache-v1-${{ matrix.os }}-main` so that is falls back to the cache on the main. I don't recall the more advance logic to get the "release" cache to work correctly for PR to the release branch.... maybe a variable for the target branch name would be appropriate as a heigher priority.

CMake/itkCCacheSupport.cmake
```diff
@@ -54,8 +54,10 @@ if(ITK_USE_CCACHE)
     message(FATAL_ERROR "ccache not found, turn off ITK_USE_CCACHE")
   endif()
   if(CCACHE_VERSION VERSION_LESS 4.0)
+    # CI actions may use sccache for increased performance, which does not return
+    # a valid CCACHE_VERSION. Provide a STATUS warning instead of outright failure
     message(
-      FATAL_ERROR
+      STATUS
```
blowekamp, line 60: I have been using cache with ITK/SimpleITK... When doing the ITK with FetchContent and cache I had to do the following:
https://github.com/SimpleITK/SimpleITK/blob/faf841a0fbf8d120186c39ff0cd5ea233a837142/CMake/sitkITKFetchContent.cmake#L56-L60

So I had to set ITK_USE_CCACHE to OFF the get the ccache to be used. There is some odd logic this this file that needs review. I think this is just a tangent, not related to the PR, but I need to check the behavior with the change.

hjmjohnson: @blowekamp Let me know if more changes will need to be made after testing with SimpleITK builds.

.github/workflows/arm.yml
hjmjohnson: I updated but dropped the ${{branch-name}} for the moment.   I want to get the cache filled.  I'll look into the other caching fallback names later today.

CMake/itkCCacheSupport.cmake
blowekamp: Adding an additional field for the branch name to the end should not necessitate the need to change the "v1" due to change in the name conventions.

Also note that for PR's made from other's forks, the cache is pull from their store not the ISC's cache store.

.github/workflows/arm.yml
hjmjohnson: Thanks @blowekamp, you suggestion prompted me to review the two actions that can deal with caching.  Your comments were relevant to `actions/cache@v4` instead of `hendrikmuhs/ccache-action@v1.2`.  Upon deeper consideration, using the github supported `actions/cache@v4` 
I defaulted to actions/cache@v4 plus explicit ccache setup to minimize inter-project dependencies and prefer more dependable tools.

.github/workflows/arm.yml
```diff
@@ -88,6 +88,33 @@ jobs:
...
+          echo "CCACHE_DIR=${{ runner.temp }}/ccache" >> "$GITHUB_ENV"
+          echo "CCACHE_MAXSIZE=11G" >> "$GITHUB_ENV"
+          if [ "$RUNNER_OS" == "Linux" ]; then
+            sudo apt-get update -qq && sudo apt-get install -y ccache
+          elif [ "$RUNNER_OS" == "macOS" ]; then
+            brew install ccache
+          fi
+
+      - name: Set up compiler cache
+        uses: actions/cache@v4
```
thewtex, line 107: We should use v5 if possible: https://github.com/actions/cache

.github/workflows/pixi.yml
```diff
@@ -47,6 +47,34 @@ jobs:
...
+            if [ "$RUNNER_OS" == "Linux" ]; then
+              sudo apt-get update -qq && sudo apt-get install -y ccache
+            elif [ "$RUNNER_OS" == "macOS" ]; then
+              brew install ccache
+            elif [ "$RUNNER_OS" == "Windows" ]; then
+              choco install ccache --yes
+            fi
+
+        - name: Set up compiler cache
+          uses: actions/cache@v4
```
thewtex, line 68: v5

hjmjohnson: Done.

.github/workflows/arm.yml
hjmjohnson: Done.

.github/workflows/arm.yml
```diff
@@ -83,11 +83,39 @@ jobs:
...
+        with:
+          path: ${{ runner.temp }}/ccache
+          key: ccache-v1-${{ matrix.os }}-${{ matrix.name }}
+          restore-keys: |
+            ccache-v1-${{ matrix.os }}-
+            ccache-v1-
+
+      - name: Evict old ccache entries
+        shell: bash
```
blowekamp, line 116: @hjmjohnson I don't understand the why this additional cache management is needed? Can you please shed some insights?

PR #5913, opened 2026-03-09
STYLE: Use DefaultPixelAccessor in itkImageAdaptorGTest

N-Dekker:
Removed the locally defined `DummyPixelAccessor`, which was very similar to `itk::DefaultPixelAccessor` anyway.

### DIFF
--- a/Modules/Core/ImageAdaptors/test/itkImageAdaptorGTest.cxx
+++ b/Modules/Core/ImageAdaptors/test/itkImageAdaptorGTest.cxx
@@ -20,6 +20,7 @@
 #include "itkImageAdaptor.h"
 
 #include "itkCovariantVector.h"
+#include "itkDefaultPixelAccessor.h"
 #include "itkImage.h"
 #include "itkImageBase.h"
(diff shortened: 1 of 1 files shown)


Modules/Core/ImageAdaptors/test/itkImageAdaptorGTest.cxx
```diff
@@ -32,27 +33,6 @@
...
-  {
-    output = input;
-  }
-
-  static TPixelType
-  Get(const TPixelType & input)
-  {
-    return input;
-  }
-};
```
N-Dekker, line 53: For comparison, here is `itk::DefaultPixelAccessor`: https://github.com/InsightSoftwareConsortium/ITK/blob/4169ee538d3b58ef2c24219752f880e13f9b6918/Modules/Core/Common/include/itkDefaultPixelAccessor.h#L53-L85

hjmjohnson: Thanks.

PR #5914, opened 2026-03-10
WIP: Test `Math::Absolute(std::numeric_limits<int>::min())`

N-Dekker:
Checks that `Absolute(INT_MIN)` returns `INT_MAX + 1U`.

Note that `std::abs(int)` does not support `INT_MIN` as argument.

----

Example at https://godbolt.org/z/dPTxjdqoK built with GCC (`-std=c++23`):

```c++
#include <climits>
#include <cmath>

constexpr auto absMax = std::abs(INT_MAX); // OK
constexpr auto absMin = std::abs(INT_MIN); // error: overflow in constant expression
```

### DIFF
--- a/Modules/Core/Common/test/itkMathGTest.cxx
+++ b/Modules/Core/Common/test/itkMathGTest.cxx
@@ -334,6 +334,9 @@ TEST(itkMath, Abs)
   static_assert(itk::Math::Absolute<double>(-5.0) == 5.0);
   static_assert(itk::Math::Absolute<float>(-5.0f) == 5.0f);
 
+  // Note that `std::abs(int)` does not support `std::numeric_limits<int>::min()` as argument.
+  static_assert(itk::Math::Absolute(std::numeric_limits<int>::min()) == std::numeric_limits<int>::max() + 1U);
+
   constexpr auto cf = std::complex<float>(-3, -4);
   EXPECT_EQ(itk::Math::Absolute(cf), 5);
   constexpr auto cd = std::complex<double>(-3, -4);


Modules/Core/Common/test/itkMathGTest.cxx
```diff
@@ -334,6 +334,9 @@ TEST(itkMath, Abs)
   static_assert(itk::Math::Absolute<double>(-5.0) == 5.0);
   static_assert(itk::Math::Absolute<float>(-5.0f) == 5.0f);
 
+  // Note that `std::abs(int)` does not support `std::numeric_limits<int>::min()` as argument.
+  static_assert(itk::Math::Absolute(std::numeric_limits<int>::min()) == std::numeric_limits<int>::max() + 1U);
```
N-Dekker, line 338: [ARMBUILD-Ubuntu-24.04-arm](https://github.com/InsightSoftwareConsortium/ITK/actions/runs/22902727872/job/66453137709?pr=5914#logs) says:

```
itkMathGTest.cxx:338:70: error: non-constant condition for static assertion
  338 |   static_assert(itk::Math::Absolute(std::numeric_limits<int>::min()) == std::numeric_limits<int>::max() + 1U);
      |                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
itkMathGTest.cxx:338:36:   in 'constexpr' expansion of 'itk::Math::Absolute<int>(std::numeric_limits<int>::min())'
itkMathGTest.cxx:338:70: error: overflow in constant expression [-fpermissive]

```

I believe this is because `Math::Absolute<int>(x)` internally does `-x`, at https://github.com/InsightSoftwareConsortium/ITK/blob/969b2ca90d754b6f93896c0a4091f9123270e4a5/Modules/Core/Common/include/itkMath.h#L900

N-Dekker: So if we want to support `Math::Absolute(INT_MIN)`, we should adjust `Math::Absolute` to do so. Moreover, we cannot use `std::abs`, if we want to support `INT_MIN`.

@hjmjohnson What do you think?

N-Dekker: For signed integer types (from int8 to int64), we could implement `Math::Absolute(x)` as follows:
```c++
using ReturnType = std::make_unsigned_t<T>;
using Limits = std::numeric_limits<T>;

return static_cast<ReturnType>(
  (x < 0) ? (x == Limits::min()) ? (Limits::max() + ReturnType{1}) : -x : x);
```

What do you think?

Issue #5823, opened 2026-02-23
Possible ODR violation from `itk::Math::Absolute(T x)` when mixing C++ standard versions?


N-Dekker:
### Description

I think it's possible to build ITK with a different C++ standard version than the user application. So for example, someone might build ITK with C++20 and the user application with C++23, or vise versa. Do we actually support such a use case?

If we do indeed support mixing C++ standard versions like that, I'm afraid that the current definition of `itk::Math::Absolute(T x)` violates the One Definition Rule ([ODR](https://en.cppreference.com/w/cpp/language/definition.html)): https://github.com/InsightSoftwareConsortium/ITK/blob/4e88c8e53c2a41aca55f4c4635a8ea4e6c8bc2e2/Modules/Core/Common/include/itkMath.h#L878-L945

The function definition is different for C++23, that is for `__cplusplus >= 202302`. Because the function is defined in a header file, its definition within the context of the user application may be different from its definition within ITK, when different C++ standard versions are used.

### Versions

This possible issue was introduced by pull request #5797 commit 3cc3f4a1e6347e0c46a6ef1e49bbcabba2005fd9 (merged on Feb 14, 2026), so it is not yet in a tagged or a released version of ITK

### Possible solution

If this is indeed an issue, we might simply remove the C++23 specific code.

dzenanz:
Welcome to the ITK community Niels 😄

PR #5917, opened 2026-03-10
BUG: Fix overflow `Math::Absolute` on a minimum signed integer argument

N-Dekker:
An attempt to evaluate `Math::Absolute(INT_MIN)` at compile-time caused a compile error triggered by the expression `-x`, saying:

    value 2147483648 is outside the range of representable values of type 'int'

`std::abs(INT_MIN)` also has integer overflow.

Fixed by having `Math::Absolute(T x)` specifically take care of the case when its argument is a signed integer with value `std::numeric_limits<T>::min()`.

----

- Aims to supersede pull request #5914
- Partially (!) addresses issue #5823

### DIFF
--- a/Modules/Core/Common/include/itkMath.h
+++ b/Modules/Core/Common/include/itkMath.h
@@ -31,6 +31,7 @@
 #include <cassert>
 #include <complex>
 #include <cmath>
+#include <limits>
 #include <type_traits>
 #include "itkMathDetail.h"
 #include "itkConceptChecking.h"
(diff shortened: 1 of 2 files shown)


dzenanz (approved):

issakomi:
> std::abs(INT_MAX) also has integer overflow

I think there is the problem with `std::abs(INT_MIN)`, not with `std::abs(INT_MAX)`

issakomi:
> Good catch, @issakomi ! I think I only made this mistake in the commit message, not in the code. Right?

I think so. Thanks for the PR.

hjmjohnson (approved):

PR #5928, opened 2026-03-12
ENH: Only save ccache on main branch, restore for all PRs

### LINKED ISSUE #5927: Investigate CCACHE utilization
### Description

Using ccache will require more careful implementation to gain benefits.  

There is a combined 10G limit (by default) for all ccaches.  We are currently round-robin-invalidating most potential benefits partially due to too many fine-grained ccaches.

### Impact analysis

For example,  if 20 fine-grained caches at 1G each were made during every PR, then almost all caches will be evacuated before they are re-used due to the limit.  Even within a single PR, only the last 10 caches would be preserved, and then when a new PR started, the first 10 caches created would evacuate the previous 10 caches.

### Expected behavior

We should maintain a set of consistent caches that are mostly re-used, perhaps only 1 cache that is only updated on the upstream/main branch.

### Actual behavior

Each PR makes it's own cache for every build.

A very careful consideration of behaviors added to actions/cache@v5  and recommended strategies should be considered.

https://github.com/actions/cache/blob/main/caching-strategies.md

### Environment

Github cache.

Review behaviors in https://github.com/InsightSoftwareConsortium/ITK/actions/caches

hjmjohnson:
## Summary
- Split `actions/cache@v5` into `actions/cache/restore@v5` (all runs) and `actions/cache/save@v5` (main branch pushes only) in both `pixi.yml` and `arm.yml`
- PR builds now **restore** the main branch's ccache but never **save** their own cache entries
- Main branch pushes **save** the cache keyed by `${{ matrix.os }}` (pixi) or `${{ matrix.os }}-${{ matrix.name }}` (arm), creating a stable set of ~6 cache entries total instead of unbounded per-PR caches

## Motivation (Issue #5927)
GitHub Actions has a combined 10GB cache limit per repository. Previously every PR build created its own per-OS cache entry (via the `key` + `restore-keys` pattern in `actions/cache@v5`). With 20+ fine-grained caches at ~1GB each, the round-robin eviction meant most caches were evicted before reuse — defeating the purpose of caching.

## How it works
| Event | Restore | Save |
|-------|---------|------|
| `push` to `main` | `actions/cache/restore@v5` with exact key | `actions/cache/save@v5` with same key |
| `pull_request` | `actions/cache/restore@v5` with exact key (matches main's cache) | **skipped** — no cache written |

The `actions/cache/save` step overwrites the previous cache for that key on each main push, keeping a fresh, stable cache per OS/build-variant.

## Cache keys
- **pixi.yml:** `ccache-v1-${{ matrix.os }}-pixi-cxx` (3 entries: ubuntu, windows, macos)
- **arm.yml:** `ccache-v1-${{ matrix.os }}-${{ matrix.name }}` (3 entries: Ubuntu-arm, x86_64-rosetta, Python)

## Test plan
- [ ] Verify main branch push creates cache entries for each OS
- [ ] Verify PR builds restore the main branch cache (check "Restore compiler cache" step output)
- [ ] Verify PR builds do NOT create new cache entries (no "Save compiler cache" step runs)
- [ ] Confirm cache usage stays well within the 10GB limit after several PRs

Closes #5927

### DIFF
--- a/.github/workflows/pixi.yml
+++ b/.github/workflows/pixi.yml
@@ -64,13 +64,12 @@ jobs:
               choco install ccache --yes
             fi
 
-        - name: Set up compiler cache
-          uses: actions/cache@v4
+        - name: Restore compiler cache
+          id: restore-ccache
+          uses: actions/cache/restore@v5
           with:
             path: ${{ runner.temp }}/ccache
-            key: ccache-v1-${{ matrix.os }}-pixi-cxx-${{ hashFiles('pixi.lock') }}
-            restore-keys: |
-              ccache-v1-${{ matrix.os }}-pixi-cxx-
+            key: ccache-v1-${{ matrix.os }}-pixi-cxx
 
         - name: Show ccache configuration, stats and maintenance
           shell: bash
@@ -149,6 +148,13 @@ jobs:
             df -h
             echo "****** df -h /"
             df -h /
+        - name: Save compiler cache
+          if: steps.restore-ccache.outputs.cache-hit != 'true' || github.ref == 'refs/heads/main'
+          uses: actions/cache/save@v5
+          with:
+            path: ${{ runner.temp }}/ccache
+            key: ccache-v1-${{ matrix.os }}-pixi-cxx
+
         - name: ccache stats
           if: always()
           run: ccache --show-stats
--- a/.github/workflows/arm.yml
+++ b/.github/workflows/arm.yml
@@ -103,14 +103,12 @@ jobs:
             brew install ccache
           fi
 
-      - name: Set up compiler cache
-        uses: actions/cache@v4
+      - name: Restore compiler cache
+        id: restore-ccache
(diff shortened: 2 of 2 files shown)


dzenanz (approved):
LGTM

.github/workflows/pixi.yml
```diff
@@ -149,6 +148,13 @@ jobs:
             df -h
             echo "****** df -h /"
             df -h /
+        - name: Save compiler cache
+          if: steps.restore-ccache.outputs.cache-hit != 'true' || github.ref == 'refs/heads/main'
```
blowekamp, line 152: With "steps.restore-ccache.outputs.cache-hit != 'true'" being here ( which is an exact key match), when will the cache actually be updated? Only when there are not caches available?

hjmjohnson: When the builds occur for the push of the main branch "github.ref == 'refs/heads/main'
"

PR #5931, opened 2026-03-12
DOC: Improve `Math::Absolute(x)` doc on `INT_MIN` and `bool` support

N-Dekker:
Removed the text, saying that the function "typecasts to a larger storage type", as those typecasts were removed by pull request https://github.com/InsightSoftwareConsortium/ITK/pull/5917 commit 381bafa4effe8f8e2a5ede51d21a9aad1d0f82b2 "BUG: Fix overflow `Math::Absolute` on a minimum signed integer argument"

### DIFF
--- a/Modules/Core/Common/include/itkMath.h
+++ b/Modules/Core/Common/include/itkMath.h
@@ -861,16 +861,17 @@ using vnl_math::squared_magnitude;
  * @brief Returns the absolute value of a number.
  *
  * This function provides a c++17 implementation of the vnl_math::abs absolute value functionality.
- * itk::Absolute() preserves backward compatibility with vnl_math::abs that was originally used in ITK.
+ * Math::Absolute() preserves backward compatibility with vnl_math::abs that was originally used in ITK.
  *
- * Key differences between itk::Absolute() and std::abs()
- *     - Where std::abs() returns the absolute value in the same type as the
- *       input, itk::Absolute() returns the absolute value in the unsigned
- *       equivalent of the input type.
- *     - std::abs() is undefined when converting the minimum storage value of unsigned
- *       integers to positive values.  itk::Absolute() typecasts to a larger storage type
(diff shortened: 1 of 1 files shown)


Modules/Core/Common/include/itkMath.h
```diff
@@ -861,16 +861,17 @@ using vnl_math::squared_magnitude;
...
- *       all cases except 'signed long long' where there is no larger storage type.
+ * Math::Absolute() preserves backward compatibility with vnl_math::abs that was originally used in ITK.
+ *
+ * Key differences between Math::Absolute() and std::abs():
+ * - Where `std::abs(x)` returns the absolute value in the same type as the input, `Math::Absolute(x)` returns the
+ * absolute value in the unsigned equivalent of the input type, when the input is an integer.
+ * - `std::abs(x)` has undefined or compiler specific behavior when its argument is the minimum value of `int`, `long`,
+ * or `long long`. On the other hand, Math::Absolute _does_ support the entire range of possible integer values. So for
+ * example `Math::Absolute(INT_MIN)` properly returns 2147483648 as unsigned integer (assuming that `INT_MIN` has the
+ * value -2147483648).
```
N-Dekker, line 872: FYI, https://cplusplus.com/reference/cstdlib/abs says: 

> If the result cannot be represented by the returned type (such as `abs(INT_MIN)` in an implementation with two's complement signed values), it causes _**undefined behavior**_.

But on the other hand, https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/abs-labs-llabs-abs64?view=msvc-170#remarks says:

> **Microsoft-specific:** [...] Specifically, `abs(INT_MIN)` returns `INT_MIN`, `labs(LONG_MIN)` returns `LONG_MIN`, `llabs(LLONG_MIN)` returns `LLONG_MIN`,... 

So that's why I concluded that it is _either_ undefined _or_ compiler specific!  🤓

dzenanz (approved):

PR #5932, opened 2026-03-12
STYLE: Use CTAD for iterators "with index", in hxx files

N-Dekker:
Used class template argument deduction when declaring "IteratorWithIndex" variables.

    Find what: ([^\w]Image\w+IteratorWithIndex)<\w+>[ ]*( [a-z]\w*[\({])
    Replace with: $1$2
    Files: itk*.hxx
    [v] Match case
    (*) Regular expression


dzenanz (approved):
Were any manual interventions needed?

PR #5935, opened 2026-03-12
PERF: Refine CI ccache behaviors

hjmjohnson:
## Summary
- Only save ccache on main branch pushes, preventing PR builds from creating cache entries that exceed the 10GB GitHub Actions cache limit
- Reduce `CCACHE_MAXSIZE` from 11G to 2.4G across all GitHub Actions and Azure Pipelines CI configurations to minimize storage of stale object files (2.4G is ~10% more than currently needed for the largest Python builds)

## Test plan
- [x] Verify GitHub Actions workflows pass on Linux, macOS, and ARM
- [x] Verify Azure Pipelines pass on Linux, macOS, and Windows (C++ and Python)
- [x] Confirm ccache hit rates remain acceptable after size reduction

### DIFF
--- a/.github/workflows/arm.yml
+++ b/.github/workflows/arm.yml
@@ -169,7 +169,7 @@ jobs:
         env:
           CTEST_OUTPUT_ON_FAILURE: 1
       - name: Save compiler cache
-        if: steps.restore-ccache.outputs.cache-hit != 'true' || github.ref == 'refs/heads/main'
+        if: github.ref == 'refs/heads/main'
         uses: actions/cache/save@v5
         with:
           path: ${{ runner.temp }}/ccache
--- a/.github/workflows/arm.yml
+++ b/.github/workflows/arm.yml
@@ -96,7 +96,7 @@ jobs:
           echo "CCACHE_NOHASHDIR=true" >> "$GITHUB_ENV"
           echo "CCACHE_NODIRECT=1" >> "$GITHUB_ENV"
           echo "CCACHE_DIR=${{ runner.temp }}/ccache" >> "$GITHUB_ENV"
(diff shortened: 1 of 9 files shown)


.github/workflows/arm.yml
```diff
@@ -169,7 +169,7 @@ jobs:
         env:
           CTEST_OUTPUT_ON_FAILURE: 1
       - name: Save compiler cache
-        if: steps.restore-ccache.outputs.cache-hit != 'true' || github.ref == 'refs/heads/main'
```
hjmjohnson, line 172: @blowekamp I was hoping to `warm-bootstrap` the cache when the `main` branch cache did not exist.  Unfortunately, this caused each PR to make a cache for each PR that would only be used for that PR, and they would not expire for a week. That pushed us over the 10G limit and caused cache evictions due to quota limits that would start evicting the main branch caches.

.github/workflows/pixi.yml
```diff
@@ -55,7 +55,7 @@ jobs:
             echo "CCACHE_NOHASHDIR=true" >> "$GITHUB_ENV"
             echo "CCACHE_NODIRECT=1" >> "$GITHUB_ENV"
             echo "CCACHE_DIR=${{ runner.temp }}/ccache" >> "$GITHUB_ENV"
-            echo "CCACHE_MAXSIZE=11G" >> "$GITHUB_ENV"
+            echo "CCACHE_MAXSIZE=2.4G" >> "$GITHUB_ENV"
```
hjmjohnson, line 58: After monitoring the caches, the 2.4G limit seems like a good amount to avoid filling up the cache with outdated objects.

blowekamp (approved):

PR #5939, opened 2026-03-13
STYLE: Convert CreateTestDriver from macro to function

hjmjohnson:
## Summary
- Convert `CreateTestDriver` from a CMake `macro()` to `function()` in `CMake/ITKModuleTest.cmake`
- Modern CMake recommends `function()` over `macro()` because functions have their own variable scope and prevent accidental variable leakage into the caller's scope
- Verified no callers depend on any variables leaking from the macro (`ADDITIONAL_SRC`, `Tests`, `CMAKE_TESTDRIVER_BEFORE_TESTMAIN`, `CMAKE_TESTDRIVER_AFTER_TESTMAIN`, `emscripten_before`, `emscripten_after`)
- `CreateGoogleTestDriver` already uses `function()` — this makes the two consistent

## Test plan
- [ ] Verify full ITK build completes successfully (all test drivers compile)
- [ ] Verify Emscripten cross-compile path still works (if applicable)
- [ ] Verify `ctest` runs pass (test discovery unchanged)

### DIFF
--- a/CMake/ITKModuleTest.cmake
+++ b/CMake/ITKModuleTest.cmake
@@ -1,4 +1,4 @@
-# This file contains CMake functions and macros used when testing ITK modules.
+# This file contains CMake functions used when testing ITK modules.
 
 #-----------------------------------------------------------------------------
 # Create source code, compile and link a test driver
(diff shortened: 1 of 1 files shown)


dzenanz (approved):
LGTM

PR #5940, opened 2026-03-13
COMP: Address scan-build warnings

hjmjohnson:
## Summary
- Suppress intentional `cplusplus.Move` warning in `itkArray2DGTest.cxx` move-semantics tests that deliberately verify moved-from object state
- Fix `cplusplus.NewDelete` zero-sized allocation in `itkQuadEdgeMeshCellInterfaceTest.cxx` by guarding with `numberOfPoints > 0`
- `cplusplus.NewDelete` findings in `itkSmartPointer.h` are confirmed false positives (refcount prevents use-after-free); no code change needed

## Test plan
- [x] `ITKCommonGTestDriver` builds and all `Array2D.*` tests pass
- [x] `ITKQuadEdgeMeshTestDriver` builds and `itkQuadEdgeMeshCellInterfaceTest` passes
- [ ] CI passes

Addresses #1261 ([scan-build report comment](https://github.com/InsightSoftwareConsortium/ITK/issues/1261#issuecomment-4023451241))

### DIFF
--- a/Modules/IO/TIFF/src/itkTIFFImageIO.cxx
+++ b/Modules/IO/TIFF/src/itkTIFFImageIO.cxx
@@ -1004,8 +1004,12 @@ TIFFImageIO::AllocateTiffPalette(uint16_t bps)
   m_ColorGreen = nullptr;
   m_ColorBlue = nullptr;
 
-  // bpp is 16 at maximum for palette image
-  const tmsize_t array_size = tmsize_t{ 1 } << bps * sizeof(uint16_t);
+  // TIFF palette images support at most 16 bits per sample per the TIFF spec.
+  if (bps > 16)
+  {
+    itkExceptionMacro("TIFF palette images require at most 16 bits per sample, but got " << bps);
+  }
+  const tmsize_t array_size = (tmsize_t{ 1 } << bps) * sizeof(uint16_t);
   m_ColorRed = static_cast<uint16_t *>(_TIFFmalloc(array_size));
   if (!m_ColorRed)
   {
--- a/Modules/Core/Common/include/itkSingleton.h
+++ b/Modules/Core/Common/include/itkSingleton.h
@@ -140,7 +140,7 @@ Singleton(const char * globalName, std::function<void()> deleteFunc)
   T *                                      instance = SingletonIndex::GetInstance()->GetGlobalInstance<T>(globalName);
   if (instance == nullptr)
   {
-    instance = new T;
+    instance = new T{};
     SingletonIndex::GetInstance()->SetGlobalInstance<T>(globalName, instance, std::move(deleteFunc));
   }
   return instance;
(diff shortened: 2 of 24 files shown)


dzenanz (approved):

Modules/IO/MeshOFF/src/itkOFFMeshIO.cxx
```diff
@@ -286,7 +286,7 @@ OFFMeshIO::ReadPoints(void * buffer)
 void
 OFFMeshIO::ReadCells(void * buffer)
 {
-  const auto data = make_unique_for_overwrite<itk::uint32_t[]>(this->m_CellBufferSize - this->m_NumberOfCells);
+  const auto data = std::make_unique<itk::uint32_t[]>(this->m_CellBufferSize - this->m_NumberOfCells);
```
N-Dekker: `make_unique_for_overwrite` was added to allow creating an uninitialized buffer. Which might be slightly faster than creating an _initialized_ buffer (`std::make_unique`). _And_ it explicitly expresses the intend to overwrite the buffer. Should we never create uninitialized buffers anymore, from now? Just wondering  🤷

hjmjohnson: @N-Dekker   I did recognize this and was going to bring it to your attention.  I wanted to look at it more closely.  The analyze-build tool stated that it detected that incomplete initialization may occur in the subsequent steps.

Perhaps that is a false positive, or an additional initialization is needed.

I also researched differences between make_unique_for_overwrite and std::make_unique, and the performance difference is almost always negligible.  Perhaps this is a case where it is not negligible.

hjmjohnson: @N-Dekker This commit has been isolated to 
- [ ] https://github.com/InsightSoftwareConsortium/ITK/pull/5943 so that it can be considered separately.

Modules/IO/TIFF/src/itkTIFFImageIO.cxx
```diff
@@ -1004,8 +1004,12 @@ TIFFImageIO::AllocateTiffPalette(uint16_t bps)
...
   m_ColorBlue = nullptr;
 
-  // bpp is 16 at maximum for palette image
-  const tmsize_t array_size = tmsize_t{ 1 } << bps * sizeof(uint16_t);
+  // TIFF palette images support at most 16 bits per sample per the TIFF spec.
+  if (bps > 16)
+  {
+    itkExceptionMacro("TIFF palette images require at most 16 bits per sample, but got " << bps);
+  }
+  const tmsize_t array_size = (tmsize_t{ 1 } << bps) * sizeof(uint16_t);
```
dzenanz, line 1012: The commit message sounds outrageous. Is this correct? @issakomi @malaterre

dzenanz (commented):
Mostly looks good.

PR #5944, opened 2026-03-13
COMP: Use setter/getter macros for TemporalProcessObject members

hjmjohnson:
## Summary
- Replace `this->TemporalProcessObject::m_*` direct member access with `Set*()`/`Get*()` macros in `DummyVideoToVideoFilter` test helper
- Fixes compilation error on compilers that cannot resolve grandparent class names through dependent template base classes (`TemporalProcessObject` is two levels up: `VideoToVideoFilter` → `VideoSource` → `TemporalProcessObject`)

## Test plan
- [x] `itkVideoToVideoFilterGTest` compiles and passes on all CI platforms
- [ ] No functional change — uses the same protected setter/getter macros already provided by `TemporalProcessObject`

### DIFF
--- a/Modules/Video/Core/test/itkVideoToVideoFilterGTest.cxx
+++ b/Modules/Video/Core/test/itkVideoToVideoFilterGTest.cxx
@@ -88,10 +88,10 @@ protected:
   /** Constructor */
   DummyVideoToVideoFilter()
   {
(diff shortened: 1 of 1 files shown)


hjmjohnson:
This PR resolves the source of most of the PR failures today on compilers that only look at the immediate parent class.  

Using the Getters and Setters is preferred.

dzenanz (approved):

PR #5945, opened 2026-03-13
COMP: Silence nodiscard warnings for Future.get() in PoolMultiThreader

hjmjohnson:
## Summary
- Cast `Future.get()` return values to `(void)` in `itkPoolMultiThreader.cxx` where the call is only used for its side effect of waiting for thread completion and propagating exceptions

Fixes:
```
ITK/Modules/Core/Common/src/itkPoolMultiThreader.cxx:148:55: warning:
  ignoring return value of function declared with 'nodiscard'
  attribute [-Wunused-result]
    exceptionHandler.TryAndCatch([this, threadLoop] { m_ThreadInfoArray[threadLoop].Future.get(); });
                                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ITK/Modules/Core/Common/src/itkPoolMultiThreader.cxx:303:11: warning:
  ignoring return value of function declared with 'nodiscard'
  attribute [-Wunused-result]
          m_ThreadInfoArray[i].Future.get();
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

## Test plan
- [ ] CI builds clean with no `-Wunused-result` warnings
- [ ] No functional change — `(void)` cast only suppresses the warning

### DIFF
--- a/Modules/Core/Common/src/itkPoolMultiThreader.cxx
+++ b/Modules/Core/Common/src/itkPoolMultiThreader.cxx
@@ -145,7 +145,7 @@ PoolMultiThreader::SingleMethodExecute()
   // so now it waits for each of the other work units to finish
   for (threadLoop = 1; threadLoop < m_NumberOfWorkUnits; ++threadLoop)
   {
-    exceptionHandler.TryAndCatch([this, threadLoop] { m_ThreadInfoArray[threadLoop].Future.get(); });
+    exceptionHandler.TryAndCatch([this, threadLoop] { (void)m_ThreadInfoArray[threadLoop].Future.get(); });
   }
 
   exceptionHandler.RethrowFirstCaughtException();
@@ -300,7 +300,7 @@ PoolMultiThreader::ParallelizeImageRegion(unsigned int         dimension,
               filter->IncrementProgress(0);
             }
           } while (status != std::future_status::ready);
-          m_ThreadInfoArray[i].Future.get();
+          (void)m_ThreadInfoArray[i].Future.get();
           reporter.CompletedPixel();
         });
       }


Modules/Core/Common/src/itkPoolMultiThreader.cxx
```diff
@@ -145,7 +145,7 @@ PoolMultiThreader::SingleMethodExecute()
   // so now it waits for each of the other work units to finish
   for (threadLoop = 1; threadLoop < m_NumberOfWorkUnits; ++threadLoop)
   {
-    exceptionHandler.TryAndCatch([this, threadLoop] { m_ThreadInfoArray[threadLoop].Future.get(); });
+    exceptionHandler.TryAndCatch([this, threadLoop] { (void)m_ThreadInfoArray[threadLoop].Future.get(); });
```
N-Dekker, line 148: What does `Future.get()` return? And why is it OK to ignore its return value?

I'm asking, because maybe we should use the `std::future<void>` specialization, for `Future`: https://en.cppreference.com/w/cpp/thread/future/get.html
`std::future<void>::get()` returns `void`. It looks like that's what we want, right? 

In general, I think it's preferable to avoid unnecessary casting  🤷

hjmjohnson:
```c++
std::future<ITK_THREAD_RETURN_TYPE> Future;
```

ITK_THREAD_RETURN_TYPE depends on the platform.

I don't understand what you are suggesting.  Could you please make an alternate proposal?

N-Dekker: Ah, I see: 

ITK_THREAD_RETURN_TYPE is a pointer type when using Pthreads:
https://github.com/InsightSoftwareConsortium/ITK/blob/badec43941c7c6ed5d87912b376edbffc781569d/Modules/Core/Common/include/itkThreadSupport.h#L47

And it's `unsigned int` when using WIN32 threads:
https://github.com/InsightSoftwareConsortium/ITK/blob/badec43941c7c6ed5d87912b376edbffc781569d/Modules/Core/Common/include/itkThreadSupport.h#L62

But "normally" it's just void:
https://github.com/InsightSoftwareConsortium/ITK/blob/badec43941c7c6ed5d87912b376edbffc781569d/Modules/Core/Common/include/itkThreadSupport.h#L77

Are Pthreads and WIN32 threads both still actively supported nowadays?

PR #5946, opened 2026-03-13
BUG: Use std::forward instead of std::move on forwarding references

hjmjohnson:
## Summary
- Replace `std::move` with `std::forward` on forwarding references (`T&&` where `T` is a deduced template type) in test utilities
- `itkArray2DGTest.cxx`: Fix `std::move` on `decltype(original)` forwarding reference in move-construct test
- `itkRangeGTestUtilities.h`: Fix `std::move` on `TRange&&` forwarding references in `ExpectMoveConstructedRangeHasSameIteratorsAsOriginalBeforeMove` and `ExpectMoveAssignedRangeHasSameIteratorsAsOriginalBeforeMove`

Detected by `bugprone-move-forwarding-reference` clang-tidy check (LLVM 22).

## Test plan
- [ ] `itkArray2DGTest` passes
- [ ] All range GTests pass
- [ ] CI builds clean with no `bugprone-move-forwarding-reference` warnings

### DIFF
--- a/Modules/Core/Common/test/itkArray2DGTest.cxx
+++ b/Modules/Core/Common/test/itkArray2DGTest.cxx
@@ -70,7 +70,7 @@ TEST(Array2D, MoveConstruct)
     const auto * const * const originalDataArray{ original.data_array() };
     const unsigned int         originalSize{ original.size() };
 
-    const auto moveConstructed = std::move(original);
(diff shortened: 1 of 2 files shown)


N-Dekker (changes requested):
These tests are specifically intended to test move semantics. So I believe they should use `std::move`.

PR #5949, opened 2026-03-14
COMP: suppress some third party warings on cdash

seanm:


thewtex (approved):

seanm:
Seems not to have worked today either: https://open.cdash.org/builds/11132842/errors

seanm:
Can do!

I added [here](https://open.cdash.org/builds/11137140/notes):

```
set(CTEST_CUSTOM_WARNING_EXCEPTION
  ${CTEST_CUSTOM_WARNING_EXCEPTION}
  ".*Modules/ThirdParty/.*warning:.*Wzero-as-null-pointer-constant"
  ".*Modules/ThirdParty/Eigen3/.*warning:.*"
  "ld: warning: ignoring duplicate libraries: '-lm'"
  "ranlib: warning: '.*' has no symbols"
  )
```

and that worked. Meaning those regexs suppressed warnings as expected, leaving [just 1](https://open.cdash.org/builds/11137140/errors).

So the regex itself is correct I guess. The question is why doesn't it work in `CMake/CTestCustom.cmake.in`?

PR #5950, opened 2026-03-15
COMP: Let `Future.get()` calls in PoolMultiThreader always return void

N-Dekker:
`Future` was declared as `std::future<ITK_THREAD_RETURN_TYPE>`, which is platform specific, and caused warnings on some platforms, as reported by Hans Johnson (@hjmjohnson) at https://github.com/InsightSoftwareConsortium/ITK/pull/5945 :

    itkPoolMultiThreader.cxx:148:55: warning:
      ignoring return value of function declared with 'nodiscard'
      attribute [-Wunused-result]
        exceptionHandler.TryAndCatch([this, threadLoop] { m_ThreadInfoArray[threadLoop].Future.get(); });
                                                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Addressed by using `std::future<void>` instead.

### DIFF
--- a/Modules/Core/Common/src/itkPoolMultiThreader.cxx
+++ b/Modules/Core/Common/src/itkPoolMultiThreader.cxx
@@ -132,7 +132,8 @@ PoolMultiThreader::SingleMethodExecute()
   {
     m_ThreadInfoArray[threadLoop].UserData = m_SingleData;
     m_ThreadInfoArray[threadLoop].NumberOfWorkUnits = m_NumberOfWorkUnits;
-    m_ThreadInfoArray[threadLoop].Future = m_ThreadPool->AddWork(m_SingleMethod, &m_ThreadInfoArray[threadLoop]);
+    m_ThreadInfoArray[threadLoop].Future = m_ThreadPool->AddWork(
+      [method = m_SingleMethod, threadInfo = &m_ThreadInfoArray[threadLoop]] { method(threadInfo); });
   }
 
   // Now, the parent thread calls this->SingleMethod() itself
@@ -175,8 +176,6 @@ PoolMultiThreader::ParallelizeArray(SizeValueType             firstIndex,
       {
         aFunc(ii);
       }
-      // make this lambda have the same signature as m_SingleMethod
-      return ITK_THREAD_RETURN_DEFAULT_VALUE;
     };
 
     SizeValueType workUnit = 1;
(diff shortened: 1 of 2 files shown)


Modules/Core/Common/src/itkPoolMultiThreader.cxx
```diff
@@ -132,7 +132,8 @@ PoolMultiThreader::SingleMethodExecute()
   {
     m_ThreadInfoArray[threadLoop].UserData = m_SingleData;
     m_ThreadInfoArray[threadLoop].NumberOfWorkUnits = m_NumberOfWorkUnits;
-    m_ThreadInfoArray[threadLoop].Future = m_ThreadPool->AddWork(m_SingleMethod, &m_ThreadInfoArray[threadLoop]);
+    m_ThreadInfoArray[threadLoop].Future = m_ThreadPool->AddWork(
+      [method = m_SingleMethod, threadInfo = &m_ThreadInfoArray[threadLoop]] { method(threadInfo); });
```
dzenanz, line 136: Will this refactoring not add another layer of indirection? Adding a lambda invocation on top of the code in AddWork method? And it seems unnecessary, with respect to the other changes in this commit.

dzenanz (commented):
Mostly looks good.

Modules/Core/Common/src/itkPoolMultiThreader.cxx
N-Dekker: > Will this refactoring not add another layer of indirection?

Yes, but it will be a very lightweight layer. `PoolMultiThreader::ParallelizeImageRegion` does something similar:

https://github.com/InsightSoftwareConsortium/ITK/blob/badec43941c7c6ed5d87912b376edbffc781569d/Modules/Core/Common/src/itkPoolMultiThreader.cxx#L268-L269

> And it seems unnecessary, with respect to the other changes in this commit.

Why do you think it seems unnecessary? `AddWork(function, arguments)` deduces the return type of the function, using it as template argument of `std::future<T>`: https://github.com/InsightSoftwareConsortium/ITK/blob/badec43941c7c6ed5d87912b376edbffc781569d/Modules/Core/Common/include/itkThreadPool.h#L87 
So in order to get `std::future<void>`, the function must return `void`, rather than `ITK_THREAD_RETURN_TYPE`. 

Another approach could be: Adjust `AddWork(function, arguments)` to always return `std::future<void>`. Because it appears that its return value is always ignored anyway  🤷 

Things will become much easier once we drop both Pthreads and Win32 threads, of course!

dzenanz: I meant, why change line 135 at all? Can't it just stay the way it is? It is less code, and less indirection.

N-Dekker: Do you mean:

```c++
m_ThreadInfoArray[threadLoop].Future = m_ThreadPool->AddWork(m_SingleMethod, &m_ThreadInfoArray[threadLoop]);
```
?

When directly using `m_SingleMethod` as the first argument of `AddWork`, `AddWork` will deduce the return type of  `m_SingleMethod` (which is `ITK_THREAD_RETURN_TYPE`), and use it as template argument for `std::future`. So it will then return `std::future<ITK_THREAD_RETURN_TYPE>`. Which is _not_ the same as `std::future<void>`, when using Pthreads or Win32 threads.

Right?

dzenanz: Ah, OK, so this change is necessary, which was not obvious to me.

dzenanz (approved):

hjmjohnson (approved):

PR #5964, opened 2026-03-18
BUG: Remove m_InputFile and CloseFile from OBJMeshIO, use RAII for file

N-Dekker:
The "Read" member functions of OBJMeshIO were not exception-safe, because they would not call `CloseFile()` in case of an exception. So then the input file would not be closed.

Fixed by using local `inputFile` variables, instead of a `m_InputFile` member variable. The `inputFile` variables close the file automatically when going out of scope (RAII), so `CloseFile()` is no longer necessary.

- This problem was found when processing pull request #5956 by Hans Johnson (@hjmjohnson).

### DIFF
--- a/Modules/IO/MeshOBJ/src/itkOBJMeshIO.cxx
+++ b/Modules/IO/MeshOBJ/src/itkOBJMeshIO.cxx
@@ -57,8 +57,8 @@ OBJMeshIO::CanWriteFile(const char * fileName)
   return true;
 }
 
-void
-OBJMeshIO::OpenFile()
+std::ifstream
+OBJMeshIO::OpenFile() const
 {
   if (this->m_FileName.empty())
   {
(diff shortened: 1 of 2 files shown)


Modules/IO/MeshOBJ/src/itkOBJMeshIO.cxx
```diff
@@ -74,26 +74,17 @@ OBJMeshIO::OpenFile()
...
 
   // Test whether the file was opened
-  if (!m_InputFile.is_open())
+  if (!inputFile.is_open())
   {
     itkExceptionMacro("Unable to open file " << this->m_FileName);
   }
 
-  // If not set to start of file stream, windows won't work properly
-  m_InputFile.seekg(0, std::ios::beg);
```
N-Dekker, line 86: I believe that this seek to the begin of the file is unnecessary now, with this PR. I think this problem that "windows won't work properly" only occurred when one and the same file stream (`m_InputFile`) was reopened and reused multiple times. With this PR, a file stream is only used once per "Read" member function call. It is not reopened afterwards anymore.

dzenanz (approved):

PR #5968, opened 2026-03-19
Add `MeshIOBase::OpenInputFile()`, use it in OBJMeshIO, BYUMeshIO, and OFFMeshIO

N-Dekker:
- Replaced `OBJMeshIO::OpenFile()` with `MeshIOBase::OpenInputFile()` 
 - Used `MeshIOBase::OpenInputFile()` in BYUMeshIO to reduced duplicate code
 - Used `MeshIOBase::OpenInputFile()` in OFFMeshIO to increase exception safety, following pull request #5964


N-Dekker:
- @hjmjohnson As this PR is merged, I think both OBJ _and_ OFF mesh IO are exception-safe, so maybe then we can reconsider adding exceptions, as you proposed with PR #5956

PR #5989, opened 2026-03-29
Remove local type aliases of Image Iterators in `*.hxx`, use CTAD

N-Dekker:

    Find what: ^[ ]+ using (\w+) = (Image\w+Iterator<\w+>);\r\n(.*?[^\w])\1
    Filters: itk*.hxx
    [v] Match case
    (*) Regular expression
    [v] . matches newline

Followed by manual replacements of uses of those type aliases.

- Follow-up to pull request #5981

### DIFF
--- a/Modules/Core/Transform/include/itkBSplineDeformableTransform.hxx
+++ b/Modules/Core/Transform/include/itkBSplineDeformableTransform.hxx
@@ -501,7 +501,7 @@ BSplineDeformableTransform<TParametersValueType, VDimension, VSplineOrder>::Tran
   const ParametersValueType *           basePointer = this->m_CoefficientImages[0]->GetBufferPointer();
   for (unsigned int j = 0; j < SpaceDimension; ++j)
   {
-    coeffIterator[j] = ImageScanlineConstIterator<ImageType>(this->m_CoefficientImages[j], supportRegion);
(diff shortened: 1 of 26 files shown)


Modules/Core/Transform/include/itkBSplineDeformableTransform.hxx
```diff
@@ -496,13 +496,12 @@ BSplineDeformableTransform<TParametersValueType, VDimension, VSplineOrder>::Tran
 
   outputPoint.Fill(ScalarType{});
 
-  using IteratorType = ImageScanlineConstIterator<ImageType>;
-  IteratorType                coeffIterator[SpaceDimension];
-  unsigned long               counter = 0;
-  const ParametersValueType * basePointer = this->m_CoefficientImages[0]->GetBufferPointer();
+  ImageScanlineConstIterator<ImageType> coeffIterator[SpaceDimension];
```
dzenanz, line 499: This commit seems to increase verbosity. I think the main point of these typedefs was to reduce verbosity. Can you do some git archeology to check this assumption of mine?

N-Dekker: The first commit might slightly increase verbosity, although it _does_ reduce the number of lines. Fortunately, the second and the third commit reduce verbosity, by using CTAD (class template argument deduction).  😃

PR #5992, opened 2026-03-31
Remove GoToBegin() calls from Image Iterators in hxx files

N-Dekker:
When an Image Iterator is just constructed by `ImageIteratorType(image, region)`, it is already initialized at the begin of the region.

- Follow-up to pull request #4830

### DIFF
--- a/Modules/Filtering/BiasCorrection/include/itkMRIBiasFieldCorrectionFilter.hxx
+++ b/Modules/Filtering/BiasCorrection/include/itkMRIBiasFieldCorrectionFilter.hxx
@@ -85,7 +85,6 @@ MRIBiasEnergyFunction<TImage, TImageMask, TBiasField>::GetValue(const Parameters
     typename TBiasField::SimpleForwardIterator bIter(m_BiasField);
 
     bIter.Begin();
-    iIter.GoToBegin();
     total = 0.0;
 
     // Fastest for full sampling
--- a/Modules/Filtering/DistanceMap/include/itkSignedMaurerDistanceMapImageFilter.hxx
+++ b/Modules/Filtering/DistanceMap/include/itkSignedMaurerDistanceMapImageFilter.hxx
@@ -253,9 +253,6 @@ SignedMaurerDistanceMapImageFilter<TInputImage, TOutputImage>::ThreadedGenerateD
     ImageRegionIterator      Ot(outputPtr, outputRegion);
     ImageRegionConstIterator It(m_InputCache, outputRegion);
 
-    Ot.GoToBegin();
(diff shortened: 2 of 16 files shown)


Modules/Segmentation/LevelSets/include/itkExtensionVelocitiesImageFilter.hxx
```diff
@@ -186,15 +186,6 @@ ExtensionVelocitiesImageFilter<TLevelSet, TAuxValue, VAuxDimension>::GenerateDat
     auxTempIt[k] = ImageRegionIterator(ptr, ptr->GetBufferedRegion());
   }
 
-  inputIt.GoToBegin();
-  outputIt.GoToBegin();
-  tempIt.GoToBegin();
-  for (unsigned int k = 0; k < VAuxDimension; ++k)
-  {
-    auxOutputIt[k].GoToBegin();
```
dzenanz, line 194: When did calls to `GoToBegin()` stop being needed?

Modules/Filtering/BiasCorrection/include/itkMRIBiasFieldCorrectionFilter.hxx
```diff
@@ -85,7 +85,6 @@ MRIBiasEnergyFunction<TImage, TImageMask, TBiasField>::GetValue(const Parameters
     typename TBiasField::SimpleForwardIterator bIter(m_BiasField);
 
     bIter.Begin();
-    iIter.GoToBegin();
```
dzenanz, line 88: Curiosity: how many false positives did this produce?

Modules/Segmentation/LevelSets/include/itkExtensionVelocitiesImageFilter.hxx
N-Dekker: - _In practice_, I don't think `GoToBegin()` was ever needed, for an Image Iterator that was freshly constructed by `T(image, region)`. However, it only became officially documented and tested with pull request #4815, Aug 28, 2024.

Modules/Filtering/BiasCorrection/include/itkMRIBiasFieldCorrectionFilter.hxx
N-Dekker: The find-and-replace (commit b1162f4aa6ce75ea60e90d5224022601b859c64e) had no false positives.

The "find" from b3a3e4b41af5c8ff515d2baa26711225a2d5192b did have a few false positives (six, I think), so I had to process them manually.

PR #5994, opened 2026-03-31
COMP: Migrate OpenCV Video Bridge from legacy C API to C++ API

hjmjohnson:
## Summary

Migrates `Modules/Video/BridgeOpenCV` from the OpenCV legacy C API to the
modern C++ API. The C API (`IplImage*`, `CvCapture*`, `CvVideoWriter*`,
`cvCaptureFromFile`, `cvQueryFrame`, `CV_CAP_PROP_*`, etc.) was deprecated in
OpenCV 2.x, moved to optional headers in OpenCV 3.x, and **removed entirely in
OpenCV 4.0**, causing build failures on any system with OpenCV 4+.

### Changes per commit

**1. Fix `OpenCVBasicTypeBridge` template deduction (OpenCV 3/4)**
- The `cv::Matx<T, VRows, VColumns>` specialization used `unsigned int` template
  parameters but `cv::Matx` uses `int`. Cast `(unsigned int)VRows` in the partial
  specialization to allow the compiler to deduce both sides correctly.

**2. Migrate `OpenCVVideoIO` to C++ API**
- Remove four custom deleter structs (`IplImageDeleter`, `CvCaptureDeleter`, etc.)
  and all `std::unique_ptr` wrappers around C API types.
- Replace member types with RAII C++ value types: `cv::Mat`, `cv::VideoCapture`,
  `cv::VideoWriter` — these manage their own lifetime with no custom cleanup needed.
- Rewrite all methods using the C++ API: `cv::VideoCapture::isOpened()`,
  `operator>>` for frame reads, `cv::VideoWriter::write()`, `cv::cvtColor()`,
  `cv::VideoWriter::fourcc()`.
- Replace `memcpy` with `std::copy_n` for type-safe buffer copy.
- Remove conditional OpenCV 2.4/3.x C header includes; use only
  `opencv2/videoio.hpp` and `opencv2/imgproc.hpp`.

**3. Migrate `OpenCVVideoCapture` and test files to C++ API**
- Replace `IplImage*` frame handling in `retrieve()` with `cv::Mat` non-owning
  constructor wrapping the ITK frame buffer directly (no copy needed).
- Replace `CV_FOURCC(...)` macro with `cv::VideoWriter::fourcc(...)` (×2 in
  constructors).
- Replace all `CV_CAP_PROP_*` macros with `cv::CAP_PROP_*` scoped enums in
  `set()`/`get()` methods.
- In test files: replace `CvCapture*`/`cvCaptureFromCAM`/`cvReleaseCapture` with
  `cv::VideoCapture`; replace all remaining `CV_CAP_PROP_*`/`CV_FOURCC` macro
  usage; remove legacy C headers (`videoio_c.h`, `imgproc_c.h`, `highgui.h`).

## Backward Compatibility

All C++ API replacements used have been available since **OpenCV 2.0** or earlier.
This PR does **not** raise the minimum required OpenCV version.

| Legacy C API (removed in OpenCV 4) | Modern C++ Replacement | Available Since |
|---|---|---|
| `IplImage*` / `cvCreateImageHeader` | `cv::Mat(rows, cols, type, ptr)` | OpenCV 1.x (C++ wrappers) |
| `CvCapture*` / `cvCaptureFromFile` | `cv::VideoCapture(filename)` | OpenCV 2.0 |
| `CvCapture*` / `cvCaptureFromCAM` | `cv::VideoCapture(index)` | OpenCV 2.0 |
| `cvQueryFrame` / `cvGetNextImage` | `cv::VideoCapture::operator>>` | OpenCV 2.0 |
| `cvGetCaptureProperty` | `cv::VideoCapture::get(propId)` | OpenCV 2.0 |
| `cvSetCaptureProperty` | `cv::VideoCapture::set(propId, val)` | OpenCV 2.0 |
| `CvVideoWriter*` / `cvCreateVideoWriter` | `cv::VideoWriter(file, fourcc, fps, size)` | OpenCV 2.0 |
| `cvWriteFrame` | `cv::VideoWriter::write(frame)` | OpenCV 2.0 |
| `cvReleaseCapture` / `cvReleaseVideoWriter` | RAII (destructor / `.release()`) | OpenCV 2.0 |
| `cvCvtColor` / `CV_BGR2RGB` | `cv::cvtColor` / `cv::COLOR_BGR2RGB` | OpenCV 2.0 |
| `CV_FOURCC('M','P','4','2')` macro | `cv::VideoWriter::fourcc('M','P','4','2')` | OpenCV 2.1 |
| `CV_CAP_PROP_POS_FRAMES` macro | `cv::CAP_PROP_POS_FRAMES` scoped enum | OpenCV 3.0 |
| `CV_CAP_PROP_FPS` macro | `cv::CAP_PROP_FPS` scoped enum | OpenCV 3.0 |
| `CV_CAP_PROP_FOURCC` macro | `cv::CAP_PROP_FOURCC` scoped enum | OpenCV 3.0 |
| `CV_CAP_PROP_FRAME_WIDTH/HEIGHT` macros | `cv::CAP_PROP_FRAME_WIDTH/HEIGHT` enums | OpenCV 3.0 |
| `CV_CAP_PROP_POS_AVI_RATIO` macro | `cv::CAP_PROP_POS_AVI_RATIO` scoped enum | OpenCV 3.0 |
| `CV_CAP_PROP_POS_MSEC` macro | `cv::CAP_PROP_POS_MSEC` scoped enum | OpenCV 3.0 |

> **Note:** The `cv::CAP_PROP_*` scoped enums were introduced in OpenCV 3.0.
> The `CV_CAP_PROP_*` macro aliases for them still exist in OpenCV 3.x (via
> `videoio_c.h`) but were removed in OpenCV 4.0. Since ITK already requires
> OpenCV 3+ for other reasons, using `cv::CAP_PROP_*` directly is safe.

## Test plan
- [ ] Verify CI builds pass on Linux, macOS, Windows with OpenCV 4.x
- [x] Verify all OpenCV Video Bridge tests pass (`ctest -R OpenCV`)

### DIFF
--- a/Modules/Video/BridgeOpenCV/include/itkOpenCVBasicTypeBridge.h
+++ b/Modules/Video/BridgeOpenCV/include/itkOpenCVBasicTypeBridge.h
@@ -161,10 +161,14 @@ struct OpenCVBasicTypeBridge<itk::Size<2>, cv::Size>
   }
 };
 
-template <typename T, unsigned int VRows, unsigned int VColumns>
-struct OpenCVBasicTypeBridge<itk::Matrix<T, VRows, VColumns>, cv::Matx<T, VRows, VColumns>>
+// cv::Matx uses int template params; itk::Matrix uses unsigned int.
+// Using int for VRows/VColumns (matching OpenCV) with a cast in the ITK
+// pattern allows the compiler to deduce both sides correctly.
+template <typename T, int VRows, int VColumns>
+struct OpenCVBasicTypeBridge<itk::Matrix<T, static_cast<unsigned int>(VRows), static_cast<unsigned int>(VColumns)>,
+                             cv::Matx<T, VRows, VColumns>>
 {
-  using ITKDataType = itk::Matrix<T, VRows, VColumns>;
+  using ITKDataType = itk::Matrix<T, static_cast<unsigned int>(VRows), static_cast<unsigned int>(VColumns)>;
   using OpenCVDataType = cv::Matx<T, VRows, VColumns>;
 
   static ITKDataType
--- a/Modules/Video/BridgeOpenCV/include/itkOpenCVVideoCapture.hxx
+++ b/Modules/Video/BridgeOpenCV/include/itkOpenCVVideoCapture.hxx
@@ -20,18 +20,8 @@
 
 #include "itkNumericTraits.h"
 
-#include "opencv2/core/version.hpp"
-#if !defined(CV_VERSION_EPOCH) // OpenCV >= 3.0
-#  include "opencv2/videoio.hpp"
-#  include "opencv2/imgproc/types_c.h" // CV_RGB2BGR, CV_BGR2GRAY, ...
-#  include "opencv2/imgproc/imgproc_c.h"
-#  if CV_VERSION_MAJOR > 3 // OpenCV 4.x
-#    include "opencv2/videoio/legacy/constants_c.h"
-#  endif
-#else
-// OpenCV 2.4.x
-#  include "cv.h"
-#endif
+#include "opencv2/videoio.hpp"
+#include "opencv2/imgproc.hpp"
 
 namespace itk
 {
@@ -50,7 +40,7 @@ OpenCVVideoCapture<TVideoStream>::OpenCVVideoCapture()
   m_FpS = 24;
 
   // Default to mp4.2
-  m_FourCC = CV_FOURCC('M', 'P', '4', '2');
+  m_FourCC = cv::VideoWriter::fourcc('M', 'P', '4', '2');
 }
 
 //
@@ -65,7 +55,7 @@ OpenCVVideoCapture<TVideoStream>::OpenCVVideoCapture(TVideoStream * videoStream)
   m_FpS = 24;
 
   // Default to mp4.2
-  m_FourCC = CV_FOURCC('M', 'P', '4', '2');
+  m_FourCC = cv::VideoWriter::fourcc('M', 'P', '4', '2');
 }
 
 
@@ -121,7 +111,7 @@ OpenCVVideoCapture<TVideoStream>::grab()
 
   // Move the current frame forward by 1
   SizeValueType currentFrame = m_VideoStream->GetRequestedTemporalRegion().GetFrameStart();
-  return set(CV_CAP_PROP_POS_FRAMES, currentFrame + 1);
+  return set(cv::CAP_PROP_POS_FRAMES, currentFrame + 1);
 }
 
 //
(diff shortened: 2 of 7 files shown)


Modules/Video/BridgeOpenCV/include/itkOpenCVBasicTypeBridge.h
```diff
@@ -161,10 +161,13 @@ struct OpenCVBasicTypeBridge<itk::Size<2>, cv::Size>
...
-template <typename T, unsigned int VRows, unsigned int VColumns>
-struct OpenCVBasicTypeBridge<itk::Matrix<T, VRows, VColumns>, cv::Matx<T, VRows, VColumns>>
+// cv::Matx uses int template params; itk::Matrix uses unsigned int.
+// Using int for VRows/VColumns (matching OpenCV) with a cast in the ITK
+// pattern allows the compiler to deduce both sides correctly.
+template <typename T, int VRows, int VColumns>
+struct OpenCVBasicTypeBridge<itk::Matrix<T, (unsigned int)VRows, (unsigned int)VColumns>, cv::Matx<T, VRows, VColumns>>
 {
-  using ITKDataType = itk::Matrix<T, VRows, VColumns>;
+  using ITKDataType = itk::Matrix<T, (unsigned int)VRows, (unsigned int)VColumns>;
```
N-Dekker: Please don't use C-style cast! Is an explicit conversion really necessary?  (If so, you may consider `unsigned{VRows}` and `unsigned{VColumns}`, which are 100% type safe.)

hjmjohnson: Good catch — replaced both C-style casts with `static_cast<unsigned int>(VRows)` and `static_cast<unsigned int>(VColumns)`. The explicit conversion is necessary here: the template parameter `VRows` is deduced as `int` (to match `cv::Matx`), but `itk::Matrix` requires `unsigned int`, so a cast in the specialization pattern is required for deduction to succeed. Squashed into the same commit.

N-Dekker: Thanks, but are you sure that the explicit cast is necessary? Do you get a warning, otherwise?

`itk::Matrix` is declared as follows:

    template <typename T, unsigned int VRows = 3, unsigned int VColumns = 3> 
    class ITK_TEMPLATE_EXPORT Matrix 

https://github.com/InsightSoftwareConsortium/ITK/blob/2b7e6d4f1d54d6775c9122c1e0a02f38cc2d8538/Modules/Core/Common/include/itkMatrix.h#L51-L52

I think it should be perfectly fine to pass a _signed_ integer as argument for VRows  or VColumns, for example `itk::Matrix<float, 2, 2>`. (`2` is a _signed_ int.)

hjmjohnson:
@dzenanz This looks like perhaps it is something that supports Kitware projects.  Do you know a good candidate to review this PR?

Modules/Video/BridgeOpenCV/include/itkOpenCVBasicTypeBridge.h
hjmjohnson: I did get a failure that this fixed.  It is hard to test this code in all scenarios, so I figured that being explicit is a "won't hurt" condition.

dzenanz (approved):
Looks good. I have some program which uses `Video\BridgeOpenCV\include\itkOpenCVBasicTypeBridge.h`, and it keeps working after these changes.

Modules/Video/BridgeOpenCV/include/itkOpenCVBasicTypeBridge.h
N-Dekker: Well, there is a guideline: [Avoid casts](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#res-casts)

So I would only add static_cast when it is really necessary  🤷 That's why I suggested to just remove the cast. Or otherwise use modern `T{ x }` conversion.

